PR #2192, opened 2024-09-23
[Issue #1868] Enforce absolute imports outside of the same directory

### LINKED ISSUE #1868: [Task]: Enforce absolute imports for frontend
### Summary

The frontend is currently using a mix of relative and absolute imports. This task is to update the frontend imports to use absolute paths. Additionally we should [add a relative path setting of `src/`](https://nextjs.org/docs/app/building-your-application/configuring/absolute-imports-and-module-aliases#module-aliases) so that absolute imports become `component/Footer` instead of `src/components/Footer`.

If there is a reason not to use absolute imports, there should at least be a consistent approach across the frontend, and this ticket can be used to document that.

### Acceptance criteria

- [ ] config updated and absolute import paths enforced
- [ ] check if we can add this to the linter 
- [ ] update docs if deemed necessary (if there is a linter than that could be considered self-documenting)

acouch:
## Summary
Fixes #1868

## Changes proposed
Enforces absolute imports outside of the same directory.

### DIFF
--- a/frontend/.eslintrc.js
+++ b/frontend/.eslintrc.js
@@ -14,6 +14,17 @@ module.exports = {
     // dependencies to work in standalone mode. It may be overkill for most projects at
     // Nava which aren't image heavy.
     "@next/next/no-img-element": "off",
+    "no-restricted-imports": [
+      "error",
+      {
+        "patterns": [
+          {
+            "group": ["../"],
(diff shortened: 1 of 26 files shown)


andycochran (dismissed):
So much more legible 👏

mdragon (approved):
Looks good!

Issue #1937, opened 2024-05-06
[Task]: Update Loading spinner "Loading results..." to use translation utility


rylew1:
### Summary

Right now the loading spinner work was done in parallel with the `next-intl` translation work and therefore didn't use the `t()` function with a translation string.

acouch:
The component has moved but the translation still not addressed: https://github.com/HHS/simpler-grants-gov/blob/main/frontend/src/app/%5Blocale%5D/search/loading.tsx#L11

PR #2202, opened 2024-09-25
[Issue #1937] Comply with new linter rules about import paths

mdragon:
## Summary
Follows #1937

## Changes proposed
Fix import path to not fail new linter rule banning relative paths


doug-s-nava (approved):
✅ ✅ ✅ ✅ ✅ ✅

PR #2229, opened 2024-09-26
Change reference location in api object for agency name on search

emilycnava:
## Summary
Fixes #2172 

## Changes proposed
I changed the render logic to pull agency name from the root of the object as specified in the original issue. I chose to leave the agency code lookup as a fallback. Open to removing that fall back if we think that'd make more sense? Thoughts r.e. leaving agency code lookup as a backup @acouch ?

## Context for reviewers
Make sure search results still have agency results where expected locally.

## Additional information
Screenshot 2024-09-26 at 11 55 14 AM

PR #2230, opened 2024-09-26
[Issue #2222] Account for empty summaries

### LINKED ISSUE #2222: [Task]: Fix API version and Rendering Empty Data on the Opportunity Listing Page
### Summary

The current opportunity listing page is using the wrong API version. It also renders some nested fields without checking so see if there is data which causes some errors. This task is to fix that.

### Acceptance criteria

- [x] correct API version used
- [x] empty data fields don't cause issues

acouch:
## Summary
Fixes #2222

## Changes proposed
* Account for empty summaries which we can expect from the API
* Fix some of the type definitions for API fields that can be null
* bonus: Add title of the opp to the page
  * note that next encourages multiple fetch (only GET) in the same component tree: https://nextjs.org/docs/app/building-your-application/caching
* bonus: Fix lack of gap with the description:

  image
  becomes:
  
image
 
* bonus: search breadcrumbs updated to link from `/search` to `/search?status=forecasted,posted`

### DIFF
--- a/frontend/src/app/api/OpportunityListingAPI.ts
+++ b/frontend/src/app/api/OpportunityListingAPI.ts
@@ -7,7 +7,7 @@ import BaseApi from "./BaseApi";
 
 export default class OpportunityListingAPI extends BaseApi {
   get version(): string {
-    return "v0.1";
+    return "v1";
   }
 
   get basePath(): string {
--- a/frontend/src/app/api/OpportunityListingAPI.ts
+++ b/frontend/src/app/api/OpportunityListingAPI.ts
@@ -7,7 +7,7 @@ import BaseApi from "./BaseApi";
 
 export default class OpportunityListingAPI extends BaseApi {
   get version(): string {
-    return "v1";
+    return "v0.1";
   }
 
   get basePath(): string {
--- a/frontend/src/components/opportunity/OpportunityDescription.tsx
+++ b/frontend/src/components/opportunity/OpportunityDescription.tsx
@@ -59,9 +59,7 @@ const OpportunityDescription = ({ opportunityData }: Props) => {
         <div
           dangerouslySetInnerHTML={{
             __html: DOMPurify.sanitize(
-              opportunityData.summary.summary_description
-                ? opportunityData.summary.summary_description
-                : "",
+              opportunityData.summary.summary_description ?? "",
             ),
           }}
         />
--- a/frontend/src/components/opportunity/OpportunityDescription.tsx
+++ b/frontend/src/components/opportunity/OpportunityDescription.tsx
@@ -87,7 +87,7 @@ const OpportunityDescription = ({ opportunityData }: Props) => {
             ),
           }}
         />
-        <p>{opportunityData.summary.agency_email_address}</p>
+        <p>{agency_email}</p>
         <p>
           <a href={`mailto:${agency_email}`}>
             {opportunityData.summary.agency_email_address_description}
--- a/frontend/src/components/opportunity/OpportunityDescription.tsx
+++ b/frontend/src/components/opportunity/OpportunityDescription.tsx
@@ -53,6 +53,9 @@ const OpportunityDescription = ({ opportunityData }: Props) => {
     .applicant_eligibility_description
     ? opportunityData.summary.applicant_eligibility_description
     : "--";
+  const agency_email = opportunityData.summary.agency_email_address
+    ? opportunityData.summary.agency_email_address
+    : "";
   return (
     <>
       <div className="usa-prose">
@@ -60,7 +63,9 @@ const OpportunityDescription = ({ opportunityData }: Props) => {
         <div
           dangerouslySetInnerHTML={{
             __html: DOMPurify.sanitize(
-              opportunityData.summary.summary_description,
+              opportunityData.summary.summary_description
+                ? opportunityData.summary.summary_description
+                : "",
             ),
           }}
         />
@@ -76,13 +81,15 @@ const OpportunityDescription = ({ opportunityData }: Props) => {
         <div
           dangerouslySetInnerHTML={{
             __html: DOMPurify.sanitize(
-              opportunityData.summary.agency_contact_description,
+              opportunityData.summary.agency_contact_description
+                ? opportunityData.summary.agency_contact_description
+                : "",
             ),
           }}
         />
         <p>{opportunityData.summary.agency_email_address}</p>
         <p>
-          <a href={`mailto:${opportunityData.summary.agency_email_address}`}>
+          <a href={`mailto:${agency_email}`}>
             {opportunityData.summary.agency_email_address_description}
           </a>
         </p>
--- a/frontend/src/components/opportunity/OpportunityDescription.tsx
+++ b/frontend/src/components/opportunity/OpportunityDescription.tsx
@@ -49,13 +49,9 @@ const OpportunityDescription = ({ opportunityData }: Props) => {
     ? opportunityData.summary.agency_phone_number.replace(/-/g, "")
     : "";
 
-  const additionalInformationOnEligibility = opportunityData.summary
-    .applicant_eligibility_description
-    ? opportunityData.summary.applicant_eligibility_description
-    : "--";
-  const agency_email = opportunityData.summary.agency_email_address
-    ? opportunityData.summary.agency_email_address
-    : "";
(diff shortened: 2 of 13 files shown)


frontend/src/app/api/OpportunityListingAPI.ts
```diff
@@ -7,7 +7,7 @@ import BaseApi from "./BaseApi";
 
 export default class OpportunityListingAPI extends BaseApi {
   get version(): string {
-    return "v1";
+    return "v0.1";
```
chouinar, line 10: I'm not sure if you want to switch this - as soon as the API deploys are fixed, you'll want to be on v1 as several fields are missing on v0.1.

acouch: @chouinar misunderstood what you said in slack. Updating.

acouch, line 10:
```suggestion
    return "v1";
```

frontend/src/components/opportunity/OpportunityDescription.tsx
```diff
@@ -76,13 +81,15 @@ const OpportunityDescription = ({ opportunityData }: Props) => {
...
           dangerouslySetInnerHTML={{
             __html: DOMPurify.sanitize(
-              opportunityData.summary.agency_contact_description,
+              opportunityData.summary.agency_contact_description
+                ? opportunityData.summary.agency_contact_description
+                : "",
             ),
           }}
         />
         <p>{opportunityData.summary.agency_email_address}</p>
```
emilycnava, line 90: Should this also be changed to `{agency_email}`

frontend/src/components/opportunity/OpportunityIntro.tsx
```diff
@@ -14,10 +14,9 @@ type Props = {
 const OpportunityIntro = ({ opportunityData }: Props) => {
   const t = useTranslations("OpportunityListing.intro");
 
-  const agencyName =
-    opportunityData.summary.agency_name === ""
-      ? "--"
-      : opportunityData.summary.agency_name;
+  const agencyName = opportunityData.summary.agency_name
```
emilycnava, line 17: Similar to https://github.com/HHS/simpler-grants-gov/pull/2229 should we change these references to use the root level agency name not the summary?

emilycnava (changes requested):

frontend/src/components/opportunity/OpportunityDescription.tsx
acouch: Updated, thanks.

frontend/src/components/opportunity/OpportunityDescription.tsx
```diff
@@ -53,14 +53,19 @@ const OpportunityDescription = ({ opportunityData }: Props) => {
     .applicant_eligibility_description
     ? opportunityData.summary.applicant_eligibility_description
     : "--";
+  const agency_email = opportunityData.summary.agency_email_address
```
doug-s-nava, line 56: seems like we can use nullish coalescing (`??`) in all of these defaulting situations, slightly legible, less verbose

frontend/tests/components/opportunity/OpportunityDescription.test.tsx
```diff
@@ -59,7 +59,9 @@ describe("OpportunityDescription", () => {
     expect(descriptionHeading).toBeInTheDocument();
 
     const sanitizedSummaryDescription = DOMPurify.sanitize(
-      mockOpportunityData.summary.summary_description,
+      mockOpportunityData.summary.summary_description
```
doug-s-nava, line 62: why do we have to do a check here (and elsewhere in the file), aren't we supplying the mock data directly to ensure that we we don't pass incomplete data in these cases?

frontend/tests/components/opportunity/OpportunityDescription.test.tsx
```diff
@@ -108,17 +112,19 @@ describe("OpportunityDescription", () => {
...
+        : "",
     });
-    expect(telLink).toHaveAttribute(
-      "href",
-      `tel:${mockOpportunityData.summary.agency_phone_number.replace(/-/g, "")}`,
-    );
+    const number = mockOpportunityData.summary.agency_phone_number
+      ? mockOpportunityData.summary.agency_phone_number.replace(/-/g, "")
+      : "";
+    expect(telLink).toHaveAttribute("href", `tel:${number}`);
```
doug-s-nava, line 128: do we want to add a test supplying null data to make sure the defaulting / null checking behavior is working right?

doug-s-nava (commented):
have not manually tested, but here are some thoughts on the code. Any suggestions on manual test steps? Just updating an opportunity manually to remove the summary data and making sure it displays without errors?

frontend/tests/components/opportunity/OpportunityDescription.test.tsx
acouch: @doug-s-nava the sanitize function won't take a null value.

doug-s-nava: Looking more closely I'm a little confused about what we're testing here, would love to have a quick call to straighten it out

acouch:
@doug-s-nava 

> Any suggestions on manual test steps? Just updating an opportunity manually to remove the summary data and making sure it displays without errors?

You can view the opps from the search page.

frontend/tests/components/opportunity/OpportunityDescription.test.tsx
acouch: Good question. I'd like to punt this to https://github.com/HHS/simpler-grants-gov/issues/2321 as I'm not sure the correct pattern.

frontend/tests/components/opportunity/OpportunityDescription.test.tsx
acouch: I updated this, thanks.

frontend/src/components/opportunity/OpportunityDescription.tsx
acouch: I updated these, thanks.

frontend/src/components/opportunity/OpportunityDescription.tsx
```diff
@@ -49,18 +49,19 @@ const OpportunityDescription = ({ opportunityData }: Props) => {
...
+  const agency_email = opportunityData.summary.agency_email_address ?? "";
   return (
     <>
       <div className="usa-prose">
         <h2>{t("description")}</h2>
         <div
           dangerouslySetInnerHTML={{
             __html: DOMPurify.sanitize(
-              opportunityData.summary.summary_description,
+              opportunityData.summary.summary_description
```
doug-s-nava, line 62: I think you can use `??` her, line 80, and in OpportunityIntro as well

frontend/tests/components/opportunity/OpportunityDescription.test.tsx
```diff
@@ -87,7 +87,9 @@ describe("OpportunityDescription", () => {
     expect(additionalInfoHeading).toBeInTheDocument();
 
     const sanitizedEligibilityDescription = DOMPurify.sanitize(
-      mockOpportunityData.summary.applicant_eligibility_description,
+      mockOpportunityData.summary.applicant_eligibility_description
```
doug-s-nava, line 90: I know these tests are not the point of this change, so feel free to ignore at this point but after thinking about this some more, I think it would be best to just remove the calls to `sanitize` on lines 61 and 89 and leave the tests the same otherwise. The tests still pass, as the values we're sending in in the mock do not change when run through the sanitize function, as it is being mocked to be just an identity.

doug-s-nava (commented):
a couple of nits remaining, but verified locally by trying to access an opportunity detail for an opportunity with no associated summary that this fixes the bug, so going to go ahead and approve

✅ ✅ ✅ ✅ ✅

frontend/src/components/opportunity/OpportunityDescription.tsx
```diff
@@ -49,18 +49,19 @@ const OpportunityDescription = ({ opportunityData }: Props) => {
...
+                ? opportunityData.summary.summary_description
+                : "",
```
acouch, line 64:
```suggestion
              opportunityData.summary.summary_description ?? "",
```

doug-s-nava (approved):

PR #2238, opened 2024-09-26
[Issue #1563] Ability to hide search filters in mobile viewport

### LINKED ISSUE #1563: Collapse sidebar/filters on small screens (mobile view)
### Summary

On small screens, the search sidebar (status, filters) should be collapsed by default, as to not take up so much screen real estate. A "Show/Hide filters" button will toggle this state. 

image

### Open Q: 
Should this also include a count (as the accordions/sections do) of all enabled filters, so that it's obvious how many filters are applied while the sidebar is collapsed? 

image

### Acceptance criteria

- [ ] User can toggle the visibility of the filters

doug-s-nava:
## Summary
Fixes #1563

Todos:
- [ ] refactor to reduce duplication of search page on error page (see https://github.com/HHS/simpler-grants-gov/issues/2260)

## Changes proposed

Adds logic for hiding / showing filter options on search page such that:

* on viewports larger than mobile
    * no change
* on mobile viewports
    * all filter options are hidden by default
    * a button appears to allow toggling of appearance of filters
    * toggle button icon changes from up to down arrow when toggled

### Technical changes

* creates ContentDisplayToggle to handle toggle functionality
* creates SearchFilters component to isolate client side state control
* refactors to search page and additional styling to allow for this change
* supporting tests and story

Also:

* adds clsx package
* updates eslint rules for no-unused-args to allow for underscore prefixed args to be unused

## Context for reviewers

To make this change the way that seemed best for me, I made a number of related but perhaps not entirely necessary changes. I'm happy to discuss any of these change further and take a different approach

### Test steps

1. visit your local search page with desktop width viewport
2. _VALIDATE_: all filter options (accordions and checkboxes) appear as normal
3. shrink to mobile size viewport (ex 375px)
4. _VALIDATE_: all filter options are hidden
5. _VALIDATE_: filter toggle button appears matching designs from ticket
6. click filter toggle button
7. _VALIDATE_: filters appear
8. click filter toggle button
7. _VALIDATE_: filters disappear
8. resize to desktop viewport size
7. _VALIDATE_: filters appear 

## Additional information

https://github.com/user-attachments/assets/0fba089b-2057-4cf5-87b9-0ae410f5b3d2


frontend/.eslintrc.js
```diff
@@ -57,7 +57,10 @@ module.exports = {
       rules: {
         camelcase: "off",
         // Prevent dead code accumulation
-        "@typescript-eslint/no-unused-vars": "error",
+        "@typescript-eslint/no-unused-vars": [
+          "error",
+          { argsIgnorePattern: "^_" },
```
doug-s-nava, line 62: I've always found this to be a helpful pattern for marking unused variables that are expected, but we can discuss as a team

frontend/package.json
```diff
@@ -30,6 +30,7 @@
     "@opentelemetry/api": "^1.8.0",
     "@trussworks/react-uswds": "^7.0.0",
     "@uswds/uswds": "^3.6.0",
+    "clsx": "^2.1.1",
```
doug-s-nava, line 33: a very small library used for applying conditional logic for class assignments

frontend/src/app/[locale]/search/page.tsx
```diff
@@ -71,6 +72,35 @@ function Search({ searchParams }: { searchParams: searchParamsTypes }) {
   const pager1key = Object.entries(searchParams).join("-") + "pager1";
   const pager2key = Object.entries(searchParams).join("-") + "pager2";
 
+  const filterConfigurations = useMemo((): SearchFilterConfiguration[] => {
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/search/page.tsx:75]: refactored to dry up the accordion implementation and hopefully improve extensibility if we need it

frontend/src/components/search/SearchFilters.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+import { useState } from "react";
+
+import ContentDisplayToggle from "src/components/ContentDisplayToggle";
+import SearchFilterAccordion, {
+  SearchFilterConfiguration,
+} from "src/components/search/SearchFilterAccordion/SearchFilterAccordion";
+import SearchOpportunityStatus from "src/components/search/SearchOpportunityStatus";
+
+export default function SearchFilters({
```
doug-s-nava, line 13: split out largely so that we can avoid making the full page component a client component when we add state for controlling the toggle

frontend/stories/components/ContentDisplayToggle.stories.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+import ContentDisplayToggle from "src/components/ContentDisplayToggle";
+
+const meta: Meta<typeof ContentDisplayToggle> = {
+  title: "Components/ContentDisplayToggle",
+  component: ContentDisplayToggle,
+};
+export default meta;
+
+export const Default = {
```
doug-s-nava, line 12: this is not fully interactive, would require adding some add ons. Do we want to explore that? https://storybook.js.org/docs/addons/addons-api#useargs

frontend/src/app/[locale]/search/page.tsx
```diff
@@ -52,6 +52,7 @@ interface searchParamsTypes {
 function Search({ searchParams }: { searchParams: searchParamsTypes }) {
   unstable_setRequestLocale("en");
   const t = useTranslations("Search");
+
   const convertedSearchParams = convertSearchParamsToProperTypes(searchParams);
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/search/page.tsx:56]: can this moved into the searchFilters?

frontend/src/styles/_content-display-toggle.scss
andycochran, line 1: Can we use utility classes instead of custom CSS?

frontend/src/components/ContentDisplayToggle.tsx
```diff
@@ -0,0 +1,32 @@
+import { Dispatch, SetStateAction } from "react";
+
+import { USWDSIcon } from "src/components/USWDSIcon";
+
+// this could implemented slightly differently to either be simpler and more generic
+// or we could opt to build the entire show / hide functionality into this component and pass the content controlled by the toggle as children
```
andycochran, line 6: USWDS has an accordion component for the more generic use case of toggling content display. This particular component needs to be more unique, so I'd name/build it with specificity — e.g. it's a toggle for the mobile sidebar layout.

frontend/src/components/search/SearchFilters.tsx
```diff
@@ -0,0 +1,63 @@
...
+          title={title}
+          key={title}
+        />
+      );
+    },
+  );
+  return (
+    <>
+      <div className="grid-row flex-wrap grants-search-filter-toggle">
+        <div className="grid-col-4" />
```
andycochran, line 44: What's the empty div for?

doug-s-nava: I used flexbox with an empty div to center the toggle. If the empty div is a concern I can take a different approach

frontend/src/components/ContentDisplayToggle.tsx
doug-s-nava: My hope was that this component could be reused in other scenarios, since all it's doing is displaying the toggle button and calling the toggle function that is passed in. If we don't want this to be a reusable component I can revert that though, and we could probably delete the story as well.

frontend/src/components/search/SearchFilters.tsx
```diff
@@ -0,0 +1,63 @@
...
+          setToggledContentVisible={setShowFilterOptions}
+          toggledContentVisible={showFilterOptions}
+          toggleText={showFilterOptions ? "Hide Filters" : "Show Filters"}
+        />
+      </div>
+      <div
+        data-testid="search-filters"
+        className={clsx({
+          "grants-search-filter-options": true,
+          "grants-search-filter-options-display-mobile": !!showFilterOptions,
```
andycochran, line 55: Use the [built-in utility class](https://designsystem.digital.gov/utilities/display/)

```suggestion
          "display-block": !!showFilterOptions,
```

andycochran: I think you can just use… 
```
"display-none tablet:display-block" : !showFilterOptions 
```
…and nix the custom CSS. I don't think you need the `!!showFilterOptions`

frontend/src/components/search/SearchFilters.tsx
andycochran: There are utility classes you might use to do this w/o the extra markup: https://designsystem.digital.gov/utilities/flex/

frontend/src/app/[locale]/search/page.tsx
Comment (acouch) [reply frontend/src/app/[locale]/search/page.tsx:74]: Don't feel strongly, but this feels less extensible and could make it harder to change in the future, ie if another component type is added among the filters, but uses the same amount of code.

frontend/src/styles/_content-display-toggle.scss
doug-s-nava: Yes, yes we can

frontend/stories/components/ContentDisplayToggle.stories.tsx
emilycnava: I wonder if we should make a followup ticket to document this idea, so we can discuss in eng sync and possibly explore it in the future?

frontend/.eslintrc.js
emilycnava: I have no issue with this

frontend/src/app/[locale]/search/page.tsx
Comment (emilycnava) [reply frontend/src/app/[locale]/search/page.tsx:74]: Impartial, although given what we were talking about this morning with how next caches data fetches, I wonder if we could make a data fetch call inside SearchFilters with no performance cost as that same call would also be made in this component, this could yield to passing less things in?

frontend/src/components/ContentDisplayToggle.tsx
```diff
@@ -0,0 +1,29 @@
...
+  toggleText,
+}: {
+  setToggledContentVisible: Dispatch<SetStateAction<boolean>>;
+  toggledContentVisible: boolean;
+  toggleText: string;
+}) {
+  const iconName = toggledContentVisible ? "arrow_drop_up" : "arrow_drop_down";
+  return (
+    <div data-testid="content-display-toggle" className="grid-col-4">
+      <button
```
emilycnava, line 19: +1 on using an unstyled uswds button instead of an anchor tag, I know we discussed that yesterday

emilycnava (commented):
Left a few comments, don't think any of them are showstoppers, agree with what a lot of Andy noted too.

andycochran (dismissed):
LGTM!

frontend/src/app/[locale]/search/page.tsx
Comment (doug-s-nava) [reply frontend/src/app/[locale]/search/page.tsx:74]: I think a better way to do it might be to move any functionality related to the search params out of `Search` and into lower level components, and just take minor hit on processing the params multiple times. That would be slightly bigger refactor with a wider surface area though.

For now, I could just move the search param logic for the filter accordions into SearchFilters, and within that file keep things looking more like they do now? Or I could just pass through the necessary param related info rather than the config idea. Open to whatever.

frontend/stories/components/ContentDisplayToggle.stories.tsx
doug-s-nava: https://github.com/HHS/simpler-grants-gov/issues/2331

frontend/src/app/[locale]/search/page.tsx
Comment (doug-s-nava) [reply frontend/src/app/[locale]/search/page.tsx]: pushed a commit with another approach, and the only thing I don't like is the sheer number of props that search filters now needs to take (9). If we're ok with that then this could work? Have a couple of ideas percolating, but this client / server boundary stuff is adding a wrench to my usual thinking on this stuff

acouch (dismissed):
Amazing. This could be part of a larger component library.

frontend/src/components/ContentDisplayToggle.tsx
```diff
@@ -0,0 +1,54 @@
...
+  visibleClassname: string;
+  children: React.ReactNode;
+}) {
+  const [toggledContentVisible, setToggledContentVisible] =
+    useState<boolean>(false);
+
+  const iconName = toggledContentVisible ? "arrow_drop_up" : "arrow_drop_down";
+  return (
+    <>
+      <div className="display-flex flex-column flex-align-center tablet:display-none">
```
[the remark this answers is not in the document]
andycochran: Good idea! You'd always need the same breakpoint to show the content and hide the toggle. Instead of `visibleClassname="tablet:display-block"`, you could pass `breakpoint="tablet"` and concat the display classes — e.g. `${breakpoint}:display-none` and `${breakpoint}:display-block`.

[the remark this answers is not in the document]
doug-s-nava: yep, I definitely missed that. I'll rework

acouch (approved):

PR #2247, opened 2024-09-27
[Issue 2498] update frontend readme for current file structure

doug-s-nava:
## Summary
Issue #2498

## Changes proposed
frontend README updated for current file structure

### Test Steps

1. read the updated README
2. _VALIDATE_: in relation to the changes made, no important information is missing and the information that's there is correct and informative


frontend/README.md
```diff
@@ -13,25 +13,22 @@
...
 │   └── styles        # Sass & design system settings
 ├── stories           # Storybook pages
 └── tests
 ```
 
 ## 💻 Development
 
-[Next.js](https://nextjs.org/docs) provides the React framework for building the web application. Pages are defined in the `pages/` directory. Pages are automatically routed based on the file name. For example, `pages/index.tsx` is the home page.
-
-Files in the `pages/api` are treated as [API routes](https://nextjs.org/docs/api-routes/introduction). An example can be accessed at [localhost:3000/api/hello](http://localhost:3000/api/hello) when running locally.
```
doug-s-nava, line 28: did we used to support this at the /api route? Is that no longer a thing or do we need to implement a route.ts file to handle API requests into Next? https://nextjs.org/docs/app/api-reference/file-conventions/route

Based on the discussion about the API today, it seems like would likely want to avoid this and instead direct traffic to the API

emilycnava: I do see a `subscribe.ts` under pages/api, and it looks like it is called from 'NewsletterForm.tsx', whether or not that means we need a routes.ts, not sure.

doug-s-nava: yeah, that's true, but I think we're in the process of deprecating that, so I was thinking we should remove it from the README at this point. Does that sound right @acouch ?

acouch: > did we used to support this at the /api route?

Yes, that is how the pages router works. With the move to the app router what you removed doesn't make sense.

acouch (approved):
@doug-s-nava Doc update looks great. The title could be updated for the `[Issue #N]` convention. 🚀

PR #2253, opened 2024-09-27
Expand scope of trivy scans

coilysiren:
## Summary
Fixes #{ISSUE}


acouch:
Not sure why the docker build is breaking GA on this PR. I rebased, see if that makes a diff.

acouch:
@coilysiren notice that the expanded scan fails on packages in `@esbuild`. That should be removed by `npm prune --production` but it isn't: https://github.com/HHS/simpler-grants-gov/actions/runs/11110004902/job/30866752825?pr=2253#step:6:35

I think this is because some of the production packages in the frontend require esbuild, so it isn't filtered out.

.github/workflows/vulnerability-scans.yml
```diff
@@ -114,7 +114,7 @@ jobs:
           format: table
           exit-code: 1
           ignore-unfixed: true
-          vuln-type: os
+          vuln-type: 'os,library'
```
acouch, line 117: Wondering why this wasn't included previously.

PR #2328, opened 2024-10-01
[Issue #2324] Adjust various small pieces of the ELT process

### LINKED ISSUE #2324: Modify/adjust the ELT process based on the testing in dev/staging
### Summary

Placeholder / catch-all for any small fixes we may need to make.

### Acceptance criteria

_No response_

chouinar:
## Summary
Fixes #2324

## Changes proposed
Changed the default chunk size for loading from 4000 to 800

Added additional logging to extract logic to make clear where in the processing it is

No longer error if the opportunity that is supposed to be deleted does not exist (matching behavior of all other tables)

Make the `expected_number_of_awards` column a big integer

## Context for reviewers
Been testing the ELT process in test and dev recently, these are various minor adjustments.

The chunk size was changed because there is a limit to the number of fields you can specify in a query to Oracle. For especially large tables, a chunk size of 4000 causes that limit to be exceeded, 800 looks to be safe. Except for the initial load of a table, this will almost never matter as we're only copying over a handful of records each run.

The additional logging is just to help make the logs clearer where in the the processing the extract step is.

There is a sort of race condition that the delete change takes into account for the opportunity table. If an opportunity is created in our staging table, and then deleted in our staging table before the transformation ever runs, it would have errored the process. If everything is working smoothly, this won't happen, but if we don't run the transformation process, that's actually pretty easy to have happen - as it did in the dev environment because we've never run the transformation process.

Expected number of awards needs to be a big integer because in the legacy test DB, there are values like `999999999999999` which exceed a regular integer.


emilycnava (commented):
(still getting familiar with python so take this with a grain of salt, and def get at least a second opinion haha)

everything makes sense to me and makes your change desc

mdragon (approved):

PR #2341, opened 2024-10-02
Deploy Opensearch

coilysiren:
## Summary

Fixes https://github.com/HHS/simpler-grants-gov/issues/1656

## Changes proposed

- Deploys an opensearch domain
- Configures the opensearch domain to be high availability in prod, and as cheap as possible (while still being similarly configured) elsewhere. Prod is currently turned off, though.
- Adds opensearch secrets to the API service

## Context for reviewers

Opensearch takes about 20 minutes to deploy, similarly to RDS Postgres

Generating the password was a pain in the butt

This style of this infra code is similar to the style of platform code, so I'm going to post it in our platform channel for other people to see

This PR tangentially includes the updated secrets handling from https://github.com/navapbc/template-infra/pull/602. I added that because I'm adding a bunch of secrets, and I don't want them to conflict with https://github.com/HHS/simpler-grants-gov/issues/2351.

## Additional information

Opensearch deployment:

(staging has since been torn down)

image

The terraform diff for the API service layer:

```
                    {
                        Action = "ssm:GetParameters"
                        Effect = "Allow"
                      ~ Resource = "arn:aws:ssm:*:*:parameter/api/dev/api-auth-token" -> [
                          + "arn:aws:ssm:us-east-1:315341936575:parameter/search/dev/username"
                          + "arn:aws:ssm:us-east-1:315341936575:parameter/search/dev/password"
                          + "arn:aws:ssm:us-east-1:315341936575:parameter/search/dev/endpoint"
                          + "arn:aws:ssm:us-east-1:315341936575:parameter/api/dev/api-auth-token"
                        ]
                        Sid = "SecretsAccess"
                    }
                ]
```

```
                    secrets = [
                        {
                        }
                      + {
                          + name = "SEARCH_ENDPOINT"
                          + valueFrom = "arn:aws:ssm:us-east-1:315341936575:parameter/search/dev/endpoint"
                        }
                      + {
                          + name = "SEARCH_PASSWORD"
                          + valueFrom = "arn:aws:ssm:us-east-1:315341936575:parameter/search/dev/password"
                        }
                      + {
                          + name = "SEARCH_USERNAME"
                          + valueFrom = "arn:aws:ssm:us-east-1:315341936575:parameter/search/dev/username"
                        }
                    ]
```

PR #2364, opened 2024-10-03
Updates checkov

coilysiren:
## Changes proposed

- Updated checkov
- Adds various TODOs and ignores to our code, in service of making checkov pass

## Context for reviewers

Checkov is a terraform security linter. I'm doing this PR now for two reasons:

1. renovate bot has been pinging me non-stop with checkov update requests
2. I got some strange checkov bugs in https://github.com/HHS/simpler-grants-gov/pull/2341, so I decided it was time to update

### DIFF
--- a/infra/modules/database/main.tf
+++ b/infra/modules/database/main.tf
@@ -82,6 +82,7 @@ resource "aws_rds_cluster_instance" "instance" {
 resource "aws_kms_key" "db" {
   description         = "Key for RDS cluster ${var.name}"
   enable_key_rotation = true
+  # checkov:skip=CKV2_AWS_64:TODO: https://github.com/HHS/simpler-grants-gov/issues/2366
 }
 
 # Query Logging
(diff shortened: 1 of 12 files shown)


infra/modules/database/main.tf
```diff
@@ -75,11 +75,14 @@ resource "aws_rds_cluster_instance" "instance" {
   monitoring_interval                   = 30
   performance_insights_enabled          = true
   performance_insights_retention_period = 93
+
+  # checkov:skip=CKV_AWS_354:Ignore the managed customer KMS key requirement for now
```
coilysiren, line 79: This is just... a common thing we say. I don't think we are ever going to get around to it. At least not here. But maybe in the template!

mdragon (approved):
Looks good. Good to see the big version bump didn't cause too much churn.

acouch (approved):
Looks good. Wonder how we should be bucketing and labeling tix like the ones created from this PR? Will bring up internally.

PR #2383, opened 2024-10-07
New search input

babebe:
## Summary
Fixes #{2289}


api/src/services/opportunities_v1/search_opportunities.py
```diff
@@ -148,7 +142,13 @@ def _get_search_request(params: SearchOpportunityParams) -> dict:
...
     # Query
     if params.query:
-        builder.simple_query(params.query, SEARCH_FIELDS)
+        match params.experimental.scoring_rule:
+            case ScoringRuleEnum.EXPANDED:
+                builder.simple_query(params.query, EXPANDED)
+            case ScoringRuleEnum.AGENCY:
+                builder.simple_query(params.query, AGENCY)
+            case ScoringRuleEnum.DEFAULT:
+                builder.simple_query(params.query, DEFAULT)
```
chouinar, line 151: Hmm, something about this is bothering me. I'm not sure I like having effectively the same function call with one param difference setup like this.

My first thought was something like:
```py
        match params.experimental.scoring_rule:
            case ScoringRuleEnum.EXPANDED:
                filter_rule = EXPANDED
            case ScoringRuleEnum.AGENCY:
                filter_rule = AGENCY
            case ScoringRuleEnum.DEFAULT:
                filter_rule = DEFAULT

         builder.simple_query(params.query, filter_rule)
```

But then this is just a more verbose version of:

```py
# defined somewhere up top
FILTER_RULE_MAPPING = {
           ScoringRuleEnum.EXPANDED: EXPANDED,
           ScoringRuleEnum.AGENCY: AGENCY,
           ScoringRuleEnum.DEFAULT: DEFAULT
}

...

filter_rule = FILTER_RULE_MAPPING.get(params.experimental.scoring_rule, ScoringRuleEnum.DEFAULT)
builder.simple_query(params.query, filter_rule)
```

Thoughts on this approach?

chouinar (commented):
Could you add a test to https://github.com/HHS/simpler-grants-gov/blob/main/api/tests/src/api/opportunities_v1/test_opportunity_route_search.py

Don't need to worry about testing that the fields are necessarily causing a specific response order (way too complex / will change significantly), mostly just validating that we can call it with all possible values for the new field uneventfully.

api/src/services/opportunities_v1/search_opportunities.py
```diff
@@ -148,7 +153,10 @@ def _get_search_request(params: SearchOpportunityParams) -> dict:
 
     # Query
     if params.query:
-        builder.simple_query(params.query, SEARCH_FIELDS)
+        filter_rule = FILTER_RULE_MAPPING.get(
```
babebe, line 156: I could alternatively access using `FILTER_RULE_MAPPING[params.experimental.scoring_rule]` since we already handling the default case

api/tests/src/api/opportunities_v1/test_opportunity_route_search.py
```diff
@@ -1305,3 +1305,55 @@ def test_search_query_facets_200(self, client, api_auth_token):
...
         }
+
+    @pytest.mark.parametrize(
+        "search_request,expected_response",
+        [
+            # default scoring rule
+            (
+                get_search_request(
+                    order_by="opportunity_id",
+                    sort_direction=SortDirection.ASCENDING,
```
chouinar, line 1316: If you wanted to make these tests even more minimal, the order_by and sort_direction can be dropped here as the method adds some defaults to make a valid request.

chouinar (commented):
Looks good, just a few minor suggestions

chouinar (approved):
LGTM!

PR #2392, opened 2024-10-07
[Issue #1397] process roadmap accessibility

### LINKED ISSUE #1397: High-level roadmap markup is not semantic
### Summary

On the [Process page](https://simpler.grants.gov/process), under the "High-level Roadmap" section, the three steps of the roadmap each are coded as a list with one item each. This is not semantic and is confusing for screen readers. 

The right-facing chevron icons between each item are also using "Launch" as screen reader text (see #1396). These might be hidden from screen readers. At a minimum, the alt text should make more sense. 

Possible solutions: 
- Code the three steps as one single list, instead of 3 lists each with one item
- Manipulate the ARIA-Label for the heading 3 such that it'll announce "Step 1 Find", "Step 2 Advanced Reporting", "Step 3 Apply" to assistive technology

### Acceptance criteria

- [ ] Markup for the high-level roadmap section is semantic and accessible for screen readers

doug-s-nava:
## Summary
Fixes #1397

## Changes proposed
* update DOM structure on process page to follow best practices for lists (previously implemented as three invidual lists with one item each, rather than one list with three items)
* update aria labels to make high level roadmap items more readable by screen readers
* a few refactors for better readability of process page code

### Test steps

1. visit localhost:3000/process
2. visit https://simpler.grants.gov/process
3. _VERIFY_: the "high-level roadmap" section of each page matches visually
4. turn on screen reader (cmd + fn + f5 probably)
5. ctrl + opt + cmd + h to cycle through headings until you find "the high level roadmap"
6. ctrl + opt + a to read the content
7. _VERIFY_: roadmap headings are read out as "Step 1: Find" etc.

## Additional information
note that requirement in the ticket specifying removal of "launch" aria text for the chevron icons seems to have already been fixed

### DIFF
--- a/frontend/src/app/[locale]/process/ProcessMilestones.tsx
+++ b/frontend/src/app/[locale]/process/ProcessMilestones.tsx
@@ -47,46 +47,51 @@ const ProcessMilestones = () => {
         bottomBorder="dark"
         gridGap={6}
       >
-        {high_level_roadmap_items.map((_unusedItem, index) => {
-          const title = t(`high_level_roadmap_items.${index}.title`);
-          const content = t.rich(`high_level_roadmap_items.${index}.content`, {
-            p: (chunks) => (
-              <p className="font-sans-md line-height-sans-4 desktop-lg:line-height-sans-6">
-                {chunks}
-              </p>
-            ),
-            italics: (chunks) => <em>{chunks}</em>,
-          });
+        <IconList className="usa-icon-list--size-lg">
+          <Grid row>
+            {high_level_roadmap_items.map((_unusedItem, index) => {
+              const title = t(`high_level_roadmap_items.${index}.title`);
+              const content = t.rich(
+                `high_level_roadmap_items.${index}.content`,
+                {
+                  p: (chunks) => (
+                    <p className="font-sans-md line-height-sans-4 desktop-lg:line-height-sans-6">
+                      {chunks}
+                    </p>
+                  ),
+                  italics: (chunks) => <em>{chunks}</em>,
+                },
+              );
 
-          return (
-            <Grid key={title + "-key"} tabletLg={{ col: 4 }}>
-              <IconList className="usa-icon-list--size-lg">
-                <IconListItem className="margin-top-4">
-                  <IconListIcon>{getIcon(index)}</IconListIcon>
-                  <IconListContent className="tablet-lg:padding-right-3 desktop-lg:padding-right-105">
-                    <IconListTitle className="margin-bottom-2" type="h3">
-                      {title}
-                    </IconListTitle>
-                    <div className="font-sans-md line-height-sans-4 desktop-lg:line-height-sans-6">
-                      {content}
-                    </div>
-                    {
-                      // Don't show the chevron in the last row item.
-                      index < high_level_roadmap_items.length - 1 ? (
-                        <USWDSIcon
-                          className="usa-icon usa-icon--size-9 display-none tablet-lg:display-block text-base-lighter position-absolute right-0 top-0 margin-right-neg-5"
-                          name="navigate_next"
-                        />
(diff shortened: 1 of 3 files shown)


frontend/src/app/[locale]/process/ProcessMilestones.tsx
```diff
@@ -38,90 +40,97 @@ const ProcessMilestones = () => {
...
+            return (
+              <IconListItem
+                key={title + "-key"}
+                className="margin-top-4 tablet-lg:grid-col-4"
+              >
+                <IconListIcon>{getIcon(index)}</IconListIcon>
+                <IconListContent className="tablet:padding-right-7">
+                  <h3
+                    aria-label={`Step ${index + 1}: ${title}`}
+                    className="margin-bottom-2 usa-icon-list__title"
```
Comment (emilycnava) [inline frontend/src/app/[locale]/process/ProcessMilestones.tsx:73]: I'm curious why the change from using the IconListTitle component to what presumably is the same thing but in class appended to native html element form? Is this the way the docs have it now?

frontend/src/app/[locale]/process/ProcessMilestones.tsx
```diff
@@ -10,17 +10,19 @@ import {
...
-  const t = useTranslations("Process");
+  const t = useTranslations("Process.milestones");
 
-  const messages = useMessages() as unknown as IntlMessages;
-  const keys = Object.keys(messages.Process.milestones.icon_list);
+  const {
+    Process: {
+      milestones: { high_level_roadmap_items },
+    },
+  } = useMessages() as unknown as IntlMessages;
```
Comment (emilycnava) [inline frontend/src/app/[locale]/process/ProcessMilestones.tsx:25]: Might just be a me thing, but I'm not sure I understand what this section is doing, can you explain it a bit?

emilycnava (commented):
Left a few questions, haven't tested locally yet.

frontend/src/app/[locale]/process/ProcessMilestones.tsx
Comment (doug-s-nava) [reply frontend/src/app/[locale]/process/ProcessMilestones.tsx:73]: I had to do this in order to add the aria-label as requested. Using the trussworks component you can add classes directly to the h3, but you can't pass anything else through. As far as I could tell, the only thing the component was really doing was adding USWDS classes, which I replicated directly. Does that make sense?

frontend/src/app/[locale]/process/ProcessMilestones.tsx
Comment (doug-s-nava) [reply frontend/src/app/[locale]/process/ProcessMilestones.tsx:25]: Sure!

As far as I can tell from the docs, `useMessages` will always return all messages in use in the app. Since we only care about the `high_level_roadmap_items` messages, I destructured the return value from the hook to dig into just that part of the object. That way we don't have to use a long chain to index into the object when we want to reference the high_level_roadmap items. Does that make sense? Happy to do a quick call to discuss

emilycnava (approved):
Verified it works locally, checked out upstream branch, enabled voiceover etc. so LGTM!

frontend/src/app/[locale]/process/ProcessMilestones.tsx
Comment (acouch) [reply frontend/src/app/[locale]/process/ProcessMilestones.tsx:25]: Thanks for asking and for the explanation. It looks like we could ditch `useMessages()` altogether and just iterate over the three roadmap items.

Comment (doug-s-nava) [reply frontend/src/app/[locale]/process/ProcessMilestones.tsx:25]: Yeah, this whole thing is kind of odd IMO. My take is we probably shouldn't be directly using the translation messages to determine this functionality. I'd rather have a concept of the roadmap item keys outside of the translation and use those to create translated text.

On the other side of things, `useMessages` seems to be created for the purpose of accessing next intl messages on the client side. Since this is a server component, we likely could access the messages directly as you're describing, but again, I think with a slightly bigger refactor that wouldn't be necessary.

I'll take a crack at it.

acouch (approved):

PR #2393, opened 2024-10-07
Update API for Trivy Scan

acouch:
Testing updating an API file

## Summary
Fixes #{ISSUE}


coilysiren (dismissed):
will approve again when hadolint is fixed

chouinar (changes requested):
Can you hold on this - I think there is more going on here - the non-local version is getting the wrong version.

acouch:
Looks like we have the same approach #2394

PR #2406, opened 2024-10-08
[Issue #2405] Account for updates when last_upd_date is null

### LINKED ISSUE #2405: Fix issue with Oracle data load that misses updates where the last_upd_date was previously null
### Summary

The legacy Oracle database has a `last_upd_date` timestamp that we use for finding updates to our system.

This field is always null when a record is first inserted, and only on receiving at least one update will it be populated.

This means that if a record is inserted, we copy over `last_upd_date=null` and then an update happens, we try to do effectively: `where null < {some timestamp}` which means we find nothing when trying to update. What we instead should do is fallback to using the `created_date` timestamp if the update date is null.

### Reproducibility

None

### How to reproduce the bug

- [ ] Fix issue
- [ ] Adjust tests to add this scenario

### Browser

None

### Code of Conduct

chouinar:
## Summary
Fixes #2405

## Changes proposed
Fallback to using `created_date` when `last_upd_date` is null

## Context for reviewers
The legacy Oracle database has a last_upd_date timestamp that we use for finding updates to our system.

This field is always null when a record is first inserted, and only on receiving at least one update will it be populated.

This means that if a record is inserted, we copy over last_upd_date=null and then an update happens, we try to do effectively: where null < {some timestamp} which means we find nothing when trying to update. What we instead should do is fallback to using the created_date timestamp if the update date is null.

## Additional information
Found this when I realized there were cases where we just weren't processing updates. Turns out, the ELT process doesn't handle updates pretty much at all right now (this is the common case)

### DIFF
--- a/api/src/data_migration/load/sql.py
+++ b/api/src/data_migration/load/sql.py
@@ -67,7 +67,15 @@ def build_select_updated_rows_sql(
             == sqlalchemy.tuple_(*source_table.primary_key.columns),
         )
         # `WHERE ...`
-        .where(destination_table.c.last_upd_date < source_table.c.last_upd_date)
+        # NOTE: The legacy system doesn't populate the last_upd_date unless at least one update
+        #       has occurred, so we need to fallback to the created_date otherwise it will always be
+        #       null on the destination table side
+        .where(
+            sqlalchemy.func.coalesce(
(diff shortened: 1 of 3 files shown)


mdragon (approved):
Data is fun... LGTM

PR #2427, opened 2024-10-09
improve header loading process

doug-s-nava:
## Summary
Fixes (unticketed)

## Changes proposed

updates to allow header nav items to remain consistent between page navigations.

## Context for reviewers

because the list of nav items was being instantiated as empty on first render, when loading the layout on navigation, there would be a flicker of the nav, waiting for the items to be created on the second render. With this change, the nav will be rendered by default without any feature flagged items, and feature flagged items will appear on second render. There are implementations that could prevent flickering, but we'd have to implement feature flags or translation a little differently, I think.

### Test steps

1. on main, with search feature flag off, navigate around different pages on the app
2. _VERIFY_: on each navigation the header nav links flicker on each navigation
3. on this branch, with search feature flag off, navigate around different pages on the app
4. _VERIFY_: nav links remain in place while navigating through the app
5. on this branch, with search feature flag on, navigate around different pages on the app
6. _VERIFY_: search link appears in nav on second render
7. turn off search feature at /dev/feature-flags
8. _VERIFY_: search link still appears in nav until page is refreshed. This is an existing issue that this change does not deal with, but we should consider refactoring the useFeatureFlags hook to respond to changes in feature flags without refreshing

### DIFF
--- a/frontend/src/components/Header.tsx
+++ b/frontend/src/components/Header.tsx
@@ -4,7 +4,7 @@ import { useFeatureFlags } from "src/hooks/useFeatureFlags";
 import { assetPath } from "src/utils/assetPath";
 
 import { useTranslations } from "next-intl";
-import { useEffect, useRef, useState } from "react";
+import { useEffect, useMemo, useRef, useState } from "react";
 import {
   GovBanner,
   NavMenuButton,
@@ -14,8 +14,9 @@ import {
 } from "@trussworks/react-uswds";
 
 type PrimaryLinks = {
-  i18nKey: string;
+  linkText: string;
   href: string;
+  flag?: string;
 }[];
 
 type Props = {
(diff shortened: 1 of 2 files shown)


acouch:
Looks good so far. I still see flickering in the navigation, even if the feature flag code is removed. I would think that Next could be more efficient about not needing to re-render navigation. Not going to look further into it myself for the time being.

doug-s-nava:
Pushing pause on this for now, may need to refactor the useFeatureFlags hook to get this working properly.

mdragon (approved):
Approving with Aaron's note to remove the stray console

doug-s-nava:
I think this work still gets at... something interesting, but @acouch solved the header consistency issue in a much simpler way in https://github.com/HHS/simpler-grants-gov/commit/29b43afb405c3eeede02998f38d67751b1fdc840, so I'm going to close this out

PR #2430, opened 2024-10-09
[Issue #2429] Various small fixes to get ELT process working in prod

### LINKED ISSUE #2429: Fixes to get opportunity ELT process functioning in production
### Summary

A catch-all ticket for various small ELT process issues encountered in testing the ELT process in prod:

* Fix the default chunk size to actually use the 800 value we've configured
* Account for ” ” as a value for opportunity category in the transform logic (make a general pass at other similar fields)
* Handle removing the current opportunity summary connection when deleting the opportunity summary object

### Acceptance criteria

- [ ] Fixes made
- [ ] Tests updated to verify fixes

chouinar:
## Summary
Fixes #2429

## Changes proposed
Actually set the default chunk size to 800

Handle cases where lookup values in the legacy system are ` ` (whitespace) and treat them the same as null/empty strings

Handle deleting the `current_opportunity_summary` record when an `opportunity_summary` is deleted

## Context for reviewers
All issues found during testing in the prod ELT runs I was doing today

Chunk size affects how "big" the batches are when we do queries to the Oracle foreign data wrapper as Oracle has a hard limit on the size of queries that we were hitting.

For the lookup values, found a dozen or so opportunities where the opportunity category was literally a ` ` (space) and we couldn't process them. This fixes that and pre-emptively does it for other similar rows (although I think most of them haven't had the issue).

The delete just takes advantage of SQLAlchemy's approach to deletes which will automatically delete through relationships when you do `db_session.delete(some_record)`.  Now it'll know it needs to first delete the current opportunity summary record as part of that.

## Additional information
Tests updated accordingly

### DIFF
--- a/api/src/data_migration/command/load_transform.py
+++ b/api/src/data_migration/command/load_transform.py
@@ -29,7 +29,7 @@ logger = logging.getLogger(__name__)
     "--set-current/--no-set-current", default=True, help="run SetCurrentOpportunitiesTask"
 )
 @click.option(
-    "--insert-chunk-size", default=4000, help="chunk size for load inserts", show_default=True
+    "--insert-chunk-size", default=800, help="chunk size for load inserts", show_default=True
 )
 @click.option("--tables-to-load", "-t", help="table to load", multiple=True)
 @flask_db.with_db_session()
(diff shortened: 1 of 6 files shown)


mdragon (approved):
Looks good. Excited to see this running against Prod dsta.

PR #2434, opened 2024-10-09
[Task 2260] refactor search error page for maintainability

### LINKED ISSUE #2260: Refactor search page and search error page for reusability and consistency
### Summary

Currently the opportunity search error page is out of sync with the non-error version of the page. The error page duplicates most of the non-error page, so the solution is likely to refactor both of these pages to allow for maximum amount of reuse and a minimum of repetition.

Some ideas:

* split shared wrapping functionality / ui into layout files
* split other shared ui / functionality into components which can be shared between error and non-error pages

See https://github.com/HHS/simpler-grants-gov/pull/2238

### Acceptance criteria

UI / functionality shared between the error and non-error versions of the opportunity page can be updated without making changes in multiple files

doug-s-nava:
## Summary
Fixes #2260

## Changes proposed

The goal of this change is to update the search error page to:

* be up to date with recent changes to the search page (collapsable filters on mobile)
* be more easily maintainable in the future, as currently the error page is set up to duplicate nearly all of the code from the main search page

To accomplish this, I:
* created a layout for the search route containing as much shared DOM and functionality as possible
* created a SearchResult component to abstract out functionality to be removed from the error page
* updated the error page to be as in line with the new search page as possible

## Context for reviewers

Note that I considered other options for approaching this change:
* error boundaries: I could not get this to work with Next's server side rendering logic
* let the layout handle shared functionality: I tried putting everything but the search result functionality into the layout file, but the layout doesn't have access to search params so I was limited in what I could do without a larger refactor

A few other ancillary changes:
* some updates to error handling messaging and logic to make my internal debugging easier during development. These can be dropped if we don't want this scope creep
* update to search error message component to leverage USWDS component
* removal of unnecessary code from SearchFilters to allow it to be rendered client side

### Test steps

1. visit the search page
2. _VERIFY_: functions as expected when performing search / filter / sort actions
3. adjust the search request such that it will return an error - for example, update the "namespace" for the SearchOpportunityAPI such that the request will 404, or use an extension like tweak to mock an error response https://github.com/HHS/simpler-grants-gov/blob/db47fec984f17d4aabd21a957ff95679c63629f2/frontend/src/app/api/SearchOpportunityAPI.ts#L22
4. refresh
5. _VERIFY_: an error state appears on the page

## Additional information

Existing issues uncovered but not addressed:

* in an error state the page does not entirely work - filter checkboxes when clicked will update query params but will not update in the UI
* the error page is relying on the state of the page in the form of query params coming back in the error that page receives. If an error is thrown that does not contain this data, the state of the page is lost


acouch (dismissed):
Awesome start at making things cleaner and easier to maintain.

frontend/src/app/[locale]/search/error.tsx
```diff
@@ -31,138 +21,91 @@ interface ErrorProps {
...
+function isValidJSON(str: string) {
+  try {
+    JSON.parse(str);
+    return true;
+  } catch (e) {
+    return false; // String is not valid JSON
+  }
+}
+
+function createBlankParsedError(): ParsedError {
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/search/error.tsx:38]: it seems like there should be a better way but I think I'd need more context into our error throwing / handling patterns to take an informed approach. For now just tried to change as little as possible to keep this working

frontend/src/components/search/SearchFilters.tsx
```diff
@@ -23,7 +22,6 @@ export default function SearchFilters({
   category: Set<string>;
   opportunityStatus: Set<string>;
 }) {
-  unstable_setRequestLocale("en");
```
doug-s-nava, line 26: TIL that `unstable_setRequestLocale` is only meant to be used on layout and page files https://next-intl-docs.vercel.app/docs/getting-started/app-router/with-i18n-routing#add-unstable_setrequestlocale-to-all-layouts-and-pages

Issue #2321, opened 2024-10-01
Ensure that the Opportunity Listing page is consistent about hiding vs rendering empty values


acouch:
### Summary

Some of the opportunity page items are hidden if there is no data, and some are not. This should be cleaned up and be consistent.

### Acceptance criteria

- [ ] check with design / product about correct pattern (ie what should be shown vs not if there is no data)
- [ ] implement pattern
- [ ] ensure testing complies

mxk0:
@andycochran to post Figma link.

crystabelrangel:
https://www.figma.com/files/team/1329563746319512353/project/196525874/Simpler-Grants-Project?fuid=1329605892244385679

doug-s-nava:
Field details for opportunities https://docs.google.com/spreadsheets/d/163cqQy9SdJIPFOBDwJV9fuqG0xougo0O_QMjl0PFod8/edit?gid=0#gid=0

andycochran:
@doug-s-nava Let me know if you would like to sync with the design team and go through all the edge cases, or if we need to make individual design tasks to resolve

doug-s-nava:
Changes to be made as part of this change after discussion with @chouinar and @crystabelrangel:

* add double dash for null values to all 4 grid boxes, prefixed by $ for money values
* replace double dash with em dash for null assistance listings (do not worry about handling cases for individual missing values within assistance listings)
* fix display of version, use -- if null (write test for date util)
* add a -- in status widget next to "Closing" when null and status is posted
* add double dash under eligible applicants when null
* for applicant_eligibility_description make sure to strip out any markup
* for agency name, ensure that we are using the top level rather than summary value
* add double dash under description when null
* for link to additional information, use link url for text if description is missing
* for contact create sections:
    * description
    *  email (email description, newline - email address linked to mailto)
    * phone
    * -- if any are absent (other than email description)


For later:

* what to do other than hiding the status box when there is no opportunity status
* we should put posted date and closing date description on a new line under the date, and hide / expand after X characters

crystabelrangel:
Figma in devmod with comments: https://www.figma.com/design/FGxWtAgToKhehLJCiuy1zL/Simpler.Grants.gov-Project?node-id=1-10&m=dev

Four mockups for:
- Null values
- Null status
- Closed status
- Forecasted status

PR #2453, opened 2024-10-11
[Issue #2321] adjust empty value behavior on opportunity detail

doug-s-nava:
…t folder, add a test for date formatter, some other minor cleanup

## Summary
Fixes #{ISSUE}


frontend/tests/components/opportunity/OpportunityDescription.test.tsx
```diff
@@ -1,12 +1,10 @@
 /* eslint-disable @typescript-eslint/unbound-method */
+/* eslint-disable testing-library/no-node-access */
```
doug-s-nava, line 2: are we ok with this, or would we rather add test-id or label to the elements we're trying to target here? I was hoping that testing library would have something like "text content for header" or something, but it doesn't seem to.

acouch (approved):
Looks great 👍🏿

PR #2458, opened 2024-10-11
[Issue #2459] Speed up e2e tests by testing against built frontend

### LINKED ISSUE #2459: Speed e2e Tests By Running Production Version of Frontend Instead of Dev
### Summary

If we run `npm run start` instead of `npm run dev` for the e2e tests they should be faster. The only caveat is there would be a faster build time.

### Acceptance criteria

- [ ] e2e tests are faster in the CI

acouch:
## Summary
Fixes #2459

## Changes proposed
This is a small fix to run the e2e tests against the built version of the frontend. The test for this PR ran a minute slower than the previous PR (3 mins 51 secs vs 4 min 49 secs).

### DIFF
--- a/.github/workflows/ci-frontend-e2e.yml
+++ b/.github/workflows/ci-frontend-e2e.yml
@@ -51,7 +51,10 @@ jobs:
         shell: bash
 
       - name: Run E2E Tests
-        run: npm run test:e2e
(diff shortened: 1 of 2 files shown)


coilysiren (approved):
makes sense!

PR #2477, opened 2024-10-15
[Issue 2497] network fronted and api docker containers

doug-s-nava:
## Summary
Issue #2497 

## Changes proposed
In order to get search working in a local dockerized environment, network Next.js and api containers together

## Context for reviewers
This is only necessary for local development, and should have no impact otherwise.

Note that it may take a while for the next to compile the search and home routes - anywhere from 200 to 700 seconds. Not sure if this is normal, or just me, something that is isolated to the local process, or if this is a sign of an actual issue

### Test steps

1. in the api directory `make stop`
2. in the api directory `make start`
3. uncomment line 19 of the frontend/.env.development file, and comment out line 20
4. in the frontend directory `make stop`
5. in the frontend directory `make start`
6. visit localhost:3000/search
7. _VERIFY_: no errors, and search results are displayed

### DIFF
--- a/api/docker-compose.yml
+++ b/api/docker-compose.yml
@@ -8,6 +8,8 @@ services:
       - "5432:5432"
     volumes:
       - grantsdbdata:/var/lib/postgresql/data
+    networks:
+      - default
 
   opensearch-node:
     image: opensearchproject/opensearch:latest
@@ -32,6 +34,8 @@ services:
     ports:
       - 9200:9200 # REST API
       - 9600:9600 # Performance Analyzer
+    networks:
+      - default
 
   opensearch-dashboards:
     image: opensearchproject/opensearch-dashboards:latest
@@ -43,13 +47,15 @@ services:
     environment:
       - 'OPENSEARCH_HOSTS=["http://opensearch-node:9200"]'
       - DISABLE_SECURITY_DASHBOARDS_PLUGIN=true # disables security dashboards plugin in OpenSearch Dashboards
+    networks:
+      - default
 
   localstack:
     container_name: "${LOCALSTACK_DOCKER_NAME:-localstack-main}"
     image: localstack/localstack
     ports:
-      - "127.0.0.1:4566:4566"            # LocalStack Gateway
-      - "127.0.0.1:4510-4559:4510-4559"  # external services port range
+      - "127.0.0.1:4566:4566" # LocalStack Gateway
+      - "127.0.0.1:4510-4559:4510-4559" # external services port range
     environment:
       # LocalStack configuration: https://docs.localstack.cloud/references/configuration/
       - DEBUG=${DEBUG:-0}
@@ -59,6 +65,8 @@ services:
     volumes:
       - "${LOCALSTACK_VOLUME_DIR:-./volume}:/var/lib/localstack"
       - "/var/run/docker.sock:/var/run/docker.sock"
+    networks:
+      - default
 
   grants-api:
     build:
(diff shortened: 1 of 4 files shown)


api/docker-compose.yml
```diff
@@ -43,13 +47,15 @@ services:
...
+      - default
 
   localstack:
     container_name: "${LOCALSTACK_DOCKER_NAME:-localstack-main}"
     image: localstack/localstack
     ports:
-      - "127.0.0.1:4566:4566"            # LocalStack Gateway
-      - "127.0.0.1:4510-4559:4510-4559"  # external services port range
+      - "127.0.0.1:4566:4566" # LocalStack Gateway
+      - "127.0.0.1:4510-4559:4510-4559" # external services port range
```
chouinar, line 58: Why was this reformatted? I have no preference either way / we have no specific linting for this file, this was just how the formatting came from the OpenSearch docs.

frontend/.env.development
```diff
@@ -14,6 +14,9 @@ SENDY_API_KEY=
 SENDY_API_URL=
 SENDY_LIST_ID=
 
+# when running the full containerized app locally in Docker, uncomment the line below to
+# enable connection to the conatainerized API server
+# API_URL=http://host.docker.internal:8080
```
chouinar, line 19: We have a similar issue in the API where the path we want inside docker and outside are different.

I use direnv to handle env vars with an `.envrc` file that auto-loads when I'm in the API directory. The local.env file always has the docker specific values, and I override for non-docker values outside.

```sh
#!/bin/bash

# This file can be used by direnv to load the local.env file
# into your current terminal session.
# Steps:
# * Install direnv: https://direnv.net/
# * Configure direnv: https://direnv.net/docs/hook.html
# * In this folder, run "direnv allow ."

set -o allexport
source local.env
set +o allexport

# If you are running outside of the Docker container, the DB can
# be found on localhost:5432. Inside the container it's referenced via
# the name of the docker container.
export DB_HOST=localhost
export S3_ENDPOINT_URL=http://localhost:4566
export PY_RUN_APPROACH=local

export SEARCH_ENDPOINT=localhost
```

I thought this was somewhere in the docs, but it looks like it wasn't copied over from the template repo: https://github.com/navapbc/template-application-flask/tree/main/docs/app#running-natively

chouinar (commented):
I did a quick test of the API/DB/OpenSearch running locally and it all seems to work the same as before, didn't run into any issues with the different parts communicating with each other.

mdragon:
Just to note on the really slow builds. I was seeing this when I first got things running locally and it was some process on the Mac going wild and eating CPU. I think it was some plugin for language parsing/whatever in VSCode and after I Quit VSCode and restarted everything was fine and super snappy.

acouch (approved):
Awesome! Only suggestion would be to make a quick ticket for the board.

api/docker-compose.yml
doug-s-nava: That was just my VS code autoformatting for YML files, I think, not something intentional. It looks like the `<code> # <comment>` pattern is more widely used in our codebase than the `<code> #    <comment>` pattern so I'm inclined to keep it this way, but let me know if you'd like me to revert

frontend/.env.development
doug-s-nava: Cool, and very good to know. This could be an option for the frontend as well, but the common use case I think is the opposite of the API side. Where on the API side it seems like we're usually running the DB in a container so that is the default value, on the frontend side we're usually running the app outside of the container so that would be the default value to commit.

Also, with an .envrc file pointing to localhost, for example, how do you deal with the case where you want to run the DB in a container? Do you unload direnv before starting up the containers? I'm having a hard time figuring out an easy system for this that doesn't require some action on the part of the developer when they're going with the less common use case, whether it's uncommenting code or disabling a plugin...

chouinar: Most things always run inside a container, there isn't a way to have the DB run outside the container in our setup.

What running locally/non-locally mostly changes is for how scripts/tests work. If I'm running the python tests in the docker container, when they hit the DB, it needs to work one way, but outside the docker container, I want a different URL.

.envrc never affects anything with Docker, just sometimes tells our scripts to run differently outside Docker. This mostly matters for the `PY_RUN_APPROACH` flag which tells some of our make commands to run natively rather than inside docker.

chouinar (approved):

PR #2483, opened 2024-10-16
[Issue #2123] Add Opportunity Attachment and Lookup Tables

### LINKED ISSUE #2123: Create tables for opportunity attachments
We want to create the following two tables in our SQLAlchemy models:

Image

Only the created_by, updated_by, and legacy_folder_id columns should be nullable. Note you can use the python typing to dictate this. `my_field: Mapped[int]` is not nullable. `my_field: Mapped[int | None]` is nullable.

---

The `lk_opportunity_attachment_type_id` lookup table should follow our conventions for setting up a lookup table. We will only have two values at the moment for this lookup enum:

* `notice_of_funding_opportunity`
* `other`

See: https://github.com/navapbc/simpler-grants-gov/blob/main/documentation/api/lookup-values.md for further details on how this works.

### Acceptance criteria
* [ ] Tables created in our SQLAlchemy models
* [ ] Migrations created via Alembic
* [ ] Lookup table values configured (should auto-populate their table when running migrations)

mikehgrantsgov:
## Summary
Fixes #2123

## Changes proposed
- Add models and supporting migrations for Opportunity attachments, including lookup table for type.

## Context for reviewers
See generated migration and screenshots of result after db upgrade. 

Only the created_by, updated_by, and legacy_folder_id columns should be nullable.

## Additional information
Screenshots of new tables after running `make db-upgrade`

Screenshot 2024-10-16 at 9 53 59 AM

Screenshot 2024-10-16 at 9 53 47 AM

Screenshot 2024-10-16 at 9 55 06 AM


api/src/db/models/opportunity_models.py
```diff
@@ -385,3 +391,29 @@ class CurrentOpportunitySummary(ApiSchemaTable, TimestampMixin):
...
+
+
+class OpportunityAttachment(ApiSchemaTable, TimestampMixin):
+    __tablename__ = "opportunity_attachment"
+
+    attachment_id: Mapped[int] = mapped_column(BigInteger, primary_key=True)
+
+    opportunity_id: Mapped[int] = mapped_column(
+        BigInteger, ForeignKey(Opportunity.opportunity_id), primary_key=True, index=True
+    )
```
chouinar, line 403: Missed this the first time, we don't need opportunity ID to be a part of the primary key here I don't think.

mikehgrantsgov: Ah, makes sense. Made that change.

chouinar (approved):
LGTM!

PR #2503, opened 2024-10-16
Require build and publish before database migrations

coilysiren:
## Context

Right now builds are failing because of a race condition between migrations and build and publish. This fixes that by making the dependency explicit.


mdragon (approved):
Looks good.

Issue #2472, opened 2024-10-15
Configure Opensearch to restrict access from AWS services


coilysiren:
### Summary

eg. scoping down this policy, without breaking opensearch

https://github.com/HHS/simpler-grants-gov/pull/2452/files#diff-b5ef403a679056e13d89fb4d322e947a82edfd50d1e6f0b330e1fbfce3230394

### Acceptance criteria

policy scoped down

coilysiren:
This requires a little bit of infra work and a large(r) amount of python work. [Right now the opensearch client can handle _**either**_ opensearch auth or AWS auth, but both](https://github.com/HHS/simpler-grants-gov/pull/2450/files#diff-6cab7949e69b1d5e141ce9e2a0c8144a7dadd3d2664b1a4ea31f870259ec7585). For security reasons, we would like it to handle both.

coilysiren:
via means of the opensearch provider:

- https://registry.terraform.io/providers/opensearch-project/opensearch/latest/docs
- https://registry.terraform.io/providers/opensearch-project/opensearch/latest/docs/resources/role
- https://registry.terraform.io/providers/opensearch-project/opensearch/latest/docs/resources/roles_mapping

coilysiren:
See https://github.com/navapbc/template-infra/blob/main/docs/decisions/infra/0006-provision-database-users-with-serverless-function.md for description of a similar problem

coilysiren:
For testing:

```bash
aws ecs run-task \
  --region=us-east-1 \
  --cluster=api-dev \
  --task-definition=api-dev \
  --started-by="$(whoami)" \
  --launch-type=FARGATE \
  --platform-version=1.4.0 \
  --network-configuration \
  '{
        "awsvpcConfiguration": {
            "subnets": [
                "GO GET THIS,
                "GO GET THIS",
                "GO GET THIS"
            ],
            "securityGroups": [
                "GO GET THIS"
            ],
            "assignPublicIp": "DISABLED"
        }
    }' \
  --overrides \
  '{
      "containerOverrides": [
      {
        "name": "api-dev",
        "command": [
          "poetry",
          "run",
          "flask",
          "load-search-data",
          "load-opportunity-data"
        ],
        "environment": [
          {
            "name": "FLASK_APP",
            "value": "src.app:create_app()"
          },
          {
            "name": "ENVIRONMENT",
            "value": "dev"
          }
        ]
      }
    ]
}'
```

[...]

Issue #2351, opened 2024-10-03
Nava infrastructure template updates


mdragon:
https://github.com/navapbc/template-infra/

coilysiren:
I recently migrated us to the platform step functions pattern. That contained a breaking change, and took roughly a day.

Next I'm going to do a more standard upgrade, eg. updating this: https://github.com/HHS/simpler-grants-gov/blob/main/.template-version, and take the average between the two.

I assume the time on both will be ~1 day.

Then I'll count up the number of releases with breaking changes between the last diff on https://github.com/HHS/simpler-grants-gov/blob/main/.template-version, and now.

coilysiren:
Ah, a blocker has emerged. Grants.gov is using a setup with multiple applications. This impedes our ability to update the platform template, which only uses a single application.

There's a WIP tool for supporting multiple applications, it lives here: https://github.com/navapbc/platform-cli/

I'll investigate how usable it is from the grants.gov use case now

coilysiren:
We are currently here https://github.com/navapbc/platform-test/commit/1985b64d8f5aec2c7ef4e554a6dd22b411156eba

On the list of updates from here https://github.com/navapbc/platform-test/commits/main/.template-version?after=95fc5e83dee811490eca96d91bcfb19243fb7379+139

coilysiren:
The next step of significance is here: https://github.com/HHS/simpler-grants-gov/commit/1bf8744d8ccd92d0fc0c258b81cd53902a1c35bc#diff-d102067c049594f7510d51fb2d308cda319bfbf275f341057668e6ea116857c2R1

coilysiren:
Update command:

```bash
curl https://raw.githubusercontent.com/navapbc/template-infra/main/template-only-bin/update-template | bash -s fe5c7cd24d3c2c9f15c342826cda0a20af4cd0a5
```

coilysiren:
🤦🏼 nevermind I was looking in the wrong folder

coilysiren:
Update - the [update template script](https://github.com/navapbc/template-infra/blob/main/template-only-bin/update-template) is too tedious for a repo of our size and complexity. We should wait until [platform CLI](https://github.com/navapbc/platform-cli) is working.

coilysiren:
Not sure how I got this wrong, but the next step of significance is actually here:

- https://github.com/navapbc/platform-test/commit/acd94c9e9b5da6ee62901614a08f0ca0d9a1c8a6

[...]

PR #2506, opened 2024-10-17
[Issue #2351, #2472] Migrate to platforms step function pattern

coilysiren:
## Summary

Relates to: #2351, #2472

## Changes proposed

There are two goals to this PR:

1. Make a material change that gets us closer in line with the platform. Specifically we are pulling in https://github.com/navapbc/template-infra/pull/745
2. Updates the step functions in advance of adding another step function - for the terraform in ECS stuff

## Context for reviewers

The majority of the diff is security related, specifically: 

- via guidance from @lorenyu, the permissions for managing the step function and workflow triggers were moved off of the task exec role. that's the diff on `infra/modules/service/access-control.tf`
- those permissions were moved to `infra/modules/service/scheduler_role.tf` and `infra/modules/service/workflow_orchestrator_role.tf`

The rest of the diff is around easy of configuration:

- `infra/modules/service/scheduled_jobs.tf` represents the new, highly configurable, way to create scheduled jobs
- `infra/api/app-config/env-config/scheduled_jobs.tf` shows how two scheduled jobs are configured

## Additional information

image

image


infra/api/app-config/env-config/scheduled_jobs.tf
```diff
@@ -0,0 +1,51 @@
...
+      "--load",
+      "--transform",
+      "--set-current",
+    ],
+  }
+  scheduled_jobs = {
+    copy-oracle-data = {
+      task_command        = ["poetry", "run", "flask", "data-migration", "copy-oracle-data"]
+      schedule_expression = "rate(2 minutes)"
+    }
```
coilysiren, line 45: @chouinar fancy new simple way to setup scheduled jobs

infra/api/app-config/env-config/scheduled_jobs.tf
```diff
@@ -0,0 +1,51 @@
...
+      "load-transform",
+      "--load",
+      "--transform",
+      "--set-current",
+    ],
+  }
+  scheduled_jobs = {
+    copy-oracle-data = {
+      task_command        = ["poetry", "run", "flask", "data-migration", "copy-oracle-data"]
+      schedule_expression = "rate(2 minutes)"
```
mdragon, line 44: Should we include enabled/disabled here too, or does a null schedule or something like that cause the job to just not run? Just thinking there may be times we want to keep the job from running, but not delete it from here entirely.

infra/modules/service/scheduled_jobs.tf
```diff
@@ -0,0 +1,86 @@
+resource "aws_scheduler_schedule" "scheduled_jobs" {
+  for_each = var.scheduled_jobs
+
+  # TODO(https://github.com/navapbc/template-infra/issues/164) Encrypt with customer managed KMS key
+  # checkov:skip=CKV_AWS_297:Encrypt with customer key in future work
+
+  name                         = "${var.service_name}-${each.key}"
+  state                        = "ENABLED"
+  schedule_expression          = each.value.schedule_expression
```
mdragon, line 9: Noting the enabled here that we might want to control via the list of jobs per the other comment.

infra/modules/service/scheduled_jobs.tf
```diff
@@ -0,0 +1,86 @@
+resource "aws_scheduler_schedule" "scheduled_jobs" {
+  for_each = var.scheduled_jobs
+
+  # TODO(https://github.com/navapbc/template-infra/issues/164) Encrypt with customer managed KMS key
+  # checkov:skip=CKV_AWS_297:Encrypt with customer key in future work
+
+  name                         = "${var.service_name}-${each.key}"
+  state                        = "ENABLED"
```
mdragon, line 8: Noting the enabled here that we might want to control via the list of jobs per the other comment.

mdragon (approved):
Approving as is. If we do want to facilitate pausing jobs via config as suggested in the comments, I'm happy to have that come in a future ticket.

infra/api/app-config/env-config/scheduled_jobs.tf
chouinar: Thanks, this is simple enough that I can set things up / not bother you for any adjustments we make.

infra/api/app-config/env-config/scheduled_jobs.tf
coilysiren: Added! https://github.com/HHS/simpler-grants-gov/pull/2506/commits/1ba3d0615fdff173d9020b34a7e7fc83cfe206e9

mdragon (approved):

PR #2502, opened 2024-10-16
[Unticketed] Esbuild failing scans

mdragon:
Attempt to manually zap esbuild since we don't need it and it's failing the security scans.

PR #2508, opened 2024-10-17
[Issue 2507] upgrade storybook to version 8

### LINKED ISSUE #2507: Upgrade Storybook to v8
There is a new major version of storybook available, which should fix some dependency vulnerabilities, add features, such as Server Component support, etc.

See more here!

https://storybook.js.org/blog/storybook-8/

doug-s-nava:
## Summary
Fixes #2507

## Changes proposed
Upgrade to the latest version of Storybook.

This PR is largely needed in order to work around a security issue with esbuild that is present in storybook until 8.4. A workaround was applied in https://github.com/HHS/simpler-grants-gov/pull/2502 that can hopefully now be reverted.

Note that to get this working I ran the suggested upgrade script mentioned below, which eventually worked, but then had to fix the reference to our next config, which was not being found for some reason, in order to get storybook to load

https://storybook.js.org/docs/migration-guide

## Context for reviewers

Most of our stories are currently broken, so set a baseline expectation on what is currently working before jumping in and assuming that the upgrade broke stuff

### Test steps

1. on main, `npm install && npm run storybook`
2. visit http://localhost:6006
3. get a sense of which stories are working and not working on main
4. run steps 1 and 2 on this branch
5. _VERIFY_: everything that worked on main still works here

## Additional information

it looks like Go 1.23 support in esbuild only came along last month in version 0.24 https://github.com/evanw/esbuild/releases/tag/v0.24.0

version 0.24 of esbuild is only supported in storybook core as of v8.4.0-alpha.4 https://github.com/storybookjs/storybook/commit/c40aba23008334cb0668fb96a2238779e25495d5

now that storybook 8.4 is officially released, we can upgrade and hopefully remove the workaround built in https://github.com/HHS/simpler-grants-gov/pull/2502

### upgrade pitfalls

- peer dependency issue with @storybook/blocks preventing an npm install
    - resolved by explicitly adding `@storybook/blocks: ^8.4.2` as a dev dependency, and deleting package-lock
- missing @types/lodash
    - previously was required by @storybook/blocks but now isn't anymore, so we need to include it explicitly
-  missing @types/node-fetch
    - previously was required by @storybook/core-common but now isn't anymore, so we need to include it explicitly


emilycnava (approved):
Verified that storybook between main and this branch are consistent and show similar failing components, unrelated to upgrade

acouch:
Testing locally, now that we are on 8 we might be able to include RSCs

acouch (approved):

PR #2509, opened 2024-10-17
[Issue #2466] Update API schema to include opportunity attachments

### LINKED ISSUE #2466: Update API schema to include opportunity attachments
### Summary

We currently use the same schema for both the search and GET opportunity endpoint. We likely want a separate schema for non-search (use inheritance so we don't need to redefine everything), as we don't want the attachments on the search schema.

The API response should look something like (trimming out some existing things for brevity):

```json
{
   "opportunity_id": 1234,
   "summary": { .. },
   "attachments": [
       {
         "file_location": "https://...",
         "mime_type": "application/pdf",
         "file_name": "my_NOFO.pdf",
         "file_description": "The full announcement NOFO",
         "file_size_bytes": "10012",
         "created_at": "2024-01-01 12:00:00",
         "updated_at": "2024-01-01 12:00:00",
         "opportunity_attachment_type": "notice_of_funding_opportunity"
       }
   ]
}
```

For futher details on the types of these fields in the API, see https://github.com/HHS/simpler-grants-gov/issues/2123 where we will set this up in the DB. Note that most of these are just strings and integers. The attachment type will be an enum that would have been created in that PR. If you are building this in parallel, make it a string temporarily and we'll switch it later.

A few fields like created_by/updated_by/legacy_folder_id are deliberately not included in the API schema as they're only for internal purposes at the moment.

### Acceptance criteria

- [ ] API schema modified
   - [ ] Examples included for fields that the default isn't helpful (ie. strings default should be changed which isn't helpful)
   - [ ] Descriptions provided for all individual fields to explain what they do
- [ ] The API should fetch the data from the database - NOTE that adjusting the file locations to be something a user can go to will be handled in a follow-up PR
- [ ] Tests added to validate data returned properly from the database for these fields (note you may need to create a factory for the DB table)

mikehgrantsgov:
## Summary
Fixes #2466

## Changes proposed
- Add opportunity attachments to API response
- Add tests including new factories and validators
- Separate response types for search results vs. opportunity GET request.

## Context for reviewers

We currently use the same schema for both the search and GET opportunity endpoint. We likely want a separate schema for non-search (use inheritance so we don't need to redefine everything), as we don't want the attachments on the search schema.

## Additional information
See attached unit test to validate attachment data coming back from API.


api/src/api/opportunities_v1/opportunity_schemas.py
```diff
@@ -308,6 +308,60 @@ class OpportunityV1Schema(Schema):
...
+        metadata={
+            "description": "The user who last updated the attachment",
+        }
+    )
+
+    legacy_folder_id = fields.Integer(
+        metadata={
+            "description": "The legacy folder ID",
+        }
+    )
```
chouinar, line 354: We don't want these fields in the API schema itself - they're just for some internal purposes at the moment for the legacy system.

We do however want the created_at/updated_at timestamp fields for these.

api/src/api/opportunities_v1/opportunity_schemas.py
```diff
@@ -308,6 +308,60 @@ class OpportunityV1Schema(Schema):
...
+    file_size_bytes = fields.Integer(
+        metadata={"description": "The size of the attachment in bytes", "example": 10012}
+    )
+    opportunity_attachment_type = fields.String(
+        metadata={
+            "description": "The type of attachment",
+            "example": "notice_of_funding_opportunity",
+        }
+    )
```
chouinar, line 338: We can make this an enum and point to the one you created in the prior PR to setup the DB table

api/tests/src/db/models/factories.py
```diff
@@ -234,6 +236,27 @@ class Meta:
...
+    attachment_id = factory.Sequence(lambda n: n)
+
+    file_location = factory.Faker("url")
+    mime_type = factory.Faker("mime_type")
+    file_name = factory.Faker("file_name")
+    file_description = factory.Faker("sentence")
+    file_size_bytes = factory.Faker("random_int", min=1000, max=10000000)
+    opportunity_attachment_type = factory.Faker(
+        "random_element", elements=OpportunityAttachmentType
+    )
```
chouinar, line 252: Another way you can do this - slightly less to type.
```suggestion
    opportunity_attachment_type = factory.fuzzy.FuzzyChoice(OpportunityAttachmentType)
```

api/tests/src/db/models/factories.py
```diff
@@ -234,6 +236,27 @@ class Meta:
         sqlalchemy_session_persistence = "commit"
 
 
+class OpportunityAttachmentFactory(BaseFactory):
```
chouinar, line 239: Could you have the opportunity factory automatically add attachments for itself?

Should be something like:
```py
    opportunity_attachments = factory.RelatedFactoryList(
        OpportunityAttachmentFactory,
        factory_related_name="opportunity",
        size=lambda: random.randint(1, 2),
    )
```

api/tests/src/db/models/factories.py
```diff
@@ -234,6 +236,27 @@ class Meta:
...
+    class Meta:
+        model = opportunity_models.OpportunityAttachment
+
+    attachment_id = factory.Sequence(lambda n: n)
+
+    file_location = factory.Faker("url")
```
chouinar, line 245: FYI - I need to make the ticket still, but there will be a follow-up where we'll want these to be s3 paths in the DB, and have the API presign the URLs to make them something clickable.

api/src/api/opportunities_v1/opportunity_schemas.py
mikehgrantsgov: Got it, made this change.

api/tests/src/db/models/factories.py
mikehgrantsgov: Yep, I played around with that earlier but was getting some SQLAlchemy errors, but this snippet works. Thanks

chouinar: Yeah, it's very precise what you need to specify for these related factories to avoid SQLAlchemy or circular dependency issues - but they sure make things easier once they work.

mikehgrantsgov: Actually, I spoke too soon. The problem seems to be transient -- tests work once, then fail again with an error like:

```
sqlalchemy.exc.InvalidRequestError: Instance <Opportunity at 0xffff81e4c350> is already associated with an instance of <class 'src.db.models.opportunity_models.OpportunityAttachment'> via its OpportunityAttachment.opportunity attribute, and is only allowed a single parent. (Background on this error at: https://sqlalche.me/e/20/bbf1)
```

chouinar: Let me take a look

api/src/api/opportunities_v1/opportunity_schemas.py
mikehgrantsgov: Updated this, thanks.

api/tests/src/db/models/factories.py
chouinar: Hmm, it's something with the `single_parent=True` in the opportunity relationship. The pattern here of the tables is effectively the same as the assistance listing table (an opportunity has multiple assistance listings / attachments), but the assistance listing table doesn't have that flag either.

I'd just switch the relationship on the attachment table to: `opportunity: Mapped[Opportunity] = relationship(Opportunity)` - I'm not sure what issue the single_parent thing is getting hung up on because that would imply it's attempting to have the opportunity changed, but if you generate several opportunities, there isn't anything that exists.

mikehgrantsgov: Works now, thanks!

chouinar (approved):
LGTM!

Issue #2494, opened 2024-10-16
Test TransformAgency ELT process in dev/staging


chouinar:
### Summary

Test process, fix any issues, create tickets as follow-ups.

### Acceptance criteria

- [ ] Note down observations

chouinar:
Created several follow-up tickets after doing one round of testing yesterday. Two issues identified:
* Need to handle top-level agency names
* Need to handle agencies without certain fields unconfigured that we don't intend to use anyways

https://github.com/HHS/simpler-grants-gov/issues/2490 - add top-level agency name to the API response
https://github.com/HHS/simpler-grants-gov/issues/2491 - add top-level agency info to the DB model
https://github.com/HHS/simpler-grants-gov/issues/2492 - populate the top-level agency info in the transform process
https://github.com/HHS/simpler-grants-gov/issues/2493 - ease up validation on some agency transformations

chouinar:
Here are the agencies that failed to process when running in dev:

```

1 | GDIT-PDCM-LEV2 | Cannot create agency GDIT-PDCM-LEV2 as required fields are missing: AgencyContactCity
-- | -- | --
2 | GDIT-AG | Cannot create agency GDIT-AG as required fields are missing: AgencyContactCity
3 | GDIT-311 | Cannot create agency GDIT-311 as required fields are missing: AgencyContactCity
4 | GDIT-21 | Cannot create agency GDIT-21 as required fields are missing: AgencyContactCity
5 | GDIT-31 | Cannot create agency GDIT-31 as required fields are missing: AgencyContactCity
6 | GDIT-2 | Cannot create agency GDIT-2 as required fields are missing: AgencyContactCity
7 | GDIT-06282011 | Cannot create agency GDIT-06282011 as required fields are missing: AgencyContactCity
8 | GDIT-070120111 | Cannot create agency GDIT-070120111 as required fields are missing: AgencyContactCity
9 | GDIT-07012011 | Cannot create agency GDIT-07012011 as required fields are missing: AgencyContactCity
10 | dxprobeyd0000BA | Cannot create agency dxprobeyd0000BA as required fields are missing: ldapGp
11 | dxprobeyd00000F | Cannot create agency dxprobeyd00000F as required fields are missing: ldapGp
12 | dxprobeyd00006C | Cannot create agency dxprobeyd00006C as required fields are missing: ldapGp
13 |   | Cannot create agency as required fields are missing: ldapGp
```

[...]

PR #2513, opened 2024-10-17
[Issue #2493] Allow certain agency fields to be null from the legacy values

### LINKED ISSUE #2493: Modify Agency table + transform job to allow certain legacy values to be null
### Summary

In our Agency table, we want to allow a few columns to be nullable. We also want to update the TransformAgency class to not include these values in the `REQUIRED_FIELDS` list that causes us to error if the field is null.

(our column + legacy field name)
* `ldap_group` / `ldapGroup`
* `description` / `description`
* `label` / `label`

### Acceptance criteria

- [ ] Agency table schema updated + Alembic migrations generated
- [ ] TransformAgency required fields updated
- [ ] Tests for TransformAgency updated accordingly

chouinar:
## Summary
Fixes #2493

## Changes proposed
Modified three fields (ldapGp, description, label) in the agency table to be nullable

Adjusted the transform agency process to allow these fields to be nullable as well

## Context for reviewers
This isn't an issue in prod right now, but is accounting for a problem I found in the dev/staging environments. The agency data is stored in a very different way (key-value pairs) in the legacy data, and these fields are sometimes not present in the test data. There isn't any reason we should require them as we likely won't use them, so I'm letting them be nullable to let these agencies with the values null get populated in the lower envs.

## Additional information
There is some additional context on https://github.com/HHS/simpler-grants-gov/issues/2494


mdragon (approved):
Looks good.

PR #2514, opened 2024-10-17
Configure Opensearch Domain via Terraform inside VPC

coilysiren:
## Summary
Fixes #{ISSUE}

### DIFF
--- a/bin/configure-monitoring-secret.sh
+++ b/bin/configure-monitoring-secret.sh
@@ -17,7 +17,7 @@ ENVIRONMENT=$2
 INTEGRATION_ENDPOINT_URL=$3
 
 terraform -chdir="infra/$APP_NAME/app-config" init > /dev/null
-terraform -chdir="infra/$APP_NAME/app-config" apply -refresh-only -auto-approve> /dev/null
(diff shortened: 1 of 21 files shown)


bin/configure-monitoring-secret.sh
```diff
@@ -17,7 +17,7 @@ ENVIRONMENT=$2
 INTEGRATION_ENDPOINT_URL=$3
 
 terraform -chdir="infra/$APP_NAME/app-config" init > /dev/null
-terraform -chdir="infra/$APP_NAME/app-config" apply -refresh-only -auto-approve> /dev/null
```
coilysiren, line 20: https://github.com/navapbc/template-infra/pull/586

infra/api/search/main.tf
```diff
@@ -0,0 +1,55 @@
...
+  index_permissions {
+    index_patterns  = ["*"]
+    allowed_actions = ["*"]
+  }
+
+  tenant_permissions {
+    tenant_patterns = ["*"]
+    allowed_actions = ["*"]
+  }
+}
```
coilysiren, line 47: this isn't much better than our current situation 😆 me and @chouinar will work on these permissions

Issue #2492, opened 2024-10-16
Create a step for the transformation process that links agencies to their top-level agencies


chouinar:
### Summary

As part of our transformation process, we transform agency data from the legacy system, but we want to do a second pass over agency data once in our system in order to make the agency data connect together properly.

In the system, there are agencies and sub-agencies. For example, `HHS` is a top-level agency, and `HHS-NIH11` is a subagency. We want `HHS-NIH11` to point to `HHS` in the `top_level_agency_id` foreign key we setup in https://github.com/HHS/simpler-grants-gov/issues/2491

Determining whether an agency is a top-level or subagency is simple, a top-level agency will have no dashes, while a sub-agency will have one or more dashes in its agency code with the top-level agency being everything before the first dash. We want to iterate over the agency data in our system and attach appropriately.

## Implementation details

The data set size of the agency data is very small (~1000 total records) so a full re-iteration over the data is perfectly fine and shouldn't take more than a handful of seconds.

We want to make a new class derived from `AbstractTransformSubtask` like the other transform jobs, and have it [run after](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/data_migration/transformation/transform_oracle_data_task.py#L82) the TransformAgency task.

We DO want to potentially update an existing connection (in case we have to fix anything later) but we DO NOT want to error if a connection cannot be made as I'm almost certain the test data has subagencies without a top-level agency (and top-level agencies themselves won't connect to anything).

### Acceptance criteria

- [ ] New transform subtask created
- [ ] Tests added for new subtask

PR #2517, opened 2024-10-18
[Issue #2491] Modify the Agency table schema to include a link to a top-level agency

### LINKED ISSUE #2491: Modify the Agency table schema to include a link to a top-level agency
### Summary

We want to add a foreign key in the agency table to the (same) agency table called `top_level_agency_id` as well as a corresponding relationship.

This field will be populated by a separate process and used in setting up the agency name in the API

In the opportunity table, add a new property called `top_level_agency_name` that pulls from agency record relationship it already has to get the top-level agency name. Something like:
```py
@property
def top_level_agency_name(self) -> str | None:
        if self.agency_record is not None and self.agency_record.top_level_agency is not None:
            return self.agency_record.top_level_agency.agency_name

        return None
```

### Acceptance criteria

- [ ] DB tables updated
- [ ] Alembic migrations created
- [ ] Do some level of sanity testing to verify the relationships work correctly (tests aren't needed as follow-up code will be tested and confirm the behavior)
- [ ] Modify the [seed_local_db](https://github.com/HHS/simpler-grants-gov/blob/main/api/tests/lib/seed_local_db.py#L121) script to connect the subagency to its top-level (`DOC` should be the top-level agency of `DOC-EDA`). This may require some small reworking of the setup script to work.

mikehgrantsgov:
## Summary
Fixes #2491

## Changes proposed
New schema fields
DB Migration
In the opportunity table, add a new property called top_level_agency_name that pulls from agency record relationship it already has to get the top-level agency name. Something like:

## Context for reviewers
We want to add a foreign key in the agency table to the (same) agency table called top_level_agency_id as well as a corresponding relationship.

This field will be populated by a separate process and used in setting up the agency name in the API

## Additional information

After seed is complete:
https://github.com/user-attachments/assets/a861d717-06ba-4731-a88f-679945ba0ae7


api/tests/src/db/models/factories.py
```diff
@@ -316,6 +316,8 @@ class Params:
 
         no_current_summary = factory.Trait(current_opportunity_summary=None)
 
+        no_attachments = factory.Trait(opportunity_attachments=None)
```
mikehgrantsgov, line 319: Added this trait and associated seeder update to make sure no PK violations hit (attachment ID was causing a problem with duplicate generation).

chouinar: You can work around that by adding something like what we have in the opportunity factory to the attachment factory:
```py
    @classmethod
    def _setup_next_sequence(cls):
        if _db_session is not None:
            value = _db_session.query(
                func.max(opportunity_models.Opportunity.opportunity_id)
            ).scalar()
            if value is not None:
                return value + 1

        return 1
```
Effectively, tell the factory to start counting differently

Alternatively - if we don't set the primary key ID at all in the factory - the DB will automatically set it (Postgres integer primary keys are automatically given a sequence they draw from).

api/tests/src/db/models/factories.py
```diff
@@ -721,6 +723,8 @@ class Meta:
     is_image_workspace_enabled = factory.Faker("boolean")
     is_validation_workspace_enabled = factory.Faker("boolean")
 
+    top_level_agency = factory.Faker("agency")
```
chouinar, line 726: I don't think this would work? The `agency` provider is written towards the top of this file, and just returns a string, but this would need to be another instance of the agency class. I think for the factory, I'd probably just leave this unset by default otherwise you'd need to have it refer to itself and run into a recursive loop issue.

api/tests/lib/seed_local_db.py
```diff
@@ -142,13 +161,28 @@ def _build_agencies(db_session: db.Session) -> None:
...
+            if not top_level_agency:
+                logger.warning(
+                    f"Top-level agency {top_level_agency_code} not found for {agency_to_create['agency_code']}"
+                )
+
+        new_agency = factories.AgencyFactory.create(
+            agency_code=agency_to_create["agency_code"],
+            agency_name=agency_to_create["agency_name"],
+            top_level_agency=top_level_agency,
+        )
```
chouinar, line 185: Fine to just set the top level agency into the dict itself and expand it like before. Means we don't have to remember to pipe things through if we add other agency values to that config.

```suggestion
        agency_to_create["top_level_agency"] = top_level_agency
        factories.AgencyFactory.create(**agency_to_create)
```

api/tests/lib/seed_local_db.py
```diff
@@ -142,13 +161,28 @@ def _build_agencies(db_session: db.Session) -> None:
     # Create a static set of agencies, only if they don't already exist
     agencies = db_session.query(Agency).all()
     agency_codes = set([a.agency_code for a in agencies])
+    agency_dict = {a.agency_code: a for a in agencies}
```
chouinar, line 164: Wouldn't this approach require running this twice?

If you run with a fresh DB, this dict will be empty on the first run.

Maybe something like:

```py
agency_dict = {a.agency_code: a for a in agencies}

for agency_to_create in AGENCIES_TO_CREATE:
     ...

     new_agency = ...
     
     agency_dict[new_agency.agency_code] = new_agency

# a second pass
for agency_to_create in AGENCIES_TO_CREATE:
     ...
```

After typing that, it seems like this is literally just a separate version of what we'll implement in https://github.com/HHS/simpler-grants-gov/issues/2492 - it might just be easier to reuse that here once written. What do you think?

api/tests/src/db/models/factories.py
mikehgrantsgov: Removed setting the PK in the factory. Seems to work.

api/tests/lib/seed_local_db.py
mikehgrantsgov: Yeah, makes sense. What do you think about the latest change? It now works in 1 pass (last time it worked only because I had previously seeded). We could circle back and add the #2492 version later.

chouinar: Another way that might actually work without any code changes here.

In the agency definition above, do something like:
```py
AGENCIES_TO_CREATE = [
    {
       "agency_id": 1,
        "agency_code": "USAID",
        "agency_name": "Agency for International Development",
    },
    {
       "agency_id": 2,
        "agency_code": "ARPAH",
        "agency_name": "Advanced Research Projects Agency for Health",
    },
    {
       "agency_id": 3,
        "agency_code": "DOC",
        "agency_name": "Agency for International Development",
    },
    {
       "agency_id": 4,
        "agency_code": "DOC-EDA",
        "agency_name": "Agency for International Development",
        "top_level_agency_id": 3, # DOC
    },
]
```

mikehgrantsgov: That makes it much easier. One loop and we can preserve the old logic.

chouinar (approved):
LGTM!

PR #2523, opened 2024-10-18
[Issue #2521] Have the error response always return an internal request ID

### LINKED ISSUE #2521: Return a request ID in the error response to aid in debugging
### Summary

In the event that a user encounters an error while using our API, we want to return a unique ID that lets one of our developers connect it back to log messages easily.

We already generate a `request.internal_id` and attach it to all log messages for a requests lifecycle, so re-using that sounds like it would work, it's just a matter of piping this through to our error_processor that constructs the error responses. We can likely use Flask contexts (`g` probably?) to handle getting this where it needs to go.

### Acceptance criteria

- [ ] Request ID piped through for all (4xx and 5xx) errors

chouinar:
## Summary
Fixes #2521

## Changes proposed
Return the internal request ID in all error responses

## Context for reviewers
A follow-up from a discussion we had yesterday. This can aid in debugging as a user can just give us the ID and we can easily look up the corresponding logs.

Note `g` in Flask is a request-local global context (ie. each request has their own separate `g`) which we use for storing a few other logging related things already safely. https://flask.palletsprojects.com/en/3.0.x/appcontext/#storing-data

## Additional information
Screenshot 2024-10-18 at 12 15 49 PM

Screenshot 2024-10-18 at 12 16 09 PM

Screenshot 2024-10-18 at 12 17 05 PM

An example of the logs that show the internal ID - note I trimmed out the full error message as it's a very long stack trace.
Screenshot 2024-10-18 at 12 17 25 PM

### DIFF
--- a/api/src/logging/flask_logger.py
+++ b/api/src/logging/flask_logger.py
@@ -154,14 +154,17 @@ def _add_global_context_info_to_log_record(record: logging.LogRecord) -> bool:
 
 
 def _get_request_context_info(request: flask.Request) -> dict:
+    internal_request_id = str(uuid.uuid4())
+    flask.g.internal_request_id = internal_request_id
+
     data = {
         "request.id": request.headers.get("x-amzn-requestid", ""),
         "request.method": request.method,
         "request.path": request.path,
         "request.url_rule": str(request.url_rule),
-        # A backup ID in case the x-amzn-requestid isn't passed in
-        # doesn't help with tracing across systems, but at least links within a request
-        "request.internal_id": str(uuid.uuid4()),
+        # This ID is used to group all logs for a given request
+        # and is returned in the API response for any 4xx/5xx scenarios
+        "request.internal_id": internal_request_id,
     }
 
     # Add query parameter data in the format request.query.<key> = <value>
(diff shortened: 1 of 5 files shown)


api/src/logging/flask_logger.py
```diff
@@ -154,14 +154,17 @@ def _add_global_context_info_to_log_record(record: logging.LogRecord) -> bool:
...
 def _get_request_context_info(request: flask.Request) -> dict:
+    internal_request_id = str(uuid.uuid4())
+    flask.g.internal_request_id = internal_request_id
+
     data = {
         "request.id": request.headers.get("x-amzn-requestid", ""),
         "request.method": request.method,
         "request.path": request.path,
         "request.url_rule": str(request.url_rule),
-        # A backup ID in case the x-amzn-requestid isn't passed in
```
mdragon, line 162: We should make sure we don't lose track of this, either using the X-amzn one in place of this or as an additional identifier (when it's present, and falling back to one we generate when it's missing). But we don't have to change that in this PR, if we want to just open a ticket to follow up in near future Sprint.

mdragon (approved):

api/src/logging/flask_logger.py
chouinar: We'll have to see whether the x-amzn-requestid field is even populated, that might have been something specific to the past project that this originally came from. I just left that one until we have a chance to dig into it more

coilysiren:
👀 !

PR #2529, opened 2024-10-21
[Issue #2492] Create an agency linking step for the transformation process

### LINKED ISSUE #2492: Create a step for the transformation process that links agencies to their top-level agencies

mikehgrantsgov:
## Summary
Fixes #2492

## Changes proposed
Add transform step to Agency transformation
Add tests

## Context for reviewers
As part of our transformation process, we transform agency data from the legacy system, but we want to do a second pass over agency data once in our system in order to make the agency data connect together properly.

In the system, there are agencies and sub-agencies. For example, HHS is a top-level agency, and HHS-NIH11 is a subagency. We want HHS-NIH11 to point to HHS in the top_level_agency_id foreign key we setup in https://github.com/HHS/simpler-grants-gov/issues/2491

Determining whether an agency is a top-level or subagency is simple, a top-level agency will have no dashes, while a sub-agency will have one or more dashes in its agency code with the top-level agency being everything before the first dash. We want to iterate over the agency data in our system and attach appropriately.

## Additional information
See attached unit test

### DIFF
--- a/api/src/data_migration/transformation/subtask/transform_agency.py
+++ b/api/src/data_migration/transformation/subtask/transform_agency.py
@@ -285,7 +285,6 @@ class TransformAgencyHierarchy(AbstractTransformSubTask):
             if top_level_agency_code and top_level_agency_code in agency_map:
                 agency.top_level_agency = agency_map[top_level_agency_code]
 
-        self.db_session.commit()
 
     def get_top_level_agency_code(self, agency_code: str) -> str | None:
         if "-" not in agency_code:
--- a/api/src/data_migration/transformation/subtask/transform_agency.py
+++ b/api/src/data_migration/transformation/subtask/transform_agency.py
@@ -272,6 +272,27 @@ class TransformAgency(AbstractTransformSubTask):
             )
 
 
+class TransformAgencyHierarchy(AbstractTransformSubTask):
+    def __init__(self, task: Task):
+        super().__init__(task)
+
+    def transform_records(self) -> None:
+        agencies = self.db_session.scalars(select(Agency)).all()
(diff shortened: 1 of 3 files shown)


mdragon (approved):
Looks good though I'm still knocking the rust (not Rust) off my Python skills

chouinar (approved):

PR #2534, opened 2024-10-21
Add top_agency_name api

### LINKED ISSUE #2490: Update API schema to include both the agency name AND top-level agency name
### Summary

We already return the `agency_name` field in an opportunity, but want to add an additional string field called `top_agency_name`.

This field won't be populated yet as there is no corresponding field in the DB, we're just getting it available in the request for the frontend to be able to use it.

### Acceptance criteria

- [ ] API schema updated

emilycnava:
## Summary
Fixes #2490 

## Changes proposed
- Add new string field to opportunity results in api

Screenshot 2024-10-21 at 1 49 11 PM


api/src/db/models/agency_models.py
```diff
@@ -41,6 +41,7 @@ class Agency(ApiSchemaTable, TimestampMixin):
     agency_id: Mapped[int] = mapped_column(BigInteger, primary_key=True)
 
     agency_name: Mapped[str]
+    top_agency_name: Mapped[str]
```
chouinar, line 44: There shouldn't be any changes at the DB level for this PR, Mike H's recent PR added a property for this already https://github.com/HHS/simpler-grants-gov/pull/2517/files#diff-9b2da6ff61feaaaccb33635caeb9059e185c4f6c9d8334989097779e48a243aa which this PR should use.

api/src/api/opportunities_v1/opportunity_schemas.py
```diff
@@ -276,6 +283,14 @@ class OpportunityV1Schema(Schema):
         },
     )
 
+    top_agency_name = fields.String(
```
chouinar, line 286: If you use the name `top_level_agency_name` - this should work with the change Mike H added in https://github.com/HHS/simpler-grants-gov/pull/2517/files#diff-9b2da6ff61feaaaccb33635caeb9059e185c4f6c9d8334989097779e48a243aa

mdragon (dismissed):
Maybe @babebe could give this a final once over with Michael out. But I see you addressed the stuff he originally raised.

emilycnava:
> Maybe @babebe could give this a final once over with Michael out. But I see you addressed the stuff he originally raised.

Agreed, not sure why the ERD diagram test is failing, looks like git is just failing to load or something weird.

r.e. the api check, i'm wonder, should i just remove the synopsis mixin db change, or do we want that and I should just make a new migration?

babebe:
LGTM

api/src/db/models/legacy_mixin/synopsis_mixin.py
```diff
@@ -31,6 +31,7 @@ class TsynopsisMixin:
     ac_email_addr: Mapped[str | None]
     ac_email_desc: Mapped[str | None]
     agency_name: Mapped[str | None]
+    top_level_agency_name: Mapped[str | None]
```
chouinar, line 34:
```suggestion
```

api/src/db/models/legacy_mixin/synopsis_mixin.py
```diff
@@ -71,6 +72,7 @@ class TsynopsisHistMixin:
...
```
chouinar, line 75: If we keep this in this file, it will break the ELT process as these columns both need to exist in the legacy system and exactly match the order of the legacy system.
```suggestion
```

chouinar (changes requested):

chouinar (approved):
LGTM!

chouinar (approved):

PR #2540, opened 2024-10-21
[Issue 2447] add relevancy sort option to opportunity search

### LINKED ISSUE #2447: Implementation for enabling users to sort by relevancy
### Summary

Implementation for the design from #2264.

### Acceptance criteria

- [ ] User can explicitly sort by relevancy, relevancy sort used when user enters search terms

doug-s-nava:
## Summary
Fixes #2447

## Changes proposed

Following note from the ticket at https://github.com/HHS/simpler-grants-gov/issues/2447#issuecomment-2427653490, this change:

* updates the default sort option on the opportunities search to "relevancy" (with a built in secondary sort of "post date descending", as previously implemented on the API)
* creates a new sort option in the UI titled "Default / Most Relevant" to be used as the default sort option

Also includes a couple of updates for stronger typing and improve testing of related components.

Note requirements update after review:

* update sort select option labels with new caps scheme
* add a "Sort by" label to the left of the sort select, shifting to above the select at mobile widths

### Testing UI behavior

1. ensure you have local opensearch seeded (if not, run `make populate-search-opportunities` in the api directory)
1. visit local search page
2. _VERIFY_ the default selected sort option is "Default / Most Relevant"
2. _VERIFY_ search results are sorted in reverse chronological order by post date
3. enter a search term such as "grant" and click submit
4. _VERIFY_ search results are sorted by relevance in reverse chronological order by post date (since relevance is not visible directly in the UI, to verify further, see below)
5. select a new sort option
6. _VERIFY_ results are now sorted by your selected option
7. hard refresh or copy and paste the url for the page into a new incognito browser
8. _VERIFY_ sort selection is retained

### Testing sort results

1. add `relevancy_score = fields.Decimal()` on line 310 of `opportunity_schemas.py` and rebuild and restart the API container. `make stop && make build && make start` in the API directory should do it
9. put a debugger at line 66 of `SearchOpportunityApi.ts` and load the search page, performing various searches adn sorts
10. looking at the response you can confirm that the response array is sorted primarily on descending relevancy, and secondarily on descending post_date by default, but updates as you choose other options

### Screenshots
Screenshot 2024-10-22 at 1 17 28 PM

Screenshot 2024-10-22 at 9 49 21 AM


api/src/services/opportunities_v1/search_opportunities.py
```diff
@@ -99,7 +99,7 @@ def _get_sort_by(pagination: PaginationParams) -> list[tuple[str, SortDirection]
 
     # Add a secondary sort for relevancy to sort by post date (matching the sort direction)
     if pagination.order_by == "relevancy":
-        sort_by.append((_adjust_field_name("post_date"), pagination.sort_direction))
+        sort_by.append((_adjust_field_name("post_date"), SortDirection.DESCENDING))
```
chouinar, line 102: Why did we change this? This will make it so reversing the sorting doesn't reverse the results.

doug-s-nava: As far as I can tell there will be no way to hit this branch of the code with the `sort_direction` set to ascending. We are not going to allow the user to sort by ascending relevancy in the UI, as showing the least relevant results first doesn't seem like a common or expected use case. Since the sort_direction would come from the primary sort option, in this case relevancy, that would mean that the secondary sort would always come out as descending in this case.

That being the case we could still leave this code as is and it would always come out as descending, no impact to the user or the functionality of the code. However, I think it is could be best to hardcode the secondary sort behavior for relevancy to descending post date here for clarity and to avoid confusion. When I looked at the code I had to do a bunch of thinking to figure out if we could end up in a case where the secondary sort would actually be post date ascending, so if our intention is for that to not be possible we may as well hard code it, is where I ended up.

This is still a draft though, so I may change my mind before I open for review.

chouinar: I agree that I'm not sure in what scenario a user would want to sort by score/relevancy ascending, but at least on the API side I'd prefer we leave it as it was - the API can be used directly as well without the front-end so we need to consider the behavior there.

Likely what will make sense long-term is to remove this and just allow for the API request to specify a secondary sort directly.

chouinar: If we wanted to restrict relevancy sort to not allow ascending at all on the API, that would also be reasonable I think - although that change would require a little more work in the API schema to setup.

doug-s-nava: I went ahead and reverted the change. I think it'd be better to be explicit about this sort behavior to avoid confusion, but that can wait for another ticket

coilysiren (commented):
Deferring my review

acouch:
Code looks good!

> VERIFY the default selected sort option is "Default / Most Relevant"

@andycochran @doug-s-nava why not default to sort by newest and then switch to relevancy if there is a term and no other sort filter? That is the way I've seen most search engines with default values work. It seems confusing to start with "Default / Most Relevant" because they are not sorted that way. "Default" seems confusing as well because we don't label other filters that way.

acouch (approved):

doug-s-nava:
@acouch I'll let Andy respond with his take, but my recollection on the decision was that this had to do with:
* the complexity of secondary sorting - currently the default sort is by post date descending, and a relevancy sort is set up with a secondary sort of post date ascending. The result is that if you sort by relevancy with no search term, you end up with a list sorted by post data descending, as relevancy comes back as 0 for all results. This gives us a way to leverage that secondary sort behavior without exposing it to users
* the complexity of changing default sort behavior based on a user interaction that users may not expect to change sort behavior - if we change the sort when a user enters a search term, I think there was a concern that this could be confusing for users

Apologies if this is mischaracterizing or overly confusing - late Friday brain over here

acouch:
@doug-s-nava Thanks, the result just seems more confusing to me for users for the reasons mentioned above.

andycochran:
I understand Aaron's concerns. With the solution we landed on, I think what has the most potential to confuse users is the label of the default sort. It will behave intuitively, as expected: defaulting to reverse chron, auto switching to relevancy upon entering terms. We switch the sort without switching the dropdown option, since the user has not made an explicit sort decision. Maybe the default sort option's label doesn't clearly communicate the two default cases. 

We'd quickly landed on "Default / most relevant" to have something shorter than the literal "Posted Date (newest) without search terms _OR_ Most Relevant if you've entered search terms." Perhaps @michellemin-nava has an idea of how to better label this default option. 

Maybe this? 
- Default: Posted Date (newest) / most relevant

Maybe we dynamically show what the current default is (tho this changing might be confusing): 
- Default: Posted Date (newest)
- Default: Most Relevant

Regardless, we'll get user feedback on this.

acouch:
@andycochran yeah it is tricky. Glad we are thinking it through. Personally would opt for not using the term "Default", but am 💚 with the approach and, as you said we can see what the users think.

PR #2549, opened 2024-10-23
[Issue 2487] relevant uswds and trussworks upgrades

### LINKED ISSUE #2487: Evaluate updating of USDWDS to version 9.1+
Following up on #2096 which determined the issue was in the library we depend on.

Investigate what an update from 7 to 9.1+ would involve. If it appears to be easy/well supported and we can get a good sense of what would be affected we can just go ahead and make the leap. If it's more complicated spin up some tickets to cover this work

doug-s-nava:
## Summary
Fixes #2487

## Changes proposed

Upgrade Trussworks to latest version, upgrade USWDS to latest version supported by Trussworks.

## Context for reviewers

See notes here: https://github.com/HHS/simpler-grants-gov/issues/2487#issuecomment-2432512375

### Test steps

**Note:** it's probably too early to implementing anything, but how do we feel about visual diffing tools like Percy?

These test steps are what I did, in order to allow an easy-ish side by side comparison of main vs this branch, but you could take screenshots of the main branch pages and compare to screenshots of this branch, or there are probably other options as well. The point is, do a visual diff of the site.

1. _VERIFY_ all CI build stuff passes
2. create an alternate simpler repo either by re-cloning into a new directory, or running a `cp` into a new directory
3. ensure that your alternate directory is on main and your regular directory is on this branch, or vice versa, and `npm ci` in both directories
4. start servers from both directories, these will start on locahost:3000 and localhost:3001
5. run the apps side by side and do a rough visual diff check on every page to make sure nothing is different
6. run some searches and look at opportunities as well, to make sure functionality is the same

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -28,9 +28,10 @@
   "dependencies": {
     "@next/third-parties": "^14.3.0-canary.59",
     "@opentelemetry/api": "^1.8.0",
-    "@trussworks/react-uswds": "^7.0.0",
-    "@uswds/uswds": "^3.6.0",
+    "@trussworks/react-uswds": "^9.1.0",
+    "@uswds/uswds": "^3.7.1",
     "clsx": "^2.1.1",
+    "focus-trap-react": "^10.2.3",
     "isomorphic-dompurify": "^2.15.0",
     "js-cookie": "^3.0.5",
     "lodash": "^4.17.21",
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -11,9 +11,10 @@
       "dependencies": {
         "@next/third-parties": "^14.3.0-canary.59",
         "@opentelemetry/api": "^1.8.0",
-        "@trussworks/react-uswds": "^7.0.0",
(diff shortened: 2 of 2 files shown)


frontend/package.json
```diff
@@ -28,9 +28,10 @@
   "dependencies": {
     "@next/third-parties": "^14.3.0-canary.59",
     "@opentelemetry/api": "^1.8.0",
-    "@trussworks/react-uswds": "^7.0.0",
-    "@uswds/uswds": "^3.6.0",
+    "@trussworks/react-uswds": "^9.0.0",
+    "@uswds/uswds": "^3.7.1",
     "clsx": "^2.1.1",
+    "focus-trap-react": "^10.2.3",
```
doug-s-nava, line 34: a new peer dependency of Trussworks that we need to include manually with the latest version

emilycnava (approved):
Duplicated repo locally and poked around for a bit, no noticeable diffs in any of the major UX component heavy pages that I could see

acouch (approved):

PR #2550, opened 2024-10-23
[Issue #2389] Test and adjust Opportunity ORM relationships

### LINKED ISSUE #2389: Validate / update our relationship logic in the opportunity model ORMs
### Summary

As a follow-up to https://github.com/HHS/simpler-grants-gov/issues/2387 - I want us to verify there aren't any other relationships between tables that would cause issues.

The best way of handling this is to create a dedicated set of tests that sets up data, deletes a record from within that data, and verifies what records remain. Having this as a test also helps work around any issue with someone changing this in the future in an unexpected way.

### Acceptance criteria

- [ ] Tests written
- [ ] Any adjustments to relationships made to our ORM models to correct issues identified

chouinar:
## Summary
Fixes #2389 

## Changes proposed
Add tests to validate the behavior of our opportunity relationships, mostly in regards to deletion events

Fix one faulty relationship

Fix a mypy issue regarding the interactive console script (a recent mypy upgrade doesn't like this script's imports, so move it to a place that won't get checked)

## Context for reviewers
Our ORM tries to be smart / we can configure our ORM to be smart when we delete an object. For example, if we do a `db_session.delete(some_opportunity_record)` the ORM will first figure out what depends on this record via our relationships, and delete other records (for example, an opportunity summary) in the proper order saving us a lot of complexity in the deletes.

However, sometimes the ORM doesn't quite choose right (deleting a child object deletes the whole thing). There have been a few bug fixes I've made to address this, and these tests confirm we've got those issues worked out + found another one.

Without the change I made here, deleting an opportunity makes the ORM try to null out a non-null column in the agency table (why that is the default behavior, I'm not sure). By making it a `viewonly` relationship, the ORM knows not to cascade across the relationships.


api/src/db/models/opportunity_models.py
```diff
@@ -70,7 +70,10 @@ class Opportunity(ApiSchemaTable, TimestampMixin):
     )
 
     agency_record: Mapped[Agency | None] = relationship(
-        Agency, primaryjoin="Opportunity.agency == foreign(Agency.agency_code)", uselist=False
+        Agency,
+        primaryjoin="Opportunity.agency == foreign(Agency.agency_code)",
+        uselist=False,
+        viewonly=True,
```
chouinar, line 76: The ELT job is failing in dev because of this:
```
"(psycopg.errors.NotNullViolation) null value in column \"agency_code\" of relation \"agency\" violates not-null constraint\nDETAIL:  Failing row contains (914, MicroHealth LLC, null, MH, 00.00, 3, 914, f, MH, MicroHealth LLC, MicroHealth LLC, f, f, t, f, f, f, 2021-09-17 16:06:49+00, 2024-10-23 15:53:13.128999+00, null).\n[SQL: UPDATE api.agency SET agency_code=%(agency_code)s::VARCHAR, updated_at=%(updated_at)s::TIMESTAMP WITH TIME ZONE WHERE api.agency.agency_id = %(api_agency_agency_id)s::BIGINT]\n[SQL parameters hidden due to hide_parameters=True]\n(Background on this error at: https://sqlalche.me/e/20/gkpj)",
```

We don't even have a true foreign key here, it's a bit odd that this is the default behavior, but this fixes it.

mikehgrantsgov (approved):
LGTM!

PR #557, opened 2023-09-28
[ Issue 548 ] Production Cert for Simpler Domain

SammySteiner:
## Summary
Fixes #548 

## Changes proposed

- Add frontend prod certificate
- Update ternary to send cert to service module

PR #2556, opened 2024-10-24
[Issue #2351] Perform first platform upgrade

coilysiren:
## Summary

Relates to https://github.com/HHS/simpler-grants-gov/issues/2351


infra/frontend/service/variables.tf
```diff
@@ -8,9 +8,3 @@ variable "image_tag" {
   description = "image tag to deploy to the environment"
   default     = null
 }
-
-variable "domain" {
-  type        = string
-  description = "DNS domain of the website managed by HHS"
-  default     = null
-}
```
coilysiren, line 16: this has always been null! isn't never been used! even inside of the PR where it was added!

https://github.com/HHS/simpler-grants-gov/pull/557

infra/frontend/service/main.tf
```diff
@@ -1,12 +1,10 @@
-# docs: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/vpc
```
coilysiren, line 1: I used to add these comments, mostly for myself, but then I found a VS code extension that provides these links automatically. 

https://marketplace.visualstudio.com/items?itemName=TylerHarris.terraform-link-docs

PR #2557, opened 2024-10-24
[Issue #2351] Move alarms into accounts folder

coilysiren:
## Summary

Relates to https://github.com/HHS/simpler-grants-gov/issues/2351

## Changes proposed

Moves the account level alarm configuration from `infra/alarms/main.tf` into `infra/accounts/alarms.tf`

## Context for reviewers

This came up while working on the platform update CLI. The way its structured, it assumed the `alarms` folder was an entire service. It very much is not!

I put some thought into this, and looked for a place where "AWS account level" terraform should go. Well, it goes in here, in `infra/accounts/`. Everything in this folder is deployed once per AWS account.


chouinar (approved):
Makes sense to me

Issue #2512, opened 2024-10-17
Setup presigning the URLs for opportunity attachments


chouinar:
### Summary

We want the API to presign the s3 URLs in our GET /opportunity endpoint. Presigning a URL makes it so for a period of time, a user can directly access the file.

As the paths we'll have stored in the attachments table are s3 paths, the process should roughly be:
* fetch the opportunity
* iterate over the opportunity attachment records, presign each file_location parameter (note - as a sanity check, make sure this isn't actually modifying the value in the DB)
* return the opportunity

---
We also want this to work with our tests and locally. in our tests we mock out s3, and locally we run localstack which is effectively a different mocked s3 that you can access. We want this approach to work fully locally. The s3 path returned should return a valid link that you can download from. I've tested previously that presigned urls do work locally with localstack. The primary complexity here is setting up the data, our factories should probably be setup to generate a set of static s3 paths (say 5 different files we pre-upload for tests + during the db-seed-local script) that can be randomly picked.

---

I've tested presigning the URLs before in this very hacky closed PR: 
https://github.com/navapbc/simpler-grants-gov/pull/68/files

The relevant code is in the opportunity routes - ignore anything regarding redirects.

Open questions:
* How will the AWS access work exactly? Do we want to assume the ecs task role for doing this?

### Acceptance criteria

- [ ] The GET /opportunity endpoint returns presigned URLs
- [ ] The factories have been modified to set the file_location to one of a set of files that a fixture will pre-upload to s3
- [ ] The db-seed-local script has been modified to upload a few files that are attached to the opportunities

babebe:
@chouinar Per your suggestion i could start working on the first portion of this with the first 2 criteria's. Do i create a new issue to segregate the two.

mdragon:
I don't think we need a separate ticket now for the seeds. If we decide different we can always split it later, but i think the seeds with the code they facilitate testing with is a good pattern to try to follow.

PR #2563, opened 2024-10-24
[Issue 2512] S3 Presign URL

babebe:
## Summary
Fixes #{[2512](https://github.com/HHS/simpler-grants-gov/issues/2512)}

## Changes proposed
Added test files for opportunity attachment
Updated local seed db with func to upload test files to S3
Updated factories to set opportunity attachment "file_location" from any value in static list 
Updated the response schema to return new field `download_path` which is the link to the pre-signed url
Added  default Expiration limit for the s3 url in s3 config
Added reusable function to get aws  boto3 session for local development
Added new env variable for local development
Added fixture to upload test files 
Added unit test. Tests if we can successfully download s3 file.

## Context for reviewers
Return a pre-signed url for an attachment a user can then use to download the s3 file


api/src/services/opportunities_v1/get_opportunity.py
```diff
@@ -31,7 +34,30 @@ def _fetch_opportunity(
...
+
+    s3_config = S3Config()
+    s3_config.s3_endpoint_url = "http://localstack:4566"
+    s3_client = get_s3_client(
+        s3_config, boto3.Session(aws_access_key_id="NO_CREDS", aws_secret_access_key="NO_CREDS")
+    )
+
+    for opp_attachment in opportunity_copy.opportunity_attachments:
+        file_loc = opp_attachment.file_location
+        object_key = file_loc.split(f"{s3_config.s3_opportunity_bucket}/")[
```
chouinar, line 50: We have a file_util which has a split function for s3 urls: https://github.com/HHS/simpler-grants-gov/blob/main/api/src/util/file_util.py#L18 - which handles getting the bucket + key to pass into the next function.

While the s3 config here has a bucket configured, we should always just use whatever is in the object itself.

api/src/services/opportunities_v1/get_opportunity.py
```diff
@@ -31,7 +34,30 @@ def _fetch_opportunity(
...
+            1
+        ]  # should be same as filename in db ?
+        pre_sign_file_loc = s3_client.generate_presigned_url(
+            "get_object",
+            Params={"Bucket": s3_config.s3_opportunity_bucket, "Key": object_key},
+            ExpiresIn=1800,
+        )  # do we want to put limit
+        opp_attachment.file_location = pre_sign_file_loc
```
chouinar, line 58: Hmm, an idea that might be better and let us work around the copy issue.

If we did `opp_attachment.file_path = pre_sign_file_loc` (or some other name that isn't one of the DB values) that would attach the value to the object, but not modify anything relevant to the ORM. We'd also have to change the API to use `file_path` (or whatever name). 

Happy to help brainstorm - this is the main thing I wasn't sure how we wanted to do properly.

api/src/services/opportunities_v1/get_opportunity.py
```diff
@@ -31,7 +34,30 @@ def _fetch_opportunity(
...
+
+    for opp_attachment in opportunity_copy.opportunity_attachments:
+        file_loc = opp_attachment.file_location
+        object_key = file_loc.split(f"{s3_config.s3_opportunity_bucket}/")[
+            1
+        ]  # should be same as filename in db ?
+        pre_sign_file_loc = s3_client.generate_presigned_url(
+            "get_object",
+            Params={"Bucket": s3_config.s3_opportunity_bucket, "Key": object_key},
+            ExpiresIn=1800,
```
chouinar, line 56: I'd say we define something in the s3_config class with a default of 1800 so we can override it if we want something different.

api/tests/lib/opportunity_attachment_test_files/test_file_2.txt
```diff
@@ -0,0 +1 @@
+Hello, world
```
chouinar, line 1: Some slight variance in the files so they're "distinct"
```suggestion
Hello, world - again!
```

api/src/services/opportunities_v1/get_opportunity.py
```diff
@@ -31,7 +34,30 @@ def _fetch_opportunity(
...
 
 def get_opportunity(db_session: db.Session, opportunity_id: int) -> Opportunity:
-    return _fetch_opportunity(db_session, opportunity_id, load_all_opportunity_summaries=False)
+    opportunity = _fetch_opportunity(
+        db_session, opportunity_id, load_all_opportunity_summaries=False
+    )
+    opportunity_copy = copy.deepcopy(opportunity)
+
+    s3_config = S3Config()
+    s3_config.s3_endpoint_url = "http://localstack:4566"
```
chouinar, line 43: This should be getting loaded from env vars properly, were you hitting an issue where it wasn't?

api/src/services/opportunities_v1/get_opportunity.py
babebe: great!

api/src/services/opportunities_v1/get_opportunity.py
babebe: We could also create a method in the model class for updating field

api/tests/lib/seed_local_db.py
```diff
@@ -173,14 +201,16 @@ def seed_local_db(iterations: int, include_history: bool) -> None:
...
+        #
+        # with db_client.get_session() as db_session:
+        #     factories._db_session = db_session
+        #
+        #     _build_opportunities(db_session, iterations, include_history)
+        #     # Need to commit to force any updates made
+        #     # after factories created objects
+        #     db_session.commit()
+        #
+        #     _build_agencies(db_session)
```
chouinar, line 216: I think you left this in from testing

api/src/services/opportunities_v1/get_opportunity.py
```diff
@@ -35,8 +40,46 @@ def _fetch_opportunity(
...
+        bucket, key = split_s3_url(file_loc)
+        pre_sign_file_loc = s3_client.generate_presigned_url(
+            "get_object",
+            Params={"Bucket": bucket, "Key": key},
+            ExpiresIn=s3_config.presigned_s3_duration,
+        )
+        if s3_config.s3_endpoint_url:
+            pre_sign_file_loc = pre_sign_file_loc.replace(
+                s3_config.s3_endpoint_url, "http://localhost:4566"
+            )
```
chouinar, line 68: I'd suggest we add a comment calling out this hacky fix as only mattering locally due to docker path issues.

chouinar (commented):
Looks good, tested it locally, just a few tidying up suggestions

api/tests/lib/seed_local_db.py
babebe: Yup!

api/src/services/opportunities_v1/get_opportunity.py
babebe: Should i mark it as todo as well ?

chouinar (approved):
LGTM!

PR #2564, opened 2024-10-24
[Unticketed] Fix agency relationship loading in the GET opp endpoint

chouinar:
## Changes proposed
Explicitly load the agency relationships upfront in the GET /opportunity endpoint

## Context for reviewers
By default SQLAlchemy lazy-loads all relationships, we had configured the opportunity fetch to load `*` relationships, but the way our agency relationships are setup don't seem to work without explicitly defining it. So I did that.

Need to do some additional configuration to make this work with a unit test which I'll follow-up with, but testing locally it stopped erroring.

## Additional information
The original error: `"Parent instance <Agency at 0x7f498eee92b0> is not bound to a Session; lazy load operation of attribute 'top_level_agency' cannot proceed (Background on this error at: https://sqlalche.me/e/20/bhk3)",`


doug-s-nava (approved):
seems to have fixed the issue on my end

PR #2566, opened 2024-10-24
[Issue #2371] Rename agency to agency_code in API + DB

### LINKED ISSUE #2371: Look into adjusting the naming / data structure of the agency data in the opportunity response
### Summary

Right now an opportunity returns the following relevant agency information in our API (fake data):

```json
{
    "agency": "HHS-NIH11",
    "top_level_agency_name": "Department of Health and Human Services",
    "agency_name": "National Institutes of Health",
    "summary": {
      "agency_code": "USAID",
      "agency_contact_description": "google.com Contact Center\nHours of operation are 24 hours a day, 7 days a week.\nchristinarosales@example.org",
      "agency_email_address": "mark90@example.net",
      "agency_email_address_description": "Contact National Housing Administration via email",
      "agency_name": "National Housing Administration",
      "agency_phone_number": "123-456-0002",
    }
  }
```

The agency data is stored across two parts of the API response, two fields in the top-level, and the rest in the summary object. The agency code, and name appear twice as well, and are not consistent with one another.

This is almost certainly going to be confusing for users of the API, and we should correct this issue.

--

Context on why it is this way:
* The top-level `agency` field is actually the agency code of the opportunity
* The top-level `agency_name` is found by joining the agency field with the agency table
* The summary fields are fields attached to the opportunity summary itself in the DB, which is partially a duplicate of the top-level values and a static copied value of the actual agency info

From prior investigation, the agency code is almost never different between the two levels, but there are a few specific cases where they are (not intentional as far as I can tell - likely just an old bug). The top-level values are better supported/accurate and we shouldn't keep the agency code / agency name of the inner summary object.

--

Recommendation:
* Remove agency_code and agency_name from the summary object
* Rename the "agency" field in the top-level object to "agency_code" to be clearer.

As the only user of our API at this time is the front-end, we would just need to temporarily add another field at the top-level, and have them switch over.

This will make the request look like:
```json
{
    "agency_code": "HHS-NIH11",
    "top_level_agency_name": "Department of Health and Human Services",
    "agency_name": "National Institutes of Health",
    "summary": {
      "agency_contact_description": "google.com Contact Center\nHours of operation are 24 hours a day, 7 days a week.\nchristinarosales@example.org",
      "agency_email_address": "mark90@example.net",
      "agency_email_address_description": "Contact National Housing Administration via email",
      "agency_phone_number": "123-456-0002",
    }
  }
```

--

Note that the DB will continue to contain a few extra fields for now, we can clean that up later.

### Acceptance criteria

- [ ] Run this idea past product + front-end folks
- [ ] Make the change

chouinar:
## Summary
Fixes #2371

## Changes proposed
High-level - renamed `agency` in the opportunity models to `agency_code` - both at the API and DB layer

Remove agency name and agency code from the opportunity summary response object (should not be used - the opportunity values are more accurate)

A lot of fixes for tests that were checking for the old name

Deleted an import-csv script that wasn't needed anymore (and would've needed fixes)

## Context for reviewers
`agency` is way too generic of a name, and `agency_code` makes more sense.

Agency info is also split across an opportunity and summary. Some of this was duplicated (code and name) and we want to avoid confusion by having these values duplicated.

## Additional information
Did a quick pass on the existing search UI, don't see anything having broken yet. Query box, sorting, and filters all work, as well as the agency itself being displayed in search. I also switched the frontend to use v1 and didn't see any issues

Screenshot 2024-10-25 at 3 33 08 PM
Screenshot 2024-10-25 at 3 33 27 PM


doug-s-nava:
Made a couple of notes on the ticket but I agree that everything seems to check out here. We'll have some minor updates to make on the frontend before deprecating the existing `opportunity.agency` but we can coordinate that later.

emilycnava (approved):
lgtm! (checked locally and didn't see any noticeable errors on FE)

emilycnava:
@doug-s-nava I wonder I know you mentioned visual diff checkers in a PR the other day, I assume something like that would have helped here as well? How complex are those tools to implement?

doug-s-nava:
Getting an error on `make migrate-db-down`: `psycopg.errors.UndefinedColumn) column "agency_code" does not exist`

doug-s-nava:
@emilycnava 🤔 I feel like a visual differ could help in situations like this to uncover bugs, but I think e2e tests would be even more effective with that since they take user interaction into account. The way we're using e2e tests at the moment, we wouldn't have caught something like a text change here, though. In any case, for a testing scenario to catch an issue in a change like this we'd have to carefully choose the examples we're using. For example, to catch a bug with how agency data is displayed we'd have to make sure to be running our tests against an opportunity that is affected by the change - an opportunity with no agency data or, perhaps, agency data that matches between the opportunity and the summary, may not show us that anything is wrong. Interesting to think about for the future.

Visual differs are not super hard to set up. However:
* they cost money
* they can be finicky similar to how e2e test are in terms of false negatives, and as a result there is some overhead in terms of maintenance

PR #2568, opened 2024-10-25
Kai/v0.3

coilysiren:
## Summary
Fixes #{ISSUE}


.template-infra/app-analytics.yml
```diff
@@ -0,0 +1,3 @@
+_commit: 929a959
```
chouinar, line 1: Not for this PR - but in the template repo, I might recommend a README.md file also be copied over that gives some context on what these files are / what the template repo is.

Just something linking back to that repo and saying "hey we manage a lot of shared configuration that you can pull in, here's the link to the docs".

chouinar (approved):
LGTM!

.template-infra/app-analytics.yml
coilysiren: > Not for this PR - but in the template repo

Big agree 👍🏼

PR #2579, opened 2024-10-25
Return notFound if opportunity not found

### LINKED ISSUE #2337: Opportunity "not found" page does not return 404
### Summary

The current implementation of the opportunity page provides a 404 if the opp id doesn't exist in the API, however the return code is a 200.

Returning the `notfound()` function in local testing fixes that:

```javascript

  try {
    const opportunityData = await getOpportunityData(id);
    const meta: Metadata = {
      title: `${t("OpportunityListing.page_title")} - ${opportunityData.opportunity_title}`,
      description: t("OpportunityListing.meta_description"),
    };
    return meta;
  } catch (error) {
    return notFound();
  }
}

```

### Acceptance criteria

- [ ] not found opportunity returns a 404 status code

mikehgrantsgov:
## Summary
Fixes #2337

## Changes proposed
Return notFound() if opportunity API returns a 404 error code.

## Context for reviewers
See screen capture to demonstrate local env returning 'Oops! Page Not Found' title vs current dev showing 'Opportunity Listing' title:
https://github.com/user-attachments/assets/31a6ba1b-8e9f-4df7-9a79-bd72ac7ca8e1

### DIFF
--- a/frontend/src/app/[locale]/opportunity/[id]/page.tsx
+++ b/frontend/src/app/[locale]/opportunity/[id]/page.tsx
@@ -6,6 +6,7 @@ import withFeatureFlag from "src/hoc/search/withFeatureFlag";
 import { Opportunity } from "src/types/opportunity/opportunityResponseTypes";
 
 import { getTranslations } from "next-intl/server";
+import { notFound } from "next/navigation";
 import { GridContainer } from "@trussworks/react-uswds";
 
 import BetaAlert from "src/components/BetaAlert";
@@ -26,6 +27,7 @@ export async function generateMetadata({ params }: { params: { id: string } }) {
     title = `${t("OpportunityListing.page_title")} - ${opportunityData.opportunity_title}`;
   } catch (error) {
     console.error("Failed to render title");
+    return notFound();
   }
   const meta: Metadata = {
     title,


frontend/src/app/[locale]/opportunity/[id]/page.tsx
```diff
@@ -26,6 +27,7 @@ export async function generateMetadata({ params }: { params: { id: string } }) {
     title = `${t("OpportunityListing.page_title")} - ${opportunityData.opportunity_title}`;
   } catch (error) {
     console.error("Failed to render title");
+    return notFound();
```
Comment (chouinar) [inline frontend/src/app/[locale]/opportunity/[id]/page.tsx:30]: More a design question, but would we want to display a notFound page if the API returned a 5xx? Looks like

Comment (mikehgrantsgov) [reply frontend/src/app/[locale]/opportunity/[id]/page.tsx:30]: Maybe a follow-up ticket to create a `500.js` error page and show that? https://nextjs.org/docs/pages/building-your-application/configuring/error-handling#handling-server-errors

doug-s-nava (approved):
This seems to work as far as returning a 404 status code to the browser on the request for the page when there is an error, so I'm good to sign off.

A couple things to think about:

* As this stands, we'll return a `not found` regardless of what the error is in the API interaction, which could result in some confusion if the error is actually a 500 and the ID provided is correct
* That problem also exists at 98
* This `notFound` call that is setting the status code is only being set in the `generateMetadata` code. I guess Next is smart enough to use that as we are not explicitly setting a code anywhere else, but I wonder if we need to try and be consistent in other places where we are making this same call out to the API.
I think it's fine if we don't want to deal with the complexity of 404 vs 500 error cases in this change - I have a branch partially implemented for the fix and can tackle it separately.

Issue #2580, opened 2024-10-25
Fix grype output to show CVE version number and fix


coilysiren:
### Summary

Followup to https://github.com/HHS/simpler-grants-gov/pull/2542/

### Acceptance criteria

\\

coilysiren:
> could just run the same scan twice with both outputs

@acouch this is exactly what I was thinking about earlier

PR #2598, opened 2024-10-28
[Issue #2580] Run vulnerability scans twice

coilysiren:
## Changes

Runs vulnerability scans twice so we can get both the version fixes and CVE locations

## Context 

This is a patchwork fix for #2580. I would rather fix this upstream, so I'm not making this PR as resolving #2580


acouch (approved):
Thanks!

PR #2599, opened 2024-10-28
[Issue ##2798] Break up e2e tests

### LINKED ISSUE #2798: Shard e2e tests to make them faster and more reliable
### Summary

The e2e tests could be sped up and made less "flaky" by sharding them so multiple tests run at the same time. Playwright supports sharding.

### Acceptance criteria

- [ ] tests split up and run faster

acouch:
## Summary
Fixes #2798

## Changes proposed
Update e2e tests so they are sharded across 3 processes that run in parallel. Time per test reduced by a couple of mins. More importantly the tests seem less flaky. This not as big a deal now but the tests should scale as we grow.

### DIFF
--- a/.github/workflows/ci-frontend-e2e.yml
+++ b/.github/workflows/ci-frontend-e2e.yml
@@ -1,5 +1,4 @@
 name: Frontend E2E Tests
-
 on:
   workflow_call:
   pull_request:
@@ -25,6 +24,11 @@ jobs:
     name: Run E2E Tests
     runs-on: ubuntu-latest
 
+    strategy:
+      matrix:
+        shard: [1, 2, 3]
+        total_shards: [3]
+
     steps:
       - uses: actions/checkout@v4
 
@@ -35,10 +39,9 @@ jobs:
           cache: ${{ env.PACKAGE_MANAGER }}
           cache-dependency-path: ${{ env.LOCKFILE_PATH }}
 
-      - run: npm ci
-
-      - name: Install Playwright Browsers
-        run: npx playwright install --with-deps
+      - run: |
+          npm ci
+          npx playwright install --with-deps
 
       - name: Start API Server for e2e tests
         run: |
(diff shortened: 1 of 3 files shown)


rylew1:
If you want to shave some time on the docker builds, have you tried to use npm commands in the Dockerfile - can remove the make install and copying of the Makefile over .

.github/linters/templates/csv.tmpl
```diff
@@ -0,0 +1,4 @@
+"Package","Version Installed","Vulnerability ID","Severity","Locations"
+{{- range .Matches}}
+"{{.Artifact.Name}}","{{.Artifact.Version}}","{{.Vulnerability.ID}}","{{.Vulnerability.Severity}}","{{ (index .Artifact.Locations 0).RealPath }}"
+{{- end}}
```
rylew1, line 4: What is this for? 🤔

coilysiren (approved):
Cool! Haven't seen the shards thing before

.github/workflows/ci-frontend-e2e.yml
```diff
@@ -50,15 +53,75 @@ jobs:
...
         run: |
-          npm run build
           cat .env.development >> .env.local
-          npm run test:e2e
+          npm run build
+
+      - name: Run e2e tests (Shard ${{ matrix.shard }}/${{ matrix.total_shards }})
+        env:
+          CI: true
+          TOTAL_SHARDS: ${{ matrix.total_shards }}
```
doug-s-nava, line 64: curious about the decision to use environment variables, which I can't find a reference to in the docs, versus the `--shards` flag that seems to be recommended https://playwright.dev/docs/test-sharding#sharding-tests-between-multiple-machines

.github/workflows/ci-frontend-e2e.yml
```diff
@@ -50,15 +53,75 @@ jobs:
           ../api/bin/wait-for-api.sh
         shell: bash
 
-      - name: Run E2E Tests
+      - name: Build a prod version of the site
         run: |
-          npm run build
           cat .env.development >> .env.local
-          npm run test:e2e
+          npm run build
```
doug-s-nava, line 59: it looks like since the sharded test runs are all going in parallel that the npm and playwright setup steps are not able to be cached, so 5 mins of processing time is being spent on each run on something that it seems like could be shared. I'm a novice with github actions, but is this a place where a reusable workflow could help? https://docs.github.com/en/actions/sharing-automations/reusing-workflows#creating-a-reusable-workflow

acouch: Yes, this would be a good follow up.

acouch: I created ticket https://github.com/HHS/simpler-grants-gov/issues/2813

Issue #2582, opened 2024-10-25
Resolve golang gh CLI CVE-2024-34158


coilysiren:
### Summary

first found here: https://github.com/HHS/simpler-grants-gov/pull/2572

### Acceptance criteria

\\

coilysiren:
we should create an issue here: https://github.com/cli/cli/issues?q=is%3Aissue+is%3Aopen+CVE

coilysiren:
The vulnerability is in the Github CLI here: https://github.com/cli/cli/blob/trunk/go.mod#L5

The fix is shown here: https://github.com/HHS/simpler-grants-gov/actions/runs/11558432017/job/32170953451?pr=2606 eg. upgrading golang to at least `1.23.1`

coilysiren:
Superseded by https://github.com/HHS/simpler-grants-gov/issues/2590

Issue #2590, opened 2024-10-25
[Analytics] Call GitHub graphql API directly from python


widal001:
### Summary

Update the `analytics.integrations.github` sub-package to call the GitHub GraphQL API directly from python instead of using the `gh` CLI as an intermediary.

### Motivation

- The `gh` CLI uses a pinned version of go that has a CVE
- Simplifies the Dockerfile
- Removes a somewhat clunky sub-process call
- Enables us to add better error handling for GitHub API responses
- Enables us to mock the GitHub API for tests

### Acceptance criteria

- [ ] The `analytics.integrations.github` sub-package calls the GitHub GraphQL API without using the `gh` CLI
- [ ] The `make-graphql-query.sh` has been replaced with a conceptually similar python function or method
- [ ] The `gh` CLI tool and `jq` are removed from the `analytics/Dockerfile`

widal001:
Bumping this to sprint 1.4 since I won't have a chance to work on this until I'm back in office and it's not blocking anything.

PR #2606, opened 2024-10-28
[Issue #2582] Ignore CVEs in upstream golang CLI

coilysiren:
## Summary

Relates to #2582

## Changes proposed

Ignores the CVE originally found in https://github.com/HHS/simpler-grants-gov/pull/2572


analytics/Makefile
```diff
@@ -200,3 +200,4 @@ percent-complete:
 sprint-reports: sprint-burndown percent-complete
 
 sprint-reports-with-latest-data: gh-data-export sprint-reports
+
```
coilysiren, line 203: no-op diff on the analytics package to test

widal001 (approved):
LGTM! Thanks for silencing these alerts that will be addressed once we complete #2590 and move away from using the gh CLI

PR #2611, opened 2024-10-28
[Issue #2603] Setup triggers on our opportunity tables which populate the search queue table

### LINKED ISSUE #2603: Setup triggers on our opportunity tables which populate the search queue table
### Summary

Follow-up to https://github.com/HHS/simpler-grants-gov/issues/2602

We want to create Postgres table triggers on several tables that write to the table created whenever an opportunity receives an update. The tables we want this to trigger for are:
* opportunity
* opportunity_assistance_listing
* current_opportunity_summary
* opportunity_summary
* link_opportunity_summary_funding_instrument
* link_opportunity_summary_funding_category
* link_opportunity_summary_applicant_type
* opportunity_attachment

Except for the 3 link tables, all of these tables have the opportunity ID in them (the link tables will need to join to their `opportunity_summary` record for the opportunity_id).

This trigger should upsert a record to the opportunity queue table for all inserts/updates to these tables, setting `has_update=True`

Some useful links:
* Postgres triggers: https://www.postgresql.org/docs/current/sql-createtrigger.html
* Setting up a trigger in SQLAlchemy: https://github.com/sqlalchemy/sqlalchemy/discussions/6806

Note that it's fine if we have to define the triggers directly in SQL within our migrations. While doing it via the ORM might be nice, these are simple enough that getting a fully ORM compliant approach seems like extra work, however, I would recommend looking into it and seeing the level of effort (having them work with unit tests may or may not be worthwhile). 

### Acceptance criteria

- [ ] Postgres triggers setup
- [ ] (if unit tests don't make sense) a writeup of your local testing for updating opportunities and the behavior encountered - should verify that doing an update across multiple parts of an opportunity in a single transaction doesn't cause any issues

mikehgrantsgov:
## Summary
Fixes #2603

## Changes proposed
Add migration which adds DB triggers to populate `opportunity_search_index_queue` based on updates made to existing tables:
opportunity
opportunity_assistance_listing
current_opportunity_summary
opportunity_summary
link_opportunity_summary_funding_instrument
link_opportunity_summary_funding_category
link_opportunity_summary_applicant_type
opportunity_attachment

## Context for reviewers
See test SQL below. Created migration will add or update entries in the opportunity_search_index_queue table until a subsequent process handles them.

## Additional information
See attached SQL file for running tests based on these changes:

```
-- Start transaction for all tests
BEGIN;

-- Test 1: Basic opportunity insert
INSERT INTO api.opportunity (opportunity_id, opportunity_title, is_draft)
VALUES (99999, 'Test Opportunity', false);

-- Verify queue entry was created
SELECT EXISTS (
    SELECT 1 FROM api.opportunity_search_index_queue 
    WHERE opportunity_id = 99999 AND has_update = true
) as "Test 1: Queue entry created for new opportunity";

-- Test 2: Multiple related inserts in single transaction
INSERT INTO api.opportunity (opportunity_id, opportunity_title, is_draft)
VALUES (99998, 'Test Multi-Update Opportunity', false);

INSERT INTO api.opportunity_summary (
    opportunity_summary_id, 
    opportunity_id, 
    summary_description,
    is_forecast
) VALUES (99998, 99998, 'Test Summary', false);

INSERT INTO api.current_opportunity_summary (
    opportunity_id,
    opportunity_summary_id,
    opportunity_status_id
) VALUES (99998, 99998, 1);

INSERT INTO api.link_opportunity_summary_funding_instrument (
    opportunity_summary_id,
    funding_instrument_id
) VALUES (99998, 1);

INSERT INTO api.opportunity_attachment (
    attachment_id,
    opportunity_id,
    opportunity_attachment_type_id,
    file_location,
    mime_type,
    file_name,
    file_description,
    file_size_bytes
) VALUES (
    99998,
    99998,
    1,
    'test/location',
    'text/plain',
    'test.txt',
    'Test file',
    100
);

-- Verify only one queue entry exists for multiple updates
SELECT 
    (SELECT COUNT(*) FROM api.opportunity_search_index_queue WHERE opportunity_id = 99998) = 1 
    as "Test 2: Single queue entry for multiple updates";

-- Test 3: Update existing record
UPDATE api.opportunity 
SET opportunity_title = 'Updated Title' 
WHERE opportunity_id = 99999;

-- Verify has_update is still true
SELECT has_update 
FROM api.opportunity_search_index_queue 
WHERE opportunity_id = 99999 
as "Test 3: has_update still true after update";

-- Test 4: Link table triggers
INSERT INTO api.opportunity_summary (
    opportunity_summary_id, 
    opportunity_id, 
    summary_description,
    is_forecast
) VALUES (99999, 99999, 'Another Test Summary', false);

INSERT INTO api.link_opportunity_summary_funding_instrument (
    opportunity_summary_id,
    funding_instrument_id
) VALUES (99999, 1);

-- Verify queue entry still exists and has_update is true
SELECT EXISTS (
    SELECT 1 FROM api.opportunity_search_index_queue 
    WHERE opportunity_id = 99999 AND has_update = true
) as "Test 4: Queue entry exists after link table insert";

-- Test 5: Verify timestamps are updating
UPDATE api.opportunity 
SET opportunity_title = 'Another Update' 
WHERE opportunity_id = 99999;

SELECT 
    updated_at > created_at 
FROM api.opportunity_search_index_queue 
WHERE opportunity_id = 99999 
as "Test 5: Updated timestamp is newer than created";

-- Output all test data for manual verification
SELECT 'Final Queue State' as description;
SELECT * FROM api.opportunity_search_index_queue WHERE opportunity_id IN (99999, 99998);

-- Cleanup
ROLLBACK;
```

### DIFF
--- a/api/src/db/migrations/versions/2024_10_28_add_opportunity_table_triggers.py
+++ b/api/src/db/migrations/versions/2024_10_28_add_opportunity_table_triggers.py
@@ -39,9 +39,9 @@ BEGIN
 
     INSERT INTO api.opportunity_search_index_queue (opportunity_id, has_update)
     VALUES (opp_id, TRUE)
-    ON CONFLICT (opportunity_id) 
+    ON CONFLICT (opportunity_id)
     DO UPDATE SET has_update = TRUE, updated_at = CURRENT_TIMESTAMP;
-    
+
     RETURN NEW;
 END;
 $$ LANGUAGE plpgsql;
--- a/api/src/db/migrations/versions/2024_10_28_add_opportunity_table_triggers.py
+++ b/api/src/db/migrations/versions/2024_10_28_add_opportunity_table_triggers.py
@@ -1,0 +1,84 @@
+"""Add opportunity table triggers 
+
+Revision ID: a8ebde13a18a
+Revises: a2e9144cdc6b
+Create Date: 2024-10-28 17:48:02.678523
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "a8ebde13a18a"
+down_revision = "a2e9144cdc6b"
+branch_labels = None
+depends_on = None
+
+create_trigger_function = """
+CREATE OR REPLACE FUNCTION update_opportunity_search_queue()
+RETURNS TRIGGER AS $$
+DECLARE
+    opp_id bigint;
+BEGIN
+    -- Determine the opportunity_id based on the table
+    CASE TG_TABLE_NAME
+        WHEN 'link_opportunity_summary_funding_instrument' THEN
+            opp_id := (SELECT opportunity_id FROM api.opportunity_summary WHERE opportunity_summary_id = NEW.opportunity_summary_id);
+        WHEN 'link_opportunity_summary_funding_category' THEN
+            opp_id := (SELECT opportunity_id FROM api.opportunity_summary WHERE opportunity_summary_id = NEW.opportunity_summary_id);
+        WHEN 'link_opportunity_summary_applicant_type' THEN
+            opp_id := (SELECT opportunity_id FROM api.opportunity_summary WHERE opportunity_summary_id = NEW.opportunity_summary_id);
+        WHEN 'opportunity_summary' THEN
+            opp_id := NEW.opportunity_id;
+        WHEN 'current_opportunity_summary' THEN
+            opp_id := NEW.opportunity_id;
+        ELSE
+            opp_id := NEW.opportunity_id;
+    END CASE;
+
+    INSERT INTO api.opportunity_search_index_queue (opportunity_id, has_update)
+    VALUES (opp_id, TRUE)
+    ON CONFLICT (opportunity_id) 
+    DO UPDATE SET has_update = TRUE, updated_at = CURRENT_TIMESTAMP;
+    
+    RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+"""
+
+# List of tables that need triggers
+tables = [
+    "opportunity",
+    "opportunity_assistance_listing",
+    "current_opportunity_summary",
+    "opportunity_summary",
+    "link_opportunity_summary_funding_instrument",
+    "link_opportunity_summary_funding_category",
+    "link_opportunity_summary_applicant_type",
+    "opportunity_attachment",
+]
+
+
+def upgrade():
+    # Create the trigger function
+    op.execute(create_trigger_function)
+
+    # Create triggers for each table
+    for table in tables:
+        op.execute(
+            f"""
+            CREATE TRIGGER {table}_queue_trigger
+            AFTER INSERT OR UPDATE ON api.{table}
+            FOR EACH ROW EXECUTE FUNCTION update_opportunity_search_queue();
+        """
+        )
+
+
+def downgrade():
+    # Drop triggers
+    for table in tables:
+        op.execute(f"DROP TRIGGER IF EXISTS {table}_queue_trigger ON {table};")
+
+    # Drop the trigger function
+    op.execute("DROP FUNCTION IF EXISTS update_opportunity_search_queue();")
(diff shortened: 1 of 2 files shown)


api/src/db/migrations/versions/2024_10_28_add_opportunity_table_triggers.py
```diff
@@ -0,0 +1,83 @@
...
+        )
+
+
+def downgrade():
+    # Drop triggers
+    for table in tables:
+        op.execute(f"DROP TRIGGER IF EXISTS {table}_queue_trigger ON {table};")
+
+    # Drop the trigger function
+    op.execute("DROP FUNCTION IF EXISTS update_opportunity_search_queue();")
```
chouinar, line 83: Running the downgrade, I got this error:
```
sqlalchemy.exc.InternalError: (psycopg.errors.DependentObjectsStillExist) cannot drop function update_opportunity_search_queue() because other objects depend on it
DETAIL:  trigger opportunity_queue_trigger on table api.opportunity depends on function update_opportunity_search_queue()
```

I think you need:
```suggestion
def downgrade():
    # Drop triggers
    for table in tables:
        op.execute(f"DROP TRIGGER IF EXISTS {table}_queue_trigger ON api.{table};")

    # Drop the trigger function
    op.execute("DROP FUNCTION IF EXISTS update_opportunity_search_queue();")
```

chouinar (commented):
Something I didn't catch on your last PR - but that we need before we can merge this.

We need the Opportunity ORM model to be aware of its record in the queue table otherwise we can't delete it.

You can see this if you run `make console` and then try to do:
```py
x = dbs.query(Opportunity).first()
dbs.delete(x)
dbs.commit()
```
which errors with:
```
sqlalchemy.exc.IntegrityError: (psycopg.errors.ForeignKeyViolation) update or delete on table "opportunity" violates foreign key constraint "opportunity_search_index_queue_opportunity_id_opportunity_fkey" on table "opportunity_search_index_queue"
```
We solve this with other foreign keys with the `cascade="all, delete-orphan"` on the relationship.

Alternatively.. I suppose there isn't anything stopping us from making this whole "has_update" column just exist in the opportunity table itself. Thoughts?

mikehgrantsgov:
I kind of like the idea of a separate table to handle the concern of the updates, which can expand if we need to store other info to store into the search queue later on. For this reason I think going the `cascade="all, delete-orphan"` route makes sense here.

> Something I didn't catch on your last PR - but that we need before we can merge this.
> 
> We need the Opportunity ORM model to be aware of its record in the queue table otherwise we can't delete it.
> 
> You can see this if you run `make console` and then try to do:
> 
> ```python
> x = dbs.query(Opportunity).first()
> dbs.delete(x)
> dbs.commit()
> ```
> 
> which errors with:
> 
> ```
> sqlalchemy.exc.IntegrityError: (psycopg.errors.ForeignKeyViolation) update or delete on table "opportunity" violates foreign key constraint "opportunity_search_index_queue_opportunity_id_opportunity_fkey" on table "opportunity_search_index_queue"
> ```
> 
> We solve this with other foreign keys with the `cascade="all, delete-orphan"` on the relationship.
> 
> Alternatively.. I suppose there isn't anything stopping us from making this whole "has_update" column just exist in the opportunity table itself. Thoughts?

api/src/db/models/opportunity_models.py
```diff
@@ -435,5 +435,7 @@ class OpportunitySearchIndexQueue(ApiSchemaTable, TimestampMixin):
     opportunity_id: Mapped[int] = mapped_column(
         BigInteger, ForeignKey(Opportunity.opportunity_id), primary_key=True, index=True
     )
-    opportunity: Mapped[Opportunity] = relationship(Opportunity)
+    opportunity: Mapped[Opportunity] = relationship(
+        Opportunity, cascade="all, delete-orphan", single_parent=True
+    )
```
chouinar, line 440: You would need this relationship to go the other way. Doing this would delete the opportunity if you deleted the queue record.

mikehgrantsgov: Ah, good call. I thought the `single_parent` property would have done this on the queue. Updated this.

chouinar (approved):
LGTM!

PR #2626, opened 2024-10-29
[Issue 1890] add GA beacon for filter use on search

### LINKED ISSUE #1890: Add GA4 tracking for number of searches made per filter combination
### Summary

Number of searches made per filter combination and the click-through-rate to the opportunity listing.

This may be done already.

### Acceptance criteria

Number of searches made per filter combination and the click-through-rate to the opportunity listing

doug-s-nava:
## Summary
Fixes #1890

## Changes proposed
Added a GA event beacon "search_attempt" to track each load of the search results page and pass along a list of filters in place.

## Context for reviewers

Note that in React strict mode, which is on for us and by default when running a DEV Next server, components basically mount, unmount, then mount again, in order to test out expected behavior around maintaining state between page loads, or something. This means that out of the box on a local server you may see multiple beacons being fired when the page is first loaded, but if you turn off strict mode here, you should see only one:
https://github.com/HHS/simpler-grants-gov/blob/fa3aabddd9d34feaa0c358888e7c5008970e7033/frontend/next.config.js#L19

For more see https://legacy.reactjs.org/docs/strict-mode.html#ensuring-reusable-state

Also these test steps will assume you've got omnibug or something similar installed for testing this PR, otherwise you can look at your network console and parse out requests being made to GA that way. https://chromewebstore.google.com/detail/omnibug/bknpehncffejahipecakbfkomebjmokl

### Test steps
1. visit a search page
2. _VERIFY_: omnibug console shows a beacon going out withe event type set to "search_attempt" and search_filters data set to "{"status":"forecasted,posted"}"
3. update filters
4. _VERIFY_: omnibug console shows a search_attempt event with search_filters reflecting the new filter selection
5. enter a search term
6. _VERIFY_: omnibug console shows a search_attempt beacon but event data does not include the search term (or any other non-filter related information)

## Additional information
We're currently firing a "search_term" event whenever a filter or search keyword is updated, however, this event is not fired on the initial search / page load, and only contains the information about the filter or keyword that is actively being updated. I'm not sure if we want to keep this around, or if we want to fold it into the new event by adding some dimensions there. See https://github.com/HHS/simpler-grants-gov/blob/fa3aabddd9d34feaa0c358888e7c5008970e7033/frontend/src/hooks/useSearchParamUpdater.ts#L44

Note that as far as a GA report, that's inconclusive as of 10/29, will need to check back to see if the data is coming in as expected and that we can actually report back on it. https://analytics.google.com/analytics/web/#/analysis/p439528520/edit/bFNLNRQKRN2eUYMJAZshfg

### Omnibug example
Screenshot 2024-10-29 at 11 46 17 AM

### DIFF
--- a/frontend/src/app/[locale]/search/page.tsx
+++ b/frontend/src/app/[locale]/search/page.tsx
@@ -1,5 +1,7 @@
 import { Metadata } from "next";
+import QueryProvider from "src/app/[locale]/search/QueryProvider";
 import withFeatureFlag from "src/hoc/search/withFeatureFlag";
+import { SearchParamsTypes } from "src/types/search/searchRequestTypes";
 import { Breakpoints } from "src/types/uiTypes";
 import { convertSearchParamsToProperTypes } from "src/utils/search/convertSearchParamsToProperTypes";
 
@@ -7,10 +9,10 @@ import { useTranslations } from "next-intl";
 import { getTranslations, unstable_setRequestLocale } from "next-intl/server";
 
 import ContentDisplayToggle from "src/components/ContentDisplayToggle";
+import SearchAnalytics from "src/components/search/SearchAnalytics";
 import SearchBar from "src/components/search/SearchBar";
 import SearchFilters from "src/components/search/SearchFilters";
 import SearchResults from "src/components/search/SearchResults";
-import QueryProvider from "./QueryProvider";
 
 export async function generateMetadata() {
   const t = await getTranslations({ locale: "en" });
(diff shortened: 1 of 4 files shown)


mdragon (approved):
Looks good and we can validate more fully once it's in Dev

acouch (approved):
Would be great to get some links or a follow up ticket to provide the report you were working on to see the results of the "search_attempt"s.

doug-s-nava:
@acouch see follow up ticket here. Also, I think I need a re-review after resolving a merge conflict

mdragon (approved):

PR #2641, opened 2024-10-29
[Issue #2528] Change job schedule ELT process to hourly

### LINKED ISSUE #2528: Adjust transform job to run hourly
### Summary

We want to adjust the rate at which the transform job runs from daily to hourly: https://github.com/HHS/simpler-grants-gov/blob/main/infra/api/app-config/env-config/scheduled_jobs.tf#L50

I'd recommend instead of specifying a rate, we instead specify a cron (the rate approach just arbitrarily chooses when the day/hour begins based on when you enable it). https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-scheduled-rule-pattern.html

### Acceptance criteria

- [ ] Config updated

chouinar:
## Summary
Fixes #2528

## Changes proposed
Modified ELT process to run hourly

Removed the load job for the v0 data

## Context for reviewers
I can't imagine there is much value in the v0 job now that v1 works. Probably will save us a bit as we won't be running a process near constantly that no one uses.

I changed the schedule expression to a cron so that the schedule is a bit more consistent. The `rate(..)` configuration seems to be based on when you turn it on, or last did a deploy. This will keep it from varying every time we deploy and instead make it so it's always at the top of the hour.

## Additional information
https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-scheduled-rule-pattern.html

### DIFF
--- a/infra/api/app-config/env-config/scheduled_jobs.tf
+++ b/infra/api/app-config/env-config/scheduled_jobs.tf
@@ -40,14 +40,10 @@ locals {
     ],
   }
   scheduled_jobs = {
-    copy-oracle-data = {
-      task_command        = ["poetry", "run", "flask", "data-migration", "copy-oracle-data"]
-      schedule_expression = "rate(2 minutes)"
-      state               = "ENABLED"
-    }
     load-transform = {
       task_command        = local.load-transform-args[var.environment]
-      schedule_expression = "rate(1 days)"
+      # Every hour at the top of the hour
+      schedule_expression = "cron(0 * * * ? *)"
       state               = "ENABLED"
     }
   }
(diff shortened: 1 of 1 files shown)


coilysiren (commented):
Approved in concept, but we'll want to solve the terraform lint error 🙏🏽

chouinar:
> Approved in concept, but we'll want to solve the terraform lint error 🙏🏽

I don't know why I always forget terraform formatting. I run the format command like 3x for every API change I make

mdragon (approved):
LGTM

coilysiren (approved):
Q for the sake of my own knowledge: the load transform job also copies Oracle data?

chouinar:
> Q for the sake of my own knowledge: the load transform job also copies Oracle data?

Load transform is the full ELT process made up of 3 "tasks"
* Load data from Oracle via the FDW into our staging tables
* Run the transformation logic between the staging and actual opportunity tables
* Set current opportunities

The job I deleted is the old v0 copy job that loads into the "transfer" tables that we never expanded past just the singular opportunity table.

PR #2686, opened 2024-10-31
[Issue #2604] Incrementally pull updates from the search queue table

### LINKED ISSUE #2604: Modify search index load script to have the incremental mode pull updates from the search queue table
### Summary

After https://github.com/HHS/simpler-grants-gov/issues/2602 and https://github.com/HHS/simpler-grants-gov/issues/2603

We partially setup an incremental load for the populate search opportunities script. We want to continue that implementation and have it determine updates based on opportunities in the queue table we setup.

The incremental process should fully handle the three cases an opportunity is in:
* Insert (opportunity ID in our DB, not in OpenSearch)
* Delete (opportunity ID no longer in our DB, in OpenSearch)
* Updates (described below)

Updates need to consider a few things.
* Updates may include inserts - by the nature of how the triggered records will work, if we just inserted an opportunity, we don't need to re-update it.
* The queue table may contain opportunities we do not want to index (the triggers are naive and may contain opps we don't want to index). For example, if an opportunity is a draft, we don't want to index it. We have these rules already in the query to fetch opportunities, we should make sure to keep that behavior as part of the fetching behavior.
* The actual "updates" to OpenSearch will just be reindexing the same record and will be a full replace. There are some ways to do a partial update, but it's not worthwhile and some work we'll do later for indexing the files would hit issues anyways.

### Acceptance criteria

- [ ] Incremental load implemented as described above
- [ ] Thorough tests written to show behavior of many of these edge cases

mikehgrantsgov:
## Summary
Fixes #2604

## Changes proposed
Change incremental index process to load from queue table.
Handle edge cases for updates - updates will re-index the whole opportunity

## Context for reviewers
WIP for now as we add testing scenarios. Initial draft to review implementation.

## Additional information
See unit tests.


api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -68,21 +72,61 @@ def run_task(self) -> None:
...
+            logger.info(f"Deleting {len(opportunity_ids_to_delete)} opportunities from search")
             self.search_client.bulk_delete(self.index_name, opportunity_ids_to_delete)
 
+        # Clear processed entries from the queue
+        if processed_opportunity_ids:
+            self.db_session.execute(
+                delete(OpportunitySearchIndexQueue).where(
+                    OpportunitySearchIndexQueue.opportunity_id.in_(processed_opportunity_ids)
+                )
+            )
```
chouinar, line 127: WIth the queue table, we had the `has_update` column we could flip rather than deleting the records.

Virtually the same behavior, updating just gives us a some auditability (which the logs could also give, so not a requirement). Happy to go with this approach, but we might as well remove that column to avoid confusion in that case.

api/tests/src/search/backend/test_load_opportunities_to_index.py
```diff
@@ -1,12 +1,52 @@
...
+    client = Mock()
+    # Mock the alias exists check
+    client.alias_exists.return_value = True
+    # Mock empty initial index
+    client.scroll.return_value = [Mock(records=[])]
+    return client
+
+
+@pytest.fixture
+def test_opportunity():
```
chouinar, line 32: Why not use the Opportunity factory for this?

api/tests/src/search/backend/test_load_opportunities_to_index.py
```diff
@@ -151,3 +200,41 @@ def test_load_opportunities_to_index_index_does_not_exist(self, db_session, sear
...
+
+    def test_new_opportunity_gets_indexed(
+        self, db_session, test_opportunity, load_opportunities_to_index
+    ):
+        """Test that a new opportunity in the queue gets indexed"""
+        # Add to queue
+        OpportunitySearchIndexQueueFactory.create(
+            opportunity_id=test_opportunity.opportunity_id, has_update=True
+        )
+        db_session.commit()
```
chouinar, line 212: You can be slightly less verbose with just:
```suggestion
        OpportunitySearchIndexQueueFactory.create(opportunity=test_opportunity)
```
has_update has a default, and the opportunity/opportunity_id all get set together anyways (both at the factory and ORM level).

The factory `create` function also commits after it runs

api/tests/src/search/backend/test_load_opportunities_to_index.py
mikehgrantsgov: Good call, this was simpler to use the factory. At first I was thinking of a shared opportunity for many tests, but the factory is easier.

api/src/search/backend/load_opportunities_to_index.py
mikehgrantsgov: I was thinking delete to avoid an endlessly growing table which might cause unknown side effects down the road. Will remove the column!

api/src/db/migrations/versions/2024_10_28_add_opportunity_table_triggers.py
```diff
@@ -36,10 +37,10 @@
             opp_id := NEW.opportunity_id;
     END CASE;
 
-    INSERT INTO api.opportunity_search_index_queue (opportunity_id, has_update)
-    VALUES (opp_id, TRUE)
+    INSERT INTO api.opportunity_search_index_queue (opportunity_id)
+    VALUES (opp_id)
```
chouinar, line 41: You can't modify an existing migration. Once it runs in an environment, it won't run again unless we manually do migrate-down and that's more hassle than is worthwhile.

You'll need to make a new migration. Since the command was a `create or replace` already, it'll basically just be a copy of this

api/src/search/backend/load_opportunities_to_index.py
mdragon: Just flagging when troubleshooting search data not updating it can be very helpful to see a history of when we thought the data changesd and when those changes purportedly updated to search. So having like data_updated and search_updated timestamps could be really helpful as it's way easier to consume that data directly than parse it out of logs both for like ongoing metrics (how many search updates and lag of data to search change) and troubleshooting. It would create the need to clean up the table but with the platform framework for scheduling jobs having a daily or weekly job that turns anything over 30, 60, 90 days or 6 months or whatever should be pretty doable? We don't have to change this round necessarily but should think through it this is valuable to do in a follow up.

mdragon (dismissed):

api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -68,21 +72,61 @@ def run_task(self) -> None:
...
+        processed_opportunity_ids = set()
+        opportunities_to_index = []
 
-        if len(opportunity_ids_to_delete) > 0:
+        for opportunity in queued_opportunities:
+            logger.info(
+                "Processing queued opportunity",
+                extra={
+                    "opportunity_id": opportunity.opportunity_id,
+                    "in_search": opportunity.opportunity_id in existing_opportunity_ids,
```
chouinar, line 100: What about something like this to get update/insert
```suggestion
                    "status": "update" if opportunity.opportunity_id in existing_opportunity_ids else "insert",
```

api/src/search/backend/load_opportunities_to_index.py
chouinar: One thing that would help and go along with what we have here. Add something like:
```
for opportunity_id in opportunity_ids_to_delete:
     logger.info("Deleting opportunity from search", extra={"opportunity_id": opportunity_id, "status": "delete")
```
It adds a lot of logs (but we already do that for inserts/updates, so whatever), but also makes finding what happened all exist in one place.

api/tests/src/search/backend/test_load_opportunities_to_index.py
```diff
@@ -1,12 +1,25 @@
...
     LoadOpportunitiesToIndexConfig,
 )
 from src.util.datetime_util import get_now_us_eastern_datetime
 from tests.conftest import BaseTestClass
-from tests.src.db.models.factories import OpportunityFactory
+from tests.src.db.models.factories import OpportunityFactory, OpportunitySearchIndexQueueFactory
+
+
+@pytest.fixture
+def mock_search_client():
```
chouinar, line 16: I don't think you use this anywhere

mikehgrantsgov: Good catch - was part of a test I wrote trying out a mock service

api/src/search/backend/load_opportunities_to_index.py
mikehgrantsgov: I added this and makes sense since we can log and report on those logs later to track, and without storing these in an ever-growing table which would need an extra task to prune.

chouinar (approved):
LGTM!

PR #2689, opened 2024-10-31
[Issue #2630] Enable search load to run hourly

### LINKED ISSUE #2630: Configure job to load data to search hourly
### Summary

TODO

(prod only) We need to make sure the shard count is a multiple of 3, and the replica count is that minus 1. 3 & 2 likely work for now.

### Acceptance criteria

- [ ] Job scheduled to run in all environments

chouinar:
## Summary
Fixes #2630

## Changes proposed
Schedule the DB -> search index load job to run every hour

Set the shard/replica count for prod as it needs to be higher as we have more nodes.

## Context for reviewers
In the future we may want to coordinate the ELT process with this (put them in the same flow?), but for now I've setup the job to run 30 minutes after the start of the hour. The job in prod takes ~9 minutes at the moment.

The ELT process takes 4-5 minutes and runs at the top of the hour - the jobs overlapping wouldn't be a big deal other than search having staler data.

### DIFF
--- a/infra/api/app-config/env-config/scheduled_jobs.tf
+++ b/infra/api/app-config/env-config/scheduled_jobs.tf
@@ -46,5 +46,11 @@ locals {
       schedule_expression = "cron(0 * * * ? *)"
       state               = "ENABLED"
     }
+    populate-search-index = {
+      task_command = ["poetry", "run", "flask", "load-search-data", "load-opportunity-data"]
+      # Every hour at the half hour
+      schedule_expression = "cron(30 * * * ? *)"
+      state               = "ENABLED"
+    }
   }
 }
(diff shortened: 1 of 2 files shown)


coilysiren:
> put them in the same flow?

Sounds like a fun configuration task!

infra/api/app-config/env-config/scheduled_jobs.tf
```diff
@@ -46,5 +46,11 @@ locals {
       schedule_expression = "cron(0 * * * ? *)"
       state               = "ENABLED"
     }
+    populate-search-index = {
```
coilysiren, line 49: Love the new easier scheduled jobs configs

coilysiren (approved):

chouinar:
Gonna hold on merging this until after the demo because we want to show some stuff in dev - in the unlikely event this breaks something

coilysiren (approved):

PR #2698, opened 2024-11-01
[Issue #2697] Enable agency data load in ELT process

### LINKED ISSUE #2697: Enable loading of TGROUPS/agency data in the ELT process
### Summary

We've tested these loads, and should enable them to actually run as part of the ELT process.

This requires two changes:
* For the extract load step, adding the `TGROUPS` table to the list of tables copied over
* For the transform step, setting the flag that enables the TransformAgency load

### Acceptance criteria

- [ ] Processes enabled
- [ ] Tests updated

chouinar:
## Summary
Fixes #2697

## Changes proposed
Enable the extract-load process on the TGROUPS table (agency data)

Enable the Agency transform steps

## Context for reviewers
I've tested this in the lower environments and prod by manually configuring the job to load the data, so all of the agency data is there already. This just gets updates from here on out.

The extract-load process has a set of tables it is configured to load, and the transformation process has specific classes for transforming tables that are handled via env vars, so the configuration change is minor.

The unit test changes I made show that the agency transforms are happening now when running the "full" job.

### DIFF
--- a/api/src/data_migration/load/load_oracle_data_task.py
+++ b/api/src/data_migration/load/load_oracle_data_task.py
@@ -37,7 +37,7 @@ TABLES_TO_LOAD = [
     "tfundinstr_forecast_hist",
     "tfundinstr_synopsis",
     "tfundinstr_synopsis_hist",
-    # tgroups,  # Want to hold on this until we have permissions
+    "tgroups",
 ]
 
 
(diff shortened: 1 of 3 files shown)


mdragon (approved):
Looks good, but wouldn't hurt to have Bruk or Mike review as well.

mikehgrantsgov (approved):
Looks good!

PR #2704, opened 2024-11-01
[Issue #2510] Update search logic for /search path

### LINKED ISSUE #2510: Going directly to search page de-selects default opportunity status filters
### Summary

We will probably often send people directly to `https://simpler.grants.gov/search`. When visiting that URL, however, it means that no opportunity statuses are selected for the filter. Is it a desired behavior for the default experience when landing on `https://simpler.grants.gov/search` to have it select "Posted" and "Forecasted" as default filters turned on?

### Reproducibility

### How to reproduce the bug

1. Visit https://simpler.grants.gov/search (not via the Nav)
2. Notice that no opportunity status filters are selected. 
3. Should "Forecasted" and "Posted" be selected by default?

Does this need selected=none?

### Browser

Chrome

### Code of Conduct

acouch:
## Summary
Fixes #2510

## Changes proposed
Change the logic to the following:

### 1. user visits `/search`

forecasted,posted selected while no changes are made the the path or query params:
Image

### 2. user deselects all statuses

statuses stay un-selected

`status=none` for no status selected in the URL

Image

### 3. user selects one status when none are selected

`status=none` is replaced by the selected status:

Image


andycochran (dismissed):

frontend/src/components/search/SearchOpportunityStatus.tsx
```diff
@@ -26,6 +27,10 @@ export default function SearchOpportunityStatus({
   const handleCheck = (value: string, isChecked: boolean) => {
     const updated = new Set(query);
     isChecked ? updated.add(value) : updated.delete(value);
+    // Add "status=NO_STATUS_VALUE" if no values are selected.
```
doug-s-nava, line 30: [nit] not really related to this change, but since this is a client side component, we probably want to wrap this function in a useCallback to limit re-renders

frontend/src/constants/search.ts
```diff
@@ -0,0 +1,2 @@
+// Show all values status for search.
+export const SEARCH_NO_STATUS_VALUE = "none";
```
doug-s-nava, line 2: 👍

frontend/src/utils/search/convertSearchParamsToProperTypes.ts
```diff
@@ -30,7 +31,18 @@ export function convertSearchParamsToProperTypes(
 }
 
 // Helper function to convert query parameters to set
-function paramToSet(param: string | string[] | undefined): Set<string> {
+function paramToSet(
+  param: string | string[] | undefined,
+  type?: string,
+): Set<string> {
+  if (type === "status") {
```
doug-s-nava, line 38: 2 (maybe 2.5) nits:
* if you put the early exit before this logic that's currently at line 46 before these conditionals I think you can save the null check on 42
* I feel slightly itchy about mutating the param argument. What about this instead?
```suggestion
  const defaultedParam = param === SEARCH_NO_STATUS_VALUE) ? "" : "forecasted,posted"
```
Or maybe this is even better since you don't have to spend time splitting the string
```suggestion
  const defaultedParam = param === SEARCH_NO_STATUS_VALUE) ? [] : ["forecasted", "posted"]
```

frontend/src/utils/search/convertSearchParamsToProperTypes.ts
```diff
@@ -30,7 +31,18 @@ export function convertSearchParamsToProperTypes(
...
+    if (param === SEARCH_NO_STATUS_VALUE) {
+      param = "";
+    } else {
+      param = param ?? "forecasted,posted";
```
doug-s-nava, line 42: always wary of being too comment heavy, but since this function says `Helper function to convert query parameters to set` and it's now doing quite a bit more than that, it probably is worth updating or adding a new comment

doug-s-nava (commented):
a couple of nonblocking nits and one request for a comment

frontend/src/utils/search/convertSearchParamsToProperTypes.ts
acouch: Updated, good catch.

frontend/src/utils/search/convertSearchParamsToProperTypes.ts
acouch: Updated, thanks.

frontend/src/utils/search/convertSearchParamsToProperTypes.ts
```diff
@@ -30,12 +32,18 @@ export function convertSearchParamsToProperTypes(
 }
 
 // Helper function to convert query parameters to set
-function paramToSet(param: string | string[] | undefined): Set<string> {
-  if (!param) return new Set();
-  if (Array.isArray(param)) {
-    return new Set(param);
+// and to reset that status params none if status=none is set
+function paramToSet(param: QuerySetParam, type?: string): Set<string> {
+  const defaultedParam =
```
doug-s-nava, line 37: ugh, this logic is more complex than I thought - apologies for the problems in my first suggestion. I think this works though???

```suggestion
function paramToSet(param: QuerySetParam, type?: string): Set<string> {\
  if (!param && type === "status") {
    return new Set(["forecasted", "posted"])
  }

  if (!param || (type === "status" && param === SEARCH_NO_STATUS_VALUE)) {
    return new Set();
  }

  if (Array.isArray(param)) {
    return new Set(param);
  }
  return new Set(param.split(","));
}
```

acouch: What I created can only be described as the twisted love-child of bad design decisions in typescript and javascript and my own addled mind the night before an important election; and the worst monstrosity I've almost ever wrought on a code base.

This looks better, thanks!

doug-s-nava (approved):
nice

✅ ✅ ✅ ✅ ✅

PR #2711, opened 2024-11-04
[Issue #2699] Skip assistance listing transform processing with empty fields

### LINKED ISSUE #2699: Modify assistance listing transform code to not load assistance listings without a program_title/assistance_listing_number
### Summary

See: https://github.com/HHS/simpler-grants-gov/issues/2325 for additional context

In brief, we want to modify the process that transforms assistance listings from the legacy system to not create a record in our assistance listings table if the program_title/assistance_listing_number are null/empty. The legacy system has these "empty" records to handle connecting other pieces of data in a way that is confusing, and can be implemented differently when we get to it.

If we see that these values aren't set (null/empty string) we should:
* Not create a record on our side
* Set the transformation_notes to `empty_assistance_listing` + set the transformed_at timestamp
* Log a note

The existing transform process already skips assistance listings in certain situations (like orphaned records), so this should just be adding to that list of `if/elif..` statements.

NOTE: Don't worry about the case where the assistance listing is an update, the idea is we won't import these at all if the value is empty. There will be values that are already empty, we'll have to go back and delete those after we implement this.

### Acceptance criteria

- [ ] Changes made to support this skip behavior
- [ ] Tests added

mikehgrantsgov:
## Summary
Fixes #2699

## Changes proposed
Modify `transform_assistance_listings` to skip listings with empty or null values for `programtitle` or `cfdanumber` values.

## Context for reviewers
We want to modify the process that transforms assistance listings from the legacy system to not create a record in our assistance listings table if the program_title/assistance_listing_number are null/empty. The legacy system has these "empty" records to handle connecting other pieces of data in a way that is confusing, and can be implemented differently when we get to it.

## Additional information
See attached unit tests

### DIFF
--- a/api/src/data_migration/transformation/subtask/transform_assistance_listing.py
+++ b/api/src/data_migration/transformation/subtask/transform_assistance_listing.py
@@ -87,6 +87,27 @@ class TransformAssistanceListing(AbstractTransformSubTask):
             source_assistance_listing.transformation_notes = transform_constants.ORPHANED_CFDA
 
         else:
+            # Check if assistance listing has empty required fields
+            if (
+                not source_assistance_listing.programtitle
+                or not source_assistance_listing.cfdanumber
+            ):
+                self.increment(
+                    transform_constants.Metrics.TOTAL_RECORDS_SKIPPED,
+                    prefix=transform_constants.ASSISTANCE_LISTING,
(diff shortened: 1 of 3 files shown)


chouinar (approved):
LGTM!

Issue #2636, opened 2024-10-29
Design for document and NOFO edge cases


doug-s-nava:
### Summary

In order to build a full UI for documents we should make decisions and draw up designs and or UI requirements for:

* case where an opportunity has no associated documents
* case where an opportunity has no documents that would fall into either the "Related documents" or "Application forms" sections
* case where an opportunity has multiple "NOFO" documents and a user clicks the "Download Full Grant Details" button

### Acceptance criteria

designs and or UI requirements are created for the cases mentioned above

crystabelrangel:
@doug-s-nava @emilycnava here is the [Figma in Dev Mode](https://www.figma.com/design/FGxWtAgToKhehLJCiuy1zL/Simpler.Grants.gov-Project?node-id=607-14094&m=dev)

For the edge case of multiple NOFO files, we discussed this with @chouinar and @emilycnava, since these seven NOFO opportunities will only have one NOFO file we will not have multiple files. We can look into only downloading the latest one or downloading a zip file down the line.

PR #2722, opened 2024-11-04
Add documents section to opportunity listing page

### LINKED ISSUE #2637: Create "documents" section on opportunity detail
### Summary

We need a place on the opportunity detail page to display associated documents

See designs here: https://www.figma.com/design/FGxWtAgToKhehLJCiuy1zL/Simpler.Grants.gov-Project?node-id=239-2733&node-type=frame&m=dev

Related tickets:

* https://github.com/HHS/simpler-grants-gov/issues/2275 depends on this change
* https://github.com/HHS/simpler-grants-gov/issues/2636 is a pre-requisite

### Acceptance criteria

A documents section is created that adheres to the designs from Figma and outcome of https://github.com/HHS/simpler-grants-gov/issues/2636

emilycnava:
## Summary
Fixes #2637 

## Changes proposed
Add documents section to opportunity listing page.

## Additional information
Screenshot 2024-11-13 at 3 59 53 PM


frontend/src/components/opportunity/OpportunityDocuments.tsx
```diff
@@ -0,0 +1,76 @@
...
+  documents: OpportunityDocument[];
+}
+
+const OpportunityDocuments = ({ documents }: OpportunityDocumentsProps) => {
+  const t = useTranslations("OpportunityListing.documents");
+
+  const documentTable = (documents_filtered: OpportunityDocument[]) => {
+    return (
+      <>
+        <thead>
```
doug-s-nava, line 21: Is there a reason why we shouldn't use the Trussworks table component here? https://trussworks.github.io/react-uswds/?path=/docs/components-table--docs

frontend/src/components/opportunity/OpportunityDocuments.tsx
```diff
@@ -0,0 +1,76 @@
...
+}
+
+interface OpportunityDocumentsProps {
+  documents: OpportunityDocument[];
+}
+
+const OpportunityDocuments = ({ documents }: OpportunityDocumentsProps) => {
+  const t = useTranslations("OpportunityListing.documents");
+
+  const documentTable = (documents_filtered: OpportunityDocument[]) => {
```
doug-s-nava, line 18: seems to me like it'd be beneficial to split this into its own component. Potentially something that could be abstracted and reused somewhere else later, and making it its own thing not reliant on the wrapping component would be a good first step.

frontend/src/components/opportunity/OpportunityDocuments.tsx
```diff
@@ -0,0 +1,76 @@
...
+              </td>
+            </tr>
+          ))}
+        </tbody>
+      </>
+    );
+  };
+
+  const related_docs = documents.filter(
+    (d) => d.opportunity_attachment_type === "notice_of_funding_opportunity",
```
doug-s-nava, line 55: maybe a bigger discussion for the team but I'm very much of the opinion that JS variable names should be descriptive. I'd advocate for `document` rather than `d` here.

emilycnava:
> What is the expected behavior if there are no docs?

It puts the '--' as is shown under 'Application Forms' in the screenshot, i.e. that screenshot shows one table with a doc and one without, if neither have docs they'll both just show '--', per design spec.

frontend/src/components/opportunity/OpportunityDocuments.tsx
```diff
@@ -0,0 +1,76 @@
...
+                  ? t("type.funding_details")
+                  : t("type.other")}
+              </th>
+              <td data-label={t("table_col_file_name")}>
+                <Link target="_blank" href={document.download_path}>
+                  {document.file_name}
+                </Link>
+              </td>
+              <td data-label={t("table_col_last_updated")}>
+                {new Date(document.updated_at).toDateString() +
```
doug-s-nava, line 43: What I'm seeing in the UI doesn't match the designs as far as the timestamp format.
`Thu Oct 24 2024 16:05:21 GMT-0400 (Eastern Daylight Time)` vs `Oct 24, 2024 4:05:21 PM EDT`

We have a helper for this that should spit out the correct date string format for display https://github.com/HHS/simpler-grants-gov/blob/main/frontend/src/utils/dateUtil.ts. Using that should help, but may need to expand it in some way

doug-s-nava (commented):
It looks like the "no documents" edge cases don't have completed designs yet, do we want to wait on the ticket for that, or go ahead with this and do a follow up later? https://github.com/HHS/simpler-grants-gov/issues/2636

emilycnava:
> It looks like the "no documents" edge cases don't have completed designs yet, do we want to wait on the ticket for that, or go ahead with this and do a follow up later? #2636

I could be mistaken about intention, but the figma design does in fact have a null values design showing '--' in the case of no documents, I operated under the assumption that was the design for 'no documents'

frontend/src/components/opportunity/OpportunityDocuments.tsx
emilycnava: I did use the trussworks table component, just this particular helper function returns only the table body which is identical to a regular html table, the <Table> component is further down in the code/return statement

frontend/src/components/opportunity/OpportunityDocuments.tsx
emilycnava: d was meant to be short for document, but if we want, I can change it?

frontend/src/components/opportunity/OpportunityDocuments.tsx
emilycnava: Yeah, I considered adding moment.js or something to get a very particular formatting, I guess my judgement call was it might be acceptable to deviate slightly from design r.e. exact formatting so long as all the info was there to avoid an external library and additional dependency, didn't know we had a dateUtil, I can look into that, but not sure if that will yield us the exact format in design either.

frontend/src/components/opportunity/OpportunityDocuments.tsx
emilycnava: You mean splitting up the table into it's own component? I did consider that too but I felt like it's be unnecessary wrapper around the existing trussworks table component since the code is like 99% the same its just the data that changes I figured just a function would serve that purpose for now, but open to a discussion about it. I guess I had the mentality of duplicate until it hurts, then abstract and taking this approach def didn't 'hurt'.

doug-s-nava: Yeah, the future reusability may never really surface, but even without that, the only reason I can see to leave it nested inside the larger component like this is to close over the `t` function defined in the outer scope, which you can get around by calling the translation hook again in the new component. Other than that it is basically being used as its own component already so it seems clearer to me to define it that way to begin with. Happy to get on a call to discuss further

frontend/src/components/opportunity/OpportunityDocuments.tsx
doug-s-nava: again, just one guy's opinion, and a larger conversation to be had, I like Uncle Bob's take on variable naming. 

https://dzone.com/articles/naming-conventions-from-uncle-bobs-clean-code-phil

In this case, though, we can avoid the discussion entirely, and just destructure the one attribute you need, which I think is the best way in any case:

```suggestion
    ({ opportunity_attachment_type }) => opportunity_attachment_type === "notice_of_funding_opportunity",
```

frontend/src/components/opportunity/OpportunityDocuments.tsx
emilycnava: Exploring libraries to help with this

emilycnava: Group coalesced around day.js

doug-s-nava (commented):
one more nit here as we wait for confirmation on designs for the null case

andycochran:
We can replace the whole table with "--" if there are no docs (just as the 'Application Forms' section)

emilycnava:
> We can replace the whole table with "--" if there are no docs (just as the 'Application Forms' section)

Perfect, that's the current behavior, @doug-s-nava does that satisfy you enough to mark this done and get it merged in?

frontend/src/components/opportunity/OpportunityDocuments.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+interface OpportunityDocumentsProps {
+  documents: OpportunityDocument[];
+}
+
+const DocumentTable = ({ documents }: OpportunityDocumentsProps) => {
+  const t = useTranslations("OpportunityListing.documents");
+
+  dayjs.extend(advancedFormat);
+  dayjs.extend(timezone);
```
doug-s-nava, line 23: thought I made a comment here yesterday but I guess it never went through. Oops!

Anyway, can we move these plugin instantiations out of the render, and maybe even into somewhere that they can be shared / only run once? I don't know if initiating these plugins is expensive but it'd be good to limit how often it happens, since they only really need to be initiated once.

doug-s-nava (commented):
If Andy's good on the "no documents" case, then I am too! One last thing on the date stuff, and I'm good to sign off

doug-s-nava (approved):
I can't speak to the design side updates, but this lgtm, so I'll go ahead and sign off.

✅ ✅ ✅ ✅

PR #2756, opened 2024-11-06
[Issue #2732] clear search input when clicking search nav link

### LINKED ISSUE #2732: Clear search bar text with a new search (e.g. clicking "Search" on nav)
### Summary

Any keywords entered in the search bar should clear when you click "Search" on the nav

### Reproducibility

None

### How to reproduce the bug

1. Click on "Search" nav item
2. enter keyword in the search bar (e.g. climate)
3. click Search/hit enter
4. Updated results show
5. to reset/start a new search, click on "Search" nav iteam
6. Search results refresh, but the keyword in the search bar remains. The keyword should also clear.

### Browser

None

### Code of Conduct

doug-s-nava:
## Summary
Fixes #2732

## Changes proposed
Fixes a bug where the search bar does not clear when clicking "search" on the nav while on the search page.

The solution I came up with was to append a query param to the search nav item when on the search page, whose presence is a signal to the page to clear the search bar.

The approach is not ideal, but a fix requires knowing when a navigation has occurred, and Next does not provide a simple way to know that without relying on something like global state.

### Test steps

1. on this branch, visit the search page
2. enter text in the search bar
3. click the "search" nav item
4. _VERIFY_ the search bar is cleared when the page refreshes

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -6,6 +6,7 @@
     "node": ">=20.0.0"
   },
   "scripts": {
+    "all-checks": "npm run test && npm run lint && npm run ts:check && npm run build",
     "build": "next build",
     "dev": "next dev",
     "debug": "NODE_OPTIONS='--inspect' next dev",
--- a/frontend/src/components/search/SearchBar.tsx
+++ b/frontend/src/components/search/SearchBar.tsx
@@ -21,6 +21,8 @@ export default function SearchBar({ query }: SearchBarProps) {
     updateQueryParams("", "query", queryTerm, false);
   };
 
+  // if we have "refresh=true" query param, clear the input
+  // this supports the expected refresh of the input if the user clicks the search link while on the search page
   useEffect(() => {
     if (searchParams.get("refresh") && inputRef.current) {
       updateQueryTerm("");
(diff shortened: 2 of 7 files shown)


frontend/package.json
```diff
@@ -6,6 +6,7 @@
     "node": ">=20.0.0"
   },
   "scripts": {
+    "all-checks": "npm run test && npm run lint && npm run ts:check && npm run build",
```
emilycnava, line 9: love this!! kudos!

frontend/src/components/search/SearchBar.tsx
```diff
@@ -36,6 +52,7 @@ export default function SearchBar({ query }: SearchBarProps) {
       </label>
       <div className="usa-search usa-search--big" role="search">
         <input
+          ref={inputRef}
```
emilycnava, line 55: Non blocking: Is it worth just adding either useState or possibly the newer server action state management for this form as opposed to using refs and essentially handling setting state manually?

emilycnava (commented):
Confirmed it works as expected, left a few thoughts. If this is the approach we wanna take, feel free to merge. I do think this bug may raise bigger questions about state management on the search page since I know there's discussions around how to handle having 'forcasted' and 'posted' checked by default too, so I think we just have some complex state on this particular page and may be worth doing a wholistic eval?

mdragon (approved):
Approving, per Emily's review

doug-s-nava:
Note that @emilycnava and I discussed offline, both agree that this solution doesn't feel great but that we don't have a better idea without creating some global state or other relatively heavy solution that would be out of scope for this ticket. We should review this with whoever is interested and figure out what to do with this sort of thing as a next step.

andycochran:
Could we use `router.reload()` or something to force a page refresh when you click the nav menu?

doug-s-nava:
@andycochran I'll look into that, if we're ok with doing a full page refresh in this case

Issue #2665, opened 2024-10-30
Load GitHub data into Postgres in Dev


widal001:
### Summary

Load the GitHub issue data into Postgres in dev so that we can start analyzing this data in Metabase. 

**Note:** Ideally we'd be loading this in using the new data schema that @DavidDudas-Intuitial has been working on, but at as a fall back, we can use the `GitHubIssue.to_sql()` method to dump the flattened data into Postgres, and we should consider switching to that fallback strategy if we can't get the new data schema working by 11/5.

TODO: 
- [x] Run daily step function to execute `make gh-transform-and-load` 
- [x] Add new step function to execute `make init-db`
- [x] Add IAM support to `/analytics` db client
- [x] Troubleshoot failed db connection attempts from ECS container 
- [x] Troubleshoot create table privilege problems

### Acceptance criteria

- [x] On a daily basis, we load data exported from GitHub into the Postgres DB in dev

widal001:
This would most likely use AWS step functions to run the CLI command for loading the data. @coilysiren can support on how to build/run the step function.

DavidDudas-Intuitial:
https://github.com/HHS/simpler-grants-gov/pull/2759

widal001:
Beep boop: Automatically closing this issue because it was marked as 'Done' in https://github.com/orgs/HHS/projects/13. This action was performed by a bot.

sarahknoppA6:
@widal001 can you tell us why this keeps getting moved to done?

DavidDudas-Intuitial:
PR https://github.com/HHS/simpler-grants-gov/pull/2778

DavidDudas-Intuitial:
FYI - This task has many PRs, and every time I merge one of them, a bot closes the ticket! That's why I keep reopening the ticket. I will close it manually when the task is actually done.

DavidDudas-Intuitial:
Related PR: https://github.com/HHS/simpler-grants-gov/pull/2779

DavidDudas-Intuitial:
PR https://github.com/HHS/simpler-grants-gov/pull/2786

DavidDudas-Intuitial:
Bumped points from 2 to 3

DavidDudas-Intuitial:
Now seeing refused connections in ECS logs:
```2024-11-08T03:26:30.934Z | │ │ exc_tb = <traceback object at 0x7f6383899b40> │ │
-- | --
[...]

PR #2759, opened 2024-11-06
[Issue 2665] Add gh-transform-and-load command to scheduled jobs

DavidDudas-Intuitial:
## Summary
Fixes #2665 

### Time to review: __1 min__

Added `gh-transform-and-load` command to existing `make gh-data-export` command. I'm not sure if this is sufficient or correct, but I'm taking a guess based on what I see in https://github.com/HHS/simpler-grants-gov/pull/2546 and https://github.com/HHS/simpler-grants-gov/pull/2506.

In the analytics work stream, we have a new CLI command `make gh-transform-and-load` for transforming and loading (some) GitHub data. Per issue #2665, that command should be run daily, after the existing `gh-data-export` command which exports data from Github. 

I see that `scheduled_jobs.tf` seems to be the mechanism by which `make gh-data-export` runs daily. In this PR I'm taking and educated guess and attempting to add `gh-transform-and-load` to the existing job, and requesting feedback from @coilysiren as to whether this is the correct approach.

### DIFF
--- a/infra/analytics/app-config/env-config/scheduled_jobs.tf
+++ b/infra/analytics/app-config/env-config/scheduled_jobs.tf
@@ -7,7 +7,7 @@ locals {
 
   scheduled_jobs = {
     sprint-reports = {
-      task_command        = ["make", "gh-data-export", "sprint-reports"]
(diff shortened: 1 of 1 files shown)


coilysiren (approved):
👍

Issue #2716, opened 2024-11-04
Create user tables for basic login.gov auth


chouinar:
### Summary

We want to create three tables in our database:

`user`
* user_id : UUID
* created_at/updated_at timestamps (like every other table, just use the mixin)

`lk_external_user_type`
This follows our usual pattern for [lookup tables](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/db/models/lookup_models.py)

The enum we create for this will only have a single value of `login_gov` at this time.

**link_external_user**
* link_external_user_id : UUID
* external_user_id: str
* user_id: UUID / foreign key to user table / indexed
* external_user_type_id: int / foreign key to the lk_external_user_type table
* email: str
* first_name: str
* last_name: str

Note that we have been using integers for a lot of our tables primary keys up to this point. However we want these to be UUIDs so they aren't just incrementally counting upward. If you make the python type `Mapped[uuid.UUID]` it should make the DB type correctly, but we'll need to validate that (it should also automatically populate the ID if none provided, we'll likely need to configure that).

### Acceptance criteria

- [x] Tables created
- [x] The two user related tables should be in a new models file
- [x] Migrations created
- [ ] Updated factories to be able to generate user with login.gov info

babebe:
@chouinar  Do I have this right >>  The UUID for table `user` is what we generate ? which is the foreign key  `external_user_type_id` in table `link_external_user` and the UUID in that table is the UUID from login.gov ?

chouinar:
@babebe 

The user table has no knowledge of anything to do with login.gov - it's a very generic table we'll add to later.

The link_external_user table is where we put info that will connect a user to an oauth service, in this case, just login.gov

The lk_external_user_type is just to store enums for different types of external users (in this case, just login.gov) in our DB

PR #2760, opened 2024-11-07
[ISSUE 2716] Create user tables for basic login.gov

babebe:
## Summary
Fixes #{[2716](https://github.com/HHS/simpler-grants-gov/issues/2716)}

## Changes proposed
3 user tables
migration script
updated factories to create new users

## Context for reviewers
Users tables will be used for Oauth2


api/src/db/models/user_models.py
chouinar, line 1: You'll need to generate migrations with `make db-migrate-create MIGRATE_MSG="some sort of small message here"`

babebe: Had not pushed the script

api/src/db/models/user_models.py
```diff
@@ -0,0 +1,35 @@
...
+    link_external_user_id: Mapped[uuid.UUID] = mapped_column(UUID, primary_key=True)
+
+    external_user_id: Mapped[str]
+
+    user_id: Mapped[uuid.UUID] = mapped_column(UUID, ForeignKey(User.user_id), index=True)
+    user: Mapped[User] = relationship(User)
+
+    external_user_type_id: Mapped[int] = mapped_column(
+        BigInteger, ForeignKey(LkExternalUserType.external_user_type_id)
+    )
```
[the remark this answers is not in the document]
babebe: Got it ! Thank you

api/src/db/models/user_models.py
```diff
@@ -0,0 +1,39 @@
...
+
+    user_id: Mapped[uuid.UUID] = mapped_column(UUID, primary_key=True)
+
+
+class LinkExternalUser(ApiSchemaTable, TimestampMixin):
+    __tablename__ = "link_external_user"
+
+    link_external_user_id: Mapped[uuid.UUID] = mapped_column(UUID, primary_key=True)
+
+    external_user_id: Mapped[str]
```
chouinar, line 23: We should probably index/unique this since it'll be what we actually query the table by most of the time.

```suggestion
    external_user_id: Mapped[str] = mapped_column(index=True, unique=True)
```

api/tests/src/db/models/factories.py
```diff
@@ -1833,3 +1835,27 @@ class Meta:
...
+
+    user_id = Generators.UuidObj
+
+
+class LinkExternalUserFactory(BaseFactory):
+    class Meta:
+        model = user_models.LinkExternalUser
+
+    link_external_user_id = Generators.UuidObj
+    external_user_id = "test111"
```
chouinar, line 1852: You'll want to make this be a unique-ish value.

Since login.gov will be the only one and they use UUIDs, we should be able to reuse that here.

api/src/db/models/user_models.py
```diff
@@ -0,0 +1,39 @@
...
+        LookupColumn(LkExternalUserType),
+        ForeignKey(LkExternalUserType.external_user_type_id),
+        index=True,
+    )
+
+    email: Mapped[str]
+
+    first_name: Mapped[str]
+
+    last_name: Mapped[str]
```
chouinar, line 39: NOTE: regardless of anything else, we should hold on merging this until I've gotten more clarification. We might not have first/last name as those require ID proofing and I don't know if we intend for users to be ID proofed 100% of the time.

api/tests/src/db/models/factories.py
babebe: using the `Generators.UuidObj` now

api/src/db/models/user_models.py
babebe: Converted this PR to a Draft PR so we don't accidentally merge

coilysiren (commented):
defering my review

api/tests/src/db/models/factories.py
```diff
@@ -1833,3 +1835,27 @@ class Meta:
...
+    external_user_id = Generators.UuidObj
+
+    user = factory.SubFactory(UserFactory)
+    user_id = factory.LazyAttribute(lambda s: s.user.user_id)
+
+    external_user_type_id = factory.fuzzy.FuzzyChoice(ExternalUserType)
+
+    email = factory.Faker("email")
+    first_name = factory.Faker("first_name")
+    last_name = factory.Faker("last_name")
```
chouinar, line 1861: You don't need first/last name anymore

chouinar (commented):
Just one minor adjustment on the factory - otherwise looks good

PR #2774, opened 2024-11-07
[no ticket] fix labeler

coilysiren:
## Context

Followup to https://github.com/HHS/simpler-grants-gov/pull/2769

The labeler was still broken, unfortunately

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -5,24 +5,21 @@
 #
 # See https://github.com/marketplace/actions/labeler
 #
-
 # Labels for programming languages.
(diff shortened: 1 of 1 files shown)


chouinar (approved):

acouch:
This looks right. Seeing: https://github.com/actions/labeler/issues/712#issuecomment-2454544085 

> the problem is that it was still using the config v4 file from the base branch, I fixed this in the PR by adding a checkout. 

Could try and see if a code checkout fixes it and do a separate PR to remove or just pul in as and hope it is the same problem.

acouch (approved):

PR #2777, opened 2024-11-07
[Issue #2775] Opportunity number partial match fix

### LINKED ISSUE #2775: Search does not successfully partial match on opportunity number
### Summary

My expectation was that if we search for a grant using *part* of the opportunity number, it would match on the result. This does not appear to be working. 

### Reproducibility

### How to reproduce the bug

1. Go to https://simpler.grants.gov/search 
2. Search for "FY25-BSB" (or a part of the opportunity number for any opportunity listing). 
3. Notice no results appear. 
4. Search for "FY25-BSB-410601" (or the full opportunity number for any opportunity listing). 
5. Notice that your result will appear. 

I would've expected the partial match to succeed. 

Image

### Browser

Chrome

### Code of Conduct

chouinar:
## Summary
Fixes #2775 

## Changes proposed
Adjusts the fields we query against and how in search

## Context for reviewers
I wrote up a pretty thorough explanation of keyword vs non-keyword on #2775 - check that out. In short, the bug here is due to how the request/field being compared are tokenized. Keyword leaves a field "whole" and non-keyword tokenizes it into pieces. In the case of opportunity number splitting on each of the dashes (so `ABC-123-XYZ` becomes `["ABC", "123", "XYZ"]`. In short, to support all the ways someone might want to query for an opportunity number, it's easiest to just include both.

## Additional information
Some testing locally where I set the opportunity number to `FY25-BSB-56789` and `FY25-BSB-12345` for two opportunities:

Screenshot 2024-11-07 at 2 27 07 PM
Screenshot 2024-11-07 at 2 27 12 PM
Screenshot 2024-11-07 at 2 27 18 PM
Screenshot 2024-11-07 at 2 27 31 PM


mdragon (approved):
Nice, was hoping it would be that simple.

PR #2787, opened 2024-11-08
[Issue #2770] cache opportunity requests

### LINKED ISSUE #2770: Next is making too many requests
### Summary

@chouinar noticed that
* for each search page load, 4 search requests are being received by the API, and 
* for each opportunity listing page load, 2 opportunity listing requests are being received

The expectation is that only one request is made in each case.

Initial research points to Next's request memoization not working as expected.

### Reproducibility

### How to reproduce the bug

1. visit a search page
2. tail your docker logs and notice that 4 requests are received
3. visit an opportunity page
4. tail your docker logs and notice that 2 requests are received

### Browser

None

### Code of Conduct

doug-s-nava:
## Summary
Fixes #2770

Note that https://github.com/HHS/simpler-grants-gov/pull/2790 also solves this issue, we only need to take one or the other

## Changes proposed

Currently two requests to the API's opportunity endpoint on each opportunity detail page load. This is because the page is making two calls to the opportunity fetch function and for reasons that are still not entirely clear Next's fetch memoization functionality is not deduplicating the calls, and allowing both requests to pass through.

This change wraps call to opportunity fetch function in a React "cache" call to get desired memoization behavior and eliminate the duplicate call.

Also included is a slight change to the "fetcher" definition behavior and pattern, which necessitates adding a `.bind()` call in order to maintain the use of "this" in the underlying API classes.

## Context for reviewers
So far I have not been able to track down why the React / Next memoization is not happening, but in the interest of time this seems like a good solution at least temporarily. Follow up can be found here https://github.com/HHS/simpler-grants-gov/issues/2789

### Test steps

1. visit an opportunity detail on main while tailing API docker logs `docker logs grants-api -f`
2. _VERIFY_: two separate "start request" logs are visible
1. visit an opportunity detail on this branch while tailing API docker logs `docker logs grants-api -f`
2. _VERIFY_: one "start request" log is visible

### DIFF
--- a/frontend/src/app/[locale]/opportunity/[id]/page.tsx
+++ b/frontend/src/app/[locale]/opportunity/[id]/page.tsx
@@ -25,8 +25,7 @@ export async function generateMetadata({ params }: { params: { id: string } }) {
   const id = Number(params.id);
   let title = `${t("OpportunityListing.page_title")}`;
   try {
-    const { data: opportunityData } =
-      await fetchers.opportunityFetcher.getOpportunityById(id);
+    const { data: opportunityData } = await fetchers.opportunityFetcher(id);
     title = `${t("OpportunityListing.page_title")} - ${opportunityData.opportunity_title}`;
   } catch (error) {
     console.error("Failed to render page title due to API error", error);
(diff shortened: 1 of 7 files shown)


doug-s-nava:
closing this in favor of https://github.com/HHS/simpler-grants-gov/pull/2790

PR #2786, opened 2024-11-08
[Issue 2665] Implement IAM for analytics db 

DavidDudas-Intuitial:
## Summary
Fixes #2665 

### Time to review: __2 mins__

According to [slack thread](https://betagrantsgov.slack.com/archives/C05TSL64VUH/p1730739915257369?thread_ts=1730737407.547249&cid=C05TSL64VUH), the `/analytics` step functions can connect to Postgres only via use of IAM. This PR mimics code patterns in `/api` to implement IAM support.

PR #2796, opened 2024-11-08
[Issue 2665] Update analytics db settings to recognize empty pwd setting in env

DavidDudas-Intuitial:
## Summary
Fixes #2665

### Time to review: __1 min__

Made `DBSettings.password` an optional field in `analytics/config.py`, thereby allowing the value to be `None`, which in turn allows IAM token-creation to occur.


coilysiren (approved):
I assume the idea is: you do use the password locally, and don't in AWS, therefore the password must be optional

DavidDudas-Intuitial:
> I assume the idea is: you do use the password locally, and don't in AWS, therefore the password must be optional

That's exactly right. When running in AWS, the ENV should presumably have a blank pwd value. The code I merged yesterday (https://github.com/HHS/simpler-grants-gov/pull/2786) has logic to generate an IAM token if the pwd value is `None`.

PR #2797, opened 2024-11-08
[Issue #2770] deduplicate search requests

### LINKED ISSUE #2770: Next is making too many requests

doug-s-nava:
## Summary
Fixes #2770

## Changes proposed

The task here was to fix the current situation in which the search page is making 4 identical requests to the API on each page load.

To do this, instead of making individual requests in each of the search components that require search result data, we make one request in the top level component, and pass the unresolved Promise down to each child component, where it can resolve and provide data upon resolution.

Note that Next / React's request memoization does not work in this case because our search call is "POST" request and memoization only works on "GET"s.

### Test steps

1. visit an search page on main while tailing API docker logs `docker logs grants-api -f`
2. _VERIFY_: 4 separate "start request" logs are visible
1. visit a search page on this branch while tailing API docker logs `docker logs grants-api -f`
2. _VERIFY_: one "start request" log is visible

### DIFF
--- a/frontend/src/components/search/SearchPaginationFetch.tsx
+++ b/frontend/src/components/search/SearchPaginationFetch.tsx
@@ -1,23 +1,25 @@
 "use server";
 
-import fetchers from "src/app/api/Fetchers";
-import { QueryParamData } from "src/types/search/searchRequestTypes";
+import { SearchAPIResponse } from "src/types/search/searchResponseTypes";
 
 import SearchPagination from "src/components/search/SearchPagination";
 
 interface SearchPaginationProps {
-  searchParams: QueryParamData;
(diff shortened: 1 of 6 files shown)


acouch (approved):
Looks great, thanks.

Issue #2668, opened 2024-10-30
Setup local oauth2 server for API to connect to


chouinar:
### Summary

We want to setup a local oauth2 server to connect to for the purposes of local development. Non-locally we'll be connecting to login.gov, but for local development we want something running on your host.

One possible implementation is https://github.com/navikt/mock-oauth2-server which looks like we can add to our docker-compose file.

### Acceptance criteria

- [ ] Local oauth2 instance stood up
- [ ] Rudimentary testing / notes about how it works (things like how we create users)

babebe:
@chouinar Since we can get the public key associated with the token via the` jwks_uri` is it still necessary for us to have a static key ?

PR #2806, opened 2024-11-12
[Issue 2668] Setup local oauth2 server for API to connect to

babebe:
## Summary
Fixes #{[2668](https://github.com/HHS/simpler-grants-gov/issues/2668)}

## Changes proposed
Added Mock Oauth2 Server to api/docker-compose.yml 
Added custom config.json file with one test user 
Added custom mock-server-key.json file with one jwks key

## Context for reviewers
Pull mock server image specified in docker compose file and run it(Can also use `make start`) 
Can hit `http://localhost:5000/default/debugger` successfully 
`http://localhost:5000/issuer1/.well-known/openid-configuration` lists endpoints you can hit.


api/config.json
chouinar, line 1: Could we put this file and the mock-server-key.json into a folder within this directory?

Something like `mock-oauth` for the folder?

chouinar (commented):
Just an organizational comment, if there are any questions I still have, we'll sort those out as we get to them in the usage of this

emilycnava (commented):
Had to change port to 5001 on my machine as 5000 was colliding with something else, but aside from the port number issue both endpoints were reachable!

api/config.json
babebe: Yes! That would be better! Done

api/docker-compose.yml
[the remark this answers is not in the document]
babebe: updated to 5001

chouinar (approved):
LGTM!

PR #2836, opened 2024-11-13
[Issue 2665] Add search path to db connection params

DavidDudas-Intuitial:
## Summary
Fixes #2665 

### Time to review: __1 min__

Added search path to db connection params.

This is latest step in a series of attempts to resolve failed db connections and privileges from /analytics step functions. See ticket history and comments for details.

### DIFF
--- a/analytics/src/analytics/integrations/db.py
+++ b/analytics/src/analytics/integrations/db.py
@@ -52,6 +52,7 @@ def get_connection_parameters(config: DBSettings) -> dict[str, Any]:
     token = (
         config.password if config.local_env is True else generate_iam_auth_token(config)
     )
+    search_path = "public" if config.local_env is True else "app"
     return {
         "host": config.db_host,
         "dbname": config.name,
@@ -60,6 +61,7 @@ def get_connection_parameters(config: DBSettings) -> dict[str, Any]:
         "port": config.port,
         "connect_timeout": 20,
         "sslmode": config.ssl_mode,
+        "options": f"-c search_path={search_path}",
     }
 
 


coilysiren (dismissed):

analytics/src/analytics/integrations/db.py
```diff
@@ -52,6 +52,7 @@ def get_connection_parameters(config: DBSettings) -> dict[str, Any]:
     token = (
         config.password if config.local_env is True else generate_iam_auth_token(config)
     )
+    search_path = "public" if config.local_env is True else "app"
```
chouinar, line 55: I'd put this more explicitly in an environment variable in the config class itself.

Something like:
```py
db_schema: str = Field ("app", alias="DB_SCHEMA")
```

Then in local.env, put `DB_SCHEMA=public`

This makes it much easier to adjust or configure.

DavidDudas-Intuitial: Done! Please see latest commit

analytics/config.py
```diff
@@ -17,6 +17,7 @@ class DBSettings(PydanticBaseEnvConfig):
      user: str = Field (alias="DB_USER")
      password: Optional[str] = Field(None, alias="DB_PASSWORD")
      ssl_mode: str = Field("require", alias="DB_SSL_MODE")
+     db_schema: str = Field ("app", alias="DB_SCHEMA")
```
chouinar, line 20: You would still need to set the search path from this variable

chouinar, line 20: You would need to set something in local.env as well to override this to be `public` locally.

DavidDudas-Intuitial: I did that locally but did not commit the changes

DavidDudas-Intuitial: I'll push it up now

chouinar (approved):

PR #2837, opened 2024-11-13
[Issue #2721] Setup logic to validate login.gov tokens against their public keys

### LINKED ISSUE #2721: Add logic to process and validate login.gov JWT
### Summary

Follow-up to https://github.com/HHS/simpler-grants-gov/issues/2673

This ticket is explicitly defined roughly as a lot of the details for it need to be figured out as we implement it. However, we want the following.

Given a JWT (assumed to be from login.gov) we want to:
* Validate that the JWT came from login.gov (see below)
* Parse the JWT for its information

To do any of this parsing or validation, we want to use the pyJWT library which has an example for how some of this works: https://pyjwt.readthedocs.io/en/latest/usage.html#oidc-login-flow

In order to validate a JWT, you need to fetch the public key from login.gov and then pass it into the token decode endpoint. A JWT can be parsed without this, but we can't guarantee that the key is valid without this. Public keys can rotate, so we need to check the endpoint if the key we used does not work (assuming we don't have a KID to link them).

For parsing the jwt, we want to pull out the following fields:
* `sub` which is the login.gov user ID
* `email`

### Acceptance criteria

- [ ] Function exists that can take in a JWT + configuration and produce a parsed token (or raise an error if it failed)
- [ ] Validations should follow all implemented validations (issuer, token age, etc.)
- [ ] Rough process implemented to handle fetching new public keys
- [ ] Thorough tests of all validation scenarios

chouinar:
## Summary
Fixes #2721

## Changes proposed
Add logic to validate the authenticity of tokens from login.gov

Add rough logic for fetching public keys from login.gov

Add logic for parsing tokens from login.gov, extracting relevant info

## Context for reviewers
We will be processing the id_token returned by this login.gov endpoint https://developers.login.gov/oidc/token/

This token follows most general JWT conventions. In order to validate the authenticity of the token, we need to fetch the public key from login.gov (see: https://developers.login.gov/oidc/certificates/ ). This is all standard Oauth stuff, we just have to deal with complexities surrounding the caching of these public keys.

We assume that if a token fails to validate with the public key, that it might be with a key we don't yet know about, and refresh the keys (this is a pretty standard approach), however this refresh runs into thread-safety issues that are documented inline. In short, the API runs multiple threads and they might all try to refresh simultaneously. Whether that is an issue, we'll have to see.

Most of the actual validation on the token is done inside of the library and just boils down to:
* Proving it was generated with the key (the public key verification the JWT library handles
* Verifying timestamps are in the right order
* Verifying who sent us this token (clientID will need to be configured to match whatever the frontend sets up).

After all this, we're certain we know who they are, and can use the token, although the current logic stops after parsing.

## Additional information
More thorough testing of this behavior needs the frontend / integration with login.gov to be a bit further along. There are a few small unknowns in the details of the token we'll be parsing that might require small adjustments.

To avoid committing tokens, I make the tests fully generate the RSA tokens in the tests. This is the same as defining a public and private key in a very long form (the "BEGIN PUBLIC KEY..." files).


.trivyignore.yaml
```diff
@@ -1,9 +1,22 @@
...
+
+# vulnerabilities:
+#   - id: CVE-1234-5678
+
+secrets:
+  # PyJWT has an example with a fake JWT that Trivy flags.
+  # see: https://github.com/aquasecurity/trivy/discussions/5772
+  id: jwt-token
+  paths:
+    - "/api/.venv/lib/python3.13/site-packages/PyJWT-2.9.0.dist-info/METADATA"
```
coilysiren, line 22: very rude that trivy won't accept a wildcard in here

coilysiren (commented):
(leaving the actually approval to someone else, the trivy stuff LGTM)

.trivyignore.yaml
mdragon: Oof... so this is going to break every time we upgrade Python or PyJWT?

coilysiren: yep!

chouinar: While their docs don't mention wildcards, they might work? https://github.com/aquasecurity/trivy/issues/5120

Will test

chouinar: Unfortunately, wild cards do not work. So yeah, any version upgrade will break this.

Skipping directories in the scan technically could work with wildcards: https://aquasecurity.github.io/trivy/v0.56/docs/configuration/skipping/ - as long as we're fine with ignoring everything for this singular metadata file.

I'll look into it more tomorrow

chouinar: I think setting up skips like that requires a configuration file: https://aquasecurity.github.io/trivy/v0.56/docs/references/configuration/config-file/

We currently set some of these values with env vars. I think the right way would be to setup a config file to use, but that is definitely veering outside of the scope of this ticket, so I'll just revert it back to the direct path.

mdragon (approved):
Looks good!

chouinar:
I'm pretty sure the anchore scan issue is fixed, but this PR is running with an old cache. I could clean that up, but doesn't seem that important since it's clearly unrelated to this PR.

PR #2870, opened 2024-11-15
[Issue 2869] Add versioning to analytics db schema

### LINKED ISSUE #2869: Implement basic versioning in analytics db schema
### Summary

In /analytics the database initialization command currently executes a single non-versioned SQL file (non-versioned in terms of data model evolution). Implement a basic versioning pattern to facilitate extensibility. 

### Acceptance criteria

- [x] `schema_version` table is added to analytics db to keep track of schema version
- [x] `initialize_database` command is updated to execute multiple sequential SQL files, and ignore already-executed files, and update `schema_version` accordingly

DavidDudas-Intuitial:
## Summary
Fixes #2869 

1. Added a simple versioning pattern to enable schema extensibility, including a table in the schema to keep track of current schema version
2. Refactored `make init-db` command to enable execution of multiple SQL files, and update schema version accordingly upon the execution of each file

This is part of an effort to enable automated deployment of analytics db to staging and prod.


analytics/src/analytics/integrations/etldb/migrations/versions/2024_11_15_update_columns_etldb.sql
```diff
@@ -0,0 +1,25 @@
+CREATE TABLE IF NOT EXISTS gh_schema_version (
```
coilysiren, line 1: The schema_version table creation needs to go into it's own file

Also it should just be called `schema_version`, with no `gh_`

analytics/src/analytics/integrations/etldb/migrations/versions/2024_11_15_update_columns_etldb.sql
```diff
@@ -0,0 +1,25 @@
...
+	version INTEGER NOT NULL UNIQUE,
+	t_created TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+	t_modified TIMESTAMP DEFAULT CURRENT_TIMESTAMP
+);
+
+INSERT INTO gh_schema_version (version) values (2) ON CONFLICT (version) DO NOTHING;
```
coilysiren, line 7: This value (the schema version) is generated by looking at your position on the list of files, rather than being set here manually

analytics/src/analytics/integrations/etldb/main.py
```diff
@@ -15,37 +15,47 @@
 from analytics.integrations.etldb.sprint_model import EtlSprintModel
 
 VERBOSE = False
+MIN_SCHEMA_VERSION = 2
```
coilysiren, line 18: Retrieve this value from the schema version table

analytics/src/analytics/integrations/etldb/main.py
```diff
@@ -15,37 +15,47 @@
...
         OperationalError,
         InsufficientPrivilege,
     ) as e:
-        message = f"FATAL: Failed to initialize db: {e}"
+        message = f"FATAL: Failed to initialize database client: {e}"
+        raise RuntimeError(message) from e
+
+    # execute sql
+    try:
+        for sql_path in get_sql_files():
```
coilysiren, line 38: 1. Get the schema version here, as a special case
2. Pass in the schema version as an input into this function
3. Start iterating based on the position in the schema version

coilysiren (changes requested):
I have a set of comments whose cohesive goal is around making this process truly automatic and resilient to user (eg. developer) error. Most of that revolves around how schema versioning is handled.

coilysiren:
One more high level comment: this schema version stuff needs to be tested on CI. Add a line into the analytics CI file that runs it

analytics/src/analytics/integrations/etldb/main.py
```diff
@@ -190,3 +209,13 @@ def sync_quads(db: EtlDb, dataset: EtlDataset) -> dict:
...
             )
     return result
+
+
+def get_sql_files() -> list[str]:
+    """Get all sql files needed for database initialization."""
+    # define the path to the sql files
+    parent_path = Path(__file__).resolve().parent
+    return [
+        f"{parent_path}/migrations/versions/2024_11_14_create_tables_etldb.sql",
```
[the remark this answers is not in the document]
DavidDudas-Intuitial: done

analytics/src/analytics/integrations/etldb/migrations/versions/2024_11_15_update_columns_etldb.sql
DavidDudas-Intuitial: done

analytics/src/analytics/integrations/etldb/migrations/versions/2024_11_15_update_columns_etldb.sql
DavidDudas-Intuitial: done

analytics/src/analytics/integrations/etldb/main.py
DavidDudas-Intuitial: done

analytics/src/analytics/integrations/etldb/migrations/versions/0000_drop_tables_etldb.sql.ignore
```diff
@@ -0,0 +1,15 @@
+DROP TABLE IF EXISTS schema_version;
```
coilysiren, line 1: Did you mean to leave this file in? I don't understand what it's for

analytics/src/analytics/integrations/etldb/main.py
```diff
@@ -190,3 +195,26 @@ def sync_quads(db: EtlDb, dataset: EtlDataset) -> dict:
...
+    result = {}
+
+    # define the path to the sql files
+    sql_file_directory = f"{Path(__file__).resolve().parent}/migrations/versions"
+
+    # get list of sorted filenames
+    filename_list = sorted(os.listdir(sql_file_directory))
+
+    # compile result
+    pattern = re.compile(r"^\d\d\d\d_.+\.sql$")
```
coilysiren, line 211: This SQL looks like it's still validating based on using datetime in SQL files, instead of 000X

coilysiren (commented):
It's looking close, but I want to see it [working in CI](https://github.com/HHS/simpler-grants-gov/blob/4e1ed8f71180e733002d62c0dd1920623b178b14/.github/workflows/ci-analytics.yml#L42-L43) to confirm that it's actually working

coilysiren:
> Not sure I understand. What is "the analytics CI file"?

https://github.com/HHS/simpler-grants-gov/blob/main/.github/workflows/ci-analytics.yml

analytics/src/analytics/integrations/etldb/migrations/versions/0000_drop_tables_etldb.sql.ignore
DavidDudas-Intuitial: Yes, it's useful during dev. Notice the `.ignore` suffix, which causes the `init-db` command to ignore it

DavidDudas-Intuitial: Having said that, I can remove it if it raises concerns. LMK

analytics/src/analytics/integrations/etldb/migrations/versions/0002_create_schema_versions.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE IF NOT EXISTS schema_version (
```
chouinar, line 1: Wouldn't we want this to be created first before anything else?

analytics/src/analytics/integrations/etldb/main.py
DavidDudas-Intuitial: No, it's looking for a file that starts with 4 decimal digits (denoted by \d) followed by an arbitrary string, and ending in ".sql"

DavidDudas-Intuitial:
> It's looking close, but I want to see it [working in CI](https://github.com/HHS/simpler-grants-gov/blob/4e1ed8f71180e733002d62c0dd1920623b178b14/.github/workflows/ci-analytics.yml#L42-L43) to confirm that it's actually working

I would expect `make init-db` to always fail in CI because it will not be able to connect to Postgres. Can you clarify? Which Postgres instance is available from CI, if any?

analytics/src/analytics/integrations/etldb/main.py
```diff
@@ -17,35 +19,38 @@
...
-        message = f"FATAL: Failed to initialize db: {e}"
-        raise RuntimeError(message) from e
+    # get connection to database
+    etldb = EtlDb()
+    current_version = etldb.get_schema_version()
+    print(f"current schema version: {current_version}")
+
+    # get all sql file paths and respective version numbers
+    sql_file_path_map = get_sql_file_paths()
+    all_versions = sorted(sql_file_path_map.keys())
```
chouinar, line 31: Rather than needing to re-sort here. What about either returning tuples or class objects in sorted order?

Something very basic like:
```py
from dataclasses import dataclass

@dataclass
MigrationFile:
    version: int
    file_path: str

```

analytics/src/analytics/integrations/etldb/main.py
```diff
@@ -17,35 +19,38 @@
...
+            print(f"applying updates for schema version: {next_version}")
+            cursor = etldb.connection()
+            cursor.execute(
+                text(sql),
+            )
+            # commit changes
+            etldb.commit(cursor)
+            # bump schema version number
+            etldb.set_schema_version(next_version)
+            current_version = next_version
```
chouinar, line 50: Minor nit: I think you can get rid of this? If there are multiple files to process, current_version will always be the last one processed (before the current run), so you just want everything after than anyways.

analytics/src/analytics/integrations/etldb/main.py
```diff
@@ -190,3 +195,26 @@ def sync_quads(db: EtlDb, dataset: EtlDataset) -> dict:
...
+    # get list of sorted filenames
+    filename_list = sorted(os.listdir(sql_file_directory))
+
+    # compile result
+    pattern = re.compile(r"^\d\d\d\d_.+\.sql$")
+    for filename in filename_list:
+        # validate filename format
+        if pattern.match(filename):
+            version = int(filename[:4])
+            result[version] = f"{sql_file_directory}/{filename}"
```
chouinar, line 216: Regexes can trip some folks up (I'm pretty mediocre with them), what about something a bit more direct like:

```py
if filename.endswith(".sql"):
     version = int(filename.split("_")[0])
     ...
```

analytics/src/analytics/integrations/etldb/migrations/versions/0002_create_schema_versions.sql
DavidDudas-Intuitial: Ideally yes, but it's too late because the schema is already deployed and in use. Therefore I'm using this as an opportunity to modify it in place without breaking anything, thereby using this as a proof of concept.

analytics/src/analytics/integrations/etldb/main.py
DavidDudas-Intuitial: That's another way to do it! :) 
I like the explicit sorting because it shows very clearly, without having to look elsewhere, that the sql files are being executed in order of version.

analytics/src/analytics/integrations/etldb/main.py
DavidDudas-Intuitial: This protects against sql files that duplicate the 4 digit prefix, e.g. 
`0007_do_stuff.sql`
`0007_do_more_stuff.sql`

analytics/src/analytics/integrations/etldb/main.py
DavidDudas-Intuitial: That would fail if the filename does not include and underscore. The regex enforces the file naming convention.

analytics/src/analytics/integrations/etldb/main.py
```diff
@@ -17,35 +19,38 @@
...
-        ProgrammingError,
-        OperationalError,
-        InsufficientPrivilege,
-    ) as e:
-        message = f"FATAL: Failed to initialize db: {e}"
-        raise RuntimeError(message) from e
+    # get connection to database
+    etldb = EtlDb()
+    current_version = etldb.get_schema_version()
+    print(f"current schema version: {current_version}")
```
coilysiren, line 27: Can you make it to print out each migration (file name, version, etc) one at a time as it's running. I just noticed that there's only one print statement being emitted in CI, which isn't enough to debug if migrations are broken.

DavidDudas-Intuitial: Yes, I had those in there and took them out after I saw your other comment about print statements, haha. I'll put them back in.

DavidDudas-Intuitial: Done

DavidDudas-Intuitial: Output looks like this now:
```
$ make init-db
=> Initializing the database schema
=====================================================
poetry run analytics etl initialize_database
initializing database
current schema version: 0
applying migration for schema version: 1
migration source file: /Users/david/Code/simpler-grants-gov/analytics/src/analytics/integrations/etldb/migrations/versions/0001_create_tables_etldb.sql
applying migration for schema version: 2
migration source file: /Users/david/Code/simpler-grants-gov/analytics/src/analytics/integrations/etldb/migrations/versions/0002_create_schema_versions.sql
applying migration for schema version: 3
migration source file: /Users/david/Code/simpler-grants-gov/analytics/src/analytics/integrations/etldb/migrations/versions/0003_alter_tables_set_default_timestamp.sql
total migrations applied: 3
new schema version: 3
done
```

DavidDudas-Intuitial: If there are no migrations to apply, it will look like this:
```
$ make init-db
=> Initializing the database schema
=====================================================
poetry run analytics etl initialize_database
initializing database
current schema version: 3
total migrations applied: 0
new schema version: 3
done
```

coilysiren:
It looks like the migrations aren't actually running? 

https://github.com/HHS/simpler-grants-gov/actions/runs/11863350979/job/33064630902#step:9:27

coilysiren:
The database is recreated on every run of CI, so it should be applying 3 migrations, not 0

DavidDudas-Intuitial:
> The database is recreated on every run of CI, so it should be applying 3 migrations, not 0

I don't think the db is being recreated. There's no way the schema version would be set to 3 in a fresh db.

analytics/src/analytics/integrations/etldb/migrations/versions/0000_drop_tables_etldb.sql.ignore
DavidDudas-Intuitial: I removed it out of an abundance of caution

coilysiren:
> I don't think the db is being recreated

Then successive runs of your PR won't produce valid test results, because you'll want to see the migrations re-run on each commit in case you change them.

Add this (or some variant thereof) to ensure that we can actually see test results in CI for your changes:

https://github.com/HHS/simpler-grants-gov/pull/2892/commits/725dec6a26f40ea6982f85b528d786c8253fbdef

It's working in CI, by the way. I'll approve after I see this one change. There's of course Michael's comments though.

DavidDudas-Intuitial:
> Then successive runs of your PR won't produce valid test results, because you'll want to see the migrations re-run on each commit in case you change them

Agreed!

analytics/src/analytics/integrations/etldb/main.py
```diff
@@ -17,35 +19,38 @@
 VERBOSE = False
 
 
-def init_db() -> None:
+def initialize_database() -> None:
     """Initialize etl database."""
```
widal001, line 23: Can we make this docstring more informative? It would be great to get a summary of how the migrations _should_ work in this function.

Something like:

```python
def initialize_database() -> None:
    """
    Sets up the etl database by applying a sequential set of migration scripts.
    
    It applies the migrations using the following steps:
    - Check the current schema version listed in the database
    - Retrieve the list of migration scripts ordered by version
    - If the current schema version is less than the version of the latest migration script
      run the remaining migrations in order
    - Bump the schema version in the database to the latest version
    - If the current schema version matches the latest script, do nothing
    """
```

I often find docstrings are most helpful when they express intent, so that future maintainers know how a function was designed to work rather than having to rely on a close reading of the code. It also helps distinguish potential bugs vs features.

widal001 (commented):
Excited to get migrations up and running!

General approach looks solid to me, though I know Kai and Michael have been following this work much more closely. Everything seems to have run as expected:

Screenshot 2024-11-18 at 10 06 10 AM

I left a few comments requesting some additional clarity in a few docstrings and code comments. I don't necessarily want to block the merge, but would love to see those added before I approve.

analytics/src/analytics/integrations/etldb/main.py
widal001: I don't have a strong preference about this but want to check my understanding on how the code is supposed to behave in the case when there are multiple migration scripts with the same version:

- Are just skipping over duplicate versions? Or raising an error? I imagine we want to due the latter in get_sql_file_paths()
- How are we ordering between versions if there are duplicates (if we're not raising an error)?
- Also if we're fetching the path from sql_file_path_map then I don't actually think we'd actually have duplicate versions since it's keyed on version.

```python
for filename in filename_list:
    # validate filename format
    if pattern.match(filename):
        version = int(filename[:4])
        result[version] = f"{sql_file_directory}/{filename}"
```

mdragon: Yeah, I would agree that I would expect a fatal error if the files don't match expectations, not running one? of the offending files per alpha sorting and silently ignoring the rest? if I'm reading this right?

analytics/src/analytics/integrations/etldb/main.py
```diff
@@ -190,3 +202,26 @@ def sync_quads(db: EtlDb, dataset: EtlDataset) -> dict:
...
+    # get list of sorted filenames
+    filename_list = sorted(os.listdir(sql_file_directory))
+
+    # execute file only if filename follows convention: 0000_summary_of_change.sql
+    pattern = re.compile(r"^\d\d\d\d_.+\.sql$")
+    for filename in filename_list:
+        # validate filename format
+        if pattern.match(filename):
+            version = int(filename[:4])
+            result[version] = f"{sql_file_directory}/{filename}"
```
widal001, line 223: Wanted to pull [this comment](https://github.com/HHS/simpler-grants-gov/pull/2870#discussion_r1846801062) to a top-level review thread since it was getting lost earlier in the PR.

If I'm understanding this correctly, it looks like if there are two SQL files with the same version, we'll end up skipping the first one (by overwriting its path in the map with the second file)

As @mdragon suggested, I'd strongly recommend that we fail and halt execution under two likely scenarios:
- Two files with the same version prefix
- A file that doesn't match this pattern

The current approach risks downstream problems by having certain migrations be silently skipped unless verbose is enabled.

widal001 (commented):
After reading a couple of the threads more closely, I'd like to either clarify the behavior of the code below and/or change it to avoid migrations being silently skipped before we merge.

analytics/src/analytics/integrations/etldb/main.py
DavidDudas-Intuitial: Good catch. I'll raise an exception if there are duplicate version, rather than silently ignore them.

analytics/src/analytics/integrations/etldb/main.py
DavidDudas-Intuitial: Thank you, I will add this to the file.

analytics/src/analytics/integrations/etldb/main.py
DavidDudas-Intuitial: Yes

analytics/src/analytics/integrations/etldb/main.py
DavidDudas-Intuitial: Done - please see latest commits

analytics/src/analytics/integrations/etldb/main.py
DavidDudas-Intuitial: Done - please see latest commits

analytics/src/analytics/integrations/etldb/main.py
DavidDudas-Intuitial: Done - please see latest commits

analytics/src/analytics/integrations/etldb/main.py
DavidDudas-Intuitial: Done - please see latest commits

DavidDudas-Intuitial:
@widal001 @mdragon @coilysiren All comments have been addressed. Can I get another review?

coilysiren (dismissed):
I just checked again, and migrations still aren't being run in CI.

See: https://github.com/HHS/simpler-grants-gov/pull/2870#issuecomment-2480232140

DavidDudas-Intuitial:
> I just checked again, and migrations still aren't being run in CI.

@coilysiren Where did you check, and what did you see?

DavidDudas-Intuitial:
I see that the migrations are indeed [running in CI](https://github.com/HHS/simpler-grants-gov/actions/runs/11918006568/job/33214427390), as evidenced by this output:
```
current schema version: 3
total migrations applied: 0
new schema version: 3
```

coilysiren:
They aren't being run _**correctly**_ in CI, in that they aren't actually applying the 3 migrations. We don't know why, but it doesn't matter why. We just want the migrations themselves to actually be tested. We can do that by dropping the docker volume that stores the postgres data, which is what my commit does.

DavidDudas-Intuitial:
> they aren't actually applying the 3 migrations

Because they've already been applied. The code is doing what it's supposed to: skip the already applied transformation. Evidently CI is *not* rebuilding the container and/or standing up a fresh db instance with each check.

> We just want the migrations themselves to actually be tested. We can do that by dropping the docker volume that stores the postgres data, which is what my commit does.

What commit are you referring to?

coilysiren:
> How?

https://github.com/HHS/simpler-grants-gov/commit/725dec6a26f40ea6982f85b528d786c8253fbdef

DavidDudas-Intuitial:
Are you saying I need to add your commit to this branch?

coilysiren:
> Are you saying I need to add your commit to this branch?

Yes, add the changes from that commit into this branch so we can see that the migrations are actually being performed as intended

chouinar:
> Are you saying I need to add your commit to this branch?

@coilysiren / @DavidDudas-Intuitial 
I've only been partially following this discussion, but why would we need to drop the docker volumes in the CI/CD? Those would be completely fresh for every run unless the analytics CI is caching that across runs.

I am wondering why `init-db` is called after all the unit tests, I would think the tests would need the DB to be setup beforehand if any of them interact with the DB.

DavidDudas-Intuitial:
> ...completely fresh for every run unless the analytics CI is caching that across runs

There is indeed some caching going on. The db instance persists across runs.
 
> I am wondering why `init-db` is called after all the unit tests, I would think the tests would need the DB to be setup beforehand if any of them interact with the DB

You are correct, the tests need to run after the db is initialized. I just changed the sequencing in my most recent commit.

DavidDudas-Intuitial:
@coilysiren see most recent [run](https://github.com/HHS/simpler-grants-gov/actions/runs/11919878649/job/33220575343?pr=2870)

```
current schema version: 0
applying migration for schema version: 1
migration source file: /analytics/src/analytics/integrations/etldb/migrations/versions/0001_create_tables_etldb.sql
applying migration for schema version: 2
migration source file: /analytics/src/analytics/integrations/etldb/migrations/versions/0002_create_schema_versions.sql
applying migration for schema version: 3
migration source file: /analytics/src/analytics/integrations/etldb/migrations/versions/0003_alter_tables_set_default_timestamp.sql
total migrations applied: 3
new schema version: 3
```

DavidDudas-Intuitial:
Screenshot 2024-11-19 at 11 13 26 AM

DavidDudas-Intuitial:
@coilysiren I've applied your commit which resets the database instance. As expected, the transformations are all applied in CI. Do you have any other change requests or can this be merged?

coilysiren:
@chouinar 

> I've only been partially following this discussion, but why would we need to drop the docker volumes in the CI/CD? Those would be completely fresh for every run unless the analytics CI is caching that across runs.
>
> I am wondering why `init-db` is called after all the unit tests, I would think the tests would need the DB to be setup beforehand if any of them interact with the DB.

I wish I could tell you 🤷🏼‍♀️ I just know that the migrations weren't being run prior to dropping the volume

coilysiren (approved):
this LGTM, professionally though I would give the other folks till EOD or something (they are all ET) to see if they have any feedback as well

analytics/src/analytics/integrations/etldb/main.py
```diff
@@ -17,35 +19,55 @@
...
+    """
+    Create and/or update an etl database by applying a sequential set of migration scripts.
+
+    It applies the migrations using the following steps:
+    - Check the current schema version listed in the database
+    - Retrieve the list of migration scripts ordered by version
+    - If the current schema version is less than the version of the latest migration script
+      run the remaining migrations in order
+    - Bump the schema version in the database to the latest version
+    - If the current schema version matches the latest script, do nothing
```
widal001, line 32: Thanks for adding this!!

analytics/src/analytics/integrations/etldb/main.py
```diff
@@ -190,3 +212,38 @@ def sync_quads(db: EtlDb, dataset: EtlDataset) -> dict:
...
+
+    # define the path to the sql files
+    sql_file_directory = f"{Path(__file__).resolve().parent}/migrations/versions"
+
+    # get list of sorted filenames
+    filename_list = sorted(os.listdir(sql_file_directory))
+
+    # expected filename format: {4_digit_version_number}_{short_description}.sql
+    # example: 0003_alter_tables_set_default_timestamp.sql
+    pattern = re.compile(r"^\d\d\d\d_.+\.sql$")
```
widal001, line 229: Thanks for adding this!!

analytics/src/analytics/integrations/etldb/main.py
```diff
@@ -190,3 +212,38 @@ def sync_quads(db: EtlDb, dataset: EtlDataset) -> dict:
...
+            message = f"FATAL: malformed db migration filename: {filename}"
+            raise RuntimeError(message)
+
+        # extrace version number from filename
+        version = int(filename[:4])
+
+        # do not allow duplicate version number
+        if version in result:
+            message = f"FATAL: Duplicate db migration version number: {version} "
+            raise RuntimeError(message)
```
widal001, line 244: Appreciate this extra check!

widal001 (approved):
LGTM! Thanks for addressing all of our requests!

Everything is looking good, and the safety measure to catch duplicate is working well:

Passed okay with existing code:

Screenshot 2024-11-19 at 4 28 36 PM

Correctly raised runtime error with duplicate migration versions:

Screenshot 2024-11-19 at 4 30 38 PM

Issue #2794, opened 2024-11-08
Modify `export_opportunity_data_task` to create `ExtractMetadata` database records


mikehgrantsgov:
Related to [2454](https://github.com/HHS/simpler-grants-gov/issues/2454)

### Summary

[2793](https://github.com/HHS/simpler-grants-gov/issues/2793) creates the API needed to send records back to the database. This ticket populates that table with results from the `export_opportunity_data_task` task.

Something like:

```
def run_task(self) -> None:

   .... existing code ....
   self.record_metadata("json", self.json_file)
   self.record_metadata("csv", self.csv_file)

   def record_metadata(self, extract_type: str, file_path: str) -> None:
    file_name = os.path.basename(file_path)
    file_size = os.path.getsize(file_path)
    metadata = ExtractMetadata(
        extract_type=extract_type,
        file_name=file_name,
        file_path=file_path,
        file_size=file_size
    )
    self.db_session.add(metadata)
    self.db_session.commit()
```

**Note:** The file path should be a public URL to the file for download. If we go a presigned URL, we will need a follow-up task to generate presigned URLs in the `ExtractMetadata` response.

### Acceptance criteria

- [ ] Opportunity export task updated to add db rows
- [ ] Tests added to `test_export_opportunity_data_task` to make sure db records are being created.

Issue #2793, opened 2024-11-08
Create the `ExtractMetadata` API


mikehgrantsgov:
Related to [2454](https://github.com/HHS/simpler-grants-gov/issues/2454)

### Summary

[2792](https://github.com/HHS/simpler-grants-gov/issues/2792) creates the input/output schemas needed to support the API. This ticket creates the API to read params from the request and send back extract metadata information.

Something like:

POST  /v1/extracts/opportunities

Input:
ExtractMetadataRequestSchema

If `extract_type` is not provided, default to `opportunity_data_extract` (we may support more later)
If `start_date` and `end_date` not provided, default to where `created_at` is within the last 7 days.

Something like (rough pseudocode):
```
@opportunity_blueprint.post("/extracts/opportunities")
@opportunity_blueprint.output(extract_metadata_schemas.ExtractMetadataListResponseSchema)
@opportunity_blueprint.auth_required(api_key_auth)
@flask_db.with_db_session()
def extract_metadata_get(db_session: db.Session) -> ApiResponse:
    # Get query parameters

    # Default to last 7 days if no date range is provided

    # Call service with params to get results

    # Serialize results

    # return ApiResponse(message="Success", data=metadata_list)
```

### Acceptance criteria

- [ ] API created with supporting service
- [ ] Test to seed the db and read these values. Tests include parameter permutations

chouinar:
We'll want to be careful about doing `GET` with a request body. It's fuzzy whether that's technically "right" in HTTP requests. Some libraries apparently won't let you do that. Either need everything to be request params or just make it a `POST` endpoint (I generally just prefer the latter).

I'm trying to think if we'll have more extracts in the future to warrant a naming schema like:
* `/v1/extracts/opportunities`
* `/v1/extracts/something-else-down-the-road`

mikehgrantsgov:
Makes sense, adapted this for `/v1/extracts/opportunities` now and can add more.

Issue #2792, opened 2024-11-08
Create input / output schemas for the `ExtractMetadata` API


mikehgrantsgov:
Related to [2454](https://github.com/HHS/simpler-grants-gov/issues/2454)

### Summary

[2791](https://github.com/HHS/simpler-grants-gov/issues/2791) creates the database table to store extracts. This ticket creates the input/output schemas needed by the API to support querying extracts and returning data.

Something like:

Input:
```
class ExtractMetadataRequestSchema(AbstractResponseSchema):
    extract_type = fields.String(
        allow_none=True,
        metadata={
            "description": "The type of extract to filter by",
            "example": "opportunity_data_extract",
        }
    )
    start_date = fields.Date(
        allow_none=True,
        metadata={
            "description": "The start date for filtering extracts",
            "example": "2023-10-01",
        }
    )
    end_date = fields.Date(
        allow_none=True,
        metadata={
            "description": "The end date for filtering extracts",
            "example": "2023-10-07",
        }
    )
```

Output:
```
class ExtractMetadataResponseSchema(AbstractResponseSchema):
    id = fields.Integer(metadata={"description": "The ID of the extract metadata", "example": 1})
    extract_type = fields.String(
        metadata={"description": "The type of extract", "example": "opportunity_data_extract"}
    )
    file_name = fields.String(
        metadata={"description": "The name of the file", "example": "data_extract.csv"}
    )
    file_path = fields.String(
        metadata={"description": "The path to the file", "example": "/path/to/data_extract.csv"}
    )
    file_size = fields.Integer(
        metadata={"description": "The size of the file in bytes", "example": 1024}
    )
    created_on = fields.DateTime(
        metadata={"description": "When the extract was created", "example": "2023-10-01T12:00:00"}
    )
    updated_on = fields.DateTime(
        metadata={"description": "When the extract was last updated", "example": "2023-10-02T12:00:00"}
    )

class ExtractMetadataListResponseSchema(Schema):
    data = fields.List(fields.Nested(ExtractMetadataResponseSchema), metadata={
        "description": "A list of extract metadata records"
    })
```

### Acceptance criteria

- [ ] Schemas created
- [ ] Test to convert DB model to Response list model

chouinar:
One minor detail - the response schema should a subclass of `AbstractResponseSchema` so that it gets some of the default API response stuff

mikehgrantsgov:
Included this, thanks.

PR #2880, opened 2024-11-15
[Issue #2791] Create database table ExtractMetadata to store extract files

### LINKED ISSUE #2791: Create database table `ExtractMetadata` to store extract files

### Summary

The legacy website provides the ability to download XML extracts of opportunity data: https://www.grants.gov/xml-extract - this is something we should also support. Having extracts (especially if they're just the same data as our search endpoints) reduces the traffic our search endpoint will need to handle as we can point anyone that would want to scrape our data to an extract file.

Create a table to track extract metadata. 

Something like:
```

# Create LkExtractType as a lookup value with opportunities_xml and opportunities_csv as initial values

class ExtractMetadata(Base):
    __tablename__ = 'extract_metadata'

    extract_metadata_id = Column(BigInteger, primary_key=True, autoincrement=True)
    extract_type = Column(LkExtractType, nullable=False)
    file_name = Column(String, nullable=False)
    file_path = Column(String, nullable=False)
    file_size_bytes = Column(BigInteger, nullable=False)
    created_on = Column(DateTime, server_default=func.now(), nullable=False)
    updated_on = Column(DateTime, server_default=func.now(), onupdate=func.now(), nullable=False)

```

### Acceptance criteria

- [ ] Model created
- [ ] Database migration generated

mikehgrantsgov:
## Summary
Fixes #2791

## Changes proposed
- Add a lookup table `LkExtractType`
- Add a `ExtractMetadata` model and associated table for future use by the backend.

## Context for reviewers
This issue unblocks many other backend tickets from creating and storing the necessary db extracts in S3.
See #2792, #2793, #2794

## Additional information
See generated migration.


api/src/db/models/extract_models.py
```diff
@@ -0,0 +1,21 @@
...
+
+    extract_metadata_id = Column(BigInteger, primary_key=True, autoincrement=True)
+    extract_type: Mapped[ExtractType] = mapped_column(
+        "extract_type_id",
+        LookupColumn(LkExtractType),
+        ForeignKey(LkExtractType.extract_type_id),
+    )
+    file_name = Column(String, nullable=False)
+    file_path = Column(String, nullable=False)
+    file_size_bytes = Column(BigInteger, nullable=False)
```
chouinar, line 21: SQLAlchemy 2 is better at type interpretation, we can leave columns out entirely if we just need types/nullability. Think of it kinda like how Pydantic or Dataclasses work, but just with needing `Mapped` around it so SQLAlchemy knows you want it to be a part of the DB object.


```suggestion
    file_name: Mapped[str]
    file_path: Mapped[str]
    file_size_bytes: Mapped[str]
```

api/src/db/models/extract_models.py
```diff
@@ -0,0 +1,21 @@
...
+from src.adapters.db.type_decorators.postgres_type_decorators import LookupColumn
+from src.constants.lookup_constants import ExtractType
+from src.db.models.base import ApiSchemaTable, TimestampMixin
+from src.db.models.lookup_models import LkExtractType
+
+
+class ExtractMetadata(ApiSchemaTable, TimestampMixin):
+    __tablename__ = "extract_metadata"
+
+    extract_metadata_id = Column(BigInteger, primary_key=True, autoincrement=True)
```
chouinar, line 13: Want to use mapped_column, I think I saw Column is being deprecated (already deprecated?) in Sqlalchemy 2.

Autoincrement ends up not needed, a primary key int column ends up as a Serial type which does that automatically.
```suggestion
    extract_metadata_id = mapped_column(BigInteger, primary_key=True)
```

chouinar (approved):
LGTM!

Issue #2673, opened 2024-10-30
Create POST /users/token stub endpoint


chouinar:
### Summary

We want to create a new POST /user/token endpoint

**Auth**
This endpoint will have the same api_auth_token auth that most of our other endpoints have

**Request params**
The only request param that this endpoint will take in is a header param named `X-OAuth-login-gov`

**Response params**
The API should return a response that contains the token we generated, as well as a user, and whether we created the user new:
```json
{
  "data": { 
    "token": "the token goes here!",
    "user": {
         "user_id": "abc-...",
         "email": "example@mail.com",
         "external_user_type": "login_gov",
     },
    "is_user_new": true
 },
  "message": "Success",
  "status_code": 200
}
```

### Acceptance criteria

- [ ] Endpoint can receive a token and return a static token
- [ ] Endpoint should error if the request param is not in the header

PR #2885, opened 2024-11-15
[Issue 2673] Users token stub endpoint

babebe:
## Summary
Fixes #{[2673](https://github.com/HHS/simpler-grants-gov/issues/2673)}

## Changes proposed

- Added `POST v1/users/user/token` endpoint 
- Added `UserTokenHeaderSchema`, `UserTokenResponseSchema` , `UserTokenSchema`, `UserSchema`
- Added `users` blueprint.  Register blueprint only when local
- Added tests

## Context for reviewers
This is a stub endpoint that returns an error response if called without the required header. Otherwise it returns a static response


api/src/api/opportunities_v1/opportunity_routes.py
```diff
@@ -212,3 +212,28 @@ def opportunity_versions_get(db_session: db.Session, opportunity_id: int) -> res
         data = get_opportunity_versions(db_session, opportunity_id)
 
     return response.ApiResponse(message="Success", data=data)
+
+
+@opportunity_blueprint.post("/user/token")
+@opportunity_blueprint.output(opportunity_schemas.OpportunityUserTokenResponseV1Schema)
+@opportunity_blueprint.auth_required(api_key_auth)
```
chouinar, line 219: This should be in a separate blueprint from opportunities.

I'd recommend we create an `auth` folder / blueprint / etc. instead.

api/src/api/opportunities_v1/opportunity_routes.py
```diff
@@ -212,3 +212,28 @@ def opportunity_versions_get(db_session: db.Session, opportunity_id: int) -> res
...
+                "external_user_type": "login_gov",
+            },
+            "is_user_new": True,
+        }
+        return response.ApiResponse(message="Success", data=data)
+
+    message = "Missing X-OAuth-login-gov header"
+    logger.error(message)
+
+    return response.ApiResponse(message=message, status_code=400)
```
chouinar, line 239: For returning a non 2xx, I'd recommend we use `raise_flask_error` which handles all of the fun building-an-error response logic (well - really, handles throwing an error in the expected form so the error handling logic works right)

api/src/api/opportunities_v1/opportunity_schemas.py
```diff
@@ -535,3 +535,39 @@ class OpportunitySearchResponseV1Schema(AbstractResponseSchema, PaginationMixinS
...
+        }
+    )
+    email = fields.String(
+        metadata={
+            "description": "The email address returned from Oauth2 provider",
+            "example": "js@gmail.com",
+        }
+    )
+    external_user_type = fields.String(
+        metadata={
```
chouinar, line 554: This should be an enum pointing to the ExternalUserType that was created

api/src/api/opportunities_v1/opportunity_schemas.py
```diff
@@ -535,3 +535,39 @@ class OpportunitySearchResponseV1Schema(AbstractResponseSchema, PaginationMixinS
...
+            "description": "The Oauth2 provider through which a user was authenticated",
+            "example": "login_gov",
+        }
+    )
+
+
+class OpportunityUserTokenResponseV1Schema(AbstractResponseSchema):
```
chouinar, line 561: The abstract response schema usage means we should have something like:

```
class UserTokenResponseSchema(AbstractResponseSchema):
     data = fields.Nested(SomeOtherSchema())
```

Because we want the response to look like:
```json
{
   "data": {
      // the actual meaningful stuff goes here
   },
   ...
}
```

api/src/api/opportunities_v1/opportunity_routes.py
babebe: Added the auth folder with the blueprint, routes and schema files

api/src/api/opportunities_v1/opportunity_schemas.py
babebe: That makes sense. Fixed!

api/src/api/opportunities_v1/opportunity_schemas.py
babebe: Updated this to enum `ExternalUserType`

api/src/api/opportunities_v1/opportunity_routes.py
babebe: Awesome! Raising 400 error with `raise_flask_error`

api/src/api/opportunities_v1/opportunity_routes.py
chouinar: I'd make a folder called users (similar to `opportunities_v1`). The blueprint you create should have the prefix `/v1/users` - the route would then be `/v1/users/token` when all put together.

api/src/api/users/user_routes.py
```diff
@@ -0,0 +1,52 @@
...
+
+This endpoint in its current form is primarily for testing and feedback.
+
+Features in this endpoint are still under heavy development, and subject to change. Not for production use.
+
+See [Release Phases](https://github.com/github/roadmap?tab=readme-ov-file#release-phases) for further details.
+"""
+
+
+@user_blueprint.post("/user/token")
```
chouinar, line 25: The endpoint should just be `/token` here - The full URL should be `/v1/users/token`

api/src/api/users/user_routes.py
```diff
@@ -0,0 +1,52 @@
...
+@user_blueprint.input(
+    user_schemas.UserTokenHeaderSchema, location="headers", arg_name="x_oauth_login_gov"
+)
+@user_blueprint.output(user_schemas.UserTokenResponseSchema)
+@user_blueprint.auth_required(api_key_auth)
+@user_blueprint.doc(  # should we include this?
+    description=SHARED_ALPHA_DESCRIPTION, responses=[200, 400]
+)
+def user_token(x_oauth_login_gov: str) -> response.ApiResponse | Response:
```
chouinar, line 34: mypy won't catch this - but the type is going to be the json/dict object.

```suggestion
def user_token(x_oauth_login_gov: dict) -> response.ApiResponse | Response:
```

api/src/api/users/user_routes.py
```diff
@@ -0,0 +1,52 @@
...
+                "user_id": "abc-...",
+                "email": "example@gmail.com",
+                "external_user_type": "login_gov",
+            },
+            "is_user_new": True,
+        }
+        return response.ApiResponse(message="Success", data=data)
+
+    message = "Missing X-OAuth-login-gov header"
+    logger.error(message)
```
chouinar, line 50: Note - even once built out, we don't want to log an error.

To a user it is an error sure, but error logs on our side would trip alerts. A user calling our endpoint and hitting a 4xx is not something we want to alert.

We only want to log non-info if it is bad for the system.
```suggestion
    logger.info(message)
```

babebe: Makes sense!

api/src/app.py
```diff
@@ -37,6 +39,13 @@
 """
 
 
+class AuthEndpointConfig(PydanticBaseEnvConfig):
+    auth_endpoint: bool = True if os.getenv("ENVIRONMENT", "local") == "local" else False
+
+
+auth_endpoint_config = AuthEndpointConfig()
```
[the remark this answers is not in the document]
babebe: Moved this to `register_blueprints` before conditional stmt

api/src/api/users/user_routes.py
```diff
@@ -0,0 +1,49 @@
...
+SHARED_ALPHA_DESCRIPTION = """
+__ALPHA VERSION__
+
+This endpoint in its current form is primarily for testing and feedback.
+
+Features in this endpoint are still under heavy development, and subject to change. Not for production use.
+
+See [Release Phases](https://github.com/github/roadmap?tab=readme-ov-file#release-phases) for further details.
+"""
+
```
chouinar, line 23: Since we're not adding this to the docs - we can remove it

api/src/api/users/user_routes.py
```diff
@@ -0,0 +1,49 @@
...
+
+@user_blueprint.post("/token")
+@user_blueprint.input(
+    user_schemas.UserTokenHeaderSchema, location="headers", arg_name="x_oauth_login_gov"
+)
+@user_blueprint.output(user_schemas.UserTokenResponseSchema)
+@user_blueprint.auth_required(api_key_auth)
+def user_token(x_oauth_login_gov: dict) -> response.ApiResponse | Response:
```
chouinar, line 31: Don't need `Response` - that's on the search endpoint due to CSV file stuff it can do.

```suggestion
def user_token(x_oauth_login_gov: dict) -> response.ApiResponse:
```

api/src/api/users/user_routes.py
babebe: done!

mdragon (approved):
Looks good.

chouinar (approved):

PR #2898, opened 2024-11-18
[Issue #2808] Create a JWT for a user

### LINKED ISSUE #2808: Setup logic to generate a JWT in our API
### Summary

TODO - need to detail how we'll do this and what we want to include in the content.

https://pyjwt.readthedocs.io/en/latest/usage.html#encoding-decoding-tokens-with-hs256

Rough notes:
* Define a private/public key as an env var
* Want to include:
   * `sub` (user ID from user table)
   * `exp` / `iat` / `nbf` (token times) https://pyjwt.readthedocs.io/en/latest/usage.html#registered-claim-names
   * `iss` / `aud`

Open questions:
* Who should the issuer/audience be (only the API should ever use these - so presumably something it cares about).
* How long should the expiration be on the jwt?

### Acceptance criteria

- [ ] Given a user object generate an encoded JWT
- [ ] Tests thoroughly written to verify behavior
- [ ] We do not need to connect this to the API yet, a follow-up will handle that

chouinar:
## Summary
Fixes #2808

## Changes proposed
Adds logic to create our own internal JWT for a user

Adds a new user session table for tracking the JWT expiration

Adds the beginning of the logic for parsing that JWT

## Context for reviewers
Generating a JWT is pretty simple, just give the JWT library a key and whatever you want encoded, and it just kinda works.

To follow some general conventions, I'm making a token that has a payload like:
```json
{
    "sub": "abc-123",
    "iat": 1234567890, // unix timestamp - issued at
    "aud": "simpler-grants-api",
    "iss": "simpler-grants-api",
}
```
The `sub` value is the main one and is a "token_id" that we store in the new `user_token_session` table and can later use to fetch the token back out.

I added some rough parsing logic just for the sake of testing, this will expand significantly in the next PR where we actually build that parsing logic back out.

Open question: would it make more sense to have audience be the user_id itself? Logically makes sense, but the parsing on the other side is a bit clunkier.


coilysiren (commented):
defering my review!

mdragon (approved):

PR #2900, opened 2024-11-18
[Issue #2792] Add schemas for extract metadata list API

### LINKED ISSUE #2792: Create input / output schemas for the `ExtractMetadata` API

mikehgrantsgov:
## Summary
Fixes #2792 

## Changes proposed
Add request/response schemas for ExtractMetadata listing API

## Context for reviewers
Creates the input/output schemas needed by the API to support querying extracts and returning data.

## Additional information
See unit tests


api/src/api/extracts_v1/extract_schema.py
```diff
@@ -0,0 +1,74 @@
...
+    extract_metadata_id = fields.Integer(
+        metadata={"description": "The ID of the extract metadata", "example": 1}
+    )
+    extract_type = fields.String(
+        metadata={"description": "The type of extract", "example": "opportunity_data_extract"}
+    )
+    file_name = fields.String(
+        metadata={"description": "The name of the file", "example": "data_extract.csv"}
+    )
+    file_path = fields.String(
```
chouinar, line 53: To follow a pattern we did on the opportunity attachments, I think we'll want to name this different than the DB - something like `download_path`

This way we can pre-sign the URL and adjust the path without risk of modifying the DB column value.

On that note - I wonder if it is worth defining a base `FileResponseSchema` class that we can re-use in different endpoint. Worth checking if there is enough overlap.

api/src/api/extracts_v1/extract_schema.py
```diff
@@ -0,0 +1,74 @@
...
+
+class ExtractMetadataFilterV1Schema(Schema):
+    extract_type = fields.String(
+        allow_none=True,
+        metadata={
+            "description": "The type of extract to filter by",
+            "example": "opportunity_data_extract",
+        },
+        validate=[validators.OneOf([e.value for e in ExtractType])],
+    )
```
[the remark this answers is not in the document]
mikehgrantsgov: Updated this, thanks.

api/src/api/extracts_v1/extract_schema.py
mikehgrantsgov: @chouinar See latest changes and see if that's what you had in mind. Added a placeholder to make a signed S3 URL, but we may want to sign the URL somewhere in the service layer rather than in the schema object.

api/src/api/extracts_v1/extract_schema.py
```diff
@@ -0,0 +1,82 @@
...
+            "description": "When the extract was last updated",
+            "example": "2023-10-02T12:00:00",
+        }
+    )
+
+    def dump(self, obj: Any, **kwargs: Any) -> dict[str, Any]:
+        data = super().dump(obj, **kwargs)
+        # In the future we will update this to use the S3 signed URL
+        data["download_path"] = obj.file_path
+        return data
```
chouinar, line 75: If you wanted it to be like what we'll do in the implementation, just have the test be:

```
extract_metadata = ExtractMetadata(...)
setattr(extract_metadata, "download_path", "http://...")
```

mikehgrantsgov: Got it, ok. I think these latest updates cover it.

chouinar: We wouldn't need the `file_path` field anymore, file path is the internal s3 path, but `download_path` is the external meaningful one.

mikehgrantsgov: Ah yeah, you're right. Removed that. Thanks!

api/src/api/schemas/response_schema.py
```diff
@@ -58,3 +58,11 @@ class ErrorResponseSchema(Schema):
...
         }
     )
+
+
+class FileResponseSchema(Schema):
+    download_path = fields.String(
+        metadata={
+            "description": "The file's download path",
+        },
+    )
```
chouinar, line 68: I don't have a strong opinion on whether we want to do this, but if we're going to setup a base class, we should copy all the common fields, and not need to define them in the two derived classes (download_path, file_size_bytes, created_at, updated_at).

At

api/tests/src/api/extracts_v1/test_extract_schema.py
```diff
@@ -0,0 +1,110 @@
...
+        schema.load(invalid_data)
+
+
+def test_response_schema_single(sample_extract_metadata):
+    schema = ExtractMetadataResponseSchema()
+
+    extract_metadata = schema.dump(sample_extract_metadata)
+
+    # The backend will set this as a S3 signed URL
+    extract_metadata["download_path"] = "http://www.example.com"
```
chouinar, line 59: You would want to set it before, not after running it through the logic. Like how we do for presigning: https://github.com/HHS/simpler-grants-gov/blob/main/api/src/services/opportunities_v1/get_opportunity.py#L60
```suggestion
    
    sample_extract_metadata.download_path = "http://www.example.com"
    extract_metadata = schema.dump(sample_extract_metadata)


    assert extract_metadata["download_path"] == "http://www.example.com"
```

api/src/api/schemas/response_schema.py
mikehgrantsgov: I think it's a good idea as we may use it in the future and it would be duplicated otherwise. I adapted these changes for the common fields and left `OpportunityAttachmentV1Schema` with the fields not relevant to the `ExtractMetadataResponseSchema`

chouinar (approved):
LGTM!

PR #2924, opened 2024-11-19
[Issue #2923] Reduce false positive alert triggering

### LINKED ISSUE #2923: Increase "slow" API response alarm threshold
### Summary

We're alerting when requests take more than 200ms but that's right about the expected response time for an API->OpenSearch request/response cycle. So bumping that up to 500ms as what would represent a problem. The lower threshold leads to alert fatigue and will mask any real performance issues that require investigation.

### Acceptance criteria

- Alarms no longer fire when the time taken is below 500ms.

mdragon:
## Summary
Fixes #2923 

## Changes proposed
Raise the alert threshold for "slow" API requests to something at is closer to the threshold at which we'd start to care how long the request is taking.

## Context for reviewers
We were getting a lot of noise from this alert firing basically every time the site got traffic, this should help avoid the noise but keep the intent of telling us if things are getting slow.

### DIFF
--- a/infra/modules/monitoring/main.tf
+++ b/infra/modules/monitoring/main.tf
@@ -55,7 +55,7 @@ resource "aws_cloudwatch_metric_alarm" "high_app_response_time" {
   namespace           = "AWS/ApplicationELB"
   period              = 60
   statistic           = "Average"
-  threshold           = 0.2
+  threshold           = 0.5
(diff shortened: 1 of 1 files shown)


coilysiren (approved):
:easy-button:

chouinar (approved):

Issue #1884, opened 2024-04-30
Load testing for up to 2,000 concurrent users


acouch:
### Summary

This task is to run a load test with [our artillery setup](https://github.com/HHS/simpler-grants-gov/blob/main/api/artillery-load-test.yml) on the production search api. 

Steps:

* Schedule time to run the load test
* Run artillery locally as an individual or group
* Share results

### Acceptance criteria

- [ ] Run load test and share results
- [ ] Create needed follow up ticket

coilysiren:
I'm going to load test at 3 different levels, of decreasing load:

level 3

```yaml
  phases:
    - duration: 300
      arrivalRate: 10
      maxVusers: 100
      name: Warm up phase
    - duration: 300
      arrivalRate: 500
      maxVusers: 5000
      name: Ramp up load
    - duration: 300
      arrivalRate: 1000
      maxVusers: 10000
      name: Spike phase
```

level 2

```yaml
  phases:
    - duration: 300
      arrivalRate: 10
      maxVusers: 100
      name: Warm up phase
    - duration: 300
      arrivalRate: 250
      maxVusers: 2500
      name: Ramp up load
    - duration: 300
      arrivalRate: 500
      maxVusers: 5000
      name: Spike phase
```

level 1:

```yaml
  phases:
    - duration: 300
      arrivalRate: 10
      maxVusers: 50
      name: Warm up phase
    - duration: 300
      arrivalRate: 100
      maxVusers: 1000
      name: Ramp up load
    - duration: 300
      arrivalRate: 200
      maxVusers: 2000
      name: Spike phase
```

Will post results when I have them, these load tests take a long time to run...

coilysiren:
So in the latest "level 1" test, we start out fine, then 500 a bunch, then finally get back to a good state. I assume what's happening is that:

- the container can handle the warm-up phase easily enough at the resting numbers
- then the resting numbers can't handle the ramp-up case
- but by the time we hit the spike phase the containers have scaled up enough to handle the spike load

So I think what I need to do is... reconfigure the load test parameters so that the warm-up / early ramp-up phase is leaving enough time for the containers to spin up in response to the incoming load. That will probably lead to a more realistic load simulation, I think.

coilysiren:
Report done, its here: https://github.com/HHS/simpler-grants-gov/pull/3120

PR #2940, opened 2024-11-19
[Issue #1884] make artillery very aggressive

coilysiren:
## Summary

Relates to #1884

## Changes proposed

Scales the artillery config up to "the site falls down" numbers, so that we can see exactly what the failure cases are for our website.

Unless we are testing for DDOS protection, we should never need to test artillery with higher numbers than this. These numbers are _**very**_ high.

## Context for reviewers

We tested this already in staging, and it does in fact break the site. Or at least, it does with 4 people running it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -152,3 +152,6 @@ dmypy.json
 
 # Python testing stuff
 *__pycache__*
+
+# Artilery
+frontend/tests/artillery/params.json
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -152,3 +152,6 @@ dmypy.json
 
 # Python testing stuff
 *__pycache__*
+
+# Artilery
+frontend/tests/artillery/params.json
```
doug-s-nava, line 157: should we add this to the frontend gitignore rather than the root?

PR #2963, opened 2024-11-20
[Issue #2793] Create Extract Metadata API

### LINKED ISSUE #2793: Create the `ExtractMetadata` API

mikehgrantsgov:
## Summary
Fixes #2793 

## Changes proposed
Add a new Blueprint / API for the extract metadata API.
Add tests for all business rules and data retrieval

## Context for reviewers
Population of the table will be addressed in a follow-on ticket.

## Additional information
Open questions:
Should this be called V1 to follow the opportunity API convention?
`extract_type` is a filter in the POST param vs an explicit path param in the URL. Is that OK?

See unit tests for additional context.


api/src/api/extracts_v1/extract_routes.py
```diff
@@ -0,0 +1,55 @@
...
+def extract_metadata_get(
+    db_session: db.Session, raw_list_params: dict
+) -> response.ApiResponse | Response:
+    # Default to last 7 days if no date range is provided
+    if not raw_list_params.get("filters", {}).get("created_on"):
+        raw_list_params.setdefault("filters", {})
+        raw_list_params["filters"]["created_on"] = {
+            "start_date": (datetime.now() - timedelta(days=7)).date(),
+            "end_date": datetime.now().date(),
+        }
```
chouinar, line 46: Be careful with `datetime.now()` without a timezone. I'd recommend we try to either explicitly do UTC or EST. We have helper methods in datetime_util for some of that.

--

I realize we can't actually set defaults for these as those would be statically defined when the app starts up. In that case, you could put the defaults on the Pydantic models you created which avoids needing to set dictionaries here.

Might be able to do something roughly like:
```py
class ExtractFilters(BaseModel):
    extract_type: ExtractType | None = None
    created_at: DateSearchFilter | None = Field(default_factory=datetime_util.get_now_us_eastern_date())

```

api/src/api/extracts_v1/extract_routes.py
```diff
@@ -0,0 +1,55 @@
...
+
+examples = {
+    "example1": {
+        "summary": "No filters",
+        "value": {
+            "pagination": {
+                "order_by": "created_at",
+                "page_offset": 1,
+                "page_size": 25,
+                "sort_direction": "ascending",
```
chouinar, line 21: It is just an example, but would descending be better here so it is "latest first" - what I guess most people would care about.

api/src/api/extracts_v1/extract_routes.py
```diff
@@ -0,0 +1,55 @@
...
+    extract_schema.ExtractMetadataRequestSchema,
+    arg_name="raw_list_params",
+    examples=examples,
+)
+@extract_blueprint.output(extract_schema.ExtractMetadataListResponseSchema)
+@extract_blueprint.auth_required(api_key_auth)
+@flask_db.with_db_session()
+def extract_metadata_get(
+    db_session: db.Session, raw_list_params: dict
+) -> response.ApiResponse | Response:
```
chouinar, line 39: Should be able to do just `response.ApiResponse` - the search endpoint has `Response` as well because it can return a csv file

api/tests/src/api/extracts_v1/test_extracts_routes.py
```diff
@@ -0,0 +1,127 @@
...
+    yield
+
+
+def test_extract_metadata_get_default_dates(
+    client, api_auth_token, enable_factory_create, db_session
+):
+    """Test that default date range (last 7 days) is applied when no dates provided"""
+    ExtractMetadataFactory.create_batch(
+        2,
+        extract_type=ExtractType.OPPORTUNITIES_XML,
```
chouinar, line 23: Not specifically for this PR, but I missed it when you added extract type - we want JSON, not XML for this file type.

With how the lookup values work, it should be safe to just change this (DB migrations will resync the value to the DB - and nothing is using it in the API yet).

api/src/services/extracts_v1/get_extracts.py
```diff
@@ -0,0 +1,54 @@
...
+                    ExtractMetadata.created_at >= list_params.filters.created_at.start_date
+                )
+            if list_params.filters.created_at.end_date:
+                stmt = stmt.where(
+                    ExtractMetadata.created_at <= list_params.filters.created_at.end_date
+                )
+
+    # Apply pagination
+    offset = list_params.pagination.page_size * (list_params.pagination.page_offset - 1)
+    stmt = stmt.offset(offset).limit(list_params.pagination.page_size)
```
chouinar, line 46: Should we use the paginator for this? You can see it in the v0.1 version of the opportunity endpoint: https://github.com/HHS/simpler-grants-gov/blob/main/api/src/services/opportunities_v0_1/search_opportunities.py#L260

https://github.com/HHS/simpler-grants-gov/blob/main/api/src/pagination/paginator.py

api/tests/src/api/extracts_v1/test_extracts_routes.py
mikehgrantsgov: Yep makes sense! I was just making this change when adding the db record when the task runs.

api/src/api/extracts_v1/extract_routes.py
mikehgrantsgov: I think UTC would be better in this case? I made an update here.

api/src/api/extracts_v1/extract_routes.py
mikehgrantsgov: Makes sense!

api/src/services/extracts_v1/get_extracts.py
mikehgrantsgov: Good ida, added this

api/src/api/extracts_v1/extract_routes.py
```diff
@@ -0,0 +1,55 @@
...
+
+    add_extra_data_to_current_request_logs(
+        {
+            "response.pagination.total_pages": pagination_info.total_pages,
+            "response.pagination.total_records": pagination_info.total_records,
+        }
+    )
+    logger.info("Successfully fetched extracts")
+
+    print(pagination_info)
```
chouinar, line 53:
```suggestion

```

api/src/services/extracts_v1/get_extracts.py
```diff
@@ -0,0 +1,69 @@
...
+                stmt = stmt.where(
+                    ExtractMetadata.created_at >= list_params.filters.created_at.start_date
+                )
+            if list_params.filters.created_at.end_date:
+                stmt = stmt.where(
+                    ExtractMetadata.created_at <= list_params.filters.created_at.end_date
+                )
+
+    # Apply pagination
+    print(list_params.pagination)
```
chouinar, line 57:
```suggestion

```

api/tests/src/api/extracts_v1/test_extract_schema.py
```diff
@@ -106,5 +111,6 @@ def test_request_schema_null_values():
 
     result = schema.load(data)
     assert result["filters"]["extract_type"] is None
-    assert result["filters"]["start_date"] == date(2023, 10, 1)
-    assert result["filters"]["end_date"] is None
+    assert result["filters"]["created_at"]["start_date"] == date(2023, 10, 1)
+    print(result["filters"]["created_at"]["end_date"])
```
chouinar, line 115:
```suggestion

```

api/tests/src/api/extracts_v1/test_extracts_routes.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+def test_extract_metadata_get_default_dates(
+    client, api_auth_token, enable_factory_create, db_session
+):
+    """Test that default date range (last 7 days) is applied when no dates provided"""
+    ExtractMetadataFactory.create_batch(
+        2,
+        extract_type=ExtractType.OPPORTUNITIES_JSON,
+    )
```
chouinar, line 25: Could you add a record that is older/newer than the last 7 days?

api/tests/src/api/extracts_v1/test_extracts_routes.py
chouinar, line 1: Could you add at least one test that verifies the file can be downloaded / the pre-signing works.

Can probably just steal some of what https://github.com/HHS/simpler-grants-gov/blob/main/api/tests/src/api/opportunities_v1/test_opportunity_route_get.py#L111 does which is testing the presigned logic for GET opportunities.

mikehgrantsgov: Sure, I just added this. Used some adaptation from that example you sent, thanks.

chouinar (approved):
LGTM!

PR #2977, opened 2024-11-21
[Unticketed] Enable ELT process in dev

chouinar:
## Changes proposed
Turns out the ELT process in dev was running while doing nothing - noticed when looking at some data there that nothing had updated recently (since I last manually ran in dev).


chouinar:
This is another good reason to have New Relic - I'd have a nice dashboard that shows records processed per run (at a lot of different levels) and would have seen this weeks ago.

coilysiren (approved):

PR #3001, opened 2024-11-22
[Issue #2819] disable ga in test

### LINKED ISSUE #2819: Local testing logging to Prod GA Property
### Summary

Running tests either locally on Dev laptops or via CI/CD should not be picking up the Prod GA Property ID. Instead it should be using the test or Dev one.

### Acceptance criteria

We no longer see GA data in Prod with hostnames of 127.0.0.1 or localhost
Possibly related: Image

doug-s-nava:
## Summary
Fixes #2819

## Changes proposed
This reorganizes how the application will grab Google Analytics IDs in difference scenarios in order to solve issues where non-production analytics data was making it into the production GA account. I laid out the scenarios in a comment on the ticket here https://github.com/HHS/simpler-grants-gov/issues/2819#issuecomment-2494258673.

The change basically consists of removing GA IDs from Next code or .env files, and moving it to SSM / terraform.

This also adds some documentation, and removes an unused environment variable, both of which should hopefully reduce confusion and help to reduce the chance of this kind of problem down the line.

### Test steps

4. run `npm run dev` and click around the app
5. run `npm run build && npm start` and click around the app
6. _VERIFY_: using omnibug or similar tool, verify that no requests to GA are going out

Beyond this it is difficult to test without merging the change and checking GA after 24 hours, but that can be done if folks think it's necessary.

1. merge the change
2. wait 24 hours
3. _VERIFY_: GA prod account no longer shows any traffic coming from localhost


mdragon (approved):
This was worse than I expected but I like where you arrived at.

Issue #2817, opened 2024-11-12
Create a /users/token/refresh endpoint


chouinar:
### Summary

Create a new endpoint that refreshes a token.

This endpoint should use the new `api_jwt_auth` from https://github.com/HHS/simpler-grants-gov/pull/2959

The only thing this endpoint needs to do is a simple update on the `expires_at` timestamp in the user token session. This should use the same/similar logic to how we set the expires_at in the api_jwt_auth.py code.

You can fetch the current user token session by doing `api_jwt_auth.current_user`

NOTE: It will seem a bit odd, but you must do `db_session.add(user_token_session)` in the endpoint in order to actually modify it. This has to do with some complex under-the-hood details on how the DB sessions work (in short - the DB session the API uses and that auth uses is different so we need to make the API one aware of it).

**Request**
There is no request body for this endpoint, the auth token itself is handled by connecting auth to the endpoint

**Response**
The response is also very simple, we just want a very basic Response Schema without a relevant data object - see what the Healthcheck endpoint does. We really just want to return a message like `Token refreshed` when successful.

**Note on testing**
For unit tests, you can make a valid token by doing something like:
```py
    user = UserFactory.create()
    token, _ = create_jwt_for_user(user, db_session)
    db_session.commit()  # need to commit here to push the session to the DB

    # then freely use the token to call the endpoint
```

### Acceptance criteria

- [ ] Endpoint created
- [ ] Token expires_at updated and validated in tests
- [ ] Basic tests (endpoint gives auth errors if token already expired / isn't valid - which is implemented in the auth logic itself)

PR #3002, opened 2024-11-22
[Issue 2817] Create a /users/token/refresh endpoint

babebe:
## Summary
Fixes #{[2817](https://github.com/HHS/simpler-grants-gov/issues/2817)}

## Changes proposed
Added  POST endpoint `token/refresh`
Added `UserTokenRefreshResponseSchema`
Added reusable func `refresh_token_expiration`
Added 2 tests. For success case  and case when token was expired


api/src/api/users/user_routes.py
```diff
@@ -34,3 +41,30 @@ def user_token(x_oauth_login_gov: dict) -> response.ApiResponse:
...
+@user_blueprint.output(UserTokenRefreshResponseSchema)
+@user_blueprint.doc(responses=[200, 401])
+@user_blueprint.auth_required(api_jwt_auth)
+@flask_db.with_db_session()
+def user_token_refresh(db_session: db.Session) -> response.ApiResponse:
+    logger.info("POST /v1/users/token/refresh")
+
+    user_token_session: UserTokenSession = api_jwt_auth.current_user  # type: ignore
+    config = get_config()
+    expiration_time = datetime_util.utcnow() + timedelta(minutes=config.token_expiration_minutes)
```
chouinar, line 56: Put this in the api jwt auth file, and follow the pattern it has for getting the config (just want to keep that located all in one place as it's a bit weird).

api/tests/src/api/users/test_user_route_token.py
```diff
@@ -25,3 +29,34 @@ def test_post_user_route_token_400(client, api_auth_token):
...
+    token, user_token_session = create_jwt_for_user(user, db_session)
+    expiration = user_token_session.expires_at
+    db_session.commit()
+
+    resp = client.post("v1/users/token/refresh", headers={"X-SGG-Token": token})
+
+    db_session.refresh(user_token_session)
+
+    assert resp.status_code == 200
+    assert user_token_session.expires_at != expiration
```
chouinar, line 47: A trick for testing time related things is to use the freeze_time library (see an example here: https://github.com/HHS/simpler-grants-gov/blob/main/api/tests/src/auth/test_api_jwt_auth.py#L160 )

You can wrap a bit of code like a call to the API (or the [whole test](https://github.com/HHS/simpler-grants-gov/blob/main/api/tests/src/auth/test_api_jwt_auth.py#L57)) in a specific time, and then know that the expires_at will always be that time + 30.

api/src/auth/api_jwt_auth.py
```diff
@@ -205,3 +206,12 @@ def example_method(db_session: db.Session) -> response.ApiResponse:
         # The message is just the value we set when constructing the JwtValidationError
         logger.info("JWT Authentication Failed for provided token", extra={"auth.issue": e.message})
         raise_flask_error(401, e.message)
+
+
+def set_token_expiration_time(config: ApiJwtConfig | None = None) -> datetime:
```
chouinar, line 211: I'd just make this method take in the user token session and set it in here.

I might also suggest naming it something like `refresh_token_expiration`.

mdragon (approved):

PR #3004, opened 2024-11-22
[Issue #2810] Connect all the components of the /users/token endpoint together

### LINKED ISSUE #2810: Fully connect the /users/token endpoint component pieces
### Summary

We want to connect all of the pieces of the implementation together in this ticket. That is:

* Using the endpoint written in #2673 
* It should take the token it received, and pass it to the logic created in #2721 
* The ID from this should be used to query the users table (by external_user_id) and create a user if doesn't exist
* We should then generate a JWT from this user using the logic in #2808 

### Acceptance criteria

- [ ] Thorough API-wide unit tests written that hit many scenarios (happy path, and error)

chouinar:
DRAFT: Needs a bit more testing/cleanup, but I'll be out next week, so getting any feedback folks might have up front.

TODO for when I get back:
* Rough testing with the local mock - I got it talking to it, but haven't actually made it work
* Cleanup of how the response gets built
* Logging/metrics stuff 

## Summary
Fixes #2810

## Changes proposed
Connected all the components of the /users/token endpoint that had been built out separately:
* Process token from login.gov
* Create/update user in our DB
* Create a JWT + session and return it

## Context for reviewers
TODO

## Additional information
TODO


api/tests/conftest.py
```diff
@@ -217,6 +220,60 @@ def opportunity_index_alias(search_client, monkeypatch_session):
...
+    return _generate_rsa_key_pair()
+
+
+@pytest.fixture(scope="session")
+def setup_login_gov_auth(monkeypatch_session, public_rsa_key):
+    # TODO - describe
+    def override_method(config):
+        config.public_keys = [public_rsa_key]
+
+    monkeypatch_session.setattr(login_gov_jwt_auth, "_refresh_keys", override_method)
```
chouinar, line 271: Note for self - what if I instead made a dummy endpoint that returned the keys? That would at least validate the parsing logic

chouinar:
Redid how this works in another PR, closing

PR #3027, opened 2024-11-26
[Issue #1884] 4x the max number of potential containers

coilysiren:
## Summary

Realtes to https://github.com/HHS/simpler-grants-gov/issues/1884

## Changes proposed

Throws even more containers at our frontend performance problem

## Context for reviewers

I tested "level 1" of https://github.com/HHS/simpler-grants-gov/issues/1884 with the following configuration

```yaml
  phases:
    - arrivalRate: 10
      maxVusers: 50
      name: Warm up phase
    - arrivalRate: 100
      maxVusers: 1000
      name: Ramp up load
    - arrivalRate: 200
      maxVusers: 2000
      name: Spike phase
```

This is a low amount of spike load (IMO) and it caused -alot- of 500 errors. But once the containers has scaled up sufficiently, the 500 errors died down. So I'm yet again adjusting the scaling, in hopes that these new numbers don't "fail" the level 1 load test again.

### DIFF
--- a/frontend/.gitignore
+++ b/frontend/.gitignore
@@ -55,3 +55,6 @@ npm-debug.log*
 # pa11y
 screenshots-output/*
 pa11y_output.txt
+
+# Artilery
+tests/artillery/params.json
--- a/infra/frontend/app-config/prod.tf
+++ b/infra/frontend/app-config/prod.tf
@@ -15,7 +15,7 @@ module "prod_config" {
   instance_desired_instance_count = 4
   instance_scaling_min_capacity   = 4
   # instance_scaling_max_capacity is 5x the instance_scaling_min_capacity
-  instance_scaling_max_capacity = 20
+  instance_scaling_max_capacity = 80
 
   instance_cpu    = 1024
   instance_memory = 2048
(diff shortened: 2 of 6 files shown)


infra/frontend/app-config/prod.tf
```diff
@@ -15,7 +15,7 @@ module "prod_config" {
   instance_desired_instance_count = 4
   instance_scaling_min_capacity   = 4
   # instance_scaling_max_capacity is 5x the instance_scaling_min_capacity
-  instance_scaling_max_capacity = 20
+  instance_scaling_max_capacity = 80
```
coilysiren, line 18: I need to do calcuations for how much this would cost. In the experience the load testing lasts about an hour, so I need to do the math for the cost of running ... 80 ... containers for an hour.

infra/modules/service/autoscaling.tf
```diff
@@ -24,7 +24,7 @@ resource "aws_appautoscaling_policy" "ecs_scale_policy_memory" {
       predefined_metric_type = "ECSServiceAverageMemoryUtilization"
     }
     scale_in_cooldown  = 300
-    scale_out_cooldown = 5
+    scale_out_cooldown = 30
```
coilysiren, line 27: I'm increasing the scale out numbers so we doesn't reach for 80 containers immediately 😆

frontend/.gitignore
```diff
@@ -55,3 +55,6 @@ npm-debug.log*
 # pa11y
 screenshots-output/*
 pa11y_output.txt
+
+# Artilery
+tests/artillery/params.json
```
coilysiren, line 60: https://github.com/HHS/simpler-grants-gov/pull/2940#discussion_r1852244950

PR #3040, opened 2024-11-26
[no ticket] fix cron vuln scan logic (hopefully)

coilysiren:
## Changes proposed

Changes the vuln scan cronjob to both:

- have an explicit `needs` keyword
- always run every vuln scan on every app

## Context for reviewers

Right now the setup for the cron vuln scans is nonideal, this is shown here:

image

There are two problems here:

- the matrix fails fast, which means that if one scan fails, the other scans won't get a chance to cache
- there is a `needs` statement missing on the slack job, so it doesn't wait for the vuln scans to complete before it alerts slack. In practice this means that it will never alert slack (and in my experience, it never has)

### DIFF
--- a/.github/workflows/ci-cron-vulnerability-scans.yml
+++ b/.github/workflows/ci-cron-vulnerability-scans.yml
@@ -22,4 +22,5 @@ jobs:
 
   send-slack-notification:
     if: failure()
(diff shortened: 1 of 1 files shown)


mdragon (approved):
LGTM!

Issue #2676, opened 2024-10-30
Create a GET /users/:userID endpoint


chouinar:
### Summary

We want to add a simple endpoint:

* `/users/:userID`

This endpoint should have authentication configured to use what we created in #2809 

The endpoint should validate that the user ID matches the user ID being queried and raise a 401 Unauthorized if not
* This is very very rough AuthZ - a more thorough approach will come later

The endpoint should return a simple user object (same as the user object in #2673 )

### Acceptance criteria

- [ ] Endpoint returns a 200 if the auth is successful
- [ ] Endpoint returns a 401 (use `raise_flask_error` in the auth logic) if not
- [ ] Scenario: If the user ID attached to the token doesn't match (and that user ID doesn't exist) it should still error - you don't necessarily need to account for this as the check to see if `user_token_session.user_id == request.user_id` can happen before any DB calls.

PR #3041, opened 2024-11-26
[Issue 2676] Create a GET /users/:userID endpoint

babebe:
## Summary
Fixes #{[2676](https://github.com/HHS/simpler-grants-gov/issues/2676)}

## Changes proposed
Created GET `v1/users/:user_id` endpoint ( expected user_id type UUID)
Created `UserGetResponseSchema` 
Created `get_user` service method to fetch user object 
Created 2 tests: 
1. Test correct user object returned 
2. Test 401 error when auth user request for another user_id 

## Context for reviewers
This endpoint should have authentication configured 
The endpoint should validate that the user ID matches the user ID being queried and raise a 401 Unauthorized if not


api/tests/src/db/models/factories.py
```diff
@@ -1858,7 +1858,7 @@ class Meta:
     user = factory.SubFactory(UserFactory)
     user_id = factory.LazyAttribute(lambda s: s.user.user_id)
 
-    external_user_type_id = factory.fuzzy.FuzzyChoice(ExternalUserType)
+    external_user_type = factory.fuzzy.FuzzyChoice(ExternalUserType)
```
mikehgrantsgov, line 1861: Was this required as part of the PR? Do tests fail if this is `external_user_type_id`

babebe: yup they do fail. This fixes it.

mikehgrantsgov (approved):
Looks good!

PR #3044, opened 2024-11-26
[no ticket] Use higher ASG targets

coilysiren:
## Context

At 50% memory, the frontend hits that memory cap easily and immediately gets cycled out. I wasn't really expecting this behavior, but you live and you learn.

## Changes proposed

Increases the memory target to 99%, so nextjs only spins up new containers when after containers max out their cache. This won't meaningfully impact the backend, because the because the backend doesn't have an infinitely* scaling cache.

\* There's not significant indication that the nextjs cache is infinitely scaling. It might just scale higher than our current memory numbers.

I also increased the CPU target to 75%, which may come back to bite me I dunno. We'll see.


mdragon (approved):
👍🏻

PR #3059, opened 2024-11-26
[no ticket] Update libc dev

coilysiren:
## Changes proposed

Adds `libc-dev` to the list of packages we update explicitly

## Context for reviewers

I'm hoping this fixes the analytics deploys

### DIFF
--- a/analytics/Dockerfile
+++ b/analytics/Dockerfile
@@ -11,15 +11,24 @@ FROM python:3.13-slim AS base
 RUN pip install --no-cache-dir poetry==1.8.2 --upgrade
 
 RUN apt-get update \
+  # Remove existing packages before installing their never versions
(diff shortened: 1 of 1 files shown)


analytics/Dockerfile
```diff
@@ -11,15 +11,24 @@ FROM python:3.13-slim AS base
 RUN pip install --no-cache-dir poetry==1.8.2 --upgrade
 
 RUN apt-get update \
+  # Remove existing packages before installing their never versions
+  && apt-get remove --yes \
```
coilysiren, line 15: This pattern (removing first) was necessary for postgres recently, so now I want to do it everywhere

DavidDudas-Intuitial (approved):

PR #3065, opened 2024-11-27
[Issue #3057] Create Agencies API

### LINKED ISSUE #3057: Add API endpoint for list of agencies
### Summary

To ensure we hit our Quad 1 AC of the search API having feature parity with the search UI, we need an API endpoint that returns the list of agencies in the database.

### Acceptance criteria

- [ ] Endpoint is live in PROD and returns available agencies

mikehgrantsgov:
## Summary
Fixes #3057

## Changes proposed
Add agencies API supporting pagination and basic filtering (future requirements around shape TBD)

## Context for reviewers
Open question: What should the schema return? Right now we are returning all Agency model fields. Maybe we should make this payload smaller depending on the frontend needs.

## Additional information
See attached unit test


api/tests/src/api/agencies_v1/test_agencies_routes.py
```diff
@@ -0,0 +1,34 @@
+import pytest
+
+from src.db.models.agency_models import Agency, LinkAgencyDownloadFileType
+from tests.src.db.models.factories import AgencyFactory
+
+
+@pytest.fixture(autouse=True)
+def cleanup_agencies(db_session):
```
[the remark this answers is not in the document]
mikehgrantsgov: Thanks I refactored this into a class

api/tests/src/db/models/factories.py
```diff
@@ -721,9 +721,21 @@ class AgencyFactory(BaseFactory):
...
+                return value + 1
+
+        return 1
+
+    agency_id = factory.Sequence(lambda n: n)
     agency_name = factory.Faker("agency_name")
 
-    agency_code = factory.Faker("agency_code")
+    # Make agency_code unique by including the sequence number
+    agency_code = factory.Sequence(lambda n: f"AG-{n:04d}")
```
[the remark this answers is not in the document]
chouinar: https://github.com/HHS/simpler-grants-gov/pull/3069

mdragon (approved):
Looks good, think we want to restrict a few of the Agency fields as more internal use only.

api/src/api/agencies_v1/agency_schema.py
```diff
@@ -0,0 +1,77 @@
...
+    is_test_agency = fields.Boolean()
+    is_multilevel_agency = fields.Boolean()
+    is_multiproject = fields.Boolean()
+    has_system_to_system_certificate = fields.Boolean()
+    can_view_packages_in_grace_period = fields.Boolean()
+    is_image_workspace_enabled = fields.Boolean()
+    is_validation_workspace_enabled = fields.Boolean()
+
+    # Optional fields
+    ldap_group = fields.String(allow_none=True)
```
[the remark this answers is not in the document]
mikehgrantsgov: Agreed. @chouinar do we have conventions to have the requestor specify fields it needs? If not I can restrict the response for now to fields only needed by the FE.

api/src/api/agencies_v1/agency_schema.py
```diff
@@ -0,0 +1,77 @@
...
+
+    contact_name = fields.String()
+    address_line_1 = fields.String()
+    address_line_2 = fields.String(allow_none=True)
+    city = fields.String()
+    state = fields.String()
+    zip_code = fields.String()
+    phone_number = fields.String()
+    primary_email = fields.String()
+    secondary_email = fields.String(allow_none=True)
```
chouinar, line 32: Could you add some descriptions to these?

api/src/api/agencies_v1/agency_schema.py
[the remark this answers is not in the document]
chouinar: On the agency table there are several fields I marked as "unclear usage" that you can exclude. It is fine if the Marshmallow model has less fields than the DB, they'll just get discarded.

We can always add stuff later.

api/src/api/agencies_v1/agency_schema.py
```diff
@@ -0,0 +1,77 @@
...
+    agency_code = fields.String()
+    sub_agency_code = fields.String(allow_none=True)
+    assistance_listing_number = fields.String()
+    agency_submission_notification_setting = fields.String()  # Enum value
+
+    # Agency contact info as nested object
+    agency_contact_info = fields.Nested(AgencyContactInfoSchema, allow_none=True)
+
+    # Boolean flags
+    is_test_agency = fields.Boolean()
```
chouinar, line 49: We probably want to filter out any test agencies (like how we filter draft opportunities), the legacy system does that to some degree. The agency transform job already handles setting this flag for known test agencies

api/src/api/agencies_v1/agency_schema.py
```diff
@@ -0,0 +1,77 @@
...
+    # Optional fields
+    ldap_group = fields.String(allow_none=True)
+    description = fields.String(allow_none=True)
+    label = fields.String(allow_none=True)
+
+    # Related agency info
+    top_level_agency_id = fields.Integer(allow_none=True)
+
+    # File types as a list of strings
+    agency_download_file_types = fields.List(fields.String())
```
chouinar, line 66: There is an enum for this

api/src/api/agencies_v1/agency_schema.py
```diff
@@ -0,0 +1,77 @@
...
+    is_image_workspace_enabled = fields.Boolean()
+    is_validation_workspace_enabled = fields.Boolean()
+
+    # Optional fields
+    ldap_group = fields.String(allow_none=True)
+    description = fields.String(allow_none=True)
+    label = fields.String(allow_none=True)
+
+    # Related agency info
+    top_level_agency_id = fields.Integer(allow_none=True)
```
chouinar, line 63: We can exclude the ID, although we will need this API to return the top level agency info for the frontend to build a hierarchy.

api/src/api/agencies_v1/agency_schema.py
mikehgrantsgov: Added!

api/src/api/agencies_v1/agency_schema.py
mikehgrantsgov: Added a filter to this and removed from the schema since this will always be `false`

api/src/api/agencies_v1/agency_schema.py
mikehgrantsgov: Thanks, using this now.

api/src/api/agencies_v1/agency_schema.py
mikehgrantsgov: I removed this but sounds like it should stay, though?

api/src/api/agencies_v1/agency_schema.py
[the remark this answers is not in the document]
chouinar: I missed the other question you originally had, there isn't any way to change the fields in the response based on what the requestor asks for, at least not right now. Schemas are strictly defined, at best we could null out a bunch of fields for that

api/src/api/agencies_v1/agency_schema.py
chouinar: We don't need the ID since that is purely internal, but the frontend would want the top level agency itself - which might be a ticket itself as that requires nesting the schema in itself.

You should be able to do that https://marshmallow.readthedocs.io/en/stable/nesting.html#nesting-a-schema-within-itself

Something like:
```
class AgencySchema(Schema):
     ...
    top_level_agency = fields.Nested(lambda: AgencySchema(exclude=("top_level_agency",)))
```
That'll nest it within itself, but only go one layer deep since you will have excluded nesting in the layer down.

mdragon (approved):
Looks good.

chouinar (approved):

PR #3068, opened 2024-12-02
[Issue #2958] Limit draft_documents s3 to just api

### LINKED ISSUE #2958: Add a condition so you don't deploy a drafts s3 bucket for every app
### Summary

//

### Acceptance criteria

//

mdragon:
## Summary
Fixes #2958

## Changes proposed
Only create a S3 bucket for draft documents in the API service, not in other services.

## Context for reviewers
When setup initially the new bucket for drafts was created as desired, but then it was noticed that it was created for all services not just the one that was intended.

### DIFF
--- a/infra/analytics/app-config/env-config/outputs.tf
+++ b/infra/analytics/app-config/env-config/outputs.tf
@@ -28,6 +28,9 @@ output "scheduled_jobs" {
   value = local.scheduled_jobs
 }
 
+output "enable_drafts_bucket" {
+  value = startswith(var.app_name, "api-")
+}
 output "domain" {
   value = var.domain
 }
--- a/infra/analytics/app-config/env-config/outputs.tf
+++ b/infra/analytics/app-config/env-config/outputs.tf
@@ -28,9 +28,6 @@ output "scheduled_jobs" {
   value = local.scheduled_jobs
 }
 
-output "enable_drafts_bucket" {
-  value = startswith(var.app_name, "api-")
-}
 output "domain" {
   value = var.domain
 }
--- a/infra/analytics/metabase/main.tf
+++ b/infra/analytics/metabase/main.tf
@@ -124,7 +124,5 @@ module "service" {
     }
   }
 
-  enable_drafts_bucket = startswith(local.service_name, "api-")
-
   is_temporary = false
 }
--- a/infra/analytics/metabase/main.tf
+++ b/infra/analytics/metabase/main.tf
@@ -123,5 +123,8 @@ module "service" {
       schema_name = local.database_config.schema_name
     }
   }
+
+  enable_drafts_bucket = startswith(local.service_name, "api-")
+
   is_temporary = false
 }
--- a/infra/api/app-config/env-config/outputs.tf
+++ b/infra/api/app-config/env-config/outputs.tf
@@ -47,10 +47,6 @@ output "scheduled_jobs" {
   value = local.scheduled_jobs
 }
 
-output "enable_drafts_bucket" {
-  value = startswith(var.app_name, "api-")
-}
-
 output "incident_management_service_integration" {
   value = var.has_incident_management_service ? {
     integration_url_param_name = "/monitoring/${var.app_name}/${var.environment}/incident-management-integration-url"
--- a/infra/api/app-config/env-config/outputs.tf
+++ b/infra/api/app-config/env-config/outputs.tf
@@ -47,6 +47,10 @@ output "scheduled_jobs" {
   value = local.scheduled_jobs
 }
 
+output "enable_drafts_bucket" {
+  value = startswith(var.app_name, "api-")
+}
+
 output "incident_management_service_integration" {
   value = var.has_incident_management_service ? {
     integration_url_param_name = "/monitoring/${var.app_name}/${var.environment}/incident-management-integration-url"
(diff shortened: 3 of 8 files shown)


infra/modules/service/draft_documents.tf
```diff
@@ -1,4 +1,5 @@
 resource "aws_s3_bucket" "draft_documents" {
+  count = var.service_name == "api" ? 1 : 0
```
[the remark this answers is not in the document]
coilysiren: Starts with `api-`, just to be safe

mdragon:
Addressed the original issues in subsequent commits

infra/modules/service/draft_documents.tf
```diff
@@ -1,4 +1,5 @@
 resource "aws_s3_bucket" "draft_documents" {
+  count         = startswith(var.service_name, "api-") ? 1 : 0
```
coilysiren, line 2: I thought about it overnight, and...

I think we should swap this to a `enable_drafts_bucket` parameter on the module

sorry for the run around >_>

infra/modules/service/draft_documents.tf
```diff
@@ -9,7 +10,8 @@ resource "aws_s3_bucket" "draft_documents" {
 }
 
 resource "aws_s3_bucket_public_access_block" "draft_documents" {
-  bucket = aws_s3_bucket.draft_documents.id
+  count  = startswith(var.service_name, "api-") ? 1 : 0
+  bucket = one(aws_s3_bucket.draft_documents[*].id)
```
coilysiren, line 14: While this does work, the usual pattern here is

```suggestion
  bucket = aws_s3_bucket.draft_documents[0].id
```

and consistency is nice!

infra/modules/service/main.tf
```diff
@@ -29,7 +29,7 @@ locals {
   environment_variables = concat(
     local.base_environment_variables,
     local.db_environment_variables,
-    [{ name : "DRAFTS_S3_BUCKET_ARN", value : aws_s3_bucket.draft_documents.arn }],
+    [{ name : "DRAFTS_S3_BUCKET_ARN", value : one(aws_s3_bucket.draft_documents[*].arn) }],
```
coilysiren, line 32:
```suggestion
    var.enable_drafts_bucket != null ? [{ name : "DRAFTS_S3_BUCKET_ARN", value : aws_s3_bucket.draft_documents[0].arn }] : [],
```

^ needs to be a ternary that resolves to empty with the input condition is null

coilysiren: Actually, flip this ^ that one around...

```suggestion
    var.enable_drafts_bucket == null ? [] : [{ name : "DRAFTS_S3_BUCKET_ARN", value : aws_s3_bucket.draft_documents[0].arn }],
```

infra/modules/service/draft_documents.tf
mdragon: I had seen some folks saying those [0].blah would fail if there was nothing in the set? But I guess if it works elsewhere and we have stuff turned off then maybe that is handled better now.

infra/modules/service/draft_documents.tf
mdragon: I'm pretty sure this isn't quite right, but it validates... 
https://github.com/HHS/simpler-grants-gov/pull/3068/commits/09a643a4fe0fdb5923d0ff41cfa732c8bbc0d6ad

infra/analytics/metabase/main.tf
```diff
@@ -123,5 +123,8 @@ module "service" {
       schema_name = local.database_config.schema_name
     }
   }
+
+  enable_drafts_bucket = startswith(local.service_name, "api-")
```
coilysiren, line 127: Simpler: set the default value to `false` and only set it to true inside of `infra/api/service/main.tf`

infra/api/app-config/env-config/outputs.tf
```diff
@@ -47,6 +47,10 @@ output "scheduled_jobs" {
   value = local.scheduled_jobs
 }
 
+output "enable_drafts_bucket" {
+  value = startswith(var.app_name, "api-")
+}
```
coilysiren, line 52: This isn't necessary

infra/analytics/app-config/env-config/outputs.tf
```diff
@@ -28,6 +28,9 @@ output "scheduled_jobs" {
   value = local.scheduled_jobs
 }
 
+output "enable_drafts_bucket" {
+  value = startswith(var.app_name, "api-")
+}
```
coilysiren, line 33: Delete this too

infra/api/service/outputs.tf
```diff
@@ -22,3 +22,7 @@ output "application_log_stream_prefix" {
 output "migrator_role_arn" {
   value = module.service.migrator_role_arn
 }
+
+output "enable_drafts_bucket" {
+  value = module.service.enable_drafts_bucket
+}
```
coilysiren, line 28: Also delete this one

coilysiren (commented):
yeah!!! (except now you must appease checkov with our traditions `skip`s)

coilysiren:
merge conflict: https://github.com/HHS/simpler-grants-gov/pull/3165

coilysiren (approved):
🚀

PR #3069, opened 2024-12-02
[Untickted] Add more agencies to the local seed dataset

chouinar:
## Changes proposed
Adds more agencies to the factories as possible values + generates more agencies into our local DB at startup

## Context for reviewers
We only had 4 possible agencies that could be generated before, I added another set based on real values (pulled a bunch that were over 100 total opportunities).

This gives us a bit more to work with locally in tests as before we only would generate 4 unique agencies.

## Additional information
Screenshot 2024-12-02 at 12 56 59 PM


mikehgrantsgov (approved):
Looks good!

Issue #680, opened 2023-11-07
Implement CloudFront CDN for Static Site


acouch:
### Summary

We are moving forward with CloudFront as a CDN tool. This work is set up CloudFront for the static site.

### Acceptance criteria

- [ ] CloudFront CDN added to static site
- [ ] Update team documentation

lorenyu:
@acouch this is [on the backlog for the infra template too](https://github.com/navapbc/template-infra/issues/459). have you thought about the technical design for this? i was thinking that we could do something like:

- add a new terraform module `infra/modules/cdn`
- call the new cdn module from the `infra/frontend/service` module

the infra template version would look similar but would be more customizable:

- add a `has_cdn` to [app-config](https://github.com/HHS/simpler-grants-gov/blob/main/infra/frontend/app-config/main.tf)
- add a new terraform module `infra/modules/cdn`
- call the new cdn module from the `infra/frontend/service` module if `app_config.has_cdn` is true

thoughts?

mxk0:
To be broken down into implementation tickets during 12/11 refinement.

PR #3082, opened 2024-12-03
[Issue #680] Add basic CDN configuration

coilysiren:
## Summary

Relates to #680

## Changes proposed

- Adds a Cloudfront CDN
- The CDN caches everything at 1 hour by default
- It is only active in staging and prod
- It logs to a (newly created) S3 bucket
- It forward to the frontend ALB
- It only caches GET, HEAD, OPTIONS

### DIFF
--- a/infra/modules/service/cdn.tf
+++ b/infra/modules/service/cdn.tf
@@ -42,7 +42,6 @@ resource "aws_cloudfront_distribution" "cdn" {
   enabled         = true
   is_ipv6_enabled = true
   aliases         = var.domain != null ? [var.domain] : []
-  staging         = var.environment_name == "prod" ? true : false
 
   origin {
     domain_name = "http://${aws_lb.alb[0].dns_name}"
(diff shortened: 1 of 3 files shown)


mdragon (dismissed):
Looks good!

infra/modules/service/cdn.tf
```diff
@@ -0,0 +1,111 @@
...
+    cached_methods         = ["GET", "HEAD", "OPTIONS"]
+    target_origin_id       = local.default_origin_id
+    cache_policy_id        = aws_cloudfront_cache_policy.default[0].id
+    compress               = true
+    viewer_protocol_policy = var.cert_arn == null ? "allow-all" : "redirect-to-https"
+
+    # Default to caching for 1 hour, with a minimum of 1 minute.
+    # The default TTL can be overriden by the `Cache-Control max-age` or `Expires` headers
+    # There's also a `max_ttl` option, which can be used to override the above headers.
+    min_ttl     = 60
```
acouch, line 83: Think we might want this as 0 is we set those in the app with the `s-maxage=` https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Expiration.html#ExpirationDownloadDist

coilysiren: I had this as 0 at first! I'll put it back

mdragon (approved):
Ship it!

PR #3085, opened 2024-12-03
[Issue #3066] Expose analytics CLI entry point in Docker

### LINKED ISSUE #3066: Expose analytics entrypoints in Docker image (i.e. without poetry)
### Summary

Find a way to install the analytics package directly in the docker image so that the entry points can be run using `docker run <image-tag> analytics` and `docker run <image-tag> db-migrate`

### Acceptance criteria

- [ ] We're able to run the DB migrations as part of the CI/CD pipeline

widal001:
## Summary

Exposes the `analytics` CLI entry point for the Docker image that allows us to run `docker run <image> etl db_migrate` as part of the CI/CD pipeline. 

Fixes #3066 

- Edits the `poetry run pip install` to use `--no-cache-dir` and use a wildcard for the `dist/*.whl` so that users don't have to change that line if we bump the version.

To prove that the entry point is available to run from the Docker container (without `poetry run`) do the following steps:

1. Checkout the current PR
2. Run `make release-build` from the root of the `/analytics/` sub-directory
3. Copy the `sha256:<image-hash>` from the resulting docker logs:
   ```
   => exporting to image                                                                                                      0.0s
   => => exporting layers                                                                                                     0.0s
   => => writing image sha256:14939cf59d75e1465ae5fd8030715d79ae33168bf0007346b571654a73914040                                0.0s
   ```
4. Use that image hash to run the following command `docker run <image-hash> analytics etl --help` and it should print out:
Screenshot 2024-12-03 at 1 55 21 PM

**Note:** Trying to run the actual DB migration with this method will fail because the migration depends on interacting with a live Postgres instance, which is what we use `docker compose` for but the current `docker-compose.yml` has the target set to `dev` based on [this Slack thread](https://betagrantsgov.slack.com/archives/C05TSL64VUH/p1733259136748469?thread_ts=1733253723.080049&cid=C05TSL64VUH) testing it with `docker compose run` may not be necessary.

There are two main strategies for installing the package and exposing the CLI entry point:

- `poetry install --only-root` Not chosen because it installs an editable version of the python package (equivalent to doing `pip install . -e`) which is usually reserved for development
- `poetry build ... && poetry run pip install ...` Chosen because it first builds a stable version of the package and installs it

### DIFF
--- a/analytics/Dockerfile
+++ b/analytics/Dockerfile
@@ -118,7 +118,8 @@ RUN poetry install --no-root --only main
 # include section in pyproject.toml. Also note that if you change the name or
 # version section in pyproject.toml, you will need to change the dist/... to match
 # or the application will not build
-RUN poetry build --format wheel && pip install 'dist/simpler_grants_gov_analytics-0.1.0-py3-none-any.whl'
+RUN poetry build --format wheel && \
+  pip install --no-cache-dir 'dist/simpler_grants_gov_analytics-0.1.0-py3-none-any.whl'
 
 # Add project's virtual env to the PATH so we can directly run poetry scripts
 # defiend in pyproject.toml
--- a/analytics/Dockerfile
+++ b/analytics/Dockerfile
@@ -115,12 +115,9 @@ RUN poetry install --no-root --only main
 
 # Build the application binary (python wheel) defined in pyproject.toml
 # Note that this will only copy over python files, and files stated in the
-# include section in pyproject.toml. Also note that if you change the name or
-# version section in pyproject.toml, you will need to change the dist/... to match
-# or the application will not build
-# RUN poetry build --format wheel && \
-#   pip install --no-cache-dir 'dist/simpler_grants_gov_analytics-0.1.0-py3-none-any.whl'
-RUN poetry install --only-root
+# include section in pyproject.toml.
+RUN poetry build --format wheel && \
+  poetry run pip install --no-cache-dir dist/*.whl
 
 # Add project's virtual env to the PATH so we can directly run poetry scripts
 # defiend in pyproject.toml
--- a/analytics/Dockerfile
+++ b/analytics/Dockerfile
@@ -13,22 +13,22 @@ RUN pip install --no-cache-dir poetry==1.8.2 --upgrade
 RUN apt-get update \
   # Remove existing packages before installing their never versions
   && apt-get remove --yes \
-    build-essential \
-    libc-dev \
-    libpq-dev \
-    postgresql \
-    wget \
-    jq \
+  build-essential \
+  libc-dev \
+  libpq-dev \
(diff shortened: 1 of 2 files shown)


analytics/Dockerfile
```diff
@@ -115,14 +115,18 @@ RUN poetry install --no-root --only main
 
 # Build the application binary (python wheel) defined in pyproject.toml
 # Note that this will only copy over python files, and files stated in the
-# include section in pyproject.toml. Also note that if you change the name or
-# version section in pyproject.toml, you will need to change the dist/... to match
-# or the application will not build
-RUN poetry build --format wheel && poetry run pip install 'dist/simpler_grants_gov_analytics-0.1.0-py3-none-any.whl'
+# include section in pyproject.toml.
+RUN poetry build --format wheel && \
+  poetry run pip install --no-cache-dir dist/*.whl
```
widal001, line 120: The change to `dist/*.whl` avoids breaking the install if the version or name changes. Happy to change this back if we want a developer to explicitly update this when they bump versions.

analytics/pyproject.toml
```diff
@@ -8,7 +8,6 @@ version = "0.1.0"
 
 [tool.poetry.scripts]
 analytics = "analytics.cli:app"
-db-migrate = "src.analytics.cli:migrate_database"
```
widal001, line 11: Removed this since the same functionality is available with `etl db_migrate` and I was having trouble exposing multiple entry points in the docker image.

analytics/Dockerfile
coilysiren: Make it narrower: simpler grants gov analytics *.whl

analytics/pyproject.toml
coilysiren: Unfortunately you will in fact need to put this one back. The problem you say likely stems from using ENTRYPOINT and the end of your docker file rather than CMD. So I would change it to CMD, and update your manual tests accordingly

coilysiren: You'll notice, for example, that the API uses CMD

(Ignore the fact that's its using poetry heh)

https://github.com/HHS/simpler-grants-gov/blob/main/api/Dockerfile#L127

analytics/Dockerfile
coilysiren: This is fine 👍!

coilysiren (commented):
Almost there!

analytics/pyproject.toml
widal001: Added this back and updates the Dockerfile to _not_ use a default entry point so that it can support both:

`docker run <image> db-migrate` and `docker run <image> analytics etl <sub-command>`

coilysiren (approved):
🚀 (even just what you have here was important to figure out)

PR #3089, opened 2024-12-03
[Issue #3078] cleanup translation related code

### LINKED ISSUE #3078: Translation middleware is not grabbing es value from paths
### Summary

The code that is supposed to be grabbing information about whether to serve an English or Spanish page is always setting the locale to English.

Note that as we do not currently have a usable Spanish version of the site, this is not urgent, but it is important that some basic translation can work to make sure future changes do not break the set up that is in place.

### Reproducibility

None

### How to reproduce the bug

- alternatively, add a console log to output the locale in src/i18n/server.tx line 14
- start a local server with either `npm run dev` or `npm run build && npm start`
- visit localhost:3000/es
- verify spanish messages do not appear, locale is logged as "en"

### Browser

None

### Code of Conduct

doug-s-nava:
## Summary
Fixes #3078
## Context for reviewers
When testing translation behavior, it was difficult to track down why the Spanish version of the site was not loading. In the end, it was a matter of us manually setting "en" as the locale in a number of places unnecessarily. Since `getMessagesWithFallbacks` handles filling any missing messages in with English, so there isn't a need to manually set 
the request locale to "en".

This change updates the entire app to support dynamically set locales.

It also:

* updates outdated next intl `set locale` methods (unstable_setRequestLocale is now stable, see https://github.com/amannn/next-intl/pull/1437)
* includes a couple of quality of life improvements related to translation code

### Test steps

1. add some spanish messages for local testing. Add the following to the spanish messages file:
```
export const messages = {
  ErrorPages: {
    page_not_found: {
      title: "No pagina nunca",
    },
  },
  Subscribe: {
    form: {
      name: "Nombre Primero",
      lastName: "Nombre Ultimo",
    },
  },
};
```
2. visit http://localhost:3000/es/subscribe
3. _VERIFY_: the spanish input titles added in step one appear
2. visit http://localhost:3000/es/subscrdfafdfibe
3. _VERIFY_: the spanish heading added in step one appear


frontend/next.config.js
```diff
@@ -1,6 +1,6 @@
 // @ts-check
 
-const withNextIntl = require("next-intl/plugin")("./src/i18n/server.ts");
+const withNextIntl = require("next-intl/plugin")();
```
doug-s-nava, line 3: see https://next-intl-docs.vercel.app/docs/usage/configuration - using a request.ts file, config will get picked up automatically. Not sure why this was in `server.ts` instead

frontend/src/i18n/config.ts
```diff
@@ -11,9 +11,9 @@ type RequestConfig = Awaited<
  * List of languages supported by the application. Other tools (Storybook, tests) reference this.
  * These must be BCP47 language tags: https://en.wikipedia.org/wiki/IETF_language_tag#List_of_common_primary_language_subtags
  */
-export const locales = ["en", "es"] as const;
+export const locales = ["en", "es"];
```
doug-s-nava, line 14: I'm of two minds here. It's important to list our supported locales, but as far typing is concerned, I am not sure we need to assert that the locale that comes through will always be one of these values. We're doing a check to see if the locale that comes through is supported (see request.ts), and that check seems strange and is complicated TS wise if we are hardening the list of locales as a type here. Fwiw I am fine to revert if we want to go that way

frontend/tests/pages/page.test.tsx
```diff
@@ -16,15 +16,15 @@ jest.mock("next-intl", () => ({
 
 describe("Home", () => {
   it("renders intro text", () => {
-    render(<Home />);
+    render(Home({ params: { locale: "en" } }));
```
doug-s-nava, line 19: this is a bit unfortunate, but seems to be the best way to deal with testing a top level page that takes params from the Next system

frontend/src/i18n/request.ts
acouch, line 1: Was there a reason for renaming this?

doug-s-nava: yeah, you can see my note in the config file - basically next intl will automatically pick up a file called request, whereas ours was called server and we had to pass a path to it to get it to work. Simplifies things slightly and makes use of built in behavior

frontend/src/app/[locale]/subscribe/page.tsx
```diff
@@ -54,12 +51,7 @@ export default function Subscribe() {
             })}
           </Grid>
           <Grid tabletLg={{ col: 6 }}>
-            <NextIntlClientProvider
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/subscribe/page.tsx:57]: had to make this change to get tests to work, but this wasn't doing anything anyway

frontend/src/i18n/request.ts
```diff
@@ -10,7 +10,16 @@ import { getMessagesWithFallbacks } from "./getMessagesWithFallbacks";
  */
 
 // @ts-expect-error TS2345: Argument of type error is expected behavior by next-intl maintainer: https://github.com/amannn/next-intl/issues/991#issuecomment-2050087509
-export default getRequestConfig(async ({ locale }) => {
+export default getRequestConfig(async ({ requestLocale }) => {
+  let locale = (await requestLocale) || "";
+
+  const isValidLocale = locales.includes(locale); // https://github.com/microsoft/TypeScript/issues/26255
```
doug-s-nava, line 16: moved this over from the `getMessagesWithFallbacks` code. Not a necessary change but just made more sense to me to do the locale default logic as high up in the call stack as possible

acouch (approved):
🎉  🇪🇸 🚀 !

PR #3090, opened 2024-12-03
[no ticket] Fix cron vulnerability scans not alerting slack

coilysiren:
## Context

The issue is visible here: https://github.com/HHS/simpler-grants-gov/actions/runs/12118942297/job/33784611951


mdragon (approved):
Good catch.

PR #3099, opened 2024-12-04
[Issue #3071] load New Relic browser code consistently

### LINKED ISSUE #3071: Get consistent New Relic Browser data
### Summary

On 11/22 we deployed out a test version of our New Relic implementation to our DEV environment after testing locally. I was able to observe that data was coming through in both our APM (backend) and Browser (frontend) views, and went ahead with implementing the change. However, when we again deployed the implementation on 11/26, the Browser view was not receiving any data from our development environment and I could confirm that the frontend New Relic script was not being added. I did some work to try and figure out why but left on vacation at end of day without a real solution.

The change that had been deployed to development was promoted to production on 11/27 and as of the morning of 12/2 morning we are seeing data come in to the production Browser view. However, we are still not seeing the script being inserted on my end, and do not see any calls to New Relic going out from the client. For what it's worth the APM integration seems to be working just fine.

We need to determine why this inconsistent behavior is happening and come up with a fix.

### Acceptance criteria

- [ ] we are confident that future DEV and PROD deployments will correctly initialize New Relic browser data collection on all clients

doug-s-nava:
## Summary
Fixes #3071

## Changes proposed
Refactors our layout situation to create a root layout which we can use to properly insert the client side New Relic code snippet into the head of the document consistently. This allows us to ensure that the snippet is generated on the server side at runtime, to fix bugs with consistency of loading the New Relic script that we have observed.

### Static rendering

We should ensure that this does not break the build time rendering we have in place for the opportunity detail pages. @acouch, what is the best way to test that?

### Translation

Note that I was slightly concerned that changing the layout situation here would negatively impact our translation related code, but was able to confirm that it works fine with these changes. For more on testing that, you can see https://github.com/HHS/simpler-grants-gov/pull/3089

### Test steps
1. Update .env.production to point to your locally running API server
2. start a "production like" build locally with New Relic creds (see 1pass). `npm run build && NEW_RELIC_APP_NAME="<1pass>" NEW_RELIC_LICENSE_KEY="<1pass>" npm run start:nr`
3. visit all pages on the app and
4. _VERIFY_: page loads as expected
5. _VERIFY_: the <head> of the document contains a populated "nr-browser-agent" script. See screenshot
6. _VERIFY_: New Relic beacons are sent. Note that Firefox seems to block these beacons by default, so try testing in Chrome. Filter Network tab of dev tools for calls to https://bam.nr-data.net

## Additional information

Screenshot 2024-12-04 at 9 19 48 AM


acouch:
Looks like this does break SSG:
image

Maybe that doesn't need to matter with the way that the CDN is working with the search page in dev, but we'd need to test that somehow.

doug-s-nava:
It might be back to the drawing board here. This also breaks accessibility.

PR #3100, opened 2024-12-04
[Issue #2625] set max-ages for app pages

### LINKED ISSUE #2625: Add cache lifetimes for frontend components
### Summary

The frontend and CDN will cache items across a couple of fronts. This task is to determine the correct cache lifetimes, document those, and implement them. The implementation should only be for adding or updating values for existing features / tools, not implementing new caches.

An in-exhaustive list includes.

* Fetch cache from server actions
* ISR cache (revalidate var) for opportunity and other pages
* cache lifetime for 404s in the CDN
* cache lifetime in the CDN for assets (images, files)

This ticket would entail updating the distribution settings in the CDN. If that is too much scope for a single ticket we could break them up.

### Acceptance criteria

- [ ] cache lifetimes documented
- [ ] cache lifetime values updated or a new ticket is created if more work is needed to implement a value

acouch:
## Summary
Fixes #2625

## Changes proposed
Explicitly set the headers for different parts of the app so they are cached by clients and the CDN appropriately.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -6,7 +6,7 @@
     "node": ">=20.0.0"
   },
   "scripts": {
-    "all-checks": "npm run test && npm run lint && npm run ts:check && npm run build",
+    "all-checks": "npm run lint && npm run ts:check && npm run test && npm run build",
     "build": "next build",
     "dev": "next dev",
     "dev:nr": "NODE_OPTIONS='-r @newrelic/next' next dev",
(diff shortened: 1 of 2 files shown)


coilysiren (approved):

frontend/package.json
```diff
@@ -6,7 +6,7 @@
     "node": ">=20.0.0"
   },
   "scripts": {
-    "all-checks": "npm run test && npm run lint && npm run ts:check && npm run build",
+    "all-checks": "npm run lint && npm run ts:check && npm run test && npm run build",
```
acouch, line 9: @doug-s-nava I moved `npm run test` last since the test takes longest and it stops if the test fails (which has happened a couple of times).

doug-s-nava (approved):

PR #3087, opened 2024-12-03
[Issue #3075] Setup /users/login redirect endpoint for OAuth

chouinar:
## Summary
Fixes #3075

## Changes proposed
Added a new /users/login endpoint that redirects you to the authorization endpoint for login.gov

Added a to-be-filled out /users/login/callback endpoint to receive the "response" from login.gov (several tickets will fill out the implementation in this)

A lot of reconfiguring of mock oauth server to make all this work locally

## Context for reviewers
This technically makes the login flow run across its entire set of endpoints, however the bulk of the logic regarding tokens is not yet included (a lot is implemented in other PRs already).

The general flow of this is:
* A user navigates to the login page
* The user is redirected to the OAuth signin page
* The user fills out the OAuth page and is redirected back to our /callback endpoint
* The callback endpoint will have the bulk of the logic (users, token generation, and validation)
* The callback endpoint redirects a user to a final page and passes off the token

Testing this was.. fun as redirects and Docker make a lot of the paths constantly differ from each other. It _should_ work in all cases, but I make no guarantees I didn't miss something.

## Additional information
An example flow with pictures.

A user navigates to the /v1/users/login endpoint:
Screenshot 2024-12-03 at 1 22 24 PM

In this case they can select the link as OpenAPI doesn't handle redirects and just dumps the redirect HTML into the response section. The endpoint builds a redirect URL to the OAuth client, and immediately redirects.

Selecting that link immediately redirects you to the Mock OAuth login page
Screenshot 2024-12-03 at 1 22 36 PM

Note that the name you fill out in the top box will become the ID we get back later (ie. the Login.gov user ID).

After you fill that out, you'll be redirected back to our API which doesn't do much yet, but at the end it does a final redirect and sends your token + a message to a page. That page I just made a quick endpoint for:
Screenshot 2024-12-03 at 1 59 03 PM

PR #3107, opened 2024-12-04
[Issue #3080] Error handling in the login endpoints

### LINKED ISSUE #3080: Error handling in the login endpoints
### Summary

When an error occurs we want to redirect to the same final location that a user goes to when they successfully login, except with some additional error details.

Rough notes:
* Putting a try/except around all of the logic in the endpoint is probably simplest, although consider a decorator for re-usability.
* If the error is one of a few common known ones (like our own jwt errors) then we should pass through its message, otherwise put a generic "Internal Server Error" style message
* The logic for redirect/building the URL from its pieces will already exist and just needs to be re-used

### Acceptance criteria

- [ ] Error handling added
- [ ] Tests added to validate the behavior for various cases

chouinar:
## Summary
Fixes #3080

## Changes proposed
Error handling added to the login endpoints using a decorator that is keeps us from needing duplicate work

## Context for reviewers
Normally putting a big `try/catch` around an entire API endpoint is a bad idea, we should let errors propagate, but if an error occurs (lets say the user has an issue in login.gov), we need the error to go somewhere - because these endpoints are all redirect endpoints, we send them to the "final" endpoint where we want to put the error.

If you aren't familiar with the decorator added here, it is just a wrapper around the function, effectively it lets us re-use a bit of code and "wrap" a function, in this case we effectively do:

```py
try:
     our_api_func(...)
except ...:
     ...
```

## Additional information
There aren't any actual places that can error without deliberately breaking things right now - so I've overriden random methods to break - as we build out the callback endpoint, plenty of errors will go through this logic and we'll change it a bit more.


api/src/auth/auth_utils.py
```diff
@@ -1,8 +1,76 @@
...
+        @functools.wraps(f)
+        def wrapper(*args: P.args, **kwargs: P.kwargs) -> flask.Response:
+            try:
+                return f(*args, **kwargs)
+            except HTTPError as e:
+                # HTTPError is what raise_flask_error raises
+                # and should encompass our "expected" errors
+                # that aren't a concern, as long as it isn't a 5xx
+                message = e.message
+                logger.info("Login flow failed")
```
mdragon, line 54: Do we want to include the message here so it's clearer what logs here correspond to what users are experiencing?

mdragon (dismissed):

api/src/auth/auth_utils.py
chouinar: I think I realized that about two PRs after this one and adjusted some of the logging here, I'll backport it after I merge the upstream changes

mdragon:
I approved it, but you noted there would be later churn, so feel free to tag me back in.

chouinar:
> I approved it, but you noted there would be later churn, so feel free to tag me back in.

@mdragon - https://github.com/HHS/simpler-grants-gov/pull/3087 needs to be approved first, these auth PRs are chained together so I need to get them approved/merged one-by-one

mdragon (approved):

PR #3114, opened 2024-12-05
[Issue #3076] Manage state value when calling/returning from login.gov auth

### LINKED ISSUE #3076: Modify /users/login to store state information in a new table
### Summary

TODO - but should use the `state` field we set and then later receive in the callback endpoint

### Acceptance criteria

- [ ] Table created
- [ ] Logic added for setting / parsing

chouinar:
## Summary
Fixes #3076

## Changes proposed
Added logic to support handling of the "state" value in OAuth

## Context for reviewers
In OAuth, state is a value you first pass the authorization endpoint that it then passes back to connect the original redirect to this new redirect (eg. lets our login endpoint connect to the callback endpoint)

Right now we only support storing the Nonce value which is another UUID we generate, which we'll use in some of the validation logic when we get to calling the token endpoint (next ticket).

There are still several components remaining to be built out which I've organized a bit of as a series of TODOs that represent next tickets.

https://developers.login.gov/oidc/authorization/ describes these nonce/state values in more detail

## Additional information
Still able to go through what we have of the login flow locally, and state gets saved to the DB:
Screenshot 2024-12-05 at 12 20 41 PM


api/src/services/users/login_gov_callback_handler.py
```diff
@@ -0,0 +1,76 @@
...
+        select(LoginGovState).where(LoginGovState.login_gov_state_id == callback_params.state)
+    ).scalar_one_or_none()
+
+    # If we don't have the state value in our DB, that either means:
+    # * login.gov is very broken and sending us bad data
+    # * Someone called this endpoint directly with a random value
+    #
+    # There isn't a way to truly separate those here, so we'll assume the more likely second one
+    # and raise a 404 to say we have no idea what they passed us.
+    if login_gov_state is None:
```
mdragon, line 63: Are these secrets? Can they not be re-used? Logging what was passed could help us troubleshoot between the two scenarios you're anticipating for lookups failing if we don't consider them dangerous to log.

mdragon (approved):
Looks good.

api/src/services/users/login_gov_callback_handler.py
chouinar: Secret, probably not. Reused, definitely no.

State is basically the way we connect the initial call to login.gov with the callback. If it is none, that means someone probably called this endpoint directly with some random value and we don't have that ID in our system.

The nonce value stored is kinda the same idea, but instead it gets returned when we call the token endpoint (a later PR).

See: https://developers.login.gov/oidc/authorization/ for some further details.

mdragon: Cool, so I think logging the raw value which the error will indicate we didn't find will help us understand/troubleshoot any failed connectivity stuff we encounter here.

mdragon (approved):

PR #3115, opened 2024-12-05
[Issue #2695] opportunity and search toggle with feature flag env vars

### LINKED ISSUE #2695: Add kill switch for Search and Opportunity data display
### Summary

In case we have an issue with bad data that rises to the level of a risk to the program we want to have a way to disable usage of the Search and Opportunity pages respectively to force people back to the existing site. 

Use content from #2735 

The Dev flags have been set up in AWS Systems Manager - Parameter Store:
/frontend/dev/feature-opportunity-off
/frontend/dev/feature-search-off

### Acceptance criteria

- [ ] With the Search flag set to True the Search page does not allow searching and does not display results
- [ ] With the Opportunity flag set to True the Opportunity page does not display data about the Opportunity
- [ ] When the flags are True a message is displayed directing the user to grants.gov until the issue can be resolved
- [ ] With the Search flag set to False the Search page behaves as normal
- [ ] With the Opportunity flag set to False the Opportunity page behaves as normal

doug-s-nava:
## Summary
Fixes #2695

## Changes proposed
The goal of the PR is the addition of two "kill switch" parameters that can be used to direct the search and opportunity pages, respectively, to a maintenance page. These parameters can be supplied either on the frontend via query params or feature flags admin panel, or via environment variables.

Setting flags via environment variables is a new functionality, so changes were made to the system to get this to work. Some deficiencies of the existing system were also patched up as part of this process.

### Test Steps

1. `FEATURE_OPPORTUNITY_OFF=true FEATURE_SEARCH_OFF=true npm run dev` on this branch
3. _VERIFY_: maintenance page appears
4. _VERIFY_: maintenance page matches designs and includes correct text from spec
2. visit http://localhost:3000/opportunity/1
3. _VERIFY_: maintenance page appears
4. _VERIFY_: maintenance page matches designs and includes correct text from spec
2. visit http://localhost:3000/search?_ff=searchOff%3Afalse
3. _VERIFY_: search page appears
4. 2. visit http://localhost:3000/opportunity/1?_ff=opportunityOff%3Afalse
3. _VERIFY_: opportunity page appears
8. Repeat 2 - x but in faux prod mode with `npm run build && FEATURE_OPPORTUNITY_OFF=true FEATURE_SEARCH_OFF=true npm start`

#### Dockerized
I would like to be able to test the changes to the Dockerfile locally, but can't without creating a docker compose that runs the production Next build and start rather than `next dev`. I feel like that's out of scope for this change, so we should just deploy out to DEV and test there.


frontend/src/services/FeatureFlagManager.ts
```diff
@@ -116,25 +122,25 @@ export class FeatureFlagsManager {
...
+    - As query param values are set in middleware on each request, query params have the highest precedence
+    - Values set in cookies will be persistent per browser session unless explicitly overwritten
+    - This means that simply removing a query param from a url will not revert the feature flag value to
+    the value specified in environment variable or default, you'll need to clear cookies or open a new private browser
+
+  */
   get featureFlags(): FeatureFlags {
     return {
       ...this.defaultFeatureFlags,
+      ...this.featureFlagsFromEnvironment,
```
doug-s-nava, line 139: note that this means that if a user in any way sets their cookie to override the flag, they'll be able to access search or opportunities

frontend/src/services/FeatureFlagManager.ts
```diff
@@ -238,14 +264,14 @@ export class FeatureFlagsManager {
...
+   * Toggle feature flag on cookie
    */
-  setFeatureFlag(featureName: string, enabled: boolean): void {
+  setFeatureFlagCookie(featureName: string, enabled: boolean): void {
     if (!this.isValidFeatureFlag(featureName)) {
       throw new Error(`\`${featureName}\` is not a valid feature flag`);
     }
     const featureFlags = {
-      ...this.featureFlags,
+      ...this.featureFlagsCookie,
```
doug-s-nava, line 274: this is a change to the existing functionality - previously whenever we set the feature flags cookie it would enshrine the default feature flag values in the cookie, but that functionality conflicts with idea of honoring flags set in the environment and using cookies as an override. Setting a flag on the cookie should, in this case, only set the single flag value specified

frontend/src/app/[locale]/opportunity/[id]/page.tsx
```diff
@@ -20,7 +22,12 @@ import OpportunityIntro from "src/components/opportunity/OpportunityIntro";
 import OpportunityLink from "src/components/opportunity/OpportunityLink";
 import OpportunityStatusWidget from "src/components/opportunity/OpportunityStatusWidget";
 
+type OpportunityListingProps = {
+  params: { id: string };
+} & WithFeatureFlagProps;
+
 export const revalidate = 600; // invalidate ten minutes
+export const dynamic = "force-dynamic";
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/opportunity/[id]/page.tsx:30]: I honestly don't understand why this needs to be here since we are already including the generateStaticParams, but without the the production build fails to render the page due to the dependence on searchParams in the withFeatureFlag code

frontend/src/constants/featureFlags.ts
```diff
@@ -2,7 +2,10 @@ import { FeatureFlags } from "src/services/FeatureFlagManager";
 
 // Feature flags should default to false
 export const featureFlags: FeatureFlags = {
-  // This is for showing the search page as it is being developed and user tested
+  // This is for hiding the search page as it is being developed and user tested
   // This should be removed when the search page goes live, before May 2024
-  showSearchV0: true,
+  hideSearchV0: false,
```
doug-s-nava, line 7: do we still need this? I changed it to a `hide` vs `show` variable to match the pattern here and allow us to more easily reason about feature flag effects

mdragon (approved):
Looks good.

acouch (dismissed):
Tested locally, looks great.

mdragon (approved):

PR #3122, opened 2024-12-06
[Issue #3102] Call Oauth token endpoint

### LINKED ISSUE #3102: Setup logic for calling the Oauth token endpoint
### Summary

See: https://developers.login.gov/oidc/token/

We only need to handle calling and getting a response, logic for parsing the `id_token` value returned is already implemented as we originally planned on the frontend making this call and passing it to us.

### Acceptance criteria

* [ ] Any errors handled
* [ ] Endpoint can be mocked out for local tests
* [ ] Code works with local mock Oauth server

chouinar:
## Summary
Fixes #3102

## Changes proposed
Added a client (and mock) for calling an OAuth token endpoint

A lot of restructuring of test utils for clearer setup

## Context for reviewers
https://developers.login.gov/oidc/token/

This gets the token from the OAuth server and parses the response. There is more work to do on this later as login.gov requires a special JWT to also be passed, but a basic version doesn't need that (our local mock doesn't care), so I'll follow-up on that later.

This approach to setting up a client is following some patterns I've used before that worked well. Building a mock version alongside the real one helps with testing.

## Additional information
Still nothing new visually, under the hood it is just one more big step remaining to process the token

### DIFF
--- a/api/local.env
+++ b/api/local.env
@@ -45,7 +45,9 @@ LOGIN_GOV_CLIENT_ID=local_mock_client_id
 
 LOGIN_GOV_JWK_ENDPOINT=http://host.docker.internal:5001/issuer1/jwks
 LOGIN_GOV_AUTH_ENDPOINT=http://localhost:5001/issuer1/authorize
+LOGIN_GOV_TOKEN_ENDPOINT=http://host.docker.internal:5001/issuer1/token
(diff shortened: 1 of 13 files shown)


mdragon (dismissed):
LGTM

chouinar:
@mdragon - could you look at this one again - had a merge conflict

mdragon (approved):

PR #3125, opened 2024-12-06
[no ticket] Adds some checkov skips

coilysiren:
## Context

follows https://github.com/HHS/simpler-grants-gov/pull/3082


mdragon (approved):
+1

PR #3130, opened 2024-12-06
[Issue #3127] Add a New Relic agent to the API

### LINKED ISSUE #3127: Add a New Relic agent to the API
### Summary

We can setup New Relic monitoring on the API using an agent + a configuration file.

https://docs.newrelic.com/install/python/?python-non-web=non-web-yes

This requires running something like:

```py
def init_newrelic() -> None:
    newrelic.agent.initialize(
        config_file=os.path.join(os.path.dirname(__file__), "../../../..", "newrelic.ini"),
        environment=os.environ.get("ENVIRONMENT", "local"),
    )
```

first before the application itself starts up, this needs to be done before any other imports

### Acceptance criteria

- [ ] Configuration .ini file added
- [ ] App initialized

mikehgrantsgov:
## Summary
Fixes #3127

## Changes proposed
Add a New Relic ini file, dependency, and lightweight utilities to send custom events.

## Context for reviewers
New Relic monitoring is needed for the application and the .ini file is the start place for configuration. Environment-specific configuration is available as well. The `NEW_RELIC_LICENSE_KEY` will be provided as an environment variable (command line?) to the application in non-local environments.

## Additional information
Local env reports to a fake collector. Other envs will report property to NR.
Screenshot 2024-12-06 at 2 32 34 PM

Local env running gunicorn showing NewRelic initializing and sending events to the fake collector:
Screenshot 2024-12-09 at 3 08 45 PM
Screenshot 2024-12-09 at 3 08 38 PM


api/src/util/newrelic/events.py
chouinar, line 1: Could we put the newrelic folder under the adapters folder?

api/src/app.py
```diff
@@ -67,6 +68,8 @@ def create_app() -> APIFlask:
     if os.getenv("ENVIRONMENT") == "local":
         initialize_jwt_auth()
 
+    init_newrelic()
```
chouinar, line 71: Doesn't this need to happen before anything else? We did that before / the New Relic docs say that needs to happen.

https://docs.newrelic.com/install/python/?python-non-web=non-web-yes#integrating

---

It might be that this needs to be done for multi-threading purposes, which we only do non-locally unless you reconfigure some commands.

You'll need to swap out the command temporarily at https://github.com/HHS/simpler-grants-gov/blob/main/api/docker-compose.yml#L78 for what we use in prod to run via Gunicorn: https://github.com/HHS/simpler-grants-gov/blob/main/api/Dockerfile#L127

api/newrelic.ini
```diff
@@ -0,0 +1,270 @@
...
+app_name = SIMPLER-GRANTS-API-STAGING
+monitor_mode = true
+
+[newrelic:prod]
+app_name = SIMPLER-GRANTS-API-PROD
+monitor_mode = true
+
+[newrelic:dev]
+app_name = SIMPLER-GRANTS-API-DEV
+monitor_mode = true
```
chouinar, line 268: @coilysiren - I think we talked about this a few weeks ago, but what landing scheme did we land on for the API app in New Relic?

mdragon (dismissed):
Looks good.

api/src/util/newrelic/events.py
```diff
@@ -0,0 +1,74 @@
...
+AGENT_ATTRIBUTE_LIMIT = 128
+EVENT_API_ATTRIBUTE_LIMIT = 255
+
+
+def record_custom_event(event_type: str, params: Dict[Any, Any]) -> None:
+    # legacy aliases
+    params["request.uri"] = params.get("request.path")
+    params["request.headers.x-amzn-requestid"] = params.get("request_id")
+
+    # legacy FineosError attribute aliases
```
chouinar, line 19: We can probably adjust some of this / remove Fineos reference.

FYI our logging is a little different now on this project, https://github.com/HHS/simpler-grants-gov/blob/main/api/src/logging/flask_logger.py#L165 has a few more things.

Might be worth adding something like this here since we don't have that x-amzn ID set anywhere yet.
```
if flask.has_request_context():
    params["api.request.internal_id"] = getattr(flask.g, "internal_request_id", None)
```

api/src/util/newrelic/events.py
```diff
@@ -0,0 +1,74 @@
+import logging
+import sys
+from types import TracebackType
+from typing import Any, Dict, Optional, Tuple, Union
```
chouinar, line 4: While we're borrowing code, let's cleanup the typing, should be able to trim out dict, optional, union for the simpler forms (`dict`, `x | None`, `x | y` respectively)

api/src/util/newrelic/events.py
mdragon: This brings up a good point, especially as we start to look at requests hitting, CDN, LB, backend, S3, DB, OpenSearch, etc. Should we be using the X-amzn one if it's present since it'll get set the earliest in that whole stack? Fall back to our own generated one if we're hitting the API first like locally?

chouinar: Right now, it isn't set. I'm not sure where that values comes from, that value is just what a prior project used, and I never figured out what was setting it in their infra.

api/newrelic.ini
coilysiren: this looks fine to me - the `SIMPLER-GRANTS-` bit is probably superfluous but if that matches the frontend then I'm down to keep it

api/src/util/newrelic/events.py
mikehgrantsgov: Yep, changed this.

api/src/util/newrelic/events.py
mikehgrantsgov: Got this too.

api/src/util/newrelic/events.py
mikehgrantsgov: > We can probably adjust some of this / remove Fineos reference.
> 
> FYI our logging is a little different now on this project, https://github.com/HHS/simpler-grants-gov/blob/main/api/src/logging/flask_logger.py#L165 has a few more things.
> 
> Might be worth adding something like this here since we don't have that x-amzn ID set anywhere yet.
> 
> ```
> if flask.has_request_context():
>     params["api.request.internal_id"] = getattr(flask.g, "internal_request_id", None)
> ```

FYI there is another Fineos reference in api/tool/console/interactive.py.

api/src/app.py
mikehgrantsgov: @chouinar What if we just lift init_newrelic() out into app.py, like I did here? Would we still need to change the Docker command?

api/src/util/newrelic/events.py
mdragon: https://kevinhakanson.com/2017-03-08-aws-request-ids/ Looks like it was API Gateway back in the day, but I wonder if CDN might be doing it too now, or if there's another one we should also have in the hierarchy.

mdragon: X-Amz-Cf-Id is the Cloudfront one. So for API code I guess, x-amzn-requestid, falling back to X-Amz-Cf-Id, falling back to our own generated one?
https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/RequestAndResponseBehaviorCustomOrigin.html#request-custom-headers-behavior

mdragon (dismissed):

api/src/app.py
chouinar: I'm suggesting you adjust the docker command to test that it works in multi-threaded/multi-process mode, not as a permanent change

chouinar (changes requested):
I'd like to verify that this works locally when running via gunicorn before we merge - the process setup is very different and likely why New Relic says it needs to run before everything

chouinar (approved):
LGTM!

PR #3131, opened 2024-12-06
[Issue #2810] Create token for auth and connect everything

### LINKED ISSUE #2810: Fully connect the /users/token endpoint component pieces

chouinar:
## Summary
Fixes #2810

## Changes proposed
Finish connecting together the Oauth callback endpoint, returning a token that can be used in other endpoints

Deleted the /token endpoint we no longer need

## Context for reviewers
There are a few cases that aren't quite handled yet that will be covered in subsequent tickets, but this does complete things working end-to-end for the common scenarios.

Selecting login will direct you to oauth, back to us, and then to a final endpoint where you get a token.

## Additional information

Go to http://localhost:8080/v1/users/login - get redirected immediately to the Oauth login flow
Screenshot 2024-12-06 at 3 17 58 PM

Get redirected back to us, and finally dump the token on another landing page:
Screenshot 2024-12-06 at 3 18 29 PM

The user exists in our DB, and locally the name you typed in becomes the external user ID (actual login.gov will have a UUID):
Screenshot 2024-12-06 at 3 20 02 PM

I can use this token to call our endpoints that expect a token, in this case the refresh endpoint:
Screenshot 2024-12-06 at 3 21 46 PM


api/openapi.generated.yml
```diff
@@ -1766,6 +1684,25 @@ components:
...
+      type: object
+      properties:
+        user_id:
+          type: string
+          description: The internal ID of a user
+          example: 861a0148-cf2c-432b-b0b3-690016299ab1
+        email:
+          type: string
+          description: The email address returned from Oauth2 provider
+          example: js@gmail.com
```
mdragon, line 1697: Nit, but make this like "user@example.com?"

mdragon (dismissed):

api/src/db/models/user_models.py
```diff
@@ -6,43 +6,44 @@
...
 
-    user_id: Mapped[uuid.UUID] = mapped_column(UUID, primary_key=True)
+    user_id: Mapped[uuid.UUID] = mapped_column(UUID, primary_key=True, default=uuid.uuid4)
 
 
 class LinkExternalUser(ApiSchemaTable, TimestampMixin):
     __tablename__ = "link_external_user"
 
-    link_external_user_id: Mapped[uuid.UUID] = mapped_column(UUID, primary_key=True)
+    link_external_user_id: Mapped[uuid.UUID] = mapped_column(primary_key=True, default=uuid.uuid4)
```
mdragon, line 23: If we need these to match to connect to an existing user do we want to make our workaround for it not existing more specific to our mock than having this just silently fill in the default? I've seen where you like MD5 the string value and then that is a valid UUID so the same identifier put through our mock twice would match the same internal user?

api/src/auth/login_gov_jwt_auth.py
```diff
@@ -164,7 +164,9 @@ def get_final_redirect_uri(
     return f"{config.login_final_destination}?{encoded_params}"
 
 
-def validate_token(token: str, config: LoginGovConfig) -> LoginGovUser:
+def validate_token(token: str, config: LoginGovConfig | None = None) -> LoginGovUser:
+    if not config:
+        config = get_config()
 
     # TODO - this iteration approach won't be necessary if the JWT we get
     #        from login.gov does actually set the KID in the header
```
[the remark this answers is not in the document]
chouinar: Yeah, I wasn't sure if it would exist because the docs never mentioned it, so I built the code with the assumption it wouldn't.

After my testing yesterday, I know it does exist, so made a ticket for a future sprint to use the KID directly

api/src/db/models/user_models.py
chouinar: This is actually something I shouldn't have added, was being a bit too copy-paste heavy with some cleanup here.

This is definitely a column that should never have a default.

chouinar: The `link_external_user_id` is just an internal ID.

The one that connects to the user in login.gov goes in the `external_user_id` field

chouinar: Happy to adjust the naming if you can help figure out a better naming scheme, I even made the same mistake a few times

mdragon: Hmm, ok, I don't have a better name... so I just missed the context of the table name. Now I'm noticing it's the PK too so that wouldn't really make sense either.

PR #3136, opened 2024-12-06
[no ticket] Initializes logs for db-migrate entry point

widal001:
## Summary

Initializes logger for `db-migrate` entry point which runs as part of the CI/CD pipeline

1. Checkout PR
2. Run `make release-build`
3. Get the image tag from the bottom of the output
4. Run `docker run <image-tag> db-migrate`

Screenshot 2024-12-06 at 4 34 42 PM

### DIFF
--- a/analytics/src/analytics/cli.py
+++ b/analytics/src/analytics/cli.py
@@ -61,23 +61,14 @@ app.add_typer(import_app, name="import", help="Import data into the database")
 app.add_typer(etl_app, name="etl", help="Transform and load local file")
 
 
-def init() -> None:
-    """Shared init function for all scripts."""
-    # Setup logging
-    init_logging(__package__)
-    init_app(logging.root)
+# Setup logging
+init_logging(__package__)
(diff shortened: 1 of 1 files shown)


coilysiren (approved):

analytics/src/analytics/cli.py
```diff
@@ -61,18 +61,14 @@
...
 
 
-def init() -> None:
-    """Shared init function for all scripts."""
-    # Setup logging
-    init_logging(__package__)
-    init_app(logging.root)
+# Setup logging
+init_logging(__package__)
+init_app(logging.root)
```
chouinar, line 66: This is potentially going to make local dev / testing more difficult if you always enable logging as soon as the module loads.

What is the context for this change / where do you want logging that you currently don't have it?

coilysiren:
The idea behind this change was merged into https://github.com/HHS/simpler-grants-gov/pull/3893

PR #3139, opened 2024-12-06
[no ticket] fix silly infinite redirect issue

coilysiren:
## Context

`default_root_object` does the following:

> [default_root_object](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudfront_distribution#default_root_object-1) (Optional) - Object that you want CloudFront to return (for example, index.html) when an end user requests the root URL.

It sends the user somewhere when they request `/`, like for example to the literal route `index.html`. Well we don't have an `index.html`. So I figured, `/` should just go to `/`, right? Wrong. When you tell `/` to go to `/` what you end up with is an infinite redirect loop. This me an undue amount of time to debug 😅 

I have no idea why AWS allows you to shoot yourself in the foot like this. Pointing `default_root_object` to `/` should return a validation error on AWS's end...


coilysiren:
I feel like checkov is trolling me

image

mdragon (approved):

PR #3147, opened 2024-12-09
[Unticketed] Allow for null deploy metadata

chouinar:
## Changes proposed
For our deploy metadata, allow the value to be null (however unlikely) to avoid breaking our jobs

## Context for reviewers
This shouldn't ever be null, but we did hit a weird case where it was due to manual testing which broke our ECS tasks in dev for a few days. This just makes a non-required field, actually non-required. Very useful, but not required.


mdragon (approved):
Good fallback.

coilysiren (approved):

PR #3150, opened 2024-12-09
[Issue #3058] Static List of Agencies for Search filter was incomplete

### LINKED ISSUE #3058: Add remaining agencies to the agency filter on search
### Summary

Update agency mappings on the frontend so we render all agencies that have opportunities associated with them.

Context and notes:

- https://nava.slack.com/archives/C057K146W8H/p1731964053115059?thread_ts=1731959195.730169&cid=C057K146W8H
- https://github.com/HHS/simpler-grants-gov/blob/main/frontend/src/components/search/SearchFilterAccordion/SearchFilterOptions.ts#L212

### Acceptance criteria

- [ ] Mappings updated and all agencies that have opportunities associated with them show in the agency filter on search

mdragon:
## Summary
Fixes #3058

## Changes proposed
Pasting in an updated JSON extract that is complete for all Agencies listing Opportunities currently. This will move to be a live API call in a follow-up ticket.

## Context for reviewers
Testing by validating that the filter list for Agencies is much more extensive now. Because we're using a static list of Prod Agencies we can't actually test that the filters work and return Opportunities

### DIFF
--- a/frontend/src/components/search/SearchFilterAccordion/SearchFilterOptions.ts
+++ b/frontend/src/components/search/SearchFilterAccordion/SearchFilterOptions.ts
@@ -212,507 +212,3516 @@ export const categoryOptions: FilterOption[] = [
 
 export const agencyOptions: FilterOption[] = [
   {
-    id: "ARPAH",
-    label: "Advanced Research Projects Agency for Health (ARPAH)",
-    value: "ARPAH",
-  },
-  {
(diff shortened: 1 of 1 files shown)


doug-s-nava (approved):
it might've been cool to move the agencies to their own file, but especially since this is a temporary fix ✅ ✅ ✅ ✅ ✅ ✅

PR #3163, opened 2024-12-10
[Issue #3162] Add EXPORT_OPP_DATA_FILE_PATH to terraform to run opportunity extracts

### LINKED ISSUE #3162: Add `EXPORT_OPP_DATA_FILE_PATH` to terraform to run opportunity extracts
### Summary

The daily extract job you setup is failing due to `EXPORT_OPP_DATA_FILE_PATH` not being set as an environment var.

### Acceptance criteria

Environment variable is added to the container.

mikehgrantsgov:
## Summary
Fixes #3162

## Changes proposed
Add env var for nightly task to run.

## Context for reviewers


api/src/task/opportunities/export_opportunity_data_task.py
```diff
@@ -59,12 +60,30 @@ def __init__(
...
-        )
-        self.csv_file = os.path.join(
-            config.file_path, f"opportunity_data-{self.current_timestamp}.csv"
-        )
+        # Get base path (either local or S3)
+        base_path = os.getenv("GENERAL_S3_BUCKET_URL", "").rstrip("/")
+
+        # Parse the path to determine if it's S3 or local
+        parsed_url = urlparse(base_path)
+        self.is_s3 = parsed_url.scheme == "s3"
```
chouinar, line 68: I thought we were fine to do this all in the terraform? This code should not be aware that it is using s3 at all. 

We should 100% not be doing it here, at worst we should be just doing `WHATEVER_FILE_PATH="s3://bucket-name` (using the actual vars) and then just append the path afterwards.

chouinar: Based on Kai's PR, could we do something like:

```
{ name : "EXPORT_OPP_DATA_FILE_PATH", value : "${aws_s3_bucket.general_purpose.bucket_regional_domain_name}/public-extracts/" }
```

mikehgrantsgov: There we go. Much cleaner!

infra/modules/service/main.tf
```diff
@@ -37,7 +37,8 @@ locals {
     { name : "DEPLOY_GITHUB_SHA", value : data.external.deploy_github_sha.result.value },
     # TODO: https://github.com/HHS/simpler-grants-gov/issues/3177
     # { name : "DEPLOY_GITHUB_REF", value : data.external.deploy_github_ref.result.value },
-    { name : "DEPLOY_WHOAMI", value : data.external.whoami.result.value }
+    { name : "DEPLOY_WHOAMI", value : data.external.whoami.result.value },
+    { name : "EXPORT_OPP_DATA_FILE_PATH", value : "${aws_s3_bucket.general_purpose.bucket_regional_domain_name}/public-extracts/" }
```
chouinar, line 41: @coilysiren - just confirming this approach works for you

mikehgrantsgov: Hi @coilysiren, would you be able to leave a review please?

coilysiren: Huh... why didn't I get notifications about this earlier. I'm usually on top of this sort of thing :/

coilysiren: So this isn't quite what we want - we want to avoid putting API specific logic into `infra/modules/whatever` unless it goes into its own file

api/src/task/opportunities/export_opportunity_data_task.py
coilysiren: So this code is actually closer to what we want. More specifically it should look like an env var that is null (or an empty string, or ./), locally, but is set to the s3 bucket path in terraform. You can use BUCKET_NAME for that. Its going to be unintuitive in the code, but that's the price of using a library that mixes the s3 paths with filesystem paths

coilysiren: To put this another way: all terraform knows is s3. And terraform is already giving you the bucket name. 

The application layer is the place where there's a package that can write to either S3 or local based on the path, so the application is where this logic should fundamentally live. But you should be able to do it with configuration, and not require an `if S3 then...` case

coilysiren: I can change the BUCKET_NAME env var to something that makes more sense given that there's dual concerns here... like... FILES_BASE_NAME? Idk gimme a few minutes to think on that

coilysiren (approved):

chouinar:
@mikehgrantsgov - tests need to be updated

chouinar (approved):

PR #3176, opened 2024-12-10
2 service deploy fixes

coilysiren:
## Context

2 quick infra fixes to unblock prod deploy

## Testing

I deployed this to dev already to test

### DIFF
--- a/infra/modules/service/main.tf
+++ b/infra/modules/service/main.tf
@@ -9,9 +9,9 @@ data "external" "whoami" {
   program = ["sh", "-c", "whoami | xargs -I {} echo '{\"value\": \"{}\"}'"]
 }
 
-data "external" "deploy_github_ref" {
-  program = ["sh", "-c", "git branch --show-current | xargs -I {} echo '{\"value\": \"{}\"}'"]
-}
+# data "external" "deploy_github_ref" {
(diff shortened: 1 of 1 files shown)


lucasmbrown-usds (approved):
I used to be a pretty good SRE - now my opinion should be taken with a grain of salt!

infra/modules/service/main.tf
```diff
@@ -9,9 +9,9 @@ data "external" "whoami" {
   program = ["sh", "-c", "whoami | xargs -I {} echo '{\"value\": \"{}\"}'"]
 }
 
-data "external" "deploy_github_ref" {
-  program = ["sh", "-c", "git branch --show-current | xargs -I {} echo '{\"value\": \"{}\"}'"]
-}
+# data "external" "deploy_github_ref" {
```
lucasmbrown-usds, line 12: Maybe add a note with a ticket number or another comment of TODO, explaining how and when we'll fix this?

coilysiren: That'll reset the review so I'll fast follow

coilysiren: https://github.com/HHS/simpler-grants-gov/issues/3177

mdragon (approved):

infra/modules/service/main.tf
lucasmbrown-usds: Perfect, thanks!!

PR #3208, opened 2024-12-13
Update to use newrelic.agent.background_task

mikehgrantsgov:
## Summary
Fixes #{ISSUE}

### DIFF
--- a/api/src/task/ecs_background_task.py
+++ b/api/src/task/ecs_background_task.py
@@ -6,8 +6,10 @@ import uuid
 from functools import wraps
 from typing import Callable, Generator, ParamSpec, TypeVar
 
+import newrelic.agent
 import requests
 
+from src.adapters.newrelic import init_newrelic
 from src.logging.flask_logger import add_extra_data_to_global_logs
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


api/src/task/ecs_background_task.py
```diff
@@ -47,8 +49,12 @@ def entrypoint(db_session: db.Session):
     def decorator(f: Callable[P, T]) -> Callable[P, T]:
         @wraps(f)
         def wrapper(*args: P.args, **kwargs: P.kwargs) -> T:
+            # Initialize New Relic at the start of any task
+            init_newrelic()
```
chouinar, line 53: We'd still keep init where it was, create_app gets called when we start up the API as a service and for background tasks already, and all that happens well before this for background tasks.

api/src/task/ecs_background_task.py
```diff
@@ -47,8 +49,12 @@ def entrypoint(db_session: db.Session):
...
+
             with _ecs_background_task_impl(task_name):
-                return f(*args, **kwargs)
+                # Finally execute the function with New Relic instrumentation
+                return newrelic.agent.background_task(name=task_name)(f)(*args, **kwargs)
```
chouinar, line 57:
```suggestion
                # Finally execute the function with New Relic instrumentation
                return newrelic.agent.background_task(name=task_name, group="Python/ECSTask")(f)(*args, **kwargs)
```

api/src/task/ecs_background_task.py
mikehgrantsgov: Makes sense, thanks

chouinar: We shouldn't need this here - create_app already is called so init_newrelic was already called well before this

mikehgrantsgov: Got it -- forgot that create_app is called on every task well

chouinar (approved):

Issue #3214, opened 2024-12-13
Review uncategorized content in Internal Wiki for migration


sarahknoppA6:
### Summary

Many items in the internal wiki are out of date and uncategorized. These items need to be sorted through and a determination needs to be made about if they should be moved to the external wiki, Google drive, or deleted and if they need to be updated first.

https://docs.google.com/spreadsheets/d/1uHWPqqWpZIPQzjZ1I1nAK4WlDo04sHR1Q31dVOBEwmc/edit?gid=1985278195#gid=1985278195

### Acceptance criteria

All content on the wiki audit spreadsheet labelled "Other" and/or with comments about the status have a new location or a decision has been made to remove them. Please track the status using the spreadsheet so the team will know if specific content is not going to move to Drive or the public wiki.

PR #3225, opened 2024-12-15
[Issue #3214] update fonts to Public Sans, Newsreader

### LINKED ISSUE #3124: [Brand FE] Update typography
### Summary

The new Grants brand uses Public Sans (for sans-serif type) and Newsreader (for serif type). The styles should be updated to use these two fonts. 

- https://fonts.google.com/specimen/Public+Sans 
- https://fonts.google.com/specimen/Newsreader 

USWDS comes with Public Sans included, but defaults to Source Sans Pro. Changing the default sans typeface should be as simple as adding `$theme-font-type-sans: "public-sans"` to the theme settings. 

Newsreader will need to be added. See [Adding fonts to USWDS](https://designsystem.digital.gov/design-tokens/typesetting/font-family/#adding-fonts-to-uswds-2). 

This ticket is only for switching the fonts. Any adjustments to typography (size, weight, font) of individual elements will be done via other issues. 

**Target the [feature/brand](https://github.com/HHS/simpler-grants-gov/tree/feature/brand) branch**

### Acceptance criteria

- [x] San-serif font is switched to Public Sans
- [x] Serf font is switched to Newsreader

andycochran:
## Summary
Fixes #3124 

## Changes proposed
This change adds the Newsreader font from the Google Fonts hosting service, sets the sans serf font to Public Sans, defines a USWDS type token for Newsreader, and sets the serif font to Newsreader (with the fallback set to Georgia, per brand guidelines). 

## Context for reviewers
When you view the site, all type should be Public Sans or Newsreader. You should no longer see Source Sans Pro or Merriweather. This change just changes the fonts. There will be subsequent PRs to use the right fonts in the right places so that the design looks better. 

## Additional information

Before (left) & after (right)
image

### DIFF
--- a/frontend/src/styles/_uswds-theme-custom-styles.scss
+++ b/frontend/src/styles/_uswds-theme-custom-styles.scss
@@ -22,6 +22,8 @@ i.e.
 
 @use "uswds-core" as *;
 
+@import url("https://fonts.googleapis.com/css2?family=Newsreader:ital,opsz,wght@0,6..72,200..800;1,6..72,200..800&display=swap");
+
 .pdf-card {
   transition: transform 0.2s;
   @include u-shadow(2);
--- a/frontend/src/styles/_uswds-theme.scss
+++ b/frontend/src/styles/_uswds-theme.scss
@@ -58,6 +58,15 @@ in the form $setting: value,
   ),
 
   // Typography settings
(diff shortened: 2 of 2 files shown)


mdragon (approved):
10/10 would typography again.

Issue #2317, opened 2024-09-30
Copy data from [source] into the opportunity table(s) in the analytics db


mxk0:
### Summary

Can be a full replace of the data already in the analytics db

### Acceptance criteria

- [x] Analytics db opportunities have data

PR #3228, opened 2024-12-16
2317/Copy data from grants-db into the opportunity table(s) in the analytics db

babebe:
## Summary
Fixes #{[2317](https://github.com/HHS/simpler-grants-gov/issues/2317)}

## Changes proposed
New Cli Function to upload opportunity tables into analytics db
S3 configuration to read `csv `opportunity tables 
Add S3 Environment variables 
Mock s3 client for testing
Added fixtures for AWS (From API code),  test-schema and opportunity-tables to aid in testing, scoped to test session.
Added fixture to delete table records after each test
Added opportunity table `csv` files for testing
Added test that checks files were successfully uploaded and records inserted into test-schema tables


analytics/src/analytics/integrations/extracts/load_opportunity_data.py
```diff
@@ -0,0 +1,58 @@
...
+    MAP_TABLES_TO_COLS,
+    OpportunityTables,
+)
+from analytics.integrations.extracts.s3_config import S3Config, get_s3_client
+
+logger = logging.getLogger(__name__)
+
+
+def get_s3_bucket(path):  # if not in env ?
+    return urlparse(path).hostname
```
chouinar, line 16: Is this used?

chouinar (commented):
I did not realize how much DB setup was needed for the tests, I had assumed that it was much further along. I would have made sure that was setup first before anything else. Good job in figuring that out.

Happy to help get any of that test setup cleaned up, I'm very familiar with it if you have any questions.

analytics/pyproject.toml
```diff
@@ -22,10 +23,11 @@ python = "~3.13"
 slack-sdk = "^3.23.0"
 typer = { extras = ["all"], version = "^0.14.0" }
 sqlalchemy = "^2.0.30"
-psycopg = ">=3.0.7"
 pydantic-settings = "^2.3.4"
 boto3 = "^1.35.56"
 boto3-stubs = "^1.35.56"
+psycopg = "^3.2.3"
+moto = "^5.0.22"
```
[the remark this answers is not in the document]
babebe: moved to dev dependencies

analytics/src/analytics/integrations/extracts/load_opportunity_data.py
babebe: Nope, not if we have it as env var. I will remove it.

analytics/tests/integrations/extracts/test_load_opportunity_data.py
```diff
@@ -0,0 +1,66 @@
...
+def upload_opportunity_tables_s3(mock_s3):
+
+    s3_config = S3Config()
+    s3_bucket = s3_config.s3_opportunity_bucket
+
+    s3_resource = mock_s3
+
+    # Create the S3 bucket for local-opportunity
+    s3_resource.create_bucket(Bucket=s3_bucket)
+    bucket = s3_resource.Bucket(s3_bucket)
```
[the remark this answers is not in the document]
babebe: Was going to remove that. But i could use it , and monkey path the env var here

analytics/tests/conftest.py
```diff
@@ -255,3 +263,118 @@ def issue(  # pylint: disable=too-many-locals
...
+    Creates a temporary PostgreSQL schema and creates a database engine
+    that connects to that schema. Drops the schema after the context manager
+    exits.
+    """
+    monkeypatch_session.setenv("DB_CHECK_CONNECTION_ON_INIT", "False")
+    monkeypatch_session.setenv("DB_HOST", "localhost")
+
+    etldb_conn = EtlDb()
+
+    schema = "test_schema"
```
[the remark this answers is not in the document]
babebe: Oh, that make sense! Updated with rand int !

analytics/Makefile
```diff
@@ -150,6 +150,12 @@ db-migrate:
 	$(POETRY) analytics etl db_migrate
 	@echo "====================================================="
```
coilysiren, line 151: The === situation here is messy IMO. But valid to keep the pattern

analytics/pyproject.toml
```diff
@@ -9,6 +9,7 @@ version = "0.1.0"
 [tool.poetry.scripts]
 analytics = "analytics.cli:app"
 db-migrate = "analytics.cli:migrate_database"
+opportunity-load = "analytics.cli:ingest_opportunity_data"
```
coilysiren, line 12: Not necessary, we can call the same function via app entrypoint. I don't remember how to do that offhand though

analytics/src/analytics/integrations/extracts/constants.py
```diff
@@ -0,0 +1,99 @@
...
+    "UPDATED_AT",
+)
+
+MAP_TABLES_TO_COLS: dict[OpportunityTables, list[str]] = {
+    OpportunityTables.LK_OPPORTUNITY_STATUS: LK_OPPORTUNITY_STATUS_COLS,
+    OpportunityTables.LK_OPPORTUNITY_CATEGORY: LK_OPPORTUNITY_CATEGORY_COLS,
+    OpportunityTables.OPPORTUNITY: OPPORTUNITY_COLS,
+    OpportunityTables.OPPORTUNITY_SUMMARY: OPOORTUNITY_SUMMARY_COLS,
+    OpportunityTables.CURRENT_OPPORTUNITY_SUMMARY: CURRENT_OPPORTUNITY_SUMMARY_COLS,
+}
```
coilysiren, line 99: Keeping this in sync sounds annoying. Nothing to do about that now though

analytics/pyproject.toml
chouinar: I think what is already setup in the Makefile makes it work, the makefile isn't even running `poetry opportunity-load` - I think this line could just be deleted

analytics/src/analytics/integrations/etldb/etldb.py
```diff
@@ -128,3 +130,18 @@ class EtlChangeType(Enum):
...
+    @wraps(func)
+    def wrapper(*args: Any, **kwargs: Any):
+        etldb_conn = kwargs.get("etldb_conn")
+        if not etldb_conn:
+            etldb_conn = EtlDb()
+            kwargs.update({"etldb_conn": etldb_conn})
+
+        return func(*args, **kwargs)
+
+    return wrapper
```
chouinar, line 147: While this util is nice to have, I might suggest not doing it in this PR as it'll probably be 30+ lines of type issues to sort through.

I'd suggest just initializing the DB connection like other jobs do for now

babebe: Removed. Main func now passes the connection to a helper func where the bulk of the code runs.

coilysiren (approved):

analytics/tests/conftest.py
```diff
@@ -255,3 +263,118 @@ def issue(  # pylint: disable=too-many-locals
...
+        sql_file_path = (
+            test_file_path.parent.parent
+            / "src"
+            / "analytics"
+            / "integrations"
+            / "etldb"
+            / "migrations"
+            / "versions"
+            / "0006_add_opportunity_tables.sql"
+        )
```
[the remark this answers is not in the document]
babebe: The migration currently runs the latest migration file  only.

analytics/src/analytics/integrations/extracts/load_opportunity_data.py
```diff
@@ -0,0 +1,66 @@
...
+                        f"{s3_config.load_opportunity_data_file_path}/{table}.csv",
+                        "r",
+                    ),
+                )
+                copy = stack.enter_context(cursor.copy(query))
+
+                while data := file.read():
+                    copy.write(data)
+
+            logger.info("Successfully loaded data for table: %s", table)
```
chouinar, line 66: There is one thing missing - we need to delete the prior data from each table first.

If you run the job twice it errors because of duplicate key errors:
```
UniqueViolation: duplicate key value violates unique constraint "lk_opportunity_status_pkey"
DETAIL:  Key (opportunity_status_id)=(1) already exists.
CONTEXT:  COPY lk_opportunity_status, line 2
```

babebe: Ah, thank you ! Fixed with a fixture to delete records prior each test

chouinar: This isn't a test problem - this is a thing you need to implement right here, not as a fixture

analytics/src/analytics/integrations/extracts/load_opportunity_data.py
```diff
@@ -0,0 +1,79 @@
...
+
+        _fetch_insert_opportunity_data(conn)
+
+    logger.info("Extract opportunity data completed successfully")
+
+
+def _trancate_opportunity_table_records(conn: Connection) -> None:
+    """Truncate existing records from all tables."""
+    cursor = conn.connection.cursor()
+    schema = os.environ["DB_SCHEMA"]
```
chouinar, line 46: Do you need to specify the schema? Since this sets the search_path, I think this would implicitly affect the below command (ie. you could just write that as `TRUNCATE TABLE {table} CASCADE`)

chouinar (approved):
LGTM!

analytics/src/analytics/integrations/extracts/load_opportunity_data.py
babebe: Just out of caution. Since it will default to public schema if i leave it out.

PR #3229, opened 2024-12-16
[Unticketed] Update package.json to remove dep i18n

acouch:
## Summary
Fixes NA

## Changes proposed
Remove deprecated i18n packages.


doug-s-nava (approved):
everything seems to run fine

✅ ✅ ✅ ✅ ✅ 

I do wonder what plans, if any, we have for translations on storybook though. I assume the idea was that the `storybook-react-i18next` package was going to be providing something there, and I don't see anything else lined up to take its place...

acouch:
@doug-s-nava good question. Looks like this might be the way to go with that: https://github.com/stevensacks/storybook-next-intl I created a ticket: https://github.com/HHS/simpler-grants-gov/issues/3236

PR #3273, opened 2024-12-17
[Issue #3183] Make search facets for agencies return more than 25 agencies

### LINKED ISSUE #3183: Modify search facet count logic to allow more than 25 values in the response
### Summary

When we query OpenSearch we tell it what facets (or aggregations as they call them) to return counts for. We also say the maximum number of facets to return and default that to 25.

We'll want to adjust that for agency_code as we can return several hundred of those.

https://github.com/HHS/simpler-grants-gov/blob/main/api/src/services/opportunities_v1/search_opportunities.py#L133 is where we specify it, the `aggregation_terms` method takes in a size and defaults to 25.

Open Question:
* Is there an max amount we want returned, or should we just set the number to something high like 1000 (for context, about 700 agencies have at least one opportunity at the time of writing)
* Do we want to do anything with [shard size](https://opensearch.org/docs/latest/aggregations/bucket/terms/#size-and-shard-size-parameters)?
* Do we want to configure any of the other options for aggregations?

### Acceptance criteria

- [ ] Adjustments made after figuring out open questions

chouinar:
## Summary
Fixes #3183

## Changes proposed
Modify the search facet count logic to allow up to 1000 unique agencies in response

## Context for reviewers
By default we configured the size to be 25 arbitrarily. There are about 700 agencies with opportunities in prod, so configured a number a bit above that for now.

We might want to consider allowing the user of the API to specify the facets they want in the future.

## Additional information
I loaded a lot of opportunities into the search index, and the response for facet counts now has all of the agencies we have configured locally with is about 80.

### DIFF
--- a/api/src/services/opportunities_v1/search_opportunities.py
+++ b/api/src/services/opportunities_v1/search_opportunities.py
@@ -137,7 +137,7 @@ def _add_aggregations(builder: search.SearchQueryBuilder) -> None:
     builder.aggregation_terms("applicant_type", _adjust_field_name("applicant_type"))
     builder.aggregation_terms("funding_instrument", _adjust_field_name("funding_instrument"))
     builder.aggregation_terms("funding_category", _adjust_field_name("funding_category"))
-    builder.aggregation_terms("agency", _adjust_field_name("agency_code"))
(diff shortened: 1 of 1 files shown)


mdragon (approved):
Approved!

PR #3281, opened 2024-12-18
[Issue #2653] Sign in component

### LINKED ISSUE #2653: Task: Implement header sign in component
### Summary

Implement header sign in component based off approved design.

https://www.figma.com/design/FGxWtAgToKhehLJCiuy1zL/Simpler.Grants.gov-Project?node-id=963-1474&node-type=frame&t=n9P2w2tLmUNok9cZ-0

Image

Note that the designs show a logged in state for the component as well, confirming requirements for that state.

### Acceptance criteria

Component created and tested

acouch:
## Summary
Fixes #2653

#### Figma
image

#### Local
image

#### Figma
image

#### Local
image


mdragon (approved):

infra/frontend/app-config/env-config/environment-variables.tf
```diff
@@ -44,6 +44,11 @@ locals {
       manage_method     = "manual"
       secret_store_name = "/${var.app_name}/${var.environment}/api-auth-token"
     },
+    # URL for the API login route. 
+    AUTH_LOGIN_URL = {
```
coilysiren, line 48: This isn't secret so it goes next to the NODE_OPTIONS key above

acouch: It is the same as the `API_URL` right?

frontend/src/components/Header.tsx
```diff
@@ -92,6 +95,7 @@ const NavLinks = ({
           className={clsx({
             "usa-nav__link": true,
             "usa-current": currentNavItemIndex === index,
+            "text-primary": link.textPrimary ?? false,
```
doug-s-nava, line 98: I don't see textPrimary being set anywhere

frontend/src/styles/_uswds-theme-custom-styles.scss
```diff
@@ -30,3 +30,8 @@ i.e.
     @include u-shadow(4);
   }
 }
+@include at-media("desktop") {
```
doug-s-nava, line 33: Imagine you've tried some things out, but is there a way to avoid this? Can we mirror styling of the navbar items, or is there something throwing this off that is too difficult to deal with?

frontend/src/components/Header.tsx
```diff
@@ -146,6 +150,28 @@ const Header = ({ logoPath, locale }: Props) => {
   const title =
     usePathname() === "/" ? t("title") : <Link href="/">{t("title")}</Link>;
 
+  const LoginLink = () => {
+    return (
+      <Link
+        href={process.env.auth_login_url as string}
+        key={process.env.auth_login_url as string}
+        className={clsx({
```
doug-s-nava, line 158: if all of these are always true we probably don't need to use `clsx` here

doug-s-nava (commented):
can we confirm that the sign in button should get the same hover indicator as the nav items? seems slightly weird to me to have two items showing the same indicator at the same time:
Screenshot 2024-12-18 at 11 59 14 AM

doug-s-nava:
Also if we're not updating the button to a logged in state when user is logged in in this change I think we need a ticket to handle that

frontend/.env.production
```diff
@@ -11,4 +11,5 @@ SENDY_API_KEY=
 SENDY_API_URL=
 SENDY_LIST_ID=
 
-API_URL=http://api.simpler.grants.gov
+API_URL=https://api.simpler.grants.gov
+AUTH_LOGIN_URL=https://api.simpler.grants.gov/v1/users/login
```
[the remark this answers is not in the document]
mdragon: Yeah, I guess we have the API one to point to that the behavior is "as expected" but that likely also reinforces that they don't need to be here?

infra/frontend/app-config/env-config/environment-variables.tf
mdragon: @coilysiren Secrets knows how to read from SSM where as the standard env vars are more "constants?" Right? Or is there a way to read from SSM there too?

frontend/src/components/Header.tsx
```diff
@@ -146,6 +150,28 @@ const Header = ({ logoPath, locale }: Props) => {
   const title =
     usePathname() === "/" ? t("title") : <Link href="/">{t("title")}</Link>;
 
+  const LoginLink = () => {
+    return (
+      <Link
+        href={process.env.auth_login_url as string}
```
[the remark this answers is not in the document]
mdragon: I wish local Next did a better job of highlighting/blocking this anti-pattern.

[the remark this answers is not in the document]
acouch: @doug-s-nava b/c these pages are statically rendered, there isn't a server side place to put the env for the static pages, unless we use wrap them in a suspense boundary. @mdragon that isn't an anti-pattern, it is the rec from [next documentation](https://nextjs.org/docs/14/app/building-your-application/configuring/environment-variables) 

That could be the way to go. As of this comment, the CDN is rendering the suspense on the search page without a loading state, meaning the header could potentially use suspense without a "flicker."

If we switch to dynamic rendering, then we could just use this pattern as is.

frontend/next.config.js
```diff
@@ -68,6 +68,9 @@ const nextConfig = {
       },
     ];
   },
+  env: {
+    auth_login_url: process.env.AUTH_LOGIN_URL,
```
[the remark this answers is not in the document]
acouch: The same thing the `src/constants/environments.ts` is doing. Initially I had this as a `NEXT_PUBLIC_` var and that file says `home for all interpreted server side environment variables`. I'll move this over there, and update that comment since `NEXT_PUBLIC` files are defined there anyway. Good catch, thanks.

frontend/tests/components/Header.test.tsx
```diff
@@ -10,6 +10,8 @@ const props = {
   locale: "en",
 };
 
+process.env.auth_login_url = "/login-url";
```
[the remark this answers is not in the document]
acouch: Seeing that there is a [`mockProcessEnv()`](https://github.com/HHS/simpler-grants-gov/blob/main/frontend/src/utils/testing/commonTestUtils.ts) function we aren't using and an example of mocking the `environment.ts` file in [FeatureFlagManager.test.ts](https://github.com/HHS/simpler-grants-gov/blob/main/frontend/tests/services/FeatureFlagManager.test.ts#L14C1-L15C1) which I started off as not using, so no clear pattern yet. I'll use the `environments.ts` file and follow the pattern in the FeatureFlagManager.test.ts file. Good catch, thanks.

frontend/src/styles/_uswds-theme-custom-styles.scss
acouch: There is no 5px margin from [USWDS](https://designsystem.digital.gov/utilities/margin-and-padding/). The extra pixel did make a difference IMHO.

frontend/src/components/Header.tsx
acouch: Updated, thanks.

acouch:
> Also if we're not updating the button to a logged in state when user is logged in in this change I think we need a ticket to handle that

@doug-s-nava this seemed like enough to chew on for a PR and the provider wasn't available when I started. I can do a follow-up for the same ticket now that the user provider is available.

frontend/.env.production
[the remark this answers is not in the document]
mdragon: It appears this gets picked up at build time, even though it's in a client component, because... I'm not sure how or why? Which also makes me question how staging is not finding Prod, as at that build it definitely doesn't have access to the SSM stuff (and I can't double check what SSM has for Stage cause I can't SSO into AWS right now).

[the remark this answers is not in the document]
doug-s-nava: I've unfortunately fought this stuff a bunch with the new relic work. It might be good to talk this out if folks have time

frontend/src/components/Header.tsx
```diff
@@ -15,9 +15,12 @@ import {
   Header as USWDSHeader,
 } from "@trussworks/react-uswds";
 
+import { USWDSIcon } from "./USWDSIcon";
```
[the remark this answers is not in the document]
acouch: Updated, thanks.

frontend/next.config.js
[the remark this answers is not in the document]
acouch: I removed this.

frontend/tests/components/Header.test.tsx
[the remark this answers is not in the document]
acouch: I updated this.

frontend/src/components/Header.tsx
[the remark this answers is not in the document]
acouch: Noting we resolved this, decided to use a route handler for the public client side envars for the time being then will remove that once we can use dynamic routing.

frontend/.env.production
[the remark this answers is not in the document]
acouch: I removed these and made the file blank after discussion.

frontend/src/components/Header.tsx
```diff
@@ -146,6 +161,25 @@ const Header = ({ logoPath, locale }: Props) => {
   const title =
     usePathname() === "/" ? t("title") : <Link href="/">{t("title")}</Link>;
 
+  const LoginLink = ({ auth_login_url }: { auth_login_url: string | null }) => {
+    const passHref = auth_login_url ? { href: auth_login_url } : "";
+
+    return (
+      <a
+        {...passHref}
```
acouch, line 168: This is the easiest way I could think of to disable the login button while it fetches.

frontend/src/app/api/env/route.ts
```diff
@@ -0,0 +1,11 @@
+import { environment } from "src/constants/environments";
+
+export const dynamic = "force-dynamic";
+
+export function GET() {
+  return Response.json({
+    api_url: environment.API_URL,
+    auth_login_url: environment.AUTH_LOGIN_URL,
+    node_env: environment.NODE_ENV,
```
mdragon, line 9: I don't see this being consumed anywhere yet, but per the discussion on the other ticket, should this use the new ENVIRONMENT since NODE_ENV is just prod everywhere except localhost?

frontend/src/components/Header.tsx
mdragon: And just to make sure I'm keeping all the different tickets straight, Doug has a separate ticket to hide this button behind a feature flag?

frontend/src/app/api/env/route.ts
acouch: The `NODE_ENV` is different than the env we set per environment. I could remove this as it isn't being used.

acouch: I removed it

frontend/src/components/Header.tsx
acouch: yes

frontend/src/components/Header.tsx
```diff
@@ -178,11 +211,19 @@ const Header = ({ logoPath, locale }: Props) => {
...
+          <div className="usa-navbar order-last desktop:display-none">
             <NavMenuButton
               onClick={handleMobileNavToggle}
               label={t("nav_menu_toggle")}
+              className="usa-menu-btn"
             />
           </div>
+          <div className="usa-nav__primary margin-top-0 margin-bottom-1 desktop:margin-bottom-5px text-no-wrap desktop:order-last margin-left-auto">
+            <div className="usa-nav__primary-item border-0">
+              <LoginLink auth_login_url={authLoginUrl} />
```
doug-s-nava, line 224: can we commit to camelCase?

frontend/src/components/Header.tsx
doug-s-nava: what about moving the fetch and state management into this component, and implementing a loading state (which in this case would just hide the link, I guess?)

doug-s-nava (commented):
a couple of nits

frontend/src/app/api/env/route.ts
```diff
@@ -0,0 +1,9 @@
+import { environment } from "src/constants/environments";
+
+export const dynamic = "force-dynamic";
+
+export function GET() {
+  return Response.json({
```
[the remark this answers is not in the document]
acouch: I'd like to keep the same pattern of getting the envar itself and pass to the component which is what we'll be moving to.

acouch:
> what about moving the fetch and state management into this component

@doug-s-nava want to avoid a loading state for every page request. this way the user just gets the link, and it isn't available if there is some server issue.

frontend/src/components/Header.tsx
acouch: Updated, thanks.

frontend/src/components/Header.tsx
```diff
@@ -119,11 +119,40 @@ const NavLinks = ({
   );
 };
 
+const LoginLink = ({ navLoginLinkText }: { navLoginLinkText: string }) => {
+  const [authLoginUrl, setAuthLoginUrl] = useState<string | null>(null);
```
doug-s-nava, line 123: no need to make a change, but for future reference this could be simplified since useState should always accept an undefined initial value
```
const [authLoginUrl, setAuthLoginUrl] = useState<string>();
```

doug-s-nava (approved):
✅ ✅ ✅ ✅ 

lgtm!

infra/frontend/app-config/env-config/environment-variables.tf
coilysiren: 👀

PR #3282, opened 2024-12-18
[Issue #3223] style beta alert to match brand

### LINKED ISSUE #3223: [Brand FE] BetaAlert
### Summary

Update the BetaAlert component to match the [Figma](https://www.figma.com/design/FGxWtAgToKhehLJCiuy1zL/Simpler.Grants.gov-Project?node-id=1467-5629&node-type=frame&t=InZzQfGYsDVKfze4-0). 

Update the Alert color settings to match the brand. 

The BetaAlert should no longer be sticky (the Header is sticky now). It should scroll out of view with the other content. 

The BetaAlert should no longer span the full-width of the viewport. It can be contained to the width of the content. It is currently using a custom component to achieve this full-width layout. We can probably switch to just using the Truss Alert component and deprecate the custom BetaAlert component. 

**Target the [feature/brand](https://github.com/HHS/simpler-grants-gov/tree/feature/brand) branch**

### Acceptance criteria

- [ ] Alert is not sticky
- [ ] Alert is width of content
- [ ] Alert colors updated to match the brand

andycochran:
## Summary
Fixes #3223

## Changes proposed
- BetaAlert is no longer full-width (replace FullWidthAlert with USWDSAlert)
- Update theme warning colors settings 
- Update theme alert settings 

## Context for reviewers
The new branded alert is not full-width. It is no longer sticky (changed when Header was made sticky). It is quite prominent on all pages. We may iterate on this by making it dismissible, which would require something like setting a global state to show/hide the BetaAlert on all pages. 

## Additional information

image

image

image

image

image


frontend/src/components/BetaAlert.tsx
```diff
@@ -1,18 +1,28 @@
 import { useTranslations } from "next-intl";
+import { GridContainer, Alert as USWDSAlert } from "@trussworks/react-uswds";
 
-import FullWidthAlert from "./FullWidthAlert";
+type Props = {
+  containerClasses?: string;
```
andycochran, line 5: Added this so that each instance of the banner can be spaced appropriately on its page with utility classes for e.g. margin

frontend/src/components/BetaAlert.tsx
```diff
@@ -1,18 +1,28 @@
...
   const t = useTranslations("Beta_alert");
   const alert = t.rich("alert", {
     LinkToGrants: (content) => <a href="https://www.grants.gov">{content}</a>,
   });
 
   return (
-    <div data-testid="beta-alert">
-      <FullWidthAlert type="info" heading={t("alert_title")}>
-        {alert}
-      </FullWidthAlert>
```
andycochran, line 15: No longer need the FullWidthAlert for the BetaAlert. We might deprecate this component in the future. But we can leave it for now in case we need it for other purposes.

doug-s-nava (approved):
looks good (much better) to me!!!

✅ ✅ ✅ ✅ ✅ ✅ ✅ 

also, probably unrelated but the color of `(optional)` on the subscribe page looks weird

Screenshot 2024-12-18 at 1 02 48 PM

PR #3310, opened 2024-12-19
[Unticketed] Ignore vulnerability for issue with pending fix

chouinar:
## Changes proposed
Ignore [GHSA-v778-237x-gjrc](https://github.com/advisories/GHSA-v778-237x-gjrc)

Cleaned up old ignores that _should_ be fixed by now

## Context for reviewers
This vulnerability is in a dependency pulled in by the Github CLI. A [fix was made](https://github.com/cli/cli/commit/1af421012e5e7a648577e091e5a094a80b75d720) but no new release has occurred, likely due to the holidays. https://github.com/cli/cli/releases

As this vulnerability already would exist in our image, ignoring it for now seems uneventful, and the issue shouldn't persist beyond the holidays.

### DIFF
--- a/.grype.yml
+++ b/.grype.yml
@@ -19,7 +19,7 @@ ignore:
   - fix-state: wont-fix
   - fix-state: unknown
 
-  # https://github.com/HHS/simpler-grants-gov/issues/2582
-  - vulnerability: CVE-2024-34158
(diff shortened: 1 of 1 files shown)


mdragon (approved):
Approving this, pending the clean vuln scans

Issue #2216, opened 2024-09-25
Setup tables for attachments in the staging and foreign table schemas


chouinar:
### Summary

We want to setup the staging & foreign table for the following tables that we are going to transform into our database:

Tables:
* TSYNOPSISATTACHMENT
* TSYNOPSISATTACHMENTFOLDER - Not needed anymore
* ~TSYNOPSIS_DOC_URL~ We're holding off on this table and the one below due to some schema issues we need to resolve (TSYN_ATT_CHANGE_COMMENTS has no primary key)
* ~TSYN_ATT_CHANGE_COMMENTS~

IMPORTANT: The field names **must** be defined in the exact order as given, we cannot rearrange them as the order is used by the process that copies from the foreign tables.

These tables can be defined at the same time as we define an identical base schema for our staging and foreign tables that the specific tables inherit from. We should create a new file called "attachment.py` in both the foreign, and staging table folders for these models.

## Mappings
Note that we have it configured so the python types for each of these columns are:
* int maps to BIG INT (eg. `my_field: Mapped[int]`)
* str maps to TEXT (eg. `my_field: Mapped[str]`)
* datetime maps to TIMESTAMP WITH TIMEZONE (eg. `my_field: Mapped[datetime]`)
* Nullability can be defined based on whether the python type is nullable (`Mapped[str | None]` vs `Mapped[str]`)
* We don't have any current mapping for the BYTEA type, we likely want that to map to `bytes` in python (which should be the default, so `my_field: Mapped[bytes]`) but you'll need to test this is the case.
```
=========================
tsynopsisattachment
=========================
syn_att_id: BIG INT PRIMARY KEY
opportunity_id: BIG INT NOT NULL
att_revision_number: BIG INT
att_type: TEXT
mime_type: TEXT
link_url: TEXT
file_name: TEXT
file_desc: TEXT
file_lob: BYTEA
file_lob_size: BIG INT
create_date: TIMESTAMP WITH TIMEZONE
created_date: TIMESTAMP WITH TIMEZONE
last_upd_date: TIMESTAMP WITH TIMEZONE
creator_id: TEXT
last_upd_id: TEXT
syn_att_folder_id: BIG INT


=========================
tsynopsisattachmentfolder
=========================
syn_att_folder_id: BIG INT PRIMARY KEY
opportunity_id: BIG INT NOT NULL
folder_type: TEXT NOT NULL
folder_name: TEXT
[...]

PR #3313, opened 2024-12-19
[2216] setup staging foreign tables

babebe:
## Summary
Fixes #{[2216](https://github.com/HHS/simpler-grants-gov/issues/2216)}

## Changes proposed
Mixin, Staging and Foreign attachment models created 
Migration file created

## Context for reviewers
Run migration. new table should be created in staging schema


api/src/db/models/foreign/attachement.py
chouinar, line 1: Typo in file name, should be `attachment.py`

api/src/db/models/foreign/attachement.py
```diff
@@ -0,0 +1,14 @@
...
+# match by oracle_fdw, but we are matching them for maintainability.
+#
+
+from src.db.models.legacy_mixin import synopsis_mixin
+
+from . import foreignbase
+
+
+class TsynopsisAttachment(foreignbase.ForeignBase, synopsis_mixin.TsynopsisAttachmentMixin):
+    __tablename__ = "tsynopsis_attachment"
```
chouinar, line 14: This needs to be exactly `tsynopsisattachment` - we keep the names 100% matching how they are in Oracle for simplicity, even when it isn't what we might want.

api/src/db/models/staging/attachment.py
```diff
@@ -0,0 +1,6 @@
+from src.db.models.legacy_mixin import synopsis_mixin
+from src.db.models.staging.staging_base import StagingBase, StagingParamMixin
+
+
+class TsynopsisAttachment(StagingBase, synopsis_mixin.TsynopsisAttachmentMixin, StagingParamMixin):
+    __tablename__ = "tsynopsis_attachment"
```
chouinar, line 6: Also needs no underscore
```suggestion
    __tablename__ = "tsynopsisattachment"
```

babebe: Fixed

api/src/db/models/foreign/attachement.py
babebe: Fixed

api/src/db/models/foreign/attachement.py
babebe: fixed

api/src/db/models/staging/attachment.py
chouinar: Still needs to be fixed here

chouinar (approved):
LGTM!

Issue #2818, opened 2024-11-12
Spike: Investigate AuthZ libraries for Python


chouinar:
### Summary

We want a library that makes configuring AuthZ rules simple for our endpoints (user X can use resource Y).

We've used [flask-bouncer](https://github.com/bouncer-app/flask-bouncer) in some of our past projects, but it has not been updated in 5 years, and depends on a library that is 10 years stale. We should see what alternatives exist.

Ideally we want something that lets us pass in:
* A user
* An object
* An action (eg. READ, WRITE, etc.)

And handles checking that the user can do that operation. Ideally it would have a bit more usage out there.

Optionally, consider porting some of the flask bouncer logic to our code with a bit of cleanup (eg. adding python typing).

### Acceptance criteria

- [ ] Libraries investigated and chosen

chouinar:
A few scenarios we'd want to account for (might not exactly match what we'd do, but just to see if a library could be used):
* An ADMIN role should be able to access most things (probably rare cases where we don't want this)
* A user can read/write anything that belongs to them based on the user_id (application, saved searches, user info endpoints)
* A user's access to something in an organization they belong to would depend on their roles in that org

Likely there will be a few different types of "roles". There are user types (system admin vs generic user), and there are roles within an organization (admin for agency XYZ or read-only in agency ABC).

PR #3330, opened 2024-12-20
[Issue #3293] Create POST /users/:userId/saved-opportunities API schema and stub endpoint

### LINKED ISSUE #3293: Create POST /users/:userId/saved-opportunities API schema and stub endpoint
### Summary

For now, the request body will just be the opportunity ID, we'll likely expand this in the future.

```json
{
   "opportunity_id": 1234
}
```

The response body will just be a basic empty-ish response with a success message (ie. inherit from the abstract response schema with nothing in it).

### Acceptance criteria

- [ ] Schema created
- [ ] Empty API which returns

mikehgrantsgov:
## Summary
Fixes #3293 

## Changes proposed
Create an API for users to save opportunities.
Make sure this API is scoped to the currently logged in user
Response body a basic success message

## Additional information
See attached unit tests


api/src/api/users/user_routes.py
```diff
@@ -146,3 +148,39 @@ def user_get(db_session: db.Session, user_id: UUID) -> response.ApiResponse:
...
+def user_save_opportunity(
+    db_session: db.Session, user_id: UUID, json_data: dict
+) -> response.ApiResponse:
+    logger.info("POST /v1/users/:user_id/saved-opportunities")
+
+    user_token_session: UserTokenSession = api_jwt_auth.current_user  # type: ignore
+
+    # Verify the authenticated user matches the requested user_id
+    if user_token_session.user_id != user_id:
+        raise_flask_error(401, "Unauthorized user")
```
chouinar, line 168: Completely unrelated to your PR, but if you're looking for something to work on while folks are out, I'd like if we had an AuthZ library or approach that was a bit more sophisticated than this once we have roles: https://github.com/HHS/simpler-grants-gov/issues/2818

Mass PFML had a setup but I looked into the library used and it was not maintained anymore.

api/tests/src/api/users/test_user_save_opportunity_post.py
```diff
@@ -0,0 +1,97 @@
...
+from src.auth.api_jwt_auth import create_jwt_for_user
+from src.db.models.user_models import UserSavedOpportunity
+from tests.src.db.models.factories import OpportunityFactory, UserFactory
+
+
+def test_user_save_opportunity_post_unauthorized_user(app, enable_factory_create, db_session):
+    # Create a user and get their token
+    user = UserFactory.create()
+    token, _ = create_jwt_for_user(user, db_session)
+    db_session.commit()
```
chouinar, line 12: We could use a test utility that is like "get_test_user" that we'll eventually be able to pass things like roles into for a given unit test. Might be worth looking into setting that up.

api/tests/src/api/users/test_user_save_opportunity_post.py
```diff
@@ -0,0 +1,97 @@
...
+
+    # Create an opportunity
+    opportunity = OpportunityFactory.create()
+
+    # Try to save an opportunity for a different user ID
+    different_user_id = uuid.uuid4()
+    response = app.test_client().post(
```
chouinar, line 19: Instead of app, we can pull in client directly
```suggestion
    response = client.post(
```

api/src/api/users/user_routes.py
mikehgrantsgov: Good idea, will take a look!

api/tests/src/api/users/test_user_save_opportunity_post.py
mikehgrantsgov: I did abstract this slightly to remove duplication. I suppose the fixture I added could be added to conftest?

chouinar (approved):
LGTM!

PR #3355, opened 2024-12-23
[Issue #3295] Create GET /users/:userId/saved-opportunities API schema and stub endpoint

### LINKED ISSUE #3295: Create GET /users/:userId/saved-opportunities API schema and stub endpoint
### Summary

There is no request body, the user ID is in the URL.

---

We want to return a list response like so:

```json
{
  "data": [
    {
       "opportunity_id": 1234,
       "opportunity_title": "my title",
       "opportunity_status": "posted",
       "summary": {
            "post_date": "2024-12-19",
            "close_date": "2024-12-19",
            "is_forecast": false
       }
    }
  ]
} 
```
We will likely adjust the schema going forward once design has figured out what they want to display, but we want the data to be a list of "mini" opportunity objects. We do not want the full opportunity, but just a few top fields that a user has. For now we'll just put the ones above

### Acceptance criteria

- [ ] Schema created
- [ ] Dummy endpoint which returns a static response created

mikehgrantsgov:
## Summary
Fixes #3295

## Changes proposed
Create GET /users/:userId/saved-opportunities API schema

## Context for reviewers
Overlaps with the implementation ticket, as it was easier to add the service and implementation vs. stub + implement in 2 different PRs

## Additional information
See attached unit tests.


api/src/api/opportunities_v1/opportunity_schemas.py
```diff
@@ -528,3 +528,29 @@ class OpportunitySearchResponseV1Schema(AbstractResponseSchema, PaginationMixinS
...
+class SavedOpportunityResponseV1Schema(AbstractResponseSchema):
+    opportunity_id = fields.Integer(
+        metadata={"description": "The ID of the saved opportunity", "example": 1234}
+    )
+    opportunity_title = fields.String(
+        metadata={"description": "The title of the opportunity", "example": "my title"}
+    )
+    opportunity_status = fields.String(
+        metadata={"description": "Current status of the opportunity", "example": "posted"}
+    )
```
chouinar, line 554: Keep enums as enums
```suggestion
    opportunity_status = fields.Enum(
        OpportunityStatus,
        metadata={
            "description": "The current status of the opportunity",
            "example": OpportunityStatus.POSTED,
        },
    )
```

api/src/services/users/get_saved_opportunities.py
```diff
@@ -0,0 +1,20 @@
...
+from uuid import UUID
+
+from src.adapters import db
+from src.db.models.opportunity_models import Opportunity
+from src.db.models.user_models import UserSavedOpportunity
+
+logger = logging.getLogger(__name__)
+
+
+def get_saved_opportunities(db_session: db.Session, user_id: UUID) -> List[UserSavedOpportunity]:
```
chouinar, line 12:
```suggestion
from uuid import UUID

from src.adapters import db
from src.db.models.opportunity_models import Opportunity
from src.db.models.user_models import UserSavedOpportunity

logger = logging.getLogger(__name__)


def get_saved_opportunities(db_session: db.Session, user_id: UUID) -> list[UserSavedOpportunity]:
```

api/src/services/users/get_saved_opportunities.py
```diff
@@ -0,0 +1,20 @@
...
+    logger.info(f"Getting saved opportunities for user {user_id}")
+
+    return (
+        db_session.query(UserSavedOpportunity)
+        .join(Opportunity)
+        .filter(UserSavedOpportunity.user_id == user_id)
+        .all()
+    )
```
chouinar, line 20: We should use the newer `db_session.execute(select(...` approach - while query isn't deprecated yet, it's considered the old-style for SqlAlchemy. Can just copy from one of the other service methods.

api/src/api/users/user_routes.py
```diff
@@ -184,3 +186,50 @@ def user_save_opportunity(
...
+                "opportunity_id": opp.opportunity_id,
+                "opportunity_title": opp.opportunity_title,
+                "opportunity_status": (
+                    opp.opportunity_status.value if opp.opportunity_status else None
+                ),
+                "summary": summary_data,
+            }
+        )
+
+    return {"message": "Success", "data": opportunities_data}
```
chouinar, line 235: We shouldn't be returning a dict, return an ApiResponse object

api/src/api/users/user_routes.py
```diff
@@ -184,3 +186,50 @@ def user_save_opportunity(
...
+        raise_flask_error(401, "Unauthorized user")
+
+    # Get all saved opportunities for the user with their related opportunity data
+    saved_opportunities = get_saved_opportunities(db_session, user_id)
+
+    # Format the response data
+    opportunities_data = []
+    for saved in saved_opportunities:
+        opp = saved.opportunity
+        summary = opp.summary
```
chouinar, line 212: This loop shouldn't be necessary, it should be fine to return the ORM objects and Marshmallow will handle iterating over/making them into JSON for you. If you're hitting an error about objects being detached from a session, then that's because SQLAlchemy lazy-loads all relationships (this loop effectively emits several SQL queries). Easiest way around that is to load in the query itself. It's why in an endpoint like `GET opportunity` we do `.options(selectinload("*"))`: https://github.com/HHS/simpler-grants-gov/blob/main/api/src/services/opportunities_v1/get_opportunity.py#L21

api/tests/src/api/users/test_user_saved_opportunities_get.py
```diff
@@ -0,0 +1,78 @@
...
+from src.auth.api_jwt_auth import create_jwt_for_user
+from src.db.models.user_models import UserSavedOpportunity
+from tests.src.db.models.factories import OpportunityFactory, UserFactory
+
+
+@pytest.fixture
+def user(enable_factory_create, db_session):
+    user = UserFactory.create()
+    db_session.commit()
+    return user
```
chouinar, line 12: Factories commit when you call create, we can remove the commit and just do 
```suggestion
    return UserFactory.create()
```

api/tests/src/api/users/test_user_saved_opportunities_get.py
```diff
@@ -0,0 +1,78 @@
...
+def user_auth_token(user, db_session):
+    token, _ = create_jwt_for_user(user, db_session)
+    return token
+
+
+@pytest.fixture(autouse=True)
+def clear_opportunities(db_session):
+    db_session.query(UserSavedOpportunity).delete()
+    db_session.commit()
+    yield
```
chouinar, line 25: I'd caution against making this an autouse that runs for every test as it adds several seconds per test, either make it explicitly used by tests in this module or do something like  `scope="module"`

api/tests/src/api/users/test_user_saved_opportunities_get.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def test_user_get_saved_opportunities(
+    client, user, user_auth_token, enable_factory_create, db_session
+):
+    # Create an opportunity and save it for the user
+    opportunity = OpportunityFactory.create()
+    saved = UserSavedOpportunity(user_id=user.user_id, opportunity_id=opportunity.opportunity_id)
+    db_session.add(saved)
+    db_session.commit()
```
chouinar, line 35: Make a factory for creating these instead to let this be a one-liner without add/commits necesary

mikehgrantsgov: Thanks, this was added

api/tests/src/api/users/test_user_saved_opportunities_get.py
mikehgrantsgov: I added scope="function" which I think should do the trick.

api/src/services/users/get_saved_opportunities.py
mikehgrantsgov: Updated this, thanks

api/src/api/users/user_routes.py
mikehgrantsgov: Yep that makes sense. I thought we had to do more work here than needed. I did update the utility to return pure `Opportunity` objects so it will fit in the shape of the response.

PR #3358, opened 2024-12-24
[Issue #3296] Create DELETE /`users/:userId/saved-opportunities/:opportunityId` API schema and endpoint

### LINKED ISSUE #3296: Create DELETE /users/:userId/saved-opportunities/:opportunityId API schema and stub endpoint
### Summary

No request body, IDs are all from the path parameters

Response body should just be a message, and we should specifically return a 200

### Acceptance criteria

- [ ] Endpoint setup, params should be parsed out
- [ ] Endpoint is a stub, no implementation yet, just return a generic success message

mikehgrantsgov:
## Summary
Fixes #3296

## Changes proposed
Add API to delete saved opportunities

## Additional information
See attached unit tests


api/src/api/users/user_routes.py
```diff
@@ -184,3 +186,31 @@ def user_save_opportunity(
...
+        raise_flask_error(401, "Unauthorized user")
+
+    with db_session.begin():
+        # Try to find and delete the saved opportunity
+        result = delete_saved_opportunity(db_session, user_id, opportunity_id)
+
+    if result == 0:
+        raise_flask_error(404, "Saved opportunity not found")
+
+    db_session.commit()
```
chouinar, line 214: This doesn't do anything as a commit automatically occurred at the end of the block with `begin()`
```suggestion

```

api/src/services/users/delete_saved_opportunity.py
```diff
@@ -0,0 +1,15 @@
...
+
+def delete_saved_opportunity(db_session: db.Session, user_id: UUID, opportunity_id: int) -> int:
+    return (
+        db_session.query(UserSavedOpportunity)
+        .filter(
+            UserSavedOpportunity.user_id == user_id,
+            UserSavedOpportunity.opportunity_id == opportunity_id,
+        )
+        .delete()
+    )
```
chouinar, line 15: Two suggestions:
* We should use the new `db_session.execute(delete(...` approach - you can get a row count from the returned object: https://docs.sqlalchemy.org/en/20/tutorial/data_update.html#getting-affected-row-count-from-update-delete
* Rather than have this method return an integer, have this method check if the row count is 0 (ie. not found) and raise the error here. Effectively, put all the DB logic in this method instead of partially here and partially in the route method

mikehgrantsgov: Makes sense, and made both of these updates!

chouinar (approved):

Issue #3166, opened 2024-12-10
Make the search index load job cleanup old search indexes


chouinar:
### Summary

Our search load job works as follows:
* Create a search index
* Load data into it
* Swap the old search index alias (the name it's found at) with the new one
* Delete the prior search index

But if we have a failure with the load partway through, the old search index remains.

We should have the process also delete other search indexes (that match the same pattern format) to avoid filling up our search index with unnecessary data.

This is only necessary on the code flow that does a full refresh, incremental is unaffected

### Acceptance criteria

- [ ] Cleanup added to job
- [ ] The index we JUST made should explicitly not be deleted

PR #3372, opened 2025-01-02
[Issue 3166] cleanup to delete all indexes using prefix

babebe:
## Summary
Fixes #{[3166](https://github.com/HHS/simpler-grants-gov/issues/3166)}


api/tests/src/search/backend/test_load_opportunities_to_index.py
```diff
@@ -137,12 +137,16 @@ def test_load_opportunities_to_index(
...
             "%Y-%m-%d_%H-%M-%S"
         )
         search_client.create_index(index_name)
         search_client.swap_alias_index(
-            index_name, load_opportunities_to_index.config.alias_name, delete_prior_indexes=True
+            index_prefix,
+            index_name,
+            load_opportunities_to_index.config.alias_name,
+            delete_prior_indexes=True,
         )
```
chouinar, line 150: We need to add a few dedicated tests for this.

We should have tests that specifically check each of the following:
* Create a few indexes with a similar name, they should have been cleaned up.
* Other indexes that might exist are not modified

api/src/adapters/search/opensearch_client.py
```diff
@@ -175,8 +180,15 @@ def swap_alias_index(
         self._client.indices.update_aliases({"actions": actions})
 
         # Cleanup old indexes now that they aren't connected to the alias
+        resp = self._client.cat.indices(format="json", h=["index"])
+        old_indexes = [
+            index["index"]
+            for index in resp
+            if index["index"].startswith(index_prefix) and index["index"] != index_name
+        ]  # omit the newly created one
+
```
chouinar, line 189: I think you can simplify this a bit, when querying OpenSearch you can specify a wildcard:

```
GET _cat/indices/opportunity-index-*
```

Should be able to do that here as well?

api/src/adapters/search/opensearch_client.py
```diff
@@ -149,7 +149,12 @@ def alias_exists(self, alias_name: str) -> bool:
         return len(existing_index_mapping) > 0
 
     def swap_alias_index(
-        self, index_name: str, alias_name: str, *, delete_prior_indexes: bool = False
+        self,
+        index_prefix: str,
```
chouinar, line 153: I might suggest making the "cleanup_old_indices" its own dedicated function we call intentionally instead of making this method do more than it already is. Seeing all the changes to the usage of this makes it seem like this is now doing too much

babebe: agreed!

api/src/adapters/search/opensearch_client.py
```diff
@@ -148,9 +148,27 @@ def alias_exists(self, alias_name: str) -> bool:
...
         return len(existing_index_mapping) > 0
 
-    def swap_alias_index(
-        self, index_name: str, alias_name: str, *, delete_prior_indexes: bool = False
+    def cleanup_old_indices(
+        self,
+        index_prefix: str,
+        index_name: str,
+        *,
+        delete_prior_indexes: bool = False,
```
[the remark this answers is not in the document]
babebe: Ah yes! removed!

chouinar (approved):

PR #3374, opened 2025-01-02
[Issue #3249] Feature flag manager refactor

### LINKED ISSUE #3249: Update client side useFeatureFlag hook to read feature flags
### Summary

The [useFeatureFlag hook](https://github.com/HHS/simpler-grants-gov/blob/main/frontend/src/hooks/useFeatureFlags.ts) doesn't directly expose any way to detect feature flag state or changes in the app.

In order to properly implement feature flagging around such things as header nav items (as will be needed for the launch of the login feature), we will need to expand this functionality.

### Acceptance criteria

- [ ] useFeatureFlags hook exposes either a static value containing state of each flag or a function that can be used to read the state of an individual flag, using the FeatureFlagManager, and thus honoring default flag values, environment variables, cookies, and search params (in ascending order of precedence)

doug-s-nava:
## Summary
Fixes #3249

### TODOS

- [ ] docs

## Changes proposed

The existing feature flag system in NextJs consists of a withFeatureFlag wrapper that is used only on the server, a useFeatureFlag hook that is used only on the client, and a FeatureFlagManager that is by both of these implementations and within NextJs middleware. The FeatureFlagManager is able to read environment variables for feature flag values, but as currently written, is unable to make these values available to client side implementations of the FeatureFlagManager.

This change rewrites the FeatureFlagManager and useFeatureFlag hook to solve this issue, primarily making everything rely on all feature flag state being maintained in cookies to allow client and server to remain in sync.

It also:

* adds the `authOff` feature flag in frontend code and terraform
* refactors the `environments` setup a bit to more easily expose feature flags
* splits functionality that does not benefit from being held in the FeatureFlagsManager class into a helper file
* moves feature flag manager file into a nested directory
* creates a placeholder "auth enabled" text in the header for testing - to be replaced with implementation from https://github.com/HHS/simpler-grants-gov/pull/3281 once that is merged

## Context for reviewers
For whoever is reviewing this, I'd appreciate an in person meeting to explain a few things and gather thoughts around the approach.

### Test Steps

1. start server on this branch with `npm run build && npm start`
3. visit http://localhost:3000
4. _VERIFY_: a sign in button appears in the header
5. visit http://localhost:3000/opportunity/33?_ff=opportunityOff:false
6. _VERIFY_: opportunity page loads
1. start a server on this branch with `npm run build && FEATURE_AUTH_OFF="true" FEATURE_OPPORTUNITY_OFF="true" npm start`
3. visit http://localhost:3000
4. _VERIFY_: no sign in button appears
5. visit http://localhost:3000/opportunity/33?_ff=opportunityOff:false
6. _VERIFY_: opportunity page redirects to maintenance
3. visit http://localhost:3000?_ff=authOff:false
4. _VERIFY_: sign in button appears in the header
5. visit http://localhost:3000/opportunity/33?_ff=opportunityOff:false
6. _VERIFY_: opportunity page loads
7. visit localhost:3000/dev/feature-flags
8. _VERIFY_: both authOff and opportunityOff are disabled
9. turn on both authOff and opportunityOff from the UI
10. refresh
11. _VERIFY_: sign in button appears
5. visit http://localhost:3000/opportunity/33
6. _VERIFY_: you are redirected to maintenance

## Additional information
With auth enabled the header should look like this

Screenshot 2025-01-02 at 3 02 18 PM


frontend/src/components/Layout.tsx
```diff
@@ -20,20 +14,14 @@ export default function Layout({ children, locale }: Props) {
...
   const t = useTranslations();
-  const messages = useMessages();
 
   return (
     // Stick the footer to the bottom of the page
     <div className="display-flex flex-column minh-viewport">
       <a className="usa-skipnav" href="#main-content">
         {t("Layout.skip_to_main")}
       </a>
-      <NextIntlClientProvider
```
doug-s-nava, line 31: this change is unrelated but at one point I was messing with this file a lot and removing this unnecessary code simplified things for me a bit

frontend/src/components/Header.tsx
```diff
@@ -27,6 +30,21 @@ type Props = {
 
 const homeRegexp = /^\/(?:e[ns])?$/;
 
+const LoginLink = () => {
```
doug-s-nava, line 33: should be removed on rebase to include actual login button

frontend/src/app/[locale]/dev/feature-flags/FeatureFlagsTable.tsx
```diff
@@ -12,13 +12,9 @@ import Loading from "src/components/Loading";
  * View for managing feature flags
  */
 export default function FeatureFlagsTable() {
-  const { featureFlagsManager, mounted, setFeatureFlag } = useFeatureFlags();
+  const { setFeatureFlag, featureFlags } = useFeatureFlags();
   const { user, isLoading, error } = useUser();
 
-  if (!mounted) {
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/dev/feature-flags/FeatureFlagsTable.tsx:18]: `mounted` is no longer a thing

frontend/src/constants/environments.ts
```diff
@@ -1,36 +1,43 @@
...
-  FEATURE_SEARCH_OFF = "false",
-  FEATURE_OPPORTUNITY_OFF = "false",
-  NEXT_BUILD = "false",
-  SESSION_SECRET = "",
+  FEATURE_SEARCH_OFF,
+  FEATURE_OPPORTUNITY_OFF,
+  FEATURE_AUTH_OFF,
 } = process.env;
 
+export const featureFlags = {
```
doug-s-nava, line 20: this could be done elsewhere if we'd rather keep this file clean

frontend/src/hooks/useFeatureFlags.ts
```diff
@@ -1,48 +1,70 @@
...
+export function useFeatureFlags(): {
+  setFeatureFlag: (flagName: string, value: boolean) => void;
+  checkFeatureFlag: (flagName: string) => boolean;
+  featureFlags: FeatureFlags;
+} {
+  const [featureFlags, setFeatureFlags] =
+    useState<FeatureFlags>(defaultFeatureFlags);
 
+  // a workaround, as setting this in default state value results in hydration error
+  // on feature flag admin page. Does it cause a blip on other pages?
```
doug-s-nava, line 31: this seems fine but it's a workaround just for the admin page so it may not be necessary

frontend/tests/services/featureFlags/featureFlagHelpers.test.ts
```diff
@@ -0,0 +1,202 @@
...
+}
+
+describe("getFeatureFlagsFromCookie", () => {
+  test('getter loads feature flags with server-side NextRequest["cookies"]', () => {
+    expect(getFeatureFlagsFromCookie(MockRequestCookiesModule())).toEqual(
+      COOKIE_VALUE,
+    );
+  });
+
+  // // do we still need to support this?
```
doug-s-nava, line 60: may need some discussion here

frontend/src/services/featureFlags/FeatureFlagManager.ts
```diff
@@ -0,0 +1,148 @@
...
+   * @param searchParams - search params to check for enabled flags
+   * @example isFeatureEnabled("featureFlagName")
+   */
+  isFeatureEnabled(
+    name: string,
+    cookies: NextRequest["cookies"] | ReadonlyRequestCookies,
+    searchParams?: ServerSideSearchParams,
+  ): boolean {
+    if (!isValidFeatureFlag(name)) {
+      throw new Error(`\`${name}\` is not a valid feature flag`);
```
mdragon, line 90: Just wondering about this down the road when we want to retire a feature flag like we did with search. I think this is throwing an error early while we're in that process and maybe missed a check because it's this throws when consuming the flag, not when trying to parse what is stuck in the user's cookies? Just trying to make sure we don't make it where we can't ever kill an old flag because some users have it set true and we will break them if it's removed from the defaults.

doug-s-nava: that's an interesting point. The way things are structured though, I'm not too worried about this causing a problem. The calls to this function should always be coming with feature flag names set internal to the app, not with names read dynamically from cookies or query params. The only issue I can see is if a feature flag was not fully removed - if we remove a flag from the list of defaults but are still attempting to use it to gate behavior, for instance.

In the case where we remove a flag but a user's cookies are still set, I think the concern is that their cookie value wouldn't be honored, but that's probably not a big issue.

In any case it's existing behavior, so we can make another ticket to look into it if you think that's warranted.

mdragon: Thanks, I agree structurally we're ok, and I don't think we need to worry about not honoring them anymore once we've "dropped them" on the configurations side.

frontend/tests/hooks/useFeatureFlags.test.ts
```diff
@@ -1,30 +1,79 @@
...
+      expect(value).toEqual(true);
+    });
+
+    test("returns false if specified flag is not present", () => {
+      mockGetCookie.mockReturnValue(mockDefaultFeatureFlagsString);
+      const { result } = renderHook(() => useFeatureFlags());
+
+      const { checkFeatureFlag } = result.current;
+
+      const value = checkFeatureFlag("someFakeFeature4");
```
mdragon, line 75: Nit, but maybe making this "nonexistantFakeFeature" or "fakeFeatureDoesNotExist" as I could see 4 getting picked up by someone wanting to extend these tests further than then that would break this test. They'd realize the problem eventually, but might be easier to try to avoid the clash proactively.

mdragon:
One thought, maybe for a follow up ticket. I've had issues in the past with end users getting odd stuff in their FF or even just App cookies and it can be helpful to have a URL that will just wipe out whatever they've got out there, basically just wipe clean the cookie. Often this happens on a Logout URL to other cookies, but FF ones in particular are usually not easily cleared by a non-technical user and potentially will break stuff all over the site if they're not working right. Might be worth throwing another URL arg on the flag manager or adding a _ff_clear URL argument that the middleware? would respond to by just setting an empty cookie over whatever the user had stuck in there. That way the help desk could send them a link with that special URL argument rather than as set of instructions for how to clear cookies on this site.

frontend/src/components/Layout.tsx
```diff
@@ -1,15 +1,9 @@
...
-} from "next-intl";
+import { useTranslations } from "next-intl";
 import { setRequestLocale } from "next-intl/server";
 
-import Footer from "./Footer";
-import GrantsIdentifier from "./GrantsIdentifier";
-import Header from "./Header";
+import Footer from "src/components/Footer";
+import GrantsIdentifier from "src/components/GrantsIdentifier";
+import Header from "src/components/Header";
```
acouch, line 6: We had discussed when creating https://github.com/HHS/simpler-grants-gov/pull/2192 that we'd allow imports that are relative if they are in the same directory. We could revisit that, but would be good to update the es lint https://github.com/HHS/simpler-grants-gov/blob/main/frontend/.eslintrc.js#L22 so we are consistent.

frontend/src/components/Layout.tsx
acouch: Yeah, good catch, this is already accounted for in the [parent component](https://github.com/HHS/simpler-grants-gov/blob/main/frontend/src/app/%5Blocale%5D/layout.tsx#L94)

frontend/src/constants/environments.ts
```diff
@@ -1,38 +1,45 @@
...
+  NEXT_BUILD,
+  SESSION_SECRET,
   NEXT_PUBLIC_BASE_URL,
-  FEATURE_SEARCH_OFF = "false",
-  FEATURE_OPPORTUNITY_OFF = "false",
-  NEXT_BUILD = "false",
-  SESSION_SECRET = "",
+  FEATURE_SEARCH_OFF,
+  FEATURE_OPPORTUNITY_OFF,
+  FEATURE_AUTH_OFF,
```
acouch, line 17: This feels cleaner to grab them here and set defaults in the export. Would like to alphabetize them as well, might do that in a follow-up PR.

frontend/src/constants/environments.ts
acouch: They touch the envars so seems good to me.

frontend/src/components/Layout.tsx
doug-s-nava: got it, forgot that we were allowing that. Happy to revert this

frontend/src/services/featureFlags/FeatureFlagManager.ts
```diff
@@ -0,0 +1,148 @@
...
+   *
+   * @param name - Feature flag name
+   * @param cookies - server side cookies to check for enabled flags
+   * @param searchParams - search params to check for enabled flags
+   * @example isFeatureEnabled("featureFlagName")
+   */
+  isFeatureEnabled(
+    name: string,
+    cookies: NextRequest["cookies"] | ReadonlyRequestCookies,
+    searchParams?: ServerSideSearchParams,
```
acouch, line 87: These are query params, not just for the search page AFAICT, so `queryParams` as an arg would make more sense IMO.

mdragon:
> @mdragon if I'm understanding your idea correctly, I think we already have it implemented here:

Yup, exactly that, I didn't notice it was already in there.

frontend/src/constants/environments.ts
doug-s-nava: agree alphabetizing is a good idea, also agree it might be cleaner to do it later on

frontend/src/services/featureFlags/FeatureFlagManager.ts
doug-s-nava: not sure if you're saying that there is a better type to be using here or that we should rename the parameter. The nomenclature in the app is a bit confusing, and I confused myself enough while thinking about this that I had to look things up. https://stackoverflow.com/a/39294675 It sounds like "query params" or "search params" are not defined in any spec, but Next uses the term "search params" in their "useSearchParams" hook, and that's backed up by terminology in the URL API. "Query params" is definitely a very common phrase though. Not sure which way we'd want to go

acouch: I meant the param name. Not sure why next uses "search" but imagine it is b/c they want to make it easier for new devs to create search pages.

acouch (approved):
Looks great, though I would address the flicker for the login link, either in this PR or other ticket.

PR #3376, opened 2025-01-02
[Issue 3167] retries

babebe:
## Summary
Fixes #{ISSUE}


api/src/search/backend/load_search_data.py
```diff
@@ -20,6 +22,13 @@
 @flask_db.with_db_session()
 @flask_opensearch.with_search_client()
 @ecs_background_task(task_name="load-opportunity-data-opensearch")
+@retry(
+    stop=stop_after_attempt(3),  # Retry up to 3 times
+    wait=wait_fixed(2),  # Wait 2 seconds between retries
+    retry=retry_if_exception_type(
+        (TransportError, ConnectionTimeout)
+    ),  # Retry on TransportError (including timeouts)
+)
```
chouinar, line 31: While I'm alright with this kind of retry, I don't think putting it on the entire task itself would make sense. We'd want this specifically on the functions that are calling OpenSearch.

I had hoped we would use the retries already built into the OpenSearch client, we can configure it to retry for connection timeouts

babebe: Since the timeout settings will be global opted for this. Moved to `load_records`

chouinar (approved):

Issue #3270, opened 2024-12-17
Setup the ingest attachment plugin locally for OpenSearch


chouinar:
### Summary

https://opensearch.org/docs/latest/install-and-configure/additional-plugins/ingest-attachment-plugin/

To install the plugin we'll need to make a dockerfile for OpenSearch derived from the existing one that we use instead: https://opensearch.org/docs/latest/install-and-configure/install-opensearch/docker/#working-with-plugins

### Acceptance criteria

- [ ] Plugin working locally
- [ ] Some quick manual testing shows that the plugin is working and attachments can be added (actual configuration will come in a follow-up)

PR #3378, opened 2025-01-02
[Issue 3270] opensearch plugin

babebe:
## Summary
Fixes #{[3270](https://github.com/HHS/simpler-grants-gov/issues/3270)}

## Changes proposed
Screenshot 2024-12-31 at 10 29 00 AM

Dockerfile For Opensearch with attachment plugin
Updated docker compose to use Dockerfile to build image


api/docker-compose.yml
```diff
@@ -12,7 +12,9 @@ services:
       - default
 
   opensearch-node:
-    image: opensearchproject/opensearch:latest
+    build:
+      context: .
+      dockerfile: Dockerfile_opensearch
```
chouinar, line 17: Minor nit:

Could we put that dockerfile in a folder (let's say `opensearch`) and just name it `Dockerfile`? That seems to be the common pattern I've seen for Dockerfiles / sorta follows what we did for mock-oauth configuration as well.

api/Dockerfile_opensearch
```diff
@@ -0,0 +1,8 @@
+# Dockerfile for building opensearch image with Ingest-attachment plugin
+
```
chouinar, line 2:
```suggestion
# Dockerfile for building opensearch image with Ingest-attachment plugin
# This image is only used for local development

```

chouinar (commented):
Just two minor nitpicks, otherwise verified it works locally

PR #3379, opened 2025-01-02
[Issue #2841] Make the robots.txt dynamic so that it can mutate per environment

### LINKED ISSUE #2841: Add a robots.txt file
### Summary

Robots.txt is a best practice, helps link to the sitemap. Simple to do: https://nextjs.org/docs/app/api-reference/file-conventions/metadata/robots

It should discourage traffic for non-prod builds.

### Acceptance criteria

- [ ] basic robots.txt added
- [ ] disallow traffic for non prod envs

mdragon:
## Summary
Fixes #2841

## Changes proposed
We need to make the robots.txt dynamic per environment so that we can ban crawling of the lower environments but still support nuanced rules for upper environments.

## Context for reviewers
As a follow up change we will also make the Sitemap dynamic as well since we don't want to point to Sitemaps outside of Production. Our Sitemap isn't currently ready for being utilized so not including it for now.

## Additional information
Tested this locally and confirmed that we do have a dynamic robots.txt that reacts to changes to the environment variable at runtime, not build time. Not sure if this will all hook together as expected when it tries to run in AWS, so having this test for "dev" temporarily until we prove it is all likely to work in Prod.


acouch (approved):
Confirmed this builds dynamic locally and has the right cache control headers so that it should be stored in the CDN (isn't dynamic for each bot).

frontend/src/app/robots.ts
```diff
@@ -0,0 +1,40 @@
+// initial rules were absorbed from static robots.txt file
+
+import type { MetadataRoute } from "next";
+
+export const dynamic = "force-dynamic";
+
+export default function robots(): MetadataRoute.Robots {
+  return {
+    rules: [
+      process.env.AWS_ENVIRONMENT === "dev" // switching this temporarily to ensure that the variable is being set at AWS runtime as expected, will make it "prod" after confirming in Dev
```
doug-s-nava, line 10: not a big deal, but all the other env vars being used by the Next app are sourced out of https://github.com/HHS/simpler-grants-gov/blob/main/frontend/src/constants/environments.ts. In some way this is to avoid the potentially expensive access to process.env any more than necessary, but Next seems to not have much of a problem with that, so this is more of an organization thing than anything else

mdragon: Done

acouch: > process.env is defined on the C side an exposed as a getter to the runtime. Every access of process.env needs to drop down to C to get the value out - that is the performance penalty.

Yeah, not sure what the penalty is here in real life but should use the same pattern of the cached versions in `[environments.ts](https://github.com/HHS/simpler-grants-gov/blob/main/frontend/src/constants/environments.ts).`

frontend/src/constants/environments.ts
```diff
@@ -31,4 +32,5 @@ export const environment: { [key: string]: string } = {
   FEATURE_OPPORTUNITY_OFF,
   FEATURE_SEARCH_OFF,
   NEXT_BUILD,
+  AWS_ENVIRONMENT,
```
acouch, line 35: Above we are using `NODE_ENV` to determine the legacy host, could use the actual env instead of `NODE_ENV` for consistency since we are introducing that.

mdragon: My thought, though not strongly held, was that it really only matters in AWS, you don't have to specify it locally cause no robots will ever hit us there, as opposed to other Env Vars where stuff wouldn't work without specifying it even when running locally. I was also trying to sus-out a name that wouldn't get confused with other reasons we might specify Environment, like NODE_ENV, other concepts of Production like our "Production" build we run in every AWS environment but not laptop environments, etc. But I'm happy to drop the AWS_

infra/frontend/app-config/env-config/environment-variables.tf
```diff
@@ -7,6 +7,8 @@ locals {
     NEW_RELIC_ENABLED = "true"
     # see https://github.com/newrelic/node-newrelic?tab=readme-ov-file#setup
     NODE_OPTIONS = "-r newrelic"
+    # expose the current AWS Env to the FE Next Node Server at Runtime
+    AWS_ENVIRONMENT = var.environment
```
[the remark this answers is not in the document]
mdragon: My thought, though not strongly held, was that it really only matters in AWS, you don't have to specify it locally cause no robots will ever hit us there, as opposed to other Env Vars where stuff wouldn't work without specifying it even when running locally. I was also trying to sus-out a name that wouldn't get confused with other reasons we might specify Environment, like NODE_ENV, other concepts of Production like our "Production" build we run in every AWS environment but not laptop environments, etc. But I'm happy to drop the AWS_

frontend/src/constants/environments.ts
mdragon: Done

infra/frontend/app-config/env-config/environment-variables.tf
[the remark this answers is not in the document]
mdragon: Done

acouch (approved):
🎉

mdragon:
Testing in Stage: 
image

Testing in Dev (which is what Prod will work like):
image

PR #3383, opened 2025-01-02
set public env vars on cookie in middleware, read in client hook

doug-s-nava:
## Summary
Fixes #{ISSUE}


doug-s-nava:
this was an interesting idea for getting environment variables into client components, but we likely won't ever need to implement it, as we worked around it via feature flags implementation. see https://github.com/HHS/simpler-grants-gov/commit/b3fca0a655966fa59a00043104970b8d7669e4a6

Issue #3160, opened 2024-12-10
Setup configuration to run job that generates extracts of Opportunity DB tables for analytics daily


chouinar:
### Summary

Scheduling the job + configuring

### Acceptance criteria

- [ ] Job should run daily (early morning, let's say 5am)
- [ ] Job needs the `ANALYTICS_DB_CSV_FILE_PATH` env var set with a path to an s3 bucket that both the API and analytics can reach (API will write output here)

chouinar:
Can't configure this until we have a bucket that the API can write to and analytics code read from.

chouinar:
@coilysiren - After the PR, I'm still seeing `(AccessDenied) when calling the GetObject operation` when calling as the `analytics-dev-app` user.

coilysiren:
The deploy failed, so that's probably the cause

coilysiren:
https://github.com/HHS/simpler-grants-gov/actions/runs/12635878312/job/35207369432#step:5:162

coilysiren:
@chouinar try it again!

chouinar:
@coilysiren  - Copying from thread:

Need to make both the API and analytics terraform both be able to reference the API-analytics bucket for an env var. Easy to do for the API, put it in [s3_buckets.tf](https://github.com/HHS/simpler-grants-gov/blob/main/infra/api/app-config/env-config/s3_buckets.tf) - but not sure how the analytics code would reference it since it doesn't have that file.

PR #3384, opened 2025-01-02
[Issue #3160] Give analytic access to the transfer bucket

coilysiren:
## Summary

Relates to #3160

## Changes proposed

- Writes the iterative s3 bucket ids and arns to ssm params
- Reads the api's variants of those ssm params into the analytics application
- Gives the analytics application access to the api's s3 buckets

## ⚠️ Deploy Warning ⚠️ 

This creates a dependency between the API infra and Analytics infra. Because of this, there will be a race condition between the two deployments. If a deploy is broken, then the fix is just to deploy again.


mdragon (approved):
🚀

PR #3389, opened 2025-01-03
[Issue #3312] delete cd-*-infra workflows

### LINKED ISSUE #3312: Deploying both infra and application at the same time will reliably result in state locks due to race conditions
### Summary

TODO

### Acceptance criteria

TODO

coilysiren:
## Summary

Fixes #3312

## Context

I realized recently that race conditions from the regular deploys and the infra deploys would reliably cause problems. So I wondered, why would platform do this? How have they solve it? Well... platform doesn't have `infra` deploys that are separate from the application deploys. Which sounds like a reasonable solution this problem I guess. So that's the solution I'm going with.


mdragon (dismissed):
Sounds good. We had talked about whether there were any corner cases where we wouldn't want to deploy the application for an infra change, but those are probably not a big deal, compared to regular deploys getting mucked up more often.

mdragon (approved):
Approved.

PR #3393, opened 2025-01-05
[Issue #2590] Replace `gh` in analytics ETL

### LINKED ISSUE #2590: [Analytics] Call GitHub graphql API directly from python

### LINKED ISSUE #3203: Update analytics export to not use file system
### Summary

### Explanation

### Acceptance criteria

widal001:
## Summary

Replaces the sub-process call to the `gh` CLI by replacing it with a `GitHubGraphqlClient` class that can make calls to the GitHub GraphQL library directly from python.

Fixes #2590 
Fixes #3203 

- Adds a `GitHubGraphqlClient` class that can make paginated calls to the GitHub GraphQL API
- Replaces the calls to a separate sub-process in `src/analytics/etl/github/main.py` with the `GitHubGraphqlClient`
- Removes the `make-graphql-call.sh` script that previously invoked the `gh` CLI
- Removes the `

### Instructions to test

1. Checkout the PR locally
2. Re-build the docker image: `make build`
3. Run the GitHub data export and sprint reports end-to-end: `make sprint-reports-with-latest-data`

### Notes

We'll want to refactor the `src/analytics/integrations/github/` sub-package a little bit further pulling most of the code in the `main.py` file in that sub-package into `src/analytics/etl/github.py` instead.

I didn't include that in this PR to try to minimize the amount of code I was changing, but we can/should tackle that refactor in #3203 because some of the functions in `main.py` still write to the local file system, but can easily be updated to pass the exported data as a python dictionary.

The local run of sprint reports with the new code matches the output of the last run triggered by AWS step functions (using code in `main`) posted to slack:

### Sprint report for HHS/13

In Slack (based on `main`)

Screenshot 2025-01-04 at 7 46 28 PM

Locally, based on this feature branch:

Screenshot 2025-01-04 at 7 47 22 PM

### Sprint burndown for HHS/17

In Slack (based on `main`)

Screenshot 2025-01-04 at 7 48 06 PM

Locally, based on this feature branch:

Screenshot 2025-01-04 at 7 47 50 PM

### Deliverable percent complete

In Slack (based on `main`)

Screenshot 2025-01-04 at 7 49 02 PM

Locally, based on this feature branch:

Screenshot 2025-01-04 at 7 48 45 PM


analytics/Dockerfile
```diff
@@ -18,7 +18,6 @@ RUN apt-get update \
     libpq-dev \
     postgresql \
     wget \
-    jq \
```
widal001, line 21: Removing `jq` because we no longer need it for transformations

analytics/Dockerfile
```diff
@@ -28,24 +27,11 @@ RUN apt-get update \
...
   # Reduce the image size by clear apt cached lists
   # Complies with https://github.com/codacy/codacy-hadolint/blob/master/codacy-hadolint/docs/description/DL3009.md
   && rm -fr /var/lib/apt/lists/* \
   && rm /etc/ssl/private/ssl-cert-snakeoil.key
 
-# Install gh CLI
-# docs: https://github.com/cli/cli/blob/trunk/docs/install_linux.md
```
widal001, line 38: Removing this script because we no longer need the `gh` CLI

analytics/config.py
```diff
@@ -19,6 +19,7 @@ class DBSettings(PydanticBaseEnvConfig):
      ssl_mode: str = Field("require", alias="DB_SSL_MODE")
      db_schema: str = Field ("app", alias="DB_SCHEMA")
      slack_bot_token: str = Field(alias="ANALYTICS_SLACK_BOT_TOKEN")
+     github_token: str = Field(alias="GH_TOKEN")
```
widal001, line 22: Added this because we now need to reference it directly within the codebase, instead of indirectly like we did previously with the `gh` CLI

analytics/local.env
```diff
@@ -31,14 +31,15 @@ MB_DB_PASS=secret123
...
-# Slack Configuration   #
+# Token Configuration   #
 ###########################
 # Do not add these values to this file
 # to avoid mistakenly committing them.
 # Set these in your shell
 # by doing `export ANALYTICS_REPORTING_CHANNEL_ID=whatever`
 ANALYTICS_REPORTING_CHANNEL_ID=DO_NOT_SET_HERE
 ANALYTICS_SLACK_BOT_TOKEN=DO_NOT_SET_HERE
+GH_TOKEN=DO_NOT_SET_HERE
```
widal001, line 42: Prevents tests from failing if someone hasn't set their GitHub token locally.

analytics/pyproject.toml
```diff
@@ -68,8 +68,6 @@ line-length = 100
 [tool.ruff.lint]
 select = ["ALL"]
 ignore = [
-  "ANN101",  # missing type annotation for self
```
widal001, line 71: Removed these because they're deprecated in ruff

analytics/pyproject.toml
```diff
@@ -68,8 +68,6 @@ line-length = 100
...
-  "ANN102",  # missing type annotation for cls
```
widal001, line 72: Removed these because they've been removed in the latest version of ruff

analytics/pyproject.toml
```diff
@@ -78,7 +76,6 @@ ignore = [
   "PTH123",  # `open()` should be replaced by `Path.open()`
   "RUF012",  # Mutable class attributes should be annotated with `typing.ClassVar`
   "TD003",   # missing an issue link on TODO
-  "PT004",   # pytest fixture leading underscore - is marked deprecated
```
widal001, line 81: Same with this one

analytics/src/analytics/integrations/github/main.py
widal001, line 1: This files is basically a complete refactor, but preserves the existing helper functions for the export to prevent this PR from getting bigger than it already is.

analytics/src/analytics/integrations/github/make-graphql-query.sh
widal001, line 1: Removes this because we no longer need it

analytics/tests/logs/test_logging.py
```diff
@@ -40,7 +40,7 @@ def test_init(
         records = caplog.records
         assert len(records) == 2
         assert re.match(
-            r"^start test_logging: \w+ [0-9.]+ \w+, hostname \S+, pid \d+, user \d+\(\w+\)$",
+            r"^start test_logging: \w+ [0-9.]+ \w+, hostname \S+, pid \d+, user \d+\([\w\.]+\)",
```
widal001, line 43: Changed this because the tests were failing locally if there was a period in the username, e.g. `billy.daly`

analytics/config.py
coilysiren: Since we are in this file, can we rename `DBSettings` to something more accurate

analytics/src/analytics/integrations/github/main.py
```diff
@@ -1,164 +1,158 @@
...
+    owner: str,
+    project: int,
+    excluded_types: tuple = (),  # By default include everything
+) -> list[dict]:
+    """Pluck and reformat relevant fields for each item in the raw data."""
+    return [
+        {
+            "project_owner": owner,
+            "project_number": project,
+            "issue_title": safe_pluck(item, "content.title"),
```
coilysiren, line 47: I don't understand why we need `safe_pluck`. If there's a bunch of fields missing, I would rather the code raise a keyerror, instead of getting us bad (eg. mostly null) data.

coilysiren (commented):
I have (1) significant question about the data formatting, everything else looks fine

analytics/src/analytics/integrations/github/main.py
widal001: I'd definitely welcome your thoughts, though, on other potential strategies here as an intermediate step to implementing more robust data quality checks!

I think the basic things we're trying to achieve in the transform step are:

- Prevent "bad" data from being inserted into the database (i.e. data that is missing _required_ columns, or data that is missing optional columns because of a bug in the ETL -- the latter one is harder to check for)
- Support inserts of data that are missing optional columns, when they are valid
- Prevent failures of a subset of data from blocking loads of the remaining valid data

Typically I've achieved these goals by using tools like Great Expectations or Anomalo which run on the entire data set to check for quality issues or anomalies, but there might be immediate steps we can take right now to block bad data or catch more programming errors upfront.

coilysiren: Since we are already using Pydantic, we can use it's nested models feature:

https://stackoverflow.com/questions/70302056/define-a-pydantic-nested-model

Then drop this translation and safe_pluck layer entirely, and rely entirely on Pydantic to do the validation and null data transformation

DavidDudas-Intuitial (approved):
LGTM. Nice work, especially on the tests

analytics/src/analytics/integrations/github/main.py
coilysiren: cc @chouinar I would be interested in your thoughts on my _"rely entirely on Pydantic"_ assertion here. Going to request changes since I feel like this is in fact blocking, would rather do it right than do it fast.

coilysiren (changes requested):

analytics/src/analytics/integrations/github/main.py
widal001: @coilysiren Sorry I haven't gotten a chance to circle back on this in a minute -- been juggling some other items this sprint!

I think we could switch to using pydantic for the initial validation of the output from GitHub's GraphQL API, but that I don't think that will remove the need for something equivalent to a `safe_pluck()` function, which is what we use to _transform_ the output from the API into the flattened and simplified dataset that serves as an input to the `transform_and_load()` step.

### Next steps

Here's my suggestions for next steps:

- I'll spend some time today incorporating your suggestion in a stacked branch
- We can chat in tomorrow's infra office hours so I can show you what I mean about still needing something that works similar to `safe_pluck()` even if it's operating on a pydantic schema instead of a dictionary
- We can decide whether to merge that stacked branch into this one, or proceed with merging this branch as is, and working on an alternative for validating with pydantic as a separate effort associated with [this ticket](https://github.com/HHS/simpler-grants-gov/issues/3543) that I just created.

### Other thoughts

Taking a step back, I do appreciate the emphasis on trying to validate the API response the _correct_ way, but want to acknowledge this is outside the scope of the original ticket which was just about replacing the GitHub CLI.

The main reasons I personally think continued work would be better tackled as a separate PR are:

- This PR already introduces a pretty significant change in how we're interacting with the GitHub API, and I'd like to make that change relatively atomic, so we don't accidentally introduce more bugs with a broader refactor.
- The approach to parsing the GitHub API output in this PR is about as close as possible to the current approach (in `main`) which uses `jq` to do the transformations. That was by design, to reduce the surface area of what was being changed in this PR.
- We'll be doing more refactoring to the GitHub ETL in [this ticket](https://github.com/HHS/simpler-grants-gov/issues/3203), which would also provide us an opportunity to think more holistically about how we're validating and transforming the GitHub output.

chouinar: For safe_pluck/pydantic, seems like that would just be making the Pydantic models have default values of the nested objects

For example, let's assume we had a JSON object like:
```json
{
   "content": {
       "inner": {
           "a": 5
      }
   }
}
```

Let's assume each object is nullable, but we want the value of `a`. We could set up the pydantic models roughly like (quickly typed up, not validated the code is structured perfectly):

```
from pydantic import Field, BaseModel)

class Inner(BaseModel):
       a: int | None = None

class Content(BaseModel):
      inner: Inner = Field(default=Inner())

class TopLevel(BaseModel)
     content = Field(default=Content())
```

There might be some distinction in how this works between a field not being present/being null (probably configurable in the behavior of Pydantic, but requires some use of validators to handle the null case).

Might also need lambdas/factories instead of class calls like that

widal001: I think the core request here was filled by #3568 which is now merged into this branch in commit: [refactor: Update GitHub API validation (](https://github.com/HHS/simpler-grants-gov/pull/3393/commits/7adb1e5dec790e861adee9955a61ed633ffd1a76)

coilysiren (approved):
💹

PR #3406, opened 2025-01-06
[Issue #3133] Add auth env vars for login.gov OAuth

### LINKED ISSUE #3133: Setup environment variables for auth work non-locally
### Summary

There are a lot of env vars we need to set, some secret, some not, for the auth code to work non-locally.

For interacting with login.gov
* `LOGIN_GOV_CLIENT_ID` - will depend on what we configure for each env TODO
* `LOGIN_GOV_ENDPOINT` - will be `https://idp.int.identitysandbox.gov` in lower envs, and `https://secure.login.gov` in prod
* `LOGIN_GOV_JWK_ENDPOINT` - will be `https://idp.int.identitysandbox.gov/api/openid_connect/certs` in lower envs, and `https://secure.login.gov/api/openid_connect/certs` in prod
* `LOGIN_GOV_AUTH_ENDPOINT` - will be `https://idp.int.identitysandbox.gov/openid_connect/authorize` in lower envs, and `https://secure.login.gov/openid_connect/authorize` in prod (unlike the other paths, `/api` is not in this one, not sure why).
* `LOGIN_GOV_TOKEN_ENDPOINT` - will be `https://idp.int.identitysandbox.gov/api/openid_connect/token` in lower envs, and `https://secure.login.gov/api/openid_connect/token` in prod
* `LOGIN_FINAL_DESTINATION` - TODO - will depend on frontend work, but differ by env

For the key/tokens that the API manages:
* `API_JWT_PRIVATE_KEY` - an RSA256 private key (secret)
* `API_JWT_PUBLIC_KEY` - an RSA256 public key (secret)
* `API_JWT_ISSUER` - should be something like `simpler-grants-api-{env}`
* `API_JWT_AUDIENCE` - should be something like `simpler-grants-api-{env}`

Other (TODO):
* `ENABLE_AUTH_ENDPOINT`
* Something for private/public key that we use with querying Login.gov, TBD after we set that up locally

### Acceptance criteria

- [ ] All env vars and secrets set up for all 3 environments

chouinar:
## Summary
Fixes #3133

## Changes proposed
Added all of the env vars to connect to login.gov

Added secrets to parameter store for all envs

## Additional information
I've previously tested the lower env values locally

### DIFF
--- a/infra/api/app-config/dev.tf
+++ b/infra/api/app-config/dev.tf
@@ -29,5 +29,10 @@ module "dev_config" {
   # https://docs.aws.amazon.com/opensearch-service/latest/developerguide/what-is.html#choosing-version
   search_engine_version = "OpenSearch_2.15"
 
-  service_override_extra_environment_variables = {}
+  service_override_extra_environment_variables = {
+
+    # Login.gov OAuth
+    ENABLE_AUTH_ENDPOINT    = 0
+    LOGIN_FINAL_DESTINATION = "frontend-dev-1739892538.us-east-1.elb.amazonaws.com/api/auth/callback"
+  }
 }
--- a/infra/api/app-config/dev.tf
+++ b/infra/api/app-config/dev.tf
@@ -32,7 +32,6 @@ module "dev_config" {
   service_override_extra_environment_variables = {
 
     # Login.gov OAuth
-    ENABLE_AUTH_ENDPOINT    = 1
-    LOGIN_FINAL_DESTINATION = "frontend-dev-1739892538.us-east-1.elb.amazonaws.com/api/auth/callback"
+    ENABLE_AUTH_ENDPOINT = 1
   }
 }
--- a/infra/api/app-config/dev.tf
+++ b/infra/api/app-config/dev.tf
@@ -32,7 +32,7 @@ module "dev_config" {
   service_override_extra_environment_variables = {
 
     # Login.gov OAuth
-    ENABLE_AUTH_ENDPOINT    = 0
+    ENABLE_AUTH_ENDPOINT    = 1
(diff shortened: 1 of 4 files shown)


infra/api/app-config/dev.tf
```diff
@@ -29,5 +29,10 @@ module "dev_config" {
   # https://docs.aws.amazon.com/opensearch-service/latest/developerguide/what-is.html#choosing-version
   search_engine_version = "OpenSearch_2.15"
 
-  service_override_extra_environment_variables = {}
+  service_override_extra_environment_variables = {
+
+    # Login.gov OAuth
+    ENABLE_AUTH_ENDPOINT    = 1
+    LOGIN_FINAL_DESTINATION = "frontend-dev-1739892538.us-east-1.elb.amazonaws.com/api/auth/callback"
```
coilysiren, line 36: I'm torn on this. I don't want to over engineer, but I would also very much prefer not to be hard coding the ALB URL inside of our terraform configuration.

coilysiren: The frontend does this with a SSM parameter that's written in manually. For consistency that's what we should do here as well

coilysiren: https://github.com/HHS/simpler-grants-gov/blob/965b08704fa0ef8c9e65acaffe19aeb545fe1676/infra/frontend/app-config/env-config/environment-variables.tf#L38-L41

chouinar: I can make it a secret if desired, this approach likely gets replaced in the next few months as the redirect will instead be determined by who is calling us most likely

mdragon (approved):
👍🏻

coilysiren (approved):

Issue #3320, opened 2024-12-19
Setup the pipeline for ingesting attachments


chouinar:
### Summary

https://opensearch.org/docs/latest/ingest-pipelines/create-ingest/

At the start of our opportunity load to OpenSearch we want to create a pipeline like so:

```py
PUT _ingest/pipeline/multi-attachment
{
  "description" : "Extract attachment information",
  "processors" : [
  {
      "foreach": {
        "field": "attachments",
        "processor": {
          "attachment": {
            "target_field": "_ingest._value.attachment",
            "field": "_ingest._value.data"
          }
        }
      }
    }
  ]
}
```
(Note - we can use the client to do this, but this is the raw form of the request we make).

This should be safe to rerun as many times as we want as it is a PUT operation and will simply overwrite every time we run the job.

We won't use the pipeline yet, but you can test it works locally by going to http://localhost:5601/app/dev_tools#/console and doing:
```
PUT /opportunity-index-alias/_doc/1?pipeline=multi-attachment
{
    "opportunity_id": 1,
    "attachments": [
    {
      "filename" : "ipsum.txt",
      "data" : "dGhpcyBpcwpqdXN0IHNvbWUgdGV4dAo="
    },
    {
      "filename" : "test.txt",
      "data" : "VGhpcyBpcyBhIHRlc3QK"
    }
  ]
}
```

Which should create attachments when you fetch the opportunity.

### Acceptance criteria

- [ ] Processor added
- [ ] Tests added (or already exist) that run the process multiple times and shows it isn't an issue to rerun the job to set this up

PR #3432, opened 2025-01-07
3320/pipeline ingest attachment

babebe:
## Summary
Fixes #{[3320](https://github.com/HHS/simpler-grants-gov/issues/3320)}

## Changes proposed
Function to create opensearch multi-attachment pipeline. Called in `src/search/backend/load_opportunities_to_index.py::load_records`
## Context for reviewers
Tests already exists that runs the process multiple times. ex: `test_load_opportunities_to_index`


api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -66,13 +66,43 @@ def __init__(
...
+                }
+            ],
+        }
+        pipeline_name = "multi-attachment"
+        resp = self.search_client._client.ingest.put_pipeline(id=pipeline_name, body=pipeline)
+        # Check the response
+        if resp["acknowledged"]:
+            logger.info(f"Pipeline '{pipeline_name}' created successfully!")
+        else:
+            logger.error(f"Error creating pipeline: {resp}")
```
chouinar, line 104: I think making a method in our search client called `put_pipeline` would be useful for the future.

api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -66,13 +66,48 @@ def __init__(
...
+        if resp["acknowledged"]:
+            logger.info(f"Pipeline '{pipeline_name}' created successfully!")
+        else:
+            status_code = resp["status"] or 500
+            error_message = resp["error"]["reason"] or "Internal Server Error"
+
+            raise Exception(
+                error_message, {"pipeline_name": pipeline_name, "status_code": status_code}
+            )
+
```
chouinar, line 110: Put the error logic in the put_pipeline method that way any callers of the method don't need to parse the errors themselves.

For the exception, you should put everything into the message or make a custom exception class as passing in random variables they won't be stored anywhere.

babebe: Put in the extra args it in custom message and moved it into put_pipeline

chouinar (approved):

PR #3443, opened 2025-01-07
[Issue #3271] Setup structure of opportunity attachment transformation (minus file logic)

### LINKED ISSUE #3271: Setup the basic structure of the attachment transformation code
### Summary

As the attachment logic is going to be quite complex, we just want to focus on the simple parts of the transformations.

We should handle transforming everything BUT the attachment itself (for now make the code write a static file for the s3 file path).

This will help make the complex logic around the transformations of the file a bit more focused.

### Acceptance criteria

- [ ] Basic structure setup
- [ ] Job NOT configured to run yet - should be blocked by an env var like the other tasks

chouinar:
## Summary
Fixes #3271 

## Changes proposed
Add a subtask to transform the opportunity attachment table

DOES NOT handle the files themselves (next PR will handle that)

## Context for reviewers
The transformation code is bulky and has a lot of repeated patterns that needed to be setup. This PR gets all of that setup and working for the opportunity attachment table, except for the logic regarding moving the file from the staging table to our s3 bucket (which has a lot of extra fun complexities that will need to be handled special). Every other column in the table we want is copied over and functionally working with these changes.

The follow-up to this likely will refactor a bit to make the file logic reasonable, but figured it was best to start with a pattern that looks like the other transformations we've done.

## Additional information
From a clean database, you can run this locally by doing `make console` and then using our factories to generate data
```py
f.StagingTsynopsisAttachmentFactory.create_batch(size=50)
exit()
```

I then modified `TransformOracleDataTask` to only run the TransformOpportunity and new TransformOpportunityAttachment classes which successfully copied to my local opportunity_attachment table as expected:
Screenshot 2025-01-07 at 3 28 25 PM


api/tests/src/data_migration/transformation/subtask/test_transform_opportunity_attachment.py
```diff
@@ -1,41 +1,101 @@
...
+    TransformOpportunityAttachment,
+)
+from tests.src.data_migration.transformation.conftest import (
+    BaseTransformTestClass,
+    setup_opportunity_attachment,
+    validate_opportunity_attachment,
 )
 
 
-def test_uploading_attachment_staging(db_session, enable_factory_create, tmp_path):
```
chouinar, line 9: Because these tests use the staging table factories heavily, we don't need the specific tests anymore

mikehgrantsgov (approved):

babebe (commented):
LGTM

PR #3458, opened 2025-01-08
[Issue #3247] logged in header

### LINKED ISSUE #3247: Implement header logged in state including user email
### Summary

When a user is logged in, the header should contain an item displaying the user's email address which, when clicked, will display a user management drop down.

See figma here for details https://www.figma.com/design/FGxWtAgToKhehLJCiuy1zL/Simpler.Grants.gov-Project?node-id=963-1474&t=E8Yr2o687YW6GQq2-0

Note that this is dependent on https://github.com/HHS/simpler-grants-gov/issues/3246, as the frontend will need access to the users email. NextJs logic will need to be updated to extract this information.

### Acceptance criteria

- [ ] logged state of header matches figma designs
- [ ] user is able to sign out using new dropdown menu

doug-s-nava:
## Summary
Fixes #3247

## Changes proposed

- updates to session management to allow for decrypting the login.gov JWT in order to pull out user email
- creation of a logged in state for the header as per requirements in https://www.figma.com/design/FGxWtAgToKhehLJCiuy1zL/Simpler.Grants.gov-Project?node-id=795-16668&p=f&t=15lzPw6dJlQGDjUi-0

### Notes

- I made as much as possible of USWDS styling, but given that we are going against the grain in a number of ways with this nav implementation I ended up needing to create a few override styles. If anyone sees a way to work around these styles with build in USWDS stuff, that'd be great, but I reached the limit of my powers on a couple of these
- while doing this work I created a hook for JS side media queries that I ended up not needing. In case we need it in the future it's here https://github.com/HHS/simpler-grants-gov/compare/use-media-query?expand=1
- note that in building this functionality I had some difficulty working around the initialization of the encoded secrets, which relates to testing. If we are ok with leaving this functionality untestable, then this complexity I created here isn't necessary, but if we want to test that the env vars are being encoded properly, we need some way to encode them independent of the file initialization. We also potentially have the difficulty of handling cases where the keys are not available (such as when the session code runs during build). Here's what I tried to work around this that didn't work:

   - make a session manager class. This did some things for us, but in the end, we'd need to initialize the class outside of the file, which would entail a lot of useless class instances floating around rather than a single class instance which would be the goal
    - use next instrumentation with an initializer function. This also did some things, and effectively stored the encoded keys in memory on server startup. The problem is that the memory and keys got wiped during runtime when the module reloaded for some reason
- does not include new e2e tests, ticket for that work is here https://github.com/HHS/simpler-grants-gov/issues/3459

### Test steps

Refer to screenshots below throughout

1. if you haven't yet, run `make setup-env-override-file` in api
2. grab the `API_JWT_PUBLIC_KEY` value from your override.env file
3. start a local server on this branch with `npm run build && SESSION_SECRET='anything' AUTH_LOGIN_URL=http://localhost:8080/v1/users/login API_JWT_PUBLIC_KEY="<yr public key>" FEATURE_AUTH_ON=true npm run start`
4. visit localhost:3000
5. click `sign in`
6. enter any value and click the button to simulate login.gov sign in
7. _VERIFY_: you are returned to the User page with a success message
8. _VERIFY_: the header shows a logged in state including a fake email
9. click the email drop down
10. _VERIFY_: a sign out option appears
11. shrink viewport to mobile size
12. _VERIFY_: email disappears from header, but drop down still exists where it should with just the icon
13. click the icon
14. _VERIFY_: email and sign out link appear in drop down
15. click sign out
16. _VERIFY_: header returns to logged out state
17. _VERIFY_: all screens match Figma designs

### Logged out
Screenshot 2025-01-08 at 10 37 35 AM

### Logged in
Screenshot 2025-01-08 at 10 40 53 AM

### Logged in drop down
Screenshot 2025-01-08 at 10 41 00 AM

### Logged in mobile
Screenshot 2025-01-08 at 10 41 13 AM

### Logged in mobile drop down
Screenshot 2025-01-08 at 10 41 20 AM


frontend/src/components/USWDSIcon.tsx
```diff
@@ -12,7 +14,7 @@ const sprite_uri = SpriteSVG.src as string;
 export function USWDSIcon(props: IconProps) {
   return (
     <svg
-      className={props.className}
+      className={clsx("usa-icon", props.className)}
```
doug-s-nava, line 17: since all usage of this component had this class added manually, figured we should add it to the shared implementation and make this component a bit more useful

frontend/src/app/[locale]/dev/feature-flags/FeatureFlagsTable.tsx
```diff
@@ -1,39 +1,18 @@
...
 import { Button, Table } from "@trussworks/react-uswds";
 
-import Loading from "src/components/Loading";
-
 /**
  * View for managing feature flags
  */
 export default function FeatureFlagsTable() {
   const { setFeatureFlag, featureFlags } = useFeatureFlags();
-  const { user, isLoading, error } = useUser();
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/dev/feature-flags/FeatureFlagsTable.tsx:16]: with a working header we no longer need these changes on the feature flag page as a proof of concept

frontend/src/app/api/auth/session/route.ts
```diff
@@ -2,12 +2,12 @@ import { getSession } from "src/services/auth/session";
 
 import { NextResponse } from "next/server";
 
+export const revalidate = 0;
```
doug-s-nava, line 5: we don't want to cache anything here

frontend/src/components/Layout.tsx
```diff
@@ -17,16 +19,18 @@ export default function Layout({ children, locale }: Props) {
...
-        {t("Layout.skip_to_main")}
-      </a>
-      <Header locale={locale} />
-      <main id="main-content" className="border-top-0">
-        {children}
-      </main>
-      <Footer />
-      <GrantsIdentifier />
-    </div>
+    <UserProvider>
```
doug-s-nava, line 22: we could be more targeted here, but I don't see a need

frontend/src/services/auth/sessionUtils.ts
```diff
@@ -0,0 +1,46 @@
+import { KeyObject } from "crypto";
```
doug-s-nava, line 1: where I ended up is a little confusing, but the idea was to split out functionality that needs access to JWT keys from functionality that doesn't in order to help with testing and organization. I think we could put everything back in one file at this point, given what I learned in the process, but leaving as is for now

frontend/src/services/auth/session.ts
```diff
@@ -1,78 +1,94 @@
...
-  try {
-    const { payload } = await jwtVerify(sessionCookie, encodedKey, {
-      algorithms: ["HS256"],
-    });
-    return payload;
-  } catch (error) {
-    console.error("Failed to decrypt session cookie", error);
-    return null;
+// isolate encoding behavior from file execution
+const initializeSessionSecrets = () => {
```
doug-s-nava, line 22: this implementation probably isn't strictly necessary, but it's a little bit cleaner than instantiating on each module load, and helps deal with situations (like our build) where the keys won't be available

mdragon (approved):
Approved

frontend/.env.development
```diff
@@ -32,4 +32,4 @@ USE_SEARCH_MOCK_DATA=false
 NEW_RELIC_APP_NAME=
 NEW_RELIC_LICENSE_KEY=
 
-SESSION_SECRET=extraSecretSessionSecretValueSssh
+# SESSION_SECRET=extraSecretSessionSecretValueSssh
```
acouch: It occurs to me we should update the dev docs for the auth work, created a separate ticket https://github.com/HHS/simpler-grants-gov/issues/3474

frontend/src/services/auth/sessionUtils.ts
```diff
@@ -0,0 +1,46 @@
...
+import { JWTPayload, jwtVerify, SignJWT } from "jose";
+
+import { cookies } from "next/headers";
+
+export const CLIENT_JWT_ENCRYPTION_ALGORITHM = "HS256";
+export const API_JWT_ENCRYPTION_ALGORITHM = "RS256";
```
acouch, line 7: Should these live in `src/constants`? If not what is our criteria for what goes there vs not?

frontend/src/app/api/auth/session/route.ts
acouch: Not sure if this has an effect since a `fetch` isn't called in the session creation AFAICT. Next says it doesn't cache `GET` routes [if cookies](https://nextjs.org/docs/14/app/building-your-application/routing/route-handlers#opting-out-of-caching) are used. [force-dynamic](https://nextjs.org/docs/14/app/building-your-application/routing/route-handlers#convention) might be the way to make sure that the `GET` request is re-rendered every time, but can't hurt to tell it to revalidate as well.

frontend/src/constants/environments.ts
```diff
@@ -41,4 +42,5 @@ export const environment: { [key: string]: string } = {
   NEXT_BUILD: NEXT_BUILD || "false",
   SESSION_SECRET: SESSION_SECRET || "",
   NEXT_PUBLIC_BASE_URL: NEXT_PUBLIC_BASE_URL || "http://localhost:3000",
+  API_JWT_PUBLIC_KEY: API_JWT_PUBLIC_KEY || "",
```
acouch, line 45: Thinking out loud, might be useful to error or write debug statements here if certain envars are missing.

frontend/src/services/auth/sessionUtils.ts
acouch: `deleteSession` seems like it belongs with `createSession` but the others make sense as `utils`. Could do as a follow-up though.

acouch (approved):
Some comments but mostly just food for thought. Looks great, lets merge!

frontend/src/app/api/auth/session/route.ts
acouch: Also wondering if this is better as a `POST` method.

frontend/src/constants/environments.ts
doug-s-nava: one problem there is that these will always be undefined in the build process, so we'd be polluting the logs there unless we check the NEXT_BUILD env var to gate the logging. We're only setting that flag in the Dockerfile at the moment, though. Worth a further discussion

frontend/src/app/api/auth/session/route.ts
doug-s-nava: yeah, I had this in there just for safety, but I just verified and the 'no-store' on the fetch call is doing all the work. We can safely pull this out. As for POST vs GET, I'm not sure I see a reason to make a change, but we can discuss

Issue #3321, opened 2024-12-19
Load attachments into the search index


chouinar:
### Summary

**This work should be behind an environment variable feature flag which defaults to NOT doing anything (locally we can enable it)**

After we've setup the attachment pipeline in the prior ticket (#3320) we want to load attachments into the index.

We'll need to do the following:
* For every attachment load it from S3
* Base64 encode the attachment and set it as an attachments list on the opportunity JSON like so:

```json
{
    "opportunity_id": 1,
    "opportunity_title": "my title",
    "summary" : {...},
     .. a bunch of other fields not included here for brevity,
    "attachments": [
    {
      "filename" : "ipsum.txt",
      "data" : "dGhpcyBpcwpqdXN0IHNvbWUgdGV4dAo="
    },
    {
      "filename" : "test.txt",
      "data" : "VGhpcyBpcyBhIHRlc3QK"
    }
  ]
}
```

To make a pipeline get used when uploaded a record, you need to specify `pipeline="whatever_we_called_the_pipeline"` when calling the `self._client.bulk` method inside of our search client (have pipeline be an optional field passed into the bulk method).

NOTE: We likely need to have infra modify the search cluster to be larger for this. We will need a large search index (disk, not CPU, maybe memory) as the data size will grow from ~1gb to 55gb+ when we do this as the attachments are about 55gb.

### Acceptance criteria

- [ ] Attachments loaded into search for each opportunity
- [ ] Thorough testing

PR #3467, opened 2025-01-08
3321/load attachment

babebe:
## Summary
Fixes [#{3321}](https://github.com/HHS/simpler-grants-gov/issues/3321)

## Changes proposed
Add `ENABLE_OPPORTUNITY_ATTACHMENT_PIPELINE` feature flag
Use multi-attachment pipeline to bulk update opportunities into openSearch
Add/Update tests: Check attachment data was encoded 
Ensure attachments are indexed properly


api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -251,6 +253,23 @@ def fetch_existing_opportunity_ids_in_index(self) -> set[int]:
             (TransportError, ConnectionTimeout)
         ),  # Retry on TransportError (including timeouts)
     )
+
+    def fetch_attachments(self, opportunity_id) -> list:
+        return self.db_session.execute(
+            select(OpportunityAttachment.file_location, OpportunityAttachment.file_name).where(
+                OpportunityAttachment.opportunity_id == opportunity_id
+            )
+        ).all()
```
chouinar, line 262: You shouldn't need to do a query - the opportunity object already has a `opportunity_attachments` relationship that you can just reference.

api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -251,6 +253,23 @@ def fetch_existing_opportunity_ids_in_index(self) -> set[int]:
...
+
+
+    def get_attachment_data(self, file_path) -> str:
+        with ExitStack() as stack:
+            file = stack.enter_context(
+                smart_open.open(file_path, "rb"),
+            )
+            file_content = file.read()
+            return  base64.b64encode(file_content).decode("utf-8")
```
chouinar, line 271: Why not just use the context manager of files themselves? ExitStack is for really complex scenarios where you need to open a lot of context managers at once. Rare that we'll ever need that.


```suggestion
        with smart_open.open(file_path, "rb") as file:
            file_content = file.read()
            return  base64.b64encode(file_content).decode("utf-8")
```

api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -274,7 +293,20 @@ def load_records(self, records: Sequence[Opportunity]) -> set[int]:
...
+
+            for attachment in attachments:
+                file_loc, file_name = attachment
+                # data = self.get_attachment_data(file_loc)
+                json_record["attachments"].append({
+                    "filename": file_name ,
+                    "data": "data"
+                })
+
+            json_records.append(json_record)
```
chouinar, line 309: I'd suggest that the method we call just return the JSON objects we'll want. So this should look like:

```
json_record=schema.dump(record)

if <whatever check we add for an env var>:
     json_record["attachments"] = get_attachment_json_for_opportunity(...)

json_records.append(json_record)
```

api/src/search/backend/load_opportunities_to_index.py
chouinar: However, there is something we do want for the attachments - a filter.

We don't need to fully implement it here, can be a follow-up ticket, but we don't want EVERY attachment encoded and uploaded. Some files aren't "text" (some opportunities have mp4 files). We'll still want a function like this just for the purposes of putting that filter somewhere.

babebe: Okay i will add  the func.

chouinar (changes requested):
There's a lot going on here that refactored existing components that shouldn't have been touched

api/openapi.generated.yml
```diff
@@ -2092,7 +2092,7 @@ components:
           format: uuid
           description: The ID of the saved search
           example: !!python/object:uuid.UUID
-            int: 82637552140693101888240202082641616217
+            int: 223701124228644688372276000513924728179
```
chouinar, line 2095: I'll have a fix for this shortly so it stops popping up on PRs

api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -269,6 +275,31 @@ def fetch_existing_opportunity_ids_in_index(self) -> set[int]:
...
         return opportunity_ids
 
+    def filter_attachments(
+        self, attachments: List[OpportunityAttachment], filters: list
+    ) -> List[OpportunityAttachment]:
+        return [attachment for attachment in attachments if attachment.mime_type in filters]
+
+    def get_attachment_json_for_opportunity(
+        self, opp_attachments: List[OpportunityAttachment]
+    ) -> list[dict]:
```
chouinar, line 285:
```suggestion
    def filter_attachments(
        self, attachments: list[OpportunityAttachment], filters: list
    ) -> list[OpportunityAttachment]:
        return [attachment for attachment in attachments if attachment.mime_type in filters]

    def get_attachment_json_for_opportunity(
        self, opp_attachments: list[OpportunityAttachment]
    ) -> list[dict]:
```

api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -269,6 +275,31 @@ def fetch_existing_opportunity_ids_in_index(self) -> set[int]:
 
         return opportunity_ids
 
+    def filter_attachments(
+        self, attachments: List[OpportunityAttachment], filters: list
+    ) -> List[OpportunityAttachment]:
+        return [attachment for attachment in attachments if attachment.mime_type in filters]
```
chouinar, line 281: The filters will be a bit different than this, I'd just make it do `return [attachment for attachment in attachments]` for now

api/tests/src/search/backend/test_load_opportunities_to_index.py
```diff
@@ -119,6 +140,49 @@ def test_load_opportunities_to_index(
...
+        record = [d for d in resp.records if d.get("opportunity_id") == opportunity.opportunity_id]
+        attachment = record[0]["attachments"][0]
+
+        # assert correct attachment was uploaded
+        assert attachment["filename"] == filename
+
+        # assert content of file was b64encoded
+        decoded = base64.b64decode(attachment["data"])
+
+        assert decoded.decode("utf-8") == content
```
chouinar, line 184: Hmm, from the testing I did, the contents of the attachment should be decoded already in the response.

Can you print out what you see when you print out the "attachment" object here? The pipeline should have added half a dozen extra things (mime type, and a few other things). If not, that would mean the pipeline isn't being used.

babebe: This is what is returned: [{'filename': 'test_file_1.txt', 'data': 'SSBhbSBhIGZpbGU='}] which is what we are saving as attachments in the json object for each record

babebe: ah yes the pipeline argument was missing in the bulk request. Here's the response object:
`{'filename':` 'test_file_1.txt', 'data': 'SSBhbSBhIGZpbGU=', 'attachment': {'content_type': 'text/plain; charset=ISO-8859-1', 'language': 'ga', 'content': 'I am a file', 'content_length': 12}}`

chouinar (approved):

Issue #3341, opened 2024-12-20
Revisit error states after the MVP launch


mxk0:
### Summary

Considerations [here](https://github.com/HHS/simpler-grants-gov/issues/3173#issuecomment-2549456710)

### Acceptance criteria

- [ ]

PR #3470, opened 2025-01-09
[Issue #3244] auth login home and error redirects

### LINKED ISSUE #3244: Redirect logged in users to home page
### Summary

In the current temporary implementation, users are directed from the auth callback route at /api/auth/callback to a User page. As the auth designs and requirements specify that the user should be returned to the home page, we should redirect there instead once the login button in the header is implemented.

For requirements see https://www.figma.com/design/FGxWtAgToKhehLJCiuy1zL/Simpler.Grants.gov-Project?node-id=963-1474&t=E8Yr2o687YW6GQq2-0

Dependent on completion of https://github.com/HHS/simpler-grants-gov/issues/2653

### Acceptance criteria

- [ ] on login success, user is redirected to home page with logged in status reflected in state of the login button in header
- [ ] on login error redirect to a generic error page, which will need to be built for handling errors on some top level pages. We may need to think about error codes, as NextJs error handling implementation will send down a 500, and we may need to be more specific / less destructive around the initial error

doug-s-nava:
## Summary
Fixes #3244

## Changes proposed
- the callback route, which previously routed to the User page for all cases with different string messages, will now redirect to the home page on success, or an unauthorized page if no token is present or error page in error cases
- created error and unauthorized pages
- added middleware to implement the correct status codes on these redirects

Note that the error handling here is temporary for phase one, we will revisit as per this ticket for the next phase https://github.com/HHS/simpler-grants-gov/issues/3341

### Testing steps
1. start with `npm run dev` on this branch
2. visit http://localhost:3000?_ff=authOn:true and log in 
3. _VERIFY_: on successful login you're returned to the home page
4. visit htp://localhost:3000/api/auth/callback (note no token is present)
5. _VERIFY_: you're redirected to the unauthorized page
6. simulate an error by throwing from line 12 of the callback route
7. attempt login
8. _VERIFY_: you're redirected to the error page


acouch (approved):
👍🏿

Issue #3097, opened 2024-12-03
CDN for our S3 buckets


mxk0:
### Summary

Once we're done with the CDN work for the frontend, we'll apply the CDN to S3.

This ticket may get broken up into multiple tickets when we start the work.

### Acceptance criteria

- [ ] TBD, most likely CDN live in prod in front of prod S3

PR #3477, opened 2025-01-09
[Issue #3097] Adds configuration for S3 CDN

coilysiren:
## Summary

Probably finishes #3097

## Changes proposed

Adds configuration to the CDN that allows it to operate in 4 modes:

1. With no fronting domain name, and an ALB origin (eg. frontend dev)
2. With a fronting domain name, and an ALB origin (eg. frontend prod)
3. With no fronting domain name, and a S3 origin (eg. API dev)
4. With a fronting domain name, and a S3 origin (eg. API prod)

## Context for reviewers

It's complicated in here 😅 

## Testing

I tested by deploying to API dev and frontend dev

### DIFF
--- a/infra/modules/service/variables.tf
+++ b/infra/modules/service/variables.tf
@@ -204,10 +204,22 @@ variable "enable_load_balancer" {
   default     = true
 }
 
-variable "enable_cdn" {
-  description = "Whether to enable a CDN for the service"
+variable "enable_alb_cdn" {
+  description = "Whether to enable an ALB origin CDN for the service"
   type        = bool
   default     = false
+}
+
+variable "enable_s3_cdn" {
+  description = "Whether to enable a S3 origin CDN for the service"
+  type        = bool
+  default     = false
+}
+
+variable "s3_cdn_bucket_name" {
+  description = "The name of the S3 bucket to use for the S3 CDN"
+  type        = string
+  default     = null
 }
 
 variable "healthcheck_command" {
--- a/infra/modules/service/variables.tf
+++ b/infra/modules/service/variables.tf
@@ -205,19 +205,25 @@ variable "enable_load_balancer" {
 }
 
 variable "enable_alb_cdn" {
-  description = "Whether to enable an ALB origin CDN for the service"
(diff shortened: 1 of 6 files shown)


infra/modules/service/variables.tf
```diff
@@ -204,12 +204,30 @@ variable "enable_load_balancer" {
...
+  type        = string
+  default     = null
+}
+
+variable "s3_cdn_domain_name" {
+  description = "The domain name of the S3 bucket to use for the S3 CDN"
+  type        = string
+  default     = null
+}
+
```
chouinar, line 230: Doesn't need to be in this PR, but we'll want this value to be an env var the API gets.

The plan I had for this was just to swap out the s3 bucket for the CDN domain name on the fly in the API (eg. `s3://public-bucket-1234/path/to/file.txt` becomes `https://the-cdn-name/path/to/file.txt`)

mdragon:
Just to make sure I'm clear. 
1. The references to API in the PR itself (not the code) are about CDN->S3 because our "API" is the service that owns the S3 buckets for files that we're looking to front with Cloudfront? So we have CDN to ALB to CDN to S3 support following this change.
2. In the scenario where frontend turns on ALB CDN, and API turns on S3 CDN, do we have 2 CDNs? Or is this able to apply both sets of configuration into a single CDN instance?

coilysiren:
> In the scenario where frontend turns on ALB CDN, and API turns on S3 CDN, do we have 2 CDNs? Or is this able to apply both sets of configuration into a single CDN instance?

2 CDNs, which is my preference.

coilysiren:
> The references to API in the PR itself (not the code) are about CDN->S3 because our "API" is the service that owns the S3 buckets for files that we're looking to front with Cloudfront? So we have frontend CDN to ALB and API CDN to S3 support following this change.

Comment edited for clarity, and yes.

mdragon:
> 2 CDNs, which is my preference.

Ok, great, that's what I was hoping, but just couldn't put all the terraform together in my head to be sure. I think that's going to be very beneficial as we look to do mass-invalidation on deploy (which wouldn't need to touch the S3 CDN at all for example).

mdragon (approved):
Looks good.

PR #3483, opened 2025-01-10
[Unticketed] Remove import that is no longer necessary

chouinar:
## Changes proposed
A recent change removed the transfer topportunity table, but an import in a utility script wasn't cleaned up. Fixing that here


doug-s-nava (approved):
✅ ✅ ✅ ✅ ✅ 

my local now starts up with no issues

doug-s-nava (approved):

PR #3485, opened 2025-01-10
[Unticketed] Add cryptopgraphy dependency so RS256 is available

chouinar:
## Changes proposed
Add dependency on crypto library so other digital signature approaches are available

## Context for reviewers
We use RS256 encryption which requires crypto to be installed: https://pyjwt.readthedocs.io/en/latest/installation.html

Locally this wasn't an issue, but getting an error non-locally. We setup dependencies differently locally, so some test dependency probably brings the library in.


mdragon (approved):
🚀

PR #3486, opened 2025-01-10
[Issue #3322] Transform opportunity attachment files

### LINKED ISSUE #3322: Implement the transformation for attachment files
### Summary

In a prior ticket (#3271 ) we setup the structure for the transformation logic, but didn't yet copy the files, here is where we want to do that.

The logic for this will vary depending on update/insert/delete/whether the opportunity is a draft as follows:

## Insert

We need to stream the bytes from the staging table into the configured s3 location described above. The way we do that streaming should be able to re-use the same idea described in the testing setup above as well.

We need to have the opportunity record to determine whether or not the file should go to the private bucket where we store drafts.

If there are multiple files, they’ll all be uploaded to the same location: s3://<whichever_bucket>/opportunities/<oppId>/attachments/<filename>

## Update

Same as insert except if the s3 path would change. For example, if the file name changed as part of the update, we’d want to change it in s3 as well. For this, we should insert the new file and then delete the old one. No other changes are necessary.

## Delete

Delete requires deleting the files from s3, much like in the update case (just without a follow-up insert). We should make sure the delete approach is easily reused.

## Opportunity Update

If is_draft on an opportunity changes from True to False, we want to move all of the attachments to the public bucket. As opportunities get processed before attachments, we may move something and then later do another update, that is perfectly fine. 

## Opportunity delete

When we delete an ORM object, we automatically handle recursively deleting all the related models, but that wouldn’t cascade into s3.  If an opportunity needs to be deleted, we want to first delete all of its attachments from s3, and then delete the opportunity.

### Acceptance criteria

- [ ] Logic added
- [ ] Very thorough tests added, likely a few dozen

chouinar:
## Summary
Fixes #3322

## Changes proposed
A lot of file utilities (used in this PR) for handling reading/writing/naming files on s3

Utility for setting up s3 file paths for the attachments

Logic to handle inserts/updates/deletes of attachments and the files that need to move around on s3.

## Context for reviewers
There are some scenarios I haven't accounted for yet when the opportunity itself is modified (deleted / is no longer a draft), I originally wanted to handle this in a single PR, but I'll split that out as this one already was getting too big.

See the ticket for details on the scenarios we need to handle.

## Additional information
Testing this is a bit tedious - there is a lot that needs to be setup exactly to test it.

I'd recommend nuking anything you already have with `make volume-recreate`

Set the env var to enable the job to run (add `TRANSFORM_ORACLE_DATA_ENABLE_OPPORTUNITY_ATTACHMENT=1` to override.env)

Run `make console` and in that do `f.StagingTsynopsisAttachmentFactory.create_batch(size=50)` and then `exit()`

Finally you can run the job by doing `make cmd args="data-migration load-transform --no-load --transform --no-set-current"`


api/bin/setup_localstack.py
```diff
@@ -26,6 +27,14 @@ def does_s3_bucket_exist(s3_client: botocore.client.BaseClient, bucket_name: str
     return False
 
 
+def create_bucket_if_not_exists(s3_client: botocore.client.BaseClient, bucket_name: str) -> None:
+    if not does_s3_bucket_exist(s3_client, bucket_name):
+        logger.info("Creating S3 bucket %s", bucket_name)
+        s3_client.create_bucket(Bucket=bucket_name)
+    else:
+        logger.info("S3 bucket %s already exists - skipping", bucket_name)
```
mdragon, line 35: "... skip creating" or some similar more explicit statement of what we're skipping so it's clear we're not skipping the whole process or the whole opportunity?

api/tests/conftest.py
```diff
@@ -380,6 +380,18 @@ def mock_s3_bucket(mock_s3_bucket_resource):
     yield mock_s3_bucket_resource.name
 
 
+@pytest.fixture
+def other_mock_s3_bucket_resource(mock_s3):
```
babebe, line 384: Why is this needed?

api/tests/conftest.py
```diff
@@ -380,6 +380,18 @@ def mock_s3_bucket(mock_s3_bucket_resource):
...
+    bucket = mock_s3.Bucket("test_bucket")
```
babebe, line 385: should this not be a different bucket?

chouinar: You're right, that should be a different name

babebe:
Run both locally and tests!! LGTM!!! Great util functions!

babebe (approved):

babebe (approved):
LGTM

mdragon (approved):
🚀

PR #3501, opened 2025-01-13
[Issue #3184] Switch from RDFa to Microdata as meta property tags weren't making it to the client

### LINKED ISSUE #3184: Resolve Google crawl issue for breadcrumb navigation.
### Summary

Image
Docs for fixing: https://developers.google.com/search/docs/appearance/structured-data/breadcrumb

### Acceptance criteria

- [ ] Structured data is added such that we expect the Google revalidation process to succeed, post release
- [ ] After Release, we kick off the async revalidation process for these specific errors in: Google Search Console, Enhancements, Breadcrumbs
- [ ] After the async revalidation process, the pages affected by breadcrumb issues goes to zero

mdragon:
## Summary
Fixes #3184 

## Changes proposed
Use Microdata syntax within the Breadcrumb component

## Context for reviewers
RDFa style metadata was being flagged as missing by Google (position field, specifically). In testing, that seems to be because `<meta property` tags are stripped out by Next's page metadata handling to ensure that the more typical page level header meta tags work as expected via Next's layer for controlling them.

## Additional information
You could see in the existing code that we were setting `<meta property="position" content="{i+1}"/>` however but Google was flagging that property as missing on our Breadcrumbs. In testing, those tags were not showing in the Element Inspector nor when curling the pages directly. So something between our JSX code and the client was stripping the necessary tag. Microdata uses `itemprop` instead of `property` and that seems to evade the tag removal Next seems to be doing. We could have jumped all the way to JSOND, but this felt like a smaller, easier step since we already were doing in via markup, just in a way that wasn't fully translating to the browser.

### DIFF
--- a/frontend/src/components/Breadcrumbs.tsx
+++ b/frontend/src/components/Breadcrumbs.tsx
@@ -16,18 +16,20 @@ type Props = {
   breadcrumbList: BreadcrumbList;
 };
 
-const rdfaMetadata = {
+const microdata = {
   ol: {
-    vocab: "http://schema.org/",
-    typeof: "BreadcrumbList",
+    itemScope: true,
+    itemType: "https://schema.org/BreadcrumbList",
   },
   li: {
-    property: "itemListElement",
-    typeof: "ListItem",
+    itemScope: true,
+    itemProp: "itemListElement",
+    itemType: "https://schema.org/ListItem",
   },
   a: {
-    property: "item",
-    typeof: "WebPage",
+    itemScope: true,
+    itemProp: "item",
+    itemType: "https://schema.org/WebPage",
   },
 };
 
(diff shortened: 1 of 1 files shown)


mdragon:
This was a weird one, and so I'm open to other ideas, but I did confirm that our code had meta tags and they just weren't served to the client as long as they had a `property` attribute.

doug-s-nava (commented):
I think this is the right approach, and the main issue of meta tags moving around seems to be solved but I'm not sure how to test that the new meta tags are actually working. iS there a way to see what a Google crawl of the new page would look like, or do we need to deploy to find out? Holding off on approval just to make sure there's nothing else we can do to test prior to merging

Otherwise a couple observations:

* on main I am seeing the meta tags moved into the `<head>` rather than being stripped out entirely. It seems others have maybe seen similar issues, but unclear exactly what is going on. Could this be related? https://github.com/vercel/next.js/pull/66996
* on this branch I see `itemProp` in the source changed to `itemprop` in the rendered DOM. Assume this doesn't cause any issues with indexing but not sure how to test
* JSON LD does seem to be a much bigger change but given what I'm seeing in Next's issues it may be the way to go down the line if we want to really fix up our structured data approach. Something to discuss at eng sync?

mdragon:
> Is there a way to see what a Google crawl of the new page would look like

I was able to validate the snippet by copy/pasting it into the Google Tool
https://github.com/HHS/simpler-grants-gov/issues/3184#issuecomment-2587557509

mdragon:
> * on main I am seeing the meta tags moved into the `<head>` rather than being stripped out entirely. It seems others have maybe seen similar issues, but unclear exactly what is going on. Could this be related? https://github.com/vercel/next.js/pull/66996

Good catch I had been. Focused on the Breadcrumb section of the DOM so the fact they're getting redirected up there seems to further confirm that it's an intentional effort by Nest that has a negative side effect for our use case.

doug-s-nava (approved):
✅ ✅ ✅ ✅ ✅ 

Nice, tested at https://search.google.com/test/rich-results and all seems to be working now

PR #3503, opened 2025-01-13
[Issue #3500] Handle opportunity attachments when opp deleted or is published

### LINKED ISSUE #3500: Handle opportunity attachments when opportunity changes is_draft or is deleted
### Summary

# Summary

In #3322 we setup the opportunity attachment transformations when processing the attachments themselves, but need to also account for when the opportunity itself is updated (the below two scenarios).

## Opportunity Update

If is_draft on an opportunity changes from True to False, we want to move all of the attachments to the public bucket. As opportunities get processed before attachments, we may move something and then later do another update, that is perfectly fine. 

## Opportunity delete

When we delete an ORM object, we automatically handle recursively deleting all the related models, but that wouldn’t cascade into s3.  If an opportunity needs to be deleted, we want to first delete all of its attachments from s3, and then delete the opportunity.

### Acceptance criteria

- [ ] Logic updated
- [ ] Tests added

chouinar:
## Summary
Fixes #3500

## Changes proposed
Handle the following two cases during the transformation process
* Opportunity deleted - clean up attachments from s3
* Opportunity stops being a draft - move all the attachments to the other s3 bucket

## Context for reviewers
Mostly just some additional file utils for moving files to handle the above scenarios. Only noteworthy callout is that there isn't really a concept of "moving" a file on s3, it's just a copy+delete.

### DIFF
--- a/api/src/data_migration/transformation/subtask/transform_opportunity.py
+++ b/api/src/data_migration/transformation/subtask/transform_opportunity.py
@@ -3,16 +3,29 @@ from typing import Tuple
 
 import src.data_migration.transformation.transform_constants as transform_constants
 import src.data_migration.transformation.transform_util as transform_util
+from src.adapters.aws import S3Config
 from src.data_migration.transformation.subtask.abstract_transform_subtask import (
     AbstractTransformSubTask,
 )
 from src.db.models.opportunity_models import Opportunity
 from src.db.models.staging.opportunity import Topportunity
+from src.services.opportunity_attachments import attachment_util
+from src.task.task import Task
+from src.util import file_util
 
 logger = logging.getLogger(__name__)
 
 
 class TransformOpportunity(AbstractTransformSubTask):
+
+    def __init__(self, task: Task, s3_config: S3Config | None = None):
+        super().__init__(task)
+
+        if s3_config is None:
+            s3_config = S3Config()
+
+        self.s3_config = s3_config
+
     def transform_records(self) -> None:
         # Fetch all opportunities that were modified
         # Alongside that, grab the existing opportunity record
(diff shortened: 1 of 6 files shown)


api/src/data_migration/transformation/subtask/transform_opportunity.py
```diff
@@ -53,11 +66,18 @@ def process_opportunity(
                 extra,
             )
 
+            # Cleanup the attachments from s3
+            if target_opportunity is not None:
+                for attachment in target_opportunity.opportunity_attachments:
```
mdragon, line 71: Just making sure I understand the flow here. `target_opportunity` is an existing record on our side, that we're replacing due to an update. So this is effectively the "old" attachments and so we're cleaning those out and then we'll put "back" any attachments that are still referenced, and any net new attachments.

api/src/data_migration/transformation/subtask/transform_opportunity.py
```diff
@@ -76,5 +96,23 @@ def process_opportunity(
                 )
                 self.db_session.merge(transformed_opportunity)
 
+                # If an opportunity went from being a draft to not a draft (published)
+                # then we need to move all of its attachments to the public bucket
+                # from the draft s3 bucket.
+                if was_draft and transformed_opportunity.is_draft is False:
```
mdragon, line 102: Maybe I am mis-understanding in my comment above, because if I'm reading that right, would we not have deleted the draft entries above?

mdragon: Never mind, I saw in the other review a little more context around the above section and see now that it's only deleting them if the record is marked as deleted.

api/src/data_migration/transformation/subtask/transform_opportunity.py
mdragon: I see now these are deleted Opportunities only so there's no putting back, deleting is the final state.

api/src/data_migration/transformation/subtask/transform_opportunity.py
```diff
@@ -53,11 +66,18 @@ def process_opportunity(
                 extra,
             )
 
+            # Cleanup the attachments from s3
+            if target_opportunity is not None:
+                for attachment in target_opportunity.opportunity_attachments:
+                    file_util.delete_file(attachment.file_location)
```
mikehgrantsgov, line 72: Does this delete all revisions of a file (is the bucket using versioning?) If not, we may need to delete all past revisions as well.

chouinar: We don't have versioning on the bucket.

PR #3507, opened 2025-01-13
[Unticketed] fix auth bug - allow non ssl cookies in non-prod

doug-s-nava:
## Summary
Unticketed

## Changes proposed
* Only set the "secure" parameter on the session cookie to `true` in prod for now, as our lower environments are not set up with SSL yet
* Also adds some logging for future debugging

### Test steps
1. deploy to DEV if necessary 
2. log in
3. _VERIFY_: logged in header state is reflected after successful login

## Additional information
It's pretty clear we need better logging on the NextJS app, creating tickets to deal with that now.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -6,7 +6,7 @@
     "node": ">=22.13.0"
   },
   "scripts": {
-    "all-checks": "npm run lint && npm run ts:check && npm run test && npm run build",
+    "all-checks": "npm run lint && npm run ts:check && npm run test && npm run build -- --no-lint",
     "build": "next build",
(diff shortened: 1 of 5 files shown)


frontend/package.json
```diff
@@ -6,7 +6,7 @@
     "node": ">=22.13.0"
   },
   "scripts": {
-    "all-checks": "npm run lint && npm run ts:check && npm run test && npm run build",
+    "all-checks": "npm run lint && npm run ts:check && npm run test && npm run build -- --no-lint",
```
doug-s-nava, line 9: incidental, but speeds things up a bit

acouch (approved):
🚀

Issue #3498, opened 2025-01-13
Modify the search response to exclude the attachments from the response


chouinar:
### Summary

We don't need the attachments in the search results, and parsing the large text responses would be a lot of extra parsing, so we shouldn't have OpenSearch return them.

This can be done by adding `_source_excludes=field1,field2` to the query we send to OpenSearch (_search_excludes is also a param that the search client we use takes in).

### Acceptance criteria

1. Modified the search method of our opensearch client to pass through an optional excludes param
2. Modified our search API to set the `attachments` field to be excluded

PR #3531, opened 2025-01-15
[Issue 3498] exclude attachments from search response

babebe:
## Summary
Fixes #{[3498](https://github.com/HHS/simpler-grants-gov/issues/3498)}

## Changes proposed
Modified the search method of our opensearch client to pass through an optional excludes param
Modified our search API to set the attachments field to be excluded


chouinar (approved):
LGTM!

PR #3532, opened 2025-01-15
[Unticketed] Fix UUID default example to not randomize each time

chouinar:
## Changes proposed
Fix schema field logic for UUID to not default to a randomly generated UUID every time and either use a static value or whatever the user passed in

## Context for reviewers
Was seeing on recent PRs that after we added UUID params to the schemas, they were randomly generating each time.

## Additional information
Generated the file again myself a few times to verify it doesn't change each run now


mdragon (approved):
👍🏻

PR #3524, opened 2025-01-15
[Issue #3522] Do not show pagination on search when no search results returned.

yebra06:
## Summary
Fixes #3522 

### Time to review: 10m

Testing through local dev environment:
1. Run the application locally.
2. Navigate to the search for opportunities page. 
3. Navigate to page 2 of results.
4. Add Agency > Consumer Product Safety Commission filter to search results.
5. Ensure that there pagination is not displayed AND that there are no search results.

Frontend unit testing:
In the frontend directory run the following command and ensure all tests pass:
- `npm run test`
Screenshot 2025-01-14 at 8 35 17 PM

Screen Recording 2025-01-14 at 8 16 00 PM

PR #3533, opened 2025-01-15
[Issue #3522] refine fix for pagination showing with no results

### LINKED ISSUE #3522: Pagination can appear with no search results
### Summary

It is possible to get search into a state where the app shows no search results, but does show a pagination menu.

Image

It is expected that the pagination menu would not appear if there are no search results for a given search.

Note that this reproducible only if you filter down to zero results from a page other than the first page of search results.

### Reproducibility

### How to reproduce the bug

* visit https://simpler.grants.gov/search?refresh=true&page=2 (puts you on page two of the default search
* check the Agency filter for "Consumer Product Safety Commission"

### Browser

Not applicable

### Code of Conduct

doug-s-nava:
## Summary
Fixes #3522

## Changes proposed
While the client facing issue for the bug mentioned in the ticket was fixed in https://github.com/HHS/simpler-grants-gov/pull/3524 this provides a more comprehensive fix by correcting some slightly funky behavior in pagination related components

## Context for reviewers
1. start the app on this branch
2. run a search with all kinds of different pagination scenarios
3. _VERIFY_: all pagination related behavior works, pagination continues to not appear if there are no search results


frontend/src/components/search/SearchSortBy.tsx
```diff
@@ -52,11 +45,13 @@ export default function SearchSortBy({
     },
   ];
 
-  const handleChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
-    const newValue = event.target.value;
-    updateTotalResults(totalResults);
```
doug-s-nava, line 57: there's no need to set total results in multiple places in the app, since both of these components are responding to the same request

frontend/src/components/search/SearchPagination.tsx
```diff
@@ -14,40 +14,61 @@ export enum PaginationPosition {
...
+
+  useEffect(() => {
+    if (!loading) {
+      updateTotalResults(String(totalResults));
+    }
+  }, [updateTotalResults, totalResults, loading]);
 
   const updatePage = (page: number) => {
-    updateTotalPages(String(total));
-    updateTotalResults(totalResults);
```
doug-s-nava, line 41: with these functions only being called in `updatePage` the client side pagination state was only being updated on pagination clicks, not in response to the search results updating from the completed API request. With this new setup, the state will be set purely based on the results of the search request whenever a new request is made

frontend/src/app/[locale]/search/QueryProvider.tsx
```diff
@@ -3,7 +3,7 @@
 import { useSearchParams } from "next/navigation";
 import { createContext, useCallback, useMemo, useState } from "react";
 
-interface QueryContextParams {
+export interface QueryContextParams {
```
Comment (acouch) [inline frontend/src/app/[locale]/search/QueryProvider.tsx:6]: Is there a reason to export this? Didn't see it called elsewhere

acouch (dismissed):
👍🏿

frontend/src/app/[locale]/search/QueryProvider.tsx
Comment (doug-s-nava) [reply frontend/src/app/[locale]/search/QueryProvider.tsx:6]: good catch, this is left over from a previous idea. Removing.

acouch (approved):

Issue #3497, opened 2025-01-13
Add an allow list for specific file types to be uploaded into the search index


chouinar:
### Summary

Before we turn on the process to load opportunities into our search index, we want to create a filter so only certain file types get uploaded. For example, we don't want to upload an mp4 file as the file needs to be roughly a text file.

Filter to only files with the following suffixes (case-insensitive):
* .txt
* .pdf
* .docx
* .doc
* .xlsx
* .xlsm
* .html
* .htm
* .pptx
* .ppt
* .rtf

Note that it's more efficient to filter by doing something like than any sort of looping:

```
ALLOWED_ATTACHMENT_SUFFIXES = set(["pdf", "docx", ...])

file_suffix = my_attachment.file_name.lower().split(".")[-1]

if file_suffix in ALLOWED_ATTACHMENT_SUFFIXES:
    # do something with it
```

### Acceptance criteria

- [ ] Filter added

chouinar:
Some data for deciding the open questions.

Mime type counts by file suffix:
```
text/richtext 1
	.rtf: 1
/ 1
	.pdf: 1
application/vnd.fdf 1
	.fdf: 1
audio/mp3 1
	.mp3: 1
audio/x-m4a 1
	.m4a: 1
multipart/related 1
	.mht: 1
application/octet-stream; name="LM2023456789123 (1)" 1
	.LM: 1
application/octet-stream; name=IMLS_Library_4_0-V4.0_F831.xls 1
	.xls: 1
application/octet-stream; name=1234-rover_and_rocks_medium.mp4 1
	.mp4: 1
text/plain; charset=us-ascii; name=RDFile1.txt 1
	.txt: 1
application/octet-stream; name=1234-instructions5.docx 1
	.docx: 1
application/octect-stream 2
	.pdf: 2
application/vnd.ms-xpsdocument 2
	.xps: 2
application/vnd.oasis.opendocument.text 2
	.odt: 2
application/vnd.openxmlformats-officedocument.wordprocessingml.template 2
	.dotx: 2
video/mp4 2
	.mp4: 2
audio/wav 2
	.wav: 2
text/calendar 3
	.ics: 3
text/plain; charset=us-ascii 3
	.txt: 2
	.LM2023456789123456: 1
image/pjpeg 5
	.jpg: 5
application/vnd.ms-excel.sheet.binary.macroEnabled.12 5
	.xlsb: 5
application/vnd.ms-word.document.macroEnabled.12 6
[...]

PR #3544, opened 2025-01-16
[Issue 3497] add allowed list to filter through attachments

babebe:
## Summary
Fixes [#{3497}](https://github.com/HHS/simpler-grants-gov/issues/3497)

## Changes proposed
Added Filter to only files with the allowed suffixes (case-insensitive)
Update test

### DIFF
--- a/api/src/search/backend/load_opportunities_to_index.py
+++ b/api/src/search/backend/load_opportunities_to_index.py
@@ -27,6 +27,9 @@ from src.util.env_config import PydanticBaseEnvConfig
 
 logger = logging.getLogger(__name__)
 
+ALLOWED_ATTACHMENT_SUFFIXES = set(
+            ["txt", "pdf", "docx", "doc", "xlsx", "xlsm", "html", "htm", "pptx", "ppt", "rtf"]
+        )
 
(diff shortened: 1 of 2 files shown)


api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -275,28 +275,31 @@ def fetch_existing_opportunity_ids_in_index(self) -> set[int]:
...
         return opportunity_ids
 
-    def filter_attachments(
-        self, attachments: list[OpportunityAttachment]
-    ) -> list[OpportunityAttachment]:
-        return [attachment for attachment in attachments]
+    def filter_attachment(self, attachment: OpportunityAttachment) -> bool:
+        ALLOWED_ATTACHMENT_SUFFIXES = set(
+            ["txt", "pdf", "docx", "doc", "xlsx", "xlsm", "html", "htm", "pptx", "ppt", "rtf"]
+        )
```
chouinar, line 281: Define this once at the top of the file, otherwise it'll re-initialize for every single attachment.

babebe: Ah of course! Moved

chouinar (approved):

PR #3554, opened 2025-01-16
[Issue #3152] properly error on client for max search term length

### LINKED ISSUE #3152: Handle search max search term length
### Summary

There is a 100 character length maximum on the search term enforced on the API. Currently if a user enters a term longer than 100 characters the UI shows them a generic error. For better UX we should:

* respond to the API error with a more specific error based on the error returned from the API
* validate search term length with a max of 100 characters on the frontend and disallow search for terms that are too long

The error from the API response looks like this

Image

### Acceptance criteria

* respond to the API error with a more specific error based on the error returned from the API
* validate search term length with a max of 100 characters on the frontend and disallow search for terms that are too long

More specific requirements around verbiage TBD

doug-s-nava:
## Summary
Fixes #3152

## Changes proposed
Main goal here is to show a helpful error message with a relevant call to action to users when they enter a search term longer than the API allows (100 characters).

To do that this change:
* introduces client side validation of the query string size, and displays an error on submit click without submitting
* to handle a case where a request does go out from the client but returns with a validation error (most likely due to a change in API behavior that the client has not yet responded to), the error handling system has been updated to pass relevant information to the client so that it can be handled

This change also fixes a bug that was preventing the search form state to be maintained in the case of an error.

I also observed when doing this work that Next seems to treat the errors thrown by the search fetcher to be unhandled, even though they redirect to the search error page. This means the error toast thing shows up on dev, and the error is logged to the console as unhandled. Is this expected?

### Test steps for client side validation

3. enter a search term longer than 99 characters and click submit
4. _VERIFY_: search results do not update
5. _VERIFY_: an error message is shown (see client validation error screenshot below)
6. enter a search term less than 99 characters long and submit
7. _VERIFY_: search is performed successfully, and error message is removed

### Test steps for error response handling

Totally synthetic but the easiest way to do this is to disable the client side check.
1. comment out lines 23-29 in SearchBar.tsx
3. enter a search term longer than 99 characters and click submit
4. _VERIFY_: page loads with no results and an error message (see server side error screenshot below)
5. _VERIFY_: query string is maintained (not cleared) in the search bar

### Server side error
Screenshot 2025-01-17 at 11 28 07 AM

### Client side validation error

Screenshot 2025-01-17 at 12 22 58 PM

### DIFF
--- a/frontend/src/app/[locale]/search/error.tsx
+++ b/frontend/src/app/[locale]/search/error.tsx
@@ -76,10 +76,11 @@ export default function SearchError({ error }: ErrorProps) {
     console.error(error);
   }, [error]);
 
+  // note that the validation error will contain untranslated strings
   const ErrorAlert =
     parsedErrorData.details && parsedErrorData.type === "ValidationError" ? (
       <Alert type="error" heading={t("validationError")} headingLevel="h4">
-        {`Error in ${parsedErrorData.details.field}: ${parsedErrorData.details.message}`}
+        {`Error in ${parsedErrorData.details.field || "a search field"}: ${parsedErrorData.details.message || "adjust your search and try again"}`}
       </Alert>
     ) : (
       <ServerErrorAlert callToAction={t("generic_error_cta")} />
--- a/frontend/src/app/[locale]/search/error.tsx
+++ b/frontend/src/app/[locale]/search/error.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import QueryProvider from "src/app/[locale]/search/QueryProvider";
-import { FrontendErrorDetails } from "src/errors";
+import { FrontendErrorDetails } from "src/types/apiResponseTypes";
 import { ServerSideSearchParams } from "src/types/searchRequestURLTypes";
 import { Breakpoints, ErrorProps } from "src/types/uiTypes";
 import { convertSearchParamsToProperTypes } from "src/utils/search/convertSearchParamsToProperTypes";
--- a/frontend/src/app/[locale]/search/error.tsx
+++ b/frontend/src/app/[locale]/search/error.tsx
@@ -1,12 +1,14 @@
 "use client";
 
 import QueryProvider from "src/app/[locale]/search/QueryProvider";
+import { FrontendErrorDetails } from "src/errors";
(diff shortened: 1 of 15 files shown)


frontend/src/app/[locale]/search/error.tsx
```diff
@@ -73,6 +76,16 @@ export default function SearchError({ error }: ErrorProps) {
     console.error(error);
   }, [error]);
 
+  // note that the validation error will contain untranslated strings
+  const ErrorAlert =
+    parsedErrorData.details && parsedErrorData.type === "ValidationError" ? (
+      <Alert type="error" heading={t("validationError")} headingLevel="h4">
+        {`Error in ${parsedErrorData.details.field || "a search field"}: ${parsedErrorData.details.message || "adjust your search and try again"}`}
```
Comment (acouch) [inline frontend/src/app/[locale]/search/error.tsx:83]: Could we do a follow-up to translate this?

frontend/src/errors.ts
```diff
@@ -1,9 +1,19 @@
 /**
  * @file Custom Error classes. Useful as a way to see all potential errors that our system may throw/catch
+ * Note that the errors defined here rely on stringifying JSON data into the Error's message parameter
+ * That data will need to be parsed back out into JSON when reading the error
  */
 
 import { QueryParamData } from "src/types/search/searchRequestTypes";
 
+export interface FrontendErrorDetails {
```
acouch, line 9: Is there a reason not to store this in `src/types/...` ?

acouch (approved):
👍🏿 with a few nit Qs

frontend/src/app/[locale]/search/error.tsx
Comment (doug-s-nava) [reply frontend/src/app/[locale]/search/error.tsx:83]: I can make a ticket, but I'm not exactly sure what the scope would need to be. To get this to work as a fully translated UI we'd need to handle translating stuff that's coming from the API, including error messages and our field names. That seems to open up a lot of questions, and since it doesn't seem like we're prioritizing translation right now I don't know when the right time for that conversation is

Comment (acouch) [reply frontend/src/app/[locale]/search/error.tsx:83]: I meant the `a search field` part.

acouch (approved):

PR #3555, opened 2025-01-16
[no ticket] Fix ordering of CDN deploy operations

coilysiren:
## Summary

Fixes https://github.com/HHS/simpler-grants-gov/actions/runs/12815925972/job/35736051716

## Changes proposed

When deploying from scratch, as opposed to the iterative deploy you would do during testing, you often find small quirks where the order to operations for the deploy are slightly incorrect. This is one of those cases.


mdragon (approved):
👍🏻

PR #3564, opened 2025-01-17
[Issue #3520] Add the ability to only process a subset of opportunity attachments

### LINKED ISSUE #3520: Make it possible to run opportunity attachment logic for a subset of attachments
### Summary

Testing the opportunity attachment logic non-locally is going to be time consuming, we should verify things work in chunks and have an env var that can be passed in that dictates the total number of opportunity attachments to process in a "batch". This way we can run 10, 100, 1000... attachments at a time.

### Acceptance criteria

- [ ] Env var added

chouinar:
## Summary
Fixes #3520

## Changes proposed
Add an env var that lets us run the opportunity attachment transform job only process a smaller set of attachments

## Context for reviewers
Attachments are going to take a long time to run, and I wanted an easy way when testing to just tell the job to "stop after X attachments". This is that mechanism. It's pretty hacky/straightforward, just tell it to stop after processing a number of values, instead of finishing. It will still save these changes.

## Additional information
If you want to test it locally, you can do the following.

First set the following env vars in override.env:
```
TRANSFORM_ORACLE_DATA_ENABLE_OPPORTUNITY_ATTACHMENT=1
TOTAL_ATTACHMENTS_TO_PROCESS=5
```

`make volume-recreate` to drop your DB state
`make console` and then:
```py
f.StagingTsynopsisAttachmentFactory.create_batch(size=50)
exit()
```
Finally: `make cmd args="data-migration load-transform --no-load --transform --no-set-current"`


mikehgrantsgov (approved):
Setting the env var in terraform would be in another PR? Or does it need to be in TF? Either way, this one looks good.

chouinar:
> Setting the env var in terraform would be in another PR? Or does it need to be in TF? Either way, this one looks good.

Won't ever set in terraform, can set manually when I run the job during testing. Can run jobs with the `run-command.sh` script we have.

A different command, but I ran: `bin/run-command.sh --environment-variables "$(jq -c . ~/env_vars/load_job.json)" api dev '["poetry", "run", "flask", "task", "create-analytics-db-csvs"]'`

Set the overriden env vars in a file that I load with `jq` on the fly. Sorta like an improved version of the `run_ecs_task` script we had on Mass PFML if you ever used that.

PR #3565, opened 2025-01-17
[Issue #3527] Modify the logic around the opportunity change tracking table to never delete records

### LINKED ISSUE #3527: Modify the logic around the opportunity change tracking table to never delete records
### Summary

We setup a table for handling an incremental search load to populate with opportunity IDs whenever the opportunity (or related table) is modified.

However, we plan to delete records from this table after loading them to search. This would be fine, but we've found a few use cases where having that information for an opportunity would be useful.

We should instead modify the search logic to track which opportunities it needs to process in another way (a timestamp for when we loaded the opportunity last?).

Additionally, we should backfill every opportunity to be in this table. This is fine to do in a migration (grab the greatest `updated_at` timestamp from the same tables we configured to trigger an update to this table in the first place).

### Acceptance criteria

- [ ] Search logic modified to use table without deleting
- [ ] Table likely should be renamed to something more generic (fine if we lose data)
- [ ] Data backfilled

mikehgrantsgov:
## Summary
Fixes #3527

## Changes proposed
Add a new table to track task progress
Rename queue table and add all opportunities to it. Add any new opportunities that come in (should basically always be 1:1)
For index job, load data based on `last_loaded_at` date rather than using `has_update` field or deleting records as we go.

## Context for reviewers
Cascade deleting orphan relationship remains unchanged
JobTable is meant to be general purpose and can be used, possibly ended, by other tasks.

## Additional information
See unit tests


api/src/db/models/task_models.py
```diff
@@ -0,0 +1,22 @@
...
+    COMPLETED = "completed"
+    FAILED = "failed"
+
+
+class JobTable(ApiSchemaTable, TimestampMixin):
+    __tablename__ = "job"
+
+    job_id: Mapped[uuid.UUID] = mapped_column(UUID, primary_key=True, default=uuid.uuid4)
+    job_type: Mapped[str]
+    job_status: Mapped[str]
```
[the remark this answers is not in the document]
mikehgrantsgov: I was thinking about the metrics object and will add it for flexibility down the road (although I think NR is place to store and query these). When trying to make this an enum the migration would try creating these enum types that I wasn't sure was right. Will make this something like LkJobStatus

api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -121,25 +123,45 @@ def incremental_updates_and_deletes(self) -> None:
...
+        if last_successful_job:
+            query = query.where(
+                (OpportunityChangeAudit.last_loaded_at.is_(None))
+                | (OpportunityChangeAudit.last_loaded_at > last_successful_job.created_at)
+            )
+        else:
+            # If no previous successful job, get unprocessed records
+            query = query.where(OpportunityChangeAudit.last_loaded_at.is_(None))
+
+        queued_opportunities = self.db_session.execute(query).scalars().all()
```
chouinar, line 159: Do we need the `last_loaded_at` column? Could we use the `updated_at` column to know when something was last updated + if there was no job, just fetch everything (maybe limit it to last X days? Could always manually make a job to handle that case).

api/src/task/task.py
```diff
@@ -28,9 +30,18 @@ class Metrics(StrEnum):
...
         self.metrics: dict[str, Any] = {}
+        self.job: JobTable | None = None
 
     def run(self) -> None:
         try:
+            # Create initial job record
+            job_id = uuid.uuid4()
+            self.job = JobTable(
+                job_id=job_id, job_type=self.cls_name(), job_status=JobStatus.STARTED
+            )
```
chouinar, line 41: Should be able to just rely on the db to set it, right?
```suggestion
            self.job = JobTable(
                job_type=self.cls_name(), job_status=JobStatus.STARTED
            )
```

api/src/search/backend/load_opportunities_to_index.py
mikehgrantsgov: I don't think we need this anymore, and modified the logic to use `updated_at`

api/src/db/models/task_models.py
```diff
@@ -0,0 +1,22 @@
...
+from sqlalchemy import ForeignKey
+from sqlalchemy.dialects.postgresql import JSONB, UUID
+from sqlalchemy.orm import Mapped, mapped_column
+
+from src.adapters.db.type_decorators.postgres_type_decorators import LookupColumn
+from src.db.models.base import ApiSchemaTable, TimestampMixin
+from src.db.models.lookup_models import JobStatus, LkJobStatus
+
+
+class JobTable(ApiSchemaTable, TimestampMixin):
```
mdragon, line 12: This will be a single record that gets updated as the status changes? I think we probably want to have at least created/updated timestamps to be able to know when the job failed, succeed, how long it's been running if it's Started, etc.

api/src/db/models/opportunity_models.py
```diff
@@ -452,10 +452,12 @@ class OpportunityAttachment(ApiSchemaTable, TimestampMixin):
...
-    __tablename__ = "opportunity_search_index_queue"
+class OpportunityChangeAudit(ApiSchemaTable, TimestampMixin):
+    __tablename__ = "opportunity_change_audit"
 
     opportunity_id: Mapped[int] = mapped_column(
         BigInteger, ForeignKey(Opportunity.opportunity_id), primary_key=True, index=True
     )
     opportunity: Mapped[Opportunity] = relationship(Opportunity)
+
+    has_update: Mapped[bool]
```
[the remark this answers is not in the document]
mikehgrantsgov: Yes, removed.

api/tests/src/search/backend/test_load_opportunities_to_index.py
```diff
@@ -314,28 +308,52 @@ def test_load_opportunities_to_index_index_does_not_exist(self, db_session, sear
...
 
         load_opportunities_to_index.run()
 
         # Verify queue was cleared
-        remaining_queue = db_session.query(OpportunitySearchIndexQueue).all()
+        remaining_queue = (
+            db_session.execute(
+                select(OpportunityChangeAudit).where(
+                    OpportunityChangeAudit.opportunity_id == test_opportunity.opportunity_id,
+                    OpportunityChangeAudit.updated_at.is_(None),
```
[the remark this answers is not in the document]
mikehgrantsgov: Ah yeah that was leftover from `last_loaded_at` which was nullable, but changed to `updated_at`. Updated these queries.

api/src/db/models/task_models.py
mikehgrantsgov: @mdragon correct and yes, the `TimestampMixin` provides `created_at`/`updated_at` for Jobs, so we can always diff `updated_at` and `created_at` to find the total job run time (or pending job run time if the status is not yet completed).

api/src/task/task.py
```diff
@@ -45,8 +56,16 @@ def run(self) -> None:
...
 
+            # Update job status to completed
+            self.update_job(JobStatus.COMPLETED, metrics=self.metrics)
+            self.job.metrics = self.metrics
+            self.db_session.commit()
+
             logger.info("Completed %s in %s seconds", self.cls_name(), duration, extra=self.metrics)
         except Exception:
+            # Update job status to failed
+            self.update_job(JobStatus.FAILED, metrics=self.metrics)
```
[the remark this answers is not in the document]
mikehgrantsgov: @chouinar I think the recent adds to `test_task.py` cover the scenarios you mention here. If a db error is thrown in `task` it will rollback the transaction (which I don't think SQLAlchemy does automatically?) and begin a new transaction so the `FAILED` state can be stored for the job.

api/src/task/task.py
```diff
@@ -28,25 +31,46 @@ class Metrics(StrEnum):
...
+                # Run the actual task
+                self.run_task()
+            except SQLAlchemyError:
+                # Rollback and begin new transaction only for DB-specific errors
+                self.db_session.rollback()
+                self.db_session.begin()
+                raise
+            except Exception:
+                # For non-DB errors, just raise without touching the transaction
+                raise
```
chouinar, line 59: In what case would we want a non-DB error to still commit changes? If the task is failing, I'd rather we just rollback anyways and not commit. A partial success state seems more problematic, and if any such scenarios do exist, we should handle that on a particular task.

What about a pattern like this:

```py
job_succeeded = True
try:
     self.run_task()
     ... # timing stuff
except Exception:
     self.db_session.rollback() # May or may not be necessary?
     job_succeeded = False
     raise

finally:
    job_status = JobStatus.COMPLETED if job_succeeded else JobStatus.FAILED
    with db_session.begin():
         self.update_job(job_status, metrics=self.metrics)
```

Would this hit any issues?

api/src/task/task.py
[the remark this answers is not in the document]
chouinar: SQLAlchemy doesn't rollback automatically, but if it fails to get to the end of the transaction, it should at least fail to commit.

api/src/task/task.py
mikehgrantsgov: If run_task raises an exception, the timing stuff and metrics would not run. Not sure if that's a big issue, though. I think the rollback() is not necessary here since the task would handle its own tx stuff (right? we are not relying on this task class for tx management?), and the `finally` block would start its new transaction with `db_session.begin`.

Otherwise I don't see any other issues here.

mikehgrantsgov: I don't think this is quite right yet, but @chouinar if you have any ideas on this, maybe we can pair to make this implementation more elegant.

api/src/task/task.py
```diff
@@ -28,27 +29,44 @@ class Metrics(StrEnum):
...
 
             # Initialize the metrics
             self.initialize_metrics()
 
             # Run the actual task
             self.run_task()
 
+            logger.info("Starting %s", self.cls_name())
+            start = time.perf_counter()
+
```
chouinar, line 51: Why did this get moved after the `run_task` command? Should be the first thing in the try block

api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -163,12 +178,12 @@ def _handle_incremental_upserts(self, existing_opportunity_ids: set[int]) -> Non
...
 
-        # Clear processed / skipped entries from the queue
-        self.db_session.execute(
-            delete(OpportunitySearchIndexQueue).where(
-                OpportunitySearchIndexQueue.opportunity_id.in_(processed_opportunity_ids)
+            # Update updated_at timestamp instead of deleting records
+            self.db_session.execute(
+                update(OpportunityChangeAudit)
+                .where(OpportunityChangeAudit.opportunity_id.in_(processed_opportunity_ids))
+                .values(updated_at=get_now_us_eastern_datetime())
```
chouinar, line 185: In the DB, we use UTC for everything, I would make this call the UTC function, not the eastern one

chouinar (commented):
Just two small things, sorry for the slow reviews on this PR

api/src/task/task.py
mikehgrantsgov: Ah was moving things around testing something out. Moved it back.

chouinar (approved):

PR #3567, opened 2025-01-17
[Unticketed] Fix issue where opportunities aren't loaded to search index

chouinar:
## Changes proposed
Tell processor to allow the record even if attachments are missing

## Context for reviewers
https://www.elastic.co/guide/en/elasticsearch/reference/current/foreach-processor.html

ignore_missing is defaulted to False. This means if the record has no attachments, it just drops it and moves on. This fix makes it not drop it, and just allow it.

## Additional information
Set `ENABLE_OPPORTUNITY_ATTACHMENT_PIPELINE=false` in the local.env

Run `make db-seed-local` and `make populate-search-opportunities`, without this fix, nothing ends up in the search index.

### DIFF
--- a/api/src/search/backend/load_opportunities_to_index.py
+++ b/api/src/search/backend/load_opportunities_to_index.py
@@ -97,7 +97,7 @@ class LoadOpportunitiesToIndex(Task):
                                 "field": "_ingest._value.data",
                             }
                         },
-                        "ignore_missing": True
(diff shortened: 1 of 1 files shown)


babebe (approved):

babebe (approved):
LGTM

acouch (approved):
Can confirm this fixes the search server error locally.

PR #3568, opened 2025-01-17
[Issue #3203] Update GitHub API validation

### LINKED ISSUE #3203: Update analytics export to not use file system

widal001:
## Summary

Changes how we validate the response data from the GitHub API.

Fixes #3203 

- Adds new schemas to validate the GitHub output
- Uses those schemas to modify the `transform_project_data()` method

This PR is stacked on top of #3393

### DIFF
--- a/analytics/src/analytics/integrations/github/main.py
+++ b/analytics/src/analytics/integrations/github/main.py
@@ -6,31 +6,17 @@ we disable writing to disk in https://github.com/HHS/simpler-grants-gov/issues/3
 """
 
 import json
-from datetime import datetime, timedelta
+import logging
 from pathlib import Path
-from typing import Any
+
+from pydantic import ValidationError
 
 from analytics.integrations.github.client import GitHubGraphqlClient
+from analytics.integrations.github.validation import CombinedProjectItem
+
+logger = logging.getLogger(__name__)
 
 PARENT_DIR = Path(__file__).resolve().parent
-
-
-def safe_pluck(data: dict, keys: str) -> Any:  # noqa: ANN401
-    """
-    Safely pluck a value from a nested dictionary using a string of dot-separated keys.
-
-    Example:
-        data = {"content": {"parent": {"url": "example.com"}}}
-        safe_pluck(data, "content.parent.url") # returns "example.com"
-        safe_pluck(data, "content.missing") # returns None
-
-    """
-    current = data
-    for key in keys.split("."):
-        if not isinstance(current, dict) or key not in current:
-            return None
-        current = current[key]
-    return current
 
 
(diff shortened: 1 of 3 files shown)


analytics/src/analytics/integrations/github/validation.py
```diff
@@ -0,0 +1,119 @@
...
+    """
+    Replace keys that are explicitly set to None with an empty dict for default_factory.
+
+    We need to do this because if a key is present, but its value is None or null,
+    it will raise a Pydantic ValidationError rather than using the default_factory.
+    """
+    for key in keys_to_replace:
+        if data.get(key) is None:
+            data[key] = {}
+    return data
```
widal001, line 19: @coilysiren @chouinar I tried a number of ways to get around using something like this, but without spending another 1-2 hours researching this seems to be the best way to force a schema validator to use the `default_factory` when a key is present, but explicitly set to `None` (`null` in the JSON).

analytics/src/analytics/integrations/github/validation.py
```diff
@@ -0,0 +1,119 @@
...
+    created_at: str = Field(alias="createdAt")
+    closed_at: str | None = Field(alias="closedAt", default=None)
+    issue_type: IssueType = Field(alias="type", default_factory=IssueType)
+    parent: IssueParent = Field(default_factory=IssueParent)
+
+    @model_validator(mode="before")
+    def replace_none_with_defaults(cls, values) -> dict:  # noqa: ANN001, N805
+        """Replace None with default_factory instances."""
+        # Replace None with default_factory instances
+        return safe_default_factory(values, ["type", "parent"])
```
widal001, line 55: Here's where we apply that `safe_default_factory()` to catch and replace the keys for which we want to use the default factory if they're explicitly set to `None`

analytics/src/analytics/integrations/github/validation.py
coilysiren: This is reasonable to me 👍🏼

analytics/src/analytics/integrations/github/main.py
```diff
@@ -40,54 +26,53 @@ def transform_project_data(
...
+                continue
+
+            # Transform into flattened format
+            transformed = {
+                # project metadata
+                "project_owner": owner,
+                "project_number": project,
+                # issue metadata
+                "issue_title": validated_item.content.title,
+                "issue_url": validated_item.content.url,
```
widal001, line 47: This does look a lot cleaner!

coilysiren (approved):
Yeah! This is what I expected to see

PR #3574, opened 2025-01-17
[Issue #3553] Modify local s3 setup in factories to be easier to use

### LINKED ISSUE #3553: Investigate local s3 setup to see if there is a way for the factories to create the files they work with
### Summary

TODO - the current setup is obnoxious, needs a better approach, see if we can somehow configure the factories to work with s3?

### Acceptance criteria

- [ ] Figure out a better approach

chouinar:
## Summary
Fixes #3553

## Changes proposed
Modified our attachment factory to create files on s3 automatically

## Context for reviewers
We had a lot of tests and local setup assume files would already be pre-generated that the opportunity attachment table would point to. Those often were wrong and broke.

This fixes things by instead making the factories create s3 files for you. It does this by taking the path you gave it, and making a file at that location. This does make assumptions that your s3 bucket exists, but I piggybacked on the `enable_factory_create` pytest fixture we have that already requires you call it to call any DB operation with create, so that shouldn't be too much of an issue. I've set it up so the factory by default now creates no attachments on an opportunity, but can be told to add a few because the performance degradation on running tests was pretty bad.

## Additional information
Running `make db-seed-local` works. You can see the files it created with `aws --endpoint-url http://localhost:4566 s3 ls local-mock-public-bucket --recursive`

The tests all work with just a few minor fixes.


api/tests/conftest.py
```diff
@@ -373,7 +349,7 @@ def mock_s3(reset_aws_env_vars):
 
 @pytest.fixture
 def mock_s3_bucket_resource(mock_s3):
-    bucket = mock_s3.Bucket("test_bucket")
+    bucket = mock_s3.Bucket("local-mock-public-bucket")
```
babebe, line 352: description here for when to use one vs the other would help.

api/tests/src/db/models/factories.py
```diff
@@ -776,6 +754,59 @@ class Meta:
...
+    file_size_bytes = factory.Faker("random_int", min=1000, max=10000000)
+    opportunity_attachment_type = factory.fuzzy.FuzzyChoice(OpportunityAttachmentType)
+
+    created_at = factory.Faker("date_time_between", start_date="-1y", end_date="now")
+    updated_at = factory.LazyAttribute(
+        lambda o: fake.date_time_between(start_date=o.created_at, end_date="now")
+    )
+
+    @classmethod
+    def _build(cls, model_class, *args, **kwargs):
```
babebe, line 784: Since it's not part of the model? neat!

babebe (commented):
Thank you. This will definitely make it easier to create attachments!!! Also, much easier to get s3 session to work locally. We can call it directly when signing the url as well.

api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -286,7 +286,7 @@ def get_attachment_json_for_opportunity(
 
         attachments = []
         for att in opp_attachments:
-            with smart_open.open(
+            with file_util.open_stream(
```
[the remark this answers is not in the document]
chouinar: Hadn't pulled the latest main before making this branch, fixed

api/tests/src/db/models/factories.py
chouinar: FYI - There are [better ways](https://factoryboy.readthedocs.io/en/stable/reference.html#factory.FactoryOptions.exclude) of putting extra parameters on models for temporarily using, but none that are available at build/create.

api/tests/conftest.py
chouinar: You're free to use either however you see fit - there's not any real specific guidance. I can add a note to the other one to say that it exists because in some cases we want to test across buckets.

api/tests/src/db/models/factories.py
babebe: Yup!

babebe (approved):
LGTM

PR #3590, opened 2025-01-21
[Issue #3097] Fix analytics deploy not having a CDN

coilysiren:
## Summary

Accidentally caused by work on #3097

## Changes proposed

Gaurds against the cause where a given service doesn't have a CDN to deploy _(which is only analytics right now)_

## Testing

Analytics deploy was broken before this change, and now it's not. I haven't tested API and frontend, but I'm hoping this doesn't break them.


mdragon (approved):
👍🏻

PR #3593, opened 2025-01-21
[Unticketed] Fix alignment for header, update logo loading, fix logo for opp pages

acouch:
### Fix header alignment

Current:
image

Fix:
image

image

### Update header image to use svg and Next Image component and fix logo on opportunity pages

Since we have SVG logo images, we should use them since they load the correct size for the image and don't require a `srcet` attribute and different size images to be included to optimize the images for different screen sizes, which we weren't using anyway.

The [unoptimized](https://nextjs.org/docs/app/api-reference/components/image#unoptimized) prop allows us to load the image directly, and the [fill](https://nextjs.org/docs/app/api-reference/components/image#fill) prop sets the image not to use widths and heights set by Next, so the SVGs are loaded optimally.

current:

image

fix:

image

### Remove `assetPath` since the header was the only active component using that

This was part of the initial setup we inherited, and might have been necessary in the past to work with SVG?


frontend/src/components/Header.tsx
```diff
@@ -168,18 +167,18 @@ const Header = ({ locale }: Props) => {
...
               <div className="display-flex flex-align-center">
-                <Link href="/" className="display-block">
-                  {logoPath && (
-                    <img
-                      src={assetPath(logoPath)}
-                      alt={t("title")}
-                      className="display-block height-4  desktop:height-auto"
-                    />
-                  )}
+                <Link href="/" className="position-relative">
```
acouch, line 174: Next requires adding a relative position width because it automatically sets `position:absolute` to images with a `fill` prop set.

frontend/src/components/Header.tsx
```diff
@@ -168,18 +167,18 @@ const Header = ({ locale }: Props) => {
         basic={true}
         className="desktop:position-sticky top-0 desktop:z-500 bg-white border-bottom-2px border-primary-vivid"
       >
-        <div className="usa-nav-container display-flex flex-align-end">
+        <div className="usa-nav-container display-flex flex-justify">
```
acouch, line 170: This was the fix for the header alignment

doug-s-nava (commented):
I'm seeing this in the footer on mobile width
Screenshot 2025-01-22 at 11 09 09 AM

frontend/src/components/Header.tsx
andycochran: Oh, weird. This must've been a regression from the auth work @doug-s-nava (cc) did. Let's make sure the layout still works with the `authOn` feature flag enabled.

andycochran:
Nice! I was just thinking about how I regret using both SVG and PNG when I applied the new brand.

acouch:
Fixed mobile update:
image

image

frontend/src/components/Header.tsx
acouch: not @doug-s-nava 
ididit

acouch: image

doug-s-nava (approved):
✅ ✅ ✅ ✅

PR #3594, opened 2025-01-21
[Unticketed] Enable context data logging in New Relic

chouinar:
## Changes proposed
Enable https://docs.newrelic.com/docs/apm/agents/python-agent/configuration/python-agent-configuration/#application_logging.forwarding.context_data.enabled

## Context for reviewers
If I'm reading docs right, this should make it so New Relic automatically includes the `extra` we add to our logs. It already is getting the logs from the API and backend tasks successfully, but only includes the message. This should fix that?


mdragon (approved):
👍🏻

PR #3599, opened 2025-01-22
[Unticketed] Add devcontainer for frontend QA

acouch:
## Changes proposed
Adds a devcontainer so non-devs or devs who don't want to spin up the site locally can QA frontend changes that don't require the API


mdragon (approved):
Approved

PR #3629, opened 2025-01-23
[Issue #3628] allow error page to be interactive without refresh

### LINKED ISSUE #3628: Search page is unresponsive after error
### Summary

When the search page gets into an error state the filters and query bar are unusable until a refresh.

### Reproducibility

### How to reproduce the bug

* visit a page that will put you in an error state such as https://simpler.grants.gov/search?status=fake
* click filter buttons or attempt to submit a search query
* url updates but nothing happens in the UI

### Browser

None

### Code of Conduct

doug-s-nava:
## Summary
Fixes #3628

## Changes proposed
Fixes bug where search error page had an unusable UI

Also introduces a usePrevious hook as pulled from https://www.developerway.com/posts/implementing-advanced-use-previous-hook (I think this hook implementation used to be in the React docs themselves, but I don't see it there anymore)

## Context for reviewers

This was fixed by correctly using NextJs's `reset` method which is not well documented https://nextjs.org/docs/app/api-reference/file-conventions/error#reset

Because NextJS's error handling uses React error boundaries under the hood, which are basically designed to set an error state and take action based on that state, it makes sense that something explicit must be done to tell the error boundary that the app is no longer in an error state. That's what reset is for, so by calling reset whenever a user interaction occurs on the error page, the problem disappears.

### Test steps

1. start on main with `npm run dev`
2. visit http://localhost:3000/search?status=fake
3. _VERIFY_: you're on the error page
4. click a filter or attempt to do a query based search
5. _VERIFY_: url updates but UI does not update
6. restart on this branch
2. visit http://localhost:3000/search?status=fake
3. _VERIFY_: you're on the error page
4. click a filter or attempt to do a query based search
5. _VERIFY_: the page performs the search and updates the UI

### DIFF
--- a/frontend/src/app/[locale]/search/error.tsx
+++ b/frontend/src/app/[locale]/search/error.tsx
@@ -1,12 +1,14 @@
 "use client";
 
 import QueryProvider from "src/app/[locale]/search/QueryProvider";
+import { usePrevious } from "src/hooks/usePrevious";
 import { FrontendErrorDetails } from "src/types/apiResponseTypes";
 import { ServerSideSearchParams } from "src/types/searchRequestURLTypes";
 import { Breakpoints, ErrorProps } from "src/types/uiTypes";
 import { convertSearchParamsToProperTypes } from "src/utils/search/convertSearchParamsToProperTypes";
 
 import { useTranslations } from "next-intl";
+import { ReadonlyURLSearchParams, useSearchParams } from "next/navigation";
 import { useEffect } from "react";
 import { Alert } from "@trussworks/react-uswds";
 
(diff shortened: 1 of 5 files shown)


frontend/src/app/[locale]/search/error.tsx
```diff
@@ -51,8 +53,25 @@ function createBlankParsedError(): ParsedError {
...
+      previousSearchParams &&
+      searchParams.toString() !== previousSearchParams?.toString()
+    ) {
+      reset();
+    }
+  }, [searchParams, reset]);
+
+  useEffect(() => {
+    console.error(error);
+  }, [error]);
```
Comment (acouch) [inline frontend/src/app/[locale]/search/error.tsx:74]: Could we use or extend one of the errors from https://github.com/HHS/simpler-grants-gov/blob/main/frontend/src/errors.ts so we have more consistent error formatting and handling?

acouch (approved):
👍🏿 with question about the error reporting.

frontend/src/app/[locale]/search/error.tsx
Comment (acouch) [reply frontend/src/app/[locale]/search/error.tsx:74]: We could punt to this https://github.com/HHS/simpler-grants-gov/issues/3651 if that is what yr thinking

Comment (doug-s-nava) [reply frontend/src/app/[locale]/search/error.tsx:74]: the errors thrown here largely will be based on the definitions in our errors file, so not sure what to do here

Comment (acouch) [reply frontend/src/app/[locale]/search/error.tsx:74]: I would think we would wrap the output. I guess I don't see how this error is formed or called.

PR #3637, opened 2025-01-23
[Issue #3156] fetch options for agency filter from api

### LINKED ISSUE #3156: When creating Agency filter, pull Agency list from API endpoint
### Summary

We need to offer a list of agencies and their details (specifically name and agency_code) that will allow for Frontends to present lists of agencies to users. The one existing use case for this is the Agency filter list on the Search page.

### Acceptance criteria

- [ ] Search page, Agency filter reflects agencies configured in that environment, instead of a hard coded list of Prod Agencies.

doug-s-nava:
## Summary
Fixes #3156

## Changes proposed

The goal of the change is to allow the frontend to serve a list of agencies to filter on that are up to date with the list of agencies in the API rather than relying on a hardcoded list. To do this, we had to do a bunch of things:

* create fetcher for agencies and utility functions for translating the response into filter options
* create an async component for fetching the agency list
* restructure the search error page to allow it to work with the asynchronous nature of the agencies filter. This involves
  * no longer relying on parsing the error message to obtain form state on the error page. This would have always been broken in prod where error messages are not preserved. Using search params instead gets us more consistent results
  * storing the agency list in global state so that they can be retrieved by the client side error page without a fetch to the API 
  * adding zustand for global state management. We may want to discuss whether that's necessary or not at this point

#### Agency filter functionality
1. start on this branch with `npm run dev`
2.


.github/workflows/ci-frontend-a11y.yml
```diff
@@ -39,7 +39,7 @@ jobs:
       - name: Start API Server for search results
         run: |
           cd ../api
-          make init db-seed-local start &
+          make init db-seed-local populate-search-opportunities start &
```
doug-s-nava, line 42: leaving this out was causing failures on the pa11y search page tests, but not sure why it only needs to be added now

frontend/src/app/[locale]/search/error.tsx
```diff
@@ -32,51 +41,34 @@ function isValidJSON(str: string) {
...
-      actionType: "initialLoad",
-    },
-    message: "Invalid error message JSON returned",
-    status: -1,
-  };
-}
-
+// note that the SearchFilters component is not used here since that is a server component
+// we work around that by including the rendered components from SearchFilters, but manually
+// passing through the agency options as received from global state rather than fetching from API
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/search/error.tsx:46]: after discussing with Matt I realized that we could work around this by not using the Next's `error.js` pattern but use a normal React ErrorBoundary in a more targeted way. I think this is fine for now, but something to consider if this gets too hard to manage

frontend/src/components/search/SearchResultsListItem.tsx
```diff
@@ -100,13 +103,7 @@ export default function SearchResultsListItem({
...
               <span className={metadataBorderClasses}>
                 <strong>{t("resultsListItem.summary.agency")}</strong>
-                {opportunity?.agency ||
-                  (opportunity?.summary?.agency_name &&
-                  opportunity?.summary?.agency_code &&
-                  agencyNameLookup
-                    ? // Use same exact label we're using for the agency filter list
-                      agencyNameLookup[opportunity?.summary?.agency_code]
-                    : "--")}
+                {lookUpAgencyName(opportunity, flattenedAgencyOptions)}
```
doug-s-nava, line 106: this functionality to map the code from the search response onto the human readable agency name was broken before since the lookup was never being passed in. I've asked to make sure that design wants it restored.

frontend/tests/services/fetch/fetchers/agenciesFetcher.test.ts
```diff
@@ -0,0 +1,234 @@
...
+      {
+        id: "FAKE",
+        label: "Completely fake",
+        value: "FAKE",
+      },
+    ]);
+  });
+
+  it("converts a complex list of nested agencies to filter options", () => {
+    const fakeAgencyResponseData: RelevantAgencyRecord[] = [
```
doug-s-nava, line 94: should we consider throwing some of this into fixture files to make the tests more readable?

frontend/src/components/search/SearchResultsListItem.tsx
mdragon: I had created https://github.com/HHS/simpler-grants-gov/issues/3646 to track this same thing from a different angle, so feel free to either leave this for that ticket or we can close that ticket and address it here.

doug-s-nava: I'm thinking we leave this in and then follow up to add the top level agency name?

frontend/src/services/fetch/fetchers/agenciesFetcher.ts
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// would have called this getAgencies, but technically it's a POST
+export const obtainAgencies = async (): Promise<RelevantAgencyRecord[]> => {
+  const response = await fetchAgencies({
+    body: {
+      pagination: {
+        order_by: "created_at", // this seems to be the only supported option
+        page_offset: 1,
+        page_size: 100, // should fetch them all. db seems to have 74 records as of 1/17/25
```
mdragon, line 22: In Prod there's more like 700 I believe.

chouinar: There are ~1150 in prod at the moment, it's about 700 with an opportunity.

Should we add a filter rule to this endpoint to only return agencies with an opportunity?

coilysiren (commented):
defering review since this is a case of github codeowners being incorrect

frontend/src/app/[locale]/search/error.tsx
Comment (doug-s-nava) [reply frontend/src/app/[locale]/search/error.tsx:47]: actually this kind of approach wouldn't really work since error boundaries require class components which don't work in Next server components 🤷

doug-s-nava:
This is on hold for now as we continue to discuss implementation options

acouch (dismissed):
👍🏿

acouch (approved):

acouch:
Pulling in for @doug-s-nava as he is on 🏖️

PR #3639, opened 2025-01-24
[Issue #3536] Add saved opportunity notifications to backend job

### LINKED ISSUE #3536: Add saved opportunity notifications to backend job
### Summary

Requires #3527 and #3534 to be completed first

We want to detect and notify when an opportunity is changed, and add it to our logic.

To our `user_saved_opportunity` table we want to add a column called `last_notified_at`, a timestamp. The value for this column should default to the current timestamp - this can be done by doing:

```py
last_notified_at: Mapped[datetime] = mapped_column(
        nullable=False,
        default=datetime_util.utcnow,
        server_default=sqlnow(),
    )
```
---

In the notification job, if last_notified_at is ever before the last modified timestamp from the update table (changed in #3527 ) then we want to notify. This info can be fetched in a query joining the opportunity and user_saved_opportunity records together. Note that it is possible that one user might be notified, and another might not be due to timing (eg. user A saves opp last week. user B saves opp today. Opp was updated today before user B saved it). This is fine.

Note that we only want to notify after processing everything, we don't want to notify as we go, so the opportunities per user mapping should be updated as we go (added in #3534 )

Not in scope for this ticket:
* Actual notifications - just log a message for now to "notify" in the notify function
* Knowing whether a user wants to receive notifications, we'll add rules around that later, assume everyone wants them for now

### Acceptance criteria

- [ ] Table updated
- [ ] Backend script updated
- [ ] Tests added (note you can detect "notifications" by using caplog)

mikehgrantsgov:
## Summary
Fixes #3536

## Changes proposed
Add `last_notified_at` to `user_saved_opportunity`
Collect opportunity notifications, log them out and test.

## Context for reviewers
Can be merged, now that #3527 is wrapped up. 

## Additional information
See attached unit tests


api/tests/src/task/notifications/test_generate_notifications.py
```diff
@@ -1,5 +1,63 @@
...
+    assert result.exit_code == 0
+
+    # Verify expected log messages
+    assert "Collected opportunity notifications" in caplog.text
+    assert "Would send notification to user" in caplog.text
+
+    # Verify the log contains the correct metrics
+    log_records = [r for r in caplog.records if "Would send notification to user" in r.message]
+    assert len(log_records) == 1
+    extra = log_records[0].__dict__
```
chouinar, line 34: If we want tests that are a bit less log-focused at the moment, I did add the notification table which we'll use as a sort of auditing table. Could start populating that with something.

api/src/task/notifications/generate_notifications.py
```diff
@@ -55,11 +59,39 @@ def run_task(self) -> None:
...
+            select(User.user_id, UserSavedOpportunity.opportunity_id)
+            .join(
+                UserSavedOpportunity,
+                User.user_id == UserSavedOpportunity.user_id,
+            )
+            .join(
+                Opportunity,
+                UserSavedOpportunity.opportunity_id == Opportunity.opportunity_id,
+            )
+            .where(Opportunity.updated_at > UserSavedOpportunity.last_notified_at)
```
chouinar, line 73: Should this be compared against the timestamp from the `updated_at` column in the opportunity change log table from your other PR?

api/src/task/notifications/generate_notifications.py
```diff
@@ -55,11 +59,39 @@ def run_task(self) -> None:
...
-        To be implemented in future ticket
-        """
-        logger.info("Opportunity notification collection not yet implemented")
-        pass
+        """Collect notifications for changed opportunities that users are tracking"""
+        stmt = (
+            select(User.user_id, UserSavedOpportunity.opportunity_id)
+            .join(
+                UserSavedOpportunity,
+                User.user_id == UserSavedOpportunity.user_id,
```
chouinar, line 67: What if this was `select(UserSavedOpportunity)...` with similar joins. You'd avoid needing to use mappings/specific columns since the UserSavedOpportunity record is the user_id + opportunity_id.

We should make the user_notification_map opportunity IDs instead just be this class object. Might also simplify the updates later?

api/src/task/notifications/generate_notifications.py
mikehgrantsgov: Update this, thanks

api/tests/src/task/notifications/test_generate_notifications.py
mikehgrantsgov: Added some new tests here to and adding logs in `generate_notifications.py`

api/src/task/notifications/generate_notifications.py
```diff
@@ -103,32 +98,42 @@ def _collect_search_notifications(self) -> None:
...
                 },
             )
 
+            # Create notification log entry
+            # TODO: Use enum for notification reason?
+            notification_log = UserNotificationLog(
+                user_id=user_id,
+                notification_reason="opportunity_updates",
+                notification_sent=True,
+            )
```
chouinar, line 120: For the notification reason, I wasn't sure how notifications might evolve over time so left it pretty freeform. For this PR and the later one to add notifications for search, I'd probably just suggest adding some constants at the top of the file like:
```
class NotificationConstants:
     OPPORTUNITY_UPDATES = "opportunity_updates"
```

At least organizes it slightly kinda like an enum

chouinar (dismissed):
LGTM! Just adding an idea about a small TODO you had

api/src/task/notifications/generate_notifications.py
chouinar: Yes, you adjusted it like I was thinking

chouinar (approved):

chouinar (approved):

Issue #3545, opened 2025-01-16
Support relative date ranges in search


chouinar:
### Summary

This requires two sets of changes:
* Our API schema needs to allow relative date ranges (negative/positive integers representing days)
* We need to be able to pass in relative date ranges to the search index

I'd recommend first reading the docs on date ranges: https://opensearch.org/docs/latest/query-dsl/term/range/ and manually determining what the query needs to be to do this.

For the API schema, we currently support:
```json
    "post_date": {
      "end_date": "2024-02-01",
      "start_date": "2024-01-01"
    }
```

We should allow:
```json
    "post_date": {
      "end_date_relative": -15,
      "start_date_relative": 15
    }
```
These should be mutually exclusive, which may require some more complex rules on the schema than what we previously have done.

### Acceptance criteria

- [ ] API supports relative date range (specific range still allowed/supported)
- [ ] Search client updated to support relative date range
- [ ] Tests added to validate this behavior

PR #3640, opened 2025-01-24
[Issue 3545] relative date search

babebe:
## Summary
Fixes #{[3545](https://github.com/HHS/simpler-grants-gov/issues/3545)}

## Changes proposed
API supports relative date range (specific range still allowed/supported)
Search client updated to support relative date range
Tests


api/src/api/schemas/search_schema.py
```diff
@@ -302,16 +310,35 @@ def _with_date_range_validator(self) -> "DateSearchSchemaBuilder":
...
+                "start_date_relative" in data or "end_date_relative" in data
+            ):
+                raise ValidationError(
+                    [
+                        MarshmallowErrorContainer(
+                            ValidationErrorType.INVALID,
+                            "Cannot have both absolute and relative date parameters.",
+                        )
+                    ]
+                )
```
chouinar, line 331: I actually wonder if we should disallow this. We should definitely error if both absolute and relative for the same value are passed in, but if someone passed in a relative start, and absolute last, would that really be an issue (assuming OpenSearch doesn't complain)?

api/src/services/opportunities_v1/search_opportunities.py
```diff
@@ -127,7 +127,13 @@ def _add_search_filters(
             builder.filter_int_range(field_name, field_filters.min, field_filters.max)
 
         elif isinstance(field_filters, DateSearchFilter):
-            builder.filter_date_range(field_name, field_filters.start_date, field_filters.end_date)
+            builder.filter_date_range(
+                field_name,
+                field_filters.start_date,
+                field_filters.end_date,
+                field_filters.start_date_relative,
+                field_filters.end_date_relative,
```
chouinar, line 135: If you go with the idea I had on the query builder, this would become something like:
```
start_date=field_filters.start_date if field_filters.start_date else field_filders.start_date_relative,
end_date=field_filters.end_date if field_filters.end_date else field_filders.end_date_relative,
```

api/src/api/schemas/search_schema.py
babebe: Opensearch does allow for both to be passed at the same time. Though i am not sure if it is optimal or will result in some inconsistencies i can run more tests to test these.

api/src/adapters/search/opensearch_query_builder.py
```diff
@@ -195,7 +195,12 @@ def filter_int_range(
         return self
 
     def filter_date_range(
-        self, field: str, start_date: datetime.date | None, end_date: datetime.date | None
+        self,
+        field: str,
+        start_date: datetime.date | None,
+        end_date: datetime.date | None,
+        start_date_relative: int | None,
+        end_date_relative: int | None,
```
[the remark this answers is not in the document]
babebe: Nice formatting ! Will adapt this, def more readable.

chouinar (approved):

PR #3659, opened 2025-01-28
[Issue #3094] Use CDN URLs for attachments instead of S3 URLs

### LINKED ISSUE #3094: Use CDN URLs for attachments instead of S3 URLs
### Summary

Once the CDN is in place and in front of S3, we'll need to update the document links in API responses to reference the CDN address for each document.

### Acceptance criteria

- [ ] API responses for attached documents reference the CDN URL, not a pre-signed S3 URL.

mikehgrantsgov:
## Summary
Fixes #3094

## Changes proposed
Branch logic for attachment URLs whether CDN_URL is set or not. If set, use CDN_URL to replace s3://<foo>/<path_to_file>

## Context for reviewers
Instead of changing actual values in the DB, evaluate at run-time if the CDN_URL is set and use that value instead of presigning S3 URLS.

## Additional information
One thing to figure out: `AttachmentConfig` relies on *.env files being set, but doesn't appear to work when setting ENV vars via Monkeypatch.


api/src/util/file_util.py
```diff
@@ -166,3 +166,17 @@ def read_file(path: str | Path, mode: str = "r", encoding: str | None = None) ->
...
+    """
+    Convert an S3 URL to a CDN URL
+
+    Example:
+        s3://bucket-name/path/to/file.txt -> https://cdn.example.com/path/to/file.txt
+    """
+    if not is_s3_path(file_path):
+        raise ValueError(f"Expected s3:// path, got: {file_path}")
+
+    _, key = split_s3_url(file_path)
```
chouinar, line 181: Can we use a str replace? - I want to be very very careful that we only do this for the public files bucket.

api/src/util/file_util.py
```diff
@@ -166,3 +166,17 @@ def read_file(path: str | Path, mode: str = "r", encoding: str | None = None) ->
...
+    return f"{cdn_url.rstrip('/')}/{key}"
```
chouinar, line 182: Would `file_util.join(cdn_url, key)` accomplish the same thing?

api/tests/src/api/opportunities_v1/test_opportunity_route_get.py
```diff
@@ -153,3 +153,34 @@ def test_get_opportunity_404_not_found_is_draft(client, api_auth_token, enable_f
         resp.get_json()["message"]
         == f"Could not find Opportunity with ID {opportunity.opportunity_id}"
     )
+
+
+def test_get_opportunity_returns_cdn_urls(
+    client, api_auth_token, monkeypatch_session, enable_factory_create, db_session, mock_s3_bucket
+):
+    monkeypatch_session.setenv("CDN_URL", "https://cdn.example.com")
```
chouinar, line 161: Could we set this in some autouse conftest value? Otherwise we'll need to set it for every API test with attachments. Fine to override it here.

Also we should set CDN_URL in local.env to be `localhost:4566` and test that this works locally as well?

api/src/util/file_util.py
mikehgrantsgov: Updated to use this.

api/src/util/file_util.py
mikehgrantsgov: Agreed, added this for safety

api/tests/src/api/opportunities_v1/test_opportunity_route_get.py
mikehgrantsgov: Tried this locally and receive:
Screenshot 2025-01-28 at 2 48 17 PM

This would be the correct path is <s3_path> was replaced with <cdn_path>

api/src/services/opportunities_v1/get_opportunity.py
```diff
@@ -8,7 +8,13 @@
...
 from src.db.models.agency_models import Agency
 from src.db.models.opportunity_models import Opportunity, OpportunityAttachment, OpportunitySummary
-from src.util.file_util import pre_sign_file_location
+from src.util.env_config import PydanticBaseEnvConfig
+from src.util.file_util import convert_public_s3_to_cdn_url, pre_sign_file_location
+
+
+class AttachmentConfig(PydanticBaseEnvConfig):
+    # If the CDN URL is set, we'll use it instead of pre-signing the file locations
+    cdn_url: str | None = None
```
coilysiren, line 17: This doesn't need to be in capslock?

api/src/util/file_util.py
```diff
@@ -166,3 +166,16 @@ def read_file(path: str | Path, mode: str = "r", encoding: str | None = None) ->
...
+    """
+    Convert an S3 URL to a CDN URL
+
+    Example:
+        s3://bucket-name/path/to/file.txt -> https://cdn.example.com/path/to/file.txt
+    """
+    if not is_s3_path(file_path):
+        raise ValueError(f"Expected s3:// path, got: {file_path}")
+
+    return file_path.replace(os.environ["PUBLIC_FILES_BUCKET"], cdn_url)
```
coilysiren, line 181: Have you testing that this URL is exactly right?

api/src/services/opportunities_v1/get_opportunity.py
chouinar: [Pydantic](https://docs.pydantic.dev/latest/concepts/pydantic_settings/#case-sensitivity) will look for `CDN_URL` as an env var. We sometimes also give values an alias if we want a cleaner python name, but doesn't matter too much here.

api/src/services/opportunities_v1/get_opportunity.py
```diff
@@ -50,7 +56,14 @@ def get_opportunity(db_session: db.Session, opportunity_id: int) -> Opportunity:
...
 
-    pre_sign_opportunity_file_location(opportunity.opportunity_attachments)
+    attachment_config = AttachmentConfig()
+    if attachment_config.cdn_url is not None:
+        for opp_att in opportunity.opportunity_attachments:
+            opp_att.download_path = convert_public_s3_to_cdn_url(  # type: ignore
+                opp_att.file_location, attachment_config.cdn_url
+            )
+    else:
+        pre_sign_opportunity_file_location(opportunity.opportunity_attachments)
```
chouinar, line 66: I'm not sure there will be a case where it's not set once we merge this. (NOTE - we should first verify with @coilysiren that the CDN is fully working before we merge this - I think it's still in progress)

api/src/util/file_util.py
chouinar, line 181: Can you use the s3 config for this instead of the `os` module? Just so we only define the env vars in a single place

api/tests/src/api/opportunities_v1/test_opportunity_route_get.py
chouinar: If you go to that path, does the file download work? I think for local, it also needs the bucket name (ie. `CDN_URL` should include the bucket for our local dev purposes).

mikehgrantsgov: I added the bucket name to CDN_URL in local.env and it works. Checking that in shortly.

api/src/util/file_util.py
mikehgrantsgov: Yes we added tests to cover a few scenarios: https://github.com/HHS/simpler-grants-gov/pull/3659/files#diff-e284f21c2216feb0f0612d0cbed5d770d98a4fccb1591714ce50bb9dc47910d0R214

Also the only scenarios where S3 -> CDN replacement happens is explicitly for files in the public files bucket.

chouinar:
@mikehgrantsgov - FYI - the change looks good, but holding approval until we can confirm the CDN itself is actually working.

chouinar (approved):
LGTM - I think we're safe to merge this now.

After this is deployed, can you check that dev is using the CDN URL instead of presigned URLs?

Here's an opportunity with two attachments already on it: http://frontend-dev-1739892538.us-east-1.elb.amazonaws.com/opportunity/70111

PR #3682, opened 2025-01-29
[Issue #3579] Update opportunity CSV generation

### LINKED ISSUE #3579: Identify CSV columns and adjust API response
### Summary

* Expose Opportunity data for matching search results as a CSV
* Ideally without breaking the existing ordering/formatting (so that folks can work with our CSVs interchangeably).
* We could add new columns, different formats, on the end to help not break existing consumers

### Acceptance criteria

- [ ] Identify gaps between current grants.gov and Simpler implementations
- [ ] Confirm plan to close gaps with Product/Design
- [ ] Implement changes to generate the desired CSV.

chouinar:
## Summary
Fixes #3579

## Changes proposed
Remove a few columns from the opportunity CSV extract that shouldn't be used (bad data / duplicative)

For the search endpoint, add a maximum to the page size to prevent someone from using it to fetch everything (if they want that, we have extract files which will be better).

Slight adjustment to CSV writing logic

Cleaned up some old unused pagination code from an earlier implementation

## Context for reviewers
For fields removed:
* Agency == agency code, but been moving us to only use agency code for clarity on what the field actually is, so removing the extra duplicate
* Agency phone number has bad data and we already removed it from the search API itself

The small adjustment to the CSV writing logic is to have one less iteration over opportunities in the endpoint. No idea if it'll add much performance gain, but will use less memory/cycles for an easy change.

### DIFF
--- a/api/openapi.generated.yml
+++ b/api/openapi.generated.yml
@@ -837,6 +837,7 @@ components:
         page_size:
           type: integer
           minimum: 1
+          maximum: 5000
           description: The size of the page to fetch
           example: 25
         page_offset:
@@ -978,6 +979,7 @@ components:
         page_size:
           type: integer
           minimum: 1
+          maximum: 5000
           description: The size of the page to fetch
           example: 25
         page_offset:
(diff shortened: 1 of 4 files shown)


mdragon (approved):
Looks good

mikehgrantsgov (approved):

PR #3686, opened 2025-01-29
[Issue #3646] Better Agency display and sort in Search

### LINKED ISSUE #3646: Display Agency Name not Code in search results
### Summary

We should use the more explanatory Agency Name in search result cards not the short "Code" identifier.

Crystabel confirmed we should be using the name whenever we have it. 

### Acceptance criteria

- [ ] Agency Name is used on the Search Result card

mdragon:
## Summary
Fixes #3646

## Changes proposed
Display both the top level agency and agency name for search results. Sort accordingly when sorting by Agency.

## Context for reviewers
Search results now display <top level agency name> - <agency name> or just agency name if the agency is a top level agency. Sorting now takes into account that an agency sort should sort by both top level and agency name fields.


mdragon:
Revisiting this on a different branch with API support for multi-column sort

PR #3158, opened 2024-12-10
Ds/renovate i18next http backend

doug-s-nava:
## Summary
Fixes unticketed

## Changes proposed
This is an update to the renovate PR to update i18n-next https://github.com/HHS/simpler-grants-gov/pull/2595

The renovate PR pushes i18next-http-backend to 2.7, which has been deprecated by the maintainer. As 2.7 contained a breaking change and introduced a possible bug, we should be pointing at the latest version, 3.0.1 instead

see https://github.com/i18next/i18next-http-backend/issues/161

## Context for reviewers
Test steps can be found in gitbook, but I can't find them at the moment. Did the private gitbook go away?

In general, make sure you can build the app, and run the app and storybook locally

Issue #3518, opened 2025-01-14
Search results: Download CSV frontend implementation


snyder-ux:
### Summary

This ticket is for the frontend implementation of #2891. The API should already support download-results-as-csv.

Keep behind a feature flag until #3579 ships.

[Design in Figma](https://www.figma.com/design/FGxWtAgToKhehLJCiuy1zL/Simpler.Grants.gov-Project?node-id=1952-4077&t=MbgftMoz67yafc6g-1)

![Image](https://github.com/user-attachments/assets/25550537-c570-4239-be44-987d2bf8955f)

### Acceptance criteria

- [ ] Users can download search results as a CSV

PR #3689, opened 2025-01-30
[Issue #3518] download search results

doug-s-nava:
## Summary
Fixes #3158

## Changes proposed
The goal here is adding a button to the search page that will allow users to download a csv file containing all (well, the first 5000) search results.

To accomplish this a few larger refactors were made:

* fetch function service and helpers updated to be more flexible in order to allow a non-json response
* some flaky e2e tests updated

### Test steps
1. seed your database with a bunch more records. Update `seed_local_db.py` to up the size of each type by a bunch. Try updating `size=5` to `size=50`. This is important for testing with a full width pagination. Then run `make db-seed-local populate-search-opportunities`
3. _VERIFY_: you see an "Export results" button that matches designs
4. resize window to table and mobile widths
5. _VERIFY_: button placement and appearance matches expectations at each viewport width
6. click the button
7. _VERIFY_: a csv file downloads with filename `grants-search-<timestamp>.csv`)
8. open the file
9. _VERIFY_: the file contains the expected search result details in csv form

## Additional information
Screenshot 2025-01-31 at 4 25 59 PM

Screenshot 2025-01-31 at 4 25 30 PM

### DIFF
--- a/frontend/src/app/[locale]/search/error.tsx
+++ b/frontend/src/app/[locale]/search/error.tsx
@@ -3,7 +3,7 @@
 import QueryProvider from "src/app/[locale]/search/QueryProvider";
 import { usePrevious } from "src/hooks/usePrevious";
 import { FrontendErrorDetails } from "src/types/apiResponseTypes";
-import { ServerSideSearchParams } from "src/types/searchRequestURLTypes";
+import { OptionalStringDict } from "src/types/searchRequestURLTypes";
 import { Breakpoints, ErrorProps } from "src/types/uiTypes";
 import { convertSearchParamsToProperTypes } from "src/utils/search/convertSearchParamsToProperTypes";
 
@@ -19,7 +19,7 @@ import ServerErrorAlert from "src/components/ServerErrorAlert";
 
 export interface ParsedError {
   message: string;
-  searchInputs: ServerSideSearchParams;
+  searchInputs: OptionalStringDict;
   status: number;
   type: string;
   details?: FrontendErrorDetails;
--- a/frontend/src/app/api/search/export/route.ts
+++ b/frontend/src/app/api/search/export/route.ts
@@ -1,4 +1,4 @@
-import { searchForOpportunities } from "src/services/fetch/fetchers/searchFetcher";
+import { downloadOpportunities } from "src/services/fetch/fetchers/searchFetcher";
 import { convertSearchParamsToProperTypes } from "src/utils/search/convertSearchParamsToProperTypes";
 
 import { NextRequest } from "next/server";
@@ -10,8 +10,12 @@ export async function GET(request: NextRequest) {
     const searchParams = convertSearchParamsToProperTypes(
       Object.fromEntries(request.nextUrl.searchParams.entries().toArray()),
     );
-    const response = await searchForOpportunities(searchParams, true);
-    return response;
+    const apiResponse = await downloadOpportunities(searchParams);
+    return new Response(apiResponse, {
+      headers: {
+        "Content-Type": "text/csv",
+      },
+    });
   } catch (e) {
     console.error("Error downloading search results", e);
     throw e;
--- a/frontend/src/app/api/search/export/route.ts
+++ b/frontend/src/app/api/search/export/route.ts
@@ -1,0 +1,19 @@
+import { searchForOpportunities } from "src/services/fetch/fetchers/searchFetcher";
+import { convertSearchParamsToProperTypes } from "src/utils/search/convertSearchParamsToProperTypes";
+
+import { NextRequest } from "next/server";
+
+export const revalidate = 0;
+
+export async function GET(request: NextRequest) {
+  try {
+    const searchParams = convertSearchParamsToProperTypes(
+      Object.fromEntries(request.nextUrl.searchParams.entries().toArray()),
+    );
+    const response = await searchForOpportunities(searchParams, true);
+    return response;
+  } catch (e) {
+    console.error("Error downloading search results", e);
+    throw e;
+  }
+}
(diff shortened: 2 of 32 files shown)


frontend/src/services/fetch/fetchers/opportunityFetcher.ts
```diff
@@ -0,0 +1,11 @@
+import { OpportunityApiResponse } from "src/types/opportunity/opportunityResponseTypes";
```
doug-s-nava, line 1: since the bare fetcher no longer returns the json result, we'll create a basic wrapper here.

frontend/src/app/[locale]/search/error.tsx
```diff
@@ -19,7 +19,7 @@ import ServerErrorAlert from "src/components/ServerErrorAlert";
 
 export interface ParsedError {
   message: string;
-  searchInputs: ServerSideSearchParams;
+  searchInputs: OptionalStringDict;
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/search/error.tsx:22]: just a rename to reflect that this type is generic rather than specific in any way

frontend/src/app/api/search/export/route.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  ExportSearchResultsButton click ->
+  /export route ->
+  downloadOpportunities ->
+  fetchOpportunitySearch ->
+  ExportSearchResultsButton (handle response by blobbing it to the location) -> user's file system
+
+*/
+
+export async function GET(request: NextRequest) {
```
doug-s-nava, line 19: making API request here in order to avoid exposing the API key

frontend/src/services/fetch/fetcherHelpers.ts
```diff
@@ -38,32 +37,6 @@ export function getDefaultHeaders(): HeadersDict {
   return headers;
 }
 
-/**
- * Send a request and handle the response
- * @param queryParamData: note that this is only used in error handling in order to help restore original page state
- */
-export async function sendRequest<ResponseType extends APIResponse>(
```
doug-s-nava, line 45: since this function wasn't doing much, and what it was doing (reading a json response body) was making our implementation less flexible, I got rid of it and moved all this functionality either into the fetcher generator or endpoint specific wrappers

frontend/src/services/fetch/fetcherHelpers.ts
```diff
@@ -127,42 +100,12 @@ function fetchErrorToNetworkError(
     : new NetworkError(error);
 }
 
-// note that this will pass along filter inputs in order to maintain the state
-// of the page when relaying an error, but anything passed in the body of the request,
-// such as keyword search query will not be included
-function handleNotOkResponse(
```
doug-s-nava, line 133: similar to the above, this wasn't doing much, moved the behavior into the fetcher generator

frontend/src/services/fetch/fetchers/fetchers.ts
```diff
@@ -52,27 +50,51 @@ export function requesterForEndpoint<ResponseType extends APIResponse>({
...
+        response.status,
+      );
+    }
 
     return response;
   };
 }
 
 export const fetchOpportunity = cache(
-  requesterForEndpoint<OpportunityApiResponse>(fetchOpportunityEndpoint),
```
doug-s-nava, line 70: these all return `Response` now, so the generic isn't necessary any more

frontend/tests/e2e/search/search.spec.ts
```diff
@@ -156,12 +156,12 @@ test.describe("Search page tests", () => {
     page,
   }: PageProps) => {
     await page.goto("/search");
+    await waitForSearchResultsInitialLoad(page);
```
doug-s-nava, line 159: not sure why, but these tests got flaky. Adding the wait on initial page load seems to solve the problem

frontend/src/components/opportunity/OpportunityDocuments.tsx
```diff
@@ -16,9 +14,6 @@ interface OpportunityDocumentsProps {
   documents: OpportunityDocument[];
 }
 
-dayjs.extend(advancedFormat);
-dayjs.extend(timezone);
```
doug-s-nava, line 20: moved this into a central location

frontend/src/utils/dateUtil.ts
```diff
@@ -23,3 +30,5 @@ export function formatDate(dateStr: string | null): string {
   };
   return date.toLocaleDateString("en-US", options);
 }
+
+export const getConfiguredDayJs = () => dayjs;
```
doug-s-nava, line 34: to support the timestamp in the filename

frontend/src/components/USWDSIcon.tsx
```diff
@@ -11,6 +11,8 @@ interface IconProps {
 // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
 const sprite_uri = SpriteSVG.src as string;
 
+// height prop doesn't seem to work
```
acouch, line 14: Seems like this could be discussion rather than a code comment. The USWDS component doesn't work with SVGs as far as I recall. We could revisit that, was a while ago this was adopted.

frontend/src/app/api/search/export/route.ts
acouch: Seems like a shame we have to do this. Maybe at some point we can open up some of the APIs for the general public. 

Noting the API is a `POST` while we're using a `GET`, I think `GET` is the correct interpretation for search since a search request isn't meaningfully creating or updating server entities.

acouch (approved):
Looks great.

frontend/src/components/USWDSIcon.tsx
doug-s-nava: true, I'll remove the comment. Ticket for follow up here: https://github.com/HHS/simpler-grants-gov/issues/3748

frontend/src/app/api/search/export/route.ts
doug-s-nava: Agree it's not great, but the best option we have for now. We should maybe talk as a team about how we want to handle this sort of thing.

As far as the GET v POST, I'd be fine to make the first request as POST as well, but since it's largely internal and not sending a request body, a GET seemed fine. The search itself being a post makes sense to me since we're sending data in the body of the request.

acouch (approved):

PR #3701, opened 2025-01-30
[Issue #3577] Modify sorting to support sorting by multiple fields

### LINKED ISSUE #3577: Modify sorting in search endpoint to support passing in multiple order_by values to support secondary+ sorting
### Summary

Needs more detail

Instead of taking in a str with allowed values, we should take in a list of str with allowed values.

Something like: `["relevancy", "post_date"]` would sort first by relevancy, and break ties based on post date

TODO - it might actually make more sense to have the ascending/descending value also be a part of this list so you can sort things opposite (ascending relevancy, descending post date).

Regardless, we should just make this value a list so if someone wants to sort by 5 values they can (probably add a max value for performance/realistic reasons).

### Acceptance criteria

- [ ] TODO

chouinar:
## Summary
Fixes #3577

## Changes proposed
Allow for sorting across multiple fields in all of our pagination/sorting endpoints

Deprecate sorting/ordering by a single field (not using the new approach)

## Context for reviewers
There have been a few features we've wanted to support that require sorting across multiple fields. The search index supports this easily, but the API doesn't have a good way of doing this. Rather than some set of specific sort rules on the API side, this instead just lets the user give a multisort. If you want the same behavior as before, send us just a single record.

I did add a hacky thing that adapts the old way into the new way for the frontend, search still works from the UI with this change, although relevancy sort loses the `post_date` secondary sort behavior we had previously built in.

I found while making this change that some of the endpoints that have pagination, don't actually do any sorting and made follow-up tickets to fix that (ie. they weren't broken by this).


babebe (approved):
LGTM !!

mdragon:
This seems to be working fine with my FE changes, and was working fine before I made my FE changes to the fallback for teh current structure seems to work as well. I think we can move this forward.

chouinar:
@mdragon - Thoughts on waiting until tomorrow (after prod deploy) to merge this? That way we can get both this change and yours out there together without immediately being in prod.

mdragon:
> @mdragon - Thoughts on waiting until tomorrow (after prod deploy) to merge this? That way we can get both this change and yours out there together without immediately being in prod.

That's probably the best plan. Not sure if @btabaska would want to alert our API Key holders this change is coming as well.

babebe (approved):
Tested locally. LGTM

PR #3731, opened 2025-01-30
[Issue #3730] Fix parsing `issueType` bug

### LINKED ISSUE #3730: Bug parsing `issueType` from GitHub export
### Summary

With some of the changes made to the workflow for validating data from the GitHub API, we introduced a bug with how `issueType` is parsed that's causing reporting problems downstream.

### Reproducibility

### How to reproduce the bug

If you checkout the latest version of `main` and run `make gh-data-export` then check the output in `data/delivery-data.json` the value of `"issueType"` is `null` for all records. This is due to an issue with the alias in the pydantic model for `IssueType`.

In reality only about 30% of issues across both projects are missing a type.

### Browser

Not applicable

### Code of Conduct

widal001:
## Summary

Fixes the type in the `IssueType` pydantic model that was causing `issueType` to be parsed incorrectly.

Fixes #3730 

- Fixes typo in `IssueType` field alias (it was incorrectly aliased as `"type"` instead of `"issueType"`)
- Replaces the string literal field aliases with module level constants to make it easier to update and monitor field aliases.

1. Checkout the PR locally
2. Run `make gh-data-export` to export the data from GitHub
3. Run the following command to check the percentage of issues that have `null` value for issue type:
    ```bash
    jq '[.[] | select(.issue_type == null)] | length' data/delivery-data.json
    ```
4. The output should be around 691
5. Spot check values in `data/delivery-data.json` where `"issue_type"` is `null` each issue that you check, should be missing a "Type" in GitHub (see screenshot below)
Screenshot 2025-01-30 at 5 02 46 PM

A major goal for Sprint 2.3 will be to set up some automated data quality checks in both Metabase and the ETL pipeline to prevent or at least catch similar regressions earlier.

### DIFF
--- a/analytics/tests/integrations/github/test_validation.py
+++ b/analytics/tests/integrations/github/test_validation.py
@@ -24,7 +24,7 @@ VALID_ISSUE_CONTENT = {
         "title": "Test Parent",
         "url": "https://github.com/test/repo/issues/2",
     },
-    "type": {
+    "issueType": {
         "name": "Bug",
     },
 }
--- a/analytics/src/analytics/integrations/github/validation.py
+++ b/analytics/src/analytics/integrations/github/validation.py
@@ -5,6 +5,12 @@ from datetime import datetime, timedelta
 
 from pydantic import BaseModel, Field, computed_field, model_validator
 
+# Declares class level constants for the fields that need to be aliased
+ISSUE_TYPE = "issueType"
+CLOSED_AT = "closedAt"
+CREATED_AT = "createdAt"
+PARENT = "parent"
+
 
 def safe_default_factory(data: dict, keys_to_replace: list[str]) -> dict:
     """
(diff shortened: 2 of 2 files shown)


coilysiren (approved):
👍🏽!

DavidDudas-Intuitial (approved):
Thanks for finding and fixing the root cause of recent data anomalies!

jcrichlake:
Should we put a unit test in place for these values? To make sure that the values are parsed correctly and not dropped due to an incorrect label?

analytics/tests/integrations/github/test_validation.py
```diff
@@ -24,7 +24,7 @@
         "title": "Test Parent",
         "url": "https://github.com/test/repo/issues/2",
     },
-    "type": {
+    "issueType": {
```
widal001, line 27: @jcrichlake This is the unit test, associated with parsing the values. Part of the problem was that the key was also incorrect here.

Open to other suggestions if you think there'd be a better way to catch or test this!

widal001: Ideally we'd implement something similar to snapshot testing, to make sure that future refactors of the ETL don't change the output. But we'll have to think of the best way to do that when the actual input (i.e. from GitHub) changes so frequently.

jcrichlake: Ah my bad, I didn't read the full PR, I was commenting on the changes I watched you make yesterday.  I do wonder if we couldn't update the GraphQL to export the data to have the column names as expected in the database.  I know SQL has the ability to alias columns and I would be surprised of GraphQL didn't have something similar.  

It wouldn't have to be in this PR specifically but something we could maybe write a card for to look into?

widal001: Ooh that's a great idea! Yeah aliasing columns in GraphQL is quite easy to do. Should have been my first thought actually.

I think there are some nuances, so maybe still best saved for a follow-up ticket but agree that would make it more robust than aliasing in the validation step.

PR #3738, opened 2025-01-31
[Issue #3464] Better agency display/sort in search

mdragon:
## Summary
Fixes #3464 

🚨🚨🚨 Diffing against Michael's API change branch for now to highlight the further changes I made, but we will probably merge direct to main not into his branch to go to main.

## Changes proposed
Added support for the API change that takes an array of sort criteria instead of individual arguments for filed and direction

## Additional information
With some logging in the API layer I was able to demonstrate we were getting the correct format passed in from the FE and results sorted as expected:
Default/relevancy sort:
``` sort_order: [SortOrderParams(order_by='relevancy', sort_direction=<SortDirection.DESCENDING: 'descending'>), SortOrderParams(order_by='post_date', sort_direction=<SortDirection.DESCENDING: 'descending'>)]```
Agency sort:
``` sort_order: [SortOrderParams(order_by='top_level_agency_name', sort_direction=<SortDirection.DESCENDING: 'descending'>), SortOrderParams(order_by='agency_name', sort_direction=<SortDirection.DESCENDING: 'descending'>)] ```

### DIFF
--- a/frontend/src/services/fetch/fetchers/searchFetcher.ts
+++ b/frontend/src/services/fetch/fetchers/searchFetcher.ts
@@ -4,6 +4,7 @@ import { fetchOpportunitySearch } from "src/services/fetch/fetchers/fetchers";
 import {
   PaginationOrderBy,
   PaginationRequestBody,
+  PaginationSortOrder,
   QueryParamData,
   SearchFetcherActionType,
   SearchFilterRequestBody,
@@ -12,12 +13,12 @@ import {
 import { SearchAPIResponse } from "src/types/search/searchResponseTypes";
 
 const orderByFieldLookup = {
-  relevancy: "relevancy",
-  opportunityNumber: "opportunity_number",
-  opportunityTitle: "opportunity_title",
-  agency: "agency_code",
-  postedDate: "post_date",
-  closeDate: "close_date",
+  relevancy: ["relevancy"],
+  opportunityNumber: ["opportunity_number"],
+  opportunityTitle: ["opportunity_title"],
+  agency: ["top_level_agency_name", "agency_name"],
+  postedDate: ["post_date"],
+  closeDate: ["close_date"],
 };
 
 type FrontendFilterNames =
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: frontend/src/services/fetch/fetchers/searchFetcher.ts [hunk 1]
```diff
@@ -4,19 +4,20 @@ import { fetchOpportunitySearch } from "src/services/fetch/fetchers/fetchers";
...
 } from "src/types/search/searchRequestTypes";
 
 const orderByFieldLookup = {
-  relevancy: "relevancy",
-  opportunityNumber: "opportunity_number",
-  opportunityTitle: "opportunity_title",
-  agency: "agency_code",
-  postedDate: "post_date",
-  closeDate: "close_date",
+  relevancy: ["relevancy"],
```
chouinar, line 15: Shouldn't this also have post_date? I see you had a default pagination sort order below, but if someone intentionally chose relevancy (not sure how that links / if that's possible) you'd lose the post_date bit, right?
```suggestion
  relevancy: ["relevancy", "post_date"],
```

frontend/src/services/fetch/fetchers/searchFetcher.ts
```diff
@@ -4,19 +4,20 @@ import { fetchOpportunitySearch } from "src/services/fetch/fetchers/fetchers";
...
+  opportunityNumber: ["opportunity_number"],
+  opportunityTitle: ["opportunity_title"],
+  agency: ["top_level_agency_name", "agency_name"],
```
chouinar, line 18: To make this fully work, we'd also need the backend to always populate top level agency name, as long as you're making the API changes, could add the fallback to the property on the ORM model

frontend/src/services/fetch/fetchers/searchFetcher.ts
mdragon: Yes, I'm torn on how to best handle the default and the post_date backup. Part of me thinks we should always have a post_date fallback, so even if sorting by agency, we show the ones closing sooner earlier amongst the peers from that same agency. But I think In the implementation you'll see that if it's hitting on relevancy it's adding the post_date backup in the pagination builder. But agree, I'm not sure if that's the right way, except to note, anything we default here, doesn't affect the "no sort" route which is possible and does happen (primarily if you've just never touched sort upon landing on the page).

doug-s-nava: [tldr: I'm good with this change as is] it seems like based on the changes to the fetcher, the default is being set the way we want with the secondary sort. Beyond that, I feel like if users want to update the sort away from the default I think it's best to not implement a hidden secondary sort behind the scenes at the moment.

doug-s-nava (commented):
haven't tested manually yet, but a couple thoughts

frontend/src/services/fetch/fetchers/searchFetcher.ts
```diff
@@ -98,24 +99,38 @@ export const buildPagination = (
...
+
   if (sortby) {
+    sort_order = [];
+    // this will need to change in a future where we allow the user to set an ordered set of sort columns.
+    // for now we're just using the multiple internally behind a single column picker drop down so this is fine.
     for (const [key, value] of Object.entries(orderByFieldLookup)) {
       if (sortby.startsWith(key)) {
-        order_by = value as PaginationOrderBy;
-        break; // Stop searching after the first match is found
+        const sort_direction =
```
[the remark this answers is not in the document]
mdragon: I was originally going to go bigger and try to support multiple arguments via the URL argument so we'd need to be parsing each argument for it's direction. I stopped short of that since we have no plans for how the UI for that would work, and it would have required a more fundamental overhaul of what this was already doing.

frontend/src/services/fetch/fetchers/searchFetcher.ts
chouinar: Don't modify that line, you'll hit a null pointer error if you do. That full check needs to be there to check if the agency record exists first, and then if the top level agency exists off of that.

Add a new if statement below it that is:
```py
if self.agency_record is not None:
     return self.agency_record.agency_name
```

mdragon: Right, thanks. I realized I was thinking about the Not None backwards after I asked you here.

chouinar (dismissed):
LGTM!

coilysiren (approved):
👍🏼

coilysiren:
(approved based purely on @chouinar's approval 10 minutes ago)

Issue #3691, opened 2025-01-30
Create API endpoint for PATCH /users/:userID/save-searches/:savedSearchID


chouinar:
### Summary

This will be an update endpoint for a saved search.

This endpoint will _only_ allow a user to update the name of the saved search for now (so, a more minimal version of the endpoint that creates the saved search). We do not need to handle anything with updating the saved search / 

Auth should work identically to the other saved search endpoints.

The endpoints should 404 if the saved search doesn't exist.

### Acceptance criteria

- [ ] Endpoint added
- [ ] Unit tests added

PR #3744, opened 2025-02-03
[Issue #3691] Update user saved search

babebe:
## Summary
Fixes #{[3691](https://github.com/HHS/simpler-grants-gov/issues/3691)}

## Changes proposed
New route PUT /users/:userID/save-searches/:saved_search_id added
Input and Output schema for route added
Pydantic Model for input update fields
Tests added


api/src/services/users/update_saved_searches.py
```diff
@@ -0,0 +1,26 @@
...
+            UserSavedSearch.saved_search_id == saved_search_id, UserSavedSearch.user_id == user_id
+        )
+        .first()
+    )
+
+    if not saved_search:
+        raise_flask_error(404, "Saved search not found")
+
+    # Update
+    saved_search.name = json_data["name"]
```
chouinar, line 24: It's minor right now, but can we have a small Pydantic model so we're not indexing dicts? Just for type safety.

api/src/services/users/update_saved_searches.py
```diff
@@ -0,0 +1,26 @@
...
+    db_session: db.Session, user_id: UUID, saved_search_id: UUID, json_data: dict
+) -> UserSavedSearch:
+    """Update saved search for a user"""
+    saved_search = (
+        db_session.query(UserSavedSearch)
+        .filter(
+            UserSavedSearch.saved_search_id == saved_search_id, UserSavedSearch.user_id == user_id
+        )
+        .first()
+    )
```
chouinar, line 18: Use `db_session.execute(select(...` - query is the old version and will be deprecated in the future. I know we use it in unit tests (we shouldn't), but want to keep that out of any non-test code.

api/tests/src/api/users/test_user_update_saved_search.py
```diff
@@ -0,0 +1,74 @@
...
+from tests.src.db.models.factories import UserSavedSearchFactory
+
+
+@pytest.fixture
+def saved_search(enable_factory_create, user, db_session):
+    search = UserSavedSearchFactory.create(
+        user=user, name="Save Search", search_query={"keywords": "python"}
+    )
+    db_session.commit()
+    return search
```
chouinar, line 15: The [DB session commits automatically for each factory](https://github.com/HHS/simpler-grants-gov/blob/main/api/tests/src/db/models/factories.py#L317), don't need to commit.

```suggestion
@pytest.fixture
def saved_search(enable_factory_create, user, db_session):
    return UserSavedSearchFactory.create(
        user=user, name="Save Search", search_query={"keywords": "python"}
    )
```

api/src/api/users/user_routes.py
```diff
@@ -318,3 +321,34 @@ def user_get_saved_searches(db_session: db.Session, user_id: UUID) -> response.A
     saved_searches = get_saved_searches(db_session, user_id)
 
     return response.ApiResponse(message="Success", data=saved_searches)
+
+
+@user_blueprint.patch("/<uuid:user_id>/saved-searches/<uuid:saved_search_id>")
```
chouinar, line 326: I know I said to make this a patch, but let's actually make it a `PUT` - it'll simplify a lot more down the road if this endpoint is just an overwrite endpoint.

api/src/services/users/update_saved_searches.py
babebe: Done

api/src/services/users/update_saved_searches.py
babebe: Done

api/tests/src/api/users/test_user_update_saved_search.py
babebe: Done

api/src/api/users/user_routes.py
babebe: Done

PR #3749, opened 2025-02-03
[Issue #3584] Remove references to the attachment type that we're no longer using

### LINKED ISSUE #3584: Modify frontend so it doesn't look for attachment type in the API response
### Summary

We aren't going to try and guess which attachment is "the NOFO", so we're no longer using the concept of an attachment type on the backend. This blocks #3264 and #3265.

### Acceptance criteria

- [ ] Frontend updated to remove logic related to attachment type
- [ ] #3264 and #3265 updated

mdragon:
## Summary
Fixes #3584

## Changes proposed
Remove references to the attachment_type which is depreciated on the backend. 

## Context for reviewers
Testing is limited locally, so I was able to verify that some of this works, but real test will be in Dev:
✅ One attachment: http://localhost:3000/opportunity/155
image
image

✅ Two attachments: http://localhost:3000/opportunity/152 
image

### DIFF
--- a/frontend/src/app/[locale]/opportunity/[id]/page.tsx
+++ b/frontend/src/app/[locale]/opportunity/[id]/page.tsx
@@ -124,11 +124,9 @@ async function OpportunityListing({ params }: OpportunityListingProps) {
   });
 
   const nofoPath =
-    opportunityData.attachments.filter(
-      (document) =>
-        document.opportunity_attachment_type ===
-        "notice_of_funding_opportunity",
-    )[0]?.download_path || "";
+    opportunityData.attachments.length === 1
+      ? opportunityData.attachments[0]?.download_path
+      : "";
 
   return (
     <div>
--- a/frontend/src/components/opportunity/OpportunityDocuments.tsx
+++ b/frontend/src/components/opportunity/OpportunityDocuments.tsx
@@ -6,7 +6,6 @@ import { useTranslations } from "next-intl";
 import { Link, Table } from "@trussworks/react-uswds";
 
 interface OpportunityDocument {
-  opportunity_attachment_type: string;
   file_name: string;
   download_path: string;
   updated_at: string;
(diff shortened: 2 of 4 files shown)


frontend/src/app/[locale]/opportunity/[id]/page.tsx
```diff
@@ -124,11 +124,9 @@ async function OpportunityListing({ params }: OpportunityListingProps) {
   });
 
   const nofoPath =
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/opportunity/[id]/page.tsx:126]: since we aren't doing anything to try to see if the attachment is a nofo anymore, I think we should remove `nofo` from any variable names, including the props on OpportunityDescription and OpportunityDownload.  Maybe `firstAttachmentPath`?

doug-s-nava (commented):
Just to be clear on the requirements here since I don't see it spelled out anywhere:

We are removing references to attachment type, and instead only referring to a file as a NOFO if its opportunity has only one attachment. In that context, we are still going ahead with the idea of identifying the NOFO one way or another on the opportunity detail page, even though for the moment it's an inexact science, so we're ok with leaving all of the NOFO terminology (variable naming and user facing verbiage) in place, just changing the logic behind identifying the NOFO.

Does that sound right? I just want to make sure we shouldn't also be removing references to "NOFO" here as well.

Also, a question for another time, but revisiting this code after implementing the csv download function, I notice there's no error handling on the NOFO download. Do we care? If not we could simplify the csv download code significantly. If so, we should make a ticket to refactor this a bit.

mdragon:
> Does that sound right? I just want to make sure we shouldn't also be removing references to "NOFO" here as well.
This is my understanding that while we don't think we need to optimize for identifying NOFOs within larger packets of attachments, if there's only one it's still fine to call it the NOFO and make it easier to find that file. However, I think you raise a good point that the download button was intended to be tied to the NOFO determination, for now we'll only use that button/language if there is only one attachment so I think we're still ok there.
 
> Also, a question for another time, but revisiting this code after implementing the csv download function, I notice there's no error handling on the NOFO download. Do we care? If not we could simplify the csv download code significantly. If so, we should make a ticket to refactor this a bit.
Ultimately that download will just be having the browser go download a file directly from the CDN. So I think there's not much we would be able to do much in terms of error handling? Where as the CSV download has an API hit that could fail after the user interacts with it? But I didn't look at the download piece so I may have missed something.

doug-s-nava (approved):
✅ ✅ ✅ ✅

PR #3756, opened 2025-02-03
[unticketed] fix webkit / chrome error with csv download filename

doug-s-nava:
## Summary
Unticketed

## Changes proposed
https://github.com/HHS/simpler-grants-gov/pull/3689 introduced search results csv downloads but on all browsers other than Firefox, the files were not downloading with the proper file names. This fix is a bit of a hack, but one that was used by FileSaver.js, which people probably still use! https://github.com/eligrey/FileSaver.js/blob/master/src/FileSaver.js

## Context for reviewers
The new function here is not covered by unit tests, but I did improve the e2e test around the functionality.
### Test steps
1. `npm run dev` in this branch
3. click "export results"
4. _VERIFY_: file downloads with filename matching `grants-search-<timestamp>.csv`
5. repeat steps 2 - 4 for every browser you have available

### DIFF
--- a/frontend/src/utils/generalUtils.ts
+++ b/frontend/src/utils/generalUtils.ts
@@ -96,3 +96,18 @@ export const encodeText = (valueToEncode: string) =>
 export const stringToBoolean = (
   mightRepresentABoolean: string | undefined,
 ): boolean => mightRepresentABoolean === "true";
+
+// a hack to get filenames to work on blob based downloads across all browsers
+// see https://stackoverflow.com/a/48968694
+export const saveBlobToFile = (blob: Blob, filename: string) => {
+  const temporaryLink = document.createElement("a");
+  document.body.appendChild(temporaryLink);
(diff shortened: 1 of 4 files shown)


frontend/src/utils/generalUtils.ts
```diff
@@ -96,3 +96,18 @@ export const encodeText = (valueToEncode: string) =>
 export const stringToBoolean = (
   mightRepresentABoolean: string | undefined,
 ): boolean => mightRepresentABoolean === "true";
+
+// a hack to get filenames to work on blob based downloads across all browsers
```
mdragon, line 100: 👀

mdragon (approved):
Looks good.... terrible, but good.

frontend/src/utils/generalUtils.ts
doug-s-nava: 😬

PR #3774, opened 2025-02-04
[Issue #3538] Modify the backend notification script to handle saved searches

### LINKED ISSUE #3538: Modify the backend notification script to handle saved searches
### Summary

Logic should be roughly:
* Fetch all saved searches since the last job ran (TODO - how to know when the last job ran?) based on the last_notified_at timestamp
* Merge all saved searches with the same query together (maintain which records in the DB these connect to)
* Query the search index with all the searches, only fetching opportunity IDs
* For any searches that produce a different set of results,  update the user notification mapping + user_saved_search table.

TODO - fill out in a bit more detail

The job does not need to actually notify, it can just write to the table created in #3546 

### Acceptance criteria

- [ ] Backend script updated

mikehgrantsgov:
## Summary
Fixes #3538

## Changes proposed
Update `generate_notifications` task to use saved searches. Run those saved searches against the search index to compare opportunity ID changes.

## Context for reviewers
The job does not actually notify. That will be done in a follow-up ticket.

## Additional information
See attached unit tests


api/tests/src/task/notifications/test_generate_notifications.py
```diff
@@ -130,3 +144,237 @@ def test_no_notification_log_when_no_updates(
...
+    factories.OpportunitySummaryFactory.build(
+        opportunity=new_opportunity,
+        summary_description="This should appear in test search results",
+    )
+    json_record = schema.dump(new_opportunity)
+    search_client.bulk_upsert(opportunity_index, [json_record], "opportunity_id")
+
+    # Run the notification task
+    result = cli_runner.invoke(args=["task", "generate-notifications"])
+    assert result.exit_code == 0
```
chouinar, line 351: We can run these with the task itself, that would give you more direct access to checking the map if we wanted to verify anything in that.

api/src/task/notifications/generate_notifications.py
```diff
@@ -93,16 +98,54 @@ def _collect_opportunity_notifications(self) -> None:
...
+                if set(current_results) != previous_results:
+                    user_id = saved_search.user_id
+                    if user_id not in self.user_notification_map:
+                        self.user_notification_map[user_id] = NotificationContainer()
+                    self.user_notification_map[user_id].saved_searches.append(saved_search)
+
+                    # Update the saved search with new results
+                    saved_search.searched_opportunity_ids = list(current_results)
+
+        logger.info(f"User notification map: {self.user_notification_map}")
```
chouinar, line 133: Do we want to log this? This could be a lot of data

mikehgrantsgov: Good call. I was using that for some initial debugging, and removed it.

api/src/task/notifications/generate_notifications.py
```diff
@@ -93,16 +98,54 @@ def _collect_opportunity_notifications(self) -> None:
...
+        )
+        saved_searches = self.db_session.execute(stmt).scalars()
+
+        # Group searches by query to minimize search index calls
+        query_map: dict[str, list[UserSavedSearch]] = {}
+        for saved_search in saved_searches:
+            query_key = str(saved_search.search_query)
+            if query_key not in query_map:
+                query_map[query_key] = []
+            query_map[query_key].append(saved_search)
```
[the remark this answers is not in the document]
mikehgrantsgov: I added a util to cover this. I think this would cover the scenario where we remove the pagination.

api/tests/src/task/notifications/test_generate_notifications.py
mikehgrantsgov: Good idea, I adapted this over.

api/tests/src/task/notifications/test_generate_notifications.py
```diff
@@ -130,3 +145,257 @@ def test_no_notification_log_when_no_updates(
...
+    # Verify both searches were updated with the same new results
+    db_session.refresh(saved_search1)
+    db_session.refresh(saved_search2)
+
+    assert saved_search1.searched_opportunity_ids == saved_search2.searched_opportunity_ids
+    assert saved_search1.last_notified_at > datetime_util.utcnow() - timedelta(minutes=1)
+    assert saved_search2.last_notified_at > datetime_util.utcnow() - timedelta(minutes=1)
+
+
+def test_search_notifications_on_index_change(
```
mikehgrantsgov, line 318: @chouinar any idea why these tests all pass locally but fail when running in CI/CD? Maybe running the full suite puts the db/search index in a state?

mikehgrantsgov: Nevermind, worked around this.

chouinar (approved):
Whoops - thought I approved this one yesterday - LGTM!

PR #3776, opened 2025-02-05
[Unticketed] Adjust the batch size of the attachment job to be 100

chouinar:
## Changes proposed
Change the opportunity attachment job to process 100 attachments at a time

## Context for reviewers
This means we'll fetch 100 attachments at a time into memory - instead of 1000+ like we were previously doing

Doing this because I hit an out-of-memory error on the ECS task when running - not 100% certain this will work (depends on how the ORM cache works?) but want to test it.


mdragon (approved):
Looks good.

PR #3777, opened 2025-02-05
DO NOT MERGE - Prototype of temporal versioning via SQLAlchemy ORM

chouinar:


chouinar:
Not going with this approach, no need to leave this draft up

PR #3787, opened 2025-02-06
[Issue #2537] Create e2e tests for subscribe page

### LINKED ISSUE #2537: Add e2e tests for /subscribe
### Summary

The subscribe page has unit but not e2e tests. To accomplish this we could either:

* use  [Experimental test mode for Playwright](https://github.com/vercel/next.js/blob/canary/packages/next/src/experimental/testmode/playwright/README.md), see also a [tutorial]https://dev.to/weamadel/configure-playwright-with-nextjs-mock-apis-for-testing-44oh]
* or use something like [wiremock](https://github.com/wiremock/wiremock)

Sendy is the only server action / API we use at the moment so would heavily favor the former since it could be the only remote API we use for e2e for a while if ever and while wiremock is pretty simple it is still another service to adopt and maintain.

### Acceptance criteria

- [ ] mocking solution for server actions adopted
- [ ] e2e test for /subscribe form

KeithNava:
## Summary
Fixes #2537

## Changes proposed
Creates the E2E tests for the /subscribe page. 

## Context for reviewers
Because of some challenges with mocking API calls within NextJS' server components in the current version (v15 as of this PR) we had to utilize some experimental features - https://github.com/vercel/next.js/tree/canary/packages/next/src/experimental/testmode/playwright

> [!WARNING]
>   There is some risk in depending on experimental features but this does appear to be the recommended path today
> 1. Subject to Breaking Changes
> - Experimental features can change without prior notice
> - May introduce unexpected behavior in future releases
> - Not recommended for production-critical applications
> 2. Performance and Stability Risks
> - Not fully tested across all use cases
> - Potential performance overhead
> - May require additional debugging
> 3. Upgrade Considerations
> - Always review release notes when updating Next.js
> - Experimental configs might require manual migration
> - Potential need to refactor code after updates

## Additional information
subscribe spec ts-passing


frontend/next.config.js
```diff
@@ -101,6 +101,9 @@ const nextConfig = {
       "types",
     ],
   },
+  experimental: {
+    testProxy: true,
```
doug-s-nava, line 105: 🆒

frontend/tests/e2e/subscribe.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+test("error during signup", async ({ next, page }) => {
+  mockAPIEndpoints(next, "Error with subscribing");
+
+  await page.getByLabel("First Name (required)").fill("Apple");
+  await page.getByLabel("Email (required)").fill("name@example.com");
+
+  await page.getByRole("button", { name: /subscribe/i }).click();
+
+  await expect(page.getByTestId("errorMessage")).toHaveCount(1);
+  await expect(
```
doug-s-nava, line 66: this test seems to be failing, and I think it's because the API request isn't being proxied for some reason. Logging out the requests that come through the proxy locally it seems to be only catching a google analytics request at the moment. My guess is that's because the request for subscription is in a server action, so it isn't being made from the client directly?

acouch: Maybe try https://storybook.js.org/addons/msw-storybook-addon

KeithNava: It's interesting that this test is failing now in CI, but on my local, it seems to be passing and showing that it's mocking the server-side request. I will take a deeper look and will definitely check out MSW. Thanks @doug-s-nava @acouch !
subscribe-spec-e2e

doug-s-nava: the e2e test action does not have the sendy variables set, so that could definitely be what's going on here. Since we're mocking the request, hopefully the sendy url won't matter, and we can set it to be something random and harmless rather than using the real sendy url.

if you want to test that out, this is where you'd want to look in the code https://github.com/HHS/simpler-grants-gov/blob/003c64c456600031ae5e363c58d2d50b80958b07/.github/workflows/ci-frontend-e2e.yml#L61

doug-s-nava: Glad you cracked the issue! This change is probably fine but I'd rather avoid any change to the application itself if we can make a change just to the testing configuration. Can you try adding the environment variable to the action definition instead of defaulting within the app itself?

KeithNava: Ok I think I got it stable now without impacting the application logic 👉  https://github.com/HHS/simpler-grants-gov/pull/3787/files#diff-def35e8365b9471bd377c2100868dd944a82954372a3a4efc7e29063be9aa844R63

doug-s-nava (approved):
✅ ✅ ✅ ✅

coilysiren (commented):
defer to others 🙏🏼

PR #3794, opened 2025-02-06
[Issue #3539] Add notifications to notifications backend

### LINKED ISSUE #3539: Add notifications to the notification backend task
### Summary

We won't worry about the shape of the email yet, just dump the info we've collected up to this point into an email and send it.

Mock utility created in https://github.com/HHS/simpler-grants-gov/issues/3540

### Acceptance criteria

- [ ] When run locally (non tests) we should make the notifications just log, there isn't a way we're ever going to mock out an email service.

mikehgrantsgov:
## Summary
Fixes #3539

## Changes proposed
Integrate AWS Pinpoint client to `generate_notifications.py` task
Set pinpoint app ID as an environment variable

## Context for reviewers
Open question: how do we handle pinpoint app id? Is it going to set as an env var?

## Additional information
See attached unit tests


api/src/task/notifications/generate_notifications.py
```diff
@@ -150,9 +173,30 @@ def _send_notifications(self) -> None:
             if not container.saved_opportunities and not container.saved_searches:
                 continue
 
-            # TODO: Implement actual notification sending in future ticket
+            # Get user email from LinkExternalUser using select pattern
+            stmt = select(LinkExternalUser).where(LinkExternalUser.user_id == user_id)
+            user = self.db_session.execute(stmt).scalar_one_or_none()
+
+            if not user or not user.email:
```
chouinar, line 180: Just because we'll likely need the email for other things in the future, and it might get pulled from a few places, what if we add a property to the user table, something like:

```py
@property
def email(self) -> str | None:
     if self.login_gov_user is not None: # TODO - would need to add a relationship that fetches this
          return self.login_gov_user.email

     return None
```

api/src/task/notifications/generate_notifications.py
```diff
@@ -150,9 +173,30 @@ def _send_notifications(self) -> None:
...
+                f"You have updates to {len(container.saved_opportunities)} saved opportunities"
+            )
+
+            send_pinpoint_email_raw(
+                to_address=user.email,
+                subject=subject,
+                message=message,
+                pinpoint_client=self.pinpoint_client,
+                app_id=self.config.app_id,
+            )
```
chouinar, line 196: Probably want a try/except around this - if it fails, we should update the `UserNotificationLog` notification_sent field.

api/src/task/notifications/generate_notifications.py
```diff
@@ -150,9 +173,30 @@ def _send_notifications(self) -> None:
...
+
             logger.info(
-                "Would send notification to user",
+                "Sending notification to user",
```
chouinar, line 199: Could we put this before the call to pinpoint?

api/src/db/models/user_models.py
```diff
@@ -46,7 +59,9 @@ class LinkExternalUser(ApiSchemaTable, TimestampMixin):
     )
 
     user_id: Mapped[uuid.UUID] = mapped_column(ForeignKey(User.user_id), index=True)
-    user: Mapped[User] = relationship(User)
+    user: Mapped["User"] = relationship(
+        "User", primaryjoin=lambda: LinkExternalUser.user_id == User.user_id
+    )
```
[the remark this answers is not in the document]
mikehgrantsgov: This change wasn't needed, but a part of the fiddling I was doing to get the sqlalchemy pieces to work. Updated this too

api/src/task/notifications/generate_notifications.py
mikehgrantsgov: Updated this

api/src/db/models/user_models.py
```diff
@@ -30,6 +30,22 @@ class User(ApiSchemaTable, TimestampMixin):
         "UserSavedSearch", back_populates="user", uselist=True, cascade="all, delete-orphan"
     )
 
+    linked_external_user: Mapped["LinkExternalUser | None"] = relationship(
```
chouinar, line 33: Just one last thing, could we call this `linked_login_gov_external_user` or something with login_gov in it?

mikehgrantsgov: Makes sense, just updated

PR #3796, opened 2025-02-06
[Issue #1485] return last page of search if requested page is out of bounds

### LINKED ISSUE #1485: Redirect to the highest valid page of search results when page query param has value beyond the max page possible
### Summary

The user can update the search URL to e.g. page=11 when there's only 7 pages

We currently have a "No results" message there and some weird logic with paging/"next".

From the original ticket: also remove all references to `maxPaginationError` as defined in https://github.com/HHS/simpler-grants-gov/issues/1517 (that ticket was closed)

### Acceptance criteria

- [ ] Frontend redirects to highest valid result page when the query param exceeds that page

doug-s-nava:
## Summary
Fixes #1485

## Changes proposed
If the page of search results requested exceeds the total number of pages returned by the API, this updates the page query param to the last page of results, triggering a new search request.

Includes removal of some leftovers from the previous "maxPageError" implementation, and cleans up a couple other things too.

Worth noting that I tried solving this on the API side first, which you can see in https://github.com/HHS/simpler-grants-gov/pull/3796/commits/423d73020a42f2c690024a1fd6678cbe7a48246e. The issue with that is that there was no easy way to update the url to reflect the change, which seemed bad.

### Test steps

1. `npm run dev` on this branch
2. visit http://localhost:3000/search?page=10000
3. _VERIFY_: you are redirected to the last page of results


frontend/src/app/[locale]/search/error.tsx
```diff
@@ -91,10 +91,6 @@ export default function SearchError({ error, reset }: ErrorProps) {
   const { agency, category, eligibility, fundingInstrument, query, status } =
     convertedSearchParams;
 
-  useEffect(() => {
-    console.error(error);
-  }, [error]);
-
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/search/error.tsx:97]: not sure how this came back, and the above change I think I also handled in a previous PR 🤷

frontend/src/components/search/SearchResultsListFetch.tsx
```diff
@@ -15,20 +16,31 @@ export default async function SearchResultsListFetch({
   searchResultsPromise,
 }: ServerPageProps) {
   const searchResults = await searchResultsPromise;
-
-  const maxPaginationError = null;
   const t = await getTranslations("Search");
 
+  if (
```
doug-s-nava, line 21: where this is being done is more or less arbitrary since the same promise is being awaited in multiple components. That seems weird, which is one of the reasons I did the refactor in https://github.com/HHS/simpler-grants-gov/pull/3796

frontend/src/i18n/messages/en/index.ts
```diff
@@ -573,10 +573,9 @@ export const messages = {
       message: "{count, plural, =1 {1 Opportunity} other {# Opportunities}}",
     },
     resultsListFetch: {
-      title: "Your search did not return any results.",
-      body: "<li>Check any terms you've entered for typos</li><li>Try different keywords</li><li>Make sure you've selected the right statuses</li><li>Try resetting filters or selecting fewer options</li>",
-      paginationError:
-        "You're trying to access opportunity results that are beyond the last page of data.",
+      noResultsTitle: "Your search did not return any results.",
```
doug-s-nava, line 576: just updating the names here as they were confusing me when working with this component

acouch (dismissed):
👍🏿

acouch (approved):
🚀

PR #3799, opened 2025-02-07
[Issue 3203] Refactor analytics ETL workflow for a read-only filesystem

### LINKED ISSUE #3203: Update analytics export to not use file system

DavidDudas-Intuitial:
## Summary
Fixes #3203 

This PR creates a new end-to-end ETL workflow that does *not* write to the filesystem.  The new workflow is derived from the previous file-based implementation but uses memory instead of the filesystem to marshall data between discrete stages of work. This PR also updates the daily scheduled job such that the new workflow will be executed instead of the legacy workflow, thereby eliminating a dependency on a write-access filesystem. 

The new workflow can be triggered on the command-line via `make gh-extract-transform-and-load` or `poetry run analytics etl extract_transform_and_load`.  Functionally it should produce the same results as running `gh-data-export` and `gh-transform-and-load` separately, but during and after the job there will be no files written to `analytics/data/`.

extract-transform-and-load


analytics/src/analytics/cli.py
```diff
@@ -174,10 +174,50 @@ def transform_and_load(
...
+    """Export data from GitHub, transform it, and load into analytics warehouse."""
+    # get configuration
+    config_path = Path(config_file)
+    if not config_path.exists():
+        typer.echo(f"Not a path to a valid config file: {config_path}")
+    config = load_config(config_path, GitHubProjectConfig)
+
+    # validate effective_date
+    try:
+        dateformat = "%Y-%m-%d"
```
jcrichlake, line 200: Should we extract this to it's own private function in case we ever need to support future differing date formats?

analytics/src/analytics/cli.py
```diff
@@ -174,10 +174,50 @@ def transform_and_load(
...
+        datestamp = (
+            datetime.strptime(effective_date, dateformat)
+            .astimezone()
+            .strftime(dateformat)
+        )
+        logger.info("running ETL workflow with effective date %s", datestamp)
+    except ValueError:
+        logger.info("FATAL ERROR: malformed effective date, expected YYYY-MM-DD format")
```
jcrichlake, line 208: Should this be logger.error?

jcrichlake:
Looks overall pretty good! Should we update the old code to remove the previous workflow? Or at least put a comment in the make file for the methods of running the app that are deprecated?

analytics/src/analytics/datasets/utils.py
```diff
@@ -35,9 +31,40 @@ def load_json_data_as_df(
...
-        json_data = json_data[key_for_nested_items]
+
+    return load_json_data_as_df_from_object(
+        json_data,
+        column_map,
+        date_cols,
+    )
+
+
+def load_json_data_as_df_from_object(
```
widal001, line 42: @DavidDudas-Intuitial How does this (in intent) differ from the base dataset [`from_dict()`](https://github.com/HHS/simpler-grants-gov/blob/85c6a464cb1620a422cc8b8041d19ded284e1a6f/analytics/src/analytics/datasets/base.py#L26-L29) method.

In the long-term would you see this method replacing the existing `from_dict()` method or would they both continue to exist?

analytics/src/analytics/datasets/etl_dataset.py
```diff
@@ -83,11 +86,46 @@ def load_from_json_file(cls, file_path: str) -> Self:
         )
 
         # transform entity id columns
+        df = EtlDataset.transform_entity_id_columns(df)
+
+        return cls(df)
+
+    @classmethod
+    def load_from_json_object(cls, json_data: list) -> Self:
```
widal001, line 94: Was there a particular reason we only added this to the `etl_dataset.py` instead of to the base dataset class?

coilysiren (commented):
deferring to other folks existing reviews here 🙏🏼

widal001 (dismissed):
Left a few non-blocking comments above about:
- Adding TODOs to remove deprecated code
- Consolidating potentially duplicative functions (or clarifying their differences) -- This can be a future change
- Adding some logging that was lost in the transition to the new in-memory methods

But overall this refactor seems to minimize the size of the changes while doing its job of allowing us to disable writes. Great work!

### Future changes

While I think it's good to keep the footprint of this PR small, once we add more robust snapshot testing and data quality checks, I'd love to revisit a broader refactor of this step of the ETL that aims to do the following:

- Move some of the use-case specific transformations that currently happen in `integrations/github/main.py` into `etl/github.py` (as described in [this TODO](https://github.com/HHS/simpler-grants-gov/blob/df2fa9920659af259e27bfd8df8d59ebaf98a6cd/analytics/src/analytics/integrations/github/main.py#L1-L6))
- Explore options for combining the write-based ETL and the in-memory ETL into a single pipeline that uses config variables to determine whether we snapshot the files or pass them to the next step in memory
- Deprecate the old entry points and Make commands

### Previous ETL with writes

Screenshot 2025-02-11 at 9 43 22 AM
Screenshot 2025-02-11 at 9 43 59 AM

### New in-memory ETL

Screenshot 2025-02-11 at 9 44 29 AM
Screenshot 2025-02-11 at 9 44 39 AM

analytics/src/analytics/cli.py
DavidDudas-Intuitial: Yes, good catch!

analytics/src/analytics/cli.py
DavidDudas-Intuitial: I see this code occurs twice in the same file. That alone is good enough reason to put it into a separate function.

DavidDudas-Intuitial:
> Should we update the old code to remove the previous workflow? Or at least put a comment in the make file for the methods of running the app that are deprecated?

I do not think those methods are deprecated. It is still useful to be able to run the extract workflow only, e.g. to dump a file to inspect the results in a development environment. It is entirely possible that in the future as we scale we will pivot to run the workflow steps separately and use S3 buckets as interim stores of flat files.

DavidDudas-Intuitial:
> Looks overall pretty good! Should we update the old code to remove the previous workflow? Or at least put a comment in the make file for the methods of running the app that are deprecated?

I do not think those methods are deprecated. It is still useful to be able to run the extract workflow only, e.g. to dump a file to inspect the results in a development environment. It is entirely possible that in the future as we scale we will pivot to run the workflow steps separately and use S3 buckets as interim stores of flat files.

analytics/src/analytics/datasets/utils.py
DavidDudas-Intuitial: Haha, I didn't know (or forgot) there's already a method in utils. I'll use that instead. Thanks for pointing it out!

analytics/src/analytics/cli.py
```diff
@@ -174,10 +174,50 @@ def transform_and_load(
...
 def load_opportunity_data() -> None:
     """Grabs data from s3 bucket and loads it into opportunity tables."""
     extract_copy_opportunity_data()
+
+
+@etl_app.command(name="extract_transform_and_load")
+def extract_transform_and_load(
+    config_file: Annotated[str, CONFIG_FILE_ARG],
+    effective_date: Annotated[str, EFFECTIVE_DATE_ARG],
+) -> None:
```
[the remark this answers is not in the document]
DavidDudas-Intuitial: IMO they should not be deprecated, for at least a couple reasons:

- Today we know we want to run the workflow as one process w/o writing to filesystem, but down the road it is entirely reasonable to assume we pivot back to interim files written to a file system. We're writing only ~2k records per day so it's no problem to keep interim transformations in memory but after a certain point that pattern will not scale or be practical. 
- It is still useful to be able to run the extract workflow only, e.g. to dump a file to inspect the results in a development environment; or to run the load workflow only, e.g. to troubleshoot in a dev environment; etc.

analytics/src/analytics/cli.py
DavidDudas-Intuitial: Thanks for the suggestion!

analytics/src/analytics/datasets/utils.py
DavidDudas-Intuitial: This was existing code in `load_json_data_as_df`, following code that reads from the file system. I abstracted out the `df` related code into its own function, to separate it from the code that reads from the file system, so that the `df` related code can be executed independent of file system access.

analytics/src/analytics/datasets/etl_dataset.py
DavidDudas-Intuitial: Yes. This is a very specific step in a specific use case (see #2 below), not a generic base class behavior.

`EtlDataset` now supports the following similar-but-different code paths:

1. Original: open file created by the export step --> convert the json from the file to a df object --> transform entity id columns --> instantiate class instance
2. New: accept as an arg the json object created by the export step --> convert the json from the file to a df object --> transform entity id columns --> instantiate class instance

analytics/src/analytics/datasets/utils.py
DavidDudas-Intuitial: This code is not similar to the generic `from_dict` method. Instead, it is a common step in multiple `EtalDataset` code paths. See in bold below:

1. Original: open file created by the export step --> **convert the json from the file to a df object** --> transform entity id columns --> instantiate class instance
2. New: accept as an arg the json object created by the export step --> **convert the json from the file to a df object** --> transform entity id columns --> instantiate class instance

widal001 (approved):
LGTM! You're responses to my non-blocking comments make sense, re-ran the pipeline after making the changes and got the same output between the in-memory and the file-based ETL:

Screenshot 2025-02-12 at 2 46 56 PM
Screenshot 2025-02-12 at 2 47 05 PM

PR #3804, opened 2025-02-07
[Issue #1485] Search page suspense refactor

### LINKED ISSUE #1485: Redirect to the highest valid page of search results when page query param has value beyond the max page possible

doug-s-nava:
## Summary
Unticketed refactor work done on top of work for #1485 in https://github.com/HHS/simpler-grants-gov/pull/3796

## Changes proposed
Relying on the search results fetch promise resolving in multiple places in the application creates some confusion and the need for multiple additional components and suspense boundaries. This work is an attempt to simplify how the search page is structured in that regard. With this change:

* 1 promise, 1 resolution
* no need for `fetch` components
* url update after search results response is no longer in an arbitrary place

### Test steps

1. see https://github.com/HHS/simpler-grants-gov/pull/3796
2. also make sure search otherwise still works as normal


acouch (approved):
👍🏿

PR #3814, opened 2025-02-10
[Unticketed] Add batching support for transforming attachments

chouinar:
## Changes proposed
Adds support to batch the opportunity attachment transforms

## Context for reviewers
Been hitting memory issues with the opportunity attachment transformations. Wanted a way to process fewer at a time (fetching/committing in batches). This approach sets up a way to support batching on all of our transformation work, without actually changing the other processes yet.

Effectively now each batch of opportunity attachment transforms will be:
* Fetch 100 attachments
* Transform each of those
* Commit
* Repeat up to 100 times


mdragon (approved):
Looks good

Issue #3692, opened 2025-01-30
Adjust GET saved-search endpoint to take in sorting/pagination and use it


chouinar:
### Summary

Right now the endpoint returns all saved searches sorted by created_at, while this is a reasonable default, we should use our pagination utilities to setup the ability to order by multiple fields (although having a default of `created_at` desc is fine).

### Acceptance criteria

- [ ] API schema updated
- [ ] API implementation updated to use paginator
- [ ] Tests updated

PR #3815, opened 2025-02-10
[Issue #3692] Saved search sorting

babebe:
## Summary
Fixes #{[3692](https://github.com/HHS/simpler-grants-gov/issues/3692)}

## Changes proposed
Updated route to POST endpoint to pass in pagination object
Added Input schema with pagination: can now sort with `created_at, updated_at, name` defaulting to `created_at`
Created util func `apply_sorting` 
Updated service func `get_saved_searches`  to implement the pagination and return pagination info as well
Updated Test


chouinar:
@acouch / @doug-s-nava - not sure who among you is working on the frontend work that would use the saved search list endpoint - we had to update the endpoint from a GET to a POST with a slight path change to handle taking in a request body. If your work hasn't started yet, shouldn't affect much, but apologies if this causes any issue

chouinar (approved):
LGTM!

Issue #3694, opened 2025-01-30
Adjust GET saved-opportunity endpoint to take in sorting/pagination and use it


chouinar:
### Summary

Right now the endpoint returns all saved opportunities without any sorting at all. We should add sorting to the endpoint.

### Acceptance criteria

- [ ] API schema updated
- [ ] API implementation updated to use paginator
- [ ] Tests updated

capriiis:
Waiting on frontend changes, @acouch will attempt to resolve by Feb. 19

PR #3846, opened 2025-02-11
[Issue #3694] Add Sorting To Saved Opportunity

babebe:
## Summary
Fixes #{[3694](https://github.com/HHS/simpler-grants-gov/issues/3694)}

## Changes proposed
Updated route to POST endpoint to pass in pagination object
Added Input schema `UserSavedOpportunitiesRequestSchema` with pagination: can now sort with` "created_at", "updated_at", "opportunity_title", "close_date"` defaulting to `agency_code`
Updated service func `get_saved_opportunities `  to implement the pagination and return also return pagination info 
Updated Test


api/src/services/users/get_saved_opportunities.py
```diff
@@ -1,27 +1,90 @@
...
+        )
+
+        if (
+            order.sort_direction == SortDirection.ASCENDING
+        ):  # defaults to nulls at the end when asc order
+            order_cols.append(asc(column))
+        elif order.sort_direction == SortDirection.DESCENDING:
+            order_col = desc(column)
+            if order.order_by == "close_date":
+                order_cols.append(order_col.nullslast())
```
chouinar, line 48: I think the type checking issues might be because of this confusing it.

I also think we want this to be `nulls_last(order_col)` based on: https://docs.sqlalchemy.org/en/20/core/sqlelement.html#sqlalchemy.sql.expression.nulls_last

babebe: Updated

babebe:
@doug-s-nava Similar to saved search list endpoint endpoint for the user_get_saved_opportunities, this will be updated to use the POST method `/v1/users/<uuid:user_id>/saved-opportunities/list` in order to accommodate the pagination and sorting request object.

doug-s-nava:
@babebe got it. I guess we need tickets to switch over our saved opportunities call?

chouinar:
@babebe - this change looks good, but want to be careful about merging it if it immediately breaks a feature the frontend is building. Can we work with Aaron/Doug to make sure they get changes in that might be needed?

babebe:
> @babebe - this change looks good, but want to be careful about merging it if it immediately breaks a feature the frontend is building. Can we work with Aaron/Doug to make sure they get changes in that might be needed?

[Talked](https://betagrantsgov.slack.com/archives/C05TSL64VUH/p1740000857557909) to Doug. I will be merging the backend, then the front-end changes will be merged in.

mdragon (approved):
Approved

Issue #3697, opened 2025-01-30
Setup GET /agencies to support sorting


chouinar:
### Summary

The API schema contains the pagination parameters, but none of the queries use them and the result we return is completely unsorted.

We should be able to sort by:
* agency_code
* created_at
* agency_name

### Acceptance criteria

- [ ] Logic updated to use sorting

mxk0:
@chouinar is this a dupe of #3576?

chouinar:
@mxk0 - Sort've - this is a more complete version of the ticket as the problem in #3576 was that we never hooked up sorting to begin with + didn't allow that value. This ticket would cover both and more

mxk0:
@chouinar cool, sounds like you and Matt discussed covering this at refinement tomorrow.

PR #3850, opened 2025-02-11
[Issue #3697] Update sorting for Agency 

babebe:
## Summary
Fixes #{[3697](https://github.com/HHS/simpler-grants-gov/issues/3697)}

## Changes proposed
Update pagination to allow sorting with `agency_code, agency_name, created_at` defaulting to `agency_code`
Use util func `apply_sorting`
Add test


api/src/services/agencies_v1/get_agencies.py
```diff
@@ -2,13 +2,14 @@
...
 from pydantic import BaseModel, Field
-from sqlalchemy import asc, select
+from sqlalchemy import select
 from sqlalchemy.orm import joinedload
 
 import src.adapters.db as db
 from src.db.models.agency_models import Agency
 from src.pagination.pagination_models import PaginationInfo, PaginationParams
 from src.pagination.paginator import Paginator
+from src.services.users.get_saved_searches import apply_sorting
```
chouinar, line 12: Could you move `apply_sorting` to a more generic location?

Maybe a file at src/services/service_utils.py?

babebe: Moved

api/tests/src/api/agencies_v1/test_agencies_routes.py
```diff
@@ -77,3 +81,41 @@ def test_agencies_get_with_sub_agencies(
...
+            {"order_by": "agency_code", "sort_direction": "descending"},
+        ]
+
+        response = client.post("/v1/agencies", headers={"X-Auth": api_auth_token}, json=payload)
+        assert response.status_code == 200
+        data = response.json["data"]
+
+        # assert order by agency_name desc then agency_code desc
+        assert data[0]["agency_name"] > data[1]["agency_name"]
+        assert data[1]["agency_code"] > data[2]["agency_code"]
```
chouinar, line 121: I don't think this would be a consistent test as it's relying on whatever randomly gets generated for the agency code value.

For the test you setup, could you make sure each of the agency factory calls sets an agency name and agency code?

babebe: I was thinking irrespective of the agency_code the sort should persist

PR #3857, opened 2025-02-11
[Issue #3822] Modify our PII logging mask to allow long floats

### LINKED ISSUE #3822: Modify our PII logging mask to allow long floats
### Summary

We've encountered an odd bug that causes random errors with our logic to mask PII from logs: https://github.com/HHS/simpler-grants-gov/actions/runs/13246291300/job/36973529031

The error is ultimately due to our process that masks PII information replacing a float/timestamp with `*********` and then the formatting in the logger breaking.

https://github.com/HHS/simpler-grants-gov/blob/main/api/src/logging/pii.py

This masking process happens on every log record, and it checks every part of the log message if it contains a certain regex that looks like an SSN (9 digits in a row, optionally separated by dashes) and replaces it with `*********`. There are already some rules built into this to not replace if we think it's a floating point number, but it only handles the integer half (eg. `999000000.5` is unmodified, but not `5.999000000`). We should fix that

### Acceptance criteria

- [ ] Bug fixed
- [ ] Unit tests updated

mikehgrantsgov:
## Summary
Fixes #3822 

## Changes proposed
Modify our PII logging mask
Add unit test

## Context for reviewers
We've encountered an odd bug that causes random errors with our logic to mask PII from logs: https://github.com/HHS/simpler-grants-gov/actions/runs/13246291300/job/36973529031

The error is ultimately due to our process that masks PII information replacing a float/timestamp with ********* and then the formatting in the logger breaking.

https://github.com/HHS/simpler-grants-gov/blob/main/api/src/logging/pii.py

This masking process happens on every log record, and it checks every part of the log message if it contains a certain regex that looks like an SSN (9 digits in a row, optionally separated by dashes) and replaces it with *********. There are already some rules built into this to not replace if we think it's a floating point number, but it only handles the integer half (eg. 999000000.5 is unmodified, but not 5.999000000). We should fix that

## Additional information
See unit test

### DIFF
--- a/api/src/logging/pii.py
+++ b/api/src/logging/pii.py
@@ -56,12 +56,12 @@ def mask_pii(record: logging.LogRecord) -> bool:
 # Regular expression to match a tax identifier (SSN), 9 digits with optional dashes.
 # Matches between word boundaries (\b), except when:
 #  - Preceded by word character and dash (e.g. "ip-10-11-12-134")
-#  - Followed by a dot and digit, for decimal numbers (e.g. 999000000.5)
-# See https://docs.python.org/3/library/re.html#regular-expression-syntax
+#  - Preceded by or followed by a decimal point (for floating point numbers)
 TIN_RE = re.compile(
     r"""
         \b          # word boundary
         (?<!\w-)    # not preceded by word character and dash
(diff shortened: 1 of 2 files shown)


chouinar (approved):
LGTM!

Issue #3848, opened 2025-02-11
Delete the GET /opportunities/{opp ID}/versions endpoint


chouinar:
### Summary

This endpoint was a half thought out attempt to make versioning work. We're refactoring how versioning is going to work, and this approach won't be kept.

### Acceptance criteria

- [ ] Endpoint deleted
- [ ] Tests removed
- [ ] Any utils for versioning that were setup, cleaned up

PR #3864, opened 2025-02-12
[Issue #3848] Delete get opportunity version endpoint

babebe:
## Summary
Fixes #{[3848](https://github.com/HHS/simpler-grants-gov/issues/3848)}

## Changes proposed
Delete current get opportunity version Endpoint and references to it. New versioning method in the works.


api/tests/src/api/opportunities_v1/test_opportunity_route_get_versions.py
```diff
@@ -1,162 +0,0 @@
...
-    non_forecast_versions = [f["version_number"] for f in non_forecasts]
-
-    assert forecast_versions == expected_forecasts
-    assert non_forecast_versions == expected_non_forecasts
-
-
-def test_get_opportunity_versions_200(client, api_auth_token, enable_factory_create):
-    # Uneventful opportunity with forecasts and non-forecasts, each with some history, nothing filtered
-    opportunity = (
-        OpportunitySummaryHistoryBuilder()
```
chouinar, line 22: While we're deleting things, can you also remove the `OpportunitySummaryHistoryBuilder`? Should only be used here, so is safe to cleanup.

babebe: It's being used in script `seed-local-db.py` do we not need it there ?

chouinar: Right, missed that when looking - we can remove it from there as well (the whole of "history" stuff from db-seed-local, doesn't need to take in the `include-history` param either).

babebe: Removed

chouinar (approved):
LGTM!

PR #3865, opened 2025-02-12
[Issue #1936] fix subscription tests somewhat

### LINKED ISSUE #1936: Fix SearchPagination tests with React 19
### Summary

* Blocked by ~~#1935~~  #2621

Next uses a canary version of React (not the React listed in package.json) and therefore `useFormStatus` (which has been in the canary build), works in source code, but does not in tests. Jest tests use the `react` and `react-dom` version in `package.json` - which does not have `useFormStatus`.

`SearchPagination` therefore had to comment out all tests since they were trying to render a `useFormStatus` - which is not a function in React 18.

Once we upgrade to React 19, these tests should be uncomments and be able to pass.

### Acceptance criteria

- Uncomment tests

doug-s-nava:
## Summary
Fixes #1936 ... kind of

## Changes proposed
The ticket in question refers to fixing tests that no longer exist in the form that they were written about. However, there is some adjustment to be made to other tests that are still referencing the deprecated and no longer used `useFormState` function. This change updates these subscription tests and makes an attempt to expand them where possible.

Note that I ran into issues with fully unit testing the server actions being used in the subscription form. We can return to that another day when perhaps the Next / React server actions implementation is a bit more stable.

### Test steps

1. _VERIFY_: tests pass CI

### DIFF
--- a/frontend/tests/pages/subscribe/SubscriptionForm.test.tsx
+++ b/frontend/tests/pages/subscribe/SubscriptionForm.test.tsx
@@ -1,7 +1,10 @@
 import { render, screen } from "@testing-library/react";
+import { ValidationError } from "src/errors";
 import { mockMessages, useTranslationsMock } from "src/utils/testing/intlMocks";
 
 import SubscriptionForm from "src/components/subscribe/SubscriptionForm";
+
+const mockSubscribeEmail = jest.fn();
 
 jest.mock("next-intl", () => ({
   useTranslations: () => useTranslationsMock(),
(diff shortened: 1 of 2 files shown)


frontend/tests/pages/subscribe/SubscriptionForm.test.tsx
```diff
@@ -42,4 +22,47 @@ describe("SubscriptionForm", () => {
...
+      { errorMessage: "", validationErrors: {} },
+      expect.any(FormData),
+    );
+  });
+
+  // unclear why, but React server actions are still not working with Jest here
+  // action function is replaced with `javascript:throw new Error('A React form was unexpectedly submitted')`
+  // See also https://github.com/facebook/react/blob/f83903bfcc5a61811bd1b69b14f0ebbac4754462/packages/react-dom-bindings/src/client/ReactDOMComponent.js#L468
+  // - DWS 2-12-25
+  it.skip("shows relevant errors returned by action", () => {
```
acouch, line 43: Thanks for stubbing this out and leaving it, hopefully support for this lands sooner than later.

acouch (approved):
👍🏿 thanks!

PR #3866, opened 2025-02-12
[Issue #3780] Snapshot and disable writes

### LINKED ISSUE #3780: Implement First Data Quality Check
### Summary

We need to implement the first data quality check as chosen by #3779. 

We will implement a data snapshot check to compare the data.

### Acceptance criteria

[ ] Snapshot checks to compare the shape and format of the data are in place
[ ] The snapshot checks are running as part of the pipeline to flag any bugs before code is merged

jcrichlake:
## Summary
Fixes #3780

## Changes proposed
Adding the syrupy plugin to the pytest framework to enable snapshot testing for the ETL pipeline.  We are testing the input of each step of the top level ETL function up to the database.   The snapshots 

The pipeline was ran with breakpoints to get sample data that went into the input directory for the snapshots to compare to.  The snapshots were then generated using the syrupy framework and stored in the __snapshots__ directory.

image


DavidDudas-Intuitial:
Looks good so far! I like the simplicity of the mock fixtures provided by syrupy. My only suggestion is that the input files can/should be shrunk to contain fewer objects, to abide by linter file size checks. Unless there's a legitimate need for such large input files? @jcrichlake

jcrichlake:
> Looks good so far! I like the simplicity of the mock fixtures provided by syrupy. My only suggestion is that the input files can/should be shrunk to contain fewer objects, to abide by linter file size checks. Unless there's a legitimate need for such large input files? @jcrichlake

The size of the files was purely due to my limited knowledge of the data set, I figured the bigger file the wider the net we'd cast to catch any issues.  But I can shrink them down that shouldn't be an issue.

analytics/pyproject.toml
```diff
@@ -28,6 +28,7 @@ boto3 = "^1.35.56"
 boto3-stubs = "^1.35.56"
 psycopg = "^3.2.3"
 smart-open = "^7.0.5"
+syrupy = "^4.8.1"
```
coilysiren, line 31: Pretty sure this is a dev dependency

jcrichlake: Updated this

widal001 (approved):
LGTM! Excited to get some snapshot testing set up to help us prevent regressions on our ETL.

The snapshots passed when I ran on the current branch:

Screenshot 2025-02-14 at 11 26 32 AM

And then they failed (as expected) when I changed the value of a field we were parsing to simulate introducing a bug to the pipeline.

Screenshot 2025-02-14 at 11 25 29 AM

PR #3882, opened 2025-02-13
[Issue #3852] Initialize background task app properly for New Relic

### LINKED ISSUE #3852: Adjust New Relic setup for backend ECS tasks to initialize an application
### Summary

We setup a New Relic background task on our backend ECS tasks: https://github.com/HHS/simpler-grants-gov/blob/main/api/src/task/ecs_background_task.py#L53

However this isn't completely working in New Relic (logs aren't being differentiated as web vs non-web). I think we need to initialize an application based on the docs: https://docs.newrelic.com/docs/apm/agents/python-agent/python-agent-api/backgroundtask-python-agent-api/#examples

This should just require adding:
```py
application = newrelic.agent.register_application(timeout=10.0)
```
And then passing that application into the background task method.

### Acceptance criteria

- [ ] New Relic setup updated

chouinar:
## Summary
Fixes #3852

## Changes proposed
Initialize an "application" for New Relic before starting the New Relic background task

## Context for reviewers
From reading the docs, initializing the application is necessary when running outside of our normal API approach (which initializes automatically on first request). The whole background app thing is to separate our API metrics from the backend ECS task metrics (for example, slow DB queries are much more impactful in an API call than some ECS task).

## Additional information
I set the newrelic key locally, and ran some of our scripts, they now appear as "transactions" under the non-web transcation type:
Screenshot 2025-02-13 at 12 32 10 PM

### DIFF
--- a/api/src/task/ecs_background_task.py
+++ b/api/src/task/ecs_background_task.py
@@ -48,11 +48,13 @@ def ecs_background_task(task_name: str) -> Callable[[Callable[P, T]], Callable[P
     def decorator(f: Callable[P, T]) -> Callable[P, T]:
         @wraps(f)
         def wrapper(*args: P.args, **kwargs: P.kwargs) -> T:
-            with _ecs_background_task_impl(task_name):
-                # Finally execute the function with New Relic instrumentation
-                return newrelic.agent.background_task(name=task_name, group="Python/ECSTask")(f)(
-                    *args, **kwargs
-                )
+            # Wrap with New Relic instrumentation
+            application = newrelic.agent.register_application(timeout=10.0)
+            with newrelic.agent.BackgroundTask(application, name=task_name, group="Python/ECSTask"):
+                # Wrap with our own logging (timing/general logs)
+                with _ecs_background_task_impl(task_name):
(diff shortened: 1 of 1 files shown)


api/src/task/ecs_background_task.py
```diff
@@ -48,11 +48,13 @@ def entrypoint(db_session: db.Session):
...
         @wraps(f)
         def wrapper(*args: P.args, **kwargs: P.kwargs) -> T:
-            with _ecs_background_task_impl(task_name):
-                # Finally execute the function with New Relic instrumentation
-                return newrelic.agent.background_task(name=task_name, group="Python/ECSTask")(f)(
-                    *args, **kwargs
-                )
+            # Wrap with New Relic instrumentation
+            application = newrelic.agent.register_application(timeout=10.0)
+            with newrelic.agent.BackgroundTask(application, name=task_name, group="Python/ECSTask"):
```
chouinar, line 53: FYI - I flipped the background task to wrap the ecs_background_task_impl method as it has some logs/operations it does that I want included in New Relic

mdragon (approved):
Approve

PR #3887, opened 2025-02-13
[no ticket] frontend has no database

coilysiren:
## Context

followup from: https://github.com/HHS/simpler-grants-gov/commit/9eca4df4f71f403c07165a5a3e71b02888daa529

The infra template changed the default value of "has database" from false to true, which resulted in the frontend trying to deploy a database

### DIFF
--- a/infra/frontend/service/main.tf
+++ b/infra/frontend/service/main.tf
@@ -145,9 +145,6 @@ module "service" {
 
   enable_alb_cdn = true
 
-  app_access_policy_arn      = null
-  migrator_access_policy_arn = null
-
   extra_environment_variables = merge({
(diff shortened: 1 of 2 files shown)


infra/frontend/service/main.tf
```diff
@@ -145,9 +145,6 @@ module "service" {
 
   enable_alb_cdn = true
 
-  app_access_policy_arn      = null
-  migrator_access_policy_arn = null
```
coilysiren, line 149: These are related to database stuff, hence them being hardcoded null

coilysiren: But we don't need them at all

chouinar (approved):

PR #3893, opened 2025-02-13
[Issue 3189] Refactor logging initialization 

### LINKED ISSUE #3189: Analytics `db-migrate` logs aren't tailing into GitHub actions
### Summary

Right now the logs for `db-migrate` aren't tailing into the GitHub actions due to how the logs are initialized. Fix this!

### Acceptance criteria

- [x] Logs will appear in the GitHub actions for deploying to staging, dev, and prod when we run the `db-migrate` step

DavidDudas-Intuitial:
## Summary
Fixes #3189

Logging was setup in such a way that assumed invocation only via the `Typer` app entry point. Thus, logging reportedly fails when invocation occurs via the `ecs_background_task` entry point. This PR refactors logger initialization and access so that logging will work irrespective of entry point.


analytics/src/analytics/cli.py
```diff
@@ -49,14 +46,25 @@
...
     init_app(logging.root)
 
 
+def get_logger() -> logging.Logger:
+    """Retrieve the logger instance, initializing it if necessary."""
+    global LOGGER  # noqa: PLW0603, pylint: disable=global-statement
+    if LOGGER is None:
+        LOGGER = logging.getLogger(__name__)
+        init_logging(__package__)
+    return LOGGER
```
coilysiren, line 65: I see what you're trying to do here, but I'm not sure this is the best way to do it. I'd want @chouinar's expert python opinion here

DavidDudas-Intuitial: Here's what I'm going for:

- lazy load logger to prevent initialization at import time
- replace global access pattern with a getter function

The lazy load should solve the problem of missing logs when db-migrate is run as an `@ecs_background_task`.

chouinar: Python logging shouldn't require lazy loading a logger.

It should always be safe to do:

```py
```
in a file even if you haven't instantiated any of the logging logic yet. Later configuring the logger (which is done globally generally) will cause its behavior to change, but since you should configure logging before any meaningful logging statements, shouldn't be an issue.

https://docs.python.org/3/library/logging.html

---

I'm not certain how this change fixes the problem you were seeing? It sounds like in the case you're talking about it's an issue of `init_logging` not always being called?

DavidDudas-Intuitial: Thanks for the feedback. Here's more context:

`cli.py` was already asynchronously (i.e. not at module load time) calling `init_logging(__package__)` within a `Typer` app callback. Why it was doing this, I don't know, but it prevented logging from being initialized for any module usage that does NOT use the `Typer` app entry point. One such usage is the `@ecs_background_task` workflow which does not use `Typer` thus the callback is never invoked and logging is never completely initialized. To fix this disparity I added a getter for accessing the logger object which lazy initializes the logger, thereby ensuring it is always initialized regardless of entry point.

DavidDudas-Intuitial: Having said that, it's possible that the previous pattern of initializing logging in the `Typer` callback is not necessary. I didn't write that code, I don't know why it was done that way. I'll test now to see if it is even necessary. If not, then this PR will get a lot simpler.

chouinar: Do you have an example of the behavior you're seeing in each scenario?

One thing that is confusing me is that ideally _every_ CLI command should be run the same way, there shouldn't be a distinction between those with/without the ecs_background_task - that was only added to one job, but should be on every one (it just adds more logging/metrics). If there's currently 2 approaches to running a command, we should fix that to be just one approach.

chouinar: It's entirely possible that the way I pulled the logging logic over conflicts with typer, and a fix is necessary, but the fix should end with `logger = logging.getLogger(__name__)` always being safe to toss at the top of any file. If anything is preventing that, then the fix is in the configuration setup somewhere

DavidDudas-Intuitial: The async logging initialization in the Typer app callback is seemingly not necessary. I just pushed a commit that initializes logs the standard way, and as a result `cli.py` logging appears to work as expected.

DavidDudas-Intuitial:
Problem solved: https://github.com/HHS/simpler-grants-gov/actions/runs/13336373477/job/37252660439

Screenshot 2025-02-14 at 11 58 53 AM

widal001 (approved):
@DavidDudas-Intuitial tested the deploy and it demonstrated the logs tailing into the CD.

PR #3902, opened 2025-02-14
[Unticketed] Adjust describe-log-streams command to not use query

chouinar:
## Changes proposed
Change the run-command script we use for running ECS tasks to filter the log stream using `--log-stream-name-prefix` instead of `--query`

## Context for reviewers
We were seeing two issues that this hopefully resolves:
* This command was taking 7+ minutes to resolve
* This command frequently was getting rate limited and failing deployments (likely due to a lot of internal queries)

Testing this change to run an ECS task locally, it takes 1~2 seconds to run now

## Additional information
Was able to successfully use the script to run the transform script (setup to no-op), and did not need to wait long for logs to start flowing.

### DIFF
--- a/bin/run-command
+++ b/bin/run-command
@@ -144,7 +144,7 @@ while true; do
     echo "Timing out task $ecs_task_id waiting for logs"
     exit 1
   fi
-  is_log_stream_created=$(aws logs describe-log-streams --no-cli-pager --log-group-name "$log_group" --query "length(logStreams[?logStreamName==\`$log_stream\`])")
+  is_log_stream_created=$(aws logs describe-log-streams --no-cli-pager --log-group-name "$log_group" --log-stream-name-prefix "$log_stream" | jq -r '.logStreams | length')
   if [ "$is_log_stream_created" == "1" ]; then
     break
   fi


bin/run-command
```diff
@@ -144,7 +144,7 @@ while true; do
     echo "Timing out task $ecs_task_id waiting for logs"
     exit 1
   fi
-  is_log_stream_created=$(aws logs describe-log-streams --no-cli-pager --log-group-name "$log_group" --query "length(logStreams[?logStreamName==\`$log_stream\`])")
+  is_log_stream_created=$(aws logs describe-log-streams --no-cli-pager --log-group-name "$log_group" --log-stream-name-prefix "$log_stream" | jq -r '.logStreams | length')
   if [ "$is_log_stream_created" == "1" ]; then
```
coilysiren, line 148: I'm wondering if this if statement down here will trip us up? Can you test that part too pls 🙏🏽

coilysiren: Oh duh I didn't read to the end of your PR description sorry 🤦🏽‍♀️

coilysiren (approved):

bin/run-command
chouinar: It looks to work fine when I ran `run-command` locally. It looped a few times before exiting the loop and printing the logs (the next part of the script).

Even just testing it outside the script, if I run with a valid log group: `aws logs describe-log-streams --no-cli-pager --log-group-name "service/api-dev" --log-stream-name-prefix "api-dev/api-dev/b83f296e29b8412e8446e831150c5518" | jq -r '.logStreams | length'` produces `1`. It I change the log stream name to a random string, it prints `0`

Issue #3874, opened 2025-02-12
Create rough tables representing an opportunity application package


chouinar:
### Summary

Note that this does not contain every column we will need, and doesn't factor in _any_ of the tables needed for actually applying (ie. a place for users forms to go).

All tables should have our usual updated_at/created_at timestamps

Competition table
* competition_id - uuid, primary key
* opportunity_id - int, foreign key, indexed
* legacy_competition_id - int | None, indexed
* public_competition_id - str | None (?? This is the one that actually displays sometimes on the legacy website, how is it used?)
* legacy_package_id - str | None (?? This also displayed on the legacy website - weirdly it's not in the package table, only in the competition table)
* competition_title - str | None
* opening_date - date | None (TODO - what does it mean when this is null?)
* closing_date - date | None (TODO - what does it mean when this is null?)
* grace_period - int | None
* contact_info - str | None
* TODO - open to applicant type? - how do we want to handle that? Do we still want the options as individual/organization/both (if so - make a lookup table with a one-to-many relationship)

Competition Instruction table
* competition_instruction_id - uuid, primary key
* competition_id - uuid, foreign key to competition table
* file_location - str (similar to opportunity attachment in how it will work)

CompetitionAssistanceListing - a many-to-many table between competition and opportunity assistance listing 
* competition_id - uuid, primary key
* opportunity_assistance_listing_id - int, primary key

Form (Do we want a more descriptive name like "Application Form"?)
* form_id - uuid, primary key
* form_name - str
* form_version - str
* is_active - bool
* description - str (nullable?)
* agency_code_id - str (TODO - can we make this a foreign key? I'm not sure it actually correlates to agencies as "Grants.gov" is a value in the current system)
* omb_number - str | None
* form_schema? - TODO - presuambly jsonb type but we need to hash out the details (can hold on adding for now?)

CompetitionForm
* competition_id - uuid, primary key, foreign key
* form_id - uuid, primary key, foreign key
* is_required - bool (whether or not the form falls into the optional or mandatory group)
* TODO - is there anything else? The legacy system had an additional table for "opportunity package group" but I'm not sure what that actually did

### Acceptance criteria

- [ ] Tables setup
- [ ] Factories setup and working for tables

PR #3906, opened 2025-02-14
[Issue #3874] Create Rough Tables for Opportunity Applicate package

babebe:
## Summary
Fixes #{[3874](https://github.com/HHS/simpler-grants-gov/issues/3874)}

### Time to review: __15 mins__

**Rough Tables  Created**
opportunity_application_form
opportunity_competition_form
opportunity_competition
opportunity_competition_assistance_listing
opportunity_competition_instruction

Factories created


api/src/db/models/opportunity_models.py
```diff
@@ -451,3 +457,80 @@ class OpportunityChangeAudit(ApiSchemaTable, TimestampMixin):
         BigInteger, ForeignKey(Opportunity.opportunity_id), primary_key=True, index=True
     )
     opportunity: Mapped[Opportunity] = relationship(Opportunity)
+
+
+class OpportunityCompetition(ApiSchemaTable, TimestampMixin):
```
chouinar, line 462: Could we put these in a `competition_models.py` file - just want to avoid the opportunity file getting really long, and there are going to be a lot of additional tables we need for this. We can also name the tables without prefixing with `Opportunity` - they're all sorta a part of it, but separate enough

api/src/db/models/opportunity_models.py
```diff
@@ -93,6 +95,10 @@ def agency(self) -> str | None:
         viewonly=True,
     )
 
+    opportunity_competition: Mapped[list["OpportunityCompetition"]] = relationship(
+        back_populates="opportunity", uselist=True, cascade="all, delete-orphan"
+    )
```
chouinar, line 100: FYI - if you move these models to another file, you might hit an issue with this having a circular import. You can work around that by adding the import to the `if TYPE_CHECKING` line towards the top of this file: https://github.com/HHS/simpler-grants-gov/blob/main/api/src/db/models/opportunity_models.py#L26

api/src/db/models/opportunity_models.py
```diff
@@ -451,3 +457,80 @@ class OpportunityChangeAudit(ApiSchemaTable, TimestampMixin):
...
+    opportunity_competition: Mapped[OpportunityCompetition] = relationship(OpportunityCompetition)
+
+    file_location: Mapped[str]
+
+
+class OpportunityCompetitionAssistanceListing(ApiSchemaTable, TimestampMixin):
+    __tablename__ = "opportunity_competition_assistance_listing"
+
+    competition_id: Mapped[uuid.UUID] = mapped_column(
+        UUID, ForeignKey(OpportunityCompetition.competition_id), index=True, primary_key=True
```
chouinar, line 500: We don't need to index if it's the primary key - I might have something like that on some of our many-to-many tables, but that was a holdover from me testing some performance ideas a while back (I don't think they ever worked?)

api/tests/src/db/models/factories.py
```diff
@@ -1821,3 +1821,64 @@ class Meta:
...
+
+
+class OpportunityCompetitionFactory(BaseFactory):
+    class Meta:
+        model = opportunity_models.OpportunityCompetition
+
+    opportunity = factory.SubFactory(OpportunityFactory)
+    opportunity_id = factory.LazyAttribute(lambda o: o.opportunity.opportunity_id)
+
+    competition_id = Generators.UuidObj
```
chouinar, line 1833: Could we put the primary key at the top of the factories? I generally just keep everything in the same order as the class itself.

api/tests/src/db/models/factories.py
```diff
@@ -1821,3 +1821,64 @@ class Meta:
...
+    legacy_competition_id = sometimes_none(factory.Faker("random_int", min=1, max=15))
+    public_competition_id = sometimes_none(factory.Faker("random_int", min=1, max=15))
+    legacy_package_id = sometimes_none(factory.Faker("random_int", min=1, max=15))
```
chouinar, line 1836: For these, my recommendation:
* don't set legacy_competition_id -> it's a pointer back to the old system, only our transform logic will ever care about it / want it set.
* public_competition_id -> It's actually a string with some variance in how it can be set, I'd just put something like `ABC<number>` to make it look sorta like the old system (keep the sometimes_none)
* legacy_package_id - This looks to be formatted as `PKG<number>`, probably just do that as well? (keep the sometimes_none).

api/tests/src/db/models/factories.py
```diff
@@ -1821,3 +1821,64 @@ class Meta:
...
+
+    competition_title = sometimes_none(factory.Faker("sentence"))
+
+    opening_date = factory.Faker("date_between", start_date="-3w", end_date="-1d")
+    closing_date = factory.LazyAttribute(
+        lambda o: fake.date_time_between(start_date=o.opening_date)
+    )
+
+    grace_period = factory.Faker("random_int", min=1, max=10)
+    contact_info = sometimes_none(factory.Faker("sentence"))
```
chouinar, line 1846: I setup a contact info faker provider previously, could use `agency_contact_description` for this to get it to roughly look like contact info

api/tests/src/db/models/factories.py
```diff
@@ -1821,3 +1821,64 @@ class Meta:
...
+class OpportunityApplicationForm(BaseFactory):
+    class Meta:
+        model = opportunity_models.OpportunityApplicationForm
+
+    form_id = Generators.UuidObj
+    form_name = factory.Faker("sentence")
+    form_version = factory.Faker("sentence")
+    is_active = False
+    description = factory.Faker("sentence")
+    agency_code_id = factory.Faker("sentence")
```
chouinar, line 1873: Some suggestions:
* `form_name` -> `bs` might be a good one: https://faker.readthedocs.io/en/master/providers/faker.providers.company.html#faker.providers.company.Provider.bs
* `form_version` -> I'd probably say we should make this formatted like `1.<random int>` - could randomize both, don't think it matters much
* `is_active` -> default to true, if it's inactive, we'll probably disallow the form for use in certain ways
* `agency_code_id` -> Just reuse the `agency_code` faker provider we made
*

api/src/db/models/opportunity_models.py
```diff
@@ -451,3 +457,80 @@ class OpportunityChangeAudit(ApiSchemaTable, TimestampMixin):
...
+
+class OpportunityApplicationForm(ApiSchemaTable, TimestampMixin):
+    __tablename__ = "opportunity_application_form"
+
+    form_id: Mapped[uuid.UUID] = mapped_column(UUID, primary_key=True, default=uuid.uuid4)
+    form_name: Mapped[str]
+    form_version: Mapped[str]
+    is_active: Mapped[bool]
+    description: Mapped[str]
+    agency_code_id: Mapped[str]
```
chouinar, line 522: On second thought, let's just name this `agency_code` like we have on the opportunity

api/src/db/models/opportunity_models.py
```diff
@@ -451,3 +457,80 @@ class OpportunityChangeAudit(ApiSchemaTable, TimestampMixin):
...
+
+
+class OpportunityApplicationForm(ApiSchemaTable, TimestampMixin):
+    __tablename__ = "opportunity_application_form"
+
+    form_id: Mapped[uuid.UUID] = mapped_column(UUID, primary_key=True, default=uuid.uuid4)
+    form_name: Mapped[str]
+    form_version: Mapped[str]
+    is_active: Mapped[bool]
+    description: Mapped[str]
```
chouinar, line 521: Yknow what, on second thought, I don't even know what this description refers to, let's drop this column. It's not on the form UI, so let's hold off on it for now

api/src/db/models/opportunity_models.py
babebe: Yes, makes sense. Removed

api/src/db/models/opportunity_models.py
babebe: Done

api/tests/src/db/models/factories.py
babebe: Moved to top

mdragon (changes requested):

api/src/db/migrations/versions/2025_02_14_add_competition_tables.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        "application_form",
+        sa.Column("form_id", sa.UUID(), nullable=False),
+        sa.Column("form_name", sa.Text(), nullable=False),
+        sa.Column("form_version", sa.Text(), nullable=False),
+        sa.Column("is_active", sa.Boolean(), nullable=False),
```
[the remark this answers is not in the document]
babebe: @mdragon These were initially considered basic columns, with the plan to add more as needed. I believe it's a good idea to include both inactive_at and active_at columns, as it's generally better for each column to serve a single purpose rather than combining logic. Thoughts?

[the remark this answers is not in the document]
mdragon: I think that makes sense for now and then yes, we absolutely shouldn't try to make perfect tables out of the gate and anticipate revisions and improvements as we work with the tables and learn more about the functionality. 👍🏻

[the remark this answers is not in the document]
babebe: Added `inactive_at` -Nullable  and `active_at`-Non-Nullable `datetime` columns and updated factory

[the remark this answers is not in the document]
mdragon: I would maybe let active_at be null too for now. I could see someone setting up a form but not knowing when it will be ready to be made active and so we'd force them to put in some random future date to be able to even start working on the form. I think it being null is a better signal that it's not scheduled to be active yet.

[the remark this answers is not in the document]
mdragon: Sorry, I thought we were replacing is_active with the active_at. I think we can just remove is_active for now and we can use a active_at <= now() kind of logic to determine if something is active rather than having the boolean too.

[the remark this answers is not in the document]
babebe: removed is_active and made active_at nullable

mdragon (approved):
👍🏻

Issue #3849, opened 2025-02-11
Remove transformations from legacy history tables to our non-history tables


chouinar:
### Summary

As part of our work to re-implement versioning, we first need to stop bringing in historical data to our non-history tables (they'll later be added to history tables).

We need to adjust 4 subtasks of the transformation job:
* [Applicant Type](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/data_migration/transformation/subtask/transform_applicant_type.py)
* [Funding Category](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/data_migration/transformation/subtask/transform_funding_category.py)
* [Funding Instrument](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/data_migration/transformation/subtask/transform_funding_instrument.py)
* [Opportunity Summary](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/data_migration/transformation/subtask/transform_opportunity_summary.py)

Each of these jobs follows the same pattern of doing 4 separate fetches on synopsis/forecast/synopsis_history/forecast_history based tables to get data that we transform. For all of these, we want to remove the transformations for the two history tables.

This change should be simple, as we just need to remove the query + processing call for each of them. However the unit tests have a lot of scenarios built into them that will need to be cleaned up.

Please consider breaking this into several PRs if the changes are growing too large/confusing.

### Acceptance criteria

- [ ] Historical tables no longer transformed
- [ ] Unit tests cleaned up to remove testing of historical scenarios

PR #3914, opened 2025-02-18
[Issue #3901] Cleanup remainder of historical transformation code

### LINKED ISSUE #3901: Cleanup remainder of historical transformation code
### Summary

After https://github.com/HHS/simpler-grants-gov/issues/3849 disabled the historical transformation code, we have a few unused bits of code remaining that we can clean up:

* We no longer need special logic for the revision_number (will be deleted soon): https://github.com/HHS/simpler-grants-gov/blob/main/api/src/data_migration/transformation/transform_util.py#L212
* We no longer need the logic around processing action_type (a historical only field): https://github.com/HHS/simpler-grants-gov/blob/main/api/src/data_migration/transformation/transform_util.py#L468 - can also remove any usage of it
* Can remove revision_number from all of the `get_log_extra_x` methods: https://github.com/HHS/simpler-grants-gov/blob/main/api/src/data_migration/transformation/transform_util.py#L500
* Can remove `_is_orphaned_historical` and `_handle_orphaned_historical` - https://github.com/HHS/simpler-grants-gov/blob/main/api/src/data_migration/transformation/subtask/abstract_transform_subtask.py#L101 - several tasks reference these, can remove the if-statements that use them.
* The `fetch_with_opportunity_summary` method takes in a field for whether the query is for historical records, that can be removed: https://github.com/HHS/simpler-grants-gov/blob/main/api/src/data_migration/transformation/subtask/abstract_transform_subtask.py#L190
* We have a few type aliases for various fields, which are unions of the historical/non-historical tables. We can remove the historical tables: https://github.com/HHS/simpler-grants-gov/blob/main/api/src/data_migration/transformation/transform_constants.py#L59
* We no longer need a metric for `TOTAL_HISTORICAL_ORPHANS_SKIPPED` - https://github.com/HHS/simpler-grants-gov/blob/main/api/src/data_migration/transformation/transform_constants.py#L49

Since we already removed any code that uses these branching paths, this _shouldn't_ break any tests.

### Acceptance criteria

- [ ] Unused code removed

mikehgrantsgov:
## Summary
Fixes #3901

## Changes proposed
Remove special logic for the revision_number
Remove logic around processing action_type (a historical only field)
Remove revision_number from all of the get_log_extra_x methods
Remove _is_orphaned_historical and _handle_orphaned_historical
Remove fetch_with_opportunity_summary param to check whether the query is for historical records
Remove some related type aliases
Remove metric TOTAL_HISTORICAL_ORPHANS_SKIPPED

## Context for reviewers
After #3849 disabled the historical transformation code, we have a few unused bits of code remaining that we can clean up.

## Additional information
Check with @chouinar if the tests removed in this PR can actually be/should have been previously removed, as stated the code changes should not have broken any tests.


api/src/db/models/legacy_mixin/forecast_mixin.py
```diff
@@ -78,7 +78,6 @@ class TforecastHistMixin:
     est_project_start_date: Mapped[datetime.datetime | None]
     fiscal_year: Mapped[int | None]
     modification_comments: Mapped[str | None]
-    action_type: Mapped[str | None]
```
chouinar, line 81: We can't change these tables - the column order needs to match the Oracle table's column order exactly or the foreign data wrapper breaks

chouinar (changes requested):
The changes look good, but need to be careful - can't change the schema of the staging/foreign tables due to how we use the foreign data wrapper.

api/src/db/models/legacy_mixin/forecast_mixin.py
mikehgrantsgov: Makes sense. I restored this

chouinar: You still have the change on the other mixin table - shouldn't have any migration from this PR

mikehgrantsgov: Ah I see. Removed the migration and reference to action_type. Thanks for the clarification

chouinar (approved):
LGTM!

Issue #3860, opened 2025-02-12
Create a table that stores opportunity versions


chouinar:
### Summary

Create a table called `opportunity_version` with the following columns:
* opportunity_version_id: uuid, primary key
* opportunity_id: int, primary key, foreign key
* opportunity_data: jsonb
* created_at/updated_at as usual

### Acceptance criteria

- [ ] Tables setup
- [ ] Factory added

PR #3917, opened 2025-02-18
[Issue #3860] Create opportunity_version Table

babebe:
## Summary
Fixes #{[3860](https://github.com/HHS/simpler-grants-gov/issues/3860)}

## Changes proposed
Added OpportunityVersion Model
Added OpportunityVersionFactory
Migration Script to create table


api/src/db/migrations/versions/2025_02_19_add_competition_tables.py
```diff
@@ -0,0 +1,55 @@
...
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        "opportunity_version",
+        sa.Column("opportunity_version_id", sa.UUID(), nullable=False),
+        sa.Column("opportunity_id", sa.BigInteger(), nullable=False),
+        sa.Column("opportunity_data", postgresql.JSONB(astext_type=sa.Text()), nullable=True),
```
mdragon, line 26: Probably want to make this not nullable.

api/src/db/migrations/versions/2025_02_19_add_competition_tables.py
```diff
@@ -0,0 +1,55 @@
+"""add competition tables
```
mdragon, line 1: Just flagging this seems to be lingering from the last round, along with the file name.

api/src/db/migrations/versions/2025_02_19_add_competition_tables.py
babebe: Done!

api/src/db/migrations/versions/2025_02_19_add_competition_tables.py
babebe: Updated!

mdragon (approved):
Approved

mdragon (approved):
👍🏻

PR #3918, opened 2025-02-18
[Issue 2968] Add execution timers to ETL workflow logging

### LINKED ISSUE #2968: Refactor insert pattern for GitHub transform and load
### Summary

Currently, the pattern adopting for upserting data in `etldb/main.py` is to:

1. Filter for the unique set of Github IDs
2. Iterate through that list of IDs
3. Then filter the _dataset_ again to get data for a given ID

See the following sections for examples:

- [Deliverables](https://github.com/HHS/simpler-grants-gov/blob/873c776f337018ce01e78bad6c40feee3d232fdd/analytics/src/analytics/integrations/etldb/main.py#L159-L163)
- [Epics](https://github.com/HHS/simpler-grants-gov/blob/873c776f337018ce01e78bad6c40feee3d232fdd/analytics/src/analytics/integrations/etldb/main.py#L171-L173)
- [Issues](https://github.com/HHS/simpler-grants-gov/blob/873c776f337018ce01e78bad6c40feee3d232fdd/analytics/src/analytics/integrations/etldb/main.py#L183-L185)

A more common pattern would be to select the distinct set of values you want to insert using pandas [DataFrame.drop_duplicates()](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.drop_duplicates.html)  method and _then_ iterate through the list of data inserting each row or _preferably_ doing a _bulk_ upsert.

In addition to requiring multiple passes through the dataset with $`On^k`$ complexity (where $`n` is the size of the source DataFrame and $`k`$ is the number of unique IDs for a given table) vs a single pass to retrieve all data we want to insert, this pattern also makes it harder to adopt bulk upserts.

Bulk transactions are helpful because they allow us to combine DML into transaction blocks more easily and prevent instances in which tables are only partially updated during a batch data load process.

### Acceptance criteria

- [ ] We've reduced the number of passes through the dataframe needed to extract the unique set of values to insert into the database
- [ ] The data is loaded using bulk DML statements rather than separate insert/update statements for each row of data.

DavidDudas-Intuitial:
## Summary
Fixes #2968 

Adds a few simple execution timers so we can easily measure and inspect how long the ETL workflow takes to execute.

It was suggested that we optimize the ETL workflow insert pattern. A precursor to any optimization is measuring the baseline performance. This PR introduces simple execution timers to measure baseline performance.

### DIFF
--- a/analytics/src/analytics/cli.py
+++ b/analytics/src/analytics/cli.py
@@ -2,9 +2,9 @@
 # pylint: disable=C0415
 """Expose a series of CLI entrypoints for the analytics package."""
 
-import time
 import logging
 import logging.config
+import time
 from datetime import datetime
 from pathlib import Path
 from typing import Annotated
@@ -89,7 +89,7 @@ def export_github_data(
     start_time = time.perf_counter()
     GitHubProjectETL(config).run()
     end_time = time.perf_counter()
-    logger.info("extract workflow executed in %.5f seconds", end_time-start_time)
+    logger.info("extract workflow executed in %.5f seconds", end_time - start_time)
 
 
 # ===========================================================
@@ -154,7 +154,7 @@ def transform_and_load(
     # sync data to db
     etldb.sync_data(dataset, datestamp)
     end_time = time.perf_counter()
-    logger.info("transform and load executed in %.5f seconds", end_time-start_time)
+    logger.info("transform and load executed in %.5f seconds", end_time - start_time)
 
 
 @etl_app.command(name="extract_transform_and_load")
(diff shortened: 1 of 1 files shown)


analytics/src/analytics/cli.py
```diff
@@ -140,16 +146,15 @@ def transform_and_load(
...
+    start_time = time.perf_counter()
     dataset = EtlDataset.load_from_json_file(file_path=issue_file)
-
     # sync data to db
     etldb.sync_data(dataset, datestamp)
-
-    # finish
-    logger.info("transform and load is done")
+    end_time = time.perf_counter()
+    logger.info("transform and load executed in %.5f seconds", end_time - start_time)
```
jcrichlake: Should this also say workflow to match the logger above?

jcrichlake (approved):

analytics/src/analytics/cli.py
DavidDudas-Intuitial: There's a unit test that checks for this output string, so I left the message as-is rather than update the test :)

PR #3929, opened 2025-02-19
[Issue #3607] Add saved tag on search results

### LINKED ISSUE #3607: Add indicators on search results for saved opportunities
### Summary

When a search result contains opportunities that a logged in user has saved to their account, the search results should reflect this by including a saved indicator in the result

## Designs

https://www.figma.com/design/FGxWtAgToKhehLJCiuy1zL/Simpler.Grants.gov-Project?node-id=1952-3669

Image

### Acceptance criteria

- [ ] for logged in users, on the search page, all search results representing an opportunity that a user has saved contain a "saved" indicator
- [ ] "saved" indicator matches designs
- [ ] "saved" indicator is non-interactive

acouch:
## Summary
Fixes #3607

## Changes proposed

1. Adds saved tag to search results

image

### Desktop

image

### Mobile

image

2. Cleans up `SearchResultListItem` to remove redundant html

3. Adds story for SearchResultListItem

## Testing

1. Spin up local, including API
2. Visit search page, see no changes to current dev or prod
3. Sign in, save an opportunity
4. See "Saved" on opportunity in search page

### DIFF
--- a/frontend/src/components/search/SearchResultsListItem.tsx
+++ b/frontend/src/components/search/SearchResultsListItem.tsx
@@ -78,7 +78,7 @@ export default function SearchResultsListItem({
                 <span className="padding-x-105 padding-y-2px bg-base-lighter display-flex flex-align-center font-sans-3xs radius-sm">
                   <USWDSIcon
                     name="star"
-                    className="text-accent-warm-dark button-icon-large padding-right-05"
+                    className="text-accent-warm-dark button-icon-md padding-right-05"
                   />
                   {t("opportunitySaved")}
                 </span>
--- a/frontend/src/components/search/SearchResultsListItem.tsx
+++ b/frontend/src/components/search/SearchResultsListItem.tsx
@@ -24,7 +24,6 @@ export default function SearchResultsListItem({
   const t = useTranslations("Search");
 
   const metadataBorderClasses = `
-    display-block
     tablet:display-inline-block
     tablet:border-left-1px
     tablet:padding-x-1
@@ -57,7 +56,7 @@ export default function SearchResultsListItem({
                 </Link>
               </h2>
             </div>
-            <div className="font-body-xs display-flex">
+            <div className="font-body-xs display-flex flex-wrap">
               <SearchResultListItemStatus
                 archiveDate={opportunity?.summary?.archive_date}
                 archivedString={t("resultsListItem.status.archived")}
@@ -67,7 +66,9 @@ export default function SearchResultsListItem({
                 postedString={t("resultsListItem.status.posted")}
                 status={opportunity?.opportunity_status}
               />
-              <span className={metadataBorderClasses}>
+              <span
+                className={`${metadataBorderClasses} tablet:order-0 order-2`}
+              >
                 <strong>{t("resultsListItem.summary.posted")}</strong>
                 {opportunity?.summary?.post_date
                   ? formatDate(opportunity?.summary?.post_date)
@@ -86,6 +87,7 @@ export default function SearchResultsListItem({
                   </div>
                 </div>
               )}
+              <div className="width-full tablet:width-auto" />
             </div>
             <div className="grid-col tablet:order-2 overflow-hidden font-body-xs">
               <strong>{t("resultsListItem.summary.agency")}</strong>
--- a/frontend/src/components/search/SearchResultsListItem.tsx
+++ b/frontend/src/components/search/SearchResultsListItem.tsx
@@ -7,14 +7,19 @@ import { AgencyNamyLookup } from "src/utils/search/generateAgencyNameLookup";
 import { useTranslations } from "next-intl";
 import Link from "next/link";
 
+import { USWDSIcon } from "src/components/USWDSIcon";
+import SearchResultListItemStatus from "./SearchResultListItemStatus";
+
 interface SearchResultsListItemProps {
   opportunity: Opportunity;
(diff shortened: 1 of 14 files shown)


andycochran:
First thought: It'd be rad if there was a Storybook story for the `SearchResultsListItem` component.

frontend/src/components/search/SearchResultsListItem.tsx
```diff
@@ -52,87 +56,67 @@ export default function SearchResultsListItem({
...
+              <span
+                className={`${metadataBorderClasses} tablet:order-0 order-2`}
+              >
                 <strong>{t("resultsListItem.summary.posted")}</strong>
                 {opportunity?.summary?.post_date
                   ? formatDate(opportunity?.summary?.post_date)
                   : "--"}
               </span>
+              {saved && (
+                <div className="bg-base-lighter padding-x-105 height-205 font-body-3xs">
```
andycochran, line 78: Looks like the saved indicator is a different height than the `SearchResultListItemStatus`, which uses `.usa-tag`. I wonder if this should also use the USWDS tag class so that they're the same. And if it used spans instead of divs, that might simplify the layout (the status and saved tags could both be spans in the div that controls the order).

snyder-ux:
> the Figma design seems to have removed the light gray metadata separator/border between the data items

@andycochran The intent was to just show the new elements fitting in and **not** change existing search results UI.
Reason for discrepancy: The latest version of Figma and what is in prod are not identical. The design was based on the latest version of Figma for that time (Archive: Search-20240508-CR-v07) and I updated the colors to the new brand style. I should have taken a screen capture instead but didn't realize they were different. This was my attempt to update an existing Figma screen to the brand styles, colors.

snyder-ux:
@doug-s-nava called out wonky UI at various breakpoints for responsive design. We discussed that with upcoming search page updates coming, @crystabelrangel would better define the UI then and it wasn't worth investing in improving those wonky bits now.

frontend/src/components/search/SearchResultsListItem.tsx
```diff
@@ -52,87 +56,67 @@ export default function SearchResultsListItem({
...
+                archivedString={t("resultsListItem.status.archived")}
+                closedDate={opportunity?.summary?.close_date}
+                closedString={t("resultsListItem.status.closed")}
+                forecastedString={t("resultsListItem.status.forecasted")}
+                postedString={t("resultsListItem.status.posted")}
+                status={opportunity?.opportunity_status}
+              />
+              <span
+                className={`${metadataBorderClasses} tablet:order-0 order-2`}
+              >
```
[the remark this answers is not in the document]
andycochran: Per Jenn's comments, let's ignore this. Search v2 can clean it all up.

frontend/src/components/search/SearchResultsListItem.tsx
acouch: @andycochran `.usa-tag` doesn't affect the height, just the padding. It also sets a number of items like font color, border radius, text-transform that need to be overwritten, so I don't think it is a good candidate here:

image

acouch:
>  First thought: It'd be rad if there was a Storybook story for the SearchResultsListItem component. With controls for previewing the variants.

@andycochran There is a storybook story for this included in the PR 🚀

snyder-ux:
Looks great! Could the tags be the same height and horizontally aligned? If it's gnarly to adjust, then it's not worth the effort considering the Search page v2 updates coming soon.
Screenshot 2025-02-21 at 10 06 01 AM

acouch:
I pushed an update to make sure they are the same size: 
image

acouch:
Noting in Figma that they aren't aligned:

image

acouch:
Pushed one more update to tweak the size and border radius

image

frontend/src/components/search/SearchResultsListItem.tsx
```diff
@@ -75,12 +75,12 @@ export default function SearchResultsListItem({
                   : "--"}
               </span>
               {saved && (
-                <span className="padding-x-105 padding-y-1px bg-base-lighter display-flex flex-align-center font-sans-2xs">
-                    <USWDSIcon
-                      name="star"
-                      className="text-accent-warm-dark button-icon-large padding-right-05"
-                    />
-                    {t("opportunitySaved")}
+                <span className="padding-x-105 padding-y-2px bg-base-lighter display-flex flex-align-center font-sans-3xs radius-sm">
```
andycochran, line 78: Nit: All the size/padding/radius classes could probably be replaced with `.usa-tag`

acouch: @acouch `.usa-tag` applies a text-transform of uppercase. USWDS doesn't have a `text-transform: capitalize` so it would need to be a custom tag:

image 

Additionally the padding is different because the font size of the "Saved" is different.

So it is `usa-tag simpler-text-capitalize color-black` plus the same padding and font-size additions.

acouch: Noting in Figma this is a "tag2". I assume that this started from the USWDS Figma file. Might be helpful to note any of the variants ahead of time to address how they can be addressed

image

andycochran: D'oh! I see. Yeah, we don't want these ALL CAPS YELLING

andycochran (approved):

PR #3937, opened 2025-02-19
[#3775] ETL Fix for broken project objects from private repos

### LINKED ISSUE #3775: Issues from Private Repos break `gh-data-export` step
### Summary

For certain users if there are issues from a private repo in a project the `gh-data-export` step will fail locally.  

### Acceptance criteria

Users can run the `gh-data-export` step regardless of if they have access to the repository

jcrichlake:
## Summary
Fixes #3775 

## Changes proposed
A check in the validation step to avoid empty objects since they were causing errors in the process.

## Context for reviewers
If an item from is added from a project to one of the ETL sprint boards that the GitHub token doesn't have access to the ETL process will fail instead of just omitting the bad data.

### Failure output
image

### Fixed output
image

### DIFF
--- a/analytics/src/analytics/integrations/github/main.py
+++ b/analytics/src/analytics/integrations/github/main.py
@@ -30,43 +30,46 @@ def transform_project_data(
 
     for i, item in enumerate(raw_data):
         try:
+            # Filter out invalid content from boards local user may not have permission to
+            if item["content"] is not None:
             # Validate and parse the raw item
-            validated_item = ProjectItem.model_validate(item)
+                validated_item = ProjectItem.model_validate(item)
+                # Skip excluded issue types
+                if validated_item.content.issue_type.name in excluded_types:
+                    continue
 
-            # Skip excluded issue types
-            if validated_item.content.issue_type.name in excluded_types:
-                continue
-
-            # Transform into flattened format
-            transformed = {
-                # project metadata
-                "project_owner": owner,
-                "project_number": project,
-                # issue metadata
-                "issue_title": validated_item.content.title,
-                "issue_url": validated_item.content.url,
-                "issue_parent": validated_item.content.parent.url,
-                "issue_type": validated_item.content.issue_type.name,
-                "issue_status": validated_item.status.name,
-                "issue_is_closed": validated_item.content.closed,
-                "issue_opened_at": validated_item.content.created_at,
-                "issue_closed_at": validated_item.content.closed_at,
-                "issue_points": validated_item.points.number,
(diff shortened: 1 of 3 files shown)


analytics/src/analytics/integrations/github/main.py
```diff
@@ -30,43 +30,45 @@ def transform_project_data(
...
-                "deliverable_pillar": validated_item.pillar.name,
-                "quad_id": validated_item.quad.iteration_id,
-                "quad_name": validated_item.quad.title,
-                "quad_start": validated_item.quad.start_date,
-                "quad_length": validated_item.quad.duration,
-                "quad_end": validated_item.quad.end_date,
-            }
-            transformed_data.append(transformed)
+            # Filter out invalid content from boards local user may not have permission to
+            if item["content"] is not None:
```
coilysiren, line 34: `if item["content"] is None: continue` would have been a but cleaner here, but that's a nit

coilysiren (approved):
👍🏼 !

analytics/src/analytics/integrations/github/main.py
widal001: @jcrichlake I would actually like to follow @coilysiren's here because it also gives us an opportunity to log a warning / info-level note that we're skipping a project item because it doesn't have any content.

Also we might want to use `.get()` instead of `["content"]` so we avoid raising an exception and halting the entire pipeline if the "content" key is missing for one item.

```python
if item.get("content") is None:
    logger.info("Row %d is missing the 'content' key, skipping.", i)
    continue 
```

analytics/tests/integrations/github/test_validation.py
```diff
@@ -42,6 +43,36 @@
 }
 
 
+class TestTransformProjectData:
```
widal001, line 46: Nice test! Clean and simple.

Would it be worth adding a line like this to the snapshot test as well?

analytics/tests/integrations/github/test_validation.py
```diff
@@ -42,6 +43,36 @@
...
+        """Test validating a project item that has none for it's content filters that item out."""
+        data = [
+            {
+                "content": None,
+                "status": None,
+                "sprint": None,
+                "points": None,
+                "quad": None,
+                "pillar": None,
+            },
```
widal001, line 59: Would it be worth adding a line like this to the snapshot input data too?

jcrichlake: updated

coilysiren (approved):

Issue #3863, opened 2025-02-12
Create a utility that given an opportunity will store it into the opportunity version table


chouinar:
### Summary

After https://github.com/HHS/simpler-grants-gov/issues/3860

This utility won't be used yet, we just want to setup a function that does the following:
* Takes in a db session + opportunity
* Converts the opportunity into JSON
* Creates a record in the opportunity summary table

### Acceptance criteria

- [ ] Utility built
- [ ] Tests added

PR #3957, opened 2025-02-20
[Issue #3863] Save opportunity version

babebe:
## Summary
Fixes #{([3863](https://github.com/HHS/simpler-grants-gov/issues/3863))}

## Changes proposed
Created util func that saves an opportunity in the opportunity_version table
Create Test for util func


api/src/services/opportunities_v1/opportuntity_version.py
```diff
@@ -0,0 +1,26 @@
...
+    creates a new OpportunityVersion record with the data, and saves it in the database.
+
+    :param  db_session: The active SQLAlchemy session used to interact with the database.
+    :param opportunity: An instance of the Opportunity model containing the data to be saved.
+    :return: This function does not return a value. It saves a new version of the opportunity in the database.
+    """
+
+    #  Extracts the opportunity data as JSON object
+    opp_obj = opportunity.for_json()
+
```
chouinar, line 19: We need to use the Opportunity schema, not `for_json`. This method just turns the opportunity (and not other parts of the object) into a dictionary. It's useful for printing, but not enough for what we want here.

api/src/services/opportunities_v1/opportuntity_version.py
```diff
@@ -0,0 +1,26 @@
...
+    # Add new OpportunityVersion instance to the database session
+    opportunity_version = OpportunityVersion(
+        opportunity_id=opp_obj["opportunity_id"],
```
chouinar, line 22: Just because I prefer not indexing if possible.

```suggestion
        opportunity_id=opportunity.opportunity_id,
```

api/src/services/opportunities_v1/opportuntity_version.py
babebe: Done

api/src/services/opportunities_v1/opportuntity_version.py
babebe: Done

api/src/services/opportunities_v1/opportunity_version.py
```diff
@@ -0,0 +1,33 @@
...
+
+    # Add new OpportunityVersion instance to the database session
+    opportunity_version = OpportunityVersion(
+        opportunity_id=opportunity.opportunity_id,
+        opportunity_data=schema_data,
+    )
+
+    db_session.add(opportunity_version)
+
+    db_session.commit()
```
chouinar, line 33: Don't commit - we shouldn't include commits, especially in utility functions. Assume we're using this as part of an endpoint where we update an opportunity, the API endpoint itself controls when it's done with a transaction (our `with db_session.begin()` blocks). There might be additional DB processing after this and if you commit, those will fail as the transaction has ended.
```suggestion

```

api/tests/src/services/opportunities_v1/test_opportunity_version.py
```diff
@@ -0,0 +1,55 @@
...
+        no_current_summary=True,
+        category=OpportunityCategory.MANDATORY,
+    )
+
+    opp_ass = OpportunityAssistanceListingFactory.create(opportunity=opp)
+
+    expected = {
+        "agency": agency.agency_code,
+        "summary": None,
+        "category": OpportunityCategory.MANDATORY.lower(),
```
chouinar, line 28: Was this causing an issue without `lower()`? OpportunityCategory.MANDATORY should == "mandatory" since that's the value.

api/tests/src/services/opportunities_v1/test_opportunity_version.py
```diff
@@ -0,0 +1,55 @@
...
+    agency = AgencyFactory.create(top_level_agency=agency_top)
+
+    opp = OpportunityFactory.create(
+        opportunity_assistance_listings=[],
+        agency_code=agency.agency_code,
+        no_current_summary=True,
+        category=OpportunityCategory.MANDATORY,
+    )
+
+    opp_ass = OpportunityAssistanceListingFactory.create(opportunity=opp)
```
chouinar, line 23: If we want a slightly friendlier way to shorten it, I've seen "AL" for assistance listing
```suggestion
    opp_al = OpportunityAssistanceListingFactory.create(opportunity=opp)
```

api/src/services/opportunities_v1/opportunity_version.py
```diff
@@ -0,0 +1,33 @@
...
+
+    This function extracts the opportunity data from the provided Opportunity model instance,
+    creates a new OpportunityVersion record with the data, and saves it in the database.
+
+    :param  db_session: The active SQLAlchemy session used to interact with the database.
+    :param opportunity: An instance of the Opportunity model containing the data to be saved.
+    :return: This function does not return a value. It saves a new version of the opportunity in the database.
+    """
+    #  Extracts the opportunity data as JSON object
+    schema = OpportunityV1Schema()
```
chouinar, line 22: I know you're just following a pattern elsewhere, but we can define this once at the top of the file like `SCHEMA = OpportunityV1Schema()` so we don't need to reinitialize the schema every time we call the method. Negligible performance improvement, but good practice (that we should correct elsewhere).

babebe: Done

api/src/services/opportunities_v1/opportunity_version.py
babebe: Done

api/tests/src/services/opportunities_v1/test_opportunity_version.py
babebe: Actually not needed. Removed

api/tests/src/services/opportunities_v1/test_opportunity_version.py
babebe: Done

api/src/services/opportunities_v1/opportunity_version.py
babebe: Makes sense! Moved to top!

chouinar (approved):

PR #3965, opened 2025-02-21
Updating validation errors and logging

### LINKED ISSUE #3867: GitHub ETL - Handle validation failures
### Summary

Currently when an exported Github project item fails the validation step, we just log an ERROR to the console to avoid halting the rest of the ETL process. We should more gracefully handle these validation issues (e.g. dead letter queue, deterministic filters, etc.)

### Acceptance criteria

- [x] We're only logging errors to the console if there's an unexpected bug in the ETL
- [x] Expected validation failures are handled some other way

jcrichlake:
## Summary
Fixes #3867 

## Changes proposed
The validation code currently logs errors for items that fail.  We don't want to log errors for those items as they will condition developers to ignore errors that show up in the logs.  This change is to update the logging to make it more specific as to what the problem is and modify the logging level.


coilysiren (dismissed):

analytics/src/analytics/etl/github.py
```diff
@@ -274,7 +274,10 @@ def populate_issue_lookup_table(
         try:
             entry = IssueMetadata.model_validate(issue)
         except ValidationError as err:
-            logger.error("Error parsing row %d, skipped.", i)  # noqa: TRY400
+            logger.info(
+                "Error parsing row for issue lookup table %d, skipping.",
+                i,
+            )
             logger.debug("Error: %s", err)
```
DavidDudas-Intuitial, line 281: It might be more valuable to include `err` in the `logger.info` message rather than in a separate `logger.debug` message. Wdyt?

analytics/src/analytics/integrations/github/main.py
```diff
@@ -73,7 +73,10 @@ def transform_project_data(
             transformed_data.append(transformed)
 
         except ValidationError as err:
-            logger.error("Error parsing row %d, skipped.", i)  # noqa: TRY400
+            logger.info(
+                "**** Can't validate project data for row %d, skipping. ****",
+                i,
+            )
             logger.debug("Error: %s", err)
```
DavidDudas-Intuitial, line 80: Same here: can `err` be included in `logger.info` message?

analytics/src/analytics/etl/github.py
DavidDudas-Intuitial: I'm thinking specifically of the following troubleshooting scenario:
1. Daily ETL process fails during prod execution
2. Developer looks at AWS step function logs to ascertain cause and/or symptoms of failure
3. Generic "skipping" messages from `logger.info` offer no insight, so developer has to run locally and turn on debug level logging in order to troubleshoot

IMO it would be better if step 3 info messages contain more info to help in troubleshooting.

jcrichlake: I liked the idea so I tried it really quickly in local.  It really clutters up the console.   I think it could be easy for folks to miss new errors if they were inside that section of the logs.  But I'd like to hear your thoughts.


image

jcrichlake: I think I like the existing implementation.  As it is the logs stay clean while still alerting devs to omitted records.  It shouldn't be to hard for devs to change the log level up or down to get the full error if they want to dig into the issue.

widal001: I wonder if there's a compromise here: Something like including the field that the validation failed on? 

I don't know how easy it would be off the top of my head to parse the field title 

Looking at this debug-level logs is helpful because it points to a change we should actually make to the validation: Changing the type of [this `number` field](https://github.com/HHS/simpler-grants-gov/blob/a12b57406f0117ebb4142f7dbae4565811481908/analytics/src/analytics/integrations/github/validation.py#L113) to: `integer | float | None`

DavidDudas-Intuitial: A couple suggestions:
1. Maybe shorten the preamble of the logging statement, so instead of 
`logger.info("**** Can't validate project data for row %d, skipping. **** %s", i, err)` 
try 
`logger.info("Skipping project row %d: %s", i, err)`
2. It appears that these rows are failing because the validation is expecting an integer for points field but received a float (!) value. Can you please troubleshoot this? There's no good reason I can think of that a points field could/should contain a float value such as 3.1 or 5.1

jcrichlake: @DavidDudas-Intuitial I think that's a good call out! I can make that change.  @widal001 should we update the validation as part of this or keep it scoped as is for now?

DavidDudas-Intuitial: The fact that these rows have been failing for weeks evidently because of a `float` value in the `int` points field, and that we are just now discovering the cause, argues strongly in favor of making the logs more verbose!

coilysiren (approved):

analytics/src/analytics/integrations/github/validation.py
```diff
@@ -112,6 +112,16 @@ class NumberValue(BaseModel):
 
     number: int | None = None
 
+    @field_validator("number", mode="before")
+    def transform_float_to_int(cls, value) -> int | None:  # noqa: ANN001, N805
```
DavidDudas-Intuitial, line 116: @jcrichlake what is the N805 for? Seems unnecessary here

jcrichlake: It's to allow the first parameter to be named 'cls' instead of self.  I was trying to match the convention established by the other functions in this file

DavidDudas-Intuitial: Got it, thanks!

Issue #3897, opened 2025-02-14
Create a utility that given a two complex nested dictionaries can provide a set of changes between the two


chouinar:
### Summary

For example:

Given value1:
```json
{
    "name": "bob",
    "nested": {
        "sub_value1": [1, 2, 3, 4],
        "sub_value2": "hello"
        "sub_value3": "initial value"
    }
}
```

and value2:
```json
{
    "name": "steve",
    "nested": {
        "sub_value1": [1, 2, 5],
        "sub_value2": "hello"
        "sub_value3": "new value"
    },
    "new_field": "yay"
}
```

Might produce:
```json
[
  {
    "field": "name",
    "before": "bob",
    "after": "steve"
  },
  {
    "field": "nested.sub_value1",
    "before": [1, 2, 3, 4],
    "after": [1, 2, 5]
  },
  {
    "field": "nested.sub_value3",
    "before": "initial value",
    "after": "new value"
  },
  {
    "field": "new_field",
    "before": None,
    "after": "yay"
  },
]
```

A few rough notes:
* Put this util in our dict util: https://github.com/HHS/simpler-grants-gov/blob/main/api/src/util/dict_util.py
* The dict util already has a util for flattening a dictionary which might help to do before trying to do comparisions
* Figure out how it should behave when a key is only in the first or second dict, we should probably treat missing as null in either case?

### Acceptance criteria

- [ ] Util added
- [ ] Tests added

PR #3968, opened 2025-02-21
[Issue #3897] Get the diff of two nested dict

babebe:
## Summary
Fixes #{[3897](https://github.com/HHS/simpler-grants-gov/issues/3897)}

## Changes proposed
Util Function: Given two complex nested dictionaries can provide a set of changes between the two
Test Added


api/src/util/dict_util.py
```diff
@@ -40,3 +40,34 @@ def flatten_dict(in_dict: Any, separator: str = ".", prefix: str = "") -> dict:
...
+    diffs: list = []
+
+    all_keys = set(flatt_dict1.keys()).union(flatt_dict2.keys())  # Does not keep order
+
+    for k in all_keys:
+        v_a = flatt_dict1.get(k, None)
+        v_b = flatt_dict2.get(k, None)
+
+        # Store diff if values are different
+        if v_a != v_b:
```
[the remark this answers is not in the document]
babebe: Ah! Yes that would have been flagged as such. Added the check for lists

api/src/util/dict_util.py
```diff
@@ -40,3 +40,36 @@ def flatten_dict(in_dict: Any, separator: str = ".", prefix: str = "") -> dict:
...
+
+    all_keys = set(flatt_dict1.keys()).union(flatt_dict2.keys())  # Does not keep order
+
+    for k in all_keys:
+        v_a = flatt_dict1.get(k, None)
+        v_b = flatt_dict2.get(k, None)
+
+        # Store diff if values are different
+        if isinstance(v_a, list):
+            if set(v_a) != set(v_b):
```
chouinar, line 69: One scenario to account for - if `v_b` is `None` this would error since `set(None)` isn't valid.

I think the safest way to deal with that is making a utility function to handle if that errors and make it an empty set

babebe: Checking if iterable first would also do the trick?

babebe: created helper func to convert to set before comparing

api/src/util/dict_util.py
```diff
@@ -40,3 +40,43 @@ def flatten_dict(in_dict: Any, separator: str = ".", prefix: str = "") -> dict:
...
+
+    diffs: list = []
+
+    all_keys = set(flatt_dict1.keys()).union(flatt_dict2.keys())  # Does not keep order
+
+    for k in all_keys:
+        values = [flatt_dict1.get(k, None), flatt_dict2.get(k, None)]
+        # convert values to set for comparison
+        v_a = _convert_to_set(values[0])
+        v_b = _convert_to_set(values[1])
```
chouinar, line 67: Sorry if when I quickly said this earlier, I'd still want to check if these are iterables before converting to a set (ie. if they're both integers, leave them as integers). Otherwise we have to make sure everything we ever feed through this is hashable into a set.

How about this - adjust/rename the `_convert_to_set` method to have an implementation like:
```py
if isinstance(data, (list, tuple)): # not iterable - see note
   return set(data)

return data
```
While the iterable check seems decent, you found that strings already have a problem that you have to build in an exception for. But bytes, dicts, strings, and a few other types end up needing special handling - it's probably easier to go without it.

chouinar (approved):

PR #3992, opened 2025-02-24
[Issue #3308] View Saved Opportunities on Saved grants page

### LINKED ISSUE #3308: Implementation for viewing saved opportunities
### Summary

Implementation for #3307 

For user / saved grants page:

- If a user does NOT have any saved opportunities
  - Show the default page as per https://github.com/HHS/simpler-grants-gov/issues/3299
- If a user DOES have any saved opportunities
  - Heading remains the same as "Saved grants"
  - List of saved opportunities is displayed in style of existing search results
    - Includes display of total number of opportunities
    - Sorts in order saved from most recent descending
    - Display at full container width
- If the application errors trying to fetch saved opportunities
    - Do something TBD 

### Designs
https://www.figma.com/design/FGxWtAgToKhehLJCiuy1zL/Simpler.Grants.gov-Project?node-id=1988-11776&t=dbhBpeeRkVxuzFMp-0 

Image

### Acceptance criteria

- [ ] List of saved opportunities on the user page is live in prod

acouch:
## Summary
Fixes #3308

## Changes proposed
Add saved opps to Saved grants page:

### Desktop

image

### Mobile
image

### Desktop (no saved opps, no change)

image

### Mobile (no saved opps, no change)

image

## Context for reviewers
To test:

1. Start API locally
2. Sign in
3. Save an opportunity
4. View opportunity and "Saved grants" page


frontend/src/services/fetch/fetchers/savedOpportunityFetcher.ts
```diff
@@ -69,3 +70,22 @@ export const getSavedOpportunity = async (
   );
   return savedOpportunity ?? null;
 };
+
+export const fetchSavedOpportunities = async (): Promise<
+  SavedOpportunity[]
+> => {
+  try {
+    const session = await getSession();
```
doug-s-nava, line 78: It may be overkill, but since the session will only be available on the server side (until now it's only been used in route handlers) maybe we should put a "server only" on this file?

frontend/src/app/[locale]/saved-grants/page.tsx
```diff
@@ -17,9 +22,56 @@ export async function generateMetadata({ params }: LocalizedPageProps) {
...
 
+const SavedOpportunitiesList = ({
+  opportunities,
+}: {
+  opportunities: Opportunity[];
+}) => {
+  return (
+    <ul className="usa-prose usa-list--unstyled">
+      {opportunities.map((opportunity) => (
+        <li key={opportunity?.opportunity_id}>
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/saved-grants/page.tsx:33]: doesn't seem like this would probably be possible, but if we could have undefined opportunities coming through that could mean non-unique keys here. We could either just assert that the opportunity will always be defined or add a unique fallback value?

frontend/src/app/[locale]/saved-grants/page.tsx
```diff
@@ -17,9 +22,56 @@ export async function generateMetadata({ params }: LocalizedPageProps) {
...
+        </Link>
+      </div>
+    </>
+  );
+};
+
 export default async function SavedGrants({ params }: LocalizedPageProps) {
   const { locale } = await params;
   const t = await getTranslations({ locale });
+  const savedOpportunities = await fetchSavedOpportunities();
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/saved-grants/page.tsx:67]: [a note for later] is there a way we could move this complexity to the API side so that we don't have to make so many round trips here, or will caching make that more or less a moot point?

frontend/src/app/[locale]/saved-grants/page.tsx
```diff
@@ -17,9 +22,56 @@ export async function generateMetadata({ params }: LocalizedPageProps) {
...
+  const opportunities = savedOpportunities.map(async (savedOpportunity) => {
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/saved-grants/page.tsx:68]: something like "opportunityDetailPromises" would be clearer to me here

doug-s-nava (commented):
a couple of nit picks and one edge case that I'm not sure we want to cover here or handle later.

If you're viewing a search result that contains a saved opportunity, then you log out, the opportunity still shows up as saved. It gets removed on refresh, so maybe we want to just refresh the page on logout?

Screenshot 2025-02-24 at 11 42 39 AM

frontend/src/app/[locale]/saved-grants/page.tsx
Comment (acouch) [reply frontend/src/app/[locale]/saved-grants/page.tsx:68]: The `getOpportunityDetails` returns:

```javascript

interface OpportunityApiResponse extends APIResponse {
  data: Opportunity;
}
```

For me it makes sense to describe the type of data we are returning, vs the function used to retrieve it. Does `opportuniyPromises` make more sense?

Comment (doug-s-nava) [reply frontend/src/app/[locale]/saved-grants/page.tsx:68]: tldr; yep, `opportunityPromises` is great. 

I'm willing to go either way, since the `promise` piece of this is in the type, and explaining the type in the variable name is also generally a no-no, but in cases like this where a synchronous map function returns the promises, and then you await them later, I find it's often confusing at what point we're dealing with unresolved promises and at what point we're dealing with actual data. Somehow making it clear that this variable doesn't yet represent opportunities, but will in the future, seems helpful to me.

frontend/src/services/fetch/fetchers/savedOpportunityFetcher.ts
doug-s-nava: Agree there's no big risk here since this would break if we tried to use it in the client, but potentially would add some dev time if someone didn't realize that. If we pop "server-only" in here, since it's a utilities file rather than anything react specific, I think we'd gain the ability to break faster and provide an annotation for devs so that the use case for the file is immediately clear. https://nextjs.org/docs/app/building-your-application/rendering/composition-patterns#keeping-server-only-code-out-of-the-client-environment

Not entirely necessary, but worth maybe discussing further and thinking about documenting as we make a decision one way or another about how we want to use these designations.

acouch: I was conflating `server-only` and `use server`, the former poisoning clients from importing while the latter endorsing. I added `server-only`.

doug-s-nava (approved):
✅ ✅ ✅ ✅

acouch:
I created a separate ticket for the logout https://github.com/HHS/simpler-grants-gov/issues/4022

PR #3999, opened 2025-02-24
[Issue #3994] Fix agency data during transforms

### LINKED ISSUE #3994: Modify agency transform logic to unset top level agencies if null, and set test agencies based on prefix
### Summary

Modify the logic for setting the hierarchy to unset it if it’s no longer the parent (I don’t know how some of these got connected previously, agency codes can’t change since that’s the key we index all the logic on)

Modify the logic to set test agencies based on prefix (eg. we configured “GDIT” as a test agency, so anything that starts with that should be a test agency, not just “GDIT”).

### Acceptance criteria

- [ ] Changes made
- [ ] Tests added

chouinar:
## Summary
Fixes #3994

## Changes proposed
Modified the way we determine an agency is a test agency to be based on a prefix rather than exact match (eg. if we configure `ABC`, `ABC-123` would now be a test agency as well)

Modified the transform hierachy task to be able to unset top level agency if it doesn't match anything (fixes some weird data issues we're seeing)

Modified the transform hierachy to also update `is_test_agency` so that everything always gets reprocessed each run

## Context for reviewers
The transform agency logic happens in two passes, first we look at all of the agency changes from the `TGROUPS` table and apply updates / create new agencies.

We then have a `TransformAgencyHierarchy` task that runs to do some post-processing on all agencies. I mostly put these fixes in the latter bit so that any of the weird issues we've noticed get cleaned up.

### DIFF
--- a/api/src/data_migration/transformation/subtask/transform_agency.py
+++ b/api/src/data_migration/transformation/subtask/transform_agency.py
@@ -299,7 +299,7 @@ class TransformAgencyHierarchy(AbstractTransformSubTask):
             if top_level_agency_code and top_level_agency_code in agency_map:
                 agency.top_level_agency = agency_map[top_level_agency_code]
             else:
-                # We want to unset the top level agency is something is incorrectly
+                # We want to unset the top level agency if something is incorrectly
                 # pointed to (just as a safeguard)
                 agency.top_level_agency = None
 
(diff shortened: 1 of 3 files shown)


api/src/data_migration/transformation/subtask/transform_agency.py
```diff
@@ -275,17 +275,43 @@ def update_agency_download_file_types(
...
+
             top_level_agency_code = self.get_top_level_agency_code(agency.agency_code)
+            logger.info(
+                "Determined top level agency code for agency",
+                extra=log_extra | {"top_level_agency_code": top_level_agency_code},
+            )
             if top_level_agency_code and top_level_agency_code in agency_map:
                 agency.top_level_agency = agency_map[top_level_agency_code]
+            else:
+                # We want to unset the top level agency is something is incorrectly
```
freddieyebra, line 302:
```suggestion
                # We want to unset the top level agency if something is incorrectly
```

freddieyebra (commented):
minor comment correction. otherwise looks correct.

freddieyebra (approved):

PR #4020, opened 2025-02-26
[Issue #3936] Product roadmap page

### LINKED ISSUE #3936: Create a new page for the product roadmap
### Summary

This ticket is to convert the content currently on simpler.grants.gov/process to a new Product Roadmap page.

- "What's happening next" becomes "What we're working on right now"
- "Our progress so far" becomes "Recent milestones reached"
- Content is mostly the same with a few tweaks
- "Recent milestones reached" will have Quad 1 stuff for now. In the future, we'll move stuff from "What we're working on right now" to "Recent milestones reached" as we deliver
- New "How we work" section at the bottom
- "Process" tab in the header is renamed to "Roadmap", which points to this new page

### Acceptance criteria

- [ ] New roadmap page is live at simpler.grants.gov/roadmap
- [ ] "Process" is changed to "Roadmap" in the nav, pointing to /roadmap

freddieyebra:
## Summary
Fixes #3936

## Changes proposed
This PR implements a new product roadmap page at /roadmap based on this design here https://www.figma.com/design/FGxWtAgToKhehLJCiuy1zL/Simpler.Grants.gov-Project?node-id=3422-31812. 

## Context for reviewers
This PR only implements the new page at /roadmap and only implements the ability to land on this page by direct URL. Follow up work is in place in #4091  to implement replacing the existing /process page with the roadmap page.

To verify this change, run the frontend locally and navigate to localhost:3000/roadmap and ensure the content renders as expected.

## Additional information
Screen Recording 2025-03-10 at 10 40 43 AM


frontend/src/components/roadmap/roadmap.scss
andycochran, line 1: I think this should all be done with [USWDS Utility](https://designsystem.digital.gov/utilities/) classes in the markup.  No need for custom CSS.

frontend/src/components/roadmap/RoadmapPageSection.tsx
andycochran, line 1: Do we need this abstraction, or can this markup just be included in the sections. This might be brittle if e.g. the layout changes or the header hierarchy needs to change.

freddieyebra: Thanks for the review and that is a good point Andy. My thoughts behind this implementation was that all of the sections on the page have the same structure/layout aside from the header. If we wanted to update one the sections, we could do it in 1 place instead of every section.

Some options here are:
1. Leave this component as is and refactor if we get to a point where most section layouts are not consistent.
2. Add conditional logic so that if we do not get a `sectionTitle` property, then just `return {content}` as is.
3. Refactor an nuke this component all together.

Let me know which one you are leaning more towards!

frontend/src/components/roadmap/roadmap.scss
freddieyebra: Good call. I was running into some issues with setting the min/max styling on some of these components. They seemed to not be honoring/applying some of these properties. I will work on removing this stytlesheet.

freddieyebra: @andycochran just addressed this and removed the custom css.

doug-s-nava (commented):
nit picky notes from a brief manual test, none of these seem like deal breakers:
* header image seems to be cut off at some point in the figma file to make it more square. Is that an issue?
* The github logo is missing by the `View all deliverables on Github` link
* Text in the coding challenge pilot section reads `Collaborative Coding Challenge` in figma and `Collaborative Code & Design Challenge` in the local app.
* Font size in intro paragraph in `How we work` looks a little small

frontend/src/components/roadmap/sections/RoadmapWhatWereWorkingOn.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const { contentItems } = messages.Roadmap.sections.progress;
+
+  return (
+    <RoadmapPageSection
+      sectionTitle={t("title")}
+      extraClasses="bg-white"
+      sectionContent={
+        <>
+          <div className="margin-top-1" />
+          {contentItems.map((i, j) => (
```
doug-s-nava, line 18: for consistency, and other reasons that I probably have unreasonably strong opinions about, I'd rather stay away from abbreviated variable names like this. Can we generally use names that points to what the variables represent instead?

frontend/src/components/roadmap/sections/RoadmapHeader.tsx
```diff
@@ -0,0 +1,49 @@
...
+            >
+              {t("pageHeaderParagraph")}
+            </p>
+          </Grid>
+          <Grid
+            className="display-none tablet:display-flex tablet:flex-justify-center tablet:flex-align-center"
+            tablet={{ col: 6 }}
+          >
+            <div className="display-flex flex-justify-center flex-align-center">
+              <img
```
doug-s-nava, line 31: Is there value in using the next js `<Image />` here? https://nextjs.org/docs/pages/api-reference/components/image

frontend/src/components/roadmap/RoadmapPageSection.tsx
doug-s-nava: I can see the value of the abstraction here but I'd largely defer to Andy - if we don't expect consistency from the design and content on these pages it doesn't make a lot of sense to try and build a template to handle them. In the end I'm fine either way as long as we keep an eye out for anywhere that we're trying too hard to put a square peg in a round hole

freddieyebra:
> nit picky notes from a brief manual test, none of these seem like deal breakers:
> 
> * header image seems to be cut off at some point in the figma file to make it more square. Is that an issue?
> * The github logo is missing by the `View all deliverables on Github` link
> * Text in the coding challenge pilot section reads `Collaborative Coding Challenge` in figma and `Collaborative Code & Design Challenge` in the local app.
> * Font size in intro paragraph in `How we work` looks a little small

@doug-s-nava just addressed your cr feedback.

frontend/src/components/roadmap/RoadmapPageSection.tsx
andycochran: This just reminds me of the [ContentLayout](https://github.com/HHS/simpler-grants-gov/blob/main/frontend/src/components/ContentLayout.tsx) component, which we've discussed getting rid of.

acouch:
@andycochran Noting that the font-weight for the heading is `72px`, but there isn't a corresponding USWDS font size that is 72px. What is the correct font-size here? 
image

frontend/src/components/roadmap/RoadmapPageSection.tsx
freddieyebra: Thanks Andy. Aaron did mention that when I picked this up so I ensured I did not utilize that.

frontend/src/components/roadmap/sections/RoadmapHeader.tsx
acouch: Yes, esp since it sets the `srcset`. We haven't used content images before, so not sure what the right initial size and sizes should be. Will take a look.

freddieyebra:
> @andycochran Noting that the font-weight for the heading is `72px`, but there isn't a corresponding USWDS font size that is 72px. What is the correct font-size here? image

what is the scale of the font size for the uswds framework? From just looking/comparing font size, looks like px value is around 68.

andycochran:
> @andycochran Noting that the font-weight for the heading is `72px`, but there isn't a corresponding USWDS font size that is 72px. What is the correct font-size here? 

I don't think @michellemin-nava meant to prescribe a specific font size. We should use a default USWDS size. Likely the same that's on the homepage hero now: an `h1` with classes `text-ls-neg-2 tablet:font-sans-2xl desktop-lg:font-sans-3xl text-balance`

frontend/src/components/roadmap/RoadmapPageSection.tsx
freddieyebra: @andycochran let me know if this is a hard block, otherwise this is ready for re-review.

frontend/src/components/roadmap/sections/RoadmapProcess.tsx
```diff
@@ -0,0 +1,83 @@
...
+  return (
+    <RoadmapPageSection
+      sectionTitle={t("title")}
+      extraClasses="bg-white"
+      sectionContent={
+        <>
+          <p className="margin-0 tablet-lg:margin-bottom-2 line-height-sans-4 tablet:font-sans-md">
+            {t("sectionSummary")}
+          </p>
+          {roadmapProcesSectionGridRows.map((i, k) => (
```
doug-s-nava, line 49: can the iterator variable names get updated in this file well?

doug-s-nava (commented):
just one final nit from me

frontend/src/components/roadmap/RoadmapPageSection.tsx
```diff
@@ -0,0 +1,34 @@
...
+  sectionContent: React.ReactNode;
+  sectionTitle?: string;
+  extraClasses?: string;
+}
+
+export default function RoadmapPageSection({
+  sectionContent,
+  sectionTitle = "",
+  extraClasses = "",
+}: RoadmapPageSectionProps) {
```
acouch, line 14: A couple of recommendations here. To clean it up a little you could remove the `extraClasses` as a prop and just add `className` to  the implementations: `<RoadmapPageSection className="bg-white"/>`.  `sectionTitle` could be simplified as `title` as it conveys the same meaning with less. Finally the `sectionContent` could be changed to `{children}` which would allow wrapping whatever the implementation wants to pass as JSX.

frontend/src/components/roadmap/RoadmapPageSection.tsx
acouch: I'm counting 8 sections that have an identical structure in the upcoming design. Using a component saves us from:

```javascript
      <GridContainer className="display-flex flex-column padding-y-4 grid-container tablet-lg:flex-row tablet-lg:padding-y-6">
        <Grid row>
          <Grid tabletLg={{ col: 4 }}>
            <h2 className="margin-0 font-sans-lg tablet:font-sans-xl tablet-lg:maxw-card-lg">
              {title}
            </h2>
          </Grid>
          <Grid
            tabletLg={{ col: 8 }}
            className="margin-top-2 margin-bottom-0 tablet-lg:margin-0"
          >
            {content}
          </Grid>
        </Grid>
      </GridContainer>
```

for each one. An alternative to using a component could be a custom class (ie `simpler-container` and `simpler-title`):


```javascript
      <GridContainer className="simpler-container">
        <Grid row>
          <Grid tabletLg={{ col: 4 }}>
            <h2 className="simpler-title">
              {title}
            </h2>
          </Grid>
          <Grid
            tabletLg={{ col: 8 }}
          >
            {content}
          </Grid>
        </Grid>
      </GridContainer>
```

If I recall correctly, the issue with [ContentLayout](https://github.com/HHS/simpler-grants-gov/blob/main/frontend/src/components/ContentLayout.tsx) was complexity. The approach above seems less complicated:

```javascript
  bottomBorder?: "light" | "dark" | "none";
  children: React.ReactNode;
  gridGap?: true | "sm" | "md" | "lg" | "2px" | "05" | 1 | 2 | 3 | 4 | 5 | 6;
  paddingTop?: boolean;
  title?: string | React.ReactNode;
  titleSize?: "l" | "m";
```

@andycochran thoughts?

acouch:
Code looks great so far. Made one comment on the pageSection component. Also noting that the image isn't aligned like the mockup: 

image

and disappears on mobile.

Locally getting rid of the `display-none` in the grid row, changing the image classes to `minh-full maxw-full tablet:maxw-mobile-lg` and updating the column containing the image to ` tablet:flex-justify-end` fixed it for me.

frontend/src/components/roadmap/RoadmapSections.tsx
```diff
@@ -0,0 +1,17 @@
+"use client";
+
```
acouch, line 2:
```suggestion
```

One last nit 😄 but don't think we need this here.

freddieyebra:
> Code looks great so far. Made one comment on the pageSection component. Also noting that the image isn't aligned like the mockup:
> 
> image
> 
> and disappears on mobile.
> 
> Locally getting rid of the `display-none` in the grid row, changing the image classes to `minh-full maxw-full tablet:maxw-mobile-lg` and updating the column containing the image to ` tablet:flex-justify-end` fixed it for me.


@acouch  For this image, I intentionally did not render it on tablet since it looked a bit off as it looked to be a bit too stretched when rendered under title and sub title and too small in the same row. That was the purpose of having 2 separate images displayed/hidden in different screen sizes. I have updated the position of the image to be flex aligned at the end and removed the `"use client"` in component. Mind re-reviewing when you have a sec?

acouch:
@freddieyebra I didn't realize we are using two images. We can get away with using just one if the `<Gridcontainer /`> is replaced with a `<div className="tablet:grid-container"/> so that there isn't a border and applying the max width to the original image.

acouch:
Noticed that the text touches the image on some widths, this could be fixed by adding `padding-right-2` to the `<p>`
image

freddieyebra:
> @freddieyebra I didn't realize we are using two images. We can get away with using just one if the `<Gridcontainer /`> is replaced with a `<div className="tablet:grid-container"/>` so that there isn't a border and applying the max width to the original image.

Looks like there may be a few bugs/inconsistencies between using the class `"grid-container"` and the component `<GridContainer>` with the padding and margin. There are some discrepancies with the 2. I will need to refactor the rest of the sections in order to utilize the grid-container class to use a single image. Going to continue poking around to use a single image.

acouch (approved):
LGTM!

PR #4029, opened 2025-02-26
[Issue #3737] Refactoring to remove transforms and unify data names in pipeline

### LINKED ISSUE #3737: Update GraphQL queries to use aliases
### Summary

We are currently exporting the graphQL data as is from Github and transforming the column names in our python code to match what the database expects. We should instead use an alias in the GraphQL queries which would allow us to avoid transforms in the python code.  

### Acceptance criteria

Data columns from GraphQL are retrieved with the proper headings and the column name transformations are no longer needed.

jcrichlake:
## Summary
Fixes #3737


analytics/src/analytics/datasets/issues.py
```diff
@@ -37,11 +37,11 @@ class IssueMetadata(BaseModel):
     """Stores information about issue type and parent (if applicable)."""
 
     # Project metadata -- attributes about the sprint project board
-    project_owner: str
-    project_number: int
+    project_name: str
+    project_ghid: int
     # Issue metadata -- attributes about the issue common to both projects
     issue_title: str
-    issue_url: str
```
widal001, line 44: While most of this renaming I feel pretty neutral about, this one I'm a little hesitant to change because until it gets parsed by the transform and load step it's still a URL.

While that's not a show-stopper for me, it's worth noting that this content is still referred to as a URL [elsewhere in the ETL](https://github.com/HHS/simpler-grants-gov/blob/5ed0231b145c16cc2e96a99072dc3d40c4fbd1e4/analytics/src/analytics/etl/github.py#L288-L328)

widal001 (commented):
While I'm not inherently opposed to this change, I do have some questions:

1. The ticket that is associated with this PR says the goal is to shift aliasing to GraphQL, but all of these changes seem to be made to the python code. Is there a reason we shifted this PR to python updates instead of GraphQL?
2.  **Taking a step back:** What is the motivation for these changes? They don't _seem_ to be saving us transformations within the transform and load step (at least I don't see corresponding changes in that section of the code).
3. On a related note, can you help me understand why these changes to the IssueMetadata attributes in the extract step _don't_ require corresponding changes in the transform and load step?


TL;DR I'm okay with making these changes, I'm just not entirely sure what problem they're helping solve and I want to make sure that the changes made to the export aren't violating the contract with the transform and load step.

analytics/src/analytics/datasets/issues.py
jcrichlake: The answer to these:
1. Changing the aliases wouldn't actually change the data for most of the pipeline since the data shape is determined when it is put into a python object.  Instead I made these changes to pull the data field names that the db is expecting as far left as possible in the python code.
2. [Here ](https://github.com/HHS/simpler-grants-gov/pull/4029/files#:~:text=df.rename(columns=-,column_map) is where we were renaming the data
3. I believe because those steps aren't referencing any of the fields directly.

The problem I was attempting to solve by making these changes was to hopefully reduce confusion in the pipeline by having the fields remain the same throughout the process.  I thought that would be simpler for tracking down any issues related to a given field, instead of having to find the field in the database figure out if it got mapped to a different name at some point and then figuring under which of the two keys did that value get dropped/ incorrectly updated.

analytics/src/analytics/datasets/utils.py
```diff
@@ -67,9 +67,9 @@ def load_json_data_as_df_from_object(
     """
     # flatten the nested json into a dataframe
     df = pd.json_normalize(json_data)
-    # reorder and rename the columns
-    df = df[column_map.keys()]
-    df = df.rename(columns=column_map)
```
widal001, line 72: I'm not sure I love the idea of us removing this step from a general utility function, especially if the parameter for `column_map` still says it will rename them.

We should either change the function signature and docstrings, or restore the rename step.

widal001 (changes requested):
Thanks for clarifying the intent behind these column name changes.

If we're going to commit to stripping the `df.rename()` step from the `load_json_data_as_df_from_object()` then we should change the type and description of the `column_map` parameter to that function as well. Otherwise future users of that function may think it still supports renaming fields.

analytics/src/analytics/datasets/issues.py
widal001: Thanks for clarifying the intent.

While I appreciate the goal you described here:

> to hopefully reduce confusion in the pipeline by having the fields remain the same throughout the process.

I wonder if thinking about it this way makes the various steps of the ETL more tightly coupled than they're really meant to be. 

Again I feel pretty neutral about the name changes themselves, so I'm not opposed to merging this in. That being said, I want us to be careful about blurring the contract between steps in the ETL.

The `IssueMetadata` dataset provides a consistent contract with subsequent steps in a given ETL pipeline and you could imagine a future scenario in which the current pipeline:

```mermaid
flowchart LR
    gh[GitHub] --> etl1([make gh-data-extract]) --> issue[IssueMetadata] -->  etl2([make transform-and-load]) --> dw[Data warehouse]
```

Might support multiple targets in the future

```mermaid
flowchart LR
    gh[GitHub] --> etl1([make gh-data-extract]) --> issue[IssueMetadata] -->  etl2([make transform-and-load]) --> pg[GitHub tables in Postgres]
    issue --> etl3([make new-command]) --> target[Some other target]
```

analytics/src/analytics/datasets/utils.py
widal001: And if we're going to change the function signature then we should also remove this [`COLUMN_MAP`](https://github.com/HHS/simpler-grants-gov/blob/6416b063a10fecbff4a067eaa92d22bd01fec2c5/analytics/src/analytics/datasets/etl_dataset.py#L35-L63) class global so we don't have folks thinking the columns are getting renamed.

coilysiren (commented):
Defer to others!

PR #4032, opened 2025-02-26
[Issue #3828] Save search query unauth

### LINKED ISSUE #3828: Basic Saved Search Pane on Search Page for Unauthenticated Users
### Summary

All users should see a panel on the search page allowing them to copy the current page URL to the clipboard.

### Designs

https://www.figma.com/design/FGxWtAgToKhehLJCiuy1zL/Simpler.Grants.gov-Project?node-id=3436-17128&t=EDqa3pOlrQ3LrWpT-1

### Acceptance criteria

- [x] new pane is added to search page
- [x] pane contains a "copy url" button
- [x] button clicks copy the current page url to clipboard
- [x] if there is an error in search results, do not show the pane
- [x] create new snackbar component as per designs, by default displayed with dissolve transition in upper right corner of screen with automatic dissolve after 4 seconds
- [ ] create new "clear" custom button component. Can use Andy's suggestion for implementation below
- [x] tooltip shows specified messages for anonymous, users with saved searches, users with no saved searches

```
.usa-button--clear-mint {
  color: color("mint-50");
  background-color: transparent;

  &:hover,
  &.usa-button--hover {
    color: color("mint-60");
    background-color: color("mint-10");
  }

  &:active,
  &.usa-button--active {
    color: color("mint-70");
    background-color: color("mint-20");
  }
}
```

acouch:
## Summary
Fixes #3828

## Changes proposed

Adds "copy" button:

image

When clicked it shows:

image

it shows "copied" and a snackbar that are set for 4 seconds.

There is a "copying" state that should not be visible to the user ordinarily, but is included in case there is an issue with the navigation window.

## Testing locally

1. run npm dev (no need for API)
2. visit the search page
3. click "Copy" button, see search is pasted and snackbar appears

## Additional information
See various notes in code.


.devcontainer/devcontainer.json
```diff
@@ -1,20 +1,20 @@
...
+  
   "portsAttributes": {
-    "3000": {
+    "6006": {
       "label": "Application",
       "onAutoForward": "openPreview"
     }
   },
-  "forwardPorts": [3000]
+  "forwardPorts": [6006]
```
acouch, line 19: I updated this so we could share storybook branch versions for QA.

frontend/src/components/TooltipWrapper.tsx
```diff
@@ -0,0 +1,19 @@
+"use client";
```
acouch, line 1: This was added so that the USWDS Tooltip could be loaded without SSR which is necessary to described here https://github.com/uswds/uswds/issues/4338 in which the tooltip loads with different IDs.

frontend/src/components/Snackbar.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { ReactNode } from "react";
+
+type SnackbarProps = {
+  children: ReactNode;
+  isVisible: boolean;
+};
+
+const Snackbar = ({ children, isVisible = false }: SnackbarProps) => {
+  return (
+    <div
```
acouch, line 12: I did some research into what the correct accessibility markup would be for a "snackbar" message. USWDS does not have a snackbar for guidance.

[Material design](https://m3.material.io/components/snackbar/accessibility#127f5547-7068-4d5d-a6bd-ff97096d1139) suggests the following:

> * When a snackbar appears, announce the message but don't move focus.
> * Don't automatically move focus.
> * Don't trap focus in the snackbar. Users should be able to freely navigate in and out. 

[Clarity design](https://clarity.design/documentation/notifications) provides similar guidance.

So the [dialog role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role) and [alertdialog role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alertdialog_role) don't apply.

Noting that **both** advise that users should be able to dismiss snackbars, and Clarity suggests a 6 second timeout.

frontend/src/components/user/SavedSearchQueryUserControl.tsx
```diff
@@ -0,0 +1,32 @@
+"use client";
```
acouch, line 1: This file might be overkill for copy/paste, but would imagine using this for managing the user state and data fetching in the upcoming work.

frontend/src/hooks/useCopyToClipboard.ts
```diff
@@ -0,0 +1,27 @@
+"use client";
```
acouch, line 1: These two hooks are user driven, so not using `useEffect()` even though they touch outside systems.

frontend/tests/components/search/SavedSearchQuery.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+    mockedUseSnackBar.mockReturnValue({
+      ...mockSnackbarValues,
+    });
+    render(<SavedSearchQuery {...SavedSearchQueryProps} />);
+    expect(screen.getByText(snackbarMessage)).toBeInTheDocument();
+  });
+
+  // Doesn't work without adding support for dynamic imports
+  // eslint-disable-next-line jest/no-disabled-tests
+  it.skip("Shows tooltip", () => {
```
acouch, line 90: This isn't hard to add support for, but doesn't seem worth adding a new library https://stackoverflow.com/a/55444300

frontend/src/components/user/SavedSearchQueryUserControl.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { useTranslations } from "next-intl";
+import { usePathname, useSearchParams } from "next/navigation";
+
+import SavedSearchQuery from "src/components/search/SavedSearchQuery";
+
+export const SearchQuerySaveUserControl = () => {
+  const path = usePathname();
+  const searchParams = useSearchParams();
+  const query = searchParams?.toString() ? `?${searchParams.toString()}` : "";
+  const url = `${environment.NEXT_PUBLIC_BASE_URL}${path}${query}`;
```
acouch, line 14: This works in production mode locally, and should be correctly set in dev, but not 100% certain. Can update if not.

frontend/src/i18n/messages/en/index.ts
```diff
@@ -660,6 +660,20 @@ export const messages = {
...
+        authenticated: "Copy",
+      },
+      copying: "Copying",
+      copied: "Copied!",
+      help: {
+        unauthenticated:
+          "Use this set of search terms and filters often? Sign in to save this query to your account and use it again later. You can also copy and share the link to this query without signing in.",
+      },
+      snackbar:
+        "This search query was copied to your clipboard.<br></br> Paste it as a link anywhere.",
```
acouch, line 675: Using a line-break looked cleaner than setting a max-width for this message.

frontend/stories/components/ButtonwithIcon.stories.tsx
acouch, line 1: This was added by accident in another commit.

frontend/src/types/generalTypes.ts
```diff
@@ -8,3 +8,248 @@ export interface LayoutProps {
 export interface OptionalStringDict {
   [key: string]: string | undefined;
 }
+
+export type UswdsIconNames =
```
doug-s-nava, line 12: this is great, two things I'm thinking about:

* I've had a lot of luck recently defining these sort of lists / enums as const arrays that we then interpret into types, so that we can iterate over the list in code if we need to. I can't decide whether I think that'd be useful here - do you think we can be pretty confident that we'll only need this list at build time, or could it be useful for interpreting things at build time as well?
* there's a file in the uswds package that contains what I imagine is this same list (uswds/packages/usa-icon/src/usa-icon.json). I can't figure out whether or not there's a way to actually reference it, though. My one attempt at using `import * as usaIcon from "@uswds/uswds/packages/usa-icon/src/usa-icon.json";` doesn't work. Do you think that's worth exploring so we don't have to maintain the list?

frontend/tests/components/search/SavedSearchQuery.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+jest.mock("src/components/TooltipWrapper", () => {
+  return {
+    __esModule: true,
+    default: () => {
+      return <div>{copyText}</div>;
+    },
+  };
+});
+
+const mockedUseCopyToClipboard = useCopyToClipboard as jest.MockedFunction<
```
doug-s-nava, line 30: I'm confused about how these mocks are working. Would it work to set up the mocks and `jest.fn`s and reference them in the mocked modules, rather than importing the real modules?

doug-s-nava (commented):
leaving my initial thoughts on the code here for now in case I don't get around to manual testing before EOD

doug-s-nava (commented):
Functionally everything looks great!

The positioning on the tooltip looks slightly off center, and the content looks a little cramped, but that probably isn't a deal breaker.

Screenshot 2025-03-06 at 3 11 08 PM

acouch:
@doug-s-nava I updated the margin so the tooltip centers 
image

frontend/src/types/generalTypes.ts
acouch: I couldn't get that to work either. I can't imagine the names change very often.

frontend/src/hooks/useSnackbar.ts
```diff
@@ -0,0 +1,20 @@
...
+    setTimeout(() => {
+      setSnackbarIsVisible(false);
+    }, visibleTime);
+  };
+
+  const hideSnackbar = () => {
+    setSnackbarIsVisible(false);
+  };
+
+  return { hideSnackbar, snackbarIsVisible, showSnackbar, Snackbar };
```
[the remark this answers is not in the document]
acouch: So you can "use" it 😄 . Other hooks return the component they control.

frontend/src/components/search/SavedSearchQuery.tsx
```diff
@@ -0,0 +1,73 @@
...
+type SavedSearchQueryProps = {
+  copyText: string;
+  copyingText: string;
+  copiedText: string;
+  helpText: ReactNode;
+  url: string;
+  snackbarMessage: ReactNode;
+};
+
+const SavedSearchQuery = ({
```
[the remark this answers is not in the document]
acouch: I thought the next steps would use this as a container, I changed it so the save panel implements it and skips this, as it seems like you weren't going to use it.

frontend/src/components/USWDSIcon.tsx
```diff
@@ -18,6 +20,7 @@ export function USWDSIcon(props: IconProps) {
       aria-hidden="true"
       height={props.height}
       role="img"
+      onClick={props.onClick}
```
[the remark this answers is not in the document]
acouch: I removed it.

frontend/src/components/search/SavedSearchQuery.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { useCopyToClipboard } from "src/hooks/useCopyToClipboard";
+import { useSnackbar } from "src/hooks/useSnackbar";
+
+import dynamic from "next/dynamic";
+import { ReactNode } from "react";
+import { Button } from "@trussworks/react-uswds";
+
+import { USWDSIcon } from "src/components/USWDSIcon";
+
+const TooltipWrapper = dynamic(() => import("src/components/TooltipWrapper"), {
```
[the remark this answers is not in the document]
acouch: I think it still re-renders twice in dev, but there is not a discrepancy between server and client since it doesn't render on the server. `ssr: false,` is the key thing here I think.

frontend/src/components/user/SavedSearchQueryUserControl.tsx
acouch: I'd lean on using the Next functions first, but we could fall back on that.

frontend/src/components/user/SavedSearchQueryUserControl.tsx
acouch: I don't really care TBH, updated to `SaveSearchPanel`

frontend/src/components/Snackbar.tsx
```diff
@@ -0,0 +1,43 @@
...
+};
+
+const Snackbar = ({ children, close, isVisible = false }: SnackbarProps) => {
+  return (
+    <div
+      aria-hidden={!isVisible}
+      role="status"
+      data-testid="snackbar"
+      className={clsx(
+        "bg-base-darkest display-flex flex-align-start font-sans-2xs position-fixed padding-2 radius-md right-0 text-left text-white top-0 usa-modal-wrapper",
```
andycochran, line 21: > Using a line-break looked cleaner than setting a max-width for this message.

Is this where you tried a max width? Why wouldn't that look as nice as a line break? I fear relying on line breaks could be fragile, since it depends on the content for layout. And on very small screens the content could wrap to additional lines in an awkward way. 

Also wondering if we should use a slightly larger offset than `right-0` and `top-0`

frontend/src/styles/_uswds-theme-custom-styles.scss
```diff
@@ -252,6 +252,32 @@ i.e.
...
+    }
+  }
+  &--bottom {
+    &::after {
+      border-top-color: transparent;
+      border-bottom-color: color("secondary-darker");
+    }
+  }
+}
+
```
andycochran, line 280: Can't we just set the theme variables? 

These are available: 
```
$theme-tooltip-background-color
$theme-tooltip-font-color
$theme-tooltip-font-size
```

andycochran: It looks like [_usa-tooltip.scss](https://github.com/uswds/uswds/blob/681e5b2d6ce6df222ec034e800ee69af407d8998/packages/usa-tooltip/src/styles/_usa-tooltip.scss) uses that theme variable exactly as you've done.

frontend/src/types/generalTypes.ts
andycochran: If we wanted to get fancier, we could even configure USWDS to only compile the icons we're actually using. This and that config could pull from the same list. But IDK if it's necessary to shave off icon file size and compile time. 

FWIW, I think the USWDS icon names are pretty established. Would be rare to even change w/ major versions.

frontend/tests/components/search/SavedSearchQuery.test.tsx
doug-s-nava: Yeah, that's exactly the sort of implementation I was thinking of. If I throw this in, all the tests still pass for me locally. Were that not passing for you?

```suggestion
jest.mock("src/hooks/useCopyToClipboard", () => ({
  useCopyToClipboard: () => mockedUseCopyToClipboard(),
}));

jest.mock("src/hooks/useSnackbar", () => ({
  useSnackbar: () => mockedUseSnackBar(),
}));

const mockedUseCopyToClipboard = jest.fn();
const mockedUseSnackBar = jest.fn();
```

frontend/src/components/user/SavedSearchQueryUserControl.tsx
doug-s-nava: yeah, I'm fine with this as is but if it causes problems in dev I think using window.location is a good alternative

frontend/src/components/user/SavedSearchQueryUserControl.tsx
doug-s-nava: did you push? I'm not seeing the change

frontend/src/components/search/SavedSearchQuery.tsx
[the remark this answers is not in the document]
doug-s-nava: might make sense for the two of us to touch base on how to reconcile our PRs - I was thinking the `SavedSearchQueryUserControl` / `SaveSearchPanel` would be the container that would hold this and the other saved search UI elements

frontend/src/components/USWDSIcon.tsx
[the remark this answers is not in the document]
doug-s-nava: I'm still seeing it on my end

frontend/tests/components/search/SavedSearchQuery.test.tsx
acouch: When testing your suggestion, there is a linting error:

```
./tests/components/search/SearchQueryCopyButton.test.tsx
24:29  Error: Unsafe return of an `any` typed value.  @typescript-eslint/no-unsafe-return
27:22  Error: Unsafe return of an `any` typed value.  @typescript-eslint/no-unsafe-return
```

which is fixed by:

```javascript

const mockedUseCopyToClipboard = jest.fn() as jest.MockedFunction<
  typeof useCopyToClipboard
>;

```

which is I guess is slightly better because it only uses the functions as types, but they are still mocked so not sure what that gains us 🤷🏿 . I updated it as I had tested it locally anyway.

frontend/src/components/Snackbar.tsx
andycochran: Can we try these classes: `maxw-full top-0 right-0 tablet:maxw-mobile-lg tablet:top-2 tablet:right-2`? 

This would make it full width on small screens and limited to the top-right corner on larger screens. I think the line break you're using makes it about the width of `mobile-lg` (480px). 

We'll also need to enable the media queries for the top/right classes: 

```
$right-settings: ( responsive: true ), 
$top-settings: ( responsive: true ), 
```

If we ever want a snackbar that's not limited width, we can refactor this or pass it a prop to make it full-width/limited-width.

frontend/tests/components/search/SavedSearchQuery.test.tsx
acouch: @doug-s-nava additional to the linting, when I use your suggestion I get:

```
  ● Test suite failed to run

    ReferenceError: Cannot access 'mockedUseCopyToClipboard' before initialization

      21 |
      22 | jest.mock("src/hooks/useCopyToClipboard", () => ({
    > 23 |   useCopyToClipboard: mockedUseCopyToClipboard,
         |                       ^
      24 | }));
      25 |
```

frontend/src/components/Snackbar.tsx
acouch: I added those classes. 

> This would make it full width on small screens

Did you mean `width-full`? It isn't full width, just won't be larger than the screen on mobile:

image

frontend/src/i18n/messages/en/index.ts
acouch: @andycochran This was part of the design, which I guess I internalized. We'd save a lot of back and forth on these if the designs represented the desired output, thanks.

frontend/src/styles/_uswds-theme-custom-styles.scss
andycochran: The [custom button styles](https://github.com/HHS/simpler-grants-gov/blob/82eb518ccf56e053e8c6d917320f573a86a86649/frontend/src/styles/_uswds-theme-custom-styles.scss#L42-L233) are there because our button design can't be achieved with theme settings only (e.g. the bottom border). The button theme variables aren't granular enough and make assumptions of the color palette that don't align with the brand guidelines. But we do use what theme settings we can to style buttons (e.g. [radius](https://github.com/HHS/simpler-grants-gov/blob/82eb518ccf56e053e8c6d917320f573a86a86649/frontend/src/styles/_uswds-theme.scss#L200)). 

When implementing a design, we should first configure whatever theme setting we can to bring us closer to the design. If settings alone are not enough, we should then apply utility classes. If that is still not enough, then we get into writing CSS. 

I'd like to establish the norm that any custom CSS is a code smell. We can do almost anything we need by using USWDS utility classes. Definitely anything we _need_ to do, as we rarely need to stray from USWDS patterns. Style changes should be thought of in a particular order:
- Can I define a global USWDS system setting to achieve the goal?
- If not, can I apply USWDS utility classes?
- If not, can the CSS I write leverage USWDS mixins and functions?
- If not, can it at least use system tokens?

Our tooltip styles can be achieved w/ theme settings only.

andycochran: I should probably have justified in a code comment why all those button styles exist.

frontend/src/components/Snackbar.tsx
andycochran: Good catch! I was assuming the content would be long enough to force it to full width. You're right, it'll look better on small screens if it's full width regardless of content length.

frontend/src/i18n/messages/en/index.ts
andycochran: I think there might be an unintended/assumed design decision here. I doubt @snyder-ux meant to specify where the text should break to a new line. This is likely just a sketch, a coincidence that each sentence is on its own line, and not an explicit design decision.

image

^ This design could be interpreted to specify that the text should wrap at a certain width. Or that each sentence should be on its own line. If it's the later, I don't agree with that decision. These are not separate paragraphs. 

(If @snyder-ux did want each sentence on its own line that should be done with two paragraph tags. But let's not do that. Let's just make sure the toast itself is styled to work with a unknown amount of content.)

andycochran: 100% agree tho, this design should have been more specific about its intent.

doug-s-nava (approved):
✅ ✅ ✅ ✅

doug-s-nava (approved):
✅ ✅ ✅ ✅

frontend/src/i18n/messages/en/index.ts
snyder-ux: Text should naturally wrap based on the width of the container. Every sentence for this message is not a new paragraph.

Issue #3899, opened 2025-02-14
Create a task that adds a version to the opportunity version table after the transformations run


chouinar:
### Summary

This ticket requires most other opportunity versioning tickets to be complete first including:
* https://github.com/HHS/simpler-grants-gov/issues/3863
* https://github.com/HHS/simpler-grants-gov/issues/3897

The task will:
* Determine the last time it ran by looking at the JobLog table (if no runs, TODO - not sure we want to fetch everything or just default to like 24 hours?)
* fetch all opportunities where their last updated at timestamp in the OpportunityChangeAudit table is greater than the last time this task ran
* For each opportunity, check if there are any changes from the latest version in the opportunity version table (if no record in that table, assume changes)
    *  Note we do this check as a future-proof for when the opportunity update endpoint(s) will handle creating this record, don't want to duplicate.
* If there are changes, write a record to the opportunity version table using the utility.

This task should run as part of the [ECS task that does transforms](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/data_migration/command/load_transform.py), as a new 4th step after the set current opportunities task.

### Acceptance criteria

- [ ] Task created
- [ ] Thorough unit tests added for various scenarios

capriiis:
Finalizing testing, may rollover to Sprint 2.5

PR #4061, opened 2025-02-28
[Issue #3899] Task to Add new record into OpportunityVersion table

babebe:
## Summary
Fixes #{[3899](https://github.com/HHS/simpler-grants-gov/issues/3899)}

## Changes proposed
`StoreOpportunityVersionTask` has been implemented to call `saved_opportunity_version` whenever there’s an update in the `OpportunityChangeAudit` table for any opportunity since the last task run. Added as a 4th step to the `load_transform` job. 
The `saved_opportunity_version` utility has been updated to invoke the `diff_nested_dicts` function. If there’s a difference between the latest saved opportunity version and the current opportunity, a new record is created in the `OpportunityVersion` table.
Added `JobLogFactory`
Added tests.


api/src/task/opportunities/store_opportunity_version_task.py
```diff
@@ -0,0 +1,80 @@
...
+            select(OpportunityChangeAudit).where(OpportunityChangeAudit.updated_at > latest_time)
+        ).all()
+
+        for opp in updated_opportunities:
+            # Get Opportunity object
+            opportunity = self.db_session.execute(
+                select(Opportunity).where(Opportunity.opportunity_id == opp.opportunity_id)
+            ).scalar_one()
+            # Get Json
+            opportunity_v1 = SCHEMA.dump(opportunity)
```
chouinar, line 65: You made a utility for loading these records into that table, we should be using that as much as possible - I know it's used below, but re-converting something into a schema multiple times is a bit wasteful.

We'd likely need to add logic to that method of "only add if there is a diff", but we should keep that contained in one place.

api/src/task/opportunities/store_opportunity_version_task.py
```diff
@@ -0,0 +1,80 @@
...
+        super().__init__(db_session)
+
+    def run_task(self) -> None:
+
+        # Get latest job that run
+        latest_job = self.db_session.scalars(
+            select(JobLog)
+            .where(JobLog.job_type == self.cls_name())
+            .where(
+                or_(JobLog.job_status == JobStatus.COMPLETED, JobLog.job_status == JobStatus.FAILED)
```
chouinar, line 44: If the job failed, we probably don't want to count it.

api/src/task/opportunities/store_opportunity_version_task.py
```diff
@@ -0,0 +1,80 @@
...
+            )
+            .order_by(JobLog.created_at.desc())  # get the latest # add job_status
+        ).first()
+        # Get opportunity ids that were updated after the latest job run
+        latest_time = (
+            latest_job.created_at
+            if latest_job
+            else get_now_us_eastern_datetime() - timedelta(hours=24)
```
chouinar, line 52: DB values are stored in UTC, use `utcnow` whenever you do a comparison with now against something in the DB.

I think if there hasn't been a prior job, we should actually just grab everything (ie. pick a datetime of like `1970-01-01`) since we want to backfill these records anyways.

api/src/task/opportunities/store_opportunity_version_task.py
```diff
@@ -0,0 +1,80 @@
...
+        updated_opportunities = self.db_session.scalars(
+            select(OpportunityChangeAudit).where(OpportunityChangeAudit.updated_at > latest_time)
+        ).all()
+
+        for opp in updated_opportunities:
+            # Get Opportunity object
+            opportunity = self.db_session.execute(
+                select(Opportunity).where(Opportunity.opportunity_id == opp.opportunity_id)
+            ).scalar_one()
+            # Get Json
```
chouinar, line 64: No need to write a query when the relationship already is that:
```suggestion
        opportunity_change_audits = self.db_session.scalars(
            select(OpportunityChangeAudit).where(OpportunityChangeAudit.updated_at > latest_time)
        ).all()

        for opp_change_audit in opportunity_change_audits:
            opportunity = opp_change_audit.opportunity

```

api/src/task/opportunities/store_opportunity_version_task.py
```diff
@@ -0,0 +1,80 @@
...
+            opportunity_v1 = SCHEMA.dump(opportunity)
+
+            # Fetch latest version stored
+            latest_versioned_opp = self.db_session.execute(
+                select(OpportunityVersion).where(
+                    OpportunityVersion.opportunity_id == opp.opportunity_id
+                )
+            ).scalar_one_or_none()
```
chouinar, line 72: What happens if there are multiple versions?

api/src/task/opportunities/store_opportunity_version_task.py
babebe: Moved the logic to check for detecting diffs to save_opportunity_version function. And the task only checks if there is a change in the OpportunityChangeAudit table.

api/src/task/opportunities/store_opportunity_version_task.py
babebe: That's right! Updated!

api/src/task/opportunities/store_opportunity_version_task.py
babebe: Ordered by created_at desc

api/src/services/opportunities_v1/opportunity_version.py
```diff
@@ -11,23 +14,34 @@
...
-        opportunity_data=schema_data,
-    )
 
-    db_session.add(opportunity_version)
+    # Fetch latest opportunity version stored
+    latest_opp_version = db_session.execute(
+        select(OpportunityVersion)
+        .where(OpportunityVersion.opportunity_id == opportunity.opportunity_id)
+        .order_by(OpportunityVersion.created_at.desc())
+    ).scalar_one_or_none()
```
chouinar, line 33: One additional recommendation:
```suggestion
    latest_opp_version = db_session.execute(
        select(OpportunityVersion)
        .where(OpportunityVersion.opportunity_id == opportunity.opportunity_id)
        .order_by(OpportunityVersion.created_at.desc())
        .options(selectinload("*"))
    ).scalar_one_or_none()
```

Won't have any obvious affect on how the code works, but makes it so SQLAlchemy will select all parts of an opportunity in one set of queries rather than lazy-loading every relationship which each individually fire a select query off to the database. When we have a job that'll iterate over tens of thousands of records, it'll make it much faster (I tested adding it with set-current-opportunities and I want to say it was 70x faster?)

api/src/services/opportunities_v1/opportunity_version.py
```diff
@@ -11,23 +14,34 @@
...
+
+    diffs = diff_nested_dicts(opportunity_new, opportunity_existing)
+    if diffs:
+        # Add new OpportunityVersion instance to the database session
+        opportunity_version = OpportunityVersion(
+            opportunity_id=opportunity.opportunity_id,
+            opportunity_data=opportunity_new,
+        )
+
+        db_session.add(opportunity_version)
```
chouinar, line 47: Can put a small optimization/short circuit in the case where there is no existing since there will always be a diff.

Something like:
```suggestion
    diffs = {}
    if latest_opp_version:
        diffs = diff_nested_dicts(opportunity_new, latest_opp_version.opportunity_data)
        
    if diffs or latest_opp_version is None:
        # Add new OpportunityVersion instance to the database session
        opportunity_version = OpportunityVersion(
            opportunity_id=opportunity.opportunity_id,
            opportunity_data=opportunity_new,
        )

        db_session.add(opportunity_version)
```

api/src/task/opportunities/store_opportunity_version_task.py
```diff
@@ -0,0 +1,59 @@
...
+logger = logging.getLogger(__name__)
+
+
+@task_blueprint.cli.command(
+    "store-opportunity-version",
+    help="Store a new opportunity version if an opportunity has been updated",
+)
+@flask_db.with_db_session()
+def store_opportunity_version(db_session: db.Session) -> None:
+    StoreOpportunityVersionTask(db_session).run()
```
chouinar, line 25: We don't need a way to run this as a task directly, we want to add it as a 4th step of the existing transform jobs: https://github.com/HHS/simpler-grants-gov/blob/main/api/src/data_migration/command/load_transform.py

api/src/task/opportunities/store_opportunity_version_task.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+SCHEMA = OpportunityV1Schema()
+
+
+class StoreOpportunityVersionTask(Task):
+    def __init__(self, db_session: db.Session) -> None:
+        super().__init__(db_session)
```
chouinar, line 33: If we're not changing the init function, can exclude it and just use the one from task
```suggestion

```

api/src/task/opportunities/store_opportunity_version_task.py
```diff
@@ -0,0 +1,59 @@
...
+
+        updated_opportunities_change_audit = self.db_session.scalars(
+            select(OpportunityChangeAudit).where(OpportunityChangeAudit.updated_at > latest_time)
+        ).all()
+
+        for oca in updated_opportunities_change_audit:
+            # Get Opportunity object
+            opportunity = self.db_session.execute(
+                select(Opportunity).where(Opportunity.opportunity_id == oca.opportunity_id)
+            ).scalar_one()
```
chouinar, line 56: We don't need to query for the opportunity separately - the change audit object has a relationship to it. Also like I commented elsewhere, performance will be a lot better adding the selectinload bit.

```suggestion
        updated_opportunities_change_audit = self.db_session.scalars(
            select(OpportunityChangeAudit)
            .where(OpportunityChangeAudit.updated_at > latest_time)
            .options(selectinload("*"))
        ).all()

        for oca in updated_opportunities_change_audit:
            opportunity = oca.opportunity
```

api/src/task/opportunities/store_opportunity_version_task.py
```diff
@@ -0,0 +1,59 @@
...
+@task_blueprint.cli.command(
+    "store-opportunity-version",
+    help="Store a new opportunity version if an opportunity has been updated",
+)
+@flask_db.with_db_session()
+def store_opportunity_version(db_session: db.Session) -> None:
+    StoreOpportunityVersionTask(db_session).run()
+
+
+SCHEMA = OpportunityV1Schema()
```
chouinar, line 28: Don't think this is used here anymore
```suggestion

```

api/tests/src/db/models/factories.py
```diff
@@ -1907,4 +1913,16 @@ class Meta:
...
-    opportunity_data = factory.LazyAttribute(lambda o: o.opportunity_data)
+    opportunity_data = factory.LazyAttribute(lambda o: SCHEMA.dump(o.opportunity))
+
+
+class JobLogFactory(BaseFactory):
+    class Meta:
+        model = task_models.JobLog
+
+    job_id = Generators.UuidObj
+    job_type = factory.LazyAttribute(
```
chouinar, line 1924: This is the sort of field that if you want to use a factory for it, you should pass the value in yourself since a random value is going to be weird with tests (since you usually care a lot about the exact value).

api/tests/src/db/models/factories.py
```diff
@@ -1907,4 +1913,16 @@ class Meta:
...
+        lambda _: random.choice(["StoreOpportunityVersionTask", "SetCurrentOpportunitiesTask"])
+    )
+    job_status = factory.Iterator(JobStatus)
```
chouinar, line 1927: I'd probably say to just make this always the success value since most tests would only care about that (or want to override it to started/failed).

api/src/task/opportunities/store_opportunity_version_task.py
```diff
@@ -0,0 +1,59 @@
...
+    def __init__(self, db_session: db.Session) -> None:
+        super().__init__(db_session)
+
+    def run_task(self) -> None:
+
+        # Get latest job that run
+        latest_job = self.db_session.scalars(
+            select(JobLog)
+            .where(JobLog.job_type == self.cls_name())
+            .where(or_(JobLog.job_status == JobStatus.COMPLETED))
```
chouinar, line 41:
```suggestion
            .where(JobLog.job_status == JobStatus.COMPLETED)
```

api/src/services/opportunities_v1/opportunity_version.py
babebe: Yes indeed! Updated , thanks

api/src/task/opportunities/store_opportunity_version_task.py
babebe: @chouinar Will this be optional like the others defaulting to True

api/src/task/opportunities/store_opportunity_version_task.py
babebe: Removed

api/src/data_migration/command/load_transform.py
```diff
@@ -32,6 +33,9 @@
     "--insert-chunk-size", default=800, help="chunk size for load inserts", show_default=True
 )
 @click.option("--tables-to-load", "-t", help="table to load", multiple=True)
+@click.option(
+    "--store-version/--no-store-version", default=True, help="run StoreOpportunityVersionTask"
+)
```
chouinar, line 38: Let us default this to not being enabled, I think we should test the behavior out manually first to be safe.

api/tests/src/task/opportunities/test_store_opportunity_version_task.py
```diff
@@ -0,0 +1,101 @@
...
+        return StoreOpportunityVersionTask(db_session)
+
+    @pytest.fixture(autouse=True)
+    def clear_db(self, db_session):
+        opportunities = db_session.query(Opportunity).all()
+        for opp in opportunities:
+            db_session.delete(opp)
+
+        db_session.execute(delete(OpportunityVersion))
+        db_session.execute(delete(OpportunityChangeAudit))
```
chouinar, line 28: Minor nitpick, technically those two deletes do nothing because they had to have been deleted for the opportunity deletes to work (the `cascade="all, delete-orphan"` on the relationships tells SQLAlchemy to cleanup orphaned records if the opportunity itself is deleted).

```suggestion
    @pytest.fixture(autouse=True)
    def clear_db(self, db_session):
        opportunities = db_session.query(Opportunity).all()
        for opp in opportunities:
            db_session.delete(opp)
```

api/src/task/opportunities/store_opportunity_version_task.py
```diff
@@ -0,0 +1,36 @@
...
+
+        opportunity_change_audits = self.db_session.scalars(
+            select(OpportunityChangeAudit).where(OpportunityChangeAudit.updated_at > latest_time)
+        ).all()
+
+        for opp_change_audit in opportunity_change_audits:
+            opportunity = opp_change_audit.opportunity
+
+            # Store to OpportunityVersion table
+            save_opportunity_version(self.db_session, opportunity)
```
chouinar, line 36: Something I hadn't thought of until now, but has been briefly mentioned before - if an opportunity is a draft, we don't want to store anything for it yet in the version table.

I _think_ it would make more sense to put that check in the save_opportunity_version function so when we use it elsewhere, we don't need to implement that twice, but I'm not 100% on that behavior.

Should be just a simple if statement, but apologies for not mentioning it sooner.

api/src/util/dict_util.py
```diff
@@ -74,5 +74,7 @@ def diff_nested_dicts(dict1: dict, dict2: dict) -> list:
 
 def _convert_iterables_to_set(data: Any) -> Any:
     if isinstance(data, (list, tuple)):
+        if data and isinstance(data[0], (dict, list)):  # test with applicant typen
```
chouinar, line 77: Should this comment still be here?

api/tests/src/task/opportunities/test_store_opportunity_version_task.py
babebe: Thats right ! Deleted

api/src/task/opportunities/store_opportunity_version_task.py
babebe: NP, added check.

api/src/util/dict_util.py
babebe: Nope. Removed

chouinar (approved):

PR #4068, opened 2025-03-03
[Issue #3862] Remove is_deleted and revision_number from OpportunitySummary

### LINKED ISSUE #3862: Modify the opportunity summary transformation job to handle versioned records
### Summary

We want to populate our `opportunity_summary_history` table with records from `tsynopsis_hist` and `tforecast_hist` similar to how we already populate the non historical records.

Some rough notes:
* We _do not_ need there to 100% be an opportunity to import these historical records, fine to import with a non-existent opportunity (it likely means it was deleted) - our historical tables won't have foreign keys between them.
* The start timestamp for the version should be set based on the `create_ts` field
* The end timestamp should be `action_date` field
* Is deleted is based on the action type (D for delete)

If this is tricky/complex to integrate into the opportunity summary transformation job, feel free to make a new task that will run.

### Acceptance criteria

- [ ] Historical opportunity summaries imported
- [ ] Tests updated thoroughly

mikehgrantsgov:
## Summary
Fixes #3862

## Changes proposed
Remove `is_deleted` and `revision_number` from `OpportunitySummary`
Add migration
Remove code from factories and relevant tests

## Context for reviewers
Supports opportunity versioning work


api/src/db/models/opportunity_models.py
```diff
@@ -160,9 +160,7 @@ class OpportunitySummary(ApiSchemaTable, TimestampMixin):
 
     __table_args__ = (
         # nulls not distinct makes it so nulls work in the unique constraint
-        UniqueConstraint(
-            "is_forecast", "revision_number", "opportunity_id", postgresql_nulls_not_distinct=True
-        ),
+        UniqueConstraint("is_forecast", "opportunity_id", postgresql_nulls_not_distinct=True),
```
chouinar, line 163: We can remove the nulls_not_distinct bit, that was because revision_number could be null and needed it to make a compound key work right.

api/tests/src/data_migration/transformation/subtask/test_transform_funding_category.py
```diff
@@ -196,127 +183,15 @@ def test_process_funding_categories(self, db_session, transform_funding_category
         assert transform_constants.Metrics.TOTAL_ERROR_COUNT not in metrics
         assert metrics[transform_constants.Metrics.TOTAL_DELETE_ORPHANS_SKIPPED] == 1
 
-    def test_process_funding_categories_delete_and_inserts(
```
chouinar, line 199: Don't think you meant to do this - these are my recently added tests - guessing a merge conflict?

api/tests/src/data_migration/transformation/subtask/test_transform_funding_category.py
```diff
@@ -61,7 +61,7 @@ def test_process_funding_categories(self, db_session, transform_funding_category
         )
 
         opportunity_summary_forecast_hist = f.OpportunitySummaryFactory.create(
```
chouinar, line 63: If we were setting the revision number before, we should just delete these records since they're tests for historical scenarios.

api/tests/src/task/opportunities/test_set_current_opportunities_task.py
```diff
@@ -571,27 +485,11 @@ def test_via_cli(cli_runner, db_session, enable_factory_create):
     )
 
     # a basic forecasted scenario with several past revisions
-    container2 = (
```
mikehgrantsgov, line 574: @chouinar I'm not sure if this is correct. Maybe some of these tests can be removed?

api/tests/src/data_migration/transformation/subtask/test_transform_funding_category.py
mikehgrantsgov: Ah, that's right. Adding this back in

api/tests/src/data_migration/transformation/subtask/test_transform_funding_category.py
```diff
@@ -196,14 +169,39 @@ def test_process_funding_categories(self, db_session, transform_funding_category
...
+        assert delete_but_current_missing.transformation_notes == "orphaned_delete_record"
+
     def test_process_funding_categories_delete_and_inserts(
         self, db_session, transform_funding_category
     ):
         """Test that if we receive an insert and delete of the same lookup value
         in a single batch, we'll delete and then insert the record (effectively no meaningful change)
         """
         opportunity_summary_forecast = f.OpportunitySummaryFactory.create(
-            is_forecast=True, revision_number=None, no_link_values=True
```
mikehgrantsgov, line 206: @chouinar running this test I see `assert link_funding_category is not None` failing, but not sure why here. Something about the changes to `setup_funding_category`?

api/src/db/models/opportunity_models.py
mikehgrantsgov: Updated this, thanks

chouinar (approved):
LGTM!

PR #4081, opened 2025-03-07
[unticketed] Refactor opportunity types

doug-s-nava:
## Summary
unticketed

## Changes proposed
* refactors typing of opportunity data in order to address
  * general flexibility and reusability of opportunity types across different use cases dealing with the same data
  * newly introduced use case of search-like functionality and display of data in a non search context on the saved grants page

Note that this based off of the agency fetching Un-revert work, until that work is merged, as that work touches these types quite a bit

### Test steps
1. since this is just a typing change, we should be fine as long as CI passes, which means the application will build correctly


mdragon (approved):
👍🏻

frontend/src/app/[locale]/opportunity/[id]/page.tsx
```diff
@@ -40,7 +40,7 @@ export async function generateMetadata({
   let title = `${t("OpportunityListing.page_title")}`;
   try {
     const { data: opportunityData } = await getOpportunityDetails(id);
-    title = `${t("OpportunityListing.page_title")} - ${opportunityData.opportunity_title}`;
+    title = `${t("OpportunityListing.page_title")} - ${opportunityData.opportunity_title || ""}`;
```
Comment (freddieyebra) [inline frontend/src/app/[locale]/opportunity/[id]/page.tsx:43]: Not a blocker since the scope of this is pr is typing, but is this translating the title twice? Could this be: 
```suggestion
    title = `${title} - ${opportunityData.opportunity_title || ""}`;
```

Comment (doug-s-nava) [reply frontend/src/app/[locale]/opportunity/[id]/page.tsx:43]: discussed with Freddy, and this would probably be a more dry implementation but we'll leave as is for now:
```
let opportunityData;
  try {
    const { data } = await getOpportunityDetails(id);
    opportunityData = data;
  } catch (error) {
    console.error("Failed to render page title due to API error", error);
    if (parseErrorStatus(error as ApiRequestError) === 404) {
      return notFound();
    }
  }
  const pageTitle = t("OpportunityListing.page_title");
  const title = opportunityData?.opportunity_title
    ? `${pageTitle} - ${opportunityData.opportunity_title}`
    : pageTitle;
```

PR #4093, opened 2025-03-10
[Unticketed] Fix columns defined in analytics opp load

chouinar:
## Changes proposed
Remove expected columns from analytics opp data load that were recently removed

## Context for reviewers
We're seeing an error when loading data into analytics because we recently removed two columns (is_deleted and revision_number) from the opportunity_summary table. This causes the COPY command to fail as everything gets shifted over a few columns.

Removing those two columns from the config fixes the issues - tested by modifying the test file and seeing the same errors unless I made the configurational change:
```
E           psycopg.errors.InvalidTextRepresentation: invalid input syntax for type boolean: "545"
E           CONTEXT:  COPY opportunity_summary, line 4, column is_deleted: "545"
```


coilysiren (approved):
🙏🏼

PR #4073, opened 2025-03-06
[Issue #3829] add ability to save a search

doug-s-nava:
## Todo

- [ ] integrate with https://github.com/HHS/simpler-grants-gov/pull/4032 post merge
- [ ] ticket for e2e tests

## Summary
Fixes #3829

## Changes proposed
* adds save button to saved searches panel
* adds modal used for saving searches
* adds route to handle pushing saved search to API
* adds helper functions for saving search
* adds fixture file to hold common test data
* adds feature flag for toggling the availability of the save search functionality
* refactors some search helper functionality to make it more easily reusable 

### Test steps

1. start a server on this branch with `npm run dev`
2. visit a search page with the two necessary feature flags set to on: http://localhost:3000/search?_ff=authOn:true;savedSearchesOn:true
3. _VERIFY_: save button appears in save searches panel
4. add filters, query and sorting option
5. click "save" button to save search
6. _VERIFY_: modal appears
7. click save button in modal
8. _VERIFY_: validation error appears, no API request made
9. enter a name and click save
10. _VERIFY_: loading state is displayed
11. _VERIFY_: success state is displayed
12. run something like `SELECT * FROM api.user_saved_search WHERE name='<the name you used to save your search>' against your local db
13. _VERIFY_: the saved search looks as expected in the DB
14. force an error in the code by altering the subpath on line 19 of savedSearchFetcher.ts
15. re-run steps 5 - 9
16. _VERIFY_: error message appears

PR #4095, opened 2025-03-10
[Issue #3828] Update copy button to support non-https

### LINKED ISSUE #3828: Basic Saved Search Pane on Search Page for Unauthenticated Users

acouch:
## Summary
Fixes #3828

## Changes proposed
The clipboard property of the Navigator interface [requires https](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/clipboard) or localhost, which was not discovered until testing in dev.

This PR fixes that, throws an error in the clipboard component so the tooltip is not shown if it fails, and updates the content URL.

### Testing locally

1. Run `npm run dev` on `main`. 

```bash
   ▲ Next.js 15.1.4
   - Local:        http://localhost:3000
   - Network:      http://192.168.1.66:3000
   - Environments: .env.local, .env.development
```  
Use the `Network` URL, go to /search and see that the copy function errors.

2. Checkout this branch. Repeat the process, see that it works.

### DIFF
--- a/frontend/src/components/search/SearchQueryCopyButton.tsx
+++ b/frontend/src/components/search/SearchQueryCopyButton.tsx
@@ -14,7 +14,7 @@ const TooltipWrapper = dynamic(() => import("src/components/TooltipWrapper"), {
   loading: () => <USWDSIcon className="margin-left-1" name="info_outline" />,
 });
 
-const SNACKBAR_VISIBLE_TIME = 60000;
+const SNACKBAR_VISIBLE_TIME = 6000;
 
 type SearchQueryCopyButtonProps = {
   copyText: string;
@@ -49,7 +49,7 @@ const SearchQueryCopyButton = ({
               showSnackbar(SNACKBAR_VISIBLE_TIME);
             })
             .catch((error) => {
-              console.error("Error copying to clipboard:", error);
+              console.error(error);
             });
         }}
       >
--- a/frontend/src/components/user/SaveSearchPanel.tsx
+++ b/frontend/src/components/user/SaveSearchPanel.tsx
@@ -2,15 +2,20 @@
 
 import { useTranslations } from "next-intl";
 import { usePathname, useSearchParams } from "next/navigation";
+import { useMemo } from "react";
 
 import SearchQueryCopyButton from "src/components/search/SearchQueryCopyButton";
 
 export const SaveSearchPanel = () => {
   const path = usePathname();
   const searchParams = useSearchParams();
-  const query = searchParams?.toString() ? `?${searchParams.toString()}` : "";
-  const origin = typeof window !== "undefined" ? window.location.origin : "";
-  const url = `${origin}${path}${query}`;
+
+  const url = useMemo(() => {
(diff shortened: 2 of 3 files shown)


frontend/src/components/search/SearchQueryCopyButton.tsx
```diff
@@ -14,7 +14,7 @@ const TooltipWrapper = dynamic(() => import("src/components/TooltipWrapper"), {
   loading: () => <USWDSIcon className="margin-left-1" name="info_outline" />,
 });
 
-const SNACKBAR_VISIBLE_TIME = 60000;
+const SNACKBAR_VISIBLE_TIME = 6000;
```
acouch, line 17: Off by order of 1

frontend/src/components/user/SaveSearchPanel.tsx
```diff
@@ -11,7 +9,8 @@ export const SaveSearchPanel = () => {
   const path = usePathname();
   const searchParams = useSearchParams();
   const query = searchParams?.toString() ? `?${searchParams.toString()}` : "";
-  const url = `${environment.NEXT_PUBLIC_BASE_URL}${path}${query}`;
+  const origin = typeof window !== "undefined" ? window.location.origin : "";
```
doug-s-nava, line 12: my gut is telling me we should memoize this but it may not matter. I ended up memoizing in my branch though, fwiw https://github.com/HHS/simpler-grants-gov/pull/4073/files#diff-00a0460a1f9ea20f824eeb39f592f53c68adfa957f1d04d64959829f9700ccecR40

doug-s-nava (approved):
working for me locally! let's get it to DEV and confirm

✅ ✅ ✅ ✅

frontend/src/components/user/SaveSearchPanel.tsx
acouch: Thanks for the suggestion. I spent way too much time thinking about this, and don't think the calculation of `.toString()` and grabbing the window object are worth caching, but prefer to have consistency so updated.

doug-s-nava (approved):

Issue #4014, opened 2025-02-25
Make incremental job safe to run more often with smaller batch sizes, pseudo-continuous


mxk0:
### Summary

When running the incremental search data load, the job should be changed in the following ways:
* Rather than fetch all unprocessed recently-updated opportunities, process them all, and commit - it should instead process in batches, committing between each batch. We should process a configurable (via an env var) number of records per batch (default to 1000 for now).
* When fetching a batch, we should join with the opportunity table, and sort by the `created_at` timestamp descending - this way we are always processing "newer" opportunities first. If we ever have to process an enormous backlog of opportunities, we'll prefer those that were created more recently / those that appear towards the top of search results.
* For the batch processing, if the total time elapsed since batch processing began is more than 50 minutes, the job should not process another batch and end. This is done because we run the job hourly and don't want two jobs running on top of each other if we have a large backlog of data to load into the index.

Note that for writing tests around the "50 minutes elapsed - stop processing" logic, the freezegun library we use to set a specific time can be used to change the current time: https://github.com/spulec/freezegun?tab=readme-ov-file#moving-time-to-specify-datetime

Also for testing, feel free to configure the batch size down to a more reasonable number (like 5 per batch), don't try to write a test with 1000 records

Leads to #4017 

### Acceptance criteria

* [ ] Listed changes made
* [ ] Tests updated to validate behavior of each

PR #4151, opened 2025-03-11
[Issue #4014] Batch process incremental job

babebe:
## Summary
Fixes #{[4014](https://github.com/HHS/simpler-grants-gov/issues/4014)}

Process in incremental load in batches, committing between each batch. 
Join with the opportunity table, and sort by the created_at timestamp descending and opportunity_id desc - this way we are always processing "newer" opportunities first
For the batch processing, if the total time elapsed since batch processing began is more than 50 minutes, the job exits


api/local.env
```diff
@@ -138,6 +138,10 @@ PUBLIC_FILES_OPPORTUNITY_DATA_EXTRACTS_PATH=/tmp
 # File path for the create-analytics-db-csvs task
 API_ANALYTICS_DB_EXTRACTS_PATH=/tmp
 
+# Batch processing configuration for load_opportunities_to_index task
+INCREMENTAL_LOAD_BATCH_SIZE=1000
+INCREMENTAL_LOAD_MAX_PROCESS_TIME=3000
```
chouinar, line 143: Instead of env vars in local.env, set these as defaults  in the config class - we want these to be the defaults in all environments, if we just put it here, you'll need to make terraform changes as well.

api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -134,56 +139,93 @@ def _handle_incremental_upserts(self, existing_opportunity_ids: set[int]) -> Non
...
+            if last_processed_opportunity:
+                query = query.where(
+                    or_(
+                        Opportunity.created_at < last_processed_opportunity.created_at,
+                        and_(
+                            Opportunity.created_at == last_processed_opportunity.created_at,
+                            Opportunity.opportunity_id < last_processed_opportunity.opportunity_id,
+                        ),
+                    )
+                )
```
chouinar, line 175: I'm not following what this is doing - if we already processed an opportunity, it should have some flag set on it in the OpportunityChangeAudit table that means we won't pick it up again on the next batch. We shouldn't be querying/sorting by opportunity_id at all, only the created_at timestamp because we want to process newer opportunities first due to them being at the top of search.

api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -134,56 +139,93 @@ def _handle_incremental_upserts(self, existing_opportunity_ids: set[int]) -> Non
...
-        # Fetch opportunities that need processing from the queue
-        query = (
-            select(Opportunity)
-            .join(OpportunityChangeAudit)
-            .join(CurrentOpportunitySummary)
-            .where(
-                Opportunity.is_draft.is_(False),
-                CurrentOpportunitySummary.opportunity_status.isnot(None),
+        last_processed_opportunity: Opportunity | None = None
+        while True:
```
chouinar, line 143: There is a lot going on in this single loop, could we move some of the implementation into functions so it's easier to parse? Maybe something like a function for building the query?

api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -134,56 +139,93 @@ def _handle_incremental_upserts(self, existing_opportunity_ids: set[int]) -> Non
...
+                logger.info(f"Indexed {len(loaded_ids)} opportunities")
+
+                # Update updated_at timestamp instead of deleting records
+                self.db_session.execute(
+                    update(OpportunityChangeAudit)
+                    .where(OpportunityChangeAudit.opportunity_id.in_(processed_opportunity_ids))
+                    .values(updated_at=datetime_util.utcnow())
+                )
+
+                self.db_session.commit()
```
chouinar, line 224: If at all possible, we should avoid directly using `commit`. Always prefer:

```py
with db_session.begin():
     # stuff
# when the block ends, commit is called
```

api/tests/src/search/backend/test_load_opportunities_to_index.py
```diff
@@ -196,12 +199,16 @@ def test_opportunity_attachment_pipeline(
 
 class TestLoadOpportunitiesToIndexPartialRefresh(BaseTestClass):
     @pytest.fixture(scope="class")
-    def load_opportunities_to_index(self, db_session, search_client, opportunity_index_alias):
+    def load_opportunities_to_index(
+        self, db_session, search_client, opportunity_index_alias, monkeypatch_class
+    ):
+        monkeypatch_class.setenv("BATCH_SIZE", "5")
```
chouinar, line 205: Don't need to monkeypatch, just pass the value into the config we're already overriding values on the next line. You gave an alias to the fields, so Pydantic expects you to do `INCREMENTAL_LOAD_BATCH_SIZE=5` to the init function to set `incremental_load_batch_size`.

api/src/search/backend/load_opportunities_to_index.py
babebe: We do need additional check to pick up the next batch of "opportunities" within this job. I am not certain OpportunityChangeAudit's updated_at column is safe to use, if this is the only job that updates this field then i could use it ?

api/local.env
babebe: Yes that makes sense

api/tests/src/search/backend/test_load_opportunities_to_index.py
babebe: Reading from config

api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -120,71 +124,74 @@ def incremental_updates_and_deletes(self) -> None:
...
+                self.db_session.execute(self._build_opportunities_to_process_query())
+                .scalars()
+                .all()
             )
 
-            # Add to index batch if it's indexable
-            opportunities_to_index.append(opportunity)
-            processed_opportunity_ids.add(opportunity.opportunity_id)
+            if len(queued_opportunities) == 0:
+                break
```
chouinar, line 162: Could we add a message sorta like this?
```suggestion
            if len(queued_opportunities) == 0:
                logger.info("No opportunities remain in queue, exiting processing")
                break
```

chouinar (commented):
One very minor logging suggestion, LGTM otherwise

chouinar (approved):

PR #4156, opened 2025-03-12
[Issue 4084] Create acceptance criteria dataset

### LINKED ISSUE #4084: [ETL:Export] Extend export output to include deliverable acceptance criteria
### Summary

This task is to be done after the roadmap export step of the ETL workflow has been extended to include deliverable bodyContent which contains acceptance criteria (see #4083).

In the export step of the ETL workflow, the `GitHubProjectETL` class exposes the method `extract_and_transfrom_in_memory` which is responsible for getting deliverable data from GitHub and storing it in an internal data structure of type `GitHubIssues`. See etl/github.py.

This task is to extend the functionality and output of `extract_and_transfrom_in_memory` as follows:

1. extract acceptance criteria data from roadmap_json
2. store acceptance criteria data in a new data structure to encapsulate description and state (boolean, i.e. is the checkbox checked)
3. return the data to caller

This task also includes integration of a markup parser to do the heavy lifting of parsing acceptance criteria from the bodyContent data (see https://github.com/HHS/simpler-grants-gov/issues/4082).

### Acceptance criteria

- [ ] Deliverable acceptance criteria are available for each deliverable in instances of GitHubIssues or in another new data structure

DavidDudas-Intuitial:
## Summary
Fixes #4084 

[WIP] Scaffolding for extracting acceptance criteria from GitHub bodyContent values.

WIP

### DIFF
--- a/analytics/src/analytics/datasets/acceptance_criteria.py
+++ b/analytics/src/analytics/datasets/acceptance_criteria.py
@@ -1,0 +1,88 @@
+"""Implement the AcceptanceCriteriaDataset class."""
+
+from typing import Self
+
+import pandas as pd
+
+from analytics.datasets.base import BaseDataset
+from analytics.datasets.utils import load_json_data_as_df_from_object
+
+
+class AcceptanceCriteriaDataset(BaseDataset):
+    """Encapsulate data exported from GitHub."""
+
+    COLUMN_MAP = {
+        "issue_url": "ghid",
+        "issue_bodycontent": "bodycontent",
+    }
+
+    @classmethod
+    def load_from_json_object(cls, json_data: list) -> Self:
+        """
+        Instantiate an instance of AcceptanceCriteriaDataset from a JSON object.
+
+        Parameters
+        ----------
+        json_data: list
+            In-memory JSON object containing data exported from GitHub.
+
+        Returns
+        -------
+        Self:
+            An instance of the AcceptanceCriteriaDataset dataset class.
+
+        """
+        # Load input dataset
+        df = load_json_data_as_df_from_object(
+            json_data=json_data,
+            column_map=cls.COLUMN_MAP,
+            date_cols=None,
+        )
+
+        # Transform entity ID columns
+        df = cls.transform_entity_id_columns(df)
+
+        return cls(df)
+
+    @classmethod
+    def transform_entity_id_columns(cls, df: pd.DataFrame) -> pd.DataFrame:
+        """Remove FQDN from URLs in the 'ghid' column."""
+        prefix = "https://github.com/"
+        for col in ("ghid",):  # Ensure correct iteration
+            if col in df.columns:
+                df[col] = df[col].str.replace(prefix, "", regex=False)
+
+        return df
+
+    def get_totals(self, ghid: str) -> dict:
+        """Get the total number of acceptance criteria and the total number done."""
+        # Ensure required columns exist
+        if "ghid" not in self.df.columns or "bodycontent" not in self.df.columns:
+            return {"total_criteria": 0, "total_done": 0}
+
+        result = self.df[self.df["ghid"] == ghid]
+
+        if result.empty:
+            return {"total_criteria": 0, "total_done": 0}
+
+        # Safely extract bodycontent
+        bodycontent = result["bodycontent"].get(result.first_valid_index(), "")
+
+        # Parse the acceptance criteria
+        criteria = self._parse_body_content(bodycontent)
+
+        total_criteria = len(criteria)
+        total_done = sum(1 for criterion in criteria if criterion["is_done"])
+
+        return {"total_criteria": total_criteria, "total_done": total_done}
+
+    def _parse_body_content(self, bodycontent: str) -> list[dict]:
+        """Parse bodycontent into structured acceptance criteria."""
+        if not isinstance(bodycontent, str) or not bodycontent.strip():
+            return []
+
+        # Placeholder for real parsing logic, returning hardcoded criteria
+        return [
+            {"title": "the box is blue", "is_done": False},
+            {"title": "the customer is happy", "is_done": False},
+        ]
--- a/analytics/src/analytics/etl/github.py
+++ b/analytics/src/analytics/etl/github.py
@@ -9,6 +9,8 @@ from pathlib import Path
 import pandas as pd
 from pydantic import BaseModel, ValidationError
 
+from analytics.datasets.acceptance_criteria import AcceptanceCriteriaDataset
+
 from analytics.datasets.issues import (
     GitHubIssues,
     IssueMetadata,
@@ -16,6 +18,7 @@ from analytics.datasets.issues import (
 )
 from analytics.datasets.utils import load_json_file
 from analytics.integrations import github
+
 
 logger = logging.getLogger(__name__)
 
@@ -205,6 +208,13 @@ class GitHubProjectETL:
             quad_field=roadmap.quad_field,
             pillar_field=roadmap.pillar_field,
         )
+
+        # extract acceptance criteria from roadmap json
+        deliverable_json = [d for d in roadmap_json if d["issue_type"] == "Deliverable"]
+        _ = AcceptanceCriteriaDataset.load_from_json_object(
+            json_data=deliverable_json,
+        )
+        # TO DO: include a.c. in function output
 
         # export sprint data
         issues = []


DavidDudas-Intuitial:
@widal001 @jcrichlake I think we can leverage existing `roadmap_json` in `extract_and_transform_in_memory()` by adding key/value pair `issue_bodycontent`. If the roadmap json is extended to contain body content for each deliverable, then the new `AcceptanceCriteriaDataset` class in this PR will encapsulate the parsing, and can be used by the follow on load step in the ETL workflow.

analytics/src/analytics/etl/github.py
```diff
@@ -206,6 +207,13 @@ def extract_and_transform_in_memory(self) -> list[dict]:
             pillar_field=roadmap.pillar_field,
         )
```
jcrichlake, line 209: I think the data needs to be bubbled up from inside of the `export_roadmap_data_to_object` function.  I tested this some yesterday and my working functions weren't populating when just using the roadmap_json. But I'll explore this more this morning

DavidDudas-Intuitial: @jcrichlake I prefer to not parse the body content during export. Instead, just include it (unparsed) as a value in the roadmap json returned by `export_roadmap_data_to_object`. Then we do not need to change the return type of `export_roadmap_data_to_object`. That method will continue to return the roadmap json but the json will contain a new key/value pair (e.g. `issue_bodycontent`) which will be parsed later during the load step.

roadmap-data

DavidDudas-Intuitial:
@jcrichlake see my latest [commit](https://github.com/HHS/simpler-grants-gov/pull/4156/commits/895d4d4acdf074d50636b0aa434c196493c15cfd) in which I add issue body to existing roadmap query and data structure without need for new query, transformation, validation, or wrapper.

DavidDudas-Intuitial:
After latest commit, roadmap data now contains raw body content for each deliverable:

roadmap-data-with-acceptance-critera

As a next step, we need to add @jcrichlake's checkbox parsing logic to `AcceptanceCriteriaDataset._parse_body_content()`

DavidDudas-Intuitial:
Latest [commit](https://github.com/HHS/simpler-grants-gov/pull/4156/commits/ce62c2430888ac94e87b17fd28d2e7a485e95fe1) writes acceptance criteria totals to data warehouse, completing the end-to-end journey of the data from GitHub.

analytics/src/analytics/datasets/acceptance_criteria.py
```diff
@@ -0,0 +1,184 @@
...
+
+        # init counters
+        total_criteria = 0
+        total_done = 0
+
+        # define regex to capture markdown sections:
+        # - captures section titles that start with "### "
+        # - captures section content until the next "###", EOF, or multiple blank lines
+        section_regex = r"^###\s+(.*?)\n([\s\S]*?)(?=\n###|\Z|\n{2,})"
+        sections = re.findall(section_regex, bodycontent, re.MULTILINE)
```
widal001, line 121: I think this regex will miss the required content in at least one, maybe two scenarios:

- A given section (e.g. Acceptance Criteria) has two successive line breaks before it gets to a new header (see the Acceptance criteria section in [this example](https://regexr.com/8d52b)) -- this is likely to happen in at least a few cases.
- Acceptance Criteria or Metrics is the last section in the issue content and there is one return before the end of string (see [this example](https://regexr.com/8d530)) -- This might not be a concern with the `strip()` call earlier but I'd love to have at least one test that specifically checks for this edge case.


Is there a reason we're not simply splitting on the headers like in [this function](https://github.com/HHS/grants-product-and-delivery/blob/358abdeebbf967869a55ed1c21f4e7d7ccf48910/automations/product_strategy_metrics/calculate_metrics.py#L23-L41)?

analytics/tests/datasets/test_acceptance_criteria.py
```diff
@@ -0,0 +1,171 @@
...
+    """Test get_totals with a valid issue URL."""
+    totals = acceptance_criteria_dataset.get_totals(
+        "org/repo/issues/123",
+        AcceptanceCriteriaType.MAIN,
+        AcceptanceCriteriaNestLevel.ALL,
+    )
+
+    assert isinstance(totals, AcceptanceCriteriaTotal)
+    assert totals.criteria == 6
+    assert totals.done == 3
```
widal001, line 129: I'm glad that we're testing for at least a few cases here!

Since so much work is being offloaded to the parser method, it might be good to pull that out to a utility function that is simply invoked by `self._parse_body_content()` That would allow us to test it separately and more easily add explicit test scenarios for known edge cases as they arise.

widal001 (commented):
Smart approach! And everything runs end to end as expected (see screenshot)

I flagged one specific scenario that I think will be a problem with the current regex to parse the sections. And left a request about testing the parser function separately so we can mock out those different scenarios.

Screenshot 2025-03-14 at 11 21 07 AM

analytics/src/analytics/datasets/acceptance_criteria.py
```diff
@@ -0,0 +1,184 @@
...
+
+        # iterate sections
+        for section_name, section_body in sections:
+
+            # determine whether this section should be included in totals
+            if (
+                ac_type != AcceptanceCriteriaType.ALL
+                and section_name.lstrip("#").strip() != ac_type.value
+            ):
+                continue
```
widal001, line 138: Oh also, this is probably fine for now, but the current code would count checkboxes in _any_ section (including those outside "Metrics" and "Acceptance criteria")

I _do think_ Lucas wants the count of checkboxes in both Acceptance Criteria and Metrics, but I'm not sure they want checkboxes outside those sections included (if they exist).

Maybe instead of a single enum value, we could check for a list of desired sections, like I have in [this function](https://github.com/HHS/grants-product-and-delivery/blob/358abdeebbf967869a55ed1c21f4e7d7ccf48910/automations/product_strategy_metrics/calculate_metrics.py#L44-L61). It could still be a list of enums.

Not a must have change right now, though, we just might want to anticipate that as potential feedback.

jcrichlake: I would discuss that change, what if say we put in checkboxes for pre-requisites that need to be in place before we start something.  I don't think those checkboxes which would be needed for starting this story should count toward the progress of it

analytics/src/analytics/integrations/etldb/main.py
```diff
@@ -124,13 +135,27 @@ def sync_data(dataset: EtlDataset, effective: str) -> None:
...
+        ac_total = (
+            ac.get_totals(
+                ghid,
+                AcceptanceCriteriaType.ALL,
+                AcceptanceCriteriaNestLevel.ALL,
+            )
+            if ac is not None
+            else AcceptanceCriteriaTotal()
+        )
+        result[ghid], _ = model.sync_deliverable(deliverable_df, ghid_map, ac_total)
```
widal001, line 158: @DavidDudas-Intuitial this looks like it's counting all sections.

analytics/src/analytics/datasets/acceptance_criteria.py
widal001: Is there a reason that we'd want to stop capture after hitting multiple new lines at all? I feel like just limiting it to either the next section or the end of the string should suffice, assuming scenario two isn't an issue with proper stripping.

analytics/src/analytics/datasets/acceptance_criteria.py
DavidDudas-Intuitial: You're right! Good catch. I'll fix it. The intention is too include BOTH sections not ALL sections

analytics/src/analytics/datasets/acceptance_criteria.py
DavidDudas-Intuitial: Good call. I'll iterate on this to improve it.

DavidDudas-Intuitial: Thanks for pointing out the much simpler approach of splitting by section header.  I added it (see next commit) and it works great while simplifying the code.

analytics/src/analytics/datasets/acceptance_criteria.py
DavidDudas-Intuitial: Fixed in next commit.

analytics/src/analytics/integrations/etldb/main.py
DavidDudas-Intuitial: Fixed

analytics/tests/datasets/test_acceptance_criteria.py
DavidDudas-Intuitial: The `AcceptanceCriteriaDataset` class exists for the purpose of encapsulating the logic of parsing, so I prefer to keep the logic in the private method in the class instead of a utility function outside the class.

widal001: Fair enough, can we at least add more explicit tests for different scenarios?

I'd like to keep each test relatively focused on a specific edge case (e.g. checkboxes in sections other than "Acceptance Criteria" and "Metrics") so we are explicit about the desired behavior rather than just adding lots of different scenarios to a single mock input dataset.

widal001 (approved):
@DavidDudas-Intuitial thanks for addressing my core concerns!

I'd love to see us add a few more explicit test case scenarios around parsing to document the expected behavior around certain key edge cases, but don't let that hold up the merge.

Re-ran the pipeline and things looked good as far as I could tell!

Screenshot 2025-03-14 at 3 39 59 PM

DavidDudas-Intuitial:
@widal001 I added 8 more tests for various edge cases

widal001 (approved):
LGTM! Thanks for adding some more tests

PR #4162, opened 2025-03-12
[ISSUE #4014] Update opportunity change audit [1/2]

babebe:
## Summary
Fixes #{4014}

## Changes proposed
Migration File: Add new col `loaded` to OpportunityChangeAudit. [Nullable, Bool]
Update Factory


api/src/db/models/opportunity_models.py
```diff
@@ -456,6 +456,7 @@ class OpportunityChangeAudit(ApiSchemaTable, TimestampMixin):
         BigInteger, ForeignKey(Opportunity.opportunity_id), primary_key=True, index=True
     )
     opportunity: Mapped[Opportunity] = relationship(Opportunity)
+    loaded: Mapped[bool | None]
```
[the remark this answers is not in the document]
babebe: Added to migration.

api/src/db/migrations/constants.py
```diff
@@ -0,0 +1,31 @@
...
+        WHEN 'current_opportunity_summary' THEN
+            opp_id := NEW.opportunity_id;
+        ELSE
+            opp_id := NEW.opportunity_id;
+    END CASE;
+
+    INSERT INTO api.opportunity_change_audit (opportunity_id)
+    VALUES (opp_id)
+    ON CONFLICT (opportunity_id)
+    DO UPDATE SET is_loaded_to_search = False;
```
chouinar, line 26: We still want to set updated_at when doing this - even if it's not used for any process, still useful from an auditing perspective.

api/src/db/migrations/constants.py
```diff
@@ -0,0 +1,31 @@
+create_trigger_function = """
```
chouinar, line 1: Could we name this something more specific in case we add more triggers?

chouinar (approved):
LGTM!

PR #4166, opened 2025-03-12
[no ticket] Adds New Relic Infrastructure Monitoring

coilysiren:
## Summary

Adds baseline new relic configuration. This doesn't solve any of our specific tickets, but is does configure the baseline AWS integration.

## Context

This was harder to setup than it looks! I had to do some fancy stuff with imports and deletes.


mdragon (approved):
LGTM

PR #4170, opened 2025-03-13
[Issue #3830] apply saved searches

### LINKED ISSUE #3830: Apply saved searches
### Summary

Allow users to apply any of their saved searches.

Application will fetch the list of the user's saved searches, translated them into URLs as necessary, and display them in alphabetical order by their display name in a select. Default option for the select should be blank.

When a saved search is selected from the select, the current URL will be replaced with the URL for the saved search, without the need for clicking a submit button.

Select should expand to the length of the longest saved name and scroll to show all saved searches.

If a saved search is chosen, and any changes are made to search terms, the select should return to its default blank state.

Appearance of this functionality will be controlled by the feature flag introduced in #3829 

Depends on https://github.com/HHS/simpler-grants-gov/issues/3829

### Designs
https://www.figma.com/design/FGxWtAgToKhehLJCiuy1zL/Simpler.Grants.gov-Project?node-id=3436-17763&t=EDqa3pOlrQ3LrWpT-1

### Acceptance criteria

- [ ] Select appears in saved searches pane for all authenticated users with at least one saved search
- [ ] Select is populated with all of the user's saved searches
- [ ] Default selected option is blank
- [ ] Saved search options are sorted alphabetically by display name
- [ ] Selecting a search applies the filters, query and sort order associated with the saved search
- [ ] Functionality is controlled by saved search feature flag
- [ ] error messaging on failure to fetch saved searches TBD

doug-s-nava:
## Summary
Fixes #3830

## Changes proposed

For detailed UI and content requirements see figma here https://www.figma.com/design/FGxWtAgToKhehLJCiuy1zL/Simpler.Grants.gov-Project?node-id=1809-14763&p=f&t=wdo6SF0jtnMNp7Uu-0

### Main changes

* creates a SaveSearchSelector component to handle displaying and applying users' saved searches from the search page
* updates to related to components to support the selector's functionality

### Secondary changes
* updates saved search API route response payload to include id of new search
* new helper functions for search format translation, query strings, and search param hook
* tightening up some typing

### Test steps

1. start a server on this branch with `npm run dev`, and restart or start the API server to pull in changes there as well
2. visit search with feature flags on http://localhost:3000/search?_ff=authOn:true;savedSearchesOn:true
3. log in
4. _VERIFY_: no select appears
5. save a search
6. _VERIFY_: on save, the select appears and the name of your saved search appears as selected in the select
7. alter your search by selecting or unselecting a filter or entering a search term
8. _VERIFY_: the select updates to its default display, no longer displaying your saved search name
9. select your saved search from the select
10. _VERIFY_: page and url update as expected to reflect your saved search
11. _VERIFY_: name of your selected search appears in the select

### On successful save
Screenshot 2025-03-14 at 12 53 19 PM

### With options available
Screenshot 2025-03-14 at 12 53 36 PM

### Error state
Screenshot 2025-03-14 at 12 44 18 PM

### Mobile

Screenshot 2025-03-14 at 12 54 21 PM


api/src/api/users/user_routes.py
```diff
@@ -282,7 +282,7 @@ def user_save_search(
         },
     )
 
-    return response.ApiResponse(message="Success")
+    return response.ApiResponse(message="Success", data=saved_search)
```
doug-s-nava, line 285: @chouinar would appreciate a look at this, as I imagine it will break some tests, and there may be a better way of going about getting the id down in the payload

frontend/src/components/SimplerAlert.tsx
```diff
@@ -5,7 +5,8 @@ import { ReactNode } from "react";
 import { USWDSIcon } from "./USWDSIcon";
 
 interface SimplerAlertProps {
-  alertClick: () => void;
+  // in some cases we want the message to be persistent
```
doug-s-nava, line 8: designs show an unclosable alert, so made this adjustment

frontend/src/types/search/searchRequestTypes.ts
```diff
@@ -12,7 +14,7 @@ export type PaginationOrderBy =
   | "opportunity_number"
   | "opportunity_title"
   | "agency_name"
-  | "top_level_agency_code"
+  | "top_level_agency_name"
```
doug-s-nava, line 17: seems to be a change made in the data that was not updated in the frontend typing

frontend/src/types/search/searchRequestTypes.ts
```diff
@@ -59,27 +61,37 @@ export type SortOption = {
...
+// used for now in the process of performing a search request. To be deprecated.
+export interface QueryParamData extends FilterQueryParamData {
+  page: number;
   sortby: SortOptions | null;
+  query?: string | null;
   actionType?: SearchFetcherActionType;
   fieldChanged?: string;
 }
 
-export interface SearchParamsTypes {
```
doug-s-nava, line 75: since everything here is optional this type wasn't doing very much for us

api/src/api/users/user_routes.py
chouinar: Could we spin this change off into a separate PR? There's a lot going on here, and I want to make sure the API change gets the right attention / is implemented for what you are actually looking for

doug-s-nava: sure

doug-s-nava: https://github.com/HHS/simpler-grants-gov/pull/4196

frontend/src/app/api/user/saved-searches/list/route.ts
```diff
@@ -0,0 +1,35 @@
+import { readError, UnauthorizedError } from "src/errors";
```
acouch, line 1: Noting it is nice we have a consistent API route format 🤓

frontend/src/components/search/SaveSearchModal.tsx
```diff
@@ -85,7 +83,10 @@ function SuccessContent({
   );
 }
 
-export function SaveSearchModal() {
+// note that this is client component, but not marked as "use client" because to do so
+// would cause an error when passing the function prop from parent client component see:
+// https://github.com/vercel/next.js/discussions/46795
```
acouch, line 88: Interesting, makes sense, thanks for the note.

Themoreyouknow

frontend/src/components/search/SaveSearchSelector.tsx
```diff
@@ -0,0 +1,156 @@
...
+  useEffect(() => {
+    if (newSavedSearches.length && fetchSavedSearches) {
+      fetchSavedSearches()
+        .then(() => {
+          // set the latest saved search as selected
+          setSelectedSavedSearch(newSavedSearches[0]);
+        })
+        .catch(handleFetchError);
+    }
+  }, [handleFetchError, fetchSavedSearches, newSavedSearches]);
```
acouch, line 99: Couldn't the first one here be removed, and the`if(newSavedSearches.length...` be added to the `.then()` step? It would still fetch the searches on page load and would update the selected saved search if `newSavedSearches `data changes.

frontend/src/components/search/SaveSearchSelector.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+  // reset saved search selector on search change
+  useEffect(() => {
```
acouch, line 102: My reading of the docs suggests not using an effect for state changes: [*If there is no external system involved (for example, if you want to update a component’s state when some props or state change), you shouldn’t need an Effect.*](https://react.dev/learn/you-might-not-need-an-effect). It looks like the `useEffect()` could be removed.

frontend/src/components/search/SaveSearchSelector.tsx
doug-s-nava: It's a good thought. I'll play around and see if I can effectively combine these two useEffects

frontend/src/components/search/SaveSearchSelector.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+  const [selectedSavedSearch, setSelectedSavedSearch] = useState<string>();
+  const [loading, setLoading] = useState<boolean>(false);
+  const [apiError, setApiError] = useState<Error | null>();
+
+  // allows us to avoid resetting the select option when params change after selection
+  const [applyingSavedSearch, setApplyingSavedSearch] =
+    useState<boolean>(false);
+
+  const handleFetchError = useCallback((e: Error) => {
```
acouch, line 44: My reading / thinking of `useCallback()` is for expensive calculations or functions that will be passed down as props so a child doesn't re-render. The docs say there is [no harm in using them](https://react.dev/reference/react/useCallback#should-you-add-usecallback-everywhere). 

For me caching state updates and fetches feels like an over-optimization, but most of all want to be consistent about it, so if we adopt here, would like to have a follow-up to update [OpportunitySaveUserControl.tsx](https://github.com/HHS/simpler-grants-gov/blob/main/frontend/src/components/user/OpportunitySaveUserControl.tsx) to be consistent.

doug-s-nava: thanks for bringing this up. In this case we can just get rid of the function definition altogether, since it really only needs to get used as part of the fetch function. If this was a case where the function was needed across multiple other memoized pieces of functionality the useCallback would be necessary though, otherwise you end up in a situation with infinite rerenders.

frontend/src/components/search/SaveSearchSelector.tsx
acouch: OK, maybe something to keep our eye on as we go. We could refactor later if we come up with a different solution.

doug-s-nava:
@acouch yeah there are a couple of things I've noticed while working on this ticket that I didn't get right on the modal PR - also need to update the success modal state. I'll create a ticket for the follow up.

acouch (approved):

PR #4180, opened 2025-03-13
[Issue #4178] Create a rough utility for building JsonSchema

### LINKED ISSUE #4178: Create a utility for generating JsonSchema
### Summary

As we look to build out an approach to at least partially automate some of the building of JsonSchema objects, we should start with a rough approach for creating those via a Python script.

We don't need to account for all possibilities with JsonSchema, just setup a rough foundation for future work to build on (ie. parsing XML and using these tools).

### Acceptance criteria

- [ ] Utility created

chouinar:
## Summary
Fixes #4178

## Changes proposed
Created a rough utility for building JsonSchema

## Context for reviewers
The idea here is that this will later get used by a process that parses the PDFs/XML of the forms and uses this builder to convert to JsonSchema. This does not include all terms supported by JsonSchema, I mostly aimed at getting all the types + validation rules setup. Additional fields (like titles, descriptions, comments) aren't included here yet.

Additional context from research - I looked into a tool that just directly converts XSD to Json and while a few exist, they largely don't appear to work - the only one I got sorta working was a plugin to PyCharm, and it only vaguely worked. Most of them were written 10 years ago and not maintained, and JsonSchema has gone through several revisions in the past few years.

I also looked into libraries like this, and came across https://github.com/wolverdude/GenSON - but it doesn't include a lot of the JsonSchema components. It seems to be focused on generating JsonSchema from an answer rather than building it programmatically like I was looking for.

### DIFF
--- a/api/poetry.lock
+++ b/api/poetry.lock
@@ -80,6 +80,46 @@ tests = ["apispec[marshmallow,yaml]", "openapi-spec-validator (==0.7.1)", "pytes
 yaml = ["PyYAML (>=3.10)"]
 
 [[package]]
+name = "arrow"
+version = "1.3.0"
+description = "Better dates & times for Python"
+optional = false
+python-versions = ">=3.8"
+groups = ["main"]
+files = [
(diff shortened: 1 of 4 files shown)


mdragon (approved):
Approved

PR #4185, opened 2025-03-13
[Issue #4169] Create `POST` `/applications/start`

### LINKED ISSUE #4169: Create POST /applications/start endpoint
### Summary

Auth: ApiTokenAuth (not user)

### Request
```json
{
  "competition_id": "<some UUID>"
}
```

### Response
```json
{
 "data": {
   "application_id": "<some UUID>"
 }
}
```

### Behavior
Very basic - it will create an application in the application table pointing to the competition passed in.

Only meaningful error conditions:
* Invalid auth token (same as search and a bunch of other endpoints)
* Competition doesn't exist (do something like `raise_flask_error(404, ...)`)

### NOTE
These endpoints should _ONLY_ be enabled in dev/staging, assuming this is the first one we do, we want to set these up like the auth endpoints and only connect the blueprint to the app based on an env var (see app.py).

### Acceptance criteria

- [ ] Endpoint created (likely need to make a new blueprint for these apply endpoints)
- [ ] Tests added
- [ ] Schema created for request/response

mikehgrantsgov:
## Summary
Fixes #4169

## Changes proposed
Create POST `application/start` route, creating an application given a competition ID.
Add tests.

## Context for reviewers
Very basic - it creates an application in the application table pointing to the competition passed in.

Only meaningful error conditions:
- Invalid auth token (same as search and a bunch of other endpoints)
- Competition doesn't exist (do something like raise_flask_error(404, ...))

## Additional information
See attached unit tests.


api/src/api/application_alpha/application_route.py
```diff
@@ -0,0 +1,33 @@
...
+    ApplicationStartRequestSchema,
+    ApplicationStartResponseSchema,
+)
+from src.auth.api_key_auth import api_key_auth
+from src.services.applications.create_application import create_application
+
+logger = logging.getLogger(__name__)
+
+
+@application_blueprint.post("/application_alpha/start")
```
chouinar, line 17: For the path, we want `/applications/...` - the alpha/versioning bit is already added by the blueprint itself.

api/src/api/application_alpha/application_schemas.py
```diff
@@ -0,0 +1,10 @@
+from src.api.schemas.extension import Schema, fields
+
+
+class ApplicationStartRequestSchema(Schema):
+    competition_id = fields.UUID(required=True)
+
+
+class ApplicationStartResponseSchema(Schema):
+    message = fields.String()
+    data = fields.Dict(keys=fields.String(), values=fields.String())
```
chouinar, line 10: I think we need to leave keys/values out as it'll run validation that those are the types and we will have any possible type here. For example, what happens if you pass this in right now?
```
{
 "my_field": {
     "a": 1,
     "b": [
        {
          "c": "hello"
        }
      ]
  }
}
```

api/src/api/application_alpha/application_schemas.py
```diff
@@ -0,0 +1,10 @@
+from src.api.schemas.extension import Schema, fields
+
+
+class ApplicationStartRequestSchema(Schema):
+    competition_id = fields.UUID(required=True)
+
+
+class ApplicationStartResponseSchema(Schema):
+    message = fields.String()
```
chouinar, line 9: Make sure response schemas are derived from AbstractResponseSchema - don't need to specify anything but data with that.
```suggestion
class ApplicationStartResponseSchema(AbstractResponseSchema):

```

api/src/services/applications/create_application.py
```diff
@@ -0,0 +1,40 @@
...
+    ).scalar_one_or_none()
+
+    if not competition:
+        raise_flask_error(404, f"Competition with ID {competition_id} not found")
+
+    # Create a new application
+    application = Application(application_id=uuid.uuid4(), competition_id=competition_id)
+
+    db_session.add(application)
+    db_session.commit()
```
chouinar, line 30: Use a transaction from the top-level route instead of commiting as we go.

```py
with db_session.begin():
     create_application(...)
```

api/tests/src/api/applications/test_application_routes.py
```diff
@@ -0,0 +1,100 @@
...
+from src.db.models.competition_models import Application, Competition
+from tests.src.db.models.factories import CompetitionFactory, OpportunityFactory
+
+
+@pytest.fixture(autouse=True)
+def clear_competitions(db_session):
+    db_session.query(Application).delete()
+    db_session.query(Competition).delete()
+    db_session.commit()
+    yield
```
chouinar, line 15: FYI - if there is no logic after the yield, it doesn't do anything (eg. the fixture that creates the test DB tables creates tables, yields, then deletes tables)

api/src/services/applications/create_application.py
```diff
@@ -0,0 +1,40 @@
...
+
+    logger.info(
+        "Created new application",
+        extra={
+            "application_id": str(application.application_id),
+            "competition_id": str(competition_id),
```
chouinar, line 36: Our logging serialization _should_ handle converting UUIDs to strings for us - can just pass it the raw uuid

api/local.env
```diff
@@ -155,3 +155,5 @@ DEPLOY_TIMESTAMP=2024-12-02T21:25:18Z
 DEPLOY_WHOAMI=local-developer
 
 CDN_URL=http://localhost:4566/local-mock-public-bucket
+
+ENABLE_APPLY_ENDPOINTS=true
```
chouinar, line 159: This is already defined on line 64

api/src/api/application_alpha/application_schemas.py
mikehgrantsgov: I left it out, leave it just `fields.Dict()` or do we want something more defined type-wise?

chouinar: Wait - I'm mixing things up right now between this and other endpoints. We don't want dict/vaguely described schemas.

For this endpoint, we want something like (with better class names):
```
class InnerSchema(Schema):
     application_id = fields.UUID()

class OuterSchema(AbstractResponseSchema):
    data = fields.Nested(InnerSchema())
```

This is just the same pattern as all of our other response schemas that have at least something. See for example, the response schema of "Get Opportunity" https://github.com/HHS/simpler-grants-gov/blob/main/api/src/api/opportunities_v1/opportunity_schemas.py#L510

mikehgrantsgov: Ah gotcha, OK made that change.

chouinar (approved):
LGTM!

PR #4196, opened 2025-03-14
[Issue #3830] send saved search id down in response payload

doug-s-nava:
## Summary
Support for issue #3830 

## Changes proposed
Send id of saved search down in payload on successful save request

## Context for reviewers
This id will be used by the frontend to trigger a refetch of a user's saved search list

See https://github.com/HHS/simpler-grants-gov/pull/4170


chouinar (approved):
LGTM!

PR #4197, opened 2025-03-14
[Issue #4171] Create PUT `/applications/:applicationId/forms/:formId`

### LINKED ISSUE #4171: Create PUT /applications/:applicationId/forms/:formId
### Summary

Auth: ApiTokenAuth (not user)

### Request
The request contains an arbitrary JSON object (as it could represent the answers to any of hundreds of forms). For the schema, use the `Dict` field type.

```json
{
  "application_response": { ... }
}
```

### Response
The response should contain a list of warnings - note that you don't need to build a schema for this as the `WarningMixinSchema` already exists with this format, just add it to your base response schema. You can populate this field by populating the `warnings` object in the `ApiResponse` class. Warnings will be added in a later PR.

```json
{
 "data": {
   "application_id": "<some UUID>"
 },
"warnings": [
   {
      "type": "",
      "message": "",
      "field": ""
   }
 ]
}
```

### Behavior
Fetch the application from the DB (404 if it doesn't exist)

For the given application - the formId passed in needs to correspond with a form ID attached to the competition that the application is linked to (404 if that form doesn't exist).

Update the application with the new response.

<In a later PR, validation will be added here>

Only meaningful error conditions:
* Invalid auth token (same as search and a bunch of other endpoints)
* Application doesn't exist
* Form ID either doesn't exist or isn't attached to the competition the app is connected to

### Acceptance criteria

- [ ] Endpoint created
- [ ] Schemas created
- [ ] Tests added for basic scenarios

mikehgrantsgov:
## Summary
Fixes #4171

## Changes proposed
Add an endpoint that contains arbitrary JSON 
Returns warnings that are TBD.
Handles error conditions

## Context for reviewers
Warnings will be filled in, in a subsequent ticket.
Borrowed slightly from PRs that are already merged.

## Additional information
See attached unit tests.


api/tests/src/db/models/factories.py
```diff
@@ -1918,3 +1918,15 @@ class Meta:
     job_id = Generators.UuidObj
     job_status = factory.lazy_attribute(lambda _: JobStatus.COMPLETED)
     metrics = None
+
+
+class FormFactory(BaseFactory):
```
chouinar, line 1923: In the PR I sent early Friday, I setup a factory for Forms: https://github.com/HHS/simpler-grants-gov/pull/4188/files#diff-b91918457716d29ecaa77e0cf43ca8b23fd909f3ff2eac7cfa24be79c883f6c0 - can just copy until I get that merged

api/src/api/application_alpha/application_route.py
```diff
@@ -32,3 +36,25 @@ def application_start(db_session: db.Session, json_data: dict) -> response.ApiRe
...
+@application_blueprint.input(ApplicationFormUpdateRequestSchema, location="json")
+@application_blueprint.output(ApplicationFormUpdateResponseSchema)
+@application_blueprint.doc(responses=[200, 401, 404])
+@application_blueprint.auth_required(api_key_auth)
+@flask_db.with_db_session()
+def application_form_update(
+    db_session: db.Session, application_id: UUID, form_id: UUID, json_data: dict
+) -> response.ApiResponse:
+    """Update an application form response"""
+    logger.info(f"PUT /alpha/applications/{application_id}/forms/{form_id}")
```
chouinar, line 51: Counterintuitively, I'd actually recommend logging:
```suggestion
    add_extra_data_to_current_request_logs({"application.application_id": application_id, "form.form_id": form_id})
    logger.info("PUT /alpha/applications/:application_id/forms/:form_id")
```
All of our log messages automatically add both the URL rule and actual path (using the form route I recently added as an example):
```
request.path=/alpha/forms/550e8400-e29b-41d4-a716-446655440000
request.url_rule=/alpha/forms/<uuid:form_id>
```
So having the log message static makes it easier to group queries together.

The `add_extra_data_to_current_request_logs` bit makes it so every log message for the rest of the request lifecycle automatically includes those values as extra params.

api/src/services/applications/update_application_form.py
```diff
@@ -0,0 +1,85 @@
...
+
+    if not application:
+        raise_flask_error(404, f"Application with ID {application_id} not found")
+
+    # Check if form exists and is attached to the competition
+    competition_form = db_session.execute(
+        select(Form)
+        .join(Competition, Competition.competition_id == application.competition_id)
+        .where(Form.form_id == form_id)
+    ).scalar_one_or_none()
```
chouinar, line 45: I think this query is missing a join somewhere.

We need to make sure a form is associated with a competition, which would require the CompetitionForm model to be somewhere.

I think we want something roughly like:
```
    competition_form = db_session.execute(
        select(Form)
        .join(CompetitionForm, Form.form_id == CompetitionForm.form_id)
        # Technically don't even need to join to the competition table, competition_form is enough I think?
        .where(CompetitionForm.competition_id == application.competition_id)
        .where(Form.form_id == form_id)
    ).scalar_one_or_none()
```
Effectively, fetch a competition_form where:
* It's associated with the competition
* It has that specific form_id

api/src/services/applications/update_application_form.py
```diff
@@ -0,0 +1,85 @@
...
+    else:
+        # Create new application form
+        application_form = ApplicationForm(
+            application_id=application_id,
+            form_id=form_id,
+            application_response=application_response,
+        )
+        db_session.add(application_form)
+
+    db_session.commit()
```
chouinar, line 72: We want to use `db_session.begin()` for this as well

api/tests/src/api/applications/test_application_routes.py
```diff
@@ -98,3 +102,306 @@ def test_application_start_invalid_request(
...
+    )
+    db_session.add(application)
+
+    # Create form
+    form = FormFactory.create(
+        form_json_schema={"type": "object", "properties": {"name": {"type": "string"}}},
+        form_ui_schema={"name": {"ui:widget": "text"}},
+    )
+    db_session.add(form)
+    db_session.commit()
```
chouinar, line 126: The create function for a factory already does add/commit in it (which is why a lot of create calls is slow as it's a lot of separate commits).

```suggestion
    # Create form
    form = FormFactory.create(
        form_json_schema={"type": "object", "properties": {"name": {"type": "string"}}},
        form_ui_schema={"name": {"ui:widget": "text"}},
    )
```

api/src/services/applications/update_application_form.py
mikehgrantsgov: Got it. Should be good to go now. Also tweaked our factory usage.

api/src/api/application_alpha/application_schemas.py
```diff
@@ -12,3 +12,12 @@ class ApplicationStartResponseDataSchema(Schema):
...
     data = fields.Nested(ApplicationStartResponseDataSchema())
+
+
+class ApplicationFormUpdateRequestSchema(Schema):
+    application_response = fields.Dict(required=True)
+
+
+class ApplicationFormUpdateResponseSchema(WarningMixinSchema):
+    message = fields.String()
+    data = fields.Dict()
```
chouinar, line 23: For the response schema, it should also be derived from the AbstractResponseSchema to get the message field, but we should also set data to be equal to something that has the application ID in it.
```suggestion
class ApplicationFormUpdateResponseSchema(AbstractResponseSchema, WarningMixinSchema):
    data = fields.Nested(something)
```

api/src/services/applications/update_application_form.py
chouinar: Can clean up the commit now that it's in a begin().

api/tests/src/api/opportunities_v1/conftest.py
```diff
@@ -19,6 +20,15 @@ def truncate_opportunities(db_session):
...
     # https://docs.sqlalchemy.org/en/20/orm/queryguide/dml.html#orm-queryguide-update-delete-caveats
     # but if we do it individually they will
+
+    application_forms = db_session.query(ApplicationForm).all()
+    for application_form in application_forms:
+        db_session.delete(application_form)
+
+    opportunity_summary = db_session.query(OpportunitySummary).all()
+    for opp_summary in opportunity_summary:
+        db_session.delete(opp_summary)
```
chouinar, line 30: I just merged a change that adds a utility for deleting - but I also don't think we need the one for OpportunitySummary since they would get deleted by the opportunity deletes.

PR #4227, opened 2025-03-17
[Unticketed] Add ignores for vulnerability scans

chouinar:
## Changes proposed
Add ignores for vulnerabilities that we just need to wait until the Debian image is updated


mdragon (approved):
👍🏻

PR #4262, opened 2025-03-19
[Issue #4260] Fix GET saved opportunity filter by user ID 

### LINKED ISSUE #4260: GET saved opportunities doesn't filter by user ID
### Summary

The query to fetch saved opportunities for a user, doesn't filter by a specific user:
https://github.com/HHS/simpler-grants-gov/blob/16db06890bdeb6e481aaae732314605b0000e3db/api/src/services/users/get_saved_opportunities.py#L57

We need to modify the query to do so

### Acceptance criteria

- [ ] Fix query
- [ ] Adjusts tests to verify that if records exist for multiple users, only one is returned

mikehgrantsgov:
## Summary
Fixes #4260

## Changes proposed
Update query to join on `user_id`

## Context for reviewers
Previously this was not added, resulting in a bug showing all user saved opportunities.

## Additional information
See attached unit test (which fails in `main`), passes here.

### DIFF
--- a/api/src/services/users/get_saved_opportunities.py
+++ b/api/src/services/users/get_saved_opportunities.py
@@ -3,7 +3,7 @@ from typing import Sequence, Tuple
 from uuid import UUID
 
 from pydantic import BaseModel
-from sqlalchemy import asc, desc, nulls_last, select, and_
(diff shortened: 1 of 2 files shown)


chouinar (approved):
LGTM!

PR #4266, opened 2025-03-19
[Issue #4265] Expand the test agency list to account for Prod data

### LINKED ISSUE #4265: Identify more test agencies
### Summary

Update test agency list to reflect test agencies in Prod

### Acceptance criteria

- [ ] Fewer obviously testing agencies are showing in Prod

mdragon:
## Summary
Fixes #4265 

## Changes proposed
Add more agencies that shouldn't show in the get agencies endpoint

### DIFF
--- a/api/src/data_migration/transformation/subtask/transform_agency.py
+++ b/api/src/data_migration/transformation/subtask/transform_agency.py
@@ -97,7 +97,38 @@ class AgencyConfig(PydanticBaseEnvConfig):
     # TODO - we might want to put this somewhere more central
     #        as we might want to filter these out in other places
     test_agency_config: set[str] = Field(
-        default={"GDIT", "IVV", "IVPDF", "0001", "FGLT", "NGMS", "NGMS-Sub1", "SECSCAN"}
+        default={
+            "GDIT",
+            "IVV",
+            "IVPDF",
+            "0001",
+            "FGLT",
+            "NGMS",
+            "NGMS-Sub1",
+            "SECSCAN",
+            "MH",
+            "TX",
+            "MN",
+            "MMC",
+            "WWC",
+            "SCRC",
+            "NRC",
+            "JL04022024",
+            "JUSFC",
+            "JMM",
+            "IAF",
+            "USIP",
+            "GCERC",
+            "ARPAH",
+            "ORD",
+            "DC",
+            "SCC800",
+            "BBG",
+            "ACR",
+            "ECP",
+            "MC",
+            "CCFF",
+        }
     )
 
 
--- a/api/src/data_migration/transformation/subtask/transform_agency.py
+++ b/api/src/data_migration/transformation/subtask/transform_agency.py
@@ -100,7 +100,7 @@ class AgencyConfig(PydanticBaseEnvConfig):
     prefix_env: str = Field(
         default="GDIT,IVV,IVPDF,0001,FGLT,NGMS,SECSCAN", alias="TEST_AGENCY_PREFIXES"
     )
-    test_agency_config: set[str] = Field(default=[])
+    test_agency_config: set[str] = Field(default=set())
 
     def model_post_init(self, _context: typing.Any) -> None:
         """Run after __init__ sets above values from env vars"""
--- a/api/src/data_migration/transformation/subtask/transform_agency.py
+++ b/api/src/data_migration/transformation/subtask/transform_agency.py
@@ -1,4 +1,5 @@
 import logging
+import os
 from dataclasses import dataclass, field
 from datetime import datetime
 from enum import StrEnum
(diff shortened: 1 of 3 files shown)


api/src/data_migration/transformation/subtask/transform_agency.py
```diff
@@ -97,7 +97,7 @@ class AgencyConfig(PydanticBaseEnvConfig):
     # TODO - we might want to put this somewhere more central
     #        as we might want to filter these out in other places
     test_agency_config: set[str] = Field(
-        default={"GDIT", "IVV", "IVPDF", "0001", "FGLT", "NGMS", "NGMS-Sub1", "SECSCAN"}
+        default={"GDIT", "IVV", "IVPDF", "0001", "FGLT", "NGMS", "NGMS-Sub1", "SECSCAN", "MH", "TX", "MN", "MMC", "WWC", "SCRC", "NRC", "JL04022024", "JUSFC", "JMM", "IAF", "USIP", "GCERC", "GSA", "FCC", "ARPAH", "ORD", "DC", "SCC800", "BBG", "ACR", "ECP", "MC", "CCFF"}
```
mdragon, line 100: `alias="TEST_AGENCY_PREFIX",`
It seemed like I could add something like this to pull from Env Var, but I wasn't sure how the conversion to dict would happen?

api/src/data_migration/transformation/subtask/transform_agency.py
```diff
@@ -97,7 +97,7 @@ class AgencyConfig(PydanticBaseEnvConfig):
     # TODO - we might want to put this somewhere more central
     #        as we might want to filter these out in other places
     test_agency_config: set[str] = Field(
-        default={"GDIT", "IVV", "IVPDF", "0001", "FGLT", "NGMS", "NGMS-Sub1", "SECSCAN"}
+        default={"GDIT", "IVV", "IVPDF", "0001", "FGLT", "NGMS", "NGMS-Sub1", "SECSCAN", "MH", "TX", "MN", "MMC", "WWC", "SCRC", "NRC", "JL04022024", "JUSFC", "JMM", "IAF", "USIP", "GCERC", "ARPAH", "ORD", "DC", "SCC800", "BBG", "ACR", "ECP", "MC", "CCFF"}
```
mdragon, line 100: `        alias="TEST_AGENCY_PREFIX",`
It seemed like I could add something like this to map the Env Var in but wasn't sure how we'd go from comma delimited list to this type of structure.

chouinar, line 100: We don't want to adjust the configuration here - we should move all of this to env vars in terraform. If we filter out stuff like `MH` in dev/staging we won't have any data at all.

It sounds like Pydantic parses lists (presumably sets) as just comma-separated: https://docs.pydantic.dev/latest/concepts/pydantic_settings/#parsing-environment-variable-values - can test that real quick by changing your local.env file for this env var.

chouinar: I can help look at this more tomorrow.

mdragon: Ok, with that pointer, I can look more and try to make it environment based tomorrow

chouinar: @mdragon 

A quick test with local.env - was able to get the value properly set by doing:
```
TEST_AGENCY_CONFIG='["A","B","C","D"]'
```
The quotes had to be exactly that for it to be happy, not sure how we'll need to format it in terraform since that usually has double-quotes on the outer bit (hopefully we don't need to escape everything).

api/src/data_migration/transformation/subtask/transform_agency.py
```diff
@@ -96,8 +97,14 @@
 class AgencyConfig(PydanticBaseEnvConfig):
     # TODO - we might want to put this somewhere more central
     #        as we might want to filter these out in other places
-    test_agency_config: set[str] = Field(
-        default={"GDIT", "IVV", "IVPDF", "0001", "FGLT", "NGMS", "NGMS-Sub1", "SECSCAN"}
+    prefix_env: str | None = os.getenv("TEST_AGENCY_PREFIXES")
```
mdragon, line 100: @chouinar Interested in your thoughts on this approach. The big upside to me is it lets the format in the env var be:
```ABC,DEF,XYZ```
So no risk of breaking the parsing, needing to escape quotes within quotes, etc. It makes the Python code a bit less glamorous but feels much safer/easier to maintain the list going forward.

Comment (chouinar) [reply api/src/data_migration/transformation/subtask/transform_agency.py:100]: First thing - you don't need to ever use os.getenv, that's what the point of these env configs is, doing that for you (and converting types).
```suggestion
    prefix_env: str | None = Field("GDIT,IVV,IVPDF,0001,FGLT,NGMS,SECSCAN", alias="TEST_AGENCY_PREFIXES")
```

For how to actually parse this into a set, I'd recommend following a pattern like we did for the deploy metadata: https://github.com/HHS/simpler-grants-gov/blob/main/api/src/util/deploy_metadata.py#L38

There's a `model_post_init` function you can setup and Pydantic will run it after it loads everything else. So you could do something like:

```py
    def model_post_init(self, _context: typing.Any) -> None:
        """Run after __init__ sets above values from env vars"""

        self.test_agency_config = set(self.prefix_env.split(","))
```

api/src/data_migration/transformation/subtask/transform_agency.py
```diff
@@ -96,10 +97,13 @@
 class AgencyConfig(PydanticBaseEnvConfig):
     # TODO - we might want to put this somewhere more central
     #        as we might want to filter these out in other places
-    test_agency_config: set[str] = Field(
-        default={"GDIT", "IVV", "IVPDF", "0001", "FGLT", "NGMS", "NGMS-Sub1", "SECSCAN"}
-    )
+    prefix_env: str = Field(default="GDIT,IVV,IVPDF,0001,FGLT,NGMS,SECSCAN", alias="TEST_AGENCY_PREFIXES")
+    test_agency_config: set[str] = Field(default=[])
```
mdragon, line 101: If I didn't give this a default it seemed to throw a validation warning because it seems it's being validated before post_init can run.

chouinar: Two options:
* Don't define it at all - which is what I did for the metadata config, although that runs into type checkers not being happy because it's not defined on the class (it can't tell that post_init is effectively a part of init)
* What you did here

No real preference personally

chouinar (dismissed):
LGTM!

api/src/data_migration/transformation/subtask/transform_agency.py
chouinar: To fix the issue here, change the default to be a set instead of a list

chouinar (approved):

PR #4275, opened 2025-03-20
[Issue #3551] Setup tables for transfer and staging tables for user data

### LINKED ISSUE #3551: Setup tables for transfer and staging tables for user data
### Summary

We want to setup two tables (4 in total, 2 in transfer, 2 in staging schemas). The columns must be named, typed, and ordered exactly as listed here for our import process to work correctly.

`TUSER_ACCOUNT_MAPPER`
* `USER_ACCOUNT_ID` - int, primary key
* `SOURCE_TYPE` - str, primary key
* `EXT_USER_ID` - str, NOT NULL
* `EXT_ISSUER` - str, NOT NULL
* `LAST_AUTH_DATE` - datetime, NOT NULL
* `CREATED_DATE` - datetime, nullable
* `CREATOR_ID` - str, nullable
* `LAST_UPD_DATE` - datetime, nullable
* `LAST_UPD_ID` - text, nullable

`TUSER_ACCOUNT`
* `USER_ACCOUNT_ID` - int, primary key
* `USER_ID` - str, NOT NULL
* `FULL_NAME` - str, nullable
* `EMAIL_ADDRESS` - str, nullable
* `PHONE_NUMBER` - str, nullable
* `FIRST_NAME` - str, nullable
* `MIDDLE_NAME` - str, nullable
* `LAST_NAME` - str, nullable
* `IS_DELETED` - str, NOT NULL
* `IS_DUPLICATE` - str, NOT NULL
* `IS_ACTIVE` - str, NOT NULL
* `CREATED_DATE` - datetime
* `CREATOR_ID` - str, nullable
* `LAST_UPD_DATE` - datetime, nullable
* `LAST_UPD_ID` - str, nullable
* `IS_EMAIL_CONFIRM_PENDING` - str, nullable
* `DEACTIVATED_DATE` - datetime, nullable
* `MOBILE_NUMBER` - str, nullable

### Acceptance criteria

- [ ] Tables created
- [ ] Migrations created (only the staging table will end up in the final tables)

mikehgrantsgov:
## Summary
Fixes #3551

## Changes proposed
Add two tables (4 in total, 2 in transfer, 2 in staging schemas). 
Add models / migration in two schemas (exactly ordered) 

## Context for reviewers
Only models and migrations needed for this task.


api/src/db/models/user_models.py
```diff
@@ -162,3 +163,81 @@ class UserOpportunityNotificationLog(ApiSchemaTable, TimestampMixin):
     opportunity: Mapped[Opportunity] = relationship(
         "Opportunity", back_populates="all_opportunity_notification_logs"
     )
+
+
+class StagingUserAccountMapper(Base):
```
chouinar, line 168: I don't think I did a good enough job explaining this ticket, the way these tables work is a bit different - they don't go with the rest of our schemas.

We should instead define a mixin (like we did for the Topportunity table and others): https://github.com/HHS/simpler-grants-gov/blob/main/api/src/db/models/legacy_mixin/opportunity_mixin.py

We then need to define derived classes for the foreign: https://github.com/HHS/simpler-grants-gov/blob/main/api/src/db/models/foreign/opportunity.py

and staging: https://github.com/HHS/simpler-grants-gov/blob/main/api/src/db/models/staging/opportunity.py

schemas. Note, these derived ones don't need anything in them except the table names. Everything else comes purely from the classes you derive them from (follow the same pattern just swap out the mixin).

---
Why are these defined in this weird way? A few reasons:
1. We really want the schemas to match for the foreign & staging tables, so defining the columns once helps.
2. The foreign tables especially need to be defined _exactly_ in the way they are for the process that loads them to work - down to the place they're defined and the classes they're derived from. To actually create the foreign tables, we run [this script](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/data_migration/setup_foreign_tables.py) which when run non-locally actually sets up a foreign data wrapper to an Oracle DB table.

api/src/db/models/user_models.py
```diff
@@ -162,3 +163,81 @@ class UserOpportunityNotificationLog(ApiSchemaTable, TimestampMixin):
...
+    __tablename__ = "TUSER_ACCOUNT_MAPPER"
+    __table_args__ = {"schema": Schemas.STAGING}
+
+    USER_ACCOUNT_ID: Mapped[int] = mapped_column(primary_key=True)
```
chouinar, line 172: FYI - for these, we can make them lowercase, Postgres is case-insensitive so we can at least have some err on the side of readability without any issue.

api/src/db/models/legacy_mixin/user_mixin.py
```diff
@@ -0,0 +1,38 @@
+from datetime import datetime
+
+from sqlalchemy.orm import Mapped, declarative_mixin, mapped_column
+
+
+@declarative_mixin
+class TuserAccountMapperMixin:
+    user_account_id: Mapped[int] = mapped_column(primary_key=True)
+    source_type: Mapped[str]
```
chouinar, line 9: Two issues, one is on me for not putting it in the docs right:
1. This should also be a primary key
2. This should be the last column defined (don't know why I put it second)

api/src/db/models/legacy_mixin/user_mixin.py
```diff
@@ -0,0 +1,38 @@
...
+    ext_user_id: Mapped[str]
+    ext_issuer: Mapped[str]
+    last_auth_date: Mapped[datetime]
```
chouinar, line 12: In double-checking the DB, I completely missed a column (I clearly was having an off day when I wrote this ticket).

Between these two we need `ext_subject: Mapped[str | None]`

api/src/db/models/legacy_mixin/user_mixin.py
```diff
@@ -0,0 +1,38 @@
...
+    full_name: Mapped[str | None]
+    email_address: Mapped[str | None]
+    phone_number: Mapped[str | None]
+    first_name: Mapped[str | None]
+    middle_name: Mapped[str | None]
+    last_name: Mapped[str | None]
+    is_deleted_legacy: Mapped[str]
+    is_duplicate: Mapped[str]
+    is_active: Mapped[str]
+    created_date: Mapped[datetime]
```
chouinar, line 32: Another mistake of mine - this should be nullable:
```suggestion
    created_date: Mapped[datetime | None]
```

api/src/db/models/legacy_mixin/user_mixin.py
mikehgrantsgov: I think this last change should do it.

chouinar (approved):

Issue #3909, opened 2025-02-18
Implementation: Show Facet Count


capriiis:
### Summary

See screenshot and notes below for details.

### Acceptance criteria

- Two spaces between item and number count
- Style update

![Image](https://github.com/user-attachments/assets/4e655481-4772-4916-9379-be52af835ae9)
![Image](https://github.com/user-attachments/assets/871b33a4-b7e4-4702-a8f6-44db3c197b54)

crystabelrangel:
Initial decisions from the call on 3/25 regarding [slack convo](https://betagrantsgov.slack.com/archives/C05TGEL3C6Q/p1742909585202319):

- do not hide options if there are 0 results

Agency filter
- update/build new end point to point to status (open/forecasted)
- will look into the count for closed status to determine if we will add this
- new task will be made for this change

cc: @doug-s-nava @mdragon @andycochran @chrisw-nava

PR #4276, opened 2025-03-21
[unticketed] refactor search filter accordion code into multiple components

doug-s-nava:
…ier to suspend

## Summary
unticketd but working in the service of #3909

## Changes proposed
split search filter accordion into multiple components to make it easier to suspend


acouch (approved):
Looks good, thanks.

PR #4290, opened 2025-03-24
[Issue #4168] Update competition/form/application factories to better support test setup

### LINKED ISSUE #4168: Setup local db seed script to create a competition with a form that the frontend can use for development
### Summary

Our seed script should add a bit of logic that:
* For one opportunity (can be lazy, make it opportunity_id 1 which will always get created)
   * Create a form in the table with the schemas described below
   * Create a competition attached to the opportunity
   * Add the form to the competition (the CompetitionForm table to link them)

JsonSchema
```json
 TODO - checking with frontend folks
```

UiSchema
```json
TODO - checking with frontend folks
```

### Acceptance criteria

- [ ] Seed script updated

chouinar:
## Summary
Fixes #4168

## Changes proposed
* Form factory now has a valid json schema + UI schema
* Adjusted schema of form response as UI schema is a list, not a dict (make it raw type to be JSON)
* Adjusted factories ordering to be a bit better organized
* Cleaned up a few tests that were doing a lot more than necessary to setup data

## Context for reviewers
Note that the rearranging of factories is a sorta-precursor to work I'm thinking we need to do to split this file up a bit (at least move the staging/legacy table factories to a separate file).

The test updates are mostly to remove redundant factory usage. For example, in some tests we were creating records, then passing those to another factory, and then passing that to a 3rd factory. Unless there is a specific setup on one of those earlier records, we don't need to do this as the factories themselves create all of their dependencies. Before this change a few of those dependencies were missing, so the tests weren't wrong, just didn't fully have the support from these changes.


api/tests/lib/db_testing.py
```diff
@@ -72,6 +72,7 @@ def cascade_delete_from_db_table(db_session: db.Session, table_model) -> None:
     with whatever data might exist in a given table.
     """
 
+    db_session.expunge_all()
```
joshtonava, line 75: 👍

joshtonava (approved):
This will be useful for my open PR

Issue #4091, opened 2025-03-10
Replace process page with the roadmap page.


freddieyebra:
### Summary

Remove feature flag from roadmap work done in #3936 and replace the process page link with the roadmap page link. 

Any dead code related to the process page should also be cleaned up as a part of this task.

### Acceptance criteria

- [x] Process page is replaced with the roadmap page.
- [x] Process page dead code is cleaned up.
- [x] All references to the process page now point to the roadmap page.

capriiis:
Convo with @andycochran needed around preliminary steps, @capriiis will follow-up

capriiis:
Freddie has advised he was able to resolve his outstanding questions, got the PR merged, and a meeting with Andy is no longer needed.

PR #4291, opened 2025-03-24
[unticketed] Update progress link on homepage.

freddieyebra:
## Summary
no ticket but related to #4091 

## Changes proposed
Fix a link on the home page that still tries to redirect to the /process and redirect it to /roadmap instead.

## Context for reviewers
There was an instance of replacing all of the links that go to /process page to /roadmap. To test this:
- Checkout this branch `git checkout fyebra/fix-roadmap-link`
- Run the frontend locally with `npm run dev`
- Open the homepage and click the "learn about our progress" button.
- Ensure it redirects to /roadmap page.

### DIFF
--- a/frontend/src/components/content/ProcessAndResearchContent.tsx
+++ b/frontend/src/components/content/ProcessAndResearchContent.tsx
@@ -20,7 +20,7 @@ const ProcessAndResearchContent = () => {
         <p className="font-sans-md line-height-sans-4 padding-bottom-2 desktop-lg:line-height-sans-6">
           {t("process_and_research.paragraph_1")}
         </p>
-        <Link href="/process" passHref>
+        <Link href="/roadmap" passHref>
           <Button className="margin-bottom-4" type="button" size="big">
             <span className="margin-right-5">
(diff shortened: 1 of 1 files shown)


andycochran (approved):
Should we also consider routing requests for `/process` to `/roadmap`?

frontend/next.config.js
```diff
@@ -104,6 +104,15 @@ const nextConfig = {
   experimental: {
     testProxy: true,
   },
+  async redirects() {
+    return [
+      {
+        source: "/process",
+        destination: "/roadmap",
+        permanent: true,
```
andycochran, line 112: Will this be a 301? Would that cause cache issues if we ever want to reuse the `/process` URL?

freddieyebra: good call. I think this would fall under a temporary redirect which would respond with 302 instead of permanent 301.

freddieyebra: Additionally, if we ever wanted to use /process, we would need to remove this redirect rule.

andycochran (approved):

Issue #4017, opened 2025-02-25
Test incremental approach to loading data (including attachments) into the search index


mxk0:
### Summary

\

### Acceptance criteria

\

capriiis:
Unable to proceed with testing

chrisw-nava:
@freddieyebra @chouinar @babebe - is this ticket still needed or are we waiting on anything in particular?

chrisw-nava:
Thanks, @freddieyebra! I moved the ticket back into To Do for the current sprint

babebe:
@chrisw-nava tested yesterday, incremental approach did run successfully and able to index the attachments. @freddieyebra we can do one more test in staging with the scheduled job disabled before we close it.

PR #4293, opened 2025-03-24
[Issue #4017] Incremental approach enhancement

babebe:
## Summary
Fixes #{[4017](https://github.com/HHS/simpler-grants-gov/issues/4017)}

## Changes proposed
Parallelizing the processing and loading  of opportunity records in batches
The search index will be refreshed after each batch (as set per `incremental_load_batch_size`) is loaded


api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -343,32 +351,49 @@ def load_records(self, records: Sequence[Opportunity]) -> set[int]:
...
-            # If the opportunity has a test agency, skip uploading it to the index
-            if record.agency_record and record.agency_record.is_test_agency:
-                logger.info(
-                    "Skipping upload of opportunity as agency is a test agency",
-                    extra=log_extra | {"agency": record.agency_code},
-                )
-                self.increment(self.Metrics.TEST_RECORDS_SKIPPED)
-                continue
+        # This will run in parallel
+        def process_opportunities(record: Opportunity) -> tuple[dict[str, Any], int] | None:
```
chouinar, line 355: Why does this only take in a single opportunity? That would mean each thread processes exactly 1 opportunity.

I was imagining something like:
* (outside of thread) - we fetch X opportunities from the DB
* Each thread takes X / thread_count opportunities to process (probably aim for like 25-50 in each?)
* (inside thread) - loop over the batch of opportunities like before

api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -343,32 +351,49 @@ def load_records(self, records: Sequence[Opportunity]) -> set[int]:
...
+            futures = []
+            for record in records:
+                futures.append(executor.submit(process_opportunities, record))
+
+            for future in as_completed(futures):
+                result = future.result()  # will raise the exception if worker thread fails
+                if result:
+                    json_record, opportunity_id = result
+                    json_records.append(json_record)
+                    loaded_opportunity_ids.add(record.opportunity_id)
```
chouinar, line 396: Why is the json_record being returned here, the call to OpenSearch should be in the threading, it's fine if 10 different threads are calling it at the same time, it can handle that

api/src/search/backend/load_opportunities_to_index.py
babebe: Updated.

api/src/search/backend/load_opportunities_to_index.py
babebe: Now being called for each batch of opportunities

chouinar:
> > The search index will be refreshed after all opportunities have been successfully uploaded.
> 
> Just wanted to flag this, not sure if we can address it easily or at all. What happens if something fails? Since the benefit of the work doesn't get made visible to the user calling the search API until we do the refresh, what happens if a batch bombs out? Can we be reasonably sure we'll still run the refresh? If not should we move the refresh to run per batch knowing that hurts our throughput, but avoids the work of the job not being visible to the user?

If I'm reading the docs right, the default refresh_interval for an index is 1 second: https://opensearch.org/docs/latest/install-and-configure/configuring-opensearch/index-settings/#dynamic-index-level-index-settings

We could always just set the refresh interval on the index to be safe, and then the backend job failing to upload something wouldn't matter anyways.

babebe:
@mdragon. @chouinar 
Good catch, it would not refresh since this is being called inside an outer try/except block. Could just leave it to refresh for each batch of records that are loaded into the search index(Current behavior) 
But, do we want the process to continue if a failure occurs for an opportunity and/or if failure occurs when a batch is processed (ex: when calling bulk_upsert for the batch to load into open search) right now it will break.

api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -335,44 +367,48 @@ def get_attachment_json_for_opportunity(
...
-                logger.info(
-                    "Skipping upload of opportunity as agency is a test agency",
-                    extra=log_extra | {"agency": record.agency_code},
-                )
-                self.increment(self.Metrics.TEST_RECORDS_SKIPPED)
-                continue
+        batch_json_records = []
+        batch_processed_opp_ids = set()
+        for record in batch:
+            try:
```
chouinar, line 378: Be careful with adding a try block here. Try to wrap as little code as possible. For example, the log_extra, checking if something is a test agency, and so on shouldn't ever fail - if they did we actually probably prefer the job to fall over entirely and not catch the error.

But if that's the case, what is this try/except catching, what errors are we fine with the processing continuing for? If an opportunity can't be turned into JSON that seems bad, indicates a real issue with the system. Maybe the attachment logic is fine to wrap for now, not sure if we've hit any issues with that yet.

api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -335,44 +367,48 @@ def get_attachment_json_for_opportunity(
             (TransportError, ConnectionTimeout)
         ),  # Retry on TransportError (including timeouts)
     )
-    def load_records(self, records: Sequence[Opportunity]) -> set[int]:
+    def load_records(self, batch: Sequence[Opportunity]) -> set[int]:
```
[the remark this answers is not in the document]
babebe: Updated

api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -335,44 +367,48 @@ def get_attachment_json_for_opportunity(
...
 
-            json_records.append(json_record)
-            self.increment(self.Metrics.RECORDS_LOADED)
+                self.increment(self.Metrics.RECORDS_LOADED)
+                batch_json_records.append(json_record)
+                batch_processed_opp_ids.add(record.opportunity_id)
 
-            loaded_opportunity_ids.add(record.opportunity_id)
+            except Exception as e:
+                logger.error(f"Error processing opportunity {record.opportunity_id}: {e}")
```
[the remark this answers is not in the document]
babebe: Updated

babebe:
> > > The search index will be refreshed after all opportunities have been successfully uploaded.
> > 
> > 
> > Just wanted to flag this, not sure if we can address it easily or at all. What happens if something fails? Since the benefit of the work doesn't get made visible to the user calling the search API until we do the refresh, what happens if a batch bombs out? Can we be reasonably sure we'll still run the refresh? If not should we move the refresh to run per batch knowing that hurts our throughput, but avoids the work of the job not being visible to the user?
> 
> If I'm reading the docs right, the default refresh_interval for an index is 1 second: https://opensearch.org/docs/latest/install-and-configure/configuring-opensearch/index-settings/#dynamic-index-level-index-settings
> 
> We could always just set the refresh interval on the index to be safe, and then the backend job failing to upload something wouldn't matter anyways.

api/src/search/backend/load_opportunities_to_index.py
babebe: Haven't run into any issues. But put it in to see if we should continue to process if a failure occurs on the opportunity level. Do we want to continue if error occurs there.

api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -180,9 +187,24 @@ def _handle_incremental_upserts(self, existing_opportunity_ids: set[int]) -> Non
 
                 processed_opportunity_ids.add(opportunity.opportunity_id)
 
-            # Bulk index the opportunities (handles both inserts and updates)
-            loaded_ids = self.load_records(queued_opportunities)
-            logger.info(f"Indexed {len(loaded_ids)} opportunities")
+            # Determine how many opportunities each thread will process
+            opportunities_per_thread = max(
+                len(queued_opportunities) // self.config.incremental_load_max_workers, 1
+            )
```
chouinar, line 193:
```suggestion
```

api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -50,6 +54,10 @@ class LoadOpportunitiesToIndexConfig(PydanticBaseEnvConfig):
 
     incremental_load_batch_size: int = Field(default=1000)
     incremental_load_max_process_time: int = Field(default=3000)
+    # Configurable max worker. Set default to ThreaPoolExecutor default.
+    # See: https://docs.python.org/dev/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor
+    incremental_load_max_workers: int = Field(default=(os.cpu_count() or 1) + 4)
+    batch_size_for_processing: int = Field(default=100)
```
chouinar, line 60: This isn't used anywhere at the moment.

api/src/adapters/search/opensearch_client.py
```diff
@@ -181,7 +181,15 @@ def cleanup_old_indices(self, index_prefix: str, indexes_to_keep: list[str]) ->
         for index in old_indexes:
             self.delete_index(index)
 
-    def swap_alias_index(self, index_name: str, alias_name: str) -> None:
+    def refresh_index(self, index_name: str) -> None:
```
chouinar, line 184: Shouldn't this be called once at the end of processing?

api/src/search/backend/load_opportunities_to_index.py
babebe: removed

api/src/search/backend/load_opportunities_to_index.py
babebe: I have minimized the try/except to the `opportunity attachment` logic. If it fails, the process will continue.

api/src/adapters/search/opensearch_client.py
babebe: Yes, getting called after each batch is processed

api/src/adapters/search/opensearch_client.py
```diff
@@ -93,7 +93,7 @@ def bulk_upsert(
         records: Iterable[dict[str, Any]],
         primary_key_field: str,
         *,
-        refresh: bool = True,
+        refresh: bool = False,
```
chouinar, line 96: There are a bunch of tests failing, I'm guessing because of this change and those tests relying on data to be setup and refreshed. I'd suggest switching this back, but changing the load script to set it to false

api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -339,40 +354,51 @@ def load_records(self, records: Sequence[Opportunity]) -> set[int]:
...
+        # Bulk upsert for the current batch
+        if batch_json_records:
             self.search_client.bulk_upsert(
-                self.index_name, json_records, "opportunity_id", pipeline="multi-attachment"
+                self.index_name, batch_json_records, "opportunity_id", pipeline="multi-attachment"
             )
 
-        return loaded_opportunity_ids
+        # refresh index after batch is processed
+        self.search_client.refresh_index(self.index_name)
```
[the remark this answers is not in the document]
babebe: Moved after thread executor run

api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -339,40 +354,51 @@ def load_records(self, records: Sequence[Opportunity]) -> set[int]:
...
-                )
+                try:
+                    json_record["attachments"] = self.get_attachment_json_for_opportunity(
+                        record.opportunity_attachments
+                    )
+                except Exception:
+                    logger.exception(
+                        "Error preparing opportunity attachment for search index", extra=log_extra
+                    )
+                    continue
```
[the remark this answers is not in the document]
babebe: Removed.

api/src/adapters/search/opensearch_client.py
babebe: Done

PR #4294, opened 2025-03-24
[no ticket] Dont run migrations inside of the sidecar

coilysiren:
## Context

Currently the migrations run in the first container in the list, instead of trying to get the primary application container


mdragon (approved):
🚀

PR #4295, opened 2025-03-24
[Unticketed] Remove dependency on pytest-watch

chouinar:
## Changes proposed
Removed pytest-watch which hadn't been updated since 2018 and was causing issues building due to a recent change in setuptools

## Context for reviewers
This just removes the ability to watch tests and have it constantly run tests everytime your code changes. If we really want that functionality, we can look into https://pypi.org/project/pytest-watcher/ which seems to be the spiritual successor to pytest-watch and is still maintained.

### DIFF
--- a/api/Makefile
+++ b/api/Makefile
@@ -225,9 +225,6 @@ test: ## Run all tests except for audit logging tests
 test-audit: ## Run audit logging tests
 	$(PY_RUN_CMD) pytest -m "audit" $(args)
 
-test-watch: ## Run tests continually and watch for changes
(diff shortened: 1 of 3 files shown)


coilysiren:
@chouinar if you want to avoid adding another python package, this command looks like it does the same thing: https://formulae.brew.sh/formula/entr

mdragon (approved):

mdragon:
I actually tried to run the watch on Friday and it was failing to execute the tests with an error I didn't try to troubleshoot as I figured out how to run the tests against just a single file instead.

PR #4304, opened 2025-03-24
[Issue #3909] add facets counts to search filter options and hide options with no results

### LINKED ISSUE #3909: Implementation: Show Facet Count

doug-s-nava:
## Summary
Fixes #3909 

Depends on https://github.com/HHS/simpler-grants-gov/pull/4276

## Changes proposed

* adds a facet count - a display showing the number of current search results related to a given filter option - to each filter option in all filter accordions except "agency"
* to support this, moves the search request call up a level, and drills the result down to the filter checkbox components
* changes status filter view to a standard accordion rather than the unique implementation previously used

### Note on tests

The search page unit test was deleted for a couple reasons:

* the page component doesn't do very much on its own, largely an entry point into other component behavior
* underlying behavior of search functionality (filters, search results, etc.) is, or should be, fully unit tested within smaller components
* use of suspendcd and unsuspended async behavior in child components makes testing that behavior here very difficult, since suspense needs to be mocked for use with testing library
* integrated behavior of the page is better tested using Playwright

As a result of deleting this the code coverage has gone down slightly. A main culprit on this page is the SearchFilterAccordion, which I restructured a bit, but still isn't structured well for testing. I don't think that's a big concern right now, as we can handle the behavior with e2e tests, but something to consider for later.

We should also consider whether we want to bother unit testing top level pages at all, or make it a clear standard that top level pages should not hold business logic whenever possible, and put that logic into child components that can be more easily isolated and tested on their own.

### Test steps
3. _VERIFY_: the status filter accordion is open with "posted" and "forecasted" options selected
4. _VERIFY_: facet counts display for all status options
5. open all filter accordions
6. _VERIFY_: all filter options show facet counts, except agencies
7. select an option with a positive facet count
8. _VERIFY_: search updates and facet counts change
9. select an option with a 0 facet count
10. _VERIFY_: option is selected but results and counts do not change

## Additional information

Screenshot 2025-03-27 at 10 16 36 AM


frontend/src/components/search/SearchFilterAccordion/SearchFilterSection/SearchFilterSection.tsx
```diff
@@ -66,6 +68,14 @@ const SearchFilterSection: React.FC<SearchFilterSectionProps> = ({
     toggleSelectAll(false, currentSelections);
   }, [toggleSelectAll, accordionTitle, searchParams, allSectionOptions]);
 
+  const childOptionsWithFacetCounts = option.children.filter((child) => {
+    return facetCounts[child.value];
+  });
+
+  if (!childOptionsWithFacetCounts.length) {
```
mdragon, line 75: Just checking that I'm reading this right. This is the piece that will hide the top level agency if none of it's sub-agencies have any facet counts, correct?

doug-s-nava: correct, I can add a code comment if you think that'd be helpful

frontend/src/app/[locale]/search/page.tsx
```diff
@@ -48,6 +49,8 @@ function Search({ searchParams, params }: SearchPageProps) {
     resolvedSearchParams.page = "1";
   }
 
+  const searchResultsPromise = searchForOpportunities(convertedSearchParams);
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/search/page.tsx:52]: moved out here to allow the filter components to consume the response to get facets

frontend/src/components/search/SearchFilters.tsx
```diff
@@ -13,36 +14,60 @@ import {
...
   const agenciesPromise = getAgenciesForFilterOptions();
 
+  let searchResults;
+  try {
+    searchResults = await searchResultsPromise;
+  } catch (e) {
+    console.error("Search error, cannot set filter facets", e);
+  }
+
+  const facetCounts = searchResults?.facet_counts || defaultFacetCounts;
```
doug-s-nava, line 50: the tricky piece here is that we have an async call in this component, but the component is not suspended. I did it this way because all we care about from the search response is the facet count, which is not vitally important to the functionality of the page. If we suspended on this, it would make the entire filter section unusable until the search results load, as fallbacks (I learned while working on this) cannot be interactive.

Also note that the promise is resolved here since beyond this in the tree we are in client component territory.

This works fine as far as I can tell, but makes testing tricky, as the test suite expects things like this to be suspended

acouch: I was thinking we'd have to do something like this, since we don't want all the filers and results flashing every time someone visits the page. This seems like a good outcome. It helps that most of the filters are hidden by default. The initial page render might be cached by a CDN at some point as well.

acouch (dismissed):
Looks great, excited we are adding this.

PR #4305, opened 2025-03-24
[no ticket] Don't wait for logs when doing `run/command`

coilysiren:
## Context

Logs don't go to cloudwatch anymore, so you can't wait for them

### DIFF
--- a/bin/run-command
+++ b/bin/run-command
@@ -118,6 +118,9 @@ echo "Task information"
 echo "  task_arn=${task_arn}"
 echo "  task_id=${ecs_task_id}"
 
+echo
+echo "Waiting for task to complete"
+
 aws ecs wait tasks-stopped --cluster "${cluster_name}" --tasks "${task_arn}"
 
 container_exit_code=$(
@@ -139,3 +142,6 @@ if [[ "${container_exit_code}" == "null" || "${container_exit_code}" != "0" ]]; 
   echo "Task status (lastStatus, stopCode, stoppedAt, stoppedReason): ${task_status}" >&2
   exit 1
 fi
+
+echo
+echo "Task complete, container exit code: ${container_exit_code}"


mdragon (dismissed):
👍🏻

bin/run-command
```diff
@@ -118,104 +118,3 @@ printf " ... %s\n" "${aws_args[@]}"
...
-
-container_exit_code=$(
-  aws ecs describe-tasks \
-    --cluster "${cluster_name}" \
-    --tasks "${task_arn}" \
-    --query "tasks[0].containers[?name=='${container_name}'].exitCode" \
-    --output text
-)
-
-if [[ "${container_exit_code}" == "null" || "${container_exit_code}" != "0" ]]; then
```
chouinar: Wouldn't removing all of this mean we no longer error if migrations fail to run? How will be verify migrations ran successfully every run?

coilysiren: @chouinar this is still in draft, I added that back :)

coilysiren: kay -now- its ready

bin/run-command
```diff
@@ -119,87 +112,6 @@ task_arn=$(aws --no-cli-pager "${aws_args[@]}" --query "tasks[0].taskArn" --outp
...
-# Wait for log stream to be created before fetching the logs.
-# The reason we don't use the `aws ecs wait tasks-running` command is because
-# that command can fail on short-lived tasks. In particular, the command polls
-# every 6 seconds with describe-tasks until tasks[].lastStatus is RUNNING. A
-# task that completes quickly can go from PENDING to STOPPED, causing the wait
-# command to error out.
-echo "Waiting for log stream to be created"
-echo "  task_arn=${task_arn}"
-echo "  task_id=${ecs_task_id}"
-echo "  log_stream=${log_stream}"
```
chouinar, line 139: Could we leave these in (feel free to change the message) - useful for grabbing the task ID and then going to cloudwatch yourself.

coilysiren: 👍🏽

bin/run-command
```diff
@@ -119,87 +112,6 @@ task_arn=$(aws --no-cli-pager "${aws_args[@]}" --query "tasks[0].taskArn" --outp
...
-    logs_start_time_millis=$((last_log_timestamp + 1))
-  fi
-
-  # Give the application a moment to generate more logs before fetching again
-  sleep 1
-done
-echo "::endgroup::"
-echo
-
 container_exit_code=$(
```
chouinar, line 115: Wouldn't this logic break if we don't wait for the job to finish? There wouldn't be an exitcode if the job took more than a few moments to run.

Not sure if the exitCode query would also hit an issue if we have two containers now.

bin/run-command
coilysiren: https://github.com/HHS/simpler-grants-gov/pull/4305/commits/5b132fc3ef8f9740e07ed0e88ff4ab1eb7ca01e7

PR #4311, opened 2025-03-25
[Issue #4310] Further clean up test agency list

### LINKED ISSUE #4310: A few more test agencies
### Summary

Need to add FMCS and CNCS as test agencies in Prod

### Acceptance criteria

- [ ] Agencies no longer show in the Filter

mdragon:
## Summary
Fixes #4310

## Changes proposed
Add two agencies that are not current. Remove two agencies that we were hiding but just because they don't have grants right now which #3909 will address.


coilysiren (approved):

api/src/data_migration/transformation/subtask/transform_agency.py
mdragon, line 1: Didn't try to deal with top_level here, because we don't have it yet and as noted it'll be re-checked later. This will however mark the top level agencies correctly at this phase, so left the check in https://github.com/HHS/simpler-grants-gov/pull/4311/files#diff-8656b1fe3992d70128a672b27869210e326df602f8c099621050fba7237bf3efR242

api/src/data_migration/transformation/subtask/transform_agency.py
```diff
@@ -533,13 +533,13 @@ def apply_updates(
...
 
-    If the agency_code starts with one of the configured values, it is a test agency
+    If the top_level_agency_code matches one of the configured values, it is a test agency
     """
     for test_agency in agency_config.test_agency_config:
-        if agency_code.startswith(test_agency):
+        if top_level_agency_code == test_agency:
             return True
 
     return False
```
chouinar, line 545: Since we're back to doing equality checks and we're working with a set, can be real simple:
```
return top_level_agency_code in agency_config.test_agency_config
```

mdragon: Thanks! Changed.

chouinar (approved):

PR #4314, opened 2025-03-25
[Issue #4172] Add JSON schema validation to the PUT /applications/:appId/forms/:formId

### LINKED ISSUE #4172: Add JSON Schema validation to PUT /applications/:applicationId/forms/:formId
### Summary

After #4171 implemented the general structure of the endpoint, we want to validate the JSON Schema and return any validation issues. This should not cause any error as we're fine accepting partially filled in schemas. We'll instead return the warnings in the response object.

A rough little prototype of how validation works:
```python
import jsonschema

validator = jsonschema.Draft202012Validator(
    schema, format_checker=jsonschema.Draft202012Validator.FORMAT_CHECKER
)

for e in validator.iter_errors(user_answers):
     # do something
```

The validator is a rough container of the schema itself, and you can iterate over each of the errors (which we'll treat as warnings). I'd recommend making a utility for this that is reusable, as we'll want it later for submission (where they won't be warnings and will be actual errors).

As for how we process the errors, we want to populate the warnings objects as such:
* type: `e.validator`
* message: `e.message`
* field: `e.json_path` -> note that this will always look like `$.path.to.field` which I guess is fine?

For more details on the error object: https://python-jsonschema.readthedocs.io/en/stable/errors/

### Acceptance criteria

- [ ] Validation added
- [ ] Tests updated to check that validation runs and adds expected warnings

chouinar:
## Summary
Fixes #4172

## Changes proposed
Adds JSON schema validation utilities

Adds JSON schema validation to the user response for a given application

## Context for reviewers
Note that the validation doesn't prevent or in any way block a users answers from being filled out, it's just the current set of open issues. This is so if a user partially fills out a form, we'll note down what is still needed/incorrect, but not block them from saving their answers. It also allows the frontend to build out its structure in whatever way it wants (eg. a user could fill out a form across multiple pages).

When we eventually have a concept of submitting a form, at that point we would error if there are any outstanding issues.


api/src/form_schema/jsonschema_validator.py
```diff
@@ -0,0 +1,43 @@
+import jsonschema
+
+from src.api.response import ValidationErrorDetail
+from src.db.models.competition_models import Form
+
+
+def _get_validator(json_schema: dict) -> jsonschema.Draft202012Validator:
```
joshtonava, line 7: Does this throw if the json_schema is invalid?

api/tests/src/form_schema/test_jsonschema_validator.py
```diff
@@ -0,0 +1,133 @@
+import pytest
```
joshtonava, line 1: Can we do a test where the json_schema is invalid? Im curious to know what happens in that case

chouinar: Adding tests right now - will update PR soon

PR #4320, opened 2025-03-25
[no ticket] Only pass in cert once

coilysiren:
## Context

In a manifestation of pure chaos... we were grabbing and inputting the cert twice... 

This is obviously bad, so I have resolved this situation

## Caveat

I probably caused this situation 😅 The curse of being solo is that its easier to make silly mistakes such as this

### DIFF
--- a/infra/api/service/main.tf
+++ b/infra/api/service/main.tf
@@ -95,7 +95,7 @@ data "aws_rds_cluster" "db_cluster" {
 }
 
 data "aws_acm_certificate" "cert" {
-  count       = local.service_config.domain_name != null ? 1 : 0
+  count       = local.service_config.enable_https != null ? 1 : 0
   domain      = local.service_config.domain_name
   most_recent = true
 }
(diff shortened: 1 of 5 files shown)


infra/api/service/main.tf
```diff
@@ -95,8 +95,9 @@ data "aws_rds_cluster" "db_cluster" {
 }
 
 data "aws_acm_certificate" "cert" {
-  count  = local.service_config.domain_name != null ? 1 : 0
-  domain = local.service_config.domain_name
+  count       = local.service_config.domain_name != null ? 1 : 0
+  domain      = local.service_config.domain_name
+  most_recent = true
```
coilysiren, line 100: I just found `most_recent` while looking into something else

It's actually tangential to the main point of the PR, but... I figured since I'm messing with certs anyways

docs: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/acm_certificate#most_recent-1

mdragon (dismissed):
👍🏻

mdragon (dismissed):
Approved

mdragon (approved):
🚀

Issue #4317, opened 2025-03-25
Remove agencies with no open or forecasted opportunities from options list


doug-s-nava:
### Summary

In order to reduce user confusion over the extremely large and messy full list of agencies currently appearing in the agencies filter on prod, we will limit the list to only agencies that currently are associated with at least one opportunity with a status of "open" or "forecasted" (or "closed"? see below).

### What about closed

As a pre-requisite, run a query to get a sense of the number of agencies that have a currently open or forecasted opportunity, as well as the number when we add the closed status as well. If adding closed does not raise the number greatly or include a lot of agencies we want to filter out, we can add "closed" to the criteria

### Acceptance criteria

- [ ] as described above, determine if "closed" status should be included in the query
- [ ] either a new endpoint is built or existing agencies endpoint is updated to allow for returning a list of agencies that currently are associated with at least one opportunity with a status of "open" or "forecasted" (or "closed"? see above)
- [ ] frontend displays only agency options matching the list returned by this endpoint

PR #4321, opened 2025-03-25
[Issue #4317] Remove agencies with no open or forecasted opportunities from options list

babebe:
## Summary
Fixes #{[4317](https://github.com/HHS/simpler-grants-gov/issues/4317)}

## Changes proposed
The `get_agencies` endpoint has been updated to accept a filter parameter: `active=true`. When this filter is applied, it will return a list of sub-agencies and respective top-level agencies that are associated with at least one opportunity in either the `forecasted` or `posted` status. If the filter is not provided, the endpoint will return all agencies.


api/src/services/agencies_v1/get_agencies.py
```diff
@@ -26,15 +29,28 @@ class AgencyListParams(BaseModel):
...
     )
 
+    if active:
+        stmt = (
+            select(Agency)
+            .join(Opportunity, onclause=Agency.agency_code==Opportunity.agency_code)
+            .join(CurrentOpportunitySummary)
+            .join(LkOpportunityStatus)
+            .where(Agency.is_test_agency.isnot(True)) # Exclude test agencies
+            .where(LkOpportunityStatus.opportunity_status_id.in_([1,2])) # Opportunities associated with posted or forecasted status
```
chouinar, line 50: I assume this is still a work in progress, but you can make this a bit simpler of a query / use the enums directly:
```suggestion
            .join(CurrentOpportunitySummary)
            .where(Agency.is_test_agency.isnot(True)) # Exclude test agencies
            .where(CurrentOpportunitySummary.opportunity_status.in_([OpportunityStatus.FORECASTED,OpportunityStatus.POSTED])) # Opportunities associated with posted or forecasted status
```

babebe: Yup, draft query, updated!

api/src/api/agencies_v1/agency_routes.py
```diff
@@ -42,9 +44,17 @@
...
     list_params: AgencyListParams = AgencyListParams.model_validate(raw_list_params)
 
+    # Retrieve query parameter
+    active_str = request.args.get("active", None)
+    active: bool
+    if active_str in ["True", "true"]:
+        active = True
+    else:
+        active = False
+
```
chouinar, line 54: Why not use the request models we already have? Why create a custom parser here instead of just using the Marshmallow schema on the endpoint?

"active" feels like something that is reasonably a filter that a user would want to pass in.
```suggestion

```

api/src/services/agencies_v1/get_agencies.py
```diff
@@ -26,7 +28,7 @@ class AgencyListParams(BaseModel):
 
 
 def get_agencies(
-    db_session: db.Session, list_params: AgencyListParams
+    db_session: db.Session, list_params: AgencyListParams, active: bool | None = None
```
chouinar, line 31: Why isn't `active` a part of the agency params request object?

api/src/services/agencies_v1/get_agencies.py
```diff
@@ -35,6 +37,20 @@ def get_agencies(
...
+            .join(Opportunity, onclause=Agency.agency_code == Opportunity.agency_code)
+            .join(CurrentOpportunitySummary)
+            .where(Agency.is_test_agency.isnot(True))  # Exclude test agencies
+            .where(
+                CurrentOpportunitySummary.opportunity_status.in_(
+                    [OpportunityStatus.FORECASTED, OpportunityStatus.POSTED]
+                )
+            )
+            .options(joinedload(Agency.top_level_agency), joinedload("*"))
+        )
```
chouinar, line 52: Don't create a new stmt, extend the one that already exists. That way any default rules we want for filters don't need to be implemented twice.
```suggestion
    if active:
        stmt = (
            stmt
            .join(Opportunity, onclause=Agency.agency_code == Opportunity.agency_code)
            .join(CurrentOpportunitySummary)
            .where(
                CurrentOpportunitySummary.opportunity_status.in_(
                    [OpportunityStatus.FORECASTED, OpportunityStatus.POSTED]
                )
            )
        )
```

I'm also not certain, but it might be worth considering a query like this instead (no idea which is better performing, could run some `EXPLAIN ANALYZE <query>` in dev/staging to test)

```sql
SELECT
	*
FROM
	api.agency a
WHERE
	is_test_agency IS FALSE
	AND a.agency_code in(
		SELECT
			agency_code FROM api.opportunity o
			JOIN api.current_opportunity_summary cos ON o.opportunity_id = cos.opportunity_id
		WHERE
			opportunity_summary_id in(1, 2))
```

api/tests/src/api/agencies_v1/test_agencies_routes.py
```diff
@@ -111,3 +119,56 @@ def test_agencies_sorting(self, client, api_auth_token, enable_factory_create, d
...
+            opportunity=opp_3, opportunity_status=OpportunityStatus.FORECASTED
+        )
+
+        # DARPA
+        opp_4 = OpportunityFactory.create(
+            agency_code=darpa.agency_code, current_opportunity_summary=None
+        )
+        CurrentOpportunitySummaryFactory.create(
+            opportunity=opp_4, opportunity_status=OpportunityStatus.ARCHIVED
+        )
```
chouinar, line 153: Something I setup to make this a lot easier, you don't need to create them separate, there's a param you can pass to the factory to make the opportunity summary a certain status:
```suggestion
        opp_4 = OpportunityFactory.create(
            agency_code=darpa.agency_code, is_archived_non_forecast_summary=True
        )
```
There is one for each of the statuses

doug-s-nava (dismissed):
can't speak to the implementation details, but as far as I can test locally, the changes are doing what they're supposed to

✅ ✅ ✅ ✅

api/tests/src/api/agencies_v1/test_agencies_routes.py
babebe: Nice! updated

api/src/services/agencies_v1/get_agencies.py
```diff
@@ -17,13 +19,47 @@
...
+                CurrentOpportunitySummary.opportunity_status.in_(
+                    [OpportunityStatus.FORECASTED, OpportunityStatus.POSTED]
+                )
+            )
+        )
+        .union(active_agency)
+        .subquery()
+    )
+
+    agency_id_stmt = select(active_top_level_agency).distinct
```
chouinar, line 59: Wouldn't this need to be `.distinct()` - does it work if you don't call the function?

api/src/services/agencies_v1/get_agencies.py
```diff
@@ -17,13 +19,47 @@
...
+            .where(Agency.is_test_agency.isnot(True))  # Exclude test agencies
+            .where(
+                CurrentOpportunitySummary.opportunity_status.in_(
+                    [OpportunityStatus.FORECASTED, OpportunityStatus.POSTED]
+                )
+            )
+        )
+        .union(active_agency)
+        .subquery()
+    )
```
chouinar, line 57: Here is one way we can shorten this a bit:

```
# Note - I don't know exactly what type field would be here, select takes in like a dozen, something column related
def _construct_active_inner_query(field: ???) -> Select:
     """Construct a query of the form:
      
     SELECT < field >
     FROM
	agency
	JOIN opportunity ON agency.agency_code = opportunity.agency_code
	JOIN current_opportunity_summary ON opportunity.opportunity_id = current_opportunity_summary.opportunity_id
      WHERE
	agency.is_test_agency IS NOT TRUE
	AND current_opportunity_summary.opportunity_status_id in(1, 2)
     """
     return select(field)
        .join(Opportunity, onclause=Agency.agency_code == Opportunity.agency_code)
        .join(CurrentOpportunitySummary)
        .where(Agency.is_test_agency.isnot(True))  # Exclude test agencies
        .where(
            CurrentOpportunitySummary.opportunity_status.in_(
                [OpportunityStatus.FORECASTED, OpportunityStatus.POSTED]
            )
        )
    )

def _construct_active_query(stmt: Select) -> Select:
     active_agency_subquery = _construct_active_inner_query(Agency.agency_id).union(_construct_active_inner_query(Agency.top_level_agency_id)).subquery()
```

api/src/api/agencies_v1/agency_routes.py
babebe: Now a filter

api/src/services/agencies_v1/get_agencies.py
babebe: adapted new query

chouinar (approved):

PR #4323, opened 2025-03-25
[no ticket] Adds fluentbit itselfs logs into cloudwatch

coilysiren:
## Context

This is a little confusing to bear with me.

Right now, AWS only allows a given container to allow logs to output to one location. And we set it up so that the primary application container logs => fluentbit => New Relic. Okay.

Well, guess what! fluentbit itself has logs as well! And those logs aren't being sent anywhere. So we can send those to cloudwatch.

## Testing

TODO


mdragon (approved):
👍🏻

PR #4328, opened 2025-03-26
[no ticket] Show how to test every deploy

coilysiren:
## Context

I'm going to start writing down every command I use before I run them


mdragon (approved):
👍🏻

PR #4338, opened 2025-03-26
[Issue #4263] Need to know if the service is HTTPS not the local container

### LINKED ISSUE #4263: Forming Login.gov redirect is un-aware we're behind HTTPs in Prod
### Summary

When formulating the URL argument for the redirect URL to pass to Login.gov, we get the full URL of the route from Flask. But since Flask is running HTTP, but is fronted by HTTPS at the load balancer, it formulates an http URL which doesn't match the allow list in Login.gov (nor would we want to do a non-secure redirect back). 
https://github.com/HHS/simpler-grants-gov/blob/e785ee9bce3dbbaf55dd93238fb10d5fe2d6ba7a/api/src/auth/login_gov_jwt_auth.py#L134 
Adding the missing http(s) in the Login.gov URL allows successful login. We haven't seen this before because the lower environments don't run HTTPS while we hashed out the method of obtaining the certificates for all environments.

### Reproducibility

### How to reproduce the bug

1. In Simpler Prod turn on the Auth Feature Flag
2. Click the Login link
3. Click the button to go to Login.gov
4. Observe the error message about the redirect_url being invalid
5. Add the missing s on the URL param for redirect and hit enter
6. Login successfully

### Browser

Not applicable

### Code of Conduct

mdragon:
## Summary
Fixes #4263

I'm inferring that var.enable_https is if the local container server should be running HTTPS (false in our apps). So I think I might have found the right higher level config that we should really be checking local.service_config.enable_https

## Changes proposed
Check the correct enable_https

### DIFF
--- a/infra/api/app-config/env-config/environment_variables.tf
+++ b/infra/api/app-config/env-config/environment_variables.tf
@@ -13,13 +13,14 @@ locals {
     # Login.gov OAuth
     # Default values point to the IDP integration environment
     # which all non-prod environments should use
+    # local.service_config not var because we want if the service is HTTPS not if this endpoint is HTTPS
     ENABLE_AUTH_ENDPOINT      = 0
     LOGIN_GOV_CLIENT_ID       = "urn:gov:gsa:openidconnect.profiles:sp:sso:hhs-${var.environment}-simpler-grants-gov"
     LOGIN_GOV_ENDPOINT        = "https://idp.int.identitysandbox.gov/"
     LOGIN_GOV_JWK_ENDPOINT    = "https://idp.int.identitysandbox.gov/api/openid_connect/certs"
     LOGIN_GOV_AUTH_ENDPOINT   = "https://idp.int.identitysandbox.gov/openid_connect/authorize"
     LOGIN_GOV_TOKEN_ENDPOINT  = "https://idp.int.identitysandbox.gov/api/openid_connect/token"
-    LOGIN_GOV_REDIRECT_SCHEME = var.enable_https ? "https" : "http"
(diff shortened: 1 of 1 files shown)


infra/api/app-config/env-config/environment_variables.tf
```diff
@@ -13,13 +13,14 @@ locals {
...
     # which all non-prod environments should use
+    # local.service_config not var because we want if the service is HTTPS not if this endpoint is HTTPS
     ENABLE_AUTH_ENDPOINT      = 0
     LOGIN_GOV_CLIENT_ID       = "urn:gov:gsa:openidconnect.profiles:sp:sso:hhs-${var.environment}-simpler-grants-gov"
     LOGIN_GOV_ENDPOINT        = "https://idp.int.identitysandbox.gov/"
     LOGIN_GOV_JWK_ENDPOINT    = "https://idp.int.identitysandbox.gov/api/openid_connect/certs"
     LOGIN_GOV_AUTH_ENDPOINT   = "https://idp.int.identitysandbox.gov/openid_connect/authorize"
     LOGIN_GOV_TOKEN_ENDPOINT  = "https://idp.int.identitysandbox.gov/api/openid_connect/token"
-    LOGIN_GOV_REDIRECT_SCHEME = var.enable_https ? "https" : "http"
+    LOGIN_GOV_REDIRECT_SCHEME = local.service_config.enable_https ? "https" : "http"
```
mdragon, line 23: Maybe fixes the issue with this being set to http in Prod still

coilysiren (approved):
🤞🏽

PR #4350, opened 2025-03-27
[Issue #4264] Create util / tests for dat to JSONSchema conversion

### LINKED ISSUE #4264: Create a utility that given a DAT file for a given form, can produce a rough JSON Schema object
### Summary

This utility can be self-contained in a function (we can hook it up to CLI processes later), and can be put in the form_schema folder with some of our other utilities. The logic to build a JSON Schema already exists in the [JSON Schema Builder](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/form_schema/jsonschema_builder.py) - this ticket mostly aims to parse a file, and use this builder to generate a JSON Schema output.

The file we will be parsing (and that you can use for some rough testing) can be found at https://grants.gov/forms/form-items-description/fid/668 - by downloading the DAT file.

Some notes:
* The JSON Schema builder will need to be modified to allow a `title` field to be passed into each of the different methods. This title isn't used for anything in the validation, but the frontend will need it.
* The tool you build should be flexible, and allow for us to customize it a bit more later - for example, there is a column `Global Library Field Names` that has a weird globLib type defined for certain things - in the future we may want to use that to know something is a particular type or group (eg. something has all the address globLibs, make an address object instead). Don't worry about implementing that now, but keep it in mind (check with Michael if this doesn't make sense, can explain better)
* This file isn't quite a CSV, you might need to do some stuff to properly parse it since the first row isn't the column headers.
* There will be things that are complex/impossible to parse, the goal is to try and get the fields into the file, but some things might just be too weirdly defined to parse properly - if so, sidelines those lines and log them out at the end. We don't expect this process to 100% convert files, we just want to automate as much as we can.
* If the `Field Implementation` column is `Label` we can _probably_ skip it

Some rough rules about how to populate json schema values from this:
* If the "Field Implementation" value is
* The field name comes from `Agency Field Name`
* Title comes from `Field Label`
* Types are... iffy - if there isn't a great way to determine it, fallback to a string.
* Validation Rules
   * Enumerations (or allowed values) are in the `List of Values` column - for some of them it gives an actual value, and for things like state/country, it doesn't. For those, make a static list with a few values (we'll get a full list later)
   * min/max length of values have a column
   * Whether or not a field is required is based on the `Required?` column
   * Whether or not a field is nullable is the `Minimum Occurrences` column - if the value is 0, it's nullable

### Acceptance criteria

- [ ] Rough utility built
- [ ] Some rough examples of JSON schema generated for a few files added

mikehgrantsgov:
## Summary
Fixes #4264

## Changes proposed
- Add CSV parser to JSONSchema
- Add util to convert XLS to CSV
- Parse these fields based on metadata from DAT entries.
- Add pandas to project for XLS parsing

## Context for reviewers
We parse multiple DAT files to test:
https://grants.gov/forms/form-items-description/fid/668
https://grants.gov/forms/form-items-description/fid/723

We either skip or set unknown data types to `string` (skipped types are defined in an array in code)

## Additional information
poetry.lock - how should this be updated?
This will eventually be rolled into a CLI tool, and will change frequently over time. Not by any means a final implementation.


api/src/form_schema/csv_to_jsonschema.py
```diff
@@ -0,0 +1,182 @@
...
+        csv_file_path: Path to the CSV file containing form field definitions
+
+    Returns:
+        A dictionary representing the JSON Schema
+    """
+    # Initialize the main schema builder
+    schema_builder = JsonSchemaBuilder()
+
+    # Group fields by section for better organization
+    sections: dict[str, list[dict[str, Any]]] = {}
```
chouinar, line 25: We do not need to solve it now, but a follow-up we'll want to figure out is how to handle "common types".

For example, addresses get repeated all over schemas. In terms of how we'll actually want to structure the data, it'll look like:

```
{
 "WhateverAddressField": {
    "Street1": "...",
    "Street2": "...",
    "City": "...",
    "County": "...",
    "State": "...",
    "Province": "...",
    "ZipPostalCode": "...",
    "Country": "..."
  }
}
```
If we see this, instead of individually adding each address field, we should just add an address object (and in the very far future, we'll just add a $ref to a central json schema file)

I'm not sure how well we can parse that from the .dat files - we can probably recognize that all the parts of an address are present, but what the "overall address field name" is, I don't see in that file (the XML schema says its called `AuthorizedRepresentativeAddress`)

api/src/form_schema/dat_to_jsonschema.py
```diff
@@ -0,0 +1,20 @@
+from typing import Any
+
+from src.form_schema.csv_to_jsonschema import csv_to_jsonschema
+from src.util.xls_to_csv import xls_to_csv
+
+
+def parse_xls_to_schema(file_path: str) -> dict[str, Any]:
```
chouinar, line 7: Can we add a way to run this from the CLI? Doesn't need to be a full backend ECS task, just something simple for local use to pass in a file path and get an output.

mikehgrantsgov: Added and documented a mechanism e.g.:
make dat-to-jsonschema args="tests/src/form_schema/SF424_Individual_2_0-V2.0_F668.xls"

api/src/form_schema/csv_to_jsonschema.py
```diff
@@ -0,0 +1,182 @@
...
+    """Add a field to the appropriate JsonSchemaBuilder based on its type."""
+    field_id = field_info["id"]
+    required = field_info["required"]
+    data_type = field_info["data_type"]
+    field_type = field_info["type"]
+    min_value = field_info["min_value"]
+    max_value = field_info["max_value"]
+    list_of_values = field_info["list_of_values"]
+    help_tip = field_info["help_tip"]
+    is_nullable = field_info["is_nullable"]
```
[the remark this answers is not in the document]
mikehgrantsgov: Added a FieldInfo class and encapsulated some of the parsing there

api/src/form_schema/csv_to_jsonschema.py
```diff
@@ -0,0 +1,151 @@
...
+
+        # Add the section as a sub-object to the main schema
+        schema_builder.add_sub_object(section_key, False, section_builder)
+
+    return schema_builder.build()
+
+
+def add_field_to_builder(builder: JsonSchemaBuilder, field_info: FieldInfo) -> None:
+    """Add a field to the appropriate JsonSchemaBuilder based on its type."""
+    field_id = field_info.id
```
chouinar, line 75: Instead of using field ID for the name of the field, can we use "Agency Field Name" - it better aligns with what the XML/pdfs call the fields internally (eg. `FederalAgency` instead of `Name of Federal Agency`)

At least quickly skimming the file, if something doesn't have an agency field name, it seems to not be an actual field and is instead either info in the doc OR a section header.

api/src/form_schema/field_info.py
```diff
@@ -0,0 +1,83 @@
...
+        cleaned_value: str = value.strip().replace("$", "").replace(",", "").replace(" ", "")
+
+        try:
+            # Try converting to int first
+            num_value = int(cleaned_value)
+            return num_value
+        except ValueError:
+            # Log the error or handle it as needed
+            logger.warning(f"Warning: Could not convert '{value}' to a numeric value")
+            return None
```
chouinar, line 46: When running locally, I get a lot of warnings for `Warning: Could not convert '1.0' to a numeric value` - seems like the numbers are stored as floats but we aren't converting them to ints as we'd expect.

Apparently `int("1.0")` isn't valid, so we'll have to do something like `int(Decimal("1.0"))`

api/src/form_schema/dat_to_jsonschema_cli.py
```diff
@@ -0,0 +1,49 @@
...
+        "--skip-rows",
+        help="Number of rows to skip from the top of the sheet (default is 2)",
+        type=int,
+        default=2,
+    )
+
+    return parser.parse_args(args)
+
+
+def main(args: list[str] | None = None) -> int:
```
chouinar, line 33: Two notes:
1. I'd recommend initializing our logging just to get nice logs - just need to wrap everything in a `with src.logging.init("dat_to_jsonschema_cli"):` call
2. It doesn't matter a ton, but `click` tends to be a bit simpler to setup CLI commands than argparse

You can see both of these in use on https://github.com/HHS/simpler-grants-gov/blob/main/api/tests/lib/seed_local_db.py#L53

api/src/form_schema/dat_to_jsonschema_cli.py
```diff
@@ -0,0 +1,49 @@
...
+    parsed_args = parse_args(args)
+
+    try:
+        result = parse_xls_to_schema(
+            file_path=parsed_args.input_file,
+            sheet_index=parsed_args.sheet,
+            skip_rows=parsed_args.skip_rows,
+        )
+
+        print(result)
```
chouinar, line 44: Could we pretty-print this? Something like:
```
json_result = json.loads(result)
print(json.dumps(json_result, indent=4))
```

api/src/form_schema/csv_to_jsonschema.py
```diff
@@ -0,0 +1,151 @@
...
+    title = field_info.label
+
+    # Determine field format
+    format_value = None
+    if data_type == "DATE":
+        format_value = "date"
+    elif "email" in field_id.lower():
+        format_value = "email"
+
+    skipped_types = ["FILE"]
```
chouinar, line 95: I think we also want to skip button?

At least on the SF424, the only button is a pop-up that gives a bunch of info, but isn't itself a field you'd fill out.

api/src/form_schema/csv_to_jsonschema.py
```diff
@@ -0,0 +1,151 @@
...
+
+    return schema_builder.build()
+
+
+def add_field_to_builder(builder: JsonSchemaBuilder, field_info: FieldInfo) -> None:
+    """Add a field to the appropriate JsonSchemaBuilder based on its type."""
+    field_id = field_info.id
+    required = field_info.required
+    data_type = field_info.data_type
+    field_type = field_info.type
```
chouinar, line 78: I think the mapping of where these come from and how they're used is off.
Field Type is mapped from the field type column that just stores values like `Agency-specific` or `Pre-populated`, while the actual type of the column is the `Field Implementation` column.

api/src/form_schema/csv_to_jsonschema.py
```diff
@@ -0,0 +1,151 @@
...
+            description=help_tip,
+            min_length=min_value,
+            max_length=max_value,
+            format=format_value,
+        )
+
+    # Add description or help text to the property
+    if help_tip:
+        # We need to manually add description since it's not part of the builder API
+        builder.properties[field_id]["description"] = help_tip
```
chouinar, line 151: Is this needed if the description is passed in above?

api/src/form_schema/csv_to_jsonschema.py
mikehgrantsgov: Made sure to add this and update tests.

api/src/form_schema/dat_to_jsonschema_cli.py
mikehgrantsgov: Didn't know about Click, good idea. Moved over to this.

api/src/form_schema/dat_to_jsonschema_cli.py
mikehgrantsgov: Added this.

api/src/form_schema/csv_to_jsonschema.py
mikehgrantsgov: I think we are skipping this Button anyway since the "Agency Field Name" is empty, but I added some additional checks to make sure we skip "Field Implementation"

api/src/form_schema/csv_to_jsonschema.py
mikehgrantsgov: Ah, removed that.

api/src/form_schema/csv_to_jsonschema.py
mikehgrantsgov: I think this most recent update solved it? Maybe we can pair on this if something is still off to clarify my understanding.

api/src/form_schema/csv_to_jsonschema.py
```diff
@@ -0,0 +1,156 @@
...
+
+def add_field_to_builder(builder: JsonSchemaBuilder, field_info: FieldInfo) -> None:
+    """Add a field to the appropriate JsonSchemaBuilder based on its type."""
+    field_id = field_info.id
+    required = field_info.required
+    data_type = field_info.data_type
+    field_type = field_info.type
+    min_value = field_info.min_value
+    max_value = field_info.max_value
+    list_of_values = field_info.list_of_values
```
chouinar, line 91: We can follow-up on it, but some enums don't have their actual values in the files we process. For example, the only value for country is `"GENC Standard Ed3.0 Update 11"` which you can see under https://apply07.grants.gov/apply/system/schemas/UniversalCodes-V2.0.xsd in the `CountryCodeDataTypeV3` section (US states are also present).

We'll want to handle swapping those out next.

api/src/form_schema/csv_to_jsonschema.py
```diff
@@ -0,0 +1,156 @@
...
+    # Initialize the main schema builder
+    schema_builder = JsonSchemaBuilder()
+
+    # Group fields by section for better organization
+    sections: dict[str, list[FieldInfo]] = {}
+    current_section = None
+
+    reader = csv.DictReader(csv_content.splitlines())
+
+    skipped_field_implementations = ["button"]
```
chouinar, line 31: Could we define a few of these static values at the top of the file just to put all of them together?

api/src/form_schema/dat_to_jsonschema_cli.py
```diff
@@ -0,0 +1,45 @@
...
+                sheet_index=sheet,
+                skip_rows=skip_rows,
+            )
+
+            print(json.dumps(result, indent=4, separators=(",", ": ")))
+            return 0
+
+        except Exception as e:
+            print(f"Error: {str(e)}", file=sys.stderr)
+            return 1
```
chouinar, line 45: Could do something like this and it'll dump the stack trace for you.
```suggestion
            logger.exception("Process failed")
            raise
```

chouinar (commented):
Just a few small tidying up suggestions, but the output looks good - I need to create a few follow-up tickets for next things

api/src/form_schema/csv_to_jsonschema.py
mikehgrantsgov: What other types should we have? I added `radio`, but not sure if that's right? I couldn't find a list of possible values we'd want to skip in the schema.

api/src/form_schema/dat_to_jsonschema_cli.py
```diff
@@ -29,7 +29,8 @@
 def main(input_file: str, sheet: int, skip_rows: int) -> int:
     """Main entry point for the script."""
 
-    with logging.init("dat_to_jsonschema_cli"):
+    with src_logging.init("dat_to_jsonschema_cli"):
+        logger = logging.getLogger(__name__)
```
chouinar, line 33: Because it makes the imports weird as you noticed, I'd recommend just importing the logging init as `import src.logging` and using the full path here of `src.loggin.init(...)`

We can also define the logger at the top of the file, it doesn't need to be after we initialize logging (initializing logging changes its behavior globally retroactively)

mikehgrantsgov: Something like the most recent commit?

chouinar (approved):

PR #4361, opened 2025-03-27
[Issue #4010] Enable attachment load+transform by default

### LINKED ISSUE #4010: Make attachment loading job part of normal automated hourly data loading process
### Summary

\

### Acceptance criteria

\

chouinar:
## Summary
Fixes #4010

## Changes proposed
Make the attachment load (Oracle -> our staging table) and transform (staging -> API table) processes on by default

Remove Oracle data copy for tables we no longer use in transformations (historical tables)

## Context for reviewers
Enabling the load + transform is just two config options, one to add the table to the load config, and the other to set the default env var for attachments.

I also cleaned up copy the historical tables over at all as we no longer transform those tables and that's just a waste to continue to copy data we don't plan to use (we should clean those up even more like removing the tables, but that's definitely a much bigger change).

Otherwise just some test additions and fixes.

### DIFF
--- a/api/tests/src/db/models/factories.py
+++ b/api/tests/src/db/models/factories.py
@@ -1333,10 +1333,10 @@ class TsynopsisAttachmentFactory(BaseFactory):
     class Meta:
         abstract = True
 
-    syn_att_id: factory.Sequence(lambda n: n)
-    opportunity_id: factory.Sequence(lambda n: n)
+    syn_att_id = factory.Sequence(lambda n: n)
+    opportunity_id = factory.Sequence(lambda n: n)
     att_revision_number = factory.Faker("random_int", min=1000, max=10000000)
-    att_type: factory.Faker("att_type")
+    att_type = factory.Faker("word")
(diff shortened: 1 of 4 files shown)


api/tests/src/db/models/factories.py
```diff
@@ -1723,6 +1723,14 @@ class StagingTsynopsisAttachmentFactory(TsynopsisAttachmentFactory, AbstractStag
     class Meta:
         model = staging.attachment.TsynopsisAttachment
 
+    @classmethod
+    def _setup_next_sequence(cls):
```
joshtonava, line 1727: For my own notes https://factoryboy.readthedocs.io/en/stable/reference.html#factory.Factory._setup_next_sequence

> This method will compute the first value to use for the sequence counter of this factory.
> It is called when the first instance of the factory (or one of its subclasses) is created.
> Subclasses may fetch the next free ID from the database, for instance.

joshtonava (approved):
👍

PR #4367, opened 2025-03-27
[Issue #4060] documents UI tuning

### LINKED ISSUE #4060: Remove button "Download Full Announcement (NOFO)"
### Summary

As we move from the attachment prototype we built during Quad 1 to a solution that scales to support all opportunity attachments, we need to revisit the UX surrounding attachment downloads.

Note: for our current purposes "attachments" refers only to files in the Related Documents tab on existing Grants.gov. We don't currently support downloading forms (i.e. files in the "Package" tab).

Users need a way to download all attachments as a .zip bundle in a single click
The majority of opportunities have one attachment. Even if there's just one attachment, we should still enable users to download it as a .zip.
The [current design](https://www.figma.com/design/FGxWtAgToKhehLJCiuy1zL/Simpler.Grants.gov-Project?node-id=3296-29235&t=01er8dEr6SH723A2-0) uses the term "Download Full Announcement (NOFO)" for the download button. We're revisiting this terminology in https://github.com/HHS/simpler-grants-gov/issues/3048.

**Decisions Made:**

remove button "Download Full Announcement (NOFO)"
If there are documents, then show the "jump to documents" link
update document table to show description column
add "Download all" button above table to download zip file of all documents
if there is only one document, then hide the "download documents" button
remove "Category" column from document table
table will only show "related documents" not application forms

Updated Figma file: https://www.figma.com/design/FGxWtAgToKhehLJCiuy1zL/Simpler.Grants.gov-Project?node-id=3296-29111&t=dh2MFQxGjWDRY2tl-1

### Acceptance criteria

Input from frontend needed

doug-s-nava:
## Summary
Fixes #4060

## Changes proposed
* remove "nofo download" button
* add description column in opportunity detail page documents table
* make documents table full width

### Test steps
2. visit an opportunity page with 1 attachment (should be 7) http://localhost:3000/opportunity/7
3. _VERIFY_: there is no download nofo button
4. _VERIFY_: documents table contains a description column

## Additional information

Screenshot 2025-03-27 at 15-28-48 Opportunity Listing - Research into Jewellery designer industry

### DIFF
--- a/frontend/src/components/opportunity/OpportunityDownload.tsx
+++ b/frontend/src/components/opportunity/OpportunityDownload.tsx
@@ -1,37 +1,21 @@
 "use client";
 
-import { useTranslations } from "next-intl";
-import { Button, Link } from "@trussworks/react-uswds";
+import { OpportunityDocument } from "src/types/opportunity/opportunityResponseTypes";
 
-import { USWDSIcon } from "src/components/USWDSIcon";
+import { useTranslations } from "next-intl";
+import { Link } from "@trussworks/react-uswds";
 
 type OpportunityDownloadProps = {
-  nofoPath: string;
-  opportunityId: number;
+  attachments: OpportunityDocument[];
 };
 
-const downloadNOFO = (nofoPath: string) => {
-  window.open(nofoPath, "_blank");
-};
-
-const OpportunityDownload = ({
-  nofoPath,
-  opportunityId,
-}: OpportunityDownloadProps) => {
+const OpportunityDownload = ({ attachments }: OpportunityDownloadProps) => {
   const t = useTranslations("OpportunityListing.description");
 
   return (
     <>
-      {nofoPath.length > 0 ? (
+      {attachments.length > 0 ? (
         <div className="grid-row flex-justify">
-          <Button
-            onClick={() => downloadNOFO(nofoPath)}
-            type="button"
-            id={`opportunity-document-button-${opportunityId}`}
-          >
-            <span>{t("nofo_download")} </span>
-            <USWDSIcon name={"file_download"} className="usa-icon--size-4" />
-          </Button>
           <Link
             className="flex-align-self-center"
             href={"#opportunity_documents"}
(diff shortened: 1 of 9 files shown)


frontend/src/components/opportunity/OpportunityDownload.tsx
```diff
@@ -1,37 +1,21 @@
...
-  nofoPath,
-  opportunityId,
-}: OpportunityDownloadProps) => {
+const OpportunityDownload = ({ attachments }: OpportunityDownloadProps) => {
   const t = useTranslations("OpportunityListing.description");
 
   return (
     <>
-      {nofoPath.length > 0 ? (
+      {attachments.length > 0 ? (
```
andycochran, line 17: (@doug-s-nava don't let this prevent merging. I just wanted to raise a Q for design.)

@crystabelrangel, should the "jump to" link be disabled instead of nonexistent when there are no attachments? I'm wondering what the preferred UI would be. Is it more confusing for the link to not be there or not be interactive? Should we consider another way to indicate at a glance that there are no documents?

acouch (approved):
Two nits:

It looks like the "Grantor contact information" and "Documents" should be the same size according to the design. In the PR the "Grantor" h2 has `.usa-prose > h2 ` applied to it, so it is `.5rem` larger. Documents in the design have a weight of `800` vs `700`:

image


Also notice the width of each column are not the same like the design:
image

andycochran:
> width of each column are not the same like the design

I doubt the column widths in the mockup were explicit design decisions. I think it's okay if the table column widths are determined by the content. It might make sense to put a no-wrap class on the table headings (e.g. so "File name" doesn't wrap).

andycochran:
For the heading sizes a couple thoughts: 
- In the last design sync we talked about how the sizes were inappropriately nested. "Documents" should appear as under "Grantor contact information" (it should prob be the same level). cc @crystabelrangel 
- Using `.usa-prose` can be difficult. Maybe we should strip it out of our app? It's really intended for content where you don't control the markup or can't add classes manually (e.g. blog posts or content managed elsewhere).

doug-s-nava:
@andycochran @acouch can we add these changes to https://github.com/HHS/simpler-grants-gov/issues/4410 and we can take care of them next sprint?

PR #4383, opened 2025-03-31
[Issue #4355] Adjust logging in LoadOracleDataTask to use less metrics

### LINKED ISSUE #4355: Adjust the logging in the LoadOracleDataTask to not put everything in metrics
### Summary

In our task classes we collect metrics (counts and such) that at the end we log out. This is great for making dashboards and alerts off of. However, we put too much into the metrics in this particular case and the logs are too big to fit into New Relic.

We still want all of this info, we just can't put it all in a big "mega log" at the end.

So, roughly we want to do the following:
* Anywhere we do `self.increment` or `self.set_metrics` for specific table metrics (eg. a line like `self.increment(f"count.insert.{staging_table.name}", total_insert_count)`) instead log a message on that line. These log messages should be kept pretty generic like `inserted records into table` and anything like the table name should be in the log extra object.
* Leave any metrics for stuff like total counts alone (eg. `self.increment("count.insert.total", total_insert_count)` leave alone) - we still want overall metrics.
* Several tests check the metrics object - we'll need to adjust that

### Acceptance criteria

- [ ] Logs adjusted
- [ ] Tests updated

chouinar:
## Summary
Fixes #4355

## Changes proposed
Updated logging in LoadOracleDataTask to not put it all in metrics / log everything at once

## Context for reviewers
New Relic has a limit of ~255 key/value pairs in a log statement. Right now we're putting about 200+ values into the metric object just for this one task - which is causing it to break when combined with the ~50 or so we always add. To work around this, instead of adding all these values as we go, just log a bunch of them as we go. Only keep metrics for the overall counts, and just log insert/update/delete metrics when we calculate them. Still have all the same info, just spread out a bit.

### DIFF
--- a/api/src/data_migration/load/load_oracle_data_task.py
+++ b/api/src/data_migration/load/load_oracle_data_task.py
@@ -122,7 +122,7 @@ class LoadOracleDataTask(src.task.task.Task):
 
     def do_insert(self, foreign_table: sqlalchemy.Table, staging_table: sqlalchemy.Table) -> int:
         """Determine new rows by primary key, and copy them into the staging table."""
-        log_extra = {"table": foreign_table.name}
+        log_extra: dict = {"table": foreign_table.name}
 
         logger.info("Fetching records to be inserted", extra=log_extra)
         select_sql = sql.build_select_new_rows_sql(foreign_table, staging_table)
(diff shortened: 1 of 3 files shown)


api/src/data_migration/load/load_oracle_data_task.py
```diff
@@ -154,19 +154,19 @@ def do_insert(self, foreign_table: sqlalchemy.Table, staging_table: sqlalchemy.T
...
-                f"count.insert.chunk.{staging_table.name}": ",".join(map(str, insert_chunk_count)),
-                f"time.insert.{staging_table.name}": round(t1 - t0, 3),
-            }
-        )
+
+        log_extra |= {
+            f"count.insert.{staging_table.name}": total_insert_count,
+            f"count.insert.chunk.{staging_table.name}": ",".join(map(str, insert_chunk_count)),
+            f"time.insert.{staging_table.name}": round(t1 - t0, 3),
+        }
```
coilysiren, line 162: Have you ever used true metrics? You're using logs here, and describing them as "metrics" in the PR title. But I'm pretty sure what you actually want are true metrics:

https://docs.newrelic.com/docs/apm/agents/python-agent/supported-features/python-custom-metrics/

chouinar: I have used metrics before - the metrics we collect here follow a pattern from a past project where we did actually do that - it might be something we want to consider, but that requires some separate setup to do. Additionally, I found just having everything in "Logs" makes things easier rather than having to understand that your data lives in multiple places in New Relic.

coilysiren (approved):

PR #4387, opened 2025-03-31
[Unticketed] Fix agency transform to not break for is_deleted rows

chouinar:
## Changes proposed
Modify the transform agency job to not error when trying to transform a row that is deleted that we also don't want to process

## Context for reviewers
Before this change, the job would fail to update/insert an agency if there was a row marked as `is_deleted=True` that also was one that we didn't even care to map. This fixes that by doing the mapping check first before any other processing, not at the end.


joshtonava (approved):
Should prevent jobs from failing for deleted agencies considering we now have more fields than were originally in place when these agencies were loaded, and need to explicitly ignore them.

PR #4389, opened 2025-03-31
[Issue #4388] Implement Column Exclusion Feature for Oracle Data

### LINKED ISSUE #4388: Implement Column Exclusion Feature for Oracle Data
### Summary

Discovered while implementing #4382

We need to enhance our Oracle to PostgreSQL data migration process to handle columns with incompatible data types, particularly encrypted columns (like `tuser_account.email_address`) that cause conversion errors. The current implementation attempts to copy all columns, causing query errors when Oracle data types can't be converted to PostgreSQL types.

Implement a column exclusion mechanism that allows specific columns to be omitted from the data transfer, preventing type conversion errors while still migrating the rest of the data.

### Acceptance criteria

[ ] LoadOracleDataTask updated to exclude columns
[ ] SQL builder functions correctly modified
[ ] Unit tests verify columns are properly excluded during migration

mikehgrantsgov:
## Summary
Fixes #4388

## Changes proposed
Set explicit columns to exclude loading oracle data
Configure SQL builder to accommodate skipped columns
Add unit tests

## Context for reviewers
Discovered while checking user data ported from Oracle.


api/src/data_migration/load/load_oracle_data_task.py
```diff
@@ -253,8 +271,14 @@ def main() -> None:
         foreign_tables = {t.name: t for t in src.db.models.foreign.metadata.tables.values()}
         staging_tables = {t.name: t for t in src.db.models.staging.metadata.tables.values()}
 
+        columns_to_exclude = {
+            "tuser_account_mapper": ["email_address"],
+        }
```
chouinar, line 276: We want to exclude the email address from the `tuser_account` table, I mixed up the table names previously in that thread.

I also think we need to figure out where we want this configuration - if it's not a parameter we can override on the CLI (is there ever a case we'd want to do that?) then I'm fine just defining it statically in the LoadOracleDataTask's init function.

chouinar (approved):
LGTM!

FYI - I created the follow-up ticket to actually run the load / test this works for dev/staging: https://github.com/HHS/simpler-grants-gov/issues/4403

Issue #4333, opened 2025-03-26
Modify extract metadata table to use UUID for its primary key instead of integer


joshtonava:
### Summary

See #4239 

This change should be relatively contained. Extract metadata could be dropped/remade.

1. First temporarily delete the [ExtractMetadata](https://github.com/HHS/simpler-grants-gov/blob/04b68bf2acb8d17af9f356ff3aecb9fa5eb18b35/api/src/db/models/extract_models.py#L10-L21) table and generate a new migration.
2. Re add the ExtractMetadata table with the `extract_metadata_id` as the primary key now as a UUID.
**TODO: ** Is there an additional staging table update that needs to happen to force data regeneration? 

Known things that will be affected:
- There are some tests that reference the `extract_metadata_id` field
- The [API Schema](https://github.com/HHS/simpler-grants-gov/blob/5d4754f92f1922762bf9dc2dde0e0c7e8d85786d/api/src/api/extracts_v1/extract_schema.py#L45-L51)  will break 

### Acceptance criteria

- [ ] DB schema changes made
- [ ] Migrations created as directed
- [ ] All fixes made and tests updated

PR #4392, opened 2025-03-31
[Issue #4333] Modify extract metadata table to use UUID for its primary key instead of integer

joshtonava:
## Summary
Fixes #{4333}

## Changes proposed
- Updated ExtractMetadata table to use UUID instead of BigInteger
- Updated ExtractMetadata API models to use UUID
- Updated tests depending on the ExtractMetadata id
- Generated openapi schema changes

## Context for reviewers
- Running the migrations will drop the agency table and recreate it with the proper ids.
- This is part of a larger effort to completely move the db from using BigInteger to UUID


chouinar:
Change looks good, but lets hold until after the next deploy to avoid this and the agency one going out together - just in case

Issue #4153, opened 2025-03-11
Populate agencies search index


mxk0:
### Summary

Search index support for #4152 

### Acceptance criteria

- [ ] Agencies are included in the search index

PR #4406, opened 2025-04-01
[Issue #4153] Load agency into search index

babebe:
## Summary
Fixes #{[4153](https://github.com/HHS/simpler-grants-gov/issues/4153)}

## Changes proposed
Load all agencies in the db to dedicated search index
Screenshot 2025-04-01 at 4 20 32 PM


api/src/search/backend/load_agencies_to_index.py
```diff
@@ -0,0 +1,78 @@
...
+    def fetch_agencies(self) -> Sequence[Agency]:
+        """
+        Fetch all agencies.
+        """
+        return self.db_session.execute(select(Agency).options(selectinload("*"))).scalars().all()
+
+    def load_agencies(self, agencies: Sequence[Agency]) -> None:
+        logger.info("Loading agencies...")
+
+        schema = AgencyResponseSchema()
```
chouinar, line 71: I'd recommend defining this statically once at the top of the file so we're not re-initializing it repeatedly.

We might also want to add is_test_agency to this schema since from talking earlier it sounds like we'd want to be able to filter by that - but that can be delayed until the API itself gets updated.

api/src/search/backend/load_agencies_to_index.py
```diff
@@ -0,0 +1,78 @@
...
+        agencies_json = [schema.dump(record) for record in agencies]
```
chouinar, line 72: Could you make it so it logs the agency ID + agency code of each agency as it processes them? That helps a lot with debugging.

Just something like:
```suggestion
        agencies_json = []
        for record in agencies:
            logger.info("Preparing agency for upload to search index", extra={"agency_id": agency.agency_id, "agency_code": "agency.agency_code"})
            agencies_json.append(schema.dump(record))
```

api/Makefile
```diff
@@ -284,8 +284,8 @@ setup-foreign-tables:
 seed-local-legacy-tables:
 	$(PY_RUN_CMD) python3 -m tests.lib.seed_local_legacy_tables
 
-populate-search-opportunities: ## Load opportunities from the DB into the search index, run "make db-seed-local" first to populate your database
-	$(FLASK_CMD) load-search-data load-opportunity-data $(args)
+populate-search-opportunities: ## Load opportunities and agencies from the DB into the search index, run "make db-seed-local" first to populate your database
+	$(FLASK_CMD) load-search-data load-opportunity-data load-agency-data $(args)
```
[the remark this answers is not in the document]
babebe: Done

api/src/search/backend/load_agencies_to_index.py
```diff
@@ -0,0 +1,81 @@
...
+    model_config = SettingsConfigDict(env_prefix="LOAD_AGENCY_SEARCH_")  # update tf /rm ?
+
+    shard_count: int = Field(default=1)
+    replica_count: int = Field(default=1)
+
+    alias_name: str = Field(default="agency-index-alias")  # LOAD_AGENCY_SEARCH_ALIAS_NAME
+    index_prefix: str = Field(default="agency-index")  # LOAD_AGENCY_INDEX_PREFIX
+
+
+class LoadAgenciesToIndex(Task):
```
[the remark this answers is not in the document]
babebe: Added

api/src/search/backend/load_agencies_to_index.py
babebe: Done

api/src/search/backend/load_agencies_to_index.py
babebe: Noted!

api/src/search/backend/load_agencies_to_index.py
```diff
@@ -0,0 +1,90 @@
...
+        return self.db_session.execute(select(Agency).options(selectinload("*"))).scalars().all()
+
+    def load_agencies(self, agencies: Sequence[Agency]) -> None:
+        logger.info("Loading agencies...")
+
+        agencies_json = []
+        for agency in agencies:
+            logger.info(
+                "Preparing agency for upload to search index",
+                extra={"agency_id": agency.agency_id, "agency_code": "agency.agency_code"},
```
chouinar, line 81:
```suggestion
                extra={"agency_id": agency.agency_id, "agency_code": agency.agency_code},
```

api/src/search/backend/load_agencies_to_index.py
```diff
@@ -0,0 +1,90 @@
...
+from src.task.task import Task, logger
+from src.util.datetime_util import get_now_us_eastern_datetime
+from src.util.env_config import PydanticBaseEnvConfig
+
+SCHEMA = AgencyResponseSchema()
+
+
+class LoadAgenciesToIndexConfig(PydanticBaseEnvConfig):
+    shard_count: int = Field(default=1)
+    replica_count: int = Field(default=1)
```
chouinar, line 20: For these, give them more specific names so we can set them in terraform. Something like `LOAD_AGENCY_SEARCH_SHARD_COUNT`

For context - these need to be 3 & 2 respectively in prod otherwise the index creation will fail.

api/src/search/backend/load_agencies_to_index.py
```diff
@@ -0,0 +1,90 @@
...
+from src.task.task import Task, logger
+from src.util.datetime_util import get_now_us_eastern_datetime
+from src.util.env_config import PydanticBaseEnvConfig
+
+SCHEMA = AgencyResponseSchema()
+
+
+class LoadAgenciesToIndexConfig(PydanticBaseEnvConfig):
+    LOAD_AGENCY_SEARCH_SHARD_COUNT: int = Field(default=1)
+    LOAD_AGENCY_SEARCH_REPLICA_COUNT: int = Field(default=1)
```
chouinar, line 20: For naming these, we don't need to make the variable all caps (we shouldn't), we can do either:

`load_agency_search_shard_count: int = Field(default=1)`
OR
`shard_count: int = Field(default=1, alias=LOAD_AGENCY_SEARCH_SHARD_COUNT)`

babebe: thought that's what you were suggesting, gave it the alias

api/src/search/backend/load_agencies_to_index.py
```diff
@@ -16,11 +16,11 @@
...
 class LoadAgenciesToIndexConfig(PydanticBaseEnvConfig):
-    LOAD_AGENCY_SEARCH_SHARD_COUNT: int = Field(default=1)
-    LOAD_AGENCY_SEARCH_REPLICA_COUNT: int = Field(default=1)
+    shard_count: int = Field(default=1, alias="LOAD_AGENCY_SEARCH_SHARD_COUNT")
+    replica_count: int = Field(default=1, alias="LOAD_AGENCY_SEARCH_REPLICA_COUNT")
 
-    alias_name: str = Field(default="agency-index-alias")
-    index_prefix: str = Field(default="agency-index")
+    alias_name: str = Field(default="agency-index-alias")  # LOAD_AGENCY_SEARCH_ALIAS_NAME
+    index_prefix: str = Field(default="agency-index")  # LOAD_AGENCY_INDEX_PREFIX
```
chouinar, line 23: This is a bit misleading, those env var names wouldn't override these values, you'd need to set the alias to those to do that. As implemented, to set these env vars you'd set `ALIAS_NAME` and `INDEX_PREFIX`

babebe: since both of these are not env specific i will remove the "alias" comment and can just directly set the values as it is in test

chouinar (approved):

Issue #4343, opened 2025-03-26
Make frontend use real GET ApplicationForm endpoint


mxk0:
### Summary

\

### Acceptance criteria

- [ ] Frontend uses real GET ApplicationForm endpoint

acouch:
Note, will be implemented with this PR #4415 4415

PR #4415, opened 2025-04-01
[Issue #4189] sfs-424 initial form implementation

### LINKED ISSUE #4189: Finish initial SF-424 form implementation
### Summary

A prototype was created during this spike: https://github.com/HHS/simpler-grants-gov/issues/3885

This task is to implement the demo with the SF-424 form data:

* finishing and document uischema
* add schema validation
* add testing
* add server action
* write to mock endpoint

### Acceptance criteria

- [x] Form contains all SF-424 individual fields
- [x] Form renders on the server and uses a server action
- [x] Form pushes updates on submit to a mock API endpoint
- [x] Form includes testing and schema validation
- [ ] Form includes documentation
- [x] Form is behind feature flag

acouch:
## Summary
Fixes #4189 #4343

## Changes proposed
* Adds initial form driven by json schema and a ui schema
* Does not include actually connecting to the APIs
* Is not fully completed

#### Follow-ups

* Add testing for all functions
* Add checkbox widget
* Add documentation on UI Schema
* Clean up init file
* Handle post-pop fields
* Probably others...

#### To test

1. Fire up the local 
2. go to /formprototype
3. Play around with form, notice that the validation works
4. Submit a full form and see the success page


infra/frontend/app-config/env-config/environment_variables.tf
```diff
@@ -83,6 +83,10 @@ locals {
       manage_method     = "manual"
       secret_store_name = "/${var.app_name}/${var.environment}/feature-saved-searches-on"
     },
+    FEATURE_APPLY_FORM_PROTOTYPE_OFF = {
+      manage_method     = "manual"
+      secret_store_name = "/${var.app_name}/${var.environment}/feature-saved-searches-on"
```
doug-s-nava, line 88: not sure what's going on here - do we actually want to have the apply forms off whenever saved search is on, or is this a typo and should point to its own variable in SSM?

Also, could we make this a positive polarity to match the rest of the recent feature flag env vars?

frontend/package.json
```diff
@@ -29,24 +29,30 @@
     "ts:check": "tsc --noEmit"
   },
   "dependencies": {
-    "axios": "^1.8.2",
     "@newrelic/next": "^0.10.0",
     "@next/third-parties": "^15.1.4",
     "@opentelemetry/api": "^1.8.0",
+    "@rjsf/core": "^5.24.8",
```
doug-s-nava, line 35: looks like some of this is unused. Can we remove  rjsf/core, rjsf/validator-ajv8 and react-jsonschema-form?

frontend/src/components/applyForm/validate.ts
```diff
@@ -0,0 +1,37 @@
+import { RJSFSchema } from "@rjsf/utils";
+import Ajv from "ajv";
+import addFormats from "ajv-formats";
```
doug-s-nava, line 3: if we're using this it seems like it should be explicitly included in the package.json

frontend/src/app/[locale]/formPrototype/page.tsx
```diff
@@ -0,0 +1,78 @@
...
+    throw error;
+  }
+
+  const { form_json_schema, form_ui_schema } = formData;
+  try {
+    validateUiSchema(form_ui_schema);
+    validateFormSchema(form_json_schema);
+  } catch (error) {
+    console.error("Error validating", error);
+    return <TopLevelError />;
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/formPrototype/page.tsx:47]: [non-blocking] eventually we should add an error page for the route

also, I think if you just rethrow the error here it will bubble up to the top level error page fwiw?

frontend/src/app/[locale]/formPrototype/success/page.tsx
```diff
@@ -0,0 +1,35 @@
...
+      <Link className="usa-button" type="link" href="/formPrototype">
+        Back
+      </Link>
+    </GridContainer>
+  );
+}
+
+export default withFeatureFlag<WithFeatureFlagProps, never>(
+  FormSuccess,
+  "applyFormPrototypeOn",
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/formPrototype/success/page.tsx:33]: I think see the confusion here - `withFeatureFlag` expects a negative polarity, but otherwise our feature flags use a positive polarity. It's dumb. I think in this case we should have the app side feature flag use a negative polarity, and the SSM param could go either way

frontend/src/app/[locale]/formPrototype/page.tsx
```diff
@@ -0,0 +1,78 @@
...
+import ApplyForm from "src/components/applyForm/ApplyForm";
+import {
+  validateFormSchema,
+  validateUiSchema,
+} from "src/components/applyForm/validate";
+import BetaAlert from "src/components/BetaAlert";
+
+export function generateMetadata() {
+  const meta: Metadata = {
+    title: `Form test for form React JSON Schema`,
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/formPrototype/page.tsx:22]: [non-blocking] do we want to go ahead and update this now so that nobody that might be looking at it gets the wrong idea about what tech we're using?

frontend/src/app/api/applications/[applicationId]/application_form/route.ts
```diff
@@ -0,0 +1,40 @@
+// NOTE: this route is only for testing, will be removed when API is ready
+// TODO: remove file
+import { readError } from "src/errors";
+
+export async function PUT(
+  request: Request,
+  { params }: { params: Promise<{ applicationId: string; appFormId: string }> },
+) {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
Comment (doug-s-nava) [inline frontend/src/app/api/applications/[applicationId]/application_form/route.ts:9]: [non-blocking nit] this can get deleted

frontend/src/app/[locale]/formPrototype/page.tsx
```diff
@@ -0,0 +1,78 @@
...
+    description: "this is the form",
+  };
+  return meta;
+}
+
+async function FormPage() {
+  let formData = {} as formDetail;
+  const id = "anyform";
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/formPrototype/page.tsx:30]: where will this id eventually be coming from? a query param? if so we should probably go ahead and put this page in a dynamic location to allow for that

frontend/src/components/applyForm/types.ts
```diff
@@ -0,0 +1,96 @@
...
+export type TextTypes =
+  | "text"
+  | "email"
+  | "number"
+  | "password"
+  | "search"
+  | "tel"
+  | "url";
+
+export interface UswdsWidgetProps<
```
doug-s-nava, line 61: could be worth a comment in here linking to what you're basing this widget typing on

frontend/src/components/applyForm/utils.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+import { JSX } from "react";
+
+import { UiSchema } from "./types";
+import { FieldsetWidget } from "./widgets/FieldsetWidget";
+import SelectWidget from "./widgets/SelectWidget";
+import TextareaWidget from "./widgets/TextAreaWidget";
+import TextWidget from "./widgets/TextWidget";
+
+export function buildFormTreeClosure(
```
doug-s-nava, line 14: [incredibly inconsequential nit] I feel like `buildForTreeRecursive` or `buildFormTreeNested` is a better term - the `buildFormTree` defined in here is closing over the accumulator array, but I don't think we need to call out the closure, especially since later on I'd expect this to be implemented as a regular reducer

frontend/src/components/applyForm/utils.tsx
```diff
@@ -0,0 +1,197 @@
...
+        // acc = parentHasChild(uiSchema) ? wrapSection(parent.label, parent.name, row) : [acc, wrapSection(parent.label, parent.name, row)];
+      }
+    }
+  };
+  buildFormTree(uiSchema, null);
+  return acc;
+}
+
+const createField = (
+  id: string,
```
doug-s-nava, line 70: wow that's a lot of arguments. I'd say this is non-blocking but it's more annoying to make this sort of change later on - what do you say to packing this into an object argument so that we don't have to worry about argument order?

doug-s-nava (commented):
haven't kicked the tires yet, but a couple minor things to think about.

doug-s-nava (commented):
it looks like http://localhost:3000/formPrototype works currently by default - seems like the flags should be set so that by default it redirects to maintenance

otherwise some notes on the UI. I don't think any of this is blocking, but we should capture any of this that we want to fix for the upcoming tickets:

* the table of contents disappears at mobile width
* first portion of the form doesn't have a table of contents entry, imagine this is just an issue with the form data
* for something like opportunity number, do we want an incrementer? I doubt that would get much use and might be confusing
* is there support for the sort of validation that would allow us to make sure the project start date is before the project end date?
* some funkiness in the following scenario:
  * leave one required field blank
  * enter a totally bogus email address in the email field, something like "notan email"
  * submit
  * "please enter an email address" notification appears that seems like it is from a library of some sort
  * enter a value in the previously blank required field
  * submit
  * notice that the previously blank required field still has the invalid message and styling
  * update the email address to something more correct like "notan@mail"
  * notice a different error is received

acouch:
Thanks for the comments:

> it looks like http://localhost:3000/formPrototype works currently by default - seems like the flags should be set so that by default it redirects to maintenance

I updated this, it took me off guard that the feature flag for pages is reversed. I might update that in another PR.

> the table of contents disappears at mobile width

Good catch. That is the same behavior as the truss component: https://trussworks.github.io/react-uswds/iframe.html?globals=&args=&id=components-in-page-navigation--default&viewMode=story . There isn't a design yet, so the final version might not include the navigation, so not going to fix that for the demo.

> first portion of the form doesn't have a table of contents entry, imagine this is just an issue with the form data

This is part of the uiSchema. The first entries don't have a wrapper component. I'm going to update that before the demo.

> is there support for the sort of validation that would allow us to make sure the project start date is before the project end date?

This would be nice to show, so will include that in #4426

> some funkiness in the following scenario

This is partly an issue from the input for the email being `type="email"` which has its own validation. At some point we might want to add "onChange" validation that would help clarify some of these types of things. I'm thinking that is out of scope for the demo, and we'll get feedback and can create a strategy post demo.

frontend/src/components/applyForm/utils.tsx
acouch: Yeah, I meant to remove this function. I updated it. Thanks for noting it.

frontend/src/components/applyForm/types.ts
acouch: Added, thanks.

frontend/src/app/[locale]/formPrototype/success/page.tsx
Comment (acouch) [reply frontend/src/app/[locale]/formPrototype/success/page.tsx]: I changed that to `Off`, but missed this page. Updated.

frontend/src/app/[locale]/formPrototype/page.tsx
Comment (acouch) [reply frontend/src/app/[locale]/formPrototype/page.tsx:22]: The next tickets involve actual routes, so will update this there.

frontend/src/components/applyForm/validate.ts
acouch: Added, thanks.

frontend/src/app/[locale]/formPrototype/[id]/page.tsx
```diff
@@ -29,20 +26,31 @@ export function generateMetadata() {
...
 
-async function FormPage() {
-  let formData = {} as ApplyFormDetail;
-  const id = "anyform";
+interface formPageProps {
+  params: Promise<{ id: string; locale: string }>;
+}
+
+async function FormPage({ params }: formPageProps) {
+  const { id } = await params;
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/formPrototype/[id]/page.tsx:34]: ❤️

frontend/src/components/applyForm/data/SF424_Individual_2_0-V2.0_F668/Form DAT-Table 1.csv
```diff
@@ -0,0 +1,76 @@
+FORM TITLE:,,Application for Federal Domestic Assistance- Individual,,,,,,,,,,,,,,,,
```
doug-s-nava, line 1: can we delete this hardcoded data at his point since we're grabbing this from the API?

doug-s-nava (approved):
When I visit http://localhost:3000/formPrototype/fd7f5921-9585-48a5-ab0f-e726f4d1ef94 I get a form that looks like this.
Screenshot 2025-04-04 at 10 11 26 AM


The inputs seem to work fine, and I can submit the form successfully. There doesn't seem to be any validation, and the seed doesn't seem to create any other forms, so it seems like this meets the AC. Glad to see it connecting to the API!

✅ ✅ ✅ ✅

Issue #3949, opened 2025-02-19
DNS records for all Simpler services


mxk0:
### Description

**BLOCKED BY #3972**

URLs for prod and lower environments for all services:

- API
- Frontend
- Metabase
- Files (S3)
- Email notifications

### Probability of success and key risks

High

### Date of delivery (estimated)

April 24, 2025

### Level of effort

None

PR #4423, opened 2025-04-02
[Issue #3949] Email Notifications

coilysiren:
## Changes proposed

- enables email notifications
- adds a domain that we can mount email notifications onto
- changes a condition to check `enable_https` instead of `domain_name` because we are going to be in a temporary state where we have a domain name but not a cert _(I really should have guessed that this would happen)_
- creates a https://simpler.grants.gov hosted zone, which isn't connected to the real https://simpler.grants.gov, but does allow us to view what routes it **would** create. This allows us to copy paste those DNS records into the DNS request form for HHS.

## Context for reviewers

Whole lot of random unforesoon consequences that can happen here... mostly related to DNS and certs. 

## Testing

The goal was to acquire these DNS routes so I can request them on the real DNS zone:

image


mdragon (approved):
🚀

PR #4431, opened 2025-04-02
add path for next to add request logging with pino-http

doug-s-nava:
## Summary
Fixes #{ISSUE}


doug-s-nava:
we're going in a different direction with logging. This branch is interesting, but closing PR for now to reduce noise

PR #4432, opened 2025-04-02
[Unticketed] Update seed script to create a competition for frontend devs

chouinar:
## Changes proposed
Create a competition in our seed script at a static ID for frontend dev of the application workflow

## Context for reviewers
We can create more competitions with more variance in the future, but this'll make a competition with the same form every time.


acouch (approved):
👍🏿

mdragon (approved):
Approved

PR #4451, opened 2025-04-03
[Unticketed] Disable developer mode for New Relic

chouinar:
## Changes proposed
Disable developer mode so New Relic stops trying to trace in tests which has a small chance to cause tests to fail due to our test setup


coilysiren (approved):
👍🏼 (saw the thread about this)

PR #4455, opened 2025-04-03
[Issue #4242] no login redirect to unauthorized

### LINKED ISSUE #4242: Allow redirecting user to intended page rather than /unauthorized after login
### Summary

I would like to see us support not redirecting and having a message about logging in required so that with our "return them to where they logged in" logic, they can sign in and get back to the page that required the log in.

But I think that could be a follow up ticket? Maybe we want to have the /unauthorized page not save itself when clicking login? Or have that page redirect to the home page if you hit it while signed in?

### Acceptance criteria

When a non authenticated user hits a page gated by auth, such as /saved-grants
- [ ] after login they are redirected back to the page they want to visit, for example /saved-grants, rather than /unauthorized

doug-s-nava:
## Summary
Fixes #4242

## Changes proposed

* replaces behavior that redirected unauthenticated users to an "unauthenticated" route with functionality to display an unauthenticated message on the existing route
* removes `/unauthorized` route
* updates some fetch functionality for consistency

Note that since the saved grants and saved opportunities pages are server rendered almost entirely, they do not respond automatically to a logout, since this happens almost entirely in the client. Because of this, to get the page data to refresh to the unauthenticated state, we are performing a page refresh after all logouts. This is somewhat heavy handed, but by far the easiest solution at this point, without making the entire page render client side.

Also note that on future work for automated logouts for expired tokens or inactivity we likely will also need to do a refresh of the page for the same reasons as here.

### Test steps

1. start this branch with `npm run dev`
2. visit http://localhost:3000/?_ff=authOn:true;savedSearchesOn:true;savedOpportunitiesOn:true
3. visit http://localhost:3000/saved-search-queries
4. _VERIFY_: you see an unauthorized message but are still on /saved-search-queries
5. log in
6. _VERIFY_: you wind up back at /saved-search-queries viewing a logged in version of the page
7. log out
4. _VERIFY_: you see an unauthorized message but are still on /saved-search-queries
5. visit http://localhost:3000/saved-grants
6. repeat steps 4- 8


frontend/src/app/api/auth/callback/route.ts
```diff
@@ -6,7 +6,7 @@ import { NextRequest } from "next/server";
 export async function GET(request: NextRequest) {
   const token = request.nextUrl.searchParams.get("token");
   if (!token) {
-    return redirect("/unauthorized");
+    return redirect("/unauthenticated");
```
doug-s-nava, line 9: this is a case we haven't really discussed so largely leaving it as is for now

frontend/.pa11yci-desktop.json
```diff
@@ -2,7 +2,7 @@
   "defaults": {
     "timeout": 240000,
     "runners": ["axe"],
-    "ignore": ["color-contrast"],
+    "ignore": ["color-contrast", "document-title"],
```
doug-s-nava, line 5: didn't want to do this, but couldn't find an easy way around it. When the unauthenticated version of the pages load, the metadata does not, even though as far as I could tell the Next internals pick it up correctly.

Seems likely to be related to the next bug mentioned here https://github.com/vercel/next.js/issues/77512

doug-s-nava: Yeah, good idea https://github.com/HHS/simpler-grants-gov/issues/4495

acouch (approved):

PR #4487, opened 2025-04-07
[no ticket] Scale down RDS

coilysiren:
## Context

We have been overscaled for quite some time, and passively aware of that fact. I'm doing something about that now, to offset the cost increase of the fluent bit containers.

### DIFF
--- a/infra/api/app-config/prod.tf
+++ b/infra/api/app-config/prod.tf
@@ -20,16 +20,23 @@ module "prod_config" {
   # The math is: 5 * max(average CPU or average Memory) * 1.3. The 1.3 is for a buffer.
   instance_desired_instance_count = 2
   instance_scaling_min_capacity   = 2
-  # instance_scaling_max_capacity is 5x the instance_scaling_min_capacity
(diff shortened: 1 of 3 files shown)


infra/api/app-config/prod.tf
```diff
@@ -20,16 +20,23 @@ module "prod_config" {
   # The math is: 5 * max(average CPU or average Memory) * 1.3. The 1.3 is for a buffer.
   instance_desired_instance_count = 2
   instance_scaling_min_capacity   = 2
-  # instance_scaling_max_capacity is 5x the instance_scaling_min_capacity
-  instance_scaling_max_capacity = 10
+  # instance_scaling_max_capacity is 4x the instance_scaling_min_capacity
+  # the "4x" number is functionally arbibrary.
+  instance_scaling_max_capacity = 12
```
chouinar, line 25: Isn't this 6x?

coilysiren: Ah thanks

mdragon (approved):
🚀

PR #4489, opened 2025-04-07
[no ticket] Remove evidently vpc endpoints

coilysiren:
## Context

I was looking for things that cost too much money, and we don't use evidently at all. In fact, its deprecated.


mdragon (approved):
👍🏻

PR #4492, opened 2025-04-07
[no ticket] enable readonly root filesystem for analytics

coilysiren:
## Contex

1. Having a writeable root filesystem is a security concern

2. As of https://github.com/HHS/simpler-grants-gov/pull/3799 we don't need one anymore


mdragon (approved):
🚀 🔒

Issue #4152, opened 2025-03-11
Add agency search endpoint


mxk0:
### Summary

API support for #3916 

### Acceptance criteria

- [ ] Endpoint is live

capriiis:
Dependent upon 4153

PR #4493, opened 2025-04-07
[Issue #4152] Agency search endpoint

### LINKED ISSUE #3916: Implementation: Add "Search" Box to Agency Filter
### Summary

In usability testing, users had trouble finding the agency they need in the long list of agencies.

To enable users to more quickly find the agency they want to filter for, we're adding a search box to the agency filter.

Open questions (maybe to wait until agencies are in the search index)

- What happens if the user types e.g. "Homeland Security"? Do we return options currently nested under DHS e.g. Region 6?
  - Select all of the sub-agencies available at the time
- What happens if the user searches e.g. "Region" and "Region" appears under multiple top-level agencies?
  - Show the associated parent agency as an item in the list (similar to USASpending.gov)
- Should we prepend all sub-options with the top-level agency name? (E.g. "Department of Homeland Security - Region 6")
  - Yes, see above

### Acceptance criteria

- [ ] Search field shows sub-agency and agency in the result list (design TBD)
- [ ] Search field is live for Agency filter
- [ ] Follow-up tickets made for open questions

### Designs

https://www.figma.com/design/FGxWtAgToKhehLJCiuy1zL/Simpler.Grants.gov-Project?node-id=5714-28206&t=7iLoeV50KqotGuA0-1

babebe:
## Summary
Fixes #3916

## Changes proposed
Added Agency Search endpoint. 
Updated load agency to search index task to include `is_active_status` field. This is used to filter agencies based on their active status.
Added test


api/src/api/agencies_v1/agency_schema.py
```diff
@@ -20,6 +20,10 @@ class AgencyListRequestSchema(Schema):
     )
 
 
+class AgencySearchRequestSchema(AgencyListRequestSchema):
+    query = fields.String()
```
chouinar, line 24: I'd recommend not deriving this from the list endpoints schema - what we want for a search request is different than what it did.

I'd recommend we follow the same pattern as the opportunity search endpoint and have filters structured the same way (using the StrSearchSchemaBuilder and such).

api/src/api/agencies_v1/agency_schema.py
```diff
@@ -39,3 +43,30 @@ class AgencyListResponseSchema(AbstractResponseSchema, PaginationMixinSchema):
         fields.Nested(AgencyResponseSchema),
         metadata={"description": "A list of agency records"},
     )
+
+
+class AgencyV1Schema(Schema):
+    agency_id = fields.UUID(
+        metadata={"description": "The internal ID of the agency", "example": "123res45"},
```
chouinar, line 50: The example you had wasn't a UUID - UUID has a default example already so we don't need to specify it generally.
```suggestion
        metadata={"description": "The internal ID of the agency"},
```

api/src/services/agencies_v1/get_agencies.py
```diff
@@ -76,3 +85,57 @@ def get_agencies(
...
+
+    # Pagination
+    builder.pagination(
+        page_size=params.pagination.page_size, page_number=params.pagination.page_offset
+    )
+
+    # Query
+    if params.query:
+        filter_rule = AGENCY_ONLY
+        builder.simple_query(params.query, filter_rule, SearchQueryOperator.OR)
```
chouinar, line 101: Should we add the operator to the request like we did for the API?

api/src/services/opportunities_v1/experimental_constant.py
```diff
@@ -46,9 +46,18 @@
     "attachments.attachment.content",
 ]
 
+AGENCY_ONLY = [
+    "agency_name",
+    "agency_code",
+    "agency_code.keyword",
+    "top_level_agency_name",
+    "top_level_agency_code" "top_level_agency_code.keyword",
+]
```
chouinar, line 55: Could you make a file in the agency service folder instead of putting in in the opportunity configuration?

api/tests/src/api/agencies_v1/test_agency_routes_search.py
```diff
@@ -0,0 +1,84 @@
...
+    )
+    def test_search_agencies(self, client, api_auth_token, search_request, expected_result):
+        resp = client.post(
+            "/v1/agencies/search", json=search_request, headers={"X-Auth": api_auth_token}
+        )
+        data = resp.json["data"]
+
+        import pdb
+
+        pdb.set_trace()
```
chouinar, line 81:
```suggestion

```

api/src/services/agencies_v1/get_agencies.py
babebe: Ys, it makes sense to add it.

api/src/services/opportunities_v1/experimental_constant.py
babebe: Moved

api/src/api/agencies_v1/agency_schema.py
```diff
@@ -20,6 +22,39 @@ class AgencyListRequestSchema(Schema):
     )
 
 
+class AgencySearchFilterV1Schema(Schema):
+    active = fields.Nested(
```
chouinar, line 26: Nitpicky - I know we called it `active` on the existing agency endpoint, but we should really make booleans be named as sorta-questions (is/has/are/was/can, whatever makes sense in the context), could we call this `is_active` here? If we wanted to be very accurate, we could name it something like `has_active_opportunity` so it's clear what "active" means.

api/src/api/agencies_v1/agency_schema.py
```diff
@@ -20,6 +22,39 @@ class AgencyListRequestSchema(Schema):
...
+    )
+
+
+class AgencySearchRequestSchema(Schema):
+    query = fields.String(
+        metadata={
+            "description": "Query string which searches against several text fields",
+            "example": "research",
+        },
+        validate=[validators.Length(min=1, max=100)],
```
chouinar, line 37: Check with the frontend on the validation here, I know us having a limit of 100 caused a few issues on regular search.

api/src/api/agencies_v1/agency_schema.py
```diff
@@ -39,3 +74,39 @@ class AgencyListResponseSchema(AbstractResponseSchema, PaginationMixinSchema):
...
         metadata={"description": "A list of agency records"},
     )
+
+
+class AgencyV1Schema(Schema):
+    agency_id = fields.UUID(
+        metadata={"description": "The internal ID of the agency"},
+    )
+    agency_name = fields.String(
+        allow_none=False,
```
chouinar, line 84: I'd be cautious about adding these allow_none values - note that they don't affect anything on the response (APIFlask does not do response validation, it'll try its best to make your response into JSON, but doesn't validate it) so this only affects incoming JSON which would only happen with this schema when we write to the search index.

api/src/search/backend/load_agencies_to_index.py
```diff
@@ -73,14 +74,27 @@ def fetch_agencies(self) -> Sequence[Agency]:
...
     def load_agencies(self, agencies: Sequence[Agency]) -> None:
         logger.info("Loading agencies...")
+        active_agency_subquery = (
+            _construct_active_inner_query(Agency.agency_id)
+            .union(_construct_active_inner_query(Agency.top_level_agency_id))
+            .subquery()
+        )
+
+        agency_id_stmt = select(active_agency_subquery).distinct()
+        active_agencies = set(self.db_session.execute(agency_id_stmt).unique().scalars().all())
```
chouinar, line 84: Does just this work?
```suggestion
        active_agencies = set(self.db_session.execute(agency_id_stmt).scalars())
```

api/src/search/backend/load_agencies_to_index.py
```diff
@@ -73,14 +74,27 @@ def fetch_agencies(self) -> Sequence[Agency]:
...
         agencies_json = []
         for agency in agencies:
             logger.info(
                 "Preparing agency for upload to search index",
                 extra={"agency_id": agency.agency_id, "agency_code": agency.agency_code},
             )
-            agencies_json.append(SCHEMA.dump(agency))
+            agency_json = SCHEMA.dump(agency)
+            if agency.agency_id in active_agencies:
+                agency_json["is_active_agency"] = True
```
chouinar, line 94: Can also set it when it's false just for completeness / one less line
```suggestion
            agency_json["is_active_agency"] = agency.agency_id in active_agencies
```

api/src/services/service_utils.py
```diff
@@ -25,3 +33,75 @@ def apply_sorting(stmt: Select, model: Type, sort_order: list) -> Select:
...
+
+def _add_search_filters(
+    builder: search.SearchQueryBuilder, filters: BaseModel | None = None
+) -> None:
+    if filters is None:
+        return
+
+    for field in filters.model_fields_set:
+        field_filters = getattr(filters, field)
+        field_name = _adjust_field_name(field)
```
chouinar, line 78: While you can re-use this function, I'd make it so you pass in the field name mapping to the function.

The mapping of field names for different search indexes should be stored separately.

api/src/search/backend/load_agencies_to_index.py
babebe: Works!

api/src/api/agencies_v1/agency_schema.py
babebe: @Doug gave the clear

api/src/api/agencies_v1/agency_schema.py
babebe: Updated to has_active_opportunity.

api/src/services/service_utils.py
babebe: Updated

api/src/api/agencies_v1/agency_schema.py
babebe: Removed. it is not necessary to to do the check to index.

api/src/api/agencies_v1/agency_routes.py
```diff
@@ -56,3 +60,55 @@ def agencies_get(db_session: db.Session, raw_list_params: dict) -> response.ApiR
...
+@agency_blueprint.post("/agencies/search")
+@agency_blueprint.input(
+    agency_schema.AgencySearchRequestSchema,
+    arg_name="raw_search_params",
+    examples=examples,
+)
+@agency_blueprint.output(agency_schema.AgencySearchResponseV1Schema)
+@agency_blueprint.doc(
+    responses={200: {"content": {"application/octet-stream": {}}}},  # type: ignore
+)
```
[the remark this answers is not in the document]
babebe: Removed

chouinar (approved):

PR #4498, opened 2025-04-07
[Issue #4452] Further attempts to prevent saved items caching 

### LINKED ISSUE #4452: Newly saved or unsaved grants/saved search queries aren't reflected on list pages until after a refresh
### Summary

_Note: these steps below are for saved opportunities, but the same applies to saved search queries._

Steps to repro:

- Visit [staging](http://frontend-staging-1506108424.us-east-1.elb.amazonaws.com/) with feature flags for auth and saved opportunities enabled
- Log in using Login.gov
- Visit /search
- Open an opportunity
- Save that opportunity
- Navigate to /saved-grants, either via the shortcut link in the toast next to the Saved button, or via the Workspace -> Saved opportunities link in the header nav

Expected:

- The grant I just saved appears in the list

Actual:

- The list is empty
- If I refresh, the grant I just saved shows up

The reverse is also an issue:

Steps to repro:

- After doing all the above, click the saved grant on /saved-grants
- Click "Saved" to unsave the grant
- Click Workspace -> Saved opportunities in the header nav

Expected: 

- Grant I just unsaved does not show up in the list

Actual:

- Grant still appears
- If I refresh, grant no longer appears

Note: these issues only happen ~90% of the time. I haven't been able to reliably reproduce a situation where the behavior works as expected.

### Reproducibility

None

### How to reproduce the bug

_No response_

### Browser

None

### Code of Conduct

mdragon:
## Summary
Fixes #4452

## Changes proposed
Use NextJS config to prevent pages we want to always re-render from caching. You don't see the effect of this in npm run dev have to build & start.

Context for reviewers
This is really tough to test, but the Cache-Control headers are better behaved under build & start following this change so 🤞🏻

### DIFF
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -76,6 +76,27 @@ const nextConfig = {
           },
         ],
       },
+      // don't cache user specific pages: saved-grants, saved-search-queries
+      {
+        source: "/saved:path*",
+        headers: [
+          {
+            key: "Cache-Control",
+            value: "no-store, must-revalidate",
+          },
+        ],
+      },
+      // don't cache if users has a session cookie
+      {
+        source: "/:path*",
+        has: [{ type: "cookie", key: "session" }],
+        headers: [
+          {
+            key: "Cache-Control",
+            value: "no-store, must-revalidate",
+          },
+        ],
+      },
     ];
   },
   basePath,


frontend/next.config.js
```diff
@@ -76,6 +76,27 @@ const nextConfig = {
           },
         ],
       },
+      // don't cache user specific pages: saved-grants, saved-search-queries
+      {
+        source: "/saved:path*",
```
acouch, line 81: Since the `/saved-*` paths are all behind pages that require a session cookie, I think this entry might not be needed. Testing locally I was able to see the refreshes happen as they are supposed to with just the cookie entry.

acouch (approved):

frontend/next.config.js
doug-s-nava: yeah, we might want to try running it rebased off of this see how things work in that world https://github.com/HHS/simpler-grants-gov/pull/4455

doug-s-nava: without the redirect, removing the /saved* declaration things work for me locally on a local prod build, so I think we can remove that declaration?

acouch: When clicking with the link component from an opportunity the `saved-grants` page is loaded through client side navigation, which in the browser is a fetch, ie `http://localhost:3000/saved-grants?_rsc=hp2fn`. That fetch was getting cached before which the cookie entry fixes.

I was able to recreate an experience where you login in tab 1, logout tab 2, click saved grants in tab 1, login in tab 2, and click saved grants it does initially load the cached page, so let's leave it in as is.

PR #4510, opened 2025-04-08
[Issue #4393] Update GET Opportunity endpoint

### LINKED ISSUE #4393: Modify Opportunity tables to use UUID for their primary key instead of integer
### Summary

See #4239, #4333 

We want to utilize UUIDs for our primary keys as they are easier to work with (can generate them or let the DB without ever hitting a conflict). However the primary key on a lot of the data we're working with was integers when we imported it.

So far the Agency #4239, and Extract Metadata #4333 tables are accounted for.

The [Opportunity tables](https://github.com/HHS/simpler-grants-gov/blob/1f08b0735b3690e487bc71f3605e975b87074145/api/src/db/models/opportunity_models.py) need to be updated to use UUID.
- Opportunity
- OpportunitySummary 
- LinkOpportunitySummaryFundingInstrument
- LinkOpportunitySummaryFundingCategory
- LinkOpportunitySummaryApplicantType
- CurrentOpportunitySummary
- OpportunityAttachment
- OpportunityChangeAudit
- OpportunityVersion

We still need to be able to query specific tables by their integer ID to the old system, so for these tables we should still include this value as an indexed field without a foreign but renamed to `legacy_<keyname>_id`
- Opportunity
- OpportunitySummary
- OpportunityAttachment

This is a large, complex change as everything that uses Opportunities data will be affected. There will be foreign keys in Schemas referencing Opportunities that will need to be updated as well.

We should have 2 migrations:

1. First create a migration that drops all of the tables using specific BigInteger Ids. 
2. Then create a migration to add back all of the tables with the updated ids.

As part of the Migration that restores all tables, 
Prime the [staging tables](https://github.com/HHS/simpler-grants-gov/blob/250c6f6a2e726ad174e9506ecef7759a236135b1/api/src/data_migration/load/load_oracle_data_task.py#L21-L41) for re processing. Do this by adding the following SQL statements for each table:
 **TODO: Double check this **
`update staging.<table_name> set transformed_at = NULL where is_deleted = FALSE`

The opportunity change audit table is populated whenever certain opportunity tables get updated. [We’ll both need to readd the triggers](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/db/migrations/versions/2024_10_28_add_opportunity_table_triggers.py), and update the [function they call.](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/db/migrations/constants.py) And put these all in the second migration.

Next it will require going through each place referencing the old id's and making sure things are updated correctly.

### Acceptance criteria

- [x] Update all opportunity models to use UUID
- [x] Updated all related models with Foreign keys to opportunity models 
- [x] Restore the opportunity change audit hooks
- [x] Update all application code to work with new opportunity ids
    - [x] get_opportunity should handle UUID or BigInteger
- [x] Update all batch jobs to work with new opportunity ids
    - [x] Ensure the search index job works as expected
    - [x] The jobs that generate CSV and JSON files
- [x] Update all affected tests
    - [x]  Update factories, `setup_next_sequence` should be no longer needed
- [ ] In the analytic system we need to make sure we are using the correct id.
    - [ ] May need to update the migrations to support this
- [ ] Turn off search index load job when we merge this
- [x] Download the saved opportunity and saved searches data to be added back for users after the change
- [x] Coordinate feature flag with frontend to put the site into a maintenance mode
- [x] manually run the job for testing
- [x] Add saved searches/saved opportunities back in before maintenance mode is disabled

joshtonava:
## Summary
Fixes #4393

## Changes proposed
- GET Opportunity endpoint was updated to support UUID and int for legacy id.
- Tests updated to handle both cases
- Open API schema Updated

## Context for reviewers

## Additional information
To run effected tests

```bash
make test args="-vv tests/src/api/opportunities_v1/test_opportunity_route_get.py"
```


api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -239,7 +240,7 @@ def _handle_incremental_delete(self, existing_opportunity_ids: set[int]) -> None
         # Fetch the opportunity IDs of opportunities we would expect to be in the index
         opportunity_ids_we_want_in_search: set[int] = set(
             self.db_session.execute(
-                select(Opportunity.opportunity_id)
+                select(Opportunity.legacy_opportunity_id)
```
chouinar, line 243: This might require changes elsewhere in the process, but we should be able to use UUIDs for the search index "ID".

api/src/services/opportunity_attachments/attachment_util.py
```diff
@@ -27,7 +27,7 @@ def get_s3_attachment_path(
     return file_util.join(
         base_path,
         "opportunities",
-        str(opportunity.opportunity_id),
+        str(opportunity.legacy_opportunity_id),
```
chouinar, line 30: For this function, it might be better to just use the UUIDs instead of the integers. Only reason to keep the integers is if we really want to come up with a way to not re-import the attachments (annoying to do, but pretty whatever).

api/tests/src/api/opportunities_v1/test_opportunity_route_get.py
```diff
@@ -88,6 +90,27 @@ def test_get_opportunity_with_attachment_200(
     validate_opportunity_with_attachments(opportunity, response_data)
 
 
+def test_get_opportunity_with_attachment_200_legacy(
+    client, api_auth_token, enable_factory_create, db_session
+):
+    # Create an opportunity with an attachment
+    opportunity = OpportunityFactory.create(has_attachments=True)
+    db_session.commit()
+
```
chouinar, line 99: I know it's done in some tests, but it's something we should cleanup - we don't need to commit after using factories, the factories auto-commit after creating things.

api/src/search/backend/load_opportunities_to_index.py
joshtonava: I think this is one I accidentally committed this with another change I wanted in this branch.

I will try to isolate the backend task updates to a branch

joshtonava: im moving to a new PR.

This will be addressed in joshtonava/4393-update-backend-tasks

PR #4513, opened 2025-04-08
[Issue #4174] Update GET /applications/:applicationId/application_form/:appFormId to include JsonSchema validations

### LINKED ISSUE #4174: Update GET /applications/:applicationId/application_form/:appFormId to include JsonSchema validations
### Summary

After #4172 and #4173 

We want to add JsonSchema validations to the GET endpoint - in #4172 we will have created a utility to easily fetch the validation issues, reuse that here and add the issues to the warnings in the response object.

### Acceptance criteria

- [ ] Validations run
- [ ] Tests updated to validate

mikehgrantsgov:
## Summary
Fixes #4174 

## Changes proposed
- Run `validate_json_schema_for_form` inside `application_form_get`
- Add tests

## Context for reviewers
Form data is now validated against its schema and any validation issues are returned as warnings in the response. The implementation includes proper error handling for invalid schemas and comprehensive test coverage for various validation scenarios.

## Additional information
Is exception handing in the route correct?

### DIFF
--- a/api/src/api/application_alpha/application_route.py
+++ b/api/src/api/application_alpha/application_route.py
@@ -14,6 +14,7 @@ from src.api.application_alpha.application_schemas import (
     ApplicationStartResponseSchema,
 )
 from src.auth.api_key_auth import api_key_auth
+from src.form_schema.jsonschema_validator import validate_json_schema_for_form
 from src.logging.flask_logger import add_extra_data_to_current_request_logs
 from src.services.applications.create_application import create_application
 from src.services.applications.get_application import get_application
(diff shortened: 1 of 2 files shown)


api/src/api/application_alpha/application_route.py
```diff
@@ -90,13 +91,33 @@ def application_form_get(
     )
     logger.info("GET /alpha/applications/:application_id/application_form/:app_form_id")
 
+    validation_warnings = []
```
chouinar, line 94: Can we put the validation warnings in the `get_application_form` function like we have for the PUT endpoint?

api/src/api/application_alpha/application_route.py
```diff
@@ -90,13 +91,33 @@ def application_form_get(
...
+                    )
+            except Exception:
+                # Log the exception but continue
+                logger.exception(
+                    "Error validating application form data against schema",
+                    extra={
+                        "application_form_id": app_form_id,
+                        "application_id": application_id,
+                    },
+                )
```
chouinar, line 115: I'm curious what issue you're accounting for here? Form is a non-nullable field on the application form.

Shouldn't need much more than:
```py
                    validation_warnings = validate_json_schema_for_form(
                        application_form.application_response, application_form.form
                    )
```

The only exception I'd reasonably expect from the validation is an invalid form schema which should be a 500 as it means our form itself is setup wrong.

mikehgrantsgov: Got it, I didn't know if we should be returning a 500 here. Will update.

api/src/services/applications/get_application_form.py
```diff
@@ -26,7 +30,12 @@ def get_application_form(
         )
     ).scalar_one_or_none()
 
+    if application_form:
+        warnings: list[ValidationErrorDetail] = validate_json_schema_for_form(
+            application_form.application_response, application_form.form
+        )
+
     if not application_form:
         raise_flask_error(404, f"Application form with ID {app_form_id} not found")
```
chouinar, line 39: To be very nitpicky, we could do:

```suggestion
    if not application_form:
        raise_flask_error(404, f"Application form with ID {app_form_id} not found")
        
    warnings: list[ValidationErrorDetail] = validate_json_schema_for_form(
             application_form.application_response, application_form.form
         )
```

PR #4523, opened 2025-04-08
[Issue #4468] Add tika local server for opportunity attachments

### LINKED ISSUE #4468: Try using Apache Tika to process opportunity attachments to search index
### Summary

https://github.com/chrismattmann/tika-python

This is the same software the search index uses to process attachments. Instead of using a pipeline + sending a b64 encoded attachment to the search index, we instead run it through on our side and add the content to what we upload to the search index. This removes the CPU bottleneck from the search index, moving it entirely to our ECS task.

At the moment, the only goal is to test this behavior and get it working locally, there may be infra setup necessary to get Tika working non-locally, and there will certainly be quite a bit of configuration even locally to make it work since Tika is a Java application that the Python library calls out to.

Roughly we should:
1. Get Tika working in the script locally
2. Look into what it takes to setup Tika conveniently locally (ie. it seems it can download the jar when it runs, but we should look to cache that in some way)
3. Figure out what we need non-locally in infra (if any)

### Acceptance criteria

- [ ] Apache Tika tested out

freddieyebra:
## Summary
Fixes #4468


api/docker-compose.yml
```diff
@@ -136,6 +137,14 @@ services:
     networks:
       - default
 
+  tika:
+    image: apache/tika:3.0.0.0
+    container_name: tika
+    ports:
+      - 9998:9998
+    networks:
+      - default
```
chouinar, line 146: The docs mention something about downloading a jar every time this runs, is that something we could leverage to install it in the dockerfile? https://github.com/chrismattmann/tika-python?tab=readme-ov-file#airgap-environment-setup

freddieyebra: I can definitely look into the local "client mode" option as well!

freddieyebra:
closing as wont do

PR #4524, opened 2025-04-09
[Unticketed] Address anchore scan

mdragon:
## Summary
Install patched library

## Changes proposed
Bump library that anchore scan is catching


doug-s-nava (approved):
✅ ✅ ✅ ✅

PR #4527, opened 2025-04-09
[Isssue #4342] Add save to form prototype

### LINKED ISSUE #4342: Make frontend use real PUT ApplicationForm endpoint - ie Form page save
### Summary

Save form data using PUT endpoint. Depends on #4340 

- [ ] Save allows form saving
- [ ] Submit saves form data and sends user to success message

### Acceptance criteria

- [ ] Frontend uses real PUT ApplicationForm endpoint

acouch:
## Summary
Fixes #4342

### Form flow
This change adds the save ability to the apply form.

The desired flow is that a user:

1. Starts at /formProtype
2. Clicks "Create new application"
3. Is directed to the "Application" page that has a list of forms (currently one)
4. *Change in this PR*

There is now a "save" button. A user can enter form info and save their place by clicking "Save":

image

Clicking "Submit" sends the user to the temporary success page. If you go back after clicking submit you should see values last entered.

## Context for reviewers
Remember this is a demo with no designs.

### DIFF
--- a/frontend/src/app/[locale]/formPrototype/[applicationId]/[formId]/page.tsx
+++ b/frontend/src/app/[locale]/formPrototype/[applicationId]/[formId]/page.tsx
@@ -84,8 +84,9 @@ async function FormPage({ params }: formPageProps) {
 
   const emptyApplicationForm = {
     application_response: {},
-    application_form_id: "",
   };
+  // TODO: this currently calls all forms for an application since we don't
+  // have the app_form_id in the path.
   const applicationForm =
     applicationData.application_forms.length > 0
       ? (applicationData.application_forms.find(
--- a/frontend/src/app/[locale]/formPrototype/[applicationId]/[formId]/page.tsx
+++ b/frontend/src/app/[locale]/formPrototype/[applicationId]/[formId]/page.tsx
@@ -3,7 +3,12 @@ import TopLevelError from "src/app/[locale]/error/page";
 import NotFound from "src/app/[locale]/not-found";
 import { ApiRequestError, parseErrorStatus } from "src/errors";
 import withFeatureFlag from "src/services/featureFlags/withFeatureFlag";
+import { getApplicationDetails } from "src/services/fetch/fetchers/applicationFetcher";
 import { getFormDetails } from "src/services/fetch/fetchers/formsFetcher";
+import {
+  ApplicationDetail,
+  ApplicationFormDetail,
+} from "src/types/applicationResponseTypes";
 import { FormDetail } from "src/types/formResponseTypes";
 
 import { redirect } from "next/navigation";
@@ -31,31 +36,71 @@ interface formPageProps {
 }
 
 async function FormPage({ params }: formPageProps) {
-  const { formId } = await params;
+  const { applicationId, formId } = await params;
   let formData = {} as FormDetail;
+  let applicationData = {} as ApplicationDetail;
 
   try {
     const response = await getFormDetails(formId);
     if (response.status_code !== 200) {
-      throw new Error(
-        `Error retrieving competition details: ${JSON.stringify(response.errors)}`,
+      console.error(
+        `Error retrieving form details for formID (${formId})`,
+        response,
       );
+      return <TopLevelError />;
     }
     formData = response.data;
-  } catch (error) {
-    if (parseErrorStatus(error as ApiRequestError) === 404) {
+  } catch (e) {
+    console.error(
+      `Error retrieving application details for formId ${formId}:`,
+      e,
+    );
+    if (parseErrorStatus(e as ApiRequestError) === 404) {
       return <NotFound />;
     }
-    throw error;
+    return <TopLevelError />;
   }
 
+  try {
+    const response = await getApplicationDetails(applicationId);
+    if (response.status_code !== 200) {
+      console.error(
+        `Error retrieving form details for applicationID (${applicationId}), formID (${formId})`,
+        response,
+      );
+      return <TopLevelError />;
+    }
+    applicationData = response.data;
+  } catch (e) {
+    if (parseErrorStatus(e as ApiRequestError) === 404) {
+      console.error(
+        `Error retrieving application details for applicationID (${applicationId}), formId ${formId}:`,
+        e,
+      );
+      return <NotFound />;
+    }
+    return <TopLevelError />;
+  }
+
+  const emptyApplicationForm = {
+    application_response: {},
+    application_form_id: "",
+  };
+  const applicationForm =
+    applicationData.application_forms.length > 0
+      ? (applicationData.application_forms.find(
+          (form) => form.form_id === formId,
+        ) as ApplicationFormDetail)
+      : emptyApplicationForm;
+
+  const { application_response } = applicationForm;
   const { form_id, form_name, form_json_schema, form_ui_schema } = formData;
 
   try {
     validateUiSchema(form_ui_schema);
     validateFormSchema(form_json_schema);
-  } catch (error) {
-    console.error("Error validating", error);
+  } catch (e) {
+    console.error("Error validating form", e);
     return <TopLevelError />;
   }
 
@@ -77,9 +122,11 @@ async function FormPage({ params }: formPageProps) {
         ).
       </p>
       <ApplyForm
+        rawFormData={application_response}
         formSchema={form_json_schema}
         uiSchema={form_ui_schema}
         formId={form_id}
+        applicationId={applicationId}
       />
     </GridContainer>
   );
--- a/frontend/src/app/[locale]/formPrototype/[applicationId]/page.tsx
+++ b/frontend/src/app/[locale]/formPrototype/[applicationId]/page.tsx
@@ -1,4 +1,5 @@
 import { Metadata } from "next";
+import TopLevelError from "src/app/[locale]/error/page";
 import NotFound from "src/app/[locale]/not-found";
 import { COMPETITION_ID } from "src/constants/competitions";
 import { ApiRequestError, parseErrorStatus } from "src/errors";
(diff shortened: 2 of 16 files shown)


frontend/src/app/[locale]/formPrototype/[applicationId]/[formId]/page.tsx
```diff
@@ -31,31 +36,71 @@ interface formPageProps {
...
-    if (parseErrorStatus(error as ApiRequestError) === 404) {
+  } catch (e) {
+    console.error(
+      `Error retrieving application details for formId ${formId}:`,
+      e,
+    );
+    if (parseErrorStatus(e as ApiRequestError) === 404) {
+      return <NotFound />;
+    }
+    return <TopLevelError />;
```
Comment (acouch) [inline frontend/src/app/[locale]/formPrototype/[applicationId]/[formId]/page.tsx:61]: There isn't a good error boundary at the moment, so not throwing errors. If it doesn't exist in the API, you get a 404, otherwise just a page error. We can revisit after the demo, once we are really building something.

frontend/src/app/[locale]/formPrototype/[applicationId]/[formId]/page.tsx
```diff
@@ -31,31 +36,71 @@ interface formPageProps {
...
+  }
+
+  try {
```
Comment (acouch) [inline frontend/src/app/[locale]/formPrototype/[applicationId]/[formId]/page.tsx:64]: I thought about separating these into separate fetch functions, but following the same data retrieval model of the opportunity page.

frontend/src/services/fetch/fetchers/applicationFetcher.ts
```diff
@@ -1,21 +1,53 @@
...
   const response = await fetchApplicationWithMethod("POST")({
     subPath: `start`,
     body: { competition_id: competitionID },
   });
 
-  return (await response.json()) as ApplicationStartResponse;
+  return (await response.json()) as ApplicationStartApiResponse;
+};
+
+export const getApplicationDetails = async (
```
doug-s-nava, line 21: the usage here sort of demonstrates the limitations of this layer of abstraction. At this point we'd probably be good just using the `fetchApplicationWithMethod` function directly within the components. Slighty harder to test but not much

frontend/src/components/applyForm/ApplyFormSuccessMessage.tsx
```diff
@@ -0,0 +1,11 @@
+import { Alert } from "@trussworks/react-uswds";
+
+export const ApplyFormSuccessMessage = ({ message }: { message: string }) => {
+  return (
```
doug-s-nava, line 4: we probably want to clear this message at some point - maybe when the user touches the form after save?

frontend/src/app/[locale]/formPrototype/[applicationId]/page.tsx
```diff
@@ -1,4 +1,5 @@
 import { Metadata } from "next";
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/formPrototype/[applicationId]/page.tsx:1]: I know we probably don't have design input on this yet which is why it hasn't been done, but a path setup like
`formPrototype/application/:applicationId/form/:formId would make a lot more sense, and we can probably lead the charge on that even without finished designs. As long as that is actually what this represents?

frontend/src/app/[locale]/formPrototype/[applicationId]/[formId]/page.tsx
```diff
@@ -77,9 +123,11 @@ async function FormPage({ params }: formPageProps) {
         ).
       </p>
       <ApplyForm
+        rawFormData={application_response}
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/formPrototype/[applicationId]/[formId]/page.tsx:126]: does the `application_response` / `rawFormData` represent a user's saved application state? If so how can we make that more clear using the variable naming here?

frontend/src/components/applyForm/actions.ts
```diff
@@ -1,71 +1,142 @@
...
   validationErrors: ErrorObject<string, Record<string, unknown>, unknown>[];
+};
+type applyFormResponse = {
+  applicationId: string;
   formData: object;
   formId: string;
+  successMessage: string;
 };
 
 export async function submitApplyForm(
```
doug-s-nava, line 25: [blocker] since this is handling both save and submit cases, can we rename this accordingly? maybe `handleFormAction`? hard to come up with a good name for this, which is probably a sign that it should be split up somehow to handle these two cases more discretely. 

at this point I think we should at least update the function name, and make a note to follow up to refactor this into two more clearly deliniated functions that will share most of their behavior through other utilities

doug-s-nava (commented):
a couple very minor blockers, and a number of things to add to the list to consider for later phases.

does what it's supposed to do though, so this is very close

frontend/src/app/[locale]/formPrototype/[applicationId]/[formId]/page.tsx
```diff
@@ -31,31 +36,72 @@ interface formPageProps {
...
-    throw error;
+    return <TopLevelError />;
   }
 
+  const emptyApplicationForm = {
+    application_response: {},
+  };
+  // TODO: this currently calls all forms for an application since we don't
+  // have the app_form_id in the path.
+  const applicationForm =
```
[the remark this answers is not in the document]
Comment (acouch) [reply frontend/src/app/[locale]/formPrototype/[applicationId]/[formId]/page.tsx:90]: This data fetching and route setup is all getting tossed. Let's focus on the most important parts of the demo. Right now we are writing data to a table that will be truncated after the demo.

frontend/src/components/applyForm/actions.ts
```diff
@@ -1,71 +1,142 @@
...
+          formData,
+          formId,
+          successMessage: "Form saved",
+          validationErrors,
+        };
+      }
+    }
+    return {
+      applicationId,
+      errorMessage: "Error saving the form",
```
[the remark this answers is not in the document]
acouch: We haven't discussed and don't have designs for the error states. I'd consider that out of scope for the demo and would want to hear from design about how we want to word things.

frontend/src/components/applyForm/actions.ts
```diff
@@ -1,71 +1,142 @@
...
-  } catch (error) {
-    if (parseErrorStatus(error as ApiRequestError) === 404) {
-      throw new Error();
+    if (response.status_code !== 200) {
+      console.error(
+        `Error retrieving form details for formID (${formId})`,
+        response,
+      );
+      return {
+        errorMessage: "Error submitting form",
```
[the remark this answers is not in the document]
acouch: This is an error requesting the form from the server. The form schema has to be retrieved to validate it. If we can't retrieve the form it can't be submitted.

frontend/src/components/applyForm/utils.tsx
```diff
@@ -203,3 +208,16 @@ const wrapSection = (
     </FieldsetWidget>
   );
 };
+
+export const filterFormData = <T extends object>(formData: FormData): T => {
```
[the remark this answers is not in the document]
acouch: I added an explanation. This function is getting changed and renamed in the next PR and has an explanatory comment.

frontend/src/services/fetch/fetchers/applicationFetcher.ts
acouch: I actually like how it provides the consistent names for the data and endpoint that is getting retrieved, even if it is extra verbose. You can see that from the endpoints as well, so don't have a strong opinion here.

frontend/src/components/applyForm/ApplyFormSuccessMessage.tsx
acouch: I'm not sure I understand your comment. Currently this is shown when a user clicks "save".

frontend/src/components/applyForm/actions.ts
acouch: The save and submit currently do the same thing, submit the data to the same place, the only difference is you get redirected if you click "submit." I updated this, but TBH doesn't seem like a blocking concern for the demo.

frontend/src/app/[locale]/formPrototype/[applicationId]/page.tsx
Comment (acouch) [reply frontend/src/app/[locale]/formPrototype/[applicationId]/page.tsx:1]: I'll add changing the route to the cleanup ticket. A bigger concern after the demo is using UUIDs for route params, as it quickly gets ugly for the user once we are a couple of params deep.

frontend/src/app/[locale]/formPrototype/[applicationId]/[formId]/page.tsx
Comment (acouch) [reply frontend/src/app/[locale]/formPrototype/[applicationId]/[formId]/page.tsx:126]: Yes, this is the saved state from the server that hasn't yet been loaded into the form. Do you have one in mind? It's not `formData` which is already accounted and "initial" or `formDataProps` doesn't seem more clear to me.

Comment (doug-s-nava) [reply frontend/src/app/[locale]/formPrototype/[applicationId]/[formId]/page.tsx:126]: something like `savedApplicationData`? `userSavedApplication`?

frontend/src/app/[locale]/formPrototype/[applicationId]/page.tsx
Comment (doug-s-nava) [reply frontend/src/app/[locale]/formPrototype/[applicationId]/page.tsx:1]: yeah, agreed - seeing those two uuids mashed up against each other in the url was what made me think of this

frontend/src/components/applyForm/ApplyForm.tsx
```diff
@@ -4,58 +4,80 @@ import { RJSFSchema } from "@rjsf/utils";
...
-        <FormGroup>{fields}</FormGroup>
+        <FormGroup>
+          <Fieldset legendStyle="large">{fields}</Fieldset>
+        </FormGroup>
         <p>
+          <Button
+            data-testid="apply-form-save"
+            type="submit"
+            secondary
+            name="apply-form-button"
```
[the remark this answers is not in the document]
doug-s-nava: ah, I see how it's being used here. Again the nomenclature is tough since one is a save button and one is an apply button, maybe it'd be clearer to use `submit-form-button`?

frontend/src/components/applyForm/ApplyFormSuccessMessage.tsx
doug-s-nava: The main scenario I'm thinking of where confusion could happen if the message isn't cleared is this one

1. user saves the application
2. user sees the save notification
3. user edits the application while the save notification is still present
4. user saves again
5. user sees the save notification
6. has the second save actually occurred or is this still the same message from the first save?

frontend/src/app/[locale]/formPrototype/[applicationId]/[formId]/page.tsx
```diff
@@ -31,31 +36,72 @@ interface formPageProps {
...
+  let applicationData = {} as ApplicationDetail;
 
   try {
     const response = await getFormDetails(formId);
     if (response.status_code !== 200) {
-      throw new Error(
-        `Error retrieving competition details: ${JSON.stringify(response.errors)}`,
+      console.error(
+        `Error retrieving form details for formID (${formId})`,
+        response,
```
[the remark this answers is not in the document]
Comment (doug-s-nava) [reply frontend/src/app/[locale]/formPrototype/[applicationId]/[formId]/page.tsx:48]: did you mean to change this?

frontend/src/services/fetch/fetchers/applicationFetcher.ts
doug-s-nava: yeah, something to discuss later on, we can see how things progress

frontend/src/components/applyForm/actions.ts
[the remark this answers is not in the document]
doug-s-nava: in that case if we can update this error to point to an error with retrieving the form rather than submitting the form that would be great

doug-s-nava (approved):

frontend/src/components/applyForm/ApplyFormSuccessMessage.tsx
acouch: The message gets cleared upon save / submit.

Issue #4253, opened 2025-03-19
Create a utility to fetch a legacy user from our staging tables


chouinar:
### Summary

Note this ticket requires https://github.com/HHS/simpler-grants-gov/issues/3551 to have been completed as it queries against the tables created in that ticket.

We want to create a utility that roughly looks like this function:
```py
def get_legacy_user_for_login_gov_id(login_gov_id: str, db_session: db.Session) -> TUserAccount | None: # Or whatever we named the staging table record
      pass
```

Given a login.gov ID, it can fetch a legacy user from our staging tables.

To fetch a user, we can do a rough query like:
```sql
select * from TUSER_ACCOUNT_MAPPER where EXT_USER_ID = <grants_gov_id> and SOURCE_TYPE = 'GOV'
```
Make sure to include the SOURCE_TYPE filter as the data has multiple different source types for different login systems, and we only care about login.gov for now.

Note that you'll need to join this record back to the TUSER_ACCOUNT table on USER_ACCOUNT_ID to return the user object.

If the user isn't found, just return `None`, we'll let the calling code determine what it wants to do in that case.

### Acceptance criteria

- [ ] Created this utility function - I recommend making a "legacy_user" folder in the services folder for this as we'll add more utilities in the future
- [ ] Added tests that verify the behavior

PR #4528, opened 2025-04-09
[Issue #4253] Create a utility to fetch a legacy user from staging tables

babebe:
## Summary
Fixes #{[4253](https://github.com/HHS/simpler-grants-gov/issues/4253)}

## Changes proposed
Utility Added
Factories Created/Updated
Added Test

### DIFF
--- a/api/tests/src/db/models/factories.py
+++ b/api/tests/src/db/models/factories.py
@@ -1938,7 +1938,7 @@ class TuserAccountFactory(BaseFactory):
         abstract = True
 
     user_account_id = factory.Sequence(lambda n: n)
-    user_id = factory.Sequence(lambda n: n)
+    user_id = Generators.UuidObj
     full_name = factory.Faker("name")
     email_address = factory.LazyAttribute(lambda o: f"{o.full_name}@example.com")
     last_upd_date = factory.Faker("date_time_between", start_date="-5y", end_date="now")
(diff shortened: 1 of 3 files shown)


api/tests/src/db/models/factories.py
```diff
@@ -1938,6 +1938,7 @@ class Meta:
         abstract = True
 
     user_account_id = factory.Sequence(lambda n: n)
+    user_id = factory.Sequence(lambda n: n)
```
chouinar, line 1941: Rather than an integer that needs the setup next sequence thing, make this a UUID, the value is actually a UUID in the real system, just shoved into a string.

You can do that on the Tuser factory as well.

api/tests/src/services/legacy_users/test_utils.py
```diff
@@ -0,0 +1,19 @@
+from src.services.legacy_users.utils import get_legacy_user_for_login_gov_id
+from tests.src.db.models.factories import (
+    LoginGovStateFactory,
+    StagingTuserAccountFactory,
+    StagingTuserAccountMapperFactory,
+)
+
+
+def test_get_legacy_user_for_login_gov_id(db_session, enable_factory_create):
```
chouinar, line 9: Could we add a test for it not being found?

api/tests/src/db/models/factories.py
babebe: update TuserAccountFactory and TuserAccountMapperFactory

chouinar (approved):

Issue #2631, opened 2024-10-29
Scope out monitoring of backend tasks in prod


chouinar:
### Summary

- Backend ELT process
- Search data load

### Acceptance criteria

- [ ] Implementation tickets created

chouinar:
I've been slow rolling this while waiting for us to get New Relic setup, but here are some rough notes.


# Failure alerts
We should be notified on any failed jobs

The jobs automatically log a message when they complete (success or failure)

### Time since last job run
We should alert if a job hasn't run in a certain period of time. Sometimes AWS experiences issues, or we may accidentally disable a job, and want to be aware of it. Additionally, a lot of our logging doesn't "work" if the job itself just fails in the startup process - a misconfigured job is the most likely cause of this alert. A slight delay shouldn't cause an alert.

We should alert on hourly jobs after 4 hours without a run.

We should alert on daily jobs after 4 hours without a run (we want to run every day).

Unknown: How best to alert on this? Something like `if most_run_log.timestamp > now - 4h: alert()`?

# Errors
We should alert if any errors occur (based on log level)

We can consider alerting on seeing stack traces, this has been iffy on past projects as backend processes often have simple try/retry logic especially when interacting with external services and we don't want to 


# Job Specific

## Load Transform (hourly)
https://github.com/HHS/simpler-grants-gov/blob/main/api/src/data_migration/command/load_transform.py

This job is made of several tasks, so will go through each individually

### Load Oracle Data
https://github.com/HHS/simpler-grants-gov/blob/main/api/src/data_migration/load/load_oracle_data_task.py

If an error occurs with loading a table, it gets skipped, but logs an error, so should be okay as is.

### Transform Oracle Data
Has several parts, all called from https://github.com/HHS/simpler-grants-gov/blob/main/api/src/data_migration/transformation/transform_oracle_data_task.py

Should alert if the `total_error_count` metric is ever non-zero as that indicates a record had to be skipped due to a data issue.

### Set Current Opportunities
https://github.com/HHS/simpler-grants-gov/blob/main/api/src/task/opportunities/set_current_opportunities_task.py

Could consider a total count metric - if this ever plummets that would be bad, but I don't think that's super necessary

## Load Opportunity Data (hourly)
https://github.com/HHS/simpler-grants-gov/blob/main/api/src/search/backend/load_search_data.py

Maybe a total count metric? We might want to hold on anything beyond the shared ones as we're going to make this job incremental soon-ish?

[...]

PR #4532, opened 2025-04-09
[Issue #2631] Add a few more tags

coilysiren:
## Context

More tags help to slice the data in new relic more effectively


mdragon (approved):
👍🏻

PR #4547, opened 2025-04-10
[no ticket] Adds back API analytics env vars

coilysiren:
## Context

I accidentally removed these env vars from the container while working on other things


mdragon (approved):
🚀

Issue #4529, opened 2025-04-09
Manually Test Opportunity Versioning in all Environments


babebe:
### Summary

Manually Test the load_transform 4th step `StoreOpportunityVersionTask` in dev/staging/prod

### Acceptance criteria

- [ ]  After running task, Opportunity Version Records are created in table `opportunity_version`

PR #4550, opened 2025-04-11
[Issue #4529] Add more logs to StoreOpportunityVersionTask [1/2]

babebe:
## Summary
Fixes #{4529}

## Changes proposed
Add more logging.
Moved check for draft status for better logging/metric


api/src/services/opportunities_v1/opportunity_version.py
```diff
@@ -26,9 +26,6 @@ def save_opportunity_version(db_session: db.Session, opportunity: Opportunity) -
     :return: This function does not return a value. It saves a new version of the opportunity in the database.
     """
 
-    if opportunity.is_draft:
-        return
```
chouinar, line 30: We can't/don't want to move this.

In the future this versioning function will be called by other places in the code, and if we take this out they also need to add this (and there likely will be more filters in the future).

One thing we could do to help with metrics, although not specifically this is to make this function return a boolean. If True, it created a new version, if False, it didn't. Doesn't give draft/not-draft, but at least can be used to get a count.

api/src/task/opportunities/store_opportunity_version_task.py
```diff
@@ -26,11 +33,26 @@ def run_task(self) -> None:
...
-            select(OpportunityChangeAudit).where(OpportunityChangeAudit.updated_at > latest_time)
+            select(OpportunityChangeAudit)
+            .where(OpportunityChangeAudit.updated_at > latest_time)
+            .options(selectinload("*"))
         ).all()
 
         for opp_change_audit in opportunity_change_audits:
+            log_extra = {
+                "opportunity_id": opp_change_audit.opportunity.opportunity_id,
+            }
```
chouinar, line 44: Not necessary, but you can also do something like this:
```suggestion
            log_extra = {
                "opportunity_id": opp_change_audit.opportunity_id,
            }
```

api/src/task/opportunities/store_opportunity_version_task.py
```diff
@@ -13,7 +15,12 @@
 
 
 class StoreOpportunityVersionTask(Task):
+    class Metrics(StrEnum):
+        OPPORTUNITIES_VERSIONED = "opportunities_versioned"
+        DRAFT_OPPORTUNITIES_SKIPPED = "draft_opportunities_skipped"
+
     def run_task(self) -> None:
+        logger.info("Fetching opportunities to version")
```
chouinar, line 23: Also, something that should have been in the initial version of this. I'd recommend adjusting run_task like so:

```py
def run_task(self) -> None:
     with self.db_session.begin():
            self.process_opportunity_versions()
```
and move the rest of this function into a new `process_opportunity_versions` function. As it's implemented right now, there isn't a point where it obviously commits to the DB which is a bit of an issue.

api/src/services/opportunities_v1/opportunity_version.py
babebe: Good point. Added bool.

api/src/task/opportunities/store_opportunity_version_task.py
babebe: Oh yes! Added

chouinar (approved):

PR #4553, opened 2025-04-11
[Issue #4546] Modify search logic to avoid querying for more than 10k records

### LINKED ISSUE #4546: Modify the search response logic to cap total results at 10,000
### Summary

This is just a change to the pagination info we return back to the user. This will prevent the frontend from trying to paginate past the 10000th item as OpenSearch won't let you query past the 10,000th thing even if it has more results:
https://opensearch.org/docs/latest/search-plugins/searching-data/paginate/

If the total records returned is more than 10,000, we should set the value to 10,000.

Additionally we need to adjust the total page value we calculate to instead be `math.floor` instead of `math.ceil` but only if that last page would contain records over 10k.

Because this logic is done as part of the search logic and testing it with 10k records would be annoying for a unit test, we should pull some of this logic into a utility and test that directly.

### Acceptance criteria

- [ ] Fixed logic for 10k+ on total_records and total_pages

chouinar:
## Summary
Fixes #4546

## Changes proposed
Modified the logic around parsing the search response to cap the total results at 10k

Modified the logic around creating a request to the search index to prevent a user from asking for anything past 10k

## Context for reviewers
If you try to query a search index for any record past 10k, it will error.

For example, if we tried to do this following query against the search index, it would error saying you tried to ask for record 10005, which is more than 10k. There is no way around this limit in a way that still supports pagination (there is a scroll API, but you have to iterate one page at a time).

```
GET opportunity-index-alias/_search
{
  "size": 10,
  "from": 9995
}
```

To work around this issue, we changed two things:
1. Even if the search index says it has more than 10k results, cap it to 10k (note our metrics for facet counts are not affected by this and can still be more than 10k). This way any user like our frontend that respects the page counts/total records won't hit any issues.
2. If a user does request a page that would go over 10k results then adjust that page's size so it only goes up to 10k, but not past it (in the above example, we'd change the size to `5` and it would work). If they fully request something past 10k, then we'll do a page size of 0, and start from 10k. This keeps a valid request and means things like facet counts still work.

## Additional information
In addition to the unit tests, I loaded more than 10k records into my local search index and queried it with a bunch of things around 10k and couldn't get any errors to happen.

### DIFF
--- a/api/openapi.generated.yml
+++ b/api/openapi.generated.yml
@@ -1426,7 +1426,7 @@ components:
           type:
           - object
           $ref: '#/components/schemas/HasActiveOpportunityFilterV1'
-    SortOrderAgencyPaginationV11:
+    SortOrderAgencySearchPaginationV1:
       type: object
       properties:
         order_by:
@@ -1445,7 +1445,7 @@ components:
       required:
       - order_by
       - sort_direction
-    AgencyPaginationV11:
+    AgencySearchPaginationV1:
       type: object
       properties:
         sort_order:
@@ -1459,7 +1459,7 @@ components:
           items:
             type:
             - object
-            $ref: '#/components/schemas/SortOrderAgencyPaginationV11'
+            $ref: '#/components/schemas/SortOrderAgencySearchPaginationV1'
         page_size:
           type: integer
           minimum: 1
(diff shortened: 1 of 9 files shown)


mdragon (approved):
Looks good

PR #4554, opened 2025-04-11
[Issue #4477] Setup TCompetition table in staging & legacy schemas

### LINKED ISSUE #4477: Setup TCompetition table in staging & legacy schemas
### Summary

TODO - docs about this are being written, but the columns we want to setup in exactly this order

* COMP_ID - int, primary key
* OPP_CFDA_ID - int
* COMPETITIONID - str
* FAMILYID - int
* COMPETITIONTITLE - str 
* OPENINGDATE - date
* CLOSINGDATE -  date
* CONTACTINFO - str
* GRACEPERIOD - int
* OPENTOAPPLICANTTYPE - int
* DIALECT - str, NOT NULL
* LAST_UPDATE - datetime
* ELECTRONIC_REQUIRED - str
* EXPECTED_APPL_NUM - int
* EXPECTED_APPL_SIZE -int
* ORIGCFDANUM - str
* ORIGOPPNUMBER - str
* CREATED_DATE - datetime
* LAST_UPD_DATE - datetime
* CREATOR_ID - str
* LAST_UPD_ID - str
* ISMULTI - str
* AGENCY_DWNLD_URL - str
* PACKAGE_ID - str
* IS_WRKSPC_COMPATIBLE - str, NOT NULL
* SENDMAIL - str
* MODIFICATION_COMMENTS - str

### Acceptance criteria

- [ ] Staging and Legacy table set
- [ ] Migrations created, note that only the staging table will get created automatically

mikehgrantsgov:
## Summary
Fixes #4477

## Changes proposed
Add competition_mixin for legacy and staging schemas.
Migration created

## Context for reviewers
Columns in specific order.


api/src/db/models/legacy_mixin/competition_mixin.py
```diff
@@ -0,0 +1,34 @@
...
+
+@declarative_mixin
+class TcompetitionMixin:
+    comp_id: Mapped[int | None] = mapped_column(primary_key=True)
+    opp_cfda_id: Mapped[int | None]
+    competitionid: Mapped[str | None]
+    familyid: Mapped[int | None]
+    competitiontitle: Mapped[str | None]
+    openingdate: Mapped[datetime | None]
+    closingdate: Mapped[datetime | None]
```
chouinar, line 14: I just double-checked these in the Oracle DB, these are actually real dates, so we should use the date type instead of datetime. The other date fields in this table are actually datetime as I said.

chouinar (commented):
Just one small thing on types - I should've looked at the types closer when I wrote up the ticket.

api/src/db/models/legacy_mixin/competition_mixin.py
mikehgrantsgov: Updated this

chouinar (approved):

PR #4556, opened 2025-04-11
[Issue #4555] Add New Relic support to Analytics code

### LINKED ISSUE #4555: Modify analytics code to better support New Relic
### Summary

Copy features over that we've added recently in the API to make New Relic support work.

### Acceptance criteria

- [ ] New Relic working

chouinar:
## Summary
Fixes #4555

## Changes proposed
Add NewRelic setup to Analytics code including:
* ini file
* Initializing New Relic at startup with that ini file
* Marking the tasks with the Background task designation
* Adding New Relic metadata to all log messages

Non-locally disable the pretty-print error messages

Make all tasks also be wrapped by ecs_background_task

## Context for reviewers
A lot of this is copied from the API code, just without some of the API-specific things configured (ie. we don't need certain New Relic features if it's not an API serving requests).

Perhaps the most important thing here is the `@ecs_background_task` added to each of the tasks, this accomplishes a few things:
* When a process errors, it makes it so it will **ALWAYS** emit at least one error level log. (It adds a try-except around the entire function and logs + rethrows the error). This means we can easily find logs in New Relic
* Configures the logging to work with New Relic a bit better by marking it as a background task
* Adds some general info to all logs like ECS info, timing metrics, and more.

`pretty_exceptions_enable` is a feature of Typer that prints a convenient error message for you and dumps the local variables, but that gets a bit broken in New Relic + could log secret values, so is disabled by default (locally it's re-enabled).

### DIFF
--- a/.dockleconfig
+++ b/.dockleconfig
@@ -3,4 +3,4 @@
 # DOCKLE_ACCEPT_FILES="file1,path/to/file2,file3/path,etc"
 # https://github.com/goodwithtech/dockle#accept-suspicious-environment-variables--files--file-extensions
 # The apiflask/settings file is a stub file that apiflask creates, and has no sensitive data in. We are ignoring it since it is unused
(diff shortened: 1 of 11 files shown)


coilysiren (approved):
`o11y` 💹

coilysiren (approved):
✅

PR #4568, opened 2025-04-14
[Issue #4530] Load opportunity attachment text extraction

### LINKED ISSUE #4530: Implement support for extracting text from opportunity attachments
### Summary

After investigation, we need to implement support for extracting text from opportunity attachments.

We need to implement text extraction for the following file types:
```python
ALLOWED_ATTACHMENT_SUFFIXES = set(
    ["txt", "pdf", "docx", "doc", "xlsx", "xlsm", "html", "htm", "pptx", "rtf"]
)
```

[This library](https://github.com/deanmalmgren/textract) can be used as a general reference for potential libraries needed for the text extraction by file type.

### Acceptance criteria

- [ ] extract the appropriate text from each of the above file types
- [x] the appropriate text extracting logic is invoked for loading opportunities into search index
- [x] txt
- [x] pdf
- [x] html
- [x] htm
- [x] docx
- [ ] doc
- [x] xlsx
- [x] xlsm
- [x] pptx
- [x] rtf

freddieyebra:
## Summary
Fixes #4530 

## Changes proposed
Add support for extracting text on text based file types to load into our search index. 

## Context for reviewers
It will be beneficial to run a test run in lower envs to populate the search index with opportunity attachments using this extractor.


api/src/util/text_extractor.py
```diff
@@ -0,0 +1,140 @@
...
+            "html": html_extractor_config,
+            "htm": html_extractor_config,
+            "rtf": TextExtractorConfig(extractor=extract_text_from_rft),
+            "xlsm": xls_extractor_config,
+            "xlsx": xls_extractor_config,
+        }[self.file_type]
+
+    def _validate_file_type(self) -> None:
+        if self.file_type.lower() not in TEXT_EXTRACTOR_SUPPORTED_FILE_TYPES:
+            raise UnsupportedTextExtractorFileType(f"Unsupported file type: {self.file_type}")
```
chouinar, line 130: Just to avoid the configs getting out of sync if we ever add more file types, is there any way we could avoid defining `TEXT_EXTRACTOR_SUPPORTED_FILE_TYPES` and just use the config's keys?

It seems like the config is statically defined, so maybe we can just define that at the top and do something like:
```py
TEXT_EXTRACTOR_SUPPORTED_FILE_TYPES = set(text_extractor_config.keys())
```

api/tests/src/util/test_text_extractor.py
```diff
@@ -0,0 +1,44 @@
...
+TEST_FILE_DIR = "/text_extractor_test_files"
+
+
+def test_text_extractor_errors_on_invalid_file_type() -> None:
+    with pytest.raises(UnsupportedTextExtractorFileType):
+        TextExtractor("/pathtounsupported/filetype.movie")
+
+
+def test_extract_text_from_file_returns_none_unsupported_file_type() -> None:
+    assert extract_text_from_file("unsopported_file_type.not_today") is None
```
chouinar, line 18:
```suggestion
    assert extract_text_from_file("unsupported_file_type.not_today") is None
```

api/tests/src/util/test_text_extractor.py
```diff
@@ -0,0 +1,44 @@
...
+            f"{TEST_FILE_DIR}/xlsx_data.xlsx",
+            "sheet1\nh1,h2\nxlsx,data\n\nsheet2\nh3,h4\nandxlsm,data",
+        ),
+        (
+            f"{TEST_FILE_DIR}/xlsm_data.xlsm",
+            "sheet1\nh1,h2\nxlsx,data\n\nsheet2\nh3,h4\nandxlsm,data",
+        ),
+    ],
+)
+def test_extract_pdf_text(fixture_file_path, fixture_file_path_val, expected) -> None:
```
chouinar, line 41: One scenario to test, what is the behavior if we try to parse a file and the suffix is wrong (eg. my "file.pdf" is actually a text file or ppt). That's a lot of possible ways it could go wrong, we don't need to test them all, as long as they all fail relatively the same way so we can skip those files.

chouinar (commented):
Looking good so far

api/src/util/text_extractor.py
freddieyebra: Thanks Michael. I like that idea!

freddieyebra:
Created a follow up ticket for support doc file types [here](https://github.com/HHS/simpler-grants-gov/issues/4683)

api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -347,17 +344,15 @@ def get_attachment_json_for_opportunity(
...
-                    )
+                file_text = extract_text_from_file(att.file_location)
+                attachments.append(
+                    {
+                        "filename": att.file_name,
+                        "attachment": {
+                            "content": file_text,
+                        },
+                    }
+                )
```
chouinar, line 355: Two thoughts here:
1. We should have a try/except wrapping this - if a single attachment is bad (I'm going to guess at least some are misnamed), we shouldn't break the process, just log a warning/increment a counter and continue.
2. We might want to consider adding a limit to the size of the response from the extract logic. A 200 page PDF we might be fine cutting off at a certain point - the ingest attachment stuff we're replacing had a limit of 100,000 by default: https://docs.opensearch.org/docs/latest/install-and-configure/additional-plugins/ingest-attachment-plugin/

api/src/util/text_extractor.py
```diff
@@ -0,0 +1,157 @@
...
+    file_path: str, file_type: str | None = None, raise_on_error: bool = False
+) -> str | None:
+    try:
+        return TextExtractor(file_path, file_type=file_type).get_text()
+    except Exception as e:
+        if raise_on_error:
+            logger.error(e)
+            raise e
+        logger.warning(e)
+        return None
```
chouinar, line 40: I'd leave the logging out of this actually - let the calling code handle logging. In some processes we might want warnings/errors, and others we might not. Best to leave that up to the calling code.

Note that as far as our alerts are concerned, warnings/errors both should trip an alert.

api/src/util/text_extractor.py
```diff
@@ -0,0 +1,157 @@
...
+    for slide in presentation.slides:
+        for shape in slide.shapes:
+            if not shape.has_text_frame:
+                continue
+            for paragraph in shape.text_frame.paragraphs:
+                for run in paragraph.runs:
+                    if text := run.text.strip():
+                        text_runs.append(text)
+    data = "\n".join(text_runs)
+    stream.close()
```
chouinar, line 71:
```suggestion
    with BytesIO(file_data) as stream:
        presentation = pptx.Presentation(stream)
        text_runs = []
        for slide in presentation.slides:
            for shape in slide.shapes:
                if not shape.has_text_frame:
                    continue
                for paragraph in shape.text_frame.paragraphs:
                    for run in paragraph.runs:
                        if text := run.text.strip():
                            text_runs.append(text)
        data = "\n".join(text_runs)

```

api/src/util/text_extractor.py
```diff
@@ -0,0 +1,157 @@
...
+            "docx": TextExtractorConfig(extractor=extract_text_from_docx, read_mode="rb"),
+            "pdf": TextExtractorConfig(extractor=extract_text_from_pdf, read_mode="rb"),
+            "pptx": TextExtractorConfig(extractor=extract_text_from_pptx, read_mode="rb"),
+            "txt": TextExtractorConfig(),
+            "html": html_extractor_config,
+            "htm": html_extractor_config,
+            "rtf": TextExtractorConfig(extractor=extract_text_from_rft),
+            "xlsm": xls_extractor_config,
+            "xlsx": xls_extractor_config,
+        }
```
chouinar, line 133: This looks like it is static, but every file we would process would need to rebuild this config multiple times - every time `get_configs` is called.

Unless there is a reason this config would be modified, could we just define this at the top of the file once?

api/src/util/text_extractor.py
```diff
@@ -0,0 +1,157 @@
...
+    stream.close()
+    return data
+
+
+def extract_text_from_rft(rtf_data: str) -> str:
+    return rtf_to_text(rtf_data)
+
+
+def extract_text_from_docx(file_data: bytes) -> str:
+    return "\n".join(paragraph.text for paragraph in Document(BytesIO(file_data)).paragraphs)
```
chouinar, line 80: Does this close the byte streams it opens?

api/src/util/text_extractor.py
```diff
@@ -0,0 +1,157 @@
...
+        logger.warning(e)
+        return None
+
+
+def extract_text_from_html(html_string: str) -> str:
+    return "\n".join(BeautifulSoup(html_string, "html.parser").stripped_strings)
+
+
+def extract_text_from_pdf(file_data: bytes) -> str:
+    reader = PdfReader(BytesIO(file_data))
```
chouinar, line 48: Does this close the bytesIO stream?

api/src/util/text_extractor.py
```diff
@@ -0,0 +1,157 @@
...
+            "html": html_extractor_config,
+            "htm": html_extractor_config,
+            "rtf": TextExtractorConfig(extractor=extract_text_from_rft),
+            "xlsm": xls_extractor_config,
+            "xlsx": xls_extractor_config,
+        }
+
+    def _read_file_data(self) -> bytes | str:
+        if self.config.reader:
+            return self.config.reader(self.file_path)
```
chouinar, line 137: If there is a reader, it also needs to be opened with `open_stream` - we can't pass a file path around as the file path is on s3 and open_stream handles that. If the underlying readers need a file path and can't take in a file object, then we can't use them.

api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -346,18 +346,29 @@ def get_attachment_json_for_opportunity(
...
+                self.increment(self.Metrics.ATTACHMENTS_PROCESSED)
+            except Exception as e:
+                self.increment(self.Metrics.ATTACHMENTS_UNPROCESSED)
+                logger.warning(
+                    "text-extractor: error extracting text", extra={"err": e, "attachment": att}
```
chouinar, line 370: Could we put the attachment ID instead (+ add opportunity ID while we're at it)? I think what this would do at the moment is look like (with some random mock data locally): `<OpportunityAttachment(attachment_id=7,opportunity_id=113,file_location='s3://local-mock-public-bucket/opportunities/113/attachments/83321342/whole.txt',mime_type='model/x3d+vrml',file_name='whole.txt' ...)>`

This wouldn't fit too well into the JSON object/be searchable.

api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -346,18 +346,29 @@ def get_attachment_json_for_opportunity(
...
+                    {
+                        "filename": att.file_name,
+                        "attachment": {
+                            "content": file_text,
+                        },
+                    }
+                )
+                self.increment(self.Metrics.ATTACHMENTS_PROCESSED)
+            except Exception as e:
+                self.increment(self.Metrics.ATTACHMENTS_UNPROCESSED)
```
chouinar, line 368: Instead of unprocessed, could we add a separate metric for error count?

api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -346,18 +346,29 @@ def get_attachment_json_for_opportunity(
...
+                attachments.append(
+                    {
+                        "filename": att.file_name,
+                        "attachment": {
+                            "content": file_text,
+                        },
+                    }
+                )
+                self.increment(self.Metrics.ATTACHMENTS_PROCESSED)
+            except Exception as e:
```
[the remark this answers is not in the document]
freddieyebra: For this iteration, we could consider any failed/unprocessed attachments as non-blocking and consider the opportunity processed and re-run full load once we have doc support. I am not too familiar with how the task queue for this job is managed so i defer to you on that.

api/tests/src/util/test_text_extractor.py
freddieyebra: added the UnprocessableFileContent exception for this case.

api/src/util/text_extractor.py
```diff
@@ -15,6 +15,9 @@
 
 logger = logging.getLogger(__name__)
 
+TEXT_EXTRACTOR_LOG_PREFIX = "TextExtractor:"
```
chouinar, line 18: Our logs have the file_path/func/line number so we know where they were emitted from - do we need this?

api/src/util/text_extractor.py
```diff
@@ -131,30 +195,31 @@ def __init__(
         self.config = get_text_extractor_configs()[self.file_type]
 
     def get_text(self) -> str:
+        logger.info(f"{TEXT_EXTRACTOR_LOG_PREFIX} start: {self.file_path=} {self.file_type=}")
         try:
-            return self.config.extractor(self._read_file_data()).strip()
+            extracted_data = self.config.extractor(self._read_file_data()).strip()
+            logger.info(
+                f"{TEXT_EXTRACTOR_LOG_PREFIX} success: {self.file_path=} {self.file_type=} characters_extracted={len(extracted_data)}"
+            )
```
chouinar, line 203: Don't try to shove info into the log message, log messages should be kept generic, put it in the extra object so it's easily parsed out.

api/tests/src/util/test_text_extractor.py
```diff
@@ -65,3 +66,14 @@ def test_extract_pdf_text(fixture_file_path, fixture_file_path_val, expected) ->
...
+
+def test_unprocessable_file_content(fixture_file_path):
+    with pytest.raises(UnprocessableFileContent, match="UnprocessableFileContent"):
+        extract_text_from_file(
+            fixture_file_path(f"{TEST_FILE_DIR}/bytes_named_as_text.txt"), raise_on_error=True
+        )
+        extract_text_from_file(
+            fixture_file_path(f"{TEST_FILE_DIR}/text_data_as_bytes_content.pdf"),
+            raise_on_error=True,
+        )
```
chouinar, line 79: If you're trying to test that both of these error, put them in separate `pytest.raises` blocks.

chouinar (approved):
LGTM! We'll need to test this a bit in the lower environments, hopefully it runs a bit faster than prior attempts.

PR #4587, opened 2025-04-15
[no ticket] Improves step function consistency

coilysiren:
## Motivation & Context

Infrastructure will always have a set of concerns regarding scheduled jobs that are slightly different than application layer concerns. Specifically, infrastructure will want to know if any of their changes have broken scheduled jobs in the recent past, via a view that displays all the scheduled jobs at once. In this view, all the jobs should have the same name on the application layer as they do on the infrastructure layer. There can be a different key that the application layer uses (`task_name` or whatever else) but the infrastructure layer dashboards are gonna look for this now, eg. `SCHEDULED_JOB_NAME` / `scheduled_job_name`

### DIFF
--- a/infra/modules/service/scheduled_jobs.tf
+++ b/infra/modules/service/scheduled_jobs.tf
@@ -1,13 +1,3 @@
-locals {
-  scheduled_job_environment_variables = [
-    for name, value in var.scheduled_jobs :
-    {
-      name  = "SCHEDULED_JOB_NAME",
-      value = name
-    }
-  ]
-}
-
 resource "aws_scheduler_schedule" "scheduled_jobs" {
   for_each = var.scheduled_jobs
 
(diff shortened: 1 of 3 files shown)


infra/modules/service/scheduled_jobs.tf
```diff
@@ -1,3 +1,13 @@
+locals {
+  scheduled_job_environment_variables = [
+    for name, value in var.scheduled_jobs :
+    {
+      name  = "SCHEDULED_JOB_NAME",
+      value = name
+    }
+  ]
+}
```
chouinar, line 9: Could be that I'm not familiar with terraform, but how does this individually give each job its specific environment variable? This reads like it would set the same env vars for every job?

coilysiren: It's going to be the same env var for each job, yeah. I just wanna have this env var piped straight into the logs into a `scheduled_job_name` json key

coilysiren: Oh yeah I did this wrong, not on my game today 🤦🏼‍♀️ please hold

coilysiren: image

chouinar (approved):

Issue #4551, opened 2025-04-11
Manual Test for Loading Agency into search index in all Environment


babebe:
### Summary

Load agencies into the search index in dev/staging
Verify that agencies are searchable in dev/staging
Load agencies into the search index in prod

### Acceptance criteria

[ ] All agencies are successfully indexed in each environment (dev, staging, prod)
[ ] v1/agencies/search endpoint returns valid results for expected queries in dev/staging

PR #4589, opened 2025-04-15
[Issue #4551] Add filter for Test agency

babebe:
## Summary
Fixes #{[4551](https://github.com/HHS/simpler-grants-gov/issues/4551)}

## Changes proposed
add is_test_agency to the filter options.


api/src/api/agencies_v1/agency_schema.py
```diff
@@ -102,6 +105,12 @@ class AgencyV1Schema(Schema):
             "example": "False",
         },
     )
+    is_test_agency = fields.Boolean(
+        metadata={
+            "description": "Indicates if the agency is a test agency .",
+            "example": "False",
```
chouinar, line 111: While it works because of the JSON parsing library we have, this (and the example added for `has_active_opportunity` above) mean the examples we end up putting in openapi has a value of `"False"` which is only valid in python and isn't actually valid JSON. just pass in a boolean like `False` instead of `"False"`
```suggestion
            "example": False,
```

chouinar (approved):

PR #4590, opened 2025-04-15
[Unticketed] Add step function name and short error logs

chouinar:
## Changes proposed
If a log message has `exc_info` put a short form of it in the logs as well for slightly nicer error log messages in New Relic

If running via a step function, add the name of it to the logs (recently piped through in terraform)

## Context for reviewers
By default, the `exc_info` looks like `(<class 'ValueError'>, ValueError('hello'), <traceback object at 0x1160d4640>)` and the traceback is generally really long. This is useful for debugging, and we'll keep that, but it's usually like 30 lines long and doesn't fit neatly into a dashboard. So I added `exc_info_short` that is the middle of those without the traceback.

## Additional information
Screenshot 2025-04-15 at 2 47 25 PM


coilysiren (approved):
🚢

PR #4591, opened 2025-04-15
[Issue #4426] Form prototype enhancements

### LINKED ISSUE #4426: Prepare form for demo
### Summary

Some follow-up tasks that are not captured in the other tickets. These are all "nice to haves":

- [ ] document ui schema
- [x] add testing for utils
- [x] remove CSV files and test schema files
- [ ] validate the project start and end dates
- [ ]  checkbox field
- [x] read-only fields
- [x] update ui schema to more consistently show headings
- [x] update routes

### Acceptance criteria

- [ ]  nice to have's addressed

### LINKED ISSUE #4679: Form demo follow-ups
- [x] document ui schema
- [x] add remaining tests
- [x] document ui schema
- [x]  ~~checkbox field~~ moved to #4807

acouch:
## Summary
Fixes #4679

## Changes proposed

1. Adds README.md explaining UI schema
2. Moves schema validation from zod to JSON Schema
  * zod didn't hold up with the need for recursive nesting of the form schema and the `anyOf` option for the field or the section entries
3. page updates which free the `<BetaAlert/>` from the `<GridContainer/>`
4. Adds feature flag to prototype landing page
5. Updates the `<SelectWidget/>` to use the `<ComboBox/>` instead of `<Select/>` if there are more than 15 enum options
6. removes `validateFormSchema()`. This was initially done by Zod. To verify that a json object conforms to JSON Schema requires some additional effort that doesn't seem worthwhile as JSON schema objects can take almost any shape. The form degrades gracefully if the form data is malformed which is tested.

## Context for reviewers
This is still in a demo phase, so on auth or designs.

## Additional information
NA


frontend/src/components/applyForm/customValidation.ts
acouch, line 1: This was added as a placeholder as I expected to add a validation that the start date would not be before the end date. We've punted on that for now as the logic would ideally be handled by the API.

frontend/src/app/[locale]/formPrototype/[applicationId]/form/[formId]/page.tsx
```diff
@@ -83,46 +83,49 @@ async function FormPage({ params }: formPageProps) {
...
-        savedFormData={application_response}
-        formSchema={formSchema}
-        uiSchema={form_ui_schema}
-        formId={form_id}
-        applicationId={applicationId}
-      />
-    </GridContainer>
+    <>
+      <BetaAlert containerClasses="margin-top-5" />
+      <GridContainer>
```
Comment (acouch) [inline frontend/src/app/[locale]/formPrototype/[applicationId]/form/[formId]/page.tsx:105]: Only change here is to move the `<GridContainer>` so it is not wrapping the `<BetaAlert/>`

frontend/src/components/applyForm/ApplyForm.tsx
```diff
@@ -42,14 +42,22 @@ const ApplyForm = ({
...
-    errors: validationErrors,
-    formData: formObject,
-    schema: formSchema,
-    uiSchema,
-  });
-
   const navFields = useMemo(() => getFieldsForNav(uiSchema), [uiSchema]);
+  let fields: JSX.Element[] = [];
+  try {
+    fields = buildForTreeRecursive({
```
doug-s-nava, line 48: recognize this is an annoying ask at this point where it's already used a bunch, but can we fix the typo now? Every time I look at it I get confused 😅 `buildFormTreeRecursive`

frontend/src/components/applyForm/README.md
```diff
@@ -0,0 +1,78 @@
+# Apply form component
```
doug-s-nava, line 1: ❤️

frontend/src/components/applyForm/README.md
```diff
@@ -0,0 +1,78 @@
...
+    definition: "/properties/TestField",
+    schema: {
+        description: "This one is different"
+    }
+}
+```
+
+This can be used for descriptions or enums, but the form will fail if the UI Schema tries to change the field type as the validation is still determined by the form schema.
+
+A field can only exist with the `schema` key in the UI schema, however it won't be saved:
```
doug-s-nava, line 38: I'm not sure what this means - is the idea that you can provide a field in the schema that is informational or internal and that won't be saved when the form is saved or submitted?

frontend/src/components/applyForm/README.md
```diff
@@ -0,0 +1,78 @@
...
+The UI Schema is an array of fields or sections.
+
+### Fields
+
+Fields can either reference a field in the form schema:
+
+```
+{
+    type: "field",
+    definition: "/properties/TestField",
```
doug-s-nava, line 20: looking at the example schema from the DB I think this makes sense - since all of the individual fields are nested under `properties`, this is pointing to something like the field defined as "TestField" here?
```
  "properties": {
    "TestField": {
      "type": "string",
      "title": "Date of application ",
      "format": "date"
    },
```
It may help to provide an example with a bit more context like this to show this would be used

frontend/tests/components/applyForm/ApplyForm.test.tsx
```diff
@@ -124,4 +124,20 @@ describe("ApplyForm", () => {
       expect.any(FormData),
     );
   });
+  it("errors when form data is bad", () => {
```
doug-s-nava, line 127: trying to figure out the error case here - it looks like it's the empty object `formSchema` prop that's causing the error rather than `savedFormData`. If that's true can that be more clear in the name of the test? Related, but separate question for later - is there also a possible error case where the formSchema is valid but the savedFormData is bad?

doug-s-nava (dismissed):
still seems to work as it did before, so unless there is something new to test, I'm good to go forward with this (with a few non-blocking comments and suggestions)

Also, are the empty files in here intentional?

✅ ✅ ✅ ✅

frontend/src/components/applyForm/README.md
acouch: I added this example. Additionally tried to clarify:

> A field can have a `schema` without a `definition`. However, this field will only exist in the UI. When the form is saved the field will not be included in the form data sent to the server.

frontend/src/components/applyForm/ApplyForm.tsx
acouch: Updated, thanks for re-spotting.

frontend/src/components/applyForm/README.md
acouch: Thanks for asking. I updated to clarify:

> A field can have a schema without a definition. However, this field will only exist in the UI. When the form is saved the field will not be included in the form data sent to the server.

Does that help?

frontend/tests/components/applyForm/ApplyForm.test.tsx
acouch: I updated the test to be more explicit and contain cases of empty objects and ones that are not JSON Schema compliant.

> Related, but separate question for later - is there also a possible error case where the formSchema is valid but the savedFormData is bad?

This should fail gracefully. The form schema is the source of truth. Down the line we might want to make the form fail if the data does not conform to the JSON schema.

I included a test to verify the above.

doug-s-nava (approved):

PR #4605, opened 2025-04-16
[Issue #4564] Modify process that turns .DAT files into JsonSchema to populate the title from the Field Label

### LINKED ISSUE #4564: Modify process that turns .DAT files into JsonSchema to populate the title from the Field Label
### Summary

Right now we populate it from the Field Id, but Field Label would be cleaner for the UI

### Acceptance criteria

- [ ] Script updated

mikehgrantsgov:
## Summary
Fixes #4564

## Changes proposed
Use field label instead of field ID in generating our schema files. 
Update tests

## Context for reviewers
The `title` field in the schema is duplicative. Should this be changed / removed? Or stay as is?

## Additional information
Modified unit tests to pass given the new schema fields.

### DIFF
--- a/api/src/form_schema/csv_to_jsonschema.py
+++ b/api/src/form_schema/csv_to_jsonschema.py
@@ -135,7 +135,7 @@ def add_field_to_builder(builder: JsonSchemaBuilder, field_info: FieldInfo) -> N
     # Add appropriate property based on type
     if field_type == "Radio Group" or data_type == "LIST" or list_of_values:
         builder.add_string_property(
-            field_id,
+            title,
             is_nullable=is_nullable,
             is_required=required,
             title=title,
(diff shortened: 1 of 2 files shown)


api/src/form_schema/csv_to_jsonschema.py
```diff
@@ -135,7 +135,7 @@ def add_field_to_builder(builder: JsonSchemaBuilder, field_info: FieldInfo) -> N
     # Add appropriate property based on type
     if field_type == "Radio Group" or data_type == "LIST" or list_of_values:
         builder.add_string_property(
-            field_id,
+            title,
```
chouinar, line 138: We don't want to change the name of the field, we want the title to be based on the field label column. We'd want to change the title we're setting here, not this field

chouinar: For example we currently generate:
```
        "FederalAgency": {
            "type": "string",
            "title": "1. NAME OF FEDERAL AGENCY:",
            "description": "Pre-populated from the Application cover sheet.",
            "minLength": 1,
            "maxLength": 60
        },
```
We want title to instead be `Name of Federal Agency` which comes from the `Field ID` column in the .dat file

mikehgrantsgov: Ah, closed this and opened the much simpler: https://github.com/HHS/simpler-grants-gov/pull/4611/files

PR #4607, opened 2025-04-16
[Issue #4551] Enable load agency cron job

babebe:
## Summary
Fixes #{[4551](https://github.com/HHS/simpler-grants-gov/issues/4551)}

## Changes proposed
Override values for shard count and replica count in prod
Set Job to run every morning 8:00 AM EST top of the hour 9:00 AM non-dst


coilysiren (approved):
I have questions but they're non-blocking and I would prefer to merge in this particular case

PR #4623, opened 2025-04-17
[Issue #4618] Add dv verification for ssl certs

### LINKED ISSUE #4618: Configure ssl certs for interacting with applicants and grantors soap apis in training environment
### Summary

We have purchased the necessary ssl certificates for registering with the grants.gov soap api. We need to configure these certs on our staging domains.

### Acceptance criteria

- [ ] ssl certs configured in staging

freddieyebra:
## Summary
Fixes #4618

## Changes proposed
This pr add a route for verifying DV ssl certs. This route just renders static .txt files specified from the certificate providers.

## Context for reviewers
This is required for setting up SSL certs for grants.gov SOAP API. This pr adds verification content for grantors and applicants soap api in training environment.


api/src/static/domain_verification_assets/05589A9E44482CE51E7B9531194F4087.txt
```diff
@@ -0,0 +1,3 @@
+16FF65F1826CF79D30889347340FE9B9307501DC3F7E24ADB7DC574297D55B57
+sectigo.com
+95n55B5q5wIF5a55bM55
```
coilysiren, line 3: These look like secrets, please don't commit them into the repo 😅 

Strictly they are not secrets, but they very much look like secrets.

api/src/app.py
```diff
@@ -191,3 +193,18 @@ def robots() -> str:
...
         """
+
+
+def register_well_known(app: APIFlask) -> None:
+    @app.route("/.well-known/pki-validation/<file_name>")
+    @app.doc(hide=True)
+    def get_dv_verification_content(file_name: str) -> tuple:
+        try:
+            with open(
+                f"{Path(__file__).resolve().parent}/static/domain_verification_assets/{file_name}"
```
coilysiren, line 204: I feel like, as written, this can easily expose and random file on our server

Like so:

```
curl api.simpler.grants.gov/.well-known/pki-validation/../../../../secret-file
```

api/src/static/domain_verification_assets/05589A9E44482CE51E7B9531194F4087.txt
chouinar: To be safe, even if they're the "public" value we should make them env vars and store them in parameter store.

api/src/app.py
chouinar: Assuming this route needs to be shaped this way, and take in a file/key name, I'd probably aim to make a mapping of keys to something loaded from an env var.

api/src/static/domain_verification_assets/05589A9E44482CE51E7B9531194F4087.txt
freddieyebra: nuked

api/src/app.py
```diff
@@ -191,3 +193,18 @@ def robots() -> str:
...
+            ) as f:
+                return f.read(), 200
+        except Exception as e:
+            message = f"Could not read {file_name} dv contents: {e}"
+            logger.warning(message)
```
chouinar, line 209: I wouldn't make this a warning. Anything above info we treat as something to alert us - 4xx scenarios aren't a problem for us (eg. we wouldn't want to log a warning/error if a user hits the GET /opportunity endpoint with an ID that doesn't exist).

api/src/app.py
```diff
@@ -52,6 +53,17 @@ class EndpointConfig(PydanticBaseEnvConfig):
     auth_endpoint: bool = Field(False, alias="ENABLE_AUTH_ENDPOINT")
 
     enable_apply_endpoints: bool = Field(False, alias="ENABLE_APPLY_ENDPOINTS")
+    domain_verification_content: str | None = Field(None, alias="DOMAIN_VERIFICATION_CONTENT")
+    domain_verification_map: dict | None = None
+
+    def model_post_init(self, _context: Any) -> None:
+        self.domain_verification_map = {}
```
chouinar, line 60: What if we just do this - make the value never None, which cleans up a few of the type checks you do with it.
```suggestion
    domain_verification_map: dict = Field(default_factory=dict)

    def model_post_init(self, _context: Any) -> None:
```

chouinar (approved):

Issue #4440, opened 2025-04-03
Simpler SOAP - Applicants SOAP API Integration [DRAFT]


freddieyebra:
### Overview
Now that we are able to handle incoming SOAP requests, we need to support integrating simpler.grants.gov data flow into the grants.gov applicants SOAP API data that is passed in and out the Simpler SOAP API. This epic represents the work needed to integrate and support the data from both simpler.grants.gov and the legacy grants.gov system for the Applicants Simpler SOAP API.

This epic will also include some of the groundwork functionality around processing, transforming, and persisting of the data flowing within these 2 systems, as well as auth.

### Applicants SOAP API
The following will need to be integrated and available at our `<SGG BASE API URL>/grantsws-applicant/services/v2/ApplicantWebServicesSoapPort`:

**Standard Operations**
- [Get Opportunities](https://www.grants.gov/system-to-system/applicant-system-to-system/web-services/get-opportunity-list)
- [Submit Application](https://www.grants.gov/system-to-system/applicant-system-to-system/web-services/submit-application)
- [Get Submitted Applications](https://www.grants.gov/system-to-system/applicant-system-to-system/web-services/get-submission-list)
- [Get Application Info](https://www.grants.gov/system-to-system/applicant-system-to-system/web-services/get-application-info)
- [Get Application Zip](https://www.grants.gov/system-to-system/applicant-system-to-system/web-services/get-application-zip)

**NIH Specific Operations (TBD)**
- [Authentication](https://www.grants.gov/system-to-system/applicant-system-to-system/web-services/authenticate-aor-expanded)
- [Submit Application as 3rd Party](https://www.grants.gov/system-to-system/applicant-system-to-system/web-services/submit-application-as-third-party-expanded)
- [Get Submissions as 3rd Party](https://www.grants.gov/system-to-system/applicant-system-to-system/web-services/get-submission-list-as-third-party)
- [Get Application Info as 3rd Party](https://www.grants.gov/system-to-system/applicant-system-to-system/web-services/get-application-info-as-third-party)

### Parking Lot:

**Data sync lag**
Discuss how/if we want to implement this API such that there will be a known lag in data synchronization when **new** data is created through this API into SGG database. For example, if an applicant submits an application to an existing opportunity in grants.gov (i.e. through grants.gov UI) and then attempts to retrieve it through Simpler SOAP Applicants API.

**Authentication**
[...]

PR #4624, opened 2025-04-17
[Issue #4478] Modify our competition table to include extra columns

### LINKED ISSUE #4478: Competition - Modify our competition table to include extra columns we'll want to transform from the old system
### Summary

Add a new table called `lk_form_family` - this should follow our usual convention of setting up a lookup table and have the following in its enum:
* `sf-424`
* `sf-424-individual`
* `r&r`
* `sf-424-mandatory`
* `sf-424-short-organization`
For context, these are different "groups" that forms can be in: https://grants.gov/forms/forms-repository

Add a lookup table called `lk_competition_open_to_applicant` which has the following values in its enum:
* `individual`
* `organization`

Add a one-to-many table called `link_competition_open_to_applicant` that handles the one-to-many behavior of a competition having multiple applicant types with the following columns (both primary key):
* `competition_id`
* `competition_open_to_applicant_id` (foreign key to `lk_competition_open_to_applicant`)

Add the following columns:
* `form_family_id` -> this should be a lookup foreign key to the lk_form_family, the actual type of the column should be the enum created, and the field should be called `form_family`
* `opportunity_assistance_listing_id` - int, nullable - foreign key to the opportunity assistance listing table
* `open_to_applicants` -> A list relationship to the `link_competition_open_to_applicant` table. We'll want an [association proxy](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/db/models/opportunity_models.py#L242) like we have in a few other cases so this list can be the OpenToApplicant enum
* `is_electronic_required` - bool, nullable
* `expected_application_count` - int, nullable
* `expected_application_size_mb` - Big Integer, nullable
* `is_multi_package` - bool, nullable
* `agency_download_url` - str, nullable
* `is_legacy_workspace_compatible` - bool, nullable (not null in the old system, but we only need it for backwards compatibility, we won't have workspaces)
* `can_send_mail` - bool, nullable

### Acceptance criteria

- [ ] Tables updated

mikehgrantsgov:
## Summary
Fixes #4478 

## Changes proposed
Add lookup and link tables
Add new columns to `Competition` model
Add migration

## Context for reviewers
This supports the data integration epic for Simpler SOAP: https://github.com/HHS/simpler-grants-gov/issues/4440


api/src/db/models/competition_models.py
```diff
@@ -28,6 +32,28 @@ class Competition(ApiSchemaTable, TimestampMixin):
     grace_period: Mapped[int | None] = mapped_column(BigInteger)
     contact_info: Mapped[str | None]
 
+    form_family: Mapped[FormFamily] = mapped_column(
+        "form_family_id",
+        LookupColumn(LkFormFamily),
+        ForeignKey(LkFormFamily.form_family_id),
+        index=True,
+    )
```
chouinar, line 40: Something we'll have to account for, there is data for a single competition in the staging environment, if we run the migration there it'll fail because this non-nullable field won't be set.

Options:
1. Something in the migration that gives it a default (I don't care what the value is, we won't use this for a while and that data was manually made).
2. We make this nullable?
3. I delete and remake the data after the migration

Can't recall what we've done before for this kind of migration

mikehgrantsgov: Making this nullable seems the least bad. We can always make sure the field is set, then run another migration making the field non nullable?

chouinar (approved):

PR #4627, opened 2025-04-17
[Issue #4056] E2E Tests for Search Results

### LINKED ISSUE #4056: E2E Tests -  confirming search result listing information
### Summary

There are currently parts of the application uncovered by E2E tests. This covers the work for filling in existing gaps, but does not cover continued work to fill gaps as new features are added.

Note that this is work that can be done by open source contributors.

### Acceptance criteria

E2E tests written and passing in CI for

- [ ] confirming search result listing information

austincanada:
## Summary
Fixes #4056 

## Changes proposed
Test id was added to SearchResultsList
e2e tests were added for Search Page

## Context for reviewers
Running tests for CI


frontend/tests/e2e/search/search-results.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+test.afterEach(async ({ context }) => {
+  await context.close();
+});
+
+test.describe("Search page results tests", () => {
+  test("should return at least 1 result when searching with valid term", async ({
+    page,
+  }) => {
+    // eslint-disable-next-line testing-library/prefer-screen-queries
+    const resultsHeading = page.getByRole("heading", {
```
doug-s-nava, line 34: this is a weird one - the eslint rule is coming from testing-library, but we're using playwright here, and the page object seems to just have methods like "getByRole" that are technically unrelated to testing-library but have the same names and functionality. In any case, I think we can probably work around this by using `page.locator` instead if we want to avoid the linting error

frontend/src/components/search/SearchResultsList.tsx
```diff
@@ -40,7 +40,7 @@ export default async function SearchResultsList({
   }
 
   return (
-    <ul className="usa-list--unstyled">
+    <ul className="usa-list--unstyled" data-testid="search-list">
```
doug-s-nava, line 43: is there a way to query for this without adding a testid? Sometimes with testing-library the limitations they put on their queries make test ids necessary but since playwright lets you do arbitrary DOM queries I feel like it shouldn't be necessary to add a test id here.

btabaska (approved):
LGTM!

PR #4630, opened 2025-04-17
[no ticket] Fix query to first or none

babebe:
Fix query to use first() to pick up the latest version if none returns None.

### DIFF
--- a/api/src/services/opportunities_v1/opportunity_version.py
+++ b/api/src/services/opportunities_v1/opportunity_version.py
@@ -30,12 +30,16 @@ def save_opportunity_version(db_session: db.Session, opportunity: Opportunity) -
         return False
 
     # Fetch latest opportunity version stored
-    latest_opp_version = db_session.execute(
-        select(OpportunityVersion)
-        .where(OpportunityVersion.opportunity_id == opportunity.opportunity_id)
-        .order_by(OpportunityVersion.created_at.desc())
-        .options(selectinload("*"))
-    ).first()
+    latest_opp_version = (
+        db_session.execute(
+            select(OpportunityVersion)
+            .where(OpportunityVersion.opportunity_id == opportunity.opportunity_id)
+            .order_by(OpportunityVersion.created_at.desc())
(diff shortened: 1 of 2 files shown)


api/src/services/opportunities_v1/opportunity_version.py
```diff
@@ -35,7 +35,7 @@ def save_opportunity_version(db_session: db.Session, opportunity: Opportunity) -
         .where(OpportunityVersion.opportunity_id == opportunity.opportunity_id)
         .order_by(OpportunityVersion.created_at.desc())
         .options(selectinload("*"))
-    ).scalar_one_or_none()
+    ).first()
```
chouinar, line 38: Could you add a test with a few versions (more than 1 previous)?

babebe: yup adding one now

api/src/services/opportunities_v1/opportunity_version.py
```diff
@@ -35,15 +35,15 @@ def save_opportunity_version(db_session: db.Session, opportunity: Opportunity) -
...
 
     # Extracts the opportunity data as JSON object
     opportunity_new = SCHEMA.dump(opportunity)
 
     diffs = []
 
     if latest_opp_version:
-        diffs = diff_nested_dicts(opportunity_new, latest_opp_version.opportunity_data)
+        diffs = diff_nested_dicts(opportunity_new, latest_opp_version[0].opportunity_data)
```
chouinar, line 46: first should only return one? I think with the execute/select approach we instead want that query to be:
```
db_session.scalars(select(...)).first()
```

babebe: this does return one but it returns a row and not the entity so have to index to grab it

chouinar: `scalars` will remove the need to do the indexing, you'll get exactly one like you expect

chouinar (approved):

Issue #4612, opened 2025-04-16
"Community" is live in the global nav with sub-items


btabaska:
'Community' should link to:
- Events (/events)
- Subscribe (/subscribe)
- Public Wiki (https://wiki.simpler.grants.gov/)
- Discussion forum (https://simplergrants.discourse.group/)

- [ ] "Community" is live in the global nav with sub-items

karinamzalez:
- Adding second navbar dropdown, Community, requires a refactor of the Header component. The header component currently can only handle one navbar dropdown at a time.

karinamzalez:
[* Adding community drop](https://www.figma.com/design/FGxWtAgToKhehLJCiuy1zL/Simpler.Grants.gov-Project?node-id=3841-2885&t=wDLHdKZUQOo9rd2i-0)

PR #4606, opened 2025-04-16
[Issue #4454] auto close nav sub menus

doug-s-nava:
## Summary
Fixes #4454

## Changes proposed
The workspace and user dropdowns in the nav are currently not closing unless you specifically click on them to close them. With this change they will close on any click after they are opened.

### Test steps

1. start a server on this branch with `npm run dev`
2. visit http://localhost:3000?_ff=authOn:true;savedSearchesOn:true
3. log in
4. click the workspace nav
5. click anywhere (you should probably test clicking a bunch of different places)
6. _VERIFY_: subnav closes
7. repeat 4 - 6 for the user dropdown nav

Issue #4454, opened 2025-04-03
Header nav items (workspace & login) do not collapse when focus changes


chrisw-nava:
### Summary

When using the workspace or login header navs, the nav remains open until you click it closed instead of closing automatically when another button is clicked or focus moves away from it by clicking elsewhere on the screen

### Reproducibility

I tried multiple times and saw the bug every time

### How to reproduce the bug

Flow 1: 
- Click "Workspace"
- Click "Saved queries"
- Arrive on saved queries screen
- See that the nav remains open

Flow 2:
- Click "Workspace"
- Click anywhere else outside of the "Workspace" nav item
- Nav remains open when it should close

### Browser

Chrome

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

andycochran:
If we're using the Truss Header component, per https://github.com/trussworks/react-uswds/issues/1930, they explored 3 options: 

> 1. Leave it alone. Let consumers of react-uswds set their own open/close togglers.
> 2. Use the [uswds-made listener](https://github.com/uswds/uswds/blob/v2.14.0/src/js/components/navigation.js)
> 3. Write our own listener, trying to mirror the behavior of the listener above.

And settles on #1:  "…leaving it to consumers to write their own togglers"

Issue #3942, opened 2025-02-19
Create new 'About' element in the global nav


mxk0:
### Summary

'About' should link to:
- Our vision (/vision)
- Product roadmap (/roadmap)

### Acceptance criteria

- [ ] "About" is live in the global nav with sub-items
- [ ] "Roadmap" element is removed from the global nav since it now appears under "About"

karinamzalez:
Picking this up!

karinamzalez:
https://www.figma.com/design/FGxWtAgToKhehLJCiuy1zL/Simpler.Grants.gov-Project?node-id=3841-2885&t=wDLHdKZUQOo9rd2i-0

karinamzalez:
Link to PR for About menu item: https://github.com/HHS/simpler-grants-gov/pull/4617/

PR #4639, opened 2025-04-18
[Issue #3942], [Issue #4612], [Issue #4789], and [#Issue 4799] - Add "About" and "Community" header dropdowns

karinamzalez:
## Summary
Fixes [#{3942}](https://github.com/HHS/simpler-grants-gov/issues/3942), [#{4612}](https://github.com/HHS/simpler-grants-gov/issues/4612),  [#4789](https://github.com/HHS/simpler-grants-gov/issues/4789), and [#4799](https://github.com/HHS/simpler-grants-gov/issues/4799)

### Time to review: __30 mins__

#3942: 
- Adds "About" to navbar menu 
    - Adds "About" submenu that includes the following submenu items: 
        - "Our vision" --> links to /vision
        - "Product roadmap" ---> links to /roadmap
 - removes "Roadmap" navbar button 
 - Adds tests 

#4612 
- Adds "Community" to navbar menu 
   - Adds "Community" submenu that includes the following submenu items: 
        - "Subscribe" --> links to /subscribe
        - "Events" ---> links to /events
        - "Public wiki" --> links to "https://wiki.simpler.grants.gov/"
        - "Discussion forum" --> links to "https://simplergrants.discourse.group/"
     - _external_ link submenu items: 
        - render USWDS' "launch" icon 
        - launch new tab to open link 

#4789 
- Refactors Nav Dropdowns, reference context section for more info. 

#4799
- related to this recently merged ticket: https://github.com/HHS/simpler-grants-gov/pull/4606
   - Issue #4454 was implemented when Header could only handle rendering one dropdown at a time. The solution was interwoven with the header code that this PR deprecates  
   -  This PR now refactors that solution to take into account multiple header dropdowns. Please reference this section of PR for solution specific to this ticket: https://github.com/HHS/simpler-grants-gov/blob/878bd0ebe1b23eb2f7b5261e67c37d6abcda7378/frontend/src/components/NavDropdown.tsx#L36-L68

NON TICKET RELATED: 
-  adds `isExternalLink` function to general utils
- add IndexType to shared types

- ADDS NEW NAVDROPDOWN COMPONENT: 
     - addition of this component allows for proper state management of nav dropdowns so that: 
         - only one navbar dropdown renders at a time 
         - double clicking one dropdown toggles it open and closed 
         - clicking a navbar link _without_ a dropdown/submenu closes any open submenus
         - clicking a submenu item, closes submenu 
      - "activeNavDropdownIndex" state is set in NavDropdowns parent  component (Header) because: 
         - we need access to this to allow for "non-dropdown" navbar items to close any open submenus 
         -  "activeNavDropdownIndex" needs to be added as a dependency on "useMemo" wrapped around `navItem's` in order for NavDropdown's `useEffect` hook to retrigger correctly. 

Demo of both dropdowns:

https://github.com/user-attachments/assets/10fd4258-955c-40cd-9815-1d96f83a53bd

Demo of external links opening a new tab: 

https://github.com/user-attachments/assets/0060f2e0-a28e-4d33-83e6-95492a723c5d

Demo of nav dropdown outside click auto-close (related to #4799) 

https://github.com/user-attachments/assets/7f8e23d7-8d5c-4bec-9cad-62086b7ddc68


frontend/tests/components/search/SaveSearchSelector.test.tsx
```diff
@@ -176,7 +174,7 @@ describe("SaveSearchSelector", () => {
       />,
     );
 
-    expect(clientFetchMock).toHaveBeenCalledTimes(3);
```
karinamzalez, line 179: @doug-s-nava I had to update "SaveSearchSelector" to address linting warnings regarding missing dependencies on "useCallback" instances. I'm not sure why adding those dependencies would affect the amount of times these mock fetches are called.

frontend/src/components/NavDropdown.tsx
```diff
@@ -0,0 +1,57 @@
+"use client";
```
karinamzalez, line 1: Note: when Header component is broken down into smaller components, this file could be moved under a the new "header" directory will all header related components.

doug-s-nava (commented):
a couple preliminary thoughts, I'll get to the rest on Monday

frontend/src/components/Header.tsx
```diff
@@ -152,6 +184,11 @@ const NavLinks = ({
     }
   }, [mobileExpanded, onToggleMobileNav]);
 
+  const closeDropdownAndMobileNav = useCallback(() => {
```
doug-s-nava, line 187: I'm not sure this is necessary - the mobile nav will only be relevant at mobile viewport widths, and the dropdowns will only be relevant at > mobile widths. Am I missing a case where these could be relevant at the same time? I was thinking this may be in place to close the dropdowns if the window is resized but that doesn't seem to be happening.

frontend/src/components/NavDropdown.tsx
```diff
@@ -0,0 +1,57 @@
...
+  setActiveNavDropdownIndex: Dispatch<SetStateAction<IndexType>>;
+}
+
+export default function NavDropdown({
+  activeNavDropdownIndex,
+  index,
+  isCurrent,
+  linkText,
+  menuItems,
+  setActiveNavDropdownIndex,
```
doug-s-nava, line 24: [just some info] passing this in shows a TS error in my VS code, though it doesn't seem to actually cause any problems in the TS compilation:
```
Props must be serializable for components in the "use client" entry file. "setActiveNavDropdownIndex" is a function that's not a Server Action. Rename "setActiveNavDropdownIndex" either to "action" or have its name end with "Action" e.g. "setActiveNavDropdownIndexAction" to indicate it is a Server Action.ts(71007)
```

It's complicated but it goes away if you remove the "use client" from the file. The component will still be a client component, but technically "use client" is only necessary when marking a component that will be included by a server component - since this one is being included by another client component it's a client component already so the "use client" isn't necessary. So, yeah, if you remove the "use client" the tooling will understand that this isn't an entrypoint client component and will be ok with you passing in a non-server-action function as a prop, but no need to update now

frontend/src/components/NavDropdown.tsx
```diff
@@ -0,0 +1,57 @@
...
+}: NavDropdownProps): JSX.Element {
+  const [isOpen, setIsOpen] = useState<boolean>(false);
+
+  function handleToggle() {
+    const activeIndex: IndexType = isOpen ? null : index;
```
doug-s-nava, line 29: I think we can manage this without the local state. 
```suggestion
    const newIndex = activeNavDropdownIndex === index ? null : index;
    setActiveNavDropdownIndex(newIndex);
```

would also need to update line 52 accordingly

frontend/src/components/workspace/DeleteSavedSearchModal.tsx
```diff
@@ -96,7 +96,7 @@ export function DeleteSavedSearchModal({
       .finally(() => {
         setLoading(false);
       });
-  }, [user?.token, replaceQueryParams, savedSearchId]);
+  }, [clientFetch, user?.token, replaceQueryParams, savedSearchId]);
```
doug-s-nava, line 99: I had some trouble with infinite re-renders on some of these, which is why I didn't include the function in the dependency arrays to begin with. Are you sure we're not getting any infinite loops on these? Sometimes hard to tell other than slow performance unless you add a log for each render. Maybe something to consider splitting into a separate pr

frontend/src/hooks/useClientFetch.ts
```diff
@@ -18,25 +18,25 @@ export const useClientFetch = <T>(
...
+          router.refresh();
+        }
+        return response;
       }
       return response;
-    }
-    return response;
-  };
+    },
+    [authGatedRequest, refreshIfExpired, refreshUser, router],
```
doug-s-nava, line 38: why did these need to get updated?

doug-s-nava (commented):
a few more thoughts, and one thing to look into generally. Since the user drop down isn't accounted for in the index state (see `UserControl`) it is still staying open when another nav is open. We'd need to refactor that a bit to get it to work the same way (drill the state and state function down into `UserDropdown`), but probably worth doing

frontend/src/components/Header.tsx
karinamzalez: @doug-s-nava added this because both instances of `NavLink` in this file had `closeMobileNav` passed as the `onClick` prop and I needed to also pass `setActiveNavDropdown(null)` as the onClick action for both to close any open `NavDropdown`s when a non-navdropdown nav item is clicked. This function makes is so that I can cleanly pass `closeDropdownAndMobileNav` as both NavLink's on click actions, as opposed to: 
```
onClick = {() => {
            closeMobileNav;
            setActiveNavDropdownIndex(null);
          }}
```

But I could remove this function and opt for the above instead if that is your preference!

frontend/src/components/NavDropdown.tsx
karinamzalez: @doug-s-nava thanks so much! I was wondering about this, but for some reason the CI's linter did not complain about this so I dropped it 😅

frontend/src/hooks/useClientFetch.ts
karinamzalez: @doug-s-nava to address warnings that were being thrown by the linter

frontend/src/components/Header.tsx
```diff
@@ -37,6 +38,10 @@ type Props = {
 
 const homeRegexp = /^\/(?:e[ns])?$/;
 
+function isExternalLink(href: string): boolean {
```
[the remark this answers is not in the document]
karinamzalez: @doug-s-nava Actually, I think I've just answered my own question haha! How do you feel about this instead
```
  return !!(href && !!href.includes("http"));
```

frontend/src/components/Header.tsx
doug-s-nava: Gotcha, thanks for the explanation. I'm good with this! It may not be necessary after the rebase to pull in my PR to auto close the sub nav, but we can revisit after you've pulled that in

frontend/src/components/Header.tsx
```diff
@@ -126,7 +162,7 @@ const NavLinks = ({
           }
           // mark as current if any child page is active
           return children.some((child) => {
-            return child?.href && isCurrentPath(child.href, currentPath);
+            return child?.isCurrentPath(child.href, currentPath);
```
doug-s-nava, line 165: this seems like a typo, and throws an error

frontend/src/components/workspace/DeleteSavedSearchModal.tsx
karinamzalez: Of course!

frontend/src/components/Header.tsx
karinamzalez: it definitely is haha

frontend/src/components/Header.tsx
```diff
@@ -167,52 +199,31 @@ const NavLinks = ({
...
+          />
         );
       }
       return (
         <NavLink
           href={link.href}
           key={link.href}
-          onClick={closeMobileNav}
+          onClick={() => {
+            closeMobileNav;
```
doug-s-nava, line 224: did you mean to call this?

frontend/src/hooks/useClientFetch.ts
doug-s-nava: in order to limit the scope on this, if you could remove the useCallback for now that'd be best

frontend/src/components/NavDropdown.tsx
```diff
@@ -0,0 +1,80 @@
...
+  let isOpen = activeNavDropdownIndex === index;
+
+  function handleToggle() {
+    const activeIndex: IndexType = isOpen ? null : index;
+    setActiveNavDropdownIndex(activeIndex);
+  }
+
+  useEffect(() => {
+    isOpen = activeNavDropdownIndex === index;
+  }, [activeNavDropdownIndex, index]);
```
doug-s-nava, line 43: when we were discussing I had a feeling I was forgetting something, and remembered later - rather than the defining the variable and changing it in the useEffect, you could just assign the variable in a useMemo. This is good for now, but something for later maybe.

frontend/src/styles/_uswds-theme-custom-styles.scss
```diff
@@ -479,3 +479,9 @@ button.usa-pagination__button.usa-button {
     }
   }
 }
+
+.icon-btn {
+  display: flex;
```
doug-s-nava, line 484: a couple of thoughts here if you have time for css:
* I think that you could accomplish all of this by wrapping the link in an unstyled button. If you do a find for "unstyled" in the project you should be able to find other examples of where we've done this. 
* In case that doesn't work, I think you can at least move the `display: flex` to a `display-flex` class and remove it from this custom css. `align-items`, probably not, I'm not sure about `gap`, but that could maybe be handled with padding or margin instead?

frontend/src/components/Header.tsx
karinamzalez: No, this was from the merge. It has been reverted back.

frontend/src/components/NavDropdown.tsx
karinamzalez: @doug-s-nava we should discuss in morning!

frontend/src/components/NavDropdown.tsx
```diff
@@ -31,37 +22,61 @@ export default function NavDropdown({
   menuItems,
   setActiveNavDropdownIndex,
 }: NavDropdownProps): JSX.Element {
-  let isOpen = activeNavDropdownIndex === index;
+  const [isOpen, setIsOpen] = useState<boolean>(false);
+
+  function eventHandler(e: any) {
+    const dropdowns = document.getElementsByName("navDropDownButton");
```
doug-s-nava, line 35: Doing a dom query here feels a little icky. I wonder if we could get around having to do a query like this by using refs, or even running this function at a higher level that knows about all of the dropdowns, and passing it through.

frontend/src/components/NavDropdown.tsx
```diff
@@ -31,37 +22,61 @@ export default function NavDropdown({
...
 
   useEffect(() => {
-    isOpen = activeNavDropdownIndex === index;
+    setIsOpen(activeNavDropdownIndex === index);
   }, [activeNavDropdownIndex, index]);
 
   return (
     <>
       <NavDropDownButton
+        id={index.toString()}
```
doug-s-nava, line 83: I see what you're doing, and it's not a bad solution, but since ids need to be unique in the document I'm not a big fan of using just a blank number for an id. Can we set this as `nav-dropdown-button-${index}` and then parse the index out of that in the handler?

frontend/src/components/NavDropdown.tsx
```diff
@@ -0,0 +1,107 @@
...
+  activeNavDropdownIndex,
+  index,
+  isCurrent,
+  linkText,
+  menuItems,
+  setActiveNavDropdownIndex,
+}: NavDropdownProps): JSX.Element {
+  const [isOpen, setIsOpen] = useState<boolean>(false);
+
+  function eventHandler(e: MouseEvent) {
```
doug-s-nava, line 34: this is nice solution, and seems to work perfectly. I couldn't help trying to clean it up just a little tho

```suggestion
  // closes all navs unless the click event falls on a currently closed drop down
  // in which case, set that drop down as active
  function closeNavs(e: MouseEvent) {
    let targetId = null;
    const dropdowns = document.getElementsByName("navDropDownButton");
    for (const dropdown of dropdowns) {
      if (dropdown.contains(e.target as Node)) {
        let targetNode: HTMLElement = e.target as HTMLElement;
        if (targetNode.localName === "span") {
          targetNode = targetNode.parentNode as HTMLElement;
        }
        targetId = toNumber(targetNode.id);
        break;
      }
    }
    setActiveNavDropdownIndex(targetId);
```

frontend/src/components/NavDropdown.tsx
```diff
@@ -0,0 +1,107 @@
...
+    <>
+      <NavDropDownButton
+        id={index.toString()}
+        name="navDropDownButton"
+        label={linkText}
+        menuId={linkText}
+        isOpen={isOpen}
+        // eslint-disable-next-line
+        // @ts-ignore
+        onClick={(e: MouseEvent<HTMLButtonElement, MouseEvent>) =>
```
doug-s-nava, line 90: I was able to work around the TS problems here
* no type assignments on the click handler
* handleToggle takes a MouseEvent
* closeNavs takes an Event

frontend/src/components/NavDropdown.tsx
karinamzalez: @doug-s-nava committing this suggestion, would you still like me to try removing the DOM query? Since this is checking to see if the dropdown was clicked via the mouse event, I'm unsure how I'd do it at a higher level.

frontend/src/components/NavDropdown.tsx
karinamzalez: @doug-s-nava I agree that a DOM query isn't ideal. An alternative to asserting that the mouseevent didn't click a dropdown wasn't immediately obvious to me. After checking in via slack, we've decided to create a backlog issue for this. Sharing here for visibility: https://github.com/HHS/simpler-grants-gov/issues/4797

frontend/src/hooks/useClientFetch.ts
```diff
@@ -55,7 +55,7 @@ export const useClientFetch = <T>(
         throw new Error(`${errorMessage}: ${response.status}`);
       }
     },
-    [],
+    [errorMessage, fetchWithAuthCheck, jsonResponse],
```
karinamzalez, line 58:
```suggestion
    [],
```

frontend/src/components/search/ExportSearchResultsButton.tsx
```diff
@@ -31,7 +31,7 @@ export function ExportSearchResultsButton() {
         );
       })
       .catch(console.error);
-  }, [searchParams]);
+  }, [clientFetch, searchParams]);
```
karinamzalez, line 34:
```suggestion
  }, [searchParams]);
```

frontend/tests/components/search/SaveSearchSelector.test.tsx
```diff
@@ -155,7 +155,7 @@ describe("SaveSearchSelector", () => {
       />,
     );
 
-    expect(clientFetchMock).toHaveBeenCalledTimes(1);
+    expect(clientFetchMock).toHaveBeenCalledTimes(2);
```
karinamzalez, line 158:
```suggestion
    expect(clientFetchMock).toHaveBeenCalledTimes(1);
```

frontend/tests/components/search/SaveSearchSelector.test.tsx
```diff
@@ -166,7 +166,7 @@ describe("SaveSearchSelector", () => {
       />,
     );
 
-    expect(clientFetchMock).toHaveBeenCalledTimes(2);
+    expect(clientFetchMock).toHaveBeenCalledTimes(3);
```
karinamzalez, line 169:
```suggestion
    expect(clientFetchMock).toHaveBeenCalledTimes(2);
```

frontend/tests/components/search/SaveSearchSelector.test.tsx
```diff
@@ -176,7 +176,7 @@ describe("SaveSearchSelector", () => {
       />,
     );
 
-    expect(clientFetchMock).toHaveBeenCalledTimes(3);
+    expect(clientFetchMock).toHaveBeenCalledTimes(4);
```
karinamzalez, line 179:
```suggestion
    expect(clientFetchMock).toHaveBeenCalledTimes(3);
```

frontend/src/components/NavDropdown.tsx
karinamzalez: done!

frontend/src/styles/_uswds-theme-custom-styles.scss
karinamzalez: @doug-s-nava because the rendering of the icon on this button is inconsistent and this button utilizes nextJS's Link, I don't think an unstyled button could be used in this case. Unless I'm missing something? Here's the link to the code in reference. 

https://github.com/HHS/simpler-grants-gov/blob/3f926bcfd691e0e8f1906b31f7dd2fca63325484/frontend/src/components/Header.tsx#L60

doug-s-nava (dismissed):
✅ ✅ ✅ ✅ 

this work was a bit of a beast, thanks for rolling with it!

frontend/tests/components/Header.test.tsx
```diff
@@ -185,4 +189,55 @@ describe("Header", () => {
...
+      ).not.toBeInTheDocument();
+
+      const aboutBtn = screen.getByRole("button", { name: /Community/i });
+
+      await userEvent.click(aboutBtn);
+
+      expect(container).toMatchSnapshot();
+      expect(aboutBtn).toHaveAttribute("aria-expanded", "true");
+
+      const visionLink = screen.getByRole("link", { name: /Events/i });
```
karinamzalez, line 239:
```suggestion
      const eventsLink = screen.getByRole("link", { name: /Events/i });
```

frontend/tests/components/Header.test.tsx
```diff
@@ -185,4 +189,55 @@ describe("Header", () => {
...
+      expect(visionLink).toBeInTheDocument();
+    });
+    it("renders Community submenu", async () => {
+      const { container } = render(<Header />);
+
+      expect(
+        screen.queryByRole("link", { name: /Events/i }),
+      ).not.toBeInTheDocument();
+
+      const aboutBtn = screen.getByRole("button", { name: /Community/i });
```
karinamzalez, line 232:
```suggestion
      const communityBtn = screen.getByRole("button", { name: /Community/i });
```

frontend/tests/components/Header.test.tsx
```diff
@@ -185,4 +189,55 @@ describe("Header", () => {
...
+    it("renders Community submenu", async () => {
+      const { container } = render(<Header />);
+
+      expect(
+        screen.queryByRole("link", { name: /Events/i }),
+      ).not.toBeInTheDocument();
+
+      const communityBtn = screen.getByRole("button", { name: /Community/i });
+
+      await userEvent.click(aboutBtn);
```
karinamzalez, line 234:
```suggestion
      await userEvent.click(communityBtn);
```

PR #4647, opened 2025-04-18
add some docs around metadata and camelCasing

doug-s-nava:
## Summary
Unticketed

## Changes proposed
Add some code style docs

### DIFF
--- a/documentation/frontend/code-style.md
+++ b/documentation/frontend/code-style.md
@@ -36,6 +36,8 @@ The main things to focus on are:
 - don't abuse (or maybe even use) abbreviations
 - think about searchability
 
+Also, `camelCase` should be used by default. Objects coming in from outside sources, such as API response payloads, may use `snake_case`, and there is no need to avoid using snake case in those instances, but any newly defined variables within the NextJS app should use camel case.
+
 ### Promises
 
 Constants that represent unresolved promises should be named `varNamePromise(s)`.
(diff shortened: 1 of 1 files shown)


documentation/frontend/code-style.md
```diff
@@ -36,6 +36,8 @@ The main things to focus on are:
 - don't abuse (or maybe even use) abbreviations
 - think about searchability
 
+Also, `camelCase` should be used by default. Objects coming in from outside sources, such as API response payloads, may use `snake_case`, and there is no need to avoid using snake case in those instances, but any newly defined variables within the NextJS app should use camel case.
```
acouch, line 39:
```suggestion
Also, `camelCase` should be used by default. Objects coming in from outside sources, such as API response payloads, may use `snake_case`, and there is no need to avoid using snake case in those instances, but any newly defined variables within the Next.js app should use camel case.
```

extreme nit, "Next.js" is how next describes itself https://nextjs.org/

documentation/frontend/code-style.md
```diff
@@ -60,3 +62,9 @@ Classes for different flavors of API errors are established in errors.ts, use th
 ### Snapshots & Page Component Tests
 
 We have observed that running typical Jest unit tests on dynamic Next JS page level components is complicated and generally does not provide much value. When creating any new pages, feel free to rely on snapshot based tests rather than functional tests for these components.
```
acouch, line 64:
```suggestion
We have observed that running typical Jest unit tests on dynamic Next.js page level components is complicated and generally does not provide much value. When creating any new pages, feel free to rely on snapshot based tests rather than functional tests for these components.
```

Out of scope extreme nit, Next.js spelling / case

acouch (approved):
Looks great, two spelling nits

acouch (approved):

PR #4694, opened 2025-04-21
[Issue #4479] Add competition transformer

### LINKED ISSUE #4479: Competition - Create a transform job for the competition table
### Summary
After https://github.com/HHS/simpler-grants-gov/issues/4478 has been completed

We want to transform the TCompetition staging table into our Competition table format. This _must_ follow the existing patterns for how we process transforming other tables. The exact way in which we process these tables needs to be followed exactly including:
* Fetching all of the relevant table rows up front in the query (performance + SQLAlchemy caching reasons)
* Structure of functions, order of operations (process deletes the same, including allowing for orphaned records)
* Inserts/updates must work _exactly_ the same, and updates should call `db_session.merge` NOT `db_session.add` (SQLAlchemy caching necessitates this behavior, updates can get lost or done incorrectly otherwise). This also means for updates we don't modify the value we fetched from the DB, but make a new one and merge it.
* Error handling structured the same, any transform issues should raise a ValueError which is caught, logged, and the next record set is processed.

One major thing to call out is how a competition connects to an opportunity in the old system. These always needed to go via an `opportunity_cfda` (now opportunity assistance listing). We don't import all opportunity assistance listings into our system because many of them are empty/null. Competitions are the reason they're null because sometimes folks wanted a competition on an opportunity, but didn't want an actual assistance listing. To account for this, we'll need to join with the `opportunity_cfda` record so we know the opportunity ID. This will require some adjustments to the logic to fetch the data upfront most likely. This is likely the most complex part of this beyond "following the general pattern".

#### Field Mapping

If a field mapping doesn't provide any further details, assume it can be copied as is without modification. Some fields require additional processing to fit into our schema and are called out.

* `comp_id` -> `legacy_competition_id`
* n/a -> `opportunity_id`- see above
* `opp_cfda_id` -> `opportunity_assistance_listing_id` - see above, should point to the assistance listing record IF we have it / it wasn't the weird null case.
* `competitionid` -> `public_competition_id`
* `package_id` -> `legacy_package_id`
* `competitiontitle` -> `competition_title`
* `openingdate` -> `opening_date`
* `closingdate` -> `closing_date`
* `graceperiod` -> `grace_period`
* `contactinfo` -> `contact_info`
* `familyid` -> `form_family_id` - See below for mapping to our enum
* `opentoapplicanttype` -> `open_to_applicants` - See below for mapping to our one-to-many enum
* `electronic_required` -> `is_electronic_required` - Values can be `Y` or `N`, use our existing `convert_yn_bool` function to convert.
* `expected_appl_num` -> `expected_application_count`
* `expected_appl_size` -> `expected_application_size_mb`
* `ismulti` -> `is_multi_package` - Values can be `Y` or `N`, use our existing `convert_yn_bool` function to convert.
* `agency_dwnld_url` -> `agency_download_url`
* `is_wrkspc_compatible` -> `is_legacy_workspace_compatible` - Values can be `Y` or `N`, use our existing `convert_yn_bool` function to convert.
* `sendmail` -> `can_send_mail` - Values can be `Y` or `N`, use our existing `convert_yn_bool` function to convert.
* `created_date`/`last_upd_date` -> `created_at`/`updated_at` - Use the `transform_update_create_timestamp` util that already exists, it handles the complexity in copying this value over.

Family ID in the Oracle DB is an integer that we should map to our enum value for FormFamily as follows:
* 12 -> SF-424 Individual
* 14 -> R&R
* 15 -> SF-424
* 16 -> SF-424 Mandatory
* 17 -> SF-424 Short Organization
* null -> null
Any other values should raise a ValueError

`opentoapplicanttype` is a single value in the old system, but we're changing it to a one-to-many lookup value as follows:
* 1 -> `organization`
* 2 -> `individual`
* 3 -> `organization` AND `individual` - we should insert two records for this case
* null -> Add neither

### Acceptance criteria

- [ ] Created the job following the same structure as our other transform jobs
- [ ] Thorough tests added for all common cases and error cases (see other transform tests)

mikehgrantsgov:
## Summary
Fixes #4479

## Changes proposed
Add competition transformer
Create relevant tests for various transformer scenarios
Update factories and conftest

## Context for reviewers
Important note is how a competition connects to an opportunity in the old system. For various reasons in the ticket, we join the opportunity_cfda record so we know the opportunity ID.  Otherwise the general transformer pattern is implemented here.

## Additional information
See unit tests for scenarios. Can be broken into smaller PRs if review proves difficult.


api/src/data_migration/transformation/subtask/transform_competition.py
```diff
@@ -0,0 +1,236 @@
...
+            ).scalar_one_or_none()
+
+            if opportunity_cfda:
+                opportunity_id = opportunity_cfda.opportunity_id
+                opportunity_assistance_listing_id = opportunity_cfda.opp_cfda_id
+
+                # Make sure the opportunity exists in our target table
+                opportunity = self.db_session.execute(
+                    select(Opportunity).where(Opportunity.opportunity_id == opportunity_id)
+                ).scalar_one_or_none()
```
chouinar, line 87: Could we move the fetching of these to the function that calls this? That way we can keep the process function just focused on processing the already fetched values / follow the rough pattern of the other transform jobs. It's fine if when they get passed into this method they're null, we can handle the cases for that

api/src/data_migration/transformation/subtask/transform_competition.py
```diff
@@ -0,0 +1,236 @@
...
+            source_competition.transformed_at = self.transform_time
+            return
+
+        # Find the related opportunity and opportunity_assistance_listing
+        opportunity_id = None
+        opportunity_assistance_listing_id = None
+
+        # One major thing to call out is how a competition connects to an opportunity in the old system.
+        # These always needed to go via opportunity_cfda (now opportunity assistance listing).
+        if source_competition.opp_cfda_id is not None:
```
chouinar, line 72: Could we invert this and raise an error if it happens? While the column is nullable, at least in prod there aren't any cases of it actually being null, so it's definitely an edge case where we should be made aware of it.


```suggestion
        # One major thing to call out is how a competition connects to an opportunity in the old system.
        # These always needed to go via opportunity_cfda (now opportunity assistance listing).
        if source_competition.opp_cfda_id is None:
             raise ValueError("Some sort of message")
        ...
```

api/src/data_migration/transformation/subtask/transform_competition.py
```diff
@@ -0,0 +1,236 @@
...
+        # To avoid incrementing metrics for records we fail to transform, record
+        # here whether it's an insert/update and we'll increment after transforming
+        is_insert = target_competition is None
+
+        if opportunity_id is None:
+            logger.warning(
+                "Competition references opportunity that doesn't exist",
+                extra={**extra, "opportunity_id": opportunity_id},
+            )
+            return
```
chouinar, line 106: We should maintain the same structure as the other transform jobs. If the opportunity were null here, that should be an error.

I know it's being very pedantic, but the if/elif/else structure is necessary for a lot of weird eccentricities in how the SQLAlchemy caching/legacy data come together (to be perfectly honest, I don't 100% understand it, I just know the pattern that works).

```
if is_deleted:
   # handle delete
elif opportunity is None:
   raise ValueError(..)
else: # insert/update
   ...

logger.info("Processed ...", extra=extra)
source_competition.transformed_at = self.transform_time
```

api/src/data_migration/transformation/subtask/transform_competition.py
```diff
@@ -0,0 +1,236 @@
...
+    target_competition.agency_download_url = source_competition.agency_dwnld_url
+
+    # is_wrkspc_compatible is a required field in the source table but optional in the target
+    # This value should always exist in the source table
+    target_competition.is_legacy_workspace_compatible = transform_util.convert_yn_bool(
+        source_competition.is_wrkspc_compatible
+    )
+
+    target_competition.can_send_mail = transform_util.convert_yn_bool(source_competition.sendmail)
+    target_competition.opportunity_assistance_listing_id = opportunity_assistance_listing_id
```
chouinar, line 192: Nit: could we put this one with the other simple copy values around like ~156?

api/src/data_migration/transformation/subtask/transform_competition.py
```diff
@@ -0,0 +1,236 @@
...
+
+    return form_family_map[form_family_id]
+
+
+def transform_open_to_applicants(applicant_type: int | None) -> set[CompetitionOpenToApplicant]:
+    """Transform the legacy open to applicant type to the new set of enum values."""
+    if applicant_type is None:
+        return set()
+
+    print(applicant_type)
```
chouinar, line 226:
```suggestion

```

api/src/data_migration/transformation/transform_oracle_data_task.py
```diff
@@ -49,6 +50,7 @@ class TransformOracleDataTaskConfig(PydanticBaseEnvConfig):
     enable_opportunity_attachment: bool = (
         True  # TRANSFORM_ORACLE_DATA_ENABLE_OPPORTUNITY_ATTACHMENT
     )
+    enable_competition: bool = True  # TRANSFORM_ORACLE_DATA_ENABLE_COMPETITION
```
chouinar, line 53: Let's disable this by default, and manually test it after we merge it. The job will probably take a while to run and we don't want two hourly jobs running on top of each other.

api/src/data_migration/transformation/subtask/transform_competition.py
mikehgrantsgov: Refactored this and updated the test, too.

api/src/data_migration/transformation/subtask/transform_competition.py
mikehgrantsgov: Updated this, too.

api/src/data_migration/transformation/subtask/transform_competition.py
mikehgrantsgov: Moved this up.

api/src/data_migration/transformation/subtask/transform_competition.py
```diff
@@ -0,0 +1,238 @@
...
+                    ).scalar_one_or_none()
+
+                    if not opportunity:
+                        logger.warning(
+                            "Competition references opportunity that doesn't exist in target schema",
+                            extra={
+                                "competition_id": source_competition.comp_id,
+                                "opportunity_id": opportunity_id,
+                            },
+                        )
```
chouinar, line 58: Let us not log a warning here - there is a non-problematic scenario where this can happen - if the opportunity was deleted, and we're processing the competition's delete, then this is actually expected (we would have recursively deleted the competition already - also why the if/elif/else is important in the other function).
```suggestion

```

Actually - we don't currently have a relationship between opportunity and competition (and opportunity should be aware of its competitions) - if we tried to delete the opportunity right now and it had a competition, we'd get a Postgres error - could we add that relationship with the cascade delete stuff?

api/tests/src/data_migration/transformation/conftest.py
```diff
@@ -801,3 +808,243 @@ def validate_opportunity_attachment(
...
+    is_already_processed: bool = False,
+    opportunity: Opportunity | None = None,
+    opportunity_assistance_listing_id: int | None = None,
+    source_values: dict | None = None,
+    all_fields_null: bool = False,
+) -> staging.competition.Tcompetition:
+    """Setup a competition record for testing transformation"""
+
+    # Default values for non-nullable fields
+    values = {
```
chouinar, line 826: There could be fields that for our test we want a certain way but would relying on the factories to generate these with some variance be fine?

mikehgrantsgov: Good call, I removed this to allow some more variability and defer to the factories.

PR #4725, opened 2025-04-22
[Issue #4622] Refactor email notifications

### LINKED ISSUE #4622: Refactor Email notification
### Summary

Refactor NotificationTask such that, we have:
-  Main task that handles sending the email notifications
- 3 Sub Classes that inherit from the Main class for  each email notification types: `update_notification`, `closing_date_notifications`, `search_notifications`

### Acceptance criteria

[ ] Code refactored
[ ] Tests updated/created

babebe:
## Summary
Fixes #4622

### Time to review: __20 mins__

* Created new task for sending email notification
* Segregate the 3 different "sub-tasks" into individual tasks 
* Created tests for each sub-tasks


api/src/task/notifications/BaseNotification.py
```diff
@@ -0,0 +1,51 @@
...
+
+import src.adapters.db as db
+from src.adapters.search import SearchClient
+from src.db.models.user_models import User, UserSavedOpportunity, UserSavedSearch
+from src.task.notifications.constants import EmailData
+
+logger = logging.getLogger(__name__)
+
+
+class BaseNotification(ABC):
```
chouinar, line 16: These classes should all be derived from the Task class, just so it comes with the automatic timing/metric stuff. We should have these be used like the transform job where we just have a config to turn tasks on/off. It's perfectly fine to have an ECS task call multiple "Task" classes (even if that ends up being a tad confusing to follow).

You could even make it so this case class has a few abstract methods as you do here, but its `run_task` method calls each of those in order.

api/src/task/notifications/BaseNotification.py
```diff
@@ -0,0 +1,51 @@
+import logging
+from abc import ABC, abstractmethod
+from typing import Optional
```
chouinar, line 3: Just use `whatever_type | None`
```suggestion

```

api/src/task/notifications/closing_date_notification.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+class ClosingDateNotification(BaseNotification):
+    collected_data: dict | None = None
+
+    def __init__(self, db_session: db.Session):
+        super().__init__(
+            db_session,
+        )
+        self.frontend_base_url = None
```
chouinar, line 32: This should be a config parameter somewhere.

api/src/task/notifications/closing_date_notification.py
```diff
@@ -0,0 +1,171 @@
...
+                "user_count": len(closing_date_opportunities),
+                "total_notifications": sum(
+                    len(closing_date_opportunities)
+                    for container in closing_date_opportunities.values()
+                ),
+            },
+        )
+        self.collected_data = closing_date_opportunities or None
+
+        return closing_date_opportunities or None
```
chouinar, line 81: It generally makes logic a bit easier when working with lists/sets/collections, just return an empty list/set/whatever.

Normally we'll have logic like:

```py
my_values = get_some_list(...)
for value in my_values:
    print(value)
```
If `get_some_list` always returns a list, this works, if we can return None, then we have to additionally add guardrails if it is None. Better to just return an empty list. If an empty list is a sign of a problem, well then we can add a `if len(my_values) == 0`.

Not having anything to iterate over is often fine, keep it simple.

api/src/task/notifications/closing_date_notification.py
```diff
@@ -0,0 +1,171 @@
+import logging
+from datetime import timedelta
+from typing import List
```
chouinar, line 3: Just use `list[...]`
```suggestion

```

api/src/task/notifications/closing_date_notification.py
```diff
@@ -0,0 +1,171 @@
...
+    ) -> EmailData:
+
+        notification: dict[UUID, str] = {}
+        to_address_list: List[str] = []
+
+        for user_id, saved_items in closing_date_data.items():
+            user_email = self._get_user_email(user_id)
+            if not user_email:
+                continue
+            if len(saved_items) == 1:
```
chouinar, line 94: Why this special logic for a single item? Why not just always iterate over the list?

api/src/task/notifications/generate_notifications.py
```diff
@@ -92,357 +53,55 @@ def __init__(
...
-                "user_count": len(self.user_notification_map),
-                "total_searches": sum(
-                    len(container.saved_searches)
-                    for container in self.user_notification_map.values()
-                ),
-            },
-        )
-
-    def _send_notifications(self) -> None:
+    def send_notifications(self, data: EmailData) -> None:
```
chouinar, line 75: Why not make `send_notifications` a function in your base notification class and just have it be called after the other functions? Keep it all contained in one class, although the actual sending is a shared util

api/src/task/notifications/BaseNotification.py
chouinar, line 1: File should be named `base_notification.py`

api/src/task/notifications/closing_date_notification.py
```diff
@@ -0,0 +1,171 @@
...
+CONTACT_INFO = (
+    "mailto:support@grants.gov\n"
+    "1-800-518-4726\n"
+    "24 hours a day, 7 days a week\n"
+    "Closed on federal holidays"
+)
+
+
+class ClosingDateNotification(BaseNotification):
+    collected_data: dict | None = None
```
chouinar, line 26: **Caution with setting values like this** - it's easy to make the mistake because we do this sort of thing with dataclasses/pydantic/SQLAlchemy models - the way those objects are built is fundamentally different and not normal.

If you define a value like this normally, it's not defined on the object, it's defined on the class itself which leads to weird behavior.

Here's an example:
```py
class MyClass:
     x: int | None = None

obj1 = MyClass()
obj1.x = 5
print(obj1.x) # prints 5 like you'd expect

obj2 = MyClass()
print(obj2.x) # prints None like you'd expect

MyClass.x = 4 # change on the class itself
print(ob1.x) # prints 4
print(obj2.x) # also prints 4
```

Effectively, don't set anything that isn't class-scoped like this. I'd recommend just making it something you initialize in an `__init__` function.

api/src/task/notifications/closing_date_notification.py
```diff
@@ -0,0 +1,185 @@
...
+                close_date_stmt = select(OpportunitySummary.close_date).where(
+                    OpportunitySummary.opportunity_id == opportunity.opportunity_id
+                )
+                close_date = self.db_session.execute(close_date_stmt).scalar_one_or_none()
+                if close_date is None:
+                    logger.warning(
+                        "No close date found for opportunity",
+                        extra={"opportunity_id": opportunity.opportunity_id},
+                    )
+                    continue
```
chouinar, line 112: This logic assumes an opportunity has a single opportunity summary which isn't correct, it can have two, and the logic of how that works is an entire task that sets the data up for you already. Don't write any queries, just do:
```py
close_date = opportunity.current_opportunity_summary.opportunity_summary.close_date if opportunity.current_opportunity_summary
```

We have relationships between these, don't write queries.

api/src/task/notifications/base_notification.py
```diff
@@ -0,0 +1,89 @@
...
+        user = self.db_session.execute(
+            select(User).where(User.user_id == user_id)
+        ).scalar_one_or_none()
+
+        if not user or not user.email:
+            logger.warning(
+                "No email found for user", extra={"user_id": user.user_id if user else None}
+            )
+            return None
+        return user.email
```
chouinar, line 50: Since all notifications are driven by tables like `UserSavedOpportunity` or `UserSavedSearch` that are connected to the user, I wonder if there is a clean way of grabbing the user from those records?

That way we can both:
* Fetch the user object when fetching the saved record (better performance)
* Not need this special logic

api/src/task/notifications/base_notification.py
```diff
@@ -0,0 +1,89 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class BaseNotification(Task):
+
+    @abstractmethod
+    def collect_notifications(
+        self,
+    ) -> dict[UUID, list[Any]]:
```
chouinar, line 22: Two ideas to avoid `Any`
* Don't pass the values around, just call the functions, each function has to grab the values at the beginning from something it setup in `__init__`
* Generics (may or may not be better in this case, but useful to be aware of)

If you want to have the typing be a tad better here, we could use [Generics](https://typing.python.org/en/latest/reference/generics.html).

Effectively, we define this class as `class BaseNotification(Task, Generic[T])` and the derived classes are like `ClosingDateNotification(BaseNotification[UserSavedOpportunity])`. You can see a really simple version of this with our Lookup classes: https://github.com/HHS/simpler-grants-gov/blob/main/api/src/db/models/lookup/lookup.py

api/src/task/notifications/base_notification.py
```diff
@@ -0,0 +1,89 @@
...
+                        message=message,
+                        pinpoint_client=pinpoint_client,
+                        app_id=app_id,
+                    )
+                    logger.info(
+                        "Successfully sent notification to user",
+                        extra={
+                            "user_id": user_id,
+                        },
+                    )
```
chouinar, line 81: I'd recommend two additions to the `extra` here:
* Add the `notification_reason`
* Add the notification log ID - this is slightly trickier as it doesn't get set until it's flushed to the DB, so instead we'd need to set the ID ourselves when we create it above - just set it to `uuid.uuid4()`

api/src/task/notifications/closing_date_notification.py
```diff
@@ -0,0 +1,185 @@
...
+        db_session: db.Session,
+        app_id: str,
+        pinpoint_client: botocore.client.BaseClient | None = None,
+        frontend_base_url: str | None = None,
+    ):
+        super().__init__(db_session)
+        self.app_id = app_id
+        self.pinpoint_client = pinpoint_client
+        self.frontend_base_url = frontend_base_url
+        self.collected_data: dict | None = None
```
chouinar, line 38: Since the pinpoint stuff is shared between every notification task, can we move the `self.x = ...` bit up to that class for app ID and pinpoint client?

api/src/task/notifications/constants.py
```diff
@@ -0,0 +1,25 @@
+from dataclasses import dataclass
+from enum import StrEnum
+from typing import List
```
chouinar, line 3:
```suggestion

```

api/src/task/notifications/constants.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class NotificationReasons(StrEnum):
```
chouinar, line 6: Nit: Singular as something has a single notification reason
```suggestion
class NotificationReason(StrEnum):
```

api/src/task/notifications/closing_date_notification.py
```diff
@@ -0,0 +1,185 @@
...
+                        "opportunity_ids": [opp.opportunity_id for opp in saved_items],
+                    },
+                )
+
+    def run_task(self) -> None:
+        """Override to define the task logic"""
+        data = self.notification_data()
+        if data:
+            self.send_notifications(data, self.pinpoint_client, self.app_id)
+            self.create_user_opportunity_notification_log()
```
chouinar, line 185: Make sure to put `with db_session.begin():` around whatever we're running so it commits at the end.

api/src/task/notifications/opportunity_notifcation.py
```diff
@@ -0,0 +1,79 @@
+import logging
+from typing import List
```
chouinar, line 2:
```suggestion
```

api/src/task/notifications/base_notification.py
```diff
@@ -0,0 +1,89 @@
...
+            )
+            return None
+        return user.email
+
+    def send_notifications(
+        self,
+        data: EmailData,
+        pinpoint_client: botocore.client.BaseClient | None,
+        app_id: str,
+    ) -> None:
```
chouinar, line 57: If the pinpoint client and app ID are already set in the init, could we exclude passing them around here and just do `self.pinpoint_client`?

api/src/task/notifications/opportunity_notifcation.py
```diff
@@ -0,0 +1,79 @@
...
+            message = f"You have updates to {len(saved_items)} saved opportunities"
+            notification[user_id] = message
+            to_address_list.append(user_email)
+
+        return EmailData(
+            to_addresses=to_address_list,
+            subject="Updates to Your Saved Opportunities",
+            content=notification,
+            notification_reason=NotificationReasons.OPPORTUNITY_UPDATES,
+        )
```
chouinar, line 73: As I'm looking at this, I'm wondering if these two parts should be merged together.

In `collect_notifications` we do the DB query and make a weird map of `user_id` to a list of whatever the saved thing is for a given notification class.

In `prepare_notifications` we then iterate over everything we collected, add an email and a message.

What if we just did one pass, and did that altogether? Would that be an issue?

api/src/task/notifications/constants.py
```diff
@@ -0,0 +1,25 @@
...
+    SEARCHES_TRACKED = "searches_tracked"
+    NOTIFICATIONS_SENT = "notifications_sent"
+
+
+@dataclass
+class EmailData:
+    to_addresses: List[str]
+    subject: str
+    content: dict
+    notification_reason: NotificationReasons
```
chouinar, line 25: I'm not following the data model for this - do we make a single EmailData object and put every notification we're going to send out into it?

If 5 people need notifications, I might be reading the implementation of this elsewhere that EVERYONE receives EVERYONEs notifications - I think there's a bit of a different way we could construct this - an EmailData should be a single one for a given user, not a single object for everyone

api/src/task/notifications/opportunity_notifcation.py
babebe: I think its much readable and easier to test if separated and one we would still need to group it by the user, for aggregated info that is logged and in the message... which the group by could be done in the query but i think this is more readable.

api/src/task/notifications/constants.py
babebe: switched data model per user

chouinar (changes requested):
I'm not following the changes you've made here - it looks to be undoing a lot of what you had already done?

At its core, the process for every one of these email notifications should:
* Collect the emails we want to send out
* Send out those emails
* Update some tables so we don't re-notify

Collecting the emails is currently split up and you're running into some weird issues with passing that data around, make it simple, don't split that up.

Sending out the emails is defined in the base class.

Updating the tables at the end is actually the trickiest part as we have different tables to do that for, that depends on how we setup the data we pass around.

All of this together means that in the base class we implement something that looks like:
```
def run_task(self) -> None:
      records = self.collect_email_notifications()
      self.send_notifications(records)
      self.post_process_records(records)
```

Every notification job then needs to implement the collect + post process functions.

I would recommend we get that pattern working for one of the 3 jobs before trying to make it work for all of them.

api/tests/src/task/notifications/test_closing_date_notification.py
```diff
@@ -0,0 +1,201 @@
...
+    return user
+
+
+@pytest.fixture(autouse=True)
+def clear_data(db_session):
+    """Clear all notification logs"""
+    db_session.query(UserNotificationLog).delete()
+    db_session.query(UserOpportunityNotificationLog).delete()
+    cascade_delete_from_db_table(db_session, Opportunity)
+    db_session.query(UserSavedOpportunity).delete()
```
chouinar, line 33: Can we use cascade_delete_from_db_table for each of these?

api/src/task/notifications/closing_date_notification.py
```diff
@@ -0,0 +1,193 @@
...
+        self.frontend_base_url = frontend_base_url
+
+    def collect_email_notifications(self) -> list[UserEmailNotification]:
+        """Collect notifications for opportunities closing in two weeks"""
+        two_weeks_from_now = datetime_util.utcnow() + timedelta(days=14)
+
+        # Find saved opportunities closing in two weeks that haven't been notified
+        stmt = (
+            select(UserSavedOpportunity)
+            .options(joinedload(UserSavedOpportunity.user))
```
chouinar, line 43: selectinload seems to work better - I've seen cases where joinedload just doesn't return for big enough queries. Might be safer to use that instead.

api/tests/src/task/notifications/test_closing_date_notification.py
```diff
@@ -0,0 +1,201 @@
...
+    _clear_mock_responses()
+    assert len(notification_opportunity_logs) == 1
+
+    # Run the notification task
+    task_again = EmailNotificationTask(db_session, search_client)
+    task_again.run()
+
+    # Verify no emails were sent
+    mock_responses = _get_mock_responses()
+    print(mock_responses)
```
chouinar, line 154:
```suggestion
```

api/src/task/notifications/closing_date_notification.py
```diff
@@ -0,0 +1,193 @@
...
+                closing_opportunities.append(
+                    {
+                        "opportunity_id": opportunity.opportunity_id,
+                        "opportunity_title": opportunity.opportunity_title,
+                        "close_date": close_date,
+                    }
+                )
+
+            if closing_opportunities:
+                message = self._build_notification_message(closing_opportunities)
```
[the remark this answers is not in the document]
babebe: Yea, separating the logic made sense here and more readable

api/src/task/notifications/closing_date_notification.py
```diff
@@ -0,0 +1,193 @@
...
+                "total_closing_opportunities": sum(
+                    len(n.notified_object_ids) for n in users_email_notifications
+                ),
+            },
+        )
+
+        return users_email_notifications
+
+    def _build_notification_message(self, closing_opportunities: list) -> str:
+        message = "Applications for the following funding opportunity are due in two weeks:\n\n"
```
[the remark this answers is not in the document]
babebe: Will fallow-up here: https://github.com/HHS/simpler-grants-gov/issues/4926

api/src/task/notifications/base_notification.py
```diff
@@ -0,0 +1,98 @@
...
+        self,
+        db_session: db.Session,
+        pinpoint_client: botocore.client.BaseClient | None = None,
+        notification_config: EmailNotificationConfig | None = None,
+    ):
+        super().__init__(
+            db_session,
+        )
+
+        self.pinpoint_client = pinpoint_client
```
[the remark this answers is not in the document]
babebe: It is None here and get_pinpoint_client retrieves it. Not sure why it was setup as such. Removed

api/src/task/notifications/base_notification.py
```diff
@@ -0,0 +1,98 @@
...
+
+                self.increment(Metrics.USERS_NOTIFIED)
+
+            except Exception as e:
+                # Notification log will be updated in the finally block
+                logger.exception(
+                    f"Failed to send notification email: {str(e)}",
+                    extra={
+                        "user_id": user_notification.user_id,
+                        "email": user_notification.user_email,
```
chouinar, line 87: We do not want to log user emails ever.
```suggestion

```

api/src/task/notifications/base_notification.py
```diff
@@ -0,0 +1,98 @@
...
+                )
+                notification_log.notification_sent = True
+                user_notification.is_notified = True
+
+                self.increment(Metrics.USERS_NOTIFIED)
+
+            except Exception as e:
+                # Notification log will be updated in the finally block
+                logger.exception(
+                    f"Failed to send notification email: {str(e)}",
```
chouinar, line 84: `logger.exception` already puts the error on the extra object, don't add it twice.
```suggestion
                    "Failed to send notification email",
```

api/tests/src/task/notifications/test_generate_notifications.py
```diff
@@ -39,13 +41,14 @@ def setup_search_data(opportunity_index, opportunity_index_alias, search_client)
     search_client.swap_alias_index(opportunity_index, opportunity_index_alias)
 
 
-@pytest.fixture
+@pytest.fixture(autouse=True)
 def clear_notification_logs(db_session):
     """Clear all notification logs"""
     db_session.query(UserNotificationLog).delete()
+    db_session.query(UserOpportunityNotificationLog).delete()
```
chouinar, line 48: Should be using cascade_delete_from_db_table for these

chouinar (approved):

Issue #4663, opened 2025-04-18
Build fluentbit container to enable movement of logs to cloudwatch and new relic


chrisw-nava:
### Summary

Fluentbit will allow us to easily push a single set of logs to multiple destinations at once

[Slack thread](https://betagrantsgov.slack.com/archives/C05TSL64VUH/p1742840374607149) for reference

### Acceptance criteria

- [ ] Fluentbit container successfully pushes logs to both CloudWatch and New Relic

PR #4792, opened 2025-04-23
[Issue #4663] Builds an entire custom fluentbit container 

coilysiren:
## Summary

Relates to #4663

## Motivation

This is a well known way to get logs pushing to multiple locations! In particular we need logs in New Relic for observability, and Cloudwatch for auditing.

## Testing

I need whoever approves this to actually run through the testing directions! I don't want to be the only person who can run this.


infra/fluentbit/app-config/main.tf
```diff
@@ -0,0 +1,13 @@
...
+  project_name = module.project_config.project_name
+
+  environment_configs = {}
+  account_names_by_environment = {}
+  shared_network_name = "dev"
+}
+
+module "project_config" {
+  source = "../../project-config"
+}
```
coilysiren, line 13: I removed ALOT of stuff from this file... I like the end result

fluentbit/fluentbit.yml
```diff
@@ -0,0 +1,27 @@
...
+      match: "**"
+      license_key: ${licenseKey}
+
+    - name: cloudwatch_logs # aws cloudwatch logs
+      match: "**"
+      log_group_name: ${log_group_name}
+      log_stream_name: ${container_name}/${container_id}
+      region: ${aws_region}
+      auto_create_group: true
+      log_format: json
```
coilysiren, line 24: This is the important part

PR #4830, opened 2025-04-24
[Unticketed] Fix scenario where user declines to use login.gov

chouinar:
## Changes proposed
Adjusts behavior when a user declines to login to login.gov via our endpoint

## Context for reviewers
When implementing this, I didn't read the detail for `access_denied` very closely and assumed it was "us" being denied access to login.gov entirely. In reality, it's the user saying no. The change here makes it so we raise a 401 instead of a 500 (which both get redirected the same) which effectively just makes us avoid logging an error message / tripping an alert. No meaningful experience change will happen for a user.


mdragon (approved):
Looks good.

PR #4833, opened 2025-04-25
[Issue #4682] Create sam.gov client and mock client for local/dev

### LINKED ISSUE #4682: Create a client for interacting with sam.gov for extracts + make a mock client for local dev+tests
# Summary
We want a client that can call https://open.gsa.gov/api/sam-entity-extracts-api/ (and likely other endpoints later)

Note we don't have access yet, so also want a mock client that we can use for local development.

# Acceptance criteria
- [ ] Client created
- [ ] Mock client created that we can use locally / in unit tests

mikehgrantsgov:
## Summary
Fixes #4682

## Changes proposed
Added SAM.gov API client implementation to interact with the SAM.gov Extract API
Created BaseSamGovClient abstract interface with real and mock implementations
Implemented SamGovConfig for managing client configuration
Added factory function for creating appropriate client instances
Included models using Pydantic for request/response data structures
Added tests for all components

## Context for reviewers
Introduce a new adapter for the SAM.gov API that allows us to download extract files. Follows our standard adapter pattern. Supports both real API interactions and a mock mode for testing and development.

## Additional information
Is requests_mock correct here? Should we use another library instead?
See attached tests
Can be broken up into smaller PRs to assist review.


api/src/adapters/sam_gov/client.py
```diff
@@ -0,0 +1,320 @@
...
+        if include_auth:
+            # Add the API key authorization
+            if self.api_key:
+                headers["x-api-key"] = self.api_key
+
+            # Add Basic Auth if username and password are provided (needed for sensitive data)
+            if self.username and self.password:
+                auth_string = f"{self.username}:{self.password}"
+                encoded_auth = base64.b64encode(auth_string.encode()).decode()
+                headers["Authorization"] = f"Basic {encoded_auth}"
```
chouinar, line 130: The entity extract endpoint takes in an api_key path parameter, it doesn't support headers or user/password as far as I see.
https://open.gsa.gov/api/sam-entity-extracts-api/

api/src/adapters/sam_gov/config.py
```diff
@@ -0,0 +1,34 @@
...
+        return cls(
+            base_url=os.environ.get(
+                "SAM_GOV_API_BASE_URL", "https://open.gsa.gov/api/sam-entity-extracts-api"
+            ),
+            extract_url=os.environ.get(
+                "SAM_GOV_EXTRACT_URL", "https://open.gsa.gov/api/sam-entity-extracts-api/extracts"
+            ),
+            api_key=os.environ.get("SAM_GOV_API_KEY"),
+            timeout=int(os.environ.get("SAM_GOV_API_TIMEOUT", "30")),
+        )
```
chouinar, line 29: Why not use the base env config so we don't need to use several `os.environ.get` calls? We shouldn't be using os getenv style calls generally.

api/src/adapters/sam_gov/config.py
```diff
@@ -0,0 +1,34 @@
...
+
+from pydantic import BaseModel
+
+
+class SamGovConfig(BaseModel):
+    """Configuration for SAM.gov API client."""
+
+    base_url: str = "https://open.gsa.gov/api/sam-entity-extracts-api"
+    extract_url: str = "https://open.gsa.gov/api/sam-entity-extracts-api/extracts"
+    api_key: Optional[str] = None
```
chouinar, line 14:
```suggestion

from pydantic import BaseModel


class SamGovConfig(BaseModel):
    """Configuration for SAM.gov API client."""

    base_url: str = "https://open.gsa.gov/api/sam-entity-extracts-api"
    extract_url: str = "https://open.gsa.gov/api/sam-entity-extracts-api/extracts"
    api_key:str | None = None
```

chouinar (commented):
I think some wires got crossed - we want to connect to this endpoint: https://open.gsa.gov/api/sam-entity-extracts-api/ which returns just a file that you ask for. There's a lot I'm seeing here that doesn't come from the specification including auth which is a path param.

We do not want an implementation for the entity (non extract) endpoint at this time as we need to research that a bit more.

api/src/adapters/sam_gov/config.py
```diff
@@ -0,0 +1,26 @@
...
+from pydantic import Field
+from pydantic_settings import SettingsConfigDict
+
+from src.util.env_config import PydanticBaseEnvConfig
+
+
+class SamGovConfig(PydanticBaseEnvConfig):
+    """Configuration for SAM.gov API client."""
+
+    model_config = SettingsConfigDict(env_prefix="SAM_GOV_", populate_by_name=True, extra="ignore")
```
chouinar, line 12: Huh, didn't know about populate_by_name, although it seems like Pydantic recommends not using it in favor of `validate_by_name` and `validate_by_alias` instead:
https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.populate_by_name

api/src/adapters/sam_gov/factory.py
```diff
@@ -0,0 +1,61 @@
...
+
+from src.adapters.sam_gov.client import BaseSamGovClient, SamGovClient
+from src.adapters.sam_gov.config import SamGovConfig
+from src.adapters.sam_gov.mock_client import MockSamGovClient
+
+
+def create_sam_gov_client(
+    use_mock: bool = False,
+    config: SamGovConfig | None = None,
+    config_override: Dict[str, Any] | None = None,
```
chouinar, line 13: What is the config_override for? If someone wanted to override the config, couldn't they just make the config object and pass that in? I don't see it used outside of a test verifying it works.

api/src/adapters/sam_gov/factory.py
```diff
@@ -0,0 +1,61 @@
...
+
+from typing import Any, Dict
+
+from src.adapters.sam_gov.client import BaseSamGovClient, SamGovClient
+from src.adapters.sam_gov.config import SamGovConfig
+from src.adapters.sam_gov.mock_client import MockSamGovClient
+
+
+def create_sam_gov_client(
+    use_mock: bool = False,
```
chouinar, line 11: For this param and the two mock params, could those just go in the SamGovConfig itself? They'll be False/null/null for anything but using it locally, which is fine.

api/src/adapters/sam_gov/client.py
```diff
@@ -0,0 +1,242 @@
...
+        """
+        pass
+
+
+class SamGovClient(BaseSamGovClient):
+    """Client for interacting with the SAM.gov API."""
+
+    def __init__(
+        self,
+        api_key: str | SamGovConfig | None = None,
```
chouinar, line 47: What scenarios would we want to pass something in that isn't the config itself, which has the key? I'm not following the complexity we're adding here instead of just passing around a SamGovConfig itself with all the relevant params in it.

api/src/adapters/sam_gov/config.py
```diff
@@ -0,0 +1,26 @@
...
+
+    base_url: str = Field(
+        default="https://open.gsa.gov/api/sam-entity-extracts-api", alias="BASE_URL"
+    )
+    extract_url: str | None = Field(
+        default="https://open.gsa.gov/api/sam-entity-extracts-api/extracts", alias="EXTRACT_URL"
```
chouinar, line 18: The API endpoints are not under gsa, they're under sam.gov - see https://open.gsa.gov/api/sam-entity-extracts-api/#api-endpoints

api/src/adapters/sam_gov/client.py
```diff
@@ -0,0 +1,242 @@
...
+            Exception: For any other errors.
+        """
+        # Validate that file_name is provided
+        if not extract_request.file_name:
+            raise ValueError("file_name must be provided for downloads")
+
+        # Validate that API key is provided
+        if not self.api_key:
+            raise ValueError("API key must be provided for SAM.gov API access")
+
```
chouinar, line 111: I was wrong about the api_key before - completely missed the small detail it had about being able to pass it as a header param. Completely missed the `The API Key value must be sent as "x-api-key" under "Headers" and not directly in the request URL.` alternative approach. Could we put it in the header? That feels like the most secure approach of the ones offered.

api/src/adapters/sam_gov/client.py
```diff
@@ -0,0 +1,242 @@
...
+            Response object with download information.
+
+        Raises:
+            ValueError: If the request parameters are invalid or the extract is not found.
+            IOError: If there is an issue saving the file.
+            Exception: For any other errors.
+        """
+        # Validate that file_name is provided
+        if not extract_request.file_name:
+            raise ValueError("file_name must be provided for downloads")
```
chouinar, line 106: This isn't a nullable field on the request class, do we need to do this check?

api/src/adapters/sam_gov/client.py
```diff
@@ -0,0 +1,242 @@
...
+        # Validate that API URL is provided
+        if not self.api_url:
+            raise ValueError("API URL must be provided for SAM.gov API access")
+
+        # Build URL and parameters for file download
+        # Use extract_url if available, otherwise use api_url/download
+        if self.extract_url:
+            url = self.extract_url
+        else:
+            url = urljoin(self.api_url, "download")
```
chouinar, line 121: Could we just make it so the URL is a required field and not need to check/pass it in to the client?

api/src/adapters/sam_gov/client.py
```diff
@@ -0,0 +1,242 @@
...
+                filename = disposition_match.group(1)
+
+            # Save the file
+            with open_stream(output_path, "wb") as f:
+                for chunk in response.iter_content(chunk_size=8192):
+                    f.write(chunk)
+
+            # Build response object
+            extract_response = SamExtractResponse(
+                file_name=filename or os.path.basename(output_path),
```
chouinar, line 163: Should we just make this the output path? Any processing we'll do with this will need to go to s3 to grab it and would need the path.

api/src/adapters/sam_gov/client.py
```diff
@@ -0,0 +1,242 @@
...
+                file_size=content_length or os.path.getsize(output_path),
+                content_type=content_type,
+                sensitivity=SensitivityLevel.PUBLIC,
+                download_date=datetime.now(),
```
chouinar, line 167: How much do we care about these values? Are we going to use them for anything? I could see file size maybe being useful, but content type, and sensitivity level are static (also we're getting the FOUO file, not the public one).

api/src/adapters/sam_gov/client.py
```diff
@@ -0,0 +1,242 @@
...
+                method, url, params=params, json=data, stream=stream, headers=headers, **kwargs
+            )
+
+            # Raise an exception for HTTP errors
+            response.raise_for_status()
+            return response
+
+        except requests.exceptions.RequestException as e:
+            logger.error(f"Error when calling SAM.gov API: {e}")
+            raise
```
chouinar, line 242: logger.exception attaches the exception in a format that works with New Relic.
```suggestion
        except requests.exceptions.RequestException:
            logger.exception("Error when calling SAM.gov API")
            raise
```

api/src/adapters/sam_gov/mock_client.py
```diff
@@ -0,0 +1,173 @@
...
+        # If we have the file in mock_extract_dir, copy it to the output path
+        if self.mock_extract_dir and os.path.exists(os.path.join(self.mock_extract_dir, file_name)):
+            if is_s3_path(output_path):
+                # For S3 paths, we need to use open_stream to copy the file
+                with open(os.path.join(self.mock_extract_dir, file_name), "rb") as source_file:
+                    with open_stream(output_path, "wb") as target_file:
+                        target_file.write(source_file.read())
+            else:
+                # For local paths, we can use shutil.copyfile
+                shutil.copyfile(os.path.join(self.mock_extract_dir, file_name), output_path)
```
chouinar, line 125: Might be able to use `copy_file` in our file_util, doesn't handle copying between disk/s3, but handles the cases where the source and destination are the same.

If it helps simplify anything, since this is just a mock for local development, it's fine if we just simplify some of this to support _just_ our local s3, since it'll all be s3 non-locally.

mikehgrantsgov: I switched this to use `copy_file` and makes it much easier knowing it'll be an S3 destination both remotely and locally

api/src/adapters/sam_gov/client.py
mikehgrantsgov: I figured they might be useful for auditing later but I removed them now

api/src/adapters/sam_gov/client.py
mikehgrantsgov: Just made this the output path

api/src/adapters/sam_gov/config.py
mikehgrantsgov: Updated this and the path

api/src/adapters/sam_gov/client.py
mikehgrantsgov: I simplified this too

chouinar (commented):
There are a few outstanding comments from my last PR still

api/tests/src/adapters/sam_gov/test_factory.py
```diff
@@ -0,0 +1,89 @@
...
+            "SAM_GOV_BASE_URL": "https://test-api.sam.gov",
+            "SAM_GOV_USE_MOCK": "false",
+        },
+    )
+    def test_create_real_client(self):
+        """Test creating a real client."""
+        # Create config directly instead of relying on environment variables
+        config = SamGovConfig(
+            api_key="test-api-key", base_url="https://test-api.sam.gov", use_mock=False
+        )
```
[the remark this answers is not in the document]
mikehgrantsgov: We don't; removed these

api/src/adapters/sam_gov/config.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class SamGovConfig(PydanticBaseEnvConfig):
+    """Configuration for SAM.gov API client."""
+
+    model_config = SettingsConfigDict(
+        env_prefix="SAM_GOV_", validate_by_name=True, validate_by_alias=True, extra="ignore"
+    )
+
+    base_url: str = Field(default="https://api.sam.gov/", alias="BASE_URL")
```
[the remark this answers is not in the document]
mikehgrantsgov: Makes sense, dropped the prefix

api/src/adapters/sam_gov/factory.py
mikehgrantsgov: It's not needed -- it was an idea I had that during a first pass at the implementation and didn't remove it

api/src/adapters/sam_gov/factory.py
mikehgrantsgov: Updated this to be fully config-driven

api/src/adapters/sam_gov/client.py
mikehgrantsgov: It's a required field now, as we'll raise an error if it's not provided.

api/src/adapters/sam_gov/client.py
```diff
@@ -0,0 +1,212 @@
...
+            raise ValueError("API key must be provided for SAM.gov API access")
+
+        # Validate that API URL is provided
+        if not self.api_url:
+            raise ValueError("API URL must be provided for SAM.gov API access")
+
+        # Build URL and parameters for file download
+        url = urljoin(self.api_url, "data-services/v1/extracts")
+
+        params = {"fileName": extract_request.file_name, "api_key": self.api_key}
```
chouinar: I had missed when first reading the docs that you can pass the API key as a header param (which I think you had on a prior version of the PR), can we switch back to that, seems the safest approach. The `x-api-key` header param is only documented in a dropdown hidden on that page, but should be the way we go.

api/src/adapters/sam_gov/models.py
```diff
@@ -0,0 +1,82 @@
+"""Models for SAM.gov API client."""
+
+from enum import Enum
+
+from pydantic import BaseModel, Field
+
+
+class EntityStatus(str, Enum):
```
chouinar: StrEnum == `str, Enum`, but slightly simpler.
```suggestion
from enum import StrEnum

from pydantic import BaseModel, Field


class EntityStatus(StrEnum):
```

api/src/adapters/sam_gov/client.py
mikehgrantsgov: Made this update

api/src/adapters/sam_gov/models.py
```diff
@@ -1,11 +1,11 @@
 """Models for SAM.gov API client."""
 
-from enum import Enum
+from enum import Enum, StrEnum
```
chouinar: Could you make the other enums below also StrEnum?

mikehgrantsgov: Ah yeah, updated this.

PR #4857, opened 2025-04-29
[Issue #4807] Checkbox and `widget` field for schemas

### LINKED ISSUE #4807: Add checkbox and ability to choose widget type from the UI Schema
### Summary

Add checkbox option and the  ability to choose widget type from the UI Schema

### Acceptance criteria

- [ ] checkbox added
- [ ] `widget` option available for the UI schema
- [ ] docs updated

acouch:
## Summary
Fixes #4807

## Changes proposed
1. Update seed data so that the test form shows working widgets
2. Add `widget` key to the ui schema to allow the ui to determine which widget is used for the field
3. Add a `CheckboxWidget` component for check boxes
4. Update the `RadioWidget` to use the enum value instead of the array key of the enum
5. Add missing `FormGroup` to some widgets
6. Update the widget determination logic to be more clear
7. Update doc

## Context for reviewers
Had intended on adding the ability to choose the field type in the UI. The checkbox use case made this more urgent as we may want to save the checked value in the form data object as a string, and thus there is not a good way to infer from the schema that it should be a checkbox.

## Additional information
The form seed is now updated to include all field types. To test:

1. `make remake-backend` in the `/api` folder
2. `npm run dev` in the `/frontend` folder
3. Go to http://localhost:3000/formPrototype and click through to the test form. The test form should now have a full variety of field types:
image
4. Bang on the form by saving and submitting. The form should save even if there is an error with a required field.


frontend/src/app/[locale]/formPrototype/[applicationId]/form/[formId]/page.tsx
```diff
@@ -86,7 +86,11 @@ async function FormPage({ params }: formPageProps) {
 
   const schemaErrors = validateUiSchema(form_ui_schema);
   if (schemaErrors) {
-    console.error("Error validating form", form_ui_schema, schemaErrors);
+    console.error(
+      "Error validating form ui schema",
```
Comment (acouch) [inline frontend/src/app/[locale]/formPrototype/[applicationId]/form/[formId]/page.tsx:90]: More specific error message ("form" vs "form ui schema")

frontend/src/components/applyForm/types.ts
```diff
@@ -40,11 +40,22 @@ export type FieldErrors = ErrorObject<
...
 >[];
 
+export type WidgetTypes = "Checkbox" | "Text" | "TextArea" | "Radio" | "Select";
+
 export type UiSchemaField = {
   type: "field";
-  definition?: `/properties/${string}`;
-  schema?: SchemaField;
-};
+  widget?: WidgetTypes;
```
acouch, line 47: Added optional `widget` but also improved type checking for the ui schema. The field must have `definition`, `schema`, or both.

frontend/src/components/applyForm/widgets/SelectWidget.tsx
```diff
@@ -43,35 +44,36 @@ function SelectWidget<
   S extends StrictRJSFSchema = RJSFSchema,
   F extends FormContextType = never,
 >({
-  schema,
   id,
-  options,
-  value,
-  required,
   disabled,
+  options,
```
acouch, line 49: Move the props to be in alpha order for consistency across the widgets.

frontend/src/components/applyForm/widgets/SelectWidget.tsx
```diff
@@ -43,35 +44,36 @@ function SelectWidget<
...
   // passing on* functions made optional
   onChange = () => ({}),
   onBlur = () => ({}),
   onFocus = () => ({}),
 }: UswdsWidgetProps<T, S, F>) {
+  const { title, description } = schema;
   // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
   const { enumOptions: opts, enumDisabled, emptyValue: optEmptyVal } = options;
-  const { title, description } = schema;
+  const enums = useMemo(() => (opts && opts?.length > 0 ? opts : []), [opts]);
```
acouch, line 66: Use an empty array of select options if none are provided.

frontend/src/components/applyForm/widgets/TextAreaWidget.tsx
acouch, line 1: Clean up: alpha order for props, camel case for `TextArea` and include missing `<FormGroup/>`

frontend/src/components/applyForm/validate.ts
```diff
@@ -106,17 +110,7 @@ export const validateUiSchema = (data: object) => {
   return validateJsonBySchema(data, UiJsonSchema);
 };
 
-const getKeysWithValues = (formData: FormData) => {
```
acouch, line 109: This was removed for `shapeFormData` which gets the keys/values but also the correct field object shape.

frontend/src/components/applyForm/widgets/SelectWidget.tsx
doug-s-nava: [nit] don't think you need the optional chain since you're doing a null check for opts right before

frontend/tests/components/applyForm/widgets/CheckboxWidget.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+  schema: {
+    title: "I agree",
+    description: "Test description",
+  },
+  value: true,
+  required: true,
+  options: {},
+};
+
+describe("RadioWidget", () => {
```
doug-s-nava, line 16: CheckboxWidget?

frontend/src/components/applyForm/utils.tsx
```diff
@@ -112,67 +108,55 @@ export function buildFormTreeRecursive({
...
+  } else if (fieldSchema.type === "boolean") {
+    return "Checkbox";
+  } else if (fieldSchema.maxLength && fieldSchema.maxLength > 255) {
+    return "TextArea";
+  }
+  return "Text";
+};
+
+// either schema or definition is required, and schema fields take precedence
+export const getFieldSchema = ({
```
doug-s-nava, line 132: would love to see some tests for this one

doug-s-nava (commented):
this looks really good, I'm starting to see how this is will all come together. A few non blocking notes on the code, but more importantly I'm not able to get the form to save if there are errors. Tests steps suggest I should be able to, so want to make sure that's all set before I approve

frontend/src/components/applyForm/types.ts
```diff
@@ -40,11 +41,22 @@ export type FieldErrors = ErrorObject<
...
+} & (
+  | {
+      definition: `/properties/${string}`;
+      schema?: undefined;
```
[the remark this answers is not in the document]
doug-s-nava: yep, predictable that this union / intersection stuff was more complicated than I thought. We could work around this in the consuming code as described here, but agreed it's better to keep the type like this to avoid having to deal with the fallout in the code 👍 https://dpericich.medium.com/how-to-fix-typescript-property-does-not-exist-on-union-type-f83349cfb570

PR #4865, opened 2025-04-29
[Issue #4542] Create ECS task to get SAM.gov extracts [DRAFT]

### LINKED ISSUE #4542: Create a basic ECS task to fetch the sam.gov daily and monthly extracts
### Summary

After https://github.com/HHS/simpler-grants-gov/issues/4682 which will create a client+mock for fetching this info.

We may not have access yet when we build this, but we want to be able to fetch the monthly and daily extracts for the FOUO files from the sam.gov API

https://open.gsa.gov/api/sam-entity-extracts-api/

Monthly extracts are effectively full datasets, while the daily files are only changes since the prior extract. The logic should be setup according to the docs to fetch the specified files on the correct dates (assume we'll run the ECS task after the 7am time they mention).

Roughly speaking we should be doing the following:
* For monthly extracts, if we haven't processed the latest, fetch that
* For daily extracts, we should fetch all extracts we haven't processed since the last monthly extract

For example, if we run on the 10th of the month without any prior runs, we should fetch the monthly extract and all daily extracts after it. Effectively the job should catch it self up to the latest set of files.

This ticket does not need to account for any processing of these files yet, just collecting them (if however that is setup needs to mark files as "processed" go ahead and do so).

A few notes:
* Definitely make a table for storing the info about files processed
* We don't need to process daily files from before the most recent monthly extract
* When running in ECS we don't have local disk, if we need to put the file somewhere, put it on s3
* We will need to mock out actually calling sam.gov's endpoint for now

### Acceptance criteria

- [ ] File processing setup in a basic ECS task

mikehgrantsgov:
## Summary

Fixes #4542


api/src/task/sam_extracts/fetch_sam_extracts.py
```diff
@@ -0,0 +1,207 @@
...
+        self.s3_client = s3_client or get_s3_client()
+
+    def run_task(self) -> None:
+        """Main task logic to fetch SAM.gov extracts"""
+        with self.db_session.begin():
+            logger.info("Attempting to fetch monthly extract.")
+            self._fetch_monthly_extract()
+
+            logger.info("Attempting to fetch daily extracts for the current month.")
+            self._fetch_daily_extracts_for_month()
```
mikehgrantsgov: @chouinar very much simplified this process to run by getting the monthly extract if we haven't gotten it yet (first Sunday of the month) then, if necessary, get any dailies since then that haven't been retrieved. It would be nice to test this with a local API key, but this is the first step that would get us extracts we can subsequently process in another task.

api/src/task/sam_extracts/fetch_sam_extracts.py
```diff
@@ -0,0 +1,207 @@
...
+@ecs_background_task("fetch-sam-extracts")
+@flask_db.with_db_session()
+def run_fetch_sam_extracts_task(db_session: db.Session) -> None:
+    """Run the SAM.gov extracts fetching task"""
+    # Create the SAM.gov client using the factory
+    sam_gov_client = create_sam_gov_client()
+
+    # Initialize and run the task
+    task = SamExtractsTask(db_session, sam_gov_client)
+    task.run()
```
chouinar, line 46: We can leave it for now, but we'll eventually make it so this task is one of 3 we run in an ECS task (similar to how the transform job is several tasks).

api/src/task/sam_extracts/fetch_sam_extracts.py
```diff
@@ -0,0 +1,207 @@
...
+        The date of the first Sunday
+    """
+    # Start with the first of the month
+    d = date(year, month, 1)
+
+    # Add days until we hit a Sunday (where weekday() == 6)
+    while d.weekday() != 6:
+        d += timedelta(days=1)
+
+    return d
```
chouinar, line 236: I think this doesn't handle needing to go to the prior month. For example, May of 2025, the first Sunday of the month was May 4th. So, if we were running any day May 4th -> May 31st, it would be fine. But if we were running this on May 3rd, we actually want the April date (April 6th) as that's the most recent monthly extract.

Maybe we leave this function alone, but change how we use it - I _think_ this would work? If the date it found is after the current date, that extract doesn't exist yet, so go to the prior month.
```
target_date = get_first_sunday_of_month(current_date.year, current_date.month)
if target_date > current_date:
    # Definitely a better way to do these checks, but roughly go to the prior month
    if current_date.month == 1:
          target_date = get_first_sunday_of_month(current_date.year - 1, current_date.month)
    else:
          target_date = get_first_sunday_of_month(current_date.year, current_date.month - 1)
```

api/src/task/sam_extracts/fetch_sam_extracts.py
```diff
@@ -0,0 +1,207 @@
...
+        if existing:
+            logger.info(f"Monthly extract for {target_date} already exists")
+            return False
+
+        # Construct the filename based on the target date
+        filename = f"SAM_FOUO_MONTHLY_V2_{target_date.strftime('%Y%m%d')}.ZIP"
+
+        # Download the extract
+        s3_key = f"{self.config.s3_prefix}monthly/{filename}"
+        s3_path = f"s3://{self.config.s3_bucket}/{s3_key}"
```
chouinar: For the s3 path / how we configure it, instead of a dedicated configuration for the s3 prefix, could we just use the bucket from the s3 config? The way the terraform is setup means setting up s3 paths is a bit clunky, easier to just use the bucket (the draft bucket in this case).

You can see how I did that for the application attachment extracts: https://github.com/HHS/simpler-grants-gov/blob/main/api/src/services/applications/create_application_attachment.py#L71 - different path stuff after the bucket, but easiest to build the path this way.

api/src/task/sam_extracts/fetch_sam_extracts.py
```diff
@@ -0,0 +1,207 @@
...
+        return True
+
+    def _fetch_daily_extracts_for_month(self) -> None:
+        """Fetch all daily extracts for the current month that we don't already have."""
+        current_date = datetime_util.utcnow().date()
+        start_of_month = current_date.replace(day=1)
+
+        # Iterate through each day of the current month
+        for day_offset in range(current_date.day):
+            process_date = start_of_month + timedelta(days=day_offset)
```
chouinar: For this iteration, I think we need to adjust slightly. I think if we calculated the start-of-month file date and then passed it into the monthly/daily extract functions that might help?

I think we want every date between the start of month (but not including it) and the current date, BUT the docs also say: `They are produced every Tuesday-Saturday. Kindly check after 7 AM Eastern time.` so if the day is a Sunday or Monday we need to skip. Implicitly because the monthly extract is on sunday, we could start on the next day (which I guess also doesn't get anything?).

I think the easiest way to do that would be like (might be some off-by-one issues here, but general idea):
```py
date_to_process = monthly_extract_date

while date_to_process <= current_date:
     date_to_process = date_to_process + timedelta(days=1)
     
     if date_to_process.weekday() in (0, 6):
          continue
     
     # do file stuff
```

api/src/task/sam_extracts/fetch_sam_extracts.py
```diff
@@ -0,0 +1,207 @@
...
+
+            # Construct the filename for the daily extract
+            filename = f"SAM_FOUO_DAILY_V2_{process_date.strftime('%Y%m%d')}.ZIP"
+
+            # Check if we already have this extract by filename
+            stmt = select(SamExtractFile).where(SamExtractFile.filename == filename)
```
chouinar: Should this have a query like the monthly one where it's the date+type instead of the filename? That way seems safer

mikehgrantsgov: Updated this query

api/src/task/sam_extracts/fetch_sam_extracts.py
mikehgrantsgov: I updated to use the S3Config. I think this matches the example you shared.

api/src/task/sam_extracts/fetch_sam_extracts.py
mikehgrantsgov: Yep this makes sense. I added a few tests to cover this tweak.

api/src/task/sam_extracts/fetch_sam_extracts.py
mikehgrantsgov: Makes sense to me!

api/src/task/sam_extracts/fetch_sam_extracts.py
```diff
@@ -0,0 +1,250 @@
...
+            select(SamExtractFile.extract_date)
+            .where(
+                SamExtractFile.extract_type == extract_type,
+                SamExtractFile.processing_status == SamGovProcessingStatus.COMPLETED,
+            )
+            .order_by(SamExtractFile.extract_date.desc())
+            .limit(1)
+        )
+        result = self.db_session.execute(stmt).scalar_one_or_none()
+        return result
```
chouinar: I think this function isn't used anymore - from a prior approach

chouinar (approved):
LGTM - might need a merge from main, I know we have a few DB table PRs out at the moment.

chouinar (approved):

chouinar (approved):

PR #4868, opened 2025-04-30
Add ApplicationUser table

### LINKED ISSUE #4686: Add an ApplicationUser table
# Summary

Create a new table `ApplicationUser` with the following columns:
* `application_id` - UUID, primary key, foreign key to the application table
* `user_id` - UUID, primary key, foreign key to the user table

Note that in the future we'll likely include role information here (eg. "this user is an admin for this application") but as we haven't scoped that out yet, we'll leave it out for now.

# Acceptance criteria
- [ ] Table created

mikehgrantsgov:
## Summary

Fixes #4686 

## Changes proposed

Adds `ApplicationUser` table model and migration

## Context for reviewers
In the future we'll likely include role information here but not now as it is not presently scoped.

## Validation steps

Run migration, see created table.


api/src/db/models/competition_models.py
```diff
@@ -139,6 +144,10 @@ class Application(ApiSchemaTable, TimestampMixin):
         "ApplicationForm", uselist=True, back_populates="application", cascade="all, delete-orphan"
     )
 
+    users: Mapped[list["ApplicationUser"]] = relationship(
```
chouinar, line 147: I'd suggest we call this `application_users` so it's sightly more specific as these don't refer to the actual user records.

api/src/db/models/user_models.py
```diff
@@ -162,3 +167,21 @@ class UserOpportunityNotificationLog(ApiSchemaTable, TimestampMixin):
...
+        UUID,
+        ForeignKey("api.application.application_id"),
+        primary_key=True,
+    )
+    user_id: Mapped[uuid.UUID] = mapped_column(
+        UUID, ForeignKey("api.user.user_id"), primary_key=True
+    )
+
+    application: Mapped["Application"] = relationship("Application", back_populates="users")
+    user: Mapped["User"] = relationship("User", back_populates="applications")
```
chouinar, line 187: The types for these are defined above/imported, can we do?
```suggestion
    application: Mapped[Application] = relationship(Application, back_populates="users")
    user: Mapped[User] = relationship(User, back_populates="applications")
```

api/src/db/models/user_models.py
```diff
@@ -40,6 +41,10 @@ class User(ApiSchemaTable, TimestampMixin):
         viewonly=True,
     )
 
+    applications: Mapped[list["ApplicationUser"]] = relationship(
```
chouinar, line 44: Small nit, missed on first pass, can we call this `application_users`?

mikehgrantsgov: Updated this too, makes sense as it is clearer.

api/src/db/migrations/versions/2025_05_02_add_application_status_table_and_field.py
```diff
@@ -11,7 +11,7 @@
 
 # revision identifiers, used by Alembic.
 revision = "462b5b67c0ea"
-down_revision = "3b0705202c54"
+down_revision = "b4aad25f47bc"
```
chouinar, line 14: We can't do this - the migration has already been merged to main and run, it's too late to change the revision history.

mikehgrantsgov: Fixed that, kind of a side effect of having multiple PRs open with migrations

chouinar (approved):

PR #4874, opened 2025-04-30
[Issue #4380] auth token refresh on request

### LINKED ISSUE #4380: Refresh token on API requests
### Summary

After https://github.com/HHS/simpler-grants-gov/issues/3739, session tokens should be checked on each API request / route change. When this check occurs, if the token is still valid but set to expire within the next 10 minutes, the token should be refreshed using the existing API endpoint.

Note that 10 minutes is a pretty aggressive refresh, which we can relax once we have a browser activity based activity monitor in https://github.com/HHS/simpler-grants-gov/issues/3891, since we can be more targeted at that point and refresh on a less aggressive schedule

### Acceptance criteria

- [ ] whenever token is checked for validity, if token is valid but expiring within the next 10 minutes, token is refreshed using the existing API endpoint
- [ ] a user is able to stay logged in as long as they make an API request or change routes within a 15 minute window

doug-s-nava:
## Summary

Fixes #4380

## Changes proposed

Adds call to refresh auth token if necessary on each client side request to the Next API.

If a user is logged in and has a token expiring within 10 minutes, on an API request, a call will be made to the python API to refresh the token on that side, and then re-encrypt the login.gov token with a new expiration.

## Context for reviewers

* Refactors `createSession` to take in the expiration date for the session in order to make the function more reusable.
* Adds constants to track timing on expiration and refresh intervals

## Validation steps

1. to make testing this easier, set the client token refresh interval to the same value as the token expiration interval. This way the app will always try to refresh your token on API requests once you're logged in. Change line 2 of src/constants/auth.ts to
```
export const clientTokenRefreshInterval = 15 * 60 * 1000;
```
3. visit http://localhost:3000/search?_ff=authOn:true
4. log in
5. open dev tools and find the expiration date for your session cookie
6. _VERIFY_: it's roughly 15 minutes in the future
7. download a search result csv
8. check the session cookie
9. _VERIFY_: new token value, and expiration has been bumped forward a bit

### Bonus

1. repeat steps 7-9 for other actions that trigger API calls (save search or opportunity, etc.)

### Bonus part 2

1. undo change from first step one and repeat test steps
2. _VERIFY_: token and expiration on cookie do not update when you download the csv

### What this looks like in the dev tools

https://github.com/user-attachments/assets/d4317833-1fe2-4536-9888-ba5657b28e54

### DIFF
--- a/frontend/src/app/api/auth/callback/route.ts
+++ b/frontend/src/app/api/auth/callback/route.ts
@@ -1,4 +1,5 @@
 import { createSession } from "src/services/auth/session";
+import { newExpirationDate } from "src/services/auth/sessionUtils";
 
 import { redirect } from "next/navigation";
 import { NextRequest } from "next/server";
@@ -9,7 +10,7 @@ export async function GET(request: NextRequest) {
     return redirect("/unauthenticated");
   }
   try {
-    await createSession(token);
+    await createSession(token, newExpirationDate());
   } catch (_e) {
     console.error("Error creating session for token", { token });
     console.error(_e);
--- a/frontend/src/app/api/auth/session/route.ts
+++ b/frontend/src/app/api/auth/session/route.ts
@@ -1,4 +1,5 @@
-import { getSession } from "src/services/auth/session";
+import { getSession, refreshSession } from "src/services/auth/session";
+import { isExpiring } from "src/utils/dateUtil";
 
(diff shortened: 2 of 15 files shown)


acouch (approved):
👍🏿 We discussed sync, looks good.

Issue #4616, opened 2025-04-16
Reporting for Saved Opportunities and Saved Search Queries


mxk0:
### Summary

The Search v2 deliverable requires:

> At least 500 saved opportunities are created by at least 50 unique users
> At least 250 searches are saved by at least 25 unique users

This ticket is to create a dashboard that shows all of these metrics.

### Acceptance criteria

- [ ] Pipe the data to metabase for opportunities and searches
- [ ] Report published and shared in metabase

chrisw-nava:
Preliminary dashboard set up in NR. We'll get one mirrored to the backend as well for future public analytics: https://onenr.io/0yw4qW0MLR3

PR #4878, opened 2025-05-01
[Issue #4616] upload user saved opportunity s3 [1/3]

babebe:
## Summary

Fixes #[4616](https://github.com/HHS/simpler-grants-gov/issues/4616) 

Added tables `user_saved_opportunity` and `user_saved_searches` to opportunity extracts for analytics
Added Test
Migration file for analytics db for both tables

### DIFF
--- a/api/tests/src/api/users/test_user_update_saved_search.py
+++ b/api/tests/src/api/users/test_user_update_saved_search.py
@@ -19,6 +19,8 @@ def clear_data(db_session):
     db_session.query(UserSavedSearch).delete()
     db_session.query(UserTokenSession).delete()
     yield
+    db_session.query(UserSavedSearch).delete()
+    db_session.commit()
 
 
 def test_user_update_saved_search(client, db_session, user, user_auth_token, saved_search):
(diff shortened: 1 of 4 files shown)


api/tests/src/api/users/test_user_update_saved_search.py
```diff
@@ -19,6 +19,8 @@ def clear_data(db_session):
     db_session.query(UserSavedSearch).delete()
     db_session.query(UserTokenSession).delete()
     yield
+    db_session.query(UserSavedSearch).delete()
+    db_session.commit()
```
mikehgrantsgov, line 23: I think we want this after the yield?

mikehgrantsgov (changes requested):
Small change and I think it's good.

api/tests/src/api/users/test_user_update_saved_search.py
babebe: need to make sure it's deleted before running test and after words

mikehgrantsgov, line 23:
```suggestion
    db_session.query(UserSavedSearch).delete()
    db_session.commit()
    yield
```

babebe: this needs to run after yield as the "last" test would leave a record that will not get cleaned up

babebe: @mdragon We do need it, we want the delete to persist to the db before another session is started.

mikehgrantsgov (approved):

PR #4837, opened 2025-04-25
[Issue #2715] Opportunity Detail Page E2E Tests

austincanada:
## Summary
Fixes #2715 

## Changes proposed
Adds E2E tests for opportunity detail page component.

## Context for reviewers
E2E tests ran in local. E2E tests complete what was described.

PR #4880, opened 2025-05-01
[unticketed] fix issue with opportunity page content display toggle e2e tests

doug-s-nava:
## Summary

unticketed

## Changes proposed

due to some changes in the wording on the opportunity page that were not taken into account with tests introduced in https://github.com/HHS/simpler-grants-gov/pull/4837 there are some failing tests. This fixes the tests

## Validation steps

1. _VERIFY_: all opportunity page e2e tests pass locally and in CI

### DIFF
--- a/frontend/src/components/opportunity/OpportunityDescription.tsx
+++ b/frontend/src/components/opportunity/OpportunityDescription.tsx
@@ -88,42 +88,41 @@ const OpportunityDescription = ({
   );
 
   return (
-    <>
-      <div className="usa-prose margin-top-3">
-        <div className="display-block tablet:display-flex usa-prose flex-align-end">
-          <h2 className="flex-1">{t("title")}</h2>
-          <OpportunityDownload attachments={attachments} />
-        </div>
-        <SummaryDescriptionDisplay
-          summaryDescription={summary.summary_description || ""}
-        />
-        <h2>{t("eligibility")}</h2>
-        <h3>{t("eligible_applicants")}</h3>
-        <OpportunityEligibility
-          applicantTypes={summary.applicant_types || []}
-        />
-        <h3>{t("additional_info")}</h3>
-        <div
-          dangerouslySetInnerHTML={{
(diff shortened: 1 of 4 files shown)


frontend/src/components/opportunity/OpportunityDescription.tsx
```diff
@@ -88,42 +88,41 @@ const OpportunityDescription = ({
...
-          }}
-        />
-        <h4>{t("email")}</h4>
-        {summary?.agency_email_address_description && (
-          <p>{summary.agency_email_address_description}</p>
-        )}
-        <p>{agencyEmailLink}</p>
+    <div
+      className="usa-prose margin-top-3"
+      data-testid="opportunity-description"
```
doug-s-nava, line 93: this is the only real change - the rest is formatting after removing an unnecessary fragment wrapper

btabaska: small nit: 
Is there not a clean way of grabbing this without adding a data-testid?

btabaska (approved):

doug-s-nava:
@btabaska since both toggles use the same text, not really. I'd have to rely purely on targeting based on classes that are there for UI styling and would be subject to change if we updated the page. Open to that if you think it's preferable

frontend/tests/e2e/search/search-copy-url.spec.ts
```diff
@@ -1,16 +1,18 @@
...
-    timeout: 10000,
-  });
-
+  // this is dumb but webkit has an issue with trying to fill in the input too quickly
+  // if the expect in here fails, we give it another shot after 5 seconds
+  // this way we avoid an arbitrary timeout, and do not slow down the other tests
+  try {
+    await fillSearchInputAndSubmit("education grants", page);
+  } catch (e) {
+    await fillSearchInputAndSubmit("education grants", page);
```
btabaska, line 14: This feels a little hack-y but I don't like it less than the 10k timeout, so seems good for now.

Issue #4886, opened 2025-05-02
Add Playwright linter for e2e tests


acouch:
### Summary

Currently the `e2e` tests are in the `tests` folder which is linted by the jest linter which is not appropriate for the playwright tests.

Installing the playwright linter:

`npm install -D eslint-plugin-playwright`

and adding:

```javascript

    {
      files: ["tests/e2e/**"],
      extends: ["plugin:playwright/recommended"]
    },

```

to the `.eslintrc.js` file applies the correct linter.

When run this exposes a number of lint errors that will need to be fixed.

### Acceptance criteria

- [ ] playwright linting setup
- [ ] lint errors addressed

PR #4887, opened 2025-05-02
[Unticketed] Quiet linter in frontend

acouch:
## Changes proposed

1. Adds `*Fetch` dependencies to several React hooks. This won't change the hooks updating but quiets the linter warnings.
2. Adds `// eslint-disable-next-line react-hooks/exhaustive-deps` where empty hook dependency is justified and documented
3. updates `.eslintrc.js` to no longer target the e2e tests
4. updates a number of e2e tests to no longer ignore testing-library notices
5. updates `tests/jest.setup.js` to include missing declaration
6. fixes a test file name error

## Validation steps

All tests should pass

Running `npm run lint` should not produce any linting errors.

### DIFF
--- a/frontend/.eslintrc.js
+++ b/frontend/.eslintrc.js
@@ -30,7 +30,7 @@ module.exports = {
   overrides: [
     // Lint config specific to Test files
     {
-      files: [ "tests/**/*.test.tsx"],
+      files: ["tests/**/*.test.tsx"],
       plugins: ["jest"],
       extends: [
         "plugin:jest/recommended",
--- a/frontend/.eslintrc.js
+++ b/frontend/.eslintrc.js
@@ -30,7 +30,7 @@ module.exports = {
   overrides: [
     // Lint config specific to Test files
     {
-      files: ["tests/**"],
+      files: [ "tests/**/*.test.tsx"],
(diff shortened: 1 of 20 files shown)


acouch:
I created two follow-ups: https://github.com/HHS/simpler-grants-gov/issues/4886 and https://github.com/HHS/simpler-grants-gov/issues/4888

frontend/.eslintrc.js
```diff
@@ -30,7 +30,7 @@ module.exports = {
   overrides: [
     // Lint config specific to Test files
     {
-      files: ["tests/**"],
+      files: ["tests/**/*.test.tsx"],
```
doug-s-nava, line 33: 👍

doug-s-nava (dismissed):
thanks for this!

a related thought is that we should discuss whether we want to use testing-library at all in our playwright tests. Playwright gives you `locator` so we shouldn't have to use `getByText` etc. in those tests. Sometimes it's more convenient, but it also feels like we're mixing the peanut butter and the jelly, and it's a little confusing the way it is now.

✅ ✅ ✅ ✅

myduong-navapbc (dismissed):
LGTM

PR #4904, opened 2025-05-02
[Issue #4892] Passthrough Client Certificates at the LB 

### LINKED ISSUE #4892: Enable mTLS on api load balancers in terraform
### Summary

We need to support passing through ssl certs from S2S api users from SGG soap proxy. In order to capture obtain the cert in our app, we need to enable mTLS on our load balancers to have the specific header provided in the request.

[This documentation](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lb_listener#mutual_authentication) holds information around enabling mTLS.

We need to update terraform resources to enable this feature in our environments. 

### Acceptance criteria

- mTLS is enabled on our api load balancers.

mdragon:
## Summary

Fixes #4892

## Changes proposed
Make it so the API LBs will passthrough the client certificate that was given to them (which will be present if it's a SOAP Proxy request)

## Context for reviewers
Untested, would see how it worked when applied to Dev.

### DIFF
--- a/infra/modules/service/load_balancer.tf
+++ b/infra/modules/service/load_balancer.tf
@@ -83,7 +83,7 @@ resource "aws_lb_listener" "alb_listener_https" {
   protocol          = "HTTPS"
   certificate_arn   = var.certificate_arn
   mutual_authentication {
-    mode = "passthrough"
+    mode = module.app_config.app_name == "api" ? "passthrough" : "off"
   }
 
   # Use security policy that supports TLS 1.3 but requires at least TLS 1.2
--- a/infra/modules/service/load_balancer.tf
+++ b/infra/modules/service/load_balancer.tf
@@ -83,7 +83,7 @@ resource "aws_lb_listener" "alb_listener_https" {
   protocol          = "HTTPS"
   certificate_arn   = var.certificate_arn
   mutual_authentication {
-    mode = module.app_config.app_name == "api" ? "passthrough" : "off"
+    mode = startswith(var.service_name, "api-") ? "passthrough" : "off"
   }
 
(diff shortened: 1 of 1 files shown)


infra/modules/service/load_balancer.tf
```diff
@@ -82,6 +82,9 @@ resource "aws_lb_listener" "alb_listener_https" {
   port              = 443
   protocol          = "HTTPS"
   certificate_arn   = var.certificate_arn
+  mutual_authentication {
+    mode = module.app_config.app_name == "api" ? "passthrough" : "off"
```
coilysiren, line 86:
```suggestion
    mode = startswith(var.service_name, "api-") ? "passthrough" : "off"
```

There's a context and variable scoping issue here, in that `module.app_config` won't be available at this location. You can access `var.service_name` though.

coilysiren (commented):
I'm not sure how this works, but I'm fine with merging it to test. We just need to make absolutely sure that it doesn't totally break the load balancer before Wednesday deploy

infra/modules/service/load_balancer.tf
mdragon: > There's a context and variable scoping issue here, in that module.app_config won't be available at this location.

I was worried about this, when nothing else in this folder referenced module. Made it two commits to make it easy to rollback.

> would be to add a enable_mutual_authentication variable that's false by default but the API sets it to true

Passthrough is just one of two ways you can configure the client auth and they require different additional config, so it's almost more like a whole object we might want to pass rather than the true/false? Maybe?

coilysiren (approved):

PR #4907, opened 2025-05-02
[Issue #4616] Load User Saved Opportunities [2/3]

### LINKED ISSUE #4616: Reporting for Saved Opportunities and Saved Search Queries

babebe:
## Summary

Fixes #4616 

## Changes proposed

Load  `user_saved_opportunity` and `user_saved_search` to analytics-db 
Update test


analytics/tests/conftest.py
```diff
@@ -387,19 +387,25 @@ def _create_opportunity_table(conn: EtlDb.connection, schema: str) -> None:
         test_file_path = Path(__file__).resolve()
 
         # Construct the path to the SQL file
-        sql_file_path = (
+        migrations_path = (
```
chouinar, line 390: I can't recall, but what was the reason we didn't just run all of the migrations in order here rather than just running the migration files we wanted?

babebe: Can't recall but i don't see an issue with creating it. Its the only test that makes use of it.

Issue #4912, opened 2025-05-02
Run nofos tests in CI


coilysiren:
### Summary

//

### Acceptance criteria

- `make test`

PR #4913, opened 2025-05-02
[Issue #4912] Adds CI for nofos

coilysiren:
## Ticket

https://github.com/HHS/simpler-grants-gov/issues/4912

## Testing

I won't actually be able to test this until after I merge, because Github Actions can only be run once they have been merged to main at least once. (As far as I know!)


pcraig3:
Great, let's merge. This should work [now that `make test` works](https://github.com/HHS/simpler-grants-pdf-builder/pull/308).

PR #4918, opened 2025-05-05
Dschrashun/4903 feature flags default

doug-s-nava:
## Summary

Fixes #ISSUE

### DIFF
--- a/frontend/src/components/LoginModal.tsx
+++ b/frontend/src/components/LoginModal.tsx
@@ -1,5 +1,6 @@
 "use-client";
 
+import { useIsSSR } from "src/hooks/useIsSSR";
 import SessionStorage from "src/services/auth/sessionStorage";
 
 import { RefObject } from "react";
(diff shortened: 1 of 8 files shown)


acouch (dismissed):
Looks good. Think it would be helpful to mention the non-issue items in the description next time around (I assume the `useIsSSR` update is not related to the FF).

Would this update allow use to use `false` as a default, ie `applyFormPrototypeOff: true` could be `applyFormPrototypeOn: false`?

doug-s-nava:
@acouch yeah, good point about the `useSsr` thing - that ended up being the fix for the issue I mentioned at `tracked down a tbd error around a reference to the document in server side pre-rendered code...`, but I forgot to go back and update the description after I figured out the fix. The issue was that with the feature flags turned on during build time for the saved search functionality, the Trussworks modals were being rendered server side, which doesn't work unless you tell it, like not to render them.

As for switching the polarity on the page level flags, this doesn't touch that. We'd need to adjust the behavior of the `withFeatureFlag` wrapper to get that working. Here's a ticket https://github.com/HHS/simpler-grants-gov/issues/5010

acouch (dismissed):

acouch (approved):

myduong-navapbc (approved):

PR #4922, opened 2025-05-05
[Issue #4519] Update headers for security

### LINKED ISSUE #4519: Add `X-Content-Type-Options` and `Content-Security-Policy` headers to better secure frontend
### Summary

The HTTP X-Content-Type-Options [response header](https://developer.mozilla.org/en-US/docs/Glossary/Response_header) indicates that the [MIME types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types) advertised in the [Content-Type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Type) headers should be respected and not changed.

Next.js has an easy way set these: 

* https://nextjs.org/docs/app/api-reference/config/next-config-js/headers#x-content-type-options
* https://nextjs.org/docs/app/building-your-application/configuring/content-security-policy

### Acceptance criteria

- [ ] headers added to better secure frontend

acouch:
## Summary

Fixes #4519

## Changes proposed

This adds the following headers to improve security:

* [X-Content-Type-Options](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Content-Type-Options) set to `no-sniff` to to avoid [MIME type sniffing](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types#mime_sniffing)
* [Strict-Transport-Security](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Strict-Transport-Security), aka [HSTS](https://developer.mozilla.org/en-US/docs/Glossary/HSTS), lets browsers know to always use HTTPS for the site instead of trying to make an initial request that is insecure
* [Content-Security-Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy) helps stop cross-site scripting by regulating what resources can be domains can be loaded and from which domains. This could be done more securely if we switch to all dynamic content by using [nonces](https://nextjs.org/docs/app/guides/content-security-policy) to ensure assets are actually coming from our servers.
* [X-Frame-Options](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Frame-Options) doesn't allow other sites to embed our site content in iframes. This is little "un-open webby", and maybe at some point we would want to allow other sites to embed our search or something, but on the other hand this helps ensure that these assets are not embedded in insecure contexts.

## Context for reviewers

This is good security policy, as well as helping to promote best practices as sites like https://dri.es/headers?url=https://simpler.grants.gov and scangov.org are publicly evaluating the site.

## Validation steps

Everything should work locally the same. To QA you could click around and try a couple of user actions. HSTS [doesn't appear to be an issue](https://issues.chromium.org/issues/41251622) with localhost anymore.

### DIFF
--- a/.github/workflows/ci-frontend-e2e.yml
+++ b/.github/workflows/ci-frontend-e2e.yml
@@ -61,7 +61,7 @@ jobs:
           cat .env.development >> .env.local
           # Use the localhost url for tests
           sed -i 's|^SENDY_API_URL=.*|SENDY_API_URL=http://localhost:3000|' .env.local
-          IS_CI=true npm run build -- --no-lint
+          npm run build -- --no-lint
 
       - name: Run e2e tests (Shard ${{ matrix.shard }}/${{ matrix.total_shards }})
         env:
--- a/.github/workflows/ci-frontend-e2e.yml
+++ b/.github/workflows/ci-frontend-e2e.yml
@@ -26,8 +26,8 @@ jobs:
 
     strategy:
       matrix:
-        shard: [1, 2, 3]
-        total_shards: [3]
+        shard: [1, 2, 3, 4]
+        total_shards: [4]
 
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/ci-frontend-e2e.yml
+++ b/.github/workflows/ci-frontend-e2e.yml
@@ -61,7 +61,7 @@ jobs:
           cat .env.development >> .env.local
           # Use the localhost url for tests
           sed -i 's|^SENDY_API_URL=.*|SENDY_API_URL=http://localhost:3000|' .env.local
-          npm run build -- --no-lint
+          IS_CI=true npm run build -- --no-lint
 
       - name: Run e2e tests (Shard ${{ matrix.shard }}/${{ matrix.total_shards }})
         env:
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -14,6 +14,22 @@ const nrExternals = require("@newrelic/next/load-externals");
 const basePath = process.env.NEXT_PUBLIC_BASE_PATH;
 const appSassOptions = sassOptions(basePath);
 
+const cspHeader = `
+    default-src 'self';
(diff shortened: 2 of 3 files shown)


doug-s-nava (dismissed):
✅ ✅ ✅ ✅ 

very cool

I can't resist casting some Next JS shade here - shouldn't a framework this heavy build in some plug and play support for this kind of thing?

acouch:
@doug-s-nava I expanded the matrix for the e2e tests.

.github/workflows/ci-frontend-e2e.yml
```diff
@@ -61,7 +61,7 @@ jobs:
           cat .env.development >> .env.local
           # Use the localhost url for tests
           sed -i 's|^SENDY_API_URL=.*|SENDY_API_URL=http://localhost:3000|' .env.local
-          npm run build -- --no-lint
+          IS_CI=true npm run build -- --no-lint
```
doug-s-nava, line 64: it looks like a `CI` env var is being set in the make task, and there's some stuff in the application using that variable so I'm assuming it's coming through. Is it possible to use that env var rather than creating a new one here?

frontend/next.config.js
```diff
@@ -14,90 +14,134 @@ const nrExternals = require("@newrelic/next/load-externals");
...
+    headers: [
+      {
+        key: "Cache-Control",
+        value: "no-store, must-revalidate",
+      },
+    ],
+  },
+];
+
+const isCi = Boolean(process.env.IS_CI);
```
acouch, line 133:
```suggestion
const isCi = Boolean(process.env.CI);
```

.github/workflows/ci-frontend-e2e.yml
acouch: But IS... it CI?....

Good catch, testing.

doug-s-nava (approved):

doug-s-nava (approved):

PR #4954, opened 2025-05-06
[Issue #4688] Multiauth support for our endpoints

### LINKED ISSUE #4688: [Spike] - Check if we can have multiple auth types on a single endpoint
# Summary
With APIFlask we can add different auth types (user or API key right now) to a given endpoint, but in the future may need auth to work in a way that can do either or - see what it would take to do that.

We would want the behavior to be that not providing auth for an endpoint only errors if neither auth type is provided. Can we do that by adding both auths to an endpoint?

Worst case, we need to build a 3rd auth type that manually handles either approach (just handling this OR logic).

# Acceptance criteria
- [ ] Figure out the behavior we want
- [ ] Create a follow-up ticket to make any necessary adjustments

chouinar:
## Summary

Fixes #4688

## Changes proposed
* Added a utility for supporting multiple distinct authentication methods on a single endpoint

## Context for reviewers
Some of our upcoming work will want a way to authenticate as either a user or with an API key. This started as a simple test to see if that was possible, and by the time I proved I could do it, most of it was written, so rather than have follow-up tickets to implement it, I just cleaned up my prototype a bit.

The main complexity here is that APIFlask does not support this, so we can't use APIFlask's decorator, and have to fallback to doing it the flask way and then manually tell APIFlask what auth we support in the docs (this only affects OpenAPI letting you choose the auth, it doesn't affect anything about the actual behavior). Mildly annoying, but it's a two-line difference from what we were doing before for any endpoint we want this supported on.

## Validation steps
None of the endpoints actually have this enabled yet, but in addition to the unit tests that verify the behavior, I temporarily modified the GET opportunity endpoint to test this, just follow the details I added in the documentation to temporarily switch it over if you want to validate it.


api/src/auth/api_key_auth.py
```diff
@@ -18,17 +18,17 @@
 
 
 @dataclass
-class User:
+class ApiKeyUser:
```
chouinar, line 21: I renamed this for now to avoid it having the same name as our User table objects which I found confusing. We'll likely redo a lot of this whenever we re-implement the API key auth - this'll probably just get replaced with something like an organization.

freddieyebra (approved):
Looks good. Verified that the @jwt_or_key_multi_auth.login_required works on different endpoints that require auth.

PR #4971, opened 2025-05-07
[Issue #4907] Add frontend CSS/USWDS guidance

### LINKED ISSUE #4097: Document frontend markup expectations
### Summary

In #4032 @andycochran noted:

> When implementing a design, we should first configure whatever theme setting we can to bring us closer to the design. If settings alone are not enough, we should then apply utility classes. If that is still not enough, then we get into writing CSS.
> I'd like to establish the norm that any custom CSS is a code smell. We can do almost anything we need by using USWDS utility classes. Definitely anything we need to do, as we rarely need to stray from USWDS patterns. Style changes should be thought of in a particular order:
> * Can I define a global USWDS system setting to achieve the goal?
> * If not, can I apply USWDS utility classes?
> * If not, can the CSS I write leverage USWDS mixins and functions?
> * If not, can it at least use system tokens?

This should be cleaned up and added to the [frontend dev documentation](https://github.com/HHS/simpler-grants-gov/tree/main/documentation/frontend.

### Acceptance criteria

- [ ] Documentation updated

andycochran:
## Summary

Fixes #4097

## Changes proposed

- Add "CSS and Design System" in frontend docs 

## Context for reviewers

This guidance is intended to help frontend contributors understand how to avoid custom CSS and diverging from USWDS defaults. 

## Validation steps

n/a


doug-s-nava (approved):
10/10 no notes

✅ ✅ ✅ ✅

PR #4989, opened 2025-05-08
[Issue #4706] Add soft delete to user-saved-opportunities-search

### LINKED ISSUE #4706: Implement soft deletes for saved opportunities and saved searches
### Summary

We want to be able to see a record of deleted rows in `user_saved_opportunity` and `user_saved_search` tables for historical, reporting and debugging purposes

### Acceptance criteria

- [ ] boolean `deleted` column is added (nullable or default to false)
- [ ] delete behavior is adjusted to updated the `deleted` column to true instead of deleting a row on DELETE actions
- [ ] GET actions return only rows where `deleted = FALSE`
- [ ] POST / PUT actions look first to see if a deleted user_id / opportunity_id pair exists, and if so, sets it to `deleted = FALSE` instead of creating a new row

babebe:
## Summary

Fixes #4706 

## Changes proposed
delete behavior is adjusted to updated the deleted column to true instead of deleting a row on DELETE actions
GET actions return only rows where deleted = FALSE
POST / PUT actions look first to see if a deleted user_id / opportunity_id pair exists, and if so, sets it to deleted = FALSE instead of creating a new row
Tests updated/created


api/src/db/models/user_models.py
```diff
@@ -113,6 +113,7 @@ class UserSavedOpportunity(ApiSchemaTable, TimestampMixin):
     opportunity: Mapped[Opportunity] = relationship(
         "Opportunity", back_populates="saved_opportunities_by_users"
     )
+    is_deleted: Mapped[bool] = mapped_column(default=False)
```
chouinar, line 116: Where is the migration for this change?

If you run this migration while records exist in the table already, what happens? I think default doesn't retroactively set the column, so it would fail.

Options are:
* Allow the column to be null - treat null and false the same
* A two-step migration, first add the column as a nullable column, then set the value for each existing to false, then make it not nullable.

api/src/services/users/create_saved_opportunity.py
```diff
@@ -0,0 +1,25 @@
...
+    record = db_session.execute(
+        select(UserSavedOpportunity).filter(
+            UserSavedOpportunity.user_id == user_id,
+            UserSavedOpportunity.opportunity_id == json_data["opportunity_id"],
+        )
+    ).scalar_one_or_none()
+
+    if record:
+        record.is_deleted = False
+    else:
```
chouinar, line 20: The if-else here is a good case for something we'd want to be able to easily track in case a user hits a weird issue. I'd recommend adding a log message for each case.

api/tests/src/api/users/test_user_delete_saved_opportunity.py
```diff
@@ -45,18 +45,10 @@ def test_user_delete_saved_opportunity(
...
-        f"/v1/users/{user.user_id}/saved-opportunities/1234567890",
-        headers={"X-SGG-Token": user_auth_token},
-    )
-
-    assert response.status_code == 404
-    assert response.json["message"] == "Saved opportunity not found"
+    # Verify it was soft deleted
+    saved_opp = db_session.query(UserSavedOpportunity).all()
+    assert len(saved_opp) == 1
+    assert saved_opp[0].is_deleted
```
chouinar, line 51: Following up on what we talked about the other day, a general suggestion for tests like this - don't rely on data being deleted / don't delete data.

Right now this:
* Deletes all existing data
* Runs
* Fetches all existing data assuming it's just what we did

This approach works, but doesn't account for a case where other data exists. What if we had a bug where we were missing a `where user_id = ..` in a query somewhere and instead fetching data for another user (we had a bug like this before). Having other data would have found that because it would have broken the test.

You don't need to do it in this PR, but I'd highly recommend a ticket be created that:
* Does not delete data up front
* The query to check what was created filters by user ID - that way any other stuff isn't fetched

api/tests/src/api/users/test_user_saved_opportunities_get.py
```diff
@@ -264,3 +264,27 @@ def test_user_get_only_own_saved_opportunities(
...
+        json={
+            "pagination": {
+                "page_offset": 1,
+                "page_size": 25,
+            }
+        },
+    )
+
+    assert response.status_code == 200
+    assert len(response.json["data"]) == 0
```
chouinar, line 290: I'd recommend making this test have a mix of deleted and not-deleted records. Testing things separately (ie. a user has only one deleted, or only one non-deleted opportunities) doesn't always work and it's good practice to test a realistic case where a user has a mix of scenarios.

api/src/db/models/user_models.py
babebe: updated as Nullable col

api/src/services/users/delete_saved_opportunity.py
```diff
@@ -9,10 +9,12 @@
 
 def delete_saved_opportunity(db_session: db.Session, user_id: UUID, opportunity_id: int) -> None:
     result = db_session.execute(
-        delete(UserSavedOpportunity).where(
+        update(UserSavedOpportunity)
+        .where(
```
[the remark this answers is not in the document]
babebe: It will return 0.

api/src/services/users/create_saved_search.py
```diff
@@ -11,8 +13,28 @@
...
     search_client: search.SearchClient, db_session: db.Session, user_id: UUID, json_data: dict
 ) -> UserSavedSearch:
+    # Check if a saved search exists
+    record = db_session.execute(
+        select(UserSavedSearch).filter(
+            UserSavedSearch.user_id == user_id,
+            UserSavedSearch.search_query == json_data["search_query"],
+            UserSavedSearch.name == json_data["name"],
+        )
+    ).scalar_one_or_none()
```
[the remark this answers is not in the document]
chrisw-nava: @babebe - I am good with Michael's suggestion. When someone "deletes" it, they probably think it's gone entirely anyways. When someone "un-saves" an opportunity, they don't expect it to go anywhere other than off their screen, so this distinction makes sense.

api/src/services/users/delete_saved_search.py
```diff
@@ -1,20 +1,20 @@
...
 def delete_saved_search(db_session: db.Session, user_id: UUID, saved_search_id: UUID) -> None:
-    saved_search = db_session.execute(
-        select(UserSavedSearch).where(
+    result = db_session.execute(
+        update(UserSavedSearch)
+        .where(
             UserSavedSearch.saved_search_id == saved_search_id, UserSavedSearch.user_id == user_id
         )
-    ).scalar_one_or_none()
+        .values(is_deleted=True)
```
chouinar, line 16: Could we leave this as it was before? Just instead of deleting it, we set the value on the saved search record returned?

api/src/services/users/delete_saved_opportunity.py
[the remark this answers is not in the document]
chouinar: I'd really prefer the fetch and update the record approach over an update query.

It's just inherently less risky / automatically validates that we fetched exactly 0 or 1 (if more got fetched it'll 500 which shows we have a bug). As written here, a bug where we update 20 rows wouldn't be noticed. It's unlikely, but easy to make that mistake when models get changed.

api/src/services/users/get_saved_searches.py
```diff
@@ -22,7 +22,9 @@ def get_saved_searches(
 
     search_params = SavedSearchListParams.model_validate(raw_search_params)
 
-    stmt = select(UserSavedSearch).where(UserSavedSearch.user_id == user_id)
+    stmt = select(UserSavedSearch).where(
+        (UserSavedSearch.user_id == user_id) & (UserSavedSearch.is_deleted.isnot(True))
+    )
```
chouinar, line 27: I'm not sure this syntax is accepted - at the very least it's not documented: https://docs.sqlalchemy.org/en/20/tutorial/data_select.html#the-where-clause

where also accepts multiple comma separated clauses, so can just trim it to:
```suggestion
    stmt = select(UserSavedSearch).where(UserSavedSearch.user_id == user_id, UserSavedSearch.is_deleted.isnot(True))
```

chouinar (approved):
LGTM, but I'd recommend testing the behavior out in the lower envs before the prod deploy

PR #4992, opened 2025-05-08
fix: fixes tables that blow out when columns are very wide

### LINKED ISSUE #4413: Long file names break documents table
### Summary

When an opportunity has long enough file names the documents table on the opportunity detail page breaks.

See https://simpler.grants.gov/opportunity/314757

Image

### Reproducibility

### How to reproduce the bug

Visit https://simpler.grants.gov/opportunity/314757

### Browser

Firefox

### Code of Conduct

myduong-navapbc:
## Summary

Fixes #4413

## Changes proposed

`
.usa-table {
  &.overflow-wrap {
    max-width: 0;
    word-wrap: break-word;
  }
}
`

I've added some styles to the  newly added `.overflow-wrap` class to allow for word breakage on overflow.

`max-width: 0;`

is a fix for tables inside of flex

## Context for reviewers
Screenshot 2025-05-08 at 8 13 47 AM

checked in firefox, chrome and safari

### DIFF
--- a/frontend/src/styles/_uswds-theme-custom-styles.scss
+++ b/frontend/src/styles/_uswds-theme-custom-styles.scss
@@ -487,8 +487,9 @@ button.usa-pagination__button.usa-button {
 }
 
 // Tables
-.usa-table--borderless td,
-.usa-table--borderless th {
-  max-width: 0; // magic to prevent tables that are nested inside of a flex element from blowing out
-  word-wrap: break-word;
+.usa-table {
+  &.overflow-wrap {
+    max-width: 0; // magic to prevent tables that are nested inside of a flex element from blowing out
+    overflow-wrap: break-word;
+  }
 }
--- a/frontend/src/styles/_uswds-theme-custom-styles.scss
+++ b/frontend/src/styles/_uswds-theme-custom-styles.scss
@@ -485,3 +485,10 @@ button.usa-pagination__button.usa-button {
   align-items: baseline;
   gap: 5px;
 }
+
+// Tables
+.usa-table--borderless td,
+.usa-table--borderless th {
+  max-width: 0; // magic to prevent tables that are nested inside of a flex element from blowing out
+  word-wrap: break-word;
+}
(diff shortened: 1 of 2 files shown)


frontend/src/styles/_uswds-theme-custom-styles.scss
```diff
@@ -485,3 +485,10 @@ button.usa-pagination__button.usa-button {
   align-items: baseline;
   gap: 5px;
 }
+
+// Tables
+.usa-table--borderless td,
```
doug-s-nava, line 490: I'd recommend adding a class specifically to designate a non-expanding table in this case rather than targeting this existing class

`.usa-table--borderless` is applied by the Trussworks component, so if they update that component to use a different class name, our styling would break, and on the other hand we may end up using a borderless <Table> somewhere else where we, for some reason don't want to wrap and have it accidentally end up wrapping.

frontend/src/styles/_uswds-theme-custom-styles.scss
```diff
@@ -485,3 +485,10 @@ button.usa-pagination__button.usa-button {
...
+.usa-table--borderless th {
+  max-width: 0; // magic to prevent tables that are nested inside of a flex element from blowing out
+  word-wrap: break-word;
```
doug-s-nava, line 493: as a clear css charlatan, I had to look this up and had some trouble. https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap

Mind changing this to `overflow-wrap` since that seems to be the more up to date term?

doug-s-nava (commented):
great solution! just a couple nits

frontend/src/styles/_uswds-theme-custom-styles.scss
myduong-navapbc: Ah yeah, I forgot that `word-wrap` was replaced by `overflow-wrap`

frontend/src/styles/_uswds-theme-custom-styles.scss
myduong-navapbc: very valid points of concern!

doug-s-nava (approved):
✅ ✅ ✅ ✅

PR #4996, opened 2025-05-08
[Issue #4706] Migration file/model update 

### LINKED ISSUE #4706: Implement soft deletes for saved opportunities and saved searches

babebe:
## Summary

Fixes #4706 

Update models with new col: is_deleted: nullable
Migration file for user_saved_opportunity and user_saved_search tables


api/src/db/migrations/versions/2025_05_08_add_is_deleted_col_to_user_saved_tables.py
```diff
@@ -0,0 +1,36 @@
...
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        "user_saved_opportunity",
+        sa.Column("is_deleted", sa.Boolean(), nullable=True),
+        schema="api",
+    )
```
chouinar, line 25: This doesn't match what the schema says, running the tests will fail because the migrations aren't fully up to date.

You have this as nullable=True, but the SQLAlchemy model doesn't have a nullable column.

If we make that column non-nullable, you have to do some additional work to make it so that column is first populated for everything

api/src/db/models/user_models.py
```diff
@@ -113,6 +113,7 @@ class UserSavedOpportunity(ApiSchemaTable, TimestampMixin):
     opportunity: Mapped[Opportunity] = relationship(
         "Opportunity", back_populates="saved_opportunities_by_users"
     )
+    is_deleted: Mapped[bool] = mapped_column(nullable=True)
```
chouinar, line 116: Use the python typing instead, otherwise I'm not sure what the behavior is if the value was null in the DB and it tries to pull it in
```suggestion
    is_deleted: Mapped[bool | None]
```

chouinar (approved):

PR #5014, opened 2025-05-12
Keith nava/4280 improve close behavior on all simpler modals

### LINKED ISSUE #4280: Improve close behavior on all Simpler Modals
### Summary

The Truss modal component does not do these out of the box, so we should wrap their component in this functionality and use that wrapped component for all Modals in the app.

We'll also fold in the building of a abstraction over the other shared aspects of our current modals, with the goal of reducing redundancy and duplicative testing in those existing components.

### Acceptance criteria

- [ ] Simpler Modal component created that:
  - [ ] includes a close button in the upper right corner
  - [ ] closes on blur (when background is clicked)
  - [ ] includes styling / functionality in common to existing modals
    - [ ] if it makes sense to do this only for the saved search related modals and creating an abstraction just over those, do that
- [ ] All existing modals on the site are changed to use new component

KeithNava:
## Summary

Fixes #4280 

## Changes proposed

Override the custom button styling in the application for the close button on the Modal Component only.

## Context for reviewers

The original intention of this issue was to ensure the Trussworks Modal component met the requirements for the Simpler Grants Modal. The Trussworks modal does have all the features required, if we remove the `forceAction` prop, but the close button's styling in the Modal is being overridden by the custom button styling in the application. To overcome this, I added a set of styles just for the Modal's close button.

## Validation steps

image


frontend/src/styles/_uswds-theme-custom-styles.scss
```diff
@@ -27,6 +27,7 @@ i.e.
 .pdf-card {
   transition: transform 0.2s;
   @include u-shadow(2);
+
```
doug-s-nava, line 30: can you revert the formatting changes here?

doug-s-nava (commented):
the close behavior doesn't seem any different - I think you need to make the change to the modals themselves as earlier discussed

PR #5015, opened 2025-05-12
[Issue #4968] Switch authentication of apply endpoints to support user auth

### LINKED ISSUE #4968: Switch authentication of apply endpoints to support user auth
### Summary

For the below endpoints, we want to change their auth to be our `api_jwt_auth` instead of the key auth we set them up with.
* POST /applications/start
* GET /applications/:applicationId
* GET /applications/:applicationId/application_form/:appFormId
* POST /alpha/applications/:applicationId/submit

We don't need to enforce any user ownership yet, but a follow-up PR will do that.

For the GET /competitions/:competionId endpoint, we want to allow either jwt or key auth for this endpoint. This can be done with the new `jwt_or_key_multi_auth` we recently added.

### Acceptance criteria

- [ ] Update endpoints auth
- [ ] Update tests to use the new proper auth

mikehgrantsgov:
## Summary

Fixes #4968

## Changes proposed

Update several application endpoints to use `api_jwt_auth` mechanisms
Update competition endpoint ` /competitions/:competionId` to use `jwt_or_key_multi_auth`.
Update tests.

## Context for reviewers

Due to @freeze_time in the first several application test, we have to create a JWT token manually since they are date-reliant (I thought this would work but it doesn't?)

## Validation steps

Run application and competition tests, see new competition test to support multi auth.

### DIFF
--- a/api/src/api/competition_alpha/competition_route.py
+++ b/api/src/api/competition_alpha/competition_route.py
@@ -6,7 +6,7 @@ import src.adapters.db.flask_db as flask_db
 import src.api.competition_alpha.competition_schema as competition_schema
 import src.api.response as response
 from src.api.competition_alpha.competition_blueprint import competition_blueprint
-from src.auth.api_key_auth import api_key_auth
+from src.auth.multi_auth import jwt_or_key_multi_auth
 from src.logging.flask_logger import add_extra_data_to_current_request_logs
 from src.services.competition_alpha.get_competition import get_competition
 
@@ -15,7 +15,7 @@ logger = logging.getLogger(__name__)
 
 @competition_blueprint.get("/competitions/<uuid:competition_id>")
 @competition_blueprint.output(competition_schema.CompetitionResponseAlphaSchema())
-@competition_blueprint.auth_required(api_key_auth)
+@competition_blueprint.auth_required(jwt_or_key_multi_auth)
 @flask_db.with_db_session()
 def competition_get(db_session: db.Session, competition_id: uuid.UUID) -> response.ApiResponse:
     add_extra_data_to_current_request_logs({"competition.competition_id": competition_id})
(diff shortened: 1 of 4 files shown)


api/src/api/competition_alpha/competition_route.py
```diff
@@ -15,7 +15,7 @@
 
 @competition_blueprint.get("/competitions/<uuid:competition_id>")
 @competition_blueprint.output(competition_schema.CompetitionResponseAlphaSchema())
-@competition_blueprint.auth_required(api_key_auth)
+@competition_blueprint.auth_required(jwt_or_key_multi_auth)
```
chouinar, line 18: It's a bit annoying, but you'll need to specify the auth a little different: https://github.com/HHS/simpler-grants-gov/blob/main/documentation/api/authentication.md#supporting-multiple-authentication-schemes-on-an-endpoint

APIFlask doesn't work with the multi-auth, so need to define auth the more generic Flask way and then tell APIFlask about which security we have in the doc decorator.

mikehgrantsgov: Ah yeah, forgot to clean that up. Just did

api/src/auth/multi_auth.py
```diff
@@ -42,4 +43,14 @@ def get_user(self) -> MultiAuthUser:
...
 jwt_or_key_security_schemes = [api_jwt_auth.security_scheme_name, api_key_auth.security_scheme_name]
+
+
+# Helper function to format security schemes for OpenAPI
+def get_security_requirement(schemes: Sequence[str | None]) -> list[str | dict[str, list[Any]]]:
+    # Only include scheme names that are not None and cast to the expected type
+    # for APIScaffold.doc's security parameter
+    return cast(
+        list[str | dict[str, list[Any]]], [{scheme: []} for scheme in schemes if scheme is not None]
+    )
```
chouinar, line 56: Rather than needing to call this function if we define multiple endpoints that use multi-auth, could we add something after this function like:

```py
jwt_or_key_security_schemes = get_security_requirement([api_jwt_auth.security_scheme_name, api_key_auth.security_scheme_name])
```
That way the security scheme gets defined once and we just need to import the list/use that rather than import a list & function.

mikehgrantsgov: I updated the method signature since it's an internal method and changed the export

chouinar (approved):

PR #5019, opened 2025-05-12
[Issue #4862]Feature Flag default button

### LINKED ISSUE #4862: Add a reset button to the feature flags page
### Summary

Image
Currently, it shows how you have each flag set in your browser. But if you've toggled them, you can lose track of what the environment set as defaults. It'd be nice to one-click back to the env variables set on the server.

### Acceptance criteria

- [ ] Added a button to the FF page that when clicked resets all flags to the server environment default
- [ ] Added tests to validate functionality

austincanada:
## Summary

Fixes #4862 

## Changes proposed

Adds a button to change the feature flags to their default options.

## Context for reviewers

This builds on the underlying feature flag implementation. Could be changed to params if necessary.

## Validation steps
Screenshot 2025-05-12 at 4 24 02 PM
Screenshot 2025-05-12 at 4 24 23 PM


frontend/src/components/dev/FeatureFlagsTable.tsx
```diff
@@ -35,15 +39,25 @@ export default function FeatureFlagsTable() {
                 <Button
                   data-testid={`enable-${featureName}`}
                   disabled={!!enabled}
-                  onClick={() => setFeatureFlag(featureName, true)}
+                  onClick={() => {
+                    setFeatureFlag(featureName, true);
+                    router.replace(
```
doug-s-nava, line 44: what is this doing exactly? Why do we need to reset the query param here?

frontend/src/components/dev/FeatureFlagsTable.tsx
```diff
@@ -53,6 +67,15 @@ export default function FeatureFlagsTable() {
           ))}
         </tbody>
       </Table>
+      <Button
+        data-testid={"reset-defaults"}
+        onClick={() => {
+          router.replace(pathname + "?" + createQueryString("_ff", "reset"));
```
doug-s-nava, line 73: the `setQueryParam` function here should accomplish what we need, no need to create a new helper function as far as I can tell https://github.com/HHS/simpler-grants-gov/blob/c92704ac25dade39c6bee538ee73c70a0c89cdd9/frontend/src/hooks/useSearchParamUpdater.ts#L59

frontend/src/components/dev/FeatureFlagsTable.tsx
austincanada: From what I can tell if we enable/disable some features and then use the reset button, the query param of `reset` stays. I can double check if that still happens with the other `setQueryParam` function you mentioned.

doug-s-nava: I don't think that the param staying in the url is a problem for this feature

frontend/src/components/dev/FeatureFlagsTable.tsx
```diff
@@ -53,6 +59,15 @@ export default function FeatureFlagsTable() {
           ))}
         </tbody>
       </Table>
+      <Button
+        data-testid={"reset-defaults"}
+        onClick={() => {
+          setQueryParam("_ff", "reset");
```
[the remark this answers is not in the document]
austincanada: Is there a refresh hook to use? When adding a refresh through useRouter, it actually doesn't refresh any of the UI, I think because it's not updating the FeatureFlag state directly.

doug-s-nava:
closing in favor of https://github.com/HHS/simpler-grants-gov/pull/5158

Issue #4951, opened 2025-05-06
Deploy NOFOs from Github Actions


coilysiren:
### Summary

//

### Acceptance criteria

//

PR #5023, opened 2025-05-12
[Issues #4951] NOFOs CD part 3

coilysiren:
## Summary

Probably finishes https://github.com/HHS/simpler-grants-gov/issues/4951

## Changes proposed

Fixes some random bits here and there that were breaking the nofos terraform apply

## Testing

`terraform apply` gets as far as looking for the specific image tag it needs to deploy


pcraig3 (approved):
Also approve!

PR #5032, opened 2025-05-13
[Issue #4969] Associate the user with the application when starting application

### LINKED ISSUE #4969: In the start application endpoint, associate the user with the application
### Summary

TODO

### Acceptance criteria

x

mikehgrantsgov:
## Summary

Fixes #4969 

## Changes proposed

Associate users with applications during creation
Added user_id parameter to create_application (should this be optional?)
Created `ApplicationUser` records, taking duplicates into account

## Context for reviewers

Part of user auth setup for our application endpoints.

Note this will eventually be merged to main when #4968 gets merged. Just creating this to get ahead.

## Validation steps

See updated unit test

### DIFF
--- a/api/src/api/application_alpha/application_route.py
+++ b/api/src/api/application_alpha/application_route.py
@@ -39,10 +39,10 @@ def application_start(db_session: db.Session, json_data: dict) -> response.ApiRe
 
     # Get user from token session
     token_session = api_jwt_auth.get_user_token_session()
-    user_id = token_session.user_id
+    user = token_session.user
 
     with db_session.begin():
-        application = create_application(db_session, competition_id, user_id)
+        application = create_application(db_session, competition_id, user)
 
     return response.ApiResponse(
         message="Success", data={"application_id": application.application_id}
--- a/api/src/api/application_alpha/application_route.py
+++ b/api/src/api/application_alpha/application_route.py
@@ -14,7 +14,7 @@ from src.api.application_alpha.application_schemas import (
     ApplicationStartResponseSchema,
 )
 from src.api.schemas.response_schema import AbstractResponseSchema
-from src.auth.api_key_auth import api_key_auth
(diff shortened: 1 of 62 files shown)


api/src/api/application_alpha/application_route.py
```diff
@@ -37,8 +37,12 @@ def application_start(db_session: db.Session, json_data: dict) -> response.ApiRe
 
     competition_id = json_data["competition_id"]
 
+    # Get user from token session
+    token_session = api_jwt_auth.get_user_token_session()
+    user_id = token_session.user_id
+
     with db_session.begin():
-        application = create_application(db_session, competition_id)
+        application = create_application(db_session, competition_id, user_id)
```
chouinar, line 45: I'd recommend just passing the user in, we'll need the user in some future tickets.

api/src/services/applications/create_application.py
```diff
@@ -68,9 +69,41 @@ def create_application(db_session: db.Session, competition_id: UUID) -> Applicat
...
+            )
+        ).scalar_one_or_none()
+
+        # Only create the association if it doesn't already exist
+        if not existing_association:
+            application_user = ApplicationUser(
+                application_id=application.application_id, user_id=user_id
+            )
+            db_session.add(application_user)
+
```
chouinar, line 90: Since this endpoint is where we create the application (right above this section), this wouldn't ever find an existing association. We can just add an application user object.

The user_id also can't be null at the moment, so this can be simplified down to something like:
```suggestion
    application_user = ApplicationUser(application=application,user=user)
    db_session.add(application_user)
```

api/src/api/application_alpha/application_route.py
mikehgrantsgov: Makes sense, I updated this

chouinar (approved):

PR #5040, opened 2025-05-14
[Issue #4696] Enforce competition close date on application submission 

### LINKED ISSUE #4696: Enforce competition close date on application submission
# Summary
If the competition close date has been reached, error if a user tries to submit.

Note that there is a `grace_period` field in the competition table, if that column is not null, that actual close date is `close_date + grace_period days`. So if the close date were Jan 1st, and there was a grace period of 3 days, it would actually close Jan 4th.

All logic should be done in EST - use the datetime util to get the current date.

If it is after the close date + grace period, use `raise_flask_error` to raise a `422`, and add a validation issue to explain the issue.

# Acceptance criteria
- [ ] Validation behavior added around close date, including accounting for optional grace period.
- [ ] Tests added

mikehgrantsgov:
## Summary

Fixes #4696

## Changes proposed

Respect competition closing_date and grace period when submitting an application.
Tweak existing tests, add new tests

## Context for reviewers

All logic done in EST

## Validation steps

See new tests and tweaks to existing tests.

### DIFF
--- a/api/src/services/applications/submit_application.py
+++ b/api/src/services/applications/submit_application.py
@@ -1,5 +1,5 @@
 import logging
-from datetime import datetime, timedelta
+from datetime import timedelta
 from uuid import UUID
 
 import src.adapters.db as db
@@ -49,10 +49,6 @@ def submit_application(db_session: db.Session, application_id: UUID) -> Applicat
             actual_closing_date = competition.closing_date + timedelta(
                 days=competition.grace_period
             )
-
-        # Ensure we're comparing dates, not mixing date and datetime
-        if isinstance(actual_closing_date, datetime):
-            actual_closing_date = actual_closing_date.date()
 
         if current_date > actual_closing_date:
             message = "Cannot submit application - competition is closed"
--- a/api/src/services/applications/submit_application.py
+++ b/api/src/services/applications/submit_application.py
@@ -1,4 +1,5 @@
 import logging
+from datetime import datetime, timedelta
 from uuid import UUID
 
 import src.adapters.db as db
@@ -7,6 +8,7 @@ from src.api.route_utils import raise_flask_error
 from src.constants.lookup_constants import ApplicationStatus
 from src.db.models.competition_models import Application
 from src.services.applications.get_application import get_application
+from src.util.datetime_util import get_now_us_eastern_date
 from src.validation.validation_constants import ValidationErrorType
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


api/src/services/applications/submit_application.py
```diff
@@ -35,6 +37,45 @@ def submit_application(db_session: db.Session, application_id: UUID) -> Applicat
...
+        # If grace_period is not null, add that many days to the closing date
+        if competition.grace_period is not None and competition.grace_period > 0:
+            actual_closing_date = competition.closing_date + timedelta(
+                days=competition.grace_period
+            )
+
+        # Ensure we're comparing dates, not mixing date and datetime
+        if isinstance(actual_closing_date, datetime):
+            actual_closing_date = actual_closing_date.date()
+
```
chouinar, line 56: What scenario would the closing date be a datetime? I don't think we need to account for that.

api/src/services/applications/submit_application.py
```diff
@@ -35,6 +37,45 @@ def submit_application(db_session: db.Session, application_id: UUID) -> Applicat
             ],
         )
 
+    # Check if competition is closed
```
chouinar, line 40: As we're going to continue adding more and more validation rules, could we put each of these different checks into a function that we call here?

Then this submit function can be something like:

```
    logger.info("Processing application submit")
    application = get_application(db_session, application_id)

    validate_application_in_progress(application)
    validate_competition_open(application)
    # ... many many more we'll add soon

    application.application_status = ApplicationStatus.SUBMITTED
    logger.info("Application successfully submitted")

```
Also could then make tests that call those specific functions if we don't want to fully setup an API call test for each scenario we add/test.

api/src/services/applications/submit_application.py
mikehgrantsgov: I hate datetimes in my tests but changed them all to dates, so backed out this change

api/tests/src/services/applications/test_submit_application.py
```diff
@@ -1,62 +1,144 @@
...
+    )
+    application = ApplicationFactory.create(
+        application_status=ApplicationStatus.IN_PROGRESS, competition=competition
+    )
+    db_session.commit()  # Commit initial state
+
+    updated_application = submit_application(db_session, application.application_id)
+    db_session.commit()  # Commit the change made by the service
+
+    db_session.refresh(updated_application)  # Refresh to get the latest state from DB
```
chouinar, line 138: Factories already commit as part of create. Any case of `SomeFactory.create(...)` followed by `db_session.commit()` isn't necessary.

What is refreshing from the DB doing here when the function returns the DB object?

```suggestion
    application = ApplicationFactory.create(
        application_status=ApplicationStatus.IN_PROGRESS, competition=competition
    )

    updated_application = submit_application(db_session, application.application_id)
```

api/tests/src/services/applications/test_submit_application.py
```diff
@@ -1,62 +1,144 @@
...
+        == ValidationErrorType.COMPETITION_ALREADY_CLOSED
+    )
+
+
+@freeze_time(TEST_DATE)
+def test_validate_competition_with_no_closing_date(enable_factory_create, db_session):
+    """Test that validating a competition with no closing date succeeds."""
+    competition = CompetitionFactory.create(closing_date=None)
+    application = ApplicationFactory.create(competition=competition)
+    db_session.commit()
```
chouinar, line 116:
```suggestion

```

api/tests/src/services/applications/test_submit_application.py
```diff
@@ -1,62 +1,144 @@
...
+    current_date = date.fromisoformat(TEST_DATE)
+
+    # Create a competition with a past closing date (5 days ago) and a grace period of 3 days
+    # This means the competition has been closed for 2 days
+    competition = CompetitionFactory.create(
+        closing_date=current_date - timedelta(days=5),
+        grace_period=3,
+    )
+    application = ApplicationFactory.create(competition=competition)
+    db_session.commit()
```
chouinar, line 98:
```suggestion

```

api/tests/src/services/applications/test_submit_application.py
```diff
@@ -1,62 +1,144 @@
...
+    # Should not raise any exception
+    validate_competition_open(application1)
+
+    # Past closing date but within grace period
+    competition2 = CompetitionFactory.create(
+        closing_date=current_date - timedelta(days=2),
+        grace_period=3,
+    )
+    application2 = ApplicationFactory.create(competition=competition2)
+    db_session.commit()
```
chouinar, line 80:
```suggestion

    # Should not raise any exception
    validate_competition_open(application1)

    # Past closing date but within grace period
    competition2 = CompetitionFactory.create(
        closing_date=current_date - timedelta(days=2),
        grace_period=3,
    )
    application2 = ApplicationFactory.create(competition=competition2)
```

api/tests/src/services/applications/test_submit_application.py
mikehgrantsgov: Yeah, I thought those references were removed. I was debugging another issue in the code that may have been related to a record not saving in the DB, and added a commit to debug that wasn't properly removed

PR #5044, opened 2025-05-14
[no ticket] Adds tags for cost allocation

coilysiren:
## Motivation

We want to be able to slice our costs in AWS by application and by team (Nava vs Bloomworks)


mdragon (approved):
Approved

PR #5045, opened 2025-05-14
[no ticket] Scales nofos down to minimums

coilysiren:
## Context

NOFOs doesn't need to amount of resources that grants.gov application does


mdragon (approved):
Approved

PR #5047, opened 2025-05-14
fix whitespace bug by adding max height to filter content

doug-s-nava:
## Summary

Fixes #ISSUE

### DIFF
--- a/frontend/src/components/search/SearchFilterAccordion/SearchFilterAccordion.tsx
+++ b/frontend/src/components/search/SearchFilterAccordion/SearchFilterAccordion.tsx
@@ -1,5 +1,6 @@
 "use client";
 
+import clsx from "clsx";
 import { useSearchParamUpdater } from "src/hooks/useSearchParamUpdater";
 import { QueryContext } from "src/services/search/QueryProvider";
 import {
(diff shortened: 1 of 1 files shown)


frontend/src/components/search/SearchFilterAccordion/SearchFilterAccordion.tsx
```diff
@@ -86,7 +84,11 @@ const AccordionContent = ({
 
   return (
     <>
-      <ul className="usa-list usa-list--unstyled">
+      <ul
+        className={clsx("usa-list usa-list--unstyled", {
+          "maxh-mobile-lg": wrapForScroll,
```
andycochran, line 89: Not sure what this is trying to do. 

Is the intent to get rid of the horizontal scrollbar at the bottom of the filter? 

image

That might be achieved by changing `overflow-scroll` to `overflow-x-hidden` on the containing `div`, not the list itself. 

image

image

andycochran: But I'm wondering if the max height and scroll should go on the `.usa-accordion__content`, which has padding that's currently making the scrollable area smaller. The scroll bar could be flush with the accordion's border.

doug-s-nava:
see https://github.com/HHS/simpler-grants-gov/pull/5050

PR #5048, opened 2025-05-14
[Issue #4924] Adjust API queries for opps and users to select less

### LINKED ISSUE #4924: Adjust queries to fetch opportunity in GET /opportunity to exclude relationships that aren't used in the response
### Summary

Our query to fetch an opportunity looks like (simplified):

```py
select(Opportunity).where(...).options(selectinload("*").options(noload(Opportunity.all_opportunity_summaries)
```
The `selectinload` makes it so relationships will be loaded as part of the query instead of lazy-loaded which improves performance. However, we tell it to load **all** relationships except those we add a `noload` for. This means that even relationships that aren't relevant to the GET /opportunity endpoint are loaded. We've added a lot of relationships from the opportunity table recently, so want to exclude the following:
* `opportunity_change_audit`
* `all_opportunity_notification_logs`
* `saved_opportunities_by_users`
* `competitions` - we might undo this later, but nothing uses it right now
* `versions`

As far as testing this goes, we should just make sure if an opportunity has these components in the DB, it can still be fetched uneventfully. If we were to put something in `noload` that is actually used to create the JSON response, it would error.

This is largely a performance improvement, we likely won't see much gain right now, but if an opportunity in the future has 1000 users that have saved it, we wouldn't want to fetch that.

### Acceptance criteria

- [ ] Loading adjusted
- [ ] Tests verify nothing has broken

chouinar:
## Summary

Fixes #4924

## Changes proposed
Adjust queries in the GET /opportunity endpoint and user auth logic to not automatically load as many relationships

## Context for reviewers
https://docs.sqlalchemy.org/en/20/orm/queryguide/relationships.html

SQLAlchemy relationships allow you to configure how they load. By default, all relationships are lazy loaded which means that if you do `opportunity.current_opportunity_summary` it will emit a select statement to the DB to fetch the current opportunity summary record unless it already has fetched it. There are several loading strategies that you can set (see the above link), but besides lazy loading, `selectinload` tends to work the best. It works by emitting several select statements alongside your primary select statement all in one query to improve performance.

When we do a `selectinload("*")` it says to load all relationships recursively. This is fine-ish, but the opportunity table especially is connected to everything. This means that before this change, the GET /opportunity endpoint actually runs roughly this set of queries (not even the full set because I didn't have any user records setup):

<details>
<summary>Click me to see the full query</summary>
<br>

```sql
SELECT * FROM api.opportunity 
WHERE api.opportunity.opportunity_id = %(opportunity_id_1)s::BIGINT AND api.opportunity.is_draft IS false

SELECT * FROM api.user_saved_opportunity 
WHERE api.user_saved_opportunity.opportunity_id IN (%(primary_keys_1)s::BIGINT)

SELECT * FROM api.opportunity AS opportunity_1 JOIN api.agency ON opportunity_1.agency_code = api.agency.agency_code 
WHERE opportunity_1.opportunity_id IN (%(primary_keys_1)s::BIGINT)

SELECT * FROM api.competition 
WHERE api.competition.opportunity_id IN (%(primary_keys_1)s::BIGINT)

SELECT * FROM api.opportunity_version 
WHERE api.opportunity_version.opportunity_id IN (%(primary_keys_1)s::BIGINT)

SELECT * FROM api.opportunity_attachment 
WHERE api.opportunity_attachment.opportunity_id IN (%(primary_keys_1)s::BIGINT)

SELECT * FROM api.opportunity_assistance_listing 
WHERE api.opportunity_assistance_listing.opportunity_id IN (%(primary_keys_1)s::BIGINT)

SELECT * FROM api.opportunity_change_audit 
WHERE api.opportunity_change_audit.opportunity_id IN (%(primary_keys_1)s::BIGINT)

SELECT * FROM api.current_opportunity_summary 
WHERE api.current_opportunity_summary.opportunity_id IN (%(primary_keys_1)s::BIGINT)

SELECT * FROM api.opportunity_summary 
WHERE api.opportunity_summary.opportunity_id IN (%(primary_keys_1)s::BIGINT)

SELECT * FROM api.user_opportunity_notification_log 
WHERE api.user_opportunity_notification_log.opportunity_id IN (%(primary_keys_1)s::BIGINT)

SELECT * FROM api.opportunity_summary 
WHERE api.opportunity_summary.opportunity_summary_id IN (%(primary_keys_1)s::BIGINT)

SELECT * FROM api.current_opportunity_summary 
WHERE api.current_opportunity_summary.opportunity_summary_id IN (%(primary_keys_1)s::BIGINT)

SELECT * FROM api.link_opportunity_summary_funding_instrument 
WHERE api.link_opportunity_summary_funding_instrument.opportunity_summary_id IN (%(primary_keys_1)s::BIGINT)

SELECT * FROM api.link_opportunity_summary_funding_category 
WHERE api.link_opportunity_summary_funding_category.opportunity_summary_id IN (%(primary_keys_1)s::BIGINT)

SELECT * FROM api.link_opportunity_summary_applicant_type 
WHERE api.link_opportunity_summary_applicant_type.opportunity_summary_id IN (%(primary_keys_1)s::BIGINT)

SELECT * FROM api.link_opportunity_summary_funding_instrument 
WHERE api.link_opportunity_summary_funding_instrument.opportunity_summary_id IN (%(primary_keys_1)s::BIGINT)

SELECT * FROM api.link_opportunity_summary_funding_category 
WHERE api.link_opportunity_summary_funding_category.opportunity_summary_id IN (%(primary_keys_1)s::BIGINT)

SELECT * FROM api.link_opportunity_summary_applicant_type 
WHERE api.link_opportunity_summary_applicant_type.opportunity_summary_id IN (%(primary_keys_1)s::BIGINT) 
```
</details>

If you notice, this contains tables like the saved opportunities table which we don't return in the API endpoint itself. As more folks saved opportunities, this means we're fetching more and more.

The change made here changes several relationships back to `lazyload` (which is the default - so effectively removes it from the "*" list). This changes the queries to roughly:

<details>
<summary>See after query</summary>
<br>

```sql
SELECT * FROM api.opportunity 
WHERE api.opportunity.opportunity_id = %(opportunity_id_1)s::BIGINT AND api.opportunity.is_draft IS false

SELECT * FROM api.opportunity AS opportunity_1 JOIN api.agency ON opportunity_1.agency_code = api.agency.agency_code 
WHERE opportunity_1.opportunity_id IN (%(primary_keys_1)s::BIGINT)

SELECT * FROM api.opportunity_attachment 
WHERE api.opportunity_attachment.opportunity_id IN (%(primary_keys_1)s::BIGINT)

SELECT * FROM api.opportunity_assistance_listing 
WHERE api.opportunity_assistance_listing.opportunity_id IN (%(primary_keys_1)s::BIGINT)

SELECT * FROM api.current_opportunity_summary 
WHERE api.current_opportunity_summary.opportunity_id IN (%(primary_keys_1)s::BIGINT)

SELECT * FROM api.opportunity_summary 
WHERE api.opportunity_summary.opportunity_summary_id IN (%(primary_keys_1)s::BIGINT)

SELECT * FROM api.link_opportunity_summary_funding_instrument 
WHERE api.link_opportunity_summary_funding_instrument.opportunity_summary_id IN (%(primary_keys_1)s::BIGINT)

SELECT * FROM api.link_opportunity_summary_funding_category 
WHERE api.link_opportunity_summary_funding_category.opportunity_summary_id IN (%(primary_keys_1)s::BIGINT)

SELECT * FROM api.link_opportunity_summary_applicant_type 
WHERE api.link_opportunity_summary_applicant_type.opportunity_summary_id IN (%(primary_keys_1)s::BIGINT)
```
</details>

Additionally, I updated two auth related queries to not use `*` and instead specify what we want to fetch. This is for a similar reason, but the fix is simpler. Both fetch a row connected to a user, and want the user, but having `*` meant that user which has saved opportunities then fetched all opportunities, and saved searches, and more. That's way, way too much. The bit of logic for each of them just needs the `user` connection.

## Validation steps
If you ever want to see the SQLAlchemy queries emitted, you can turn `echo=True` on when we call `sqlalchemy.create_engine` in the postgres_client.py file. Warning, the info it emits is way, way more than shown above, I trimmed these massively for readability.

### DIFF
--- a/api/src/auth/api_jwt_auth.py
+++ b/api/src/auth/api_jwt_auth.py
@@ -153,7 +153,7 @@ def parse_jwt_for_user(
     token_session: UserTokenSession | None = db_session.execute(
         select(UserTokenSession)
         .where(UserTokenSession.token_id == sub_id)
-        .options(selectinload("*"))
+        .options(selectinload(UserTokenSession.user))
     ).scalar()
 
     # We check both the token expires_at timestamp as well as an
--- a/api/src/services/opportunities_v1/get_opportunity.py
+++ b/api/src/services/opportunities_v1/get_opportunity.py
@@ -1,5 +1,5 @@
 from sqlalchemy import select
-from sqlalchemy.orm import noload, selectinload
+from sqlalchemy.orm import lazyload, selectinload
 
 import src.adapters.db as db
 from src.adapters.aws import S3Config
@@ -15,9 +15,7 @@ class AttachmentConfig(PydanticBaseEnvConfig):
     cdn_url: str | None = None
 
 
-def _fetch_opportunity(
-    db_session: db.Session, opportunity_id: int, load_all_opportunity_summaries: bool
-) -> Opportunity:
+def _fetch_opportunity(db_session: db.Session, opportunity_id: int) -> Opportunity:
     stmt = (
         select(Opportunity)
         .where(Opportunity.opportunity_id == opportunity_id)
@@ -27,10 +25,20 @@ def _fetch_opportunity(
         # we need to explicitly join here as the "*" approach doesn't
         # seem to work with the way our agency relationships are setup
         .options(selectinload(Opportunity.agency_record).selectinload(Agency.top_level_agency))
+        # Do not load the following relationships, they aren't necessary for
+        # our opportunity endpoints, and would make the query much larger/slower
+        # if we were to fetch them.
+        # This effectively undoes the `selectinload("*")` above for these relationships
+        # and makes them lazily loaded (the default for relationships) - keeping them out of the query entirely.
+        .options(
+            lazyload(Opportunity.opportunity_change_audit),
+            lazyload(Opportunity.all_opportunity_summaries),
+            lazyload(Opportunity.all_opportunity_notification_logs),
+            lazyload(Opportunity.saved_opportunities_by_users),
+            lazyload(Opportunity.competitions),
+            lazyload(Opportunity.versions),
+        )
     )
-
-    if not load_all_opportunity_summaries:
-        stmt = stmt.options(noload(Opportunity.all_opportunity_summaries))
 
     opportunity = db_session.execute(stmt).unique().scalar_one_or_none()
 
@@ -50,9 +58,7 @@ def pre_sign_opportunity_file_location(
 
 
 def get_opportunity(db_session: db.Session, opportunity_id: int) -> Opportunity:
-    opportunity = _fetch_opportunity(
-        db_session, opportunity_id, load_all_opportunity_summaries=False
-    )
+    opportunity = _fetch_opportunity(db_session, opportunity_id)
 
     attachment_config = AttachmentConfig()
     if attachment_config.cdn_url is not None:
--- a/api/src/adapters/db/clients/postgres_client.py
+++ b/api/src/adapters/db/clients/postgres_client.py
@@ -48,7 +48,6 @@ class PostgresDBClient(DBClient):
             pool=conn_pool,
             hide_parameters=db_config.hide_sql_parameter_logs,
             execution_options={"schema_translate_map": db_config.get_schema_translate_map()},
-            echo=True,
             # TODO: Don't think we need this as we aren't using JSON columns, but keeping for reference
             # json_serializer=lambda o: json.dumps(o, default=pydantic.json.pydantic_encoder),
         )
--- a/api/src/adapters/db/clients/postgres_client.py
+++ b/api/src/adapters/db/clients/postgres_client.py
@@ -48,6 +48,7 @@ class PostgresDBClient(DBClient):
             pool=conn_pool,
             hide_parameters=db_config.hide_sql_parameter_logs,
             execution_options={"schema_translate_map": db_config.get_schema_translate_map()},
+            echo=True,
             # TODO: Don't think we need this as we aren't using JSON columns, but keeping for reference
             # json_serializer=lambda o: json.dumps(o, default=pydantic.json.pydantic_encoder),
         )
--- a/api/src/services/users/login_gov_callback_handler.py
+++ b/api/src/services/users/login_gov_callback_handler.py
@@ -141,7 +141,7 @@ def _process_token(db_session: db.Session, token: str, nonce: str) -> LoginGovCa
         # We only support login.gov right now, so this does nothing, but let's
         # be explicit just in case.
         .where(LinkExternalUser.external_user_type == ExternalUserType.LOGIN_GOV)
-        .options(selectinload("*"))
+        .options(selectinload(LinkExternalUser.user))
     ).scalar()
 
     is_user_new = external_user is None


api/src/services/opportunities_v1/get_opportunity.py
```diff
@@ -27,11 +25,21 @@ def _fetch_opportunity(
...
+            lazyload(Opportunity.all_opportunity_summaries),
+            lazyload(Opportunity.all_opportunity_notification_logs),
+            lazyload(Opportunity.saved_opportunities_by_users),
+            lazyload(Opportunity.competitions),
+            lazyload(Opportunity.versions),
+        )
     )
 
-    if not load_all_opportunity_summaries:
-        stmt = stmt.options(noload(Opportunity.all_opportunity_summaries))
```
chouinar, line 33: FYI - I got rid of `noload` because the docs recommend never using it - it exists for legacy reasons. It makes it so the value is forced to be `None` and has weird behaviors. It's better to just change set things to `lazyload` - worst case there is we try to fetch something that has to be lazyloaded.

freddieyebra (approved):
looks good!

api/src/auth/api_jwt_auth.py
```diff
@@ -153,7 +153,7 @@ def parse_jwt_for_user(
     token_session: UserTokenSession | None = db_session.execute(
         select(UserTokenSession)
         .where(UserTokenSession.token_id == sub_id)
-        .options(selectinload("*"))
+        .options(selectinload(UserTokenSession.user))
```
joshtonava, line 156: Reiterating my own understanding, this is going to preload the fields from the user table

api/src/services/opportunities_v1/get_opportunity.py
```diff
@@ -27,11 +25,21 @@ def _fetch_opportunity(
...
+        # if we were to fetch them.
+        # This effectively undoes the `selectinload("*")` above for these relationships
+        # and makes them lazily loaded (the default for relationships) - keeping them out of the query entirely.
+        .options(
+            lazyload(Opportunity.opportunity_change_audit),
+            lazyload(Opportunity.all_opportunity_summaries),
+            lazyload(Opportunity.all_opportunity_notification_logs),
+            lazyload(Opportunity.saved_opportunities_by_users),
+            lazyload(Opportunity.competitions),
+            lazyload(Opportunity.versions),
```
joshtonava, line 39: Do we need to think about specifying the load behavior of the remaining relationships on the Opportunity table not mentioned here?

- current_opportunity_summary: Mapped["CurrentOpportunitySummary | None"]
- opportunity_attachments: Mapped[list["OpportunityAttachment"]]
- opportunity_assistance_listings: Mapped[list["OpportunityAssistanceListing"]]
- agency_record: Mapped[Agency | None]

joshtonava (approved):
Makes sense to me, I just had a small question

api/src/services/opportunities_v1/get_opportunity.py
chouinar: Those are the ones we want to load, and are picked up by the `*` above - I originally looked at flipping this logic, and only setting the ones we want to load, but that's surprisingly more verbose. The current opportunity summary links to the opportunity summary which links to several other relationships - that's a lot more we have to end up defining, so I flipped it to turn off the relationships we don't want to fetch.

api/src/auth/api_jwt_auth.py
chouinar: Yeah. Without this, when we do `token_session.user` a few lines later, it would emit a separate select statement to the DB to fetch it. This just makes it so they're both fetched up front.

I changed this from the current `*` because otherwise it was fetching a lot. Just think about these relationships - the user has a list of saved opportunities, opportunities have a lot of relationships including other saved opportunity records and other users.

I wouldn't be surprised if we see some faster API calls after this change with it currently fetching a whole bunch of stuff.

api/src/services/opportunities_v1/get_opportunity.py
joshtonava: Thanks for the explanation I see now what you are doing with the lazyload

PR #5049, opened 2025-05-14
[Issue #4982] Add application_name to application table

### LINKED ISSUE #4982: Add application_name field to the application table
### Summary

We want to add a new column named `application_name` to the application table - we'll likely make it non-null in the future, but existing data needs to be cleared first.

### Acceptance criteria

- [ ] Column added

mikehgrantsgov:
## Summary

Fixes #4982

## Changes proposed

Add non-nullable application_name column to application table. 

## Context for reviewers

We'll likely make it non-null in the future, but existing data needs to be cleared first.

## Validation steps

See migration


api/src/db/models/competition_models.py
```diff
@@ -150,7 +150,7 @@ class Application(ApiSchemaTable, TimestampMixin):
         ForeignKey(LkApplicationStatus.application_status_id),
     )
 
-    application_name: Mapped[str | None]
+    application_name: Mapped[str]
```
chouinar, line 153: There are some applications already in the staging environment, so making this non-null would require changes - also would have a lot of broken tests right now if nothing is setting this.

I'm fine with it being nullable for now, I think there are a few fields we'll have to circle back on later and fix that, but we can do that then.

mikehgrantsgov: I flipped them over to nullable for now

chouinar (approved):

PR #5053, opened 2025-05-15
[Issue #4726] Agency filter by status

### LINKED ISSUE #4726: Update API to include all agencies with past opportunities when "Closed" or "Archived" status is selected
### Summary:
Currently, the agency filter only shows agencies with **Open** or **Forecasted** opportunities, so agencies with only **Closed** or **Archived** opportunities will not be shown in the filter list

### Acceptance Criteria:
- [ ] Update the search endpoint to allow for different searches based on status
- [ ] Ensure agencies make it into the index

babebe:
## Summary

Fixes #4726 

## Changes proposed

Update the load agency task to index a boolean flag on the agency for each associated opportunity status.
Update the search agency route to support filtering by multiple opportunity statuses.
Update the get agency route to include filtering based on opportunity status.


api/src/search/backend/load_agencies_to_index.py
```diff
@@ -72,25 +73,39 @@ def fetch_agencies(self) -> Sequence[Agency]:
...
                 extra={"agency_id": agency.agency_id, "agency_code": agency.agency_code},
             )
+
             agency_json = SCHEMA.dump(agency)
-            agency_json["has_active_opportunity"] = agency.agency_id in active_agencies
+
+            agency_json["has_open_opportunity"] = agency.agency_id in posted_agencies
+            agency_json["has_forecasted_opportunity"] = agency.agency_id in forecasted_agencies
+            agency_json["has_closed_opportunity"] = agency.agency_id in closed_agencies
+            agency_json["has_archived_opportunity"] = agency.agency_id in archived_agencies
```
chouinar, line 108: I know this would take a bit to change from what you've done, but what if instead of 4 separate booleans, the dataset was a list of statuses? Instead we have effectively `"opportunity_statuses": ["posted", "archived"]`? Instead of passing in 4 separate booleans when doing the search request, you pass in the statuses.

I mostly ask because based on recent-ish discussions, we might want to completely change the opportunity statuses, if we add/remove one, we'd have to make a change in all of the same places. If we just change the values, then that's a much smaller change.

This also already has support for the search queries, that's effectively the same "overlap" sort of query we do for things like applicant type.

api/src/api/agencies_v1/agency_schema.py
```diff
@@ -23,9 +24,9 @@ class AgencyListRequestSchema(Schema):
 
 
 class AgencySearchFilterV1Schema(Schema):
-    has_active_opportunity = fields.Nested(
-        BoolSearchSchemaBuilder("HasActiveOpportunityFilterV1Schema")
-        .with_one_of(example=True)
+    opportunity_statuses = fields.Nested(
+        StrSearchSchemaBuilder("OpportunityStatusesFilterV1Schema")
+        .with_one_of(example="archived")
```
chouinar, line 29: Just be careful with making this change if it breaks the frontend.

Might temporarily need to leave that field as-is, and just treat it as if a user had supplied `["posted", "forecasted"]`

api/src/search/backend/load_agencies_to_index.py
```diff
@@ -72,25 +77,55 @@ def fetch_agencies(self) -> Sequence[Agency]:
...
+                opportunity_statuses.append(OpportunityStatus.POSTED.value)
+
+            if agency.agency_id in forecasted_agencies:
+                opportunity_statuses.append(OpportunityStatus.FORECASTED.value)
+
+            if agency.agency_id in closed_agencies:
+                opportunity_statuses.append(OpportunityStatus.CLOSED.value)
+
+            if agency.agency_id in archived_agencies:
+                opportunity_statuses.append(OpportunityStatus.ARCHIVED.value)
```
chouinar, line 121: OpportunityStatus is a string and an enum, should be able to exclude value
```suggestion
            if agency.agency_id in posted_agencies:
                opportunity_statuses.append(OpportunityStatus.POSTED)

            if agency.agency_id in forecasted_agencies:
                opportunity_statuses.append(OpportunityStatus.FORECASTED)

            if agency.agency_id in closed_agencies:
                opportunity_statuses.append(OpportunityStatus.CLOSED)

            if agency.agency_id in archived_agencies:
                opportunity_statuses.append(OpportunityStatus.ARCHIVED)
```

api/tests/conftest.py
```diff
@@ -257,7 +257,9 @@ def agency_index(search_client, monkeypatch_session):
     # with an actual one, similar to how we create schemas for database tests
     index_name = f"test-agency-index-{uuid.uuid4().int}"
 
-    search_client.create_index(index_name)
+    search_client.create_index(
+        index_name, mappings={"properties": {"opportunity_statuses": {"type": "keyword"}}}
+    )
```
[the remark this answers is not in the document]
babebe: Yes, you can do one or the other. This seemed better, can be expanded on if we have other fields we want to define.

api/Makefile
```diff
@@ -113,10 +113,9 @@ volume-recreate: clean-volumes init ## Destroy current volumes, setup new ones -
 stop:
 	docker compose down
 
-check: format-check lint db-check-migrations test
-
-remake-backend: volume-recreate db-seed-local populate-search-opportunities ## Completely recreate API services, load data into the DB and search index
+check: format-check lint db-check-migratons test
```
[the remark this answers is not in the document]
babebe: I did actually. I saw the typo.

api/src/api/agencies_v1/agency_schema.py
babebe: Updated

api/src/services/agencies_v1/search_agencies.py
```diff
@@ -67,7 +69,22 @@ def get_search_request(params: AgencySearchParams) -> dict:
...
+        ):
+            value = params.filters.has_active_opportunity.one_of
+            if True in value:
+                params.filters.opportunity_statuses = StrSearchFilter(
+                    one_of=[OpportunityStatus.POSTED, OpportunityStatus.FORECASTED]
+                )
+            elif False in value:
+                params.filters.opportunity_statuses = StrSearchFilter(
+                    one_of=[OpportunityStatus.CLOSED, OpportunityStatus.ARCHIVED]
+                )
```
chouinar, line 86: I think to keep the same behavior as before, we'd just set nothing for the False and None cases? Don't know if it matters much or if the frontend would ever call with that value to matter.

api/src/services/agencies_v1/get_agencies.py
```diff
@@ -53,16 +55,39 @@ def get_agencies(
...
+                _construct_active_inner_query(
+                    Agency.agency_id, [OpportunityStatus.CLOSED, OpportunityStatus.ARCHIVED]
+                )
+                .union(
+                    _construct_active_inner_query(
+                        Agency.top_level_agency_id,
+                        [OpportunityStatus.CLOSED, OpportunityStatus.ARCHIVED],
+                    )
+                )
+                .subquery()
```
chouinar, line 85: Same as elsewhere, we might just want to treat False as "no filter" to keep the behavior roughly the same. Won't matter much as it's just while the frontend swaps its usage.

chouinar (commented):
One minor set of suggestions, otherwise looks good.

Change feels like one we should merge after deploy tomorrow in case it breaks anything / give the frontend a chance to swap over if there is a ticket queued up for that.

api/src/services/agencies_v1/search_agencies.py
babebe: Updated.

chouinar (approved):

Issue #3920, opened 2025-02-18
Implementation: Email notification content for close date matches content design


mdragon:
### Summary

Update the content of the emails being sent for opportunities closing soon

### Acceptance criteria

- [x] Email content matches the content provided [here](https://docs.google.com/document/d/19soxXfps4VFcNUsvdo5xH_qctPTx5j5EpBJpysIq7vY/edit?usp=sharing)

PR #5061, opened 2025-05-15
[Issue #3920] Email notification content for new matches to a saved search content

### LINKED ISSUE #4059: Update "New matches" Email Content after opp-status changes available
### Summary

Add to your bookmarked opportunities to be notified by email when the status or details are updated.

#3602 Content for the email draft about new opportunity matches is available now. However, we are striking out copy referring to the ability to be notified about future status updates about an opportunity until that functionality is available to us, a few sprints from now. 

This ticket is to update the copy when that functionality is ready.

Current:
`Submission period: [MM/DD/YYYY–MM/DD/YYYY] [To be announced.]`

New: 
`Submission period: [MM/DD/YYYY–MM/DD/YYYY] [To be announced. Add to your bookmarked opportunities to be notified by email when the status or details are updated.]`

### Acceptance criteria

- [ ] Update copy as described.

joshtonava:
## Summary

Fixes #4059

## Changes proposed

Content on the Email for saved search notifications was improved.
Added tests to show the email output.

## Context for reviewers

These are example emails generated from inspecting the tests.

### Singular

**Subject**
```
New Grant Published on 5/15/2025
```

**Body**
```
A funding opportunity matching your saved search query was recently published.

2025 Port Infrastructure Development Program

Status: Posted
Submission period: 1/31/2025–4/30/2025
Award range: $1,000,000-$112,500,000
Expected awards: 40
Cost sharing: Yes

To unsubscribe from email notifications for this query, delete it from your saved search queries.
```

### Multiple
**Subject**
```
2 New Grants Published on 5/15/2025
```

**Body**
```
The following funding opportunities matching your saved search queries were recently published.

2025 Port Infrastructure Development Program

Status: Posted
Submission period: 1/31/2025–4/30/2025
Award range: $1,000,000-$112,500,000
Expected awards: 40
Cost sharing: Yes

Cooperative Agreement for affiliated Partner with Rocky Mountains Cooperative Ecosystem Studies Unit (CESU)

Status: Forecasted
Submission period: To be announced.
Award range: $1-$30,000
Expected awards: --
Cost sharing: No

To unsubscribe from email notifications for a query, delete it from your saved search queries.
```

## Validation steps

To run relevant tests:
```bash
make test args="-vv tests/src/task/notifications/test_search_notification.py
```

### DIFF
--- a/api/src/task/notifications/search_notification.py
+++ b/api/src/task/notifications/search_notification.py
@@ -79,13 +79,44 @@ class SearchNotificationTask(BaseNotificationTask):
                 "Created changed search email notification",
                 extra={"user_id": user_id, "changed_search_count": len(saved_items)},
             )
-            message = self._build_notification_message(saved_items)
 
+            all_opportunity_ids = set()
+            for saved_search in saved_items:
+                all_opportunity_ids.update(saved_search.searched_opportunity_ids)
+            stmt = (
+                select(Opportunity)
+                .join(
+                    CurrentOpportunitySummary,
+                    CurrentOpportunitySummary.opportunity_id == Opportunity.opportunity_id,
+                )
+                .join(
+                    OpportunitySummary,
+                    OpportunitySummary.opportunity_summary_id
+                    == CurrentOpportunitySummary.opportunity_summary_id,
+                )
+                .where(Opportunity.opportunity_id.in_(all_opportunity_ids))
+                .options(
+                    selectinload(Opportunity.current_opportunity_summary).selectinload(
+                        CurrentOpportunitySummary.opportunity_summary
+                    )
+                )
+            )
+
+            opportunities = self.db_session.execute(stmt).scalars().all()
+
+            message = self._build_notification_message(opportunities)
+
+            formated_date = datetime_util.utcnow().strftime("%-m/%-d/%Y")
+            subject = (
+                f"New Grant Published on {formated_date}"
+                if len(opportunities) == 1
+                else f"{len(opportunities)} New Grants Published on {formated_date}"
+            )
             users_email_notifications.append(
                 UserEmailNotification(
                     user_id=user_id,
                     user_email=user_email,
-                    subject="Updates to Your Saved Opportunities",
+                    subject=subject,
                     content=message,
                     notification_reason=NotificationReason.SEARCH_UPDATES,
                     notified_object_ids=[
@@ -108,32 +139,7 @@ class SearchNotificationTask(BaseNotificationTask):
 
         return users_email_notifications
 
-    def _build_notification_message(self, user_notifications: list[UserSavedSearch]) -> str:
-        # Get the unique opportunity IDs across all saved searches
-        all_opportunity_ids = set()
-        for saved_search in user_notifications:
-            all_opportunity_ids.update(saved_search.searched_opportunity_ids)
-
-        stmt = (
-            select(Opportunity)
-            .join(
-                CurrentOpportunitySummary,
-                CurrentOpportunitySummary.opportunity_id == Opportunity.opportunity_id,
-            )
-            .join(
-                OpportunitySummary,
-                OpportunitySummary.opportunity_summary_id
-                == CurrentOpportunitySummary.opportunity_summary_id,
-            )
-            .where(Opportunity.opportunity_id.in_(all_opportunity_ids))
-            .options(
-                selectinload(Opportunity.current_opportunity_summary).selectinload(
-                    CurrentOpportunitySummary.opportunity_summary
-                )
-            )
-        )
-
-        opportunities = self.db_session.execute(stmt).scalars().all()
+    def _build_notification_message(self, opportunities: list[Opportunity]) -> str:
 
         # Build message intro based on number of opportunities
         if len(opportunities) == 1:
--- a/api/src/task/notifications/search_notification.py
+++ b/api/src/task/notifications/search_notification.py
@@ -6,6 +6,11 @@ from sqlalchemy.orm import selectinload
 
 import src.adapters.db as db
 import src.adapters.search as search
+from src.db.models.opportunity_models import (
+    CurrentOpportunitySummary,
+    Opportunity,
+    OpportunitySummary,
+)
 from src.db.models.user_models import UserSavedSearch
 from src.services.opportunities_v1.search_opportunities import search_opportunities_id
 from src.task.notifications.base_notification import BaseNotificationTask
@@ -74,12 +79,14 @@ class SearchNotificationTask(BaseNotificationTask):
                 "Created changed search email notification",
                 extra={"user_id": user_id, "changed_search_count": len(saved_items)},
             )
+            message = self._build_notification_message(saved_items)
+
             users_email_notifications.append(
                 UserEmailNotification(
                     user_id=user_id,
                     user_email=user_email,
                     subject="Updates to Your Saved Opportunities",
-                    content="",
+                    content=message,
                     notification_reason=NotificationReason.SEARCH_UPDATES,
                     notified_object_ids=[
                         saved_search.saved_search_id for saved_search in saved_items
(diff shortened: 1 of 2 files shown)


api/src/task/notifications/search_notification.py
```diff
@@ -74,12 +80,45 @@ def collect_email_notifications(self) -> list[UserEmailNotification]:
                 "Created changed search email notification",
                 extra={"user_id": user_id, "changed_search_count": len(saved_items)},
             )
+
+            all_opportunity_ids = set()
```
joshtonava, line 84: Let's discuss if this makes sense or if I am missing the boat on requirements.

api/src/task/notifications/search_notification.py
```diff
@@ -74,12 +80,45 @@ def collect_email_notifications(self) -> list[UserEmailNotification]:
...
+                    )
+                )
+            )
+
+            opportunities = self.db_session.execute(opportunities_stmt).scalars().all()
+
+            message = self._build_notification_message(opportunities)
+
+            formated_date = datetime_util.utcnow().strftime("%-m/%-d/%Y")
+            subject = (
```
joshtonava, line 111: Also its unclear to me what date I should use here. If this is a broader product question I can move the conversation.

api/src/task/notifications/search_notification.py
chouinar: Saved searches was built a bit differently than what I think the content expects.

We saved/compare against the top 1000 results from the saved search query. If any of those 1000 change, we send out a notification.

Here the `saved_search.searched_opportunity_ids` value is the result of the "current" query and is what a user would get if they searched now, above it was added because there was something different, but what is different isn't specified.

How we actually want to do this, no idea. We're probably working with way too many opportunities anyways.

joshtonava: I updated this based on discussion from slack to be the opportunities that were not previously in the list before.

api/src/task/notifications/search_notification.py
joshtonava: Using the current date per discussion in [slack](https://betagrantsgov.slack.com/archives/C08MTT10SKU/p1747761611140069)

api/src/task/notifications/search_notification.py
```diff
@@ -50,16 +56,29 @@ def collect_email_notifications(self) -> list[UserEmailNotification]:
...
         for searches in query_map.values():
             current_results = search_opportunities_id(self.search_client, searches[0].search_query)
             for saved_search in searches:
-                previous_results = set(saved_search.searched_opportunity_ids)
-                if set(current_results) != previous_results:
+                previous_results = set(saved_search.searched_opportunity_ids or [])
+                current_results_set = set(current_results)
+
+                # Find NEW opportunities (in current but not in previous)
+                new_opportunities = current_results_set - previous_results
```
chouinar, line 71: This is where we might want one additional filter - because these result sets are both up to 1000 opportunities (assume someone saved a really generic search query), this could be a really, really big list.

I'd really recommend making this just be the first 10 or so of the current result set. Something like:
```py
current_results_set = set(current_results[:10])
```
We'd still want to save the full up-to-1000 list at the end for the next time around, but this way the actual things we notify a user on are on the first page of search results.

api/src/task/notifications/search_notification.py
```diff
@@ -101,6 +166,76 @@ def collect_email_notifications(self) -> list[UserEmailNotification]:
...
+
+            # Add opportunity title (empty line before title)
+            message += f"{opportunity.opportunity_title}\n\n"
+
+            # Add status
+            status = (
+                str(opportunity.opportunity_status).capitalize()
+                if opportunity.opportunity_status
+                else "Unknown"
+            )
```
chouinar, line 198: FYI - if it has a summary it has a status since the field isn't nullable on the `current_opportunity_summary` object, but MyPy might not be able to recognize that so it's fine to leave this.

api/src/task/notifications/search_notification.py
```diff
@@ -101,6 +166,76 @@ def collect_email_notifications(self) -> list[UserEmailNotification]:
...
+            message += f"Status: {status}\n"
+
+            # Add submission period
+            if summary.is_forecast:
+                message += "Submission period: To be announced.\n"
```
chouinar, line 203: If it's a forecast, the `forecasted_post_date` and `forecasted_close_date` may be set and could tell this info. Don't need to handle necessarily in this PR.

api/src/task/notifications/search_notification.py
```diff
@@ -101,6 +166,76 @@ def collect_email_notifications(self) -> list[UserEmailNotification]:
...
+            else:
+                if summary.post_date and summary.close_date:
+                    formatted_post = summary.post_date.strftime("%-m/%-d/%Y")
+                    formatted_close = summary.close_date.strftime("%-m/%-d/%Y")
+                    message += f"Submission period: {formatted_post}–{formatted_close}\n"
+                else:
+                    message += "Submission period: To be announced.\n"
```
chouinar, line 210: Post date being null shouldn't be a concern - if it's null it won't have a current opportunity summary.

Close date being null isn't too common, but usually means "accepting submissions indefinitely". Once something is not a forecast, there isn't really an expectation that the submission period would be announced still, it's "live" and I'd recommend not putting this sort of content as it's misleading.

api/src/task/notifications/search_notification.py
mdragon: Good catch here.

api/src/task/notifications/search_notification.py
joshtonava: I dont fully understand the required changes. Let me know if I am off:

If posted:
- Show the submission period range if both dates are defined
- Show just a post date and no end date if close date not defined ?
-- Alternatively, if there is a post date and no close date, show "accepting submissions indefinitely"
-  Remove block where if post date and close date are not defined that shows "Submission period: To be announced.\n""

api/src/task/notifications/search_notification.py
chouinar: Enforcing it at the DB level won't give you useful results - if we only want to show opportunities that are actually new (and not just "new to the query several pages deep") you'd have to trim the sorted list straight from the search query.

Because you're doing set operations, the order is lost, and it could be you're randomly grabbing opportunities 800 deep in the 1000 item list. 

This circles back to a content question of what are we looking to display to a user - as written it's "10 opportunities different between these two 1000-item queries" which isn't meaningful.

joshtonava: Are you saying it wouldnt work to limit the opportunities returned from the entire set of new opportunities like this?

https://github.com/HHS/simpler-grants-gov/pull/5061/files#diff-fea5b608266f965db4039b37021df3abd3f8377582ef7362153d6b8d1878ff39R126

If we care about the order of the results we could sort the results of this query.

joshtonava: If we trim the sorted list straight from the search query.

```

  current_results_set = set(current_results[10:])

  # Find NEW opportunities (in current but not in previous)
  new_opportunities = current_results_set - previous_results
  
```

then we may not find any new opportunities in the set difference because it could be that the first 10 results are in both the current results and the previous results

mdragon: Seems like we should get the list of differences, and then enforce the 10 limit for display in the email, rather than in the raw lists?  I think the queries are sorted by newest first, so there is some sorting already happening in both of the source lists?

api/src/task/notifications/search_notification.py
chouinar: "Submission period to be announced" on a non-forecast isn't something that would ever make sense. If it's not a forecast, there isn't any future announcement that would happen, the non-forecast _is_ the announcement. While a grantor can adjust these dates whenever, saying "to be announced" is misleading because they probably aren't ever going to change.

The post date can't be null - if something has no post date, it can't get loaded to search. Except for something having a post date deleted, and us having the stale cached version in the index, no scenario can produce this scenario. The close date being null is actually valid, sometimes opportunities are just open indefinitely. So, having logic that requires both of them is an issue.

My recommendation (and run this by content folks):
* if post date is null - skip this entirely, it's unlikely but indicates the data is probably stale and about to change really soon
* if close date is null - something like `Submission period opens: {formatted_post}` - maybe say something about it not having a close date.
* otherwise - `Submission period: {formatted_post}–{formatted_close}`

api/src/task/notifications/search_notification.py
mdragon: Since the saved search results could change for reasons other than a new posting, I think even things posted a while ago might still be relevant? For example if someone updates an opportunity and now it matches a saved search it didn't previously? Or if we fix something about search that wasn't correctly finding something previously but now it is? "New results in your search" not "Newly posted opportunities that match your search."

api/src/task/notifications/search_notification.py
joshtonava: Updated logic from Slack discussion

When close date exists: "Submission period: 1/31/2025–4/30/2025"
When close date is null: "Submission period: 1/31/2025–(To be determined)"

api/src/task/notifications/search_notification.py
```diff
@@ -50,16 +56,28 @@ def collect_email_notifications(self) -> list[UserEmailNotification]:
...
             for saved_search in searches:
-                previous_results = set(saved_search.searched_opportunity_ids)
-                if set(current_results) != previous_results:
+                previous_results = set(saved_search.searched_opportunity_ids or [])
+                # Find NEW opportunities (in current but not in previous)
+                new_opportunities = set(current_results) - previous_results
+                top_new_opportunities = set(current_results[:25]) - previous_results
+
+                # Only add searches that have new opportunities
+                if new_opportunities:
```
chouinar, line 73: Should this be:
```suggestion
                top_new_opportunities = set(current_results[:25]) - previous_results

                # Only add searches that have new opportunities
                if top_new_opportunities:
```

api/src/task/notifications/search_notification.py
```diff
@@ -70,16 +88,62 @@ def collect_email_notifications(self) -> list[UserEmailNotification]:
...
+
+            opportunities = self.db_session.execute(opportunities_stmt).scalars().all()
+
+            if not opportunities:
+                logger.info("No opportunities found to notify about", extra={"user_id": user_id})
+                continue
+
+            message = self._build_notification_message(opportunities)
+
+            formated_date = datetime_util.utcnow().strftime("%-m/%-d/%Y")
```
chouinar, line 135:
```suggestion
            formatted_date = datetime_util.utcnow().strftime("%-m/%-d/%Y")
```

PR #5063, opened 2025-05-15
[Issue #4393] Update GET Opportunity endpoint

### LINKED ISSUE #4393: Modify Opportunity tables to use UUID for their primary key instead of integer

joshtonava:
## Summary

Fixes #4393

## Changes proposed

- GET Opportunity endpoint was updated to support UUID and int for legacy id.
- Tests updated to handle both cases
- Open API schema Updated

## Context for reviewers

This is an incremental change for updating Opportunity models to use UUID. We are going to merge changes into to [joshtonava/4393-modify-opportunity-tables-to-use-uuid-for-primary-key](https://github.com/HHS/simpler-grants-gov/tree/joshtonava/4393-modify-opportunity-tables-to-use-uuid-for-primary-key). I expect there to be linting and test failures outside these changes.

## Validation steps

To run effected tests

```bash
make test args="-vv tests/src/api/opportunities_v1/test_opportunity_route_get.py"
```

### DIFF
--- a/api/tests/src/api/opportunities_v1/test_opportunity_route_get.py
+++ b/api/tests/src/api/opportunities_v1/test_opportunity_route_get.py
@@ -95,7 +95,6 @@ def test_get_opportunity_with_attachment_200_legacy(
 ):
     # Create an opportunity with an attachment
     opportunity = OpportunityFactory.create(has_attachments=True)
-    db_session.commit()
 
     # Make the GET request
     resp = client.get(


api/src/services/opportunities_v1/get_opportunity.py
```diff
@@ -49,11 +50,7 @@ def pre_sign_opportunity_file_location(
     return opp_atts
 
 
-def get_opportunity(db_session: db.Session, opportunity_id: int) -> Opportunity:
-    opportunity = _fetch_opportunity(
-        db_session, opportunity_id, load_all_opportunity_summaries=False
-    )
-
+def _setup_attachements(opportunity: Opportunity) -> None:
```
chouinar, line 53:
```suggestion
def _setup_attachments(opportunity: Opportunity) -> None:
```

chouinar (approved):
Two minor things, but LGTM otherwise!

PR #5064, opened 2025-05-15
[Issue #4983] Take in an optional application name to start application endpoint

### LINKED ISSUE #4983: Modify start application endpoint to take in an optional application name
### Summary

After #4982 

Start application should have the request body take in an application name. We'll set this value in the application name column of the application table.

If the user doesn't provide a value, we should instead default this column to having the opportunity number from the opportunity set for that value when an application is created (A competition is always attached to an opportunity).

### Acceptance criteria

- [ ] Updated endpoint
- [ ] Added tests for both scenarios

mikehgrantsgov:
## Summary

Fixes #4983

## Changes proposed

Add optional application name to start application endpoint. Updated API schema, added validation, and included tests for both custom name and default name scenarios.

## Context for reviewers

This change allows users to specify a custom name when creating applications, with the opportunity number used as the default when no name is provided. 

## Validation steps

See modified existing and new unit tests.

### DIFF
--- a/api/src/services/applications/create_application.py
+++ b/api/src/services/applications/create_application.py
@@ -16,7 +16,9 @@ from src.validation.validation_constants import ValidationErrorType
 logger = logging.getLogger(__name__)
 
 
-def create_application(db_session: db.Session, competition_id: UUID, user: User) -> Application:
+def create_application(
+    db_session: db.Session, competition_id: UUID, user: User, application_name: str | None = None
+) -> Application:
     """
     Create a new application for a competition.
     """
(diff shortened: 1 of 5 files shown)


api/src/services/applications/create_application.py
```diff
@@ -67,8 +69,20 @@ def create_application(db_session: db.Session, competition_id: UUID, user: User)
...
             )
 
+    # Get default application name if not provided
+    if (
+        application_name is None
+        and competition.opportunity
+        and competition.opportunity.opportunity_number
+    ):
+        application_name = competition.opportunity.opportunity_number
+
```
chouinar, line 79: I think we can just simplify to this?
```suggestion
    # Get default application name if not provided
    if application_name is None:
        application_name = competition.opportunity.opportunity_number
```
Opportunity can't be none, opportunity number shouldn't ever be none, but even if it is, grabbing None instead of using the "None" in application name is the same thing.

Issue #5065, opened 2025-05-15
Sort out ENV vars for the NOFO Builder


pcraig3:
### Summary

The Nofo Builder app is deploying but it needs some configuration variables to get up and running properly.

### Acceptance criteria

There are a few things to sort out here:

- [ ] add default env vars (like `DEBUG`)
- [ ] sort out `ALLOWED_HOSTS` (needs to dynamically look up the internal docker container IP and the external domain)
- [ ] add in secrets (like `DOCRAPTOR_API_KEY`)

pcraig3:
This is now complete, and [a document exists here](https://docs.google.com/document/d/1zRXPz9qFn2TO5wKG7_EuxKUK7bKz6QLk50ipmf_cRUs/edit?tab=t.0#heading=h.5jkfizrvdjc3) (ask for access).

PR #5066, opened 2025-05-15
[Issue #5065] Add 2 static env vars to the nofos app

pcraig3:
## Summary

This is part of #5065 

## Changes proposed

Right now, the `ALLOWED_HOSTS` var is causing Django to throw a 500 error.

Until we figure that out, let's add this wildcard.


coilysiren (approved):
✅

PR #5073, opened 2025-05-16
[Issue #4695] Add form validation to the submit application endpoint

### LINKED ISSUE #4695: Update the application submit endpoint to validate all forms and error if there are any issues
# Summary

After #4689 

We want to iterate over every form, and add any validation issues together. We want to add a single `ValidationErrorDetail` for any form that fails validation with the following information:
* `type` -> `form_validation_error`
* `message` -> `Form validation failed`
* `field` -> The UUID of the ApplicationForm record
* `details` -> The list of validation error details in the format TODO

We don't currently have a `details` field on the `ValidationErrorDetail` class and will need to add it, it should have a type like: `details: dict = dataclasses.field(default_factory=dict)`

Validating the forms has been done in other endpoints, but we always warned - we want to instead error, but only after validating all forms.

A few callouts:
1. How do we organize the errors so that each form is clearly defined in the error response?
2. If a form is required (from the `CompetitionForm` table) then an empty/null response for a given ApplicationForm is a validation error and should be added to the list.

If there are any validation issues, raise a 422 using `raise_flask_error` and pass the validation issues in (note we'll likely need to modify/add some way of prefixing these?)

# Acceptance criteria
- [ ] Form validation added
- [ ] Many tests added for various scenarios (form validation itself, forms being required/not required, form not even being started at all yet).

chouinar:
## Summary

Fixes #4695

## Changes proposed

* Validates that the required forms on the competition exist
* Validates the json schema against the application forms answers

## Context for reviewers
Validating that required forms are filled out is pretty straightforward - likely will be much more complex in the future (ie. user said yes to a specific question, an optional form becomes required).

Validating that an application form is correct is also not too bad, the main complexity addressed in this PR is communicating that information easily to the user. Since we might have validation issues across several forms we need a way to make it clear which form each validation issue goes to. To solve that, the normal errors object just gets an error object like:
```json
{
   "type": "application_form_validation",
   "message": "The application form has outstanding errors.",
   "field": "application_form_id",
   "value": "<the UUID of the application form>"
}
```
The value can be used to cross-reference it with the data object (that up until now wasn't meaningfully used on errors). The data object will now look like:
```json
{
   "form_validation_errors": {
       "<the UUID of the application form>": [{
                    "field": "$.name",
                    "message": "5 is not of type 'string'",
                    "type": "type",
                    "value": null,
            }]
   }
}
```
The format of the validation errors

## Validation steps
Endpoints are usable locally. First you'll need to authenticate, on the localhost:8080/docs page, go to the login endpoint and there is a link you can click to authenticate, grab the full token from the response page and set it for the `ApiJwtAuth`.

Make sure you've recently run `make db-seed-local` which will create a competition with a form, this competition always has a static ID at the moment of `fd7f5921-9585-48a5-ab0f-e726f4d1ef94`.

Go to the start application endpoint and pass it the above competition ID.

If you take the application ID from the start application endpoint response and immediately try to submit it, you should get an error roughly like:
```json
{
  "data": {},
  "errors": [
    {
      "field": "form_id",
      "message": "Form Test form is required",
      "type": "missing_required_form",
      "value": "3421aeb3-8013-449a-8a04-a1d88e1b32bc"
    },
    {
      "field": "form_id",
      "message": "Form Application For Federal Assistance SFS 424 - Individual is required",
      "type": "missing_required_form",
      "value": "e1cffadb-b316-4685-81b3-a5a43cce3c5a"
    }
  ],
  "internal_request_id": "35e2a368-1220-44fd-a34a-a1abf0864746",
  "message": "The application has issues in its form responses.",
  "status_code": 422
}
```


api/src/services/applications/create_application.py
```diff
@@ -77,6 +78,7 @@ def create_application(
         application_id=uuid.uuid4(),
         competition_id=competition_id,
         application_name=application_name,
+        application_status=ApplicationStatus.IN_PROGRESS,
```
chouinar, line 81: This was a small feature / ticket I missed making. It's a one-line change and I had to fix it to test the endpoints worked together anyways.

api/src/api/schemas/response_schema.py
```diff
@@ -10,6 +10,9 @@ class ValidationIssueSchema(Schema):
     field = fields.String(
         metadata={"description": "The field that failed", "example": "summary.summary_description"}
     )
+    value = fields.String(
+        metadata={"description": "The value that failed", "example": "invalid string"}
```
mikehgrantsgov, line 14: Do we have any requirements around not logging PII? I wonder if any validated fields might possibly contain any PII

mikehgrantsgov (approved):
Looks good! Left just one curiosity around logging user input. Disregard if it's not a consideration.

api/src/api/schemas/response_schema.py
chouinar: We definitely don't want to log PII. So far, only what I've added in this PR would ever get set to the value field, which is just some UUIDs.

We don't log the error object anywhere right now. For example, the logs if someone were to call submit with a missing form would be roughly (just the messages):
```
start request
JWT Authentication Successful
POST /alpha/applications/:application_id/submit
Processing application submit
end request
```
The error would be thrown between those last two logs. The only error info we put on the "end request" message is the status code. Now that I think about it, with so many validation scenarios possible, we should probably put some sort of log message for when those get hit, just so we know what cases are getting hit most often.

PR #5074, opened 2025-05-17
redirect to homepage on auth page logout

### LINKED ISSUE #5059: Redirect users who intentionally signs out from auth-dependent pages
### Summary

Redirect users after sign-out if user intentionally signs out from authenticated pages (e.g., in Workspace)

Current behavior: If users are on an authenticated experience page (e.g., workspace pages like saved opportunities or saved searches pages), they see a warning that they are trying to access a page that is for signed-in users: [Screenshot](https://drive.google.com/file/d/1kY_nQqo23v8dxNrqW3thkjtJ1Cy-lJWb/view?usp=drive_link)

Proposed behavior: If users sign out from an authenticated experience page, they should be redirected to the homepage (simpler.grants.gov), not shown an error message as they do now. 

No need to show a toast.

### Acceptance criteria

- [ ] Redirect set up

GlennTatum:
## Summary

Fixes #5059 

## Changes proposed

Redirect user to homepage on logout

## Context for reviewers

Removed `router.refresh()` in preference to `router.push("/")` where the user is returned back to the simpler.grants.gov homepage.

> Proposed behavior: If users sign out from an authenticated experience page, they should be redirected to the homepage (simpler.grants.gov), not shown an error message as they do now. No need to show a toast.

Since the below is called when the `logout()` [callback](https://github.com/GlennTatum/simpler-grants-gov/blob/9882251551d21755e14e0b33af618dfa107548d1/frontend/src/components/user/UserControl.tsx#L128) is triggered, the JWT session token is still deleted. The behavior in the session being deleted remains the same between calling `router.refresh()` and `router.push("/")`.
```js
    await fetch("/api/auth/logout", {
      method: "POST",
    });
```

If @doug-s-nava could chime and give feedback in on this change as this partially removes the need of `AuthenticationGate` at routes which use the mechanism for making it protected from unauthenticated users to the system that would be great. 

## Validation steps

A basic user saving a grant and then logging out.

auth-logout

### DIFF
--- a/frontend/src/components/user/UserControl.tsx
+++ b/frontend/src/components/user/UserControl.tsx
@@ -132,7 +132,7 @@ export const UserControl = () => {
     });
 
     logoutLocalUser();
-    router.refresh();
+    router.push("/")
   }, [logoutLocalUser, router]);
 
   return (


doug-s-nava (commented):
the issue I see here is that with this change users will be redirected back to the home page no matter where they are when they log out (ex. if a user logs out while on /search). Since we only want to redirect to the homepage if the user logs out from a page that is gated by authentication status, I think a better approach would be to implement this change within the <AuthenicationGate> component, so that the redirect only happens on pages where we specifically want to gate based on authentication

GlennTatum:
> the issue I see here is that with this change users will be redirected back to the home page no matter where they are when they log out (ex. if a user logs out while on /search). Since we only want to redirect to the homepage if the user logs out from a page that is gated by authentication status, I think a better approach would be to implement this change within the component, so that the redirect only happens on pages where we specifically want to gate based on authentication

By adding `redirect("/")` inside of the `AuthenticationGate` the `AuthenticationGate` will no longer render any page when a user ends up on a protected route. 

I think it's an either or situation.

Option A is to keep the Sign In Alert **without** redirecting the user back to the home page on an auth page.

Option B is to remove the Sign In Alert and redirect to the homepage. This option does not notify the user they must be signed in, however, the only way they would even end up on a protected URL would be by typing it in, which is unlikely.

Currently Option B is implemented and the tests still have to be rewritten with [MemoryRouter](https://v5.reactrouter.com/web/api/MemoryRouter) for coverage.

Option B also benefits of making it very simple to **not** redirect the user back to the home page when on a route like `/search`, as `redirect("/")` only gets triggered on pages wrapped with `AuthenticationGate`

frontend/src/components/user/UserControl.tsx
```diff
@@ -132,8 +134,11 @@ export const UserControl = () => {
     });
 
     logoutLocalUser();
+
+    setQueryParam("auth_logout", "ok");
```
doug-s-nava, line 138: Can we change this to follow the pattern of the rest of the app and use camelCase for query param keys? 

Also, since this param is used by the app to determine when to redirect a user upon logout, a more descriptive name, such as `redirectOnLogout` would be nice

Lastly, I think "true" is more in line with general query param value practices, can we use that instead of "ok"?

doug-s-nava, line 138: I'm wary of adding this query param on all logouts, rather than only when we're logging out from a gated page. I can't reproduce any weird behavior based on this at the moment, but adding an unused query param just seems like bad practice and could come back to bite us later. I'm not sure immediately how we would want to implement only adding the param when logging out from gated pages, though. It may be worth some thought.

frontend/src/components/user/AuthenticationGate.tsx
```diff
@@ -1,24 +1,24 @@
...
-  const t = await getTranslations("Errors");
   if (!session?.token) {
     return (
-      <GridContainer className="margin-top-4">
-        <Alert type="error" heading={t("unauthenticated")} headingLevel="h4">
-          {t("signInCTA")}
-        </Alert>
-      </GridContainer>
-    );
+      <Suspense fallback={<Loading message={t("loading")} />}>
```
doug-s-nava, line 18: this is very interesting to me - since there is no async behavior going on in <RequireLogin> I would expect that the fallback would never load here, but it does. Any idea why?

In any case, I'm not sure the brief loading state improves the user experience at all here, so I'd be in favor of removing the suspense unless there's a compelling reason for it that I'm missing.

frontend/src/components/user/RequireLogin.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+import { useTranslations } from "next-intl";
+import { redirect, useSearchParams } from "next/navigation";
+import { Alert, GridContainer } from "@trussworks/react-uswds";
+
+export function RequireLogin() {
+  const t = useTranslations("Errors");
+  const searchParams = useSearchParams();
+
+  if (searchParams.get("auth_logout") != null) {
```
doug-s-nava, line 11: can we make this more exact? I wouldn't want to redirect accidentally in edge cases here
```suggestion
  if (searchParams.get("redirectOnLogout") === "true") {
```

doug-s-nava (commented):
definitely a step in the right direction. This is tricky to implement, thanks for bearing with my comments

GlennTatum:
Closing this PR in reference to [https://github.com/HHS/simpler-grants-gov/pull/5214](https://github.com/HHS/simpler-grants-gov/pull/5214) which has a more suitable solution to this issue.

PR #5076, opened 2025-05-19
[Issue #4970] Set apply endpoints to only allow the owner of an application to interact

### LINKED ISSUE #4970: Modify the apply endpoints to only allow the owner of an application to interact
### Summary

For the following endpoints, we want to only allow someone that is an `ApplicationUser` attached to the application to interact with the application:
* GET /applications/:applicationId
* GET /applications/:applicationId/application_form/:appFormId
* POST /alpha/applications/:applicationId/submit

This logic should be identical for all of them and will be significantly expanded on in the future (roles, different types of users), so I'd recommend making it a shared utility function for now that they all use that takes in the `application` and `user` and does the check. The application users are already a relationship on the application itself, so no additional DB queries will be needed.

This check should be done in the service layer right after we grab the application, and after handling the 404 case for the app not existing, but before any other operations are done.

If the user is not one of the application users, we should raise a 403 error using `raise_flask_error`.

### Acceptance criteria
- [ ] Added authentication to endpoints
- [ ] A lot of test updates, both in the happy path ones (need the user to be associated with app) and for adding tests to verify this behavior for each endpoint

mikehgrantsgov:
## Summary

Fixes #4970 

## Changes proposed

Add `auth_utils` file that checks user authorized access to an application
Add `ApplicationUser` factory
Update existing and add new tests
Adjusted SQLAlchemy relationship cascade rules to prevent primary key errors when associating users with applications

## Context for reviewers

After this is merged, other new tickets (e.g. 4984) may need to be updated to account for these changes.

## Validation steps

See updated and new unit tests

### DIFF
--- a/api/src/services/applications/auth_utils.py
+++ b/api/src/services/applications/auth_utils.py
@@ -1,0 +1,37 @@
+import logging
+
+from src.api.response import ValidationErrorDetail
+from src.api.route_utils import raise_flask_error
+from src.db.models.competition_models import Application
+from src.db.models.user_models import User
+
+logger = logging.getLogger(__name__)
+
+
+def check_user_application_access(application: Application, user: User) -> None:
+    """
+    Check if a user has access to an application.
+    """
+    user_has_access = any(
+        app_user.user_id == user.user_id for app_user in application.application_users
+    )
+
+    if not user_has_access:
+        logger.info(
+            "User attempted to access an application they are not associated with",
+            extra={
+                "user_id": user.user_id,
+                "application_id": application.application_id,
+            },
+        )
+        raise_flask_error(
+            403,
+            "Unauthorized",
+            validation_issues=[
+                ValidationErrorDetail(
+                    type="unauthorized_application_access",
+                    message="Unauthorized",
+                    field="application_id",
+                )
+            ],
+        )
--- a/api/src/services/applications/auth_utils.py
+++ b/api/src/services/applications/auth_utils.py
@@ -1,6 +1,5 @@
 import logging
 
-from src.api.response import ValidationErrorDetail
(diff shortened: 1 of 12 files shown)


api/src/services/applications/auth_utils.py
```diff
@@ -0,0 +1,37 @@
...
+        raise_flask_error(
+            403,
+            "Unauthorized",
+            validation_issues=[
+                ValidationErrorDetail(
+                    type="unauthorized_application_access",
+                    message="Unauthorized",
+                    field="application_id",
+                )
+            ],
```
chouinar, line 36: We can check with the frontend folks, but I think just returning a 403 is probably enough? Never sure what is the right amount of info for something like auth errors, but we're usually putting these validation issues for the frontend to present specific info to the user, and I think "access denied" in some form is about as specific as we need to be?

api/src/services/applications/get_application_form.py
```diff
@@ -4,20 +4,24 @@
...
-    application = db_session.execute(
-        select(Application).where(Application.application_id == application_id)
-    ).scalar_one_or_none()
+    """
+    Get an application form by ID, checking if the user has access to it.
+    """
+    # Get the application
+    application = get_application(db_session, application_id, user)
 
     if not application:
```
chouinar, line 26: Since we have this 404 logic in the get_application function, could remove it here

api/src/services/applications/auth_utils.py
mikehgrantsgov: Makes sense, I just updated this

chouinar (dismissed):

chouinar (approved):

PR #5110, opened 2025-05-20
[Issue #4984] Add an endpoint for updating an application

### LINKED ISSUE #4984: Add an endpoint for updating an application
### Summary

After #4982 and #4983 

This endpoint exists for updating info on the application itself - at the moment the only field that we'll allow a user to update is the application_name field.

This endpoint should be called `PUT /applications/:applicationId`.

Example request:
```json
{
   "application_name": "my new application"
}
```
The application_name field should not be nullable, and must be required (this will differ slightly from the create application endpoint - effectively we won't let someone not set the value here).

As this is a PUT endpoint, we'll update all values on the application from the request - which is just the one value for now.

Auth:
* This endpoint should use user auth
* This endpoint should reject anyone that doesn't own the application like we added in #4970 

### Acceptance criteria

- [ ] Endpoint created
- [ ] Tests added for auth, and basic usage

mikehgrantsgov:
## Summary

Fixes #4984

## Changes proposed

Add endpoint to update an application. For now this only supports `application_name`, but it flexible to support a `dict` of values.
Add request and response schemas
Add unit tests for endpoint

## Context for reviewers

We use user auth here. Built to support additional fields in the future.

## Validation steps

See unit tests


api/src/services/applications/update_application.py
```diff
@@ -0,0 +1,66 @@
...
+                422,
+                "Application name cannot be empty",
+                validation_issues=[
+                    ValidationErrorDetail(
+                        type=ValidationErrorType.REQUIRED,
+                        message="Application name is required",
+                        field="application_name",
+                    )
+                ],
+            )
```
chouinar, line 56: I don't think this is possible from the API since the field is required + not nullable. I think it's fine to just assume it's not null here.

api/src/services/applications/update_application.py
```diff
@@ -0,0 +1,66 @@
...
+            - List of validation warnings (if any)
+
+    Raises:
+        Flask error with appropriate status code if validation fails
+    """
+    # Get application (this will check if it exists and if user has access)
+    application = get_application(db_session, application_id, user)
+
+    # List to collect warnings (non-blocking validation issues)
+    warnings: list[ValidationErrorDetail] = []
```
chouinar, line 40: I don't think we need the warnings list, don't think we'll add the validations to this endpoint for things like the forms.

api/src/services/applications/update_application.py
mikehgrantsgov: Yeah I was curious about this edge case if it would pop up. Adjusted this.

api/src/services/applications/update_application.py
mikehgrantsgov: Got it. I made this change, didn't know how detailed we wanted to be in the update.

chouinar (approved):

PR #5113, opened 2025-05-20
NOFOs vuln scans

coilysiren:
## Changes proposed

Adds a vulnerability scan workflow to the NOFOs deployment process. Currently it's optional.

## Context for reviewers

We have an obligation to vuln scan the apps that run within our infrastructure. 

## Validation steps

CI outputs:

https://github.com/HHS/simpler-grants-gov/actions/runs/15147743231/job/42587548429?pr=5113

https://github.com/HHS/simpler-grants-gov/actions/runs/15147743231/job/42587548461?pr=5113

https://github.com/HHS/simpler-grants-gov/actions/runs/15147743231/job/42587548416?pr=5113

The scans are working, but all 3 are failing

### DIFF
--- a/.github/workflows/cd-nofos.yml
+++ b/.github/workflows/cd-nofos.yml
@@ -36,6 +36,7 @@ jobs:
 
   vulnerability-scans:
     name: Vulnerability Scans
+    needs: [checks]
     uses: ./.github/workflows/vulnerability-scans-nofos.yml
 
   deploy:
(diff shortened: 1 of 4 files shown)


pcraig3 (commented):
General question, why save the 'container to scan' somewhere else than the 'container to deploy'?

coilysiren:
The container to deploy is tagged with the git hash, whereas the container to scan is tagged latest

I'm being a bit lazy here in that I could use the latest tagged container inside the deployment as well, and just re-tag it with the git hash. Which is to say, we don't need to be saving 2 different container images

coilysiren:
Since I can tell from the questions that this is unclear, I will refactor it

pcraig3 (approved):
We talked through this so approved 👍

PR #5117, opened 2025-05-21
[Unticketed] Attempt to fix vulnerability scans for anchore/grype

chouinar:
## Changes proposed

Updates grype/anchore to v6

Fixes a small 

## Context for reviewers

v4 of the anchore scan was last updated in August of last year. From reading some docs they changed the internal DB schema format at some point (around v5 or v6?) so I'm guessing they stopped making new iterations of the db compatible with v4 recently. I can't find anything describing a deprecation or a way of finding that the versions aren't supported anymore though.

The current database can be fetched from:
https://grype.anchore.io/databases/v6/latest.json

But it doesn't give anything for earlier version numbers, so might be entirely unrelated.


mdragon (dismissed):
Approved

coilysiren (approved):
Note to self to do this here too

https://github.com/HHS/simpler-grants-gov/pull/5113

PR #5118, opened 2025-05-21
[Issue #4393] Application Updates for migrating Opportunity tables to use UUID

joshtonava:
## Summary
Fixes #{4393}

## Changes proposed
- This PR contains uuid updates for Opportunity tables on the application side.
- Included a field mapping for opportunity_id in opportunity_search
- Duplicated DELETE saved opportunities endpoint for legacy id
- Updated open API schema
- Fixed application tests

## Context for reviewers

```bash
make test args="-vv tests/src/adapters/db/type_decorators/test_postgres_type_decorators.py tests/src/api/competition_alpha/test_competition_route_get.py tests/src/api/opportunities_v1/test_opportunity_route_search.py tests/src/api/users/test_user_delete_saved_opportunity.py tests/src/api/users/test_user_save_opportunity_post.py tests/src/api/users/test_user_saved_opportunities_get.py tests/src/api/users/test_user_save_search_post.py tests/src/services/opportunities_v1/test_opportunity_version.py"
```


api/src/services/opportunities_v1/get_opportunity.py
```diff
@@ -1,4 +1,5 @@
 import uuid
+from typing import Optional
```
chouinar, line 2: Instead of `Optional` just do `WhateverType | None`

api/tests/src/services/opportunities_v1/test_opportunity_version.py
```diff
@@ -25,12 +25,13 @@ def test_save_opportunity_version(db_session, enable_factory_create):
     expected = {
         "agency": agency.agency_code,
         "summary": None,
-        "category": OpportunityCategory.MANDATORY,
+        "category": str(OpportunityCategory.MANDATORY),
```
chouinar, line 28: This is already a string / shouldn't be hitting any issues with comparison. If it is, that's 
```suggestion
        "category":OpportunityCategory.MANDATORY,
```
If you saw a weird comparison / it looked like it wouldn't match when you were fixing the UUIDs, that's normal. Because they're `StrEnum` it'll be like "dict1 doesn't match dict2" and print the two dicts, one will be a string, and the other is like `<MyEnum.SOMETHING: 'something'>` but it's really also a string, that's just the repr function.

api/src/api/users/user_schemas.py
```diff
@@ -75,7 +75,7 @@ class UserGetResponseSchema(AbstractResponseSchema):
 
 
 class UserSaveOpportunityRequestSchema(Schema):
-    opportunity_id = fields.Integer(required=True)
+    opportunity_id = fields.UUID(required=True)
```
chouinar, line 78: If we change saved opportunity request to a UUID - it would break the frontend. I think we might need the saved opportunity endpoints to handle the ID?

I guess that depends on how the frontend passes that ID around / how strongly type checked their stuff is. If they call GET /opportunity and we return a UUID will that in turn get passed to the call to the saved opportunity endpoints?

Might be worth just spinning up the frontend locally and clicking around the pages and see if anything is broken.

joshtonava: good call, will see what happens

api/src/services/opportunities_v1/get_opportunity.py
joshtonava: If thats preferred I can do that :)

api/src/api/users/user_schemas.py
joshtonava: Looks like the frontend does a validation check on the ID,  when loading the opportunity page,  that will need to be addressed 

image

Additionally

image

and 

image

I think this can be done separately in a different frontend only PR

joshtonava: After addressing those, I was able to save the opportunity just fine. @chouinar

chouinar (approved):

PR #5119, opened 2025-05-21
[Issue #4985] Don't allow a user to modify an application if it's not in-progress

### LINKED ISSUE #4985: Do not allow a user to modify an application or its components once submitted
### Summary

If the current application status of an application is not `in_progress`, we don't want to allow a user to modify the application.

This logic should apply to the following endpoints:
* PUT /applications/:applicationId - This is being added in #4984 

We already added logic for this in the submit application endpoint, we should reuse that logic here, moving it to a shared utility that all 3 endpoints use.

This logic should occur after we fetch the application, and after we verify a user can interact with the application.

### Acceptance criteria

- [ ] Endpoints updated
- [ ] Tests added to verify behavior

chouinar:
## Summary

Fixes #4985

## Changes proposed

This applies to the endpoints for submitting an application, as well as updating an application, and updating an application form.

Cleaned up some of the unit tests to use user/auth token utils instead of redoing the logic each time (a bit more cleanup could be done, but I'll separate that out later). Also removed some of the date freezing logic because it broke the auth cleanup - not necessary to freeze time if we already compare to "today".

## Context for reviewers

Once a user submits an app, they shouldn't be able to modify it. Pretty uneventful change, we already had the logic on the submit endpoint, just moved it to a central location and reused it.


mikehgrantsgov:
Looks good to me, too. Is there ever a scenario where we would want a submitted application to be modified (i.e. a privileged user who needs to make an edit).  Overall, looks good!

chouinar:
> Looks good to me, too. Is there ever a scenario where we would want a submitted application to be modified (i.e. a privileged user who needs to make an edit). Overall, looks good!

Potentially yes, but we'll handle adjusting that whenever roles start to be a thing.

PR #5122, opened 2025-05-21
[Issue #5106] Modify application response schema to include additional fields

### LINKED ISSUE #5106: Modify application response schema to include additional fields
### Summary

We want to modify our `ApplicationGetResponseDataSchema` to include additional fields:
* `application_status` - should be an enum of the ApplicationStatus type
* `application_name` - string
* `users` - see note below

Users should be a list of user objects that for now only contain the:
* `user_id`
* `email`
Fields from the user object.

Roughly, the data model should now look like:
```json
{
    "application_forms": [
      {
        "application_form_id": "123e4567-e89b-12d3-a456-426614174000",
        "application_id": "123e4567-e89b-12d3-a456-426614174000",
        "application_response": {...},
        "form_id": "123e4567-e89b-12d3-a456-426614174000"
      }
    ],
    "application_id": "123e4567-e89b-12d3-a456-426614174000",
    "competition_id": "123e4567-e89b-12d3-a456-426614174000",
    # new stuff below here
    "application_status": "in_progress",
    "application_name": "My new application",
    "users": [
        {
            "user_id": "123e4567-e89b-12d3-a456-426614174000",
            "email":"example@example.com"
        }
    ]
  }
```

Because the application_users relationship is to the `application_user` table and not the user table, a trick we can do to get a list of user objects is to add a property to the application table like so:

```py
@property
def users() -> list[User]:
     return [app_user.user for app_user in self.application_users]
```
This way when the Marshmallow model tries to turn the DB models into JSON, it fetches the right user object from the property.

### Acceptance criteria

- [ ] Updated schema of application GET response
- [ ] Tested that fields get piped through as expected

mikehgrantsgov:
## Summary

Fixes #5106

## Changes proposed

Modify the `ApplicationGetResponseDataSchema` to include additional fields:

application_status
application_name
users: [user_id, email]

## Context for reviewers

User association implemented as a property to the application table. Also had to add `selectinload` for `application_users` to avoid detached objects from SQLAlchemy sessions.

## Validation steps

See new and updated unit tests.


api/src/services/applications/get_application.py
```diff
@@ -1,24 +1,32 @@
...
     """
-    application = db_session.execute(
+    result = db_session.execute(
         select(Application)
-        .options(selectinload(Application.application_forms))  # Fetch the application forms
+        .options(
+            selectinload(Application.application_forms),
+            joinedload(Application.application_users)
+            .joinedload(ApplicationUser.user)
+            .joinedload(User.linked_login_gov_external_user),
```
chouinar, line 23: Could we use selectinload instead?

From some prior testing of joinedload, it sometimes just locks queries up if big enough - plus adds complexity for needing things like the unique you added below.

api/src/api/application_alpha/application_schemas.py
```diff
@@ -38,10 +38,20 @@ class ApplicationFormGetResponseSchema(AbstractResponseSchema, WarningMixinSchem
     data = fields.Nested(ApplicationFormGetResponseDataSchema())
 
 
+class UserSchema(Schema):
```
chouinar, line 41: Could you name this something more specific like `ApplicationUserSchema`? Marshmallow registers all the class names globally and it'll start logging warnings whenever the app starts up if there is a name overlap.

api/src/services/applications/get_application.py
mikehgrantsgov: Makes sense, I updated this

api/src/api/application_alpha/application_schemas.py
mikehgrantsgov: Ah makes sense. I saw it was named `User1`.. not ideal. Updated this

chouinar (approved):

chouinar (approved):

PR #5127, opened 2025-05-21
[Issue #4785] abstract sub agencies for saved search

### LINKED ISSUE #4785: Abstract sub-agencies into parent agency for saved search purposes
### Summary:
When a user saves a search that includes "All" of a given agency or department, that saved search will only currently account for agencies which have an active opportunity. Given the changes that are expected between Agencies/Departments and sub-agencies over time, we need to adjust the logic to account for this

### Acceptance Criteria:

- [ ] Abstract all known sub-agencies into a parent agency or department
- [ ] Update the saved search capability so that if "All" is selected, we save the search as the top level agency/department for future searches to pick up new sub-agency results

babebe:
## Summary

Fixes #4785 

Add top level agency opportunity search schema filter
Use open search prefix filter to filter for sub agencies for given top level agency along with the given sub agencies. 
Update tests


api/src/adapters/search/opensearch_query_builder.py
```diff
@@ -177,12 +177,23 @@ def filter_terms(self, field: str, terms: list) -> typing.Self:
...
 
+    def filter_should_prefix(self, field: str, value: str) -> typing.Self:
+        """
+        Adds a prefix filter under a should clause for the specified field.
+
+        This is used to match documents where the field value starts with the given prefix.
+        The filter is added to a `bool.should` context, allowing OR-style matching when combined
+        with other should clauses.
+        """
+        self.should.append({"prefix": {field: f"{value}-"}})
```
chouinar, line 188: Don't put the `-` in the builder, if we wanted to have prefix filters for things that aren't agencies they wouldn't work unless they had the exact same behavior.

api/src/services/service_utils.py
```diff
@@ -77,3 +77,29 @@ def _add_search_filters(
                 start_date,
                 end_date,
             )
+
+
+def _add_top_level_agency_prefix(
```
chouinar, line 82: This function doesn't feel like it should be in this generic util as it's very specific to the use case of the search opportunities. It expects filters to basically be a specific type, but you're having it take in a generic type and then doing several hasattr checks. I'd just move this function to the search opportunities logic and not worry about making it generic.

chouinar (approved):

Issue #5141, opened 2025-05-22
Solve issues raised by vulnerability scans for NOFO Builder


pcraig3:
### Summary

Now that the vulnerability scans are running (#5113), we want them to be resolved so that we can make vuln scans a required part of the NOFO app deployment workflow.

### Acceptance criteria

Three green checkmarks when running vulnerability plans!

- [x] Trivy
- [x] Anchore/Grype
- [ ] Dockle

PR #5142, opened 2025-05-22
[#5141] Ignore settings.py python files flagged by Dockle

pcraig3:
## Summary

Addresses #5141, but doesn't completely resolve it.

This should cause the Dockle scans to go green by ignoring `settings.py` files in the NOFO apps.

## Changes proposed

Dockle's automated linting rules are flagging what it says are "Suspicious filenames", which turn out to be settings.py files. 

Example here: https://github.com/HHS/simpler-grants-gov/actions/runs/15190778398/job/42722888417

`settings.py` is a Python convention, in Django in particular.

I think this "FATAL" message is flagging these files because they might contain plaintext passwords or other configuration, but they don't, so all good.

This PR adds 5 settings.py files to the .dockleignore DOCKLE_ACCEPT_FILES string.


mdragon (approved):
🚀

PR #5150, opened 2025-05-23
[Unticketed] Fix error when a non-string is passed for an enum

chouinar:
## Changes proposed

Fixes an issue where if a non-string is passed in a request body for something that is a str enum, it would cause a 500 error when failing to hash the value

## Context for reviewers
We have a custom handler for str enums in our Marshmallow schema definitions. Internally we setup a dict that is a mapping of strings to enum values. If the value a user passes in is not hashable (like `{}`), it would break and return a 500 error. This fix adds a guardrail to shortcut the check to not break when a non-string is passed in.

## Validation steps
Calling the opportunity search endpoint, a request like:
```json
{
  "format": {},
  "pagination": {
    "page_offset": 1,
    "page_size": 25,
    "sort_order": [
      {
        "order_by": "opportunity_id",
        "sort_direction": "ascending"
      }
    ]
  }
}
```
Would cause the error before because the format value which is a str enum would hit the above issue. Now it just says it's an invalid value and returns a 422 instead.

### DIFF
--- a/api/src/api/schemas/extension/schema_fields.py
+++ b/api/src/api/schemas/extension/schema_fields.py
@@ -252,6 +252,11 @@ class Enum(MixinField):
     ) -> typing.Any:
         val = self.field._deserialize(value, attr, data, **kwargs)
 
+        # If the value isn't a string, we know
+        # it can't be a StrEnum
+        if not isinstance(val, str):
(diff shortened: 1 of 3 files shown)


mdragon (approved):
Approved

PR #5160, opened 2025-05-23
[Issue #4994] Allow feature flags to know defaults and user values independently

### LINKED ISSUE #4994: Allow env vars to take precedence over client side cookie when no feature flag client overrides are set
### Summary

After dealing with a situation in the current system where changing the feature flag env var value in prod did not result in the feature flag change taking effect for users that had a pre-existing feature flag cookie on their system, we will update feature flag override logic to support the default use case more directly

Technical note - the easiest way to do this is probably with a separate "feature flag overrides" cookie. That cookie will only be set
* in the middleware if a query param comes through
* in a client side cookie if set through the UI (or maybe we could just do this all through query params and have the middleware take care of setting the cookie?)

The new cookie will be read in the middleware and used to override default / env var values and set the main cookie. The main cookie will be sent to the frontend and used as the source of truth for the frontend, but will not be read by the middleware. After being used the override cookie should be sent on every request, it can have a decently long expiration.

We discussed doing this by allowing for a "default" setting on the cookie, but since the frontend needs a clear "true / false" value to know what to do, that seems difficult to implement

See https://betagrantsgov.slack.com/archives/C08MTT10SKU/p1746715675590439 and https://github.com/HHS/simpler-grants-gov/pull/4991 for more context

### Acceptance criteria

- [ ] when a user has set no client side feature flag overrides (either from a query param or the /dev/feature-flags UI), feature flags should always be set to the value provided by the environment variable or hardcoded default
- [ ] expiration time of main ff cookie is reduced to ca 1 week

mdragon:
## Summary

Fixes #4994

## Changes proposed
We need to store user set feature flags separately from system/default values. This is important so that user's cookies don't lock out new features from when the default was set to off.

### DIFF
--- a/frontend/src/app/[locale]/dev/feature-flags/page.tsx
+++ b/frontend/src/app/[locale]/dev/feature-flags/page.tsx
@@ -2,6 +2,7 @@ import { Metadata } from "next";
 
 import Head from "next/head";
 import React from "react";
+import { Button } from "@trussworks/react-uswds";
 
 import FeatureFlagsTable from "src/components/dev/FeatureFlagsTable";
 
@@ -22,9 +23,16 @@ export default function FeatureFlags() {
       <Head>
         <title>Manage Feature Flags</title>
       </Head>
-      <div>
+      <div className="grid-container">
         <h1>Manage Feature Flags</h1>
+
         <FeatureFlagsTable />
+
+        <a href="?_ff=reset">
+          <Button type="button" data-testid="reset-defaults">
+            Reset all flags to defaults
+          </Button>
+        </a>
       </div>
     </>
   );
--- a/frontend/src/services/featureFlags/FeatureFlagManager.ts
+++ b/frontend/src/services/featureFlags/FeatureFlagManager.ts
@@ -6,10 +6,13 @@ import {
   defaultFeatureFlags,
   FeatureFlags,
 } from "src/constants/defaultFeatureFlags";
-import { environment, featureFlags } from "src/constants/environments";
+import { featureFlags } from "src/constants/environments";
 import {
   assignBaseFlags,
+  deleteCookie,
+  FEATURE_FLAGS_DEFAULTS_KEY,
   FEATURE_FLAGS_KEY,
+  getDefaultFlagsFromCookie,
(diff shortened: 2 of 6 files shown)


frontend/src/app/[locale]/dev/feature-flags/page.tsx
```diff
@@ -2,6 +2,7 @@ import { Metadata } from "next";
 
 import Head from "next/head";
 import React from "react";
+import { Button } from "@trussworks/react-uswds";
```
Comment (mdragon) [inline frontend/src/app/[locale]/dev/feature-flags/page.tsx:5]: Based this off the reset button branch so you can ignore these as they're in a different PR.

frontend/src/app/[locale]/dev/feature-flags/page.tsx
```diff
@@ -22,9 +23,16 @@ export default function FeatureFlags() {
       <Head>
         <title>Manage Feature Flags</title>
       </Head>
-      <div>
+      <div className="grid-container">
```
Comment (mdragon) [inline frontend/src/app/[locale]/dev/feature-flags/page.tsx:26]: Based this off the reset button branch so you can ignore these as they're in a different PR.

frontend/tests/pages/dev/feature-flags/page.test.tsx
```diff
@@ -71,4 +71,32 @@ describe("Feature flags page", () => {
       expect(statusElement).toHaveTextContent("Enabled");
     });
   });
+
+  it("should set feature flags to their default state when clicking reset to default button", () => {
```
mdragon, line 75: Based this off the reset button branch so you can ignore these as they're in a different PR.

frontend/src/services/featureFlags/FeatureFlagManager.ts
```diff
@@ -93,6 +96,7 @@ export class FeatureFlagsManager {
 
     // Start with the default feature flag setting
     const currentFeatureFlags = {
+      ...getDefaultFlagsFromCookie(cookies),
```
mdragon, line 99: Unclear if we need this, but if we do, it should be before the local defaults in `this.featureFlags`

mdragon:
Evaluating a better way to deal with this so closing the PR for now.

PR #5171, opened 2025-05-27
[Issue #5138] Add validation to the GET /application endpoint

### LINKED ISSUE #5138: Modify GET /applications to add validation warnings
### Summary

Similar to the form validation we do in the submit application endpoint, we want to run the same validation on each of the application forms.

The errors that it produces instead attach as warnings. The detail object attach in the data model (effectively the data model will be an application/validation detail hybrid).

### Acceptance criteria

- [ ] GET /applications logic updated to do validation checks
- [ ] Tests updated

chouinar:
## Summary

Fixes #5138

## Changes proposed
Runs validation and returns warnings in the GET /application endpoint

## Context for reviewers
Nothing too noteworthy, same validation we're already running elsewhere, just piping it through here as well. Uses the same rough pattern (specific validation issues for a given form are stored in the data object) as the submit endpoint.

## Validation steps
`make volume-recreate db-seed-local run-logs` to remake your local DB + populate it (recreate might not be necessary but useful if you've got some odd stuff in your local DB). This creates a competition you can use with a static ID of `fd7f5921-9585-48a5-ab0f-e726f4d1ef94`

If you start an application and immediately fetch it, you'll get a set of warnings like below to indicate you're missing forms.
```json
[
    {
      "field": "form_id",
      "message": "Form Test form is required",
      "type": "missing_required_form",
      "value": "90923e7e-c1bb-410e-babf-9593dfdcfa71"
    },
    {
      "field": "form_id",
      "message": "Form Application For Federal Assistance SFS 424 - Individual is required",
      "type": "missing_required_form",
      "value": "eb65c432-f696-4818-902a-36451598f5db"
    }
  ]
```

If there are specific validation issues like fields missing of something that has actually started, then we'll add that list to the application warnings object in the application itself (Not an actual DB column).


api/openapi.generated.yml
```diff
@@ -2932,6 +2932,16 @@ components:
             type:
             - object
             $ref: '#/components/schemas/ApplicationUser'
+        form_validation_warnings:
+          type: object
+          description: Specific form validation issues
+          example:
+            123e4567-e89b-12d3-a456-426614174000:
+            - field: $
+              message: '''name'' is a required property'
```
joshtonava, line 2941: **question:** What is with the double single quotes? `''`

joshtonava (approved):
This is neat! I like that we put this type of data validation in the api responses.

PR #5174, opened 2025-05-27
Dschrashun/4280 modal refactor

### LINKED ISSUE #4280: Improve close behavior on all Simpler Modals

doug-s-nava:
## Summary

Fixes #4280


frontend/src/components/workspace/EditSavedSearchModal.tsx
```diff
@@ -183,17 +181,15 @@ export function EditSavedSearchModal({
...
         className="text-wrap"
-        aria-labelledby={`${modalId}-heading`}
-        aria-describedby={`${modalId}-description`}
-        id={modalId}
+        modalId={modalId}
+        titleText={updated ? undefined : `${t("title")} ${queryName}`}
         onKeyDown={(e) => {
           if (e.key === "Enter") handleSubmit();
         }}
+        onClose={onClose}
```
doug-s-nava, line 192: allows resetting the state of the form when the modal closes

myduong-navapbc (approved):

acouch (approved):
Nice, lightweight wrapper! 👍🏿 

If I understand correctly, the server rendered modals aren't as accessible, b/c they are rendered outside the accessible app tree. We could try and adopt something like https://github.com/jesstelford/react-portal-universal to fix that, though would be a heavy lift.

PR #5184, opened 2025-05-28
Double the RAM of nofo app containers to 4096 MB

pcraig3:
## Summary

This is to help us migrate our data to the NOFOs environments. Once the data is migrated, we can bring this back down to 2048 where it was originally.

### DIFF
--- a/infra/nofos/app-config/env-config/variables.tf
+++ b/infra/nofos/app-config/env-config/variables.tf
@@ -98,7 +98,7 @@ variable "instance_cpu" {
 variable "instance_memory" {
   description = "Memory in MiB for the ECS container instances"
   type        = number
-  default     = 2048
+  default     = 4096
 }
 
 variable "instance_desired_instance_count" {


infra/nofos/app-config/env-config/variables.tf
```diff
@@ -98,7 +98,7 @@ variable "instance_cpu" {
 variable "instance_memory" {
   description = "Memory in MiB for the ECS container instances"
   type        = number
-  default     = 2048
+  default     = 4096
```
mdragon, line 101: Just FYI, there's a matrix of CPU vs Memory in AWS ECS: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#task_size 

It looks like going to 4GB is fine for our 1024 (1 vCPU) CPU config, so this is just a heads up.

mdragon (approved):
👍🏻

infra/nofos/app-config/env-config/variables.tf
pcraig3: Thanks for the tip.

PR #5189, opened 2025-05-29
[Issue #5188] Lookup the app folder specific most recent hash (no later than the current ref being deployed)

### LINKED ISSUE #5188: Use app specific git hash to check if we've already built this container
### Summary

Currently we use the top level repo git hash to determine which container tags to check for each of our services. But we don't build every service for every commit, so we miss finding an already built container for 2/3rds of our apps in Prod Releases. We find the existing container for the most recently committed app in main, but miss finding the already built container in the other 2.

### Acceptance criteria

- The most recent container logic finds the correct hash per app folder, not at the top level of the repo.

mdragon:
## Summary

Fixes #5188

## Changes proposed

This change attempts to locate the most recent git commit hash within each app folder (which is not the same as the top level commit for 2 of 3 apps) and use that to check if the container already exists and can just be deployed as already built.

## Context for reviewers

Currently the deploy steps attempt to re-use already built containers (built for the lower environment). But due to it checking by the top level git hash associated with the merge/release, we will only ever find an existing container for the app that had the most recent git commit.

## Validation steps

image
image
Previously, the build and deploy steps would have checked if the top level repo HEAD existed as a frontend container. But c60e4a2a0b113fd5d91551b1ae676a7a7930844a was created by my commits working on the deploy scripts, so there wouldn't be a FE container representing that hash. So we would rebuild the FE container even though there was no changes to the FE. Now it will use the hash from the most recent commit to frontend folder specifically, 6f51edeb383bc1b0bc46da1fe2d3087f3e102a97 which is found, and deployed without rebuilding.

This is a bit contrived because I'm force running the FE deploy on a branch with no FE changes. But this mirrors how the Prod release works, where we deploy all 3 services, but usually at most one of those 3 services actually had changes in the most HEAD so we were rebuilding 2 of 3 services for every Prod Release.

### DIFF
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -40,9 +40,10 @@ jobs:
       - name: Get commit hash
         id: get-commit-hash
         run: |
+          APP_COMMIT_HASH=$(git log --pretty=format:'%H' -n 1 ${{inputs.ref}} ${{ inputs.app_name}})
           COMMIT_HASH=$(git rev-parse ${{ inputs.ref }})
-          echo "Commit hash: $COMMIT_HASH"
-          echo "commit_hash=$COMMIT_HASH" >> "$GITHUB_OUTPUT"
+          echo "Commit hash: $COMMIT_HASH, App: $APP_COMMIT_HASH"
+          echo "commit_hash=$APP_COMMIT_HASH" >> "$GITHUB_OUTPUT"
   build-and-publish:
     name: Build and publish
     runs-on: ubuntu-22.04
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-publish.yml
```diff
@@ -37,12 +37,14 @@ jobs:
       - uses: actions/checkout@v4
         with:
           ref: ${{ inputs.ref }}
+          fetch-depth: 1000
```
mdragon, line 40: This is needed because the default for the checkout action is to fetch a depth of zero, the most recent commit only. This makes all files and folders in the checkout report that hash as their most recent commit. Initially I tested this with 100 not 1000 and that was only taking 2 seconds, so I bumped it up higher. The risk of having this number too low is if an App hasn't had a change in more than this number of commits it will not return an existing container hash and we'll rebuild, so not a huge deal, but something we want to avoid as much as possible.

coilysiren (approved):
lets see what happens!

PR #5190, opened 2025-05-29
[Issue #5132] Modify GET /application to also return the competition

### LINKED ISSUE #5132: Modify GET /application to also return the competition
### Summary

Right now our GET /application schema returns the competition ID - instead make it return the competition itself - reuse the existing competition schema from the GET /competition endpoint.

### Acceptance criteria

- [ ] Remove `competition_id` from the GET /application schema
- [ ] Add `competition` to the schema

mikehgrantsgov:
## Summary

Fixes #5132

## Changes proposed

Return the full competition in the GET /application response rather than just the competition_id in the `ApplicationGetResponseDataSchema`.

## Context for reviewers

Without all the new added `selectinload` calls in `get_application`, I would get lazy loading errors in unit tests. Maybe there's a better way than all these new `selectinload` calls?

## Validation steps

See updated unit tests.

### DIFF
--- a/api/src/services/applications/get_application.py
+++ b/api/src/services/applications/get_application.py
@@ -4,9 +4,11 @@ from sqlalchemy import select
 from sqlalchemy.orm import lazyload, selectinload
 
 import src.adapters.db as db
+from src.api.response import ValidationErrorDetail
 from src.api.route_utils import raise_flask_error
 from src.db.models.competition_models import Application, Competition
 from src.db.models.user_models import ApplicationUser, User
+from src.services.applications.application_validation import get_application_form_errors
 from src.services.applications.auth_utils import check_user_application_access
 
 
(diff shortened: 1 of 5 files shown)


api/src/services/applications/get_application.py
```diff
@@ -23,6 +28,15 @@ def get_application(db_session: db.Session, application_id: UUID, user: User) ->
...
             .selectinload(ApplicationUser.user)
             .selectinload(User.linked_login_gov_external_user),
+            selectinload(Application.competition)
+            .selectinload(Competition.competition_forms)
+            .selectinload(CompetitionForm.form),
+            selectinload(Application.competition).selectinload(
+                Competition.opportunity_assistance_listing
+            ),
+            selectinload(Application.competition).selectinload(
+                Competition.link_competition_open_to_applicant
```
chouinar, line 38: I wonder if at this point it is better to just do a `selectinload("*")` and then exclude the ones we don't want.

We wouldn't want some of the stuff on the user table (saved searches/opportunities), we wouldn't want the competition.opportunity or competition.applications - but I think that's all we'd need to exclude? Worth seeing if that's any clearer, but no need to look into that too much.

FYI - I just merged the PR where the application forms point to the competition forms table, so this might need to be adjusted slightly with that pulled in.

chouinar (commented):
Change looks good, but likely need a small adjustment if you merge from main + check if we can minimize that selectinload a little (try it, not a hard requirement if it's more complex the other way)

chouinar (approved):

PR #5191, opened 2025-05-29
[Issue #5078] Create an endpoint for adding an attachment to an application

### LINKED ISSUE #5078: Create a PUT /applications/:application_id/attachments endpoint
### Summary

Create `PUT /applications/:application_id/attachments` which is used to add an attachment to an application.

### Request
Unlike our other endpoints, we actually don't want to use a standard JSON request body, instead we want to use the `form_and_files` type which allows a user to upload a file in a multipart stream.

You can see a very rough prototype of this behavior at: https://github.com/HHS/simpler-grants-gov/compare/chouinar/prototype-upload

The form should take in the following:
* `application_attachment` - the file itself - see the prototype above for setting up the "file" type object we want.

### Response
The primary thing we need to return is the attachment_id.

```json
{
  "data": {
    "application_attachment_id": "<some uuid>"
  },
  "message": "Success",
  "status_code": 200
}
```

### Logic

* Fetch the application, 404 if not found
* User authentication, only an owner of the application can modify it (same as other endpoints, reuse their logic)
* Adjust the filename using werkzeug's `secure_filename` function
* Upload the file to s3
   * The file should be uploaded to our "draft" bucket which is private
   * The filepath should be something like `s3://<draft_bucket>/applications/<application_id>/attachments/<attachment_id>/<file_name>`
* Create the application attachment record in the DB.
* Return the application attachment ID in the response.

### Acceptance criteria

- [ ] Endpoint added
- [ ] Thorough tests added

chouinar:
## Summary

Fixes #5078

## Changes proposed
Create `POST /applications/:application_id/attachments` which allows adding a file to your application.

## Context for reviewers
This is our first endpoint for uploading files, so a few utilities needed to be setup to support that. A few things to call out:
* To support uploading large files, we don't want to use a standard JSON body request, but instead use a "form" request which allows for multi-part upload of the file which should perform better than having to base64 encode a file into a JSON object.
* The file we upload to s3 has a slightly modified filename to keep things simple (removes spaces and non-ascii characters) - the users actual file name is stored in the DB so on the UI it matches whatever they might have on disk.
* The mimetype of the file is currently pulled from the form and is set by the browser actually, it can be a generic "application/octet-stream" if not specified which is fine. We could look into guessing it, but I started down that path and it's definitely its own dedicated ticket to setup.

## Validation steps
If you create an application locally, you can easily add files to it via the swagger docs - it handles the form type just fine and you can select files:
Screenshot 2025-05-29 at 10 50 58 AM

I did also do some testing of running curls to upload files (swagger will give you the curl command) and removed the content type - it still works fine but just makes things be "application/octet-stream" in most cases.


api/src/services/applications/create_application_attachment.py
```diff
@@ -0,0 +1,93 @@
...
+    # Fetch the application - handles checking if application exists & user can access
+    application = get_application(db_session, application_id, user)
+
+    # This uses a werkzeug FileStorage object for managing the file operations
+    # Mimetype is set if the user specifies it when uploading the file which
+    # if it's done via a standard HTML file box would include it.
+    file_attachment: FileStorage = cast(FileStorage, form_and_files_data.get("file_attachment"))
+
+    # secure_filename makes the file safe in path operations and removes non-ascii characters
+    secure_file_name = file_util.get_secure_file_name(file_attachment.filename)
```
mikehgrantsgov, line 29: Is there any value in preserving the original file name? I think it can be stored as metadata on the S3 object if so.

mikehgrantsgov (approved):
It looks good if we don't want to preserve the original file name. If we do want to, consider uploading it as additional metadata in the PutObject request

api/src/services/applications/create_application_attachment.py
chouinar: We do store it on the attachment table itself so when a user is looking at what they've uploaded that will match what they might have locally.

As for adding that as metadata on the s3 object, would that be used by anything (downloading the file as that name)? If we really want that, seems we can do it when we presign the URL as well: https://stackoverflow.com/questions/2611432/amazon-s3-change-file-download-name#

mikehgrantsgov: Got it, makes sense. No need to add as metadata if we are already storing it in the database. Just wanted to make sure it wasn't being lost.

PR #5192, opened 2025-05-29
[Issue #4393] Backend Updates for migrating Opportunity tables to use UUID

joshtonava:
## Summary

Fixes #{4393}

## Changes proposed

This PR handles backend tasks and processes updates related to UUID migration. 
There were also some changes related to recent changes from main but for the most part this is for backend tasks.

## Context for reviewers

## Validation steps

All tests should pass

```bash
make test
```

### DIFF
--- a/api/src/data_migration/transformation/subtask/transform_opportunity_attachment.py
+++ b/api/src/data_migration/transformation/subtask/transform_opportunity_attachment.py
@@ -209,7 +209,11 @@ def transform_opportunity_attachment(
     if source_attachment.file_lob_size is None:
         raise ValueError("Opportunity attachment does not have a file size, cannot process.")
 
-    attachment_id = uuid.uuid4()
+    if incoming_attachment:
+        attachment_id = incoming_attachment.attachment_id
+    else:
+        attachment_id = uuid.uuid4()
+
     file_location = attachment_util.get_s3_attachment_path(
         file_name, attachment_id, opportunity, s3_config
     )
@@ -231,9 +235,6 @@ def transform_opportunity_attachment(
         updated_by=source_attachment.last_upd_id,
         legacy_folder_id=source_attachment.syn_att_folder_id,
     )
-
-    if incoming_attachment:
-        target_attachment.attachment_id = incoming_attachment.attachment_id
 
     transform_util.transform_update_create_timestamp(
         source_attachment, target_attachment, log_extra=log_extra


api/tests/src/data_migration/transformation/conftest.py
```diff
@@ -103,11 +102,12 @@ def setup_cfda(
     return source_cfda
 
 
+# TODO: Clean this up
```
joshtonava: I think I can resolve this issue by using the source_values instead of passing an opportunity

api/src/data_migration/transformation/subtask/transform_competition.py
```diff
@@ -41,15 +42,29 @@ def transform_records(self) -> None:
...
+                            OpportunityAssistanceListing.legacy_opportunity_assistance_listing_id
+                            == opportunity_cfda.opp_cfda_id
+                        )
+                    ).scalar_one_or_none()
+
+                    opportunity_assistance_listing_id = (
+                        opportunity_assistance_listing.opportunity_assistance_listing_id
+                        if opportunity_assistance_listing
+                        else None
+                    )
```
chouinar: I don't think we need to add this query - we just want to adjust the for-loop below to compare the assistance listing's legacy ID with the opp_cfda_id. This query is sort've an alternate to what we did, but this is already finnicky enough to get working

api/src/data_migration/transformation/subtask/transform_opportunity.py
```diff
@@ -105,7 +105,7 @@ def process_opportunity(
                         file_name = attachment_util.adjust_legacy_file_name(attachment.file_name)
                         s3_path = attachment_util.get_s3_attachment_path(
                             file_name,
-                            attachment.attachment_id,
+                            attachment.legacy_attachment_id,
```
chouinar: Could we make the calls to this function use the new UUIDs? It's fine to make it so the function takes in UUIDs. It does mean when we create the attachment record in the transformation process, we'll need to do `application_attachment_id=uuid.uuid4()` ourselves so its set.

api/src/data_migration/transformation/transform_util.py
```diff
@@ -104,14 +104,19 @@ def is_empty_str(value: str | None) -> bool:
 def transform_opportunity(
     source_opportunity: Topportunity, existing_opportunity: Opportunity | None
 ) -> Opportunity:
-    log_extra = {"opportunity_id": source_opportunity.opportunity_id}
+    log_extra = {
+        "legacy_opportunity_id": source_opportunity.opportunity_id,
+        "opportunity_id": existing_opportunity.opportunity_id if existing_opportunity else None,
```
chouinar, line 109: The source will always have the ID like it did before, so can keep this simpler
```suggestion
        "opportunity_id": source_opportunity.opportunity_id,
```

api/src/services/opportunity_attachments/attachment_util.py
```diff
@@ -27,7 +27,7 @@ def get_s3_attachment_path(
     return file_util.join(
         base_path,
         "opportunities",
-        str(opportunity.opportunity_id),
+        str(opportunity.legacy_opportunity_id),
```
chouinar: Could we have this be the opportunity_id? Thinking long-term, I'm not certain if we'll always want to have a legacy ID.

api/src/task/notifications/generate_notifications.py
```diff
@@ -157,9 +157,11 @@ def _collect_search_notifications(self) -> None:
 
         # For each unique query, check if results have changed
         for searches in query_map.values():
-            current_results: list[uuid.UUID] = search_opportunities_id(
+            current_results_str = search_opportunities_id(
                 self.search_client, searches[0].search_query
             )
+            # Convert string UUIDs to UUID objects for database compatibility
+            current_results = [uuid.UUID(uuid_str) for uuid_str in current_results_str]
```
chouinar: I see you had to do this in a few places, are we able to make `search_opportunities_id` return a list of UUIDs, moving this list generator into that function? If it breaks other things that might expect strings, feel free to ignore

api/tests/src/data_migration/transformation/conftest.py
```diff
@@ -920,9 +947,12 @@ def setup_competition(
...
             opportunity_id=opportunity.opportunity_id if opportunity else None,
-            opportunity_assistance_listing_id=(
-                opportunity_assistance_listing_id or competition.opp_cfda_id
-            ),
+            # opportunity_assistance_listing_id=(
+            #     opportunity_assistance_listing.opportunity_assistance_listing_id
+            #     if opportunity_assistance_listing
+            #     else competition.opportunity_assistance_listing_id
+            #     # DO I Need this??
+            # ),
```
chouinar: The competition's opportunity assistance listing is a complex mess because it works different between the systems. If we can set it, to show it can change, that'd be great, but I know it's a complex thing.

Context on what is going on with this.

In our system, an opportunity has a 1-many relationship with competitions. A competition then has an optional opportunity assistance listing record.

In the legacy system, an opportunity has a list of CFDA (what assistance listings used to be called) which might each have one or more competition. Their data model made the CFDA table pull double duty as recording CFDAs as well as a sorta link between opportunities and competitions. BUT because sometimes a competition didn't want a CFDA, they'd make empty CFDAs with none of the info set. Rather than repeat this, we simplified to the above model. This means that the transformations/setup is a gnarly mess as we're having to reference their structure to adapt to ours.

That's all to say, yeah, I get this might be a bit difficult to make work happily.

api/src/data_migration/transformation/subtask/transform_competition.py
joshtonava: I can try messing with this and report back here

joshtonava: Yea that makes sense and worked :)

api/src/data_migration/transformation/subtask/transform_opportunity.py
joshtonava: Yes we can take care of that here.

chouinar:
Did a bit of testing this morning. Backend tasks seemed good, found the following frontend things we need to follow-up on.

* Search loads fine, but clicking an opportunity takes you to the UUID page which doesn't work - frontend either needs to use the int OR allow the opportunity page to support UUID
* Saved opportunities do not work, it can't find an endpoint, probably is trying to save an int instead of UUID?
* Saved searches seems fine (frontend never used the opp ids here)
* Navigating to the saved search page seems to work and fetch saved searches, but like search itself the UUID links don't work.

api/src/services/opportunity_attachments/attachment_util.py
joshtonava: Yes this is done now :)

api/src/data_migration/transformation/transform_util.py
joshtonava: I updated the log to have a new key `legacy_opportunity_id` that holds`source_opportunity.opportunity_id`

since `existing_opportunity` is optional, I think we have to have the more complicated ternary for `opportunity_id`

Let me know if im misunderstanding

api/src/task/notifications/generate_notifications.py
joshtonava: Moving this to `search_opportunities_id` worked great :)

api/tests/src/data_migration/transformation/conftest.py
joshtonava: Thanks for the additional context.

After reviewing the transformation code, I can see that the actual transformation logic in `transform_competition.py` is correctly handling the `opportunity_assistance_listing_id` assignment. The field gets set when we find a matching assistance listing for the competition's CFDA record.


The commented-out code in the test setup was causing confusion for me as well because the tests were still passing without it. 

Im thinking we may want to circle back to this with a follow up ticket with a new test to show this behavior:

https://github.com/HHS/simpler-grants-gov/issues/5522

joshtonava:
> Did a bit of testing this morning. Backend tasks seemed good, found the following frontend things we need to follow-up on.
> 
> * Search loads fine, but clicking an opportunity takes you to the UUID page which doesn't work - frontend either needs to use the int OR allow the opportunity page to support UUID
> * Saved opportunities do not work, it can't find an endpoint, probably is trying to save an int instead of UUID?
> * Saved searches seems fine (frontend never used the opp ids here)
> * Navigating to the saved search page seems to work and fetch saved searches, but like search itself the UUID links don't work.

Perfect thanks for the feedback on the frontend @chouinar. I have created a ticket for this as well: 
https://github.com/HHS/simpler-grants-gov/issues/5523 

Ill work on this and reach out to @doug-s-nava for help as needed

api/src/data_migration/transformation/subtask/transform_opportunity_attachment.py
```diff
@@ -208,15 +209,17 @@ def transform_opportunity_attachment(
     if source_attachment.file_lob_size is None:
         raise ValueError("Opportunity attachment does not have a file size, cannot process.")
 
+    attachment_id = uuid.uuid4()
     file_location = attachment_util.get_s3_attachment_path(
-        file_name, source_attachment.syn_att_id, opportunity, s3_config
+        file_name, attachment_id, opportunity, s3_config
```
chouinar: Could we instead do:

```
if incoming_attachment:
   attachment_id = incoming_attachment.attachment_id
else:
   attachment_id = uuid.uuid4()
```
And remove the if-statement from lines 235-236.

I think as written here we'd end up putting updated attachments into a folder not matching their ID.

api/src/data_migration/transformation/transform_util.py
chouinar: No, you're right, I was complicating things.

api/src/db/migrations/versions/2025_05_02_add_application_status_table_and_field.py
chouinar, line 1: For the DB migration files, when we make the final merge-to-main PR, can we update the dates so they're roughly in the right order by filename?

chouinar (commented):
Just one small transform opportunity attachment adjustment and I think we're good to go

api/src/data_migration/transformation/subtask/transform_opportunity_attachment.py
joshtonava: good call!!

api/src/db/migrations/versions/2025_05_02_add_application_status_table_and_field.py
joshtonava: After renaming the two migrations to be more recent I had to adjust this migration. This migration is unrelated to my changes, my added migrations were right here before renaming them so I had to adjust the linking between files to make everything continue to work.

chouinar (approved):

PR #5201, opened 2025-05-29
[Issue #4986] Add an application form status to the response model for relevant endpoints

### LINKED ISSUE #4986: Calculate and return a "form status" for the frontend to display to a user
### Summary

After #5138 

Update the application form API schema to include a new `ApplicationFormStatus` str enum with the following values:

* `not_started`
* `in_progress`
* `complete`

This value needs to be calculated and added to the application form DB model (not as an actual column, just add the value to the object so the Marshmallow json logic can get it). The logic is simple:

Not started the form -> `not_started`
1+ validation issues form the form -> `in_progress`
No validation issues for the form -> `complete`

 if there are no validation issues, then the status is `complete`, otherwise it's `in_progress`.

See if we can put this logic into the validation logic itself - we'll want this to at the very least work in the following endpoints:
* GET /application/:appId
* GET /application/:appId/applicationForm/applicationFormId

### Acceptance criteria

- [ ] Update the application form schema to include this new enum
- [ ] Update the logic to set this field when validation is done

chouinar:
## Summary

Fixes #4986

## Changes proposed
Calculate an applicant form status and return it in:
* GET /applications/:appId
* PUT /applications/:appId/form/:formId
* GET /applications/:appId/application_forms/:appFormId

## Context for reviewers
There is still a bit of ongoing discussion with product/design about the exact way we calculate the form status, this might shift a bit, but that's not a blocker for getting something out.

The statuses are meant to help guide the user in filling out forms, even if every form has an application status of "complete" the app could still be rejected on submission for validations outside of the forms (eg. the submission window has closed).


api/src/api/application_alpha/application_route.py
```diff
@@ -104,13 +104,11 @@ def application_form_update(
...
         # Call the service to update the application form
-        _, warnings = update_application_form(
+        application_form, warnings = update_application_form(
             db_session, application_id, form_id, application_response, user
         )
 
-    return response.ApiResponse(
-        message="Success", data={"application_id": application_id}, warnings=warnings
-    )
+    return response.ApiResponse(message="Success", data=application_form, warnings=warnings)
```
mikehgrantsgov, line 111: ApplicationFormUpdateResponseDataSchema looks like it just has `application_form_status` and `application_id`. Passing the whole `application_form` to the response would strip out any non-schema fields?

mikehgrantsgov (approved):
Looks good, I was just confused returning the whole `application_form` in the update response but I don't think it's an issue.

api/src/api/application_alpha/application_route.py
chouinar: Yep, this is fine. If a field isn't defined in our Marshmallow schema, it won't be pulled out and put in the JSON response.

Under the hood the Marshmallow code basically loops over defined fields and does `getattr(object, "field_name")` on everything defined, if there is anything extra it's just ignored. That's why some of our DB schemas can have more stuff than what a user sees and not expose it (useful for internal metadata fields)

PR #5214, opened 2025-05-30
[Issue #5059] redirect users on protected routes using x-redirectOnLogout header

### LINKED ISSUE #5059: Redirect users who intentionally signs out from auth-dependent pages

GlennTatum:
## Summary

Fixes #5059 

## Changes proposed

This PR adds a `x-redirectOnLogout` header to `next.config.js` to routes which content are protected.

## Context for reviewers

The precedent to this pull request is #5074 which used url search parameters as an alternative to being able to log out users gracefully back to the homepage when authenticated. 

However, #5074 incurs the issue of being hacky as if `useSearchParams()` is called, a `<Suspense />` boundary **must** be placed around it ([https://nextjs.org/docs/messages/missing-suspense-with-csr-bailout](https://nextjs.org/docs/messages/missing-suspense-with-csr-bailout)) which for handling authentication hinders the UX with a temporary loading page on logouts [feedback](https://github.com/HHS/simpler-grants-gov/pull/5074#discussion_r2116064542).

As a solution to determining whether or not a route (or a series of routes that fall under a path name) are authenticated routes that desire the behavior of logging out back to the homepage, the `x-redirectOnLogout` header is used. By using the header instead of a search param or modified anonymous session token this prevents any unwanted mutation of the current user session or window location [feedback](https://github.com/HHS/simpler-grants-gov/pull/5074#discussion_r2116058548).

If any route wants the desired behavior of being able to log back out to the homepage the only change they must add to their route is adding the `x-redirectOnLogout` header in `next.config.js`. This allows for future authenticated pages to opt-out of the `redirectOnLogout` behavior if it is not necessary but still preserves the default behavior of refreshing the page and displaying the Need to Sign In Call To Action (CTA) if the headers is not applied.

## Validation steps
headers-auth-control

### DIFF
--- a/frontend/src/components/user/UserControl.tsx
+++ b/frontend/src/components/user/UserControl.tsx
@@ -4,7 +4,7 @@ import { useUser } from "src/services/auth/useUser";
 import { UserProfile } from "src/types/authTypes";
 
 import { useTranslations } from "next-intl";
-import { useRouter } from "next/navigation";
+import { redirect, usePathname, useRouter } from "next/navigation";
 import { useCallback, useState } from "react";
 import {
   IconListContent,
@@ -124,6 +124,7 @@ export const UserControl = () => {
 
   const { user, logoutLocalUser } = useUser();
   const router = useRouter();
+  const pathname = usePathname();
 
   const logout = useCallback(async (): Promise<void> => {
     // this isn't using the clientFetch hook because we don't really need all that added functionality here
@@ -131,9 +132,18 @@ export const UserControl = () => {
       method: "POST",
     });
 
+    const res = await fetch(pathname, {
+      method: "GET",
+    });
+
     logoutLocalUser();
-    router.refresh();
-  }, [logoutLocalUser, router]);
+
+    if (res.headers.has("x-redirectOnLogout")) {
+      redirect("/");
+    } else {
+      router.refresh(); // call router.refresh() before so we can grab the header in AuthenticationGate as 1 and redirect while also logging out
+    }
+  }, [logoutLocalUser, router, pathname]);
 
   return (
     <>
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -115,6 +115,10 @@ const headers = [
         key: "Cache-Control",
         value: "no-store, must-revalidate",
       },
+      {
+        key: "x-redirectOnLogout",
+        value: "1",
+      },
     ],
   },
   // don't cache if users has a session cookie


doug-s-nava (commented):
This is a cool idea, but I am not a big fan of making an extra network request to the page in order to grab a header value. I think we have two paths forward here:

* instead of adding a header on responses to pages that should redirect on logout, we store this information in a hard coded variable that the application can reference from both the client and server. On logout, we check the current pathname against the hardcoded list, and if present, we redirect.
* we go back to your original solution. Now that I understand why the Suspense was necessary there (Next being inscrutable as usual), I think that solution can work, but I think a better user experience would be supplying the original page (the original `children` prop in the AuthenticationGate) as the fallback UI rather than a loading message. With that change I think your last implementation could work, but we I think we may be working toward a more elegant solution here in any case

GlennTatum:
I also agree with that doing another HTTP request just to grab a header value is a bit weird. I wish NextJS had an interceptor to automatically populate the Response Headers in some sort of context. In the most recent change, I switched the HTTP Method to `HEAD` to prevent downloading the entire page which only retrieves the Response headers from the path instead of downloading an entire page.

In terms of other solutions though, here are the potential pathways that are being discussed:

**Hard Coded Values**
Hard coded values could be represented as a list like:
```typescript
export const redirectOnLogout = [
  '/saved:path*',
  ...
]
```

and the `UserControl` callback would reference them.

It would be an equal amount of toil to the developer on adding their route to this list as it would be adding the header to `next.config.js`. 

So this would be a good tradeoff if we want to head for a static rather than dynamic solution going forward in terms of being able to strategically determine how to automatically redirect users upon logging out from protected routes.

**URL Parameters**

The primary reason for the new PR was to solve the issue of populating the url parameters, as the solution was 100% dependent upon them being present.
> I'm wary of adding this query param on all logouts, rather than only when we're logging out from a gated page. I can't reproduce any weird behavior based on this at the moment, but adding an unused query param just seems like bad practice and could come back to bite us later. [Reply](https://github.com/HHS/simpler-grants-gov/pull/5074/files#r2116058548)

As a solution, this was an option which presented itself to be completely dynamic in nature. As it would tack on the url parameter at the time of calling logout where the `AuthenticationGate` would then determine how to redirect from there. This prevents the need to manually add on any paths or routes.

However, when using `useSearchParams()` and running `next build` in CI, the CSR Bailout error began to pop up on other unrelated pages (Reference) [1](https://github.com/HHS/simpler-grants-gov/actions/runs/15221185635/job/42816910866#step:6:88), [2](https://github.com/HHS/simpler-grants-gov/actions/runs/15334230628/job/43147991216#step:6:102). 

I'm not sure if this was a side-effect of adding `useSearchParams()` to a client side component inside of a server component causing the compilation to start having erroneous builds.

**Summary**

Overall, the hard coded solution is looking better. Primarily because if in the future, a path doesn't need the `redirectOnLogout` capability anymore, all that needs to be done is to remove the path from the list.

Having to add a `<Suspense />` boundary causes odd behavior and potentially raises accessibility issues on the client, being that a loading screen has to briefly pop up, and then the client gets redirected back to the home page.

Since the start of the original PR, finding a non-hard coded solution was preferred. However, looking at it now, if we don't want to completely overhaul `AuthenticationGate` the hard coded solution looks more reliable.

btabaska:
@doug-s-nava What is the status on this PR? I see a thumbs up on Glenns last comment, but unsure if this is the direction you want to go.

doug-s-nava:
@btabaska yeah, I am on board with the idea we landed on here to hardcode a list of paths that we want to redirect away from on logout

GlennTatum:
Hello, sorry for the untimely response and inactivity on this PR.

If were all go ahead for the hardcoded list of protected paths that sounds great!

Will be on the task very soon. :+1:

GlennTatum:
It would be preferable to not have [frontend/src/components/user/redirectLogout.ts](https://github.com/HHS/simpler-grants-gov/pull/5214/files#diff-c7fa41f2558e32907291f972e93d59652ab781167b7e4ada9cf31420ce67c9d0) under the `components/` directory as it is more of a configuration directive. Any suggestions on if it should go anywhere else within the folder structure? Or should it just remain where it is?

frontend/src/components/user/redirectLogout.ts
```diff
@@ -0,0 +1 @@
+export const redirectLogout = ["/saved*"];
```
doug-s-nava: a good place for this might be src/constants/routes

would you mind renaming to `redirectLogoutPaths`?

frontend/src/components/user/UserControl.tsx
```diff
@@ -124,16 +125,31 @@ export const UserControl = () => {
...
 
   const logout = useCallback(async (): Promise<void> => {
     // this isn't using the clientFetch hook because we don't really need all that added functionality here
     await fetch("/api/auth/logout", {
       method: "POST",
     });
 
+    let redirectToLogout = false;
+    redirectLogout.forEach((path) => {
+      const base = path.substring(1); // remove the forward-trailing '/' for the basepath to have a valid regex
```
doug-s-nava: would it make more sense to store and reference regexs directly rather than having to convert from strings? 

Another alternative would be to just do direct string matches, but that would fail if we ever want to include dynamic paths here, which we probably will down the line.

frontend/src/components/user/UserControl.tsx
```diff
@@ -124,16 +125,31 @@ export const UserControl = () => {
...
+  const pathname = usePathname();
 
   const logout = useCallback(async (): Promise<void> => {
     // this isn't using the clientFetch hook because we don't really need all that added functionality here
     await fetch("/api/auth/logout", {
       method: "POST",
     });
 
+    let redirectToLogout = false;
+    redirectLogout.forEach((path) => {
```
doug-s-nava: I think this can be simplified a bit, how do you feel about

```suggestion
const redirectToLogout = redirectLogoutPaths.some((pathRegexp) => pathRegexp.test(pathname))
```

frontend/src/components/user/UserControl.tsx
GlennTatum: I was looking to take a hybrid approach in between what appears in: https://github.com/HHS/simpler-grants-gov/blob/757992cd984e0813187f0a76912cc8ac8be98033/frontend/next.config.js#L112

But using the NextJS internal router url parsing functions  would be too complex for a task such as matching path names. For example, having `:path` within the string. So only a valid regular expression is used `saved*` instead of `saved:path*`.

In regards to how the patterns that should be stored, using Regex objects directly (which is what I assume is being asked for) is an option.

Instead of this:
```ts
export const redirectLogout = [
    "/saved*"
]
```
Do this:
```ts
export const redirectLogoutPaths = [
   new RegExp("saved*")
]
```

Overall this makes it more abundantly clear that this matching is dependent on Regex, and all that must be done is write a proper expression that matches with the path. All that would be changed is that instead of converting the string, `re.test(exp)` would be used immediately.

frontend/src/components/user/UserControl.tsx
GlennTatum: Much better, I was indifferent about using forEach since it would just keep on going even if a match was made to the end of the array.

frontend/src/constants/redirectLogout.ts
GlennTatum, line 1: Is there any preference on using an escaped slash to capture the `/` in the path name?

Example: `/saved-opportunities`

```ts
/saved*/  ✅
```
```ts
/\/saved*/ ✅
```

Both match on `/saved-opportunities` but it would probably be better to catch the forward trailing slash.

btabaska:
This has been open for a few months and is now stale. Closing. Thanks so much for you help with these code updates.

PR #5229, opened 2025-06-02
[Issue #5137] Return the form instructions in the form response object

### LINKED ISSUE #5137: Return the form instructions in the form response object
### Summary

After #5136 

Similar to #5134 - we want to add a `form_instruction` object to the form API schema which has:
* `file_name` - str
* `download_path` - str - see note below
* `created_at` - datetime
* `updated_at` - datetime

We should follow what we implemented in that ticket for this one, using the same approach for getting the s3 download path set on the response object.

### Acceptance criteria

- [ ] Add the property for the file location to the form instruction DB model
- [ ] Add tests for verifying calling the GET /form endpoint gets the instruction object set

mikehgrantsgov:
## Summary

Fixes #5137

## Changes proposed

Add `FormInstruction` schema and return with the `Form`
Move `CDNConfig` to more centralized place
Add `presign_or_s3_cdnify_url` which can be referenced from `FormInstruction` and future models as properties
Add tests

## Context for reviewers

We modified the way we do this presigning/s3 stuff as needing to iterate over records to do it is going to get very tedious/error prone as more and more endpoints need this.

## Validation steps

See additional unit tests.


api/src/services/opportunities_v1/get_opportunity.py
```diff
@@ -59,7 +53,7 @@ def pre_sign_opportunity_file_location(
 def get_opportunity(db_session: db.Session, opportunity_id: int) -> Opportunity:
     opportunity = _fetch_opportunity(db_session, opportunity_id)
 
-    attachment_config = AttachmentConfig()
+    attachment_config = CDNConfig()
```
chouinar: Could we adjust the opportunity attachment class to also use this same property function? That way we don't have to iterate over it to set it, the property itself handles it letting us delete the below code.

api/src/util/file_util.py
```diff
@@ -198,3 +203,16 @@ def convert_public_s3_to_cdn_url(file_path: str, cdn_url: str, s3_config: S3Conf
...
 
     return file_path.replace(s3_config.public_files_bucket_path, cdn_url)
+
+
+def presign_or_s3_cdnify_url(file_path: str) -> str:
+    """
+    Generates a URL for file download, either using CDN or pre-signed URL.
+    """
+    cdn_config = CDNConfig()
+    s3_config = S3Config()
```
chouinar: As a small optimization, could we instead have these be something like:

```suggestion
def presign_or_s3_cdnify_url(file_path: str, s3_config: S3Config | None) -> str:
    if s3_config is None:
        s3_config = get_s3_config()
```

The function would then look something like:
```py
_s3_config = None

def get_s3_config() -> S3Config:
    global _s3_config
    if _s3_config is None:
        _s3_config = S3Config()

    return _s3_config
```

This keeps us from reloading the env vars every time. We also let someone pass in an s3 config for making unit tests a bit easier to test (otherwise they'd have a static config that is more annoying to adjust).

api/src/util/file_util.py
```diff
@@ -10,12 +10,17 @@
...
 from src.adapters.aws import S3Config, get_boto_session, get_s3_client
+from src.util.env_config import PydanticBaseEnvConfig
 
 ##################################
 # Path parsing utils
 ##################################
 
 
+class CDNConfig(PydanticBaseEnvConfig):
+    cdn_url: str | None = None
```
chouinar: Could we put this in the s3 config? Only ever used in the context of s3

api/src/db/models/competition_models.py
```diff
@@ -147,6 +147,12 @@ class FormInstruction(ApiSchemaTable, TimestampMixin):
     file_location: Mapped[str]
     file_name: Mapped[str]
 
+    @property
+    def download_path(self) -> str:
+        from src.util.file_util import presign_or_s3_cdnify_url
```
chouinar: Can the import go at the top of the file?

mikehgrantsgov: Updated this

api/src/util/file_util.py
mikehgrantsgov: Makes sense, updated this too

api/src/services/opportunities_v1/get_opportunity.py
mikehgrantsgov: Made this change, and it's much cleaner now.

api/src/db/models/competition_models.py
```diff
@@ -291,6 +296,10 @@ class ApplicationAttachment(ApiSchemaTable, TimestampMixin):
     mime_type: Mapped[str]
     file_size_bytes: Mapped[int] = mapped_column(BigInteger)
 
+    @property
+    def download_path(self) -> str:
+        return presign_or_s3_cdnify_url(self.file_location)
```
chouinar: For this table, I added an equivalent in the PR I sent out earlier today, I'd exclude it here as it actually needs to be presign-only because the attachments won't ever be on the public s3 bucket.

api/src/services/extracts_v1/get_extracts.py
```diff
@@ -78,8 +79,9 @@ def get_extracts(
     extracts = paginator.page_at(page_offset=list_params.pagination.page_offset)
     pagination_info = PaginationInfo.from_pagination_params(list_params.pagination, paginator)
 
+    s3_config = S3Config()
     for extract in extracts:
         file_loc = extract.file_path
-        setattr(extract, "download_path", pre_sign_file_location(file_loc))  # noqa: B010
+        setattr(extract, "download_path", pre_sign_file_location(file_loc, s3_config))  # noqa: B010
```
chouinar: Could we make the extracts also use the presign/cdnify approach?

mikehgrantsgov: Added this

api/src/db/models/competition_models.py
mikehgrantsgov: Got it, I removed this.

api/tests/src/api/extracts_v1/test_extract_schema.py
```diff
@@ -54,15 +55,20 @@ def test_request_schema_validation():
...
     schema = ExtractMetadataResponseSchema()
 
-    sample_extract_metadata.download_path = "http://www.example.com"
-    extract_metadata = schema.dump(sample_extract_metadata)
+    # Mock the download_path property
+    with mock.patch(
+        "src.db.models.extract_models.ExtractMetadata.download_path",
+        new_callable=mock.PropertyMock,
+        return_value="http://www.example.com",
+    ):
```
chouinar: Missed this in earlier revisions / saw it in your follow-up PR. If we are mocking this behavior out, then are we certain that it works? Mocking this means we're not testing that the property itself works - is there a reason for doing it this way?

We likely need to adjust the extract metadata used here to have a valid s3 path so we can test it

mikehgrantsgov: Ah yeah, I added some tests that cover the s3 and CDN scenarios

api/src/util/file_util.py
```diff
@@ -83,8 +83,10 @@ def open_stream(
         return smart_open.open(path, mode, encoding=encoding)
 
 
-def pre_sign_file_location(file_path: str) -> str:
-    s3_config = S3Config()
+def pre_sign_file_location(file_path: str, s3_config: S3Config | None = None) -> str:
+    if s3_config is None:
```
mikehgrantsgov, line 100: @chouinar I don't want to read env vars every time this is called if no s3_config is passed in (from competition_models`ApplicationAttachment`). Is there a better way to do this?

chouinar: An earlier comment I had put this snippet:

```py
_s3_config = None

def get_s3_config() -> S3Config:
    global _s3_config
    if _s3_config is None:
        _s3_config = S3Config()

    return _s3_config
```

Then you do:
```py
if s3_config is None:
    s3_config = get_s3_config()
```
Effectively, we only initialize / fetch the env vars once ever globally, and reuse that. This can sometimes make the unit tests a bit funky if you want to change the values, but that's why we pass in an optional config to the file so we can test those functions directly with a manually defined config.

mikehgrantsgov: Makes sense. It's just that if we call the `download_path` property, it will reference the global s3 config without giving us the opportunity in the unit test to override it at the param level. Let me fiddle with this.

PR #5237, opened 2025-06-03
[Issue #5134] Add the competition instruction object to the competition response

### LINKED ISSUE #5134: Add the competition instruction object to the competition response object
### Summary

After #5133 created a competition instruction table, we want to add that to the competition API response

We want to modify the competition API schema to include a `competition_instruction` object which should include:
* `file_name` - str
* `download_path` - str - see note below
* `created_at` - datetime
* `updated_at` - datetime

The download_path should be a path on our s3 CDN based on the `file_location` value from the competition instruction DB object. Similar to what we do with setting the s3 cdn/presigned urls for opportunity attachments in GET /opportunities. Ideally we should move that logic to somewhere more generic.

We should modify the way we do this presigning/s3 stuff as needing to iterate over records to do it is going to get very tedious/error prone as more and more endpoints need this. We should instead consider putting it into a property like so on the competition instruction table:
```py
@property
def download_path(self) -> str:
    return presign_or_s3_cdnify_url(self.file_location)
```
Where that function is the logic that we currently use to make an s3/cdn path for download.

### Acceptance criteria

- [ ] Move the logic to presign/s3 CDNify a file's location to somewhere more central
- [ ] Modify the usage of the presigning/CDN url generation to be a property (including on the opportunity attachment)
- [ ] For each of the endpoints that gets a competition, make sure the competition instruction response object has the instructions presigned/CDN converted
- [ ] Add tests for each of these endpoints

mikehgrantsgov:
## Summary

Fixes #5134

## Changes proposed

Add `CompetitionInstructionAlphaSchema`
Add `competition_instructions` to `CompetitionAlphaSchema`
Add `download_path` to `CompetitionInstruction`
Add units tests

## Context for reviewers

Since the majority of this was done in 5137, putting up a PR to merge into that branch since that's not yet in `main`.

## Validation steps

See new unit tests


api/tests/src/api/competition_alpha/test_competition_route_get.py
```diff
@@ -34,6 +35,66 @@ def test_competition_get_200_with_api_key(client, api_auth_token, enable_factory
...
+    # Mock the download_path property for CompetitionInstruction
+    presigned_url = "https://example.com/competition-instructions/file.pdf"
+    with mock.patch(
+        "src.db.models.competition_models.CompetitionInstruction.download_path",
+        new_callable=mock.PropertyMock,
+        return_value=presigned_url,
+    ):
+        resp = client.get(
+            f"/alpha/competitions/{competition.competition_id}", headers={"X-Auth": api_auth_token}
+        )
```
chouinar: Instead of patching this - could we just rely on the actual logic / check the file exists?

With how the factory is setup, the instruction object should be downloadable without any additional work, we might pass in file_contents to the competition instruction factory and check the contents of the file we download to be safe.

mikehgrantsgov: I used `with_instruction=True` param in the factory and tested that we got some data back. Would it be better to explicitly create the instruction and check the content? Or is that done in another test?

api/tests/src/api/competition_alpha/test_competition_route_get.py
```diff
@@ -35,64 +35,96 @@ def test_competition_get_200_with_api_key(client, api_auth_token, enable_factory
     assert response_competition["open_to_applicants"] == competition.open_to_applicants
 
 
-def test_competition_get_with_instructions_200(client, api_auth_token, enable_factory_create):
-    # Create a competition with instructions
+def test_competition_get_with_instructions_200(
+    client, api_auth_token, enable_factory_create, monkeypatch_session, mock_s3_bucket
```
chouinar: We don't want to use `monkeypatch_session` on an individual test, this changes the behavior for every single test that runs after it.

If we wanted to have the things you're setting be the defaults for all tests, that could be fine, but doing it this way is going to have weird side effects between tests.

api/tests/src/api/competition_alpha/test_competition_route_get.py
```diff
@@ -35,64 +35,96 @@ def test_competition_get_200_with_api_key(client, api_auth_token, enable_factory
...
+):
+    monkeypatch_session.setattr(file_util, "_s3_config", None)
+
+    # Set environment variables for S3 configuration
+    monkeypatch_session.setenv("CDN_URL", "")  # Empty string to ensure no CDN is used
+    monkeypatch_session.setenv("S3_ENDPOINT_URL", "http://localstack:4566")
```
chouinar: Don't set this to localstack - we don't use localstack in unit tests - we should never set this value to anything but null in our unit tests. In `reset_aws_env_vars` we intentionally delete this to avoid ever using localstack.

I see I missed that you had done that on your prior PR, we need to undo that.

api/tests/src/api/competition_alpha/test_competition_route_get.py
```diff
@@ -35,64 +35,96 @@ def test_competition_get_200_with_api_key(client, api_auth_token, enable_factory
     assert response_competition["open_to_applicants"] == competition.open_to_applicants
 
 
-def test_competition_get_with_instructions_200(client, api_auth_token, enable_factory_create):
-    # Create a competition with instructions
+def test_competition_get_with_instructions_200(
```
chouinar, line 38: Here is my suggestion for how to simplify things / make presigned URLs the default locally:
* In the `reset_aws_env_vars` fixture, add `monkeypatch.delenv("CDN_URL")` to the end. This will make it so everything is using presigning by default, and we shouldn't need any special setup to fetch/work with these files in our tests
* In tests where we are testing that the CDNify approach works - set this env var using `monkeypatch` (not the session one)
* Go through tests, remove any usage of `monkeypatch_session` from individual tests

api/tests/src/api/competition_alpha/test_competition_route_get.py
```diff
@@ -34,6 +35,96 @@ def test_competition_get_200_with_api_key(client, api_auth_token, enable_factory
...
 
 
+def test_competition_get_with_instructions_200(
+    client, api_auth_token, enable_factory_create, monkeypatch, mock_s3_bucket
+):
+    monkeypatch.setattr(file_util, "_s3_config", None)
+
+    # Set environment variables for S3 configuration
+    monkeypatch.setenv("PUBLIC_FILES_BUCKET", f"s3://{mock_s3_bucket}")
+    monkeypatch.delenv("CDN_URL")
```
chouinar: If you add `monkeypatch.delenv("CDN_URL")` to the `reset_aws_env_vars` fixture in the conftest file, I think we can remove all of these.

api/tests/src/api/competition_alpha/test_competition_route_get.py
mikehgrantsgov: Made these changes and cleaned up a lot of manual setenv's

chouinar (approved):

PR #5264, opened 2025-06-06
[Issue #4842] Setup a process for handling pre/post population and custom validations on an application form

### LINKED ISSUE #4842: Setup a utility that given a form, will handle pre and post population of fields on a
### Summary

After #4838 

TODO - add details for how this will work

Roughly should have a function like:
```
def auto_populate_fields(application_form: ApplicationForm, population_method: <lookup enum added in prior PR>) -> None:
     ... 
```
This method will first check in the form if it has any pre/post population fields, if it does, we'll then iterate over them and handle setting values.

When we get to that point in the logic, we'll just do a set of if statements like so:

```
if form_field_type == FormFieldType.OPPORTUNITY_NUMBER:
     # Probably a much better way to pull this out, but use the relationships
     value = application_form.application.competition.opportunity.opportunity_number
```

We also need to account for setting the value in the correct path which needs some more thought.

### Acceptance criteria

- [ ] Utility function built
- [ ] Thorough testing added that verifies all field mapping types work
- [ ] Various scenarios for the state of the existing data (already exists, already set, completely null/empty) including with nested path structures.

chouinar:
## Summary

Fixes #4842

## Changes proposed
Create a utility that given an application form, runs pre-population, post-population, and validation rules on it based on a configuration.

## Context for reviewers
I added a README that explains some of this, let me know if there are any other questions you have or more detail I should add. The README also details some things not currently handled in this iteration (whether we need to handle some of the later ideas depends on the forms we need to support - so unknown for now).

The recursive configuration approach means adding new rules only requires changes to the mappers we have defined + a function to handle it. For most field population that should be very simple. Validation will depend on the scenario.

## Validation steps
Mainly the tests at the moment, this isn't yet hooked up to any API or process outside the tests yet.


freddieyebra:
this looks awesome! There is great re-usable functionality here. I do wonder; and would be happy to discuss async, if there would/could be any use cases for this functionality of other forms not related to opportunities/competitions that could leverage the `form_schema` module.

An idea could be to create a separate table `api.form_configs` with `id,form_ui_schema,form_json_schema` etc. that could be referenced by id from other forms for handling form rules in other cases, and could potentially configured by admin tool of some sort.

chouinar:
> this looks awesome! There is great re-usable functionality here. I do wonder; and would be happy to discuss async, if there would/could be any use cases for this functionality of other forms not related to opportunities/competitions that could leverage the `form_schema` module.
> 
> An idea could be to create a separate table `api.form_configs` with `id,form_ui_schema,form_json_schema` etc. that could be referenced by id from other forms for handling form rules in other cases, and could potentially configured by admin tool of some sort.

The form table is effectively just "form configs", so I'd leave it either connected to / or part of the form table as we have it here (note that I do think in the future we might want flexibility to override configs for a given competition, but that could go in competition_form).

I did originally try to make this a bit more generic and not need to be effectively tied to an application form object, but since one of the underlying purposes of this is fetching values out of the application/competition/opportunity, that proved tricky. We could potentially swap out the mappers + container class for that.

At the moment, I'm not familiar with any cases for populating form-related JSON outside of application forms. In the current grants.gov forms are only used in two contexts:
* People applying
* Some sort of post-award form process - https://grants.gov/forms/forms-repository/post-award-reporting-forms
I know very little about the latter, haven't even looked into it, might not need to be accounted for at all.

api/tests/src/search/backend/test_load_agencies_to_index.py
```diff
@@ -33,7 +33,7 @@ def test_load_agencies_to_index(
         enable_factory_create,
     ):
         # Create Agencies to load into the search index
-        agencies = [AgencyFactory.create(agency_code="DOD")]
+        agencies = [AgencyFactory.create(agency_code="LOADAGENCY1")]
```
chouinar, line 36: These tests started failing because the agency code iterator for the agency factory just happened to be making the same agency code and breaking things. Changed the agency code in the tests to values not in the iterator so they won't clash.

api/src/form_schema/rule_processing/json_rule_validator.py
```diff
@@ -0,0 +1,69 @@
...
+    if isinstance(value, str):
+        if value not in application_attachment_ids:
+            context.validation_issues.append(
+                ValidationErrorDetail(
+                    type=ValidationErrorType.UNKNOWN_APPLICATION_ATTACHMENT,
+                    message="Field references application_attachment_id not on the application",
+                    field=build_path_str(path),
+                    value=value,
+                )
+            )
```
freddieyebra:
```suggestion
    if isinstance(value, str) value not in application_attachment_ids:
        context.validation_issues.append(
            ValidationErrorDetail(
                type=ValidationErrorType.UNKNOWN_APPLICATION_ATTACHMENT,
                message="Field references application_attachment_id not on the application",
                field=build_path_str(path),
                value=value,
            )
        )
```

api/src/form_schema/rule_processing/json_rule_util.py
```diff
@@ -0,0 +1,84 @@
...
+def build_path_str(path: list[str]) -> str:
+    """Build a path string for validation issues
+
+    We prepend $. on the front to mimic the format of
+    the JSON Schema validation
+    """
+    return ".".join(["$"] + path)
+
+
+def get_nested_value(data: dict, path: list[str]) -> Any:
```
freddieyebra, line 13: how is the `path` param here actually built?

api/src/form_schema/rule_processing/json_rule_field_population.py
```diff
@@ -0,0 +1,82 @@
...
+
+def get_agency_name(context: JsonRuleContext) -> str | None:
+    if context.opportunity.agency_name is None:
+        return context.opportunity.agency_code
+
+    return context.opportunity.agency_name
+
+
+def get_current_date(context: JsonRuleContext) -> str:
+    return get_now_us_eastern_date().isoformat()
```
freddieyebra, line 27: do we need the get_current_date here?

If not we could consider something like 
```suggestion
def get_current_date(_: JsonRuleContext) -> str:
    return get_now_us_eastern_date().isoformat()
```

freddieyebra: ah i see the pattern and why it is needed :thumbsup:

api/src/form_schema/rule_processing/json_rule_util.py
chouinar: When we call `_process_rule_schema(context=context, rule_schema=v, path=path + [k])` this gradually builds the path list up as we recursively iterate down the json configuration.

freddieyebra (dismissed):
looks good!

freddieyebra (dismissed):

freddieyebra (approved):

PR #5269, opened 2025-06-06
[Issue #4994] Better handling of Feature Flag Defaults

### LINKED ISSUE #4994: Allow env vars to take precedence over client side cookie when no feature flag client overrides are set

mdragon:
## Summary

Fixes #4994

## Changes proposed

1. We need to separate out the FF defaults from the values the user has changed themselves so when we update the defaults we don't have old values stuck in folks' cookies. 
2. As part of that work we're standardizing on only using the middleware to set the FF cookie values, and it will check to make sure the value being "set" isn't the current default (it will just remove that flag instead).
3. Instead of trying to have the FF page "react" to a user's changes immediately we pop a message warning them they'll need to refresh the page to apply the FF changes. 
   - This is because while we could make this features buttons and content react to the FF changes, the rest of the site (including the Nav) won't reflect those changes reliably until the page is refreshed.

## Context for reviewers
This is not an end-user facing page so we leaned into simplicity not perfect UX or seamless rendering.

## Validation steps
1. Check out the branch
2. Visit http://localhost:3000/dev/feature-flags
3. Validate that now there are two columns for each feature, noting the default and the current setting. This allows for easy scanning for differences when trying to test features.
4. Click to enable/disable a feature flag.
    - This should pop up the warning message that you need to refresh the page to fully apply the flags you're changing

image

### DIFF
--- a/frontend/src/components/Layout.tsx
+++ b/frontend/src/components/Layout.tsx
@@ -1,4 +1,3 @@
-import { CookiesNextProvider } from "cookies-next";
 import { defaultFeatureFlags } from "src/constants/defaultFeatureFlags";
 import UserProvider from "src/services/auth/UserProvider";
 
@@ -20,20 +19,18 @@ export default function Layout({ children, locale }: Props) {
   const t = useTranslations();
 
   return (
-    <CookiesNextProvider pollingOptions={{ enabled: true, intervalMs: 1000 }}>
-      <UserProvider featureFlagDefaults={defaultFeatureFlags}>
-        <div className="display-flex flex-column minh-viewport">
-          <a className="usa-skipnav" href="#main-content">
-            {t("Layout.skipToMain")}
-          </a>
-          <Header locale={locale} />
-          <main id="main-content" className="border-top-0">
-            {children}
-          </main>
-          <Footer />
-          <GrantsIdentifier />
-        </div>
-      </UserProvider>
-    </CookiesNextProvider>
+    <UserProvider featureFlagDefaults={defaultFeatureFlags}>
+      <div className="display-flex flex-column minh-viewport">
+        <a className="usa-skipnav" href="#main-content">
+          {t("Layout.skipToMain")}
+        </a>
+        <Header locale={locale} />
+        <main id="main-content" className="border-top-0">
+          {children}
+        </main>
+        <Footer />
+        <GrantsIdentifier />
+      </div>
+    </UserProvider>
   );
 }
--- a/frontend/src/components/Layout.tsx
+++ b/frontend/src/components/Layout.tsx
@@ -1,3 +1,5 @@
+import { CookiesNextProvider } from "cookies-next";
+import { defaultFeatureFlags } from "src/constants/defaultFeatureFlags";
 import UserProvider from "src/services/auth/UserProvider";
 
 import { useTranslations } from "next-intl";
@@ -18,18 +20,20 @@ export default function Layout({ children, locale }: Props) {
   const t = useTranslations();
 
   return (
-    <UserProvider>
-      <div className="display-flex flex-column minh-viewport">
-        <a className="usa-skipnav" href="#main-content">
-          {t("Layout.skipToMain")}
-        </a>
-        <Header locale={locale} />
-        <main id="main-content" className="border-top-0">
-          {children}
-        </main>
-        <Footer />
-        <GrantsIdentifier />
-      </div>
-    </UserProvider>
+    <CookiesNextProvider pollingOptions={{ enabled: true, intervalMs: 1000 }}>
+      <UserProvider featureFlagDefaults={defaultFeatureFlags}>
+        <div className="display-flex flex-column minh-viewport">
+          <a className="usa-skipnav" href="#main-content">
+            {t("Layout.skipToMain")}
+          </a>
+          <Header locale={locale} />
+          <main id="main-content" className="border-top-0">
+            {children}
+          </main>
+          <Footer />
+          <GrantsIdentifier />
+        </div>
+      </UserProvider>
+    </CookiesNextProvider>
   );
 }
--- a/frontend/src/components/dev/FeatureFlagsTable.tsx
+++ b/frontend/src/components/dev/FeatureFlagsTable.tsx
@@ -1,22 +1,38 @@
 "use client";
 
+import { defaultFeatureFlags } from "src/constants/defaultFeatureFlags";
 import { useFeatureFlags } from "src/hooks/useFeatureFlags";
 
-import React from "react";
+import React, { useState } from "react";
 import { Button, Table } from "@trussworks/react-uswds";
+
+import BetaAlert from "src/components/BetaAlert";
 
 /**
  * View for managing feature flags
  */
 export default function FeatureFlagsTable() {
   const { setFeatureFlag, featureFlags } = useFeatureFlags();
+  const [pendingChanges, setPendingChanges] = useState(false);
+
+  const handleFeatureChange = (flagName: string, value: boolean) => {
+    setPendingChanges(true);
+    setFeatureFlag(flagName, value);
+  };
 
   return (
     <>
+      {pendingChanges && (
+        <BetaAlert
+          heading="Refresh your page"
+          alertMessage="Hard refresh your page when done changing Flags for the changes to fully apply."
+        />
+      )}
       <Table>
         <thead>
           <tr>
-            <th scope="col">Status</th>
+            <th scope="col">Default</th>
+            <th scope="col">Current</th>
             <th scope="col">Feature Flag</th>
             <th scope="col">Actions</th>
           </tr>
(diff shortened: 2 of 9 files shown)


frontend/src/services/featureFlags/FeatureFlagManager.ts
```diff
@@ -125,25 +126,35 @@ export class FeatureFlagsManager {
...
+
+    const featureFlags = {
+      ...getFeatureFlagsFromCookie(request.cookies),
+      ...featureFlagsFromQuery,
+    };
+
+    const nonDefaultFlags: { [key: string]: boolean } = {};
+    Object.keys(featureFlags).forEach((key) => {
+      const value = featureFlags[key];
+      if (value !== defaultFeatureFlags[key]) {
```
doug-s-nava: I think we need to look at env var based feature flag values here (this.featureFlags), since the the default values will only come into play if there are no env vars, which shouldn't ever happen in a deployed env

doug-s-nava:
updated broken tests, feel like this needs some additional testing on the user provider and the middleware, lemme know if you'd like me to try and tackle that

frontend/src/components/ClientMaintenanceCheck.tsx
```diff
@@ -0,0 +1,28 @@
...
+    checkFeatureFlag, // An instance of FeatureFlagsManager
+  } = useFeatureFlags();
+
+  useEffect(() => {
+    if (
+      !checkFeatureFlag("searchOff") &&
+      !checkFeatureFlag("opportunityOff") &&
+      checkFeatureFlag("authOn")
+    ) {
+      // console.log("would redir");
```
doug-s-nava: delete me

frontend/src/components/ClientMaintenanceCheck.tsx
```diff
@@ -0,0 +1,28 @@
+"use client";
+
+import { useFeatureFlags } from "src/hooks/useFeatureFlags";
+
+import { redirect, RedirectType } from "next/navigation";
+import { useEffect } from "react";
+
+/**
+ * View for managing feature flags
```
doug-s-nava: this comment is a little confusing - I'd say this component is designed to redirect away from the maintenance page if all maintenance related feature flags are off

frontend/src/components/Layout.tsx
```diff
@@ -18,7 +19,7 @@ export default function Layout({ children, locale }: Props) {
   const t = useTranslations();
 
   return (
-    <UserProvider>
+    <UserProvider featureFlagDefaults={defaultFeatureFlags}>
```
doug-s-nava: similar to what we discussed in the middleware, I think this also needs to include env var related flags

frontend/src/components/dev/FeatureFlagsTable.tsx
```diff
@@ -1,29 +1,55 @@
...
             <th scope="col">Actions</th>
           </tr>
         </thead>
         <tbody>
           {Object.entries(featureFlags).map(([featureName, enabled]) => (
             <tr key={featureName}>
+              <td
+                data-testid={`${featureName}-default`}
+                style={{
+                  background: defaultFeatureFlags?.[featureName]
```
doug-s-nava, line 46: I think we need to look at env vars here as well

frontend/src/hooks/useFeatureFlags.ts
```diff
@@ -24,19 +18,12 @@ export function useFeatureFlags(): {
...
-    ) as FeatureFlags;
-    setFeatureFlags(flagsFromCookie);
-  }, []);
+  const userContext = useUser();
+  const pathname = usePathname() || "";
+  const router = useRouter();
 
   // Note that values set in cookies will be persistent per browser session unless explicitly overwritten
-  const setFeatureFlag = useCallback(
+  /*   const setFeatureFlag = useCallback(
```
doug-s-nava: delete me

frontend/src/services/featureFlags/FeatureFlagManager.ts
```diff
@@ -125,25 +126,35 @@ export class FeatureFlagsManager {
...
-        : {
-            ...this.featureFlags,
-            ...getFeatureFlagsFromCookie(request.cookies),
-            ...featureFlagsFromQuery,
-          };
-
-    setCookie(JSON.stringify(featureFlags), response.cookies);
+    // flag admin view, which depends on reading all flags from cookies, so the logic has been removed
+
+    const featureFlags = {
```
doug-s-nava: consider renaming to `userFeatureFlags` to more firmly differentiate from `this.featureFlags`?

frontend/tests/hooks/useFeatureFlags.test.ts
```diff
@@ -35,39 +43,35 @@ describe("useFeatureFlags", () => {
...
       const { checkFeatureFlag } = result.current;
 
-      const value = checkFeatureFlag("someFakeFeature1");
+      const value = checkFeatureFlag("someFlag");
       expect(value).toEqual(true);
     });
 
     test("returns false if specified flag is not present", () => {
-      mockGetCookie.mockReturnValue(mockDefaultFeatureFlagsString);
+      // mockGetCookie.mockReturnValue(mockDefaultFeatureFlagsString);
```
doug-s-nava: delete me

frontend/tests/pages/maintenance/page.test.tsx
```diff
@@ -20,22 +21,146 @@ jest.mock("react", () => ({
...
   useMessages: () => mockMessages,
 }));
 
 describe("Maintenance", () => {
   it("renders intro text", () => {
-    render(<Maintenance params={localeParams} />);
-
+    // set a major feature offline so we don't redirect to homepage
+    render(
+      <UserContext
```
doug-s-nava: this is a lot of mocking to do - worth looking into the pattern used in the `FakeQueryProvider` in case we want to make this more reusable

frontend/tests/services/featureFlags/FeatureFlagManager.test.ts
```diff
@@ -131,23 +128,28 @@ describe("FeatureFlagsManager", () => {
...
-      expect(mockSet).toHaveBeenCalledWith({
-        expires: expect.any(Date) as jest.Expect,
-        name: FEATURE_FLAGS_KEY,
-        value: JSON.stringify({
-          feature1: true, // from default
-          feature2: false, // from default
-          feature3: false, // from env var
-        }),
-      });
+      // expect(mockSet).toHaveBeenCalledWith({
```
doug-s-nava: delete me

frontend/tests/services/featureFlags/FeatureFlagManager.test.ts
```diff
@@ -131,23 +128,28 @@ describe("FeatureFlagsManager", () => {
...
+      jest.spyOn(NextResponse.prototype, "cookies", "get").mockReturnValue({
+        set: mockSet,
+        delete: mockDelete,
+      } as object as NextResponse["cookies"]);
 
       const featureFlagsManager = new FeatureFlagsManager({
         feature3: "false",
       });
       featureFlagsManager.middleware(request, NextResponse.next());
-      expect(mockSet).toHaveBeenCalledWith({
```
doug-s-nava, line 142: would love to see another test here covering the new functionality in the middleware

frontend/src/components/dev/FeatureFlagsTable.tsx
```diff
@@ -1,29 +1,55 @@
...
+                    ? "#81cc81"
+                    : "#fc6a6a",
+                }}
+              >
+                {defaultFeatureFlags?.[featureName] ? "Enable" : "Disable"}
```
doug-s-nava: I think we want "Enabled / Disabled" here

doug-s-nava (commented):
some weird behavior to call out:
* http://localhost:3000/search?_ff=searchOff:true - seems to direct to maintenance, but then redirects to the homepage
* set searchOff to true in the ui, click apply, click search - not redirected to maintenance
* still need to test with env vars

doug-s-nava (commented):
confirmed that while the search drawer does appear when started with `FEATURE_SEARCH_DRAWER_ON=true npm run dev`, the ui still shows the flag as disabled

with `FEATURE_SEARCH_OFF=true npm run dev`, visiting /search redirects back to the home page

frontend/src/components/ClientMaintenanceCheck.tsx
mdragon: Copypasta ❌

mdragon:
> * still need to test with env vars

I think we do have to create an API endpoint and fetch the "env" ones when on the client:
image
image

frontend/tests/pages/maintenance/page.test.tsx
mdragon: Took a slightly different approach that felt like it would allow easier configuration of the feature flags across different tests, but let me know what you think when I push this up next.

PR #5281, opened 2025-06-09
[Issue #5280] Align naming with what's configured for the jobs in Terraform

### LINKED ISSUE #5280: Infra/API config values mismatch for Pinpoint
### Summary

It seems like the Python code is using PINPOINT_APP_ID and the Terraform configured it as AWS_PINPOINT_APP_ID.

Image

### Acceptance criteria

- Both things use the same name for PINPOINT

mdragon:
## Summary

Fixes #5280

## Changes proposed
Align to the Env Var names Terraform is actually assigning.

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -46,7 +46,14 @@ See [development.md](../documentation/api/development.md) for installation and d
 ```bash
 make run-generate-notifications
 # executes poetry run flask task generate-notifications
+
+# more generically, you can construct poetry run flask calls with make cmd
+make cmd args="data-migration setup-foreign-tables"
+make cmd args="data-migration load-transform --no-load --transform --no-set-current"
+make cmd args="task create-analytics-db-csvs"
 ```
+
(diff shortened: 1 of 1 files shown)


api/README.md
```diff
@@ -39,6 +39,30 @@ root
...
 See [development.md](../documentation/api/development.md) for installation and development instructions.
 
+## Running jobs/tasks
+
+### Locally
+
+```bash
+make run-generate-notifications
+# executes poetry run flask task generate-notifications
+```
```
chouinar: A slightly more generic way if you haven't setup a specific make command: `make cmd args=""` where args are whatever you want after `poetry run flask`.

A few from my own history:
* `make cmd args="data-migration setup-foreign-tables"`
* `make cmd args="data-migration load-transform --no-load --transform --no-set-current"`
* `make cmd args="task create-analytics-db-csvs"`

And maybe useful to note that all the Poetry CLI commands are of the form `<task group> <task name> <any other params>`. So in the above, `data-migration` is the task group for the first two, but then the task name is `setup-foreign-tables` and `load-transform`

mdragon: Added.

PR #5295, opened 2025-06-10
[Issue #4935] add cost sharing filter

### LINKED ISSUE #4935: Add Filter for Cost Sharing Yes/No
### Summary

Cost sharing can be an important decision point in whether a person or organization wants to pursue an opportunity, so we should allow users to filter those options in or out

🏁 Place behind a feature flag: newFilterViewOn (Enable/Disable, disabled by default until view is complete and tested)

### Acceptance criteria

- [ ] Add new filter for "Cost Sharing" which has the options "Any cost sharing", "Yes", and "No" represented as radio buttons
- [ ] When selected the search should refine down to the relevant results

### Notes
- Use Radio Buttons

doug-s-nava:
## Summary

Fixes #4935

Review https://github.com/HHS/simpler-grants-gov/pull/5231 first

## Changes proposed

Adds a filter for cost sharing (yes/no) - note that this is behind the searchDrawerOn feature flag

## Context for reviewers

This includes the changes from https://github.com/HHS/simpler-grants-gov/pull/5231 in order to pull in the radio button based components that are added there

## Validation steps

2. visit http://localhost:3000/search?_ff=searchDrawerOn:true
3. open the filter drawer
4. _VERIFY_: there is a cost sharing filter using radio buttons to filter for opportunties with or without cost sharing (yes and no options)
5. use the close date filter
6. _VERIFY_: search results and url update as expected
7. save a search that includes a close date filter
8. visit the saved searches page and click on your saved search
9. _VERIFY_: you're correctly delivered back to the search results page with close date filter enabled


frontend/src/errors.ts
```diff
@@ -174,6 +174,7 @@ type SearchInputsSimple = {
   [key: string]: string[] | string | number | null | undefined;
 };
 
+// Is this still being used? Can we get rid of this?
```
freddieyebra, line 177: FWIW I see this referenced in some error classes.

freddieyebra (approved):
looks good!

Verified 
Screenshot 2025-06-12 at 11 00 01 AM
Screenshot 2025-06-12 at 10 59 53 AM

frontend/src/errors.ts
doug-s-nava: I am pretty sure this was being used by the old search error page to reproduce the form state, and is no longer needed, but that's definitely a topic for another ticket

PR #5300, opened 2025-06-11
[Issue #5248] Create a task that creates organizations from ebiz POCs

### LINKED ISSUE #5248: Create a task that creates orgs from sam.gov entity data + ebiz POCs
### Summary

Organizations are created in our system automatically when we are able to join the sam.gov entity table's ebiz POC email column with the email of a user in the `link_external_user` table.

For this logic, we first want to do a join like:
```
select * from sam_gov_entity join link_external_user on sam_gov_entity.ebiz_poc_email = link_external_user.email
```
NOTE: This is not necessarily all we want in the query, we likely want to join and fetch additional info in the query for optimization purposes (ie. fetch the org, org users, user record).

For every result we then want to do the following:
* See if an organization already exists attached to that sam.gov entity record
   * if not, create that organization.
* Check if the user is an `organization_user` of the organization.
  * If not, create that connection (ie. add that user to that org)
* Check if the user has the `is_ebiz_poc` flag set in that table
  * If not, add that flag

NOTE: Try to compartmentalize this logic a bit to be reusable, a subsequent ticket will need to do something similar for a single user when they login.

This logic effectively says, create an org and/or make that user the ebiz POC. If a different user already exists as the ebiz POC, that's fine, they won't be affected. If the user themselves already exists as a non-ebiz POC, then they just get the flag added. This is a very rudimentary role approach which will change in the future, but is fine for our pilot.

Effectively, there are a few scenarios that this logic should account for (and doing it like above should cover):
* The organization does not exist
* The organization exists, but the user isn't a member of the org
* The organization exists, the user is a member, but not an ebiz POC
* The organization exists, the user is a member, and already the ebiz POC (this should do nothing - this means we already did it all a previous run).

If there is no one with the email of a sam.gov entity, or a users email doesn't correspond to anything in the sam.gov table, nothing should happen to any of them. Most of our sam.gov entities and users will not be affected by this - this is effectively adding admins to organizations, which only specific people are.

### Acceptance criteria

- [ ] Task created for doing this connection (note - we don't need to setup a CLI command, it'll run alongside something else in a future PR)
- [ ] Tests added that test each scenario - make sure to load random users / entities in that won't be a part of the join to verify that this is only getting what we expect.

chouinar:
## Summary

Fixes #5248

## Changes proposed
Created a task that sets up organizations when the EBIZ POC email has a corresponding user

## Context for reviewers
Organizations aren't created by users, instead if the sam.gov entity record we imported happens to have an EBIZ POC email that matches one in our system we want to setup the user as an owner of that organization.

This implementation handles a few scenarios:
* Org doesn't exist at all -> Org is created, user is added as an owner of the organization
* Org exists, user isn't a member -> User is added as an owner of the organization
* Org exists, user is a member, but not an owner -> User is made owner of the organization
* Org exists, user is already an owner -> nothing happens

Any existing users in an organization are unaffected, including other owners. This only adds someone as an owner, it doesn't kick out the old owners (when user management is built, we will let them do that potentially).

## Validation steps
Just the tests which handle various scenarios / data


api/src/db/models/entity_models.py
```diff
@@ -35,8 +35,8 @@ class SamGovEntity(ApiSchemaTable, TimestampMixin):
     import_records: Mapped[list["SamGovEntityImportType"]] = relationship(
         back_populates="sam_gov_entity", cascade="all, delete-orphan"
     )
-    organizations: Mapped[list["Organization"]] = relationship(
-        back_populates="sam_gov_entity",
+    organization: Mapped["Organization | None"] = relationship(
```
mikehgrantsgov, line 40: I didn't see anything in the ticket regarding changing this from a list to a single organization. Is there any other impact making this a single org?

mikehgrantsgov:
I had one question around the organizations -> organization change, but should be good after the conflicts are resolved.

api/src/db/models/entity_models.py
chouinar: I think when we first setup the tables there was an idea that we might have some different organization/sam.gov entity connection (multiple orgs for one UEI?) which conversations over the past month or two have made clear isn't needed / would make things way too complex.

I also double-checked that UEI is a unique value, and since an org generally is thought of as "a UEI" - adjusting this to follow that makes sense.

mikehgrantsgov (dismissed):

PR #5302, opened 2025-06-11
[Unticketed] Add created_at/updated_at to app form and app attachment schemas

chouinar:
## Changes proposed
Adds created_at/updated_at timestamps to the application form and application attachment data API models.

## Context for reviewers
Frontend has a few places they might show these timestamps, just piping them through

### DIFF
--- a/api/openapi.generated.yml
+++ b/api/openapi.generated.yml
@@ -3147,7 +3147,6 @@ components:
           - complete
           type:
           - string
-<<<<<<< HEAD
         created_at:
           type: string
           format: date-time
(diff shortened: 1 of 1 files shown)


api/openapi.generated.yml
```diff
@@ -3692,11 +3700,11 @@ components:
         created_at:
           type: string
           format: date-time
-          readOnly: true
```
mikehgrantsgov: Safe to remove `readOnly` here?

chouinar: The read-only doesn't really matter (if we were reusing schemas for read and write it would, but we haven't really done that yet), BUT this happened because I'm effectively overwriting created_at/updated_at that are already on the schema it's derived from.

mikehgrantsgov (dismissed):

mikehgrantsgov (approved):

PR #5315, opened 2025-06-11
[Issue #5280] Add logging of Pinpoint response to aid troubleshooting 

### LINKED ISSUE #5280: Infra/API config values mismatch for Pinpoint

mdragon:
## Summary

We're getting log messages that we "successfully" sent the message to Pinpoint, but Pinpoint doesn't send any emails. Adding logging of Pinpoint response values to help understand what's happening.

Fixes #5280 

## Changes proposed

Add one additional Pinpoint response field to the model and log all 4 response values to what we include in the success log.

### DIFF
--- a/api/src/task/notifications/base_notification.py
+++ b/api/src/task/notifications/base_notification.py
@@ -65,24 +65,27 @@ class BaseNotificationTask(Task):
                     message=user_notification.content,
                     app_id=self.notification_config.app_id,
                 )
+
+                email_response = response.results.get(user_notification.user_email, None)
+
                 logger.info(
                     "Successfully delivered notification to user",
                     extra={
                         "user_id": user_notification.user_id,
                         "notification_reason": user_notification.notification_reason,
                         "notification_log_id": notification_log.user_notification_log_id,
-                        "pinpoint_delivery_status": response.results[
-                            user_notification.user_email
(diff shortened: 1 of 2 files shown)


api/src/task/notifications/base_notification.py
```diff
@@ -71,6 +71,18 @@ def send_notifications(self, notifications: list[UserEmailNotification]) -> None
...
+                        ].delivery_status,
+                        "pinpoint_message_id": response.results[
+                            user_notification.user_email
+                        ].message_id,
+                        "pinpoint_status_code": response.results[
+                            user_notification.user_email
+                        ].status_code,
+                        "pinpoint_status_message": response.results[
+                            user_notification.user_email
+                        ].status_message,
```
chouinar: In the event that the results don't have the users email (which if it's not working is possible), should we be safer and do something like:

```py
pinpoint_response = response.results.get(email, None)
...
# in the log extra here
    "pinpoint_delivery_status": pinpoint_response.delivery_status if pinpoint_response else None
```

chouinar (approved):

PR #5327, opened 2025-06-12
Allow a user to connect an organization to their application in the create application endpoint 

### LINKED ISSUE #5251: Allow a user to connect an organization to their application in the create application endpoint
### Summary

After #5246 

We should take in a non-required (but also non-null) `organization_id` param to our create application endpoint.

If this parameter is supplied, we want to do the following:
* Fetch the organization (404 if not found)
* Check if the user is a member of the org (403 forbidden if not - note in the future we'll have actual roles we check against, but for now just being in the org is enough).
* Add the organization ID to the application record

If the user does not supply the parameter, then we don't do any of these checks and the application won't have this field set.

### Acceptance criteria

- [ ] Add parameter to create application endpoint
- [ ] Add logic for fetching org / checking membership
- [ ] Add tests

mikehgrantsgov:
## Summary

Fixes #5251

## Changes proposed

Add a `_validate_organization_membership` to validate org membership
Add new tests to cover scenarios
Return `organization_id` in `ApplicationStartRequestSchema`

## Context for reviewers

The org is optional but is validated and returned if provided.

## Validation steps

See added unit tests.

### DIFF
--- a/api/src/services/applications/create_application.py
+++ b/api/src/services/applications/create_application.py
@@ -9,7 +9,8 @@ import src.adapters.db as db
 from src.api.route_utils import raise_flask_error
 from src.constants.lookup_constants import ApplicationStatus
 from src.db.models.competition_models import Application, ApplicationForm, Competition
-from src.db.models.user_models import ApplicationUser, User
+from src.db.models.entity_models import Organization
+from src.db.models.user_models import ApplicationUser, OrganizationUser, User
 from src.services.applications.application_validation import (
     ApplicationAction,
     validate_competition_open,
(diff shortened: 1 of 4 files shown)


api/src/services/applications/create_application.py
```diff
@@ -18,8 +19,36 @@
...
+
+    if not organization:
+        raise_flask_error(404, "Organization not found")
+
+    # Check if the user is a member of the organization
+    is_member = db_session.execute(
+        select(OrganizationUser)
+        .where(OrganizationUser.organization_id == organization_id)
+        .where(OrganizationUser.user_id == user.user_id)
+    ).scalar_one_or_none()
```
chouinar: I think if we add a selectinload for the organization user, then the second query here wouldn't even have to hit the DB as we'd already have the values cached. Very minor optimizations, but quick to switch.

```suggestion
    organization = db_session.execute(
        select(Organization).where(Organization.organization_id == organization_id).options(selectinload(Organization.organization_users))
    ).scalar_one_or_none()

    if not organization:
        raise_flask_error(404, "Organization not found")

    # Check if the user is a member of the organization
    is_member = db_session.execute(
        select(OrganizationUser)
        .where(OrganizationUser.organization_id == organization_id)
        .where(OrganizationUser.user_id == user.user_id)
    ).scalar_one_or_none()
```

api/tests/src/api/applications/test_application_routes.py
```diff
@@ -1762,3 +1764,301 @@ def test_application_get_without_organization(
...
+        .all()
+    )
+    assert len(applications_count) == 0
+
+
+def test_application_start_organization_membership_validation_works_with_multiple_users(
+    client, enable_factory_create, db_session, user, user_auth_token
+):
+    """Test that organization membership validation works correctly with multiple users"""
+    from tests.src.db.models.factories import UserFactory
```
chouinar: Should this be at the top with the other imports?

mikehgrantsgov: Ah yeah, I had some transient import issues but think it was an temporary environment issue. Moved this up to the other imports.

chouinar (approved):

PR #5330, opened 2025-06-12
[Issue #5280] Ensure TraceId is always set since it can't be None and it's helpful to have it

mdragon:
## Changes proposed

Default the TraceId, but also fix the bug where we weren't passing in the one we were logging.

This is now ready for review. I validated that the SES IAM error hiding inside a Pinpoint Success will now throw. 

One thing we'll want to do in a follow up ticket is not update the list of search IDs if the send fails. Currently if the job fails to send and then you run the job again it doesn't find anything to send because the row is updated and there's no "changes."

### DIFF
--- a/api/src/adapters/aws/pinpoint_adapter.py
+++ b/api/src/adapters/aws/pinpoint_adapter.py
@@ -149,6 +149,9 @@ def send_pinpoint_email_raw(
                     "pinpoint_trace_id": trace_id,
                 },
             )
+            raise Exception(
+                f"Failed to send email pinpoint_trace_id: {trace_id} with pinpoint_delivery_status: {email_response.delivery_status if email_response else None}"
+            )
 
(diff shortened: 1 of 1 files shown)


api/src/adapters/aws/pinpoint_adapter.py
```diff
@@ -62,7 +62,7 @@ def send_pinpoint_email_raw(
     message: str,
     app_id: str,
     pinpoint_client: botocore.client.BaseClient | None = None,
-    trace_id: str | None = None,
+    trace_id: str = str(uuid.uuid4()),
```
chouinar: I think this'll break / statically define a value when the module is loaded.

Instead do something like:
```py
if trace_id is None:
    trace_id = str(uuid.uuid4())
```

chouinar (approved):

PR #5331, opened 2025-06-12
[Issue $5219] Update the application attachment data model to have a user_id foreign key for who uploaded it

### LINKED ISSUE #5219: Update the application attachment data model to have a user_id foreign key for who uploaded it
### Summary

Added a foreign key form the application attachment table to the user table called `user_id` - don't allow it to be null (might break some tests, but no actual attachments exist yet).

In the application attachment response model we want to add a "user" model that has the user_id + email address so that the frontend can display something.

When an application attachment is created/updated, set the user_id from the user that made the request.

### Acceptance criteria

- [ ] Table updated
- [ ] Response schema updated
- [ ] Application attachment creation updated

mikehgrantsgov:
## Summary

Fixes #5219 

## Changes proposed

Add user_id as a FK to `application_attachment` table
Update tests and factories

## Context for reviewers

In the application attachment response model we want to add a "user" model that has the user_id + email address so that the frontend can display something.

Couldn't find a location where attachments are 'updated', per se.

## Validation steps

See migration / updated tests


api/src/services/applications/create_application_attachment.py
```diff
@@ -48,6 +48,9 @@ def create_application_attachment(
     # so instead get the content-length from the file on s3 after we upload it.
     file_size_bytes = file_util.get_file_length_bytes(s3_file_location)
 
+    # If we don't do this, we see `Can't attach instance of 'User' to session` errors.
+    user = db_session.merge(user)
```
chouinar, line 52: Right - haven't hit this yet, but it's expected - we technically fetch the user in an entirely different DB session during the auth logic, I believe merge is right since it moves the state between the sessions.

PR #5333, opened 2025-06-12
NOFOs: Build a "prod" image after tests pass

pcraig3:
## Summary

The NOFO app container we use to run tests is not the one we should deploy.

We have an [`IS_PROD`](https://github.com/HHS/simpler-grants-pdf-builder/blob/main/Dockerfile#L6) variable that sets up our prod configuration, but if we turn it on from the get go, some of our tests fail.

## Changes proposed

Previously, we would do this:

- Build container
- Run tests
- Cache image
- Another workflow deploys cached image

Now, we do this:

- Build container
- Run tests
- Build PROD container
- Cache prod image
- Another workflow deploys cached image

## Validation steps

It works!

https://github.com/HHS/simpler-grants-gov/actions/runs/15619933756

### DIFF
--- a/.github/workflows/ci-nofos.yml
+++ b/.github/workflows/ci-nofos.yml
@@ -27,7 +27,7 @@ jobs:
           repository: HHS/simpler-grants-pdf-builder
           ref: ${{ inputs.version || 'main' }}
 
-      - name: Build docker image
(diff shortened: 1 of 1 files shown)


mdragon (approved):
👍🏻

PR #5342, opened 2025-06-13
[Issue #5252] Create an endpoint for a user to be able to fetch their organizations

### LINKED ISSUE #5252: Create an endpoint for a user to be able to fetch their organizations
### Summary

We want to create a `GET /v1/users/:user_id/organizations` endpoint that will return a list of organizations that the user is associated with.

### Request
Nothing beyond the user ID in the path

### Response
A list of organizations / with sam.gov entities in them.
```json
{
  "data": [
       {
         "organization_id": "<some uuid>",
         "sam_gov_entity": {
            "uei": "str",
            "legal_business_name": "",
            "expiration_date": "",
         }
       }
   ]
}
```

### Logic
Our usual auth logic from our other user endpoints (a user can only access their own records)

To get the above information, we can determine the orgs a user is in by connecting through the `organization_user` table to the organization table. We then just need to return a list of the organization records which should have the same shape as the above response already.

If no organizations are found, just return an empty list, no 404.

### Acceptance criteria

- [ ] Endpoint created
- [ ] Tests added

mikehgrantsgov:
## Summary

Fixes #5252

## Changes proposed

Create `"/<uuid:user_id>/organizations` endpoint to return a user's organizations
Add service to get user's organizations
Add tests

## Context for reviewers

We can determine the orgs a user is in by connecting through the organization_user table to the organization table. We then just need to return a list of the organization records which should have the same shape as the above response already.

## Validation steps

See unit tests

### DIFF
--- a/api/src/services/users/get_user_organizations.py
+++ b/api/src/services/users/get_user_organizations.py
@@ -14,7 +14,9 @@ def _fetch_user_organizations(db_session: db.Session, user_id: UUID) -> list[Org
     stmt = (
         select(OrganizationUser)
         .where(OrganizationUser.user_id == user_id)
-        .options(selectinload(OrganizationUser.organization).selectinload(Organization.sam_gov_entity))
+        .options(
+            selectinload(OrganizationUser.organization).selectinload(Organization.sam_gov_entity)
+        )
     )
 
     organization_users = db_session.execute(stmt).scalars().all()
--- a/api/src/services/users/get_user_organizations.py
+++ b/api/src/services/users/get_user_organizations.py
@@ -39,7 +39,7 @@ def get_user_organizations(db_session: db.Session, user_id: UUID) -> list[dict[s
             org_data["sam_gov_entity"] = {
                 "uei": sam_gov_entity.uei,
                 "legal_business_name": sam_gov_entity.legal_business_name,
-                "expiration_date": sam_gov_entity.expiration_date.isoformat(),
+                "expiration_date": sam_gov_entity.expiration_date,
             }
 
         organizations.append(org_data)
--- a/api/src/services/users/get_user_organizations.py
+++ b/api/src/services/users/get_user_organizations.py
@@ -2,7 +2,7 @@ from typing import Any
 from uuid import UUID
 
 from sqlalchemy import select
-from sqlalchemy.orm import joinedload
+from sqlalchemy.orm import selectinload
 
 from src.adapters import db
 from src.db.models.entity_models import Organization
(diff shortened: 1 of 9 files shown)


api/src/services/users/get_user_organizations.py
```diff
@@ -0,0 +1,47 @@
...
+from src.db.models.user_models import OrganizationUser
+
+
+def _fetch_user_organizations(db_session: db.Session, user_id: UUID) -> list[OrganizationUser]:
+    """Fetch all organizations for a user with their SAM.gov entity data."""
+    stmt = (
+        select(OrganizationUser)
+        .where(OrganizationUser.user_id == user_id)
+        .options(joinedload(OrganizationUser.organization).joinedload(Organization.sam_gov_entity))
+    )
```
chouinar: Instead of `joinedload`, could we use `selectinload`? Joinedload tends to lock up when there is more data in a given table (no idea why - but the docs generally recommend `selectinload` anyways).

api/src/services/users/get_user_organizations.py
```diff
@@ -0,0 +1,47 @@
...
+        }
+
+        if sam_gov_entity:
+            org_data["sam_gov_entity"] = {
+                "uei": sam_gov_entity.uei,
+                "legal_business_name": sam_gov_entity.legal_business_name,
+                "expiration_date": sam_gov_entity.expiration_date,
+            }
+
+        organizations.append(org_data)
```
chouinar: I forgot to put one parameter we want in the response, can we pipe through the `is_organization_owner` param in the object as well? It's the precursor to eventual roles which we'd want to return.

So the response would look like:
```json
{
  "data": [
       {
         "organization_id": "<some uuid>",
         "is_organization_owner": true,
         "sam_gov_entity": {
            "uei": "str",
            "legal_business_name": "",
            "expiration_date": "",
         }
       }
   ]
}
```
---
As for actually building a response object, we don't need to pull the values out if they're already in the shape we want (even if that shape is our DB models), can do something like:
```py
org_data = {
  "organization_id": organization.organization_id,
  "is_organization_owner": organization_user.is_organization_owner,
  "sam_gov_entity": organization.sam_gov_entity # even if null, should be fine to do this
}
```
If it weren't for needing that organization user bit, could just return a list of `Organization` DB model objects and be perfectly fine as that's fully in the right shape.

api/src/services/users/get_user_organizations.py
mikehgrantsgov: Makes sense, updated this.

api/src/services/users/get_user_organizations.py
mikehgrantsgov: I made this change to add `is_organization_owner` and updated the sam_gov_entity in the response.

chouinar (approved):

PR #5356, opened 2025-06-16
[Issue #5085] Create PUT `/applications/:application_id/attachments/:attachment_id` endpoint

### LINKED ISSUE #5085: Create PUT /applications/:application_id/attachments/:attachment_id
### Summary

We want to create `PUT /applications/:application_id/attachments/:attachment_id` which will update an attachment. It should effectively be the same as the other one, but will handle "updating" the file for the given attachment ID.

### Request
Same as the other PUT application endpoint

### Response
Same as the other PUT application endpoint

### Logic
Most of this logic should be reusable from the GET + PUT endpoints.

* Fetch the application, 404 if not found
* 404 if the attachment ID is also not found
* User authentication, only an owner of the application can modify it (same as other endpoints, reuse their logic)
* Adjust the filename using werkzeug's secure_filename function
* Upload the file to s3
   * The file should be uploaded to our "draft" bucket which is private
   * The filepath should be something like `s3://<draft_bucket>/applications/<application_id>/attachments/<attachment_id>/<file_name>`
   * IF the filepath on s3 is different than before, delete the old file from s3, if it's the same, fine to just overwrite it.
* Update the application attachment record in the DB.

### Acceptance criteria

- [ ] Created endpoint
- [ ] Tests added

mikehgrantsgov:
## Summary

Fixes #5085

## Changes proposed

Add route and associated service.
Add `ApplicationAttachmentUpdateRequestSchema` and `ApplicationAttachmentUpdateResponseSchema` to handle request/response.
Add tests covering multiple scenarios.

## Context for reviewers

Much of this has been reused from other existing routes / services.

Route updates an attachment. It should effectively be the same as the other one, but will handle "updating" the file for the given attachment ID. Deletes old attachment path if new path is provided. Uses secure_filename to regenerate a file name, if needed.

## Validation steps

See added unit tests.

### DIFF
--- a/api/src/services/applications/create_application_attachment.py
+++ b/api/src/services/applications/create_application_attachment.py
@@ -88,9 +88,7 @@ def upsert_application_attachment(
     application_attachment.file_size_bytes = file_size_bytes
     application_attachment.user = user
 
-    # Only add to session if it's a new attachment (checking if it already exists in the session)
-    if application_attachment not in db_session:
-        db_session.add(application_attachment)
+    db_session.add(application_attachment)
 
     logger.info(
         "Created/updated application attachment",
--- a/api/src/services/applications/create_application_attachment.py
+++ b/api/src/services/applications/create_application_attachment.py
@@ -7,7 +7,7 @@ from werkzeug.datastructures import FileStorage
 import src.adapters.db as db
 import src.util.file_util as file_util
 from src.adapters.aws import S3Config
-from src.db.models.competition_models import ApplicationAttachment
+from src.db.models.competition_models import Application, ApplicationAttachment
 from src.db.models.user_models import User
 from src.services.applications.get_application import get_application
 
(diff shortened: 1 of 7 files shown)


api/src/services/applications/update_application_attachment.py
```diff
@@ -0,0 +1,85 @@
...
+    application_id: uuid.UUID,
+    application_attachment_id: uuid.UUID,
+    user: User,
+    form_and_files_data: dict,
+) -> ApplicationAttachment:
+    # Fetch the application attachment, handles verifying
+    # it exists, and that the user has access attachment.
+    application_attachment = get_application_attachment(
+        db_session, application_id, application_attachment_id, user
+    )
```
chouinar, line 25: Like some of the other application endpoints, could we try to reuse as much of create_application_attachment function as we can? That way if we adjust something in the behavior, we adjust one spot and it affects both endpoints.

Something like - make a new function like `upsert_application_attachment` that takes in an optional `ApplicationAttachment` and creates/updates it.

Then this update method is roughly:
```py
    application_attachment = get_application_attachment(
        db_session, application_id, application_attachment_id, user
    )

    old_s3_file_location = application_attachment.file_location
    updated_application_attachment = upsert_application_attachment(application_attachment, ...)

    if old_s3_file_location != updated_application_attachment.file_location:
         # delete stuff
    
    return updated_application_attachment
```

The upsert method would just be the current create method but we'd have to slightly adjust the creation of the attachment object:
```py
    if application_attachment is None:
         application_attachment = ApplicationAttachment()

    application_attachment.file_name = ...
```

mikehgrantsgov: Makes sense, I refactored this to reuse and follow that update logic above.

api/src/services/applications/create_application_attachment.py
```diff
@@ -51,22 +78,22 @@ def create_application_attachment(
...
+    application_attachment.file_location = s3_file_location
+    # In the file_name column we store the users actual file name unmodified
+    # so when we display it on the UI it matches whatever they uploaded.
+    application_attachment.file_name = file_util.get_file_name(file_attachment.filename)
+    application_attachment.mime_type = file_attachment.mimetype
+    application_attachment.file_size_bytes = file_size_bytes
+    application_attachment.user = user
+
+    # Only add to session if it's a new attachment (checking if it already exists in the session)
+    if application_attachment not in db_session:
```
chouinar: I think you're safe to just always do `db_session.add(..)` on something already in the session since it really just changes internal state to the same one as it would be when we fetch it from the DB.

chouinar (dismissed):

Issue #5337, opened 2025-06-13
BE - Implement LIKE results in agency search component


eric-serrato95:
## Summary 
A user should now be able to introduce a vague term and get LIKE results E.G. User inputs HAZ - Agency retrieves results for all possible matches that include HAZ somewhere in the agency name like HAZmat or super HAZardous 

### Acceptance criteria

- [ ] While using the agency search box the user should be able to get all potential matches to the term being searched for 
- [ ] Results should be scrollable

eric-serrato95:
@mdragon  this is the implementation for the fuzzy search terms

mdragon:
This should be working in Dev/Staging now: https://staging.simpler.grants.gov/search

PR #5357, opened 2025-06-16
[Issue #5337] Make Agency Search perform a more LIKE type match

mdragon:
## Changes proposed

Use the OpenSearch prefix * indicator to get OpenSearch to treat the search term like a prefix not a full word. Basic support for not doing that if the user quotes any of their term.

## Context for reviewers

This is the smallest delta to accomplish the AC and allow us to evaluate how the behavior is understood by users. More helpful implementation might use fuzziness or incorporate some sort of logic to switch to fuzzy when a certain word or character count is reached, apply the prefix/fuzzy indicator to every word in the search not just the last one, etc.


api/Makefile
```diff
@@ -103,7 +103,7 @@ start-debug:
 run-logs: start ## Start the API and follow the logs
 	docker compose logs --follow --no-color $(APP_NAME)
 
-init: setup-env-override-file build init-db init-opensearch init-localstack init-mock-soap-services
+init: setup-env-override-file build init-db init-opensearch init-localstack init-mock-soap-services ## Run all of the init tasks to set up DB, Search, Local Stack, and Mock services from scratch
```
chouinar: That message might be misleading to new developers, running init doesn't set things up front scratch as the DB + search index both get cached.

To recreate things from scratch, you'll need to run `make clean-volumes` or `make volume-recreate` which just adds init to the end.

api/src/services/agencies_v1/search_agencies.py
```diff
@@ -65,8 +65,13 @@ def get_search_request(params: AgencySearchParams) -> dict:
 
     # Query
     if params.query:
+        wrapped_query = params.query
         filter_rule = DEFAULT
-        builder.simple_query(params.query, filter_rule, SearchQueryOperator.OR)
+        # Don't add the prefix indicator if they searched for quoted text anywhere in the search query
+        if wrapped_query.find('"') == -1:
```
chouinar: Much easier/pythonic/readable to do:

```py
if '"' not in wrapped_query:
```

api/src/services/agencies_v1/search_agencies.py
```diff
@@ -65,8 +65,13 @@ def get_search_request(params: AgencySearchParams) -> dict:
...
+            # For now, use the prefix indicator. ~ for fuzzy could be better in the future, but is too confusing to the user for now
+            wrapped_query = "" + wrapped_query + "*"
```
chouinar:
```suggestion
            wrapped_query = f"{wrapped_query}*"
```

chouinar (approved):

Issue #5166, opened 2025-05-27
[GitHub updates] Propagate deliverable field


widal001:
### Summary

Propagate the "Deliverable" field in the Simpler.Grants.Gov product backlog GitHub project from the parent issue to all sub-issues.

**Note:** Should be a similar pattern to #5144 with a few key differences:
- GitHub project field values are stored on the project item rather than the issue itself, which makes plucking and setting this value a bit more complicated than the milestone.
- Changes to GitHub project fields don't trigger GitHub actions in the same way that changing a milestone value does

### Acceptance criteria

- [ ] When a new sub-issue is created in the same project, it inherits the value of the "Deliverable" field from its parent issue
- [ ] On a regular basis (nightly, weekly?) all of the sub-issues are linted to make sure their "Deliverable" field matches the value of their parent issue.

margaretspring:
@widal001 - Thank you for getting this together. Looks good, I'd say the sub-issue (tasks) assignment of the Deliverable field from the parent should run at least nightly. Without it, tasks getting created in some view won't show up unless manually created. Actually... given that would there be a way to also include logic to copy the Deliverable assignment of a parent when an issue is created, if created with a parent? Thanks!

PR #5359, opened 2025-06-16
[Issue #5166] Fix failure without parent

widal001:
## Summary

Fixes the script that makes an issue inherit its parent deliverable, so that it doesn't fail if an issue doesn't have a parent.

Relates to #5166 

## Changes proposed

- Adds a check to `inherit-parent-deliverable.sh` to log if an issue doesn't have a parent (avoids GitHub action failure)
- Adds a check to `inherit-parent-milestone.sh` to log if an issue doesn't have a parent (improves logging clarity)

## Context for reviewers

I've tested this locally with an issue that previously failed (see screenshots below), but it's hard to test full end-to-end until we merge it in.

## Validation steps

Previous failures and unclear log output for an issue without a parent:

Screenshot 2025-06-16 at 11 09 46 AM
Screenshot 2025-06-16 at 11 10 05 AM

New logs with updated checks for the same issue:

Screenshot 2025-06-16 at 11 10 43 AM
Screenshot 2025-06-16 at 11 11 01 AM

Double checking that it still works when we have a parent issue that was updated:

Screenshot 2025-06-16 at 11 13 05 AM

After unsetting the deliverable field and re-running it:

Screenshot 2025-06-16 at 11 13 52 AM

### DIFF
--- a/.github/linters/scripts/inherit-parent-milestone.sh
+++ b/.github/linters/scripts/inherit-parent-milestone.sh
@@ -110,6 +110,12 @@ parent_milestone_title=$(pluck_field "$parent_issue" '.milestone.title // empty'
 # Check if the parent issue is in the same repo
 # #######################################################
 
+# Check if parent exists
+if [[ $(pluck_field "$data" ".parent") == "null" ]]; then
+  log "No parent issue found. Exiting."
+  exit 0
+fi
+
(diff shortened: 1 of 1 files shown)


mdragon (approved):
🚀

PR #5361, opened 2025-06-16
[Issue #5003] Content change to saved opportunities

### LINKED ISSUE #5003: Implementation: Email notification content for changes to saved opportunities
### Summary

### Acceptance criteria

babebe:
## Summary

Fixes #5003 

## Changes proposed

Update content per [doc](https://docs.google.com/document/d/1ypejg8hL7uOXdrI7auwf9SWQ1Ag07o7rCjnrVCLfA-s/edit?tab=t.0)

### DIFF
--- a/api/src/search/backend/load_opportunities_to_index.py
+++ b/api/src/search/backend/load_opportunities_to_index.py
@@ -218,7 +218,7 @@ class LoadOpportunitiesToIndex(Task):
             # refresh index
             self.search_client.refresh_index(self.index_name)
 
-            # Update updated_at timestamp instead of deleting records
+            # Update updated_at timestamp instead of deleting records             ## this is not updating the updated_at column
             self.db_session.execute(
                 update(OpportunityChangeAudit)
                 .where(OpportunityChangeAudit.opportunity_id.in_(processed_opportunity_ids))
--- a/api/src/services/opportunities_v1/opportunity_version.py
+++ b/api/src/services/opportunities_v1/opportunity_version.py
@@ -4,13 +4,13 @@ from sqlalchemy import select
 from sqlalchemy.orm import selectinload
 
 from src.adapters import db
-from src.api.opportunities_v1.opportunity_schemas import OpportunityV1Schema
+from src.api.opportunities_v1.opportunity_schemas import OpportunityV1Schema, OpportunityWithAttachmentsV1Schema
 from src.db.models.opportunity_models import Opportunity, OpportunityVersion
 from src.util.dict_util import diff_nested_dicts
 
 logger = logging.getLogger(__name__)
 
-SCHEMA = OpportunityV1Schema()
+SCHEMA = OpportunityWithAttachmentsV1Schema()
 
 
 def save_opportunity_version(db_session: db.Session, opportunity: Opportunity) -> bool:
--- a/api/src/task/notifications/constants.py
+++ b/api/src/task/notifications/constants.py
@@ -16,6 +16,7 @@ class Metrics(StrEnum):
     SEARCHES_TRACKED = "searches_tracked"
     NOTIFICATIONS_SENT = "notifications_sent"
     FAILED_TO_SEND = "failed_to_send"
+    NONE_VERSIONED_OPPORTUNITIES = "none_versioned_opportunities"
 
 
 @dataclass
--- a/api/src/task/notifications/opportunity_notifcation.py
+++ b/api/src/task/notifications/opportunity_notifcation.py
@@ -1,69 +1,175 @@
+import json
 import logging
+from datetime import datetime
+from typing import Tuple
 from uuid import UUID
 
-from sqlalchemy import select, update
-from sqlalchemy.orm import selectinload
+from numpy.random.mtrand import Sequence
+from sqlalchemy import select, update, func, desc, tuple_
+from sqlalchemy.orm import selectinload, aliased
+
 
 from src.adapters import db
-from src.db.models.opportunity_models import OpportunityChangeAudit
+from src.api.opportunities_v1.opportunity_schemas import OpportunityV1Schema
+from src.constants.lookup_constants import OpportunityStatus
+from src.db.models.opportunity_models import OpportunityChangeAudit, OpportunityVersion
 from src.db.models.user_models import UserSavedOpportunity
 from src.task.notifications.base_notification import BaseNotificationTask
 from src.task.notifications.constants import Metrics, NotificationReason, UserEmailNotification
 from src.util import datetime_util
+from src.util.dict_util import diff_nested_dicts
 
 logger = logging.getLogger(__name__)
 
+SCHEMA = OpportunityV1Schema()
+
+IMPORTANT_DATE_FIELDS = {
+    "close_date": "The application due date changed from",
+    "forecasted_award_date": "The estimated award date changed from",
+    "forecasted_project_start_date": "The estimated project start date changed from",
+    "fiscal_year": "The fiscal year changed from",
+}
+
+AWARD_FIELDS = {
+    "estimated_total_program_funding": "Program funding changed from",
+    "expected_number_of_awards": "The number of expected awards changed from",
+    "award_floor": "The award minimum changed from",
+    "award_ceiling": "The award maximum changed from",
+}
+
+CATEGORIZATION_FIELDS = {
+    "is_cost_sharing": "Cost sharing or matching requirement has changed from",
+    "funding_instruments": "The funding instrument type has changed from",
+    "category": "The opportunity category has changed from",
+    "category_explanation": "Opportunity category explanation has changed from",
+    "funding_categories": "The category of funding activity has changed from",
+    "funding_category_description": "The funding category explanation has been changed from"
+}
+
+ELIGIBILITY_FIELDS = {
+    "applicant_types": "eligibility criteria include:",
+    "applicant_eligibility_description": "Additional information was",
+}
+
+GRANTOR_CONTACT_INFORMATION_FIELDS = {
+    "agency_name": "The updated grantor’s name is",  # we dont have this
+    "agency_email_address": "The updated email address is",
+    "agency_contact_description": "New description:",
+}
+
+DOCUMENTS_FIELDS = {
+    "attachments": "One or more new documents were ",
+    "additional_info_url": "A link to additional information was",
+}
+
+CONTACT_INFO = (
+    "<div>"
+    "If you have questions, please contact the Grants.gov Support Center:<br><br>"
+    "<a href='mailto:support@grants.gov'>support@grants.gov</a><br>"
+    "1-800-518-4726<br>"
+    "24 hours a day, 7 days a week<br>"
+    "Closed on federal holidays"
+    "</div>"
+)
+
+FIELD_VALUE_MAP = {
+    OpportunityStatus.POSTED: "Open",
+    True: "Yes",
+    False: "No"
+}
+
+SECTION_STYLING = '<p style="margin-left: 20px;">{}</p>'
+BULLET_POINTS_STYLING = f'<p style="margin-left: 40px;">• '
 
 class OpportunityNotificationTask(BaseNotificationTask):
-    def __init__(self, db_session: db.Session):
-        super().__init__(
-            db_session,
-        )
+    def __init__(self, db_session: db.Session, frontend_base_url: str | None = None):
+        super().__init__(db_session)
+        self.frontend_base_url = frontend_base_url
+        self.opportunity_content: dict[UUID, str] = {}
 
     def collect_email_notifications(self) -> list[UserEmailNotification]:
         """Collect notifications for changed opportunities that users are tracking"""
-        stmt = (
-            select(UserSavedOpportunity)
-            .options(selectinload(UserSavedOpportunity.user))
-            .join(
-                OpportunityChangeAudit,
-                OpportunityChangeAudit.opportunity_id == UserSavedOpportunity.opportunity_id,
-            )
-            .where(UserSavedOpportunity.is_deleted.isnot(True))
-            .where(OpportunityChangeAudit.updated_at > UserSavedOpportunity.last_notified_at)
-            .distinct()
+        latest_versions_subq = (
+            select(OpportunityVersion)
+            .distinct(OpportunityVersion.opportunity_id)
+            .order_by(OpportunityVersion.opportunity_id, desc(OpportunityVersion.created_at))
+            .subquery()
         )
 
-        results = self.db_session.execute(stmt).scalars().all()
-        changed_saved_opportunities: dict[UUID, list[UserSavedOpportunity]] = {}
+        latest_opp_version = aliased(OpportunityVersion, latest_versions_subq)
 
-        for result in results:
-            user_id = result.user_id
-            changed_saved_opportunities.setdefault(user_id, []).append(result)
+        stmt = (
+            select(UserSavedOpportunity,latest_opp_version)
+            .options(selectinload(UserSavedOpportunity.user))
+            .outerjoin(
+                latest_opp_version,
+                UserSavedOpportunity.opportunity_id == latest_opp_version.opportunity_id
+            )
+        )
+
+        results = self.db_session.execute(stmt).all()
+        changed_saved_opportunities: dict[UUID, dict] = {}
+        user_opportunity_pairs : list = []
+        none_versioned_opportunities: list = []
+
+        for user_saved_opp, latest_opp_ver in results:
+            if latest_opp_ver is None:
+                logger.error(
+                    "No prior version recorded for this opportunity;",
+                    extra={"opportunity_id": user_saved_opp.opportunity_id},
+                )
+                none_versioned_opportunities.append(user_saved_opp.opportunity_id)
+                continue
+
+            if user_saved_opp.last_notified_at < latest_opp_ver.created_at:
+                user_id = user_saved_opp.user_id
+                opp_id = user_saved_opp.opportunity_id
+                if user_id not in changed_saved_opportunities:
+                    changed_saved_opportunities[user_id] = {
+                        "email": user_saved_opp.user.email,
+                        opp_id: {}
+                    }
+
+                changed_saved_opportunities[user_id][opp_id] = {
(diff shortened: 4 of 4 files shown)


api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -218,7 +218,7 @@ def _handle_incremental_upserts(self, existing_opportunity_ids: set[int]) -> Non
             # refresh index
             self.search_client.refresh_index(self.index_name)
 
-            # Update updated_at timestamp instead of deleting records
+            # Update updated_at timestamp instead of deleting records             ## this is not updating the updated_at column
```
chouinar: Update queries won't ever update it since it's set when we update an individual record as it's done by the ORM, not by any sort of DB trigger.

If you want that updated, you need to fetch the records, and change them individually.

api/src/services/opportunities_v1/opportunity_version.py
```diff
@@ -4,13 +4,13 @@
...
 from src.adapters import db
-from src.api.opportunities_v1.opportunity_schemas import OpportunityV1Schema
+from src.api.opportunities_v1.opportunity_schemas import OpportunityV1Schema, OpportunityWithAttachmentsV1Schema
 from src.db.models.opportunity_models import Opportunity, OpportunityVersion
 from src.util.dict_util import diff_nested_dicts
 
 logger = logging.getLogger(__name__)
 
-SCHEMA = OpportunityV1Schema()
+SCHEMA = OpportunityWithAttachmentsV1Schema()
```
chouinar: I'll caution using this schema, otherwise we're storing the path on s3 to the version table and we likely don't want that.

It might be that we need a separate schema for versioning specifically.

api/src/task/notifications/constants.py
```diff
@@ -16,6 +16,7 @@ class Metrics(StrEnum):
     SEARCHES_TRACKED = "searches_tracked"
     NOTIFICATIONS_SENT = "notifications_sent"
     FAILED_TO_SEND = "failed_to_send"
+    NONE_VERSIONED_OPPORTUNITIES = "none_versioned_opportunities"
```
chouinar: Maybe name this something like `versionless_opportunity_count`?

api/src/task/notifications/opportunity_notifcation.py
```diff
@@ -1,86 +1,443 @@
...
-        changed_saved_opportunities: dict[UUID, list[UserSavedOpportunity]] = {}
+        results = self.db_session.execute(stmt).all()
+        if not results:
+            logger.info("No user-opportunity tracking records found.")
+            return []
 
-        for result in results:
-            user_id = result.user_id
-            changed_saved_opportunities.setdefault(user_id, []).append(result)
+        changed_saved_opportunities: dict[UUID, dict] = {}
```
chouinar: Instead of dictionaries inside of dictionaries with very arbitrary typings you later need to unnest, would making a simple container class help here instead? Trying to parse and debug this if we ever need to change or fix something would require a lot of manually digging through how the data was built without any concrete data structures to look at.

Just use a dataclass that contains whatever information you want - that way you don't need to do any dictionary checks, and are just working with actual typed objects. You might have some dictionaries in the object (since the JSON itself is a dict), but data itself shouldn't be nested layer deep in an object this way

api/src/task/notifications/opportunity_notifcation.py
```diff
@@ -1,86 +1,443 @@
...
+        user_opportunity_pairs : list = []
+        none_versioned_opportunities: list = []
+        for user_saved_opp, latest_opp_ver in results:
+            if latest_opp_ver is None:
+                logger.error(
+                    "No prior version recorded for this opportunity;",
+                    extra={"opportunity_id": user_saved_opp.opportunity_id},
+                )
+                none_versioned_opportunities.append(user_saved_opp.opportunity_id)
```
chouinar: If the only thing we're making this list for is adding a metric, just cut the 3 lines of code for creating, adding, and incrementing to just:

```py
self.increment(Metrics.WHATEVER)
```
in this loop.

api/src/task/notifications/opportunity_notifcation.py
```diff
@@ -1,86 +1,443 @@
...
+
+        result = self.db_session.execute(stmt).all()
+        prior_notified_versions = {
+            (user_saved_opp.user_id, user_saved_opp.opportunity_id): version
+            for user_saved_opp, version in result
+        }
+
+        return prior_notified_versions
+
+    def _build_opportunity_content(self,  opp_id: UUID, diffs: list) -> str:
```
chouinar: Break this function into many functions, way too much is going on in it, and it's hard to follow. Testing this will be incredibly difficult as written as I'd say there are hundreds of scenarios possible as written, and we need to test that things work for all of those scenarios.

I'd suggest at least making each of the if-statement chunks a dedicated function that you can then write tests for (Note: I'd argue that building these is complex enough that you might want to completely separate these - the scope of this ticket is way larger than one ticket should be)

api/src/task/notifications/opportunity_notifcation.py
```diff
@@ -1,86 +1,443 @@
...
+        status_section = ""
+        important_section = ""
+        award_section = ""
+        category_section = ""
+        description_section = ""
+        eligibility_section = ""
+        grantor_section = ""
+        documents_section = ""
```
chouinar: Instead of this, why not just do something like:

```
sections = []

if <whatever>:
   sections.append("whatever text"}

...
return "<br>".join(sections)
```
I think you did it because you might process the diffs in a random order, BUT just invert the logic. You'll need to reshape the diff structure to be something like:
```
{
 "field": {"before": x, "after" y}
}
```
But then you can do something like:
```
if "opportunity_status" in diffs:
    # do stuff

elif award_diffs := diffs.keys() & AWARD_FIELDS.keys():
    # award_diffs is the exact set of keys that overlapped that we'd want to add
```

api/src/task/notifications/opportunity_notifcation.py
```diff
@@ -1,86 +1,443 @@
...
+    "24 hours a day, 7 days a week<br>"
+    "Closed on federal holidays"
+    "</div>"
+)
+
+FIELD_VALUE_MAP = {
+    OpportunityStatus.POSTED: "Open",
+    True: "Yes",
+    False: "No"
+}
```
chouinar: This map serves two different purposes, one to handle booleans, and one to handle opportunity status, make those separate.

api/src/task/notifications/opportunity_notifcation.py
```diff
@@ -1,86 +1,443 @@
...
+                    if not after:
+                        continue
+
+                category_section += f'{BULLET_POINTS_STYLING} {CATEGORIZATION_FIELDS[field]} {before} to {after}.<br>'
+
+            if field == "summary_description":
+                max_char = 250
+                truncated_after = after[:max_char]
+                if len(after) > max_char:
+                    truncated_after += f"... <a href='{self.frontend_base_url}/opportunity/{opp_id}' style='color:blue;'>Read full description</a>"
```
chouinar: I'm not familiar with HTML nuances, but the summary description often has a lot of HTML, what happens if the description is something like:

```
<p>
    some text that is 250+ chars long
</p>
```
You'd cut it off in the middle of the text and never close the p tag (or whatever tag), that might completely break the formatting of the rest of the email. I, don't know what the best approach is here, either somehow trimming or removing the HTML might help, but no idea.

api/src/task/notifications/opportunity_notifcation.py
```diff
@@ -1,86 +1,443 @@
...
+
+CONTACT_INFO = (
+    "<div>"
+    "If you have questions, please contact the Grants.gov Support Center:<br><br>"
+    "<a href='mailto:support@grants.gov'>support@grants.gov</a><br>"
+    "1-800-518-4726<br>"
+    "24 hours a day, 7 days a week<br>"
+    "Closed on federal holidays"
+    "</div>"
+)
```
chouinar: When we send an email, we send HTML + plain text and I think the email program handles rendering. What happens for the plain-text scenarios?

api/src/task/notifications/opportunity_notifcation.py
```diff
@@ -1,86 +1,443 @@
...
+        ).label("rn")
+
+        # Subquery that assigns row numbers
+        latest_versions_subq = (
+            select(OpportunityVersion, row_number)
+            .subquery()
+        )
+
+        latest_opp_version = aliased(OpportunityVersion, latest_versions_subq)
+
```
chouinar: For this, and the other big DB query function you added, can you add dedicated tests for just testing their functionality - that they return the expected results in a variety of scenarios? They're definitely using some more complex/advanced things that I'm less certain about whether they're right from just reading them.

api/src/task/notifications/opportunity_notifcation.py
```diff
@@ -1,86 +1,443 @@
...
+            if field == "opportunity_status":
+                before = FIELD_VALUE_MAP.get(before, before)
+                after = FIELD_VALUE_MAP.get(after, after)
+
+                status_section += (
+                    SECTION_STYLING.format("Status") +
+                    f"{BULLET_POINTS_STYLING} The status changed from {before} to {after}.<br><br>"
+                )
+
+            if field in IMPORTANT_DATE_FIELDS:
```
chouinar: Minor thing, but can you make all these `field in X` statements be `if/elif/elif...` to make clear it's only going to be one of them.

api/src/task/notifications/opportunity_notifcation.py
```diff
@@ -1,86 +1,443 @@
...
+                + eligibility_section
+                + grantor_section
+                + documents_section
+        )
+
+        return   change_sections
+
+    def _build_notification_content(
+        self, updated_saved_opps: dict[UUID, dict[str, OpportunityVersion]]
+    ) -> tuple[str ,str] | tuple[str,str, list] | tuple[None,None, None]:
```
chouinar: If a function is returning that many tuples of generic types, I suggest making a class that contains those so it's obvious what we're working with

chouinar (commented):
There is a lot going on here, the work to fetch the data and the work to generate an email are both massive in scope and entirely different problems. I'd really really recommend focusing this PR on one or the other. The number of tests that each needs is immense (I'll push back on any approval until many, many scenarios are accounted for in the tests with regards to how the data is setup).

api/src/services/opportunities_v1/opportunity_version.py
babebe: using separate schema

api/src/task/notifications/opportunity_notifcation.py
babebe: Want to avoid duplicates opportunity ids here and should have had it as a set but since we are logging the opportunity_ids that are not versioned, don't need to keep this metric? Should technically not run into this scenario but higher chance first time this job runs.

api/src/task/notifications/opportunity_notifcation.py
babebe: Using dataclass.

api/src/task/notifications/constants.py
babebe: updated

api/src/task/notifications/opportunity_notifcation.py
babebe: Added comments and tests

api/tests/src/task/notifications/test_opportunity_notification.py
```diff
@@ -4,178 +4,279 @@
...
+        self,
+        cli_runner,
+        db_session,
+        search_client,
+        enable_factory_create,
+        user,
+        user_with_email,
+        caplog,
+        monkeypatch,
+        email_notification_task,
```
chouinar:
```suggestion
        self,
        db_session,
        enable_factory_create,
        user,
        user_with_email,
        caplog,
        email_notification_task,
```
Cleanup the fixtures you're using, you're pulling in a bunch of stuff that goes unused

api/tests/src/task/notifications/test_opportunity_notification.py
```diff
@@ -4,178 +4,279 @@
...
+    @pytest.fixture
+    def email_notification_task(self, db_session, search_client, monkeypatch):
+        monkeypatch.setenv("AWS_PINPOINT_APP_ID", "test-app-id")
+        monkeypatch.setenv("FRONTEND_BASE_URL", "http://testhost:3000")
+        monkeypatch.setenv("ENABLE_OPPORTUNITY_NOTIFICATIONS", "true")
+        monkeypatch.setenv("ENABLE_SEARCH_NOTIFICATIONS", "false")
+        monkeypatch.setenv("ENABLE_CLOSING_DATE_NOTIFICATIONS", "false")
+
+        config = EmailNotificationConfig()
+        return EmailNotificationTask(db_session, search_client, config)
```
chouinar: For these tests, why not run the OpportunityNotificationTask directly rather than the general email task? Would be less configuration necessary.

Also note that you can just pass the values into the config instead of setting env vars first, you just have to use the alias names like:

```py
config = EmailNotificationConfig(AWS_PINPOINT_APP_ID="whatever")
```

api/tests/src/task/notifications/test_opportunity_notification.py
```diff
@@ -4,178 +4,279 @@
...
+        user,
+        user_with_email,
+        caplog,
+        monkeypatch,
+        email_notification_task,
+    ):
+        """Test that latest opportunity version is collected for each saved opportunity"""
+        # create a different user
+        user_2 = factories.UserFactory.create()
+        user_2 = link_user_with_email(user_2)
```
chouinar: Could do something like this if you want a 1-line approach `user_2 = LinkExternalUserFactory.create(...).user` since the link external user record has the user in it - wouldn't need to create them separately.

api/tests/src/task/notifications/test_opportunity_notification.py
```diff
@@ -4,178 +4,279 @@
...
+
+        for user_saved_opp, latest_opp_ver in results:
+            opp_id = user_saved_opp.opportunity_id
+
+            if opp_id == opp_1.opportunity_id:
+                assert latest_opp_ver == opp_1_v_2
+            elif opp_id == opp_2.opportunity_id:
+                assert latest_opp_ver == opp_2_v_1
+            elif opp_id == opp_3.opportunity_id:
+                assert latest_opp_ver == opp_3_v_2
```
chouinar: I'd suggest comparing ids rather than doing `lastest_opp_ver == <some DB model>` - not sure that behavior is defined. Just use the primary key ID instead

api/src/task/notifications/opportunity_notifcation.py
```diff
@@ -1,59 +1,105 @@
...
+                        entry.opportunities.append(version_change)
+                        break
+                else:
+                    changed_saved_opportunities.append(
+                        ChangedSavedOpportunity(
+                            user_id=user_id,
+                            email=user_saved_opp.user.email,
+                            opportunities=[version_change],
+                        )
+                    )
```
chouinar: I don't think this else would do anything? It doens't have an if-statement lined up with it (Is it somehow doing something from the for statement?)

api/src/task/notifications/opportunity_notifcation.py
```diff
@@ -63,24 +109,109 @@ def collect_email_notifications(self) -> list[UserEmailNotification]:
...
+        # Grabs latest version per (user, opportunity_id) pairs
+        stmt = select(
+            subq.c.user_id,
+            subq.c.opportunity_id,
+            opp_version_from_subq,
+        ).where(subq.c.rn == 1)
+
+        results = self.db_session.execute(stmt).all()
+
+        return {(row.user_id, row.opportunity_id): row[2] for row in results}
```
chouinar, line 213: Is `row[2]` the `opp_version_from_subq` value above in the select statement?

api/src/task/notifications/opportunity_notifcation.py
babebe: It is for the for-loop, runs if no break occurs within the for-loop

api/src/task/notifications/opportunity_notifcation.py
babebe: yes, added the comment

api/src/task/notifications/opportunity_notifcation.py
```diff
@@ -1,59 +1,105 @@
...
+
 
 class OpportunityNotificationTask(BaseNotificationTask):
-    def __init__(self, db_session: db.Session):
-        super().__init__(
-            db_session,
-        )
+    def __init__(self, db_session: db.Session, frontend_base_url: str | None = None):
+        super().__init__(db_session)
+        self.frontend_base_url = frontend_base_url
```
chouinar: Is this used anywhere?

api/src/task/notifications/opportunity_notifcation.py
chouinar: I've been staring at this for a few minutes, and I'm genuinely struggling to follow it.

Is this the same?

```py
                saved_opp_exists = False
                for entry in changed_saved_opportunities:
                    if entry.user_id == user_id:
                        entry.opportunities.append(version_change)
                        saved_opp_exists = True
                        break

               if not saved_opp_exists:
                   changed_saved_opportunities.append(...
```
I think I'd prefer a few more lines of code for clarity - at the very least, can you add a comment of what you're doing here?

api/tests/src/task/notifications/test_opportunity_notification.py
chouinar: Checking on this again

api/src/task/notifications/opportunity_notifcation.py
babebe: No. Removed.

api/src/task/notifications/opportunity_notifcation.py
babebe: yes it is the same logic. Replaced for more clarity.

api/tests/src/task/notifications/test_opportunity_notification.py
babebe: I need the first 2 env vars to be set since it will look for it when setting the config inside BaseNotificationTask. Updated.

PR #5370, opened 2025-06-17
[issue #5128] application attachments card

### LINKED ISSUE #5128: Create Required documentation card
### Summary

See Figma for design.

### Acceptance criteria

- [ ] Component
- [ ] Jest test
- [ ] Review with design/product

myduong-navapbc:
## Summary

Fixes #5128 

## Changes proposed

Added:
- The Application Attachment section
- api routes / handlers for `POST` & `DELETE` endpoints
- Simple Popover menu
- file size calculator (util)

Modified:
- The fetcher to also accept `POST` & `DELETE`
- spinner scss to include a small version

## Context for reviewers

Added Context to the Attachment section to allow for some basic & local state management.

Added basic state for uploading and deleting states 

## Validation steps

rebuild and start API (make remake-backend && docker compose up) and start dev server (npm run dev) and ensure that the feature flag applyFormPrototypeOff is disabled
go to http://localhost:3000/opportunity/114 or http://localhost:3000/opportunity/115
login
click start app button and fill out modal
application page should populate and the attachments card should be located at the bottom


frontend/src/app/api/applications/[applicationId]/attachments/[applicationAttachmentId]/handler.ts
```diff
@@ -0,0 +1,40 @@
...
+export const deleteAttachmentHandler = async (
+  _request: Request,
+  {
+    params,
+  }: {
+    params: Promise<{ applicationId: string; applicationAttachmentId: string }>;
+  },
+) => {
+  const applicationId = (await params).applicationId;
+  const applicationAttachmentId = (await params).applicationAttachmentId;
```
Comment (doug-s-nava) [inline frontend/src/app/api/applications/[applicationId]/attachments/[applicationAttachmentId]/handler.ts:16]: probably makes no difference but makes me a little confused to be awaiting params twice here
```suggestion
  const { applicationAttachmentId, applicationId } = await params;
```

frontend/src/app/api/applications/[applicationId]/attachments/handler.ts
```diff
@@ -0,0 +1,53 @@
...
+
+function createFormData(filename: string, buffer: Buffer, mimeType: string) {
+  const form = new FormData();
+  const file = new File([buffer], filename, { type: mimeType });
+  form.append("file_attachment", file);
+  return form;
+}
+
+export const postAttachmentHandler = async (
+  _request: Request,
```
Comment (doug-s-nava) [inline frontend/src/app/api/applications/[applicationId]/attachments/handler.ts]: if we're using the request arg in this handler we should remove the underscore from the arg name

frontend/src/app/api/applications/[applicationId]/attachments/handler.ts
```diff
@@ -0,0 +1,53 @@
...
+    return NextResponse.json({ error: "No file uploaded" }, { status: 400 });
+  }
+
+  const arrayBuffer = await file.arrayBuffer();
+  const buffer = Buffer.from(arrayBuffer);
+
+  try {
+    const session = await getSession();
+    if (!session || !session.token) {
+      throw new UnauthorizedError("No active session to get saved opportunity");
```
Comment (doug-s-nava) [inline frontend/src/app/api/applications/[applicationId]/attachments/handler.ts]: this looks like copy pasta?

frontend/src/app/api/applications/[applicationId]/attachments/handler.ts
```diff
@@ -0,0 +1,53 @@
...
+  const file = formData.get("file_attachment") as File;
+
+  if (!file) {
+    return NextResponse.json({ error: "No file uploaded" }, { status: 400 });
+  }
+
+  const arrayBuffer = await file.arrayBuffer();
+  const buffer = Buffer.from(arrayBuffer);
+
+  try {
```
Comment (doug-s-nava) [inline frontend/src/app/api/applications/[applicationId]/attachments/handler.ts:24]: does it make sense to move this to the top of the handler so we can fail fast?

frontend/src/components/PopoverMenu.tsx
```diff
@@ -0,0 +1,80 @@
+"use client";
+
+import { ReactNode, Ref, useEffect, useRef, useState } from "react";
+import { Button } from "@trussworks/react-uswds";
+
+import { USWDSIcon } from "./USWDSIcon";
+
+interface Props {
```
doug-s-nava: I think `React.PropsWithChildren` will work here, no need to create a new interface

frontend/src/components/PopoverMenu.tsx
```diff
@@ -0,0 +1,80 @@
...
+        handleOnTrigger={() => setOpen(!open)}
+        popoverState={open}
+      >
+        <USWDSIcon name="more_horiz" className="button-icon-md" />
+      </PopoverTrigger>
+
+      {open && (
+        <div
+          className="border border-base-light bg-white shadow-2 radius-md padding-1 text-no-wrap position-absolute z-top"
+          style={{ top: "100%", right: 0 }}
```
doug-s-nava: would adding `.top-full` and `right-0` work here instead?

frontend/src/components/application/attachments/AttachmentDeleteButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+  useSetPendingDeleteName,
+} from "src/context/application/AttachmentsContext";
+import { Attachment } from "src/types/attachmentTypes";
+
+import { RefObject } from "react";
+import { ModalRef, ModalToggleButton } from "@trussworks/react-uswds";
+
+import { USWDSIcon } from "src/components/USWDSIcon";
+
+interface Props {
```
doug-s-nava: can this be more descriptive? `AttachmentDeleteButtonProps`?

frontend/src/components/application/attachments/AttachmentDeleteModal.tsx
```diff
@@ -0,0 +1,83 @@
...
+  modalId,
+  modalRef,
+}: Props) => {
+  const pendingDeleteId = usePendingDeleteId();
+  const pendingDeleteName = usePendingDeleteName();
+  const deleteAttachment = useDeleteAttachment();
+  const setDeletingIds = useSetDeletingIds();
+
+  const handleDelete = async () => {
+    if (!pendingDeleteId) return;
```
doug-s-nava: [?] what's the use case where this modal would be visible with no pending delete id?

frontend/src/components/application/attachments/AttachmentDeleteModal.tsx
```diff
@@ -0,0 +1,83 @@
...
+    setDeletingIds((prev) => new Set(prev).add(pendingDeleteId));
```
doug-s-nava: do we ever remove pending delete ids from this set, or is that not necessary?

frontend/src/context/application/AttachmentsContext.tsx
```diff
@@ -0,0 +1,230 @@
...
+        }
+      } finally {
+        await refresh();
+      }
+    },
+    [applicationId, refresh, remove, updateStatus],
+  );
+
+  /* ------------------------------ delete ------------------------------ */
+  const deleteAttachment = useCallback(
```
doug-s-nava: I don't think there's going to be any way to wrap this in a client fetch since that requires a hook, which can only be used within a component. We could pass the client fetch function into the context, but that sounds like a potential nightmare. Is there a way to rethink this so that the fetch can come from the component, or is there a compelling reason to store the fetch function within the context?

frontend/src/components/application/attachments/AttachmentDeleteModal.tsx
```diff
@@ -0,0 +1,83 @@
...
+      modalRef.current?.toggleModal(),
+    );
+  };
+
+  return (
+    <SimplerModal
+      modalId={modalId}
+      modalRef={modalRef}
+      titleText={
+        pendingDeleteName
```
doug-s-nava: what is the use case where we wouldn't have a `pendingDeleteName` here?

frontend/src/components/application/attachments/AttachmentDeleteModal.tsx
```diff
@@ -0,0 +1,83 @@
...
+          : "Caution, deleting attachment"
+      }
+      className="text-wrap"
+    >
+      <p className="font-sans-2xs margin-y-4">{descriptionText}</p>
+      <ModalFooter>
+        <ButtonGroup>
+          <Button
+            type="button"
+            onClick={() => handleDelete() as unknown as void}
```
doug-s-nava: fwiw you can avoid this typing if you don't make handleDelete an async function and do it like this
```suggestion
    deleteAttachment(pendingDeleteId).catch(e => { throw e }).finally(() =>
      modalRef.current?.toggleModal(),
    );
```

frontend/src/components/application/attachments/AttachmentsCard.tsx
```diff
@@ -0,0 +1,249 @@
...
+  const deletingIds = useDeletingIds();
+
+  const t = useTranslations("Application.attachments");
+
+  const modalRef = useRef<ModalRef>(null);
+
+  const sortedAttachments = useMemo(() => {
+    return [...attachments].sort((a, b) => {
+      const aTyped = a;
+      const bTyped = b;
```
doug-s-nava: why rename these?

frontend/src/components/application/attachments/AttachmentsCard.tsx
```diff
@@ -0,0 +1,249 @@
...
+
+      if (aComparable < bComparable) return sortDirection === "asc" ? -1 : 1;
+      if (aComparable > bComparable) return sortDirection === "asc" ? 1 : -1;
+      return 0;
+    });
+  }, [attachments, sortBy, sortDirection]);
+
+  const handleUpload = async (file: File) => await uploadAttachment(file);
+
+  const handleAttachmentUpload = async (e: FormEvent<HTMLInputElement>) => {
```
doug-s-nava: two functions this similar makes me itchy. how would you feel about this
```suggestion
  const handleUpload = (files: FileList | null) => {
    if (!files || files.length === 0) return;
    uploadAttachment(files[0]).catch((e) => {
      throw e;
    });
  };


  <FileInput
    id="application-attachment-upload"
    name="application-attachment-upload"
    onChange={(e) => {
      const files = e.currentTarget.files;
      handleUpload(files || null);
    }}
    onDrop={(e) => {
      const files = e.dataTransfer.files;
      handleUpload(files);
    }}
    ref={fileInputRef}
  />;
```

frontend/src/components/application/attachments/AttachmentsCard.tsx
```diff
@@ -0,0 +1,249 @@
...
+    const files = event.dataTransfer.files;
+    if (!files || files.length === 0) return;
+    await handleUpload(files[0]);
+  };
+
+  const handleCancelUpload = (attachment: Attachment) => {
+    attachment.cancelToken?.abort();
+  };
+
+  const TableHeader = ({
```
doug-s-nava: let's move this to it's own file, we will likely want to use it on the search table at some point. Understood that that would mean bringing along the sort logic... Essentially I guess I'm saying we should separate the table from the upload form, but I think that's a good idea in any case

frontend/src/components/application/attachments/AttachmentsCard.tsx
```diff
@@ -0,0 +1,249 @@
...
+                    }
+                    key={file.application_attachment_id}
+                  >
+                    <td>
+                      {file.status === "uploading" && (
+                        <Spinner className="sm" />
+                      )}{" "}
+                      {file.file_name}
+                    </td>
+                    <td>
```
doug-s-nava: this cell is complex enough that I feel like it deserves its own component (or 2 or 3). That doesn't seem impossible from what I can see here

frontend/src/components/application/attachments/AttachmentsCard.tsx
```diff
@@ -0,0 +1,249 @@
...
+                                file={file}
+                                buttonText={t("delete")}
+                                modalRef={modalRef}
+                              />
+                            </PopoverMenu>
+                          </>
+                        )}
+                    </td>
+                    <td>
+                      {file.status !== "uploading" &&
```
doug-s-nava: `file.status !== "uploading" && !deletingIds.has(file.application_attachment_id)` is a popular value. Can we calculate that once and for all?

frontend/src/components/application/attachments/FormattedDate.tsx
```diff
@@ -0,0 +1,13 @@
...
+import { useEffect, useState } from "react";
+
+export const FormattedDate = ({ date }: { date: string }) => {
+  const [formatted, setFormatted] = useState("");
+
+  useEffect(() => {
+    setFormatted(formatDateTime(date));
+  }, [date]);
+
+  return <>{formatted}</>;
```
doug-s-nava: why make this its own component and why use state to control it?

frontend/src/app/api/applications/[applicationId]/handler.ts
```diff
@@ -0,0 +1,34 @@
...
+  try {
+    const session = await getSession();
+    if (!session || !session.token) {
+      throw new UnauthorizedError("No active session to get saved opportunity");
+    }
+
+    const response = await getApplicationDetails(applicationId, session.token);
+    const attachments = response.data.application_attachments;
+
+    return NextResponse.json(attachments, {
```
Comment (doug-s-nava) [inline frontend/src/app/api/applications/[applicationId]/handler.ts:22]: this looks like it's returning application attachments rather than application details. Can we update the naming and location of the endpoint? Would likely live more happily at `/api/applications/[applicationId]/attachments`

frontend/src/context/application/AttachmentsContext.tsx
```diff
@@ -0,0 +1,230 @@
...
+      setAttachments(json);
+    } catch (err) {
+      console.error("refreshAttachments:", (err as Error).message);
+    } finally {
+      clearInputAndRefreshRoute();
+    }
+  }, [applicationId, clearInputAndRefreshRoute]);
+
+  /* -------------------------- local state helpers ----------------------- */
+  const updateStatus = useCallback(
```
doug-s-nava: can this be more descriptive? `updateLocalAttachmentStatus`?

frontend/src/context/application/AttachmentsContext.tsx
```diff
@@ -0,0 +1,230 @@
...
+      setAttachments((prev) =>
+        prev.map((a) =>
+          a.application_attachment_id === id ? { ...a, status } : a,
+        ),
+      );
+    },
+    [],
+  );
+
+  const remove = useCallback((id: string) => {
```
doug-s-nava: `removeAttachment`?

frontend/src/context/application/AttachmentsContext.tsx
```diff
@@ -0,0 +1,230 @@
...
+
+// ---------------------------- consumer hooks -----------------------------
+export const useAttachmentsCtx = () => {
+  const ctx = useContext(AttachmentsContext);
+  if (!ctx)
+    throw new Error("useAttachmentsCtx must be inside AttachmentsProvider");
+  return ctx;
+};
+
+export const useDeletingIds = () => useAttachmentsCtx().deletingIds;
```
doug-s-nava: admittedly not using context, but our pattern with hooks so far is to have them return everything, rather than having individual hooks for each piece of data or function, and have components just destructure what they need. I'm not sure which approach is better but we should have a conversation about future direction

frontend/src/context/application/AttachmentsContext.tsx
```diff
@@ -0,0 +1,230 @@
...
+export const useSetDeletingIds = () => useAttachmentsCtx().setDeletingIds;
+export const usePendingDeleteId = () => useAttachmentsCtx().pendingDeleteId;
+export const useSetPendingDeleteId = () =>
+  useAttachmentsCtx().setPendingDeleteId;
+export const usePendingDeleteName = () => useAttachmentsCtx().pendingDeleteName;
+export const useSetPendingDeleteName = () =>
+  useAttachmentsCtx().setPendingDeleteName;
+
+export const useAttachmentsList = () => useAttachmentsCtx().attachments;
+export const useUploadAttachment = () => {
```
doug-s-nava: why is this one defined outside of the context?

frontend/src/styles/_uswds-theme-custom-styles.scss
```diff
@@ -499,10 +499,25 @@ button.usa-pagination__button.usa-button {
...
 .usa-table {
   table-layout: fixed;
-
   &.overflow-wrap {
     overflow-wrap: break-word;
   }
+  // specific to applications attachment
+  &.application-attachments-table {
+    td {
+      height: 65px;
```
doug-s-nava, line 508: any way to avoid hardcoding the height? does overflow work for especially long file names?

frontend/src/utils/formatFileSizeUtil.ts
```diff
@@ -0,0 +1,10 @@
+export const formatFileSize = (bytes: number, decimals = 1): string => {
```
doug-s-nava, line 1: not sure this needs its own file. Maybe we make a "fileUtils" and move `saveBlobToFile` over there as well?

frontend/src/components/PopoverMenu.tsx
myduong-navapbc: i have never actually used this one!! i like it

frontend/src/components/application/attachments/AttachmentDeleteModal.tsx
myduong-navapbc: this one, is a simple check- it should never have a null value but my ocd wont let it go since the pendingDeleteId is string | null

frontend/src/context/application/AttachmentsContext.tsx
myduong-navapbc: Well, technically, the provider is a React component. I dont think its much different than writing a wrapping or container component that passes functions down to its children. 

As of right now, this component is the only one using these functions but, a little further down the line- we will have forms that require users to upload files, which would end up in the attachments table- which they can reference the id. or delete from the attachment table would be useful.

I dont think its bad practice to put these calls in a context- they are wrapped in a useCallback, so, that does prevent  re-renders unless the data changes.

is there a reason why you want it to come directly from a component?

frontend/src/components/application/attachments/AttachmentDeleteModal.tsx
myduong-navapbc: there shouldnt be BUT it is typed as a string | null, we're saying theres a chance

frontend/src/components/application/attachments/AttachmentDeleteModal.tsx
myduong-navapbc: i like that ^

frontend/src/components/application/attachments/AttachmentsCard.tsx
myduong-navapbc: i was testing out a bunch of things and never cleaned this up.

PR #5375, opened 2025-06-17
[Issue #5286] Enable mTLS on a SOAP Proxy/Router specific ALB

### LINKED ISSUE #5286: Enable TLS passthrough on load balancers in all environments and ensure there is no unwarranted side effects
### Summary

To prepare for the authenticated SOAP Proxy to work with external partners, we need to do more preparation before sharing with them to test.

This work is to enable TLS passthrough on load balancers in all environments and ensure there is no unwarranted side effects.

### Acceptance criteria

- [ ] Enabled TLS Passthrough on load balancers
- [ ] Tested and confirmed for accuracy that there are no unwanted effects or abnormalities

mdragon:
Fixes #5286

## Summary
We need an ALB that has mutual TLS Client Certificate Passthrough on for the SOAP Proxy/Router but that prompts browsers to ask for a client certificate.

## Changes proposed

Move the SOAP Proxy/Router off to it's own ALB that is configured almost identically, but supports mTLS passthrough. This also gives us the flexibility to use a unique DNS name for this and split it's traffic load off to other containers in the future if needed.

## Validation steps

Ran the terraform apply locally to validate everything was working and happy both in API (which should get an mTLS ALB and frontend which should be totally unaffected by this change).

### DIFF
--- a/infra/modules/service/load_balancer.tf
+++ b/infra/modules/service/load_balancer.tf
@@ -6,8 +6,8 @@
 resource "aws_lb" "alb" {
   # we need an identical alb, with mtls enabled
   # so we piggy back off existing and just spin up two when the api sets this true
-  count = var.enable_load_balancer ? var.enable_mtls_load_balancer ? 2 : 1 : 0
-  depends_on      = [aws_s3_bucket_policy.access_logs]
+  count      = var.enable_load_balancer ? var.enable_mtls_load_balancer ? 2 : 1 : 0
+  depends_on = [aws_s3_bucket_policy.access_logs]
   # adjust name for the mtls alb that's in slot 1
   name            = count.index == 0 ? var.service_name : format("%s-mtls", var.service_name)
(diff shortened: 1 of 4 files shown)


infra/modules/service/load_balancer.tf
```diff
@@ -117,6 +126,7 @@ resource "aws_lb_listener_rule" "app_https_forward" {
   }
 }
 
+# these get referenced from the service/main file so we want two distinct onces for easier referencing by app_tg vs mtls_tg names
```
freddieyebra:
```suggestion
# these get referenced from the service/main file so we want two distinct ones for easier referencing by app_tg vs mtls_tg names
```

freddieyebra (commented):
minor spelling check otherwise looks good. may want second set of eyes since we verified this together.

freddieyebra (approved):

PR #5378, opened 2025-06-17
[Unticketed] Add sam.gov env vars

chouinar:
## Changes proposed
Adds sam.gov env vars to terraform

## Context for reviewers
We just got access to sam.gov - want to test it / see what failures we'll encounter before we test it more thoroughly. Need to have the key passed in via a secret + set the right URL.

I have set the secret in parameter store based on the value from 1password - for prod I just set a TODO value for when we get that key.

### DIFF
--- a/api/src/task/sam_extracts/fetch_sam_extracts.py
+++ b/api/src/task/sam_extracts/fetch_sam_extracts.py
@@ -25,7 +25,6 @@ from src.util import datetime_util
 logger = logging.getLogger(__name__)
 
 
-
 @task_blueprint.cli.command("fetch-sam-extracts", help="Fetch SAM.gov daily and monthly extracts")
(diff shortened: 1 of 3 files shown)


api/src/task/sam_extracts/fetch_sam_extracts.py
```diff
@@ -22,17 +21,10 @@
 from src.task.task import Task
 from src.task.task_blueprint import task_blueprint
 from src.util import datetime_util
-from src.util.env_config import PydanticBaseEnvConfig
 
 logger = logging.getLogger(__name__)
 
 
-class SamExtractsConfig(PydanticBaseEnvConfig):
```
chouinar, line 30: This isn't used - all the config we use is in the sam.gov client class. Removed to reduce confusion

freddieyebra (approved):
looks good!

Issue #4718, opened 2025-04-22
Introduce "pilling" for selected filter values


chrisw-nava:
### Summary:
Users should be able to quickly see which filter options they have selected without going through each filter item or opening the drawer to view them, so we will introduce "Pill" views of the filters 

🧑‍🎨 [Figma link](https://www.figma.com/design/FGxWtAgToKhehLJCiuy1zL/Simpler.Grants.gov-Project?node-id=5454-29793&t=uTmNMy5JzTQYy4sO-4)

🏁 Place behind a feature flag: newFilterViewOn (Enable/Disable, disabled by default until view is complete and tested)

### Acceptance Criteria:
- [ ] When a user selects filter options and clicks "Apply", add the selected filters as pills in the search view
- [ ] Pills should be stackable in a vertical manner. 
- [ ] If a user selects "All Department of X" or "All Agency X", show "All Department of X" or "All Agency of X" in the pill and not individual pills for subagencies, just one for the top level agency 
- [ ] If a user selects multiple of the same item, each selection otherwise should get its own pill
- [ ] The default status options of "Forecasted" and "Open" should display in the pill format when the user arrives on the screen

Notes 
5/20 - - @crystabelrangel to create a design with vertical stacking of pills to review for usability, will revisit AC after a decision is made
6/12 -- We are keeping stackable pills for the moment as it reduces a bit of scope

eric-serrato95:
does this also include the X to remove individual Pills as filters? @chrisw-nava

eric-serrato95:
"Basically do we want to require the user to click the right/left arrows or be able to scroll?" - Chris W
- Should we limit the amount of characters for the pills to avoid super long pills?
- Should we consider a design where the pills not only stack horizontally but also vertically? 
- Consideration for mobile or tablet view for when a user is querying with a long list of filters being applied

eric-serrato95:
@doug-s-nava AC is now updated and it reflects our conversation around just showing the top level agency as a pill and not each individual sub agency

PR #5380, opened 2025-06-17
[Issue #5355] top level agency search frontend implementation

### LINKED ISSUE #5355: Perform search for top-level agencies when selected
### Summary

Currently when a user selects "All <top level agency>" as a filter option in the agency filter, each sub agency of the top level agency is added to the query string and passed to the backend.

With https://github.com/HHS/simpler-grants-gov/issues/4785 we are now able to pass the top level agency instead of the list of sub agencies, and let the API sort out which sub agencies should be applied in the filter

The work here is to correctly pass the top level agency to the API from the frontend in this case, and correctly parse saved searches that contain a top level agency search.

### Acceptance criteria

- [ ] when a user selects an "All <top level agency>" filter option in the agency filter, the top level agency code is added to a "topLevelAgency" query parameter
- [ ] topLevelAgency value is passed correctly to API to allow for proper search results
- [ ] any saved searches containing topLevelAgency are correctly parsed by the frontend when retrieved for reapplication to search

doug-s-nava:
## Summary

Fixes #5355

## Changes proposed

Allows for the frontend to search explicitly by top level agency.

* fixes alignment of filter drawer button when logged in
* refactors agency filter components to establish the agency filter as a "unicorn" implementation and separate its specific needs from those of the other more straightforward filters as much as possible
* updates the AllOption and AnyOption check boxes to support the new top level query functionality
* updates the SearchFilterCheckbox to correctly reflect state where a parent top level option has been selected
* adds related helper functions

## Context for reviewers

This change is being made for two main reasons:

* to future proof saved searches against changes to a top level agency's sub agencies. If a user wants to save a search for a top level agency, they should receive, at any time, the search results related to the current sub agencies of that top level, not the frozen in time list of sub agencies from when the search was initially saved
* To allow for filter pills (see https://github.com/HHS/simpler-grants-gov/issues/4718) to list only top level agencies when a top level agency is selected, rather than overwhelming the UI with the full list of selected sub agencies

## Validation steps

3. open the agency filter and select an "all <top level agency>" check box
4. _VERIFY_: "Any" checkbox is unchecked, all child boxes are checked, agency code for top level agency is added as "topLevelAgency" query param value, agency query param is not present, search results update as expected
5. uncheck a checked child agency box
6. _VERIFY_: top level box is unchecked, all other child boxes remain checked, topLevelAgency query param is empty, agency query param is populated with list of remaining child agencies, search results update as expected
7. check two top level agency boxes
8. _VERIFY_: all expected children are selected, topLevelAgency query param contains comma separated values for each selected top level agency, search results update as expected
9. select a child agency with an unselected parent
10. select the parent
11. _VERIFY_: the agency query param for the previously selected child agency is removed while the top level agency param is added

BONUS: repeat from step 2 with the searchDrawerOn feature flag enabled using the agency filter from the drawer

### DIFF
--- a/frontend/src/components/search/Filters/AgencyFilterBody.tsx
+++ b/frontend/src/components/search/Filters/AgencyFilterBody.tsx
@@ -9,7 +9,7 @@ import {
   getSiblingOptionValues,
 } from "src/utils/search/searchUtils";
 
-import { useContext, useMemo } from "react";
+import { useCallback, useContext, useMemo } from "react";
 
 import { AnyOptionCheckbox } from "src/components/search/SearchFilterAccordion/AnyOptionCheckbox";
 import { SearchFilterAccordionProps } from "src/components/search/SearchFilterAccordion/SearchFilterAccordion";
@@ -19,29 +19,33 @@ import SearchFilterSection from "src/components/search/SearchFilterAccordion/Sea
 interface AgencyFilterBodyProps extends SearchFilterAccordionProps {
   referenceOptions?: FilterOption[];
   topLevelQuery?: Set<string>;
-  isParentSelected?: (value: string) => boolean;
 }
 
 // can rename to NestedCheckboxFilterBody if we can generalize a bit more
 export function AgencyFilterBody({
   includeAnyOption,
   title,
-  defaultEmptySelection,
   filterOptions,
   query,
   facetCounts,
   referenceOptions,
   topLevelQuery,
-  isParentSelected = () => false,
 }: AgencyFilterBodyProps) {
   const { queryTerm } = useContext(QueryContext);
   const { updateQueryParams, setQueryParams } = useSearchParamUpdater();
+
+  const isParentAgencySelected = useCallback(
+    (subAgencyCode: string): boolean => {
+      return topLevelQuery?.has(getAgencyParent(subAgencyCode)) || false;
+    },
+    [topLevelQuery],
+  );
 
   const toggleOptionChecked = (value: string, isChecked: boolean) => {
     const newParamValue = new Set(query);
     isChecked ? newParamValue.add(value) : newParamValue.delete(value);
     // happy path a normal checkbox experience
-    if (isChecked || !topLevelQuery || !isParentSelected(value)) {
+    if (isChecked || !topLevelQuery || !isParentAgencySelected(value)) {
       return updateQueryParams(newParamValue, "agency", queryTerm);
     }
     // handle unchecking a child box when top level parent is selected
@@ -71,7 +75,6 @@ export function AgencyFilterBody({
               title={title.toLowerCase()}
               checked={isNoneSelected}
               queryParamKey={"agency"}
-              defaultEmptySelection={defaultEmptySelection}
             />
           </li>
         )}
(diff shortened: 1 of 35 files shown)


frontend/src/components/search/Filters/CheckboxFilter.tsx
```diff
@@ -62,31 +53,13 @@ export function CheckboxFilterBody({
         )}
         {filterOptions.map((option) => (
           <li key={option.id}>
-            {/* If we have children, show a "section", otherwise show just a checkbox */}
-            {option.children ? (
```
doug-s-nava, line 66: since the agency filter is the only thing that uses this nested setup, moved that into the agency filter body to clean this up for vanilla implementations

frontend/src/components/search/SearchFilterAccordion/AgencyFilterAccordion.tsx
```diff
@@ -31,44 +34,16 @@ export async function AgencyFilterAccordion({
...
         title={t("accordion.titles.agency")}
         allAgencies={allAgencies}
         facetCounts={facetCounts}
+        topLevelQuery={topLevelQuery}
       />
     </BasicSearchFilterAccordion>
   );
 }
-
-export async function AgencyFilter({
```
doug-s-nava, line 46: this wasn't really doing much in the end

frontend/src/components/search/SearchVersionTwo.tsx
```diff
@@ -47,7 +47,7 @@ export function SearchVersionTwo({
       <QueryProvider>
         <div className="grid-container">
           <div className="desktop:display-flex desktop:margin-bottom-2">
-            <div className="flex-6">
+            <div className="flex-6 flex-align-self-end">
```
doug-s-nava, line 50: unrelated styling fix to correctly align the filter drawer button when logged in

frontend/src/components/search/Filters/AgencyFilterBody.tsx
```diff
@@ -0,0 +1,122 @@
...
+    const newParamValue = new Set(query);
+    isChecked ? newParamValue.add(value) : newParamValue.delete(value);
+    // happy path a normal checkbox experience
+    if (isChecked || !topLevelQuery || !isParentAgencySelected(value)) {
+      return updateQueryParams(newParamValue, "agency", queryTerm);
+    }
+    // handle unchecking a child box when top level parent is selected
+    // need to add children of selected top level, minus the one that has just been removed
+    // since they previously were not present anywhere in the url, but should remain selected
+    const siblingOptions = getSiblingOptionValues(value, filterOptions);
```
doug-s-nava, line 55: this is a complex edge case, but without it unchecking a child box when a parent box is selected wouldn't work right

acouch:
Seeing an issue with the agency count, the count doesn't reflect the number selected:

image

image

acouch:
I'd recommend a little more padding or something between "Any agency" checkbox and the first agency:

image

doug-s-nava:
@acouch good call out on the count for selected items. I noticed this at one point then forgot about it. Since the counts are based purely on the number of query param values present in the url, this is a little complex to fix and I'm not sure it's a high priority. @eric-serrato95 @crystabelrangel , can you weigh in on whether the agency count needs to be correct to proceed with this change, or whether we can ticket it for later?

doug-s-nava:
confirmed that @eric-serrato95 will create a follow up ticket to address the selected option count

Issue #5250, opened 2025-06-05
When a user logs in for the first time - check if they are an ebiz POC and create their org


chouinar:
### Summary

After #5248 has setup most of this logic for a backend task, we want to have a similar bit of logic added to our login flow.

When we first create a new login.gov user (see: https://github.com/HHS/simpler-grants-gov/blob/main/api/src/services/users/login_gov_callback_handler.py#L150 ) we also want to check if the email address is in the sam.gov entity table.

We ONLY want to do this if the user is new (a boolean we already determine in this logic and can be reused).

We want to do a query roughly like:
```sql
select * from sam_gov_entity where ebiz_poc_email='<whatever the user's email is>'
```

With the results we want to do the following if there are 1 or more results (note - most users won't be an EBIZ POC, so won't hit this logic):
* If an organization record for the sam.gov entity does not exist, create it
* Create an organization user record connecting the user to the organization just created, and mark that user as the ebiz POC
* Note that it's possible the org might have other EBIZ POCs already, that's perfectly fine, they won't be changed in any way, we're only potentially adding to the list.

This should be basically the same logic as #5248 and should reuse as much code from that as possible. We aren't going to worry about updating any time a user logs in because at least for now, we just want to check in two scenarios:
* We just created sam.gov entity records (which is when #5248 will run)
* We just made a new user (this ticket)
Any other time won't matter.

### Acceptance criteria

- [ ] EBIZ POC / org creation logic added when a user logs in the first time
- [ ] Tests updated

mikehgrantsgov:
@chouinar I think #5248 implement this ticket also? https://github.com/HHS/simpler-grants-gov/blob/main/api/src/services/users/login_gov_callback_handler.py#L165C9-L165C50

chouinar:
@mikehgrantsgov - I think on your PR you put the other ticket's ID instead of this one. In any case, this ticket has been finished

PR #5385, opened 2025-06-17
[Issue #5248] When a user logs in for the first time - check if they are an ebiz POC and create their org

### LINKED ISSUE #5248: Create a task that creates orgs from sam.gov entity data + ebiz POCs

mikehgrantsgov:
## Summary

Fixes #5248

## Changes proposed

For new users, link to organization as organization owner if applicable

Refactor `link_sam_gov_entity_if_not_exists` to be reused with optional metrics.
Add tests

## Context for reviewers

Heavily reuses logic in https://github.com/HHS/simpler-grants-gov/issues/5248, as most of the complexity in this task is reusing the `link_sam_gov_entity_if_not_exists` function.
Additional context provided in ticket: https://github.com/HHS/simpler-grants-gov/issues/5250

## Validation steps

See new unit tests.

### DIFF
--- a/api/src/services/users/login_gov_callback_handler.py
+++ b/api/src/services/users/login_gov_callback_handler.py
@@ -14,6 +14,7 @@ from src.auth.auth_errors import JwtValidationError
 from src.auth.login_gov_jwt_auth import get_login_gov_client_assertion, validate_token
 from src.constants.lookup_constants import ExternalUserType
 from src.db.models.user_models import LinkExternalUser, LoginGovState, User
+from src.services.users.organization_from_ebiz_poc import handle_ebiz_poc_organization_during_login
 from src.util.string_utils import is_valid_uuid
 
 logger = logging.getLogger(__name__)
@@ -157,6 +158,23 @@ def _process_token(db_session: db.Session, token: str, nonce: str) -> LoginGovCa
     # prior to us trying to work with the user further.
     # NOTE: This doesn't commit yet - but effectively moves the cache from memory to the DB transaction
     db_session.flush()
+
+    # Check if the user is an ebiz POC and create/link their organization
+    # Only do this for new users
+    if is_user_new:
+        org_user = handle_ebiz_poc_organization_during_login(db_session, external_user.user)
+
+        if org_user:
+            logger.info(
+                "User linked to organization as organization owner (ebiz POC)",
+                extra={
+                    "user_id": str(external_user.user.user_id),
+                    "organization_id": str(org_user.organization_id),
+                    "is_organization_owner": org_user.is_organization_owner,
+                },
+            )
+            # Flush again to ensure organization data is persisted
+            db_session.flush()
 
     token, user_token_session = create_jwt_for_user(
         external_user.user, db_session, email=external_user.email
--- a/api/src/services/users/login_gov_callback_handler.py
+++ b/api/src/services/users/login_gov_callback_handler.py
@@ -162,19 +162,7 @@ def _process_token(db_session: db.Session, token: str, nonce: str) -> LoginGovCa
     # Check if the user is an ebiz POC and create/link their organization
     # Only do this for new users
     if is_user_new:
-        org_user = handle_ebiz_poc_organization_during_login(db_session, external_user.user)
-
-        if org_user:
-            logger.info(
-                "User linked to organization as organization owner (ebiz POC)",
-                extra={
-                    "user_id": str(external_user.user.user_id),
-                    "organization_id": str(org_user.organization_id),
-                    "is_organization_owner": org_user.is_organization_owner,
-                },
-            )
-            # Flush again to ensure organization data is persisted
-            db_session.flush()
+        handle_ebiz_poc_organization_during_login(db_session, external_user.user)
 
     token, user_token_session = create_jwt_for_user(
         external_user.user, db_session, email=external_user.email
--- a/api/src/services/users/organization_from_ebiz_poc.py
+++ b/api/src/services/users/organization_from_ebiz_poc.py
@@ -1,0 +1,63 @@
+import logging
+
+from sqlalchemy import select
+from sqlalchemy.orm import selectinload
+
+import src.adapters.db as db
+from src.db.models.entity_models import SamGovEntity
+from src.db.models.user_models import OrganizationUser, User
+from src.util.sam_gov_utils import link_sam_gov_entity_if_not_exists
+
+logger = logging.getLogger(__name__)
+
+
+def find_sam_gov_entity_for_ebiz_poc(
+    db_session: db.Session, user_email: str
+) -> SamGovEntity | None:
+    """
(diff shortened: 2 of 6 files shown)


api/src/util/sam_gov_utils.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+    log_extra = {"sam_gov_entity_id": sam_gov_entity.sam_gov_entity_id, "user_id": user.user_id}
+    logger.info("Processing sam.gov entity record connection to user", extra=log_extra)
+
+    # If an organization does not already exist for the sam.gov entity
+    # create and associate it
+    organization = sam_gov_entity.organization
+    if organization is None:
+        if increment_metric:
+            increment_metric("new_organization_created_count")
```
chouinar: I like the idea here for making the metrics work. One idea if we want to have one less line for each of them could be to do something like:

```py
# Define a function that does nothing
def no_op(metric_name: str) -> None:
     pass

# in this function at the top
if increment_metric is None:
    increment_metric = no_op
```
Then we don't need the if statements in each spot, we just call a function that does nothing instead.

api/src/services/users/login_gov_callback_handler.py
```diff
@@ -158,6 +159,23 @@ def _process_token(db_session: db.Session, token: str, nonce: str) -> LoginGovCa
...
+            logger.info(
+                "User linked to organization as organization owner (ebiz POC)",
+                extra={
+                    "user_id": str(external_user.user.user_id),
+                    "organization_id": str(org_user.organization_id),
+                    "is_organization_owner": org_user.is_organization_owner,
+                },
+            )
+            # Flush again to ensure organization data is persisted
+            db_session.flush()
```
chouinar: I'd actually suggest trimming most of this - the function we call logs the same thing, and I don't think the flush is needed (really the flush above also could be worked around too, not sure why I did it like that).
```suggestion
    if is_user_new:
        handle_ebiz_poc_organization_during_login(db_session, external_user.user)
```

api/src/services/users/organization_from_ebiz_poc.py
```diff
@@ -0,0 +1,63 @@
...
+
+    # User is an ebiz POC, link them to the organization
+    logger.info(
+        "User is ebiz POC, linking to organization",
+        extra={
+            "user_id": str(user.user_id),
+            "user_email": user.email,
+            "sam_gov_entity_id": str(sam_gov_entity.sam_gov_entity_id),
+        },
+    )
```
chouinar: The function we call logs anything relevant already - I think we can trim most of this too. Also - don't log the users email.


```suggestion
     if not sam_gov_entities:
          return
```

api/src/services/users/organization_from_ebiz_poc.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+    return (
+        db_session.execute(
+            select(SamGovEntity)
+            .where(SamGovEntity.ebiz_poc_email == user_email)
+            .options(selectinload(SamGovEntity.organization))
+        )
+        .scalars()
+        .first()
+    )
```
chouinar: This should return a list - one person could be the owner of multiple organizations - we'd want to add them all.

api/src/util/sam_gov_utils.py
mikehgrantsgov: Good idea, I added this.

api/src/services/users/login_gov_callback_handler.py
mikehgrantsgov: Also updated this, much cleaner.

api/src/services/users/organization_from_ebiz_poc.py
mikehgrantsgov: Removed the logging and cleaned this up, too

api/src/services/users/organization_from_ebiz_poc.py
mikehgrantsgov: Got it, updated this and relevant tests

api/src/task/sam_extracts/create_orgs_from_sam_entity.py
```diff
@@ -31,55 +31,26 @@ def run_task(self) -> None:
...
-            self.db_session.add(organization_user)
-            logger.info("Added user to organization", extra=log_extra)
 
-        # As we know they're the ebiz POC from the initial query,
-        # make them the owner if they aren't already
-        if organization_user.is_organization_owner is not True:
-            self.increment(self.Metrics.NEW_ORGANIZATION_OWNER_COUNT)
-            organization_user.is_organization_owner = True
-            logger.info("Made user an owner of the organization", extra=log_extra)
+        def increment_with_task_metrics(metric_name: str) -> None:
```
mikehgrantsgov, line 44: @chouinar I fixed a merge conflict from main here, just flagging for review.

chouinar: @mikehgrantsgov - I removed a TODO from the logic that you copied - could you remove that as well?

PR #5395, opened 2025-06-18
[Issue #5123] Update start application auth handling

### LINKED ISSUE #5123: Implement correct error states applications
### Summary

If a user is on the prototype form or if when they are on the application page, their auth time out needs to be properly handled so they can no longer access the form and are prompted to re-login

### Acceptance criteria

.

acouch:
## Summary

Fixes #5123

## Changes proposed

1. Start app button is now visible for anon users and directs user to login if they aren't already logged in: image
2. Provides a specific error for "unauthorized" for if the user times out while submitting the app:
image

Keeps the same error if there is another error with the API:

image

## Context for reviewers

It would be ideal perhaps to provide more than an error state, but that would blow up the scope of the ticket.

## Validation steps

1. start dev server and api server locally
2. go to an opportunity with a competition (112 or 114). You can check the `opportunity_id` in the competition table in the database to be sure
3. click "Start new application" when not logged in. You should be directed to login
4. login and find you are rerouted back to the opp page
5. click "Start new application" and confirm the form still works
6. go back to opp and click button again. log out in another tab before submitting. submit the modal form and see that it directs you to login.

### DIFF
--- a/frontend/src/components/LoginModal.tsx
+++ b/frontend/src/components/LoginModal.tsx
@@ -39,40 +39,64 @@ export const LoginModal = ({
       titleText={titleText}
       className="text-wrap"
     >
-      <>
-        <p>{helpText}</p>
-        <p className="font-sans-2xs margin-y-4">{descriptionText}</p>
-        <ModalFooter>
-          <ButtonGroup>
-            <a
-              href={LOGIN_URL}
-              key="login-link"
-              className="usa-button"
-              onClick={() => {
-                const startURL = `${location.pathname}${location.search}`;
-                if (startURL !== "") {
-                  SessionStorage.setItem("login-redirect", startURL);
-                }
-              }}
-            >
-              {buttonText}
-              <USWDSIcon
-                className="usa-icon margin-right-05 margin-left-neg-05"
-                name="launch"
-                key="login-gov-link-icon"
-              />
-            </a>
-            <ModalToggleButton
-              modalRef={modalRef}
-              closer
-              unstyled
-              className="padding-105 text-center"
-            >
-              {closeText}
-            </ModalToggleButton>
-          </ButtonGroup>
-        </ModalFooter>
-      </>
+      <LoginModalBody
+        buttonText={buttonText}
+        closeText={closeText}
+        descriptionText={descriptionText}
+        helpText={helpText}
+        modalRef={modalRef}
+      />
     </SimplerModal>
+  );
+};
+
+export const LoginModalBody = ({
+  buttonText,
+  closeText,
+  descriptionText,
+  helpText,
+  modalRef,
+}: {
+  buttonText: string;
+  closeText: string;
+  descriptionText: string;
+  helpText: string;
+  modalRef: RefObject<ModalRef | null>;
+}) => {
+  return (
+    <>
+      <p>{helpText}</p>
+      <p className="font-sans-2xs margin-y-4">{descriptionText}</p>
+      <ModalFooter>
+        <ButtonGroup>
+          <a
+            href={LOGIN_URL}
+            key="login-link"
+            className="usa-button"
+            onClick={() => {
+              const startURL = `${location.pathname}${location.search}`;
+              if (startURL !== "") {
+                SessionStorage.setItem("login-redirect", startURL);
+              }
+            }}
+          >
+            {buttonText}
+            <USWDSIcon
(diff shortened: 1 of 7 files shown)


frontend/src/components/LoginModal.tsx
```diff
@@ -39,40 +39,64 @@ export const LoginModal = ({
...
+        closeText={closeText}
+        descriptionText={descriptionText}
+        helpText={helpText}
+        modalRef={modalRef}
+      />
     </SimplerModal>
   );
 };
+
+export const LoginModalBody = ({
```
acouch, line 53: There is no change to the body here, I just made it a separate component for re-use

frontend/src/components/user/OpportunityCompetitionStart.tsx
```diff
@@ -13,15 +12,10 @@ export const OpportunityCompetitionStart = ({
...
-  const { user } = useUser();
   const { checkFeatureFlag } = useFeatureFlags();
   const openCompetitions = selectOpenCompetitions({ competitions });
 
-  if (
-    !openCompetitions.length ||
-    !user?.token ||
-    checkFeatureFlag("applyFormPrototypeOff")
-  ) {
+  if (!openCompetitions.length || checkFeatureFlag("applyFormPrototypeOff")) {
```
acouch, line 18: Anon users can now see the start app button and are directed to login if they are not.

frontend/src/components/workspace/StartApplicationModal.tsx
```diff
@@ -82,22 +87,36 @@ const StartApplicationModal = ({
         modalRef={modalRef}
         className="text-wrap"
         modalId={modalId}
-        titleText={t("startAppplicationModal.title")}
+        titleText={
+          token
+            ? t("startAppplicationModal.title")
+            : t("startAppplicationModal.login")
```
acouch: The modal has a diff title if they are not logged in

frontend/src/components/workspace/StartApplicationModal.tsx
```diff
@@ -25,15 +26,19 @@ const StartApplicationModal = ({
   const { user } = useUser();
   const router = useRouter();
   const t = useTranslations("OpportunityListing");
+  const headerTranslation = useTranslations("HeaderLoginModal");
   const modalId = "start-application";
   const [validationError, setValidationError] = useState<string>();
   const [savedSearchName, setSavedSearchName] = useState<string>();
   const [error, setError] = useState<string>();
   const [loading, setLoading] = useState<boolean>();
+  const token = user && user.token ? user.token : null;
```
freddieyebra: You could simplify this a bit with:
```suggestion
  const token = user?.token || null;
```

freddieyebra (commented):
minor feedback, otherwise looks good!

frontend/src/components/workspace/StartApplicationModal.tsx
```diff
@@ -25,15 +26,19 @@ const StartApplicationModal = ({
...
 
   const handleSubmit = useCallback(() => {
-    if (!user?.token) return;
-
+    if (!token) {
+      setValidationError(t("startAppplicationModal.loggedOut"));
```
doug-s-nava: [sp] 
```suggestion
      setValidationError(t("startApplicationModal.loggedOut"));
```

doug-s-nava (approved):
I've been unable to reproduce a situation that would trigger the "You must be logged in to proceed." message. I get the standard error message if my token expires or I've logged out in another tab, which seems fine for now, as the case where a user's token has been deleted while the modal is open but before the request to create the application was made is pretty unlikely.

One thing to add to the list of todos is to use the `useClientFetch` hook for the call to `startApplication` so that an expired token on that call will log the user out

frontend/src/components/workspace/StartApplicationModal.tsx
acouch: Updated, thanks.

frontend/src/components/workspace/StartApplicationModal.tsx
acouch: Great catch, thanks.

acouch:
@doug-s-nava  I added a test to confirm. here is the workflow that worked for me. 

https://github.com/user-attachments/assets/c30c7c7f-7a5c-47af-b5f3-b22ee89d709d

PR #5402, opened 2025-06-20
[Unticketed] Update package versions for vulnerability scans

chouinar:
## Changes proposed

Updated packages for vulnerability scans - which then needed the following fixes:
* MyPy ignore for how we setup Metrics class in a few cases (might be a better way to handle typing, didn't look too deep and just ignored)
* MyPy recognized that the filename can be null for our attachment work, added some guardrails and tests to account for it
* Alembic had a new warning for a config value they added being recommended, set it to the default (shouldn't matter for us I think?)
* An ignore we had (B1) for some logging stuff is no longer in the ruff rules

### DIFF
--- a/.grype.yml
+++ b/.grype.yml
@@ -27,6 +27,9 @@ ignore:
   # Last checked: May 20th, 2025
   - vulnerability: CVE-2025-4516
 
+  # This is a vulnerability in html.parser.HTMLParser which we don't use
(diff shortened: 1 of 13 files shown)


mikehgrantsgov (approved):
Looks good!

freddieyebra (approved):

PR #5404, opened 2025-06-20
[Issue #5003] Update content opportunity notification - Opportunity Status

### LINKED ISSUE #5003: Implementation: Email notification content for changes to saved opportunities

babebe:
## Summary

Fixes #5003

## Changes proposed

Updated User Notification Content per [Doc] For opportunity status only(https://docs.google.com/document/d/1ypejg8hL7uOXdrI7auwf9SWQ1Ag07o7rCjnrVCLfA-s/edit?tab=t.0)


api/src/task/notifications/opportunity_notifcation.py
```diff
@@ -1,86 +1,432 @@
...
+            else "Your saved funding opportunity changed on "
+            f"<a href='{self.frontend_base_url}' target='_blank' style='color:blue;'>Simpler.Grants.gov</a>"
+        )
+        return UserOpportunityUpdateContent(
+            subject=subject,
+            message=intro + all_sections + closing_msg,
+            updated_opportunity_ids=updated_opp_ids,
+        )
+
+    def _build_sections(self, opp_change: OpportunityVersionChange) -> str:
```
chouinar: Pre-emptively before I review this in much depth, I'm going to be looking for thorough tests verifying the behavior of this building logic, that is - for every obvious scenario - we need to verify that it behaves as expected and structures data as we expect. I'd guess we're looking at easily 100+ scenarios across all of these fields.

I'd recommend we have very focused tests that are just checking the string logic works as expected without any of the DB/querying stuff around it.

api/tests/src/task/notifications/test_opportunity_notification.py
```diff
@@ -1,181 +1,528 @@
...
+        cascade_delete_from_db_table(db_session, Opportunity)
+        cascade_delete_from_db_table(db_session, UserSavedOpportunity)
+
+    @pytest.fixture
+    def user_with_email(self, db_session, user, monkeypatch):
+        return link_user_with_email(user)
+
+    def test_email_notifications_collection(
+        self,
+        cli_runner,
```
chouinar: For these tests, take a look at the things you're pulling in, only pull in what you need - we don't need a cli_runner for these tests

api/tests/src/task/notifications/test_opportunity_notification.py
```diff
@@ -1,181 +1,528 @@
...
+        opp_summary.close_date = date(2025, 7, 19)
+        opp_summary.forecasted_award_date = date(2025, 10, 10)
+        opp_summary.forecasted_project_start_date = date(2025, 5, 5)
+        opp_summary.fiscal_year = 2025
+
+        opp_p = factories.OpportunityVersionFactory.create(opportunity=opp)
+        # Update opp and create second version
+        opp.current_opportunity_summary.opportunity_summary.close_date += timedelta(days=10)
+        opp.current_opportunity_summary.opportunity_summary.forecasted_award_date += timedelta(
+            days=1
```
chouinar: This test is doing way too much, and not enough.

For testing the individual functions that build the strings, minimize how much you're setting up, don't setup full DB objects (don't use any factories). All you want to test here is that give two JSON objects, it diffs them and builds strings. Just test that, don't add all of the factory layers to the top of this.

I say this because the actual number of scenarios we need to test for something like this is immense, just for the dates:
* Various values being set (null is possible for several date fields).
* Having varying fields having changed (or none at all).

api/tests/src/task/notifications/test_opportunity_notification.py
babebe: cli_runner needed for checking log_records.

api/src/task/notifications/opportunity_notifcation.py
```diff
@@ -205,13 +265,107 @@ def _get_last_notified_versions(
...
+        all_sections = ""
+        updated_opp_ids = []
+        opp_numb = 0
+        # Get sections statement
+        for opp in updated_opportunities:
+            opp_id = opp.opportunity_id
+            sections = self._build_sections(opp)
+            if not sections:
+                continue
+            assert opp.previous is not None
```
chouinar: Asserts in non-tests are generally bad form (depending on how you run python, they might do nothing). If this is a scenario where encountering it would be a very big problem that should cause the job to fall over, just make it an explicit exception we're raising.

api/src/task/notifications/opportunity_notifcation.py
```diff
@@ -205,13 +265,107 @@ def _get_last_notified_versions(
...
+        closing_msg = (
+            "<div>"
+            "<strong>Please carefully read the opportunity listing pages to review all changes.</strong> <br><br>"
+            f"<a href='{self.notification_config.frontend_base_url}' target='_blank' style='color:blue;'>Sign in to Simpler.Grants.gov to manage your saved opportunities.</a>"
+            "</div>"
+        ) + CONTACT_INFO
+
+        all_sections = ""
+        updated_opp_ids = []
+        opp_numb = 0
```
chouinar: Just because it tripped me up, since this count is used for the email content, can we name it something like `opp_count` as I was interpreting it as the opportunity number field.

Also, if we're putting the count in the email content, should we start at 1 instead of 0?

api/tests/src/task/notifications/test_opportunity_notification.py
```diff
@@ -253,3 +426,146 @@ def test_last_notified_version(
 
         assert results[user.user_id, opp.opportunity_id] == v_1
         assert results[user_2.user_id, opp.opportunity_id] == v_2
+
+    @pytest.mark.parametrize(
+        "diff_dict,expected_dict",
+        [
+            (
+                # qSingle field with nested name
```
chouinar:
```suggestion
                # Single field with nested name
```

api/src/task/notifications/opportunity_notifcation.py
babebe: This is for the type checker. I have it as possibly `None` since i grab that at a later time in collect_email_notifications. But for this to be called it would need to have not been None.

api/src/task/notifications/opportunity_notifcation.py
babebe: Ah yes, updated

api/src/task/notifications/opportunity_notifcation.py
chouinar: Can we change it to a non-assert? asserts should not be used outside of tests as they're not guaranteed to actually do anything. It might be purely for type-checking, but it's bad practice.

If we know for certain it's never None here we can `cast`, if there is a technical possibility it could be None, do an if statement.

api/tests/src/task/notifications/test_opportunity_notification.py
chouinar: If you want to check logs you just need to pull in caplog. By default caplog doesn't catch info logs, so you need to add a line like:
```py
caplog.set_level(logging.INFO)
```
to your test. Pulling in cli_runner is setting up a whole bunch of stuff we don't want for most tests as it effectively runs the entire `create_app` function https://github.com/HHS/simpler-grants-gov/blob/main/api/src/app.py#L70 - which we don't want that overhead on tests that don't need it.

api/src/task/notifications/opportunity_notifcation.py
babebe: asserting non null value using cast

api/tests/src/task/notifications/test_opportunity_notification.py
babebe: Got it ! Thanks! Updated

chouinar (approved):

PR #5405, opened 2025-06-20
[Issue #5306] Set up table for short-lived token

### LINKED ISSUE #5306: Set up table for short-lived token
### Summary

This work is to set up a table for the short-lived tokens for each form that will be submitted.

Create a DB table named `short_lived_internal_token` with the following columns:
* `short_lived_internal_token_id` - uuid, primary key
* `expires_at` - datetime, not nullable
* `is_valid` - boolean, not nullable, default True
* Our usual created_at/updated_at from the mixin

### Notes from conversation 6/11/2025
- Information needed: Application ID, Token, Form ID
- FE looking for a single route: PDF/AppID-FormID-Token

### Acceptance criteria

- [ ] Table for short-lived token is created

mikehgrantsgov:
## Summary

Fixes #5306 

## Changes proposed

Create `ShortLivedInternalToken` model.
Add migration

## Context for reviewers

This work is to set up a table for the short-lived tokens for each form that will be submitted.

## Validation steps

See applied migration


api/src/db/models/competition_models.py
```diff
@@ -324,6 +324,27 @@ def download_path(self) -> str:
...
+
+    application_id: Mapped[uuid.UUID] = mapped_column(
+        UUID, ForeignKey(Application.application_id), nullable=False, index=True
+    )
+    application: Mapped[Application] = relationship(Application)
+
+    form_id: Mapped[uuid.UUID] = mapped_column(
+        UUID, ForeignKey(Form.form_id), nullable=False, index=True
+    )
+    form: Mapped[Form] = relationship(Form)
```
chouinar: We don't need any reference to these tables - while the token will be used for fetching that info, no need to complicate it (I think I was planning to just reuse a token for all forms in an app)

mikehgrantsgov: Removed these!

chouinar (approved):

PR #5410, opened 2025-06-23
[Issue #5393] Setup an SF424 Form's JSON Schema

### LINKED ISSUE #5393: Create JSON for form SF424 locally
### Summary

To ensure our approach to building forms will work as expected, we want to start on building out a form in it's entirety even though we have not identified our pilot opportunity.

This work is to finish the backend work needed to create form SF424 locally for testing purposes.

### Acceptance criteria

- [ ] JSON created (locally) for form SF424 so that we can create and test the form on the front-end

chouinar:
## Summary

Fixes #5393 

## Changes proposed
Setup the JSON schema for the SF424

Some utilities in our DB seed local script for updating the SF424 DB record (may expand in a future PR to also update non-local DB).

## Context for reviewers
Long-term, I don't know what exactly our approach to defining JSON schemas will be, we probably can't fully commit to doing it in code like I did here, but for the pilot I think it's fine / gives us some simple ways to test and verify it's setup correctly. Nothing built here can't be moved elsewhere when we figure that out.

## Validation steps
I did verify this works with the frontend, although needs a lot of cleanup on the text displayed, various types, and other UI schema fun.

Run `make db-seed-local` - this will create a competition with this new form attached to it, and print out a link to the opportunity on your local frontend server. Go to that, and if you login + have feature flags setup right, you can start an application. You can then navigate to the app form page, and see all the fields - a few aren't working, but I'll defer to the frontend to fix/cleanup that.
Screenshot 2025-06-23 at 12 01 55 PM
Screenshot 2025-06-23 at 12 06 44 PM


api/src/form_schema/forms/sf424.py
```diff
@@ -0,0 +1,781 @@
...
+        "CountryCode": {
+            "type": "string",
+            "title": "Country",
+            "description": "Country Code",
+            "enum": shared_schema.COUNTRIES,
+        },
+    },
+}
+
+FORM_UI_SCHEMA = [
```
chouinar, line 658: @acouch / @myduong-navapbc - I setup the laziest UI schema possible and just shoved everything in one section - just so you wouldn't need to figure out which fields we have.

Wasn't sure how nested fields need to be formatted, what I have commented out for those fields didn't work, but also didn't look too much into it further.

freddieyebra:
@chouinar working through validation and wondering if the `applyFormPrototypeOff` is the needed feature flag, and what would be the url of the form?

chouinar:
> @chouinar working through validation and wondering if the `applyFormPrototypeOff` is the needed feature flag, and what would be the url of the form?
@freddieyebra 
This is how I have my feature flags setup at http://localhost:3000/dev/feature-flags 

Screenshot 2025-06-26 at 11 53 25 AM

Once you're logged in (which requires first setting `LOGIN_FINAL_DESTINATION=http://localhost:3000/api/auth/callback` as an env var on the API side + some private key stuff on the frontend) - if you just go to the opportunity page you should see an apply button that lets you navigate to the form. Note - the form is rough, really not ready to be looked at on the frontend, that aspect of setting it up will be mostly done by the frontend. I just verified that nothing was broken and it could interpret my JSON schema - probably not too important to try and validate from that perspective yet.

frontend/src/components/applyForm/validate.ts
```diff
@@ -29,7 +29,7 @@ export const UiJsonSchema: RJSFSchema = {
         },
         definition: {
           type: "string",
-          pattern: "^/properties/[a-zA-Z0-9]+$",
+          pattern: "^/properties/[a-zA-Z0-9_]+$",
```
chouinar: @acouch / @myduong-navapbc  and any other frontend folks. Let me know if you'd like to defer making this change to the frontend folks when you get into adjusting the form page.

The UI schema didn't allow property names to have underscores in them, and I named many with that. I don't need to make this in this PR, just needed if you want to see the form schema in the UI right now with my really lazily made UI schema

chouinar: I think I'll undo this so I don't need to fix frontend tests

PR #5417, opened 2025-06-23
Mikehgrantsgov/5307 add auth short lived token

### LINKED ISSUE #5307: Implement auth approach for using short-lived token
### Summary

This work is to define an approach for authentication for using the short-lived token.

The short-lived token table was created in ticket #5306 .

---
We want to create a new JWT-auth approach for a short-lived token. This is effectively a much more focused and simpler version of the user JWT auth we have already in `api_jwt_auth.py`. For implementing this change, reusing some of that code is highly recommended, you might need to refactor it slightly to not include certain things (anything login.gov, or user based isn't needed for this type of auth).

We'll want to create a new `JwtUserHttpTokenAuth` with:
* scheme: `ApiKey`
* header: `X-SGG-Internal-Token`
* security_scheme_name: `InternalApiJwtAuth`
You'll also need to modify the `get_app_security_scheme` function in auth_utils to include this new auth approach as that's needed by the API to route auth appropriately.

We'll need to create two main functions:
* A way to create the JWT (similar to the user-based `create_jwt_for_user`, but without any user stuff) - let the user pass in an expiration time - we'll set that in the process that is going to call this. This should create a record in the table setup in #5306
* A way to validate the JWT (similar to the `decode_token` function) which should fetch the record from the DB after doing basic JWT validation. The decode_token function should return the `short_lived_internal_token` table object from #5306

Other details:
* Feel free to reuse the JWT config we already have, we don't need to add new private/public keys for this.
* All the rules / options for JWT parsing should be the same (same `verify_x` options).

We DO NOT need any of the following:
* Token refreshing
* Anything connected to users/emails
* Any way for a user to actually get this token
* Add it to an endpoint (a follow-up ticket will do that).

### Acceptance criteria

- [ ] Authentication approach for using short-lived token is defined and reviewed with engineering
- [ ] Link to any resulting documentation included in this ticket

mikehgrantsgov:
## Summary

Fixes #5307

## Changes proposed

Added `JwtUserHttpTokenAuth` and utility functions to create, parse, and decode. 
New auth method added to `get_app_security_scheme`
Add test cases for wide variety of scenarios
Add technical Markdown documentation

## Context for reviewers

Implements a JWT-based short-lived token authentication system for internal API use, providing the foundational infrastructure needed to support form-to-PDF transformation, possibly other use cases.

## Validation steps

See unit tests


api/src/auth/internal_jwt_auth.py
```diff
@@ -0,0 +1,159 @@
...
+    """
+
+    try:
+        short_lived_token = parse_jwt_for_internal_token(token, db_session)
+
+        add_extra_data_to_current_request_logs(
+            {
+                "auth.token_id": str(short_lived_token.short_lived_internal_token_id),
+            }
+        )
```
chouinar: Just so we can easily distinguish the user types in our New Relic logs/graphs, could we call this `auth.short_lived_internal_token_id`?

mikehgrantsgov: Makes sense, updated this.

chouinar (approved):

PR #5420, opened 2025-06-23
[Issue #5364] Multiple top level agency filter

### LINKED ISSUE #5364: Allow filtering on multiple top level agencies
### Summary

Currently the `top_level_agency` param passed to the API search endpoint only accepts a single string. Since this endpoint needs to support searching based on multiple top level agencies, this should be adjusted to accept multiple values, likely following the same pattern as the "one_of" filters.

Also, the `top_level_agency` parameter should be moved from the top level of the payload body into the "filters" section

### Acceptance criteria

- [ ] "top_level_agency" filter accepts multiple values
- [ ] "top_level_agency" filter is positioned within "filters"

babebe:
## Summary

Fixes #5364 

## Changes proposed

Update top_level_agency to be a filter param
Update filter_should_terms to take a list of filter prefixes
Updates tests


api/src/services/opportunities_v1/search_opportunities.py
```diff
@@ -182,15 +186,16 @@ def _get_search_request(params: SearchOpportunityParams, aggregation: bool = Tru
...
         filter_rule = FILTER_RULE_MAPPING.get(params.experimental.scoring_rule, DEFAULT)
         builder.simple_query(params.query, filter_rule, params.query_operator)
 
     # Filter Prefix
-    if params.top_level_agency:
-        _add_top_level_agency_prefix(builder, params.top_level_agency, params.filters)
+    if params.filters and params.filters.top_level_agency:
+        _add_top_level_agency_prefix(builder, params.filters.top_level_agency, params.filters)
+        # remove top level agency from filter
+        params.filters.top_level_agency = None
```
chouinar: Rather than pass in a specific filter + the whole filter object, could we just pass in the filter object?

Something like this here:
```py
_add_top_level_agency_prefix(builder, params.filters)
```

And modify that function to instead have the checks if the values are null.

api/src/api/opportunities_v1/opportunity_schemas.py
```diff
@@ -467,12 +473,7 @@ class OpportunitySearchRequestV1Schema(Schema):
     )
 
     filters = fields.Nested(OpportunitySearchFilterV1Schema())
-    top_level_agency = fields.String(
-        metadata={
-            "description": "Top level agency for querying against database for it's sub_agencies",
-        },
-        validate=[validators.Length(min=2)],
-    )
```
chouinar, line 475: @doug-s-nava - just confirming that you aren't currently using the top-level-agency in the GET /opportunity endpoint. I don't think I see anything from a quick search of the frontend code, but want to be certain. Otherwise gotta work around that and keep this temporarily.

doug-s-nava: confirmed - the frontend isn't using `top_level_agency` at all yet

api/src/services/opportunities_v1/search_opportunities.py
babebe: Ah yes! Updated

api/src/services/opportunities_v1/search_opportunities.py
```diff
@@ -182,15 +182,16 @@ def _get_search_request(params: SearchOpportunityParams, aggregation: bool = Tru
...
         filter_rule = FILTER_RULE_MAPPING.get(params.experimental.scoring_rule, DEFAULT)
         builder.simple_query(params.query, filter_rule, params.query_operator)
 
     # Filter Prefix
-    if params.top_level_agency:
-        _add_top_level_agency_prefix(builder, params.top_level_agency, params.filters)
+    if params.filters and params.filters.top_level_agency:
+        _add_top_level_agency_prefix(builder, params.filters)
+        # remove top level agency from filter
+        params.filters.top_level_agency = None
```
chouinar: Could we move all of this inside the `_add_top_level_agency_prefix` function? Make this just: `_add_top_level_agency_prefix(builder, params.filters)` and in that function return if the values are None. That way the logic of how this works isn't split across multiple places, and is contained in one place.

PR #5422, opened 2025-06-23
[Unticketed]: Add automation for local soap proxy to training env

freddieyebra:
## Summary

Fixes unticketed

## Changes proposed

This pr adds support for setting up the soap proxy to request authenticated requests locally to the existing GG SOAP API.

## Context for reviewers

Currently, there is only a mock service that responds with fake data, this pr will support proxying data against the existing GG SOAP API in the training environment.

## Validation steps

Checkout this branch and follow the steps documented [here](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/1933639704/Simpler+SOAP+API+Dev+Setup#Proxy-GG-SOAP-API-Locally)


api/Makefile
```diff
@@ -354,3 +356,6 @@ init-mock-soap-services: start-mock-soap-services ## Initialize the applicants s
 
 start-mock-soap-services: ## start applicants soap api
 	docker compose up --detach mock-applicants-soap-api
+
+setup-soap-proxy:
+	$(SETUP_OVERRIDE_ENV_CMD) --setup-soap
```
chouinar: Just throwing out another idea if it might make sense. Rather than have a configurational soap proxy part of the setup override env file, we could add another .env file for this.

Could be simpler / more self-contained as I doubt most folks running the API will interact with it.

Would just need to add it as another env file in the docker-compose file.

api/bin/setup-env-override-file.sh
```diff
@@ -95,9 +121,39 @@ EOF
...
+  "`openssl x509 -in $applicantspub -noout -fingerprint -sha256 | sed 's/://g' | cut -d'=' -f2 | tr '[:upper:]' '[:lower:]'`" "`cat ${applicantspk}`" "`cat ${applicantspub}`" \
+  "`openssl x509 -in $grantorspub -noout -fingerprint -sha256 | sed 's/://g' | cut -d'=' -f2 | tr '[:upper:]' '[:lower:]'`" "`cat ${grantorspk}`" "`cat ${grantorspub}`")
+
+  print_log "Setting soap proxy auth variables in $OVERRIDE_FILE"
+  cat <<EOF >> $OVERRIDE_FILE
+
+#############################################
+# SOAP Auth for proxying training locally
+#############################################
+SOAP_AUTH_CONTENT=$(printf '%s' "$soap_auth_content" | python3 -c 'import json,sys; print(json.dumps(sys.stdin.read()))')
```
chouinar: If you want to do bash operations on JSON, you can use `jq` for that usually.

api/bin/setup-env-override-file.sh
```diff
@@ -95,9 +121,39 @@ EOF
...
+  applicantspk=$SOAP_CERT_FOLDER_PATH$APPLICANTS_BASE_CERT_NAME.key
+  grantorspub=$SOAP_CERT_FOLDER_PATH$GRANTORS_BASE_CERT_NAME.crt
+  grantorspk=$SOAP_CERT_FOLDER_PATH$GRANTORS_BASE_CERT_NAME.key
+
+  for file in $applicantspub $applicantspk $grantorspub $grantorspk; do
+    if [ ! -e "$file" ]; then
+      print_err "Missing $file aborting soap setup. Double check SOAP_CERT_FOLDER_PATH var"
+      return 0
+    fi
+  done
```
chouinar: What would I need locally setup in order to run this?

api/bin/setup-env-override-file.sh
```diff
@@ -7,11 +7,17 @@
...
 #   ./setup-env-override-file.sh --recreate
+#   ./setup-env-override-file.sh --recreate --setup-soap
+#   ./setup-env-override-file.sh --setup-soap
 #
 
 set -o errexit -o pipefail
 
 PROGRAM_NAME=$(basename "$0")
+OVERRIDE_FILE="override.env"
+SOAP_CERT_FOLDER_PATH=$HOME/Downloads/grants_s2s_soap_certs/localsetup/
```
chouinar: This doesn't seem like a path we should have defined in a script, should we make the SOAP cert setup a bit more intentional? If I want to setup the proxy to work, maybe the command I run should be something like:

```sh
setup-soap-proxy-overrides /path/to/folder
```

The setup-env-override-file script is meant to be mostly automated and almost never run manually, while you've set it up to require that. I think we should separate it out

api/bin/setup-env-override-file.sh
freddieyebra: there was some issues with trying to use that and this was the only parsing that made loading it env var happy.

api/Makefile
freddieyebra: updated to add separate script.

api/bin/setup-env-override-file.sh
freddieyebra: I added some docs [here](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/1933639704/Simpler+SOAP+API+Dev+Setup#Proxy-GG-SOAP-API-Locally)

api/bin/setup-env-override-file.sh
freddieyebra: refactored this.

api/bin/setup-soap.sh
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env bash
+
+# This is a script to run setup for doing development on the simpler soap api.
+# The certificates must be downloaded from OnePassword and then unzipped.
+# The path of that unzipped file must be passed to this script and defaults to
+# the ~/Downloads folder.
```
chouinar: Could you specify which thing in 1password we'd want to download / what files?

freddieyebra: updated!

chouinar (approved):

PR #5434, opened 2025-06-24
[Issue #5308] Add auth to get application endpoints

### LINKED ISSUE #5308: Add auth approach to Get-Application form endpoint
### Summary

This work is to integrate the auth approach defined in #5307 to the GET APPLICATION FORM endpoint.

---
We want to modify the GET /application form endpoint to allow for a user using our new Internal JWT auth to be allowed to request. To add this auth to the endpoint, use our MultiAuth (like we do on [competition](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/api/competition_alpha/competition_route.py#L19), but note we'll need a different multi-auth object for JWT or internal JWT auth).

The multi-auth class itself may need to be modified to be able to fetch either the UserTokenSession or the new table object.

The main complexity with this change is that we need to make it so if the user is of this type of auth, then we don't do any of the auth checks we have for fetching the application form itself. We should roughly change the logic to:

```
if isinstance(user_auth_obj, UserTokenSession):
    # validate user can access as we already do
else:
    # Do not check - think of them like an admin user
```

### Acceptance criteria

- [ ] Authentication approach added GET APPLICATION FORM endpoint

mikehgrantsgov:
## Summary

Fixes #5308

## Changes proposed

Added internal JWT authentication to the GET application form endpoint.
Created `jwt_key_or_internal_multi_auth` that supports User JWT, API Key, and Internal JWT authentication.
Tweak permissions in the event the endpoint is called by an internal JWT token.
Add unit tests

## Context for reviewers

This extends the existing multi-authentication system to allow internal JWT tokens to bypass user access checks while maintaining all existing authentication methods.

## Validation steps

See updated unit tests

### DIFF
--- a/api/src/services/applications/get_application_form.py
+++ b/api/src/services/applications/get_application_form.py
@@ -12,7 +12,6 @@ from src.services.applications.application_validation import (
     is_form_required,
     validate_application_form,
 )
-from src.services.applications.auth_utils import check_user_application_access
 from src.services.applications.get_application import get_application
 
 
@@ -24,19 +23,11 @@ def get_application_form(
 
     If user is None (for internal JWT tokens), access checks are bypassed.
     """
-    if user is not None:
-        # Regular user access - get the application with user access checks
-        application = get_application(db_session, application_id, user)
-    else:
-        # Internal JWT access - get the application without user access checks
-        application_result = db_session.execute(
-            select(Application).where(Application.application_id == application_id)
-        ).scalar_one_or_none()
+    # Determine if this is an internal user request (user is None)
+    is_internal_user = user is None
 
-        if application_result is None:
-            raise_flask_error(404, f"Application with ID {application_id} not found")
-
-        application = application_result
+    # Ensure the application exists and user has access (if not internal)
+    get_application(db_session, application_id, user, is_internal_user)
 
     # Get the application form with eagerly loaded application and its attachments
     application_form = db_session.execute(
@@ -54,10 +45,6 @@ def get_application_form(
 
     if not application_form:
         raise_flask_error(404, f"Application form with ID {app_form_id} not found")
-
-    # Check if the user has access to the application (only for regular users)
-    if user is not None:
-        check_user_application_access(application, user)
 
     # Get a list of validation warnings (also sets form status)
     warnings: list[ValidationErrorDetail] = validate_application_form(application_form)
--- a/api/src/services/applications/get_application.py
+++ b/api/src/services/applications/get_application.py
@@ -16,7 +16,12 @@ from src.services.applications.application_validation import (
 from src.services.applications.auth_utils import check_user_application_access
 
 
-def get_application(db_session: db.Session, application_id: UUID, user: User) -> Application:
+def get_application(
+    db_session: db.Session,
+    application_id: UUID,
+    user: User | None = None,
+    is_internal_user: bool = False,
+) -> Application:
     """
     Get an application by ID, checking if the user has access to it.
     """
@@ -43,20 +48,24 @@ def get_application(db_session: db.Session, application_id: UUID, user: User) ->
     if not application:
         raise_flask_error(404, f"Application with ID {application_id} not found")
 
-    # Check if the user has access to the application
-    check_user_application_access(application, user)
+    # Check if the user has access to the application (skip for internal users or when user is None)
+    if not is_internal_user and user is not None:
+        check_user_application_access(application, user)
 
     return application
 
 
 def get_application_with_warnings(
-    db_session: db.Session, application_id: UUID, user: User
+    db_session: db.Session,
+    application_id: UUID,
+    user: User | None = None,
+    is_internal_user: bool = False,
 ) -> tuple[Application, list[ValidationErrorDetail]]:
     """
     Fetch an application along with validation warnings
     """
     # Fetch an application, handles the auth checks as well
-    application = get_application(db_session, application_id, user)
+    application = get_application(db_session, application_id, user, is_internal_user)
 
     # See what validation issues remain on the application's forms
     form_warnings, form_warning_map = get_application_form_errors(application)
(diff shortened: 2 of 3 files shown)


api/src/api/application_alpha/application_route.py
```diff
@@ -129,8 +131,8 @@ def application_form_update(
     "/applications/<uuid:application_id>/application_form/<uuid:app_form_id>"
 )
 @application_blueprint.output(ApplicationFormGetResponseSchema)
-@application_blueprint.doc(responses=[200, 401, 404])
-@application_blueprint.auth_required(api_jwt_auth)
+@application_blueprint.doc(responses=[200, 401, 404], security=jwt_key_or_internal_security_schemes)
+@jwt_key_or_internal_multi_auth.login_required
```
chouinar, line 135: We want this to be only JWT auth or internal key auth, we don't want our generic key auth to be valid for this.

EIther you need to be a user OR we have a weird alternative route for accessing this for internal use only that no one outside of our system will ever get. If we allow our generic keys, we'd be letting anyone who ever asked for a key be able to access other users applications

api/src/services/applications/get_application_form.py
```diff
@@ -17,13 +17,26 @@
...
+    else:
+        # Internal JWT access - get the application without user access checks
+        application_result = db_session.execute(
+            select(Application).where(Application.application_id == application_id)
+        ).scalar_one_or_none()
+
+        if application_result is None:
+            raise_flask_error(404, f"Application with ID {application_id} not found")
+
+        application = application_result
```
chouinar: Rather than have some custom things here, could we add the user check to `get_application`?

If we want to be a bit safer, could have a boolean we pass down from the route function that says "is_internal_user" and if that's True, we're fine with user being null, otherwise error.

api/src/auth/multi_auth.py
```diff
@@ -57,3 +72,12 @@ def _get_security_requirement(schemes: Sequence[str | None]) -> list[str | dict[
...
 )
+
+# List of security scheme names for application endpoints
+jwt_key_or_internal_security_schemes = _get_security_requirement(
+    [
+        api_jwt_auth.security_scheme_name,
+        api_key_auth.security_scheme_name,
+        internal_jwt_auth.security_scheme_name,
+    ]
+)
```
chouinar:
```suggestion
jwt_key_or_internal_security_schemes = _get_security_requirement(
    [
        api_jwt_auth.security_scheme_name,
        internal_jwt_auth.security_scheme_name,
    ]
)
```

api/src/auth/multi_auth.py
```diff
@@ -44,6 +50,15 @@ def get_user(self) -> MultiAuthUser:
 jwt_or_key_multi_auth = MultiHttpTokenAuth(api_jwt_auth, api_key_auth)
 
 
+# Define the multi auth that supports
+# * User JWT auth
+# * API Key Auth
+# * Internal JWT auth
+#
+# This is specifically for application endpoints that need to support internal services
+jwt_key_or_internal_multi_auth = MultiHttpTokenAuth(api_jwt_auth, api_key_auth, internal_jwt_auth)
```
chouinar:
```suggestion
jwt_key_or_internal_multi_auth = MultiHttpTokenAuth(api_jwt_auth, internal_jwt_auth)
```

api/src/api/application_alpha/application_route.py
```diff
@@ -144,9 +146,15 @@ def application_form_get(
...
-    user = token_session.user
+    # Get user from the multi-auth and determine auth type
+    multi_auth_user = jwt_key_or_internal_multi_auth.get_user()
+
+    if isinstance(multi_auth_user.user, UserTokenSession):
+        # validate user can access as we already do
+        user = multi_auth_user.user.user
+    else:
+        # Do not check
+        user = None
```
freddieyebra, line 157: Wondering if this would be used in other routes in the future? If so, it may be good to put this into its own method.

api/src/services/applications/get_application_form.py
```diff
@@ -17,13 +17,26 @@
...
+        # Regular user access - get the application with user access checks
+        application = get_application(db_session, application_id, user)
+    else:
+        # Internal JWT access - get the application without user access checks
+        application_result = db_session.execute(
+            select(Application).where(Application.application_id == application_id)
+        ).scalar_one_or_none()
+
+        if application_result is None:
+            raise_flask_error(404, f"Application with ID {application_id} not found")
```
freddieyebra: Could we just return an empty response here and handle raising the 404 at the router layer? Just an idea for instances where we would need to re use this service method outside of the get application form endpoint like future internal tools logic.

Still getting familiar with the `services/` layer but curious on your thoughts as well as @chouinar

api/src/auth/multi_auth.py
```diff
@@ -4,19 +4,22 @@
 
 from flask_httpauth import MultiAuth
 
+from ..db.models.competition_models import ShortLivedInternalToken
```
freddieyebra, line 7: Not a blocker, just noting that as a reader of this, `ShortLivedInternalToken` seems auth related but also seems related to competitions? May be something to put into our todo backlog.

api/src/api/application_alpha/application_route.py
mikehgrantsgov: This makes sense. I updated this to be either `api_jwt_auth, internal_jwt_auth`

api/src/services/applications/get_application_form.py
mikehgrantsgov: Lifted this into `get_application`

api/src/services/applications/get_application_form.py
chouinar: In a perfectly ideal approach, probably could do it like that, but this is following the convention we already do in most of our service code.

api/src/api/application_alpha/application_route.py
chouinar: This _should_ only be used on this one route, this is purely for PDF generation to work as it lets the frontend call this endpoint without needing to be logged in as the user. It's a very focused exception in our auth approach as any other auth will eventually have users associated with it.

I suppose we _could_ associate the internal token with a user if we wanted some consistency, but we're building this a bit quick, can always circle back to that later.

api/src/auth/multi_auth.py
mikehgrantsgov: We can create a follow up ticket to move it, if needed. This was added in #5306

chouinar (approved):

Issue #5436, opened 2025-06-24
Firm up email notification DB logging and adjust some logic following testing


mdragon:
### Summary

During testing we identified some changes to the logic and DB logging for the notifications emails.

### Acceptance criteria

Job tests and runs successfully.

PR #5437, opened 2025-06-24
[Issue #5436] Adjust Closing Date Email logic and DB logging

mdragon:
## Summary
Adjust the logic and dupe email prevention logic for the closing date email notifications

## Changes proposed

Utilize the Notification Log to allow for notifying on anything with close_date less than 14 days rather than only that specific 14 day window. This way if the job fails or something prevents the notifications from firing today, they'll be sent tomorrow. Checking if they've already been send is made smarter to avoid re-sending the same notification.

## Validation steps

Tests pass, not sure if additional new tests are needed.

### DIFF
--- a/api/src/task/notifications/closing_date_notification.py
+++ b/api/src/task/notifications/closing_date_notification.py
@@ -54,8 +54,8 @@ class ClosingDateNotificationTask(BaseNotificationTask):
             .where(
                 # Check if closing date is within 24 hours of two weeks from now
                 and_(
-                    OpportunitySummary.close_date >= two_weeks_from_now - timedelta(hours=24),
-                    OpportunitySummary.close_date <= two_weeks_from_now + timedelta(hours=24),
+                    OpportunitySummary.close_date <= two_weeks_from_now,
+                    OpportunitySummary.close_date >= datetime_util.utcnow()
                 ),
                 # Ensure we haven't already sent a closing reminder
                 ~exists().where(
--- a/api/src/task/notifications/closing_date_notification.py
+++ b/api/src/task/notifications/closing_date_notification.py
@@ -7,7 +7,7 @@ from sqlalchemy.orm import selectinload
 
 from src.adapters import db
 from src.db.models.opportunity_models import Opportunity, OpportunitySummary
-from src.db.models.user_models import UserOpportunityNotificationLog, UserSavedOpportunity
+from src.db.models.user_models import UserSavedOpportunity, UserNotificationLog
 from src.task.notifications.base_notification import BaseNotificationTask
 from src.task.notifications.constants import NotificationReason, UserEmailNotification
 from src.util import datetime_util
@@ -60,9 +60,12 @@ class ClosingDateNotificationTask(BaseNotificationTask):
                 # Ensure we haven't already sent a closing reminder
                 ~exists().where(
                     and_(
-                        UserOpportunityNotificationLog.user_id == UserSavedOpportunity.user_id,
-                        UserOpportunityNotificationLog.opportunity_id
+                        UserNotificationLog.user_id == UserSavedOpportunity.user_id,
+                        UserNotificationLog.opportunity_id
                         == UserSavedOpportunity.opportunity_id,
+                        UserNotificationLog.created_at >= two_weeks_from_now,
+                        UserNotificationLog.notification_reason == NotificationReason.CLOSING_DATE_REMINDER,
+                        UserNotificationLog.notification_sent == True
                     )
                 ),
             )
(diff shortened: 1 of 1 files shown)


api/src/task/notifications/closing_date_notification.py
```diff
@@ -54,15 +54,19 @@ def collect_email_notifications(self) -> list[UserEmailNotification]:
             .where(
                 # Check if closing date is within 24 hours of two weeks from now
                 and_(
-                    OpportunitySummary.close_date >= two_weeks_from_now - timedelta(hours=24),
-                    OpportunitySummary.close_date <= two_weeks_from_now + timedelta(hours=24),
+                    OpportunitySummary.close_date <= two_weeks_from_now,
+                    OpportunitySummary.close_date >= datetime_util.utcnow(),
```
chouinar: For any of the date comparisons (including the existing `two_weeks_from_now` value), I'd recommend we adjust to using date instead of datetime + use US eastern. It doesn't matter if the job is run at like 6am, but if we run this job anytime between 7pm~5am, time zone issues might present themselves. Since anything user-relevant in our system is always US Eastern (like what the closing date is compared against), best to keep that the same.


```suggestion
                    OpportunitySummary.close_date >= datetime_util.get_now_us_eastern_date(),
```

api/src/task/notifications/closing_date_notification.py
```diff
@@ -54,15 +54,19 @@ def collect_email_notifications(self) -> list[UserEmailNotification]:
...
                 # Ensure we haven't already sent a closing reminder
                 ~exists().where(
                     and_(
                         UserOpportunityNotificationLog.user_id == UserSavedOpportunity.user_id,
                         UserOpportunityNotificationLog.opportunity_id
                         == UserSavedOpportunity.opportunity_id,
+                        UserOpportunityNotificationLog.created_at <= two_weeks_from_now,
+                        # TODO Add this to the table
+                        # UserOpportunityNotificationLog.notification_reason
+                        # == NotificationReason.CLOSING_DATE_REMINDER
```
chouinar: Did we want to add that in this PR? Might be better than a time comparison as that'll conflict with other reasons we might have notified a user.

mdragon: I think we'd still need both as we may send multiple Close Date notifications if they change the Close Date or if we ever expanded to 30, 60, 90 or something like that in the future? Right now this is the only job writing to this table, but I think we'd want to have the Opp Changes job also log to this table in the future, so we could consider this change when we do that follow up work? That would likely happen after the feature launch? Though It would be good to get it in before the features live, but we have other higher priority stuff like the reset so we don't send 3 months worth of updates in the first email.

chouinar:
Could we add tests to verify this works? Surprised none of the tests required adjustments as is

mdragon:
> Could we add tests to verify this works? Surprised none of the tests required adjustments as is

I think because the existing test lands the close date exactly at 14 days, the changes to the logic don't change anything. I think I can look to add other opportunities with close dates that are in the past, less than 14 days, and greater than 14 days to make sure we send vs. don't send in the right places.

api/src/task/notifications/closing_date_notification.py
```diff
@@ -54,15 +52,20 @@ def collect_email_notifications(self) -> list[UserEmailNotification]:
             .where(
                 # Check if closing date is within 24 hours of two weeks from now
                 and_(
-                    OpportunitySummary.close_date >= two_weeks_from_now - timedelta(hours=24),
-                    OpportunitySummary.close_date <= two_weeks_from_now + timedelta(hours=24),
+                    OpportunitySummary.close_date <= text("now() + INTERVAL '14 day'"),
+                    OpportunitySummary.close_date >= text("now()"),
```
chouinar: Why switch this to DB statements? I don't know if I like mixing python and sql like that / that's not really the pattern we follow.

Having the dates calculated in the code means we can manipulate time for testing much easier - we have a test library (freezegun) that lets us change the time that the python date libraries return. If we calculate dates in the DB, then we lose that.

mdragon: Good question. I couldn't find a way to do it not in the DB below in the EXISTS, because it needs to be specific to the close date/record in question. So here I was figuring this would make it consistent across this query. But I'm willing to flip these back to as they were and just leave the one where we have no other choice.

chouinar: Ah, hadn't looked at that one too close, let me see if there is a more SQLAlchemy way of doing it

chouinar: I think you can just use timedeltas and it converts them to intervals under the hood.

Not quite as complex of a query, but doing `results = db_session.query(Opportunity).where(Opportunity.created_at + timedelta(days=3) >= Opportunity.updated_at).all()`

SQLAlchemy runs the following SQL query against the DB:
```sql
select ...
WHERE api.opportunity.created_at + %(created_at_1)s >= api.opportunity.updated_at
```
And has params `{'created_at_1': datetime.timedelta(days=3)}`.

https://docs.sqlalchemy.org/en/20/core/type_basics.html#sqlalchemy.types.Interval mentions that in Postgres timedelta translates to the interval type

mdragon: I don't think we can do it in the app code because it's date math on the opportunity close date itself. But i'm going to try to draw a timeline out on a piece of paper to see if that shows a different way to think about it.

mdragon: image
I think it has to stay in the DB because the close date is specific to each opportunitiy, not now? Now being within the 14 days isn't the important part, the time we sent the last notification of this type being in that date window is and that's also specific to the DB record, not now.

chouinar: I get the latter bit being in relation to the individual opportunities close date.

If you do `Opportunity.close_date + timedelta(...)` it converts that to the interval you have on the latter query. Your query right now is right, but is mixing SQLAlchemy and raw SQL in a way I don't like. I'm saying that you can get that without any `text("...interval")`

mdragon: Fixed

chouinar (approved):

PR #5443, opened 2025-06-25
[Issue #5387] Add rule validation to API endpoints

### LINKED ISSUE #5387: Adding rule validation to our form validation logic
### Summary

We will need to add rules to the validation that enable us to verify if all Attachment IDs have an uploaded document associated with them. This ensures that if a user deletes an attachment in the Application view, they cannot successfully submit an application without confirming that the attachment has either been replaced or is no longer needed.

### Acceptance criteria

- [ ] TBD

chouinar:
## Summary

Fixes #5387 

## Changes proposed
Connect the rule validation to our application and application form logic

## Context for reviewers
The rule JSON schema we've setup can do pre/post population, and custom validation. For the pilot, we 100% need custom validation. This ticket enables that. It deliberately does not enable pre/post-population, but does note where we'd have to adjust some things to support that, mostly configuration related.

For pre/post/validation - we need to run them at certain times (eg. post-population only on submit, OR only validation in GET endpoints) so we actually have a few different configurations. To help select the configuration, I made the calling methods pass in the ApplicationAction they're doing which we then can use to determine which config is relevant.


api/src/services/applications/application_validation.py
```diff
@@ -109,10 +152,30 @@ def get_application_form_errors(
...
 
-def validate_application_form(application_form: ApplicationForm) -> list[ValidationErrorDetail]:
+def _get_json_rule_config_for_action(action: ApplicationAction) -> JsonRuleConfig:
+    config = ACTION_RULE_CONFIG_MAP.get(action, None)
+    if config is None:
+        raise Exception(
+            f"Action {action} does not have a configured JsonRuleConfig - cannot validate."
+        )
+
+    return config
```
freddieyebra, line 162: Not a blocker but this could be simplified with:
```suggestion
    if config := ACTION_RULE_CONFIG_MAP.get(action, None):
        return config
    raise Exception(
        f"Action {action} does not have a configured JsonRuleConfig - cannot validate."
    )
```

api/src/services/applications/application_validation.py
```diff
@@ -15,6 +17,47 @@ class ApplicationAction(StrEnum):
     START = "start"
     SUBMIT = "submit"
     MODIFY = "modify"
+    GET = "get"
+
+
+START_JSON_RULE_CONFIG = JsonRuleConfig(
+    # When starting an application, we only do pre-population
+    do_pre_population=False,  # TODO - when we enable pre-population, flip this to True
```
freddieyebra, line 25: Do we have follow ups for these todos?

api/src/services/applications/application_validation.py
```diff
@@ -109,10 +152,30 @@ def get_application_form_errors(
...
+    form_validation_errors: list[ValidationErrorDetail] = []
+
+    context = JsonRuleContext(application_form, config=_get_json_rule_config_for_action(action))
+    process_rule_schema_for_context(context)
+    form_validation_errors.extend(context.validation_issues)
+
+    # TODO pre/post-populate should update the value here
+
+    form_validation_errors.extend(
+        validate_json_schema_for_form(application_form.application_response, application_form.form)
```
freddieyebra, line 178: Will there be any side effects of validating the form without pre/post population?

api/src/services/applications/application_validation.py
chouinar: There are tickets for support pre/post population for late in the quad

api/src/services/applications/application_validation.py
chouinar: Not yet - any of the fields we will do pre/post population for that happen to be required will just be something a user has to fill out themselves. Eventually pre/post populated fields would be made read-only on the frontend, but the backend can't ever make those fields truly read-only (no way to do so), so when we enable pre/post population they'll just start overwriting anything the user may have passed in.

api/src/services/applications/application_validation.py
chouinar: Generally speaking, I try to avoid `:=` except when it really really makes things cleaner because it's not a super well known feature of Python. When the calculation line you're saving is something that would be pretty annoying to work into if-statements otherwise (like I recently recommended for some [email work](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/task/notifications/opportunity_notifcation.py#L460)).

I'd err on the side of what I have as it's slightly more intuitive "no config, that's an error"

PR #5447, opened 2025-06-25
[Issue #5456] Fix transform competition task.

### LINKED ISSUE #5456: Fix competition transform bug when no associated cfda listing.
### Summary

Currently, when the transform job is run, there are failures when the case of an existing cfda listing exists in staging schema but not in the api schema. 

This fix will prevent failures and ensure that the competition is still associated with appropriate opportunity.

### Reproducibility

### How to reproduce the bug

Run the competition transform job in lower environments with the following env variables configured: 

```json
[
    { "name" : "TRANSFORM_ORACLE_DATA_ENABLE_OPPORTUNITY", "value" : "false" },
    { "name" : "TRANSFORM_ORACLE_DATA_ENABLE_ASSISTANCE_LISTING", "value" : "false" },
    { "name" : "TRANSFORM_ORACLE_DATA_ENABLE_OPPORTUNITY_SUMMARY", "value" : "false" },
    { "name" : "TRANSFORM_ORACLE_DATA_ENABLE_APPLICANT_TYPE", "value" : "false" },
    { "name" : "TRANSFORM_ORACLE_DATA_ENABLE_FUNDING_CATEGORY", "value" : "false" },
    { "name" : "TRANSFORM_ORACLE_DATA_ENABLE_FUNDING_INSTRUMENT", "value" : "false" },
    { "name" : "TRANSFORM_ORACLE_DATA_ENABLE_AGENCY", "value" : "false" },
    { "name" : "TRANSFORM_ORACLE_DATA_ENABLE_OPPORTUNITY_ATTACHMENT", "value" : "false" },
    { "name": "TRANSFORM_ORACLE_DATA_ENABLE_COMPETITION", "value": "true"}
]
```

### Browser

None

### Code of Conduct

freddieyebra:
## Summary

Fixes #5456 

## Changes proposed

This change updates the transform logic for the competition transform task to fix a bug.

## Context for reviewers

Before this change, the transformation was failing due to the competition not being able to find the opportunity based on a cfda listing entry that existed in staging db schema but not in the api schema.

This fix ensures that the correct cfda listing is associated to the correct competition and opportunity and will not be set if there is not match.

## Validation steps

This validation step is handled through unit tests. 

For a sanity check, you can revert changes in the transform class and see that the test cases will fail with 
```
sqlalchemy.exc.IntegrityError: (psycopg.errors.ForeignKeyViolation) insert or update on table "competition"
```


api/src/data_migration/transformation/subtask/transform_competition.py
```diff
@@ -41,16 +42,25 @@ def transform_records(self) -> None:
...
-                        select(Opportunity).where(Opportunity.opportunity_id == opportunity_id)
+                        select(Opportunity)
+                        .where(Opportunity.opportunity_id == opportunity_cfda.opportunity_id)
+                        .options(selectinload(Opportunity.opportunity_assistance_listings))
                     ).scalar_one_or_none()
 
-                    if not opportunity:
-                        opportunity_id = None
-                        opportunity_assistance_listing_id = None
+                    if opportunity:
```
chouinar, line 57: Could we add a comment above this to explain why we're doing this? Something like:

```
In order to know which opportunity to connect the competition to, we reference the CFDA record which links the two tables in the Oracle DB. We cannot use our assistance listing number table as we skip transforming opportunity assistance listings with null program title which only existed for the purposes of connecting the tables.
```
Feel free to adjust - just want future developers to know why we're doing this weird logic.

api/src/data_migration/transformation/subtask/transform_competition.py
```diff
@@ -41,16 +42,25 @@ def transform_records(self) -> None:
...
+                        for assistance_listing in opportunity.opportunity_assistance_listings:
+                            if (
+                                assistance_listing.opportunity_assistance_listing_id
+                                == opportunity_cfda.opp_cfda_id
+                            ):
+                                opportunity_id = opportunity.opportunity_id
+                                opportunity_assistance_listing_id = (
+                                    opportunity_assistance_listing_id
+                                )
```
chouinar: I don't think what you have right now sets the value right.
```suggestion
                                opportunity_assistance_listing_id = (
                                    assistance_listing.opportunity_assistance_listing_id
                                )
```

api/tests/src/data_migration/transformation/subtask/test_transform_competition.py
```diff
@@ -90,19 +90,30 @@ def test_process_competitions(self, db_session, transform_competition):
...
+            create_existing=False,
+            opportunity=opportunity,
+            opportunity_assistance_listing_id=cfda_opp_id_in_staging_only,
+        )
+
         # Run the transformation
         transform_competition.run_subtask()
 
         # Validate the results
+        validate_competition(db_session, cfda_opp_listing_only_in_staging)
```
chouinar: I think the validate method needs a way to validate the assistance listing value was set correctly (or not at all). Probably fine to do a boolean and have it check whether it expects the value to be none or not.

So something like:
```py
validate_competition(db_session, cfda_opp_listing_only_in_staging, expect_assistance_listing=False)
validate_competition(db_session, basic_inser, expect_assistance_listing=True)
...
```

api/src/data_migration/transformation/subtask/transform_competition.py
freddieyebra: updated

api/tests/src/data_migration/transformation/subtask/test_transform_competition.py
freddieyebra: updated

PR #5448, opened 2025-06-25
[Issue #5003] Update important dates notification

### LINKED ISSUE #5003: Implementation: Email notification content for changes to saved opportunities

babebe:
## Summary

Fixes #5003 

Update content generation for Important dates section per [DOC](https://docs.google.com/document/d/1ypejg8hL7uOXdrI7auwf9SWQ1Ag07o7rCjnrVCLfA-s/edit?tab=t.0#heading=h.sl8i9r9yex9j)
Add tests


api/src/task/notifications/opportunity_notifcation.py
```diff
@@ -290,6 +284,28 @@ def _build_opportunity_status_content(self, status_change: dict) -> str:
             + f"{BULLET_POINTS_STYLING} The status changed from {before} to {after}.<br>"
         )
 
+    def _normalize_date_field(self, value: str | int | None) -> str | int | None:
+        if isinstance(value, str):
+            return datetime.strptime(value, "%Y-%m-%d").strftime("%B %-d, %Y")
+        return value
```
chouinar: In what case would a date field be an integer / do we want to do something with that?

api/tests/src/task/notifications/test_opportunity_notification.py
```diff
@@ -458,6 +461,61 @@ def test_build_opportunity_status_content(
...
+        "imp_dates_diffs,expected_html",
+        [
+            # close_date
+            (
+                {"close_date": {"before": "2035-10-10", "after": "2035-10-30"}},
+                '<p style="padding-left: 20px;">Important dates</p><p style="padding-left: 40px;">•  The application due date changed from October 10, 2035 to October 30, 2035.<br>',
+            ),
+            (
+                {"close_date": {"before": "2025-10-10", "after": None}},
+                '<p style="padding-left: 20px;">Important dates</p><p style="padding-left: 40px;">•  The application due date changed from October 10, 2025 to None.<br>',
```
chouinar: We should check with design on what we should use to represent None/null. Both words feel a bit "technical" so maybe something like "nothing"?

api/src/task/notifications/opportunity_notifcation.py
babebe: The fiscal year is an Int

api/tests/src/task/notifications/test_opportunity_notification.py
babebe: @michellemin-nava . Any preference here ?

michellemin-nava: @chouinar, I recommended "not specified" to @babebe.

"The award maximum changed from $5,000 to not specified."

babebe: Updated to be "not specified"

api/src/task/notifications/opportunity_notifcation.py
chouinar: Ah, right. You might want to remove fiscal year (and any other forecast only fields). Here is a scenario you're going to encounter any time an opportunity has a forecast.

An opportunity is created as a forecast. It has a fiscal year. An opportunity reaches its post date, making the non-forecast be used (this'll switch out the entire summary object in the JSON). The fiscal year will go from `2025` to null. This is going to be confusing to a user as they'll wonder why the fiscal year went away.

Alternatively, for some fields like fiscal year, if they go from set to null, we just don't flag that.

Issue #5446, opened 2025-06-25
(Placeholder) Fix verbiage on mailto: verbiage


eric-serrato95:
### Summary

Currently our opportunity closing email is showing some verbiage and hyper text to users that should be behind scenes following the mailto: verbiage 

### Acceptance criteria

Placeholder

mdragon:
Confirmed mailto link is resolved in testing yesterday.

PR #5452, opened 2025-06-25
[Issue #5446] Fix mailto link in email notifications

mdragon:
## Changes proposed

Make our mailto links actual links so that they display and work as expected

## Context for reviewers

Tests pass but real proof is in an email client


api/src/task/notifications/closing_date_notification.py
```diff
@@ -16,7 +16,7 @@
 
 CONTACT_INFO = (
     "If you encounter technical issues while applying on Grants.gov, please reach out to the Contact Center:\n"
-    "mailto:support@grants.gov\n"
+    '<a href="mailto:support@grants.gov">support@grants.gov</a>\n'
```
freddieyebra, line 19: is this a UI change or functionality change? If this text is being interpreted as html, I would suggest replacing `\n` with `<br>`.

chouinar: I think the email work has some general inconsistencies in HTML vs. plain text formatting that probably should be addressed.

mdragon: Yeah we're kinda in-between right now. We're sending HTML capable emails, so we can do some light HTML. We're across the board doing `\n -> <br/>` replacement so we're ok there.

api/src/task/notifications/generate_notifications.py
```diff
@@ -37,7 +37,7 @@ class GenerateNotificationsConfig(PydanticBaseEnvConfig):
 
 
 CONTACT_INFO = (
-    "mailto:support@grants.gov\n"
+    '<a href="mailto:support@grants.gov">supoort@grants.gov</a>\n'
```
chouinar:
```suggestion
    '<a href="mailto:support@grants.gov">support@grants.gov</a>\n'
```
```suggestion
    '<a href="mailto:support@grants.gov">supoort@grants.gov</a>\n'
```

chouinar (approved):

PR #5458, opened 2025-06-26
[Issue #5426] Add is_simpler_grants_enabled flag and logic check

### LINKED ISSUE #5426: Add a flag called is_simpler_grants_enabled to the competition table, and check it in our is_open logic
### Summary

To ensure we are only showing the "Start an Application" button on the opportunity page for available Simpler Opportunities (pilot), we need to create some backend logic that the frontend team can read.

---

Add the following column to the competition table:
* `is_simpler_grants_enabled` - boolean, nullable, default False

In the `is_open` function on the competition model, adjust the logic to include:
```py
if self.is_simpler_grants_enabled is not True:
    return False
```
at the beginning before we do any of the date checks.

To avoid needing to update hundreds of competition related tests, I'd recommend in the competition factory, we just set `is_simpler_grants_enabled` to True.

### Acceptance criteria

- [ ] Add a flag called 'is_simpler_grants_enabled' to the competition table
- [ ] Check it in our 'is_open' logic

mikehgrantsgov:
## Summary

Fixes #5426

## Changes proposed

Add `is_simpler_grants_enabled` flag to competition table.
Add check `is_open` for a competition
Add migration and unit tests

## Context for reviewers

To ensure we are only showing the "Start an Application" button on the opportunity page for available Simpler Opportunities (pilot), we need to create some backend logic that the frontend team can read.

## Validation steps

See migration and tests


api/tests/src/api/competition_alpha/test_competition_route_get.py
```diff
@@ -193,6 +193,106 @@ def test_competition_get_200_is_open(
...
+    api_auth_token,
+    enable_factory_create,
+):
+    """Test that competitions with is_simpler_grants_enabled=False are not open"""
+    competition = CompetitionFactory.create(
+        opening_date=date(2025, 1, 1),
+        closing_date=date(2025, 12, 31),
+        grace_period=10,
+        is_simpler_grants_enabled=False,
+    )
```
chouinar: For this test and the next one, I'd recommend not setting the opening/closing/grace period. The factory already sets those as valid open dates by default. If we statically define these, technically when we get to 2026, this test won't start failing because the `is_simpler_grants_enabled` check happens first, but it'll look like something is off about the test.

```suggestion
def test_competition_get_200_is_simpler_grants_enabled_false(
    client,
    api_auth_token,
    enable_factory_create,
):
    """Test that competitions with is_simpler_grants_enabled=False are not open"""
    competition = CompetitionFactory.create(
        is_simpler_grants_enabled=False,
    )
```
```suggestion
def test_competition_get_200_is_simpler_grants_enabled_false(
    client,
    api_auth_token,
    enable_factory_create,
):
    """Test that competitions with is_simpler_grants_enabled=False are not open"""
    competition = CompetitionFactory.create(
        opening_date=date(2025, 1, 1),
        closing_date=date(2025, 12, 31),
        grace_period=10,
        is_simpler_grants_enabled=False,
    )
```

mikehgrantsgov: Makes sense, updated this.

chouinar (approved):

PR #5463, opened 2025-06-26
[Unticketed] Update `update_opportunity_search_queue ` explicitly to api schema

mikehgrantsgov:
## Changes proposed

Somewhere along the way the `make db-init` command broke due to schema not properly being declared in some migrations. Update this to resolve.

## Context for reviewers

Screenshot 2025-06-26 at 4 22 12 PM

## Validation steps

See passing ci/cd, local `make init-db` should succeed.


chouinar (approved):
LGTM! Surprised this hadn't caused us issues. Did we ever make it in the `api` schema? Do we need to fix the trigger non-locally to be in that schema or is it already there?

mikehgrantsgov:
> LGTM! Surprised this hadn't caused us issues. Did we ever make it in the `api` schema? Do we need to fix the trigger non-locally to be in that schema or is it already there?

This should already be there.

PR #5497, opened 2025-06-27
[Issue #5495] Update competition model col to big int.

### LINKED ISSUE #5495: Update api.Competition.expected_application_count column to BigInt.
### Summary

When running the transform competition in dev environment, the following error: 
```(psycopg.errors.NumericValueOutOfRange) integer out of range```

This is due to an integer in the staging environment being our of range for a normal integer size in postgresql.

This pr updates that column to a big int.

### Reproducibility

### How to reproduce the bug

See #5494

### Browser

None

### Code of Conduct

### PR DESCRIPTION
**Pull request by freddieyebra**
## Summary

Fixes #5495

## Changes proposed

Update Competition column

## Context for reviewers

Bug found during transform competition ecs task run.

## Validation steps

Unit tests.

### CODE REVIEW

File: api/src/db/migrations/versions/2025_06_27_increase_column_size_for_competition_.py [hunk 1]
```diff
@@ -0,0 +1,42 @@
+"""Increase column size for Competition.expected_application_count
+
+Revision ID: cff668b42a9f
+Revises: 1012e0cb7c7c
+Create Date: 2025-06-27 18:18:46.862621
+
+"""
+
+import sqlalchemy as sa
```
chouinar, line 9: Linter doesn't like this - some migrations need this import, some don't - those that don't get yelled at by the inter to remove this

PR #5498, opened 2025-06-27
[Issue #5003] Categorization notification content

babebe:

Update content generation for Categorization section per [DOC](https://docs.google.com/document/d/1ypejg8hL7uOXdrI7auwf9SWQ1Ag07o7rCjnrVCLfA-s/edit?tab=t.0#heading=h.sl8i9r9yex9j)
Add tests


api/src/task/notifications/opportunity_notifcation.py
```diff
@@ -273,6 +273,62 @@ def _get_last_notified_versions(
...
+        category_section = SECTION_STYLING.format("Categorization")
+        for field, change in category_change.items():
+            before = change["before"]
+            after = change["after"]
+            if field == "is_cost_sharing":
+                before = self._normalize_bool_field(before)
+                after = self._normalize_bool_field(after)
+            elif field in ["funding_instruments", "funding_categories"]:
+                before = ", ".join([value.capitalize() for value in before])
+                after = ", ".join([value.capitalize() for value in after])
```
chouinar: Funding categories are often multiple words with underscores, for example `science_technology_and_other_research_and_development` which with just capitalize would be `Science_technology_and_other_research_and_development`

I think what we'd want is something that does replaces the underscores with spaces as well at least. Still will be a bit wordy/not exactly match whatever the frontend has, but at least gets rid of underscores.

I'd suggest making a function that given a string replaces all underscores and capitalizes that you can just reuse.

babebe: Great! Applied func

chouinar (approved):

PR #5506, opened 2025-06-27
[unticketed] move options constants to constants directory

doug-s-nava:
## Summary
unticketed

## Changes proposed

Moves filter option definition constants file from components to constants directory

## Validation steps

1. ci passes


freddieyebra (approved):
looks good!

Issue #5377, opened 2025-06-17
Reset updates for Saved Searches


eric-serrato95:
### Summary

We need to introduce a reset for the updates we send over to our users related to their saved searches to only reflect changes that happen in the last 24 hours 

### Acceptance criteria

- [ ] The reset is implemented at least 24 hours prior to testing in prod 
- [ ] Updates for saved searches only reflect updates for the last 24 hours

PR #5512, opened 2025-06-30
 [Issue #5377] Add ability to "reset" email notifications to only pick up changes since the reset was run

mdragon:
## Changes proposed

We need a config option to enable/disable the actual sending of emails out across the wire. We still want to do the rest of the process, and act like we sent the email, so that we "clear" a backlog of changes we don't want to send emails for.

## Context for reviewers

In this case we want to reset because we haven't had the email job live in Prod for the last couple months. But changes have been happening over that time and we don't want to send a huge update email. For search results because we need to update the list of opportunities or we'll send the big update the next time, we want to run the job end-to-end EXCEPT not have an email go out.  Because we want to make it hard to send emails until we've run the reset, the default is to run in reset mode and skip sending the notification.

## Validation steps

We'll want to test the reset and email versions of the job in Dev to validate it's all working as expected.

### DIFF
--- a/api/tests/src/task/notifications/test_opportunity_notification.py
+++ b/api/tests/src/task/notifications/test_opportunity_notification.py
@@ -326,16 +326,19 @@ class TestOpportunityNotification:
         )
 
     def test_get_latest_opportunity_versions_deleted(
-        self, db_session, search_client, user, set_env_var_for_email_notification_config
+        self,
+        db_session,
+        search_client,
+        user,
+        set_env_var_for_email_notification_config,
+        notification_task,
     ):
         """Test that the user notification does not pick up opportunities that have been marked as deleted"""
         opp = factories.OpportunityFactory.create(is_posted_summary=True)
         factories.UserSavedOpportunityFactory.create(user=user, opportunity=opp, is_deleted=True)
         factories.OpportunityVersionFactory.create(opportunity=opp)
         # Instantiate the task
-        task = OpportunityNotificationTask(db_session=db_session)
-
-        results = task._get_latest_opportunity_versions()
+        results = notification_task._get_latest_opportunity_versions()
 
         # assert deleted saved opportunity is not picked up
         assert len(results) == 0
@@ -361,7 +364,7 @@ class TestOpportunityNotification:
         assert len(results) == 0
 
     def test_with_no_relevant_notifications(
-        self, db_session, user, set_env_var_for_email_notification_config, caplog
+        self, db_session, user, set_env_var_for_email_notification_config, caplog, notification_task
     ):
         """Test that only relevant (tracked) changes to opportunities generate email notifications"""
         caplog.set_level(logging.INFO)
@@ -401,10 +404,7 @@ class TestOpportunityNotification:
             opportunity=opp_2,
         )
 
-        # Instantiate the task
-        task = OpportunityNotificationTask(db_session=db_session)
-
-        results = task.collect_email_notifications()
+        results = notification_task.collect_email_notifications()
 
         # assert only the change to Opportunity 2 is tracked and should result in a notification.
         assert len(results) == 1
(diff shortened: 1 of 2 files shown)


mdragon:
I'm going to keep this PR open since every time Bruk merges another PR this branch starts failing tests as a result.

api/src/task/notifications/base_notification.py
```diff
@@ -56,42 +56,58 @@ def send_notifications(self, notifications: list[UserEmailNotification]) -> None
...
+
             try:
-                response = send_pinpoint_email_raw(
-                    to_address=user_notification.user_email,
-                    subject=user_notification.subject,
-                    message=user_notification.content,
-                    app_id=self.notification_config.app_id,
-                    trace_id=trace_id,
-                )
+                if not self.notification_config.reset_emails_without_sending:
```
chouinar: Thoughts on doing:

```
if self.notification_config.reset_emails_without_sending:
    # everything currently in the else statement
    return

# everything currently in the if-statement
```
Just as a slightly less-indented approach.

api/src/task/notifications/base_notification.py
```diff
@@ -56,42 +56,58 @@ def send_notifications(self, notifications: list[UserEmailNotification]) -> None
...
+                            "pinpoint_status_code": (
+                                email_response.status_code if email_response else None
+                            ),
+                            "pinpoint_status_message": (
+                                email_response.status_message if email_response else None
+                            ),
+                            "pinpoint_trace_id": trace_id,
+                        },
+                    )
+                    notification_log.notification_sent = True
```
chouinar: I don't think we use it anywhere yet, but in the else-statement we might want to set this too? I think we setup that notification log table as a way to send emails that failed, if we ever get around to doing that, probably don't want to randomly send them for the reset ones (but likely we'd build that out and figure it out then).

mdragon: Good catch. I had done that, but must have stashed it or something as it didn't make it in here.

api/src/task/notifications/config.py
```diff
@@ -13,3 +13,4 @@ class EmailNotificationConfig(PydanticBaseEnvConfig):
         default=True, alias="ENABLE_OPPORTUNITY_NOTIFICATIONS"
     )
     enable_search_notifications: bool = Field(default=True, alias="ENABLE_SEARCH_NOTIFICATIONS")
+    reset_emails_without_sending: bool = Field(default=True, alias="RESET_EMAILS_WITHOUT_SENDING")
```
chouinar, line 16: Are you sure you want this to default to true? Do we want the default behavior in our environments to be to do this reset logic unless you configure it to not (which without any added terraform, would be all of them)?

api/tests/src/task/notifications/test_opportunity_notification.py
```diff
@@ -184,6 +185,9 @@ def set_env_var_for_email_notification_config(self, monkeypatch):
         monkeypatch.setenv("AWS_PINPOINT_APP_ID", "test-app-id")
         monkeypatch.setenv("FRONTEND_BASE_URL", "http://testhost:3000")
 
+        self.notification_config = EmailNotificationConfig()
+        self.notification_config.reset_emails_without_sending = False
```
chouinar: looking at this, depending on how Pytest builds the fixture dependency tree, this can be null when used below.

Make this a fixture that gets pulled in (or just define it in the notification_task fixture), relying on the side effect of another fixture without anything to enforce the ordering is going to cause issues potentially if there is some randomness to the fixture order resolution

api/src/task/notifications/config.py
mdragon: Yeah, we wanted to make sure that if you ran the job without knowing about the reset path we'd do the reset as that's technically reversible/undoable while the emails going out when you meant to run the reset instead is something we can't correct after the fact. After launch, it might make sense to make this a different task name or something like that if it would be easy to base the logic off that.

chouinar (approved):

PR #5515, opened 2025-06-30
[Issue #5003] Update documents section

babebe:

Update content generation for Documents section per [DOC](https://docs.google.com/document/d/1ypejg8hL7uOXdrI7auwf9SWQ1Ag07o7rCjnrVCLfA-s/edit?tab=t.0#heading=h.sl8i9r9yex9j)
Add tests


api/src/task/notifications/opportunity_notifcation.py
```diff
@@ -332,6 +332,31 @@ def _build_categorization_fields_content(self, category_change: dict) -> str:
...
+                after_set = set(att["attachment_id"] for att in after)
+
+                if after_set - before_set:
+                    documents_section += (
+                        f"{BULLET_POINTS_STYLING} {DOCUMENTS_FIELDS["attachments"]} added.<br>"
+                    )
+                if before_set - after_set:
+                    documents_section += (
+                        f"{BULLET_POINTS_STYLING} {DOCUMENTS_FIELDS["attachments"]} removed.<br>"
+                    )
```
chouinar, line 376: What happens if an attachment is updated, but none are added or removed? Would we end up just having a header with no contents?

babebe: That would not be picked up by the diff actually since we are only looking at the attachment_id when versioning

PR #5525, opened 2025-06-30
[Issue #5003] Description notification

babebe:

Update content generation for Description section per [DOC](https://docs.google.com/document/d/1ypejg8hL7uOXdrI7auwf9SWQ1Ag07o7rCjnrVCLfA-s/edit?tab=t.0#heading=h.sl8i9r9yex9j)
Add tests

This PR introduces `beautifulsoup4` to handle HTML truncation instead of implementing a custom truncation function.
Using BeautifulSoup avoids reinventing complex parsing logic and reduces long-term maintenance risk.
It allows us to: 
Truncate based on text content while preserving valid HTML structure
Avoid edge cases around nested elements, inline tags, and mixed content

Attempt at creating custom truncation utility, required:
- Proper HTML parsing
- Handling nested tags correctly
- Ensuring valid output after truncation
- Maintaining and testing edge cases over time

The additional dependency with this change is small and widely adopted.

### DIFF
--- a/api/local.env
+++ b/api/local.env
@@ -134,6 +134,8 @@ AWS_PINPOINT_APP_ID=dummy-local-id
 ############################
 ENABLE_OPPORTUNITY_LOG_MSG=false
 
+ENABLE_WORKFLOW_API=True
+
 ############################
 # Endpoint Configuration
 ############################
--- a/api/src/api/workflows/__init__.py
+++ b/api/src/api/workflows/__init__.py
@@ -1,5 +1,8 @@
 from .workflow_blueprint import workflow_blueprint
 
+# import workflow_routes module to register the API routes on the blueprint
+import src.api.workflows.workflow_routes  # noqa: F401 isort:skip
+
 # Load all CLI commands used by the workflow CLI command
 # so that Flask is aware of them
 import src.workflow.cli.workflow_main  # noqa: F401 isort:skip
--- a/api/src/api/workflows/workflow_routes.py
+++ b/api/src/api/workflows/workflow_routes.py
@@ -1,0 +1,37 @@
+import logging
+
+import src.adapters.db as db
+from src.adapters.db import flask_db
+from src.api import response
+from src.api.workflows.workflow_blueprint import workflow_blueprint
+from src.api.workflows.workflow_schemas import (
+    WorkflowEventRequestSchema,
+    WorkflowEventResponseSchema,
+)
+from src.auth.multi_auth import jwt_or_user_api_key_multi_auth, jwt_or_user_api_key_security_schemes
+from src.logging.flask_logger import add_extra_data_to_current_request_logs
+from src.services.workflows.ingest_workflow_event import ingest_workflow_event
+
+logger = logging.getLogger(__name__)
+
+
+@workflow_blueprint.put("/events")
+@workflow_blueprint.input(WorkflowEventRequestSchema)
+@workflow_blueprint.output(WorkflowEventResponseSchema)
+@workflow_blueprint.doc(
+    summary="Send Workflow Event",
+    description="Trigger a new workflow or progress an existing one.",
+    security=jwt_or_user_api_key_security_schemes,
+    responses=[200, 401, 403, 404, 422],
+)
+@jwt_or_user_api_key_multi_auth.login_required
+@flask_db.with_db_session()
+def workflow_event_put(db_session: db.Session, json_data: dict) -> response.ApiResponse:
+
+    add_extra_data_to_current_request_logs({"event_type": json_data.get("event_type")})
+    logger.info("PUT /v1/workflows/events")
+
+    with db_session.begin():
+        event_id = ingest_workflow_event(json_data)
+
+    return response.ApiResponse(message="Event received", data={"event_id": event_id})
--- a/api/src/api/workflows/workflow_schemas.py
+++ b/api/src/api/workflows/workflow_schemas.py
@@ -1,0 +1,138 @@
+from typing import Any
+
+from marshmallow import ValidationError, validates_schema
+
+from src.api.schemas.extension import Schema, fields, validators
+from src.api.schemas.extension.schema_common import MarshmallowErrorContainer
+from src.api.schemas.response_schema import AbstractResponseSchema
+from src.constants.lookup_constants import WorkflowEntityType, WorkflowEventType, WorkflowType
+from src.validation.validation_constants import ValidationErrorType
+
+
+class WorkflowEntitySchema(Schema):
+    entity_type = fields.Enum(
+        WorkflowEntityType,
+        required=True,
+        metadata={
+            "description": "The type of entity to associate with the workflow - different workflows require different entities."
+        },
+    )
+    entity_id = fields.UUID(
+        required=True,
+        metadata={"description": "The ID of the entity, for example an opportunity_id"},
+    )
+
+
+class StartWorkflowContextSchema(Schema):
+    workflow_type = fields.Enum(
+        WorkflowType, required=True, metadata={"description": "The type of workflow to initiate"}
+    )
+    entities = fields.List(
+        fields.Nested(WorkflowEntitySchema),
+        required=True,
+        validate=[validators.Length(min=1, max=5)],
+        metadata={"description": "A list of entities (1-5) to associate with the workflow."},
+    )
+
+
+class ProcessWorkflowContextSchema(Schema):
+    workflow_id = fields.UUID(
+        required=True, metadata={"description": "The ID of the existing workflow to progress"}
+    )
+    event_to_send = fields.String(
+        required=True,
+        validate=validators.Length(min=1),
+        metadata={
+            "description": "The specific event/action to trigger in the state machine",
+            "example": "approve",
+        },
+    )
+
+
+class WorkflowEventRequestSchema(Schema):
+    event_type = fields.Enum(
+        WorkflowEventType,
+        required=True,
+        metadata={
+            "description": "The category of event: either starting a new workflow or processing an existing one"
+        },
+    )
+    start_workflow_context = fields.Nested(
+        StartWorkflowContextSchema,
+        required=False,
+        metadata={
+            "description": "Context and entities required to initialize a new workflow. Only allowed if event_type is 'start_workflow'."
+        },
+    )
+    process_workflow_context = fields.Nested(
+        ProcessWorkflowContextSchema,
+        required=False,
+        metadata={
+            "description": "Information required to progress an existing workflow state. Only allowed if event_type is 'process_workflow'."
+        },
+    )
+    metadata = fields.Dict(
+        required=False,
+        load_default={},
+        metadata={
+            "description": "A freeform JSON object for the particular event, if needed[cite: 127].",
+            "example": {"source_system": "grants_solutions", "priority": "high"},
+        },
+    )
+
+    @validates_schema
+    def validate_context(self, data: dict[str, Any], **kwargs: Any) -> None:
+        event_type = data.get("event_type")
+        start_ctx = data.get("start_workflow_context")
+        process_ctx = data.get("process_workflow_context")
+
+        if event_type == WorkflowEventType.START_WORKFLOW:
+            if not start_ctx:
+                raise ValidationError(
+                    [
+                        MarshmallowErrorContainer(
+                            ValidationErrorType.REQUIRED, "start_workflow_context is required"
+                        )
+                    ],
+                    "start_workflow_context",
+                )
+
+            if process_ctx:
+                raise ValidationError(
+                    [
+                        MarshmallowErrorContainer(
+                            ValidationErrorType.INVALID,
+                            "process_workflow_context should not be provided",
+                        )
+                    ],
+                    "process_workflow_context",
+                )
+
+        if event_type == WorkflowEventType.PROCESS_WORKFLOW:
+            if not process_ctx:
+                raise ValidationError(
+                    [
+                        MarshmallowErrorContainer(
+                            ValidationErrorType.REQUIRED, "process_workflow_context is required"
+                        )
+                    ],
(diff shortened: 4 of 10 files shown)


api/src/task/notifications/opportunity_notifcation.py
```diff
@@ -274,6 +274,16 @@ def _get_last_notified_versions(
 
         return {(row.user_id, row.opportunity_id): row[2] for row in results}
 
+    def _build_description_fields_content(self, description_change: dict, opp_id: int) -> str:
+        after = description_change["after"]
+        if after:
+            description_section = SECTION_STYLING.format("Description")
+            if len(after) > TRUNCATION_THRESHOLD:
+                after += f"<a href='{self.notification_config.frontend_base_url}/opportunity/{opp_id}' style='color:blue;'>...Read full description</a>"
+            description_section += f"{BULLET_POINTS_STYLING} <i>New Description:</i> {after}<br>"
```
chouinar: A technical problem that I don't know an answer to (maybe ask frontend in our engineer channel) - how do we truncate HTML?

The description field has HTML in it most of the time, if we truncate it - we'll be putting half-written HTML into our HTML, potentially partially through a tag. For example, on this random opportunity: https://simpler.grants.gov/opportunity/359874 it uses bold. If we happened to truncate in the middle of that text, suddenly the rest of our email is all bold.

Maybe we want to escape the HTML or remove the tags? I don't know - but we'll definitely run into some problems if we don't do anything.

babebe: @doug-s-nava . Any suggestions ?

doug-s-nava: so... we had to deal with this on the frontend as well. I built a function to handle truncating and being smart about closing tags. Reimplementing this in python would likely be gnarly, but maybe that's our best option?

https://github.com/HHS/simpler-grants-gov/blob/53d4ec408d501b984e33aa618a326ef4fc963d6b/frontend/src/utils/generalUtils.ts#L12

mdragon: For version one and initial launch, can we just not truncate the content? Or not show the content changes just note that "The description was changed." On our current timeline we don't really have time to implement HTML striping or closing tags automatically before we want to be sending emails?

@eric-serrato95 Ok just saying "The Description was changed" without showing what changed like we do for other fields for now?

eric-serrato95: @mdragon I think that's ok for launch, we can figure out how to handle large description changes for example after launch

api/src/task/notifications/opportunity_notifcation.py
```diff
@@ -274,6 +289,71 @@ def _get_last_notified_versions(
 
         return {(row.user_id, row.opportunity_id): row[2] for row in results}
 
+    def _contains_regex(self, value: str, regex: str) -> bool:
+        return bool(re.search(regex, value))
+
+    def _truncate_html_safe(self, html_str: str) -> str:
```
chouinar: Could you move these to the string_utils file? They generic enough for reuse.

api/tests/src/task/notifications/test_opportunity_notification.py
```diff
@@ -790,6 +801,94 @@ def test_build_eligibility_content(
...
         assert res == expected_html
 
+    @pytest.mark.parametrize(
+        "html_str,expected_html",
+        [
+            # Truncate mid-text inside inline tag <strong>, no closing tags
+            (
+                "<p>This is a <strong>very big description here!!!!",
+                "<p>This is a <strong>very big description here!!!!</strong>",
+            ),
```
chouinar: Could you add a test for multiple of the same tag nested? Something like `<div>first<div>second layer<div>yet another layer</div></div></div>`

api/src/task/notifications/opportunity_notifcation.py
```diff
@@ -274,6 +289,71 @@ def _get_last_notified_versions(
...
+                else:
+                    # Add non self-closing tags
+                    tag_name = re.match(r"<(\w+)", token)
+                    if tag_name and not token.endswith("/>"):
+                        tag_stack.append(tag_name.group(1))
+
+        # Remove block-level closing tags from the end
+        while output_tokens:
+            last = output_tokens[-1]
+            closing_match = re.match(r"</(\w+)>", last)
```
chouinar: For any of these regexes, especially one we pass to compile, could we define them statically with a note of what they're looking for? You don't necessarily need to go piece-by-piece like https://github.com/HHS/simpler-grants-gov/blob/main/api/src/logging/pii.py#L60 but at least something like "words ending in a slash" or whatever.

api/tests/src/task/notifications/test_opportunity_notification.py
babebe: Added

api/src/task/notifications/opportunity_notifcation.py
```diff
@@ -276,11 +278,28 @@ def _get_last_notified_versions(
...
+            SECTION_STYLING.format("Description"),
+            f"{BULLET_POINTS_STYLING} <i>New Description:</i>",
+        ]
+
+        if len(after) > TRUNCATION_THRESHOLD:
+            truncated = after[:TRUNCATION_THRESHOLD]
+
+            if contains_regex(truncated, r"<[^>]+>"):
+                truncated = truncate_html_safe(truncated)
+
```
chouinar: Might be a naming thing, but should `truncate_html_safe` take in the truncation length?

I'm imagining I'd want to use it like so:
```py
my_html = "..."

truncated_html = truncate_html_safe(my_html, TRUNCATION_THRESHOLD)
```
And move the bit you have here to that utility function.

api/tests/src/task/notifications/test_opportunity_notification.py
```diff
@@ -787,13 +794,85 @@ def test_build_eligibility_content(
...
+        self,
+        db_session,
+        html_str,
+        expected_html,
+        set_env_var_for_email_notification_config,
+        set_truncation_threshold,
+    ):
+
+        res = truncate_html_safe(html_str)
+        assert res == expected_html
```
chouinar: This test should be in test_string_utils

kmk142789 (approved):

api/src/util/string_utils.py
```diff
@@ -1,5 +1,20 @@
...
+    "div",
+    "section",
+    "article",
+    "header",
+    "footer",
+    "main",
+    "nav",
+    "aside",
+    "blockquote",
+    "ul",
```
joshtonava: **comment:** one additional tag I can think of is `li`, there may be more. I want to confirm this list is exhaustive.

api/tests/src/task/notifications/test_opportunity_notification.py
joshtonava: issue: Test is failing I believe here 
```
TypeError: OpportunityNotificationTask._build_description_fields_content() missing 1 required positional argument: 'opp_id'
```

api/src/util/string_utils.py
joshtonava: I think missing block tags are

li, ol, h1, h2, h3, h4, h5, h6, pre, hr, table, thead, tbody, tr, td, th, form, fieldset

api/src/util/string_utils.py
```diff
@@ -20,3 +35,88 @@ def is_valid_uuid(value: str) -> bool:
...
+        i -= 1
+
+    return html
+
+
+def _truncate_preserving_html(html_str: str) -> str:
+    tag_stack: list = []  # Keep track of open tags
+    output_tokens: list = []  # list of tokens
+
+    # Match any tags or text
```
joshtonava:
```suggestion
    # Match any HTML tag or text content between tags
    # Pattern breakdown:
    # - <[^>]+>  : Matches complete HTML tags (opening, closing, or self-closing)
    # - |        : OR operator
    # - [^<]+    : Matches any text content (non-tag characters)
```

api/src/util/string_utils.py
```diff
@@ -20,3 +35,88 @@ def is_valid_uuid(value: str) -> bool:
...
+            return html
+        if html[i] == "<":
+            # Found start of a tag with no closing '>' → strip it
+            return html[:i]
+        i -= 1
+
+    return html
+
+
+def _truncate_preserving_html(html_str: str) -> str:
```
joshtonava: **suggestion:**
Im not sure how crazy we need to go with testing but some edge cases I can think of:

- Self-closing tags with attributes (e.g., ``)
- Comments (`<!-- ... -->`)
- CDATA sections
- Malformed/mismatched tags
- Tags with complex attributes (quotes, special characters)

api/src/util/string_utils.py
```diff
@@ -20,3 +35,88 @@ def is_valid_uuid(value: str) -> bool:
...
+        token = match.group(0)  # substring matched by regex
+        output_tokens.append(token)
+
+        if token.startswith("<"):  # Open Tag
+            tag_name = re.match(
+                r"</"  # Matches the begining of closing tag of an element
+                r"(\w+)",  # Matches one or more word characters
+                token,
+            )
+            if tag_name and token[1] == "/":  # Closing tag
```
joshtonava: **question:** The closing tag matching logic checks `token[1] == "/"` after already matching with a regex that starts with r"</". This seems redundant. Is the double check intentional for safety, or can it be simplified?

api/src/util/string_utils.py
```diff
@@ -20,3 +35,88 @@ def is_valid_uuid(value: str) -> bool:
         return True
     except ValueError:
         return False
+
+
+def contains_regex(value: str, regex: str) -> bool:
```
joshtonava: **nit:** This is only used once, consider prefixing with an underscore for internal use or even inlining.

api/src/util/string_utils.py
```diff
@@ -20,3 +35,88 @@ def is_valid_uuid(value: str) -> bool:
...
+        r"[^<]+"  # Match one or more characters (excluding <)
+    )
+
+    for match in tag_regex.finditer(html_str):
+        token = match.group(0)  # substring matched by regex
+        output_tokens.append(token)
+
+        if token.startswith("<"):  # Open Tag
+            tag_name = re.match(
+                r"</"  # Matches the begining of closing tag of an element
```
joshtonava:
```suggestion
                r"</"  # Matches the beginning of closing tag of an element
```

api/src/util/string_utils.py
```diff
@@ -20,3 +35,88 @@ def is_valid_uuid(value: str) -> bool:
...
+        if closing_match and closing_match.group(1) in BLOCK_TAGS:
+            output_tokens.pop()
+        else:
+            break
+    # Close any remaining open tags (excluding block-level)
+    closing_tags = [f"</{tag}>" for tag in reversed(tag_stack) if tag not in BLOCK_TAGS]
+    return "".join(output_tokens) + "".join(closing_tags)
+
+
+def truncate_html_safe(value: str, max_length: int) -> str:
```
joshtonava: **suggestion:** Consider adding a doc string here.

api/src/util/string_utils.py
```diff
@@ -20,3 +35,88 @@ def is_valid_uuid(value: str) -> bool:
...
+                # Add non self-closing tags
+                tag_name = re.match(
+                    r"<"  # Matches the opening angle bracket of an HTML tag,
+                    r"(\w+)",  # Matches one or more word characters
+                    token,
+                )
+                if tag_name and not token.endswith("/>"):
+                    tag_stack.append(tag_name.group(1))
+
+    # Remove block-level closing tags from the end
```
joshtonava: suggestion: A comment explaining why block tags are left open would help future maintainers.

api/src/util/string_utils.py
chouinar: I think this logic is getting quite complex / there's a lot we'd have to account for. Have we looked into any libraries that might handle this for us?

Something like https://stackoverflow.com/a/9205760 ?

api/src/util/string_utils.py
babebe: Now, using bs4 to handle truncation

api/tests/src/task/notifications/test_opportunity_notification.py
babebe: udpated

api/src/util/string_utils.py
babebe: Added new func

api/src/util/string_utils.py
babebe: Now, using bs4 to handle truncation

api/src/util/string_utils.py
babebe: Now, using bs4 to handle truncation

api/src/util/string_utils.py
babebe: Removed

api/src/util/string_utils.py
babebe: added doc string to new func

babebe: Rewrote logic to now, use bs4 to handle truncation.

api/src/util/string_utils.py
```diff
@@ -20,3 +23,83 @@ def is_valid_uuid(value: str) -> bool:
         return True
     except ValueError:
         return False
+
+
+def contains_regex(value: str, regex: str) -> bool:
+    return bool(re.search(regex, value))
+
+
+def _strip_partial_tag(html: str) -> str:
```
joshtonava: **suggestion:** If we aren't using this we should remove it

api/src/util/string_utils.py
joshtonava: is this removed?

api/src/util/string_utils.py
babebe: Removed

babebe:
> I think this looks good. Can you maybe put something in the PR description for why beautiful soup was chosen?

Updated Description

Issue #5524, opened 2025-06-30
Add (BETA) Banner to first batch of production email notifications


eric-serrato95:
### Summary

for our first test run in prod for email notifications, we should add a beta banner so that our users would know we are currently developing the email notifications

### Acceptance criteria

- [ ] all 3 emails should have the BETA as the starting verbiage in the subject line

eric-serrato95:
@mdragon updated from BETA to [Test Email] on subject line!

PR #5553, opened 2025-07-01
[Issue #5524] Add flag on email subjects to convey to users we're testing new emails

mdragon:
## Changes proposed

Per Product/Design add prefix of "[Test email]" to each of the outgoing emails until we're more stable and solid on the email contents and they are sending regularly.

## Validation steps

Will test as part of larger email testing in Dev/Stage after deploy.

### DIFF
--- a/api/src/task/notifications/closing_date_notification.py
+++ b/api/src/task/notifications/closing_date_notification.py
@@ -121,9 +121,9 @@ class ClosingDateNotificationTask(BaseNotificationTask):
                     extra={"user_id": user_id, "closing_opp_count": len(closing_opportunities)},
                 )
                 subject = (
-                    "[Test email] Applications for your bookmarked funding opportunity are due soon"
+                    "[This is a test email from the Simpler.Grants.gov alert system. No action is required] Applications for your bookmarked funding opportunity are due soon"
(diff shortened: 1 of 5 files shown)


api/src/task/notifications/closing_date_notification.py
```diff
@@ -121,9 +121,9 @@ def collect_email_notifications(self) -> list[UserEmailNotification]:
                     extra={"user_id": user_id, "closing_opp_count": len(closing_opportunities)},
                 )
                 subject = (
-                    "Applications for your bookmarked funding opportunity are due soon"
+                    "[Test email] Applications for your bookmarked funding opportunity are due soon"
```
chouinar: Not a blocker, but do we want to consider an environment variable for determining this behavior?

chouinar (approved):
LGTM - just throwing an idea out

api/src/task/notifications/closing_date_notification.py
mdragon: They're not really test emails, it's more of a for the first week flag explain why the format or contents might be a little wonky. So I think we probably wouldn't put this back ever after we clear it out in a week or two.

chouinar (approved):

PR #5554, opened 2025-07-01
[Unticketed] Clean up docs for better steps to get someone from 0 to 1 on local FE

mdragon:
## Summary
Update the API make commands referenced in the FE docs to ones that are more wholistic. Addresses some snags Chris Freeman hit while trying to poke around in our E2E testing process.


myduong-navapbc (approved):
LGTM

PR #5560, opened 2025-07-01
[Unticketed] Update docs with better API test steps

mdragon:
## Summary

Update the docs since I always forget and have to re-figure out how to run the API tests (and limit which it tries to run) every time I return to the API code.

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -39,6 +39,15 @@ root
 
 See [development.md](../documentation/api/development.md) for installation and development instructions.
 
+## Running tests locally
+1. Start the services with `docker compose up`
+2. In another terminal run the tests `make test` or if you've set your PY Approach to local you probably want to run the tests in Docker so you don't have to deal with Env Vars and other config `PY_RUN_APPOACH=docker make test`
+
+You can also run only certain tests by pattern matching the file name:
+```bash
+args=tests/src/task/notifications/* make test
+```
+
 ## Running jobs/tasks
 
 ### Locally


api/README.md
```diff
@@ -39,6 +39,15 @@ root
 
 See [development.md](../documentation/api/development.md) for installation and development instructions.
 
+## Running tests locally
+1. Start the services with `docker compose up`
+2. In another terminal run the tests `make test` or if you've set your PY Approach to local you probably want to run the tests in Docker so you don't have to deal with Env Vars and other config `PY_RUN_APPOACH=docker make test`
+
+You can also run only certain tests by pattern matching the file name:
+```bash
+args=tests/src/task/notifications/* make test
```
chouinar: Thoughts on having this be flipped a bit to have args at the end? That way the bit you'll likely edit is on the right-side.

For example, my command with some other stuff thrown in:

```sh
make test args="-x -s -vv tests/src/api/users/test_user_route_login.py"
```

Although now that I have PyCharm working, I rarely run individually from the CLI, but that's a whole fun separate setup to get env vars and packages happy.

api/README.md
```diff
@@ -39,6 +39,15 @@ root
 
 See [development.md](../documentation/api/development.md) for installation and development instructions.
 
+## Running tests locally
+1. Start the services with `docker compose up`
```
mdragon: @chouinar I think this isn't a needed step, actually?

chouinar: Only need to do `make init`. Running the service is only necessary to query it locally, the tests don't actually hit the local API (but do hit the DB, and search index that `make init` will setup).

api/README.md
```diff
@@ -39,6 +39,19 @@ root
...
+2. In another terminal run the tests `make test` or if you've set your PY Approach to local you probably want to run the tests in Docker so you don't have to deal with Env Vars and other config `PY_RUN_APPOACH=docker make test`
```
chouinar: Can exclude the docker compose up bit and the running in another terminal. Just `make init` and then you can run `make test`.

PR #5578, opened 2025-07-02
convert agency search results to filter options

doug-s-nava:
## Summary

unticketed

## Changes proposed

Properly converts agency search result data into filter option data

## Context for reviewers

Without this, searching for agencies results in unlabeled agency filter options

## Validation steps

1. start a local server on this branch
3. perform an agency search in the agency filter for a term that will give you results
4. _VERIFY_: results display with correct labels
5. selected displayed option
6. _VERIFY_: search updates as expected


mdragon (approved):
Looks good and tested successfully

Issue #5608, opened 2025-07-03
Reduce noisiness of health checks in logs


mdragon:
### Summary

We get a lot of health checks logged, as we have multiple containers being checked by Load Balancers across different AZs, etc. For now, only log 1 out of every 10 healthchecks.

### Acceptance criteria

Healthcheck volume in logs is reduced. Currently it's 16/minute in Dev and Staging and 40/minute in Prod just for the FE

mdragon:
<img width="1477" alt="Image" src="https://github.com/user-attachments/assets/26dd0268-f3c1-4be6-802f-b5441973acde" />
Accomplished desired 10x reduction in FE health check logging, confirmed in Dev.

PR #5609, opened 2025-07-03
[Issue #5608] Sample logs for /health

mdragon:
## Summary

Work for #5608

## Changes proposed

If the middleware logger detects it's a request to /health, pull a random number to only log 1/10 actual hits to /health.

If my math is right, we're sending about 622MB of health check logs a month to NewRelic/CloudWatch.

## Validation steps
1. Checkout the branch
2. Run `npm run dev | grep "/health"`
3. Refresh http://localhost:3000/health multiple times
    a. Note that not every request logs the JSON info style log line
4. Validate that other URLs still log as expected 

image


freddieyebra (approved):
looks good. Verified the cadence of the expected rate of logging for the /health endpoint.

Screenshot 2025-07-07 at 9 53 41 AM

Issue #5335, opened 2025-06-12
Adjust the PUT application form endpoint to take in whether to include the value


chris-kuryak:
### Summary

Conditional forms may be needed by some applicants. Applicants may start on a conditional form that they don't need to submit. So we need to provide a way for the applicant to indicate which conditional forms they would like to attach to their application.

This work follows #5334 and is to adjust the PUT application form endpoint to take in whether to include the value.

### Acceptance criteria
- [ ] Sync with FE and align on whether we should create a new endpoint or just pass more information on an existing endpoint (the one used for updating user's answers in a form)
- [ ] Adjust the PUT application form endpoint to take in whether to include the value

PR #5611, opened 2025-07-07
[Issue #5335] Adjust the PUT application form endpoint to take in whether to include the value

mikehgrantsgov:
## Summary

Work for #5335

## Changes proposed

Add a new endpoint: `/applications/<uuid:application_id>/forms/<uuid:form_id>/inclusion`, which  provides a way for the applicant to indicate which conditional forms they would like to attach to their application.
Add new request/response schemas
Modify existing `update_application_form.py` to cover both new parameter and existing functionality.
Add new unit tests to cover functionality

## Context for reviewers

Due to the changes required by the frontend to modify the existing endpoint, we decided to add a new endpoint instead of modify an existing one. 

## Validation steps

See new unit tests.


api/src/api/application_alpha/application_schemas.py
```diff
@@ -224,3 +247,24 @@ class ApplicationAttachmentUpdateRequestSchema(Schema):
...
+class ApplicationFormInclusionUpdateResponseDataSchema(Schema):
+    application_id = fields.UUID()
+    form_id = fields.UUID()
+    is_included_in_submission = fields.Boolean(
+        metadata={"description": "Whether this form is included in the application submission"},
+    )
+
+
+class ApplicationFormInclusionUpdateResponseSchema(AbstractResponseSchema):
+    data = fields.Nested(ApplicationFormInclusionUpdateResponseDataSchema())
```
chouinar: It might not be something we want to do in this PR, but updating the application form (both the normal one, and your new one) should probably just use the same response schema as GET application form (`ApplicationFormGetResponseDataSchema`). If switching to doing that is simple, I'd say we should just make that change.

We'd also want this new field in the GET endpoint response anyways so the frontend can render the value.

api/src/api/application_alpha/application_route.py
```diff
@@ -121,12 +124,56 @@ def application_form_update(
...
+            is_included_in_submission=is_included_in_submission,
+        )
+
+    return response.ApiResponse(
+        message="Success",
+        data={
+            "application_id": application_form.application_id,
+            "form_id": application_form.form_id,
+            "is_included_in_submission": application_form.is_included_in_submission,
+        },
```
chouinar: Marshmallow can fetch the values out of the DB models, as long as the shape of our DB models matches the Marshmallow response model, we shouldn't be building a custom response.
```suggestion
        data=application_form
```

api/src/services/applications/update_application_form.py
```diff
@@ -72,23 +82,42 @@ def update_application_form(
...
-        application_form, ApplicationAction.MODIFY
-    )
+    # Only validate if we're updating form content
+    if application_response is not None:
+        warnings: list[ValidationErrorDetail] = validate_application_form(
+            application_form, ApplicationAction.MODIFY
+        )
+    else:
+        warnings = []
+
```
chouinar: We should always do the validation, even if they're not changing their answers.

Because of how the validation will work, if a form is marked as "included" that changes the validation as we suddenly want to enforce it, so even if their answers didn't change, the way we validate does.

api/src/services/applications/update_application_form.py
```diff
@@ -72,23 +82,42 @@ def update_application_form(
 
     if application_form:
         # Update existing application form
-        application_form.application_response = application_response
+        if application_response is not None:
+            application_form.application_response = application_response
+        if is_included_in_submission is not None:
+            application_form.is_included_in_submission = is_included_in_submission
```
chouinar, line 88: An option, we can come back to this later.

For handling updates, what if instead we passed in the request dict (instead of pulling the values out), and did something roughly like:

```py
request_dict = {"application_response": {...}, "is_included_in_submission": True}

for field, value in request_dict.items():
    # TODO - we'd likely want some vague sort of validation that the fields
    # are ones we'd actually want, maybe a check if the field isn't an expected one
    setattr(application_form, field, value)
```

api/src/api/application_alpha/application_schemas.py
mikehgrantsgov: I made some changes to consolidate this. Makes sense!

api/src/api/application_alpha/application_route.py
mikehgrantsgov: Updated this

api/src/services/applications/update_application_form.py
mikehgrantsgov: Also updated this

api/src/services/applications/update_application_form.py
mikehgrantsgov: Got it, I didn't make this change but can create a follow up ticket.

chouinar (dismissed):

chouinar (approved):

PR #5619, opened 2025-07-07
[Issue #4393] Temporary disable auto deploy to dev

joshtonava:
## Summary

Work for #4393

## Changes proposed

Commented out the push trigger in the cd-api and cd-frontend workflows

## Context for reviewers

Testing UUID migration in dev first. We want control over the environment and when the deploy happens so turning off auto deploy.


mdragon:
Do we want to do analytics too so it's consistent and nothing deploys or do we want to limit it to just stuff we actively need to hold back?

joshtonava:
> Do we want to do analytics too so it's consistent and nothing deploys or do we want to limit it to just stuff we actively need to hold back?

Going to limit it to the stuff we actively need to hold back for now

Issue #5277, opened 2025-06-09
Add unit tests for QueryProvider


doug-s-nava:
### Summary

The QueryProvider does not have any unit tests

### Acceptance criteria

- [ ] unit tests are created to test all of QueryProvider's functionality

PR #5625, opened 2025-07-08
Issue #5277: Add query provider unit tests

elliottbattle-nava:
## Summary

Work for #5277

## Changes proposed

- Created a new file (`frontend/tests/services/search/QueryProvider.test.tsx`)
- Added 4 new unit tests  

## Context for reviewers
The `frontend/src/services/search/QueryProvider.tsx` file is currently only at 19.04% code coverage, with lines 10-41 all untested
image

## Validation steps

All new unit tests are passing and test runner confirms that `frontend/src/services/search/QueryProvider.tsx` is now at 100% code coverage
image


frontend/tests/services/search/QueryProvider.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+};
+
+// Generic child component that creates a context object, updates it and displays dynamic content based-on its current state
+function ChildWithHandlers(props: contextHandlers) {
+  const context = useContext(QueryContext);
+  props.onContextUpdate(context);
+  return props.onContextDisplay(context);
+}
+
+function renderQueryProviderWithHandlers(props: contextHandlers) {
```
doug-s-nava: since this is a function that takes props and returns React nodes, can we treat it like a component?

frontend/tests/services/search/QueryProvider.test.tsx
```diff
@@ -0,0 +1,73 @@
+import QueryProvider, { QueryContext } from "src/services/search/QueryProvider";
+import { QueryContextParams } from "src/types/search/searchQueryTypes";
+import { render, screen } from "tests/react-utils";
+
+import { useContext } from "react";
+
+type contextHandlers = {
```
doug-s-nava: [nit] we've been using capitalized camelCase for types and interfaces - can you update to
```suggestion
type ContextHandlers = {
```

doug-s-nava (commented):
testing context / providers is tricky, and I like this approach

can you add a test for the default query term behavior?

frontend/tests/services/search/QueryProvider.test.tsx
elliottbattle-nava: So currently this function returns void and is intended to reduce repeated boilerplate in our tests. I suppose I could refactor it to return the React node, treat it like a component and then manually call `render()` in each test, but that sort of seems antithetical to what I was going for here.

elliottbattle-nava: Wound-up completely refactoring how the util method worked to resolve another issue, so gonna resolve this comment chain since it's no longer relevant to what's checked-in.

elliottbattle-nava:
> testing context / providers is tricky, and I like this approach
> 
> can you add a test for the default query term behavior?

~~Tried implementing this but ran into a blocker with the `useSearchParams()` not mocking properly. Moving this section of the convo into Slack.~~

Update 7/12: Looks like it was an issue with me trying to mock the hook inside the body of the test itself. Just pushed a change with the test now passing.

doug-s-nava (approved):
✅ ✅ ✅ ✅

PR #5630, opened 2025-07-08
[Issue #5003] Bug no prior version opportunity update

babebe:
## Summary
if no prior version exists prior to a user having saved an opportunity log the error and skip the opportunity.


api/src/task/notifications/opportunity_notifcation.py
```diff
@@ -144,12 +144,19 @@ def collect_email_notifications(self) -> list[UserEmailNotification]:
...
                 opp.previous = prior_notified_versions.get(
                     (user_changed_opp.user_id, opp.opportunity_id)
                 )
+                opp.previous = prior_notified_versions.get((user_id, opp.opportunity_id))
+                if opp.previous is None:
+                    logger.error(
+                        "No previous version found for this opportunity",
+                        extra={"user_id": user_id, "opportunity_id": opp.opportunity_id},
+                    )
+                    continue
```
chouinar: Is this an error / do we want to alert an engineer that there is a problem, or is this a scenario we could reasonably expect to happen?

api/src/task/notifications/opportunity_notifcation.py
```diff
@@ -144,12 +144,19 @@ def collect_email_notifications(self) -> list[UserEmailNotification]:
...
 
-            updated_opps = user_changed_opp.opportunities
+            updated_opps: list[OpportunityVersionChange] = []
 
-            for opp in updated_opps:
+            for opp in user_changed_opp.opportunities:
                 opp.previous = prior_notified_versions.get(
                     (user_changed_opp.user_id, opp.opportunity_id)
                 )
+                opp.previous = prior_notified_versions.get((user_id, opp.opportunity_id))
```
chouinar: This line is duplicated.

api/src/task/notifications/opportunity_notifcation.py
babebe: In prod this should never be the case as we already have opportunity versions created for users to have saved the opportunity. If none exists, its probably because the version job failed

api/tests/src/task/notifications/test_opportunity_notification.py
```diff
@@ -360,6 +359,31 @@ def test_with_no_prior_version_email_collections(
...
+
+        assert len(results) == 0
+        # Verify the log contains the correct metrics
+        log_records = [
+            r
+            for r in caplog.records
+            if "No previous version found for this opportunity" in r.message
+        ]
+
+        assert len(log_records) == 1
```
chouinar, line 385: I'm afraid I'm not following this change / whether it's an error case. What you setup here seems pretty uneventful/ordinary.

A user has a saved opportunity and it has exactly one saved version - what about that should error? That'll be _all_ opportunities when they first get created.

babebe: the version created is after the user has saved the opportunity hence it would be picked up as a new opportunity to notify on but to do the diff we collect the last opportunity the user was last notified about. In this scenario that version does not exist, since we will not be able to do a diff we omit this opportunity.

chouinar: So, if I'm understanding, this would hit the error scenario:
* A user saved an opportunity on Jan 10th
* The only opportunity version we have is on Jan 11th or later

If so - that'll happen for any saved opportunities from before July 9th since we're remaking the version table. I know that @mdragon was also trying to setup a "reset" approach, make sure that handles this (or we just quickly update the saved opportunity table to have the last notified be July 10th for everyone).

mdragon: We were only going to reset for folks we would have otherwise emailed (primarily for the search ones were we need to calculated the new list of IDs, not just timestamp). So we could run that reset and then manually run a query to update all the last notified dates without any other logic being applied.

Issue #5255, opened 2025-06-05
[Placeholder] Adjusting logic for organizations if sam.gov entity is expired


chouinar:
### Summary

TODO: 
* Do we want to not create the org/add the EBIZ POC?
* Do we want to skip returning expired orgs, or just return a flag saying whether it’s expired?
* Do we want to block someone from starting and/or submitting an application if the org is expired?
* Anything else?

### Acceptance criteria

TODO

PR #5632, opened 2025-07-08
[Issue #5255] In Start Application raise an error if organization is expired

mikehgrantsgov:
## Summary

Work for #5255

## Changes proposed

Add `_validate_organization_expiration` and integrate into service `create_application`
Add unit tests

## Context for reviewers

If a user is starting an application as an organization (organization ID is not null), verify it is not expired.

## Validation steps

See unit tests


api/src/services/applications/create_application.py
```diff
@@ -99,7 +138,9 @@ def create_application(
 
     # Validate organization if provided
     if organization_id is not None:
-        _validate_organization_membership(db_session, organization_id, user)
+        organization = _validate_organization_membership(db_session, organization_id, user)
+        # Validate that the organization is not expired
+        _validate_organization_expiration(organization)
```
chouinar: Could we do:

```py
# just copying the query already being done to here
organization = db_session.execute(...)
if not organization:
    raise_flask_error(404, "Organization not found")

_validate_organization_membership(organization, user)
_validate_organization_expiration(organization)
```
That way all the validate methods have the same pattern of expecting the org rather than one having to also fetch it.

mikehgrantsgov: Makes sense, updated this.

chouinar:
Hmm, your test failures are unrelated, probably the UUID work causing some randomness where it wasn't before. Let me try to fix those to get things unblocked

chouinar (approved):

Issue #5634, opened 2025-07-08
Redirect Opportunities from numeric to UUIDs


mdragon:
### Summary

Redirect integer IDs we were using previously to UUIDs we're using now

### Acceptance criteria

- Page loads with UUIDs
- Page redirects to correct UUID when passed ID

PR #5635, opened 2025-07-08
[Issue #5634] Redirect prior int Opp ID to new UUID

mdragon:
## Summary

Work for #5634

## Changes proposed

If an Opportunity page is loaded using an Int ID, redirect to the same opportunity URL but with the new UUID

## Validation steps

1. Checkout the branch
2. Go to an int ID: https://localhost:3000/opportunity/3
3. Verify you're redirected to a UUID URL
4. Do a search
5. Click a search result
6. Verify the page loads without any redirect

### DIFF
--- a/frontend/src/app/[locale]/opportunity/[id]/page.tsx
+++ b/frontend/src/app/[locale]/opportunity/[id]/page.tsx
@@ -8,7 +8,7 @@ import { OpportunityDetail } from "src/types/opportunity/opportunityResponseType
 import { WithFeatureFlagProps } from "src/types/uiTypes";
 
 import { getTranslations } from "next-intl/server";
-import { notFound, redirect } from "next/navigation";
+import { notFound, redirect, RedirectType } from "next/navigation";
 
 import BetaAlert from "src/components/BetaAlert";
 import Breadcrumbs from "src/components/Breadcrumbs";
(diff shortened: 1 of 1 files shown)


doug-s-nava:
why?

mdragon:
> why?

We don't want to split our Google Juice as they get mad when you have two URLs with the same content. Over all we'd also like to not have to support the Int IDs forever. @doug-s-nava  good with the why and the how to approve?

frontend/src/app/[locale]/opportunity/[id]/page.tsx
```diff
@@ -113,6 +113,13 @@ async function OpportunityListing({ params }: OpportunityListingProps) {
     ? opportunityData.summary
     : emptySummary();
 
+  if (id !== opportunityData.opportunity_id) {
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/opportunity/[id]/page.tsx:112]: [nit] this could be moved up to line 112

doug-s-nava (approved):
✅ ✅ ✅ ✅

frontend/src/app/[locale]/opportunity/[id]/page.tsx
Comment (mdragon) [reply frontend/src/app/[locale]/opportunity/[id]/page.tsx:112]: @doug-s-nava nit addressed and rebased to main

doug-s-nava (approved):

Issue #5172, opened 2025-05-27
Fix typography on `SearchListItem`


andycochran:
### Summary

The recent [typography improvements](https://github.com/HHS/simpler-grants-gov/pull/5115) adjusted the type scale of headings. The headings are too large: 

![Image](https://github.com/user-attachments/assets/3bdc295c-aa87-45b2-af02-a2eb85baafd3)

The Search result items should be `h3` instead of `h2`. The "1,555 Opportunities" is an `h2` and the items are nested under it. 

### Acceptance criteria

- [ ] `SearchListItem` heading are changed from `h2` to `h3`

andycochran:
No longer relevant w/ the latest table view change

PR #5115, opened 2025-05-20
clean up typography

andycochran:
## Summary

Fixes #5082 

## Changes proposed

TBD — THIS PR IS STILL IN PROGRESS

## Context for reviewers

… Gonna be a bog boy. This touches a lot.

## Validation steps

…

PR #5645, opened 2025-07-09
[Issue #5172] Fix typography on `SearchListItem`

andycochran:
## Summary

Work for #5172

## Changes proposed

- Add `headingLevel` to `SearchResultListItem` (defaults to `h3`) 
- Pass "h2" to `headingLevel` on the saves opportunities page
- Add explicit font size class to heading 

## Context for reviewers

With the recent [typography improvements](https://github.com/HHS/simpler-grants-gov/pull/5115), the headings on the search results are too large. 

We can't simply make them `h3` instead of `h2`. On the Search page, they can be `h3` since there's an `h2` above the list (the opportunity count). However, they can't be `h3` on the Saved Opportunities page since it would be an a11y violation to skip heading levels (there's no `h2` between the page's `h1` and the list). 

But the search result list item heading should be visually the same size, regardless of which tag is used. 

## Validation steps

- Go to the Search page and note that the headings are `h3`
- Go to the Saved Opportunities page and note that the headings are `h2` 
- Note that the headings are the same size (`1.38rem`) on both pages

### DIFF
--- a/.github/workflows/cd-api.yml
+++ b/.github/workflows/cd-api.yml
@@ -2,13 +2,13 @@ name: Deploy API
 run-name: Deploy ${{ inputs.version || 'main' }} to API ${{ inputs.environment || (github.event_name == 'release' && 'prod') || 'nonprod' }}
 
 on:
-  # push:
-  #   branches:
-  #     - "main"
-  #   paths:
-  #     - "api/**"
-  #     - "infra/api/**"
-  #     - "infra/modules/**"
+  push:
+    branches:
+      - "main"
+    paths:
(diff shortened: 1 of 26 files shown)


doug-s-nava:
should have brought this up sooner, but this display is going away next week (with the exception of the saved opportunities page)

andycochran:
@doug-s-nava, up to you if you wanna just toss this PR. But it does make the Saved Opportunities page less jumbo easy-reader

image

andycochran:
No longer relevant w/ the recent table view change

Issue #5035, opened 2025-05-13
Utilize tinstructions table to generate and include the InstructionsURL in the GetOpportunityList response


freddieyebra:
### Summary

In order to generate the InstructionsURL for the GetOpportunityList response, we will need import the tinstructions table from the legacy oracle database.

### Acceptance criteria
- [ ] Create logic to query the table

freddieyebra:
Consider running these transformations with the competitions.

PR #5649, opened 2025-07-09
[Issue #5035] Utilize tinstruction data for GetOpportunityList response

freddieyebra:
## Summary

Work for #5035

## Changes proposed

This PR queries the staging.tinstructions table to build urls for the GetOpportunityList soap response. Specifically, this is required for building the schema_url, and instructions_url for the opportunity.

## Context for reviewers

The tinstructions data is queried from the staging schema since we are not transforming this data into the api schema.

### DIFF
--- a/infra/api/app-config/dev.tf
+++ b/infra/api/app-config/dev.tf
@@ -7,6 +7,7 @@ module "dev_config" {
   network_name                    = "dev"
   domain_name                     = "api.dev.simpler.grants.gov"
   s3_cdn_domain_name              = "files.dev.simpler.grants.gov"
+  mtls_domain_name                = "soap.dev.simpler.grants.gov"
   enable_https                    = true
   has_database                    = local.has_database
   database_enable_http_endpoint   = true
--- a/infra/api/app-config/env-config/outputs.tf
+++ b/infra/api/app-config/env-config/outputs.tf
@@ -28,6 +28,7 @@ output "service_config" {
     service_name                    = "${local.prefix}${var.app_name}-${var.environment}"
     domain_name                     = var.domain_name
     s3_cdn_domain_name              = var.s3_cdn_domain_name
+    mtls_domain_name                = var.mtls_domain_name
     enable_https                    = var.enable_https
     region                          = var.default_region
     cpu                             = var.instance_cpu
--- a/infra/api/app-config/env-config/variables.tf
+++ b/infra/api/app-config/env-config/variables.tf
@@ -30,6 +30,11 @@ variable "s3_cdn_domain_name" {
   default     = null
 }
 
+variable "mtls_domain_name" {
+  type        = string
+  description = "The domain name for the mTLS side-by-side ALB for the API"
+  default     = null
+}
 variable "enable_command_execution" {
   type        = bool
   description = "Enables the ability to manually execute commands on running service containers using AWS ECS Exec"
--- a/infra/api/app-config/prod.tf
+++ b/infra/api/app-config/prod.tf
@@ -8,6 +8,7 @@ module "prod_config" {
   domain_name                     = "api.simpler.grants.gov"
   enable_https                    = true
   s3_cdn_domain_name              = "files.simpler.grants.gov"
+  mtls_domain_name                = "soap.simpler.grants.gov"
   has_database                    = local.has_database
   database_enable_http_endpoint   = true
   has_incident_management_service = local.has_incident_management_service
--- a/infra/api/app-config/staging.tf
+++ b/infra/api/app-config/staging.tf
@@ -7,6 +7,7 @@ module "staging_config" {
   network_name                    = "staging"
   domain_name                     = "api.staging.simpler.grants.gov"
   s3_cdn_domain_name              = "files.staging.simpler.grants.gov"
+  mtls_domain_name                = "soap.staging.simpler.grants.gov"
   enable_https                    = true
   has_database                    = local.has_database
   database_enable_http_endpoint   = true
--- a/infra/api/service/main.tf
+++ b/infra/api/service/main.tf
@@ -109,6 +109,12 @@ data "aws_acm_certificate" "s3_cdn_cert" {
   most_recent = true
 }
 
+data "aws_acm_certificate" "mtls_cert" {
+  count       = local.service_config.mtls_domain_name != null ? 1 : 0
+  domain      = local.service_config.mtls_domain_name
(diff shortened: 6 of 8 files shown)


api/src/legacy_soap_api/applicants/services/get_opportunity_list_response.py
```diff
@@ -38,6 +49,43 @@ def get_competitions_from_opportunity_list_request(
...
+            select(Tinstructions.extension, Tinstructions.comp_id).where(
+                Tinstructions.comp_id.in_(legacy_competition_ids)
+            )
+        )
+        .mappings()
+        .all()
+    )
+    if not tinstructions:
+        return {}
+    return {tinstruction.comp_id: tinstruction.extension for tinstruction in tinstructions}
```
chouinar: I believe if the list is empty, the generator should produce the same result.
```suggestion
    return {tinstruction.comp_id: tinstruction.extension for tinstruction in tinstructions}
```

api/src/legacy_soap_api/applicants/services/get_opportunity_list_response.py
```diff
@@ -1,22 +1,32 @@
...
     OpportunityFilter,
 )
+from src.legacy_soap_api.legacy_soap_api_config import get_soap_config
+from src.legacy_soap_api.legacy_soap_api_utils import bool_to_string, ensure_dot_prefix
 
 logger = logging.getLogger(__name__)
 
 
+@dataclass
+class TinstructionsURLs:
```
chouinar: I wouldn't make a class name plural, reading the usage of it below I interpreted it as a list at first even though you did `TinstructionsURLs()`.

api/src/legacy_soap_api/applicants/services/get_opportunity_list_response.py
```diff
@@ -38,6 +49,43 @@ def get_competitions_from_opportunity_list_request(
...
+
+
+def get_tinstructions_urls(competition: Competition, tinstructions_map: dict) -> TinstructionsURLs:
+    urls = TinstructionsURLs()
+    if not competition.legacy_package_id:
+        return urls
+    base_url = f"{get_soap_config().grants_gov_uri}/apply/opportunities"
+    urls.schema_url = f"{base_url}/schemas/applicant/{competition.legacy_package_id}.xsd"
+    if extension := tinstructions_map.get(competition.legacy_competition_id):
+        urls.instructions_url = f"{base_url}/instructions/{competition.legacy_package_id}-instructions{ensure_dot_prefix(extension)}"
```
chouinar: For building path strings, could you use `file_util.join()` - it _should_ handle any issues with having an extra slash in one of the strings, just in case.

api/src/legacy_soap_api/applicants/services/get_opportunity_list_response.py
```diff
@@ -68,26 +116,38 @@ def _get_competitions(
...
+) -> GetOpportunityListResponse:
+    if not competitions:
+        return GetOpportunityListResponse(opportunity_details=[])
+    return GetOpportunityListResponse(
+        opportunity_details=[
+            _build_get_opportunity_details(
+                competition, get_tinstructions_urls(competition, tcompetitions_map)
+            )
+            for competition in competitions
+        ]
```
chouinar: Suggestion - build the parts of the response conditionally, not the whole thing.
```suggestion
    opportunity_details = []
    if competitions:
        opportunity_details = [
            _build_get_opportunity_details(
                competition, get_tinstructions_urls(competition, tcompetitions_map)
            )
            for competition in competitions
        ]

   return GetOpportunityListResponse(opportunity_details=opportunity_details)
```

api/src/legacy_soap_api/legacy_soap_api_config.py
```diff
@@ -25,3 +32,9 @@ def model_post_init(self, _context: Any) -> None:
             except Exception:
                 # This except is to make sure the API still starts up, even if the value is malformed
                 logger.exception("Could not load soap auth content")
+
+
+@cache
+def get_soap_config() -> LegacySoapAPIConfig:
+    # This is cached since any changes the config will require app restart anyways.
+    return LegacySoapAPIConfig()
```
chouinar, line 40: Huh, wasn't aware of `@cache` - probably better than the approach we have been doing using globals to avoid re-making these env var config classes like https://github.com/HHS/simpler-grants-gov/blob/main/api/src/util/file_util.py#L22

Good find.

api/tests/src/db/models/factories.py
```diff
@@ -2454,6 +2454,11 @@ class Params:
         )
 
 
+class StagingTinstructionsFactory(AbstractStagingFactory):
+    class Meta:
+        model = staging.instructions.Tinstructions
+
```
chouinar, line 2460: Could we add a few fields to this factory so it could be used without having to pass in everything you want?

api/src/legacy_soap_api/applicants/services/get_opportunity_list_response.py
```diff
@@ -38,6 +50,47 @@ def get_competitions_from_opportunity_list_request(
...
+    urls.schema_url = file_util.join(
+        base_url, "schemas", "applicant", f"{competition.legacy_package_id}.xsd"
+    )
+    if extension := tinstructions_map.get(competition.legacy_competition_id):
+        urls.instructions_url = file_util.join(
+            base_url,
+            "instructions",
+            f"{competition.legacy_package_id}-instructions{ensure_dot_prefix(extension)}",
+        )
+    return urls
```
chouinar: Just to go alongside renaming the class, could you make these all `url` instead of `urls`?

api/tests/src/db/models/factories.py
```diff
@@ -2454,6 +2454,20 @@ class Params:
...
+    competition = factory.SubFactory(StagingTcompetitionFactory)
+    comp_id = factory.LazyAttribute(lambda c: c.competition.comp_id)
+    extension = factory.LazyFunction(lambda: random.choice(["pdf", "docx", ".docx"]))
+    mimetype = factory.LazyFunction(
+        lambda: random.choice(["application/pdf", "application/octet-stream"])
+    )
```
chouinar:
```suggestion
    mimetype = factory.Faker("mime_type")
```

freddieyebra: updated

chouinar (approved):

PR #5653, opened 2025-07-10
[Unticketed] Fix store opportunity versioning to be batched

chouinar:
## Changes proposed
Modify the store opportunity versioning task to be batched

## Context for reviewers
With the recreation of our DB in the UUID work, this task is failing because it is running out of memory. It does need to fetch a lot of data, we could probably adjust the query to fetch less, BUT batching it like this should also work, and except for the initial backfill, we're normally processing like 20 records an hour.

This general structure is identical to what I did for opportunity attachment batching, it _should_ work to reduce memory. Even if it doesn't, it should commit every 5k records, and the worst case scenario would be running the job a few times.

## Validation steps
Loaded some opportunities in locally to verify batching works (with a batch size of like 5) and will exit after 100 iterations.

### DIFF
--- a/api/src/task/opportunities/store_opportunity_version_task.py
+++ b/api/src/task/opportunities/store_opportunity_version_task.py
@@ -5,22 +5,55 @@ from enum import StrEnum
 from sqlalchemy import select
 from sqlalchemy.orm import selectinload
 
+import src.adapters.db as db
 from src.constants.lookup_constants import JobStatus
 from src.db.models.opportunity_models import OpportunityChangeAudit
 from src.db.models.task_models import JobLog
 from src.services.opportunities_v1.opportunity_version import save_opportunity_version
 from src.task.task import Task
+from src.util.env_config import PydanticBaseEnvConfig
 
 logger = logging.getLogger(__name__)
 
 
+class StoreOpportunityVersionConfig(PydanticBaseEnvConfig):
+    store_opportunity_version_batch_size: int = 5000
+
+
 class StoreOpportunityVersionTask(Task):
+
+    def __init__(self, db_session: db.Session):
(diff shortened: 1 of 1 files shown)


api/src/task/opportunities/store_opportunity_version_task.py
```diff
@@ -5,22 +5,55 @@
...
+
     class Metrics(StrEnum):
         OPPORTUNITIES_VERSIONED = "opportunities_versioned"
 
     def run_task(self) -> None:
-        with self.db_session.begin():
-            self.process_opportunity_versions()
+
+        batch_num = 0
+        while True:
```
joshtonava, line 36: Discussion: Instead of while True, could we also do 

```
while self.has_unprocessed_records:
```

api/src/task/opportunities/store_opportunity_version_task.py
```diff
@@ -5,22 +5,55 @@
...
+            batch_num += 1
+            with self.db_session.begin():
+                self.process_opportunity_versions()
+                logger.info("Ran a batch of store opportunity versions - committing results")
+
+            if not self.has_unprocessed_records:
+                break
+
+            if batch_num > 100:
```
joshtonava, line 45: Nit: 100 could be a parameter for max_batches or something similar incase we need to tweak the upper bound in the future.

chouinar: The attachment job does have that, was just making something quick here.

api/src/task/opportunities/store_opportunity_version_task.py
chouinar: I should clean this and the attachment one up, will make a ticket to follow-up

Issue #4940, opened 2025-05-06
Touch up layout to align with designs


chrisw-nava:
### Summary

Due to there being an interim design between introducing the newest filtering options (#4585) and introducing the table view (#4586) we may need some UI cleanup before calling this transition "done"

### Acceptance criteria

- [ ] TBD

eric-serrato95:
@doug-s-nava  This ticket was bumped around 3.6 to address any clean up once we implement new filters and table view component to search

doug-s-nava:
see https://docs.google.com/document/d/1uTtTHo_xyHn8ovjaC1VVOxtOx3UgWr-IMANvP3f07yA/edit?usp=sharing

doug-s-nava:
also fixed flash of label on search bar due to feature flag implementation

PR #5660, opened 2025-07-11
[Issue #4940] search results table always full width

doug-s-nava:
## Summary

Work for #4940

## Changes proposed

ensure search results table is always full width regardless of the size of contents

## Validation steps

1. start a local server on main
2. update local data so that one opportunity has very short agency name and opportunity title. I used `USDA-CSREE` for the agency
3. run `make populate-search-opportunities` in api
4. visit http://localhost:3000/search?_ff=searchTableOn:true
5. enter the name of your renamed opportunity from step 2
6. _VERIFY_: the table is not full width
7. restart server on this branch and refresh page
8. _VERIFY_: the table is full width

#### Before
Screenshot 2025-07-11 at 10 47 17 AM

#### After
Screenshot 2025-07-11 at 10 46 52 AM

### DIFF
--- a/frontend/src/components/TableWithResponsiveHeader.tsx
+++ b/frontend/src/components/TableWithResponsiveHeader.tsx
@@ -71,7 +71,7 @@ export const TableWithResponsiveHeader = ({
                   {headerContent[j].cellData}
                 </div>
                 <div
-                  className="flex-2 tablet:flex-3"
+                  className="flex-2"
                   data-testid={`responsive-data-${i}-${j}`}
                 >
                   {tableCell.cellData}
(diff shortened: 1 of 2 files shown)


doug-s-nava:
here is what i see in firefox at the width right before breaking into mobile view. It looks ok to me, though it's also unlikely any users will be using the site at this width. How does this compare to what you're seeing @freddieyebra ? (github seems to not like this long image - recommend downloading for a comparison)

Screenshot 2025-07-11 at 12-17-37 Search Funding Opportunities Simpler Grants gov

freddieyebra (approved):

PR #5668, opened 2025-07-11
[Unticketed] Fix opportunity versioning to use a bool queue

chouinar:


api/src/db/migrations/versions/2025_07_11_add_version_queue_column.py
```diff
@@ -0,0 +1,49 @@
...
+down_revision = "448ee11789b1"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        "opportunity_change_audit",
+        sa.Column("has_been_versioned", sa.Boolean(), nullable=True),
```
mikehgrantsgov: Should we index on this too? Does the trigger function `update_opportunity_search_queue()` need an update?

api/src/db/models/opportunity_models.py
```diff
@@ -477,6 +477,7 @@ class OpportunityChangeAudit(ApiSchemaTable, TimestampMixin):
     )
     opportunity: Mapped[Opportunity] = relationship(Opportunity)
     is_loaded_to_search: Mapped[bool | None]
+    has_been_versioned: Mapped[bool | None]
```
mikehgrantsgov: Just a small nit, should we be consistent with `is_` naming, like `is_versioned`. Definitely not a blocking thing.. more philosophical

api/src/db/migrations/versions/2025_07_11_add_version_queue_column.py
chouinar: Yes, we should add an index. I updated the trigger function to set this.

api/src/db/models/opportunity_models.py
chouinar: Yeah, I should be consistent, will adjust

mdragon (approved):
Looks good.

Issue #5239, opened 2025-06-04
[Placeholder] Adjusting the API to allow users to skip validation for non-required forms


chouinar:
### Summary

TODO - design is working on how this'll work

But most likely:
* Boolean on app form table
* Let someone update that boolean in the app form update endpoint
* When doing validation, if a form isn't required, and isn't meant to be included, skip it

### Acceptance criteria

- [ ] TODO

PR #5671, opened 2025-07-14
[Issue #5239] Allow users to skip validation for non-required forms

mikehgrantsgov:
## Summary

Work for #5239

## Changes proposed

Added new validation logic for `is_included_in_submission` flag on application forms
Added new validation error type `MISSING_INCLUDED_IN_SUBMISSION`
Updated `validate_application_form` to implement conditional validation logic based on form requirement status / inclusion flag
Add tests

## Context for reviewers

Gives users explicit control over validation for non-required forms. Previously, all forms (required and non-required) would always run JSON schema validation

## Validation steps

See updated unit tests

### DIFF
--- a/api/src/services/applications/application_validation.py
+++ b/api/src/services/applications/application_validation.py
@@ -168,15 +168,44 @@ def validate_application_form(
     """Validate an application form, and set the current application form status"""
     form_validation_errors: list[ValidationErrorDetail] = []
 
+    # Always run rule validation
     context = JsonRuleContext(application_form, config=_get_json_rule_config_for_action(action))
     process_rule_schema_for_context(context)
     form_validation_errors.extend(context.validation_issues)
 
     # TODO pre/post-populate should update the value here
 
-    form_validation_errors.extend(
-        validate_json_schema_for_form(application_form.application_response, application_form.form)
-    )
+    # Check if the form is required
+    is_required = is_form_required(application_form)
+
+    # Handle validation based on form requirement and is_included_in_submission flag
+    should_run_json_schema_validation = True
+
+    if not is_required:
+        # For non-required forms, check is_included_in_submission
+        if application_form.is_included_in_submission is None:
+            # Add validation issue but don't run JSON schema validation
+            form_validation_errors.append(
+                ValidationErrorDetail(
+                    message="is_included_in_submission must be set on all non-required forms",
+                    type=ValidationErrorType.MISSING_INCLUDED_IN_SUBMISSION,
+                    field="is_included_in_submission",
+                    value=None,
+                )
+            )
+            should_run_json_schema_validation = False
+        elif application_form.is_included_in_submission is False:
+            # Don't run JSON schema validation if form is not included in submission
+            should_run_json_schema_validation = False
+        # If is_included_in_submission is True, run validation (default behavior)
+
+    # Run JSON schema validation only if required
+    if should_run_json_schema_validation:
+        form_validation_errors.extend(
+            validate_json_schema_for_form(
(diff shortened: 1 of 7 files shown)


api/src/services/applications/application_validation.py
```diff
@@ -174,9 +174,37 @@ def validate_application_form(
 
     # TODO pre/post-populate should update the value here
 
-    form_validation_errors.extend(
-        validate_json_schema_for_form(application_form.application_response, application_form.form)
-    )
+    # Check if the form is required
+    is_required = is_form_required(application_form)
```
chouinar, line 177: Another adjustment we need to make that I didn't document well, right now in the `get_application_form_errors` function that calls this we have a check that does:

```py
        if len(application_form.application_response) == 0:
            if is_app_form_required:
                # add an error

            continue
```
This is actually something we want to replace with the change you have here. The idea before was if someone hadn't started a form, and it wasn't required, we'd let that through, but now we want to be more explicit that they have to say yes/no for now-required forms.

This is all to say, I think we can remove that if-statement above. Even the case where a form is required, but we'd put "missing_required_form" is a bit clunky of a check as once we start pre-populating fields, it'll be rare that a form is ever truly empty.

api/src/services/applications/application_validation.py
```diff
@@ -164,21 +148,61 @@ def _get_json_rule_config_for_action(action: ApplicationAction) -> JsonRuleConfi
...
+    # Handle validation based on form requirement and is_included_in_submission flag
+    should_run_json_schema_validation = True
+
+    if not is_required:
+        # For non-required forms, check is_included_in_submission
+        if application_form.is_included_in_submission is None:
+            # If the form has no content and is_included_in_submission is None,
+            # treat it as not included in submission (don't validate)
+            if len(application_form.application_response) == 0:
+                should_run_json_schema_validation = False
```
chouinar: I think we shouldn't have this `len(application_form.application_response) == 0` check. If a form isn't required, but hasn't set the `is_included_in_submission` flag, its just an error regardless of whether they've started it.

mikehgrantsgov: I think this recent change captures what you're referring to

api/src/services/applications/application_validation.py
```diff
@@ -190,7 +209,7 @@ def validate_application_form(
 
     application_form.application_form_status = application_form_status  # type: ignore[attr-defined]
 
-    return form_validation_errors
+    return form_validation_errors, inclusion_errors
```
chouinar: Could we adjust this to return just one error list as before? I get that for the form status we want to work around it, but anything that calls this method just wants all the errors in one list and we shouldn't need to add the lists together in each of the calling places.

api/src/services/applications/application_validation.py
```diff
@@ -114,31 +114,26 @@ def get_application_form_errors(
...
-                )
-            # Don't add the form validation errors below if they haven't started yet
-            continue
-
-        if form_validation_errors:
-            form_error_map[str(application_form.application_form_id)] = form_validation_errors
+        validation_errors = validate_application_form(application_form, action)
+
+        # Separate inclusion errors from other validation errors
+        inclusion_errors = [
```
mikehgrantsgov, line 120: @chouinar this break up is slightly clunky but I thought better than iterating over the list and adding to two groups separately?

chouinar: I think this is fine, we might want to check with the frontend on how they'd want to parse this particular error out, but I'm guessing they wanted it separated out like you've done here.

chouinar (approved):

Issue #5052, opened 2025-05-15
Generate metadata file for ZIP


mdragon:
### Summary

After #5263 

We want to generate a metadata file during our process that creates an application submission. This metadata file should be a simple text file named `manifest.txt`. It should roughly look like:

```
Manifest for Grant Application

Forms included in ZIP (total X):
 1. Form FORM_ABC.pdf (size 123 bytes)
 2. Form FORM_XYZ.pdf (size 456 bytes)

Attachments included in ZIP (total X):
 1. my_attachment.pdf (size 678 bytes)
 2. my_file.docx (size 123 bytes)
```
There should be information for every form and attachment included in the file.

Each form should have:
* The form name
* The size in bytes

Each attachment should have:
* The attachment name (note - logic elsewhere may require we slightly rename things for duplicates, do that here as well)
* The size in bytes

This file should be written to the ZIP that the rest of the files are.

### Acceptance criteria

- [ ] Process setup to create a manifest when creating application submission ZIP
- [ ] Tests added

chris-kuryak:
This is needed to prove out that handling PDFs in this manner will work for our purposes.

chris-kuryak:
This will only be needed if the simpler API approach

chris-kuryak:
Repurposed per Michael C for backend use since we didn't need the FE ticket anymore.

PR #5679, opened 2025-07-14
[Issue #5052] Generate a metadata file in the application submission ZIP

chouinar:
## Summary

Work for #5052 

## Changes proposed
Creates a metadata/manifest file in the application submission ZIP saying what files are present, and their size.

Adds a new column to the form table for storing the short form name

## Context for reviewers
Had to make a change to the form table to have a "short form name" which we use for the file name - values I set for the forms I've setup come from the old system which had this column. Also added a legacy form ID column while I was at it as we'll need that eventually.

I added a bit more code to the logic for processing application forms to make a dummy file so we can have the manifest see it - the application forms will be PDFs we generate in a later PR, so we'll be able to just plug that in when it is time.

The file contents match what grants.gov does for its manifest.

## Validation steps
You can setup some applications by doing `make console` and then:
```py
from src.constants.lookup_constants import ApplicationStatus
f.ApplicationFactory.create_batch(size=10, application_status=ApplicationStatus.SUBMITTED, with_attachments=True, with_forms=True)
```

You can run the job by doing: `make cmd args="task create-application-submission"`

To fetch an application submission you can either use the s3 CLI, or just build the URL yourself and go to `http://localhost:4566/local-mock-draft-bucket/applications/{application_id}/submissions/{submission_id}/submission-{application_id}.zip` which will download for you (can get values from the logs).

Here are the contents of one zip + the manifest generated:
Screenshot 2025-07-14 at 4 49 28 PM


mikehgrantsgov (approved):
Looks good!

Issue #5321, opened 2025-06-11
Create admin endpoint to update forms in the db


chris-kuryak:
### Summary

To make editing/updating forms easier as we create them, we want to create an admin endpoint for updating forms in the database.

### Acceptance criteria

- [ ]

PR #5686, opened 2025-07-15
[Issue #5321] Create admin endpoint to update forms in the db

mikehgrantsgov:
## Summary

Work for #5321

## Changes proposed

Added `update_form.py` service for form upsert operations
Added PUT `/alpha/forms/{form_id}` admin endpoint
Updated form schemas with request/response fields
Added unit tests

## Context for reviewers

The endpoint allows administrators (a specific token only) to create new forms or update existing ones with proper validation.

## Validation steps

See new unit tests

### DIFF
--- a/api/src/services/form_alpha/update_form.py
+++ b/api/src/services/form_alpha/update_form.py
@@ -62,5 +62,4 @@ def update_form(db_session: db.Session, form_id: uuid.UUID, form_data: dict) -> 
     # Set the form instruction relationship
     form.form_instruction = form_instruction
 
-
     return form


api/src/services/form_alpha/update_form.py
```diff
@@ -0,0 +1,87 @@
...
+            form_version=form_data["form_version"],
+            agency_code=form_data["agency_code"],
+            omb_number=form_data.get("omb_number"),
+            form_json_schema=form_data["form_json_schema"],
+            form_ui_schema=form_data["form_ui_schema"],
+            form_instruction_id=form_instruction_id,
+            form_rule_schema=form_data.get("form_rule_schema"),
+        )
+        db_session.add(new_form)
+        db_session.flush()  # Flush to get the ID assigned
```
chouinar: Don't need to flush since we set the ID ourselves

api/src/api/form_alpha/form_schema.py
```diff
@@ -70,6 +84,89 @@ class FormAlphaSchema(Schema):
...
+    updated_at = fields.DateTime(
+        metadata={"description": "The timestamp when the form was last updated"}
+    )
+
+
+class FormResponseAlphaSchema(WarningMixinSchema, AbstractResponseSchema):
+    data = fields.Nested(FormAlphaSchema)
+
+
+class FormUpdateRequestSchema(Schema):
```
chouinar, line 115: I just added two new fields `short_form_name` and `legacy_form_id` to the form table, could you add them to this schema as well?

Didn't realize until my recent PR that we needed those fields.

api/src/services/form_alpha/update_form.py
chouinar: Since we're just modifying a single DB object from a dict, we can be a bit lazy and use setattr. Also means if we add any new fields we don't even need to change this code.
```suggestion
    if form is None:
        logger.info("Creating new form", extra={"form_id": form_id})
        form = Form()
        db_session.add(form)
   else:
        logger.info("Updating existing form", extra={"form_id": form_id})
        
    for field, value in form_data.items():
         setattr(form, field, value)
```

api/src/services/form_alpha/update_form.py
```diff
@@ -0,0 +1,87 @@
...
+        # Re-fetch with relationships loaded
+        form_with_relationships = db_session.scalar(
+            select(Form).options(selectinload(Form.form_instruction)).where(Form.form_id == form_id)
+        )
+
+        # This should never be None since we just created it, but handle it to satisfy mypy
+        if form_with_relationships is None:
+            raise_flask_error(500, "Failed to create form")
+
+        return form_with_relationships
```
chouinar: If we want the form to have the instructions added, if we have the back-populates configured right, then
```py
form.form_instruction = form_instruction
```
Should set it? Rather than setting the ID, we set the relationship of the form instruction object we found. We'd want to pop out the `form_instruction_id` above and not set it directly.

api/src/services/form_alpha/update_form.py
```diff
@@ -0,0 +1,77 @@
...
+    if existing_form is None:
+        form_with_relationships = db_session.scalar(
+            select(Form).options(selectinload(Form.form_instruction)).where(Form.form_id == form_id)
+        )
+
+        # This should never be None since we just created it, but handle it to satisfy mypy
+        if form_with_relationships is None:
+            raise_flask_error(500, "Failed to create form")
+
+        return form_with_relationships
```
chouinar: This is fetching the form we already have and the form instruction we already have too. The line right above this `form.form_instruction = form_instruction` makes certain that the relationship is set already.

```suggestion
```

api/src/services/form_alpha/update_form.py
mikehgrantsgov: Ah got it, I think we're good now.

Issue #5555, opened 2025-07-01
In Proxy endpoint, flag who fulfills the request: legacy vs Simpler


chris-kuryak:
### Summary

To meet our SOAP Proxy/Router analytics needs, we need to be able to identify requests fulfilled by the legacy SOAP API or the new Simpler REST API. To assist with this, we want to add a flag indicating which API fulfilled the request.

Note that for the initial implementation, all responses will be fulfilled by GG.

### Acceptance criteria

-[ ] Visibility in new relic logs or dashboard around whether the response was from simpler or GG

PR #5709, opened 2025-07-17
[Issue #5555] Add flag for how soap response returned

freddieyebra:
## Summary

Work for #5555

## Changes proposed

This PR includes adding logs around who the SOAP response was returned as well as refactoring. These changes include the following improvements:

1. Ensure all unhandled exceptions return valid xml responses instead of flask json.
2. Move soap proxy request/response earlier in endpoint and ensure no simpler logic prevents responding with the proxy response.
3. Entry point for dict comparison.
4. Improved logging.

## Context for reviewers

This is largely an effort of cleaning up the soap proxy work and better error handling and logging.


api/src/legacy_soap_api/simpler_soap_api.py
```diff
@@ -0,0 +1,78 @@
...
+            logger.info(
+                "simpler_soap_api: Successfully processed request and returning Simpler SOAP response",
+                extra={"used_simpler_response": use_simpler},
+            )
+            return simpler_soap_response
+    except (SOAPOperationNotSupported, SOAPInvalidEnvelope, SOAPInvalidRequestOperationName) as e:
+        err = f"Unable to initialize Simpler SOAP client: {e}"
+        logger.info(
+            f"simpler_soap_api: {err}",
+            extra={"simpler_soap_api_error": err, "used_simpler_response": use_simpler},
```
chouinar: If the request is invalid, does it dump the request in the error message?

api/src/legacy_soap_api/simpler_soap_api.py
```diff
@@ -0,0 +1,78 @@
...
+            },
+        )
+        return soap_proxy_response
+
+    try:
+        simpler_soap_response, use_simpler = simpler_soap_client.get_simpler_soap_response(
+            soap_proxy_response
+        )
+
+        if simpler_soap_response is not None and use_simpler:
```
chouinar: I believe this is where we'd know if we were using the simpler response or not, right, based on what was returned here?

What if we add this here and then don't need to worry about it anywhere below?
```py
        add_extra_data_to_current_request_logs({"used_simpler_response": use_simpler})
```

Actually, there might be another place we want this. While we might be able to say we used simpler here, what do we actually return? What if there is a difference (when that logic gets added in the future), do we return our response or grants.gov's? What about when an endpoint might make sense to merge their responses (application list probably needs that?).

It might not be a binary "used simpler or not" but we might want to mark both of them? That might be a detail we circle back on later.

api/src/legacy_soap_api/legacy_soap_api_proxy.py
```diff
@@ -0,0 +1,78 @@
...
+            cert = soap_request.auth.certificate.get_pem(config.soap_auth_map)
+        except SOAPClientCertificateLookupError as e:
+            # This exception handles invalid client certs. We will continue to return the response
+            # from GG.
+            cert = ""
+            logger.info(f"soap_client_certificate: Unknown or invalid client certificate: {e}")
+        except SOAPClientCertificateNotConfigured as e:
+            # This exception handles the case of a valid cert being passed, but not configured
+            # to use Simpler SOAP API.
+            logger.info(f"soap_client_certicate: Certificate validated but not configured: {e}")
```
chouinar, line 61: Idea: If we wanted to have a good way to create graphs/separate out logs for when these happen, we could add some consistent value to the extra object of these logs.

Something like:
```py
logger.info("whatever message", extra={"soap_error": "soap_client_certificate_lookup"})
```
Then we could easily see how often each of the different error cases happen on a graph that just facets by `soap_error`.

api/src/legacy_soap_api/legacy_soap_api_config.py
```diff
@@ -38,3 +40,46 @@ def model_post_init(self, _context: Any) -> None:
...
+            request_operation_name="GetOpportunityListRequest",
+            response_operation_name="GetOpportunityListResponse",
+            force_list_attributes=("OpportunityDetails",),
+        )
+    },
+    SimplerSoapAPI.GRANTORS: {},
+}
+
+
+@lru_cache(maxsize=SIMPLER_SOAP_OPERATION_CONFIG_CACHE_SIZE)
```
chouinar: I assume we'll eventually have a config for each endpoint of which there are 30-40ish? I think we'd want the cache size to be at least that big, the default of 128 is probably fine.

api/src/legacy_soap_api/simpler_soap_api.py
freddieyebra: It will not. It will only log the message passed to the exception:

```
grants-api  | 15:17:28.660  src.legacy_soap_api.simpler_soap_api get_simpler_soap_response    INFO     simpler_soap_api: Simpler grantors SOAP API does not support GetOpportunityListRequest simpler_soap_api_error=Simpler grantors SOAP API does not support GetOpportunityListRequest used_simpler_response=False app.name=src.app app_name=api run_mode=service environment=local deploy_whoami=local-developer request.id= request.method=POST request.path=/grantsws-agency/services/v2/AgencyWebServicesSoapPort request.url_rule=/<service_name>/services/v2/<service_port_name> request.internal_id=47deab73-bb5d-47ab-bd03-b31b36e1ad41 soap_api=grantors soap_request_operation_name=GetOpportunityListRequest
```

api/src/legacy_soap_api/legacy_soap_api_proxy.py
freddieyebra: I created a follow up ticket to improve the logs to add this [here](https://github.com/HHS/simpler-grants-gov/issues/5731)

api/src/legacy_soap_api/legacy_soap_api_config.py
freddieyebra: updated

Issue #5711, opened 2025-07-17
Address Vuln Scan results


mdragon:
### Summary

https://github.com/HHS/simpler-grants-gov/actions/runs/16344604273

Deal with Docker base image vulnerabilities 

### Acceptance criteria

Builds pass.

PR #5712, opened 2025-07-17
[Issue #5711] Update package installs for vulnerabilities

mdragon:
## Summary

Work for #5711

## Changes proposed
Didn't actually change what we force to install, actually removed a couple older ones since we probably didn't need them anymore.

## Validation steps

PR tests are passing


joshtonava (approved):

doug-s-nava (approved):
not sure i understand how this clears things up (addition by subtraction?) but good to unblock

PR #5723, opened 2025-07-18
Add a new env var (AWS_STORAGE_BUCKET_NAME) to NOFOs app container

pcraig3:
## Summary

This is prep work for an upcoming task to build an image uploading feature.

## Changes proposed

Add this new env var to the NOFOs app container task definition: `AWS_STORAGE_BUCKET_NAME`

Should be self-evident what it is for.

### DIFF
--- a/infra/nofos/app-config/env-config/environment_variables.tf
+++ b/infra/nofos/app-config/env-config/environment_variables.tf
@@ -26,5 +26,10 @@ locals {
       manage_method     = "manual"
       secret_store_name = "/nofos/${var.environment}/secret-key"
     }
+
+    AWS_STORAGE_BUCKET_NAME = {
+      manage_method     = "manual"
+      secret_store_name = "/nofos/${var.environment}/aws-storage-bucket-name"
+    }
   }
 }


infra/nofos/app-config/env-config/environment_variables.tf
```diff
@@ -26,5 +26,10 @@ locals {
       manage_method     = "manual"
       secret_store_name = "/nofos/${var.environment}/secret-key"
     }
+
+    AWS_STORAGE_BUCKET_NAME = {
+      manage_method     = "manual"
+      secret_store_name = "/nofos/${var.environment}/aws-storage-bucket-name"
+    }
```
chouinar, line 33: I'm not sure if the NOFOs stuff is setup different, but if you want the s3 bucket to be an env var, Kai had setup the bucket to become an env var automatically: https://github.com/HHS/simpler-grants-gov/blob/main/infra/api/app-config/env-config/s3_buckets.tf

This both handles creating the s3 bucket and making an env var. I'm not sure if there was config elsewhere needed to make this work (probably), but lets the s3 definition handle setting its own env var as well and not needing to configure it.

pcraig3: Thanks for this! I didn't realize but it makes sense to do it this way. 

For posterity, the env name is `GENERAL_S3_BUCKET_URL`.

pcraig3:
Closing this because the bucket name already exists as another variable.

Issue #5698, opened 2025-07-16
Create JSON for Budget Narrative Attachment form


chris-kuryak:
### Summary

Create JSON for Budget Narrative Attachment form:
https://apply07.grants.gov/apply/forms/readonly/BudgetNarrativeAttachments_1_2-V1.2.pdf

### Acceptance criteria

- [ ] JSON ready to be picked up by FE Engr for implementation

Issue #5697, opened 2025-07-16
Create JSON for Project Narrative Attachment Form


chris-kuryak:
### Summary

Create JSON for the Project Narrative Form:

https://apply07.grants.gov/apply/forms/readonly/ProjectNarrativeAttachments_1_2-V1.2.pdf

### Acceptance criteria

- [ ] JSON ready to be picked up by FE engineers for implementation
- [ ] Form class should be created in https://github.com/HHS/simpler-grants-gov/tree/main/api/src/form_schema/forms
- [ ] All validation should be added including required fields, field lengths, enums, and conditionally required if-else fields
- [ ] Tests are added to verify the validation was setup and works as expected
- [ ] If any fields need pre/post/custom validation, a rule schema is configured
- [ ] Fields are renamed to be formatted as `my_example_field` instead of `MyExampleField`

PR #5726, opened 2025-07-18
[Issue #5697 & #5698] Add Budget & Project Narrative Attachment Forms

chouinar:
## Summary

Work for #5697
Work for #5698 

## Changes proposed
Add JSON schema for Budget Narrative Attachment Form
Add JSON schema for Project Narrative Attachment Form

## Context for reviewers
These two forms are identical as far as the schemas go with just a minor difference in file name and some of the text a user will see.

Note that in the current PDF the attachment is setup as two fields a "required" attachment and then an "optional" list of attachments. But in the underlying data schema, it's a list like what I've done here.

## Validation steps
`make db-seed-local` will create an opportunity+competition with these attached. They don't quite render right because the attachment logic isn't fully implemented on the frontend / UI schema would need to be updated too.

Screenshot 2025-07-18 at 1 38 42 PM
Screenshot 2025-07-18 at 1 51 58 PM

### DIFF
--- a/api/src/form_schema/forms/budget_narrative_attachment.py
+++ b/api/src/form_schema/forms/budget_narrative_attachment.py
@@ -1,0 +1,53 @@
+import uuid
+
+from src.db.models.competition_models import Form
+
+FORM_JSON_SCHEMA = {
+    "type": "object",
+    "required": ["attachments"],
+    "properties": {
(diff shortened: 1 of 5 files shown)


mikehgrantsgov (approved):
Looks good!

Issue #5729, opened 2025-07-19
Use full www URL when linking back to grants.gov


mdragon:
### Summary

During Simpler testing on Production, Andrew happened to notice that Simpler site uses href=https://grants.gov and not https://www.grants.gov .
 
Currently this is not an issue but, in the past, URL without the initial “www.” has led to problems sometimes. Ref: SCR-3738,  SCR-12199
 
Our recommendation is to consider including “www.” in the URL so the redirection can happen consistently.

### Acceptance criteria

Links back to grants.gov use www

mdragon:
Confirmed changes work as expected in Staging.

PR #5730, opened 2025-07-21
[Issue #5729] Fix up links to Grants.gov

mdragon:
## Summary

Work for #5729

## Changes proposed

Make the links to Grants.gov use full www.grants.gov in Prod and make it more aware of other environments.

## Validation steps

1. Checkout the branch
2. Load http://localhost:3000/search?utm_source=Grants.gov
3. Validate that the link for "Return to Grants.gov" is to "dev.grants.gov" (previously it would have been to "grants.gov"

### DIFF
--- a/frontend/src/constants/environments.ts
+++ b/frontend/src/constants/environments.ts
@@ -36,10 +36,24 @@ export const featureFlags = {
   searchTableOn: FEATURE_SEARCH_TABLE_ON,
 };
 
+const legacyHost = (): string => {
+  switch (ENVIRONMENT) {
+    case "prod":
+      return "https://www.grants.gov";
+    case "training":
+      return "https://training.grants.gov";
+    case "staging":
+      return "https://staging.grants.gov";
+    case "test":
+      return "https://test.grants.gov";
+    default:
+      return "https://dev.grants.gov";
+  }
+};
+
 // home for all interpreted server side environment variables
 export const environment: { [key: string]: string } = {
-  LEGACY_HOST:
-    ENVIRONMENT === "prod" ? "https://grants.gov" : "https://test.grants.gov",
+  LEGACY_HOST: legacyHost(),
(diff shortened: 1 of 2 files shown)


frontend/src/constants/environments.ts
```diff
@@ -36,10 +36,24 @@ export const featureFlags = {
...
+    case "prod":
+      return "https://www.grants.gov";
+    case "training":
+      return "https://training.grants.gov";
+    case "staging":
+      return "https://staging.grants.gov";
+    case "test":
+      return "https://test.grants.gov";
+    default:
+      return "https://dev.grants.gov";
```
chouinar: Our staging and dev environments should both point to `test` if we want the link to be useful.

frontend/src/constants/environments.ts
```diff
@@ -36,10 +36,24 @@ export const featureFlags = {
...
+    case "training":
+      return "https://training.grants.gov";
+    case "staging":
+      return "https://test.grants.gov";
+    case "test":
+      return "https://test.grants.gov";
+    default:
+      return "https://test.grants.gov";
+  }
+};
```
freddieyebra, line 52: small suggestion:
```suggestion
const legacyHost = (): string => {
  switch (ENVIRONMENT) {
    case "prod":
      return "https://www.grants.gov";
    case "training":
      return "https://training.grants.gov";
    case "staging":
    case "test":
    default:
      return "https://test.grants.gov";
  }
};
```

mdragon: I'm going to not do this, because our end state is that each of these is a unique 1:1 mapping, or we get HHS to agree to retire the environment. So I agree this is better if we were going to always collapse the bottom 3, but we're actively moving toward seperating them or killing them off.

freddieyebra (approved):
Looks good. Confirmed that the "Return to grants.gov" link goes to test.grants.gov locally.
Screenshot 2025-07-22 at 10 10 14 AM

Issue #5429, opened 2025-06-24
Return facet counts for close date filter


doug-s-nava:
### Summary

To support new filters for close date and cost sharing, the API should return facet counts for the following on all responses from the opportunity search endpoint:

```
{
  "close_date": {
    "end_date_relative": [7, 30, 90, 120]
  }
}
```

The ideal facet count response would look something like:

```
{
  "facet_counts": {
     "close_date": {
         "7": x,
         "30": x,
         "90": x,
         "120": x,
      }
   }
}
```

If this arbitrary bucketing is not possible or prohibitively difficult, we can work with something like this, and parse results on the frontend (I guess the main concern here would be overly bloating the payload):

```
{
  "facet_counts": {
     "close_date": {
         "1": x,
         "2": x,
         "3": x,
         ...
         "365": x,
      }
   }
}
```

Another option might be taking the bucket info in from the request - if something like this gets passed maybe it could inform how the buckets are populated in the response. The response could then just be an indexed array that the frontend can unpack based on the order in which the buckets were assigned in the request:

```
{
   "facet_count_buckets": {
      "close_date": [
         {
            "start": 0,
            "end": 7,
         },
         {
            "start": 8,
            "end": 30,
         },
         {
            "start": 31,
            "end": 90,
         },
         {
            "start": 91,
            "end": 120,
         },
      ]
   },
   "pagination": {
      ...
   },
   "filters": {
      ...
   }
}
```

This will require some preliminary research to determine approach, as we have not encountered bucketing something like date ranges for facet counts previously.

### Acceptance criteria

* response payload from opportunity search endpoint includes facet count information for close_date as described in one of the options above

doug-s-nava:
After discussing with @chouinar it looks like this can be accomplished using customized keys in the open search query. See https://betagrantsgov.slack.com/archives/C08MTT10SKU/p1750784661864059?thread_ts=1750782964.872999&cid=C08MTT10SKU

PR #5732, opened 2025-07-21
[Issue #5429] Facet count closed date

babebe:
Fixes https://github.com/HHS/simpler-grants-gov/issues/5429
Added  aggregation_relative_date_range func that allows for  static date-range aggregations. 
Currently handles only static ranges for [7, 30, 60, 90, 120] days from now. 
Updated aggregation parser to also handle date range aggregates
Added Test

### DIFF
--- a/frontend/src/components/ReturnToGrantsNotification.tsx
+++ b/frontend/src/components/ReturnToGrantsNotification.tsx
@@ -1,5 +1,6 @@
 "use client";
 
+import { environment } from "src/constants/environments";
 import SessionStorage from "src/services/sessionStorage/sessionStorage";
 
 import { useTranslations } from "next-intl";
@@ -10,7 +11,7 @@ export function ReturnToGrantsNotification() {
   const searchParams = useSearchParams();
   const GrantsLink = (
     <div className="display-flex flex-1 tablet:text-right tablet:margin-bottom-0 margin-bottom-3 tablet:padding-top-3 tablet:flex-justify-end">
-      <a href="https://grants.gov">{t("message")}</a>
+      <a href={environment.LEGACY_HOST}>{t("message")}</a>
     </div>
   );
   if (searchParams.get("utm_source") === "Grants.gov") {
--- a/frontend/src/components/search/SearchResultsTable.tsx
+++ b/frontend/src/components/search/SearchResultsTable.tsx
@@ -52,15 +52,22 @@ const CloseDateDisplay = ({ closeDate }: { closeDate: string }) => {
 const TitleDisplay = ({
   opportunity,
   saved,
+  page,
+  index,
 }: {
   opportunity: BaseOpportunity;
   saved: boolean;
+  page: number;
+  index: number;
 }) => {
   const t = useTranslations("Search.table");
   return (
     <>
       <div className="font-sans-lg text-bold">
-        <a href={getOpportunityUrl(opportunity.opportunity_id)}>
+        <a
+          href={getOpportunityUrl(opportunity.opportunity_id)}
+          id={`search-result-link-${page}-${index + 1}`}
+        >
           {opportunity.opportunity_title}
         </a>
       </div>
@@ -105,6 +112,8 @@ const AgencyDisplay = ({ opportunity }: { opportunity: BaseOpportunity }) => {
 const toSearchResultsTableRow = (
   result: BaseOpportunity,
   saved: boolean,
+  page: number,
+  index: number,
 ): TableCellData[] => {
   return [
     {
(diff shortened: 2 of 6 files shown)


api/src/adapters/search/opensearch_query_builder.py
```diff
@@ -345,3 +346,48 @@ def build(self) -> dict:
...
+
+        from_val = format_bound(date_filter.start_date, date_filter.start_date_relative)
+        to_val = format_bound(date_filter.end_date, date_filter.end_date_relative)
+
+        def key_label(val: datetime.date | int | None) -> str:
+            if isinstance(val, datetime.date):
+                return val.isoformat()
+            elif isinstance(val, int):
+                return str(val)
+            return "*"
```
chouinar: Lets avoid defining functions inline, just define them at the bottom of the file after the class if they don't need anything from the class.

api/src/adapters/search/opensearch_query_builder.py
```diff
@@ -345,3 +346,48 @@ def build(self) -> dict:
             request["aggs"] = self.aggregations
 
         return request
+
+    def aggregation_relative_date_range(
+        self, aggregation_name: str, field_name: str, date_filter: DateSearchFilter
+    ) -> typing.Self:
```
chouinar: The date filter should not be passed into this function - that wouldn't be a list of values like the ticket is saying.

I don't think anything from the request should actually get piped here, if I'm reading right it wants 4 buckets of relative dates for this (7, 30, 90, 120), so we should instead pass in a list of those which would just be statically defined in the API.

We'd want the request to the search index to be something like:
```json
{
  "aggs": {
    "summary.close_date": {
      "date_range": {
        "field": "date",
        "format": "YYYY-MM-dd",
        "ranges": [
            {"from": "now-7d/d", "to": "now", "key": "7"},
            {"from": "now-30d/d", "to": "now", "key": "30"},
            {"from": "now-90d/d", "to": "now", "key": "90"},
            {"from": "now-120d/d", "to": "now", "key": "120"}
          ]
      }
    }
  }
}
```

Because each range needs to define a start, end and key, perhaps we create a simple container class like:
```py
@dataclass
class DateRangeAgg:
     from: str
     to: str
     key: str
```
And then just statically define + use that and pass in a list of those values?

api/src/adapters/search/opensearch_response.py
```diff
@@ -106,24 +107,43 @@ def _parse_aggregations(
...
+            # Date_range aggregation
+            for k, v in buckets.items():
+                field_aggregation[k] = v.get("doc_count", 0)
+            aggregations[field] = field_aggregation
+
+        elif isinstance(buckets, list):
+            # Terms aggregation
+            for bucket in buckets:
+                key = bucket.get("key")
+                count = bucket.get("doc_count", 0)
```
chouinar: If we set the key to what we want it in the response, we should be able to make zero changes to the response parsing logic here

api/src/adapters/search/opensearch_query_builder.py
babebe: Removed not needed

api/src/adapters/search/opensearch_query_builder.py
babebe: From is reserved word so linter picks it up.

api/src/adapters/search/opensearch_response.py
babebe: Removed

api/src/adapters/search/opensearch_query_builder.py
```diff
@@ -3,6 +3,14 @@
...
 from src.pagination.pagination_models import SortDirection
 
+STATIC_DATE_RANGES: list = [
+    {"from": "now", "to": "now+7d/d", "key": "7"},
+    {"from": "now", "to": "now+30d/d", "key": "30"},
+    {"from": "now", "to": "now+60d/d", "key": "60"},
+    {"from": "now", "to": "now+90d/d", "key": "90"},
+    {"from": "now", "to": "now+120d/d", "key": "120"},
+]
+
```
chouinar: I'd make this a parameter we pass into the builder and we have it statically defined for the API itself. That way the builder supports it being flexible, we just don't use it yet.

api/src/services/opportunities_v1/search_opportunities.py
```diff
@@ -124,7 +124,9 @@ def _get_sort_by(pagination: PaginationParams) -> list[tuple[str, SortDirection]
     return sort_by
 
 
-def _add_aggregations(builder: search.SearchQueryBuilder) -> None:
+def _add_aggregations(
+    builder: search.SearchQueryBuilder, filters: OpportunityFilters | None
+) -> None:
```
chouinar: Don't think filters needs to be here anymore
```suggestion
def _add_aggregations(
    builder: search.SearchQueryBuilder
) -> None:
```

Issue #5744, opened 2025-07-22
Handle parsing GetOpportunityListResponse from proxy response


freddieyebra:
### Summary

Currently, the dates returned from the grants.gov s2s soap api are not formatted the same as in the database. This causes issues when trying to validate and diff simpler soap responses to gg soap api responses.

We need to handle the case of this format being different from what we store and persist in the database in order to be able diff the data.

To reproduce this locally:
1. Run the backend /api
2. Send the following curl command to proxy a request to gg training:
```
curl --verbose  'http://0.0.0.0:8080/grantsws-applicant/services/v2/ApplicantWebServicesSoapPort' \
--header 'X-GG-S2S-URI: https://trainingws.grants.gov' \
--header 'Content-Type: application/xml' \
--data '<soapenv:Envelope xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/" xmlns:app="http://apply.grants.gov/services/ApplicantWebServices-V2.0" xmlns:gran="http://apply.grants.gov/system/GrantsCommonElements-V1.0" xmlns:app1="http://apply.grants.gov/system/ApplicantCommonElements-V1.0">
   <soapenv:Header/>
   <soapenv:Body>
      <app:GetOpportunityListRequest>
            <app1:OpportunityFilter>
            <gran:FundingOpportunityNumber>O-BJA-2025-202930-STG</gran:FundingOpportunityNumber>
         </app1:OpportunityFilter>
      </app:GetOpportunityListRequest>
   </soapenv:Body>
</soapenv:Envelope>'
```
3. You will see the following info logs:
`Could not parse proxy response <class 'src.legacy_soap_api.applicants.schemas.get_opportunity_list_schemas.GetOpportunityListResponse'> into Simpler SOAP response pydantic schema. {('date_from_datetime_parsing', ('OpportunityDetails', 0, 'ClosingDate')), ('date_from_datetime_parsing', ('OpportunityDetails', 0, 'OpeningDate'))}
`

---
The issue is that grants.gov formats dates as `2025-09-16-04:00` and Pydantic expects dates to not have a time zone attached when parsing a date from a string. We'll need to look into a way to coerce dates with Pydantic to be parseable in this format.

### Acceptance criteria

- Able to run the above curl command and parse the proxy response.

freddieyebra:
@mikehgrantsgov you may be able to add something like 
```python
class OpportunityDetails(BaseSOAPSchema):
    ...
[...]

Issue #5224, opened 2025-06-02
Create method that diffs simpler soap response with grants.gov soap response


freddieyebra:
### Summary

In order to have visibility into any differences in data between GG and SGG soap responses, we first need to be able to supporting diffing the data in each response.

There are pydantic schemas that are used to validate response data from simpler. Since we are already retrieving the soap response from GG and have ability to convert that to a dict, we need to implement a method that puts both the simpler response dict, and the grants.gov response dict into these schemas to get the difference.

Use cases to support:
- Nested dicts
- Parameter to only retrieve key differences without values in order to keep PII out of logs.
- Support getting diff of dict keys in a list given a key for finding the dicts in each response to compare to.

### Acceptance criteria

- Method implementation
- Diff method is called and appropriate data is logged for soap requests.

PR #5745, opened 2025-07-22
[Issue #5224] Add dict diff for soap responses

freddieyebra:
## Summary

Work for #5224

## Changes proposed

This PR adds logic for diffing SOAP dict responses from the data queried within simpler against the dict from the proxy response and logs it for soap requests.

## Context for reviewers

This provides insights into any differences between the 2 responses. 

## Validation steps
- Unit tests pass
- Run the /api service locally
- Send the following curl command that proxys to training environment:
```
curl --verbose  'http://0.0.0.0:8080/grantsws-applicant/services/v2/ApplicantWebServicesSoapPort' \
--header 'X-GG-S2S-URI: https://trainingws.grants.gov' \
--header 'Content-Type: application/xml' \
--data '<soapenv:Envelope xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/" xmlns:app="http://apply.grants.gov/services/ApplicantWebServices-V2.0" xmlns:gran="http://apply.grants.gov/system/GrantsCommonElements-V1.0" xmlns:app1="http://apply.grants.gov/system/ApplicantCommonElements-V1.0">
   <soapenv:Header/>
   <soapenv:Body>
      <app:GetOpportunityListRequest>
            <app1:OpportunityFilter>
            <gran:FundingOpportunityNumber>O-BJA-2025-202930-STG</gran:FundingOpportunityNumber>
         </app1:OpportunityFilter>
      </app:GetOpportunityListRequest>
   </soapenv:Body>
</soapenv:Envelope>'
```
Ensure you see the following logs:
`soap_api_diff complete                             soap_api_diff={'keys_only_in_sgg': {'OpportunityDetails': 'hidden'}} app.name=src.app app_name=api run_mode=service environment=local deploy_whoami=local-developer request.id= request.method=POST request.path=/grantsws-applicant/services/v2/ApplicantWebServicesSoapPort request.url_rule=/<service_name>/services/v2/<service_port_name> request.internal_id=3a6793c2-06a7-4cea-bcb7-fc0246588edb soap_api=applicants soap_request_operation_name=GetOpportunityListRequest soap_response_operation=GetOpportunityListResponse`

Note: You may see log about opening and closing date for the proxy response. That will be addressed [here](https://github.com/HHS/simpler-grants-gov/issues/5744)

### DIFF
--- a/api/src/legacy_soap_api/legacy_soap_api_client.py
+++ b/api/src/legacy_soap_api/legacy_soap_api_client.py
@@ -114,7 +114,27 @@ class BaseSOAPClient:
         """
         proxy_response_soap_dict = self.get_proxy_soap_response_dict(proxy_response)
         simpler_response_soap_dict = self.get_soap_response_dict()
-        diff_soap_dicts(simpler_response_soap_dict, proxy_response_soap_dict)
+
+        try:
+            diff_results = diff_soap_dicts(
+                sgg_dict=get_envelope_dict(
+                    simpler_response_soap_dict, self.operation_config.response_operation_name
+                ),
+                gg_dict=get_envelope_dict(
+                    proxy_response_soap_dict, self.operation_config.response_operation_name
+                ),
+                key_indexes=self.operation_config.key_indexes,
+                keys_only=True,
+            )
+            logger.info("soap_api_diff complete", extra={"soap_api_diff": diff_results})
+        except Exception as e:
+            logger.info(
+                "soap_api_diff incomplete",
+                {
+                    "soap_api_error": "soap_response_diff",
+                    "soap_traceback": "".join(traceback.format_tb(e.__traceback__)),
+                },
+            )
 
         try:
             simpler_soap_response_payload = SOAPPayload(
--- a/api/src/legacy_soap_api/legacy_soap_api_utils.py
+++ b/api/src/legacy_soap_api/legacy_soap_api_utils.py
@@ -8,6 +8,7 @@ from src.legacy_soap_api.legacy_soap_api_schemas import FaultMessage, SOAPRespon
 BASE_SOAP_API_RESPONSE_HEADERS = {
     "Content-Type": 'multipart/related; type="application/xop+xml"',
 }
+HIDDEN_VALUE = "hidden"
 
 
 def format_local_soap_response(response_data: bytes) -> bytes:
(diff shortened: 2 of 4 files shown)


api/src/legacy_soap_api/legacy_soap_api_client.py
```diff
@@ -114,7 +114,27 @@ def get_simpler_soap_response(self, proxy_response: SOAPResponse) -> tuple:
...
+                sgg_dict=get_envelope_dict(
+                    simpler_response_soap_dict, self.operation_config.response_operation_name
+                ),
+                gg_dict=get_envelope_dict(
+                    proxy_response_soap_dict, self.operation_config.response_operation_name
+                ),
+                key_indexes=self.operation_config.key_indexes,
+                keys_only=True,
+            )
+            logger.info("soap_api_diff complete", extra={"soap_api_diff": diff_results})
```
chouinar: I don't think I see it in what is returned, but could we make sure there is a field in the diff that just says whether it matches or not as a boolean for the whole model?

Just so we have something for a simple graph of percentage of records that match for a given endpoint that doesn't need to check things like whether the results themselves are empty or not.

api/src/legacy_soap_api/legacy_soap_api_utils.py
```diff
@@ -125,6 +126,85 @@ def get_invalid_path_response() -> SOAPResponse:
...
+    only_in_gg = {k: gg_dict[k] for k in gg_dict_keys - sgg_dict_keys}
+    if keys_only:
+        return {
+            "index_key": index_key,
+            "count_found_only_in_sgg": len(only_in_sgg.keys()),
+            "count_found_only_in_gg": len(only_in_gg.keys()),
+            "count_different_values": len(
+                [k for k in sgg_dict_keys & gg_dict_keys if sgg_dict[k] != gg_dict[k]]
+            ),
+        }
```
chouinar, line 204: For this, what if instead of counts we returned the keys themselves as a list? That way when we query it in New Relic we can see which fields are different.

freddieyebra: I started making this change and realized that this case is only for a list of dicts, that would not indicate difference in number of entries returned by each response.

freddieyebra: I did add a flag for if the objects do match before the diff.

api/src/legacy_soap_api/legacy_soap_api_client.py
```diff
@@ -114,7 +140,12 @@ def get_simpler_soap_response(self, proxy_response: SOAPResponse) -> tuple:
         """
         proxy_response_soap_dict = self.get_proxy_soap_response_dict(proxy_response)
         simpler_response_soap_dict = self.get_soap_response_dict()
-        diff_soap_dicts(simpler_response_soap_dict, proxy_response_soap_dict)
+
+        # We will only run diffs for responses that do not match.
+        if proxy_response_soap_dict == simpler_response_soap_dict:
+            logger.info("soap_api_diff responses match", extra={"soap_responses_match": True})
+        else:
+            self.log_diffs(proxy_response_soap_dict, simpler_response_soap_dict)
```
chouinar, line 148: If the response has lists in it that happen to be sorted differently, wouldn't the first if-statement here potentially contain fewer matches than it should?

Assume the responses were `[1, 2 3]` and `[2, 3, 1]` and the sorting order didn't matter, we would say they didn't match.

I think this is a reasonable shortcut to keep, but should we check in `log_diffs` if any diff results came back, and if not, set the `soap_responses_match` to True in that case?

freddieyebra: Good catch. Just updated this.

Issue #4674, opened 2025-04-21
Email schedule sender


chrisw-nava:
### Summary

TBD

### Acceptance criteria

TBD

eric-serrato95:
Matt D will be our schedule sender until this is implemented, 
@mdragon to write up command to make the sender run manually

PR #5751, opened 2025-07-22
[Issue #4674] Email schedule sender

babebe:
Fixes https://github.com/HHS/simpler-grants-gov/issues/4674
Scheduler for email notifications. 
Disable reset in all environments


infra/api/app-config/env-config/scheduled_jobs.tf
```diff
@@ -73,5 +73,11 @@ locals {
       schedule_expression = "cron(0 13 * * ? *)"
       state               = "ENABLED"
     }
+    email_notification_opportunity = {
+      task_command = ["poetry", "run", "flask", "task", "email-notifications"]
+      # Every day at 11:40am Eastern Time during DST. 12:40pm during non-DST.
+      schedule_expression = "cron(40 15 * * ? *)"
+      state               = "ENABLED"
+    }
```
chouinar: This would run at 10:40 / 11:40am I believe?

The ticket mentions something about weekends, did we want to exclude weekends?

babebe: thats right. updated. And it's saying don't handle creating exceptions for the weekend

chouinar (approved):

Issue #5724, opened 2025-07-18
USWDS FileUpload component broken in dev mode


doug-s-nava:
### Summary

When running the app in dev mode (`npm run dev`) rather than with a production build, the USWDS FileUpload component (see https://github.com/trussworks/react-uswds/blob/c18d70efec1cad0fafa8f9205e10fc84a387b211/src/components/forms/FileInput/FileInput.tsx) will error when uploading a file.

This is due to a bug in the FilePreview component used under the hood, which uses a one time on-mount useEffect to read in a file. This useEffect runs twice in dev mode, and thus the error. A solution for this bug has been proposed here: https://github.com/trussworks/react-uswds/issues/2433

Our job is to create a PR against USWDS with the fix and get it merged.

Note that this is not a bug that will affect production or other deployed environments and can be worked around locally by developing with a production build.

### Reproducibility

I tried multiple times and saw the bug every time

### How to reproduce the bug

* start a local server with `npm run dev`
* go to an application and attempt to upload an attachment

### Browser

None

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

doug-s-nava:
Other Nava folks have a PR up with a fix for this within the platform project, but not a comprehensive fix https://github.com/navapbc/template-application-nextjs/pull/406

When this fix is merged into the trussworks library, an extra step for this ticket is to open a PR with the nava template to bump the trussworks version

doug-s-nava:
https://github.com/trussworks/react-uswds/pull/3236

PR #5756, opened 2025-07-23
[Issue 5128]: Attachment Card

myduong-navapbc:
## Summary

Work for #ISSUE 5128

## Changes proposed

This PR contains the Attachments Card Feature
- It has the USWDS File Input
- Contains a table that shows the attachments that the user has uploaded
- - Ability to sort the Attachment Table
- Actions to Delete Attachments
- Shows temp line item when an item is uploading or deleting with spinner

## Validation steps

- For local testing: you cannot run `npm run dev` strictmode throws an error when trying to attach a file to upload. 
- https://github.com/HHS/simpler-grants-gov/issues/5724
- 
- to run locally, you would need to run `npm run build && npm start`
- make remake-backend && docker compose up
- navigate to an opportunity that has competition data associated with it
- start application
- scroll to the bottom where you can attach files


frontend/src/app/api/applications/[applicationId]/attachments/[applicationAttachmentId]/handler.ts
```diff
@@ -0,0 +1,53 @@
...
+    const response = await deleteAttachment(
+      applicationId,
+      applicationAttachmentId,
+      session.token,
+    );
+    const status = response.status_code;
+
+    revalidateTag(`application-${applicationId}`);
+
+    if (!response || (status !== 200 && status !== 422)) {
```
Comment (doug-s-nava) [inline frontend/src/app/api/applications/[applicationId]/attachments/[applicationAttachmentId]/handler.ts]: what is the situation where we are getting a 422 back from the delete endpoint?

frontend/src/components/application/attachments/attachmentUtils.tsx
```diff
@@ -0,0 +1,49 @@
+import { Attachment, AttachmentCardItem } from "src/types/attachmentTypes";
+import { v4 as uuidv4 } from "uuid";
+
+export type SortKey = "file_name" | "file_size_bytes" | "updated_at";
```
doug-s-nava: can these be moved to a types file?

frontend/src/components/TableHeader.tsx
```diff
@@ -0,0 +1,35 @@
+import { ReactElement } from "react";
+
+import { SortDirection } from "src/components/application/attachments/attachmentUtils";
+
+interface TableHeaderProps<SortKey extends string> {
```
doug-s-nava: I think I'm missing something with the generics here - is this to future proof for sortable headers that have different sets of possible sort keys?

frontend/src/components/application/attachments/AttachmentsCard.tsx
```diff
@@ -0,0 +1,258 @@
...
+interface AttachmentsCardProps {
+  applicationId: string;
+  attachments: Attachment[];
+}
+
+export const AttachmentsCard = ({
+  applicationId,
+  attachments,
+}: AttachmentsCardProps) => {
+  /**
```
doug-s-nava, line 42: these seem like constants, can we define them outside the component?

frontend/src/components/application/attachments/AttachmentsCard.tsx
```diff
@@ -0,0 +1,258 @@
...
+  const deleteModalRef = useRef<ModalRef | null>(null);
+  const fileInputRef = useRef<FileInputRef | null>(null);
+  const lastDeletedIdRef = useRef<string | null>(null);
+
+  /**
+   * Local state
+   */
+
+  const [uploads, setUploads] = useState<AttachmentCardItem[]>([]);
+  const [sortBy, setSortBy] = useState<SortKey>("updated_at");
```
doug-s-nava: everything related to sorting could be moved into the AttachmentsCardTable

frontend/src/components/application/attachments/AttachmentsCard.tsx
```diff
@@ -0,0 +1,258 @@
...
+    startTransition(() => {
+      formAction({ formData, tempId, abortController });
+    });
+  };
+
+  /**
+   * useActionStates
+   */
+
+  const [state, formAction] = useActionState(
```
doug-s-nava:
```suggestion
  const [uploadState, uploadFormAction] = useActionState(
```

frontend/src/components/application/attachments/AttachmentsCard.tsx
```diff
@@ -0,0 +1,258 @@
...
+      deleteModalRef.current?.toggleModal();
+    }
+  }, [deleteState]);
+
+  // Upload State
+  useEffect(() => {
+    if (state?.success) {
+      fileInputRef.current?.clearFiles();
+      setUploads((prev) => prev.filter((u) => u.id !== state?.uploads.tempId));
+      setFileInputErrorText(state?.error);
```
doug-s-nava: is this necessary? seems like the same condition would be hit on 210, unless this is meant to clear errors? if that's the case I'd make it more explicit

alternatively could this be simplified like
```
 fileInputRef.current?.clearFiles();
setFileInputErrorText(state.error);
if (state.success) {
  do upload thing
}
```

frontend/src/components/application/attachments/AttachmentsCardForm.tsx
```diff
@@ -0,0 +1,39 @@
...
+  inputRef,
+}: Props) => {
+  return (
+    <>
+      <input type="hidden" name="application_id" value={applicationId} />
+      <FileInput
+        id="application-attachment-upload"
+        name="application-attachment-upload"
+        errorText={errorText}
+        accept=".pdf"
```
doug-s-nava: we're cool with only accepting pdfs for now even though the directions mention other file types?

frontend/src/components/application/attachments/AttachmentsCardTableHeaders.tsx
```diff
@@ -0,0 +1,55 @@
...
+  return (
+    <thead>
+      <tr>
+        <TableHeader<SortKey>
+          isSortable
+          sortKey="file_name"
+          value={t("attachedDocument")}
+          currentSortKey={sortBy}
+          currentSortDirection={sortDirection}
+          onSort={handleAttachmentSort}
```
doug-s-nava, line 31: i wonder if we could work around the complexity of sort key typing by removing the idea entirely, and just leading the onSort hander do all the work - if we fold the idea of the key into the function (since it's a static value) we wouldn't have to track the key separately...

```
const getSortFnForKey = (key) => sortFn
...
onSort={(event) => getSortFnForKey(key)(event)}
```

frontend/src/components/application/attachments/DeleteAttachmentButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+}: AttachmentDeleteButtonProps) => {
+  return (
+    <ModalToggleButton
+      disabled={false}
+      modalRef={modalRef}
+      opener
+      className="usa-nav__link font-sans-2xs display-flex text-normal border-0"
+      data-testid="delete-attachment-modal-toggle-button"
+      onClick={() =>
+        handleDeleteAttachment(file.application_attachment_id, file.file_name)
```
doug-s-nava: this is a little misleading - seems like the purpose of this function is to mark a file for deletion within the list rather than to actually delete the attachment. If I'm understanding right, can we rename?

frontend/src/components/application/attachments/DeleteAttachmentModal.tsx
```diff
@@ -0,0 +1,81 @@
...
+  ModalFooter,
+  ModalRef,
+  ModalToggleButton,
+} from "@trussworks/react-uswds";
+
+import { SimplerModal } from "src/components/SimplerModal";
+import Spinner from "src/components/Spinner";
+
+interface Props {
+  buttonCtaText: string;
```
doug-s-nava: since this is already specific to deleting attachments do we need these values to be flexible? can we cut down on the props by just getting these strings within the component? looks like we could cut this down to 4 props

frontend/src/components/application/attachments/DeleteAttachmentModal.tsx
```diff
@@ -0,0 +1,81 @@
...
+}: Props) => {
+  const t = useTranslations("Application.attachments");
+  return (
+    <SimplerModal
+      modalId={modalId}
+      modalRef={modalRef}
+      titleText={
+        pendingDeleteName
+          ? `${titleText} ${pendingDeleteName}?`
+          : "Caution, deleting attachment"
```
doug-s-nava: should move to `messages`

frontend/src/components/application/attachments/actions.ts
```diff
@@ -0,0 +1,134 @@
...
+  abortController: AbortController;
+}
+
+export const uploadAttachmentAction = async (
+  _prevState: UploadAttachmentActionState | undefined,
+  {
+    formData,
+    tempId,
+    abortController,
+  }: {
```
doug-s-nava: seems the same as `UploadAttachmentAction`

frontend/src/components/application/attachments/actions.ts
```diff
@@ -0,0 +1,134 @@
...
+    formData: FormData;
+    tempId: string;
+    abortController: AbortController;
+  },
+): Promise<UploadAttachmentActionState> => {
+  const session = await getSession();
+
+  if (!session || !session.token) {
```
doug-s-nava, line 47: might be worth building out a wrapper to do with `useClientFetch` is doing but for server actions, so that we don't have this boilerplate, and we can be bumping session tokens etc.

frontend/src/services/fetch/fetchers/applicationFetcher.ts
```diff
@@ -81,6 +74,31 @@ export const getApplicationDetails = async (
   return (await response.json()) as ApplicationDetailApiResponse;
 };
 
+export const updateApplicationFilingName = async (
```
doug-s-nava, line 77: is this supposed to be in this pr or did it leak from somewhere else

frontend/src/utils/fileUtils/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./createFormData";
```
doug-s-nava: we're not using this type of pattern at this point. For now can you reference files directly and we can discuss a change?

doug-s-nava:
for later - can we catch the error that happens in strict mode (related to the FileInput bug called out elsewhere) so that it doesn't break the app, and add a useful error message like ("run this in a production build")

see https://github.com/HHS/simpler-grants-gov/issues/5724

frontend/src/services/fetch/fetchers/applicationFetcher.ts
myduong-navapbc: updateApplicationFilingName is in main now, so i'm not sure why it thinks that it is new. perhaps adding comment blocks did something. or it was reordered in this branch

doug-s-nava (commented):
a couple things worth calling out:

* uploading over 1mb files apparently doesn't work, I get this error message
```
[Error: Body exceeded 1 MB limit.
                To configure the body size limit for Server Actions, see: https://nextjs.org/docs/app/api-reference/next-config-js/serverActions#bodysizelimit] {
  statusCode: 413
}
 ⨯ uncaughtException:  [Error: Body exceeded 1 MB limit.
                To configure the body size limit for Server Actions, see: https://nextjs.org/docs/app/api-reference/next-config-js/serverActions#bodysizelimit] {
  statusCode: 413
}
```
* uploading an unsupported file type still allows the upload, but displays an error in the input. Is it worth limiting the file types at all if the uploads of all file types actually works fine?
*

frontend/src/components/application/attachments/actions.ts
myduong-navapbc: i will address this comment in https://github.com/HHS/simpler-grants-gov/issues/5724

frontend/src/components/application/attachments/AttachmentsCardTableHeaders.tsx
myduong-navapbc: I will address this in https://github.com/HHS/simpler-grants-gov/issues/5724

frontend/src/components/TableHeader.tsx
myduong-navapbc: I removed the generics from this component/props as the app does have anything like it.

But to answer the question:
It was to help scale or future proof the header component for other uses ie
`<TableHeader<UserSortKey> ...>, <TableHeader<OpportunitySortKey> ...>`

but essentially, to keep the type safety and allows the compiler to only us valid keys for the tables using the headers.

something like:
`<TableHeader<UserSortKey> onSort={(key) => doSomethingWithUserSort(key)} /> `

myduong-navapbc:
> a couple things worth calling out:
> 
> * uploading over 1mb files apparently doesn't work, I get this error message
> 
> ```
> [Error: Body exceeded 1 MB limit.
>                 To configure the body size limit for Server Actions, see: https://nextjs.org/docs/app/api-reference/next-config-js/serverActions#bodysizelimit] {
>   statusCode: 413
> }
>  ⨯ uncaughtException:  [Error: Body exceeded 1 MB limit.
>                 To configure the body size limit for Server Actions, see: https://nextjs.org/docs/app/api-reference/next-config-js/serverActions#bodysizelimit] {
>   statusCode: 413
> }
> ```
> 
> * uploading an unsupported file type still allows the upload, but displays an error in the input. Is it worth limiting the file types at all if the uploads of all file types actually works fine?

This, I've updated the config file to allow for big files.

doug-s-nava (approved):
two notes on behavior, neither of which are blocking:
* sorting by attachment title seems a little off. In my testing "Internal Wiki.pdf" is sorted before "dashboard-export-11-37-pm-2025-06-09.pdf", when I would expect the opposite based on alphabetical sorting
* I'm not sure that upload cancellation works. I tried uploading a large file and canceling. On clicking cancel the upload was removed from the list, but then it popped in later, likely when the upload completed. Maybe a timing issue

let's get this merged!!!! 

✅ ✅ ✅ ✅

Issue #5752, opened 2025-07-22
Change agency search schedule to populate data hourly


mdragon:
### Summary

Currently the load-search-agency-data job is running daily at about 9am, but we really want it to run every hour, after the ELT jobs run.

### Acceptance criteria

Agency to Search index job is running hourly instead of daily.

PR #5757, opened 2025-07-23
[Issue #5752] Change agency search schedule to populate data hourly (#5752)

prasnava:
## Summary

Work for #5752

## Changes proposed

Updated the cron schedule for load-search-agency-data to run every 1 hour.

## Context for reviewers

Currently the load-search-agency-data job is running daily at about 9am, but we want it to run every hour

## Validation steps

- [ ] The dev environment "api-dev-load-search-agency-data" - event bridge schedule cron should be scheduled to run hourly

### DIFF
--- a/api/tests/src/adapters/search/test_opensearch_query_builder.py
+++ b/api/tests/src/adapters/search/test_opensearch_query_builder.py
@@ -1,12 +1,12 @@
 import uuid
-from datetime import date
+from datetime import date, timedelta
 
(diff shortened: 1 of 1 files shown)


chouinar (approved):
LGTM - We'll want to update the alert while we're at it which errors if the job hasn't run in the past day to be 3 hours like the other hourly jobs: 
https://onenr.io/0qwLDDXerw5

Issue #5739, opened 2025-07-21
Create ADR on virus scanning solution for Simpler Application Attachments


chris-kuryak:
### Summary

We want to write an ADR to determine how we are going to scan files uploaded by users for viruses.

Ideally the approach should be able to scan a file as it is uploaded (file is in memory) before going to s3. While approaches exist for scanning at rest in s3, any such approach would need additional logic to somehow update the record in our DB otherwise we would have a pointer from our DB table to a bad/deleted s3 file. This async approach would also need a way to alert a user that the file was bad - we shouldn't fully rule out these approaches BUT if we go with any at-rest approach the approach needs to describe how we'll handle this.

### Acceptance criteria

- [ ] ADR created include historical decisions, current solution, proposed solution, and justification

PR #5761, opened 2025-07-23
[Issue #5739] Create ADR on virus scanning solution for Simpler Application Attachments

mikehgrantsgov:
## Summary

Work for #5739

## Changes proposed

Add ADR for virus scanning solution

## Context for reviewers

Explore ways of implementing antivirus for application file attachments.

## Validation steps

Confirm ADR satisfies all approaches we would want to tackle for this effort.

### DIFF
--- a/documentation/wiki/product/decisions/adr/2025-07-23-file-upload-virus-scanning.md
+++ b/documentation/wiki/product/decisions/adr/2025-07-23-file-upload-virus-scanning.md
@@ -1,5 +1,7 @@
 # File Upload Virus Scanning Architecture
 
+- **Status:** Active
+- **Last Modified:** 2025-07-23
 * **Related Issue:** [#5686](https://github.com/HHS/simpler-grants-gov/pull/5686)
 
 ## Context and Problem Statement
--- a/documentation/wiki/product/decisions/adr/2025-07-23-file-upload-virus-scanning.md
+++ b/documentation/wiki/product/decisions/adr/2025-07-23-file-upload-virus-scanning.md
@@ -1,0 +1,146 @@
+# File Upload Virus Scanning Architecture
+
+* **Related Issue:** [#5686](https://github.com/HHS/simpler-grants-gov/pull/5686)
+
+## Context and Problem Statement
+
+The application allows users to upload file attachments as part of their applications. Currently, uploaded files are stored directly in S3 without any virus scanning, creating a security vulnerability where malicious files could be stored and potentially downloaded by other users or administrators.
+
+To ensure the security and integrity of the platform, we need to implement virus scanning for all uploaded files before they are stored and made available for download.
+
+_What is the best architectural approach to implement virus scanning for file uploads?_
+
+## Options Considered
+
+* **In-App Server ClamAV Scanning** - Run ClamAV daemon directly in the Flask application container
+* **Asynchronous S3 Event-Driven Scanning** - Upload to staging bucket, scan via S3 events, move to final bucket
+* **Third-party SaaS Solution** - Use services like VirusTotal API or AWS GuardDuty Malware Protection
+
+## Decision Recommendation
+
+Recommend **In-App Server ClamAV Scanning** by running the ClamAV daemon directly within our existing Flask application containers.
+
+### Architecture Components
+
+1. **ClamAV Daemon** - Runs as a background process in each Flask app container
+2. **Enhanced Upload Endpoint** - Modified `/applications/{id}/attachments` endpoint with virus scanning
+3. **Python ClamAV Client** - Uses `clamd` library to communicate with local ClamAV daemon
+4. **Definition Updates** - Scheduled `freshclam` updates within containers every 6 hours
+5. **Error Handling** - Return immediate 422 errors for infected files
+
+### Implementation Approach
+
+```python
+# Add to app initialization
+import clamd
+
+# Initialize ClamAV connection (once at app startup)
+clam = clamd.ClamdUnixSocket()
+
+# Modified create_application_attachment service
+def create_application_attachment(db_session, application_id, user, form_and_files_data):
+    file_attachment = form_and_files_data.get("file_attachment")
+    
+    # Pre-scan validation
+    if file_attachment.content_length > MAX_SCAN_SIZE:
+        raise_flask_error(422, "File too large for virus scanning")
+    
+    # Perform virus scan using local ClamAV daemon
+    scan_result = clam.instream(file_attachment.stream)
+    
+    if scan_result['stream'][0] == 'FOUND':
+        threat_name = scan_result['stream'][1]
+        logger.warning(f"Malware detected in upload: {threat_name}")
+        raise_flask_error(422, f"File rejected: malware detected ({threat_name})")
+    
+    if scan_result['stream'][0] == 'ERROR':
+        logger.error(f"Virus scan failed: {scan_result['stream'][1]}")
+        raise_flask_error(500, "File upload temporarily unavailable")
+    
+    # Proceed with S3 upload only if clean
+    s3_key = file_util.open_stream(file_attachment)
+    # ... rest of attachment creation logic
+```
+
+### Docker Configuration
+
+```dockerfile
+# Add to existing Dockerfile
+RUN apt-get update && apt-get install -y \
+    clamav \
+    clamav-daemon \
+    clamav-freshclam \
+    && rm -rf /var/lib/apt/lists/*
+
+# Update virus definitions
+RUN freshclam
+
+# Start ClamAV daemon
+RUN mkdir -p /var/run/clamav && \
+    chown clamav:clamav /var/run/clamav
+    
+# Add to container startup script
+CMD ["sh", "-c", "clamd --config-file=/etc/clamav/clamd.conf & python app.py"]
+```
+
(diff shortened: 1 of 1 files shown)


documentation/wiki/product/decisions/adr/2025-07-23-file-upload-virus-scanning.md
```diff
@@ -0,0 +1,148 @@
...
+# Initialize ClamAV connection (once at app startup)
+clam = clamd.ClamdUnixSocket()
+
+# Modified create_application_attachment service
+def create_application_attachment(db_session, application_id, user, form_and_files_data):
+    file_attachment = form_and_files_data.get("file_attachment")
+    
+    # Pre-scan validation
+    if file_attachment.content_length > MAX_SCAN_SIZE:
+        raise_flask_error(422, "File too large for virus scanning")
```
chouinar: How large of files can clam AV support?

documentation/wiki/product/decisions/adr/2025-07-23-file-upload-virus-scanning.md
```diff
@@ -0,0 +1,148 @@
...
+    
+    # Perform virus scan using local ClamAV daemon
+    scan_result = clam.instream(file_attachment.stream)
+    
+    if scan_result['stream'][0] == 'FOUND':
+        threat_name = scan_result['stream'][1]
+        logger.warning(f"Malware detected in upload: {threat_name}")
+        raise_flask_error(422, f"File rejected: malware detected ({threat_name})")
```
chouinar: Any idea what the performance is for scanning files? How long does it take for something small (a few KB), a mildly complex PDF (maybe a few MB), or just a full video (1gb)?

documentation/wiki/product/decisions/adr/2025-07-23-file-upload-virus-scanning.md
```diff
@@ -0,0 +1,148 @@
...
+# Update virus definitions
+RUN freshclam
+
+# Start ClamAV daemon
+RUN mkdir -p /var/run/clamav && \
+    chown clamav:clamav /var/run/clamav
+    
+# Add to container startup script
+CMD ["sh", "-c", "clamd --config-file=/etc/clamav/clamd.conf & python app.py"]
+```
```
chouinar: Is there a way we could only have clam AV running when we run as the API (some env var we could set?). We wouldn't need virus scanning on backend tasks since it doesn't interact with live user data and adding more RAM to support something we aren't going to call seems wasteful.

Alternatively, is there any way to run clamav as a sidecar in its own container so it can get its own resources, and then maybe when we startup a task we say whether we want that sidecar (if that's possible?).

documentation/wiki/product/decisions/adr/2025-07-23-file-upload-virus-scanning.md
```diff
@@ -0,0 +1,148 @@
...
+* **User experience:** Instant feedback on file uploads (clean or rejected)
+* **Simple architecture:** No additional AWS services or infrastructure
+* **Lower latency:** No network calls to external scanning services (~100-500ms faster than Lambda)
+* **Cost-effective:** No Lambda execution costs, uses existing container resources
+* **No cold starts:** ClamAV daemon stays warm between requests
+* **Easy testing:** Can run and test locally during development
+
+### Negative Consequences
+
+* **Container resource usage:** Each app container needs additional memory (~200MB) for ClamAV
```
chouinar: I'm seeing 3-4gb in their docs: https://docs.clamav.net/manual/Installing/Docker.html

documentation/wiki/product/decisions/adr/2025-07-23-file-upload-virus-scanning.md
mikehgrantsgov: Add some capacity updates

documentation/wiki/product/decisions/adr/2025-07-23-file-upload-virus-scanning.md
mikehgrantsgov: Also added some performance benchmarks based on my findings

documentation/wiki/product/decisions/adr/2025-07-23-file-upload-virus-scanning.md
mikehgrantsgov: There is a way to run this as a sidecar, and added a new approach to the docs

documentation/wiki/product/decisions/adr/2025-07-23-file-upload-virus-scanning.md
mikehgrantsgov: Ah yeah, updated this

chouinar:
@mikehgrantsgov - could you add chatting about this to the engineering sync agenda for this afternoon? Just to get a few more eyes on it.

mikehgrantsgov:
@chouinar I thihk we landed on creating some tickets to explore and benchmark the sidecar performance / memory footprint, then take it from there is an asynchronous method is needed for our load requirements.

Does anything else need to be added to the ADR to capture this?

documentation/wiki/product/decisions/adr/2025-07-23-file-upload-virus-scanning.md
```diff
@@ -0,0 +1,277 @@
...
+**File Size Support:**
+* **ClamAV Maximum:** 4GB (configurable via MaxFileSize)
+* **Practical Limit:** 1GB for user experience
+* **Files exceeding limit:** Return clear error message suggesting file compression or alternative delivery
+
+**Performance Benchmarks:**
+* **Small files (few KB):** ~50-100ms scanning time
+* **Medium files (1-10MB):** ~200-500ms scanning time  
+* **Large files (100MB-1GB):** ~2-15 seconds scanning time
+* **Memory usage:** 3-4GB RAM for ClamAV daemon (virus database + processing)
```
chouinar, line 144: Any idea how much memory the API itself had to use while testing this? Not sure if there is a good way to see current memory used by a docker container.

Also worth calling out that the performance times are likely a bit higher as uploading a file to a local s3 will be faster than s3 when running in AWS

documentation/wiki/product/decisions/adr/2025-07-23-file-upload-virus-scanning.md
```diff
@@ -0,0 +1,277 @@
...
+
+## Pros and Cons of the Options
+
+### ClamAV Sidecar Container ✅ **CHOSEN**
+
+* **Pros**
+  * Immediate feedback to users
+  * Malicious files never reach permanent storage
+  * Resource isolation - dedicated 4GB for ClamAV
+  * Selective deployment (API only, not backend tasks)
```
chouinar, line 174: We'll need to check if we actually have support for that, I think right now sidecars were setup so they're on both the API and backend tasks since most of our infra doesn't even differentiate the two as being distinct. I'm sure it's doable, but will be a bit of infra most likely.

documentation/wiki/product/decisions/adr/2025-07-23-file-upload-virus-scanning.md
```diff
@@ -0,0 +1,277 @@
...
+      - CLAMAV_NO_FRESHCLAMD=false
+    memory: 4096m  # 4GB as recommended by ClamAV docs
+    networks:
+      - app-network
+    healthcheck:
+      test: ["CMD", "clamdscan", "--ping"]
+      interval: 30s
+      timeout: 10s
+      retries: 3
+
```
chouinar, line 114: Just thinking if we set this up, I'd probably err on the side of not by default having your local docker-compose enable clamav just to avoid needing to give it 4gb for something we don't need most of the time. Not something we need in these docs, but if we make the follow up tickets, probably call it out as a separate extended docker-compose we can optionally use

prasnava:
> @mdragon / @prasnava Do you have any thoughts after our discussion the other day?

I'm thinking should we take this endpoint out of the api container and have a separate ecs service/tasks scheduled on ECS? So that way we can implement custom autoscaling by the number of concurrent uploads? 

We can use path based routing at the load balancer to route for the uploads endpoint.

documentation/wiki/SUMMARY.md
```diff
@@ -89,6 +89,7 @@
     * [Cross-Program Team Health Survey Tool](product/decisions/adr/2024-12-06-team-health-survey-tool.md)
     * [Adding Slack Users to SimplerGrants Slack Workspace](product/decisions/adr/2024-12-17-adding-slack-users.md)
     * [Repo organization](product/decisions/adr/2025-01-02-repo-organization.md)
+    * [File Upload Virus Scanning Architecture](product/decisions/adr/2025-07-23-file-upload-virus-scanning.md)
```
mdragon, line 92: Can you put this last so they're in roughly chronological order?

mdragon:
> > @mdragon / @prasnava Do you have any thoughts after our discussion the other day?
> 
> We can use path based routing at the load balancer to route for the uploads endpoint.

@prasnava I think we can consider that for beyond the pilot, but even in the medium term we're probably going to have fewer than 5 concurrent uploads at any given time. So I would like to have a ticket representing this idea so we don't lose it, but hold off on implementing that until we get a better sense of if we have a level of load that justifies it.

Issue #5759, opened 2025-07-23
Add the ability to save an opportunity from the Search Results Table


btabaska:
### Summary

Currently to save an opportunity you need to click into the opportunity from a search result. This task is to implement the front-end feature to save an opportunity directly from the search results table. 

Proposed Design
https://github.com/HHS/simpler-grants-gov/issues/5544

### Acceptance criteria

- [ ] You can save an opportunity from the search results table
- [ ] Unit Tests updated

PR #5762, opened 2025-07-24
[Issue #5759] Add ability to save opportunities from search results table

btabaska:
## Summary

Work for #5759 

## Changes proposed

Added ability to opportunities from search results table. 

## Context for reviewers

Had to manually implement CSS classes to change how button was displayed in this context without impacting how the button is used on the opportunity page. 

## Validation steps
Screenshot 2025-07-30 at 4 06 34 PM

### DIFF
--- a/frontend/src/components/search/SearchResultsTable.tsx
+++ b/frontend/src/components/search/SearchResultsTable.tsx
@@ -117,13 +117,7 @@ const toSearchResultsTableRow = (
       stackOrder: 1,
     },
     {
-      cellData: (
-        <TitleDisplay
-          opportunity={result}
-          page={page}
-          index={index}
-        />
-      ),
+      cellData: <TitleDisplay opportunity={result} page={page} index={index} />,
       stackOrder: 0,
     },
     {
@@ -167,11 +161,7 @@ export const SearchResultsTable = ({
     { cellData: t("headings.awardMax") },
   ];
   const tableRowData = searchResults.map((result, index) =>
-    toSearchResultsTableRow(
-      result,
-      page,
-      index,
-    ),
+    toSearchResultsTableRow(result, page, index),
   );
   return (
     <TableWithResponsiveHeader
--- a/frontend/src/components/search/SearchResultsTable.tsx
+++ b/frontend/src/components/search/SearchResultsTable.tsx
@@ -72,7 +72,7 @@ const TitleDisplay = ({
         <span className="text-bold">{t("number")}:</span>{" "}
         {opportunity.opportunity_number}
       </div>
-      <div className="margin-top-2">
+      <div className="margin-top-2 search-table-save-wrapper">
         <OpportunitySaveUserControl
           opportunityId={opportunity.opportunity_id}
         />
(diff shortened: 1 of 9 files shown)


btabaska:
@andycochran Created a PR with the work so far if you wanted to review and recommend alternatives for the button.

andycochran:
A few thoughts: 
* We're putting a lot besides the title in the "Title" column, which is making the rows taller so that fewer results are on screen at once. Not sure this button makes sense in any of the other columns. Wondering if there's a better layout we might use. 
* There are multiple patterns for indicating save/saved. On the saved opportunities page, there's a smaller element. It's not clickable, but could be. Then we could follow the same pattern on the search results as on the saved opps page. It's kinda confusing that the table rows are _almost_ the same in these two views. 
* I don't think we need the inline alert that tells user that the opp has been saved. I think the button state changing from "[empty] save" to "[filled] saved" is enough visual feedback for users. And the alert adds even more content to the "Title" column. 

cc @crystabelrangel maybe we can discuss this at Monday's Design Sync?

btabaska:
Updated with final design

btabaska:
@andycochran Ready for code review

frontend/src/components/search/SearchResultsTable.tsx
```diff
@@ -51,41 +50,38 @@ const CloseDateDisplay = ({ closeDate }: { closeDate: string }) => {
...
-      {saved && (
-        <div className="margin-top-2 display-inline-block">
-          <span className="display-flex flex-align-center font-sans-2xs">
-            <USWDSIcon
-              name="star"
-              className="text-accent-warm-dark button-icon-md padding-right-05"
-            />
-            {t("saved")}
-          </span>
+      <div className="display-flex">
```
andycochran, line 63: Does this need to be flex? Can it just be a standard grid row?

frontend/src/components/user/OpportunitySaveUserControl.tsx
```diff
@@ -93,6 +96,42 @@ export const OpportunitySaveUserControl = ({
 
   const { checkFeatureFlag } = useFeatureFlags();
   if (!checkFeatureFlag("savedOpportunitiesOn")) return null;
+
+  if (type === "icon") {
+    return (
+      <>
+        {user?.token ? (
+          <SaveIcon
```
doug-s-nava, line 104: [nit] since these returned components are very similar, I wonder if we could move the differentiation into the <SaveButton> and avoid the repetition here. Not necessary at all, just a thought

frontend/src/components/search/SearchResultsTable.tsx
```diff
@@ -169,11 +157,6 @@ export const SearchResultsTable = async ({
     return <FilterSearchNoResults useHeading={true} />;
   }
 
-  const savedOpportunities = await fetchSavedOpportunities();
```
doug-s-nava, line 172: do we lose out on any real efficiency from making these requests server side here? Probably not.

doug-s-nava (approved):
a couple of thoughts, but this works, and I love the test coverage

✅ ✅ ✅ ✅

andycochran (approved):
Looks fantastic. Great tests!

frontend/src/components/search/SearchResultsTable.tsx
btabaska: It has to be flex, tried it both ways

PR #5766, opened 2025-07-24
[Issue #4515] Add unique ID to certs in SOAP calls

### LINKED ISSUE #4515: Simpler SOAP Applicants Analytics
### Summary

We will need to setup metrics/analytics around the use of this endpoint. As part of this ticket we will need to have visibility into the following:

- Simpler SOAP API usage based on URL path.
- Which SOAP operation the payload specified. For example the `GetOpportunityList` operation takes in a payload that contains a `GetOpportunityListRequest` XML property and each other SOAP operation has a corresponding XML tag that is indicative of its SOAP operation.
- Which systems or users are utilizing the API.

### Acceptance criteria

Either logging or dashboards are implemented to enable visibility into the above metrics.

chouinar:
## Summary

Fixes #4515

## Changes proposed
Adjust how we define certs to include an ID that we can log

## Context for reviewers
We want to have an ID for each cert that isn't secret. Now the env var JSON will have an ID in it alongside the cert for logging. 

Before we can merge this, we'll need to update the env var in dev/staging/prod in parameter store, but holding on that until right before I merge to avoid breaking things / confirm this approach.

## Validation steps
Delete your existing soap-api.env file and run `make setup-soap-proxy` to generate the env var in the same format.

Testing using the example SOAP call in the cert secret from 1password you'll now see the logs have the ID value of that cert:
Screenshot 2025-07-24 at 2 30 20 PM

### DIFF
--- a/api/bin/setup-soap.sh
+++ b/api/bin/setup-soap.sh
@@ -53,7 +53,7 @@ for file in $applicantspub $applicantspk $grantorspub $grantorspk; do
   fi
 done
 
-soap_auth_content=$(printf '{"%s": "%s\n\n%s","%s": "%s\n\n%s"}' \
+soap_auth_content=$(printf '{"%s": {"id": "applicant_cert", "cert": "%s\n\n%s"},"%s": {"id": "grantor_cert", "cert": "%s\n\n%s"}}' \
 "`openssl x509 -in $applicantspub -noout -fingerprint -sha256 | sed 's/://g' | cut -d'=' -f2 | tr '[:upper:]' '[:lower:]'`" "`cat ${applicantspk}`" "`cat ${applicantspub}`" \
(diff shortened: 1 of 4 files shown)


mdragon (approved):
Looks good.

freddieyebra (approved):
looks good!

`grants-api  | 19:20:13.699  src.logging.flask_logger             _log_end_request             INFO     end request                                        response.status_code=200 response.content_length=840 response.content_type=multipart/related; type="application/xop+xml"; boundary="uuid:b7924bb7-13fb-4830-b774-1e1b1e6f45d9"; start="<root.message@cxf.apache.org>"; start-info="text/xml" response.mimetype=multipart/related response.time_ms=646.4957919997687 app.name=src.app app_name=api run_mode=service environment=local deploy_whoami=local-developer request.id= request.method=POST request.path=/grantsws-applicant/services/v2/ApplicantWebServicesSoapPort request.url_rule=/<service_name>/services/v2/<service_port_name> request.internal_id=627d254e-4417-42d7-b99f-78e215bcdfcf soap_api=applicants soap_request_operation_name=GetApplicationInfoRequest cert_id=applicant_cert`

freddieyebra:
@chouinar these changes will need to be reflected in the param store prior to merge.

Issue #5727, opened 2025-07-18
Add metrics to the submission logic


chris-kuryak:
### Summary

Before adding visualizations in New Relic (#5612), we want to add a few more metrics to the API, such as "why wasn’t the submission allowed”?

Anywhere in our application logic (submit, update, start) where we would error and prevent a user from proceeding, add some level of detail of what error was encountered to the logs as an extra object. We should have the key for this be consistent so we can easily make graphs.

For example when a user tries to start an application, we reject if they aren't a member of the organization they passed in. We should have code like:
```py
if not is_member:
    # feel free to come up with better names than this
    logger.info("Some sort of log message", extra={"submission_issue": "not_a_member_of_org"})
    raise_flask_error(...)
```
Just adding that log message with a consistent key will give us a lot of data for which scenarios get hit.

We do not need to add any detail about which endpoint was calling (already automatically attached to logs).

### Acceptance criteria

- [ ] Metrics added and ready to be queried in New Relic

Issue #5612, opened 2025-07-07
Create New Relic dashboards & alerts for the new API endpoints


chris-kuryak:
### Summary

We have created new API endpoints as part of the Simpler Apply pilot process. We need to create/update the New Relic dashboards and alerts for better monitoring of these API endpoints.

Sam.gov integration endpoints will be handled in a separate ticket: #5613 

---

We already have a [rough dashboard](https://onenr.io/0qwyMYk6Xwn) for the Competition and Application endpoints - we should:
* Make sure the queries are setup to get all of the endpoints we have
* Because we have a lot of endpoints, it might make sense to break up a few of these graphs, especially for something like error scenarios
* For each of our endpoints, any metrics we record

### Acceptance criteria

- [ ] All new API endpoints from Simpler Apply pilot
- [ ] Links to the dashboard and alerts added to this ticket for awareness

chouinar:
Updated [the dashboard](https://onenr.io/0LREm44nbRa) - no metrics in prod yet, but if you switch it to dev you'll see some from testing.

I didn't add any sort of alerts yet, our API has an alert already in place for any error logs, not sure what else we might want that that wouldn't cover?

<img width="1684" height="662" alt="Image" src="https://github.com/user-attachments/assets/995b51db-c90c-4660-a72b-edcee774a728" />

<img width="1674" height="654" alt="Image" src="https://github.com/user-attachments/assets/faa0929f-db91-4842-9027-b6d5ace2a4da" />

PR #5770, opened 2025-07-25
Mikehgrantsgov/5727 add metrics to submission logic

mikehgrantsgov:
## Summary

Work for #5727

## Changes proposed

Add `SubmissionIssue` class to capture new logging message types
Add logging in strategic places across application logic to capture reasons for specific actions, which can later go into NR.

## Context for reviewers

Before adding visualizations in New Relic (https://github.com/HHS/simpler-grants-gov/issues/5612), we want to add a few more metrics to the API, such as "why wasn’t the submission allowed”?

Anywhere in our application logic (submit, update, start) where we would error and prevent a user from proceeding, add some level of detail of what error was encountered to the logs as an extra object. We should have the key for this be consistent so we can easily make graphs.

## Validation steps

Unit tests not updated since we are just adding logging and augmenting existing logging. Visual review.

### DIFF
--- a/api/src/constants/lookup_constants.py
+++ b/api/src/constants/lookup_constants.py
@@ -148,6 +148,7 @@ class CompetitionOpenToApplicant(StrEnum):
 
 class SubmissionIssue(StrEnum):
     """Submission issue types for metrics logging"""
+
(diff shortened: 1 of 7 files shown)


chouinar (approved):
LGTM!

PR #5717, opened 2025-07-17
[Issues #5574] add budget form

acouch:
## Summary

Work for #5574

## Changes proposed

Adds the initial budget form A:

image

as well as fixes from @doug-s-nava s PR: https://github.com/HHS/simpler-grants-gov/pull/5807

as well as:

1. support for "multifield" 
2. update to TextWidget to pass on class and other info
3. smaller fixes

## Context for reviewers

Note that:

1. Section B has been stubbed out, it can be ignored for this PR
2. There are frontend validation issues that will be fixed in the next PR:
image

## Validation steps

1. run general setup tasks
4. go to [Local Pilot-equivalent Opportunity](http://localhost:3000/opportunity/ac07b63e-1a11-40c5-a6ec-549b04d9a2b0) and click: [Budget Information for Non-Construction Programs (SF-424A)](http://localhost:3000/workspace/applications/application/f7dcdb9e-e25f-4c54-9da1-7a9b5448d3da/form/1765ddb2-2169-4c5b-8dd0-652bdbae7270)
5. fill in values

Issue #5701, opened 2025-07-16
Create form SF-LLL


chris-kuryak:
### Summary

Create form SF-LLL from JSON in #5599

### Acceptance criteria

- [ ] SF-LLL form is created and working as expected

PR #5807, opened 2025-07-28
[Issue #5701] SF-LLL form

doug-s-nava:
## Summary

Work for #5701 

## Changes proposed

Adds UI implementation for the "Disclosure of Lobbying Activities (SF-LLL)" form

## Context for reviewers

This relies on some changes from https://github.com/HHS/simpler-grants-gov/pull/5717, and is based off of that branch. This PR will merge into that branch rather than main.

Other than form implementation, this PR handles:
* properly referencing nested json schema properties. HTML input ids will now reflect the nested nature of the schema, using `--` as a delimiter for nested paths.
* transforming a flat set of form data into a properly nested format. This draws on the library here, but required some customization for our use case, so copy pasted it over and made the necessary changes https://github.com/ArturKot95/FormData2Json

### Not implemented

* proper validation indicators for some nested fields. Address/city and Address/street1 are always required in "Name and Address of Reporting Entity", but will not be indicated as such in the UI. Same situation with signatureBlock/name/firstName and lastName
* proper "required" indicators for any nested or conditionally required fields
* postpopulation stuff for signature and signed date has not been implemented. Those fields should not be editable but currently look like they are

### Required fields

Since parsing the schema is a little difficult, this is is how the required fields break down for this form

```

- "federal_action_type",
- "federal_action_status",
- "report_type",
- federal dept or agency

- reporting entity / entity type
- reporting entity / applicant reporting entity / organization name
- reporting entity / applicant reporting entity / address / city
- reporting entity / applicant reporting entity / address / street 1

- lobbying registrant / first name
- lobbying registrant / last name

- individual performing service / first name
- individual performing service / last name

- signature / first name
- signature / last name

- if report type == material change
- - material change year
- - material change quarter
- - (material change) last report date

- if reporting entity / entity type == subawardee
- - reporting entity / tier
- - reporting entity / prime reporting entity / organization name
- - reporting entity / prime reporting entity / address / city
- - reporting entity / prime reporting entity / address / street 1
```

Reference for the form (PDF, DAT file) can be found here https://grants.gov/forms/form-items-description/fid/670

1. on this branch, run `make db-seed-local` from /api in order to pull in form changes to the DB
2. on this branch start a local fronted server with `npm run dev`
3. find the local "pilot type" opportunity in your local db. reference your local "competition" table, the opportunity id will match up with the competition that has more forms in the "competition_form" table
4. visit http://localhost:3000/opportunity/<opportunity id from 3>?_ff=applyFormPrototypeOff:false
5. log in
6. start a new application
7. click on the sf-LLL form in the list
8. _VERIFY_: form matches screenshot with all forms present and matching the definitions from the PDF / DAT file
9. click save
10. _VERIFY_: validation warnings returned largely match the required fields as mentioned above, with the caveats mentioned further above
11. enter "materialChange" as the "Report type" and save
12. _VERIFY_: all 3 material change fields are now failing validation
13. enter "subawardee" as "Entity Type"
14. _VERIFY: "tier" and section 4 "organization_name" are now failing validation
15. fill in all fields and save
16. _VERIFY_: no validation warnings

### Outstanding work and questions

- should state be a select?
- how to represent proper award amount and zip as strings?

### Screenshot
Screenshot 2025-07-31 at 11-57-38 Form test for form React JSON Schema

### DIFF
--- a/api/src/form_schema/forms/sflll.py
+++ b/api/src/form_schema/forms/sflll.py
@@ -134,18 +134,7 @@ FORM_JSON_SCHEMA = {
             "maxLength": 120,
         },
         "award_amount": {
-            # Represents a monetary amount. We use a string instead of number
-            # to avoid any floating point rounding issues.
-            "type": "string",
-            # Pattern here effectively says:
-            # * Any number of digits
-            # * An optional decimal point
-            # * Then exactly 2 digits - if there was a decimal
-            "pattern": r"^\d*([.]\d{2})?$",
-            # Limit the max amount based on the length (11-digits, allows up to 99 billion)
-            "maxLength": 14,
-            "title": "Award Amount",
-            "description": "For a covered Federal action where there has been an award or loan commitment by the Federal agency, enter the Federal amount of the award/loan commitment of the prime entity identified in item 4 or 5.",
+            "$ref": "#/$defs/budget_monetary_amount",
         },
         "lobbying_registrant": {
             "type": "object",
@@ -321,6 +310,20 @@ FORM_JSON_SCHEMA = {
                     "maxLength": 6,
                 },
             },
+        },
+        "budget_monetary_amount": {
+            # Represents a monetary amount. We use a string instead of number
+            # to avoid any floating point rounding issues.
+            "type": "string",
+            # Pattern here effectively says:
+            # * Any number of digits
+            # * An optional decimal point
+            # * Then exactly 2 digits - if there was a decimal
+            "pattern": r"^\d*([.]\d{2})?$",
+            # Limit the max amount based on the length (11-digits, allows up to 99 billion)
+            "maxLength": 14,
+            "title": "Award Amount",
+            "description": "For a covered Federal action where there has been an award or loan commitment by the Federal agency, enter the Federal amount of the award/loan commitment of the prime entity identified in item 4 or 5.",
         },
         "state_code": {
             "type": "string",
--- a/api/src/form_schema/forms/sflll.py
+++ b/api/src/form_schema/forms/sflll.py
@@ -96,10 +96,10 @@ FORM_JSON_SCHEMA = {
                     "maximum": 99,
                 },
                 "applicant_reporting_entity": {
-                    "allOf": [{"$ref": "#/$defs/reporting_entity_awardee"}],
+                    "$ref": "#/$defs/reporting_entity_awardee",
                 },
                 "prime_reporting_entity": {
-                    "allOf": [{"$ref": "#/$defs/reporting_entity_awardee"}],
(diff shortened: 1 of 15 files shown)


api/src/form_schema/forms/sflll.py
```diff
@@ -96,10 +96,10 @@
                     "maximum": 99,
                 },
                 "applicant_reporting_entity": {
-                    "allOf": [{"$ref": "#/$defs/reporting_entity_awardee"}],
+                    "$ref": "#/$defs/reporting_entity_awardee",
```
chouinar, line 99: Heads up, these being `allOf` instead of just ref is intentional. I intentionally don't use ref directly because allOf allows for multiple inheritance (not used yet), but more importantly lets you override things from the value like the title or description which _is_ necessary on some of these. This schema doesn't seem to use that, but on other schemas I made a base type that had variations of the title/description.

Was there any issue you encountered to change this?

api/src/form_schema/forms/sflll.py
```diff
@@ -134,7 +134,16 @@
...
+            # Represents a monetary amount. We use a string instead of number
+            # to avoid any floating point rounding issues.
+            "type": "string",
+            # Pattern here effectively says:
+            # * Any number of digits
+            # * An optional decimal point
+            # * Then exactly 2 digits - if there was a decimal
+            "pattern": r"^\d*([.]\d{2})?$",
+            # Limit the max amount based on the length (11-digits, allows up to 99 billion)
+            "maxLength": 14,
```
chouinar: This should stay a ref, any refs I added, even if they're only used once is because soon we'll move those to a single schema that every schema references, those types were separated because they're already in multiple forms (the SF424A for example has a lot of this same type).

api/src/form_schema/forms/sflll.py
doug-s-nava: adding the "allOf" here makes working with the UI schema definitions difficult. Decision is to push this difficulty down the road a bit and move forward with this change for now. See https://betagrantsgov.slack.com/archives/C08NCEXRSPJ/p1753806394533519

frontend/src/components/applyForm/widgets/TextAreaWidget.tsx
```diff
@@ -63,7 +63,16 @@ function TextAreaWidget<
     <FormGroup error={error} key={`wrapper-for-${id}`}>
       <FieldLabel idFor={id} title={title} required={required} />
 
-      {error && <ErrorMessage>{rawErrors[0]}</ErrorMessage>}
+      {error && (
```
acouch, line 66: Thanks for this. we could maybe make this its own component in the future.

Issue #5740, opened 2025-07-21
BE work to limit attachment size


chris-kuryak:
### Summary

Limit file attachment size on applications to 2 GB

We should be able to do this in one of two ways:
* https://apiflask.com/request/#file-uploading - add a validator to the file upload schema - NOTE - you cannot use a validator directly from the APIFlask library and would need to rewrite it due to how the error handling works. See for example how we rewrote several validators already in https://github.com/HHS/simpler-grants-gov/blob/main/api/src/api/schemas/extension/field_validators.py
* https://flask.palletsprojects.com/en/stable/patterns/fileuploads/#improving-uploads - this is likely simpler, but I'm not certain it works / what the errors would end up as. Test this out and see what happens - if this way works in a way compatible with our system, I'd probably prefer it.

### Acceptance criteria

TBD

PR #5808, opened 2025-07-29
[Issue #5740] Limit backend attachment size

mikehgrantsgov:
## Summary

Fixes / Work for #5740 

## Changes proposed

Use Flask's built-in file size limits to restrict upload size to 2GB.
Add unit tests

## Context for reviewers

Explored a few avenues to implement this, first via custom validators then with the Flask built-in approach. Chose this method due to its simplicity / small changeset.

## Validation steps

See unit tests.


api/src/api/application_alpha/application_schemas.py
```diff
@@ -5,8 +5,13 @@
     FileResponseSchema,
     WarningMixinSchema,
 )
+from src.app_config import AppConfig
 from src.constants.lookup_constants import ApplicationFormStatus, ApplicationStatus
 
+# Get the maximum file size from config
+app_config = AppConfig()
+MAX_FILE_SIZE_BYTES = app_config.max_file_upload_size_bytes
+
```
chouinar: I don't think we need this here?

api/tests/src/api/applications/test_flask_file_size_limits.py
```diff
@@ -0,0 +1,107 @@
...
+    assert response.json["message"] == "Success"
+    assert "application_attachment_id" in response.json["data"]
+
+
+def test_flask_max_content_length_configuration():
+    """Test that the Flask app is configured with the correct MAX_CONTENT_LENGTH."""
+    from src.app import create_app
+    from src.app_config import AppConfig
+
+    app = create_app()
```
chouinar: The `app` fixture calls `create_app` should be able to use that - calling it multiple times can do some weird things (the logging will double every time you call it).

```suggestion
def test_flask_max_content_length_configuration(app):
    """Test that the Flask app is configured with the correct MAX_CONTENT_LENGTH."""
```

api/src/api/application_alpha/application_schemas.py
mikehgrantsgov: Ah yeah, left that around from when I was experimenting with custom Validators. Removed it

chouinar (approved):

PR #5814, opened 2025-07-29
feat(form): sf424

myduong-navapbc:
## Summary

Fixes / Work for #ISSUE


api/src/form_schema/forms/sf424.py
```diff
@@ -347,6 +347,8 @@
             "maxLength": 255,
         },
         "areas_affected": {
+            "type": "string",
+            "format": "attachment-id",
```
chouinar: FYI - we cannot add a format here that isn't specified in the JSON schema specification, if we need some way of communicating that this is an attachment ID to the frontend, we cannot do it here. I don't want to extend the JSON schema to require any users of the the schema to also need to extend their usage of JSON schema libraries.

myduong-navapbc: removed!

Issue #5810, opened 2025-07-29
Create a able for storing suppressed emails


babebe:
### Summary

Create a table to track suppressed emails called `suppressed_email` with the following columns. 
suppressed_email_id - UUID | Primary Key
email -  str | email address on the suppression list.
reason - str | reason for being added to suppression list 

### Acceptance criteria

- [ ] Table created
- [ ] Migration created

PR #5824, opened 2025-07-30
[Issue #5810] Create suppressed emails table

babebe:
## Summary

Fixes for #5810  

## Changes proposed

Created suppressed emails table and factory


api/src/db/migrations/versions/2025_07_30_add_suppressed_email_table.py
```diff
@@ -0,0 +1,66 @@
...
+Create Date: 2025-07-30 14:46:46.144742
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "6cae8331ab57"
+down_revision = "c59a945c96c5"
```
chouinar: Mike Merged in a PR this morning that adds some new migrations that you'd need to either adjust this to / regenerate the migration.

PR #5825, opened 2025-07-30
[Unticketed] Make all post-populated fields non-required

chouinar:
## Changes proposed
Make all fields that are going to be post-populated non-required.

## Context for reviewers
This doesn't technically change what fields will be populated when submission occurs, but will remove error messages complaining about "missing a required field" for a field that a user can't fill in. We know the field will be set by our post-population (when that gets enabled in the next few days) so giving a user an error they can't do anything about is just noise.


acouch (approved):
Looks great, thanks.

Issue #5244, opened 2025-06-04
when searching agencies, top level agencies with opportunities are not usable


doug-s-nava:
### Summary

if a top level agency has opportunities assigned to it, and you run a search against the agency filter options list that contains sub agencies for the agency, you will only be able to select sub agencies of the agency, and not the agency itself

### Reproducibility

None

### How to reproduce the bug

* visit https://simpler.grants.gov/opportunity/334079
* _VERIFY_: agency assigned is top level (DOC)
* visit https://simpler.grants.gov/search
* select agency filter to see all DOC opportunities
* _VERIFY_: 334079 does not appear in results

### Notes 
This addresses the filtering issues, not the search aspect, search will be handled on a separate ticket. 
None

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

eric-serrato95:
Could be fixed by both FE and BE, after consulting, might be easier to fix this on the BE 
we have API consumers that aren't on the FE, one possible solutions could be add more data to thee search index

eric-serrato95:
2 possible solutions, 
1.- when selecting top level agency, we return every opportunity associated with both top level agency and with every single sub-agency 
2.- Grants does offer a separate box for supporting only opportunities associated with a top level agency.

joshtonava:
@eric-serrato95 @chrisw-nava This is something that can be UATed

eric-serrato95:
@joshtonava UAT ✅  moving to done

PR #5826, opened 2025-07-30
[Issue #5244] Include opportunities assigned to top level agency when filtering by top level agency

joshtonava:
## Summary

Work for #5244 

## Changes proposed

- Added exact match filtering for top-level agency codes (e.g., "DOC") in addition to existing prefix matching for sub-agencies (e.g., "DOC-EDA")
- Modified `_add_top_level_agency_prefix` function in `search_opportunities.py` to include both exact matches and prefix matches when filtering by top-level agency
- Added new test opportunity `DOC_TOP_LEVEL` and updated all relevant test cases to verify the enhanced functionality

## Context for reviewers

Previously, when searching for opportunities by top-level agency (e.g., "DOC"), the search would only return opportunities from sub-agencies (e.g., "DOC-EDA", "DOC-MANUFACTURING") but would miss opportunities directly assigned to the top-level agency itself with the exact code "DOC".

The fix adds an exact match filter alongside the existing prefix match filter in the OpenSearch query. This ensures that:
1. Opportunities with exact agency codes like "DOC" are found via exact match
2. Opportunities with sub-agency codes like "DOC-EDA" are found via prefix match on "DOC-"
3. Both types of opportunities are returned in a single search query using OpenSearch's `should` clause

The implementation uses the existing `filter_should_terms` method for exact matching and maintains the existing `filter_should_prefix` method for sub-agency matching, ensuring backward compatibility while extending functionality.

## Validation steps

Search for opportunities using a top-level agency filter (e.g., "DOC") and verify that both top-level agency opportunities and sub-agency opportunities are returned

### DIFF
--- a/api/tests/src/api/opportunities_v1/test_opportunity_route_search.py
+++ b/api/tests/src/api/opportunities_v1/test_opportunity_route_search.py
@@ -1732,36 +1732,3 @@ class TestOpportunityRouteSearch(BaseTestClass):
         assert [opp["opportunity_id"] for opp in data] == [
             opp.opportunity_id for opp in [DOS_DIGITAL_LITERACY, DOC_SPACE_COAST, DOC_MANUFACTURING]
         ]
-
-    def test_search_agency_filter_includes_top_level_agency_when_sub_agencies_present_200(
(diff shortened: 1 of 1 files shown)


chouinar (approved):
LGTM!

Issue #5815, opened 2025-07-29
Mock list-suppressed-destinations API


babebe:
### Summary

Created a mock SESv2 client to simulate the list-suppressed-destinations API for testing purposes.
This is necessary because moto does not currently expose a direct method under the standard SES client.

The mock client will have:
1. a function that will be used to seed mock response data  
2. a function that mocks `list_suppressed_destinations` method that filters the seeded data and return only the records matching the request

[Reference](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sesv2/client/list_suppressed_destinations.html?utm_source=chatgpt.com)

### Acceptance criteria

[ ] Mock client created

PR #5829, opened 2025-07-31
[Issue #5815] Mock sesv2 client

babebe:
## Summary
Fixes / Work for #5815 
Created Mock sesv2 client. 

## Changes proposed

Create a mock SESv2 client to simulate the list-suppressed-destinations API for testing purposes.

## Context for reviewers
Moto does not support SESv2.


api/src/adapters/aws/sesv2_adapter.py
```diff
@@ -0,0 +1,33 @@
...
+logger = logging.getLogger(__name__)
+
+
+class SuppressedDestination(BaseModel):
+    email_address: str = Field(alias="EmailAddress")
+    reason: str = Field(alias="Reason")
+    last_update_time: datetime = Field(alias="LastUpdateTime")
+
+
+class MockSESV2Client:
```
chouinar: For the mock class to be something we can use, it needs to have the real class alongside it. We should make a base class that the mock one is derived from (alternatively can override the real one, depends on what we might be doing):

For example:
```py

class BaseClient(abc.ABC, metaclass=abc.ABCMeta):
    @abstractmethod 
    def call_endpoint(request: Request) -> Response:
          pass
    
class RealClient(BaseClient):
    def call_endpoint(request: Request) -> Response:
          aws_call(...)

class MockClient(BaseClient):
    def call_endpoint(request: Request) -> Response:
          return self.mock_responses[...]

    def add_mock_response(...):
          pass
```
Unless we have a real client, it can be hard to make a mock client because we need to test it with a real one to know if we're mocking the right thing.

And to make it so the mock client can be used, we can do something like:

```py
def get_client() -> BaseClient:
     # Determine this based on an env var so we can turn it on locally / in tests, but non-locally it's the real one
     config = MyConfig() # a pydantic config class
     
```
And finally when we want to run tests, if we setup the task class that is going to use this like so:
```py
class MyTask(Task):
    def __init__(self, db_session: db.Session, client: BaseClient | None):
           super().__init__(db_session)
           if client is None:
               client = RealClient()
           self.client = client

# and using that in a test
def test_example(db_session):
     mock_client = MockClient()
     mock_client.add_mock_response(...)
     task = MyTask(db_session, mock_client)
```

babebe: Ah yes, it makes sense to add a common interface. Updated.

api/src/adapters/aws/sesv2_adapter.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+class SESV2Client(BaseSESV2Client):
+    def list_suppressed_destinations(self, start_date: datetime | None = None) -> SESV2Response:
+        client = boto3.client("sesv2")
+        response = (
+            client.list_suppressed_destinations(StartDate=start_date)
+            if start_date
+            else client.list_suppressed_destinations()
+        )
```
chouinar: Rather than alternative call methods depending on params, make the params themselves the thing we vary like so.

```suggestion
        request_params = {}
        if start_date:
           request_params["StartDate"] = start_date
           
        response = client.list_suppressed_destinations(**request_params)
        )
```

api/src/adapters/aws/sesv2_adapter.py
```diff
@@ -0,0 +1,59 @@
...
+
+class BaseSESV2Client(ABC, metaclass=ABCMeta):
+    @abstractmethod
+    def list_suppressed_destinations(self) -> SESV2Response:
+        pass
+
+
+class SESV2Client(BaseSESV2Client):
+    def list_suppressed_destinations(self, start_date: datetime | None = None) -> SESV2Response:
+        client = boto3.client("sesv2")
```
chouinar: It won't matter much with how we're planning to use it, but better practice for AWS clients is to initialize once and reuse since the startup for them is surprisingly long.

Could you do
```py
    def __init__(self):
        self.client = boto3.client("sesv2")
```

PR #5831, opened 2025-07-31
Infra 5775 - Standup API Gateway configuration

### LINKED ISSUE #5775: INFRA – Stand‑up API Gateway REST API (HTTP proxy via VPC Link)
### Summary

Stand‑up a REST API in Amazon API Gateway that forwards every request (/{proxy+}) to the existing public Application Load Balancer in the production VPC using an HTTP proxy integration and a VpcLink. Follow the “proxy resource” pattern so path & query strings are transparently passed through.

### Acceptance criteria

Terraform plan creates:
  - [ ] aws_api_gateway_rest_api named simpler-grants-api-prod.
  - [ ] aws_api_gateway_resource /{proxy+} with ANY method.
  - [ ] aws_api_gateway_integration type HTTP_PROXY pointing at the ALB DNS via aws_api_gateway_vpc_link.

- [ ] Deployment to stage prod succeeds; GET /health returns 200 via Gateway and mirrors direct ALB response.
- [ ] Gateway logging enabled to CloudWatch; logs show target URL & latency.
- [ ] Terraform plan is idempotent; destroy preserves ALB.

### LINKED ISSUE #5776: INFRA – Add Internal & Developer Usage Plans + Header Key Source
### Summary

Plan | Throttle | Quota | Purpose
-- | -- | -- | --
internal | Unlimited | ∞ | First‑party services
developer | 10 req/s | 250K req/month | External keys

Set the API‑level API‑Key source to HEADER (X‑API‑Key) as recommended.
[AWS Documentation 1](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-api-key-source.html)
[AWS Documentation 2](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-api-key-call.html)

### Acceptance criteria

- [ ] Terraform aws_api_gateway_usage_plan resources with limits above.
- [ ] aws_api_gateway_rest_api property api_key_source = "HEADER".
- [ ] Stage prod attached to both usage plans.
- [ ] Verification:
  - [ ] Without key ⇒ 403 Forbidden.
  - [ ] With internal key ⇒ 200 at >10 req/s.
  - [ ] With developer key ⇒ throttled at 11 req/s.

Nava-JoshLong:
## Summary

Fixes #5775 
Fixes #5776 

## Changes proposed

- Configures the base API gateway with
    - Logging
    - Controlling traffic from the gateway to the ALB
    - Proxies all endpoints for the API to reduce terraform maintenance
    - Creates API keys to used for internal and public systems

## Context for reviewers

This PR will add the configuration the API gateway needs to work, since right now it is not forwarding or responding to any requests. This PR will enable

- Full proxy from the gateway to the API. Defining what endpoints the gateway can hit and how is very cumbersome
    - We're hoping to lock down access for public vs internal systems by using API key rate limiting. This will need some testing
- Logging for the gateway so we can review requests. The format is a modified example from AWS, with the required fields added from the ticket's ACs
- Auto-deploys the gateway changes when it detects the file is changed via terraform workflows
- Comments added for more info, as well as possible issues for the proposed setup
    - The main issue I'm seeing is if we set the API gateway to use the same domain name as the ALB, it will cause looping and issues
- API keys and usage plans for internal vs public users/systems
- I possibly might need to add some changes for VPC link networking to the ALB
    - This is usually for private LBs, but we might want to point to the ALB's AWS DNS instead of the public domain to reduce hops and latency
    - It might also handle the domain conflict issue

## Validation steps

Once this PR is approved, I will see what steps are needed to deploy this to dev and test to ensure everything is configured correctly before merging


infra/modules/service/api_gateway.tf
```diff
@@ -9,8 +9,119 @@ resource "aws_api_gateway_rest_api" "api" {
...
+  access_log_settings {
+    destination_arn = aws_cloudwatch_log_group.api_gateway_logs[0].arn
+    format          = "{ \"requestId\":\"$context.requestId\", \"extendedRequestId\":\"$context.extendedRequestId\",\"ip\": \"$context.identity.sourceIp\", \"caller\":\"$context.identity.caller\", \"apiKeyId\":\"$context.identity.apiKeyId\", \"requestTime\":\"$context.requestTime\", \"httpMethod\":\"$context.httpMethod\", \"resourcePath\":\"$context.resourcePath\", \"status\":\"$context.status\", \"protocol\":\"$context.protocol\", \"responseLength\":\"$context.responseLength\", \"responseLatency\": \"$context.responseLatency\" }"
+  }
+
+  cache_cluster_enabled = true
+
+  # checkov:skip=CKV_AWS_73:X-Ray can increase costs greatly, and aren't always necessary
+  # checkov:skip=CKV2_AWS_29:WAF can be enabled at a later time if needed
+  # checkov:skip=CKV2_AWS_51:Mutual TLS increases complexity for downstream systems
```
Nava-JoshLong, line 94: For these, I think it's okay to not have, but I wanted to run it by you to make sure

- For X-Ray, we can enable this when there's problems, but it doesn't always give a ton of insight when there's problems. The logs give more info IMO
- WAF can be costly if the account already isn't using it, and can need a good amount of time to setup
- Mutual TLS seems like it might be overkill for this setup, but please correct me if I'm wrong

infra/modules/service/api_gateway.tf
```diff
@@ -9,8 +9,119 @@ resource "aws_api_gateway_rest_api" "api" {
...
+}
+
+resource "aws_cloudwatch_log_group" "api_gateway_logs" {
+  count = var.enable_api_gateway ? 1 : 0
+
+  name              = "service/${var.service_name}-api-gateway/v1-stage"
+  retention_in_days = 1827
+
+  # TODO(https://github.com/navapbc/template-infra/issues/164) Encrypt with customer managed KMS key
+  # checkov:skip=CKV_AWS_158:Encrypt gateway logs with customer key in future work
```
Nava-JoshLong, line 106: Copied the todo from another file, I think that's fine? Should I update that ticket with this file?

infra/modules/service/api_gateway.tf
```diff
@@ -9,8 +9,119 @@ resource "aws_api_gateway_rest_api" "api" {
...
+  rest_api_id = aws_api_gateway_rest_api.api[0].id
+  stage_name  = aws_api_gateway_stage.api_v1_stage[0].stage_name
+  method_path = "*/*"
+
+  settings {
+    metrics_enabled      = true
+    logging_level        = "INFO"
+    caching_enabled      = true
+    cache_ttl_in_seconds = 300
+    cache_data_encrypted = true
```
Nava-JoshLong: Some of the findings were related to caching. I think it's okay to enable, but if the endpoints will be hit less often than every 5 minutes, this might not be worth it

infra/modules/service/api_gateway.tf
mdragon: If we don't call out other projects in that main ticket, I don't know that we need to be the first. We'd pick this up via in infra-template update in the future?

mdragon (approved):

infra/modules/service/api_gateway.tf
Nava-JoshLong: I'm not 100% sure, I'll check. I don't think this would get one since I don't think API gateway is in the template infra repo. Once the logic is in this repo, it would just be copying over the logic into this file and gate it behind the count param

Nava-JoshLong: Forgot to circle back on this, the template repo isn't using API gateway, so we wouldn't automatically get CMKs for this. We can have that be a follow-up ticket here if you want, it's pretty straight forward to implement

mdragon: Ok, I don't think we have to do any more on this in the short term

Nava-JoshLong:
First pass shows it working as expected. Doing some work now to test the public usage plan and limiting what it can and can't hit. I have a few tweaks I'll need to push and some logic for the account level change required

```bash
curl https://wxuxv45ooj.execute-api.us-east-1.amazonaws.com/v1/health -H 'x-api-key: ********' 
{
  "data": {
    "commit_link": "https://github.com/HHS/simpler-grants-gov/commit/ce6559be4fde8e99f5e774f50e614a5c03245c08",
    "commit_sha": "ce6559be4fde8e99f5e774f50e614a5c03245c08",
    "deploy_whoami": "runner",
    "last_deploy_time": "2025-08-01T16:44:38-04:00",
    "release_notes_link": "https://github.com/HHS/simpler-grants-gov/releases"
  },
  "message": "Service healthy",
  "status_code": 200
}
```

infra/modules/service/api_gateway.tf
mdragon: I think this could be an issue if we can't force the cache to clear. For example a used adds or deletes a saved opportunity. But the cache has the old data so it appears it didn't work when the user looks at their Saved Opportunities.

mdragon: Maybe if we could limit it to the "unauthenticated" endpoints like Opportunities and Search that would work better. But maybe we just leave it off for now?

Nava-JoshLong:
Findings from the testing

- Base infra works as is, some things I noticed/tweaked
    - I'll need to run another test later to make sure the logging works as expected since it has to be done at the account level
        - I created a file with an account level logging role, since it isn't enabled right now
    - Is there a cache size that makes sense to have? It's in GBs, I was thinking either 0.5, 1.6 (only certain values are allowed)

We'll need to slightly pivot what we have if we want to ensure that there are certain endpoints that the public can't hit. Since we're doing a full proxy, I can't tell the gateway to restrict access to paths it doesn't know exists or not. We could implement this by having specific internal only endpoints defined in a `internal/*` or `private/*` path for both the gateway and ECS. These would have a 0 burst and rate limit and I think this would let us lock down access so public users can't hit them. Once that's set, we would have a `public/*` path that they can use with the normal rate limiting. I'd need to test to see, I can't do that with the code as is right now unless I do both pointing to the normal endpoints and see if the private path is restricted to the public usage plan

infra/modules/service/api_gateway.tf
Nava-JoshLong: I'll turn it off for now, I'm not as familiar with caching, so I don't want it bogging the PR down. We can spin up a follow-up to look into it, and see what is and isn't possible

mdragon:
> We'll need to slightly pivot what we have if we want to ensure that there are certain endpoints that the public can't hit. Since we're doing a full proxy, I can't tell the gateway to restrict access to paths it doesn't know exists or not. 

@Nava-JoshLong Just noting we want to do the opposite, I think. Allow un-authenticated "public" access through API GW to the BE for a few endpoints like: /health, the login flow paths, etc.

Nava-JoshLong:
> @Nava-JoshLong Just noting we want to do the opposite, I think. Allow un-authenticated "public" access through API GW to the BE for a few endpoints like: /health, the login flow paths, etc.

@mdragon I think we're getting our wires crossed. For the gateway, it's setup to require API key usage on any endpoints the gateway forwards to, but it will forward any traffic it gets to the ECS service. The ECS service we are forwarding to can have its own authentication, but I thought there was some logic we wanted to add for restricting what endpoints the gateway can and can't hit, which would have to be from the usage plan on the API gateway key unless we want to define each endpoint in the API gateway. 

That's where the current issue is, since we're doing a fully proxy, we can't do any form of locking down because the gateway doesn't know what paths it's proxying to. That's why I brought up the internal vs public proxy endpoints we could do instead. That would look something like this

```terraform
resource "aws_api_gateway_resource" "api_public_proxy" {
  count = var.enable_api_gateway ? 1 : 0

  rest_api_id = aws_api_gateway_rest_api.api[0].id
  parent_id   = aws_api_gateway_rest_api.api[0].root_resource_id
  # This would be different in actual code, but putting like this to try and make it clearer
  path_part   = "public/{proxy+}"
}

resource "aws_api_gateway_resource" "api_internal_proxy" {
  count = var.enable_api_gateway ? 1 : 0

  rest_api_id = aws_api_gateway_rest_api.api[0].id
  parent_id   = aws_api_gateway_rest_api.api[0].root_resource_id
  # This would be different in actual code, but putting like this to try and make it clearer
  path_part   = "internal/{proxy+}"
}

# Other resources to forward that traffic down to the ALB/public and ALB/internal endpoints

resource "aws_api_gateway_usage_plan" "api_public_usage_plan" {
  count = var.enable_api_gateway ? 1 : 0

  name = "${var.service_name}-public-usage-plan"

  api_stages {
    api_id = aws_api_gateway_rest_api.api[0].id
    stage  = aws_api_gateway_stage.api_v1_stage[0].stage_name

    throttle {
      path        = "/internal/ANY"
      burst_limit = 0
      rate_limit  = 0
    }
  }

  # Global level throttling for this usage plan
  throttle_settings {
    burst_limit = 5
    rate_limit  = 10
  }
}
```

We can hop on a call to try and iron things out if you want too

Nava-JoshLong:
We were able to confirm logging is working as expected. @btabaska can you confirm this looks like all of the fields you want? I want to call out the resource path shows just proxy, and we aren't sure if there could be any PII as a path param that we might want to log

```json
{
    "requestId": "10bb0992-2883-4b2a-a8d6-3065355fec23",
    "extendedRequestId": "PNjhhE10oAMEglw=",
    "ip": "REDACTED",
    "caller": "-",
    "apiKeyId": "6n7v7jc9r0",
    "requestTime": "12/Aug/2025:21:32:06 +0000",
    "httpMethod": "GET",
    "resourcePath": "/{proxy+}",
    "status": "200",
    "protocol": "HTTP/1.1",
    "responseLength": "411",
    "responseLatency": "29"
}
```

Nava-JoshLong:
Public usage plan rate limiting test

```
seq 1 50 | xargs -P50 -I '$' curl -s -o /dev/null -w "%{http_code}\n" https://wxuxv45ooj.execute-api.us-east-1.amazonaws.com/v1/health -H 'x-api-key: REDACTED'
429
200
429
429
200
429
502
200
429
200
429
200
200
200
200
200
200
200
200
200
200
429
200
429
429
200
200
429
200
200
429
200
200
200
200
200
200
200
200
200
200
200
200
200
200
200
200
200
200
200
```

btabaska:
> We were able to confirm logging is working as expected. @btabaska can you confirm this looks like all of the fields you want? I want to call out the resource path shows just proxy, and we aren't sure if there could be any PII as a path param that we might want to log
> 
> ```json
> {
>     "requestId": "10bb0992-2883-4b2a-a8d6-3065355fec23",
>     "extendedRequestId": "PNjhhE10oAMEglw=",
>     "ip": "REDACTED",
>     "caller": "-",
>     "apiKeyId": "6n7v7jc9r0",
>     "requestTime": "12/Aug/2025:21:32:06 +0000",
>     "httpMethod": "GET",
>     "resourcePath": "/{proxy+}",
>     "status": "200",
>     "protocol": "HTTP/1.1",
>     "responseLength": "411",
>     "responseLatency": "29"
> }
> ```

This looks good to start. I'd rather log less and then have fast follows if additional logging is needed in the future. Thanks!

Issue #5489, opened 2025-06-27
Run post-population during submit flow


chris-kuryak:
### Summary
For example, signature and current date would be disabled while filling out the application, and then we would auto-populate them upon submission.

---

We want to enable post-population. To do this we need to do a few things in [application_validation.py](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/services/applications/application_validation.py)
* For the submit config, swap the config to enable post population where noted (should just be the one config)
* In `validate_application_form` if we haven't already connected it, we want the json from the context we pass into `process_rule_schema_for_context` to overwrite the JSON in application form
  * Note - pre-processing makes a copy of the JSON and its here where we actually "apply" the changes. 

### Acceptance criteria

- [ ] Post populated fields are set correctly during the process initiated for the SUBMIT endpoint

PR #5846, opened 2025-08-01
[Issue #5489] Run post-population during submit flow

mikehgrantsgov:
## Summary

Fixes / Work for #5489 

## Changes proposed

Set `do_post_population=True`
Set `application_form.application_response = context.json_data` in validate
Update tests

## Context for reviewers

Signature and current date would be disabled while filling out the application, and then we would auto-populate them upon submission.

## Validation steps

See updated unit tests

### DIFF
--- a/api/src/services/applications/application_validation.py
+++ b/api/src/services/applications/application_validation.py
@@ -171,9 +171,7 @@ def validate_application_form(
     process_rule_schema_for_context(context)
     form_validation_errors.extend(context.validation_issues)
 
-    # Apply post-population changes to the application form during submit
-    if action == ApplicationAction.SUBMIT and context.config.do_post_population:
-        application_form.application_response = context.json_data
+    application_form.application_response = context.json_data
 
     # Check if the form is required
     is_required = is_form_required(application_form)
--- a/api/src/services/applications/application_validation.py
+++ b/api/src/services/applications/application_validation.py
@@ -48,7 +48,7 @@ SUBMISSION_JSON_RULE_CONFIG = JsonRuleConfig(
     # During submission, we do post-population (the only place we ever do)
     # and field validation
     do_pre_population=False,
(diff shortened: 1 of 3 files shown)


api/tests/src/api/applications/test_application_routes.py
```diff
@@ -46,6 +46,21 @@
...
+        "signature_field": {"type": "string"},
+        "date_field": {"type": "string"},
+        "existing_field": {"type": "string"},
+    },
+}
+
+POST_POPULATION_RULE_SCHEMA = {
+    "signature_field": {"gg_post_population": {"rule": "signature"}},
+    "date_field": {"gg_post_population": {"rule": "current_date"}},
+}
```
chouinar: These aren't used

api/src/services/applications/application_validation.py
```diff
@@ -171,7 +171,9 @@ def validate_application_form(
     process_rule_schema_for_context(context)
     form_validation_errors.extend(context.validation_issues)
 
-    # TODO pre/post-populate should update the value here
+    # Apply post-population changes to the application form during submit
+    if action == ApplicationAction.SUBMIT and context.config.do_post_population:
+        application_form.application_response = context.json_data
```
chouinar: It'll seem like potentially too much, but we actually want to just always do:
```suggestion
    application_form.application_response = context.json_data
```
The configuration itself already determines whether we actually run pre/post population rules. With pre-population disabled (until the next ticket) we _shouldn't_ ever be change anything outside of submit. When we do add pre-population, we'll have a lot of places where we do this.

api/tests/src/api/applications/test_application_routes.py
mikehgrantsgov: Right, I need to set my IDE to pop those up as linter errors.

Issue #5314, opened 2025-06-11
Testing of Sam.gov functionality


chris-kuryak:
### Summary

This ticket is a placeholder to test the Sam.gov functionality once our work is complete.

- Pick up files
- Parse files

### Acceptance criteria
- [ ] BE alignment on what is a successful test
- [ ] Test of Sam.gov functionality in Simpler Apply Pilot to ensure working as expected

chouinar:
Noting down what I did for testing of sam.gov.

We received access to sam.gov extracts in prod. We were able to test the first part of the sam.gov task in prod which fetches the extracts. I did confirm that for when we ran it (July 31st) that we fetched all of the expected files by comparing to the DB table in grants.gov that notes down the files it processed - we grabbed the exact same ones for that time period.

I downloaded the extracts to test the processing logic locally. There were a few issues that I fixed, most notably handling UEI being duplicated in the extract files and various performance issues that likely would have caused the job to fall over.

Next steps:
* Verify that running the extract fetching step will behave correctly now that we're on the next month.
* Verify that the performance improvements made keep the job working when processing a monthly extract.
* Verify logic for setting up organizations works correctly.
* Ideally some of this testing would be possible in our lower environments so we don't need to test in production.

PR #5854, opened 2025-08-04
[Issue #5314] Adjust sam.gov extract parsing based on testing

chouinar:
## Summary

Fixes / Work for #5314 

## Changes proposed
Modified sam.gov extract parsing to account for duplicate UEIs in a given extract file (see below for context)

Modified extract processing to use a bit less memory, not loading entire entity table into memory, and processing file line-by-line instead of as a big text blob.

A few other refactors for simplifying logic.

## Context for reviewers
I assumed that we would only receive a single row in a given extract file for a UEI, but it turns out that was wrong. If a given entity has multiple "EFT Indicators" then each of those will be a separate row in the data we receive. This indicator is related to bank account information according to the docs, and is for the most part irrelevant to us. From validating the monthly extract, I verified that the only two fields that ever differ across these duplicate records are EFT indicator and cage code (a value we do not use). I also verified that every UEI has at least one record where the EFT indicator is blank, which is the one we will use, and skip the others (I added validation/logging in the job to catch if this is ever not true).

Additionally, I changed some of the parsing logic because the way it was written, we'd load the entire file, and the entire set of sam.gov entities into the processes memory, and I know that is going to break when we try to run it for real. Instead we fetch 10k entities at a time (note that the monthly extract has just shy of 900k records) and the file processing logic was restructured a bit to use proper file parsing instead of loading the text of the file fully into memory all at once.

## Validation steps
For verifying the changes I made, I actually used extract files from production running locally, and that data is sensitive, so can't quite share how you can do it, but here's what I did.

I downloaded all extract files we fetched (one monthly file, 16 daily files over the entirety of July), setup my DB to have all of them loaded in, pointing to the files on disk. Running locally, this takes ~10 minutes, about 7.5 of which are the monthly extract. I expect it'll take a bit longer non-locally, but we are fine with that.

I ran just the processing part, and skipped setting up files since I did that manually.
`make cmd args="task sam-extracts --no-fetch-extracts --no-create-orgs"`

From running locally, I had zero errors processing rows, and none of the bad scenarios occurred with data, so the assumptions I've made seem to at least be fairly valid.

### DIFF
--- a/api/src/task/sam_extracts/process_sam_extracts.py
+++ b/api/src/task/sam_extracts/process_sam_extracts.py
@@ -302,6 +302,7 @@ class ProcessSamExtractsTask(Task):
         sam_extract_file: SamExtractFile,
         extract_log_extra: dict,
     ) -> None:
+        """Load sam.gov entities to the DB"""
         total_to_process = len(sam_gov_entity_container.processed_entities)
         total_processed = 0
         for batch in itertools.batched(
--- a/api/src/task/sam_extracts/process_sam_extracts.py
+++ b/api/src/task/sam_extracts/process_sam_extracts.py
@@ -1,12 +1,15 @@
+import itertools
 import logging
+import time
 import zipfile
 from dataclasses import dataclass, field
 from datetime import date, datetime
 from enum import StrEnum
-from typing import Sequence
+from typing import IO, Sequence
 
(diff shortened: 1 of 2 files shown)


mikehgrantsgov (approved):
Looks good! Confirmed the make command works on my machine as well

PR #5856, opened 2025-08-04
[Issue #4994] Drop redirect when not in maintenance mode for now

mdragon:
## Summary
Work for #4994

## Changes proposed

Stop redirecting from /maintenance to the home page if no "major" features are off. This wasn't working reliably and we've seen from past downtime that folks don't seem to end up "stuck" on the maintenance page.

### DIFF
--- a/frontend/src/components/ClientMaintenanceCheck.tsx
+++ b/frontend/src/components/ClientMaintenanceCheck.tsx
@@ -2,7 +2,6 @@
 
 import { useFeatureFlags } from "src/hooks/useFeatureFlags";
 
-import { redirect, RedirectType } from "next/navigation";
 import { useEffect } from "react";
 
 /**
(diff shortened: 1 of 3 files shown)


frontend/src/components/ClientMaintenanceCheck.tsx
```diff
@@ -19,7 +18,8 @@ export default function ClientMaintenanceCheck() {
       !checkFeatureFlag("opportunityOff") &&
       checkFeatureFlag("authOn")
     ) {
-      redirect("/", RedirectType.push);
+      // This piece does not seem reliable, maybe due to re-render or rehyrdrate cylces where we redirect away before the actual client live feature flag is set
+      // redirect("/", RedirectType.push);
     }
   }, [checkFeatureFlag]);
```
freddieyebra, line 24: I would suggest replacing this useEffect with a comment or wrap it entirely in a comment to prevent unnecessary logic executions.

mdragon: I just took out the call to the Component in the caller, since we don't need to do any of this if the redirect isn't live. Good catch.

Issue #5394, opened 2025-06-18
Front-end work to create form SF424 locally


chris-kuryak:
### Summary

To ensure our approach to building forms will work as expected, we want to start on building out a form in it's entirety even though we have not identified our pilot opportunity.

This is the front-end work is to use the JSON (#5393 ) to create form SF424 locally for testing purposes.

- Form SF424: https://apply07.grants.gov/apply/forms/readonly/SF424_4_0-V4.0.pdf
- Info about form SF424: https://grants.gov/forms/form-items-description/fid/713

### Acceptance criteria

- [ ] Form SF424 is available for review and testing by team
- [ ] Concerns or needed changes are documented for product review and team discussion

acouch:
Is this ticket to create a similar automation for the UI schema or just get it working with the update from https://github.com/HHS/simpler-grants-gov/pull/5410 ?

chris-kuryak:
@acouch I believe this is just to get the form UI working for SF424. I don't think the original intent was to work on automation right now, but I will let @chouinar chime in if I am mis-interpreting

chouinar:
My understanding was it's the frontend half of setting up the SF424 form. I made a very rough UI schema as a start, but it doesn't fully work due to frontend regexes not liking underscores in names/nested data models.

Issue #5760, opened 2025-07-23
Address Attachment Card PR Comments


myduong-navapbc:
### Summary

These PR comments https://github.com/HHS/simpler-grants-gov/pull/5756

### Acceptance criteria

Address all the PR comments
https://github.com/HHS/simpler-grants-gov/pull/5756/files#r2226066037
https://github.com/HHS/simpler-grants-gov/pull/5756/files#r2226041806
https://github.com/HHS/simpler-grants-gov/pull/5756#pullrequestreview-3072475211

https://github.com/HHS/simpler-grants-gov/pull/5860#discussion_r2257517264

PR #5860, opened 2025-08-05
feat(forms): sf-424 standard form

myduong-navapbc:
## Summary

Work for #[5394](https://github.com/HHS/simpler-grants-gov/issues/5394)

## Changes proposed

SF424 json schema:
- fixed spelling and general text casing 

SF424 ui schema:
- added attachment widget
- added multi attachment widget
- created [sections based off pdf](https://apply07.grants.gov/apply/forms/readonly/SF424_4_0-V4.0.pdf)

Apply form:
- Attachment / Multiple Attachment Widgets

## Context for reviewers
- Updates to util, validation and types to accommodate the new widgets
- Adds context provider for attachments
- - I ran into an issue where the data that was saved for attachments were just the attachmentId, the designs showed the file name. in order to do that, we needed to know what the attachment array of objects were.
- 

## Validation steps

- For local testing: you cannot run `npm run dev`, strictmode throws an error when trying to attach a file to upload.
- https://github.com/HHS/simpler-grants-gov/issues/5724
- to run locally, you would need to run `npm run build && npm start`
- navigate to an opportunity that has competition data associated with it
- start application
- search for the **Application for Federal Assistance (SF-424)** in forms
- form should render and validate


frontend/src/app/[locale]/workspace/applications/application/[applicationId]/form/[appFormId]/page.tsx
```diff
@@ -33,17 +34,49 @@ export function generateMetadata() {
...
+interface UiSchemaSection {
+  type: "section";
+  name: string;
+  label?: string;
+  children: (UiSchemaField | UiSchemaSection)[];
+}
+
+type UiSchema = (UiSchemaField | UiSchemaSection)[];
+
+export function formHasAttachmentFields(uiSchema: UiSchema): boolean {
```
Comment (myduong-navapbc) [inline frontend/src/app/[locale]/workspace/applications/application/[applicationId]/form/[appFormId]/page.tsx]: What we get back from saved data is the attachment id,
before we make the determination and make a needless call to the backend to refresh the list of attachments, we check to see if the form even has attachments.

frontend/src/app/[locale]/workspace/applications/application/[applicationId]/form/[appFormId]/page.tsx
```diff
@@ -137,13 +174,14 @@ async function FormPage({ params }: formPageProps) {
           ]}
         />
         <h1>{form_name}</h1>
-        <ApplyForm
+        <ClientApplyForm
```
Comment (myduong-navapbc) [inline frontend/src/app/[locale]/workspace/applications/application/[applicationId]/form/[appFormId]/page.tsx]: Since we cannot use context hook on a server component, i had to create a wrapping client component

frontend/src/components/applyForm/utils.tsx
```diff
@@ -128,6 +146,34 @@ export const determineFieldType = ({
 }): WidgetTypes => {
   const { widget } = uiFieldObject;
   if (widget) return widget;
+
```
myduong-navapbc, line 138: this comes from the json schema, the attachments are set as `type of string` and `format of uuid`

frontend/src/components/applyForm/utils.tsx
```diff
@@ -509,3 +595,55 @@ export const getApplicationResponse = (
     return {};
   }
 };
+
```
myduong-navapbc: added some helpers to grab params from url

frontend/src/components/applyForm/utils.tsx
```diff
@@ -509,3 +595,55 @@ export const getApplicationResponse = (
...
+export const getApplicationIdFromUrl = (): string | null => {
+  if (typeof window === "undefined") return null;
+  const match = window.location.pathname.match(
+    /\/applications\/application\/([a-f0-9-]+)\/form\//,
+  );
+  return match?.[1] ?? null;
+};
+
+// getLabelComponent
+export function getLabelComponent<
```
myduong-navapbc: for future use, we can hide helper text if we set `hide-helper-text`

frontend/src/components/applyForm/widgets/FieldLabel.tsx
```diff
@@ -12,16 +12,20 @@ export const FieldLabel = ({
...
-      {required && (
-        <span className="usa-hint usa-hint--required text-no-underline">*</span>
-      )}
-      {description && (
-        <span>
-          <br /> {description}
-        </span>
-      )}
-    </Label>
+    <>
```
myduong-navapbc: this is one of a few steps to make the labels behave properly.

labels should be short and descriptive to each input. by adding the description inside the label, it creates an  unnecessarily long labels that do not help the user.

https://www.w3.org/WAI/WCAG22/Understanding/labels-or-instructions.html
https://www.w3.org/WAI/WCAG22/Techniques/general/G131
https://www.w3.org/WAI/WCAG22/Understanding/info-and-relationships

and while there are no hard and fast rules for length of labels. it is common practice to separate the label and helper text 

https://www.w3.org/WAI/tutorials/forms/instructions/

frontend/src/app/[locale]/workspace/applications/application/[applicationId]/form/[appFormId]/page.tsx
```diff
@@ -33,17 +34,49 @@ export function generateMetadata() {
   return meta;
 }
 
+interface UiSchemaField {
```
Comment (acouch) [inline frontend/src/app/[locale]/workspace/applications/application/[applicationId]/form/[appFormId]/page.tsx]: These are defined in `'./types.ts`

frontend/src/components/applyForm/widgets/FieldLabel.tsx
acouch: 👍🏿

acouch: USWDS has the label and description separate, but they look the same https://designsystem.digital.gov/templates/form-templates/name-form/

myduong-navapbc: i created this ticket to enhance how the assistive technologies behave with our forms
https://github.com/HHS/simpler-grants-gov/pull/5860

frontend/src/app/[locale]/workspace/applications/application/[applicationId]/form/[appFormId]/page.tsx
Comment (acouch) [reply frontend/src/app/[locale]/workspace/applications/application/[applicationId]/form/[appFormId]/page.tsx]: Does the context need to wrap the whole form, or could it just wrap the widget?

Comment (doug-s-nava) [reply frontend/src/app/[locale]/workspace/applications/application/[applicationId]/form/[appFormId]/page.tsx]: ApplyForm is also a client component - was the issue that you wanted to isolate the context from the ApplyForm? That doesn't seem necessary

frontend/src/app/api/applications/[applicationId]/attachments/handler.ts
```diff
@@ -0,0 +1,49 @@
...
+      applicationId,
+      session.token,
+      fileFormData,
+    );
+    if (res.status_code !== 200) throw new Error("Upload failed");
+
+    return NextResponse.json({
+      application_attachment_id: res.data?.application_attachment_id,
+    });
+  } catch (err) {
```
Comment (doug-s-nava) [inline frontend/src/app/api/applications/[applicationId]/attachments/handler.ts]: I think we've got a pretty good pattern set down for error handling in our route handlers - any reason not to use the pattern here? https://github.com/HHS/simpler-grants-gov/blob/5659e592a8f90c8c884bce5af01a16087a525aae/frontend/src/app/api/applications/start/handler.ts#L36

I'd be worried that the handling in this change could swallow 400s

frontend/src/components/applyForm/ApplyForm.tsx
```diff
@@ -54,8 +54,16 @@ const ApplyForm = ({
 
   const formObject = !isEmpty(formData) ? formData : savedFormData;
   const navFields = useMemo(() => getFieldsForNav(uiSchema), [uiSchema]);
-  let fields: JSX.Element[] = [];
 
+  if (!formSchema || !formSchema.properties || isEmpty(formSchema.properties)) {
```
doug-s-nava, line 58: 👍

frontend/src/components/applyForm/utils.tsx
```diff
@@ -128,6 +136,35 @@ export const determineFieldType = ({
...
+      item !== null &&
+      item.type === "string" &&
+      item.format === "uuid"
+    ) {
+      return "AttachmentArray";
+    }
+  }
+
+  // eslint-disable-next-line no-console
+  console.debug(
```
doug-s-nava: can we remove this or do you want to keep it in for future debugging?

frontend/src/components/applyForm/utils.tsx
```diff
@@ -146,19 +183,52 @@ export const getFieldSchema = ({
...
-      ...schema,
-    } as RJSFSchema;
-  } else if (definition) {
-    return getByPointer(formSchema, definition) as RJSFSchema;
+  if (!definition && !schema) {
+    throw new Error("Missing schema definition");
   }
-  return schema as RJSFSchema;
+
+  // If schema is provided, merge it on top of resolved pointer
```
doug-s-nava: I think the idea here is that there are sometimes fields defined by both a schema and a definition, and in those cases we need to walk through the paths of the definition to see if there are schema attributes at that path that should be applied to create the final schema. Is that right? Can you point to an example where that's happening so I can better understand?

frontend/src/components/applyForm/utils.tsx
```diff
@@ -509,3 +598,59 @@ export const getApplicationResponse = (
     return {};
   }
 };
+
+// getApplicationIdFromUrl
+export const getApplicationIdFromUrl = (): string | null => {
+  if (typeof window === "undefined") return null;
+  const match = window.location.pathname.match(
```
doug-s-nava: if only because it'll be easier to test, could use `usePathname` rather than `window.location.pathname`?

frontend/src/components/applyForm/utils.tsx
```diff
@@ -509,3 +598,59 @@ export const getApplicationResponse = (
...
+    /\/applications\/application\/([a-f0-9-]+)\/form\//,
+  );
+  return match?.[1] ?? null;
+};
+
+// getLabelComponent
```
doug-s-nava: can delete

frontend/src/components/applyForm/widgets/AttachmentUploadWidget.tsx
```diff
@@ -0,0 +1,203 @@
...
+  useEffect(() => {
+    if (deleteState?.success) {
+      setShowFile(false);
+      setFileName(null);
+      setAttachmentId(null);
+      onChange?.(undefined);
+    }
+  }, [deleteState, onChange]);
+
+  useEffect(() => {
```
doug-s-nava, line 77: I'm a bit lost on what this useEffect is doing, and what the significance of "Previously uploaded file" is, since that seems to drive some behavior later. Can you leave some comments?

frontend/src/components/applyForm/widgets/AttachmentUploadWidget.tsx
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    fileInputRef.current?.clearFiles();
+  };
+
+  const hasError = rawErrors.length > 0;
+  const describedBy = hasError ? `${id}-error` : undefined;
+  const isPreviouslyUploaded = fileName === "(Previously uploaded file)";
+
+  const isValidationWarning = (e: unknown): e is FormValidationWarning => {
```
doug-s-nava: this seems to be generally asserting whether something is an `Error`, right? Can we move this to a general utils file?

frontend/src/components/applyForm/widgets/MultipleAttachmentUploadWidget.tsx
```diff
@@ -0,0 +1,272 @@
...
+  useEffect(() => {
+    if (hasHydratedRef.current) return;
+
+    let parsedValue: string[] = [];
+
+    if (Array.isArray(initialValue)) {
+      parsedValue = initialValue as string[];
+    } else if (typeof initialValue === "string") {
+      try {
+        // casting doesn’t fully satisfy the linter because it treats schema as possibly any underneath
```
doug-s-nava, line 66: since this has come up in a couple places I wonder if it's worth digging into to try to fix, or if it's worth generally documenting somewhere that we're cool with working around it by disabling the linter

frontend/src/components/applyForm/widgets/MultipleAttachmentUploadWidget.tsx
```diff
@@ -0,0 +1,272 @@
...
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+        const parsed = JSON.parse(initialValue);
+        if (
```
doug-s-nava, line 69: with this logic, any case where the value comes in as an object, or comes in as a string but doesn't resolve as json to an array with all string elements, we'll set the parsedValue as an empty array. Just want to make sure that's the intention here and we're covering all our edge cases

frontend/src/components/applyForm/utils.tsx
doug-s-nava: can we make this its own component since it's functioning like one in any case?

frontend/src/components/applyForm/widgets/MultipleAttachmentUploadWidget.tsx
```diff
@@ -0,0 +1,272 @@
...
+      />
+
+      <input
+        type="hidden"
+        name={id}
+        value={JSON.stringify(uploadedFiles.map((f) => f.id))}
+      />
+
+      {uploadedFiles.length > 0 && (
+        <ul className="usa-list usa-list--unstyled margin-top-2">
```
doug-s-nava: this seems complex enough that it could be its own component

frontend/src/services/attachments/upload.ts
```diff
@@ -0,0 +1,27 @@
+import { AttachmentUploadResponse } from "src/types/attachmentTypes";
```
doug-s-nava: can we use `useClientFetch` for this?

myduong-navapbc: i went ahead and created an upload and delete hook; the upload hook uses the useclientfetch hook.
_(and will update the attachment card component to use both in a future PR)_

frontend/src/components/applyForm/widgets/MultipleAttachmentUploadWidget.tsx
myduong-navapbc: yeah that is a good callout,  it was intentional. The idea was to be defensive: for instance, if the value isnt a stringified array of all strings, we just fall back to an empty array instead of trying to handle something unexpected.

im sure we can make this better, let me know if you have other ideas

frontend/src/components/applyForm/widgets/MultipleAttachmentUploadWidget.tsx
myduong-navapbc: totally! I'll add it to https://github.com/HHS/simpler-grants-gov/issues/5760, to come up with a solution.

frontend/src/components/applyForm/widgets/AttachmentUploadWidget.tsx
myduong-navapbc: totally! this is basically trying to detect if there's a previously uploaded file and hydrate the component. If we can't match it to a known attachment, we show (previously uploaded file) just to indicate something's been uploaded before, even if we don't have the original file name. 

one of the scenarios is:
a user uploads a file here, saves, then leaves. 
comes back, and from the application page where the list of attachments are shown, deletes a file that is referenced in our sf424 form.
navigates back to the sf424 form, the name of the file would be lost, therefore, we put **(Previously uploaded file)**. this entire flow was not in the design so, i kind of made it up. looking for design input after this PR.

ill go back through the widget and put some comments in code

frontend/src/components/applyForm/utils.tsx
myduong-navapbc: my intent here is to resolve the full schema definition based on the definition path (like a JSON Pointer), even if the field is also partially defined in the schema prop. 

this allows us to support both deep schema references and give us the ability to override it in the ui schema

here's one of the places that i ran into this issue:
also, in address, authorized_representative & contact_person

```
// JSON Schema
{
  type: "object",
  properties: {
    authorized_representative: {
      $ref: "#/$defs/person_name"
    }
  },
  $defs: {
    person_name: {
      type: "object",
      properties: {
        first_name: { type: "string", title: "First name" },
        last_name: { type: "string", title: "Last name" }
      },
      required: ["first_name", "last_name"]
    }
  }
}
```

```
// UI Schema
[
  {
    type: "field",
    definition: "/properties/authorized_representative/properties/first_name",
    schema: {
      title: "Authorized rep's first name" 
    }
  }
]
```

in this case, we use the definition to traverse the schema (resolving $ref to person_name) to get the full field type and structure. Then, we merge the override from the schema block (example: the title aka helper text). This gives us a clean way to reuse schema definitions while still allowing field-level additions in the UI schema.

frontend/src/components/applyForm/utils.tsx
myduong-navapbc: ive created a hook to get the pathname since its used in multiple components

frontend/src/components/applyForm/utils.tsx
myduong-navapbc: i intended it to be for future debugging but i will remove it for now.

frontend/src/components/applyForm/utils.tsx
myduong-navapbc: i converted the component and fieldLabel component to one called DynamicFieldLabel.

frontend/src/app/[locale]/workspace/applications/application/[applicationId]/form/[appFormId]/page.tsx
Comment (myduong-navapbc) [reply frontend/src/app/[locale]/workspace/applications/application/[applicationId]/form/[appFormId]/page.tsx]: actually, i removed it and the context

api/src/form_schema/forms/sf424.py
```diff
@@ -240,29 +232,29 @@
...
             "maxItems": 3,
             "items": {
                 "type": "string",
                 "enum": [
-                    "A: state Government",
-                    "B: county Government",
-                    "C: city or Township Government",
+                    "A: State Government",
+                    "B: County Government",
+                    "C: City or Township Government",
```
chouinar, line 257: Changing this is correct (this was actually a mistake I made when renaming things), but you'll also need to fix the tests to use the correct values - just update the strings in the test setup in these two places:
https://github.com/HHS/simpler-grants-gov/blob/main/api/tests/src/form_schema/forms/test_sf424.py#L208
https://github.com/HHS/simpler-grants-gov/blob/main/api/tests/src/form_schema/forms/test_sf424.py#L228

chouinar (commented):
No issues with the form changes, just a note on fixing the test

doug-s-nava (dismissed):
this is looking good!

✅ ✅ ✅ ✅ 

Just a couple of notes:
* nice job on the prepopulation, but should the prepopulated fields be disabled?
* on the UEI field, it's being properly pre-populated, but when saved with the pre-populated value, it comes back with a "too long" validation error. Do we need to adjust the validation, or are the seeded values weird?

chouinar:
> * on the UEI field, it's being properly pre-populated, but when saved with the pre-populated value, it comes back with a "too long" validation error. Do we need to adjust the validation, or are the seeded values weird?

We can adjust the factory, we set the factory up before we knew the format of a UEI - if you want to do that in this PR change this line to be `uei = factory.Sequence(lambda n: f"UEI{n:09d}")` and that should make it always exactly 12-chars and pass the validation:
https://github.com/HHS/simpler-grants-gov/blob/main/api/tests/src/db/models/factories.py#L2569

Hopefully we don't have any API tests that expect a certain format

doug-s-nava (approved):

Issue #5811, opened 2025-07-29
Setup a job that creates an email suppression record


babebe:
### Summary

Create a job called `SyncSuppressedEmailsTask` that fetches suppressed email addresses from AWS SES using the `ListSuppressedDestinations` API and update the `suppressed_email` table in the database accordingly. 

1. Fetch suppressed emails based on most recent updated_at we have in the DB.
2. For each entry :
    - If email exists in `link_external_user` table.
       - If record exists and `last_updated_at` changed - update record
       - Otherwise - insert record

### Acceptance criteria

[ ] Task created
[ ] Tests created

PR #5864, opened 2025-08-05
[Issue #5811] Sync suppressed emails task

babebe:
Created a job called SyncSuppressedEmailsTask that fetches suppressed email addresses from AWS SES using the ListSuppressedDestinations API and update the suppressed_email table in the database accordingly.
This job runs as the first step the EmailNotificationTask, so we can catch new updates prior to sending emails again


api/src/task/notifications/email_notification.py
```diff
@@ -29,14 +31,21 @@ def __init__(
...
             notification_config = EmailNotificationConfig()
         self.notification_config = notification_config
+        if sesv2_client is None:
+            sesv2_client = SESV2Client()
+        self.sesv2_client = sesv2_client
 
     def run_task(self) -> None:
+        # Run the suppressed email job first
+        SyncSuppressedEmailsTask(db_session=self.db_session, sesv2_client=self.sesv2_client).run()
+
```
chouinar: Should put a configuration parameter to turn this job on/off when running like the other tasks below.

api/src/task/notifications/email_notification.py
```diff
@@ -29,14 +31,21 @@ def __init__(
...
+        sesv2_client: BaseSESV2Client | None = None,
     ) -> None:
         super().__init__(db_session)
         self.search_client = search_client
         if notification_config is None:
             notification_config = EmailNotificationConfig()
         self.notification_config = notification_config
+        if sesv2_client is None:
+            sesv2_client = SESV2Client()
+        self.sesv2_client = sesv2_client
```
chouinar: How about:
```suggestion
        self.sesv2_client = sesv2_client
```
If it's `None`, then this same logic is already in the Sync task, but this still lets you pass through for setting up unit tests.

api/src/task/notifications/sync_suppressed_emails.py
```diff
@@ -0,0 +1,68 @@
...
+            else:
+                self.db_session.add(
+                    SuppressedEmail(
+                        email=email,
+                        reason=destination.reason,
+                        last_update_time=destination.last_update_time,
+                    )
+                )
+
+        self.db_session.commit()
```
chouinar: Avoid commits, use `with db_session.begin()` blocks if possible. I like to follow a pattern of:
```py
def run_task(self) -> None:
     with db_session.begin():
            self.process_stuff()
```
Just to keep all the db session stuff in one place, and then the `process_stuff` function doesn't need to be tabbed over a bunch for whatever it does.

api/src/task/notifications/sync_suppressed_emails.py
```diff
@@ -0,0 +1,68 @@
...
+                existing.reason = destination.reason
+                existing.last_update_time = destination.last_update_time
+            else:
+                self.db_session.add(
+                    SuppressedEmail(
+                        email=email,
+                        reason=destination.reason,
+                        last_update_time=destination.last_update_time,
+                    )
+                )
```
chouinar: What about this structure?
```suggestion
            existing = suppression_map.get(email)
            
            suppressed_email_record = suppression_map.get(email)
            if not suppressed_email_record:
                suppressed_email_record = SuppressedEmail(email=email)
                
            suppressed_email_record.reason = destination.reason
            suppressed_email_record.last_update_time = destination.last_update_time
            self.db_session.add(suppressed_email_record)
```
Note that it's fine to call `db_session.add` on something that already exists in the DB, that just tells SQLAlchemy to be aware of it and adds it to its "state" if not already there.

api/src/adapters/aws/sesv2_adapter.py
```diff
@@ -22,22 +25,23 @@ class SESV2Response(BaseModel):
...
-        self.client = boto3.client("sesv2")
+        self.client = get_sesv2_client()
 
     def list_suppressed_destinations(self, start_date: datetime | None = None) -> SESV2Response:
         request_params = {}
         if start_date:
             request_params["StartDate"] = start_date
 
         response = self.client.list_suppressed_destinations(**request_params)
+        response_object = SESV2Response.model_validate(response["SuppressedDestinationSummaries"])
```
chouinar: Because it always seems to cause issues when parsing other AWS/boto responses, what does a response look like if it's an error? Lets say we passed in nonsense to the API and it had an error response, what would we get back here?

While we probably still want an exception thrown, we probably don't want a key error. Maybe something like:
```suggestion
        if "SuppressedDestinationSummaries" in response:
            response_object = SESV2Response.model_validate(response["SuppressedDestinationSummaries"])
       else:
            # do some parsing on whatever it actually returned
            raise SomeError("...")
```

api/src/task/notifications/sync_suppressed_emails.py
```diff
@@ -0,0 +1,64 @@
...
+class SyncSuppressedEmailsTask(Task):
+    def __init__(self, db_session: db.Session, sesv2_client: BaseSESV2Client | None = None) -> None:
+        super().__init__(db_session)
+
+        self.sesv2_client = sesv2_client or SESV2Client()
+
+    def run_task(self) -> None:
+        # Get the most recent suppression timestamp from DB
+        stmt = select(SuppressedEmail).order_by(SuppressedEmail.last_update_time.desc()).limit(1)
+        with self.db_session.begin():
```
chouinar: This'll let you unindent all the code one level while still having the db_session.begin() wrapper. 

```suggestion
    def run_task(self) -> None:
        with self.db_session.begin():
            self.process_suppressed_emails()
            
   def process_suppressed_emails(self) -> None:
        # Get the most recent suppression timestamp from DB
        stmt = select(SuppressedEmail).order_by(SuppressedEmail.last_update_time.desc()).limit(1)
        ...
```

api/src/adapters/aws/sesv2_adapter.py
```diff
@@ -22,22 +25,23 @@ class SESV2Response(BaseModel):
 
 class BaseSESV2Client(ABC, metaclass=ABCMeta):
     @abstractmethod
-    def list_suppressed_destinations(self) -> SESV2Response:
+    def list_suppressed_destinations(self, start_date: datetime | None = None) -> SESV2Response:
```
chouinar: If this is a datetime and not a date, could we call this `start_time` instead of `start_date`? I know the underlying call we make calls it that, but that's not well named

api/src/adapters/aws/sesv2_adapter.py
```diff
@@ -22,22 +25,23 @@ class SESV2Response(BaseModel):
...
     def __init__(self) -> None:
-        self.client = boto3.client("sesv2")
+        self.client = get_sesv2_client()
 
     def list_suppressed_destinations(self, start_date: datetime | None = None) -> SESV2Response:
         request_params = {}
         if start_date:
             request_params["StartDate"] = start_date
 
         response = self.client.list_suppressed_destinations(**request_params)
```
chouinar: Something to test / follow-up on - I see that there is pagination support in this with a page size + next token. How large of pages can it return / do we need to support calling this paginated? Probably need something?

api/tests/src/task/notifications/test_sync_suppressed_emails.py
```diff
@@ -0,0 +1,88 @@
...
+
+@pytest.fixture
+def sync_suppressed_emails(db_session):
+    client = MockSESV2Client()
+    task = SyncSuppressedEmailsTask(db_session, client)
+    return task, client
+
+
+def test_sync_suppressed_emails_update(sync_suppressed_emails, db_session, enable_factory_create):
+    task, client = sync_suppressed_emails
```
chouinar: Split this into two fixtures and you won't need to unpack the tuple in each test. The `mock_client` passed to the `task` fixture and to the test are the exact same object.
```suggestion
@pytest.fixture
def mock_client():
    return MockSESV2Client()
    
@pytest.fixture
def task(db_session, mock_client):
    task = SyncSuppressedEmailsTask(db_session, client)
    return task

def test_sync_suppressed_emails_update(task, mock_client, db_session, enable_factory_create):
```

api/tests/conftest.py
```diff
@@ -396,6 +397,14 @@ def api_auth_token(monkeypatch, all_api_auth_tokens):
     return auth_token
 
 
+@pytest.fixture
+def mock_sesv2_client():
+    with patch("src.task.notifications.sync_suppressed_emails.SESV2Client") as MockSESV2Client:
+        mock_instance = MagicMock()
+        MockSESV2Client.return_value = mock_instance
+        yield mock_instance
+
```
chouinar: I had hoped that setting up the mock/regular client approach would allow us to not need to pipe this sort of thing through for all of our tests in the email code.

Would an approach like this work instead?

In the `__init__` function of the sync suppressed emails task, we do something like:

```py
self.sesv2_client = sesv2_client if sesv2_client else get_sesv2_client()
```

This function is defined in the sesv2 adapter module as:
```py
class SesConfig(PydanticBaseEnvConfig):
     use_mock_ses_client: bool = False

# You have something named this already for getting the actual boto client, so one would need to be renamed
def get_sesv2_client() -> BaseSESV2Client:
    config = SesConfig()
    if config.use_mock_ses_client:
         return MockSESV2Client()
    return SESV2Client()
```
Then in local.env we set `USE_MOCK_SES_CLIENT=TRUE` and it _should_ work. What we effectively setup is that by default we use the real client but locally we change it to the mock client.

babebe: This is a better setup! Thanks!!

api/src/adapters/aws/sesv2_adapter.py
babebe: If it errors a ClientError is raised. Updated with try/except.

api/src/adapters/aws/sesv2_adapter.py
babebe: It is paginated and by default returns the first page. Can't find docs on how many will be returned per page by default. We are only searching since the last time we check so results should be small but since it is account level, supporting pagination is best in this case. Added pagination support

api/src/adapters/aws/sesv2_adapter.py
```diff
@@ -18,45 +23,93 @@ class SESV2Response(BaseModel):
...
+        if start_time:
+            results = [r for r in results if r.last_update_time >= start_time]
+        # Where the page begins in the list
+        start_index = int(next_token) if next_token else 0
+        # Where the page ends in the list
+        end_index = start_index + self.page_size
+        # Subset of suppressed destinations
+        page = results[start_index:end_index]
+        # Offset for the next page
+        next_token = str(end_index) if end_index < len(results) else None
```
chouinar: If the way we handle the pagination is by just doing multiple calls in our `list_suppressed_destinations` methods like you did above, it's fine if our mock doesn't do any sort of pagination. Under the hood the list method might make multiple calls, but that's fine to not mock since it doesn't affect the usage of our client.

api/src/adapters/aws/sesv2_adapter.py
```diff
@@ -18,45 +23,93 @@ class SESV2Response(BaseModel):
...
+    def list_suppressed_destinations(self, start_time: datetime | None = None) -> SESV2Response:
+        request_params: dict[str, datetime | str] = {}
+        if start_time:
+            request_params["StartDate"] = start_time
+
+        all_summaries = []
+        next_token = None
+
+        try:
+            while True:
```
chouinar, line 49: As a safety net, could we put some way for this to eventually break out after like 100 iterations?

Just something like:
```py
iterations = 0
while True:
     iterations += 1

     # existing logic

     if iterations > 100:
         logger.error("Stopping iteration of response from list suppression API after 100 iterations")
         break
```
It's unlikely, but the alternative is the code endlessly calling if there were some edge case we weren't aware of.

api/src/adapters/aws/sesv2_adapter.py
babebe: That's true. Removed test also.

PR #5866, opened 2025-08-06
[Issue #5779] Create api keys table

### LINKED ISSUE #5779: BE – Add api_keys table & model
### Summary

| Column     | Type      | Constraints    |
| ---------- | --------- | -------------- |
| token\_id  | UUID      | PK             |
| key\_name  | VARCHAR   | not null       |
| key\_id    | VARCHAR   | AWS identifier |
| user\_id   | UUID      | FK users(id)   |
| last\_used | TIMESTAMP | nullable       |
| is\_active | BOOLEAN   | default true   |

### Acceptance criteria

- [ ] Migration generated & applied in all environments.
- [ ] SQL schema documented.
- [ ] Unit tests cover create/read/update/deactivate.

btabaska:
## Summary

Fixes #5779  

## Changes proposed

Created api keys table and factory


api/src/db/models/user_models.py
```diff
@@ -244,3 +248,18 @@ class SuppressedEmail(ApiSchemaTable, TimestampMixin):
...
     reason: Mapped[str]
     last_update_time: Mapped[datetime] = mapped_column(index=True)
+
+
+class UserApiKey(ApiSchemaTable, TimestampMixin):
+    """API Key table for user authentication to the API"""
+
+    __tablename__ = "user_api_key"
+
+    token_id: Mapped[uuid.UUID] = mapped_column(UUID, primary_key=True, default=uuid.uuid4)
```
chouinar: General pattern we follow is that the primary key ID should be `<table_name>_id` - makes stuff like foreign keys and writing queries by hand slightly easier to remember.

api/src/db/models/user_models.py
```diff
@@ -244,3 +248,18 @@ class SuppressedEmail(ApiSchemaTable, TimestampMixin):
...
+    key_name: Mapped[str] = mapped_column(nullable=False)
+    key_id: Mapped[str] = mapped_column(nullable=False, comment="AWS API Gateway key identifier")
```
chouinar: FYI - the type that goes in `Mapped` can be used instead of specifying nullable/the type in the `mapped_column` function. It's fine to do just: `key_name: Mapped[str]` and it'll know that is a text column (the type string maps to) and not nullable.

api/src/db/models/user_models.py
```diff
@@ -244,3 +248,18 @@ class SuppressedEmail(ApiSchemaTable, TimestampMixin):
...
+    user_id: Mapped[uuid.UUID] = mapped_column(UUID, ForeignKey("api.user.user_id"), index=True)
```
chouinar: Having to use a string in the ForeignKey bit is only needed if there's a circular dependency of the types - it works the same either way
```suggestion
    user_id: Mapped[uuid.UUID] = mapped_column(UUID, ForeignKey(User.user_id), index=True)
```

api/src/db/models/user_models.py
btabaska: Updated based on feedback. changed token_id to api_key_id

api/src/db/models/user_models.py
btabaska: Updated based on feedback

api/src/db/models/user_models.py
btabaska: Merged suggestion

chouinar (approved):

PR #5876, opened 2025-08-06
[Unticketed] Change the GET Application query to load explicitly

chouinar:
## Changes proposed
Changes how we do the GET Application query used in most of our endpoints to be very explicit with which relationships we load.

## Context for reviewers
I noticed in dev that the GET application endpoint (and others that hit this get app logic) were taking more than a second. When digging into the New Relic trace, I saw it was querying every table multiple times.

I removed the `selectinload("*")` entirely because although we can use `lazyload` to remove some relationships from this join, that wasn't quite working. For example, the application_form table has a relationship back to the application table. If we just have `*` then even if we removed the `application.competition.opportunity`, we didn't remove `application.application_forms[0].application.competition.opportunity` or other route in our relationships to get back to opportunity and/or other tables that blow up the size of this query.

We could have tried to remove these various other paths, but it seemed like less work to just set the paths we want instead.

Note that we do this selectinload for two reasons:
* Performance (it's faster to fetch things all at once)
* The API logic to make the JSON response needs everything fetched before we can return, if we haven't loaded everything, we'll get an error in the response.

That second one is how I made sure we have all the fields we use in the various endpoints, otherwise this would have errored a bunch of tests.

## Validation steps
Was able to see the query by adding `echo=True` to the Postgres DB client and then did a bit of log parsing.

Here is every table we queried before (some multiple times):
```
opportunity_summary
user_saved_opportunity
opportunity_change_audit
link_opportunity_summary_funding_category
user_opportunity_notification_log
competition_instruction
current_opportunity_summary
form
application_attachment
user
organization_user
link_competition_open_to_applicant
opportunity_attachment
application_submission
link_opportunity_summary_applicant_type
application
application_form
opportunity
competition
application_user
user_saved_search
opportunity_assistance_listing
competition_form
link_opportunity_summary_funding_instrument
opportunity_version
```

Here is what it is now queried:
```
application
application_user
opportunity_assistance_listing
competition_form
application_form
competition_instruction
application_attachment
link_competition_open_to_applicant
user
form
competition
```

### DIFF
--- a/api/src/services/applications/get_application.py
+++ b/api/src/services/applications/get_application.py
@@ -1,12 +1,18 @@
 from uuid import UUID
 
 from sqlalchemy import select
-from sqlalchemy.orm import lazyload, selectinload
+from sqlalchemy.orm import selectinload
 
 import src.adapters.db as db
 from src.api.response import ValidationErrorDetail
 from src.api.route_utils import raise_flask_error
-from src.db.models.competition_models import Application, ApplicationForm, Competition
+from src.db.models.competition_models import (
+    Application,
+    ApplicationForm,
+    Competition,
+    CompetitionForm,
+    Form,
+)
 from src.db.models.entity_models import Organization
 from src.db.models.user_models import ApplicationUser, User
 from src.services.applications.application_validation import (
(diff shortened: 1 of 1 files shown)


mdragon (approved):
Looks good

PR #5885, opened 2025-08-07
[Issue #5795] Add service to create API key in DB

### LINKED ISSUE #5795: BE – Repository helper: create_api_key()
### Summary

Added a helper function to create an API key in the DB table 

### Acceptance criteria

- [ ] Inserts new record; returns ApiKey instance.
- [ ] Enforces unique (user_id, name); raises DuplicateKeyError.
- [ ] Test: factory creates key; duplicate raises error.

btabaska:
## Summary

Fixes #5795 

## Changes proposed

Added service to create API key in DB
Added utility to create API key


api/src/services/users/create_api_key.py
```diff
@@ -0,0 +1,63 @@
...
+            key_id=key_id,
+            is_active=is_active,
+        )
+
+        db_session.add(api_key)
+
+        logger.info(
+            "Created new API key",
+            extra={
+                "api_key_id": api_key.api_key_id,
```
chouinar: This will always be null when you log it, anything that has a generated default (like our UUID primary keys) doesn't get set until it flushes to the DB.

Alternatively, if you just set it yourself by doing:
```py
api_key = UserApiKey(api_key_id=uuid.uuid4(), ...)
```
Then it'll be set.

api/src/services/users/create_api_key.py
```diff
@@ -0,0 +1,63 @@
...
+            UserApiKey.key_name == key_name,
+        )
+    ).scalar_one_or_none()
+
+    if existing_key:
+        logger.warning(
+            "Attempted to create duplicate API key",
+            extra={"user_id": user_id, "key_name": key_name},
+        )
+        raise DuplicateKeyError(f"API key with name '{key_name}' already exists for user {user_id}")
```
chouinar: A few questions:
* Is the key name something we need to be unique for a given user? If so, we should add a uniqueness constraint to the table - I wouldn't necessarily see having two keys for one user with the same name as something problematic though, should we restrict this at all?
* If this scenario is hit, is that user error (eg. a 4xx) or a system error (eg. a 5xx)? If it's a user thing, don't log a warning. Warnings/errors are for flagging issues to our developers.
*

api/src/util/api_key_gen.py
```diff
@@ -0,0 +1,7 @@
+import secrets
+import string
+
+
+def generate_api_key_id(length: int = 25) -> str:
+    alphabet = string.ascii_letters + string.digits  # a-z, A-Z, 0-9
+    return "".join(secrets.choice(alphabet) for _ in range(length))
```
chouinar, line 7: However unlikely, I don't think this guarantees a key will be unique. We'll need to check that it doesn't already exist in the DB where we use this.

Actually, double-checking the user_api_key schema, we don't enforce that key_id should be unique + indexed, which we should be doing.

api/src/services/users/create_api_key.py
```diff
@@ -0,0 +1,63 @@
...
+    """Raised when attempting to create an API key with a duplicate (user_id, key_name) combination."""
+
+    pass
+
+
+def create_api_key(
+    db_session: db.Session,
+    user_id: UUID,
+    key_name: str,
+    is_active: bool = True,
```
chouinar: I know this will sound counter-intuitive, but I'd recommend not building these utilities before building the endpoints.

The rough rough implementation for something like a create API key route will be:

```py
# in the route file
def api_key_create(db_session, json_data):
     # NOTE - except for grabbing the user + starting a DB session, we keep pretty much all logic out of the route level

    token_session = api_jwt_auth.get_user_token_session()
    user = token_session.user
     with db_session.begin():
             api_key = create_api_key(db_session, user, json_data)

# In a service
def create_api_key(db_session, user, json_data):
    # if there are enough params in a JSON request, parsing them into a Pydantic object can make things easier - not required, but convenient
    params = CreateApiKeyParams.model_validate(json_data)

    # probably some auth stuff here
    
    # actually create the API key here
    api_key = UserApiKey(....)
    db_session.add(api_key)
    return api_key
```

While I've excluded some of what you had in this function here like logging, that service method and the one you made here are the same, but the params passed in are different (request params as a raw dictionary versus individual params).

Unless we think we'll use this service logic in a lot of different places besides these API endpoints, breaking those out into dedicated utilities isn't strictly necessary (I could maybe see some CLI utils for local development wanting that, but not thinking of many others?)

---
Not looking for you to refactor this all right now, BUT building the API endpoints as stubs (request/response schema stuff, static response) before the actual backing implementation can help set this up a bit.

api/src/services/users/create_api_key.py
btabaska: Good point that I suppose it is a non-issue if a user decides to name their keys all the same name. Seems odd to me, but the behavior shouldn't be an issue that we check for. Removing this code.

api/src/services/users/create_api_key.py
btabaska: I'll just set it myself as suggested, this makes sense.

api/src/util/api_key_gen.py
btabaska: Great callout, I am going back in and adding a uniqueness check with a max retry. I'll also go ahead and update the schema for the table in this PR to resolve the second issue.

api/src/services/users/create_api_key.py
btabaska: Good point, I went through and refactored the PR based on this feedback and implemented the route as well. Let me know if my approach is correct, as it may not be.

btabaska:
Updated PR based on all initial feedback. Ready for second pass checks. 

Also unsure if the database migration has been applied to prod from the last PR. I updated the existing migration, but I could move the changes into a new migration if that is more correct.

api/src/db/migrations/versions/2025_08_06_add_user_api_key_table.py
```diff
@@ -47,11 +47,15 @@ def upgrade():
     op.create_index(
         op.f("user_api_key_user_id_idx"), "user_api_key", ["user_id"], unique=False, schema="api"
     )
+    op.create_index(
+        op.f("user_api_key_key_id_idx"), "user_api_key", ["key_id"], unique=True, schema="api"
+    )
```
chouinar: Can't modify an existing migration once merged. While you asked if it has gotten to prod, as soon as you merged it, it went to dev/staging, so this change would not be applied there.

api/src/api/users/user_routes.py
```diff
@@ -452,3 +455,52 @@ def user_update_saved_search(
...
+@user_blueprint.input(UserApiKeyCreateRequestSchema, location="json")
+@user_blueprint.output(UserApiKeyCreateResponseSchema)
+@user_blueprint.doc(responses=[200, 400, 401, 403])
+@user_blueprint.auth_required(api_jwt_auth)
+@flask_db.with_db_session()
+def user_create_api_key(
+    db_session: db.Session, user_id: UUID, json_data: dict
+) -> response.ApiResponse:
+    """Create a new API key for the authenticated user"""
+    logger.info("POST /v1/users/:user_id/api-keys")
```
chouinar: If you do this, every log for the rest of the request's lifecycle will have user_id automatically included.

```suggestion
) -> response.ApiResponse:
    """Create a new API key for the authenticated user"""
    add_extra_data_to_current_request_logs({"user_id": user_id})
    logger.info("POST /v1/users/:user_id/api-keys")
```

api/src/services/users/create_api_key.py
```diff
@@ -0,0 +1,93 @@
...
+        if not self.key_name:
+            raise ValueError("key_name is required")
+
+        # Use the same validator as the schema
+        length_validator = validators.Length(min=1, max=255)
+        try:
+            length_validator(self.key_name)
+        except Exception as e:
+            raise ValueError(f"key_name validation failed: {str(e)}") from e
+
```
chouinar: Try to avoid custom validation unless it's something we can't do in our Marshmallow schemas (those rules do run on the request passed in to us). Otherwise you'll need to make sure that the way we format the error details in a way consistent with how we do it elsewhere (which would be very complex).

Since the request schema already validates this, just leave it out here. Since the only path to this code IS the API, we are fine assuming what comes in is valid - if not, that'll raise an exception somewhere, and 500 alerting us.

api/src/api/users/user_routes.py
```diff
@@ -452,3 +455,52 @@ def user_update_saved_search(
...
+            extra={"user_id": user_id, "error": str(e)},
+        )
+        raise_flask_error(400, str(e))
+
+    except Exception as e:
+        logger.error(
+            "Unexpected error creating API key",
+            extra={"user_id": user_id, "error": str(e)},
+        )
+        raise_flask_error(500, "Internal server error")
```
chouinar: Shouldn't need this try/except. The validation issue that you check for is already checked by the Marshmallow schema, and if there is an exception thrown in our code that is uncaught, that already turns into a 500 + logs it unless it's raised by `raise_flask_error(4xx)`

There's a whole mess of code around exceptions/errors, but generally summarizing the behavior in the API that you can assume:
* `raise_flask_error(4xx, ...)` will raise an error with the 4xx code you pass it, and you can pass other details back to the user
* Any uncaught exception automatically get turned into a 500 response to the user and logs an error. We have alerts setup if an error logs are produced, so it's fine to throw exceptions for scenarios that should alert someone.

This means that unless an exception is a somewhat expected case / part of logic (eg. using try/except as a sort of if-statement like with checking if something is an integer), you can ignore a lot of exception handling in our endpoints. They're setup to be resilient and alert us when things go weird.

api/src/db/migrations/versions/2025_08_06_add_user_api_key_table.py
btabaska: That is what I thought, thanks for explaining. I migrated the changes into a new migration and reverted these from the old migration.

api/src/services/users/create_api_key.py
btabaska: This makes sense, thanks for explaining. I went in and removed the custom validation across the PR to simplify the code. This update was also made in user_routes.py since we won't need the validation there either. I went in and updated the unit tests to no longer include tests for this validation.

api/src/api/users/user_routes.py
btabaska: As noted above, thanks for this explanation. I went ahead and removed the try/except. I wasn't aware of the schema so that is really helpful to understand. I went ahead and simplified the code down significantly.

btabaska:
All feedback has been iterated on and the PR is ready for third pass review.

api/tests/src/services/users/test_create_api_key.py
```diff
@@ -0,0 +1,248 @@
...
+from src.adapters import db
+from src.services.users.create_api_key import (
+    MAX_KEY_GENERATION_RETRIES,
+    KeyGenerationError,
+    create_api_key,
+)
+from tests.src.db.models.factories import UserApiKeyFactory, UserFactory
+
+
+def test_create_api_key_success(enable_factory_create, db_session: db.Session):
```
chouinar, line 15: Missed it on the prior PR - could we make some API tests so we're testing the API functionality? Or at least add a few api tests in addition to these (at least for basic success + validation issue + auth cases).

That would basically be:
* Add tests to tests/src/api/users folder
* Instead of calling `create_api_key` call the endpoint using the test client
* Can rely on the user + user_auth_token fixtures for setting up the user + auth token for you and just pass those in

So, if you wanted an equivalent happy-path test, something like
```py
def test_create_api_key_success(enable_factory_create, db_session: db.Session, client, user, user_auth_token):
    json_data = {"key_name": "Test API Key"}
    
    response = client.post(
        f"/v1/users/{user.user_id}/api-keys",
        headers={"X-SGG-Token": user_auth_token},
        data=json_data,
    )
    assert response.status_code == 200

    response_data = response.json["data"]
    # whatever checks you want to do
    assert response_data["user_id"] == user.user_id
    ...
    
    # If you want to verify it's in the DB too
    api_key = db_session.execute(select(UserApiKey).where(UserApiKey.api_key_id == response_data["api_key_id"])).scalar_one_or_none()
    # whatever checks you want to do
    assert api_key is not None
```

btabaska: Thanks for the boilerplate. I went ahead and added tests for the API including the suggested one above.

PR #5887, opened 2025-08-07
[Issue #5797] Add service to get and return API keys from logged in user in DB

### LINKED ISSUE #5797: BE – Repository helper: list_api_keys()
### Summary

Added a helper function to get all keys for the currently logged in user from the API Keys DB table for that user

### Acceptance criteria

- [ ] Returns keys for the currently logged in user ordered created DESC.
- [ ] Test covers

btabaska:
## Summary

Fixes #5797 

## Changes proposed

Added service to get and return all user API keys


api/tests/src/services/users/test_get_user_api_keys.py
```diff
@@ -0,0 +1,71 @@
...
+
+    api_key1 = UserApiKeyFactory.create(
+        user=user, key_name="First Key", created_at=datetime.utcnow() - timedelta(hours=2)
+    )
+    api_key2 = UserApiKeyFactory.create(
+        user=user, key_name="Second Key", created_at=datetime.utcnow() - timedelta(hours=1)
+    )
+    api_key3 = UserApiKeyFactory.create(
+        user=user, key_name="Third Key", created_at=datetime.utcnow()
+    )
```
chouinar: datetime.utcnow() is deprecated. We have a datetime_util with a utcnow() function that uses the newer way of setting this.

Won't have any meaningful difference except using the deprecated method puts warnings in our test output: https://github.com/HHS/simpler-grants-gov/actions/runs/16806507662/job/47600224925#step:8:191


```suggestion
    api_key1 = UserApiKeyFactory.create(
        user=user, key_name="First Key", created_at=datetime_util.utcnow() - timedelta(hours=2)
    )
    api_key2 = UserApiKeyFactory.create(
        user=user, key_name="Second Key", created_at=datetime_util.utcnow() - timedelta(hours=1)
    )
    api_key3 = UserApiKeyFactory.create(
        user=user, key_name="Third Key", created_at=datetime_util.utcnow()
    )
```

btabaska:
Updated PR based on feedback and fully implemented endpoints.

api/src/api/users/user_routes.py
```diff
@@ -490,3 +492,32 @@ def user_create_api_key(
     )
 
     return response.ApiResponse(message="Success", data=api_key)
+
+
+@user_blueprint.get("/<uuid:user_id>/api-keys")
```
chouinar: Do we want to support filtering in this endpoint at some point? If so, I'd make this a POST so we can have a request body, even if we won't have any filters yet.

api/src/api/users/user_routes.py
```diff
@@ -490,3 +492,32 @@ def user_create_api_key(
...
+    with db_session.begin():
+        api_keys = get_user_api_keys(db_session, user_id)
+
+    logger.info(
+        "Retrieved API keys for user",
+        extra={
+            "user_id": user_id,
+            "api_key_count": len(api_keys),
+        },
+    )
```
chouinar: You have this log message here and in the `get_user_api_keys` method - can we remove one?

btabaska: Thanks for the callout. I should have caught that. Updated to remove the duplicate log.

api/src/api/users/user_routes.py
btabaska: I don't know if we ever will need to support filtering, but it doesn't really cost anything to make this update now. I just pushed the change to support POST instead of get.

btabaska:
Updated based on feedback. Ready for next pass review.

chouinar (approved):

PR #5889, opened 2025-08-07
[Unticketed] Add form schema to app form API model

chouinar:
## Changes proposed
Add form the the API schema for an application form

## Context for reviewers
I thought we already did this, but it turns out we don't return the form from the application form endpoint. When talking with @doug-s-nava - realized the frontend was having to do a roundabout way of getting the form's schema on the app form page, this should help fix that weird logic to get it via a competition.

### DIFF
--- a/api/openapi.generated.yml
+++ b/api/openapi.generated.yml
@@ -3806,6 +3806,10 @@ components:
           type: string
           format: uuid
           example: 123e4567-e89b-12d3-a456-426614174000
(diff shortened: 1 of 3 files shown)


doug-s-nava (approved):
tested with Postman, and the payload comes back with all the info the frontend needs to eliminate the call to the competitions endpoint

✅ ✅ ✅ ✅

Issue #5859, opened 2025-08-05
Testing Sam.gov integration


chris-kuryak:
### Summary

* Verify that running the extract fetching step will behave correctly now that we're on the next month.
* Verify that the performance improvements made keep the job working when processing a monthly extract.
* Verify logic for setting up organizations works correctly.
* Ideally some of this testing would be possible in our lower environments so we don't need to test in production.

### Acceptance criteria

TBD

chris-kuryak:
Pointing at 5 to be conservative

PR #5895, opened 2025-08-08
[Issue #5859] Modify sam extract fetching to not refetch pending extracts

chouinar:
## Summary
Work for #5859

## Changes proposed
When checking if we already fetched an extract, factor in pending extracts

## Context for reviewers
There are 3 extract status values with the following purpose:
* `pending` - We just downloaded this - the process extract task fetches everything in this status
* `completed` - We've processed this - the process extract task puts everything into this after running
* `failed` - Nothing sets this status, if we had an issue processing a file, we could move it to this status manually

Before this change it only considered a file in the `completed` status as valid for seeing if we already handled a file. But that means if we rerun the fetch process two times in a row, we end up creating duplicate records in our DB which would be bad as we'd then try to process every file twice. To protect against that, we want pending and completed processing statuses to both keep us from fetching an extract again. Only failed should be ignored and re-downloaded.


mdragon (approved):
Approved

Issue #5238, opened 2025-06-03
Move User records from API to Analytics DB


mdragon:
### Summary
Move non-identifying user record data from Operational to Analytics DB to enable reporting in Metabase

Example query for one-time charts in #4615 
```
select count(*) accounts, created_at::date created_on 
from api.user 
where created_at::date < now()::date
group by created_at::date order by created_on desc
```
But likely we want to move more raw data like user_id and creation_at over at a row-by-row level and then do this bucketing via Metabase. As long as we don't ship identifying data like email address (which is safely tucked away in another table) we don't have to pre-rollup or obscure the data before getting it to analytics. This will help with other analysis we might do about users later like, reporting on time from first login to saving an opportunity or otherwise engaging, etc.

### Acceptance criteria

- [ ] User data is being moved from API DB to Analyics DB

eric-serrato95:
Bumping out to Past July 19, as this should not take priority over emails and other features at the moment

eric-serrato95:
No PII in this information

joshtonava:
@eric-serrato95 when this deploys to lowers we should be able to see user data in metabase to UAT this.

PR #5904, opened 2025-08-11
[Issue #5238] Move user records from api to analytics db

joshtonava:
## Summary

Work for #5238 

## Changes proposed

- Added new `user` table to analytics database with migration `0012_add_user_tables.sql`
- Updated analytics constants to include `USER` table and column definitions
- Added user table to CSV extraction process with explicit column ordering for data safety
- Updated test configuration and data to support the new user table

## Context for reviewers

This PR adds basic user tracking to the analytics database by creating a new `user` table that stores user_id, created_at, and updated_at timestamps. The table includes performance indexes on both timestamp fields for efficient date range queries.

## Validation steps

To verify data ingestion works locally:
- in the `api` folder
  - ensure the db is setup and seeded correctly `make remake-backend`
  - make a folder `tmp`
  - Edit the `local.env` field `API_ANALYTICS_DB_EXTRACTS_PATH` to be `./tmp`
  - Run `make cmd args="task create-analytics-db-csvs"` which should generate the output csvs
  - Stop the api project `docker compose down --volumes`
- Copy or move the `tmp` folder in the `api` folder to the `analytics` folder
- In the `analytics` folder:
  - Edit the `local.env` field `API_ANALYTICS_DB_EXTRACTS_PATH` to be `./tmp` 
  - run `docker-compose up -d` to start the project
  - run `make db-migrate` to apply the migrations
  - run `make opportunity-load` to load the opportunity data from the `tmp` folder.

To verify the tests work `make test-audit` in the `analytics` folder


api/src/task/analytics/create_analytics_db_csvs.py
```diff
@@ -96,9 +97,13 @@ def generate_csv(self, table: sqlalchemy.Table) -> None:
         cursor = self.db_session.connection().connection.cursor()
         schema = table.schema if self.config.db_schema is None else self.config.db_schema
 
-        with cursor.copy(
-            f"COPY {schema}.{table.name} TO STDOUT with (DELIMITER ',', FORMAT CSV, HEADER TRUE, FORCE_QUOTE *, encoding 'utf-8')"
-        ) as cursor_copy:
+        # Use explicit column ordering for user table to ensure column safety
+        if table.name == "user":
+            copy_query = f"COPY (SELECT user_id, created_at, updated_at FROM {schema}.{table.name} ORDER BY created_at) TO STDOUT with (DELIMITER ',', FORMAT CSV, HEADER TRUE, FORCE_QUOTE *, encoding 'utf-8')"
```
joshtonava, line 102: For doing the rest of the files Im thinking of doing something simple and just setting up a config of constants like used in analytics. Its not dynamic but it prevents data changes from breaking the analytics system and requires the user make sure columns are added to both sides of the system when needed.

analytics/tests/integrations/extracts/test_load_opportunity_data.py
```diff
@@ -58,42 +58,65 @@ def test_extract_copy_opportunity_data(
 
     # Verify that the data was inserted into the database
     with conn.begin():
-        lk_opp_sts_result = conn.execute(
```
joshtonava, line 61: There was a lint rule enforcing max number of local variables, hence the series of changes.

api/src/task/analytics/create_analytics_db_csvs.py
```diff
@@ -96,9 +97,13 @@ def generate_csv(self, table: sqlalchemy.Table) -> None:
         cursor = self.db_session.connection().connection.cursor()
         schema = table.schema if self.config.db_schema is None else self.config.db_schema
 
-        with cursor.copy(
-            f"COPY {schema}.{table.name} TO STDOUT with (DELIMITER ',', FORMAT CSV, HEADER TRUE, FORCE_QUOTE *, encoding 'utf-8')"
-        ) as cursor_copy:
+        # Use explicit column ordering for user table to ensure column safety
+        if table.name == "user":
+            copy_query = f"COPY (SELECT user_id, created_at, updated_at FROM {schema}.{table.name} ORDER BY created_at) TO STDOUT with (DELIMITER ',', FORMAT CSV, HEADER TRUE, FORCE_QUOTE *, encoding 'utf-8')"  # nosec B608
```
joshtonava: For doing the rest of the files Im thinking of doing something simple and just setting up a config of constants like used in analytics. Its not dynamic but it prevents data changes from breaking the analytics system and requires the user make sure columns are added to both sides of the system when needed.

let me know if something more brilliant comes to mind

joshtonava: Also im getting a lint security error by specifying the columns to select.

analytics/src/analytics/integrations/etldb/migrations/versions/0012_add_user_tables.sql
```diff
@@ -0,0 +1,18 @@
...
+
+CREATE TABLE IF NOT EXISTS "user"
+(
+    user_id    UUID                     NOT NULL
+        PRIMARY KEY,
+    created_at TIMESTAMP WITH TIME ZONE NOT NULL,
+    updated_at TIMESTAMP WITH TIME ZONE NOT NULL
+);
+
+-- Create index on created_at for performance when querying by date ranges
```
joshtonava, line 12: LMK if this is overkill

api/src/task/analytics/create_analytics_db_csvs.py
chouinar: Do we need the query to have an order by?

An idea that we can maybe evolve into a config, what if the place where we define the tables was instead formatted as table_name -> columns:
```py
TABLES_TO_EXTRACT = [
    ...
    "user_saved_search": ["*"],
    "user_saved_opportunity": ["*"],
    "user": ["user_id", "created_at", "updated_at"]
]
```
Then here we have something like:
```py
columns = ", ".join(TABLES_TO_EXTRACT[table.name])
copy_query = f"COPY (SELECT {columns} FROM ..."
```
For any of the existing ones, we'd just do `*` and can change that later.

api/src/task/analytics/create_analytics_db_csvs.py
chouinar, line 1: Could you update the test for this class to include user?

api/src/task/analytics/create_analytics_db_csvs.py
joshtonava: So you are saying * for now and then set the actual columns in the next ticket if im understanding correctly?

api/src/task/analytics/create_analytics_db_csvs.py
joshtonava: Sure thing!

joshtonava: Actually, this should be taken care of here:

image

api/src/task/analytics/create_analytics_db_csvs.py
joshtonava: and no we dont really need an order by

joshtonava: this is done now

chouinar (approved):

PR #5909, opened 2025-08-12
[Issue #5858] Handle empty response

### LINKED ISSUE #5858: Handle empty response translation when comparing S2S and Simpler data
### Summary

If an empty response is returned for the list of opportunities, grants.gov returns nothing, but we have an empty list, so the comparison fails - there is not a ticket for this.

### Acceptance criteria

S2S empty response and Simpler empty array are seen as "matching"

mikehgrantsgov:
## Summary

Fixes #5858 

## Changes proposed

Add `_normalize_soap_dict_for_comparison` to normalize a SOAP dictionary for comparison.
Apply this method where necessary, add tests.

## Context for reviewers

If an empty response is returned for the list of opportunities, grants.gov returns nothing, but we have an empty list, so the comparison fails.

This can be seen by setting up your local API to proxy a request to training and passing it an opportunity number that doesn't exist (a 404-like case). The comparison between our response and grants.gov's doesn't match. I'm assuming this is because we setup an "empty" response differently. We should fix them to match

## Validation steps

See new unit tests.

### DIFF
--- a/api/src/legacy_soap_api/legacy_soap_api_utils.py
+++ b/api/src/legacy_soap_api/legacy_soap_api_utils.py
@@ -130,6 +130,50 @@ def is_list_of_dicts(data: list) -> bool:
     return isinstance(data, list) and all(isinstance(item, dict) for item in data)
 
 
+def _normalize_soap_dict_for_comparison(soap_dict: dict, is_root_level: bool = True) -> dict:
+    """
+    Normalize a SOAP dictionary for comparison.
+
+    This function handles differences in empty response formatting between
+    Simpler and Grants.gov SOAP responses by:
+    1. Filtering out XML namespace attributes (@xmlns:*)
+    2. Normalizing OpportunityDetails field (missing/None -> empty list) only at root level
+
+    Args:
+        soap_dict: Dictionary representation of SOAP response
+        is_root_level: Whether this is the root level response dict
+
+    Returns:
+        Normalized dictionary for comparison
+    """
+    # Filter out namespace attributes and copy other keys
+    normalized = {
+        k: v for k, v in soap_dict.items()
+        if not k.startswith("@xmlns:")
+    }
+
+    # Handle OpportunityDetails normalization for empty responses - ONLY at root level
+    # This prevents adding OpportunityDetails to nested dictionaries
+    if is_root_level:
+        if "OpportunityDetails" not in normalized:
+            normalized["OpportunityDetails"] = []
+        elif normalized["OpportunityDetails"] is None:
+            normalized["OpportunityDetails"] = []
+
+    # Recursively normalize nested dictionaries (not at root level)
+    for key, value in normalized.items():
+        if isinstance(value, dict):
+            normalized[key] = _normalize_soap_dict_for_comparison(value, is_root_level=False)
+        elif isinstance(value, list) and value and isinstance(value[0], dict):
+            # Normalize list of dictionaries
+            normalized[key] = [
+                _normalize_soap_dict_for_comparison(item, is_root_level=False) if isinstance(item, dict) else item
+                for item in value
+            ]
+
+    return normalized
+
+
 def diff_soap_dicts(
     sgg_dict: dict, gg_dict: dict, key_indexes: dict | None = None, keys_only: bool = False
 ) -> dict:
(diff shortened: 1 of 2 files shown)


api/src/legacy_soap_api/legacy_soap_api_utils.py
```diff
@@ -130,52 +130,108 @@ def is_list_of_dicts(data: list) -> bool:
...
 
 
+def _normalize_soap_dict_for_comparison(soap_dict: dict, is_root_level: bool = True) -> dict:
+    """
+    Normalize a SOAP dictionary for comparison.
+
+    This function handles differences in empty response formatting between
+    Simpler and Grants.gov SOAP responses by:
+    1. Filtering out XML namespace attributes (@xmlns:*)
+    2. Normalizing OpportunityDetails field (missing/None -> empty list) only at root level
```
chouinar: @freddieyebra - For this ticket, the issue we want to solve is that if you call the SOAP proxy and get what would normally be a 404 (let's say you pass a random opportunity number), our API gives a slightly different response. Grants.gov returns nothing, while we return a list (effectively ` ` versus `{}`)

Would we want to just account for that in the diff OR would we also want to change the response we calculate as well? I think both would be helpful?

api/src/legacy_soap_api/legacy_soap_api_utils.py
```diff
@@ -130,52 +130,108 @@ def is_list_of_dicts(data: list) -> bool:
...
+    # Filter out namespace attributes and copy other keys
+    normalized = {k: v for k, v in soap_dict.items() if not k.startswith("@xmlns:")}
+
+    # Handle OpportunityDetails normalization for empty responses - ONLY at root level
+    # This prevents adding OpportunityDetails to nested dictionaries
+    if is_root_level:
+        if "OpportunityDetails" not in normalized:
+            normalized["OpportunityDetails"] = []
+        elif normalized["OpportunityDetails"] is None:
+            normalized["OpportunityDetails"] = []
```
chouinar: Should we do this on the object we return rather than just in the comparison logic? I'm betting we'll need this same behavior in a bunch of models in the future. Maybe something we can configure per model or just how the logic that builds a response works for GetOpportunityListResponse?

mikehgrantsgov: I lifted this up toward the `GetOpportunityListResponse`

api/src/legacy_soap_api/legacy_soap_api_utils.py
```diff
@@ -130,52 +130,108 @@ def is_list_of_dicts(data: list) -> bool:
     return isinstance(data, list) and all(isinstance(item, dict) for item in data)
 
 
+def _normalize_soap_dict_for_comparison(soap_dict: dict, is_root_level: bool = True) -> dict:
```
freddieyebra: The dicts for both simpler and the proxy soap responses should already be normalized. The `BaseSOAPClient.get_proxy_soap_response_dict` and `BaseSOAPClient.get_soap_response_dict` method both do the following:

1. Convert xml to xml dict (includes things like "@xmlns:app")
2. Validates and converts the dict to the pydantic schema
3. Compares the dicts of the pydantic schemas against each other.

We mostly need to ensure that the XML looks similar and that an empty response in simpler is denoted/structured the same as the proxy.

For example, the proxy response of no results is `<GetOpportunityListResponse></GetOpportunityListResponse>` where grants.gov response with a single tag `<GetOpportunityListResponse/>`

api/src/legacy_soap_api/legacy_soap_api_utils.py
freddieyebra: @chouinar Yeah, as i mentioned in my [comment](https://github.com/HHS/simpler-grants-gov/pull/5909/files#r2270788323) the dicts should already be normalized for the most part, but in the simpler dict we are getting:
```python
...
"GetOpportunityListResponse": null
...
```

And in the gg response we are resolving that to be 
```python
...
"GetOpportunityListResponse": {}
...
```

api/src/legacy_soap_api/legacy_soap_api_client.py
```diff
@@ -149,7 +149,23 @@ def get_simpler_soap_response(self, proxy_response: SOAPResponse) -> tuple:
...
+
+        # Use our normalization logic for the main comparison
+        diff_results = diff_soap_dicts(
+            sgg_dict=simpler_body,
+            gg_dict=proxy_body,
+            key_indexes=self.operation_config.key_indexes,
+            keys_only=True,
+        )
+
+        if diff_results == {}:
```
mikehgrantsgov: @freddieyebra I think this addresses your feedback?

freddieyebra: In the log_diffs method, we are doing the diff. I do not understand why we need a separate call to the diff method.

api/src/legacy_soap_api/legacy_soap_api_client.py
```diff
@@ -149,10 +149,8 @@ def get_simpler_soap_response(self, proxy_response: SOAPResponse) -> tuple:
         simpler_response_soap_dict = self.get_soap_response_dict()
 
         # We will only run diffs for responses that do not match.
-        if proxy_response_soap_dict == simpler_response_soap_dict:
-            logger.info("soap_api_diff responses match", extra={"soap_responses_match": True})
-        else:
-            self.log_diffs(proxy_response_soap_dict, simpler_response_soap_dict)
+        # Use normalized comparison to handle namespace attributes and empty response differences
+        self.log_diffs(proxy_response_soap_dict, simpler_response_soap_dict)
```
freddieyebra: I would leave this check in place so we can not run diff if not needed.

freddieyebra:
```suggestion
        if proxy_response_soap_dict == simpler_response_soap_dict:
            logger.info("soap_api_diff responses match", extra={"soap_responses_match": True})
        else:
            self.log_diffs(proxy_response_soap_dict, simpler_response_soap_dict)
```

api/src/legacy_soap_api/legacy_soap_api_utils.py
freddieyebra: Since the dicts are normalized already they should not include "@xmlns:". That would allow us to omit this normalization method all together.

api/src/legacy_soap_api/legacy_soap_api_client.py
mikehgrantsgov: Thanks, added this

Issue #5812, opened 2025-07-29
Update current email sending logic


babebe:
### Summary

Update email sending logic to exclude any users whose email exists in the `suppressed_email` table. 

`LEFT JOIN suppressed_email s ON u.email = s.email

WHERE s.email IS NULL`

Alternate query syntax (Sometimes one plan is much better than the other)

`WHERE NOT EXISTS(SELECT email from suppressed_email se WHERE se.email = link_external_user.email)`

### Acceptance criteria

[ ] Suppressed users excluded in all 3 email types
[ ] Add/Update Tests

PR #5922, opened 2025-08-12
[Issue #5812] Exclude suppressed emails form email notifications

babebe:
## Summary

Fixes / Work for #5812  

## Changes proposed

Update email sending logic to exclude any users whose email exists in the suppressed_email table.

### DIFF
--- a/api/src/task/notifications/closing_date_notification.py
+++ b/api/src/task/notifications/closing_date_notification.py
@@ -7,7 +7,12 @@ from sqlalchemy.orm import selectinload
 
 from src.adapters import db
 from src.db.models.opportunity_models import Opportunity, OpportunitySummary
-from src.db.models.user_models import UserOpportunityNotificationLog, UserSavedOpportunity
+from src.db.models.user_models import (
+    LinkExternalUser,
+    SuppressedEmail,
+    UserOpportunityNotificationLog,
+    UserSavedOpportunity,
+)
 from src.task.notifications.base_notification import BaseNotificationTask
 from src.task.notifications.config import EmailNotificationConfig
 from src.task.notifications.constants import NotificationReason, UserEmailNotification
@@ -73,6 +78,7 @@ class ClosingDateNotificationTask(BaseNotificationTask):
                         # == NotificationReason.CLOSING_DATE_REMINDER
                     )
                 ),
+                ~exists().where(SuppressedEmail.email == LinkExternalUser.email),
             )
         )
 
--- a/api/src/task/notifications/opportunity_notifcation.py
+++ b/api/src/task/notifications/opportunity_notifcation.py
@@ -3,14 +3,14 @@ from datetime import datetime
 from typing import Sequence, cast
 from uuid import UUID
 
-from sqlalchemy import and_, desc, func, select, tuple_, update
+from sqlalchemy import and_, desc, exists, func, select, tuple_, update
 from sqlalchemy.orm import aliased, selectinload
 
 from src.adapters import db
 from src.api.opportunities_v1.opportunity_schemas import OpportunityVersionV1Schema
 from src.constants.lookup_constants import FundingCategory, OpportunityCategory, OpportunityStatus
 from src.db.models.opportunity_models import OpportunityVersion
-from src.db.models.user_models import UserSavedOpportunity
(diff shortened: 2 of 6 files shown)


api/src/task/notifications/closing_date_notification.py
```diff
@@ -73,6 +78,7 @@ def collect_email_notifications(self) -> list[UserEmailNotification]:
                         # == NotificationReason.CLOSING_DATE_REMINDER
                     )
                 ),
+                ~exists().where(SuppressedEmail.email == LinkExternalUser.email),
```
joshtonava, line 81: **Question**: Do we also need to join the external user id to the user saved opportunity id or is that overkill?

```
.where(LinkExternalUser.user_id == UserSavedOpportunity.user_id)
```

api/src/task/notifications/opportunity_notifcation.py
```diff
@@ -232,7 +232,10 @@ def _get_latest_opportunity_versions(self) -> Sequence:
                     latest_versions_subq.c.rn == 1,
                 ),
             )
-            .where(UserSavedOpportunity.is_deleted.isnot(True))
+            .where(
+                UserSavedOpportunity.is_deleted.isnot(True),
+                ~exists().where(SuppressedEmail.email == LinkExternalUser.email),
```
joshtonava, line 237: **question**: same as above, do we also need to join on the `LinkExternalUser.user_id == UserSavedOpportunity.user_id`, or is this sufficient?

api/src/task/notifications/closing_date_notification.py
joshtonava: im also wondering if we need to care about the case sensitivity of the email fields

api/tests/src/task/notifications/test_opportunity_notification.py
```diff
@@ -1126,3 +1127,29 @@ def test_build_notification_content_all_changes(
...
+        user,
+    ):
+        """Test that the user notification does not pick up user on suppression_list"""
+        # create a suppressed email
+        user_1 = factories.LinkExternalUserFactory.create().user
+        factories.SuppressedEmailFactory(email=user_1.email)
+
+        # create opportunity
+        opp = factories.OpportunityFactory.create(is_posted_summary=True)
+        factories.UserSavedOpportunityFactory.create(
```
joshtonava, line 1144: **question:** Maybe I dont fully understand the test, is it intentional that you didnt link user_1 here, instead linking user from the fixture?

api/src/task/notifications/closing_date_notification.py
chouinar: I think it's fine to not use the user ID. The emails themselves are the problem. Theoretically if two users had the same email and it was suppressed, we'd exclude both anyways.

As for case sensitivity - we could maybe look at the data to see if that's an issue. If our DB says `AbC@example.com` and AWS always changes that to `abc@example.com` that could be a problem. @babebe - maybe worth throwing a bad email intentionally into the data and seeing how it comes back from AWS when we test this in the lower envs?

babebe: We are joining the LinkExternalUser.user_id here since the API will return emails on the account level (all envs), so this will ensure we are only checking for emails for the env we are running in.

api/tests/src/task/notifications/test_opportunity_notification.py
babebe: Ah. I actually did not need to create the user_1 since using the user fixture. Removed it.

api/src/task/notifications/closing_date_notification.py
babebe: As for the case sensitivity - it comes back as is , not changed. Hit the endpoint via CLI. 
`  aws sesv2 put-suppressed-destination --email-address bounCE@simulator.amazonses.com --reason BOUNCE
`
Returns : 
` [..., {
            "EmailAddress": "bounCE@simulator.amazonses.com",
            "Reason": "BOUNCE",
            "LastUpdateTime": "2025-08-13T16:56:10.438000-04:00"
        },
...]
`

joshtonava: So if its unchanged then we dont really need to worry about case sensitivity is what im gathering

joshtonava (approved):

Issue #5913, opened 2025-08-12
DB setup to store opportunity review status data


joshtonava:
### Summary

A table is needed to store the opportunity ids in review as well as there status

A table could be `opportunity_review_status` with fields:
  - `opportunity_id` (FK to opportunity table)
  - `omb_review_status` (enum: Reviewable, Returned, Approved, N/A, blank)
  - `last_upd_date` (from grants.gov view)
  - Standard audit fields

**For discussion**: should this just be a part of an existing table or represented as a join table?

### Acceptance criteria

- Create table to store the opportunity review status
- Create database migration
- Add appropriate indexes for performance

babebe:
do we need these fields? 
1. `omb_review_status_date`: what format is it ? Maybe can be added later ? 
2. `last_upd_date`: what format is it ? Maybe can be added later ? 

Since this will be used to filter out opportunities in different processes keep it very simple ?

joshtonava:
> do we need these fields?
> 
> 1. `omb_review_status_date`: what format is it ? Maybe can be added later ?
> 2. `last_upd_date`: what format is it ? Maybe can be added later ?
> 
> Since this will be used to filter out opportunities in different processes keep it very simple ?

Im assuming a datetime? for `omb_review_status_date`
and if we store `last_update_date` that will probably also be a datetime.


I talked to @chouinar about this and while its true we dont need the data for anything right now it might help us with auditing.

also im not sure the table name `opportunity_review_status` make the most sense. its holding opportunities that are in a  review state that should filter them from the search index

babebe:
Added the two columns: with respective type per @chouinar. 
`omb_review_status_date` - date
`last_upd_date` - datetime

PR #5949, opened 2025-08-14
[Issue #5913] add  excluded opportunity review table

babebe:
## Summary
Fixes for https://github.com/HHS/simpler-grants-gov/issues/5913

## Changes proposed

Created table for ExcludedOpportunityReview table and factory.


api/src/db/models/opportunity_models.py
```diff
@@ -493,3 +493,15 @@ class OpportunityVersion(ApiSchemaTable, TimestampMixin):
...
     opportunity_data: Mapped[dict] = mapped_column(JSONB)
+
+
+class ExcludedOpportunityReview(ApiSchemaTable, TimestampMixin):
+    __tablename__ = "excluded_opportunity_review"
+
+    opportunity_id: Mapped[uuid.UUID] = mapped_column(
+        UUID, ForeignKey(Opportunity.opportunity_id), primary_key=True, index=True
+    )
+    opportunity: Mapped[Opportunity] = relationship(Opportunity)
```
chouinar: The ID we store in this table shouldn't be a foreign key & will be the legacy integer ID.

If it's a foreign key, when we build the process to copy over, we have to change the order of operations to do this after we copy opportunities which could be more complex, and have to resolve the legacy ID to map to our new ID.

I'd say we just make this an integer and don't have it be a foreign key as we're just going to do an `where not exists` with it.

api/src/db/models/opportunity_models.py
```diff
@@ -493,3 +493,12 @@ class OpportunityVersion(ApiSchemaTable, TimestampMixin):
     opportunity: Mapped[Opportunity] = relationship(Opportunity, back_populates="versions")
 
     opportunity_data: Mapped[dict] = mapped_column(JSONB)
+
+
+class ExcludedOpportunityReview(ApiSchemaTable, TimestampMixin):
+    __tablename__ = "excluded_opportunity_review"
+
+    opportunity_id: Mapped[int] = mapped_column(BigInteger, primary_key=True, index=True)
```
chouinar: A primary key is already indexed - can exclude that
```suggestion
    opportunity_id: Mapped[int] = mapped_column(BigInteger, primary_key=True)
```

PR #5952, opened 2025-08-15
[Issue #5584] Rework seed script to add users and more flexibility

### LINKED ISSUE #5584: Create all tests and seed data for Application Page
### Summary

To prepare the Application page for production, we need to create all tests and seed data in dev/staging to test the workflow.
Requires access to AWS.

### Acceptance criteria

- [ ] Create 3 users: one with no orgs, one with single org, one with multiple orgs
- [ ] Documentation added for the users and how to use them in the lower environments (how to log in)

chouinar:
## Summary
Fixes  #5584

## Changes proposed
* Add users/organization setup to seed script
* Fix a few issues with the seed script when trying to rerun it
* Fix some redundant seed script setup where it was creating an opportunity for every agency on every run
* Allow users to specify which parts of the seed script to run
* Rework form setup to allow for setting up several competitions easier
* Add a unit test to verify that you can safely rerun the seed script without it breaking

## Context for reviewers
All of these fixes look to address various annoyances with the seed script. Hopefully this is a lot easier for everyone to use, and sets up some building blocks for adding even more setup (we'll need a lot more for users when roles exist).

## Validation steps
You should be able to run `make db-seed-local` and then follow the guide for logging in as a particular user. The `no_org_user` should have some saved searches and opportunities you can see if you're running the frontend locally.

If you want to try running parts of the seed script like just running the form+competition part, you can do `make db-seed-local args="-s forms"`.


api/tests/src/api/agencies_v1/test_agencies_routes.py
```diff
@@ -10,7 +10,6 @@
 class TestAgenciesRoutes(BaseTestClass):
     @pytest.fixture(autouse=True)
     def cleanup_agencies(self, db_session):
-        yield
```
chouinar, line 13: Because the seed script test runs before this, it had agencies setup which broke these tests - made it so removing the yield makes this cleanup before the tests run, and solve that issue.

api/tests/src/db/models/factories.py
```diff
@@ -2576,7 +2576,7 @@ class Meta:
     ebiz_poc_last_name = factory.Faker("last_name")
     has_debt_subject_to_offset = sometimes_none(factory.Faker("boolean"), none_chance=0.8)
     has_exclusion_status = sometimes_none(factory.Faker("boolean"), none_chance=0.8)
-    eft_indicator = sometimes_none(factory.Faker("pystr", min_chars=3, max_chars=3))
```
chouinar, line 2579: This will only ever be `None` with some recent changes we made to the parsing of sam.gov extracts

joshtonava (approved):
👏

PR #5953, opened 2025-08-15
refactor LoginLink

### LINKED ISSUE #5950: Login menu item uses invalid markup
### Summary

The "Sign in" button in the header nav is made from an `a` element that contains a `button` element. This is not valid HTML. 

> The `a` element may be wrapped around entire paragraphs, lists, tables, and so forth, even entire sections, so long as there is no interactive content within (e.g. buttons or other links).

But this also causes a keyboard navigation issue. When using tab to navigate to the "Sign in" button, the `a` element gets focus before the `button` element gets focus. The user should not have to navigate thru 2 interactive elements that have the same function. 

Image

**AC:** 
- [ ] The "Sing in" menu item should be an `a` element without a button inside it 
- [ ] The responsive layout should still function the same

### Reproducibility

### How to reproduce the bug

Use the keyboard to navigate to (and past) the "Sign in" button 

### Browser

Chrome

### Code of Conduct

andycochran:
## Summary

Fixes #5950

## Changes proposed

Refactors the `LoginLink` to remove the `button` element from the `a`

## Context for reviewers

Be sure to review this on various screen sizes

### DIFF
--- a/frontend/src/components/Header.tsx
+++ b/frontend/src/components/Header.tsx
@@ -332,7 +332,7 @@ const Header = ({ locale }: Props) => {
             />
           </div>
           {!!showLoginLink && (
-            <div className="usa-nav__primary margin-top-0 padding-bottom-05 text-no-wrap desktop:order-last margin-left-auto desktop:height-auto height-6">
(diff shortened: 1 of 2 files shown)


doug-s-nava (approved):
THANK YOU

✅ ✅ ✅ ✅

Issue #5839, opened 2025-08-01
Allow usage of "allOf" blocks for referenced schemas


doug-s-nava:
### Summary

When referencing shared schemas within a form's json schema we would like to be able to wrap the reference in an "allOf" in order to allow for any future overriding or alteration of individual parts of the schema. However, when we do this currently the frontend code is unable to properly reference the schema when building a form due to the extra level of nesting it requires. 

One solution would be to look into using a library such as https://github.com/mokkabonna/json-schema-merge-allof to compile the "allOf"s prior to processing the schemas

If this library doesn't work for our purposes we may need to look into doing our own work to compile "allOf"s as referencing the schemas by explicitly including the "allOf" arrays in the path is not a sustainable solution.

### Acceptance criteria

- [ ] adding "allOf" blocks around refs (see changes to SF-LLL form here https://github.com/HHS/simpler-grants-gov/pull/5807/files#r2240355310 for an example) does not break rendering or saving forms

myduong-navapbc:
I agree. I think we should use the library @doug-s-nava mentioned above.



If we cannot find a solution:

I created a fix in a [previous draft PR](https://github.com/HHS/simpler-grants-gov/blob/3af894a96579eac64c60b065ec6f4de1f5410582/frontend/src/components/applyForm/utils.tsx)

```
    const resolveRef = (schema: JSONSchema7, ref: string): JSONSchema7 => {
      const refPath = ref.replace(/^#/, "");
      const resolved = getPointer(schema, refPath);
      if (!isObjectSchema(resolved)) return {};
      if (resolved.$ref) {
        return resolveRef(schema, resolved.$ref);
      }
      return resolved;
    };
```

```
    const resolveAllOf = (schema: JSONSchema7): JSONSchema7 => {
      if (!schema.allOf) return schema;

      const merged = schema.allOf.reduce((acc, subschema) => {
        if (!isObjectSchema(subschema)) return acc;

        let resolved = subschema;
        if (resolved.$ref) {
          resolved = resolveRef(rootSchema, resolved.$ref);
        }
        return merge(acc, resolveAllOf(resolved));
      }, {} as JSONSchema7);

      return merge({}, merged, { ...schema, allOf: undefined });
    };

    let current: any = rootSchema;

    for (const token of tokens) {
      if (!current) return undefined;

      while (current?.$ref) {
        current = resolveRef(rootSchema, current.$ref);
      }

      if (current?.allOf) {
[...]

PR #5955, opened 2025-08-15
[Issue #5839] resolve allof references

doug-s-nava:
## Summary

Fixes / Work for #5839 

## Changes proposed

Using the "json-schema-merge-allof" library, this change resolves all "allOf" references within our form schemas to allow easier traversal into individual fields in the "properties" areas of the schema.

This allows us to add back some removed "allOf" definitions in existing forms that were removed for expediency, and will help us down the line tackle some issues with validation and proper handling of fields and field values

## Context for reviewers

The "json-schema-merge-allof", when passed the full schema, will attempt to resolve every "allOf" reference, including some that are not problematic at the moment, such as top level "required" related "allOfs". Especially since some of these involve "if"/"then" logic that the library does not know how to resolve, this change works around things by only running against "properties" (which will include anything referenced from "$defs", since those paths will have already been resolved prior to processing the "allOfs"). If we end up adding "if"/"then" stuff into those areas we may need to revisit things.

## Validation steps

1. run `make volume-recreate db-seed-local populate-search-opportunities populate-search-agencies start` in /api on this branch
2. run `npm run dev` on this branch
3. follow directions from the readme updated in this PR to get to an application

for each form:
1. load the form
2. _VERIFY_: form loads
3. save the form
4. _VERIFY_: form saves with expected validation errors

### DIFF
--- a/frontend/src/components/applyForm/utils.tsx
+++ b/frontend/src/components/applyForm/utils.tsx
@@ -579,15 +579,9 @@ export const processFormSchema = async (
     const condensedProperties = mergeAllOf({
       properties: dereferenced.properties,
     } as JSONSchema7);
-    const condensedDefs = dereferenced.$defs
-      ? mergeAllOf({
-          $defs: dereferenced.$defs,
-        } as JSONSchema7)
-      : {};
     const condensed = {
       ...dereferenced,
       ...condensedProperties,
-      ...condensedDefs,
     };
     return condensed as RJSFSchema;
   } catch (e) {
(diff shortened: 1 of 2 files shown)


frontend/src/components/applyForm/utils.tsx
```diff
@@ -561,6 +564,36 @@ const flatFormDataToArray = (field: string, data: Record<string, unknown>) => {
   );
 };
 
+// resolve and "allOf" references within "properties" or "$defs" fields
+// not merging the entire schema because many schemas have top level
+// "allOf" blocks that often contain "if"/"then" statements or other things
+// that the mergeAllOf library can't handle out of the box, and we don't need
+// to condense in any case
+export const processFormSchema = async (
```
myduong-navapbc, line 572: VERY NICE!

myduong-navapbc (approved):

Issue #5933, opened 2025-08-13
Update data dump extract to filter out opportunities in review


joshtonava:
### Summary

There is an API that can create extracts for the entire dataset of opportunities. We should not include opportunities that are in review.  

### Tech Notes

- I believe the extract code is under `extract_routes.py`
- Add join for items that do not exist in the opportunity_review_status table

### Acceptance criteria

- When generating extracts through the API, opportunities should be filtered not to include those in review.
- Add / Update tests

PR #5962, opened 2025-08-18
[Issue #5933] Extract opportunity exclude review status

babebe:
## Summary

Fixes for #5933  

## Changes proposed

Update query to exclude opportunities in excluded_opportunity_review table from extracts.
Updated test


api/src/db/migrations/versions/2025_08_15_create_excluded_opportunity_review_table.py
chouinar, line 1: I think this is a holdover from your prior PR

chouinar (approved):

joshtonava (approved):

PR #5963, opened 2025-08-18
[Issue #6372] remove widget rendering from apply utils

doug-s-nava:
## Summary

Fixes for #6372 

## Changes proposed

 The big change here is replacing the "buildField" function, which rendered a widget for a form field, with the "getFieldConfig" function, which will return a widget type and set of props to use to render the widget. This allows us to render the form separately from building out the shape of the form, and makes testing the utility file much easier.

## Context for reviewers

Next step - factor out the calls to `renderWidget` and anything else that returns JSX so we can move the tree formation into a fully backend / non react based process

## Validation steps

1. start a local server on this branch using `npm run dev`
2. start an application using this [readme](https://github.com/HHS/simpler-grants-gov/blob/main/frontend/src/app/%5Blocale%5D/workspace/applications/application/%5BapplicationId%5D/README.md)
3. open and mess with each form
4. _VERIFY_: forms all function as expected

### DIFF
--- a/frontend/src/components/applyForm/widgets/WidgetRenderers.tsx
+++ b/frontend/src/components/applyForm/widgets/WidgetRenderers.tsx
@@ -1,3 +1,5 @@
+import { omit } from "lodash";
+
 import { JSX } from "react";
 
 import {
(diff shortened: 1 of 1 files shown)


frontend/src/components/applyForm/builFormTree.tsx
```diff
@@ -0,0 +1,133 @@
...
+            acc.forEach((item, key) => {
+              if (item) {
+                if (item.key === `${node.name}-wrapper`) {
+                  keys.push(key);
+                }
+              }
+            });
+            return null;
+          } else {
+            const widgetConfig = getFieldConfig({
```
doug-s-nava: this part replaces the previous "buildField"

acouch:
This looks great so far.

PR #5983, opened 2025-08-19
[Issue #5798] Add service to rename API keys from logged in user in DB

### LINKED ISSUE #5798: BE – Repository helper: rename_api_key()
### Summary

Added a helper function to rename the submitted API keys name

### Acceptance criteria

- [ ] Updates key_name; enforces uniqueness per user.
- [ ] Returns updated row count (1) or raises NotFoundError.
- [ ] Unit tests: rename success; duplicate name error.

btabaska:
## Summary

Fixes #5798
 

## Changes proposed

Added endpoint and utilities to rename an API key

### DIFF
--- a/api/src/api/users/user_routes.py
+++ b/api/src/api/users/user_routes.py
@@ -13,10 +13,10 @@ from src.api.users.user_blueprint import user_blueprint
 from src.api.users.user_schemas import (
     UserApiKeyCreateRequestSchema,
     UserApiKeyCreateResponseSchema,
-    UserApiKeyRenameRequestSchema,
-    UserApiKeyRenameResponseSchema,
     UserApiKeyListRequestSchema,
     UserApiKeyListResponseSchema,
+    UserApiKeyRenameRequestSchema,
+    UserApiKeyRenameResponseSchema,
     UserApplicationListRequestSchema,
     UserApplicationListResponseSchema,
     UserDeleteSavedOpportunityResponseSchema,
@@ -529,8 +529,9 @@ def user_rename_api_key(
     )
 
     return response.ApiResponse(message="Success", data=api_key)
-  
-  @user_blueprint.post("/<uuid:user_id>/api-keys/list")
+
+
+@user_blueprint.post("/<uuid:user_id>/api-keys/list")
 @user_blueprint.input(UserApiKeyListRequestSchema, location="json")
 @user_blueprint.output(UserApiKeyListResponseSchema)
 @user_blueprint.doc(responses=[200, 401, 403])
--- a/api/src/api/users/user_routes.py
+++ b/api/src/api/users/user_routes.py
@@ -519,15 +519,6 @@ def user_rename_api_key(
     with db_session.begin():
         api_key = rename_api_key(db_session, user_id, api_key_id, json_data)
 
-    logger.info(
-        "Renamed API key for user",
-        extra={
-            "user_id": user_id,
-            "api_key_id": api_key.api_key_id,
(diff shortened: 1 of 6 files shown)


api/src/services/users/rename_api_key.py
```diff
@@ -0,0 +1,49 @@
...
+    old_key_name = api_key.key_name
+    api_key.key_name = params.key_name
+
+    logger.info(
+        "Renamed API key",
+        extra={
+            "api_key_id": api_key.api_key_id,
+            "user_id": user_id,
+            "old_key_name": old_key_name,
+            "new_key_name": params.key_name,
```
chouinar: While someone shouldn't be putting PII in their key names, I wouldn't be surprised if someone names their keys "Bob Smith's API Key". I'd err on the side of not logging the api key names just so when we look in New Relic, we won't see someone's name potentially.

api/src/api/users/user_routes.py
```diff
@@ -490,3 +493,36 @@ def user_create_api_key(
     )
 
     return response.ApiResponse(message="Success", data=api_key)
+
+
+@user_blueprint.patch("/<uuid:user_id>/api-keys/<uuid:api_key_id>")
```
chouinar: Very nitpicky - doesn't technically matter, but we've made similar endpoints "PUT" endpoints. When we're dealing with one field, put/patch would behave identically either way, but if we add anything else, PUT ends up being simpler assuming we follow the convention:
* PUT - full replace all values, assume anything missing is null
* PATCH - only update passed-in values

api/src/services/users/rename_api_key.py
btabaska: This makes sense, I went in and removed the old and new key names to remove this possibility.

api/src/api/users/user_routes.py
btabaska: Wouldn't it make more sense to use patch on this since this endpoint is only for rename so we would only use it for that single field? Like if we were to make this endpoint generic and just an update endpoint then I guess that would make sense. However, if we keep this limited to rename for the scope of this MVP I don't see a reason to not use patch.

btabaska:
This has been updated and is ready for review second pass.

api/src/api/users/user_routes.py
```diff
@@ -495,6 +498,39 @@ def user_create_api_key(
...
+        api_key = rename_api_key(db_session, user_id, api_key_id, json_data)
+
+    logger.info(
+        "Renamed API key for user",
+        extra={
+            "user_id": user_id,
+            "api_key_id": api_key.api_key_id,
+            "key_name": api_key.key_name,
+        },
+    )
```
chouinar: Same as the other PR, we have this log happening twice in a row + have the key name here.

api/src/api/users/user_routes.py
chouinar: I'm mainly looking to avoid any PATCH endpoints in the API as a whole because of how complex their implementations end up being - like I said, with one field replace/update end up being mostly identical, but technically with a PATCH, `{}` could be a valid request which isn't supported as you made the field required.

btabaska:
Did not mean to re-request review. Wrong button. Ignore and set to request follow up

api/src/api/users/user_routes.py
btabaska: Resolved

btabaska:
PR has been updated based on feedback and is ready for re-review.

chouinar (approved):

PR #5984, opened 2025-08-19
[Issue #5800] Add service to delete API keys from logged in user in DB

### LINKED ISSUE #5800: BE – Repository helper: delete_api_key()
### Summary

Deletes an API Key row from the users API key table

### Acceptance criteria

- [ ] Hard‑deletes row; returns count of deleted rows.
- [ ] 0 rows ⇒ raise NotFoundError.
- [ ] Test verifies cascade on user delete.

btabaska:
## Summary

Fixes  #5800 

## Changes proposed

Adds endpoint and related utilities to delete an API key


api/src/services/users/delete_api_key.py
```diff
@@ -0,0 +1,35 @@
...
+def delete_api_key(db_session: db.Session, user_id: UUID, api_key_id: UUID) -> None:
+    api_key = db_session.execute(
+        select(UserApiKey).where(
+            UserApiKey.api_key_id == api_key_id,
+            UserApiKey.user_id == user_id,
+        )
+    ).scalar_one_or_none()
+
+    if not api_key:
+        raise_flask_error(404, "API key not found")
```
chouinar: Not required, just an idea: One pattern we've done in other endpoints is to reuse the GET service method (so `get_api_key` you have in another PR) so we don't even need to re-implement this bit. It's like 2 things, but if we add more as more features are needed, it can be convenient.

api/src/services/users/delete_api_key.py
```diff
@@ -0,0 +1,35 @@
...
+    key_name = api_key.key_name
+
+    db_session.delete(api_key)
+
+    logger.info(
+        "Deleted API key",
+        extra={
+            "api_key_id": api_key_id,
+            "user_id": user_id,
+            "key_name": key_name,
```
chouinar: Like the other PR, can we exclude the API key name just in case a user puts something like their name in the field?

btabaska: Done.

api/src/services/users/delete_api_key.py
btabaska: This makes sense. I'll reuse the list api keys utilities and filter the results by id. Updated with that feedback.

api/src/services/users/delete_api_key.py
```diff
@@ -0,0 +1,30 @@
...
+logger = logging.getLogger(__name__)
+
+
+def delete_api_key(db_session: db.Session, user_id: UUID, api_key_id: UUID) -> None:
+    """Delete an API key for a user"""
+    # Reuse the get_user_api_keys service to fetch all user's API keys
+    api_keys = get_user_api_keys(db_session, user_id)
+
+    # Find the specific API key to delete
+    api_key = next((key for key in api_keys if key.api_key_id == api_key_id), None)
```
chouinar: Could we use the get api key method you built instead of list here?

btabaska: I only built a method to get api keys in a list instead of getting a single key. At least I don't remember creating a method to get a single API key. I could pull out the functionality from rename api key into its own separate get single api key utility.

chouinar: I might have mixed it up with other work, doing a lot of PRs - my bad.

Could you make it fetch just one like before?

chouinar: @btabaska - not sure if you saw this comment - been holding on the review for this

btabaska: Shifted to working on the auth wrapper while I still had you last week. Made these changes for review by Josh O.

joshtonava (approved):
This looks good!

Completely minor but is there any value in also logging the `api_key.key_name`?

Issue #5754, opened 2025-07-22
Review Infra Platform and Simpler Terraform and environment and determine plan for Training


mdragon:
### Summary

Setup a plan and/or tickets for establishing a Training environment and/or Spike out an environment by manually running the Terraform locally.

### Acceptance criteria

- [ ] Plans or tickets created to identify the work involved in standing up Training. Alternatively, we have spiked out Terraform but not the CI/CD layers.

prasnava:
Created separate tickets for building for the training env: 
#5853 
#4465 
#6147 
#6150 
#6148 
#6149 
#6151 
#6153

prasnava:
Closing this ticket as we have the training environment stood up and created individual tickets for each task.

PR #5986, opened 2025-08-19
[#5754] - Create Training Environment

prasnava:
## Summary 

#5754 

## Changes proposed

Created new Training Environment


infra/api/app-config/env-config/scheduled_jobs.tf
chouinar, line 1: For training, could we disable `load-transform` and then come back to it? That's the one we'll need to run for like a day before it's backfilled, the hourly job would run on top of itself if we enable it.

infra/api/app-config/env-config/scheduled_jobs.tf
```diff
@@ -30,6 +30,17 @@ locals {
...
+      "poetry",
+      "run",
+      "flask",
+      "data-migration",
+      "load-transform",
+      "--load",
+      "--transform",
+      "--set-current",
+      "--store-version"
+    ],
```
chouinar, line 45: If we want to just make this command be what is set by default, and not have the env-specific values, that would be fine - all the envs have the same config.

infra/modules/database/main.tf
```diff
@@ -18,7 +18,7 @@ locals {
   # See https://aws.amazon.com/blogs/database/using-iam-authentication-to-connect-with-pgadmin-amazon-aurora-postgresql-or-amazon-rds-for-postgresql/
   db_user_arn_prefix = "arn:aws:rds-db:${data.aws_region.current.name}:${data.aws_caller_identity.current.account_id}:dbuser:${aws_rds_cluster.db.cluster_resource_id}"
 
-  engine_version       = "15.4"
+  engine_version       = "15.10"
```
chouinar: Will this have any impact on the currently running DBs in each environment?

infra/project-config/networks.tf
```diff
@@ -42,6 +42,19 @@ locals {
         certificate_configs = {}
       }
     }
+    training = {
+      account_name                 = "simpler-grants-gov"
+      database_subnet_group_name   = "training"
+      vpc_name                     = "training"
+      second_octet                 = 9               # The second octet our the VPC CIDR block
+      grants_gov_oracle_cidr_block = "10.220.0.0/16" # MicroHealth managed CIDR block where the dev origin Oracle database for Grants.gov is located
```
chouinar:
```suggestion
      grants_gov_oracle_cidr_block = "10.220.0.0/16" # MicroHealth managed CIDR block where the training origin Oracle database for Grants.gov is located
```

infra/api/app-config/training.tf
```diff
@@ -0,0 +1,56 @@
...
+  search_engine_version = "OpenSearch_2.15"
+
+  service_override_extra_environment_variables = {
+    # Login.gov OAuth
+    ENABLE_AUTH_ENDPOINT   = 1
+    ENABLE_APPLY_ENDPOINTS = 1
+    ENABLE_SOAP_API        = 1
+    # Email notification
+    RESET_EMAILS_WITHOUT_SENDING = "false"
+  }
```
chouinar, line 52: @mdragon - for login.gov and sam.gov, do we want training to hit the same data as prod?

infra/api/app-config/env-config/scheduled_jobs.tf
prasnava: load-transform has been disabled.

infra/modules/database/main.tf
prasnava: Will parameterize this option instead of hardcoding it. We would want to keep the version same for now, we can schedule the minor upgrades for the other environments.

mdragon (dismissed):
🚀

PR #5988, opened 2025-08-19
[Issue #5918]  Search Filtering - Update Opportunities search Index to use review status

### LINKED ISSUE #5918: Search Filtering - Update Opportunities search Index to use review status
### Summary

Now that the opportunity review statuses are available, we need to modify search indexing to filter opportunities based on review status.

### Tech Notes

- Add join to `opportunity_review_status` table
- Filter out opportunities with status 'Reviewable' or 'Returned'
- Update both full refresh and incremental load logic

### Acceptance criteria

- Modify search indexing to filter opportunities based on review status.
- Add/update tests

joshtonava:
## Summary

Fixes #5918  

## Changes proposed

- Modified `load_opportunities_to_index.py` to filter out opportunities that exist in the `ExcludedOpportunityReview` table
- Added test coverage

## Context for reviewers

We have a new table `excluded_opportunity_review` that signifies all opportunities that are in a review state and should not be available in search results.

## Validation steps

To run these specific tests 

```bash
make test args="-vv tests/src/search/backend/test_load_opportunities_to_index.py"
```

### DIFF
--- a/api/tests/src/search/backend/test_load_opportunities_to_index.py
+++ b/api/tests/src/search/backend/test_load_opportunities_to_index.py
@@ -22,6 +22,7 @@ from tests.src.db.models.factories import (
     OpportunityChangeAuditFactory,
     OpportunityFactory,
 )
+from tests.src.db.models.factories import ExcludedOpportunityReviewFactory
 
 
 class TestLoadOpportunitiesToIndexFullRefresh(BaseTestClass):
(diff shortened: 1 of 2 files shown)


api/tests/src/search/backend/test_load_opportunities_to_index.py
```diff
@@ -226,6 +227,61 @@ def test_opportunity_attachment_pipeline(
...
+        self,
+        db_session,
+        truncate_opportunities,
+        enable_factory_create,
+        search_client,
+        opportunity_index_alias,
+        load_opportunities_to_index,
+    ):
+        # Why is truncate_opportunities not working as expected?
+        cascade_delete_from_db_table(db_session, Opportunity)
```
joshtonava: Any idea as to why `truncate_opportunities` doesn't take care of this already?

If I omit this line, the search results have 32 entries in them from previous tests.

chouinar: `truncate_opportunities` is scoped to the class - it only runs once for all of the classes in this test.

Unless it is otherwise an issue, I'd actually recommend not truncating if possible. Just add a few more opportunities, and ignore the others. Both to avoid the time it takes to truncate and to verify none of the other cases get messed up.

chouinar (approved):

Issue #5919, opened 2025-08-12
Search Filtering - Update Agency search Index to use review status


joshtonava:
### Summary

Now that the opportunity review statuses are available, we need to modify search indexing to filter agencies based on review status.
Load agencies to search relies on opportunity status from the db so we need to also filter by whatever this is.

### Tech Notes

- Ensure agency search respects opportunity review status filtering
- Update opportunity status counts for agencies

### Acceptance criteria

- Modify search indexing to filter agencies based on review status
- Add/update tests

PR #5989, opened 2025-08-19
[Issue #5919] Search agency  review status

babebe:
## Summary

Fixes for #5919 

## Changes proposed

Update query to exclude opportunities in review status.


api/tests/src/search/backend/test_load_agencies_to_index.py
```diff
@@ -115,3 +119,27 @@ def test_load_agencies_with_status(
             OpportunityStatus.FORECASTED.value
         ]
         assert archived_agency_resp["opportunity_statuses"] == [OpportunityStatus.ARCHIVED.value]
+
+    def test_load_agencies_to_index_review_status(
```
chouinar, line 123: Could we add one test for the scenario we talked about where an agency has two opportunities, but one is filtered? This should effectively not matter, the agency would still get the status.

PR #5999, opened 2025-08-20
[Issue #5055] Call the "Application render" endpoint and send it to Docraptor for "PDFing"

### LINKED ISSUE #5055: Call the "Application render" endpoint and send it to Docraptor for "PDFing"
### Summary

Stitch together the calling of the FE page to render a full Application to HTML and the integration of Docraptor to yield a PDF

---

We want to make a logic that given an application form record, can do the following:
* Generate a short-lived token (logic implemented in #5307 - just need to call a function here )
* Call a frontend URL passing the application ID, application form ID + the short-lived token
* With the contents of that response, call Docraptor
* Return this value (should be a bytestream if I'm reading Docraptor docs right)

We'll call this logic as part of a backend task that generates PDFs. We should implement this and manually verify we can make PDF data, BUT don't need to worry about the front-end URL yet. We can call Docraptor with their dummy public key and make sure we get a response.

We also need an easy way to mock this logic out as our unit tests shouldn't actually make either of these calls. We don't need to check the contents of the response of either one, just status codes to make sure it was successful.

### Acceptance criteria

- [ ] Call a FE URL and expose that HTML for sending to Docraptor
- [ ] Call that FE URL in the sidecar rather than the FE ECS service
- [ ] Call the actual Application page URL created in #5052 in the sidecar
- [ ] Send the HTML received to Docraptor

mikehgrantsgov:
## Summary

Fixes #5055

## Changes proposed

Create 2 new clients: `FrontendClient` for HTML retrieval from the frontend, `DocRaptorClient` for PDF conversion, with mock implementations for testing.
Add PDF Generation Service. This calls frontend /print endpoints and converts HTML to PDF via DocRaptor API
Add env config for frontend URL, DocRaptor API settings, and mock usage
Add new unit tests

## Context for reviewers

Implements the PDF generation client/service portion, intentionally separated from task integration. 

The service generates PDF bytes that tasks can consume. The frontend integration requires the /print/application/[id]/form/[formId] route to render application forms as HTML. DocRaptor integration uses their API to convert HTML to PDF.

## Validation steps

See relevant new unit tests that cover the client and service functionality.


api/src/services/pdf_generation/clients.py
```diff
@@ -0,0 +1,186 @@
...
+        except requests.exceptions.RequestException as e:
+            logger.error(
+                "Error requesting application form HTML from frontend",
+                extra={
+                    "application_id": str(request.application_id),
+                    "application_form_id": str(request.application_form_id),
+                    "url": url,
+                    "error": str(e),
+                },
+            )
```
chouinar: logger exception will add the stack trace to the log record - we have special log parsing added recently that'll even pull class, message and others out.

```suggestion
        except requests.exceptions.RequestException:
            logger.exception(
                "Error requesting application form HTML from frontend",
                extra={
                    "application_id": str(request.application_id),
                    "application_form_id": str(request.application_form_id),
                    "url": url,
                },
            )
```

api/local.env
```diff
@@ -179,3 +179,11 @@ SAM_GOV_API_URL=https://api.sam.gov/
 SAM_GOV_USE_MOCK=true
 SAM_GOV_MOCK_DATA_FILE=/path/to/mock/data.json
 SAM_GOV_MOCK_EXTRACT_DIR=/path/to/mock/extracts
+
+# PDF Generation Configuration
+FRONTEND_SIDECAR_URL=http://localhost:3000
```
chouinar: NIT: We aren't running as a sidecar, just calling the actual frontend

api/src/services/pdf_generation/service.py
```diff
@@ -0,0 +1,189 @@
...
+        except Exception as e:
+            logger.error(
+                "Error generating PDF",
+                extra={
+                    "application_id": str(request.application_id),
+                    "application_form_id": str(request.application_form_id),
+                    "error": str(e),
+                },
+                exc_info=True,
+            )
```
chouinar:
```suggestion
        except Exception as e:
            logger.exception(
                "Error generating PDF",
                extra={
                    "application_id": str(request.application_id),
                    "application_form_id": str(request.application_form_id),
                },
            )
```

api/src/services/pdf_generation/service.py
```diff
@@ -0,0 +1,189 @@
...
+            minutes=self.config.short_lived_token_expiration_minutes
+        )
+
+        # Create the JWT token and database record
+        token, short_lived_token = create_jwt_for_internal_token(
+            expires_at=expires_at, db_session=db_session
+        )
+
+        # Commit the token to the database
+        db_session.commit()
```
chouinar: I'd move the commit out of this function and let whoever is calling control that. We'll need to do it very carefully for the job so it happens in a separate transaction from actually processing forms.

api/src/services/pdf_generation/service.py
```diff
@@ -0,0 +1,189 @@
...
+        use_mocks: Whether to use mock clients for testing
+
+    Returns:
+        Configured PdfGenerationService instance
+    """
+    if config is None:
+        config = get_config()
+
+    frontend_client = MockFrontendClient() if use_mocks else FrontendClient(config)
+    docraptor_client = MockDocRaptorClient() if use_mocks else DocRaptorClient(config)
```
chouinar, line 147: Should this instead use the env var instead of passing a value in?
```suggestion
    if config is None:
        config = get_config()

    frontend_client = MockFrontendClient() if config.pdf_generation_use_mocks else FrontendClient(config)
    docraptor_client = MockDocRaptorClient() if config.pdf_generation_use_mocks else DocRaptorClient(config)
```

api/src/services/pdf_generation/clients.py
```diff
@@ -0,0 +1,186 @@
...
+    def __init__(self, config: PdfGenerationConfig | None = None):
+        if config is None:
+            config = PdfGenerationConfig()
+        self.config = config
+
+    def html_to_pdf(self, html_content: str) -> bytes:
+        """Convert HTML content to PDF using DocRaptor."""
+
+        # DocRaptor API endpoint
+        url = "https://docraptor.com/docs"
```
chouinar: Shouldn't this come from the config class?
```suggestion

```

api/src/services/pdf_generation/clients.py
```diff
@@ -0,0 +1,186 @@
...
+                "Converting HTML to PDF using DocRaptor",
+                extra={
+                    "test_mode": self.config.docraptor_test_mode,
+                    "html_length": len(html_content),
+                },
+            )
+
+            response = requests.post(url, json=payload, headers=headers, auth=auth, timeout=60)
+            response.raise_for_status()
+
```
chouinar: Docraptor has a python class that does a lot of what you did here - are we able to use that?
https://docraptor.com/documentation/python

mikehgrantsgov: I added this. Nice to know it's not necessary to roll our own.

api/src/services/pdf_generation/service.py
mikehgrantsgov: Updated this

api/src/services/pdf_generation/clients.py
mikehgrantsgov: Yep, that was left over. Updated this

api/local.env
mikehgrantsgov: I updated this, config and var references

api/pyproject.toml
```diff
@@ -42,6 +42,8 @@ python-pptx = "^1.0.2"
 openpyxl = "^3.1.5"
 defusedxml = "^0.7.1"
 xmltodict = "^0.14.2"
+requests = "^2.31.0"
```
mikehgrantsgov: @chouinar I've seen other imports for `requests` in the project but didn't see it in pyproject earlier. Maybe don't need this here?

chouinar: I think we don't have it explicitly added, but a bunch of other libraries use it so we have it implicitly.

api/src/services/pdf_generation/clients.py
```diff
@@ -137,20 +128,34 @@ def html_to_pdf(self, html_content: str) -> bytes:
...
-                    "html_length": len(html_content),
+                    "status_code": e.status,
+                    "reason": e.reason,
+                    "body": e.body if hasattr(e, "body") else None,
                 },
             )
             raise
+        except Exception as e:
+            logger.error("Unexpected error calling DocRaptor API", extra={"error": str(e)})
+            raise
```
chouinar:
```suggestion
        except Exception:
            logger.exception("Unexpected error calling DocRaptor API")
            raise
```

chouinar (commented):
Just one minor suggestion

chouinar (approved):

PR #6018, opened 2025-08-20
[Issue #5051] Add PDF fetching to zip process

### LINKED ISSUE #5051: Add PDF fetching logic to the ZIP process
### Summary

After #5055 - add the PDF generation logic calls to the application submission ZIP creation process

---
We want to actually use the PDF generation calling logic in our zip creation process. We will want to update the `process_application_forms` method to call the logic we created in #5055.

This function already has the general pattern for how we want to put the files into the ZIP, but creates a static file with `TODO` as the contents. Instead we want the bytes we write to that file to be what was returned by the PDF generation logic.

We'll likely also need to setup the short lived token somewhere in this task which has a few challenges. The token needs to be written and committed to the DB so the frontend can reference it, that likely means we need to do that first in a dedicated transaction before anything else regarding processing the applications as we don't want to be committing outside of the batch we're doing.

Additionally, we should test that this works locally 

### Acceptance criteria

- [ ] Implement above
- [ ] Short lived internal token is created before processing a batch of applications in a separate transaction
- [ ] Add tests

mikehgrantsgov:
## Summary

Fixes #5051 

## Changes proposed

Create a single short-lived token for this batch (if not using mocks)
Replace TODO with implemented `generate_application_form_pdf`

## Context for reviewers

`generate_application_form_pdf` has tests associated from a previous PR.

## Validation steps

We tested PDF generation locally, which works with CSS issues. After this is merged we will test in dev/staging to see if the Base URL to Docraptor will resolve the CSS issue.

### DIFF
--- a/api/tests/src/services/pdf_generation/test_service.py
+++ b/api/tests/src/services/pdf_generation/test_service.py
@@ -92,8 +92,8 @@ class TestPdfGenerationService:
         call_args = mock_create_jwt.call_args
         expires_at = call_args[1]["expires_at"]
         expected_expires_at = datetime(
-            2024, 1, 1, 12, 15, 0, tzinfo=expires_at.tzinfo
-        )  # 15 minutes later with same timezone
+            2024, 1, 1, 13, 0, 0, tzinfo=expires_at.tzinfo
+        )  # 60 minutes later with same timezone
         assert expires_at == expected_expires_at
 
         # Note: commit is no longer called within the token generation function


api/src/task/apply/create_application_submission_task.py
```diff
@@ -200,12 +249,39 @@ def process_application_forms(self, submission: SubmissionContainer) -> None:
...
-            #        call it here and pass the contents below.
+
+            # Generate PDF from the application form
+            pdf_response = generate_application_form_pdf(
+                db_session=self.db_session,
+                application_id=submission.application.application_id,
+                application_form_id=application_form.application_form_id,
+                use_mocks=self.pdf_generation_config.pdf_generation_use_mocks,
+                token=batch_token,
+            )
```
chouinar: Could we move some of this into a function to hide away the PDF stuff? Something like having this be:

```py
pdf_response = self.get_pdf_for_app_form( ... )
```

And in that function we check success/error. I think if it's an error we should just raise the error as we can't create a submission in that case, we wouldn't want to continue.

api/src/task/apply/create_application_submission_task.py
```diff
@@ -104,9 +115,12 @@ def process_batch(self) -> None:
         """Process a batch of application submissions"""
         submitted_applications = self.fetch_applications()
 
+        # Create a single short-lived token for this batch (if not using mocks)
+        batch_token = self._create_batch_token()
```
chouinar: I think if we put this here we might hit DB session issues since you're committing in that function in a transaction that we later commit more into.

I'm thinking instead of even a batch-token, we just make a token for the entire job life and set it up in the init function. Then we just have this in the init function, and don't pass it around, just settting `self.internal_token = ...`.

---
In that case we'd probably want the expiration time to be 60 minutes, but we can configure that in terraform if needed

api/src/task/apply/create_application_submission_task.py
```diff
@@ -139,7 +153,40 @@ def fetch_applications(self) -> Sequence[Application]:
...
+        """Create a single short-lived token for the entire batch.
+
+        Returns:
+            JWT token string, or None if using mocks
+        """
+        if self.pdf_generation_config.pdf_generation_use_mocks:
+            return None  # Mock clients don't need real tokens
+
+        # Create token in its own transaction to avoid conflicts
+        from datetime import datetime, timedelta, timezone
```
chouinar: Can we put these at the top of the file?

api/src/task/apply/create_application_submission_task.py
```diff
@@ -139,7 +153,40 @@ def fetch_applications(self) -> Sequence[Application]:
...
+
+        expires_at = datetime.now(timezone.utc) + timedelta(
```
chouinar: To avoid needing to pull in timezone
```suggestion
        expires_at = datetime_util.utcnow() + timedelta(
```

api/src/task/apply/create_application_submission_task.py
```diff
@@ -139,7 +153,40 @@ def fetch_applications(self) -> Sequence[Application]:
...
 
-    def process_application(self, application: Application) -> None:
+    def _create_batch_token(self) -> str | None:
+        """Create a single short-lived token for the entire batch.
+
+        Returns:
+            JWT token string, or None if using mocks
+        """
+        if self.pdf_generation_config.pdf_generation_use_mocks:
+            return None  # Mock clients don't need real tokens
```
chouinar: Could we make the token always set so that it getting passed around is tested? Even if the mock doesn't use it, just to verify that the behavior works.

api/src/task/apply/create_application_submission_task.py
mikehgrantsgov: Updated this

api/src/task/apply/create_application_submission_task.py
mikehgrantsgov: Updated this as well

api/src/task/apply/create_application_submission_task.py
mikehgrantsgov: Updated this as well

PR #6049, opened 2025-08-21
[Unticketed] Add PDF Terraform env vars

mikehgrantsgov:
## Changes proposed

Add relevant env vars for PDF generation 

## Context for reviewers

Currently unticketed. Will close / rename this if a ticket is created.

## Validation steps

Confirm env vars are set after TF is applied.

### DIFF
--- a/infra/api/app-config/env-config/environment_variables.tf
+++ b/infra/api/app-config/env-config/environment_variables.tf
@@ -34,6 +34,13 @@ locals {
 
     # Sam.gov
     SAM_GOV_BASE_URL = "https://api-alpha.sam.gov"
+
+    # PDF Generation Configuration
+    FRONTEND_URL = "https://${var.environment}.simpler.grants.gov"
+    DOCRAPTOR_TEST_MODE = "false"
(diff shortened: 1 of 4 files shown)


infra/api/app-config/env-config/environment_variables.tf
```diff
@@ -34,6 +34,13 @@ locals {
 
     # Sam.gov
     SAM_GOV_BASE_URL = "https://api-alpha.sam.gov"
+
+    # PDF Generation Configuration
+    FRONTEND_URL = "https://${var.environment}.simpler.grants.gov"
+    DOCRAPTOR_TEST_MODE = "false"
```
chouinar: I'd say we should default to test mode since only prod will ever turn that off.

mikehgrantsgov: Makes sense, updated this

chouinar (approved):

chouinar:
@mikehgrantsgov - can you make sure to add the parameter store value before merging? Fine to just put the same test key in all envs for now

Issue #5782, opened 2025-07-28
BE – Create `@require_api_key` decorator (new module)


btabaska:
### Summary

Add a fresh Flask decorator in auth/api_key.py that authenticates proxied traffic from API Gateway.

### Acceptance criteria

- [ ] Reads X‑API‑Key header; rejects if missing.
- [ ] SHA‑256 compares against active hash in api_keys table (redis cached).
- [ ] Returns JSON 401 with { "error": "invalid_api_key" }.
- [ ] Does not modify existing session/JWT decorators.
- [ ] Unit tests cover valid, missing, inactive, revoked cases.

PR #6051, opened 2025-08-21
[Issue #5782] Add API key auth

btabaska:
## Summary

Fixes / Work for #5782 

## Changes proposed

Added the ability for API keys to multi_auth with both API Gateway keys and legacy API keys


api/src/api/form_alpha/form_route.py
```diff
@@ -32,17 +34,23 @@ def form_get(db_session: db.Session, form_id: uuid.UUID) -> response.ApiResponse
...
-    current_user = api_key_auth.current_user
-    if current_user is None or current_user.username != "auth_token_0":
+    # Check if user is the internal admin user (auth_token_0) - only for environment API keys
+    multi_user = api_key_multi_auth.get_user()
+    if multi_user.auth_type == AuthType.API_KEY_AUTH:
+        # Environment-based API key - check for auth_token_0
+        api_key_user = cast(ApiKeyUser, multi_user.user)
+        if api_key_user.username != "auth_token_0":
+            raise_flask_error(403, "Only internal admin users can update forms")
+    else:
```
chouinar: Can we just hold on adding the new key auth to this endpoint? I'd rather add it later when we support roles rather than allow the key and then error it in all cases. This endpoint _needs_ to be locked down as anyone using it that isn't one of our developers could break the apply process.

api/src/auth/api_gateway_key_auth.py
```diff
@@ -0,0 +1,78 @@
...
+from src.db.models.user_models import UserApiKey
+from src.logging.flask_logger import add_extra_data_to_current_request_logs
+
+logger = logging.getLogger(__name__)
+
+# Initialize the authorization context for API Gateway key authentication
+# This uses the X-API-Key header which is the standard header that AWS API Gateway
+# forwards when api_key_required is set to true
+api_gateway_key_auth = JwtUserHttpTokenAuth(
+    "ApiKey", header="X-API-Key", security_scheme_name="ApiGatewayKeyAuth"
```
chouinar: This shouldn't be a `JwtUserHttpTokenAuth` as it's not using a JWT.


```suggestion
api_gateway_key_auth = HttpTokenAuth(
    "ApiKey", header="X-API-Key", security_scheme_name="ApiGatewayKeyAuth"
```

api/src/auth/api_gateway_key_auth.py
```diff
@@ -0,0 +1,78 @@
...
+    logger.info("Authenticating API Gateway key")
+
+    try:
+        api_key = validate_api_key_in_db(token, db_session)
+
+        # Update last_used timestamp
+        api_key.last_used = datetime_util.utcnow()
+        db_session.add(api_key)
+        # Note: We don't commit here as the transaction will be managed
+        # by the calling context (e.g., @flask_db.with_db_session)
```
chouinar: We should commit here. There isn't actually any logic we've built into the `with_db_session` that does anything to close the transaction and while I think SQLAlchemy does do that when the session isn't being used anymore, I've seen inconsistencies in how that works. Also, if there were any issues with doing the commit, we'd rather the stack trace be in our code and not 12 layers deep in SQLAlchemy code.

Just add a:

```py
with db_session.begin():
     try:
         ...
```
around this section of code.

api/src/auth/auth_utils.py
```diff
@@ -19,6 +19,7 @@ def get_app_security_scheme() -> dict[str, Any]:
         "ApiKeyAuth": {"type": "apiKey", "in": "header", "name": "X-Auth"},
         "ApiJwtAuth": {"type": "apiKey", "in": "header", "name": "X-SGG-Token"},
         "InternalApiJwtAuth": {"type": "apiKey", "in": "header", "name": "X-SGG-Internal-Token"},
+        "ApiGatewayKeyAuth": {"type": "apiKey", "in": "header", "name": "X-API-Key"},
```
chouinar: Instead of `ApiGatewayKeyAuth` can we call this something that doesn't reference the AWS service? Something like `ApiUserKeyAuth` maybe? The user/our own API doesn't care that it's API gateway.

api/tests/src/auth/test_api_gateway_key_auth.py
```diff
@@ -0,0 +1,135 @@
...
+    db_session.commit()
+
+    resp1 = mini_app.test_client().get("/dummy_auth_endpoint", headers={"X-API-Key": "user-key-1"})
+    assert resp1.status_code == 200
+
+    resp2 = mini_app.test_client().get("/dummy_auth_endpoint", headers={"X-API-Key": "user-key-2"})
+    assert resp2.status_code == 200
+
+    # Note: last_used timestamps are updated during authentication but not committed
+    # in the test environment due to transaction management
```
chouinar: They should be committed - we don't do anything that prevents commits from working normally with how our tests are setup.

api/tests/src/auth/test_api_gateway_key_auth.py
```diff
@@ -0,0 +1,135 @@
...
+    assert resp.status_code == 401
+    assert resp.get_json()["message"] == "Invalid API key"
+
+
+def test_api_gateway_key_auth_multiple_active_keys(mini_app, enable_factory_create, db_session):
+    """Test that different API keys for the same user work independently"""
+    user = UserFactory.create()
+    UserApiKeyFactory.create(user=user, key_id="user-key-1", is_active=True)
+    UserApiKeyFactory.create(user=user, key_id="user-key-2", is_active=True)
+    db_session.commit()
```
chouinar: The factory create method calls commit every time it runs - no need to commit.

```suggestion
    user = UserFactory.create()
    UserApiKeyFactory.create(user=user, key_id="user-key-1", is_active=True)
    UserApiKeyFactory.create(user=user, key_id="user-key-2", is_active=True)
```

api/src/api/opportunities_v1/opportunity_routes.py
```diff
@@ -180,7 +180,7 @@
     examples=examples,
 )
 @opportunity_blueprint.output(opportunity_schemas.OpportunitySearchResponseV1Schema())
-@opportunity_blueprint.auth_required(api_key_auth)
+@api_key_multi_auth.login_required
```
chouinar, line 183: If we're adding a new auth to these endpoints, can you add some tests for each endpoint with the new auth? I'd say we want at least:
* Happy path - endpoint works with the new token
* Invalid token - endpoint raises a 403

api/src/api/form_alpha/form_route.py
btabaska: This makes sense, I'll go ahead and revert the changes on this endpoint.

api/src/auth/auth_utils.py
btabaska: Sounds good, will do a quick find replace.

api/src/auth/api_gateway_key_auth.py
btabaska: Adding just the with db_session.begin() around that block of code causes I issues because it is already in a transaction with the higher up with flask_db.get_db we call above. I can still commit the changes though. 
` # Get database session
    with flask_db.get_db(current_app).get_session() as db_session:
        try:
            api_key = validate_api_key_in_db(token, db_session)
            api_key.last_used = datetime_util.utcnow()
            db_session.add(api_key)
            db_session.commit()`

api/src/api/opportunities_v1/opportunity_routes.py
btabaska: Added tests

api/src/auth/api_gateway_key_auth.py
chouinar: We want this as a decorator, don't switch that, that decorator doesn't affect anything with transactions.

The `with db_session.begin()` needs to wrap the entire contents of the function, because above this highlighted text we call the DB, that automatically starts a transaction (and db queries implicitly start transactions, but it's better to explicitly wrap your transaction for clarity).

api/tests/src/api/opportunities_v1/test_opportunity_auth.py
```diff
@@ -19,3 +21,138 @@ def test_opportunity_unauthorized_401(client, api_auth_token, method, url, body)
...
+
+def test_opportunity_get_legacy_success_with_api_user_key(
+    client, enable_factory_create, db_session
+):
+    """Test legacy opportunity get endpoint with valid API user key"""
+    user = UserFactory.create()
+    api_key = UserApiKeyFactory.create(user=user, key_id="valid-legacy-get-key", is_active=True)
+    # Create an opportunity to get
+    opportunity = OpportunityFactory.create()
+    db_session.commit()
```
chouinar:
```suggestion
```

api/tests/src/api/opportunities_v1/test_opportunity_auth.py
```diff
@@ -19,3 +21,138 @@ def test_opportunity_unauthorized_401(client, api_auth_token, method, url, body)
...
+    assert api_key.last_used is not None
+
+
+def test_opportunity_get_success_with_api_user_key(client, enable_factory_create, db_session):
+    """Test opportunity get endpoint with valid API user key"""
+    user = UserFactory.create()
+    api_key = UserApiKeyFactory.create(user=user, key_id="valid-get-key", is_active=True)
+    # Create an opportunity to get
+    opportunity = OpportunityFactory.create()
+    db_session.commit()
```
chouinar:
```suggestion
```

api/tests/src/api/opportunities_v1/test_opportunity_auth.py
```diff
@@ -19,3 +21,138 @@ def test_opportunity_unauthorized_401(client, api_auth_token, method, url, body)
...
+
+    # Verify the API key's last_used was updated
+    db_session.refresh(api_key)
+    assert api_key.last_used is not None
+
+
+def test_opportunity_get_success_with_api_user_key(client, enable_factory_create, db_session):
+    """Test opportunity get endpoint with valid API user key"""
+    user = UserFactory.create()
+    api_key = UserApiKeyFactory.create(user=user, key_id="valid-get-key", is_active=True)
```
chouinar: Don't create the API key in the unit test, we'll need this in hundreds of tests eventually, make a fixture in the tests/conftest.py file that does this and pull it in.

Ideally the key ID should not be static as it'll be re-made for each separate unit test.

Something like:
```py
@pytest.fixture
def user_api_key(enable_factory_create):
     return UserApiKeyFactory.create() # don't need to specify anything as the factory makes the user, and happy defaults

@pytest.fixture
def user_api_key_id(user_api_key):
     """Convenience fixture as most of the time we just care about the ID"""
     return user_api_key.key_id
```
Then in tests we just need:
```py
def test_thing(client, db_session, user_api_key_id):
     # whatever setup for opportunities or non-user stuff
     response = client.get(
        f"/whatever-route", headers={"X-API-Key": user_api_key_id}
```

api/src/auth/api_user_key_auth.py
```diff
@@ -0,0 +1,77 @@
...
+            )
+
+            logger.info("API Gateway key authentication successful")
+
+            return api_key
+
+        except ApiKeyValidationError as e:
+            logger.info("API Gateway key authentication failed", extra={"auth.issue": e.message})
+            raise_flask_error(401, e.message)
+
```
chouinar: This should work / handle the commits. Also moved the except up a bit since it would only alert on that particular function.

```suggestion
@api_user_key_auth.verify_token
@flask_db.with_db_session()
def verify_api_key(db_session: db.Session, token: str) -> UserApiKey:
    logger.info("Authenticating API Gateway key")

    with db_session.begin():
        try:
            api_key = validate_api_key_in_db(token, db_session)
        except ApiKeyValidationError as e:
            logger.info("API Gateway key authentication failed", extra={"auth.issue": e.message})
            raise_flask_error(401, e.message)
            
        api_key.last_used = datetime_util.utcnow()
        db_session.add(api_key)

        add_extra_data_to_current_request_logs(
            {
                "auth.user_id":api_key.user_id,
                "auth.api_key_id": strapi_key.api_key_id,
            }
        )

        logger.info("API Gateway key authentication successful")

        return api_key



```

api/src/auth/api_user_key_auth.py
```diff
@@ -0,0 +1,77 @@
...
+    ).scalar_one_or_none()
+
+    if user_api_key is None:
+        raise ApiKeyValidationError("Invalid API key")
+
+    if not user_api_key.is_active:
+        raise ApiKeyValidationError("API key is inactive")
+
+    if user_api_key.user is None:
+        raise ApiKeyValidationError("API key is not associated with a valid user")
```
chouinar: User ID (and thus the user relationship) are non-nullable and the foreign key constraint enforces that it'll point to something, can skip this check.

```suggestion
```

api/src/api/form_alpha/form_route.py
chouinar: I think you missed removing this?

api/src/api/agencies_v1/agency_routes.py
```diff
@@ -108,7 +108,7 @@ def agencies_get(db_session: db.Session, raw_list_params: dict) -> response.ApiR
     examples=examples,
 )
 @agency_blueprint.output(agency_schema.AgencySearchResponseV1Schema)
-@agency_blueprint.auth_required(api_key_auth)
+@api_key_multi_auth.login_required
```
chouinar, line 113: You added tests for the opportunity routes, could you add one for this agency route?

api/src/api/agencies_v1/agency_routes.py
```diff
@@ -41,7 +41,7 @@
     examples=examples,
 )
 @agency_blueprint.output(agency_schema.AgencyListResponseSchema)
-@agency_blueprint.auth_required(api_key_auth)
+@api_key_multi_auth.login_required
```
chouinar: Actually, we should skip this endpoint, this one should be deprecated in favor of the below endpoint - I think the old search experience is the only user of this.

api/src/api/form_alpha/form_route.py
btabaska: I did, removed now.

api/src/api/agencies_v1/agency_routes.py
btabaska: Reverted this back to the original api auth.

api/src/auth/multi_auth.py
```diff
@@ -5,21 +5,23 @@
...
+from ..db.models.user_models import UserApiKey, UserTokenSession
 from .api_jwt_auth import api_jwt_auth
 from .api_key_auth import ApiKeyUser, api_key_auth
+from .api_user_key_auth import api_user_key_auth
 from .internal_jwt_auth import internal_jwt_auth
 
 
 class AuthType(StrEnum):
     API_KEY_AUTH = "api_key_auth"
+    api_user_key_auth = "api_user_key_auth"
```
chouinar:
```suggestion
    API_USER_KEY_AUTH = "api_user_key_auth"
```

api/openapi.generated.yml
```diff
@@ -65,12 +65,6 @@ paths:
               schema:
                 $ref: '#/components/schemas/ErrorResponse'
           description: Validation error
-        '401':
-          content:
-            application/json:
-              schema:
-                $ref: '#/components/schemas/ErrorResponse'
-          description: Authentication error
```
chouinar: Ah - forgot about one other thing we have to do with the multiauth to make it build the swagger docs happily.

Could you add this to the routes you changed to multiauth?:
```py
@whatever_blueprint.doc(security=api_key_multi_auth_security_schemes)
```
The `@whatever_blueprint.auth_required(whatever)` adds this normally, but that doesn't work with multiauth, so we have to manually add the docs ourselves.

btabaska:
Ready for final check.

Issue #6047, opened 2025-08-21
Update table v_opportunity_summary


babebe:
### Summary

Create migration file to drop existing table, and create new migration file for table `v_opportunity_summary`:

- Rename `opportunity_id` column to `legacy_opportunity_id`
- Update code/test that reference the column name
- Update type on field `omb_review_status_date` to be of type `datetime`

### Acceptance criteria

[ ] migration file to drop table and new migration file created for table
[ ] code and tests updated

PR #6076, opened 2025-08-22
[Issue #6047] Update vopporuntiy summary

babebe:
## Summary

Fixes / Work for #ISSUE


chouinar:
@babebe - is the title of this ticket/PR right? I think you put the wrong table - the release notes we generate use the PR title

babebe:
Yes i had the wrong table name! thank you!!

chouinar (approved):
LGTM!

When the deploy happens next week, there's a likely chance this could cause errors briefly. Because we deploy DB migrations before code changes by a few minutes, the queries might be trying to fetch the old table column name and fail. As all of these changes are either in backend tasks or an agency endpoint we don't use, that should be fine. Just give a heads up to Matt before the deploy to keep a lookout

PR #6078, opened 2025-08-22
Myduong/6042 multiselect widget

myduong-navapbc:
## Summary

Fixes / Work for #ISSUE

### DIFF
--- a/frontend/src/components/applyForm/widgets/MultiSelectWidget.tsx
+++ b/frontend/src/components/applyForm/widgets/MultiSelectWidget.tsx
@@ -51,12 +51,12 @@ export default function MultiSelect<
     enumOptions?: ComboBoxOption[];
     enumDisabled?: Array<string | number>;
   };
-// Future:
-// allows us to inforce the minimum number
-// we can later pass it as prop if design likes the idea
-// this will be used for some dropdowns that have the possibility of having one option
-// example would be starting an application and it is defaulted to "individual" and no other options are available
-const enforceMinInUI = false;
+  // Future:
+  // allows us to inforce the minimum number
+  // we can later pass it as prop if design likes the idea
+  // this will be used for some dropdowns that have the possibility of having one option
+  // example would be starting an application and it is defaulted to "individual" and no other options are available
+  const enforceMinInUI = false;
   const labelType = getLabelTypeFromOptions(options?.["widget-label"]);
   const allOptions: ComboBoxOption[] = useMemo(() => opts ?? [], [opts]);
 
@@ -65,7 +65,7 @@ const enforceMinInUI = false;
   const comboRef = useRef<ComboBoxRef>(null);
 
   const maxSelections = typeof maxItems === "number" ? maxItems : 3;
-    const minSelectionsEff = typeof minItems === 'number' ? minItems : 0;
+  const minSelectionsEff = typeof minItems === "number" ? minItems : 0;
   const atCap = selected.length >= maxSelections;
 
   const isOptDisabled = useCallback(
@@ -120,11 +120,11 @@ const enforceMinInUI = false;
     return true;
   };
 
-const removeValue = (v: string) => {
-  const next = selected.filter((s) => s !== v);
-  if (enforceMinInUI && next.length < minSelectionsEff) return;
-  syncUpstream(next);
-};
+  const removeValue = (v: string) => {
+    const next = selected.filter((s) => s !== v);
+    if (enforceMinInUI && next.length < minSelectionsEff) return;
+    syncUpstream(next);
+  };
 
   const getLabelForValue = (v: string) =>
     allOptions.find((o) => String(o.value) === v)?.label ?? v;
--- a/frontend/src/components/applyForm/widgets/MultiSelectWidget.tsx
+++ b/frontend/src/components/applyForm/widgets/MultiSelectWidget.tsx
@@ -1,0 +1,216 @@
+"use client";
+
+import {
+  enumOptionsValueForIndex,
+  FormContextType,
+  RJSFSchema,
+  StrictRJSFSchema,
+} from "@rjsf/utils";
+
+import React, { useCallback, useMemo, useRef } from "react";
+import {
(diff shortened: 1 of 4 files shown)


frontend/src/components/applyForm/widgets/MultiSelectWidget.tsx
```diff
@@ -0,0 +1,216 @@
...
+  type ComboBoxOption,
+  type ComboBoxRef,
+} from "@trussworks/react-uswds";
+
+import { UswdsWidgetProps } from "src/components/applyForm/types";
+import { Pill } from "src/components/Pill";
+import { DynamicFieldLabel } from "./DynamicFieldLabel";
+import { getLabelTypeFromOptions } from "./getLabelTypeFromOptions";
+
+function toArray(selectedValue: unknown): string[] {
```
doug-s-nava: [nit] since the first case here is to turn an array into an array of strings, maybe the name should be `toStringArray`?

frontend/src/components/applyForm/widgets/MultiSelectWidget.tsx
```diff
@@ -0,0 +1,216 @@
...
+  onFocus = () => ({}),
+}: UswdsWidgetProps<T, S, F>) {
+  const { title, description, maxItems, minItems } = schema;
+
+  const { enumOptions: opts, enumDisabled } = (options ?? {}) as {
+    enumOptions?: ComboBoxOption[];
+    enumDisabled?: Array<string | number>;
+  };
+  // Future:
+  // allows us to inforce the minimum number
```
doug-s-nava: [sp] 
```suggestion
  // allows us to enforce the minimum number
```

frontend/src/components/applyForm/widgets/MultiSelectWidget.tsx
```diff
@@ -0,0 +1,216 @@
...
+  // example would be starting an application and it is defaulted to "individual" and no other options are available
+  const enforceMinInUI = false;
+  const labelType = getLabelTypeFromOptions(options?.["widget-label"]);
+  const allOptions: ComboBoxOption[] = useMemo(() => opts ?? [], [opts]);
+
+  const [selected, setSelected] = React.useState<string[]>(toArray(value));
+
+  const comboRef = useRef<ComboBoxRef>(null);
+
+  const maxSelections = typeof maxItems === "number" ? maxItems : 3;
```
doug-s-nava, line 69: how'd you arrive on 3 as the default? seems like we would want to set no max items if it isn't specified in the schema

frontend/src/components/applyForm/widgets/MultiSelectWidget.tsx
```diff
@@ -0,0 +1,216 @@
...
+    (opt: ComboBoxOption) =>
+      !!enumDisabled &&
+      (enumDisabled.includes(opt.label) || enumDisabled.includes(opt.value)),
+    [enumDisabled],
+  );
+
+  const availableOptions = useMemo(
+    () =>
+      allOptions.filter(
+        (o) => !selected.includes(String(o.value)) && !isOptDisabled(o),
```
doug-s-nava: [nit but maybe more than that] would love to not abbreviate argument and variable names like this here and elsewhere in this file. Can we use `option` instead?

frontend/src/components/applyForm/widgets/MultiSelectWidget.tsx
```diff
@@ -0,0 +1,216 @@
...
+      : undefined;
+
+  const syncUpstream = (next: string[]) => {
+    setSelected(next);
+    onChange(
+      enumOptionsValueForIndex<S>(next, allOptions, undefined) as unknown as T,
+    );
+  };
+
+  const addByLabelOrValue = (raw: string): boolean => {
```
doug-s-nava: it doesn't look like we're using the return value here - I think it's potentially confusing to return a boolean in that case. Can we remove that?

frontend/src/components/applyForm/widgets/MultiSelectWidget.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+    const next = [...selected, v];
+    syncUpstream(next);
+    comboRef.current?.clearSelection();
+    return true;
+  };
+
+  const removeValue = (v: string) => {
+    const next = selected.filter((s) => s !== v);
+    if (enforceMinInUI && next.length < minSelectionsEff) return;
```
doug-s-nava, line 127: are we potentially swallowing an error here? If we don't show an error and we don't update the field so that the error can get caught in validation on save, where would the user see that something is wrong?

doug-s-nava (approved):
a little nervous that none of this broke any tests. Can we make a follow up ticket to add some testing around this?

from a ui perspective, I wonder if the select should still be usable if you've reached the maximum, but wouldn't let you select more options? something to consider for later

hey, it works, nice job! nothing I have is blocking, though the I'd appreciate a thought for the non-nit comments, maybe as a fast follow

✅ ✅ ✅ ✅

frontend/src/components/applyForm/widgets/MultiSelectWidget.tsx
myduong-navapbc: this one, i didnt have any direction. i chose 3 at because it just happened to be the max amount for the `applicant_type` field. this is for sure something we can discuss.

frontend/src/components/applyForm/widgets/MultiSelectWidget.tsx
myduong-navapbc: good point. ^

doug-s-nava (approved):

frontend/src/components/applyForm/widgets/MultiSelectWidget.tsx
acouch: This got merged already, but this should come from the schema, most likely [maxItems](https://www.learnjsonschema.com/2020-12/validation/maxitems/).

Issue #5783, opened 2025-07-28
FE – Developer portal route (/developer)


btabaska:
### Summary

Add a developer page for logged in users

### Acceptance criteria

- [ ] Navigation item “Developer”.
- [ ] Page describes API, links to docs & rate limits.

PR #6094, opened 2025-08-25
[#5783] - Developer Page & API Dashboard

btabaska:
## Summary

Fixes for #5783
This PR implements a comprehensive Developer Portal and API Dashboard feature that enables developers to access API documentation and manage their API keys for the Simpler.Grants.gov API.
## Changes proposed

New developer information page with API capabilities overview
Explains what developers can and cannot do with the current API
Features contextual authentication - shows "Manage API Keys" button for logged-in users, login modal for anonymous users
Includes icon sections with links to external developer resources
Added "Developers" link to the header navigation under "Connect" section

API Dashboard (/api-dashboard)
Full CRUD operations for API key management
View all API keys with creation dates, last used dates, and status
Create new API keys with custom names
Rename existing API keys
Visual table display showing key names, key IDs, and usage information
Real-time loading states and error handling

Complete API route handlers for key management (/api/user/api-keys/*)
TypeScript types for API key objects
Client-side fetching utilities with proper error handling
Comprehensive test coverage for all new components and API routes

## Context for reviewers

delete functionality for API keys is still in review on backend and will be implemented in a separate PR, all other functionality should be working. 

[Figma](https://www.figma.com/design/sgKE2z5uFU6R6HVxpt224Y/Simpler.Grants.gov-API-key-portal?node-id=8103-618&t=02rPrXGt4new799l-0)

## Validation steps

- [ ] Test the Developer Portal:
  - [ ] Navigate to /developer
    - [ ] As anonymous user: Verify you see the API information and a login modal when clicking "Manage API Keys"
    - [ ] As authenticated user: Verify you see a direct link to the API Dashboard
- [ ] Test the API Dashboard:
  - [ ] Log in and navigate to /api-dashboard
    - [ ] Create API Key:
      - [ ] Click "Create New API Key" button
      - [ ] Enter a name and submit
      - [ ] Verify success message and key appears in table
      - [ ] Try creating a key with an empty name to test validation
    - [ ] Rename API Key:
      - [ ] Click edit button next to any key
      - [ ] Change the name and save
      - [ ] Verify success message and updated name in table
      - [ ] Test renaming a key to the same name to verify validation
    - [ ] View Key Details:
      - [ ] Verify table shows: key name, key ID, creation date, last used date
      - [ ] Check responsive behavior on mobile devices
- [ ] Test Navigation:
  - [ ] Verify "Developers" appears under "Connect" in the main navigation
  - [ ] Test that clicking it navigates to the developer portal
  - [ ] Verify navigation works on both desktop and mobile viewports


frontend/src/app/[locale]/(base)/api-dashboard/page.tsx
```diff
@@ -0,0 +1,89 @@
+"use client";
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/api-dashboard/page.tsx]: I'd suggest doing what we can to keep this from rendering client side for as long as we can. Some of the client side behavior here can be moved to the backend (making the request for keys), and the loading behavior can be handled with suspense

I do see the idea here of creating the fetch function once, then passing it down for the modal to use, but that has its own issues potentially. If we do the whole thing client side you'd just have to remove the "use client" from the modal component to let Next know that we'll already be in a client rendering context for that file, in order to allow passing non server action functions as props. If we refactor to use a server component here, I think the best course of action would be to make a server side call here, and then refresh the page after any updates from the modal to trigger a refetch

frontend/src/components/Header.tsx
```diff
@@ -93,6 +93,7 @@ const NavLinks = ({
   const { checkFeatureFlag } = useFeatureFlags();
   const showSavedSearch = checkFeatureFlag("savedSearchesOn");
   const showSavedOpportunities = checkFeatureFlag("savedOpportunitiesOn");
+  const showDeveloperPortal = !checkFeatureFlag("developerPageOff");
```
doug-s-nava, line 96: when a feature flag is only being referenced within a component, rather than with the "withFeatureFlag" hook to enable page level behavior, we should use positive polarity on the flag. Can we change this to `developerPageOn`?

frontend/src/components/developer/DeveloperSections.tsx
```diff
@@ -0,0 +1,9 @@
+import DeveloperInfo from "src/components/developer/sections/DeveloperInfo";
```
doug-s-nava, line 1: do we have plans to add more to this file? If not I'd say we can delete it and go straight to DeveloperInfo

frontend/src/components/developer/sections/DeveloperInfo.tsx
```diff
@@ -0,0 +1,87 @@
...
+            className="margin-y-2 usa-button--secondary"
+            type="button"
+            size="big"
+          >
+            {t("apiDashboardLink")}
+          </Button>
+        </Link>
+      ) : (
+        <>
+          <ModalToggleButton
```
doug-s-nava: I'd hope that we could use `LoginButtonModal` for any situation where we want a button to open a login modal, but now I'm seeing that it looks like that component isn't being used anywhere. 😮‍💨  

What if you updated that component so that you could use it here? Or if that's not worth it, maybe just delete the `LoginButtonModal` and be done with it?

frontend/src/components/developer/sections/DeveloperInfo.tsx
```diff
@@ -0,0 +1,87 @@
...
+  const { iconSections } = messages.Developer;
+  const { user } = useUser();
+  const modalRef = useRef<ModalRef>(null);
+
+  return (
+    <HomePageSection className="bg-base-white" title={t("infoTitle")}>
+      <h3 data-testid="developer-info">{t("canDoHeader")}</h3>
+      <h4>{t("canDoSubHeader")}</h4>
+      <p>{t("canDoParagraph")}</p>
+      {user?.token ? (
```
doug-s-nava: if you split this part out, you can have a server component here, and render everything but the buttons server side

frontend/src/components/developer/sections/DeveloperInfo.tsx
```diff
@@ -0,0 +1,87 @@
...
+const DeveloperInfoContent = () => {
+  const t = useTranslations("Developer");
+  const headerTranslations = useTranslations("HeaderLoginModal");
+  const messages = useMessages() as unknown as IntlMessages;
+  const { iconSections } = messages.Developer;
+  const { user } = useUser();
+  const modalRef = useRef<ModalRef>(null);
+
+  return (
+    <HomePageSection className="bg-base-white" title={t("infoTitle")}>
```
doug-s-nava: Is this home page section meant to be reused like this? I'm a little nervous about reusing something that doesn't seem like it was meant to be reused outside the context of the homepage

frontend/src/components/developer/apiDashboard/ApiKeyModal.tsx
```diff
@@ -0,0 +1,365 @@
...
+    ? "create-api-key"
+    : `edit-api-key-${apiKey?.api_key_id ?? "unknown"}`;
+  const titleText = success
+    ? undefined
+    : isCreate
+      ? t("createTitle")
+      : t("editTitle", { keyName: apiKey?.key_name ?? "" });
+
+  // Default trigger button configurations
+  const defaultTriggerProps = {
```
doug-s-nava, line 335: [not requesting a change!] I'm somewhat on board with the mode acting as a switch between two different modal experiences up until this point. I'd prefer a world where all of the create vs edit based configuration is passed in explicitly rather than derived from a string based switch, but it's really a philosophical difference. I wish we had a more fleshed out modal implementation that would handle cases like this better without creating a lot of very similar but crucially different components. oh well

frontend/src/app/api/user/api-keys/route.ts
```diff
@@ -0,0 +1,5 @@
+import { respondWithTraceAndLogs } from "src/utils/apiUtils";
```
doug-s-nava, line 1: if we refactor to a server component, we can remove this

frontend/tests/api/user/api-keys/route.test.ts
```diff
@@ -0,0 +1,190 @@
...
+      message: string;
+    }>,
+}));
+
+interface JsonData {
+  message: string;
+  data?: ApiKey;
+}
+
+const mockApiKeyResponse: ApiKey = {
```
doug-s-nava: could be worth creating an API key fixture that we can reuse across tests in the fixtures file

doug-s-nava (commented):
a couple points to talk about, but everything works as expected. The only thing to call out on the functionality side is that maybe we don't add the "delete" column until we're ready to implement it? At this point the user will only see a disabled hover state on what looks like a functional delete button, which may be confusing.

btabaska:
> a couple points to talk about, but everything works as expected. The only thing to call out on the functionality side is that maybe we don't add the "delete" column until we're ready to implement it? At this point the user will only see a disabled hover state on what looks like a functional delete button, which may be confusing.

The delete functionality was originally going to be a fast follow, but because of how this PR lined up I just implemented it. It is now in this PR as well and also reuses the modal component and much of the existing rename functionality.

frontend/src/components/developer/DeveloperSections.tsx
btabaska: Yes, this is just an MVP with plans in the future to be expanded like we did with the roadmap and vision page.

frontend/src/components/developer/sections/DeveloperInfo.tsx
btabaska: I think this falls outside the scope of this PR. I agree that the login modal isn't used here or anywhere else. We could update it for it to be more generic to be used in multiple places, or deprecate it. I just don't think that is within the scope of an MVP PR like this that is already implementing so much.

frontend/src/components/developer/sections/DeveloperInfo.tsx
btabaska: Great point, I just went ahead and created a new component to follow the pattern set in other pages like this one. I think in the future it might make sense to go ahead and make a generic pageSection component for reuse since it is such a commonly used pattern. In this case it doesn't matter much though.

frontend/src/components/developer/sections/DeveloperInfo.tsx
btabaska: This makes sense, I updated the page based on feedback and I see it performs much better.

frontend/src/components/Header.tsx
btabaska: This feature flag was implemented in another PR and is already live in prod. Because this FF will be short lived and will likely be deprecated shortly after this feature goes to prod I'll leave it as-is. If we planned on keeping this around for a long time I agree that it could be named better.

frontend/tests/api/user/api-keys/route.test.ts
btabaska: Yeah that makes sense, I went ahead and did that. Updated all unit tests to now use the fixture to remove code duplication.

btabaska:
Ready for next pass review. All feedback has been iterated on.

frontend/src/app/[locale]/(base)/api-dashboard/page.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export default async function ApiDashboardPage({ params }: LocalizedPageProps) {
+  const { locale } = await params;
+  const t = await getTranslations({ locale, namespace: "ApiDashboard" });
+  let apiKeys;
+
+  try {
+    apiKeys = await fetchApiKeys();
+  } catch (e) {
+    return (
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/api-dashboard/page.tsx:22]: would be a good idea to log errors here as well

frontend/src/components/developer/sections/DeveloperInfoLoading.tsx
```diff
@@ -0,0 +1,43 @@
+import { Grid, GridContainer } from "@trussworks/react-uswds";
+
+export default function DeveloperInfoLoading() {
```
doug-s-nava: is this being used anywhere?

frontend/src/services/fetch/fetchers/apiKeyClientHelpers.ts
```diff
@@ -0,0 +1,42 @@
+// Client-side API key fetching helpers for use with useClientFetch
```
doug-s-nava, line 1: it looks like most (all?) of the stuff in this file isn't being used anymore. can you double check?

frontend/src/app/api/user/api-keys/route.ts
doug-s-nava: it looks like we're going straight from a server component to the API for this data, so I don't think we need this route anymore

doug-s-nava (commented):
everything works great - thanks for digging into the review, and love all the tests. I think I found a few things that aren't being used anymore, but otherwise this is ready to go

doug-s-nava:
something i forgot to add in the review, not a blocker, but the create button looks big on mobile
Screenshot 2025-09-03 at 7 51 59 AM

andycochran:
w/o having looked at the code, that big button is likely a result of using unnecessary `flex` classes. Our grids don't need to be flex grids.

btabaska:
> w/o having looked at the code, that big button is likely a result of using unnecessary `flex` classes. Our grids don't need to be flex grids.

This is correct, I went and removed the classes and the button is now looking as expected in desktop and mobile.

btabaska:
> it looks like we're going straight from a server component to the API for this data, so I don't think we need this route anymore

You're absolutely correct in your observation - the server component does go directly to the backend API and doesn't need this Next.js API route. However, the client components (like ApiKeyModal) still use these routes for client-side operations.
The architecture is:
Server Component (page.tsx): Uses fetchApiKeys() → Direct backend API call
Client Components (ApiKeyModal.tsx): Use Next.js API routes → Then to backend API

PR #6107, opened 2025-08-26
Infra 6106 - Attach additional ALB certs and tidy up gateway

### LINKED ISSUE #6106: API Gateway - Attach Secondary Cert to ALB & Enable Logging
### Summary

We need to update the Application Load Balancer (ALB) to include the secondary certificate and enable logging. The IAM role has already been created and attached. This PR will focus on wiring up the cert and logging configuration, ensuring variables are properly passed through despite current complexity in the setup.
The next/last PR in this sequence will likely involve setting the domain association to the API Gateway. However, that cannot be fully tested until DNS is cut over. For now, this PR can be tested independently without touching DNS, since DNS management occurs outside of the account.

### Acceptance criteria

- [ ]  Secondary cert successfully attached to ALB.
- [ ]  Logging enabled and verified on ALB.
- [ ]  IAM role usage confirmed.
- [ ]  Variable passing reviewed and updated for maintainability.
- [ ]  PR validated without requiring DNS cutover.

Nava-JoshLong:
## Summary

Fixes #6106

## Changes proposed

- Logic for optionally attaching extra certs to ALB
- Enable API gateway logging
- Point API gateway to secondary domain on ALB

## Context for reviewers

This PR has two main parts to it, adding logic for optionally attaching secondary certs to the ALB, and tidying up the API gateway setup. 

### ALB changes

For this part, I wanted to setup the ALB changes to be modular enough that it won't break in case we pass no extra domains to it, but allow it to take in more than a single extra domain in case more are added in the future. This is done in preparation for the cutover for the public DNS for the api.*.simpler... to point from the ALB to the API gateway. The DNS is already setup for alb.*.simpler..., so we can point the API gateway to that, and be ready for the cutover when it happens

### API Gateway changes

This part has a change for enabling logging, and a change to the logic for what domain the proxy forwards to. For the logging change, there was an issue from my first PR where I thought the accounts playbook would be deployed via terraform, but that didn't happen. I merged a second PR that disabled it until I got accounts deployed, which needed to be done manually and is now set. Now we can enable it again and have it log as expected. The other change is to point the gateway proxy from the main domain, `api...`, to the secondary, `alb...`, so that the cutover is only required to be done on the DNS side. This will let us test the cutover before going live

There might be another change to enable the domain association, which is currently commented out, but I'll have to test to make sure it won't break anything first. I'll edit the PR body if I can enable it with a blurb about it

## Validation steps

Validation will be done manually. I'll call out in Slack before testing, then deploy the changes and test. Once things are validated, I'll update the PR and revert the changes so that I don't introduce drift

### DIFF
--- a/infra/modules/service/api_gateway.tf
+++ b/infra/modules/service/api_gateway.tf
@@ -44,8 +44,8 @@ resource "aws_api_gateway_integration" "api_proxy_integration" {
 
   integration_http_method = aws_api_gateway_method.api_proxy_method[0].http_method
   type                    = "HTTP_PROXY"
-  # This will be changed to use a new ALB DNS and cert
-  uri = "https://${var.domain_name}/{proxy}"
+
+  uri = "https://${var.optional_extra_alb_domains[0]}/{proxy}"
 
   request_parameters = {
     "integration.request.path.proxy" = "method.request.path.proxy"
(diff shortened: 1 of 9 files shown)


mdragon (dismissed):
👍🏻

Nava-JoshLong:
Latest commit changes from defining endpoints via terraform resources to the JSON definition. This is for a few reasons

- Much less lines of code required to create these endpoints
- Easier to convert manual changes/testing/new features into code
- Much less messy
- No need to add counts to every resource

This hasn't been tested yet, but was dumped from testing with Matt and Brandon, so it should work. I just need to run a deploy or two to confirm

infra/modules/service/api_gateway.tf
```diff
@@ -6,61 +6,118 @@ resource "aws_api_gateway_rest_api" "api" {
...
+        "get" : {
+          "x-amazon-apigateway-integration" : {
+            "type" : "http_proxy",
+            "httpMethod" : "GET",
+            "uri" : "https://${var.optional_extra_alb_domains[0]}/health",
+            "passthroughBehavior" : "when_no_match"
+          }
+        }
+      },
+      "/{proxy+}" : {
```
mdragon, line 22: Does this need to go last as the catch all? Should it go last for humans to make it clearer it's a catch all even if the technology doesn't need it that way?

mdragon (dismissed):
Approved, with one question about the ordering on the JSON.

infra/modules/service/api_gateway.tf
Nava-JoshLong: I'm not sure it matters? I think the logic is set so that any matching endpoints defined that aren't proxies are used first, then proxies. I'll be able to tell with deploying and testing it. I can put it at the end if you want, this was literally just taken from the export of the gateway and put here, so any updates would need to reorder them then

Nava-JoshLong: I tested by moving this to the last in the list, and it disappears from AWS somehow. I'm moving it back to where it was and seeing if things are happy again

Screenshot 2025-09-02 at 2 25 43 PM

Nava-JoshLong: Got things working again. The order in the JSON doesn't map to what's in AWS, and the ordering doesn't seem to matter since it's alphabetical on AWS

PR #6108, opened 2025-08-26
[#5780] - API Gateway Client & Local Mock

### LINKED ISSUE #5780: BE – Implement ApiGatewayService for key CRUD
### Summary

Create a service class (ApiGatewayService) using boto3 to call import_api_keys, update_api_key, delete_api_key and attach/detach usage plans.

### Acceptance criteria

- [ ] Wrapper methods return structured result & raise custom errors.
- [ ] Integration tests hit AWS sandbox account via mocked STS role api-key-sync.
- [ ] Unit‑test coverage.

btabaska:
## Summary

Fixes #5780  
Adds AWS API Gateway integration for API key management, enabling automatic import of user-generated API keys into AWS API Gateway with usage plan association and comprehensive error handling.

## Changes proposed

- Created api_gateway_adapter.py with functions to import API keys into AWS API Gateway using CSV format, associate them with usage plans, and handle both real AWS and local mock environments
-Modified create_api_key.py service to automatically import API keys to AWS API Gateway after local database creation, with graceful degradation if AWS is unavailable
- Added full test suites for both the API Gateway adapter and enhanced API key creation service, including mock response handling and error scenarios
- Added API Gateway configuration support through environment variables for usage plan association

### DIFF
--- a/api/src/adapters/aws/api_gateway_adapter.py
+++ b/api/src/adapters/aws/api_gateway_adapter.py
@@ -26,9 +26,7 @@ class ApiGatewayConfig(PydanticBaseEnvConfig):
     """Configuration for AWS API Gateway integration"""
 
     # Usage plan ID for newly created API keys (typically the public usage plan)
-    default_usage_plan_id: str | None = Field(
-        default=None, alias="API_GATEWAY_DEFAULT_USAGE_PLAN_ID"
-    )
+    default_usage_plan_id: str = Field(alias="API_GATEWAY_DEFAULT_USAGE_PLAN_ID")
 
 
 def get_boto_api_gateway_client(session: boto3.Session | None = None) -> botocore.client.BaseClient:
--- a/api/src/services/users/create_api_key.py
+++ b/api/src/services/users/create_api_key.py
@@ -82,23 +82,14 @@ def _import_api_key_to_aws_gateway(api_key: UserApiKey) -> None:
             usage_plan_id=config.default_usage_plan_id,
         )
 
-        if config.default_usage_plan_id:
-            logger.info(
-                "Successfully imported API key to AWS API Gateway and associated with usage plan",
-                extra={
-                    "api_key_id": api_key.api_key_id,
-                    "gateway_key_id": gateway_response.id,
-                    "usage_plan_id": config.default_usage_plan_id,
-                },
-            )
-        else:
-            logger.info(
-                "Successfully imported API key to AWS API Gateway (no usage plan configured)",
-                extra={
-                    "api_key_id": api_key.api_key_id,
-                    "gateway_key_id": gateway_response.id,
-                },
-            )
+        logger.info(
+            "Successfully imported API key to AWS API Gateway and associated with usage plan",
+            extra={
+                "api_key_id": api_key.api_key_id,
+                "gateway_key_id": gateway_response.id,
+                "usage_plan_id": config.default_usage_plan_id,
+            },
+        )
 
     except Exception as e:
         # Re-raise as a domain-specific exception without additional logging
--- a/api/tests/src/adapters/aws/test_api_gateway_adapter.py
+++ b/api/tests/src/adapters/aws/test_api_gateway_adapter.py
@@ -15,11 +15,16 @@ from src.adapters.aws.api_gateway_adapter import (
 class TestApiGatewayConfig:
     """Test the configuration class for API Gateway."""
 
-    def test_config_defaults(self):
-        """Test that configuration has reasonable defaults."""
-        config = ApiGatewayConfig()
+    def test_config_defaults(self, monkeypatch):
+        """Test that configuration requires default_usage_plan_id to be set."""
+        from pydantic import ValidationError
 
-        assert config.default_usage_plan_id is None
+        # Clear any existing environment variable
+        monkeypatch.delenv("API_GATEWAY_DEFAULT_USAGE_PLAN_ID", raising=False)
+
+        # Should raise a ValidationError when required field is missing
+        with pytest.raises(ValidationError):
+            ApiGatewayConfig()
 
     def test_config_from_environment(self, monkeypatch):
         """Test that configuration can be loaded from environment variables."""
--- a/api/local.env
+++ b/api/local.env
@@ -115,9 +115,9 @@ IS_LOCAL_AWS=1
 # AWS API Gateway
 ############################
 
-# Optional: Default usage plan ID for newly created API keys
-# Leave commented out if no default usage plan should be associated
-# API_GATEWAY_DEFAULT_USAGE_PLAN_ID=your-usage-plan-id-here
+# Default usage plan ID for newly created API keys
+# For local development, use a placeholder value since we mock AWS calls
+API_GATEWAY_DEFAULT_USAGE_PLAN_ID=local-dev-usage-plan
 
 ############################
 # Feature Flags


api/src/adapters/aws/api_gateway_adapter.py
```diff
@@ -0,0 +1,156 @@
...
+
+    Returns:
+        ApiKeyImportResponse with the imported key details
+    """
+
+    if api_gateway_client is None:
+        api_gateway_client = get_boto_api_gateway_client()
+
+    if is_local_aws():
+        return _handle_mock_import_response(api_key, name, description, enabled, usage_plan_id)
```
chouinar: Should these be flipped? No need to make a client if we don't use it in the mock.

api/src/adapters/aws/api_gateway_adapter.py
```diff
@@ -0,0 +1,156 @@
...
+
+    try:
```
chouinar, line 81: Could we put the try block around just the API gateway queries?

api/src/adapters/aws/api_gateway_adapter.py
```diff
@@ -0,0 +1,156 @@
...
+            "Error importing API key to AWS API Gateway",
+            extra={
+                "error_code": e.response["Error"]["Code"],
+                "error_message": e.response["Error"]["Message"],
+            },
+        )
+        raise
+    except Exception as e:
+        logger.exception("Unexpected error importing API key", extra={"error": str(e)})
+        raise
```
chouinar, line 113: logger.exception adds the error automatically to extra in a few different formats (this is what we specifically do with the exc_info object): https://github.com/HHS/simpler-grants-gov/blob/088b3a217c9d6f1c8d2cd448668b9b719cd94e87/api/src/logging/flask_logger.py#L221

Assuming the stack trace of the error has the response object somewhere in it, we can skip that and it'll get put in the proper place in our logs.

api/src/adapters/aws/api_gateway_adapter.py
```diff
@@ -0,0 +1,156 @@
...
+
+        if response.get("warnings"):
+            logger.warning("API Gateway import warnings", extra={"warnings": response["warnings"]})
+
+        imported_key_ids = response.get("ids", [])
+        if not imported_key_ids:
+            raise Exception("No API key IDs returned from import operation")
+
+        key_id = imported_key_ids[0]
+        key_details = api_gateway_client.get_api_key(apiKey=key_id, includeValue=False)
```
chouinar: It shouldn't happen, but should we add a sanity check that it returned exactly 1 key?

api/src/services/users/create_api_key.py
```diff
@@ -46,6 +53,21 @@ def create_api_key(db_session: db.Session, user_id: UUID, json_data: dict) -> Us
     )
     db_session.add(api_key)
 
+    # Import the API key to AWS API Gateway
+    try:
+        _import_api_key_to_aws_gateway(api_key)
+    except Exception as e:
+        # Log the error but don't fail the API key creation
+        # This allows the system to still function if AWS API Gateway is temporarily unavailable
```
chouinar: Can we do that? In the short-term before API Gateway fronts our API it would be fine, but later the key wouldn't be valid as it's not in API Gateway so no queries would work.

api/src/services/users/create_api_key.py
```diff
@@ -59,6 +81,50 @@ def create_api_key(db_session: db.Session, user_id: UUID, json_data: dict) -> Us
...
+        logger.exception(
+            "Error importing API key to AWS API Gateway",
+            extra={
+                "api_key_id": api_key.api_key_id,
+                "key_name": api_key.key_name,
+            },
+        )
+        raise ApiGatewayIntegrationError(
+            f"Failed to import API key to AWS API Gateway: {str(e)}"
+        ) from e
```
chouinar: If we have an except that logs this in the library, and above this, do we need this try-except?

api/src/adapters/aws/api_gateway_adapter.py
```diff
@@ -0,0 +1,156 @@
...
+        api_key_response = ApiKeyImportResponse.model_validate(key_details)
+
+        if usage_plan_id:
+            api_gateway_client.create_usage_plan_key(
+                usagePlanId=usage_plan_id, keyId=api_key_response.id, keyType="API_KEY"
+            )
+            logger.info(
+                "Associated API key with usage plan",
+                extra={"api_key_id": api_key_response.id, "usage_plan_id": usage_plan_id},
+            )
```
chouinar: I think we can add a usage plan in the API gateway CSV import approach and not need a follow-up query: https://docs.aws.amazon.com/apigateway/latest/developerguide/api-key-file-format.html

api/src/adapters/aws/api_gateway_adapter.py
```diff
@@ -0,0 +1,156 @@
...
+    if api_gateway_client is None:
+        api_gateway_client = get_boto_api_gateway_client()
+
+    if is_local_aws():
+        return _handle_mock_import_response(api_key, name, description, enabled, usage_plan_id)
+
+    try:
+        # Format the API key data as CSV for import
+        # AWS API Gateway expects CSV format: name,key,description,enabled
+        csv_data = f"{name},{api_key},{description or ''},{'true' if enabled else 'false'}"
```
chouinar: Should the format have the header row in the CSV?

api/tests/src/adapters/aws/test_api_gateway_adapter.py
```diff
@@ -0,0 +1,297 @@
...
+class TestApiGatewayConfig:
+    """Test the configuration class for API Gateway."""
+
+    def test_config_defaults(self):
+        """Test that configuration has reasonable defaults."""
+        config = ApiGatewayConfig()
+
+        assert config.default_usage_plan_id is None
+
+    @patch.dict("os.environ", {"API_GATEWAY_DEFAULT_USAGE_PLAN_ID": "test-plan-123"})
```
chouinar: We usually use monkeypatch for setting env vars in tests.

```py
def test_thing(monkeypatch):
    monkeypatch.setenv("MY_ENV_VAR", "value")
```

api/tests/src/services/users/test_create_api_key.py
```diff
@@ -12,7 +12,8 @@
 from tests.src.db.models.factories import UserApiKeyFactory, UserFactory
 
 
-def test_create_api_key_success(enable_factory_create, db_session: db.Session):
+@patch("src.services.users.create_api_key.import_api_key")
```
chouinar: What is this patch doing? With the logic defaulting to the mock with the local AWS stuff, does it not work?

api/src/adapters/aws/api_gateway_adapter.py
btabaska: Yep, that makes sense. I just flipped it so that we only create the API Gateway client when we need it for real AWS operations.

api/src/adapters/aws/api_gateway_adapter.py
btabaska: I went through and modified the try block to wrap only the API GW queries and kept the response processing outside of them.

api/src/adapters/aws/api_gateway_adapter.py
btabaska: Per AWS docs it will always return exactly 1 API Key. 
https://docs.aws.amazon.com/apigateway/latest/api/API_GetApiKey.html

So it is only possible to get 0 or 1 key.

api/src/services/users/create_api_key.py
btabaska: Good point, I'll update the code accordingly to fail API Key creation if API gateway is unavailable.

api/src/services/users/create_api_key.py
btabaska: Removed the redundant exception logging.

api/src/adapters/aws/api_gateway_adapter.py
btabaska: Updated to do this all in one query instead of the double call approach I used prior.

api/src/adapters/aws/api_gateway_adapter.py
btabaska: It is optional to do this, but I can see how it could be useful for logging, etc. I just went ahead and added it.

api/tests/src/services/users/test_create_api_key.py
btabaska: You're right, I think this is an artifact from an earlier iteration of this work that I forgot to remove since it does nothing. I have gone ahead and removed it.

api/src/adapters/aws/api_gateway_adapter.py
```diff
@@ -0,0 +1,167 @@
...
+    try:
+        response = api_gateway_client.import_api_keys(
+            body=csv_data.encode("utf-8"), format="csv", failOnWarnings=True
+        )
+    except ClientError:
+        logger.exception("Error importing API key to AWS API Gateway")
+        raise
+    except Exception as e:
+        logger.exception("Unexpected error importing API key", extra={"error": str(e)})
+        raise
```
chouinar: logger.exception adds the error automatically. Same with the except a bit below here
```suggestion
    except Exception:
        logger.exception("Unexpected error importing API key")
        raise
```

api/src/services/users/create_api_key.py
```diff
@@ -46,6 +53,24 @@ def create_api_key(db_session: db.Session, user_id: UUID, json_data: dict) -> Us
...
     db_session.add(api_key)
 
+    # Import the API key to AWS API Gateway
+    try:
+        _import_api_key_to_aws_gateway(api_key)
+    except Exception as e:
+        # Remove the API key from the database since AWS API Gateway integration failed
+        # This ensures consistency - if the key can't be created in API Gateway,
+        # it shouldn't exist in our database either
+        db_session.rollback()
```
chouinar: If this logic is wrapped in a `with db_session.begin()` then we don't need to do a rollback as that already makes a transaction that wasn't committed. When the `with` block ends it'll commit. If there is any error, it already does the rollback.

api/src/services/users/create_api_key.py
```diff
@@ -46,6 +53,24 @@ def create_api_key(db_session: db.Session, user_id: UUID, json_data: dict) -> Us
...
+        logger.error(
+            "Failed to import API key to AWS API Gateway, rolling back database transaction",
+            extra={
+                "api_key_id": api_key.api_key_id,
+                "key_name": key_name,
+                "error": str(e),
+            },
+        )
```
chouinar: I think we don't need this try/except or log message. If it fails, let the error propagate up, we already logged a more specific error in the API gateway code / will log the error message when the API itself catches in to construct the error response.

api/src/services/users/create_api_key.py
```diff
@@ -59,6 +84,45 @@ def create_api_key(db_session: db.Session, user_id: UUID, json_data: dict) -> Us
...
+                },
+            )
+        else:
+            logger.info(
+                "Successfully imported API key to AWS API Gateway (no usage plan configured)",
+                extra={
+                    "api_key_id": api_key.api_key_id,
+                    "gateway_key_id": gateway_response.id,
+                },
+            )
```
chouinar: Is there a scenario where we wouldn't have a default usage plan ID defined? Should the value be nullable?

It's fine if we have to define a value locally even though it doesn't talk to AWS. Just a string in local.env

api/src/services/users/create_api_key.py
```diff
@@ -59,6 +84,45 @@ def create_api_key(db_session: db.Session, user_id: UUID, json_data: dict) -> Us
...
+
+    except Exception as e:
+        # Re-raise as a domain-specific exception without additional logging
+        # since the AWS adapter already logs the underlying error
+        raise ApiGatewayIntegrationError(
+            f"Failed to import API key to AWS API Gateway: {str(e)}"
+        ) from e
```
chouinar: The `from e` bit should attach the stacktrace/error stuff to the error and not require adding it to the message of the error.

I appreciate you adding through error handling and logging, but we can really keep things simple, we don't need to parse exceptions in our try-excepts in most cases. The API's error logic should just handle adding all the contents of an error to our logs.

api/src/services/users/create_api_key.py
btabaska: This makes sense, I didn't know that is how it works. I'll update the code to simplify down.

api/src/services/users/create_api_key.py
btabaska: This makes sense. I'll go ahead and update the code and simplify down to remove this extra try/except block.

api/src/services/users/create_api_key.py
btabaska: For local development, you can either:
Leave it commented out (recommended for pure local development)
Set a dummy string value like API_GATEWAY_DEFAULT_USAGE_PLAN_ID=local-dev-plan if you want to test the logging paths****

Was my thought process when setting this up. I think it makes it easier to have local development set up for developers since I assume a minority will event want to interact with this code.

api/src/services/users/create_api_key.py
btabaska: I simplified out the approach and removed it from the message of the error.

api/src/services/users/create_api_key.py
chouinar: Why do you recommend commenting it out? Having the value just set in local.env and removing these extra code-paths feels simpler / better setup for developers.

btabaska: Yeah, that is ugly. I think I misunderstood this comment the first time that I read it. 
I went in and made default_usage_plan_id required, then updated local.env with a placeholder. 
I then removed the extra logging in create_api_key and then updated the unit tests. 

I think this matches the behavior that you were talking about. I think there may still be some questions about how we plumb through the default usage plan.

Issue #5538, opened 2025-07-01
Embed the Latest Newsletter on the “Subscribe” Page


btabaska:
### Summary

**What problem are we solving?**
New users often land on the Subscribe page but have no sense of the valuable content they’ll receive. They must leave the page to hunt for past newsletters, leading to drop-offs.

**What do we propose?**
Automatically display (embed) the most recent Simpler Grants newsletter directly beneath the signup form. The embed updates whenever a new newsletter is published—no manual work required.

**Why does it matter to grant seekers/writers?**
Seeing real examples of funding tips, deadline alerts, and success stories builds trust and clarifies the benefit of subscribing, increasing conversions and ensuring more stakeholders stay informed.

**How we’ll know it worked (success metrics)**

Increase in newsletter sign-ups three months post-launch
Lower bounce rate on the Subscribe page

**Effort & dependencies (non-technical view)**
Low effort: content already exists; requires one-time template update plus automated feed.

**Risks / considerations**

Need clear fallback if the feed fails.
Confirm accessibility (alt text, heading levels).

### Acceptance criteria

- [ ] Added the ability to view the latest newsletter directly from the Subscribe page

PR #6109, opened 2025-08-26
[Issue #6103] Sync Fider with GitHub for co-planning

### LINKED ISSUE #6103: [Rollout] Set up Fider for HHS/simpler-grants-gov
### Summary

Set up a Fider board for the HHS/simpler-grants-gov repository

### Acceptance criteria

- [ ] Users can load GitHub issues with the `Coplanning Proposal` issue label into SimplerGrants.Fider.io by manually triggering a GitHub action
- [ ] Data from Fider is automatically added to the GitHub issue body at least once a day

widal001:
## Summary

Adds scripts to sync GitHub issue data with Fider for co-planning.

Fixes #6103 

## Changes proposed

- Adds a `.github/linters/coplanning/` sub-directory with a CLI script to:
  - Populate a Fider board with GitHub issues that have the `Coplanning Proposal` label
  - Update those GitHub issues with the link to the corresponding Fider posts and vote count
- Adds two new GitHub actions that trigger these CLI commands:
  - `coplanning-sync-gh-to-fider.yml`
  - `coplanning-sync-fider-to-gh.yml`

## Context for reviewers

- The step to sync GitHub to Fider (i.e. populate the Fider board with GitHub issue data) is a manual trigger.
- The step to sync Fider to GitHub (i.e. updating GitHub issues with Fider vote counts) runs once per day.
- You can read more about this syncing behavior in [the README](https://github.com/HHS/simpler-grants-gov/blob/77b70611300e03c1591c3aa985f233d6089931cc/.github/linters/coplanning/README.md)

## Validation steps

- I've already run this once manually and populated the issues in [this Fider board](https://simplergrants.fider.io/)
- You can see the corresponding Fider section at the bottom of [this issue](https://github.com/HHS/simpler-grants-gov/issues/5538)
- You can also see the dry run in the CI checks:
  - [Co-planning: Sync GitHub to Fider / sync (pull_request)](https://github.com/HHS/simpler-grants-gov/actions/runs/17245471374/job/48933870565?pr=6109)
  - [Co-planning: Sync Fider to GitHub / sync (pull_request)](https://github.com/HHS/simpler-grants-gov/actions/runs/17245471321/job/48933870587?pr=6109)

### DIFF
--- a/.github/linters/coplanning/run.py
+++ b/.github/linters/coplanning/run.py
@@ -142,9 +142,6 @@ def sync_fider_to_github(args: CliArgs) -> None:
         dry_run=args.dry_run,
     )
 
-    if args.dry_run:
-        log("Dry run: Would update GitHub issues from Fider posts")
-
 
 # #######################################################
 # Main


.github/linters/coplanning/README.md
```diff
@@ -0,0 +1,59 @@
...
+    --update-existing
+```
+
+### Syncing Fider to GitHub
+
+By default, this workflow will run once a day at 12:00 AM UTC. You can manually trigger the "Co-planning: Sync Fider to GitHub" workflow to update the GitHub issue description with the Fider post description and the vote count.
+
+Behind the scenes this workflow makes the following CLI calls:
+
+```bash
```
btabaska, line 35: This block is duplicated from above

widal001: The `--sync-direction` value is actually different between these calls, but maybe I can highlight that as the key difference.

widal001: Added a sentence highlighting the difference here: https://github.com/HHS/simpler-grants-gov/pull/6109/commits/9d46268da36776e69a33085cfb983be77e9b55d6

.github/linters/coplanning/utils.py
```diff
@@ -0,0 +1,154 @@
...
+            return json.loads(response.read().decode())
+    except urllib.request.HTTPError as e:
+        err(f"HTTP request failed: {e.code} - {e.reason}")
+        sys.exit(1)
+    except json.JSONDecodeError:
+        err("Failed to parse JSON response")
+        sys.exit(1)
+    except Exception as e:
+        err(f"Request failed: {e}")
+        sys.exit(1)
```
chouinar: `err()` itself does a sys.exit(1) - so each of these can't get hit - should the error logging method do the exit, or just rely on the calling code?

Alternatively, make `err` take in a param like `exit: bool = False` and combine the idea

.github/linters/coplanning/run.py
```diff
@@ -0,0 +1,169 @@
...
+    update_existing: bool = False
+    dry_run: bool = False
+
+
+def parse_args() -> CliArgs:
+    """Parse command line arguments and return a CliArgs dataclass."""
+    parser = argparse.ArgumentParser(
+        description="Load GitHub issues into Fider or FeatureBase board"
+    )
+    parser.add_argument("--org", required=True, help="GitHub organization")
```
chouinar, line 48: I know we probably don't want a script like this to have a bunch of dependencies, but thoughts on using Click which we use in most other cases / tends to be a bit easier to work with? Not a strong opinion, if this already works, then that's fine.

.github/linters/coplanning/run.py
```diff
@@ -0,0 +1,169 @@
...
+    # Update GitHub issues based on Fider posts
+    log("Updating GitHub issues based on Fider posts")
+    github.update_github_issues(
+        issues=github_issues,
+        posts=fider_posts,
+        dry_run=args.dry_run,
+    )
+
+    if args.dry_run:
+        log("Dry run: Would update GitHub issues from Fider posts")
```
chouinar: Should this be in the above function? It looks like when that iterate over issues, it logs for each already.

.github/linters/coplanning/utils.py
widal001: Great catch! I like the combined approach, in case we want to be able to log without exiting in the future. I made that change in [refactor(linter): Fix duplicate sys.exit(1)](https://github.com/HHS/simpler-grants-gov/pull/6109/commits/21e63dc3e75a7dbab19f2f3873b175e5bb93fecc)

.github/linters/coplanning/run.py
widal001: Good catch as well! This was left over from when the check for dry run wasn't pulled up into that function. Removed it in [refactor(linter): Removes old dry-run log](https://github.com/HHS/simpler-grants-gov/pull/6109/commits/25297037b0d1c32e3d016ba6b3b995edcb14d5b7)

.github/linters/coplanning/run.py
widal001: Yeah it's a good point and one I considered because I really like the DX of click, but I worked really hard to not use any external dependencies to minimize the start up time and (more importantly) avoid locking us into perpetual dependency updates.

If we hit a point with this script where we really have to use some external dependency, then at that point I'd probably update this to use click as well. But for now I'd love to keep this pure python.

Issue #5916, opened 2025-08-12
Create ETL Task for Opportunity Review Status Data


joshtonava:
### Summary

A task is needed to transfer data from the foreign data mapper for V_OPPORTUNITY_SUMMARY into the new opportunity review status table.

### Tech Notes
- Create a new task in `api/src/task/opportunities/` to pull foreign data from the V_OPPORTUNITY_SUMMARY view into the opportunity review status table with status `Reviewable` or `Returned`
- This will be similar to existing opportunity loading tasks but customized for the view structure
- Truncate the entire table each time before loading 
- Add this task to the LoadTransform job. By default keep it off. After LoadOracleData

### Acceptance criteria

- Create new task that loads data into the opportunity review status table
- Add unit tests 
- Add appropriate logging and metrics

joshtonava:
Lets talk before picking this up. 

I think we may want to adjust our excluded_opportunity_review table

```
class ExcludedOpportunityReview(ApiSchemaTable, TimestampMixin):
    __tablename__ = "excluded_opportunity_review"

    opportunity_id: Mapped[int] = mapped_column(BigInteger, primary_key=True)
    omb_review_status_display: Mapped[str]
    omb_review_status_date: Mapped[date | None]
    last_update_date: Mapped[datetime | None]
```

Either rename opportunity_id to legacy_opportunity_id or switch to UUID

PR #6128, opened 2025-08-26
[Issue #5916] sync opportunity review status task  

babebe:
## Summary

Fixes for #5916  

## Changes proposed

A task that handles truncating the excluded_opportunity_review table and add all the opportunities in review status by filtering the foreign v_opportunity_summary table. 

## Context for reviewers
This task will setup the table that is queried downstream to filter out the opportunities listed here. Note: This task will run as the last step in the load_transform job, failure that occurs here shouldn't affect the other tasks. 

## Validation steps

Test added


api/src/task/opportunities/sync_opportunity_review_status_task.py
```diff
@@ -0,0 +1,63 @@
...
+        for row in result:
+            record = ExcludedOpportunityReview(
+                legacy_opportunity_id=row.opportunity_id,
+                omb_review_status_display=row.omb_review_status_display,
+                omb_review_status_date=row.omb_review_status_date,
+                last_update_date=row.fo_last_upd_dt,
+            )
+
+            self.db_session.add(record)
+            self.increment(self.Metrics.OPPORTUNITIES_IN_REVIEW)
```
jakobpederson, line 74: What happens if the record is already in the db? Is that a possibility?

babebe: Since each row in the result is unique and the table is emptied beforehand, there’s no chance of a duplicate or pre-existing record being present at insert time

api/tests/src/task/opportunities/test_sync_opportunity_review_status_task.py
```diff
@@ -0,0 +1,34 @@
+from src.db.models.opportunity_models import ExcludedOpportunityReview
+from src.task.opportunities.sync_opportunity_review_status_task import SyncOpportunityReviewStatus
+from tests.src.db.models.factories import ForeignVopportunitySummaryFactory
+
+
+def test_sync_opportunity_review_status(db_session, test_api_schema, enable_factory_create):
```
joshtonava, line 16: Can we add an additional test for no records? something like:

```
def test_sync_opportunity_review_status_empty_result(db_session, test_api_schema):
    """Test behavior when no opportunities match the review status criteria."""
    ForeignVopportunitySummaryFactory.create(omb_review_status_display="APPROVED")
    
    task = SyncOpportunityReviewStatus(db_session, test_api_schema)
    task.run()
    
    result = db_session.query(ExcludedOpportunityReview).all()
    assert len(result) == 0
    assert task.metrics[task.Metrics.OPPORTUNITIES_IN_REVIEW] == 0
```

api/src/task/opportunities/sync_opportunity_review_status_task.py
```diff
@@ -0,0 +1,63 @@
...
+@flask_db.with_db_session()
+@ecs_background_task(task_name="sync-opportunity-review-status")
+def sync_opportunity_review_status(db_session: db.Session) -> None:
+    SyncOpportunityReviewStatus(db_session).run()
+
+
+REVIEW_STATUS = ["REVIEWABLE", "RETURNED"]
+
+
+class SyncOpportunityReviewStatus(Task):
```
joshtonava, line 32: Question: Do we need to be logging the metrics at the end or is this automatically done in the parent class?

Also any other logging we may want?

api/src/task/opportunities/sync_opportunity_review_status_task.py
```diff
@@ -0,0 +1,63 @@
...
+    "sync-opportunity-review-status",
+    help="Pull opportunities in review status from foreign table v-opportunity_summary into our database",
+)
+@flask_db.with_db_session()
+@ecs_background_task(task_name="sync-opportunity-review-status")
+def sync_opportunity_review_status(db_session: db.Session) -> None:
+    SyncOpportunityReviewStatus(db_session).run()
+
+
+REVIEW_STATUS = ["REVIEWABLE", "RETURNED"]
```
joshtonava, line 29: nitpick: We could consider making this configurable

joshtonava (dismissed):
Overall looks good, I did have some nonblocking suggestions / thoughts.

api/src/task/opportunities/sync_opportunity_review_status_task.py
babebe: Yes that is done in the parent class.

api/tests/src/task/opportunities/test_sync_opportunity_review_status_task.py
babebe: adopted

api/src/task/opportunities/sync_opportunity_review_status_task.py
babebe: Are these values likely to change ? They are not env specific.

joshtonava: probably not, you are right

joshtonava (approved):

Issue #6097, opened 2025-08-25
[BUG] Form SF-424: Delinquent Federal Debt not functioning as expected


chris-kuryak:
### Summary

On form SF-424, the Delinquent Federal Debt question is currently set as a checkbox, but it needs to allow the user to select "Yes" or "No". And if "Yes", then the attachment is required.

This does not include showing the error in the field UI for the conditionally required attachment. That will be solved in ticket #6129 

### Acceptance criteria

- [ ] Functions as expected (above)
- [ ] Delinquent Federal Debt is a required field
- [ ] Functionality reviewed/approved by Product Manager

chris-kuryak:
Will wait until Aaron's work gets merged.
#6127

PR #6132, opened 2025-08-26
[Issue #6097]: Delinquent Federal Debt Radio button

myduong-navapbc:
## Summary

This is only a partial fix to: Delinquent Federal Debt [#6097](https://github.com/HHS/simpler-grants-gov/issues/6097)

## Changes proposed

I've added the Radio widget manually to the ui schema
it now displays, the YES / NO options
I tried to make it as type safe as possible
added some utils and helpers to work with the options that are returned/created 

## Context for reviewers

This does not contain the validation portion. That will come in a second PR-

## Validation steps

3. navigate to `search`
4. search for `pilot`
6. sign in
7. start new application
10. scroll down to Delinquent Federal Debt
11. when making no selection, and saving, you should not see any change to the radio and a validation error
12. when making a selection, for example, Yes, and saving, the value should remain Yes. same with No
13. Keep in mind validation is coming in a separate PR

### DIFF
--- a/frontend/src/components/applyForm/widgets/RadioWidget.tsx
+++ b/frontend/src/components/applyForm/widgets/RadioWidget.tsx
@@ -1,3 +1,5 @@
+"use client";
+
 import {
   enumOptionsIsSelected,
   enumOptionsValueForIndex,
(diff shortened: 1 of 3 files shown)


frontend/src/components/applyForm/widgets/RadioWidget.tsx
```diff
@@ -1,24 +1,70 @@
...
 
 import { TextTypes, UswdsWidgetProps } from "src/components/applyForm/types";
 import { DynamicFieldLabel } from "./DynamicFieldLabel";
 import { getLabelTypeFromOptions } from "./getLabelTypeFromOptions";
 
-/** The `RadioWidget` is a widget for rendering a radio group.
- *  It is typically used with a string property constrained with enum options.
+/**
+ * The portion of `options` we care about here, with safe typing.
+ * We shape our data since RJSF’s options are loosely typed.
```
doug-s-nava, line 21: 🆒

frontend/src/components/applyForm/widgets/RadioWidget.tsx
```diff
@@ -1,24 +1,70 @@
...
+    if (current === true) return "true";
+    if (current === false) return "false";
+    if (current === "true" || current === "false") return current;
+    return undefined;
+  }
+  return current;
+}
+
+/**
+ * parseFromInputValue
```
doug-s-nava, line 48: should this be `coerceFromString`?

frontend/src/components/applyForm/widgets/RadioWidget.tsx
```diff
@@ -1,24 +1,70 @@
...
  *
- * @param props - The `WidgetProps` for this component
+ * Convert an HTML input value (always a string) to either a boolean (for
+ * boolean radio groups) or the original enum value if we can find it.
  */
+function coerceFromString<S extends StrictRJSFSchema>(
+  raw: string,
+  enumOptions: ReadonlyArray<EnumOptionsType<S>>,
+): unknown {
+  const usesBoolStrings = enumOptions.some(
```
doug-s-nava, line 57: [food for thought] I wonder if this function could be split out and used elsewhere, like in `normalizeForCompare` as well. I can envision something like

```
  const booleanStringToBoolean = (value: unknown): unknown => {
    if (value === "true" || value === "false") {
      return value.toString();
    }
    return value;
  }
```

doug-s-nava (approved):
looks great

(would love to see some tests to see how this will work with non-boolean values, but that's not a blocker at all)

✅ ✅ ✅ ✅

frontend/src/components/applyForm/widgets/RadioWidget.tsx
myduong-navapbc: I will rename this in the upcoming PR, TY

Issue #6129, opened 2025-08-26
Conditional fields show validation in the UI (at the field level) when required


chris-kuryak:
### Summary

Currently, when a conditional fields is required, they are not indicated as such in the UI at the field leve. 

For example, on the SF-LLL form, if SubAwardee is selected, then Name and Address in Section 4 should become required visually, which it does not. (Congressional District remains optional, as expected.) While it does become required in the top error message, it is very difficult to navigate and find since it is not highlighted 'red' on the form.

Another example from Lucas on 8/28:

> Clicking on "'applicant_type_code' is a required property" at the top of the form's validation errors does NOT take me to the actual part of the form that has the error, unlike most of the other clicks I can complete on the error messages.

### Acceptance criteria

- [x] If conditionally required fields fail validation, they are marked visually (at the form level) as required
- [x] Merged to Staging and completed UAT with Product

PR #6095, opened 2025-08-25
[Issue #6045] Update flatFormDataToArray and upload widget

acouch:
## Summary

Fixes #6045 

## Changes proposed

Updates `flatFormDataToArray()` to return the correct errors for the upload widgets.

The updated form behavior:

image

## Context for reviewers

We have an issue with array support. Currently arrays are supported through the budget form and the upload widget. They both want data in a different form, so the `flatFormDataToArray()` function returns an array of objects for the budget form or a array of strings for the upload widget. 

We currently don't support generic arrays. I created #6092 to address this.

PR #6143, opened 2025-08-27
[Issue #6097]: Delinquent Federal Debt Radio button (part 2)

myduong-navapbc:
## Summary

Fixes to: Delinquent Federal Debt [#6097](https://github.com/HHS/simpler-grants-gov/issues/6097)

## Changes proposed

Dynamic conditional error mirroring (generic):

- Walks the JSON Schema to collect if/then rules and caches them (WeakMap) per dereferenced schema.
- When rendering each field, if it appears in an if.properties rule and the condition is satisfied, we mirror the server’s 'required' errors for then.required fields onto the trigger field.
- Uses the server’s exact error strings (e.g., '<field>' is a required property)—no fabricated messages.

Normalization helpers:
- Uses stringifyOrEmpty to compare current values to schema const values reliably (handles booleans and empties).
- (No change to user-facing copy; errors are whatever the server returns.)

## Context for reviewers

This works for any rule shaped like:

```
{
  "if": { "properties": { "<trigger>": { "const": <value> } }, "required": ["<trigger>"] },
  "then": { "required": ["<dependent-1>", "<dependent-2>", ...] }
}

```

- If an if rule doesn’t specify a const, we treat “trigger is set (non-empty)” as the condition.
- We do not alter the server errors; we simply display them also on the trigger to guide users.

## Validation steps

3. navigate to `search`
4. search for `pilot`
6. sign in
7. start new application
10. scroll down to Delinquent Federal Debt

- Leave it unselected, click Save, see the standard required error on the radio.
- Select Yes, do not attach/enter a debt explanation, click Save:
- You should still see server errors on the dependent field(s) (e.g., explanation/attachment).
- Additionally, the radio shows the server’s exact error text, e.g.:
- 'debt_explanation' is a required property
- Change selection to No and Save → the mirrored error disappears.


frontend/src/components/applyForm/utils.tsx
```diff
@@ -194,6 +199,51 @@ export const getByPointer = (target: object, path: string): unknown => {
   }
 };
 
+const conditionalRuleCache = new WeakMap<object, ConditionalRule[]>();
+
+function collectConditionalRules(
+  node: unknown,
+  acc: ConditionalRule[] = [],
+): ConditionalRule[] {
+  if (!node || typeof node !== "object") return acc;
```
doug-s-nava, line 208: does this support arrays? (since `typeof [] === 'object'`)

frontend/src/components/applyForm/utils.tsx
```diff
@@ -194,6 +199,51 @@ export const getByPointer = (target: object, path: string): unknown => {
   }
 };
 
+const conditionalRuleCache = new WeakMap<object, ConditionalRule[]>();
+
+function collectConditionalRules(
```
doug-s-nava, line 204: looks like this only handles condtional requires, which is totally cool, but maybe put that into the name: `collectConditionalRequiredRules`?

frontend/src/components/applyForm/utils.tsx
```diff
@@ -194,6 +199,51 @@ export const getByPointer = (target: object, path: string): unknown => {
...
+  node: unknown,
+  acc: ConditionalRule[] = [],
+): ConditionalRule[] {
+  if (!node || typeof node !== "object") return acc;
+  const obj = node as Record<string, unknown>;
+
+  if ("if" in obj && "then" in obj) {
```
doug-s-nava, line 211: do we want to break / return from this branch? as it is it looks like line 233 will run even if the object target is an `if/then` conditional object

frontend/src/components/applyForm/utils.tsx
```diff
@@ -231,12 +281,7 @@ export const buildField = ({
           ? (result as Record<string, unknown>)
           : {};
       })
-      .reduce<Record<string, unknown>>(
-        (acc, value) => ({ ...acc, ...value }),
-        {},
-      );
-    // multifield needs to retain field location for errors.
-    // this may not work with nested required fields
+      .reduce<Record<string, unknown>>((acc, v) => ({ ...acc, ...v }), {});
```
doug-s-nava, line 284: can you revert this variable naming?

frontend/src/components/applyForm/utils.tsx
```diff
@@ -250,14 +295,69 @@ export const buildField = ({
...
+
+    const types =
       typeof fieldSchema?.type === "string"
         ? fieldSchema.type
         : Array.isArray(fieldSchema?.type)
           ? (fieldSchema.type[0] ?? "")
           : "";
-    rawErrors = formatFieldWarnings(errors, name, fieldType, requiredField);
+
+    rawErrors = formatFieldWarnings(errors, name, types, requiredField);
```
doug-s-nava, line 311: if we're in the mood for refactoring things a bit, can we move this whole section of buildField into a "formatErrors" function to keep this function from ballooning more? if there's not an appetite for more refactors that's cool, and we can handle it later

frontend/src/components/applyForm/utils.tsx
```diff
@@ -250,14 +295,69 @@ export const buildField = ({
...
+
+        const missingRequiredWarnings = errors.filter(
+          (warning) =>
+            warning.type === "required" &&
+            rule.thenRequired.includes(warning.field),
+        );
+
+        if (missingRequiredWarnings.length) {
+          const serverMessages = Array.from(
+            new Set(
```
doug-s-nava, line 347: is `Array.from(new Set([]))` being done to remove duplicates, or is there another reason? If that's all I feel like `_.uniq()` is a more readable option (though it may just be doing this casting under the hood)

frontend/tests/components/applyForm/utils.test.tsx
```diff
@@ -234,6 +234,88 @@ describe("buildField", () => {
   });
 });
 
+describe("buildField – conditional error mirroring", () => {
```
doug-s-nava, line 237: seems like this would be more easily tested on a function by function level, if you export `collectConditionalRules` and move the trigger calculation into their own functions we could test those without worrying about the rest of buildField's functionality

frontend/src/components/applyForm/utils.tsx
```diff
@@ -570,27 +667,25 @@ export const isFieldRequired = (
 export const flatFormDataToArray = (
   field: string,
   data: Record<string, unknown>,
-) => {
-  if (field in data) return [data[field]];
```
doug-s-nava, line 574: removing this breaks the fix for multi file uploads done here https://github.com/HHS/simpler-grants-gov/pull/6095

frontend/src/components/applyForm/utils.tsx
```diff
@@ -570,27 +667,25 @@ export const isFieldRequired = (
...
-      if (fieldName === field && value) {
-        const match = key.match(/[0-9]+/);
-        const arrayNumber = match ? Number(match[0]) : -1;
-        if (!values[arrayNumber]) {
-          values[arrayNumber] = {};
-        }
-        values[arrayNumber][itemName] = value;
-      }
+): Array<Record<string, unknown> | undefined> => {
+  const out: Array<Record<string, unknown> | undefined> = [];
```
doug-s-nava, line 671: what is this refactor accomplishing? I'm still confused about what this does

myduong-navapbc:
We've decided to shift this PR into https://github.com/HHS/simpler-grants-gov/issues/6129

PR #6144, opened 2025-08-27
[Issue #6072] Add more logging and metrics to Start Application and Submit Application endpoints

### LINKED ISSUE #6072: Add more logging and metrics to Start Application and Submit Application endpoints
### Summary

To give more insight into our New Relic dashboard metrics (and potentially filter out internal users from our prod metrics), we want to log more information around our Start Application and Submit Application endpoints.

---
When we fetch an application or application related data, we want to make sure to call `add_extra_data_to_current_request_logs` with the following fields:

We want the following fields added:
* `organization_id` from the application (fine to be null)
* `competition_id` from the application
* `agency_code` from the application's competition's opportunity
* `opportunity_id` from the application's competition

Put where we call `add_extra_data_to_current_request_logs` somewhere fairly central so several endpoints get this, right after we fetch the application from the DB and do auth checks might be a good spot? For creating an application, it doesn't go through that flow so we'll need to add this in at least two places.

### Acceptance criteria

- [ ] Data listed above is logged and can be queried in New Relic dashboards on the Start Application Endpoint
- [ ] Data listed above is logged and can be queried in New Relic dashboards on the Submit Application Endpoint

mikehgrantsgov:
## Summary

Fixes #6072 

## Changes proposed

Add `_add_application_metadata_to_logs` function to add metadata to logging
Hook in that method call to relevant areas to capture more logging.
Add/update tests

## Context for reviewers

To give more insight into our New Relic dashboard metrics (and potentially filter out internal users from our prod metrics), we want to log more information around our Start Application and Submit Application endpoints.

## Validation steps

See updated unit tests

### DIFF
--- a/api/src/services/applications/create_application.py
+++ b/api/src/services/applications/create_application.py
@@ -188,7 +188,7 @@ def create_application(
     # Create a new application
     application = Application(
         application_id=uuid.uuid4(),
-        competition_id=competition_id,
+        competition=competition,
         application_name=application_name,
         application_status=ApplicationStatus.IN_PROGRESS,
         organization_id=organization_id,  # Set the organization ID if provided
@@ -219,9 +219,6 @@ def create_application(
             "organization_id": organization_id,
         },
     )
-
-    # Ensure the competition relationship is loaded for the returned application
-    application.competition = competition
 
     # Add application metadata to logs
     add_application_metadata_to_logs(application)


api/src/api/application_alpha/application_route.py
```diff
@@ -42,6 +43,29 @@
...
+            if application.competition and application.competition.opportunity
+            else None
+        ),
+        "agency_code": (
+            str(application.competition.opportunity.agency_code)
+            if application.competition
+            and application.competition.opportunity
+            and application.competition.opportunity.agency_code
+            else None
+        ),
```
chouinar: `application.competition` and `competition.opportunity` can't be null, we should be fine to simplify these to:
```suggestion
        "opportunity_id": application.competition.opportunity_id
        "agency_code": application.competition.opportunity.agency_code
        ),
```

api/src/api/application_alpha/application_route.py
```diff
@@ -42,6 +43,29 @@
 logger = logging.getLogger(__name__)
 
 
+def _add_application_metadata_to_logs(application: Application) -> None:
+    """Add application metadata to the current request logs for New Relic dashboards."""
+    from typing import Dict, Union
+
+    extra_data: Dict[str, Union[str, int, float, bool, UUID, None]] = {
```
chouinar:
```suggestion
def _add_application_metadata_to_logs(application: Application) -> None:
    """Add application metadata to the current request logs for New Relic dashboards."""

    extra_data: dict[str, str | UUID | None]] = {
```

api/src/api/application_alpha/application_route.py
```diff
@@ -67,6 +91,9 @@ def application_start(db_session: db.Session, json_data: dict) -> response.ApiRe
             db_session, competition_id, user, application_name, organization_id
         )
 
+        # Add application metadata to logs after creation
+        _add_application_metadata_to_logs(application)
```
chouinar: Could we put the calls to these in the service methods? I assume you didn't do that because our service tests fail because the `add_extra_data_to_current_request_logs` fails if it's not in a Flask request context?

We could adjust that function to instead do:
```py
if not flask.has_request_context():
    return
...
```
It doesn't need to be an assert, similar util functions don't do that.
---
In the service methods, we can just put these right next to the DB queries (eg. in the `get_application` endpoint which'll automatically make this work on most non-start/appform endpoints - we'll want to update those too)

api/src/services/applications/create_application.py
```diff
@@ -219,4 +220,10 @@ def create_application(
         },
     )
 
+    # Ensure the competition relationship is loaded for the returned application
+    application.competition = competition
```
chouinar: Could we just adjust when we make the application above, just change `competition_id=competition_id` to `competition=competition`, it'll set the ID too.

mikehgrantsgov: Just updated that, makes sense

Issue #6130, opened 2025-08-26
Get SF-424a working in staging


chris-kuryak:
### Summary

Currently, SF-424a is not working in staging. In order to perform UAT with Design and Product, we need it working in staging.

Slack chat showing a potential solution? - [LINK](https://betagrantsgov.slack.com/archives/C08NCEXRSPJ/p1755890841125569)

Current error seen after starting an application and opening the form:

<img width="1187" height="346" alt="Image" src="https://github.com/user-attachments/assets/3bfc74e2-33b5-4b1b-9954-3028130c4f6d" />

### Acceptance criteria

- [ ] Form SF-424a renders in staging environment

chris-kuryak:
I see it working in staging now. Thank you!

PR #6154, opened 2025-08-28
[ISSUE #6130]:  SF424A - form not rendering

myduong-navapbc:
## Summary

Work for #6130 - Get SF-424a working in staging

## Changes proposed

- Hardened `removePropertyPaths` to safely handle non-string inputs
- Updated i`sFieldRequired` to support `string | string[]` definitions

## Context for reviewers

SF424A was not rendering in Staging due to path handling (path.replace is not a function). 

## Validation steps

3. navigate to `search`
4. search for `pilot`
6. sign in
7. start new application
9. click Budget Information for Non-Construction Programs (SF-424A)


frontend/src/components/applyForm/utils.tsx
```diff
@@ -559,11 +560,16 @@ export const getRequiredProperties = (
 };
 
 export const isFieldRequired = (
-  definition: string,
+  definition: string | string[] | undefined,
   requiredFields: string[],
 ): boolean => {
-  const path = removePropertyPaths(definition);
-  return requiredFields.indexOf(path) > -1;
+  if (!definition) return false;
```
doug-s-nava, line 566: when I wrote this I figured it was simple enough that it didn't need testing 😬 . I'm still good with this as is, but would be even happier with some tests at this point

doug-s-nava (approved):
✅ ✅ ✅ ✅

frontend/src/components/applyForm/utils.tsx
myduong-navapbc: Will do, thanks for the feedback doug. I will add some tests around this in the main form PR

PR #6162, opened 2025-08-28
[Issue #5869] Cleanup old sam.gov extract files

### LINKED ISSUE #5869: Cleanup old sam.gov extract files
### Summary

We fetch sam.gov extract files daily. Files are either monthly extracts (full snapshots of the entity data) or daily (all updates since the last extract file).

These files are large, and contain some sensitive information. We want to clean these files up after a period of time.

What we want to do is:
* Add a new processing status called `deleted`
* Fetch all `sam_extract_file` records older than 45 days that aren't already deleted
* For each of these, delete the file from s3 (use `file_util.delete_file()`)
* Mark the record in our DB as deleted

We'll keep the records in the DB even when deleted, just as a reference, but won't keep the file on s3.

---
This should be a new Task class that runs as a new 4th step on the sam extract task.

### Acceptance criteria

- [ ] Task class created
- [ ] Tests written
- [ ] Class configured to run as 4th step of job

mikehgrantsgov:
## Summary

Fixes #5869

## Changes proposed

Add a new processing status called `deleted` to `SamGovProcessingStatus` and table.
Create `CleanupOldSamExtractsTask` which runs on the 4th step of the sam extract task which delete files > 45 days old.
Add tests.

## Context for reviewers

We fetch sam.gov extract files daily. Files are either monthly extracts (full snapshots of the entity data) or daily (all updates since the last extract file).

These files are large, and contain some sensitive information. We want to clean these files up after a period of time.

## Validation steps

See updated unit tests.


api/src/db/migrations/versions/2025_08_23_add_deleted_status_to_sam_gov_processing_status.py
```diff
@@ -0,0 +1,40 @@
...
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    # Remove the DELETED status from the lookup table
+    op.execute(
+        """
+        DELETE FROM api.lk_sam_gov_processing_status
+        WHERE sam_gov_processing_status_id = 4 AND description = 'deleted'
+        """
+    )
```
chouinar: We do not need to use a migration to add new lookup values. The migration process already sync lookup values. By just adding the `LookupStr` to the list of values, running `make db-migrate` will automatically do basically this.

`make db-migrate` calls this function:
* https://github.com/HHS/simpler-grants-gov/blob/main/api/src/db/migrations/run.py#L25
* which in turn calls https://github.com/HHS/simpler-grants-gov/blob/main/api/src/db/models/lookup/sync_lookup_values.py#L12

Which is hooked into the whole setup for how lookup values work.


```suggestion

```

api/src/task/sam_extracts/cleanup_old_sam_extracts.py
```diff
@@ -0,0 +1,112 @@
...
+                    "Failed to cleanup old SAM.gov extract file",
+                    extra={
+                        "sam_extract_file_id": sam_extract_file.sam_extract_file_id,
+                        "s3_path": sam_extract_file.s3_path,
+                        "extract_date": sam_extract_file.extract_date,
+                        "extract_type": sam_extract_file.extract_type,
+                        "error": str(e),
+                    },
+                    exc_info=True,
+                )
```
chouinar: I'd say if we hit an error, we just let the process fail - with what we're doing here being just a simple file delete, I think the only possible errors are permissions and AWS outages - so if we hit one, we probably just aren't getting anything done.

```suggestion
            except Exception:
                logger.exception(
                    "Failed to cleanup old SAM.gov extract file",
                    extra={
                        "sam_extract_file_id": sam_extract_file.sam_extract_file_id,
                        "s3_path": sam_extract_file.s3_path,
                        "extract_date": sam_extract_file.extract_date,
                        "extract_type": sam_extract_file.extract_type,
                    },
                )
                raise
```

api/src/task/sam_extracts/cleanup_old_sam_extracts.py
```diff
@@ -0,0 +1,112 @@
...
+    def _cleanup_file(self, sam_extract_file: SamExtractFile) -> None:
+        """Cleanup a single SAM.gov extract file"""
+        # Delete the file from S3
+        file_util.delete_file(sam_extract_file.s3_path)
+
+        # Mark the record as deleted in the database
+        sam_extract_file.processing_status = SamGovProcessingStatus.DELETED
+
+        # Commit the changes
+        self.db_session.commit()
```
chouinar: Can we just throw one `with db_session.begin()` in the run_task function?
```suggestion

```

api/src/task/sam_extracts/cleanup_old_sam_extracts.py
```diff
@@ -0,0 +1,109 @@
...
+                    raise
+
+            logger.info(
+                "Completed cleanup of old SAM.gov extract files",
+                extra={
+                    "total_files_processed": len(old_files),
+                    "files_deleted": deleted_count,
+                    "errors_encountered": error_count,
+                },
+            )
```
chouinar: The incrementing count of errors wouldn't ever get used since we raise from where we add to it, can we trim that out?

```suggestion
                except Exception:
                    logger.exception(
                        "Failed to cleanup old SAM.gov extract file",
                        extra={
                            "sam_extract_file_id": sam_extract_file.sam_extract_file_id,
                            "s3_path": sam_extract_file.s3_path,
                            "extract_date": sam_extract_file.extract_date,
                            "extract_type": sam_extract_file.extract_type,
                        },
                    )
                    raise

            logger.info(
                "Completed cleanup of old SAM.gov extract files",
                extra={
                    "total_files_processed": len(old_files),
                },
            )
```

mikehgrantsgov: Ah that makes sense. Updated this.

PR #6204, opened 2025-09-02
[Unticketed] Explicitly mark updated rows as not-deleted in load-oracle logic

chouinar:
## Changes proposed
When doing an update from the Oracle DB, set is_deleted and deleted_at back to default (undeleted) values

## Context for reviewers
Had an opportunity synopsis not get updated in prod because the synopsis was deleted and later recreated. The primary key is just the opportunity ID, so our logic saw them as the same row. It did properly update the row, but the `is_deleted` flag was not changed so it left it deleted.


mdragon (approved):
Approved

Issue #4870, opened 2025-04-30
saved search queries page values should be translated from backend values to human readable format, matching names used in filters UI


btabaska:

btabaska:
@chrisw-nava could you give more context to this task.

chrisw-nava:
And here's the expectations per the designs

![Image](https://github.com/user-attachments/assets/e04a797d-17a4-4c87-bc76-05956e90f0e1)

doug-s-nava:
@vergenzt is taking this one

PR #6252, opened 2025-09-02
[Issue #4870] human readable saved search

doug-s-nava:
## Summary

Work for #4870 

## Changes proposed

Updates saved search list to display "human readable" display values for each selected parameter for each search

## Context for reviewers

This requires moving sort options out of a translatable context, as we've done previously for all other filter options. If we ever want to translate these we'll have to figure out how to do that and keep things useful and iterable the way we need them to be for this sort of functionality.

## Validation steps

3. log in
4. add some of every possible search param including a search term and a non-default sort
5. save the search
6. visit the /saved-searches page
7. _VERIFY_: search params all appear with human readable values

### Screenshots

Screenshot 2025-09-02 at 5 26 38 PM


frontend/src/constants/searchFilterOptions.ts
```diff
@@ -213,3 +213,57 @@ export const allFilterOptions = {
   category: categoryOptions,
   fundingInstrument: fundingOptions,
 };
+
```
myduong-navapbc, line 216: [not a blocker]
One question, since we pulled these out of i18n, are we going to sunset the i18n file?

doug-s-nava: good question - the answer I've gotten is no. Even though we don't have it in the roadmap to add spanish translations, the team wants to leave all the infrastructure behind the translations in place for now

PR #6258, opened 2025-09-03
[Unticketed] Make Javascript configurable with DocRaptor

chouinar:
## Changes proposed
Make javascript configurable for DocRaptor

## Context for reviewers
We want to try testing enabling javascript for docraptor - made it an environment variable defaulted to false (what it already would be). Will manually override to test.


joshtonava (approved):
Not super familiar with DocRaptor but this looks to accomplish passing down the configuration.

mdragon (approved):
👍🏻

Issue #6111, opened 2025-08-26
Create tables linking users to resources and roles


joshtonava:
### Summary

With the basic roles and privileges modeled in the db #6110 , we need to be able to related these roles to entities in the database.

### Tech Notes

See Tech Spec for reference
https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/1977221432/Tech+Spec+User+Roles+on+the+Backend#Create-tables-linking-users-to-resources-and-roles

- `internal_user_role`
- `application_user_role`
- `organization_user_role`
- `agency_user`
- `agency_user_role`

### Acceptance criteria

- Create tables that will be used to model the relationship between roles and entities in the database
- Add data migration

eric-serrato95:
Dependant on #6110

PR #6268, opened 2025-09-03
[Issue #6111] Add tables for user resources and roles

babebe:
## Summary

Fixes  for #6111  

## Changes proposed
Create tables linking users to resources and roles
Create Migration file
Create Factories


api/src/db/models/user_models.py
```diff
@@ -322,3 +323,60 @@ class LinkRolePrivilege(ApiSchemaTable, TimestampMixin):
...
+    user: Mapped[User] = relationship(User)
+
+
+class AgencyUserRole(ApiSchemaTable, TimestampMixin):
+    __tablename__ = "agency_user_role"
+
+    agency_user_id: Mapped[uuid.UUID] = mapped_column(
+        ForeignKey(AgencyUser.agency_user_id), primary_key=True
+    )
+    agency_user: Mapped[User] = relationship(AgencyUser)
```
chouinar:
```suggestion
    agency_user: Mapped[AgencyUser] = relationship(AgencyUser)
```

api/tests/src/db/models/factories.py
```diff
@@ -2723,3 +2746,36 @@ class Meta:
...
+
+
+class ApplicationUserRoleFactory(BaseFactory):
+    class Meta:
+        model = user_models.ApplicationUserRole
+
+    application_user = factory.SubFactory(ApplicationUserFactory)
+    application_user_id = factory.LazyAttribute(lambda o: o.application_user.application_user_id)
+
+    role = factory.SubFactory(RoleFactory)
```
chouinar: For each of these factories, we should make it so the role they get makes sense for the type of user they are. It'll likely evolve over time, but could we add a few traits to the RoleFactory, then here (and the below two factories), we can do:

```py
role = factory.SubFactory(RoleFactory, is_application_role=True)
```

I imagine the traits would look something like this in the role table:
```py
class Params:
    is_application_role = factory.Trait(privileges = [Privilege.VIEW_APPLICATION,Privilege.MODIFY_APPLICATION,Privilege.SUBMIT_APPLICATION]
    role_types = [RoleType.APPLICATION])

    is_org_role = factory.Trait(privileges = [Privilege.VIEW_APPLICATION,Privilege.MODIFY_APPLICATION,Privilege.SUBMIT_APPLICATION,Privilege.MAANGE_ORG_MEMBERS] # maybe more, can adjust later
    role_types = [RoleType.ORGANIZATION])

    # We don't have any agency/internal privileges defined yet, so leave it blank for now?
    is_agency_role = factory.Trait(privileges = [], role_types=[RoleType.AGENCY])
    is_internal_role = factory.Trait(privileges = [], role_types=[RoleType.INTERNAL])
```

api/tests/src/db/models/factories.py
```diff
@@ -2723,3 +2746,36 @@ class Meta:
     omb_review_status_display = factory.Faker("random_element", elements=["RETURNED", "REVIEWABLE"])
     omb_review_status_date = factory.Faker("date_time_between", start_date="-5y", end_date="-3y")
     last_update_date = factory.Faker("date_time_between", start_date="-5y", end_date="-3y")
+
+
+class ApplicationUserRoleFactory(BaseFactory):
```
chouinar: Could we move these factories to be next to the other app/org/etc factories?

api/src/db/models/user_models.py
```diff
@@ -322,3 +323,60 @@ class LinkRolePrivilege(ApiSchemaTable, TimestampMixin):
...
+    role: Mapped[Role] = relationship(Role)
+
+
+class ApplicationUserRole(ApiSchemaTable, TimestampMixin):
+    __tablename__ = "application_user_role"
+
+    application_user_id: Mapped[uuid.UUID] = mapped_column(
+        ForeignKey(ApplicationUser.application_user_id), primary_key=True
+    )
+    application_user: Mapped[User] = relationship(ApplicationUser)
```
joshtonava:
```suggestion
    application_user: Mapped[ApplicationUser] = relationship(ApplicationUser)
```

api/src/db/models/user_models.py
```diff
@@ -322,3 +323,60 @@ class LinkRolePrivilege(ApiSchemaTable, TimestampMixin):
...
+    role: Mapped[Role] = relationship(Role)
+
+
+class OrganizationUserRole(ApiSchemaTable, TimestampMixin):
+    __tablename__ = "organization_user_role"
+
+    organization_user_id: Mapped[uuid.UUID] = mapped_column(
+        ForeignKey(OrganizationUser.organization_user_id), primary_key=True
+    )
+    organization_user: Mapped[User] = relationship(OrganizationUser)
```
joshtonava:
```suggestion
    organization_user: Mapped[OrganizationUser] = relationship(OrganizationUser)
```

api/src/db/migrations/versions/2025_09_03_add_user_resource_role_tables.py
```diff
@@ -0,0 +1,175 @@
...
+        ),
+        sa.ForeignKeyConstraint(
+            ["application_user_id"],
+            ["api.application_user.application_user_id"],
+            name=op.f("application_user_role_application_user_id_application_user_fkey"),
+        ),
+        sa.ForeignKeyConstraint(
+            ["role_id"], ["api.role.role_id"], name=op.f("application_user_role_role_id_role_fkey")
+        ),
+        sa.PrimaryKeyConstraint(
```
joshtonava: This goes for all of the join tables, Do we need to add a unique constraint with both of the role_id and {entity}_user_id?

The migrations make me think we could apply multiple of the same role to an entity

cc: @chouinar

chouinar: A primary key of `role_id` + `<entity>_user_id` is also a unique constraint. We wouldn't be able to add multiple of the same role to a user with this

Issue #5441, opened 2025-06-25
Implement temporary warning of unsaved changes


chris-kuryak:
### Summary

Per Aaron's suggestion, there may be a quick way to implement a warning for users who have unsaved changes on a form:
https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event

### Acceptance criteria

- [ ] Warning for unsaved changes is triggered for users

acouch:
This was fixed with #6292

chris-kuryak:
Checked and working as expected in:
- Chrome on Mac
- Safari on Mac

PR #6292, opened 2025-09-04
[Issue #5441] Update form to prevent lost changes

acouch:
## Summary

Fixes / Work for #5441

## Changes proposed

Adds https://github.com/LayerXcom/next-navigation-guard and wraps application forms in `NavigationGuardProvider`.

## Context for reviewers

I have a separate branch (https://github.com/HHS/simpler-grants-gov/tree/acouch/issue-5441-unsaved-form) that handled refresh and tab closing without this library, however the back button was more difficult. Adding the `popState` listener allowed intercepting the back button, but handling the "cancel" button in the dialog would mean intercepting the router which would have taken more custom logic. This library handles a lot of edge cases.

This only checks if a change is made to the form and the save button has not been pressed. Someone could add to a field, then remove what they added, returning to the original state. Deep comparing the formData for changes seems like outside of the scope of the ticket.

## Validation steps

Edit a form without saving, try and navigate away.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -33,6 +33,7 @@
         "newrelic": "^12.7.0",
         "next": "^15.4.7",
         "next-intl": "^3.2.1",
+        "next-navigation-guard": "^0.2.0",
         "pino": "^9.6.0",
         "react": "^19.0.0",
         "react-dom": "^19.0.0",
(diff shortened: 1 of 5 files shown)


frontend/tests/components/applyForm/ApplyForm.test.tsx
```diff
@@ -46,6 +46,15 @@ jest.mock("src/services/auth/session", () => ({
...
 }));
 
+jest.mock("src/services/auth/session", () => ({
+  getSession: (): unknown => getSessionMock(),
+}));
+
+jest.mock("next-navigation-guard", () => ({
+  useNavigationGuard: () => jest.fn(),
+}));
+// import { useNavigationGuard } from "next-navigation-guard";
```
doug-s-nava: can delete this

doug-s-nava (dismissed):
simple solution, love it

✅ ✅ ✅ ✅

doug-s-nava (approved):

Issue #6298, opened 2025-09-05
Update github workflows for Training


prasnava:
### Summary

Update github workflow nonprod deploys to include the new training environment in the deployment matrix.

### Acceptance criteria

- [ ]  Github deploy matrix is updated to include the training environment.

PR #6299, opened 2025-09-05
[Issue #6298]: Add Training Env to release deploy

prasnava:
## Summary

Update the github release action to deploy to training environment.

## Validation steps

Verify github workflow deploy step includes the training env.

### DIFF
--- a/api/src/form_schema/forms/sf424.py
+++ b/api/src/form_schema/forms/sf424.py
@@ -168,7 +168,7 @@ FORM_JSON_SCHEMA = {
             # A user will never fill this in, it's just on the form for agencies to use
             "type": "string",
             "title": "Date Received By State",
-            "description": "Enter the date received by the state, if applicable.",
+            "description": "Leave blank, to be filled out by state.",
             "format": "date",
             "readOnly": True,
         },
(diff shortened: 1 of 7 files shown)


.github/workflows/cd-analytics.yml
```diff
@@ -48,7 +48,7 @@ jobs:
       max-parallel: 1
       fail-fast: false
       matrix:
-        envs: ${{ fromJSON(inputs.environment != null && format('["{0}"]', inputs.environment) || github.event_name == 'release' && '["prod"]' || github.ref_name == 'main' && '["dev", "staging"]' || '["dev"]')  }}
+        envs: ${{ fromJSON(inputs.environment != null && format('["{0}"]', inputs.environment) || github.event_name == 'release' && '["training", "prod"]' || github.ref_name == 'main' && '["dev", "staging"]' || '["dev"]')  }}
```
mdragon: I think the order here matters and we'd want Prod to be first so that we're not waiting on Training if we're fixing downtime, etc.

mdragon (changes requested):

.github/workflows/cd-analytics.yml
prasnava: Will reverse the order, so that prod is first.

Issue #3793, opened 2025-02-06
Allow spoofed logins for e2e tests


doug-s-nava:
### Summary

As we will soon begin introducing complex behavior in the application that relies on a logged in state, we would like to be able to test the behavior in the e2e suite without first having to automate logging in.

This can likely be accomplished through a combination of manually introduced client side cookies (see https://playwright.dev/docs/api/class-browsercontext#browser-context-add-cookies) and API side seeding of test users.

Note that we are currently unable to test login in the E2E suite running in Github actions (see https://betagrantsgov.slack.com/archives/C05TSL64VUH/p1738852240921799)

### Implementation Notes

Playwright allows you to set cookies on the browser before running tests. Using this functionality we should be able to simulate a login, but we will need to encrypt the cookie using the same keys used by the API and Next if we want the token to successfully do anything. This means providing these keys to the e2e process somehow.

#### How does app know if you're logged in?

- it makes a call to the session endpoint
- the Next API decryipts the "session" cookie sent up in the request and returns a token from the decrypted cookie
- as long as this returns a token, the frontend will log you in

### Acceptance criteria

- [ ] e2e tests of auth-dependent behavior are able to run successfully in CI without first automating a manual login process

eric-serrato95:
Push it off until we can have a conversation with fellow NAVA QA folks about how to implement e2e tests

eric-serrato95:
Dependent on FE work related to enhanced deliverables

PR #6318, opened 2025-09-08
[Issue #3793] spoof e2e login

doug-s-nava:
## Summary

Work for #3793 

## Changes proposed

Allows e2e tests to run on functionality or pages that require a user to be logged in, without scripting the e2e test through the login process

## Context for reviewers

This change is being made for two reasons:

* there is currently an undiagnosed issue with the authentication flow when running the application in a CI environment that is preventing us from using any e2e tests that actually log a user in using the usual UI. This change works around that issue
* as we build more functionality in the future that is only available to logged in users, we will often want to jump straight into testing the most relevant portions of the system without having to log the user in through the UI. This will allow for more targeted and performant tests of things like applications and forms

Tests of functionality that require a user to be logged in can now be run in CI and locally, though locally you will need to run through the steps listed in the readme which you can find in the diff.

## Validation steps

1. _VERIFY_: e2e tests pass in CI on this PR
2. use steps from documentation in this branch to set up for running spoofed login on e2e tests locally
3. run the `saved-opportunities` spec file against a local server
4. _VERIFY_: tests pass

### DIFF
--- a/documentation/frontend/testing.md
+++ b/documentation/frontend/testing.md
@@ -1,0 +1,17 @@
+# Testing
+
+## E2E testing
+
+### Spoofing logins
+
+There are situations where we want to be able to test a "logged in" experience without having to script the test through the full login flow. In order to support this we have built a system to spoof the user login by placing a session cookie into the browser context.
+
+This system is based on a `createSpoofedSessionCookie` which will create a client side cookie on the browser context that will effectively log in a fake user with the API.
+
+Using this function, tests should work automatically in CI, but they will require a bit of manual setup to work locally.
+
+#### Local setup
+
+- run `make db-seed-local` in the /api directory. This will create the necessary DB records for the spoofed user and spit out an API auth token in a file at /api/e2e_token.tmp.
+- copy the token variable declaration from the e2e_token.tmp file into your frontend .env.local file
+- that's it! Running e2e tests using the functionality mentioned above should now work locally


api/tests/lib/seed_e2e.py
```diff
@@ -0,0 +1,35 @@
...
+# * create an authorization token entry tied to that user, seeding the database accordingly
+# * write the resulting token into the overrides.env file (where it can be retrieved by CI processes for frontend / e2e use)
+###############################
+
+
+def _append_token_to_override(token: str) -> None:
+    path_to_override = os.path.join(os.path.dirname(__file__), "..", "..", "override.env")
+    if Path(path_to_override).exists():
+        token_declaration = 'E2E_USER_AUTH_TOKEN="' + token + '"'
+        append_to_file(path_to_override, token_declaration)
```
chouinar: If this appends to the file every time it runs, that's going to gradually add a lot to the file if you're doing a lot of testing. Would this add it on a new line or just write to the end of the file?

Since the API doesn't need this env var, what if we threw it in a different file so you can do whatever you want with it without affecting the API's configuration setup? Can just add a file to the .gitignore and then do whatever you want.

api/src/util/file_util.py
```diff
@@ -202,6 +202,14 @@ def read_file(path: str | Path, mode: str = "r", encoding: str | None = None) ->
         return input_file.read()
 
 
+def append_to_file(
+    path: str | Path, content: str, mode: str = "a", encoding: str | None = None
+) -> str:
+    """Simple function for adding a line to end of a file"""
+    with open_stream(path, mode, encoding) as file_to_write_to:
+        return file_to_write_to.write(content)
```
chouinar: If this is called append, we shouldn't let the mode change - we also would want to add tests.

api/tests/lib/seed_e2e.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def _build_users_and_tokens(db_session: db.Session) -> None:
+    USER_E2E_BASE = factories.UserFactory.build(
+        user_id=uuid.UUID("7edb5704-9d3b-4099-9e10-fbb9f2729aff")
+    )
```
chouinar: Can you define this in the same pattern as the other user setup file? We'd additionally want to have at least a record in the link external user table so you could verify this works by logging into it.

doug-s-nava: I tried, but run into a unique constraint error when following the pattern (see commented out code). Not sure what I'm missing

Issue #6135, opened 2025-08-27
SF-424a: Ensure data is being saved to the right place


chris-kuryak:
### Summary

Since we have created the budget form design, we now need to ensure that all of the data input fields are being saved to the right location in the data schema.

### Acceptance criteria

- [ ] All data fields in the SF-424a Budget Form are connected to the correct data locations in the JSON schema

Issue #6131, opened 2025-08-26
SF-424a: Get validation working as expected


chris-kuryak:
### Summary

For the SF-424a form, to get validation working as expected after running validation:

- Show all errors at the top (like all other forms)
- Do not pass validation error text to table
- Hide the thick red bar
- Highlight required fields (input boxes) in red that fail validation

### Acceptance criteria

- [ ] Validation is working as expected
- [ ] Changes merged to staging
- [ ] UAT completed by product and design

chris-kuryak:
UAT Passed

Issue #5163, opened 2025-05-27
Combobox not retaining proper state in apply form


acouch:
### Summary

The comboxbox adopted from USWDS does not save entries properly in the apply form. It has been temporarily disabled.

The javascript used to create the combo box creates an input field that the user uses and retains the select list. The select list might need to be updated properly in an onchange function. 

### Reproducibility

None

### Browser

None

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

myduong-navapbc:
this one is in progress.
as part of the change to the ui, we use a confirmation checkbox vs a radio select.
https://github.com/HHS/simpler-grants-gov/pull/6324

ErinPattisonNava:
@hayeshorus I know you worked on this recently, is this still an issue?  If not, can you close it?  cc @chris-kuryak @myduong-navapbc

hayeshorus:
Yes, this is not an issue.  It was done in #8239.

PR #6324, opened 2025-09-08
[ISSUE #6131]: SF 424-A validation and data saved properly

myduong-navapbc:
### Summary

Work for #6131 - SF-424a: Get validation working as expected
Work for #6135 - SF-424a: Ensure data is being saved to the right place
Work for #5163 - Combobox not retaining proper state in apply form

### Changes proposed

Budget sections A–F work end-to-end

Composite data access:
Passed formContext from ApplyForm to widgets ({ rootSchema, rootFormData }) so composite budget sections can read across schema fragments without special-casing everywhere.

Fixed FormFields logic so required is computed from a proper JSON Pointer even when a node lacks definition. We now synthesize a pointer from node.name (replacing -- with /properties/) which matches getRequiredProperties(schema) output.

### Context for reviewers

- Composite budget sections (A–F) need access to broader form context. 
- Pass formContext from ApplyForm ({ rootSchema, rootFormData }), so sections can read complete values 
- Updates to Radio Component
- Updates to Checkbox components

### Validation steps

3. navigate to `search`
4. search for `pilot`
6. sign in
7. start new application
9. click Budget Information for Non-Construction Programs (SF-424A)

Screenshot 2025-09-12 at 10 39 30 AM
Screenshot 2025-09-12 at 10 39 43 AM
Screenshot 2025-09-12 at 10 39 53 AM
Screenshot 2025-09-12 at 10 40 05 AM
Screenshot 2025-09-12 at 10 40 18 AM
Screenshot 2025-09-12 at 10 40 27 AM
Screenshot 2025-09-12 at 10 40 36 AM
Screenshot 2025-09-12 at 10 40 51 AM
Screenshot 2025-09-12 at 10 40 57 AM


frontend/src/components/applyForm/actions.ts
myduong-navapbc, line 1: I changed it to false and only a successful round-trip to the API should mark the UI as “saved”.

frontend/src/components/applyForm/ApplyForm.tsx
myduong-navapbc, line 1: we needed to forced the re-render after saving so the ui reflects the changes

frontend/src/components/applyForm/ApplyForm.tsx
```diff
@@ -35,9 +35,17 @@ const ApplyForm = ({
   validationWarnings: FormValidationWarning[] | null;
   attachments: Attachment[];
 }) => {
+  type RichRenderer = (chunks: ReactNode) => ReactNode;
```
doug-s-nava, line 47: can we define types in a types file, or at least outside of the function call?

frontend/src/components/applyForm/ApplyForm.tsx
```diff
@@ -62,13 +70,17 @@ const ApplyForm = ({
...
       // eslint-disable-next-line no-alert
-      window.confirm(t("unsavedChangesWarning")),
+      window.confirm(translate("unsavedChangesWarning")),
   });
 
   const { formData, error, saved } = formState;
 
   const formObject = !isEmpty(formData) ? formData : savedFormData;
   const navFields = useMemo(() => getFieldsForNav(uiSchema), [uiSchema]);
+  const formContextValue = useMemo(
```
doug-s-nava, line 89: can we avoid the word `context` outside of situations where we're using a React context? I think that might get confusing

frontend/src/components/applyForm/ApplyForm.tsx
```diff
@@ -35,9 +35,17 @@ const ApplyForm = ({
...
+  type Translator = ((
+    key: string,
+    values?: Record<string, unknown>,
+  ) => string) & {
+    rich: (key: string, values: Record<string, RichRenderer>) => ReactNode;
+  };
   const { pending } = useFormStatus();
   const t = useTranslations("Application.applyForm");
-  const required = t.rich("required", {
+  const translate = t as unknown as Translator;
```
doug-s-nava, line 56: is the change to the translation function to satisfy typing? did something change there? (honestly I'm good with pulling the translation stuff out completely for now if it's increasing the complexity)

frontend/src/components/applyForm/FormFields.tsx
```diff
@@ -88,16 +97,15 @@ export const FormFields = ({
...
           const childAcc: JSX.Element[] = [];
           const keys: number[] = [];
           const row = uiSchema.map((node) => {
             if ("children" in node) {
               acc.forEach((item, key) => {
-                if (item) {
-                  if (item.key === `${node.name}-wrapper`) {
-                    keys.push(key);
-                  }
+                if (item && item.key === `${node.name}-wrapper`) {
```
doug-s-nava, line 103: ❤️

frontend/src/components/applyForm/actions.ts
doug-s-nava: what's the effect of this change within the flow of the form?

frontend/src/components/applyForm/formDataToJson.ts
```diff
@@ -18,30 +18,55 @@ type FormDataToJsonOptions = {
...
-  if (value === "true") return true;
-  if (
-    (type === "integer" || type === "number") &&
-    value !== "" &&
-    !isNaN(Number(value))
-  )
-    return Number(value);
-  if (type === "string" && !isNaN(Number(value))) {
-    return value || undefined;
+const parseValue = (rawValue: unknown, type: string) => {
```
doug-s-nava, line 31: curious if there was a case the previous implementation wasn't handling correctly, or if this is a refactor with the same functionality

frontend/src/components/applyForm/widgets/CheckboxWidget.tsx
```diff
@@ -29,56 +23,73 @@ function CheckboxWidget<
...
+    [onFocus, id],
   );
 
   return (
-    <FormGroup error={error} key={`form-group__checkbox--${id}`}>
+    <FormGroup error={hasError} key={`form-group__checkbox--${id}`}>
       <Checkbox
         id={id}
-        label={label}
-        labelDescription={description}
```
doug-s-nava, line 74: want to confirm that removing the description is intentional here

frontend/src/components/applyForm/widgets/budget/Budget424aSectionA.tsx
```diff
@@ -1,49 +1,146 @@
+"use client";
+
 /* eslint-disable @typescript-eslint/no-unsafe-assignment */
-// file adapted from https://github.com/rjsf-team/react-jsonschema-form/blob/main/packages/core/src/components/templates/BaseInputTemplate.tsx
```
acouch, line 2: Not sure about losing this, but maybe be best to move it to a README.md

frontend/src/components/applyForm/ApplyForm.tsx
```diff
@@ -110,14 +122,16 @@ const ApplyForm = ({
           />
           <AttachmentsProvider value={attachments ?? []}>
             <FormFields
+              key={saved ? "after-save" : "before-save"}
               errors={saved ? validationWarnings : null}
               formData={formObject}
               schema={formSchema}
               uiSchema={uiSchema}
+              formContext={formContextValue}
```
doug-s-nava, line 140: it looks like the root value as being used to create default values for some of the sections, and the root schema is being used only in section F in order to fetch some necessary schema data that is not otherwise passed through or properly determined from the tree and field building process.

I'm assuming / hoping that the schema passing is something that we'll find a better solution for down the line since passing in the entire schema into every widget seems like something we don't want to do, but if it's getting this working for now I'm good to ticket that for a later day.

Why do we need the default value behavior?

frontend/src/components/applyForm/ApplyForm.tsx
doug-s-nava: where is this rerender logic?

frontend/src/components/applyForm/widgets/budget/Budget424aSectionA.tsx
myduong-navapbc: okie dokie, moved it to the readme

myduong-navapbc:
> two UI things to note off of the bat:
> 
> * when loading the page, click on a field, enter a key -> no data goes into the field, and the field loses focus
> * enter data in a table field (confirmation seems to work fine), save -> data is removed from form

fixed:
i moved the Cell/DataCell component out and that fixed the strange loss of focus issue and data not saving

frontend/src/components/applyForm/widgets/budget/budgetValueGuards.ts
```diff
@@ -0,0 +1,17 @@
+import { get } from "lodash";
+
+export function isRecord(value: unknown): value is Record<string, unknown> {
```
doug-s-nava, line 3: [nit] could replace with https://lodash.com/docs/4.17.15#isObjectLike

frontend/src/components/applyForm/widgets/budget/budgetValueGuards.ts
```diff
@@ -0,0 +1,17 @@
...
+  return typeof value === "object" && value !== null;
+}
+
+export function asMoney(value: unknown): string | undefined {
```
doug-s-nava, line 7: `asMoney` makes some sense in terms of the scope in which this is used, but seems better to think of this as `toStringOrUndefined`

frontend/src/components/applyForm/widgets/budget/budgetValueGuards.ts
```diff
@@ -0,0 +1,17 @@
...
+  return typeof value === "string" ? value : undefined;
+}
+
+export function getStringOrUndefined(
```
doug-s-nava, line 11: ok I just proposed `getStringOrUndefined` so this one would be... `getStringOrUndefinedAtPath`?

frontend/src/components/applyForm/widgets/budget/Budget424aSectionA.tsx
```diff
@@ -1,49 +1,133 @@
...
+    return {
+      items: raw.activity_line_items.map(pickActivityItemA),
+      totals: pickBudgetSummary(raw.total_budget_summary),
+    };
+  }
+
+  if (isRecord(raw)) {
+    const items: ActivityItem[] = [];
+    for (let i = 0; i < 4; i++) {
+      items.push(pickActivityItemA(raw[String(i)]));
```
doug-s-nava, line 84: is there a reason we can't index into this using numbers?

frontend/src/components/applyForm/widgets/budget/Budget424aSectionA.tsx
```diff
@@ -1,49 +1,133 @@
...
+  const getItemVal = (
+    row: number,
+    path: keyof ActivityItem,
+  ): string | undefined => itemAt(row)[path] as string | undefined;
+  const getBudgetVal = (
+    row: number,
+    path: keyof BudgetSummary,
+  ): string | undefined => itemAt(row).budget_summary?.[path];
+
+  const COLUMNS = BUDGET_ACTIVITY_COLUMNS;
```
doug-s-nava, line 130: why reassign?

frontend/src/components/applyForm/widgets/budget/Budget424aSectionA.tsx
```diff
@@ -1,49 +1,133 @@
...
+  formContext,
+}: UswdsWidgetProps<T, S, F>): JSX.Element {
+  const rootFormDataFromContext = (
+    formContext as { rootFormData?: unknown } | undefined
+  )?.rootFormData;
+  const rawValue: unknown = rootFormDataFromContext ?? value ?? {};
+  const errors = (rawErrors as FormValidationWarning[]) || [];
+  const { items, totals } = normalizeSectionAValue(rawValue);
+  const getErrorsA = getErrorsForSection("A");
+  const itemAt = (row: number): ActivityItem => items[row] ?? {};
```
doug-s-nava, line 120: [suggestion] useCallback or move this out of the component and pass in items?

frontend/src/components/applyForm/widgets/budget/Budget424aSectionB.tsx
```diff
@@ -9,110 +9,110 @@ import {
...
   // Section B rows (A-K)
-  const ROW_I_KEY: FieldKey = "total_direct_charge_amount"; // row i
-  const ROW_K_KEY: FieldKey = "total_amount"; // row k
+  // row i
+  const ROW_I_KEY: FieldKey = "total_direct_charge_amount";
+  // row k
+  const ROW_K_KEY: FieldKey = "total_amount";
 
-  interface HelperTextProps {
+  const ColHelper: React.FC<{
```
doug-s-nava, line 93: most of the stuff between line 89 and 160 can move out of the component render

frontend/src/components/applyForm/widgets/budget/Budget424aSectionD.tsx
```diff
@@ -64,6 +64,7 @@ function Budget424aSectionD<
       label: "E",
     },
   ] as const;
+
   type ColKey = (typeof quarters)[number]["key"];
```
doug-s-nava, line 68: move type out (most of this can come out but that'd be the first thing I'd move)

frontend/src/components/applyForm/widgets/budget/Budget424aSectionE.tsx
```diff
@@ -20,47 +22,100 @@ interface FederalFundEstimates {
...
+      return {
+        items: itemsProp.map(pickActivityItem),
+        totals: pickFederalFundEstimates(totalsProp),
+      };
+    }
 
-type TotalFederalFundEstimates = FederalFundEstimates | undefined;
+    const items: ActivityItem[] = [];
+    for (const index of BUDGET_ACTIVITY_COLUMNS) {
+      items.push(pickActivityItem(rawValue[String(index)]));
```
doug-s-nava, line 81: same question here as elsewhere around the need for coercing - does the copy pasta here speak to something that can be split out into a shared function?

frontend/src/components/applyForm/widgets/budget/Budget424aSectionF.tsx
```diff
@@ -1,176 +1,153 @@
...
-      {helper ? (
-        <div className="font-sans-2xs text-italic margin-bottom-1">
-          {helper}
-        </div>
-      ) : null}
-      {children}
-    </div>
+    | undefined;
+
+  const directChargesSchema = properties?.direct_charges_explanation as
```
doug-s-nava, line 60: lots of equivocation around whether these schemas exist - is that to guard against the schema changing? Aren't we pretty sure this stuff will be there / if it's not, wouldn't all of this need to change anyway? Is this all to work around typing?

frontend/src/components/applyForm/widgets/budget/Budget424aSectionF.tsx
```diff
@@ -1,176 +1,153 @@
...
+  if (context && typeof context === "object" && "rootSchema" in context) {
+    const candidate = (context as { rootSchema?: unknown }).rootSchema;
+    return candidate && typeof candidate === "object"
+      ? (candidate as RJSFSchema)
+      : undefined;
+  }
+  return undefined;
+}
 
 function Budget424aSectionF<
```
doug-s-nava, line 28: I may be missing something but these all seem like pretty normal fields. Why are these being represented as this custom section?

frontend/src/components/applyForm/widgets/budget/budgetConstants.ts
```diff
@@ -0,0 +1,20 @@
+export const BUDGET_ACTIVITY_COLUMNS = [0, 1, 2, 3] as const;
```
doug-s-nava, line 1: seems like lots of stuff can move into this file

frontend/src/components/applyForm/widgets/budget/budgetErrorLabels.ts
```diff
@@ -0,0 +1,327 @@
+import { FormValidationWarning } from "src/components/applyForm/types";
```
doug-s-nava, line 1: they aren't paying you enough

frontend/src/components/applyForm/widgets/budget/budgetErrorLabels.ts
```diff
@@ -0,0 +1,327 @@
...
+
+  const position = parseRowColumn(section, id);
+  if (!position) {
+    return (errors ?? []).filter((e) => e.field === id).map((e) => e.message);
+  }
+
+  const columnString =
+    columnLabel === "letter"
+      ? toColumnLetter(position.columnNumber)
+      : String(position.columnNumber);
```
doug-s-nava, line 53: since these are being consumed in a template string later on, I'm pretty sure that that will handle any type coercion of numbers that needs to happen

frontend/src/components/applyForm/widgets/budget/budgetSchemas.ts
```diff
@@ -0,0 +1,31 @@
+// currency-like string;
```
doug-s-nava, line 1: man, hardcoding form schemas into the frontend makes me itchy. Let's keep thinking about a way to do this where we don't have to disregard the form schema coming from the API

frontend/src/components/applyForm/widgets/budget/budgetSchemas.ts
```diff
@@ -0,0 +1,31 @@
...
+export const amountSchema = {
+  type: "string" as const,
+  pattern: "^\\d*([.]\\d{2})?$",
+  maxLength: 14,
+};
+
+// Text schemas used
+// in A/F (tweak limits if needed)
+export const shortText50 = {
```
doug-s-nava, line 10: looks like this and the one below it are unused

frontend/src/components/applyForm/widgets/budget/budgetUiComponents.tsx
```diff
@@ -0,0 +1,183 @@
...
+export const DATA_CELL_BASE_CLASS =
+  "border-bottom-0 border-top-0 padding-05 verticle-align-top sf424a__cell";
+
+export type DataCellProps = {
+  children: React.ReactNode;
+  className?: string;
+  height?: string | number;
+};
+
+export const DataCell: React.FC<DataCellProps> = ({
```
doug-s-nava, line 17: can we use the type declaration for components that we've got elsewhere in the app `(props): {proptypes}` rather than the generic?

frontend/src/components/applyForm/widgets/budget/budgetUiComponents.tsx
```diff
@@ -0,0 +1,183 @@
...
+    formClassName={formClassName}
+    inputClassName={`${inputClassName}${bordered ? " border-2px" : ""}${
+      rawErrors && rawErrors.length ? " usa-input--error" : ""
+    }`}
+    inputMode="decimal"
+    pattern="-?[0-9]*[.,]?[0-9]{0,2}"
+    maxLength={14}
+    value={value}
+    placeholder="0.00"
+    onInput={(e: React.FormEvent<HTMLInputElement>) => {
```
doug-s-nava, line 79: what's this doing? seems complex enough we should split it out and test it separately

frontend/src/components/applyForm/ApplyForm.tsx
myduong-navapbc: `key={saved ? "after-save" : "before-save"}`, the key swaps and React re-renders

frontend/src/components/applyForm/ApplyForm.tsx
myduong-navapbc: yes, after i grabbed the latest from main. i kept getting type errors. - but if we want to move away from using translate i am happy to remove this :D

frontend/src/components/applyForm/README.md
```diff
@@ -6,6 +6,10 @@ This component takes form and UI schemas to create a form. The form schema confo
 
 The UI schema is custom for this project. It was inspired by [react-jsonschema-form](https://github.com/rjsf-team/react-jsonschema-form). RJSF was not chosen because it doesn't work with server actions, it doesn't support hierarchical UI layouts natively, and the validation will move to the API which reduces the value of the tightly coupled validation in RJSF. There is a [plugin for layouts](https://github.com/audibene-labs/react-jsonschema-form-layout), however it hasn't been updated in 4 years and still requires a lot of abstraction. This component provides most of the UI rendering with a ~100 line recursive function and uses the same library for validation. The widgets however are patterned after the RJSF widgets.
 
+## UI Template
```
doug-s-nava, line 9: the UI template you're referring to is the basic format for the widgets? Maybe make that more clear here since `ui template` as a term isn't found anywhere else in our docs or code

PR #6329, opened 2025-09-08
[Issue #6155] Add core XML generation service foundation

### LINKED ISSUE #6155: Initial Proof of Concept for JSON to XML conversion
### Summary

To prepare to build and utilize a JSON to XML conversion tool, we want to build a proof of concept on a smaller scale.

Proposed proof-of-concept form: @chouinar to develop a list of suggested POC fields from [SF-424](https://apply07.grants.gov/apply/forms/readonly/SF424_2_1-V2.1.pdf)

Scope:
- Generate XML from Simpler JSON
- Rename layer
- Configuration (likely JSON?) for defining transformation rules for a given field

When transforming, we need a way to configure what to do with each field (whether we define it as `our_field->rule` or `xml_field->rule` I'll leave up to the developer, both approaches are probably viable, but have different tradeoffs).

It is fine if there are multiple passes on the transformation (eg. one to rename things, and another to transform to XML), whatever approach is the most flexible will be best.

Goal: Get the majority (~60%) of Simpler XML matching the existing Grants.gov XML output.

The remaining 40% will be more complex: structure/heirachy, XML attributes, removing stuff from JSON that has no place in the XML, etc. We will have a follow-on ticket for that work.

### Acceptance criteria

Prototype is created that shows: 
- [ ] Simpler JSON can be converted to XML
- [ ] Most of the XML output of Simpler proof-of-concept form matches Grants.gov XML output
- [ ] Reviewed and aligned with Matt D and Michael C
- [ ] Collaborate with Chris K to create follow-on tickets

mikehgrantsgov:
## Summary

Fixes #6155 

This PR implements the core XML generation service, providing JSON to XML conversion for SF-424 forms. This is a POC that demonstrates the foundational architecture without advanced features like attachment handling, XSD validation, and developer tooling (these will come later).

## Changes proposed

Implements initial proof of concept for JSON to XML conversion.
Add `XMLGenerationService` with basic field mappings. 
Configuration-driven transformations loaded from `sf424.py`
`BaseTransformer` for applying field mappings
Basic field mappings for SF-424 forms
Comprehensive tests

## Context for reviewers

This PR combines 3 previous PRs into one, including value transformations, data handling, conditional logic and light business rule handling. A future set of PRs will cover CLI utilities, parameterized XSD validation, and full verification against the XSD.

## Validation steps

See added tests primarily in `tests/src/services/xml_generation/`

### DIFF
--- a/api/tests/src/form_schema/rule_processing/test_json_rule_util.py
+++ b/api/tests/src/form_schema/rule_processing/test_json_rule_util.py
@@ -4,7 +4,6 @@ from src.form_schema.rule_processing.json_rule_util import (
     _get_index_from_str,
     build_path_str,
     get_field_values,
-    get_nested_value,
     make_relative_path_absolute,
     populate_nested_value,
 )
@@ -134,46 +133,6 @@ def test_populate_nested_value_value_is_not_array():
         ValueError, match="Unable to populate nested value, value in path is not a list"
     ):
         populate_nested_value({"my_field": 10}, ["my_field[*]"], "hello")
-
-
-@pytest.mark.parametrize(
-    "json_data,path,expected_value",
-    [
-        ({"my_field": 5}, ["my_field"], 5),
-        ({"nested": {"path": {"to": {"value": 10}}}}, ["nested", "path", "to", "value"], 10),
-        # Path doesn't fully exist
-        ({}, ["whatever", "path"], None),
-        ({"whatever": {}}, ["whatever", "path"], None),
-        # Can fetch a whole chunk
-        (
-            {"nested": {"path": {"to": {"value": "hello"}}}},
-            ["nested"],
-            {"path": {"to": {"value": "hello"}}},
-        ),
-        (
-            {"nested": {"path": ["hello", "there", "this is a text"]}},
-            ["nested", "path"],
-            ["hello", "there", "this is a text"],
-        ),
-        # Passing in an empty path returns itself
-        ({"example": 5, "nested": {"field": 100}}, [], {"example": 5, "nested": {"field": 100}}),
-    ],
-)
-def test_get_nested_value(json_data, path, expected_value):
-    assert get_nested_value(json_data, path) == expected_value
-
-
-@pytest.mark.parametrize(
-    "path,expected_value",
-    [
-        (["array_field[0]", "x"], 1),
-        (["array_field[*]", "x"], [1, 3]),
-        (["array_field[*]", "nested_array[*]", "a"], [10, 15, 5]),
-        (["array_field[*]", "nested_array[*]", "b"], [4, 6]),
-    ],
-)
-def test_get_nested_value_of_arrays(path, expected_value):
-    assert get_nested_value(COMPLEX_ARRAY_DATA, path) == expected_value
 
 
 @pytest.mark.parametrize(
--- a/api/tests/src/util/test_dict_util.py
+++ b/api/tests/src/util/test_dict_util.py
@@ -1,6 +1,6 @@
 import pytest
 
-from src.util.dict_util import diff_nested_dicts, flatten_dict
+from src.util.dict_util import diff_nested_dicts, flatten_dict, get_nested_value
 
 
 @pytest.mark.parametrize(
@@ -139,3 +139,54 @@ def test_diff_nested_dicts(dict1, dict2, expected_output):
     sorted_result = sorted(result, key=lambda x: x["field"])
 
     assert expected_sorted == sorted_result
+
+
+# Test data for get_nested_value tests
+COMPLEX_ARRAY_DATA = {
+    "array_field": [
+        {"x": 1, "y": "hello", "nested_array": [{"a": 10, "b": 4}, {"a": 15}]},
+        {"x": 3, "y": "there", "z": "words", "nested_array": [{"a": 5, "b": 6}]},
+        {"nested_array": [{"g": 100}]},
+        {"e": "text"},
+    ]
+}
+
+
+@pytest.mark.parametrize(
+    "json_data,path,expected_value",
+    [
+        ({"my_field": 5}, ["my_field"], 5),
+        ({"nested": {"path": {"to": {"value": 10}}}}, ["nested", "path", "to", "value"], 10),
+        # Path doesn't fully exist
+        ({}, ["whatever", "path"], None),
+        ({"whatever": {}}, ["whatever", "path"], None),
+        # Can fetch a whole chunk
+        (
+            {"nested": {"path": {"to": {"value": "hello"}}}},
+            ["nested"],
+            {"path": {"to": {"value": "hello"}}},
+        ),
+        (
+            {"nested": {"path": ["hello", "there", "this is a text"]}},
+            ["nested", "path"],
+            ["hello", "there", "this is a text"],
+        ),
+        # Passing in an empty path returns itself
+        ({"example": 5, "nested": {"field": 100}}, [], {"example": 5, "nested": {"field": 100}}),
+    ],
+)
+def test_get_nested_value(json_data, path, expected_value):
+    assert get_nested_value(json_data, path) == expected_value
+
+
+@pytest.mark.parametrize(
+    "path,expected_value",
+    [
+        (["array_field[0]", "x"], 1),
+        (["array_field[*]", "x"], [1, 3]),
+        (["array_field[*]", "nested_array[*]", "a"], [10, 15, 5]),
+        (["array_field[*]", "nested_array[*]", "b"], [4, 6]),
+    ],
+)
+def test_get_nested_value_of_arrays(path, expected_value):
+    assert get_nested_value(COMPLEX_ARRAY_DATA, path) == expected_value


api/src/services/xml_generation/models.py
```diff
@@ -0,0 +1,21 @@
...
+from uuid import UUID
+from pydantic import BaseModel
+
+
+class XMLGenerationRequest(BaseModel):
+    """Request model for XML generation."""
+    
+    application_id: UUID
+    application_form_id: UUID
+    form_name: str = "SF424_4_0"
```
chouinar, line 12: Can we adjust how the input works to this logic?

Anywhere we'd be doing this processing we'll have already fetched the application form record which is connected to a form already. We could either have it _just_ take in an application form OR have it take in the JSON of the application form + form object.

That way none of this logic even needs to reference a DB, it shouldn't need to.

api/src/services/xml_generation/models.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class XMLGenerationResponse(BaseModel):
+    """Response model for XML generation."""
+    
+    success: bool
+    xml_data: Optional[str] = None
+    error_message: Optional[str] = None
```
chouinar:
```suggestion
    xml_data: str | None = None
    error_message: str | None = None
```

api/src/services/xml_generation/service.py
```diff
@@ -0,0 +1,112 @@
...
+    from src.db.models.competition_models import ApplicationForm
+
+logger = logging.getLogger(__name__)
+
+
+class XMLGenerationService:
+    """Service for generating XML from JSON application data."""
+
+    def __init__(self):
+        self.logger = logger
```
chouinar: Why add this to the class?

api/src/services/xml_generation/transformers/base_transformer.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class BaseTransformer:
+    """Base class for transforming JSON data according to configuration rules."""
+
+    def __init__(self, config: XMLTransformationConfig):
+        self.config = config
+        self.field_mappings = config.get_field_mappings()
+
+    def transform(self, source_data: Dict[str, Any]) -> Dict[str, Any]:
```
chouinar:
```suggestion
    def transform(self, source_data: dict[str, Any]) -> dict[str, Any]:
```

api/src/form_schema/forms/sf424.py
```diff
@@ -1057,6 +1057,86 @@
...
+    "field_mappings": {
+        # Core application information
+        "submission_type": "SubmissionType",
+        "application_type": "ApplicationType",
+        "date_received": "DateReceived",
+        
+        # Applicant information
+        "organization_name": "OrganizationName",
+        "employer_taxpayer_identification_number": "EmployerTaxpayerIdentificationNumber",
+        "sam_uei": "SAMUniqueEntityIdentifier",
```
chouinar: We'll want to double-check these, this is just `SAMUEI` in the XML

api/src/form_schema/forms/sf424.py
```diff
@@ -1057,6 +1057,86 @@
...
+    },
+
+    "xml_structure": {
+        "root_element": "SF424_4_0",
+        "version": "4.0"
+    },
+
+    "field_mappings": {
+        # Core application information
+        "submission_type": "SubmissionType",
```
chouinar: Might be something you handle in a later PR, but can we make these an object so we can potentially add configuration for transforms like:

```py
"submission_type": {"name": "SubmissionType"}
```

api/src/services/xml_generation/config.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class XMLTransformationConfig:
+    """Manages transformation configuration for converting JSON to XML."""
+
+    def __init__(self, form_name: str):
+        self.form_name = form_name
+        self._config = self._load_config()
+
+    def _load_config(self) -> dict[str, Any]:
```
chouinar: How do you expect this to work in later PRs? We should have the config as something on the Form table (fine that it isn't yet, no need to adjust DB schemas while hashing it out).

api/src/services/xml_generation/service.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def generate_xml(self, db_session: "db.Session", request: XMLGenerationRequest) -> XMLGenerationResponse:
```
chouinar: If this function only took in the request containing the users JSON + the form object, we could avoid anything with the DB in this code. Since anywhere that would call this would have already handled fetching these from the DB (eg. when making the ZIP file) we should be able to just not do any DB code.

api/src/form_schema/forms/sf424.py
```diff
@@ -1057,6 +1057,86 @@
...
+        "organization_name": "OrganizationName",
+        "employer_taxpayer_identification_number": "EmployerTaxpayerIdentificationNumber",
+        "sam_uei": "SAMUniqueEntityIdentifier",
+        
+        # Address information
+        "address_line_1": "AddressLine1",
+        "address_line_2": "AddressLine2",
+        "city": "City",
+        "county": "County",
+        "state_code": "StateCode",
```
chouinar: These fields are in an object, is that handled yet?

mikehgrantsgov: This is going to be handled in a subsequent PR around more complex transformations. Goal is to review this for general direction, then build on the more complex pieces.

api/src/services/xml_generation/service.py
mikehgrantsgov: Makes sense, removed the DB dependency here

api/src/form_schema/forms/sf424.py
mikehgrantsgov: Makes sense, updated this.

api/src/services/xml_generation/transformers/base_transformer.py
mikehgrantsgov: Updated these refs

api/src/services/xml_generation/models.py
mikehgrantsgov: Updated this as well.

api/src/form_schema/forms/sf424.py
mikehgrantsgov: Actually, scratch that. I added some of this functionality now (with a test) to account for these nested objects. More complex transforms would come later. See updated test.

api/src/form_schema/forms/sf424.py
mikehgrantsgov: I cross-referenced this against the Global Library XSD and form XSD and made a few mapping updates.

api/src/services/xml_generation/service.py
mikehgrantsgov: No need in particular, just got started copying from an older class. Removed now.

api/src/services/xml_generation/models.py
mikehgrantsgov: Makes sense. The current XMLGenerationRequest model was updated to take the application data directly.

api/src/services/xml_generation/config.py
mikehgrantsgov: I'm thinking something like?:

This PR, basic:
`from src.form_schema.forms.sf424 import FORM_XML_TRANSFORM_RULES`

```
# Future PR: Load config from from Form table. Add a column like `xml_transformation_config`
def _load_config(self) -> dict[str, Any]:
    form = db.session.query(Form).filter_by(form_name=self.form_name).first()
    return form.xml_transformation_config
```

chouinar: We can probably forgo having a config that needs to be loaded in the full implementation. If the form table has the config, then we'll already have things connected.

This config could just be a dict for now if that simplifies things.

api/src/services/xml_generation/service.py
```diff
@@ -0,0 +1,92 @@
...
+            root = ET.Element(root_element_name)
+
+        # Add data elements
+        for field_name, value in data.items():
+            if value is not None:
+                self._add_element_to_parent(root, field_name, value)
+
+        # Generate XML string
+        ET.indent(root, space="  ")
+        xml_string = ET.tostring(root, encoding="unicode", xml_declaration=True)
```
chouinar, line 82: Doesn't need to be handled now, but probably want some way to easily switch the output between a pretty-print XML format, and a condensed one. While I put the pretty-print version on the ticket, the actual format in the endpoint/submission zip is the condensed form.

api/tests/src/services/xml_generation/test_xml_generation_service.py
```diff
@@ -0,0 +1,173 @@
...
+        response = service.generate_xml(request)
+
+        # Verify error response
+        assert response.success is False
+        assert response.xml_data is None
+        assert "No application data provided" in response.error_message
+
+    def test_generate_xml_with_none_application_data(self):
+        """Test XML generation when application data is None (validation error)."""
+        from pydantic import ValidationError
```
chouinar: Can we put imports at the top of the file?

Is there some convention/IDE change? - I've seen this more frequently on a few different people's PRs

api/src/form_schema/forms/sf424.py
```diff
@@ -1057,6 +1057,74 @@
...
+        "application_type": {"name": "ApplicationType"},
+        "date_received": {"name": "DateReceived"},
+        # Applicant information
+        "organization_name": "OrganizationName",
+        "employer_taxpayer_identification_number": "EmployerTaxpayerIdentificationNumber",
+        "sam_uei": "SAMUEI",
+        # Address information (nested under Applicant element)
+        "applicant_address": {
+            "name": "Applicant",
+            "type": "nested_object",
```
chouinar: Seeing some of this, it's oddly similar to what I've setup for the JSON rule config down to the recursion itself.

Might be able to borrow some of the patterns for that?

Importantly, it assumes everything is an object unless it sees a specific key. So for example:
```json
{
  "x": {
      "y": {
          "z": {
              "gg_pre_population": {"rule": "something"}
           }
       }
   }
}
```
Means I can define the path without any complexity (except arrays) and only when I get to an actual value do I need to do anything special. Might have some reusability.

api/tests/src/services/xml_generation/test_xml_generation_service.py
mikehgrantsgov: I have a VSCode plugin that auto imports, which I'm assuming defaults to function-scope to avoid any possible circular dependencies? I configured it to just apply to the top in all cases.

api/src/form_schema/forms/sf424.py
mikehgrantsgov: I'm playing around with this and will see how we can possible share some implementation or, at very least, some of the recursive patterns.

api/src/services/xml_generation/service.py
mikehgrantsgov: It wasn't difficult to add a pretty-print mode for this. Added as an option.

api/src/form_schema/forms/sf424.py
mikehgrantsgov: So I made a few changes that are in the spirit of the JSON rule stuff without refactoring both into a shared utility (which we could possibly do later?). We used the rule handler pattern (e.g. `gg_pre_population`) analogously as `xml_transform` (possible we could add other rules to this later), a path-based recursive structure where we identify destination targets. I think this is flexible, uses existing patterns but doesn't explicitly (yet) borrow code.

api/src/services/xml_generation/constants.py
```diff
@@ -0,0 +1,15 @@
+"""Constants for XML generation services.
+
+This module contains shared constants, regex patterns, and other values
+used across the XML generation functionality.
+"""
+
+# Currency validation regex from SF-424A form schema
+# Allows: empty string, integers, decimals with exactly 2 places, decimals starting with dot
+CURRENCY_REGEX = r"^\d*([.]\d{2})?$"
```
chouinar: Do we need the non-negative currency regex? I think each of the forms allows negatives.

api/src/services/xml_generation/constants.py
```diff
@@ -0,0 +1,15 @@
...
+
+CURRENCY_REGEX_WITH_NEGATIVES = r"^(-)?\d*([.]\d{2})?$"
```
chouinar: Could we copy the comment explaining the regex?

```suggestion
# * An optional negative sign
# * Any number of digits
# * An optional decimal point
# * Then exactly 2 digits - if there was a decimal
CURRENCY_REGEX_WITH_NEGATIVES = r"^(-)?\d*([.]\d{2})?$"
```

api/src/services/xml_generation/service.py
```diff
@@ -0,0 +1,100 @@
...
+
+            # Log transformation results for development
+            logger.info(
+                f"XML generation successful: {len(transformed_data)} fields transformed from {len(request.application_data)} input fields for {request.form_name}"
+            )
+
+            return XMLGenerationResponse(success=True, xml_data=xml_string)
+
+        except Exception as e:
+            logger.error(f"XML generation failed: {e}")
```
chouinar:
```suggestion
            logger.exception("XML generation failed")
```

api/src/services/xml_generation/transformers/base_transformer.py
```diff
@@ -0,0 +1,207 @@
...
+            rules: Transformation rules at this level
+            path: Current path in the data structure
+
+        Returns:
+            Transformed data at this level
+        """
+        result: dict[str, Any] = {}
+
+        # Iterate over rules at this level
+        for key, rule_config in rules.items():
```
chouinar, line 60: Any way we might adjust the logic in this loop to pass functionality off to functions? Almost 100 lines, and in some places it goes 4 if-statements deep.

For example, could we move the chunk of code regarding handling of None (lines 73-101) into a function? Not sure if that would work with the continues you have, but it seems like we might be able to pull out chunks like that to make this a bit shorter.

api/src/services/xml_generation/transformers/base_transformer.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def _get_nested_value(self, data: dict[str, Any], path: list[str]) -> Any:
+        """Get a nested value from a dictionary using a path."""
+        current = data
+        for part in path:
+            if isinstance(current, dict) and part in current:
+                current = current[part]
+            else:
+                return None
+        return current
```
chouinar: I see this is similar to the get nested value I had for the JSON rules. I had to rewrite this function to handle arrays, while the SF424 doesn't have arrays, how would the transformation logic handle an array?

For example, assume we need to transform:
```json
{
  "my_field": [
    {"x": 1},
    {"x": 2},
    {"x": 3}
   ]
}
```
What I ended up doing was rewriting the function to just use the jsonpath_ng library. If that'd work for this use case, maybe we should move that `get_nested_value` function somewhere more central and just share it.

api/src/services/xml_generation/constants.py
mikehgrantsgov: We don't, good call. Removed this

api/src/services/xml_generation/transformers/base_transformer.py
mikehgrantsgov: I moved `get_nested_value` to `dict_util` and updated imports.

api/src/services/xml_generation/transformers/base_transformer.py
mikehgrantsgov: I cleaned this up a little,

chouinar: Much better, thanks

api/src/util/dict_util.py
```diff
@@ -78,3 +80,64 @@ def _convert_iterables_to_set(data: Any) -> Any:
             return {tuple(d.items()) for d in data}
         return set(data)
     return data
+
+
+def get_nested_value(data: dict, path: list[str]) -> Any:
```
chouinar, line 85: Could you move the tests for this from the `test_json_rule_util` over to `test_dict_util`?

chouinar (commented):
Just one last thing

api/src/util/dict_util.py
mikehgrantsgov: Yep, makes sense. Just moved!

PR #6361, opened 2025-09-09
[Issue #6155] Add value transforms and data formatting [2/4]

mikehgrantsgov:
## Summary

Work for #6155

## Changes proposed

Add value transformations - error handling by falling back if transform fails. Integrated with existing rule processing patterns
Boolean yes/no conversion
Currency formatting
String utilities (truncation, case conversion if necessary e.g. state codes, agency codes, )

## Context for reviewers

This builds on the core PR by adding some extra functionality to the JSON -> XML conversion. Subsequent PRs will tackle more complex. The following PR, if deemed necessary, will include more advanced transforms and conditional logic.

## Validation steps

See added tests.

### DIFF
--- a/api/src/form_schema/forms/sf424.py
+++ b/api/src/form_schema/forms/sf424.py
@@ -1108,21 +1108,64 @@ FORM_XML_TRANSFORM_RULES = {
     },
     "project_start_date": {"xml_transform": {"target": "ProjectStartDate"}},
     "project_end_date": {"xml_transform": {"target": "ProjectEndDate"}},
-    # Funding information - direct field mappings
-    "federal_estimated_funding": {"xml_transform": {"target": "FederalEstimatedFunding"}},
-    "applicant_estimated_funding": {"xml_transform": {"target": "ApplicantEstimatedFunding"}},
-    "state_estimated_funding": {"xml_transform": {"target": "StateEstimatedFunding"}},
-    "local_estimated_funding": {"xml_transform": {"target": "LocalEstimatedFunding"}},
-    "other_estimated_funding": {"xml_transform": {"target": "OtherEstimatedFunding"}},
-    "program_income_estimated_funding": {
-        "xml_transform": {"target": "ProgramIncomeEstimatedFunding"}
+    # Funding information - with currency formatting
+    "federal_estimated_funding": {
+        "xml_transform": {
+            "target": "FederalEstimatedFunding",
+            "value_transform": {"type": "currency_format", "params": {"ensure_decimal": False}},
+        }
     },
-    "total_estimated_funding": {"xml_transform": {"target": "TotalEstimatedFunding"}},
-    # Review and certification - direct field mappings
+    "applicant_estimated_funding": {
+        "xml_transform": {
+            "target": "ApplicantEstimatedFunding",
+            "value_transform": {"type": "currency_format", "params": {"ensure_decimal": False}},
+        }
+    },
+    "state_estimated_funding": {
+        "xml_transform": {
+            "target": "StateEstimatedFunding",
+            "value_transform": {"type": "currency_format", "params": {"ensure_decimal": False}},
+        }
+    },
+    "local_estimated_funding": {
+        "xml_transform": {
+            "target": "LocalEstimatedFunding",
+            "value_transform": {"type": "currency_format", "params": {"ensure_decimal": False}},
+        }
+    },
+    "other_estimated_funding": {
+        "xml_transform": {
+            "target": "OtherEstimatedFunding",
+            "value_transform": {"type": "currency_format", "params": {"ensure_decimal": False}},
+        }
+    },
+    "program_income_estimated_funding": {
+        "xml_transform": {
+            "target": "ProgramIncomeEstimatedFunding",
+            "value_transform": {"type": "currency_format", "params": {"ensure_decimal": False}},
+        }
+    },
+    "total_estimated_funding": {
+        "xml_transform": {
+            "target": "TotalEstimatedFunding",
+            "value_transform": {"type": "currency_format", "params": {"ensure_decimal": False}},
+        }
+    },
+    # Review and certification - with value transformations
     "state_review": {"xml_transform": {"target": "StateReview"}},
     "state_review_available_date": {"xml_transform": {"target": "StateReviewAvailableDate"}},
-    "delinquent_federal_debt": {"xml_transform": {"target": "DelinquentFederalDebt"}},
-    "certification_agree": {"xml_transform": {"target": "CertificationAgree"}},
+    "delinquent_federal_debt": {
+        "xml_transform": {
+            "target": "DelinquentFederalDebt",
+            "value_transform": {"type": "boolean_to_yes_no"},
+        }
+    },
+    "certification_agree": {
+        "xml_transform": {
+            "target": "CertificationAgree",
+            "value_transform": {"type": "boolean_to_yes_no"},
+        }
+    },
     # Authorized representative - direct field mappings
     "authorized_representative_title": {
         "xml_transform": {"target": "AuthorizedRepresentativeTitle"}
--- a/api/src/form_schema/forms/sf424.py
+++ b/api/src/form_schema/forms/sf424.py
@@ -2,7 +2,7 @@ import uuid
 
 import src.form_schema.forms.shared_schema as shared_schema
 from src.db.models.competition_models import Form
-from src.services.xml_generation.value_transformers import NO_VALUE
+from src.services.xml_generation.constants import NO_VALUE
 
 FORM_JSON_SCHEMA = {
     "type": "object",
--- a/api/src/form_schema/forms/sf424.py
+++ b/api/src/form_schema/forms/sf424.py
@@ -2,6 +2,7 @@ import uuid
 
 import src.form_schema.forms.shared_schema as shared_schema
 from src.db.models.competition_models import Form
+from src.services.xml_generation.value_transformers import NO_VALUE
 
 FORM_JSON_SCHEMA = {
     "type": "object",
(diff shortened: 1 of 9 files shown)


api/src/form_schema/forms/sf424.py
```diff
@@ -1108,21 +1108,64 @@
...
-    "applicant_estimated_funding": {"xml_transform": {"target": "ApplicantEstimatedFunding"}},
-    "state_estimated_funding": {"xml_transform": {"target": "StateEstimatedFunding"}},
-    "local_estimated_funding": {"xml_transform": {"target": "LocalEstimatedFunding"}},
-    "other_estimated_funding": {"xml_transform": {"target": "OtherEstimatedFunding"}},
+    # Funding information - with currency formatting
+    "federal_estimated_funding": {
+        "xml_transform": {
+            "target": "FederalEstimatedFunding",
+            "value_transform": {"type": "currency_format", "params": {"ensure_decimal": False}},
+        }
```
chouinar: Something to think about for the structure of this - If a field in our data corresponded to multiple fields in the XML, how would we go about handling that?

There are two that we'll have to account for in existing forms. The `applicant_type_code` in the SF424 (an array of strings becomes 3 separate string fields) and the `activity_title` in the SF424A which is a bit more complex and becomes an attribute on multiple parts of an XML object

api/src/services/xml_generation/value_transformers.py
```diff
@@ -0,0 +1,175 @@
...
+        value: Boolean value to transform
+
+    Returns:
+        "Yes" for True, "No" for False
+
+    Raises:
+        ValueTransformationError: If value is not a boolean
+    """
+    if isinstance(value, bool):
+        return "Yes" if value else "No"
```
chouinar: We'd actually want to transform these to:
* True -> `Y: Yes`
* False -> `N: No`

The `YesNoDataType` from:
https://apply07.grants.gov/apply/system/schemas/GlobalLibrary-V2.0.xsd

Probably useful to define a bunch of constants for these strings if multiple transforms need the values.
There are other yes/no variants, but the SF424 doesn't use those, so can wait on dealing with those.

api/src/services/xml_generation/value_transformers.py
```diff
@@ -0,0 +1,175 @@
...
+        value: Currency value (string or number)
+        ensure_decimal: Whether to ensure .00 decimal places
+
+    Returns:
+        Formatted currency string
+
+    Raises:
+        ValueTransformationError: If currency formatting fails
+    """
+    if isinstance(value, (int, float)):
```
chouinar: I almost want to say not to support floats here, if we're using floats to store monetary amounts we're doing something very wrong upstream.

api/src/services/xml_generation/value_transformers.py
```diff
@@ -0,0 +1,175 @@
...
+        cleaned = re.sub(r"[^\d.]", "", value)
+
+        if not cleaned:
+            raise ValueTransformationError(f"Invalid currency value: '{value}'")
+
+        try:
+            # Convert to float and back to ensure proper format
+            float_value = float(cleaned)
+            if ensure_decimal:
+                return f"{float_value:.2f}"
```
chouinar: Do not use floats on these values, if we want to do any sort of conversion, we want to use Decimal.

Here is what I am thinking - if the value is a string, do nothing. We should assume that the XML transformation is happening on data that is validated. Monetary amounts are already constrained to a regex that makes them either an integer or two-decimal value.

api/src/services/xml_generation/value_transformers.py
```diff
@@ -0,0 +1,175 @@
...
+
+    # Extract additional parameters
+    params = transformation_config.get("params", {})
+
+    try:
+        return transformer(value, **params)
+    except Exception as e:
+        logger.warning(f"Value transformation failed for {transform_type}: {e}")
+        # Return original value if transformation fails (graceful degradation)
+        return value
```
chouinar: If a transformation fails, we should probably just let that error propagate up. We wouldn't want to partially transform the data as there isn't much value in that. Any transformation error likely is a configurational problem or bug.

api/src/services/xml_generation/transformers/base_transformer.py
```diff
@@ -112,8 +114,24 @@ def _apply_transform_rule(
...
+                    transformed_value = apply_value_transformation(
+                        source_value, transform_rule["value_transform"]
+                    )
+                    logger.debug(
+                        f"Applied value transformation at {'.'.join(path)}: {source_value} -> {transformed_value}"
+                    )
+                except ValueTransformationError as e:
+                    logger.warning(f"Value transformation failed at {'.'.join(path)}: {e}")
+                    # Use original value if transformation fails
+                    transformed_value = source_value
```
chouinar: Just a suggestion to rearrange slightly. I don't think we want the try/except, if there is a transformation error, let it go, we don't want to transform if we're hitting errors.

```suggestion
            # Simple transformation - apply value transformation if specified
            transformed_value = source_value

            # Check if there's a value transformation specified
            rule = transform_rule.get("value_transform")
            if rule is not None:
                transformed_value = apply_value_transformation(source_value, rule)
                logger.debug(
                    f"Applied value transformation at {'.'.join(path)}: {source_value} -> {transformed_value}"
                )

```

api/src/services/xml_generation/value_transformers.py
```diff
@@ -0,0 +1,175 @@
...
+    return value[:max_length] if len(value) > max_length else value
+
+
+# Registry of available transformations
+TRANSFORMATION_REGISTRY: dict[str, Callable[..., Any]] = {
+    "boolean_to_yes_no": transform_boolean_to_yes_no,
+    "currency_format": transform_currency_format,
+    "string_case": transform_string_case,
+    "truncate_string": transform_truncate_string,
+}
```
chouinar, line 131: How do these transforms handle `None`? Is that handled before the transform functions get called?

Guessing we'll want a way to configure how `None` behaves in the config for a given field, I could see one of two ways we'd want it to work:
* Exclude the field entirely from the XML output
* Include the field, set to null

I think the first of those would be the most common / could be the default?

mikehgrantsgov: I added a proposal for how we can handle `None`. `"null_handling"` in the `xml_transform` rule can be `exclude`, `include_null`, and `default_value` to provide a sensible default when it's None (if we want it included but not blank). By default we `exclude`

api/src/services/xml_generation/value_transformers.py
mikehgrantsgov: Yep, makes sense. I made sure the error bubbles up.

api/src/services/xml_generation/value_transformers.py
mikehgrantsgov: Updated this to accommodate those changes.

api/src/services/xml_generation/value_transformers.py
mikehgrantsgov: Makes sense, removed any float stuff

api/src/services/xml_generation/value_transformers.py
mikehgrantsgov: These were externalized into constants

api/src/services/xml_generation/transformers/base_transformer.py
```diff
@@ -60,17 +62,47 @@ def _process_transform_rules(
...
+                        default_value = transform_rule.get("default_value", "")
+                        source_value = default_value
+                        logger.debug(
+                            f"Using default value '{default_value}' for {'.'.join(current_path)}"
+                        )
+                    else:
+                        logger.warning(
+                            f"Unknown null_handling '{none_handling}' for {'.'.join(current_path)}, excluding field"
+                        )
+                        continue
```
chouinar: If there is an unexpected configuration, we should just error. This code only will run in two ways I can see:
* A backend process says to convert JSON to XML (can add a try-except, and probably log an error and skip that application for the batch)
* Unit tests - we'd want to easily know that something is off

And the former shouldn't happen if we have unit tests, where it'll be easier to validate errors. In general if data or configuration is unexpected, just let it fail, let whatever is calling the XML transformer handle how to deal with errors.

api/src/services/xml_generation/value_transformers.py
```diff
@@ -0,0 +1,167 @@
...
+    Raises:
+        ValueTransformationError: If value is not a string or cannot be parsed
+    """
+    if not isinstance(value, str):
+        raise ValueTransformationError(
+            f"Currency value must be string (validated upstream), got {type(value)}"
+        )
+
+    # Remove any non-numeric characters except decimal point
+    cleaned = re.sub(r"[^\d.]", "", value)
```
chouinar: This removes a leading negative sign. I think I'd rather we err on the side of running it through a regex to verify the format, and erroring if the format is invalid, but not changing it.

api/src/services/xml_generation/value_transformers.py
```diff
@@ -0,0 +1,167 @@
...
+
+    if not cleaned:
+        raise ValueTransformationError(f"Invalid currency value: '{value}'")
+
+    # Validate format
+    if not re.match(r"^\d+(\.\d{1,2})?$", cleaned):
```
chouinar: We can move it somewhere more central, but I've got a regex in the schemas of `r"^(-)?\d*([.]\d{2})?$"`

api/src/services/xml_generation/transformers/base_transformer.py
```diff
@@ -60,17 +62,47 @@ def _process_transform_rules(
...
+                        result[target_field] = None
                         logger.debug(
-                            f"Transformed {'.'.join(current_path)} -> {target_field}: {source_value}"
+                            f"Including None value for {'.'.join(current_path)} -> {target_field}"
                         )
+                        continue
+                    elif none_handling == "default_value":
+                        # Use configured default value
+                        default_value = transform_rule.get("default_value", "")
+                        source_value = default_value
```
chouinar: If we say it should have a default value, should we error if there was no default provided?

mikehgrantsgov: Yep, makes sense. Made that change.

api/src/services/xml_generation/value_transformers.py
mikehgrantsgov: I put this in a constants file and referenced that from the 2 places it's used.

api/src/services/xml_generation/value_transformers.py
mikehgrantsgov: Updated this as well.

api/src/services/xml_generation/transformers/base_transformer.py
mikehgrantsgov: Makes sense. We are now just erroring

Issue #6194, opened 2025-08-29
[BUG] False error message in form UI


chris-kuryak:
### Summary

From Lucas 8/28:

> This error message is confusing because (a) it says 'applicant' is required on the field 'Applicant District', and (b) it's on a field I've properly filled out. I assume it's because the project end date is missing, but I'm not sure.

> <img width="380" height="480" alt="Image" src="https://github.com/user-attachments/assets/72376bb4-6cff-4006-9a13-912d94ad5934" />

### Acceptance criteria

- [ ] Error message does not appear on Applicant District when it is appropriately filled out and even though a nested field is triggering an error
- [ ] UAT passed

chris-kuryak:
UAT passed

Issue #6187, opened 2025-08-29
[BUG] Telephone number validation not functioning as expected


chris-kuryak:
### Summary

**To reproduce:** 

1. On form SF424, Fill out Contact "Telephone Number" with some information. Keep "AOR Telephone Number" blank.
2. Click "Save"

**Current response:** Validation error in the form UI shows Contact "Telephone Number" as "authorized representative phone number is required" even though it passes validation.

<img width="548" height="163" alt="Image" src="https://github.com/user-attachments/assets/655968cd-93c5-4a91-bb06-0b48c97a4c9a" />

### Acceptance criteria

- [ ] Expected behavior is achieved upon following steps to reproduce
- [ ] UAT passed

chris-kuryak:
UAT passed

Issue #6185, opened 2025-08-29
Nested field errors show appropriately in the form UI


chris-kuryak:
### Summary
Wait until #6127  is complete before trying to fix this, since it may be fixed by #6127 

As a user, I want the validation warning in the form UI to reflect what the actual error is.

Currently, a nested field shows the parent objects error in the UI.

From Lucas on 8/28:
"applicant" as the missing field for the three different fields is shown here. I assume it's because behind the scenes it's making an applicant object, but more clear messaging is helpful.

<img width="455" height="720" alt="Image" src="https://github.com/user-attachments/assets/262e71f3-f409-490b-977d-aaf43c30dc92" />

### Acceptance criteria

- [ ] Nested field validation errors display the field name, not the parent object name in the form UI
- [ ] UAT testing complete

PR #6363, opened 2025-09-10
[Issue #6185] error handling

acouch:
## Summary

Fixes / Work for #6383 #6187 #6194 #6185

## Changes proposed

1. Add [FieldErrors.tsx](https://github.com/HHS/simpler-grants-gov/pull/6363/files#diff-2c6da2929e6933b72f330188b10d41a001a0df19cf3c7a781bdd24b295baa0a7) and update all widgets to use that function that prints all errors instead of just the first one. None of the forms currently print more than one error per field, but this component supports that.
2. add helper functions for converting pointer vs path
3. adds handling for errors that uses the uischema to check the errors
4. adds a `formatted` and `htmlField` to the warnings
5. adds formatting for form errors!
6. updates `getFieldName` to `getFieldNameForHtml` to be more explicit

## Validation steps

test the forms

### DIFF
--- a/frontend/src/components/applyForm/widgets/MultiSelectWidget.tsx
+++ b/frontend/src/components/applyForm/widgets/MultiSelectWidget.tsx
@@ -155,6 +155,7 @@ export default function MultiSelect<
           value={v}
         />
       ))}
+      {/* ComboBox widget changes the id which breaks handling of idFor and anchor links */}
       <span id={id}></span>
       <ComboBox
         ref={comboRef}


frontend/src/components/applyForm/widgets/AttachmentUploadWidget.tsx
```diff
@@ -108,20 +108,23 @@ const AttachmentUploadWidget = (props: UswdsWidgetProps) => {
     fileInputRef.current?.clearFiles();
   };
 
-  const hasError = rawErrors.length > 0;
-  const describedBy = hasError ? `${id}-error` : undefined;
+  const error = rawErrors.length ? true : undefined;
```
acouch, line 111: This is how the other widgets describe the error.

frontend/src/components/applyForm/ApplyForm.tsx
```diff
@@ -65,9 +65,9 @@ const ApplyForm = ({
       window.confirm(t("unsavedChangesWarning")),
   });
 
-  const { formData, error, saved } = formState;
+  const { error, saved } = formState;
 
-  const formObject = !isEmpty(formData) ? formData : savedFormData;
+  const formObject = savedFormData || new FormData();
```
acouch, line 70: We aren't mutating the formData on the server, we fetch it every time from the page after the form load.

frontend/src/components/applyForm/PrintForm.tsx
```diff
@@ -8,15 +8,15 @@ import { FormFields } from "./FormFields";
 import { UiSchema } from "./types";
 
 export default function PrintForm({
+  attachments,
```
acouch, line 11: Nothing beats alphabetical order.

frontend/src/components/applyForm/types.ts
```diff
@@ -35,19 +34,19 @@ export interface SetFormDataFunction {
   (data: FormData): void;
 }
 
-export type FieldErrors = ErrorObject<
```
acouch, line 38: This wasn't in use

frontend/src/components/applyForm/utils.ts
```diff
@@ -87,10 +314,10 @@ export const getFieldSchema = ({
   schema,
   formSchema,
 }: {
-  definition: `/properties/${string}` | undefined;
```
acouch, line 90: Adequately describing json path with typescript requires iterations it wasn't built for.

frontend/src/components/applyForm/utils.ts
```diff
@@ -210,17 +480,29 @@ export const getFieldConfig = ({
...
-    name = getFieldName({ definition, schema });
-    const path = getFieldPath(name);
-    value = getByPointer(formData, path) as string | number | undefined;
+    fieldSchema = getFieldSchema({
+      definition,
+      schema,
+      formSchema,
+    }) as RJSFSchema;
+    fieldName = getNameFromDef({ definition, schema });
+    htmlFieldName = getFieldNameForHtml({ definition, schema });
```
acouch, line 481: Updated `name` to `htmlFieldName` to be more explicit.

myduong-navapbc:
on initial run through:
there are some fields that return 2 of the same error:

the SF 424 LLL and SF 424 B
Screenshot 2025-09-12 at 2 19 53 PM

and the SF 424 (standard)
Delinquent Federal Debt, states that it has an error but the error class isnt passed through. This is also in a different ticket https://github.com/HHS/simpler-grants-gov/issues/6097

Screenshot 2025-09-12 at 2 25 00 PM
Screenshot 2025-09-12 at 2 25 11 PM

frontend/src/components/applyForm/widgets/AttachmentUploadWidget.tsx
```diff
@@ -108,20 +108,23 @@ const AttachmentUploadWidget = (props: UswdsWidgetProps) => {
...
     <React.Fragment key={`${id}-key`}>
       <input type="hidden" name={id} value={attachmentId ?? ""} />
-
-      {error && (
-        <ErrorMessage id={`error-for-${id}`}>
-          {String(rawErrors[0])}
-        </ErrorMessage>
-      )}
-
+      {error && <ErrorMessage id={`error-for-${id}`}>{errors}</ErrorMessage>}
```
doug-s-nava: [before approval] what happens if we do this instead? I think React render logic will take care of things so the useMemo stuff  isn't necessary
```suggestion
      {error && <ErrorMessage id={`error-for-${id}`}><FieldErrors type={schema.type} fieldName={id} rawErrors={rawErrors} /></ErrorMessage>}
```

frontend/src/components/applyForm/FieldErrors.tsx
```diff
@@ -0,0 +1,37 @@
...
+  type,
+  fieldName,
+  rawErrors,
+}: {
+  type: JSONSchema7TypeName | JSONSchema7TypeName[] | string | undefined;
+  fieldName: string;
+  rawErrors: string[] | Record<string, string>[] | undefined;
+}): React.ReactNode | string | null => {
+  const safeType = Array.isArray(type) && type.length > 0 ? type[0] : type;
+  if (!Array.isArray(rawErrors) || rawErrors.length < 0 || !rawErrors)
```
doug-s-nava: [before approval] a negative length? do we want `!rawErrors.length` instead?

frontend/src/components/applyForm/PrintForm.tsx
doug-s-nava: we can add a linting rule for that, I'm sure. Should we?

frontend/src/components/applyForm/utils.ts
```diff
@@ -18,6 +19,232 @@ import {
...
+  path,
+}: {
+  definition: string;
+  errors: FormValidationWarning[];
+  fieldName: string;
+  fieldSchema: SchemaField;
+  formSchema: RJSFSchema;
+  path: string;
+}): FormattedFormValidationWarning | null => {
+  const parent = definition.replace(/\/properties\/\w+$/, "");
```
doug-s-nava, line 38: [non blocking] can this use `removePropertyPaths`?

frontend/src/components/applyForm/utils.ts
```diff
@@ -18,6 +19,232 @@ import {
...
+    }
+  }
+};
+
+export const buildWarningTree = (
+  formUiSchema: UiSchema,
+  formValidationWarnings: FormValidationWarning[],
+  formSchema: RJSFSchema,
+) => {
+  const e: FormattedFormValidationWarning[] = [];
```
doug-s-nava: [non blocking] Probably a personal failing but closing over something that is passed as an argument into a function instead of relying on return values tends to break my brain. Can we set up a reducer and return the errors that way (as I'll try to demonstrate above?)

frontend/src/components/applyForm/utils.ts
```diff
@@ -18,6 +19,232 @@ import {
...
+        label: string;
+        name: string;
+        description?: string;
+      },
+  parent: { label: string; name: string; description?: string } | null,
+  formValidationWarnings: FormValidationWarning[],
+  errors: FormattedFormValidationWarning[] = [],
+  formSchema: RJSFSchema,
+) => {
+  if (
```
doug-s-nava, line 166: an attempt at doing this with a return value

```suggestion
{
  if (
    !Array.isArray(uiSchema) &&
    typeof uiSchema === "object" &&
    "children" in uiSchema
  ) {
    return formatValidationWarnings(
      uiSchema.children,
      {
        label: uiSchema.label,
        name: uiSchema.name,
        description: uiSchema.description,
      },
      formValidationWarnings,
      formSchema,
    );
  } else if (Array.isArray(uiSchema)) {
    const childErrors = uiSchema.reduce((errors, node) => {
      if ("children" in node) {
        const nodeError = formatValidationWarnings(
          node.children,
          {
            label: node.label,
            name: node.name,
            description: node.description,
          },
          formValidationWarnings,
          errors,
          formSchema,
        );
        return errors.concat([nodeError])
      } else if (!parent && ("definition" in node || "schema" in node)) {
        const error = findValidationError(
          formValidationWarnings,
          Array.isArray(node.definition) ? node.definition[0] : node.definition,
          node.schema,
          formSchema,
        );
        if (error) {
          return errors.concat([error])
        }
        return errors;
      }
    }, []);
    if (parent) {
      const parentErrors = uiSchema.reduce((errors, node) => {
        if ("children" in node) {
          const nodeError = formatValidationWarnings(
            node.children,
            {
              label: node.label,
              name: node.name,
              description: node.description,
            },
            formValidationWarnings,
            errors,
            formSchema,
          );
          return errors.concat(nodeError)
        } else {
          const error = findValidationError(
            formValidationWarnings,
            Array.isArray(node.definition)
              ? node.definition[0]
              : node.definition,
            node.schema,
            formSchema,
          );
        if (error) {
          return errors.concat([error])
        }
        return errors;
        }
      }, []);
    }
    return childErrors.concat(parentErrors)
  }
```

frontend/src/components/applyForm/utils.ts
```diff
@@ -18,6 +19,232 @@ import {
...
+const formatValidationWarnings = (
+  uiSchema:
+    | UiSchema
+    | {
+        children: UiSchema;
+        label: string;
+        name: string;
+        description?: string;
+      },
+  parent: { label: string; name: string; description?: string } | null,
```
doug-s-nava: [non blocking] can we move the optional arg to the end, or make this a destructured object argument signature?

frontend/tests/components/applyForm/utils.test.ts
```diff
@@ -1,16 +1,26 @@
+/* eslint-disable @typescript-eslint/no-unsafe-argument */
+/* eslint-disable testing-library/no-node-access */
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
 import { RJSFSchema } from "@rjsf/utils";
 import sflllSchema from "tests/components/applyForm/sflll.mock.json";
 
-import { UiSchemaField } from "src/components/applyForm/types";
+import { UiSchema, UiSchemaField } from "src/components/applyForm/types";
 import {
+  buildWarningTree,
```
doug-s-nava, line 9: [non blocking] would love to see some tests for nestedWarningForField, formatValidationError, formatValidationWarnings

frontend/src/components/applyForm/widgets/AttachmentUploadWidget.tsx
acouch: I updated that, thanks.

frontend/src/components/applyForm/utils.ts
acouch: @doug-s-nava The above removes just that last property in the line, not all of them.

acouch:
@myduong-navapbc can you retest what you put in your comment? I don't get dupes and the Delinquent Budget works correctly for me.

image

frontend/src/components/applyForm/FieldErrors.tsx
acouch: This logic was removed

frontend/src/components/applyForm/utils.ts
acouch: I updated this. We should do `buildFormTree` next

frontend/tests/components/applyForm/utils.test.ts
acouch: I added tests for `buildWarningTree` and removed `formatValidationWarnings` and consider the other private, they are only called in `buildWarningTree`.

frontend/src/components/applyForm/widgets/CheckboxWidget.tsx
```diff
@@ -64,6 +65,9 @@ function CheckboxWidget<
 
   return (
     <FormGroup error={error} key={`form-group__checkbox--${id}`}>
+      {error && (
+        <FieldErrors fieldName={id} rawErrors={rawErrors as string[]} />
```
acouch, line 69: The errors are strings for the widgets, but that was expanded to keep the error object for the multiField.

It would be simpler to just keep the same structure which we could do by updating `getWarningsForField`.

frontend/src/components/applyForm/FieldErrors.tsx
```diff
@@ -0,0 +1,29 @@
...
+  fieldName,
+  rawErrors,
+}: {
+  fieldName: string;
+  rawErrors: string[] | undefined;
+}): React.ReactNode | null => {
+  if (rawErrors) {
+    if (rawErrors.length > 1) {
+      return (
+        <ErrorMessage id={`error-for-${fieldName}`}>
```
acouch, line 13: This was simplified to only handle array of strings from the field widgets. The multiField widgets currently have their own logic. This should be simplified at some point.

frontend/src/components/applyForm/PrintForm.tsx
acouch: yes

frontend/src/components/applyForm/utils.ts
acouch: I updated this to return errors. Thanks.

acouch:
@myduong-navapbc @doug-s-nava I updated the budget logic 
image

I'm seeing 

image
 
but it is an issue on `main` as well.

frontend/src/components/applyForm/widgets/MultiSelectWidget.tsx
```diff
@@ -158,7 +155,7 @@ export default function MultiSelect<
           value={v}
         />
       ))}
-
+      <span id={id}></span>
```
doug-s-nava, line 159: what's this for?

edit: oh, is this for the warning link to have the correct id to anchor to? That seems worth a comment in the code

doug-s-nava (approved):
looking good! some things to ticket for later:

* delinquent federal debt explanation file uploader missing left border on error
* whatever is going on with the competition number field
* linting / prettier rule for alphabetical argument order 
* enabling use of combobox within select widget? I'm a little confused about the logic there but it seems like something we just aren't supporting yet, but should

✅ ✅ ✅ ✅

frontend/src/components/applyForm/widgets/MultiSelectWidget.tsx
acouch: I commented on it higher up, wasn't sure if was too verbose. I added here one in my last commit.

doug-s-nava: ok yeah I saw that comment but didn't connect the dots that we'd need to add an empty span with the original id in order to get the anchors to work

PR #6368, opened 2025-09-10
[Issue #6208] Create a utility for generating test opportunities

### LINKED ISSUE #6208: Create a utility to generate SQL to make a test opportunity
### Summary

The current work needed to create a test opportunity is very time consuming.
So we want to create a script that generates data needed for creating opportunities.

Line of thinking: Create a prototype script that generates SQL to generate test opportunities

### Acceptance criteria

- [ ] Utility generates SQL to create a test opportunity
- [ ] Reviewed and aligned with necessary engineering stakeholders

chouinar:
## Summary
Fixes #6208

## Changes proposed
Created a script that fetches an opportunity from any environment, and outputs SQL to create an opportunity like it.

## Context for reviewers
We've had an increased need to setup test opportunities, but up until now that process required manually building SQL, and an opportunity is a lot of data across a lot of tables and the process is incredibly tedious, error-prone and all around not fun. This script helps automate most of that by using an existing opportunity as a baseline, swapping out the primary keys, and printing out the SQL you'll need to do this.

It does not yet handle copying opportunity attachment or competition instruction files as that'll require converting the s3 CDN paths back to s3 paths, and also generating s3 copy commands. Doable, but not that urgent to handle, can follow-up on that when needed.

The tests aren't too elaborate, and I don't validate the SQL generated - this is meant for helping setup test data, setting up a bunch of tests for a utility isn't something I think is valuable time (and would take a lot of time and be very complex).

## Validation steps
For example, let's say I wanted to copy this opportunity: https://staging.simpler.grants.gov/opportunity/dab5a586-69e3-4710-91d5-9dfcc40ae9d4

I could run `make cmd args="task generate-opportunity-sql --environment=staging --opportunity-id=dab5a586-69e3-4710-91d5-9dfcc40ae9d4"` and it would output a bunch of SQL statements to make the opportunity (this screenshot isn't nearly all of it):
Screenshot 2025-09-10 at 3 17 35 PM

### DIFF
--- a/api/src/task/opportunities/generate_opportunity_sql.py
+++ b/api/src/task/opportunities/generate_opportunity_sql.py
@@ -71,6 +71,7 @@ class GenerateOpportunitySqlTaskConfig(PydanticBaseEnvConfig):
 class OppSqlRequest:
     opportunity_id: str
     environment: str
+    forms: list[str] = dataclasses.field(default_factory=list)
     # This is just so we can avoid printing in tests
     print_output: bool = True
 
@@ -195,6 +196,9 @@ class GenerateOpportunitySqlProcessor:
         data_to_insert["opportunity_summary_id"] = self.opp_data_container.opportunity_summary_id
         data_to_insert["opportunity_id"] = self.opp_data_container.opportunity_id
         data_to_insert["legacy_opportunity_id"] = self.opp_data_container.legacy_opportunity_id
+        data_to_insert["agency_contact_description"] = "Bob Smith"
+        data_to_insert["agency_email_address"] = "fake_mail@fake.com"
+        data_to_insert["agency_email_address_description"] = "fake contact info"
 
         self.add_section_to_sql("Opportunity Summary")
         self.sql_statements.append(build_sql("opportunity_summary", data_to_insert))
@@ -295,6 +299,7 @@ class GenerateOpportunitySqlProcessor:
             data_to_insert["is_simpler_grants_enabled"] = False
             # This isn't returned, but putting in SQL to be easily found
             data_to_insert["grace_period"] = 0
+            data_to_insert["contact_info"] = "Bob Smith"
 
             # Figure out the assistance listing ID by iterating over
             # what we inserted and finding the matching assistance listing number
@@ -326,6 +331,7 @@ class GenerateOpportunitySqlProcessor:
             ### Competition Forms
             competition_forms = competition.get("competition_forms", [])
 
+            form_ids = set()
             for competition_form in competition_forms:
                 form = competition_form.get("form", {})
                 competition_form_data = {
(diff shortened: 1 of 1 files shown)


joshtonava:
What do I need to do to use the utility locally ? I think im forgetting a step

When I run 

`bin/run-command --environment-variables "$(jq -c . ./config.json)" api dev '["poetry", "run", "flask", "data-migration", "load-transform", "--no-load"]'`

I get the error 

```
Exception: Error fetching opportunity from staging: {
  "data": {},
  "errors": [],
  "internal_request_id": "953c85ea-f85c-4f3d-b6e2-54ba37e1f657",
  "message": "Invalid API key",
  "status_code": 401
}
```

api/src/task/opportunities/generate_opportunity_sql.py
```diff
@@ -0,0 +1,507 @@
...
+
+
+class OppDataContainer:
+    def __init__(self) -> None:
+        self.opportunity_id = str(uuid.uuid4())
+        self.opportunity_summary_id = str(uuid.uuid4())
+        # Note that this legacy ID could technically be non-unique, these values
+        # are all above the actual ones in our environments, but there's no check here
+        # to verify that.
+        self.legacy_opportunity_id = random.randint(1_000_000, 10_000_000)
```
joshtonava, line 86: Note: I see your comment acknowledging this is non unique. This likely wont be an issue often. Hopefully if/when there is a collision its obvious to the user. 

IF I understand how this is supposed to be used, someone would be pasting in the SQL somewhere so they should understand the error right?

api/src/task/opportunities/generate_opportunity_sql.py
```diff
@@ -0,0 +1,507 @@
...
+        self.build_opportunity_assistance_listing_sql(opportunity_json)
+        # Competition + Competition Forms + Competition Open to Applicant
+        self.build_competition_sql(opportunity_json)
+
+        # Opportunity Attachments and Competition Instructions
+        # are not yet handled as they would also require moving
+        # files around on s3, and we can follow-up on that if desired.
+
+        if self.request.print_output:
+            print("####### SQL STATEMENTS\n")
```
joshtonava, line 122: Question: Why print and not logger?

chouinar:
> What do I need to do to use the utility locally ? I think im forgetting a step
> 
> When I run
> 
> `bin/run-command --environment-variables "$(jq -c . ./config.json)" api dev '["poetry", "run", "flask", "data-migration", "load-transform", "--no-load"]'`
> 
> I get the error
> 
> ```
> Exception: Error fetching opportunity from staging: {
>   "data": {},
>   "errors": [],
>   "internal_request_id": "953c85ea-f85c-4f3d-b6e2-54ba37e1f657",
>   "message": "Invalid API key",
>   "status_code": 401
> }
> ```

I think you copied the wrong command, but if you got that error, you probably haven't set the API key correctly. It needs to be the API auth token for the given environment you want to call which you can grab from 1password.

api/src/task/opportunities/generate_opportunity_sql.py
chouinar: Yeah, there isn't really a way we could guarantee / check that this and some of the other IDs are unique without looking at the DB itself which for a utility to help generate SQL is beyond what we'd want to do.

If you hit the issue, you'd get a uniqueness constraint complaint and could adjust the number.

api/src/task/opportunities/generate_opportunity_sql.py
chouinar: I didn't want any of the formatting from the logger, just want something I can copy and paste from the console.

joshtonava:
> I think you copied the wrong command, but if you got that error, you probably haven't set the API key correctly. It needs to be the API auth token for the given environment you want to call which you can grab from 1password.

I think my config.json was missing as well. I added that and I tried the command with both of the staging API Auth Tokens in 1password.  

I apologize could you maybe DM me instructions for how to run this step by step?

chouinar:
> > I think you copied the wrong command, but if you got that error, you probably haven't set the API key correctly. It needs to be the API auth token for the given environment you want to call which you can grab from 1password.
> 
> I think my config.json was missing as well. I added that and I tried the command with both of the staging API Auth Tokens in 1password.
> 
> I apologize could you maybe DM me instructions for how to run this step by step?

You need to set the env var yourself - how are you doing that? The two options I use:
* Add it to your override.env
* Do `export NON_LOCAL_API_AUTH_TOKEN=<whatever>`

Issue #6352, opened 2025-09-09
Adjust how we determine the operation to not parse the entire attachment


chris-kuryak:
### Summary

We need to adjust how we determine the operation to not parse the entire attachment -- the information we need is appearing very early in the parsing.

### Acceptance criteria
- [ ] Submit application is able to find application name
- [ ] Entire attachment is not parsed to determine the operation

PR #6370, opened 2025-09-10
[Issue #6352] get operation name without parsing whole xml

jakobpederson:
## Summary

Work for #ISSUE 6352

## Changes proposed

We need to adjust how we determine the operation to not parse the entire attachment -- the information we need is appearing very early in the parsing.


api/src/legacy_soap_api/soap_payload_handler.py
```diff
@@ -209,16 +209,20 @@ def get_soap_operation_name(soap_xml: str | bytes) -> str:
     Get the SOAP operation name. Every valid SOAP request Body should
     have the global SOAP envelope namespace.
     """
+    operation_name = ""
```
chouinar, line 243: Could we add a test or two for this function to verify it works in the normal case, as well as when the request is formatted like the multipart one in the attachment endpoints?

jakobpederson: I'll add them

api/tests/src/legacy_soap_api/test_soap_payload_handler.py
```diff
@@ -223,3 +224,106 @@ def test_lists_xml_to_dict_with_ns(self):
         """
             ).encode()
         )
+
+
+class TestGetSoapOperationName(unittest.TestCase):
+    def test_get_soap_operation_name_ignores_header_and_attachment(self):
+        header = (
```
chouinar: Could we define these statically up top? I think we can also do a multiline string/bytes like:
```
whatever = b"""
abc
xyz
"""
```
If the newlines that'd introduce are an issue, I think we can also define a string like:
```
x = (
b"abc"
b"xyz"
b"lmno"
)

print(x) # b'abcxyzlmno'
```

api/src/legacy_soap_api/soap_payload_handler.py
```diff
@@ -203,22 +203,54 @@ def get_soap_envelope_from_payload(soap_message: str) -> SOAPEnvelopeData:
...
 
 
+def extract_soap_xml(soap_bytes: bytes) -> bytes:
+    soap_patterns = [b"<soap:Envelope", b"<soapenv:Envelope"]
+    max_tag_size = len(soap_patterns[1])
+    bytes_stream = io.BytesIO(soap_bytes)
+    chunk_size = 1000
+    buffer = b""
+    total_bytes_read = 0
+    while True:
```
chouinar: Should we have a max number of iterations just in case? If we're 100 iterations deep, we're probably not going to find it, but I'm not sure what a good number might be (10?)

api/tests/src/legacy_soap_api/test_soap_payload_handler.py
jakobpederson: Sure thing! I'll move them up top.

api/src/legacy_soap_api/soap_payload_handler.py
jakobpederson: I feel like maybe 5 would be all we need given it's near the top of the body. I'll add a max number.

chouinar (approved):

Issue #6013, opened 2025-08-20
[Task]  Policy Update to Allow ECS Task to hit ses v2 ListSuppressedDestinations API


babebe:
### Description

We are currently seeing an AccessDeniedException when the ECS task tries to call the SES v2 ListSuppressedDestinations API.

The same call works locally via the CLI, which confirms the issue is with the IAM role assigned to the ECS task.
Update IAM policy attached to the ECS task role with permissions needed for ses:ListSuppressedDestinations

https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonsimpleemailservicev2.html

### Risks and assumptions

Will not be able to track bouncing emails in our database until this is updated.

babebe:
error: ["<class 'botocore.exceptions.ClientError'>","An error occurred (AccessDeniedException) when calling the ListSuppressedDestinations operation: User: arn:aws:sts::315341936575:assumed-role/api-prod-app/5309b306a2994775a275f9ac8805df05 is not authorized to perform: ses:ListSuppressedDestinations on resource: * because no identity-based policy allows the ses:ListSuppressedDestinations action","<traceback object at 0x7fe2f190fe00>"]

mdragon:
Manually `terraform apply` and testing of the job shows we have the permissions right now.

PR #6374, opened 2025-09-11
[Issue #6013] Add perms for ListSuppresssedDestinations at the account level

mdragon:
## Summary

Work for #6013 

## Changes proposed

Add permissions for the API Tasks user to call SES and get the Suppressed Email List content

## Validation steps

Will apply in Dev manually and then run the job to validate it fixes the issue.


mdragon:
Tested successfully in Dev by manually running `terraform apply` and testing the job

PR #6378, opened 2025-09-11
Temp fully proxy API Gateway

Nava-JoshLong:
## Summary

Fixes / Work for #ISSUE 

## Changes proposed

- Temp comment out full proxy requirement for API token
- Temp comment out other endpoints since they are redundant right now

## Context for reviewers

This PR will comment out the majority of the logic for the API gateway so that it acts like a full proxy to the underlying ECS service. This will allow us to change the DNS from the ALB to the gateway, but not turn the gateway on, so that we can test the functionality and have a fast way to revert back in case any issues arise. Otherwise, we'd need to change the DNS, which can take time due to it being a long process

## Validation steps

Validation would be to hit the API gateway URL and see if things work as expected. I'm tempted to deploy the proxy as the none stage so we can test fully functionality before the dev DNS change occurs

### DIFF
--- a/infra/modules/service/api_gateway.tf
+++ b/infra/modules/service/api_gateway.tf
@@ -9,16 +9,17 @@ resource "aws_api_gateway_rest_api" "api" {
   body = jsonencode({
     "openapi" : "3.0.1",
     "paths" : {
-      "/health" : {
-        "get" : {
-          "x-amazon-apigateway-integration" : {
-            "type" : "http_proxy",
-            "httpMethod" : "GET",
-            "uri" : "https://${var.optional_extra_alb_domains[0]}/health",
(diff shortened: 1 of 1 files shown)


Nava-JoshLong:
I tested the changes, but since we don't have the DNS pointing to the gateway, I can't confirm anything yet. Should we still merge this @mdragon, or wait for the DNS?

mdragon:
@Nava-JoshLong ~~Let's hold off on merging just because we don't know when the DNS will actually change and we want to make sure we have time to test before the Wed Release.~~
Because this is really gated by the DNS change not the deploy of the API Gateway in Proxy mode, it's fine for this to go forward.

PR #6379, opened 2025-09-11
[Issue #6275] Add CommonGrants routes to API

### LINKED ISSUE #6275: [CommonGrants adoption] Address PR comments for CommonGrants endpoints
### Summary

Incorporate Nava's feedback on the PR to add CommonGrants routes to SGG API and merge in the PR.

### Acceptance criteria

- [x] PR comments have been addressed

### Dependencies

Must be done after https://github.com/HHS/simpler-grants-gov/issues/6273

DavidDudas-Intuitial:
## Summary

Fixes #6275

## Changes proposed

This PR implements the CommonGrants Protocol by adding 3 new routes to the existing API:

* `GET /common-grants/opportunities/`
* `GET /common-grants/opportunities/{oppId}`
* `POST /common-grants/opportunities/search/`

The implementation and tests are encapsulated in the following new directories:

* `api/src/api/common_grants/`
* `api/src/services/common_grants/`
* `api/tests/src/api/common_grants/`
* `api/tests/src/services/common_grants/`

The following existing files have been modified to support the new implementation:

* `api/Makefile` – add commands to generate and validate an isolated OpenAPI spec for new routes
* `api/pyproject.toml` – add dependency on `common-grants-sdk`
* `api/src/app.py` – add new blueprint and feature flag

The following new workflow file has been created to validate the implementation in a CI environment:

* `.github/workflows/ci-api-common-grants.yml`

Comprehensive documentation of the integration is in the following file:

* `api/src/api/common_grants/COMMON_GRANTS_INTEGRATION.md`

## Context for reviewers

The implementation in this PR leverages existing patterns:

* Uses `@api_key_multi_auth.login_required` decorator to require authentication for new routes
* Uses `@flask_db.with_db_session()` decorator for database dependency injection
* Uses `@flask_opensearch.with_search_client()` decorator for search client dependency injection 

## Validation steps

Installation:

```bash
poetry install
make init
```

Validation:

```bash
make check
make check-spec
```


api/src/api/common_grants/common_grants_routes.py
```diff
@@ -0,0 +1,152 @@
...
+            }
+        ),
+        404,
+    )
+
+
+@common_grants_blueprint.get("/opportunities")
+@common_grants_blueprint.input(PaginatedQueryParamsSchema, location="query")
+@common_grants_blueprint.output(OpportunitiesListResponseSchema)
+@api_key_multi_auth.login_required
```
chouinar, line 42: We should make the auth just be our new API key auth, we have some endpoints using this multi-auth that allows both the old and new approach while we transition, but we'll have the old one fully deprecated in about a month.

And because our naming is a bit confusing, that's the `api_user_key_auth` that uses the `X-API-Key` header.

api/src/api/common_grants/common_grants_routes.py
```diff
@@ -0,0 +1,152 @@
...
+import src.adapters.search as search
+import src.adapters.search.flask_opensearch as flask_opensearch
+from src.api.common_grants.common_grants_blueprint import common_grants_blueprint
+from src.auth.multi_auth import api_key_multi_auth, api_key_multi_auth_security_schemes
+from src.services.common_grants.opportunity_service import CommonGrantsOpportunityService
+
+logger = logging.getLogger(__name__)
+
+
+def generate_422_error(e: Exception) -> tuple[dict, int]:
```
chouinar: We chatted about errors, and I know you're figuring out whether we want to but some additional details.

If an API endpoint has an exception raised, that will always result in a 500 if allowed to propagate up. In most cases, we are fine with that, an unexpected error should be a 500 as it's probably a bug. If it's a 500, the message is also removed from the error. Note that any 500s will log the error, and we have alerts setup for any error logs, so we'll be notified. In short, if you want a 500, just raise an exception.

However, if the exception raised is an `HTTPError` from the APIFlask library, that will be caught and used to build the error message. We have a utility called `raise_flask_error` that we use for these cases to help build that error. For example, if we want to raise a 404, we don't bother with returning None to the route and then return a 404, we just do `raise_flask_error(404, "Whatever message")`. This function allows you to pass in a detail object which gets put in the `data` of the response, and a list of `validation_issues` which get put in the error list. In a way, we use `raise_flask_error` to build the error message and return it, even if we're a few layers down in the code.

api/src/api/common_grants/common_grants_routes.py
```diff
@@ -0,0 +1,152 @@
...
+def search_opportunities(search_client: search.SearchClient, json_data: dict) -> tuple[dict, int]:
+    """Search for opportunities based on the provided filters."""
+
+    # Validate input
+    request_schema = OpportunitySearchRequestSchema()
+    try:
+        validated_input = request_schema.load(json_data)
+        search_request = OpportunitySearchRequest(**validated_input)
+    except Exception as e:
+        return generate_422_error(e)
```
chouinar: If there is an exception here, is that an error we'd want to pass to the user, or would it indicate an issue with our schema/transformation? Should this be a 500?

api/src/api/common_grants/common_grants_routes.py
```diff
@@ -0,0 +1,152 @@
...
+    security=api_key_multi_auth_security_schemes,
+    responses=[200, 404],
+)
+@flask_db.with_db_session()
+def get_opportunity(db_session: db.Session, oppId: str) -> tuple[dict, int]:
+    """Get a specific opportunity by ID."""
+
+    # Create service and get query result
+    service = CommonGrantsOpportunityService(db_session)
+    response_object = service.get_opportunity(oppId)
```
chouinar: Generally the pattern we want to follow for our endpoints is:
```python
with db_session.begin():
      call_service(...)
```
This makes the DB request a transaction, which means if there is any error in that block, we automatically roll it back. Even those these endpoints are all GET endpoints and it doesn't matter, I like to keep that pattern just in case. There may be cases that some of our GET endpoints do write data, we're planning to add records to an audit table for some of our GET endpoints in the apply code this quad.

api/src/api/common_grants/common_grants_routes.py
```diff
@@ -0,0 +1,152 @@
...
+    service = CommonGrantsOpportunityService(db_session)
+    response_object = service.list_opportunities(
+        page=int(query_data.get("page", 1)), page_size=int(query_data.get("pageSize", 10))
+    )
+
+    # Hydrate response model
+    response_json = response_object.model_dump(by_alias=True, mode="json")
+    response_schema = OpportunitiesListResponseSchema()
+    validated_response = response_schema.load(response_json)
+    return validated_response, 200
```
chouinar: Do we need to convert it back and forth here? APIFlask, or really Marshmallow can convert just about any object to the JSON it needs, we usually just return the DB model record. This does require we have all the SQLAlchemy relationships loaded (the various `selectinload` calls we add to some of the DB queries) but if you have an object that pulled values from the DB objects, that shouldn't be an issue.

If we're wanting to validate the response object (which APIFlask does not do beyond coercing the types), then it's fine, although maybe we could add a function like `build_common_grants_response(data: Any, schema: Type[Schema], status_code: int = 200)` that we could call here like `return build_common_grants_response(response_object, OpportunitiesListResponseSchema, 200)`?

api/src/app.py
```diff
@@ -174,6 +176,10 @@ def register_blueprints(app: APIFlask) -> None:
         app.register_blueprint(form_blueprint)
         app.register_blueprint(competition_blueprint)
 
+    # CommonGrants Protocol endpoints
+    if endpoint_config.enable_common_grants_endpoints:
+        app.register_blueprint(common_grants_blueprint)
```
chouinar, line 181: If we want to enable this in dev/staging, can add to our terraform like so:
https://github.com/HHS/simpler-grants-gov/blob/2afbcc45902185f2b36a50df2851c515f3c7a2b0/infra/api/app-config/dev.tf#L48
And 
https://github.com/HHS/simpler-grants-gov/blob/2afbcc45902185f2b36a50df2851c515f3c7a2b0/infra/api/app-config/staging.tf#L48

api/src/app.py
```diff
@@ -54,6 +55,7 @@ class EndpointConfig(PydanticBaseEnvConfig):
     auth_endpoint: bool = Field(False, alias="ENABLE_AUTH_ENDPOINT")
 
     enable_apply_endpoints: bool = Field(False, alias="ENABLE_APPLY_ENDPOINTS")
+    enable_common_grants_endpoints: bool = Field(True, alias="ENABLE_COMMON_GRANTS_ENDPOINTS")
```
chouinar: Do we want this disabled in prod initially? Just while we test it in dev/staging, can flip later.

api/src/api/common_grants/common_grants_routes.py
```diff
@@ -0,0 +1,152 @@
...
+@common_grants_blueprint.doc(
+    summary="List opportunities",
+    description="Get a paginated list of opportunities, sorted by `lastModifiedAt` with most recent first.",
+    security=api_key_multi_auth_security_schemes,
+    responses=[200],
+)
+@flask_db.with_db_session()
+def list_opportunities(db_session: db.Session, query_data: dict) -> tuple[dict, int]:
+    """Get a paginated list of opportunities."""
+
```
chouinar: Could we add two lines related to logging here? First, a call to `add_extra_data_to_current_request_logs(...)` - this takes in a dict, and will add that information to every log message for the rest of this request's lifecycle. We usually put things like query parameters (assuming they're IDs or other non-PII values). It can make it very helpful for creating graphs and finding data in New Relic.

Secondly, a log message like `logger.info("GET /opportunities")`. While we automatically put the URL rule in the log messages, this helps us when debugging see that the log message got to our implementation of the API (past APIFlask having Marshmallow validate the request) but before the implementation. Also ends up being the first log message with the `add_extra_data_to_current_request_logs` params added.

api/src/services/common_grants/opportunity_service.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+class CommonGrantsOpportunityService:
+    """Service for managing opportunities in CommonGrants Protocol format."""
+
+    def __init__(self, db_session: Session) -> None:
+        """Initialize the service."""
+        self.db_session = db_session
+
+    def get_opportunity(self, opportunity_id: str) -> OpportunityResponse | None:
```
chouinar: As we chatted about, we have a get opportunity function: https://github.com/HHS/simpler-grants-gov/blob/main/api/src/services/opportunities_v1/get_opportunity.py

Note that we use `raise_flask_error`, depending on what we do with the error schema here, might need to catch the 404 it can raise and do something with that.

api/src/services/common_grants/transformation.py
```diff
@@ -0,0 +1,464 @@
...
+
+    Returns:
+        OppStatusOptions: The CommonGrants status enum value
+    """
+
+    STATUS_TO_CG_MAP = {
+        OpportunityStatus.FORECASTED.value: OppStatusOptions.FORECASTED,
+        OpportunityStatus.POSTED.value: OppStatusOptions.OPEN,
+        OpportunityStatus.CLOSED.value: OppStatusOptions.CLOSED,
+        OpportunityStatus.ARCHIVED.value: OppStatusOptions.CUSTOM,
```
chouinar: Should be able to exclude the `.value` from these since they're StrEnums
```suggestion
        OpportunityStatus.FORECASTED: OppStatusOptions.FORECASTED,
        OpportunityStatus.POSTED: OppStatusOptions.OPEN,
        OpportunityStatus.CLOSED: OppStatusOptions.CLOSED,
        OpportunityStatus.ARCHIVED: OppStatusOptions.CUSTOM,
```

api/src/services/common_grants/transformation.py
```diff
@@ -0,0 +1,464 @@
...
+    """
+
+    STATUS_FROM_CG_MAP = {
+        OppStatusOptions.FORECASTED: OpportunityStatus.FORECASTED,
+        OppStatusOptions.OPEN: OpportunityStatus.POSTED,
+        OppStatusOptions.CLOSED: OpportunityStatus.CLOSED,
+        OppStatusOptions.CUSTOM: OpportunityStatus.ARCHIVED,
+    }
+
+    return STATUS_FROM_CG_MAP.get(status, OppStatusOptions.CUSTOM)
```
chouinar: Should the fallback here instead be an OpportunityStatus (posted probably?)

---

For these mapping functions, should we have a fallback for the converting function, or should it error?

I feel like if the value isn't present, we probably want to be made aware to fix to code, maybe something like this to :
```py
value = STATUS_FROM_CG_MAP.get(status, None)
if value is None:
      logger.error(f"Unexpected OppStatusOptions {status} in transform")
      value = OpportunityStatus.POSTED # not sure what the fallback should be

return value
```

api/src/services/common_grants/transformation.py
```diff
@@ -0,0 +1,464 @@
...
+    timeline = OppTimeline(
+        postDate=(
+            SingleDateEvent(
+                name="Opportunity Posted",
+                date=_transform_date_to_cg(opportunity.created_at),
+                description="Date when the opportunity was first posted",
+            )
+            if opportunity.created_at
+            else None
+        ),
```
chouinar: The post date for an opportunity is in the summary. We'd want `summary.post_date` instead of the opportunity created_at, an opportunity can be created weeks or months before the post date.

api/src/services/common_grants/transformation.py
```diff
@@ -0,0 +1,464 @@
...
+        closeDate=(
+            SingleDateEvent(
+                name="Application Deadline",
+                date=_transform_date_to_cg(summary.close_date),
+                description="Deadline for submitting applications",
```
chouinar: (This isn't suggesting a change, more a heads up, something we probably want to discuss separately) `close_date` despite the name isn't an application deadline, it's just a value that says whether an opportunity is closed. An opportunity can have multiple deadlines and those values are stored on the competitions. 

For example, look at https://www.grants.gov/search-results-detail/340237 - it has a close date of Mar 14th, but if you go to the package tab, you can see all of the different close dates for the competitions.

api/src/services/common_grants/transformation.py
```diff
@@ -0,0 +1,464 @@
...
+    if not url:
+        return None
+
+    # If the URL already has a protocol, return it as is
+    if url.startswith(("http://", "https://")):
+        return url
+
+    # If it's a domain without protocol, add https://
+    if "." in url and not url.startswith(("http://", "https://", "ftp://", "file://")):
+        return f"https://{url}"
```
chouinar: Heads up that the additional info URL is frequently not a URL at all. Some of the most common values are things like `See Invitation Letter`, `See Related Documents` or `THIS IS NOT AN ACTIVE LINK. PLEASE CLICK ON THE WORDS "FULL ANNOUNCEMENT" AT THE TOP OF THIS PAGE.`. Unfortunately grants.gov doesn't make sure the link is actually a link. I think that last one having a `.` in it would get `https://` prepended.

api/src/services/common_grants/transformation.py
```diff
@@ -0,0 +1,464 @@
...
+        # Create timeline
+        timeline = OppTimeline(
+            postDate=(
+                SingleDateEvent(
+                    name="Opportunity Posted",
+                    date=_transform_date_to_cg(opp_data.get("created_at")),
+                    description="Date when the opportunity was first posted",
+                )
+                if opp_data.get("created_at")
+                else None
```
chouinar, line 256: Same as above, use post date from the summary.

api/src/services/common_grants/transformation.py
```diff
@@ -0,0 +1,464 @@
...
+
+    # If it's a domain without protocol, add https://
+    if "." in url and not url.startswith(("http://", "https://", "ftp://", "file://")):
+        return f"https://{url}"
+
+    # If it's not a valid URL format, return None
+    return None
+
+
+def transform_opportunity_to_cg(opportunity: Opportunity) -> OpportunityBase:
```
chouinar: Looks like we have two similar transforms for an opportunity from the DB versus an opportunity from search. Could combine them into one by converting an opportunity to JSON first. That's all we do for putting an opportunity in the search index in the first place: https://github.com/HHS/simpler-grants-gov/blob/main/api/src/search/backend/load_opportunities_to_index.py#L425 (there's some extra attachment stuff that probably isn't relevant yet)

api/src/services/common_grants/transformation.py
```diff
@@ -0,0 +1,464 @@
...
+            status=OppStatus(value=opp_status),
+            keyDates=timeline,
+            funding=funding,
+            source=validate_url(summary.get("additional_info_url")),
+            custom_fields={},
+            createdAt=opp_data.get("created_at"),
+            lastModifiedAt=opp_data.get("updated_at"),
+        )
+    except Exception as e:
+        logger.error(f"Failed to transform search result to CommonGrants format: {e}")
```
chouinar: If you do `logger.exception` that adds a stack trace and a few [other formats](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/logging/flask_logger.py#L235) to the logs automatically.

```suggestion
        logger.exception("Failed to transform search result to CommonGrants format")
```

api/src/services/common_grants/opportunity_service.py
```diff
@@ -0,0 +1,172 @@
...
+
+        # Use default values if not provided
+        if filters is None:
+            filters = OppFilters()
+        if sorting is None:
+            sorting = OppSorting(sort_by=OppSortBy.LAST_MODIFIED_AT)
+        if pagination is None:
+            pagination = PaginatedBodyParams()
+
+        # Convert search request to legacy format
```
chouinar: Nitpicky: We generally refer to grants.gov as legacy, can we avoid calling the simpler code that to avoid confusion?

api/tests/src/api/common_grants/test_opportunities_routes.py
```diff
@@ -0,0 +1,369 @@
...
+    assert "postDate" in key_dates
+
+    # Validate status structure
+    status = opportunity["status"]
+    assert "customValue" in status
+    assert "description" in status
+    assert "value" in status
+
+    # Validate date formats (ISO 8601)
+    from datetime import datetime
```
chouinar: Can we move this to the top of the file?

api/tests/src/services/common_grants/test_opportunity_service.py
```diff
@@ -0,0 +1,218 @@
...
+            assert response.status == 200
+            assert response.message == "Opportunities searched successfully using search client"
+            # Note: We can't easily verify the text search was applied without checking the actual data
+
+    def test_search_opportunities_with_all_parameters(self, mock_search_client):
+        """Test searching opportunities with all parameters specified."""
+        from common_grants_sdk.schemas.pydantic import OppFilters, OppSorting, PaginatedBodyParams
+        from common_grants_sdk.schemas.pydantic.filters import ArrayOperator, StringArrayFilter
+        from common_grants_sdk.schemas.pydantic.models import OppStatusOptions
+        from common_grants_sdk.schemas.pydantic.sorting import OppSortBy, SortOrder
```
chouinar: Could we have these at the top of the file? Seems like several tests use variations of these

api/src/api/common_grants/common_grants_routes.py
DavidDudas-Intuitial: Fixed in https://github.com/DavidDudas-Intuitial/simpler-grants-gov/commit/05618510fcda86ca9d62b10acd02c2f313223452

api/src/api/common_grants/common_grants_routes.py
DavidDudas-Intuitial: The conversion is necessary to go from pydantic models to marshmallow models, but the logic belongs in the service layer rather than the route layer. Fixed in https://github.com/DavidDudas-Intuitial/simpler-grants-gov/commit/05618510fcda86ca9d62b10acd02c2f313223452.

api/src/app.py
DavidDudas-Intuitial: Done in https://github.com/DavidDudas-Intuitial/simpler-grants-gov/commit/05618510fcda86ca9d62b10acd02c2f313223452

api/src/app.py
DavidDudas-Intuitial: Done in https://github.com/DavidDudas-Intuitial/simpler-grants-gov/commit/05618510fcda86ca9d62b10acd02c2f313223452

api/src/api/common_grants/common_grants_routes.py
DavidDudas-Intuitial: Done in https://github.com/DavidDudas-Intuitial/simpler-grants-gov/commit/05618510fcda86ca9d62b10acd02c2f313223452

api/src/services/common_grants/transformation.py
DavidDudas-Intuitial: Fixed in https://github.com/DavidDudas-Intuitial/simpler-grants-gov/commit/05618510fcda86ca9d62b10acd02c2f313223452

api/src/services/common_grants/transformation.py
DavidDudas-Intuitial: Done in https://github.com/DavidDudas-Intuitial/simpler-grants-gov/commit/05618510fcda86ca9d62b10acd02c2f313223452

api/src/services/common_grants/transformation.py
DavidDudas-Intuitial: Good catch! Thanks for pointing this out. Fixed in https://github.com/DavidDudas-Intuitial/simpler-grants-gov/commit/05618510fcda86ca9d62b10acd02c2f313223452

api/src/services/common_grants/transformation.py
DavidDudas-Intuitial: Fixed in https://github.com/DavidDudas-Intuitial/simpler-grants-gov/commit/05618510fcda86ca9d62b10acd02c2f313223452

api/src/services/common_grants/transformation.py
DavidDudas-Intuitial: Fixed in https://github.com/DavidDudas-Intuitial/simpler-grants-gov/commit/05618510fcda86ca9d62b10acd02c2f313223452

api/src/services/common_grants/transformation.py
DavidDudas-Intuitial: Done in https://github.com/DavidDudas-Intuitial/simpler-grants-gov/commit/05618510fcda86ca9d62b10acd02c2f313223452

api/src/services/common_grants/transformation.py
DavidDudas-Intuitial: Done in https://github.com/DavidDudas-Intuitial/simpler-grants-gov/commit/05618510fcda86ca9d62b10acd02c2f313223452

api/tests/src/api/common_grants/test_opportunities_routes.py
DavidDudas-Intuitial: Done in https://github.com/DavidDudas-Intuitial/simpler-grants-gov/commit/05618510fcda86ca9d62b10acd02c2f313223452

api/tests/src/services/common_grants/test_opportunity_service.py
DavidDudas-Intuitial: Done in https://github.com/DavidDudas-Intuitial/simpler-grants-gov/commit/05618510fcda86ca9d62b10acd02c2f313223452

api/src/services/common_grants/opportunity_service.py
DavidDudas-Intuitial: Done in https://github.com/DavidDudas-Intuitial/simpler-grants-gov/commit/05618510fcda86ca9d62b10acd02c2f313223452

api/src/services/common_grants/opportunity_service.py
DavidDudas-Intuitial: Done in https://github.com/DavidDudas-Intuitial/simpler-grants-gov/commit/05618510fcda86ca9d62b10acd02c2f313223452

api/src/services/common_grants/opportunity_service.py
DavidDudas-Intuitial: Oops, found more. Will fix those in next commit.

api/src/api/common_grants/common_grants_routes.py
DavidDudas-Intuitial: Fixed in [2f03a8f](https://github.com/DavidDudas-Intuitial/simpler-grants-gov/commit/2f03a8f651f0ffa2069455cfe1788c45f76bbeba)

api/src/api/common_grants/common_grants_routes.py
DavidDudas-Intuitial: @chouinar If I make this change, how can I generate/find a key that I can use in my local env for development? 

For example, I often use curl commands like this for testing: 
`curl -X 'GET' 'http://localhost:8080/common-grants/opportunities' -H 'accept: application/json' -H 'X-Auth: LOCAL_AUTH_12345678'`

Is there a token in `local.env` or elsewhere similar to `LOCAL_AUTH_12345678` that works in a dev env for the `X-API-Key` header? Or a utility to generate a token?

chouinar: Hmm, there would be ways to get a key, but they wouldn't be super convenient at the moment. I can make a quick utility this afternoon to generate a key a bit easier for development. That'll be needed for other work we have coming up anyways.

In the meantime, if you run the local API + frontend together you could login and then go to http://localhost:3000/api-dashboard to generate a key. The username you put in on the mock oauth won't matter for your purposes yet, it'll just make a new user.

chouinar: https://github.com/HHS/simpler-grants-gov/pull/6540 - here's a change that'll add some users with the new API key approach for local development.

DavidDudas-Intuitial: Thank you!

DavidDudas-Intuitial:
Closing this PR to replace it with a cleaned-up PR with changes on a branch rather than a fork. See https://github.com/HHS/simpler-grants-gov/pull/6542

PR #6380, opened 2025-09-12
Update vpc config for ipv6 and  elb security group.

prasnava:
## Summary

Updated the VPC config to disable auto assigning of ipv6 address to eni's within the subnet and updated the service module to allow inbound traffic at the load balancer via ipv6.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -33,6 +33,7 @@
         "newrelic": "^12.7.0",
         "next": "^15.4.7",
         "next-intl": "^3.2.1",
+        "next-navigation-guard": "^0.2.0",
         "pino": "^9.6.0",
         "react": "^19.0.0",
         "react-dom": "^19.0.0",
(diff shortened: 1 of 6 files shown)


infra/api/service/notifications.tf
```diff
@@ -1,29 +1,27 @@
...
-  domain_identity_arn = null
-  #     module.notifications_email_domain[0].domain_identity_arn :
-  #     module.existing_notifications_email_domain[0].domain_identity_arn
-  #   ) : null
-  #   notifications_environment_variables = local.notifications_config != null ? {
-  #     AWS_PINPOINT_APP_ID = module.notifications[0].app_id
-  #   } : {}
-  notifications_environment_variables = {}
-  notifications_app_name              = local.notifications_config != null ? "${local.prefix}${local.notifications_config.name}" : ""
-  #   pinpoint_app_id        = module.notifications[0].app_id
```
mdragon: Can we revert this in its own branch because this will take IPv6 to Prod with the Pinpoint fix right?

PR #6386, opened 2025-09-12
[Issue #6155] Apply conditional logic / business rule enforcement [3/4]

mikehgrantsgov:
## Summary

Work for #6155

## Changes proposed

Add conditional transformers which handle some `if_then_else`, `computed_field`, `required_when`, `one_to_many` that handle some cases we see in JSON.
Add those conditional rules to `revision_type`, `federal_award_identifier`, `debt_explanation`, other fields as needed.
Add tests for conditions types, edge cases, transformations

## Context for reviewers

The SF-424 has some business rules defined in the JSON schema that should be accommodated in the XML transform. We built some transformers and logic to handles these business rules / transformations with added tests. 
After this, I believe we will be 60% or more complete with the task. The final PR will document:
- what's remaining,
- the current process and
- extensions that include file attachments and other complex parts

## Validation steps

See new unit tests.

### DIFF
--- a/api/src/services/xml_generation/transformers/base_transformer.py
+++ b/api/src/services/xml_generation/transformers/base_transformer.py
@@ -85,17 +85,20 @@ class RecursiveXMLTransformer:
                         )
                         continue
                     elif none_handling == "default_value":
-                        # Use configured default value
-                        default_value = transform_rule.get("default_value", "")
+                        # Use configured default value - error if not provided
+                        if "default_value" not in transform_rule:
+                            raise ValueError(
+                                f"null_handling 'default_value' specified but no default_value provided for {'.'.join(current_path)}"
+                            )
+                        default_value = transform_rule["default_value"]
                         source_value = default_value
                         logger.debug(
                             f"Using default value '{default_value}' for {'.'.join(current_path)}"
                         )
                     else:
-                        logger.warning(
-                            f"Unknown null_handling '{none_handling}' for {'.'.join(current_path)}, excluding field"
+                        raise ValueError(
+                            f"Unknown null_handling '{none_handling}' for {'.'.join(current_path)}"
                         )
-                        continue
 
                 # Apply the transformation (conditional transformations can handle None source values)
                 if transform_type == "conditional" or source_value is not None:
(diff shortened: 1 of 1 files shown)


api/src/services/xml_generation/conditional_transformers.py
```diff
@@ -0,0 +1,262 @@
...
+            return search_value in actual_value
+        return False
+
+    elif condition_type == "and":
+        conditions = condition.get("conditions", [])
+        return all(evaluate_condition(cond, source_data) for cond in conditions)
+
+    elif condition_type == "or":
+        conditions = condition.get("conditions", [])
+        return any(evaluate_condition(cond, source_data) for cond in conditions)
```
jakobpederson, line 71: It seems like you could combine the "and" and "or" section here since the code looks identical.

api/src/form_schema/forms/sf424.py
```diff
@@ -1194,6 +1193,107 @@
...
+                "type": "required_when",
+                "condition": {
+                    "type": "field_equals",
+                    "field": "application_type",
+                    "value": "Revision",
+                },
+                "value": {"type": "field_value", "field": "revision_type"},
+            },
+        }
+    },
```
chouinar: There might be some need for conditional transformations, but I don't think it needs to be aligned with our JSON schemas if-then rules. Assuming our data is validated properly, these fields would already be setup properly. There might be a need to exclude fields (eg you said "no" so the field isn't required and we don't want to copy whatever they may have put), but I think most of those are in the SF-LLL.

api/src/form_schema/forms/sf424.py
```diff
@@ -1194,6 +1193,107 @@
...
+                    "fields": [
+                        "federal_estimated_funding",
+                        "applicant_estimated_funding",
+                        "state_estimated_funding",
+                        "local_estimated_funding",
+                        "other_estimated_funding",
+                        "program_income_estimated_funding",
+                    ],
+                },
+            },
```
chouinar: Summing would have been done already in the JSON schema, while there could theoretically be a need to sum in this that we don't have already upstream, I don't think we have any need for this in any existing form.

api/src/services/xml_generation/transformers/base_transformer.py
```diff
@@ -60,11 +65,17 @@ def _process_transform_rules(
...
-                # Get the source value from the input data
-                source_value = self._get_nested_value(source_data, current_path)
+                # Check if this is a conditional transformation that doesn't need source value
+                transform_type = transform_rule.get("type", "simple")
+                if transform_type == "conditional":
+                    # For conditional transformations, always attempt to process
+                    source_value = self._get_nested_value(source_data, current_path)
+                else:
+                    # Get the source value from the input data
+                    source_value = self._get_nested_value(source_data, current_path)
```
chouinar: Should these match?

api/src/services/xml_generation/transformers/base_transformer.py
```diff
@@ -91,18 +102,37 @@ def _process_transform_rules(
...
+                            result[target_field] = transformed_value
+                            logger.debug(
+                                f"Transformed {'.'.join(current_path)} -> {target_field}: {source_value}"
+                            )
+                    else:
+                        target_field = transform_rule["target"]
+                        result[target_field] = transformed_value
+                        logger.debug(
+                            f"Transformed {'.'.join(current_path)} -> {target_field}: {source_value}"
+                        )
```
chouinar: I think these two elses are the same, any way to adjust the if-statements above to consolidate this a bit?

chouinar (approved):
Just some minor if-statement things, LGTM to merge into the main PR.

api/src/services/xml_generation/transformers/base_transformer.py
mikehgrantsgov: Yeah I consolidated this. It was originally split when this PR was more complex, but it was consolidated when we simplified

api/src/services/xml_generation/transformers/base_transformer.py
mikehgrantsgov: Updated this

PR #6407, opened 2025-09-16
[ISSUE #6113] Modify logic where we create organization users to make them the organization owner role

### LINKED ISSUE #6113: Modify logic where we create organization users to make them the organization owner role
### Summary

We need to ensure that when creating organizations, the user created has the organization owner role. We also need to ensure existing users have this role as well.

### Tech Notes
- We need to modify the logic where we create organization users (from [sam.gov](http://sam.gov/) entity data) to give them an organization owner role.
- For now, leave the `is_organization_owner` flag too, we’ll trim that out later.

### Acceptance criteria
- When creating organization users, ensure they have organization owner role.  
- Add a DB migration that automatically adds this for everyone already marked as an owner.
- Add / update tests

joshtonava:
## Summary

Fixes #6113 

## Changes proposed

- Added `organization_user_roles` relationship to the `OrganizationUser` model
- Modified organization user creation logic to automatically assign the Organization Admin role when creating organization users from SAM.gov entity data
-  Updated`sam_gov_utils.py` to include role assignment logic
- Added database migration to assign Organization Admin role to all existing organization owners who don't already have it
- Expanded test coverage to verify that organization owners receive the Organization Admin role in addition to the boolean flag

## Context for reviewers

This PR implements a dual approach to organization ownership - maintaining the existing `is_organization_owner` boolean flag while also assigning the proper Organization Admin role from the role-based access control (RBAC) system.

1. The `_assign_organization_admin_role()` function checks for existing role assignments to avoid db consistency issues.
2 Added `db_session.flush()` after role assignment to ensure the `OrganizationUser` gets its primary key before creating the `OrganizationUserRole` record, preventing foreign key constraint violations.
3. The migration uses `NOT EXISTS` to avoid duplicate role assignments and imports the role ID from `static_role_values.py` to maintain consistency with application code. Using bind params for security.
4. The existing `is_organization_owner` flag is preserved, allowing for a gradual transition to the RBAC system.

## Validation steps

These are the two test files that were affected

- make test args="-vv tests/src/services/users/test_login_gov_callback_handler_ebiz_poc.py"
- make test args="-vv tests/src/services/users/test_organization_from_ebiz_poc.py"


api/src/util/sam_gov_utils.py
```diff
@@ -82,4 +83,35 @@ def link_sam_gov_entity_if_not_exists(
...
+    org_user_role = OrganizationUserRole(
+        organization_user=organization_user, role_id=ORG_ADMIN.role_id
+    )
+    db_session.add(org_user_role)
+    logger.info("Assigned Organization Admin role to organization user", extra=log_extra)
+
+    # Flush to ensure the OrganizationUser gets its primary key (organization_user_id)
+    # before the OrganizationUserRole is created. This prevents foreign key constraint
+    # violations when the role references the organization_user_id.
+    db_session.flush()
```
chouinar: Alternatively you can add the IDs when we create the objects (eg. `organization_user = OrganizationUser(organization_user_id=uuid.uuid4(), ...)`). Sometimes flushes can be a bit iffy.

Setting the IDs would also help with logging.

api/src/util/sam_gov_utils.py
```diff
@@ -82,4 +83,35 @@ def link_sam_gov_entity_if_not_exists(
...
+    """Assign the Organization Admin role to an organization user if they don't already have it."""
+
+    # Check if the user already has the Organization Admin role
+    for role_assignment in organization_user.organization_user_roles:
+        if role_assignment.role_id == ORG_ADMIN.role_id:
+            # User already has the role, nothing to do
+            return
+
+    # User doesn't have the role, assign it
+    increment_metric("new_organization_admin_role_assigned_count")
```
chouinar, line 107: @eric-serrato95 - FYI, the metric we have for tracking users being added as an owner to an organization will be a little iffy while we make this cutover. Currently that's tracked by `new_organization_owner_count` but that's tied to the old `is_owner` flag and will be removed in the next sprint I believe. Afterwards, we'll want to adjust our New Relic dashboards to use this, but it won't have the historical value.

While they both exist, we might be double-counting owners (I suppose we could remove the other metric and just add the old + new together?). I think I also use the log messages though for the API case, so it'll be a bit off regardless. If we want an exact number, it's a simple DB query.

api/src/db/migrations/versions/2025_09_16_assign_organization_admin_role_to_.py
```diff
@@ -0,0 +1,65 @@
...
+        DELETE FROM api.organization_user_role
+        WHERE role_id = '{org_admin_role_id}'::uuid
+        AND organization_user_id IN (
+            SELECT organization_user_id
+            FROM api.organization_user
+            WHERE is_organization_owner = true
+        );
+    """
+    )
+
```
chouinar: I think the downgrade to undo this might be more complex, depending on how long passes for such a rollback, we might need some additional filters. Might be best to just exclude a downgrade and we'd fix it manually.

api/src/util/sam_gov_utils.py
eric-serrato95: @chouinar could we make then 2 separate counts when we make the cutover? having a count of is_owner until the cutover and then add the count under new_organization_owner_count ?
would this cover for this iffy state? (If so I guess we would just have to be cognizant on when it gets merged into prod no?

api/src/util/sam_gov_utils.py
joshtonava: That is reasonable, I can make that change

joshtonava: So far I havent had success with this change but ill try a bit longer

joshtonava: I have been trying to get this to work and I think I am mischaracterizing the issue. 

The save works fine without any foreign key errors, but its the test that fails because the mapped field 

`org_user.organization_user_roles` 

is empty unless I flush.

I have tried explicity setting the organization user id 

```
organization_user = OrganizationUser(
            organization_user_id=uuid.uuid4(),
            organization=organization,
            user=user,
            is_organization_owner=True,
        )
 ```
 
 but in the test 
 
 `org_user.organization_user_roles` is empty

api/src/util/sam_gov_utils.py
chouinar: They are two separate counts, but they'll both be incrementing at the same time until we remove the other way.

chouinar (approved):

PR #6410, opened 2025-09-16
[ISSUE #6114] Modify logic where we create application users to make them the application owner role

### LINKED ISSUE #6114: Modify logic where we create application users to make them the application owner role
### Summary

We need to ensure that when creating applications, the user creating the application is given the application owner role. We also need to ensure existing application owner users have this role as well.

### Tech Notes
- We need to modify logic in create application where we make someone the owner of an application to give them the owner application role.
- For now, leave the `is_application_owner` flag too, we’ll trim that out later.

### Acceptance criteria
- When creating organization users, ensure they have organization owner role.  
- Add a DB migration that automatically adds this for everyone already marked as an owner.
- Add / update tests

joshtonava:
## Summary

Fixes #6114 

## Changes proposed

- Added `application_user_roles` relationship to the `ApplicationUser` model
- Modified application creation logic to automatically assign the Application Owner role when creating applications
- Updated `create_application.py` to include role assignment logic
- Added database migration to assign Application Owner role to all existing application owners who don't already have it
- Expanded test coverage to verify that application owners receive the Application Owner role in addition to the boolean flag

## Context for reviewers

This PR implements a dual approach to application ownership - maintaining the existing `is_application_owner` boolean flag while also assigning the proper Application Owner role from the role-based access control (RBAC) system.

- The `_assign_application_owner_role()` function checks for existing role assignments to avoid database consistency issues.
- Added `db_session.flush()` after role assignment to ensure the ApplicationUser gets its primary key before creating the ApplicationUserRole record, preventing foreign key constraint violations.
- The migration uses `NOT EXISTS` to avoid duplicate role assignments and imports the role ID from `static_role_values.py`. Using sqlalchemy text with role_id replacement for security.
- The existing `is_application_owner` flag is preserved, allowing for a gradual transition to the RBAC system.
-  Analysis confirmed that `create_application.py` is the only service location where `is_application_owner=True` is set

## Validation steps

This is the main test file that was affected:

make test args="-vv tests/src/services/applications/test_create_application.py"

### DIFF
--- a/api/src/db/models/user_models.py
+++ b/api/src/db/models/user_models.py
@@ -238,13 +238,6 @@ class ApplicationUser(ApiSchemaTable, TimestampMixin):
     def roles(self) -> list["Role"]:
         return [aur.role for aur in self.application_user_roles]
 
-    application_user_roles: Mapped[list["ApplicationUserRole"]] = relationship(
-        "ApplicationUserRole",
-        back_populates="application_user",
-        uselist=True,
-        cascade="all, delete-orphan",
-    )
-
 
 class OrganizationUser(ApiSchemaTable, TimestampMixin):
     __tablename__ = "organization_user"


api/src/services/applications/create_application.py
```diff
@@ -26,6 +27,36 @@
...
+def _assign_application_owner_role(
+    db_session: db.Session, application_user: ApplicationUser
+) -> None:
+    """Assign the Application Owner role to an application user if they don't already have it."""
+
+    # Check if the user already has the Application Owner role
+    for role_assignment in application_user.application_user_roles:
+        if role_assignment.role_id == APPLICATION_OWNER.role_id:
+            # User already has the role, nothing to do
+            return
```
chouinar: Since we only do this particular logic when an application is first created, we can skip this. We just need the two lines of creating the ApplicationUserRole + db_session.add

api/src/services/applications/create_application.py
```diff
@@ -26,6 +27,36 @@
...
+        extra={
+            "application_id": application_user.application_id,
+            "user_id": application_user.user_id,
+        },
+    )
+
+    # Flush to ensure the ApplicationUser gets its primary key (application_user_id)
+    # before the ApplicationUserRole is created. This prevents foreign key constraint
+    # violations when the role references the application_user_id.
+    db_session.flush()
```
chouinar: Similar to the other PR, we could also set the ID directly on the objects we create and avoid a flush.

joshtonava: I removed this. I am doing the tests different anyway so its not necessary to get them to pass

joshtonava:
@chouinar I updated the migration heads so we dont get a conflict after merging

PR #6420, opened 2025-09-17
fix: applies errors to each section

myduong-navapbc:
## Summary

Fixes / Work for #ISSUE

### DIFF
--- a/frontend/src/components/applyForm/widgets/budget/Budget424aSectionC.tsx
+++ b/frontend/src/components/applyForm/widgets/budget/Budget424aSectionC.tsx
@@ -2,7 +2,7 @@
 import { FormContextType, RJSFSchema, StrictRJSFSchema } from "@rjsf/utils";
 import { get } from "lodash";
 
-import React, { JSX } from "react";
+import React, { JSX, useEffect } from "react";
 import { Table } from "@trussworks/react-uswds";
 
 import {
@@ -136,7 +136,9 @@ function Budget424aSectionC<
 
   const ROWS = BUDGET_ACTIVITY_COLUMNS;
 
-  const getErrors = getBudgetErrors({ errors, id, section: "C" });
+  function getErrorMessagesForField(fieldId: string): string[] {
+    return getBudgetErrors({ errors, id: fieldId, section: "C" });
+  }
 
   // column labels for the money fields
   const fields: { key: AmountKey; label: string }[] = [
@@ -153,6 +155,10 @@ function Budget424aSectionC<
         activityItems,
         `[${rowIndex}].assistance_listing_number`,
       ) ?? "";
+
+      useEffect(() => {
+        console.log("Section C: ", errors, rawErrors)
+      }), [errors, rawErrors]
 
     return (
       <div className="display-flex flex-column">
@@ -174,7 +180,7 @@ function Budget424aSectionC<
     return (
       <CurrencyInput
         id={idPath}
-        rawErrors={getErrors}
+        rawErrors={getErrorMessagesForField(idPath)}
         value={get(
           activityItems,
           `[${rowIndex}].non_federal_resources.${fieldKey}`,
@@ -190,7 +196,6 @@ function Budget424aSectionC<
         <HelperText>Sum of row {rowIndex + 8}</HelperText>
         <CurrencyInput
           id={idPath}
-          rawErrors={getErrors}
           value={get(
             activityItems,
             `[${rowIndex}].non_federal_resources.total_amount`,
@@ -212,7 +217,6 @@ function Budget424aSectionC<
         </HelperText>
         <CurrencyInput
           id={idPath}
-          rawErrors={getErrors}
           value={totals ? totals[fieldKey] : undefined}
           bordered
         />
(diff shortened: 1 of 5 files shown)


frontend/src/components/applyForm/widgets/budget/Budget424aSectionC.tsx
```diff
@@ -190,7 +192,6 @@ function Budget424aSectionC<
         <HelperText>Sum of row {rowIndex + 8}</HelperText>
         <CurrencyInput
           id={idPath}
-          rawErrors={getErrors}
```
doug-s-nava, line 193: is it safe to assume these fields will never have errors?

frontend/src/components/applyForm/widgets/budget/Budget424aSectionE.tsx
```diff
@@ -114,6 +116,10 @@ function Budget424aSectionE<
   const errors = (rawErrors as FormValidationWarning[]) || [];
   const { items: activityItems, totals } = normalizeSectionEValue(rawValue);
 
+  function getErrorMessagesForField(fieldId: string): string[] {
```
doug-s-nava, line 119: if we make this change this function can be more reusable

```suggestion
  const getErrorMessagesForFieldAndSection(id: string, section: string) => (errors) => string[] {
        return getBudgetErrors({ errors, id, section });
   }
```

frontend/src/components/applyForm/widgets/budget/Budget424aSectionF.tsx
```diff
@@ -66,8 +102,30 @@ function Budget424aSectionF<
   const remarksSchema = properties?.remarks as RJSFSchema | undefined;
   const confirmationSchema = properties?.confirmation as RJSFSchema | undefined;
 
-  const getFieldErrors = (fieldId: string) =>
-    getBudgetErrors({ errors: validationWarnings, id: fieldId, section: "F" });
+  function getWarningsFromContext(context: unknown): FormValidationWarning[] {
```
doug-s-nava, line 105: this seems like it can be split out from the render

frontend/src/components/applyForm/widgets/budget/Budget424aSectionF.tsx
```diff
@@ -25,6 +27,41 @@ function getRootSchemaFromContext(context: unknown): RJSFSchema | undefined {
   return undefined;
 }
 
+type WarningsFromContext =
+  | FormValidationWarning[]
+  | FormattedFormValidationWarning[]
+  | [];
+
+function coerceFormValidationWarnings(
```
doug-s-nava, line 35: looks like this is necessary because sometimes the errors are coming through formatted using the new buildWarningTree, and some are coming in more raw. That's strange though, as any errors for the budget form should not be going through buildWarningTree. 

Or is it that sometimes the warnings are not being properly passed as props? Any idea why this is happening? It may make sense to get to the bottom of that rather than putting on this bandaid, unless it's a tough fix

frontend/src/components/applyForm/widgets/budget/Budget424aSectionC.tsx
myduong-navapbc: yes, as soon as the auto-summation these fields will be disabled and would never get errors

frontend/src/components/applyForm/widgets/budget/Budget424aSectionE.tsx
myduong-navapbc: I'll add this in a new PR

frontend/src/components/applyForm/widgets/budget/Budget424aSectionF.tsx
myduong-navapbc: I believe the two shapes come from two different data paths (as far as i understand it, and please correct me if i made a mistake or misunderstood):

Non-budget forms:
warnings flow through buildWarningTree -> FormattedFormValidationWarning  (htmlField / formatted).

Budget forms:
warnings are produced “raw” by the budget validator -> FormValidationWarning (field / message) and were bypassed by buildWarningTree.

the coerceFormValidationWarnings is a bit of a shim (currently) can handle the mix of the two BBBUUUT i think the long term-ish answer would be something kind of like this:

```
function toRawWarnings(
  input: FormattedFormValidationWarning[] | FormValidationWarning[] | null,
): FormValidationWarning[] {
  if (!Array.isArray(input)) return [];
  return input.map((item) => {
    if ("field" in item && typeof item.field === "string") {
      return { field: item.field, message: item.message, type: item.type ?? "custom", value: item.value ?? null };
    }
    if ("htmlField" in item && typeof item.htmlField === "string") {
      return { field: item.htmlField, message: item.formatted ?? item.message, type: "custom", value: null };
    }
    return { field: "", message: "", type: "custom", value: null };
  }).filter(w => w.field && w.message);
}
```

where we, unify the warnings/errors in a single normalization boundary in ApplyForm (this is just me thinking about it and not fully fleshed out the details).

My overall idea or thought behind this is:
All sections should receive validation warnings in the same way, through formContext, instead of some sections relying on props and others not.

hopefully that wasnt too much rambling

Issue #5511, opened 2025-06-28
Upgrade Postgres RDSes to 17.5


mdragon:
### Summary

We're on PG 15.10 in RDS. That loses AWS Support in March 2026. 17.5 only has support through Sept 2026, but once we're on 17 we can keep up with minor revisions more readily.

### Acceptance criteria

- [ ] Plan the upgrade
- [ ] Run the upgrade in Dev
- [ ] Test the upgrade
- [ ] Create follow on tickets for applying the upgrade to Staging and Prod (and Training if it exists yet)

prasnava:
Update: All the lower environment's database's have been successfully upgraded to postgres version: 17.5. Will create a separate ticket for the prod database upgrade.

PR #6423, opened 2025-09-17
[Issue #5511] Upgrade Postgres from v15.10 to v17.5

prasnava:
## Summary

Upgrade postgres from version 15.10 to version 17.5.


prasnava:
Let's keep this pr open for a few days so we can observe the behavior in the lower environments.

PR #6425, opened 2025-09-17
[Issue #6404] Fix prepopulation when value is null

### LINKED ISSUE #6404: [BUG] Pre-population of "null" as value causing issues
### Summary

Slack chat for reference: [link](https://betagrantsgov.slack.com/archives/C08NCEXRSPJ/p1758033382127669)
This is not a blocker for pilot launch since the pilot does not contain any "null" fields.

**Issue:** Let’s assume we pre-population the competition number, that comes from the competition table and that field is nullable.

So when we populate it into the JSON we get something like:
```
{
 "competition_id": null
}
```
But we don’t let any fields in the JSON schema be null, they can be missing, but not null (technically could allow that for non-required fields, that would maybe work, but not sure if it breaks things for the frontend).

### Proposed solution
What we want to do is twofold:
- If the value is null, don’t populate it (so the data would be just {}) - that’s easy.
- If the value already exists, delete it - this is necessary if someone populated it like an API user could potentially do. This is trickier and needs some rework to the underlying population logic.

### Acceptance criteria

- [ ] Solution implemented per above guidance and "null" fields no longer create issues.

chouinar:
## Summary
Fixes  #6404

## Changes proposed
Adjust behavior of prepopulation when a value is null, by default the field should not be populated AND any existing value is removed.

Adjust some of our prepopulation logic to not let a field be null (opportunity number, opportunity title, and agency name) to avoid issues where a required field can't be populated

Add some additional tests to the SF424 validation pre-population behavior

## Context for reviewers
Some prepopulation fields can be null as the field is not required, for example the competition title is a commonly nullable field. But our JSON schema doesn't like those fields being null, even if they're not required. Null != missing, missing is fine, null would require we change the type from `string` to `["string", "null"]` which is a much bigger change. Instead we'll change the default to not include the field. Additionally, it will delete an existing value, this is because an API user could pass in any value for it, and we want to make sure it's unset (there are a few very specific edge cases still unhandled, but not needed any time soon).

For the fields I made not be able to prepopulate to null, they do have nullable columns but in prod opportunity number and title are never actually null. Agency is only null on about 40 opportunities, most of those from before 2010. We should eventually fix the schema to not let those fields be null, but that's not something we should worry about yet.

### DIFF
--- a/api/src/form_schema/rule_processing/README.md
+++ b/api/src/form_schema/rule_processing/README.md
@@ -101,6 +101,36 @@ rules currently implemented:
 * `current_date` - The current date in the US EST timezone
 * `signature` - The name of the creator of the application - NOTE - currently defaults to a random application user's email due to not having names in our system yet.
 
+## Null Population
+Several of our rules like `public_competition_id` and `competition_title` can be
+null due to them being nullable in their corresponding DB tables. We don't generally
+configure our JSON schema to allow null values, instead preferring a value to be excluded
+if there is no value (eg. We want `{}` over `{"my_field": null}`). However, if the rule
+returns null, we have to do something. For that reason, all pre/post population rules
+can take in an optional `null_population` rule with the following values:
+* `exclude_value` (the default if no rule configured) - This will not add the value to the JSON schema AND remove any existing values.
+* `set_as_null` - This will set the value as null, behaving the same as any other population
+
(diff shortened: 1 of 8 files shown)


mikehgrantsgov (approved):
Looks good, and very similar to some of the XML stuff!

Issue #6116, opened 2025-08-26
Create a utility for enforcing privileges


joshtonava:
### Summary

With the groundwork in place to model user roles and privileges, and access to specific organizations and applications given, we now need to be able to put restrictions on specific endpoints for their usage.

### Tech Notes

See tech spec for reference: [Tech Spec: User Roles on the Backend | Enforcement](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/1977221432/Tech+Spec+User+Roles+on+the+Backend#Enforcement)

### Acceptance criteria

- Implement a method for enforcing access to a resource (organization, application) by user
- Add / Update tests

eric-serrato95:
will require a ton of tests - this is the core of our logic and will require a bulk of testing work/effort

PR #6426, opened 2025-09-17
[Issue #6116] Access endpoint utility

babebe:
## Summary

Fixes for #6116  

## Changes proposed
Utility added for determining access to resources (Application | Agency | Organization)
Added Privileges for Agency and Internal roles
Updated relationships between users and roles
Added test


api/src/auth/endpoint_access_util.py
```diff
@@ -0,0 +1,78 @@
...
+    resource: Organization | Application | Agency | None,
+) -> bool:
+    roles = get_roles_for_resource(user, resource)
+
+    for role in roles:
+        privilege_overlap = allowed_privileges.intersection(role.privileges)
+        if len(privilege_overlap) > 0:
+            logger.info(
+                "Access granted",
+                extra={"user_id": user.user_id, "role_id": role.role_id, "resource": resource},
```
chouinar: For putting info about the resource, we probably want a utility function to grab info, this would just log some SQLAlchemy memory object most of the time.

Roughly like:
```py
def get_log_info_for_resource(resource: Organization | Application | Agency | None) -> dict:
     log_info = {}
     if resource is None:
         return log_info
    
    if isinstance(resource, Organization):
        log_info["resource_type"] = "organization
        log_info["organization_id"] = resource.organization_id
    elif isinstance(resource, Application):
        log_info["resource_type"] = "organization
        log_info["application_id"] = resource.application_id
    elif isinstance(resource, Agency):
        log_info["resource_type"] = "organization
        log_info["agency_id"] = resource.agency_id
        log_info["agency_code"] = resource.agency_code

     return log_info
```

api/src/auth/endpoint_access_util.py
```diff
@@ -0,0 +1,78 @@
...
+            )
+            return True
+
+    logger.warning("Access denied", extra={"user_id": user.user_id, "resource": resource})
```
chouinar: A warning indicates some sort of minor issue in our system, someone trying to access something they don't have permissions to will be fairly routine/uneventful. We can make this an info log
```suggestion
    logger.info("Access denied", extra={"user_id": user.user_id, "resource": resource})
```

api/tests/src/auth/test_endpoint_access_util.py
```diff
@@ -0,0 +1,278 @@
...
+        assert len(result) == 1
+        assert result[0].role_id == internal_user_role.role_id
+
+    def test_get_internal_role_access(self, internal_role_privileges, app):
+        """User has required internal privilege; should be returned"""
+        internal_user_role = InternalUserRoleFactory(
+            role=internal_role_privileges,
+        )
+
+        can_access(internal_user_role.user, {INTERNAL_PRIVILEGES[0]}, app)
```
chouinar: I think an assert is missing here

api/tests/src/auth/test_endpoint_access_util.py
```diff
@@ -0,0 +1,278 @@
...
+    def app_role_privileges(self):
+        return RoleFactory(privileges=APP_PRIVILEGES)
+
+    @pytest.fixture(scope="class")
+    def internal_role_privileges(self):
+        return RoleFactory(privileges=INTERNAL_PRIVILEGES)
+
+    @pytest.fixture(scope="class")
+    def agency_role_privileges(self):
+        return RoleFactory(privileges=AGENCY_PRIVILEGES)
```
chouinar, line 91: Could we add some tests where a user doesn't have all privileges? For example, a user in org A trying to access application A for a specific privilege they don't have. Don't need to check every possible value, but just verify that they have privilege X and Y and can access that app for X/Y, but they don't have privilege Z and that gets denied as expected for Z.

babebe: Added more of these

api/src/auth/endpoint_access_util.py
babebe: Adapted

api/src/auth/endpoint_access_util.py
```diff
@@ -0,0 +1,104 @@
...
+        roles.extend(get_roles_for_org(user, resource))
+    elif isinstance(resource, Application):
+        roles.extend(get_roles_for_app(user, resource))
+    elif isinstance(resource, Agency):
+        roles.extend(get_roles_for_agency(user, resource))
+
+    return roles
+
+
+def can_access(
```
chouinar, line 61: An idea for the following tickets where we'll add usage of this function, we might want a function that wraps this like:

```py
def validate_privileges(...same params):
     if not can_access(...):
          raise_flask_error(403, "Access Denied - Missing required privilege") # maybe a validation error added.
```
That way we don't need to copy-paste that `raise_flask_error` to a bunch of places. Not needed yet, but an idea for the next tickets.

chouinar (approved):
LGTM!

Issue #6359, opened 2025-09-09
If it is Simpler format, skip sending request to Grants.gov


chris-kuryak:
### Summary

If it is Simpler format, skip sending request to Grants.gov.
- Endpoint/Operation: GetApplicatonZip

May need to change the way we handle the Simple boolean, since it may default to false.

This will help simplify the response and ensure that the user does not need to wait on responses from both systems, if not necessary.

### Acceptance criteria

- [x] Simpler formatted tracking IDs do not send request to Grants.gov

PR #6434, opened 2025-09-18
[Issue #6359] Skip sending to grants.gov if simpler formatted id

jakobpederson:
## Summary

If it is Simpler format, skip sending request to Grants.gov

## Changes proposed

Added a check to see if the incoming request meets the following criteria:
- It is either GetApplicationZip or GetApplication
- It has a GrantsGovTrackingNumber that starts with an 8 or a 9

If both those critera are met, do not forward the request on to grants.gov. 

## Context for reviewers

Hitting grants.gov can be slow, so if we know for sure the data is not in grants.gov we won't bother actually making the request.

### DIFF
--- a/api/src/legacy_soap_api/legacy_soap_api_routes.py
+++ b/api/src/legacy_soap_api/legacy_soap_api_routes.py
@@ -10,6 +10,7 @@ from src.legacy_soap_api.legacy_soap_api_constants import LegacySoapApiEvent
 from src.legacy_soap_api.legacy_soap_api_proxy import get_proxy_response
 from src.legacy_soap_api.legacy_soap_api_schemas import SimplerSoapAPI, SOAPRequest
 from src.legacy_soap_api.legacy_soap_api_utils import (
+    get_alternate_proxy_response,
     get_invalid_path_response,
     get_soap_error_response,
 )
@@ -60,7 +61,10 @@ def simpler_soap_api_route(
             auth=get_soap_auth(request.headers.get(MTLS_CERT_HEADER_KEY)),
             operation_name=operation_name,
         )
-        soap_proxy_response = get_proxy_response(soap_request)
+        if alternate_proxy_response := get_alternate_proxy_response(soap_request):
+            soap_proxy_response = alternate_proxy_response
+        else:
+            soap_proxy_response = get_proxy_response(soap_request)
     except Exception:
         logger.exception(
             msg="Error getting soap proxy response",
--- a/api/src/legacy_soap_api/legacy_soap_api_utils.py
+++ b/api/src/legacy_soap_api/legacy_soap_api_utils.py
@@ -1,13 +1,17 @@
+import io
 import json
 import logging
 import uuid
+from enum import StrEnum
 from typing import Any, Callable
 
 import requests
 from defusedxml import minidom
+from lxml import etree
 
 from src.legacy_soap_api.legacy_soap_api_config import get_soap_config
-from src.legacy_soap_api.legacy_soap_api_schemas import FaultMessage, SOAPResponse
+from src.legacy_soap_api.legacy_soap_api_schemas import FaultMessage, SOAPRequest, SOAPResponse
+from src.legacy_soap_api.soap_payload_handler import extract_soap_xml
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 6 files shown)


api/src/legacy_soap_api/legacy_soap_api_routes.py
```diff
@@ -60,7 +60,10 @@ def simpler_soap_api_route(
             auth=get_soap_auth(request.headers.get(MTLS_CERT_HEADER_KEY)),
             operation_name=operation_name,
         )
-        soap_proxy_response = get_proxy_response(soap_request)
+        if skip_soap_proxy(soap_request.operation_name, soap_request.data):
```
jakobpederson: Putting "500" here for the status code because that's the status code for not found from the soap proxy

api/tests/src/legacy_soap_api/test_legacy_soap_api_routes.py
```diff
@@ -43,3 +58,156 @@ def test_invalid_xml_server_error_500(client) -> None:
...
+        "<soap:Body><soap:Fault>"
+        "<faultcode>soap:Server</faultcode>"
+        f"<faultstring>Failed to get application zip.(Grant Application not found for tracking number:{tracking_number.text})"
+        "</faultstring></soap:Fault></soap:Body></soap:Envelope>\r\n"
+        f"--uuid:{test_uuid}--"
+    ).encode("utf-8")
+    mock_get_soap_response.assert_not_called()
+    assert response.status_code == 500
+    assert expected == response.data
+    assert response.headers["Content-Length"] == "527"
```
jakobpederson, line 91: Got this content-length from using Postman.

api/src/legacy_soap_api/legacy_soap_api_routes.py
```diff
@@ -60,7 +64,15 @@ def simpler_soap_api_route(
             auth=get_soap_auth(request.headers.get(MTLS_CERT_HEADER_KEY)),
             operation_name=operation_name,
         )
-        soap_proxy_response = get_proxy_response(soap_request)
+        if tracking_number := get_simpler_grants_gov_tracking_number(
+            soap_request.operation_name, soap_request.data
+        ):
+            is_get_application_zip = operation_name == "GetApplicationZipRequest"
```
chouinar: I think we'd want to make this a bit more generic.

What if we had a function like `get_alternate_response` (feel free to come up with a better name) that we pass info into and if it returns something, we use that, otherwise we make the actual call.

Roughly using it like:
```py
alternate_response = get_alternate_response(soap_request)
if alternate_response is not None:
    soap_proxy_response = alternate_response
else:
    soap_proxy_response = get_proxy_response(soap_request)
```

Then we can put any operation specific logic. I would suggest we flip the logic and check the operation before we try to parse something out of it.

api/src/legacy_soap_api/legacy_soap_api_utils.py
```diff
@@ -17,18 +17,18 @@
...
         f"""
---uuid:{response_id}
-Content-Type: application/xop+xml; charset=UTF-8; type=\"text/xml\"
-Content-Transfer-Encoding: binary
+--uuid:{response_id}\r
+Content-Type: application/xop+xml; charset=UTF-8; type=\"text/xml\"\r
+Content-Transfer-Encoding: binary\r
 Content-ID: <root.message@cxf.apache.org>{response_data.decode()}
---uuid:{response_id}--
+--uuid:{response_id}--\r\n
```
chouinar: Could we add a note about the `/r/n` being the windows-style new lines?

---

If we wanted to put `/r/n` on each of these lines instead of having the `\n` come from the actual line breaks, we could do a slightly different way of defining a string. Instead of:
```py
my_str = """my\r
multi\r
line\r
string\r
"""
```
You can also use string concatenation:
```py
my_str = (
"my\r\n" # NOTE - no comma or +, Python will know to concatenate this
"multi\r\n"
"line\r\n"
"string\r\n"
)
```
Or use join: `\r\n".join(["my", "multi", "line", "string", ""])`

api/src/legacy_soap_api/legacy_soap_api_utils.py
```diff
@@ -37,6 +37,35 @@ def format_local_soap_response(response_data: bytes) -> bytes:
...
+        "Content-Type": (
+            "multipart/related;"
+            ' type="application/xop+xml";'
+            f' boundary="uuid:{boundary_id}";'
+            ' start="<root.message@cxf.apache.org>";'
+            ' start-info="text/xml"'
+        ),
+    }
+    if cookie := headers.get("Cookie"):
+        response_headers["Set-Cookie"] = f"{cookie}; Path=/grantsws-agency; Secure; HttpOnly"
```
chouinar: What does this cookie do?

api/src/legacy_soap_api/soap_payload_handler.py
```diff
@@ -228,6 +228,28 @@ def extract_soap_xml(soap_bytes: bytes) -> bytes:
     return b""
 
 
+def get_simpler_grants_gov_tracking_number(
+    operation_name: str, soap_xml: str | bytes
+) -> str | None:
+    if operation_name.lower() not in ["getapplicationziprequest", "getapplicationrequest"]:
+        return None
```
chouinar: Not specifically a problem here, but for operations, we should really have a defined list of them somewhere rather than redefining the strings. Just to avoid a case where I accidentally type `getaplicationziprequest` (missing a `p`) somewhere and can't figure it out. Maybe a StrEnum?

---

Are operation names case-sensitive? If I were to call the soap API with `getAPPLICATIONzipREQUEST` as the operation, would it work? If not, I think we should leave them in their ordinary case.

api/src/legacy_soap_api/soap_payload_handler.py
```diff
@@ -228,6 +228,28 @@ def extract_soap_xml(soap_bytes: bytes) -> bytes:
...
+    xml_file = io.BytesIO(xml_bytes)
+    value = None
+    try:
+        for event, elem in etree.iterparse(xml_file, events=("end",)):
+            if elem.tag.endswith("GrantsGovTrackingNumber") and event == "end":
+                value = elem.text
+                elem.clear()
+    except etree.XMLSyntaxError:
+        return None
+    return value if value and value.startswith("GRANT8") else None
```
chouinar: Could we do tracking numbers that start with an 8 or 9? While we plan to have the sequence start at 8000000, we might have cases where we manually set up test data. Easiest way to avoid conflicting with the counter is to not start with an 8, so we could start with a 9.

```suggestion
    if value.startswith("GRANT8") or value.startswith("GRANT9"):
         return value
    return None
```

api/src/legacy_soap_api/legacy_soap_api_utils.py
jakobpederson: The cookie showed up in Postman testing and SoapUI, and it looks like if a cookie came in on the request it was added to the response as Set-Cookie. I wanted to preserve that behavior as closely as I could.

api/src/legacy_soap_api/soap_payload_handler.py
jakobpederson: That makese sense, I'll leverage a StrEnum.

api/src/legacy_soap_api/soap_payload_handler.py
jakobpederson: Updating this to include 8 or 9

api/src/legacy_soap_api/legacy_soap_api_utils.py
jakobpederson: I included a note and a date, because I like to indicate that on notes. Let me know if I should take that out.

api/src/legacy_soap_api/legacy_soap_api_utils.py
chouinar: Is it present on other operations / should we be setting it on all responses by default?

api/src/legacy_soap_api/legacy_soap_api_utils.py
```diff
@@ -262,3 +304,37 @@ def log_local(msg: str, data: Any | None = None, formatter: Callable | None = No
...
+            else None
+        )
+        if simpler_tracking_number:
+            is_zip = soap_request.operation_name.startswith(
+                AlternateSoapOperation.GET_APPLICATION_ZIP
+            )
+            return get_soap_proxy_grant_application_not_found_response(
+                simpler_tracking_number, headers=soap_request.headers, is_get_application_zip=is_zip
+            )
+    return None
```
chouinar: I think we could shrink this a bit like so by short-circuiting the logic.
```suggestion
    xml_bytes = extract_soap_xml(soap_request.data)
    if not xml_bytes:
        return None

    # Could make this one if-statement, but I think this is fairly readable
    # to say we only need to get the tracking number for a certain operation
    # And I think we'll want to do this on other operations eventually, so just need to change it to
    # "if soap_request.operation_name in [...]"
    if soap_request.operation_name == AlternateSoapOperation.GET_APPLICATION_ZIP:
        tracking_number = get_gov_grants_tracking_number(xml_bytes)

        if tracking_number.startswith("GRANT8") or tracking_number.startswith("GRANT9"):
            return get_soap_proxy_grant_application_not_found_response(tracking_number, headers=soap_request.headers, is_get_application_zip=is_zip)

    return None
```

While the `:=` notation can sometimes make things simpler, if there is a simple case that lets you skip the if-statement, just exit early.

api/src/legacy_soap_api/legacy_soap_api_utils.py
```diff
@@ -17,26 +21,64 @@
...
+    GET_APPLICATION_ZIP = "GetApplicationZipRequest"
+    GET_APPLICATION = "GetApplicationRequest"
+
+
+def format_local_soap_response(response_data: bytes, boundary_id: str | None = None) -> bytes:
     # This is a format string for formatting local responses from the mock
     # soap server since it does not support manipulating the response.
     # The grants.gov SOAP API currently includes this data.
-    response_id = str(uuid.uuid4())
+    # (09/23/25) Note: /r/n is how Windows encodes a newline
```
chouinar: Unless the date matters (ie. you expect things to change, or something is temporary), can leave dates out of comments. I assume Windows isn't going to change this behavior anytime soon.

```suggestion
    # Note: /r/n is how Windows encodes a newline
```

api/src/legacy_soap_api/legacy_soap_api_utils.py
jakobpederson: I'm currently accomodating for GET_APPLICATION_ZIP and GET_APPLICATION, so I was keeping the `if soap_request.operation_name in AlternateSoapOperation`. I do like your changes, it does read a lot better.

api/src/legacy_soap_api/legacy_soap_api_utils.py
jakobpederson: removed

api/src/legacy_soap_api/legacy_soap_api_utils.py
jakobpederson: I've seen it on all requests that aren't tests or hitting the localhost. It looks like it's basically checking if it's a test or not. I'll just add it in by default.

chouinar (approved):

Issue #6314, opened 2025-09-08
Develop Configuration files for existing forms and build the header XML


chris-kuryak:
### Summary

TBD

### Acceptance criteria

TBD

Issue #6313, opened 2025-09-08
Set proper namespace for fields that differ from the form namespace


chris-kuryak:
### Summary

We want to be able to configure how the XML namespaces get defined in our JSON->XML mapping logic. While most fields within a given form will have a namespace defined at the form level (eg. for the SF424 fields will be like `<SF424_4_0:SubmissionType>Application</SF424_4_0:SubmissionType>`), some fields that are defined in a different XSD file require a different namespace.

For example, the address fields should instead look like:
```xml
    <SF424_4_0:Applicant>
        <globLib:Street1
            xmlns:globLib="http://apply.grants.gov/system/GlobalLibrary-V2.0">123 Main St
        </globLib:Street1>
        <globLib:Street2
            xmlns:globLib="http://apply.grants.gov/system/GlobalLibrary-V2.0">Apt 123
        </globLib:Street2>
        <globLib:City
            xmlns:globLib="http://apply.grants.gov/system/GlobalLibrary-V2.0">Exampleburg
        </globLib:City>
        <globLib:County
            xmlns:globLib="http://apply.grants.gov/system/GlobalLibrary-V2.0">Example County
        </globLib:County>
        <globLib:State
            xmlns:globLib="http://apply.grants.gov/system/GlobalLibrary-V2.0">NY: New York
        </globLib:State>
        <globLib:ZipPostalCode
            xmlns:globLib="http://apply.grants.gov/system/GlobalLibrary-V2.0">10001-1000
        </globLib:ZipPostalCode>
        <globLib:Country
            xmlns:globLib="http://apply.grants.gov/system/GlobalLibrary-V2.0">USA: UNITED STATES
        </globLib:Country>
    </SF424_4_0:Applicant>
```

We want a way to configure a field in our mapping configuration so that we can say to use a different namespace than the default + add the `xmlns` attribute.

### Acceptance criteria
- [ ] Logic updated to support altering the namespace for specific fields
- [ ] Tests added

PR #6435, opened 2025-09-18
[Issue #6313] Implement configurable XML namespaces [DRAFT]

mikehgrantsgov:
## Summary

Fixes / Work for #6313 

## Changes proposed

Added `lxml` for proper namespace handling
Added `globLib` namespace definition to _xml_config.namespaces
Update / add tests.

## Context for reviewers

We want to be able to configure how the XML namespaces get defined in our JSON->XML mapping logic. While most fields within a given form will have a namespace defined at the form level (eg. for the SF424 fields will be like <SF424_4_0:SubmissionType>Application</SF424_4_0:SubmissionType>), some fields that are defined in a different XSD file require a different namespace.

Configurable XSD-driven testing and validating is coming in a subsequent PR for #6314 

## Validation steps

See updated unit tests.

### DIFF
--- a/api/src/form_schema/forms/sf424.py
+++ b/api/src/form_schema/forms/sf424.py
@@ -1066,7 +1066,10 @@ FORM_XML_TRANSFORM_RULES = {
         "description": "XML transformation rules for converting Simpler SF-424 JSON to Grants.gov XML format",
         "version": "1.0",
         "form_name": "SF424_4_0",
-        "namespaces": {"default": "http://apply.grants.gov/forms/SF424_4_0-V4.0", "prefix": ""},
+        "namespaces": {
+            "default": "http://apply.grants.gov/forms/SF424_4_0-V4.0",
+            "globLib": "http://apply.grants.gov/system/GlobalLibrary-V2.0",
+        },
         "xml_structure": {"root_element": "SF424_4_0", "version": "4.0"},
         "null_handling_options": {
             "exclude": "Default - exclude field entirely from XML (recommended)",
@@ -1092,13 +1095,13 @@ FORM_XML_TRANSFORM_RULES = {
     # Address information - nested structure
     "applicant_address": {
         "xml_transform": {"target": "Applicant", "type": "nested_object"},
-        "address_line_1": {"xml_transform": {"target": "Street1"}},
-        "address_line_2": {"xml_transform": {"target": "Street2"}},
-        "city": {"xml_transform": {"target": "City"}},
-        "county": {"xml_transform": {"target": "County"}},
-        "state_code": {"xml_transform": {"target": "State"}},
-        "country_code": {"xml_transform": {"target": "Country"}},
-        "zip_code": {"xml_transform": {"target": "ZipPostalCode"}},
+        "street1": {"xml_transform": {"target": "Street1", "namespace": "globLib"}},
+        "street2": {"xml_transform": {"target": "Street2", "namespace": "globLib"}},
+        "city": {"xml_transform": {"target": "City", "namespace": "globLib"}},
+        "county": {"xml_transform": {"target": "County", "namespace": "globLib"}},
+        "state": {"xml_transform": {"target": "State", "namespace": "globLib"}},
+        "country": {"xml_transform": {"target": "Country", "namespace": "globLib"}},
+        "zip_postal_code": {"xml_transform": {"target": "ZipPostalCode", "namespace": "globLib"}},
     },
     # Contact information - direct field mappings
     "phone_number": {"xml_transform": {"target": "PhoneNumber"}},
@@ -1183,6 +1186,26 @@ FORM_XML_TRANSFORM_RULES = {
             "target": "CertificationAgree",
             "value_transform": {"type": "boolean_to_yes_no"},
         }
+    },
+    # Authorized representative - nested structure for name fields
+    "authorized_representative": {
+        "xml_transform": {"target": "AuthorizedRepresentative", "type": "nested_object"},
+        "first_name": {
+            "xml_transform": {
+                "target": "FirstName",
+                "namespace": "globLib",
+                "null_handling": "default_value",
+                "default_value": "John",
+            }
+        },
+        "last_name": {
+            "xml_transform": {
+                "target": "LastName",
+                "namespace": "globLib",
+                "null_handling": "default_value",
+                "default_value": "Doe",
+            }
+        },
     },
     # Authorized representative - direct field mappings
     "authorized_representative_title": {
--- a/api/src/services/xml_generation/service.py
+++ b/api/src/services/xml_generation/service.py
@@ -4,6 +4,8 @@ import logging
 import xml.etree.ElementTree as ET
 from typing import Any
 
+from lxml import etree as lxml_etree
+
 from .config import load_xml_transform_config
 from .models import XMLGenerationRequest, XMLGenerationResponse
 from .transformers.base_transformer import RecursiveXMLTransformer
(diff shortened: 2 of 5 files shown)


api/src/form_schema/forms/sf424.py
```diff
@@ -1066,7 +1066,10 @@
         "description": "XML transformation rules for converting Simpler SF-424 JSON to Grants.gov XML format",
         "version": "1.0",
         "form_name": "SF424_4_0",
-        "namespaces": {"default": "http://apply.grants.gov/forms/SF424_4_0-V4.0", "prefix": ""},
+        "namespaces": {
+            "default": "http://apply.grants.gov/forms/SF424_4_0-V4.0",
+            "globLib": "http://apply.grants.gov/system/GlobalLibrary-V2.0",
+        },
```
chouinar, line 1073: For the SF424, I believe we'd also need the `att` namespace pointing to https://apply07.grants.gov/apply/system/schemas/Attachments-V1.0.xsd

api/src/services/xml_generation/service.py
```diff
@@ -83,6 +146,94 @@ def _generate_xml_string(
...
+            # Create nested element for dictionary values
+            if field_name in namespace_fields:
+                # Use configured namespace
+                namespace_prefix = namespace_fields[field_name]
+                namespace_uri = nsmap.get(namespace_prefix, "")
+                element_name = f"{{{namespace_uri}}}{field_name}"
+                nested_element = lxml_etree.SubElement(parent, element_name)
+            else:
+                # Use default namespace (SF424_4_0)
+                default_namespace = nsmap.get("SF424_4_0", "")
```
chouinar: The default shouldn't correspond to any particular form, each form needs a configured default as we need to support 100+ forms with this logic eventually.

api/src/services/xml_generation/service.py
```diff
@@ -83,6 +146,94 @@ def _generate_xml_string(
...
+                namespace_uri = nsmap.get(namespace_prefix, "")
+                element_name = f"{{{namespace_uri}}}{field_name}"
+                lxml_etree.SubElement(parent, element_name)
+            else:
+                default_namespace = nsmap.get("SF424_4_0", "")
+                if default_namespace:
+                    element_name = f"{{{default_namespace}}}{field_name}"
+                    lxml_etree.SubElement(parent, element_name)
+                else:
+                    lxml_etree.SubElement(parent, field_name)
```
chouinar: A lot of this logic is copied multiple times in this function with only slight variation. Can we adjust this logic to move some of it to a function (a `get_element_name` function?).

api/src/services/xml_generation/service.py
```diff
@@ -66,11 +68,72 @@ def _generate_xml_string(
...
+        }
+
+        # Add additional namespaces
+        for prefix, uri in namespace_config.items():
+            if prefix != "default":
+                nsmap[prefix] = uri
+
+        # Add globLib namespace if any fields use it
+        if any(ns == "globLib" for ns in namespace_fields.values()):
+            nsmap["globLib"] = "http://apply.grants.gov/system/GlobalLibrary-V2.0"
```
chouinar: Wouldn't the `globLib` value for the nsmap already have this if it's pulled from the form's config? Is this just a leftover of setup before you added the config?

api/src/services/xml_generation/service.py
```diff
@@ -66,11 +68,72 @@ def _generate_xml_string(
...
+
+        # Create root element with proper namespace prefix
         if default_namespace:
-            root = ET.Element(root_element_name, xmlns=default_namespace)
+            root_element_with_namespace = f"{{{default_namespace}}}{root_element_name}"
+            root = lxml_etree.Element(root_element_with_namespace, nsmap=nsmap)
+
+            # Add FormVersion attribute with proper namespace prefix
+            root.set(f"{{{default_namespace}}}FormVersion", "4.0")
```
chouinar: Can we set/pull this out of the configuration? Each form will need a different value.

api/tests/src/services/xml_generation/test_configurable_namespaces.py
```diff
@@ -0,0 +1,93 @@
...
+        assert response.success
+        assert response.xml_data is not None
+
+        # Verify that address fields use globLib namespace
+        xml_content = response.xml_data
+        assert "globLib:Street1" in xml_content
+        assert "globLib:City" in xml_content
+        assert "globLib:State" in xml_content
+        assert "globLib:ZipPostalCode" in xml_content
+        assert "globLib:Country" in xml_content
```
chouinar, line 44: Do these get the `xmlns:globLib` bit added? For example: `<globLib:Street1 xmlns:globLib="http://apply.grants.gov/system/GlobalLibrary-V2.0">123 Main St</globLib:Street1>`

mikehgrantsgov: It does now. I initially tried this using lxml's attribute setting, but it looks like lxml automatically optimizes namespace declarations and strips them out, so in the end used a regex, which isn't ideal but couldn't find a way to override this using lxml.

api/src/services/xml_generation/service.py
mikehgrantsgov: Yep, I externalized this.

api/src/services/xml_generation/service.py
mikehgrantsgov: This was leftover, and removed it.

api/src/services/xml_generation/service.py
mikehgrantsgov: Makes sense, refactored this!

api/src/services/xml_generation/service.py
mikehgrantsgov: Removed this and some other default lookups (throwing an error if we can't find a value we expect vs. picking a default)

api/src/form_schema/forms/sf424.py
mikehgrantsgov: Added this as well

api/src/form_schema/forms/sf424.py
```diff
@@ -1092,13 +1096,13 @@
...
-        "state_code": {"xml_transform": {"target": "State"}},
-        "country_code": {"xml_transform": {"target": "Country"}},
-        "zip_code": {"xml_transform": {"target": "ZipPostalCode"}},
+        "street1": {"xml_transform": {"target": "Street1", "namespace": "globLib"}},
+        "street2": {"xml_transform": {"target": "Street2", "namespace": "globLib"}},
+        "city": {"xml_transform": {"target": "City", "namespace": "globLib"}},
+        "county": {"xml_transform": {"target": "County", "namespace": "globLib"}},
+        "state": {"xml_transform": {"target": "State", "namespace": "globLib"}},
+        "country": {"xml_transform": {"target": "Country", "namespace": "globLib"}},
+        "zip_postal_code": {"xml_transform": {"target": "ZipPostalCode", "namespace": "globLib"}},
```
chouinar:
```suggestion
        "zip_code": {"xml_transform": {"target": "ZipPostalCode", "namespace": "globLib"}},
```

api/src/form_schema/forms/sf424.py
```diff
@@ -1092,13 +1096,13 @@
...
-        "county": {"xml_transform": {"target": "County"}},
-        "state_code": {"xml_transform": {"target": "State"}},
-        "country_code": {"xml_transform": {"target": "Country"}},
-        "zip_code": {"xml_transform": {"target": "ZipPostalCode"}},
+        "street1": {"xml_transform": {"target": "Street1", "namespace": "globLib"}},
+        "street2": {"xml_transform": {"target": "Street2", "namespace": "globLib"}},
+        "city": {"xml_transform": {"target": "City", "namespace": "globLib"}},
+        "county": {"xml_transform": {"target": "County", "namespace": "globLib"}},
+        "state": {"xml_transform": {"target": "State", "namespace": "globLib"}},
+        "country": {"xml_transform": {"target": "Country", "namespace": "globLib"}},
```
chouinar, line 1105: There is also a province field

Issue #6396, opened 2025-09-15
Add logic to transform an attachment


chris-kuryak:
### Summary

TBD

### Acceptance criteria

- [ ] Attachment is transformed as expected

PR #6436, opened 2025-09-18
[Issue #6314] Setup tests for validating JSON conversion with XSD files [DRAFT]

### LINKED ISSUE #6314: Develop Configuration files for existing forms and build the header XML

mikehgrantsgov:
## Summary

Fixes #6314

## Validation steps

To run external tests:
`make test-xml-validation`
`make test-xml-validation xsd`
Shortcut command I have used: `docker compose run --rm grants-api poetry run pytest -m "not audit" tests/src/services/xml_generation/ -v`

Min test: `make generate-xml json='{"submission_type": "Application", "organization_name": "Test Org"}'`
Complete test: `make generate-xml json='{"submission_type": "Application", "application_type": "New", "date_received": "2025-01-15", "organization_name": "Example Organization", "employer_taxpayer_identification_number": "123456789", "sam_uei": "TEST12345678", "applicant_address": {"street1": "123 Main Street", "street2": "Suite 100", "city": "Washington", "county": "District of Columbia", "state": "DC", "country": "USA", "zip_postal_code": "20001-1234"}, "phone_number": "555-123-4567", "email": "test@example.org", "applicant_type_code": ["A: State Government"], "agency_name": "Test Agency", "funding_opportunity_number": "TEST-FON-2024-001", "funding_opportunity_title": "Test Funding Opportunity", "project_title": "Example Project with Configurable Namespaces", "congressional_district_applicant": "DC-00", "congressional_district_program_project": "DC-00", "project_start_date": "2025-04-01", "project_end_date": "2026-03-31", "federal_estimated_funding": "100000.00", "applicant_estimated_funding": "0.00", "state_estimated_funding": "0.00", "local_estimated_funding": "0.00", "other_estimated_funding": "0.00", "program_income_estimated_funding": "0.00", "total_estimated_funding": "100000.00", "state_review": "c. Program is not covered by E.O. 12372.", "delinquent_federal_debt": false, "certification_agree": true, "authorized_representative": {"first_name": "John", "last_name": "Doe"}, "authorized_representative_title": "CEO", "authorized_representative_phone_number": "555-123-4567", "authorized_representative_email": "john@example.org", "aor_signature": "John Doe Signature", "date_signed": "2025-01-15"}'`

External verification available here: https://www.freeformatter.com/xml-validator-xsd.html using SF424 XSD.


api/run_xml_validation.py
```diff
@@ -0,0 +1,133 @@
...
+        datefmt="%Y-%m-%d %H:%M:%S",
+    )
+
+
+def main() -> int:
+    """Main entry point for the validation script."""
+    parser = argparse.ArgumentParser(
+        description="Run XML validation tests against XSD schemas",
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        epilog="""
```
chouinar: Thoughts on making these flask CLI commands? Would handle setting up logging/db/etc and can use the click arg parsing.

api/run_xml_validation.py
```diff
@@ -0,0 +1,133 @@
...
+        "-v",
+        action="store_true",
+        help="Enable verbose logging",
+    )
+
+    args = parser.parse_args()
+
+    # Set up logging
+    setup_logging(args.verbose)
+    logger = logging.getLogger(__name__)
```
chouinar: We should be able to put the logger at the top of the file. I think that it only matters that you've setup logging before making a `logger.info` call. You can get the logger, but the config can be changed/modified before doing the log messages and it's fine. Otherwise all of the files we have `logging.getLogger` in wouldn't quite work as they'd be initialized at import before we've run the logging setup.

api/src/services/xml_generation/models/attachment.py
```diff
@@ -0,0 +1,202 @@
...
+
+        Args:
+            file_path: Path to the file
+
+        Returns:
+            Base64-encoded SHA-1 hash string
+        """
+        sha1_hash = hashlib.sha1(usedforsecurity=False)
+
+        with open(file_path, "rb") as f:
```
chouinar: Could we use `file_util.open_stream` - might have the files be on (local) s3 depending on how we set things up.

api/src/services/xml_generation/models/attachment.py
```diff
@@ -0,0 +1,202 @@
...
+            result["DebtExplanation"] = self._attachment_to_dict(self.debt_explanation)
+
+        # Multiple attachment field
+        if self.additional_project_title and self.additional_project_title.attached_files:
+            result["AdditionalProjectTitle"] = {
+                "AttachedFile": [
+                    self._attachment_to_dict(file)
+                    for file in self.additional_project_title.attached_files
+                ]
+            }
```
chouinar: Could we make this a thing in the configuration? Just some way of saying whether the field is an attachment, and whether it's a single one or an array?

api/src/services/xml_generation/models/attachment.py
```diff
@@ -0,0 +1,202 @@
...
+        if not file_path.exists():
+            raise FileNotFoundError(f"File not found: {file_path}")
+
+        # Generate base64-encoded SHA-1 hash
+        hash_value = cls.compute_base64_sha1(file_path)
+
+        # Determine MIME type
+        mime_type, _ = mimetypes.guess_type(str(file_path))
+        if mime_type is None:
+            mime_type = "application/octet-stream"  # Default fallback
```
chouinar: Could we prefer the mime type that we store on the application attachment table?

api/src/services/xml_generation/models/attachment.py
```diff
@@ -0,0 +1,202 @@
...
+            raise ValueError("MIME type must contain a slash (e.g., 'application/pdf')")
+        return v
+
+    @field_validator("hash_algorithm")
+    @classmethod
+    def validate_hash_algorithm(cls, v: str) -> str:
+        """Ensure hash algorithm is SHA-1."""
+        if v != "SHA-1":
+            raise ValueError("Only SHA-1 hash algorithm is supported by Grants.gov")
+        return v
```
chouinar: Should we just hardcode this value rather than validate what should be static?

api/src/services/xml_generation/models/attachment.py
```diff
@@ -0,0 +1,202 @@
...
+        """Compute base64-encoded SHA-1 hash from file content.
+
+        Args:
+            content: File content as bytes
+
+        Returns:
+            Base64-encoded SHA-1 hash string
+        """
+        sha1_hash = hashlib.sha1(content, usedforsecurity=False)
+        return base64.b64encode(sha1_hash.digest()).decode("utf-8")
```
chouinar: Just double-checking since the sha1 hash they required a bit of tinkering to figure out. Does this return the same results as `cat some_file.txt | openssl sha1 -binary | base64` does on the command-line? That's the command I found that matched what I was seeing grants.gov generate.

api/src/services/xml_generation/models/attachment.py
```diff
@@ -0,0 +1,202 @@
...
+            file_path: Path to the file
+            file_location: Custom file location URI (defaults to relative path)
+
+        Returns:
+            AttachmentFile instance with computed hash and MIME type
+        """
+        file_path = Path(file_path)
+
+        if not file_path.exists():
+            raise FileNotFoundError(f"File not found: {file_path}")
```
chouinar: A few things about the file path - we should assume the files are on s3 and any file logic should use our file utils. I think it's fine to assume a file does exist, we have validation of that behavior in our rule schemas.

But, for the actual path we put in the XML, that shouldn't come from the file itself. If you look at the logic for how we create the application ZIP, you'll see that there is special logic adjusting the file names, including handling duplicate file names. We'll need to pass that into this.

We might want a way to just take in a list/mapping of attachment IDs to objects that contain file info (the application attachment model + overriden name).

api/src/services/xml_generation/service.py
```diff
@@ -83,6 +157,381 @@ def _generate_xml_string(
...
+        """
+        try:
+            # Download and parse the main XSD
+            response = requests.get(xsd_url, timeout=30)
+            response.raise_for_status()
+            xsd_root = lxml_etree.fromstring(response.content)
+
+            # XSD namespace
+            xs_ns = {"xs": "http://www.w3.org/2001/XMLSchema"}
+
```
chouinar, line 271: I wonder if a utility that pre-fetches XSD files as a sort of cache would be helpful?

Like if I could run `make fetch-xsds` and it iterated through all the XSDs we might care about, putting them in a folder that we then use in these tests. Remove the network calls from anything but the setup.

api/tests/src/services/xml_generation/test_attachment_models.py
```diff
@@ -0,0 +1,342 @@
...
+            or not computed_hash.endswith("=")
+        )  # Base64 padding
+
+    def test_from_file_path_with_temp_file(self):
+        """Test creating AttachmentFile from temporary file path."""
+        test_content = b"This is test file content for attachment testing"
+
+        with tempfile.NamedTemporaryFile(suffix=".pdf", delete=False) as temp_file:
+            temp_file.write(test_content)
+            temp_file_path = Path(temp_file.name)
```
chouinar: Instead of tempfile, what about `tmp_path`?
https://docs.pytest.org/en/6.2.x/tmpdir.html

It'll handle cleaning up the files, can trim out a bit of the cleanup

api/tests/src/services/xml_generation/test_attachment_models.py
```diff
@@ -0,0 +1,342 @@
...
+        expected_sha1 = hashlib.sha1(test_content)  # nosec B324
+        expected_base64 = base64.b64encode(expected_sha1.digest()).decode("utf-8")
+
+        assert computed_hash == expected_base64
+        assert len(computed_hash) > 0
+        assert (
+            computed_hash.endswith("=")
+            or computed_hash.endswith("==")
+            or not computed_hash.endswith("=")
+        )  # Base64 padding
```
chouinar: What if for the test we just statically define the expected value? Just run `echo "This is test content for SHA-1 hash computation" | openssl sha1 -binary | base64` on the command-line and put the output here.

api/tool/xml_cli.py
```diff
@@ -0,0 +1,179 @@
+#!/usr/bin/env python3
+"""
+Simple CLI tool for generating SF-424 XML from JSON input.
+"""
```
chouinar: I know we're using the SF424 as a first schema for testing, but we should avoid building these tools to assume the schema is the SF424. This tool will need to support 100+ different forms/configurations.

I think what I was looking for for this tool was something that didn't need to take in parameters, it just had test cases that it ran through, taking in JSON and comparing the XML to some static expected value. This doesn't need to be the actual approach, but here's a rough rough idea of what I was thinking:

```py
# Define some static JSON
SF424_4_0_FULL_JSON = { ... }
SF424_4_0_MINIMAL_JSON = { ... }

SF424_4_0_FULL_XML = "..."
SF424_4_0_MINIMAL_XML = "..."

class XmlValidator:
    def __init__(self):
         tests_run = 0
         tests_failed = 0

    def validate_json_to_xml(self, scenario_name: str, json_data: dict, expected_xml: str, form: Form) -> None:
         self.tests_run += 1
         # setup/run
         result = ...
         # This comparison probably is more complex
         # than a simple string-comparison
         if result != expected_xml:
             logger.error(f"{scenario_name} did not pass validation: {error details}")
             self.tests_failed += 1

def run():
    xml_validator = XmlValidator()
    xml_validator.validate_json_to_xml("Full SF424 4.0", SF424_4_0_FULL_JSON, SF424_4_0_FULL_XML, SF424_4_0)
    xml_validator.validate_json_to_xml("Minimal SF424 4.0", SF424_4_0_MINIMAL_JSON, SF424_4_0_MINIMAL_XML, SF424_4_0)
```

api/tool/xml_cli.py
```diff
@@ -0,0 +1,179 @@
...
+    # Show example JSON
+    python tool/xml_cli.py --example
+    
+    # Generate from JSON string
+    python tool/xml_cli.py --json '{"organization_name": "Test Org"}'
+    
+    # Generate from file
+    python tool/xml_cli.py --file input.json
+        """
+    )
```
chouinar: Why not use the Flask CLI instead of a main function here?

api/src/services/xml_generation/service.py
mikehgrantsgov: The `XSDValidator` caches downloaded XSD files to disk (configurable via --cache-dir). After first download, validation runs offline. For example: `docker compose run --rm grants-api poetry run flask task validate-xml-generation --cache-dir ./xsd_cache`. I'll put some notes in the PR description when I flip to non-draft.

mikehgrantsgov: However, a helper could be nice. I think it's an easy add, too. I'll timebox it

api/src/cli/xml_generation_cli.py
```diff
@@ -0,0 +1,372 @@
...
+        click.echo(f"Error: Validation failed: {e}", err=True)
+        sys.exit(1)
+
+
+@task_blueprint.cli.command("fetch-xsds")
+@click.option(
+    "--cache-dir",
+    type=click.Path(),
+    help="Directory to cache XSD files (default: system temp directory)",
+)
```
chouinar, line 236: Thoughts on making the cache dir default to be next to the tests we run, but we put them in a folder we add to .gitignore?

api/src/services/xml_generation/models/attachment.py
chouinar: Re-adding comment, we can't do any sort of `file_path.exists()` check as that wouldn't work with s3 paths. Our file_util has a `file_exists` function that can do this for s3/local.

api/src/services/xml_generation/models/attachment.py
```diff
@@ -0,0 +1,193 @@
...
+
+    areas_affected: AttachmentFile | None = Field(None, description="Geographic areas affected")
+    additional_project_title: AttachmentGroup | None = Field(
+        None, description="Additional project title attachments"
+    )
+    additional_congressional_districts: AttachmentFile | None = Field(
+        None, description="Additional congressional districts"
+    )
+    debt_explanation: AttachmentFile | None = Field(None, description="Debt explanation document")
+
```
chouinar: We wouldn't want to define these as we'll need a bunch of different fields for different forms. Is there a way we could do this without some sort of class with specifically named fields?

api/src/services/xml_generation/models/attachment.py
```diff
@@ -0,0 +1,193 @@
...
+        # Determine MIME type - prefer provided mime_type from database
+        if mime_type is None:
+            # Fall back to guessing from file extension
+            mime_type, _ = mimetypes.guess_type(str(file_path))
+            if mime_type is None:
+                mime_type = "application/octet-stream"  # Default fallback
+
+        # Use provided file_location or generate relative path
+        if file_location is None:
+            file_location = f"./attachments/{file_path.name}"
```
chouinar: I'd make it so `file_location` can't be null. Making a value up here would likely just not point to anything (ie. we don't put attachments in a folder in the ZIP).

api/src/services/xml_generation/transformers/attachment_transformer.py
```diff
@@ -0,0 +1,316 @@
...
+            parent: Parent XML element
+            data: Data dictionary containing attachment information
+            nsmap: Namespace map for XML generation
+        """
+        # Handle single attachment fields
+        single_attachment_fields = [
+            ("AreasAffected", "areas_affected"),
+            ("AdditionalCongressionalDistricts", "additional_congressional_districts"),
+            ("DebtExplanation", "debt_explanation"),
+        ]
```
chouinar: Isn't whether a field is a single/multi-attachment in the configuration now? I think between the attachment file and this one, there is a bit of duplicated / extra code.

api/src/services/xml_generation/transformers/attachment_transformer.py
```diff
@@ -0,0 +1,316 @@
...
+        # Process file path-based attachment
+        if "file_path" in attachment_info:
+            try:
+                attachment_file = AttachmentFile.from_file_path(
+                    attachment_info["file_path"], attachment_info.get("file_location")
+                )
+                return self._attachment_file_to_dict(attachment_file)
+            except (FileNotFoundError, Exception):
+                # If file doesn't exist, create a placeholder structure
+                return self._create_placeholder_attachment(attachment_info)
```
chouinar: If a file isn't found, the process should fail. That would indicate a major issue with the data that we need to fix.

api/src/services/xml_generation/validation/test_cases.py
```diff
@@ -0,0 +1,582 @@
...
+            "total_estimated_funding": "100000.00",
+            "state_review": "c. Program is not covered by E.O. 12372.",
+            "delinquent_federal_debt": True,
+            "debt_explanation": {
+                "filename": "debt_explanation.pdf",
+                "mime_type": "application/pdf",
+                "file_location": "./attachments/debt_explanation.pdf",
+                "hash_value": "ZGVidCBleHBsYW5hdGlvbiBkb2N1bWVudA==",  # Base64-encoded SHA-1
+                "hash_algorithm": "SHA-1",
+            },
```
chouinar: I'm not following this test setup, the input JSON should have the value be a UUID. The transformation should change the values into this in XML. That would mean the test setup needs to be a bit more complex to setup creating the file (doesn't need to 100% create an application, but at least have that ID correspond to a file).

Effectively we would need the attachment transformation somewhere to handle converting a UUID into the application attachment object and then use that to do the transformation.

api/src/services/xml_generation/validation/xsd_validator.py
```diff
@@ -0,0 +1,450 @@
...
+logger = logging.getLogger(__name__)
+
+
+class XSDValidationError(Exception):
+    """Exception raised when XSD validation fails."""
+
+    pass
+
+
+class XSDValidator:
```
chouinar, line 19: I've tried re-reading through this class a few times and I'm just not following how it is used as there are a lot of branching paths and scenarios. It's for testing, we don't need to handle every edge case, it should have a simple way of running it.

Can we simplify it a bit? A few ideas:
* It needs to have the XSD file cache, no additional configs/overrides/etc. If the files are missing, let it error. Make it so running the script to fetch the XSDs is a pre-req to running this. That'll remove a lot of logic around managing those files and alternate ways of checking for them. One script to setup. One script to use those files to run tests.
* Have one way of getting a schema, if we assume the schemas are defined in one folder as files, just have that. `get_schema` isn't used for example.
* Separate the logic for fetching XSDs from the validation. This is doing both, fetching has its own complexities that aren't validation specific.

api/tool/xml_validator.py
chouinar, line 1: This seems similar to what is defined in test_cases.py - should only one of these exist?

chouinar (commented):
I'm having a hard time following some of this. I'm seeing things implemented multiple times, and some of those aren't referenced from anywhere else (guessing prior implementations that haven't been removed?). Can you take a pass and cleanup any extra pieces?

There's also a lot going on here, the attachment work really should be its own PR - its not quite doing what I would expect it to. The attachment fields in our JSON are UUIDs, but I'm not seeing any logic around mapping those UUIDs to the files which would be the first step in the attachment transformation logic.

mikehgrantsgov:
> I'm having a hard time following some of this. I'm seeing things implemented multiple times, and some of those aren't referenced from anywhere else (guessing prior implementations that haven't been removed?). Can you take a pass and cleanup any extra pieces?
> 
> There's also a lot going on here, the attachment work really should be its own PR - its not quite doing what I would expect it to. The attachment fields in our JSON are UUIDs, but I'm not seeing any logic around mapping those UUIDs to the files which would be the first step in the attachment transformation logic.

I'm digging into it. I think the problem is this branch has been around a while and has had lots of merges into it from older PRs that have been merged already. I think the duplication is the result of some odd merging. I'm going to split this up for attachments, potentially close this PR and open something cleaner.

api/src/services/xml_generation/validation/xsd_validator.py
mikehgrantsgov: Thanks for the feedback. Made a bunch of changes here:

- Attachment stuff is in its own branch and will be tackled as part of #6396
- We now assume the `xsd_fetcher` script to pull/cache XSDs will run with a static list of known XSDs (e.g. `docker compose run --rm grants-api poetry run flask task fetch-xsds`)
- After that step is done, we can use those XSDs later (e.g. `docker compose run --rm grants-api poetry run flask task validate-xml-generation`)
- XSD Validator much more straightforward and errors / avoid checking edge cases.

api/src/cli/xml_generation_cli.py
```diff
@@ -0,0 +1,383 @@
...
+    """
+    # Set up logging level
+    logger = logging.getLogger(__name__)
+    if verbose:
+        logging.getLogger().setLevel(logging.DEBUG)
+
+    try:
+        # Use default cache directory if not specified
+        if not cache_dir:
+            cache_dir = "./xsd_cache"
```
chouinar, line 173: Could we add this to the .gitignore?

api/tests/src/services/xml_generation/test_none_handling.py
```diff
@@ -8,6 +8,7 @@
 from src.services.xml_generation.transformers.base_transformer import RecursiveXMLTransformer
 
 
+@pytest.mark.xml_validation
```
chouinar: Since the new tests you added aren't actual unit tests, and these ones are fine to run already, do we need the pytest mark / filtering it out of our pytest command?

api/Makefile
```diff
@@ -336,6 +336,19 @@ console: ## Start interactive Python console
...
 
+##################################################
+# XML Validation Tests
+##################################################
+
+test-xml-validation: ## Run XML validation tests against XSD schemas (usage: make test-xml-validation form=SF424_4_0)
+	$(PY_RUN_CMD) flask task validate-xml-generation $(if $(form),--form $(form)) $(args)
+
+test-xml-validation-xsd: ## Run XML validation with custom XSD URL (usage: make test-xml-validation-xsd xsd=URL form=SF424_4_0)
+	$(PY_RUN_CMD) flask task validate-xml-generation --xsd-url "$(xsd)" $(if $(form),--form $(form)) $(args)
```
chouinar: I think this might be an old approach, doesn't look like there is an xsd-url parameter anymore.

api/src/services/xml_generation/validation/test_runner.py
```diff
@@ -0,0 +1,201 @@
...
+        Args:
+            summary: Test results summary
+        """
+        print("\n" + "=" * 60)
+        print("XML VALIDATION TEST SUMMARY")
+        print("=" * 60)
+        print(f"Total Tests: {summary['total_tests']}")
+        print(f"Successful: {summary['successful_tests']}")
+        print(f"Failed: {summary['failed_tests']}")
+        print(f"Success Rate: {summary['success_rate']:.1f}%")
```
chouinar: I wonder if this script should either a custom logging configuration to strip out some extra stuff (like app_name or run_mode) or just use click.echo/print. It logs a bunch of stuff out when running it, but this bit is the most readable of it.

Could also have the output result default to a file in the `xsd_cache` folder?

mikehgrantsgov: Good idea, I made the change to output the results to a configurable json file location

api/Makefile
mikehgrantsgov: Ah yeah, missed that piece to remove. Just did

api/tests/src/services/xml_generation/test_none_handling.py
mikehgrantsgov: I was on the fence about adding this but glad you called it out. I removed it.

api/src/cli/xml_generation_cli.py
mikehgrantsgov: Added!

Issue #6343, opened 2025-09-09
Add Name Fields to User Model and Database


joshtonava:
### Summary

Add first name, middle name, and last name fields to the User table to store user profile information. This will allow users to provide their full name information which can be used throughout the system.

### Tech Notes
- We can either put the fields directly on the user model or we can create a new user_profile 1-1 relationship. After speaking with @chouinar and @doug-s-nava I think there is some consensus about doing the user_profile join table approach 
- If we put a foreign key on the user model to the user_profile, it needs to be nullable for existing users

### Acceptance criteria
- Add three new nullable fields to the data model 
  - `first_name: Mapped[str]`
  - `middle_name: Mapped[str | None]`
  - `last_name: Mapped[str]`
- Include new migration

babebe:
Confirmed with @chouinar . 
A new `user_profile` table will be created. This table will be joined to the existing `user` table via a foreign key on `user_id`. By decoupling profile-specific data we ensure that existing entries in the users table are not impacted by the constraints introduced in the new table.

PR #6447, opened 2025-09-19
{Issue #6343] User profile table

babebe:
## Summary

Fixes / Work for #6343  

## Changes proposed
Model for `UserProfile` and migration file created.
Updated factory


api/src/db/models/user_models.py
```diff
@@ -421,3 +424,16 @@ class AgencyUserRole(ApiSchemaTable, TimestampMixin):
...
     role_id: Mapped[uuid.UUID] = mapped_column(ForeignKey(Role.role_id), primary_key=True)
     role: Mapped[Role] = relationship(Role)
+
+
+class UserProfile(ApiSchemaTable, TimestampMixin):
+    __tablename__ = "user_profile"
+
+    user_id: Mapped[uuid.UUID] = mapped_column(
+        ForeignKey(User.user_id), primary_key=True, index=True
+    )
```
chouinar: A primary key column is indexed automatically by Postgres - can skip doing that.
```suggestion
    user_id: Mapped[uuid.UUID] = mapped_column(
        ForeignKey(User.user_id), primary_key=True
    )
```
Some tables might have an index on it when we have a group of columns serving as a primary key, although I'm not sure how much that helps.

api/src/db/models/user_models.py
```diff
@@ -421,3 +424,16 @@ class AgencyUserRole(ApiSchemaTable, TimestampMixin):
 
     role_id: Mapped[uuid.UUID] = mapped_column(ForeignKey(Role.role_id), primary_key=True)
     role: Mapped[Role] = relationship(Role)
+
+
+class UserProfile(ApiSchemaTable, TimestampMixin):
+    __tablename__ = "user_profile"
+
+    user_id: Mapped[uuid.UUID] = mapped_column(
```
chouinar: Should we have this table have its own primary key, and then have user_id be unique? I generally like that approach better as it gives a bit more flexibility to change the table. I'm not sure we'd ever need it, but if we wanted a user to be able to have multiple profiles, all we would have to do is remove the uniqueness constraint to support that. If it's the primary key, the migration is a bit trickier.

babebe: Yes we can add it, to add the flexibility. I was working under the assumption that it would be 1-1 relationship.

Issue #6331, opened 2025-09-08
Update SF424A and SF424 configs to add auto-summation


chris-kuryak:
### Summary

Update SF424A and SF424 configs to add auto-summation.

All changes made are just in the library, configs need to be updated now.

We want to:
* Add the configuration to the rule schema for the SF424 and SF424A
* Add unit tests to verify the behavior works - should test with data in various states (full filled out, empty, in-between)

### Acceptance criteria

- [x] SF424A and SF424 auto summation functions as expected
- [ ] Unit tests added to verify the JSON rules are working as expected
- [x] UAT passed

chris-kuryak:
UAT passed.
Added a follow-on ticket to improve the UX on the SF-424A form as it requires the user to hit "save" before auto-summation occurs, which could be confusing and is not readily apparent to the user: #6875

PR #6451, opened 2025-09-19
[Issue #6331] Add auto summation to sf424 and sf424a

chouinar:
**NOTE: we will hold merging this for a few days, we don't need this for launch, and don't want to interfere with testing**
## Summary
Work for #6331

## Changes proposed
Add auto-summation to the SF424 and SF424A forms

## Context for reviewers
The actual implementation of auto-summation was already implemented, I did fix one minor issue with rounding, but otherwise it just required a whole bunch of configuration as there are 36 fields in the SF424A that have to be summed.

## Validation steps
Need both the frontend and API running locally.

Run `make db-seed-local` to get form changes in your DB. Navigate to the pilot-like opportunity with an application (via search or look at the link the seed script gives you).

If you go to the 424 or 424a and populate any of the monetary fields, the summations should work. The SF424A doesn't have the summed fields as non-fillable yet (needs frontend changes due to how that form works), but I was able to adjust the SF424:
Screenshot 2025-09-19 at 1 30 28 PM

Screenshot 2025-09-19 at 1 32 51 PM

### DIFF
--- a/api/src/form_schema/forms/sf424.py
+++ b/api/src/form_schema/forms/sf424.py
@@ -949,7 +949,7 @@ FORM_UI_SCHEMA = [
                 "definition": "/properties/program_income_estimated_funding",
             },
             {
-                "type": "field",
+                "type": "null",
                 "definition": "/properties/total_estimated_funding",
             },
         ],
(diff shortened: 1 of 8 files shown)


chouinar:
Holding on merging this until after we've launched the apply pilot. This is a nice-to-have feature and we don't want to introduce any complexity / don't have time to UAT it at the moment.

PR #6455, opened 2025-09-19
[Unticketed] Fix SF424 title fields to allow required setup properly

chouinar:
## Changes proposed
Move title fields up a level outside the `person_name` object so we can independently control whether they're required.

Made `authorized_representative_title` required.

## Context for reviewers
This was a data shape change I made from the XSD, effectively making it match the old shape while fixing this.

## Validation steps

Screenshot 2025-09-19 at 3 38 59 PM
Screenshot 2025-09-19 at 3 38 43 PM


mdragon (approved):
Looks good.

PR #6459, opened 2025-09-19
[Unticketed] Remove tier from required fields in SFLLL

chouinar:
## Changes proposed
Make tier field in SFLLL never required

## Context for reviewers
Double-checked the dat and XSD and it's not required regardless of other fields.
https://grants.gov/forms/form-items-description/fid/670
https://apply07.grants.gov/apply/forms/schemas/SFLLL_2_0-V2.0.xsd


mdragon (approved):
Approved

PR #6461, opened 2025-09-19
[Issue #6344] update user service schema

### LINKED ISSUE #6344: Update User Service and Schema to Include Name Fields
### Summary

Modify the user retrieval service and response schema to include the new name fields in API responses. 

### Tech Notes
- Currently the service returns LinkExternalUser but should return User data with name fields.
- Update `UserSchema` in `api/src/api/users/user_schemas.py` to include:
  - `first_name = fields.String(allow_none=True)`
  - `middle_name = fields.String(allow_none=True)`
  - `last_name = fields.String(allow_none=True)`

### Acceptance criteria

- Ensure the GET `/v1/users/{user_id}` endpoint returns the new user fields if they exist
- Add / Update unit tests

babebe:
## Summary

Fixes  #6344  

## Changes proposed
Update `user_get` endpoint to return `UserProfile` fields. 
Update schema


api/src/api/users/user_schemas.py
```diff
@@ -38,6 +38,27 @@ class UserSchema(Schema):
...
+            "example": "Hoover",
+        },
+    )
+    last_name = fields.String(
+        allow_none=True,
+        metadata={
+            "description": "The last name of the user",
+            "example": "Smith",
+        },
+    )
```
chouinar, line 47: We currently use the user schema in the GET users endpoint, and as part of the application schema for a few different application endpoints. If we set this schema up like this, we'll have to change every endpoint to do what you did in the get user service code (pull it out of the profile).

Two options:
* Make the shape of this schema match our db models (effectively add a profile object with the first/middle/last name in it) - that's been our general approach and is the simplest.
* Instead of handling the transformation of a user into the shape it needs to be in the service, we can do it on the schema itself. We can add a [pre_dump](https://marshmallow.readthedocs.io/en/stable/marshmallow.decorators.html#marshmallow.decorators.pre_dump) decorator and effectively transform the object in that - this might be more complex as nested objects we may eventually add (like roles) would also need to be handled. But at the same time it gives a lot of flexibility (note we're still on Marshmallow 3 so some of those docs might not match exactly)

api/src/services/users/get_user.py
```diff
@@ -1,18 +1,47 @@
...
+from src.db.models.user_models import LinkExternalUser, User
+
+
+class UserWithProfile(BaseModel):
+    user_id: UUID
+    email: str
+    external_user_type: ExternalUserType
+    first_name: str | None = None
+    middle_name: str | None = None
+    last_name: str | None = None
```
chouinar: For simple data containers, could we use dataclasses instead? The issue we had with the email job and aliases makes me think we should avoid Pydantic when we just want simple containers

api/src/services/users/get_user.py
```diff
@@ -1,18 +1,47 @@
...
     return user
 
 
-def get_user(db_session: db.Session, user_id: UUID) -> LinkExternalUser | None:
-    return _fetch_user(db_session, user_id)
+def get_user(db_session: db.Session, user_id: UUID) -> UserWithProfile | None:
+    external_user = _fetch_user(db_session, user_id)
+
+    if not external_user:
+        return None
```
chouinar: Hmm - if user is None, we should 404, that's not necessarily something new to the behavior of this PR, but we shouldn't ever return None from this function because not finding it should error.

babebe: Yes ! Of course ! Updated to raise error

api/src/api/users/user_schemas.py
babebe: I only saw it being used here. Application Schema uses `ApplicationUserSchema`

chouinar: Ah, you're right that we only use it in this endpoint at the moment, but we'll be adding several more user-related endpoints soon (like all the org endpoints, think of the endpoint to list users in an org) where we'd need to do this again.

babebe: Extracted`UserProfile` into its own schema class. And using pre_dump to flatten out the nested structure.

PR #6465, opened 2025-09-22
Frontend API Gateway key and secret

Nava-JoshLong:
## Summary

Fixes / Work for #ISSUE 

## Changes proposed

- API Gateway key for frontend
- SSM param for frontend to store key in
- Update frontend ECS task def

## Context for reviewers

This PR will create an API key for the frontend to use, and a SSM param that will require someone to copy the key value into (terraform will not do this automatically). I setup the param in the frontend playbook because it would cause a race condition if it was made in the API gateway file, since frontend and API would be deployed at the same time

## Validation steps

Dependent on code and steps from @btabaska to be tested via frontend. The gateway has been set to full proxy mode, so I would need to set an endpoint to have auth required and then test via that path

### DIFF
--- a/infra/frontend/service/secrets.tf
+++ b/infra/frontend/service/secrets.tf
@@ -14,3 +14,20 @@ module "secrets" {
   )
   manage_method = each.value.manage_method
 }
+
+resource "aws_ssm_parameter" "frontend_api_access_token" {
+  #checkov:skip=CKV_AWS_337:Use AWS-managed KMS for terraform-managed SSM parameters
+  name  = "/${module.app_config.app_name}/${var.environment_name}/X-API-KEY"
+  type  = "SecureString"
+  value = "Manually update from internal-${module.app_config.app_name}-${var.environment_name}-key"
+
+  description = "Token that the frontend uses to authenticate when making Grants API fetch requests via API Gateway."
+
+  # Won't overwrite value once changed from the placeholder, this value is set manually from the
+  # created key in the API playbook. Terraform cannot dynamically pull the token's value
+  lifecycle {
+    ignore_changes = [
+      value,
+    ]
+  }
+}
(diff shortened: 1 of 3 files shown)


infra/frontend/service/secrets.tf
```diff
@@ -14,3 +14,20 @@ module "secrets" {
   )
   manage_method = each.value.manage_method
 }
+
+resource "aws_ssm_parameter" "frontend_api_access_token" {
```
chouinar: Are we not able to define this secret / env var like our other secrets? https://github.com/HHS/simpler-grants-gov/blob/main/infra/frontend/app-config/env-config/environment_variables.tf#L35

If this is manually set, I think we just need to add it to that list unless there's something more complex I'm missing here

Nava-JoshLong: I wanted to do it this way to force the ECS task to wait on the SSM param to be created. With how the repo is set up, the ECS task def could be deployed before the SSM param is created. This isn't a huge deal since the image needs to be created, but I wanted to force sequencing. I can swap it to follow that logic if we really want

chouinar: Wouldn't we need to generate and store the key manually before we merge this either way?

Nava-JoshLong: @chouinar I have a change local moving it to that env var file, and it runs into an ordering issue. For it to work, I need to either make all of the SSM params in the account manually, force an ordering with a depends_on, or keep the PR as is. I can't set the param to generated since it will show drift when I set the API token in the param itself. 

Screenshot 2025-09-23 at 12 14 40 PM

I'm not sure what the normal process is, I can make the params if the normal process is to make them before merging the PR. The setup for env vars is a little brittle IMO, I think there could be an opportunity to create a param and ignore changes if the manage method is a third typing

chouinar: Yeah, our usual process assumes you setup the env var in parameter store before merging the change. That's how they've all behaved, I wouldn't consider it brittle, I think that it failing like that and saying "You merged a change without fully setting it up" is fine. If you setup a manual SSM value, you have to manually set it up.

Nava-JoshLong: Screenshot 2025-09-23 at 2 31 31 PM

Got it, params have been made, and terraform creating it has been removed. Plan is working as expected now

PR #6467, opened 2025-09-22
[No ticket] Fix inherit parent metadata linter

widal001:
## Summary

Fixes a bug in how the deliverable option ID was passed to one of our linter scripts which caused it to fail when the option ID looked like a number (i.e. no alpha characters)

## Changes proposed

- Switches flag passing option ID value to `--raw-field` to preserve the escaping which makes it a string
- Switches to the more verbose (and clear) `--field` and `--raw-field` instead of (`-F` and `-f` respectively)

## Context for reviewers

This error was flagged by @chris-kuryak after he noticed this GitHub action failed when he was creating some issues. It only failed on issues created for the "Scale apply user workflow *" deliverable because that option ID doesn't contain any alpha characters (`35970549`) and the gh CLI was parsing it as a number instead of a string even when escaping it with quotes.

## Validation steps

https://github.com/user-attachments/assets/cf4d5dc5-b7c3-4ff6-a03e-2981d13069ff

### DIFF
--- a/.github/linters/scripts/inherit-parent-deliverable.sh
+++ b/.github/linters/scripts/inherit-parent-deliverable.sh
@@ -138,7 +138,7 @@ fi
 
 # Update the deliverable field
 # Note: We need to pass the value as a raw field to prevent the gh CLI from parsing
-#       the option ID as a number if there are no alphanumeric characters e.g. 360523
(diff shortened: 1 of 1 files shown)


mdragon (approved):
Approved

Issue #6345, opened 2025-09-09
Create Update User Profile Endpoint


joshtonava:
### Summary

Create a new API endpoint that allows authenticated users to update their profile information, specifically their name fields. 

### Tech Notes
- This endpoint should follow the existing patterns in the codebase for user-specific operations.
- Create new service api/src/services/users/update_user_profile.py with function to update user name fields
- Add new schemas to `api/src/api/users/user_schemas.py`:
  - `UserUpdateProfileRequestSchema` with optional name fields and validation
  - `UserUpdateProfileResponseSchema` extending AbstractResponseSchema
- Follow existing patterns for logging and error handling

### Acceptance criteria

- PUT `/v1/users/{user_id}/profile` endpoint exists
- Request accepts optional `first_name`, `middle_name`, `last_name` fields
- Returns appropriate HTTP status codes (200 for success, 403 for forbidden, 400 for validation errors)
- Add unit tests

PR #6479, opened 2025-09-23
[Issue #6345] Update user profile endpoint

babebe:
## Summary

Fixes / Work for #6345  

## Changes proposed

Added PUT endpoint to update user profile 
Added Schema to support endpoint
Added Test

### DIFF
--- a/frontend/src/components/applyForm/widgets/AttachmentUploadWidget.tsx
+++ b/frontend/src/components/applyForm/widgets/AttachmentUploadWidget.tsx
@@ -78,6 +78,7 @@ const AttachmentUploadWidget = (props: UswdsWidgetProps) => {
       setAttachmentId(null);
       onChange?.(undefined);
       deleteModalRef.current?.toggleModal();
+      setAttachmentsChanged(true);
     }
   }, [deleteState, onChange]);
 
@@ -95,10 +96,6 @@ const AttachmentUploadWidget = (props: UswdsWidgetProps) => {
     );
 
     setShowFile(!!newAttachmentId);
-
-    if (newAttachmentId) {
-      setAttachmentsChanged(true);
-    }
   }, [value, attachments, setAttachmentsChanged]);
 
   const handleChange = async (
@@ -113,7 +110,6 @@ const AttachmentUploadWidget = (props: UswdsWidgetProps) => {
       setFileName(file.name);
       setShowFile(true);
       onChange?.(uploadedId);
-      setAttachmentsChanged(true);
     }
 
     fileInputRef.current?.clearFiles();


api/src/api/users/user_routes.py
```diff
@@ -577,3 +580,39 @@ def user_list_api_keys(
...
+        raise_flask_error(403, "Forbidden")
+
+    with db_session.begin():
+        updated_user_profile = update_user_profile(db_session, user_id, json_data)
+
+    logger.info(
+        "Updated profile for user",
+        extra={
+            "user_id": user_id,
+            "user_profile_id": updated_user_profile.user_profile_id,
```
chouinar, line 614: I think this'll always be null for new profiles (unless the commit at the end of `begin()` does a flush?). We could explicitly set an ID when creating the object to avoid that issue.

api/src/services/users/update_user_profile.py
```diff
@@ -0,0 +1,33 @@
...
+    middle_name: str | None = None
+
+
+def update_user_profile(db_session: db.Session, user_id: UUID, json_data: dict) -> UserProfile:
+    """Update user profile for a user"""
+    user_input = UpdateUserProfileInput.model_validate(json_data)
+
+    user_profile = db_session.query(UserProfile).filter(UserProfile.user_id == user_id).first()
+    if not user_profile:
+        logger.info(f"Creating new user profile for user_id: {user_id}")
```
chouinar: The user ID will be in the extra, can have a static log message.
```suggestion
        logger.info("Creating new user profile for user")
```

api/src/services/users/update_user_profile.py
```diff
@@ -0,0 +1,33 @@
...
+    first_name: str
+    last_name: str
+    middle_name: str | None = None
+
+
+def update_user_profile(db_session: db.Session, user_id: UUID, json_data: dict) -> UserProfile:
+    """Update user profile for a user"""
+    user_input = UpdateUserProfileInput.model_validate(json_data)
+
+    user_profile = db_session.query(UserProfile).filter(UserProfile.user_id == user_id).first()
```
chouinar: Could we do `one_or_none` + use select? First implies we expect more than 1 to be possible, but the user ID is unique in the table. Using select is the newer way, and while `query` isn't deprecated, SQLAlchemy refers to it as the old way and probably will eventually deprecate it.

```suggestion
    user_profile = db_session.execute(select(UserProfile).where(UserProfile.user_id == user_id)).scalar_one_or_none()
```

api/tests/src/api/users/test_user_profile_put.py
```diff
@@ -0,0 +1,67 @@
+import pytest
+
+from src.db.models.user_models import UserProfile
+from tests.lib.db_testing import cascade_delete_from_db_table
+from tests.src.db.models.factories import UserFactory, UserProfileFactory
+
+
+@pytest.fixture(autouse=True)
+def clear_data(db_session):
+    cascade_delete_from_db_table(db_session, UserProfile)
```
chouinar: I don't think we would need this, presumably any users we need in the tests would be created during the test? These deletes are slow and unless our tests absolutely need them (like jobs that process data across an entire table), we should avoid them.
```suggestion

```

api/tests/src/api/users/test_user_profile_put.py
```diff
@@ -0,0 +1,67 @@
...
+
+    res = db_session.query(UserProfile).first()
+    assert res.first_name == "Henry"
+    assert res.last_name == "Ford"
+    assert not res.middle_name
+
+
+def test_user_update_profile_update(client, db_session, user_auth_token, user):
+    user_profile = UserProfileFactory.create(user=user, first_name="Everett", last_name="Child")
+    db_session.commit()
```
chouinar: Factory create calls commit, don't need to add it.
```suggestion
    user_profile = UserProfileFactory.create(user=user, first_name="Everett", last_name="Child")
```

api/tests/src/api/users/test_user_profile_put.py
```diff
@@ -0,0 +1,67 @@
...
+        f"/v1/users/{UserFactory.create().user_id}/profile",
+        headers={"X-SGG-Token": user_auth_token},
+        json={"first_name": "Henry"},
+    )
+
+    assert response.status_code == 403
+    assert response.json["message"] == "Forbidden"
+
+    # Verify no record created
+    res = db_session.query(UserProfile).first()
```
chouinar: Could you verify that it wasn't created for the ID passed in?

api/tests/src/api/users/test_user_profile_put.py
```diff
@@ -0,0 +1,67 @@
...
+    assert res.first_name == data["first_name"]
+    assert res.last_name == data["last_name"]
+    assert res.middle_name == data["middle_name"]
+
+
+def test_user_update_profile_unauthorized(
+    client, db_session, user_auth_token, user, enable_factory_create
+):
+    response = client.put(
+        f"/v1/users/{UserFactory.create().user_id}/profile",
```
chouinar:
```suggestion
):
    user_id = uuid.uuid4()
    response = client.put(
        f"/v1/users/{user_id}/profile",
```

api/src/api/users/user_routes.py
babebe: Yes. Setting the ID now.

api/tests/src/api/users/test_user_profile_put.py
babebe: Filtering on the user id

api/tests/src/api/users/test_user_profile_put.py
babebe: Removed. Yes i don't need them actually fixtures are scoped to function.

PR #6486, opened 2025-09-23
[ISSUE #6371] Adjust agency transform job to warn on fields that are not accounted for and continue processing

### LINKED ISSUE #6371: Adjust agency transform job to warn on fields that are not accounted for and continue processing
### Summary

In the agency transform job if there are any fields that are not accounted for, an error is thrown and we do not continue processing that agency.
We would like to skip rows for unknown fields and just log a warning so we don’t have to deal with the error logs.

### Tech Notes
A potential implementation could look like this
```python
has_mapping = (
    tgroup_field_name == "AgencyDownload" or
    tgroup_field_name in AGENCY_FIELD_MAP or
    tgroup_field_name in AGENCY_CONTACT_INFO_FIELD_MAP
)

if not has_mapping:
    if tgroup_field_name in NOT_MAPPED_FIELDS:
        logger.info("Skipping processing of field %s for %s", ...)
    else:
        logger.warning("Skipping unmapped field %s for %s - consider adding to NOT_MAPPED_FIELDS if intentional", ...)
    continue
```

Thread for context
https://betagrantsgov.slack.com/archives/C05TSL64VUH/p1757533560025859?thread_ts=1757529881.087119&cid=C05TSL64VUH

### Acceptance criteria

- If there is a non mapped field in the agency transfer job that is not ignored, we should log a warning and continue processing the agency.
- Update tests

joshtonava:
## Summary

Fixes #6371 

## Changes proposed

- Removed `raise ValueError("Unknown tgroups agency field %s")` that would stop processing the agency
- Added two-tier field handling:
  - __Known unmapped fields__ (in `NOT_MAPPED_FIELDS`): Log INFO message and skip
  - __Truly unknown fields__: Log WARNING message with suggestion to add to `NOT_MAPPED_FIELDS` if intentional, then skip
- Updated tests

## Context for reviewers

This change improves the robustness of the data migration process by preventing unknown fields from causing entire agency records to fail processing

## Validation steps

```bash
make test args="-vv tests/src/data_migration/transformation/subtask/test_transform_agency.py"
```

### DIFF
--- a/api/tests/src/data_migration/transformation/subtask/test_transform_agency.py
+++ b/api/tests/src/data_migration/transformation/subtask/test_transform_agency.py
@@ -1,5 +1,5 @@
-from datetime import datetime
 import logging
+from datetime import datetime
 
 import pytest
 


api/src/data_migration/transformation/subtask/transform_agency.py
```diff
@@ -426,6 +427,21 @@ def get_agency_updates(tgroup_agency: TgroupAgency) -> AgencyUpdates:
...
+            or tgroup_field_name in AGENCY_CONTACT_INFO_FIELD_MAP
+        )
+
+        if not has_mapping:
+            logger.warning(
+                "Skipping unmapped field %s for %s - consider adding to NOT_MAPPED_FIELDS if intentional",
+                tgroup_field_name,
+                tgroup_agency.agency_code,
+            )
+            continue
```
chouinar, line 443: If we skip the record here, we'll set the `transformed_at` timestamp and not process it again.

I could go either way on this, but do we want to keep receiving these each run or stop? If we stop, but actually wanted some new field, we'd need to mark those fields to be transformed again, which isn't that big a deal. Thoughts?

joshtonava: It might be better keep receiving the alerts and choose to add the field to the NOT_MAPPED_FIELDS fields.

One the plus side, we avoid having to get into the DB but on the downside, it requires a deploy.

@chouinar

joshtonava: My main concern is will we ack the alertbut the agency slips through the cracks

joshtonava: We can try another approach if this becomes unmanagable, lets give this a go for now

PR #6504, opened 2025-09-24
[Issue 5415] Restrict outbound traffic on security groups

prasnava:
## Summary

Restrict outbound traffic on security groups to resolve the failing check ov rule: CKV_AWS_382 

**Changes proposed**
Restricted outbound traffic on security groups to ports 443, 80 and 5432 to resolve the check ov finding CKV_AWS_382.

- Load balancer outbound security group allows traffic back to the app security group for local health checks 
- App security group is only the 3 ports (443, 80 and 5432) outbound.

**Context for reviewers**
Currently, the outbound security groups for the service module is allowing all outbound traffic by default. We want to limit traffic to only secure ports and destinations.

**Validation steps**
Run terraform apply in the training or dev environment for the api service module. The step functions and app should continue to work without any errors.

### DIFF
--- a/api/openapi.generated.yml
+++ b/api/openapi.generated.yml
@@ -24,6 +24,7 @@ tags:
 - name: Opportunity v1
 - name: Extract v1
 - name: Agency v1
+- name: Organization v1
 - name: User v1 - Internal Only
 - name: Application Alpha
 - name: Form Alpha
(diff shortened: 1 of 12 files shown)


mdragon:
Can you flesh out the PR body a bit more with what was changed (high level like AWS Services Impacted) and how it was tested?

mdragon (approved):
🚀

Issue #6115, opened 2025-08-26
Remove the is_owner flag from both the application_user and organization_user table


joshtonava:
### Summary

With the application and organization roles assigned to existing users https://github.com/HHS/simpler-grants-gov/issues/6113 and https://github.com/HHS/simpler-grants-gov/issues/6114 complete, we no longer need the is_owner flag on the application user and organization user tables.

### Acceptance criteria

- Remove usage of `is_application_owner` from application_user table
- Remove usage of `is_organization_owner` from organization_user table
- Update tests
- Create a data migration to remove the columns from both tables

joshtonava:
TODO: @joshtonava double check this ticket is still needed

PR #6505, opened 2025-09-24
[ISSUE #6338] Organization endpoints - Get organization details

### LINKED ISSUE #6338: Organization endpoints - Get organization details
### Summary

Implement an endpoint to retrieve organization information by ID. This endpoint will return basic organization details including the SAM.gov entity information (UEI, legal business name, expiration date).

https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/1977221432/Tech+Spec+User+Roles+on+the+Backend#GET-%2Forganizations%2F%3Aorganization_id

### Tech Notes
- **Required Permission**: User must have any role in the organization (any organization-scoped privilege grants read access to basic org info). Alternatively, could require `view_org_membership` privilege if you want to be more restrictive. The spec leaves this as "TODO" - needs decision on whether all org members can view basic org details

### Acceptance criteria

- Implement GET /organizations/:organization_id
- Returns organization data including SAM.gov entity information
- Include unit tests

joshtonava:
## Summary

Fixes #6338  

## Changes proposed

- Added GET /v1/organizations/:organization_id endpoint to retrieve organization information by ID, including associated SAM.gov entity data.
- Privilege-based authorization using `VIEW_ORG_MEMBERSHIP`

## Context for reviewers

This endpoint implements privilege-based access control using the existing endpoint_access_util.can_access() function. Users must have the VIEW_ORG_MEMBERSHIP privilege for the specific organization they're trying to access. 

## Validation steps

To run the affected tests

```bash
make test args="-vv tests/src/api/organizations_v1/test_organization_routes.py"
```

### DIFF
--- a/api/openapi.generated.yml
+++ b/api/openapi.generated.yml
@@ -3771,22 +3771,22 @@ components:
     UserApplicationListRequest:
       type: object
       properties: {}
-    SamGovEntity:
+    SamGovEntityResponse:
       type: object
       properties:
         uei:
           type: string
-          description: Unique Entity Identifier from SAM.gov
-          example: ABCDEFGHIJK1
+          description: Unique Entity Identifier
+          example: 000123456789
         legal_business_name:
           type: string
-          description: Legal business name as registered in SAM.gov
-          example: Example Research Corporation
+          description: Legal business name from SAM.gov
+          example: Example Inc.
         expiration_date:
           type: string
           format: date
-          description: The expiration date of the SAM.gov registration
-          example: '2025-12-31'
+          description: SAM.gov registration expiration date
+          example: '2025-08-11'
     UserOrganization:
       type: object
       properties:
@@ -3804,7 +3804,7 @@ components:
           - object
           - 'null'
           anyOf:
-          - $ref: '#/components/schemas/SamGovEntity'
+          - $ref: '#/components/schemas/SamGovEntityResponse'
           - type: 'null'
     UserApplicationCompetition:
       type: object
@@ -4231,22 +4231,6 @@ components:
           type: integer
           description: The HTTP status code
           example: 200
-    SamGovEntityResponse:
-      type: object
-      properties:
-        uei:
-          type: string
-          description: Unique Entity Identifier
-          example: 000123456789
-        legal_business_name:
-          type: string
-          description: Legal business name from SAM.gov
-          example: Example Inc.
-        expiration_date:
-          type: string
-          format: date
-          description: SAM.gov registration expiration date
-          example: '2025-08-11'
     OrganizationData:
       type: object
       properties:
@@ -4413,7 +4397,7 @@ components:
           - object
           - 'null'
           anyOf:
-          - $ref: '#/components/schemas/SamGovEntity'
+          - $ref: '#/components/schemas/SamGovEntityResponse'
           - type: 'null'
     ApplicationGetResponseData:
       type: object


api/tests/src/api/organizations_v1/test_organization_routes.py
```diff
@@ -0,0 +1,396 @@
...
+            f"/v1/organizations/{organization.organization_id}",
+            headers={"X-SGG-Token": token},
+        )
+
+        assert resp.status_code == 200
+        data = resp.get_json()
+        assert data["message"] == "Success"
+        assert data["data"]["organization_id"] == str(organization.organization_id)
+
+    @pytest.mark.parametrize(
```
joshtonava, line 243: let me know if this is overkill

api/tests/src/api/organizations_v1/test_organization_routes.py
```diff
@@ -0,0 +1,396 @@
...
+
+        # Create organization with SAM.gov entity
+        organization = OrganizationFactory.create(sam_gov_entity=sam_gov_entity)
+
+        # Create role with VIEW_ORG_MEMBERSHIP privilege
+        role = RoleFactory.create(privileges=[Privilege.VIEW_ORG_MEMBERSHIP], is_org_role=True)
+
+        # Create organization-user relationship with role
+        org_user = OrganizationUserFactory.create(
+            user=user, organization=organization, is_organization_owner=True
```
joshtonava: I have a ticket to clean this up. https://github.com/HHS/simpler-grants-gov/issues/6115 which is slated for this sprint.

api/src/api/organizations_v1/organization_routes.py
```diff
@@ -0,0 +1,84 @@
...
+    user_id = user_token_session.user_id
+
+    # Query organization with SAM.gov entity data
+    organization = (
+        db_session.query(Organization)
+        .options(joinedload(Organization.sam_gov_entity))
+        .filter(Organization.organization_id == organization_id)
+        .first()
+    )
+
```
chouinar: Can we move everything from here down to return the api response into a separate service file like our other APIs? We don't put DB logic in our routes directly to keep them simple / reusable.

We also want to do this in a transaction, I know it's a get endpoint, but when we add audit log tables (later this quad), even GET endpoints will write to the DB.

```py
with db_session.begin():
     # we'll probably have an underlying get_organization function, this one just adds the auth checking
     organization = get_organization_and_verify_access()

return response.ApiResponse(message="Success", data=organization)
```

Having a singular `get_organization` method that fetches an organization and raises a 404 is something we'll want on half a dozen different endpoints, so we should make it something central and reusable.

api/src/api/organizations_v1/organization_routes.py
```diff
@@ -0,0 +1,84 @@
...
+
+@organization_blueprint.get("/<uuid:organization_id>")
+@organization_blueprint.output(OrganizationGetResponseSchema)
+@organization_blueprint.doc(responses=[200, 401, 403, 404])
+@organization_blueprint.auth_required(api_jwt_auth)
+@flask_db.with_db_session()
+def organization_get(db_session: db.Session, organization_id: UUID) -> response.ApiResponse:
+    """Get organization information by ID"""
+    logger.info("GET /v1/organizations/:organization_id")
```
chouinar, line 25:
```suggestion
def organization_get(db_session: db.Session, organization_id: UUID) -> response.ApiResponse:
    """Get organization information by ID"""
    add_extra_data_to_current_request_logs({"organization_id": organization_id})
    logger.info("GET /v1/organizations/:organization_id")
```

api/src/api/organizations_v1/organization_routes.py
```diff
@@ -0,0 +1,84 @@
...
+    user_token_session: UserTokenSession = api_jwt_auth.get_user_token_session()
+    user_id = user_token_session.user_id
+
+    # Query organization with SAM.gov entity data
+    organization = (
+        db_session.query(Organization)
+        .options(joinedload(Organization.sam_gov_entity))
+        .filter(Organization.organization_id == organization_id)
+        .first()
+    )
```
chouinar: Can we use `select` instead of `query`? SQLAlchemy recommends the newer style queries and implies they'll eventually deprecate the old style queries.

Also, we should always use `one_or_none` instead of `first` if we expect exactly 1 thing like when querying by a primary key. First implies multiple records could be possible and fetches a list of records instead of a single row.
```suggestion
    organization = (
        db_session.execute(select(Organization)
        .options(joinedload(Organization.sam_gov_entity))
        .where(Organization.organization_id == organization_id))
        .scalar_one_or_none()
    )
```

api/src/api/organizations_v1/organization_routes.py
```diff
@@ -0,0 +1,84 @@
...
+    # Check if user has VIEW_ORG_MEMBERSHIP privilege for this organization
+    if not can_access(user, {Privilege.VIEW_ORG_MEMBERSHIP}, organization):
+        logger.info(
+            "User does not have VIEW_ORG_MEMBERSHIP privilege for organization",
+            extra={
+                "organization_id": organization_id,
+                "user_id": user_id,
+            },
+        )
+        raise_flask_error(403, "Forbidden")
```
chouinar: Since this pattern is going to be incredibly common going forward, can we make a function that wraps `can_access` and adds the logging/raise_flask_error?

Then we just have to do:
```py
verify_can_access(user, {Privilege.VIEW_ORG_MEMBERSHIP}, organization)
```

api/src/api/organizations_v1/organization_routes.py
```diff
@@ -0,0 +1,84 @@
...
+    user = db_session.query(User).filter(User.user_id == user_id).first()
+    if not user:
+        logger.error(
+            "User not found for token session",
+            extra={
+                "user_id": user_id,
+                "organization_id": organization_id,
+            },
+        )
+        raise_flask_error(401, "Unauthorized")
```
chouinar: Are we not able to use the user from the `user_token_session`? It is fetched by a different db session, but I think there are some ways to put it in this session + fetch the privileges ahead of time by changing the logic where we grab the user in the auth.

api/src/api/organizations_v1/organization_routes.py
```diff
@@ -0,0 +1,84 @@
...
+            },
+        )
+        raise_flask_error(403, "Forbidden")
+
+    logger.info(
+        "Successfully retrieved organization",
+        extra={
+            "organization_id": organization_id,
+            "user_id": user_id,
+            "has_sam_gov_entity": organization.sam_gov_entity is not None,
```
chouinar: While the DB model might allow it, practically this won't happen. Unless we create an alternative approach to setting up orgs from sam.gov, we'll always have this.

api/tests/src/api/organizations_v1/test_organization_routes.py
chouinar: It might be helpful for us to make a utility that wraps a bunch of what you have here. a function that we can call like:

```py
def create_user_in_org(privileges: list[Privilege], ...):
```
That creates a user in an org with specific privileges (and probably some other values). Just because that takes like 7 factories to do.

api/src/api/organizations_v1/organization_routes.py
joshtonava: I just realized this is already done in the can_access so my logging is redundant

api/tests/src/api/organizations_v1/test_organization_routes.py
joshtonava: Done!

api/src/api/organizations_v1/organization_routes.py
joshtonava: Done!

api/src/api/organizations_v1/organization_routes.py
joshtonava: Taking note for the future

api/src/api/organizations_v1/organization_routes.py
joshtonava: Just figured this out thanks for the tip

api/src/api/organizations_v1/organization_routes.py
joshtonava: Im not doing this log anymore

api/src/api/organizations_v1/organization_routes.py
```diff
@@ -0,0 +1,40 @@
...
+    add_extra_data_to_current_request_logs({"organization_id": organization_id})
+    logger.info("GET /v1/organizations/:organization_id")
+
+    # Get authenticated user
+    user_token_session: UserTokenSession = api_jwt_auth.get_user_token_session()
+
+    with db_session.begin():
+        # Add the user from the token session to our current session
+        # This prevents DetachedInstanceError when accessing user relationships
+        db_session.add(user_token_session)
```
chouinar, line 33: FYI - I added something to the eng sync agenda today to chat about fetching roles/privileges/etc. in our auth logic - not a blocker here, but might simplify some stuff under the hood & improve performance.

api/src/api/organizations_v1/organization_schemas.py
```diff
@@ -0,0 +1,40 @@
+from src.api.schemas.extension import Schema, fields
+from src.api.schemas.response_schema import AbstractResponseSchema
+
+
+class SamGovEntityResponseSchema(Schema):
```
chouinar, line 5: I just noticed that we have two schemas for org + sam gov entity because we made an endpoint to list orgs for a user:
https://github.com/HHS/simpler-grants-gov/blob/0078965dd15de6e111d751bbd5564bad34831725/api/src/api/application_alpha/application_schemas.py#L239

I think these match? I do think they should be defined in their own file like you have here, especially as we add more stuff. Can you move/make this one the only one that exists?

api/src/api/organizations_v1/organization_routes.py
joshtonava: I couldnt make it today is there something that came out of it I should change?

api/src/api/organizations_v1/organization_schemas.py
joshtonava: That can be done

api/src/api/organizations_v1/organization_routes.py
chouinar: I'm going to look into some details regarding the logic where we fetch a user in our auth decorators. Want to see if we can/should fetch user roles/privileges there so that this user token session object has all of that already loaded.

joshtonava: Oh sounds cool. Let me know if I / we can help with the implementation.

chouinar: Made a thread in the simpler engineering channel with some details. Not a blocker here, but will give the caveat that right now as implemented, when its doing the `can_access` check, this would be going back to the DB for every relationship since that's not already loaded.

We can adjust that later, but should address before launch as I'm guessing that's at least a few hundred milliseconds.

api/src/api/application_alpha/application_schemas.py
```diff
@@ -216,26 +217,6 @@ class ApplicationUserSchema(Schema):
...
-    )
-    expiration_date = fields.Date(
-        metadata={
-            "description": "The expiration date of the SAM.gov registration",
-            "example": "2025-12-31",
-        }
-    )
-
-
 class OrganizationSchema(Schema):
```
chouinar, line 220: Could this also be consolidated/use the one you added?

api/src/api/organizations_v1/organization_routes.py
joshtonava: Alright we can crystallize that implementation in the upcoming organization endpoints that need implementing

api/src/api/application_alpha/application_schemas.py
joshtonava: Yes I can do this in the list organization endpoint I am working on now.

Issue #6325, opened 2025-09-08
User endpoints - Get user privileges and roles


joshtonava:
### Summary

Implement an endpoint that returns all roles and privileges for a user across all resources (organizations, applications, agencies) they have access to. This endpoint enables the frontend to conditionally show/hide UI elements based on user permissions, such as graying out "manage users" links for users without the appropriate privileges.

The endpoint should support filtering by specific resource IDs to get permissions for just that context (e.g., "what can this user do in organization X?") or return all permissions when no filter is provided.

- [user privileges endpoint](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/1977221432/Tech+Spec+User+Roles+on+the+Backend#POST-%2Fusers%2F%3Auser_id%2Fprivileges)
- [data concept](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/1977221432/Tech+Spec+User+Roles+on+the+Backend#Data-Concept)

### Tech Notes

- Authentication: Users can only access their own privileges (uses existing auth pattern where user can access their own data by user_id)
- No specific privilege required - this is about users viewing their own permissions
- The response structure from the spec is nested and comprehensive - consider if this needs flattening for easier frontend consumption
- Empty request body should return all user roles across all resources

### Acceptance criteria

- Implement `POST /users/:user_id/privileges`
- Users can only access their own privileges (return 403 for other user IDs)
- Support filtering by organization_id in request body (future)
- Support filtering by application_id in request body (future)
- Support filtering by agency_id in request body (future)
- Empty request returns all roles/privileges across all resources
- Response includes user_organizations with embedded roles and privileges
- Response includes user_applications with embedded roles and privileges
- Response includes user_agencies with embedded roles and privileges
- Response includes internal_roles if user has any
- Each role object includes role_id, role_name, and privileges array
- Validate that filtered resource exists (return 404 if not)
- Add unit tests

eric-serrato95:
This ticket assumes no filtering is available and handled separately at a later point.

babebe:
Spoke with @joshtonava . Confirmed that filtering on organization_id will not be captured in this ticket. And would only need to return the ID's for the resources and not the full object.

PR #6520, opened 2025-09-25
[Issue #6325] Get user privilege role endpoint 

babebe:
## Summary

Fixes for #6325  

## Changes proposed

POST endpoint for getting a users role and privileges. Does not support filtering yet. 
Added Response Schema
Added Test


api/src/api/users/user_schemas.py
```diff
@@ -389,3 +395,130 @@ class UserUpdateProfileRequestSchema(UserProfile):
 
 class UserUpdateProfileResponseSchema(AbstractResponseSchema):
     data = fields.Nested(UserProfile, metadata={"description": "The updated user profile"})
+
+
+class UserRolesAndPrivileges(Schema):
```
chouinar: Thoughts on calling this `UserRoleSchema`? We should make at least the role bit singular as this is a single role object.

api/src/services/users/get_roles_and_privileges.py
```diff
@@ -0,0 +1,82 @@
...
+) -> User | None:
+    stmt = (
+        select(User)
+        .where(User.user_id == user_id)
+        .options(
+            # Organization-related relationships
+            selectinload(User.organizations),
+            selectinload(User.organizations).selectinload(
+                OrganizationUser.organization
+            ),  # Load Organization inside each OrganizationUser
```
chouinar: I know we haven't really used them up to this point, but I wonder if putting some of these load options on the relationships themselves as defaults might be a good idea.

Using organization as an example, we effectively have to connect Organization -> OrganizationUser -> OrganizationUserRole -> Role. Then repeat that for agency + application.

While we don't need to fetch everything up front every time, I feel like if someone fetches the `OrganizationUserRole` they probably always want the Role.

api/src/services/users/get_roles_and_privileges.py
```diff
@@ -0,0 +1,82 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def get_roles_and_privileges(
+    db_session: db.Session,
+    user_id: UUID,
+) -> User | None:
+    stmt = (
+        select(User)
```
chouinar, line 19: This'll sound strange but.. do we even need to make this DB query? (We should still go into the service code because we'll eventually want filters, but bear with me for a sec)

Since we fetch the user during auth, we _should_ already have the user and these relationships loaded and could basically have this API call do:

```py
    user_token_session: UserTokenSession = api_jwt_auth.get_user_token_session()

    # Verify the authenticated user matches the requested user_id
    if user_token_session.user_id != user_id:


    return response.ApiResponse(message="Success", data=user_token_session.user)
```
Assuming we load all the relationships in the API JWT checking logic, would that work?
@joshtonava - this is related to something I mentioned on your PR about how we shouldn't fetch a user that we already fetched. It might be that the first change we make before any of these endpoint changes is actually adjusting the loading of the user. Effectively moving the below query to the auth logic. I'll put this on the eng sync discussion.

api/src/services/users/get_roles_and_privileges.py
```diff
@@ -0,0 +1,82 @@
...
+            .selectinload(OrganizationUser.organization_user_roles)
+            .selectinload(
+                OrganizationUserRole.role
+            )  # Load Role and associated privileges inside each OrganizationUserRole
+            .selectinload(Role.link_privileges),
+            # Application-related relationships
+            selectinload(User.application_users),  # Load list of ApplicationUser
+            selectinload(User.application_users).selectinload(
+                ApplicationUser.application
+            ),  # Load Application inside each ApplicationUser
```
chouinar: Do we need to fetch the organization/application/agency? From the schema, it looks like just the ID is returned which can be fetched from the `<resource>_users` record

babebe: @chouinar . Yes, if we only want the IDs we won't need this, but we would need to change the structure of the response schema 
`"user_applications": [
    {
      "application_id": ...., 
      "roles": [...]
    }
  ],`
  
  instead of :
  `"user_applications": [
    {
      "application": {"application_id":...}, 
      "roles": [...]
    }
  ],`
  
  or would need do a pre_dump function to build the nested structure manually

api/src/services/users/get_roles_and_privileges.py
babebe: Updated to always fetch the role and the privileges

api/src/services/users/get_roles_and_privileges.py
```diff
@@ -0,0 +1,47 @@
...
+from src.adapters import db
+from src.db.models.user_models import AgencyUser, ApplicationUser, OrganizationUser, User
+
+logger = logging.getLogger(__name__)
+
+
+def get_roles_and_privileges(
+    db_session: db.Session,
+    user_id: UUID,
+) -> User | None:
```
chouinar, line 17: If this is null, we should raise a 404 just so this function can always say it returns something. Now, in the context of this implementation, we can't actually hit this scenario (we know the user ID matches the user we authed, so that user MUST exist or auth would have failed), but should add that just to follow the convention. If we use this function for a different endpoint where that isn't true, it'd be already setup.

api/src/services/users/get_roles_and_privileges.py
```diff
@@ -0,0 +1,47 @@
...
+            selectinload(User.application_users).selectinload(
+                ApplicationUser.application_user_roles
+            ),
+            # Agency-related relationships
+            selectinload(User.user_agencies),  # Load list of AgencyUser
+            selectinload(User.user_agencies).selectinload(
+                AgencyUser.agency
+            ),  # Load Agency inside each AgencyUser
+            selectinload(User.user_agencies).selectinload(AgencyUser.agency_user_roles),
+            # Internal roles
```
chouinar: I think with the change you made, we can trim out a bit more. I think we can make the selectinload section just:
```py
        .options(
            # Organization-related relationships
            selectinload(User.organizations).selectinload(OrganizationUser.organization),
            # Application-related relationships
            selectinload(User.application_users).selectinload(ApplicationUser.application),
            # Agency-related relationships
            selectinload(User.user_agencies).selectinload(AgencyUser.agency),
            # Internal roles
            selectinload(User.internal_user_roles),
        )
```
Shouldn't need to do `User.user_agencies` twice (first line + in the middle chain) and the last selectinload is for the agency user roles which should already be fetched. Testing it, it doesn't seem hit any issues with the detached state error at this amount (removing any of this does).

api/src/api/users/user_schemas.py
```diff
@@ -389,3 +395,130 @@ class UserUpdateProfileRequestSchema(UserProfile):
...
+            UserRoleSchema, metadata={"description": "Roles the user has for this agency"}
+        )
+    )
+
+
+class GetRolesAndPrivilegesResponseSchema(Schema):
+    """Schema for the full response containing user roles and privileges."""
+
+    user_id = fields.UUID()
+    organizations = fields.List(
```
chouinar: I know this just matches what the DB model has setup, but could we rename this `organization_users` to match the others? Ideally we'd rename that relationship, but if that's more than like 5 minutes of work to do, we can add a property like this to the user model for now.

```py
def organization_users(self) -> list[OrganizationUser]:
     return self.organizations
```

doug-s-nava: [peanut gallery]: aren't these all pointing to something more like `organization_user_roles` / `application_user_roles`? As it stands these sound more like they are pointing to the a relationship between the resource and the user rather than to the roles and underlying privileges

api/openapi.generated.yml
```diff
@@ -3768,6 +3806,146 @@ components:
...
+            - submit_application
+            - update_form
+            - manage_agency_members
+            - get_submitted_applications
+            type:
+            - string
+    UserOrganization:
+      type: object
+      properties:
+        organization:
```
doug-s-nava, line 3921: it looks like this will result in a response object that looks like:
```
{
    user_id: '1',
    organizations: [{
        organization: {
            organization_id: '1'
        },
        organization_user_roles: [{
            role_id: '1',
            ...
        }]
    }];
}
```
might be nice if it looked more like

```
{
    user_id: '1',
    organization_roles: [{
        organization_id: '1'
        organization_user_roles: [{
            role_id: '1',
            ...
        }]
    }];
}
```

or even

```
{
    user_id: '1',
    user_roles: [{
        resource_type: 'organization',
        resource_id: '1'
        resource_user_roles: [{
            role_id: '1',
            ...
        }]
    }];
}
```

Unless I'm missing something and there is something unique about the way roles are defined between the resource types it'd make things much cleaner on the other end to be able to reason about things generically. I can work with whatever, throwing this out as a suggestion

babebe: @doug-s-nava @chouinar The first one is how it is. If we wanted to capture more information from the organization that would be easy to capture. And is much cleaner structure that reflects the relationships: user belongs to an organization and  in that context they have roles

api/src/api/users/user_schemas.py
babebe: updated the names

api/openapi.generated.yml
joshtonava: What if we did something like this for the response body. Its a generalized list of all of the roles for the user and can handle any resource type or internal

```json
{
  "data": {
    "user_id": "123e4567-e89b-12d3-a456-426614174000",
    "user_roles": [
      {
        "role_id": "446bafb9-41ee-46ac-8584-889aedcd5142",
        "role_name": "Organization Admin",
        "privileges": [
          "MANAGE_ORG_MEMBERS",
          "MANAGE_ORG_ADMIN_MEMBERS",
          "VIEW_ORG_MEMBERSHIP",
          "START_APPLICATION",
          "LIST_APPLICATION",
          "VIEW_APPLICATION",
          "MODIFY_APPLICATION",
          "SUBMIT_APPLICATION"
        ],
        "resource": {
          "resource_id": "org-123",
          "resource_type": "organization"
        }
      },
      {
        "role_id": "f1e72876-ad13-4734-be5c-b43c8cef8d67",
        "role_name": "Application Owner",
        "privileges": [
          "START_APPLICATION",
          "LIST_APPLICATION",
          "VIEW_APPLICATION",
          "MODIFY_APPLICATION",
          "SUBMIT_APPLICATION"
        ],
        "resource": {
          "resource_id": "app-789",
          "resource_type": "application"
        }
      },
      {
        "role_id": "f1e72876-ad13-4734-be5c-b43c8cef8d67",
        "role_name": "Application Owner",
        "privileges": [
          "START_APPLICATION",
          "LIST_APPLICATION",
          "VIEW_APPLICATION",
          "MODIFY_APPLICATION",
          "SUBMIT_APPLICATION"
        ],
        "resource": {
          "resource_id": "app-789",
          "resource_type": "application"
        }
      },
      {
        "role_id": "agency-role-1",
        "role_name": "Agency Program Officer",
        "privileges": ["REVIEW_APPLICATIONS", "MANAGE_COMPETITIONS"],
        "resource": {
          "resource_id": "agency-555",
          "resource_type": "agency"
        }
      },
      {
        "role_id": "internal-role-1",
        "role_name": "System Administrator",
        "privileges": ["SYSTEM_ADMIN", "MANAGE_USERS", "VIEW_ALL_DATA"],
        "resource": null
      }
    ]
  }
}

```

babebe: resolved to keeping the current structure

doug-s-nava: I know I said we were good here, but I would appreciate it if these attribute names were standardized. ex:
```
- organizations
- applications
- agencies
```
or

```
- organization_users
- application_users
- agency_users
```

or perhaps my preferred
```
- organization_privileges
- application_privileges
- agency_privileges
```

chouinar: @doug-s-nava - I had also suggested we adjust the naming of those objects to be more consistent which Bruk changed, it's now `<resource>_users` for each of them.

chouinar (approved):

api/src/api/users/user_schemas.py
doug-s-nava: i'm still seeing the old names in the diff

chouinar (approved):

PR #6523, opened 2025-09-25
[Issue #6492] Create a script to list the status of forms in an env

### LINKED ISSUE #6492: Create / modify the update-form script to be able to list the state of all forms in an environment
### Summary

Should output a neat table that has the following for each form:

- form_id
- form_name
- form_version
- updated_at
- Whether the form has any differences from what we have locally.

### Acceptance criteria
- [ ] A list form script exists that can give helpful information about the status of forms in a specified environment
- [ ] Tests added

chouinar:
## Summary
Fixes  #6492  

## Changes proposed
* Creates a script to list forms in a given environment and tell you whether they are up-to-date
* Remove the dry-run/diff logic from the update-form script (made redundant by this script)
* A bunch of refactoring to move code shared by both scripts to a central location

## Context for reviewers
The update-form script has been very helpful in checking forms and updating them, but is a bit clunky to check multiple forms. So, I made a new script that just checks everything and makes it even less effort to update forms by giving you an output.

The goal of this script is to generally run this before the update-form script, then copy the output to make it so you don't even need to find commands to run.

## Validation steps
First make sure to run `make db-seed-local` so all of your forms locally are up-to-date.

If you make sure to set the env var as described in the README, then you can use the tool.

For example, if I run `make cmd args="task list-forms --environment=dev --verbose"` I get the following output (note I manually adjusted a form for testing + added a new dummy form, you won't see these):
Screenshot 2025-09-25 at 3 52 40 PM

### DIFF
--- a/api/poetry.lock
+++ b/api/poetry.lock
@@ -2072,7 +2072,7 @@ version = "3.16.0"
 description = "A simple Python library for easily displaying tabular data in a visually appealing ASCII table format"
 optional = false
 python-versions = ">=3.9"
-groups = ["dev"]
+groups = ["main"]
 files = [
     {file = "prettytable-3.16.0-py3-none-any.whl", hash = "sha256:b5eccfabb82222f5aa46b798ff02a8452cf530a352c31bddfa29be41242863aa"},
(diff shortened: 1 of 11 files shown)


mdragon (approved):
Looks good.

Issue #6360, opened 2025-09-09
Create prototype for testing SOAP endpoints and validating schemas


chris-kuryak:
### Summary

Run the same request against the proxy and the S2S API directly and confirm the responses are identical
    - BPS can mostly do this for us, but that's a manual ad-hoc process, not something we can automate

This can potentially be done in Postman or a Script.
Documenting this can become a model for something going forward.

### Acceptance criteria

- [ ] Reviewed with engineering team at engineering sync
- [ ] Feedback from engineering addressed and included

chris-kuryak:
Will be discussed at engineering sync.

PR #6535, opened 2025-09-26
[Issue #6360] Create process for testing soap endpoints

jakobpederson:
## Summary

Fixes #ISSUE 6360

## Changes proposed

Create prototype for testing SOAP endpoints and validating schemas

## Context for reviewers

Run the same request against the proxy and the S2S API directly and confirm the responses are identical

This can potentially be done in Postman or a Script.
Documenting this can become a model for something going forward.

### DIFF
--- a/api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
+++ b/api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
@@ -1,0 +1,248 @@
+import logging
+import os
+import re
+import tempfile
+from contextlib import ExitStack
+from dataclasses import dataclass, field
+
+import click
+import requests
+from lxml import etree
+
+import src.adapters.db as db
+import src.logging
+import tests.src.db.models.factories as factories
+from src.adapters.db import PostgresDBClient
+from src.legacy_soap_api.legacy_soap_api_config import SimplerSoapAPI
+from src.legacy_soap_api.legacy_soap_api_schemas import SOAPRequest
+from src.legacy_soap_api.legacy_soap_api_utils import get_alternate_proxy_response
+from src.util.local import error_if_not_local
+
+logger = logging.getLogger(__name__)
+
+
+"""
+Add following to soap-api.env
+CERT_DATA =
+KEY_DATA =
+SOAP_URI =
+For cert and key use awk command to replace newlines
+`awk 'NF {sub(/\r/, ""); printf "%s\\n",$0;} bps_grantors.crt`
+`awk 'NF {sub(/\r/, ""); printf "%s\\n",$0;} bps_grantors.key`
+May need to run `poetry install` to update command
+To run use: `make validate-simpler-endpoints``
+"""
+
+
+SOAP_URI = os.getenv("SOAP_URI", b"")
+HEADERS = {"Content-Type": "application/xml"}
+
+
+@dataclass
+class ValidateSoapContext:
+    cert: str = field(init=False)
+    key: str = field(init=False)
+    stack: ExitStack
+    db_session: db.Session
+
+    def __post_init__(self) -> None:
+        self.cert, self.key = get_credentials(self.stack)
+
+
+def get_temp_files(stack: ExitStack) -> dict:
+    dependencies = {
+        "GrantsFundingSynopsis-V2.0": "https://trainingapply.grants.gov/apply/system/schemas/GrantsFundingSynopsis-V2.0.xsd",
+        "GrantsCommonElements-V1.0": "https://trainingapply.grants.gov/apply/system/schemas/GrantsCommonElements-V1.0.xsd",
+        "GrantsCommonTypes-V1.0": "https://trainingapply.grants.gov/apply/system/schemas/GrantsCommonTypes-V1.0.xsd",
+        "GrantsOpportunity-V1.0": "https://trainingapply.grants.gov/apply/system/schemas/GrantsOpportunity-V1.0.xsd",
+        "GrantsForecastSynopsis-V1.0": "https://trainingapply.grants.gov/apply/system/schemas/GrantsForecastSynopsis-V1.0.xsd",
+        "GrantsPackage-V1.0": "https://trainingapply.grants.gov/apply/system/schemas/GrantsPackage-V1.0.xsd",
+        "GrantsTemplate-V1.0": "https://trainingapply.grants.gov/apply/system/schemas/GrantsTemplate-V1.0.xsd",
+        "GrantsRelatedDocument-V1.0": "https://trainingapply.grants.gov/apply/system/schemas/GrantsRelatedDocument-V1.0.xsd",
+        "AgencyManagePackage-V1.0": "https://trainingapply.grants.gov/apply/system/schemas/AgencyManagePackage-V1.0.xsd",
+        "AgencyUpdateApplicationInfo-V1.0": "https://trainingapply.grants.gov/apply/system/schemas/AgencyUpdateApplicationInfo-V1.0.xsd",
+        "AgencyWebServices-V2.0": "https://trainingws.grants.gov/grantsws-agency/services/v2/AgencyWebServicesSoapPort?wsdl",
+    }
+    temp_files = {}
+    for key, url in dependencies.items():
+        filepath = f"/api/tests/validate_legacy_soap_api/cache/{key}"
+        if not os.path.isfile(filepath):
+            f = stack.enter_context(open(filepath, mode="wb"))
+            response = requests.get(url=url, timeout=10)
+            f.write(response.content)
+            f.flush()
+        temp_files[key] = f"/api/tests/validate_legacy_soap_api/cache/{key}"
+    return temp_files
+
+
+def get_xml_schema_bytes(temp_files: dict, soap_context: ValidateSoapContext) -> bytes:
+    TARGET_URI = "http://apply.grants.gov/services/AgencyWebServices-V2.0"
+    filepath = "/api/tests/validate_legacy_soap_api/cache/AgencyWebServices-V2.0"
+    f = soap_context.stack.enter_context(open(filepath, "rb"))
+    wsdl_root = etree.fromstring(f.read())
+    NS = {"wsdl": "http://schemas.xmlsoap.org/wsdl/", "xs": "http://www.w3.org/2001/XMLSchema"}
+    schema_xpath_query = f".//xs:schema[@targetNamespace='{TARGET_URI}']"
+    schema = wsdl_root.find(schema_xpath_query, namespaces=NS)
+    import_xpath_query = "./xs:import"
+    import_elements = schema.findall(import_xpath_query, namespaces=NS)
+    for import_elem in import_elements:
+        target_name = import_elem.get("namespace")
+        filename_from_url = os.path.basename(target_name)
+        if filename_from_url in temp_files:
+            import_elem.set("schemaLocation", temp_files[filename_from_url])
+    return etree.tostring(schema)
+
+
+def get_grantors_get_application_zip_request_data(gov_grants_tracking_number: str) -> bytes:
+    return (
+        '<soapenv:Envelope xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/" xmlns:agen="http://apply.grants.gov/services/AgencyWebServices-V2.0" xmlns:gran="http://apply.grants.gov/system/GrantsCommonElements-V1.0">'
+        "<soapenv:Header/>"
+        "<soapenv:Body>"
+        "<agen:GetApplicationZipRequest>"
+        f"<gran:GrantsGovTrackingNumber>{gov_grants_tracking_number}</gran:GrantsGovTrackingNumber>"
+        "</agen:GetApplicationZipRequest>"
+        "</soapenv:Body>"
+        "</soapenv:Envelope>"
+    ).encode("utf-8")
+
+
+def validate_grantor_get_application_zip_request_not_found(
+    soap_context: ValidateSoapContext,
+) -> None:
+    """
+    Get legacy xml
+    """
+    legacy_tracking_number = "GRANT60000001"
+    legacy_resp = requests.post(
+        SOAP_URI,
+        data=get_grantors_get_application_zip_request_data(legacy_tracking_number),
+        headers=HEADERS,
+        cert=(soap_context.cert, soap_context.key),
+        timeout=10,
+    )
+    assert legacy_resp.status_code == 500
+    legacy_match = re.search(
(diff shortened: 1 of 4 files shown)


api/tests/validate_legacy_soap_api/xsd/AgencyManagePackage-V1.0.xsd
```diff
@@ -0,0 +1,169 @@
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:xmime="http://www.w3.org/2005/05/xmlmime" xmlns:AgencyManagePackage="http://apply.grants.gov/system/AgencyManagePackage-V1.0" xmlns:GrantsCommonTypes="http://apply.grants.gov/system/GrantsCommonTypes-V1.0" xmlns:GrantsCommonElements="http://apply.grants.gov/system/GrantsCommonElements-V1.0" targetNamespace="http://apply.grants.gov/system/AgencyManagePackage-V1.0" elementFormDefault="qualified" attributeFormDefault="qualified" version="1.0">
```
chouinar: Could we make them be fetched by the script (and/or make a script to fetch and cache them)? That way we don't need to copy them all into our codebase / if things change we have the updates.

It's fine if it's a script that we have to hardcode the filepaths.

api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
```diff
@@ -0,0 +1,183 @@
...
+"""
+Add following to soap-api.env
+CERT_DATA_ENCODED =
+KEY_DATA_ENCODED =
+SOAP_URI =
+For cert and key use awk command to replace newlines
+`awk 'NF {sub(/\r/, ""); printf "%s\\n",$0;} bps_grantors.crt`
+`awk 'NF {sub(/\r/, ""); printf "%s\\n",$0;} bps_grantors.key`
+May need to run `poetry install` to update command
+To run use: `docker compose run --rm grants-api poetry run validate-simpler-endpoints`
```
chouinar: Could you add a line to the makefile like

```
validate-simpler-endpoints:
      $(PY_RUN_CMD) validate-simpler-endpoints
```

The `PY_RUN_CMD` resolves to either the above docker-compose command OR if you've setup your local development to work outside of Docker (complex, but way way faster) then it ends up being just the `poetry run` bit.

api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
```diff
@@ -0,0 +1,183 @@
...
+    CERT_DATA = CERT_DATA_ENCODED.replace("\\n", "\n")
+    KEY_DATA = KEY_DATA_ENCODED.replace("\\n", "\n")
+else:
+    raise Exception("No cert data")
+with tempfile.NamedTemporaryFile(delete=False) as temp_cert_file:
+    temp_cert_file.write(CERT_DATA.encode("utf-8"))
+    CERT = temp_cert_file.name
+with tempfile.NamedTemporaryFile(delete=False) as temp_key_file:
+    temp_key_file.write(KEY_DATA.encode("utf-8"))
+    KEY = temp_key_file.name
```
chouinar: I feel like we shouldn't be keeping temporary files with cert info around.

Could we move these to be in the main loop / put some of this in some sort of context we pass around? That way we can use a tempfile to make the file and let it delete it when we're done.

api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
```diff
@@ -0,0 +1,183 @@
...
+    zip_buffer = io.BytesIO(simpler_xml)
+    with zipfile.ZipFile(zip_buffer, "r") as zip_file:
+        assert zip_file.testzip() is None
+
+
+@click.command()
+def validate_simpler_endpoints() -> None:
+    print("=start=")
+    with src.logging.init("seed_local_db"):
+        logger.info("Running seed script for local DB")
```
chouinar:
```suggestion
    with src.logging.init("validate_simpler_endpoints"):
        logger.info("Running validation for SOAP endpoints")
```

api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
```diff
@@ -0,0 +1,183 @@
...
+            VALIDATIONS = {
+                "Grantor GetApplicationZipRequest Not Found": validate_grantor_get_application_zip_request_not_found,
+                "Grantor GetApplicationZipRequest": validate_grantors_get_application_zip_request,
+            }
+            for msg, validate in VALIDATIONS.items():
+                print("=" * 25)
+                print(f"{msg} started")
+                validate(db_session)
+                print(f"{msg} passed")
+    print("=end=")
```
chouinar: As written, if we hit a validation issue on the first test, would it continue running? We should have a try/except around this and at least log what we ran. Can also grab the function name so we just need a list.

I think we'd want to make this loop work a little different. Something like:
```py
for validation in validations:
    test_name = validation.__name__ # The function name
    try:
        validation(db_session)
        logger.info(f"{test_name} passed")
    except Exception:
        # This'll have the error info appended
        logger.exception(f"{test_name} failed")
```

If we wanted to be very fancy, you could even avoid setting up the list of functions by doing something like:
```py
        validations = [obj for name, obj in inspect.getmembers(sys.modules[__name__]) if inspect.isfunction(obj) and name.startswith("validate_") and name != "validate_simpler_endpoints"]
```
Effectively loop through everything defined in this file, grab the functions that start with `validate_` that aren't this function.

api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
```diff
@@ -0,0 +1,183 @@
...
+    schema_validator = etree.XMLSchema(schema_tree)
+    schema_validator.assertValid(operation_element)
+    zip_buffer = io.BytesIO(simpler_xml)
+    with zipfile.ZipFile(zip_buffer, "r") as zip_file:
+        assert zip_file.testzip() is None
+
+
+@click.command()
+def validate_simpler_endpoints() -> None:
+    print("=start=")
```
chouinar: Could we use either logging or print statements? If you want to control the formatting to be more than just text, print statements are fine, although for errors you might want to make them color-coded which is a bit trickier.

api/tests/validate_legacy_soap_api/xsd/AgencyManagePackage-V1.0.xsd
jakobpederson: I think I got this running by fetching and caching the xsds.

api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
jakobpederson: Got to leverage ExitStack and I think it's working now that they're automatically deleted once you exit the context.

api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
jakobpederson: switching to logging.

api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
```diff
@@ -0,0 +1,260 @@
...
+        stack.enter_context(src.logging.init("validate_simpler_endpoints"))
+        logger.info("Running validation for SOAP endpoints")
+        error_if_not_local()
+        db_client = PostgresDBClient()
+        db_session = stack.enter_context(db_client.get_session())
+        factories._db_session = db_session
+        VALIDATIONS = [
+            validate_grantor_get_application_zip_request_not_found,
+            validate_grantors_get_application_zip_request,
+        ]
```
chouinar: Could we define these above the function? Not a problem yet, but if we have like 30 of these in the future, would make the function very long.

api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
```diff
@@ -0,0 +1,260 @@
...
+        raise Exception("No cert data")
+    temp_cert_file = stack.enter_context(tempfile.NamedTemporaryFile(delete=True, mode="wb"))
+    temp_cert_file.write(CERT_DATA.encode("utf-8"))
+    temp_cert_file.flush()
+    CERT = temp_cert_file.name
+    temp_key_file = stack.enter_context(tempfile.NamedTemporaryFile(delete=True, mode="wb"))
+    temp_key_file.write(KEY_DATA.encode("utf-8"))
+    temp_key_file.flush()
+    KEY = temp_key_file.name
+    return CERT, KEY
```
chouinar: Some blank lines for readability/grouping.
```suggestion

    temp_cert_file = stack.enter_context(tempfile.NamedTemporaryFile(delete=True, mode="wb"))
    temp_cert_file.write(CERT_DATA.encode("utf-8"))
    temp_cert_file.flush()
    CERT = temp_cert_file.name
    
    temp_key_file = stack.enter_context(tempfile.NamedTemporaryFile(delete=True, mode="wb"))
    temp_key_file.write(KEY_DATA.encode("utf-8"))
    temp_key_file.flush()
    KEY = temp_key_file.name
    
    return CERT, KEY
```

api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
```diff
@@ -0,0 +1,260 @@
...
+    schema_file.flush()
+    schema_file_name = schema_file.name
+    schema_tree = etree.parse(schema_file_name)
+    schema_validator = etree.XMLSchema(schema_tree)
+    schema_validator.assertValid(operation_element)
+
+
+def get_credentials(stack: ExitStack) -> tuple:
+    CERT_DATA_ENCODED = os.getenv("CERT_DATA")
+    KEY_DATA_ENCODED = os.getenv("KEY_DATA")
```
chouinar: If a variable isn't static (these I guess are psuedo-static) can we avoid making them all-caps?

api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
```diff
@@ -0,0 +1,260 @@
...
+        f"<gran:GrantsGovTrackingNumber>{gov_grants_tracking_number}</gran:GrantsGovTrackingNumber>"
+        "</agen:GetApplicationZipRequest>"
+        "</soapenv:Body>"
+        "</soapenv:Envelope>"
+    ).encode("utf-8")
+
+
+def validate_grantor_get_application_zip_request_not_found(
+    db_session: db.Session, cert: str, key: str, stack: ExitStack
+) -> None:
```
chouinar, line 111: Since a decent number of tests will need a db session, cert/key and the stack for file stuff. What about moving either moving these functions into a class or make a "context" class that we pass around. Benefit here is if we need to pass another thing around you only need to change one thing instead of a bunch of different tests.

```py
class ValidateSoap:
    def __init__(self, cert: str, key: str, ...):
         self.cert = cert
         self.key = key
    
    def validate_whatever(self):
           requests.post(..., cert=(self.cert, self.key))
```
or
```py
@dataclass
class ValidateSoapContext:
     cert: str
     key: str
     ...

def validate_whatever(context: ValidateSoapContext):
      requests.post(..., cert=(context.cert, context.key))
```

api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
```diff
@@ -0,0 +1,260 @@
...
+            timeout=10,
+        ).content,
+    }
+    temp_files = {}
+    for key, dependency in dependencies.items():
+        temp_file = stack.enter_context(tempfile.NamedTemporaryFile(delete=True, mode="wb"))
+        temp_file.write(dependency)
+        temp_file.flush()
+        temp_files[key] = temp_file.name
+    return temp_files
```
chouinar, line 75: Rather than making these with temp files, could we write them to a real directory (like `tests/validate_legacy_soap_api/cache` and then add that path to the .gitignore? That way we don't need to refetch them every time but it is automated. Could then make this logic check if the file exists there already and download it if not.

api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
```diff
@@ -0,0 +1,260 @@
...
+"""
+Add following to soap-api.env
+CERT_DATA =
+KEY_DATA =
+SOAP_URI =
+For cert and key use awk command to replace newlines
+`awk 'NF {sub(/\r/, ""); printf "%s\\n",$0;} bps_grantors.crt`
+`awk 'NF {sub(/\r/, ""); printf "%s\\n",$0;} bps_grantors.key`
+May need to run `poetry install` to update command
+To run use: `docker compose run --rm grants-api poetry run validate-simpler-endpoints`
```
chouinar:
```suggestion
To run use: `make validate-simpler-endpoints`
```

api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
```diff
@@ -0,0 +1,260 @@
...
+    return CERT, KEY
+
+
+@click.command()
+def validate_simpler_endpoints() -> None:
+    with ExitStack() as stack:
+        CERT, KEY = get_credentials(stack)
+        stack.enter_context(src.logging.init("validate_simpler_endpoints"))
+        logger.info("Running validation for SOAP endpoints")
+        error_if_not_local()
```
chouinar, line 237: Init logging first so we can use it, handle credentials once we know we actually want to run.
```suggestion
    with ExitStack() as stack:
        stack.enter_context(src.logging.init("validate_simpler_endpoints"))
        logger.info("Running validation for SOAP endpoints")

        error_if_not_local()
        CERT, KEY = get_credentials(stack)

```

api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
jakobpederson: Went with the dataclass

api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
jakobpederson: makes sense

api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
jakobpederson: I like that solution

api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
jakobpederson: No problem

chouinar (approved):

PR #6540, opened 2025-09-26
[Unticketed] Give our local test users API keys by default

chouinar:
## Changes proposed
When creating users in our seed script, give them some API keys.

## Context for reviewers
As we switch over to the new API key approach, we can't rely on our statically defined API key for hitting the local DB. We might want to have a generic statically defined key (especially when the frontend switches over), but for now we can just add it to the users we're already generating in our seed script.

## Validation steps
Running `make db-seed-local` will create the keys, you can copy one of them any hit an endpoint like GET opportunity uneventfully with it. The `key_id` field we set is what you put in the auth header.


DavidDudas-Intuitial (approved):
LGTM! Thanks for the super quick turnaround :)

PR #6542, opened 2025-09-26
[Issue #6553] Add CommonGrants routes to API (from branch not fork)

### LINKED ISSUE #6553: [CommonGrants adoption] Merge CommonGrants endpoints PR into SGG codebase
### Summary

Incorporate Nava's feedback on the PR to add CommonGrants routes to SGG API and merge in the PR.

### Acceptance criteria

- [ ] PR is merged in
- [ ] CommonGrants endpoints are available in staging (maybe behind a feature flag)

### Dependencies

Must be done after https://github.com/HHS/simpler-grants-gov/issues/6275

DavidDudas-Intuitial:
## Summary

Fixes #6553 

## Changes proposed

NOTE: This PR is a cleaned-up version of https://github.com/HHS/simpler-grants-gov/pull/6379

This PR implements the CommonGrants Protocol by adding 3 new routes to the simpler.grants.gov API:

* `GET /common-grants/opportunities/`
* `GET /common-grants/opportunities/{oppId}`
* `POST /common-grants/opportunities/search/`

The implementation and tests are encapsulated in the following new directories:

* `api/src/api/common_grants/`
* `api/src/services/common_grants/`
* `api/tests/src/api/common_grants/`
* `api/tests/src/services/common_grants/`

The following existing files have been modified to support the new routes:

* `api/Makefile` – add commands to generate and validate an isolated OpenAPI spec for CG routes
* `api/local.env` – add environment variable to enable CG routes
* `api/pyproject.toml` – add dependency on `common-grants-sdk`
* `api/src/api/schemas/extension` – add new fields needed by CG routes
* `api/src/app.py` – add CG blueprint and feature flag
* `infra/api/app-config/dev.tf` – add feature flag to enable CG routes in dev env
* `infra/api/app-config/staging.tf` – add feature flag to enable CG routes in staging env

A new GitHub Actions workflow has been created to validate the implementation:

* `.github/workflows/ci-api-common-grants.yml`

Documentation for the implementation is in the following file:

* `api/src/api/common_grants/COMMON_GRANTS_INTEGRATION.md`

## Context for reviewers

The implementation in this PR leverages existing patterns:

* `@api_key_multi_auth.login_required` for authentication on CG routes
* `@flask_db.with_db_session()` for database dependency injection
* `@flask_opensearch.with_search_client()` for search client dependency injection 
* `src.api.schemas.extension` base classes for input/output schema validation
* `src.api.schemas.response_schema` and `raise_flask_error()` for error handling

## Validation steps

Installation:

```bash
poetry install
make init
make run-logs
```

Validation:

```bash
make check
make check-spec
```

Test endpoints (happy path):

```
curl -X 'GET' 'http://localhost:8080/common-grants/opportunities' -H 'accept: application/json' -H 'X-Auth: LOCAL_AUTH_12345678'
curl -X 'GET' 'http://localhost:8080/common-grants/opportunities/ff976d71-f88c-46f3-9ff5-36508666b643' -H 'accept: application/json' -H 'X-Auth: LOCAL_AUTH_12345678'
curl -X 'POST' 'http://localhost:8080/common-grants/opportunities/search' -H 'accept: application/json' -H 'Content-Type: application/json' -H 'X-Auth: LOCAL_AUTH_12345678' -d '{ "filters": { "closeDateRange": { "operator": "between", "value": { "max": "2025-12-31", "min": "2025-01-01" } }, "status": { "operator": "in", "value": ["open", "forecasted"] }, "totalFundingAvailableRange": { "operator": "between", "value": { "max": { "amount": "1000000", "currency": "USD" }, "min": { "amount": "10000", "currency": "USD" } } } }, "pagination": { "page": 1, "pageSize": 10 }, "search": "local", "sorting": { "sortBy": "lastModifiedAt", "sortOrder": "desc" } }'
```

Test endpoints (validation errors):

```
curl -X 'GET' 'http://localhost:8080/common-grants/opportunities/not-a-valid-uuid' -H 'accept: application/json' -H 'X-Auth: LOCAL_AUTH_12345678'
curl -X 'POST' 'http://localhost:8080/common-grants/opportunities/search' -H 'accept: application/json' -H 'Content-Type: application/json' -H 'X-Auth: LOCAL_AUTH_12345678' -d '{ "filters": { "closeDateRange": { "operator": "between", "value": { "max": "2025-12-31", "min": "2025-01-01" } }, "status": { "operator": "in", "value": ["open", "forecasted"] }, "totalFundingAvailableRange": { "operator": "between", "value": { "max": { "amount": "1000000", "currency": "USD" }, "min": { "amount": "10000", "currency": "USD" } } } }, "pagination": { "page": "one", "pageSize": 10 }, "search": "local", "sorting": { "sortBy": "lastModifiedAt", "sortOrder": "desc" } }'
curl -X 'POST' 'http://localhost:8080/common-grants/opportunities/search' -H 'accept: application/json' -H 'Content-Type: application/json' -H 'X-Auth: LOCAL_AUTH_12345678' -d '{ "filters": { "closeDateRange": { "operator": "between", "value": { "max": "2025-12-31", "min": "2025-01-01" } }, "status": { "0perator": "in", "value": ["open", "forecasted"] }, "totalFundingAvailableRange": { "operator": "between", "value": { "max": { "amount": "1000000", "currency": "USD" }, "min": { "amount": "10000", "currency": "USD" } } } }, "pagination": { "page": 1, "pageSize": 10 }, "search": "local", "sorting": { "sortBy": "lastModifiedAt", "sortOrder": "desc" } }'
curl -X 'POST' 'http://localhost:8080/common-grants/opportunities/search' -H 'accept: application/json' -H 'Content-Type: application/json' -H 'X-Auth: LOCAL_AUTH_12345678' -d '{ "filters": { "closeDateRange": { "operator": "between", "value": { "max": "EOY", "min": "2025-01-01" } }, "status": { "operator": "in", "value": ["open", "forecasted"] }, "totalFundingAvailableRange": { "operator": "between", "value": { "max": { "amount": "1000000", "currency": "USD" }, "min": { "amount": "10000", "currency": "USD" } } } }, "pagination": { "page": 1, "pageSize": 10 }, "search": "local", "sorting": { "sortBy": "lastModifiedAt", "sortOrder": "desc" } }'
```

## Known Issues

1. **OpenAPI Spec Validation**: Spec validation currently fails due to the following issues. These might require an update to CLI to loosen `additionalProperties` rules and allow alias for `status` property:
    - Implementation schema has extra property 'data' not defined in base schema
    - Implementation schema has extra property 'internal_request_id' not defined in base schema
    - Implementation schema has extra property 'status_code' not defined in base schema
    - Missing required property 'status' 
2. **Custom Fields**: All opportunities return empty `customFields` object `{}`
3. **Close Date Accuracy**: The close date mapping uses `summary.close_date` which may not be the correct deadline value (deadlines are stored in competitions)
4. **Sorting**: Sorting by the `updated_at` or `created_at` timestamps might not give expected results because an opportunity is defined across multiple tables each with their own timestamps, i.e. there isn't a singular timestamp that says when an opportunity was last updated


api/src/api/common_grants/common_grants_schemas.py
```diff
@@ -0,0 +1,952 @@
+"""Complete Marshmallow schemas for CommonGrants Protocol models.
+
+This file contains all Marshmallow schemas that correspond to Pydantic models,
+organized by category for better maintainability.
```
chouinar, line 4: Should we add a comment to re-evaluate when APIFlask 3.0 comes out since it'll support Pydantic directly? While we'll probably still need to rewrite some of the error parsing logic, that's a lot easier than the complex internals-rewriting we're doing for Marshmallow.

Whenever Pydantic is made available, I'll test out the complexity of starting to switch over and give you a heads up.

api/src/api/common_grants/common_grants_openapi.py
```diff
@@ -0,0 +1,109 @@
...
+IMPLEMENTATION
+--------------
+- `transform_spec_composition_to_cg()`: Main function that orchestrates the post-processing
+- `fix_schema_composition()`: Generic recursive function that fixes composition patterns
+"""
+
+from typing import Any, Dict
+
+
+def transform_spec_composition_to_cg(spec: Dict[str, Any]) -> Dict[str, Any]:
```
chouinar:
```suggestion
def transform_spec_composition_to_cg(spec: dict[str, Any]) -> dict[str, Any]:
```

api/src/services/common_grants/opportunity_service.py
```diff
@@ -0,0 +1,170 @@
...
+    @staticmethod
+    def list_opportunities(
+        search_client: search.SearchClient,
+        pagination: PaginatedBodyParams,
+    ) -> OpportunitiesListResponseSchema:
+        """Get a paginated list of opportunities."""
+
+        # Set empty search params
+        filters = OppFilters()
+        sorting = OppSorting(sort_by=OppSortBy.LAST_MODIFIED_AT)
```
chouinar, line 56: (No action needed, but a heads up) Sorting by any of the updated_at/created_at timestamps might not give results you'd expect. Because an opportunity from the perspective of a user is really made up of an opportunity/opportunity summary and more across several tables each with their own timestamps, there isn't a singular timestamp that says when an opportunity was last updated.

I have ideas for how we might consolidate the update timestamps a bit (we have a versioning table, but that info doesn't make its way to search right now), and some aspects of that probably needs to wait until we control opportunity creation.

api/src/api/common_grants/common_grants_routes.py
```diff
@@ -0,0 +1,148 @@
...
+
+    # Fetch data from service
+    try:
+        with db_session.begin():
+            response_object = CommonGrantsOpportunityService.get_opportunity(db_session, opp_uuid)
+    except ValidationError as e:
+        error_details = transform_validation_error_from_cg(e)
+        raise_flask_error(
+            HTTPStatus.UNPROCESSABLE_ENTITY, "Validation error", validation_issues=error_details
+        )
```
chouinar: Since this pattern would be needed for each common grants endpoint, what if we moved it to a decorator that wraps this function?

I did something similar with our login endpoints because they needed to always do a redirect even if there were 4xx/5xx scenarios. See: https://github.com/HHS/simpler-grants-gov/blob/main/api/src/auth/auth_utils.py#L26

Then we'd put the `try/except ValidationError` in that decorator function. If you wanted to put other error conversion logic (like the 404s) could even handle that with specific exceptions.

api/src/api/common_grants/common_grants_routes.py
```diff
@@ -0,0 +1,148 @@
...
+    description="Get a paginated list of opportunities, sorted by `lastModifiedAt` with most recent first.",
+    security=api_key_multi_auth_security_schemes,
+    responses=[HTTPStatus.OK],
+)
+@flask_opensearch.with_search_client()
+def list_opportunities(
+    search_client: search.SearchClient, query_data: dict
+) -> tuple[OpportunitiesListResponseSchema, HTTPStatus]:
+    """Get a paginated list of opportunities."""
+    add_extra_data_to_current_request_logs(query_data)
```
chouinar, line 66: Might need to flatten the info we add to this (see our search endpoint which does that), New Relic doesn't quite parse nested JSON objects in a reasonable manner, but we can see how it looks in the lower environments.

api/src/api/common_grants/common_grants_routes.py
```diff
@@ -0,0 +1,148 @@
...
+    add_extra_data_to_current_request_logs({"oppId": oppId})
+    logger.info("GET /common-grants/opportunities/{oppId}")
+
+    message_404 = "The server cannot find the requested resource"
+
+    # Validate input
+    try:
+        opp_uuid = UUID(oppId)
+    except ValueError:
+        raise_flask_error(HTTPStatus.NOT_FOUND, message_404)
```
chouinar: If you specify the type in the path, it'll basically do what you had and verify it's a UUID, and 404 otherwise.

```suggestion
@common_grants_blueprint.get("/opportunities/<uuid:oppId>")
@common_grants_blueprint.output(OpportunityResponseSchema)
@api_key_multi_auth.login_required
@common_grants_blueprint.doc(
    summary="View opportunity details",
    description="View details about an opportunity",
    security=api_key_multi_auth_security_schemes,
    responses=[HTTPStatus.OK, HTTPStatus.NOT_FOUND],
)
@flask_db.with_db_session()
def get_opportunity(
    db_session: db.Session, oppId: UUID
) -> tuple[OpportunityResponseSchema, HTTPStatus]:
    """Get a specific opportunity by ID."""
    add_extra_data_to_current_request_logs({"oppId": oppId})
    logger.info("GET /common-grants/opportunities/{oppId}")

    message_404 = "The server cannot find the requested resource"
```

api/src/api/schemas/extension/schema_fields.py
```diff
@@ -4,6 +4,7 @@
 
 from apiflask import fields as original_fields
 from marshmallow import ValidationError
+from marshmallow import fields as marshmallow_fields
```
chouinar: Could you use the `original_fields` import above? looks like the APIFlask fields are for the most part just an unmodified re-export of the marhsmallow fields (eg. just doing `from marshmallow.fields import URL as URL`).  The 3 you added are all unmodified.

api/src/api/schemas/extension/schema_fields.py
```diff
@@ -277,3 +278,31 @@ class File(original_fields.File, MixinField):
     error_mapping: dict[str, MarshmallowErrorContainer] = {
         "invalid": MarshmallowErrorContainer(ValidationErrorType.INVALID, "Not a valid file."),
     }
+
+
+class Time(marshmallow_fields.Time, MixinField):
```
chouinar: Could you add tests for these fields? We have this test: https://github.com/HHS/simpler-grants-gov/blob/main/api/tests/src/schemas/extension/test_schema_fields.py#L55

Which basically just validates a schema defined in https://github.com/HHS/simpler-grants-gov/blob/main/api/tests/src/schemas/schema_validation_utils.py - just would need to add a few fields with these types and update the data / expected error objects.

api/src/services/common_grants/opportunity_service.py
```diff
@@ -0,0 +1,170 @@
...
+    @staticmethod
+    def get_opportunity(
+        db_session: db.Session, opportunity_id: UUID
+    ) -> OpportunityResponseSchema | None:
+        """Get a specific opportunity by ID."""
+
+        # Get response data from v1 service
+        opportunity_data_v1 = get_opportunity(db_session, opportunity_id)
+        if not opportunity_data_v1:
+            return None
```
chouinar: We wouldn't return None, we'd raise a 404 from this function, if you want a different format for a 404, you'd have to catch and restructure it.

api/src/services/common_grants/transformation.py
```diff
@@ -0,0 +1,461 @@
...
+    if not v1_sort_by:
+        logger.error(
+            f"Transform failed for field `sort_by`: unexpected OppSortBy value: {cg_sort_by}"
+        )
+        v1_sort_by = OppSortBy.LAST_MODIFIED_AT
+
+    return v1_sort_by
+
+
+def _get_opportunity_summary(opportunity: Opportunity) -> OpportunitySummary | None:
```
chouinar: The opportunity model has a property called `summary` that is this utility function, could do `opportunity.summary` and get the same result

api/src/services/common_grants/transformation.py
```diff
@@ -0,0 +1,461 @@
...
+                "close_date": summary.close_date if summary else None,
+                "estimated_total_program_funding": (
+                    summary.estimated_total_program_funding if summary else None
+                ),
+                "award_ceiling": summary.award_ceiling if summary else None,
+                "award_floor": summary.award_floor if summary else None,
+                "additional_info_url": summary.additional_info_url if summary else None,
+            }
+            if summary
+            else {}
```
chouinar: Just to reduce the `if summary else None` bits, could you do this above?

```py
if summary is not None:
    summary_section = { ... }
else:
    summary_section = {}
```

api/src/api/common_grants/common_grants_openapi.py
```diff
@@ -0,0 +1,109 @@
...
+
+SOLUTION
+--------
+Post-process the generated OpenAPI spec to transform the composition patterns from APIFlask's format to the CommonGrants protocol's expected format. This is done by:
+
+1. **Schema Inlining**: Replace `$ref` references in response schemas with the actual schema definitions to enable composition pattern fixes.
+
+2. **Composition Pattern Fixing**: Recursively find and fix the problematic pattern:
+   - Find: `type: [object]` + `allOf` (or `type: "object"` + `allOf`)
+   - Fix: Remove the `type` field, keep only `allOf`
```
widal001, line 33: This is fine to get us to a place where we can merge in this PR, but seems like something we should handle in the CommonGrants CLI .

We're already doing some flattening with `allOf` in the CLI as part of the `cg check spec` and assuming the output of APIFlask is a compliant OpenAPI spec, we should figure out what's happening in the CLI that's causing this to fail.

Not saying we need to change this code, but we should create a backlog ticket to investigate this further on the CLI.

api/src/services/common_grants/opportunity_service.py
widal001: It looks like the CommonGrants router is already checking for `None` returned by this service and raising a 404 at the router/transport layer.

api/src/services/common_grants/opportunity_service.py
```diff
@@ -0,0 +1,170 @@
...
+
+        # Transform response data to CG model
+        opportunity_data_cg = transform_opportunity_to_cg(opportunity_data_v1)
+        opportunity_response = OpportunityResponse(
+            status=HTTPStatus.OK,
+            message="Success",
+            data=opportunity_data_cg,
+        )
```
widal001: Suggestion (non-blocking): The pattern I’ve seen in most API codebases is to keep services HTTP-agnostic, i.e. returning domain data or raising exceptions, and then letting the controller/router handle status codes and response formatting.

That keeps services reusable and aligns with a [hexagonal architecture approach](https://www.geeksforgeeks.org/system-design/hexagonal-architecture-system-design). Specifically it could be helpful if we decide we need/want to pull the CommonGrants routes into a separate APIFlask app with different error handling.

I know a big part of these services _is_ focused on massaging response formats, but I wonder if there's a way to keep these more clearly separated?

That definitely doesn't have to happen before this is merged in, but moving forward I'd like to align on the dividing line between business/transformation logic handled by the CommonGrants service and formatting handled by the router, so that we can keep that bright line clear and easier to maintain. 

My $0.02 on that dividing line would be to manage all transformations that happen within the `data` or `items` response attribute in the service and then limit transformations of the response wrapper to the route layer.

.github/workflows/ci-api-common-grants.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Install Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: "20.x"
+
+      - name: Install CommonGrants CLI
+        run: npm install -g @common-grants/cli
+
+      - name: Generate and validate spec
+        run: make check-spec
```
widal001, line 34: Since this is failing right now because of the error response adjustments, and it won't pass unless we make `cg check spec` [more configurable](https://github.com/HHS/simpler-grants-protocol/issues/331), It might be worth at least adding a `continue-on-error: true` to this step.

Ideally we'd check the output of this step for any failures outside of these expected ones, but that might be overkill for now.

api/src/services/common_grants/opportunity_service.py
chouinar: I mean that the `get_opportunity` method called here can't return `None` as it either returns an opportunity or raises an exception. You couldn't hit the None check in the route itself

api/src/services/common_grants/opportunity_service.py
chouinar: The general approach we've taken so far is to put as little as possible in the API routes themselves, really just grabbing whatever is relevant (request, db session) and putting everything else in the service layer including raising errors. Skimming the hexagonal architecture doc, I've seen systems that followed those ideas and always found them incredibly difficult to parse and develop, perhaps there is a way you've seen it work, but I would see something like a simple GET endpoint go from being quite easy to follow to something monumentally complex. I tried to build a dedicated persistence layer on a prior project and everyone hated working with it.

I'm curious if you've got any examples of systems that have built this pattern?

api/src/api/common_grants/common_grants_schemas.py
DavidDudas-Intuitial: Thanks. The common-grants-sdk already has a full set of Pydantic models which we will probably switch to if and when simpler.grants.gov eventually supports Pydantic.

api/src/api/common_grants/common_grants_openapi.py
DavidDudas-Intuitial: Good catch. Thanks!

api/src/api/common_grants/common_grants_routes.py
DavidDudas-Intuitial: Great suggestion, thank you!

api/src/services/common_grants/opportunity_service.py
DavidDudas-Intuitial: Good catch, you're right

api/src/services/common_grants/transformation.py
DavidDudas-Intuitial: Good to know! Thanks

api/src/services/common_grants/transformation.py
DavidDudas-Intuitial: Fair enough. Done.

.github/workflows/ci-api-common-grants.yml
DavidDudas-Intuitial: Good call. Done.

api/src/api/schemas/extension/schema_fields.py
DavidDudas-Intuitial: For sure. Thanks for verifying that they're unmodified.

api/src/api/schemas/extension/schema_fields.py
DavidDudas-Intuitial: Sure thing. Done.

api/src/services/common_grants/opportunity_service.py
DavidDudas-Intuitial: Ack. Thanks for the heads up

api/src/services/common_grants/opportunity_service.py
widal001: Here are some conversations about the importance of separating the service and transport layer so that services are only handling business logic and HTTP exceptions are raised by the routers (transport layer).

- [This selected answer](https://stackoverflow.com/a/73211663/7338319)
- The [comments on this question](https://softwareengineering.stackexchange.com/q/393307) and [the first answer](https://softwareengineering.stackexchange.com/a/418707)
- [This answer about avoiding tightly coupled services and controllers](https://stackoverflow.com/a/77169063/7338319)

I can try to dig up some good examples in Python code!

But all that said, consistency with the rest of the codebase is more important, so if we're following a different pattern elsewhere it's probably more helpful to remain consistent with that.

api/src/api/common_grants/common_grants_routes.py
DavidDudas-Intuitial: Thanks for the tip. This endpoint does not have nested json input, but the search endpoint does, so I will flatten there.

api/src/services/common_grants/opportunity_service.py
widal001: Here are some examples of what that pattern might look like in Python code:

- [Example of a user service returning a user or none](https://github.com/vinicinhus/fastapi-clean-architecture-template/blob/fb98d6bd42eb8d4fc98a6d394c972b7937eff437/app/services/user_service.py#L103-L114), then a route wrapping that in `404` error if a user isn't found
- [Example of an article service](https://github.com/almeidaraphael/realworld-fastapi-ddd/blob/8673b6140d76f95071c8cf87dd973516a6aa725b/app/api/articles.py#L35-L52) returning domain-specific values or errors when trying to create an article, with the router dumping the successful result to a specific model or catching the error translating them to HTTP specific errors using a separate utility.
- [Flask example of a todo service creating a todo](https://github.com/microsoft/cookiecutter-python-flask-clean-architecture/blob/97cd0d33fa512bdb7a320a4bc7767cd0cb7257f2/%7B%7Bcookiecutter.repo_name%7D%7D/src/api/controllers/todo.py#L15-L24), with the controller formatting the response and status code

But again, I'd rather be consistent with the _current_ API design, so think of these just as potential references if we wanted to switch to the pattern described above -- probably not something we'd do in the short term (or maybe ever)

api/src/api/common_grants/common_grants_routes.py
DavidDudas-Intuitial: I like this idea. Done in [9635691](https://github.com/HHS/simpler-grants-gov/pull/6542/commits/9635691537efd5474b3e7a3f601ba971c6281eef)

api/src/services/common_grants/opportunity_service.py
DavidDudas-Intuitial: @widal001 This is reasonable feedback. I just pushed [b147556](https://github.com/HHS/simpler-grants-gov/pull/6542/commits/b147556eef30fa43171ab9d572b24182c8224d7a) to make the routes instantiate the response objects (`OpportunityResponse`, `OpportunitiesListResponse`, `OpportunitiesSearchResponse`) instead of the service, and the service methods return `OpportunityBase` objects.

DavidDudas-Intuitial:
@chouinar @widal001 I have addressed all of your comments and suggestions. Can you please review and approve?

widal001 (dismissed):
Thanks for addressing the points we raised @DavidDudas-Intuitial 

Seems like you've resolved all of the major blockers and have done a great job flagging the known issues in the PR which will can be addressed separately either through changes to CommonGrants protocol or CLI tool, or future PRs.

LGTM pending final pass by @chouinar !

chouinar (approved):
LGTM!

DavidDudas-Intuitial:
@chouinar @widal001  I had to fix a merge conflict which dismissed your review, will you please approve again?

chouinar (approved):

Issue #5874, opened 2025-08-06
For the API, use the CPU and Memory we are provisioning


chris-kuryak:
### Summary

This will ensure that we have a way to grow into the capacity we have provisioned.

### Acceptance criteria

TBD

PR #6545, opened 2025-09-29
[Issue #5874] Update cpu and mem config for ecs task definition

prasnava:
## Summary

Updated the memory and cpu config is to be relative to the total vcpu and memory profile of the Fargate instance. 

## Context for reviewers

Currently, the full vCPU and memory of the app task definition is not being utilized. 

## Validation steps

After deploying the changes in the lower environment we should inspect the task definition to check the full cpu and mem is utilized in the task definition.

### DIFF
--- a/infra/modules/service/variables.tf
+++ b/infra/modules/service/variables.tf
@@ -242,26 +242,26 @@ variable "is_temporary" {
 }
 
 variable "fluent_bit_memory" {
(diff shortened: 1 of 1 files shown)


prasnava:
This branch is successfully deployed in api-training.

chouinar (approved):

PR #6549, opened 2025-09-29
[ISSUE #6339] Organization endpoints - List organization members

### LINKED ISSUE #6339: Organization endpoints - List organization members
### Summary

Implement an endpoint to fetch all members of an organization. Returns users along with their roles and privileges within the organization.

https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/1977221432/Tech+Spec+User+Roles+on+the+Backend#POST-%2Forganizations%2F%3Aorganization_id%2Fusers

### Tech Notes

- Required Permission: `view_org_membership` privilege
- Response should include user roles and privileges for each member
- Consider adding user name filtering when user names are added to the system

### Acceptance criteria

- Implement `POST /organizations/:organization_id/users`
- Returns users with with their organization roles and privileges
- Enforces `view_org_membership` privilege check
- Include unit tests

joshtonava:
## Summary

Fixes #6339 

## Changes proposed

- Added `POST /organizations/:organization_id/users` endpoint to retrieve organization information by ID, including associated SAM.gov entity data.
- Privilege-based authorization using VIEW_ORG_MEMBERSHIP

## Context for reviewers

This endpoint will enable organization management features in the frontend, specifically:
- Display section showing all organization users
- Show each member's email, roles, and permissions
- Enable organizations members to see who has access to what functionality

## Validation steps

To run the tests 

```bash
make test args="-vv tests/src/api/organizations_v1/test_organization_users_list_routes.py"
```

### DIFF
--- a/api/src/api/organizations_v1/organization_schemas.py
+++ b/api/src/api/organizations_v1/organization_schemas.py
@@ -32,9 +32,58 @@ class OrganizationDataSchema(Schema):
     )
 
 
+class OrganizationUserRoleSchema(Schema):
+    """Schema for user role information with privileges"""
+
+    role_id = fields.UUID(
+        metadata={
+            "description": "Role unique identifier",
+            "example": "123e4567-e89b-12d3-a456-426614174000",
+        }
+    )
+    role_name = fields.String(
+        metadata={"description": "Role name", "example": "Organization Admin"}
+    )
+    privileges = fields.List(
+        fields.String(),
+        metadata={
+            "description": "List of privileges for this role",
+            "example": ["manage_org_membership"],
+        },
+    )
+
+
+class OrganizationMemberSchema(Schema):
+    """Schema for organization member information"""
+
+    user_id = fields.UUID(
+        metadata={
+            "description": "User unique identifier",
+            "example": "123e4567-e89b-12d3-a456-426614174000",
+        }
+    )
+    email = fields.String(
+        allow_none=True,
+        metadata={"description": "User email from login.gov", "example": "user@example.com"},
+    )
+    roles = fields.List(
+        fields.Nested(OrganizationUserRoleSchema),
+        metadata={"description": "User roles in this organization"},
+    )
+
+
 class OrganizationGetResponseSchema(AbstractResponseSchema):
     """Schema for GET /organizations/:organization_id response"""
 
     data = fields.Nested(
         OrganizationDataSchema, metadata={"description": "Organization information"}
+    )
+
+
+class OrganizationUsersResponseSchema(AbstractResponseSchema):
+    """Schema for POST /organizations/:organization_id/users response"""
+
+    data = fields.List(
+        fields.Nested(OrganizationMemberSchema),
+        metadata={"description": "List of organization members"},
     )
--- a/api/src/services/organizations_v1/organization_users_list.py
+++ b/api/src/services/organizations_v1/organization_users_list.py
@@ -1,0 +1,84 @@
+from typing import Any
+from uuid import UUID
+
+from sqlalchemy import select
+from sqlalchemy.orm import joinedload
+
+import src.adapters.db as db
+from src.db.models.user_models import OrganizationUser, OrganizationUserRole, User
+from src.services.organizations_v1.get_organization import get_organization_and_verify_access
+
+
+def _fetch_organization_users(db_session: db.Session, organization_id: UUID) -> list[User]:
+    """Fetch all users in an organization with their roles and privileges eagerly loaded."""
+    stmt = (
+        select(User)
+        .join(OrganizationUser)
+        .options(
+            # Only load the organization membership for the specific organization we're querying
+            joinedload(User.organizations.and_(OrganizationUser.organization_id == organization_id))
+            .joinedload(OrganizationUser.organization_user_roles)
+            .joinedload(OrganizationUserRole.role),
+            joinedload(User.linked_login_gov_external_user),
+        )
+        .where(OrganizationUser.organization_id == organization_id)
+    )
+    users = db_session.execute(stmt).scalars().unique().all()
+    return list(users)
+
+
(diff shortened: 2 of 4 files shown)


api/src/services/organizations_v1/organization_users_list.py
```diff
@@ -0,0 +1,84 @@
...
+
+    Returns:
+        list[dict]: List of user data with roles and privileges
+
+    Raises:
+        FlaskError: 404 if organization not found, 403 if access denied
+    """
+
+    # Check if user has VIEW_ORG_MEMBERSHIP privilege for this organization
+    get_organization_and_verify_access(db_session, user, organization_id)
```
chouinar, line 48: Hmm, not specific to this PR, but should GET organization have that privilege, or should something else? I'd be confused if I couldn't see an org I was in because I was missing a privilege about viewing org membership. Should we add a `view_organization` privilege instead?

api/src/services/organizations_v1/organization_users_list.py
```diff
@@ -0,0 +1,84 @@
...
+            # Only load the organization membership for the specific organization we're querying
+            joinedload(User.organizations.and_(OrganizationUser.organization_id == organization_id))
+            .joinedload(OrganizationUser.organization_user_roles)
+            .joinedload(OrganizationUserRole.role),
+            joinedload(User.linked_login_gov_external_user),
+        )
+        .where(OrganizationUser.organization_id == organization_id)
+    )
+    users = db_session.execute(stmt).scalars().unique().all()
+    return list(users)
```
chouinar: Would this be simpler if we did the query against the OrganizationUser table? Then we'd not need to do the filter in the load to only fetch org users for a specific user.

I think the iteration we do on the data might be simpler since the org users is what we want to iterate over anyways.

---
Can we use `selectinload`? `joinedload` has caused some issues in the past - I've found that some calls just lock up entirely with `joinedload` across a few tables once the data gets slightly bigger.

api/src/services/organizations_v1/organization_users_list.py
```diff
@@ -0,0 +1,84 @@
...
+    users = _fetch_organization_users(db_session, organization_id)
+
+    return [
+        {
+            "user_id": user.user_id,
+            "email": (
+                user.linked_login_gov_external_user.email
+                if user.linked_login_gov_external_user
+                else None
+            ),
```
chouinar: There is a property on the user table that does this.
```suggestion
            "email": user.email,
```

But, if someone has a null email, is that something we want to do something about?

api/src/api/organizations_v1/organization_schemas.py
```diff
@@ -32,9 +32,58 @@ class OrganizationDataSchema(Schema):
     )
 
 
+class OrganizationUserRoleSchema(Schema):
```
chouinar: Could we just make this a generic `UserRoleSchema`, other than maybe the example values, the object would be the same.

api/src/api/organizations_v1/organization_schemas.py
```diff
@@ -32,9 +32,58 @@ class OrganizationDataSchema(Schema):
...
+    role_name = fields.String(
+        metadata={"description": "Role name", "example": "Organization Admin"}
+    )
+    privileges = fields.List(
+        fields.String(),
+        metadata={
+            "description": "List of privileges for this role",
+            "example": ["manage_org_membership"],
+        },
+    )
```
chouinar:
```suggestion
    privileges = fields.List(
        fields.Enum(Privilege),
        metadata={
            "description": "List of privileges for this role",
            "example": [Privilege.MANAGE_ORG_MEMBERSHIP],
        },
    )
```

api/src/services/organizations_v1/organization_users_list.py
joshtonava: Yea now that you mention it, the get organization endpoint's permission does feel a little off.

api/src/services/organizations_v1/organization_users_list.py
joshtonava: This should be good now

api/src/services/organizations_v1/organization_users_list.py
joshtonava: I switched to using the user email.

What cases would allow a user to have a null email?

api/src/api/organizations_v1/organization_schemas.py
joshtonava: I moved this to a shared location as well. Let me know if this is overkill

api/src/services/organizations_v1/organization_users_list.py
chouinar: Could we make a ticket to follow-up on that?

api/src/services/organizations_v1/organization_users_list.py
chouinar: Not yet, but late this quad/early next quad. When we need to support API keys for organizations/agencies, the way we'll implement that will be that we create a record in the user + api key table, but won't create anything in the linked external user table. Effectively these keys will be users (so all roles logic just works for them), but they won't ever log in via login.gov and thus won't have an email set in that table.

For now, that won't happen, short of us manually inserting user into the DB, the only way a user is created is via the login.gov login flow.

api/src/services/organizations_v1/organization_users_list.py
joshtonava: Yes here is the link https://github.com/HHS/simpler-grants-gov/issues/6552

chouinar (approved):

api/src/services/organizations_v1/organization_users_list.py
joshtonava: We may want to return a little more info about the user like API user and give back a different identifier for them for the frontend to display

Issue #6342, opened 2025-09-09
Organization endpoints - List available organization roles


joshtonava:
### Summary

Implement an endpoint to fetch all roles available for an organization. Currently returns core organization roles, but designed to support custom roles in the future.

https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/1977221432/Tech+Spec+User+Roles+on+the+Backend#POST-%2Forganizations%2F%3Aorganization_id%2Froles%2Flist

### Tech Notes
- Required Permission: The spec says "TODO privilege" - likely `view_org_membership` or any role in the org
- Should return only roles with organization role_type
- In the future will include organization-specific custom roles
- Response should include role details and associated privileges

### Acceptance criteria

- Implement `POST /organizations/:organization_id/roles/list`
- Returns all core organization roles (`ORG_ADMIN`, `ORG_MEMBER`). Returns role_id, role_name, and privileges for each role.
- Returns appropriate error codes (403 for forbidden, 404 for not found)
- Include unit tests

babebe:
Currently there is no link between a role to a resource to create custom roles so this will only return the core roles for the organization.

PR #6556, opened 2025-09-30
[Issue #6520] List available organization roles

babebe:
## Summary

Fixes  for #[6342](https://github.com/HHS/simpler-grants-gov/issues/6342) 

## Changes propose
`POST` endpoint for returning organization roles given the user has the `view_org_membership` privilege on the requested organization. 
For the time being we only have core roles and those are being returned.
Schema for the response
Test added


api/src/api/organizations_v1/organization_schemas.py
```diff
@@ -67,3 +68,19 @@ class OrganizationUsersResponseSchema(AbstractResponseSchema):
         fields.Nested(OrganizationMemberSchema),
         metadata={"description": "List of organization members"},
     )
+
+
+class RoleSchema(Schema):
```
joshtonava: I have something really similar here

`api/src/api/schemas/shared_schema.py`

babebe: @josh Sure.  Renamed it to `RoleSchema`, so it is more generic..

api/src/services/organizations_v1/list_organization_roles.py
```diff
@@ -0,0 +1,25 @@
...
+    # We are currently only grabbing the core roles
+    get_organization_and_verify_access(db_session, user, organization_id)
+    return get_organization_roles()
+
+
+def get_organization_roles() -> list[Role]:
+    """Returns the organization core roles."""
+    core_org_roles = [ORG_ADMIN, ORG_MEMBER]
+
+    return core_org_roles
```
chouinar: Could we have this do the DB query to fetch the roles? That way when we add more core roles, we don't have to remember to add it here. Should be something roughly like:

```py
db_session.execute(select(Role).join(LinkRoleRoleType).where(LinkRoleRoleType.role_type == RoleType.ORGANIZATION)).scalars().all()
```

PR #6561, opened 2025-10-01
switch applications on by default

doug-s-nava:
## Summary

Fixes / Work for unticketed

## Changes proposed

sets application / form related behavior and pages to be visible by default. Should only affect local development since this has been switched on in deployed envs already

## Validation steps

1. run `npm run dev` on this branch
3. search for "pilot"
4. click into the opportunity detail
5. _VERIFY_: you see a button for starting an application


mdragon (approved):
Approved. Kicked the checks since they failed for something goofy so assuming they'll pass this time.

PR #6562, opened 2025-10-01
HOTFIX: 20251001 banner

andycochran:
## Summary

This is a hotfix to add a temporary banner for Oct 01, 2025.

## Changes proposed

* Hardcode a message in the `Layout` that displays below the site header

## Context for reviewers

As this is temporary, the content is hardcoded and does not make use of i18n. This change is sandwiched in a comment so the code can easily be removed. 

## Validation steps

Load the app and you should see the following message on a yellow background.

image

### DIFF
--- a/frontend/src/components/Layout.tsx
+++ b/frontend/src/components/Layout.tsx
@@ -37,6 +37,28 @@ export default function Layout({ children, locale }: Props) {
         <LoginModalProvider>
           <Header locale={locale} />
           <main id="main-content" className="border-top-0">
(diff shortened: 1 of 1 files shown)


mdragon:
Code looks good, I'm going to test it out real quick. But let's just make sure we don't merge until we have final approval.

chrisw-nava:
Looks good to me! Second Matt that we should wait to merge until we get word. Thanks for the quick turnaround, @andycochran

mdragon (approved):
Verified change, awaiting final deploy approval.

PR #6564, opened 2025-10-01
[Issue #6490] Modify update-form script to not need form instruction ID

### LINKED ISSUE #6490: Make the form-instruction-id parameter optional, and use the value fetched from the environment as a default instead of null.
### Summary

We want to update the update form task: https://github.com/HHS/simpler-grants-gov/blob/main/api/src/task/forms/update_form_task.py

This task takes in a form instruction ID that points to a form instruction record. Except for when we first set up a form, this form instruction record isn't too likely to change as it's just pointing to a file on s3. Rather than needing to pass that ID in every single time, we should instead make it so if the field is not passed in, it'll fetch the existing form and reuse its ID. That means we can leave it out most of the time and not negatively affect the form logic.

### Acceptance criteria
- [ ] Script updated to make it so if form instruction ID is missing, it will default to whatever exists in that environment.

chouinar:
## Summary
Fixes #6490

## Changes proposed
* On each statically defined form, add a form instruction ID
* Adjust the seed script to make a form instruction record for anything with a form instruction ID
* Adjust the form scripts to expect form instruction ID on the form we have locally

## Context for reviewers
Because form instructions work a little different than other fields on the form, needing the ID to point to a different record you have to manually setup, I had made it so the script required you pass it in. But that's really not needed, so made it so we define the ID on the forms themselves and just use that when diffing/updating forms. Most of the logic changes are just simplifying that.

Note that in the db-seed-local changes, I made it so it generates txt files instead of PDFs. Making a PDF with a little bit of content is way too complex for a seed script. Surprisingly there isn't just a way to say "make a PDF with a little text" that isn't a 5 step process or requires several other PDF libraries.

## Validation steps
Checking each environment, these changes (as expected) don't show any changes to the forms as we already had the form instruction ID set on all of them.

Had to do a bit of adjusting to make sure `make db-seed-local` would setup the instruction forms as expected, but the script works when re-running it.

The output of `make cmd args="task list-forms --environment=prod --verbose"`
```
+------------------------------------------------------------+---------+-------------------------+----------------+
| Form Name                                                  | Version | Last Updated            | Changed Fields |
+------------------------------------------------------------+---------+-------------------------+----------------+
| Assurances for Non-Construction Programs (SF-424B)         | 1.1     | 2025-08-21 16:10:27 UTC |                |
| Project Abstract Summary                                   | 2.0     | 2025-08-21 16:10:41 UTC |                |
| Project Narrative Attachment Form                          | 1.2     | 2025-08-21 16:10:47 UTC |                |
| Budget Narrative Attachment Form                           | 1.2     | 2025-08-21 16:10:53 UTC |                |
| Application for Federal Assistance (SF-424)                | 4.0     | 2025-09-24 16:35:21 UTC |                |
| Budget Information for Non-Construction Programs (SF-424A) | 1.0     | 2025-09-24 16:35:38 UTC |                |
| Disclosure of Lobbying Activities (SF-LLL)                 | 2.0     | 2025-09-24 16:35:49 UTC |                |
+------------------------------------------------------------+---------+-------------------------+----------------+
```


jakobpederson (dismissed):
LGTM

PR #6569, opened 2025-10-01
[Issue #6563] Update co-planning sync script

### LINKED ISSUE #6563: [Co-planning] Update GitHub to Fider sync
### Summary

Update the process of syncing GitHub issues to Fider to allow us to pull in multiple sections from a GitHub issue to fit the new issue template.

### Acceptance criteria

- [ ] We can populate the Fider ticket with content from multiple sections in the GitHub issue body

widal001:
## Summary

Updates the GitHub to Fider Co-planning sync script to allow us to populate Fider posts with multiple sections from the GitHub issue body.

Fixes #6563 

## Changes proposed

- Fixes a few small type checking warnings
- Adds support for passing multiple sections to the `--issue-sections` flag
- Updates the CI script to use all of the sections in the GitHub issue template for co-planning proposals

## Context for reviewers

In the "Internal - Co-planning Proposal" issue template, we replaced the "Summary" section with a new set of sections that we want to be copied over. The previous syncing pattern only allowed us to copy over a single section (or would default to the entire ticket if that section wasn't found).

This new pattern allows us to copy over multiple sections and includes any matching sections in the Fider post. If no matching sections are found, it still defaults to use the entire issue body.

## Validation steps

Here's an example of what would be copied over to Fider based on the PR run:

Screenshot 2025-10-01 at 1 03 22 PM

### DIFF
--- a/.github/linters/coplanning/utils.py
+++ b/.github/linters/coplanning/utils.py
@@ -95,37 +95,39 @@ def format_title(title: str) -> str:
 def format_post_description(
     url: str,
     description: str,
-    section: str | None = None,
+    sections: list[str],
 ) -> str:
-    """Format the post description by extracting content from a specific section or the entire description.
+    """Format the post description by extracting content from multiple sections.
 
     Args:
         url: The GitHub issue URL
         description: The issue description text
-        section: Optional section header to extract (e.g., "Description", "Context").
-                If None, extracts content between first and second ### headers.
-                If specified but not found, uses the entire description.
+        sections: List of section headers to extract (e.g., ["Summary", "Context"]).
+                 If specified sections not found, uses the entire description.
     """
-    # Set the appropriate pattern based on whether a section is specified
-    if section:
+    # Extract content from multiple sections
+    extracted_sections = []
+
+    for section in sections:
         pattern = rf"^###\s+{re.escape(section)}\s*\n+(?P<content>.*?)(?=\n###|\Z)"
-    else:
-        # Original behavior: extract text between first and second ### headers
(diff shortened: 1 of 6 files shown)


.github/linters/coplanning/utils.py
```diff
@@ -72,11 +72,14 @@ def make_request(
         with urllib.request.urlopen(req) as response:
             return json.loads(response.read().decode())
     except urllib.request.HTTPError as e:
-        err(f"HTTP request failed: {e.code} - {e.reason}")
+        err(f"HTTP request failed: {e.code} - {e.reason}", exit=True)
+        return {}  # For type checking, never reached due to sys.exit() in err()
```
chouinar: What if we split the case where `err` should exit into a separate function like:
```py
def err(message: str) -> None:
    logger.error(message)

def err_and_exit(message: str) -> typing.Never:
    err(message)
    sys.exit(1)
```
Can add a return of `typing.Never` on the one that exits so the type-checker knows that function won't return so you don't need the `return {}` here.

chouinar (approved):
LGTM, but an idea if you want to slightly simplify the error cases

.github/linters/coplanning/utils.py
widal001: I like that pattern! I'll try it out and if it's an easy swap I'll re-request a review

widal001: Worked like a charm! I added this pattern in this commit: [feat(coplanning): Adds err_and_exit() logger func](https://github.com/HHS/simpler-grants-gov/pull/6569/commits/87c82ad24db539ef3802ef23294be04ec2c8db35)

chouinar (approved):

Issue #5873, opened 2025-08-06
Modify our run task command to be able to pass CPU and Memory Override commands


chris-kuryak:
### Summary

TBD

### Acceptance criteria

TBD

PR #6571, opened 2025-10-01
[Issue #5873] Update scheduled jobs cpu and mem override options

prasnava:
## Summary

Currently the schedule jobs terraform config doesn't support adding cpu and mem configurations for each job. This change adds the ability to pass in optional cpu and mem params. The default is configured to 1024 cpu and 2048 ram.

## Changes proposed

Updated the scheduled_jobs .tf with the new cpu and mem params.

```
scheduled_jobs_config = {
    training = {
      cpu = 1024
      mem = 4096
    }
    prod = {
      cpu = 1024
      mem = 4096
    }
  }
```

## Validation steps

After deploying the changes the scheduled cron jobs / step functions should function as normal.

### DIFF
--- a/infra/modules/service/variables.tf
+++ b/infra/modules/service/variables.tf
@@ -281,8 +281,8 @@ variable "scheduled_jobs" {
     task_command        = list(string)
     schedule_expression = string
     state               = string
-    cpu                 = optional(number, 256)
-    mem                 = optional(number, 512)
+    cpu                 = optional(number, 1024)
+    mem                 = optional(number, 2048)
   }))
   default = {}
 }


infra/modules/service/variables.tf
```diff
@@ -281,6 +281,8 @@ variable "scheduled_jobs" {
     task_command        = list(string)
     schedule_expression = string
     state               = string
+    cpu                 = optional(number, 256)
+    mem                 = optional(number, 512)
```
chouinar: Wouldn't this lower what the jobs run with currently? I think right now it's 768 (.75 cpu) and 2gb (4gb in prod). Could we set the defaults to be higher / use everything we have available?

prasnava:
This branch has been deployed in training, verified the step function: 
api-training-load-search-opportunity-data is working. The execution id is: 0868dffa-085f-453a-95a4-0c50f65972e3

PR #6573, opened 2025-10-02
[Issue #6477] Build Header XML

### LINKED ISSUE #6477: Build Header XML
### Summary

The XML file contains a header that includes data. We need to create a header that meets the needs of existing grants.gov XML output.

---

We want to generate XML that will represent a header. This header looks like:
```xml
<header:GrantSubmissionHeader glob:schemaVersion="1.0"
	xmlns:glob="http://apply.grants.gov/system/Global-V1.0"
	xmlns:header="http://apply.grants.gov/system/Header-V1.0">
	<glob:HashValue glob:hashAlgorithm="SHA-1">XzrJcEg0I6Xszvu3Zt+1bGVAVg4=</glob:HashValue>
	<header:AgencyName>Simpler Grants.gov</header:AgencyName>
	<header:OpportunityID>SIMP-TESTING-092225</header:OpportunityID>
	<header:OpportunityTitle>Testing Forms</header:OpportunityTitle>
	<header:CompetitionID>SIMP-TESTING-090225-COMPID</header:CompetitionID>
	<header:OpeningDate>2025-09-02</header:OpeningDate>
	<header:ClosingDate>2025-09-30</header:ClosingDate>
	<header:SubmissionTitle>Test App</header:SubmissionTitle>
</header:GrantSubmissionHeader>
````
The namespaces can be considered static, and the fields within can be pulled from the application or the competition/opportunity associated with the competition. We should assume that when we generate this, the input is an application object.
* `AgencyName` - `opportunity.agency_name`  -if that value is null, use the `opportunity.agency_code`
* `OpportunityID` - `opportunity.opportunity_number`
* `OpportunityTitle` - `opportunity.opportunity_title`
* `CompetitionID` - `competition.public_competition_id`
* `OpeningDate` - `competition.opening_date` - see format in example above
* `ClosingDate` - `competition.closing_date` - see format in example above
* `SubmissionTitle` - `application.application_name`
* `CFDANumber` - `competition.opportunity_assistance_listing.assistance_listing_number` - Note that not every competition has a `opportunity_assistance_listing` record.
* `HashValue` - How this is calculated isn't clear, we might need to test a bit. This might be a hash of the java object which we wouldn't be able to mimic, or it's a hash of the rest of the XML? It's not clear, it might be best to circle back on this later, calculate something from the other values.

If a field is null, it should be excluded from the header entirely.

### Acceptance criteria

- [ ] XML generated for header from an application
- [ ] Utility built that can do this, does not need to be connected to any process yet

### PR DESCRIPTION
**Pull request by mikehgrantsgov**
## Summary

Fixes #6477 

## Changes proposed

Added `HeaderXMLGenerator` which, given an `Application`, will produce header information similar to the example in the ticket, e.g.

```
<header:GrantSubmissionHeader glob:schemaVersion="1.0"
	xmlns:glob="http://apply.grants.gov/system/Global-V1.0"
	xmlns:header="http://apply.grants.gov/system/Header-V1.0">
	<glob:HashValue glob:hashAlgorithm="SHA-1">XzrJcEg0I6Xszvu3Zt+1bGVAVg4=</glob:HashValue>
	<header:AgencyName>Simpler Grants.gov</header:AgencyName>
	<header:OpportunityID>SIMP-TESTING-092225</header:OpportunityID>
	<header:OpportunityTitle>Testing Forms</header:OpportunityTitle>
	<header:CompetitionID>SIMP-TESTING-090225-COMPID</header:CompetitionID>
	<header:OpeningDate>2025-09-02</header:OpeningDate>
	<header:ClosingDate>2025-09-30</header:ClosingDate>
	<header:SubmissionTitle>Test App</header:SubmissionTitle>
</header:GrantSubmissionHeader>
```

Add a flask CLI task to execute this via command line. 
Add tests

## Context for reviewers

Note that `HashValue` calculation may change (also noted in code). We didn't spend too much time on that value and instead combine a few fields from the application knowing it will change in the future.

## Validation steps

Create an application / competition using`make console`
Run `docker compose run --rm grants-api poetry run flask task generate-submission-xml --application-id <application_id>`
See generated output.
```
<?xml version='1.0' encoding='utf-8'?>
<header:GrantSubmissionHeader xmlns:header="http://apply.grants.gov/system/Header-V1.0" xmlns:glob="http://apply.grants.gov/system/Global-V1.0" glob:schemaVersion="1.0">
  <glob:HashValue glob:hashAlgorithm="SHA-1">+B5PB5GV2zH9m/PlADORTnd9WX4=</glob:HashValue>
  <header:AgencyName>Bureau of Reclamation - Mid-Pacific Region</header:AgencyName>
  <header:OpportunityID>USDOJ-OJP-BJS-32-093</header:OpportunityID>
  <header:OpportunityTitle>Thomas, Crane and Porter 1986 award</header:OpportunityTitle>
  <header:CompetitionID>ABC-134-56789</header:CompetitionID>
  <header:OpeningDate>2025-09-15</header:OpeningDate>
  <header:ClosingDate>2025-10-12</header:ClosingDate>
  <header:SubmissionTitle>Important wrong bit.</header:SubmissionTitle>
  <header:CFDANumber>30.115</header:CFDANumber>
</header:GrantSubmissionHeader>
```
Also see new unit tests.

### CODE REVIEW

File: api/src/services/xml_generation/header_generator.py [hunk 1]
```diff
@@ -0,0 +1,117 @@
...
+            self.opportunity.opportunity_number or "",
+            str(self.competition.competition_id) if self.competition.competition_id else "",
+        ]
+        hash_data = "|".join(hash_data_parts)
+
+        # Calculate SHA-1 hash
+        sha1_hash = hashlib.sha1(hash_data.encode("utf-8"), usedforsecurity=False)
+
+        # Return base64-encoded hash
+        import base64
```
chouinar: Could we add this to the top of the file?

api/src/task/xml_generation/generate_xml_cli.py
```diff
@@ -0,0 +1,116 @@
...
+    "--pretty/--no-pretty",
+    default=True,
+    help="Pretty print the XML output (default: True)",
+)
+@click.option(
+    "--output-file",
+    help="Optional file path to write the XML to (prints to stdout if not specified)",
+)
+@flask_db.with_db_session()
+def generate_xml(
```
chouinar: Fine if we keep it if it helps with building it, but the only place we'll be using this is in our greater XML generation logic which'll only ever get called when we create the application zip.

api/src/services/xml_generation/header_generator.py
```diff
@@ -0,0 +1,117 @@
...
+    # TODO: this is likely going to be changed based on additional research into how the hash value is calculated
+    def _calculate_hash_value(self) -> str | None:
+        # Create a string from key application data
+        hash_data_parts = [
+            str(self.application.application_id) if self.application.application_id else "",
+            self.application.application_name or "",
+            self.opportunity.opportunity_number or "",
+            str(self.competition.competition_id) if self.competition.competition_id else "",
+        ]
+        hash_data = "|".join(hash_data_parts)
```
chouinar, line 109: I couldn't confirm after digging through the code for a while, but my best guess at what the hash value is in grants.gov is either a Java class object hash OR some sort of hash of an XML object since it seems like it sometimes gets set upstream of the java logic.

I suppose one way we could try to get it to match would be to put the same data into XML and see if we can produce the same result, but at the same time.. if it's such an obtuse concept, we might just be fine putting whatever in it.

That's all to say, this is fine.

api/src/services/xml_generation/header_generator.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+class HeaderXMLGenerator:
+
+    def __init__(self, application: Application):
+        self.application = application
+        self.competition = application.competition
+        self.opportunity = self.competition.opportunity
+
+    def generate_header_xml(self, pretty_print: bool = True) -> str:
```
chouinar: Since a lot of this overlaps with the footer work, especially some of the util methods for setting up the XML itself, should we just make this something we'll extend to add a `generate_footer_xml` method as well?

api/src/task/xml_generation/generate_xml_cli.py
mikehgrantsgov: Yeah I'll remove this. I added it quickly to get something running from the command line vs just in unit tests. Figured it might be useful but we can revive it later from history if it's needed/useful.

api/src/services/xml_generation/header_generator.py
mikehgrantsgov: I made the class more generic and added a stub for the footer method. Should make the footer implementation much more straightforward.

Issue #6544, opened 2025-09-29
Create Read Only Versions of Apply Form Widgets


doug-s-nava:
### Summary

Some widgets are having problems displaying properly when used to generate PDFs, likely due to some complex render timing issues.

Rather than work around these timing issues, we have decided to solve this and any future similar issues by building read only versions of all apply form widgets.

The change will probably look like:
* building static versions of each widget that take in as little info as possible and do as little as possible to render the field with the provided value
* updating each main widget to take a "readOnly" prop and render a static version of the widget when set to true
* setting the readOnly prop to true on each widget when rendered from the /print route, likely by passing the flag through into FormFields into calls to `renderWidget`

We could build the switch for static / writeable functionality into the existing widget components themselves, but some of the widget components are already too big and adding more logic would likely only confuse things more.

This change should handle the errors mentioned in https://github.com/HHS/simpler-grants-gov/issues/6138 and  https://github.com/HHS/simpler-grants-gov/issues/6141 and https://github.com/HHS/simpler-grants-gov/issues/6142

**No changes should be made to handle the budget form widgets specifically as part of this ticket.** A separate ticket should be created for the budget form, or reuse https://github.com/HHS/simpler-grants-gov/issues/6506

### Acceptance criteria

- [x] issues mentioned in https://github.com/HHS/simpler-grants-gov/issues/6138 and  https://github.com/HHS/simpler-grants-gov/issues/6141 and https://github.com/HHS/simpler-grants-gov/issues/6142 are resolved
- [ ] no new issues with generated form pdfs are created

chris-kuryak:
UAT not passed.
Tagging @doug-s-nava @acouch to look into. Let me know if you would like me to create separate, new tickets for the below issues instead.

Everything looks good except:
1. The Certification Agreement (Section 21) question no longer appears on form SF-424. So I cannot check it.
2. There is an issue with radio button selection not showing up in the PDF.
Screenshot from SF-424:

<img width="583" height="284" alt="Image" src="https://github.com/user-attachments/assets/97f873d4-c1e9-4380-b623-2178f84630c1" />

Screenshot from generated PDF:

<img width="579" height="265" alt="Image" src="https://github.com/user-attachments/assets/3f95c696-73e8-404b-8f5b-2281e4b276fa" />

chris-kuryak:
Per feedback, a new bug ticket is being made for the radio button issue above, which will allow us to close out this ticket.

#6771

Issue #6142, opened 2025-08-27
PDF generation of Abstract summary field appears as expected


chris-kuryak:
### Summary

The Project Abstract Summary form includes a free response field. When the number of characters exceeds the length/size of the box, it cuts off the any remaining words in the generated PDF.

**Expected behavior:** The generated PDF shows all text in the Project Abstract Summary in a readable presentation.

### Acceptance criteria

- [x] Expected behavior achieved
- [x] UAT passed

chris-kuryak:
Checked. It did not work as expected.

chris-kuryak:
Working as expected now. UAT passed.

<img width="829" height="712" alt="Image" src="https://github.com/user-attachments/assets/26592e76-0ceb-4af9-8bc2-f554859b6bb3" />

Issue #6141, opened 2025-08-27
[BUG] - The generated PDF does not show drop-down selections in a readable way


chris-kuryak:
### Summary

**Current Behavior:** Drop-down selections are shown in a way that sometimes obstructs other fields. Multi-select drop-down selections (longer ones) block other fields completely and do not clearly show which fields were selected.

**Expected Behavior:** Drop-down selections and Multi-select Drop-down selections are shown on the generated PDF clearly and without obstructing other fields.

Example:

<img width="617" height="557" alt="Image" src="https://github.com/user-attachments/assets/d7b9a82c-3059-405a-ac1f-8086e9165db2" />

### Acceptance criteria

- [x] Expected behavior achieved
- [x] Reviewed and approved by Product

chris-kuryak:
My: Could use inline styles?

chris-kuryak:
From Aaron: https://www.npmjs.com/package/react-to-print

From Michael: https://docraptor.com/documentation/article/1067818-css-stylesheets

From Michael: https://docraptor.com/documentation/article/1067832-enabling-javascript

chris-kuryak:
Michael to try flipping feature flag for Javascript to see if it fixes the issue.

chris-kuryak:
From Aaron, NGROK is a potential other solution

chouinar:
Here is what it now looks like - format is a bit off, but it's showing the value at least.
<img width="774" height="764" alt="Image" src="https://github.com/user-attachments/assets/afe2d49c-0a65-45a7-bea6-9ffb19a4efab" />

chris-kuryak:
Drop downs working as expected. UAT passed.

Issue #6138, opened 2025-08-27
[BUG] - The generated PDF does not show filename of attached file


chris-kuryak:
### Summary

Current behavior: The attachment filename is not shown on the PDF generated from a submitted application.

Expected behavior: The PDF generated from a submitted application needs to show the attachment filename in all attachment fields.

<img width="605" height="447" alt="Image" src="https://github.com/user-attachments/assets/675ad55f-9d7e-42b8-afb7-bbe6f83d2afa" />

### Acceptance criteria

- [ ] Attachment filename is shown in submitted application PDF
- [ ] Reviewed with Product for approval

chris-kuryak:
UAT passed. Filename's showing as expected now.

<img width="824" height="229" alt="Image" src="https://github.com/user-attachments/assets/8b125f11-c512-4391-b87d-75e8f58d2615" />

PR #6574, opened 2025-10-02
[Issue #6544] Update print page with print widgets

acouch:
## Summary

Fixes / Work for #6544 #6138 #6141 #6142

## Changes proposed

Creates new print widgets and print component to make the print view look better. Updates uiSchema for print to display printed versions. Ignores `multiField`, so no update to budget forms.

### Figma

image

https://www.figma.com/design/zPlDWcYM396szw2yL3UXvk/Simpler-Grants-2025?node-id=5219-57461&m=dev

### Screenshots

image

image

## Validation steps

1. Create a new application, fill out the forms.
2. Create a token in the API folder `make generate-internal-token`
3. Replace `workspace/applications` with `print` in form URL
4. View the form (I use https://modheader.com/) to add the internal token so I can see it in the web browser

### DIFF
--- a/frontend/src/components/applyForm/widgets/PrintAttachmentWidget.tsx
+++ b/frontend/src/components/applyForm/widgets/PrintAttachmentWidget.tsx
@@ -1,0 +1,120 @@
+// file adapted from https://github.com/rjsf-team/react-jsonschema-form/blob/main/packages/core/src/components/templates/BaseInputTemplate.tsx
+// changes made to include USWDS and allow to functional as non-reactive form field
+import { FormContextType, RJSFSchema, StrictRJSFSchema } from "@rjsf/utils";
+import { useApplicationAttachments } from "src/hooks/ApplicationAttachments";
+
+import { useEffect, useState } from "react";
+import { FormGroup } from "@trussworks/react-uswds";
+
+import { FieldErrors } from "src/components/applyForm/FieldErrors";
+import { UploadedFile, UswdsWidgetProps } from "src/components/applyForm/types";
+import { DynamicFieldLabel } from "./DynamicFieldLabel";
+
+/** The `TextWidget` component uses the `BaseInputTemplate`.
+ *
+ * @param props - The `WidgetProps` for this component
+ */
+function PrintAttachmentWidget<
+  T = unknown,
+  S extends StrictRJSFSchema = RJSFSchema,
+  F extends FormContextType = never,
+>({
+  id,
+  required,
+  schema,
+  value,
+  rawErrors = [],
+  formClassName,
+}: UswdsWidgetProps<T, S, F>) {
+  const { title, description } = schema as S;
+  const { attachments } = useApplicationAttachments();
+  const [uploadedFiles, setUploadedFiles] = useState<UploadedFile[]>([]);
+
+  useEffect(() => {
+    let parsedValue: string[] = [];
+
+    if (Array.isArray(value)) {
+      parsedValue = value as string[];
+    } else if (typeof value === "string") {
+      try {
+        // casting doesn’t fully satisfy the linter because it treats schema as possibly any underneath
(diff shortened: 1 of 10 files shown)


frontend/src/components/applyForm/widgets/PrintAttachmentWidget.tsx
```diff
@@ -0,0 +1,101 @@
+// file adapted from https://github.com/rjsf-team/react-jsonschema-form/blob/main/packages/core/src/components/templates/BaseInputTemplate.tsx
+// changes made to include USWDS and allow to functional as non-reactive form field
```
doug-s-nava: [sp] 
```suggestion
// changes made to include USWDS and allow to function as non-reactive form field
```

frontend/src/components/applyForm/widgets/PrintAttachmentWidget.tsx
```diff
@@ -0,0 +1,101 @@
...
+import { FormContextType, RJSFSchema, StrictRJSFSchema } from "@rjsf/utils";
+import { useApplicationAttachments } from "src/hooks/ApplicationAttachments";
+
+import { useEffect, useState } from "react";
+import { FormGroup } from "@trussworks/react-uswds";
+
+import { UploadedFile, UswdsWidgetProps } from "src/components/applyForm/types";
+
+/** The `TextWidget` component uses the `BaseInputTemplate`.
```
doug-s-nava: not sure what this comment means - maybe we can pull it out?

frontend/src/components/applyForm/widgets/PrintAttachmentWidget.tsx
```diff
@@ -0,0 +1,101 @@
...
+  schema,
+  value,
+  rawErrors = [],
+  formClassName,
+}: UswdsWidgetProps<T, S, F>) {
+  const { title } = schema as S;
+  const { attachments } = useApplicationAttachments();
+  const [uploadedFiles, setUploadedFiles] = useState<UploadedFile[]>([]);
+
+  useEffect(() => {
```
doug-s-nava: since the state on here isn't changing in the print version, this calculation could just be done in the render without the state. if we wanted to go real crazy this could be moved out to a utility as well, to share with the non-print version
```suggestion
 let parsedValue: string[] = [];

  if (Array.isArray(value)) {
    parsedValue = value as string[];
  } else if (typeof value === "string") {
    try {
      // casting doesn’t fully satisfy the linter because it treats schema as possibly any underneath
      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
      const parsed = JSON.parse(value);
      if (
        Array.isArray(parsed) &&
        parsed.every((item) => typeof item === "string")
      ) {
        parsedValue = parsed;
      }
    } catch {
      console.warn("Invalid JSON string for attachment:", value);
    }
  }

  if (parsedValue.length > 0) {
    const hydrated = parsedValue.map((uuid) => {
      const match = attachments?.find(
        (a) => a.application_attachment_id === uuid,
      );
      return {
        id: uuid,
        name: match?.file_name || "(Previously uploaded file)",
      };
    });

    uploadedFiles = hydrated;
  }
```

acouch: I removed useEffect, thanks.

frontend/src/components/applyForm/widgets/PrintAttachmentWidget.tsx
acouch: I removed this.

frontend/src/components/applyForm/widgets/PrintAttachmentWidget.tsx
acouch: I removed these. Don't think they are useful for the print widgets.

Issue #6145, opened 2025-08-27
Section Numbers on the Simpler Forms should Match Section Numbers on Form Instructions


chris-kuryak:
### Summary

The numbering of sections in the Simpler forms that we have created for the pilot do not always match the numbering of sections in the PDF instructions document.

Since we are expecting users to reference the PDF instruction documents as they fill out the forms on Simpler, the numbering of the fields should match. At minimum, the section title numbers (1, 2, 3, etc) should match 

Here is an example -
**Simpler's SF-424:**

<img width="714" height="610" alt="Image" src="https://github.com/user-attachments/assets/534312ed-8660-447e-96f5-69880a0f05a2" />

**PDF version of the form:** 

<img width="1097" height="622" alt="Image" src="https://github.com/user-attachments/assets/8fc27e99-0e70-43aa-8cee-ed4a41f02506" />

**Corresponding PDF instruction document:**

<img width="660" height="566" alt="Image" src="https://github.com/user-attachments/assets/ec3baaee-7da1-40cd-89b3-780936558c35" />

### Acceptance criteria

SF-424 ([PDF Form](https://apply07.grants.gov/apply/forms/readonly/SF424_4_0-V4.0.pdf), [PDF Instructions](https://apply07.grants.gov/apply/forms/instructions/SF424_4_0-V4.0-Instructions.pdf)):
- [x] Section numbers are present in the form UI and form sidebar nav
- [x] Section numbers match those on the PDF instructions
- [x] Section 8f is included as it's own section with title "Name and contact information of person to be contacted on matters involving this application"
- [x] UAT passed

SF-424a ([PDF Form](https://apply07.grants.gov/apply/forms/readonly/SF424A-V1.0.pdf), [PDF Instructions](https://apply07.grants.gov/apply/forms/instructions/SF424A-V1.0-Instructions.pdf)):
- [x] Section numbers are present in the form UI and form sidebar nav
- [x] Section numbers match those on the PDF instructions
- [x] UAT passed

SF-LLL ([PDF Form](https://apply07.grants.gov/apply/forms/readonly/SFLLL_2_0-V2.0.pdf), [PDF Instructions](https://apply07.grants.gov/apply/forms/sample/SFLLL_2_0-V2.0.pdf)):
- [x] Section numbers are present in the form UI and form sidebar nav
- [x] Section numbers match those on the PDF instructions
- [x] Add Title and Telephone number fields to Section 11
[...]

PR #6584, opened 2025-10-03
[ISSUE #6145]: updates the sections in sf-lll to match documents provided (1/3 parts)

myduong-navapbc:
## Summary

Work for #I6145

## Context for reviewers

Re-number of the sections in SF-LLL form

## Validation steps

3. navigate to `search`
4. search for `pilot`
6. sign in
7. start new application
9. click Disclosure of Lobbying Activities (SF-LLL)

### DIFF
--- a/api/src/form_schema/forms/sflll.py
+++ b/api/src/form_schema/forms/sflll.py
@@ -352,20 +352,22 @@ FORM_JSON_SCHEMA = {
 FORM_UI_SCHEMA = [
     {
         "type": "section",
-        "label": "1. Background",
-        "name": "Background",
-        "children": [
-            {"type": "field", "definition": "/properties/federal_action_type"},
-            {"type": "field", "definition": "/properties/federal_action_status"},
-            {"type": "field", "definition": "/properties/report_type"},
-        ],
+        "label": "1. Type of Federal Action",
+        "name": "Type of Federal Action",
+        "children": [{"type": "field", "definition": "/properties/federal_action_type"}],
     },
     {
         "type": "section",
-        "label": "2. For Material Change Only",
-        "name": "For Material Change Only",
+        "label": "2. Status of Federal Action",
+        "name": "Status of Federal Action",
+        "children": [{"type": "field", "definition": "/properties/federal_action_status"}],
+    },
+    {
+        "type": "section",
+        "label": "3. Report Type",
+        "name": "Report Type",
         "children": [
-            # Material Change
+            {"type": "field", "definition": "/properties/report_type"},
             {"type": "field", "definition": "/properties/material_change_year"},
             {"type": "field", "definition": "/properties/material_change_quarter"},
             {"type": "field", "definition": "/properties/last_report_date"},
@@ -373,10 +375,9 @@ FORM_UI_SCHEMA = [
     },
     {
         "type": "section",
-        "label": "3. Name and Address of Reporting Entity",
+        "label": "4. Name and Address of Reporting Entity",
         "name": "Name and Address of Reporting Entity",
         "children": [
-            # Reporting Entity
             {"type": "field", "definition": "/properties/reporting_entity/properties/entity_type"},
             {"type": "field", "definition": "/properties/reporting_entity/properties/tier"},
             {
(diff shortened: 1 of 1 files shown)


api/src/form_schema/forms/sflll.py
```diff
@@ -353,31 +353,32 @@
...
     },
     {
         "type": "section",
-        "label": "3. Name and Address of Reporting Entity",
+        "label": "4. Name and Address of Reporting Entity",
         "name": "Name and Address of Reporting Entity",
         "children": [
-            # Reporting Entity
             {"type": "field", "definition": "/properties/reporting_entity/properties/entity_type"},
             {"type": "field", "definition": "/properties/reporting_entity/properties/tier"},
```
doug-s-nava, line 383: never noticed this before, but this field doesn't seem to be in the pdf. 🤷 guess we keep it in

doug-s-nava: also the `organization name` field is just called `name` in the pdf and instructions, should we change that? `organization name` is more helpful, but I don't know how important it is for us to have an exact match against directions. See also `zip` vs `zip / postal code` and `congressional district if known` vs `congressional district`

api/src/form_schema/forms/sflll.py
```diff
@@ -447,23 +448,36 @@
     },
     {
         "type": "section",
-        "label": "5. Details",
-        "name": "Details",
+        "label": "6. Federal Department/Agency",
```
doug-s-nava, line 451: directions have likely outdated names here (CFDA number and title), and in a different order. I think what's here is better, and it matches pdf, but wanted to call it out, especially since this means we really shouldn't always be trusting the instructions.

api/src/form_schema/forms/sflll.py
```diff
@@ -508,10 +522,9 @@
     },
     {
         "type": "section",
-        "label": "7. Individual Performing Services (including address if different from No. 6)",
-        "name": "Individual Performing Services (including address if different from No. 6)",
+        "label": "10b. Individual Performing Services",
```
doug-s-nava, line 525: love that the instructions have `zip code` here, but `zip` for other addresses...

api/src/form_schema/forms/sflll.py
```diff
@@ -556,10 +569,9 @@
     },
     {
         "type": "section",
-        "label": "8. Signature",
+        "label": "11. Signature",
```
doug-s-nava, line 572: seems to be missing title and telephone number. do we need to add those, or are they left out on purpose?

doug-s-nava, line 572: what happened to `Information requested through
this form is authorized by title
31 U.S.C. section 1352. This
disclosure of lobbying activities
is a material representation of
fact upon which reliance was
placed by the tier above when
the transaction was made or
entered into. This disclosure is
required pursuant to 31 U.S.C.
1352. This information will be
reported to the Congress semi-
annually and will be available
for public inspection. Any
person who fails to file the
required disclosure shall be
subject to a civil penalty of not
less than $10,000 and not more
than $100,000 for each such
failure.`?

doug-s-nava (commented):
a few questions here. In general I'd question how closely we should be trying to adhere to the instructions, since they are outdated and inconsistent. The sections labels in many situations are also redundant with the field labels. That can be a larger question for later, though

api/src/form_schema/forms/sflll.py
chouinar: `tier` is in the PDF, but it only appears if you select SubAwardee.
Image

doug-s-nava: tricky

api/src/form_schema/forms/sflll.py
myduong-navapbc: added the title and telephone

myduong-navapbc: i added this in as a description for the section

api/src/form_schema/forms/sflll.py
```diff
@@ -575,11 +592,15 @@
...
+                "definition": "/properties/signature_block/properties/name/properties/suffix",
             },
             {
                 "type": "field",
-                "definition": "/properties/signature_block/properties/name/properties/suffix",
+                "definition": "/properties/signature_block/properties/title",
+            },
+            {
+                "type": "field",
+                "definition": "/properties/signature_block/properties/telephone",
```
doug-s-nava, line 603: The labels for telephone and title in the PDF and instructions say "Telephone No." and "Title" - can update ours to match (currently they read "Signature Telephone Number" and "Signature Title"

PR #6586, opened 2025-10-03
[Issue #6494] Update migration for `form_type_id` foreign key [2/2]

### LINKED ISSUE #6494: Add a sgg_version and is_deprecated column to the form table and a connect to a new lk_form_type table
### Summary

We want a new lk_form_type table that stores a type of form. The form types we would store correspond to the unversioned forms:

- SF424
- SF424A
- SF424B
- SFLLL
- ProjectNarrativeAttachment
- BudgetNarrativeAttachment
- ProjectAbstractSummary

---
We want to add the following columns:
* `sgg_version` - str, nullable
* `is_deprecated` - boolean, nullable

We should set these values on all of our defined forms, we should have the version be `1.0` for everything, and all of them have `False` for `is_deprecated`. Later tickets will handle actually getting these values merged to the right environment.

### Acceptance criteria
- [ ] Columns added
- [ ] Factories updated
- [ ] Form definitions updated

mikehgrantsgov:
## Summary

Fixes #6494 

## Changes proposed

Update migration to account for foreign key relationship.

## Context for reviewers

There was an error running the migration on the merge of the first PR for this ticket. This fixes the issue.

## Validation steps

run `make db-migrate` successfully.

### DIFF
--- a/api/src/db/models/competition_models.py
+++ b/api/src/db/models/competition_models.py
@@ -218,11 +218,10 @@ class Form(ApiSchemaTable, TimestampMixin):
     form_rule_schema: Mapped[dict | None] = mapped_column(JSONB)
     json_to_xml_schema: Mapped[dict | None] = mapped_column(JSONB)
 
-    form_type: Mapped[FormType] = mapped_column(
+    form_type: Mapped[FormType | None] = mapped_column(
         "form_type_id",
         LookupColumn(LkFormType),
         ForeignKey(LkFormType.form_type_id),
-        nullable=True,
     )
     sgg_version: Mapped[str | None]
     is_deprecated: Mapped[bool | None]


chouinar:
@mikehgrantsgov - we'll need to adjust the definition of the column on the form table itself. It needs to have the same pattern as our other lookup table foreign keys.

api/src/db/models/competition_models.py
```diff
@@ -218,7 +218,12 @@ class Form(ApiSchemaTable, TimestampMixin):
     form_rule_schema: Mapped[dict | None] = mapped_column(JSONB)
     json_to_xml_schema: Mapped[dict | None] = mapped_column(JSONB)
 
-    form_type: Mapped[FormType | None] = mapped_column(LookupColumn(LkFormType))
+    form_type: Mapped[FormType] = mapped_column(
+        "form_type_id",
+        LookupColumn(LkFormType),
+        ForeignKey(LkFormType.form_type_id),
+        nullable=False,
+    )
```
chouinar: This field has to be nullable, at least for now since they aren't set in the values we already have in the environments. Even though your other PR updated our statically defined values, those aren't updated automatically.

api/src/db/migrations/versions/2025_10_03_update_form_type_id.py
```diff
@@ -0,0 +1,30 @@
...
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "5060847aba92"
+down_revision = "5f8e99a128b7"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
```
chouinar: We shouldn't need a second migration, I think this just was making it non-nullable which we wouldn't want.

mikehgrantsgov: Makes sense, running it again with one migration

api/src/db/models/competition_models.py
mikehgrantsgov: Makes sense, and made this nullable

api/src/db/models/competition_models.py
```diff
@@ -218,7 +218,12 @@ class Form(ApiSchemaTable, TimestampMixin):
     form_rule_schema: Mapped[dict | None] = mapped_column(JSONB)
     json_to_xml_schema: Mapped[dict | None] = mapped_column(JSONB)
 
-    form_type: Mapped[FormType | None] = mapped_column(LookupColumn(LkFormType))
+    form_type: Mapped[FormType] = mapped_column(
+        "form_type_id",
+        LookupColumn(LkFormType),
+        ForeignKey(LkFormType.form_type_id),
+        nullable=True,
+    )
```
chouinar: Can keep nullable=True, but want to make it `Mapped[FormType | None]` otherwise type checkers will get confused when we start using it.

```suggestion
    form_type: Mapped[FormType | None] = mapped_column(
        "form_type_id",
        LookupColumn(LkFormType),
        ForeignKey(LkFormType.form_type_id),
    )
```

chouinar (approved):

Issue #6340, opened 2025-09-09
Organization endpoints - Update user roles


joshtonava:
### Summary

Implement an endpoint to update a user's roles within an organization. This is a full replacement operation (PUT) where all specified roles replace existing roles for that user in the organization.

https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/1977221432/Tech+Spec+User+Roles+on+the+Backend#PUT-%2Forganizations%2F%3Aorganization_id%2Fusers%2F%3Auser_id

### Tech Notes

- This is a PUT operation - all existing roles are replaced with the provided role_ids
- Should validate all role_ids exist and are valid for organizations
- Should validate target user exists in the organization
- **Required Permission**:
    - `manage_org_members` privilege (minimum)
    - `manage_org_admin_members` for admin role assignment (pending clarification)

### Acceptance criteria

- Implement `PUT /organizations/:organization_id/users/:user_id`
- Returns updated user with new roles
- Validates `role_ids` exist and are organization-type roles
- Returns appropriate error codes (403 for forbidden, 404 user not in organization)
- Include unit tests

joshtonava:
@joshtonava 
Can we confirm this is the general role id or the join role id

joshtonava:
Lets rework this ticket to make sure its feasible for the frontend 
cc: @doug-s-nava, @joshtonava

joshtonava:
In order for this to work we may need to provide an optional resource type and optional resource id along with the role id

joshtonava:
We are not going to use the  `manage_org_admin_members` privilege as part of the enforcement logic for the time being as it overly complicates the solution.

cc @babebe

PR #6591, opened 2025-10-03
[Issue #6340] Update user roles

babebe:
## Summary

Fixes / Work for #6340  

## Changes proposed

`PUT /v1/organizations/:organization_id/users/:user_id` endpoint for updating organization user roles.

- Ensures acting user has correct privileges to assign roles
- Validates target user is a member of the organization
- Updates target user's roles accordingly
- Adds request and response schemas
- Adds tests for permission checks and role updates


api/src/services/organizations_v1/update_user_organization_roles.py
```diff
@@ -0,0 +1,71 @@
...
+    org_user = validate_organization_user(db_session, target_user_id, organization)
+    # Permission checks
+    role_ids = data["role_ids"]
+    if not can_access(user, {Privilege.MANAGE_ORG_MEMBERS}, organization):
+        raise_flask_error(403, "Forbidden")
+
+    if any(role_id in ADMIN_ROLES for role_id in role_ids):
+        logger.info("Admin role assignment requested")
+        if not can_access(user, {Privilege.MANAGE_ORG_ADMIN_MEMBERS}, organization):
+            raise_flask_error(403, "Forbidden")
```
chouinar: This isn't something we can do with our current implementation. While we have a role called an admin role, there is not such thing as an "admin role". When future roles get added, including user-defined roles, this would not work.

The `manage_org_admin_members` privilege has been something I've been asking about since it was first added to the list of privileges as it's not something we have a way to enforce with what we've built. We either need to get an answer (from who, I don't know), or get rid of it.

api/src/services/organizations_v1/update_user_organization_roles.py
```diff
@@ -0,0 +1,71 @@
...
+    db_session: db.Session, user_id: UUID, organization: Organization
+) -> OrganizationUser:
+    """Validate that the user is a member of the specified organization"""
+    org_user = db_session.execute(
+        select(OrganizationUser)
+        .where(OrganizationUser.organization_id == organization.organization_id)
+        .where(OrganizationUser.user_id == user_id)
+    ).scalar_one_or_none()
+    if not org_user:
+        raise_flask_error(403, "Forbidden")
```
chouinar: This shouldn't be a forbidden error, it should be a 404 as the user you want to update doesn't exist.

api/src/services/organizations_v1/update_user_organization_roles.py
```diff
@@ -0,0 +1,71 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+ADMIN_ROLES = [str(ORG_ADMIN_ID)]
+
+
+def get_role(db_session: db.Session, role_ids: list[UUID]) -> Sequence[Role]:
+    """Retrieve Role objects matching the given role_ids"""
+    return db_session.execute(select(Role).where(Role.role_id.in_(role_ids))).scalars().all()
```
chouinar: We'd need this query to do a bit more filtering.

The roles need to also:
* Be an org role
* Be either a core role OR role owned by the org (we don't have the latter built yet)

If any of the roles passed in don't exist, we need to error (not sure if that would be a 404 or 422? Probably 422?)

api/src/services/organizations_v1/update_user_organization_roles.py
```diff
@@ -0,0 +1,71 @@
...
+
+    org_user_roles = [OrganizationUserRole(organization_user=org_user, role=role) for role in roles]
+
+    for our in org_user_roles:
+        db_session.add(our)
+
+    org_user.organization_user_roles = org_user_roles
+    db_session.add(org_user)
+
+    return org_user.roles
```
chouinar: Does this handle deletes? Can't recall if doing `org_user.organization_user_roles = something` would handle adding/deleting. If it does, does it completely replace and make a new record even if something was already there (ie. change the `updated_at` timestamps?). If it does that, we'd want to be a bit more explicit so the timestamps don't change. If I update a user to whatever they already are, nothing should actually change in the DB.

api/src/services/organizations_v1/update_user_organization_roles.py
```diff
@@ -0,0 +1,71 @@
...
+    """Update roles of a user in an organization, after validating permissions and membership."""
+    logger.info("Attempting to update roles for user")
+
+    # Lookup organization and validate target user exists
+    organization = get_organization(db_session, organization_id)
+    org_user = validate_organization_user(db_session, target_user_id, organization)
+    # Permission checks
+    role_ids = data["role_ids"]
+    if not can_access(user, {Privilege.MANAGE_ORG_MEMBERS}, organization):
+        raise_flask_error(403, "Forbidden")
```
chouinar, line 68: Can we move this to just after we fetch the org? Should always check privileges as soon as possible, don't want to hit other error cases and give info to folks without access.

api/src/api/organizations_v1/organization_schemas.py
```diff
@@ -73,3 +74,13 @@ class OrganizationListRolesResponseSchema(AbstractResponseSchema):
     """Schema for POST /organizations/:organization_id/roles/list response"""
 
     data = fields.List(fields.Nested(RoleSchema), metadata={"description": "Role information"})
+
+
+class OrganizationUpdateUserRolesRequestSchema(Schema):
+    role_ids = fields.List(fields.String(required=True), validate=Length(min=1))
```
chouinar: I actually noticed a few schemas recently that had UUIDs as strings, can we adjust those? That way the example/validation we do works right.
```suggestion
    role_ids = fields.List(fields.UUID(required=True), validate=Length(min=1))
```

api/src/services/organizations_v1/update_user_organization_roles.py
```diff
@@ -0,0 +1,91 @@
...
+    OrganizationUserRole,
+    Role,
+    User,
+)
+from src.services.organizations_v1.get_organization import get_organization
+
+logger = logging.getLogger(__name__)
+
+
+def get_role(db_session: db.Session, role_ids: set[str]) -> Sequence[Role]:
```
chouinar: Nit: To go along with the request schema change / pluralize function for readability
```suggestion
def get_roles(db_session: db.Session, role_ids: set[uuid.UUID]) -> Sequence[Role]:
```

api/src/services/organizations_v1/update_user_organization_roles.py
```diff
@@ -0,0 +1,91 @@
...
+    # Permission checks
+    new_role_ids = set(data["role_ids"])
+    if not can_access(user, {Privilege.MANAGE_ORG_MEMBERS}, organization):
+        raise_flask_error(403, "Forbidden")
+
+    # Validate target user exists
+    org_user = validate_organization_user(db_session, target_user_id, organization)
+
+    existing_role_ids = {our.role_id for our in org_user.organization_user_roles}
+    if existing_role_ids == new_role_ids:
```
chouinar: Just organizing to group related code a bit more.
```suggestion
    if not can_access(user, {Privilege.MANAGE_ORG_MEMBERS}, organization):
        raise_flask_error(403, "Forbidden")

    # Validate target user exists
    org_user = validate_organization_user(db_session, target_user_id, organization)

    existing_role_ids = {our.role_id for our in org_user.organization_user_roles}
    new_role_ids = set(data["role_ids"])
    if existing_role_ids == new_role_ids:
```

api/src/services/organizations_v1/update_user_organization_roles.py
```diff
@@ -0,0 +1,91 @@
...
+
+    # Fetch and assign new roles
+    roles = get_role(db_session, new_role_ids)
+    org_user_roles = [OrganizationUserRole(organization_user=org_user, role=role) for role in roles]
+
+    for our in org_user_roles:
+        db_session.add(our)
+
+    org_user.organization_user_roles = org_user_roles
+    db_session.add(org_user)
```
chouinar: I think this does handle deleting roles a user no longer has, but I think it also deletes and creates a new record for anything they already have. While that's fine from an auth perspective, from an audit perspective it'll recreate all of the DB records meaning we wouldn't be able to see that User A got role X assigned at some specific time.


---
I think we need to do this a little more like this (feel free to adjust, probably can make this a bit simpler)

```py
# Delete roles
role_ids_to_delete = existing_role_ids - new_role_ids
roles_to_delete = [r for r in  user.organization_user_roles if r.role_id in role_ids_to_delete]
for org_role in roles_to_delete:
     db_session.delete(org_user_role)

# new roles
role_ids_to_add = new_role_ids - existing_role_ids
for role_id in role_ids_to_add:
    db_session.add(OrganizationUserRole(organization_user=org_user, role_id=role_id))
```

chouinar (approved):

Issue #6585, opened 2025-10-03
Stream soap proxy response data don't chunk


jakobpederson:
### Summary

Currently we take the soap proxy response data and chunk it and turn it into a string before passing it back as a response. What we want to do going forward is just stream the soap proxy response data as bytes.

### Acceptance criteria

- [ ] The performance should improve
- [ ] There should be no noticeable change in behavior

PR #6592, opened 2025-10-03
[Issue #6585] Stream soap proxy response data instead of chunking

jakobpederson:
## Summary

Fixes for #ISSUE6585 

## Changes proposed

Modify SoapResponse to take in bytes or an iterator for data. This is so we can stream data from the soap proxy instead of chunking it into a string.

## Context for reviewers

Currently modifying SoapResponse to perform identically if you pass it an iterator or bytes. Adding the iterator did add some challenges because that iterator could be exhausted, so I modified SoapResponse to cache that data. I also put in a boolean check for the part of the client that compares the two responses (and logs the difference) since you have to stop streaming in order to do that. In the future if we want to, we can flip the boolean based on the operation_name.

### DIFF
--- a/api/src/legacy_soap_api/legacy_soap_api_client.py
+++ b/api/src/legacy_soap_api/legacy_soap_api_client.py
@@ -85,7 +85,7 @@ class BaseSOAPClient:
 
         try:
             proxy_response_payload = SOAPPayload(
-                proxy_response.data.decode(errors="replace"),
+                proxy_response.to_bytes().decode(errors="replace"),
                 operation_name=self.operation_config.response_operation_name,
                 force_list_attributes=self.operation_config.force_list_attributes,
             )
@@ -150,7 +150,9 @@ class BaseSOAPClient:
                 },
             )
 
-    def get_simpler_soap_response(self, proxy_response: SOAPResponse) -> tuple:
+    def get_simpler_soap_response(
+        self, proxy_response: SOAPResponse, compare: bool = False
+    ) -> tuple:
         """
         This method is responsible getting the simpler soap xml payload.
 
@@ -158,13 +160,6 @@ class BaseSOAPClient:
         proxy responses. We then utilize the SOAPPayload class to get the
         XML soap response from the validated XML dicts.
         """
-        proxy_response_soap_dict = self.get_proxy_soap_response_dict(proxy_response)
-        log_local(
-            msg="proxy response validated dict",
-            data=proxy_response_soap_dict,
-            formatter=json_formatter,
-        )
-
         simpler_response_soap_dict = self.get_soap_response_dict()
         log_local(
             msg="simpler response dict", data=simpler_response_soap_dict, formatter=json_formatter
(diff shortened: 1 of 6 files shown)


api/src/legacy_soap_api/legacy_soap_api_schemas.py
```diff
@@ -1,3 +1,5 @@
+from typing import Iterator, List
```
chouinar: Can use `list`, no need for typing List anymore.

```suggestion
from typing import Iterator
```

api/src/legacy_soap_api/legacy_soap_api_client.py
```diff
@@ -150,21 +150,16 @@ def log_diffs(self, proxy_response_soap_dict: dict, simpler_response_soap_dict:
                 },
             )
 
-    def get_simpler_soap_response(self, proxy_response: SOAPResponse) -> tuple:
+    def get_simpler_soap_response(
+        self, proxy_response: SOAPResponse, compare: bool = False
+    ) -> tuple:
```
chouinar: Where does compare get set/configured? We'd want a configuration per operation saying whether to do a comparison or not.

chouinar (commented):
I think some wires got crossed. The issue we wanted to fix was streaming the request to the SOAP endpoint. While this probably also is needed, the request is where the problem is occurring, not the response (at least not yet, it never gets to resolving the response for large requests).

Could we make sure to create another ticket for handling request-side streaming?

api/src/legacy_soap_api/legacy_soap_api_client.py
jakobpederson: Added a check for request_operation_name in the method instead of compare bool

api/src/legacy_soap_api/legacy_soap_api_client.py
```diff
@@ -180,18 +176,28 @@ def get_simpler_soap_response(self, proxy_response: SOAPResponse) -> tuple:
...
+                self.log_diffs(proxy_response_soap_dict, simpler_response_soap_dict)
 
         use_simpler_response = False
         return (
             get_soap_response(data=simpler_response_xml),
             use_simpler_response,
         )
 
+    def compare_endpoints(self) -> bool:
+        return self.operation_config.request_operation_name in CompareRequestOperationName
```
chouinar: In the interest of having one place we store configuration for an operation, could you instead add this to the `SOAPOperationConfig` class?

api/src/legacy_soap_api/legacy_soap_api_config.py
```diff
@@ -87,6 +90,9 @@ class SOAPOperationConfig:
     # namespaces or namespace prefixes.
     namespace_keymap: dict[str, str | None] = field(default_factory=dict)
 
+    def compare_endpoints(self) -> bool:
+        return self.request_operation_name in CompareRequestOperationName
```
chouinar: I was thinking we'd add this to the configuration below that just the `GetOpportunityListRequest` is configured for at the moment, no enum at all.

jakobpederson: That's a better idea, I'll get on that.

Issue #6119, opened 2025-08-26
Apply endpoints - Modify start application endpoint to require privileges


joshtonava:
### Summary

Add user enforcement logic to the `POST /alpha/applcations/start` endpoint

### Tech Notes

Should require `start_application` privilege IF the user is trying to start an application for an organization (if no organization, no privilege is required)

### Acceptance criteria

- Add user enforcement logic to the `POST /alpha/applcations/start` endpoint
- Add / update unit tests
    - We also want to ensure that we test the case where a user tries to start an application they dont have access to 
- Requires UAT

eric-serrato95:
potentially think around a test where we actually hit the API and the Endpoint to ensure proper dataflow

joshtonava:
Depends on https://github.com/HHS/simpler-grants-gov/issues/6116

PR #6645, opened 2025-10-07
[Issue #6119] Update Start application endpoint to check for privileges

babebe:
## Summary

Work for #6119 

## Changes proposed

- Enforce access control when a user attempts to create an application under an organization.
- The user must have the start_application privilege for that organization.
- Skip privilege check when creating a standalone (non-organization-owned) application.

## Context for reviewers

Updated tests.
Added Test for Forbidden Access when user does not have privilege.


api/src/services/applications/create_application.py
```diff
@@ -200,6 +202,10 @@ def create_application(
         _validate_organization_membership(db_session, organization, user)
```
chouinar: With this change we can remove this check, which is effectively the precursor to the new RBAC approach.

Hmm, if it doesn't break any tests, could we move the `validate_competition_open`, `_validate_organization_expiration` and `_validate_applicant_type` to all happen _after_ the `can_access` check? The first meaningful check we should do (after request validation + 404s) is checking roles to avoid giving info to users that shouldn't have it.

babebe: Made the update. It broke some tests so i updated them.

chouinar: We can remove `_validate_organization_membership` as the `can_access` you added below does that and more.

Issue #6117, opened 2025-08-26
Modify our local seed script to create users in various scenarios


joshtonava:
### Summary

TODO

### Acceptance criteria

TODO

joshtonava:
Dependent on https://github.com/HHS/simpler-grants-gov/issues/6110 https://github.com/HHS/simpler-grants-gov/issues/6111 & https://github.com/HHS/simpler-grants-gov/issues/6112

eric-serrato95:
just extending initial set up Michael put together, in the effort of helping set up testing, adjust it to be role based vs current state

PR #6650, opened 2025-10-07
[ISSUE #6117] Modify our local seed script to create users in various scenarios

joshtonava:
## Summary

Fixes [#6117]

### DIFF
--- a/api/tests/lib/seed_orgs_and_users.py
+++ b/api/tests/lib/seed_orgs_and_users.py
@@ -103,7 +103,7 @@ USER_ONE_ORG_ORG_USER1 = factories.OrganizationUserFactory.build(
     organization_user_id=uuid.UUID("3ab87af3-66d3-4a44-9eb1-7da598ffb05b"),
     organization=ORG1,
     user=USER_ONE_ORG,
-    is_organization_owner=True,
+    as_admin=True,  # This automatically sets up the role and is_organization_owner=True
 )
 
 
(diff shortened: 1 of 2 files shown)


api/tests/lib/seed_orgs_and_users.py
```diff
@@ -130,17 +132,98 @@
...
+
+USER_MIXED_ORG_ROLES_ORG_USER2 = factories.OrganizationUserFactory.build(
+    organization_user_id=uuid.UUID("d9e0f1a2-b3c4-4d5e-6f7a-8b9c0d1e2f3a"),
+    organization=ORG2,
+    user=USER_MIXED_ORG_ROLES,
+    as_member=True,
+)
+
+
+def _assign_organization_role(
```
joshtonava, line 206: So I did update the factories with traits but we dont get to utilize that behavior in the seed scripts because here we are doing the build + merge but the traits really only work on create. 

Also merge is not working as expected here in CI especially for updating OrganizationUserRole its like trying to update part of the composite primary key

joshtonava: Was able to figure this out

api/tests/src/db/models/factories.py
```diff
@@ -1560,6 +1566,26 @@ class Meta:
     user = factory.SubFactory(UserFactory)
     user_id = factory.LazyAttribute(lambda o: o.user.user_id)
 
+    class Params:
+        # New traits for role assignment
+        as_owner = factory.Trait(
```
doug-s-nava, line 1571: I'm probably missing something - don't see these being called directly, are they being used under the hood somewhere, or are they for future use?

doug-s-nava (approved):
passes my use case of allowing one_org_user to access the organization details endpoint for the organization assigned to that user. One minor question around usage

api/tests/src/db/models/factories.py
joshtonava: This pattern cant actually be used for the seed unfortunately :( The factory traits only seem to persist when using .create but the seeds are using .build.


but I still think this pattern could be useful for our tests as we migrate away from the is_organization_owner field.

PR #6651, opened 2025-10-07
[ISSUE #6341] Organization endpoints - Remove user from organization

### LINKED ISSUE #6341: Organization endpoints - Remove user from organization
### Summary

Implement an endpoint to remove a user from an organization completely. This deletes the `organization_user` record and all associated `organization_user_role` records.

https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/1977221432/Tech+Spec+User+Roles+on+the+Backend#DELETE-%2Forganizations%2F%3Aorganization_id%2Fusers%2F%3Auser_id

### Tech Notes
- Required Permission: `manage_org_members` (minimum)
- Open Question from spec: Should `manage_org_admin_members` be required to remove admin users?
- Must cascade delete all `organization_user_role` records
- Should prevent users from removing themselves if they're the last admin
- Consider soft delete for audit trail (though spec indicates hard delete)

### Acceptance criteria

- Implement `DELETE /organizations/:organization_id/users/:user_id`
- Deletes appropriate records, returns success response with appropriate status code
- Prevents removing last admin from organization
- Include unit tests

joshtonava:
## Summary

Fixes #6341 

## Changes proposed

Implements `DELETE /organizations/:organization_id/users/:user_id` endpoint to remove users from organizations.

## Context for reviewers

This endpoint completes the organization member management CRUD operations. 
- Requires `MANAGE_ORG_MEMBERS` privilege for access
- Prevents removal of last user with admin privileges (last admin protection)

## Validation steps

```bash
make test args="-vv tests/src/api/organizations_v1/test_remove_user_from_organization.py"
```

### DIFF
--- a/api/src/services/organizations_v1/update_user_organization_roles.py
+++ b/api/src/services/organizations_v1/update_user_organization_roles.py
@@ -7,12 +7,7 @@ from src.adapters import db
 from src.api.route_utils import raise_flask_error
 from src.auth.endpoint_access_util import can_access
 from src.constants.lookup_constants import Privilege, RoleType
-from src.db.models.user_models import (
-    LinkRoleRoleType,
-    OrganizationUserRole,
-    Role,
-    User,
-)
+from src.db.models.user_models import LinkRoleRoleType, OrganizationUserRole, Role, User
 from src.services.organizations_v1.get_organization import get_organization
 from src.services.organizations_v1.organization_user_utils import validate_organization_user_exists
 


api/src/services/organizations_v1/remove_user_from_organization.py
```diff
@@ -0,0 +1,126 @@
...
+
+    Args:
+        db_session: Database session
+        target_user_id: ID of user being removed
+        organization: Organization the user is being removed from
+
+    Raises:
+        FlaskError: 403 if this would remove the last admin
+    """
+    from src.db.models.user_models import LinkRolePrivilege, Role
```
chouinar: Can we put imports at the top of the file?

api/src/services/organizations_v1/remove_user_from_organization.py
```diff
@@ -0,0 +1,126 @@
...
+    db_session.delete(org_user)
+
+    logger.info(
+        "Successfully removed user from organization",
+        extra={
+            "requesting_user_id": user.user_id,
+            "target_user_id": target_user_id,
+            "organization_id": organization_id,
+        },
+    )
```
chouinar: Nitpicky, but we can exclude the extra here. You added `target_user_id` and `organization_id` at the start of the request. The `requesting_user_id` sort've already exists as we put the auth user ID automatically on requests as well.

Same with the log message earlier.

api/src/services/organizations_v1/remove_user_from_organization.py
```diff
@@ -0,0 +1,126 @@
...
+
+    admin_user_ids = set(db_session.execute(admin_user_ids_stmt).scalars().all())
+
+    # Check if target user is an admin and if they're the only admin
+    if target_user_id in admin_user_ids and len(admin_user_ids) == 1:
+        raise_flask_error(403, message="Cannot remove the last administrator from organization")
+
+
+def validate_organization_user_exists(
```
chouinar: I think the update user service also has this function, could we move it to a central location and both of them use it?

api/src/services/organizations_v1/remove_user_from_organization.py
```diff
@@ -0,0 +1,126 @@
...
+
+    # Check if requesting user has permission to manage org members
+    if not can_access(user, {Privilege.MANAGE_ORG_MEMBERS}, organization):
+        raise_flask_error(403, "Forbidden")
+
+    # Validate target user is a member of the organization
+    org_user = validate_organization_user_exists(db_session, target_user_id, organization)
+
+    # Check last admin protection
+    check_last_admin_protection(db_session, target_user_id, organization)
```
chouinar, line 85: A question to take to design/product. Do we want a user to be able to remove themselves from an organization? If so, should they require any privileges to do so (ie. if I did not have `manage_org_members` but wanted to leave an org, should I be allowed to?).

Not something we need to solve for this particular PR

joshtonava: Interesting question though

cc: @eric-serrato95

api/src/services/organizations_v1/remove_user_from_organization.py
joshtonava: Oops..

api/src/services/organizations_v1/remove_user_from_organization.py
joshtonava: made a util file we could probably move a few more things into here as part of a cleanup

PR #6690, opened 2025-10-08
Bug/opportunity emails query

babebe:
## Summary

Update query such that it only exclude the opportunity if the user associated with it is suppressed.
Updated tests

### DIFF
--- a/api/src/task/notifications/search_notification.py
+++ b/api/src/task/notifications/search_notification.py
@@ -12,12 +12,7 @@ from src.db.models.opportunity_models import (
     Opportunity,
     OpportunitySummary,
 )
-from src.db.models.user_models import (
-    LinkExternalUser,
-    SuppressedEmail,
-    UserSavedOpportunity,
-    UserSavedSearch,
-)
+from src.db.models.user_models import LinkExternalUser, SuppressedEmail, UserSavedSearch
 from src.services.opportunities_v1.search_opportunities import search_opportunities_id
 from src.task.notifications.base_notification import BaseNotificationTask
 from src.task.notifications.config import EmailNotificationConfig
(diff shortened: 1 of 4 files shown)


mdragon:
@babebe I think there's an easy way to get the SELECT this is running to spit out? I'm just curious why the email match is insufficient but also matching on user_id fixes that so I was trying to understand the query that's actually being built by the ORM.

chouinar:
@mdragon / @babebe - If you add `echo=True` to where we create the sqlalchemy engine https://github.com/HHS/simpler-grants-gov/blob/main/api/src/adapters/db/clients/postgres_client.py#L46 - then it'll print out all DB queries (it's incredibly verbose).

An alternative way we could implement this if the exists is too complex is something like:

```sql
select * from ... left join suppressed_email on link_external_user.email = suppressed_email.email WHERE suppressed_email.email is null
```

Basically do the join, but if the join isn't found then exclude it. Probably also some sort of subquery (`where email not in (select email from suppressed_email.email)`).

Since we'll need this pattern in all of our email logic going forward, might be good to have a util to just add this.

babebe:
@mdragon So it is evaluating `SELECT * 
FROM api.suppressed_email, api.link_external_user 
WHERE api.suppressed_email.email = api.link_external_user.email`
and if there is one match this will evaluate as `True` and `Not Exist(...)` will be `False` so it will always return empty rows. 
The ` LinkExternalUser.user_id == UserSavedOpportunity.user_id` is just telling it to reference the outer query where it also matches  the `user_id` of the current row being processed in the outer query.

mdragon:
> @mdragon So it is evaluating ...

Right ok, I see the difference now. It's not the same table in both of the clauses which is what gives the the extra specificity.

Issue #2973, opened 2024-11-21
Build debian from scratch, rather than using re-built debian


coilysiren:
### Summary

motivation: https://betagrantsgov.slack.com/archives/C05TSL64VUH/p1732211068387059?thread_ts=1732207965.167499&cid=C05TSL64VUH

### Acceptance criteria

//

PR #6703, opened 2025-10-09
[Issue #2973]: Update the API docker image to use a custom image 

prasnava:
## Summary

Use a custom api docker image with self managed dependencies.

## Changes proposed

Built release image from scratch using the amazonlinux:2023 image.  The new image compressed size in ECR s 219mb, which is less than half of the old image size. (580 mb) 

## Validation steps

After deploying in dev, the api and step functions should continue to run with out any errors.

### DIFF
--- a/api/src/form_schema/forms/sf424.py
+++ b/api/src/form_schema/forms/sf424.py
@@ -732,6 +732,28 @@ FORM_UI_SCHEMA = [
             {"type": "field", "definition": "/properties/applicant/properties/province"},
             {"type": "field", "definition": "/properties/applicant/properties/country"},
             {"type": "field", "definition": "/properties/applicant/properties/zip_code"},
+        ],
+    },
+    {
+        "type": "section",
+        "name": "organizational_unit",
+        "label": "8e. Organizational Unit",
+        "children": [
+            {
+                "type": "field",
+                "definition": "/properties/department_name",
+            },
+            {
+                "type": "field",
+                "definition": "/properties/division_name",
+            },
+        ],
+    },
+    {
+        "type": "section",
+        "name": "contact_person",
+        "label": "8f. Name and contact information of person to be contacted on matters involving this application",
+        "children": [
             {"type": "field", "definition": "/properties/contact_person/properties/prefix"},
             {"type": "field", "definition": "/properties/contact_person/properties/first_name"},
             {"type": "field", "definition": "/properties/contact_person/properties/middle_name"},


prasnava:
This branch has been successfully [deployed](https://github.com/HHS/simpler-grants-gov/actions/runs/18422843863/job/52500239704) in training.

api/Dockerfile
```diff
@@ -1,135 +1,321 @@
...
-    && apt-get clean \
-    && rm -fr /var/lib/apt/lists/* \
-    && rm /etc/ssl/private/ssl-cert-snakeoil.key
+FROM amazonlinux:2023 AS builder
+
+ARG PYTHON_VERSION=3.13.8
+ARG PY_MM=3.13
+ARG POETRY_VERSION=2.2.1
+ARG PIP_VERSION=25.2
+
```
chouinar, line 7: Is there later work to move most of this first step to a separate Dockerfile + image building process?

It'd speed up local development to just pull in this built image + the analytics repo would need it as well.

api/Dockerfile
```diff
@@ -1,135 +1,321 @@
...
 
 COPY . /api
 
-# Set the host to 0.0.0.0 to make the server available external
-# to the Docker container that it's running in.
 ENV HOST=0.0.0.0
 
-# Run the application.
-CMD ["poetry", "run", "python", "-m", "src"]
+CMD ["poetry", "run", "gunicorn", "--worker-tmp-dir=/dev/shm", "src.app:create_app()"]
```
chouinar: Is this still the command for running the API locally (via our docker-compose file)? If so, we intentionally do not want gunicorn for local development.

api/Dockerfile
```diff
@@ -1,135 +1,321 @@
...
+    TMPDIR="/dev/shm" \
+    SSL_CERT_FILE="/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem" \
+    SSL_CERT_DIR="/etc/ssl/certs:/etc/pki/ca-trust/extracted/pem"
 
+# Run as the non-root user created in release
 USER ${RUN_USER}
 
-# Run the application.
-CMD ["poetry", "run", "gunicorn", "src.app:create_app()"]
+CMD ["gunicorn", "--worker-tmp-dir=/dev/shm", "src.app:create_app()"]
```
chouinar: I'm curious how this works without the `poetry run` which to my understanding would be needed for any python command so it has the necessary dependencies.

api/Dockerfile
```diff
@@ -1,135 +1,321 @@
...
 
-#---------
-# Release
-#---------
 
-FROM base AS release
+#==============================================
+# Stage 5: Minimal release runtime
+#==============================================
+FROM scratch AS release
```
chouinar: Do we not want to reuse what we made step 2 to make a base image? I assume that would let us trim most of the COPY commands below as they look to be repeated.

api/Dockerfile
prasnava: Will create another ticket for moving the base image into a separate Dockerfile + image building process. It will definitely speed up the local dev builds.

api/Dockerfile
prasnava: Updated the dev stage command to:

```
CMD ["poetry", "run", "flask", "--app", "src.app", "run", "--host", "0.0.0.0", "--port", "8080", "--reload"]
```

api/Dockerfile
prasnava: Updated the release stage; it's now using the base image.

api/Dockerfile
prasnava: In the release image we set the PATH variable to include the /api/.venv/bin. In an earlier step where we run poetry install it downloaded the binary into the /api/.venv folder.

Issue #6705, opened 2025-10-09
Remove beta banner


andycochran:
### Summary

Per #6254, we no longer need the beta banner. The site is considered in-production and its messaging should not dissuade users from relying on its features. 

We should also make sure the pages have the appropriate top margin and we're not relying on the beta banner for spacing the page content away from the header/hero (https://betagrantsgov.slack.com/archives/C05TGEL3C6Q/p1759348605537979). 

### Acceptance criteria

- Beta banner is removed
- Every page has the appropriate top-margin

PR #6706, opened 2025-10-09
[Issue #6705] Nix beta banner

andycochran:
## Summary

Addresses #6705. This PR removes the default `BetaAlert` from all the pages. It does not remove the `BetaAlert` from the Search page or the Feature Flag page — where the default content is overridden. 

It also cleans up some spacing and layout issues 

## Changes proposed

- Remove `BetaAlert` from… 
  - `/newsletter/confirmation`
  - `/newsletter/unsubscribe`
  - 404 page
  - Opportunity page
  - Application success page
- Remove the test looking for the alert on the 404 page
- Remove `Breadcrumbs` from the search page (these are unnecessary and waste layout space) 
- Adjust padding of search page alert to account for removing breadcrumbs
- Improve layout and formatting of the global layout's "October 1, 2025" message (making the line lengths more legible by using columns) 

image

image

## Context for reviewers

The `BetaAlert` is now only being used for the Search and FF pages. The beta message content is no longer being usesd. The component is really no longer a "beta" alert. It might make sense to refactor or rename the component. 

## Validation steps

- Look through all the pages and verify that the default beta message no longer appears.

### DIFF
--- a/frontend/src/components/Layout.tsx
+++ b/frontend/src/components/Layout.tsx
@@ -43,8 +43,8 @@ export default function Layout({ children, locale }: Props) {
                 <div className="grid-row grid-gap-4 font-sans-2xs">
                   <div className="tablet-lg:grid-col-5">
                     There has been a lapse in appropriated federal funds as of
-                    October 1, 2025. <a href="https://grants.gov/">Grants.gov</a>{" "}
-                    and{" "}
+                    October 1, 2025.{" "}
+                    <a href="https://grants.gov/">Grants.gov</a> and{" "}
                     <a href="https://simpler.grants.gov/">Simpler.Grants.gov</a>{" "}
                     will still be available, but service may be delayed with
                     reduced Federal support staff presence.
@@ -53,9 +53,9 @@ export default function Layout({ children, locale }: Props) {
                     For those programs affected by the funding lapse, the{" "}
                     <a href="https://grants.gov/">Grants.gov</a> and{" "}
                     <a href="https://simpler.grants.gov/">Simpler.Grants.gov</a>{" "}
-                    systems will accept and store applications until such time as
-                    the responsible awarding agency has the authority and funding
-                    to return to normal business operations.
+                    systems will accept and store applications until such time
+                    as the responsible awarding agency has the authority and
+                    funding to return to normal business operations.
                   </div>
                 </div>
               </div>


andycochran:
cc @chrisw-nava — you might wanna take a look at this, as I've added a couple small design changes to the issue scope

frontend/src/app/[locale]/(base)/search/page.tsx
```diff
@@ -98,10 +96,6 @@ function Search({ searchParams, params }: SearchPageProps) {
           })}
         />
         <div className="grid-container">
-          <Breadcrumbs
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/search/page.tsx:101]: what's the deal with breadcrumbs at this point? There has been chatter at times about removing them, but they seem pretty key to the way that things are being thought about in terms of the new user based page IA. No qualms about removing this one here, but curious if there is a holistic plan for when to show them an when not to

frontend/src/app/[locale]/(base)/search/page.tsx
```diff
@@ -1,5 +1,4 @@
 import { Metadata } from "next";
-import { SEARCH_CRUMBS } from "src/constants/breadcrumbs";
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/search/page.tsx:2]: can delete this constant as well

frontend/src/app/[locale]/(base)/search/page.tsx
Comment (andycochran) [reply frontend/src/app/[locale]/(base)/search/page.tsx:101]: Agree we need better guidance about when to have them. But I don't think that guidance should recommend them on the top-level search page.

Comment (chrisw-nava) [reply frontend/src/app/[locale]/(base)/search/page.tsx:101]: @andycochran @doug-s-nava - as long as this doesn't affect the "Return to Grants.gov" link when a user navigates with this UTM param then I have no opinion (just wanted to check) 

https://simpler.grants.gov/search?utm_source=Grants.gov

chrisw-nava:
> cc @chrisw-nava — you might wanna take a look at this, as I've added a couple small design changes to the issue scope

@andycochran Looks good to me. I don't have a strong opinion about the columns for the shutdown alert, so if these are better for legibility then great!

frontend/src/app/[locale]/(base)/search/page.tsx
Comment (andycochran) [reply frontend/src/app/[locale]/(base)/search/page.tsx:101]: D'oh! Good callout. I think I just killed the return-to-gg link. I'll revert this.

Comment (andycochran) [reply frontend/src/app/[locale]/(base)/search/page.tsx:101]: Or, maybe I'll try to put the return link back (in a better spot in the layout) w/o the breadcrumbs

Comment (andycochran) [reply frontend/src/app/[locale]/(base)/search/page.tsx:101]: I opened #6716 to correct my mistake. Will patch it before the next deploy. Thanks for helping catch this, @chrisw-nava!

PR #6712, opened 2025-10-10
[Issue #6396] Support XML Attachment transformations

### LINKED ISSUE #6396: Add logic to transform an attachment

mikehgrantsgov:
## Summary

Fixes #6396 

## Changes proposed

Adds an `AttachmentTransformer` with logic to transform attachments to XML.
Ingrates attachments into `_generate_xml_with_namespaces`. 
Add config for known attachment fields of the sf424:
- `debt_explanation`
- `areas_affected`
- `additional_congressional_districts`
- `additional_project_title`
Add tests

## Context for reviewers

Attachments were left separate from the initial XML generation work due to their complexity and research needed around them e.g. HashAlgorithm. This PR adds attachment support to the XML generation work.

## Validation steps

See updated unit tests and `test_cases` for independent XML generation/validation.

### DIFF
--- a/api/src/services/xml_generation/transformers/attachment_transformer.py
+++ b/api/src/services/xml_generation/transformers/attachment_transformer.py
@@ -159,21 +159,26 @@ class AttachmentTransformer:
         """
         group_elem = lxml_etree.SubElement(parent, element_name)
 
-        # Handle direct list of attachments
+        # Normalize to list of file data
+        files_to_add: list[Any] = []
         if isinstance(attachment_data, list):
-            for file_data in attachment_data:
-                file_elem = lxml_etree.SubElement(group_elem, "AttachedFile")
-                self._populate_attachment_content(file_elem, file_data, nsmap)
+            files_to_add = attachment_data
         elif isinstance(attachment_data, dict) and "AttachedFile" in attachment_data:
             attached_files = attachment_data["AttachedFile"]
-            if isinstance(attached_files, list):
-                for file_data in attached_files:
-                    file_elem = lxml_etree.SubElement(group_elem, "AttachedFile")
-                    self._populate_attachment_content(file_elem, file_data, nsmap)
-            else:
-                # Single file in the list
-                file_elem = lxml_etree.SubElement(group_elem, "AttachedFile")
-                self._populate_attachment_content(file_elem, attached_files, nsmap)
+            files_to_add = attached_files if isinstance(attached_files, list) else [attached_files]
+
+        # Add each file
+        for file_data in files_to_add:
+            self._add_attached_file_element(group_elem, file_data, nsmap)
+
+    def _add_attached_file_element(
+        self,
+        parent: lxml_etree._Element,
+        file_data: Any,
+        nsmap: dict[str, str],
+    ) -> None:
+        file_elem = lxml_etree.SubElement(parent, "AttachedFile")
+        self._populate_attachment_content(file_elem, file_data, nsmap)
 
     def _populate_attachment_content(
         self,


api/src/services/xml_generation/validation/test_cases.py
```diff
@@ -205,6 +205,239 @@
...
+            "delinquent_federal_debt": True,
+            "debt_explanation": {
+                "FileName": "debt_explanation.pdf",
+                "MimeType": "application/pdf",
+                "FileLocation": {"@href": "./attachments/debt_explanation.pdf"},
+                "HashValue": {
+                    "@hashAlgorithm": "SHA-1",
+                    "#text": "aGVsbG8gd29ybGQxMjM0NTY3ODk=",
+                },
+            },
```
chouinar: In the JSON we have, our attachment fields would be UUIDs, which are just the application attachment IDs. We need this transformation to start with those UUIDs and handle processing an application attachment record into this and then the XML.

It's fine if this is an intermediate step, but the valid/base JSON should be UUIDs, and this test should have UUIDs in it.

How we want to map those UUIDs we can do a few ways, it's fine if you want this code to not need to know what an application attachment is. We could just have a function that given an application, converts all app attachments into a map of UUID to this info, and then the transformation code takes that in (that would probably help with having the tests work as you'd otherwise need an actual application setup for the UUIDs).

api/src/services/xml_generation/transformers/attachment_transformer.py
```diff
@@ -0,0 +1,316 @@
...
+
+        # Process single attachment fields
+        single_fields = [
+            ("areas_affected", "AreasAffected"),
+            ("additional_congressional_districts", "AdditionalCongressionalDistricts"),
+            ("debt_explanation", "DebtExplanation"),
+        ]
+
+        for input_key, xml_key in single_fields:
+            if input_key in input_data and input_data[input_key] is not None:
```
jakobpederson: We could possibly truncate this to 
`if input_data.get(input_key, None) is not None`

api/src/services/xml_generation/transformers/attachment_transformer.py
```diff
@@ -0,0 +1,316 @@
...
+        Returns:
+            Processed data ready for XML generation
+        """
+        result = {}
+
+        # Process single attachment fields
+        single_fields = [
+            ("areas_affected", "AreasAffected"),
+            ("additional_congressional_districts", "AdditionalCongressionalDistricts"),
+            ("debt_explanation", "DebtExplanation"),
```
chouinar: These shouldn't be in here, they should be in the configuration.

api/src/services/xml_generation/transformers/attachment_transformer.py
```diff
@@ -0,0 +1,316 @@
...
+            attachment_data: Attachment group data dictionary
+            nsmap: Namespace map
+        """
+        group_elem = lxml_etree.SubElement(parent, element_name)
+
+        # Handle direct list of attachments
+        if isinstance(attachment_data, list):
+            for file_data in attachment_data:
+                file_elem = lxml_etree.SubElement(group_elem, "AttachedFile")
+                self._populate_attachment_content(file_elem, file_data, nsmap)
```
jakobpederson: (optional) It seems like you're functionally calling _add_single_attachment_element three times in this method if you wanted to swap these out.

api/src/services/xml_generation/models/attachment.py
```diff
@@ -0,0 +1,193 @@
...
+                      of guessing from file extension. This should be preferred when
+                      available from the application attachment table in the database.
+
+        Returns:
+            AttachmentFile instance with computed hash and MIME type
+        """
+        file_path = Path(file_path)
+
+        if not file_path.exists():
+            raise FileNotFoundError(f"File not found: {file_path}")
```
chouinar: All files we'll be working with are on s3, you'd have to use our file utils, but I think it's fine to skip this sort of check regardless and if there is a filenotfound issue just let that error happen

mikehgrantsgov:
@chouinar I refactored some stuff taking your suggestions into account. Added a utility `attachment_mapping` which creates a `dict` of attachment mappings from a list of `ApplicationAttachment` objects and lots of associated downstream changes to simplify.

api/src/services/xml_generation/utils/attachment_mapping.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def create_attachment_mapping(
+    application: Application, filename_overrides: dict[UUID, str] | None = None
+) -> dict[UUID, AttachmentInfo]:
+    filename_overrides = filename_overrides or {}
+    mapping: dict[UUID, AttachmentInfo] = {}
+
+    # Get all attachments for this application
+    attachments = getattr(application, "attachments", [])
```
chouinar: Attachments are just a relationship on the application
```suggestion
    attachments = application.application_attachments
```

api/src/services/xml_generation/transformers/attachment_transformer.py
```diff
@@ -0,0 +1,389 @@
...
+logger = logging.getLogger(__name__)
+
+
+class AttachmentTransformer:
+    """Transformer for handling attachment data in XML generation."""
+
+    def __init__(
+        self,
+        attachment_namespace: str = "http://apply.grants.gov/system/Attachments-V1.0",
+        attachment_mapping: dict[UUID, Any] | None = None,
```
chouinar: Should this be `dict[UUID, AttachmentInfo]` - is there a case where it wouldn't be?

api/src/services/xml_generation/transformers/attachment_transformer.py
```diff
@@ -0,0 +1,389 @@
...
+            ValueError: If UUID format is invalid or UUID not found in mapping
+        """
+        # Convert string UUID to UUID object
+        if isinstance(uuid_value, str):
+            try:
+                uuid_value = UUID(uuid_value)
+            except (ValueError, AttributeError) as e:
+                raise ValueError(
+                    f"Invalid UUID format for field '{field_name}': {uuid_value}"
+                ) from e
```
chouinar: Idea to simplify this, instead of the mapping being keyed on UUID, make the uuids into strings when we create the mapping.

Then we're not having to worry about all the UUID typing/conversion, especially since the JSON data is already strings.

api/src/services/xml_generation/transformers/attachment_transformer.py
```diff
@@ -0,0 +1,389 @@
...
+
+        attachment_info = self.attachment_mapping[uuid_value]
+
+        # Convert AttachmentInfo to dict
+        if hasattr(attachment_info, "to_dict"):
+            return attachment_info.to_dict()
+
+        # If it's already a dict, return it
+        if isinstance(attachment_info, dict):
+            return attachment_info
```
chouinar: Is there a way to have it always be one or the other?

api/src/services/xml_generation/transformers/attachment_transformer.py
```diff
@@ -0,0 +1,389 @@
...
+            try:
+                attachment_file = AttachmentFile.from_file_path(
+                    attachment_info["file_path"], attachment_info.get("file_location")
+                )
+                return self._attachment_file_to_dict(attachment_file)
+            except (FileNotFoundError, Exception):
+                # If file doesn't exist, create a placeholder structure
+                return self._create_placeholder_attachment(attachment_info)
+
+        # Process direct attachment data
```
chouinar: If we assume that attachments will only ever be in one format (the attachment info class), can we trim out these other paths?

There's just a lot of different paths through this code, and I'm not sure we'd ever hit them with how we plan to use this.

---

If there were any sort of issue, we wouldn't want to have a placeholder attachment, if data is ever malformed just let it error.

api/src/services/xml_generation/transformers/attachment_transformer.py
mikehgrantsgov: Makes sense, removed these.

api/src/services/xml_generation/models/attachment.py
mikehgrantsgov: Updated this to let the error happen if we encounter it

api/src/services/xml_generation/transformers/attachment_transformer.py
mikehgrantsgov: I updated these types.

api/src/services/xml_generation/transformers/attachment_transformer.py
mikehgrantsgov: That makes sense and cleaner since JSON will be strings. Updated some typing and conversion pieces to simplify this.

api/src/services/xml_generation/transformers/attachment_transformer.py
mikehgrantsgov: With the typing change we can collapse this and assume to_dict will be available

api/src/services/xml_generation/transformers/attachment_transformer.py
mikehgrantsgov: Updated this to error if data is malformed, and made this generally simpler

Issue #6120, opened 2025-08-26
Apply endpoints - Modify get application endpoints to require privileges


joshtonava:
### Summary

Add user enforcement logic to the `GET` application endpoints

### Tech Notes

Requires `view_application` privilege in either the organization that owns it (if there is one) or the application. ~~(except for `GET /alpha/applications/:application_id` - see comment below)~~

### Acceptance criteria

- Add user enforcement logic to the following endpoints:
    - `GET /alpha/applications/:application_id`
    - `GET /alpha/applications/:application_id/application_form/:app_form_id`
    - `GET /alpha/applications/:application_id/attachments/:application_attachment_id`
   

- Add / update unit tests
- Requires UAT

joshtonava:
Depends on https://github.com/HHS/simpler-grants-gov/issues/6116

doug-s-nava:
~~Problem: the frontend needs to confirm application access authorization (particularly in relation to write and submit actions) via either application or organization based privileges. To check organization based privileges, the frontend would need an organization id related to the application, and the frontend will only have the application id to work with.~~

~~It'd be nice to make this determination without making any extra API calls, but since the organization id will not be available without making a call for application details, we can assume that a call will need to be made.~~

~~The issue to work around in that case is that a call to get application details could result in a 403, even if the user has a write or submit privilege.~~

~~Solution: update the get application endpoint (/applications/<uuid:application_id>) to authorize a user that has any of the following permissions on the application or the application's organization:~~

~~* START_APPLICATION,~~
~~* VIEW_APPLICATION,~~
~~* MODIFY_APPLICATION,~~
~~* SUBMIT_APPLICATION,~~

~~related to github.com/HHS/simpler-grants-gov/issues/6587 (a first attempt at solving this problem)~~

PR #6714, opened 2025-10-10
[Issue #6120] get application endpoints access check

babebe:
## Summary

Work for #6120  

## Changes proposed

Updated endpoints to check for user access. User needs to have `view_application` privilege

- GET /alpha/applications/:application_id
- GET /alpha/applications/:application_id/application_form/:app_form_id
- GET /alpha/applications/:application_id/attachments/:application_attachment_id

Updates tests
Added access denied tests


api/src/services/applications/get_application.py
```diff
@@ -85,6 +87,7 @@ def get_application(
 
     # Check if the user has access to the application (skip for internal users or when user is None)
     if not is_internal_user and user is not None:
+
         check_user_application_access(application, user)
```
chouinar, line 94: I think we should put the `can_access` check here (or in `check_user_application_access`). That way anything that tries to fetch an application needs `view_application`.

While this will mean that this affects basically every application endpoint, I think that's fine - someone with `modify_application` shouldn't really be able to do anything without `view_application`.

It also already handles the weird internal user token from the application form route.

api/tests/src/api/applications/test_application_attachment_routes.py
```diff
@@ -148,7 +151,10 @@ def test_application_attachment_get_200(
     file_contents = "this is text in my file"
 
     application = ApplicationFactory.create()
-    ApplicationUserFactory.create(application=application, user=user)
+    ApplicationUserRoleFactory.create(
+        application_user=ApplicationUserFactory.create(user=user, application=application),
+        role=RoleFactory.create(privileges=[Privilege.VIEW_APPLICATION]),
+    )
```
chouinar: Similar to the utility created for the organization endpoint tests, do we want one to help setup application users in an app?

api/tests/src/api/applications/test_application_routes.py
chouinar, line 1: Could we add tests for a user being able to access an application via their organization?

No application user role, but the org owns the app, so they can view it.

api/src/services/applications/get_application.py
```diff
@@ -119,3 +125,26 @@ def get_application_with_warnings(
...
+    if user and not can_access(user, {Privilege.VIEW_APPLICATION}, application):
+        logger.info(
+            "User attempted to access an application they are not associated with",
+            extra={
+                "user_id": user.user_id,
+                "application_id": application.application_id,
+                "submission_issue": SubmissionIssue.UNAUTHORIZED_APPLICATION_ACCESS,
+            },
+        )
+        raise_flask_error(403, "Forbidden")
```
chouinar, line 150: Just to make sure the weird internal auth scenario isn't going to cause issues / won't be used incorrectly - could we add this before this check?

```py
if not user and not is_internal_user:
    raise Exception("No user found, but not marked as internal auth")
```
Fine to raise an exception instead of `raise_flask_error` as this would mean something in our code is misconfigured and it should cause a 500 / alert us to fix it.

api/src/services/applications/get_application_attachment.py
```diff
@@ -42,3 +36,33 @@ def get_application_attachment(
...
+
+def get_application_attachment(
+    db_session: db.Session,
+    application_id: uuid.UUID,
+    application_attachment_id: uuid.UUID,
+    user: User,
+) -> ApplicationAttachment:
+
+    # Fetch the application which also validates if the user can access it
+    application = get_application(db_session, application_id, user)
```
chouinar: Isn't `get_application` the one without auth (except for the old one we'll remove soon)?

api/src/services/applications/get_application.py
babebe: makes sense

api/src/services/applications/get_application_attachment.py
babebe: Yes, updated doc string

api/src/services/applications/get_application.py
chouinar: Don't see a change for this in the latest iteration

PR #6718, opened 2025-10-10
Dockle ignore Python "settings.py" files wrongly flagged as configuration

pcraig3:
## Summary

Very small update, just updating filepaths for `.dockleignore`, and then we can deploy the NOFO Builder again.

## Changes proposed

Old: `app/.venv/lib/python3.13/site-packages/cssutils/settings.py`
New: `app/.venv/lib/python3.12/site-packages/cssutils/settings.py`

(3.13 changes to 3.12)

## Context for reviewers

I am updating the filepaths for various `**/settings.py` files needed by our various python libraries that Dockle wrongly fails us on with a message like this:

```
FATAL	- CIS-DI-0010: Do not store credential in environment variables/files
	* Suspicious filename found : app/.venv/lib/python3.12/site-packages/cssutils/settings.py (You can suppress it with "-af settings.py")
...
```

There is a Python vuln identified by anchore in our current `3.13.x` version of Python, which is resolved in `3.12.12` and will be fixed in `3.14.1`, but the `14.1` variant is not released yet.  There is no `.13.x` version identified that has the fix, so we have to move minor versions.

What this means is that when we change minor versions, the dockle ignore string is out of sync, and we need to update it. 

So this is relatively small update, just updating filepaths for `.dockleignore`, and then we can deploy our app again.

## Validation steps

New deployments using this branch are working: https://github.com/HHS/simpler-grants-gov/actions/runs/18412199715


mdragon (approved):
Approved

PR #6721, opened 2025-10-10
[Issue #6665] Define schemas for GetApplicationZip

jakobpederson:
## Summary

Fixes for #ISSUE6665

## Changes proposed

Added schemas for request and response for GetApplicationZip.

## Context for reviewers

This is to prove we can consume request xml and convert it to a dict and take a response dict and convert it to xml, with some additional MTOM header.

### DIFF
--- a/api/src/legacy_soap_api/grantors/schemas/__init__.py
+++ b/api/src/legacy_soap_api/grantors/schemas/__init__.py
@@ -1,0 +1,17 @@
+from src.legacy_soap_api.grantors.schemas.get_application_zip_schemas import (
+    FileDataHandler,
+    GetApplicationZipRequest,
+    GetApplicationZipResponse,
+    GetApplicationZipResponseSOAPBody,
+    GetApplicationZipResponseSOAPEnvelope,
+    XOPIncludeData,
+)
+
+__all__ = [
+    "GetApplicationZipResponseSOAPEnvelope",
+    "GetApplicationZipResponseSOAPBody",
+    "GetApplicationZipResponse",
+    "FileDataHandler",
+    "XOPIncludeData",
+    "GetApplicationZipRequest",
+]
--- a/api/src/legacy_soap_api/grantors/services/get_application_zip_response.py
+++ b/api/src/legacy_soap_api/grantors/services/get_application_zip_response.py
@@ -1,0 +1,21 @@
+import uuid
+
+import src.adapters.db as db
+from src.legacy_soap_api.grantors import schemas as grantor_schemas
+
+
+def get_application_zip_response(
+    db_session: db.Session, get_application_zip_request: grantor_schemas.GetApplicationZipRequest
+) -> grantor_schemas.GetApplicationZipResponseSOAPEnvelope:
+    xop_data_instance = grantor_schemas.XOPIncludeData(
+        **{"@href": f"cid:{uuid.uuid4()}-0001@apply.grants.gov"}
+    )
+    file_handler_instance = grantor_schemas.FileDataHandler(**{"xop:Include": xop_data_instance})
+    get_response_instance = grantor_schemas.GetApplicationZipResponse(
+        **{"ns2:FileDataHandler": file_handler_instance}
+    )
+    return grantor_schemas.GetApplicationZipResponseSOAPEnvelope(
+        Body=grantor_schemas.GetApplicationZipResponseSOAPBody(
+            **{"ns2:GetApplicationZipResponse": get_response_instance}
+        )
+    )
--- a/api/src/legacy_soap_api/legacy_soap_api_client.py
+++ b/api/src/legacy_soap_api/legacy_soap_api_client.py
@@ -1,4 +1,5 @@
 import logging
+import uuid
 from typing import Any
 
 from pydantic import ValidationError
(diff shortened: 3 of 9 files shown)


api/src/legacy_soap_api/grantors/services/get_application_zip_response.py
```diff
@@ -0,0 +1,22 @@
...
+from src.legacy_soap_api.grantors import schemas as grantor_schemas
+
+
+def get_application_zip_response(
+    db_session: db.Session, get_application_zip_request: grantor_schemas.GetApplicationZipRequest
+) -> grantor_schemas.GetApplicationZipResponseSOAPEnvelope:
+    cid_uuid = f"{uuid.uuid4()}"
+    xop_data_instance = grantor_schemas.XOPIncludeData(
+        **{"@href": f"cid:{cid_uuid}-3975@apply.grants.gov"}
+    )
```
chouinar: `f"{uuid.uuid4()}"` is a just `str(uuid.uuid4())` and if we only need it once, might as well put that in the next line. This should still work with the patching you did in tests.

```suggestion
    xop_data_instance = grantor_schemas.XOPIncludeData(
        **{"@href": f"cid:{uuid.uuid4()}-3975@apply.grants.gov"}
    )
```

Also, where does that 3975 come from? Is that supposed to be a static number here?

api/src/legacy_soap_api/legacy_soap_api_client.py
```diff
@@ -221,4 +225,40 @@ class SimplerGrantorsS2SClient(BaseSOAPClient):
...
+        if not self.operation_config.is_mtom:
+            return super().get_simpler_soap_response(proxy_response)
+        simpler_response_soap_dict = self.get_soap_response_dict()
+        log_local(
+            msg="simpler response dict", data=simpler_response_soap_dict, formatter=json_formatter
+        )
+        simpler_response_xml = build_mtom_xml_from_dict(
+            self.operation_config.namespaces, simpler_response_soap_dict
+        )
+        boundary_uuid = f"{uuid.uuid4()}"
```
chouinar:
```suggestion
        boundary_uuid = str(uuid.uuid4())
```

api/src/legacy_soap_api/legacy_soap_api_client.py
```diff
@@ -221,4 +225,40 @@ class SimplerGrantorsS2SClient(BaseSOAPClient):
...
+        }
+        boundary = b"--uuid:" + boundary_uuid.encode("utf-8") + b"\n"
+        mime_message = boundary
+        mime_message += (
+            b'Content-Type: application/xop+xml; charset=UTF-8; type="text/xml"\n'
+            b"Content-Transfer-Encoding: binary\n"
+            b"Content-Id: <root.message@cxf.apache.org>\n\n"
+        )
+        mime_message += simpler_response_xml + b"\n"
+        mime_message += boundary
```
chouinar: Can you do something like this? Basically just a concatenation of all the parts?
```suggestion
        mime_message = (
            boundary
            b'Content-Type: application/xop+xml; charset=UTF-8; type="text/xml"\n'
            b"Content-Transfer-Encoding: binary\n"
            b"Content-Id: <root.message@cxf.apache.org>\n\n"
            simpler_response_xml + b"\n"
            boundary
        )
```

api/src/legacy_soap_api/soap_payload_handler.py
```diff
@@ -320,3 +320,43 @@ def _build_xml_elements(
         else:
             el = SubElement(parent, tag)
             el.text = str(value)
+
+
+def _build_mtom_nested_elements(
+    parent_element: etree._Element, data_dict: dict, namespaces: dict, bound_prefixes: set
+) -> None:
```
chouinar, line 331: Could you document/give an example of the mtom format that we're building here?

Just so future devs can better understand what we've taken a while to figure out.

api/src/legacy_soap_api/soap_payload_handler.py
```diff
@@ -320,3 +320,43 @@ def _build_xml_elements(
...
+            continue
+        prefix, local_name = local_name.split(":")
+        ns_uri = namespaces.get(prefix) if prefix else None
+        new_element_nsmap = {}
+        if ns_uri:
+            tag = QName(ns_uri, local_name)
+            if local_name == "GetApplicationZipResponse":
+                nsmap_to_apply = {p: namespaces[p] for p in namespaces if p not in ("soap", "xop")}
+                child_element = SubElement(parent_element, tag, nsmap=nsmap_to_apply)
+                bound_prefixes = set(namespaces.keys()).union(bound_prefixes)
```
chouinar: Any way to make this chunk configurable? We'll have at least a few more endpoints that return the mtom approach (`GetApplication` I think?)

api/src/legacy_soap_api/grantors/services/get_application_zip_response.py
jakobpederson: The 3975 is randomized from what I've seen based on the responses I've seen, I believe to help uniquify the content id (cid). Would it be best for us to generate a random number? I was assuming it wouldn't matter.

chouinar: If it's just random and part of the key, I think it's fine to just be the UUID. If the old system has that format and we really want to keep it, I'd just make it a slightly more static number like `0001`

api/src/legacy_soap_api/grantors/schemas/__init__.py
chouinar, line 1: Threading arbitrarily - unless you're hitting a git issue, could you not force-push to your PRs? I often use the "Changes since your last review" feature in PRs to check what changes have been made, and force pushing makes that not work.

No need to do anything special with your PR commits, if you have 500 of them, we always squash them to 1 when merging to main.

jakobpederson: You got it! I picked up the habit from another place, but I'll stop doing them.

Issue #6689, opened 2025-10-08
Make the Use Simpler Flag configurable by an environment variable


chris-kuryak:
### Summary

Make the Use Simpler Flag configurable by an environment variable.

This ensures that we will not be using these Simpler changes in production until we are ready.

### Acceptance criteria

- [ ] Update the code to use the environment variable instead of the hard-coded "false" that we currently have
- [ ] Update the terraform

chris-kuryak:
Changed estimate from 1pt to 2pt based on Jakob P via Slack

PR #6729, opened 2025-10-14
[Issue #6689] Add USE_SIMPLER flag to env variables

jakobpederson:
## Summary

Fixes for #ISSUE6689 

## Changes proposed

Added USE_SIMPLER variable to the environment and set them to True in lower environments and False in prod. This should allow us to eventually start testing.

## Context for reviewers

Prior to this the variable `use_simpler_response` was hardcoded to False so we were never getting a simpler response.

### DIFF
--- a/api/src/legacy_soap_api/simpler_soap_api.py
+++ b/api/src/legacy_soap_api/simpler_soap_api.py
@@ -62,17 +62,19 @@ def get_simpler_soap_response(
             },
         )
         return soap_proxy_response
-    if use_simpler:
+
+    if use_simpler or simpler_soap_client.operation_config.always_call_simpler:
         simpler_soap_response = simpler_soap_client.get_simpler_soap_response(soap_proxy_response)
-        if simpler_soap_response is not None:
-            logger.info(
-                "simpler_soap_api: Successfully processed request and returning Simpler SOAP response",
-                extra={
-                    "soap_api_event": LegacySoapApiEvent.RETURNING_SIMPLER_RESPONSE,
-                    "used_simpler_response": use_simpler,
-                },
-            )
-            return simpler_soap_response
+
+    if use_simpler and simpler_soap_response is not None:
+        logger.info(
+            "simpler_soap_api: Successfully processed request and returning Simpler SOAP response",
+            extra={
+                "soap_api_event": LegacySoapApiEvent.RETURNING_SIMPLER_RESPONSE,
+                "used_simpler_response": use_simpler,
+            },
+        )
+        return simpler_soap_response
 
     logger.info(
         "simpler_soap_api: Successfully processed request and returning SOAP proxy response",
--- a/api/src/legacy_soap_api/legacy_soap_api_config.py
+++ b/api/src/legacy_soap_api/legacy_soap_api_config.py
@@ -70,6 +70,7 @@ class SOAPOperationConfig:
     response_operation_name: str
     compare_endpoints: bool = False
     is_mtom: bool = False
+    always_call_simpler: bool = False
 
     # Some SOAP XML payloads will not force a list of objects when converting to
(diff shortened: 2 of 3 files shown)


api/src/legacy_soap_api/simpler_soap_api.py
```diff
@@ -28,7 +29,7 @@ def get_simpler_soap_response(
         else SimplerGrantorsS2SClient
     )
 
-    use_simpler = False
+    use_simpler = os.environ.get("USE_SIMPLER") == "true"
```
chouinar: Don't use environment variables like this, we instead use pydantic to load them. It handles defaults/better as well as [conversion from strings ](https://docs.pydantic.dev/2.3/usage/types/booleans/) 

See any of our classes that derive from `PydanticBaseEnvConfig`.

Having a default would also mean we only need to configure environments that differ (eg. set it to `False`, flip that only locally for now)

api/tests/src/legacy_soap_api/test_legacy_soap_api_routes.py
```diff
@@ -60,40 +60,41 @@ def test_invalid_xml_server_error_500(client) -> None:
...
-    assert response.headers["Content-Length"] == "527"
-    assert expected == response.data
-    assert (
-        response.headers["Content-Type"]
-        == f'multipart/related; type="application/xop+xml"; boundary="uuid:{test_uuid}"; start="<root.message@cxf.apache.org>"; start-info="text/xml"'
-    )
+# TODO: add back once the GetApplicationZip endpoint can return not found
+# @mock.patch("uuid.uuid4")
+# @mock.patch("src.legacy_soap_api.legacy_soap_api_proxy._get_soap_response")
+# def test_getapplicationzip_operation_returns_not_found_response_if_simpler_id_is_used(
```
chouinar: What about this change made this need to be removed?

In our tests we can override environment variables, we usually use `monkeypatch` for that like so:

```py
def my_test(monkeypatch):
    monkeypatch.setenv("MY_ENV_VAR", "some-value")

    # run code
```
It'll reset the env var after the test


---

If it does need to be temporarily removed, instead of commenting it out, you can mark something to be skipped by pytest.

jakobpederson: I removed the test because currently Simpler GetApplicationZip isn't wired up to handle a not found scenario. But we will definitely need that eventually.

api/src/legacy_soap_api/simpler_soap_api.py
```diff
@@ -61,20 +62,17 @@ def get_simpler_soap_response(
...
-        logger.info(
-            "simpler_soap_api: Successfully processed request and returning Simpler SOAP response",
-            extra={
-                "soap_api_event": LegacySoapApiEvent.RETURNING_SIMPLER_RESPONSE,
-                "used_simpler_response": use_simpler,
-            },
-        )
-        return simpler_soap_response
+    if use_simpler:
+        simpler_soap_response = simpler_soap_client.get_simpler_soap_response(soap_proxy_response)
```
chouinar: This would adjust the logic for GetApplicationList - it would only do the diff if we enable this flag. 

But for most of the upcoming endpoints we're building, it wouldn't make sense to even bother with the call.

Maybe another thing we configure per endpoint?

Something like:

```py
simpler_soap_response = None
if endpoint_config.always_call_simpler or use_simpler:
    simpler_soap_response = ...

if simpler_soap_response is not None and use_simpler:
    return simpler_soap_response

return soap_proxy_response
```

jakobpederson: I was probably trying to be too clever. I could just roll it back to the way it was and it should do what we want without having to add the configure. But I could do that endpoint config if that's what we'll probably be leveraging going forward.

Issue #6122, opened 2025-08-26
Apply endpoints - Modify submit application endpoint to require privileges


joshtonava:
### Summary

Add user enforcement logic to the application submit endpoints

### Tech Notes

Requires submit_application privilege in either the organization that owns it (if there is one) or the application.

### Acceptance criteria

- Add user enforcement logic to the following endpoints:
    - `POST /alpha/applications/:application_id/submit`
- Add / update unit tests

joshtonava:
Depends on https://github.com/HHS/simpler-grants-gov/issues/6116

PR #6732, opened 2025-10-14
[Issue #6122] update submit application endpoint with user access check 

babebe:
## Summary

Works  for #6122  

## Changes proposed
Updated endpoints to check for user access. User needs to have `submit_application` privilege

- POST /alpha/applications/:application_id/submit

Update tests
Added access denied tests

### DIFF
--- a/api/src/legacy_soap_api/grantors/schemas/get_application_zip_schemas.py
+++ b/api/src/legacy_soap_api/grantors/schemas/get_application_zip_schemas.py
@@ -1,6 +1,6 @@
-from typing import Self
+from typing import Any, Optional, Self
 
-from pydantic import BaseModel, Field, model_validator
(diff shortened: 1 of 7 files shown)


chouinar (dismissed):

api/tests/src/api/applications/test_application_routes.py
```diff
@@ -2004,6 +1990,9 @@ def test_application_form_update_forbidden_not_in_progress(
         privileges=[Privilege.SUBMIT_APPLICATION],
         status=initial_status,
     )
+    _, application, token = create_user_in_app(
+        db_session, privileges=[Privilege.MODIFY_APPLICATION], status=initial_status
+    )
```
chouinar, line 2015: I think this line got duplicated from somewhere else? It looks like it overwrites the lines immediately above it

chouinar (approved):

Issue #6588, opened 2025-10-03
Show in-line error message in required forms table after attempting submission


chris-kuryak:
### Summary

**Current functionality:**
1. Required form Z has no pre-populated fields
2. Required form Z has some required fields
3. Required form Z has not been opened or saved
4. User attempts to submit application and triggers an error at the top of the application page indicating that form Z needs to be completed.
5. Required form Z DOES NOT show any indication in-line of the Required Forms table that some required fields are missing

<img width="1162" height="392" alt="Image" src="https://github.com/user-attachments/assets/3edbf057-a522-4858-b036-71215f869287" />
<img width="1200" height="408" alt="Image" src="https://github.com/user-attachments/assets/f843c4d4-4e43-41f8-83cc-0937dedf6f9f" />

**Expected functionality:**
5. Required form Z DOES show an indication in-line of the Required Forms table indicating "Some issues found. Check your entries."

#### Solution discussed on 10/15 between Michael C and Doug S and Chris K
- Changing the backend to no longer return "Not Started" status should solve the issue. Unit tests will need to be updated. UAT still needs to be passed before deployment.

### Acceptance criteria
- [ ] Update backend to no longer return "Not Started" status upon validation. Only return "In Progress" or "Complete"
- [ ] Update unit tests
- [ ] UAT passed

doug-s-nava:
The current logic is to show the "some issues found" message if the form is "in progress". I believe there is a quirk where prepopulating fields on a form updates the status to "in progress", which is why it displays on some forms and not others. 

Two possible solutions:
* run logic to "save" all forms when the application is created. This will pick up any errors around required fields and mark the form "in progress" so you get the same behavior regardless of prepopulation. Seems kinda weird to set everything "in progress" though
* handle a "not started" status. At this point there is no indication if the form has not been started. We could add a message for required forms saying "this form has not been started" which should be enough indication. If we need to check for required fields that's a bunch more work, and I'd like that to happen on the backend if we need to do it

chris-kuryak:
@doug-s-nava 
I assumed that the first solution was what was currently happening. Or at least, I think that is what was discussed when we put the application page together originally. But I acknowledge that we were moving quickly without a lot of documentation. I think it was discussed as the easiest way to get the functionality we expected, but open to being convinced otherwise.

Having the status say "Some issues found. Check your entries." even though they haven't started it yet is fine for now.

[...]

PR #6745, opened 2025-10-15
[Issue #6588] Remove not-started application form status

chouinar:
## Summary
Work for #6588

## Changes proposed
Remove application form status of `not_started` and logic that would set it.

## Context for reviewers
This status was causing issues with the frontend and was calculated in a very obtuse way. This basically merges what was `not_started` into the `in_progress` status.

The frontend already only uses `in_progress` so this fixes some issues for them.


babebe (approved):
LGTM

PR #6750, opened 2025-10-16
[unticketed] make room for long names and emails in org info

doug-s-nava:
## Summary

unticketed

## Changes proposed

some org info didn't have enough room to breathe. there may be a better solution but this is a quick fix before demos

## Validation steps

1. start a server with `npm run dev`
2. visit http://localhost:3000?_ff=userAdminOff:false
3. navigate to an organization page via workspace
4. _VERIFY_: user info is not squished

#### Before
Screenshot 2025-10-16 at 8 55 35 AM

#### After

Screenshot 2025-10-16 at 8 55 28 AM

### DIFF
--- a/frontend/tests/components/organization/__snapshots__/OrganizationInfo.test.tsx.snap
+++ b/frontend/tests/components/organization/__snapshots__/OrganizationInfo.test.tsx.snap
@@ -7,7 +7,7 @@ exports[`OragnizationInfo matches snapshot 1`] = `
     data-testid="grid"
   >
     <div
-      class="tablet:grid-col-3"
+      class="tablet:grid-col-4"
       data-testid="grid"
     >
       <span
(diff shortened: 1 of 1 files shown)


andycochran:
What's this look like if they're all `tablet={{ col: "auto" }}`?

andycochran:
I don't think these need to be dispersed across the horizontal space. It's fine if they're aligned left. Using `auto` would fit them to the content (here's a quick test of classes in the USWDS docs):  

image

…and any really long items would go to a second row if necessary: 

image

cc @crystabelrangel — Would you be okay with this layout change?

doug-s-nava:
with auto it looks like this. I can work with that but easier to go with what I have here just to get things working for demo today and we can iterate?

Screenshot 2025-10-16 at 11 40 44 AM

andycochran:
Looks like you need to add `gap` to the grid

Issue #6746, opened 2025-10-15
Optimize the new API Container Base Image


prasnava:
### Summary

Update the API docker image to pull the base image from the repo, instead of building it.

Related to the #2973

### Acceptance criteria

API and schedule step functions should continue to run without any errors.

PR #6756, opened 2025-10-16
[Issue #6746]: Optimize API docker build to use a cached base image

prasnava:
## Summary

Update the API docker image to use a cached base image instead of building each time.

## Changes proposed

Created a GitHub workflow to build and release the python base image. Updated the api dockerfile to use the new base image resulting in faster api docker builds.

## Validation steps

The api and step functions should continue to run without any errors.


prasnava:
Successfully deployed in training: [here](https://github.com/HHS/simpler-grants-gov/actions/runs/18595834301/job/53023137256)

Waiting for the scheduled cron step functions to run successfully before merging.

mdragon (approved):
👍🏻

Issue #6758, opened 2025-10-16
Organization application admin changes


eric-serrato95:
### Summary

Capturing the effort to prevent an individual to have admin privileges on an application when the application belongs to an organization 

### Acceptance criteria

- [ ] When a user starts an application as part of an organization, do not grant them application admin privileges

PR #6761, opened 2025-10-16
[Issue #6718] Update admin assignment

babebe:
## Summary

Work for #[ISSUE](https://github.com/HHS/simpler-grants-gov/issues/6758) 

## Changes proposed

When creating application, if not owned by organization, assign application owner role to user

### DIFF
--- a/api/tests/src/api/users/test_user_application_routes.py
+++ b/api/tests/src/api/users/test_user_application_routes.py
@@ -5,10 +5,12 @@ from src.constants.lookup_constants import ApplicationStatus, Privilege
 from tests.lib.application_list_utils import create_user_in_app
 from tests.lib.organization_test_utils import create_user_in_org
 from tests.src.db.models.factories import (
+    AgencyFactory,
     ApplicationFactory,
     ApplicationUserFactory,
     ApplicationUserRoleFactory,
     CompetitionFactory,
+    OpportunityFactory,
     OrganizationFactory,
     OrganizationUserFactory,
     OrganizationUserRoleFactory,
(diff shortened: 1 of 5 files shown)


api/src/services/applications/create_application.py
```diff
@@ -33,7 +33,7 @@ def _assign_application_owner_role(
     db_session: db.Session, application_user: ApplicationUser
 ) -> None:
     """Assign the Application Owner role to an application user."""
-
+    application_user.is_application_owner = True
```
chouinar: We could probably just drop this, I don't think it's used anymore in any of our code checks outside of unit tests - which should just stop checking for it as well.

babebe: dropped column

api/tests/src/api/users/test_user_application_routes.py
```diff
@@ -79,8 +79,14 @@ def test_user_get_applications_success(
     )
 
     # Associate user with applications
-    ApplicationUserFactory.create(user=user, application=application1, as_owner=True)
-    ApplicationUserFactory.create(user=user, application=application2, as_owner=True)
```
chouinar, line 94: The `as_owner` approach should still work, that wasn't using the flag we removed here, it's just giving them the owner role.

api/tests/src/services/applications/test_create_application.py
```diff
@@ -92,40 +90,8 @@ def test_validate_organization_expiration_expires_today():
...
-    )
-
-    assert role_assignment is not None
-    assert role_assignment.role_id == APPLICATION_OWNER.role_id
-
-
-def test_create_application_assigns_owner_role(db_session, enable_factory_create):
-    """Test that creating an application assigns the Application Owner role"""
+def test_create_application_assigns_owner_role_lone_application(db_session, enable_factory_create):
+    """Test that creating an application assigns the Application Owner role if lone application"""
```
chouinar: Non-organization applications are "individual applications" in our terminology.

Do we have a test for verifying a user doesn't get a role when starting an application? Fine to tack onto a test that creates an org app and just verify `application_users` on the app is empty.

babebe: added [here](https://github.com/HHS/simpler-grants-gov/pull/6761/files#diff-6874acc8eb4eb298ef1ec697add6a8d9b9da0fc3575c492178a0e41289c92786R2594)

Issue #6763, opened 2025-10-16
Wire up AuthorizationGate to user permission check endpoint


doug-s-nava:
### Summary
Once https://github.com/HHS/simpler-grants-gov/issues/6686 is complete, wire up AuthorizationGate to user permission check endpoint

### Acceptance criteria

Wire up AuthorizationGate to user permission check endpoint

Issue #6686, opened 2025-10-08
Create an API endpoint that can say whether a user has access to a specific privilege on a resource


chouinar:
### Summary

The general idea for this endpoint is a way for someone calling the API (our frontend mainly) can ask "does this user have this privilege in this resource", effectively an endpoint wrapping the `can_access` method.

We'll call this endpoint `POST /v1/users/:user_id/can_access`

# Request
```json
{
  "resource_id": "a UUID",
  "resource_type": "agency",
  "privileges": [ ... ]
}
```
Resource type should be an enum that can be any of:
* `agency`
* `application`
* `organization`

Privilege should be a list of privileges (minimum 1) that corresponds to our privilege enum.

TODO - Do we need to handle the weird case of our internal roles? We only have one planned use of that in the near-term and that won't matter to frontend at this time.

# Response
Success should just return a simple empty response, if the user doesn't have access, that'll be a standard 403 error.

# Auth
The only auth (beyond the logic described below) will be that the JWT user_id should match the user_id like all of our other endpoints.

# Logic
The ultimate goal of this endpoint is to make a call to our `can_access` method for the user, resource and privileges passed in.

For the resource logic, we'll need to fetch the resource from the agency/organization/application table based on the ID + resource type. The resource type should say which query we make. If the resource with that ID does not exist, the endpoint should return a 404.

### Acceptance criteria

- [ ] Endpoint created per the above guidelines
- [ ] Tests added that verify the behavior

babebe:
Does not include internal roles

PR #6765, opened 2025-10-16
[Issue #6674] authorization gate refactors

### LINKED ISSUE #6674: Update AuthorizationGate to hit new API endpoint for individual privilege checks
### Summary

Problem: the frontend needs to confirm application access authorization (particularly in relation to write and submit actions) via either application or organization based privileges. To check organization based privileges, the frontend would need an organization id related to the application, and the frontend will only have the application id to work with.

It'd be nice to make this determination without making any extra API calls, but since the organization id will not be available without making a call for application details, we can assume that a call will need to be made.

As such, we've decided to leverage the API's ability to easily determine access already, and make a call to a newly built API endpoint to check for any required permissions.

Solution: refactor AuthorizationGate into AuthorizedContentWrapper as described in the tech spec here
https://navasage.atlassian.net/wiki/x/UABFeg

In short, if an <AuthorizedContentWrapper> is called with requiredPermissions, instead of fetching the user's permissons and checking them against the required permissions, we will make calls to the endpoint built in https://github.com/HHS/simpler-grants-gov/issues/6686 for each required permission.

This also involves changes for:
* fetching permissions data as described above
* making onUnauthorized optional, and adding an ability for it to take in arguments for children and request results
* updating what is passed into context and props, and what shape it takes

This also means that the AuthorizedContentWrapper no longer needs to make a call to fetch the user's permissions, and any logic and types built out to support checking permissions on the frontend can be removed in this ticket

related to github.com/HHS/simpler-grants-gov/issues/6587 (a first attempt at solving this problem) and https://github.com/HHS/simpler-grants-gov/issues/6686 (implementing api endpoint to check individual perm for a user)

### Acceptance criteria

- [ ] given an an organization with id "1" and an application with id "2" and organization id "1", a user with no application based privileges but an organization privilege of 'modify_application' for organization id "1" is able to access content wrapped in an <AuthorizationGate> passed requiredPermissions of `[{ resourceId: "2", resourceType: "application", privileges: ["modify_application"]}]`
- [ ] types and code related to frontend privilege checks are removed
- [x] update tech spec with new approach
- [ ] refactors mentioned in tech spec are completed

doug-s-nava:
## Summary

Fixes #6674 

## Changes proposed

Updates the AuthorizationGate component in a number of ways to make it more flexible to handle real world auth situations, particularly around application permissions.

## Context for reviewers

* Instead of checking for permissions based access on the frontend, hit new API endpoint to check access (note that this endpoint is actively being built out in ticket #6686. This change uses a dummy response, wiring up will happen in #6763)
* make onUnauthorized optional
* unless there is an error or an unauthorized response handled with the onUnauthorized handler, always pass BOTH fetched resource data and fetched permission check results as child prop and via context
* passes children and resource call results to onUnauthorized to allow for more customized behavior

Note that this change relies on any resource calls returning throwing errors with a status code in the cause. This will likley mean refactoring some fetching methods down the line as we begin gating pages.

## Validation steps

1. _VALIDATE_: tests pass

### DIFF
--- a/frontend/src/components/user/AuthorizationGate.tsx
+++ b/frontend/src/components/user/AuthorizationGate.tsx
@@ -50,16 +50,17 @@ type AuthorizationGateProps = {
     }
 );
 
-// unpack the fetched resource maps to get the first actual error
-const findFirstError = (
+// unpack the fetched resource maps to determine if any of calls
+// resulted in a 403
+const findUnauthorizedError = (
   fetchedResources: FetchedResourceMap[],
-): FetchedResource | undefined => {
-  const firstError = fetchedResources.find((resource) =>
+): boolean => {
+  const hasUnauthorizedError = fetchedResources.some((resource) =>
     Object.values(resource).some(
-      (resourceDefinition) => resourceDefinition.statusCode !== 200,
+      (resourceDefinition) => resourceDefinition.statusCode === 403,
     ),
   );
-  return firstError ? Object.values(firstError)[0] : undefined;
+  return hasUnauthorizedError;
 };
 
 // we need to expose status codes and possible errors as well as fetched data
@@ -96,6 +97,9 @@ const resolveAndFormatResources = (
 
 // calls the API to check all required privileges,
 // and formats results into a format to be used downstream
+// note that if a non-403 is returned, that error will be passed along
+// since we'll consider a non-403 as unauthorized, children should check for errors
+// first before checking "authorized"
 const checkRequiredPrivileges = async (
   token: string,
   userId: string,
@@ -111,7 +115,7 @@ const checkRequiredPrivileges = async (
           if (parseErrorStatus(e) === 403) {
             return { ...privilege, authorized: false };
           }
-          throw e;
+          return { ...privilege, authorized: false, error: e.message };
         });
     }),
   );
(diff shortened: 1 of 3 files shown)


frontend/src/components/user/AuthorizationGate.tsx
```diff
@@ -52,69 +145,70 @@ export async function AuthorizationGate({
...
-        </FetchedResourcesProvider>
-      );
-    } catch (e) {
-      const error = e as Error;
-      if ((error.cause as FrontendErrorDetails).status === 403) {
-        return onUnauthorized(children);
+      // Note: we will handle errors based on the order that their respective promise was passed in. If multiple promises throw errors
+      // any after the first error encountered will be swallowed, as we have not accounted for the ability to handle multiple errors.
+      // We can improve this later
+      const firstError = findFirstError(fetchedResources);
```
doug-s-nava: erin suggested letting downstream consumers handle their own non 403 related errors in which case this should iterate through to see if there are any 403s

frontend/src/hooks/useAuthorizedData.tsx
```diff
@@ -0,0 +1,30 @@
+"use client";
```
doug-s-nava, line 1: this is just renaming from useFetchedResources

PR #6772, opened 2025-10-17
[Issue #6478] Build Footer XML

### LINKED ISSUE #6478: Build Footer XML
### Summary

The XML file contains a footer that includes data. We need to create a footer that meets the needs of existing grants.gov XML output.

### Acceptance criteria

- [ ] XML generated for header

mikehgrantsgov:
## Summary

Fixes #6478 

## Changes proposed

Add `generate_footer_xml` and supporting methods to `header_generator.py`.
Add tests
Clean up class / remove redundant comments

## Context for reviewers

The XML file contains a footer that includes data. We need to create a footer that meets the needs of existing grants.gov XML output.

Note the `HashValue` computation is still TBD.

## Validation steps

See new tests added.

### DIFF
--- a/api/tests/src/services/xml_generation/test_footer_generator.py
+++ b/api/tests/src/services/xml_generation/test_footer_generator.py
@@ -25,8 +25,8 @@ def application_with_submission(enable_factory_create, db_session):
     """Create an application with submitted_at, submitted_by, and submission."""
     # Create user with profile
     user = UserFactory.create()
-    UserProfileFactory.create(user=user, first_name="Michael", last_name="Chouinard")
-    LinkExternalUserFactory.create(user=user, email="michael@example.com")
+    UserProfileFactory.create(user=user, first_name="Example", last_name="User")
+    LinkExternalUserFactory.create(user=user, email="test@example.com")
 
     # Create application with submission data
     submitted_time = datetime(2025, 9, 2, 16, 24, 28, 0, tzinfo=timezone.utc)
@@ -81,7 +81,7 @@ class TestFooterGeneration:
         # Verify SubmitterName
         submitter_name_elem = root.find(f"{{{FOOTER_NAMESPACES['footer']}}}SubmitterName")
         assert submitter_name_elem is not None
-        assert submitter_name_elem.text == "Michael Chouinard"
+        assert submitter_name_elem.text == "Example User"
 
         # Verify Grants_govTrackingNumber
         tracking_number_elem = root.find(


api/src/services/xml_generation/header_generator.py
```diff
@@ -97,24 +134,75 @@ def _format_date(self, date_value: date | None) -> str | None:
...
+        """Get the submitter name from the submitted_by user.
+
+        Returns first_name + last_name if available, otherwise email, or 'unknown' if no user.
+        """
+        if not self.application.submitted_by:
+            return "unknown"
+
+        user = self.application.submitted_by_user
+        if not user:
+            return "unknown"
```
chouinar: Can just check the relationship directly.

```suggestion

        user = self.application.submitted_by_user
        if not user:
            return "unknown"
```

api/src/services/xml_generation/header_generator.py
```diff
@@ -97,24 +134,75 @@ def _format_date(self, date_value: date | None) -> str | None:
...
+
+        if user.profile and user.profile.first_name and user.profile.last_name:
+            return f"{user.profile.first_name} {user.profile.last_name}"
```
chouinar: First/last name aren't nullable, if they have a profile, they'll have a first/last name.

```suggestion
        if user.profile:
            return f"{user.profile.first_name} {user.profile.last_name}"
```

api/src/services/xml_generation/header_generator.py
```diff
@@ -97,24 +134,75 @@ def _format_date(self, date_value: date | None) -> str | None:
...
+        return base64.b64encode(sha1_hash.digest()).decode("utf-8")
+
+    def _calculate_footer_hash_value(
+        self, application_submission: ApplicationSubmission | None
+    ) -> str | None:
+        hash_data_parts = [
+            str(self.application.application_id) if self.application.application_id else "",
+            self._get_received_datetime() or "",
+            self._get_submitter_name() or "",
+        ]
```
chouinar, line 179: Could we calculate and pass in these values? That way we aren't re-calculating things like the timestamp? If an application didn't have a `submitted_at` assigned, this would actually be a different timestamp because it'd be a different call to `utcnow`

api/tests/src/services/xml_generation/test_footer_generator.py
```diff
@@ -0,0 +1,393 @@
...
+    UserProfileFactory.create(user=user, first_name="Michael", last_name="Chouinard")
+    LinkExternalUserFactory.create(user=user, email="michael@example.com")
+
+    # Create application with submission data
+    submitted_time = datetime(2025, 9, 2, 16, 24, 28, 0, tzinfo=timezone.utc)
+    application = ApplicationFactory.create(
+        competition=competition,
+        submitted_at=submitted_time,
+        submitted_by=user.user_id,
+    )
```
chouinar: The application factory can create the competition/opportunity, if we don't need to set anything on them, let's leave that out.

Can we adjust the example name? I feel weird having it be me / any real person.

```suggestion
    # Create user with profile
    user = UserFactory.create()
    UserProfileFactory.create(user=user, first_name="Michael", last_name="Chouinard")
    LinkExternalUserFactory.create(user=user, email="michael@example.com")

    # Create application with submission data
    submitted_time = datetime(2025, 9, 2, 16, 24, 28, 0, tzinfo=timezone.utc)
    application = ApplicationFactory.create(
        submitted_at=submitted_time,
        submitted_by=user.user_id,
    )
```

api/tests/src/services/xml_generation/test_footer_generator.py
```diff
@@ -0,0 +1,393 @@
...
+
+    # Create application submission
+    submission = ApplicationSubmissionFactory.create(
+        application=application,
+        legacy_tracking_number=80000001,
```
chouinar: The sequence that backs this starts counting from 80000000 - if you set it to that, any usage of the factory that doesn't set it will immediately conflict. We either shouldn't set the value and let the sequence handle it, or pick a number way more than the sequence number that our factory won't ever count to.

api/src/services/xml_generation/header_generator.py
```diff
@@ -69,17 +55,68 @@ def generate_header_xml(self, pretty_print: bool = True) -> str:
...
+            root, "ClosingDate", self._format_date(self.competition.closing_date), header_ns
+        )
+        self._add_field(root, "SubmissionTitle", self.application.application_name, header_ns)
+        self._add_field(root, "CFDANumber", self._get_cfda_number(), header_ns)
+
+        return self._generate_xml_string(root, pretty_print)
+
+    def generate_footer_xml(
+        self, pretty_print: bool = True, application_submission: ApplicationSubmission | None = None
+    ) -> str:
```
chouinar, line 89: Application submission should be required always, the footer is based on it, not having it isn't a valid scenario.

api/src/services/xml_generation/header_generator.py
mikehgrantsgov: Makes sense, I updated `_calculate_footer_hash_value` to take in `received_datetime` as a param

api/tests/src/services/xml_generation/test_footer_generator.py
mikehgrantsgov: I set this much higher for now e.g. `99000001` to avoid any of these conflicts.

api/src/services/xml_generation/header_generator.py
mikehgrantsgov: Got it, I tightened this up.

api/tests/src/services/xml_generation/test_footer_generator.py
```diff
@@ -0,0 +1,341 @@
...
+
+
+@pytest.fixture
+def application_with_submission(enable_factory_create, db_session):
+    """Create an application with submitted_at, submitted_by, and submission."""
+    # Create user with profile
+    user = UserFactory.create()
+    UserProfileFactory.create(user=user, first_name="Michael", last_name="Chouinard")
+    LinkExternalUserFactory.create(user=user, email="michael@example.com")
+
```
chouinar, line 30: Could we adjust the test examples to not be me / any real people?

Issue #6685, opened 2025-10-08
CI/CD build runner running out of disk space


mdragon:
### Summary

We saw Prod builds running out of space today. (Yes they shouldn't be building because we already have them built but vuln scans need to change).

### Acceptance criteria

We have a sense of the amount of space our builds use and we trim as much as we can as we progress through the build.

PR #6773, opened 2025-10-17
[Issue #6685] Better reporting on and conservation of disk space during CI/CD runner

mdragon:
## Summary

Work for #6685 

## Changes proposed

- Better checking and skipping of unnecessary steps if we don't actually need to build anything
- If we do need to build, additional logging of free space on disk throughout the process so we can know if we're running low on space
- Adopt Docker Build Cache approach from recent Platform Template changes to ensure we are trimming the cache run to run

## Context for reviewers

We were running out of disk space on the runner that does the vulnerability scans. This was partially due to:
- the amount of tools pre-installed on the runner
- the images we downloaded/ran as part of this step
- the build artifacts we kept around on disk after we didn't need them any more.

### DIFF
--- a/.github/workflows/vulnerability-scans.yml
+++ b/.github/workflows/vulnerability-scans.yml
@@ -77,9 +77,11 @@ jobs:
           key: ${{ inputs.app_name }}-docker-img-${{ needs.get-commit-hash.outputs.commit_hash }}
 
       - name: Set up Docker Buildx
+        if: steps.check-docker-image-cached.outputs.cache-hit != 'true'
         uses: docker/setup-buildx-action@master
 
       - name: Restore Docker layers
+        if: steps.check-docker-image-cached.outputs.cache-hit != 'true'
         id: restore-buildx
         uses: actions/cache/restore@v4
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/vulnerability-scans.yml
```diff
@@ -95,26 +109,53 @@ jobs:
...
+
+          rm -rf /tmp/.buildx-cache
+
+          echo Post-clean docker build cache file sizes
+          df -h
+          du -h -d 3 | grep -v ".git"
+
           docker save ${{ steps.shared-output.outputs.image }} > /tmp/docker-image.tar
+          echo Post-tar file sizes
+          df -h
```
chouinar, line 150: Could be a follow-up, but any thoughts on having something "if disk_space > 90% -> error".  Maybe not an error, but it'd be good if we had some alert that said we were creeping up again before it was a problem especially since disk space issues manifest in weird ways like OpenSearch auth

mdragon: I have been thinking about this. It feels like there's two pieces:
- An action step at the end of each job that would do that kind of check and do something to get our attention, maybe fire a Slack alert or something like that.
- A `if: failure()` action that would dump the free space info whenever a job fails so that we can see more easily if the failure may be related to running out of space.

I think it should be a follow up, but it's something that we don't want to lose track of so I'll spin up a ticket.

Issue #6666, opened 2025-10-08
Fetching the data from Simpler for GetApplicationZip


chris-kuryak:
### Summary

Fetching the data (database call) from Simpler

Make a query, grab the file from S3, return it.

### Acceptance criteria

- [ ] If you put in a Simpler tracking number to GetApplicationZip, you get a response from Simpler

PR #6774, opened 2025-10-17
[Issue #6666] Fetch data for simpler GetApplicationZipRequest

jakobpederson:
## Summary

Fixes for #ISSUE6666

## Changes proposed

Fetch data for simpler GetApplicationZipRequest

## Context for reviewers

Now actually doing a query against the DB to get the file location on s3 to add the file information to the mtom response.

### DIFF
--- a/api/src/legacy_soap_api/legacy_soap_api_client.py
+++ b/api/src/legacy_soap_api/legacy_soap_api_client.py
@@ -229,6 +229,22 @@ class SimplerGrantorsS2SClient(BaseSOAPClient):
             ),
         )
 
+    def _gen_response_data(
+        self, mime_message: bytes, boundary: str, mtom_file_stream: BinaryIO
+    ) -> Iterator:
+        yield mime_message
+        CHUNK_SIZE = 4000
+        try:
+            chunk = mtom_file_stream.read(CHUNK_SIZE)
+            while chunk:
+                yield chunk
+                chunk = mtom_file_stream.read(CHUNK_SIZE)
+                if not chunk:
+                    break
+        finally:
+            mtom_file_stream.close()
+        yield b"\n" + boundary.encode("utf-8") + b"--"
+
     def get_simpler_soap_response(self, proxy_response: SOAPResponse) -> SOAPResponse:
         if proxy_response.status_code != 500:
             return proxy_response
@@ -267,26 +283,9 @@ class SimplerGrantorsS2SClient(BaseSOAPClient):
             f"{simpler_response_xml}\n"
             f"{boundary}"
         ).encode("utf-8")
-
-        def _gen_response_data(
-            mime_message: bytes, boundary: str, mtom_file_stream: BinaryIO
-        ) -> Iterator:
-            yield mime_message
-            CHUNK_SIZE = 4000
-            try:
-                chunk = mtom_file_stream.read(CHUNK_SIZE)
-                while chunk:
-                    yield chunk
-                    chunk = mtom_file_stream.read(CHUNK_SIZE)
-                    if not chunk:
-                        break
-            finally:
-                mtom_file_stream.close()
-            yield b"\n" + boundary.encode("utf-8") + b"--"
-
         if mtom_file_stream:
             return get_soap_response(
-                data=_gen_response_data(mime_message, boundary, mtom_file_stream),
+                data=self._gen_response_data(mime_message, boundary, mtom_file_stream),
                 headers=update_headers,
             )
         return proxy_response


api/tests/src/legacy_soap_api/test_legacy_soap_api_client.py
```diff
@@ -341,3 +353,194 @@ def test_get_simpler_soap_response_when_operation_is_not_get_opportunity_list_re
                         mock_get_proxy_soap_response_dict.assert_not_called()
                     client.get_simpler_soap_response(proxy_response)
                     assert len(caplog.records) == 0
+
+
+class TestSimplerSOAPGetApplicationZip:
+    @pytest.fixture(autouse=True)
+    def truncate_competitions(self, db_session):
+        # This will truncate the competitions and related data for each test within this test class.
+        cascade_delete_from_db_table(db_session, ApplicationSubmission)
```
chouinar: You shouldn't need this. This should only be needed if you absolutely need a table to be empty before adding data.

Since this is a GET endpoint that given an ID, it returns a response, if there is other random data in the table, that shouldn't be an issue. Was there an issue you encountered?

api/src/legacy_soap_api/legacy_soap_api_client.py
```diff
@@ -264,4 +269,46 @@ def get_simpler_soap_response(self, proxy_response: SOAPResponse) -> SOAPRespons
             f"{simpler_response_xml}\n"
             f"{boundary}"
         ).encode("utf-8")
-        return get_soap_response(data=mime_message, headers=update_headers)
+
+        legacy_tracking_number = self.get_soap_request_dict()["GrantsGovTrackingNumber"]
```
chouinar: This looks to be the generic function for getting a response, not every call to this function would want the behavior below here. This should be in the service function that was implemented for GetApplicationZipResponse.

I am guessing you did this because you need the file object to start the stream - could you just have the service function return the file stream and process it here? That seems like a generic enough approach that we'd be able to reuse when we have to do this for other endpoints. Then this becomes:

```py
if not self.operation_config.is_mtom:
    ...

# build mtom response stuff

return get_soap_response(data=_gen_response_data(mime_message, boundary, proxy_response.file_stream), ...)
```

I'm not sure how closing the file handle will work, if that's something Flask does while returning a response?

api/src/legacy_soap_api/legacy_soap_api_client.py
```diff
@@ -264,4 +269,46 @@ def get_simpler_soap_response(self, proxy_response: SOAPResponse) -> SOAPRespons
...
+                yield chunk
+            yield b"\n" + boundary.encode("utf-8") + b"--"
+
+        if application:
+            response = requests.get(application.download_path, timeout=10)
+            if response.status_code == 200:
+                return get_soap_response(
+                    data=_gen_response_data(mime_message, boundary, response),
+                    headers=update_headers,
+                )
```
chouinar: Don't build a custom approach to fetching files from s3. The download_path is meant for public users and uses a presigned URL and is a very roundabout way of getting the file.

We have several file utils that can handle this - `file_util.open_stream` is probably what you'd want?

api/tests/src/legacy_soap_api/test_legacy_soap_api_client.py
jakobpederson: No, I didn't run into an issue. It's something I was intending to pull out if it wasn't doing anything.

api/src/legacy_soap_api/legacy_soap_api_client.py
jakobpederson: Flask _should_ be closing the files from what I can tell, but I'll close it manually just in case.

api/src/legacy_soap_api/legacy_soap_api_client.py
```diff
@@ -264,4 +267,26 @@ def get_simpler_soap_response(self, proxy_response: SOAPResponse) -> SOAPRespons
...
+            try:
+                chunk = mtom_file_stream.read(CHUNK_SIZE)
+                while chunk:
+                    yield chunk
+                    chunk = mtom_file_stream.read(CHUNK_SIZE)
+                    if not chunk:
+                        break
+            finally:
+                mtom_file_stream.close()
+            yield b"\n" + boundary.encode("utf-8") + b"--"
```
chouinar: Could we define this outside of this function? Unless we need a nested function definition for some sort of access (ie. the nested function accessing something from the context of the outer function), best to avoid nesting/complexity.

PR #6777, opened 2025-10-17
[Issue #6742] Enable CG routes in prod

### LINKED ISSUE #6742: [CommonGrants adoption] Enable CG routes in prod
### Summary

Resolve all remaining blockers to exposing the CommonGrants routes in the prod environment (potentially as sub-tickets), then update the feature flag to turn on the CG routes.

### Acceptance criteria

- [x] CommonGrants routes have been tested in staging
- [x] CommonGrants routes are available in prod

DavidDudas-Intuitial:
## Summary

Fixes #6742 

## Changes proposed

This PR: 
- enables CommonGrants routes in Prod via feature flag
- updates auth scheme to `api_user_key_auth`


api/openapi.generated.yml
```diff
@@ -614,20 +614,11 @@ paths:
...
-                $ref: '#/components/schemas/ErrorResponse'
-          description: Authentication error
       tags:
       - CommonGrants Protocol
       summary: List opportunities
       description: Get a paginated list of opportunities, sorted by `lastModifiedAt`
         with most recent first.
-      security:
-      - ApiUserKeyAuth: []
-      - ApiKeyAuth: []
```
widal001, line 630: Hmm it seems a little odd that swapping out one auth decorator for another would cause both security schemes to drop out of the OpenAPI docs.

Does @chouinar have any insight into why this might be happening?

DavidDudas-Intuitial: I assume it's because the new decorator `@api_user_key_auth` embeds the security scheme name in the auth token, whereas the old decorator `@api_multi_auth_key` does not and instead specifies the security scheme name in the route blueprint

DavidDudas-Intuitial: Notice that `security` schema name is no longer explicitly defined in the route blueprints:

Screenshot 2025-10-17 at 3 07 31 PM

DavidDudas-Intuitial: The OpenAPI docs are generated from the APIFlask route blueprints. If the route blueprints no longer contain an explicit security scheme name, because the scheme name is instead embedded in the auth token, then APIFlask has no knowledge of security scheme name when generating the OpenAPI docs.

widal001: Gotcha! Will this mean that OpenAPI visitors won't be able to test these endpoints?

My understanding was that the route-specific security details enabled users to pass the auth details to the endpoint correctly when they hit the "execute" button in the docs:

Screenshot 2025-10-20 at 9 21 13 AM

See how the other [routes in prod](https://api.simpler.grants.gov/openapi.json) still have a "security" attribute

chouinar: There are two ways that the auth can be attached to a route in a way that makes the right OpenAPI docs:
* You use `@common_grants_blueprint.auth_required(api_user_key_auth)` decorator - this tells API flask what auth you're using. This is the recommended way to do it, but unfortunately doesn't work with the multi-auth (like what you had before)
* If using multi-auth or setting up auth with `.login_required`, you have to define the `security=...` in the doc decorator. Unfortunately, APIFlask doesn't have something configured for the multiauth object and gets confused by it, so have to do it slightly more manually.

api/src/api/common_grants/common_grants_routes.py
```diff
@@ -49,11 +49,10 @@
 @common_grants_blueprint.get("/opportunities")
 @common_grants_blueprint.input(PaginatedQueryParamsSchema, location="query")
 @common_grants_blueprint.output(OpportunitiesListResponseSchema)
-@api_key_multi_auth.login_required
+@api_user_key_auth.login_required
```
chouinar: To get the docs working as expected for these routes

```suggestion
@common_grants_blueprint.auth_required(api_user_key_auth)
```

api/tests/conftest.py
```diff
@@ -410,6 +410,12 @@ def api_auth_token(monkeypatch, all_api_auth_tokens):
     return auth_token
 
 
+@pytest.fixture
+def api_user_key_token(user_api_key):
+    """Fixture providing a valid API user key for testing CG routes."""
+    return user_api_key.key_id
```
chouinar: There is an existing `user_api_key_id` fixture that is this already - it's just not quite with the other auth, not sure why it's at the top of the file.

widal001:
One other suggestion from my conversation with @mdragon this morning: It would be great if we could add a description to the "CommonGrants Protocol" tag. I think this should be as simple as replacing [this line](https://github.com/HHS/simpler-grants-gov/blob/96375a49986864dbc45715de18863f3a36e85c9e/api/src/api/common_grants/common_grants_blueprint.py#L11) with the following or something similar:

```python
tag={
    "name": "CommonGrants Protocol",
    "description": "CommonGrants-compliant API routes for searching opportunities to promote interoperability across grant systems. Learn more by visiting CommonGrants.org"
}
```

See the [tag attribute of the blueprint API](https://apiflask.com/api/blueprint/#apiblueprint)

api/tests/conftest.py
DavidDudas-Intuitial: Good catch. I removed this redundant code in https://github.com/HHS/simpler-grants-gov/pull/6777/commits/363e22672b6d372e7cd6702c55db6f2e134bfdc1

api/src/api/common_grants/common_grants_routes.py
DavidDudas-Intuitial: Thanks for the pointer. I made this change in https://github.com/HHS/simpler-grants-gov/pull/6777/commits/cb98792959408df2ea431982e5af078966596e36

api/openapi.generated.yml
DavidDudas-Intuitial: Thanks for the explanation. I updated the routes to use `@common_grants_blueprint.auth_required(api_user_key_auth)` in https://github.com/HHS/simpler-grants-gov/pull/6777/commits/cb98792959408df2ea431982e5af078966596e36

DavidDudas-Intuitial:
@chouinar @widal001 Thanks for the suggestions! PR updated. PTAL when you have a moment.

api/src/api/common_grants/common_grants_blueprint.py
```diff
@@ -8,7 +8,10 @@
 common_grants_blueprint = APIBlueprint(
     "common_grants",
     __name__,
-    tag="CommonGrants Protocol",
+    tag={
+        "name": "CommonGrants Protocol",
+        "description": "CommonGrants-compliant API routes for searching opportunities to promote interoperability across grant systems. Learn more by visiting CommonGrants.org",
```
chouinar, line 13: If you'd like the URL to be clickable, can use markdown like so:

```suggestion
        "description": "CommonGrants-compliant API routes for searching opportunities to promote interoperability across grant systems. Learn more by visiting [CommonGrants.org](https://commongrants.org)",
```

PR #6783, opened 2025-10-20
[Issue #6686] create can access endpoint

### LINKED ISSUE #6686: Create an API endpoint that can say whether a user has access to a specific privilege on a resource

babebe:
## Summary

Fixes #6686  

## Changes proposed

POST `/v1/users/:user_id/privileges` endpoint for checking user access on a resource ( organization, application or  agency) 
Tests added


api/src/api/users/user_schemas.py
```diff
@@ -20,6 +21,12 @@
 from src.pagination.pagination_schema import generate_pagination_schema
 
 
+class ResourceSchema(Enum):
```
chouinar: Some of our tooling prefers StrEnum / it handles printing out/converting to a string a bit easier.

```suggestion
class ResourceSchema(StrEnum):
```

api/src/services/users/user_can_access.py
```diff
@@ -0,0 +1,43 @@
...
+from src.db.models.competition_models import Application
+from src.db.models.entity_models import Organization
+from src.db.models.user_models import User
+
+# Map ResourceSchema to models
+RESOURCE_MAP = {
+    ResourceSchema.AGENCY: (Agency, "agency_id"),
+    ResourceSchema.APPLICATION: (Application, "application_id"),
+    ResourceSchema.ORGANIZATION: (Organization, "organization_id"),
+}
```
chouinar, line 20: SQLAlchemy is weird in convenient ways - you can avoid the getattr approach and just directly define the column like this and it'll work (some of the transform code on the backend does similar).

```suggestion
RESOURCE_MAP = {
    ResourceSchema.AGENCY: (Agency, Agency.agency_id),
    ResourceSchema.APPLICATION: (Application, Application.application_id),
    ResourceSchema.ORGANIZATION: (Organization, Organization.organization_id),
}
```

api/src/services/users/user_can_access.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def _get_resource(
+    db_session: db.Session, resource_type: ResourceSchema, resource_id: UUID
+) -> ApiSchemaTable | None:
```
chouinar: Let's have this specifically defined / it's not able to be null.

```suggestion
) -> Organization | Application | Agency:
```

api/src/services/users/user_can_access.py
```diff
@@ -0,0 +1,43 @@
...
+
+    resource = db_session.execute(stmt).scalar_one_or_none()
+    if not resource:
+        raise_flask_error(404, f"Resource ID {resource_id} of type {resource_type} not found")
+
+    return resource
+
+
+def check_user_can_access(db_session: db.Session, user_id: UUID, json_data: dict) -> None:
+    user = db_session.execute(select(User).where(User.user_id == user_id)).scalar_one_or_none()
```
chouinar: The user is attached to the token, we can pass that in and save this query.

```suggestion
def check_user_can_access(db_session: db.Session, user: User, json_data: dict) -> None:

```

api/src/api/users/user_routes.py
```diff
@@ -627,7 +630,6 @@ def user_profile_update(
 @flask_db.with_db_session()
 def user_get_roles_and_privileges(db_session: db.Session, user_id: UUID) -> response.ApiResponse:
     """Get the roles and privileges for the authenticated user"""
-    logger.info("POST /v1/users/:user_id/privileges")
```
chouinar, line 630: Why was this removed?

babebe: duplicate

PR #6786, opened 2025-10-20
[Issue #6785] Add new X-API-KEY to frontend

### LINKED ISSUE #6785: Update frontend env variable to call api
### Summary

We need to make the frontend call the API using the new x-api key instead of the old API key. 

### Acceptance criteria

Updated frontend.

btabaska:
## Summary

Fixes #6785 

## Changes proposed

Updated frontend to grab new X-API-KEY param from AWS param store & use it when making calls to the backend. 

## Context for reviewers

Used same implementation as we are already using for the old API. Left it in because we want there to be a fallback as we roll out API GW.


btabaska:
@chouinar This change is for the requested API gateway env variable change.

infra/frontend/app-config/env-config/environment_variables.tf
```diff
@@ -119,9 +119,9 @@ locals {
       manage_method     = "manual"
       secret_store_name = "/api/${var.environment}/api-jwt-public-key"
     },
-    API_AUTH_X_API_KEY = {
+    API_GW_AUTH = {
       manage_method     = "manual"
-      secret_store_name = "/${var.app_name}/${var.environment}/X-API-KEY"
```
doug-s-nava, line 124: [nit] you shouldn't need to hardcode `frontend` here, though I suppose it doesn't particularly matter. Curious if something was breaking otherwise though

chouinar:
> testing locally, I get a failure on a search call when putting in a random value for the "API_GW_AUTH" env var. I imagine that's expected, but wanted to make sure that what we want is:
> 
> * gateway auth token: none, regular auth token: valid - success
> * gateway auth token: invalid, regular auth token: valid - failure
> * gateway auth token: valid, regular auth token: valid - success (haven't tested this case but assuming)
> 
> want to make sure I'm understanding the use case here

The way the API handles the two auths is roughly:
* If both are provided, it'll just check the new API gateway related auth
* If only one is provided, it'll use that one

So if the API gateway auth is invalid, the value of the other auth doesn't matter. It just iterates over them in a specified order, checking if the header is present and validates the first one it finds.

btabaska:
> testing locally, I get a failure on a search call when putting in a random value for the "API_GW_AUTH" env var. I imagine that's expected, but wanted to make sure that what we want is:
> 
> * gateway auth token: none, regular auth token: valid - success
> * gateway auth token: invalid, regular auth token: valid - failure
> * gateway auth token: valid, regular auth token: valid - success (haven't tested this case but assuming)
> 
> want to make sure I'm understanding the use case here

This is meant to be paired with the PR https://github.com/HHS/simpler-grants-gov/pull/6787/ to seed the local user. without that change also merged this one will have issues. Sorry for not adding the context to the top level.

infra/frontend/app-config/env-config/environment_variables.tf
```diff
@@ -119,7 +119,7 @@ locals {
       manage_method     = "manual"
       secret_store_name = "/api/${var.environment}/api-jwt-public-key"
     },
-    API_AUTH_X_API_KEY = {
+    API_GW_AUTH = {
       manage_method     = "manual"
       secret_store_name = "/${var.app_name}/${var.environment}/X-API-KEY"
```
mdragon, line 124: Dos this need to be fixed to reference a different param store name? Do we need to be able to run with both keys configured in the codebase/AWS so that setting the params on and off lets us test with both keys and then remove the old key?

PR #6787, opened 2025-10-20
 [Issue #6784] Add new X-API-KEY to seed script

### LINKED ISSUE #6784: Update local seed script to create api user
### Summary

When we move to the new API we will need to update the local seed script for local development to create an API user in the DB with a new x-api key. 

### Acceptance criteria

Seed script updated

btabaska:
## Summary

Fixes #6784 

## Changes proposed

Updates the backend seed scripts to create DB users with new API keys for use in local dev

### DIFF
--- a/api/tests/lib/seed_local_db.py
+++ b/api/tests/lib/seed_local_db.py
@@ -274,7 +274,7 @@ def seed_local_db(iterations: int, cover_all_agencies: bool, steps: list[str]) -
         seed_agencies="ALL" in steps or "agencies" in steps,
         seed_opportunities="ALL" in steps or "opps" in steps,
         seed_forms="ALL" in steps or "forms" in steps,
-        seed_users="ALL" in steps or "users" in steps or "api" in steps,
+        seed_users="ALL" in steps or "users" in steps,
         seed_e2e="ALL" in steps or "e2e" in steps,
     )
 


api/tests/lib/seed_orgs_and_users.py
```diff
@@ -203,6 +203,25 @@
...
+###############################
+# API-only user for local development
+###############################
+API_USER = factories.UserFactory.build(user_id=uuid.UUID("12345678-1234-5678-9abc-123456789abc"))
+
+LINK_API_USER = factories.LinkExternalUserFactory.build(
+    link_external_user_id=uuid.UUID("87654321-4321-8765-cba9-987654321cba"),
+    external_user_id="api_user",  # THIS IS WHAT WE USE TO LOGIN
+    user=API_USER,
+)
```
chouinar: We shouldn't add this. This is the thing that connects it to OAuth, which we didn't do non-locally.

api/tests/lib/seed_local_db.py
```diff
@@ -274,7 +274,7 @@ def seed_local_db(iterations: int, cover_all_agencies: bool, steps: list[str]) -
         seed_agencies="ALL" in steps or "agencies" in steps,
         seed_opportunities="ALL" in steps or "opps" in steps,
         seed_forms="ALL" in steps or "forms" in steps,
-        seed_users="ALL" in steps or "users" in steps,
+        seed_users="ALL" in steps or "users" in steps or "api" in steps,
```
chouinar: I think I'd just leave `api` as a choice out, we always want it with the other users anyways.

documentation/api/local-seed-data.md
chouinar, line 1: Thank you for updating this (with more than just your change even!)

api/tests/lib/seed_local_db.py
btabaska: Updated based on feedback;

Issue #6817, opened 2025-10-22
Remove graphviz from the api docker image


prasnava:
### Summary

Reduce the api docker image size by removing graphviz from the api.

PR #6818, opened 2025-10-22
[Issue #6817]: Remove graphviz deps from the api docker image

prasnava:
## Summary

Removed graphviz deps from the api docker image to reduce the image size. The ci-openapi.yml github actions workflow has been updated to run manually instead of automatically on each pull request. We will use a separate Dockerfile for running the ci-openapi.yml workflow.

## Validation steps

The api and step functions should continue to run.

### DIFF
--- a/api/openapi.generated.yml
+++ b/api/openapi.generated.yml
@@ -540,7 +540,8 @@ paths:
       tags:
       - Form Alpha
       summary: Form Get
-      security:
+      security: &id001
+      - ApiUserKeyAuth: []
       - ApiKeyAuth: []
     put:
       parameters:
@@ -582,8 +583,7 @@ paths:
           application/json:
             schema:
               $ref: '#/components/schemas/FormUpdateRequest'
-      security:
-      - ApiKeyAuth: []
+      security: *id001
   /common-grants/opportunities:
     get:
       parameters:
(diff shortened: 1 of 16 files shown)


prasnava:
Successfully deployed to training: [here](https://github.com/HHS/simpler-grants-gov/actions/runs/18730691388/job/53427866088).

.github/workflows/ci-openapi.yml
```diff
@@ -2,11 +2,12 @@
...
-    paths:
-      - api/**
-      - Makefile
-      - .github/workflows/ci-openapi.yml
+  workflow_dispatch:
+    inputs:
+      version:
+        required: true
+        default: "main"
+        description: Tag or branch or SHA to deploy
```
chouinar, line 10: Is this only triggered manually now?

api/Dockerfile
```diff
@@ -1,54 +1,64 @@
...
 # Stage 3: Dev image
 #==============================================
 FROM base AS dev
 ARG RUN_USER
+ARG PY_MM=3.13
 
-# Graphviz is already present in the base image; no extra COPYs needed for dev.
-# Rely on the base image for Graphviz binaries, plugins, fonts, and runtime libs.
-# (Removing redundant COPYs avoids duplicating OS libraries and reduces image size.)
+# Disable Python optimization to preserve docstrings required by jsonpath_ng
```
chouinar, line 58: I assume this is referring to https://github.com/h2non/jsonpath-ng?tab=readme-ov-file#special-note-about-ply-and-docstrings ?

Could we add that link for future reference

api/Dockerfile
```diff
@@ -74,33 +81,43 @@ CMD ["poetry", "run", "flask", "--app", "src.app", "run", "--host", "0.0.0.0", "
 #==============================================
 # Stage 5: Minimal release runtime
 #==============================================
+# hadolint ignore=DL3006
 FROM base AS release
-ARG RUN_USER
+ARG PY_MM=3.13
```
chouinar: No idea if this is possible, but are we able to define the Python version in the base image and reference that elsewhere? Thinking about how we'll need to update to Python 3.14 soon and hoping we can do that with as few changes as possible.

.github/workflows/ci-openapi.yml
prasnava: yeh manually for now, until the next pull request after the new container is created.

api/Dockerfile
prasnava: added the link for future reference.

api/Dockerfile
prasnava: Yes,  args are not persisted in multistage build but we can pass args via text file from stage to stage. Working on that.

chouinar (approved):

PR #6820, opened 2025-10-22
[Issue #6453] Make forms endpoints use new API key auth

### LINKED ISSUE #6453: Modify form endpoints and scripts to rely on new RBAC auth instead of hacky key approach
### Summary

We want to make a few updates to our process for updating forms to handle changes to the way API key auth works.

We want to add a new role like:
```py
NAVA_INTERNAL_ROLE_ID = uuid.UUID("some uuid")
NAVA_INTERNAL_ROLE = Role(
    role_id=NAVA_INTERNAL_ROLE_ID,
    role_name="Nava Internal Admin Role",
    is_core=True,
    link_privileges=get_link_privileges(NAVA_INTERNAL_ROLE_ID, [Privilege.UPDATE_FORM]),
    link_role_types=[LinkRoleRoleType(role_id=NAVA_INTERNAL_ROLE_ID, role_type=RoleType.INTERNAL)],
)

# Also add this to the CORE_ROLES list in static_role_values.py
```
---
We want to modify the GET form endpoint to use the new API key auth (`api_user_key_auth` instead of `api_key_auth`).

We also want to modify the PUT forms endpoint to use the new type of API key auth. We can remove the extra logic that verified it was the "first" token and instead have it verify that the calling user has `Privilege.UPDATE_FORM` privilege using the new logic for validating privileges (pass in no resource, only internal roles should be able to access this).

---
Finally, we want to adjust the script that updates forms to change the auth header we pass in.

### Acceptance criteria
- [ ] New role created
- [ ] Tests updated for each endpoint to cut over to new auth (it's fine to not support the old auth anymore, these are only used internally at the moment)
- [ ] Tests updated for the update-form script

chouinar:
## Summary
Fixes #6453

## Changes proposed
* Modify the GET form and PUT form endpoints to use new API key auth with RBAC
* Modify the list/update form scripts to be able to pass in the auth
* Add a new role for internal developers to use

## Context for reviewers
Because of the clunky way we need to do the two different approaches to auth in update-forms, the auth logic is in the route as piping through all the various types that the auth token can be would be a headache. When we later remove the old auth approach, we can move it and simplify a bunch of the logic / type issues.

## Validation steps
Will do some manual testing after merging in dev/staging once I've setup my own personal user to have the roles.

### DIFF
--- a/api/openapi.generated.yml
+++ b/api/openapi.generated.yml
@@ -538,7 +538,8 @@ paths:
       tags:
       - Form Alpha
       summary: Form Get
-      security:
+      security: &id001
+      - ApiUserKeyAuth: []
       - ApiKeyAuth: []
     put:
       parameters:
@@ -580,8 +581,7 @@ paths:
           application/json:
             schema:
               $ref: '#/components/schemas/FormUpdateRequest'
-      security:
-      - ApiKeyAuth: []
+      security: *id001
   /common-grants/opportunities:
     get:
       parameters:
@@ -2463,7 +2463,7 @@ components:
       properties:
         pagination_info:
           description: The pagination information for paginated endpoints
-          type: &id001
+          type: &id002
           - object
           $ref: '#/components/schemas/PaginationInfo'
         message:
@@ -2578,7 +2578,7 @@ components:
       properties:
         pagination_info:
           description: The pagination information for paginated endpoints
-          type: *id001
+          type: *id002
           $ref: '#/components/schemas/PaginationInfo'
         message:
           type: string
@@ -2769,7 +2769,7 @@ components:
       properties:
         pagination_info:
           description: The pagination information for paginated endpoints
-          type: *id001
+          type: *id002
           $ref: '#/components/schemas/PaginationInfo'
         message:
           type: string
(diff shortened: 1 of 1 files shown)


api/src/api/form_alpha/form_route.py
```diff
@@ -32,20 +38,32 @@ def form_get(db_session: db.Session, form_id: uuid.UUID) -> response.ApiResponse
...
     with db_session.begin():
+        multi_auth_user = api_key_multi_auth.get_user()
+        # If the old API key approach is used (only allowed temporarily)
+        # then we need to verify it's the internal auth token we haven't shared
+        if multi_auth_user.auth_type == AuthType.API_KEY_AUTH:
+            # Check if user is the internal admin user (auth_token_0)
+            if cast(ApiKeyUser, multi_auth_user.user).username != "auth_token_0":
+                raise_flask_error(403, "Only internal admin users can update forms")
+        else:
+            # This is only temporary until we remove the other way of calling
```
mikehgrantsgov, line 59: If a ticket isn't there to move this to the service layer, should we create it?

mikehgrantsgov (approved):
Looks good!

api/src/api/form_alpha/form_route.py
chouinar: I'll check with Brandon, we'll have a general "remove the old auth" set of tickets that we'll handle this in.

Issue #6662, opened 2025-10-08
Define schemas for GetSubmissionListExpanded


chris-kuryak:
### Summary

Define schemas (processing request from user and the thing that we shove the response in at the end).

Given the XML, it converts into python data types, and vice versa.

### Acceptance criteria
- [ ] Converts request XML to Python (in a format that we can use)
- [ ] Converts response Python into XML
- [ ] Write tests to confirm it works as expected

PR #6823, opened 2025-10-22
[Issue #6662] Define schemas for GetSubmissionListExpanded

jakobpederson:
## Summary

Fixes  #ISSUE6662 

## Changes proposed

Added schemas and response schemas for GetSubmissionListExpanded

## Context for reviewers

This should allow us to consume the incoming xml on the request and transform it into a python object, and take a response model and transform it into a dictionary that can be turned into xml.

### DIFF
--- a/api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
+++ b/api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
@@ -63,21 +63,24 @@ def get_submission_list_expanded_response(
     db_session: db.Session,
     get_submission_list_expanded_request: schemas.GetSubmissionListExpandedRequest,
 ) -> schemas.GetSubmissionListExpandedResponse:
-    stmt = select(ApplicationSubmission).options(
-        selectinload(ApplicationSubmission.application).options(
-            selectinload(Application.organization),
-            selectinload(Application.competition).options(
-                selectinload(Competition.opportunity),
-                selectinload(Competition.opportunity_assistance_listing),
-            ),
-        )
-    )
     stmt = (
-        stmt.join(ApplicationSubmission.application)
+        select(ApplicationSubmission)
+        .join(ApplicationSubmission.application)
+        # Joins for where clauses
         .join(Application.competition)
         .join(Opportunity, Competition.opportunity)
         .join(
             OpportunityAssistanceListing, Competition.opportunity_assistance_listing, isouter=True
+        )
+        # Prefetch values for better performance
+        .options(
+            selectinload(ApplicationSubmission.application).options(
+                selectinload(Application.organization),
+                selectinload(Application.competition).options(
+                    selectinload(Competition.opportunity),
+                    selectinload(Competition.opportunity_assistance_listing),
+                ),
+            )
         )
     )
     grants_gov_tracking_numbers = None


api/src/legacy_soap_api/grantors/services/get_application_list_expanded_response.py
```diff
@@ -0,0 +1,85 @@
...
+
+def get_application_list_expanded_response(
+    db_session: db.Session,
+    get_application_list_expanded_request: schemas.GetApplicationListExpandedRequest,
+) -> schemas.GetApplicationListExpandedResponse:
+    tracking_numbers = None
+    if application_filter := get_application_list_expanded_request.expanded_application_filter:
+        tracking_numbers = application_filter.grants_gov_tracking_numbers
+
+    stmt = select(ApplicationSubmission).options(
```
jakobpederson: Did we want to filter on some other attributes for the models?

api/src/legacy_soap_api/grantors/schemas/get_application_list_expanded_schemas.py
```diff
@@ -0,0 +1,56 @@
+from datetime import datetime
+
+from pydantic import BaseModel, Field, field_validator
+
+from src.legacy_soap_api.legacy_soap_api_schemas import BaseSOAPSchema
+
+
+class ExpandedApplicationInfo(BaseSOAPSchema):
```
chouinar: I don't think this is the right endpoint. There's a `GetSubmissionListExpanded` and also an older `GetApplicationListExpanded`. I assume we'd want the submission endpoint.

https://grants.gov/system-to-system/grantor-system-to-system/web-services

api/src/legacy_soap_api/grantors/services/get_application_list_expanded_response.py
```diff
@@ -0,0 +1,85 @@
...
+        "CFDANumber": assistance_listing.assistance_listing_number if assistance_listing else None,
+        "GrantsGovTrackingNumber": f"GRANT{submission.legacy_tracking_number}",
+        "GrantsGovApplicationStatus": GRANTS_APPLICATION_STATUSES[application.application_status],
+        "SubmissionTitle": application.application_name,
+        "PackageID": competition.legacy_package_id,
+        "n2:ReceivedDateTime": (
+            application.submitted_at.astimezone(ZoneInfo("America/New_York"))
+            if application.submitted_at
+            else None
+        ),
```
chouinar: It's similar, but we have a `datetime_util.adjust_timezone` function that would help simplify this a bit / has been tested to make sure it actually converts right.

api/src/legacy_soap_api/grantors/services/get_application_list_expanded_response.py
```diff
@@ -0,0 +1,85 @@
...
+        sam_gov_entity = organization.sam_gov_entity
+        application_list_obj.update(
+            {
+                "DelinquentFederalDebt": (
+                    "Yes" if sam_gov_entity and sam_gov_entity.has_debt_subject_to_offset else "No"
+                ),
+                "ActiveExclusions": (
+                    "Yes" if sam_gov_entity and sam_gov_entity.has_exclusion_status else "No"
+                ),
+            }
```
chouinar: Might be good to have a utility for common conversions, guessing we'll have a decent number of bool to yes/no conversions needed.

api/src/legacy_soap_api/grantors/services/get_application_list_expanded_response.py
```diff
@@ -0,0 +1,85 @@
...
+    tracking_numbers = None
+    if application_filter := get_application_list_expanded_request.expanded_application_filter:
+        tracking_numbers = application_filter.grants_gov_tracking_numbers
+
+    stmt = select(ApplicationSubmission).options(
+        joinedload(ApplicationSubmission.application).options(
+            joinedload(Application.organization),
+            joinedload(Application.competition).options(
+                joinedload(Competition.opportunity),
+                joinedload(Competition.opportunity_assistance_listing),
```
chouinar: Don't use `joinedload`, it hits issues with large joins from my experience when used at scale (queries locking up).

Use `selectinload`.

api/src/legacy_soap_api/grantors/services/get_application_list_expanded_response.py
chouinar: The other filters in the request schema would be needed.

api/src/legacy_soap_api/grantors/services/get_application_list_expanded_response.py
```diff
@@ -0,0 +1,85 @@
...
+            ),
+        )
+    )
+    if tracking_numbers:
+        stmt = stmt.where(ApplicationSubmission.legacy_tracking_number.in_(tracking_numbers))
+
```
chouinar: You can't do it yet, but can you add a comment that the agency from the certificate needs to be added to the query here?

Also - can we make sure this endpoint specifically is not enabled anywhere non-locally?

jakobpederson: I didn't do anything to implement the endpoint, so in the config there's only two so any calls to the endpoint should throw an operation not found exception.
```
SIMPLER_SOAP_OPERATION_CONFIGS: dict[SimplerSoapAPI, dict[str, SOAPOperationConfig]] = {
    SimplerSoapAPI.APPLICANTS: {
        "GetOpportunityListRequest": SOAPOperationConfig()
    },
    SimplerSoapAPI.GRANTORS: {
        "GetApplicationZipRequest": SOAPOperationConfig()
}
```

api/src/legacy_soap_api/grantors/schemas/get_submission_list_expanded_schemas.py
```diff
@@ -0,0 +1,113 @@
...
+                consolidated[filter_item.filter_type].extend(filter_item.filter_value)
+        elif len(type_list) > len(value_list):
+            raise SOAPInvalidEnvelope(
+                "The content of element 'ExpandedApplicationFilter' is not complete. One of '{\"http://apply.grants.gov/system/GrantsCommonElements-V1.0\":FilterValue}' is expected."
+            )
+        elif len(type_list) == 0 and len(value_list) > 0:
+            raise SOAPInvalidEnvelope(
+                "The content of element 'ExpandedApplicationFilter' is not complete. One of '{\"http://apply.grants.gov/system/GrantsCommonElements-V1.0\":FilterType}' is expected."
+            )
+        return {"filters": consolidated}
```
chouinar: Want to highlight something in how you've styled your code - This code has zero empty lines or comments in it which makes it hard to read. Could you adjust it a bit to group code together that goes together?

For example, in this you have this rough pattern:
```python
if x:
   pass
if y:
   pass
if z:
   pass
other code
if a:
   pass
elif b:
   pass
elif c:
   pass
```

It's hard to tell at a glance which if-statement those elif belong to. If instead that looks like:

```python
# Fix the inputs
if x:
   pass
if y:
   pass
if z:
   pass

other code

# Process the models
if a:
   pass
elif b:
   pass
elif c:
   pass
```
It's a lot easier to parse at a glance.

Without any comments, I'm not sure what exactly the lists being processed here are / what is being transformed.

api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
```diff
@@ -0,0 +1,119 @@
...
+                    )
+                ),
+                "ActiveExclusions": (
+                    convert_bool_to_yes_no(
+                        sam_gov_entity is not None and sam_gov_entity.has_exclusion_status
+                    )
+                ),
+                "UEI": sam_gov_entity.uei if sam_gov_entity else None,
+            }
+        )
```
chouinar, line 58: Since the sam gov entity is really what you want, don't check if it's null in each separate place, just don't do anything if it's null.

```suggestion
    # something roughly like this syntax, not certain this is exactly valid
    if application.organization and sam_gov_entity := application.organization.sam_gov_entity:
        application_list_obj["DelinquentFederalDebt"] = convert_bool_to_yes_no(sam_gov_entity.has_debt_subject_to_offset)
        application_list_obj["ActiveExclusions"] = convert_bool_to_yes_no(sam_gov_entity.has_exclusion_status)
        application_list_obj["UEI"] = sam_gov_entity.uei 

```

api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
```diff
@@ -0,0 +1,119 @@
...
+            .where(ListingAlias.assistance_listing_number.in_(cfda_numbers))
+        )
+    if funding_opportunity_numbers:
+        OpportunityAlias = aliased(Opportunity)
+        stmt = (
+            stmt.join(ApplicationSubmission.application)
+            .join(Application.competition)
+            .join(OpportunityAlias, Competition.opportunity)
+            .where(OpportunityAlias.opportunity_number.in_(funding_opportunity_numbers))
+        )
```
chouinar: Do you need to use aliased queries here?

It might sound a bit odd, but what if we just always had the joins in the default query? Then these would just need to be `stmt = stmt.where(...)`

Also, you'll want to be careful about joining the Competition to its assistance listing, that can be null (don't have exact numbers off the top of my head, but like 30% of competitions have no assistance listing number). You'll need to make sure you do an outer join like `.join(OpportunityAssistanceListing, isouter=True)`

api/src/legacy_soap_api/grantors/schemas/get_submission_list_expanded_schemas.py
jakobpederson: That's a good point, I'll try to avoid that in the future. Luckily, I discovered that I made the wrong assumption about the formatting of the filter which has simplified this a lot.

api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
jakobpederson: Good catch.

api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
jakobpederson: It turns out I don't need the aliased queries. I also consolidated the joins with the isouter and it works.

api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
```diff
@@ -0,0 +1,110 @@
...
+        "PackageID": competition.legacy_package_id,
+        "n2:ReceivedDateTime": (
+            adjust_timezone(application.submitted_at, "America/New_York")
+            if application.submitted_at
+            else None
+        ),
+        "SubmissionMethod": "web",
+        "DelinquentFederalDebt": None,
+        "ActiveExclusions": None,
+        "UEI": None,
```
chouinar, line 45: We can follow-up on it, but this might actually not default to null, usually the old system makes individuals have a UEI of `00000000INDV` - but we'd need to test that probably. I tried looking at their code and it calls a view - which I couldn't get to ever return anything.

api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
```diff
@@ -0,0 +1,110 @@
...
+            ),
+        )
+    )
+    stmt = (
+        stmt.join(ApplicationSubmission.application)
+        .join(Application.competition)
+        .join(Opportunity, Competition.opportunity)
+        .join(
+            OpportunityAssistanceListing, Competition.opportunity_assistance_listing, isouter=True
+        )
```
chouinar: We can put the boilerplate query altogether. If you want to split it up a bit for organizing, can use comments. 

```suggestion
    stmt = select(ApplicationSubmission)
        # Joins for where clauses
        .join(ApplicationSubmission.application)
        .join(Application.competition)
        .join(Opportunity, Competition.opportunity)
        .join(OpportunityAssistanceListing, Competition.opportunity_assistance_listing, isouter=True)
        # Prefetch values for better performance
        .options(
        selectinload(ApplicationSubmission.application).options(
            selectinload(Application.organization),
            selectinload(Application.competition).options(
                selectinload(Competition.opportunity),
                selectinload(Competition.opportunity_assistance_listing),
            ),
        )
    )

```

api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
jakobpederson: Would it be best to hold of merging until we get that resolved?

chouinar: I don't think we need to hold, worst case we need to come back to this and change the value, but should note down to test it.

PR #6831, opened 2025-10-23
[ISSUE #6621] Get User Invitations Endpoint

### LINKED ISSUE #6621: Get User Invitations Endpoint
### Summary

Implement the endpoint that allows users to view their pending invitations by matching their email address, following the existing user route patterns.

#### Endpoint Specification
- **Method**: POST
- **Path**: `/v1/users/{user_id}/invitations/list`
- **Authentication**: Required (JWT)
- **Authorization**: User can only see their own invitations (authenticated user must match user_id)
- **Email Matching**: Match invitations using user's login.gov email

Request Schema
```json
{
  "filters": {
    "status": {
      "one_of": ["pending", "accepted", "rejected", "expired"]
    },
}
```

#### Response Schema
```json
{
  "message": "Success",
  "data": [
    {
      "organization_invitation_id": "uuid",
      "organization": {
        "organization_id": "uuid",
        "organization_name": "Example Organization",
      },
      "status": "pending",
      "created_at": "2024-01-08T10:30:00Z",
      "expires_at": "2024-01-15T10:30:00Z",
      "inviter": {
        "user_id": "uuid"
        "first_name": "John",
        "last_name": "Doe",
        "email": "admin@org.com"
      },
      "roles": [
        {
          "role_id": "uuid",
          "role_name": "Organization Member",
          "privileges": ["view_org_membership", "start_application"]
        }
      ]
    }
  ]
}
```

### Acceptance criteria

- Implement `POST /v1/users/{user_id}/invitations/list` endpoint
- Returns appropriate error codes (400, 401, 403, 404, 422)
- Returns a list of user's invitations and related data on 200 response
- unit tests created

joshtonava:
## Summary

Fixes #6621 

## Changes proposed
- Added `POST /v1/users/{user_id}/invitations/list` to retrieve all organization invitations for a user

## Context for reviewers

- Implemented `get_user_invitations()` service to fetch invitations by matching user's login.gov email
- Endpoint requires JWT authentication and validates user can only access their own invitations
- Uses a single query with eager loading to fetch all related data (organization, inviter, roles)
- Returns invitations in all statuses (pending, accepted, rejected, expired)
- Includes complete invitation details: organization info, inviter details, assigned roles, and timestamps

## Validation steps

to run the tests 

```bash
make test args="-vv tests/src/api/users/test_user_invitations_list.py"
```

### DIFF
--- a/api/src/db/models/user_models.py
+++ b/api/src/db/models/user_models.py
@@ -65,7 +65,7 @@ class User(ApiSchemaTable, TimestampMixin):
     api_keys: Mapped[list["UserApiKey"]] = relationship(
         "UserApiKey", back_populates="user", uselist=True, cascade="all, delete-orphan"
     )
-    profile: Mapped["UserProfile"] = relationship(
+    profile: Mapped["UserProfile | None"] = relationship(
         "UserProfile", back_populates="user", uselist=False, cascade="all, delete-orphan"
     )
 
@@ -79,11 +79,13 @@ class User(ApiSchemaTable, TimestampMixin):
     def first_name(self) -> str | None:
         if self.profile is not None:
             return self.profile.first_name
+        return None
 
     @property
     def last_name(self) -> str | None:
         if self.profile is not None:
             return self.profile.last_name
+        return None
 
     @property
     def internal_roles(self) -> list["Role"]:


api/src/services/users/get_user_invitations.py
```diff
@@ -0,0 +1,138 @@
...
+        return []
+
+    # Transform to API response format
+    invitation_data: list[InvitationItemData] = []
+    for invitation in invitations:
+        # Get organization name - for now we'll use the organization_id as name
+        # since Organization model doesn't have a name field in the current schema
+        organization_data: InvitationOrganizationData = {
+            "organization_id": invitation.organization.organization_id,
+            "organization_name": f"Organization {invitation.organization.organization_id}",
```
chouinar: Organization has a name, but it's in the sam.gov entity record as the `legal_business_name`. While the sam.gov entity is nullable, it shouldn't ever be.

api/src/services/users/get_user_invitations.py
```diff
@@ -0,0 +1,138 @@
...
+            "last_name": (
+                invitation.inviter_user.profile.last_name
+                if invitation.inviter_user.profile
+                else None
+            ),
+            "email": (
+                invitation.inviter_user.linked_login_gov_external_user.email
+                if invitation.inviter_user.linked_login_gov_external_user
+                else None
+            ),
```
chouinar: The user table has a property that shortens this.
```suggestion
            "email": invitation.inviter_user.email,
```

api/src/services/users/get_user_invitations.py
```diff
@@ -0,0 +1,138 @@
...
+            "first_name": (
+                invitation.inviter_user.profile.first_name
+                if invitation.inviter_user.profile
+                else None
+            ),
+            "last_name": (
+                invitation.inviter_user.profile.last_name
+                if invitation.inviter_user.profile
+                else None
+            ),
```
chouinar: Since we're going to need these in a fair number of places, can we add a property to the user table that just encapsulates this so we can do `user.first_name` / `user.last_name`?

api/src/services/users/get_user_invitations.py
```diff
@@ -0,0 +1,138 @@
...
+from sqlalchemy.orm import selectinload
+
+from src.adapters import db
+from src.db.models.entity_models import LinkOrganizationInvitationToRole, OrganizationInvitation
+from src.db.models.user_models import LinkExternalUser, User
+
+
+class InvitationOrganizationData(TypedDict):
+    organization_id: UUID
+    organization_name: str
```
chouinar: Thoughts on using dataclasses (or Pydantic, same syntax for either) for this instead of dicts?

Makes the syntax slightly simpler:
```py
x =  InvitationOrganizationData(organization_id=a, organization_name=b)

# versus needing quotes/{}
x: InvitationOrganizationData = {"organization_id": a, "organization_name": b}
```

Marshmallow can convert either uneventfully back to JSON without any additional work.

api/src/services/users/get_user_invitations.py
```diff
@@ -0,0 +1,138 @@
...
+                else None
+            ),
+        }
+
+        # Get roles information
+        roles_data: list[InvitationRoleData] = [
+            {
+                "role_id": role.role_id,
+                "role_name": role.role_name,
+                "privileges": [priv.value for priv in role.privileges],
```
chouinar: What happens if you just do `"privileges": role.privileges`?

They're StrEnums so Marshmallow should easily make them into strings without any additional work since with a StrEnum this is true `MyEnum.thing.value == MyEnum.thing == str(MyEnum.thing)`

joshtonava: worth a shot

api/src/db/migrations/versions/2025_10_23_rename_responded_by_user_id_to_invitee_.py
chouinar, line 1: I assume this is the same as from your other PR? Will need to resync from main for it

api/src/db/models/user_models.py
```diff
@@ -75,6 +75,16 @@ def email(self) -> str | None:
...
+    @property
+    def first_name(self) -> str | None:
+        if self.profile is not None:
+            return self.profile.first_name
+
+    @property
+    def last_name(self) -> str | None:
+        if self.profile is not None:
+            return self.profile.last_name
+
```
chouinar, line 89: Very nitpicky, but could we be explicit here for the null case? 

```suggestion
    @property
    def first_name(self) -> str | None:
        if self.profile is not None:
            return self.profile.first_name
            
        return None

    @property
    def last_name(self) -> str | None:
        if self.profile is not None:
            return self.profile.last_name
            
       return None

```

api/src/db/migrations/versions/2025_10_23_rename_responded_by_user_id_to_invitee_.py
joshtonava: Yes this is the same I did it in all of them. Ill do a resync now

api/src/db/models/user_models.py
joshtonava: I can try, perhaps I was missing new line but It was giving me lint errors that the return statement was unreachable

joshtonava: Im getting this error

```
src/db/models/user_models.py: note: In member "first_name" of class "User":
src/db/models/user_models.py:83:9: error: Statement is unreachable  [unreachable]
            return None
            ^~~~~~~~~~~
src/db/models/user_models.py: note: In member "last_name" of class "User":
src/db/models/user_models.py:90:9: error: Statement is unreachable  [unreachable]
            return None
            ^~~~~~~~~~~
Found 2 errors in 1 file (checked 498 source files)
make: *** [lint-mypy] Error 1
```

joshtonava: I think the issue is that the profile is marked as always being defined but really it could be None.

I updated the type of profile to be "UserProfile | None" and this seemed to do the trick

chouinar (approved):

PR #6840, opened 2025-10-23
setup documentation updates 

ErinPattisonNava:
## Summary

Updating the setup documentation for readability, changing some env vars and Makefile commands to make things easier, and adding me to the CODEOWNERS file

## Changes proposed

* Updating the local env file for the API to have the frontend url by default.
* New make command to load all data sources.
* Updates to various readme/documentation files for updated commands and instructions.
* Fixing the typo for the auth key associated with two_org_user to match the user handle.
* Adding me to the CODEOWNERS and MAINTAINERS files.  

## Context for reviewers

N/A

## Validation steps

- [ ] Ensure scripts run locally successfully.
- [ ] Read documentation in preview.

### DIFF
--- a/documentation/frontend/development.md
+++ b/documentation/frontend/development.md
@@ -18,12 +18,20 @@ Optionally, disable [telemetry data collection](https://nextjs.org/telemetry)
 npx next telemetry disable
 ```
 
+### Configuration
+
+Create a local enviornment file in the frontend directory to hold your frontend application overrides.  To allows you to make specializations to your local setup outside of GitHub.
+
+`touch .env.local`
+
+For more information about environments, take a look at [environments.md](./environments.md).
+
 ### Authentication
 
 Running authentication locally requires running the API and sharing the correct JWT keys.
 
 1. Ensure you've completed the [API setup](../api/development.md), including creating the `override.env` file
-2. Copy the `API_JWT_PUBLIC_KEY` value from `/api/override.env` file to `/frontend/.env.local` file which creates the necessary keys
+2. Copy the `API_JWT_PUBLIC_KEY` value from `/api/override.env` file to your `/frontend/.env.local` file which creates the necessary keys
 3. Restart the API (if necessary reseed the database, then `make start`) and frontend (`npm run dev`) for development
 
 ### 🏛️ "Built" version


api/tests/lib/seed_orgs_and_users.py
```diff
@@ -138,7 +138,7 @@ def _build_organizations_and_users(
     ###############################
     UserBuilder(
         uuid.UUID("0f4ae584-c310-472d-9d6c-57201b5f84cc"), db_session, "user with two orgs"
-    ).with_oauth_login("two_org_user").with_api_key("two_orgs_user_key").with_organization(
+    ).with_oauth_login("two_org_user").with_api_key("two_org_user_key").with_organization(
```
ErinPattisonNava, line 141: FYI @chouinar

api/local.env
```diff
@@ -50,9 +50,9 @@ LOGIN_GOV_AUTH_ENDPOINT=http://localhost:5001/issuer1/authorize
 LOGIN_GOV_TOKEN_ENDPOINT=http://host.docker.internal:5001/issuer1/token
 LOGIN_GOV_ENDPOINT=http://host.docker.internal:5001/issuer1
 
-LOGIN_FINAL_DESTINATION=http://localhost:8080/v1/users/login/result
+LOGIN_FINAL_DESTINATION=http://localhost:3000/api/auth/callback
 
-FRONTEND_BASE_URL=http://localhost:8080
+FRONTEND_BASE_URL=http://localhost:3000
```
ErinPattisonNava: @chouinar @doug-s-nava I decided to update the local env file instead of the override file, ~~and also updated the frontend base url to keep them in sync -- let me know if you'd prefer just to update LOGIN_FINAL_DESTINATION/~~ only change the override file.

nvm about the FRONTEND_BASE_URL, that  broke a number of unittests so I reverted.

chouinar: FYI `FRONTEND_BASE_URL` is used in a backend process to generate email notifications - wouldn't meaningfully affect anything with local development to change it.

api/Makefile
```diff
@@ -319,6 +319,11 @@ populate-search-agencies: ## Load agencies from the DB into the search index, ru
 run-generate-notifications: ## Send email notifications
 	$(FLASK_CMD) generate-notifications $(args)
 
+db-seed-local-all-data: ## Load all the data, especially useful for inital setup
+	$(PY_RUN_CMD) db-seed-local --cover_all_agencies=true
+	$(FLASK_CMD) load-search-data load-opportunity-data $(args)
+	$(FLASK_CMD) load-search-data load-agency-data $(args)
```
chouinar: Could we reuse the existing commands? Note that in `db-seed-local` the `cover_all_agencies` flag is defaulted to true, so excluding that won't change how it runs.

Also wondering if there's a shorter way to write the command, something like `setup-api-data`?  The 2nd and 3rd command here are search index loads

```suggestion
db-seed-local-all-data: db-seed-local populate-search-opportunities populate-search-agencies ## Load all the data, especially useful for initial setup

```

documentation/api/development.md
```diff
@@ -27,7 +31,7 @@ This stands up the following services:
 
 ### Seed data
 
-Run `make db-seed-local && make populate-search-opportunities` to create local data in the database and make it available in the API.
+Run `make db-seed-local-all-data` to create local data in the database and make it available in the API. This basically creates everything; if you want to be selective about what data you're seeding, see the Makefile for ways you can populate selective data.
```
chouinar:
```suggestion
Run `make db-seed-local-all-data` to create local data in the database, search index and make it available in the API. This basically creates everything; if you want to be selective about what data you're seeding, see the Makefile for ways you can populate selective data.
```

documentation/api/development.md
```diff
@@ -37,6 +41,14 @@ This API uses a very simple [ApiKey authentication approach](https://apiflask.co
 
 Run `make setup-env-override-file` to create the `override.env` file which will include the necessary JWT keys for running user authentication within the app.
 
+### Accessing the API through swagger docs
+
+To see the current API definiton, you can go to the local [swagger docs](http://localhost:8080/docs).
+
+Each endpoint specifies which auth token it needs behind the lock icon. The `ApiJwtAuth` token is printed as part of user creation in `make db-seed-local-all-data`, search for `create_jwt_for_user` to see the `auth.token_id` value. For the `ApiUserKeyAuth` token, search for `X-API-Key` for the different values for each role (hint: they typically are the user name + \_key, for instance `one_org_user` the key value is `one_org_user_key`). You can set these tokens per end point, or set them globally at the top on the lock icon labeled Authorize.
```
chouinar: I have a note to myself to update the auth docs, I want to explain the auth from both an internal way and how it gets used externally (eg. the JWT auth happens via login, the key auth happens when a user makes a key).

documentation/api/development.md
```diff
@@ -14,7 +14,11 @@ This section covers development using Docker. There are a number of Docker comma
 
 ### Setup
 
-Run `make init && make run-logs` to start the local containers. The application will be available at `http://localhost:8080` and API documentation at `http://localhost:8080/docs`.
+These scripts depend on having `postgresql` installed in your enviornment as a pre-requisite.
```
chouinar: A helpful link to the website will help users know what to do - and looks to be pretty thorough in handling all the various OS differences in installing (brew, apt, whatever WIndows is doing)

```suggestion
These scripts depend on having `postgresql` installed in your environment as a pre-requisite. Install instructions for your particular operating system can be found on the [Postgres website](https://www.postgresql.org/download/)
```

api/local.env
ErinPattisonNava: @chouinar Would you prefer I keep the values in sync and change the unittests, or leave it as is?

documentation/api/development.md
ErinPattisonNava: Sounds good -- can I leave this here and you can update it later?

api/Makefile
ErinPattisonNava: Fixed in https://github.com/HHS/simpler-grants-gov/pull/6840/commits/0cf64ce6071e88e3d58e4ad0acf0127544c7e97f

documentation/api/development.md
ErinPattisonNava: fixed in https://github.com/HHS/simpler-grants-gov/pull/6840/commits/0cf64ce6071e88e3d58e4ad0acf0127544c7e97f

frontend/README.md
```diff
@@ -30,34 +30,6 @@
...
-   ```bash
-   npm ci
-   ```
-1. Optionally, disable [telemetry data collection](https://nextjs.org/telemetry)
-   ```bash
-   npx next telemetry disable
-   ```
-1. Create local environment file
-   ```bash
-   cp .env.development .env.local
```
doug-s-nava, line 46: what's here probably isn't the right place or the right way to document this, but I think we should cover the importance of creating a local env file. It is not strictly necessary, but I feel like we should act like it is in order to limit the risk of someone leaking secrets into github via changes to the dev env file. Thoughts?

api/local.env
chouinar: Leave it as is

frontend/README.md
ErinPattisonNava: Most folks are only using the .env.local file for authentication with the API, so I moved the reference into that section: https://github.com/HHS/simpler-grants-gov/blob/442f7decc5ed0262f294931570137119d976ec12/documentation/frontend/development.md?plain=1#L26

But if you'd like a more general reference that a .env.local file exists and can be used, I can add it to the other frontend documentation file outside of the authentication section too.  I think copying the whole file over makes it difficult to debug issues of setup because it isn't clear what has changed/is specified for local.

doug-s-nava: Agree it doesn't make sense to copy the whole file. In general the .env.local should be used any time you need to update environment variables locally through a file. Looking more closely, I feel like this section is out of date and should probably direct folks to create a .env.local file and use that for any local env var changes they need to make for their environment https://github.com/HHS/simpler-grants-gov/blob/442f7decc5ed0262f294931570137119d976ec12/documentation/frontend/environments.md?plain=1#L22

ErinPattisonNava: Changes made in  2cfe6ecc7d2555fa6297ade39c7368b4aa5e00e0

PR #6841, opened 2025-10-24
[ISSUE #6620] List Organization Invitations Endpoint

### LINKED ISSUE #6620: List Organization Invitations Endpoint
### Summary

Implement the endpoint that allows organization admins to view all invitations for their organization with filtering and pagination capabilities, following the existing POST pattern for list operations.

#### Endpoint Specification
- **Method**: POST
- **Path**: `/v1/organizations/{organization_id}/invitations/list`
- **Authentication**: Required (JWT)
- **Authorization**: `VIEW_ORG_MEMBERSHIP` privilege for the organization

#### Request Schema
```json
{
  "filters": {
    "status": {
      "one_of": ["pending", "accepted", "rejected", "expired"]
    },
    "email": "user@example.com",
    "created_at": {
      "start_date": "2024-01-01",
      "end_date": "2024-01-31"
    }
  },
  "pagination": {
    "sort_order": [
      {
        "order_by": "created_at",
      }
    ],
    "page_offset": 1,
    "page_size": 25
  }
}
```

#### Response Schema
```json
{
  "message": "Success",
  "data": [
    {
      "organization_invitation_id": "uuid",
      "invitee_email": "user@example.com",
      "status": "pending",
      "created_at": "2024-01-08T10:30:00Z",
      "expires_at": "2024-01-15T10:30:00Z",
      "accepted_at": null,
      "rejected_at": null,
      "inviter": {
        "user_id": "uuid",
        "email": "admin@org.com",
        "first_name": "John",
        "last_name": "Doe"
      },
      "invitee": null,
      "roles": [
        {
          "role_id": "uuid",
          "role_name": "Organization Member"
        }
      ]
    }
  ],
  "pagination_info": {
    "page_offset": 1,
    "page_size": 25,
    "total_records": 45,
    "total_pages": 2,
    "sort_order": [
      {
        "order_by": "created_at",
      }
    ]
  }
}
```

### Acceptance criteria

- Implement `POST /v1/organizations/{organization_id}/invitations/list` endpoint
- Returns appropriate error codes (400, 401, 403, 404, 422)
- Returns a list of invitations and related data on 200 response
- Filtering by status (in order to filter out "accepted" invitations)
- Pagination
- unit tests created

joshtonava:
## Summary

Fixes #6620 

## Changes proposed

- Added POST /v1/organizations/{organization_id}/invitations/list to retrieve and filter organization invitations

## Context for reviewers

- Implemented list_organization_invitations_and_verify_access() service to fetch invitations with status-based filtering
- Endpoint requires JWT authentication and VIEW_ORG_MEMBERSHIP privilege
- Uses optimized queries with selectinload to eager load relationships (inviter, invitee, roles) 
- Returns invitations in all statuses (pending, accepted, rejected, expired) with optional filtering
- Includes complete invitation details: inviter/invitee info, assigned roles, and timestamps

## Validation steps

for running the tests

```bash
make test args="-vv tests/src/api/organizations_v1/test_list_organization_invitations.py"
```

### DIFF
--- a/api/src/services/organizations_v1/list_organization_invitations.py
+++ b/api/src/services/organizations_v1/list_organization_invitations.py
@@ -97,7 +97,7 @@ def list_organization_invitations_and_verify_access(
     user: User,
     organization_id: uuid.UUID,
     filters: dict | None = None,
-) -> list[OrganizationInvitationData]:
+) -> Sequence[OrganizationInvitation]:
     """
     List organization invitations with access control and filtering.
 
--- a/api/src/services/users/list_user_invitations.py
+++ b/api/src/services/users/list_user_invitations.py
@@ -10,9 +10,6 @@ from src.db.models.entity_models import (
     OrganizationInvitation,
 )
 from src.db.models.user_models import LinkExternalUser, User
-from src.services.organizations_v1.organization_invitation_response_utils import (
-    OrganizationInvitationData,
-)
 
 
 def _fetch_user_invitations_by_user_id(
@@ -44,9 +41,7 @@ def _fetch_user_invitations_by_user_id(
     return list(invitations)
 
 
-def list_user_invitations(
-    db_session: db.Session, user_id: UUID
-) -> list[OrganizationInvitationData]:
+def list_user_invitations(db_session: db.Session, user_id: UUID) -> list[OrganizationInvitation]:
     """Get all invitations for a user by matching their login.gov email.
 
     Args:


api/src/services/organizations_v1/list_organization_invitations.py
```diff
@@ -0,0 +1,256 @@
...
+    created_at: Any  # datetime
+    expires_at: Any  # datetime
+    accepted_at: Optional[Any]  # datetime
+    rejected_at: Optional[Any]  # datetime
+    inviter: InviterData
+    invitee: Optional[InviteeData]
+    roles: List[RoleData]
+
+
+def parse_request_data(data: Dict[str, Any]) -> OrganizationInvitationListRequest:
```
joshtonava: Let me know if there is a different / better pattern to do this with Marshmallow / if we have already solved this in some way. I was trying to have type safety while accessing the properties in the filters

api/src/api/organizations_v1/organization_routes.py
```diff
@@ -156,3 +161,38 @@ def organization_update_user_roles(
...
+    """List organization invitations with filtering"""
+    add_extra_data_to_current_request_logs({"organization_id": organization_id})
+    logger.info("POST /v1/organizations/:organization_id/invitations/list")
+
+    # Get authenticated user
+    user_token_session: UserTokenSession = api_jwt_auth.get_user_token_session()
+
+    # Parse the validated JSON data into a proper request object
+    schema = OrganizationInvitationListRequestSchema()
+    request_data = schema.load(json_data)
```
chouinar: This is already being done automatically by having the schema as an input. Any type conversions we have in the schemas (enums, decimal, date, whatever else) will already be the proper python types. I think this actually would fail if we did any conversions as it'd expect a different type.

api/src/services/organizations_v1/list_organization_invitations.py
```diff
@@ -0,0 +1,238 @@
+import uuid
+from dataclasses import dataclass
+from datetime import datetime
+from typing import Any, Dict, List, Optional, Sequence
```
chouinar: We don't need Dict/List/Optional anymore from the typing library. Can just use dict/list/ `| None`

```suggestion
from typing import Any, Sequence
```

api/src/services/organizations_v1/list_organization_invitations.py
```diff
@@ -0,0 +1,238 @@
...
+                OrganizationInvitationStatus(status) for status in status_data["one_of"]
+            ]
+            status_filter = StatusFilter(one_of=status_values)
+
+        filters = InvitationFilters(status=status_filter)
+
+    return OrganizationInvitationListRequest(filters=filters)
+
+
+def transform_invitation_to_data(invitation: OrganizationInvitation) -> OrganizationInvitationData:
```
chouinar: The GET endpoint has a similar set of logic, can that be shared? Would the schemas be any different besides this one being a list of invitations versus that one being just one?

api/src/services/organizations_v1/list_organization_invitations.py
```diff
@@ -0,0 +1,238 @@
...
+    # First get the organization
+    stmt = select(Organization).where(Organization.organization_id == organization_id)
+    organization = db_session.execute(stmt).scalar_one_or_none()
+
+    if organization is None:
+        raise_flask_error(404, message=f"Could not find Organization with ID {organization_id}")
+
+    # Check if user has required privilege for this organization
+    if not can_access(user, {Privilege.VIEW_ORG_MEMBERSHIP}, organization):
+        raise_flask_error(403, "Forbidden")
```
chouinar, line 32: Do we want this to be more restrictive, should someone who doesn't have the ability to invite be able to see who has a pending invite?

api/src/services/organizations_v1/list_organization_invitations.py
```diff
@@ -0,0 +1,238 @@
...
+    # Execute query to get all invitations (we'll filter in Python since status is computed)
+    all_invitations = db_session.execute(stmt).scalars().all()
+
+    # Apply status filtering if provided
+    filtered_invitations = all_invitations
+    if filters and filters.status and filters.status.one_of:
+        status_filters = filters.status.one_of
+        filtered_invitations = [
+            invitation for invitation in all_invitations if invitation.status in status_filters
+        ]
```
chouinar: I suppose as long as we don't care about pagination for this endpoint, this would be fine to not be something in the query itself.

But, if we ever want pagination for this endpoint, we'd have to implement the status filter into the DB query to make it work.

api/src/services/organizations_v1/list_organization_invitations.py
joshtonava: Are you referring to the list endpoint for users to get organization invites?

Or the endpoint for creating an organization invite?

api/src/services/organizations_v1/list_organization_invitations.py
joshtonava: That is a good question. Probably yes you should have to have the same permissions needed to create an invitation

api/src/services/organizations_v1/list_organization_invitations.py
joshtonava: Do you think its worth to move the logic into the DB right now ? At some point we may want pagination.

I agree since we arent doing pagination or ordering at this current point in time I thought this would be a little easier.

api/src/services/organizations_v1/list_organization_invitations.py
joshtonava: I made this be manage org membership

api/src/services/organizations_v1/list_organization_invitations.py
joshtonava: I can try to set these up to be shared

joshtonava: These are now in a shared module

api/src/services/organizations_v1/organization_invitation_response_utils.py
```diff
@@ -0,0 +1,130 @@
...
+    # Fields used by organization invitation endpoints
+    invitee_email: str | None = None
+    accepted_at: datetime | None = None
+    rejected_at: datetime | None = None
+    invitee: InviteeData | None = None
+    # Fields used by user invitation endpoints
+    organization: OrganizationData | None = None
+
+
+def transform_invitation_to_response(
```
joshtonava: I think I can get rid of this all together if I change the response and alter the query

api/src/services/users/list_user_invitations.py
```diff
@@ -0,0 +1,62 @@
+from uuid import UUID
```
joshtonava, line 1: renamed this to be more consistent

Issue #6771, opened 2025-10-17
[BUG] PDF generation not showing radio button selection


chris-kuryak:
### Summary

On form SF-424, there is a radio button.

After submitting the application, my selection to the radio button field does not appear in the generated PDF.
Screenshot from SF-424:

<img width="583" height="284" alt="Image" src="https://github.com/user-attachments/assets/97f873d4-c1e9-4380-b623-2178f84630c1" />

Screenshot from generated PDF:

<img width="579" height="265" alt="Image" src="https://github.com/user-attachments/assets/3f95c696-73e8-404b-8f5b-2281e4b276fa" />

### Acceptance criteria

- [ ] Radio button selection appears in generated PDF after successful submission
- [ ] UAT passed

chris-kuryak:
UAT passed

PR #6844, opened 2025-10-24
fix: add checks for value type, radio specific

myduong-navapbc:
## Summary

Fixes / Work for #6771 

## Changes proposed

I've added some checks in the PrintWidget
```
    if (inputValue == null) return "";
    if (typeof inputValue === "string") return inputValue;
    if (typeof inputValue === "boolean") return String(inputValue);
    if (typeof inputValue === "number" || typeof inputValue === "bigint") return String(inputValue);
    if (Array.isArray(inputValue)) return inputValue.filter(v => v != null).join(", ");

    console.warn("Unknown value type ", typeof inputValue)
    return JSON.stringify(inputValue); 
```

Screenshot 2025-10-24 at 11 41 06 AM

## Validation steps

3. navigate to `search`
4. search for `pilot`
6. sign in
7. start new application

### DIFF
--- a/frontend/src/components/application/InformationCard.tsx
+++ b/frontend/src/components/application/InformationCard.tsx
@@ -6,7 +6,6 @@ import {
   Status,
 } from "src/types/applicationResponseTypes";
 import { Competition } from "src/types/competitionsResponseTypes";
-import { getConfiguredDayJs } from "src/utils/dateUtil";
 
 import { useTranslations } from "next-intl";
 import { Button, Grid, GridContainer, Link } from "@trussworks/react-uswds";
@@ -84,13 +83,7 @@ export const InformationCard = ({
 }) => {
   const t = useTranslations("Application.information");
   const hasOrganization = Boolean(applicationDetails.organization);
-
-  const isCompetitionClosed = () => {
-    const isPastCloseDate = getConfiguredDayJs()().isAfter(
-      getConfiguredDayJs()(applicationDetails.competition.closing_date),
-    );
-    return !applicationDetails.competition.is_open || isPastCloseDate;
-  };
+  const { is_open } = applicationDetails.competition;
 
   const ApplicationInstructionsDownload = () => {
     return (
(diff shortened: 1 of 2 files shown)


kmk142789 (approved):

frontend/src/components/applyForm/widgets/PrintWidget.tsx
```diff
@@ -17,6 +17,16 @@ function PrintWidget<
   inputClassName,
 }: UswdsWidgetProps<T, S, F>) {
   const { title } = schema as S;
+  const toDisplay = (inputValue: unknown): string => {
```
doug-s-nava, line 20: can you add a test around this?

frontend/src/components/applyForm/widgets/PrintWidget.tsx
```diff
@@ -17,6 +17,16 @@ function PrintWidget<
...
+    if (inputValue == null) return "";
+    if (typeof inputValue === "string") return inputValue;
+    if (typeof inputValue === "boolean") return String(inputValue);
```
doug-s-nava: Want to make sure that at least for now we good with displaying "true" or "false" here, rather than "yes" or "no" or otherwise making use of labels? That'd be more complex and we should likely move forward with this change regardless but if it's necessary we should ticket it now cc @chris-kuryak

doug-s-nava (commented):
not directly related to this ticket, as this seems to also be an issue on main, but I do see a bug with file upload components showing "previously uploaded file" in the print form when the field is empty
Screenshot 2025-10-27 at 10 29 08 AM

frontend/src/components/applyForm/widgets/PrintWidget.tsx
myduong-navapbc: @chris-kuryak @doug-s-nava, I believe the preferred display is Yes/No. i've made changes to reflect this and updated the tests to capture it too

frontend/src/components/applyForm/widgets/PrintWidget.tsx
```diff
@@ -17,6 +17,18 @@ function PrintWidget<
   inputClassName,
 }: UswdsWidgetProps<T, S, F>) {
   const { title } = schema as S;
+  const toDisplay = (inputValue: unknown): string => {
+    if (inputValue == null) return "";
```
doug-s-nava: I had to refresh my memory, but I think this means something like `inputValue === null || inputValue === undefined`. Personally I think that writing this the long way is much easier to understand (and generally think we should make it a documented code style point to not use `==`), but if you disagree we can save that conversation for a later date

frontend/tests/components/applyForm/widgets/PrintWidget.test.tsx
```diff
@@ -65,22 +73,18 @@ describe("PrintWidget", () => {
   });
 
   it("displays empty string when value is null or undefined", () => {
```
doug-s-nava: this doesn't test `undefined`, can we add that case here?

doug-s-nava (commented):
everything is working well, two minor requests on code

doug-s-nava (approved):
looks like there's a prettier error somewhere, but otherwise lgtm!

✅ ✅ ✅ ✅

PR #6847, opened 2025-10-24
Fix Frontend E2E test failures 

prasnava:
## Summary

Fixed the frontend e2e test failures. 

## Changes Proposed.

Disabled the opensearch dashboard and commented out in the docker-compose file. The dashboard was straining open search from serving requests. 

Updated the api docker image to use glibc instead of musl.  

## Validation steps

Frontend E2E tests should be passing, api and step functions should continue to run.

### DIFF
--- a/.github/workflows/ci-frontend-e2e.yml
+++ b/.github/workflows/ci-frontend-e2e.yml
@@ -55,10 +55,6 @@ jobs:
           echo "/tmp/cores/core.%e.%p.%t" | sudo tee /proc/sys/kernel/core_pattern
           # Run init first to generate JWT keys in override.env
           make init
-          # Then append CI-specific configuration
-          # Set replica count to 0 for single-node cluster in CI
-          echo "LOAD_AGENCY_SEARCH_REPLICA_COUNT=0" >> override.env
-          echo "LOAD_OPP_SEARCH_REPLICA_COUNT=0" >> override.env
           # Now run the seeding and populate commands
           make db-seed-local-with-agencies populate-search-opportunities populate-search-agencies start
           cd ../frontend


kmk142789 (approved):

prasnava:
Deployed to training [here](https://github.com/HHS/simpler-grants-gov/actions/runs/18845355808/job/53767833527). 

Will wait for the step functions to successfully run before merging.

.github/workflows/ci-frontend-e2e.yml
```diff
@@ -19,15 +19,14 @@ env:
...
-  cancel-in-progress: false
+  cancel-in-progress: true
 
 jobs:
   e2e-tests:
     name: Run E2E Tests
     runs-on: ubuntu-22.04
 
     strategy:
-      fail-fast: false
```
doug-s-nava: what's the reasoning behind turning this off? I think we still want to fail fast here

prasnava: That was for my testing, will turn that back on.

.github/workflows/ci-frontend-e2e.yml
```diff
@@ -46,9 +49,18 @@ jobs:
...
+          # Enable core dumps for debugging segfaults
+          ulimit -c unlimited
+          mkdir -p /tmp/cores
+          echo "/tmp/cores/core.%e.%p.%t" | sudo tee /proc/sys/kernel/core_pattern
+          # Run init first to generate JWT keys in override.env
+          make init
+          # Then append CI-specific configuration
+          # Set replica count to 0 for single-node cluster in CI
+          echo "LOAD_AGENCY_SEARCH_REPLICA_COUNT=0" >> override.env
+          echo "LOAD_OPP_SEARCH_REPLICA_COUNT=0" >> override.env
```
chouinar: It seems running this doesn't have an issue, but I'm not sure what "0 shards" means, can't find any docs about what it does when it has 0 as 1 is what I thought the minimum was. At best I can maybe find details that say it'll just automatically add a shard when you index data, but not certain if that's right.

chouinar: Actually, double-checking it locally, this fails as the value has to be 1 or higher.

```
opensearchpy.exceptions.RequestError: RequestError(400, 'illegal_argument_exception', 'Failed to parse value [0] for setting [index.number_of_shards] must be >= 1')
```

It doesn't fail here because you ran `make init` above this so adding the value to the `override.env` won't matter as the container already had the previously created `override.env` file attached.

prasnava: Was debugging changing the opensearch config settings to get it working. will remove it.

Issue #6649, opened 2025-10-07
User respond to Invitation Endpoint


joshtonava:
### Summary

Implement the endpoint that allows users to accept or reject organization invitations, creating the necessary organization membership on acceptance.

#### Endpoint Specification
- **Method**: PUT
- **Path**: `/v1/users/{user_id}/invitations/{invitation_id}`
- **Authentication**: Required (JWT)
- **Authorization**: User's email must match invitation email
- **Invitation Validation**: 
   - Invitation exists and is pending
   - Not expired
   - User's email matches invitation email
   - Invitation belongs to the specified user
- **Accept Logic** (status = "accepted"):
   - Create `OrganizationUser` record
   - Create `OrganizationUserRole` records for each invited role
   - Set `accepted_at` timestamp
   - Set `responded_by_user_id`
- **Reject Logic** (status = "rejected"):
   - Set `rejected_at` timestamp
   - Set `responded_by_user_id`
   - No organization access created

#### Request Schema
```json
{
  "status": "accepted" // or "rejected"
}
```

#### Response Schema
```json
{
  "message": "Success",
  "data": {
    "organization_invitation_id": "uuid",
    "status": "accepted",
    "responded_at": "2024-01-10T14:30:00Z",
    "organization": {
        "organization_id": "uuid",
        "organization_name": "Example Organization",
      },
    "roles_granted": [
      {
        "role_id": "uuid",
        "role_name": "Organization Member"
      }
    ]
  }
}
```

### Acceptance criteria

- Implement `PUT /v1/users/{user_id}/invitations/{invitation_id}`
- Returns appropriate error codes (400, 401, 403, 404, 422)
- Invitation updated and user added to organization with correct roles on successful 200 response
- unit tests created

PR #6848, opened 2025-10-24
[ISSUE #6649] user invitation response endpoint

babebe:
## Summary

Fixes / Work for #6649  

## Changes proposed

Added endpoint `POST /v1/users/:user_id/invitations/{invitation_id}`
Added Request and Response Schema
Added Tests
## Context for reviewers

Allows a user to respond to an organization invitation by accepting or rejecting it. 
The endpoint validates:

- that the invitation exists, is still pending, and is not expired.
- ensures that the authenticated user matches the invitee’s ID and email. 
- ensures that the user's ID matches the invitee’s user ID.

On acceptance, the user is added to the organization with any linked roles.

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -38,6 +38,7 @@ COPY newrelic.ini /api/newrelic.ini
 RUN PY_MM=$(cat /tmp/py_version.txt) && \
     /usr/local/bin/python"${PY_MM}" -m poetry build --format wheel && \
     /api/.venv/bin/python -m pip install --no-cache-dir dist/simpler_grants_gov_api-0.1.0-py3-none-any.whl && \
+    /api/.venv/bin/python -m pip install --no-cache-dir "virtualenv>=20.35.3" && \
     SITE_BASE="/api/.venv/lib/python${PY_MM}/site-packages" && \
     mkdir -p "${SITE_BASE}/src" && \
     cp "/api/newrelic.ini" "${SITE_BASE}/src/newrelic.ini" && \
@@ -113,11 +114,7 @@ RUN PY_MM=$(cat /tmp/py_version.txt) && \
       /api/.venv/bin/pip* \
       /api/.venv/bin/easy_install* \
       /root/.local/share/virtualenv \
-      "/home/${RUN_USER}/.local/share/virtualenv" \
-      /usr/local/lib/python"${PY_MM}"/site-packages/virtualenv/seed/wheels/embed/pip-*.whl \
-      /usr/local/lib/python"${PY_MM}"/site-packages/virtualenv/seed/wheels/embed/setuptools-*.whl \
-      /usr/local/lib/python"${PY_MM}"/site-packages/virtualenv/seed/wheels/embed/wheel-*.whl \
-      /usr/local/lib/python"${PY_MM}"/site-packages/virtualenv || true
+      "/home/${RUN_USER}/.local/share/virtualenv" || true
 
 ENV PATH="/api/.venv/bin:/usr/local/bin:/usr/bin:/usr/sbin:/bin" \
     HOST=0.0.0.0 \


api/src/services/users/org_invitation_response.py
```diff
@@ -0,0 +1,83 @@
...
+    # Validate status status
+    if invitation.status != OrganizationInvitationStatus.PENDING or invitation.is_expired:
+        raise_flask_error(
+            400, f"Invitation cannot be responded to; current status is {invitation.status}"
+        )
+
+    status = json_data.get("status")
+    now = utcnow()
+    if status == OrganizationInvitationStatus.ACCEPTED:
+        # Create organization and assign role
```
jakobpederson: It looks like this was meant to be "organization user" instead of organization

api/src/api/users/user_routes.py
```diff
@@ -709,3 +712,35 @@ def user_get_invitations(
     )
 
     return response.ApiResponse(message="Success", data=invitations)
+
+
+@user_blueprint.put("/<uuid:user_id>/invitations/<uuid:invitation_id>")
```
chouinar: Hmm, `PUT` feels off here, we're not putting an invitation into the DB with this, but responding to the invitation and putting a user into an organization. Maybe POST instead?

Since this endpoint only handles organizations, and we'll eventually need an endpoint for agencies and applications as well, could we maybe do: `POST /v1/users/:user_id/invitations/:invitation_id/organizations`? Might need to think a little on that pattern exactly

cc @joshtonava

api/src/api/users/user_schemas.py
```diff
@@ -680,3 +686,72 @@ class UserInvitationListResponseSchema(AbstractResponseSchema):
...
+    responded_at = fields.DateTime(metadata={"description": "Time User responded to invitation"})
+    organization = fields.Nested(
+        OrganizationSchema, metadata={"description": "Organization information"}
+    )
+    roles_granted = fields.List(
+        fields.Nested(RoleSchema), metadata={"description": "Roles granted"}
+    )
+
+    @pre_dump
+    def invitation_response(self, invitation: OrganizationInvitation, **kwargs: Any) -> dict:
```
chouinar: Do we need this? I'd really push for not doing _any_ special conversion of responses if it isn't needed as it's more complexity to change and maintain. The only parts I see that don't match 1:1 with the Organization Invite DB model are:
* `responded_at` - Does this need to be two different fields consolidated? If so, can add a property to the class to handle that.
* `organization_name` - We should really just add a property to the org table for this.

I'd always prefer properties in the DB models over having to build a complex object.

api/src/services/users/org_invitation_response.py
```diff
@@ -0,0 +1,83 @@
...
+        .where(OrganizationInvitation.organization_invitation_id == invitation_id)
+    ).scalar_one_or_none()
+
+    # Validate invitation
+    if not invitation:
+        raise_flask_error(404, "Invitation not found")
+
+    # Verify response user is the same
+    if user.user_id != invitation.invitee_user_id:
+        raise_flask_error(403, "Forbidden, invitation user does not match user's ID on record")
```
chouinar: This will fail 100% of the time if we include it.

We don't know who the user being invited is until this moment because it's based purely on email. An email may be invited before we even have that user in the system. 

One of the last lines in this function is to set this value, so this reads `if user.user_id != None`

api/src/services/users/org_invitation_response.py
```diff
@@ -0,0 +1,83 @@
...
+
+    # Verify responder email
+    if user.email != invitation.invitee_email:
+        raise_flask_error(403, "Forbidden, invitation email does not match user's email on record")
+
+    # Validate status status
+    if invitation.status != OrganizationInvitationStatus.PENDING or invitation.is_expired:
+        raise_flask_error(
+            400, f"Invitation cannot be responded to; current status is {invitation.status}"
+        )
```
chouinar: Should this be a 422? 400 is usually for a malformed request.

api/tests/src/api/users/test_org_invitation_response.py
```diff
@@ -0,0 +1,198 @@
...
+    return user
+
+
+def test_org_invitation_response_accepted(
+    client, db_session, user, user_auth_token, enable_factory_create
+):
+    """Test that an invitation can be successfully ACCEPTED by the correct user."""
+
+    # Create an invitation
+    inv = OrganizationInvitationFactory.create(invitee_user=user, invitee_email=user.email)
```
chouinar: We should not add the invitee user to this test data we create, that value should be null before they accept the request. If anything, if invitee_user is set, we should error.

api/src/api/users/user_routes.py
joshtonava: My first intuition was since we are updating the invitation record we should do a PATCH / PUT but we don't use PATCH in this project. 

I suppose this isnt a true CRUD operation and we are performing an action as the perspective of the user so I could see why POST makes more sense in this situation.

Also I agree with your sentiment on needing the invitation endpoint for users to work for organizations and applications. I like `/v1/users/:user_id/invitations/:invitation_id/organizations`

api/src/api/users/user_schemas.py
joshtonava: Check out what I did for organization_name in user_schemas.py. Not saying this is also perfect but you can use the attribute field to map down to a specific field:

```python
class UserInvitationOrganizationSchema(Schema):
    """Schema for organization information in invitation responses"""

    organization_id = fields.UUID(
        metadata={
            "description": "Organization unique identifier",
            "example": "123e4567-e89b-12d3-a456-426614174000",
        }
    )
    organization_name = fields.String(
        allow_none=True,
        attribute="sam_gov_entity.legal_business_name",
        metadata={
            "description": "Organization name",
            "example": "Example Organization",
        },
    )
    ```

File: api/src/services/users/org_invitation_response.py (hunk 4 above)
Comment (joshtonava) [reply api/src/services/users/org_invitation_response.py]: This is correct. The user_id is set as a result of this API call succeeding

File: api/src/api/users/user_routes.py (hunk 2 above)
Comment (babebe) [reply api/src/api/users/user_routes.py]: It does make sense to have this be a POST endpoint.

File: api/src/api/users/user_schemas.py (hunk 3 above)
Comment (babebe) [reply api/src/api/users/user_schemas.py]: Created property for both.

File: api/src/services/users/org_invitation_response.py (hunk 4 above)
Comment (babebe) [reply api/src/services/users/org_invitation_response.py]: Of course, missed that! Removing this step @joshtonava

File: api/src/services/users/org_invitation_response.py (hunk 5 above)
Comment (babebe) [reply api/src/services/users/org_invitation_response.py]: It should be 422! Updated

File: api/src/services/users/org_invitation_response.py (hunk 4 above)
Comment (babebe) [reply api/src/services/users/org_invitation_response.py]: @chouinar  If the user does not exist when invited, during what step do we add that user to the system

File: api/tests/src/api/users/test_org_invitation_response.py (hunk 6 above)
Comment (babebe) [reply api/tests/src/api/users/test_org_invitation_response.py]: Adjusted all tests

Comment (babebe) [reply api/tests/src/api/users/test_org_invitation_response.py]: @chouinar @joshtonava . Do we need additional check here if invitee.user_id has already been set. It does exit early on if we already captured a valid status  and trying to "reset"

File: api/src/api/users/user_schemas.py (hunk 3 above)
Comment (babebe) [reply api/src/api/users/user_schemas.py]: @joshtonava  what happens if here is no sam gov entity, this will still work ?

Comment (joshtonava) [reply api/src/api/users/user_schemas.py]: > what happens if here is no sam gov entity, this will still work ?

Well for one im not sure if that is even possible not to have a sam gov entity but the field is nullable in the response object.  The select statement can be tweaked to mitigate some of this i believe. 

We could add a test case as well to really understand this behavior.

File: api/src/services/users/org_invitation_response.py (hunk 4 above)
Comment (joshtonava) [reply api/src/services/users/org_invitation_response.py]: Its likely the user will not exist when invited. First we set the email on the invitation. They will get an email which will tell them they need to setup an account. After that they will login and see their invitation waiting for them. Its only after they accept the invitation (we know at that point the user exists in the system) that we link them to the invitation and setup their roles based on the invitation.

File: api/tests/src/api/users/test_org_invitation_response.py (hunk 6 above)
Comment (joshtonava) [reply api/tests/src/api/users/test_org_invitation_response.py]: Im not sure I understand the question but we should probably have a test case for if invitee.user_id already being set, if you haven't covered that case already.

File: api/src/api/users/user_schemas.py (hunk 3 above)
Comment (chouinar) [reply api/src/api/users/user_schemas.py]: The data model allows for no sam gov entity, although there shouldn't be a scenario in prod at this time where an org exists without the entity.

In theory we might one day support an alternative approach to the sam.gov entity for setting up an organization - for now you can generally assume it has an entity, but make sure the code handles the null case

Comment (babebe) [reply api/src/api/users/user_schemas.py]: If sam.gov entity is null it throws an error. Opted to keep it as a property and handling null

File: api/tests/src/api/users/test_org_invitation_response.py (hunk 6 above)
Comment (babebe) [reply api/tests/src/api/users/test_org_invitation_response.py]: I was trying to imply that the only way that this value would be set is if an invitation has already been responded to. And since we have a validation on the status it would fail, so it is redundunt?. I have added the extra validation just incase.

File: api/src/services/users/org_invitation_response.py [hunk 7]
```diff
@@ -0,0 +1,89 @@
...
+    db_session.add(invitation)
+    return {
+        "organization_invitation_id": invitation.organization_invitation_id,
+        "status": invitation.status,
+        "responded_at": invitation.responded_at,
+        "organization": {
+            "organization_id": invitation.organization_id,
+            "organization_name": invitation.organization.name,
+        },
+        "organization_id": invitation.organization_id,
```
joshtonava: **minor nit:** This field is parsed out by marshmallow on serialization.

api/src/services/users/org_invitation_response.py
```diff
@@ -0,0 +1,89 @@
...
+        # Update response time
+        invitation.accepted_at = now
+
+    if status == OrganizationInvitationStatus.REJECTED:
+        # Update response time
+        invitation.rejected_at = now
+
+    invitation.invitee_user_id = user.user_id
+    db_session.add(invitation)
+    return {
```
joshtonava: I think we can just return the invitation itself instead of constructing an object if we do a few tweaks In user_schemas.py

api/src/api/users/user_schemas.py
```diff
@@ -681,3 +686,55 @@ class UserInvitationListResponseSchema(AbstractResponseSchema):
...
+    status = fields.Enum(
+        OrganizationInvitationStatus, metadata={"description": "User response to invitation"}
+    )
+    responded_at = fields.DateTime(metadata={"description": "Time User responded to invitation"})
+    organization = fields.Nested(
+        OrganizationSchema, metadata={"description": "Organization information"}
+    )
+    roles_granted = fields.List(
+        fields.Nested(InvitationRoleSchema), metadata={"description": "Roles granted"}
+    )
```
joshtonava, line 738:
```suggestion
    roles_granted = fields.List(
        fields.Nested(InvitationRoleSchema), metadata={"description": "Roles granted"}, attribute="roles"
    )
```

This way you dont have to specify roles_granted explicitly in the response

api/src/api/users/user_schemas.py
```diff
@@ -681,3 +686,55 @@ class UserInvitationListResponseSchema(AbstractResponseSchema):
...
+            "example": "456e7890-e12c-34f5-b678-901234567890",
+        }
+    )
+    organization_name = fields.String(
+        allow_none=True,
+        metadata={
+            "description": "Organization name",
+            "example": "Legal business name of the corresponding Sam Gov Entity",
+        },
+    )
```
joshtonava, line 717:
```suggestion
    organization_name = fields.String(
        allow_none=True,
        attribute="name"
        metadata={
            "description": "Organization name",
            "example": "Legal business name of the corresponding Sam Gov Entity",
        },
    )
```

This should allow `organization_name` to work without explicitly specifying it in the response.

api/src/api/users/user_routes.py
```diff
@@ -709,3 +712,35 @@ def user_get_invitations(
...
+def user_response_org_invitation(
+    db_session: db.Session, user_id: UUID, invitation_id: UUID, json_data: dict
+) -> response.ApiResponse:
+    add_extra_data_to_current_request_logs(
+        {
+            "user_id": user_id,
+            "invitation_id": invitation_id,
+        }
+    )
+    logger.info("POST /v1/users/:user_id/invitations/{invitation_id}/organizations")
```
chouinar: Make the pattern of the log format the same
```suggestion
    logger.info("POST /v1/users/:user_id/invitations/:invitation_id/organizations")
```

api/src/db/models/entity_models.py
```diff
@@ -92,6 +92,10 @@ class Organization(ApiSchemaTable, TimestampMixin):
         "Application", uselist=True, back_populates="organization", cascade="all, delete-orphan"
     )
 
+    @property
+    def name(self) -> str | None:
+        return self.sam_gov_entity.legal_business_name if self.sam_gov_entity else None
```
chouinar, line 97: What about naming this `organization_name` so you don't need to specify a mapping in the schema?

joshtonava (approved):

chouinar (approved):

PR #6849, opened 2025-10-24
[Unticketed] Add additional API linting rules

chouinar:
## Changes proposed
Configure additional linting rules for the API

## Context for reviewers
A large portion of this change was made by letting Ruff correct the issues, as most of the issues were fixing typing related things, this shouldn't have any meaningful impact on the API functionality.

I added 3 main groups of rules:
* `UP` - This adds various upgrade related rules like "don't use typing.Dict, instead use dict" or "don't use typing.Sequence, use collections.abc.Sequence".
* `RUF` - Various rules as defined by Ruff (the linter). These vary, but are mostly small consistency fixes like telling you to do `str | None` instead of `None | str`
* `PT` - Some pytest related linting rules - mostly avoids things like duplicate tests, clunky behavior, and 

Added various rules that are either too much effort to switch right now, OR that just didn't seem like stuff we want to follow (too strict, too annoying) without enough upside to the ignore rules. Some rules are specifically skipped for just unit tests.

## Validation steps
This is largely just a linting change, despite the number of files, the actual amount of changes is not too big, most of this just adjusted imports / type definitions (removing `Optional`, `Tuple`, and switching `Sequence`/`Iterable`/`Callable` to a different import path). I triple-checked every change to make sure it wasn't going to impact logic, the only logic changes I made were in our tests, and I kept the same purpose, while fixing some inconsistencies that ruff pointed out.

Nearly all of these were fixed by ruff, not me. I added `make lint-ruff-fix` to be able to run the ruff fix command simply.

### DIFF
--- a/api/src/legacy_soap_api/grantors/schemas/get_submission_list_expanded_schemas.py
+++ b/api/src/legacy_soap_api/grantors/schemas/get_submission_list_expanded_schemas.py
@@ -81,11 +81,9 @@ class ConsolidatedFilter(BaseModel):
             for value in ["FilterType", "FilterValue"]:
                 if value not in item:
                     raise SOAPInvalidEnvelope(
-
-                            "The content of element 'ExpandedApplicationFilter' is not complete. One of "
-                            '\'{"http://apply.grants.gov/system/GrantsCommonElements-V1.0":'
-                            f"{value}}}' is expected."
-
+                        "The content of element 'ExpandedApplicationFilter' is not complete. One of "
+                        '\'{"http://apply.grants.gov/system/GrantsCommonElements-V1.0":'
+                        f"{value}}}' is expected."
                     )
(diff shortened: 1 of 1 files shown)


api/src/data_migration/transformation/transform_constants.py
```diff
@@ -54,21 +54,21 @@ class Metrics(StrEnum):
 S = TypeVar("S", bound=StagingParamMixin)
 D = TypeVar("D", bound=ApiSchemaTable)
 
-SourceSummary: TypeAlias = Tforecast | Tsynopsis
+type SourceSummary = Tforecast | Tsynopsis
```
babebe, line 57: Nice, was not aware of this syntax

joshtonava: yea reminds me of typescript

Issue #6806, opened 2025-10-21
Create tcertificates staging/legacy table


chris-kuryak:
### Summary
https://github.com/HHS/simpler-grants-gov/blob/main/documentation/api/legacy-transforms.md

Table must be created following our normal conventions for setting up a staging/legacy table and be defined in exactly the order listed below.

(columns are nullable unless otherwise stated)

- `currentcertid` - text, primary key
- `previouscertid` - text
- `orgduns` - text
- `orgname` - text
- `expirationdate` - date
- `certemail` - text, not null
- `agencyid` - text
- `requestorlname` - text
- `requestorfname` - text
- `created_date` - datetime, not null
- `creator_id` - text, not null
- `last_upd_date` - datetime
- `last_upd_id `- text
- `is_selfsigned` - text, not null
- `serial_num` - text
- `system_name` - text

### Acceptance criteria

- [ ] Table schema created for tcertificates as described above
- [ ] Migration generated for staging.tcertificates

PR #6858, opened 2025-10-27
[Issue #6806] Create tcertificate legacy staging tables

jakobpederson:
## Summary

Fixes / Work for #6806 

## Changes proposed

Create tcertificate legacy staging tables.

## Context for reviewers

We'll need the tcertificate table for future work using certificates with the legacy SOAP api.

### DIFF
--- a/api/src/db/models/legacy_mixin/certificate_mixin.py
+++ b/api/src/db/models/legacy_mixin/certificate_mixin.py
@@ -12,21 +12,21 @@ from sqlalchemy.orm import Mapped, declarative_mixin, mapped_column
 
 @declarative_mixin
 class TcertificateMixin:
-    currentcertid: Mapped[str] = mapped_column(
-        primary_key=True
-    )  # currentcertid - text, primary key
-    previouscertid: Mapped[str | None]  # previouscertid - text
-    orgduns: Mapped[str | None]  # orgduns - text
-    orgname: Mapped[str | None]  # orgname - text
-    expirationdate: Mapped[date | None]  # expirationdate - date
-    certemail: Mapped[str]  # certemail - text, not null
-    agencyid: Mapped[str | None]  # agencyid - text
-    requestorlname: Mapped[str | None]  # requestorlname - text
-    requestorfname: Mapped[str | None]  # requestorfname - text
-    created_date: Mapped[datetime]  # created_date - datetime, not null
-    creator_id: Mapped[str]  # creator_id - text, not null
-    last_upd_date: Mapped[datetime | None]  # last_upd_date - datetime
-    last_upd_id: Mapped[str | None]  # last_upd_id - text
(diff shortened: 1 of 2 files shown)


api/src/db/models/legacy_mixin/certificate_mixin.py
```diff
@@ -0,0 +1,32 @@
...
+    previouscertid: Mapped[str | None]  # previouscertid - text
+    orgduns: Mapped[str | None]  # orgduns - text
+    orgname: Mapped[str | None]  # orgname - text
+    expirationdate: Mapped[date | None]  # expirationdate - date
+    certemail: Mapped[str]  # certemail - text, not null
+    agencyid: Mapped[str | None]  # agencyid - text
+    requestorlname: Mapped[str | None]  # requestorlname - text
+    requestorfname: Mapped[str | None]  # requestorfname - text
+    created_date: Mapped[datetime]  # created_date - datetime, not null
+    creator_id: Mapped[str]  # creator_id - text, not null
```
chouinar: Whoops, I missed copying them to the ticket, but we also need these two columns. Noticed when double-checking the DB schema.

```suggestion
    requestorlname: Mapped[str | None]  # requestorlname - text
    requestorfname: Mapped[str | None]  # requestorfname - text
    requestoremail: Mapped[str | None]
    requestorphone: Mapped[str | None]
    created_date: Mapped[datetime]  # created_date - datetime, not null
    creator_id: Mapped[str]  # creator_id - text, not null
```

api/src/db/models/legacy_mixin/certificate_mixin.py
```diff
@@ -0,0 +1,32 @@
...
+
+from sqlalchemy.orm import Mapped, declarative_mixin, mapped_column
+
+
+@declarative_mixin
+class TcertificateMixin:
+    currentcertid: Mapped[str] = mapped_column(
+        primary_key=True
+    )  # currentcertid - text, primary key
+    previouscertid: Mapped[str | None]  # previouscertid - text
```
chouinar: Could we trim out the comments? The table itself says the same thing albeit in the python types.

documentation/api/legacy-transforms.md
```diff
@@ -96,6 +96,21 @@ from . import foreignbase
 class TExample(foreignbase.ForeignBase, TExampleMixin):
     __tablename__ = "texample"
 ```
+and update the staging __init__.py and the legacy/foreign __init__.py
```
chouinar, line 99: Thanks for adding this to the docs.

Unfortunately this is a general gotcha of SQLAlchemy, our routes and python imports as a whole. If nothing imports the file, things scanning for what exists won't see it. One of those things that when you know to do it, you don't even think about it, but if you don't it's real confusing.

chouinar (approved):

Issue #5931, opened 2025-08-13
Hide "save" button on forms after submission


chris-kuryak:
### Summary

If an application has been "Submitted", we want to hide the "Save" button on all forms of that competition.

This will allow users to see all of their submitted form information, but avoid any errors if they clicked the "Save" button again.

### Acceptance criteria

- [ ] "Save" button does not appear on forms after application submission

PR #6870, opened 2025-10-28
[Issue #5931] Hide save button when application is submitted

glenblosser-nava:
## Summary

Work for #5931 

## Changes proposed

Adds conditional rendering to the save button, reliant on the application's status not being submitted or accepted 

## Context for reviewers

https://github.com/HHS/simpler-grants-gov/issues/5931

## Validation steps

Submit an application in a dev environment, and verify that the save button disappears

### DIFF
--- a/frontend/src/components/applyForm/ApplyForm.tsx
+++ b/frontend/src/components/applyForm/ApplyForm.tsx
@@ -71,7 +71,7 @@ const ApplyForm = ({
   const { pending } = useFormStatus();
   const t = useTranslations("Application.applyForm");
   const translate = t as unknown as Translator;
-  const isFormLocked = applicationStatus !== "in_progress" || true;
+  const isFormLocked = applicationStatus !== "in_progress";
   const required = translate.rich("required", {
     abr: (content) => (
       <abbr


frontend/src/app/[locale]/(base)/workspace/applications/application/[applicationId]/form/[appFormId]/page.tsx
```diff
@@ -43,6 +46,25 @@ interface formPageProps {
 async function FormPage({ params }: formPageProps) {
   const { applicationId, appFormId } = await params;
   const { data, error } = await getFormData({ applicationId, appFormId });
+  let details = {} as ApplicationDetailsCardProps;
```
Comment (glenblosser-nava) [inline frontend/src/app/[locale]/(base)/workspace/applications/application/[applicationId]/form/[appFormId]/page.tsx]: This block is copied from the top-level application page, using the same logic to fetch the application information

.github/CODEOWNERS
```diff
@@ -3,7 +3,7 @@
 
 * @mdragon # project lead
 
-/documentation/  @acouch @andycochran @btabaska @chouinar @doug-s-nava @joshtonava @mdragon @myduong-navapbc @widal001 @prasnava @jakobpederson @ErinPattisonNava @glenblosser-nava # everyone
+/documentation/  @acouch @andycochran @btabaska @chouinar @doug-s-nava @joshtonava @mdragon @myduong-navapbc @widal001 @prasnava @jakobpederson @ErinPattisonNava @glenblosser-nava @brianlee-nava # everyone
```
doug-s-nava: github is saying that @brianlee-nava may not exist or have access to the project. Certainly seems like the user exists, can we confirm access?

MAINTAINERS.md
```diff
@@ -38,6 +39,7 @@ This is a list of maintainers for this project. See [CODEOWNERS](/.github/CODEOW
 * Billy Daly
 * Eric Valenzuela
 * David Dudas
+* Brian Lee
```
doug-s-nava: Brian should go in the "Product and Program Management" section

frontend/src/components/applyForm/ApplyForm.tsx
```diff
@@ -131,19 +133,22 @@ const ApplyForm = ({
...
-          className="margin-top-0"
-          value="save"
-          onClick={() => {
-            setFormChanged(false);
-            setAttachmentsChanged(false);
-          }}
-        >
-          {pending ? "Saving..." : "Save"}
-        </Button>
+        {applicationStatus !== "submitted" &&
```
doug-s-nava: can you add a unit test for this new behavior?

.github/CODEOWNERS
```diff
@@ -3,15 +3,15 @@
...
 
 /analytics/               @acouch @widal001
 /documentation/analytics/ @acouch @widal001
 
 /api/               @chouinar @mdragon @joshtonava @jakobpederson
 /documentation/api/ @chouinar @mdragon @joshtonava @jakobpederson
 
-/frontend/               @acouch @andycochran @btabaska @doug-s-nava @myduong-navapbc @ErinPattisonNava
+/frontend/               @acouch @andycochran @btabaska @doug-s-nava @myduong-navapbc @ErinPattisonNava @glenblosser-nava
```
doug-s-nava: 👍

frontend/src/app/[locale]/(base)/workspace/applications/application/[applicationId]/form/[appFormId]/page.tsx
```diff
@@ -43,6 +45,24 @@ interface formPageProps {
 async function FormPage({ params }: formPageProps) {
   const { applicationId, appFormId } = await params;
   const { data, error } = await getFormData({ applicationId, appFormId });
+  const userSession = await getSession();
+
+  if (!userSession || !userSession.token) {
+    return <TopLevelError />;
+  }
+  const response = await getApplicationDetails(
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/workspace/applications/application/[applicationId]/form/[appFormId]/page.tsx:53]: what do you think about moving this call into `getFormData`?

pros:

* don't have to duplicate the session check logic and related error handling
* all fetches necessary for the page are encapsulated in the same place

cons:
* getFormData is already too big and complex

Other ideas:

* move the session check out of getFormData and pass the token in?
* encapsulate the logic for this new call in its own function, and make parallel calls with Promise.all?

This is also ok as is, just wanted to make sure we've thought through the alternatives.

.github/CODEOWNERS
glenblosser-nava: I'll double check. He asked me to add him to the codeowners list since I was doing a PR anyway

frontend/src/app/[locale]/(base)/workspace/applications/application/[applicationId]/form/[appFormId]/page.tsx
Comment (glenblosser-nava) [reply frontend/src/app/[locale]/(base)/workspace/applications/application/[applicationId]/form/[appFormId]/page.tsx:53]: I think it's best to keep this separate, because fetching form data and application data shouldn't return the same information. While we need the status here, I think the best practice would be to only pull the minimum required data for each page.

frontend/src/components/applyForm/ApplyForm.tsx
glenblosser-nava: Will do!

frontend/src/components/applyForm/FormFields.tsx
```diff
@@ -81,7 +83,7 @@ export const FormFields = ({
 
             const field = renderWidget({
               type: widgetConfig.type,
-              props: { ...widgetConfig.props, formContext },
+              props: { ...widgetConfig.props, formContext, readonly },
```
doug-s-nava: this won't disable fields on the 424a budget form without more work - since it uses the multiField widget and handles passing down props to individual widgets in a custom way, we'd need to think about that a bit more. The budget form work can be a separate ticket but worth considering in terms of our approach here in case there is something to be gained later by taking a particular approach now

frontend/src/components/applyForm/widgets/AttachmentUploadWidget.tsx
```diff
@@ -142,7 +143,8 @@ const AttachmentUploadWidget = (props: UswdsWidgetProps) => {
           id={id}
           name={id}
           required={required}
-          disabled={disabled}
+          disabled={disabled || readonly}
```
doug-s-nava: this approach means each widget needs to implement disabling fields based on the "readonly" prop individually. Since each field (except in the budget form) is rendered through the "renderWidget" function, and each widget takes a "disabled" prop, we could do this calculation there instead, and pass "disabled: true" from "renderWidget" into each widget whenever "readonly" is true. We'd still have to figure out the budget form, but this should handle every other widget type without touching the individual widgets themselves

frontend/src/components/applyForm/FormFields.tsx
glenblosser-nava: I've added support for 424a budget form

frontend/src/components/applyForm/widgets/AttachmentUploadWidget.tsx
glenblosser-nava: This approach was copied from our other widgets, I can change it if need be. I've gone ahead and had formFields pass both readonly and disabled to its children to simplify future implementation

.github/CODEOWNERS
doug-s-nava: also wanted to make sure that @brianlee-nava knows what he's signing up for here - any time a PR comes out with an update to documentation he'll be pinged with an email for a review. I'm all for that if he wants to contribute that way, but since none of the other PMs are in the list, wanted to check

frontend/src/components/applyForm/FormFields.tsx
```diff
@@ -25,12 +25,14 @@ export const FormFields = ({
...
   uiSchema,
   formContext,
+  readonly,
 }: {
   errors: FormattedFormValidationWarning[] | null;
   formData: object;
   schema: RJSFSchema;
   uiSchema: UiSchema;
   formContext?: RootBudgetFormContext;
+  readonly?: boolean;
```
doug-s-nava: sry to not have mentioned this earlier - can we change this to `readOnly` for camelCase consistency?

frontend/src/components/applyForm/widgets/AttachmentUploadWidget.tsx
doug-s-nava: not sure what you mean about copying from other widgets. My concern is that this way each widget needs to interpret readonly separately when we could potentially isolate that logic in the widget renderer. Maybe we can get on a call to show you what I'm thinking?

frontend/src/components/applyForm/FormFields.tsx
glenblosser-nava: Got it

frontend/src/components/applyForm/widgets/WidgetRenderers.tsx
```diff
@@ -50,6 +50,9 @@ export const renderWidget = ({
     console.error(`Unknown widget type: ${type}`, definition);
     throw new Error(`Unknown widget type: ${type}`);
   }
+  if (props.isFormLocked) {
```
doug-s-nava, line 53: thanks for making this change - with this in place I think you can roll back the `readOnly` changes to each of the individual widgets other than the ones that weren't already handling disabled field states (which I think is the attachment widgets and budget form widgets)

doug-s-nava:
split "readonly / disabled" into its own ticket https://github.com/HHS/simpler-grants-gov/issues/7067

doug-s-nava:
@glenblosser-nava can we close this PR at this point or are you going to come back to it for the disabled fields?

doug-s-nava (approved):
✅ ✅ ✅ ✅

Issue #6869, opened 2025-10-28
Hiding the submit button should only happen once the competition is closed, and time should default to end of day


chris-kuryak:
### Summary

User Story: As an applicant, I would like the date time on Close date to reflect more accurately what the competition allows, and not hide the Submit button until the competition is officially closed.

Currently, it appears that the Submit button on the application page is referencing the Close date on the competition and hiding prematurely (at the beginning of the day, instead of the end of the day).

<img width="334" height="148" alt="Image" src="https://github.com/user-attachments/assets/a51b33db-03fd-4af9-b7c1-ee22fa5b28c1" />

Expected Behavior: The hiding of the Submit Button and the text indicating "No Longer Accepting Applications" should look at the `is_open` flag and not do any date comparison at all. And the time shown next to the date should be 11:59pm ET, not 12am.

### Acceptance criteria

- [ ] Hiding the Submit button and triggering "No Longer Accepting Applications" triggers when `is_open` is false
- [ ] The close date now defaults/shows 11:59pm instead of 12am
- [ ] UAT passed

PR #6871, opened 2025-10-28
[ISSUE #6869]: Hiding the submit button should only happen once the competition is closed

myduong-navapbc:
## Summary

Work for #6869  

## Changes proposed

Hiding the Submit button and triggering "No Longer Accepting Applications" triggers when is_open is false
The close date now defaults/shows 11:59pm instead of 12am

## Validation steps

3. navigate to `search`
4. search for `pilot`
6. sign in
7. start new application

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -38,6 +38,7 @@ COPY newrelic.ini /api/newrelic.ini
 RUN PY_MM=$(cat /tmp/py_version.txt) && \
     /usr/local/bin/python"${PY_MM}" -m poetry build --format wheel && \
     /api/.venv/bin/python -m pip install --no-cache-dir dist/simpler_grants_gov_api-0.1.0-py3-none-any.whl && \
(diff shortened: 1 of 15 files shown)


frontend/src/components/application/InformationCard.tsx
```diff
@@ -202,17 +195,17 @@ export const InformationCard = ({
                 <span className="text-bold text-orange">
                   {applicationDetails.competition.closing_date}
                 </span>{" "}
-                (12:00am ET)
+                (11:59pm ET)
               </dd>
             </div>
-            {isCompetitionClosed() ? <SpecialInstructions /> : null}
+            {is_open ? <SpecialInstructions /> : null}
```
doug-s-nava: these both seem reversed. Shouldn't this be shown when `!is_open` (reverse below)

doug-s-nava (commented):
it doesn't look like there are any tests around the behavior of hiding the submit button, can we add some?

PR #6874, opened 2025-10-28
[Issue #6780] Add a local-only endpoint that returns user info/keys

### LINKED ISSUE #6780: build local endpoint for returning login data for test users
### Summary

In order to allow the frontend to "log in" as a local test user without going through the motions of the mock auth login process, we will build an endpoint to return relevant information for all users including an API side JWT.

This can be a GET endpoint - there is no need for pagination or filtering.

This endpoint should go into a new "local blueprint" for grouping routes we'll only enable locally.

This endpoint will only function in a local environment. To make certain it never is used outside of a local environment we should:
* Put our `error_if_not_local()` check at the start of the route implementation
* Only connect the blueprint if the environment is `local` AND an env var flag (that defaults to False) is set to True

#### Logic
We want to fetch every user from the DB and return them in a list. Along with that we want to grab a JWT and user API key to return. For keeping things simple for those using this endpoint we want to return just one of each (null if they have none).

For both of them, we'll want to avoid anything already expired/inactive, and for the JWT grab whatever is the furthest one from expiring.

We want to actually return the JWT that the API would generate, which isn't stored in the DB, we'll need to run it through the same JWT logic in `create_jwt_for_user` but without making a new one, just passing in the existing token ID.

#### Response
```json
{
  "data": [
     {
        "user_id": "<some uuid>",
        "user_jwt": "abc123",
        "user_api_key": "abc123",
        // including first/last name as we could use them to store a description of the user if we want to display that
        "first_name": "bob",
        "last_name": "smith"
      }
   ]
}
```

### Acceptance criteria

- [ ] At minimum, endpoint exists that will return a list of all users (in a local dev environment, all users are test users) including an API side JWT token
- [ ] **endpoint does no exist / returns error in all deployed envs**

chouinar:
## Summary
Fixes #6780

## Changes proposed
Create a local-only endpoint that returns all users from the DB along with auth information to help make development simpler locally

Make it so our local users have a JWT that is set to expire very far in the future

Reorganize our JWT creation logic so we can make a JWT in this new endpoint

## Context for reviewers
This is an endpoint that will only ever work locally, I put half a dozen different checks in place to make sure this endpoint both doesn't exist non-locally, and will error if run non-locally (if you have any other ideas, let me know). It basically dumps our user table and all their auth information. Locally this will help us more easily manage users. The frontend will start building a UI for easily switching users, and for purely backend work we can use it to get keys on the swagger/openapi docs.

As this implementation is a massive security problem if it were ever deployed, I added the following safety-nets:
* The route is only enabled locally, by default it is disabled. Unless we add to our terraform configuration, it can't be enabled in other envs.
* The environment needs to be exactly `local` or every function will error.
* If somehow it gets passed either of those, I added a safety net that will make it so if more than 500 users are returned, it'll error. This is because prod already has about 8000 users.
* Added error logs so we'd be alerted if this were ever running non-locally.

I'm open to other ideas, definitely the sort of thing I think we can't be too careful on.

## Validation steps
The endpoint can be tested locally on the swagger docs:
Screenshot 2025-10-28 at 3 48 36 PM


mikehgrantsgov (approved):

api/src/services/local/get_local_users.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def _get_user_jwt(db_session: db.Session, user: User) -> str | None:
+    # This is really hacky and meant for local dev only
+    error_if_not_local()
+
+    # Grab the token session that expires the furthest in
+    # the future. In our seed script we usually generate
+    # these to expire many years out, so it's probably that one.
+    user_token_session = db_session.execute(
```
mikehgrantsgov, line 64: Does SQLAlchemy batch these selects so it doesn't run a query per user? Not that performance really matters but more curious about SQLAlchemy than anything.

chouinar: If you do `db_session.execute(select(User)).scalars().all()` SQLAlchemy will do `select col1, col2, ... from api.user`. When you add `selectinload` to the queries, that initial query is left alone, but for every table it does a query like:

```sql
SELECT * FROM user_api_key 
WHERE user_api_key.user_id IN (%(primary_keys_1)s::UUID, %(primary_keys_2)s::UUID, %(primary_keys_3)s::UUID, %(primary_keys_4)s::UUID, %(primary_keys_5)s::UUID, %(primary_keys_6)s::UUID)
```

Basically after the initial select completes, it grabs all the relevant IDs and does a `select * from table where id in (...)`. It does that for each of them and then puts all the objects back together for you.

If you're curious about `joinedload` which I know I've told people not to use, instead of doing these follow-up queries, it [does joins in the initial query](https://docs.sqlalchemy.org/en/20/orm/queryguide/relationships.html#joined-eager-loading). That's fine in some cases, but causes uniqueness issues and when you add enough of them, the join becomes pretty awful / I've seen it just lock up the query entirely. `selectinload` is nice because the queries stay really simple, it's literally just a bunch of `select * from x where id in ()` and at worst the performance is mediocre.

---

If you're ever curious about the exact queries being run, in our postgres client code, pass `echo=True` into `create_engine` and it'll dump all of its queries out. Just be aware that it is a lot since it both names every column and each SQL statements gets logged twice + if you run in something like our tests there are a lot of SQL commands run before anything else. I usually put a print statement before/after the SQL gets run to find it.

chouinar: Note, if we didn't have the `selectinload`, it would do the join per relationship since the default is to lazyload everything (eg. doing `user.api_keys` emits a SQL query to fetch the api keys for just that user).

PR #6881, opened 2025-10-29
[Issue #6677] Create an endpoint for adding an organization to an application

### LINKED ISSUE #6677: Create an endpoint for adding an organization to an application
### Summary

`PUT /alpha/applications/:application_id/organizations/:organization_id`

## Response
`{"data": {"application_id": "<some uuid>"}}` 

basically the same as the update application endpoint

## Authentication

User must have the `Privilege.MODIFY_APPLICATION` for the application

User must ALSO have `Privilege.START_APPLICATION` for the organization

Note we have to do two checks for privileges because we need to make sure they can do two separate actions, actually change the app AND that they’re even in the org that is being passed in. 

## Logic

We want to error if ANY of the following are true:

* The application already has an organization (422)
* The competition ONLY accepts individual applications (don’t want users to be put in a bad spot) (422)
* The app/org don’t exist (404)
* User doesn’t have required privileges as described above (403)

When we add an organization to an application we want to also do the following:
* Remove ALL application user records (in the future we might make this a bit more refined, but this is necessary for now)
* Rerun the pre population logic so that values from the org get updated into a users responses (like UEI) - similar to what we do in the start-application logic (but on the already-existing forms, don't erase a users answers).

### Acceptance criteria

- [ ] Endpoint created meeting above criteria
- [ ] Unit tests

mikehgrantsgov:
## Summary

Fixes #6677 

## Changes proposed

Create `PUT /alpha/applications/:application_id/organizations/:organization_id` endpoint
Allow that endpoint, with authz applied, to add an organization to an application
Add supporting tests.

## Context for reviewers

There are many new tests to cover the success/error scenarios based on the below logic.

For auth:

User must have the Privilege.MODIFY_APPLICATION for the application

User must ALSO have Privilege.START_APPLICATION for the organization

There are many error scenarios like:
- The application already has an organization (422)
- The competition ONLY accepts individual applications (don’t want users to be put in a bad spot) (422)
- The app/org don’t exist (404)
- User doesn’t have required privileges as described above (403)

Also some other business rules to consider when we add an organization to an application we want to also do the following::

- Remove ALL application user records (in the future we might make this a bit more refined, but this is necessary for now)
- Rerun the pre population logic so that values from the org get updated into a users responses (like UEI) - similar to what we do in the start-application logic (but on the already-existing forms, don't erase a users answers).

## Validation steps

See new / updated unit tests.

### DIFF
--- a/api/src/services/applications/add_organization_to_application.py
+++ b/api/src/services/applications/add_organization_to_application.py
@@ -99,7 +99,8 @@ def add_organization_to_application(
     # Delete all application user records
     # We do this because we need to reset the application to the state it would be in
     # if an organization started a brand new application. Individual user associations
-    # are not relevant when an application is associated with an organization.
+    # need to be cleared so that the user who started the application doesn't have
+    # more privileges than they would have had they created it in the organization to begin with.
     for app_user in application.application_users:
         db_session.delete(app_user)
 


api/src/services/applications/add_organization_to_application.py
```diff
@@ -0,0 +1,127 @@
...
+                "existing_organization_id": application.organization_id,
+            },
+        )
+        raise_flask_error(422, "Application already has an organization")
+
+    # Get the organization
+    organization = get_organization(db_session, organization_id)
+
+    # Check user has START_APPLICATION privilege for the organization
+    if not can_access(user, {Privilege.START_APPLICATION}, organization):
```
chouinar, line 62: Should we group the validation so the get/404 and can_modify/403s happen together?:
* get app or 404
* get org or 404
* can modify app or 403
* can start app in org or 403
* the rest of the validation rules

I could go either way, I think that way would be easier to parse, but not sure if it's "right" from an auth perspective.

---

Alternatively, if we put some of these chunks into functions so the logic in this function was a bit briefer:
```python

application = get_application(...) # gets + 404s if not found

if not can_access(... application):
    ...

organization = get_organization(...)

if not can_access(... organization):
    ...

validate_whatever(...)
remove_app_users(...)
...
```

api/src/services/applications/add_organization_to_application.py
```diff
@@ -0,0 +1,127 @@
...
+    # Validate application doesn't already have an organization
+    if application.organization_id is not None:
+        logger.info(
+            "Application already has an organization",
+            extra={
+                "application_id": application_id,
+                "existing_organization_id": application.organization_id,
+            },
+        )
+        raise_flask_error(422, "Application already has an organization")
```
chouinar, line 84: Could we put this after the auth checks, next to the other 422 check?

I think we'd want to verify a user can do the operation first before validating whether the operation is valid (except for the 404s as we can't really work around that).

api/src/services/applications/add_organization_to_application.py
```diff
@@ -0,0 +1,127 @@
...
+                "application_id": application_id,
+                "competition_id": application.competition.competition_id,
+                "submission_issue": SubmissionIssue.COMPETITION_NO_ORG_APPLICATIONS,
+            },
+        )
+        raise_flask_error(422, "This competition only accepts individual applications")
+
+    # Delete all application user records
+    for app_user in application.application_users:
+        db_session.delete(app_user)
```
chouinar, line 105: Feel free to reword, but I think we should add some sort of comment here to explain why we're just deleting users off of the app here.

```suggestion
    # Delete all application user records
    # We do this because we need to reset this
    # to the state that an application would be in
    # if an organization started a brand new application.
    for app_user in application.application_users:
        db_session.delete(app_user)
```

api/tests/src/services/applications/test_add_organization_to_application.py
```diff
@@ -0,0 +1,406 @@
...
+    FormFactory,
+    OrganizationFactory,
+    OrganizationUserFactory,
+    OrganizationUserRoleFactory,
+    RoleFactory,
+    UserFactory,
+)
+
+
+def test_add_organization_to_application_success(enable_factory_create, db_session):
```
chouinar, line 31: Could we add a few basic tests at the route level, don't need to be as thorough as you were here, just basic happy test, 404, some that hit the auth scenarios. Just as a sanity check that everything is connected together correctly.

api/src/services/applications/add_organization_to_application.py
mikehgrantsgov: Good call, so it's 404 app/org, 403, validation errors, then validate, then remove_app_users

api/src/services/applications/add_organization_to_application.py
mikehgrantsgov: Yep, this was moved.

api/src/services/applications/add_organization_to_application.py
mikehgrantsgov: Added a comment based off your suggestion, thanks.

api/tests/src/services/applications/test_add_organization_to_application.py
mikehgrantsgov: Added a couple quick tests that test the end to end flow from the routes down.

api/src/services/applications/add_organization_to_application.py
```diff
@@ -0,0 +1,130 @@
...
+                "competition_id": application.competition.competition_id,
+                "submission_issue": SubmissionIssue.COMPETITION_NO_ORG_APPLICATIONS,
+            },
+        )
+        raise_flask_error(422, "This competition only accepts individual applications")
+
+    # Delete all application user records
+    # We do this because we need to reset the application to the state it would be in
+    # if an organization started a brand new application. Individual user associations
+    # are not relevant when an application is associated with an organization.
```
chouinar: This isn't quite true, they are relevant, there will be cases where an organization application wants non-org members added to the application as people often bring consultants in to help them write applications. We need to do it in this case because we at the very least need to remove the person who started the application. Removing everyone might be more than we want to do BUT there are a lot of edge cases that we agreed with design weren't worth figuring out now and we'd just start it fresh.

This doesn't quite capture all of that nuance, but something like:
```suggestion
    # if an organization started a brand new application. Individual user associations
    # need to be cleared so that the user who started the application doesn't have
    # more privileges than they would have had they created it in the organization to begin with.
```

mikehgrantsgov: Got it, thanks for clarifying that nuance

Issue #6124, opened 2025-08-26
Apply endpoints - Modify list application endpoint to support pagination


joshtonava:
### Summary

Include pagination as part of the list application endpoint request and response.

### Tech Notes

- Request and response models should reuse our existing pagination approach
- We have a pagination utility for making DB queries paginated that we should use for the query we created.
- This for listing applications by user

### Acceptance criteria

- Add pagination to the application list endpoint (TODO put endpoint path here)
- Add / update unit tests

joshtonava:
- This probably needs to be a post endpoint which can take in a bunch of different filters.
    - package id, opportunity id, application status, submission method (came through API or Web)
- Will likely need to version this API to not break existing behavior.

mdragon:
Existing SOAP endpoint supports more like a search level of filtering:
https://www.grants.gov/system-to-system/grantor-system-to-system/web-services/get-submission-list-expanded 
https://www.grants.gov/system-to-system/grantor-system-to-system/schemas/grants-common-elements#ExpandedApplicationFilter

eric-serrato95:
List application by user @joshtonava

PR #6891, opened 2025-10-30
[Issue #6124] user list application endpoint add pagination

babebe:
## Summary

Fixes / Work for #6124  

## Changes proposed

* Pagination added to endpoint
* Filter on `application_status`, `organization_id`, `competion_id` added


api/src/api/users/user_schemas.py
```diff
@@ -255,10 +255,36 @@ class UserOrganizationsResponseSchema(AbstractResponseSchema):
...
+            ApplicationStatus,
+            metadata={"description": "Status of the application"},
+        )
+    )
+    organization_id = fields.List(
+        fields.UUID(metadata={"description": "The internal ID of the organization"})
+    )
+    competition_id = fields.List(
+        fields.UUID(metadata={"description": "The internal ID of the competition"})
+    )
```
chouinar: For filters, we should use the same filter pattern as our other endpoints for consistency. So for application status, can do:

```py
application_status = fields.Nested(
   StrSearchSchemaBuilder("UserApplicationApplicationStatusFiilterSchema")
      .with_one_of(ApplicationStatus)
      .build()
)
```
When we need to parse this in the service layer, we already have some pydantic models for this, so it'd just need to be:

```py
class ApplicationFilters(BaseModel):
     application_status: StrSearchFilter | None = None
     ...
```
This should still work with our enums uneventfully

---

Although it doesn't look like we have a similar builder for UUIDs - but I think that one would be basically just this:

```py
class UuidSearchSchemaBuilder(BaseSearchSchemaBuilder):

    def with_one_of(self) -> "UuidSearchSchemaBuilder":
        self.schema_fields["one_of"] = fields.List(
            fields.UUID(), allow_none=True
        )
        return self
```

api/src/services/users/get_user_applications.py
```diff
@@ -70,8 +103,22 @@ def get_user_applications(db_session: db.Session, user_id: UUID) -> list[Applica
             selectinload(Application.organization).selectinload(Organization.sam_gov_entity),
         )
     )
+    # Filter
+    filters = list_params.filters
+    if filters:
+        filter_clauses = build_filter_clauses(filters)
+        if filter_clauses:
+            stmt = stmt.where(and_(*filter_clauses))
```
chouinar: A bit nitpicky - Just to avoid some nested if-statements. What if we just made it so the first line of `build_filter_clauses` checked if filters was None (returning []) and then we make this one less nested if-statement.

```suggestion
    filter_clauses = build_filter_clauses(list_params.filters)
    if filter_clauses:
        stmt = stmt.where(and_(*filter_clauses))
```

api/src/api/schemas/search_schema.py
```diff
@@ -349,3 +350,37 @@ def validate_date_range(_: Any, data: dict, **kwargs: Any) -> None:
...
+    """
+
+    def with_one_of(
+        self,
+        example: uuid.UUID | None = None,
+    ) -> "UuidSearchSchemaBuilder":
+        metadata = {}
+        if example is not None:
+            metadata["example"] = example
+        self.schema_fields["one_of"] = fields.List(fields.UUID(), allow_none=True)
```
chouinar: I'd probably just exclude example for UUID. The UUID type itself automatically has an example value default that I don't see much reason to override.
```suggestion
    ) -> "UuidSearchSchemaBuilder":
        self.schema_fields["one_of"] = fields.List(fields.UUID(), allow_none=True)
```
If we did want the example, it would need to be passed into the `fields.UUID` function.

Issue #6893, opened 2025-10-30
Test building docker image on the fly for vuln scan vs loading a cached one


mdragon:
### Summary

Currently we build the docker image, cache it in GH, and then restore from that cache for each of the vulnerability scan runners. The hypothesis is that it's faster (better use of cache) to cache the docker layers the build creates and rebuild within any future jobs that need the Docker image.

### Acceptance criteria

- If faster/more reliable enable this as the default method of getting docker images to test against

mdragon:
This isn't faster than status quo, so abandoning this for now.

PR #6895, opened 2025-10-30
[Issue #6893] Rebuild Docker image from Docker Build Cache for Vulnerability Scan

mdragon:
…l uses the previously built cached image

## Summary

Fixes / Work for #ISSUE


mdragon:
This doesn't significantly improve the build times so shelving for now.

Issue #6597, opened 2025-10-03
Build table on the Application List page


chris-kuryak:
### Summary

This ticket is to build the table on the application list page.

Design in [Figma for reference](https://www.figma.com/design/zPlDWcYM396szw2yL3UXvk/Simpler-Grants-2025?node-id=4803-44917&t=VFv6ubMWwkVscLwZ-1).

- Application Type: should display the org name or "Individual" if no org name is returned by the endpoint.
- Status: We will have only two statuses shown on the front-end in the MVP
  - "Draft" = (In-Progress)
  - "Submitted" = (Submitted, Accepted)

### Acceptance criteria

- [ ] Table exists per Figma design
- [ ] Application type displays org name or "Individual"
- [ ] Status displays either "Draft" or "Submitted" per the guidance above in the ticket
- [ ] Design UAT passed
- [ ] Product UAT passed

chelsea-nava:
@chris-kuryak @ErinPattisonNava [Figma has been updated here](https://www.figma.com/design/zPlDWcYM396szw2yL3UXvk/Simpler-Grants-2025?node-id=6147-101014&t=zN8gMJqz1mSUJxxF-1) for the more fleshed out designs, just for your reference, please reach out with questions/issues. Thanks!

ErinPattisonNava:
@eric-serrato95 Ready for UAT!

eric-serrato95:
@ErinPattisonNava UAT Done, moving to closed!

Issue #6595, opened 2025-10-03
Create Application list view page


chris-kuryak:
### Summary

This is the ticket to make the page exist, but does not add table or information to the page.

Design in [Figma for reference](https://www.figma.com/design/zPlDWcYM396szw2yL3UXvk/Simpler-Grants-2025?node-id=4803-44917&t=VFv6ubMWwkVscLwZ-1).

Unless told otherwise, a text link of "My Applications" appears on the Workspace page.

### Acceptance criteria

- [ ] Page exists in the website structure as expected
- [ ] Link/Button exists on the Workspace page

ErinPattisonNava:
@chris-kuryak @eric-serrato95 Not sure if you want to do UAT on the skeleton, or just review once we have the full page/make it accessible to users.  Note that I did pull in a small bug about shifting font sizes based on feedback from design, see the PR for details in the description.

chris-kuryak:
@ErinPattisonNava No need from my perspective to do a UAT on just the blank page -- let's do UAT once the table is on there. Thanks!

PR #6907, opened 2025-10-31
feat: adding the skeleton of the Applications page

ErinPattisonNava:
## Summary

Work for [#6595 ](https://github.com/HHS/simpler-grants-gov/issues/6595)

## Changes proposed

- [ ] Adding the skeleton for the Applications page at /applications
- [ ] Fixing the shifting h1 font on the Saved Opportunities, Saved Searches, and API Dashboard page 

## Context for reviewers

Will be followed up with [#6597](https://github.com/HHS/simpler-grants-gov/issues/6597) to build the table with data and the authentication check, currently the page is accessible without logging in.

The shifting h1 fonts on the Saved Opportunities, Saved Searches, and API Dashboard page was discovered when we standardizing the look across similar Workspace pages.  The font shifts smaller for tablet, then pops bigger again for smaller breakpoints.  Design asked to remove that, apparently there had been an earlier initiative to do that and these pages just got missed.

## Validation steps

- [ ] Go to http://localhost:3000/applications to see the skeleton page.  There is no link to it yet on the website.
- [ ] Check that the font looks ok at  all breakpoints on these pages: 
   - [ ] Saved Opportunities as a logged in user at  http://localhost:3000/saved-opportunities
   - [ ]  Save Searches page as a logged in user at http://localhost:3000/saved-search-queries
   - [ ] API Dashboard page as a logged in user at http://localhost:3000/api-dashboard (note that the odd spacing at the top is current behaviour, this is just about the shifting of the font size)

### DIFF
--- a/frontend/src/app/[locale]/(base)/applications/page.tsx
+++ b/frontend/src/app/[locale]/(base)/applications/page.tsx
@@ -22,16 +22,14 @@ export default function Applications({ params }: LocalizedPageProps) {
   const t = useTranslations("Applications");
 
   return (
-    <>
-      <GridContainer>
-        <h1 className="margin-top-9 margin-bottom-7">{t("pageTitle")}</h1>
-        <div className="margin-bottom-15">
-          <div className="font-sans-xl text-bold margin-bottom-3">
-            {t("noApplicationsMessage.primary")}
-          </div>
-          <div>{t("noApplicationsMessage.secondary")}</div>
+    <GridContainer>
+      <h1 className="margin-top-9 margin-bottom-7">{t("pageTitle")}</h1>
+      <div className="margin-bottom-15">
+        <div className="font-sans-xl text-bold margin-bottom-3">
+          {t("noApplicationsMessage.primary")}
         </div>
-      </GridContainer>
-    </>
+        <div>{t("noApplicationsMessage.secondary")}</div>
+      </div>
+    </GridContainer>
   );
 }


frontend/src/app/[locale]/(base)/api-dashboard/page.tsx
```diff
@@ -22,9 +22,7 @@ export default async function ApiDashboardPage({ params }: LocalizedPageProps) {
     return (
       <>
         <div className="grid-container">
-          <h1 className="tablet-lg:font-sans-xl desktop-lg:font-sans-2xl margin-top-0">
-            {t("heading")}
-          </h1>
+          <h1 className="margin-top-0">{t("heading")}</h1>
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/api-dashboard/page.tsx:25]: no objection to these updates, but curious where the requirements came from. I feel like definitely need to adjust this spacing in some other places as well...

frontend/src/app/[locale]/(base)/applications/page.tsx
```diff
@@ -0,0 +1,37 @@
...
+  return meta;
+}
+
+export default function Applications({ params }: LocalizedPageProps) {
+  const { locale } = use(params);
+  setRequestLocale(locale);
+  const t = useTranslations("Applications");
+
+  return (
+    <>
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/applications/page.tsx]: GridContainer can be the wrapper here, no need for the fragment

frontend/src/app/[locale]/(base)/api-dashboard/page.tsx
Comment (ErinPattisonNava) [reply frontend/src/app/[locale]/(base)/api-dashboard/page.tsx:25]: Since the Applications page was going to live under the Workspace area, I was looking at other pages there to steal some of their similar styling.  What I found was that the h1 sizing would shrink at one breakpoint, then bounce back to the original size at a lower breakpoint.  When I talked to design, they said that was a bug that they had changed a while back and just missed some cases.  

I should have put that context in the description, fixed!

ErinPattisonNava:
> looks good, one minor code cleanup request.
> 
> Also, a question around when we plan to implement authentication gating here. It probably makes sense to do it in #6597 but the ticket doesn't call that out. Probably worth updating the ticket with the requirement for handling unauthenticated users

@doug-s-nava Good call on the other ticket, I'll get that added along with an additional error scenario.  I'm assuming that for unauthenticated user it just takes them to the login page; is there a different pattern we've used other places?

doug-s-nava:
@ErinPattisonNava in general for unauthenticated users we don't redirect, but display this message instead https://github.com/HHS/simpler-grants-gov/blob/main/frontend/src/components/user/UnauthenticatedMessage.tsx . If there's no particular behavior specified for a page, I would follow how other pages use this component.

frontend/src/app/[locale]/(base)/applications/page.tsx
Comment (ErinPattisonNava) [reply frontend/src/app/[locale]/(base)/applications/page.tsx]: Fixed in https://github.com/HHS/simpler-grants-gov/pull/6907/commits/a1d0647230b458d35439a60c5a8ebc2b0213d3a8

doug-s-nava (approved):
✅ ✅ ✅ ✅

chouinar:
@ErinPattisonNava - Something funky happened with your latest commit, a bunch of changes from main seem to have been added as if they came from this branch.

ErinPattisonNava:
> @ErinPattisonNava - Something funky happened with your latest commit, a bunch of changes from main seem to have been added as if they came from this branch.

My bad, I goofed when I was trying to grab the most recent version of main. All fixed.  @doug-s-nava can you approve again?  Nothing changed from the last time you looked at it.

doug-s-nava (approved):
✅ ✅ ✅ ✅

Issue #6675, opened 2025-10-08
Create validation logic to verify an application that requires an org has an org


chris-kuryak:
### Summary

Should verify that if the competition on an application requires exactly an organization (not an individual), and no organization is on the application, a validation issue is added.

This logic should be called in both our GET application endpoint and SUBMIT application logic.
This logic could be added to get_application_form_errors which would handle making it available in both endpoints.

Validation Error

- `message` - Application requires organization in order to submit
- `type` - organization_required - add to our ValidationErrorType enum

### Acceptance criteria

- [ ] GET application endpoint includes logic to produce a validation warning (2xx) with `message` and `type` in above scenario
- [ ] SUBMIT application endpoint includes logic to produce a validation error (422) with `message` and `type` in above scenario
- [ ] Unit tests

PR #6911, opened 2025-10-31
[Issue #6678] Modify the start application endpoint to allow an individual to start an application when only orgs are allowed

### LINKED ISSUE #6678: Modify the start application endpoint to allow an individual to start an application when only orgs are allowed
### Summary

Note - if only individuals are allowed, we should still prevent an org from starting an app

In #6675, we will add logic that prevents an expired/inactive org from submitting to be a validation issue in GET/SUBMIT app. This ticket is to remove the blocker that currently exists for starting an app.

- Currently we have logic in our start application logic contained in `_validate_organization_expiration` - we should keep the general logic, but instead of erroring, it should add a validation issue to the list (which will be a warning in GET, and error for submit)

- This could be handled in both endpoints by adding logic to get_application_form_errors

### Acceptance criteria
- [ ] Logic that prevents an expired/inactive org from submitting in `_validate_organization_expiration` instead provides a validation issue but does not error
- [ ] If opportunity is for individuals only, orgs are prevented from starting an application

mikehgrantsgov:
## Summary

Fixes #6678

## Changes proposed

Remove individual application restriction for org-only competitions
Refactor organization expiration validation
Add new validation error types
Add/update unit tests

## Context for reviewers

Builds on prior work in #6675 and #6677 that enable users to add organizations to applications after starting them.

In our start-application logic we want to remove the check that prevents an individual from starting an application against an organization-only competition. 

Additionally, we adjusted the logic that prevents an organization from starting an application when their entity is expired (the logic contained in `_validate_organization_expiration`). We moved this logic out of the start-application logic and instead
added list of validation issues to our GET application / submit application logic.

## Validation steps

See new/updated unit tests.

### DIFF
--- a/api/tests/src/api/applications/test_application_routes.py
+++ b/api/tests/src/api/applications/test_application_routes.py
@@ -4080,9 +4080,7 @@ def test_add_organization_no_start_application_privilege(
     assert "Forbidden" in response.json["message"]
 
 
-def test_application_get_shows_expired_org_warning(
-    client, enable_factory_create, db_session
-):
+def test_application_get_shows_expired_org_warning(client, enable_factory_create, db_session):
     """Test that GET application shows warning when organization has expired SAM.gov entity"""
     today = get_now_us_eastern_date()
     past_expiration_date = today - timedelta(days=30)
@@ -4098,13 +4096,13 @@ def test_application_get_shows_expired_org_warning(
     )
 
     competition = CompetitionFactory.create()
-    
+
     # Start an application with the expired org
     application = ApplicationFactory.create(
         competition=competition,
         organization=organization,
     )
-    
+
     # Associate user with application
     ApplicationUserRoleFactory.create(
         application_user=ApplicationUserFactory.create(user=user, application=application),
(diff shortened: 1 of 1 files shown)


api/tests/src/services/applications/test_create_application.py
```diff
@@ -19,52 +15,51 @@
 
 
 def test_validate_organization_expiration_no_sam_gov_entity():
```
chouinar: Since we moved the `get_organization_expiration_errors` function out of create application, could we move the tests too since they aren't used with create app anymore.

api/tests/src/api/applications/test_application_routes.py
chouinar, line 1: Could we add a test for GET application and submit application to verify if they have an expired org, the error appears (as a warning / error+422 respectively)? Doesn't need to be all the scenarios, just a sanity test that we did connect things in the right place.

api/tests/src/services/applications/test_create_application.py
mikehgrantsgov: Makes sense, moved these!

api/tests/src/api/applications/test_application_routes.py
mikehgrantsgov: Added `test_application_submit_fails_with_expired_org` and `test_application_get_shows_expired_org_warning` to account for this

chouinar (approved):

PR #6915, opened 2025-10-31
[Issue #6908] Only fail the weekly vuln reporting job if something fails, not if it reports vulnerabilities

### LINKED ISSUE #6908: Expand skip ignore for weekly vuln scan to not fail the job
### Summary

We run the vuln scan Fridays with no ignores to ensure we're tracking ongoing fixed vulns we're ignoring. But the job "fails" every Friday because anything above medium fails the job. We might miss the scans not running because that's also a failure. 

### Acceptance criteria

"Successful runs" that find vulnerabilities but run to completion will not "fail" the job.

mdragon:
## Summary

Fixes #6908  

## Changes proposed

Allow the option for the vuln scans to not fail job if they find vulnerabilities

## Context for reviewers

When the job runs weekly, we bypass the ignore files so that we get a full report of all fixed vulnerabilities, even if we're ignoring them. But that would cause the job to fail as long as there were any active ignores. This makes is so ignored vulnerabilities don't fail the job, but an issue running the scanners would still fail the job.

### DIFF
--- a/.github/workflows/vulnerability-scans.yml
+++ b/.github/workflows/vulnerability-scans.yml
@@ -4,6 +4,22 @@
 name: Vulnerability Scans
 
 on:
+  workflow_dispatch:
+    inputs:
+      app_name:
+        description: "name of application folder under infra directory"
+        required: true
+        type: string
+      bypass_ignore:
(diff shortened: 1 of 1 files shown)


prasnava (approved):
👍

Issue #6812, opened 2025-10-21
Create a legacy_certificate table for storing certs


chris-kuryak:
### Summary

We want to create a `legacy_certificate` API table with the following columns:
- legacy_certificate_id - UUID, primary key
- cert_id - text, not null, indexed
- serial_number - text, not null, indexed (it is nullable in [grants.gov](http://grants.gov/) but we can’t meaningfully use it without this)
- expiration_date - date, not null (also nullable technically in [grants.gov](http://grants.gov/), but we won’t allow that, seems to only be 10+ year old certs).
- user_id - UUID, foreign key to the user table, not null
- agency_id - UUID, foreign key to agency table, nullable
  - If this is set, this is the default agency for queries
- organization_id - UUID, foreign key to organization table, nullable
  - If this is set, this is the default organization for queries

### Acceptance criteria
- [ ] Table created
- [ ] Factory created for table
- [ ] Migration generated

PR #6918, opened 2025-10-31
[Issue #6812] Create legacy_certificate table

jakobpederson:
## Summary

Fixes / Work for #6812 

## Changes proposed

Add legacy_certificate table.

## Context for reviewers

Links users, organization, agencies to certificates from legacy system. Will be needed for RBAC stuff going forward.

### DIFF
--- a/api/src/db/models/user_models.py
+++ b/api/src/db/models/user_models.py
@@ -461,17 +461,14 @@ class LegacyCertificate(ApiSchemaTable, TimestampMixin):
     legacy_certificate_id: Mapped[uuid.UUID] = mapped_column(
         UUID, primary_key=True, default=uuid.uuid4
     )
-    cert_id: Mapped[str] = mapped_column(index=True, nullable=False)
-    serial_number: Mapped[str] = mapped_column(index=True, nullable=False)
-    expiration_date: Mapped[date] = mapped_column(index=True, nullable=False)
-
-    user_id: Mapped[uuid.UUID] = mapped_column(UUID, ForeignKey(User.user_id), nullable=False)
+    cert_id: Mapped[str] = mapped_column(index=True)
+    serial_number: Mapped[str] = mapped_column(index=True)
+    expiration_date: Mapped[date] = mapped_column(index=True)
+    user_id: Mapped[uuid.UUID] = mapped_column(UUID, ForeignKey(User.user_id))
     user: Mapped[User] = relationship(User)
-
-    agency_id: Mapped[uuid.UUID] = mapped_column(UUID, ForeignKey(Agency.agency_id))
-    agency: Mapped[Agency] = relationship(Agency)
-
-    organization_id: Mapped[uuid.UUID] = mapped_column(
+    agency_id: Mapped[uuid.UUID | None] = mapped_column(UUID, ForeignKey(Agency.agency_id))
+    agency: Mapped[Agency | None] = relationship(Agency)
+    organization_id: Mapped[uuid.UUID | None] = mapped_column(
         UUID, ForeignKey(Organization.organization_id)
     )
-    organization: Mapped[Organization] = relationship(Organization)
+    organization: Mapped[Organization | None] = relationship(Organization)
--- a/api/tests/src/db/models/factories.py
+++ b/api/tests/src/db/models/factories.py
@@ -2930,19 +2930,29 @@ class ExcludedOpportunityReviewFactory(BaseFactory):
     last_update_date = factory.Faker("date_time_between", start_date="-5y", end_date="-3y")
 
 
-class LegacyCertificateFactory(BaseFactory):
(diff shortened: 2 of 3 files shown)


api/src/db/models/user_models.py
```diff
@@ -453,3 +453,25 @@ class UserProfile(ApiSchemaTable, TimestampMixin):
...
+    user_id: Mapped[uuid.UUID] = mapped_column(UUID, ForeignKey(User.user_id), nullable=False)
+    user: Mapped[User] = relationship(User)
+
+    agency_id: Mapped[uuid.UUID] = mapped_column(UUID, ForeignKey(Agency.agency_id))
+    agency: Mapped[Agency] = relationship(Agency)
+
+    organization_id: Mapped[uuid.UUID] = mapped_column(
+        UUID, ForeignKey(Organization.organization_id)
+    )
+    organization: Mapped[Organization] = relationship(Organization)
```
chouinar: SQLAlchemy respects the types you define in the `Mapped` parameter. It'll use that to determine whether a column is nullable. The python type also has a whole thing about how it gets mapped (eg. in Postgres, a python str defaults to the `Text` column type. These are overrideable, and if you look at the `Base` class, that's what the `type_annotation_map` is).

So, we can exclude `nullable=False` if we want. Additionally, because of that, for columns that are nullable, we have to do `| None` otherwise it won't make them nullable. If you look at the migration, you'll notice agency ID is not nullable.

I'd just always exclude the `nullable` from the `mapped_column` and rely on the python typing, at best it's duplicative, at worst it's incorrect.

```suggestion
    cert_id: Mapped[str] = mapped_column(index=True)
    serial_number: Mapped[str] = mapped_column(index=True)
    expiration_date: Mapped[date] = mapped_column(index=True)

    user_id: Mapped[uuid.UUID] = mapped_column(UUID, ForeignKey(User.user_id))
    user: Mapped[User] = relationship(User)

    agency_id: Mapped[uuid.UUID | None] = mapped_column(UUID, ForeignKey(Agency.agency_id))
    agency: Mapped[Agency | None] = relationship(Agency)

    organization_id: Mapped[uuid.UUID | None] = mapped_column(
        UUID, ForeignKey(Organization.organization_id)
    )
    organization: Mapped[Organization | None] = relationship(Organization)
```

api/tests/src/db/models/factories.py
```diff
@@ -2928,3 +2928,21 @@ class Meta:
...
+    serial_number = factory.Sequence(lambda n: n)
+    expiration_date = factory.Faker("future_date", end_date="+2y")
+
+    user_id = factory.LazyAttribute(lambda s: s.user.user_id)
+
+    agency_id = factory.LazyAttribute(lambda a: a.agency.agency_id)
+    agency = factory.SubFactory(AgencyFactory)
+
+    organization_id = factory.LazyAttribute(lambda o: o.organization.organization_id)
+    organization = factory.SubFactory(OrganizationFactory)
```
chouinar, line 2958: What you've defined here is invalid - there won't ever be a cert that has an agency and organization attached. We shouldn't add both.

A few approaches we could take to make a factory for this particular scenario:
* By default, always add one of the two (lets say agency), add a Param like `is_organization` that makes agency null, but adds an org.
* Don't put either of them, make this an abstract factory and make two factories called `LegacyAgencyCertificateFactory` and `LegacyOrganizationCertificateFactory` that each just add these.

api/tests/src/db/models/factories.py
```diff
@@ -2928,3 +2928,21 @@ class Meta:
...
     last_update_date = factory.Faker("date_time_between", start_date="-5y", end_date="-3y")
+
+
+class LegacyCertificateFactory(BaseFactory):
+    class Meta:
+        model = user_models.LegacyCertificate
+
+    legacy_certificate_id = Generators.UuidObj
+    cert_id = factory.Sequence(lambda n: n)
+    serial_number = factory.Sequence(lambda n: n)
```
chouinar: Should cert ID and serial number be unique? I think to have our logic make sense for how we'll use these, we'd want these unique right? Didn't think of that when writing the spec/tickets.

I ask here because in that case, we'd want to avoid a sequence, it'd be better to just make this a random number as otherwise each time we run the factory locally it starts from 1 again.

api/tests/src/db/models/factories.py
jakobpederson: I liked the second option so went with that.

api/src/db/models/user_models.py
jakobpederson: Sounds good, I like that a lot better.

api/tests/src/db/models/factories.py
jakobpederson: Swapped in a uuid instead of a sequence.

PR #6932, opened 2025-11-04
[Issue #6836] Make app attachments be soft-deleted

### LINKED ISSUE #6836: Make attachments soft deleted
### Summary

This will allow us to track attachment deletions/removals in activity history.

TBD

### Acceptance criteria

TBD

chouinar:
## Summary
Fixes #6836

## Changes proposed
* Make application attachments be soft-deleted
* Make it so the `application_attachments` relationship list on an application by default excludes the deleted attachments
* Test that the above relationship change works / is used everywhere

## Context for reviewers
In general, we want to move towards using soft-deletes as it simplifies some data integrity issues in our system. In this case, we want to make it so when we setup logic to add auditing to the application table, if an attachment is deleted, we can continue to track that.

We still want to delete the actual file from s3, if someone deleted an attachment because they accidentally uploaded the wrong file, we don't want to hold onto it.

## Validation steps
Added tests to all places in the app endpoints that work with application attachments to verify that deleted attachments aren't seen by their logic.

### DIFF
--- a/api/src/db/migrations/versions/2025_11_03_add_soft_delete_to_app_attachment.py
+++ b/api/src/db/migrations/versions/2025_11_03_add_soft_delete_to_app_attachment.py
@@ -1,7 +1,7 @@
 """add soft delete to app attachment
 
 Revision ID: cdfb9202130e
-Revises: 43622a46494e
+Revises: 75c42327a370
 Create Date: 2025-11-03 15:11:30.558599
 
 """
@@ -11,7 +11,7 @@ from alembic import op
 
 # revision identifiers, used by Alembic.
 revision = "cdfb9202130e"
-down_revision = "43622a46494e"
+down_revision = "75c42327a370"
 branch_labels = None
 depends_on = None
 


api/src/services/applications/delete_application_attachment.py
```diff
@@ -32,5 +32,9 @@ def delete_application_attachment(
...
     file_util.delete_file(application_attachment.file_location)
 
-    # Delete the attachment from the DB
-    db_session.delete(application_attachment)
+    # Mark the attachment as deleted
+    # We keep the record in the DB for auditing purposes
+    # But still remove the actual file
+    application_attachment.is_deleted = True
+    # Remove the s3 path to make clear it's gone.
+    application_attachment.file_location = "DELETED"
```
mikehgrantsgov, line 40: Do we do anything with file locations where we'd need to filter out attachments where file_location = "DELETED"? I don't necessarily think we should make this column nullable, but wondering if there are downstream implications of having some of our data contain "DELETED" in the column. Maybe there's value in storing the old file location instead of overwriting?

chouinar: The two ways we use the file location that I can think of:
* In various endpoints, we'll convert the file location to a presigned URL. The frontend uses that to make a download link. Whether it's blank/null/"Deleted", those would all not work in the same way.
* In our own logic, using our file utils to interact with the files like the application submission logic. 

In both cases, we will exclude these deleted attachments from the logic, that's what the tests I added were for verifying. I'm open to adjusting exactly what we do with this field, I don't think I want to make it null as that makes using the field clunky.

I could just leave it alone, while it won't point to anything, that's probably fine?

mikehgrantsgov: I think either way is fine. If there is any historical value I'd vote to leave it, but `DELETED` is good too.

mikehgrantsgov (approved):

Issue #6813, opened 2025-10-21
Add a user_type concept to the user table


chris-kuryak:
### Summary

We want to create a new lk_user_type lookup table following our normal lookup table conventions. This table should have the following enum values at this time:

- `standard`
- `internal_frontend`
- `legacy_certificate`

We also want to add a user_type (user_type_id for the actual column name) to the user table. This should be nullable and reference the above table. Default this to `standard` on the column as well.

In the migration we should automatically set ALL user records to have standard for their lookup value except for the following user IDs which should get internal_frontend for their user type.

- dev - `6234b6c2-334e-42c4-8e8b-0c7faba549dd`
- staging - `21b7ac4c-3acb-4975-bd4d-6c76592d8507`
- training - `30bbfd53-f34a-401c-b3b7-a2712e32e072`
- prod - `aa59a18f-86e9-47b2-a80e-8c2a4ddc9287`

You don’t need to have the query figure out the env, just do a where user_id in (…) with all of them, they should be unique across envs.

### Acceptance criteria
- [ ] Lookup table added with new enum
- [ ] Updated user table to have the user type lookup value referenced
- [ ] Manually adjusted the migration to handle setting the user type for all users as described above
- [ ] Adjust the user factory to also default the user type to `standard`

PR #6933, opened 2025-11-04
[Issue #6813] Add lookup user_type

jakobpederson:
## Summary

Fixes / Work for #6813

Ran the migration locally and checked the db
```
app=# SELECT user_id, user_type_id FROM api.user;
               user_id                | user_type_id
--------------------------------------+--------------
 a3f77afe-c293-414b-a2c0-53c1be5f2936 |            1
 f15c7491-7ebc-4f4f-8de6-3ac0594d9c63 |            1
 0f4ae584-c310-472d-9d6c-57201b5f84cc |            1
 b1c2d3e4-f5a6-4b7c-8d9e-0f1a2b3c4d5e |            1
 f5a6b7c8-d9e0-4f1a-2b3c-4d5e6f7a8b9c |            1
 5b4807c5-57d4-4867-b722-1658b47c59ba |            1
 12345678-1234-5678-9abc-123456789abc |            1
 7edb5704-9d3b-4099-9e10-fbb9f2729aff |            1
 6234b6c2-334e-42c4-8e8b-0c7faba549dd |            2  # dev
 21b7ac4c-3acb-4975-bd4d-6c76592d8507 |            2  # staging
 30bbfd53-f34a-401c-b3b7-a2712e32e072 |            2  # training
 aa59a18f-86e9-47b2-a80e-8c2a4ddc9287 |            2  # prod
 ```
Querying the database session it looks like the enums are being set correctly by id:
```
│   User(
│   │   user_id=UUID('7edb5704-9d3b-4099-9e10-fbb9f2729aff'),
│   │   user_type=<UserType.STANDARD: 'standard'>,
│   │   created_at=datetime.datetime(2025, 11, 4, 21, 14, 43, 907167, tzinfo=datetime.timezone.utc),
│   │   updated_at=datetime.datetime(2025, 11, 4, 21, 14, 43, 907167, tzinfo=datetime.timezone.utc)
│   ),
│   User(
│   │   user_id=UUID('6234b6c2-334e-42c4-8e8b-0c7faba549dd'),
│   │   user_type=<UserType.INTERNAL_FRONTEND: 'internal_frontend'>,
│   │   created_at=datetime.datetime(2025, 11, 4, 22, 11, 14, 542228, tzinfo=datetime.timezone.utc),
│   │   updated_at=datetime.datetime(2025, 11, 4, 22, 11, 14, 542228, tzinfo=datetime.timezone.utc)
│   )
```

Can query against the UserType:
```
>>> list(dbs.query(User).where(User.user_type==UserType.INTERNAL_FRONTEND))
[
│   User(
│   │   user_id=UUID('6234b6c2-334e-42c4-8e8b-0c7faba549dd'),
│   │   user_type=<UserType.INTERNAL_FRONTEND: 'internal_frontend'>,
│   │   created_at=datetime.datetime(2025, 11, 4, 22, 11, 14, 542228, tzinfo=datetime.timezone.utc),
│   │   updated_at=datetime.datetime(2025, 11, 4, 22, 11, 14, 542228, tzinfo=datetime.timezone.utc)
│   )
]
```

## Changes proposed

- Add user_type Lookup model
- Add user_type_id to User model
- Update UserFactory
- Manually adjust user_type_id on Users (use list of uuids to set to internal frontend users)

### DIFF
--- a/api/src/db/migrations/versions/2025_11_06_add_lk_user_type_and_add_to_user.py
+++ b/api/src/db/migrations/versions/2025_11_06_add_lk_user_type_and_add_to_user.py
@@ -11,7 +11,7 @@ from alembic import op
 
 # revision identifiers, used by Alembic.
 revision = "618216bb724a"
-down_revision = "cdfb9202130e"
+down_revision = "878613347a9c"
 branch_labels = None
 depends_on = None
 

### CODE REVIEW

File: api/src/db/models/user_models.py [hunk 1]
```diff
@@ -68,6 +68,10 @@ class User(ApiSchemaTable, TimestampMixin):
     profile: Mapped["UserProfile | None"] = relationship(
         "UserProfile", back_populates="user", uselist=False, cascade="all, delete-orphan"
     )
+    user_type_id: Mapped[int | None] = mapped_column(
+        ForeignKey(LkUserType.user_type_id),
+        default=1,
+    )
```
chouinar: I saw on prior versions you had this with the enum, which is what we'd want to do. We don't want  lookup values to be used directly as integers as it becomes very difficult to work with.

Was there an issue you were running into that prevents this from being `user_type: Mapped[UserType | None]` ?

For context, what the lookup column approach does is let us do something like:

```py
db_session.execute(select(Opportunity).where(Opportunity.opportunity_category == OpportunityCategory.DISCRETIONARY))
```
It handles converting enums to the underlying integers seamlessly.

api/src/db/migrations/versions/2025_11_04_add_lk_user_type_to_user.py
```diff
@@ -0,0 +1,67 @@
...
+                WHEN user_id = ANY(:internal_user_list) THEN :internal_frontend_type_id
+                ELSE :standard_type_id
+            END;
+            """
+        ).bindparams(
+            bindparam("internal_user_list", value=INTERNAL_USER_UUIDS, type_=ARRAY(UUID)),
+            bindparam("internal_frontend_type_id", INTERNAL_FRONTEND_TYPE_ID),
+            bindparam("standard_type_id", STANDARD_TYPE_ID),
+        )
+    )
```
chouinar: Have you tested that this works locally via a migration? We'd want to be pretty thorough for any custom code we're doing locally in migrations. If so, could you add how it behaved to the PR summary?

api/src/db/models/user_models.py
jakobpederson: I did run into some issues but I was doing the foreign key wrong.

api/src/db/migrations/versions/2025_11_04_add_lk_user_type_to_user.py
jakobpederson: Added above.

api/tests/src/db/models/factories.py
```diff
@@ -896,6 +897,7 @@ class Meta:
         model = user_models.User
 
     user_id = Generators.UuidObj
+    user_type = factory.LazyFunction(lambda: UserType.STANDARD)
```
chouinar: Should be able to just directly assign it.
```suggestion
    user_type = UserType.STANDARD
```

api/src/db/migrations/versions/2025_11_06_add_lk_user_type_and_add_to_user.py
chouinar, line 1: Could this be one migration file?

jakobpederson: I'm not sure, it seems like when I put it in one file I run into issues where the id numbers don't seem to exists for the user_type_ids when I'm trying assign them with the SQL query command. It seems like running that part first and separately fixes that issue.

chouinar: If you run both of these migrations together (eg. run `make db-migrate-down` twice and then run `make db-migrate`), does that fail? I think it would.

Here's the issue you're hitting, when we run the DB migrations, the process first runs any outstanding DB migrations and then it runs the script that loads lookup values to the tables.

Even if these are in two migrations, if we merged this to dev/staging/etc. it would fail because the migrations would run before the lookup values exist.

This was a nice-to-have, if we really want it, we'd need to manually do that backfill I think.

jakobpederson: Yeah it wasn't working when I did `make db-migration down` twice and then ran `make db-migrate`. I'll pull it out.

jakobpederson: We could definitely do it in Django I think, so a good thing to learn about Flask.

chouinar: Note - Flask isn't involved in our migrations at all. Unlike Django, or Flask-SQLAlchemy (which we don't use, despite using the two parts of it), our API and DB integrations are completely independent. The migrations are via Alembic, but the way our lookup values work are purely a custom approach I built.

jakobpederson: How do we go about running this manually? Is it something I'll add a command for?

chouinar: We'd login to the database and run the SQL command to do the update.

api/src/db/migrations/versions/2025_11_04_add_lk_user_type_to_user.py
```diff
@@ -0,0 +1,40 @@
...
+Revises: 9f919a9e045e
+Create Date: 2025-11-04 18:43:34.137899
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "20eb0b848e1d"
```
chouinar: Could we merge these two migrations now that we removed the SQL command? Should work fine now.

A different PR merged a DB change right as you sent the latest commit, so you'll need to pull from main and fix the migrations regardless.

PR #6954, opened 2025-11-04
[Issue #6834] Modify application submission logic to put XML file in the Zip

### LINKED ISSUE #6834: Modify application submission logic to put XML file in the Zip
### Summary
Following up from #6657 

- If an application has a form we don't support, skip generating XML altogether and do not add to Zip

- Filename needs to be exactly `GrantApplication.xml`
- Always add XML to the Zip (for now, until we hear otherwise from Grantor Agencies)
- Header, forms, and footer need to be constructed in the right way
- Plug this logic into the submission logic

(Michael to provide example of what this all looks like)

### Acceptance criteria

- [ ] XML file is created and added to Zip file per above criteria
- [ ] Create feature flag that skips the process of XML generation that can be turned on/off in Prod
- [ ] Tests created

mikehgrantsgov:
## Summary

Fixes #6834

## Changes proposed

Add `enable_xml_generation` to `ApplicationSubmissionConfig`.
Add `SubmissionXMLAssembler` to handle assembling application submission forms, header and footer into a `GrantApplication.xml` file, included in the zip file.
When FF is enabled, `SubmissionXMLAssembler` will generate XML and include in the ZIP file.
Add relevant tests to the task and assembler.
Update the submission manifest to reflect `GrantApplication.xml` is added to the zip.

## Context for reviewers

We moved up creating the `ApplicationSubmission` record earlier in the `create_application_submission_task` because we need the `legacy_tracking_number` in the XML footer, so we update the record after creating it and the zip file.
The process is designed to *not* block the task if there was an error generating the XML, so any forms that are not supported or fail are skipped.

## Validation steps

See unit tests for task and assembler.

### DIFF
--- a/api/src/task/apply/create_application_submission_task.py
+++ b/api/src/task/apply/create_application_submission_task.py
@@ -235,8 +235,8 @@ class CreateApplicationSubmissionTask(Task):
         # Get the tracking number from the sequence before creating the record
         tracking_number = self.db_session.scalar(ApplicationSubmission.legacy_tracking_number_seq)
 
-        # Create a temporary submission object for XML generation (not yet persisted)
-        temp_submission = ApplicationSubmission(
+        # Create the submission object (file size will be updated after zip creation)
+        application_submission = ApplicationSubmission(
             application_submission_id=submission_id,
             application=application,
             file_location=s3_path,
@@ -248,7 +248,7 @@ class CreateApplicationSubmissionTask(Task):
             with zipfile.ZipFile(outfile, "w") as submission_zip:
 
                 submission_container = SubmissionContainer(
-                    application, submission_zip, temp_submission
+                    application, submission_zip, application_submission
                 )
 
                 self.process_application_forms(submission_container)
@@ -256,17 +256,8 @@ class CreateApplicationSubmissionTask(Task):
                 self.process_xml_generation(submission_container)
                 self.create_manifest_file(submission_container)
 
-        # Get the size of the zip from s3
-        zip_length = file_util.get_file_length_bytes(s3_path)
-
-        # Now create the final submission record with all correct data
-        application_submission = ApplicationSubmission(
-            application_submission_id=submission_id,
-            application=application,
-            file_location=s3_path,
-            file_size_bytes=zip_length,
-            legacy_tracking_number=tracking_number,
-        )
+        # Update the file size now that the zip is complete
+        application_submission.file_size_bytes = file_util.get_file_length_bytes(s3_path)
         self.db_session.add(application_submission)
 
         # Mark the app as accepted
@@ -382,7 +373,7 @@ class CreateApplicationSubmissionTask(Task):
             submission.application, submission.application_submission, attachment_mapping
         )
 
-        xml_content = xml_assembler.generate_complete_submission_xml(pretty_print=True)
+        xml_content = xml_assembler.generate_complete_submission_xml()
 
         if not xml_content:
             logger.warning(


api/src/task/apply/create_application_submission_task.py
```diff
@@ -325,6 +339,50 @@ def process_application_attachments(self, submission: SubmissionContainer) -> No
...
+                    "application_id": submission.application.application_id,
+                    "competition_id": submission.application.competition_id,
+                },
+            )
+            return
+
+        log_extra = {
+            "application_id": submission.application.application_id,
+            "competition_id": submission.application.competition_id,
+        }
```
chouinar:
```suggestion
        log_extra = {
            "application_id": submission.application.application_id,
            "competition_id": submission.application.competition_id,
        }
        if not self.app_submission_config.enable_xml_generation:
            logger.info(
                "Skipping XML generation - feature flag disabled",
                extra=log_extra,
            )
            return


```

api/src/task/apply/create_application_submission_task.py
```diff
@@ -387,6 +445,9 @@ def create_manifest_text(submission: SubmissionContainer) -> str:
         Attachments included in ZIP (total #)
         1. my-attachment.txt (size X)
         2. another-attachment.docx (size X)
+
+        XML Files included in ZIP (total #)
+        1. GrantApplication.xml (size X)
```
chouinar: I checked what the manifest looks like when it has an xml file. Looks like it adds a whole section like so:

```
Manifest for Grant Application # GRANT00838603

Grant Application XML file (total 1): 
 1. GrantApplication.xml. (size 13390 bytes)

Forms Included in Zip File(total 2):
 1. Form SFLLL_2_0-V2.0.pdf (size 20927 bytes)
 2. Form SF424_Short_3_0-V3.0.pdf (size 21985 bytes)

Attachments Included in Zip File (total 0):
```

Could we add that (or make a follow-up ticket to do that?)

api/src/services/xml_generation/config.py
```diff
@@ -3,23 +3,29 @@
...
     try:
-        if form_name.upper() == "SF424_4_0":
-            from src.form_schema.forms.sf424 import FORM_XML_TRANSFORM_RULES
+        form_name_upper = form_name.upper()
+
+        if form_name_upper == "SF424_4_0":
 
-            logger.info(f"Loaded transformation config for {form_name} from sf424.py")
+            logger.info(f"Loaded transformation config for {form_name}")
             return FORM_XML_TRANSFORM_RULES
```
chouinar, line 19: Is there a way we could have this logic, at least for the submission XML work, just check whether `form.json_to_xml_schema` is not null / get the value from that?

We actually would have issues with this approach - we don't always have the latest form in a given environment, if we had added some recent transform rules for a form, but weren't ready for it to be in prod, we'd not want to import the value like you've done here as it wouldn't be the one deployed to the env.

The underlying XML logic doesn't need to know what a form is, but we need some point where we can pass in a config from the form DB object.

api/src/services/xml_generation/submission_xml_assembler.py
```diff
@@ -0,0 +1,209 @@
...
+
+        Returns:
+            List of tuples (application_form, form_name) for supported forms only
+        """
+        supported_forms = []
+
+        for app_form in self.application.application_forms:
+            form_name = app_form.form.short_form_name
+            if self.is_form_xml_supported(form_name):
+                supported_forms.append((app_form, form_name))
```
chouinar: Instead of cross-referencing a config, can we just do this and then pipe through this json_to_xml_schema?

```suggestion
        for app_form in self.application.application_forms:
            form_name = app_form.form.short_form_name
            if app_form.form.json_to_xml_schema is not None:
                supported_forms.append((app_form, form_name))
```

api/src/services/xml_generation/submission_xml_assembler.py
```diff
@@ -0,0 +1,209 @@
...
+        )
+
+        request = XMLGenerationRequest(
+            application_data=app_form.application_response,
+            form_name=form_name,
+            pretty_print=False,
+            attachment_mapping=attachment_mapping,
+        )
+
+        response = self.xml_service.generate_xml(request)
```
chouinar, line 144: Could we change the XMLGenerationRequest to swap out the `form_name` for the `json_to_xml_schema` dict? From looking at the code, the first thing we use the form name for is getting that schema in a different way, this would just be doing it a bit more directly and removing the need for the underlying code to even know what a form is.

api/src/task/apply/create_application_submission_task.py
```diff
@@ -224,26 +229,35 @@ def process_application(self, application: Application) -> None:
...
+        # Create the submission record now so we can get the tracking number. We'll update the size later.
+        application_submission = ApplicationSubmission(
+            application_submission_id=submission_id,
+            application=application,
+            file_location=s3_path,
+            file_size_bytes=0,  # Placeholder, will update after zip is created
+        )
+        self.db_session.add(application_submission)
+        # Flush so we can get the tracking number
+        self.db_session.flush()
```
chouinar: If we want to get the value of the sequence without needing to rearrange things and flush, surprisingly this works:

```py
tracking_number = db_session.execute(ApplicationSubmission.legacy_tracking_number_seq)
application_submission = ApplicationSubmission(
   ...
   legacy_tracking_number=tracking_number
)
```

api/src/services/xml_generation/submission_xml_assembler.py
```diff
@@ -0,0 +1,209 @@
...
+    def _parse_xml_string(self, xml_string: str) -> lxml_etree.Element:
+        """Parse XML string into element tree."""
+        try:
+            # Parse the XML string
+            parser = lxml_etree.XMLParser(remove_blank_text=True)
+            root = lxml_etree.fromstring(xml_string.encode("utf-8"), parser=parser)
+            return root
+        except Exception as e:
+            logger.error(f"Failed to parse XML string: {e}")
+            raise ValueError(f"Invalid XML string: {e}") from e
```
chouinar: Could this error contain the contents of the XML? We'd have to be careful about logging as these forms can contain a lot of PII.

mikehgrantsgov: Made this change.

api/src/task/apply/create_application_submission_task.py
mikehgrantsgov: Good call, I updated this too.

api/src/services/xml_generation/submission_xml_assembler.py
mikehgrantsgov: Updated this and tests associated with this behavior

api/src/task/apply/create_application_submission_task.py
mikehgrantsgov: Updated to manifest to match this in this PR.

api/src/services/xml_generation/config.py
mikehgrantsgov: I updated this so the XML generation now reads form.json_to_xml_schema directly from the database

api/src/services/xml_generation/submission_xml_assembler.py
```diff
@@ -0,0 +1,204 @@
...
+
+class SubmissionXMLAssembler:
+    """Assembles complete GrantApplication.xml with header, forms, and footer."""
+
+    def __init__(self, application: Application, application_submission: ApplicationSubmission):
+        self.application = application
+        self.application_submission = application_submission
+        self.xml_service = XMLGenerationService()
+
+    def get_supported_forms(self) -> list[tuple[Any, str]]:
```
chouinar: Since we return the application form objects, could we adjust this type, I think it'd simplify a few other places that then use this list.

Can keep the form name, but it's also something we can get from the app form so, could remove the tuple potentially
```suggestion
    def get_supported_forms(self) -> list[ApplicationForm]:
```

api/src/services/xml_generation/submission_xml_assembler.py
```diff
@@ -0,0 +1,204 @@
...
+            form_name = app_form.form.short_form_name
+            if app_form.form.json_to_xml_schema is not None:
+                supported_forms.append((app_form, form_name))
+            else:
+                logger.info(
+                    f"Skipping form {form_name} for XML generation - no transform config",
+                    extra={
+                        "application_id": self.application.application_id,
+                        "form_name": form_name,
+                    },
```
chouinar, line 52:
```suggestion
                    extra={
                        "application_id": self.application.application_id,
                        "form_id": app_form.form.form_id,
                        "form_name": form_name,
                    },
```

api/src/services/xml_generation/submission_xml_assembler.py
```diff
@@ -0,0 +1,204 @@
...
+        )
+
+        return complete_xml
+
+    def _generate_form_xml(self, app_form: Any, form_name: str) -> str:
+        """Generate XML for a single application form."""
+        # Build attachment mapping for this form's attachments
+        attachment_mapping = create_attachment_mapping_from_list(
+            self.application.application_attachments
+        )
```
chouinar: Could we pass in the application attachment mapping we have from the application submission logic? We'd want that for the overrides in this function. That said, we don't really store a mapping right now, so if we want to make a follow-up ticket to pipe that through from where we calculate it in `process_application_attachments`, I'm fine with that.

I think we'd also want to generate this mapping only once rather than for each form since it's application wide.

api/src/services/xml_generation/submission_xml_assembler.py
```diff
@@ -0,0 +1,204 @@
...
+                    f"Skipping form {form_name} for XML generation - no transform config",
+                    extra={
+                        "application_id": self.application.application_id,
+                        "form_name": form_name,
+                    },
+                )
+
+        return supported_forms
+
+    def generate_complete_submission_xml(self, pretty_print: bool = True) -> str | None:
```
chouinar: For the XML we put in the file, I'm not sure whether the default should be pretty or not. The current system seems to be halfway between the two:

Here's a rough screenshot of what it looks like:
Image

The header is across multiple lines, then each form is a single line.

I have no idea if this matters, I would hope no ones parsing cares, but something we should keep in mind.

api/src/task/apply/create_application_submission_task.py
```diff
@@ -224,26 +229,45 @@ def process_application(self, application: Application) -> None:
 
         submission_id = uuid.uuid4()
         s3_path = build_s3_application_submission_path(self.s3_config, application, submission_id)
+
+        # Get the tracking number from the sequence before creating the record
+        tracking_number = self.db_session.scalar(ApplicationSubmission.legacy_tracking_number_seq)
+
+        # Create a temporary submission object for XML generation (not yet persisted)
+        temp_submission = ApplicationSubmission(
```
chouinar: Do we want both this and the object we create below? If we need to create the object here for things to work, that's fine, but might as well just fully make it and not have the duplication of setting it up, since I think just the file size would change.

mikehgrantsgov: I consolidated this into one object

api/src/services/xml_generation/submission_xml_assembler.py
mikehgrantsgov: Makes sense. I made `pretty_print` default to `False` and we can use True when it's needed for debugging etc.

api/src/services/xml_generation/submission_xml_assembler.py
mikehgrantsgov: Good idea, I updated this.

Issue #6600, opened 2025-10-03
Develop longer term design for directing users to existing applications


chris-kuryak:
### Summary

For the longer term solution, we want the Opportunity Page to alert/show the user if they have any In-Progress or Submitted applications for the opportunity and how to access them. This should not deny them the ability to start a new application as well.

### Acceptance criteria

- [ ] Content/Design created for and longer term solution linked in this ticket
- [ ] Work with Product to create follow-on engineering implementation tickets for longer term solution

andycochran:
Thanks for including "should not deny them the ability to start a new app" as some users might help multiple organizations with apps for the same opportunity.

Issue #6832, opened 2025-10-23
Add pagination to Application List view page


chris-kuryak:
### Summary

Application List view page (/applications) currently returns a full list of applications. For users that have a lot of applications, we want to implement pagination to the table.

Pagination is now supported by the backend team via ticket #6124 . 
This work is for frontend team to support that pagination on this table.

[Figma design](https://www.figma.com/design/zPlDWcYM396szw2yL3UXvk/Simpler-Grants-2025?node-id=4978-47646&m=dev) for reference.

Pagination component exists in other tables.

### Acceptance criteria
- [ ] Table displays up to 10 records in the table
- [ ] Pagination added per Figma design
- [ ] Product UAT passed
- [ ] Design UAT passed

Issue #6463, opened 2025-09-19
Create instructions in New Application start modal how users can access existing applications


chris-kuryak:
### Summary

When a user starts a new application, we want to let them know how they can access any existing applications.

Product Guidance:
- Since we will be improving the experience more holistically later this quad (with an Application List feature), please time-box this improvement to 1-2 points of effort, but is sufficient to meet the user need for the near term.

From Lucas 9/17:

> If I've already started an application, but I'm back on the opportunity page, the only option is to "Start new application", there's no option to resume the application I was working on (or even see that I was working on one). I assume this relates to https://github.com/HHS/simpler-grants-gov/issues/5476 but I'm not actually seeing a ticket that directly addresses this. For the pilot, are we just assuming people know to bookmark their page (as the header suggests)? Might be worth triple-emphasizing this in instructional materials because otherwise they're locked out until they contact our support email. I am guessing this will come up during the pilot.

### Acceptance criteria

- [x] Design completed 
- [x] Design includes a link to the user's Applications List view
- [x] Aligned with product and engineering
- [ ] Work with Product to create follow-on implementation ticket

lucasmbrown-usds:
Adding this comment that was [also on Slack](https://betagrantsgov.slack.com/archives/C08NCEXRSPJ/p1758641032010899?thread_ts=1758145838.242109&cid=C08NCEXRSPJ): 

> Should we add some text here as a placeholder that instructs people who might be confused about where their existing funding is? Depends how long we think it'll take to get the links deployed with access to the old applications.
We could say something like, "Looking for your existing application? Currently, existing applications are only accessible through a direct link. You may be able to find this link by searching in the history of the browser you used to create the application, by searching for "[simpler.grants.gov/workspace/applications](http://simpler.grants.gov/workspace/applications)". Or, you can contact us for support at [simpler@grants.gov](mailto:simpler@grants.gov)."

crystabelrangel:
Here is the link to the figma file: https://www.figma.com/design/zPlDWcYM396szw2yL3UXvk/Simpler-Grants-2025?node-id=5421-60813&t=opAM6PJPPpxc79S1-1

fyi: @chris-kuryak @doug-s-nava @chelsea-nava

ErinPattisonNava:
@chris-kuryak When we build out the implementation ticket, will it also include removing the banner telling users to bookmark the application page?  cc @doug-s-nava

chris-kuryak:
[...]

PR #6962, opened 2025-11-05
feat: the Applications page with data

ErinPattisonNava:
## Summary

Work for #6597 

## Changes proposed

A table was added to the Applications page including ability to fetch the /{user_id}/applications endpoint.  Figma designs are found [here](https://www.figma.com/design/zPlDWcYM396szw2yL3UXvk/Simpler-Grants-2025?node-id=6147-97509).  

Testing coverage for touched components:

Screenshot 2025-11-05 at 1 46 10 PM

## Context for reviewers

This pr builds on #6907 to flesh out the Applications page with data.   The feature has already completed design review. 

One note is that when you go to an Opportunity where you've already created an Application, it does not remember you have already begun and application.  So you'll see the "start an application" button and you can have multiple applications opened as an individual against a single opportunity.  Looks like that will be fixed in #6600 

## Validation steps

- [ ] A logged out user should get an error message that they need to sign in
- [ ] A logged in user who has started no applications should see the appropriate content
- [ ] Easiest to test seeing applications in localhost with the `many_app_user` user
  - [ ] Close date should be formatted
  - [ ] Status for applications in draft should say "Draft"
  - [ ] Status for applications in submitted should say "Submitted"
  - [ ] Status for applications in approved should say "Submitted"
  - [ ] Application name should be a link to the already-started application
  - [ ] Type should read as Individual for applications started as an individual
  - [ ] Type should read as the organization name for applications started as an organization
  - [ ] Opportunity name should be listed with a link, and Agency should be listed as well
  - [ ] If an application is opened in a separate browser and this page is refreshed, it should appear (no caching issues)  
- [ ] Error can be tested in localhost only, I put ` raise_flask_error(500, f"EJP error!!”)` [here](https://github.com/HHS/simpler-grants-gov/blob/2ccb772aa0930263173046f17708e1e7571e8f6f/api/src/services/users/get_user_applications.py#L66) to force a 500

### DIFF
--- a/frontend/src/app/[locale]/(base)/applications/page.tsx
+++ b/frontend/src/app/[locale]/(base)/applications/page.tsx
@@ -122,14 +122,7 @@ const ApplicationsTable = ({
 
       <TableWithResponsiveHeader
         headerContent={headerTitles}
-        tableRowData={transformTableRowData(
-          userApplications.sort(
-            (a: ApplicationDetail, b: ApplicationDetail) =>
-              new Date(b.competition.closing_date).getTime() -
-              new Date(a.competition.closing_date).getTime(),
-          ),
-          t,
-        )}
+        tableRowData={transformTableRowData(userApplications, t)}
       />
     </div>
   );
--- a/frontend/src/services/fetch/fetchers/applicationsFetcher.ts
+++ b/frontend/src/services/fetch/fetchers/applicationsFetcher.ts
@@ -16,6 +16,12 @@ export const getApplications = async (
     pagination: {
       page_offset: 1,
       page_size: 5000,
+      sort_order: [
+        {
+          order_by: "created_at",
+          sort_direction: "descending",
+        },
+      ],
     },
   };
   const subPath = `${userId}/applications`;
--- a/frontend/tests/services/fetch/fetchers/applicationsFetcher.test.ts
+++ b/frontend/tests/services/fetch/fetchers/applicationsFetcher.test.ts
@@ -37,6 +37,12 @@ describe("getApplications", () => {
         pagination: {
           page_offset: 1,
           page_size: 5000,
+          sort_order: [
+            {
+              order_by: "created_at",
+              sort_direction: "descending",
+            },
+          ],
         },
       },
     });
@@ -65,6 +71,12 @@ describe("fetchApplications", () => {
         pagination: {
           page_offset: 1,
           page_size: 5000,
+          sort_order: [
+            {
+              order_by: "created_at",
+              sort_direction: "descending",
+            },
+          ],
         },
       },
     });


frontend/src/services/fetch/fetchers/applicationsFetcher.ts
```diff
@@ -0,0 +1,34 @@
...
+  });
+  const json = (await resp.json()) as { data: [] };
+  return json.data;
+};
+
+export const fetchApplications = async (): Promise<ApplicationDetail[]> => {
+  const session = await getSession();
+  if (!session || !session.token) {
+    // we shouldn't get there because the page should be checking authentication
+    throw new UnauthorizedError("No active session");
```
ErinPattisonNava, line 41: I've seen other cases return an empty array if the session isn't valid, but in our case an empty array triggers content about how they haven't started applications.  Seemed safest to throw an error and let the top level infrastructure catch it if the page authentication doesn't.

frontend/tests/pages/applications/page.test.tsx
```diff
@@ -20,26 +23,216 @@ jest.mock("next-intl", () => ({
...
+    });
+
+    it("renders headings", async () => {
+      applications.mockResolvedValue([basicApplication]);
+      const component = await Applications({ params: localeParams });
+      render(component);
+
+      expect(
+        screen.getAllByText("Applications.tableHeadings.closeDate"),
+      ).toHaveLength(2);
```
ErinPattisonNava, line 127: In a fascinating twist, our standard table includes the column name as an element for each row but with css to make sure it isn't visible (I'm guessing for a11y concerns?).  But react testing library isn't smart enough to realize that it won't be visible, so it "appears" twice.

frontend/src/types/competitionsResponseTypes.ts
```diff
@@ -24,4 +27,5 @@ export type Competition = {
   opening_date: string;
   opportunity_assistance_listings: OpportunityAssistanceListing[];
   opportunity_id: number;
+  opportunity: BaseOpportunity;
```
ErinPattisonNava, line 30: I need the `agency_name`, which is why I added BaseOpportunity instead of MinimalOpportunity.

After some thought, I think I'm leaning towards viewing these types as a vague superset of what can be returned by different APIs, rather than a tight type definition.  So the opportunity may be available in some API calls and not in others, and they only way you'd know is by looking up the endpoint information in swagger.  That gives us the most flexibility, minimizes proliferation of similarly-named types, and creates the least amount of work.  But open to thoughts!  

Long term if we do want to go with accurate type definitions, doing something that lets the backend and the frontend share a definition to auto-generate the types seems like the right way to go, but not sure if we need that/priority of it.

frontend/src/services/fetch/fetchers/applicationsFetcher.ts
```diff
@@ -0,0 +1,39 @@
...
+  token: string,
+  userId: string,
+): Promise<ApplicationDetail[]> => {
+  const ssgToken = {
+    "X-SGG-Token": token,
+  };
+  const body = {
+    pagination: {
+      page_offset: 1,
+      page_size: 5000,
```
ErinPattisonNava, line 18: Will be fixed in https://github.com/HHS/simpler-grants-gov/issues/6832

frontend/src/app/[locale]/(base)/applications/page.tsx
```diff
@@ -1,35 +1,155 @@
...
-}
+import {
+  TableCellData,
+  TableWithResponsiveHeader,
+} from "src/components/TableWithResponsiveHeader";
 
-export default function Applications({ params }: LocalizedPageProps) {
-  const { locale } = use(params);
-  setRequestLocale(locale);
+const PageLayout: FunctionComponent<PropsWithChildren> = ({ children }) => {
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/applications/page.tsx]: the pattern we're generally following is to type the props for components rather than the function. In this case I think we'd want something like
```suggestion
const PageLayout = ({ children }: PropsWithChildren) => {
```

Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/applications/page.tsx]: since `layout` is sort of an overloaded term in Nextjs, can we rename this to avoid confusion? Maybe `ApplicationsPageWrapper`?

frontend/src/app/[locale]/(base)/applications/page.tsx
```diff
@@ -1,35 +1,155 @@
...
   return (
     <GridContainer>
       <h1 className="margin-top-9 margin-bottom-7">{t("pageTitle")}</h1>
+      {children}
+    </GridContainer>
+  );
+};
+
+const NoStartedApplications = () => {
+  const t = useTranslations("Applications");
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/applications/page.tsx]: [nit] since this component is only using strings namespaced below "Application.noApplicationsMessage" you can save a bit of real estate by calling `useTranslations("Applications.noApplicationMessage")` and referencing only "primary" and "secondary" below

frontend/src/app/[locale]/(base)/applications/page.tsx
```diff
@@ -1,35 +1,155 @@
...
+        tableRowData={transformTableRowData(userApplications, t)}
+      />
+    </div>
+  );
+};
+
+export default async function Applications({ params }: LocalizedPageProps) {
+  const { locale } = await params;
+  setRequestLocale(locale);
+  await getTranslations({ locale });
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/applications/page.tsx]: since this component isn't using any translations I think you can safely cut out 132 - 134

frontend/src/app/[locale]/(base)/applications/page.tsx
```diff
@@ -1,35 +1,155 @@
...
+        </Alert>
       </div>
-    </GridContainer>
+    </PageLayout>
+  );
+};
+
+const transformTableRowData = (
+  userApplications: ApplicationDetail[],
+  t: TFn,
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/applications/page.tsx:56]: [something to consider] normally I like the idea of passing around functions, but the "t" function kinda weirds me out, as it's really only meant to be used inside a component function, and passing it around seems like breaking a rule, even if it works.

Working around this limitation we usually will pass in the necessary translated strings as individual props instead. 

🤷 not a big deal

frontend/tests/pages/applications/page.test.tsx
```diff
@@ -20,26 +23,216 @@ jest.mock("next-intl", () => ({
...
+    });
   });
 
-  it("passes accessibility scan", async () => {
-    const { container } = render(<Applications params={localeParams} />);
-    const results = await waitFor(() => axe(container));
+  describe("an application is returned", () => {
+    let basicApplication: DeepPartial<ApplicationDetail>;
+    beforeEach(() => {
+      basicApplication = {
```
doug-s-nava, line 92: it looks like we don't have an example of `ApplicationDetail` in our fixtures - is this a good example to add?

frontend/tests/pages/applications/page.test.tsx
doug-s-nava: this is to handle how the mobile version works - it hides at non-mobile sizes. I'm fine with the test case being kind of wonky as a result, since I've also encountered testing library's extreme lack of awareness of when something is visible

frontend/tests/pages/applications/page.test.tsx
```diff
@@ -20,26 +23,216 @@ jest.mock("next-intl", () => ({
...
+
+    it("renders close date", async () => {
+      applications.mockResolvedValue([basicApplication]);
+      const component = await Applications({ params: localeParams });
+      render(component);
+
+      expect(await screen.findByText("November 11, 2025")).toBeVisible();
+    });
+
+    it("renders applicaiton filing name", async () => {
```
doug-s-nava: [sp] 
```suggestion
    it("renders application filing name", async () => {
```

frontend/tests/pages/applications/page.test.tsx
```diff
@@ -20,26 +23,216 @@ jest.mock("next-intl", () => ({
...
+        applications.mockResolvedValue([basicApplication]);
+        const component = await Applications({ params: localeParams });
+        render(component);
+
+        expect(
+          await screen.findByText("Applications.tableContents.individual"),
+        ).toBeVisible();
+      });
+
+      it("if started as an orgnaization", async () => {
```
doug-s-nava: [sp] 
```suggestion
      it("if started as an organization", async () => {
```

frontend/src/app/[locale]/(base)/applications/page.tsx
Comment (ErinPattisonNava) [reply frontend/src/app/[locale]/(base)/applications/page.tsx]: Good call on removing `getTranslations`.  I thought that every server component had to call `setRequestLocale`?

frontend/src/app/[locale]/(base)/applications/page.tsx
Comment (ErinPattisonNava) [reply frontend/src/app/[locale]/(base)/applications/page.tsx:56]: Hrm.  Good food for thought, although I do prefer grabbing the strings where they are needed instead of passing them in for a helper function.  If you are cool I'll leave it like this for now but keep it in mind for the future.

ErinPattisonNava:
@doug-s-nava 

> * with this work being done, is there a ticket to remove the notification on the application page about bookmarking the link?

I think it makes the most sense to include it in https://github.com/HHS/simpler-grants-gov/issues/6463, will add a comment there.  

> * the designs seems to suggest that long organization and opportunity names should be truncated, but I don't think that was discussed as a requirement. If that's necessary let's do it as a fast follow

Checked with design, they are good with it as is.

> * what is the default sort here, and what should it be? In my testing it looks like it's not really sorted in any logical way. Seems in scope to make sure things have some discernible default sort

Good call out.  This will get overridden by pagination, but in the short term I'll sort them by close date.  (Looking at how pagination was done I don't think it supports sorting by the fields we need, so I'll do it in the front end for the time being.)

frontend/tests/pages/applications/page.test.tsx
ErinPattisonNava: Not really, because it has a smaller subset of the data that the ApplicationDetail could possibly return.  I played around some with trying to grab a better one/compose one out of already mocked objects, but in each case I ran into missing fields for the type.  I can create a huge one out of whole cloth (with data I may not be sure would be in possible combination), or I can leave this as a representative of this specific API call and add an Application mock the next time we add a test to the Application page.

doug-s-nava: I'm good with this. If we find we need something like this again in the future, let's figure out a way to set it as a fixture that makes sense

frontend/src/app/[locale]/(base)/applications/page.tsx
Comment (doug-s-nava) [reply frontend/src/app/[locale]/(base)/applications/page.tsx]: if we were actually going to be translating anything I would say maybe keep it but as things are I'd rather take that out too

frontend/src/app/[locale]/(base)/applications/page.tsx
```diff
@@ -1,35 +1,161 @@
...
+      <span className="font-sans-lg text-bold">
+        {t("Applications.numApplications", { num: userApplications.length })}
+      </span>
+
+      <TableWithResponsiveHeader
+        headerContent={headerTitles}
+        tableRowData={transformTableRowData(
+          userApplications.sort(
+            (a: ApplicationDetail, b: ApplicationDetail) =>
+              new Date(b.competition.closing_date).getTime() -
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/applications/page.tsx]: [not blocking] if a default sort hasn't been decided along with the pagination work we should make sure that's decided - what's here looks like "close date - farthest to soonest", and I think it would make more sense to flip that so that the soonest closing date floats to the top

Comment (ErinPattisonNava) [reply frontend/src/app/[locale]/(base)/applications/page.tsx]: Per discussion we decided to remove the frontend sorting and the [backend sorting](https://github.com/HHS/simpler-grants-gov/pull/7002) has been changed to be created_at so at least the list stays stable.  We'll revisit backend sorting in #7006 before we tackle frontend sorting with pagination.  It means the list still looks a little odd because the dates aren't ordered, but hopefully it will be most useful to see their newest applications at the top.

doug-s-nava (approved):
✅ ✅ ✅ ✅

Issue #6744, opened 2025-10-15
Table to modify roles in user management


eric-serrato95:
### Summary

Capturing the effort to put together the table for active users and to modify current roles for organization users 

### Acceptance criteria

- [ ] Active users table is created 
- [ ] 3 columns are available - Email - Name - Role
- [ ] IF user has not completed their First and Last name Then Name should appear blank 
- [ ] Role provides a dropdown menu to select role for a user

eric-serrato95:
@myduong-navapbc moving to done, UAT ✅

PR #6974, opened 2025-11-06
[ISSUE #6744]: Manage users table

myduong-navapbc:
## Summary

Work for #6744 

## Changes proposed

- This PR adds the routes and handlers for getting and updating the users from manage users page
- Created the Table
- Modal to verify change

## Validation steps

- to run locally, you would need to `run npm run build && npm start
- navigate to feature flags (http://localhost:3000/dev/feature-flags)
- make sure that `userAdminOff` is set to `disabled`
- sign into the app as "two_org_user"
- navigate to Workspace > Activity Dashboard
- scroll to the bottom and "View Organization Details"
- click "Manage Users"
- VERIFY that you see the table
- VERIFY that you can trigger the modal upon change of role
- VERIFY that the role updates upon change


frontend/src/app/[locale]/(base)/organization/[id]/manage-users/page.tsx
```diff
@@ -14,19 +19,49 @@ export async function generateMetadata({
   params,
 }: {
   params: Promise<{ locale: string; id: string }>;
-}) {
-  const { locale } = await params;
+}): Promise<Metadata> {
+  const { locale, id } = await params;
   const t = await getTranslations({ locale });
-
+  let title = `${t("ManageUsers.pageTitle")}`;
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/organization/[id]/manage-users/page.tsx:25]: doesn't seem like this needs to be in a template

frontend/src/app/api/user/organizations/[organizationId]/users/[userId]/handler.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  const { organizationId, userId } = await options.params;
+
+  try {
+    const deletedUser = await deleteOrganizationUser(
+      session.token,
+      organizationId,
+      userId,
+    );
+    return NextResponse.json({ data: deletedUser }, { status: 200 });
```
Comment (doug-s-nava) [inline frontend/src/app/api/user/organizations/[organizationId]/users/[userId]/handler.ts:27]: is status 200 necessary here? What does the response do if you don't provide it?

frontend/src/components/manageUsers/ActiveUsers.tsx
```diff
@@ -0,0 +1,129 @@
...
+}
+
+function formatFullName(user: UserDetail): string {
+  if (!user || !user.first_name) return "-";
+
+  const parts = [user.first_name, user.middle_name, user.last_name].filter(
+    Boolean,
+  );
+
+  return parts.length > 0 ? parts.join(" ") : "-";
```
doug-s-nava, line 27: seems like logically you've already confirmed that `user.first_name` exists, so `parts.length` should always be > 0

frontend/src/components/manageUsers/ActiveUsers.tsx
```diff
@@ -0,0 +1,129 @@
...
+}
+
+function getRoleId(role: UserRole): string {
+  return (role as { role_id: string }).role_id;
```
doug-s-nava, line 31: ??? `role_id` should already exist in UserRole, why is this necessary?

oh I think I see it - I think you can handle the typing within the parent function

```
  return match ? getRoleId(match as UserRole) : "";
```

?

frontend/src/components/manageUsers/ManageUsersClient.tsx
```diff
@@ -0,0 +1,211 @@
...
+export interface ManageUsersClientProps {
+  organizationId: string;
+  roles: UserRole[];
+  activeUsers: UserDetail[];
+  legacySystemUsers: UserDetail[];
+  pendingUsers: UserDetail[];
+}
+
+function getRoleId(role: UserRole): string {
+  return (role as { role_id: string }).role_id;
```
doug-s-nava, line 33: duplicate?

frontend/src/components/manageUsers/ActiveUsers.tsx
```diff
@@ -0,0 +1,129 @@
...
+  roles,
+  tableDescription,
+  tableTitle,
+  users,
+  busyUserId,
+  isSubmitting,
+  errorMessage,
+  onRequestRoleChange,
+}: UsersTableProps) {
+  const translatedText = useTranslations("ManageUsers.usersTable");
```
doug-s-nava, line 50: why not `t` as we do everywhere else?

frontend/src/components/manageUsers/ActiveUsers.tsx
```diff
@@ -0,0 +1,129 @@
...
+                  <td>{formatFullName(user)}</td>
+                  <td>{user.email}</td>
+                  <td>
+                    <label
+                      className="usa-sr-only"
+                      htmlFor={`role-select-${user.user_id}`}
+                    >
+                      {translatedText("selectRoleFor")} {formatFullName(user)}
+                    </label>
+                    <select
```
doug-s-nava, line 102: in some cases we're using the truss component for selects - does that do anything for us?

frontend/src/components/manageUsers/ManageUsersClient.tsx
```diff
@@ -0,0 +1,211 @@
...
+  type RefObject,
+} from "react";
+import { ModalRef } from "@trussworks/react-uswds";
+
+import { ActiveUsers } from "./ActiveUsers";
+import { ConfirmRoleChangeModal } from "./ConfirmRoleChangeModal";
+import { LegacySystemUsers } from "./LegacySystemUsers";
+import { PendingUsers } from "./PendingUsers";
+
+type Bucket = "active" | "legacy" | "pending";
```
doug-s-nava, line 22: ???

```suggestion
type UserStatusBuckets = "active" | "legacy" | "pending";
```

frontend/src/components/manageUsers/ManageUsersClient.tsx
```diff
@@ -0,0 +1,211 @@
...
+  legacySystemUsers,
+  pendingUsers,
+}: ManageUsersClientProps) {
+  const t = useTranslations("ManageUsers");
+  // Local copy of users split into buckets.
+  // We keep this in client state so we can:
+  // - Optimistically update role changes
+  // - Avoid blanking out a row when the API response doesn't
+  //   include the full user object (only updated roles, etc.)
+  const [rows, setRows] = useState<{
```
doug-s-nava, line 49: this is nitpicky, but it's confusing for me to think of these as rows, since they are more groups of users. Can we call it userGroups or userBuckets or something?

frontend/src/components/manageUsers/ManageUsersClient.tsx
```diff
@@ -0,0 +1,211 @@
...
+    <>
+      <>
+        <ActiveUsers
+          roles={roles}
+          tableDescription={t("activeUsersTableDescription")}
+          tableTitle={t("activeUsersHeading")}
+          users={rows.active}
+          busyUserId={busyUserId}
+          isSubmitting={isSubmitting}
+          onRequestRoleChange={handleActiveRoleChange}
```
doug-s-nava, line 186: is there a way to isolate the stuff here that's specific to the ActiveUsers table (everything around making changes, the modal) to the ActiveUsers component, and just passing down the necessary state management functions? Or is that idea that this stuff would be shared by the other table components as they get built out?

frontend/src/components/manageUsers/LegacySystemUsers.tsx
```diff
@@ -0,0 +1,62 @@
...
+import React from "react";
+import { Table } from "@trussworks/react-uswds";
+
+export interface UsersTableProps {
+  tableDescription: string;
+  tableTitle: string;
+  users: UserDetail[];
+}
+
+function formatFullName(user: UserDetail): string {
```
doug-s-nava, line 15: why is this different from the other implementation? can we make one general implementation and share it in a utils file?

frontend/src/app/[locale]/(base)/organization/[id]/manage-users/page.tsx
```diff
@@ -14,19 +19,49 @@ export async function generateMetadata({
...
+    if (!session?.token) {
+      throw new Error("not logged in");
+    }
+    const organizationDetails = await getOrganizationDetails(session.token, id);
+    title = `${t("ManageUsers.pageTitle")} - ${organizationDetails.sam_gov_entity.legal_business_name || ""}`;
+  } catch (error) {
+    console.error("Failed to render page title due to API error", error);
+    if (parseErrorStatus(error as ApiRequestError) === 404) {
+      return notFound();
+    }
```
Comment (ErinPattisonNava) [inline frontend/src/app/[locale]/(base)/organization/[id]/manage-users/page.tsx:37]: What happens for non 404 errors, like 500?  If we just want it to fail silently here, maybe a comment to that effect?

frontend/src/app/[locale]/(base)/organization/[id]/manage-users/page.tsx
```diff
@@ -14,19 +19,49 @@ export async function generateMetadata({
...
   params: Promise<{ locale: string; id: string }>;
-}) {
-  const { locale } = await params;
+}): Promise<Metadata> {
+  const { locale, id } = await params;
   const t = await getTranslations({ locale });
-
+  let title = `${t("ManageUsers.pageTitle")}`;
+  try {
+    const session = await getSession();
```
Comment (ErinPattisonNava) [inline frontend/src/app/[locale]/(base)/organization/[id]/manage-users/page.tsx:27]: This feels like this check might be better in the fetcher?  Also wondering if the whole page should be wrapped in `AuthenticationGate` like [this](https://github.com/HHS/simpler-grants-gov/blob/815137620fce456022e929e9a37216a0884930d9/frontend/src/app/%5Blocale%5D/(base)/saved-search-queries/layout.tsx#L22-L23)

frontend/src/app/[locale]/(base)/organization/[id]/manage-users/page.tsx
```diff
@@ -14,19 +19,49 @@ export async function generateMetadata({
...
-function ManageUsersPage({ params }: ManageUsersPageProps) {
-  const { id: organizationId } = use(params);
-  return <ManageUsersPageContent organizationId={organizationId} />;
+async function ManageUsersPage({ params }: ManageUsersPageProps) {
+  const session = await getSession();
+  const { id } = await params;
+  return (
+    <AuthorizationGate
+      resourcePromises={{
+        organizationDetails: getOrganizationDetails(session?.token || "", id),
```
Comment (ErinPattisonNava) [inline frontend/src/app/[locale]/(base)/organization/[id]/manage-users/page.tsx:51]: Why wouldn't they have a token, is this page available to non-logged in users?

frontend/src/components/manageUsers/ActiveUsers.tsx
```diff
@@ -0,0 +1,129 @@
...
+  tableTitle: string;
+  users: UserDetail[];
+  busyUserId?: string;
+  isSubmitting?: boolean;
+  errorMessage?: string;
+  onRequestRoleChange: (userId: string, nextRoleId: string) => void;
+}
+
+function formatFullName(user: UserDetail): string {
+  if (!user || !user.first_name) return "-";
```
ErinPattisonNava, line 21: Why wouldn't we return the empty string in this case?

frontend/src/components/manageUsers/ActiveUsers.tsx
```diff
@@ -0,0 +1,129 @@
...
+}
+
+function getRoleId(role: UserRole): string {
+  return (role as { role_id: string }).role_id;
+}
+
+function currentRoleIdForUser(user: UserDetail, allRoles: UserRole[]): string {
+  const currentRoleName = user.roles?.[0]?.role_name ?? "";
+  const match = allRoles.find((role) => role.role_name === currentRoleName);
+  return match ? getRoleId(match) : "";
```
ErinPattisonNava, line 37: Would it be better to return undefined?

frontend/src/components/manageUsers/ActiveUsers.tsx
```diff
@@ -0,0 +1,129 @@
...
+      <h2 className="margin-bottom-1 font-sans-lg">{tableTitle}</h2>
+      <p className="margin-bottom-2 margin-top-1">{tableDescription}</p>
+
+      {errorMessage ? (
+        <div className="margin-bottom-2">
+          <ErrorMessage>{errorMessage}</ErrorMessage>
+        </div>
+      ) : null}
+
+      <Table className="width-full overflow-wrap simpler-application-forms-table">
```
ErinPattisonNava, line 63: Why not use `TableWithResponsiveHeader` here?

frontend/src/components/manageUsers/ActiveUsers.tsx
```diff
@@ -0,0 +1,129 @@
...
+  tableTitle,
+  users,
+  busyUserId,
+  isSubmitting,
+  errorMessage,
+  onRequestRoleChange,
+}: UsersTableProps) {
+  const translatedText = useTranslations("ManageUsers.usersTable");
+
+  return (
```
ErinPattisonNava, line 52: nit: This feels long enough that maybe we should break out some helper methods.

frontend/src/components/manageUsers/ManageUsersPageContent.tsx
```diff
@@ -23,20 +29,39 @@ export async function ManageUsersPageContent({
...
+  ]);
+
+  let users: UserDetail[] = [];
+  let userOrganization: Organization | undefined;
+  let roles: UserRole[] = [];
+
+  if (usersResult.status === "fulfilled") {
+    users = usersResult.value;
+  } else {
+    console.error("Unable to fetch organization users", usersResult.reason);
```
ErinPattisonNava, line 46: For each of these errors, what's the error look like on the page?

frontend/src/components/manageUsers/ManageUsersClient.tsx
```diff
@@ -0,0 +1,211 @@
...
+import { ModalRef } from "@trussworks/react-uswds";
+
+import { ActiveUsers } from "./ActiveUsers";
+import { ConfirmRoleChangeModal } from "./ConfirmRoleChangeModal";
+import { LegacySystemUsers } from "./LegacySystemUsers";
+import { PendingUsers } from "./PendingUsers";
+
+type Bucket = "active" | "legacy" | "pending";
+
+export interface ManageUsersClientProps {
```
ErinPattisonNava, line 24: Is "Client" here referring to the type of component, e.g. client vs server?  If so, I'd rather rather it be named after the object it represents, maybe around ManageUsersRoleManager?  

I have to admit it took me a while to get my head what this represented, so you may want to think about ways to separate out behaviour into smaller components/structure the file differently for clarity if possible.

frontend/src/components/manageUsers/ManageUsersClient.tsx
ErinPattisonNava: Or UserStatusTypes?

frontend/src/i18n/messages/en/index.ts
```diff
@@ -1541,15 +1541,18 @@ export const messages = {
...
-    grantsGovUsersHeading: "Grants.gov Users",
-    grantsGovUsersTableDescription:
+    legacySystemUsersHeading: "Grants.gov Users",
+    legacySystemUsersTableDescription:
       "Manage users from your Grants.gov organization here in Simpler.",
     pendingUsersHeading: "Pending Users",
     pendingUsersTableDescription:
       "Any users waiting to join your organization will appear below.",
     errors: {
+      activeUsersFetchError: "Failed to get list of active users",
```
ErinPattisonNava, line 1551: I don't think these are used anywhere?

frontend/src/components/manageUsers/ManageUsersClient.tsx
```diff
@@ -0,0 +1,211 @@
...
+                  }
+                : user,
+            ),
+          };
+        });
+      }
+
+      handleCloseDialog();
+    } catch (error) {
+      console.error("Failed to update user role", error);
```
ErinPattisonNava, line 145: Should this also show an error on the page so they know the action didn't take?

frontend/src/components/manageUsers/ManageUsersClient.tsx
```diff
@@ -0,0 +1,211 @@
...
+          tableDescription={t("activeUsersTableDescription")}
+          tableTitle={t("activeUsersHeading")}
+          users={rows.active}
+          busyUserId={busyUserId}
+          isSubmitting={isSubmitting}
+          onRequestRoleChange={handleActiveRoleChange}
+        />
+
+        <LegacySystemUsers
+          tableDescription={t("legacySystemUsersHeading")}
```
ErinPattisonNava, line 190: Why is the Description and the Title for each of these tables is being passed in from the parent, vs living in the child components directly?  Better encapsulation of content within each group/table.

frontend/src/components/manageUsers/ManageUsersClient.tsx
```diff
@@ -0,0 +1,211 @@
...
+      pending: pendingUsers,
+    });
+  }, [activeUsers, legacySystemUsers, pendingUsers]);
+
+  // Modal state
+  const [pendingUserId, setPendingUserId] = useState<string>("");
+  const [pendingNextRoleId, setPendingNextRoleId] = useState<string>("");
+  const [pendingBucket, setPendingBucket] = useState<Bucket | null>(null);
+  const [busyUserId, setBusyUserId] = useState<string | undefined>(undefined);
+  const [isSubmitting, setIsSubmitting] = useState(false);
```
ErinPattisonNava, line 79: I don't quite get the pattern we're using to lock rows/deal with concurrent changes/prevent double clicks.  Are we only allowing them to change the role of one user at a time?  Does the modal stay up until we we have confirmation that the action is completed? If it does, then it feels like much of this could be simplified.  If we are allowing multiple users to change their role at once, should these be arrays?

Issue #6537, opened 2025-09-26
Implement Auth-Based Role Access with Preloaded Roles


babebe:
### Summary

Refactor the current access control logic by creating a wrapper function  that calls `get_roles_and_privileges` to prefetch the user’s roles and privileges for the provided resource. It then calls can_access method  for permission checks without relying on get_roles_for_resource.

The new wrapper function` check_user_access` should:

- Call  `get_roles_and_privileges` for the given user and resource.

- Invokes `can_access` with the filtered roles to perform the permission check.

Update can_access:

- Remove usage of the current `get_roles_for_resource` method. 

Reference: https://betagrantsgov.slack.com/archives/C05TSL64VUH/p1758895674020429

### Acceptance criteria

- [ ] A new wrapper function created.

- [ ] can_access method refactored

- [ ] Add/Update test

eric-serrato95:
need to do before lunch, but not urgent prior to that

PR #6977, opened 2025-11-06
[Issue #6537] pre load roles when checking access

babebe:
## Summary

Fixes / Work for #6537 

## Changes proposed

Wrapper function around can_access method to reduce database calls to fetch roles.
Update code to use wrapper function
Tests updated

### DIFF
--- a/api/src/constants/lookup_constants.py
+++ b/api/src/constants/lookup_constants.py
@@ -145,11 +145,13 @@ class FormType(StrEnum):
     SF424 = "SF424"
     SF424A = "SF424A"
     SF424B = "SF424B"
+    SF424D = "SF424D"
     SFLLL = "SFLLL"
     PROJECT_NARRATIVE_ATTACHMENT = "ProjectNarrativeAttachment"
     BUDGET_NARRATIVE_ATTACHMENT = "BudgetNarrativeAttachment"
+    OTHER_NARRATIVE_ATTACHMENT = "OtherNarrativeAttachment"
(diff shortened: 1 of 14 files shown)


chouinar:
@babebe - the tests on this are failing

api/src/auth/endpoint_access_util.py
```diff
@@ -113,3 +115,16 @@ def get_log_info_for_resource(resource: Organization | Application | Agency | No
...
+def check_user_access(
+    db_session: db.Session,
+    user: User,
+    allowed_privileges: set[Privilege],
+    resource: Organization | Application | Agency | None,
+) -> None:
+    """Wrapper function to preload all roles and privileges for the given user in a single query,
+    then checks access using the in-memory data."""
+    user_with_role = get_roles_and_privileges(db_session, user.user_id)
+    if user_with_role and not can_access(user_with_role, allowed_privileges, resource):
```
chouinar: It shouldn't happen, but can we instead make this error if no user was found? With the way our auth works, the user needs to exist to even be authenticated in the first place (minus that one weird endpoint with the alt auth, but it shouldn't call this for that scenario).

```suggestion
    if not user_with_role or not can_access(user_with_role, allowed_privileges, resource):
```

chouinar:
Change looks good, but just because it's a wide ticket, want to hold until after the prod deploy. Just in case this actually makes the performance of our various endpoints get much worse.

chouinar:
@babebe looks like merging from main has a test failure, might be a flaky test as it seems unrelated

Issue #6775, opened 2025-10-17
option to turn off token expiration / auto log out in local env


doug-s-nava:
### Summary

In order to ease the pain of being logged out while working on local development, we want to give developers the option to turn off auto-log-out / token expiration in local environments

To accomplish this we'll need to:
- establish a way to turn the feature on and off on the frontend app, likely an environment variable
- ~~since the API side token has its own expiration, establish a way to turn the feature on and off on the API~~
- build in a failsafe that will prevent this feature from being turned on in the frontend in any deployed environments

### Implementation notes

Disabling automatic logouts means setting frontend and API side token expiration dates to the far future
   
- this is already controlled by an env var on the API side, as well as in each JWT
    - since the JWTs for test users locally will be defined by seed logic going forward, we can count on the expiration of those individual JWT expirations being set to the far future at seed time, and so there's no extra work to be done here 
- on the frontend, we will need to introduce an env var to control the expiration time
    - as well as either defaults or env vars set through an npm task  

### Acceptance criteria

- [ ] local users can easily make it so that logins will not expire in their local environment, and as a result a user will not be logged out due to expired tokens
- [ ] this feature is not available / not possible to turn on in deployed envs
- [ ] relevant documentation created or updated

PR #6983, opened 2025-11-06
[Issue #6775] Option to turn off token expiration / auto log out in local env

glenblosser-nava:
## Summary

Fixes / Work for #6775 

## Changes proposed

Added a new environment variable that allows the developer to set their own timeout `AUTH_EXPIRATION_TIME`

## Context for reviewers

Previously the token would time out after 15 minutes. This made for some annoyances during local testing. Now, you can set the timeout in the .env file, with the example given in .env.development being 1 year.

## Validation steps

Log in to a local testing environment with AUTH_EXPIRATION_TIME set, and confirm that it does not auto log you within 15 minutes.


frontend/src/services/auth/sessionUtils.ts
```diff
@@ -3,15 +3,25 @@
...
 export const CLIENT_JWT_ENCRYPTION_ALGORITHM = "HS256";
 export const API_JWT_ENCRYPTION_ALGORITHM = "RS256";
 
 // returns a new date 15 minutes from time of function call
-export const newExpirationDate = () =>
-  new Date(Date.now() + clientTokenExpirationInterval);
+export const newExpirationDate = () => {
+  if (
+    environment.AUTH_EXPIRATION_TIME !== "0" &&
+    environment.ENVIRONMENT === "dev"
```
doug-s-nava: looks like "ENVIRONMENT" in dev will come through as "dev". To limit this to local only, can we tie this to something that for sure will only ever be set in a local environment? `ENVIRONMENT === 'local'` could work but we would need to make sure to set that when running the local server. 

Not sure why we're defaulting that to 'dev', but probably best to change that, since that suggests a deployed environment.

I'd advocate for creating a new NPM script that adds ENVIRONMENT="local", and we can get in the habit of using that when we don't want to be logged out?

frontend/.env.production
```diff
@@ -2,3 +2,5 @@
 # This file is checked into your git repo and provides defaults for non-sensitive
 # env vars when running `next dev` only.
 # Learn more: https://nextjs.org/docs/app/building-your-application/configuring/environment-variables
+
+ENVIRONMENT=prod
```
doug-s-nava: Forgot this file existed, and don't think it needs to exist. Let's maybe talk through in a frontend sync, but in any case I don't think we need to make this update now

documentation/frontend/environments.md
```diff
@@ -27,6 +27,8 @@ When running Next in a non-dockerized situation, variables are also pulled direc
 
 Since the .env.development file is tracked in git, any sensitive environment variables used for local development or testing - for example to connect to a non-dev API instance - **SHOULD NOT** be entered into this file, even temporarily. Instead, pass any necessary env vars directly from the command line or look into something like [direnv](https://direnv.net/).
 
+The `AUTH_EXPIRATION_TIME` environment variable will allow you to control the time your login token will remain valid. The value in .env.development is 1 year, but you can set a different value as needed in .env.local.
```
doug-s-nava: probably good to note that this value is only honored where ENVIRONMENT === 'local'

doug-s-nava (approved):
✅ ✅ ✅ ✅

PR #6987, opened 2025-11-06
[Issue #6982] Improve validation in CommonGrants services

### LINKED ISSUE #6982: [CommonGrants] Fix `validate_url()` in CommonGrants API routes
### Summary

Currently `validate_url()` uses a different pattern for validating URLs than the `HttpUrl` field supported by Pydantic, which causes some invalid URLs to get through and raise an error on `OpportunityBase.model_validate()` which gets caught and logged as an error in the API logging.

While this doesn't bubble up to API consumers it does trigger an alert in New Relic.

### Acceptance criteria

- [ ] `validate_url()` uses `HttpUrl` to validate URLs
- [ ] We log validation errors as warnings instead of errors
- [ ] We add more information to the logger

widal001:
## Summary

Improves how validation handling and logging works for CommonGrants services, in order to address a bug that allowed URLs with problematic characters to slip through `validate_url()` but then fail Pydantic validation.

For additional context see [this conversation in Slack](https://betagrantsgov.slack.com/archives/C05TB4ZSEGN/p1762341552084149).

Fixes #6982 

## Changes proposed

- **Fixes bug with `validate_url()` in CommonGrants routes**
    Previous implementation used `urlparse`, which allowed some URLs to pass through that ultimately failed `HttpUrl` validation in Pydantic with strict mode enabled. This now uses the same Pydantic field configuration to validate URLs.
- **Improves logging for CommonGrants routes**
  Adds `extras` argument to logging with input data and `CommonGrantsEvents` enum when validation fails. Also downgrades validation logging from `exception` to `warning` so we aren't tripping the New Relic alerts on every validation issue.
- **Tests validation handling and logging**
  Confirms how CommonGrants services handle validation errors and logging

## Context for reviewers

A couple of notes:

- Eventually we may want to relax the `strict=true` configuration for URL fields in the CommonGrants SDK and/or switch to a single UrlField implementation that we can update `validate_url` to use (after we publish a new SDK version)
- Since this downgrades `logging.exception()` to `logging.warning()` we might want to set up a dashboard in New Relic that filters for these validation warnings.

## Validation steps

I added a set of unit tests that specifically recreate this bug and check for the updated handling:

- [Test validating fix for URL bug](https://github.com/HHS/simpler-grants-gov/pull/6987/commits/29979920df6b2900a8f57bc2978d5ba9abca910f#diff-5c60920b2b8f217c3e0db84eaf262c0829eea35bf11387b798cd4ef39e382b16)
- [Test validating updates to the logging](https://github.com/HHS/simpler-grants-gov/pull/6987/commits/4efbc52a2dcf6765a43a2dcafa528b44dcee8eae)

### DIFF
--- a/api/src/services/common_grants/transformation.py
+++ b/api/src/services/common_grants/transformation.py
@@ -2,7 +2,6 @@
 
 import logging
 from datetime import date, datetime, timezone
-from typing import Annotated
 
 from common_grants_sdk.schemas.pydantic import (
     FilterInfo,
@@ -19,7 +18,7 @@ from common_grants_sdk.schemas.pydantic import (
     PaginatedBodyParams,
     SingleDateEvent,
 )
-from pydantic import Field, HttpUrl, TypeAdapter, ValidationError
+from pydantic import BaseModel, Field, HttpUrl, ValidationError
 
 from src.api.response import ValidationErrorDetail
 from src.constants.lookup_constants import CommonGrantsEvent, OpportunityStatus
@@ -27,6 +26,16 @@ from src.db.models.opportunity_models import Opportunity
 from src.validation.validation_constants import ValidationErrorType
 
 logger = logging.getLogger(__name__)
+
+
+class UrlValidator(BaseModel):
+    """Validator for a URL string using Pydantic's HttpUrl with strict mode enabled.
+
+    Mirrors the HttpUrl field in OpportunityBase and other CommonGrants models.
+    TODO(@widal001): Replace this with a new field from SDK or relax strictness in SDK
+    """
+
+    url: HttpUrl = Field(strict=True)
 
 
 def transform_status_to_cg(v1_status: OpportunityStatus) -> OppStatusOptions:
@@ -155,19 +164,14 @@ def validate_url(value: str | None) -> str | None:
     """
     if value is None or value == "":
         return None
-
-    # Mirrors the HttpUrl field in OpportunityBase
-    # TODO(@widal001): Replace this with a new field from SDK or relax strictness in SDK
-    UrlField = Annotated[HttpUrl, Field(strict=True)]
-    url_adapter = TypeAdapter(UrlField)
-
     try:
-        return str(url_adapter.validate_python(value))
+        valid = UrlValidator.model_validate({"url": value})
+        return str(valid.url)
     except ValidationError:
         logger.info(
             f"URL validation failed for: {value}",
             extra={
-                "event": CommonGrantsEvent.URL_VALIDATION_ERROR,
+                "cg_event": CommonGrantsEvent.URL_VALIDATION_ERROR,
                 "url": value,
             },
         )
@@ -296,7 +300,7 @@ def transform_search_result_to_cg(opp_data: dict) -> OpportunityBase | None:
         logger.warning(
             f"Failed to transform search result to CommonGrants format: {e}",
             extra={
-                "event": CommonGrantsEvent.OPPORTUNITY_VALIDATION_ERROR,
+                "cg_event": CommonGrantsEvent.OPPORTUNITY_VALIDATION_ERROR,
                 "opportunity_id": opportunity_id,
             },
         )
--- a/api/tests/src/services/common_grants/test_transformation.py
+++ b/api/tests/src/services/common_grants/test_transformation.py
@@ -688,8 +688,8 @@ class TestTransformation:
             None,
         )
         assert log_record is not None
-        assert hasattr(log_record, "event")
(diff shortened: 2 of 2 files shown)


api/tests/src/services/common_grants/test_transformation.py
```diff
@@ -31,6 +32,30 @@
...
+def _legacy_validate_url(value: str | None) -> str | None:
+    """
+    Validate a URL string.
+
+    Args:
+        value: The string to validate
+
+    Returns:
+        A valid URL string or None
+    """
```
widal001, line 44: Added the old implementation to the tests so we could recreate the bug and check that the new `validate_url()` handles it differently.

api/src/services/common_grants/transformation.py
```diff
@@ -139,26 +139,39 @@ def _transform_date_to_cg(date_value: date | datetime | None) -> date | None:
...
-    if not parsed.scheme and parsed.netloc:
-        return f"https://{value}"
+    # Mirrors the HttpUrl field in OpportunityBase
+    # TODO(@widal001): Replace this with a new field from SDK or relax strictness in SDK
+    UrlField = Annotated[HttpUrl, Field(strict=True)]
+    url_adapter = TypeAdapter(UrlField)
 
-    return None
+    try:
+        return str(url_adapter.validate_python(value))
```
chouinar: Thoughts on instead making a Pydantic object like:

```py
class UrlValidator(CommonGrantsBaseModel):
    url_field: HttpUrl | None
```
And then using that to do the validation? Makes it basically identical to what we're going to put it in, and feels a bit more intuitive to read as defining types like you have here is getting into some Pydantic internals.

---

Since the Pydantic object is static, could we define it statically rather than remaking it each time this function is called?

api/src/services/common_grants/transformation.py
```diff
@@ -280,7 +293,13 @@ def transform_search_result_to_cg(opp_data: dict) -> OpportunityBase | None:
             lastModifiedAt=opp_data.get("updated_at") or datetime.now(timezone.utc),
         )
     except Exception as e:
-        logger.exception(f"Failed to transform search result to CommonGrants format: {e}")
+        logger.warning(
+            f"Failed to transform search result to CommonGrants format: {e}",
+            extra={
+                "event": CommonGrantsEvent.OPPORTUNITY_VALIDATION_ERROR,
```
chouinar: Could we make this event have a more specific name? Something like `common_grants_event` or `cg_event`? Something so that we can really easily separate different events in our dashboards (ie. not getting these events mixed up with events from apply, or wherever else we might add this pattern).

api/src/services/common_grants/transformation.py
widal001: Yeah I think that could be a good shift -- it was my understanding that instantiating and then validating with a class-based schema had a bit more overhead, which is why I opted for this approach, but I can appreciate that this might not be as intuitive. Plus moving the instantiation out of the function should offset any performance tradeoff.

api/src/services/common_grants/transformation.py
widal001: Sure! You're just talking about the key name right?

api/src/services/common_grants/transformation.py
widal001: Updated this pattern here, does seem a little easier to reason about: [refactor(api): Incorporate PR feedback](https://github.com/HHS/simpler-grants-gov/pull/6987/commits/4d0170d26f0d0fde50f534e9757ba36233d86a64)

api/src/services/common_grants/transformation.py
widal001: Changed both instances of `event` to `cg_event` in this commit: [refactor(api): Incorporate PR feedback](https://github.com/HHS/simpler-grants-gov/pull/6987/commits/4d0170d26f0d0fde50f534e9757ba36233d86a64)

api/src/services/common_grants/transformation.py
```diff
@@ -155,19 +164,14 @@ def validate_url(value: str | None) -> str | None:
...
-
-    # Mirrors the HttpUrl field in OpportunityBase
-    # TODO(@widal001): Replace this with a new field from SDK or relax strictness in SDK
-    UrlField = Annotated[HttpUrl, Field(strict=True)]
-    url_adapter = TypeAdapter(UrlField)
-
     try:
-        return str(url_adapter.validate_python(value))
+        valid = UrlValidator.model_validate({"url": value})
+        return str(valid.url)
```
widal001, line 169: I tried doing this `return str(UrlValidator(url=value).url)` but the type checker threw an error because url should be a `HttpUrl` (the whole point of the validator lol)

PR #6996, opened 2025-11-07
[ISSUE #6809]: Add invitation instructions email sending to user invitation endpoint 

### LINKED ISSUE #6809: Add invitation instructions email sending to user invitation endpoint
### Summary

Implement email sending functionality for the user invitation creation endpoint that sends setup instructions to invited users. 

### Technical Requirments 
- Send invitation email with setup instructions when invitation record is created
- We should be able to at least partially utilize the `send_pinpoint_email_raw()` function from `src/adapters/aws/pinpoint_adapter.py` 

### Discussion
- For now should we just log an error on the email?
- Do we care to store any email sending statuses on the invitation record?  
 

### Acceptance criteria

- As part of user invite endpoint, an email is sent to the invited user
- Email contains proper setup instructions
- Error handling captures and logs email sending failures
- unit tests created / updated

joshtonava:
## Summary

Fixes #6809 

## Changes proposed
- Added email notification when organization invitations are created
- Created minimal email template with setup instructions

## Context for reviewers
- Email failures are logged but don't prevent invitation creation
- Generated UUID trace_id is passed to Pinpoint and logged as pinpoint_trace_id for correlation

## Validation steps

Start the local API server:
```bash
make run-logs
```
1: Get an Authentication Token
```bash
TOKEN=$(curl -s http://localhost:8080/local/local-users | jq -r '.data[] | select(.oauth_id == "one_org_user") | .user_jwt')

# Verify you got the token
echo $TOKEN
```

2: Set Up Variables
```
# Organization ID (Sally's Soup Emporium - the org that one_org_user is admin of)
ORG_ID="47d95649-c70d-44d9-ae78-68bf848e32f8"

# Role ID (Organization Member role)
ROLE_ID="336a530f-a356-4fbd-85ef-bc0ce18f89c8"
```

3: Create an Organization Invitation
```bash
curl -X POST "http://localhost:8080/v1/organizations/${ORG_ID}/invitations" \
    -H "X-SGG-Token: ${TOKEN}" \
    -H "Content-Type: application/json" \
    -d "{
      \"invitee_email\": \"newuser@example.com\",
      \"role_ids\": [\"${ROLE_ID}\"]
    }" | jq '.'
```

In the terminal server logs you should see log lines for:
- Sending invitation email
- Invitation email sent successfully

### DIFF
--- a/api/src/task/notifications/config.py
+++ b/api/src/task/notifications/config.py
@@ -30,3 +30,13 @@ def get_email_config() -> EmailNotificationConfig:
     if _config is None:
         _config = EmailNotificationConfig()
     return _config
+
+
+def _reset_email_config() -> None:
+    """Reset the singleton EmailNotificationConfig instance.
+
+    This is primarily used for testing to ensure environment variable changes
+    are picked up between test cases.
+    """
+    global _config
+    _config = None
--- a/api/tests/src/task/notifications/test_search_notification.py
+++ b/api/tests/src/task/notifications/test_search_notification.py
@@ -10,7 +10,7 @@ from src.api.opportunities_v1.opportunity_schemas import OpportunityV1Schema
 from src.constants.lookup_constants import OpportunityStatus
 from src.db.models.opportunity_models import Opportunity
 from src.db.models.user_models import SuppressedEmail, UserNotificationLog, UserSavedSearch
-from src.task.notifications.config import EmailNotificationConfig
+from src.task.notifications.config import EmailNotificationConfig, _reset_email_config
 from src.task.notifications.constants import NotificationReason
 from src.task.notifications.email_notification import EmailNotificationTask
 from src.task.notifications.search_notification import (
@@ -49,7 +49,8 @@ def setup_opensearch_data(opportunity_index_alias, search_client):
 
 @pytest.fixture(autouse=True)
 def clear_data(db_session):
-    """Clear all notification logs"""
+    """Clear all notification logs and reset email config singleton"""
+    _reset_email_config()
     cascade_delete_from_db_table(db_session, UserNotificationLog)
     cascade_delete_from_db_table(db_session, Opportunity)
     cascade_delete_from_db_table(db_session, UserSavedSearch)


api/src/services/organizations_v1/create_organization_invitation.py
```diff
@@ -20,6 +23,55 @@
...
+    error handling, and logging. Email failures do not raise exceptions.
+
+    Args:
+        invitation: The created invitation record
+        organization: The organization the user is being invited to
+        inviter: The user who created the invitation
+        invitee_email: Email address of the invitee
+    """
+    try:
+        config = EmailNotificationConfig()
```
chouinar: Nitpicky - could we add a function to the email notification config file like:

```py
_config: EmailNotificationConfig | None = None
def get_email_config() -> EmailNotificationConfig:
   global _config
   if _config is None:
       _config = EmailNotificationConfig()

   return _config
```
Effectively, we only make the config once globally, since the values for it can only change by deploying, no need to refetch the env vars every request. We do this pattern for a few of our other configs.

api/src/services/organizations_v1/create_organization_invitation.py
```diff
@@ -20,6 +23,55 @@
...
+    This function handles all email sending logic including configuration,
+    error handling, and logging. Email failures do not raise exceptions.
+
+    Args:
+        invitation: The created invitation record
+        organization: The organization the user is being invited to
+        inviter: The user who created the invitation
+        invitee_email: Email address of the invitee
+    """
+    try:
```
chouinar: A little nitpicky, rather than wrapping everything here in a try block, could we just wrap the email send which is the only part I think we would meaningfully expect a failure.

api/src/services/organizations_v1/create_organization_invitation.py
```diff
@@ -129,5 +181,8 @@ def create_organization_invitation(
 
     db_session.add(invitation)
 
+    # Send invitation email (non-blocking)
```
chouinar: Non-blocking as in async, or non-blocking as in - won't cause any issues even if it fails? I think you mean the latter, right?

joshtonava: As in it won't cause issues if it fails. I can make that more clear

api/src/services/organizations_v1/create_organization_invitation.py
joshtonava: Yea no problem

PR #7005, opened 2025-11-07
Gblosser/5931 hide save button button only

glenblosser-nava:
## Summary

Work for #5931

## Changes proposed

When the form is not in progress, the Save button is hidden
This PR includes work that will be implemented in a future PR for #5931

## Context for reviewers

This PR is part of larger work on the #5931 ticket, but _only_ features the hidden save button change. The readonly changes are also present in [this](https://github.com/HHS/simpler-grants-gov/pull/6870/) PR.

## Validation steps

After submitting an application, the Save button should not be visible on forms.


frontend/src/components/applyForm/widgets/WidgetRenderers.tsx
```diff
@@ -50,6 +50,9 @@ export const renderWidget = ({
     console.error(`Unknown widget type: ${type}`, definition);
     throw new Error(`Unknown widget type: ${type}`);
   }
+  if (props.isFormLocked) {
```
doug-s-nava, line 53: note that `isFormLocked` is not being passed, so this won't do anything yet

frontend/src/components/applyForm/types.ts
```diff
@@ -124,7 +124,8 @@ export interface UswdsWidgetProps<
   maxLength?: number;
   required?: boolean;
   disabled?: boolean;
-  readonly?: boolean;
+  readOnly?: boolean;
```
doug-s-nava, line 127: note that `readOnly` is not being passed yet, so implementation of this will not be used, but may later be depending on the requirements for https://github.com/HHS/simpler-grants-gov/issues/7067

PR #7019, opened 2025-11-10
[Unticketed] Remove network calls to fetch XSDs at runtime

mikehgrantsgov:
## Summary

Fixes issue where XSDs were being pulled at runtime, which is used by our XML generation logic to enforce element ordering (required in some cases). Instead of XSD ordering, rely on ordering from our configs.

Additionally, remove some SF424 hard-coding.

## Changes proposed

Update order of `applicant_type_code_mapping` to fix XSD validation.
Remove `_get_xsd_element_order` and `_get_complex_type_element_order` which previously used network requests at runtime to fetch XSDs.
Add helper method to inspect config and extract element order.
Add a test for the scenario.

## Validation steps

See new test and run `make fetch-xsds` and `make test-xml-validation` to see CLI tests passing.

### DIFF
--- a/api/src/services/xml_generation/service.py
+++ b/api/src/services/xml_generation/service.py
@@ -14,6 +14,11 @@ from .utils.attachment_mapping import AttachmentInfo
 logger = logging.getLogger(__name__)
 
 
+def _is_attribute_metadata_key(field_name: str) -> bool:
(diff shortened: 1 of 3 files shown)


chouinar (dismissed):
Thanks for the quick fix!

chouinar (approved):

Issue #6949, opened 2025-11-04
Setup foreign tables and staging tables for legacy user data discovery


joshtonava:
### Summary

We need to set up the Oracle db foreign tables and staging tables for legacy users data discovery. 

Tech Notes:
- [Legacy Data Mixins](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2174681171/Tech+Spec+Make+it+easy+for+Organizations+to+invite+users+from+existing+system#Legacy-Data-Mixins) 
- [Foreign Table Models](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2174681171/Tech+Spec+Make+it+easy+for+Organizations+to+invite+users+from+existing+system#Foreign-Table-Models
)
- [Configure Hourly Data Load](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2174681171/Tech+Spec+Make+it+easy+for+Organizations+to+invite+users+from+existing+system#Configure-Hourly-Data-Load
) 

### Acceptance criteria

- Foreign table models created 
- Staging table models created with required indexes  
- Migrations exist for staging tables
- Foreign Tables successfully connect to the Oracle DB
- Hourly load task configured to sync both tables

PR #7020, opened 2025-11-10
[Issue #6949] setup foreign staging tables legacy users[1/2]

babebe:
## Summary

Fixes  for #6949  

## Changes proposed

Foreign table models created
Staging table models created
Migration for staging tables

Will Fallow-up with a PR for Adding tables `vuser_account` and `tuser_profile ` to `load_oracle_data_task` after running manual test in dev/staging

### DIFF
--- a/api/tests/src/db/models/factories.py
+++ b/api/tests/src/db/models/factories.py
@@ -1443,6 +1443,11 @@ class ApplicationFactory(BaseFactory):
             organization=factory.SubFactory("tests.src.db.models.factories.OrganizationFactory")
         )
 
+        has_submitted_by_user = factory.Trait(
+            submitted_by=factory.LazyAttribute(lambda a: a.submitted_by_user.user_id),
+            submitted_by_user=factory.SubFactory(UserFactory),
+        )
+
 
 class ApplicationFormFactory(BaseFactory):
     class Meta:
--- a/api/src/api/application_alpha/application_route.py
+++ b/api/src/api/application_alpha/application_route.py
@@ -1,5 +1,8 @@
 import logging
+from typing import Never
 from uuid import UUID
+
+from apiflask.exceptions import HTTPError
 
 import src.adapters.db as db
 from src.adapters.db import flask_db
@@ -29,11 +32,13 @@ from src.api.application_alpha.application_schemas import (
 from src.api.schemas.response_schema import AbstractResponseSchema
 from src.auth.api_jwt_auth import api_jwt_auth
 from src.auth.multi_auth import jwt_key_or_internal_multi_auth, jwt_key_or_internal_security_schemes
+from src.constants.lookup_constants import ApplicationAuditEvent
 from src.db.models.user_models import UserTokenSession
 from src.logging.flask_logger import add_extra_data_to_current_request_logs
 from src.services.applications.add_organization_to_application import (
     add_organization_to_application,
 )
+from src.services.applications.application_audit import add_audit_event_by_id
 from src.services.applications.create_application import create_application
 from src.services.applications.create_application_attachment import create_application_attachment
 from src.services.applications.delete_application_attachment import delete_application_attachment
@@ -284,6 +289,35 @@ def application_get(
     )
 
 
+def _handle_submit_error(
+    db_session: db.Session, error: HTTPError, application_id: UUID, user_id: UUID
+) -> Never:
+    """Handle adding an audit event whenever someone attempts to submit and fails"""
+    # We don't add audit events for 403 (auth) or 404 (not found) cases as
+    # that wouldn't be useful for a user from an auditing perspective
+    if error.status_code != 422:
+        raise error
+    try:
+        # Rollback the existing transaction
+        if db_session.is_active:
+            db_session.rollback()
+        # Make a new transaction that we just add the audit event to
+        # This does rely on the application and user to both exist, but
+        # since we check that before any 422 cases in the service logic
+        # this should be safe to just pass the IDs in.
+        with db_session.begin():
+            add_audit_event_by_id(
+                db_session,
+                application_id=application_id,
+                user_id=user_id,
+                audit_event=ApplicationAuditEvent.APPLICATION_SUBMIT_REJECTED,
+            )
+    except Exception:
+        logger.exception("Failed to add audit event for failed submission")
+
+    raise error
+
+
 @application_blueprint.post("/applications/<uuid:application_id>/submit")
 @application_blueprint.output(AbstractResponseSchema)
 @application_blueprint.doc(responses=[200, 401, 404])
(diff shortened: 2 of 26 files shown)


api/src/data_migration/load/load_oracle_data_task.py
```diff
@@ -39,6 +39,8 @@
     "tcompetition",
     "tinstructions",
     "tcertificates",
+    "v_user_account",
+    "tuser_profile",
```
chouinar: You can't add these when first adding the tables. The first time we do a load has to be manual as you have to run the setup-foreign-tables command and then verify the load works manually.
```suggestion
```
We'll need a follow-up PR for this later.

api/src/db/models/legacy_mixin/user_mixin.py
```diff
@@ -76,3 +77,32 @@ class TsubscriptionOpportunityMixin:
...
+    user_profile_id: Mapped[int] = mapped_column(BigInteger, primary_key=True)
+    user_account_id: Mapped[int] = mapped_column(BigInteger, nullable=False)
+    profile_duns: Mapped[str]
+    profile_type_id: Mapped[int]
+    profile_name: Mapped[str]
+    title: Mapped[str]
+    is_ebiz_poc: Mapped[str]
+    is_deleted_legacy: Mapped[str]
+    created_date: Mapped[datetime] = mapped_column(nullable=False)
+    last_upd_date: Mapped[datetime] = mapped_column(nullable=False)
```
chouinar: The default type for integers in our staging tables is `BigInteger`, so don't need to specify the type. Don't specify nullable via `mapped_column`, use the python type otherwise the type checker is going to be confused when we use it and you try to null out a field we say isn't nullable.

There are also several columns missing and the order doesn't match what grants.gov has. Here's what we'd want.

```suggestion
    profile_name: Mapped[str | None]
    profile_duns: Mapped[str | None]
    profile_agency_code: Mapped[str | None]
    title: Mapped[str | None]
    is_ebiz_poc: Mapped[str | None]
    is_validate_mpin: Mapped[str | None]
    is_hidden: Mapped[str | None]
    is_deleted_legacy: Mapped[str | None]
    is_default: Mapped[str | None]
    email_preference: Mapped[str | None]
    profile_type_id: Mapped[int]
    user_account_id: Mapped[int]
    created_date: Mapped[datetime | None]
    creator_id: Mapped[str | None]
    last_upd_date: Mapped[datetime | None]
    last_upd_id: Mapped[str |None]
```

If you want to double-check the columns, here is this table's info:

Image

api/tests/src/db/models/factories.py
```diff
@@ -2617,6 +2617,91 @@ def _setup_next_sequence(cls):
         return 1
 
 
+class VuserAccountFactory(BaseFactory):
+    class Meta:
+        abstract = True
+
+    user_account_id = factory.Sequence(lambda n: n)
+    user_id = Generators.UuidObj
+    email = factory.LazyAttribute(lambda o: f"{o.full_name}@example.com")
```
chouinar:
```suggestion
    email = factory.Faker("email")
```

api/tests/src/db/models/factories.py
```diff
@@ -2617,6 +2617,91 @@ def _setup_next_sequence(cls):
...
+    class Meta:
+        abstract = True
+
+    user_profile_id = factory.Sequence(lambda n: n)
+    user_account_id = factory.Sequence(lambda n: n)
+    profile_duns = factory.Faker("sentence")
+    profile_type_id = factory.Faker("random_int", min=1, max=3)
+    profile_name = factory.Faker("name")
+    title = factory.Faker("sentence")
+    is_ebiz_poc = factory.Faker("sentence")
```
chouinar:
```suggestion
    is_ebiz_poc = factory.Faker("yn_boolean")
```

api/src/db/models/staging/user.py
```diff
@@ -28,3 +30,31 @@ class TsubscriptionOpportunityMixin(
     StagingBase, user_mixin.TsubscriptionOpportunityMixin, StagingParamMixin
 ):
     __tablename__ = "tsubscription_opportunity"
+
+
+class VuserAccount(StagingBase, VuserAccountMixin, StagingParamMixin):
+    __tablename__ = "v_user_account"
```
chouinar:
```suggestion
    __tablename__ = "vuser_account"
```

api/src/db/models/legacy_mixin/user_mixin.py
```diff
@@ -76,3 +77,32 @@ class TsubscriptionOpportunityMixin:
...
+    user_id: Mapped[str]
+    email: Mapped[str]
+    first_name: Mapped[str]
+    last_name: Mapped[str]
+    full_name: Mapped[str]
+    is_active: Mapped[str]
+    is_deleted_legacy: Mapped[str]
+    is_duplicate: Mapped[str]
+    created_date: Mapped[datetime] = mapped_column(nullable=False)
+    last_upd_date: Mapped[datetime] = mapped_column(nullable=False)
```
chouinar, line 99: Missing several columns from the view / not the right nullability.

```suggestion
    user_account_id: Mapped[int] = mapped_column(BigInteger, primary_key=True)
    user_id: Mapped[str]
    full_name: Mapped[str | None]
    email: Mapped[str]
    phone_number: Mapped[str | None]
    first_name: Mapped[str | None]
    middle_name: Mapped[str | None]
    last_name: Mapped[str | None]
    is_deleted_legacy: Mapped[str]
    is_duplicate: Mapped[str]
    is_active: Mapped[str]
    is_email_confirm_pending: Mapped[str | None]
    deactivated_date: Mapped[date | None] # TODO - check whether date or datetime
    mobile_number: Mapped[str | None]
    created_date: Mapped[datetime | None]
    creator_id: Mapped[str | None]
    last_upd_date: Mapped[datetime | None]
    last_upd_id: Mapped[str | None]
```

cc @joshtonava - for these kinds of tickets we really want to look this info up ahead of time and put it on the ticket as the info isn't available to anyone without access to the grants.gov Oracle DB.

If you want to double-check that I updated this right, here are the columns for this view.

Image

api/src/data_migration/load/load_oracle_data_task.py
babebe: That makes sense. Removed

api/src/db/models/legacy_mixin/user_mixin.py
babebe: @chouinar @joshtonava . Do we need that last column `last_upd_id` ?

chouinar: Include every column - we need them to match.

api/src/db/models/legacy_mixin/user_mixin.py
```diff
@@ -76,3 +76,46 @@ class TsubscriptionOpportunityMixin:
...
+    email: Mapped[str | None]
+    phone_number: Mapped[str | None]
+    first_name: Mapped[str | None]
+    middle_name: Mapped[str | None]
+    last_name: Mapped[str | None]
+    is_deleted_legacy: Mapped[str]
+    is_duplicate: Mapped[str]
+    is_active: Mapped[str]
+    is_email_confirm_pending: Mapped[str | None]
+    deactivated_date: Mapped[datetime | None]  # TODO - check whether date or datetime
```
chouinar: I checked, this is actually a date.
```suggestion
    deactivated_date: Mapped[date | None]
```

Issue #6814, opened 2025-10-21
Create an ECS task for setting up a cert user


chris-kuryak:
### Summary

We want an ECS task that we can run that takes in the following parameters:

- Cert ID
- Role ID(s)

This task will take that cert ID, and query the `staging.tcertificates` table against the `currentcertid` parameter, erroring if it doesn’t find the ID.

If no role exists for a provided ID, it should also error.

With that cert record, we’ll do the following:

If the cert is already expired, error - no reason to copy an expired cert

If the cert is for an organization, error - we’ll only handle agencies at this time. This is based on whether the `agencyid` parameter is set.

Fetch the agency based on the `agencyid` parameter - error if no agency is found with that agency code in our agency table.

Create a user in our user table
- Should have the `legacy_certificate` user type.

Create a record in our `legacy_certificate` table with the following values:

- `cert_id` - the cert ID passed in / queried with earlier
- `serial_number` - Copied from serial_num - if null in the legacy system, error
- `expiration_date` - Copied from expirationdate - if null, set this to a very far future date (2050?)
- `user_id` - the user above
- `agency_id` - agency ID from the agency fetched above (the UUID, not that agency code)

For the user, we want to add that user to the agency, add them as an agency user and give that agency user the role passed in.

IF the agency has `is_multilevel` marked as `True` then we also want to do the following:

Fetch every agency that starts with the same prefix as the agency (eg. `select * from agency where agency_code LIKE ‘ABC-XYZ-%'` - Make this string by doing the agency code + - so that it doesn’t accidentally get other agencies).

For each of these other subagencies, give them an agency user + the role as well.

### Acceptance criteria

TBD

PR #7021, opened 2025-11-10
[Issue #6814] Create ecs task for setting up cert user

jakobpederson:
## Summary

Fixes / Work for #6814 

## Changes proposed

Added a task/command to create a LegacyCertificate based on a `tcertificate.currentcertid` and some `Role.role_ids`. Creates a `User` of `UserType.LEGACY_CERTIFICATE` and assigns it to the agency associated with the tcertificate along with adding the `AgencyUser` and `AgencyUserRoles`.

```
make cmd args="task setup-cert-user --cert-id 1001 --role-ids <UUID> --role-ids <UUID>"
```

## Context for reviewers

This will take a tcertificate and create an associated user with roles attached to it.


api/src/task/certificates/setup_cert_user_task.py
```diff
@@ -0,0 +1,124 @@
...
+def setup_cert_user(db_session: db.Session, cert_id: str, role_ids: list[str]) -> None:
+    SetupCertUserTask(db_session, cert_id, role_ids).run_task()
+
+
+class SetupCertUserTask(Task):
+
+    def __init__(self, db_session: db.Session, cert_id: str, role_ids: list[str]):
+        super().__init__(db_session)
+        self.cert_id = cert_id
+        self.role_ids = [uuid.UUID(r) for r in role_ids]
```
chouinar: Could we leave these as strings? I think SQLAlchemy doesn't have an issue querying against a UUID column with strings (but double check that).

api/src/task/certificates/setup_cert_user_task.py
```diff
@@ -0,0 +1,124 @@
...
+        self.process_cert_user()
+
+    def process_cert_user(self) -> None:
+        logger.info("setup cert user start")
+        roles = list(
+            self.db_session.scalars(select(Role).where(Role.role_id.in_(self.role_ids))).all()
+        )
+        if invalid_role_uuids := self.get_invalid_role_uuids(roles):
+            logger.warning(f"Invalid role ids: {invalid_role_uuids}")
+            return
```
chouinar: A way you could shortcut this check is just check the length `if len(self.role_ids) != len(roles)`.

api/src/task/certificates/setup_cert_user_task.py
```diff
@@ -0,0 +1,124 @@
...
+            select(staging.certificates.Tcertificates).where(
+                staging.certificates.Tcertificates.currentcertid == self.cert_id
+            )
+        ).one_or_none()
+        if tcertificate is None:
+            logger.warning("Tcertificate not found")
+            return
+        if not tcertificate.expirationdate:
+            logger.warning("Cert is missing expiration date")
+            return
```
chouinar: Missing an expiration date is actually fine / expected. I checked with MicroHealth and they said that if it's null, it's just not got an expiration date. Since we made that field not-null on our table, we should just pick an arbitrary date in the future like `2050-01-01`.

api/src/task/certificates/setup_cert_user_task.py
```diff
@@ -0,0 +1,124 @@
...
+
+        agency = self.db_session.scalar(
+            select(Agency).where(Agency.agency_code == tcertificate.agencyid)
+        )
+        if not agency:
+            logger.warning("Agency not found")
+            return
+        agencies: list[Agency] = [agency]
+        if agency.is_multilevel_agency:
+            search_pattern = f"{agency.agency_code}-%"
```
chouinar, line 183: Could you add an explanation to this that we're doing this to mimic behavior from grants.gov?

api/src/task/certificates/setup_cert_user_task.py
```diff
@@ -0,0 +1,124 @@
...
+            agencies.extend(agency_query_results)
+
+        if (
+            self.db_session.scalars(
+                select(LegacyCertificate.legacy_certificate_id).where(
+                    LegacyCertificate.cert_id == tcertificate.currentcertid
+                )
+            ).one_or_none()
+            is None
+        ):
```
chouinar: Just in the interest of not having a DB query in an if statement, can we do:

```py
        existing_legacy_cert = self.db_session...
        if existing_legacy_cert:
            logger.warning("LegacyCertificate already exists")
            return
```

api/src/task/certificates/setup_cert_user_task.py
```diff
@@ -0,0 +1,124 @@
...
+            user = self.create_user_with_agency_roles(agencies, roles)
+            legacy_certificate = LegacyCertificate(
+                agency=agency,
+                cert_id=tcertificate.currentcertid,
+                expiration_date=tcertificate.expirationdate,
+                serial_number=tcertificate.serial_num,
+                user=user,
+            )
+            self.db_session.add(legacy_certificate)
+            self.db_session.commit()
```
chouinar: Don't use commit directly.

We prefer to do everything in a transaction as it makes rollbacks simpler.

In the `run_task` function, just do:
```
with self.db_session.begin():
     self.process_cert_user()
```

api/src/task/certificates/setup_cert_user_task.py
```diff
@@ -0,0 +1,124 @@
...
+)
+from src.task.ecs_background_task import ecs_background_task
+from src.task.task import Task
+from src.task.task_blueprint import task_blueprint
+
+logger = logging.getLogger(__name__)
+
+
+@task_blueprint.cli.command("setup-cert-user", help="Setup the LegacyCertificate and User")
+@click.option("--cert-id", "-t", help="currentcertid on Tcertificate")
```
chouinar: Probably a bad time to ask this question, but is this cert ID something secretive that we should avoid logging? If so, we might need to do things a bit different as even if you didn't explicitly log it, the command gets logged.

api/tests/src/db/models/factories.py
```diff
@@ -2836,6 +2836,23 @@ class Meta:
...
+class StagingTcertificatesFactory(AbstractStagingFactory):
+    class Meta:
+        model = staging.certificates.Tcertificates
+
+    previouscertid = factory.Sequence(lambda n: f"{n}")
+    currentcertid = factory.Sequence(lambda n: f"{1000 + n}")
+    orgduns = None
+    orgname = None
+    expirationdate = factory.Faker("date_between", start_date="now", end_date="+1y")
+    agencyid = factory.Faker("pystr", min_chars=15, max_chars=15)
```
chouinar: We have a custom provider that will generate an agency code defined already.

```suggestion
    agencyid = factory.Faker("agency_code")
```

api/src/task/certificates/setup_cert_user_task.py
```diff
@@ -0,0 +1,124 @@
...
+                staging.certificates.Tcertificates.currentcertid == self.cert_id
+            )
+        ).one_or_none()
+        if tcertificate is None:
+            logger.warning("Tcertificate not found")
+            return
+        if not tcertificate.expirationdate:
+            logger.warning("Cert is missing expiration date")
+            return
+        if tcertificate.expirationdate <= datetime.datetime.now(datetime.UTC).date():
```
chouinar: Use our datetime util to avoid needing to do all the weird date shenanigans. I think we'd want US Eastern as grants.gov generally sets things up with that (although really if 4~5 hours matters, it's so close to expiration that we won't be able to use it anyways).

```suggestion
        if tcertificate.expirationdate <= datetime_util.get_now_us_eastern_date():
```

api/tests/src/task/certificates/test_setup_cert_user_task.py
```diff
@@ -0,0 +1,222 @@
...
+
+class TestSetupCertUserTask(BaseTestClass):
+    def test_setup_cert_user_raises_error_if_no_role_matching_uuid(
+        self, enable_factory_create, db_session, caplog
+    ):
+        role_id = str(uuid.uuid4())
+        tcertificate = StagingTcertificatesFactory()
+        cert_id = str(tcertificate.currentcertid)
+        SetupCertUserTask(db_session, cert_id, [role_id]).run_task()
+        warning_messages = [r.getMessage() for r in caplog.records if r.levelname == "WARNING"]
```
chouinar: If you wanted to make checking the scenario it hit a bit easier, we could make the `process_cert_user` return a string/enum/whatever and check that in the test. The task itself wouldn't use it, but could check it a bit more directly.

api/src/task/certificates/setup_cert_user_task.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def __init__(self, db_session: db.Session, cert_id: str, role_ids: list[str]):
+        super().__init__(db_session)
+        self.cert_id = cert_id
+        self.role_ids = [uuid.UUID(r) for r in role_ids]
+
+    def run_task(self) -> None:
+        self.process_cert_user()
+
+    def process_cert_user(self) -> None:
```
chouinar: Any way to break this function up a bit, it's doing a lot (well, everything).

api/tests/src/task/certificates/test_setup_cert_user_task.py
```diff
@@ -0,0 +1,222 @@
...
+        role = RoleFactory.create(is_agency_role=True)
+        agency = AgencyFactory(agency_code=f"XYZ-{uuid.uuid4()}", is_multilevel_agency=False)
+        tcertificate = StagingTcertificatesFactory(
+            expirationdate=date(2023, 12, 31), agencyid=agency.agency_code
+        )
+        cert_id = str(tcertificate.currentcertid)
+        SetupCertUserTask(db_session, cert_id, [str(role.role_id)]).run_task()
+        warning_messages = [r.getMessage() for r in caplog.records if r.levelname == "WARNING"]
+        msg = "Cert is expired"
+        assert msg in warning_messages
```
chouinar: Just do the check in one line, it's easier to read.
```suggestion
        assert "Cert is expired" in warning_messages
```

api/src/task/certificates/setup_cert_user_task.py
jakobpederson: They do need to be converted to UUIDs from what I can tell, but I moved the conversion to the query.

api/src/db/models/staging/certificates.py
```diff
@@ -1,6 +1,12 @@
...
+from sqlalchemy.orm import Mapped, mapped_column
+
 from src.db.models.legacy_mixin.certificates_mixin import TcertificatesMixin
 from src.db.models.staging.staging_base import StagingBase, StagingParamMixin
 
 
 class Tcertificates(StagingBase, TcertificatesMixin, StagingParamMixin):
     __tablename__ = "tcertificates"
+
+    tcertificates_id: Mapped[uuid.UUID] = mapped_column(default=uuid.uuid4)
```
chouinar: You'll need to make this column nullable otherwise when the migration runs non-locally, it would error trying to add the column.

I think with the way our copy logic works, we also need to have the ID have a server default to be set. `default` is the ORM default when SQLAlchemy is handling things, but `server_default` puts it as a default on the table itself. Since we don't use SQLAlchemy for copying from the foreign data wrapper, we need the DB to set the default. I don't know the exact way, but something like `server_default = sqlalchemy.sql.functions.gen_random_uuid()` might work? Exactly how we call the UUID function, I'm not 100% certain on.

api/src/task/certificates/setup_cert_user_task.py
```diff
@@ -0,0 +1,169 @@
...
+class SetupCertUserTask(Task):
+
+    def __init__(self, db_session: db.Session, tcertificates_id: str, role_ids: list[str]):
+        super().__init__(db_session)
+        self.tcertificates_id = tcertificates_id
+        self.role_ids = role_ids
+        self.valid_expiration_date = FUTURE_DATE
+        self.top_agency: Agency | None = None
+
+    def run_task(self) -> SetupCertUserTaskStatus:  # type: ignore [override]
```
chouinar: Don't change the type of this function. Do this pattern:

```py
def run_task(self) -> None:
     with db_session.begin():
          self.setup_cert()

def setup_cert(self) -> SetupCertUserTaskStatus:
    ...
```
That also makes it so we don't have to indent all of this.

api/src/task/certificates/setup_cert_user_task.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def is_existing_certificate(self, tcertificate: staging.certificates.Tcertificates) -> bool:
+        existing_tcertificate = self.db_session.scalars(
+            select(LegacyCertificate.legacy_certificate_id).where(
+                LegacyCertificate.cert_id == tcertificate.currentcertid
+            )
+        ).one_or_none()
+        if existing_tcertificate:
+            return True
+        return False
```
chouinar:
```suggestion
        return existing_tcertificate is not None
```

api/src/task/certificates/setup_cert_user_task.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+class SetupCertUserTask(Task):
+
+    def __init__(self, db_session: db.Session, tcertificates_id: str, role_ids: list[str]):
+        super().__init__(db_session)
+        self.tcertificates_id = tcertificates_id
+        self.role_ids = role_ids
+        self.valid_expiration_date = FUTURE_DATE
+        self.top_agency: Agency | None = None
```
chouinar: We probably don't want to set these in this way, global (as far as this script is concerned) paramaters that change are always an iffy implementation detail.

Just pass them around between the functions, having one function call cause a side-effect that affects another one is one we should avoid.

For something like the agency, just have the get agencies function return `tuple[Agency, list[Agency]]`

api/src/task/certificates/setup_cert_user_task.py
```diff
@@ -0,0 +1,169 @@
...
+        return user
+
+    def get_roles(self) -> list[Role] | None:
+        roles = list(
+            self.db_session.scalars(
+                select(Role).where(Role.role_id.in_([uuid.UUID(r) for r in self.role_ids]))
+            ).all()
+        )
+        if len(self.role_ids) != len(roles):
+            logger.warning("Invalid role ids")
```
chouinar: For log messages, let's assume you were running this task and saw this message, what would you do with that information? Which role ID was invalid?

We have a very simple/flexible way of adding more info, just put it in the extra:

```suggestion
            logger.warning("Invalid role ids", extra={"found_role_ids": [role.role_id for role in roles]})
```

api/src/task/certificates/setup_cert_user_task.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def create_user_with_agency_roles(self, agencies: list[Agency], roles: list[Role]) -> User:
+        user = User(user_type=UserType.LEGACY_CERTIFICATE)
+        self.db_session.add(user)
+        for a in agencies:
+            agency_user = AgencyUser(user=user, agency=a)
+            self.db_session.add(agency_user)
+            agency_roles = [AgencyUserRole(agency_user=agency_user, role=r) for r in roles]
+            self.db_session.add_all(agency_roles)
+        return user
```
chouinar: Let's assume I wanted to look in the DB to find the user that was just created, how would I do that?

I could cross-reference the cert but that's a pain, log the user's IDs, make it so if we need to debug something going wrong, we have a direct way to get to that record.

Note that because the IDs of any `default` field don't get set until we push them to the DB, have to manually set them to get the IDs if we want to log (alternatively `db_session.flush` but that's clunkier).

```suggestion
        user = User(user_id=uuid.uuid4(), user_type=UserType.LEGACY_CERTIFICATE)
        self.db_session.add(user)
        
        log_extra = {"user_id": user.user_id}
        logger.info("Created legacy cert user", extra=log_extra)
        
        for agency in agencies:
            agency_user = AgencyUser(user=user, agency=agency)
            self.db_session.add(agency_user)

            agency_roles = [AgencyUserRole(agency_user=agency_user, role=r) for r in roles]
            self.db_session.add_all(agency_roles)
            
            logger.info("Added user to agency", extra=log_extra | {"agency_code": agency.agency_code, "role_ids": [r.role_id for r in roles]})
            
        return user
```

api/src/task/certificates/setup_cert_user_task.py
```diff
@@ -0,0 +1,169 @@
...
+        user = self.create_user_with_agency_roles(agencies, roles)
+        legacy_certificate = LegacyCertificate(
+            agency=self.top_agency,
+            cert_id=tcertificate.currentcertid,
+            expiration_date=self.valid_expiration_date,
+            serial_number=tcertificate.serial_num,
+            user=user,
+        )
+        self.db_session.add(legacy_certificate)
+
```
chouinar, line 121: Always helps to log what we just made.

```suggestion
        legacy_certificate = LegacyCertificate(
            legacy_certificate_id=uuid.uuid4(),
            agency=self.top_agency,
            cert_id=tcertificate.currentcertid,
            expiration_date=self.valid_expiration_date,
            serial_number=tcertificate.serial_num,
            user=user,
        )
        self.db_session.add(legacy_certificate)
        
        logger.info("Created legacy certificate", extra={"legacy_certificate_id": legacy_certificate.legacy_certificate_id, "user_id": user.user_id, "agency_code": agency.agency_code})

```

api/src/task/certificates/setup_cert_user_task.py
jakobpederson: I like that a lot.

api/src/db/migrations/versions/2025_11_17_add_staging_tcertificate_uuid.py
```diff
@@ -0,0 +1,37 @@
...
+        "tcertificates",
+        sa.Column(
+            "tcertificates_id",
+            sa.Uuid(),
+            server_default=sa.text("gen_random_uuid()"),
+            nullable=True,
+        ),
+        schema="staging",
+    )
+    # ### end Alembic commands ###
```
chouinar, line 38: I didn't finish typing a comment here from my last PR - we'll also want to backfill the UUIDs in this column.

That should be doable with something like `update staging.tcertificates set tcertificates_id=gen_random_uuid()` - but we'll want to double-check that that gives each one a different ID.

api/src/db/models/staging/certificates.py
```diff
@@ -1,6 +1,15 @@
...
 from src.db.models.legacy_mixin.certificates_mixin import TcertificatesMixin
 from src.db.models.staging.staging_base import StagingBase, StagingParamMixin
 
 
 class Tcertificates(StagingBase, TcertificatesMixin, StagingParamMixin):
     __tablename__ = "tcertificates"
+
+    tcertificates_id: Mapped[uuid.UUID | None] = mapped_column(
+        Uuid(as_uuid=True), server_default=text("gen_random_uuid()")
+    )
```
chouinar, line 16: It might not matter, but for consistency could we use the UUID type from `from sqlalchemy.dialects.postgresql`- that's what we've done elsewhere / should be the default anyways. It probably is just an alias, but to be safe/ consistent.

Also, we should make the column unique to be safe + adds an index for us to query against.

```suggestion
    tcertificates_id: Mapped[uuid.UUID | None] = mapped_column(
        UUID, server_default=text("gen_random_uuid()"), unique=True
    )
```

api/src/task/certificates/setup_cert_user_task.py
```diff
@@ -0,0 +1,196 @@
...
+    Role,
+    User,
+)
+from src.task.ecs_background_task import ecs_background_task
+from src.task.task import Task
+from src.task.task_blueprint import task_blueprint
+
+logger = logging.getLogger(__name__)
+
+FUTURE_DATE = datetime_util.make_timezone_aware(datetime(2050, 1, 1), "US/Eastern").date()
```
chouinar: Just make the date directly - making a timezone aware datetime object and then making it a date isn't going to do anything. Dates themselves aren't (generally) timezone aware. The `.date()` function is `return date(self._year, self._month, self._day)` - so the timezone gets ignored.

```suggestion
FUTURE_DATE = date(2050, 1, 1)
```

api/src/task/certificates/setup_cert_user_task.py
```diff
@@ -0,0 +1,196 @@
...
+        if self.is_existing_certificate(tcertificate):
+            logger.warning("LegacyCertificate already exists")
+            return SetupCertUserTaskStatus.LEGACY_CERTIFICATE_ALREADY_EXISTS
+
+        agency, related_agencies = self.get_agencies(tcertificate)
+        if agency is None:
+            return SetupCertUserTaskStatus.AGENCY_NOT_FOUND
+
+        else:
+            self.process_cert_user(roles, tcertificate, agency, related_agencies)
```
chouinar: Don't store the date on the class if we calculate it in the function that calls the one place it uses it. Just pass that into that function.

```suggestion
        valid_expiration_date = tcertificate.expirationdate or FUTURE_DATE
        if valid_expiration_date <= datetime_util.get_now_us_eastern_date():
            logger.warning("Cert is expired")
            return SetupCertUserTaskStatus.TCERTIFICATE_IS_EXPIRED
        if self.is_existing_certificate(tcertificate):
            logger.warning("LegacyCertificate already exists")
            return SetupCertUserTaskStatus.LEGACY_CERTIFICATE_ALREADY_EXISTS

        agency, related_agencies = self.get_agencies(tcertificate)
        if agency is None:
            return SetupCertUserTaskStatus.AGENCY_NOT_FOUND

        else:
            self.process_cert_user(roles, tcertificate, agency, related_agencies, valid_expiration_date)
```

api/src/db/migrations/versions/2025_11_17_add_staging_tcertificate_uuid.py
jakobpederson: I ran the command a few times in the console and it gave me a different uuid each time
```
dbs.execute(text("SELECT gen_random_uuid(), gen_random_uuid(), gen_random_uuid()")).one()
(UUID('c03706e5-5587-4016-8564-341fc5202932'), UUID('9db03967-441c-4b77-b3c7-a1366cc31b1d'), UUID('28623907-2758-46a5-8947-cee2cf090ce4'))
```

chouinar: It's not about running it individually to see if it gets different results, let's assume there are 1000 records in the table.

Does `UPDATE staging.tcertificates SET tcertificates_id = gen_random_uuid()` get 1000 different results in that single query, or does it generate a single UUID and try to set it for all of them? Easiest way to check, add a few rows to your DB before running this migration and see if it works.

jakobpederson:
I added some tcertificates to the db, and made some of the `tcertificates_id` `None` and then ran the migration SQL command to update the values and they do appear to be unique:
```
>>> dbs.query(staging.certificates.Tcertificates.tcertificates_id).all()
[
│   (UUID('d140df1d-b6a6-4abe-ab54-124b1ee596bb'),),
│   (UUID('bc2f8bb8-54cd-439b-8a87-e6f96060f47b'),),
│   (UUID('668f3425-c43f-4ba5-bed3-86e56ea8947b'),),
│   (UUID('154ad3fe-0f56-43ce-a770-2c8491ea2988'),),
│   (UUID('353bdb5c-d116-413d-ac93-68ef8f2051e9'),),
│   (None,),
│   (None,),
│   (None,)
]
>>> dbs.query(staging.certificates.Tcertificates.tcertificates_id).all()
[
│   (UUID('ef0a2d05-a00d-46c2-986f-cb04a56d31d5'),),
│   (UUID('ae270cff-c337-4e68-b547-1e69d22de050'),),
│   (UUID('f35d0a62-2be6-40d9-b5d5-c72751832069'),),
│   (UUID('a4b7e87f-28bd-4ba8-ab1e-e1cf4b412fcb'),),
│   (UUID('110926ea-82c7-4cc5-b4a0-e2478c7af9bc'),),
│   (UUID('2b0ac6a4-6e7c-4d92-845a-6659500f3f99'),),
│   (UUID('d9c4ba03-60bc-44b0-b85a-ff7106eb7351'),),
│   (UUID('50d0d25f-bdac-4227-810d-d920fd25aa99'),)
]
```

Issue #6856, opened 2025-10-27
Reorganize workspace -> dashboard navigation and page


doug-s-nava:
### Summary

As a first step to reorganizing our nav items and page routing we need to:

- change the name of the "workspace" page to "dashboard" everywhere that this appears
- change routing of "workspace" page from /user/workspace to /dashboard
- move the "workspace / dashboard" nav link from the user dropdown menu to the "workspace" dropdown menu

Figma: https://www.figma.com/design/zPlDWcYM396szw2yL3UXvk/Simpler-Grants-2025?node-id=6004-80101&m=dev

### Acceptance criteria

- page that was previously available as the "User Workspace" at /user/workspace is now available at /dashboard
- text changes to the page TBD
- "workspace" nav item in user dropdown is removed
- "Activity dashboard" nav item is added to workspace dropdown linking to /dashboard

doug-s-nava:
note that we'll want to look at this once work on the workspace page is merged, in order to minimize complex merge conflicts

* https://github.com/HHS/simpler-grants-gov/pull/6863
* github.com/HHS/simpler-grants-gov/issues/6792

ErinPattisonNava:
@eric-serrato95 ready for UAT!

eric-serrato95:
@ErinPattisonNava UAT complete, moving to done

Issue #6768, opened 2025-10-16
Applications sections in workspace


eric-serrato95:
### Summary

we want to have a section to highlight the ongoing applications that an org is undergoing in the workspace section 

### Acceptance criteria

- [ ] A section for applications is available in workspaces alongisde saved opportunities and saved searches 
- [ ] A link is provided to take a user into the list of applications 
- [ ] A link is added in the Navigation to point to the new applications page.  While the rest of the work is behind the feature flag for the user admin work, the link on the applications page should appear regardless of feature flag.

ErinPattisonNava:
@eric-serrato95 Just confirming that this is the right [figma link](https://www.figma.com/design/zPlDWcYM396szw2yL3UXvk/Simpler-Grants-2025?node-id=5756-214163&t=URvrtrvdVygtUSG1-1)?

eric-serrato95:
@ErinPattisonNava yes that is the right figma link!

ErinPattisonNava:
@eric-serrato95 Ready for UAT!

eric-serrato95:
@ErinPattisonNava UAT complete, moving to done!

PR #7028, opened 2025-11-11
moving the user/workspace page to dashboard (including text reference…

ErinPattisonNava:
…s) and adding applications links

## Summary

Fixes / Work for #6856 #6768 

## Changes proposed

* Rename the Workspace page to Activity Dashboard at `/dashboard`
* Add a section to point to the new Applications page
* List both pages appropriately in the header
* Remove caching for the `/applications` page

## Context for reviewers

We want the new Workspace page to be called the Activity Dashboard, and the Applications page to be integrated with it.  The dashboard is behind the `userAdminOff` feature flag, so you'll need to make sure it is disabled (?!?)  in your environment.  

I did no move/rename the workspace components directory because it seemed to have a broader scope than just this page. 

There is one reference to "Workspace" on the modal after saving a search query.  I left this reference as the link goes directly to the saved searches page, which is under the "Workspace" header item.

## Validation steps

- [ ] The workspace page can now be found at `/dashboard`
- [ ] The title of the new tab is the Activity Dashboard 
- [ ] The breadcrumbs (which are likely to eventually be removed) list the Activity Dashboard.
- [ ] The title of the new page is "Welcome to your activity dashboard."  Design requested we remove the color around Welcome.
- [ ] There is a new Applications card to the left of the "Saved queries" card, and the link should take you to `/applications`.
- [ ] In the header the entry under their email to "Workspace" has been removed.
- [ ] The link to "Activity Dashboard" now appears in the Workspace header when the feature flag is disabled and the links are in alphabetical order
- [ ] The link to "Applications" now appears in the Workspace header regardless of feature flag status


frontend/src/components/workspace/UserInviteForm.tsx
```diff
@@ -3,7 +3,7 @@
 import {
   inviteUserAction,
   OrganizationInviteValidationErrors,
-} from "src/app/[locale]/(base)/user/workspace/actions";
+} from "src/app/[locale]/(base)/dashboard/actions";
```
doug-s-nava, line 6: since the user invite form will be on the manage users page going forward, a heads up to @myduong-navapbc to handle moving this actions file again in your PR

frontend/next.config.js
```diff
@@ -117,6 +117,16 @@ const headers = [
...
+  // don't cache the user specific page for applications
+  {
+    source: "/applications",
+    headers: [
+      {
+        key: "Cache-Control",
+        value: "no-store, must-revalidate",
+      },
+    ],
+  },
```
doug-s-nava, line 129: it seems like this case would be handled by the rule below that looks for session cookies, but I also see similar entries for the saved-searches and saved-opportunities pages, which suggests that maybe it's not. I think we need a survey of this stuff and to document what the proper cache posture is for our authenticated pages. We may need to add some rules to other pages as well (workspace, manage users, etc.)

frontend/src/components/Header.tsx
```diff
@@ -122,6 +123,17 @@ const NavLinks = ({
     }
 
     const workspaceSubNavs = [];
+
```
doug-s-nava, line 126: I know it's not part of these tickets in particular, but wondering how urgent it is for us to update the name and path of the current "account" page. Currently the link is "My account" and it takes the user to "user/account" - looks like in the designs the new set up is a "Profile" link that goes to "/settings"? Want to make sure this change isn't necessary immediately, and that we have the requirements for the change in a ticket somewhere

doug-s-nava (approved):
✅ ✅ ✅ ✅

frontend/next.config.js
ErinPattisonNava: Ticket: https://github.com/HHS/simpler-grants-gov/issues/7040 on the platform workstream.

frontend/src/components/Header.tsx
ErinPattisonNava: Great catch, that had slipped through the cracks.  Added a ticket [here](https://github.com/HHS/simpler-grants-gov/issues/7044)

PR #7062, opened 2025-11-14
[Issue #7061] Create an ECS task that can generate opportunities for each form

### LINKED ISSUE #7061: Create a utility for setting up opportunities for each form in a given environment
### Summary

To help speedup testing, we want a task that will automatically generate an opportunity for every single form we have in our system.

### Acceptance criteria

- [ ] A task can be manually run to setup opportunities/competitions in an environment for each form
- [ ] Task does not run in production, but can run in other environments

chouinar:
## Summary
Fixes #7061 

## Changes proposed
Make an ECS task that for each form makes sure an opportunity exists

Also always will generate a new opportunity with all forms

## Context for reviewers
Setting up and keeping opportunities up-to-date with every form is a very tedious process involving a lot of DB SQL inserts. To save time, and make the process something that more people can do without needing much knowledge, made a task that just does all of that for us.

In the future it might make sense to add some flexibility to this job to create opportunities with a subset of forms, but the idea was to get something really quick put together and refine it once we have it.

## Validation steps
Can run `make cmd args="task build-automatic-opportunities"` to run the task locally.

If you want these in search, run `make populate-search-opportunities populate-search-agencies`.

Once you've done that, you can easily see all of them by searching for `SGG`:
Screenshot 2025-11-14 at 2 28 37 PM

Here's what the opportunity page looks like:
Screenshot 2025-11-14 at 2 28 52 PM
Screenshot 2025-11-14 at 2 28 55 PM

### DIFF
--- a/api/src/task/__init__.py
+++ b/api/src/task/__init__.py
@@ -11,6 +11,7 @@ import src.task.generate_internal_token  # noqa: F401 isort:skip
 import src.task.forms.update_form_task  # noqa: F401 isort:skip
 import src.task.forms.list_forms_task  # noqa: F401 isort:skip
 import src.task.opportunities.generate_opportunity_sql  # noqa: F401 isort:skip
+import src.task.opportunities.build_automatic_opportunities  # noqa: F401 isort:skip
 import src.cli.xml_generation_cli  # noqa: F401 isort:skip
 
 __all__ = ["task_blueprint"]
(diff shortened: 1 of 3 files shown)


jakobpederson (approved):
LGTM!

I was curious though, what was the reason you used frozenset? I've never seen that before.

chouinar:
> LGTM!
> 
> I was curious though, what was the reason you used frozenset? I've never seen that before.

I had sets of sets of strings. The set class isn't hashable because it's modifiable (generally modifiable collections aren't hashable because if you add/remove from them their hash gets messed up). A frozenset is a set that once created, can't be modified, so can be hashable. It's basically one of the main purposes of it.

Issue #6007, opened 2025-08-20
Update `$theme-link-[state-]color` settings


andycochran:
### Summary

We recently darkened `$theme-link-color` to `mint-60` (see #5316). However, we should have also adjusted the hover state and active state. Currently on prod there is no visual feedback for the hover state of links. 

### Acceptance criteria

- [x] `$theme-link-hover-color` = `mint-70`
- [x] `$theme-link-active-color` = `mint-80`
- [x] Figma template has been updated accordingly

Issue #5316, opened 2025-06-11
Improve link/button color contrast


andycochran:
### Summary

In order for links and buttons to pass WCAG 2.0 AA contrast requirements, they need to have what USWDS calls a 50+ [magic numbers](https://designsystem.digital.gov/design-tokens/color/overview/#magic-number-2), which is the difference in the color grades of the background and text. 

Our links use the `mint-50` system color token, which is fine on a white background (white being a color grade of `0`). However, when our links appear on the light grey background we're using, they do not have enough contrast (`gray-warm-5` and `mint-50` have a color grade difference of only 45). This same problems exists for other background colors like `yellow-5v` for alerts. 

We should use a darker default link color with a grade of 60, so that we can place our links on any background color with a `5` or `10` grade. 

This issue also exists for our `unstyled` and `outline` button variants. 

### Acceptance criteria

- [ ] Links meet WCAG 2.0 AA contrast requirements on light gray backgrounds and in Alert components
- [ ] Unstyled and outline button variants meets WCAG 2.0 AA contrast requirements on Alert components

eric-serrato95:
@doug-s-nava @andycochran UAT ✅ contrast passes AA 4.5:1 requirement

andycochran:
cc @crystabelrangel, noting that this issue changed our link color to`mint-60` to pass WCAG AA and we should make sure that Figma is updated to match. @eric-serrato95, think we need a ticket for Figma?

andycochran:
D'oh! I just realized that we darkened `$theme-link-color` to `mint-60`, but did not also adjust the hover state (`$theme-link-hover-color`) which is also `mint-60` (should prob be `mint-70`) or the active state (`$theme-link-active-color`, prob `mint-80`). The visited state is fine (`violet-60` has enough contrast). The reverse colors should also be fine, so long as we keep them on backgrounds that are dark enough (≥ `-60`). 

https://github.com/HHS/simpler-grants-gov/blob/e2db7ef9ecc74b7b6f208e60b5cc27f7ffb92191/frontend/src/styles/_uswds-theme.scss#L215-L221

Currently there is not visual feedback for the hover state in prod. I'll make a subsequent ticket for this…

andycochran:
https://github.com/HHS/simpler-grants-gov/issues/6007 (Sorry I missed this while we were pairing, Doug.)

PR #7065, opened 2025-11-14
[Issue #6007] Improve link hover and active colors

andycochran:
## Summary

Work for #6007 

## Changes proposed

This is a very small USWDS theme change that darkens the color of text links' `hover` and `active` states. 

## Context for reviewers

#5316 darkened the `$theme-link-color` to from `mint-50` to `mint-60` so that the default link color is accessible on the light gray background we're using. We forgot to also darken the hover and active states. 

Note: This PR will not close the issue, as Figma still needs to be updated to reflect this change. 

## Validation steps

Load the app and view a text link: 
- On `main`/prod there is no color change on hover, tho it does get slightly darker on click
- On this branch the color gets slightly darker on hover, and even darker still on click (it's subtle)


andycochran:
Good Q, @ErinPattisonNava. I think pa11y doesn't check the dynamic hover and active states. It might only catch the default color of links.

andycochran:
Oh. Also, they were already dark enough to pass AA contrast requirements. There was just no visual feedback.

Issue #6953, opened 2025-11-04
Hide Legacy Users Endpoint


joshtonava:
### Summary

Implement the API endpoint that allows organizations to hide legacy users from discovery results.

### Tech Notes
- [Hide Legacy Users Endpoint](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2174681171/Tech+Spec+Make+it+easy+for+Organizations+to+invite+users+from+existing+system#Hide-Legacy-User)

### Acceptance criteria

- Hidden users no longer appear in discovery results
- unit tests created

eric-serrato95:
@joshtonava 
- Add error responses on the techspech

PR #7068, opened 2025-11-17
[Issue #6953] Hide legacy users endpoint

babebe:
## Summary

Fixes / Work for #6953  

## Changes proposed
Add new endpoint `POST /v1/organizations/:organization_id/legacy-users/ignore`

- Normalizes email input (strip().lower()).
- Verifies user has access to the organization with `MANAGE_ORG_MEMBERS` privilege.
- Only inserts a record when the user has not been previously ignored.
- Checks for existing ignored-user entries to ensure the operation can be safely repeated without creating duplicates.
- Added tests


api/src/api/organizations_v1/organization_schemas.py
```diff
@@ -321,3 +321,15 @@ class OrganizationInvitationListResponseSchema(AbstractResponseSchema):
...
         metadata={"description": "List of organization invitations"},
     )
+
+
+class OrganizationIgnoreLegacyUserRequestSchema(Schema):
+    """Schema for handling requests to ignore a legacy user within an organization."""
+
+    email = fields.String(
+        required=True, metadata={"description": "Email address of the legacy user to ignore"}
+    )
```
chouinar, line 333: We should validate that the email is formatted properly:
```suggestion
    email = fields.String(
        required=True, metadata={"description": "Email address of the legacy user to ignore"}, validate=[validators.Email()]
    )
```

api/src/services/organizations_v1/ignore_legacy_user_organization.py
```diff
@@ -0,0 +1,51 @@
...
+    ).scalar_one_or_none()
+    if existing_record:
+        logger.info(
+            "User is already ignored",
+            extra={
+                "organization_id": organization_id,
+                "ignored_by_user_id": user.user_id,
+            },
+        )
+        return
```
chouinar: If they are already ignored, should we have some sort of 4xx error? The frontend/anyone calling the endpoint would probably want to know that they have already been ignored and that nothing new was done. Generally if an endpoint does not do what the user asked, we should tell the user why.

Alternatively, have the response from this endpoint return the `ignored_legacy_organization_user_id` and we just return this object in that case if we don't want a 4xx scenario.

api/tests/src/api/organizations_v1/test_ignore_legacy_user_organization.py
```diff
@@ -0,0 +1,121 @@
...
+
+    @pytest.fixture
+    def user_org_a(self, db_session):
+        user, org, token = create_user_in_org(db_session, privileges=[Privilege.MANAGE_ORG_MEMBERS])
+        return user, org, token
+
+    @pytest.fixture(autouse=True)
+    def clear_data(self, db_session):
+        cascade_delete_from_db_table(db_session, IgnoredLegacyOrganizationUser)
+        cascade_delete_from_db_table(db_session, Organization)
```
chouinar: I don't believe the tests you have would need data to be cleared before running. Except for cases where we delete data because we have a script or something that does a select-all on a table, don't clean up prior data. It makes our tests slower as these queries are immensely complex.

api/tests/src/api/organizations_v1/test_ignore_legacy_user_organization.py
```diff
@@ -0,0 +1,121 @@
...
+        """Test successfull addition of user to ignore table"""
+        user, org, token = user_org_a
+        resp = client.post(
+            f"/v1/organizations/{org.organization_id}/legacy-users/ignore",
+            headers={"X-SGG-Token": token},
+            json={"email": "test@gmail.com"},
+        )
+        assert resp.status_code == 200
+
+        ignored_users = db_session.query(IgnoredLegacyOrganizationUser).all()
```
chouinar: In tests, don't do a select-all, fetch the thing you want to see if it exists, query for what you want. In this case you have an email, query against that OR instead of having the endpoint return nothing, we return the `ignored_legacy_organization_user_id` from the table so we have an exact value to query against.

babebe: Updated to query for that particular record.

api/tests/src/api/organizations_v1/test_ignore_legacy_user_organization.py
babebe: Removed

api/src/services/organizations_v1/ignore_legacy_user_organization.py
babebe: Raising a 400 error if previously ignored

api/src/api/organizations_v1/organization_schemas.py
babebe: Updated this. Although this raises a "500" error when invalid, which is not really a server error.

api/src/api/organizations_v1/organization_schemas.py
```diff
@@ -321,3 +321,17 @@ class OrganizationInvitationListResponseSchema(AbstractResponseSchema):
...
     )
+
+
+class OrganizationIgnoreLegacyUserRequestSchema(Schema):
+    """Schema for handling requests to ignore a legacy user within an organization."""
+
+    email = fields.String(
+        required=True,
+        metadata={"description": "Email address of the legacy user to ignore"},
+        validate=[validators.Email()],
```
joshtonava, line 332: **Question:** should we add a test with invalid email or is this overkill just testing the framework

api/src/services/organizations_v1/ignore_legacy_user_organization.py
```diff
@@ -0,0 +1,54 @@
...
+
+from sqlalchemy import select
+
+from src.adapters import db
+from src.api.route_utils import raise_flask_error
+from src.db.models.entity_models import IgnoredLegacyOrganizationUser
+from src.db.models.user_models import User
+from src.services.organizations_v1.list_organization_invitations import (
+    get_organization_and_verify_access,
+)
```
joshtonava, line 12: Note:

I just realized we have two functions for get_organization_and_verify_access.

One in get_organization.py and one in list_organization_invitations

joshtonava (dismissed):

api/src/api/organizations_v1/organization_schemas.py
babebe: Added test

Issue #6951, opened 2025-11-04
Domain Model for Ignored Legacy Users


joshtonava:
### Summary

Create the domain model that tracks which legacy users an organization has chosen to hide from discovery results.

### Tech Notes
- [Domain-Model:IgnoredLegacyOrganizationUser](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2174681171/Tech+Spec+Make+it+easy+for+Organizations+to+invite+users+from+existing+system#Domain-Model%3A-IgnoredLegacyOrganizationUser)

### Acceptance criteria

- Add domain model to track which legacy users an organization has chosen to hide
- migration created
- factories created

PR #7069, opened 2025-11-17
[Issue #6951] New migration for table IgnoredLegacyOrganizationUser

babebe:
## Summary

Fixes / Work for #6951  

## Changes proposed
Update model to automatically generate a new UUID for `ignored_legacy_organization_user`
Generate migration file


api/src/db/migrations/versions/2025_11_17_add_ignored_legacy_organization_user_.py
```diff
@@ -1,16 +1,16 @@
...
+Create Date: 2025-11-17 06:12:31.841617
 
 """
 
 import sqlalchemy as sa
 from alembic import op
 
 # revision identifiers, used by Alembic.
-revision = "19e80334aaae"
+revision = "53eee8fd684e"
```
chouinar, line 13: We can't change migrations once merged, but also this only changes the revision, what is it we're trying to do here?

Adding a default value doesn't require a migration as it's a SQLAlchemy default - not a DB default.

babebe:
Change not needed

Issue #6776, opened 2025-10-17
build test user login button and hardcoded application link


doug-s-nava:
### Summary

In order to allow a local user to easily access a test application, we will build a way for the user to log in with a test user with one click, and after logging in, access a test application with one click.

### Login UI

This can be ugly. I think we have two options and we should do whatever will be simplest to implement and use:

- a nav / sub nav in the header that will appear in local environments containing links for each test user
- a select with options for each test user in the header but separate from the nav - maybe next to the grants logo

To populate this we will need to call out to the new endpoint created in https://github.com/HHS/simpler-grants-gov/issues/6780 to fetch a list of all test users along with JWTs. On the next js side, after fetching from the API, we will need to create client side tokens for each of these JWTs

When a user selects a test user, the client token for that user will need to be set on the cookie to be sent back and forth in all requests and responses to mimic a logged in state.

### Test application link

When logged in as the "test form user" a link should be added to the user nav dropdown titled "Test application" which will take the user to the the test application set up for that user. If necessary this url can be hardcoded using the application id set in the seed changes from https://github.com/HHS/simpler-grants-gov/issues/6779. Alternatively a call could be made to an API endpoint to fetch the user's applications.

### Acceptance criteria

- [ ] test user quick login UI is visible in local environments
- [ ] when logged in as "test form user", link to their test application is available from user nav drop down
- [ ] it is not possible to access this functionality in a deployed env

PR #7073, opened 2025-11-17
[Issue #6776] local form dev tools

doug-s-nava:
## Summary

Fixes for #6776

## Changes proposed

For local environments ONLY:

* adds a test account selector to use for logging in to test accounts locally
* adds a hardcoded test application link to user drop down when logged in as the test `many_app_user`

## Context for reviewers

Since these features allow users to at least see some data for EVERY user in the environment, and possibly would allow users to log in as other users without entering any credentials, these features will only be available when:

* the application is started with the `ENVIRONMENT` environment variable set to 'local'
* the API url includes 'localhost'
* the list of fetched users includes fewer than 500 users - following the backend's lead here for extra security

These conditions should prevent the features from ever being live in a deployed environment.

## Validation steps

1. start a local server on this branch with `npm run local` or `ENVIRONMENT=local npm run dev`
2. visit http://localhost:3000
3. _VERIFY_: you see a select in the header containing test users
4. select `many_app_user`
5. _VERIFY_: the select is disabled while logging in, then the app reflects that the user is logged in
6. click the user menu
7. _VERIFY_: contains a link to `Test application`
8. click the link
9. _VERIFY_: takes you to an application page showing all forms that exist in the system
10. click a different name in the user login dropdown
11. _VERIFY_: the user dropdown no longer contains the `Test application` link

### Screenshot

Screenshot 2025-11-17 at 1 39 40 PM

### DIFF
--- a/frontend/src/components/Layout.tsx
+++ b/frontend/src/components/Layout.tsx
@@ -26,6 +26,7 @@ export default async function Layout({ children, locale }: Props) {
   // to populate local user quick login dropdown
   if (environment.LOCAL_DEV) {
     try {
+      // if this returns more than 500 users we likely are in a deployed env and will disable the feature
       testUsers = await getTestUsers();
     } catch (e) {
       console.error("unable to fetch test users, oh well", e);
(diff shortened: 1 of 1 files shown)


myduong-navapbc:
I went through all of the validation checkpoints without any issue. This will definitely take some of the sting out of locally testing forms

PR #7094, opened 2025-11-17
[ISSUE #6744]: Active Users Table

myduong-navapbc:
## Summary

Work for #6744 

This is only for the table
The action of editing will come in the next PR

## Changes proposed

This PR displays the users who have received an invitation
Unit test will come in after the page is put together

## Validation steps

- to run locally, you would need to `run npm run build && npm start
- navigate to feature flags (http://localhost:3000/dev/feature-flags)
- make sure that `manageUsersOff` is set to `disabled`
- sign into the app as "two_org_user"
- navigate to Workspace > Activity Dashboard
- scroll to the bottom and "View Organization Details"
- click "Manage Users"
- VERIFY that the table is shown

### DIFF
--- a/frontend/src/services/fetch/fetchers/organizationsFetcher.ts
+++ b/frontend/src/services/fetch/fetchers/organizationsFetcher.ts
@@ -61,7 +61,17 @@ export const getOrganizationUsers = async (
     additionalHeaders: ssgToken,
   });
   const json = (await resp.json()) as { data: UserDetail[] };
-  return json.data;
+
+  // Sort by email, this will be temp until we get the results from the backend with sorting applied
+  const sorted = [...json.data].sort((first, second) => {
+    const a = (first.email ?? "").toLowerCase();
+    const b = (second.email ?? "").toLowerCase();
+    if (a < b) return -1;
+    if (a > b) return 1;
+    return 0;
+  });
+
+  return sorted;
 };
 
 export const getOrganizationRoles = async (
@@ -130,5 +140,15 @@ export const getOrganizationPendingInvitations = async (
   const json = (await response.json()) as {
     data: OrganizationPendingInvitation[];
   };
-  return json.data;
+
+  // Sort by email, this will be temp until we get the results from the backend with sorting applied
+  const sorted = [...json.data].sort((first, second) => {
+    const a = (first.invitee_email ?? "").toLowerCase();
+    const b = (second.invitee_email ?? "").toLowerCase();
+    if (a < b) return -1;
+    if (a > b) return 1;
+    return 0;
+  });
+
+  return sorted;
 };


frontend/src/i18n/messages/en/index.ts
```diff
@@ -1538,19 +1538,18 @@ export const messages = {
   ManageUsers: {
     pageTitle: "Manage Users | Simpler.Grants.gov",
     pageHeading: "Manage Users",
-    activeUsersHeading: "Active Users",
+    activeUsersHeading: "Active Users Roster",
     activeUsersTableDescription:
       "Adjust permissions for active users so they have the right access for their role.",
+    activeUsersFetchError:
+      "We have encountered an error retrieving the Active Users roster, please try again later.",
+    activeUsersTableZeroState: "There are no active users",
```
ErinPattisonNava: Nit: Could we add a period?

frontend/src/components/manageUsers/ActiveUsersSection.tsx
```diff
@@ -0,0 +1,89 @@
+import { UnauthorizedError } from "src/errors";
```
ErinPattisonNava: Since this is likely to be a longer term component, could we include basic unittests?

frontend/src/services/fetch/fetchers/organizationsFetcher.ts
```diff
@@ -1,15 +1,20 @@
...
 import { OrganizationInviteRecord } from "src/types/organizationTypes";
 import { UserDetail, UserRole } from "src/types/userTypes";
 
 import { fetchOrganizationWithMethod, fetchUserWithMethod } from "./fetchers";
 
 export const getOrganizationDetails = async (
-  token: string,
   organizationId: string,
 ): Promise<Organization> => {
+  const session = await getSession();
```
ErinPattisonNava, line 17: Unittest coverage?

frontend/src/components/manageUsers/ActiveUsersSection.tsx
myduong-navapbc: I will add the unit test in for these components

frontend/src/components/manageUsers/ActiveUsersSection.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { getTranslations } from "next-intl/server";
+import React from "react";
+import { Alert } from "@trussworks/react-uswds";
+
+import {
+  TableCellData,
+  TableWithResponsiveHeader,
+} from "src/components/TableWithResponsiveHeader";
+
+function formatFullName(user: UserDetail): string {
```
doug-s-nava: can we reconcile this with the `formatFullName` in your previous PR?

frontend/src/components/manageUsers/ActiveUsersSection.tsx
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  return roles.map((role) => role.role_name).join(", ");
+}
+
+interface ActiveUsersSectionProps {
+  organizationId: string;
+}
+
+export async function ActiveUsersSection({
```
doug-s-nava, line 19: this file, including the component, helper functions, props, seems identical to InvitedUsersSection with two differences:

* the function used to call to get the necessary data
* the data used for the ransofmrTableRowData function
* content strings

With this in mind can we think about an implementation that is more dry?

frontend/src/components/manageUsers/ActiveUsersSection.tsx
```diff
@@ -0,0 +1,89 @@
...
+  organizationId,
+}: ActiveUsersSectionProps) {
+  const t = await getTranslations("ManageUsers");
+  let userData: UserDetail[] = [];
+  let hasError = false;
+
+  try {
+    userData = await getOrganizationUsers(organizationId);
```
doug-s-nava: can we move these calls (on the invited users table as well) into the authorization gate since they are authorized calls?

frontend/src/components/manageUsers/ActiveUsersSection.tsx
myduong-navapbc: i moved this and the display roles into a util

frontend/src/components/manageUsers/ActiveUsersSection.tsx
myduong-navapbc: this file will have a lot of other props and functionality added to it, namely the role update functionality, i wanted to get the table approved before doing a pr for the functionality

frontend/src/components/manageUsers/ActiveUsersSection.tsx
myduong-navapbc: absolutely!

frontend/src/components/manageUsers/ActiveUsersSection.tsx
```diff
@@ -0,0 +1,79 @@
...
+  TableCellData,
+  TableWithResponsiveHeader,
+} from "src/components/TableWithResponsiveHeader";
+
+interface ActiveUsersSectionProps {
+  authorizedData?: AuthorizedData;
+}
+
+export async function ActiveUsersSection({
+  authorizedData,
```
ErinPattisonNava, line 20: Nit: I wish this was named more descriptively, so we understood what data was required by the component.  But I see that it is done this way in `OrganizationDetail`.  @doug-s-nava Thoughts on how we could make a clearer pattern of what data is expected?

ErinPattisonNava (commented):
There are some spacing issues at the top -- there is too much space around the breadcrumbs/H1, and not enough space between the invite box and the Active Users Roster.  

Screenshot 2025-11-21 at 4 29 38 PM


I'd also like to ask design if we could switch the Name and the Email columns; if names aren't required to sign up, I think that will make more sense visually and sorting.  

Speaking of which, are we supposed to be sorting based off of email address?  If that's already built in I'd like to add it, and if its not then create a followup ticket/ask Eric to track it.

frontend/src/components/manageUsers/ActiveUsersSection.tsx
myduong-navapbc: I created a ticket for the discussion with design: https://github.com/HHS/simpler-grants-gov/issues/7210

It is not sorted by email by default, i've added some code in the fetcher to sort it but this should come from the BE, imo, here is a ticket to track: https://github.com/HHS/simpler-grants-gov/issues/7211

Discussion on auth wrapper authorizedData was more descriptive:  https://github.com/HHS/simpler-grants-gov/issues/7212

myduong-navapbc:
> There are some spacing issues at the top -- there is too much space around the breadcrumbs/H1, and not enough space between the invite box and the Active Users Roster.
> 
> Screenshot 2025-11-21 at 4 29 38 PM
> I'd also like to ask design if we could switch the Name and the Email columns; if names aren't required to sign up, I think that will make more sense visually and sorting.
> 
> Speaking of which, are we supposed to be sorting based off of email address? If that's already built in I'd like to add it, and if its not then create a followup ticket/ask Eric to track it.

here are the style updates as well

Screenshot 2025-11-22 at 4 27 39 PM

frontend/src/components/manageUsers/ActiveUsersSection.tsx
doug-s-nava: the way that the authorization gate is set up to work, it passes on data based on the fetch functions and permissions data that are passed in, so it is pretty loose. I think there's an improvement to be made here to make the shape of the data returned respond to a generic https://github.com/HHS/simpler-grants-gov/blob/9d21d5445daf94545a0aec2aa0093acda9e69170/frontend/src/types/authTypes.ts#L82 Would that scratch the itch that you're feeling or is there something more?

frontend/src/components/manageUsers/ActiveUsersSection.tsx
```diff
@@ -0,0 +1,81 @@
...
+}: ActiveUsersSectionProps) {
+  const t = await getTranslations("ManageUsers");
+  let userData: UserDetail[] = [];
+
+  if (!authorizedData) {
```
doug-s-nava, line 25: [nit?] this can go through as is, but I think in general a better pattern here is to error out in the first child of the authorization gate, since that is the component that actually is required to be wrapped in the gate. These section components are getting fed the prop through normal drilling from their parent so they theoretically could be used without the gate (though in practice that wouldn't really work).

With a change in place to error out in the most immediate child, you can then destructure and only pass in the necessary data for the section components, which should make them a little lighter and easier to understand, offloading the weird auth handling stuff to the parents.

That said, the error handling around the  **particular** data that each section component requires should still happen here - so this component would need to take in the activeUsersList and any error related to it

Issue #6955, opened 2025-11-04
Create Web and Print views for CD511 form


chris-kuryak:
### Summary

This work is to use the JSON from #6903 to finish the CD511 form.

Form information: https://www.grants.gov/forms/form-items-description/fid/276

UI schema already created by Michael C.

### Acceptance criteria

- [ ] Webform matches expected layout from PDF form
- [ ] Print View generates as expected
- [ ] Backend has created an Opportunity for testing this form and it is deployed to **staging**
- [ ] Product and Design notified for UAT
- [ ] Product UAT complete
  - [ ] Webform
  - [ ] PDF
- [ ] Design UAT complete
  - [ ] Webform
  - [ ] PDF

ErinPattisonNava:
@chris-kuryak Here's the opportunity Michael set up in staging: https://staging.simpler.grants.gov/opportunity/a306c591-0e7d-46df-8629-44f64d87e36c

ErinPattisonNava:
@chris-kuryak Do we have a separate ticket for promoting this to production?

chris-kuryak:
> [@chris-kuryak](https://github.com/chris-kuryak) Do we have a separate ticket for promoting this to production?

Made one for end of quad to sync across environments:
https://github.com/HHS/simpler-grants-gov/issues/7250

chris-kuryak:
UAT with Chris and Jes:

1. Move the "As the duly authorized . . ." section into normal text similar to the rest of the document (not a section title) under a new section titled "Acknowledgement"
2. Add "Enter legal name of applicant, which will undertake the assistance activity." as a description to the Name of Applicant field
3. Change "1. Award" section title to "1. Applicant Information"

chris-kuryak:
@chouinar Can you send me the PDF output from Application ID: d96ae603-4794-4267-a07c-169ab31f4c6f

ErinPattisonNava:
@chris-kuryak #1 and #3 are super easy.  

For #2, easy to do but want to triple check we are on the same page:  @chouinar has a comment in the code saying that we map that to organization_name in the legacy system.  Do we really want that clarification?

ErinPattisonNava:
> [@chouinar](https://github.com/chouinar) Can you send me the PDF output from Application ID: d96ae603-4794-4267-a07c-169ab31f4c6f

I'd also love to understand how we get PDFs on the backend, and maybe add it to the documentation where we outline how we use docraptor in localhost for the equivalent in staging.

[...]

PR #7099, opened 2025-11-18
[Issue #6955] frontend adjustments to form cd511

ErinPattisonNava:
## Summary

Work for #6955 

## Changes proposed

Prettifying the layout for the Form CD511.

## Context for reviewers

Most of this work was previously done by @chouinar, this is just making the large text section at the top more readable.

## Validation steps

A local opportunity linked to this form can be found by running `make db-seed-local` and searching the output for "CD511".  

- [ ] Successfully start an application against this form
- [ ] Ensure the form looks correct
- [ ] Ensure the form can be saved
- [ ] Ensure the application can be submitted
- [ ] Check the print setup to make sure that the PDF looks correct


api/src/form_schema/forms/cd511.py
```diff
@@ -143,7 +167,7 @@
     # https://www.grants.gov/forms/form-items-description/fid/276
     form_id=uuid.UUID("7057eaee-f043-4029-b7f2-c932f11ce900"),
     legacy_form_id=276,
-    form_name="CD511",
+    form_name="CD511 Form",
```
chouinar: Changing the name of the form is something we'd have to discuss. This is a value we copy over from grants.gov, and is a "can we change this?" sort of problem. I know this is very minor change, but I'd want to be cautious.

ErinPattisonNava: Got it, thanks.  I made the change so that the print version looked more like the actual form, but I'll revert.

ErinPattisonNava: change is 427edf2355435d385f1f7a4804f31d02919767ff

myduong-navapbc:
The form renders and looks great!

I did notice two things:
the error messaging for the `First Name` returns as `Contact First Name `and the Error Alert box returns `Contact Name First Name`

Screenshot 2025-11-19 at 5 07 33 PM

Screenshot 2025-11-19 at 5 04 30 PM

ErinPattisonNava:
> The form renders and looks great!
> 
> I did notice two things: the error messaging for the `First Name` returns as `Contact First Name `and the Error Alert box returns `Contact Name First Name`
> 
> Screenshot 2025-11-19 at 5 07 33 PM Screenshot 2025-11-19 at 5 04 30 PM

Great catch @myduong-navapbc !  @chouinar I think I can just make that change [here](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/form_schema/forms/cd511.py#L79) to call it `Contact Person` instead of `Contact Name` so the error would read as "Contact Person First Name" which I think make sense.  But since this is in the `FORM_JSON_SCHEMA` wanted to check if there are any backend concerns with that?

chouinar:
> > The form renders and looks great!
> > I did notice two things: the error messaging for the `First Name` returns as `Contact First Name `and the Error Alert box returns `Contact Name First Name`

> Great catch @myduong-navapbc ! @chouinar I think I can just make that change [here](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/form_schema/forms/cd511.py#L79) to call it `Contact Person` instead of `Contact Name` so the error would read as "Contact Person First Name" which I think make sense. But since this is in the `FORM_JSON_SCHEMA` wanted to check if there are any backend concerns with that?

Title/description are purely used by the frontend and don't affect validation/anything on the backend. You're free to adjust those values as you see fit. One caveat / something maybe to discuss later - some titles/descriptions are defined in shared schemas (eg. the first name field used in many of the forms). Changing that would affect every form that uses it - still a frontend only thing, but a bit wider of an impact.

ErinPattisonNava:
> > > The form renders and looks great!
> > > I did notice two things: the error messaging for the `First Name` returns as `Contact First Name `and the Error Alert box returns `Contact Name First Name`
> 
> > Great catch @myduong-navapbc ! @chouinar I think I can just make that change [here](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/form_schema/forms/cd511.py#L79) to call it `Contact Person` instead of `Contact Name` so the error would read as "Contact Person First Name" which I think make sense. But since this is in the `FORM_JSON_SCHEMA` wanted to check if there are any backend concerns with that?
> 
> Title/description are purely used by the frontend and don't affect validation/anything on the backend. You're free to adjust those values as you see fit. One caveat / something maybe to discuss later - some titles/descriptions are defined in shared schemas (eg. the first name field used in many of the forms). Changing that would affect every form that uses it - still a frontend only thing, but a bit wider of an impact.

Fixed in 523bc97f0fa85f03cc33a3f7e1840b7926d30721 @myduong-navapbc

Issue #6997, opened 2025-11-07
Implement ECR lifecycle policies to purge ECR images that are no longer in use


prasnava:
Implement ECR lifecycle policies to purge ECR images that are no longer in use.

PR #7123, opened 2025-11-18
[Issue #6997]: Clean up unused ECR images 

prasnava:
## Summary

Clean up unused ecr images. Scheduled as a daily cron at 5am. 

## Validation steps

The new workflow job: ECR Cleanup should run and print out what ecr tags/images where deleted.

### DIFF
--- a/bin/cleanup-ecr
+++ b/bin/cleanup-ecr
@@ -3,7 +3,7 @@
 set -euo pipefail
 
 REGION="us-east-1"
-DRY_RUN=true
+DRY_RUN=false
 
 declare -A active_tags_map
 


prasnava:
Tested the dry run locally. We check against the task definitions to see what images to keep vs delete. 
Screenshot 2025-11-19 at 1 07 55 PM

mdragon:
Can we not delete an inactive image if it's not older than 30 days old or something like that? Wondering about if we ever wanted to rollback a container temporarily, or if HHS would want to have recent images still available for scanning, etc?

prasnava:
> Can we not delete an inactive image if it's not older than 30 days old or something like that? Wondering about if we ever wanted to rollback a container temporarily, or if HHS would want to have recent images still available for scanning, etc?

yes, we can update it to retain images that are less than 30 days old and delete the rest.  Will update the script.

prasnava:
Updated the code to preserve images that are less than 30 days old. Below is the local --dry-run: 

Screenshot 2025-11-20 at 11 20 20 AM

mdragon (approved):
Looks good.

Issue #6905, opened 2025-10-30
Create JSON schema for EPA Form 4700-4


chris-kuryak:
### Summary

This ticket is to complete the backend work for the EPA Form 4700-4.

Form information: https://www.grants.gov/forms/form-items-description/fid/773

Rule schema to include:
- Post-population:
  - Signature
  - Date

Readme: https://github.com/HHS/simpler-grants-gov/tree/main/api/src/form_schema/forms

Form metadata: https://docs.google.com/spreadsheets/d/1nGolGAVMPgDQ3kaDwE9CtRPqfR6Eh40LG2WZyGAM8O8/edit?gid=1515759995#gid=1515759995

### Acceptance criteria
- [ ] JSON Schema
- [ ] Rule Schema includes signature and date post-population
- [ ] Shared JSON Schema changes

chouinar:
@chris-kuryak - for UAT - https://staging.simpler.grants.gov/opportunity/caff069f-4b42-4af2-8670-b016fee7cbb5

chris-kuryak:
This is already in UAT on the FE ticket.

PR #7138, opened 2025-11-18
[Issue #6905] Add EPA Form 4700-4

chouinar:
## Summary
Work for #6905 

## Changes proposed
* Adds the EPA Form 4700-4

## Context for reviewers
https://www.grants.gov/forms/form-items-description/fid/773

This form has a lot of fields, and handles addresses/names a bit different than other forms, so did a little restructuring of the shared schemas to leverage what does overlap.

Note that the exact way we want to format this probably needs a bit of cleanup, I got all the fields down and ordered, but didn't try to make it look nice.

## Validation steps
If you want to see the form locally:

Run `make db-migrate db-seed-local` to get form setup locally.

Login as the `many_app_user` and go to http://localhost:3000/applications - find the application associated with the EPA form. It should look something like (not all of it, it's pretty long):
Screenshot 2025-11-18 at 2 24 46 PM
Screenshot 2025-11-18 at 2 24 52 PM
Screenshot 2025-11-18 at 2 24 56 PM
Screenshot 2025-11-18 at 2 25 03 PM


api/src/form_schema/forms/project_abstract.py
```diff
@@ -61,7 +61,7 @@
     form_json_schema=FORM_JSON_SCHEMA,
     form_ui_schema=FORM_UI_SCHEMA,
     form_rule_schema=FORM_RULE_SCHEMA,
-    form_instruction_id=uuid.UUID("b09ac657-f629-431e-a7ba-3f3017a8a466"),
+    form_instruction_id=uuid.UUID("116b3ef7-d2bb-430c-860b-3277164dea7b"),
```
chouinar, line 64: I realized on this and another new form I messed up the form instruction ID and fixed it. This one had the same ID as another, and the supplementary form doesn't have instructions at all.

Issue #6815, opened 2025-10-21
Modify the SOAP proxy endpoints to enforce RBAC by fetching the user of the cert from the DB


chris-kuryak:
### Summary

For this we need to process the cert (logic should already be present) and pull the serial number out of it. 

We should take that serial number and query the `legacy_certificate` table by that serial number.

- If no cert is found, error.
- If that cert is expired (based on current EST date), error.
- If the certificate is not connected to an agency, error (NOTE: this should only be the behavior for grantor endpoints)

Then for the given operation, we need to figure out which privilege a user requires which should be configured per operation.

- `GetSubmissionListExpanded` → `legacy_agency_viewer`
- `GetApplication` → `legacy_agency_grant_retriever`
- `GetApplicationZip` → `legacy_agency_grant_retriever`
- `ConfirmApplicationDelivery` → `legacy_agency_grant_retriever`
- `UpdateApplicationInfo` → `legacy_agency_assigner`

We then just need to call our RBAC code passing in the user associated with the cert, the privilege defined for the operation, and the Agency.

--
**NOTE:** 
Some systems that parse serial numbers (like [Home | Grants.gov](http://grants.gov/) ) seem to add leading zeroes in some cases to serial numbers. Our cert in training doesn’t start with 00 but it does in what [Home | Grants.gov](http://grants.gov/)  stored. Seems this depends on the particular system and special octet/int parsing to make it “valid”. Need to verify what the library we use to parse the cert does for these and potentially adjust it.

At the very least, `openssl x509 -noout -serial -in <some_file>.crt` doesn’t have the leading zeroes.

--

### Acceptance criteria

- [ ] Process the cert dnd pull the serial number out of it
- [ ] Throw error If no cert is found
- [ ] Throw error if cert is expired
- [ ] Throw error if cert is not connected to agency (grantor only)
- [ ] Figure out which privilege a user requires per operation
- [ ] Calls RBAC code with the user associated with the cert, privilege for operation, and agency

PR #7149, opened 2025-11-19
[Issue #6815] Modify SOAP endpoint to enforce RBAC

jakobpederson:
## Summary

Fixes / Work for #6815 

## Changes proposed

Add validation check for certificate. 

## Context for reviewers

Using the serial_number on a certificate, get the LegacyCertificate, which gives you the Agency and the User. Using the operation_name you can get the correct Privilege. At that point you can check the User access based on that Privilege. This should enforce RBAC.

- [x] Process the cert dnd pull the serial number out of it
- [x] Throw error If no cert is found
- [x] Throw error if cert is expired
- [x] Throw error if cert is not connected to agency (grantor only)
- [x] Figure out which privilege a user requires per operation
- [x] Calls RBAC code with the user associated with the cert, privilege for operation, and agency

### DIFF
--- a/api/src/legacy_soap_api/legacy_soap_api_config.py
+++ b/api/src/legacy_soap_api/legacy_soap_api_config.py
@@ -77,12 +77,12 @@ class SOAPOperationConfig:
     is_mtom: bool = False
     always_call_simpler: bool = False
 
-    # ENDPOINT_PRIVILEGES:
-    # GetSubmissionListExpandedRequest={Privilege.LEGACY_AGENCY_VIEWER},
-    # GetApplicationRequest={Privilege.LEGACY_AGENCY_GRANT_RETRIEVER},
-    # GetApplicationZipRequest={Privilege.LEGACY_AGENCY_GRANT_RETRIEVER},
-    # ConfirmApplicationDeliveryRequest={Privilege.LEGACY_AGENCY_GRANT_RETRIEVER},
-    # UpdateApplicationInfoReqest={Privilege.LEGACY_AGENCY_ASSIGNER},
+    # These are the privileges needed for these endpoints:
+    # GetSubmissionListExpandedRequest = {Privilege.LEGACY_AGENCY_VIEWER}
+    # GetApplicationRequest = {Privilege.LEGACY_AGENCY_GRANT_RETRIEVER}
+    # GetApplicationZipRequest = {Privilege.LEGACY_AGENCY_GRANT_RETRIEVER}
+    # ConfirmApplicationDeliveryRequest = {Privilege.LEGACY_AGENCY_GRANT_RETRIEVER}
+    # UpdateApplicationInfoReqest = {Privilege.LEGACY_AGENCY_ASSIGNER}
     privileges: set[Privilege] | None = None
 
     # Some SOAP XML payloads will not force a list of objects when converting to


api/src/legacy_soap_api/grantors/services/get_application_zip_response.py
```diff
@@ -35,10 +40,21 @@ def get_application_zip_response(
...
-            ApplicationSubmission.legacy_tracking_number == int(legacy_tracking_number)
+            ApplicationSubmission.legacy_tracking_number == int(legacy_tracking_number),
         )
     ).scalar()
+
     if application:
+        certificate = validate_certificate(
+            db_session, soap_auth=soap_request.auth, api_name=soap_request.api_name
+        )
+        if soap_request.operation_name in ENDPOINT_PRIVILEGES:
```
chouinar: Two questions:
* Can we put the configuration for the privileges in the endpoint config?
* If there are no configured privileges, can we error? Note that empty privileges could be fine (there's the get opportunity endpoint on the grantor side that needs no privilege).


```suggestion
        endpoint_privilege = ENDPOINT_PRIVILEGES.get(soap_request.operation_name, None)
        if endpoint_privilege is None:
             raise Exception(...) # or however we want to handle error case
        if len(endpoint_privilege) > 0:
             ...   
```

api/tests/src/legacy_soap_api/test_legacy_soap_api_auth.py
```diff
@@ -21,6 +36,13 @@
 )
 
 
+@pytest.fixture(autouse=True)
+def cleanup_agencies(db_session):
+    cascade_delete_from_db_table(db_session, LegacyCertificate)
+    cascade_delete_from_db_table(db_session, AgencyUser)
+    cascade_delete_from_db_table(db_session, Agency)
```
chouinar: I'll put this on any PR that adds these cleanup calls. Is there any way to could implement this that does not require you do this / why did you add this? The performance hit from doing this is immense and we should avoid it at all costs. Nothing I see in the tests you've written looks like this would be necessary.

The only scenario in which we should use these is when the code implemented does a `select * from table` and processes all of it like some of our backend processes. Any endpoint that fetches specific values, do not do this.
```suggestion

```

api/tests/src/legacy_soap_api/test_legacy_soap_api_client.py
```diff
@@ -39,6 +50,13 @@
 BOUNDARY_UUID = "cccccccc-1111-2222-3333-dddddddddddd"
 
 
+@pytest.fixture(autouse=True)
+def cleanup_agencies(db_session):
+    cascade_delete_from_db_table(db_session, LegacyCertificate)
+    cascade_delete_from_db_table(db_session, AgencyUser)
+    cascade_delete_from_db_table(db_session, Agency)
```
chouinar:
```suggestion

```

api/tests/src/legacy_soap_api/test_legacy_soap_api_client.py
```diff
@@ -357,10 +375,23 @@ def test_get_simpler_soap_response_when_operation_is_not_get_opportunity_list_re
...
+        legacy_certificate = LegacyAgencyCertificateFactory.create(agency=agency)
+        mock_validate_certificate.return_value = legacy_certificate
         submission = ApplicationSubmissionFactory.create()
+        agency_user = AgencyUserFactory.create(agency=agency, user=legacy_certificate.user)
+        role = RoleFactory.create(privileges=[Privilege.LEGACY_AGENCY_GRANT_RETRIEVER])
+        AgencyUserRoleFactory.create(agency_user=agency_user, role=role)
+        application_user = ApplicationUserFactory.create(
+            application=submission.application, user=legacy_certificate.user
+        )
+        ApplicationUserRoleFactory.create(application_user=application_user, role=role)
```
chouinar, line 387: Every test of a SOAP endpoint that needs valid auth from here on out is going to need a lot of this, can we make utility function that removes all the cert/user/role setup?

Something you can call like (with most of these values optional):
```py
user, cert = setup_cert_user(agency=agency, privileges=[Privilege.X, Privilege.Y])
```

api/tests/src/legacy_soap_api/test_legacy_soap_api_client.py
```diff
@@ -357,10 +375,23 @@ def test_get_simpler_soap_response_when_operation_is_not_get_opportunity_list_re
 
 
 class TestSimplerSOAPGetApplicationZip:
+    @patch(
+        "src.legacy_soap_api.grantors.services.get_application_zip_response.validate_certificate"
+    )
```
chouinar: Is the patch needed? If we setup the data in the DB, would it not be fetched? If at all possible, we should avoid mocking our own system, always best to verify it works by letting it work.

api/src/legacy_soap_api/legacy_soap_api_auth.py
```diff
@@ -100,3 +113,45 @@ def get_soap_client_certificate(urlencoded_cert: str) -> SOAPClientCertificate:
...
+        )
+        raise SOAPClientCertificateLookupError("no soap auth")
+
+    serial_number_str = str(soap_auth.certificate.serial_number)
+    legacy_certificate = (
+        db_session.query(LegacyCertificate)
+        .filter(LegacyCertificate.serial_number == serial_number_str)
+        .first()
+    )
+    extra_data = {"serial_number": serial_number_str}
```
chouinar: I believe this is the same serial number we wanted to not log elsewhere.
```suggestion

```

api/src/legacy_soap_api/legacy_soap_api_auth.py
```diff
@@ -100,3 +113,45 @@ def get_soap_client_certificate(urlencoded_cert: str) -> SOAPClientCertificate:
...
+
+    if not legacy_certificate:
+        logger.warning(
+            "soap_client_certificate: could not retrieve client cert for serial number",
+            extra=extra_data,
+        )
+        raise SOAPClientCertificateLookupError("could not retrieve client cert for serial number")
+
+    extra_data.update({"legacy_certificate_id": str(legacy_certificate.legacy_certificate_id)})
```
chouinar: Slightly shorter way to do this - don't need to stringify UUIDs, our logging automatically converts most types in the extra - can see how those work in https://github.com/HHS/simpler-grants-gov/blob/main/api/src/logging/formatters.py#L29

Can also do `|=` instead of update to trim a little out.

```suggestion
    extra_data |= {"legacy_certificate_id":legacy_certificate.legacy_certificate_id}
```

api/src/legacy_soap_api/grantors/services/get_application_zip_response.py
```diff
@@ -35,10 +40,21 @@ def get_application_zip_response(
...
+            verify_access(
+                certificate.user,
+                {ENDPOINT_PRIVILEGES[soap_request.operation_name]},
+                application.application,
+            )
```
chouinar: Seeing this made me realize we didn't talk about the error cases here - Do we handle catching and transforming the error this returns into whatever format it needs to be? Not sure what format we'll actually want. If there were an auth issue, would we want to error completely if we got results from grants.gov (less relevant for this endpoint, but worth thinking about).

api/src/legacy_soap_api/legacy_soap_api_auth.py
```diff
@@ -100,3 +113,45 @@ def get_soap_client_certificate(urlencoded_cert: str) -> SOAPClientCertificate:
...
+            "soap_client_certificate: no soap auth",
+        )
+        raise SOAPClientCertificateLookupError("no soap auth")
+
+    serial_number_str = str(soap_auth.certificate.serial_number)
+    legacy_certificate = (
+        db_session.query(LegacyCertificate)
+        .filter(LegacyCertificate.serial_number == serial_number_str)
+        .first()
+    )
```
chouinar: We should avoid query, it's the legacy way of querying, also avoid `first` unless there's a scenario where we expect multiple results and only want one. This should be `one_or_none`, if there were multiple, we'd want that to error as there would be a fundamentally incorrect assumption.

```suggestion
    legacy_certificate = (
        db_session.execute(select(LegacyCertificate).where(LegacyCertificate.serial_number == serial_number_str)).scalar_one_or_none()
    )
```

api/src/legacy_soap_api/grantors/services/get_application_zip_response.py
jakobpederson: Yeah, we can move the privileges to the endpoint config. It makes a lot of sense there. I've added a check now to see if the privileges are None. If you want it to be empty you can set it to [].

api/tests/src/legacy_soap_api/test_legacy_soap_api_auth.py
jakobpederson: After introducing the agency_code fix I was still having unrelated tests fail, and this was what I found was the workaround. I've managed strip out more of it, but I haven't found a way around to remove at least one `cascade_delete_from_db_table(db_session,  AgencyUser)`.

api/tests/src/legacy_soap_api/test_legacy_soap_api_client.py
jakobpederson: It's not needed, I've removed it.

api/src/legacy_soap_api/legacy_soap_api_auth.py
jakobpederson: Sorry, I just spaced on that one.

api/src/legacy_soap_api/grantors/services/get_application_zip_response.py
jakobpederson: I believe that when this fails it raises an HTTPError (which should be loggeD) which should bubble up to legacy_soap_api_routes, where it's logged and then the proxy_response is returned. If we want different behavior I can put that in.

I think we should just send back the results for the grants.gov. If the certificate is somehow valid there, then it's data that should be available, and if it's not, we'll just pass back whatever error data comes from grants.gov.

api/src/legacy_soap_api/legacy_soap_api_schemas.py
```diff
@@ -74,6 +74,11 @@ def get_soap_request_operation_config(self) -> SOAPOperationConfig:
                 f"Simpler {self.api_name.value} SOAP API does not support {operation_name}"
             )
 
+        if operation_config.privileges is None:
```
jakobpederson, line 77: I thought it made sense to put the check here.

api/src/legacy_soap_api/grantors/services/get_application_zip_response.py
```diff
@@ -35,10 +42,32 @@ def get_application_zip_response(
         legacy_tracking_number = legacy_tracking_number.split("GRANT")[1]
     application = db_session.execute(
         select(ApplicationSubmission).where(
-            ApplicationSubmission.legacy_tracking_number == int(legacy_tracking_number)
+            ApplicationSubmission.legacy_tracking_number == int(legacy_tracking_number),
         )
     ).scalar()
     if application:
+        soap_config = get_soap_operation_config(soap_request.api_name, soap_request.operation_name)
```
chouinar: Any way we could make that config available rather than re-fetch it? Might need to restructure things a bit from the look of it, but would feel cleaner than re-calculating it.

Also avoids the problem below where we check if it's null again, it shouldn't ever be null at this point.

api/src/legacy_soap_api/legacy_soap_api_config.py
```diff
@@ -7,12 +7,26 @@
...
 logger = logging.getLogger(__name__)
 
 
+ENDPOINT_PRIVILEGES = dict(
+    GetSubmissionListExpandedRequest=Privilege.LEGACY_AGENCY_VIEWER,
+    GetApplicationRequest=Privilege.LEGACY_AGENCY_GRANT_RETRIEVER,
+    GetApplicationZipRequest=Privilege.LEGACY_AGENCY_GRANT_RETRIEVER,
+    ConfirmApplicationDeliveryRequest=Privilege.LEGACY_AGENCY_GRANT_RETRIEVER,
+    UpdateApplicationInfoReqest=Privilege.LEGACY_AGENCY_ASSIGNER,
+)
```
chouinar: Could make these sets instead of single things so we don't have to do `set(..)` elsewhere. Not sure if an endpoint we will build for SOAP needs multiple ever.

api/src/legacy_soap_api/grantors/services/get_application_zip_response.py
```diff
@@ -35,10 +42,32 @@ def get_application_zip_response(
...
+        certificate = validate_certificate(
+            db_session, soap_auth=soap_request.auth, api_name=soap_request.api_name
+        )
+        if soap_config and soap_config.privileges is not None:
+            try:
+                verify_access(
+                    certificate.user,
+                    set(soap_config.privileges),
+                    application.application,
```
chouinar: Could we adjust the naming of what we fetch above to be `application_submission = db_session.execute(..)`.

api/src/legacy_soap_api/grantors/services/get_application_zip_response.py
```diff
@@ -35,10 +42,32 @@ def get_application_zip_response(
...
+                )
+            except HTTPError as e:
+                logger.info(
+                    "User did not have permission to access this application",
+                    extra={
+                        "user_id": certificate.user.user_id,
+                        "application_submission_id": application.application_submission_id,
+                        "priviliges": soap_config.privileges,
```
chouinar:
```suggestion
                        "privileges": soap_config.privileges,
```

api/src/legacy_soap_api/legacy_soap_api_config.py
```diff
@@ -119,13 +134,15 @@ class SOAPOperationConfig:
...
             },
+            privileges=[],
         )
     },
     SimplerSoapAPI.GRANTORS: {
         "GetApplicationZipRequest": SOAPOperationConfig(
             request_operation_name="GetApplicationZipRequest",
             response_operation_name="GetApplicationZipResponse",
             is_mtom=True,
+            privileges=[ENDPOINT_PRIVILEGES["GetApplicationZipRequest"]],
```
chouinar: Might be better to just define the privilege here directly rather than with the dict so you can see all the config for an endpoint in one place and not need to jump around.

api/src/legacy_soap_api/legacy_soap_api_schemas.py
chouinar: What happens in the opportunities endpoint without privileges?

jakobpederson: I set the opportunities privileges to an empty set, so it should function as normal because privileges are not None.

chouinar (approved):

Issue #7159, opened 2025-11-19
Don’t post deployment failures to slack unless they are on main or are a release


prasnava:
Currently, all the GitHub workflows are configured to notify slack on failures for all branches. To reduce noise in the alerts-infra slack channel we want to do only send notifications for the main branch or when a new release is created.

PR #7162, opened 2025-11-19
[Issue #7159]: Update workflows to reduce slack notifications

prasnava:
## Summary

Updated GitHub workflows to reduce slack notifications. 

## Validation steps

All workflows should continue to work as is except slack notifications should only be triggered when merged into main or when a release is created.

### DIFF
--- a/.github/workflows/cd-analytics.yml
+++ b/.github/workflows/cd-analytics.yml
@@ -55,7 +55,7 @@ jobs:
       version: ${{ inputs.version || github.ref }}
 
   send-slack-notification:
-    if: failure()
+    if: failure() && (github.event_name == 'schedule' || (github.event_name == 'push' && github.ref == 'refs/heads/main') || github.event_name == 'release')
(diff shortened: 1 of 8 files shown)


.github/workflows/cd-analytics.yml
```diff
@@ -55,7 +55,7 @@ jobs:
       version: ${{ inputs.version || github.ref }}
 
   send-slack-notification:
-    if: failure()
+    if: failure() && (github.event_name == 'schedule' || (github.event_name == 'push' && github.ref == 'refs/heads/main') || github.event_name == 'release')
```
chouinar, line 58: Which events do these cut out from our existing notifications?

Would it just be manual deploys to an environment from a dev branch that are excluded or are there other cases?

prasnava: This would cut out only manual deploys.

mdragon (approved):
Looks good

Issue #6782, opened 2025-10-20
Debug CloudFront cache issues


prasnava:
### Summary

With our current cloudfront cdn config we are seeing cacheing issues. The goal is to identify the issues with our current config, and make changes to our terraform iac.

### Acceptance criteria

Create tickets for next steps related to fixing the cdn config issues.

prasnava:
Update: Sent out a dns change request form to test the changes in dev. Once the zone has been updated we can validate the cloudfront cacheing behavior.

prasnava:
Update: The dns zone has been updated. It's now pointing at cloud front.

Issue #5980, opened 2025-08-19
Server actions don't work behind CDN


doug-s-nava:
### Summary

Follow up to https://github.com/HHS/simpler-grants-gov/issues/5910

This will need to be handled before enabling the CDN in prod

### Acceptance criteria

- [ ] server actions (saving forms, subscribing to newsletter) work in deployed env with CDN enabled

PR #7181, opened 2025-11-20
[Issue #6782] Changes to Next/CDN to make them work correctly together

mdragon:
## Summary

Fixes / Work for #6782 

## Changes proposed

We're working to get the CDN back in place and working correctly with the FE code.

## Context for reviewers

* Previously we'd had issues where signed in user's pages were cached and so the user would see stale pages (like the stars on the search results) or the Saved Opportunities page. 
* This works to address those issues via changes to the CDN config and how the NextJS code returns cache related headers. It also adds things like "Clear cache" to the deploy flow so that we clear the cache when a new FE rolls out, etc.
* This also supplements the FE "request" logging to include information from the response like the status code, whether the session cookie was present and what the cache headers back to the CDN where.
* We also add a /cdn endpoint that's only usable locally or in Dev that allows the caller to control some of the caching headers to help validate/troubleshoot the CDN behavior.

## Validation steps
1. Test on https://dev.simpler.grants.gov
2. Use the site not signed in and check that your page requests are coming back as "Cache hit" in the X-Cache header when revisiting or refreshing pages you've already visited. You'll also see the Age header which is in seconds and for common pages you should see an Age older than when you started testing indicating you're receiving a page that was cached not based on a request you even made.
3. Sign in
4. Now the Age header should not be present, and the X-Cache header should indicate "Cache miss" even for pages like the home page that was previously serving from Cache hits. 
5. Go to the search page, and start or unstar some Opps.
6. Refresh the page and validate that the stars remain correct.
7. Go to the Saved Opps page and remove an opportunity or two.
8. Refresh the page to validate that those Opps don't "come back"
9. Sign out
10. Validate that pages like the home page serve as "Cache hit" again.

You can also validate this on the command line using this command which will run parallel requests with and without the session cookie that drives this no caching while logged in behavior.
```
curl -v "https://dev.simpler.grants.gov/?asdf1" -b "session=abc;" -o /dev/null 2>&1 | egrep -i "cache|age:" && curl -v "https://dev.simpler.grants.gov/?asdf1"  -o /dev/null 2>&1 | egrep -i "cache|age:"
```

### DIFF
--- a/frontend/src/middleware.ts
+++ b/frontend/src/middleware.ts
@@ -83,9 +83,9 @@ export default function middleware(request: NextRequest): NextResponse {
   // only allow for cdn testing/troubleshooting in lower envs
 
   if (isACdnTestRequest(request)) {
-    const response = handleCdnTest(request);
-    logRequest(request, response);
-    return response;
+    const testResponse = handleCdnTest(request);
+    logRequest(request, testResponse);
+    return testResponse;
   }
   const response = request.url.match(/api\//)
     ? featureFlagsManager.middleware(request, NextResponse.next())


frontend/src/middleware.ts
```diff
@@ -42,6 +43,42 @@ const i18nMiddleware = createIntlMiddleware({
...
+
+  // only allow for cdn testing/troubleshooting in lower envs
+  if (
+    request.url.includes("/cdn") &&
+    ["dev", "local"].indexOf(environment.ENVIRONMENT) > -1
+  ) {
+    const url = new URL(request.url);
+    const params = new URLSearchParams(url.search);
+
+    const cacheControl: string[] = [];
```
doug-s-nava: why is this defined twice?

frontend/src/middleware.ts
```diff
@@ -42,6 +43,42 @@ const i18nMiddleware = createIntlMiddleware({
...
 
 export default function middleware(request: NextRequest): NextResponse {
+  const cacheControl: string[] = [];
+
+  // only allow for cdn testing/troubleshooting in lower envs
+  if (
+    request.url.includes("/cdn") &&
+    ["dev", "local"].indexOf(environment.ENVIRONMENT) > -1
+  ) {
+    const url = new URL(request.url);
```
doug-s-nava: can we split this into its own function?

doug-s-nava (commented):
have not tested but I am sure that these changes haven't fixed all of our frontend / CDN related bugs, so want to make sure that once this merges there's no plan to turn the CDN back on before fixing those

frontend/src/middleware.ts
mdragon: Copypasta

frontend/src/middleware.ts
mdragon: The URL and params parsing? Just making sure I know which part you're looking to package up.

mdragon:
> have not tested but I am sure that these changes haven't fixed all of our frontend / CDN related bugs, so want to make sure that once this merges there's no plan to turn the CDN back on before fixing those

@doug-s-nava no plans to turn it back on (that won't happen until we change the CNAME for the environment). However this is pretty broad in terms of not caching anything if the user is logged in. Can we chat about other caching related bugs that we think that wouldn't cover? We'll move this forward but want to make sure we have a clearer list of stuff to solve before we do turn it on.

frontend/src/middleware.ts
doug-s-nava:
```suggestion
    const handleCdnTest = (request) => {
    const url = new URL(request.url);
        const params = new URLSearchParams(url.search);

    const cacheControl: string[] = [];
         cacheControl.push(`max-age=${params.get("max-age") || "10"}`);
    cacheControl.push(
      params.get("cache") ||
        (request.cookies.has("session") &&
        request.cookies.get("session")?.value !== ""
          ? "no-store"
          : "public"),
    );

    const response = new NextResponse(
      JSON.stringify({ params: params.entries(), cacheControl }),
      {
        status: 200,
        headers: {
          "Cache-Control": cacheControl.join(", "),
          "Content-Type": "application/json",
        },
      },
    );

    return response
  }
    }
    
    ...
    
    if (isACdnTestRequest) {
      const response = handleCdnTest(request}
          logRequest(request, response);
          return response
    }
```

doug-s-nava:
@mdragon this is the main thing I have in mind https://github.com/HHS/simpler-grants-gov/issues/5980

prasnava:
> @mdragon this is the main thing I have in mind #5980

For the server actions endpoint, we can add a behavior to bypass all /api routes.

frontend/src/middleware.ts
```diff
@@ -42,78 +42,51 @@ const i18nMiddleware = createIntlMiddleware({
...
-      },
-    );
-    logRequest(request, response);
+export default function middleware(request: NextRequest): NextResponse {
+  const cacheControl: string[] = [];
+
+  // only allow for cdn testing/troubleshooting in lower envs
 
+  if (isACdnTestRequest(request)) {
+    const response = handleCdnTest(request);
```
doug-s-nava: would be surprised if the linter didn't complain about this shadowed variable name (since we define `response` again later in the file). Maybe change to `testResponse`?

mdragon:
> > @mdragon this is the main thing I have in mind #5980
> 
> For the server actions endpoint, we can add a behavior to bypass all /api routes.

I do see that the middleware is intentionally not running on API routes. Might be worth thinking about if this should be in new a middleware that runs in additional/different places? The API for newsletter and saving/submitting Applications should be happening in POSTs, PUTs, etc. so they shouldn't be caching, which makes me suspect the issues were more likely to be related to headers/cookies not being passed to the origin rather than caching. So still CDN to blame but a different cause within the CDN config/layer.

doug-s-nava (approved):
I think we need to have more discussion about what's going on with Next and the CDN, but this seems like a step in the right direction

PR #7202, opened 2025-11-21
[ISSUE #6952] discover legacy users endpoint

### LINKED ISSUE #6952: Discover Legacy Users endpoint
### Summary

Implement the API endpoint that allows organizations to discover legacy users from the Oracle database.

### Tech Notes
- [Discover Legacy Users Endpoint](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2174681171/Tech+Spec+Make+it+easy+for+Organizations+to+invite+users+from+existing+system#Discover-Legacy-Users)

### Acceptance criteria

- Endpoint returns legacy users matching organization UEI
- unit tests created

joshtonava:
## Summary

Fixes #6952  

## Changes proposed

- `POST /v1/organizations/:organization_id/legacy-users` - Lists legacy Oracle db users associated with an
  organization's UEI that can be invited
- Supports pagination (default: 10 per page), sorting by email/full_name/created_date, and
  filtering by status
- Added `create_legacy_user_with_status()` helper to reduce test boilerplate 

## Context for reviewers

- Uses SQL window functions (ROW_NUMBER()) for efficient email deduplication
- Queries staging tables (staging.v_user_account, staging.tuser_profile) which are synced hourly from Oracle
- Status precedence: member > pending_invitation > available (computed via SQL CASE statement)

## Validation steps

Run the tests:
```bash
make test args="tests/src/api/organizations_v1/test_list_legacy_users.py -v"
```

### DIFF
--- a/api/openapi.generated.yml
+++ b/api/openapi.generated.yml
@@ -1795,6 +1795,61 @@ paths:
               $ref: '#/components/schemas/OrganizationCreateInvitationRequest'
       security:
       - ApiJwtAuth: []
+  /v1/organizations/{organization_id}/legacy-users:
+    post:
+      parameters:
+      - in: path
+        name: organization_id
+        schema:
+          type: string
+        required: true
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/LegacyUsersListResponse'
+          description: Successful response
+        '422':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+          description: Validation error
+        '401':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+          description: Authentication error
+        '404':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+          description: Not found
+        '400':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+          description: Bad Request
+        '403':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+          description: Forbidden
+      tags:
+      - Organization v1
+      summary: List legacy users that can be invited to the organization.
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/LegacyUsersListRequest'
+      security:
+      - ApiJwtAuth: []
   /alpha/applications/{application_id}/attachments:
     post:
       parameters:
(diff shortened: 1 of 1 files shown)


api/src/api/organizations_v1/organization_schemas.py
```diff
@@ -335,3 +337,105 @@ class OrganizationIgnoreLegacyUserRequestSchema(Schema):
...
+        "sort_direction": fields.Enum(
+            SortDirection,
+            required=True,
+            metadata={"description": "Whether to sort the response ascending or descending"},
+        ),
+    },
+    name="LegacyUserSortOrderSchema",
+)
+
+LegacyUserPaginationSchema = BasePaginationSchema.from_dict(
```
chouinar: Why build this instead of using the `generate_pagination_schema`? Except for changing the default, I don't see anything else that is different, and if we want to be able to set a default, can just adjust the function to allow that as an optional parameter.

api/src/services/organizations_v1/list_legacy_users.py
```diff
@@ -0,0 +1,225 @@
...
+        404: Organization not found
+        400: Organization does not have a UEI
+    """
+    # Get organization and verify access
+    organization = get_organization_and_verify_access(db_session, user, organization_id)
+
+    # Get organization UEI from SAM.gov entity
+    uei = organization.sam_gov_entity.uei if organization.sam_gov_entity else None
+    if not uei:
+        raise_flask_error(400, "Organization does not have a UEI")
```
chouinar, line 244: Not sure if 400 would be the right status code, the request isn't bad, something with the data in our system is bad. It might be a 500?

api/src/services/organizations_v1/list_legacy_users.py
```diff
@@ -0,0 +1,225 @@
...
+
+    # Parse pagination parameters and filters
+    params = LegacyUserListParams.model_validate(request_data)
+
+    # Build subqueries to check membership and invitation status
+
+    # Subquery to check if user is a member
```
chouinar: High-level question for these queries, I always err against trying to do everything in SQL because it makes it harder to parse - is there a way we could do less in SQL and do a little in python in a way that would significantly simplify this?

The way I'm thinking of it, we want the following logic:
* (SQL Query) Fetch all pending organization invitations
* (SQL Query) Fetch all user emails for users already in the organization user table
* (SQL Query) A query to fetch users from the legacy tables like `select * from vuser_account join tuser_profile on ... where vuser_account.is_deleted != 'Y' ...`

Then I think the statuses are roughly:
* email in the organization user table -> Member
* email in the pending user table -> Pending Invitation
* else -> Available

Same result, more, but simpler SQL queries, might add a few ms to the time. Fine with trading a bit of time for readability / ease to debug.

api/src/services/organizations_v1/list_legacy_users.py
```diff
@@ -0,0 +1,225 @@
...
+            ),
+        )
+    ).subquery()
+
+    # Get total count of deduplicated users for pagination info (before status filtering)
+    count_query = (
+        select(func.count())
+        .select_from(ranked_subquery)
+        .where(ranked_subquery.c.duplicate_rank == 1)
+    )
```
chouinar: If we simplify this query to just be at its core `select * from vuser_account where <some filters>` - would that simplify things in a way that lets us use the paginator?

api/src/services/organizations_v1/list_legacy_users.py
```diff
@@ -0,0 +1,225 @@
...
+        )
+    ).label("has_pending_invitation")
+
+    # Build deduplicated query with status computation using window function
+    # Use ROW_NUMBER() to rank duplicates by created_date (most recent first)
+    ranked_subquery = (
+        select(
+            VuserAccount.user_account_id,
+            VuserAccount.email,
+            VuserAccount.full_name,
```
chouinar: Should we use the first/last name fields instead? We might want to check, when grants.gov tables have these duplicative columns, usually one of them is missing some data.

api/src/api/organizations_v1/organization_schemas.py
joshtonava: Getting the defaults in was the issue. I can adjust this.

api/src/services/organizations_v1/list_legacy_users.py
joshtonava: I see where you are coming from. I was going for speed here but I can do this more in memory if that is preffered.

api/src/services/organizations_v1/list_legacy_users.py
joshtonava: I can give it a shot and see what comes of it

api/src/services/organizations_v1/list_legacy_users.py
joshtonava: I dont have a strong intuition for the data on what is more reliable or if they are identical. We can switch to first name and last name. It also gives flexibility to sort on more columns if we wanted.

api/src/services/organizations_v1/list_legacy_users.py
joshtonava: Let me try a hybrid approach. I will try to have something up soon

api/src/services/organizations_v1/list_legacy_users.py
joshtonava: 500 would imply a server problem rather than a data state issue.

400 Indicates the request can't be processed due to the organization's data state which is consistent with client errors where the resource state prevents the operation. but that is my opinion

api/src/services/organizations_v1/list_legacy_users.py
joshtonava: Can't use the paginator with doing the status filtering outside of SQL unfortunately.

api/src/services/organizations_v1/list_legacy_users.py
```diff
@@ -0,0 +1,286 @@
...
+
+    if email_lower in member_emails:
+        return LegacyUserStatus.MEMBER
+    elif email_lower in invitation_emails:
+        return LegacyUserStatus.PENDING_INVITATION
+    else:
+        return LegacyUserStatus.AVAILABLE
+
+
+def filter_and_paginate_users(
```
joshtonava, line 88: Note for reviewerers that may not have the full context. We split out the more complicated all SQL query into a few queries and are doing the filtering in code. The side effect is we cant use paginator since we are filtering status outside of SQL

PR #7222, opened 2025-11-24
[unticketed] allow playwright to point to deployed env

doug-s-nava:
## Summary

unticketed

## Changes proposed

Sets up env var to control where playwright tests will run

## Context for reviewers

For the moment this is only for local testing, but will support running against deployed env in CI in the future

## Validation steps

1. set `PLAYWRIGHT_BASE_URL="https://staging.simpler.grants.gov"` in your .env.local
2. run the playwright test suite
3. _VERIFY_: tests run against staging and pass as expected

### DIFF
--- a/frontend/.env.development
+++ b/frontend/.env.development
@@ -41,3 +41,5 @@ NEW_RELIC_LICENSE_KEY=
 SESSION_SECRET=extraSecretSessionSecretValueSssh
 
 E2E_USER_AUTH_TOKEN=  # used for spoofing logins in e2e need to sync with API via seed
+
+PLAYWRIGHT_BASE_URL= # used to point playwright testing to a deployed environment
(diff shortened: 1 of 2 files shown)


ErinPattisonNava:
This feels like it should be a command line option rather than an environment variable.  It would allow us to jump between local tests and staging tests without having to modify configuration.  Thoughts?

myduong-navapbc:
> This feels like it should be a command line option rather than an environment variable. It would allow us to jump between local tests and staging tests without having to modify configuration. Thoughts?

that is a great point, i didnt even think about jumping between multiple envs tests

mdragon (approved):
Looks good

mdragon:
> This feels like it should be a command line option rather than an environment variable. It would allow us to jump between local tests and staging tests without having to modify configuration. Thoughts?

I think we still have both options right? Running it without the `PLAYWRIGHT_BASE_URL` prefix would execute locally, so you could flip back and forth pretty easily?

ErinPattisonNava:
> I think we still have both options right? Running it without the `PLAYWRIGHT_BASE_URL` prefix would execute locally, so you could flip back and forth pretty easily?

I guess that is another reason I'd rather have it passed in as a parameter when I run the command, I'll know exactly what environment I'm testing against rather than having to check an env var first/forgetting what I've set my local to.  It is also clearer for new people with fewer configuration steps.  Just my $.02.

Issue #6935, opened 2025-11-04
Creating Organizations page


eric-serrato95:
### Summary

Capturing the effort for putting together a page showing a list of organizations (in case a user belongs to multiple organizations) 

### Acceptance criteria

- [ ] A page is created for showing a list of organizations a user belongs to 
- [ ] For each organization, a CTA banner is created with the name of the org 
- [ ] The CTA also contains a link to access the specific organization page

eric-serrato95:
@ErinPattisonNava moving to done, UAT ✅

PR #7230, opened 2025-11-25
[Issue #6935]: adding the organizations page

ErinPattisonNava:
## Summary

Work for #6935 

## Changes proposed

We're creating a new Organizations page behind the `manageUsersOff` feature flag so users can see the organizations where they are members, and have easy access to link to the organization details and (if applicable) manage the users in that organization. 

## Context for reviewers

We'll be reusing a component already created on the Activity Dashboard and updating it as well to have the Manage Users functionality.  

I don't have unittests right now for any of the work gated by feature flags due to the lack of examples in our code base for how to create them.  I've created a follow up ticket [here](https://github.com/HHS/simpler-grants-gov/issues/7231).  

## Validation steps

For setup, first set the feature flags (http://localhost:3000/dev/feature-flags) page 

- [ ] If manageUsersOff is set to enabled, then link should not appear in header and when going to [/organizations](http://localhost:3000/organizations) it should redirect you to the maintenance page

For all other instructions, manageUsersOff is set to disabled

- [ ] If user is logged out when going to   [/organizations](http://localhost:3000/organizations) then user is prompted to log in
- [ ] Organizations always appears in the header under workspace alphabetically
- [ ] A user with no organizations (sign in with random new user name) should see appropriate message on  [/organizations](http://localhost:3000/organizations) and  [/dashboard](http://localhost:3000/dashboard)
- [ ] A user with mixed organizations (sign in with `mixed_roles_user`) sees two organizations
   - [ ] The org where they are an admin they see the view details and the manage users links
   - [ ] The org where they are member they only see the view details links
- [ ] Links to view organization details are correct
- [ ] Links to the manage users page are correct 
- [ ] The Activity Dashboard Organizations section appears correct for users as well
- [ ] In the case of error an appropriate error message is shown -- You can do this by adding ` raise_flask_error(500, f"BREAKING API”)` to this [line](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/api/users/user_routes.py#L208)

### DIFF
--- a/frontend/src/app/[locale]/(base)/organizations/layout.tsx
+++ b/frontend/src/app/[locale]/(base)/organizations/layout.tsx
@@ -1,0 +1,25 @@
+import { Metadata } from "next";
+import { LayoutProps } from "src/types/generalTypes";
+import { LocalizedPageProps } from "src/types/intl";
+
+import { getTranslations } from "next-intl/server";
+
+import { AuthenticationGate } from "src/components/user/AuthenticationGate";
+
+export async function generateMetadata({ params }: LocalizedPageProps) {
+  const { locale } = await params;
+  const t = await getTranslations({ locale });
+  const meta: Metadata = {
+    title: t("Organizations.pageTitle"),
+    description: t("Organizations.metaDescription"),
+  };
+  return meta;
+}
+
+export default function OrganizationsLayout({ children }: LayoutProps) {
+  return (
+    <>
+      <AuthenticationGate>{children}</AuthenticationGate>
+    </>
+  );
+}
--- a/frontend/src/app/[locale]/(base)/organizations/page.tsx
+++ b/frontend/src/app/[locale]/(base)/organizations/page.tsx
@@ -1,0 +1,86 @@
+import { UnauthorizedError } from "src/errors";
+import { getSession } from "src/services/auth/session";
+import withFeatureFlag from "src/services/featureFlags/withFeatureFlag";
+import { getUserOrganizations } from "src/services/fetch/fetchers/organizationsFetcher";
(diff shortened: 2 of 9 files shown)


frontend/src/app/[locale]/(base)/organizations/layout.tsx
```diff
@@ -0,0 +1,25 @@
...
+  const meta: Metadata = {
+    title: t("Organizations.pageTitle"),
+    description: t("Organizations.metaDescription"),
+  };
+  return meta;
+}
+
+export default function OrganizationsLayout({ children }: LayoutProps) {
+  return (
+    <>
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/organizations/layout.tsx:21]: no need for the fragment wrapper

frontend/src/app/[locale]/(base)/organizations/page.tsx
```diff
@@ -0,0 +1,86 @@
...
+          {t("errorMessage")}
+        </Alert>
+      </div>
+    </OrganizationsPageWrapper>
+  );
+};
+
+async function OrganizationsPage({ params }: LocalizedPageProps) {
+  const { locale } = await params;
+  setRequestLocale(locale);
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/organizations/page.tsx:44]: something for further discussion, but `setRequestLocale` seems to be only relevant for pages where we're doing static renders. We haven't been too focused on the question of static rendering recently as we've been working with very dynamic content, but it's likely worth having a better shared understanding of where we're at in relation to that aspect of the app.

In any case since this page won't be statically rendered, I think we can remove the call here

myduong-navapbc:
I checked off all the boxes in your validation list. works as intended

Issue #7234, opened 2025-11-25
Fix the public files cloud front issue - unable to download files


prasnava:
The public files endpoint is not returning file not found error. We are unable to download files.

PR #7235, opened 2025-11-25
[Issue #7234]: Hotfix for public files endpoint

prasnava:
## Summary

The public files endpoint in all lower environments are not working. The issue was we were applying to the same origin policy as the frontend and the files endpoint. This was resulting in the aws authorization headers not being passed correctly to the s3 origin. 

## Validation steps

The public files download functionality should continue to work as is.


prasnava:
Deployed and tested the changes in the dev and staging environments. The public files endpoint is working.

mdragon (approved):
Looks good

PR #7246, opened 2025-11-26
[Issue #7081] Create new endpoint for form instructions documents

### LINKED ISSUE #7081: Create new endpoint for form instructions documents (handles s3 + DB)
### Summary

Managing form instructions documents is slow and cumbersome as it requires manually uploading the file to s3 with very specific info, and then manually insert it into the DB. To alleviate this, we want to create a new endpoint and script to automate this process.

We want to create a `PUT /alpha/forms/{form_id}/form_instructions/{form_instruction_id}` endpoint that will take in a file and handle uploading it to s3.

#### Request
The request for this won't quite be JSON as it needs to be able to take in a file, we've done this before with endpoints like our POST /application/../attachments endpoint (can reuse quite a bit of the logic from that probably). The only parameters that the request needs to take

#### Response
A simple success, no info needs to be returned.

#### Auth
This endpoint should ONLY accept `api_user_key_auth` (the other form endpoints accept the older deprecated auth, no need to add that here).
 
Similar to our `PUT /alpha/forms/:form_id` endpoint, we want to require a user have the `UPDATE_FORM` privilege to use the endpoint. Because we aren't putting the other legacy auth, we don't need the very convoluted approach to auth that the form update endpoint had. The auth should also go in the service function, not the route itself.

#### Logic
This is a PUT endpoint and if a form instruction file already exists, we'll want to update it. Roughly the logic should be:
* See if an existing form instruction record exists with the ID provided
* If not - create it
* Upload the file to s3 (copy a similar pattern from the POST app attachment endpoint including how to get the file name). We want to upload this file to `s3://{public_files_bucket_path}/forms/{form_id}/instructions/{file_name}`
* If the form instruction record already existed, if the file path on s3 wasn't exactly the same, delete the old file path.
* NOTE: At no point do we need to check if the form itself exists, due to the way the data models connect, we actually would use this endpoint first. We only include the form ID in the path so we can put it in the s3 path

This ticket is to create the endpoint.

For context, here’s what it takes to setup forms the first time in an environment: https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2259157056/Setup+Forms+-+Nov+17th+2025

### Acceptance criteria

- [ ] Endpoint created
- [ ] Tests added

mikehgrantsgov:
## Summary

Fixes #7081 

## Changes proposed

Add a new PUT `/alpha/forms/{form_id}/form_instructions/{form_instruction_id}` endpoint to handle uploading and updating form instruction documents.
Implemented `upsert_form_instruction` service to manage S3 file uploads, database updates for FormInstruction
Added tests

## Context for reviewers

This endpoint enables the management of form instruction files. It requires the UPDATE_FORM privilege.

Note: The form_id in this endpoint is used solely for constructing the S3 path (/forms/{form_id}/...). We intentionally do not link the instruction to the Form record or validate the Form's existence at this stage.

## Validation steps

See new tests

### DIFF
--- a/api/src/api/form_alpha/form_route.py
+++ b/api/src/api/form_alpha/form_route.py
@@ -90,7 +90,7 @@ def form_instruction_upsert(
     # Get the file from the validated form data
     file_obj = form_and_files_data["file"]
 
-    user = cast(UserApiKey, api_user_key_auth.current_user).user
+    user = api_user_key_auth.get_user()
 
     with db_session.begin():
         db_session.add(user)
--- a/api/src/auth/api_user_key_auth.py
+++ b/api/src/auth/api_user_key_auth.py
@@ -1,4 +1,5 @@
 import logging
+from typing import cast
 
 from apiflask import HTTPTokenAuth
 from sqlalchemy import select
@@ -8,15 +9,36 @@ import src.util.datetime_util as datetime_util
 from src.adapters import db
 from src.adapters.db import flask_db
 from src.api.route_utils import raise_flask_error
-from src.db.models.user_models import UserApiKey
+from src.db.models.user_models import User, UserApiKey
 from src.logging.flask_logger import add_extra_data_to_current_request_logs
 
 logger = logging.getLogger(__name__)
 
+
+class ApiUserKeyHttpTokenAuth(HTTPTokenAuth):
+    """Custom HTTPTokenAuth that provides typed access to the current user API key and user."""
+
+    def get_user_api_key(self) -> UserApiKey:
+        """Wrapper method around the current_user value to handle type issues.
+
+        Note that this value gets set based on whatever is returned from the method
+        you configure for @<your ApiUserKeyHttpTokenAuth obj>.verify_token
+        """
+        return cast(UserApiKey, self.current_user)
+
+    def get_user(self) -> User:
+        """Get the User associated with the current API key.
+
+        This is a convenience method since we typically only care about
+        the user, not the API key itself.
+        """
+        return self.get_user_api_key().user
+
+
 # Initialize the authorization context for API Gateway key authentication
 # This uses the X-API-Key header which is the standard header that AWS API Gateway
 # forwards when api_key_required is set to true
-api_user_key_auth = HTTPTokenAuth(
+api_user_key_auth = ApiUserKeyHttpTokenAuth(
     "ApiKey", header="X-API-Key", security_scheme_name="ApiUserKeyAuth"
 )
 


api/src/api/form_alpha/form_route.py
```diff
@@ -67,3 +70,43 @@ def form_update(
         form = update_form(db_session, form_id, json_data)
 
     return response.ApiResponse(message="Success", data=form)
+
+
+@form_blueprint.put("/forms/<uuid:form_id>/form_instructions/<uuid:form_instruction_id>")
+@form_blueprint.output(AbstractResponseSchema)
+@api_key_multi_auth.login_required
+@flask_db.with_db_session()
+@form_blueprint.doc(security=api_key_multi_auth_security_schemes)
```
chouinar: For the auth, don't add both, just add our new user API key auth, the other one that requires the strange workaround where we check that it's the first API key defined should not be included.

Only have both in the other endpoint as it was built before we had fully switched over.

api/src/api/form_alpha/form_route.py
```diff
@@ -67,3 +70,43 @@ def form_update(
...
+) -> response.ApiResponse:
+    add_extra_data_to_current_request_logs(
+        {"form_id": form_id, "form_instruction_id": form_instruction_id}
+    )
+    logger.info("PUT /alpha/forms/:form_id/form_instructions/:form_instruction_id")
+
+    # Get the file from the request
+    # The request should be multipart/form-data
+    if "file" not in request.files:
+        raise_flask_error(400, "No file part in the request")
```
chouinar: For the file upload, can we use the same approach we used for uploading files in the application attachment upload endpoint? That uses marshmallow/handles a bunch of the complexities for us.

api/src/api/form_alpha/form_route.py
```diff
@@ -67,3 +70,43 @@ def form_update(
...
+            # Check if user is the internal admin user (auth_token_0)
+            if cast(ApiKeyUser, multi_auth_user.user).username != "auth_token_0":
+                raise_flask_error(403, "Only internal admin users can update form instructions")
+        else:
+            # Temporary auth check until legacy auth is removed
+            user = cast(UserApiKey, multi_auth_user.user).user
+            db_session.add(user)
+            verify_access(user, {Privilege.UPDATE_FORM}, None)
+
+        from src.services.form_alpha.upsert_form_instruction import upsert_form_instruction
```
chouinar: Move to the top
```suggestion
```

api/src/api/form_alpha/form_route.py
```diff
@@ -67,3 +70,43 @@ def form_update(
...
+        multi_auth_user = api_key_multi_auth.get_user()
+        if multi_auth_user.auth_type == AuthType.API_KEY_AUTH:
+            # Check if user is the internal admin user (auth_token_0)
+            if cast(ApiKeyUser, multi_auth_user.user).username != "auth_token_0":
+                raise_flask_error(403, "Only internal admin users can update form instructions")
+        else:
+            # Temporary auth check until legacy auth is removed
+            user = cast(UserApiKey, multi_auth_user.user).user
+            db_session.add(user)
+            verify_access(user, {Privilege.UPDATE_FORM}, None)
```
chouinar: Move this to the service function, we don't need to support both types of auth, so don't need to put it here.

api/src/services/form_alpha/upsert_form_instruction.py
```diff
@@ -0,0 +1,81 @@
...
+    secure_file_name = file_util.get_secure_file_name(file_obj.filename)
+
+    # Check if form instruction exists
+    form_instruction = db_session.get(FormInstruction, form_instruction_id)
+
+    if not form_instruction:
+        form_instruction = FormInstruction(
+            form_instruction_id=form_instruction_id,
+            file_name=file_util.get_file_name(file_obj.filename),
+            file_location="",  # Will be set below
```
chouinar: Just exclude the file location if we always set it below. SQLAlchemy won't complain as long as we don't flush.
```suggestion
```

api/src/services/form_alpha/upsert_form_instruction.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+    Upsert a form instruction.
+    If the form instruction exists, update it.
+    If not, create it.
+    Upload the file to S3.
+    """
+    if file_obj.filename is None:
+        logger.info(
+            "Invalid file name, cannot parse",
+            extra={"submission_issue": SubmissionIssue.INVALID_FILE_NAME},
```
chouinar: This isn't a submission endpoint
```suggestion
```

api/tests/src/api/form_alpha/test_form_instruction_route.py
```diff
@@ -0,0 +1,143 @@
...
+    # Setup user with privilege
+    user = UserFactory.create()
+    InternalUserRoleFactory.create(user=user, role__privileges=[Privilege.UPDATE_FORM])
+    api_key = UserApiKeyFactory.create(user=user)
+
+    form = FormFactory.create()
+    form_instruction_id = uuid.uuid4()
+
+    # Mock S3
+    with patch("src.util.file_util.open_stream") as mock_open_stream:
```
chouinar: We have a mock s3 already in our test fixtures - `mock_s3_bucket` - shouldn't need any special mock for everything to work.

api/tests/src/api/form_alpha/test_form_instruction_route.py
```diff
@@ -0,0 +1,143 @@
...
+    )
+
+
+def test_form_instruction_upsert_create_new(
+    client, db_session: db.Session, form_instruction_file, enable_factory_create
+):
+    # Setup user with privilege
+    user = UserFactory.create()
+    InternalUserRoleFactory.create(user=user, role__privileges=[Privilege.UPDATE_FORM])
+    api_key = UserApiKeyFactory.create(user=user)
```
chouinar: Can avoid creating a user, we have a fixture for making an internal user which has these privileges already. `internal_admin_user` for the user and `internal_admin_user_api_key` for the api key.

api/src/api/form_alpha/form_route.py
mikehgrantsgov: Makes sense, I updated this.

api/src/api/form_alpha/form_route.py
mikehgrantsgov: Added a schema and ported that over.

api/src/api/form_alpha/form_route.py
mikehgrantsgov: I moved the `verify_access` there. Let me know if you wanted the whole block, but the verify check made sense to move.

api/tests/src/api/form_alpha/test_form_instruction_route.py
mikehgrantsgov: Updated!

api/tests/src/api/form_alpha/test_form_instruction_route.py
mikehgrantsgov: Cleaned up the tests in general.

api/src/api/form_alpha/form_route.py
```diff
@@ -67,3 +69,31 @@ def form_update(
...
+) -> response.ApiResponse:
+    add_extra_data_to_current_request_logs(
+        {"form_id": form_id, "form_instruction_id": form_instruction_id}
+    )
+    logger.info("PUT /alpha/forms/:form_id/form_instructions/:form_instruction_id")
+
+    # Get the file from the validated form data
+    file_obj = form_and_files_data["file"]
+
+    user = cast(UserApiKey, api_user_key_auth.current_user).user
```
chouinar: Hmm, similar to what we have for the JWT key, we should probably make a function on the `api_user_key_auth` to hide this cast away.

https://github.com/HHS/simpler-grants-gov/blob/main/api/src/auth/jwt_user_http_token_auth.py

Just make something similar to that and have api_user_key_auth derive from it. Also, probably should just have a function that returns the user itself since we never care about the key/jwt.

api/src/services/form_alpha/upsert_form_instruction.py
```diff
@@ -0,0 +1,85 @@
...
+    # s3://{public_files_bucket_path}/forms/{form_id}/instructions/{file_name}
+    s3_config = S3Config()
+    base_path = s3_config.public_files_bucket_path
+
+    new_s3_location = file_util.join(
+        base_path, "forms", str(form_id), "instructions", secure_file_name
+    )
+
+    # If updating, check if we need to delete old file
+    if form_instruction.file_location and form_instruction.file_location != new_s3_location:
```
chouinar: Nitpicky: the second check in the if-statement is fine to do if the first part is null since `new_s3_location` won't ever be null. At worst the second part would be `if None != "s3://..."

```suggestion
    if form_instruction.file_location != new_s3_location:
```

api/src/api/form_alpha/form_route.py
mikehgrantsgov: Added a helper function and using that here.

Issue #7183, opened 2025-11-20
[BUG] Align "Download Instructions" button with "Submit Application" button on Application Page per Figma design


chris-kuryak:
### Summary

The "Submit Application" button and "Download Instructions" button are not aligned on the Application page.

<img width="1063" height="211" alt="Image" src="https://github.com/user-attachments/assets/96aba45d-ca0c-4b2d-8e90-fca963fe50f0" />

They should be aligned per the [Figma design](https://www.figma.com/design/zPlDWcYM396szw2yL3UXvk/Simpler-Grants-2025?node-id=1497-21194&m=dev):

<img width="872" height="209" alt="Image" src="https://github.com/user-attachments/assets/7e5abd23-5952-47ff-8783-11c82e88684b" />

### Acceptance criteria

- [ ] "Download Instructions" and "Submit Application" button on Application Page are aligned per design
- [ ] Design UAT passed

chris-kuryak:
UAT passed

PR #7248, opened 2025-11-26
add more margin to align with the submit button

chay-REIsys:
## Summary

Fixes / Work for #7183 

## Changes proposed
The "Submit Application" button and "Download Instructions" button are not aligned on the Application page.

## Context for reviewers

The "Submit Application" button and "Download Instructions" button are not aligned on the Application page.
Aligning both buttons to appear on the same line for this i have updated the bootstrap class to add more padding.

## Validation steps

when submitting an application download application instructions and submit application are not aligned. this has been updated

### DIFF
--- a/frontend/src/components/application/InformationCard.tsx
+++ b/frontend/src/components/application/InformationCard.tsx
@@ -87,7 +87,7 @@ export const InformationCard = ({
 
   const ApplicationInstructionsDownload = () => {
     return (
-      <div className="margin-top-4">
+      <div className="margin-top-5">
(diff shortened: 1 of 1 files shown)


ErinPattisonNava:
@chay-REIsys We typically do PRs from branches rather than forks, as I need to pull down your code in order to test.  Can you change this to be off a branch?  The branch should be called {your user name}/{ticket number}-{description}

You'll also need to change the title of the PR to be `[ISSUE #{ticketnumber}]: {description}`

It should only be WORK or FIXES, not both.

Issue #7036, opened 2025-11-12
Upgrade Trussworks to address bugs


doug-s-nava:
### Summary

Trussworks has released a version of react-uswds that solves the issue mentioned in https://github.com/HHS/simpler-grants-gov/issues/5724

To pull in the fix we need to upgrade to >= 10.2.  (10.1.0 had a bug that got in the way of installing the package. See https://github.com/trussworks/react-uswds/issues/3322)

~~Once this bug has been addressed~~, we should upgrade to the latest version (after bug addressed in https://github.com/trussworks/react-uswds/pull/3323#event-20918376535)

### Acceptance criteria

- [ ] app is using @trussworks/react-uswds > 10.1
- [ ] installation of the package goes through without errors

andycochran:
Related to #4871

PR #7251, opened 2025-11-26
[Issue #7036] upgrade trussworks to fix file upload dev bug

doug-s-nava:
## Summary

Work for #7036 

## Changes proposed

Upgrades trussworks/uswds package in order to fix bug identified in https://github.com/HHS/simpler-grants-gov/issues/5724, which was preventing testing of file uploads in normal NextJS dev mode.

## Context for reviewers

Note that there are not many changes in this upgrade, but a general regression test of the site to make sure nothing is broken is a good idea.

## Validation steps

1. start a local server on this branch using `npm run dev`
2. start an application using this [readme](https://github.com/HHS/simpler-grants-gov/blob/main/frontend/src/app/%5Blocale%5D/workspace/applications/application/%5BapplicationId%5D/README.md)
3. open the budget narrative attachment form
4. upload a file
5. _VERIFY_: file uploads as expected without any errors

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -14,7 +14,7 @@
         "@next/third-parties": "^15.1.4",
         "@opentelemetry/api": "^1.8.0",
         "@rjsf/utils": "^5.24.8",
-        "@trussworks/react-uswds": "github:trussworks/react-uswds#58f3089739ac5ee415c11cfbe179daad7a62597",
+        "@trussworks/react-uswds": "^10.2.0",
         "@uswds/uswds": "^3.7.1",
         "@zip.js/zip.js": "^2.7.60",
         "ajv": "^8.17.1",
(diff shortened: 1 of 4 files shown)


andycochran:
This change bumps Truss to require USWDS `^3.8.2`. Our dependencies explicitly list `^3.7.1`. The latest USWDS is `3.13.0`. Are we getting the latest with that caret symbol? Should we bump that manually in our dependencies to make sure we get the latest CSS? Should we confirm we don't need to restrict to a specific version?

doug-s-nava:
for whatever reason, it looks like the version actually resolved in our package at this point is 3.11. I am not sure why it isn't pulling 3.13 since the caret version we're specifying should automatically pull new versions up until 4.0. This is confused by the fact that trussworks is pinned at 3.10 for their build. Using 3.11 hasn't caused any trouble for us so far, so I don't imagine specifying pulling 3.13 would break anything in the trussworks library. Would want to test that assumption though.

myduong-navapbc:
Everything looks good:
What I checked:

Build & Tested:
- build works
- tests pass

Components:
- GridContainer, Grid
- Alert
- ErrorMessage
- Button, ButtonGroup
- Breadcrumb,  BreadcrumbBar, BreadcrumbLink
- Footer
- Checkbox, Radio, ComboBox, ComboBoxRef, ComboBoxOption
- ModalRef, ModalToggleButton, ModalToggleButton, ModalFooter
- Table
- Tooltip
- FileInputRef, FileInput, TextArea
- Link
- Label, FormGroup, FieldSet, Select, TextInput
- Pagination
- Accordion

PR #7252, opened 2025-12-01
[Issue #7157] XML for SF-LLL Form matches legacy Grants.gov

### LINKED ISSUE #7157: XML for SF-LLL Form matches legacy Grants.gov
### Summary

This ticket is to ensure that XML matches legacy Grants.gov XML.

https://www.grants.gov/forms/form-items-description/fid/670

Example to be provided by Michael C.

### Acceptance criteria

- [ ] XML generation matches existing Grants.gov XML output
- [ ] Update documentation with any noteworthy transformation details: https://github.com/HHS/simpler-grants-gov/tree/main/api/src/services/xml_generation

mikehgrantsgov:
## Summary

Fixes #7157 

## Changes proposed

Add SF-LLL form schema for XML generation
Added a `transform_map_values()` function for configurable value mappings
Improved base transformer for complex nested structures
Added form type registry

## Context for reviewers

Had to make a few changes to base transformer to satisfy some of the nuance of the schema and some changes might be needed when we see a real example, but we are producing XSD-conforming XML. 

## Validation steps

See new unit tests and CLI:

```
make fetch-xsds form=SFLLL_2_0
make test-xml-validation form=SFLLL_2_0
```

### DIFF
--- a/api/src/services/xml_generation/config.py
+++ b/api/src/services/xml_generation/config.py
@@ -3,27 +3,29 @@
 import logging
 from typing import Any
 
-from src.form_schema.forms.sf424 import FORM_XML_TRANSFORM_RULES as SF424_RULES
-from src.form_schema.forms.sf424a import FORM_XML_TRANSFORM_RULES as SF424A_RULES
-from src.form_schema.forms.sflll import FORM_XML_TRANSFORM_RULES as SFLLL_RULES
+from src.form_schema.forms import get_active_forms
 
 logger = logging.getLogger(__name__)
+
+
+def _build_xml_form_map() -> dict[str, dict[str, Any]]:
+    """Build a dynamic map of form names to their XML transformation rules."""
+    xml_form_map: dict[str, dict[str, Any]] = {}
+    for form in get_active_forms():
+        if form.json_to_xml_schema is not None:
+            xml_form_map[form.short_form_name] = form.json_to_xml_schema
+    return xml_form_map
 
 
 def load_xml_transform_config(form_name: str) -> dict[str, Any]:
     """Load XML transformation rules for a given form."""
     try:
+        xml_form_map = _build_xml_form_map()
         form_name_upper = form_name.upper()
 
-        if form_name_upper == "SF424_4_0":
+        if form_name_upper in xml_form_map:
             logger.info(f"Loaded transformation config for {form_name}")
-            return SF424_RULES
-        elif form_name_upper == "SF424A":
-            logger.info(f"Loaded transformation config for {form_name}")
-            return SF424A_RULES
-        elif form_name_upper == "SFLLL_2_0":
-            logger.info(f"Loaded transformation config for {form_name}")
-            return SFLLL_RULES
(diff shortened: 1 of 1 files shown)


api/src/form_schema/forms/sflll.py
```diff
@@ -491,6 +497,461 @@
     },
 }
 
+# NOTE: This is an initial implementation that provides the basic structure for XML generation.
+# Some complex structures require further refinement once we have example legacy XML to validate against:
+#
+# TODO:
+# - Validate XML output against actual legacy Grants.gov XML examples
+#
```
chouinar: There are examples on https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2189590562/Form+XML+Comparison#SFLLL-2.0 - is there a scenario you are missing/need more examples on?

api/src/form_schema/forms/sflll.py
```diff
@@ -491,6 +497,461 @@
...
+                        "target": "SuffixName",
+                        "namespace": "globLib",
+                    }
+                },
+            },
+            "address": {
+                "xml_transform": {
+                    "target": "Address",
+                    "type": "nested_object",
+                },
```
chouinar, line 855: Not something we need in this PR, but maybe a future improvement - for common data types like address/name, might be useful to have a shortcut for transforming them rather than needing to copy-paste the same 5~8 fields each time.

api/src/services/xml_generation/config.py
```diff
@@ -14,9 +16,14 @@ def load_xml_transform_config(form_name: str) -> dict[str, Any]:
...
-
             logger.info(f"Loaded transformation config for {form_name}")
-            return FORM_XML_TRANSFORM_RULES
+            return SF424_RULES
+        elif form_name_upper == "SF424A":
+            logger.info(f"Loaded transformation config for {form_name}")
+            return SF424A_RULES
+        elif form_name_upper == "SFLLL_2_0":
+            logger.info(f"Loaded transformation config for {form_name}")
+            return SFLLL_RULES
```
chouinar: There are a few places that have these configs, one way we could make this dynamic is to generate it when we run the code. There is a [get_active_forms](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/form_schema/forms/__init__.py) function to get every form, could do something like:

```python
xml_form_map = {}
for form in get_active_forms():
    if form.json_to_xml_schema is not None:
        xml_form_map[form.short_form_name] = form.json_to_xml_schema
```

mikehgrantsgov: Makes sense, updated!

api/src/form_schema/forms/sflll.py
mikehgrantsgov: I agree. Marked this down for a future enhancement.

api/src/form_schema/forms/sflll.py
mikehgrantsgov: Ah, just saw these now. I think we are good!

Issue #7093, opened 2025-11-17
Upgrade to latest Python version for the API


ychoy:
### Summary

Upgrade to latest Python version 3.14

### Reference
https://github.com/HHS/simpler-grants-gov/blob/main/documentation/api/package-dependency-management.md#upgrading-python

### Acceptance criteria
- [ ] Update https://github.com/HHS/simpler-grants-gov/blob/main/documentation/api/package-dependency-management.md#upgrading-python documentation (re: Dockerfile)
- [ ] Upgraded to latest Python version
- [ ] API and Python scripts functions as expected

PR #7268, opened 2025-12-01
[Issue #7093]: Upgrade API python version to 3.14

prasnava:
## Summary

Upgraded the API's python version to 3.14. 

Will create another pull request / ticket for analytics. It's not using the python base image yet. 

## Validation steps

The api and step functions should continue to function.

### DIFF
--- a/api/src/db/models/opportunity_models.py
+++ b/api/src/db/models/opportunity_models.py
@@ -60,7 +60,7 @@ class Opportunity(ApiSchemaTable, TimestampMixin):
     revision_number: Mapped[int | None]
     modified_comments: Mapped[str | None]
 
-    opportunity_attachments: Mapped[list["OpportunityAttachment"]] = relationship(
+    opportunity_attachments: Mapped[list[OpportunityAttachment]] = relationship(
         back_populates="opportunity", uselist=True, cascade="all, delete-orphan"
     )
 


prasnava:
Successfully deployed to dev: [here](https://github.com/HHS/simpler-grants-gov/actions/runs/19833831813)

.dockleconfig
```diff
@@ -4,4 +4,4 @@
...
+DOCKLE_ACCEPT_FILES=api/.venv/lib/python3.13/site-packages/newrelic/api/settings.py,analytics/.venv/lib/python3.13/site-packages/newrelic/api/settings.py,api/.venv/lib/python3.13/site-packages/apiflask/settings.py,analytics/.venv/lib/python3.13/site-packages/jedi/settings.py,api/.venv/lib/python3.13/site-packages/docx/oxml/settings.py,api/.venv/lib/python3.13/site-packages/docx/parts/settings.py,api/.venv/lib/python3.13/site-packages/docx/settings.py,app/nofos/bloom_nofos/settings.py,app/.venv/lib/python3.12/site-packages/easyaudit/settings.py,app/.venv/lib/python3.12/site-packages/constance/settings.py,app/.venv/lib/python3.12/site-packages/isort/settings.py,app/.venv/lib/python3.12/site-packages/martor/settings.py,app/.venv/lib/python3.12/site-packages/cssutils/settings.py,api/.venv/lib/python3.14/site-packages/newrelic/api/settings.py,api/.venv/lib/python3.14/site-packages/apiflask/settings.py,api/.venv/lib/python3.14/site-packages/docx/oxml/settings.py,api/.venv/lib/python3.14/site-packages/docx/parts/settings.py,api/.venv/lib/python3.14/site-packages/docx/settings.py
```
chouinar: We should be able to remove all of the 3.12/3.13 paths from this. They haven't happened to add wildcard support for this since I last checked, have they? Would be great if we could do `api/.venv/lib/python3.*/site-packages/...`

api/pyproject.toml
```diff
@@ -9,7 +9,7 @@ include = ["newrelic.ini"]
 [tool.poetry.dependencies]
 # See /documentation/api/package-depedency-management.md#Upgrading Python
 # for details on upgrading your Python version
-python = "~3.13"
+python = "~3.14"
```
chouinar, line 12: Threading arbitrarily, were there any details that should be added to the upgrade docs https://github.com/HHS/simpler-grants-gov/blob/2818fbb7e0413f31cb4e22b13858bf35c6b692f6/documentation/api/package-dependency-management.md#upgrading-python

Based on what you have here, maybe mentioning that the linter/formatter should be run would be a good idea. Also the base image adjustments.

.dockleconfig
prasnava: yep, will update the file.

api/pyproject.toml
prasnava: Yep, will update the upgrade docs.

.dockleconfig
prasnava: Updated the dockle config, we still have python version 3.13 in there because of the analytics service. Will upgrade analytics in another pull request. An alternative I found that works is instead of manually using the file paths we can just use DOCKLE_ACCEPT_FILES=settings.py and the dockleconfig would work.

api/pyproject.toml
prasnava: Updated the docs.

chouinar (dismissed):
Changes look good - if we're about to hotfix to prod though we should hold. Whenever we merge this, should make sure we have a few days before we go to prod.

Also probably should make a callout post in the engineering channel as it'll potentially break some peoples local setup depending on how they've installed python.

PR #7273, opened 2025-12-01
[Issue: #7231] adding unittests for header and organizations work

### LINKED ISSUE #7231: create unittests for /organizations page
### Summary

Because the `/organizations` page is behind a feature flag, I couldn't find sufficient examples of unittest to be successful.  Adding a ticket to follow up later so the majority of the work can get in before the end of sprint. 

### Acceptance criteria

- [ ] Full unittests for the OrganizationsPage
- [ ] Full unittests for the feature flag gated parts of the header

ErinPattisonNava:
## Summary

Fixes #7231  

## Changes proposed

Added utilities for testing with feature flags, and examples around the organizations work

## Context for reviewers

Last week I had to get in a PR for the organizations page, but didn't have time to tackle the complexity of the unittests.  Shout out to @myduong-navapbc for helping me out with 

## Validation steps

- [ ] If the tests run successfully, we're good

### DIFF
--- a/frontend/tests/pages/organizations/page.test.tsx
+++ b/frontend/tests/pages/organizations/page.test.tsx
@@ -2,10 +2,13 @@ import { render, screen } from "@testing-library/react";
 import OrganizationsPage from "src/app/[locale]/(base)/organizations/page";
 import { Organization } from "src/types/applicationResponseTypes";
 import { LocalizedPageProps } from "src/types/intl";
+import { FeatureFlaggedPageWrapper } from "src/types/uiTypes";
 import { UserDetail } from "src/types/userTypes";
 import { wrapForExpectedError } from "src/utils/testing/commonTestUtils";
 
-import { FunctionComponent } from "react";
+import { FunctionComponent, ReactNode } from "react";
+
+type onEnabled = (props: LocalizedPageProps) => ReactNode;
 
 const redirectMock = jest.fn();
 
@@ -61,11 +64,15 @@ jest.mock("src/services/featureFlags/withFeatureFlag", () => ({
     (
       WrappedComponent: FunctionComponent<LocalizedPageProps>,
       featureFlagName: string,
-      onEnabled: () => void,
+      onEnabled: onEnabled,
     ) =>
     (props: LocalizedPageProps) =>
-      // eslint-disable-next-line @typescript-eslint/no-unsafe-call
-      withFeatureFlagMock(
+      (
+        withFeatureFlagMock as FeatureFlaggedPageWrapper<
+          LocalizedPageProps,
+          ReactNode
+        >
+      )(
         WrappedComponent,
         featureFlagName,
         onEnabled,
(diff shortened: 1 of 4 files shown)


frontend/tests/pages/organizations/page.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+  getUserOrganizations: () => organizations() as Promise<Organization[]>,
+}));
+
+jest.mock("src/components/workspace/UserOrganizationsList", () => ({
+  UserOrganizationsList: () => <div data-testid="user-org-list" />,
+}));
+
+const organizationsPageLocation = "src/app/[locale]/(base)/organizations/page";
+
+describe("Organizations page feature flag wiring", () => {
```
doug-s-nava, line 82: I wonder if some of this complexity could be dealt with more simply by testing the raw page component, unwrapped by the `withFeatureFlag` function

frontend/tests/utils/featureFlagUtils.tsx
```diff
@@ -0,0 +1,85 @@
...
+      WithFeatureFlagArgs
+    >;
+
+  withFeatureFlagMock.mockImplementation(
+    (wrappedComponent: unknown, _flagName: string, _onDisabled: () => void) =>
+      wrappedComponent,
+  );
+
+  const navigationModule = await import("next/navigation");
+  const redirectMock = navigationModule.redirect as unknown as jest.Mock<
```
doug-s-nava: why does it not work to reference the mock function directly?

frontend/tests/pages/organizations/page.test.tsx
```diff
@@ -0,0 +1,96 @@
+import { render, screen } from "@testing-library/react";
```
doug-s-nava, line 1: you might notice that there aren't a lot of page tests in the system, and this is kind of by design. At some point we decided that in most cases it was not worth the effort to run unit tests on the page level. The mocking here isn't super complex, though, so this seems like a decent thing to test. The alternative would, I suppose, be removing the business logic / fetch logic into a child component. We can talk more about what our standards should be around page components and testing, but I'm good with these tests existing here!!!

ErinPattisonNava: Given how much functionality we've put into several of our pages, it feels like we should be able to do that without having to pull out subcomponents (especially as all our components currently live in a different directory and are automatically considered "shared").    Happy to have the conversation about standards, but I think we need a way to unittest everything in our system.

frontend/tests/pages/organizations/page.test.tsx
ErinPattisonNava: I did a quick test, and even if I export the unwrapped component in addition to the default page jest has a heart attack and says it can't parse things.  Happy to show you tomorrow

frontend/tests/utils/featureFlagUtils.tsx
ErinPattisonNava: fixed!

Issue #6864, opened 2025-10-27
wire up invitation acceptance


doug-s-nava:
### Summary

after https://github.com/HHS/simpler-grants-gov/issues/6648 and https://github.com/HHS/simpler-grants-gov/issues/6649 we will need to connect the frontend to the api

### Acceptance criteria

- [ ] frontend invitation accept and reject actions communicate with api to persist selections and add user to organization when applicable
- [ ] reject and accept confirmation message flavors of the widget are hidden after a 5 second timeout

joshtonava:
I dont think this is supposed to be assigned to me ?

cc: @doug-s-nava

ErinPattisonNava:
@doug-s-nava Since the backend work is done, would it be best to roll the hook-up logic into #6648 so we can test end to end?

doug-s-nava:
@ErinPattisonNava yep, this is unblocked

eric-serrato95:
@myduong-navapbc UAT complete moving to done

eric-serrato95:
@ErinPattisonNava @myduong-navapbc @joshtonava 
Iooks like when I tested this in staging, accepting the invite does not actually add me, at least not using our test orgs in staging, 
can any of you confirm if this issue is staging specific or if this is going to extend to the prod version?

chris-kuryak:
Per Eric and DOug, closing

PR #7279, opened 2025-12-01
[Issue #6864] wire up organization invitiation response

doug-s-nava:
## Summary

Work for #6864 

## Changes proposed

Hooks up activity dashboard page to invitations endpoints to allow for accepting or rejecting invitations

## Context for reviewers

Note that these validation steps are a little involved due to a bug found here: https://github.com/HHS/simpler-grants-gov/issues/7287. These steps work around the bug where emails assigned to test users are overwritten by the default 0auth email address on login by logging in first, then updating the email address

Also note that a refresh is required after accepting the invitations before they will show up on the dashboard

## Validation steps

1. start a local server with npm run dev
2. visit http://localhost:3000?_ff=manageUsersOff:false;userAdminOff:false
5. log in as two_org_user
6. visit the manage users page for any of the orgs you're a member of
7. invite an email to the organization as an admin
4. visit the manage users page for the other org you're a member of
6. invite the same email to the organization as an member
8. log out
3. log in as new user, enter anything you'd like in the user name field
4. find your user in the line_external_user table and update the email address to match the email you invited
5. refresh the page
10. _VERIFY_: you see invitiations to two organizations
11. accept both invitiations
12. refresh
13. _VERIFY_: you see both organizations in the list on the dashboard page
14. _VERIFY_: the organization to which you were added as admin has a "manage users" button, and the other one doesn't

### DIFF
--- a/frontend/src/services/auth/session.ts
+++ b/frontend/src/services/auth/session.ts
@@ -86,7 +86,6 @@ export const getSession = async (): Promise<UserSession | null> => {
   }
   const { token, exp } = payload;
   const session = await decryptLoginGovToken(token);
-  console.log(token);
   return session
     ? {
         ...session,
--- a/frontend/tests/api/user/organization-invitations/route.test.ts
+++ b/frontend/tests/api/user/organization-invitations/route.test.ts
@@ -3,11 +3,8 @@
  */
 
 import { updateOrganizationInvitation } from "src/app/api/user/organization-invitations/[organizationInvitationId]/handler";
-import { OrganizationInvitation, UserOrganization } from "src/types/userTypes";
-import {
-  fakeOrganizationInvitation,
-  fakeUserOrganization,
-} from "src/utils/testing/fixtures";
+import { OrganizationInvitation } from "src/types/userTypes";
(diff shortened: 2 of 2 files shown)


frontend/src/app/[locale]/(base)/dashboard/page.tsx
```diff
@@ -72,8 +72,10 @@ async function ActivityDashboard() {
           ),
         })}
       </h1>
-      {userInvitations?.length && (
+      {userInvitations?.length ? (
```
Comment (glenblosser-nava) [inline frontend/src/app/[locale]/(base)/dashboard/page.tsx:75]: Is there a standard for when we use one or the other? I've seen both forms used throughout the project

Comment (doug-s-nava) [reply frontend/src/app/[locale]/(base)/dashboard/page.tsx:75]: I think it's whichever works best or is easiest to read. In this case unfortunately the `&&` version puts a `0` on your screen when there aren't any invitations so 😬

Comment (ErinPattisonNava) [reply frontend/src/app/[locale]/(base)/dashboard/page.tsx:75]: In a past team we normed on always using the ternary operator because the && can produce unexpected results.  I'd love it if that could be the standard here too.

PR #7280, opened 2025-12-01
[Issue  #7183] Change to align Instructions and submit button

chay-REIsys:
Summary
Fixes for https://github.com/HHS/simpler-grants-gov/issues/7183

Changes proposed
The "Submit Application" button and "Download Instructions" button are not aligned on the Application page.

Context for reviewers
The "Submit Application" button and "Download Instructions" button are not aligned on the Application page.
Aligning both buttons to appear on the same line for this i have updated the bootstrap class to add more padding.

Validation steps
when submitting an application download application instructions and submit application are not aligned. this has been updated

### DIFF
--- a/frontend/src/components/application/InformationCard.tsx
+++ b/frontend/src/components/application/InformationCard.tsx
@@ -85,17 +85,13 @@ export const InformationCard = ({
   const hasOrganization = Boolean(applicationDetails.organization);
   const { is_open } = applicationDetails.competition;
 
-  const ApplicationInstructionsDownload = ({
-    hasOrganization,
-  }: {
-    hasOrganization: boolean;
-  }) => {
+  const ApplicationInstructionsDownload = () => {
     return (
-      <div className={hasOrganization ? "margin-top-auto" : "margin-top-5"}>
+      <div className="margin-bottom-1 margin-left-0">
         <dt className="usa-sr-only">
           {t("applicationDownloadInstructionsLabel")}:{" "}
         </dt>
-        <dd>
+        <dd className="margin-left-0">
           {instructionsDownloadPath && (
             <Link href={instructionsDownloadPath}>
               <Button
@@ -115,7 +111,7 @@ export const InformationCard = ({
 
   const NoApplicationInstructionsDownload = () => {
     return (
-      <div className="margin-bottom-1">
+      <div className="margin-bottom-1 margin-left-0">
         <dt className="margin-right-1 text-bold">
           {t("applicationDownloadInstructionsLabel")}:{" "}
         </dt>
@@ -178,13 +174,6 @@ export const InformationCard = ({
               hasOrganization={hasOrganization}
               samGovEntity={applicationDetails.organization?.sam_gov_entity}
             />
-            {applicationDetails.competition.competition_instructions.length ? (
-              <ApplicationInstructionsDownload
-                hasOrganization={hasOrganization}
-              />
-            ) : (
-              <NoApplicationInstructionsDownload />
-            )}
           </dl>
         </Grid>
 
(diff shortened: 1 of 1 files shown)


ErinPattisonNava:
The spacing looks ok if you are applying as an individual, but is incorrect if you apply as an organization.  I'm wondering if we need to change the grid layout of that section to keep the buttons in one row, but maybe just setting the margin-top-auto would work.  Don't forget to check mobile in that case and double check with design if it shifts there.  

Screenshot 2025-12-01 at 3 53 20 PM

chaitanya-duddella:
image
image

frontend/src/components/application/InformationCard.tsx
```diff
@@ -175,7 +179,9 @@ export const InformationCard = ({
               samGovEntity={applicationDetails.organization?.sam_gov_entity}
             />
             {applicationDetails.competition.competition_instructions.length ? (
-              <ApplicationInstructionsDownload />
+              <ApplicationInstructionsDownload
+                hasOrganization={hasOrganization}
+              />
             ) : (
               <NoApplicationInstructionsDownload />
```
ErinPattisonNava: I'm guessing this usecase would have the same problem, could you fix it for both?

frontend/src/components/application/InformationCard.tsx
```diff
@@ -85,9 +85,13 @@ export const InformationCard = ({
...
 
-  const ApplicationInstructionsDownload = () => {
+  const ApplicationInstructionsDownload = ({
+    hasOrganization,
+  }: {
+    hasOrganization: boolean;
+  }) => {
     return (
-      <div className="margin-top-4">
+      <div className={hasOrganization ? "margin-top-auto" : "margin-top-5"}>
```
ErinPattisonNava: This doesn't feel like a good responsive design pattern, as if other information gets added the buttons will get out of sync again.  Can you make it more of a generalize pattern?

chaitanya-duddella: pushed latest changes..

ErinPattisonNava:
Thanks for making the changes @chay-REIsys !  This is looking in a better direction.  Two things:

1) Can you do a design review with @jesababon to make sure that she's aware of the different mobile button layout?  I think the change is fine especially as we don't prioritize the mobile experience, but it would be a good reason to checking in and start building that relationship.

2) @doug-s-nava, I don't know how to test the `NoApplicationInstructionsDownload` use case on localhost.  Do we seed a specific application like that?

doug-s-nava:
It doesn't look like there is a competition seeded without instructions. To test that you could either:
* remove a row from `competition_instruction`
* if all you want to do is check the UI, edit the frontend code temporarily to always display the "no instructions" version

chaitanya-duddella:
> Thanks for making the changes @chay-REIsys ! This is looking in a better direction. Two things:
> 
> 1. Can you do a design review with @jesababon to make sure that she's aware of the different mobile button layout?  I think the change is fine especially as we don't prioritize the mobile experience, but it would be a good reason to checking in and start building that relationship.
> 2. @doug-s-nava, I don't know how to test the `NoApplicationInstructionsDownload` use case on localhost.  Do we seed a specific application like that?

@ErinPattisonNava  i have tested NoApplicationInstructionDownload on local by switching the condition though, not with actual data .. it is just showing '-' symbol

ErinPattisonNava:
> > Thanks for making the changes @chay-REIsys ! This is looking in a better direction. Two things:
> > 
> > 1. Can you do a design review with @jesababon to make sure that she's aware of the different mobile button layout?  I think the change is fine especially as we don't prioritize the mobile experience, but it would be a good reason to checking in and start building that relationship.
> > 2. @doug-s-nava, I don't know how to test the `NoApplicationInstructionsDownload` use case on localhost.  Do we seed a specific application like that?
> 
> @ErinPattisonNava i have tested NoApplicationInstructionDownload on local by switching the condition though, not with actual data .. it is just showing '-' symbol

Wow, that is terrible UI and kinda feels like a bug too.  Can you talk to Jes and see if she wants you to file a separate but on that?

ErinPattisonNava (approved):
Approving conditional that you get a design review and that you determine with design if the no instructions use case should be filed as a separate bug or fixed here, as it looks pretty weird.

chaitanya-duddella:
> Approving conditional that you get a design review and that you determine with design if the no instructions use case should be filed as a separate bug or fixed here, as it looks pretty weird.

image

Issue #7286, opened 2025-12-02
snack time for user invite widget does not appear to work as intended


eric-serrato95:
### Summary

Desired behaviour: 
After a user is successfully invited, the widget should display a short message saying that the user has been invited and keep the email and role visible while the snack time is active
Current behaviour: 
After a user is successfully invited, the widget goes gray still with the information of user, this remains for a bit maybe a second or so, but then a second message saying user has been added succesfully but the information does not remain visible 

### Reproducibility

None

### How to reproduce the bug

- Add a user leveraging the user invite in staging.simpler.grants.gov 

### Browser

None

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

doug-s-nava:
I'd like to clarify requirements here. This is what I've got:
* when a user clicks "add to organization"
* fields and buttons are disabled during loading state
* once loading state clears, on success, button changes to success message. Fields are still disabled, and reset to defaults / placeholders
* after timeout completes (3) seconds, fields and button are re-enabled 

What are the points here that we need to change?

https://github.com/user-attachments/assets/a8ad28f6-371c-4ed4-801e-a2d3ed940767

eric-serrato95:
@doug-s-nava 
Ideally we would only have one state with snacktime, where the checkmark and the message saying that a user has been added is present

doug-s-nava:
the loading state was not entirely clear, but this implementation is fine for now cc @eric-serrato95. closing.

Issue #7285, opened 2025-12-02
Error when trying to add more than 1 user at a time on the manage users page


eric-serrato95:
### Summary

When trying to send an invite to a user leveraging the user invite widget, the first time it works as expected, but after the quick snack time and reset to default state, the widget does not seem to allow to send an invite even with a valid Email and a role selected
The page has to be fully refreshed to add an additional user 

### Reproducibility

None

### How to reproduce the bug

- Go to staging.simpler.grants.gov 
- log in 
- Access to one of your test organizations 
- go to manage users 
- use the user invite to send an invite to a user
- wait for the widget to reset 
- try to add another user once the widget resets

### Browser

Chrome

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

eric-serrato95:
@ErinPattisonNava @doug-s-nava first issue of the bugs found in the manage users table

PR #7308, opened 2025-12-02
[Issue #7285] invite organization users bugfixes

doug-s-nava:
## Summary

Work for #7285 and maybe #7286

## Changes proposed

Fixes bugs on organization invitation form where:

* on second submission using the same form, error was received about "role" parameter not being supplied
* on successful submission, input values were not reset to default
* on second submission, success message was not displayed

## Context for reviewers

Underlying bugs were:

* state of selected role in organization invitation form was not being properly tracked
* form state was not being properly reset on success
* success state was not reset when making a second request with the form

## Validation steps

1. start a local server on this branch with npm run dev
2. visit http://localhost:3000/?_ff=userAdminOff:false;manageUsersOff:false
3. log in as two_org_user
4. vist the user management page for any organization
5. invite a user
6. _VERIFY_: after sending invite, form fields are returned to their default state
7. invite another user
8. _VERIFY_: invitiation goes through without error
9. _VERIFY_: success message is displayed

### DIFF
--- a/frontend/src/app/[locale]/(base)/organization/[id]/manage-users/actions.ts
+++ b/frontend/src/app/[locale]/(base)/organization/[id]/manage-users/actions.ts
@@ -41,6 +41,7 @@ export const inviteUserAction = async (
   _prevState: unknown,
   formData: FormData,
   organizationId: string,
+  genericErrorMessage: string,
 ): Promise<OrganizationInviteResponse> => {
   const session = await getSession();
 
@@ -80,7 +81,7 @@ export const inviteUserAction = async (
       `Error inviting user to org - ${error.message} ${error.cause?.toString() || ""}`,
     );
     return {
-      errorMessage: error.message,
+      errorMessage: `${genericErrorMessage}: ${error.message}`,
     };
   }
 };
--- a/frontend/src/components/manageUsers/UserInviteForm.tsx
+++ b/frontend/src/components/manageUsers/UserInviteForm.tsx
@@ -34,8 +34,9 @@ const OrganizationInviteValidationError =
 
 // in order to pass the organizationId into the form action call
 export const inviteUserActionForOrganization =
-  (organizationId: string) => (_prevState: unknown, formData: FormData) =>
-    inviteUserAction(_prevState, formData, organizationId);
+  (organizationId: string, errorMessage: string) =>
+  (_prevState: unknown, formData: FormData) =>
+    inviteUserAction(_prevState, formData, organizationId, errorMessage);
 
 export const RoleOptions = ({ roles }: { roles: UserRole[] }) => {
   const t = useTranslations("ManageUsers.inviteUser.inputs.role");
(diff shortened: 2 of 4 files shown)


frontend/tests/pages/organization/manage-users/actions.test.ts
```diff
@@ -48,7 +48,7 @@ describe("user profile form action", () => {
   it("returns result of update on success", async () => {
     getSessionMock.mockResolvedValue({ token: "logged in", user_id: "1" });
     mockCreateInvitation.mockImplementation((_token, data) => {
-      return data as unknown;
+      return { ...data, organization_invitation_id: "fun_id" } as unknown;
```
ErinPattisonNava, line 66: Fun ids are the best

frontend/src/components/manageUsers/UserInviteForm.tsx
```diff
@@ -85,22 +92,39 @@ export function UserInviteForm({
...
+  const [selectedRole, setSelectedRole] = useState<string | undefined>();
 
   const inviteUser = useMemo(
     () => inviteUserActionForOrganization(organizationId),
     [organizationId],
   );
 
   const [formState, formAction, isPending] = useActionState(inviteUser, {
-    success: false,
+    invitationCreated: "",
```
ErinPattisonNava: Nit: Could we set this to undefined?

ErinPattisonNava (commented):
We can now add multiple users in a row!  Also the a11y issues around the role dropdown seem to have been magically fixed.  Some other things I found:

1) The success state says "user added to pending" but we don't have a pending section anymore.  Could we update that text to match the UI?

2) After we click "Add to organization", a bunch of things flash on the screen

a) the text buttons are first grayed out with their current content and the "add to organization" button is disabled
b) The text buttons are back enabled but blank and the "add to organization" button is enabled 
c) then it flashes to the buttons grayed out but with the default content with the success message
d) then it flashes to the reset form ready for input

Can we get rid of step (b) at least, and maybe (a) especially as errors reset the fields anyways?  All the flashing makes me think as a user that I've missed something or that the page is behaving unexpectedly, it doesn't have a super solid feel.  

3) When we get an error from the backend, we just feed that text directly into the error message even it doesn't make sense to a user.  It would be nicer to have user-friendly error message instead of a technical one.

Let me know if you're cool fixing these as part of this ticket or if you'd rather spin them off into their own tickets.

doug-s-nava:
I've fixed up:
* the undefined nit
* the "pending" message
* added a generic error message in addition to the message returned by the backend

The flash before the success message is some react timing stuff that I'd rather not get into the weeds with on this PR, can make a separate ticket.

For anything beyond this I think I need some guidance from design, so maybe a new ticket is in order for followups.

@jesababon @chelsea-nava how do you want to deal with these points:

* language for confirmation message. For now I've replaced "user added to pending" with "User invited to organization"
* language for error handling. This is a bit complex since sometimes the error we get back from the server action is human readable and actionable, such as when an existing user has been reinvited. In other cases it will be less readable. For now I have a generic message prefixing any message coming from the server as in the screenshot below
* loading state. I feel like this component needs a loading state, but the one I've implemented is evidently confusing - both Erin and Eric mentioned it. What should that look like?

frontend/src/app/[locale]/(base)/organization/[id]/manage-users/actions.ts
```diff
@@ -70,14 +71,17 @@ export const inviteUserAction = async (
...
+      invitationCreated: inviteUserResponse.organization_invitation_id,
+    };
   } catch (e) {
     const error = e as Error;
     console.error(
       `Error inviting user to org - ${error.message} ${error.cause?.toString() || ""}`,
     );
     return {
-      errorMessage: error.message,
+      errorMessage: `${genericErrorMessage}: ${error.message}`,
```
Comment (ErinPattisonNava) [inline frontend/src/app/[locale]/(base)/organization/[id]/manage-users/actions.ts:84]: I'd personally recommend just showing the generic error message and swallowing the internal one, but I leave that call up to design.

doug-s-nava:
here's a very rough design follow up ticket https://github.com/HHS/simpler-grants-gov/issues/7352

Issue #6816, opened 2025-10-21
Modify the SOAP Proxy endpoints to use a new env variable for managing private keys


chris-kuryak:
### Summary

We want to add a new environment variable for storing private keys. This is only needed temporarily until [grants.gov](http://grants.gov/) makes changes that won’t require we work with the private keys at all.

We already have a environment variable called `SOAP_AUTH_CONTENT` that we’ll want to make an alternative one. Instead of storing the  ID / private key, we just want a map from legacy certificate ID (a UUID) to private key like so:

`SOAP_PRIVATE_KEYS="{\"my-uuid\": \"-----BEGIN PRIVATE KEY ...\"}"`

We should call this logic after we have fetched the certificate, and then we can try to grab the private key from the environment variable that will have been parsed into a dict.

If no private key is found in the configuration, we should error.

### Acceptance criteria

- [ ] New environment variable added to environment variables
- [ ] Use the map from legacy certificate ID to a private key
- [ ] Logic is called after we have fetched certificate
- [ ] Throw error if no private key is found
- [ ] Need directions on setting up the soap-api.env file with this env var. Does not need to be fully automated, fine to be instructions to fetch info from our 1password.

PR #7314, opened 2025-12-02
[Issue #6816] Modify soap proxy endpoints to use new env var

jakobpederson:
## Summary

Fixes / Work for #6816 

## Changes proposed

We already have a environment variable called SOAP_AUTH_CONTENT that we’ll want to make an alternative one. Instead of storing the ID / private key, we just want a map from legacy certificate ID (a UUID) to private key like so:

SOAP_PRIVATE_KEYS="{\"my-uuid\": \"-----BEGIN PRIVATE KEY ...\"}"

## Context for reviewers

Now that we have a legacy_certificate table, we should be able to associate a private key with a legacy_certificate_id instead of using a fingerprint.

### DIFF
--- a/api/src/legacy_soap_api/legacy_soap_api_auth.py
+++ b/api/src/legacy_soap_api/legacy_soap_api_auth.py
@@ -6,7 +6,7 @@ from urllib.parse import unquote
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import hashes
 from cryptography.x509 import load_pem_x509_certificate
-from pydantic import BaseModel
+from pydantic import BaseModel, ConfigDict
 from requests.adapters import HTTPAdapter
 from sqlalchemy import select
 
@@ -34,7 +34,9 @@ class SOAPClientCertificate(BaseModel):
     cert: str
     serial_number: int
     fingerprint: str
-    legacy_certificate_id: str
+    legacy_certificate: LegacyCertificate | None = None
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
 
     def get_pem(self, key_map: dict) -> str:
         """Note that this auth mechanism will only be configured in lower environments
@@ -42,9 +44,12 @@ class SOAPClientCertificate(BaseModel):
         There will be no prod configurations for this auth mechanism.
         TODO - is the above true? I think this happens for prod as well?
         """
-
+        if not self.legacy_certificate:
+            raise SOAPClientCertificateLookupError(
+                "could not retrieve legacy cert for serial number"
+            ) from None
         try:
-            value = key_map[self.legacy_certificate_id]
+            value = key_map[str(self.legacy_certificate.legacy_certificate_id)]
             return f"{value}\n\n{self.cert}"
         except KeyError:
             raise SOAPClientCertificateNotConfigured("cert is not configured") from None
@@ -106,28 +111,25 @@ def get_soap_client_certificate(
     legacy_certificate = db_session.execute(
         select(LegacyCertificate).where(LegacyCertificate.serial_number == str(cert.serial_number))
     ).scalar_one_or_none()
-    if not legacy_certificate:
-        raise SOAPClientCertificateLookupError(
-            "could not retrieve legacy cert for serial number"
-        ) from None
-    add_extra_data_to_current_request_logs(
-        {
-            "legacy_certificate_id": legacy_certificate.legacy_certificate_id,
-        }
-    )
-    if legacy_certificate.agency:
+    if legacy_certificate:
         add_extra_data_to_current_request_logs(
             {
-                "agency_code": legacy_certificate.agency.agency_code,
+                "legacy_certificate_id": legacy_certificate.legacy_certificate_id,
             }
         )
+        if legacy_certificate.agency:
+            add_extra_data_to_current_request_logs(
+                {
+                    "agency_code": legacy_certificate.agency.agency_code,
+                }
+            )
 
     return SOAPClientCertificate(
         cert=cert_str,
         fingerprint=cert.fingerprint(hashes.SHA256()).hex(),
         issuer=cert.issuer.rfc4514_string(),
         serial_number=cert.serial_number,
-        legacy_certificate_id=legacy_certificate.legacy_certificate_id,
+        legacy_certificate=legacy_certificate,
     )
 
 
@@ -140,10 +142,7 @@ def validate_certificate(
         )
         raise SOAPClientCertificateLookupError("no soap auth")
 
-    serial_number_str = str(soap_auth.certificate.serial_number)
-    legacy_certificate = db_session.execute(
-        select(LegacyCertificate).where(LegacyCertificate.serial_number == serial_number_str)
-    ).scalar_one_or_none()
+    legacy_certificate = soap_auth.certificate.legacy_certificate
 
     if not legacy_certificate:
         logger.warning(
--- a/api/tests/lib/data_factories.py
+++ b/api/tests/lib/data_factories.py
@@ -7,7 +7,7 @@ with only a few alterations.
 from src.constants.lookup_constants import Privilege
 from src.db.models.agency_models import Agency
 from src.db.models.competition_models import ApplicationForm
-from src.db.models.user_models import LegacyCertificate, Role, User
+from src.db.models.user_models import Role, User
 from src.legacy_soap_api.legacy_soap_api_auth import SOAPClientCertificate
 from tests.src.db.models.factories import (
     AgencyFactory,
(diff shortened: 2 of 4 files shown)


api/src/legacy_soap_api/legacy_soap_api_auth.py
```diff
@@ -34,17 +34,26 @@ class SOAPClientCertificate(BaseModel):
...
         """Note that this auth mechanism will only be configured in lower environments
 
         There will be no prod configurations for this auth mechanism.
         TODO - is the above true? I think this happens for prod as well?
         """
+
+        legacy_certificate_id = db_session.execute(
+            select(LegacyCertificate.legacy_certificate_id).where(
+                LegacyCertificate.serial_number == str(self.serial_number)
+            )
```
chouinar: We fetch the certificate elsewhere already, we shouldn't fetch it twice, is there a way we can fetch it once and pass it around? Perhaps attach it to this SOAPClientCertificate class? Might be tricky to change to work that way from the look of things because the place where we create the SoapClientCertificate is pretty nested in non-db code, but would simplify quite a bit if we could.

We wouldn't want to set it in this function (generally bad practice to have side effects of calling a function on this type of class, especially a function called `get_pem`). Might just need to adjust the logic all the way up to get_soap_auth?

api/src/legacy_soap_api/legacy_soap_api_proxy.py
```diff
@@ -80,7 +83,6 @@ def get_proxy_response(soap_request: SOAPRequest, timeout: int = PROXY_TIMEOUT)
         temp_cert_file.write(cert)
         temp_cert_file.flush()
 
-        add_extra_data_to_current_request_logs({"cert_id": cert_id})
```
chouinar, line 83: Adding the cert ID was something we really wanted, I don't think we need it necessarily in the same way, but if we aren't already doing this, can we add the `legacy_certificate_id` (the primary key of our cert table) to the request logs? If we adjust where we fetch the legacy cert from my other comment, let's assume we add something like:

```py
legacy_certificate = db_session.execute(select(...))..

add_extra_data_to_current_request_logs({"legacy_certificate_id": legacy_certificate.legacy_certificate_id, "agency_code": legacy_certificate.agency.agency_code}) # you'd need to add a quick relationship to agency for this, but I think logging that would be useful.

```

api/src/legacy_soap_api/legacy_soap_api_auth.py
jakobpederson: I can put it in the `get_soap_client_certificate` which is called in the `get_soap_auth` and put the legacy_certificate_id on the SoapClientCertificate

api/src/legacy_soap_api/legacy_soap_api_auth.py
```diff
@@ -33,18 +34,18 @@ class SOAPClientCertificate(BaseModel):
     cert: str
     serial_number: int
     fingerprint: str
+    legacy_certificate_id: str
```
chouinar: Make this the `LegacyCertificate` itself.

Then in `validate_certificate` remove the DB logic and pull the cert out of this object.

api/src/legacy_soap_api/legacy_soap_api_routes.py
```diff
@@ -58,7 +58,7 @@ def simpler_soap_api_route(
             full_path=request.full_path,
             headers=dict(request.headers),
             data=request.data,
-            auth=get_soap_auth(request.headers.get(MTLS_CERT_HEADER_KEY)),
+            auth=get_soap_auth(request.headers.get(MTLS_CERT_HEADER_KEY), db_session=db_session),
```
chouinar, line 61: I think we need to be careful here - this is the logic for calling the legacy system, with these changes, if a cert was passed into our system and not in our DB, we'd error.

Since we only will have certs we manually setup, that could be an issue. If we haven't even implemented our version of an operation, we definitely shouldn't error, the request should proxy through uneventfully. It should only hit any issues going to our half of the implementation.

I think the way we'd want to handle this is:
* Allow the cert to be null in the soap auth object - nothing on the proxy to grants.gov uses that value besides the get_pem function
* The get_pem function will fail (might want a `if cert_id is null` check), that's fine - that'll hopefully go away in the future so we don't need this env var config at all.
* If it is null when we go to our implementation of the route, then we error, you can't get anything from our system until the cert is setup.

api/src/legacy_soap_api/legacy_soap_api_auth.py
```diff
@@ -33,18 +34,23 @@ class SOAPClientCertificate(BaseModel):
     cert: str
     serial_number: int
     fingerprint: str
+    legacy_certificate: LegacyCertificate | None = None
 
-    def get_pem(self, key_map: dict) -> tuple[str, str]:
+    model_config = ConfigDict(arbitrary_types_allowed=True)
```
chouinar, line 39: Ugh, pydantic is probably more in the way here than helpful, we aren't using it for validation, just as a container. Could we just make this a dataclass instead to avoid needing the config?

```python
@dataclass
class SOAPClientCertificate:
    cert: str
    serial_number: int
    fingerprint: str
    legacy_certificate: LegacyCertificate | None = None

   def get_pem(...)...
```

jakobpederson: I tried using dataclass but it didn't fix the pydantic issue. I could swap in Any if type checking isn't a priority.

chouinar: If you made it a dataclass, what pydantic issue did you hit? It wouldn't be a pydantic object anymore, did you remove the `BaseModel`

jakobpederson: I swapped in this:
```
@dataclass
class SOAPClientCertificate:
    cert: str
    serial_number: int
    fingerprint: str
    legacy_certificate: LegacyCertificate | None = None
```
And this is what I got:
```
# /home/jakobpederson/.cache/pypoetry/virtualenvs/simpler-grants-gov-api-cCrPfAjT-py3.13/lib/python3.13/site-packages/pydantic/_internal/_generate_schema.py:659: in _unknown_type_schema
#     raise PydanticSchemaGenerationError(
# E   pydantic.errors.PydanticSchemaGenerationError: Unable to generate pydantic-core schema for <class 'src.legacy_soap_api.legacy_soap_api_auth.SOAPAuth'>. Set `arbitrary_types_allowed=True` in the model_config to ignore this error or implement `__get_pydantic_core_schema__` on your type to fully support it.
# E
# E   If you got this error by calling handler(<some type>) within `__get_pydantic_core_schema__` then you likely need to call `handler.generate_schema(<some type>)` since we do not call `__get_pydantic_core_schema__` on `<some type>` otherwise to avoid infinite recursion.
# E
# E   For further information visit https://errors.pydantic.dev/2.12/u/schema-for-unknown-type
```

jakobpederson: I wonder if it's because this certificate obj is nested in another pydantic object that is SOAPAuth.

jakobpederson: Or perhaps SoapRequest given some more of the error message:
```
src/legacy_soap_api/legacy_soap_api_schemas.py:37: in <module>
    class SOAPRequest(BaseModel):
/home/jakobpederson/.cache/pypoetry/virtualenvs/simpler-grants-gov-api-cCrPfAjT-py3.13/lib/python3.13/site-packages/pydantic/_internal/_model_construction.py:255: in __new__
    complete_model_class(
/home/jakobpederson/.cache/pypoetry/virtualenvs/simpler-grants-gov-api-cCrPfAjT-py3.13/lib/python3.13/site-packages/pydantic/_internal/_model_construction.py:648: in complete_model_class
    schema = gen_schema.generate_schema(cls)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

chouinar: Ah, we could make `SOAPAuth` a dataclass, and SOAPRequest as well. We can either leave it as-is, or adjust all of the classes. I'll defer to you.

Really, unless you want validation beyond type-checking, we shouldn't be using Pydantic, dataclasses are much better for any sort of simple container object, really we're just using these to avoid needing to manually define an `__init__` function (that's all dataclasses do under the hood is generate it for you).

chouinar (approved):
Something I forgot - right now the SOAP_AUTH_CONTENT env var gets setup by running a script https://github.com/HHS/simpler-grants-gov/blob/main/api/bin/setup-soap.sh

We don't need to go through that process necessarily again, especially since this is a temporary set of changes that hopefully we'll remove soon. But we should:
* Cleanup that script (and reference from the Makefile)
* Maybe remove soap-api.env as well from the docker-compose?
* Say how to setup this new env var (it can be as simple as "copy X from 1password")

This is fine to be a follow-up ticket.

api/src/legacy_soap_api/legacy_soap_api_auth.py
jakobpederson: I'll leave it as-is if that's acceptable for now. I'd be down for a refactor later.

jakobpederson:
> Something I forgot - right now the SOAP_AUTH_CONTENT env var gets setup by running a script https://github.com/HHS/simpler-grants-gov/blob/main/api/bin/setup-soap.sh
> 
> We don't need to go through that process necessarily again, especially since this is a temporary set of changes that hopefully we'll remove soon. But we should:
> 
> * Cleanup that script (and reference from the Makefile)
> * Maybe remove soap-api.env as well from the docker-compose?
> * Say how to setup this new env var (it can be as simple as "copy X from 1password")
> 
> This is fine to be a follow-up ticket.

I'll create a ticket.

PR #7315, opened 2025-12-02
[Unticketed] Fix typo in setup-cert-user task

jakobpederson:
## Changes proposed

Fix typo in setup-cert-user task 

## Context for reviewers

The setup-cert-user task had a typo in the variable names

### DIFF
--- a/api/Makefile
+++ b/api/Makefile
@@ -384,9 +384,6 @@ setup-soap-proxy:
 validate-simpler-endpoints:
 	$(PY_RUN_CMD) validate-simpler-endpoints
 
-setup-cert-user:
-	$(PY_RUN_CMD) flask task setup-cert-user
-
 ####################
 # XML Generation CLI
 ####################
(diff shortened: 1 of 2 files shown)


api/Makefile
```diff
@@ -384,6 +384,9 @@ setup-soap-proxy:
 validate-simpler-endpoints:
 	$(PY_RUN_CMD) validate-simpler-endpoints
 
+setup-cert-user:
+	$(PY_RUN_CMD) flask task setup-cert-user
```
chouinar: How do you pass in the parameters to this command?

I'd actually say we should not add this to avoid bloating the makefile / the primary way this will be run is against an environment non-locally which is an entirely different command, and those aren't in the makefile either.

For running locally, `make cmd args="task setup-cert-user --tcertificates-id X --role-ids X"` would do this.

jakobpederson: Oh, I forgot to add the args. I'm ok with pulling it out the makefile, I just thought I needed to add it but that command you have works fine.

chouinar (approved):

PR #7326, opened 2025-12-03
[ISSUE #6938] Supporting pagination for list organization user endpoint

### LINKED ISSUE #6938: BE - Supporting pagination for organization roster endpoint
### Summary

Capturing the effort to introduce the BE capabilities to enable pagination in the organization roster page 

[API Pagination Implementation Guide](https://navasage.atlassian.net/wiki/x/OYDxhg) 

### Acceptance criteria

- [ ] BE supports pagination for the organization roster table in the organization page 
- [ ] Ensure test coverage for pagination 
- [ ] 10 row count for the table 
- [ ] Sorting by default - email asc (A-Z)

joshtonava:
## Summary

Fixes #6938 

## Changes proposed

- Did some renaming to add more consistency
- Added pagination support to the `POST /v1/organizations/:organization_id/users` endpoint
- Implemented sorting by `email`, `first_name`, `last_name`, and `created_at` with default sort of email ascending
- Set default page size to 10 rows per page
- Configured NULLS LAST handling for null-safe sorting of name and email fields
- Refactored serialization to use a dataclass

## Context for reviewers

- Added explicit eager loading for `User.profile` to prevent lazy loading errors when serialization happens outside the database session context.
- The query now includes explicit joins to `User`, `LinkExternalUser`, and `UserProfile` tables to enable sorting by email and name fields. These are `outerjoin` since in theory users may not have profiles or external user records.

## Validation steps

1. **Basic Pagination** - Request first page with 10 users:
  ```bash
  curl -X POST https://your-api/v1/organizations/{org_id}/users \
    -H "X-SGG-Token: YOUR_TOKEN" \
    -H "Content-Type: application/json" \
    -d '{
      "pagination": {
        "page_offset": 1,
        "page_size": 10,
        "sort_order": [{"order_by": "email", "sort_direction": "ascending"}]
      }
    }'
```

2. Run tests:
```bash
  make test args="tests/src/api/organizations_v1/test_list_organization_users.py"
```

### DIFF
--- a/api/src/api/organizations_v1/organization_schemas.py
+++ b/api/src/api/organizations_v1/organization_schemas.py
@@ -1,13 +1,9 @@
-from dataclasses import dataclass
-from typing import TypedDict
-from uuid import UUID
-
 from src.api.schemas.extension import Schema, fields
 from src.api.schemas.extension.field_validators import Email, Length, validators
 from src.api.schemas.response_schema import AbstractResponseSchema, PaginationMixinSchema
 from src.api.schemas.search_schema import StrSearchSchemaBuilder
 from src.api.schemas.shared_schema import RoleSchema
-from src.constants.lookup_constants import LegacyUserStatus, OrganizationInvitationStatus, Privilege
+from src.constants.lookup_constants import LegacyUserStatus, OrganizationInvitationStatus
 from src.pagination.pagination_schema import generate_pagination_schema
 
 
@@ -57,31 +53,6 @@ class OrganizationDataSchema(Schema):
         allow_none=True,
         metadata={"description": "SAM.gov entity information"},
     )
-
-
-class RoleDict(TypedDict):
-    """Type definition for role dictionary in EnrichedOrganizationUser."""
-
-    role_id: UUID
-    role_name: str
-    privileges: set[Privilege]
-
-
-@dataclass
-class EnrichedOrganizationUser:
-    """OrganizationUser enriched with computed fields for serialization.
-
-    This dataclass provides a simple data container that Marshmallow can
-    serialize automatically. All transformation logic is handled in the
-    service layer when constructing instances.
-    """
-
-    user_id: UUID
-    email: str | None
-    roles: list[RoleDict]
-    first_name: str | None
-    last_name: str | None
-    is_ebiz_poc: bool
 
 
 class OrganizationUserSchema(Schema):
--- a/api/src/services/organizations_v1/list_organization_users.py
+++ b/api/src/services/organizations_v1/list_organization_users.py
@@ -1,4 +1,5 @@
 from collections.abc import Sequence
+from dataclasses import dataclass
 from uuid import UUID
 
 from pydantic import BaseModel
@@ -7,8 +8,7 @@ from sqlalchemy.orm import selectinload
 from sqlalchemy.sql.selectable import Select
 
 import src.adapters.db as db
-from src.api.organizations_v1.organization_schemas import EnrichedOrganizationUser
-from src.constants.lookup_constants import ExternalUserType
+from src.constants.lookup_constants import ExternalUserType, Privilege
 from src.db.models.entity_models import Organization
 from src.db.models.user_models import (
     LinkExternalUser,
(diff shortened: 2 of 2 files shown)


api/tests/src/api/organizations_v1/test_list_organization_roles.py
```diff
@@ -72,6 +72,6 @@ def test_list_organization_roles_404_organization(client, db_session, enable_fac
 def test_list_organization_roles_401_no_token(client):
     """Test that accessing organization roles without auth token returns 401"""
     # Make request
-    resp = client.post(f"/v1/organizations/{uuid4()}/users")
+    resp = client.post(f"/v1/organizations/{uuid4()}/roles/list")
```
joshtonava, line 75: This was a bug in the test

api/src/api/organizations_v1/organization_schemas.py
```diff
@@ -73,18 +87,43 @@ class OrganizationMemberSchema(Schema):
...
             "description": "Whether the user is the EBiz POC (Electronic Business Point of Contact) for the organization in SAM.gov",
             "example": False,
-        }
+        },
     )
 
+    @pre_dump
+    def flatten_organization_user(
+        self, enriched_user: EnrichedOrganizationUser, **kwargs: Any
+    ) -> dict:
```
chouinar: I am curious why we moved the structuring of the dict here? If we already need to transform the shape of the DB objects we fetch, why do DB object -> enriched object -> response object?

Seems simplest to just get it in the right shape when iterating over the db response, otherwise we have more places where something could get transformed incorrectly/lost. Assume we have to add a new field in the future, we'd have to remember to update multiple different parts of the code.

---

A middleground that might be best - when Marshmallow converts an object to JSON, it can parse dicts, or any python object, it's just recursively fetching things with a bunch of `getattr` calls. If you make the `EnrichedOrganizationUser` have the same shape as this schema - then we don't need to do anything here, it'll parse it just fine.

api/src/services/organizations_v1/list_organization_users.py
```diff
@@ -0,0 +1,177 @@
...
+        if column is None:
+            # This shouldn't happen due to schema validation, but handle gracefully
+            continue
+
+        if order.sort_direction == SortDirection.ASCENDING:
+            order_cols.append(asc(column).nulls_last())
+        elif order.sort_direction == SortDirection.DESCENDING:
+            order_cols.append(desc(column).nulls_last())
+
+    return stmt.order_by(*order_cols)
```
chouinar: Since this part is mostly our existing `apply_sorting` but handles the sort being on different tables (which is something generally useful), could we move this to a utility that takes in a mapping?

I would also caution against the `if column is None -> continue` - I think erroring would be better as it means we have a configurational problem we want to know about. If we want to split the difference, continue, but log an error message to alert us.

api/src/api/organizations_v1/organization_schemas.py
joshtonava: I was trying to do the least amount of special transformation as possible in the service. Perhaps that was misguided.

But to your point, having things in multiple places (adding the is_ebiz_poc, transforming to response format) could potentially add more places for introducing a mistake.

Im going to rework this and get back to you

api/src/services/organizations_v1/list_organization_users.py
joshtonava: I like the suggestion of moving this to a utility. I will give it a go.

And think erroring would imply that the user supplied us with a sort order that we didnt provide a mapping for, which should be preventable since schema validation ensures its one of ["email", "first_name", "last_name", "created_date"]. So Im fine with erroring because that means we really have something we need to address. It would be better to find the issue sooner then have the endpoint be defective.

api/src/api/organizations_v1/organization_schemas.py
```diff
@@ -55,14 +59,39 @@ class OrganizationDataSchema(Schema):
...
+    serialize automatically. All transformation logic is handled in the
+    service layer when constructing instances.
+    """
+
+    user_id: UUID
+    email: str | None
+    roles: list[RoleDict]
+    first_name: str | None
+    last_name: str | None
+    is_ebiz_poc: bool
```
chouinar: Anything stopping these from both being dataclasses just for consistency?

Also, since the schema doesn't need to know about this type anymore, could move it over to the service class where it's used.

---

FYI for the future - if we upgrade our APIFlask version, we'll be able to define schemas with Pydantic which could make all of this a lot smoother as we'd just use that.

joshtonava: done and done

PR #7328, opened 2025-12-03
[unticketed] add basic unit testing docs

doug-s-nava:
## Summary

unticketed

## Changes proposed

Adds some basic unit testing docs

## Validation steps

1. review additions to docs
2. _VERIFY_: all additions are correct and make sense

### DIFF
--- a/documentation/frontend/testing.md
+++ b/documentation/frontend/testing.md
@@ -22,7 +22,7 @@ Using this function, tests should work automatically in CI, but they will requir
 - We strive for high unit test coverage (but not 100%)
 - We expect engineers to write unit tests for any changes they make in the same PR that contains the code changes
 - We use data fixtures when relevant (see https://github.com/HHS/simpler-grants-gov/blob/main/frontend/src/utils/testing/fixtures.ts)
-- We strive to include snapshot tests and axe tests on all components
+- We strive to include axe tests on all components
 
 ### Helpful common patterns
 


documentation/frontend/testing.md
```diff
@@ -15,3 +15,52 @@ Using this function, tests should work automatically in CI, but they will requir
...
 - copy the token variable declaration from the e2e_token.tmp file into your frontend .env.local file
 - that's it! Running e2e tests using the functionality mentioned above should now work locally
+
+## Unit testing
+
+- We use Jest and testing-library for our unit testing
+- We strive for high unit test coverage (but not 100%)
+- We expect engineers to write unit tests for any changes they make in the same PR that contains the code changes
+- We use data fixtures when relevant (see https://github.com/HHS/simpler-grants-gov/blob/main/frontend/src/utils/testing/fixtures.ts)
+- We strive to include snapshot tests and axe tests on all components
```
ErinPattisonNava: I personally believe we should be moving away from snapshot tests; they provide a very weak testing guarantee, require deep knowledge of the component to interpret, and depend on inspection to determine if any change is acceptable.  I'd really like us to have explicit tests, where we point out what is important and and implement the check in a way that ignores underlying page implementation details.

Could we remove this line to discuss more, but get the rest of the doc in so we can start building on top of it?

PR #7334, opened 2025-12-04
[Unticketed] Fix bug in jsonref resolver which incorrectly resolves schemas

chouinar:
## Changes proposed
* Ports a bugfix from the jsonref library in to fix an issue with resolving schemas incorrectly.

## Context for reviewers
Added context in the ticket - but the issue this was causing was that depending on the order we resolved schemas, it would sometimes incorrectly resolve a schema to have the wrong fields. The linked fix changes some internal method for how it tracks references to fix it, but it hasn't been made into a release yet, so manually patched/overrode the function to the newer fixed one.

## Validation steps
Before, running the tests in the forms folder could fail ( `make test args="-x -s -vv tests/src/form_schema/forms"`).

### DIFF
--- a/api/src/form_schema/jsonschema_resolver.py
+++ b/api/src/form_schema/jsonschema_resolver.py
@@ -1,8 +1,53 @@
+from collections.abc import Callable, Mapping, Sequence
 from typing import Any
 
 import jsonref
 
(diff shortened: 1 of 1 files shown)


jakobpederson (approved):
Tests are now passing for me!

PR #7219, opened 2025-11-24
[Issue #7163]: Remove environment variables and use cli args instead for step functions  

prasnava:
## Summary

Updated the step functions to  use cli args instead of environment vars. 

## Validation steps

The step functions should continue to work.

Issue #7163, opened 2025-11-19
Remove environment variables in favor of command line args in step functions


prasnava:
Update the aws step functions to use command line args instead of relying on environment variables.

PR #7338, opened 2025-12-04
Revert "[Issue #7163]: Remove environment variables and use cli args instead for step functions  "

prasnava:
Reverts HHS/simpler-grants-gov#7219

The analytics meta base load is breaking.


prasnava:
As a part of this commit revert, the extra clean up we added was removing the unused env var scheduled_job_name.

chouinar (approved):

PR #7342, opened 2025-12-04
Added notes for Front End Windows Documentation

dghazvini:
## Summary

Added Windows front end documentation notes based on @dghazvini @Dao-REI @chay-REIsys setup (No associated ticket).

## Changes proposed

Updates to development.md to include Windows setup steps for Front End.

## Context for reviewers

Based on the setup experience for us REI Systems folks currently using Windows systems.


ErinPattisonNava (commented):
Thanks for writing documentation!  Can you refactor this so there isn't so much repetition/it is easier to update long term?  Maybe a section on setup at the top for windows vs mac, then in each section just list out the two different commands so we don't have multiple places with the node version or the authentication setup.

dghazvini:
@ErinPattisonNava Thanks for the feedback; definitely can be condensed. I've pushed an update to keep it to one section and just mention the alternative steps for Windows users where needed.

Issue #7247, opened 2025-11-26
Adding table data to manage legacy users page


ErinPattisonNava:
### Summary

Capturing the effort to put together a page to show a list of users that belonged to that organization in the grants.gov side

Figma: https://www.figma.com/design/zPlDWcYM396szw2yL3UXvk/Simpler-Grants-2025?node-id=7993-90708&t=kgz5u9g8av1W0aSh-1

This adds the calls for data and creates the first two columns in the table, the interaction component will be handled in a later ticket.

Note that the table can never be empty; in order create an organization, there must be at least one member (the EBIZ POC) listed.

### Acceptance criteria

- [ ] All changes below are made to `/manage-users/legacy` (note page is behind manageUsersOff feature flag)
- [ ] Page is only accessible if the user has permission of organization admin, otherwise show don't-have-access error 
- [ ] Create a table with users pulled from 
   - [ ] Data for table comes from endpoint created [here](https://github.com/HHS/simpler-grants-gov/issues/6952)
   - [ ] Table is sorted based on email address (check with @joshtonava that it is built, otherwise work with him to add functionality that will automagically work when the backend is complete or spin up follow up frontend ticket) 
   - [ ] No pagination
   - [ ] Table contains 2 columns 
      - [ ] Email column contains email from the user
      - [ ] Name column contains a concatenated value of First and Last name in the same row 
         - [ ] If name not available, leave blank
      - [ ] Third column of actions will be covered in a follow up [ticket](https://github.com/HHS/simpler-grants-gov/issues/7220)  
- [ ] If API returns 500 show something-wrong error
- [ ] Full unittests
- [ ] Do a Design review with @chelsea-nava before submitting the PR

ErinPattisonNava:
@eric-serrato95 Ready for UAT!

ErinPattisonNava:
@eric-serrato95 is on vacation, but I know he did the UAT before he left so moving this ticket to done to unblock later tickets

Issue #6942, opened 2025-11-04
Building page to support showing a list of legacy users from grants.gov


eric-serrato95:
### Summary

Capturing the effort to put together a page to show a list of users that belonged to that organization in the grants.gov side

Figma: https://www.figma.com/design/zPlDWcYM396szw2yL3UXvk/Simpler-Grants-2025?node-id=7993-90708&t=kgz5u9g8av1W0aSh-1

This builds out the page skeleton, and the table will be filled out in the next [ticket](https://github.com/HHS/simpler-grants-gov/issues/7247).  

Note that the table can never be empty; in order create an organization, there must be at least one member (the EBIZ POC) listed.

### Acceptance criteria

- [ ] Build a new page at `/manage-users/legacy` (see `/managed-users` for similar constraints)
   - [ ] Page is only accessible if you are authenticated
   - [ ] Page is only accessible if the feature flag manageUsersOff is disabled, otherwise redirects to maintenance page
- [ ] Add button on `manage-users` page to point to this new page.
- [ ] Page matches mocks on 
   - [ ] H1 has page heading
      - [ ]  H1 header spans organization name and page title
   - [ ] Breadcrumbs
   - [ ] Metadata is set
   - [ ] Content at top, uses the empty use case for text
- [ ] Full unittests
- [ ] Do a Design review with @chelsea-nava before submitting the PR

eric-serrato95:
@ErinPattisonNava UAT Complete, moving to Done

PR #7346, opened 2025-12-04
[Issue #6942] adding the skeleton page for manage users legacy invite

ErinPattisonNava:
## Summary

Work for #6942 

## Changes proposed

Building out a skeleton for the Invite legacy page that shows the organization information and the empty use case, plus the error use case if we can't load the data.

## Context for reviewers

This is the first ticket out of 3 for building out this page, the authorized data/table data will be added in the [next ticket](https://github.com/HHS/simpler-grants-gov/issues/7247).  

A word on unittests: We can't currently unittest pages that are behind feature flags.   Working with @doug-s-nava in a different PR to try to sort that.  So in the short term I don't have unittests for that, I'll include them in the follow on PR with the table.

I also wanted to add a unittest for adding the button to the manage users page, but the unittests in ManageUsersPageContent.test.tsx only check calls to mocks not the rendered output, and apparently there is a problem there which renders and empty div.  I didn't have time to debug that, so I couldn't add a test. 

I also slipped in switching the view details button and the manage users button on the organizations page, that's so it looks nicer when a user is sometimes a member and sometimes an admin the UI will stay stable. 

## Validation steps

- [ ] Make sure that the `manageUsersOff` feature flag is disabled
- [ ] Log on as a user who is an admin of an organization (e.g. one_org_user on localhost)
- [ ] Go through the Header: Workspace -> Organizations
- [ ] Click on "Manage users" for an org
- [ ] Click on the button saying "Invite Grants.gov users"
- [ ] Note that the breadcrumbs  are visible and the links work as expected
- [ ] Note the org title is listed in the header
- [ ] See the message implying that the request for data is empty and that the link works correctly
- [ ] Save the URL, log out, and then attempt to go to the URL again.  It should take you to the page notifiying you to log in
- [ ] To see the error case, open `organization_routes.py`, include the import `from src.api.route_utils import raise_flask_error` at the top, and add `raise_flask_error(500, f"temp error!!")` at line 60
   - [ ] Note the page displays with limited data (no org name, change in breadcrumbs) and a polite error message 
- [ ] Check the organizations page that the manage users button and view detail button are in that order (the manage users button will not be present if they are only the member of an org)

### DIFF
--- a/frontend/src/app/[locale]/(base)/organization/[id]/manage-users/legacy/page.tsx
+++ b/frontend/src/app/[locale]/(base)/organization/[id]/manage-users/legacy/page.tsx
@@ -1,4 +1,3 @@
-import { UnauthorizedError } from "src/errors";
 import withFeatureFlag from "src/services/featureFlags/withFeatureFlag";
 import { getOrganizationDetails } from "src/services/fetch/fetchers/organizationsFetcher";
 
@@ -8,11 +7,10 @@ import { redirect } from "next/navigation";
 import { GridContainer } from "@trussworks/react-uswds";
 
 import Breadcrumbs from "src/components/Breadcrumbs";
-import { AuthenticationGate } from "src/components/user/AuthenticationGate";
 
-interface InviteLegacyUsersPageProps {
+type InviteLegacyUsersPageProps = {
   params: Promise<{ locale: string; id: string }>;
-}
+};
 
 const InviteLegacyUsersErrorPage = ({
   organizationId,
@@ -57,9 +55,6 @@ async function InviteLegacyUsersPage({ params }: InviteLegacyUsersPageProps) {
   try {
     organizationDetails = await getOrganizationDetails(organizationId);
   } catch (error) {
-    if (error instanceof UnauthorizedError) {
-      throw error;
-    }
     return <InviteLegacyUsersErrorPage organizationId={organizationId} />;
   }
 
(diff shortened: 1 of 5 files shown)


frontend/src/app/[locale]/(base)/organization/[id]/manage-users/legacy/page.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+import { useTranslations } from "next-intl";
+import { getTranslations } from "next-intl/server";
+import { redirect } from "next/navigation";
+import { GridContainer } from "@trussworks/react-uswds";
+
+import Breadcrumbs from "src/components/Breadcrumbs";
+import { AuthenticationGate } from "src/components/user/AuthenticationGate";
+
+interface InviteLegacyUsersPageProps {
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/organization/[id]/manage-users/legacy/page.tsx]: not a big deal and we can revisit, but we documented a style decision on interfaces vs types a while back, and except in very special circumstances I think props can either be a declared type, or just generically defined in the function definition https://github.com/HHS/simpler-grants-gov/blob/main/documentation/frontend/code-style.md#interfaces-vs-types

frontend/src/app/[locale]/(base)/organization/[id]/manage-users/legacy/page.tsx
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  const organizationName =
+    organizationDetails?.sam_gov_entity?.legal_business_name;
+
+  const organizationLink = `/organization/${organizationId}`;
+  const manageUsersLink = `${organizationLink}/manage-users`;
+
+  return (
+    <AuthenticationGate>
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/organization/[id]/manage-users/legacy/page.tsx]: can we get rid of this now that it's in the layout?

frontend/src/app/[locale]/(base)/organization/[id]/manage-users/legacy/page.tsx
```diff
@@ -0,0 +1,114 @@
...
+async function InviteLegacyUsersPage({ params }: InviteLegacyUsersPageProps) {
+  const { locale } = await params;
+  const t = await getTranslations({ locale });
+  const { id: organizationId } = await params;
+
+  let organizationDetails;
+  try {
+    organizationDetails = await getOrganizationDetails(organizationId);
+  } catch (error) {
+    if (error instanceof UnauthorizedError) {
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/organization/[id]/manage-users/legacy/page.tsx]: do we need to handle this separately? My hope is that the authentication gate would handle this

frontend/src/components/manageUsers/ManageUsersPageContent.tsx
```diff
@@ -40,26 +44,47 @@ export async function ManageUsersPageContent({
     console.error("Unable to fetch organization information", error);
   }
   const name = userOrganizations?.sam_gov_entity?.legal_business_name;
+
+  const InviteLegacyUsersButton = ({
```
doug-s-nava: can this be split out from the parent component?

frontend/tests/components/organization/manage-users/ManageUsersPageContent.test.tsx
```diff
@@ -8,14 +8,16 @@ import type { AuthorizedData } from "src/types/authTypes";
 
 import { ManageUsersPageContent } from "src/components/manageUsers/ManageUsersPageContent";
 
-type TranslationFn = (key: string) => string;
+type TranslationFunction = (key: string, values?: { name?: string }) => string;
 
-const getTranslationsMock = jest.fn<Promise<TranslationFn>, [string]>(
-  (_namespace: string) => Promise.resolve((key: string) => key),
+const useTranslationsMock = jest.fn<TranslationFunction, [string]>(
```
doug-s-nava: is there a reason https://github.com/HHS/simpler-grants-gov/blob/113828a0bc73a5a219c7387d8fd88d4ca9abd7bc/frontend/src/utils/testing/intlMocks.ts#L3 doesn't work here?

ErinPattisonNava:
> a couple relatively minor requests, and a few thoughts and things I uncovered:
> 
> * we should probably be suspending children of authorization gate in order to add loading states. This is gonna be tricky! Something to think about down the line when we have time to think about things like loading states

I can write a ticket up, but is it authorization gate or authentication gate?  Or both?

> * there is a bug on the manage users page related to the usage of `useTranslation` vs `getTranslation`. It may not show up in prod, but throws an error locally if you use `useTranslation` in an async server component.

Changed.

> * related to our previous discussion, not sure if it will actually help anything or not, but using `<Link>` for the internal links in the new page may make things load slightly quicker. If this is something we want to dig into more later, we can look into building our own breadcrumb component that uses `<Link>` instead of `<a>` tags

I'm going to leave this one alone because I'm kinda mad at  Breadcrumbs right now.

Issue #5097, opened 2025-05-19
Create Activity History section


mxk0:
### Summary
This ticket is to add the Activity History section to the Application page.
- Reference figma design from #6965 
- Figma: https://www.figma.com/design/zPlDWcYM396szw2yL3UXvk/Simpler-Grants-2025?node-id=7423-44108&t=PrVAhuH4L0EivccL-4

Note that you can never have a scenario when there is no activity history, as the first event is "application_created".

### Acceptance criteria

- [ ] Activity History section created per design and added to application page below the Attachments section
   - [ ] use the `/applications/{application_id}/audit_history` endpoint
- [ ] Table is automatically sorted by latest "timestamp" and is displayed in ET (regardless of user preference)
- [ ] Ensure that that these use cases display correctly.  We should be filtering the request for events to these specific events:
  - [ ] `application_created`
  - [ ] `application_name_changed`
  - [ ] `application_submitted`
  - [ ] `application_submit_rejected`
  - [ ] `attachment_added`
  - [ ] `attachment_deleted`
  - [ ] `user_added` (can be tested by creating an application as an individual rather than an organization)
  - [ ] `form_updated`
- [ ] These events will be excluded from the list above because they aren't testable/shouldn't be shown to the user: `attachment_updated`, `submission_create`, `user_updated`, `user_removed`,  `organization_added` (will be added in #6644 )
- [ ] No pagination required, just display all items
- [ ] Add error scenario if the application history fails to load
- [ ] Full unittests
- [ ] UAT with Design
  - [ ] Table design matches Figma
  - [ ] All supported activity types (above) have correct design
  - [ ] Initial load state
  - [ ] Error state
- [ ] UAT with Product
  - [ ] Multiple user contributions appear as expected in the activity history
  - [ ] All supported activities (above) appear as expected
  - [ ] Initial load state
  - [ ] Error state

mxk0:
Topics for refinement:

- Prioritization (probably not for pilot)
- Is activity already tracked on the backend? Do we need to revisit scope before we start building the frontend (when we get to it)?

glenblosser-nava:
Work proceeding on this ticket. I've attached a WIP screenshot as an update.

<img width="1145" height="707" alt="Image" src="https://github.com/user-attachments/assets/cf3e6ce6-fe36-4755-83bd-bebb9ee92d0b" />

jesababon:
[...]

PR #7362, opened 2025-12-05
[Issue #5097] Create Activity History section on Applications page

glenblosser-nava:
## Summary

Work  for #5097  

## Changes proposed

Adds Application History at the bottom of the application view, according to [this](https://www.figma.com/design/zPlDWcYM396szw2yL3UXvk/Simpler-Grants-2025?node-id=442-4117&p=f) figma, except that pagination is not required to be supported yet.

## Validation steps

Create an application, and confirm that the history section is present, and updates as you make changes to the application

Screenshot 2025-12-05 at 3 06 58 PM

### DIFF
--- a/frontend/src/services/fetch/fetchers/applicationFetcher.ts
+++ b/frontend/src/services/fetch/fetchers/applicationFetcher.ts
@@ -96,6 +96,29 @@ export const getApplicationHistory = async (
         page_size: 5000,
         sort_order: [{ order_by: "created_at", sort_direction: "descending" }],
       },
+      // The following events exist, but should not be displayed yet (See #6644 for implementation):
+      // attachment_updated, submittion_created, user_updated, user_removed, organization_added
+      // events are left in the array, but commented out to make the follow-up ticket easier and
+      // make it clear that these events are being excluded deliberately for now
+      filters: {
+        application_audit_event: {
+          one_of: [
+            "application_created",
+            "application_name_changed",
+            "attachment_added",
+            "attachment_deleted",
+            // "attachment_updated",
+            "application_submitted",
+            "form_updated",
+            "user_added",
+            // "user_updated",
+            // "user_removed",
+            // "organization_added",
+            "application_submit_rejected",
+            // "submission_created",
+          ],
+        },
+      },
     },
   });
 


frontend/src/services/fetch/fetchers/applicationFetcher.ts
```diff
@@ -79,6 +80,28 @@ export const getApplicationDetails = async (
...
+  const ssgToken = {
+    "X-SGG-Token": token,
+  };
+  const response = await fetchApplicationWithMethod("POST")({
+    subPath: `${applicationId}/audit_history`,
+    additionalHeaders: ssgToken,
+    body: {
+      pagination: {
+        page_offset: 1,
+        page_size: 500,
```
ErinPattisonNava: Let's set this higher, I suspect this table is going to be chatty.  Maybe 5000?

frontend/src/components/application/ApplicationHistoryTable.tsx
```diff
@@ -0,0 +1,111 @@
+import { ApplicationHistory } from "src/types/applicationResponseTypes";
```
ErinPattisonNava, line 1: Unittests?

frontend/src/components/application/ApplicationHistoryTable.tsx
```diff
@@ -0,0 +1,111 @@
...
+}: {
+  applicationHistory: ApplicationHistoryCardProps;
+}) => {
+  const t = useTranslations("Application.historyTable");
+  const activityTranslations = useTranslations(
+    "Application.historyTable.activities",
+  );
+
+  return (
+    <Table className="width-full overflow-wrap simpler-application-forms-table">
```
ErinPattisonNava: Is there a reason we don't want to use our TableWithResponsiveHeader component?

frontend/src/services/fetch/fetchers/applicationFetcher.ts
```diff
@@ -79,6 +80,28 @@ export const getApplicationDetails = async (
...
+        sort_order: [{ order_by: "created_at", sort_direction: "descending" }],
+      },
+    },
+  });
+
+  return (await response.json()) as ApplicationHistoryApiResponse;
```
ErinPattisonNava, line 125: I was talking with @doug-s-nava about how we want to[ standardize around errors in fetchers](https://github.com/HHS/simpler-grants-gov/pull/7377#discussion_r2600358677), not sure if we want to incorporate anything here.

frontend/stories/components/application/history.mock.json
```diff
@@ -0,0 +1,10 @@
+[
+  {
+    "application_audit_event": "application_name_changed",
```
ErinPattisonNava, line 3: Do we want full coverage here of all the different types of events?

ErinPattisonNava:
In the Summary I believe this needs to be "Work" not "Fixes" so it can go through UAT.

ErinPattisonNava:
When testing I got an error in next on the console:
Screenshot 2025-12-09 at 5 49 36 PM

I also got an error adding an attachment, can you confirm that isn't part of your PR?

For the error case, It seems wrong to take down the entire apply page just because we can't show the history.  Can we have an error message just for the history section, and let the rest of the application function as normal?

frontend/src/components/application/ApplicationHistoryTable.tsx
glenblosser-nava: Just consistency with other pages in the Application view, such as ApplicationFormsTable. If we should be moving to TableWithResponsiveHeader, I can do so

frontend/src/components/application/ApplicationHistoryTable.tsx
glenblosser-nava: I don't follow what you mean here. Should the type have unit tests around it? I've added unit tests to the application history view itself.

frontend/src/services/fetch/fetchers/applicationFetcher.ts
glenblosser-nava: Setting it too high results in getting no response from the backend. Future work on this feature will include pagination, so 500 should be plenty for now.

frontend/stories/components/application/history.mock.json
glenblosser-nava: The key feature I am aiming to test is the logic around displaying the results at all. While I can add additional coverage for each localization, I felt it wasn't really needed. If you think it is, I can update it

glenblosser-nava:
> When testing I got an error in next on the console: Screenshot 2025-12-09 at 5 49 36 PM
> 
> I also got an error adding an attachment, can you confirm that isn't part of your PR?
> 
> For the error case, It seems wrong to take down the entire apply page just because we can't show the history. Can we have an error message just for the history section, and let the rest of the application function as normal?

I've fixed the data-testid error message. The attachment error exists in main, so is not part of this PR. I have also moved the error case to just display an error in the history section.

frontend/src/services/fetch/fetchers/applicationFetcher.ts
glenblosser-nava: I think that is something we should roll out across our fetchers, but I'd prefer it to be a cleanup ticket rather than doing it piecemeal

frontend/src/services/fetch/fetchers/applicationFetcher.ts
ErinPattisonNava: Pagination isn't going to happen for a while, but I'm ok leaving this at 500.  I am concerned why a larger number results in an error/empty return.  Can you follow up with the backend to understand why?

frontend/src/components/application/ApplicationHistoryTable.tsx
ErinPattisonNava: I would expect the unittests for this component to map to this component, so if we decide to use it somewhere else coverage is still clear. It also keeps test file size manageable.  Could you separate them out?

frontend/src/components/application/ApplicationHistoryTable.tsx
ErinPattisonNava: @doug-s-nava Thoughts?  I would say shift so we can use what will hopefully be standardized pagination data and patterns especially for lists.

frontend/stories/components/application/history.mock.json
ErinPattisonNava: I'd like better test coverage here, were we confirm that each use case works as we expect it.  I also think you have the best context right now to create the mock object, so let's expand it out to cover all cases.  

We do have a fixtures file just FYI, but I kinda like that this is separate as I feel like that file is too big.  But you may want to expose this info through that, but keep it a separate file.

frontend/src/components/application/ApplicationHistoryTable.tsx
glenblosser-nava: Unit tests expanded to cover all events and split into a standalone file

frontend/src/services/fetch/fetchers/applicationFetcher.ts
glenblosser-nava: Followed up with Michael, maximum page size is 5000 but he suggested 500 should be plenty for now.

frontend/stories/components/application/history.mock.json
glenblosser-nava: Test coverage expanded

frontend/src/components/application/ApplicationHistoryTable.tsx
glenblosser-nava: Updated table component

frontend/src/components/TableWithResponsiveHeader.tsx
```diff
@@ -7,6 +7,7 @@ export type TableCellData = {
   cellData: ReactNode;
   stackOrder?: number; // -1 to hide
   className?: string;
+  style?: object;
```
ErinPattisonNava, line 10: Out of curiosity, did design push back on using the standard styling?  If we can simplify to standardized components that would reduce complexity.

frontend/tests/components/application/ApplicationHistoryTable.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+import { ApplicationHistoryTable } from "src/components/application/ApplicationHistoryTable";
+
+const applicationHistory = historyMock as ApplicationHistory[];
+
+jest.mock("next-intl", () => ({
+  useTranslations: () => useTranslationsMock(),
+}));
+
+describe("ApplicationHistoryTable", () => {
```
ErinPattisonNava, line 15: I'd like to see more coverage here, especially of the alert and the transforms.

MAINTAINERS.md
```diff
@@ -22,6 +22,7 @@ This is a list of maintainers for this project. See [CODEOWNERS](/.github/CODEOW
 * Glen Blosser
 * Dao Nguyen
 * Nathan Stricker
+* Amit Desai
```
ErinPattisonNava, line 25: Normally we have new folks add themselves as a good first PR?

frontend/src/app/[locale]/(base)/workspace/applications/application/[applicationId]/page.tsx
```diff
@@ -68,6 +73,18 @@ async function ApplicationLandingPage({ params }: ApplicationLandingPageProps) {
     }
     opportunity = opportunityResponse.data;
     attachments = response.data.application_attachments;
+    const historyResponse = await getApplicationHistory(
+      applicationId,
+      userSession?.token,
+    );
+    if (response.status_code !== 200) {
```
Comment (ErinPattisonNava) [inline frontend/src/app/[locale]/(base)/workspace/applications/application/[applicationId]/page.tsx]: If this fails, the whole page fails to load.  Can we please only have the error show on the history section?
Screenshot 2025-12-16 at 9 45 06 AM

Comment (glenblosser-nava) [reply frontend/src/app/[locale]/(base)/workspace/applications/application/[applicationId]/page.tsx]: I just re-tested it, and the error only shows on the history section. This logic statement only prevents applicationHistory from being set, it doesn't throw a top level error

MAINTAINERS.md
glenblosser-nava: Amit asked me to add him. It's not a significant thing either way

frontend/src/components/TableWithResponsiveHeader.tsx
glenblosser-nava: The figma shows the Activity section having more space allocated to it than the other columns. I didn't ask design if it's okay for them to be the same

glenblosser-nava:
> On application fail, the even doesn't show up when the error does on the top of the page, it shows up on the next refresh. I'm not sure how easy that is to fix, but it isn't ideal.

The next ticket I'm working on will resolve this issue, so it doesn't need to hold up this PR

frontend/src/app/[locale]/(base)/workspace/applications/application/[applicationId]/page.tsx
Comment (glenblosser-nava) [reply frontend/src/app/[locale]/(base)/workspace/applications/application/[applicationId]/page.tsx]: Screenshot 2025-12-16 at 5 04 47 PM

ErinPattisonNava (approved):
Looks good!  Tested all cases locally.

ErinPattisonNava (changes requested):
Sorry I missed this in my initial review.  We need to be filtering the fetcher as specified in the ticket to avoid showing users "Application submission created"

PR #7363, opened 2025-12-05
[Unticketed] Add soap-private-keys to terraform

jakobpederson:
## Summary

Fixes / Work for #6816 

## Changes proposed

Added env vars to the terraform file.

## Context for reviewers

We're going to change over from SOAP_AUTH_CONTENT to SOAP_PRIVATE_KEYS and part of this is updating the terraform file as well as on AWS parameter store.


prasnava (approved):
👍

Issue #7367, opened 2025-12-08
Update OpenAPI Base Container


prasnava:
Update the bae container with the new upgraded python version 3.14.

PR #7368, opened 2025-12-08
[Issue #7367]: Update Open API base container

prasnava:
## Summary

Updated the base container with the new python 3.14 version.

## Validation steps

The Open API GitHub actions should continue to run.

### DIFF
--- a/infra/api/app-config/env-config/environment_variables.tf
+++ b/infra/api/app-config/env-config/environment_variables.tf
@@ -93,9 +93,9 @@ locals {
       secret_store_name = "/api/${var.environment}/domain-verification-content"
     }
 
-    SOAP_AUTH_CONTENT = {
+    SOAP_PRIVATE_KEYS = {
       manage_method     = "manual"
-      secret_store_name = "/api/${var.environment}/soap-auth-content"
+      secret_store_name = "/api/${var.environment}/soap-private-keys"
     }
 
     SAM_GOV_API_KEY = {


prasnava:
Successfully tested here: https://github.com/HHS/simpler-grants-gov/actions/runs/20031905611/job/57442967831

api/docker-python-openapi
```diff
@@ -7,7 +7,7 @@ RUN dnf -y install cpio && \
     for rpm in /tmp/rpms/*.rpm; do rpm2cpio "$rpm" | cpio -idmv; done && \
     rm -rf /tmp/rpms
 
-FROM ghcr.io/hhs/python-base-image:72ba391@sha256:1e2d72700ecfbff3a65435a64ca321d65c7bb82128e91c5825d69997cb54c8cd
+FROM ghcr.io/hhs/python-base-image:b8b6289@sha256:fadc641c704cc98386246e531be93c35162ea6a56f1f92489d4ea71e016c65f0
```
chouinar, line 10: Not specific to this PR - is there a way we can make it so that the base image is built / referred to like stuff we pull from Dockerhub?

That is, we don't need to re-define this each time but could instead do `FROM ghcr.io/hhs/python-base-image:1.0` and just get everything until we change the version upstream due to significant changes.

I assume we haven't done this yet because it's complex - I think a lot of the dockerhub ones are built separately by different pipelines, which might be the

prasnava: We can add the version semantics to it, it's a little overhead, not something too complicated. Will add it to the backlog.

PR #7371, opened 2025-12-08
[Issue #7372]: Update urllib3 in the base container and update the API

prasnava:
## Summary

Updated the urlib3 version from 2.5.0 to 2.6.0 to resolve anchore scan CVE findings.

## Validation steps

The API should continue to work.

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -1,6 +1,5 @@
 
-FROM ghcr.io/hhs/python-base-image:b8b6289@sha256:fadc641c704cc98386246e531be93c35162ea6a56f1f92489d4ea71e016c65f0 AS base
-
+FROM ghcr.io/hhs/python-base-image:868756e@sha256:f1ee3ce8b993a266c6721d6c024b2e41b8aa4a70199536ead1f4b0b5fc18d33c AS base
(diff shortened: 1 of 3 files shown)


api/docker-python-base
```diff
@@ -62,6 +62,8 @@ RUN rm -rf "Python-${PYTHON_VERSION}" "Python-${PYTHON_VERSION}.tgz" && \
 # Ensure virtualenv embeds patched pip (25.3) for offline venv creation
 RUN python3 -m pip install --no-cache-dir --upgrade "virtualenv>=20.35.3"
 
+# Upgrade urllib3 to fix security vulnerabilities (GHSA-2xpw-w6gg-jr37, GHSA-gm62-xv2j-4w53)
+RUN python3 -m pip install --no-cache-dir --upgrade "urllib3>=2.6.0"
```
mdragon, line 66: Is there any value in combining this line with the prior? (Assuming that's possible with pip install) Or is the idea that once this version ships without needing to force the upgrade this makes it easier to remove gain?

prasnava: Yes and also for documentation we know why we added that on a separate line later.

Issue #7220, opened 2025-11-24
invite interactions on manage legacy users page 


ErinPattisonNava:
### Summary

We previously [created](https://github.com/HHS/simpler-grants-gov/issues/7247) a page with a table of the legacy users from grants.gov.  Now we want to add their statuses and allow actions to be performed on them.

Figma: https://www.figma.com/design/zPlDWcYM396szw2yL3UXvk/Simpler-Grants-2025?node-id=7993-90708&t=kgz5u9g8av1W0aSh-1

### Acceptance criteria

- [ ] Add third column to table
   - [ ] If user is active, don't show buttons but show message
   - [ ] If user is invited, don't show buttons but show message
   - [ ] Otherwise show invite button
      - [ ] Invite button 
         - [ ] If clicked call  `organizations/{organization_id}/invitations`
         - [ ]  Disable button until return from API
         - [ ] If API fails, enable the button and display the appropriate error message
         - [ ] If API succeeds, refresh page to get new table
      - [ ] Hide button will be covered in a follow up [ticket](https://github.com/HHS/simpler-grants-gov/issues/7728) 
- [ ] Unittests
- [ ] Design review before submitting a PR

Issue #6876, opened 2025-10-28
Automatically add X-SGG-Token header to server side API requests


doug-s-nava:
### Summary

When making a request to the simpler API, the X-SGG-Token header must be present on all authenticated requests to identify the user. Currently there is a decent amount of boilerplate around this functionality that should be simplified.

An idea of how this would work:

- update the endpoint configs with an "authenticated" property
- mark "authenticated: true" for all routes requiring user authentication
- whenever calling a server side fetch function with `requesterForEndpoint` for a configuration denoted as `authenticated` the `getDefaultHeaders` function will fetch the user's session and add their token on the X-SGG-Token header

Note that this will simplify some work around moving fetch calls from page components into the AuthorizationGate

### Acceptance criteria

- [ ] Adding the "X-SGG-Token" header is not done within any individual fetch function, but handled by the general fetcher based on new arguments

doug-s-nava:
(note, are there dynamic endpoints that are not used? If so we should remove them.)

Endpoints currently requiring the auth token:

- POST /users/[user-id]/invitations/[org-id]/organizations (toDynamicUsersEndpoint) : api/user/organization-invitations/id/handler
- GET  /users/[user-id]/organizations (toDynamicUsersEndpoint) : api/user/organizations/handler
- POST /users/token/refresh (userRefreshEndpoint) : services/auth/session
- PUT /applications/[app-id]/organizations/[org-id] (toDynamicApplicationsEndpoint) : /services/fetch/fetchers/addOrganizationToApplication
- POST /users/]user-id]/agencies (toDynamicUsersEndpoint) : /services/fetch/fetchers/agenciesFetcher
- POST /users/[user-id]/api-keys/list (toDynamicUsersEndpoint) : /services/fetch/fetchers/apiKeyFetcher
- POST /users/[user-id]/api-keys/ (toDynamicUsersEndpoint) : /services/fetch/fetchers/apiKeyFetcher
- PUT /users/[user-id]/api-keys/[api-key] (toDynamicUsersEndpoint) : /services/fetch/fetchers/apiKeyFetcher
- DELETE /users/[user-id]/api-keys/[api-key] (toDynamicUsersEndpoint) : /services/fetch/fetchers/apiKeyFetcher
- POST /application/start  (toDynamicApplicationsEndpoint) : /services/fetch/fetchers/applicationFetcher
[...]

PR #7377, opened 2025-12-08
[Issue #7247] invite legacy table data

ErinPattisonNava:
## Summary

Work for #7247  

## Changes proposed

This adds the table data to the skeleton page built out in [here](https://github.com/HHS/simpler-grants-gov/pull/7346).  Interactions on this table will be coming in a[ later ticket](https://github.com/HHS/simpler-grants-gov/issues/7220).  

## Context for reviewers

This PR doesn't include unittests yet as we want to get early feedback.  I'll either add unittests after the initial review or create a ticket to do them as an immediate followup.  

We don't have the information seeded yet for this table in a permanent way, so please ping me for the setup script to test.  There is a backend ticket for next sprint to make this more official. 

## Validation steps

- [ ] With `manageUsersOff`, login as a user with admin privileges to an org.
- [ ] Confirm you can get to the page by in the Header -> Workspace -> Organizations, Click Manage Users then click "Invite Grants.gov users.
- [ ] You should see a message saying the request is empty.
- [ ] Run the script to populate data.
- [ ] Refresh the page. 
- [ ] You now should see the full list of users.
- [ ] Check that the page isn't available for unauthenticated users.
- [ ] Check that the page isn't available for orgs where you don't have admin permissions.
- [ ] Confirm that errors for pulling either the [organization information](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/api/organizations_v1/organization_routes.py#L61) or the[ legacy user information](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/api/organizations_v1/organization_routes.py#L248C5-L248C35) results in a nice error page

### DIFF
--- a/frontend/tests/components/organization/manage-users/inviteLegacyUsers/InviteLegacyUsersTable.test.tsx
+++ b/frontend/tests/components/organization/manage-users/inviteLegacyUsers/InviteLegacyUsersTable.test.tsx
@@ -43,7 +43,7 @@ describe("InviteLegacyUsersTable", () => {
     expect(screen.getByText("keyInformationDetails")).toBeVisible();
   });
 
-  it.only("has a table with two columns", () => {
+  it("has a table with two columns", () => {
     const component = InviteLegacyUsersTable({
       organizationLegacyUsers: legacyUsers,
     });


frontend/src/app/[locale]/(base)/organization/[id]/manage-users/legacy/page.tsx
```diff
@@ -46,41 +72,94 @@ const InviteLegacyUsersErrorPage = ({
...
-  let organizationDetails;
-  try {
-    organizationDetails = await getOrganizationDetails(organizationId);
-  } catch (error) {
+const LegacyUsersTable = ({
+  organizationLegacyUsers,
+}: {
+  organizationLegacyUsers: OrganizationLegacyUser[];
+}) => {
+  const t = useTranslations();
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/organization/[id]/manage-users/legacy/page.tsx]: [nit] to dry up `t()` calls
```suggestion
  const t = useTranslations('InviteLegacyUsers');
```

frontend/src/app/[locale]/(base)/organization/[id]/manage-users/legacy/page.tsx
```diff
@@ -46,41 +72,94 @@ const InviteLegacyUsersErrorPage = ({
...
+            { cellData: user.email },
+            { cellData: formatFullName(user) },
+          ],
+        )}
+      />
+    </div>
+  );
+};
+
+const PageContentWithData = ({
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/organization/[id]/manage-users/legacy/page.tsx]: [nit] can we make this name more specific? maybe `LegacyUsersPageContentWIthData`? `LegacyUsersDataDisplay`?

frontend/src/app/[locale]/(base)/organization/[id]/manage-users/legacy/page.tsx
```diff
@@ -46,41 +72,94 @@ const InviteLegacyUsersErrorPage = ({
...
+      })}
+    </div>
+  );
+};
 
-  let organizationDetails;
-  try {
-    organizationDetails = await getOrganizationDetails(organizationId);
-  } catch (error) {
+const LegacyUsersTable = ({
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/organization/[id]/manage-users/legacy/page.tsx]: thoughts on putting some of these components in a separate file so that we can test them and potentially think about them separately from the context of the page? NextJS page files have their weird conventions and I sometimes feel like it's a good idea to just get away from them

frontend/src/services/fetch/fetchers/organizationsFetcher.ts
```diff
@@ -216,3 +218,52 @@ export const removeOrganizationUser = async (
...
+            order_by: "email",
+            sort_direction: "ascending",
+          },
+        ],
+      },
+    },
+  });
+
+  if (!resp.ok) {
+    await throwOnApiError(resp, {
```
doug-s-nava, line 261: cool new pattern! something we should roll out across our fetchers? I created this ticket as a starting point for building some more rigor into these functions, but that kind of error handling is a good candidate as well https://github.com/HHS/simpler-grants-gov/issues/6876

ErinPattisonNava: I stole this code from earlier in the file, so I cannot claim the new pattern 😆 Thanks for creating the ticket

doug-s-nava (approved):
✅ ✅ ✅ ✅ 

following directions here was able to get the page populated. https://betagrantsgov.slack.com/archives/C08MTT10SKU/p1764963084998479?thread_ts=1764954691.301459&cid=C08MTT10SKU

Looks good though the "Legacy invite" heading doesn't match designs. In any case, that doesn't seem like a blocker


Screenshot 2025-12-08 at 9 14 27 PM

ErinPattisonNava:
> Looks good though the "Legacy invite" heading doesn't match designs. In any case, that doesn't seem like a blocker
> 

@doug-s-nava Can you be more specific about what isn't matching the designs?

frontend/src/app/[locale]/(base)/organization/[id]/manage-users/legacy/page.tsx
Comment (ErinPattisonNava) [reply frontend/src/app/[locale]/(base)/organization/[id]/manage-users/legacy/page.tsx]: I did end up pulling them out into components because I couldn't get reasonable testing around the AuthorizationGate.   I think this pattern will feel better once we can have sub-page components instead of pulling them out into a shared components folder, as most of these aren't really intended to be shared.

doug-s-nava:
The page header in designs says "Manage users", implemented header says "Legacy invite"
Screenshot 2025-12-09 at 7 39 11 AM

ErinPattisonNava:
> The page header in designs says "Manage users", implemented header says "Legacy invite" 

That's the archived designs; the newer ones are right above it under "Fast Follow MVP".  That is the correct title, I did pass these pages by Chelsea before submitting the PR.  But glad you mentioned it so we could make sure we are aligned.

Issue #7381, opened 2025-12-08
Validate opportunity existence before saving or deleting a user saved opportunity


babebe:
### Summary

Currently, if we attempt to save or delete an opportunity that does not exist, the API throws an error.

 Update the POST /v1/users/:user_id/saved-opportunities endpoint to:

Check if the opportunity exists before performing any save or delete operation
Return a 404 `Not Found` response if the opportunity does not exist
Update/Create test

### Reproducibility

None

### How to reproduce the bug

You can use an invalid or deleted opportunity ID.
Attempt to save the opportunity.

### Browser

None

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #7385, opened 2025-12-09
[Issue #7381] Add opportunity existence check to saved-opportunities endpoint

babebe:
## Summary

Fixes for #7381  

## Changes proposed
The API now validates the opportunity ID before performing any save logic.
If the opportunity does not exist, the endpoint responds with 404 Not Found.
Added new test to confirm this behavior


api/src/services/users/create_saved_opportunity.py
```diff
@@ -5,12 +5,20 @@
...
 
 def create_saved_opportunity(db_session: db.Session, user_id: UUID, json_data: dict) -> None:
-    # Check if the record exists
+    """
+    Create or reactivate a saved opportunity for a user.
+
+    """
+    # Validate opportunity exists
+    opportunity_id = json_data["opportunity_id"]
+    get_opportunity(db_session, opportunity_id)
```
mdragon: Just checking to make sure I understand how this is working.  `get_opportunity` is throwing in a way that generates the 404 already, so just by calling it here explicitly we get that behavior in this endpoint?

chouinar: I'd actually say we shouldn't reuse `get_opportunity` in this case, mostly because `get_opportunity` does way too much right now, and fetches too much data that we don't need (was looking at metrics, our GET opportunity endpoint is getting slower, need to fix some joins). We also might rework our get opportunity endpoint in the next quad for publishing, and that'll mess with permissions a bit.

I'd say we should just be a bit more direct here and do basically `select * from opportunity where opportunity_id = '...' and is_draft is false`.

babebe: @mdragon  Yes get_opportunity already throws the 404 for us. @chouinar Agreed. I replaced it with a helper that does a straightforward existence check instead.

Issue #7330, opened 2025-12-04
Update fluent bit filter to reduce sql stack trace for new relic ingestion


prasnava:
When a step function fails we are are not able to see the logs in new relic, because the logs size exceeds the new relic ingestion size limit. 

The goal of this ticket is to reduce the log size for sql stack traces so we can ingest them in new relic. 

Fluent allows us to create custom filters in Lua to filter out stack traces that we don't need.

PR #7388, opened 2025-12-09
[Issue #7330]: Update fluentbit logs with sql truncation - draft

prasnava:
## Summary

Added a custom fluent bit filter to truncate sql logs.

## Validation steps

The logs should continue to work as normal and long sql errors should be truncated.

### DIFF
--- a/api/src/api/users/user_schemas.py
+++ b/api/src/api/users/user_schemas.py
@@ -14,6 +14,7 @@ from src.api.schemas.search_schema import StrSearchSchemaBuilder, UuidSearchSche
 from src.constants.lookup_constants import (
     ApplicationStatus,
     ExternalUserType,
+    OpportunityStatus,
     OrganizationInvitationStatus,
     Privilege,
 )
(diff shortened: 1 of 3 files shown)


prasnava:
Deployed to api dev [here: ](https://github.com/HHS/simpler-grants-gov/actions/runs/20069806297)

fluentbit/test-fluentbit.yml
```diff
@@ -0,0 +1,42 @@
+#
+# Test configuration for Fluent Bit Lua filter
+#
+# Run with:
+#   cd fluentbit
+#   docker run --rm -v $(pwd):/fluent-bit/etc fluent/fluent-bit:4.0.8 -c /fluent-bit/etc/test-fluentbit.yml
+#
```
prasnava: Created this for local testing for the fluent bit filter.

fluentbit/fluentbit.yml
```diff
@@ -19,6 +21,23 @@ pipeline:
...
+  filters:
+    # Parse the JSON in the 'log' field and merge fields to the record level
+    # This allows New Relic to query fields like aws.ecs.task_id directly
+    # NOTE: Must run BEFORE the Lua filter, otherwise escaped newlines break JSON parsing
+    - name: parser
+      match: "**"
+      key_name: log
+      parser: json
+      reserve_data: true
+      preserve_key: false
```
chouinar, line 33: How exactly does this adjust the data, do you have an example of how the data is restructured? Want to make sure it's not changing the shape of something we use in New Relic / if it does, we can figure out what changes we need to make elsewhere.

prasnava: The logs coming in have nested fields. The main key is "log" and then ecs adds extra keys to it. So originally it looks like this: 

```
 {
    "ecs_task_id": "task_id_here",
    "container_name": "api",
    "log": "{\"message\": \"log entry\", \"level\": \"INFO\", \"timestamp\": \"...\"}"
  }
```

The parser will merge them into one record at the root level with no nested fields so we can query them easier in new relic. We don't lose any data with this approach. Everything stays intact. Before we added this, the logs were coming into new relic, but we were not able to query them by the ecs task id.

The result is the original log gets converted to below: 

```
 {
    "ecs_task_id": "task_id_here",
    "container_name": "api",
    "message": "log entry",
    "level": "INFO",
    "timestamp": "..."
  }
```

Issue #7290, opened 2025-12-02
[CommonGrants] Fix timestamps in opportunity routes


widal001:
### Summary

We received the following message via the contact us inbox:

> I wanted to flag an issue in the API as a bug to fix: When I query latest modified opportunities using the common-grants/opportunities end point, the returned json sets createdAt and lastModifiedAt values to the current UTC time for all opportunities returned instead of the timestamp it was actually created/modified. When I double check specific opportunity_ids in common-grants with the common-grants/opportunities/{oppId} endpoint, it shows the correct lastModifiedAt and createdAt values (or at least values that are not the current time).

We should diagnose and fix this discrepancy.

### Acceptance criteria

- [ ] The `lastModifiedAt` and `createdAt` reflect the timestamps of the opportunity record rather than the current timestamp

DavidDudas-Intuitial:
The problem is in [this code](https://github.com/HHS/simpler-grants-gov/blob/08dd3895e0675d7c706e14eb9463225cd308aa35/api/src/services/common_grants/transformation.py#L296-L297)

DavidDudas-Intuitial:
Relevant context: https://github.com/HHS/simpler-grants-gov/pull/6542#discussion_r2388061787

<img width="807" height="508" alt="Image" src="https://github.com/user-attachments/assets/12645847-80f9-4799-8763-c58ede2488ef" />

PR #7396, opened 2025-12-09
Fix bug for created_at and updated_at in common grants search

jcrichlake:
## Summary

Fixes  [bug reported in common-grants sdk ](https://github.com/HHS/simpler-grants-gov/issues/7290)

## Changes proposed

Fixed mapping for `updated_at` and `created_at` fields in the service for the common-grants wrapper APIs.  Instead of using the root level object we are updating the transform to use the dates from the summary object field. 

## Context for reviewers

This change only impacts the common-grants specific code and leaves untouched the v1 routes.

## Validation steps

Call either the common grants search route using curl or postman and validate that the `updated_at` and `created_at` dates are populating correctly.  Most of these dates should not be todays date, which is what was happening prior to this code change


api/tests/src/services/common_grants/test_transformation.py
```diff
@@ -436,8 +450,8 @@ def test_transform_search_result_to_cg_with_missing_data(self):
         assert result.title == "Test Opportunity"
         assert result.description == "No description available"
         assert result.status.value == "open"
-        assert result.created_at == datetime(2024, 1, 1, 12, 0, 0)
-        assert result.last_modified_at == datetime(2024, 1, 2, 12, 0, 0)
+        assert result.created_at.date() == datetime.now().date()
+        assert result.last_modified_at.date() == datetime.now().date()
```
chouinar: I'd recommend if you want to validate the behavior of any sort of `now()` check, use the freezegun library (already installed) to freeze time on the tests.

https://github.com/spulec/freezegun

api/src/services/common_grants/transformation.py
```diff
@@ -293,8 +295,8 @@ def transform_search_result_to_cg(opp_data: dict) -> OpportunityBase | None:
             ),
             source=validate_url(summary.get("additional_info_url")),
             custom_fields={},
-            createdAt=opp_data.get("created_at") or datetime.now(timezone.utc),
-            lastModifiedAt=opp_data.get("updated_at") or datetime.now(timezone.utc),
+            createdAt=summary.get("created_at") or datetime.now(timezone.utc),
+            lastModifiedAt=summary.get("updated_at") or datetime.now(timezone.utc),
```
chouinar: We have a datetime_util I'd recommend using for any datetime fetching - handles making sure it's doing it the current recommended way as there are several deprecated methods to fetching the current time stamp. While this is the current way, when they adjust it again and change where the imports are again, easiest way is to have it in one place we've found:

https://github.com/HHS/simpler-grants-gov/blob/main/api/src/util/datetime_util.py#L7

api/tests/src/services/common_grants/test_transformation.py
jcrichlake: Added, thanks for tip!

api/src/services/common_grants/transformation.py
jcrichlake: Good call out, updated in latest push!

DavidDudas-Intuitial (approved):
LGTM

PR #7397, opened 2025-12-09
[Issue #7082] Create script to automate usage of the new form instructions endpoint

### LINKED ISSUE #7082: Create script to automate usage of the new form instructions endpoint
### Summary

Managing form instructions documents is slow and cumbersome as it requires manually uploading the file to s3 with very specific info, and then manually insert it into the DB. To alleviate this, we want to create a new endpoint and script to automate this process.

This ticket is to create the script to automate the usage of the new endpoint from ticket #7081 

We want to create a script similar to our [update-form script](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/task/forms/update_form_task.py) but for updating the form instructions.

When running the script, we should have it take in 4 parameters:
* form_id - the ID of the form
* form_instruction_id - the ID of the instructions
* file_path - A path to a file that we'll upload
* environment - Which environment we want to update the form in (local, dev, staging, training, prod)

These are the 3 parameters that the endpoint we created will use + the environment. 

This script should:
* Require an auth token be setup via an env var (see how the existing update-form script does it, do the same thing)
* Error if not being run locally
* Handle sending the request to the URL which we'll need to configure per environment - see also how the other scripts do this.
* Error if the response is not a 200.

Really, follow roughly a similar pattern to the update form task, although you don't need to first grab the form, all the info needed is passed into the request.

For context, here’s what it takes to setup forms the first time in an environment: https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2259157056/Setup+Forms+-+Nov+17th+2025

### Acceptance criteria

- [ ] script created
- [ ] Simple tests created
- [ ] Verified a form instruction can be uploaded to our dev environment with the script - doesn't need to be a real instruction file or connected to a real form, just make up some IDs

mikehgrantsgov:
## Summary

Fixes #7082  

## Changes proposed

Add a `update_form_instruction_task` that calls an endpoint to call the form_instructions endpoint
Add docs to README
Add unit tests

## Context for reviewers

Managing form instructions documents is slow and cumbersome as it requires manually uploading the file to s3 with very specific info, and then manually insert it into the DB. To alleviate this, we want to create a new endpoint and script to automate this process.

## Validation steps

See tests for usage and e.g.:
`make cmd args="task update-form-instruction --environment=dev --form-id=c3c5c7e9-0b24-41f8-8da3-98241fb341fe --form-instruction-id=a1b2c3d4-5e6f-7890-abcd-ef1234567890 --file-path=/path/to/EXAMPLE-V1.0-Instructions.pdf"`

### DIFF
--- a/api/src/task/forms/form_task_shared.py
+++ b/api/src/task/forms/form_task_shared.py
@@ -61,14 +61,15 @@ class BaseFormTask(abc.ABC):
 
         Note: Content-Type is not set here because requests will automatically
         set the correct Content-Type with boundary for multipart/form-data.
+
+        Note: X-Auth is not included here because the form instruction endpoint
+        only supports X-API-Key authentication.
         """
         headers = {
             "Accept": "application/json",
         }
         if self.config.form_x_api_key_id is not None:
             headers["X-API-Key"] = self.config.form_x_api_key_id
-        if self.config.non_local_api_auth_token is not None:
-            headers["X-Auth"] = self.config.non_local_api_auth_token
 
         return headers
 
--- a/api/tests/src/task/forms/test_form_task_shared.py
+++ b/api/tests/src/task/forms/test_form_task_shared.py
@@ -105,10 +105,11 @@ def test_build_headers():
 def test_build_file_upload_headers():
     dummy_form_task = DummyFormTask()
 
+    # X-Auth is not included because the form instruction endpoint
+    # only supports X-API-Key authentication
     assert dummy_form_task.build_file_upload_headers() == {
         "Accept": "application/json",
         "X-API-Key": "fake-x-api-key",
-        "X-Auth": "fake-auth-token",
     }
 
 


api/src/task/forms/form_task_shared.py
```diff
@@ -47,6 +56,22 @@ def build_headers(self) -> dict:
...
+        Note: Content-Type is not set here because requests will automatically
+        set the correct Content-Type with boundary for multipart/form-data.
+        """
+        headers = {
+            "Accept": "application/json",
+        }
+        if self.config.form_x_api_key_id is not None:
+            headers["X-API-Key"] = self.config.form_x_api_key_id
+        if self.config.non_local_api_auth_token is not None:
+            headers["X-Auth"] = self.config.non_local_api_auth_token
```
chouinar: If we have this header function separate anyways, we should skip the `X-Auth` as we didn't add that to the endpoint.

mikehgrantsgov: Makes sense, updated this

chouinar (approved):

Issue #6940, opened 2025-11-04
BE - Support for pagination for the pending users endpoint (list organization invitations)


eric-serrato95:
### Summary

Capturing the effort to support pagination for list organization invitations endpoint.

[API Pagination Implementation Guide](https://navasage.atlassian.net/wiki/x/OYDxhg) 

### Acceptance criteria

- [ ] API provides pagination support for the pending users endpoint 
- [ ] Table allows for 10 rows 
- [ ] Table support sorting by Email Asc (A-Z) by default

joshtonava:
We have filtering to this endpoint so we can filter by, but we need pagination 

```python
class OrganizationInvitationStatus(StrEnum):
    ACCEPTED = "accepted"
    REJECTED = "rejected"
    EXPIRED = "expired"
    PENDING = "pending"
```

PR #7406, opened 2025-12-09
[Issue #6940] Add Pagination and Default Email Sorting to List Organization Invitations Endpoint

babebe:
## Summary

Fixes / Work for #6940 

## Changes proposed

- Added pagination support to `POST /v1/organizations/:organization_id/invitations/list`
- Introduced default sorting by `invitee_email`. 
- Set defaults for page_size and page_offset.
- Updated the endpoint to use the new filters and pagination Pydantic models.
- Added default sort order and pagination behavior tests and updated current tests to include pagination


api/src/api/organizations_v1/organization_schemas.py
```diff
@@ -245,6 +245,17 @@ class OrganizationInvitationListRequestSchema(Schema):
         allow_none=True,
         metadata={"description": "Filters to apply to the invitation list"},
     )
+    pagination = fields.Nested(
```
joshtonava, line 248: My only concern with this change is it will cause the frontend that is consuming this to break since now pagination is required. 

We can create a new version of this api and make a follow up ticket for the frontend to use that version.

I want to circle back to our pagination implementation (in a separate ticket) and allow for this to be optional with defaults provided.

chouinar: I think making it so pagination has a default would be easier than rebuilding the entire endpoint, all we really need to do is set a default here, we already do it for the sorting half of pagination anyways.

What happens if on the page_size/page_offset, we add a default (of 25 and 1 respectively), and make this field not required. Does that work, or do we need to make a default of this entire object?

babebe: @joshtonava  @chouinar I removed the required tag from pagination, making it optional. Defaults for page_size and page_offset are applied if not provided so the API works even when the frontend omits the values

api/src/pagination/pagination_schema.py
```diff
@@ -90,12 +92,12 @@ class MyPaginationSchema(Schema):
             **additional_sort_order_params,
         ),
         "page_size": fields.Integer(
-            required=True,
+            load_default=default_page_size,
```
chouinar: I don't think we can remove required from this - we wouldn't want to break assumptions in other endpoints. We'd need to make it so whether the field is required is passed into the function, defaulting to True.

babebe: Updated.  Required flag now configurable to support defaults.

chouinar (approved):

Issue #7379, opened 2025-12-08
Implement system for logging in to deployed site with playwright


doug-s-nava:
### Summary

In order to test that our log in system is working we need to be able to run E2E tests against a deployed environment where a test user goes through the full login.gov based login process successfully.

We are currently limited by a couple of factors:

* no suitable test account
* no ability to bypass MFA requirement

@Bhavna-Ramachandran has developed a solution for the second hurdle here, using an auth key along with a node library to generate MFA codes within the test suite.

For creating a test account we should

* create a test user email using Google Groups
* store credentials for the test email in 1pass

In order to prevent leaking the email, password and auth key for the test user we will need to

* store the values in github actions secrets or AWS SSM parameters
* reference them in the `env` section of the github actions definition https://github.com/HHS/simpler-grants-gov/blob/7c33bd2730b8a06c7c651f3c0a4b8560999f2700/.github/workflows/ci-frontend-e2e.yml#L80

To ensure that this works

* write a Playwright test in the E2E suite that asserts that the user can log in successfully
* make sure that this test passes when run against staging (note that it will need to be skipped before merge, since we are not able to run tests against staging until https://github.com/HHS/simpler-grants-gov/issues/7257, and https://github.com/HHS/simpler-grants-gov/issues/7255 + followup OR https://github.com/HHS/simpler-grants-gov/issues/7307 + followup are complete)

Note that in order to test in CI, we will also:
* add secret and reference to playwright base URL in order to direct CI playwright runs to staging (this may be a separate ticket)

### Acceptance criteria

- [ ] login test against staging passes locally and in CI
- [ ] all necessary credentials are available in a secure location

Bhavna-Ramachandran:
PR [7411](https://github.com/HHS/simpler-grants-gov/pull/7411) for the login flow automation using Login.gov with MFA - has been reviewed and approved by @doug-s-nava and merged into main. Hence marking this ticket as Done.

PR #7411, opened 2025-12-10
Add login folder structure, BDD feature, step definitions, and standalone Playwright spec

Bhavna-Ramachandran:
## Summary
Fixes / Work for #[7379](https://github.com/HHS/simpler-grants-gov/issues/7379) 
This PR creates the folder structure, BDD feature, step definitions, and a standalone Playwright spec for the login flow using Login.gov with MFA.
Here, the same login scenario exists in both approaches for convenience: the BDD feature + step definitions and the standalone Playwright script.

## Changes proposed
- Created directory: tests/e2e/login/ to isolate all login-related automation.
- Added subfolders:
- features/ — contains the BDD .feature file: login_with_login_gov.feature
- steps/ — contains the corresponding step definitions: login_with_login_gov.ts
- Added standalone Playwright test spec: login_with_login_gov.spec.ts for quick end-to-end login testing with MFA.

## Rationale for having both approaches
•	Standalone Playwright script: Provides fast iteration and is useful for CI/CD runs.
•	BDD feature + step definitions: Provides readable, documented, and reusable tests that align with business rules, making it easy for the team to review, maintain, and collaborate.

## Context for reviewers
This PR implements the folder structure, BDD feature, step definitions, and standalone Playwright spec for automating login using Login.gov with MFA, to simulate a real-world user flow on any deployed environment.
•	Environment variables (LOGIN_EMAIL, LOGIN_PASSWORD, LOGIN_MFA_KEY, STAGING_BASE_URL) must be set in .env.local or CI; no secrets are committed.
•	The structure allows both Cucumber BDD tests and standalone Playwright tests to run independently.

## Validation steps
- Ensure environment variables are present in .env.local
- Run the standalone Playwright test using command below:
- npx playwright test frontend/tests/e2e/login/login_with_login_gov.spec.ts
- Should navigate to staging, perform login + MFA, and assert the user email is visible post-login.
- (Optional) Run the Cucumber scenario (BDD flow) using command below:
- npm run test:cucumber
- Should execute the login_with_login_gov.feature successfully.

### DIFF
--- a/.github/workflows/ci-frontend-e2e.yml
+++ b/.github/workflows/ci-frontend-e2e.yml
@@ -26,9 +26,6 @@ jobs:
     name: Run Local E2E Tests
     runs-on: ubuntu-22.04
 
-    env:
-      PLAYWRIGHT_TARGET_ENV: local
-
     strategy:
       fail-fast: true
       matrix:


frontend/tests/e2e/apply/features/happy_path_apply.feature
```diff
@@ -0,0 +1,76 @@
+Feature: Happy Path – Apply Workflow
```
doug-s-nava, line 1: can you take this file out of this PR?

frontend/tests/e2e/login/login_with_login_gov.spec.ts
```diff
@@ -0,0 +1,51 @@
+/* eslint-disable no-console */
+import { test, expect } from '@playwright/test';
+import { authenticator } from 'otplib';
+import * as dotenv from 'dotenv';
+import * as path from 'path';
+
+// Load environment variables from frontend/.env.local
+dotenv.config({ path: path.resolve(__dirname, '../../../.env.local') });
+
+// Validate required env variables
```
doug-s-nava: 👍 

No need to think about this now, but as a next step we'll want to figure out how we want to properly use Playwright's system to get environment variables into the test system without referencing them directly. I did it this way in my branch, but we'd likely want to remove the references from the spec files into a config of some sort eventually

https://github.com/HHS/simpler-grants-gov/compare/main...e2e-staging#diff-3a85e4e5812fd9da5b2e1925fa0ddd321f3b121c673da0601c11f82a584fdd65R32

Some relevant docs here as well https://playwright.dev/docs/test-fixtures

frontend/tests/e2e/login/login_with_login_gov.spec.ts
```diff
@@ -0,0 +1,51 @@
+/* eslint-disable no-console */
+import { test, expect } from '@playwright/test';
+import { authenticator } from 'otplib';
+import * as dotenv from 'dotenv';
+import * as path from 'path';
+
+// Load environment variables from frontend/.env.local
+dotenv.config({ path: path.resolve(__dirname, '../../../.env.local') });
```
doug-s-nava: I'm not sure if running dotenv.config twice in the project will work - in general it should probably be only run once. What happens if we remove this?

frontend/tests/e2e/login/login_with_login_gov.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+
+// Validate required env variables
+const email = process.env.LOGIN_EMAIL;
```
doug-s-nava: can we make this env var names more specific? ex:
* PLAYWRIGHT_TEST_USER_EMAIL
* PLAYWRIGHT_TEST_USER_PASSWORD
* PLAYWRIGHT_TEST_USER_MFA_KEY

frontend/tests/e2e/login/login_with_login_gov.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+const password = process.env.LOGIN_PASSWORD;
+const authKey = process.env.LOGIN_MFA_KEY;
+const baseUrl = process.env.STAGING_BASE_URL;
```
doug-s-nava: you should be able to use the PLAYWRIGHT_BASE_URL env var already defined here https://github.com/HHS/simpler-grants-gov/blob/8f9e91e3eaf5dd9b7ad20212b54e8effcf510f63/frontend/tests/playwright.config.ts#L25

frontend/tests/e2e/login/steps/login_with_login_gov.ts
```diff
@@ -0,0 +1,90 @@
+/* eslint-disable no-console */
```
doug-s-nava: I like where you're going with this but integrating with cucumber or another system is out of scope for now. Feel free to stash this file in a branch for later reference, but it can be removed from this PR

frontend/tests/e2e/apply/features/happy_path_apply.feature
Bhavna-Ramachandran: Acknowledged. I have removed this. Resolving this comment.

frontend/.gitignore
```diff
@@ -55,3 +55,9 @@ npm-debug.log*
 # pa11y
 screenshots-output/*
 pa11y_output.txt
+
+# BDD cucumber steps (stashed for future use)
+/tests/e2e/login/steps/
```
doug-s-nava: if you don't want to commit these files, you can remove them from the PR instead of adding this to the ignore - someone should be able to help with that

.github/workflows/ci-frontend-e2e.yml
```diff
@@ -115,6 +121,46 @@ jobs:
           path: /home/runner/work/simpler-grants-gov/simpler-grants-gov/frontend/blob-report
           retention-days: 1
 
+  e2e-login-staging:
```
doug-s-nava: It might be easier to set up a completely new file for testing this. Alternatively, you can add the test script to the e2e directory, and add the env vars to the main workflow definition here. We'll need to do that eventually, up to you if you want to do that now or isolate things for testing at this point

frontend/next.config.js
```diff
@@ -54,7 +54,7 @@ const headers = [
   {
     // static pages are stored for 6 hours, refreshed in the background for
     // up to 10 minutes, and up to 12 hours if there is an error
-    source: "/:path*",
+    source: "/(.*)",
```
doug-s-nava: what's the story behind this change? Did you notice issues with the caching headers that this fixes?

frontend/next.config.js
```diff
@@ -153,13 +153,6 @@ const nextConfig = {
   async headers() {
     return headers;
   },
-  // see https://nextjs.org/docs/messages/api-routes-response-size-limit
```
doug-s-nava, line 156: it looks like the format here has changed, from what I'm seeing in the docs. Instead of making the change in this PR, can you open a new ticket? https://nextjs.org/docs/messages/api-routes-response-size-limit

frontend/package.json
```diff
@@ -65,6 +67,7 @@
   },
   "devDependencies": {
     "@chromatic-com/storybook": "^1.9.0",
+    "@cucumber/cucumber": "^10.9.0",
```
doug-s-nava: I'd like to avoid adding any of these packages unless they're necessary for this change. Are any of this critical to getting the login path script working?

frontend/src/components/RootLayoutWrapper.tsx
```diff
@@ -7,6 +7,7 @@ import * as newrelic from "newrelic";
 import { environment } from "src/constants/environments";
 import { NewRelicWithCorrectTypes } from "src/types/newRelic";
 
+import Head from "next/head";
```
doug-s-nava: I know we see warnings around this, but had problems with this change in the past. If there is not an existing ticket for this, can you make one and revert the change here?

frontend/src/lib/newrelic-server.ts
```diff
@@ -0,0 +1,42 @@
+// No "use server"
```
doug-s-nava: what is this doing? can we remove it from here and maybe put it in another ticket?

frontend/src/lib/useActionStateShim.ts
```diff
@@ -0,0 +1,10 @@
+// Compatibility shim for useActionState (React Canary API)
```
doug-s-nava: what is this for? Can this be a separate PR?

frontend/tests/components/workspace/StartApplicationModal/StartApplicationModal.test.tsx
```diff
@@ -1,4 +1,4 @@
-import { act, render, screen, waitFor } from "@testing-library/react";
```
doug-s-nava, line 1: can we move this to a new pr

frontend/tests/e2e/login/features/login_with_login_gov.feature
```diff
@@ -0,0 +1,11 @@
...
+  As a user
+  I want to sign in using Login.gov
+  So that I can access my Simpler Grants account
+
+  Background:
+    Given I navigate to the Simpler Grants staging site
+
+    @Login-LoginDotGov
+    Scenario: Successful Login.gov authentication with MFA
+        Given the user is logged in
```
doug-s-nava: is this complete? Seems like it is missing some steps

frontend/tests/e2e/login/steps/world.ts
```diff
@@ -0,0 +1,11 @@
+// /workspaces/simpler-grants-gov/frontend/tests/e2e/login/steps/world.ts
```
doug-s-nava: as we have not implemented cucumber yet, can we stash this for later?

frontend/src/services/fetch/fetchers/organizationsFetcher.test.ts
```diff
@@ -31,33 +31,33 @@ type FetchImpl = (args: FetchArgs) => FetchResponse;
 
 type FetchWithMethodFn = (type: string) => FetchImpl;
 
-const fetchOrganizationMock = jest.fn<FetchResponse, [FetchArgs]>();
```
doug-s-nava, line 34: I love these changes but can we move them to a new PR to avoid scope creep?

frontend/tests/e2e/login/login_with_login_gov.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+  await page.screenshot({ path: preMfaScreenshot, fullPage: true });
+  await test
+    .info()
+    .attach("before-mfa", { path: preMfaScreenshot, contentType: "image/png" });
+
+  // --- Step 5: Locate MFA input with retries ---
+  let mfaInput: Locator | undefined;
+
+  for (let attempt = 1; attempt <= 3; attempt++) {
+    await test.step(`MFA retry attempt ${attempt}`, async () => {
```
doug-s-nava: if you think there's a chance this could be reusable, can we define this step separately from the test?

frontend/tests/e2e/login/steps/login_with_login_gov.ts
```diff
@@ -0,0 +1,91 @@
...
```
doug-s-nava: we may want to think a little bit more about how we'll use step files before we introduce them. Would you object to holding off on this?

frontend/src/services/fetch/fetchers/organizationsFetcher.test.ts
Bhavna-Ramachandran: This change was minor and was done to unblock login test. The failure on this was blocking tests for the login spec. Hence I just updated it a bit to resolve errors and unblock login spec test. I suggest we would have to do a thorough cleanup as maintenance task sometime in the future

frontend/next.config.js
Bhavna-Ramachandran: I think I made that as I saw some root path errors - With /(.*) I did not see those errors.

frontend/tests/e2e/login/steps/login_with_login_gov.ts
Bhavna-Ramachandran: Acknowledged

frontend/.gitignore
Bhavna-Ramachandran: Acknowledged.

frontend/tests/e2e/login/features/login_with_login_gov.feature
Bhavna-Ramachandran: This is complete and was intended to be a reusable login feature - the Login.gov + MFA user flow details were abstracted into the step definitions to keep scenarios concise for use in other user flows. However, since I’ve removed those steps to address one of your comments (as they would require Cucumber), it makes sense to add the detailed login steps in this feature file. Please let me know your preference.

Proposed feature scenario (instead of the current one liner) is below:
Scenario: User signs in using Login.gov with MFA
    Given the user launches the application
    When the user clicks the "Sign in" button
    And the user enters a valid email and password
    And the user submits the login form
    And the user enters the authentication code
    And the user submits the MFA form
    Then the user is successfully logged in

frontend/tests/playwright.config.ts
```diff
@@ -1,82 +1,126 @@
...
+  return process.env[name];
+};
+
+// Base URLs with defaults
+const localBaseUrl = getEnv("LOCAL_BASE_URL") || "http://127.0.0.1:3000";
+const stagingBaseUrl =
+  getEnv("STAGING_BASE_URL") || "https://staging.simpler.grants.gov";
+
+// Use PLAYWRIGHT_BASE_URL or fallback based on environment
+const baseUrl =
```
doug-s-nava: I'm approaching this same complexity differently in my PR here https://github.com/HHS/simpler-grants-gov/pull/7435/files#diff-2c4a771dcf6db3f335f43dc0a2fd9ea8dd8e658e6668cf8245f67d11752ff599R11

but looking at your approach, I think there is a middle ground that probably better than either one. Where my branch sets two env vars - one for the base url, and one for the target env - and your sets two different ones - one for local base url, and one for staging base url - maybe a better solution is to:

* only pass in the target environment as an env var
* hardcode the base urls for each target environment within the code

it is worth handling this now as it will mean fewer changes to the existing e2e GH action definition

frontend/tests/playwright.config.ts
```diff
@@ -1,82 +1,126 @@
...
     },
-
     {
-      name: "firefox",
+      name: "local-e2e-firefox",
+      testDir: "./e2e",
+      grepInvert: /@login/,
+      testIgnore: "login/**",
       use: {
-        ...devices["Desktop Firefox"], // firefox doesn't support clipboard-write or clipboard-read
```
doug-s-nava, line 51: I think these comments are still valid - would you recommend another place to document this?

frontend/tests/playwright.config.ts
```diff
@@ -1,82 +1,126 @@
...
 
-  /* Run your local dev server before starting the tests */
-  webServer: process.env.PLAYWRIGHT_BASE_URL
-    ? undefined
-    : {
-        command: "npm run start",
-        url: "http://127.0.0.1:3000",
-        reuseExistingServer: !process.env.CI,
+    // Staging login tests
+    {
```
doug-s-nava: Is the intention to run against local and staging whenever we run tests? I think that's what this will do

frontend/.env.development
```diff
@@ -21,7 +21,6 @@ ENVIRONMENT=dev
 
 # Hardcode public auth key for local to local calls
 # This is also hardcoded and checked-in on the API
-API_AUTH_TOKEN=LOCAL_AUTH_12345678
```
doug-s-nava, line 24: why was this removed?

.github/workflows/ci-frontend-e2e.yml
```diff
@@ -82,7 +89,10 @@ jobs:
           TOTAL_SHARDS: ${{ matrix.total_shards }}
           CURRENT_SHARD: ${{ matrix.shard }}
         run: |
-          npm run test:e2e
+          npx playwright test \
+            --config=tests/playwright.config.ts \
+            ./e2e \
+            --shard="${CURRENT_SHARD}/${TOTAL_SHARDS}"
```
doug-s-nava: what is this doing?

.github/workflows/ci-frontend-e2e-login-staging.yml
```diff
@@ -0,0 +1,66 @@
+name: Staging Login E2E Tests
```
doug-s-nava: I don't see this job in the github actions UI - has this been tested or used for testing? Do you think it's valuable to keep around given this follow up ticket, where we would be removing it anyway? https://github.com/HHS/simpler-grants-gov/issues/7973

frontend/tests/e2e/login/login_with_login_gov.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+  // --- Test ---
+  test("Login.gov authentication with MFA (PR CI)", async ({
+    page,
+    context,
+  }: {
+    page: Page;
+    context: BrowserContext;
+  }) => {
+    // Validate and load envs inside the test so we don't throw at import time
+    const loginEmail: string = requireEnv(email, "STAGING_TEST_USER_EMAIL");
```
doug-s-nava, line 42: why not export these values from the config file instead?

frontend/tests/e2e/login/login_with_login_gov.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+      path: step3Path,
+      contentType: "image/png",
+    });
+
+    // --- Step 6: Generate OTP and submit ---
+    const oneTimeCode: string = authenticator.generate(loginAuthKey);
+    await mfaInput.fill(oneTimeCode);
+
+    const mfaSubmitButton: Locator = page
+      .locator('button[type="submit"]:not(:has-text("Cancel"))')
```
doug-s-nava, line 169: worth discussing later on our approach to locators - Playwright prefers using the testing-library style locators such as `getByRole` but this suite tends to not use those much at this point

frontend/tests/e2e/login/login_with_login_gov.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+      .filter({ visible: true })
+      .first();
+    await mfaSubmitButton.click();
+
+    // --- Step 7: Handle mobile dropdown if needed and confirm login success ---
+    const viewport = page.viewportSize();
+    const isMobile = viewport ? viewport.width <= 480 : false;
```
doug-s-nava: could be worth standardizing how we do this - in the tests I've written I've done something like this, which I prefer as it doesn't rely on a magic number
```
const isMobile = !!project.name.match(/[Mm]obile/);
```
see https://github.com/HHS/simpler-grants-gov/blob/9d78eed6f51317c822f7a0fcee30f8ad866368e5/frontend/tests/e2e/search/search.spec.ts#L48 as an example

doug-s-nava (commented):
some non blocking thoughts on the test itself

.github/workflows/ci-frontend-e2e.yml
Bhavna-Ramachandran: I added this during debugging to get visibility into how shards and the Playwright config were being applied. I will swtich it back to what it was before

frontend/.env.development
Bhavna-Ramachandran: I have updated this file to add this.

.github/workflows/ci-frontend-e2e-login-staging.yml
Bhavna-Ramachandran: Initially run against staging was a second job in the test workflow file, then per our discussion I moved it to a separate file. Given the follow-up ticket (https://github.com/HHS/simpler-grants-gov/issues/7973) proposes removing it, I don’t think it’s valuable to keep, hence I will remove it.

frontend/tests/e2e/login/login_with_login_gov.spec.ts
Bhavna-Ramachandran: I saw issues when CI was trying to run local tests - even though the local test runner skips the Login spec file, the setup process was still trying to import or validate variables from everything under e2e. That’s why I added the environment validation inside the test function. This approach improves test isolation and error reporting, making it easier to debug missing environment variables and prevents global test suite failures caused by import-time errors.

frontend/tests/playwright.config.ts
Bhavna-Ramachandran: You are right. Originally the intention was to run local E2E and staging (login/MFA) together.
After our discussion, we split the staging/login flow into a separate workflow, but I missed updating playwright.config.ts at the same time. I’ll update it now to be environment specific.

frontend/tests/e2e/login/login_with_login_gov.spec.ts
Bhavna-Ramachandran: Acknowledged. I have made that change.

frontend/tests/e2e/login/login_with_login_gov.spec.ts
Bhavna-Ramachandran: I understand. I went with a more defensive CSS-based locator here because the MFA submit button doesn’t always have stable accessible text. Do you want me to use the below instead and revert to this above approach if we see any issues: 

const mfaSubmitButton = page.getByRole('button', { name: 'Submit' });

frontend/tests/playwright.config.ts
Bhavna-Ramachandran: Acknowledged. I have added the comments back.

frontend/tests/playwright.config.ts
Bhavna-Ramachandran: Acknowledged and I have made updates.

frontend/tests/playwright.config.ts
```diff
@@ -1,9 +1,32 @@
...
 
-dotenv.config({ path: path.resolve(__dirname, "..", ".env.local") });
+// Load environment variables from .env.local if it exists
+const envPath = path.resolve(__dirname, "..", ".env.local");
+if (fs.existsSync(envPath)) {
+  dotenv.config({ path: envPath });
+}
 
+// Determine environment: can be overridden via PLAYWRIGHT_TARGET_ENV
+const ENV = process.env.PLAYWRIGHT_TARGET_ENV || "local";
```
doug-s-nava, line 13: 👍

.github/workflows/ci-frontend-e2e.yml
```diff
@@ -23,17 +23,23 @@ concurrency:
 
 jobs:
   e2e-tests:
-    name: Run E2E Tests
+    name: Run Local E2E Tests
     runs-on: ubuntu-22.04
 
+    env:
+      PLAYWRIGHT_TARGET_ENV: local
+      LOCAL_BASE_URL: http://127.0.0.1:3000
```
doug-s-nava: since these match the defaults set in the playwright config I think you can delete them, but not a blocker

doug-s-nava (approved):
✅ ✅ ✅ ✅ 

thank you for bearing with my reviews!

.github/workflows/ci-frontend-e2e.yml
Bhavna-Ramachandran: Done

Issue #6229, opened 2025-09-02
Frontend Engineering Implement Group Saved Opportunities by Status


btabaska:
### Summary

Develop and ship the “Group Saved Opportunities by Status” feature on the frontend, based on final design mockups.

### Acceptance criteria

- [ ]  Code implementation aligns with final design.
- [ ]  Unit and integration tests written and passing.

PR #7415, opened 2025-12-10
[Issue #6229]: Implement Filter for Saved Opportunities

btabaska:
## Summary
Work for #6229 

## Changes proposed

- Implemented the ability to sort saved opportunities by status 
- Implemented unit tests for that new expanded functionality

## Context for reviewers

- This was a Fider board issue that the community selected last quad to be implemented this quad
- Since the saved opportunities page is old, and will need a revamp in the future to add new components, pagination, etc this is a lightest-lift implementation.
- Sort was implemented on the frontend using a hook, rather than the backend since this will likely be revamped in the not-far future.

### DIFF
--- a/frontend/src/constants/searchFilterOptions.ts
+++ b/frontend/src/constants/searchFilterOptions.ts
@@ -225,26 +225,10 @@ export const savedOpportunityStatusOptions: FilterOption[] = [
     label: "Any opportunity status",
     value: "",
   },
-  {
-    id: "saved-status-forecasted",
-    label: "Forecasted",
-    value: "forecasted",
-  },
-  {
-    id: "saved-status-open",
-    label: "Open",
-    value: "posted",
-  },
-  {
-    id: "saved-status-closed",
-    label: "Closed",
-    value: "closed",
-  },
-  {
-    id: "saved-status-archived",
-    label: "Archived",
-    value: "archived",
-  },
+  ...statusOptions.map((option) => ({
+    ...option,
+    id: `saved-${option.id}`,
+  })),
 ];
 
 export const sortOptions: FilterOption[] = [


frontend/src/constants/searchFilterOptions.ts
```diff
@@ -219,6 +219,34 @@ export const allFilterOptions = {
   fundingInstrument: fundingOptions,
 };
 
+export const savedOpportunityStatusOptions: FilterOption[] = [
```
doug-s-nava, line 222: is there a reason not to reuse the existing status options? If the stumbling block is applying all statuses, there's an example of how we're doing that with agencies here https://github.com/HHS/simpler-grants-gov/blob/f88cbfad7994cffa74fc2bd0722f0141f69218ee/frontend/src/utils/search/searchUtils.ts#L118

frontend/src/services/fetch/fetchers/savedOpportunityFetcher.ts
```diff
@@ -34,11 +34,21 @@ export const handleSavedOpportunity = async (
 export const getSavedOpportunities = async (
   token: string,
   userId: string,
+  statusFilter?: string,
 ): Promise<MinimalOpportunity[]> => {
   const ssgToken = {
     "X-SGG-Token": token,
   };
-  const body = {
+  const body: {
```
doug-s-nava, line 42: does `SearchRequestBody` work here? https://github.com/HHS/simpler-grants-gov/blob/f88cbfad7994cffa74fc2bd0722f0141f69218ee/frontend/src/types/search/searchRequestTypes.ts#L47

doug-s-nava (commented):
have not tested, a couple thoughts on the code

frontend/src/services/fetch/fetchers/savedOpportunityFetcher.ts
btabaska: No, SearchRequestBody won't work directly here. The issue is with the sort_order type is mismatched. The saved opportunities endpoint uses "created_at" as the order_by value, which is not included in PaginationOrderBy.

frontend/src/constants/searchFilterOptions.ts
btabaska: Just updated based on this feedback to reuse the existing status options.

frontend/src/services/fetch/fetchers/savedOpportunityFetcher.ts
doug-s-nava: Gotcha. This isn't a dealbreaker, but I wonder if:
* this explicit typing is necessary if we utilize something like the agency filter building setup, where we're always sending the sending the filter value, and just setting it to send all status values if the filter isn't being used
* we should take a look at our typing here and make it more generic, or build out the necessary static types for this use case

PR #7425, opened 2025-12-10
[Unticketed] Document how we should be removing columns from DB tables

chouinar:
## Changes proposed
* Add docs about how to delete a column from a DB table

## Context for reviewers
We had an issue with the prod deploy today. Because our migrations run before code is deployed, what happened is:
* Migration ran, deleting a DB column
* Code querying that table failed because the DB said "I have no idea what that column is"
* API code was updated
* Code stopped failing

For about 5 minutes any calls to our opportunity endpoints were 500ing. 

Unfortunately, the proper solution requires handling it a bit different, and a bit more annoyingly. We need to tell SQLAlchemy to ignore a column first, do the prod deploy THEN delete the column. Meaning any column deletes take two weeks to do. These docs (and the way of resolving the issue) are pulled from a prior project where we hit this issue.

## Validation steps
Admittedly, haven't verified this, but can't really test it without doing it. At the very least, this is based on hitting this issue several times on a prior project, and at worst, I'm slightly off on what we need to configure.

### DIFF
--- a/documentation/api/database/database-management.md
+++ b/documentation/api/database/database-management.md
@@ -133,3 +133,39 @@ a new migration. If there would be anything created, it errors.
 
 This check also runs as part of our CI/CD, so even if you forget to run it yourself during development,
 it will be caught when you send a pull request.
+
+# Operations
+## Removing a column
+
+**First Deploy**
+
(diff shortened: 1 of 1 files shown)


prasnava (approved):
👍

Issue #7429, opened 2025-12-11
Develop CI job to run e2e tests against deployed env


doug-s-nava:
### Summary

Prerequisite for https://github.com/HHS/simpler-grants-gov/issues/7257, this work entails:

* creating a new job in github actions that will support running our playwright test suite against a deployed environment
* fixing up (or potentially skipping for later work) any tests that break when running against a deployed environment

This does not include enabling the new job to run anywhere, that will be handled in #7257 

### Acceptance criteria

- [ ] new job exists in github actions which can run e2e suite against staging on demand
- [ ] new job does not run on any triggers
- [ ] all e2e tests pass with reasonable consistency or are skipped for later ticketed work when run against staging in new job

PR #7435, opened 2025-12-11
[Issue #7429] job for running e2e tests against staging

doug-s-nava:
## Summary

Work for #7429 

## Changes proposed

The goals of this PR are:

* create a job to run in Github actions that will run the e2e test suite targeted against the staging environment
* update the e2e test suite to support targeting the staging environment, including updating tests to allow them to pass in this scenario

## Context for reviewers

A number of caveats and notes on this work:

* the job created in this PR is somewhat temporary. In https://github.com/HHS/simpler-grants-gov/issues/7973 this job will be refined so that it works better with the existing ecosystem. As it stands now it can be run locally but will not run on any automation or triggers
* the target for e2e runs is now controlled by the environment variable "PLAYWRIGHT_TARGET_ENV", if this is not provided the run will default to running against local
* global and test specific timeouts are now configured based on the target environment, since tests running against staging sometimes took longer to resolve for reasons we were not able to determine at this point. We should revisit these timeouts from time to time to see if we can reduce them.
* this isolates all environment variable based constants for use in e2e tests to a new file and updates file imports as necessary

Some notes for future consideration regarding the use of the staging test user in this change, as there may be issues as we scale this approach with that approach not being sufficient:

* we will need to be able to test the site behavior when logging in as users with different privileges, and thus will need to support using multiple users
* each time the e2e system logs in with a test user, an email is sent to the email set up for that user. Is there a way to disable this? Or is this helpful?
* still trying to figure out flakiness where the mfa generation works on some tests and not others (the saved opportunities test is currently failing)

## Validation steps

1. _VERIFY_: staging (dynamic) and local (frontend e2e) tests all pass in CI

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,7 +1,7 @@
 name: Dynamic E2E Tests
 on:
   push:
-    branches: ["e2e-staging-run"]
+    branches: ["e2e-staging"]
 
   workflow_dispatch:
     inputs:


api/tests/lib/seed_local_db.py
```diff
@@ -69,6 +70,22 @@ def _build_opportunities(
             size=2, agency_code="ARCHIVED", is_archived_non_forecast_summary=True
         )
 
+        # hardcoded id for e2e usage
+        # Check if it already exists before creating to allow running seed multiple times
+        e2e_opportunity_id = uuid.UUID("6a483cd8-9169-418a-8dfb-60fa6e6f51e5")
```
doug-s-nava, line 74: I don't remember where we landed on with this. From the backend / API perspective is this alright, or do we need to refactor?

Bhavna-Ramachandran: per @chouinar suggestion here: https://github.com/HHS/simpler-grants-gov/pull/7688/files#r2662550372, we should use the existence check pattern to avoid integrity errors. Pasting the recommended code here for reference:

existing_e2e_opportunity_id = "6a483cd8-9169-418a-8dfb-60fa6e6f51e5"
existing_e2e_opportunity = db_session.get(Opportunity, existing_e2e_opportunity_id)
if existing_e2e_opportunity is None:
    factories.OpportunityFactory.create(
        has_long_descriptions=True, opportunity_id=existing_e2e_opportunity_id
    )

doug-s-nava: should I replace the code that's currently there with what you've pasted here?

chouinar: Yes, don't try to check for integrity errors, just do:

```py
e2e_opportunity_id = uuid.UUID("6a483cd8-9169-418a-8dfb-60fa6e6f51e5")
existing_e2e_opportunity = db_session.get(Opportunity, e2e_opportunity_id)

if not existing_e2e_opportunity:
    factories.OpportunityFactory.create(
        has_long_descriptions=True, opportunity_id=e2e_opportunity_id
    )
```

Bhavna-Ramachandran: @doug-s-nava Do you want to include this change as part of this PR or handle it separately?

frontend/tests/e2e/playwrightUtils.ts
```diff
@@ -44,12 +43,19 @@ export async function waitForURLContainsQueryParamValue(
...
-    await waitForURLChange(page, changeCheck, timeout);
+    await waitForURLChange(page, changeCheck);
   } catch (_e) {
     throw new Error(
       `Url did not change to contain ${queryParamName}:${queryParamValue} as expected`,
     );
   }
+
+  return expect(page.url()).toContain(`${queryParamName}=${queryParamValue}`);
+
```
Bhavna-Ramachandran: can we try changing this line to soemthing like below:

const url = new URL(page.url());
const params = new URLSearchParams(url.search);

return expect(params.get(queryParamName)).toBe(queryParamValue);

Rationale: Current line checks raw url string, above code checks the decoded query param. This should help reduce flakiness.

frontend/tests/e2e/search/search-error.spec.ts
```diff
@@ -20,6 +20,7 @@ test.describe("Search error page", () => {
 
     await page.waitForURL(/closed/, {
       timeout: 5000,
+      waitUntil: "domcontentloaded",
     });
```
Bhavna-Ramachandran, line 26: I saw some failures for this part. I suspect that the URL changes, but this does not trigger a full load event (causing waitUntil: "domcontentloaded" to timeout??). Can we try using this: await page.waitForURL(/closed/, { timeout: 15000 }).

Rationale: I have noticed that generous timeout helps with CI runs. If the URL updateds without a full page reload, then not using 'waitUntil: "domcontentloaded"' can help.

frontend/tests/e2e/search/searchSpecUtil.ts
```diff
@@ -4,11 +4,14 @@
```
Bhavna-Ramachandran, line 4: I am not able to add comment for the line# 270 - this currently has : await filterOptions.isVisible(); , and can be changed to : await filterButton.waitFor({ state: "visible", timeout: 15000 });

Rationale: isVisible() just checks once and returns true/false, does not wait for element to appear and the click fails when element isnt rendered yet.
Reference: https://playwright.dev/docs/api/class-locator#locator-isvisible

frontend/tests/e2e/search/searchSpecUtil.ts
```diff
@@ -4,11 +4,14 @@
 
 import { expect, Locator, Page } from "@playwright/test";
```
Bhavna-Ramachandran, line 5: I am not able to add comment on line# 268.Currently the 'waitForFilterOptions' attempts to click immediately after creating locator object. Can we make it wait for element to be actually rendered and visible before attempting the click. By replacing below lines(268,269):
  await filterButton.click();
  const filterOptions = page.locator(`input[name="${filterType}-*"]`);

  to something like:
  await filterButton.waitFor({ state: "visible", timeout: 15000 });
  await filterButton.click();
  const filterOptions = page.locator(`input[name="${filterType}-*"]`);

api/tests/lib/seed_local_db.py
doug-s-nava: added

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,144 @@
+name: Dynamic E2E Tests
+on:
+  push:
+    branches: ["e2e-staging-run"]
```
Bhavna-Ramachandran: This needs to be updated

Issue #7398, opened 2025-12-09
Scaling our pipeline to address increase in searches


ychoy:
### Description

Beginning Tuesday Dec 16, we’re expecting to see some increase in search when MH re-directs users to Simpler Grants. Most likely search volume will be fully increased by January. 

Updates need to be limited to Search. 

Ideally we deploy this by Wed, or do a Terraform apply to deploy by Friday. 

### Risks and assumptions

Search index performance will be slower if this doesn't happen by Dec 16th. 

### Acceptance Criteria
- [x] Search index has adequate resources

mdragon:
Opened two PRs for the two potential scaling approaches, up vs out. I'm favoring up right now but wanted us to have flexibility as we decide what to do and each is only a one line change.

mdragon:
Testing applying the scale up PR wasn't showing the image type change in the terraform diff. Will follow up with Sean and Prash when folks are back in the office.

mdragon:
We determined the instance type scaling up/down does work from `terraform apply` and the search instance responds to queries throughout the ~12 minute apply.

PR #7438, opened 2025-12-11
[Issue #7398] Scale OpenSearch out

mdragon:
## Summary

Work for #7398  

## Changes proposed
This is one of two scaling approaches we're preparing for scaling Search traffic.

This approach will scale our search nodes out by increasing the nodes from 3 to 6. This is the smallest increment we can make as we must adjust in multiples of our availability zone count which is 3.

We are reviewing the change, so we are ready, but not planning to apply/deploy the change yet.


prasnava (approved):
👍

mdragon:
We didn't fully test this path as scaling up seemed preferable

PR #7439, opened 2025-12-11
[Issue #7398] Scale OpenSearch up

mdragon:
## Summary

Work for #7398  

## Changes proposed
This is one of two scaling approaches we're preparing for scaling Search traffic.

This approach will scale our search nodes up by increasing the size of the node from medium to large.

We are reviewing the change, so we are ready, but not planning to apply/deploy the change yet.


prasnava (approved):
👍

mdragon:
We tested applying this approach in Dev and it worked and search was accessible during the full up scaling and down scaling processes. Not merging this now as we don't need to scale Prod but this would be a valid approach in the future

Issue #6146, opened 2025-08-28
Audit and Create Plan for priority E2E tests needed for Simpler Apply workflow


chris-kuryak:
### Summary

To identify what is needed for E2E tests in the Simpler Apply workflow, we want to audit the existing workflow and develop a plan of priority E2E tests needed. We may not get a team of folks to do rigorous testing of forms and workflows in this quad, so we want to create some essential tests that can be used to catch risky errors.

Goal: What are the E2E tests (or other integrations) we should focus on building in Quad 4 that significantly reduces of our risk of breaking the Apply workflow unintentionally?

In evaluating user interactions, build test cases for each scenario and rank them by their importance to the proper functioning of the apply functionality. Test cases can be stored, for now, in google drive document or spreadsheet until we have a better place to put them

### Acceptance criteria

- [ ] Primary apply related user flows are documented as test cases and ranked in order of priority for testing

chris-kuryak:
From Aaron: 
- Forms
- Login state
- Where can we test certain things

chris-kuryak:
From My:
- Should we grow into Storybook to use "Chromium" to do a visual check before deploy?

aoysimmons-GH:
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/AOYAO_~1/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/AOYAO_~1/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">

</head>

<body link="#467886" vlink="#96607D">


Scenario   Name | Description | Risk Level to have for work-flow| Notes
-- | -- | -- | --
Start   Application | User selects   "Start Application" and the Apply shell loads (layout, header,   footer, navigation, or data context) | Critical | Main entry point
Form   Navigation | User can go Next/Back   across all steps without data loss | Critical | Most common break
Fields   Required Validation | Empty required fields   block progress and show errors | Critical | Core safety net
Cross Form   Required Validation | Empty required fields   block to submit and show errors | Critical | Core safety net
[...]

PR #7456, opened 2025-12-12
Add happy_path_apply feature file for Apply workflow

Bhavna-Ramachandran:
**Summary**
Work for : Issue #[6146 ](https://github.com/HHS/simpler-grants-gov/issues/6146)

**Changes proposed**
Created Folder structure: frontend/tests/e2e/apply/features for adding all feature files
Created draft happy path scenario for submitting an application
Created frontend/tests/e2e/apply/features/happy_path_apply.feature

**Context for reviewers**
This is the first BDD feature file for the Apply workflow.
Reviewers should verify if all critical user flows are covered.
The happy path will later be automated using Playwright for E2E regression tests.

**Validation steps**
Confirm the feature file exists in: frontend/tests/e2e/apply/features/happy_path_apply.feature
Review the Gherkin scenarios to ensure the happy path covers all critical user flows.
Verify the folder structure allows for future feature files to be added consistently.

### DIFF
--- a/frontend/tests/e2e/apply/features/happy_path_apply.feature
+++ b/frontend/tests/e2e/apply/features/happy_path_apply.feature
@@ -128,8 +128,7 @@ Then all required and optional forms have no validation errors
 
 # --- Attachment Uploads From the Attachment Section ---
 When the user uploads all required documents under the "Attachment" header
-Then each file meets size and format requirements
-And the documents are successfully attached
+Then the documents are successfully attached
 And the uploaded files appear in the "Attached documents" table
 
 # --- Submitting the Application ---


frontend/tests/e2e/apply/features/happy_path_apply.feature
```diff
@@ -0,0 +1,149 @@
...
+    And the system is available
+
+  @Happy-Path
+  Scenario: Complete the Apply workflow for an Organization user
+    
+    # --- Login ---
+    Given the user is logged in
+
+    # --- Search the Opportunity ---
+    When the user searches for funding opportunity
```
doug-s-nava: do we need to specify that the user searches for and finds an opportunity with an open competition that can be applied to?

frontend/tests/e2e/apply/features/happy_path_apply.feature
```diff
@@ -0,0 +1,149 @@
...
+    Given the user has completed the Required forms
+
+    # ---- Completing Conditionally Required Forms ---  
+    Given the user has completed the conditionally Required forms
+
+    # --- Confirm the Optional Form for Submission ---
+    Given the user has confirmed the optional Form for Submission
+
+    # --- Validate All Required and Optional Forms ---
+    Given the user completes validation of Required and Optional Forms
```
doug-s-nava: How is this different from the step in line 29? Is it that in line 29 the user has filled in the required fields on all required forms, and that on line 38 any further validation problems have been resolved? Is it possible these two could be combined?

frontend/tests/e2e/apply/features/happy_path_apply.feature
```diff
@@ -0,0 +1,149 @@
...
+
+    # --- Complete Additional Attachments ---
+    Given the user completes the additional attachments
```
doug-s-nava: I'm not aware of any forms that will require attachments which are not part of a form. In any case, we don't have a way of enforcing that and it's something that hasn't really come up. We may add something like this in the future, but I think that for now it's ok to leave this step out

frontend/tests/e2e/apply/features/happy_path_apply.feature
```diff
@@ -0,0 +1,149 @@
...
+When the user clicks the Funding Opportunity number
+Then the user is taken to the Funding Opportunity details page
+
+# ---- Completing Forms (Required or Conditionally Required) ---
+Given there are <form_type> forms in the application forms list
+When the user clicks on a <form_type> form
+Then the <form_type> form opens
+When the user completes all required fields, including uploading a file in any available attachment field if present
+And the user clicks "choose from folder" or drags and drops the file
+Then the file meets size and format requirements
```
doug-s-nava: despite what the directions on the site may say, we're not enforcing any size or format requirements on files at this point

frontend/tests/e2e/apply/features/happy_path_apply.feature
```diff
@@ -0,0 +1,149 @@
...
+Then the user scrolls to the search results table
+When the user clicks the Funding Opportunity number
+Then the user is taken to the Funding Opportunity details page
+
+# ---- Completing Forms (Required or Conditionally Required) ---
+Given there are <form_type> forms in the application forms list
+When the user clicks on a <form_type> form
+Then the <form_type> form opens
+When the user completes all required fields, including uploading a file in any available attachment field if present
+And the user clicks "choose from folder" or drags and drops the file
```
doug-s-nava, line 107: should we split the file upload portion of this into its own "statement"? (sorry, not clear on the terminology with feature files yet)

frontend/tests/e2e/apply/features/happy_path_apply.feature
```diff
@@ -0,0 +1,149 @@
...
+And the system saves the <form_type> form
+When the user clicks on the Application Name link
+Then the system navigates back to the Application landing page
+
+# --- Confirm the Optional Form for Submission ---
+When the user scrolls down to the "Conditionally-Required forms" section
+And under "Submit with application" the user selects "Yes" for the completed optional form
+Then no validation errors should appear
+And remaining optional forms have "No" selected
+And the user is taken to the next step
```
doug-s-nava: not sure what the next step is referring to here

frontend/tests/e2e/apply/features/happy_path_apply.feature
```diff
@@ -0,0 +1,149 @@
...
+
+# --- Validate All Required and Optional Forms ---
+When the user completes all required and optional forms
```
doug-s-nava, line 126: Since the optional forms aren't required for a successful submission, the minimum viable happy path might technically not include completing the optional forms?

frontend/tests/e2e/apply/features/happy_path_apply.feature
Bhavna-Ramachandran: From the Search page, I see that 'Posted' opportunities are published and open for application. So specifying an additional 'open competition' status shouldn’t be necessary - we can just ensure the scenario selects a funding opportunity with status Posted (Sample text: When the user goes to the search page; And the user selects a funding opportunity with status "Posted"). Please confirm if I can go ahead and make that change.

If the feature will later be expanded to include statuses between “Posted” and “Open” or “Accepting Applications,” we can then add clarification about the “open competition” status. Please let me know.

frontend/tests/e2e/apply/features/happy_path_apply.feature
Bhavna-Ramachandran: Line 29 covers completing required forms on the forms page, including field entry, attachments, and validations. Line 38 happens on the Application landing page after all forms are completed and the optional form for submission has been confirmed. Line 38 ensure that the application passes validation. While it may look like a duplicate, this step is included as a final check to confirm everything is correct.

frontend/tests/e2e/apply/features/happy_path_apply.feature
Bhavna-Ramachandran: Thanks for the clarification. I understand that currently this is not a requirement.
However, the Attachments section on the Application landing page states: "if opportunity instructions require documentation not covered by one of the forms above, upload files here..." Since this feature allows users to upload extra forms, I thought it would be useful to include this step in the E2E test as a check. Please confirm whether we should remove this step, or if it’s okay to keep it in the test.

frontend/tests/e2e/apply/features/happy_path_apply.feature
Bhavna-Ramachandran: Acknowledged. Are we considering adding file size limits to prevent issues with large uploads and format restrictions to ensure compatibility and reduce the risk of unsupported or potentially harmful files?

I’m was planning to use the EICAR test file in negative test cases to simulate virus detection check (format restrictions), should I pause that plan?

frontend/tests/e2e/apply/features/happy_path_apply.feature
Bhavna-Ramachandran: Sure. Thanks for the suggestion. It would make it easier to add future validations like file size, format, or virus checks where needed, without updating the whole block. 

The shared workflow would read like below. Please let me know your thoughts on this.

# ---- Completing Forms (Required or Conditionally Required) ---
Given there are <form_type> forms in the application forms list
When the user clicks on a <form_type> form
Then the <form_type> form opens
And the user completes all required fields

# --- Uploading Attachments (if present) ---
When the user uploads a file in any available attachment field
And the user clicks "choose from folder" or drags and drops the file
Then the file is successfully attached and appears in the appropriate attachment table
Then the file meets size and format requirements

# --- Saving the Form ---
And the user clicks "Save"
Then no validation errors should appear
And the system saves the <form_type> form

# --- Returning to Application Landing Page ---
When the user clicks on the Application Name link
Then the system navigates back to the Application landing page

frontend/tests/e2e/apply/features/happy_path_apply.feature
Bhavna-Ramachandran: Acknowledged. I will remove this line : 'And the user is taken to the next step' . It doesn't reflect the system behavior, I just checked and confirmed that everything depends on user actions.

frontend/tests/e2e/apply/features/happy_path_apply.feature
Bhavna-Ramachandran: You are right. Technically the happy path could exclude completing the optional forms, since they aren’t required for a successful submission.
However, including them in the scenario helps ensure that the system handles optional forms correctly and identify potential validation issues. Please confirm whether it is ok to keep this step in the test or should we remove it

frontend/tests/e2e/apply/features/happy_path_apply.feature
doug-s-nava: That all makes sense! At the moment not all "open" opportunities can be applied to because our application functionality is still in beta, but as we go forward it will be true that any "open" opportunity should be "open for application on simpler grants"

It would be clearer to me if this read "When the user searches for an open funding opportunity" but understand if you feel that's not necessary

frontend/tests/e2e/apply/features/happy_path_apply.feature
doug-s-nava: Got it. Is it considered a bad practice to write a statement like this that is more "state" based rather than "action" based? Such as:

`Given the required and optional forms contain no validation errors`?

frontend/tests/e2e/apply/features/happy_path_apply.feature
doug-s-nava: I'm fine with leaving it in, as it's important to make sure the functionality works regardless of whether we're actually able to "require" additional attachments.

Do you mind rewording?
```suggestion
    Given the user uploads additional attachments
```

frontend/tests/e2e/apply/features/happy_path_apply.feature
doug-s-nava: that's at least a good start. 

Unless I'm misunderstanding your intent, we may not want to include the part about validation errors after saving the form, since most forms will have required fields other than the attachment upload that we would expect to error in this case. That case could be more about the correct file name appearing in the form if you refresh the page after saving

frontend/tests/e2e/apply/features/happy_path_apply.feature
doug-s-nava: The frontend is not enforcing anything in this regard, but I know there was work going on in relation to virus scanning. @chouinar - any thoughts here?

frontend/tests/e2e/apply/features/happy_path_apply.feature
doug-s-nava: I'm good with leaving it here, just wanted to make sure my definition of "happy path" was in line with yours

doug-s-nava (commented):
a couple more thoughts and requests. I'd be fine with merging this as is, but can you take one more pass - interested to see what this looks like with the file upload piece split out

frontend/tests/e2e/apply/features/happy_path_apply.feature
Bhavna-Ramachandran: I see your point. I will update this to say, 'When the user searches for an open funding opportunity.'

frontend/tests/e2e/apply/features/happy_path_apply.feature
Bhavna-Ramachandran: Thank you. I will update this to say 'Given the required and optional forms contain no validation errors'

PR #7467, opened 2025-12-15
[Issue #7470]: Update package versions w/ precompiled wheels for 3.14 

prasnava:
## Summary

The docker build for the api is taking too long because we have to compile / build some of the wheels locally. Upgraded the below deps which include prebuilt wheels. 

```
xmltodict = "^1.0.0"
lxml = "^6.0.0"
```

## Validation steps

The api should continue to work as is.


chouinar (approved):
LGTM - had some engineers take a look at the packages who were more familiar with the XML work and check whether it would cause any issues upgrading these, seems like we're fine.

Issue #7317, opened 2025-12-02
Improve Pagination Schema Validation to Return 422 Instead of 500


joshtonava:
### Summary

When clients send malformed `sort_order` parameters (e.g., passing a string instead of an object/list), the API returns a 500 Internal Server Error instead of a proper 4xx validation error.

### Example Error

```
AttributeError: 'str' object has no attribute 'get'
  File "/api/.venv/lib/python3.13/site-packages/src/pagination/pagination_schema.py", line 13, in before_load
    if item.get("sort_order") is not None:
        ^^^^^^^^
```

This occurs in the `BasePaginationSchema.before_load()` method when it attempts to call `.get()` on what it expects to be a dictionary, but receives a string instead.

## Root Cause

The `@pre_load` hook in `BasePaginationSchema.before_load()` assumes the incoming `item` parameter is always a dictionary. However, when Marshmallow encounters malformed input (like `sort_order="invalid_string"` instead of `sort_order=[{...}]`), the type coercion fails and passes the raw string value to the pre_load hook before proper validation occurs.

**Location**: `api/src/pagination/pagination_schema.py:10-22`

### Acceptance criteria

- [ ] When `sort_order` is passed as a string, API returns 422 with clear error message
- [ ] When `sort_order` is passed as invalid JSON, API returns 422 with clear error message
- [ ] When any pagination parameter has wrong type, API returns 422 (not 500)
- [ ] Existing valid pagination requests continue to work
- [ ] Add test cases covering these malformed input scenarios

PR #7468, opened 2025-12-15
[Issue #7317] Schema validation on pagination

babebe:
## Summary

Fixes / Work for #7317 

## Changes proposed

- Updated BasePaginationSchema @pre_load hook to validate that input is a dict before processing.
- Added proper validation for pagination parameters to raise ValidationError when invalid.
- API now returns 422 for:
   - Wrong types for pagination fields (e.g., negative page, string page size)
   - Missing required pagination fields
   - Malformed JSON in pagination payload
- Added/updated tests for all invalid pagination scenarios, ensuring required fields are included to trigger type validation.

## Context for reviewers
Previously, malformed pagination input could cause 500 errors due to assumptions in pre_load.

## Validation steps

Send valid pagination requests → should succeed.
Send wrong types for page_offset, page_size → 422.
Omit required pagination fields → 422.
Send malformed JSON → 422.

### DIFF
--- a/.github/workflows/update-frontend-feature-flag.yml
+++ b/.github/workflows/update-frontend-feature-flag.yml
@@ -1,0 +1,132 @@
+name: Update Frontend Feature Flag
+run-name: Update ${{ inputs.feature_flag }} to ${{ inputs.value }} in ${{ inputs.environment }}
+
+on:
+  workflow_dispatch:
+    inputs:
+      feature_flag:
+        description: Feature flag to update
+        required: true
+        type: choice
+        options:
+          - feature-search-off
+          - feature-opportunity-off
+          - feature-auth-on
+          - feature-saved-opportunities-on
+          - feature-saved-searches-on
+          - feature-apply-form-prototype-off
+          - feature-search-drawer-on
+          - feature-search-table-on
+          - feature-developer-page-off
+          - feature-user-admin-off
+          - feature-manage-users-off
+      environment:
+        description: Target environment
+        required: true
+        type: choice
+        options:
+          - dev
+          - staging
+      value:
+        description: Feature flag value
+        required: true
+        type: choice
+        options:
+          - "true"
+          - "false"
+
+jobs:
+  update-feature-flag:
+    name: Update Feature Flag
+    runs-on: ubuntu-22.04
+    permissions:
+      contents: read
+      id-token: write
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Terraform
+        uses: ./.github/actions/setup-terraform
+
+      - name: Configure AWS credentials
+        uses: ./.github/actions/configure-aws-credentials
+        with:
+          app_name: frontend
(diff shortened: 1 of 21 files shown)


api/src/pagination/pagination_schema.py
```diff
@@ -8,6 +9,10 @@ class BasePaginationSchema(Schema):
 
     @pre_load
     def before_load(self, item: dict, many: bool, **kwargs: dict) -> dict:
+        # Schema-level error: payload is not an object
+        if not isinstance(item, dict):
+            raise ValidationError("Invalid pagination parameters: expected an object.", "_schema")
+
```
chouinar: I don't think we'd want to raise a Pydantic validation error in Marshmallow code.

If someone passes in an invalid type, what if we just did:
```py
if not isinstance(item, dict):
     return item
```
Marshmallow itself should do its validation check after this that complains about the type being incorrect. We don't care that the below code doesn't run since it is invalid anyways.

babebe: Updated the hook to remove raising a Pydantic validation error. Now it simply returns the data, for Marshmallow to validate.

api/tests/src/pagination/test_paginator.py
```diff
@@ -149,3 +151,60 @@ def test_paginator(db_session, create_opportunities):
...
+
+def test_sort_order_string_returns_422(client):
+    # Sending sort_order as a string instead of list/dict
+    payload = {"pagination": {"sort_order": "invalid_string"}}
+    response = client.post(f"/v1/organizations/{uuid.uuid4()}/legacy-users", json=payload)
+    assert response.status_code == 422
+
+    data = response.get_json()
+    assert "errors" in data
+    assert any("sort_order" in str(err) for err in data["errors"])
```
chouinar: Nit: I think the error you caused in this would be the only error, could you check the `type` of the error object, I can't recall what Marshmallow defaults to for bad type, but should be something consistent.

Just want to make sure it's failing in the way we expect, and not in some other way.

babebe: Updated test, asserting on the presence of ‘valid’ in the error message to confirm a schema-level type validation error.

chouinar: I'm looking for an approach (on all these new tests) that does not do `str(err)` in it. The error object is a well-formatted object with specific fields in it. All this current check does is say "sort_order is somewhere in some part of the error message", is it the field? Is it in the error message?

I'm looking for us to check that the type of the error is properly checked for, that is, we do something like:
```py
assert {"type": "whatever", "message": "something", "field": "x"} in response.get_json()["errors"]
```

Now, I don't think doing a contains check is necessarily the right way, but looking for us to verify that the `type` field is set in an expected manner. If there are other errors, check for them too.

babebe: Ah, that makes sense. Updated accordingly.

Issue #7058, opened 2025-11-14
Handle merging nested allOfs with multiple items on frontend


doug-s-nava:
### Summary

The library that we're using to manage merging "allOf" references in the form json schema definition, in order to allow for easier parsing, seemingly is not able to handle situations where:
* the allOf is defined within a property definition AND
* the allOf contains multiple items
* ex https://github.com/HHS/simpler-grants-gov/pull/7056/files#diff-fb95c16498447c8cef4c653f086f75b126254e9fc79a774835ab502ceeed60ecR282

When trying to generate a form defined this way on the frontend we hit an error that looks like "no resolver for key if".

Note that configurations that are either not nested in a property (such as ) or that contain only a single item (such as ) work properly already.

see https://betagrantsgov.slack.com/archives/C05TSL64VUH/p1763132223940399

note that issue was first encountered in this work, testing can proceed against the form defined here https://github.com/HHS/simpler-grants-gov/pull/7056

### Technical approach

Relevant code is here, dealing with this library
* https://github.com/HHS/simpler-grants-gov/blob/eb8009123159f82647cfbdf0e0ce1faa54d30f4a/frontend/src/components/applyForm/utils.ts#L784
* https://github.com/mokkabonna/json-schema-merge-allof

We may be able to add a custom resolver to the call here to either get the library to somehow properly merge the allOf or, more likely, to ignore it.

Another option would be to prune out the allOf entirely, but that would prevent us from properly handling pre-emptive recognition of these conditional validation rules in the future.

If we end up leaving the allOf in place, there may be functions that break when they see the "allOf" here, we may need to adjust those functions accordingly to be able to ignore the "allOf" as well

### Acceptance criteria

- [ ] Supplementary NEH Cover Sheet Form loads in UI without errors
- [ ] Lightweight regression testing across forms (Product can help)

PR #7482, opened 2025-12-15
[Issue #7058] allow frontend to process schema with complex nested conditionals

doug-s-nava:
## Summary

Work for #7058 

## Changes proposed

The [mergeAllOf](https://github.com/mokkabonna/json-schema-merge-allof) library the frontend uses to manage processing "allOf" nodes in the form Json schema is not able to deal with nodes that represent complex conditional logic. As a result, forms that rely on this logic will not load in the frontend. (See the NEH cover sheet form here as an example https://github.com/HHS/simpler-grants-gov/pull/7056/files#diff-fb95c16498447c8cef4c653f086f75b126254e9fc79a774835ab502ceeed60ecR282)

This change removes any relevant allOf nodes from the form json schema, and returns them as a separate list of conditional validations to be referenced later on (though this functionality is currently unused).

## Context for reviewers

Looked into working around this issue using a custom resolver passed to "mergeAllOf" but hit a snag due to the library's inability to correctly track the path of the nodes it is processing.

## Validation steps

2. use the quick login dropdown to log in as "many_app_user"
3. click the user drop down and click "test application"
4. click the "NEH cover sheet" form
5. _VERIFY_: form loads
1. Bonus regression testing: make sure all other forms still load and work as expected

### DIFF
--- a/frontend/tests/pages/applications/application/form/actions.test.ts
+++ b/frontend/tests/pages/applications/application/form/actions.test.ts
@@ -1,7 +1,6 @@
-import { getFormDetails } from "src/services/fetch/fetchers/formsFetcher";
+import { RJSFSchema } from "@rjsf/utils";
 
 import { handleFormAction } from "src/components/applyForm/actions";
-import { shapeFormData } from "src/components/applyForm/utils";
 
 const mockGetSession = jest.fn();
 const mockProcessFormSchema = jest.fn();
@@ -9,16 +8,18 @@ const mockShapeFormData = jest.fn();
 const mockGetFormDetails = jest.fn();
 
 jest.mock("src/services/auth/session", () => ({
-  getSession: () => mockGetSession(),
+  getSession: () => mockGetSession() as unknown,
 }));
 
 jest.mock("src/components/applyForm/utils", () => ({
-  processFormSchema: (schema) => mockProcessFormSchema(schema),
-  shapeFormData: (data, schema) => mockShapeFormData(data, schema),
+  processFormSchema: (schema: RJSFSchema) =>
+    mockProcessFormSchema(schema) as unknown,
+  shapeFormData: (data: FormData, schema: RJSFSchema) =>
+    mockShapeFormData(data, schema) as unknown,
 }));
 
 jest.mock("src/services/fetch/fetchers/formsFetcher", () => ({
-  getFormDetails: (id) => mockGetFormDetails(id),
+  getFormDetails: (id: string) => mockGetFormDetails(id) as unknown,
 }));
 
 const genericPayload = {
@@ -32,7 +33,7 @@ const genericPayload = {
 describe("handleFormAction", () => {
   beforeEach(() => {
     mockGetSession.mockResolvedValue({ token: "a token" });
-    mockProcessFormSchema.mockImplementation((schema) => {
+    mockProcessFormSchema.mockImplementation((schema: RJSFSchema) => {
       return Promise.resolve({
         formSchema: schema,
       });


frontend/src/utils/generalUtils.ts
```diff
@@ -129,3 +129,15 @@ export function isExternalLink(href: string): boolean {
 export const isSubset = <T>(subset: T[], superset: T[]) => {
   return !difference(subset, superset).length;
 };
+
+export const isBasicallyAnObject = (mightBeAnObject: unknown): boolean => {
```
doug-s-nava, line 133: moved from applyForm utils so it can be reused

mdragon:
Tested this successfully locally. Just noting to get the user drop down working I had to do a make build in the API folder and then restart the services as just starting my existing API container didn't seem to have the right endpoints available.

frontend/src/components/applyForm/actions.ts
```diff
@@ -115,11 +116,11 @@ async function getFormSchema(formId: string): Promise<RJSFSchema | undefined> {
...
-  let formSchema = {};
   try {
-    formSchema = await processFormSchema(formDetail.form_json_schema);
+    const { formSchema } = await processFormSchema(formDetail.form_json_schema);
+    return formSchema;
   } catch (e) {
     console.error("Error parsing JSON schema", e);
   }
-  return formSchema;
+  return {};
```
mdragon: Will these console.errors be captured in NewRelic? Should this be throwing not returning an empty object? Maybe that's not possible in the short term the way the calling code is written but is that something we should open a ticket for to improve ability to discover issues and troubleshoot them?

doug-s-nava: * Since this is an actions file it will be run on the server, so the error logs will be captured in NR.
*  As for what this should be returning, I think you're right. There is code here that seems to expect a faulty value if something goes wrong, rather than an empty object. Could argue that's out of scope of this PR to fix since that's the existing behavior, but I'll go ahead and fix it https://github.com/HHS/simpler-grants-gov/blob/914a38ab51bbc75ca08b4ae3868889337cd1fdc6/frontend/src/components/applyForm/actions.ts#L37

doug-s-nava: also realizing there are not tests for this action, so I'll write something up there too

mdragon: Great that it's logging to NR. What else would we need besides the error message to know what form, for what opportunity/competition, for what user, for what application, etc to be able to act on this if we saw it in the logs (ie not while you're testing a form directly locally)? Can we make sure that's logging as structured JSON in addition to the error message?

doug-s-nava: for this process in particular, the user / application will not be especially relevant, since the only input is the form schema itself. Any error here would be present on all usages of a form. That said, it would be good to have a better sense of error handling within the forms / application process. To really do this right would depend on refactoring a number of parts of the system, though, so I don't think it's particularly in scope. For now I'll make sure the form name is included in the logs and we can revisit the rest?

mdragon: I was thinking being able to know who applicants and what agencies were affected if we were seeing this in Prod. Good point that if it's broken for one it's broken for all, but in terms of outreach/reporting it could be valuable to have these available to us. That being said if we don't have that context right now, we don't have it, so we can improve as we go.

doug-s-nava: Fair point that we should be logging that info wherever we can. Bigger question - how do we go about getting something like that kind of work scheduled?

Issue #7452, opened 2025-12-12
application attachments not working in deployed envs


doug-s-nava:
### Summary

requests are not hitting the python api

next js logging is not capturing relevant data (generic error message, erroneous 500 status code)

### Reproducibility

None

### How to reproduce the bug

reproducible in training, staging, dev

- open an application
- upload a file on the application page
- endless spinner, timeout

- open a form with an attachment field such as sf424
- upload an attachment
- 500

### Browser

None

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

doug-s-nava:
issue is that the fetch function here is not using the fetchers system, and as a result not sending the X-API-KEY header in order to get through the API gateway

To fix, update the code here to use a fetcher function or pass the X-API-KEY header directly https://github.com/HHS/simpler-grants-gov/blob/05192430e2b97961f48e1d53c41dbc2784d623a9/frontend/src/services/fetch/fetchers/applicationFetcher.ts#L203

ErinPattisonNava:
@mdragon Did you say you were popping up a fix for this, or is it something our team should take?  cc @chris-kuryak

doug-s-nava:
initial issue was that the attachments upload function was not setting the X-API-KEY header to get through the API gateway. now that we're sending the token, locally it looks like the backend is sending back a 422. Not sure why that's happening, but in any case we're swallowing that as well, due to a config option for application endpoint fetchers that is allowing 422 errors to pass through - something related to application submit that is having unintended effects elsewhere

mdragon:
@chris-kuryak this is headed into Staging for UAT

mdragon:
I confirmed uploads are working again in Staging.

chris-kuryak:
UAT passed

PR #7486, opened 2025-12-16
[Issue #7452] Attachments failing with API Gateway

mdragon:
## Summary

Work for #7452

## Changes proposed

API Key was missing on request causing API Gateway to reject the request. That rejection was getting eaten and no details about what was actually happening was getting logged or returned to the FE. The requests were never reaching the Python API as a result.

## Context for reviewers

After discovering the initial cause, we made a few additional improvements to how the Next API layer was calling the Python API in this case.

## Validation steps 
(This testing has already been validated locally on the branch by Doug and Matt, and in Dev with this branch deployed by Matt. So this list is purely for UAT, reviewers don't have to take these steps locally on this branch)

1. Log in
1. Go to an existing Application or start a new Application
1. Upload a file to the Attachments section at the bottom of the Application page
1. Validate that the file uploads successfully (doesn't get stuck on the waiting spinner)
1. Go into a form that requires an attachment, like the budget narrative attachement form and upload a file there.
1. Validate that the file uploads successfully (doesn't get stuck on the waiting spinner)
1. Save the form
1. Refresh both pages to validate the files still show up as appropriate

### DIFF
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -161,7 +161,7 @@ const nextConfig = {
     },
   },
   basePath,
-  reactStrictMode: false,
+  reactStrictMode: true,
   // Output only the necessary files for a deployment, excluding irrelevant node_modules
   // https://nextjs.org/docs/app/api-reference/next-config-js/output
   output: "standalone",
--- a/frontend/src/services/fetch/endpointConfigs.ts
+++ b/frontend/src/services/fetch/endpointConfigs.ts
@@ -6,7 +6,6 @@ export interface EndpointConfig {
   version: string;
   namespace: string;
   method: ApiMethod;
-  allowedErrorStatuses?: number[];
 }
 
 export const opportunitySearchEndpoint = {
--- a/frontend/src/services/fetch/fetchers/applicationFetcher.ts
+++ b/frontend/src/services/fetch/fetchers/applicationFetcher.ts
@@ -50,6 +50,7 @@ export const handleSubmitApplication = async (
   const response = await fetchApplicationWithMethod("POST")({
     subPath: `${applicationId}/submit`,
     additionalHeaders: ssgToken,
+    // want to allow responses with failed validations through so we can properly handle displaying validation errors
     allowedErrorStatuses: [422],
   });
 
@@ -189,10 +190,8 @@ export const uploadAttachment = async (
   const additionalHeaders = {
     "X-SGG-Token": token,
     Accept: "application/json",
-    // "Content-Type": "multipart/form-data",
   };
 
-  // this is throwing a 422, and not sure why. Maybe an issue with content type headers?
   const response = await fetchApplicationWithMethod("POST")({
     subPath: `${applicationId}/attachments`,
     additionalHeaders,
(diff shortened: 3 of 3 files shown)


frontend/next.config.js
```diff
@@ -161,7 +161,7 @@ const nextConfig = {
     },
   },
   basePath,
-  reactStrictMode: true,
+  reactStrictMode: false,
```
doug-s-nava: my bad - I left this in from my local testing, should be reverted

frontend/package.json
```diff
@@ -32,7 +32,7 @@
   "dependencies": {
```
doug-s-nava, line 32: package changes likely got pulled in from the other PR? probably shouldn't be in here

frontend/src/services/fetch/endpointConfigs.ts
```diff
@@ -2,11 +2,11 @@ import { environment } from "src/constants/environments";
 import { ApiMethod } from "src/services/fetch/fetcherHelpers";
 
 export interface EndpointConfig {
-  allowedErrorStatuses?: number[];
   basePath: string;
   version: string;
   namespace: string;
   method: ApiMethod;
+  allowedErrorStatuses?: number[];
```
doug-s-nava: not sure how this made it back in, should remove from this interface

frontend/src/services/fetch/fetchers/applicationFetcher.ts
```diff
@@ -188,23 +187,41 @@ export const uploadAttachment = async (
   file: FormData,
 ): Promise<ApplicationAttachmentUploadResponse> => {
   const additionalHeaders = {
-    Accept: "application/json",
     "X-SGG-Token": token,
+    Accept: "application/json",
+    // "Content-Type": "multipart/form-data",
```
doug-s-nava: can remove this

frontend/src/services/fetch/fetchers/applicationFetcher.ts
```diff
@@ -188,23 +187,41 @@ export const uploadAttachment = async (
...
-    `${environment.API_URL}`,
-    "alpha",
-    "applications",
-    `${applicationId}/attachments`,
-  );
-
-  const response = await fetch(url, {
-    method: "POST",
-    headers: additionalHeaders,
+  // this is throwing a 422, and not sure why. Maybe an issue with content type headers?
```
doug-s-nava: I figured out why! we can remove the comment

frontend/src/services/fetch/fetchers/applicationFetcher.ts
```diff
@@ -52,6 +50,7 @@ export const handleSubmitApplication = async (
   const response = await fetchApplicationWithMethod("POST")({
     subPath: `${applicationId}/submit`,
     additionalHeaders: ssgToken,
+    allowedErrorStatuses: [422],
```
doug-s-nava, line 54: probably good to add a comment here about why this is happening
```suggestion
    // want to allow responses with failed validations through so we can properly handle displaying validation errors
    allowedErrorStatuses: [422],
```

frontend/src/services/fetch/fetchers/applicationFetcher.ts
```diff
@@ -188,23 +187,41 @@ export const uploadAttachment = async (
...
+  // this is throwing a 422, and not sure why. Maybe an issue with content type headers?
+  const response = await fetchApplicationWithMethod("POST")({
+    subPath: `${applicationId}/attachments`,
+    additionalHeaders,
     body: file,
+    addContentType: false,
   });
 
   return (await response.json()) as ApplicationAttachmentUploadResponse;
+  // const additionalHeaders = {
```
doug-s-nava: oops

frontend/package.json
mdragon: Yeah they won't really apply now as that's in main. Maybe a rebase will pick that up?

frontend/src/services/fetch/endpointConfigs.ts
mdragon: Oh you know what this was failing because you removed it but it was still being passed that way in the tests. I thought you had changed it and it was working without it but failing the type checks.

Issue #7048, opened 2025-11-13
Record Organization admin actions


joshtonava:
### Summary

When performing an action as an organization admin, we want to track the action that was done using the 
 [Activity History](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2037416076/Tech+Spec+Activity+History) functionality.

Related: https://github.com/HHS/simpler-grants-gov/pull/7034/files
Application audit tracking is currently implemented with 13 event types covering application lifecycle events

Questions:
- do we want a central place in confluence where we can share what events are being tracked and with what info? 

### Acceptance criteria

- Track when a user accepts an invitation and joins the organization
- Track when an admin changes a user's roles in the organization
- Track when an admin removes a user from the organization 
- Unit tests updated / created

PR #7498, opened 2025-12-17
[Issue #7048] organization admin audit [1/2]

babebe:
## Summary

Fixes / Work for #7048  

## Changes proposed

Model created for storing Organization audit history
Added factory
Migration File


chouinar:
@babebe Heads up, another PR just merged a DB migration, you'll need to resync from main

api/src/constants/lookup_constants.py
```diff
@@ -290,3 +290,9 @@ class UserType(StrEnum):
     STANDARD = "standard"
     INTERNAL_FRONTEND = "internal_frontend"
     LEGACY_CERTIFICATE = "legacy_certificate"
+
+
+class OrganizationAuditEvent(StrEnum):
+    USER_JOINED = "user_joined"
+    USER_ROLE_UPDATED = "user_role_updated"
+    USER_REMOVED = "user_removed"
```
chouinar, line 298: Do we want these to have the same values as what we have for application audit of:
```
    USER_ADDED = "user_added"
    USER_UPDATED = "user_updated"
    USER_REMOVED = "user_removed"
```
Admittedly, not completely sold on how we phrased those, but a bit late to change them.

chouinar (commented):
Change is fine, although one discussion point.

Will need to resync from main after another DB migration was added

api/src/constants/lookup_constants.py
babebe: @chouinar No disagreement there, for keeping it consistent. For the updates, do we want it to be specific or any updates would then fall under "USER_UPDATED"

chouinar: @joshtonava - probably can weigh in a bit more. Consistency across these is a nice-to-have, but certainly not a hard requirement.

On the application side, we barely even have these scenarios yet because we don't have much user management (at least nothing a user directly works with).

joshtonava: I think its fine if everything is not 100% 1:1 consistent between applications and organizations if that gives us specificity on the organization events.

Im under the impression that  `USER_ROLE_UPDATED` more accurately describes the operation then `USER_UPDATED`. But at the same time, we could put the specific update information in the audit metadata.

On that front, 

Do we care about intermediate states for the user joining the org like

```python
USER_INVITED = "user_invited"
USER_REJECTED_INVITATION = "user_rejected_invitation"
```

Which would make 
```python
USER_ADDED = "user_added"
```
more accurately
```
USER_ACCEPTED_INVITATION = "user_accepted_invitation"
```

**Note**: we are already tracking this invitation specific info on the organization invitation record. 

Im open to being wrong on this what are your thoughts cc: @chouinar @babebe

babebe: If we’re just tracking high-level admin actions for now, I think `USER_ADDED = "user_added" and USER_UPDATED = "user_updated" ` are totally reasonable. As we add more detail over time, things like what specifically changed on the user can live in the metadata, which gives us room to get more granular without overcomplicating the event names upfront.

PR #7504, opened 2025-12-17
[Issue #7048] organization admin audit implement [2/2]

babebe:
## Summary

Fixes / Work for #7048  

## Changes proposed
- User accepting an organization invitation emit an audit event 
- User role updates emit a single audit event when user role is either added/removed
- User being removed from an organization emit an audit event
- Tests updated

### DIFF
--- a/api/src/form_schema/forms/epa_form_4700_4.py
+++ b/api/src/form_schema/forms/epa_form_4700_4.py
@@ -75,7 +75,7 @@ FORM_JSON_SCHEMA = {
         },
         "civil_rights_lawsuit_question1": {  # CivilRightsLawSuits1
             "allOf": [{"$ref": "#/$defs/civil_rights_lawsuit_question"}],
-            "title": "List all pending civil rights lawsuits and administrative complaints filed under federal law against the applicant/recipient that allege discrimination based on race, color, national origin, sex, age, or disability. (Do not include employment complaints not covered by 40 C.F.R. Parts 5 and 7.) IV. List all civil rights",
+            "title": "List all pending civil rights lawsuits and administrative complaints filed under federal law against the applicant/recipient that allege discrimination based on race, color, national origin, sex, age, or disability. (Do not include employment complaints not covered by 40 C.F.R. Parts 5 and 7.)",
         },
         "civil_rights_lawsuit_question2": {  # CivilRightsLawSuits2
             "allOf": [{"$ref": "#/$defs/civil_rights_lawsuit_question"}],
@@ -109,7 +109,7 @@ FORM_JSON_SCHEMA = {
         },
         "notice3": {  # Notice3
             "type": "boolean",
-            "title": "b. Is the notice posted in a prominent place in the applicant's/recipient’s website, in the offices or facilities or, for education programs and activities, in appropriate periodicals and other written communications?",
+            "title": "b. Is the notice posted in a prominent place in the applicant's/recipient's website, in the offices or facilities or, for education programs and activities, in appropriate periodicals and other written communications?",
         },
         "notice4": {  # Notice4
             "type": "boolean",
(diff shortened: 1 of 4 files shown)


api/src/services/organizations_v1/organization_audit.py
```diff
@@ -0,0 +1,39 @@
...
+
+def add_audit_event(
+    db_session: db.Session,
+    organization: Organization,
+    user: User,
+    audit_event: OrganizationAuditEvent,
+    target_user: User | None = None,
+) -> None:
+    """Add an organization audit event and log it"""
+    audit = OrganizationAudit(
```
joshtonava, line 19: **discussion**: The audit_metadata field exists in the `OrganizationAudit` model but is never populated. Should role changes (when updating user roles) include metadata about what roles were added/removed?

api/tests/src/api/users/test_org_invitation_response.py
```diff
@@ -62,6 +62,12 @@ def test_org_invitation_response_accepted(
...
     assert data["responded_at"] == res.accepted_at.isoformat()
 
+    # Verify audit history recorded
+    org = inv.organization
+    assert len(org.organization_audits) == 1
+    assert org.organization_audits[0].user.user_id == user.user_id
+    assert org.organization_audits[0].organization_audit_event == OrganizationAuditEvent.USER_ADDED
+
 
 def test_org_invitation_response_rejected(
```
joshtonava, line 72: **question:** Im wondering if we should be explicit here and assert that an audit event is not created. This is probably overkill and could be annoying to manage now that I think about it

api/src/services/organizations_v1/organization_audit.py
```diff
@@ -0,0 +1,39 @@
...
+from src.constants.lookup_constants import OrganizationAuditEvent
+from src.db.models.entity_models import Organization, OrganizationAudit
+from src.db.models.user_models import User
+
+logger = logging.getLogger(__name__)
+
+
+def add_audit_event(
+    db_session: db.Session,
+    organization: Organization,
```
joshtonava, line 13: **nitpick(non blocking):** I think this is totally fine but also noticed the application_audit event just passes the application_id, should we just pass the organization id?

chouinar: The application audit event equivalent functions has one that passes in ID and one that passes in the objects, I always prefer the objects since we usually have them anyways, and if we ever need want to add more to the logic, we can easily do that.

The ID case is needed to handle a very specific case on the apply side when a user tries to submit and it gets rejected as we only have IDs in that layer of the code.

api/src/services/organizations_v1/organization_audit.py
chouinar: The idea I had for the metadata is that for any specific details that wouldn't make sense to be tracked as part of the data model (eg. the before/after state like what roles a user got), but we didn't add any of those in the apply side yet.

I'd say it makes sense to skip that for now, and we can do it in a later iteration, especially if the frontend doesn't have plans to use it yet like on the apply side. Whatever gets done, we should just try to make the pattern the same in both places.

api/tests/src/api/users/test_org_invitation_response.py
chouinar: Could write it as something like checking that the user added audit event is not present, and not a catch-all "no events" check since I don't think user_added would ever be added to that path in the future.

api/src/services/organizations_v1/organization_audit.py
babebe: Also, not sure if we want to define a specific structure for the metadata.

api/tests/src/api/users/test_org_invitation_response.py
babebe: @chouinar , @joshtonava I saw the other audit events  as tracking actual membership changes, but a declined invite doesn’t change the org itself so we did not need to capture that here ?

chouinar: Can add to an existing test that verifies that the user-add event _wasn't_ added.

api/src/services/organizations_v1/organization_audit.py
chouinar: Yeah, I think we'd want something specific for the metadata so that it's consistent, would want to think through that and probably work with frontend folks to make sure it's in a format that makes sense to them.

api/tests/src/api/users/test_org_invitation_response.py
babebe: Added to test.

api/tests/src/api/users/test_org_invitation_response.py
```diff
@@ -98,6 +104,12 @@ def test_org_invitation_response_rejected(
     assert data["status"] == OrganizationInvitationStatus.REJECTED
     assert data["responded_at"] == res.rejected_at.isoformat()
 
+    # Verify user added audit event was not created
+    user_added_audits = [
+        audit for audit in inv.organization.organization_audits if audit.audit_event == "USER_ADDED"
```
chouinar:
```suggestion
        audit for audit in inv.organization.organization_audits if audit.audit_event == OrganizationInvitationStatus.USER_ADDED
```

chouinar (approved):

PR #7505, opened 2025-12-17
[Issue #6642] Create Get Application Submissions endpoint

### LINKED ISSUE #6642: Create Get Application Submissions endpoint
### Summary

We want an endpoint to get the application submissions for an application. This will enable the frontend to show a link to the application package objects (Zip, PDF, XML, etc).

We should call it `POST /alpha/applications/<uuid:application_id>/submissions`

#### Request
The request body will take in pagination info.

```json
{
  "pagination": {
    "page_offset": 1,
    "page_size": 25,
    "sort_order": [
      {
        "order_by": "created_at",
      }
    ]
  }
}
```
For defining this, use our usual `generate_pagination_schema` util like other pagination endpoints. For now we only need to meaningfully support `created_at` for the order_by fields supported, make the default `created_at` + `descending`.

#### Response
```json
{
 "data": [
   {
     "application_submission_id": "some uuid",
     "download_path": "https://...",
     "file_size_bytes": 1234,
     "legacy_tracking_number": 123456789,
   }
],
"pagination_info": {
    "page_offset": 1,
    "page_size": 25,
    "sort_order": [
      {
        "order_by": "created_at",
      }
    ],
    "total_pages": 2,
    "total_records": 42
  },
  "status_code": 200
}
```
We'll want to return a list of application submission objects (fields are all in the ApplicationSubmission model) + pagination info.

#### Auth
The user needs to have the `view_application` privilege on the application to call this endpoint.

#### Logic
Fairly straightforward / similar to many of our other application endpoints:
* Fetch the application -> 404 if not found
* Enforce RBAC -> will 403
* Fetch every application submission, sorting by the provided value
* Use our paginator utility for doing the pagination/getting the results

Should be pretty uneventful

### Acceptance criteria
- [ ] Endpoint created with pagination support and returning a list of application submissions sorted by created_at desc by default
- [ ] Tests added

mikehgrantsgov:
## Summary

Fixes #6642

## Changes proposed

Create new endpoint to get application submissions, service and routes
Add tests

## Context for reviewers

**Auth**
Both user key AND JWT auth should be allowed on this endpoint. Note that is our newer API key auth attached to a user, not the legacy API key auth. These both connect to a user, so the implementation shouldn't meaningfully change.

The user needs to have the view_application privilege on the application to call this endpoint.

**Logic**
Fairly straightforward / similar to many of our other application endpoints:

Fetch the application -> 404 if not found
Enforce RBAC -> will 403
Fetch every application submission, sorting by the provided value
Use our paginator utility for doing the pagination/getting the results

## Validation steps

See updated tests

### DIFF
--- a/api/src/api/application_alpha/application_route.py
+++ b/api/src/api/application_alpha/application_route.py
@@ -517,6 +517,7 @@ def application_submissions_list(
     logger.info("POST /alpha/applications/:application_id/submissions")
 
     # Get user from the multi-auth (supports both JWT and User API Key)
+    # TODO: Simplify this after multi-auth is updated to return user directly
     multi_auth_user = jwt_or_user_api_key_multi_auth.get_user()
 
     if isinstance(multi_auth_user.user, UserTokenSession):


api/src/auth/multi_auth.py
```diff
@@ -102,3 +102,20 @@ def _get_security_requirement(schemes: Sequence[str | None]) -> list[str | dict[
...
+# This is useful for application endpoints that need to support both auth types.
+jwt_or_user_api_key_multi_auth = MultiHttpTokenAuth(api_jwt_auth, api_user_key_auth)
+
+# List of security scheme names for JWT or User API Key multi-auth
+jwt_or_user_api_key_security_schemes = _get_security_requirement(
+    [
+        api_jwt_auth.security_scheme_name,
+        api_user_key_auth.security_scheme_name,
+    ]
+)
```
chouinar, line 121: FYI @hao10282025-sudo is working on a PR to add a slightly different way of setting of multi-auth for these two auth types that should make it so the `get_user` method actually returns the user and eliminates what you had to do with the type checking in the route.

We'll want to switch this over at some point (TODO for myself, add that ticket).

api/src/api/application_alpha/application_route.py
```diff
@@ -491,3 +499,47 @@ def application_audit_list(
...
+
+    # Get user from the multi-auth (supports both JWT and User API Key)
+    multi_auth_user = jwt_or_user_api_key_multi_auth.get_user()
+
+    if isinstance(multi_auth_user.user, UserTokenSession):
+        user = multi_auth_user.user.user
+    elif isinstance(multi_auth_user.user, UserApiKey):
+        user = multi_auth_user.user.user
+    else:
+        raise Exception(f"Unknown user type: {type(multi_auth_user.user)}")
```
chouinar, line 528: Could you add a note to adjust this after we adjust the auth (see my other comment)?

api/src/services/applications/list_application_submissions.py
```diff
@@ -0,0 +1,59 @@
...
+    paginator: Paginator[ApplicationSubmission] = Paginator(
+        ApplicationSubmission, stmt, db_session, page_size=params.pagination.page_size
+    )
+    paginated_results = paginator.page_at(page_offset=params.pagination.page_offset)
+    pagination_info = PaginationInfo.from_pagination_params(params.pagination, paginator)
+
+    return _transform_submissions(paginated_results), pagination_info
+
+
+def _transform_submissions(submissions: Sequence[ApplicationSubmission]) -> list[dict]:
```
chouinar: Do we need this? Should be fine to return the application submission objects since we're not changing the shape at all / Marshmallow can convert them to JSON uneventfully.

mikehgrantsgov: Makes sense, updated this

api/src/api/application_alpha/application_route.py
mikehgrantsgov: Yep, added a TODO here.

chouinar (approved):

PR #7523, opened 2025-12-17
[Issue #7394] setup simpler multiauth for JWT and new API key

### LINKED ISSUE #7394: Setup our multiauth to allow JWT auth or our new API key auth
### Summary

For our endpoints, we can configure multiple auth types to be allowed, we want to add JWT and API User Key auth to many of our endpoints, but first need to setup the multiauth to support this.

For this we want to:
* Create a [multiauth](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/auth/multi_auth.py) class alternative to `MultiHttpTokenAuth` - it should have a `get_user` function that returns an actual user - not one of the auth types like the existing one (we don't need the key, we want the user itself). The only possible auth types it should support are `UserApiKey` and `UserTokenSession`
* Define an auth object in the file that allows for either of the two auths
* Add tests that verify this auth can be used on a generic route (don't add it to a real route yet)
* Add docs on how to use it - calling out that you can't use the APIFlask decorator due to it not supporting multiauth yet

Later tickets will swap out the auth we have on endpoints to use this new multiauth

### Acceptance criteria

- [ ] New multiauth container class created which has a `get_user` method that always returns a user record regardless of which of the two auths it is.
- [ ] Tests added that verify the auth works

hao10282025-sudo:
## Summary

Fixes #7394  

## Changes proposed

- Create a [multiauth](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/auth/multi_auth.py) class alternative to MultiHttpTokenAuth - it should have a get_user function that returns an actual user - not one of the auth types like the existing one (we don't need the key, we want the user itself). The only possible auth types it should support are UserApiKey and UserTokenSession
- Define an auth object in the file that allows for either of the two auths
- Add tests that verify this auth can be used on a generic route (don't add it to a real route yet)
- Add docs on how to use it - calling out that you can't use the APIFlask decorator due to it not supporting multiauth yet

## Context for reviewers

For our endpoints, we can configure multiple auth types to be allowed, we want to add JWT and API User Key auth to many of our endpoints, but first need to setup the multiauth to support this.

## Validation steps

- New multiauth container class created which has a get_user method that always returns a user record regardless of which of the two auths it is.
- Tests added that verify the auth works

### DIFF
--- a/api/src/auth/multi_auth.py
+++ b/api/src/auth/multi_auth.py
@@ -93,7 +93,7 @@ api_key_multi_auth = MultiHttpTokenAuth(api_user_key_auth, api_key_auth)
 # We define the JWT auth first as the frontend will pass us a users JWT
 # and the frontend's API key for all user-based requests in order to validate
 # with our API gateway that handles rate limiting.
-jwt_or_api_user_key_multi_auth_simpler = MultiHttpTokenAuthSimpler(api_jwt_auth, api_user_key_auth)
+jwt_or_api_user_key_multi_auth = MultiHttpTokenAuthSimpler(api_jwt_auth, api_user_key_auth)
 
 
 # Helper function to format security schemes for OpenAPI
@@ -127,7 +127,7 @@ api_key_multi_auth_security_schemes = _get_security_requirement(
 )
 
 # List of security scheme names for JWT or API User Key multi-auth
-jwt_or_api_user_key_simpler_security_schemes = _get_security_requirement(
+jwt_or_api_user_key_security_schemes = _get_security_requirement(
     [
         api_jwt_auth.security_scheme_name,
         api_user_key_auth.security_scheme_name,
--- a/api/tests/src/auth/test_multi_auth.py
+++ b/api/tests/src/auth/test_multi_auth.py
@@ -9,7 +9,7 @@ from src.auth.internal_jwt_auth import create_jwt_for_internal_token
 from src.auth.multi_auth import (
     AuthType,
     jwt_key_or_internal_multi_auth,
-    jwt_or_api_user_key_multi_auth_simpler,
+    jwt_or_api_user_key_multi_auth,
     jwt_or_key_multi_auth,
 )
 from src.util import datetime_util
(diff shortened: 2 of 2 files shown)


api/src/auth/multi_auth.py
```diff
@@ -73,6 +83,16 @@ def get_user(self) -> MultiAuthUser:
...
 
 
+# Define the multi auth that supports
+# * User JWT auth
+# * API User Key Auth
+#
+# Note that the order defined matters - earlier ones will take precedence in
+# the event a user provides us with multiple auth approaches at once, only the first
+# relevant one will be used
+jwt_or_api_user_key_multi_auth = MultiHttpTokenAuthSimpler(api_jwt_auth, api_user_key_auth)
```
chouinar: A bit more context we could add for why we want this exact order:

```suggestion
# Define the multi auth that supports
# * User JWT auth
# * API User Key Auth
#
# Note that the order defined matters - earlier ones will take precedence in
# the event a user provides us with multiple auth approaches at once, only the first
# relevant one will be used
# We define the JWT auth first as the frontend will pass us a users JWT
# and the frontend's API key for all user-based requests in order to validate
# with our API gateway that handles rate limiting.
jwt_or_api_user_key_multi_auth = MultiHttpTokenAuthSimpler(api_jwt_auth, api_user_key_auth)
```

api/tests/src/auth/test_multi_auth.py
```diff
@@ -49,6 +54,17 @@ def dummy_internal_endpoint():
...
 
+    @mini_app.get("/dummy_auth_endpoint_simpler")
+    @mini_app.auth_required(jwt_or_api_user_key_multi_auth)
+    def dummy_endpoint_simpler():
+        user = jwt_or_api_user_key_multi_auth.get_user()
+
+        return {
+            "message": "ok",
+            "user_id": getattr(user, "user_id", None),
+            "username": getattr(user, "username", None),
```
chouinar: Nitpick - we don't have a username field, so returning and checking this isn't going to get anything.

api/src/auth/multi_auth.py
hao10282025-sudo: updated with commit: https://github.com/HHS/simpler-grants-gov/pull/7523/commits/da10c7a4181ea0bca5fd3d7fe332a57d1eb67079

api/tests/src/auth/test_multi_auth.py
hao10282025-sudo: Updated with commit: https://github.com/HHS/simpler-grants-gov/pull/7523/commits/da10c7a4181ea0bca5fd3d7fe332a57d1eb67079

Issue #6663, opened 2025-10-08
Fetching the data (database call) from Simpler for GetSubmissionListExpanded


chris-kuryak:
### Summary
https://grants.gov/system-to-system/grantor-system-to-system/web-services/get-submission-list-expanded

Fetching the data (database call) from Simpler

Make a query, grab the application submissions from the DB and restructure them into the XML response format.

These are the filters that exist:
* The automatic / implicit filter from the certificate (ie. needs to filter by agency code which may be prefix based)
* Status - Needs to be mapped (See below)
* Funding Opportunity Number - Already added
* CFDA Number - Already added
* Submission Title - skip for now
* Grants.gov Tracking Number - Already added
* Opportunity ID - Need to add, queries against the `legacy_opportunity_id` on the opportunity
* Agency Code - Need to add, it should be in addition to the automatic/implicit one from the cert
* Competition ID - skip for now
* Package ID - skip for now

Any of the filters we don't yet support, log a metric when those filters are provided.

For status, we'll map the statuses to our DB values as follows when doing queries. Some statuses won't ever return anything, if they say Nothing, we can lazily never do the query as we know we have nothing in those statuses:
* Receiving -> Nothing
* Received -> Nothing
* Processing -> `submitted`
* Validated -> `accepted`
* Rejected with Errors -> Nothing 
* Download Preparation -> Nothing
* Received by Agency -> `agency_fetched`
* Agency Tracking Number Assigned -> `agency_fetched` + we've added the agency tracking number to the submission.

### Acceptance criteria

- [ ] GetSubmissionListExpanded filters can query against our submissions and return our data in the grants.gov XML format
- [ ] Endpoint works end-to-end, can query the route and get results

jakobpederson:
* CFDA Number - Already added
* Grants.gov Tracking Number - Already added
* Funding Opportunity Number - Already added

Add:
- [x] Status - Needs to be mapped
- [x] Opportunity ID - Need to add, queries against the `legacy_opportunity_id` on the opportunity
- [x] Agency Code - Need to add, it should be in addition to the automatic/implicit one from the cert

Add logging:
- [x]  Submission Title
- [x] Competition ID
- [x] Package ID

PR #7525, opened 2025-12-17
[Issue #6663] Fetch simpler data for get submission list expanded

jakobpederson:
## Summary

Fixes / Work for #6663 

## Changes proposed

Wiring up GetSubmissionListExpanded which means putting it into the SimplerGrantorsS2SClient and add tests that prove that the client can handle a request. The grantors services part had to be updated to produce a dictionary correctly to be turned into xml.

## Context for reviewers

Basically, a request to GetSubmissionListExpanded should get an xml response based on submissions from our db. Currently this does not happen because GetSubmissionListExpanded is not an operation that exists. This PR should add that operation, and then it should be something that produces data from our db and accepts the filters from the legacy endpoint.

### DIFF
--- a/api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
+++ b/api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
@@ -1,5 +1,4 @@
 import logging
-from collections.abc import Sequence
 from datetime import datetime
 
 from sqlalchemy import select
@@ -73,11 +72,11 @@ def transform_submission(submission: ApplicationSubmission) -> dict[str, str | d
     return application_list_obj
 
 
-def get_submission_list_expanded_response(
+def get_submissions(
     db_session: db.Session,
-    get_submission_list_expanded_request: schemas.GetSubmissionListExpandedRequest,
+    request: schemas.GetSubmissionListExpandedRequest,
     soap_request: SOAPRequest,
-) -> schemas.GetSubmissionListExpandedResponse | list:
+) -> list[ApplicationSubmission]:
     stmt = (
         select(ApplicationSubmission)
         .join(ApplicationSubmission.application)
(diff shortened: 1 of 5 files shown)


api/src/legacy_soap_api/grantors/schemas/get_submission_list_expanded_schemas.py
```diff
@@ -22,20 +22,33 @@ class SubmissionInfo(BaseSOAPSchema):
...
+        cleaned_submissions = []
+        if body.get("ns2:GetSubmissionListExpandedResponse"):
+            for item in body["ns2:GetSubmissionListExpandedResponse"].get("ns2:SubmissionInfo", []):
+                cleaned_submissions.append(
+                    {key: val for key, val in item.items() if val is not None}
+                )
+            body["ns2:GetSubmissionListExpandedResponse"][
+                "ns2:SubmissionInfo"
+            ] = cleaned_submissions
+        return {"Body": body}
```
jakobpederson: This is to remove fields that are None which seems to be behavior we see from the endpoints, instead of the fields appearing as blank like `<CFDANumber></CFDANumber>`

api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
```diff
@@ -83,18 +86,48 @@ def get_submission_list_expanded_response(
...
+
+    STATUS_TRANSFORM = {
+        "Receiving": None,
+        "Received": None,
+        "Processing": ApplicationStatus.SUBMITTED,
+        "Validated": ApplicationStatus.ACCEPTED,
+        "Rejected with Errors": None,
+        "Download Preparation": None,
+        "Received by Agency": ApplicationStatus.ACCEPTED,
+        "Agency Tracking Number Assigned": ApplicationStatus.ACCEPTED,
```
chouinar: Note that these will require new statuses - although fine if we wait to add those in the later endpoints we build out.

But I'd add a comment to say we need to adjust that later.

api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
```diff
@@ -83,18 +86,48 @@ def get_submission_list_expanded_response(
...
+    }
+    # Getting only one status from a potential list of statuses because multiple statuses on a filter don't make sense'
+    if isinstance(status, list) and len(status) > 0:
+        simpler_status = STATUS_TRANSFORM.get(status[0]) if isinstance(status[0], str) else None
+    else:
+        simpler_status = None
```
chouinar: How does grants.gov respond to multiple statuses, it would make sense if it's an OR (a list being "Status A" or "Status B"). If their logic is conceptually "A AND B" then I'd almost say the appropriate thing is to immediately skip everything else.

api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
```diff
@@ -83,18 +86,48 @@ def get_submission_list_expanded_response(
...
             grants_gov_tracking_numbers = submission_filters.get("GrantsGovTrackingNumber")
             cfda_numbers = submission_filters.get("CFDANumber")
             funding_opportunity_numbers = submission_filters.get("FundingOpportunityNumber")
+            opportunity_ids = submission_filters.get("OpportunityID")
+            competition_ids = submission_filters.get("CompetitionID")
+            package_ids = submission_filters.get("PackageID")
+            status = submission_filters.get("Status")
+            submission_titles = submission_filters.get("SubmissionTitle")
+
+    STATUS_TRANSFORM = {
```
chouinar: Can you move this to the top of the file?

api/src/legacy_soap_api/legacy_soap_api_config.py
```diff
@@ -141,7 +141,13 @@ class SOAPOperationConfig:
             response_operation_name="GetApplicationZipResponse",
             is_mtom=True,
             privileges={Privilege.LEGACY_AGENCY_GRANT_RETRIEVER},
-        )
+        ),
+        "GetSubmissionListExpandedRequest": SOAPOperationConfig(
+            request_operation_name="GetSubmissionListExpandedRequest",
+            response_operation_name="GetSubmissionListExpandedResponse",
+            is_mtom=True,
```
jakobpederson, line 148: We may want to turn this off for now till we get the syncing of the two responses going. Right now it should be dererring to the simpler response, which may not be the behavior we want.

api/tests/src/legacy_soap_api/grantors/services/test_legacy_soap_api_get_submission_list_expanded_response.py
```diff
@@ -621,23 +716,529 @@ def test_get_submission_list_expanded_response_handles_competition_with_no_oppor
...
+        )
+        submission = setup_application_submission(
+            agency,
+            legacy_package_id="PKG00118065",
+            sam_gov_entity=sam_gov_entity_1,
+            application_status=ApplicationStatus.SUBMITTED,
+        )
+        sam_gov_entity = submission.application.organization.sam_gov_entity
+        sam_gov_entity.has_debt_subject_to_offset = True
+        sam_gov_entity.has_exclusion_status = True
```
chouinar: I'm curious about how this code is setup. You use the factory to create a Sam gov entity. Pass it into the setup function, then pull it out of the submission created and then change the entity?

Why not just set those parameters first? Actually, do these parameters get used by anything you implemented, I don't think we currently use those anywhere, why are setting these at all (before or after)?

api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
jakobpederson: I've been testing it and I believe we're looking ad `A AND B` behavior, so I'll put in something to just skip.

api/tests/src/legacy_soap_api/grantors/services/test_legacy_soap_api_get_submission_list_expanded_response.py
jakobpederson: It's leftover code from other tests that got mixed up. I've tried to go through and clean it up.

These fields are impacted by the sam gov entity:
```
<UEI>
<DelinquentFederalDebt>
<ActiveExclusions>
```

api/src/legacy_soap_api/grantors/schemas/get_submission_list_expanded_schemas.py
chouinar: I think the `to_soap_envelope_dict` method in the base function accounted for this with the `exclude_none` call, if you add that does it solve the problem?

api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
```diff
@@ -83,31 +98,88 @@ def get_submission_list_expanded_response(
...
+        logger.info(f"GetSubmissionListExpanded Filter: PackageIDs {package_ids}")
+    if submission_titles:
+        logger.info(f"GetSubmissionListExpanded Filter: SubmissionTitles {submission_titles}")
+
+    simpler_status = None
+    submissions_query: Sequence[ApplicationSubmission] = []
+    # If more than one status in filter then just return nothing
+    if status and len(status) > 1:
+        submissions_query = []
+    else:
```
chouinar: To avoid some of this nesting, a suggestion. Could you make a function like:

```
get_application_submissions( ... ) -> Sequence[ApplicationSubmission]:
    if status and len(status) > 1:
        return []

    # the else statement here
```
Then we can simplify the `for submission in submissions` bit below.

api/src/legacy_soap_api/legacy_soap_api_config.py
chouinar: If we have the merging of the responses coming up, I agree that it should probably just return legacy grants.gov instead, although if no one is using it, doesn't matter that much

api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
```diff
@@ -83,31 +98,81 @@ def get_submission_list_expanded_response(
             )
         )
     )
+
     grants_gov_tracking_numbers = None
     cfda_numbers = None
     funding_opportunity_numbers = None
+    opportunity_ids = None
```
chouinar: Didn't quite see the pattern before, but I am curious why we are doing:

```
field = None

field = submission_filters.get("field")

if field:
    stmt = stmt.where( ... )
```

This is a lot of extra code when you really just care about the if statement at the end.

You can do something like:
```py
# It might be a good idea to rename get_submission_list_expanded_request to just request for readability.
if get_submission_list_expanded_request.expanded_application_filter and get_submission_list_expanded_request.expanded_application_filter.filters:
     submission_filters = get_submission_list_expanded_request.expanded_application_filter.filters

     if field := submission_filters.get("field"):
         stmt = stmt.where( ... )
```

api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
```diff
@@ -62,7 +76,8 @@ def transform_submission(submission: ApplicationSubmission) -> dict[str, str | d
 def get_submission_list_expanded_response(
     db_session: db.Session,
     get_submission_list_expanded_request: schemas.GetSubmissionListExpandedRequest,
-) -> schemas.GetSubmissionListExpandedResponse:
+    soap_request: SOAPRequest,
+) -> schemas.GetSubmissionListExpandedResponse | list:
```
chouinar: In what case would we return something other than the response type? If there are no results like the status case, wouldn't we still return a response object, but the submission info list would be empty?

The idea I suggested in a prior PR was to make a new function that handles the DB query logic and returns the results (and can shortcut to return `[]`) and this logic just uses that list.


This function should just be:
```py
submissions = get_submissions( ... )

for submission in submissions:
    submission_list_obj = transform_submission(submission)
    info.append(schemas.SubmissionInfo(**submission_list_obj))
return schemas.GetSubmissionListExpandedResponse(
    success=True, available_application_number=len(info), submission_info=info
    )
```

api/src/legacy_soap_api/legacy_soap_api_client.py
```diff
@@ -232,6 +235,18 @@ def GetApplicationZipRequest(self) -> grantors_schemas.GetApplicationZipResponse
             soap_config=self.operation_config,
         )
 
+    def GetSubmissionListExpandedRequest(
```
chouinar: Functions shouldn't be capitalized, this would be misinterpreted as a class.

```suggestion
    def get_submission_list_expanded(
```

Why does this optionally return a list?

jakobpederson: This was a consequence of code calling the operation_name 
```
        operation_method = getattr(self, self.operation_config.request_operation_name))
```
Where the operation_name is camel case like `GetApplicationZipRequest`. I've updated the methods to snake case.

jakobpederson: It no longer returns a list, once the other comment was resolved it no longer needed to have an optional list return.

api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
jakobpederson: That's a lot better, I did overcomplicate it.

api/src/legacy_soap_api/legacy_soap_api_client.py
```diff
@@ -56,7 +57,9 @@ def get_soap_response_dict(self) -> dict:
         Example:
             {'Envelope': {'Body': {'GetOpportunityListResponse': {...}}}}
         """
-        operation_method = getattr(self, self.operation_config.request_operation_name)
+        operation_method = getattr(
+            self, to_snake_case(self.operation_config.request_operation_name)
+        )
```
chouinar, line 62: Ah, hadn't realized that the function name was being used in that way, but good way of fixing it. Alternatively could have a mapping config to the function, but this works.

Issue #7340, opened 2025-12-04
Cleanup TuserAccount table setup


joshtonava:
### Summary

The TuserAccount and TuserAccountMapper tables are no longer being synced from the Oracle database. These tables were part of the legacy data migration pipeline, but they are no longer receiving updates from Oracle.

related #7329 

### Tech Notes

These models are currently defined in:
  - Foreign tables: api/src/db/models/foreign/user.py:13-19 (Oracle FDW connection)
  - Staging tables: api/src/db/models/staging/user.py:14-20 (local copies)
  - Legacy mixins: api/src/db/models/legacy_mixin/user_mixin.py:7-40 (shared column definitions)

  The models are still being referenced in:
  1. The Oracle data load task (load_oracle_data_task.py:34) - tuser_account_mapper is in TABLES_TO_LOAD
  2. The get_legacy_user_for_login_gov_id() function in src/services/legacy_users/utils.py:8-22
  3. Test files for the above function

### Acceptance criteria

- Remove the tables with a migration
- Remove all related code
- Drop tables in each environment

PR #7553, opened 2025-12-18
[Issue #7340] user account and tuser account mapper cleanup

babebe:
## Summary

Fixes / Work for #7340  

## Changes proposed
- Removed legacy models for `tuser_account` and `tuser_account_mapper`
- Removed util func `get_legacy_user_for_login_gov_id ` that is no longer needed 
- Migration file for dropping tables
- Removed related tests and factories


joshtonava (approved):
Verified changes locally

Issue #7079, opened 2025-11-17
E2E testing for "and / or" search term switch


doug-s-nava:
### Summary

"And / or" search term verb switch is untested in e2e suite

### Acceptance criteria

Test(s) written which:

* ensure proper functionality of "and / or" switch in UI / url when search term is entered
* ensure proper functionality of "and / or" switch in UI / url when search term is not entered (should not change anything)

ychoy:
@aoysimmons-GH please check with @eric-serrato95 when your test plans are ready for review

Issue #4076, opened 2025-03-07
Add e2e tests for saving search


doug-s-nava:
### Summary

As a follow up to building the save search feature, e2e tests should be written to verify all functionality

Related to https://github.com/HHS/simpler-grants-gov/pull/4073

~~**Note:** blocked until we can log in with the e2e test suite using login.gov (related tickets, https://github.com/HHS/simpler-grants-gov/issues/3459, https://github.com/HHS/simpler-grants-gov/issues/3793)~~

### Acceptance criteria

Tests written to test:

- [x] copy search to clipboard (can be done without auth)
- [x] save search to account
- [x] view saved searches in search page
- [x] apply saved search from search page
- [x] view saved searches in workspace page
- [x] apply saved search from workspace page
- [x] access workspace pages from nav

aoysimmons-GH:
Verify the search query is correctly copied to the clipboard, which can be tested without authentication.

I login to 'https://simpler.grants.gov/'
I click on 'Search' tab
I enter 'K01 Mentored Research Scientist' at 'Search' feild
I click on 'Search' button
I see '9 Opportunities'
I click on 'Copy' button
I see 'Copied!' button displayed
I see 'dialog' displayed text 'This search query was copied to your clipboard. Paste it as a link anywhere.'
I verify the data in the clipboard should be 'https://simpler.grants.gov/search?query=K01+Mentored+Research+Scientist'

I logout
**Note: I see the text of copy button is different when login showing 'Copy' Vs not login showing 'Copy this search query'**
I navigate to the 'https://simpler.grants.gov/search?query=K01+Mentored+Research+Scientist'
I see '9 Opportunities'

Screenshot below:

<img width="3433" height="1422" alt="Image" src="https://github.com/user-attachments/assets/d8f27bf9-6791-4e4c-8697-562dd5301960" />

aoysimmons-GH:
**Verify the user can save the search to the Workspace.**

I login to 'https://simpler.grants.gov/'
I click on 'Search' tab
I enter 'K01 Mentored Research Scientist' at 'Search' feild
I click on 'Search' button
I see '9 Opportunities'
I click on 'Save' button

I see 'dialog' displayed text
'Name this search query'
'Save these search terms and filters with a name for easy access later.'
'Name (required)'
I see 'saved-search-input' is displayed
I see 'Save' button
I see 'Cancel' button

I click on 'Save button
I see alert text 'Please name this query.'

[...]

PR #7556, opened 2025-12-18
Feature file - Search functions

aoysimmons-GH:
Feature file - Search functions

## Summary

Fixes / Work for #4076, #7079

### DIFF
--- a/frontend/tests/e2e/search/search-SQLi-XSS.feature
+++ b/frontend/tests/e2e/search/search-SQLi-XSS.feature
@@ -1,0 +1,24 @@
+@7079
+Feature: Open search field secure from SQL Injection (SQLi) and Cross-Site Scripting (XSS) Attack
+
+Background: Navigate to search page after login
+    Given I click 'Search' tab
+
+@login @PendingAutomate
+Scenario Outline: Verify search result when search with query injection
+    When I enter <search parameter> at 'Search' field
+    And I click 'Search' button
+    Then I see each entry in 'Search Results' section contains <search parameter>
+    And I verify if response is <Likely Response>
+    And I verify if 403 is expected as <403 Expected?>
+    Examples:
+        | search parameter                      | Backend / Frontend | Likely Response     | 403 Expected?              | Notes                                                                                | Our result                             |
+        | K01'; DROP TABLE users;--             | Backend            | 403 / 400           | ✅ 403 can be expected      | Web application firewall (WAF) may block SQLi attempts                               | Your search didn't return any results. |
+        | '; EXEC xp_cmdshell('dir');--         | Backend            | 403 / 400           | ✅ Expected                 | Dangerous OS command, WAF should block                                               | Your search didn't return any results. |
+        | ' OR '1'='1                           | Backend            | 200 / 401 / 403     | ⚠ Depends                  | If unauthenticated, 403 can be expected; if logged in, 403 may indicate overblocking | Your search didn't return any results. |
+        | ; WAITFOR DELAY '00:00:05'--          | Backend            | 403 / 500           | ✅ 403 safe                 | Delay attempts can trigger WAF throttling                                            | 403 Forbidden                          |
+        | <script>alert('XSS')</script>         | Frontend           | 200 + browser alert | ❌ 403 usually not expected | XSS is normally blocked via sanitization, not HTTP code                              | 403 Forbidden                          |
+        | %3Cscript%3Ealert('XSS')%3C/script%3E | Frontend           | 200                 | ❌ 403 usually not expected | Encoded scripts should be escaped, not rejected with 403                             | 403 Forbidden                          |
+        | ../../../../etc/passwd                | Backend            | 403 / 400           | ✅ Expected                 | Path traversal blocked by server access control                                      | 403 Forbidden                          |
+        | %27%20OR%20%271%27%3D%271             | Backend            | 403 / 400           | ✅ Expected                 | Encoded SQLi blocked by WAF                                                          | Your search didn't return any results. |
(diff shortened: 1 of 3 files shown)


frontend/tests/e2e/search/search-SQLi-XSS.feature
```diff
@@ -0,0 +1,24 @@
+@7079
```
doug-s-nava, line 1: can we move this to a separate PR?

frontend/tests/e2e/search/search-save.feature
```diff
@@ -0,0 +1,140 @@
...
+@login @PendingAutomate
+Scenario Outline: Verify copy clipboard
+    When I click 'Copy' button
+    And I see 'button'
+        | button    |
+        | 'Copied!' |
+    And I see 'text' at dialog
+        |text|
+        |This search query was copied to your clipboard. Paste it as a link anywhere.|
+    Then I see 'clipboard'
```
doug-s-nava, line 36: not sure what this means - can we verify that the correct url is pasted when the user does a "paste" after clicking "copy"?

frontend/tests/e2e/search/search-save.feature
```diff
@@ -0,0 +1,140 @@
...
+    And I see dialog displayed
+    And I 'type' 'field' in dialog
+        | type                    | field |
+        | Test-2-Test save search | Name  |
+    And I click 'Save' button
+    When I click 'Workspace' link
+    And I click 'hyperlink'
+        | hyperlink               |
+        | Test-2-Test save search |
+    Then I see 'search parameter' section populated
```
doug-s-nava, line 138: can we test the display for all filter options, as well as sort order, not just the search query?

frontend/tests/e2e/search/search-switch-andor.feature
```diff
@@ -0,0 +1,570 @@
...
+        |/search|
+
+@login @PendingAutomate
+Scenario: Verify the search defaults to AND condition
+    Then I see 'search-verb-switch' selected
+        |search-verb-switch|
+        |Must include all words (ex. transportation AND safety)|
+
+@login @PendingAutomate
+Scenario: Verify no search term entered when switching from AND to OR
```
doug-s-nava, line 29:
```suggestion
Scenario: Verify no change occurs when no search term entered and switching from AND to OR

```

frontend/tests/e2e/search/search-switch-andor.feature
```diff
@@ -0,0 +1,570 @@
...
+    Then I see each entry in 'Search Results' section contains 'ti'
+    And I select 'May include any words (ex. transportation OR safety)' at 'search-verb-switch'
+    Then I see each entry in 'Search Results' section contains 'ti'
+
+@login @PendingAutomate
+Scenario: Verify multiple search results when switching from OR to AND
+    When I select 'May include any words (ex. transportation OR safety)' at 'search-verb-switch'
+    And I enter 'Research Scientist' at 'Search' field
+    And I click 'Search' button
+    Then I see each entry in 'Search Results' section contains 'Research' OR 'Scientist'
```
doug-s-nava, line 119: this is going to be difficult to verify, since our search function looks at the opportunity description, and all other opportunity metadata to determine matches. If you can put some time in to a test case that is somewhat easier to verify as far as the difference between and AND based and OR based search that would be good. You could also pinpoint particular opportunities that will appear in an OR case but not an AND case and assert on that. If it's not really possible or reliable enough of a test case, it might be enough to assert that the number of results is different between the two cases

frontend/tests/e2e/search/search-switch-andor.feature
```diff
@@ -0,0 +1,570 @@
...
+    Then I see 'search-verb-switch' selected
+        |search-verb-switch|
+        |Must include all words (ex. transportation AND safety)|
+    And I see each entry in 'Search Results' section contains 'Research' AND 'Scientist' 
+
+@login @PendingAutomate
+Scenario: Verify search result count when switching from AND to OR
+    When I enter 'K01 Mentored Research Scientist' at 'Search' field
+    And I click 'Search' button
+    And I see '9' at 'Search Results' section
```
doug-s-nava, line 165: these numbers will change over time. Probably best to assert that the number for an AND search will always be less than or equal to the number for an OR search

frontend/tests/e2e/search/search-switch-andor.feature
```diff
@@ -0,0 +1,570 @@
...
+        | results           |
+        | 9                 |
+    And I select 'May include any words (ex. transportation OR safety)' at 'search-verb-switch'
+    And I see '15' at 'Search Results' section
+        | results           |
+        | 15                |
+
+@login @PendingAutomate
+Scenario: Verify search result count remains the same when no search term entered
```
doug-s-nava, line 195: this doesn't seem like a useful test case

frontend/tests/e2e/search/search-switch-andor.feature
```diff
@@ -0,0 +1,570 @@
...
+    And I see '1000+' at 'Search Results' section
+        | results           |
+        | 1000+             |
+    And I select 'Must include all words (ex. transportation AND safety)' at 'search-verb-switch'
+    And I see '1000+' at 'Search Results' section
+        | results           |
+        | 1000+             |
+
+@login @PendingAutomate
+Scenario Outline: Verify search result when search with arithmetic operators when switching from AND to OR
```
doug-s-nava, line 211: what is this testing? We won't expect users to use these sorts of search terms, but maybe bots will?

frontend/tests/e2e/search/search-SQLi-XSS.feature
aoysimmons-GH: Yes, I can do that.

PR #7563, opened 2025-12-18
[unticketed] specify max upload file size in a second place

doug-s-nava:
## Summary

unticketed

## Changes proposed

Adds a specification to allow uploading files with sizes greater than 10mb and less than 2gb

## Context for reviewers

We have one config option set up to allow files up to 2gb, but apparently there is a second config option for upload file size that pertains specifically to middleware for some reason that we also need to set. It may not have come up before because the default in the second setting is 10mb instead of 1mb on the other setting

## Validation steps

2. visit http://localhost:3000
3. log in as many_app_user
4. access the test application through the user dropdown
5. upload a file > 10 mb < 2gb in size
6. _VERIFY_: success


doug-s-nava:
@mdragon I have not tested extensively to see what is going on here yet, but it's definitely worth looking into:
* are we actually not running middleware on api routes? I see a comment suggesting that but nothing in the code that backs it up
* should these requests be going through middleware?
* where the error is coming from - the docs say it shouldn't error so it's possible there's something in our implementation that is causing an error based on the underlying middleware processes they're talking about

Issue #7107, opened 2025-11-18
Create web and print views for form EPA 4700-4


chris-kuryak:
### Summary

This work is to use the JSON from # 

Form information: https://www.grants.gov/forms/form-items-description/fid/

### Acceptance criteria

- [ ] Webform matches expected layout from PDF form
- [ ] Print View generates as expected
- [ ] Backend has created an Opportunity for testing this form and it is deployed to **staging**
- [ ] Product and Design notified for UAT
- [ ] Product UAT complete
  - [ ] Webform
  - [ ] PDF
- [ ] Design UAT complete
  - [ ] Webform
  - [ ] PDF

chris-kuryak:
@myduong-navapbc 
UAT not passed. I don't see the Signature or Signature Date populating at all, whether "unknown" or not, on the PDF or Webform after submission.

I know we have had issues with the Signature saying "unknown", which is fine for now, but this appears to not populate any data in these fields.

<img width="813" height="575" alt="Image" src="https://github.com/user-attachments/assets/9c18611f-c459-4f4c-aa19-655d0bbc41bd" />
<img width="673" height="487" alt="Image" src="https://github.com/user-attachments/assets/3ccda614-d60f-43b7-8736-adb550640b3a" />

chris-kuryak:
New PDF UAT Application ID: ee657bd2-9f1f-4aed-8560-5bc17e8b3d4e

chris-kuryak:
Working now. UAT passed.

<img width="822" height="584" alt="Image" src="https://github.com/user-attachments/assets/92a5cc65-04f7-4abd-8cb3-f552fb633c6b" />
<img width="734" height="514" alt="Image" src="https://github.com/user-attachments/assets/40878ac1-2076-4ed9-a2c3-fb3674b71dea" />

PR #7627, opened 2025-12-19
[Issue #7107] EPA 4700-4 v5 Form UI Schema

myduong-navapbc:
## Summary

Work for #7107 

## Changes proposed

Minor changes proposed: 
- typo fixed
- each section is named by numbering. there will be later improvements to name the actual sections

## Context for reviewers

The current form
[EPA4700_4_5_0-V5.0.pdf](https://github.com/user-attachments/files/24266361/EPA4700_4_5_0-V5.0.pdf)

Example of print form (from doc raptor)
[doc-570385337.pdf](https://github.com/user-attachments/files/24266364/doc-570385337.pdf)

## Validation steps
3. navigate to `search`
4. search for `test`
5. you should see an opportunity called `Test Opportunity for GG_LobbyingForm 1.1`
6. sign in
7. start new application
9. click Grants.gov Lobbying Form
10. VERIFY the form loads
11. fill out the form, save
12. run `make generate-internal-token` from `api` directory
13. grab the token and place in request header "X-SGG-Internal-Token" and the token (I use a plugin called Modheader)
14. navigate to http://localhost:3000/print/application/APP_ID/form/FORM_ID
15. VERIFY the form data saves correctly

### DIFF
--- a/api/src/form_schema/forms/epa_form_4700_4.py
+++ b/api/src/form_schema/forms/epa_form_4700_4.py
@@ -184,7 +184,7 @@ FORM_UI_SCHEMA = [
         "label": "I. B. Unique Entity Identifier (UEI)",
         "name": "uei",
         "children": [
-            {"type": "field", "definition": "/properties/sam_uei"},
+            {"type": "null", "definition": "/properties/sam_uei"},
         ],
     },
     {


ErinPattisonNava:
Also meant to confirm that the state should be formatted like this?  
Screenshot 2025-12-19 at 6 56 46 PM

chouinar:
> Can you confirm that none of the fields in section III-XI are required? It looks a little weird in the final PDF if none of that stuff is set.

https://grants.gov/forms/form-items-description/fid/773 has the required fields in a semi-readable format, most of them aren't


> I'm confused about `Unique Entity Identifier (UEI)`, it feels like that should be a field that we pre-fill and they can't edit OR a field they need to fill in and we don't do that for them.

It is pre-populated, it shouldn't be something a user should be able to fill out (in my first pass as the UI schema, I didn't set the type to `null` like it should be).

myduong-navapbc:
> > I'm confused about `Unique Entity Identifier (UEI)`, it feels like that should be a field that we pre-fill and they can't edit OR a field they need to fill in and we don't do that for them.
> 
> It is pre-populated, it shouldn't be something a user should be able to fill out (in my first pass as the UI schema, I didn't set the type to `null` like it should be).

The SF424 (standard) SAM UEI is treated the same way- So, I treated this the same. I'll change it here and create a ticket to update the 424 in a separate PR.

myduong-navapbc:
> Can you confirm that none of the fields in section III-XI are required? It looks a little weird in the final PDF if none of that stuff is set.
> 
Can confirm. Everything from II to XII is optional

myduong-navapbc:
> The error message for the `B. Title of Authorized Official is required` looks weird, but I assume that is expected?
> 
The decision:
The decision was to leave it as-is.

Issue #7356, opened 2025-12-05
Seed legacy users for local development


joshtonava:
### Summary

It would be useful for testing the legacy users invite flow to have legacy users seeded in the DB.

### Acceptance criteria

- Add to the script for existing organizations to add in legacy users
- Update tests

PR #7637, opened 2025-12-22
[Issue #7356] seed local data org legacy user

babebe:
## Summary
Added seed data and test scenarios for legacy users, including AVAILABLE, MEMBER, and PENDING_INVITATION states. This allows testing the invite process, membership management, and org-specific permissions for legacy users.

Fixes / Work for #7356  

## Changes proposed

- Created AVAILABLE legacy users per organization (eligible for invite, no invite sent yet)
- Created MEMBER legacy users per organization (already part of the org)
- Created PENDING_INVITATION legacy users for specific org2 (invite sent, not yet accepted)
- Updated user_scenarios to append one entry per created legacy user 

## Context for reviewers

AVAILABLE legacy users are tied to a specific organization but have not been invited yet
MEMBER legacy users simulate existing org members for role and permission testing
PENDING_INVITATION legacy users allow testing the invite lifecycle

This data enables testing of:

- Invite acceptance/rejection flows
- Organization membership state changes
- user_scenarios updates reflect the actual users created to make the test setup traceable

## Validation steps

Run `db-seed-local` in the Makefile, should see:
- legacy users created in staging `v_user_account` and `t_user_profile` table. 
- should see `organization_user` created for the member legacy users (3) - one for each org

### DIFF
--- a/documentation/wiki/product/release-notes.md
+++ b/documentation/wiki/product/release-notes.md
@@ -1,5 +1,47 @@
 # Release Notes
 
+## [Simpler.Grants.Gov](http://simpler.grants.gov) Release Notes - December 23, 2025
+
+December 23rd’s update brings [several new changes](https://github.com/HHS/simpler-grants-gov/releases): &#x20;
+
+{% columns %}
+{% column width="25%" %}
+#### New Feature
+{% endcolumn %}
+
+{% column width="75%" %}
+### [Simpler.Grants.Gov](http://simpler.grants.gov) is now the default for searching for opportunities.
+
+With Simpler Search launching as the default experience, we can get a wider array of users providing feedback about their experience. It also solidifies Simpler.Grants.Gov as the standard for search and helps to iteratively build and deliver working software methodically. Just over the last week, we have seen a 200% increase in users visiting the search page.&#x20;
+
+Users can still access the classic search by first logging into Grants.gov and then searching for opportunities. Some funding opportunities may be removed or modified by the agency that created them, and are no longer accessible from any Grants.gov systems (Classic or Simpler). These changes are unrelated to the switch over to the Simpler Search.
+{% endcolumn %}
+{% endcolumns %}
+
+{% columns %}
+{% column width="25%" %}
+#### New Feature
+{% endcolumn %}
+
+{% column width="75%" %}
+### Activity history is now available on applications.
+
+The new activity history section of the application page shows what changes have been made to an application, when they were made, and who made each change. This surfaces important information to applicants collaborating on applications.
+{% endcolumn %}
+{% endcolumns %}
+
+{% columns %}
+{% column width="25%" %}
+#### New Improvement
+{% endcolumn %}
+
+{% column width="75%" %}
+### A clearer, persistent confirmation after applying.
+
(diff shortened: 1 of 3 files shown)


api/tests/lib/seed_orgs_and_users.py
```diff
@@ -210,6 +211,51 @@ def _build_organizations_and_users(
 
     user_scenarios.append("many_app_user - Has many applications across many orgs")
 
+    ########################
+    # Legacy users
+    ########################
+    # Available members
+    for i, org in enumerate([org1, org2, org3], start=1):
+        create_legacy_user_with_status(
+            uei=f"AVAILABLE_UEI_{i}",
```
chouinar: Wouldn't the UEI need to be the UEI of the org?

I'd recommend moving some of this newly added code to a separate function call as it's for a fairly specific user-scenario while this function is for general purpose users that we want to login as for testing.

babebe: That is correct! updated and moved to a helper func

api/tests/lib/seed_orgs_and_users.py
```diff
@@ -500,3 +510,52 @@ def handle_static_application_forms(application: Application, competition: Compe
         )
 
         validate_application_form(application_form, ApplicationAction.START)
+
+
+def _build_legacy_users_for_orgs(
+    orgs: list[Organization],
+    inviter: User,
+    user_scenarios: list[str],
```
chouinar: I'd leave the user_scenarios out of this, that is meant more for logging what users we created that you can login as, these wouldn't be those kinds of users.

api/tests/lib/seed_orgs_and_users.py
```diff
@@ -500,3 +510,52 @@ def handle_static_application_forms(application: Application, competition: Compe
...
+    """
+    # AVAILABLE legacy users
+    for i, org in enumerate(orgs, start=1):
+        create_legacy_user_with_status(
+            uei=org.sam_gov_entity.uei,
+            email=f"legacy_available_org{i}@example.com",
+            status=LegacyUserStatus.AVAILABLE,
+            organization=org,
+            first_name=f"Legacy{i}",
+            last_name="Available",
```
chouinar: Something to be aware of, if you define the emails/names like this and run the the seed script multiple times, you're going to see a lot of duplicate users since each time you run it, it'll make users with the exact same email/name every run. If that's fine, nothing to worry about, but just making sure you're aware.

babebe: Thanks for flagging. I’ve updated the emails to use Faker so we don’t create the same email on each seed run. Names may still repeat, but that’s ok for local testing.

api/tests/lib/seed_orgs_and_users.py
```diff
@@ -524,38 +525,36 @@ def _build_legacy_users_for_orgs(
     for i, org in enumerate(orgs, start=1):
         create_legacy_user_with_status(
             uei=org.sam_gov_entity.uei,
-            email=f"legacy_available_org{i}@example.com",
+            email=f"{faker.email()}@example.com",
```
chouinar: This would make `a@something.com@something.com`, the faker email function is already the formatted email.
```suggestion
            email=faker.email(),
```

PR #7683, opened 2025-12-24
[Issue #6980] ADR for SGM Data Flow

### LINKED ISSUE #6980: ADR for infrastructure setup and scaling
### Summary

ADR for managing data flow, in a secure, scalable way

TODO -- add rubric 

### Acceptance criteria

- [ ] ADR includes context and problem statement, decision drivers, options considered and decision outcome (with positive and negative consequences considered)
- [ ] ADR created in the project’s ADR directory.
- [ ] ADR reviewed and approved by stakeholders.
- [ ] ADR published in repository.

mdragon:
## Summary

Fixes #6980  

Adds an ADR describing the ways we've identified we'll access GS data directly from SGM

### DIFF
--- a/documentation/wiki/product/decisions/adr/2025-12-24-sgm-data-flow.md
+++ b/documentation/wiki/product/decisions/adr/2025-12-24-sgm-data-flow.md
@@ -8,13 +8,13 @@
 
 ## Context and Problem Statement
 
-Given the different approach for modernization of GrantSolutions we need an ongoing interoperability with the existing system, not a strangler, rebuild and replace pattern. To allow for that we need to have strategies for accessing data across the existing and modernized system that were not already designed and built for the Simpler Grants.gov work. How will we allow for bi-directional, near real-time, data flow between GrantSolutions (GS) and Simpler Grants Management (SGM).
+Given the different approach for modernization of GrantSolutions we need an ongoing interoperability with the existing system, not a strangler, rebuild and replace pattern. To allow for that we need to have strategies for accessing data across the existing and modernized system that were not already designed and built for the Simpler Grants.gov work. How will we allow for bi-directional, near real-time, data flow between GrantSolutions (GS) and Simpler Grants Management (SGM)?
 
 ## Decision Drivers
 
 - Optimize User Experience
-  - User's should be able to move between systems as seamlessly as possible
-  - Workflows or processes should move from system to system as needed without long delays or user intervention
+  - Users should be able to move between systems as seamlessly as possible
+  - Automated processes should proceed in both systems as needed without long delays or user intervention
 - Minimize time data is not in sync
   - Hourly is too slow for some data, but may be fine for other data
 - Minimize needless data duplication
@@ -49,7 +49,7 @@ For this ADR we are not picking a single option, but rather identifying everythi
 
 ### Bulk data copy on a scheduled basis
 
-This is the approach we took on Simpler Grants.gov. Hourly, we Extract, Load, and Transform (ELT) all of the table data we need from Grants.gov's database into Simpler Grants.gov's database. Whenever possible we only pull records updated since the last run to minimize data volume and the associated load on the existing system. We do not currently send data back to Grants.gov's database but in the SGM work that would be a requirement as well. We would modify our existing processes to support bi-directional data transfer. We could also consider improving the existing code base to allow it to run more frequently without collision and add more filtering to avoid when we fetch rows that we didn't see their FK records and so we fail to create the records (we could just only process something if we've already seen the parent record this run).
+This is the approach we took on Simpler Grants.gov. Hourly, we Extract, Load, and Transform (ELT) all of the table data we need from Grants.gov's database into Simpler Grants.gov's database. Whenever possible we only pull records updated since the last run to minimize data volume and the associated load on the existing system. We do not currently send data back to Grants.gov's database but in the SGM work that would be a requirement as well. We would modify our existing processes to support bi-directional data transfer. We could also consider improving the existing code base to allow it to run more frequently without collision and address some of the nuisance alters that self resolve on the following sync.
 
 - **Pros**
   - Simple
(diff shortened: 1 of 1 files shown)


mdragon:
Not sure why the check is failing, the file it's mad about is linked in the Summary, just like all the other pages Brandon added along with it...

documentation/wiki/product/decisions/adr/2025-12-24-sgm-data-flow.md
```diff
@@ -0,0 +1,95 @@
...
+
+- **Status:** Active
+- **Last Modified:** 2025-12-24
+- **Related Issue:** #6980
+- **Deciders:** Matt, Dani, Jay, Julius, Lucus
+- **Tags:** sgm, data
+
+## Context and Problem Statement
+
+Given the different approach for modernization of GrantSolutions we need an ongoing interoperability with the existing system, not a strangler, rebuild and replace pattern. To allow for that we need to have strategies for accessing data across the existing and modernized system that were not already designed and built for the Simpler Grants.gov work. How will we allow for bi-directional, near real-time, data flow between GrantSolutions (GS) and Simpler Grants Management (SGM).
```
doug-s-nava: Assuming this last sentence is a question

```
How will we allow for bi-directional, near real-time, data flow between GrantSolutions (GS) and Simpler Grants Management (SGM)?
```

documentation/wiki/product/decisions/adr/2025-12-24-sgm-data-flow.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Decision Drivers
+
+- Optimize User Experience
+  - User's should be able to move between systems as seamlessly as possible
```
doug-s-nava:
```suggestion
  - Users should be able to move between systems as seamlessly as possible
```

documentation/wiki/product/decisions/adr/2025-12-24-sgm-data-flow.md
```diff
@@ -0,0 +1,95 @@
...
+  - Workflows or processes should move from system to system as needed without long delays or user intervention
```
doug-s-nava: does this mean the workflows themselves should move, as in a workflow defined in simpler should be able to be accessed from grant solutions? Or is this more that as part of a user's "workflow" they should be able to track data between the two systems? "Workflow" being a tricky term I think this needs some clarification

documentation/wiki/product/decisions/adr/2025-12-24-sgm-data-flow.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Negative Consequences
+
+- n/a
+
+## Pros and Cons of the Options
+
+### Bulk data copy on a scheduled basis
+
+This is the approach we took on Simpler Grants.gov. Hourly, we Extract, Load, and Transform (ELT) all of the table data we need from Grants.gov's database into Simpler Grants.gov's database. Whenever possible we only pull records updated since the last run to minimize data volume and the associated load on the existing system. We do not currently send data back to Grants.gov's database but in the SGM work that would be a requirement as well. We would modify our existing processes to support bi-directional data transfer. We could also consider improving the existing code base to allow it to run more frequently without collision and add more filtering to avoid when we fetch rows that we didn't see their FK records and so we fail to create the records (we could just only process something if we've already seen the parent record this run).
```
doug-s-nava: what are `FK records` here?

chouinar: I'd probably leave out the detail about filtering based on the foreign key issue, while that is an annoying issue, it's probably too specific to call out here.

I will say that our current approach is heavily reliant on grants.gov's method of storing data in their DB (having created/updated timestamps), it wouldn't be possible to replicate if SGM doesn't have that working properly.

Also, bidirectional might not work well at all with our current approach, not only would writing back potentially put us in an endless loop (we write an update to legacy, which now has an updated timestamp and we pull it, and then process that update, and now we have an update to write back to legacy ... ).

Batch processing is likely something we'll want where timeliness isn't a concern, but we might need to start at least partially fresh depending on how SGM works.

documentation/wiki/product/decisions/adr/2025-12-24-sgm-data-flow.md
```diff
@@ -0,0 +1,95 @@
...
+### Call GS APIs directly as needed (no additional data points tracked)
+
+Where existing GS APIs exist we'll call those. We will have to figure out how API keys/permissions would work in this scenario. We may wrap these in our existing API, or just call them directly from the NextJS API "backend" server side routes. This will be the most streamlined way to build specific screens out that are able to be integrated into existing GS usage.
+
+- **Pros**
+  - Data is always fresh, exactly as if the user was still in GS
+  - Data written back via existing APIs has all of the consistency and
+- **Cons**
+  - Still learning what data has existing APIs that will make this possible
+
```
chouinar, line 72: What happens if a call fails? A con of using APIs might be that we end up in a weird state because an API call failed, but we don't have a way pipe that through? There's also some level of complexity about how to identify that two records across systems representing the same thing, opportunities have 3 different IDs (legacy integers, our UUIDs, and opportunity number which isn't unique, but often what other systems use). Even if we have IDs in sync, those IDs might not even exist to connect records (a legacy integer opportunity ID can't be created anywhere BUT grants.gov adding an entire dependency there).

documentation/wiki/product/decisions/adr/2025-12-24-sgm-data-flow.md
```diff
@@ -0,0 +1,95 @@
...
+### Call GS APIs directly as needed, store additional data points in SGM (without duplicating existing data)
+
+We won't be able to always mutate the GS data model as quickly as we'd want to iterate on SGM. In those cases we would store new fields in the Simpler DB, with the identifier of the record in GS. This would allow the data from both systems to be pulled together either in the API layer or in the FE via 2 API calls depending on whether we're wrapping API calls to GS in the Simpler API.
```
chouinar: I'd say we should keep the frontend from being aware of SGM as much as possible, since the problem is largely a data issue, having data be "merged" in the API itself makes more sense. It also means if something is only in our system or across two systems, the frontend (and user experience), doesn't really need to be aware of that.

documentation/wiki/product/decisions/adr/2025-12-24-sgm-data-flow.md
mdragon: Even more convoluted, I was wondering if initially we would want to utilize GS to get Opportunity data into Grants.gov since those are up and running integrations, with specific keys and permissions, etc.

mdragon:
HHS team approval noted by Dani via email

PR #7688, opened 2025-12-26
Follow-up to PR#7435: stabilize e2e stagging tests

Bhavna-Ramachandran:
## Summary

This PR is branched from PR#[7435 ](https://github.com/HHS/simpler-grants-gov/pull/7435) (branch: e2e-staging) to layer additional changes without modifying the original work directly.

### DIFF
--- a/frontend/tests/e2e/opportunity.spec.ts
+++ b/frontend/tests/e2e/opportunity.spec.ts
@@ -148,12 +148,7 @@ test("can navigate to grants.gov", async ({ page, context }) => {
   await page.getByRole("button", { name: "View on Grants.gov" }).click();
 
   const newPage = await newTabPromise;
-<<<<<<< HEAD
   expect(newPage.url()).toContain(
     "https://test.grants.gov/search-results-detail/",
-=======
-  await expect(newPage).toHaveURL(
-    `https://test.grants.gov/search-results-detail/${testOpportunityIdMap[targetEnv].legacy}`,
->>>>>>> 1727bf55f (further refinements for broken tests)
   );
 });


api/tests/lib/seed_local_db.py
```diff
@@ -69,9 +70,20 @@ def _build_opportunities(
         )
 
         # hardcoded id for e2e usage
-        factories.OpportunityFactory.create(
-            has_long_descriptions=True, opportunity_id="6a483cd8-9169-418a-8dfb-60fa6e6f51e5"
-        )
+        # Check if it already exists before creating to allow running seed multiple times
```
doug-s-nava, line 74: good call, and an important improvement on my solution. I think @chouinar may have another way of creating a record that will work around this issue though - might be nice to make sure we're using the same pattern everywhere

doug-s-nava (commented):
if bumping up the timeouts is all we need to do to reliably get tests passing in staging then I'm all for it. However! Is there a way to make the timeouts dynamic based on whether we are running against staging versus local? In cases where we are running against a local server it would be nice to keep the lower timeout to keep run times down.

Can you see if that's possible and make sure that the seed script check for existing records is in line with what is happening elsewhere?

api/tests/lib/seed_local_db.py
Bhavna-Ramachandran: I noticed there were errors in this part specifically, possibly due to parallel sessions running in different shards, so I updated this section to check if the opportunity already exists before creating it. This makes the seed script safe to run multiple times without causing duplicates or integrity errors. The has_long_descriptions=True record is still created for E2E tests if it doesn’t exist, but now it’s done safely, so concurrent runs or repeated seeds won’t fail.

We can check with @chouinar to see if his workaround helps avoiding conflicts or we could consider applying the same pattern across other seed data to keep everything consistent.

chouinar: I don't think the rollback would work, you didn't start a transaction so it wouldn't know what to rollback to / once an integrity error is hit, the DB session won't let you use it anymore generally. Even if you setup a guardrail properly, you'd need to do that for a dozen more such static data values elsewhere in the script - this script isn't and can't reasonably be made safe to run in parallel, if you want that support, we're going to need to start from scratch on this.

At best you can check if it exists, and create it if it doesn't, anything more than that is not going to work in this setup.

Is it possible to run the seed script before splitting the tests into parallel batches?

doug-s-nava: Currently each shard sets up its own DB. Not saying that's a good thing, but in this case it should mean that we don't need to worry about the issue of records being recreated. Can we move forward with this if we replace the rollback call with an error log? If we're still seeing errors in that case we can revisit, but I don't logically see how that would be happening

chouinar: If each shard sets up its own DB, then we wouldn't need an error message, as that wouldn't be an error case. it would just be a "don't create it if it already exists" sort of check that we have a few dozen of already. That'd just be:

```py
existing_e2e_opportunity_id = "..."
existing_e2e_opportunity = db_session.get(Opportunity, existing_e2e_opportunity_id)
if existing_e2e_opportunity is None:
    OpportunityFactory.create(opportunity_id=existing_e2e_opportunity_id, ...)
```

Issue #6950, opened 2025-11-04
Connect the agency in the cert to the GetSubmissionListExpanded filter


chris-kuryak:
### Summary

When we query the GetSubmissionListExpanded, we need to add always add a filter for the agency code to the query, even if the user didn't supply a filter. This way an agency can only see what they own.

The agency should be attached to the cert that we would have processed in prior tickets. We should always add a `where opportunity.agency_code = 'XYZ'` to every request, BUT if the agency is marked `is_multilevel` (in the agency table) then we want to have that instead be `where opportunity.agency_code = 'XYZ' OR opportunity.agency_code LIKE 'XYZ-%'`, adding an OR LIKE. Make sure to add `-%` to the end of the agency code, it has to have the dash to make sure it's a subagency and not another agency at the same level (eg. avoiding `XYZA` but getting `XYZ-ABC`).

### Acceptance criteria
- [ ] The agency code from the cert agency is automatically added to the query in our GetSubmissionListExpanded logic
- [ ] Consider making some of these filter setups generic and reusable as we'll need this same pattern in a few endpoints

PR #7689, opened 2025-12-26
[Issue #6950] Enforce agency/sub-agency filter in GetSubmissionListExpanded

babebe:
## Summary

Fixes / Work for #6950  

## Changes proposed

Added support for multi-level agencies so parent agencies can view submissions owned by their sub-agencies.
Added  test to validate agency scoping for both single-level and multi-level agencies.
## Context for reviewers
Previously, GetSubmissionListExpanded always applied an agency code filter, but it was limited to the requesting agency’s exact agency_code. This change expands the agency scoping logic to support multi-level agencies. When the authenticated agency is marked as multi-level, the query now includes submissions owned by both the parent agency and its sub-agencies.

## Validation steps
Requests without an agency filter for single-level agencies: Returned submissions are limited to the requesting agency’s ownership
Requests without an agency filter for multi-level agencies: Multi-level agencies receive both parent and sub-agency submission


jakobpederson (approved):
LGTM

Issue #6664, opened 2025-10-08
Merge the two responses together for GetSubmissionListExpanded


chris-kuryak:
### Summary

Connect with Mike H since he has become an XML wizard recently.

This ticket to merge the two responses together
- Return all submissions, even if an application has been submitted multiple times

From Matt D: We are moving toward streaming the response, so this could be a potential hazard to be aware of.

### Acceptance criteria
- [ ] Solution is configurable by endpoint, but only GetSubmissionListExpanded is operational
- [ ] Unit Tests to confirm a successful merge of the two data sources
- [ ] Solution does not break the current functionality or meaningfully change anything

PR #7699, opened 2025-12-29
[Issue #6664] Merge two GetSubmissionListExpanded responses

jakobpederson:
## Summary

Fixes / Work for #6664 

## Changes proposed

Merge the two responses together for GetSubmissionListExpanded. Take the proxy response, and extract out the application byte data and inject that into a new mtom message with the simpler data added in too.

## Context for reviewers

We want when people to hit the GetSubmissionListExpanded endpoint that they get a single response with data from both the proxy and the simpler db.

### DIFF
--- a/api/src/legacy_soap_api/legacy_soap_api_client.py
+++ b/api/src/legacy_soap_api/legacy_soap_api_client.py
@@ -270,8 +270,6 @@ class SimplerGrantorsS2SClient(BaseSOAPClient):
         yield b"\n" + boundary.encode("utf-8") + b"--"
 
     def get_simpler_soap_response(self, proxy_response: SOAPResponse) -> SOAPResponse:
-        if proxy_response.status_code != 500 and not self.operation_config.has_merge_list_response:
-            return proxy_response
         if self.operation_config.is_mtom is False:
             return proxy_response
         # MTOM message is assembled here
--- a/api/src/legacy_soap_api/legacy_soap_api_config.py
+++ b/api/src/legacy_soap_api/legacy_soap_api_config.py
@@ -76,7 +76,6 @@ class SOAPOperationConfig:
     compare_endpoints: bool = False
     is_mtom: bool = False
     always_call_simpler: bool = False
-    has_merge_list_response: bool = False
 
     # These are the privileges needed for these endpoints:
     # GetSubmissionListExpandedRequest = {Privilege.LEGACY_AGENCY_VIEWER}
@@ -147,7 +146,6 @@ SIMPLER_SOAP_OPERATION_CONFIGS: dict[SimplerSoapAPI, dict[str, SOAPOperationConf
             request_operation_name="GetSubmissionListExpandedRequest",
             response_operation_name="GetSubmissionListExpandedResponse",
             is_mtom=True,
-            has_merge_list_response=True,
             privileges={Privilege.LEGACY_AGENCY_GRANT_RETRIEVER},
         ),
     },
--- a/api/tests/src/legacy_soap_api/test_legacy_soap_api_client.py
+++ b/api/tests/src/legacy_soap_api/test_legacy_soap_api_client.py
@@ -553,35 +553,6 @@ class TestSimplerSOAPGetApplicationZip:
         assert response.data == mock_proxy_response.data
         assert response.status_code == mock_proxy_response.status_code
 
-    def test_get_simpler_soap_response_returns_proxy_response_if_proxy_response_status_code_is_not_500(
-        self, db_session
-    ):
-        request_xml_bytes = (
-            '<soapenv:Envelope xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/" xmlns:agen="http://apply.grants.gov/services/AgencyWebServices-V2.0" xmlns:gran="http://apply.grants.gov/system/GrantsCommonElements-V1.0">'
-            "<soapenv:Header/>"
-            "<soapenv:Body>"
-            "<agen:GetApplicationZipRequest>"
-            f"<gran:GrantsGovTrackingNumber>{GRANTS_GOV_TRACKING_NUMBER}</gran:GrantsGovTrackingNumber>"
-            "</agen:GetApplicationZipRequest>"
-            "</soapenv:Body>"
-            "</soapenv:Envelope>"
-        ).encode("utf-8")
-        soap_request = SOAPRequest(
-            data=request_xml_bytes,
-            full_path="x",
-            headers={},
-            method="POST",
-            api_name=SimplerSoapAPI.GRANTORS,
-            operation_name="GetApplicationZipRequest",
-        )
-        mock_proxy_response = SOAPResponse(data=b"soap", status_code=200, headers={})
-        with patch.object(uuid, "uuid4") as mock_uuid4:
-            mock_uuid4.side_effect = [CID_UUID, BOUNDARY_UUID]
-            client = SimplerGrantorsS2SClient(soap_request, db_session)
-            result = client.get_simpler_soap_response(mock_proxy_response)
-            assert result.data == mock_proxy_response.data
-            assert result.status_code == mock_proxy_response.status_code
-
     def test_get_simpler_soap_response_returns_proxy_response_if_is_mtom_is_false_on_operation_config(
         self, db_session
     ):


api/src/legacy_soap_api/legacy_soap_api_client.py
```diff
@@ -265,9 +266,12 @@ def _gen_response_data(
         yield b"\n" + boundary.encode("utf-8") + b"--"
 
     def get_simpler_soap_response(self, proxy_response: SOAPResponse) -> SOAPResponse:
-        if proxy_response.status_code != 500:
+        if (
+            proxy_response.status_code != 500
+            and self.operation_config.response_operation_name != "GetSubmissionListExpandedResponse"
```
chouinar: I'm thinking instead of hard-coding the operation name for doing the merge, we add another configuration value like `has_merge_list_response` to the operation config  - we can check the operations below to map to the appropriate function.

api/src/legacy_soap_api/legacy_soap_api_client.py
```diff
@@ -265,9 +266,12 @@ def _gen_response_data(
         yield b"\n" + boundary.encode("utf-8") + b"--"
 
     def get_simpler_soap_response(self, proxy_response: SOAPResponse) -> SOAPResponse:
-        if proxy_response.status_code != 500:
+        if (
+            proxy_response.status_code != 500
```
chouinar: Just double-checking, if there are no results in the legacy SOAP endpoint, is that returned as just an empty list, OR as a 500?

api/src/legacy_soap_api/soap_payload_handler.py
```diff
@@ -1,14 +1,23 @@
...
+from typing import TYPE_CHECKING, Any
 
 import xmltodict
 from lxml import etree
 from lxml.etree import Element, QName, SubElement
 
+from src.legacy_soap_api.legacy_soap_api_constants import LegacySoapApiEvent
+
+if TYPE_CHECKING:
+    from src.legacy_soap_api.legacy_soap_api_schemas import SOAPResponse
```
chouinar: Guessing this is due to a circular dependency between that file and this one? Can you move `SOAPResponse` to its own file to work around that?

api/src/legacy_soap_api/soap_payload_handler.py
```diff
@@ -389,3 +398,136 @@ def get_prefix_and_local_name(key: str) -> tuple:
...
+                yield chunk[start_idx:]
+        else:
+            yield chunk
+
+
+class GeneratorStream(io.RawIOBase):
+    """
+    This takes the response stream and converts into an object
+    that is file-like that can be used by Python tools
+    """
```
chouinar: Could you add a bit more detail for what this stream is doing / how it differs from whatever basic stream implementations exist in the IO libraries?

api/src/legacy_soap_api/soap_payload_handler.py
```diff
@@ -389,3 +398,136 @@ def get_prefix_and_local_name(key: str) -> tuple:
...
+) -> Generator[bytes]:
+    response_data = input_data.get("Envelope", {}).get("Body", {})
+    simple_data = response_data.get("ns2:GetSubmissionListExpandedResponse", {}).get(
+        "ns2:SubmissionInfo", {}
+    )
+
+    boundary = f"uuid:{raw_uuid}"
+    yield (
+        f"--{boundary}\n"
+        f'Content-Type: application/xop+xml; charset=UTF-8; type="text/xml"\n'
```
chouinar: Could we statically define some of this up-top just to keep the function shorter?

api/src/legacy_soap_api/soap_payload_handler.py
```diff
@@ -389,3 +398,136 @@ def get_prefix_and_local_name(key: str) -> tuple:
...
+        reader = io.BufferedReader(GeneratorStream(clean_mtom_generator(proxy_data.stream())))
+        context = etree.iterparse(reader, events=("end",))
+        try:
+            yield from stream_expanded_submissions_response(context, len(simple_data))
+        except etree.XMLSyntaxError:
+            logger.info(
+                msg="Could not parse proxy xml response",
+                exc_info=True,
+                extra={"soap_api_event": LegacySoapApiEvent.UNPARSEABLE_SOAP_PROXY_RESPONSE},
+            )
```
chouinar: Should this be an error / would we want to be made aware of it / consider it a bug if it happened?

api/src/legacy_soap_api/soap_payload_handler.py
```diff
@@ -389,3 +398,136 @@ def get_prefix_and_local_name(key: str) -> tuple:
...
+            yield simple_count_response
```
chouinar: If we were partially through the `yield from` above, would this be a valid count? Could we even make something valid if halfway through the stream data was bad?

api/src/legacy_soap_api/soap_payload_handler.py
```diff
@@ -389,3 +398,136 @@ def get_prefix_and_local_name(key: str) -> tuple:
...
+        'xmlns:ns3="http://apply.grants.gov/system/GrantsTemplate-V1.0" '
+        'xmlns:ns2="http://apply.grants.gov/services/AgencyWebServices-V2.0" '
+        'xmlns="http://apply.grants.gov/system/GrantsCommonElements-V1.0">'
+        "<ns2:Success>true</ns2:Success>"
+    ).encode("utf-8")
+
+    simple_count_response = f"<ns2:AvailableApplicationNumber>{len(simple_data)}</ns2:AvailableApplicationNumber>".encode(
+        "utf-8"
+    )
+    if proxy_data.status_code == 200:
```
chouinar: Nit:

If you have code like:

```
if X:
    # a
    # bunch
    # of
    # code
    return 1
else:
    return 0
```

Just short-circuit it and do:

```
if not X:
   return 0

# whatever
return 1
```

Keeps indents from being a bit much.

api/src/legacy_soap_api/legacy_soap_api_client.py
jakobpederson: If there's zero applications found it returns a 200 with `<ns2:AvailableApplicationNumber>0</ns2:AvailableApplicationNumber>`

api/src/legacy_soap_api/soap_payload_handler.py
jakobpederson: I was thinking that we would want to know if the incoming xml from the proxy was invalid, especially since right now it really messes with the count.

api/src/legacy_soap_api/soap_payload_handler.py
jakobpederson: The count would not be valid and we couldn't make something valid from the stream if halfway bad. I'm trying out a solution that uses NamedTemporaryFiles in order to extract what we can.

api/src/legacy_soap_api/soap_payload_handler.py
jakobpederson: I've removed this for now assuming the NamedTemperaryFile approach is going to work.

api/src/legacy_soap_api/legacy_soap_api_client.py
chouinar: Got it, just wanted to make sure if we were merging `[]` + `[a, b, c]`, it would return `[a, b, c]`.

api/src/legacy_soap_api/legacy_soap_api_client.py
chouinar: We'd want that check here, not checking a specific operation.

Down below we could check a specific operation to map to the proper function for doing the merge.

api/src/legacy_soap_api/soap_payload_handler.py
```diff
@@ -389,3 +416,98 @@ def get_prefix_and_local_name(key: str) -> tuple:
...
+        "</soap:Body>"
+        "</soap:Envelope>\n"
+        f"--{boundary}--"
+    ).encode("utf-8")
+
+
+def get_proxy_submission_data(
+    proxy_data: SOAPResponse, simple_count_response: bytes, simple_data_count: int
+) -> Generator[bytes]:
+    with tempfile.NamedTemporaryFile(mode="w+b") as tmp:
```
chouinar: Weren't we having an issue with tempfiles for the certs? Do we need to write to a file for this? This is just XML, isn't it? What stops us from just parsing it as text (MTOM file stuff aside, which I'd say leave alone as much as possible).

I'm having a hard time following how the parsing/merging logic here works as it's a lot of buffers/ streams/files without any explanation ("the data we're processing here looks like"). 

Stepping back, I assumed the implementation of this would be roughly:
* If ours is empty, return proxy
* If proxy is empty, return ours
* Using ours as a baseline, grab whatever is in the proxy, parse it (converting to not XML to do so), and add it to the XML we already have (potentially doing all of this not as XML, we could just not generate the XML for ours until this point potentially).

If we don't even generate the XML until this point, that might let us avoid needing so many regexes and parsers. 

That said, I think that would require the legacy response to get passed to our response creation logic to do that.

api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
```diff
@@ -171,12 +184,41 @@ def get_submission_list_expanded_response(
...
+        parser = etree.XMLParser(recover=True)
+        root = etree.fromstring(xml_bytes, parser=parser)
+        for element in root.iter("{*}SubmissionInfo"):
+            try:
+                element_bytes = etree.tostring(element)
+                submission_info_dict = xmltodict.parse(element_bytes).get("ns2:SubmissionInfo")
+                info.append(schemas.SubmissionInfo(**submission_info_dict))
+            except PydanticValidationError:
+                logger.exception("Skipping invalid submission due to validation error")
+                continue
```
chouinar, line 218: I'd probably put this into a function and then here we just do (feel free to adjust naming):

```py
proxy_submissions = convert_proxy_response(proxy_response)

info.extend(proxy_submissions)
```

Mostly because I think we might want a try block around this conversion. Just in case our XML parsing hits an issue, we want to at least be able to return our response and log the error (the "the whole XML is bad" scenario).

api/src/legacy_soap_api/legacy_soap_api_client.py
```diff
@@ -60,6 +60,10 @@ def get_soap_response_dict(self) -> dict:
         operation_method = getattr(
             self, to_snake_case(self.operation_config.request_operation_name)
         )
+        if self.operation_config.has_merge_list_response:
```
chouinar: Do we still need this with how the approach was adjusted? I think a few of the file adjustments you made probably aren't needed anymore as well.

jakobpederson: I took it out and it seems like it works fine, so I'll remove it.

chouinar: Is the `has_merge_list_response` param needed at all anymore? I don't think it is since that's now just built into the logic of making the response, I think you can remove all references of it / the config that was added.

jakobpederson: My concern is 
```
        if proxy_response.status_code != 500 and not self.operation_config.has_merge_list_response:
            return proxy_response
```
Because without that, the proxy_response would take priority and we'd never get to a merged response from what I can tell.

jakobpederson: Oh wait, maybe you're right. I'm going to check.

jakobpederson: I've removed it. I don't think we need that 500 check since we've got the `use_simpler` on the config.

Issue #7557, opened 2025-12-18
Create an endpoint for giving users an internal role


chouinar:
### Summary

We want an endpoint that allows one of our internal users to be able to give another internal user an internal role. Right now to do this we have to login to the DB and follow these instructions: https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2372730985/Form+Management#Setup-an-internal-admin-user

Effectively, we want that DB query set in an API endpoint.

This endpoint will be `PUT /v1/internal/roles`. We'll need to create a new "internal" blueprint and connect it to our API so it can be called.

### Request
```json
{
  "internal_role_id": "<uuid>",
  "user_email": "example@example.com"
}
```
Note - importantly, do not log the users email.

### Response
A generic success message, no response body needed.
```json
{
  "message": "Success",
  "data": null
}
```

### Auth
This endpoint will only accept API User Key Auth (no JWT auth needed at this time).

The user must have the `manage_internal_roles` privilege, a new privilege we'll want to create AND add to the internal user role that we have defined in https://github.com/HHS/simpler-grants-gov/blob/main/api/src/constants/static_role_values.py

We should use our usual logic for checking that a user has access (see the update form instruction endpoint which similarly checks a different internal role privilege), this will 403 if they don't.

### Logic
Query the role table with the internal role ID passed in filtering roles to just those that are internal roles (this cannot give users non-internal roles). If nothing is found, 404.

Query the link_external_user table by email. If no user is found with that email, 404. If more than one user is found (which can happen, although shouldn't), we should also error with a 422.

Remember that we have our `raise_flask_error` util for raising errors in the appropriate format.

If we find the role and one user, we want to add the role to the user. if the user already has the role, it should raise a 422 with an appropriate error message.

Otherwise, add the role to the user.

### Acceptance criteria

- [ ] Endpoint added
- [ ] Tests added for the endpoint testing basic functionality, error cases, and auth errors

PR #7704, opened 2025-12-29
[Issue #7557] Create internal role endpoint

kkrug:
## Summary

Fixes for #7557 

## Changes proposed

Created an endpoint for adding an internal role to a user.
Added tests to check endpoint for success and proper error responses.

## Context for reviewers

This is an endpoint that allows one of our internal users to be able to give another internal user and internal role (this was previously only possible via DB login). The new endpoint is PUT /v1/internal/roles. A new manage_internal_roles privilege was created. The endpoint only accepts API UserKey Auth. 

## Validation steps

Confirm the endpoint works as intended via Swagger. These are some helpful SQL commands for testing:

1. Confirm an admin user with the proper privilege (id=15) exists - be sure to note the returned email address:
`SELECT 
    ak.key_id AS your_api_key,
    leu.email AS admin_email,
    p.privilege_id
FROM api.user_api_key ak
JOIN api.link_external_user leu ON ak.user_id = leu.user_id
JOIN api.internal_user_role iur ON ak.user_id = iur.user_id
JOIN api.link_role_privilege lrp ON iur.role_id = lrp.role_id
JOIN api.lk_privilege p ON lrp.privilege_id = p.privilege_id
WHERE ak.is_active = true 
  AND p.privilege_id = 15;`
2. Get and email to use as a target (first part of payload for request):
`SELECT email FROM api.link_external_user WHERE email != 'EMAIL_FROM_FIRST_QUERY@example.net' LIMIT 1;`
3. Get the role ID of an internal role (second part of payload request):
`SELECT r.role_id FROM api.role r 
JOIN api.link_role_role_type lrrt ON r.role_id = lrrt.role_id 
WHERE lrrt.role_type_id = 3 LIMIT 1;`
4. After testing endpoint, check for success using target email from query 2:
`SELECT leu.email, r.role_name, r.role_id
FROM api.internal_user_role iur
JOIN api.role r ON iur.role_id = r.role_id
JOIN api.link_external_user leu ON iur.user_id = leu.user_id
WHERE leu.email = 'david65@example.org';`

### DIFF
--- a/api/src/api/internal/internal_routes.py
+++ b/api/src/api/internal/internal_routes.py
@@ -19,20 +19,21 @@ logger = logging.getLogger(__name__)
 @internal_blueprint.output(internal_schema.InternalRoleAssignmentResponseSchema)
 @internal_blueprint.auth_required(api_user_key_auth)
 @flask_db.with_db_session()
-def update_internal_roles(db_session: db.Session, json_data: dict) -> response.ApiResponse:
-    internal_role_id = str(json_data.get("internal_role_id"))
-    user_email = str(json_data.get("user_email"))
+def update_internal_roles(
+    db_session: db.Session, json_data: dict[str, str]
+) -> response.ApiResponse:
+    internal_role_id = json_data.get("internal_role_id", "")
+    user_email = json_data.get("user_email", "")
 
     add_extra_data_to_current_request_logs({"internal_role_id": internal_role_id})
     logger.info("PUT /v1/internal/roles")
 
-    user = api_user_key_auth.get_user()
-    user = db_session.merge(user, load=False)
+    with db_session.begin():
+        user = api_user_key_auth.get_user()
+        user = db_session.merge(user, load=False)
 
-    verify_access(user, {Privilege.MANAGE_INTERNAL_ROLES}, None)
+        verify_access(user, {Privilege.MANAGE_INTERNAL_ROLES}, None)
 
-    update_internal_user_role(db_session, internal_role_id, user_email)
-
-    db_session.commit()
+        update_internal_user_role(db_session, internal_role_id, user_email)
 
     return response.ApiResponse(message="Success")
--- a/api/src/services/internal/update_internal_user_role.py
+++ b/api/src/services/internal/update_internal_user_role.py
@@ -43,8 +43,7 @@ def update_internal_user_role(
         raise_flask_error(422, "User already has this role.")
 
     # Add the new internal role assignment
-    with db_session.begin_nested():
-        new_assignment = InternalUserRole(user_id=target_user_id, role_id=internal_role_id)
-        db_session.add(new_assignment)
+    new_assignment = InternalUserRole(user_id=target_user_id, role_id=internal_role_id)
+    db_session.add(new_assignment)
 
     logger.info(f"Successfully assigned role {internal_role_id} to user {user_email}")


api/src/services/internal/update_internal_user_role.py
```diff
@@ -0,0 +1,53 @@
...
+def update_internal_user_role(
+    db_session: db.Session, internal_role_id: str, user_email: str
+) -> None:
+    # Query the role table and verify it is an internal role
+    role = (
+        db_session.query(Role)
+        .join(LinkRoleRoleType, Role.role_id == LinkRoleRoleType.role_id)
+        .filter(Role.role_id == internal_role_id, LinkRoleRoleType.role_type == RoleType.INTERNAL)
+        .first()
+    )
```
jakobpederson, line 21: I think we prefer to use `select` rather than `query`. So this would look more like 
```
select(Role).where(Role.role_id==LinkRoleRoleType.role_id).scalar_or_none()
```

api/src/services/internal/update_internal_user_role.py
```diff
@@ -0,0 +1,53 @@
...
+
+    if existing_assignment:
+        raise_flask_error(422, "User already has this role.")
+
+    # Add the new internal role assignment
+    new_assignment = InternalUserRole(user_id=target_user_id, role_id=internal_role_id)
+    db_session.add(new_assignment)
+
+    db_session.commit()
```
jakobpederson: You'd probably want to wrap this with a context manager instead of directly committing, something like
`with db.session.begin()` then you don't need the `db_session.commit()`.

api/src/api/internal/internal_routes.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+@internal_blueprint.put("/roles")
+@internal_blueprint.input(internal_schema.InternalRoleAssignmentRequestSchema, location="json")
+@internal_blueprint.output(internal_schema.InternalRoleAssignmentResponseSchema)
+@internal_blueprint.auth_required(api_user_key_auth)
+@flask_db.with_db_session()
+def update_internal_roles(db_session: db.Session, json_data: dict) -> response.ApiResponse:
+    internal_role_id = str(json_data.get("internal_role_id"))
+    user_email = str(json_data.get("user_email"))
```
jakobpederson: Isn't this already a string?

api/src/api/internal/internal_routes.py
```diff
@@ -0,0 +1,35 @@
...
+
+    add_extra_data_to_current_request_logs({"internal_role_id": internal_role_id})
+    logger.info("PUT /v1/internal/roles")
+
+    user = api_user_key_auth.get_user()
+    db_session.add(user)
```
jakobpederson: What does adding a user here do?

api/src/api/internal/internal_routes.py
kkrug: yes but lint didn't like it without the str wrapper

api/src/api/internal/internal_routes.py
kkrug: It was needed because the user returned by api_user_key_auth.get_user() is detached from the current db_session due to the way verify_api_key creates its own temporary session for validation. I changed it to use db_session.merge(user) instead.

api/src/services/internal/update_internal_user_role.py
kkrug: I updated all instances of query to be select instead.

api/src/services/internal/update_internal_user_role.py
kkrug: I updated this to use db_session.begin_nested (begin doesn’t work because the transaction is already started by the route). Db_session_commit was added to the route to complete the transaction.

api/src/api/internal/internal_routes.py
jakobpederson: I think the typing of the `json:dict` is the source of the problem. If you define that then you don't have to cast the variables like above.
```
    db_session: db.Session, json_data: dict[str, str]
) -> response.ApiResponse:
    internal_role_id = json_data.get("internal_role_id", "")
    user_email = json_data.get("user_email", "")
```

api/src/api/internal/internal_routes.py
jakobpederson: I think you can move the begin up all the way to the routes file where you start doing db stuff and it seems to work:
```
    with db_session.begin():
        user = api_user_key_auth.get_user()
        user = db_session.merge(user, load=False)

        verify_access(user, {Privilege.MANAGE_INTERNAL_ROLES}, None)

        update_internal_user_role(db_session, internal_role_id, user_email)
```

api/src/api/internal/internal_routes.py
kkrug: Fixed

api/src/api/internal/internal_routes.py
kkrug: Fixed - this also allowed me to remove the db_session.begin_nested from the service layer

api/src/services/internal/update_internal_user_role.py
kkrug: db_session.being_nested was able to be removed after the db_session change in the route

Issue #7495, opened 2025-12-17
when an application is started as an organization, the proper event isn't published


ErinPattisonNava:
### Summary

When an application is started as being owned by an organization, two events should be published: the application created event and the organization added event.  (Similar to what happens when an application is started as being owned by an individual, where the two events published are application created and user added.)

### Reproducibility

I tried multiple times and saw the bug every time

### How to reproduce the bug

Start an application as an organization (log in as a user who is part of an organization, and chose the organization as part of the drop down on the apply modal after clicking "start new application button").  On the application page, scroll down to the history section.  You should see two events, not one.

<img width="1063" height="204" alt="Image" src="https://github.com/user-attachments/assets/a02491b9-d3d7-4918-928c-da19759242b1" />

### Browser

None

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #7714, opened 2025-12-30
[Issue #7495] Add organization added event on application creation

babebe:
## Summary

Fixes for #7495  
Ensure the correct audit events are published when an application is created with an organization owner.
## Changes proposed
Added logic to publish an ORGANIZATION_ADDED audit event when an application is created and owned by organization

## Context for reviewers
Per the issue requirements, when an application is started as being owned by an organization, two audit events should be published:
`APPLICATION_CREATED`
`ORGANIZATION_ADDED`
Previously, only the application created event was emitted.

## Validation steps

Create an application owned by an organization 
Verify that  both events exist in the application history page

### DIFF
--- a/api/src/api/internal/__init__.py
+++ b/api/src/api/internal/__init__.py
@@ -1,0 +1,6 @@
+from src.api.internal.internal_blueprint import internal_blueprint
+
+# import internal_routes module to register the API routes on the blueprint
+import src.api.internal.internal_routes  # noqa: F401 isort:skip
+
+__all__ = ["internal_blueprint"]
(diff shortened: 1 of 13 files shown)


api/src/services/applications/create_application.py
```diff
@@ -162,7 +162,13 @@ def create_application(
...
         audit_event=ApplicationAuditEvent.APPLICATION_CREATED,
     )
-
+    if organization_id:
+        add_audit_event(
+            db_session=db_session,
+            application=application,
+            user=user,
+            audit_event=ApplicationAuditEvent.ORGANIZATION_ADDED,
+        )
```
jakobpederson, line 171: Just a question: do we need to attach the organization info to the event audit?

babebe: We are currently not capturing that information here. Perhaps it could be added later in the metadata.

Issue #7380, opened 2025-12-08
Disable conditionally-required radio buttons in submitted applications


jesababon:
### Summary

**Current experience:** The conditionally-required forms section's yes/no radio buttons are editable on the frontend. Can currently be toggled to "No" and also does not allow for re-assigning to "Yes."

**Expected behavior:** After successful application submission, conditionally-required forms radio buttons cannot be changed, i.e. they are disabled. They should retain the "Yes" or "No" user choice from the successful application submission event.

**Screen video of current experience:**

https://github.com/user-attachments/assets/3755fb8b-deb1-4754-9300-87834eada52b

Designer expected behavior visual if helpful:

<img width="603" height="222" alt="Image" src="https://github.com/user-attachments/assets/54ad7e25-5521-4119-86d8-5739db1a1a16" />

### Reproducibility

None

### How to reproduce the bug

- Submit an application successfully
- Interact with the conditionally required forms radio buttons.

### Browser

None

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

chris-kuryak:
Tested successfully with the following applications when submitted with both "Yes" and "No" radio selections:

https://staging.simpler.grants.gov/workspace/applications/application/09117a59-e56d-4c08-b2fe-b290f965395d
https://staging.simpler.grants.gov/workspace/applications/application/2eb2872b-9be3-4fd9-bb1e-9078f92d2b72

Issue #7109, opened 2025-11-18
Create web and print views for Supplementary Cover Sheet for NEH Grant Programs Form


chris-kuryak:
### Summary

This work is to use the JSON from #7011 

Form information: https://www.grants.gov/forms/form-items-description/fid/530

- Do not include Burden Statement's that are linked via a button in the PDF.
- Include additional description/instructions below Project funding that states: "A non-zero amount is required for either Outright Funds or Federal Match."
- "Additional Funding" question should be able to select yes or no, not a single checkbox.

### Acceptance criteria

- [ ] Webform matches expected layout from PDF form, and PDF Instructions section and numbers align with webform
- [ ] Print View generates as expected
- [ ] Product and Design approved screenshots before PR review
- [x] Backend has created an Opportunity for testing this form and it is deployed to **staging**
- [x] Product and Design notified for UAT
- [x] Product UAT complete
  - [x] Webform
  - [x] PDF
- [ ] Design UAT complete
  - [ ] Webform
  - [ ] PDF

chris-kuryak:
Form UAT complete. Still awaiting PDF for review.

https://staging.simpler.grants.gov/workspace/applications/application/9745ce9e-3863-47b2-bae3-90644a584641

chris-kuryak:
PDF looks good.
Moving to done.

PR #7687, opened 2025-12-24
[Issue #7109] Supplementary Cover Sheet for NEH Grant Programs Form

myduong-navapbc:
## Summary

Work for #7109  

## Changes proposed

- [ ] Include additional description/instructions below Project funding that states: "A non-zero amount is required for either Outright Funds or Federal Match."

- [ ] "Additional Funding" question should be able to select yes or no, not a single checkbox.

screenshot of html page for docraptor
Screenshot 2025-12-24 at 4 36 23 PM

## Changes proposed

A bit of a unique situation. A non-zero amount is required for either Outright Funds or Federal Match. They are required but not really. 

My vote is to mark them both as required

## Validation steps

1. to run locally, you would need to `run npm run build && npm start`
2. `make remake-backend && docker compose up`
3. navigate to `search`
4. search for `test`
5. you should see an opportunity called `Test Opportunity for SupplementaryCoverSheetforNEHGrantPrograms 3.0`
6. sign in
7. start new application
8. navigate to form table of the application
9. click Supplementary Cover Sheet for NEH Grant Programs Form
10. VERIFY the form loads
11. fill out the form, save
12. run `make generate-internal-token` from `api` directory
13. grab the token and place in request header "X-SGG-Internal-Token" and the token (I use a plugin called Modheader)
14. navigate to http://localhost:3000/print/application/APP_ID/form/FORM_ID
15. VERIFY the form data saves correctly

PR #7716, opened 2025-12-30
[Issue #7380] Conditionally disable radio buttons based off status

hayeshorus:
### **Summary**
Work for https://github.com/HHS/simpler-grants-gov/issues/7109

### **Changes proposed**

 Updated logic when application status is submitted radio buttons are disabled

 "Submit with Application" radio should be conditionally disabled

screenshot of html page for docraptor
Screenshot 2025-12-30 at 11 24 31 AM

### **Changes proposed**

Added logic to conditionally disable radio buttons when application status is submitted

### **Validation** steps

1. Create a new application
2. Verify radio buttons are enabled when status is set to in progress.   
3. Select radio button to no or yes.  
4. Fill out CD511 if yes 
5. Submit application.  
6. Verify application status is set to submitted and radio buttons are disabled.


ErinPattisonNava:
Not sure what happened, but it seems that the PR template got confused?  Can you put it back?

It should list either fixes or work, not both.  That determines if UAT in needed.  For this PR, since it is a bug, UAT isn't required.

The name of the PR needs to follow `[Issue #xxxx] description` to look correct in our release notes, see example [here](https://github.com/HHS/simpler-grants-gov/pull/7687)

frontend/src/components/application/IncludeFormInSubmissionRadio.tsx
```diff
@@ -55,15 +57,15 @@ export const IncludeFormInSubmissionRadio = ({
     : includeFormInSubmission === false
       ? "No"
       : undefined;
-
+  const disabledValue = applicationStatus === "submitted" ? true : loading;
```
ErinPattisonNava: There are other states after submitted, see [here](https://github.com/HHS/simpler-grants-gov/blob/main/frontend/src/types/applicationResponseTypes.ts#L12).  So let's change this test to only be enabled if in_progress and use the enum?

frontend/src/components/application/ApplicationContainer.tsx
```diff
@@ -172,6 +172,7 @@ const ApplicationContainer = ({
         forms={forms}
         errors={validationErrors}
         competitionInstructionsDownloadPath={instructionsDownloadPath}
+        applicationStatus={applicationStatus}
```
ErinPattisonNava: Thoughts on passing down the applicationDetails object?  If we are going to do prop drilling, seems easier to send the larger object so for future changes it will be more useful.

frontend/src/components/application/IncludeFormInSubmissionRadio.test.tsx
```diff
@@ -19,6 +19,7 @@ jest.mock("next/navigation", () => ({
 describe("IncludeFormInSubmissionRadio", () => {
   const applicationId = "app-123";
   const formId = "form-456";
+  const applicationStatus = "submitted";
```
ErinPattisonNava, line 22: Can you add tests specifically to test the disabled state you added?

frontend/src/components/application/ApplicationContainer.tsx
hayeshorus: I agree it would be more useful.   Should that be another ticket?  Is that out of scope for this defect?  Or should I just go ahead and work on that change and include it in this PR?

frontend/src/components/application/IncludeFormInSubmissionRadio.test.tsx
hayeshorus: I have added the test.

frontend/src/components/application/ApplicationContainer.tsx
ErinPattisonNava: That's part of the fix here, so please include it in this PR.

frontend/src/components/application/ApplicationValidationAlert.test.tsx
```diff
@@ -84,7 +84,7 @@ describe("ApplicationValidationAlert", () => {
           },
           {
             field: "form_id",
-            value: "123e4567-e89b-12d3-a456-426614174001",
+            value: "123e4567-e89b-12d3-a456-426614174003",
```
ErinPattisonNava: Why did the value need to change?

frontend/src/components/application/ApplicationContainer.tsx
```diff
@@ -167,11 +168,8 @@ const ApplicationContainer = ({
       />
       <OpportunityCard opportunityOverview={opportunity} />
       <ApplicationFormsTable
-        applicationForms={applicationForms}
-        applicationId={applicationId}
-        forms={forms}
         errors={validationErrors}
-        competitionInstructionsDownloadPath={instructionsDownloadPath}
```
ErinPattisonNava, line 174: I don't think this is necessarily a good thing to refactor and have the translation to the download path living in two places.  Normally I'd say pull it out as a general utility, but in this case I'd just leave it as something you drill down through the props.

frontend/src/components/application/ApplicationFormsTable.tsx
```diff
@@ -69,12 +68,7 @@ export const ApplicationFormsTable = ({
   return (
     <>
       <h3>{t("requiredForms")}</h3>
-      <ApplicationTable
-        forms={forms}
-        applicationForms={requiredForms}
-        applicationId={applicationId}
-        formsAreOptional={false}
-      />
+      <ApplicationTable applicationDetailsObject={applicationDetailsObject} />
```
ErinPattisonNava: I believe you need to leave the `formsAreOptions={false}` here

frontend/src/components/application/ApplicationFormsTable.tsx
```diff
@@ -90,11 +84,8 @@ export const ApplicationFormsTable = ({
             })}
           </p>
           <ApplicationTable
-            forms={forms}
-            applicationForms={conditionalRequiredForms}
-            applicationId={applicationId}
-            formsAreOptional={true}
             errors={errors}
+            applicationDetailsObject={applicationDetailsObject}
```
ErinPattisonNava, line 97: And you'll need the `formsAreOptional={true}` here, as that is context specific and not covered in the applicaiton details object.

frontend/src/components/application/ApplicationFormsTable.tsx
```diff
@@ -129,18 +120,17 @@ const ApplicationTableColumnError = ({
...
+  applicationDetailsObject,
 }: {
-  applicationForms: ApplicationFormDetail[];
-  applicationId: string;
-  forms: CompetitionForms;
-  formsAreOptional: boolean;
   errors?: FormValidationWarning[] | null;
+  applicationDetailsObject: ApplicationDetail;
 }) => {
+  const formsAreOptional = false;
```
ErinPattisonNava: And this shouldn't have a default but be sent in

frontend/src/components/application/ApplicationValidationAlert.test.tsx
hayeshorus: I reverted.  I was getting errors before with duplicate ID's.   therefore the changes.  I have reverted.

frontend/src/components/application/ApplicationContainer.tsx
hayeshorus: I have reverted the changes and fixed this.

frontend/src/components/application/ApplicationFormsTable.tsx
```diff
@@ -70,10 +66,8 @@ export const ApplicationFormsTable = ({
     <>
       <h3>{t("requiredForms")}</h3>
       <ApplicationTable
-        forms={forms}
-        applicationForms={requiredForms}
```
ErinPattisonNava, line 74: Note that this is just the required forms.

frontend/src/components/application/ApplicationFormsTable.tsx
```diff
@@ -90,11 +84,9 @@ export const ApplicationFormsTable = ({
             })}
           </p>
           <ApplicationTable
-            forms={forms}
-            applicationForms={conditionalRequiredForms}
```
ErinPattisonNava, line 94: And this is supposed to just be the conditional forms

ErinPattisonNava (commented):
All the forms show up both under the required list and the conditional list, regardless of whether they are conditional or required.

frontend/src/components/application/ApplicationFormsTable.tsx
hayeshorus: ok, then forms only required will show, else nothing will show?

ErinPattisonNava (approved):
Approving again after merge issues.

Issue #7560, opened 2025-12-18
Create an endpoint for setting the is_simpler_grants_enabled flag on a competition


chouinar:
### Summary

To avoid manual DB work, we want to create endpoints for working with data in our system. This endpoint just needs to handle setting a boolean on our competition table.

We want to create an endpoint `PUT /alpha/competitions/{competition_id}/flags`

### Request
```json
{
  "is_simpler_grants_enabled": true
}
```

### Response
Make the response the same as the GET competition endpoint, same schema, can just return the competition DB object.

### Auth
This endpoint will only accept API User Key Auth (no JWT auth needed at this time).

The user must have the `manage_competition` privilege, a new privilege we'll want to create AND add to the internal user role that we have defined in https://github.com/HHS/simpler-grants-gov/blob/main/api/src/constants/static_role_values.py

We should use our usual logic for checking that a user has access (see the update form instruction endpoint which similarly checks a different internal role privilege), this will 403 if they don't.

### Logic
Fetch the competition, 404 if not found.

Set the value of the `is_simpler_grants_enabled` flag on the competition as passed in by the user (fine if the user passed in the value it's already at). That's it, should be a very simple change.

Note -- this work supports Simpler Apply

### Acceptance criteria

- [ ] Endpoint Created
- [ ] Test written that verify the basic behavior, auth is working, and basic validation (competition not found, invalid request parameter, missing the one field it needs)

PR #7736, opened 2025-12-31
Add endpoint to update is_simpler_grants_enabled flag for competitions

kkrug:
## Summary

Fixes for #7560  

## Changes proposed

Added a PUT endpoint to update the is_simpler_grants_enabled boolean flag for competitions

## Context for reviewers

In order to prevent this flag from needing to be set manually in the DB, a PUT endpoint was created so users with an internal role can set this flag via the API.

## Validation steps

1. Get a competition_id from the database, noting whether the is_simpler_grants_enabled flag is currently set to true or false
2. Execute the endpoint via Swagger using the competition_id and switching the flag to true/false
3. Confirm in the database that the flag for that competition_id has flipped


api/src/services/competition_alpha/update_competition_flag.py
```diff
@@ -0,0 +1,45 @@
...
+        .options(
+            # Chained selectinload fetches nested children: Competition -> CompetitionForm -> Form
+            # This is required in order to return the competition DB object
+            selectinload(Competition.competition_forms)
+            .selectinload(CompetitionForm.form)
+            .selectinload(Form.form_instruction),
+            # Fetch other required top-level relationships
+            selectinload(Competition.opportunity_assistance_listing),
+            selectinload(Competition.link_competition_open_to_applicant),
+            selectinload(Competition.competition_instructions),
```
kkrug: This is necessary due to the criteria of returning the competition db object in the response, which has a bunch of nested data. I am wondering if altering the response to only return the competition_id and the new state of the flag instead would make more sense and allow us to avoid loading all of this extra data.

api/tests/src/api/competition_alpha/test_competition_update_flag.py
```diff
@@ -0,0 +1,97 @@
...
+    db_session.commit()
+
+    payload = {"is_simpler_grants_enabled": True}
+    url = f"/alpha/competitions/{competition.competition_id}/flag"
+
+    resp = client.put(url, headers={"X-API-Key": internal_admin_user_api_key}, json=payload)
+
+    assert resp.status_code == 200
+    data = resp.get_json()["data"]
+    assert data["is_simpler_grants_enabled"] is True
```
jakobpederson, line 35: What about a test that proves you can set "is_simpler_grants_enabled" to False?

api/src/api/competition_alpha/competition_route.py
```diff
@@ -26,3 +30,23 @@ def competition_get(db_session: db.Session, competition_id: uuid.UUID) -> respon
...
+@flask_db.with_db_session()
+def update_competition_flag_route(
+    db_session: db.Session, competition_id: uuid.UUID, json_data: dict[str, bool]
+) -> response.ApiResponse:
+    is_enabled = json_data.get("is_simpler_grants_enabled", False)
+
+    with db_session.begin():
+        user = api_user_key_auth.get_user()
+        user = db_session.merge(user, load=False)
+        verify_access(user, {Privilege.MANAGE_COMPETITION}, None)
```
chouinar: Could you move this into the `update_competition_flag` function? We try to put all of the AuthZ into the service layer so that components are reusable in a way that enforces access-rights.

api/src/api/competition_alpha/competition_route.py
```diff
@@ -26,3 +30,23 @@ def competition_get(db_session: db.Session, competition_id: uuid.UUID) -> respon
...
+@competition_blueprint.auth_required(api_user_key_auth)
+@flask_db.with_db_session()
+def update_competition_flag_route(
+    db_session: db.Session, competition_id: uuid.UUID, json_data: dict[str, bool]
+) -> response.ApiResponse:
+    is_enabled = json_data.get("is_simpler_grants_enabled", False)
+
+    with db_session.begin():
+        user = api_user_key_auth.get_user()
+        user = db_session.merge(user, load=False)
```
chouinar: I don't think these would necessarily be different in function, but merge has a lot of very specific technical complexity that I like to avoid it if possible.

```suggestion
        user = db_session.add(user)
```

api/src/api/competition_alpha/competition_route.py
```diff
@@ -26,3 +30,23 @@ def competition_get(db_session: db.Session, competition_id: uuid.UUID) -> respon
...
+
+@competition_blueprint.put("/competitions/<uuid:competition_id>/flag")
+@competition_blueprint.input(competition_schema.CompetitionFlagUpdateSchema, location="json")
+@competition_blueprint.output(competition_schema.CompetitionResponseAlphaSchema())
+@competition_blueprint.auth_required(api_user_key_auth)
+@flask_db.with_db_session()
+def update_competition_flag_route(
+    db_session: db.Session, competition_id: uuid.UUID, json_data: dict[str, bool]
+) -> response.ApiResponse:
+    is_enabled = json_data.get("is_simpler_grants_enabled", False)
```
chouinar, line 44: Generally, we'll put the IDs into the logs / log the endpoint first. Both to make sure the relevant data is tracked on the rest of the logs (useful to see all logs for a given competition) as well as for debugging to see that it reached our code in the event of a failure.

```suggestion
) -> response.ApiResponse:
    add_extra_data_to_current_request_logs({"competition_id": competition_id})
    logger.info("PUT /alpha/competitions/:competition_id/flag")

    is_enabled = json_data.get("is_simpler_grants_enabled", False)
```

api/src/services/competition_alpha/update_competition_flag.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def update_competition_flag(
+    db_session: db.Session, competition_id: str, is_simpler_grants_enabled: bool
+) -> Competition:
+    comp_uuid = uuid.UUID(competition_id)
+
+    stmt = (
+        select(Competition)
+        .where(Competition.competition_id == comp_uuid)
```
chouinar: The ID comes in as a UUID, you converted it to a string and then back - just leave it a UUID. Also, to keep things simple, use the `get_competition` function from the other competition route. It'd already have the selectinloads it needs + handles the 404, all things we also want the exact same.

```suggestion
    db_session: db.Session, competition_id: uuid.UUID, is_simpler_grants_enabled: bool
) -> Competition:

    competition = get_competition(db_session, competition_id)
```

api/src/services/competition_alpha/update_competition_flag.py
```diff
@@ -0,0 +1,45 @@
...
+
+    competition = db_session.execute(stmt).scalar_one_or_none()
+
+    if not competition:
+        raise_flask_error(404, f"Competition with ID {competition_id} not found.")
+
+    # Apply the change
+    competition.is_simpler_grants_enabled = is_simpler_grants_enabled
+
+    logger.info(f"Updated is_simpler_grants_enabled for competition {competition_id}")
```
chouinar: We generally try to avoid putting any variables into the log messages themselves. That way we can easily query for specific log messages in New Relic. We usually put any variables into the extra section as that can easily be queried against (eg. fetching all logs where competition ID is some value). Along with the suggestion I had in the route to add the competition ID to the extra for the request lifecycle, we could adjust this like so:

```suggestion
    logger.info("Updated is_simpler_grants_enabled for competition", extra={"is_simpler_grants_enabled": is_simpler_grants_enabled})
```

api/tests/src/api/competition_alpha/test_competition_update_flag.py
```diff
@@ -0,0 +1,97 @@
...
+        user = factories.UserFactory.create()
+        factories.UserApiKeyFactory.create(user=user, key_id=api_key_value)
+
+    admin_role = factories.RoleFactory.create(
+        role_name=f"Competition Manager {uuid.uuid4()}", privileges=[Privilege.MANAGE_COMPETITION]
+    )
+    factories.LinkRoleRoleTypeFactory.create(role=admin_role, role_type=RoleType.INTERNAL)
+    factories.InternalUserRoleFactory.create(user=user, role=admin_role)
+
+    db_session.commit()
```
chouinar: The create function of our factories calls `commit` when they run, can omit.

```suggestion
```

api/tests/src/api/competition_alpha/test_competition_update_flag.py
```diff
@@ -0,0 +1,97 @@
...
+from src.db.models.user_models import UserApiKey
+
+
+def setup_admin_privileges(db_session, api_key_value):
+    """
+    Ensures the user associated with the API key has the MANAGE_COMPETITION privilege.
+    """
+    existing_key = db_session.execute(
+        select(UserApiKey).where(UserApiKey.key_id == api_key_value)
+    ).scalar_one_or_none()
```
chouinar: The `api_key_value` being passed into this is from the `internal_admin_user_api_key` fixture which is an API key's ID. This fixture calls the `internal_admin_user` fixture which is the user. If you want the user, just pull in the `internal_admin_user` fixture as well.

But, I'm not sure you need to do any of what is in this function, those fixtures setup a user with the internal role which has this privilege already added. Either just use that user unmodified, or make a completely new user. You're basically adding a duplicate role to that user that they already have.

api/src/services/competition_alpha/update_competition_flag.py
kkrug: This was replaced with a call to the existing get function.

api/tests/src/api/competition_alpha/test_competition_update_flag.py
kkrug: Added the new test.

api/src/api/competition_alpha/competition_route.py
kkrug: Access verification moved to the service layer.

api/src/api/competition_alpha/competition_route.py
kkrug: Fixed

api/src/api/competition_alpha/competition_route.py
kkrug: Fixed logging

api/src/services/competition_alpha/update_competition_flag.py
kkrug: Fixed

api/src/services/competition_alpha/update_competition_flag.py
kkrug: Fixed

api/tests/src/api/competition_alpha/test_competition_update_flag.py
kkrug: Fixed

api/tests/src/api/competition_alpha/test_competition_update_flag.py
kkrug: I made some changes to address this redundancy. Some of my tests were returning 403s after the changes as the privilege was still missing, so I added a helper function to check for and add the privilege if it is missing.

kkrug:
@chouinar are you ok with all the changes here?

chouinar (approved):

Issue #7645, opened 2025-12-23
Cleanup incremental opportunity search load logic


chouinar:
### Summary

A while back we modified our logic for[ loading opportunities to search](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/search/backend/load_opportunities_to_index.py) to support a way of just loading opportunities that changed. That logic is mostly finished, but we never went forward with the work, and if we were to proceed, we may want to support it a different way.

We want to cleanup the code we added, but not lose it in the event we want to revive it later. For this work, make sure to leave a comment in the load-opportunity code to the PR itself saying something along the lines of "this is where you can find incremental search logic". We can use the PR itself as a way of tracking this.

What we want to remove:
* In the load-opportunity code, remove anything that isn't used by the full-refresh path.
* In the load-opportunity code, remove any references or logic specifically for attachments, while the full-refresh path has some of this, it's not enabled ever because it would make the hourly load job take a day to run.
* Remove [text_extractor.py](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/util/text_extractor.py) which is only used by the incremental logic.
  * This text extraction logic has multiple dependencies that only it uses, we should be able to remove the following dependencies from our pyproject.toml file: `pypdf python-docx beautifulsoup4 striprtf python-pptx openpyxl`
* In the load-opportunity code, remove the `enable_opportunity_attachment_pipeline` env var and any reference, also remove the incremental env vars. 
* Remove [the Dockerfile we made specifically for Opensearch](https://github.com/HHS/simpler-grants-gov/blob/main/api/opensearch/Dockerfile), this was only necessary for an attachment plugin.
  * Modify our [docker-compose.yml](https://github.com/HHS/simpler-grants-gov/blob/main/api/docker-compose.yml) for the opensearch-node service. Remove the build section that references the removed Dockerfile, and instead add `image:  opensearchproject/opensearch:2` 

### Acceptance criteria

- [ ] References to incremental search are removed
- [ ] Code contains a reference to the created PR saying where to find the incremental code if needed

PR #7748, opened 2026-01-02
[Issue #7645] Cleanup Incremental Load

babebe:
## Summary

Fixes / Work for #7645  

## Changes proposed

- Removed all incremental / changed-only opportunity loading logic from the load-opportunity code path, leaving only logic required for the full-refresh flow.
- Added a comment in the load-opportunity code referencing this PR as the location where the incremental search logic previously lived, so it can be revived later if needed.
- Removed all attachment-specific logic from the load-opportunity code
- Deleted text_extractor.py, which was only used by the incremental opportunity logic.
- Removed unused dependencies from pyproject.toml that were only required for text extraction
- Removed the following environment variables and all related references:
  -   ENABLE_OPPORTUNITY_ATTACHMENT_PIPELINE   
  -   All incremental opportunity loading env vars
- Updated docker-compose.yml to use image: opensearchproject/opensearch:2

## Context for reviewers

A previous effort introduced incremental opportunity loading to only index changed records, but that work was never completed or rolled out. Rather than keeping unused and partially integrated logic in the codebase, this PR cleans it up while preserving a breadcrumb for future work.

To avoid losing that investment, a comment has been left in the load-opportunity code pointing back to this PR as the reference for where incremental search logic previously existed.

This cleanup also removes attachment-related indexing and its supporting infrastructure (text extraction, dependencies, custom Opensearch image), which are not viable for the current hourly full-refresh job due to performance constraints.

## Validation steps

- Confirmed no references remain to incremental or attachment-specific environment variables.
- Ran dependency installation to ensure removed libraries are no longer required.
- Confirmed docker-compose up successfully starts Opensearch using the official opensearchproject/opensearch:2 image.


api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -82,193 +58,19 @@ def __init__(
...
     def run_task(self) -> None:
-        logger.info("Creating multi-attachment pipeline")
-        self._create_multi_attachment_pipeline()
+        # NOTE:
+        # Incremental (changed-only) opportunity loading logic previously lived here.
+        # It was removed as part of PR https://github.com/HHS/simpler-grants-gov/pull/7748 during code cleanup.
+        # See that PR if incremental search support needs to be revived in the future.
         if self.is_full_refresh:
             logger.info("Running full refresh")
             self.full_refresh()
```
chouinar: Could we remove the full refresh bool here and in the passed in parameter to the class as well?
```suggestion
        logger.info("Running full refresh")
        self.full_refresh()
```

Issue #7336, opened 2025-12-04
403s not handled on form page


doug-s-nava:
### Summary

Errors in the call to `getApplicationDetails` on the form page are not being handled

https://github.com/HHS/simpler-grants-gov/blob/0c0747e6150ce19c0238e13ecf443f0700722b43/frontend/src/app/%5Blocale%5D/(base)/workspace/applications/application/%5BapplicationId%5D/form/%5BappFormId%5D/page.tsx#L53

Wrapping this call in a try block and handling errors should fix the issue.
See Doug S for more information if need.

### Reproducibility

I tried multiple times and saw the bug every time

### How to reproduce the bug

- login as user A
- create an application and access a form
- log out
- login as user B
- see unhandled error message

<img width="1039" height="641" alt="Image" src="https://github.com/user-attachments/assets/1dfb5a66-9265-4f4d-972e-7c47b5a079d6" />

### Browser

None

### Acceptance Criteria
- [ ] Following reproduction steps and the app does not break
- [ ] Covers all error message (not just 403)
- [ ] Show user friendly error messages
- [ ] Review with Design before submitting PR
- [ ] Unit tests added

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

chris-kuryak:
@chay-REIsys Should this be "In Review"?

chris-kuryak:
@doug-s-nava @chay-REIsys 

How can I test this? I only have one account in Staging and the reproduction steps seem to indicate I need a second account. I don't think that a user have two valid accounts is a scenario that will occur but very rarely, so I assume there is another way to test this issue in a more common user experience.

doug-s-nava:
I guess the actual test case is a user sending a link to someone that doesn't have access? You could make sure someone else sees the right error message for your application, or you could try accessing someone else's?

chris-kuryak:
Thanks @doug-s-nava !

chris-kuryak:
Error message now shown is user-friendly:

<img width="1504" height="288" alt="Image" src="https://github.com/user-attachments/assets/f59db33d-c289-4f04-bde9-e5f1efe5b118" />

PR #7749, opened 2026-01-02
[Issue #7336] Handle unauthorized exceptions in getApplicationDetails…

chay-REIsys:
## Summary

Fixes / Work for #7336 

## Changes proposed

get Application Details is included in try catch to handle exceptions much better
added unit tests for the method

## Context for reviewers

- login with user A
- create application
- access any form from forms table
- logout
- login with other user it will redirect to the previous link when we logged out but now we show error page as below

image

### DIFF
--- a/frontend/src/services/fetch/fetchers/applicationFetcher.ts
+++ b/frontend/src/services/fetch/fetchers/applicationFetcher.ts
@@ -66,28 +66,18 @@ export const getApplicationDetails = async (
   applicationId: string,
   token: string,
 ): Promise<ApplicationDetailApiResponse> => {
-  try {
-    const ssgToken = {
-      "X-SGG-Token": token,
-    };
-    const response = await fetchApplicationWithMethod("GET")({
-      subPath: applicationId,
-      additionalHeaders: ssgToken,
-      nextOptions: {
-        tags: [`application-${applicationId}`, "application-details"],
-      },
-    });
+  const ssgToken = {
+    "X-SGG-Token": token,
+  };
+  const response = await fetchApplicationWithMethod("GET")({
+    subPath: applicationId,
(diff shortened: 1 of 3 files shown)


frontend/tests/services/fetch/fetchers/applicationFetcher.test.ts
```diff
@@ -0,0 +1,185 @@
+import {
```
doug-s-nava, line 1: ❤️

frontend/src/services/fetch/fetchers/applicationFetcher.ts
```diff
@@ -66,18 +66,28 @@ export const getApplicationDetails = async (
...
-    "X-SGG-Token": token,
-  };
-  const response = await fetchApplicationWithMethod("GET")({
-    subPath: applicationId,
-    additionalHeaders: ssgToken,
-    nextOptions: {
-      tags: [`application-${applicationId}`, "application-details"],
-    },
-  });
+  try {
```
doug-s-nava: The most correct fix for this issue, I'm realizing, is to wrap this page in an AuthorizationGate component. This would mean a pretty sizeable refactor of the existing page, and the application page needs that work done as well. I can make a separate ticket for you to pick up as a follow up to implement that.

For now, can you follow the pattern that exists on the application page, and add the try catch logic within the page component, where the call to the fetch function is being made, and return the `<TopLevelError>` if an error occurs? https://github.com/HHS/simpler-grants-gov/blob/c7d4a7aa99677b304b9fe0676dc50dbb8e579c92/frontend/src/app/%5Blocale%5D/(base)/workspace/applications/application/%5BapplicationId%5D/page.tsx#L84

Issue #7646, opened 2025-12-23
Remove agency list endpoint


chouinar:
### Summary

We have two different endpoints for fetching agency info.

The `POST /v1/agencies` endpoint is not meant to be used anymore and doesn't use a supported authentication method anymore. We built a better agency search endpoint that we use instead.

We want to remove this older legacy endpoint as it shouldn't be used anymore.

NOTE: There is a very small amount of traffic to this endpoint, we have contact the person using it several times and asked them to switch to the better endpoint, but haven't gotten a response after several months, we are okay with removing this regardless.

### Acceptance criteria

- [ ] `POST /v1/agencies` and any references to it removed including the service layer functions + unit tests

PR #7752, opened 2026-01-05
[Issue #7646] Remove agency list endpoint

babebe:
## Summary

Fixes / Work for #7646  

## Changes proposed

- Removed legacy POST /v1/agencies endpoint
- Deleted associated service-layer functions used exclusively by this endpoint
- Removed related unit tests


babebe:
@ErinPattisonNava @doug-s-nava @chouinar There is still a front end component that makes reference to this endpoint: https://github.com/HHS/simpler-grants-gov/blob/c7d4a7aa99677b304b9fe0676dc50dbb8e579c92/frontend/src/services/fetch/fetchers/clientAgenciesFetcher.ts#L7 . Not sure why we are not using the `/agencies/search` endpoint here

doug-s-nava:
@babebe what you've got here is a call to the NextJS API, which proxies a call out to /agencies/search on the python API. We should be good for this backend change

babebe:
> what you've got here is a call to the NextJS API, which proxies a call out to /agencies/search on the python API. We should be good for this backend change

Ah, got it — I didn’t realize this was proxying to /agencies/search. Thank you @doug-s-nava

api/tests/src/api/opportunities_v1/test_opportunity_route_search.py
```diff
@@ -1748,6 +1750,7 @@ def test_search_top_level_agency_200(
         api_auth_token,
     ):
         # setup-data
+        cascade_delete_from_db_table(db_session, Agency)
```
chouinar: What issue were you hitting if you didn't add this? Want to always check if it's absolutely required as we should avoid these cascade deletes unless we have no other options.

I don't think we need this actually, the `AgencyFactory` calls don't matter - we don't use the DB in search, so creating anything in the DB wouldn't affect the results returned. I think you can remove any factory stuff from these two functions and be perfectly fine.

PR #7757, opened 2026-01-05
Remove email from logs in internal role endpoint

kkrug:
## Summary

Fixes / Work for #7557  

## Changes proposed

Removed user email from logs in the internal role endpoint.

## Context for reviewers

The internal role endpoint was recently created with a logging issue that exposes user PII (email). This removes the PII.


api/src/services/internal/update_internal_user_role.py
```diff
@@ -46,4 +46,4 @@ def update_internal_user_role(
     new_assignment = InternalUserRole(user_id=target_user_id, role_id=internal_role_id)
     db_session.add(new_assignment)
 
-    logger.info(f"Successfully assigned role {internal_role_id} to user {user_email}")
+    logger.info(f"Successfully assigned role {internal_role_id} to user.")
```
chouinar: The internal role ID is already added to the extra based on the `add_extra_data_to_request_logs` call. We like to have log messages be static generally as it makes querying for them a bit easier in New Relic.

```suggestion
    logger.info("Successfully assigned role to user.")
```

kkrug: Fixed

kkrug:
@chouinar are we all set to merge this?

api/src/services/internal/update_internal_user_role.py
mdragon: It would be good to have the user ID receiving the role and the user ID granting the role in the logs too in the future for troubleshooting purposes. Any reason not to open a ticket to add those values to the add_extra_data_to_current_request_logs call in the handler you mentioned? Would that have to be called in the handler, so we'd need to change this function to return the target_user_id? @chouinar

Issue #7647, opened 2025-12-23
Binary media file uploads failing for apply attachments


mikehgrantsgov:
### Summary

API Gateway is not allowing successful saving of binary media types. Investigation uncovered that API Gateway has no binary media types set, so API Gateway is likely encoding these uploads and storing them malformed.

Conversation in Slack: https://betagrantsgov.slack.com/archives/C08NCEXRSPJ/p1766503052919959

### Acceptance criteria
- [ ] Change deployed to API Gateway
- [ ] Binary and non-binary file uploads in Staging can be uploaded, then downloaded successfully.

ErinPattisonNava:
@mdragon Moving to platform and feel free to combine this with the ones you are creating.

ychoy:
Duplicate

PR #7758, opened 2026-01-05
Fix API Gateway pdf uploads for alpha/applications

Nava-JoshLong:
## Summary

This PR fixes the issue for the alpha application attachments not being useable downstream

Fixes / Work for #7647 

## Changes proposed

- Defined endpoint for `/alpha/applications/{app_id}/attachments`
- Defined endpoint for `/alpha/applications/{application_id}/attachments/{application_attachment_id}`
- Defined endpoint for `/alpha/forms/{form_id}/form_instructions/{form_instruction_id}`
- Model for PDF uploads
- Allow uploads for specific file types on the gateway

## Context for reviewers

API Gateway encodes all request bodies by default, which was causing the attachments file to "work", but we couldn't open the uploads. This PR introduces and implements an infra side change to allow attachments, and not encode them, so downstream systems can use them as intended

## Validation steps

This PR's changes have been deployed so we can allow the reviewers to test

- Login to the dev site
- Start a new application
- Upload a file
- Go into S3, find your upload, and then download and open it
    - If the file is either blank, or the file fails to open, then ping me. The fix isn't working for your upload

### DIFF
--- a/infra/modules/service/api_gateway.tf
+++ b/infra/modules/service/api_gateway.tf
@@ -138,7 +138,7 @@ locals {
         "method" : "PUT",
         "api_key_required" : true,
         "method_parameters" : {
-          "method.request.path.application_id" = true,
+          "method.request.path.application_id"            = true,
           "method.request.path.application_attachment_id" = true,
         },
         "request_parameters" : {


infra/modules/service/api_gateway.tf
```diff
@@ -60,15 +66,29 @@ locals {
...
+      "grantsws-agency/services/v2"         = [],
+      "grantsws-applicant/services/v2"      = [],
+      "v1/users/login"                      = [{ "method" : "GET" }],
+      "v1/users/token"                      = [],
   }][var.enable_api_gateway ? 1 : 0]
 
   third_level_endpoints = [
     {},
     {
+      "alpha/applications/{application_id}/attachments" = [{
```
chouinar, line 81: We need this for more than just this endpoint, the `PUT /alpha/forms/:form_id/form_instructions/:form_instruction_id` endpoint is also currently broken.

Any way to generically define some of this so we just need a list of paths? Guessing that might be tricky with the request params

infra/modules/service/api_gateway.tf
```diff
@@ -158,13 +182,39 @@ resource "aws_api_gateway_rest_api" "api" {
   endpoint_configuration {
     types = ["REGIONAL"]
   }
+
+  binary_media_types = [
+    "application/pdf",
+    "application/octet-stream",
+    "multipart/form-data",
+  ]
+}
```
chouinar, line 241: What happens if the file is something else like a word doc?

Nava-JoshLong: It's really difficult to get a non-AI generated list of allowed types, but I believe most file types that aren't images, videos, or PDFs will fall under the `multipart/form-data` or `application/octet-stream` buckets. We can test that on the attachments endpoint to double check me though. Worst case, I can allow all, but I don't feel like that should be our first option

infra/modules/service/api_gateway.tf
Nava-JoshLong: I'll add the PUT in the next commit. As for generically defining things, the current setup is as generic as I can make it without making it difficult down the line to update things

Nava-JoshLong: Forms endpoint is now live, let me know if it doesn't work

infra/modules/service/api_gateway.tf
chouinar: Does it allow wildcards? Can you do `image/*` for both `image/png` and `image/jpeg`?

For reference, I went to grants.gov's DB to see what mimetypes they have recorded for application attachments over the past 2 years, there were 77 distinct values. The top 20 or so:
```
application/pdf
application/vnd.openxmlformats-officedocument.wordprocessingml.document
application/vnd.openxmlformats-officedocument.spreadsheetml.sheet
image/jpeg
image/png
application/msword
application/unknown
application/x-zip-compressed
application/vnd.ms-excel
text/plain
application/vnd.openxmlformats-officedocument.presentationml.presentation
application/vnd.ms-excel.sheet.binary.macroEnabled.12
application/vnd.google-earth.kmz
application/vnd.ms-excel.sheet.macroEnabled.12
application/zip
application/vnd.oasis.opendocument.text
application/x-iwork-pages-sffpages
application/vnd.google-earth.kml+xml
text/csv
image/tiff
text/html
video/mp4
```

Nava-JoshLong: Bringing some of the slack thread back here, no we cannot do wildcards. We need to explicitly say what is allowed. This might be changed to allow all, and the filtering on what is and isn't allowed will be done on the ECS side

chouinar: Have we tested whether `multipart/form-data` encompasses all the different file types with the way we upload? If I upload a PDF or .docx file, it doesn't break?

infra/modules/service/api_gateway.tf
chouinar: Sorry, missed another one. Along with our `POST /alpha/applications/:application_id/attachments` endpoint, we have a separate endpoint for updating an existing application attachment that is `PUT /alpha/applications/:application_id/attachments/:application_attachment_id`

infra/modules/service/api_gateway.tf
Nava-JoshLong: I forgot to move the other part of that thread here, I tested with PDF, docx, and image, and all 3 were able to be downloaded from S3 and worked as expected

infra/modules/service/api_gateway.tf
Nava-JoshLong: Got it, I'll update the code and push once I have it set

Nava-JoshLong: @chouinar Code is updated, let me know when you might have some time to test it, and I can deploy these changes

Issue #7721, opened 2025-12-30
Remove snapshot tests


ErinPattisonNava:
### Summary

We want to establish good unittest examples so we all agree where we are going/what is expected when we write unittests.   We'll start by removing snapshot tests, as they are low value and not a pattern we want to continue.

### Acceptance criteria

- [ ] Remove all snapshot tests

PR #7767, opened 2026-01-05
[Issue 7721] Remove snapshot tests

myduong-navapbc:
## Summary

Work for [#7721](https://github.com/HHS/simpler-grants-gov/issues/7721)

Removal of snapshots and shifting of some test files to colocate with the thing they are testing

## Validation steps

- [ ] Tests pass, `npm run test`

### DIFF
--- a/frontend/src/utils/opportunity/zipUtils.test.ts
+++ b/frontend/src/utils/opportunity/zipUtils.test.ts
@@ -2,11 +2,11 @@
  * @jest-environment node
  */
 
+import { HttpReader as FakeHttpReader } from "@zip.js/zip.js";
 import {
   attachmentsToZipEntries,
   deduplicateFilename,
 } from "src/utils/opportunity/zipUtils";
-import { HttpReader as FakeHttpReader } from "@zip.js/zip.js";
 
 describe("deduplicateFilename", () => {
   it("returns original filename if no previous usage of name is found", () => {


frontend/src/components/application/ApplicationFormsTable.test.tsx
```diff
@@ -0,0 +1,165 @@
...
+    render(
+      <ApplicationFormsTable
+        competitionInstructionsDownloadPath="http://path-to-instructions.com"
+        forms={competitionForms}
+        applicationForms={applicationForms}
+        applicationId={applicationId}
+      />,
+    );
+
+    // Required form: ABC shows Unavailable
```
ErinPattisonNava, line 76: Future thought: I'd prefer if these were their own `it`s instead of adding a comment it would be in the test name.  But we can defer that until we get to standards/examples.

frontend/tests/jest.setup.ts
```diff
@@ -0,0 +1,7 @@
+import "@testing-library/jest-dom";
+
+import { toHaveNoViolations } from "jest-axe";
+
+expect.extend(toHaveNoViolations);
+
+process.env.apiUrl = "http://localhost";
```
ErinPattisonNava, line 7: Nit: I did find the note in the deleted file that indicates this is where we mock environment variables useful, could we bring it back?

frontend/tests/utils/next-intl-messages.ts
```diff
@@ -0,0 +1,20 @@
+import type { AbstractIntlMessages } from "next-intl";
```
ErinPattisonNava, line 1: I feel like maybe these should live with the other intl mocks [here](https://github.com/HHS/simpler-grants-gov/blob/main/frontend/src/utils/testing/intlMocks.ts)?  

Or maybe a better question: is the plan that the `frontend/tests` directory is mostly about e2es going forward and unittests/their utilities are stored in `frontend/src`?  Or a different directory strategy?  We don't need to do everything in this PR, but would like to understand the direction.

frontend/tests/utils/mock-feature-flags.ts
```diff
@@ -0,0 +1,14 @@
+type Flags = Record<string, boolean>;
```
ErinPattisonNava, line 1: Perhaps these should live with the other feature flag testing utils [here](https://github.com/HHS/simpler-grants-gov/blob/main/frontend/src/utils/testing/FeatureFlagTestUtils.ts)?

frontend/tests/utils/intl.ts
```diff
@@ -0,0 +1,16 @@
+export type MessagesTree = Record<string, unknown>;
+
+export function getMessage(tree: MessagesTree, dottedKey: string): string {
```
ErinPattisonNava, line 3: Should this replace the utilities [here](https://github.com/HHS/simpler-grants-gov/blob/main/frontend/src/utils/testing/intlMocks.ts#L1)?  I'm unsure when we want to use each/what's the long term plan.

frontend/src/components/RootLayoutWrapper.tsx
```diff
@@ -54,13 +55,13 @@ export default async function RootLayoutWrapper({
   // to do that in commit 46566b4c0ad but later removed. We can bring it back if it is ever useful. - DWS
   return (
     <html lang={locale} suppressHydrationWarning>
-      <head>
+      <Head>
```
ErinPattisonNava, line 58: This would have been a good thing to call out in the PR description that it needed to be tested specifically.  I think this was the only non-testing file change, is that correct?

I no longer see the meta tag for `google-site-verification` in the script tag on the homepage, I think that's going to cause problems for our integration?

ErinPattisonNava: And just to be clear, this is the comment why I didn't approve and needs to be addressed.

Issue #7724, opened 2025-12-30
remove the developerPageOff feature flag


ErinPattisonNava:
### Summary

The developerPageOff feature flag is disabled in production (terrible double negative there), so we want to remove the flag from the code base to simplify complexity and reduce confusion. 

### Acceptance criteria

- [ ] Remove all mentions of the developerPageOff from the code base, leaving the code in the state that the flag is effectively disabled (i.e. the behaviour is enabled).
- [ ] Remove any unused components, tests, strings, or types that have become outdated.
- [ ] Solemnly promise out loud that we will never name flags with a double negative again.

ErinPattisonNava:
@btabaska Any concerns with removing this feature flag?

btabaska:
> [@btabaska](https://github.com/btabaska) Any concerns with removing this feature flag?

Nope, let's get rid of it.

ErinPattisonNava:
@btabaska Since the PR is merged, can we close this ticket?

PR #7788, opened 2026-01-06
[Issue #7724] remove the developerPageOff feature flag

btabaska:
## Summary

Removed the developer page FF
Updated the snapshots in the header now that the dev page is showing correctly

## Changes proposed

Remove FF
Update Snaps

TODO: After this goes to prod, remove the FF from AWS 

## Context for reviewers

We are deprecating no longer used feature flags. 

## Validation steps

Check that baseurl/dev/feature-flags does not show developerPageOff feature flag
Check that header has link under Community to the Developer page that loads.

### DIFF
--- a/infra/frontend/app-config/env-config/environment_variables.tf
+++ b/infra/frontend/app-config/env-config/environment_variables.tf
@@ -107,10 +107,6 @@ locals {
       manage_method     = "manual"
       secret_store_name = "/${var.app_name}/${var.environment}/feature-search-table-on"
     },
-    FEATURE_DEVELOPER_PAGE_OFF = {
-      manage_method     = "manual"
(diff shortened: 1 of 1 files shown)


andycochran:
Shouldn't hurt if we don't remove the FF from AWS, right? Can we add that as a AC on #7724?

btabaska:
> would be good to remove from the terraform as well
> 
> https://github.com/HHS/simpler-grants-gov/blob/8e3d0d3f40893dcd4fb28aacbb032c3dd7349e97/infra/frontend/app-config/env-config/environment_variables.tf#L110
Screenshot 2026-01-06 at 4 18 14 PM

It is removed in the PR, unless I'm missing something? @doug-s-nava

PR #7818, opened 2026-01-07
[Issue #7747] Add Field to Application model to signify the applicant intends to add an organization

### LINKED ISSUE #7747: Add Field to Application model to signify the applicant intends to add an organization
### Summary

We want to build the transfer application ownership functionality. To do this we need to persist whether users plan to add an organization later (vs. applying as an individual permanently) so we can provide appropriate messaging in the UI.

A nullable boolean field needs to be added to the Application model to track whether a user intends to add an organization later when starting an application as an individual.

The field's meaning based on state:

| intends_to_add_organization | organization_id | Interpretation                                  |
|-----------------------------|-----------------|-------------------------------------------------|
| null                        | null            | Unknown intent / Not specified                  |
| true                        | null            | Individual who plans to add org later           |
| false                       | null            | Individual applicant (no org planned)           |
| null (or any)               | UUID          | Organization application (field not applicable) |

Note: We will want to prevent organization_id being set and intends_to_add_organization set at the same time. This can probably be done on the api / validation side of things

### Acceptance criteria

- Application model updated to include nullable intends_to_add_organization boolean
- Migrations created

mikehgrantsgov:
## Summary

Fixes #7747

## Changes proposed

Added a nullable boolean field `intends_to_add_organization` to the Application model to track whether users plan to add an organization later when starting an application as an individual.
Update schemas
Adds validation to prevent both `organization_id` and `intends_to_add_organization` from being set simultaneously
Add migration

## Context for reviewers

We want to build the transfer application ownership functionality. To do this we need to persist whether users plan to add an organization later (vs. applying as an individual permanently) so we can provide appropriate messaging in the UI.

## Validation steps

See updated unit tests.

### DIFF
--- a/api/src/api/application_alpha/application_route.py
+++ b/api/src/api/application_alpha/application_route.py
@@ -72,14 +72,7 @@ logger = logging.getLogger(__name__)
 @flask_db.with_db_session()
 def application_start(db_session: db.Session, json_data: dict) -> response.ApiResponse:
     """Create a new application for a competition"""
-    competition_id = json_data["competition_id"]
-    # application_name is optional, so we use get to avoid a KeyError
-    application_name = json_data.get("application_name", None)
-    # organization_id is optional, so we use get to avoid a KeyError
-    organization_id = json_data.get("organization_id", None)
-    # intends_to_add_organization is optional, so we use get to avoid a KeyError
-    intends_to_add_organization = json_data.get("intends_to_add_organization", None)
-    add_extra_data_to_current_request_logs({"competition_id": competition_id})
+    add_extra_data_to_current_request_logs({"competition_id": json_data.get("competition_id")})
     logger.info("POST /alpha/applications/start")
 
     # Get user from token session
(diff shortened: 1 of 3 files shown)


api/src/api/application_alpha/application_route.py
```diff
@@ -77,6 +77,8 @@ def application_start(db_session: db.Session, json_data: dict) -> response.ApiRe
     application_name = json_data.get("application_name", None)
     # organization_id is optional, so we use get to avoid a KeyError
     organization_id = json_data.get("organization_id", None)
+    # intends_to_add_organization is optional, so we use get to avoid a KeyError
+    intends_to_add_organization = json_data.get("intends_to_add_organization", None)
```
chouinar: With this many values being pulled out of the json_data, should we instead just pass the json_data into the create_application function?

api/src/services/applications/create_application.py
```diff
@@ -92,10 +92,25 @@ def create_application(
     user: User,
     application_name: str | None = None,
     organization_id: UUID | None = None,
+    intends_to_add_organization: bool | None = None,
 ) -> Application:
     """
     Create a new application for a competition.
     """
+    # Validate that organization_id and intends_to_add_organization are not both set
+    if organization_id is not None and intends_to_add_organization is not None:
```
chouinar: Would it be an issue if they had an org and said No to the question? Should we instead do:

```suggestion
    if organization_id is not None and intends_to_add_organization is True:
```

api/src/api/application_alpha/application_route.py
mikehgrantsgov: Makes sense, this is a lot cleaner now

chouinar (approved):

PR #7558, opened 2025-12-18
Feature files - Work-flow

aoysimmons-GH:
Feature files - Work-flow

## Summary

Fixes / Work for #6146

PR #7836, opened 2026-01-08
Add negative path scenarios for application workflow

aoysimmons-GH:
This feature file contains various negative path scenarios for the application workflow, including cases for application shell loading failures, form navigation data retention, required field validation, submission errors, data persistence on refresh, save failures, invalid file uploads, and workflow status retrieval issues.

## Summary

Fixes / Work for #7848  

## Changes proposed

 As per the peer-review from this PR: https://github.com/HHS/simpler-grants-gov/pull/7558/files. I updated the path to be under ->_frontend/tests/e2e/apply/features_ (same place of the happy path which Bhavna created) and I updated the scenarios as per the reviewed and comments

### DIFF
--- a/negative_path_apply
+++ b/negative_path_apply
@@ -23,7 +23,6 @@ Scenario: Application shell fails to load when starting a new application
   ###########################################################################
   # 2. Successful Submission – Backend Submission Error
   ###########################################################################
-  @negative @submission @workflow
   Scenario: Submission fails due to backend error
     Given the user has completed all required fields
     When the user clicks Submit
@@ -33,7 +32,6 @@ Scenario: Application shell fails to load when starting a new application
   ###########################################################################
   # 3. Data Persist on Refresh – Data Loss After Refresh
   ###########################################################################
-  @negative @persistence @workflow
   Scenario: Data does not persist after page refresh
     When the user attempts to refresh the page
     Then a warning dialog is displayed about unsaved changes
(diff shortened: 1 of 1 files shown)


negative_path_apply
```diff
@@ -0,0 +1,65 @@
+Feature: Application Workflow - Negative Path Scenarios
```
doug-s-nava, line 1: it looks like this file is in the root path of the project - can you move it to the intended folder, which I think is at /frontend/tests/e2e/apply

can we rename the file to `negative-path.feature`

negative_path_apply
```diff
@@ -0,0 +1,65 @@
...
+  Scenario: Submission fails due to backend error
+    Given the user has completed all required fields
+    When the user clicks Submit
+    Then the system displays "Submission failed"
+    And the user stays on the application page
+
+  ###########################################################################
+  # 3. Data Persist on Refresh – Data Loss After Refresh
+  ###########################################################################
+  Scenario: Data does not persist after page refresh
```
doug-s-nava, line 35: the title of the case talks about data loss, but the actual test case seems to be handling the expected use case. This use case is a good positive case - do you want to add a negative one where data that is expected to be saved is not saved? I am not sure how we would ever test this, but could be worth documenting

negative_path_apply
aoysimmons-GH: I updated and create a new PR here: https://github.com/HHS/simpler-grants-gov/pull/7997/files

negative_path_apply
aoysimmons-GH: I removed it

PR #7876, opened 2026-01-09
[Issue #7090] Deprecate unused agency columns - first deploy

### LINKED ISSUE #7090: Remove unused agency fields from Opportunity Summary table
### Summary

In our OpportunitySummary table we want to remove the following columns:
* `agency_phone_number`
* `agency_code`
* `agency_name`

These fields are unused and often incorrect. Data for these instead comes from elsewhere in the opportunity/agency models rather than these fields.

We also need to adjust our transformation code to not transform these fields anymore - that is done here https://github.com/HHS/simpler-grants-gov/blob/main/api/src/data_migration/transformation/transform_util.py#L210

### Acceptance criteria

- [ ] `agency_phone_number` removed from Opportunity Summary table
- [ ] `agency_code`, `agency_name` removed from Opportunity Summary table
- [ ] Database migration created
- [ ] Updated transformation jobs to no longer populate these fields.

hao10282025-sudo:
## Summary

Fixes #7090  

## Changes proposed

In our OpportunitySummary table we want to remove the following columns:

- agency_phone_number
- agency_code
- agency_name

## Context for reviewers

These fields are unused and often incorrect. Data for these instead comes from elsewhere in the opportunity/agency models rather than these fields.

**Note:** There are 2 steps/deployments for removing columns, and this PR is doing the 1st step/deployment. 

## Validation steps

Validate the deprecation of 3 columns for first deploy step, as mentioned in database-management.md

### DIFF
--- a/api/openapi.generated.yml
+++ b/api/openapi.generated.yml
@@ -3499,7 +3499,7 @@ components:
           minItems: 1
           items:
             type: string
-            pattern: ^\d{2}\.\d{2,3}$
+            pattern: ^\d{2}\.[A-Za-z0-9]{2,3}$
             example: '45.149'
     IsCostSharingFilterV1:
       type: object
--- a/frontend/.env.development
+++ b/frontend/.env.development
@@ -21,6 +21,7 @@ ENVIRONMENT=dev
 
 # Hardcode public auth key for local to local calls
 # This is also hardcoded and checked-in on the API
+API_AUTH_TOKEN=LOCAL_AUTH_12345678
 API_GW_AUTH=local-dev-api-key
 
 AUTH_LOGIN_URL=http://localhost:8080/v1/users/login
--- a/infra/modules/service/api_gateway.tf
+++ b/infra/modules/service/api_gateway.tf
@@ -1,6 +1,10 @@
 locals {
+  alpha_applications_model_name = "alphaApplicationsAttachments"
   # Root level paths, and their method types. If the method is an empty list, this has sub-paths, which will be defined
   # in another variable
+  # Note that by default, API gateway will encode all bodies of requests, which can cause issues for attachments/uploads.
+  # To handle this, you need to create an explicit endpoint that is expecting an upload, and have it use a model. See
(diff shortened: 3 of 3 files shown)


api/src/db/models/opportunity_models.py
```diff
@@ -214,7 +214,8 @@ class OpportunitySummary(ApiSchemaTable, TimestampMixin):
     funding_category_description: Mapped[str | None]
     applicant_eligibility_description: Mapped[str | None]
 
-    agency_phone_number: Mapped[str | None]
+    # SGG-7090, deprecate agency_phone_number usage - first deploy
+    agency_phone_number: Mapped[str | None] = mapped_column(Text().evaluates_none(), deferred=True)
```
chouinar, line 218: So, testing this a bit, I'm not sure if the `evaluates_none` does anything - at least in these cases. https://docs.sqlalchemy.org/en/20/core/type_api.html#sqlalchemy.types.TypeEngine.evaluates_none

Reading the docs a bit more, I think evaluates_none is for a specific use case where you have a server_default which would be set if the value isn't passed in, and you want to force it to null.

For example, assume a table with two columns, ID + description, where description is nullable but has a server default of "example".

If I do:
```sql
insert into my_table(id) VALUES (1);
```
Then description gets the default of `example`. But if I wanted that to be `null`, I'd have to explicitly insert null. 

---
That's all to say, I don't think in these cases we need it at all, but in the event we were deleting a column with a server_default, we would. In the docs, can you:
* Update the example to specify the evalulates none in the correct format (ie `Text().` instead of `Text.`)
* Add the following info about evaluates_none:

```
`evaulates_none` is required if the column has a server_default as it will make it so if a field isn't set, rather than excluding it from the insert into the DB, it sets it to `null` so the server_default isn't used. If the column has no `server_default`, it isn't technically required to use it, but we recommend adding it because it's simpler to have a single pattern to follow rather than multiple.
```

hao10282025-sudo: database-management.md file updated, thanks

Issue #7781, opened 2026-01-06
Support alphanumeric Assistance Listing Numbers in search


joshtonava:
### Summary

NIH has started submitting Assistance Listing Numbers (ALNs) with alphanumeric characters (e.g., 93.KT1, 93.AA1, 93.NR2). Our search API currently rejects these with a 422 validation error because the regex pattern only accepts numeric formats like "45.149". The database already stores these values correctly, but users cannot filter by them in search.

Related thread: https://betagrantsgov.slack.com/archives/C08MTT10SKU/p1767627355343079 

### Tech Notes
- Location: api/src/api/opportunities_v1/opportunity_schemas.py:359
- Current pattern: ^\d{2}\.\d{2,3}$ (numeric only)
- Potential pattern: 
  - ^\d{2}\.[A-Z0-9]{2,3}$ (supports alphanumeric)
  - Or simply make sure the character length is correct 

### Acceptance criteria

- Users can filter opportunities by alphanumeric ALNs (e.g., "93.KT1")
- API accepts both numeric (45.149) and alphanumeric (93.KT1) formats
- Test factory updated to generate alphanumeric test data
- New test case(s) added for alphanumeric ALN search filtering
- All existing tests continue to pass

PR #7877, opened 2026-01-09
[Issue #7781] allow alpha numeric ALN on opportunity search 

babebe:
## Summary

Fixes / Work for #7781 

NIH has started submitting Assistance Listing Numbers (ALNs) with alphanumeric characters (e.g., 93.KT1, 93.AA1, 93.NR2).
Currently, the Opportunity Search API rejects these with a 422 validation error because the regex pattern only accepts numeric formats like 45.149. The database already stores these values correctly, but users cannot filter by them in search.

This PR updates the schema and tests so that both numeric and alphanumeric ALNs are accepted.
## Changes proposed

- New pattern: `r"^\d{2}\.[A-Za-z0-9]{2,3}$"`
Supports both numeric and alphanumeric ALNs. Schema validation allows both uppercase and lowercase letters in alphanumeric ALNs.
- Generate alphanumeric ALNs for test data.
- Added service-layer normalization to uppercase ALNs before building the search request.

## Validation steps

Run existing tests

### DIFF
--- a/analytics/Dockerfile
+++ b/analytics/Dockerfile
@@ -1,4 +1,4 @@
-FROM ghcr.io/hhs/python-base-image:8b8ce02@sha256:70a2337a709022d95effde6f19c9fe31e0f3d8966f401290ba9557148798c477 AS base
+FROM ghcr.io/hhs/python-base-image:8b8ce02@sha256:f5822def7eddfe5163ead820534c78012bcd4f333f553e03831a22f4cc5c9e3b AS base
 
 ARG RUN_UID
 ARG RUN_USER
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -1,5 +1,5 @@
 
-FROM ghcr.io/hhs/python-base-image:8b8ce02@sha256:70a2337a709022d95effde6f19c9fe31e0f3d8966f401290ba9557148798c477 AS base
+FROM ghcr.io/hhs/python-base-image:8b8ce02@sha256:f5822def7eddfe5163ead820534c78012bcd4f333f553e03831a22f4cc5c9e3b AS base
 
 ARG RUN_UID
 ARG RUN_USER


api/tests/src/db/models/factories.py
```diff
@@ -774,7 +775,10 @@ def _setup_next_sequence(cls):
 
     program_title = factory.Faker("company")
     assistance_listing_number = factory.LazyFunction(
-        lambda: f"{fake.random_int(min=1, max=99):02}.{fake.random_int(min=1, max=999):03}"
+        lambda: (
+            f"{fake.random_int(min=1, max=99):02}."
+            f"{''.join(random.choices(string.ascii_uppercase + string.digits, k=random.randint(2, 3)))}"
+        )
```
chouinar: Just for consistency, could we keep this all in Faker and use a provider? Can have multiple formats that it can pick from - and could have `##-???` and `##-###`, you can see how we do that for opportunity numbers and agency codes above in the `CustomProvider` class. Then this whole thing can be `assistance_listing_number = factory.Faker("assistance_listing_number")`

api/src/api/opportunities_v1/opportunity_schemas.py
```diff
@@ -356,7 +356,7 @@ class OpportunitySearchFilterV1Schema(Schema):
     assistance_listing_number = fields.Nested(
         StrSearchSchemaBuilder("AssistanceListingNumberFilterV1Schema")
         .with_one_of(
-            example="45.149", pattern=r"^\d{2}\.\d{2,3}$"
+            example="45.149", pattern=r"^\d{2}\.[A-Z0-9]{2,3}$"
         )  # Always of the format ##.## or ##.###
```
joshtonava: **nitpick:** Can we update this comment

joshtonava:
Discussion: We may want to update https://github.com/HHS/simpler-grants-gov/blob/90f9d44f7f5766af5b9637327d7a0330607e65cb/documentation/wiki/product/api/search-opportunities.md?plain=1#L80-L82

joshtonava:
Discussion:

The legacy soap XSD is using the old numeric only pattern.

https://github.com/HHS/simpler-grants-gov/blob/49f4a48cdc7b9e9a5bacb4a308eebb27c2bd60ea/api/mock-soap-services/applicants/schemas/GrantsCommonTypes-V1.0.xsd#L300

Do we care about this cc: @chouinar

chouinar:
> Discussion:
> 
> The legacy soap XSD is using the old numeric only pattern.
> 
> https://github.com/HHS/simpler-grants-gov/blob/49f4a48cdc7b9e9a5bacb4a308eebb27c2bd60ea/api/mock-soap-services/applicants/schemas/GrantsCommonTypes-V1.0.xsd#L300
> 
> Do we care about this cc: @chouinar

This is a change that grants.gov will have to make as well since they also use assistance listing numbers. Note that particular file isn't used anywhere I think? I don't see a change yet in grants.gov's schema. It might be something worth asking about to someone like Lucas or Julius if we know their timelines for making the changes as it would impact our apply logic.

api/tests/src/db/models/factories.py
babebe: Nice! Updated

api/src/api/opportunities_v1/opportunity_schemas.py
babebe: updated

babebe:
> Discussion: We may want to update
> 
> https://github.com/HHS/simpler-grants-gov/blob/90f9d44f7f5766af5b9637327d7a0330607e65cb/documentation/wiki/product/api/search-opportunities.md?plain=1#L80-L82

@joshtonava looks like I need to make CR through gitbook to make updates there ?

api/src/api/opportunities_v1/opportunity_schemas.py
```diff
@@ -356,8 +356,8 @@ class OpportunitySearchFilterV1Schema(Schema):
     assistance_listing_number = fields.Nested(
         StrSearchSchemaBuilder("AssistanceListingNumberFilterV1Schema")
         .with_one_of(
-            example="45.149", pattern=r"^\d{2}\.\d{2,3}$"
-        )  # Always of the format ##.## or ##.###
+            example="45.149", pattern=r"^\d{2}\.[A-Za-z0-9]{2,3}$"
+        )  # Always of the format ##.## or ##.???
```
chouinar: I would exclude the format comment or rewrite it more directly, since `##.##` and `##.??` and `##.?#?` are all allowed, and specifying every variant wouldn't make sense anymore.

api/src/services/opportunities_v1/search_opportunities.py
```diff
@@ -244,11 +244,24 @@ def _search_opportunities(
...
+
+
 def search_opportunities(
     search_client: search.SearchClient, raw_search_params: dict
 ) -> tuple[Sequence[dict], dict, PaginationInfo]:
 
     search_params = SearchOpportunityParams.model_validate(raw_search_params)
+
+    # Normalize ALN casing before building the search query
+    _normalize_aln(search_params.filters)
```
chouinar: Can we put this in the `_add_search_filters` function instead? Makes sense to adjust the filters as we're adding them. Can do it after we check whether the filters exist, but before we start to iterate over them.

api/src/api/opportunities_v1/opportunity_schemas.py
babebe: Agreed, I’ve removed the format comment. The regex itself, along with the example, should be sufficient

api/src/services/opportunities_v1/search_opportunities.py
babebe: Updated

Issue #7880, opened 2026-01-09
Null byte in tcertificate.is_selfsigned causing an error in load transform


jakobpederson:
### Summary

Getting this error in lower environments:
```
DataError('(psycopg.errors.CharacterNotInRepertoire) invalid byte sequence for encoding "UTF8": 0x00\nCONTEXT: converting column "is_selfsigned" for foreign table scan of "tcertificates", row 1')
```
https://onenr.io/0oQDxO53NQy

It looks like there's invalid values in `is_selfsigned` that's causing things to break when trying to convert them into UTF-8. 

Note: `columns_to_exclude` fixes this issue on an `INSERT` but does not work on an `UPDATE.` 

Once we reproduced this and tested in the dev environment we attempted a fix to manually create the tables adding in the columns option `strip_zeros` in order to remove the null bytes from the column values and prevent the error:
```
DROP FOREIGN TABLE IF EXISTS legacy.tcertificates;
CREATE FOREIGN TABLE legacy.tcertificates (
    currentcertid TEXT OPTIONS (key 'true') NOT NULL,
    previouscertid TEXT,
    orgduns TEXT,
    orgname TEXT,
    expirationdate DATE,
    certemail TEXT NOT NULL,
    agencyid TEXT,
    requestorlname TEXT,
    requestorfname TEXT,
    requestoremail TEXT,
    requestorphone TEXT,
    created_date TIMESTAMP WITH TIME ZONE NOT NULL,
    creator_id TEXT NOT NULL,
    last_upd_date TIMESTAMP WITH TIME ZONE,
    last_upd_id TEXT,
    is_selfsigned TEXT OPTIONS (strip_zeros 'true'),
    serial_num TEXT,
    system_name TEXT
) SERVER grants OPTIONS (schema 'EGRANTSADMIN', table 'TCERTIFICATES', readonly 'true', prefetch '1000');
```

That resulted in a new permissions related error when the `UPDATE` happened: 
```
ProgrammingError('(psycopg.errors.InsufficientPrivilege) permission denied for foreign table tcertificates')
```
https://onenr.io/0dQemAOWZQe

After bringing in Prash we added the permissions to the command and running it BEFORE we drop/created the table:
```
ALTER DEFAULT PRIVILEGES IN SCHEMA legacy GRANT SELECT ON TABLES TO app;
```

This does not occur in the Prod environment because the Prod environment columns do not have null bytes (tested by directly querying the Oracle DB).

### Reproducibility

None

[...]

PR #7881, opened 2026-01-09
[Issue #7880] Manually create the legacy.tcertificates table to add strip_zeros option

jakobpederson:
## Summary

Fixes / Work for #7880  

## Changes proposed

Manually created the `legacy.tcertificates` table in order to add the `strip_zeros` option to address an issue that occurs in the lower envs. I'm adding comments to help explain the situation. Currently the `load-transform` breaks whenever there's an `UPDATE` on the `staging.tcertificates` table because of the invalid value and the `exclude_columns` only addresses this issue on an `INSERT`.

## Context for reviewers

The load for moving the tcertificates from legacy to staging breaks because the null bytes cannot be converted to UTF-8 and so are incompatible.

### DIFF
--- a/api/src/db/models/foreign/certificates.py
+++ b/api/src/db/models/foreign/certificates.py
@@ -8,11 +8,8 @@ option to the is_selfsigned column due to null bytes being present for is_selfsi
 exclude columns option does not work on UPDATES so it causes errors every time an UPDATE is
 attempted when there's a null byte in is_selfsigned. Since null bytes don't seem to be an
 issue in production for that column we can let that be automatically generated.
-Note: the first line with the ALTER is to ensure that the user running the task will have correct access
-to the table.
 
 The command:
-ALTER DEFAULT PRIVILEGES IN SCHEMA legacy GRANT SELECT ON TABLES TO app;
 DROP FOREIGN TABLE IF EXISTS legacy.tcertificates;
 CREATE FOREIGN TABLE legacy.tcertificates (
     currentcertid TEXT OPTIONS (key 'true') NOT NULL,
@@ -22,7 +19,8 @@ CREATE FOREIGN TABLE legacy.tcertificates (
     expirationdate DATE,
     certemail TEXT NOT NULL,
     agencyid TEXT,
-    requestorlname TEXT,                                                                                                                                                                                                  requestorfname TEXT,
+    requestorlname TEXT,
+    requestorfname TEXT,
     requestoremail TEXT,
     requestorphone TEXT,
     created_date TIMESTAMP WITH TIME ZONE NOT NULL,


chouinar:
A few questions:
* Have you tested this non-locally? We have no Oracle database locally - I can't approve this until we've manually vetted that this will work with the Oracle DB.
* Do you have a plan for rolling this change out? This will require deleting and recreating the foreign tables which needs to be carefully timed as you need to avoid the hourly job when doing that.

jakobpederson:
> A few questions:
> 
> * Have you tested this non-locally? We have no Oracle database locally - I can't approve this until we've manually vetted that this will work with the Oracle DB.
> * Do you have a plan for rolling this change out? This will require deleting and recreating the foreign tables which needs to be carefully timed as you need to avoid the hourly job when doing that.

I haven't tested it non-locally and I don't have a plan for rolling it out yet. I'll get those together.

jakobpederson:
Plan for rolling this fix out:

1. Wait to a bout +10 min to +15 min after the top of the hour to avoid the task
2. Run command in dev
3. Wait for the task to run
4. Verify fix
5. Repeat with staging
6. Finally repeat with training

api/src/db/models/foreign/certificates.py
```diff
@@ -2,6 +2,40 @@
...
+DROP FOREIGN TABLE IF EXISTS legacy.tcertificates;
+CREATE FOREIGN TABLE legacy.tcertificates (
+    currentcertid TEXT OPTIONS (key 'true') NOT NULL,
+    previouscertid TEXT,
+    orgduns TEXT,
+    orgname TEXT,
+    expirationdate DATE,
+    certemail TEXT NOT NULL,
+    agencyid TEXT,
+    requestorlname TEXT,                                                                                                                                                                                                  requestorfname TEXT,
```
chouinar:
```suggestion
    requestorlname TEXT,
    requestorfname TEXT,
```

api/src/db/models/foreign/certificates.py
```diff
@@ -2,6 +2,40 @@
...
+This table needs to be manually created in lower environments in order to add the strip_zeros
+option to the is_selfsigned column due to null bytes being present for is_selfsigned. The
+exclude columns option does not work on UPDATES so it causes errors every time an UPDATE is
+attempted when there's a null byte in is_selfsigned. Since null bytes don't seem to be an
+issue in production for that column we can let that be automatically generated.
+Note: the first line with the ALTER is to ensure that the user running the task will have correct access
+to the table.
+
+The command:
+ALTER DEFAULT PRIVILEGES IN SCHEMA legacy GRANT SELECT ON TABLES TO app;
```
chouinar: I'd leave this and the note about it out, this is something we should correct with our setup so all tables already behave this way.

Issue #7142, opened 2025-11-18
Implement "PIV Required" message


ychoy:
### Summary

Design ticket: https://github.com/HHS/simpler-grants-gov/issues/7135

Figma: https://www.figma.com/design/zPlDWcYM396szw2yL3UXvk/Simpler-Grants-2025?node-id=8990-150893&p=f&t=yVycmxXeuP0eIQZd-0

See also this thread: https://betagrantsgov.slack.com/archives/C09QV6E3RJ8/p1765416907688459

<img width="2226" height="998" alt="Image" src="https://github.com/user-attachments/assets/279cf90b-9092-4197-9e0e-063b86b06b21" />

After the user authenticates, they are routed back to whichever SGG page they were on when they started the authentication process. So, the message needs to be displayed in a standard location that works in every page layout. Per USWDS guidance, design is recommending a modal: 

<img width="2600" height="1822" alt="Image" src="https://github.com/user-attachments/assets/90ed39da-faf7-49b4-b864-2a78ec27b59a" />

#### Content: 
> ### Your account requires additional identity verification.
> You must sign in with your government employee ID. Make  sure you’ve set up your Personal Identity Verification (PIV)  or Common Access Card (CAC) as a two-factor  authentication method.
> 
> [ Sign in using PIV/CAC ] 

#### To be refined: 

Is it possible to show this modal without using query parameter? Doing so would make the message accessible in browser history, and allow it to be loaded manually. It would be best if the message did NOT persist when the user refreshes the app,  or if they close the modal, navigate to another page, then navigate back. 

### Acceptance criteria

- [ ] When a user associated with an agency signs in without using PIV method, they see a modal displaying a message 
- [ ] The "PIV Required" modal contains a link to Login.gov that passes a parameter (e.g. `require_piv=true`) to the endpoint that sets `hsdp12` before routing the user to Login.gov 
- [ ] The site header's "Sign in" link does NOT include the `require_piv=true` parameter

brianlee-nava:
@crystabelrangel surfacing this comment here as you're refining the ticket. From the kickoff notes, there was mention that the error message could be in the form of toast message. Do we have existing UX/UI for toast messages in the pattern library?

crystabelrangel:
@brianlee-nava I have updated this to be an implementation ticket. We'll update the description once we have the designs complete. We do have an existing toast message pattern.

brianlee-nava:
[...]

PR #7885, opened 2026-01-09
[Issue #7142] Adding PIV warning modal and logic

glenblosser-nava:
## Summary

Work for #7142

## Changes proposed

- When an agency user logs in without using PIV, they are not logged in
- Instead, they are returned to their prior page, and a modal opens informing them of the PIV requirement
- The button in the modal includes a query string parameter to force piv login

## Context for reviewers

This will only work if PIV Required is enabled for the backend.

## Validation steps
In api/src/auth/login_gov_jwt_auth.py on line 43 set is_piv_required to default to True
Then deploy to Dev, and attempt to log in with an agency user. It should display the error modal and a login button that will force a piv login


doug-s-nava (commented):
having some difficulty testing, I think that the code you mentioned is not hit in local development - can you let me know how you enabled the login.gov flow from local?

frontend/src/app/[locale]/(base)/login/page.tsx
```diff
@@ -10,6 +10,12 @@ import { USWDSIcon } from "src/components/USWDSIcon";
 
 export default function Login() {
   const router = useRouter();
+  if (typeof window !== "undefined") {
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/login/page.tsx]: `useSearchParams` should make this easier https://nextjs.org/docs/app/api-reference/functions/use-search-params

frontend/src/app/api/auth/callback/route.ts
```diff
@@ -7,6 +7,12 @@ import { NextRequest } from "next/server";
 export async function GET(request: NextRequest) {
   const token = request.nextUrl.searchParams.get("token");
   if (!token) {
+    const errorMessage = request.nextUrl.searchParams.get("error_description");
+    if (
+      errorMessage === "Agency users must authenticate using a PIV/CAC card"
+    ) {
+      return redirect("/login?piverror=true");
```
doug-s-nava: I'd love to have this match the camelCase that we use for the rest of our query params (other than sortby but that's been kinda grandfathered in at this point)

frontend/src/components/application/editAppFilingName/PivRequiredModal.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+import { LoginLink } from "src/components/LoginButton";
+import { SimplerModal } from "src/components/SimplerModal";
+
+export const PivRequiredModal = () => {
+  const [showPivError, setShowPivError] = useState<string | null>("false");
+  const t = useTranslations("PivRequiredModal");
+  const pivModalRef = useRef<ModalRef | null>(null);
+  useEffect(() => {
+    setShowPivError(SessionStorage.getItem("showPivError"));
```
doug-s-nava: seems like this will only run once on mount, will want to test to make sure that this reruns when the session storage is updated

frontend/src/components/LoginButton.tsx
```diff
@@ -6,10 +6,18 @@ import { PropsWithChildren } from "react";
...
   className,
-}: { className?: string } & PropsWithChildren) {
+  queryParameters,
+}: {
+  className?: string;
+  queryParameters?: Record<string, string>;
+} & PropsWithChildren) {
+  let linkUrl = LOGIN_URL;
+  if (queryParameters) {
+    linkUrl += `?${new URLSearchParams(queryParameters).toString()}`;
```
doug-s-nava, line 16: seems like this will have unintended consequences in places like search.

frontend/src/components/SimplerModal.tsx
```diff
@@ -48,6 +48,7 @@ export function SimplerModal({
   children,
   onKeyDown,
   onClose,
+  isInitiallyOpen,
```
doug-s-nava: default to false?

doug-s-nava (commented):
haven't been able to test yet, but some things to look into once I can, and some code that could be punched up in the meantime

frontend/src/services/auth/session.test.ts
```diff
@@ -166,7 +166,7 @@ describe("createAndReturnSession", () => {
   });
 
   it("calls decrypt with the correct arguments and returns successfully", async () => {
-    decryptMock.mockReturnValueOnce({
+    decryptMock.mockReturnValue({
```
doug-s-nava: doesn't seem like there are any changes to this functionality or test otherwise, why this change?

frontend/src/app/api/auth/callback/route.ts
```diff
@@ -7,6 +7,12 @@ import { NextRequest } from "next/server";
 export async function GET(request: NextRequest) {
   const token = request.nextUrl.searchParams.get("token");
   if (!token) {
+    const errorMessage = request.nextUrl.searchParams.get("error_description");
+    if (
+      errorMessage === "Agency users must authenticate using a PIV/CAC card"
```
doug-s-nava: are we expecting the error message to come through in the query param like this, with spaces and the slash? I wouldn't be surprised if those sometimes get escaped. Is there another way we can identify the error? Or can we just look for a match for `PIV` or something like that, which would be a little risky, but maybe more reliable in the end?

doug-s-nava (commented):
I'd like to humbly request some unit testing around new behavior in:

* login page
* callback route
* piv required modal
* login button

frontend/src/app/api/auth/login/route.ts
```diff
@@ -1,13 +1,18 @@
...
 
-export function GET() {
+export function GET(request: NextRequest) {
   if (!environment.AUTH_LOGIN_URL) {
     return new NextResponse("AUTH_LOGIN_URL not defined", { status: 500 });
   }
-  redirect(environment.AUTH_LOGIN_URL);
+  const pivRequired =
+    request.nextUrl.searchParams.get("piv_required") ||
+    request.nextUrl.searchParams.get("require_piv");
```
doug-s-nava: we have control over these params, right? why do we need to support both?

frontend/src/components/LoginButton.tsx
doug-s-nava: I know we resolved this in March but I don't remember why and there is no note here. Can you leave a note as to what this is doing and why? My guess is that for some reason this is not picking up all of the query params from the page and forwarding them to the login route, but that seems like what it's doing?

frontend/src/services/auth/sessionUtils.ts
```diff
@@ -58,3 +60,10 @@ export async function deleteSession() {
   const cookie = await cookies();
   cookie.delete("session");
 }
+
+export function loginGovLogout(pivRequired = false) {
```
doug-s-nava: doesn't look like this is used anywhere

frontend/src/app/api/auth/callback/route.ts
doug-s-nava: doesn't look like this was ever addressed

frontend/src/app/api/auth/login/route.ts
glenblosser-nava: We don't, this is leftover code that I meant to remove. Fixed!

frontend/src/services/auth/sessionUtils.ts
glenblosser-nava: for the Login button: This is just passing along any query parameters passed to the button component. It is not reading the query parameters from the page and passing them along.

For the logout endpoint: We ultimately want to have a logout endpoint to make our user login/logout workflow work as expected. A previous iteration of this PIV required workflow used the logout endpoint to force a login, but it ultimately didn't work for the needed purpose. I've left the code for the logout endpoint in despite it being currently unused, because it works perfectly well for the future improvements we want to make and there's no sense re-doing the work in the future. If this needs to be pulled out, I can remove it

frontend/src/app/api/auth/callback/route.ts
doug-s-nava: rather than checking the error_description, let's set an "X-Unauth-Reason" or something header to pass this information instead

api/src/services/users/login_gov_callback_handler.py
```diff
@@ -130,7 +130,11 @@ def _validate_piv_requirement(user: User, x509_presented: bool | None) -> None:
                     "x509_presented": x509_presented,
                 },
             )
-            raise_flask_error(422, "Agency users must authenticate using a PIV/CAC card")
+            raise_flask_error(
+                422,
+                "Agency users must authenticate using a PIV/CAC card",
+                headers={"login_piv_required_error": "true"},
```
doug-s-nava: so setting the header ends up creating a query param on the other end? interesting...

doug-s-nava (commented):
this all checks out for me from a frontend perspective, would like to get a backend check before I approve cc @chouinar @dghazvini

infra/api/app-config/env-config/environment_variables.tf
```diff
@@ -17,6 +17,7 @@ locals {
     LOGIN_GOV_ENDPOINT               = "https://idp.int.identitysandbox.gov/"
     LOGIN_GOV_JWK_ENDPOINT           = "https://idp.int.identitysandbox.gov/api/openid_connect/certs"
     LOGIN_GOV_AUTH_ENDPOINT          = "https://idp.int.identitysandbox.gov/openid_connect/authorize"
+    LOGIN_GOV_LOGOUT_ENDPOINT        = "https://idp.int.identitysandbox.gov/openid_connect/logout"
```
chouinar: I don't see any usage of this env var - is this needed for this PR?

api/src/auth/login_gov_jwt_auth.py
```diff
@@ -126,34 +130,42 @@ def _refresh_keys(config: LoginGovConfig) -> None:
...
+        "acr_values": config.acr_value,
+        "scope": config.scope,
+        # These are statically defined by the spec
+        "prompt": "select_account",
+        "response_type": "code",
+    }
+    if redirect_params.piv_required:
+        url_params["acr_values"] = (
+            config.acr_value + " http://idmanagement.gov/ns/assurance/aal/2?hspd12=true"
+        )
```
chouinar: Could we move the definition of this URL to a variable at the top of the file / add a link to the login.gov docs so when we look back at this later, we know what it means.

api/src/auth/login_gov_jwt_auth.py
```diff
@@ -36,7 +40,7 @@ class LoginGovConfig(PydanticBaseEnvConfig):
     client_id: str = Field(alias="LOGIN_GOV_CLIENT_ID")
     acr_value: str = Field(alias="LOGIN_GOV_ACR_VALUE", default="urn:acr.login.gov:auth-only")
     scope: str = Field(alias="LOGIN_GOV_SCOPE", default="openid email x509:presented")
-    is_piv_required: bool = Field(alias="IS_PIV_REQUIRED", default=False)
+    is_piv_required: bool = Field(alias="IS_PIV_REQUIRED", default=True)
```
chouinar: If we default this to true for all environments, that means no one who is associated with an agency will be able to login. I think a decent number of engineers have accounts in staging associated with an agency.

This would also enable it in training, we'd probably want to let folks like Chris Weekley know that they won't be able to look at anything there.

api/src/auth/auth_utils.py
```diff
@@ -64,7 +64,9 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> flask.Response:
                     )
 
                 return response.redirect_response(
-                    get_final_redirect_uri("error", error_description=message)
+                    get_final_redirect_uri(
+                        "error", error_description=message, additional_params=dict(e.headers)
```
chouinar: I don't think we want to use the headers to pass this around - we're not actually using them as headers here, just as a place to put some info on the error object. The error we throw does actually have a dict for that called `extra_data`.

Could we instead do something like this:
* Modify `raise_flask_error` to take in an optional `extra_data` field - and if it's set, merge it into the one we already set for validation issues
* Here, only grab out the fields that you want in the redirect
* Rather than additional_params for the redirect uri, just add `login_piv_required_error` to that function and set it the same as you are now. Don't have it created 3-jumps of error handling away, do it here.

api/src/auth/login_gov_jwt_auth.py
glenblosser-nava: This is intended to be switched back to False after review, so will not release in this state. I left it as True so we can deploy to dev for testing if needed

infra/api/app-config/env-config/environment_variables.tf
glenblosser-nava: No, I meant to remove all references to it but missed one. Thank you!

chouinar (approved):

PR #7899, opened 2026-01-10
[unticketed] Specify platform for api container.

freddieyebra:
## Summary

Fixes / Work for unticketed

## Changes proposed

Specify the platform for avoiding discrepancies in local dev builds.
 
## Context for reviewers

to prevent the following occurring in local dev. 
```
ERROR [internal] load metadata for ghcr.io/hhs/python-base-image:8b8ce02@sha256:70a2337a709022d
http://ghcr.io/hhs/python-base-image:8b8ce02@sha256:70a2337a709022d
```

## Validation steps

- checkout branch
- stop api container if running
- run `make start` or `make run-logs`
- ensure app runs as expected.

### DIFF
--- a/api/docker-compose.yml
+++ b/api/docker-compose.yml
@@ -72,6 +72,7 @@ services:
       - default
 
   grants-api:
+    platform: linux/amd64
     build:
       context: .
       target: dev


api/docker-compose.yml
```diff
@@ -72,6 +72,7 @@ services:
       - default
 
   grants-api:
+    platform: linux/amd64
```
chouinar, line 75: Before we can consider merging this, we'd need everyone to test it as I don't know what would happen if we ran this on Windows. I remember we used to have this on a prior project, but it caused us a lot of issues and we removed it.

This isn't a "just make a quick change" sort of things, we would want to figure out the right way of doing things, I'm guessing the way building works as a whole would need to be looked at.

freddieyebra:
wont do since underlying local build issue resolved.

Issue #7921, opened 2026-01-12
Update all modals to have matching padding


myduong-navapbc:
### Summary

Per discussion with design, Modals should have a `24pt grid`. or 3 rems, 24 pixels.

### Acceptance criteria

update global styles for modals

Issue #7729, opened 2025-12-31
build the transfer application ownership modal


ErinPattisonNava:
### Summary

We want to build the transfer application ownership modal.  To break this in to pieces, we're going to first build out the modal without hooking it up to actions so the look and feel  are right, then in a follow up ticket we'll connect it to the apply page.

In order to test this, ideally we'd use storybook but I don't believe storybook supports translations.  You can create a temp button on the apply page to pop up the modal, but don't commit this button!  For testing, include instructions on where they can manually add the button locally for testing purposes.

Design in Figma: https://www.figma.com/design/zPlDWcYM396szw2yL3UXvk/Simpler-Grants-2025?node-id=8410-72140&m=dev

### Acceptance criteria

- [x] Build out the transfer application ownership modal component so that it matches the mock
   - [x] Dropdown should list the organizations the user is a member of
      - [x] If there is an error retrieving organizations, leave dropdown empty
   - [x] Return to application button should return to the application
   - [x] Transfer ownership button is disabled 
- [x] Make sure the temporary button isn't committed! 
- [x] unittests
- [x] design review before submitting PR 
- [x] no UAT

myduong-navapbc:
removed shared footer AC's per redesign

PR #7922, opened 2026-01-12
[Issue #7729] Transfer ownership modal

myduong-navapbc:
## Summary

Work for #7729  

## Context for reviewers

You will need to create a button that triggers the modal in order to test it.
To test this locally, i suggest creating a throwaway client component.
you can put this anywhere.

```

"use client";

import { useRef } from "react";
import { ModalRef, ModalToggleButton } from "@trussworks/react-uswds";

import { TransferOwnershipModal } from "src/components/application/transferOwnership/TransferOwnershipModal";

export const ThrowAwayComponent = () => {
  const modalRef = useRef<ModalRef>(null);

  return (
    <>
      <ModalToggleButton
        modalRef={modalRef}
        opener
        className="usa-nav__link font-sans-2xs display-flex text-normal border-0"
        data-testid="sign-in-button"
      >
        OPEN TRANSFER OWNERSHIP
      </ModalToggleButton>
      <TransferOwnershipModal modalId={""} modalRef={modalRef} />
    </>
  );
};

```

Screenshot 2026-01-13 at 11 45 04 AM

NOTE:
the dropdown will throw if you are not logged into a user with an org (if you put the component on a non auth page).

The modal header does not match design, I opt to use the existing header sizes that all other modals use

The modal padding does not match design, i opt to use the existing padding; ticket to update all modal padding to match  design here https://github.com/HHS/simpler-grants-gov/issues/7921

## Validation steps

- [ ] unit test pass
- [ ] matches designs


freddieyebra (commented):
This looks good! I verified the following locally:

No orgs
Screenshot 2026-01-16 at 12 17 09 PM

logged in + has at least one org
Screenshot 2026-01-16 at 12 18 15 PM

org select state works
Screenshot 2026-01-16 at 12 18 25 PM

I am wondering if Is it intentional for the large white space at the top of the modal for the logged in user with orgs.

myduong-navapbc:
> This looks good! I verified the following locally:
> 
> No orgs Screenshot 2026-01-16 at 12 17 09 PM
> 
> logged in + has at least one org Screenshot 2026-01-16 at 12 18 15 PM
> 
> org select state works Screenshot 2026-01-16 at 12 18 25 PM
> 
> I am wondering if Is it intentional for the large white space at the top of the modal for the logged in user with orgs.

That is intentional, and maybe something we may want to revisit- since that white space is kind of distracting to me-

The idea was that the contents of the modal wouldn't suddenly shift if an error occurred.

PR #7961, opened 2026-01-13
[Draft] Add training env to New Relic config + adjust expected errors

chouinar:
## Changes proposed
Adjusts error codes that New Relic will consider an error
Adds training env to New Relic config

### DIFF
--- a/api/newrelic.ini
+++ b/api/newrelic.ini
@@ -183,15 +183,15 @@ error_collector.expected_classes =
 # 405: Method Not Allowed
 # 406: Not Acceptable - Invalid Accept header (API does not support it)
 # 415: Unsupported Media Type - Invalid media types for upload
-# 503: Service Unavailable - Temporary service unavailability. High volume should trigger a New Relic alarm.
+# 416: Range Not Satisfiable - We don't support ranges, if our API returns this, is on the user
+# 422: Unprocessable Content - API Request validation issues fall under this
 #
 # Status codes that we may want to ignore in the future:
 #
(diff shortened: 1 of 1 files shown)


api/newrelic.ini
```diff
@@ -183,15 +183,15 @@ error_collector.expected_classes =
...
+# 422: Unprocessable Content - API Request validation issues fall under this
 #
 # Status codes that we may want to ignore in the future:
 #
 # - 400: Bad Request - Validation and extra parameter errors. Unclear if we want to catch pydantic ValidationErrors here so we're keeping them for now. Instead, selectively ignore specific error classes above.
-# - 422: Unprocessable Entity - Haven't seen these yet so we'll report them.
 # - 504: Gateway Timeout - We do not expect any 504s to be thrown from the API server itself. This should come from the API Gateway that sits in front of it.
 #
-error_collector.ignore_status_codes = 401 402 403 404 405 406 415 503
+error_collector.ignore_status_codes = 401 402 403 404 405 406 415 416 422
```
mdragon, line 194: Just making sure I understand this list. This is status codes we want to opt out of the automatic handling that NR sees error statuses on the response and flags them as an error vs regular log entry?

chouinar: @mdragon - Effectively in the APM section on New Relic for our services has an error rate / we also have an alert connected to it. It currently thinks if the API returns a 422, that's some sort of problem (those are largely request validation issues). Conversely, a 503 is seen as not an error - copied that from Mass PFML and I don't think we want that although I don't think our API ever returns it.

PR #334, opened 2023-07-25
[ADR]: Back-end production server#245

daphnegold:
## Summary
Fixes #245 

## Changes proposed
ADR proposing Python production server and implementation.

PR #7965, opened 2026-01-13
[Issue #334]: add lychee markdown link checker 

prasnava:
## Summary

Added a markdown link checker ci job. Lychee will check each link in the documentation to ensure it still exists.

## Validation steps

Ensure the github actions workflow is successfully running: [here](https://github.com/HHS/simpler-grants-gov/actions/workflows/ci-docs.yml)

The pipeline will fail as we multiple links to fix, but it should print the summary report and a detailed list of what to fix.

### DIFF
--- a/bin/lint-markdown
+++ b/bin/lint-markdown
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -euo pipefail
 
 # To make things simpler, ensure we're in the repo's root directory (one directory up) before
 # running, regardless where the user is when invoking this script.


prasnava:
The nava template infra repo is using the npm package markdown-link-check which is slower than lychee.  Will create a ticket there to update it there.

.github/workflows/ci-docs.yml
```diff
@@ -0,0 +1,21 @@
+name: CI Documentation Checks
+
+on:
+  push:
+    branches:
+      - main
```
mdragon: Wondering if this should be more of a weekly thing or something? Seems like a lot of crawling to have this fire on every commit to main and PR?

prasnava: Yep, will change it to weekly.  The pipeline is working [here](https://github.com/HHS/simpler-grants-gov/actions/runs/20970134584/job/60271031886) It prints the summary table: 

# Summary
| Status         | Count |
|----------------|-------|
| 🔍 Total       | 1670  |
| ✅ Successful  | 1358  |
| ⏳ Timeouts    | 3     |
| 🔀 Redirected  | 99    |
| 👻 Excluded    | 10    |
| ❓ Unknown     | 0     |
| 🚫 Errors      | 200   |
| ⛔ Unsupported | 0     |

.github/workflows/ci-docs.yml
```diff
@@ -0,0 +1,21 @@
+name: CI Documentation Checks
+
+on:
+  schedule:
+    # Run weekly on Monday at 9:00 AM EST
+    - cron: "0 14 * * 1"
+  workflow_dispatch:
```
prasnava, line 7: Updated the workflow to run as a weekly cron job.

.github/workflows/ci-docs.yml
```diff
@@ -0,0 +1,21 @@
+name: CI Documentation Checks
```
mdragon: Will this be expanded to do more than link checks and grow into this name? Or should we rename to be more narrow?

prasnava: Changed the name to: CI Documentation Linker Checks

We can rename back down the road as we add more checks to it.

bin/lint-markdown.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
sean-navapbc, line 1: Missing shell safety flags in bin/lint-markdown.sh

  Add at the top:
  #!/bin/bash
  set -euo pipefail

sean-navapbc (changes requested):
Looks good except one minor change:

Suggestions

  3. Minimal lychee config - Consider adding common exclusions to .github/workflows/lychee.toml to reduce false positives:

  timeout = 30
  exclude = ["localhost", "127.0.0.1", "example.com", "example.org"]

  4. Consider adding path filters to the workflow - Right now it only runs on schedule/manual. You might want to also trigger on
  PRs that modify .md files:

  on:
    pull_request:
      paths:
        - "**/*.md"
    schedule:
      - cron: "0 14 * * 1"
    workflow_dispatch:

bin/lint-markdown.sh
prasnava: added

prasnava:
> Looks good except one minor change:
> 
> Suggestions
> 
> 3. Minimal lychee config - Consider adding common exclusions to .github/workflows/lychee.toml to reduce false positives:
> 
> timeout = 30 exclude = ["localhost", "127.0.0.1", "example.com", "example.org"]
> 
> 4. Consider adding path filters to the workflow - Right now it only runs on schedule/manual. You might want to also trigger on
>    PRs that modify .md files:
> 
> on: pull_request: paths: - "**/*.md" schedule: - cron: "0 14 * * 1" workflow_dispatch:

Updated the lychee.toml file with the exclude, right now we have too many fixes to make to make this auto run for each pull request. I think we should work on fixing all the links in a future ticket then turn on the auto run on pull request so we don't block development.

Issue #7562, opened 2025-12-18
Create an endpoint for adding forms to a competition


chouinar:
### Summary

To avoid manual DB work to setup forms on a competition, we want to be able to do these operations via our API endpoint.

We want to create an endpoint `PUT /alpha/competitions/{competition_id}/forms` that will take in a list of forms, and whether they're required and update the competition accordingly. Importantly, it will replace any forms that might already be there (the endpoint expects the full set of forms you want on the competition).

### Request
```json
{
  "forms": [
     {"form_id": "<some uuid>", "is_required": true}
   ]
}
```
We'll take in a list of forms, make the minimum length be 1. Every form object must have a form ID and whether it is required.

### Response
Make the response the same as the GET competition endpoint, can reuse the same schema and return the competition itself.

### Auth
Exact same as what we implemented in https://github.com/HHS/simpler-grants-gov/issues/7560 - a user must have the `manage_competition` privilege.

### Logic
Fetch the competition, 404 if not found.

For each form, fetch it from the DB (fine to fetch all of them at once in a single query). If ANY of them are not found, 404.

For each form, we want to create a competition form record for the competition, BUT may need to account for it already existing on the competition. If it already exists, you should update it (although the only update would be whether it's required, if that's the same, nothing needs to be done). This'll require a bit of iteration over the lists provided by the user, and what is currently in the data. Prefer slightly more verbose readable code over clever short code, we're at most dealing with a dozen or so form objects, no need for anything significant in optimizing the logic.

Return the competition, with now updated competition forms.

### Acceptance criteria

- [ ] Endpoint created
- [ ] Tests created for various scenarios including significant form changes, passing in forms that are the same as what it already has, and everything in-between. Also auth / general request validation cases.

ychoy:
It's possible that engineers from @kevlyu @brianlee-nava teams will be able to pick this work up soon. so I'm moving it to general onboarding. cc @chris-kuryak

PR #7987, opened 2026-01-14
[Issue #7562] Add forms to competiton

babebe:
## Summary

Fixes / Work for #7562  

## Changes proposed

- Added new endpoint: `PUT /alpha/competitions/{competition_id}/forms`
- Implemented request payload validation
- Fetch and validate competition — return 404 if not found
- Fetch requested forms — return 404 if any are not found or deprecated
- Enforced user access check with `manage_competition` privilege
- If `competition_form` exists — update` is_required`
- If `form_id` is not synced to competition — create a new` competition_form`
- Remove existing `competition_form` records not included in the request
- Perform all create/update/delete operations atomically
- Added test coverage for add, update, remove, and authorization scenarios

## Context for reviewers

To avoid manual DB work to setup forms on a competition, we want to be able to do these operations via our API endpoint.

We want to create an endpoint PUT /alpha/competitions/{competition_id}/forms that will take in a list of forms, and whether they're required and update the competition accordingly. Importantly, it will replace any forms that might already be there (the endpoint expects the full set of forms you want on the competition).

## Validation steps

Run existing tests


api/src/services/competition_alpha/put_competition_forms.py
```diff
@@ -0,0 +1,85 @@
...
+def set_competition_forms(
+    db_session: db.Session, user: User, competition_id, json_data: dict
+) -> Competition:
+    # Check user access
+    verify_access(user, {Privilege.MANAGE_COMPETITION}, None)
+
+    competition = get_competition(db_session, competition_id)
+
+    if not competition:
+        raise_flask_error(404, f"Competition with ID {competition_id} not found.")
```
chouinar: I think we want something like this. We don't need the `is None` check, `get_competition` has that.

The verify_access bit is a bit of future proofing. Eventually, for a grantor to modify this, they'd need the `manage_competition` privilege in the agency that owns the opportunity. Just to avoid us forgetting this later, we should enforce that restriction now. It doesn't matter for our internal role as they aren't checked against the agency, so this should still work.

```suggestion
    competition = get_competition(db_session, competition_id)
    
    # Check user access
    verify_access(user, {Privilege.MANAGE_COMPETITION}, competition.opportunity.agency)
```

api/src/services/competition_alpha/put_competition_forms.py
```diff
@@ -0,0 +1,85 @@
...
+
+    requested_forms = json_data["forms"]
+    requested_form_ids = [f["form_id"] for f in requested_forms]
+
+    # Validate all forms exist
+    forms_db = db_session.query(Form).filter(Form.form_id.in_(requested_form_ids)).all()
```
chouinar: We don't have any deprecated forms at the moment, but also shouldn't let those be set. Alternatively, can fetch them, but do this check in the reconcile function to give a more specific error.

```suggestion
    forms_db = db_session.query(Form).filter(Form.form_id.in_(requested_form_ids), Form.is_deprecated.isnot_(True)).all()
```

api/src/services/competition_alpha/put_competition_forms.py
```diff
@@ -0,0 +1,85 @@
...
+        form_id = form_data["form_id"]
+        is_required = form_data["is_required"]
+
+        target_form_ids.add(form_id)
+
+        existing_cf = existing_by_form_id.get(form_id)
+
+        if existing_cf:
+            if existing_cf.is_required != is_required:
+                existing_cf.is_required = is_required
```
chouinar: Even if they match already, can just set it. I _think_ SQLAlchemy will see that nothing changed and won't even bother updating the row, but we should double-check that. This really only matters for the `updated_at` timestamp.

```suggestion
            existing_cf.is_required = is_required
```

api/src/services/competition_alpha/put_competition_forms.py
```diff
@@ -0,0 +1,85 @@
...
+        else:
+            cf = CompetitionForm(
```
chouinar, line 46: Could we add a bit of logging to this, and the if statement above, just something like:
```py
logger.info("Adding competition form to competition", extra={"form_id": form_id, "is_required": is_required})
```

api/src/services/competition_alpha/put_competition_forms.py
babebe: Updated. Test confirms that no update occurs.

api/src/services/competition_alpha/put_competition_forms.py
babebe: Added more logs

api/src/services/competition_alpha/put_competition_forms.py
babebe: Updated

api/src/services/competition_alpha/put_competition_forms.py
babebe: Updated

chouinar (approved):

Issue #7673, opened 2025-12-24
Negative test scenario cases feature file for primary Apply-related user flows


ychoy:
### Summary

Negative test scenario cases feature file is documented

Parent #6146 

### Acceptance criteria

- [x] Negative scenario cases feature file (primary Apply-related user flows) test cases and ranked in order of priority for testing is documented @aoysimmons-GH 
- [x] Negative scenario cases feature file is reviewed @doug-s-nava 
- [x] Negative scenario cases feature file is refined based on review feedback @aoysimmons-GH 
- [ ] Negative scenario cases feature file is approved @doug-s-nava 
- [ ] Negative scenario cases feature file is merged to main @aoysimmons-GH

PR #7997, opened 2026-01-14
Add negative path scenarios for application workflow

aoysimmons-GH:
This feature file contains negative path scenarios for the application workflow, detailing various failure cases such as application shell loading issues, backend submission errors, save failures, invalid file uploads, and workflow status retrieval failures.

## Summary

Fixes / Work for #7673 

## Changes proposed

Move file to /frontend/tests/e2e/apply and change file name to negative-path.feature
Remove scenario 3 as per it is positive scenario- 3. Data Persist on Refresh – Data Loss After Refresh

## Context for reviewers

 Ref. to the peer-review of this ticket: https://github.com/HHS/simpler-grants-gov/pull/7836/files

### DIFF
--- a/frontend/tests/e2e/apply/features/negative-path.feature
+++ b/frontend/tests/e2e/apply/features/negative-path.feature
@@ -49,7 +49,7 @@ Scenario: Application shell fails to load when starting a new application
   ###########################################################################
   # 5. Workflow Status – Status Fails to Load
   ###########################################################################
-  Scenario: Workflow status cannot be retrieved
+  Scenario: Application status cannot be retrieved
     Given the user views the application summary
     When the system loads workflow status
     Then an error occurs


frontend/tests/e2e/apply/features/negative-path.feature
```diff
@@ -0,0 +1,57 @@
...
+  Scenario: User uploads invalid file type
+  Given the user is on the file upload step
+  When the user uploads a ".exe" file
+  Then the system rejects the file
+  And the user sees an error "Invalid file type"
+
+  ###########################################################################
+  # 5. Workflow Status – Status Fails to Load
+  ###########################################################################
+  Scenario: Workflow status cannot be retrieved
```
doug-s-nava: We should avoid the term "workflow" here
```suggestion
  Scenario: Application status cannot be retrieved
```

Issue #7691, opened 2025-12-26
Enforce deployment order


mdragon:
### Summary

Currently the deploys all race in parallel when doing a Release deploy to Prod (or if you touch multiple services/apps in a single branch merge), Instead, they should deploy in a consistent order API -> FE -> NOFO -> Analytics. However since each deploy currently does linting, testing, vulnerability scans, etc we should still allow as much of that as possible to run in parallel and only enforce ordering for the migration and deploy steps to keep the deploy time window as short as possible.

### Acceptance criteria

- [ ] Developers can rely on API deploying before everything else, so backwards compatibility can be centralized in that single service and everything else deploying can safely assume any changes they rely on from API have already been deployed.
- [ ] Jobs run as parallel as possible (lint, tests, scans) up to the point where they must be serialized (migration and deploy for each service in order)
- [ ] Failures to deploy an earlier service should halt further service deployments.

PR #7998, opened 2026-01-14
[Issue #7996] Improvements following testing of deploy order CD process

### LINKED ISSUE #7996: Follow up for enforcing deploy order
### Summary

Following up on #7691 refine the process a bit based on testing

### Acceptance criteria

- [ ] NOFOs is skipped for now
- [ ] Analytics can start deploying as soon as API is deployed
- [ ] Environments for workflow dispatch include dev

mdragon:
## Summary

Fixes #7996  

## Changes proposed

- Allow deploys to dev
- Don't deploy NOFOs (they deploy manually)
- Allow Analytics to kick off when API is done, there's no reason for it to wait for FE deploy

## Context for reviewers

Follow up to #7691 with tweaks following testing.

## Validation steps

Action succeeds

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -23,14 +23,6 @@ on:
 concurrency: cd-${{inputs.app_name}}-${{ inputs.environment }}
 
 jobs:
-  force-breaking-on-api:
-    name: Force breaking for API to test Release Deploy
-    runs-on: ubuntu-22.04
-    if: ${{ inputs.app_name == 'api' }}
-    steps:
-      - name: Exit with error
-        run: exit 1
-
   # Don't need to call the build-and-publish workflow since the database-migrations
   # workflow already calls it
   database-migrations:


.github/workflows/cd-release.yml
```diff
@@ -127,7 +136,7 @@ jobs:
   # 4. Deploy Analytics last, after NOFOs completes successfully
   deploy-analytics:
     name: Deploy Analytics
-    needs: [analytics-checks, analytics-vulnerability-scans, deploy-nofos]
+    needs: [analytics-checks, analytics-vulnerability-scans, deploy-api]
```
chouinar, line 139: We could make analytics go without the API potentially. The only dependency between them to consider at the moment is the CSVs we generate to send over. But that's a backend process anyways that we probably wouldn't really consider, and only runs early in the morning when we wouldn't be deploying.

It doesn't matter a ton regardless, but if we want some parallelization.

mdragon: I was thinking if we made data model changes in API and then needed that to be applied so Analytics stuff wouldn't fail. Now there's a limited window there where Analytics could be ahead of API and get fired to actually run a data pull, but this felt safer. I guess we could hit the opposite, we haven't updated Analytics but API has updated and so the current Analytics version fails.

chouinar (approved):

PR #8022, opened 2026-01-15
[Issue #7968] Remove unused agency fields from Opportunity Summary

### LINKED ISSUE #7968: Follow up - Removing unused agency fields from Opportunity Summary Table
### Summary

For the necessary second deploy of removing DB columns #7090 

### Acceptance criteria

- [ ] agency_phone_number removed from Opportunity Summary table
- [ ] agency_code, agency_name removed from Opportunity Summary table
- [ ] Database migration created
- [ ] Updated transformation jobs to no longer populate these fields.

hao10282025-sudo:
## Summary

Fixes #7968  

## Changes proposed

In our OpportunitySummary table we want to remove the following columns:

- agency_phone_number
- agency_code
- agency_name

## Context for reviewers

This is the second deploy changes for the main ticket #7090

## Validation steps

- agency_phone_number removed from Opportunity Summary table
- agency_code, agency_name removed from Opportunity Summary table
- Database migration created
- Updated transformation jobs to no longer populate these fields.


api/src/data_migration/transformation/transform_util.py
```diff
@@ -251,14 +251,11 @@ def transform_opportunity_summary(
...
     target_summary.funding_category_description = source_summary.oth_cat_fa_desc
     target_summary.applicant_eligibility_description = source_summary.applicant_elig_desc
-    target_summary.agency_name = source_summary.ac_name
     target_summary.agency_email_address = source_summary.ac_email_addr
     target_summary.agency_email_address_description = source_summary.ac_email_desc
     target_summary.can_send_mail = convert_yn_bool(source_summary.sendmail)
 
     target_summary.summary_description = source_summary.description
-    target_summary.agency_code = source_summary.agency_code
-    target_summary.agency_phone_number = source_summary.agency_phone_number
```
chouinar, line 261: Hmm, we probably should have removed these in the first part of the PR. If this bit of code is running when we run the migration that drops the columns, it could fail. We should probably have the docs say to remove all setting of a field in the first PR for this.

This isn't that big of a deal, I'd guess there's about a 10 second window every hour where this could be an issue, and even if it happens, the job will run again the next hour uneventfully.

FYI @mdragon - When we do the next prod deploy after this, it could cause a very minor set of failures in the transformation job. If we wanted to be incredibly safe we can just make sure the deploy doesn't happen exactly on top of this. Rough guess, this happens at about xx:05 after the hour, which if we start the deploy at noon, shouldn't overlap.

hao10282025-sudo: Ok interesting.
I think I got the general idea, will understand it more from the code, and update the docs with a new PR.

Issue #6964, opened 2025-11-06
Hook up activity history to Required and Conditionally-Required form tables and Remove `Last Updated By` column on Application page


chris-kuryak:
### Summary

The tables on the Application page do not currently show any activity history in the `last updated` and `last updated by`. `last updated` should be fairly straightforward since we capture that information already. However, `last updated by` may require extra backend work to support. Therefore, we will focus on the light lift for now and focus on supporting `last updated`.

- `Last updated` shows date and time and timezone when the form was last updated.
- Remove `last updated by` column from table to align with new design

Formatting should be aligned with design from #7816 

This work does NOT include adding sorting to the tables.

Activity History Tech Spec for reference: https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2037416076/Tech+Spec+Activity+History

<img width="1055" height="425" alt="Image" src="https://github.com/user-attachments/assets/6a6139f2-4e88-4916-bee5-86a554980e7a" />

<img width="1058" height="384" alt="Image" src="https://github.com/user-attachments/assets/e0a0965f-6c29-445f-b3a7-35e2c25075b4" />

### Acceptance criteria
- [x] Remove `last updated by` column from the required and conditionally-required form tables to align with new design
- [x] When a form has never been updated (saved), no information is shown in the `Last Updated`  field and instead show the existing dashes
- [x] When a user updates a required form, the required forms table shows last updated (date, time, timezone)
   - [ ] Formatting and timezone should match what is currently shown in the application history
- [x] When a user updates a conditionally-required form, the conditionally-required forms table shows last updated (date, time, timezone)
- [ ] Unittests
- [x] Review with Design before PR is submitted

doug-s-nava:
* timestamp for "last updated" should be present on each form - looks like that is coming down in the payload from the application endpoint so adding that should be easy
* "updated by" - would need more detail on where we should be getting that from. The linked tech spec doesn't speak to a payload shape for the activity history endpoint. Would want to see that before assessing feasibility

chris-kuryak:
@chouinar I think you said even with the new Activity History work we are implementing this sprint, that the "Last updated by" would be difficult to show in these tables, right?

chris-kuryak:
Requires another sync and alignment on solution from a technical perspective

ErinPattisonNava:
[...]

PR #8050, opened 2026-01-16
[Issue #6964]auditcolumns update for forms table

chay-REIsys:
## Summary

Work for #6964  

## Changes proposed

Application Forms Table has been updated (both required and conditional)

- Remove Updated by
- Populating Updated Date column with most updated date for the form

## Context for reviewers

while creating or editing an application for an oppurtunity
data in the forms table (both  required and conditional) should have updated date for the respective form also remove update by column as per figma

## Validation steps

create application
access any form update any value and check update date column reflects the same.

### DIFF
--- a/frontend/src/components/application/ApplicationFormsTable.tsx
+++ b/frontend/src/components/application/ApplicationFormsTable.tsx
@@ -3,7 +3,7 @@ import {
   ApplicationFormDetail,
 } from "src/types/applicationResponseTypes";
 import { CompetitionForms } from "src/types/competitionsResponseTypes";
-import { formatTimestamp } from "src/utils/generalUtils";
+import { getModifiedTimeDisplay } from "src/utils/generalUtils";
 
 import { useTranslations } from "next-intl";
 import Link from "next/link";
@@ -216,7 +216,9 @@ const ApplicationTable = ({
               />
             </td>
             <td data-label={t("updated")}>
-              <div>{formatTimestamp(form.updated_at)}</div>
+              <div>
+                {getModifiedTimeDisplay(form.updated_at, form.created_at, "--")}
+              </div>
             </td>
           </tr>
         ))}
--- a/frontend/src/types/applicationResponseTypes.ts
+++ b/frontend/src/types/applicationResponseTypes.ts
@@ -50,6 +50,7 @@ export interface ApplicationFormDetail {
   is_required: boolean;
   is_included_in_submission?: boolean | null;
   updated_at: RegexMatchedString<typeof iso8601Date>;
+  created_at: RegexMatchedString<typeof iso8601Date>;
 }
 
 export interface ApplicationDetail {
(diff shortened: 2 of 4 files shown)


myduong-navapbc:
Please add unit test to cover the function you created

frontend/src/components/application/ApplicationFormsTable.tsx
```diff
@@ -297,6 +291,19 @@ const InstructionsLink = ({
   );
 };
 
+const formatTimestamp = (time: string) => {
```
ErinPattisonNava: Instead of copying the code from the history table, please pull out a utility that can be used in both places.

frontend/src/types/applicationResponseTypes.ts
```diff
@@ -4,6 +4,11 @@ import { FormValidationWarning } from "src/components/applyForm/types";
 import { Attachment } from "./attachmentTypes";
 import { Competition } from "./competitionsResponseTypes";
 import { FormDetail } from "./formResponseTypes";
+import { RegexMatchedString } from "./generalTypes";
+
+// ISO 8601 date, ie "1978-10-08T19:23:01.165Z"
+const iso8601Date =
```
ErinPattisonNava: In my original review I didn't realize this is also copied code.  Can you refactor this into generalTypes so it is more reusable?

chaitanya-duddella: pushed new changes for this refactor

frontend/src/components/application/ApplicationFormsTable.tsx
```diff
@@ -218,10 +216,7 @@ const ApplicationTable = ({
               />
             </td>
             <td data-label={t("updated")}>
-              <div> -- </div>
-            </td>
-            <td data-label={t("updatedBy")}>
-              <div> -- </div>
+              <div>{formatTimestamp(form.updated_at)}</div>
```
ErinPattisonNava: I think we need to keep something here, as per the ticket `--` should still show if a form hasn't been updated yet.  Whether the API should return empty for this value or if the frontend should compare created_at to updated_at is being discussed, but either way I think we'll need to change something here -- and it would be a good use case to add to the unit tests.

From the ticket: 
`When a form has never been updated (saved), no information is shown in the Last Updated field and instead show the existing dashes`

chaitanya-duddella: image

frontend/src/types/generalTypes.ts
```diff
@@ -1,5 +1,8 @@
 import { FrontendErrorDetails } from "src/types/apiResponseTypes";
 
+export const iso8601Date =
+  /^\d{4}(-\d\d(-\d\d(T\d\d:\d\d(:\d\d)?(\.\d+)?(([+-]\d\d:\d\d)|Z)?)?)?)?$/;
+
```
freddieyebra, line 5: nice!

Issue #7402, opened 2025-12-09
Feature files test cases for and/or functionality


ychoy:
### Summary

Writing feature files test cases for the "and/or" functionality 

### Acceptance criteria

- [ ] Feature files test cases are documented for the "and/or" functionality

ychoy:
@aoysimmons-GH please check with @eric-serrato95 when this test plan is ready for review

aoysimmons-GH:
@eric-serrato95 Can you please review the test scenarios?

PR #8053, opened 2026-01-16
Add feature tests for 'and/or' search term switch

aoysimmons-GH:
This feature tests the functionality of the 'and/or' search term switch, covering various scenarios including URL verification, search defaults, result validation, and persistence after navigation.

## Summary

Fixes / Work for #7402 
## Context for reviewers

 Please review the test scenarios

### DIFF
--- a/frontend/tests/e2e/search/features/filterDrawer/andor-searchtermswitch.feature
+++ b/frontend/tests/e2e/search/features/filterDrawer/andor-searchtermswitch.feature
@@ -1,0 +1,554 @@
+@7079
+Feature: "and / or" search term switch
+
+#Summary
+  ###########################################################################
+     #Every scenario below tests the search verb switch between AND and OR conditions in various situations:
+    # - Verifying default selection
+    # - Ensuring no search term is entered when switching with empty input
+    # - Checking URL updates when switching between AND and OR with/without search terms
+    # - Validating search results for single and multiple terms when switching
+    # - Testing persistence of search verb switch after page reload and navigation
+    # - Confirming accurate search result counts when switching
+    # - Handling special characters and long input validations
+
+  ###########################################################################
+# Every scenario must run after login and click 'Search' tab
+  ###########################################################################
+Background: Navigate to search page after login
+    Given I click 'Search' tab
+
+Scenario: Verify the URL of the search page
+    Then I see 'url'
+        |url|
+        |/search|
+
+Scenario: Verify the search defaults to AND condition
+    Then I see 'search-verb-switch' selected
+        |search-verb-switch|
+        |Must include all words (ex. transportation AND safety)|
+
+Scenario: Verify no search term entered when switching from AND to OR
+    When I clear 'Search' field
+    And I click 'Search' button
+    Then I see 'url'
+        |url|
+        |/search|
+    And I select 'May include any words (ex. transportation OR safety)' at 'search-verb-switch'
+    Then I see 'url'
+        |url|
+        |/search|
+    And I select 'Must include all words (ex. transportation AND safety)' at 'search-verb-switch'
+    Then I see 'url'
+        |url|
+        |/search|
+
+Scenario: Verify the URL when switching the selection from AND to OR
+    When I enter 'ti' at 'Search' field
+    Then I see 'url'
+        |url|
+        |/search|
+    And I click 'Search' button
+    And I see 'url'
+        |url|
+        |/search?query=ti|
+    And I select 'May include any words (ex. transportation OR safety)' at 'search-verb-switch'
+    And I see 'url'
+        |url|
+        |/search?query=ti&andOr=OR|
+    And I select 'Must include all words (ex. transportation AND safety)' at 'search-verb-switch'
+    And I see 'url'
+        |url|
+        |/search?query=ti&andOr=AND|
+
+Scenario: Verify the URL when switching the selection from OR to AND
+    When I select 'May include any words (ex. transportation OR safety)' at 'search-verb-switch'
+    Then I see 'url'
+        |url|
+        |/search|
+    And I enter 'ti' at 'Search' field
+    And I click 'Search' button
+    And I see 'url'
+        |url|
+        |/search?andOr=OR&query=ti|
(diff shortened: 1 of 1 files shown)


eric-serrato95:
@aoysimmons-GH @ychoy 
test scenarios and files have been reviewed ✅

frontend/tests/e2e/search/features/filterDrawer/andor-searchtermswitch.feature
aoysimmons-GH, line 1: Eric Serrato

  [Tuesday at 3:48 PM](https://betagrantsgov.slack.com/archives/C08MTT10SKU/p1768942109335119)
[@Aoy Simmons](https://betagrantsgov.slack.com/team/U09QNQT3ENM) I have reviewd the test scenarios, looks good to me, thanks for sharing!
one note that I have is
When we are testing for the results that come off from searches (for example a given scenario using search and we see 9 results) I think it would be a good idea to try to catch what those 9 results where, if I test in the future for example there is a risk that new opportunities get posted so the count 9 would not match but I still want to make sure the test scenario is functioning properly
any idea here how to ensure this?
(might be too much but just looking to establish continuity for these tests)
also maybe time machine for the browser could solve this? (to re run the query based on the exact date iit was run for tests?)
2 replies


Aoy Simmons

  [Tuesday at 5:29 PM](https://betagrantsgov.slack.com/archives/C08MTT10SKU/p1768948187716129?thread_ts=1768942109.335119&cid=C08MTT10SKU)
[@Eric Serrato](https://betagrantsgov.slack.com/team/U08Q45C3VS7) Thanks for the review and the thoughtful feedback!
That’s a great point about search result counts changing over time. I agree that relying solely on a static count (e.g., 9 results) could make the tests brittle as new opportunities are added.
One idea could be to assert on specific, known attributes of the results instead of (or in addition to) the total count—for example, validating that certain expected items appear in the result set, or that all returned results meet defined criteria (filters, keywords, status, etc.). This would help ensure the search logic is working correctly even if the total number changes.
Another option might be to mock or seed the data used for these tests, if that’s feasible, so the dataset is deterministic. That would give us stronger continuity without depending on live data.
Using a “time machine” approach (running the query against a snapshot as of a specific date) could also work conceptually, but it may add complexity depending on browser and backend support.
I will discuss with test team of these makes the most sense given our current test setup.
CC: [@Bhavna Ramachandran](https://betagrantsgov.slack.com/team/U09T1KPUBKL) [@Doug Schrashun](https://betagrantsgov.slack.com/team/U07NNAXH0F2) (edited) 

Erin Pattison

  [Tuesday at 6:55 PM](https://betagrantsgov.slack.com/archives/C08MTT10SKU/p1768953312734569?thread_ts=1768942109.335119&cid=C08MTT10SKU)
cc [@Josh Osborne](https://betagrantsgov.slack.com/team/U08CDGAMS2H) as we're going to be making some changes to the search algorithm for acronyms, it would be an interesting test case to work through these issues.  Do we want to look for one specific opportunity that we've seeded in local and staging to be a high match to the query, but then not be as strict on other matches?

frontend/tests/e2e/search/features/filterDrawer/andor-searchtermswitch.feature
```diff
@@ -0,0 +1,554 @@
+@7079
```
doug-s-nava, line 1: can we use kebab case for the directory and filenames here `filter-drawer/and-or-search-term-switch`

frontend/tests/e2e/search/features/filterDrawer/andor-searchtermswitch.feature
```diff
@@ -0,0 +1,554 @@
...
+    Then I see each entry in 'Search Results' section contains 'ti'
+    And I select 'Must include all words (ex. transportation AND safety)' at 'search-verb-switch'
+    Then I see each entry in 'Search Results' section contains 'ti'
+
+Scenario: Verify multiple search results when switching from AND to OR
+    When I enter 'Research Scientist' at 'Search' field
+    And I click 'Search' button
+    Then I see each entry in 'Search Results' section contains 'Research' AND 'Scientist'
+    And I select 'May include any words (ex. transportation OR safety)' at 'search-verb-switch'
+    Then I see each entry in 'Search Results' section contains 'Research' OR 'Scientist'
```
doug-s-nava, line 97: I think I've mentioned before this is going to be very difficult to test, as our search algorithm queries fields that are not available in the search results view, such as description. We can try to figure this out longer term, for now when can skip this when it comes to writing scripts

frontend/tests/e2e/search/features/filterDrawer/andor-searchtermswitch.feature
```diff
@@ -0,0 +1,554 @@
...
+    And I select 'May include any words (ex. transportation OR safety)' at 'search-verb-switch'
+    And I see '1000+' at 'Search Results' section
+        | results           |
+        | 1000+             |
+    And I select 'Must include all words (ex. transportation AND safety)' at 'search-verb-switch'
+    And I see '1000+' at 'Search Results' section
+        | results           |
+        | 1000+             |
+
+Scenario Outline: Verify search result when search with arithmetic operators when switching from AND to OR
```
doug-s-nava, line 201: testing this should be a very low priority at this point

frontend/tests/e2e/search/features/filterDrawer/andor-searchtermswitch.feature
```diff
@@ -0,0 +1,554 @@
...
+    And I see field 'validation text'
+        | validation text                                      |
+        | Search terms must be no longer than 100 characters. |
+    And I select 'Must include all words (ex. transportation AND safety)' at 'search-verb-switch'
+    And I see field 'validation text'
+        | validation text                                      |
+        | Search terms must be no longer than 100 characters. |
+
+
+Scenario: Verify search validation with 101 characters (no spaces) when switching from OR to AND
```
doug-s-nava, line 474: I think covering this scenario generally, rather than in relation to the and/or switch is sufficient. Can leave this here for now, but actual testing around this functionality should be kept general

PR #8055, opened 2026-01-16
Windows Front End Documentation Update

dghazvini:
Realized that I originally put `npx run dev` instead of `npx next dev` when I created the Windows documentation. Running `npx run dev` in the front end directory will actually throw an error that the node module cannot be found, `npx next dev` will automatically grab the node package and run it.

--Edit: made a new branch instead of reusing old one to avoid conflict


documentation/frontend/development.md
```diff
@@ -11,7 +11,7 @@ This [Next.js](https://nextjs.org) application can be run natively (or locally)
 For version 0.1.0, please install and use node <= v22.13.0.
 
 * **For Mac** - Run `npm install && npm run local` to install and start the application.
-* **For Windows** - First follow [this guide](https://www.freecodecamp.org/news/node-version-manager-nvm-install-guide/) for installing Node Version Manager (How to Install NVM on Windows). Then in Windows PowerShell in the \simpler-grants-gov\frontend directory, run `npm install` to install the application. Run `npx run dev` afterwards to start the application.
+* **For Windows** - First follow [this guide](https://www.freecodecamp.org/news/node-version-manager-nvm-install-guide/) for installing Node Version Manager (How to Install NVM on Windows). Then in Windows PowerShell in the \simpler-grants-gov\frontend directory, run `npm install` to install the application. Run `npx next dev` afterwards to start the application.
```
doug-s-nava, line 16: Is there a reason `npm run dev` doesn't work on windows? That should be the preference as it gives us more control over any app specific behavior for running a local server. All we have in there now is disabling New Relic, which isn't crucial, but we could add more in the future

dghazvini: It is related to New Relic. When trying to run `npm run dev` on Windows through WSL, this is the displayed error:

```
> frontend@0.1.0 dev
> NEW_RELIC_ENABLED=false next dev

'NEW_RELIC_ENABLED' is not recognized as an internal or external command
```

The script in package.json doesn't seem to run natively on Windows even through WSL. If we want to keep commands consistent I imagine we'd have to update package.json to include a cross-env New Relic package. Or it might also be an issue with how the WSL environment is configured. I agree it's probably not ideal to keep `npx next dev` since from what I can tell that comment bypasses the packages.json scripts completely.

In any case, the `npx run dev` I originally put in the Doc isn't a valid command; typo'd between reading the other 2 commands.

doug-s-nava: Interesting. I vaguely remember this problem from troubleshooting with someone a few months ago. How do you normally set inline environment variables using WSL? Not a blocker for this change, but ideally our NPM scripts would work for everyone, so if we can adjust them so that we don't have separate directions for everyone that'd be best

dghazvini:
Just resolved merge conflict, need reapproval

Issue #7787, opened 2026-01-06
Test SOAP Endpoint and certificate


jakobpederson:
### Summary

We've introduced a new certificate setup on the backend and we want to confirm it's working.

Currently using this curl:
```
curl -v -X POST 'https://soap.dev.simpler.grants.gov/grantsws-agency/services/v2/AgencyWebServicesSoapPort' \
--header 'Content-Type: application/xml' \
--cert ~/Downloads/grants_s2s_soap_certs/bps_keys/bps_grantors.crt \
--key ~/Downloads/grants_s2s_soap_certs/bps_keys/bps_grantors.key \
--data '
<soapenv:Envelope xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/" xmlns:agen="http://apply.grants.gov/services/AgencyWebServices-V2.0" xmlns:gran="http://apply.grants.gov/system/GrantsCommonElements-V1.0">
   <soapenv:Header/>
   <soapenv:Body>
      <agen:GetApplicationZipRequest>
         <gran:GrantsGovTrackingNumber>GRANT10611719</gran:GrantsGovTrackingNumber>
      </agen:GetApplicationZipRequest>
   </soapenv:Body>
</soapenv:Envelope>'
```

### Acceptance criteria

- [ ] Hit the 'https://soap.dev.simpler.grants.gov/grantsws-agency/services/v2/AgencyWebServicesSoapPort' endpoint with the current certificate data and get a successful response

jakobpederson:
Currently hitting this error:
https://onenr.io/0nQx25OZaRV

chrisw-nava:
CC: @ychoy

jakobpederson:
After talking with Sean Thomas it looks like the issue could be the order in the Docker copies permissions but not preserving the permissions. He's got a PR here https://github.com/HHS/simpler-grants-gov/pull/7819 and I'm trying to test locally.

jakobpederson:
The issue seems to be continuing. I hit up Sean so hopefully we'll have a fix soon.

PR #8056, opened 2026-01-16
[Issue #7787] Fix SOAP/Proxy certificate

jakobpederson:
## Summary

Fixes / Work for #7787 

## Changes proposed

Removed the NamedTemporaryFile context manager and created a NamedTemporaryFile that we then write to and then close. This is to prevent an OpenSSL error where it tries to open a NamedTemporaryFile that is locked. 

Also changed the `cert.serial_number` from an `int` to `hex` because the `hex` value matches the serial number on the `serial_num` field on the `tcertificates`. 

## Context for reviewers

Request to the SOAP/Proxy were failing on certificates despite certificates being legitimate. After testing it was because of an OpenSSL error of trying to open a locked file and because the `serial_number` extracted from the certificates was in the wrong format. This PR addresses both of these issues.

## Validation steps

Use this curl:
```
curl -v 'https://soap.staging.simpler.grants.gov/grantsws-agency/services/v2/AgencyWebServicesSoapPort' \
--header 'Content-Type: application/xml' \
--header 'X-GG-S2S-URI: https://trainingws.grants.gov' \
--cert "/Users/$(whoami)/Downloads/grants_s2s_soap_certs/bps_keys/bps_grantors.crt" \
--key "/Users/$(whoami)/Downloads/grants_s2s_soap_certs/bps_keys/bps_grantors.key" \
--data '<soapenv:Envelope xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/" xmlns:agen="http://apply.grants.gov/services/AgencyWebServices-V2.0" xmlns:gran="http://apply.grants.gov/system/GrantsCommonElements-V1.0">
   <soapenv:Header/>
   <soapenv:Body>
      <agen:GetApplicationZipRequest>
         <gran:GrantsGovTrackingNumber>GRANT00837443</gran:GrantsGovTrackingNumber>
      </agen:GetApplicationZipRequest>
   </soapenv:Body>
</soapenv:Envelope>'
```
And confirm you got a successful response back from the proxy in New Relic:
`simpler_soap_api: Successfully processed request and returning SOAP proxy response`


api/src/legacy_soap_api/legacy_soap_api_proxy.py
```diff
@@ -50,39 +53,46 @@ def get_proxy_response(soap_request: SOAPRequest, timeout: int = PROXY_TIMEOUT)
...
+    finally:
+        if temp_file_path and os.path.exists(temp_file_path):
+            os.remove(temp_file_path)
+
+
+def get_cert_file(soap_auth: SOAPAuth, config: LegacySoapAPIConfig) -> _TemporaryFileWrapper:
+    temp_cert_file = NamedTemporaryFile(mode="w", delete=False)
+    cert = soap_auth.certificate.get_pem(config.soap_auth_map)
+    temp_cert_file.write(cert)
+    temp_cert_file.close()
```
chouinar, line 95: What was the OpenSSL error you were seeing with a locked file? This seems like you are doing file operations that were already handled by the context manager, if there's a genuine problem with that, it should be thoroughly explained as seeing file logic without a context manager immediately reads as problematic code.

jakobpederson: The error we were getting on NewRelic
```
["<class 'requests.exceptions.SSLError'>","HTTPSConnectionPool(host='trainingws.grants.gov', port=443): Max retries exceeded with url: /grantsws-applicant/services/v2/ApplicantWebServicesSoapPort (Caused by SSLError(SSLError(524297, '[SSL] PEM lib (_ssl.c:4158)')))","<traceback object at 0x7feafeeb8c80>"]
```

And it turned out `SSLError(524297, '[SSL] PEM lib (_ssl.c:4158)')))` was an error that was indicating that there was an error trying to read the file that the cert data was on. I thought it was possible that this was because we were attempting to read the file in the context manager before it was closed, so we tried to close the file and then try and read it and that stopped the error.

What I think was happening was that while we were in the context manager, it was "locked" by Python, but OpenSSL tried to open that file from the filepath that was passed to it, which had not yet been unlocked and written to, so it failed.

jakobpederson: From what I'm reading OpenSSL is in the C-Library so it doesn't have access to Python internal memory.

jakobpederson: Should I add this as comment?

chouinar: @jakobpederson - Yes, please add that explanation. A good rule of thumb, if you do something intentionally outside the normal pattern, explain why. Otherwise someone coming along might think to "fix" it, and unknowingly break it again.

For example, this first example would be bad because it's commenting and explaining something too obvious. You don't need to explain why you're doing something the routine / by-the-book.

```py
# Opening a file with the context manager
# to ensure it gets closed properly when we are done with it
with open("my_file.txt") as my_file:
     ...
```

While this example's comment is helpful and says why you're not following a standard approach.

```py
# We don't use a context manager for opening this file
# as the xyz library can't read a context manager file
# due to an issue:
# <link to relevant docs / a github issue>
my_file = open("my_file.txt")

xyz.process_file(my_file)

my_file.close()
``

Issue #6945, opened 2025-11-04
Remove deprecated political review/sync status functionality and ExcludedOpportunityReview model


joshtonava:
### Summary

Now that the --sync-status flag has been removed from the load-transform command across all environments, the `SyncOpportunityReviewStatus` task will no longer run. This task was responsible for populating the `excluded_opportunity_review` table with opportunities in review status.

### Acceptance criteria

Since this table will never be populated, all code that filters against it is now obsolete. We need to remove:
-  The sync task implementation
- The `ExcludedOpportunityReview` database model
- All queries that filter against this table (they're now no-ops)
- Test factories and test cases
- Create a migration to drop the table from the database

mdragon:
Please hold on merging this. I was under the impression this was a feature that was coming back. So we wouldn't want to fully remove it if that's the case.

mdragon:
Confirmed that this functionality is not coming back so you may proceed with removing it

PR #8058, opened 2026-01-16
[Issue #6945] Remove sync opportunity review status and references[1/2]

babebe:
## Summary

Fixes / Work for #6945  

## Changes proposed

Now that the --sync-status flag has been removed from the load-transform command across all environments, the SyncOpportunityReviewStatus task will no longer run. This task was responsible for populating the excluded_opportunity_review table with opportunities in review status.

**TODO: **Once PR merged will need to drop the "v_opportunity_summary" table from foreign schema in lower envs, then prod post deploy**

## Validation steps

Run existing tests
Run migration to remove table locally
Run tasks `export-opportunity-data`, `load_agency_data`, `load_opportunity_data` successfully


api/src/search/backend/load_opportunities_to_index.py
```diff
@@ -105,12 +100,6 @@ def fetch_opportunities(self) -> Iterator[Sequence[Opportunity]]:
                 .where(
                     Opportunity.is_draft.is_(False),
                     CurrentOpportunitySummary.opportunity_status.isnot(None),
-                    ~exists(
-                        select(ExcludedOpportunityReview.legacy_opportunity_id).where(
-                            ExcludedOpportunityReview.legacy_opportunity_id
-                            == Opportunity.legacy_opportunity_id
-                        )
-                    ),
```
chouinar, line 113: Something we need to think about for each of these is the order of operations. If we remove the DB table these reference in the same PR as the queries adjusted, it might actually cause the queries to fail. Our deployments do this in the following order:
* Run DB migrations
* Update the running API code (assume this happens like 2-5 minutes later)

So, we'll remove the tables before we change the queries. That means for 2-5 minutes, those queries will try to query against a table that does not exist and fail. Any endpoint will 500 for several minutes/any ECS task will fail.

So, my recommendation is to split this PR into two parts:
* Remove all references to the ExcludedOpportunityReview table - fine to also cleanup the other VOpportunitySummary since it's not used anywhere by running code.
* After the subsequent prod deploy we can merge a change that drops the ExcludedOpportunityReview table as nothing will reference it.

babebe: Updated PR to remove migration file and put back the ExcludedOpportunityReview model. This change is now in this [PR](https://github.com/HHS/simpler-grants-gov/pull/8076/changes)

Issue #8042, opened 2026-01-16
Setup terraform infra iac code for the new test environment.


prasnava:
AC

- [ ] Set up additional environment for Grantee UAT

PR #8131, opened 2026-01-26
[Issue #8042]: new test1 env iac 

prasnava:
## Summary

Created new test1 env. 

## Validation steps

The test1 env infra should successfully be applied.


mdragon:
Can we name this grantee-1 or grantee1 unless you're intentionally creating one first that will prove the point but not actually be turned over to the team for actual usage?

mdragon:
What are our thoughts about domain names and SSL for these?

prasnava:
> What are our thoughts about domain names and SSL for these?

I set the domain name prefix to "*.grantee1.simpler.grants.gov".  Are we ok with this cname, or do we want to change it to something else? 

Once we have the environment up, we request an ssl cert to import into acm.

infra/api/app-config/env-config/scheduled_jobs.tf
```diff
@@ -41,6 +41,17 @@ locals {
...
+      "poetry",
+      "run",
+      "flask",
+      "data-migration",
+      "load-transform",
+      "--load",
+      "--transform",
+      "--set-current",
+      "--store-version"
+    ],
```
chouinar, line 54: Unless you're going through the entire process of setting up DB connectivity to grants.gov, this job is going to perpetually fail.

Have we considered the idea of copying the dev/staging DB for this environment? If so, we could eliminate a lot of these jobs in this environment.

prasnava: yes we are copying the staging db. The test/grantee1 env will not have the vpc peering connection to grants.gov. Will make that block empty.

chouinar: If we're copying the DB, for the jobs, we'd want the following disabled:
* load-transform
* sam-extracts - in lower envs, this just makes each user an org, but staging will already have that
* build-automatic-opportunities - not a big deal to have it enabled, it'll just do nothing as the opportunities it makes will already exist

Maybe want these disabled?
* create-analytics-db-csvs - Will we even have an analytics grantee1 env? If not, disable it.
* email_notification_opportunity - This might just be annoying to have enabled as you'd double up any staging emails
*

PR #8175, opened 2026-01-26
[Issue #7657] Upgrade APIFlask to 3.1

### LINKED ISSUE #7657: Migrate to APIFlask 3.0.x
### Summary

We're current on APIFlask 2.x - APIFlask 3 adds a few new features:
* Support for Pydantic for API routes (in addition to existing Marshmallow support)
* Actual support for MultiAuth (which we've had to handle in a hacky manner up until now)
* Slightly better auth configuration

https://apiflask.com/migration_guide/#migrate-to-apiflask-3x

As part of this upgrade, we'll want to adjust our auth to be defined using the new classes as detailed. We won't make any changes yet for starting to use Pydantic which will require some rewriting of our error handling logic to be usable.

As part of these changes, there should be no changes to our auth behavior, we should not require any changes to our unit tests. If there is any issue that requires modifications to our tests, halt work, we'll need to evaluate how we proceed.

With better support of multi-auth, we should no longer need to define auth like:

```py
@example_blueprint.post("/example")
@example_blueprint.doc(security=jwt_or_key_security_schemes)
@jwt_or_key_auth.login_required
def my_route( ... ) -> ...
```

And should instead be able to do it like:
```py
@example_blueprint.post("/example")
@example_blueprint.auth_required(jwt_or_key_auth)
def my_route( ... ) -> ...
```

No longer needing to define the docs or use the `login_required` attribute instead of `auth_required`

### Acceptance criteria

- [ ] APIFlask upgraded to latest version (3.0.2 as of the writing of this ticket, but may be higher)
- [ ] All auth works the same as before from a user perspective, no tests needed to be modified to support the auth

chouinar:
## Summary
Fixes #7657 

## Changes proposed
* Upgrade to APIFlask 3.0.x
* Cleanup how we setup auth now that the approach can be done a bit simpler with new version of equivalent classes

## Context for reviewers
We can remove quite a bit of boilerplate for our auth - no longer needing security schemes as MultiAuth is fully supported automatically. This does mean any endpoint that uses 2 types of auth required a slight adjustment to the definition, but no functionality should actually change with this.

There were a handful of endpoints using an alternative approach to auth for JWT + User API Key as we built 2 similar things around the same time, I removed the duplicate that was clunkier to use in favor of the simpler common one.

## Validation steps
Relying on the unit tests + e2e tests to verify I didn't fundamentally break auth which is where most changes are. Only changed a few tests that switched auth slightly (just text changing).

The main thing I needed to manually check is that our OpenAPI docs work correctly. The reason we previously had to specify the schemes on every API was otherwise the auth wouldn't connect properly for multi auth. All of the endpoints locally have auth still viewable without issue - the info is properly connected to every endpoint automatically.

Screenshot 2026-03-27 at 4 14 44 PM

### DIFF
--- a/api/src/api/workflows/workflow_routes.py
+++ b/api/src/api/workflows/workflow_routes.py
@@ -73,10 +73,9 @@ def workflow_get(db_session: db.Session, workflow_id: uuid.UUID) -> response.Api
 @workflow_blueprint.doc(
     summary="Get Workflow Audit Events",
     description="Retrieve a paginated list of audit events for a specific workflow. Access is controlled by entity-based privileges (VIEW_OPPORTUNITY or VIEW_APPLICATION).",
-    security=jwt_or_api_user_key_security_schemes,
     responses=[200, 401, 403, 404],
 )
-@jwt_or_api_user_key_multi_auth.login_required
+@workflow_blueprint.auth_required(jwt_or_api_user_key_multi_auth)
 @flask_db.with_db_session()
 def workflow_audit_post(
     db_session: db.Session, workflow_id: uuid.UUID, json_data: dict
--- a/api/tests/src/db/models/factories.py
+++ b/api/tests/src/db/models/factories.py
@@ -1018,7 +1018,7 @@ class AwardRecommendationAttachmentFactory(BaseFactory):
                 Does this location exist? If you are running in unit tests, make sure
                 `enable_factory_create` is pulled in as a fixture to your test.
 
-                If you are running locally outside of unit tests, make sure that `make init-localstack` has run.
+                If you are running locally outside of unit tests, make sure that `make init-s3mock` has run.
                 """
             ) from e
 


api/src/api/agencies_v1/agency_routes.py
```diff
@@ -59,8 +59,7 @@
     examples=examples,
 )
 @agency_blueprint.output(agency_schema.AgencySearchResponseV1Schema)
-@agency_blueprint.doc(security=jwt_or_user_api_key_security_schemes)
-@jwt_or_user_api_key_multi_auth.login_required
+@agency_blueprint.auth_required(api_key_multi_auth)
 @flask_opensearch.with_search_client()
```
kkrug, line 63: I know there were a bunch of auth changes that didn't functionally change how anything works, but this one removes JWT altogether. I just want to confirm that this is intentional, and recommend noting the change in the ticket/PR notes.

api/tests/src/api/agencies_v1/test_agency_auth.py
```diff
@@ -57,7 +57,7 @@ def test_agency_search_unauthorized_401_jwt(client, method, url, body):
     )
 
     assert response.status_code == 401
-    assert response.get_json()["message"] == "Unable to process token"
+    assert response.get_json()["message"] == "Invalid API key"
```
kkrug: With the change in auth type noted in my other comment, is this test still necessary? I think the 3 other 401 tests below this cover the realistic scenarios.

api/src/api/agencies_v1/agency_routes.py
chouinar: No, it was not - good catch. Think the merge conflicts I resolved the wrong way around as we changed the auth on some of these endpoints between when I first made this and now.

api/tests/src/api/agencies_v1/test_agency_auth.py
chouinar: The change in auth wasn't intentional, so I shouldn't have changed this.

api/src/api/award_recommendations_alpha/award_recommendation_routes.py
```diff
@@ -24,7 +24,6 @@
 @award_recommendation_blueprint.doc(
     summary="Get Award Recommendation",
     description="Retrieve an award recommendation by ID, including opportunity details, attachments, and reviews.",
-    security=jwt_or_api_user_key_security_schemes,
     responses=[200, 401, 403, 404],
 )
 @jwt_or_api_user_key_multi_auth.login_required
```
kkrug:
```suggestion
@award_recommendation_blueprint.auth_required(jwt_or_api_user_key_multi_auth)
```

Issue #7349, opened 2025-12-05
Implement design to add a link to a download a submitted application


chris-kuryak:
### Summary

Based on design #7348 
Figma: https://www.figma.com/design/zPlDWcYM396szw2yL3UXvk/Simpler-Grants-2025?node-id=8920-37425&m=dev

We want to show a button that allows users to download the Zip file of their application after they have submitted it.

While the underlying API is a list, we only need the most recent submission

### Acceptance criteria

- [ ] Show the button only when the application is in `accepted` state
   - [ ] In the `submitted` state the application has been submitted but the Zip isn't ready.  Show nothing.
- [ ]  Clicking on the link calls the submissions API (POST /alpha/applications/<uuid:application_id>/submissions)
   - [ ] API pagination should be sorted to most recent, and only 1 record should be requested to get the most recent
   - [ ] If the API returns 200, download the the document to the user's device
   - [ ] If the API returns anything else, show the specified generic error locally next to the button
- [ ] Review with Design before PR is submitted
- [ ] Unittests

chris-kuryak:
UAT passed, but could not check error state.

UAT performed on: https://staging.simpler.grants.gov/applications/8ba4ff7b-d86b-4c4e-8ecc-d0339544efcb

<img width="1057" height="202" alt="Image" src="https://github.com/user-attachments/assets/6b5286fa-a837-4142-bb00-c1e670205dd6" />
<img width="1051" height="217" alt="Image" src="https://github.com/user-attachments/assets/72bcd1d3-973d-43e8-92cf-5141e3e79762" />

PR #8177, opened 2026-01-26
[Issue #7349] Add support for downloading latest application submission zip

freddieyebra:
## Summary

Work for #7349 

## Changes proposed

This pr adds support for download content for an application in different states (in progress, submitted, accepted).

## Context for reviewers

This feature utilizes the get submissions list response from the api to get the latest application submission.

## Validation steps

1. Ensure db has been seeded.
2. Run the api and frontend
3. Login to user with applications with applications in the 3 status (submitted, in progress, accepted)
4. Navigate to the application list page
5. Navigate to an application with the Accepted status and verify the following UI:
Accepted application
Screenshot 2026-01-27 at 11 24 48 AM

Mobile accepted
Screenshot 2026-01-27 at 11 26 28 AM

6. Click the Download submission zip button and ensure the download zip is downloaded and available.
7. Navigate to an application in each of the following statuses and verify the following ui:

In progress application
in-progress

Submitted application
submitted
8. To verify an application that is in IN_PROGRESS status to > ACCEPTED locally:
- Set PDF_GENERATION_USE_MOCKS=true in local.env
- Submit <x type of form> form with application
- Run `make cmd args="task create-application-submission"`
- Reload application and download submission zip


jesababon:
On the frontend - we don’t want to say “accepted” as that would be confusing as a status for users. We can keep the copy “submitted” as the status copy for “accepted.”
image

jesababon:
Thanks for the update, Freddie! And yes, we discussed at the onsite about reporting the error occurred on the backend and defaulting to the "preparing download" message while investigating the error.
LGTM! Passes design review.

frontend/src/app/[locale]/(base)/applications/[applicationId]/page.tsx
```diff
@@ -112,12 +125,52 @@ async function ApplicationLandingPage({ params }: ApplicationLandingPageProps) {
...
           attachments={attachments}
           applicationHistory={historyDetails}
+          latestApplicationSubmission={latestApplicationSubmission}
         />
       </GridContainer>
     </>
   );
 }
 
+const LATEST_APPLICATION_SUBMISSION_REQUEST: ApplicationSubmissionsRequestBody =
```
Comment (ErinPattisonNava) [inline frontend/src/app/[locale]/(base)/applications/[applicationId]/page.tsx]: It feels a little weird to have the RequestBody object exposed into the component, vs having a function in the fetcher similar to latestApplicationSubmissionRequest.  That would let the details of how we call the API be encapsulated more cleanly.

frontend/src/components/application/InformationCard.tsx
```diff
@@ -109,6 +112,45 @@ export const InformationCard = ({
...
+   *
+   * IN_PROGRESS
+   *  - Has not been submitted, so do not render anything related to submission.
+   * SUBMITTED
+   *  - Submitted but does not yet have submission db entry and S3 download not available yet.
+   *  - Show message that download is being prepared.
+   * ACCEPTED
+   *  - Download available, render button to download submission zip.
+   */
+  const ApplicationSubmissionDownload = () => {
```
ErinPattisonNava, line 127: Worth pulling out as its own component?

frontend/src/app/[locale]/(base)/applications/[applicationId]/page.tsx
```diff
@@ -112,12 +125,52 @@ async function ApplicationLandingPage({ params }: ApplicationLandingPageProps) {
...
+  {
+    pagination: {
+      page_offset: 1,
+      page_size: 1,
+      sort_order: [{ order_by: "created_at", sort_direction: "descending" }],
+    },
+  };
+
+async function getLatestSubmission(
```
Comment (ErinPattisonNava) [inline frontend/src/app/[locale]/(base)/applications/[applicationId]/page.tsx]: It seems odd that we need to do this top level rather than down in the component that needs the data. Is that because of weird server/client component interactions, or could we encapsulate the data call within the ApplicationSubmissionDownload component?

ErinPattisonNava (commented):
When verifying locally, I noticed two things:

a) When I download the zip file the pdf of the form is corrupted.  I'm doing this on localhost so it shouldn't have the API Gateway issue with attachments.  Were you able to correctly view the pdfs of the forms?  If not, this feels like a backend issue so can you file a bug and post in the channel so we can discuss?

b) If the API returns an error nothing is displayed -- is this expected?  I'm ok if design wants to swallow the error or if we want to display the "We are preparing your submitted application for download. Please check back later.", just wanted to check what was intended.

c) Also checking that this is what we wanted to see in tablet breakpoints?  I would have expected different ordering of the buttons with the submit on top (and them centered rather than left aligned), but defer to design. 

Image

freddieyebra:
@ErinPattisonNava 

> a) When I download the zip file the pdf of the form is corrupted. I'm doing this on localhost so it shouldn't have the API Gateway issue with attachments. Were you able to correctly view the pdfs of the forms? If not, this feels like a backend issue so can you file a bug and post in the channel so we can discuss?

Looking into this now good catch.

> b) If the API returns an error nothing is displayed -- is this expected? I'm ok if design wants to swallow the error or if we want to display the "We are preparing your submitted application for download. Please check back later.", just wanted to check what was intended.

Yes it is intended. I have created a [follow up ticket](https://github.com/HHS/simpler-grants-gov/issues/8282) to setup alerting for any failures around this. 

Additionally, i want to discuss with the team around having a dedicated endpoint for getting the latest submission since the response from the submission list endpoint is not necessarily failing if multiple or 0 submissions are returned. Having a dedicated endpoint for the latest submission should minimize edge cases the FE needs to handle.

> c) Also checking that this is what we wanted to see in tablet breakpoints? I would have expected different ordering of the buttons with the submit on top (and them centered rather than left aligned), but defer to design.

The submit button would not be visible if the application is in the `SUBMITTED` or `ACCEPTED` state which are the only states this functionality would render the download button or zip in progress message.

frontend/src/app/[locale]/(base)/applications/[applicationId]/page.tsx
Comment (freddieyebra) [reply frontend/src/app/[locale]/(base)/applications/[applicationId]/page.tsx]: yes it is because of weird server/client component.

frontend/src/app/[locale]/(base)/applications/[applicationId]/page.tsx
Comment (freddieyebra) [reply frontend/src/app/[locale]/(base)/applications/[applicationId]/page.tsx]: updated to address this.

freddieyebra:
@ErinPattisonNava i was able to download an app submission zip with the following contents 
Screenshot 2026-01-30 at 3 57 40 PM

i wonder if corruption of zip was related to the specific application or form?

Issue #7768, opened 2026-01-05
Endpoint to deliver test user auth token for playwright in lower envs


doug-s-nava:
### Summary

See https://github.com/HHS/simpler-grants-gov/issues/7256 for background

In order to allow for circumventing the full login.gov authorization process in staging for purposes of efficient end to end testing, we want to be able to fetch a viable auth token that Playwright can use to spoof cookies on the frontend.

Related frontend ticket https://github.com/HHS/simpler-grants-gov/issues/7769

### Acceptance criteria

- [ ] new user privilege read_test_user_token is created
- [ ] POST endpoint exists at /v1/internal/e2e-token
- [ ] endpoint can be accessed only by API key auth, using a key with the `read_test_user_token` privilege attached
- [ ] endpoint creates a necessary DB changes to reflect creation of a session for test user account
- [ ] endpoint returns a viable auth token for a test user account
- [ ] endpoint is NOT available in production / endpoint returns an instant 404 in production

kkrug:
@doug-s-nava @chouinar I read through the linked background ticket - do we want to create a new role for this (e.g. E2E_TEST_USER_ROLE) or will this new privilege be added to an existing role for test users?

doug-s-nava:
Yeah, this should be a new role

PR #8186, opened 2026-01-26
Endpoint to create e2e test user auth token

kkrug:
## Summary

Fixes / Work for #7768  

## Changes proposed

Created a new internal endpoint for the creation of a test user auth token in lower environments. A new role (E2E_TES_USER_ROLE) was created with the new privilege (READ_TEST_USER_TOKEN). This endpoint should only be available outside of production and to a user with this role and privilege. 

**NOTE**
A small change was made to the other internal endpoint to move the verify_access step from the route into the service layer for consistency with our coding patterns. No functionality was affected. 

## Context for reviewers

In order to allow for circumventing the full login.gov authorization process in staging for purposes of efficient end to end testing, we want to be able to fetch a viable auth token that Playwright can use to spoof cookies on the frontend.

## Validation steps

Confirm the endpoint returns the auth key only when provided the API key 'e2e-test-key'. The endpoint will also return a 404 if called from production, a 403 if the user does not have the correct privilege, and a 401 if the API key is missing/invalid.

### DIFF
--- a/api/tests/src/api/internal/test_internal_e2e_token.py
+++ b/api/tests/src/api/internal/test_internal_e2e_token.py
@@ -53,6 +53,7 @@ def test_get_e2e_token_unauthorized(client, enable_factory_create):
 
 def test_get_e2e_token_production_guard_403(client, db_session, enable_factory_create, monkeypatch):
     """Test that the endpoint returns a 403 in production, even with valid auth"""
+    monkeypatch.setenv("ENVIRONMENT", "production")
     api_key = "prod-test-key"
     setup_e2e_privileges(db_session, api_key, has_privilege=True)
 


kkrug:
@doug-s-nava I added you as a reviewer here in case you wanted to take a look as it is relevant to the Playwright testing work.

api/src/api/internal/internal_routes.py
```diff
@@ -30,10 +29,27 @@ def update_internal_roles(
...
+    Endpoint to fetch a viable auth token for a test user account.
+    Only accessible via API key auth with the read_test_user_token privilege.
+    """
+    logger.info("POST /v1/internal/e2e-token")
+
+    user = api_user_key_auth.get_user()  #
+    user = db_session.merge(user, load=False)
+
+    token_data = create_e2e_token(db_session, user)
+
```
chouinar:
```suggestion
    with db_session.begin():
        user = api_user_key_auth.get_user()
        user = db_session.add(user)

        token_data = create_e2e_token(db_session, user)

```

api/src/constants/static_role_values.py
```diff
@@ -136,11 +136,28 @@ def get_link_privileges(role_id: uuid.UUID, privilege: list[Privilege]) -> list[
...
+E2E_TEST_USER_ROLE = Role(
+    role_id=E2E_TEST_USER_ROLE_ID,
+    role_name="E2E Test User",
+    is_core=True,
+    link_privileges=get_link_privileges(
+        E2E_TEST_USER_ROLE_ID,
+        [
+            Privilege.READ_TEST_USER_TOKEN,
+            Privilege.LIST_APPLICATION,
+            Privilege.VIEW_APPLICATION,
```
chouinar: Internal roles are special, they aren't checked against entities (eg, an application). If you give a user an internal role with these privileges, they'll be able to read any application, even those that they didn't create. That's not what we want here I don't think. A user can have multiple roles in different things, they can get the read app privileges via their org.

```suggestion
```

api/src/services/internal/create_e2e_token.py
```diff
@@ -0,0 +1,25 @@
...
+from src.constants.lookup_constants import Privilege
+from src.db.models.user_models import User
+
+logger = logging.getLogger(__name__)
+
+
+def create_e2e_token(db_session: db.Session, user: User) -> dict:
+    if os.getenv("APP_ENV") == "production":
+        logger.warning("Attempted to access E2E token endpoint in production")
+        raise_flask_error(404)
```
chouinar: Our environment is defined in `ENVIRONMENT`. It's also safer to make this only allow specific environments (we call it prod, not production).

Instead, do something like:

```suggestion
    if os.getenv("ENVIRONMENT") not in ["local", "dev", "staging", "training"]:
        logger.warning("Attempted to access E2E token endpoint in production")
        raise_flask_error(403, "Forbidden")
```

We've done this enough times, it might be a good idea to make a utility for doing this check so if we add more envs, it's just one place to adjust.

api/src/services/internal/create_e2e_token.py
```diff
@@ -0,0 +1,25 @@
...
+
+    verify_access(user, {Privilege.READ_TEST_USER_TOKEN}, None)
+
+    logger.info("Generating E2E bypass token", extra={"auth.user_id": str(user.user_id)})
```
chouinar: When we do auth, we add an `auth.user_id` value already with the user ID, so can leave it out.

```suggestion
    logger.info("Generating E2E bypass token")
```

api/src/api/internal/internal_routes.py
```diff
@@ -30,10 +29,27 @@ def update_internal_roles(
...
-        update_internal_user_role(db_session, internal_role_id, user_email)
+        update_internal_user_role(db_session, internal_role_id, user_email, user)
 
     return response.ApiResponse(message="Success")
+
+
+@internal_blueprint.post("/e2e-token")
+@internal_blueprint.output(internal_schema.E2ETokenResponseSchema)
+@internal_blueprint.auth_required(api_user_key_auth)
+@flask_db.with_db_session()
```
chouinar, line 43: We might want to do this - this'll make it so it doesn't appear in the OpenAPI/swagger docs. @doug-s-nava - any opposition to hiding this endpoint?

```suggestion
@internal_blueprint.auth_required(api_user_key_auth)
@internal_blueprint.doc(hide=True)
@flask_db.with_db_session()
```

api/tests/lib/seed_e2e.py
```diff
@@ -23,8 +25,19 @@ def _write_token_to_file(token: str) -> None:
 
 
 def _build_users_and_tokens(db_session: db.Session) -> None:
+    e2e_role = factories.RoleFactory.create(
+        role_name="E2E Test Role", is_core=True, privileges=[Privilege.READ_TEST_USER_TOKEN]
+    )
+    factories.LinkRoleRoleTypeFactory.create(role=e2e_role, role_type=RoleType.INTERNAL)
```
chouinar: I'd suggest adding a function to the UserBuilder, something like `with_internal_role` - might be useful elsewhere later and saves a bit of boilerplate.

api/tests/lib/seed_e2e.py
```diff
@@ -23,8 +25,19 @@ def _write_token_to_file(token: str) -> None:
...
+
     builder = UserBuilder(
         uuid.UUID("7edb5704-9d3b-4099-9e10-fbb9f2729aff"), db_session, "user for e2e"
     ).with_jwt_auth()
-    builder.build()
+
+    builder.with_api_key("e2e-test-key")
```
chouinar: Add it to the builder calls
```suggestion
     .with_api_key("e2e-test-key")
```

api/tests/src/api/internal/test_internal_e2e_token.py
```diff
@@ -0,0 +1,65 @@
...
+    user = factories.UserFactory.create()
+    factories.UserApiKeyFactory.create(user=user, key_id=api_key_value)
+
+    privileges = [Privilege.READ_TEST_USER_TOKEN] if has_privilege else [Privilege.VIEW_APPLICATION]
+
+    role = factories.RoleFactory.create(role_name="E2E Role", privileges=privileges)
+    factories.LinkRoleRoleTypeFactory.create(role=role, role_type=RoleType.INTERNAL)
+    factories.InternalUserRoleFactory.create(user=user, role=role)
+
+    db_session.commit()
```
chouinar: Factories commit at the end of create - should rarely need commits in our tests
```suggestion

```

api/src/services/internal/create_e2e_token.py
kkrug: Do we want to change to a 403 here? The last AC on the ticket called for a 404: 
> endpoint is NOT available in production / endpoint returns an instant 404 in production

I don't have a particularly strong opinion either way.

api/src/api/internal/internal_routes.py
kkrug: Do we want to hide the other internal endpoint (and any future internal endpoints) as well? If so, I can add a comment in the route file to indicate that they should be left out moving forward.

api/src/services/internal/create_e2e_token.py
chouinar: At the end of the day, this is an internal endpoint, so it doesn't matter a ton, but I'd say 403 makes more sense? If we wanted it to be a 404, we'd want to figure out a way to make the endpoint just not exist in prod, but we don't have a pattern for that (something like https://flask.palletsprojects.com/en/stable/blueprints/#nesting-blueprints maybe?).

doug-s-nava: 403 is fine by me

api/src/api/internal/internal_routes.py
doug-s-nava: no objection, hiding from the docs makes sense to me

api/src/api/internal/internal_routes.py
kkrug: Fixed

api/src/constants/static_role_values.py
kkrug: Got it, fixed

api/src/services/internal/create_e2e_token.py
kkrug: Fixed and changed to 403 (including tests). Also fixed any other references to "APP_ENV" to be "ENVIRONMENT."

api/src/services/internal/create_e2e_token.py
kkrug: Fixed

api/src/api/internal/internal_routes.py
kkrug: Hid from docs, and made the same change to the other existing internal endpoint

api/tests/lib/seed_e2e.py
kkrug: New function created for the UserBuilder

api/tests/lib/seed_e2e.py
kkrug: Fixed

api/tests/src/api/internal/test_internal_e2e_token.py
kkrug: Fixed

api/src/api/internal/internal_routes.py
```diff
@@ -17,6 +16,7 @@
 @internal_blueprint.put("/roles")
 @internal_blueprint.input(internal_schema.InternalRoleAssignmentRequestSchema, location="json")
 @internal_blueprint.output(internal_schema.InternalRoleAssignmentResponseSchema)
+@internal_blueprint.doc(hide=True)
```
chouinar: We probably shouldn't hide this one as we expect some of our engineers to use it directly and the openapi docs are the most convenient way. Otherwise they'd need to curl or setup something like postman.

Fine with hiding the e2e one since it'll never be used by a person directly.

api/src/api/internal/internal_routes.py
```diff
@@ -30,10 +30,29 @@ def update_internal_roles(
...
+def get_e2e_token(db_session: db.Session) -> response.ApiResponse:
+    """
+    Endpoint to fetch a viable auth token for a test user account.
+    Only accessible via API key auth with the read_test_user_token privilege.
+    """
+    logger.info("POST /v1/internal/e2e-token")
+
+    with db_session.begin():
+        user = api_user_key_auth.get_user()
+        user = db_session.merge(user, load=False)
```
chouinar:
```suggestion
        user = api_user_key_auth.get_user()
        user = db_session.add(user)
```

api/tests/src/api/internal/test_internal_e2e_token.py
```diff
@@ -0,0 +1,64 @@
...
+    role = factories.RoleFactory.create(role_name="E2E Role", privileges=privileges)
+    factories.LinkRoleRoleTypeFactory.create(role=role, role_type=RoleType.INTERNAL)
+    factories.InternalUserRoleFactory.create(user=user, role=role)
+
+    return user
+
+
+def test_get_e2e_token_success(client, db_session, enable_factory_create, monkeypatch):
+    """Test successful token retrieval in a lower environment"""
+    monkeypatch.setenv("ENVIRONMENT", "local")
```
chouinar: In tests, the env should already be local since they use our local.env which sets it that way.
```suggestion

```

kkrug: Removed all instances of setting env except setting to prod for prod guard check

api/src/api/internal/internal_routes.py
kkrug: Fixed

api/src/api/internal/internal_routes.py
kkrug: Unhid from docs

chouinar (approved):

PR #8187, opened 2026-01-26
[Issue #8067] Create a minimal workflow main function

### LINKED ISSUE #8067: Create a barebones workflow management service script
### Summary

Should have the logic built into it to handle receiving a SIGTERM. Implementation can be a loop that logs a message and sleeps for 10 seconds repeatedly - this is just to give a basic service setup that the infra can point to, it does not need to do anything for workflows yet.

### Acceptance criteria

- [ ] Script runs and logs a message every 10 seconds
- [ ] Script can be interrupted with both a SIGTERM and keyboard interrupt

chouinar:
## Summary
Fixes #8067

## Changes proposed
Creates a minimal entrypoint/main function for the workflow management service

## Context for reviewers
This is just an event loop, it doesn't actually process SQS events, just randomly makes events that it then logs out. Has the rough pattern of how we'll do this, without any meaningful implementation beyond the signal handling. Note that it's likely the logs won't work with New Relic yet, a follow-up ticket will address that.

We want something to build on, and something for infra folks to have that they can get stood up. This doesn't really do anything yet (it likely doesn't work with New Relic as we're missing some config normally handled by `@ecs_background_task`, separate ticket will handle that), but it's something that roughly works like the main function, without any core functionality needed yet. Let's us and infra work be built separately if we merge this.

## Validation steps
Locally you can run this with `make cmd args="workflow workflow-main"`. A keyboard interrupt (CTRL+C or CMD+C) will immediately interrupt it, even if it's in the middle of a sleep. If you want to test SIGINT, you can do `ps aux | grep workflow-main` to find the process ID, and then do `kill <process ID>` (running inside Docker, not 100% if this'll work).

### DIFF
--- a/api/src/workflow/manager/workflow_manager.py
+++ b/api/src/workflow/manager/workflow_manager.py
@@ -56,6 +56,10 @@ class WorkflowManager:
     def _register_signal_handlers(self) -> None:
         """Register signal handlers to handle expected
         signals like keyboard interrupts and kill commands.
+
+        This changes the default behavior of "end the program instantly"
+        into a more graceful approach. Note that not all signals
+        can be caught, and many that indicate hardware faults shouldn't.
         """
         # Make it so if a SIGTERM is received, it doesn't
         # cause the process to instantly exit so we can gracefully
@@ -74,7 +78,7 @@ class WorkflowManager:
 
         # Most other signals indicate either errors or
         # more significant kill signals that we are fine
-        # just killing the program outright.
+        # with causing the program to exit instantly as normal.
 
     def handle_exit(self, signum: int, frame: FrameType | None) -> None:
         logger.info(
@@ -101,7 +105,10 @@ class WorkflowManager:
         return events
 
     def process_events(self) -> None:
-        """Process events constantly,"""
+        """Process workflow events constantly
+
+        The 'main' loop of the workflow manager.
+        """
         logger.info("Processing workflow events")
 
         batch_count = 0
@@ -124,6 +131,8 @@ class WorkflowManager:
                 logger.info("Exiting after receiving SIGTERM.")
                 break
 
+            # For the purposes of testing, we can configure a maximum batch
+            # size count to break the loop after a certain number of iterations
             if (
                 self.config.workflow_maximum_batch_count is not None
                 and batch_count >= self.config.workflow_maximum_batch_count
--- a/infra/api/app-config/dev.tf
+++ b/infra/api/app-config/dev.tf
@@ -46,7 +46,7 @@ module "dev_config" {
     # Login.gov OAuth
     ENABLE_XML_GENERATION = 1
 
-
+    ENABLE_WORKFLOW_ENDPOINTS = 1
 
     # Email notification
     RESET_EMAILS_WITHOUT_SENDING = "false"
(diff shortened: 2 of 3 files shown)


api/src/workflow/manager/workflow_manager.py
```diff
@@ -0,0 +1,187 @@
...
+
+class WorkflowManager:
+
+    def __init__(self, config: WorkflowManagerConfig | None = None):
+        self.sigterm_received = False
+        self._register_signal_handlers()
+
+        if config is None:
+            config = WorkflowManagerConfig()
+        self.config = config
```
jakobpederson, line 46: (optional) We could also do
`self.config = config or WorkflowManagerConfig()` but I don't think it's necessarily much better. I do like how explicit `is None` is.

api/src/workflow/manager/workflow_manager.py
```diff
@@ -0,0 +1,187 @@
...
+                logger.warning("Encountered retryable workflow error", exc_info=True)
+            except NonRetryableWorkflowError:
+                logger.warning("Encountered non-retryable workflow error", exc_info=True)
+                # In a future ticket, we'll write back the event to SQS
+            except Exception:
+                logger.exception("Failed to process event")
+
+        # Very simple metrics for test purposes
+        self.metrics["batches_processed"] += 1
+        self.metrics["events_processed"] += len(events)
```
jakobpederson, line 173: Are events processed if they throw an exception? Because I kind of assumed that those wouldn't count. But I get this is also super basic metrics for testing.

chouinar: When we set this up properly, we'll probably have "events processed" and "events processed successfully" in some manner. That way we have an event count and a success count. But yeah, this is intentionally really simple, just wanted a hacky thing to validate against for the test.

api/src/workflow/manager/workflow_manager.py
chouinar: I've put it this way in most places, generally prefer slightly more verbose to make it clear.

api/src/workflow/manager/workflow_manager.py
```diff
@@ -0,0 +1,187 @@
...
+                break
+
+            # We want batches to run every 10 seconds at most
+            # so sleep for the remaining time
+            time_to_sleep = round(self.config.workflow_cycle_duration - batch_duration, 3)
+            if time_to_sleep <= 0:
+                time_to_sleep = 0
+            logger.info("Sleeping after processing events", extra={"time_to_sleep": time_to_sleep})
+            time.sleep(time_to_sleep)
+            logger.info("Finished sleeping.")
```
hao10282025-sudo, line 150: Overall looks good to me!
As I see the lines 148 - 150 are printing same "time_to_sleep": 0, when the batch running time >= 10s
may print the actual batch time for these cases to be more helpful?

chouinar: I'll rework this once we connect it to meaningful code. I don't think we'll even keep the sleep in the future as that's mimicking the SQS client polling and waiting for messages. This is all just to avoid ten thousand log messages per second (from my local testing, this time_to_sleep is ~9.95).

Issue #8197, opened 2026-01-27
Ensure jobs run as the app user, unless overridden


mdragon:
### Summary

As part of PR #7481 work for #2500 we attempted to set the jobs to run as the migrator role when updating OpenSearch (because that ticket implemented fine grained permissions/access for OpenSearch). However instead all the jobs are running as the migrator role. Instead, the default task definition should run as the app role. The Migration job during deploys should continue to run as the migrator role (that should require no change in this ticket). The jobs that run to put data in OpenSearch will need to be modified to pass a new OpenSearch write role that will take over the permissions originally granted to the migrator role to write data to OpenSearch.

### Acceptance criteria

- [ ] Migrator Task Definition is removed
- [ ] By default Jobs execute using the single Task Definition with no role change (default app role)
- [ ] For Migration jobs that run during CI/CD, the single Task Definition, but the migrator role is used instead. (This should be the existing behavior, and the OpenSearch permissions should be removed from this role)
- [ ] OpenSearch sync jobs run using the single Task Definition, but use a new opensearch-write role instead. (This role has the OpenSearch permissions that are currently on the migrator role).

PR #8204, opened 2026-01-28
[Issue #8197]: Remove migrator ecs task definition and pass in task role arn overrides 

prasnava:
## Summary

Removed the migrator task definition and added a new role for opensearch writes: open search write role. Previously all jobs where using the migrator role, now we have defaulted for the app role instead and using the new opensearch role for just opensearch writes. 

## Validation steps

The step functions and the api should continue to be functional using the new ecs tasks overrides for the iam role.


prasnava:
Successfully deployed to dev: [here](https://github.com/HHS/simpler-grants-gov/actions/runs/21484171451)

The step function: api-dev-load-search-agency-data is working: [here](https://us-east-1.console.aws.amazon.com/states/home?region=us-east-1#/v2/executions/details/arn:aws:states:us-east-1:315341936575:execution:api-dev-load-search-agency-data:bb1542d6-3d70-4392-a03d-43bd87ba0ff4)

Will wait for all the step functions to successfully run.

After deploying the API we have to manually deploy the database module for the opensearch iam changes to take affect.

infra/modules/service/scheduled_jobs.tf
```diff
@@ -52,6 +68,7 @@ resource "aws_sfn_state_machine" "scheduled_jobs" {
           "Overrides" : {
             "Cpu" : tostring(each.value.cpu),
             "Memory" : tostring(each.value.mem),
+            "TaskRoleArn" : local.scheduled_job_role_arns[each.key],
```
chouinar, line 71: Not sure where to thread, but want to check. The migrator role isn't configured on anything as it is used mainly by setting it directly for DB migrations https://github.com/HHS/simpler-grants-gov/blob/main/bin/run-database-migrations#L65

The other place we would want it is the create foreign table script, which we run manually by calling the `run-command` script. For this, we'd need the migrator arn, but what is the easiest way to get that? We'll want to document that so we can still run this script.

infra/api/app-config/env-config/scheduled_jobs.tf
```diff
@@ -132,6 +133,7 @@ locals {
       cpu                 = try(local.scheduled_jobs_config[var.environment].cpu, null)
       mem                 = try(local.scheduled_jobs_config[var.environment].mem, null)
       environment_vars    = try(local.scheduled_jobs_config[var.environment].environment_vars, null)
+      role_override       = "opensearch-write"
```
sean-navapbc, line 136: Silent fallback in scheduled_job_role_arns                                      
                                                                                                                            
  If a job specifies role_override = "opensearch-write" but opensearch_ingest_policy_arn is null (so                        
  aws_iam_role.opensearch_write doesn't exist), the ternary silently falls through to aws_iam_role.app_service.arn. This could mask misconfiguration — a job intended to have OpenSearch write permissions would run with the app role without any error or warning.

could be solved with adding validation to

role_override values — variables.tf:292                                                               
                                                                                                                            
  role_override is typed as optional(string, null) with no validation. If someone types "opensearch_write" (underscore)     
  instead of "opensearch-write" (hyphen), it silently falls back to the app role. Adding a validation block would catch     
  this:                                                                                                                     
```                                                                                                                            
  role_override = optional(string, null)                                                                                    
  # Would benefit from:                                                                                                     
  # validation {                                                                                                            
  #   condition     = var.role_override == null || contains(["opensearch-write", "migrator"], var.role_override)            
  #   error_message = "role_override must be null, \"opensearch-write\", or \"migrator\""                                   
  # }      
```

infra/modules/service/variables.tf
```diff
@@ -291,6 +291,13 @@ variable "scheduled_jobs" {
     # If not specified or null, defaults to the app role.
     role_override = optional(string, null)
   }))
+  validation {
+    condition = alltrue([
```
prasnava, line 295: @sean-navapbc added the validation here.

infra/modules/service/scheduled_jobs.tf
prasnava: updated the documentation and created a new make command: bin/run-setup-foreign-tables

chouinar (commented):
The foreign table changes LGTM, I'll defer to @sean-navapbc on the infra portion as I'm less familiar with that side

PR #8206, opened 2026-01-28
[Issue #8023] Add support for hard-coded opportunity IDs when building automatic opportunities for lower environments

### LINKED ISSUE #8023: Add support for hard-coded opportunity IDs when building automatic opportunities for lower environments
### Summary

Currently, the `build_automatic_opportunities.py` script generates random UUIDs for `opportunity_id` each time it creates opportunities, making it impossible to reference specific test opportunities consistently across lower environments. To support testing and E2E's that need stable opportunity IDs, we should allow for explicit hard-coded UUIDs for any opportunity scenario and clean up existing randomly-generated records in lower environments that would conflict.

### Tech Notes

- Add configurable field `opportunity_id: uuid.UUID = field(default_factory=uuid.uuid4) ` to `OpportunityContainer` class
- Update opportunity creation logic to use the provided ID instead of `uuid.uuid4()`
- Add explicit UUIDs to all opportunity scenarios in `_create_opportunity_scenarios()` and the dynamic form-based opportunities
- Discussion: Cleanup strategy for existing records:
  - ~~Add a --force-recreate flag that deletes existing opportunities by number and recreates with new IDs~~
  - Update the task to detect ID mismatches and update/replace records
  - Manually go into AWS log into the DB for each environment and clean up the records
- How to run: 
  - `make cmd args="task build-automatic-opportunities"` (local) 
  - `bin/run-command api staging '["poetry", "run", "flask", "task", "build-automatic-opportunities"]'`

### Acceptance criteria

- [ ] `OpportunityContainer` accepts optional `opportunity_id` parameter
- [ ] Opportunity creation uses provided ID when specified, generates random UUID otherwise
- [ ] A set of predefiined existing opportunity scenarios have explicit, hard-coded UUIDs defined
- [ ] Existing lower-environment records with conflicting random IDs are cleaned up (across dev, staging, training)
- [ ] Script creates opportunities with consistent IDs, when id is specified, across all runs
- [ ] Tests verify that re-running the script maintains the same IDs for opportunities with ID specified

mikehgrantsgov:
## Summary

Fixes #8023 

## Changes proposed

- Adds a configurable field opportunity_id: uuid.UUID = field(default_factory=uuid.uuid4)  to OpportunityContainer class
- Update opportunity creation logic to use the provided ID instead of uuid.uuid4()
- Add explicit UUIDs to all opportunity scenarios in _create_opportunity_scenarios() and the dynamic form-based opportunities
- Add a --force-recreate flag that deletes existing opportunities by number and recreates with new IDs
- Update the task to detect ID mismatches and update/replace records

## Context for reviewers

The only remaining part of the AC is `Manually go into AWS log into the DB for each environment and clean up the records`, which I'm not sure should happen after this is merged, or the timing of this action.

## Validation steps

See updated unit tests for task to confirm consistent IDs across runs and mismatches replace records.

### DIFF
--- a/api/src/task/opportunities/build_automatic_opportunities.py
+++ b/api/src/task/opportunities/build_automatic_opportunities.py
@@ -213,7 +213,6 @@ class BuildAutomaticOpportunitiesTask(Task):
             OpportunityContainer(
                 opportunity_title=f"Opportunity with ALL forms - {datetime_util.get_now_us_eastern_date().isoformat()}",
                 opportunity_number=f"SGG-ALL-Forms-{datetime_util.get_now_us_eastern_date().isoformat()}",
-                opportunity_id=uuid.uuid5(uuid.NAMESPACE_DNS, "simpler-grants-gov.all-forms"),
             ),
             competitions=[CompetitionContainer(optional_form_ids=[form.form_id for form in forms])],
         )
--- a/api/tests/src/task/opportunities/test_build_automatic_opportunities.py
+++ b/api/tests/src/task/opportunities/test_build_automatic_opportunities.py
@@ -2,6 +2,7 @@ import uuid
 from copy import deepcopy
 
 import pytest
+from freezegun import freeze_time
 from sqlalchemy import select, update
 
 from src.db.models.competition_models import Form
@@ -105,7 +106,7 @@ def test_opportunity_ids_are_consistent_across_runs(enable_factory_create, db_se
         first_run_ids[opp.opportunity_number] = opp.opportunity_id
 
     # Second run - should skip existing opportunities
-    
+
     with freeze_time("2026-02-04 12:00:00"):
         task2 = BuildAutomaticOpportunitiesTask(db_session)
         task2.run()
@@ -141,17 +142,12 @@ def test_opportunity_ids_are_consistent_across_runs(enable_factory_create, db_se
             f"expected {expected_id}, got {db_ids[opp_number]}"
         )
 
-    # Verify the ALL forms opportunity has the expected UUID5
     # The opportunity number is dynamic (includes date), so find it by prefix
-    all_forms_opp_number = next(
-        (num for num in db_ids.keys() if num.startswith("SGG-ALL-Forms-")), None
-    )
-    assert all_forms_opp_number is not None, "ALL forms opportunity not found"
-    expected_all_forms_id = uuid.uuid5(uuid.NAMESPACE_DNS, "simpler-grants-gov.all-forms")
-    assert db_ids[all_forms_opp_number] == expected_all_forms_id, (
-        f"ALL forms opportunity has incorrect ID: "
-        f"expected {expected_all_forms_id}, got {db_ids[all_forms_opp_number]}"
-    )
+    all_forms_opp_numbers = [num for num in db_ids.keys() if num.startswith("SGG-ALL-Forms-")]
+    # With freeze_time testing different dates, we should have at least 2 ALL forms opportunities
+    assert (
+        len(all_forms_opp_numbers) >= 2
+    ), f"Expected at least 2 ALL forms opportunities, found {len(all_forms_opp_numbers)}"
 
 
 def test_does_not_work_in_prod(db_session, monkeypatch):


api/src/task/opportunities/build_automatic_opportunities.py
```diff
@@ -344,17 +367,29 @@ def create_opportunity(
...
+                    extra={
+                        "opportunity_id": existing_opportunity.opportunity_id,
+                        "opportunity_number": data.opportunity_number,
+                        "force_create": force_create,
+                        "id_mismatch": existing_opportunity.opportunity_id != data.opportunity_id,
+                    },
+                )
+                self.db_session.delete(existing_opportunity)
+                self.db_session.flush()  # Ensure deletion happens before we try to insert
+            else:
```
chouinar: I'd err on the side of not including deletes at all. What happens if someone was doing testing with an opportunity created by this script? You just deleted it and all of its applications. I met with the e2e folks and showed them how this script works and told them also not to do this.

I intentionally did not include this because if you run this and didn't check with everyone potentially doing testing / UAT / etc. in the environment, you just broke everything they were doing with no way to recover it.

At the very least, don't make this the global behavior, make it the behavior of very specific opportunities, and unless explicitly setup to be deleted, it isn't affected.

api/src/task/opportunities/build_automatic_opportunities.py
```diff
@@ -186,13 +194,20 @@ def create_opportunities(self) -> None:
...
+                uuid.NAMESPACE_DNS, f"simpler-grants-gov.form.{form.form_id}"
+            )
             self.create_opportunity(
                 OpportunityContainer(
                     opportunity_title=f"Opportunity for form {form.short_form_name}",
                     opportunity_number=f"SGG-{form.short_form_name}",
+                    opportunity_id=form_opportunity_id,
                 ),
                 competitions=[CompetitionContainer(optional_form_ids=[form.form_id])],
+                force_create=self.force_recreate,
```
chouinar: Is there ever a need to recreate this? This gets made once for each form, that's it. If a form later changes, that doesn't need to be remade, it's fine to leave it alone.

mikehgrantsgov:
> #8023

Here's a recap of the changes I made:
- `--force-recreate` only affects opportunities that are explicitly marked as safe to recreate
- Form-based opportunities and any other opportunities are completely protected - they will never be deleted, even with the flag.

Also cc @joshtonava to confirm this is in the spirit of the AC.

chouinar:
> > #8023
> 
> Here's a recap of the changes I made:
> 
> * `--force-recreate` only affects opportunities that are explicitly marked as safe to recreate
> * Form-based opportunities and any other opportunities are completely protected - they will never be deleted, even with the flag.
> 
> Also cc @joshtonava to confirm this is in the spirit of the AC.

What is the use case for the recreate logic? Do we want the ability to completely delete and remake an opportunity for a particular purpose? I just want to make sure that all folks (@joshtonava  / @ErinPattisonNava  / @chris-kuryak ) know that this will delete all the applications on these opportunities as well. If you're wanting that behavior, sure, but if that's not desired, I'd say we should regroup a bit.

api/tests/src/task/opportunities/test_build_automatic_opportunities.py
```diff
@@ -93,7 +94,196 @@ def test_build_automatic_opportunities(enable_factory_create, db_session, forms)
...
+    # Verify that all opportunities from first run have the same IDs in the database
+    for opp_number, opp_id in first_run_ids.items():
+        assert opp_number in db_ids, f"Opportunity {opp_number} not found in database"
+        assert db_ids[opp_number] == opp_id, (
+            f"Opportunity ID mismatch for {opp_number}: "
+            f"expected {opp_id}, got {db_ids[opp_number]}"
+        )
+
+    # Verify specific hard-coded UUIDs for the named scenarios
+    import uuid
```
jakobpederson: Did you want this specifically down here instead of up at the top?

joshtonava:
> > > #8023
> > 
> > 
> > Here's a recap of the changes I made:
> > 
> > * `--force-recreate` only affects opportunities that are explicitly marked as safe to recreate
> > * Form-based opportunities and any other opportunities are completely protected - they will never be deleted, even with the flag.
> > 
> > Also cc @joshtonava to confirm this is in the spirit of the AC.
> 
> What is the use case for the recreate logic? Do we want the ability to completely delete and remake an opportunity for a particular purpose? I just want to make sure that all folks (@joshtonava / @ErinPattisonNava / @chris-kuryak ) know that this will delete all the applications on these opportunities as well. If you're wanting that behavior, sure, but if that's not desired, I'd say we should regroup a bit.

For some additional context I was initially thinking we would need to specify a hard coded id for all opportunities in our configs. It turns out that is impractical. I was wanting a way for us to be able to handle regenerating existing opportunities with the hard coded ID. I think the better approach we are landing on is to default to not having an opportunity id. If one is provided we will check if it exists and skip otherwise create the opportunity and related data.

cc:  @chouinar, @ErinPattisonNava

mikehgrantsgov:
> > > > #8023
> > > 
> > > 
> > > Here's a recap of the changes I made:
> > > 
> > > * `--force-recreate` only affects opportunities that are explicitly marked as safe to recreate
> > > * Form-based opportunities and any other opportunities are completely protected - they will never be deleted, even with the flag.
> > > 
> > > Also cc @joshtonava to confirm this is in the spirit of the AC.
> > 
> > 
> > What is the use case for the recreate logic? Do we want the ability to completely delete and remake an opportunity for a particular purpose? I just want to make sure that all folks (@joshtonava / @ErinPattisonNava / @chris-kuryak ) know that this will delete all the applications on these opportunities as well. If you're wanting that behavior, sure, but if that's not desired, I'd say we should regroup a bit.
> 
> For some additional context I was initially thinking we would need to specify a hard coded id for all opportunities in our configs. It turns out that is impractical. I was wanting a way for us to be able to handle regenerating existing opportunities with the hard coded ID. I think the better approach we are landing on is to default to not having an opportunity id. If one is provided we will check if it exists and skip otherwise create the opportunity and related data.
> 
> cc: @chouinar, @ErinPattisonNava

@joshtonava makes sense. I made these changes:
- Removed the `force-recreate` flag. 
- Removed `allow_force_recreate` and made the logic: If an opportunity exists by number, skip or otherwise create with the provided id.

joshtonava (dismissed):
Thanks for iterating on this

api/src/task/opportunities/build_automatic_opportunities.py
```diff
@@ -200,11 +211,11 @@ def create_opportunities(self) -> None:
         # and we don't have to adjust something existing
         self.create_opportunity(
             OpportunityContainer(
-                opportunity_title=f"Opportunity with ALL forms - {datetime_util.get_now_us_eastern_date().isoformat()}",
-                opportunity_number=f"SGG-ALL-Forms-{datetime_util.get_now_us_eastern_date().isoformat()}",
+                opportunity_title="Opportunity with ALL forms",
+                opportunity_number="SGG-ALL-Forms",
+                opportunity_id=uuid.uuid5(uuid.NAMESPACE_DNS, "simpler-grants-gov.all-forms"),
```
chouinar: I don't think we want to make these changes. The idea for this opportunity is we make a new one every day. Since the logic diffs on opportunity number, setting a static opportunity number would break that.

That means we no longer have a new opportunity every day.

mikehgrantsgov: @chouinar would something like `opportunity_number=f"SGG-ALL-Forms-{current_date_str}"`

where `current_date_str` is something like `%Y%m%d` work?

chouinar: I'd just say to revert the change back to what it was before.

mikehgrantsgov: Got it, reverted and updated the test.

api/src/task/opportunities/build_automatic_opportunities.py
```diff
@@ -202,9 +213,9 @@ def create_opportunities(self) -> None:
             OpportunityContainer(
                 opportunity_title=f"Opportunity with ALL forms - {datetime_util.get_now_us_eastern_date().isoformat()}",
                 opportunity_number=f"SGG-ALL-Forms-{datetime_util.get_now_us_eastern_date().isoformat()}",
+                opportunity_id=uuid.uuid5(uuid.NAMESPACE_DNS, "simpler-grants-gov.all-forms"),
```
chouinar: I'd revert this too - just let it be a random UUID. If I'm reading how this works, this wouldn't make a new one each run, it'd be static still since you pass in the same string each time.

api/tests/src/task/opportunities/test_build_automatic_opportunities.py
```diff
@@ -80,17 +82,73 @@ def test_build_automatic_opportunities(enable_factory_create, db_session, forms)
...
+    task1.run()
+
+    # Collect opportunity IDs from first run
+    first_run_ids = {}
+    for opp in task1.opportunities:
+        first_run_ids[opp.opportunity_number] = opp.opportunity_id
+
+    # Second run - should skip existing opportunities
+    task2 = BuildAutomaticOpportunitiesTask(db_session)
+    task2.run()
```
chouinar: If you do this, do you see any failures? This would simulate the job running on different days, thus making the all-forms opportunity get made twice.

```suggestion
    with freeze_time("2026-02-03 12:00:00"):
        task1 = BuildAutomaticOpportunitiesTask(db_session)
        task1.run()

    # Collect opportunity IDs from first run
    first_run_ids = {}
    for opp in task1.opportunities:
        first_run_ids[opp.opportunity_number] = opp.opportunity_id

    # Second run - should skip existing opportunities
    
    with freeze_time("2026-02-04 12:00:00"):
        task2 = BuildAutomaticOpportunitiesTask(db_session)
        task2.run()
```

mikehgrantsgov: It does work, it creates a new ALL forms opportunity on each different day (one for 2/3/26 and another for 2/4/2026)

PR #8207, opened 2026-01-28
[unticketed] next version bump 15.4.10 > 15.5.10

myduong-navapbc:
anchore scan failed

Screenshot 2026-01-28 at 10 03 14 AM


frontend/next.config.js
```diff
@@ -109,7 +109,7 @@ const headers = [
   },
   // don't cache user specific pages: saved-opportunities, saved-search-queries
   {
-    source: "/saved:path*",
+    source: "/saved/:path*",
```
myduong-navapbc, line 112: Fix invalid rewrite matcher (/saved:path* → /saved/:path*)

The matcher was technically invalid but tolerated by older versions. 
Newer Next.js versions enforce stricter route parsing.

will work for paths like this:
```
/saved-opportunities
/saved/anything/else
```

Issue #7651, opened 2025-12-23
Change auth on our GET competition endpoint to JWT + User Key Auth


chouinar:
### Summary

We want to modify our `GET /alpha/competitions/{competitionId}` endpoint to remove the legacy API key auth, we should instead make it support JWT auth + our new api key auth.

### Acceptance criteria

- [ ] Adjust the endpoint to use the new `jwt_or_user_api_key_multi_auth` auth which has JWT + the new API User Auth
- [ ] Replace existing tests that use the legacy auth, and add support for the new auth approach

PR #8211, opened 2026-01-28
Change auth on our GET competition endpoint to JWT and User Key Auth

kkrug:
## Summary

Fixes / Work for #7651  

## Changes proposed

Modified the GET /alpha/competitions/{competitionId} endpoint to remove the legacy API key auth and instead support JWT auth + our new api key auth. Tests were also updated to reflect this auth change.

This was also the last endpoint using the legacy authentication, so multi_auth.py and its tests were updated to remove jwt_or_key_multi_auth and its schema.

## Validation steps

Confirm the GET competition endpoint still works as intended using either the JWT or User Key auth and that all tests pass.

### DIFF
--- a/documentation/api/authentication.md
+++ b/documentation/api/authentication.md
@@ -159,8 +159,7 @@ def user_token_logout(db_session: db.Session) -> response.ApiResponse:
 ```
 
 # System-to-system Auth
-Our system-to-system (S2S) authentication approach right now is in development, while a basic key auth
-approach currently exists, further improvements are being scoped out.
+Our system supports database-based API Key authentication. These are linked to specific users and can be passed via the X-API-Key header.
 
 
 # Supporting multiple authentication schemes on an endpoint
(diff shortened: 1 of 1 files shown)


api/src/api/competition_alpha/competition_route.py
```diff
@@ -7,7 +7,7 @@
 import src.api.response as response
 from src.api.competition_alpha.competition_blueprint import competition_blueprint
 from src.auth.api_user_key_auth import api_user_key_auth
-from src.auth.multi_auth import jwt_or_key_multi_auth, jwt_or_key_security_schemes
+from src.auth.multi_auth import jwt_or_user_api_key_multi_auth, jwt_or_user_api_key_security_schemes
```
chouinar, line 10: Since I think this is the last endpoint with `jwt_or_key_multi_auth` - could you remove that + the security scheme from the multi_auth file?

mdragon:
Past changes like this broke FE calls to the API endpoints so if we can test out the FE locally as much as possible before merge and then do more of a full stack smoke test after this goes to Staging that'd be great.

api/src/api/competition_alpha/competition_route.py
kkrug: Multi_auth.py has been updated and the tests that depend on jwt_or_key_multi_auth have been removed

chouinar:
> Past changes like this broke FE calls to the API endpoints so if we can test out the FE locally as much as possible before merge and then do more of a full stack smoke test after this goes to Staging that'd be great.

This was the endpoint that specifically broke. @kkrug - could you run the e2e tests on your branch? That should catch the issue if it happens again.

kkrug:
@mdragon @chouinar I ran the e2e test locally on both main and this branch and they had the same test failures (actually main had 2 more than the branch). I think we should be good.

Issue #8015, opened 2026-01-15
Enhanced Agency fields search behavior for Opportunities


joshtonava:
### Summary

Users searching for top-level agencies like "HHS" need results that include all sub-agencies (e.g., "HHS" should return "HHS-NIH", "HHS-CDC", etc.). Currently the search only matches exact agency codes - "HHS" returns only "HHS" opportunities, not sub-agencies. Additionally, agency full names are not searchable in the default scoring rule. This adds a `top_level_agency_code` field to enable hierarchical agency matching, adds agency name fields to the default search, and replaces the deprecated agency field name with `agency_code`.

#### Model & Schema Changes

- Add `top_level_agency_code` property to Opportunity model, following pattern of `top_level_agency_name`
- Expose field in `OpportunityV1Schema` for search index serialization
- Update `OPP_REQUEST_FIELD_NAME_MAPPING` to include: `top_level_agency_code":"top_level_agency_code.keyword`

### OpenSearch Scoring Rules Changes

`api/src/services/opportunities_v1/experimental_constant.py`

DEFAULT Rule (lines 3-15)

Remove:
```
"agency^16",
"agency.keyword^16",
```

Add (in this order):
```
"agency_code^16",
"agency_code.keyword^16",
"top_level_agency_code^16",
"top_level_agency_code.keyword^16",
"agency_name",
"top_level_agency_name",
```

Result: 6 agency-related fields instead of 2, with code fields boosted ^16 and name fields unboosted

EXPANDED Rule (lines 17-31)

Remove:
```
"agency",
"agency.keyword",
```

Add (after the removed lines):
```
"agency_code",
"agency_code.keyword",
"top_level_agency_code",
"top_level_agency_code.keyword",
```

Note: `agency_name` and `top_level_agency_name` already exist in EXPANDED - leave them unchanged

AGENCY Rule (lines 34-43)

Remove:
```
"agency",
"agency.keyword",
```

Add (after the removed lines):
```
"agency_code",
"agency_code.keyword",
"top_level_agency_code",
"top_level_agency_code.keyword",
```

Note: `agency_name` and `top_level_agency_name` already exist in AGENCY - leave them unchanged

### Why These Changes
- **Deprecated field cleanup:** agency is an alias for `agency_code`, using correct field name
- **Hierarchical matching:** `top_level_agency_code` enables "HHS" to match "HHS-NIH", "HHS-CDC"
[...]

PR #8212, opened 2026-01-28
[Issue #8015] enhanced search opportunities [1/2]

babebe:
## Summary

Fixes for #8015 

## Changes proposed

Update Opportunity model to derive top_level_agency_code from its agency_record.
Update search index loading to include top_level_agency_code for opportunities.

## Context for reviewers

Previously, top_level_agency_code was not included in the search index. With this update, the Opportunity model provides a property that resolves the top-level agency code from its associated agency_record (or falls back to agency_code if none exists). The index load now includes this value so that searches  by top-level agency are possible.

No other model behavior is changed; this is purely an additive update to support search functionality.

 *** This change will trigger creation of a new OpportunityVersion record because a new field is being added. However, it will not generate a notification, since this field is not tracked for notification purposes.***
## Validation steps

Build an opportunity with a multi-level agency
Confirm top_level_agency_code is indexed


api/docker-compose.yml
```diff
@@ -40,18 +40,18 @@ services:
       - default
 
   # opensearch-dashboards disabled to save resources in CI
-  # opensearch-dashboards:
```
joshtonava, line 43: **question:** Was this meant for testing or are we going to re enable this ?

I see the note above `disabled to save resources in CI` but im not sure of the context

cc: @prasnava

joshtonava:
**issue:** I think we need to add top_level_agency_code to the OPP_REQUEST_FIELD_NAME_MAPPING for sorting/filtering by top_level_agency_code.

api/docker-compose.yml
prasnava: The pipelines where failing due to memory constraints of the github runners. So we disabled this to reduce the memory usage. We can still use it locally, but just not in the pipeline.

chouinar: If we uncomment it, but don't add it to the make init command, and instead make a make command to turn the dashboards on, I think that would be safe / wouldn't get run in CI?

chouinar:
> **issue:** I think we need to add top_level_agency_code to the OPP_REQUEST_FIELD_NAME_MAPPING for sorting/filtering by top_level_agency_code.

Yeah - I suggest we add some tests for the search endpoint that verify we can query and get expected results.

api/docker-compose.yml
babebe: Added command to turn it on.

api/src/db/models/opportunity_models.py
```diff
@@ -159,6 +159,13 @@ def all_non_forecasts(self) -> list[OpportunitySummary]:
         # Note this will include historical and deleted records.
         return [summary for summary in self.all_opportunity_summaries if not summary.is_forecast]
 
+    @property
+    def top_level_agency_code(self) -> str | None:
+        if self.agency_record is not None and self.agency_record.top_level_agency is not None:
```
jakobpederson, line 164: I don't this warrants a change but I think could also say
```
if self.agency_record and self.agency_record.top_level_agency:
```

api/tests/src/search/backend/test_load_opportunities_to_index.py
```diff
@@ -0,0 +1,179 @@
...
+        all_indexed_opportunity_ids = set([record["opportunity_id"] for record in resp.records])
+
+        # Convert our test opportunities to string IDs for comparison
+        expected_included_ids = set([str(opp.opportunity_id) for opp in included_opportunities])
+
+        # Verify that ALL of our expected opportunities are present in the index
+        missing_included = expected_included_ids - all_indexed_opportunity_ids
+        assert (
+            not missing_included
+        ), f"Expected opportunities missing from index: {missing_included}"
```
jakobpederson, line 179: (optional) I feel like I understand but a possible alternative could be `len(missing_included) == 0`

api/docker-compose.yml
```diff
@@ -39,19 +39,18 @@ services:
...
-  #   ports:
-  #     - 5601:5601 # Map host port 5601 to container port 5601
-  #   expose:
-  #     - "5601" # Expose port 5601 for web access to OpenSearch Dashboards
-  #   environment:
-  #     - 'OPENSEARCH_HOSTS=["http://opensearch-node:9200"]'
-  #     - DISABLE_SECURITY_DASHBOARDS_PLUGIN=true # disables security dashboards plugin in OpenSearch Dashboards
-  #   networks:
-  #     - default
+  opensearch-dashboards:
```
joshtonava, line 41: nit: Can you fix the indentation in this block 😬 

Otherwise I think we are good

api/src/db/models/opportunity_models.py
babebe: Yes that is technically correct. Since the field is Nullable, checking for not None is safer.

api/tests/src/search/backend/test_load_opportunities_to_index.py
babebe: Yup. that's the same. This is shorter, and works for lists

Issue #7966, opened 2026-01-13
Add "Posted date" filter to Search drawer


myduong-navapbc:
## Summary

Add a Posted date filter to the Search page filters drawer

Design Files : https://www.figma.com/design/zPlDWcYM396szw2yL3UXvk/Simpler-Grants-2025?node-id=9291-55169&t=wYtCQlA0hK4UMX7k-4

### Background / Context
- The Search page renders the filter drawer via `SearchDrawerFilters`:
  - `frontend/src/app/[locale]/search/page.tsx` (Search page)
  - `frontend/src/components/search/SearchDrawerFilters.tsx` (drawer contents)
- Existing date-range filter pattern:
  - Close Date uses `RadioButtonFilter` + `SearchFilterRadio`
  - Query param wiring uses `useSearchParamUpdater` (`setQueryParam`)
- Search param types / keys:
  - `ValidSearchQueryParam` / `validSearchQueryParamKeys` in `src/types/search/searchQueryTypes`
  - `convertSearchParamsToProperTypes` and `formatSearchRequestBody` handle query param conversion + API body creation.
- Current filters supported include `postedDate` (relative days).

### Acceptance Criteria

- [ ] Add a new accordion section in the filter drawer titled **"Posted date"**.
- [ ] Use the same UX pattern as Close Date:
- [ ]   Radio options (single select)
- [ ]   Includes "Any posted date" (or "Any posted date" equivalent) option when nothing is selected
- [ ] Selecting an option updates the URL query param and triggers search updates (same behavior as other filters).
- [ ] Clearing the selection resets the query param (same behavior as "Any" option for Posted Date).
- [ ] Works in mobile drawer and desktop layout (where applicable).
- [ ] Unit tests added/updated for the filter UI wiring.
- [ ] UAT

nathan-stricker:
@eric-serrato95 @chelsea-nava This is ready for UAT!

eric-serrato95:
@nathan-stricker UAT looks good, closing ticket now!

Issue #7967, opened 2026-01-13
Add “Posted date” pill labels in FilterPillPanel


myduong-navapbc:
## Summary

See [issue 7966](https://github.com/HHS/simpler-grants-gov/issues/7966) for more context.
Design Files: https://www.figma.com/design/zPlDWcYM396szw2yL3UXvk/Simpler-Grants-2025?node-id=9291-55169&t=wYtCQlA0hK4UMX7k-4

- Pills are rendered in `FilterPillPanel`:
  - `frontend/src/components/search/FilterPillPanel.tsx`
- Pill formatting logic lives in:
  - `frontend/src/utils/search/searchFilterUtils.ts` (functions like `formatPillLabel`, `formatPillLabels`)
- Existing pill formatting includes special handling:
  - `closeDate` > `Closing within last ${value} days`

### Acceptance Criteria

- [ ] When `postedDate` query param is selected, render a pill for it.
- [ ] The pill label should be human readable and consistent with Close Date style.
- [ ] `Any` date selected should not render a pill.
- [ ] `Posted within last 7 days`
- [ ] Removing the pill clears the `postedDate` query param and updates results (same behavior as other pills).
- [ ] If the postedDate value is missing/unknown, avoid crashing and fail gracefully (empty/omitted label).
- [ ] Unit tests added/updated.
- [ ] UAT

chelsea-nava:
@chay-REIsys I'm available if you need support or clarity

eric-serrato95:
@chay-REIsys UAT complete, moving to done!

PR #8214, opened 2026-01-28
[Issue #7966] Add Posted Date filter to Search Drawer

nathan-stricker:
## Summary

Work for #7966 

## Changes proposed

- Added "Posted Date" accordion to search filter drawer
- Added radio options under accordion
- Implemented request formatting for posted-date filtering (looks backward from start date, vs close-date looking forwards from end date)
- Added backend support for `post_date` facet counts
- Added and updated unit tests

## Context for reviewers

Design Files: https://www.figma.com/design/zPlDWcYM396szw2yL3UXvk/Simpler-Grants-2025?node-id=9291-55169&t=wYtCQlA0hK4UMX7k-4 

## Validation steps

1. Navigate to the search page and open the Filters drawer.
2. Validate the new filter exists and is placed correctly:
    - [x] A new accordion section titled “Posted date” appears between “Category” and “Closing date range”
    - [x] It uses radio buttons (single select) and shows an “Any posted date” option when nothing is selected
3. Validate URL/query-param behavior:
    - [x] Selecting “Within the last 7 days” (or any option) updates the URL with postedDate=<N>
    - [x] Selecting “Any posted date” removes/clears the postedDate query param
4. Validate results update behavior:
    - [x] After selecting a posted-date option, the search results refresh and reflect filtering
    - [x] Clearing back to “Any posted date” restores the broader result set (removes the posted-date constraint)
5. Validate facet counts:
    - [x] Each posted-date radio option displays a count in brackets (like Close Date range does)
    - [x] Counts change appropriately when other filters are applied (e.g., add a Category filter and confirm Posted date counts update)

### DIFF
--- a/frontend/src/components/workspace/SavedSearchesList.test.tsx
+++ b/frontend/src/components/workspace/SavedSearchesList.test.tsx
@@ -58,6 +58,7 @@ const fakeParamDisplayMapping = {
   page: "page",
   sortby: "sortby",
   closeDate: "closeDate",
+  postedDate: "postedDate",
   costSharing: "costSharing",
   topLevelAgency: "topLevelAgency",
   andOr: "andOr",
--- a/frontend/src/errors.test.ts
+++ b/frontend/src/errors.test.ts
@@ -11,6 +11,7 @@ describe("BadRequestError (as an example of other error types)", () => {
     assistanceListingNumber: new Set(["15.817"]),
     category: new Set(["science"]),
     closeDate: new Set(["500"]),
+    postedDate: new Set(),
     costSharing: new Set(["true"]),
     topLevelAgency: new Set(["CDC"]),
     query: "space exploration",


chaitanya-duddella:
validated 
image

image

api/src/services/opportunities_v1/search_opportunities.py
```diff
@@ -80,6 +80,14 @@
     {"from": "now", "to": "now+120d/d", "key": "120"},
 ]
 
+STATIC_POSTED_DATE_RANGES: list = [
+    {"from": "now-3d/d", "to": "now", "key": "3"},
+    {"from": "now-7d/d", "to": "now", "key": "7"},
+    {"from": "now-14d/d", "to": "now", "key": "14"},
+    {"from": "now-30d/d", "to": "now", "key": "30"},
+    {"from": "now-60d/d", "to": "now", "key": "60"},
+]
```
chouinar, line 89: Should we have the keys be `-3` / `-7` / etc? That way it's clear they're past dates, not future ones like the close date range we have?

nathan-stricker: I could see an argument for either way, but would lean toward leaving them as they are to stay consistent across the frontend when these are later used for the formatted pill labels in https://github.com/HHS/simpler-grants-gov/issues/7967 (I.e. it will say something like "Posted within ${value} days"), but happy to change if you think negative values are clearer

chouinar: I'm also thinking of folks that use the API directly, I'd say negative makes more sense. I get the UX won't care much with how it's worded (there's an intuitive, posted must be in past, closed is in the future with this). Could adjust the frontend to to an absolute value on it to merge the two together?

Issue #8021, opened 2026-01-15
Database updates


brianlee-nava:
### Summary

There are 4 main tables within the Opportunity Data Model that Plan and Publish will need to work with. Those are Opportunity, Opportunity Attachment, Opportunity Summary, and Opportunity Assistance Listing.

In the past, opportunities were imported from Grants.gov into Simpler.Grants.gov. In order to make Opportunity Publishing work in Simpler Grants Management, we need:
1. A way to distinguish which system (Grants.gov or Simpler.Grants.gov) created the opportunity. 
2. Allow the legacy_opportunity_id column in the opportunity table to accept nulls. This column is the Grants.gov opportunity_id which is N/A for opportunities created by Simpler.Grants.gov

Additionally, since most opportunities are created by copying an existing opportunity and updating it, we want:
3. A way to link the original opportunity with the derived opportunity.

### Acceptance criteria

- [ ] Add a SGG Opportunity flag to indicate that it is a Simpler.Grants.gov created opportunity
- [ ] Update Legacy_opportunity_id column to allow nulls
- [ ] Create the Referenced Opportunity table to link the original opportunity with the derived opportunity

brianlee-nava:
@glenblosser-nava consult @chouinar who is authorized to make db changes

chouinar:
@brianlee-nava / @glenblosser-nava 
Any engineer can change the DB, that's not something we do manually. You change the SQLAlchemy DB model, generate a DB migration (all locally) and that migration runs when we merge to an environment. It's not restricted, an open source contributor could make a PR with a DB change.

I go over this in my walkthrough from November: https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/1234370636/Engineering#Walkthroughs

It is also documented in https://github.com/HHS/simpler-grants-gov/blob/main/documentation/api/database/database-management.md

This is a very simple task (I had Hao and Kenny make DB changes as beginner tasks). We've made _many_ changes to the DB: https://github.com/HHS/simpler-grants-gov/tree/main/api/src/db/migrations/versions

Dao-REI:
Watching [Michael's API/Backend Walkthrough](https://drive.google.com/file/d/1MShOxg9mRqEs-ZyLRl_qjnIyxMlykES3/view?usp=drive_link) I've figured out most of this implementation:

1. opportunity_models.py - update the definition for legacy_opportunity_id to allow nulls
[...]

PR #8234, opened 2026-01-29
[Issue #8021] Database Updates for Opportunity Data Model

dghazvini:
## Summary

Fixes for #8021  

Database updates for the Opportunity Model as part of Phase 1 of implementing Opportunity Publishing in Simpler Grants.

## Changes proposed

Performing Database updates as part of the #8021 ticket, for preliminary work for enabling Opportunity Publishing to work in Simpler Grants Management.

We are making updates to the Opportunity table to fulfill AC1 and AC2 as defined in ticket #8021, as well as creating a new ReferencedOpportunity table to fulfill AC3.

-Updating Legacy_opportunity_id column to be nullable
-Added a new column to serve as a flag to indicate that it is a Simpler.Grants.gov created opportunity
-Created the Referenced Opportunity table to link the original opportunity with the derived opportunity

## Context for reviewers

More context in Ticket #8021 regarding Opportunity Publishing and in the [Tech Spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2612789251/Phase+1+Core+Foundations) for Phase 1 of Opportunity Publishing in Simpler Grants.

## Validation steps

No manual verification; added Factory method for ReferencedOpportunity.

### DIFF
--- a/api/src/services/xml_generation/constants.py
+++ b/api/src/services/xml_generation/constants.py
@@ -15,7 +15,8 @@ NO_VALUE = "N: No"
 
 # Base URL for schema locations
 # This matches the legacy Grants.gov URL structure
-SCHEMA_LOCATION_BASE_URL = "https://apply07.grants.gov/apply/opportunities/schemas/applicant"
+# Format: https://apply07.grants.gov/apply/opportunities/schemas/agency/oppOPP_NUMBER-cfdaCFDA_NUMBER.xsd
+SCHEMA_LOCATION_BASE_URL = "https://apply07.grants.gov/apply/opportunities/schemas/agency"
 
 # Namespace Definitions
 # These match the legacy Grants.gov XML structures
--- a/api/src/services/xml_generation/submission_xml_assembler.py
+++ b/api/src/services/xml_generation/submission_xml_assembler.py
@@ -204,8 +204,8 @@ class SubmissionXMLAssembler:
         # Add header (must strip XML declaration and use just the element)
         root.append(header_element)
 
-        # Add Forms wrapper
-        forms_element = lxml_etree.SubElement(root, "Forms")
+        # Add Forms wrapper with grant: namespace prefix for consistency with legacy
+        forms_element = lxml_etree.SubElement(root, f"{{{grant_ns}}}Forms")
         for form_element in form_elements:
             forms_element.append(form_element)
 
@@ -225,11 +225,55 @@ class SubmissionXMLAssembler:
     def _get_schema_location_url(self) -> str | None:
         """Get the schema location URL for xsi:schemaLocation attribute.
 
-        Constructs the URL from competition's legacy_package_id.
+        Constructs the URL from competition's opportunity number and CFDA number.
+        Format: https://apply07.grants.gov/apply/opportunities/schemas/agency/oppOPP_NUMBER-cfdaCFDA_NUMBER.xsd
+
+        Example: https://trainingapply.grants.gov/apply/opportunities/schemas/agency/oppSIMP-QUAD4-FORMS-10202025-cfda00.000.xsd
+
+        Returns:
+            Schema location URL or None if opportunity_number is not set
         """
-        return (
-            f"{SCHEMA_LOCATION_BASE_URL}/" f"{self.application.competition.legacy_package_id}.xsd"
-        )
+        opportunity_number = self.application.competition.opportunity.opportunity_number
+
+        if not opportunity_number:
+            logger.warning(
+                "Competition opportunity_number is not set - cannot generate schema location",
+                extra={
+                    "application_id": self.application.application_id,
+                    "competition_id": self.application.competition.competition_id,
+                },
+            )
+            return None
+
+        # Get assistance listing number from the competition
+        # Note: A competition is associated with a single assistance listing (or none)
+        cfda_suffix = ""
+        opportunity_assistance_listing = self.application.competition.opportunity_assistance_listing
+
+        if opportunity_assistance_listing is not None:
+            assistance_listing_number = opportunity_assistance_listing.assistance_listing_number
+            if assistance_listing_number:
+                cfda_suffix = f"-cfda{assistance_listing_number}"
+            else:
+                logger.info(
+                    "Assistance listing exists but assistance_listing_number is None - XSD filename will not include CFDA",
+                    extra={
+                        "application_id": self.application.application_id,
+                        "competition_id": self.application.competition.competition_id,
+                    },
+                )
+        else:
+            logger.info(
+                "No assistance listing found for competition - XSD filename will not include CFDA",
+                extra={
+                    "application_id": self.application.application_id,
+                    "competition_id": self.application.competition.competition_id,
+                },
+            )
+
+        # Construct XSD filename with 'opp' prefix + opportunity number + optional CFDA suffix
+        xsd_filename = f"opp{opportunity_number}{cfda_suffix}.xsd"
+        return f"{SCHEMA_LOCATION_BASE_URL}/{xsd_filename}"
 
     def _parse_xml_string(self, xml_string: str) -> lxml_etree.Element:
         """Parse XML string into element tree."""
--- a/api/src/services/xml_generation/validation/test_cases.py
+++ b/api/src/services/xml_generation/validation/test_cases.py
@@ -13,11 +13,11 @@ SF424_TEST_CASES = [
             "organization_name": "Test Organization",
             "employer_taxpayer_identification_number": "123456789",
             "sam_uei": "TEST12345678",
-            "applicant_address": {
+            "applicant": {
                 "street1": "123 Main St",
                 "city": "Washington",
                 "state": "DC: District of Columbia",
-                "zip_postal_code": "20001",
+                "zip_code": "20001",
                 "country": "USA: UNITED STATES",
             },
             "phone_number": "555-123-4567",
@@ -61,11 +61,11 @@ SF424_TEST_CASES = [
             "organization_name": "Test Organization",
             "employer_taxpayer_identification_number": "123456789",
             "sam_uei": "TEST12345678",
-            "applicant_address": {
+            "applicant": {
                 "street1": "123 Main St",
                 "city": "Washington",
                 "state": "DC: District of Columbia",
-                "zip_postal_code": "20001",
+                "zip_code": "20001",
                 "country": "USA: UNITED STATES",
             },
             "phone_number": "555-123-4567",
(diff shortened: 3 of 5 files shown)


api/src/db/models/opportunity_models.py
```diff
@@ -36,7 +36,9 @@ class Opportunity(ApiSchemaTable, TimestampMixin):
 
     opportunity_id: Mapped[uuid.UUID] = mapped_column(UUID, primary_key=True, default=uuid.uuid4)
 
-    legacy_opportunity_id: Mapped[int] = mapped_column(BigInteger, index=True, unique=True)
+    legacy_opportunity_id: Mapped[int | None] = mapped_column(BigInteger, index=True, unique=True, nullable=True)
```
chouinar: Just rely on the `Mapped` type. It's not wrong to add `nullable=True`, but technically `my_field: Mapped[int] = mapped_column(nullable=False)` (or vice-versa) won't make anything complain until we try to use it. Easiest to just rely on the typing for that. Same reason we generally leave the type out and use the default - this case being an exception as we want a `BigInteger` instead of a regular one which int would normally map to.

```suggestion
    legacy_opportunity_id: Mapped[int | None] = mapped_column(BigInteger, index=True, unique=True)
```

api/src/db/models/opportunity_models.py
```diff
@@ -36,7 +36,9 @@ class Opportunity(ApiSchemaTable, TimestampMixin):
...
+
+    is_simpler_grants_opportunity: Mapped[bool] = mapped_column(default=False, index=True)
```
chouinar: This has to be nullable as we already have records in our environments. If this migration ran against a DB, it'd fail. Technically we could make it not null by:
* Adding the column as nullable
* Populate the column
* Make the column not-null

But that's annoying, and we can also just be fine with null values and assume False/null are equivalent.

```suggestion
    is_simpler_grants_opportunity: Mapped[bool | None] = mapped_column(index=True)
```

api/src/db/models/opportunity_models.py
```diff
@@ -458,6 +470,39 @@ class OpportunityChangeAudit(ApiSchemaTable, TimestampMixin):
...
+
+    referenced_opportunity_id: Mapped[uuid.UUID] = mapped_column(UUID, primary_key=True, default=uuid.uuid4)
+    
+    original_opportunity_id: Mapped[uuid.UUID] = mapped_column(
+        UUID, ForeignKey("api.opportunity.opportunity_id"), index=True
+    )
+    
+    derived_opportunity_id: Mapped[uuid.UUID] = mapped_column(
+        UUID, ForeignKey("api.opportunity.opportunity_id"), index=True
+    )
```
chouinar, line 493: Prefer actual objects over strings when possible.

```suggestion
    original_opportunity_id: Mapped[uuid.UUID] = mapped_column(
        UUID, ForeignKey(Opportunity.opportunity_id), index=True
    )
    
    derived_opportunity_id: Mapped[uuid.UUID] = mapped_column(
        UUID, ForeignKey(Opportunity.opportunity_id), index=True
    )
```

api/tests/src/db/models/factories.py
```diff
@@ -492,6 +492,16 @@ class Meta:
...
 
+class ReferencedOpportunityFactory(BaseFactory):
+    class Meta:
+        model = opportunity_models.ReferencedOpportunity
+
+    referenced_opportunity_id = Generators.UuidObj
+    
+    original_opportunity_id = factory.LazyFunction(lambda: OpportunityFactory.create().opportunity_id)
+    derived_opportunity_id = factory.LazyFunction(lambda: OpportunityFactory.create().opportunity_id)
+
```
chouinar: Don't need to call factories like this from within factories. We can use the relationships like so:


```suggestion
    original_opportunity = factory.SubFactory(OpportunityFactory)
    original_opportunity_id = factory.LazyAttribute(lambda a: a.original_opportunity.opportunity_id)
    
    derived_opportunity = factory.SubFactory(OpportunityFactory)
    derived_opportunity_id = factory.LazyAttribute(lambda a: a.derived_opportunity.opportunity_id)
```

LazyAttributes get evaluated on a second pass after everything else is initialized. Note that it's not strictly necessary to do this, there's a few use cases of factories that do require them (specifically if you do `build()` rather than `create()` which makes the object, but doesn't put it in the DB). In most cases we use `create()` so the ID would get populated automatically by the ORM by setting it on the relationship.

jakobpederson:
Did you want to add or update a test just to show that the derived opportunities are showing up correctly? I don't think it's necessary but it just struck me you had the factories and updated model, but no tests actually including them.

chouinar:
> Did you want to add or update a test just to show that the derived opportunities are showing up correctly? I don't think it's necessary but it just struck me you had the factories and updated model, but no tests actually including them.

Tests for tables aren't something we generally do since they become unnecessary as soon as we add an implementation that uses them which usually closely follows.

api/src/db/models/opportunity_models.py
dghazvini: Updated based on this feedback.

api/src/db/models/opportunity_models.py
dghazvini: Updated based on feedback.

api/src/db/models/opportunity_models.py
dghazvini: Updated based on feedback.

api/tests/src/db/models/factories.py
```diff
@@ -492,6 +492,27 @@ class Meta:
...
 
+class ReferencedOpportunityFactory(BaseFactory):
+    class Meta:
+        model = opportunity_models.ReferencedOpportunity
+        exclude = ('original_opportunity', 'derived_opportunity')
+
+    referenced_opportunity_id = Generators.UuidObj
+    
+    @classmethod
+    def _create(cls, model_class, *args, **kwargs):
```
chouinar: I'm not quite following this change, it should be rare we need to use `_create` ourselves with factories, I think we only do it for some file related factories because they need to go outside of what you have here is a fairly standard table.

Was there an issue with this:
```py
    original_opportunity = factory.SubFactory(OpportunityFactory)
    original_opportunity_id = factory.LazyAttribute(lambda a: a.original_opportunity.opportunity_id)
    
    derived_opportunity = factory.SubFactory(OpportunityFactory)
    derived_opportunity_id = factory.LazyAttribute(lambda a: a.derived_opportunity.opportunity_id)
```

dghazvini: Sorry, meant to leave a comment yesterday evening after committing this but got busy...

I tried implementing with SubFactory as suggested, but encountered a SQLAlchemy bidirectional attribute conflict:
`ValueError: Bidirectional attribute conflict detected: Passing object <Opportunity at 0x725e54a796a0> to attribute "ReferencedOpportunity.original_opportunity" triggers a modify event on attribute "ReferencedOpportunity.derived_opportunity" via the backref "Opportunity.derived_opportunities".`

Which from what I understood is an error with SQLAlchemy trying to establish both sides of the relationship at the same time, and is happening because the ReferencedOpportunity model has two relationships to the same 
Opportunity model.
The exclude statement from what I understand is for preventing Factory Boy from setting the original_opportunity and derived_opportunity relationship attributes on the ReferencedOpportunity instance. Then `_create` is overwritten to manually create the Opportunity objects first. Then foreign key IDs are set without triggering SQLAlchemy's relationship machinery.
There probably is a better solution that follows the convention of other Factories but I didn't see one that seemed to fit for this particular case.

chouinar: That issue isn't a factory one, you'd hit that issue just working with the SQLAlchemy models normally.

The issue is having back_populates on both side of the relationship. That param effectively makes it so if I do:
```py
my_opportunity.derived_opportunities.append(referenced_opp)

# this is now set
assert referenced_opp.derived_opportunity is my_opportunity
```

It's a nice-to-have, and gets tricky with complex models. Since there's a few opportunities here, it doesn't like trying to sync this to the other opportunity. opportunity -> referenced_opportunity -> opportunity -> referenced_opportunity becomes a never ending loop. Easiest fix, remove the back_populates on one side so it's not a loop.

Since I'm going to guess that the most likely way we'll use a referenced opportunity is like:

```py
referenced_opportunity = ReferencedOpportunity(original_opportunity=opp_a, derived_opportunity=opp_b)
```

I think removing it from the opportunity side is what we'd want.

dghazvini: Got it, makes sense. I wasn't getting the issue when running the migrations so figured it was just a Factory setup issue. I removed the back_populates on the opportunity side and now the factory is running with the original code block you mentioned.

dghazvini:
Looking at the build check fail it looks like some code formatting check is failing. Does a formatter need to be run on these files before committing? Not sure if I missed that in some documentation.

chouinar:
> Looking at the build check fail it looks like some code formatting check is failing. Does a formatter need to be run on these files before committing? Not sure if I missed that in some documentation.

`make format lint` will format and then run the linter for anything that isn't automatically handled. The CI runs these and errors if anything would change.

api/tests/src/db/models/factories.py
chouinar: The `back_populates` issues wouldn't pop up in the migrations, changing that won't even affect the migrations since those only include what goes in the DB. Any ORM specific stuff like this isn't - basically foreign keys would go in a migration, but the relationships are just in Python so if you change those, nothing happens to the DB itself.

dghazvini:
> > Looking at the build check fail it looks like some code formatting check is failing. Does a formatter need to be run on these files before committing? Not sure if I missed that in some documentation.
> 
> `make format lint` will format and then run the linter for anything that isn't automatically handled. The CI runs these and errors if anything would change.

Resolved the formatting issues.

api/src/db/models/opportunity_models.py
```diff
@@ -458,6 +472,38 @@ class OpportunityChangeAudit(ApiSchemaTable, TimestampMixin):
     is_loaded_to_version_table: Mapped[bool | None] = mapped_column(index=True)
 
 
+class ReferencedOpportunity(ApiSchemaTable, TimestampMixin):
+    __tablename__ = "referenced_opportunity"
+
+    __table_args__ = (UniqueConstraint("derived_opportunity_id"), ApiSchemaTable.__table_args__)
```
chouinar: Shouldn't this unique constraint have derived opportunity ID + original opportunity ID?

dghazvini: Ah makes sense... so the way I set it up before is it would just constrain it so that the same derived opp can't be linked to multiple original opps, which isn't what we want. Adding original opportunity to this makes it so that the specific combination of original + derived IDs can only appear once, which is what we want. Thanks, pushed an update for this.

PR #8235, opened 2026-01-29
[Issue #8065] Add lookup tables enums for workflow management

### LINKED ISSUE #8065: Add lookup tables and enums for Workflow Management
### Summary

We want the following lookup tables created:

* `lk_workflow_type` - connected to WorkflowType enum
* `lk_approval_type` - connected to ApprovalType enum

We also want the following enums, with the following values:

WorkflowType
* `opportunity_publish`
* `application_submission`
* `initial_prototype` - FYI this is the one we’ll actually build in the near term for testing/demoing

ApprovalType - Not sure what we want here yet, just as a placeholder add
* `initial_prototype_approval`

WorkflowEntityType - this won’t be connected to a table, but will use it elsewhere.
* `opportunity`
* `application`

WorkflowEventType
* `start_workflow`
* `process_workflow`

### Acceptance criteria

- [ ] Lookup enums created and connected to LK tables for workflow type and approval type
- [ ] Database migration generated

hao10282025-sudo:
## Summary

Fixes #8065  

## Changes proposed

Add lookup tables and enums for the Workflow Management

## Context for reviewers

tables with enums:
- lk_workflow_type 
- lk_approval_type 
- lk_approval_response_type 

entity with enums (no table):
- WorkflowEntityType 
- WorkflowEventType

## Validation steps

- Lookup enums created and connected to LK tables for workflow type and approval type
- Database migration generated


api/src/constants/lookup_constants.py
```diff
@@ -300,3 +300,29 @@ class OrganizationAuditEvent(StrEnum):
...
+
+
+class WorkflowEntityType(StrEnum):
+    OPPORTUNITY = "opportunity"
+    APPLICATION = "application"
+
+
+class WorkflowEventType(StrEnum):
+    START_WORKFLOW = "start_workflow"
+    PROCESS_WORKFLOW = "process_workflow"
```
kkrug, line 328: Do these need to be registered in lookup_models.py?

api/src/db/models/lookup_models.py
```diff
@@ -573,3 +598,45 @@ def from_lookup(cls, lookup: Lookup) -> LkOrganizationAuditEvent:
...
             organization_audit_event_id=lookup.lookup_val, description=lookup.get_description()
         )
+
+
+@LookupRegistry.register_lookup(WORKFLOW_TYPE_CONFIG)
+class LkWorkflowType(LookupTable, TimestampMixin):
+    __tablename__ = "lk_workflow_type"
+
+    workflow_type_id: Mapped[int] = mapped_column(primary_key=True)
+    description: Mapped[str]
```
kkrug, line 608: Would this be better off called 'name' instead of 'description?' And should it have a unique constraint?

chouinar: Description is what we've called it on all the other lookup tables - uniqueness isn't really needed, that's technically enforced by the python enums, and this field actually isn't used by anything. It exists just in case you're in the DB and want to see what the lookup value string is, but we never actually query these tables, just use them for foreign keys to enforce enums.

api/src/constants/lookup_constants.py
chouinar: They won't be written to the DB, need them for some other workflow logic outside of the DB.

The WorkflowEntityType is going to be for defining what kind of entity a workflow expects / is used in in-code validation. Won't be stored in the DB, at least not yet.

I suppose workflow event type could be written to the DB, although I believe I was planning for this to be inside a JSON object anyways so wouldn't need that foreign key.

hao10282025-sudo: Hmm nope, these two entities do not connect to tables as the task summary
As I understand, these two may not have that strong "DB characteristics"

PR #8244, opened 2026-01-29
[Issue #8068] Add New Relic transcation logging to workflow script

### LINKED ISSUE #8068: Setup a new equivalent for ECS background task for this service and get New Relic transaction logging working
### Summary

TODO - write this up

### Acceptance criteria

- [ ] TODO

chouinar:
## Summary

Fixes #8068 

## Changes proposed
* Adds New Relic APM support to the workflow script
* Adds general logging setup similar to our other ECS tasks to the workflow script

## Context for reviewers
On our ECS tasks, we always have an `@ecs_background_task` decorator. This decorator helps get our logs and metrics working in New Relic, connected to the APM entity. We need that as well for the workflow script, but it works slightly different in that it'll be an always on service processing events. To get things to work, we need New Relic to have a "transaction" concept with the task, our other ECS tasks just have this transaction be "the entire job", but since we'll be processing events, makes sense to do that by event, so that's what we did here.

Note, some of the event type stuff I added will be rewritten when we actually build that, just need something to test with.

## Validation steps
Unfortunately validation requires access to APM in New Relic which very few folks have, so just sharing what I did as it's not easily repeated. I first adjusted the `newrelic.ini` to have our New Relic key and adjust the config so our local instance would report to New Relic, and then ran the script with `make cmd args="workflow workflow-main"`. To help with validating things, I added "event types" so I could have more than one thing in the UI. After running this for a few minutes, the UI now has events, and even separates them by type.

Screenshot 2026-01-29 at 1 04 22 PM

Screenshot 2026-01-29 at 1 04 18 PM


chouinar:
> Implementing New Relic transactions are new to me but also it looks like nothing is really happening beyond basic logging, so I'm going to approve.

Transactions are useful to have as they show a bunch of information about what the code was doing. The API has transactions for requests and so if GET opportunity is taking a long time, it'll break down all the queries/function calls and give us a breakdown of what is taking so long. I've used this in the past to find that we had a poorly implemented endpoint (not indexing a column we queried on, queries that fetched too much).

Very useful for debugging and general system investigation.

For example, our login callback endpoint has this breakdown for a given request (can expand a lot of this):
Screenshot 2026-01-29 at 3 34 27 PM

PR #8251, opened 2026-01-29
[Issue #8269] update next v to 15.5.10

myduong-navapbc:
### Summary

Work for #8269 

This PR upgrades Next.js to v15.5.10 and documents an exception for a newly reported Next.js vulnerability that the scanner indicates is fixed only in a canary prerelease.

### Details

The vulnerability scanner reports GHSA-5f7q-jpqc-wp7h as fixed in 15.6.0-canary.61. I attempted to upgrade to the canary release and align all related Next.js packages, but npm is unable to resolve the dependency tree.

**This occurs because:**

- Canary (prerelease) versions do not satisfy the peer dependency range ^15.x
- Even when next and @next/third-parties are pinned to the same canary version, npm fails with an ERESOLVE error
- As a result, upgrading to the canary release is not viable in this repository.
- Next.js v15.5.10 is the latest stable 15.x release and contains the relevant backported fixes. However, the scanner does not currently recognize the stable fix and continues to recommend the canary version.

### Resolution
- Upgrade to Next.js v15.5.10
- Add a documented scanner ignore for GHSA-5f7q-jpqc-wp7h
- Revisit the ignore once a stable release is recognized by the scanner


frontend/next.config.js
```diff
@@ -109,7 +109,7 @@ const headers = [
   },
   // don't cache user specific pages: saved-opportunities, saved-search-queries
   {
-    source: "/saved:path*",
+    source: "/saved/:path*",
```
mdragon, line 112: Just flagging that this is the only change here so want to double check it was intentionally and/or needed.

myduong-navapbc: It is intentional

The original route pattern was **_invalid_** under newer Next.js versions and caused a runtime error during build/startup:

`TypeError: Can not repeat "path" without a prefix and suffix`

In Next.js 15.5.10, route source patterns are parsed more strictly.
A repeatable parameter like :path* must be separated from the static path by a slash.

PR #8279, opened 2026-01-30
Scale to 6 search nodes (has to be a multiple of 3)

btabaska:
## Summary

Fixes / Work for #ISSUE


infra/api/app-config/prod.tf
```diff
@@ -52,7 +52,7 @@ module "prod_config" {
   # Use the AWS Console to determine the number of availability zones in your region.
   search_data_instance_type      = "or1.medium.search"
   search_data_volume_size        = 20
-  search_data_instance_count     = 3
```
btabaska, line 55: test

Issue #8284, opened 2026-01-30
Create the opportunity and competition in local data, where the competition has a single required form of the SF424B and a conditional form of Lobbying. The competition should accept organizations and individuals.


Bhavna-Ramachandran:
### Summary

Add local test data for an opportunity and competition that supports both organizations and individuals, with SF424B as a required form and Lobbying as a conditional form, to enable validation and testing of [7953](https://github.com/HHS/simpler-grants-gov/issues/7953)

### Acceptance criteria

- [ ]  Local opportunity and competition are created in the local environment for testing (#7953).
- [ ]  Competition supports both organizations and individual applicants.
- [ ]  SF424B is marked as a required form.
- [ ]  Lobbying is marked as a conditional form.

Bhavna-Ramachandran:
**Status:** Pending reviewer confirmation on naming
Per reviewer comment in PR#[8286](https://github.com/HHS/simpler-grants-gov/pull/8286), I've proposed renaming _build_8037_custom_competitions to _build_custom_test_competitions to align with scenario based naming. Awaiting confirmation.

Bhavna-Ramachandran:
**Status:** Pending final review of PR#[8286](https://github.com/HHS/simpler-grants-gov/pull/8286)
Completed updates per reviewer comment.

Bhavna-Ramachandran:
PR#[8286 ](https://github.com/HHS/simpler-grants-gov/pull/8286) for creating local test data for an opportunity and competition that supports both organizations and individuals, with SF424B as a required form and Lobbying as a conditional form - has been reviewed and approved by @chouinar and merged into main. Hence marking this ticket as Done.

As part of this ticket, below data were created-
    - TEST-APPLY-ORG-IND-CT01 -> open to both organizations and individuals
    - TEST-APPLY-ORG-CT01 -> organizations only
    - TEST-APPLY-IND-CT01 -> individuals only

Issue #8037, opened 2026-01-16
Create opportunity/competition for e2e to test the SF424B and conditional Lobbying form


ErinPattisonNava:
### Summary

We want to create an e2e that tests the SF424B form and tests conditional forms.  To do that efficiently, we'd like to create the opportunity with the same guid (and therefore the same URL) across multiple environments.

This competition should have as its only required form the SF424B.  There should be a single conditional form of the Lobbying form. 

For questions on text/details, please work with Bhavna on e2e or Chris to get the information you need.

### Acceptance criteria

- [x] Create the opportunity and competition in local data, where the competition has a single required form of the SF424B and a conditional form of Lobbying.  The competition should accept organizations and individuals.
- [x] Create an opportunity and the corresponding competition in staging having the opportunity referred to by the same guid.
- [x] Create an opportunity and the corresponding competition in dev having the opportunity referred to by the same guid.
- [x] Create an opportunity and the corresponding competition in training having the opportunity referred to by the same guid.

Bhavna-Ramachandran:
Please can you provide more details on the below:

1. User Type: 
- Can this same SF424B + conditional Lobbying form configuration be used for both Individual and Organization users, or is this opportunity intended to be Organization-only?
- If Individuals require a different setup, which mandatory / conditional forms should be used for that scenario?
2. Field level validation / business rules
- For SF424B and the Lobbying form, which fields are:
    - required vs optional
    - validations (format, length, allowed values, etc.)

ErinPattisonNava:
@Bhavna-Ramachandran 

> Please can you provide more details on the below:
> 
> 1. User Type:
> 
> * Can this same SF424B + conditional Lobbying form configuration be used for both Individual and Organization users, or is this opportunity intended to be Organization-only?
> * If Individuals require a different setup, which mandatory / conditional forms should be used for that scenario?
> 

I've clarified the requirements above to specify that the competition should allow for both organizations and individuals, so this opportunity/competition can be used/should be used for both tests.

> 2. Field level validation / business rules
> 
> * For SF424B and the Lobbying form, which fields are:
>   
>   * required vs optional
>   * validations (format, length, allowed values, etc.)

This should be specified in the follow up e2e test, as the form requirements are independent from the competition setup (the competition just specifies which forms it needs, but the rules around fields should be consistent across all uses of that form).

Bhavna-Ramachandran:
[...]

Issue #7953, opened 2026-01-13
Playwright Tests for Apply Happy Path


doug-s-nava:
### Summary

A feature file (see https://github.com/HHS/simpler-grants-gov/pull/7478) has been created for the Apply feature's happy path. This work is to create two Playwright tests based on this feature file:

* happy path for application as an organization
* happy path for application as an individual

This work should entail creating reusable functionality for use in future testing - aspects of the testing script that represent actions that will be taken in other testing scenarios should be split out into their own functions for future reuse.

Depends on https://github.com/HHS/simpler-grants-gov/pull/7411

## Open Questions

- which opportunities can be used both in staging and in a locally seeded environment to accomplish these test cases? In each environment we will need to identify an opportunity with a minimal set of forms, ideally one very simple form, which can be applied to as either an individual or an organization.

### Acceptance criteria

- [x] Playwright test written for `Complete the Apply workflow for an Organization user` test case from feature file
- [ ] Playwright test written for `Complete the Apply workflow for an Individual user` test case from feature file
- [ ] Tests pass consistently against local environment and staging environment in CI

Note: The intent of this happy-path spec is to validate the end-to-end Apply workflow, rather than to exercise complex form logic. Since this path will run in CI, using an opportunity with simple forms (one required form and one conditionally required form) helps reduce flakiness, keeps the tests stable, and makes any failures easier to diagnose. Coverage for more complex forms can be added in separate tests later.

ychoy:
@aoysimmons-GH @Bhavna-Ramachandran will be creating opportunities, so it's not necessarily blocked by #8036 and #8037. 

This work takes priority over Search-related test plans. cc @doug-s-nava

Bhavna-Ramachandran:
**Status:**
- Local test data creation (related task: #[8284](https://github.com/HHS/simpler-grants-gov/issues/8284)) - completed, PR merged to main.
- Playwright script for required form using local data (sub-task: #[8117](https://github.com/HHS/simpler-grants-gov/issues/8117)) -  completed, pending review.

**Next steps** (after #8117 -PR #8431 is reviewed):
- Staging data creation
- Update spec file to work for Staging environment

Bhavna-Ramachandran:
PR #8431 for sub-task #8117 - was reviewed and approved by @doug-s-nava and merged into main and the sub-task has been marked completed.

PR #8286, opened 2026-01-30
Create local data requirements in 8037 for testing 7953

Bhavna-Ramachandran:
## Summary

Work for : Task #[8284](https://github.com/HHS/simpler-grants-gov/issues/8284) 

## Changes proposed
- Added _build_8037_custom_competitions() to create local test competitions for #8037, required for scripting work in #7953.
- The competitions include:
    - TEST-BR-8037-IOU-OT01 -> open to both organizations and individuals
    - TEST-BR-8037-OU-OT01 -> organizations only
    - TEST-BR-8037-IT-OT01 -> individuals only
- Forms attached to each competition:
    - SF424B -> required
    - SFLLL_2_0 -> conditional
- Log statements include local URLs for quick verification in the browser.

## Context for reviewers
- This change is for #8037 local test data setup only and does not affect deployed environments
- Enables testing and validation for #7953 scripting work.
- Ensures each competition scenario matches the applicant types and forms required.

## Validation steps
- Verify the three new competitions exist in local DB:
    - TEST-BR-8037-IOU-OT01 -> both applicant types
    - TEST-BR-8037-OU-OT01 -> organizations only
    - TEST-BR-8037-IT-OT01 -> individuals only
- Check that SF424B is required and SFLLL_2_0 is conditional in each competition.
- Confirm log output contains URLs for each competition that can be opened in local environment.

### DIFF
--- a/api/tests/lib/seed_local_db.py
+++ b/api/tests/lib/seed_local_db.py
@@ -1,9 +1,3 @@
-def fetch_competition(db_session, competition_id):
-    return db_session.scalar(
-        select(Competition).where(Competition.competition_id == competition_id)
-    )
-
-
 import dataclasses
 import logging
 import uuid
@@ -27,6 +21,13 @@ from tests.lib.seed_agencies import _build_agencies
 from tests.lib.seed_data_utils import CompetitionContainer
 from tests.lib.seed_e2e import _build_users_and_tokens
 from tests.lib.seed_orgs_and_users import _build_organizations_and_users, seed_internal_admin
+
+
+def fetch_competition(db_session, competition_id):
+    return db_session.scalar(
+        select(Competition).where(Competition.competition_id == competition_id)
+    )
+
 
 logger = logging.getLogger(__name__)
 


api/tests/lib/seed_local_db.py
```diff
@@ -299,11 +299,62 @@ def _build_competition_with_all_forms(forms: list[Form]) -> Competition:
     return competition
 
 
+# Build custom competitions 8037 for testing 7953
+def _build_8037_custom_competitions(forms: dict[str, Form]) -> None:
```
chouinar: I'd suggest naming this more for the use case rather than the ticket number - we won't otherwise know what it is for in the future. The logging reads like you're creating 8037 separate opportunities, it wouldn't be clear as-is that this is for a ticket. I'd say we should just name these based on the scenarios like the other data we create.

Not sure how we'd describe these scenarios though.

chouinar (commented):
You'll want to run `make format lint` to fix / tell you any format/lint issues.

api/tests/lib/seed_local_db.py
Bhavna-Ramachandran: Acknowledged! Since the purpose is to seed data for test automation, I’m thinking of renaming it to either _build_custom_test_competitions or _build_custom_automation_competitions. Do you have a preference, or is there another naming pattern you’d recommend? Otherwise, I can proceed with _build_custom_test_competitions.

api/tests/lib/seed_local_db.py
```diff
@@ -299,11 +299,81 @@ def _build_competition_with_all_forms(forms: list[Form]) -> Competition:
...
+        opportunity__opportunity_title="TEST-APPLY-ORG-IND-OT01",
+        opportunity__opportunity_number="TEST-APPLY-ORG-IND-ON01",
+        competition_title="TEST-APPLY-ORG-IND-CT01",
+        competition_forms=[],
+        open_to_applicants=[
+            CompetitionOpenToApplicant.ORGANIZATION,
+            CompetitionOpenToApplicant.INDIVIDUAL,
+        ],
+        with_instruction=True,
+    )
```
chouinar: For any of this to work with static IDs, you have to first check if the opportunity ID already exists and only create it if it doesn't.

The seed script has to be capable of safely running multiple times in a row as there are several use cases that depend on it.

To keep this a bit simpler, I'd suggest adding a function like:

```py
def does_opportunity_exist(db_session: db.Session, opportunity_id: uuid.UUID) -> bool:
    opportunity = db_session.get(Opportunity, opportunity_id)
    return opportunity is not None
```

Then for each of these, you can do:
```py
if does_opportunity_exist(db_session, opp_id):
    x = factories.CompetitionFactory.create( ... )
```

Bhavna-Ramachandran: Thank you will work on this.

api/tests/lib/seed_local_db.py
```diff
@@ -299,11 +299,129 @@ def _build_competition_with_all_forms(forms: list[Form]) -> Competition:
...
 
 
+# Build custom competitions 8037 for testing 7953
+def does_competition_form_exist(db_session, competition_id, form_id):
+    return (
+        db_session.query(CompetitionForm)
+        .filter_by(competition_id=competition_id, form_id=form_id)
+        .first()
+        is not None
+    )
```
chouinar, line 319: This works, but is the older legacy way of querying. Just prefer the newer way that uses execute/scalar.

```suggestion
    return (
        db_session.scalar(select(CompetitionForm)
        .where(competition_id=competition_id, form_id=form_id))
        is not None
    )
```

api/tests/lib/seed_local_db.py
```diff
@@ -299,11 +299,129 @@ def _build_competition_with_all_forms(forms: list[Form]) -> Competition:
...
+                CompetitionOpenToApplicant.INDIVIDUAL,
+            ],
+            with_instruction=True,
+        )
+    else:
+        both_competition = (
+            db_session.query(Competition)
+            .filter_by(competition_id=uuid_map["TEST-APPLY-ORG-IND-CT01"])
+            .one()
+        )
```
chouinar: We could also make the function that fetches the opportunity to instead fetch the competition and do something like:

```py
both_competition = fetch_competition(...)
if not both_competition:
    both_competition = factories.CompetitionFactory.create(...)
```

Would mean we don't need the if-else / makes it a bit shorter.

api/tests/lib/seed_local_db.py
Bhavna-Ramachandran: Thank you. I will update that to use use db_session.scalar(select(...))

api/tests/lib/seed_local_db.py
Bhavna-Ramachandran: Thank you. I will update this per your comment.

api/tests/lib/seed_local_db.py
Bhavna-Ramachandran: Completed updates per reviewer comment.

api/tests/lib/seed_local_db.py
Bhavna-Ramachandran: Completed updates per reviewer comment.

api/tests/lib/seed_local_db.py
Bhavna-Ramachandran: Completed updates per reviewer comment.

PR #8293, opened 2026-02-01
HOTFIX: 20260131 banner

andycochran:
## Summary

This is a hotfix to add a temporary banner for Jan 31, 2026.

## Changes proposed

* Hardcode a message in the `Layout` that displays below the site header

## Context for reviewers

As this is temporary, the content is hardcoded and does not make use of i18n. This change is sandwiched in a comment so the code can easily be removed.

## Validation steps

Load the app and you should see the following message on a yellow background.

image


mdragon (approved):
Approved

PR #8345, opened 2026-02-03
Add privileges and roles to create and publish opportunity

### LINKED ISSUE #7711: AuthZ/RBAC Enforcement
### Summary

**Background/Context**

**Resources**

**USER STORIES** 

**Backend Example**

**Frontend Example**

**TECHNICAL NOTES** 

### Acceptance criteria
- [x] Define the agency-user-roles and the list of permissions for each role.
- [x] Add/insert the list of agency-user-roles and the list of permissions for each role into the DB.

Role type association: Agency

Dao-REI:
## Summary

Fixes #7711

## Changes proposed

Four privileges added for view, create, update and publish opportunity.
Two roles added, one to edit and one to publish opportunities, both for role type = Agency.

## Context for reviewers

Opportunity editor is given view, create and update privileges
Opportunity publisher is given view and publish privileges
AC #3 (temporary approaches to assign roles to users) in #7711 will be moved to a separate ticket and is not included in this PR.

## Validation steps

1. Review the code changes.
2. Locally, run make make db-migrate-create MIGRATE_MSG="Add privileges and roles for opportunity publishing" and make db-migrate (or make remake-backend) to apply changes to the DB. Then verify that the records were added to these tables:
- select * from api.lk_privilege;
- select * from api.role;

### DIFF
--- a/api/src/constants/static_role_values.py
+++ b/api/src/constants/static_role_values.py
@@ -150,6 +150,8 @@ OPPORTUNITY_PUBLISHER = Role(
         [
             Privilege.VIEW_OPPORTUNITY,
             Privilege.PUBLISH_OPPORTUNITY,
+            Privilege.CREATE_OPPORTUNITY,
+            Privilege.UPDATE_OPPORTUNITY,
         ],
     ),
     link_role_types=[LinkRoleRoleType(role_id=OPPORTUNITY_PUBLISHER_ID, role_type=RoleType.AGENCY)],


api/src/constants/static_role_values.py
```diff
@@ -120,6 +120,41 @@ def get_link_privileges(role_id: uuid.UUID, privilege: list[Privilege]) -> list[
...
+    role_id=OPPORTUNITY_PUBLISHER_ID,
+    role_name="Opportunity Publisher",
+    is_core=True,
+    link_privileges=get_link_privileges(
+        OPPORTUNITY_PUBLISHER_ID,
+        [
+            Privilege.VIEW_OPPORTUNITY,
+            Privilege.PUBLISH_OPPORTUNITY,
+        ],
+    ),
```
chouinar, line 156: Should a publisher have create/update as well?

Either approach is fine, if someone has multiple roles, we just merge their privileges together for a given entity. Just whether we expect a user to need both roles to go through the whole process, or if they should be able to do it with just one - which is a UX question.

Dao-REI: Referencing [Notes on Publishing Opportunities](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2638381057/Notes+on+publishing+opportunities), it says "Changes may require additional approvals depending on agency policy." That implies that there is an approval process for all updates. Thus, I assumed the Publisher should not have create/update privileges, less their changes are published without getting reviewed. 

@crystabelrangel Would you be able to make a determination on this?

chouinar: Anything with approvals will be separate - that's getting into the workflow management work I've been doing.

Publish won't quite be publish, that endpoint will just put it on the workflow to get approved/actually published.

This isn't a blocker - feel free to circle back to this later. Main reason I was recommending having one role with all the privileges is that it'll be more work to manually setup users otherwise.

crystabelrangel: @Dao-REI you're correct that these will likely include an approval process, but we will get more information as we learn more about roles and permissions. 

For this initial work, it makes sense to keep it to a single role to go through the publishing process, and then we can work on the approval process.

Dao-REI: @chouinar Create and Update privileges have been added to Opportunity Publisher. Please review.

Issue #8356, opened 2026-02-03
Create Framework for Opportunity Listing Page


glenblosser-nava:
### Summary

While we don't yet have the endpoints to support fetching the actual user data, we can create the table using placeholder data to reduce the work that needs to be done later. This ticket will only cover the basic existence of the Opportunity List table.

### Acceptance criteria

- [ ] User can navigate to /opportunities
- [ ] Page contains a table of 0-5 example opportunities
- [ ] Each row displays the Agency, Title, Status, and Action Buttons for Edit, Copy, and Delete.
- [ ] The Action Buttons should be disabled
- [ ] Clicking on the Title should take you to that Opportunity's page

Issue #8009, opened 2026-01-15
List of Opportunities Frontend Page (Functionality)


brianlee-nava:
### Summary

**Background and Context:** 
Resources: [Tech Spec](https://navasage.atlassian.net/wiki/x/CQA-kw
) 

- We need to build a new page in Simpler Grants Management. 
- The page will display all opportunities for one of the user's agencies
- This page will be similar to the My Applications page. 
- The columns to display are TBD but should include Title and Status (e.g. draft, published, ended/funded, archived, etc.). 
- Action buttons are TBD but should include copy and edit for the applicable opportunity. (For example, the edit button should only be enabled for drafts/unpublished opportunities and only for users with permission to edit that record). 

**User Story**
As a Grantor user
I want to view and manage opportunities by agency in Simpler Grants Management
So that I can easily find, review, and take action on opportunities I am responsible for

### Acceptance criteria

Background:
    Given I am an authenticated Grantor user
    And I have access to one or more agencies

Scenario: View list of agencies
  Given I navigate to the Agency Opportunities page
  Then I see a list of agencies that I belong to
  And at least one agency is selected by default

Scenario: View opportunities for a selected agency
  Given I am on the Agency Opportunities page
  And I have selected an agency
  When the opportunities load
  Then I see a table of opportunities for the selected agency
  And each opportunity displays a Title column
  And each opportunity displays a Status column

  Scenario: Opportunity status values are displayed
    Given I am viewing the opportunities table
    Then each opportunity shows one of the following statuses:
      | draft |
      | published |
      | ended |
      | funded |
      | archived |

  Scenario: Edit action is enabled for editable opportunities
    Given I am viewing the opportunities table
    And I have permission to edit an opportunity
    When an opportunity has a status of "draft"
    Then the Edit action is enabled for that opportunity

  Scenario: Edit action is disabled for non-editable opportunities
    Given I am viewing the opportunities table
    When an opportunity has a status other than "draft"
    Then the Edit action is disabled for that opportunity

  Scenario: Edit action is hidden or disabled without permission
    Given I am viewing the opportunities table
    And I do not have permission to edit an opportunity
    Then the Edit action is hidden or disabled for that opportunity

  Scenario: Copy action is available for applicable opportunities
    Given I am viewing the opportunities table
    When an opportunity supports copying
    Then the Copy action is available for that opportunity

- [ ] Hidden behind feature flag
- [ ] Call GET API to retrieve list of opportunities for user
[...]

PR #8359, opened 2026-02-03
[Issue #8356] List of opportunities page

glenblosser-nava:
## Summary

Work for #8009 
Work for #8356 

## Changes proposed

Basic page for listing opportunities created. Does not have full functionality, but will be able to link into other tickets to unblock work.

## Context for reviewers

Fetches 5 random opportunities, rather than opportunities linked to a specific user. Action buttons are placeholders as their functionality is not implemented yet.
Adds a feature flag to control access to the page. It defaults to allowing the page in dev but not in prod

## Validation steps

- [ ] Manually navigate to /opportunities
- [ ] Page contains 5 opportunities in a table
- [ ] Clicking the title takes you to the opportunity page


frontend/src/app/[locale]/(base)/opportunities/page.test.tsx
```diff
@@ -0,0 +1,159 @@
...
+jest.mock("react", () => ({
+  ...jest.requireActual<typeof import("react")>("react"),
+  use: jest.fn(() => ({
+    locale: "en",
+  })),
+}));
+
+jest.mock("next-intl/server", () => ({
+  getTranslations: () => identity,
+  setRequestLocale: identity,
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunities/page.test.tsx]: this can be left out now

frontend/src/app/[locale]/(base)/opportunities/page.test.tsx
```diff
@@ -0,0 +1,159 @@
...
+}));
+
+jest.mock("next-intl", () => ({
+  useTranslations: () => useTranslationsMock(),
+}));
+
+const opportunities = jest.fn().mockResolvedValue({ data: [] });
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunities/page.test.tsx]: this would be easier to track if it was called something like "mockSearchForOpportunities"

frontend/src/app/[locale]/(base)/opportunities/page.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+      <TableWithResponsiveHeader
+        headerContent={headerTitles}
+        tableRowData={transformTableRowData(userOpportunities, t)}
+      />
+    </div>
+  );
+};
+
+export default async function Opportunities(_props: LocalizedPageProps) {
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunities/page.tsx]: For ease of searching, I'd name this something like "OpportunitiesListPage". "Opportunities" is thrown around a lot in this project, good to keep things somewhat unique

frontend/src/app/[locale]/(base)/opportunities/page.test.tsx
Comment (glenblosser-nava) [reply frontend/src/app/[locale]/(base)/opportunities/page.test.tsx]: removed

frontend/src/app/[locale]/(base)/opportunities/page.test.tsx
Comment (glenblosser-nava) [reply frontend/src/app/[locale]/(base)/opportunities/page.test.tsx]: renamed

frontend/src/app/[locale]/(base)/opportunities/page.tsx
Comment (glenblosser-nava) [reply frontend/src/app/[locale]/(base)/opportunities/page.tsx]: renamed

frontend/.env.development
```diff
@@ -44,3 +44,5 @@ SESSION_SECRET=extraSecretSessionSecretValueSssh
 E2E_USER_AUTH_TOKEN=  # used for spoofing logins in e2e need to sync with API via seed
 
 PLAYWRIGHT_TARGET_ENV= # used to determine which environment to target in playwright tests. defaults to "local", "staging" also supported
+
+FEATURE_OPPORTUNITIES_LIST_OFF=false # enables opportunities list page in lower environment
```
doug-s-nava: I see what you're getting at here, but the better way to do this is to set the feature flag values in SSM for the lower environments. This value would be overridden by the SSM value in any case.

One other point is that builds for all deployed environments use the .env.production file rather than .env.development. The development env file is only used when running `next dev`, which should only happen locally

frontend/src/app/[locale]/(base)/opportunities/page.tsx
```diff
@@ -0,0 +1,140 @@
...
+    </div>
+  );
+};
+
+async function OpportunitiesListPage(_props: OpportunitiesListProps) {
+  const searchParams = convertSearchParamsToProperTypes({});
+
+  let userOpportunities: BaseOpportunity[];
+  try {
+    userOpportunities = (await searchForOpportunities(searchParams)).data
```
Comment (chouinar) [inline frontend/src/app/[locale]/(base)/opportunities/page.tsx:115]: A warning, the endpoint that'll actually get built will be quite a bit different than search. This will need to be reworked.

I'd suggest trying to build backend before frontend where possible.

doug-s-nava:
make sure the failing e2e tests is a blip before merging (run locally and make sure you can get it to pass)

Issue #8263, opened 2026-01-30
Build unittest utilities -- feature flags


ErinPattisonNava:
### Summary

We want to build some unittest utilities around feature flags and modify enough of the existing examples so that they because the established way to move forward.

### Acceptance criteria

- [x] Feature flag utilities are built
- [x] Draft PR with 3 examples is circulated
- [x] Full PR will as many examples that can be converted in the timeboxed time is committed.

myduong-navapbc:
Per discussion, this PR no longer introduces or commits to feature-flag–specific testing behavior. The underlying infrastructure is generic and applies to any wrapper or import-time side effect used by App Router pages.

I've decided to close this PR and the task that belongs to it; opted to not introduce code that would not be used immediately and pollute the codebase.

PR #8361, opened 2026-02-03
[Issue #8263] frontend testing utilities

myduong-navapbc:
## Summary

Work for #8263  

---

Per discussion, this PR no longer introduces or commits to feature-flag–specific testing behavior. The underlying infrastructure is generic and applies to any wrapper or import-time side effect used by App Router pages.

---

### `PageModule`

Page tests need to interact with App Router pages in a limited, intentional way:

- call the page’s default export
- optionally call generateMetadata
- avoid coupling to unrelated exports or implementation details

The `PageModule` type formalizes this boundary.

```
type PageModule = {
  default: PageFn;
  generateMetadata?: (args: { params: Promise<Params> }) => Promise<Metadata>;
};
```

This allows tests to:

- focus on inputs (params) and outputs (JSX, metadata, redirects)
- avoid re-testing framework behavior
- remain stable as page internals evolve

---

### `loadIsolatedPageModule`

`loadIsolatedPageModule` is the core primitive introduced in this PR.

It combines two responsibilities that always occur together in App Router page tests:

- Registering mocks
- Loading the page module in an isolated module graph

This guarantees the correct ordering:

register mocks
-> isolate module graph
-> evaluate page module
-> execute test

Without this, tests become dependent on:

- implicit import order
- Jest’s global module cache
- whether another test imported the page earlier in the run

This utility should removes that uncertainty.

### DIFF
--- a/frontend/src/test/fixtures/sessionUser.ts
+++ b/frontend/src/test/fixtures/sessionUser.ts
@@ -3,16 +3,32 @@ export type GetSession = typeof import("src/services/auth/session").getSession;
 type SessionValue = Awaited<ReturnType<GetSession>>;
 export type SessionUser = NonNullable<SessionValue>;
 
-export function buildSessionUser(
-  overrides?: Partial<SessionUser>,
-): SessionUser {
-  return {
+export const SESSION_USERS = {
+  default: {
     user_id: "user-1",
     email: "user-1@example.com",
     first_name: "Test",
     last_name: "User",
     token: "fake-token",
     session_duration_minutes: 0,
+  } satisfies SessionUser,
+
+  // Example of another user
+  otherUser: {
+    user_id: "user-2",
+    email: "user-2@example.com",
+    first_name: "Other",
+    last_name: "User",
+    token: "fake-token-2",
+    session_duration_minutes: 0,
+  } satisfies SessionUser,
+} as const;
+
+export function buildSessionUser(
+  overrides?: Partial<SessionUser>,
+): SessionUser {
+  return {
+    ...SESSION_USERS.default,
     ...overrides,
   };
 }


documentation/frontend/featureFlags.md
```diff
@@ -162,7 +162,7 @@ test file that is affected
 
 ```typescript
```
myduong-navapbc, line 163: Moved `jsdomNodeEnvironment.ts` to `src/test/jest-environments/` to keep custom Jest environments grouped with other test-only tooling. 

Functionality is unchanged; only the file location and the @jest-environment path reference were updated.

frontend/src/app/[locale]/(base)/organizations/[id]/manage-users/legacy/page.test.tsx
```diff
@@ -1,110 +1,63 @@
 import { render } from "@testing-library/react";
-import InviteLegacyUsersPage from "src/app/[locale]/(base)/organizations/[id]/manage-users/legacy/page";
-import { LocalizedPageProps } from "src/types/intl";
-import { FeatureFlaggedPageWrapper } from "src/types/uiTypes";
-import { UserDetail } from "src/types/userTypes";
+import {
+  buildSessionUser,
+  type GetSession,
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/organizations/[id]/manage-users/legacy/page.test.tsx]: is the type assertion necessary?

frontend/src/app/[locale]/(base)/organizations/[id]/manage-users/legacy/page.test.tsx
```diff
@@ -1,110 +1,63 @@
...
-});
+    const sessionUser = buildSessionUser();
+    const authenticationMock: jest.MockedFunction<GetSession> = jest
+      .fn()
+      .mockResolvedValue(sessionUser);
 
-jest.mock("src/services/auth/session", () => ({
-  getSession: () => authentication() as Promise<UserDetail>,
-}));
+    jest.doMock("next-intl/server", () => ({
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/organizations/[id]/manage-users/legacy/page.test.tsx]: why are imports being mocked within the test body?

frontend/src/app/[locale]/(base)/organizations/[id]/manage-users/legacy/page.test.tsx
```diff
@@ -1,110 +1,63 @@
...
-        WrappedComponent: FunctionComponent<LocalizedPageProps>,
-        _featureFlagName: "",
-        _onEnabled: () => void,
-      ) =>
-        (props: { params: Promise<{ locale: string }> }) =>
-          WrappedComponent(props) as unknown,
-    );
-  });
+    const { pageModule, featureFlagHarness } =
+      loadPageWithFeatureFlagHarness<PageModule>(
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/organizations/[id]/manage-users/legacy/page.test.tsx]: can we assume that `loadPageWithFeatureFlagHarness` should always load the page module with the `PageModule` type? If so can we avoid the generics?

frontend/src/test/helpers/loadPageWithFeatureFlagHarness.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Loads a module in isolation with the feature flag harness applied.
+ */
+export function loadPageWithFeatureFlagHarness<PageModule>(
+  modulePath: string,
+  initialMode: FeatureFlagMode,
+): {
+  pageModule: PageModule;
+  featureFlagHarness: ReturnType<typeof createFeatureFlagHarness>;
```
doug-s-nava: why not 
```suggestion
  featureFlagHarness: FeatureFlagHarness
```

frontend/src/app/[locale]/(base)/organizations/[id]/manage-users/legacy/page.test.tsx
```diff
@@ -1,110 +1,63 @@
...
-  useTranslations: () => (key: string) => key,
-}));
+  it("redirects to maintenance if manageUsersOff is enabled", async () => {
+    const redirectMock = jest.fn<void, [string]>();
 
-const authentication = jest.fn().mockResolvedValue({
-  token: "fake-token",
-  user_id: "user-1",
-});
+    const sessionUser = buildSessionUser();
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/organizations/[id]/manage-users/legacy/page.test.tsx]: two things:
* what's the advantage of this versus mocking the `<AuthenticationGate`? This is more flexible, but that is more clearly isolating specific page behavior...
* why not expand the user mocking tooling to include the mocking itself.
```
const mockSessionUser = () => {
  const sessionUser = buildSessionUser();
  const authenticationMock: jest.MockedFunction<GetSession> = jest
      .fn()
      .mockResolvedValue(sessionUser);
  return authenticationMock
}
...
const authenticationMock = mockSessionUser()
jest.doMock("src/services/auth/session", () => ({
  getSession: () => authenticationMock(),
}));
```

frontend/src/app/[locale]/(base)/organizations/[id]/manage-users/page.test.tsx
```diff
@@ -130,13 +87,66 @@ describe("manage-users page", () => {
...
+        getOrganizationUsersMock(...args),
+      getOrganizationRoles: (...args: Parameters<GetOrgRolesFn>) =>
+        getOrganizationRolesMock(...args),
+    }));
+
+    const { pageModule, featureFlagHarness } =
+      loadPageModuleWithMocks("flagDisabled");
+
+    // Assert that the page is wrapped by withFeatureFlag without coupling to a specific flag name.
+    expectAnyFeatureFlagWiring(featureFlagHarness);
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/organizations/[id]/manage-users/page.test.tsx]: seems like a lot of work to do to not even assert anything particularly meaningful here. My hope is we can build a system that will not require a lot of complexity for cases where we do not need to dig in too much to the feature flag behavior

frontend/src/test/fixtures/sessionUser.ts
```diff
@@ -0,0 +1,18 @@
+export type GetSession = typeof import("src/services/auth/session").getSession;
+
+type SessionValue = Awaited<ReturnType<GetSession>>;
+export type SessionUser = NonNullable<SessionValue>;
+
+export function buildSessionUser(
+  overrides?: Partial<SessionUser>,
+): SessionUser {
+  return {
+    user_id: "user-1",
```
doug-s-nava, line 8: can we store this user as a fixture so that everything can be referencing the same data?

frontend/src/test/harness/featureFlagHarness.ts
```diff
@@ -0,0 +1,108 @@
...
+};
+
+/**
+ * Test harness for withFeatureFlag.
+ *
+ * - Simulates flag enabled/disabled branches
+ * - Captures how the wrapper is wired
+ * - Does not read cookies, env vars, or Next internals
+ */
+export function createFeatureFlagHarness(
```
doug-s-nava: quite conflicted here:
* this is a very clever use of closures to get a class like behavior without using classes which is :chef's kiss:
* it's so confusing

frontend/src/app/[locale]/(base)/organizations/[id]/manage-users/legacy/page.test.tsx
Comment (myduong-navapbc) [reply frontend/src/app/[locale]/(base)/organizations/[id]/manage-users/legacy/page.test.tsx]: No, not absolutely necessary- BUT, it is intentional.
In this particular instance you get the full type safety without setting `AS TYPE.`

```
const authenticationMock: jest.MockedFunction<GetSession> = jest
  .fn()
  .mockResolvedValue(sessionUser);
```

I ran into an linting / TS issue where userSession, the error was "Conversion of type { token: string; user_id: string } to type UserDetail may be a mistake…” and to get it to pass I had to force set it to UserDetail.

So, in short, what i am doing is binding the mock `getSession` to the real `getSession` signature and there’s no drift in typecasting to avoid the `AS TYPE` or worse `AS UNKNOWN AS TYPE`.

Let me know if this is a good enough reason to keep it, if not, I will remove it.

frontend/src/app/[locale]/(base)/organizations/[id]/manage-users/legacy/page.test.tsx
Comment (myduong-navapbc) [reply frontend/src/app/[locale]/(base)/organizations/[id]/manage-users/legacy/page.test.tsx]: Mocking `getSession` keeps the test closer to how the page actually decides what to render, and it lets us exercise unauthenticated cases by returning null. Mocking `AuthenticationGate` is a good tool when the page delegates auth entirely to the gate component. In this file, we’re validating the page’s own session-dependent behavior, so mocking `getSession` is the better fit.

_Why not expand the session tooling to include the mocking itself?_

I like that idea and agree it would reduce repetition (e.g. a helper like `createGetSessionMock()` / `registerMockGetSession()`). I kept `buildSessionUser()` as a pure fixture in this PR (no Jest side effects) and scoped this change to feature flag testing. I’d prefer to introduce Jest-aware auth helpers in a follow-up “auth testing utilities” ticket/PR so we can review that direction separately.

frontend/src/app/[locale]/(base)/organizations/[id]/manage-users/legacy/page.test.tsx
Comment (myduong-navapbc) [reply frontend/src/app/[locale]/(base)/organizations/[id]/manage-users/legacy/page.test.tsx]: This is something I learned digging into Next App Router behavior. The answer is a little long-winded.

In App Router, a `page.tsx` isn’t just a React component that gets rendered later. it’s a module that gets evaluated when it’s imported. That means any top-level code runs immediately, including wrapper application like:

```
export default withFeatureFlag(
  PageComponent,
  "manageUsersOff",
  () => redirect("/maintenance"),
);
```


That `withFeatureFlag(...)` call executes at import time, not when the component renders.

So if we rely on top-of-file `jest.mock(...)`, it only works reliably if:

- the page module hasn’t been imported yet (directly or indirectly)
- no other test imported it earlier in the run
- we don’t hit Jest’s module caching in a way that “locks in” the real module before our mocks apply

In our suite we hit this a lot because pages pull in dependencies indirectly (examples):

- src/components/manageUsers/ManageUsersPageContent
- src/components/user/AuthorizationGate
- src/services/fetch/fetchers/organizationsFetcher
- next-intl/server
- next/navigation

If the page is imported earlier (even indirectly), the wrapper is evaluated before our mocks are registered - and then later mocks either don’t apply or apply too late.

That’s why we use `jest.doMock` inside the test body + load the page within `jest.isolateModules()`: it guarantees the order we need:

1. register mocks
2. import the page module (fresh graph)
3. the wrapper code runs using the mocked dependencies

This is a Next/App Router constraint, not a stylistic preference.

frontend/src/app/[locale]/(base)/organizations/[id]/manage-users/legacy/page.test.tsx
Comment (myduong-navapbc) [reply frontend/src/app/[locale]/(base)/organizations/[id]/manage-users/legacy/page.test.tsx]: Absolutely! Love it! agreed. Most pages only need the default export, so we can make `loadPageWithFeatureFlagHarness` assume the common `PageModule` shape by default and avoid generics at the call site. We’ll keep an explicit generic only for cases like metadata tests where we need additional exports.

frontend/src/app/[locale]/(base)/organizations/[id]/manage-users/page.test.tsx
Comment (myduong-navapbc) [reply frontend/src/app/[locale]/(base)/organizations/[id]/manage-users/page.test.tsx]: Agreed. I removed this entirely

frontend/src/test/harness/featureFlagHarness.ts
myduong-navapbc: I think the complexity is not from closures themselves but from the need to coordinate import-time behavior. 

Closures are just the cleanest way to keep state local - TOTALLY open to ideas here. This is actually the one part of the entire PR that I got really conflicted with as well.

frontend/src/test/helpers/loadPageWithFeatureFlagHarness.ts
myduong-navapbc: That is a really good catch. That was me dropping a bunch of console and debuggers to make sure I got the right types when I wrote it.

It should be typed as `FeatureFlagHarness`

myduong-navapbc:
> some general thoughts:
> 
> * great stuff in the documentation - for any of this that we end up committing to, the docs should move into permanent documentation so it doesn't just live in this ticket

Agreed. if we commit to this approach I will updaate the test README.


> * re: `__mocks__` - I'd like to either get rid of this and standardize what's there now (just focus-trap and zipjs) to use a more functional and less platform dependent mocking mechanism, the way we generally mock dependencies elsewhere, or make it clear when and how we expect global mocks to be used. I personally have never written one, as it never occurs to me, but if other folks on the team like working with these let's make sure everyone understands how and when to use them

I’m aligned with this. The only reason __mocks__ exists today is for a small number of third-party libraries (eg, focus-trap, zipjs) where Jest’s resolution behavior makes inline mocking awkward.

For application code, my preference is the same as yours: functional, explicit mocks declared in the test or test utilities.

The longer I work in NEXT, specifically testing, I would wholeheartedly push away from  global mocks. 

> * re: new `src/test` directory structure, `utils/testing` should migrate over there, can we make a ticket for that?

I didn’t want to move it prematurely before we agreed on the pattern.


> * other than the session mocking, which I think is good, all of this seems to be a workaround for testing pages that are wrapped in `withFeatureFlag`, which nobody likes as far as I can tell. What if instead of all of this we created a new way to override page behavior based on feature flag values that would be easier to test? (what about switch cases within the page component render? what about redirects from the middleware?) Or have you been down that road and figured this is as close to a good system as we're going to get?

This is one comment, I asked myself while working on the task, and it is a totally fair question- and I did think about alternative designs (middleware redirects, inline conditionals, switch statements inside the page).

The conclusion I came to is - given the current architecture, withFeatureFlag is already widely used and not something I can easily unwind without touching many pages. This task isn’t meant to defend the current design forever. As the project exists today, these utilities are meant to make testing it tractable and predictable (as i possibly could).

I do agree that:

Middleware-level redirects are often easier to test

Inline conditional rendering is simpler in Next than HOCs / Wrappers



If we decide to move in that direction, I think that should be a separate architectural change. 

The harness here is intentionally isolated so that if we later replace withFeatureFlag, the number of test changes would be minimal.



> * a lot of the complexity here seems to be about navigating the timing and order of operations around when things get mocked. This seems to be necessary because we want to mock out behavior that is happening during the render (?) because of the way we are wrapping the default components with `withFeatureFlag`. If the best way to work around that complexity involves moving all of our mocking behavior into the body of each test that gives me a feeling that we need to go back to the drawing board and implement our feature flag management differently. My thinking at this point is that `withFeatureFlag` is a good implementation... for a vanilla react app. Next complicates the matter of testing a non default (non wrapped) export of the page, and I think at this point we should maybe invest time in doing things in a way that places nice with Next rather than jumping through hoops to make this work

Its one of those, the longer I work in Next, the more I am starting to understand Next’s import time and execution orders and how Next evals modules and page modules. BUT, I agree that the complexity around mocking timing is not great or ideal by any means.

What I constantly saw was:

reset modules & inline a custom withFeatureFlag mock

I dont think I could ever really eliminate that complexity but I could contain it in one place and I tried to give the us a way to declare flag on / flag off. But to me i wanted to avoid retesting wrapper internals.

AND I don’t disagree that there are hoops here. The tradeoff I was (hoping for) optimizing for was:

- keep page tests readable
- avoid global state
- avoid asserting on wrapper internals
- make behavior deterministic

The harness approach felt like the smallest surface area change that accomplished that, without requiring a rewrite of how feature flags are applied across the app. 

I’m very open to continuing on a more Next-native feature flag approach. I just ran out of time for the scope of this task.

frontend/src/app/[locale]/(base)/organizations/[id]/manage-users/page.test.tsx
```diff
@@ -1,16 +1,24 @@
...
 
-import React, { JSX } from "react";
+import React, { type JSX } from "react";
 
 import "@testing-library/jest-dom";
 
-import Page, {
-  generateMetadata,
-} from "src/app/[locale]/(base)/organizations/[id]/manage-users/page";
+import { expectFeatureFlagWiring } from "src/test/harness/featureFlagHarness";
```
Comment (ErinPattisonNava) [inline frontend/src/app/[locale]/(base)/organizations/[id]/manage-users/page.test.tsx]: Alas, I believe Hayes has a PR up right now to remove the manageUsersOff flag.  If this code isn't needed anymore, what's the best way to preserved the examples?

frontend/src/app/[locale]/(base)/organizations/[id]/manage-users/legacy/page.test.tsx
Comment (ErinPattisonNava) [reply frontend/src/app/[locale]/(base)/organizations/[id]/manage-users/legacy/page.test.tsx]: I had a much harder time reading these tests because of the distance between the descriptive test (what we should be testing) and the page specific checks (what we are testing) with all the mocks in between. Is there another pattern we could use that would improve readability?

frontend/src/test/types/featureFlagTypes.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Feature flag wrapper behavior to simulate in tests.
+ *
+ * - flagDisabled: wrapper renders the wrapped component
+ * - flagEnabled: wrapper executes the onEnabled callback
+ */
+export type FeatureFlagMode = "flagEnabled" | "flagDisabled";
```
ErinPattisonNava: Does this lock us in to a polarity about how we design flags going forward?  I must admit, when I saw `flagEnabled` I didn't think it meant we executed a callback but in how we were interpreting the flag behaviour.  If you want this to be more functionality oriented is this the right name?

frontend/src/test/fixtures/sessionUser.ts
myduong-navapbc: added an example

frontend/src/app/[locale]/(base)/organizations/[id]/manage-users/page.test.tsx
Comment (myduong-navapbc) [reply frontend/src/app/[locale]/(base)/organizations/[id]/manage-users/page.test.tsx]: i removed the feature flag harness

frontend/src/test/types/featureFlagTypes.ts
myduong-navapbc: i removed the feature flag harness

myduong-navapbc:
I've decided to close this PR and the task that belongs to it; opted to not introduce code that would not be used immediately and pollute the codebase.

Issue #8071, opened 2026-01-20
Create a client for calling SQS


chouinar:
### Summary

Similar to some of our other clients, we want a client that makes interacting with SQS simple. It should be using the [boto SQS client](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs.html) and have functions for interacting with:

* `receive_message` - we should be able to pass in:
  * How many messages (default 10)
  * Wait time (default 10sec)
  * Visibility timeout (default 5 min)
* `delete_message_batch` - which should take in a list of receipt handles, and should return an easy to parse response (giving us a mapping of receipt handles to success/failure status codes).

For testing this, we should be able to interact with a mock SQS queue in our unit tests using [moto](https://docs.getmoto.org/en/latest/docs/getting_started.html) - a library we already use for mocking s3. We should have a fixture for an sqs queue that we can reference, similar to the s3 bucket fixture we have.

### Acceptance criteria

- [x] SQS client created
- [x] Tests added that show the functionality of receiving/sending messages, mocked out with moto3 similar to our s3 bucket

PR #8378, opened 2026-02-04
Create an adapter for calling SQS with tests

kkrug:
## Summary

Fixes / Work for #8071  

## Changes proposed

- Created a new SQSClient adapter and SQSConfig to provide an interface for receiving and batch-deleting messages.
- Added a suite of unit tests that utilize moto to validate success cases, partial batch failures, and error logging.

## Context for reviewers

This client is a core dependency for the upcoming Workflow Management architecture.

## Validation steps

Confirm that unit tests pass and sufficiently cover potential scenarios.


api/src/adapters/aws/sqs_adapter.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class SQSClient:
+    def __init__(self, queue_url: str, sqs_client: botocore.client.BaseClient | None = None):
+        self.queue_url = queue_url
+        self.client = sqs_client or get_boto_sqs_client()
+
+    def receive_messages(
+        self, max_messages: int = 10, wait_time: int = 10, visibility_timeout: int = 300
+    ) -> list[dict]:
```
chouinar: For the responses from these functions, instead of return dicts, could we make an object of some sort (Pydantic usually works well as it can do JSON -> itself). That way it's easy to know what the format of the data we're working with is.

One of the main reasons for these wrapper functions is that the boto libraries are really annoyingly typed and we want to simplify that.

api/src/adapters/aws/sqs_adapter.py
```diff
@@ -0,0 +1,76 @@
...
+            response = self.client.delete_message_batch(QueueUrl=self.queue_url, Entries=entries)
+
+            results = {}
+            for success in response.get("Successful", []):
+                idx = int(success["Id"])
+                results[receipt_handles[idx]] = "Success"
+
+            for failure in response.get("Failed", []):
+                idx = int(failure["Id"])
+                results[receipt_handles[idx]] = failure.get("Code", "UnknownError")
```
chouinar: Similar to the above, could we have this be a typed object rather than a dict?

Something as simple as:
```py
@dataclass
class SqsDeleteBatchResponse:
    successful_deletes: set[str]
    failed_deletes: set[str]
```
Rather than have a map with receipt->status, just make a set of success/failure with the IDs.

api/src/adapters/aws/sqs_adapter.py
```diff
@@ -0,0 +1,76 @@
...
+            {"Id": str(i), "ReceiptHandle": handle} for i, handle in enumerate(receipt_handles)
+        ]
+
+        try:
+            response = self.client.delete_message_batch(QueueUrl=self.queue_url, Entries=entries)
+
+            results = {}
+            for success in response.get("Successful", []):
+                idx = int(success["Id"])
+                results[receipt_handles[idx]] = "Success"
```
chouinar: I think this works, but something about it feels iffy with indexing into the list passed in.

What if above we did:

```py

receipt_mapping = {}
entries = []
for i, handle in enumerate(receipt_handles):
   i_str = str(i)
   receipt_mapping[i_str] = handle
   entries.append({"Id": i_str, "ReceiptHandle": handle})
```

And then here you could do:
```py
    results.successful_deletes.add(receipt_mapping[success["Id"]])
```

api/tests/src/adapters/aws/test_sqs_adapter.py
```diff
@@ -0,0 +1,114 @@
...
+
+from src.adapters.aws.sqs_adapter import SQSClient, SQSConfig, get_boto_sqs_client
+
+
+class TestSQSConfig:
+    """Tests for the SQS configuration class."""
+
+    def test_config_defaults(self, monkeypatch):
+        """Verify that SQSConfig raises a validation error if required environment variables are missing."""
+        from pydantic import ValidationError
```
chouinar: Could we move this to the top of the file?

api/tests/conftest.py
```diff
@@ -484,6 +484,15 @@ def s3_config(mock_s3_bucket, other_mock_s3_bucket):
...
 
 
+@pytest.fixture
+def mock_sqs():
+    with moto.mock_aws():
+        sqs = boto3.client("sqs", region_name="us-east-1")
+        # Create a default queue for tests
+        queue = sqs.create_queue(QueueName="test-workflow-queue")
+        yield {"client": sqs, "queue_url": queue["QueueUrl"]}
+
```
chouinar, line 499: We want `reset_aws_env_vars` to be called, it makes sure it doesn't use real AWS - see https://docs.getmoto.org/en/latest/docs/getting_started.html#how-do-i-avoid-tests-from-mutating-my-real-infrastructure

Separate the mocking from the queue creation, that way we can have multiple queues 

Don't return the boto3 client from the fixture, as written, you shouldn't even need the client as the SQSClient class will make it.

```suggestion
@pytest.fixture
def mock_sqs(reset_aws_env_vars):
    with moto.mock_aws(config={"core": {"service_whitelist": ["sqs"]}}):
        yield

def workflow_sqs_queue(mock_sqs):
        sqs = boto3.client("sqs", region_name="us-east-1")
        # Create a default queue for tests
        queue = sqs.create_queue(QueueName="test-workflow-queue")
        yield queue["QueueUrl"]
```

We could also do what the moto docs recommend and just make a `mock_all_aws` fixture, remove the config bit from `moto.mock_aws` and use this for both s3/sqs. I can't recall why exactly we added the specific config to s3.

api/src/adapters/aws/sqs_adapter.py
kkrug: I added an `SQSMessage` class to serve as the return object.

api/src/adapters/aws/sqs_adapter.py
kkrug: I added an `SQSDeleteBatchResponse` class to serve as the return object.

api/src/adapters/aws/sqs_adapter.py
kkrug: Updated

api/tests/src/adapters/aws/test_sqs_adapter.py
kkrug: Fixed - also made a bunch of changes to the tests to account for the other changes

api/tests/conftest.py
kkrug: Updated

api/tests/src/adapters/aws/test_sqs_adapter.py
```diff
@@ -0,0 +1,135 @@
...
+        mock_session.client.assert_called_once_with("sqs")
+
+
+class TestSQSClient:
+    """Tests for the SQSClient adapter methods."""
+
+    def test_receive_messages_success(self, workflow_sqs_queue):
+        """Verify that receive_messages successfully retrieves and parses messages as SQSMessage objects."""
+        boto_client = boto3.client("sqs", region_name="us-east-1")
+        sqs_client = SQSClient(queue_url=workflow_sqs_queue, sqs_client=boto_client)
```
chouinar, line 58: Just to keep things slightly less verbose, could we leave out creating the boto3 sqs client? Since you get a client if one isn't passed, it should be the same effectively.

```suggestion
        sqs_client = SQSClient(queue_url=workflow_sqs_queue)
```

Although if a test wants to send a message using the client rather than our wrapper, then we can leave it alone - but a few look like they could drop it since they read from empty queues/error anyways.

kkrug: I was able to clean up the tests a bit, but only made this change for `test_receive_messages_logs_on_error`. The empty queue test needs the boto client to be instantiated within the test to ensure the SQSClient to see the virtual queue, and all of the other tests explicitly reference the boto client.

Issue #8379, opened 2026-02-04
Add environment variables for SOAP partner gateway auth


freddieyebra:
### Summary

We need to setup necessary environment variables and secrets in AWS that will be needed to test this feature in lower environments. The following variables need to be setup in AWS:

Secrets:
`SOAP_PARTNER_GATEWAY_AUTH_KEY` - auth key
- Default value = ""
- This will have the same value in all non-prod environments
- [Reference PR](https://github.com/HHS/simpler-grants-gov/pull/4623/changes#diff-1bca772b07c864a2f2607398972500aa8b3d377926f3c5d54fcac6222bff6d08)

Environment variables (parameter store):
`SOAP_PARTNER_GATEWAY_URI `= https://simpler.ws.dev.grants.gov
- [Reference PR](https://github.com/HHS/simpler-grants-gov/commit/d8f9c665f7b3815642731046b1c77d279e85863a)
- Default value = ""
- This will have same value in all non-prod environments

### Acceptance criteria

- [ ] AWS env vars and secrets exist in all environments (this must be completed prior to merging)
- [ ] Necessary local env vars are configured
- [ ] Updates to support these variables in the `LegacySoapAPIConfig` class
- [ ] Unit tests pass

jakobpederson:
<img width="1030" height="888" alt="Image" src="https://github.com/user-attachments/assets/b8e28c39-4224-4d05-adf6-2cb489d1840a" />

PR #8392, opened 2026-02-05
[Issue #8379] Add environmental variables for Soap Partner Gateway

jakobpederson:
## Summary

Fixes / Work for #8379  

## Changes proposed

Added two variables to the LegacySoapApiConfig: `soap_partner_gateway_uri` and `soap_partner_gateway_auth_key`. Updated the environmental variables to pull those values from AWS Parameter Store. Added those entries to the Parameter Store.

## Context for reviewers

As we move to use the JWT auth for MH in the SOAP/Proxy we need to store the key and uri we will need to use those values.

### DIFF
--- a/infra/api/app-config/env-config/environment_variables.tf
+++ b/infra/api/app-config/env-config/environment_variables.tf
@@ -96,6 +96,16 @@ locals {
       secret_store_name = "/api/${var.environment}/soap-private-keys"
     }
 
+    SOAP_PARTNER_GATEWAY_URI = {
+      manage_method     = "manual"
+      secret_store_name = "/api/${var.environment}/soap-partner-gateway-uri"
+    }
+
+    SOAP_PARTNER_GATEWAY_AUTH_KEY = {
+      manage_method     = "manual"
+      secret_store_name = "/api/${var.environment}/soap-partner-gateway-auth-key"
+    }
+
     SAM_GOV_API_KEY = {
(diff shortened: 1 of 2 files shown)


infra/api/app-config/env-config/environment_variables.tf
```diff
@@ -96,6 +96,16 @@ locals {
...
 
+    SOAP_PARTNER_GATEWAY_URI = {
+      manage_method     = "manual"
+      secret_store_name = "/api/${var.environment}/soap-partner-gateway-uri"
+    }
+
+    SOAP_PARTNER_GATEWAY_AUTH_KEY = {
+      manage_method     = "manual"
+      secret_store_name = "/api/${var.environment}/soap-partner-gateway-auth-key"
+    }
```
chouinar, line 107: Have you added values for all of these into parameter store for every env?

Even if we don't have actual values yet, if they don't exist, deploys will fail as it tries to fetch them.

jakobpederson: I have added these to the parameter store for dev, staging, training, prod
image

freddieyebra: @jakobpederson SOAP_PARTNER_GATEWAY_URI should have been an environment variable not a secret. As Michael mentioned, the environment variable must exist prior to merge.

Issue #8403, opened 2026-02-05
Update tests for next-intl changes related to next 16 upgrade


doug-s-nava:
### Summary

Upgrading to Next 16 involves some breaking changes to test behavior, particularly around usage of the `<NextIntlClientProvider>`. 

Unclear exactly why, but some part of Next 16 upgrade creates situation where `<NextIntlClientProvider>` cannot be used. Rather than trying to fix that, move to a more universal standard around next-intl mocking.

Update test and testing utilities so that tests pass as expected.

### Acceptance criteria

- [ ] all tests pass without usage of `<NextIntlClientProvider>`

PR #8404, opened 2026-02-05
[Issue #8403] fix up tests for Next 16 upgrade and universal next-intl mocking

doug-s-nava:
## Summary

Fixes for #8403

## Changes proposed

Updates testing for:

* standardizing next-intl mocking
* some type related fixes around `ApplicationStatus`

## Context for reviewers

These are changes related to the Next 16 upgrade, but can be enacted outside of the context of the upgrade

The main change here is a move away from the approach to next-intl mocking established in the react-utils file, which wrapped the testing-library render function with a next-intl provider, and toward a global mock using `jest.mock`. The react-utils method broke during the Next 16 upgrade for unknown reasons, but it make sense to standardize this approach regardless so that our tests are not mocking the same thing in two different ways

## Validation steps

1. _VERIFY_: tests pass in CI

### DIFF
--- a/frontend/src/components/search/SearchFilterAccordion/AgencyFilterAccordion.test.tsx
+++ b/frontend/src/components/search/SearchFilterAccordion/AgencyFilterAccordion.test.tsx
@@ -10,6 +10,10 @@ import { ReadonlyURLSearchParams } from "next/navigation";
 
 import { AgencyFilterAccordion } from "src/components/search/SearchFilterAccordion/AgencyFilterAccordion";
 
+jest.mock("next-intl/server", () => ({
+  getTranslations: () => identity,
+}));
+
 jest.mock("src/hooks/useSearchParamUpdater", () => ({
   useSearchParamUpdater: () => ({
     searchParams: new ReadonlyURLSearchParams(),
@@ -20,10 +24,6 @@ jest.mock("src/hooks/useSearchParamUpdater", () => ({
 jest.mock("react", () => ({
   ...jest.requireActual<typeof import("react")>("react"),
   Suspense: ({ fallback }: { fallback: React.Component }) => fallback,
-}));
-
-jest.mock("next-intl/server", () => ({
-  getTranslations: () => identity,
 }));
 
 describe("AgencyFilterAccordion", () => {


frontend/tests/jest.setup.js
```diff
@@ -1,14 +1,28 @@
...
 
 /**
  * Mock environment variables
  */
 process.env.apiUrl = "http://localhost";
+
+/**
+ * Always mock next-intl functions
+ */
+jest.mock("next-intl", () => ({
```
doug-s-nava, line 25: almost all changes are downstream updates to support this global mock, in favor of the deleted mock that it is replacing

frontend/src/components/SaveButton.tsx
```diff
@@ -8,7 +8,7 @@ import Spinner from "./Spinner";
 import { USWDSIcon } from "./USWDSIcon";
 
 interface SaveButtonProps {
-  buttonClick?: (() => Promise<void>) | (() => void);
+  buttonClick?: () => void;
```
doug-s-nava, line 11: dunno how this got in here, should have been in the types PR, but seems like a good change

ErinPattisonNava: Huh, why is this optional?  I guess I'm wondering how to test this, as I'd assume it would be a default everyone would override.

frontend/src/app/[locale]/(base)/events/EventsHero.test.tsx
```diff
@@ -1,11 +1,11 @@
+import { render, screen } from "@testing-library/react";
 import EventsHero from "src/app/[locale]/(base)/events/EventsHero";
-import { render, screen } from "tests/react-utils";
 
 describe("Events Hero Content", () => {
   it("Renders without errors", () => {
     render(<EventsHero />);
     const H1 = screen.getByRole("heading", {
-      name: /Events/i,
+      name: /header/i,
```
Comment (ErinPattisonNava) [inline frontend/src/app/[locale]/(base)/events/EventsHero.test.tsx]: Why did this test pass before?

frontend/src/app/[locale]/(base)/newsletter/action.test.tsx
```diff
@@ -1,7 +1,7 @@
 import { identity } from "lodash";
 import { Response } from "node-fetch";
 import { subscribeEmailAction } from "src/app/[locale]/(base)/newsletter/actions";
-import { mockMessages, useTranslationsMock } from "src/utils/testing/intlMocks";
+import { useTranslationsMock } from "src/utils/testing/intlMocks";
```
Comment (ErinPattisonNava) [inline frontend/src/app/[locale]/(base)/newsletter/action.test.tsx:4]: Can we remove this line?

frontend/src/app/[locale]/(base)/saved-opportunities/page.test.tsx
```diff
@@ -6,24 +7,12 @@ import {
...
-  localeParams,
-  mockUseTranslations,
-  useTranslationsMock,
-} from "src/utils/testing/intlMocks";
-import { render, screen, waitFor } from "tests/react-utils";
-
-import { ReactNode } from "react";
+import { localeParams, mockUseTranslations } from "src/utils/testing/intlMocks";
 
 jest.mock("next-intl/server", () => ({
```
Comment (ErinPattisonNava) [inline frontend/src/app/[locale]/(base)/saved-opportunities/page.test.tsx:12]: Why can't we remove this?  It would be nice to have this be handled automagically too.

frontend/src/app/[locale]/(base)/not-found.test.tsx
```diff
@@ -13,11 +12,6 @@ jest.mock("react", () => ({
 
 jest.mock("next-intl/server", () => ({
   getTranslations: () => identity,
```
Comment (ErinPattisonNava) [inline frontend/src/app/[locale]/(base)/not-found.test.tsx:14]: This feels odd, given that identity is a lodash library.

frontend/src/components/developer/apiDashboard/ApiKeyModal.test.tsx
```diff
@@ -42,65 +40,16 @@ const mockApiKey: ApiKey = {
   api_key_id: "test-api-key-id",
 };
 
-const messages = {
```
ErinPattisonNava, line 45: YAY.

frontend/src/components/newsletter/SubscriptionForm.test.tsx
```diff
@@ -42,7 +36,6 @@ describe("SubscriptionForm", () => {
   // action function is replaced with `javascript:throw new Error('A React form was unexpectedly submitted')`
   // See also https://github.com/facebook/react/blob/f83903bfcc5a61811bd1b69b14f0ebbac4754462/packages/react-dom-bindings/src/client/ReactDOMComponent.js#L468
   // - DWS 2-12-25
-  // eslint-disable-next-line jest/no-disabled-tests
```
ErinPattisonNava, line 45: Nit: Seems odd, is that no longer part of our linting rules?

frontend/src/components/search/SearchFilterAccordion/AgencyFilterAccordion.test.tsx
```diff
@@ -31,6 +22,10 @@ jest.mock("react", () => ({
   Suspense: ({ fallback }: { fallback: React.Component }) => fallback,
 }));
 
+jest.mock("next-intl/server", () => ({
```
ErinPattisonNava: Same as above, wondering why we can't remove all translation mocking.

Nit: If we need to leave it, it would be nice if this could be left in the same place in the file so it doesn't register as a change made in this PR.

frontend/src/app/[locale]/(base)/newsletter/action.test.tsx
Comment (doug-s-nava) [reply frontend/src/app/[locale]/(base)/newsletter/action.test.tsx:4]: the way the test is written it is using the mock function directly for some reason. That should be rewritten, but I'd rather not expand the scope of this PR beyond where it is

frontend/src/app/[locale]/(base)/saved-opportunities/page.test.tsx
Comment (doug-s-nava) [reply frontend/src/app/[locale]/(base)/saved-opportunities/page.test.tsx:12]: we can and probably should include this in the global mocks, but it's not necessary for replacing the existing global mock which only handled client side next-intl mocking https://github.com/HHS/simpler-grants-gov/blob/36417100aed9fdf5bac01829caf7df5d9d01244e/frontend/tests/react-utils.tsx#L22

Maybe we can create a new ticket for coming up with a more comprehensive global mocking system?

frontend/src/app/[locale]/(base)/not-found.test.tsx
Comment (doug-s-nava) [reply frontend/src/app/[locale]/(base)/not-found.test.tsx:14]: how so? This accomplishes the general goal of our next-intl mocks, which is to pass through the key as the content

frontend/src/components/newsletter/SubscriptionForm.test.tsx
doug-s-nava: I think with the new eslint setup it no longer complains about disabled tests for some reason? However, we haven't merged the new eslint setup yet so I honestly don't know why it would have stopped complaining at this point.

frontend/package-lock.json
```diff
@@ -101,18 +101,6 @@
       "dev": true,
       "license": "MIT"
     },
-    "node_modules/@ampproject/remapping": {
```
doug-s-nava: there's no change to the package in this branch, so not sure why this is here. Must have been a missed update from a previous change 🤷

frontend/src/components/SaveButton.tsx
doug-s-nava: It would make sense for this to be a required prop, not sure why it isn't. The only change here is to remove the option to set it as an async function, which the linter doesn't support since it's a click handler

frontend/src/app/[locale]/(base)/not-found.test.tsx
Comment (ErinPattisonNava) [reply frontend/src/app/[locale]/(base)/not-found.test.tsx:14]: But all of our translations are stored within our repository, not in a lodash library.  So I would expect this to be mocked out with something local, or preferably not speciality mocked at all.

frontend/src/app/[locale]/(base)/saved-opportunities/page.test.tsx
Comment (ErinPattisonNava) [reply frontend/src/app/[locale]/(base)/saved-opportunities/page.test.tsx:12]: Eh, I'm not sure that is enough of a priority to follow up on.  I leave it to you.  Just weird that we removed the mocks for "next-intl" but kept select mocks for "next-intl/server"

frontend/src/components/newsletter/SubscriptionForm.test.tsx
ErinPattisonNava: That seems to be a problem, as I do find that a useful linting rule for making sure development code doesn't make it into production (even development test code).  Can you file a follow up ticket and link it here?

frontend/src/app/[locale]/(base)/not-found.test.tsx
Comment (doug-s-nava) [reply frontend/src/app/[locale]/(base)/not-found.test.tsx:14]: the intention here is to mock useTranslations with a function that returns an identity function in order to get useTranslations to spit out the key passed into "t". That way we can assert on the translation key showing up in our tests and not worry about the idea of translation. Lodash identity seems like a perfect use case for this. I could create our own identity function, but since the app is already dependent on lodash, I'm not sure what that would get us

frontend/src/components/newsletter/SubscriptionForm.test.tsx
doug-s-nava: this may resolve with. I'll make a note there to look into again as part of that https://github.com/HHS/simpler-grants-gov/issues/8412

doug-s-nava: https://github.com/HHS/simpler-grants-gov/issues/8412#issuecomment-3955252718

frontend/src/app/[locale]/(base)/not-found.test.tsx
Comment (ErinPattisonNava) [reply frontend/src/app/[locale]/(base)/not-found.test.tsx:14]: Ahh, I get it now.  This does feel like it should be a global mock, rather than each file figuring it out.  Which is why I thought it would be covered by "universal next-intl mocking".  But approving this PR as is for forward momentum.

Issue #7527, opened 2025-12-17
Signature field when submitting on behalf of an organization fills in as expected in the PDF


chris-kuryak:
### Summary

Until a clear requirement is indicated by the client, we need to fill in the signature field on an organization application with something other than "unknown". Instead, we will fill in with the email address of the user that "submits" the application. In other words, the email address we have for the user that clicks the "Submit Application" button.

Current behavior: Shows as "unknown"
<img width="1182" height="460" alt="Image" src="https://github.com/user-attachments/assets/11775588-b591-4c0a-8897-5f1960024b22" />

Expected behavior: Similar to what happens when an individual submits an individual application, the email address of the user who clicks "submit application" the application should be shown.

### Acceptance criteria

- [ ] For Organization applications, the email address of the user who clicks "submit application" is shown after submission in:
  - [ ] The PDF
  - [ ] On the webform (in the read-only field)
  - [ ] In the XML
- [ ] UAT complete 
- [ ] Identify and implement any missing tests to cover for this

chris-kuryak:
Working to confirm expected behavior with client in sprint 5.1, but we should move forward in 5.2 if no confirmation is received.

freddieyebra:
@chris-kuryak since there are use cases where the user that submits may not have an email, the logic will attempt to sign the form with the email if it exists otherwise, it will fallback to unknown. I am open to other fall backs as well.

chris-kuryak:
@freddieyebra 
- Falling back to unknown in the case above is okay for now. Can you create a follow-on stub ticket to create an alert for us to know when someone submits an application and does not have an associated email address? Let me know the ticket number.
- Per our discussion, it seems logical at this time to simplify the signature logic to show the person's email address that clicked submit in the signature field for both Individual and Organization applications

PR #8408, opened 2026-02-05
[Issue #7527] Update signature method to use submitting user email.

freddieyebra:
## Summary

Work for #7527 

## Changes proposed

This pr changes the post population functionality of the signature form field on application submission.

## Context for reviewers

Prior to this change, the signature form field was populated with the first application users email that was associated to the application. This will ensure that the user's email that submits the application will be populated in the signature field.

## Validation steps

1. In the /api directory run `make db-seed-local`
2. Run the api and frontend
3. Open localhost:3000 and login as `many_app_user`
4. Go to workspaces > applications
5. Search for `App for SF424bv1.1`
6. Scroll to forms and select "Include in form submission"
7. Fill out the 2 form fields and save form
8. Submit form
9. Scroll down to the sf424b form and click to open it
10. Verify the signature is "fake_mail@mail.com".

### DIFF
--- a/api/openapi.generated.yml
+++ b/api/openapi.generated.yml
@@ -231,6 +231,52 @@ paths:
       security:
       - ApiUserKeyAuth: []
       - ApiKeyAuth: []
+  /v1/workflows/events:
+    put:
+      parameters: []
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/WorkflowEventResponse'
+          description: Successful response
+        '422':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+          description: Validation error
+        '401':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+          description: Authentication error
+        '403':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+          description: Forbidden
+        '404':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+          description: Not Found
+      tags:
+      - Workflow
+      summary: Send Workflow Event
+      description: Trigger a new workflow or progress an existing one.
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/WorkflowEventRequest'
+      security:
+      - ApiJwtAuth: []
+      - ApiUserKeyAuth: []
   /v1/users/token/logout:
     post:
       parameters: []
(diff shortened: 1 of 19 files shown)


api/src/services/applications/submit_application.py
```diff
@@ -37,6 +37,10 @@ def submit_application(db_session: db.Session, application_id: UUID, user: User)
         application,
     )
 
+    # This assignment will not persist if any of the form validations fail.
+    application.submitted_by_user = user
+    application.submitted_by = user.user_id
```
chouinar: Shouldn't need both, setting a relationship will set the ID.

```suggestion
    # This assignment will not persist if any of the form validations fail.
    application.submitted_by_user = user
```

api/tests/src/api/applications/test_application_routes.py
```diff
@@ -1851,6 +1851,8 @@ def test_application_submit_success(
         == ApplicationAuditEvent.APPLICATION_SUBMITTED
     )
     assert application.application_audits[0].user_id == user.user_id
+    assert application.submitted_by == user.user_id
+    assert application.submitted_by_user == user
```
chouinar: For checking a foreign-key/relationship, just `assert application.submitted_by_user.user_id == user.user_id` like the line above. It effectively tests both in one.

api/src/form_schema/rule_processing/json_rule_field_population.py
```diff
@@ -130,13 +130,16 @@ def get_competition_title(context: JsonRuleContext, json_rule: JsonRule) -> str
...
+    """Get the signature of the user submitting the application
 
+    Signatures occur during the POST_PROCESSING of application submissions.
+
+    If the submitting user has an associated email, we will sign with that, otherwise
+    log that we signed with the unknown value.
+    """
+    application = context.application_form.application
+    if application.submitted_by_user and application.submitted_by_user.email:
+        return application.submitted_by_user.email
```
chouinar, line 142: I know this is more focused on just getting things unbroken, but any interest in putting the persons name if the user has a profile, or is that a later ticket?

freddieyebra: Id say a later ticket. I defer to @chris-kuryak though.

api/src/services/applications/submit_application.py
freddieyebra: It seems that the relationship used in the context to access the user through the application_forms is required and not picked up since the application has not persisted at the time of accessing.

chouinar: I'm not following. What bit of code is having an issue if you don't have both?

Setting a relationship should also immediately/implicitly set the ID, we almost never directly set IDs as it wouldn't ever do anything. If you're seeing an issue by not having both, my immediate assumption would be that there is a bug somewhere and this is just masking that.

Especially since you refer to it by the `submitted_by_user` elsewhere.

freddieyebra: its failing [here](https://github.com/HHS/simpler-grants-gov/pull/8408/changes#diff-3d42fbe2416b78fa9d8238c2e9664aceb2cad54e9ea078605a4aa32de7df0c78R277)

When i just set the submitting user on the application like this
```python
    application.submitted_by = user.user_id
```

That passes all of the tests but setting it like `application.submitted_by_user = user` causes fails

chouinar: Testing it, surprised we don't see this more often, but here's what how it behaves.

Doing `application.submitted_by_user = user` is working just fine. The foreign key doesn't get populated until the DB does a flush, this happens periodically (when the cache fills up enough), or when committing, but you can force it by doing `db_session.flush()` as well.

It's a case that it isn't wrong to do it, but it's unnecessary unless you try to immediately access the ID. In all honesty, I'd just change the test to not check both the ID and the object (Also, don't do `obj == obj` with the DB models). I'd say just change the test to `assert application.submitted_by_user.user_id == user.user_id` - it checks that the DB model is set / and the ID all in one.

Issue #8410, opened 2026-02-06
Typing fixes related to Next 16 upgrade


doug-s-nava:
### Summary

As we upgrade to Next 16 there are some changes to tsconfig and linting necessary that bring up new typing issues. The work here is to fix those issues ahead of the upgrade so as to unblock it.

### Acceptance criteria

- [ ] linting passes
- [ ] ts check passes
- [ ] app builds successfully

PR #8411, opened 2026-02-06
[Issue #8410] typing fixes for Next 16 upgrade

doug-s-nava:
## Summary

Fixes for #8410 

## Changes proposed

Updates / improvements / bugfixes related to typescript typing errors noted in the process of updating Next to v16.

In the sequence of Next upgrade PRs, this PR should be merged first.

## Validation steps

1. _VERIFY_: all CI checks pass
2. `npm run local` on this branch
3. _VERIFY_: app works generally as expected locally


chaitanya-duddella:
this has some conflicts @doug-s-nava 

Also i tested pulling changes onto local
did npm run build and ran locally

tested
local signin
searching oppourtunities
create application
updating forms
uploading attachments etc

Please resolve the conflict and i will approve the changes

frontend/src/components/application/IncludeFormInSubmissionRadio.tsx
```diff
@@ -25,7 +25,7 @@ export const IncludeFormInSubmissionRadio = ({
   >(includeFormInApplicationSubmission ?? null);
   const [loading, setLoading] = useState<boolean>(false);
 
-  const handleChange = (value: string | unknown) => {
+  const handleChange = (value: string) => {
```
ErinPattisonNava, line 28: I worry about how to test this, or why it was `| unknown` in the first place.  Thoughts?

frontend/src/components/applyForm/types.ts
```diff
@@ -9,6 +9,15 @@ import {
 
 import { HTMLAttributes } from "react";
 
+export type GeneralRecord = Record<string, unknown>;
```
ErinPattisonNava, line 12: Can you give some context on why we're using a GeneralRecord and removing the specific examples in each budget form?

frontend/src/components/search/Filters/AgencyFilterBody.tsx
```diff
@@ -44,7 +44,11 @@ export function AgencyFilterBody({
 
   const toggleOptionChecked = (value: string, isChecked: boolean) => {
     const newParamValue = new Set(query);
-    isChecked ? newParamValue.add(value) : newParamValue.delete(value);
+    if (isChecked) {
```
ErinPattisonNava, line 47: If it is preferred to only use ternary operators in assignments and not in logic, can you add that to the frontend coding conventions?  Or is this something that will now error in linting?  I guess I don't quite understand why this is important, so want it captured somewhere.

frontend/src/components/search/SearchError.tsx
```diff
@@ -12,7 +12,7 @@ function isValidJSON(str: string) {
   try {
     JSON.parse(str);
     return true;
-  } catch (e) {
+  } catch (_e) {
```
ErinPattisonNava, line 15: Have we just been doing it wrong all this time?

frontend/src/components/search/SaveSearchModal.tsx
```diff
@@ -77,7 +78,7 @@ function SuccessContent({
       <p>
         {t.rich("successDescription", {
           workspaceLink: (chunks) => (
-            <a href="/saved-search-queries">{chunks}</a>
+            <Link href="/saved-search-queries">{chunks}</Link>
```
ErinPattisonNava, line 81: Could you also add this to frontend conventions, including the extra benefit we get by using Next's Link?

frontend/src/components/user/OpportunitySaveUserControl.tsx
```diff
@@ -59,23 +59,25 @@ export const OpportunitySaveUserControl = ({
...
-  const userSavedOppCallback = async () => {
+  const userSavedOppCallback = () => {
     setLoading(true);
 
     const method = displayAsSaved ? "DELETE" : "POST";
-    try {
-      const data = await updateSaved("/api/user/saved-opportunities", {
-        method,
-        body: JSON.stringify({ opportunityId }),
+    updateSaved("/api/user/saved-opportunities", {
```
ErinPattisonNava, line 66: What's the benefit of this change?

frontend/src/components/application/IncludeFormInSubmissionRadio.tsx
doug-s-nava: not sure why it was like this in the first place, but adding `| unknown` to the type basically nullifies any benefit of adding anything else to the type. `string | unknown` evaluates to `unknown`. In this case we didn't need to be that open with the typing

frontend/src/components/applyForm/types.ts
doug-s-nava: the most confusing part of this refactor was figuring out one standard typing for `UswdsWidgetProps` value. We had defined it in the main type as `string | Array<T> | unknown`, which as mentioned in the previous comment basically made it `unknown`. And it turns out that `unknown` was doing some heavy lifting, as some widgets rely on boolean, number or "object" values as well. Easy enough to account for the number and boolean types, but the object types were a little trickier. 

One place to look, where I didn't bother refactoring to use GeneralRecord but could, is https://github.com/HHS/simpler-grants-gov/blob/52e164d2cde28bdb6bbf1f4d92c6e210d85a9fff/frontend/src/components/applyForm/utils.ts#L460. Here we're dealing with determining values for budget form fields, and typing them as GeneralRecords, even though that type was not mentioned in the type definition for value elsewhere (see https://github.com/HHS/simpler-grants-gov/pull/8411/changes#diff-e50934b0c8bef690db6ecf413e94c4ba14d8fce72db461de583c1b5cdd2750caR441).

Though the actual budget form widgets type each value more strongly, they all generally take a "GeneralRecord" type value prop, so it seems like a valuable generalization.

As to why I removed the more specific typing from the tests, that's to some degree personal preference on my part. I don't like over determining types in unit tests when we can get by with something more general - it adds more confusion that's besides the point of the unit tests.

frontend/src/components/search/Filters/AgencyFilterBody.tsx
doug-s-nava: Yeah, this is to address an error that started coming up once I started updating linting and typing configurations for this upgrade. The idea is that ternaries should produce values and not just statements. I believe the error is from no-unused-expression? https://eslint.org/docs/latest/rules/no-unused-expressions#allowternary

Not a huge fan of the rule, but figured easier to just satisfy the linter for now

frontend/src/components/search/SearchError.tsx
doug-s-nava: yeah, not sure why this was never caught, must have been an issue with the version of the type check / linter we had

frontend/src/components/user/OpportunitySaveUserControl.tsx
doug-s-nava: click handlers complain if you pass in a function that returns anything. Any async function by default returns a promise even if there's no return statement, so had to convert this into a normal function in order to satisfy the type check. Another one where I'm not sure why it wasn't already complaining about this

frontend/src/components/application/IncludeFormInSubmissionRadio.tsx
ErinPattisonNava: Should this go into the conventions as an anti-pattern?

frontend/src/components/applyForm/types.ts
ErinPattisonNava: Thanks for the explanation.  I agree with the more generalized typing in unittests.

frontend/src/components/user/OpportunitySaveUserControl.tsx
ErinPattisonNava: Can this get added to the frontend conventions as well?

frontend/src/components/application/transferOwnership/TransferOwnershipModal.tsx
```diff
@@ -121,12 +121,12 @@ export const TransferOwnershipModal = ({
     () =>
       t.rich("contactSupport", {
         tel: (content) => {
-          const phoneNumber = String(content);
-          return <a href={`tel:${phoneNumber}`}>{phoneNumber}</a>;
+          const telephoneNumber = content as string;
```
ErinPattisonNava, line 124: Is this also a pattern to be added to the frontend conventions?

frontend/src/components/applyForm/widgets/SelectWidget.tsx
```diff
@@ -66,7 +66,7 @@ function SelectWidget<
   // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
   const { enumOptions: opts, enumDisabled, emptyValue: optEmptyVal } = options;
   const enums = useMemo(() => (opts && opts.length > 0 ? opts : []), [opts]);
-  const selectValue = value ? String(value) : "";
+  const selectValue = value ? String(value as string) : "";
```
ErinPattisonNava: Nit: Should this just be `selectValue = value ? (value as string) : ""` as the `String(value as string)` seems oddly redundant.

frontend/src/components/search/Filters/AgencyFilterBody.tsx
ErinPattisonNava: Hrm.  That is obscure, and I'm not a fan either.  But since this is the only case that was existing in our code base, yeah I can see letting it through for now.

frontend/src/components/application/IncludeFormInSubmissionRadio.tsx
doug-s-nava: This should now be caught by the linter after the upgrade that's in the coming PRs, so maybe we can just rely on that? Should avoid bloating the conventions too much

frontend/src/components/application/transferOwnership/TransferOwnershipModal.tsx
doug-s-nava: eh, I don't know if we need to be opinionated here. This change may not even be necessary to satisfy the linter, now that I look at it.

frontend/src/components/user/OpportunitySaveUserControl.tsx
doug-s-nava: done

ErinPattisonNava (commented):
Tested 
* Search filters
* API key modal
* Submit application radio button
* Error handling in application submissions
* attachments in forms
* all widgets in SF424

Everything looks good!  But maybe don't merge this until after deploy tomorrow, just to give things time to burn in.

Issue #8380, opened 2026-02-04
Implement method for building auth request


freddieyebra:
### Summary

Implement a method for building JWTs with necessary payload. Use sample script that was a result from [this spike](https://github.com/HHS/simpler-grants-gov/issues/8270?issue=HHS%7Csimpler-grants-gov%7C8271)

This should just be the function implementation with unit tests.

### Acceptance criteria

- [ ] Function for building JWT to use in the partner gateway auth implementation
- [ ] Unit tests pass
- [ ] JWT generated with this new method is used to hit the service and get successful response as part of pr verification.

Issue #8401, opened 2026-02-05
Remove stale artifacts from old auth


freddieyebra:
### Summary

We will need to remove the following artifacts from the simpler soap code base:

1. Temporary code needed for transitioning into the newer soap
2. Stale environment variables and secrets.
3. Stale code from old auth.

### Acceptance criteria

- Above resources removed from api

PR #8447, opened 2026-02-09
[Issue #8380] Implement method to build soap jwt request

jakobpederson:
## Summary

Fixes / Work for #8380  

## Changes proposed

Added a method in the proxy that creates a jwt token and attaches it to a request that gets sent to grants.gov.

## Context for reviewers

In order to transition to the new jwt auth the MH supports we have to be able to get the correct `cert_id` from our `legacy_certificate` table and put that in the payload of a jwt. We use that instead of the certificate/key pair we have been using previously.


api/src/legacy_soap_api/legacy_soap_api_proxy.py
```diff
@@ -40,9 +45,25 @@ def get_proxy_response(soap_request: SOAPRequest, timeout: int = PROXY_TIMEOUT)
...
+            )
+            return _get_soap_response(_request, timeout=timeout)
+        else:
+            logger.info(
+                "soap_client_certificate: soap jwt auth certificate not found",
+                extra={"soap_api_event": LegacySoapApiEvent.NOT_CONFIGURED_CERT},
+            )
+            return get_soap_error_response(
+                faultstring="Client certificate not configured for Simpler SOAP jwt auth."
+            )
```
chouinar: I think we could make this quite a bit simpler. All we need to do in this added logic is add a header param, you're instead setting up an entirely different branch of request building, replacing what comes below this.

Can we instead do something like:
```py
if use_jwt_auth:
    proxy_headers["S2S_PARTNER_CERTID_JWT_B64"] = get_soap_jwt(soap_auth.certificate.legacy_certificate)
    # whatever logging
```

Just adjust the headers here, don't do the request. Otherwise you have to dupe all of the error handling logic. You can then adjust the if-statements below so this flows into the first one I guess? Or at least something like that, the logging might need to adjust a bit.

jakobpederson: That sounds a lot better, I was struggling with the error handling.

chouinar:
@jakobpederson - I'm curious, with this we should be able to call with a JWT, right? have you tested that by proxying to their dev env?

api/src/legacy_soap_api/legacy_soap_api_proxy.py
```diff
@@ -37,14 +42,22 @@ def get_proxy_response(soap_request: SOAPRequest, timeout: int = PROXY_TIMEOUT)
...
+    if use_jwt_auth and soap_auth and soap_auth.certificate.legacy_certificate:
+        proxy_headers = {
+            "S2S_PARTNER_CERTID_JWT_B64": get_soap_jwt_auth_jwt(
+                config, soap_auth.certificate.legacy_certificate
+            )
+        }
+    else:
+        proxy_headers = filter_headers(
+            soap_request.headers, [config.gg_s2s_proxy_header_key, MTLS_CERT_HEADER_KEY]
+        )
```
chouinar, line 57: With this change, does the MTLS cert still need to be passed? I think it does? @freddieyebra - I think you're more familiar with this.

jakobpederson:
> @jakobpederson - I'm curious, with this we should be able to call with a JWT, right? have you tested that by proxying to their dev env?

I've been able to hit the grants.gov endpoint ("https://simpler.ws.dev.grants.gov/grantsws-agency-partner/services/v2/AgencyWebServicesSoapPort") but haven't tried proxying it through to the dev env. I'll test it out in staging.

api/tests/src/legacy_soap_api/test_legacy_soap_api_proxy.py
```diff
@@ -140,3 +149,122 @@ def test_get_proxy_response_logs_soap_client_certificate_not_configured_error_an
...
+        factories.LegacyAgencyCertificateFactory.create()
+    )
+    with freeze_time("2023-05-10 12:00:00", tz_offset=0):
+        s2s_partner_certid_jwt_b64 = get_soap_jwt_auth_jwt(
+            mock_config,
+            mock_soap_request.auth.certificate.legacy_certificate,
+        )
+        decoded_base64 = base64.b64decode(s2s_partner_certid_jwt_b64).decode("utf-8")
+        original_claims = jwt.decode(
+            decoded_base64, key=mock_config.soap_partner_gateway_auth_key, algorithms=["HS256"]
```
chouinar: You're doing quite a lot of mocking for things that you can do directly.

Except for calls to external systems, I'd argue mocking is generally more confusing/worse for testing.

```suggestion
    config = LegacySoapAPIConfig(SOAP_PARTNER_GATEWAY_URI="https://grants.gov", SOAP_PARTNER_GATEWAY_AUTH_KEY="X-Gg-S2S-Uri")
    cert = factories.LegacyAgencyCertificateFactory.create()

    with freeze_time("2023-05-10 12:00:00", tz_offset=0):
        s2s_partner_certid_jwt_b64 = get_soap_jwt_auth_jwt(
            config, cert
        )
        decoded_base64 = base64.b64decode(s2s_partner_certid_jwt_b64).decode("utf-8")
        original_claims = jwt.decode(
            decoded_base64, key=config.soap_partner_gateway_auth_key, algorithms=["HS256"]
```

If you don't have a direct way to pass parameters into a config like this test, use monkeypatch to set the env vars. And for most tests that aren't testing something other than the default env var, just set a default somewhere (local.env, a fixture).

api/tests/src/legacy_soap_api/test_legacy_soap_api_proxy.py
```diff
@@ -140,3 +149,122 @@ def test_get_proxy_response_logs_soap_client_certificate_not_configured_error_an
...
+    mock_soap_request.data = SOAP_PAYLOAD
+    mock_soap_request.auth = MagicMock()
+    mock_soap_request.auth.certificate.legacy_certificate = (
+        factories.LegacyAgencyCertificateFactory.create()
+    )
+    with patch("src.legacy_soap_api.legacy_soap_api_proxy._get_soap_response") as _, patch(
+        "src.legacy_soap_api.legacy_soap_api_proxy.get_soap_config"
+    ) as mock_get_config, patch(
+        "src.legacy_soap_api.legacy_soap_api_proxy.get_soap_jwt_auth_jwt"
+    ) as mock_soap_jwt_auth:
```
chouinar: Mocking the soap response is fine, but why mock the auth function or the request object? Is there anything stopping us from just doing things directly?

Mocking means you're not running something for real, which means you're testing less.

api/src/legacy_soap_api/legacy_soap_api_proxy.py
jakobpederson: I'm leaving this in at the moment because there's [a card to pull all the stale code ](https://github.com/HHS/simpler-grants-gov/issues/8401)

api/tests/src/legacy_soap_api/test_legacy_soap_api_proxy.py
jakobpederson: I'll strip out the extraneous mocks.

api/src/legacy_soap_api/legacy_soap_api_proxy.py
freddieyebra: Nothing should change for the proxy requests that do not pass in the `use-jwt-auth` flag. For the new auth, we got a successful response without passing the cert and just the jwt.

chouinar: We'll want to test that in the lower envs, I thought the MTLS cert was needed for our server to reach their server (before any meaningful auth), but I wasn't directly involved in that.

freddieyebra: Yeah I agree. @mdragon From reviewing the resources we were provided, there is currently only one environment (non-prod) for the new s2s API, but can you confirm?

jakobpederson:
> > @jakobpederson - I'm curious, with this we should be able to call with a JWT, right? have you tested that by proxying to their dev env?
> 
> I've been able to hit the grants.gov endpoint ("https://simpler.ws.dev.grants.gov/grantsws-agency-partner/services/v2/AgencyWebServicesSoapPort") but haven't tried proxying it through to the dev env. I'll test it out in staging.

I had to hardcode in a cert id but I managed to hit my local endpoint and get data back using the jwt.

api/tests/conftest.py
```diff
@@ -116,6 +116,12 @@ def set_env_var_defaults(monkeypatch_session):
 
     # We will set this to false so we skip logs during unit tests and keep enabled during dev.
     monkeypatch_session.setenv("SOAP_ENABLE_VERBOSE_LOGGING", "0")
+    monkeypatch_session.setenv("SOAP_PARTNER_GATEWAY_URI", "soap_partner_gateway_uri")
+    monkeypatch_session.setenv("SOAP_PARTNER_GATEWAY_AUTH_KEY", "soap_partner_gateway_auth_key")
+    monkeypatch_session.setenv(
+        "SOAP_PRIVATE_KEYS",
+        '{"0000000000000000000000000000000000000000000000000000000000000000": {}}',
+    )
```
chouinar: Any of these that make sense to set as defaults for local dev? If we put something in local.env, it gets loaded for unit tests.

Most of what we set in this function is specific overrides for tests that differ from local.env. Not a big deal to add, but just checking.

api/tests/src/legacy_soap_api/test_legacy_soap_api_proxy.py
```diff
@@ -116,27 +126,120 @@ def test_get_proxy_response_logs_soap_client_certificate_not_configured_error_an
...
+    mock_soap_request.data = SOAP_PAYLOAD
+    mock_soap_request.auth = MagicMock()
+    mock_soap_request.auth.certificate.legacy_certificate = (
+        factories.LegacyAgencyCertificateFactory.create()
+    )
+    mock_soap_config = get_soap_config()
+    s2s_partner_certid_jwt_b64 = get_soap_jwt_auth_jwt(
+        mock_soap_config,
+        mock_soap_request.auth.certificate.legacy_certificate,
+    )
```
chouinar: I'm going to start being a bit more strict on tests (this applies to all the tests here).

If we wanted a request object, why mock instead of making it directly? And we don't even use the request object. You build an entire thing and then pull something out of it.

```suggestion

    legacy_cert = factories.LegacyAgencyCertificateFactory.create()
    mock_soap_config = get_soap_config()
    s2s_partner_certid_jwt_b64 = get_soap_jwt_auth_jwt(
        mock_soap_config,
        legacy_cert,
    )
```

My suggestion, except for external calls, you should just not mock. If you want to easily be able to make a SOAPRequest, make a util that has some reasonable defaults (a sort of factory function). That's the general pattern we follow as it ends up less verbose and more flexible. Even if there is a strong reason to use the mock, make a utility for generating the object, copy-pasting the same 15 lines of boilerplate to each test isn't how we want to write these.

api/tests/conftest.py
jakobpederson: I'll see about setting them in the local dev, I was just having issues with tests failing on when I pushed to the branch despite passing locally.

chouinar (approved):

Issue #8239, opened 2026-01-29
SF424 applicant type code not properly referenced


doug-s-nava:
### Summary

looks like there's a bug with the combo box input. Saving the sf424 I get an error `Undefined field type shaping form data applicant_type_code__combobox_input`. Indeed, this path `applicant_type_code__combobox_input` does not match any path in the form schema. The result is that the json is being stored with a key of `applicant_type_code__combobox_input` rather than `applicant_type_code`. This may not be causing any real issues at the moment, but it will once we get logic in place that will be expecting the submitted JSON to match up with the form schema shape. Should be an easy fix to just use the id here instead of adding combo_box but maybe that's being done for other reasons? https://github.com/HHS/simpler-grants-gov/blob/5f124a33fc14c6644cdcb7c2f8c563266744445b/frontend/src/components/applyForm/widgets/MultiSelectWidget.tsx#L163

### Reproducibility

Always

### How to reproduce the bug

- save the sf424 on an application with applicant type code populated
- in the db, the entry for the `application_response` column in the the `application_form` table contains `application_type_code__combobox_input` rather than the expected `application_type_code`

### Browser

None

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

hayeshorus:
I have been researching this and testing various changes to see if I could restrict the application_type_code_combobox_input data from being included in the output from the form save.   I tried using atMaxSelection value to restrict the combo box from creating the application_type_code_combobox_input with an additional 4th value.  I have tried to modify the addByLabelOrValue logic to not added any additional elements after the atMaxSelection value is 3.  I also tried checking variables and not doing anything when selected values match and max is full.   I have not found a working solution so that the output data is correct when saved.  The application_type_code_combobox_input is not used anywhere and therefore does not need to be in the output data.  I am testing more, but feel just making the name field an empty string still feels like a good solution.

chris-kuryak:
UAT passed

PR #8458, opened 2026-02-09
[#8239] Application Type Code updated to remove combobox_input

hayeshorus:
[#8239] Application Type Code updated to remove combobox_input from name

## Summary
updated logic to remove the combobox_input from the application type code name field to fix DB save error

Work for #8239

## Changes proposed

Removed the combobox_input text from the application type code field

Screenshot 2026-02-09 at 11 22 00 AM

## Validation steps

1	Search for a new opportunity
2	Type in Test Opportunity for in the search bar
3	Click on Test Opportunity for SF424 4_0 4.0
4	Start a new application
5	Enter data in for Who’s Applying and type in a name. The click create application button
6	Find the Application for Federal Assistance (SF-424) in the Conditionally-Required Forms section
7	Click on the Application for Federal Assistance (SF-424) Form link
8	Enter in data for all required fields
9	Click on save, to save the form the DB
10   Open up Postgres
11   Find the Application_Form table and search for the above entry in the table
12   Find the Appilication_Response column and verify the data show application_type_code

### DIFF
--- a/api/src/api/agencies_v1/agency_routes.py
+++ b/api/src/api/agencies_v1/agency_routes.py
@@ -5,7 +5,7 @@ import src.api.response as response
 from src.adapters import search
 from src.adapters.search import flask_opensearch
 from src.api.agencies_v1.agency_blueprint import agency_blueprint
-from src.auth.multi_auth import api_key_multi_auth, api_key_multi_auth_security_schemes
+from src.auth.multi_auth import jwt_or_user_api_key_multi_auth, jwt_or_user_api_key_security_schemes
 from src.logging.flask_logger import add_extra_data_to_current_request_logs
 from src.services.agencies_v1.search_agencies import search_agencies
 from src.util.dict_util import flatten_dict
@@ -59,8 +59,8 @@ examples = {
     examples=examples,
 )
 @agency_blueprint.output(agency_schema.AgencySearchResponseV1Schema)
-@agency_blueprint.doc(security=api_key_multi_auth_security_schemes)
-@api_key_multi_auth.login_required
+@agency_blueprint.doc(security=jwt_or_user_api_key_security_schemes)
+@jwt_or_user_api_key_multi_auth.login_required
 @flask_opensearch.with_search_client()
 def agency_search(
     search_client: search.SearchClient, raw_search_params: dict
--- a/api/src/workflow/listener/workflow_audit_listener.py
+++ b/api/src/workflow/listener/workflow_audit_listener.py
@@ -4,6 +4,7 @@ from statemachine.event_data import EventData
 
 from src.adapters import db
 from src.db.models.workflow_models import WorkflowAudit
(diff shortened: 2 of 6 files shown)


hayeshorus:
> this works, and I figured out why here https://betagrantsgov.slack.com/archives/C09RKTHP19N/p1771880439999259?thread_ts=1771616893.513899&cid=C09RKTHP19N - given the complexity here, I think this component needs some further documentation. Can you add a jsdoc for the component that explains how it works, and why we're setting the name to an empty string? Feel free to work with @myduong-navapbc since he built the component if anything is unclear

I added a comment to the component that is under the node modules code.   Is that where we want it?

hayeshorus:
@doug-s-nava - I have added the comment for the discussion on the combo box.   This is ready for another review.

frontend/src/components/applyForm/widgets/MultiSelectWidget.tsx
```diff
@@ -132,6 +132,9 @@ export default function MultiSelect<
     allOptions.find((option) => String(option.value) === value)?.label ?? value;
 
   // ComboBox widget changes the id which breaks handling of idFor and anchor links
+  // this component creates a <select> with the name attached to the <ComboBox> component
+  // when selecting the code takes the value and creates a hidden input based on the index of the selection and the id
+  // component tracks selected option internally, set the name to "" to not set the value of the select.
```
doug-s-nava: this isn't super clear - can you mention the interaction with the truss component, and put this in a jsdoc rather than in the body of the component? https://jsdoc.app/

the comment should answer the questions:
* how does this component communicate / store selected values to the DOM?
* why is the ComboBox name set to an empty string?

frontend/src/components/applyForm/widgets/MultiSelectWidget.tsx
```diff
@@ -131,7 +131,16 @@ export default function MultiSelect<
   const getLabelForValue = (value: string) =>
     allOptions.find((option) => String(option.value) === value)?.label ?? value;
 
-  // ComboBox widget changes the id which breaks handling of idFor and anchor links
+  /** ComboBox widget changes the id which breaks handling of idFor and anchor links
```
doug-s-nava: sorry to send this back, but JSDocs need to be located above the function definition that they are annotating, and some of this language is still unclear to me.

i'd suggest something like this:
```
This component denotes selected items in the DOM by creating hidden inputs whose names correspond to the id passed into the component, along with an index for each selected item. Ex

<an example of what the hidden inputs will look like for a given component id>

Note that the Trussworks ComboBox creates a select component that will take a name based on the "name" prop passed in. Based on the component's internal state management logic, it will maintain a value for the select on its own. For some reason, it will select the first entry in the option list on close, and store that as its value. In order to prevent his value to be picked up on save along with the hidden inputs mentioned above, we set the "name" for the ComboBox to an empty string.
```

PR #8467, opened 2026-02-10
[Unticketed] Add a unit test that validates all factories work

chouinar:
## Changes proposed
Add a unit test to our factories that validates every factory works and `create()` can be called.

## Context for reviewers
Something that can be tricky for folks is adding/modifying our DB tables, and setting up factories not knowing if their factories are right. Factories are ultimately just for test data, so writing unit tests for them isn't something we generally do (testing tests leads to testing test tests, and other confusing ideas). But it's useful to have at least the most basic of validation that what you setup works. So, this test I added does that. For each factory it calls `.create()` on it, verifying that at least the defaults defined work, and can in fact be inserted into the DB.

I removed a factory that wasn't used anywhere that otherwise needed fixing, and fixed another one that didn't work. A handful can't be run directly, which is explained in the test why.

## Validation steps
The test itself

### DIFF
--- a/api/tests/src/db/models/factories.py
+++ b/api/tests/src/db/models/factories.py
@@ -63,8 +63,6 @@ from src.constants.static_role_values import (
     ORG_MEMBER,
 )
 from src.db.models import agency_models
-from src.db.models.lookup.lookup_registry import LookupRegistry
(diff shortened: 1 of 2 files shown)


kkrug:
@chouinar Checking in on some older review requests in my queue - is this PR still relevant?

Issue #8475, opened 2026-02-10
Investigate flakiness of saved opportunities test


doug-s-nava:
### Summary

See https://betagrantsgov.slack.com/archives/C0ABAM9AB6X/p1770737243164669

The "Working saved opportunities page link appears in nav when logged in" test has been failing intermittently.

It looks like the issue is `createSpoofedSessionCookie` function is not always working properly

### Acceptance criteria

- [ ]. a timeboxed attempt (1 day) is made to improve reliability of the test in question

PR #8481, opened 2026-02-10
[Issue #8475] reenable skipped saved opportunity e2e test and fix it

doug-s-nava:
## Summary

Work for #8475 

## Changes proposed

Fixes bugs that were causing the "Working saved opportunities page link appears in nav when logged in" e2e test to fail, and re-enables the test

## Context for reviewers

Two things seemed to have been causing this test to fail:

* The expiration time set on the dummy user token that is used in the failing test was too short (5 minutes), and was causing the frontend to try and refresh the token (which failed because the token doesn't exist on the API side)
* The environment file generated for the e2e tests was coming out malformed due to the lack of a newline at the end of the checked in .env.development file. As a result, the necessary env vars for login functionality were not being loaded. This is a regression from https://github.com/HHS/simpler-grants-gov/commit/534f73201cd6de40eb48dbd867fd076abeecc573#diff-8418659e03d1d6494d6c53bd2f591905fdc4369cf108de7b249bb796fb88aa1bR46. Adding the newline back to the end of the file allows the correct env vars to come through.
## Validation steps

1. _VERIFY_: all e2e tests pass consistently in CI


myduong-navapbc (approved):
Yay

Issue #8099, opened 2026-01-22
Create a stub version of the events API


chouinar:
### Summary

We want to have the endpoint created as `PUT /v1/workflows/events` - this will require setting up a new workflow API blueprint and connecting it to the API. There should be a feature flag that gates this blueprint from being available. We should enable it locally and in dev/staging, but default it to off, and leave it off in training/prod.

Specific implementation details are defined in the [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2374369297/Tech+Spec+Workflow+Management+with+Python+StateMachine#Event-API).

This endpoint needs to have the following:
* Request and response schemas defined
* Allows API User Key or JWT auth
* Converts the request into the event model object in the service layer, raising a 422 if there are any issues (ideally the request schema should catch any issues, but the dual nature of the events might make it tricky).

The service layer of the implementation should be created, but other than converting the request model, won’t actually send events yet. We won’t have any AuthZ yet for this endpoint either (this is fine as the endpoint won't be usable outside of dev/staging AND won't do anything beyond schema validation).

This can return a static/stub response.

### Acceptance criteria

- [ ] Workflow PUT api created, including making sure the workflow blueprint is behind a feature flag (default to False, Terraform needs to enable dev/staging)
- [ ] Tests added for happy path, request schema validation, and auth

PR #8487, opened 2026-02-11
Create events API stub

kkrug:
## Summary

Fixes for #8099  

## Changes proposed

Created the events API, including the route, schema with validation, and a service layer stub to be expanded upon. The blueprint is behind a feature flag, and tests were also added.

## Context for reviewers

This events API is to support Workflow Management as defined in the [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2374369297/Tech+Spec+Workflow+Management+with+Python+StateMachine).

## Validation steps

Confirm that tests run successfully and that the API is correctly hidden by the feature flag.

### DIFF
--- a/api/src/api/workflows/workflow_schemas.py
+++ b/api/src/api/workflows/workflow_schemas.py
@@ -30,13 +30,8 @@ class StartWorkflowContextSchema(Schema):
     entities = fields.List(
         fields.Nested(WorkflowEntitySchema),
         required=True,
-        validate=[validators.Length(min=1, max=5, error="maximum length 5")],
-        metadata={
-            "description": "A list of entities (1-5) to associate with the workflow.",
-            "example": [
-                {"entity_type": "opportunity", "entity_id": "3fa85f64-5717-4562-b3fc-2c963f66afa6"}
-            ],
-        },
+        validate=[validators.Length(min=1, max=5)],
+        metadata={"description": "A list of entities (1-5) to associate with the workflow."},
     )
 
 
@@ -66,27 +61,14 @@ class WorkflowEventRequestSchema(Schema):
         StartWorkflowContextSchema,
         required=False,
         metadata={
-            "description": "Context and entities required to initialize a new workflow. Only allowed if event_type is 'start_workflow'.",
-            "example": {
-                "workflow_type": "opportunity_publish",
-                "entities": [
-                    {
-                        "entity_type": "opportunity",
-                        "entity_id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
-                    }
-                ],
-            },
+            "description": "Context and entities required to initialize a new workflow. Only allowed if event_type is 'start_workflow'."
         },
     )
     process_workflow_context = fields.Nested(
         ProcessWorkflowContextSchema,
         required=False,
         metadata={
-            "description": "Information required to progress an existing workflow state. Only allowed if event_type is 'process_workflow'.",
-            "example": {
-                "workflow_id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
-                "event_to_send": "approve",
-            },
+            "description": "Information required to progress an existing workflow state. Only allowed if event_type is 'process_workflow'."
         },
     )
     metadata = fields.Dict(


api/src/api/workflows/workflow_schemas.py
```diff
@@ -0,0 +1,95 @@
...
+                    ],
+                    "start_workflow_context",
+                )
+
+            entities = start_ctx.get("entities", [])
+            if len(entities) < 1 or len(entities) > 5:
+                raise ValidationError(
+                    [MarshmallowErrorContainer(ValidationErrorType.INVALID, "maximum length 5")],
+                    "start_workflow_context",
+                )
```
chouinar: For length, you don't need to make a custom validation rule like this, just add `validate=[validators.Length(min=1, max=5)]` to the entities definition above.

api/src/api/workflows/workflow_routes.py
```diff
@@ -0,0 +1,29 @@
...
+@workflow_blueprint.doc(
+    summary="Send Workflow Event",
+    description="Trigger a new workflow or progress an existing one.",
+    security=jwt_or_user_api_key_security_schemes,
+    responses=[200, 401, 422],
+)
+@jwt_or_user_api_key_multi_auth.login_required
+def workflow_event_put(json_data: dict) -> response.ApiResponse:
+
+    event_id = ingest_workflow_event(json_data)
```
chouinar: Add the DB session, even if we don't need it yet.

Add a bit for logging.

```suggestion
@jwt_or_user_api_key_multi_auth.login_required
@flask_db.with_db_session()
def workflow_event_put(db_session: db.Session, json_data: dict) -> response.ApiResponse:
    # there might be more we could add here, but that might be a bit more parsing than we want to do before turning it into the models.
    add_extra_data_to_current_request_logs({"event_type": json_data.get("event_type")})
    logger.info("PUT /v1/workflows/events")
    
    with db_session.begin():
         event_id = ingest_workflow_event(json_data)
```

api/src/api/feature_flags/feature_flag_config.py
```diff
@@ -21,6 +21,9 @@ class FeatureFlagConfig(PydanticBaseEnvConfig):
         False, alias=FeatureFlag.ENABLE_OPPORTUNITY_LOG_MSG.get_env_var_name()
     )
 
+    # ENABLE_WORKFLOW_API
+    enable_workflow_api: bool = Field(False, alias="ENABLE_WORKFLOW_API")
+
```
chouinar: Don't put this in the feature flag config, put it in the endpoint config. We don't really use this, this is more of a placeholder feature that we haven't ever built out.

api/src/api/workflows/workflow_schemas.py
```diff
@@ -0,0 +1,95 @@
...
+from src.api.schemas.extension import Schema, fields, validators
+from src.api.schemas.extension.schema_common import MarshmallowErrorContainer
+from src.api.schemas.response_schema import AbstractResponseSchema
+from src.constants.lookup_constants import WorkflowEntityType, WorkflowEventType, WorkflowType
+from src.validation.validation_constants import ValidationErrorType
+
+
+class WorkflowEntitySchema(Schema):
+    entity_type = fields.Enum(WorkflowEntityType, required=True)
+    entity_id = fields.UUID(required=True)
```
chouinar: For our schemas, we like to include examples/descriptions. Not every field needs an example (generally just basic string fields, others get reasonable defaults). But we should always have a description.

```suggestion
class WorkflowEntitySchema(Schema):
    entity_type = fields.Enum(WorkflowEntityType, required=True, metadata={"description": "The type of entity to associate with the workflow - different workflows require different entities."})
    entity_id = fields.UUID(required=True, metadata={"description": "The ID of the entity, for example an opportunity_id"})
```

These ones wouldn't need examples as enums would put one of them, and uuid gets defaulted to a static UUID. Something like `event_to_send` would want a default as a string field defaults to "string" which isn't too helpful.

api/src/api/workflows/workflow_routes.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+@workflow_blueprint.put("/events")
+@workflow_blueprint.input(WorkflowEventRequestSchema)
+@workflow_blueprint.output(WorkflowEventResponseSchema)
+@workflow_blueprint.doc(
+    summary="Send Workflow Event",
+    description="Trigger a new workflow or progress an existing one.",
+    security=jwt_or_user_api_key_security_schemes,
+    responses=[200, 401, 422],
```
chouinar: 403 (AuthZ) and 404 (not finding an entity / workflow) will also be possible. I suppose 500 as well, although not sure if we should call that out.

api/src/api/feature_flags/feature_flag_config.py
kkrug: What do you mean exactly by "put it in the endpoint config?" Does that mean to create a workflow_config.py and put something like this?:
`class WorkflowConfig(BaseConfig):
    enable_workflow_api: bool = Field(False, alias="ENABLE_WORKFLOW_API")`

api/src/api/workflows/workflow_schemas.py
kkrug: Fixed, and updated test to reflect change

api/src/api/workflows/workflow_routes.py
kkrug: Added db session and logging

api/src/api/workflows/workflow_schemas.py
kkrug: Added descriptions and examples

api/src/api/workflows/workflow_routes.py
kkrug: Added 403 and 404

api/src/api/feature_flags/feature_flag_config.py
chouinar: In the app.py file, there's an EndpointConfig class that we use for enabling/disabling endpoints - it's used in the function where you added the blueprint.

chouinar:
@kkrug - I see you replied saying you fixed a few things, but I don't think you pushed the commit(s)

kkrug:
> @kkrug - I see you replied saying you fixed a few things, but I don't think you pushed the commit(s)

They should be there now, including the feature flag change.

api/src/api/feature_flags/feature_flag_config.py
kkrug: Feature flag has been updated

api/src/api/workflows/workflow_schemas.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+class StartWorkflowContextSchema(Schema):
+    workflow_type = fields.Enum(
+        WorkflowType, required=True, metadata={"description": "The type of workflow to initiate"}
+    )
+    entities = fields.List(
+        fields.Nested(WorkflowEntitySchema),
+        required=True,
+        validate=[validators.Length(min=1, max=5, error="maximum length 5")],
```
chouinar: I don't think the error field gets used, the errors have a formatter built into the validator (note we override/re-implement the validator classes). Whatever you passed in here isn't used based on your unit tests getting the normal error message.

We should keep error formats for these consistent regardless and I don't think we'd want to change it.

api/src/api/workflows/workflow_schemas.py
```diff
@@ -0,0 +1,156 @@
...
+            "example": {
+                "workflow_type": "opportunity_publish",
+                "entities": [
+                    {
+                        "entity_type": "opportunity",
+                        "entity_id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
+                    }
+                ],
+            },
+        },
```
chouinar: I'd suggest for any of these nested objects to leave the examples out - it'll build an example for you from the fields within it, and it just ends up being extra work / more to maintain.

kkrug: Got it, removed examples from the three nested fields

api/src/api/workflows/workflow_schemas.py
kkrug: Correct that was left over from the previous iteration. It is now removed

PR #8491, opened 2026-02-11
[Issue #8069] Build core workflow state machine logic

### LINKED ISSUE #8069: Build core workflow state machine logic
### Summary

This is really just cleaning up and setting up patterns from the prototype foundation.

This probably needs to be broken up a bit, but includes:
* Creating a state machine that we can use for testing / building onto
* Setting up workflow configuration
* Setting up mapping of events to workflow logic
* Basic error handling

### Acceptance criteria

- [ ] Clean up the prototype version of workflow state machine implementation with complete code coverage
- [ ] Create follow-up tickets for components that make sense to be implemented separately

chouinar:
## Summary
Fixes #8069

## Changes proposed
* Adds a significant amount of core workflow components and utilities to build upon
* Adds a simple integration with the Python StateMachine library (future PRs will expand the prototype)
* Adds logic to make one of our SQS events validate and be passed to a state machine

## Context for reviewers
This change looks far larger than it is, don't be intimidated by the file count, I put a lot of these components in separate folders, so a lot of `__init__.py` files were added that heavily inflate the count. Many of these files contain only a handful of pieces.

The most important piece of code is the event handler. This takes an SQS event (as a Pydantic object) and does a significant amount of validation and conversion to make it into a format for our underlying state machine. This is the core of the whole event processing logic as it has to:
* Do various validations that the event makes sense (several of these validations are separated into a workflow service as we will do these before sending an event as well).
* Fetch entities/users and the workflow itself from the DB.
* Reshape the object into something a bit easier to use now that we've verified everything (eg. so the state machine won't need to do any not-null checks)

Effectively, the event handler is mostly a series of guard rails and error handling, which all is very important, and hopefully very rarely relevant to the development of the state machines.

The state machine that I have setup is very very barebones, future tickets will add emails and approval logic to it.

## Validation steps
All contained within the unit tests - no way to directly call it elsewhere at the moment. see the event handler logic for most of the core functionality testing.

### DIFF
--- a/api/src/workflow/handler/event_handler.py
+++ b/api/src/workflow/handler/event_handler.py
@@ -12,7 +12,11 @@ from src.workflow.base_state_machine import BaseStateMachine
 from src.workflow.event.state_machine_event import StateMachineEvent
 from src.workflow.event.workflow_event import WorkflowEvent
 from src.workflow.registry.workflow_registry import WorkflowRegistry
-from src.workflow.service.workflow_service import get_workflow_entities, is_event_valid_for_workflow
+from src.workflow.service.workflow_service import (
+    get_and_validate_workflow,
+    get_workflow_entities,
+    is_event_valid_for_workflow,
+)
 from src.workflow.workflow_config import WorkflowConfig
 from src.workflow.workflow_constants import WorkflowConstants
 from src.workflow.workflow_errors import (
@@ -20,7 +24,6 @@ from src.workflow.workflow_errors import (
     InvalidWorkflowTypeError,
     UnexpectedStateError,
     UserDoesNotExist,
-    WorkflowDoesNotExistError,
 )
 
 logger = logging.getLogger(__name__)
@@ -175,17 +178,9 @@ class EventHandler:
             )
             raise InvalidEventError("Process workflow event has a null process workflow context")
 
-        workflow = self.db_session.scalar(
-            select(Workflow).where(
-                Workflow.workflow_id == self.event.process_workflow_context.workflow_id
-            )
+        workflow = get_and_validate_workflow(
+            self.db_session, self.event.process_workflow_context.workflow_id, log_extra
         )
-
-        if workflow is None:
-            logger.warning("Workflow does not exist - cannot process event", extra=log_extra)
-            raise WorkflowDoesNotExistError(
-                "Workflow does not exist, cannot process events against it"
-            )
 
         config, state_machine_cls = self._get_state_machine_for_workflow_type(
             workflow.workflow_type
--- a/api/src/workflow/service/workflow_service.py
+++ b/api/src/workflow/service/workflow_service.py
@@ -1,4 +1,5 @@
 import logging
+import uuid
 from typing import Any
 
 from sqlalchemy import select
@@ -8,13 +9,16 @@ from src.constants.lookup_constants import WorkflowEntityType
 from src.db.models.base import ApiSchemaTable
 from src.db.models.competition_models import Application
 from src.db.models.opportunity_models import Opportunity
+from src.db.models.workflow_models import Workflow
 from src.workflow.base_state_machine import BaseStateMachine
 from src.workflow.event.workflow_event import WorkflowEntity
 from src.workflow.workflow_config import WorkflowConfig
 from src.workflow.workflow_errors import (
     EntityNotFound,
     ImplementationMissingError,
+    InactiveWorkflowError,
     InvalidEntityForWorkflow,
+    WorkflowDoesNotExistError,
 )
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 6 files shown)


api/src/workflow/handler/event_handler.py
```diff
@@ -0,0 +1,225 @@
...
+            workflow_id=uuid.uuid4(),
+            workflow_type=context.workflow_type,
+            # When initializing a workflow, grab the current state from
+            # the state machine class. Note the .value is needed as it
+            # otherwise won't realize the StrEnum is also a string and error.
+            current_workflow_state=state_machine_cls.initial_state.value,
+            is_active=True,
+            **workflow_entities
+        )
+        self.db_session.add(workflow)
```
kkrug, line 150: Maybe I missed this in the tech spec, but at what point is the final commit() handled?

api/src/workflow/service/workflow_service.py
```diff
@@ -0,0 +1,87 @@
...
+
+    workflow_entities: dict[str, list[ApiSchemaTable]] = {"opportunities": [], "applications": []}
+
+    for entity in entities:
+        log_extra |= {
+            "entity_type": entity.entity_type,
+            "entity_id": entity.entity_id,
+        }
+
+        if entity.entity_type == WorkflowEntityType.OPPORTUNITY:
```
kkrug, line 56: The README mentions updating this to add new entity types. Do we envision this occurring with any kind of frequency during future development? If so, maybe we could move the entity fetching logic to where we don't have to touch the core service logic often.

api/src/workflow/handler/event_handler.py
```diff
@@ -0,0 +1,225 @@
...
+    def _pre_process_existing_workflow_event(self) -> StateMachineEvent:
+        """Pre-process a process_workflow event.
+
+        Parse and validates that the information passed in makes sense
+        and is something we want to pass to the underlying state machine.
+
+        This includes:
+        * Validating that the process workflow context is present
+        * Fetching and verifying that the workflow exists in our DB
+        * Fetching the acting user
```
kkrug, line 170: Should we check if a workflow is active here? If an event comes in for a completed workflow, should we treat that as a NonRetryableWorkflowError immediately to avoid unnecessary instantiation?

api/tests/workflow/workflow_test_util.py
```diff
@@ -0,0 +1,97 @@
...
+    workflow_type: WorkflowType,
+    user: User | None,
+    entities: list,
+    exclude_start_workflow_context: bool = False,
+) -> WorkflowEvent:
+    user_id = user.user_id if user else uuid.uuid4()
+
+    entity_list = []
+    for entity in entities:
+        if isinstance(entity, Opportunity):
```
kkrug, line 50: Is there a reason why this only supports Opportunity currently and not also Application?

api/tests/workflow/workflow_test_util.py
```diff
@@ -0,0 +1,97 @@
...
+
+    return event
+
+
+def build_process_workflow_event(
+    workflow_id: uuid.UUID,
+    user: User | None,
+    event_to_send: str,
+    exclude_process_workflow_context: bool = False,
+):
```
kkrug: Should this also have a `-> WorkflowEvent` type?

kkrug (dismissed):

api/src/workflow/handler/event_handler.py
chouinar: It'll be handled above this layer of the code. A later PR will connect this to the SQS fetching/main loop of the code. https://github.com/HHS/simpler-grants-gov/blob/7e72033863c3e75760c3d355ac9463e94d66bb93/api/src/workflow/manager/workflow_manager.py#L184 is where that is right now.

All of this will eventually go into the `with db_session.begin()` of handle_event.

api/src/workflow/service/workflow_service.py
chouinar: It will be somewhat common, no idea how common though. I did separate this here out of the event handler into this service function as this will be used in the event API (along with the below function). Might make sense to move it to a different folder, but we can shuffle that later when we get to adding it to the API.

api/src/workflow/handler/event_handler.py
chouinar: At the moment it'd get an error from the workflow service complaining that it's an invalid event, but I should pull that up a bit to make sure the errors are right. Will adjust

api/tests/workflow/workflow_test_util.py
chouinar: I didn't fully setup the other entity types yet - figure we can add the remaining logic when needed.

api/tests/workflow/workflow_test_util.py
chouinar: I can add it - tests don't need to be typed (in some cases, the typing can be quite tricky), but I tend to add it as it makes IDE completion a bit better.

Issue #8005, opened 2026-01-15
Opportunity Table CRUD Endpoint- Create


brianlee-nava:
### Summary

**Background and Context:**
There are 4 main tables within the Opportunity Data Model that Plan and Publish will need to work with. Those are Opportunity, Opportunity Attachment, Opportunity Summary, and Opportunity Assistance Listing.

In order to make Opportunity Publishing work in Simpler Grants Management, we need to rebuild endpoints. The basic CRUD operations for each table account for 16 endpoints, with additional simple endpoints for publishing and querying the assistance listing table. This ticket is scoped for building the Create endpoint for the Opportunity table.

**User Story**
As a grantor system, I want to create a new Opportunity record so that a funding opportunity can be drafted and managed in Simpler Grants Management

### Acceptance criteria

Given a valid Opportunity payload
When the create Opportunity endpoint is called
Then a new Opportunity record is persisted
And the response returns a 201 status
And the response includes the Opportunity ID

Given a missing required field
When the create Opportunity endpoint is called
Then the response returns a 400 status
And the error message identifies the invalid field

dghazvini:
@Dao-REI @chouinar 

QQ - Based on the [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2612789251/Phase+1+Core+Foundations) on this part:

CRUD endpoints
NOTE: on the API documentation page, update the section title for the existing endpoints from “Opportunity v1” to “Opportunity v1 - for Applicants” and put these in a different section titled “Opportunity v1 - for Grantors”:

GET /v1/opportunities/:opportunity_id/grantor

POST /v1/opportunities/

PUT /v1/opportunities/:opportunity_id

--

Just want to make sure my understanding of the implementation is correct; we will create a new blueprint, route, and schema file that these new CRUD endpoints will consume, and the existing Opportunities v1 structure will later be renamed to be specified to be Opportunities v1 for Applicants.

<img width="375" height="274" alt="Image" src="https://github.com/user-attachments/assets/7a5a470a-d816-474c-bf32-846e801ace9e" />


And based on the convention I see, we also want to create a new directory under services for grantor opportunities? I just want to make sure that makes sense and that there won't potentially be a bunch of reused code we end up using.

<img width="396" height="564" alt="Image" src="https://github.com/user-attachments/assets/03c9de56-abc6-4e1c-b02f-02efc1a64428" />

chouinar:
[...]

PR #8498, opened 2026-02-11
8005 Opportunity Table Create Endpoint

dghazvini:
## Summary

Fixes for #8005 

## Changes proposed

Adding Create Endpoint for Creating new opportunities.

## Context for reviewers

[Tech Spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2612789251/Phase+1+Core+Foundations)

Opportunity Publishing in Simpler Grants will enable the creation of funding opportunities directly into Simpler Grants. With this capability, key opportunity data can be entered into Simpler Grants and displayed to applicants in a consistent, unified experience. This PR is focused on making the Create endpoint for creating a new opportunity.

Endpoint information:

`POST /v1/grantors/opportunities/`

Expected responses:

- 200 - Successful response (opportunity is created)
- 401 - Authentication error
- 403 - User does not have the `create_opportunity` for their assigned agency
- 404 - Agency ID doesn't exist
- 422 - User attempts to create opportunity with number that already exists/Missing Data

## Validation steps

- [x] 5 new Unit Tests pass
- [x] Confirmed all expected request+responses on local
- [x] Confirmed all information populates in local DB


api/src/api/opportunities_grantor_v1/opportunity_grantor_blueprint.py
```diff
@@ -0,0 +1,9 @@
+from apiflask import APIBlueprint
+
+opportunity_grantor_blueprint = APIBlueprint(
+    "opportunity_grantor_v1",
+    __name__,
+    tag="Opportunity v1 - for Grantors",
+    cli_group="opportunity_grantor_v1",
+    url_prefix="/v1/grantor",
```
chouinar: Any entity in a URL should be plural. https://restfulapi.net/resource-naming/ (among many other style guides) recommend naming with plural. While the grantor bit isn't quite the entity, looking at what we did for organization we did `organizations` so let's keep that consistent.

```suggestion
    url_prefix="/v1/grantors",
```

api/src/api/opportunities_grantor_v1/opportunity_grantor_routes.py
```diff
@@ -0,0 +1,187 @@
...
+__ALPHA VERSION__
+
+This endpoint in its current form is primarily for testing and feedback.
+
+Features in this endpoint are still under heavy development, and subject to change. Not for production use.
+
+See [Release Phases](https://github.com/github/roadmap?tab=readme-ov-file#release-phases) for further details.
+"""
+
+examples = {
```
chouinar: I think you copied this from the other file, this is just some examples for search requests that we add, you can delete this.

api/src/api/opportunities_grantor_v1/opportunity_grantor_routes.py
```diff
@@ -0,0 +1,187 @@
...
+    """Create a new opportunity"""
+    add_extra_data_to_current_request_logs(flatten_dict(json_data, prefix="request.body"))
+    logger.info("POST /v1/grantor/opportunities/")
+
+    with db_session.begin():
+        from flask_login import current_user
+
+        from src.services.opportunities_grantor_v1.opportunity_creation import create_opportunity
+
+        opportunity = create_opportunity(db_session, current_user, json_data)
```
chouinar: The flask current user isn't necessarily the user, don't use that. You can get the user directly from the auth. Do this.

```py
with db_session.begin():
    user = jwt_or_api_user_key_multi_auth.get_user()
    db_session.add(user)
```

That user is a User object from the user table, and is what we need for basically anything user related (including AuthZ). The `db_session.add()` is needed because the auth runs in a different DB session, so the record it fetched is not connected to the DB session so certain operations will fail. That just moves it over to this session.

Don't put imports in the middle of code, just put them at the top of the file. The few cases where we don't do that are usually for some circular dependency / import resolution issue (pretty rare).

api/src/api/opportunities_grantor_v1/opportunity_grantor_routes.py
```diff
@@ -0,0 +1,187 @@
...
+)
+@opportunity_grantor_blueprint.output(opportunity_grantor_schemas.OpportunityCreateResponseSchema())
+@jwt_or_api_user_key_multi_auth.login_required
+@opportunity_grantor_blueprint.doc(
+    responses=[200, 403, 404, 422, 500], security=jwt_or_api_user_key_security_schemes
+)
+@flask_db.with_db_session()
+def opportunity_create(db_session: db.Session, json_data: dict) -> response.ApiResponse:
+    """Create a new opportunity"""
+    add_extra_data_to_current_request_logs(flatten_dict(json_data, prefix="request.body"))
```
chouinar: We probably don't want to put all of the JSON data in our logs, we do that on the search endpoint as we want to track what folks are searching. We want to be careful as we do not want to log anything that could be considered sensitive or personal. While the data at the top-level of an opportunity isn't quite sensitive, we also don't want to add the description field to half a dozen logs.

Generally we just want to throw a few useful things like IDs. We don't quite have an opportunity ID because we add that later, so probably just the agency ID?

```suggestion
    add_extra_data_to_current_request_logs({"agency_id":  json_data.get("agency_id")})
```

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+class SearchResponseFormat(StrEnum):
+    JSON = "json"
+    CSV = "csv"
+
+
+class SearchQueryOperator(StrEnum):
+    AND = "AND"
+    OR = "OR"
```
chouinar: I think you copied a few search things over
```suggestion
```

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
```diff
@@ -0,0 +1,164 @@
...
+    )
+
+
+class OpportunityGrantorSchema(Schema):
+    """Schema for opportunity data in grantor-facing API responses
+
+    This schema defines the opportunity fields that are returned in grantor-facing API responses.
+    """
+
+    opportunity_id = fields.UUID(
```
chouinar: If at all possible, don't duplicate the entire schema here, we should try to reuse the schema from the primary GET opportunity endpoint, or if we want the endpoint to just return the opportunity itself and not everything under it, at least make it derived from the create schema + add a bit more. Note that validation is not done on responses, it just does its best to shape it into JSON from the schema (so if you had a validation for 50 characters max, it wouldn't error/would be fine if a field had 60 characters).

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
```diff
@@ -0,0 +1,164 @@
...
+        metadata={
+            "description": "The title of the opportunity",
+            "example": "Research Grant for Climate Innovation",
+        },
+    )
+    agency_id = fields.UUID(
+        required=True,
+        metadata={
+            "description": "The UUID of the agency that will own this opportunity",
+            "example": "add4b88f-e895-4ca9-92f4-38ed34055247",
```
chouinar: It's fine to include, but for any field with a very specific type like an enum or UUID, they actually have defaults based on their type (eg. an enum will pick one of the enum values, a UUID will be properly formatted). Can save a bit of time and skip those. Basic string fields are the main one to add examples for as they default to "string".

api/src/services/opportunities_grantor_v1/opportunity_creation.py
```diff
@@ -0,0 +1,37 @@
...
+        category=opportunity_data["category"],
+        category_explanation=opportunity_data.get("category_explanation"),
+        legacy_opportunity_id=None,
+        is_simpler_grants_opportunity=True,
+        is_draft=True,
+    )
+
+    db_session.add(opportunity)
+    # Push changes
+    db_session.flush()
```
chouinar, line 54: It doesn't end up meaningfully different, but can do this - I assume you did a flush so we'd have an opportunity_id to return.

```suggestion
    # Create the opportunity
    opportunity = Opportunity(
        opportunity_id=uuid.uuid4(),
        opportunity_number=opportunity_data["opportunity_number"],
        opportunity_title=opportunity_data["opportunity_title"],
        agency_id=agency.agency_id,
        agency_code=agency.agency_code,
        category=opportunity_data["category"],
        category_explanation=opportunity_data.get("category_explanation"),
        legacy_opportunity_id=None,
        is_simpler_grants_opportunity=True,
        is_draft=True,
    )

    db_session.add(opportunity)
```

api/src/services/opportunities_grantor_v1/opportunity_creation.py
```diff
@@ -0,0 +1,37 @@
...
+
+def create_opportunity(db_session: db.Session, user: User, opportunity_data: dict) -> Opportunity:
+    # Get agency and verify it exists
+    agency = get_agency(db_session, opportunity_data["agency_id"])
+
+    # Check if user has permission to create opportunities for this agency
+    verify_access(user, {Privilege.CREATE_OPPORTUNITY}, agency)
+
+    # Check if opportunity number already exists (raises 422 if it does)
+    check_opportunity_number_exists(db_session, opportunity_data["opportunity_number"])
```
chouinar: An idea - not a requirement - we sometimes prefer to make a pydantic object rather than work with the dictionary for ease of use:

```py
class OpportunityCreateRequest(BaseModel):
    agency_id: uuid.UUID
    opportunity_number: str
     # .. and so on
```

And then convert the dict into that - and use that so we can do `request.agency_id` instead. It's not strictly required, but it can make things a little nicer to work with.

Note that a future version of our API would actually let us define the API schemas in pydantic (but that's blocked at the moment by the APIFlask library needing to make a change).

api/src/api/opportunities_grantor_v1/opportunity_grantor_blueprint.py
dghazvini: Updated here and in the request/response schema and logger based on feedback.

api/src/api/opportunities_grantor_v1/opportunity_grantor_routes.py
dghazvini: Yeah I copied over the opportunity_v1 api and services directories and built off that, some of the unneeded code I still hadn't cleaned up... removed.

api/src/api/opportunities_grantor_v1/opportunity_grantor_routes.py
dghazvini: Updated based on feedback.

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
dghazvini: Yes; removed

api/src/api/opportunities_grantor_v1/opportunity_grantor_routes.py
dghazvini: Understood, yeah this block was just some temporary workaround I was working with while I didn't have proper authentication set up on the endpoint, did intend to move/remove it. Revised based on feedback.

api/src/services/opportunities_grantor_v1/opportunity_creation.py
dghazvini: I wasn't exactly sure when `db_session.flush()` is necessary; so if we explicitly set all required values it's not necessary then. So SQLAlchemy automatically flushes the session when the commit is done it looks like. Updated based on feedback.

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
dghazvini: Removed some of the unneeded example fields for UUID, enum, and bool.

api/src/services/opportunities_grantor_v1/opportunity_creation.py
dghazvini: Made updates to `get_opportunity.py` and `opportunity_creation.py` to reflect this.

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
dghazvini: Agree with this; noticed OpportunityV1Schema already has all the fields we need except for is_draft. I had some issues trying to implement this though...
Initially tried this in `opportunity_grantor_schemas`:

```
class OpportunityGrantorSchema(OpportunityV1Schema):
    
     is_draft = fields.Boolean(
         metadata={
             "description": "Whether the opportunity is a draft",
         },
     )
```

But was receiving errors about lazy loading after opportunity creation for several fields (it DID perform the insertion but is returning a 500 error):
`sqlalchemy.orm.exc.DetachedInstanceError: Parent instance <Opportunity at 0x75462596b390> is not bound to a Session; lazy load operation of attribute 'agency_record' cannot proceed `

I'm assuming the logic we should use is similar to what is being used in `in_get_opportunity` in the opportunities_v1 service layer (I see options for lazyload being used there)? Not entirely sure.

dghazvini:
Opened PR for full Review.

Notes:
- I added asserts for each field in the test case for a successful opportunity creation; probably overkill so any guidance on what constitutes good enough would be great.
- Not sure if you already committed what you mentioned for the util for setting up a user in a particular agency with a specific role, if so I can use that @chouinar 
- Any input on where we might want to log more information than what's already currently in the endpoint is appreciated.

api/tests/lib/opportunity_test_utils.py
```diff
@@ -0,0 +1,109 @@
...
+        agency_id: UUID of existing agency to use (creates new one if None)
+        privileges: List of privileges to grant the user
+        email: Optional email address for the user
+        **kwargs: Additional arguments passed to factory creation
+
+    Returns:
+        tuple: (user, agency, token, api_key_id) - The created user, agency, JWT token, and API key ID
+    """
+    # Initialize factory db_session
+    factories._db_session = db_session
```
chouinar: Don't initialize the factory db session like this, any unit test using the factories should add `enable_factory_create` - we do that so you need to be intentional that you want the factories to write to the DB (they can also be used with the `.build()` function to just create an object).

api/tests/lib/opportunity_test_utils.py
```diff
@@ -0,0 +1,109 @@
...
+    AgencyFactory,
+    AgencyUserFactory,
+    AgencyUserRoleFactory,
+    RoleFactory,
+    UserApiKeyFactory,
+    UserFactory,
+)
+
+
+def create_user_with_agency_privileges(
```
chouinar: Here is what I created - mine is tied up in a larger PR, could break it out, but also fine to copy what I have.

Notably, I'd recommend not trying to fetch an agency like you have, either have it take in an agency (if you want to setup other stuff first on it) or make one for you. Any ID lookup shouldn't be necessary for our unit tests since you'd almost always make the agency as part of that unit test.

I do like that you added the API key stuff, although like what I did for JWT here, make it a separate function as some tests don't need the JWT/key.

```py
from src.adapters import db
from src.auth.api_jwt_auth import create_jwt_for_user
from src.constants.lookup_constants import Privilege
from src.db.models.user_models import Role, User
from tests.src.db.models.factories import (
    AgencyFactory,
    AgencyUserFactory,
    AgencyUserRoleFactory,
    LinkExternalUserFactory,
    RoleFactory,
    UserFactory,
    UserProfileFactory,
)


def create_user_in_agency(
    agency: Agency | None = None,
    role: Role | None = None,
    privileges: list[Privilege] | None = None,
    email: str | None = None,
    first_name: str | None = None,
    last_name: str | None = None,
) -> tuple[User, Agency]:
    user = UserFactory.create()

    link_kwargs = {"user": user}
    if email is not None:
        link_kwargs["email"] = email
    LinkExternalUserFactory.create(**link_kwargs)

    # Create user profile if first_name or last_name provided
    if first_name is not None or last_name is not None:
        profile_kwargs = {"user": user}
        if first_name is not None:
            profile_kwargs["first_name"] = first_name
        if last_name is not None:
            profile_kwargs["last_name"] = last_name
        UserProfileFactory.create(**profile_kwargs)

    if agency is None:
        agency = AgencyFactory.create()

    agency_user = AgencyUserFactory.create(user=user, agency=agency)

    if privileges:
        role = RoleFactory.create(privileges=privileges, is_agency_role=True)

    if role is not None:
        AgencyUserRoleFactory.create(agency_user=agency_user, role=role)

    return user, agency


def create_user_in_agency_with_jwt(
    db_session: db.Session,
    agency: Agency | None = None,
    role: Role | None = None,
    privileges: list[Privilege] | None = None,
    email: str | None = None,
    first_name: str | None = None,
    last_name: str | None = None,
) -> tuple[User, Agency, str]:
    user, agency = create_user_in_agency(
        agency=agency,
        role=role,
        privileges=privileges,
        email=email,
        first_name=first_name,
        last_name=last_name,
    )

    token, _ = create_jwt_for_user(user, db_session)
    db_session.commit()  # commit to push JWT to DB

    return user, agency, token

```

api/tests/lib/opportunity_test_utils.py
```diff
@@ -0,0 +1,109 @@
...
+        agency_id = str(uuid.uuid4())
+
+    # Ensure agency_id is a string, not a UUID object
+    if not isinstance(agency_id, str):
+        agency_id = str(agency_id)
+
+    request = {
+        "opportunity_number": opportunity_number,
+        "opportunity_title": opportunity_title,
+        "agency_id": agency_id,
```
chouinar: Slightly lazier way, just always make it a string. 

```suggestion
    if not agency_id:
        agency_id = uuid.uuid4()
    
    request = {
        "opportunity_number": opportunity_number,
        "opportunity_title": opportunity_title,
        "agency_id": str(agency_id),
```

api/tests/lib/opportunity_test_utils.py
```diff
@@ -0,0 +1,109 @@
...
+        agency_id: UUID of the agency (as string)
+        **kwargs: Additional fields to include in the request
+
+    Returns:
+        dict: A valid opportunity creation request
+    """
+
+    opportunity_number = "TEST-2026-001"
+    opportunity_title = "Research Grant for Climate Innovation"
+    category_explanation = "Competitive research grant"
```
chouinar: Move the defaults to the function that way you can easily override them.

Also, might suggest that opportunity number not get a default since it'll conflict if you have two tests both with it set the same.

```suggestion
def create_opportunity_request(
    agency_id=None,
    opportunity_number = "TEST-2026-001",
    opportunity_title = "Research Grant for Climate Innovation",
    category =  OpportunityCategory.DISCRETIONARY,
    category_explanation = "Competitive research grant",
    ):
    """Create a valid opportunity creation request.

    Args:
        agency_id: UUID of the agency (as string)
        **kwargs: Additional fields to include in the request

    Returns:
        dict: A valid opportunity creation request
    """


```

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
```diff
@@ -0,0 +1,147 @@
+from src.api.schemas.extension import Schema, fields, validators
+from src.api.schemas.response_schema import AbstractResponseSchema, FileResponseSchema
+from src.constants.lookup_constants import OpportunityCategory
+
+
+class OpportunityAttachmentV1Schema(FileResponseSchema):
```
chouinar: I don't think this gets used

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
chouinar: The lazyload is related, but not quite what you want.

An explanation. Let's assume I have this bit of code:

```
with db_session.begin():
    opportunity = db_session.execute(select(Opportunity)) ...
    
    # this would work
    print(opportunity.opportunity_attachments)

# this would give a detached instance error
db_session.close() # this happens some time after we we return a response
print(opportunity.opportunity_assistance_listings)
```

There's nothing different about the two relationships, but if a SQLAlchemy model is no longer attached to a session (like what happens once that session gets cleaned up after we return a response), it can't be used to fetch any relationships. All relationships by default for SQLAlchemy are lazily loaded. That is, doing `opportunity.opportunity_attachments` emits a select statement to the DB as part of getting that list.

Well, if the DB session has finished, it can't emit a select statement. The error complains about a detached state because the opportunity object isn't attached to a session anymore.

How do we fix this? Tell SQLAlchemy to not lazy-load the variables. If you look at what get opportunity does, it fetches a whole bunch of extra stuff (albeit, it's not well implemented, we should redo it to be more explicit).

---
However, in this case, I'm not 100% certain of how we should do that, we could query the DB again for the opportunity that we already have to load everything, or we could do something like:
```py
SCHEMA =  OpportunityV1Schema()
# this'll make the JSON, but won't do anything with it. Importantly, it'll cause SQLAlchemy to fetch everything. Note we can't return the JSON since otherwise it'll try to convert the JSON into JSON and run into a bunch of typing issues.
SCHEMA.dump(opportunity)
```

api/src/services/opportunities_grantor_v1/get_opportunity.py
```diff
@@ -0,0 +1,27 @@
...
+from src.constants.lookup_constants import OpportunityCategory
+from src.db.models.opportunity_models import Opportunity
+
+
+class OpportunityCreateRequest(BaseModel):
+    agency_id: uuid.UUID
+    opportunity_number: str
+    opportunity_title: str
+    category: OpportunityCategory
+    category_explanation: str | None = None
```
chouinar: Could we put this in the opportunity creation file where it is used?

api/src/services/opportunities_grantor_v1/opportunity_creation.py
```diff
@@ -0,0 +1,43 @@
...
+        agency_code=agency.agency_code,
+        category=request.category,
+        category_explanation=request.category_explanation,
+        legacy_opportunity_id=None,
+        is_simpler_grants_opportunity=True,
+        is_draft=True,
+    )
+
+    db_session.add(opportunity)
+
```
chouinar, line 70: Useful for debugging/metrics to add a message at the end of one of these service functions. That way when product asks "how many / which opportunities were created", we've got a really easy log message to look for.

```suggestion

    db_session.add(opportunity)
    
    logger.info("Created opportunity", extra={"opportunity_id": opportunity.opportunity_id, "agency_code": agency.agency_code})

```

dghazvini: Added

api/src/services/opportunities_grantor_v1/get_opportunity.py
dghazvini: Moved to opportunity_creation.py

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
dghazvini: It's currently being used in the request and response schema on lines 64 and 106:

```
    category = fields.Enum(
        OpportunityCategory,
        required=True,
        metadata={
            "description": "The opportunity category",
        },
    )
```

api/tests/lib/opportunity_test_utils.py
dghazvini: Updated based on feedback. Removed default assignment to opportunity number and added a check to generate a number if no value is assigned (after the agency id check).

api/tests/lib/opportunity_test_utils.py
dghazvini: Adjusted based on feedback

api/tests/lib/opportunity_test_utils.py
dghazvini: Copied over the create_user_in_agency and create_user_in_agency_with_jwt functions, as well as made a create_user_in_agency_with_jwt_and_api_key function for using both.
Added a test agency fixture in test_opportunity_route_create.py. and updated the other fixture/test to take that in instead of fetching an agency.

api/tests/lib/opportunity_test_utils.py
dghazvini: Removed the initialize

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
dghazvini: Based on what I read and tried, was able to get it working properly using this setup; see if this makes sense.

First added what I mentioned before in ```opportunity_grantor_schemas```:
```
class OpportunityGrantorSchema(OpportunityV1Schema):
    
     is_draft = fields.Boolean(
         metadata={
             "description": "Whether the opportunity is a draft",
         },
     )
```

Then in ```create_opportunity``` in ```opporunity_creation.py```, made the following updates:

```
    db_session.add(opportunity)
    db_session.flush()

    # Reload the opportunity with all necessary relationships
    opportunity = (
        db_session.query(Opportunity)
        .options(
            joinedload(Opportunity.agency_record).joinedload(Agency.top_level_agency),
            joinedload(Opportunity.opportunity_attachments),
            joinedload(Opportunity.opportunity_assistance_listings),
            joinedload(Opportunity.current_opportunity_summary),
            joinedload(Opportunity.all_opportunity_summaries),
            joinedload(Opportunity.competitions),
        )
        .filter(Opportunity.opportunity_id == opportunity.opportunity_id)
        .one()
    )
```

The ```joinloaded()``` function is meant to tell SQLAlchemy to eager load rather than lazy load later. ```db_session.flush()``` sets the opportunity ID and is meant to ensure the state before querying again with these joinloads (then we query with that ID).

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
chouinar: I'm not seeing this schema below. This is an attachment schema, your example is the opportunity category?

api/src/services/opportunities_grantor_v1/opportunity_creation.py
chouinar: Yeah, there's probably more to it than this, but with SQLAlchemy an object can be in a few states:

```python
# At this point, it's not in the session, it's just a python object
# This is transient
opportunity = Opportunity(...)

# This is now pending. It's in SQLAlchemy's cache, but not in the DB
db_session.add(opportunity)

# This moves the object to the DB - note that until you commit this is all
# in a transaction so the DB doesn't actually have this
db_session.flush()

# if you load a value from the DB, it's also persistent.
other_opp = db_session.execute(select(Opportunity)) ...
```

Note that a flush happens in a few cases:
* Explicitly by calling `flush`
* As part of a commit
* SQLAlchemy sometimes just does it as part of other operations - usually if the cache is growing a bunch, it'll flush - this isn't predictable

As far as we care, a flush does the following:
* All default values get set (we mostly only use defaults for primary keys) - note these are not DB defaults, they're still all in python.
* It sometimes will cause an error if you did something wrong - this can present on random lines of code. If I did a `db_session.add()` on an object that had a unique constraint that it was violating, it wouldn't fail until later.

More info
https://docs.sqlalchemy.org/en/21/orm/session_state_management.html#quickie-intro-to-object-states

api/src/services/opportunities_grantor_v1/opportunity_creation.py
```diff
@@ -0,0 +1,75 @@
...
+        .options(
+            joinedload(Opportunity.agency_record).joinedload(Agency.top_level_agency),
+            joinedload(Opportunity.opportunity_attachments),
+            joinedload(Opportunity.opportunity_assistance_listings),
+            joinedload(Opportunity.current_opportunity_summary),
+            joinedload(Opportunity.all_opportunity_summaries),
+            joinedload(Opportunity.competitions),
+        )
+        .filter(Opportunity.opportunity_id == opportunity.opportunity_id)
+        .one()
```
chouinar: Don't use `query`, use execute, query is the legacy way, we don't want to use it going forward.

---

Don't use joinedload. It's got a lot of issues including:
* It does not handle uniqueness very well and you will end up with duplicate data if that's possible on a join - there are workarounds but it's a lot of complexity to support
* Large queries can and will lock up

selectinload does not have these issues and produces very simple queries under the hood without any sort of scaling problems. Docs will say that it's slower than joinedload in certain cases, but it's minimal (sub-millisecond) for any cases we've encountered that I've tested.

api/tests/src/api/opportunities_grantors_v1/test_opportunity_route_create.py
```diff
@@ -0,0 +1,138 @@
...
+@pytest.fixture
+def test_agency(db_session, enable_factory_create):
+    """Create test agency"""
+    return AgencyFactory.create(agency_id=uuid.uuid4())
+
+
+@pytest.fixture
+def grantor_auth_data(db_session, enable_factory_create, test_agency):
+
+    agency = test_agency
```
chouinar: The fixture is just a single line factory call. Guessing an earlier iteration did a bit more in it, but can trim that out. The way the`create_user_in_agency_with_jwt_and_api_key` util works, you can even just leave the agency out and let it create it.

```suggestion
@pytest.fixture
def grantor_auth_data(db_session, enable_factory_create):

    agency = AgencyFactory.create()
```

api/tests/lib/opportunity_test_utils.py
```diff
@@ -0,0 +1,140 @@
...
+        agency_id = uuid.uuid4()
+
+    if not opportunity_number:
+        opportunity_number = f"TEST-{uuid.uuid4().hex[:3]}"
+
+    request = {
+        "opportunity_number": opportunity_number,
+        "opportunity_title": opportunity_title,
+        "agency_id": str(agency_id),
+        "category": OpportunityCategory.DISCRETIONARY,
```
chouinar: Think you forgot to connect this.
```suggestion
        "category": category,
```

dghazvini: Yes, good catch 🙃 Fixed it.

api/tests/src/api/opportunities_grantors_v1/test_opportunity_route_create.py
dghazvini: Got it; yeah I suppose it makes more sense to just do it in ```grantor_auth_data```. Updated based on feedback and removed references to the previous fixture.

api/src/services/opportunities_grantor_v1/opportunity_creation.py
dghazvini: Updated based on feedback and verified functionality with ```selectinload``` and ```execute```

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
dghazvini: My bad, this was a holdover from when i copied the file over from ```opportunity_schemas.py```. 🙃 Somehow misread this twice. Removed it.

PR #8501, opened 2026-02-11
[Issue #8405] Seed agency users and roles

### LINKED ISSUE #8405: seed agency users and opportunity roles
### Summary

related to #8333 

### Acceptance criteria

- [ ] create seed script to assign users to agencies and roles

Dao-REI:
## Summary

Fixes #8405

## Changes proposed

Added agency users with roles for opportunity editing and/or publishing. This is dummy/test data for local and dev environments.

## Context for reviewers

1. Added support for two roles to AgencyUserFactory
2. Updated UserBuilder (seed_data_util.py) to support user to agency association with roles
3. Updated the seed_agency.py file to create some users associated with a few agencies with some roles.

## Validation steps

1. Review the code
2. Run make setup-api-data to populate the database (note: if your DB is not up to date, run make volume-recreate first.)
3. Check that the database tables have been populated.
- select * from api.agency_user;
- select * from api.agency_user_role;
- Agencies are in this table: select * from api.agency;

### DIFF
--- a/api/tests/lib/seed_agencies.py
+++ b/api/tests/lib/seed_agencies.py
@@ -1,12 +1,9 @@
 # Called by seed_local_db when running `make seed-db-local`
 import logging
-import uuid
 
 import src.adapters.db as db
 import tests.src.db.models.factories as factories
-from src.constants.static_role_values import OPPORTUNITY_EDITOR, OPPORTUNITY_PUBLISHER
 from src.db.models.agency_models import Agency
-from tests.lib.seed_data_utils import UserBuilder
 
 logger = logging.getLogger(__name__)
 
@@ -48,11 +45,6 @@ AGENCIES_TO_CREATE = [
         "agency_code": "NSF",
         "agency_id": "16945d39-1564-479c-b438-ef8d0804f051",
         "agency_name": "National Science Foundation",
-    },
-    {
-        "agency_code": "USAID",
-        "agency_id": "094f7d5c-afe6-4e40-823b-d830076e9144",
-        "agency_name": "Agency for International Development",
     },
     {
         "agency_code": "USDA",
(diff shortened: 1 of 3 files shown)


api/tests/lib/seed_agencies.py
```diff
@@ -462,10 +466,135 @@ def _build_agencies(db_session: db.Session) -> None:
...
+    (
+        UserBuilder(
+            uuid.UUID("25dea202-fac8-48f6-ac52-0ec06d7176e0"),
+            db_session,
+            "user with one agency with the opportunity editor role",
+        )
+        .with_oauth_login("one_agency_opp_edit")
+        .with_api_key("one_agency_opp_edit_key")
+        .with_jwt_auth()
+        .with_agency(agencies_created[0], roles=[OPPORTUNITY_EDITOR])
```
chouinar: I'd suggest rather than setup agencies from the list of generated ones we setup a set of statically defined ones, similar to what we do for orgs in `_build_organizations_and_users` otherwise the agencies will change every time you run this and actually users would just keep getting added to more and more agencies.

Rather than passing agencies into this function, we just do something like:

```py
def setup_agency(db_session: db.Session, agency_id: uuid.UUID, agency_code: str, top_level_agency: Agency | None = None):
    agency = db_session.execute(select(Agency).where(Agency.agency_id == agency_id).scalar_one_or_none()

    if agency is None:
        # pass whatever else you want here as well.
        agency = AgencyFactory.create(agency_id=agency_id, agency_code=agency_code, top_level_agency=top_level_agency)

    else:
         agency.agency_code = agency_code
         agency.top_level_agency = top_level_agency
         # whatever else you want to set

     return agency
```

Then we can use it like:
```
agency_top_level = setup_agency(db_session, uuid.UUID(".. something"), "EXAMPLE")

agency_sub = setup_agency(db_session, uuid.UUID(".. something"), "EXAMPLE-123", top_level_agency=agency_top_level)
```

Dao-REI: @chouinar Actually, these Agencies are not generated but statically defined. 

```
AGENCIES_TO_CREATE = [
    {
        "agency_code": "DOC",
        "agency_id": "add4b88f-e895-4ca9-92f4-38ed34055247",
        "agency_name": "Department of Commerce",
    },
    { ...
```

And every time it is executed, it checks if they are already in the DB. Similarly with my code for agency users, it will not add new ones if they already exist.

chouinar: The way the logic is written, it isn't going to be the same agency every time. The first time it runs, it'll create the agencies, and that list is the order. The second time it runs, it'll fetch an unsorted list of agencies from the DB. Those won't be the same. If you run this twice, a user will be in two agencies despite the config saying they should be in 1.

Let me flip this around a bit - can I easily look at this code and see "what agency did we assign to this user"?

If when we later want to setup more on these agencies (creating opportunities in various states) what agency is `agencies_create[3]`? You can count through the list in the file, but that's painful to maintain - and as above, isn't the same list on every run.

This seed script is complex, and has a lot of connections across different pieces, we don't just want "some agency" on these, we want specific ones (e2e tests would need to change it like I'm suggesting).

Dao-REI: I see. I didn't realize that it'll fetch an unsorted list of agencies from the DB on the second run. Okay, then I'll create a separate file for this in case we want to setup more test data associated with agencies in the future. (seed_agencies_and_users.py)

Dao-REI: @chouinar Updates completed. Re-review please.

chouinar (approved):

Issue #8382, opened 2026-02-04
Implement functionality to send request to new URL with new auth


freddieyebra:
### Summary

Utilize the method from [auth functionality implementation](https://github.com/HHS/simpler-grants-gov/issues/8379) to build and send a request to the new partner gateway url. The result of the response should be logged in local environment only. This can be achieved by utilizing the `log_local()` method. Note, that this should only be invoked with the incoming request if the `use_soap_jwt` is set.

### Acceptance criteria

- [ ] Unit tests pass
- [ ] Able to invoke locally

PR #8506, opened 2026-02-12
[Issue #8382] Locally log proxy response from soap jwt path

jakobpederson:
## Summary

Fixes / Work for #8382 

## Changes proposed

I added a new way to get the `proxy_url` based on the `use_soap_jwt` parameter. It gets the url from the `full_path` on the request and the `SOAP_PARTNER_GATEWAY_URI` secret on the parameter store. I also added a `log_local` which logs the response data.

## Context for reviewers

As part of transitioning to the new soap jwt auth we need to use a slightly different `proxy_url` and local logging is very helpful for future local testing going forward.

Example of local logging:
```
grants-api             | 21:47:02.612  src.legacy_soap_api.legacy_soap_api_utils log_local                    INFO
grants-api             | soap jwt proxy response:
grants-api             | --uuid:a0c20d08-4fe1-4ad9-b7a5-4bb903188514
grants-api             | Content-Type: application/xop+xml; charset=UTF-8; type="text/xml"
grants-api             | Content-Transfer-Encoding: binary
grants-api             | Content-ID: <root.message@cxf.apache.org>
grants-api             |
grants-api             | <soap:Envelope xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/"><soap:Body><ns2:GetSubmissionListExpandedResponse xmlns:ns12="http://schemas.xmlsoap.org/wsdl/soap/" xmlns:ns11="http:/
/schemas.xmlsoap.org/wsdl/" xmlns:ns10="http://apply.grants.gov/system/GrantsFundingSynopsis-V2.0" xmlns:ns9="http://apply.grants.gov/system/AgencyUpdateApplicationInfo-V1.0" xmlns:ns8="http://apply.grants.gov/syst
em/GrantsForecastSynopsis-V1.0" xmlns:ns7="http://apply.grants.gov/system/AgencyManagePackage-V1.0" xmlns:ns6="http://apply.grants.gov/system/GrantsPackage-V1.0" xmlns:ns5="http://apply.grants.gov/system/GrantsOppo
rtunity-V1.0" xmlns:ns4="http://apply.grants.gov/system/GrantsRelatedDocument-V1.0" xmlns:ns3="http://apply.grants.gov/system/GrantsTemplate-V1.0" xmlns:ns2="http://apply.grants.gov/services/AgencyWebServices-V2.0"
 xmlns="http://apply.grants.gov/system/GrantsCommonElements-V1.0"><ns2:Success>false</ns2:Success><ErrorDetails><Code>SYS-1001</Code></ErrorDetails><ns2:AvailableApplicationNumber>0</ns2:AvailableApplicationNumber>
</ns2:GetSubmissionListExpandedResponse></soap:Body></soap:Envelope>
grants-api             | --uuid:a0c20d08-4fe1-4ad9-b7a5-4bb903188514--
```

### DIFF
--- a/api/src/legacy_soap_api/legacy_soap_api_proxy.py
+++ b/api/src/legacy_soap_api/legacy_soap_api_proxy.py
@@ -9,6 +9,7 @@ from requests import Request, Session
 
 from src.db.models.user_models import LegacyCertificate
 from src.legacy_soap_api.legacy_soap_api_auth import (
+    LOG_LOCAL_RESPONSE_HEADER_KEY,
     MTLS_CERT_HEADER_KEY,
     S2S_PARTNER_CERTID_JWT_B64_HEADER_KEY,
     USE_SOAP_JWT_HEADER_KEY,
@@ -40,6 +41,7 @@ def get_proxy_response(soap_request: SOAPRequest, timeout: int = PROXY_TIMEOUT) 
 
     soap_auth = soap_request.auth
     use_soap_jwt = soap_request.headers.get(USE_SOAP_JWT_HEADER_KEY) == "1"
+    should_log_response = soap_request.headers.get(LOG_LOCAL_RESPONSE_HEADER_KEY) == "1"
     if use_soap_jwt and soap_auth and soap_auth.certificate.legacy_certificate:
         proxy_headers = {
             S2S_PARTNER_CERTID_JWT_B64_HEADER_KEY: get_soap_jwt_auth_jwt(
@@ -66,7 +68,7 @@ def get_proxy_response(soap_request: SOAPRequest, timeout: int = PROXY_TIMEOUT) 
             extra={"soap_api_event": LegacySoapApiEvent.CALLING_WITHOUT_CERT},
         )
         response = _get_soap_response(_request, timeout=timeout)
-        if use_soap_jwt:
+        if use_soap_jwt and should_log_response:
             log_local(
                 msg="soap jwt proxy response",
                 data=response.to_bytes().decode("utf-8"),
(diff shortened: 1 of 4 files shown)


api/src/legacy_soap_api/legacy_soap_api_proxy.py
```diff
@@ -63,7 +65,13 @@ def get_proxy_response(soap_request: SOAPRequest, timeout: int = PROXY_TIMEOUT)
...
             extra={"soap_api_event": LegacySoapApiEvent.CALLING_WITHOUT_CERT},
         )
-        return _get_soap_response(_request, timeout=timeout)
+        response = _get_soap_response(_request, timeout=timeout)
+        if use_soap_jwt:
+            log_local(
+                msg="soap jwt proxy response",
+                data=response.to_bytes().decode("utf-8"),
+            )
+        return response
```
chouinar: If the response is a very large response, this'll be very slow - this also will run even if you're not local because you're converting the value before passing it into the function. If you're going to add anything like this, the order of operations needs to be changed a bit.

Maybe something like a header param you pass in to turn on the logging?
```py
# this is turned on by a configuration, like a header
should_log_response = soap_request.headers.get("Log-Local-Response")
if should_log_response:
    log_local(...) # same as now
```

That way it only happens when you intend for it to / testing an endpoint with a big response doesn't always convert even locally.

jakobpederson: I'll throw that parameter on. The response can be very large to the point that the logs won't be useful (like asking for GetApplication). This will mean this log is only really helpful for some the endpoints.

chouinar (approved):

Issue #8267, opened 2026-01-30
Update backend to support PIV Required Login


glenblosser-nava:
### Summary

Look at the **Implementation** section for details
https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2265940001/Tech+Spec+Grantors+Require+PIV+Login#Implementation

### Acceptance criteria

- [ ] When an Agency user attempts to log in without a PIV, sends error to frontend callback
- [ ] When a non-Agency user attempts to log in without a PIV, login succeeds
- [ ] When an Agency user attempts to log in with a PIV, login succeeds

brianlee-nava:
@desaiamit This ticket was brought up, but we didn't get to story point it. Can you please assign a story point value? When you're finished with this ticket, please ping @glenblosser-nava so Glen can finish #7142

desaiamit:
@brianlee-nava Assigned the story points

PR #8512, opened 2026-02-12
[Issue #8267] Add PIV/CAC login requirement for agency users

desaiamit:
## Summary

Work for #8267

## Changes proposed

- Add `x509:presented` to login.gov scope to detect PIV/CAC authentication
- Add `IS_PIV_REQUIRED` config flag (default: false) for environment control
- Add `x509_presented` field to `LoginGovUser` dataclass
- Extract `x509_presented` from login.gov token payload
- Add `_validate_piv_requirement()` function to validate agency users used PIV
- Eager load `agency_users` relationship for PIV validation
- Add PIV validation tests

## Context for reviewers

When `IS_PIV_REQUIRED=true`, agency users (users with `agency_users` records) must authenticate using PIV/CAC. Non-agency users are unaffected.

The validation happens after user lookup but before JWT generation in the login callback flow. If an agency user attempts to login without PIV when required, they receive a 422 error that redirects to the frontend with the message "Agency users must authenticate using a PIV/CAC card".

Default behavior (`IS_PIV_REQUIRED=false`) allows all users to login without PIV - safe for dev/staging environments.

## Validation steps

1. Run tests: `make test args="tests/src/api/users/test_user_route_login.py -v -k piv"`
2. All 4 PIV tests should pass:
   - `test_agency_user_without_piv_fails_when_required`
   - `test_agency_user_with_piv_succeeds_when_required`
   - `test_non_agency_user_without_piv_succeeds_when_required`
   - `test_agency_user_without_piv_succeeds_when_not_required`

### DIFF
--- a/api/src/services/users/login_gov_callback_handler.py
+++ b/api/src/services/users/login_gov_callback_handler.py
@@ -113,23 +113,36 @@ def _validate_piv_requirement(user: User, x509_presented: bool | None) -> None:
     """
     config = get_config()
 
-    # Skip validation if PIV is not required (lower environments)
-    if not config.is_piv_required:
-        return
-
     # Check if user is an agency user
     is_agency_user = len(user.agency_users) > 0
 
-    # If user is an agency user and didn't use PIV, reject login
+    # If user is an agency user and didn't use PIV, reject or log
     if is_agency_user and not x509_presented:
+        if config.is_piv_required:
+            logger.info(
+                "Agency user attempted login without PIV",
+                extra={
+                    "user_id": user.user_id,
+                    "x509_presented": x509_presented,
+                },
+            )
+            raise_flask_error(422, "Agency users must authenticate using a PIV/CAC card")
+        else:
+            logger.info(
+                "Agency user login would have been blocked if PIV were required",
+                extra={
+                    "user_id": user.user_id,
+                    "x509_presented": x509_presented,
+                },
+            )
+    elif is_agency_user and x509_presented:
         logger.info(
-            "Agency user attempted login without PIV",
+            "Agency user logged in with PIV",
             extra={
-                "user_id": str(user.user_id),
+                "user_id": user.user_id,
                 "x509_presented": x509_presented,
             },
         )
-        raise_flask_error(422, "Agency users must authenticate using a PIV/CAC card")
 
 
 def handle_login_gov_token(
--- a/api/tests/src/api/users/test_user_route_login.py
+++ b/api/tests/src/api/users/test_user_route_login.py
@@ -11,7 +11,12 @@ from src.auth.api_jwt_auth import parse_jwt_for_user
 from src.db.models.user_models import LinkExternalUser, LoginGovState
 from src.util import datetime_util
 from tests.lib.auth_test_utils import create_jwt
-from tests.src.db.models.factories import LinkExternalUserFactory, LoginGovStateFactory
+from tests.src.db.models.factories import (
+    AgencyFactory,
+    AgencyUserFactory,
+    LinkExternalUserFactory,
+    LoginGovStateFactory,
+)
 
 ##########################################
 # Full login flow tests
@@ -511,12 +516,6 @@ def test_agency_user_without_piv_fails_when_required(
     client, db_session, enable_factory_create, mock_oauth_client, private_rsa_key, monkeypatch
 ):
     """Agency user logging in without PIV should fail when IS_PIV_REQUIRED=true"""
-    from tests.src.db.models.factories import (
-        AgencyFactory,
-        AgencyUserFactory,
-        LinkExternalUserFactory,
-    )
-
     # Enable PIV requirement
     monkeypatch.setattr("src.auth.login_gov_jwt_auth._config.is_piv_required", True)
 
@@ -558,12 +557,6 @@ def test_agency_user_with_piv_succeeds_when_required(
     client, db_session, enable_factory_create, mock_oauth_client, private_rsa_key, monkeypatch
 ):
     """Agency user logging in with PIV should succeed when IS_PIV_REQUIRED=true"""
-    from tests.src.db.models.factories import (
-        AgencyFactory,
-        AgencyUserFactory,
-        LinkExternalUserFactory,
-    )
-
     # Enable PIV requirement
     monkeypatch.setattr("src.auth.login_gov_jwt_auth._config.is_piv_required", True)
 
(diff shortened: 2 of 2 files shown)


api/tests/src/api/users/test_user_route_login.py
```diff
@@ -500,3 +500,184 @@ def test_user_callback_token_fails_validation_no_valid_key_302(
...
+
+def test_agency_user_without_piv_fails_when_required(
+    client, db_session, enable_factory_create, mock_oauth_client, private_rsa_key, monkeypatch
+):
+    """Agency user logging in without PIV should fail when IS_PIV_REQUIRED=true"""
+    from tests.src.db.models.factories import (
+        AgencyFactory,
+        AgencyUserFactory,
+        LinkExternalUserFactory,
+    )
```
chouinar: Put any imports at the top of the file.

api/src/services/users/login_gov_callback_handler.py
```diff
@@ -101,6 +101,37 @@ def handle_login_gov_callback_request(
...
+    # If user is an agency user and didn't use PIV, reject login
+    if is_agency_user and not x509_presented:
+        logger.info(
+            "Agency user attempted login without PIV",
+            extra={
+                "user_id": str(user.user_id),
+                "x509_presented": x509_presented,
+            },
+        )
+        raise_flask_error(422, "Agency users must authenticate using a PIV/CAC card")
```
chouinar: @glenblosser-nava - just double-checking that this is the message the frontend is looking for to know specifically it was a PIV issue.

api/src/services/users/login_gov_callback_handler.py
```diff
@@ -101,6 +101,37 @@ def handle_login_gov_callback_request(
...
+
+    # Check if user is an agency user
+    is_agency_user = len(user.agency_users) > 0
+
+    # If user is an agency user and didn't use PIV, reject login
+    if is_agency_user and not x509_presented:
+        logger.info(
+            "Agency user attempted login without PIV",
+            extra={
+                "user_id": str(user.user_id),
```
chouinar: Nitpicky: can leave a lot of common types alone for the extra in our logging, we have a mapper for types [here](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/logging/formatters.py#L29) - I know we do this in some places in the code, I try to clean them up when I can.
```suggestion
                "user_id": user.user_id,
```

api/tests/src/api/users/test_user_route_login.py
```diff
@@ -500,3 +500,184 @@ def test_user_callback_token_fails_validation_no_valid_key_302(
...
+    """Non-agency user logging in without PIV should succeed even when IS_PIV_REQUIRED=true"""
+    # Enable PIV requirement
+    monkeypatch.setattr("src.auth.login_gov_jwt_auth._config.is_piv_required", True)
+
+    # Create state (user will be created as new, non-agency user)
+    login_gov_state = LoginGovStateFactory.create()
+
+    code = str(uuid.uuid4())
+    id_token = create_jwt(
+        user_id="new-non-agency-user",
```
jakobpederson, line 609: Is there a user with this user_id or does it just not matter what id you put in here?

api/src/services/users/login_gov_callback_handler.py
```diff
@@ -101,6 +101,37 @@ def handle_login_gov_callback_request(
...
+        x509_presented: Whether the user authenticated with a certificate (PIV/CAC)
+
+    Raises:
+        HTTPError: If an agency user attempts to login without PIV when required
+    """
+    config = get_config()
+
+    # Skip validation if PIV is not required (lower environments)
+    if not config.is_piv_required:
+        return
```
chouinar: An idea - not required, but what if we did something like below. Just to have a log record in environments where we have this disabled.

```py
if is_agency_user and not x509_presented:
    if config.is_piv_required:
        # what you already have
    else:
        logger.info("User would have been blocked from login if piv required were enabled", extra={...})
```

api/src/services/users/login_gov_callback_handler.py
```diff
@@ -101,6 +101,37 @@ def handle_login_gov_callback_request(
...
+
+    # Check if user is an agency user
+    is_agency_user = len(user.agency_users) > 0
+
+    # If user is an agency user and didn't use PIV, reject login
+    if is_agency_user and not x509_presented:
+        logger.info(
```
chouinar, line 139: Just for metrics sake, if a user is an agency user and x509_presented is true, can we add a log message?

Then we can easily track agency user logins that were successful/rejected by this.

api/tests/src/api/users/test_user_route_login.py
desaiamit: Fixed - moved `AgencyFactory`, `AgencyUserFactory` imports to the top of the file and removed inline imports from test functions.

api/src/services/users/login_gov_callback_handler.py
desaiamit: Good callout - will wait for @glenblosser-nava to confirm this message.

api/src/services/users/login_gov_callback_handler.py
desaiamit: Good point - applied the suggestion. Thanks for the link to the type mapper!

api/tests/src/api/users/test_user_route_login.py
desaiamit: No existing user - this test simulates a brand new user's first login. The user_id is arbitrary and gets created during the login callback flow.

api/src/services/users/login_gov_callback_handler.py
desaiamit: Added - now logging successful agency user logins with PIV. This enables tracking both successful and rejected agency logins for metrics.

api/src/services/users/login_gov_callback_handler.py
desaiamit: Good idea - added logging for when agency users would have been blocked if PIV were required. This helps track potential issues in lower environments before enabling in production.

api/src/services/users/login_gov_callback_handler.py
glenblosser-nava: This is an acceptable message. The frontend work was waiting for this to be complete, so I can use whatever you send.

desaiamit: Thanks @glenblosser-nava

chouinar (approved):

PR #8521, opened 2026-02-12
Bug/email notification

babebe:
## Summary

Bug fix: https://onenr.io/0OQM9Xb4NQG
## Changes proposed

Added a null-safe helper function to safely format opportunity status values.

### DIFF
--- a/api/src/task/notifications/opportunity_notifcation.py
+++ b/api/src/task/notifications/opportunity_notifcation.py
@@ -471,12 +471,11 @@ class OpportunityNotificationTask(BaseNotificationTask):
         before = status_change["before"]
         after = status_change["after"]
 
-        def safe_status_display(value: str | None) -> str:
-            if value in OPPORTUNITY_STATUS_MAP:
-                return OPPORTUNITY_STATUS_MAP[value]
-            if isinstance(value, str):
-                return value.capitalize()
-            return NOT_SPECIFIED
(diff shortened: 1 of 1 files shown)


mdragon (approved):
👍🏻

api/src/task/notifications/opportunity_notifcation.py
```diff
@@ -471,8 +471,15 @@ def _build_opportunity_status_content(self, status_change: dict) -> str:
         before = status_change["before"]
         after = status_change["after"]
 
-        before = OPPORTUNITY_STATUS_MAP.get(before, before.capitalize())
-        after = OPPORTUNITY_STATUS_MAP.get(after, after.capitalize())
+        def safe_status_display(value: str | None) -> str:
+            if value in OPPORTUNITY_STATUS_MAP:
+                return OPPORTUNITY_STATUS_MAP[value]
```
jakobpederson: (optional) You could also do something like `OPPORTUNITY_STATUS_MAP.get(value, None)`

Also I feel like there's a case for checking for `None` right away, like:
```
if value is None:
    return NOT_SPECIFIED
```
Just because you shouldn't need to check for a `str` and I'm assuming `None` is not a key on that dict. But again, this is all optional.

babebe: Updated with correct type. Checking for None early and If not in the dict it will just capitalize

Issue #8010, opened 2026-01-15
Endpoint to get list of user's agencies


brianlee-nava:
### Summary

Resource: Knowledge Transfer: Opportunity Data Model Recording ([25:53](https://drive.google.com/file/d/1rmvWZwjRaQ8QyKhLPcvw_jSPw4lBvvmR/view?usp=sharing&t=1553))

**Background and Context**
We do have a foreign key on the Agency Table, but its not really used at this point. When creating an opportunity, Grantor or Agency users need to pick an agency to associate with the opportunity. We need to create an endpoint to enable users to select from a dropdown list of agencies and enforce whatever agency is being passed in (it exists) and Grantor has `create_opportunity` privilege. 

**Michael's Note:** We have an endpoint for listing orgs for a user, should be able to follow that convention and use that model for this ticket.

**User Story**
As a Grantor or Agency user
I want to select an Agency from a dropdown when creating an Opportunity
So that the Opportunity is correctly associated to a valid Agency and access is properly enforced

### Acceptance criteria

Feature: Select and validate Agency for Opportunity creation

  Background:
    Given the user is authenticated
    And the user is a Grantor or Agency user

  Scenario: Retrieve list of agencies for dropdown selection
    Given the user has access to one or more agencies
    When the user requests the list of agencies for Opportunity creation
    Then the system returns a list of agencies the user belongs to
    And each agency includes a valid agency identifier and display name
    And the response follows the same access rules as the existing list orgs endpoint

  Scenario: Successfully create an Opportunity with a valid agency
    Given the user has Create or Manage Opportunity privilege for an agency
    And the agency exists in the system
    When the user submits an Opportunity creation request with the agency identifier
    Then the system validates that the agency exists
    And the system validates the user has permission for the agency
    And the Opportunity is created with the agency foreign key populated
    And the system returns a success response

  Scenario: Fail to create an Opportunity with a non-existent agency
    Given the user submits an Opportunity creation request
    When the provided agency identifier does not exist
    Then the system rejects the request
    And the system returns an error indicating the agency is invalid

  Scenario: Fail to create an Opportunity without required privileges
    Given the agency exists
    And the user does not have Create or Manage Opportunity privilege for the agency
    When the user submits an Opportunity creation request with that agency identifier
    Then the system rejects the request
    And the system returns an authorization error

  Scenario: Fail to create an Opportunity without an agency selected
    When the user submits an Opportunity creation request without an agency identifier
    Then the system rejects the request
[...]

PR #8525, opened 2026-02-12
[Issue #8010] Add endpoint to get list of user's agencies

desaiamit:
## Summary

Work for #8010

## Changes proposed

- New service layer: `src/services/users/get_user_agencies.py` - uses user.agency_users relationship
- Added `UserAgencySchema` and `UserAgenciesResponseSchema` to user_schemas.py
- Added `user_get_agencies` route handler to user_routes.py
- Added 5 tests covering success, empty list, 403, and 401 cases

## Context for reviewers

This endpoint follows the same pattern as the existing `user_get_organizations` endpoint. It enables a dropdown for agency selection when creating opportunities.

The endpoint:
- `POST /v1/users/<user_id>/agencies`
- Returns agency_id, agency_name, and agency_code for each agency
- Supports both JWT and API user key authentication
- Returns 403 if user tries to access another user's agencies

## Validation steps

1. Start the API server
2. Generate a test token for a user with agency associations
3. Test the endpoint:
  curl -X POST \
    'http://localhost:8080/v1/users/<user_id>/agencies' \
    -H 'X-SGG-Token: <token>'

4. Verify response returns list of agencies with agency_id, agency_name, and agency_code
5. Test 403 by using a different user_id than the token's user
6. Test 401 by omitting the token header

### DIFF
--- a/api/tests/src/api/users/test_user_agencies_get.py
+++ b/api/tests/src/api/users/test_user_agencies_get.py
@@ -85,13 +85,13 @@ class TestUserAgenciesGet:
 
         assert resp.status_code == 401
 
-    def test_get_user_agencies_200_api_key_happy_path(self, enable_factory_create, client, db_session):
+    def test_get_user_agencies_200_api_key_happy_path(
+        self, enable_factory_create, client, db_session
+    ):
         """Test getting user agencies works with API key auth"""
         user, agency, _, api_key_id = create_user_in_agency_with_jwt_and_api_key(db_session)
 
-        resp = client.post(
-            f"/v1/users/{user.user_id}/agencies", headers={"X-API-Key": api_key_id}
-        )
+        resp = client.post(f"/v1/users/{user.user_id}/agencies", headers={"X-API-Key": api_key_id})
 
         assert resp.status_code == 200
         data = resp.get_json()
@@ -99,7 +99,9 @@ class TestUserAgenciesGet:
         assert len(data["data"]) == 1
         assert data["data"][0]["agency_id"] == str(agency.agency_id)
 
-    def test_get_user_agencies_403_api_key_wrong_user(self, enable_factory_create, client, db_session):
+    def test_get_user_agencies_403_api_key_wrong_user(
+        self, enable_factory_create, client, db_session
+    ):
         """Test that API key auth returns 403 when user_id doesn't match authenticated user"""
         _, _, _, api_key_id = create_user_in_agency_with_jwt_and_api_key(db_session)
         other_user, _, _ = create_user_in_agency_with_jwt(db_session)


api/src/services/users/get_user_agencies.py
```diff
@@ -0,0 +1,42 @@
...
+
+        agency_data: dict[str, Any] = {
+            "agency_id": str(agency.agency_id),
+            "agency_name": agency.agency_name,
+            "agency_code": agency.agency_code,
+        }
+
+        agencies.append(agency_data)
+
+    return agencies
```
chouinar: Rather than return a list of dicts, return a list of agencies. Marshmallow schemas can convert any type of the same shape into JSON, unless we are structuring data differently in the DB than in the API response, we can just return the DB objects directly.

```suggestion
    agencies = [a.agency for a in agency_users]

    return agencies
```

api/src/services/users/get_user_agencies.py
```diff
@@ -0,0 +1,42 @@
...
+
+import src.adapters.db as db
+from src.db.models.user_models import AgencyUser
+
+logger = logging.getLogger(__name__)
+
+
+def get_user_agencies(db_session: db.Session, user_id: UUID) -> list[dict[str, Any]]:
+    """Get user agencies in the format expected by the API response."""
+    agency_users = _fetch_user_agencies(db_session, user_id)
```
chouinar: Something we could do without needing to write a DB query. Pass the user into this rather than the ID (user_token_session is attached to the user). Then we just do `user.user_agencies` to get this.

api/src/api/users/user_routes.py
```diff
@@ -227,6 +229,34 @@ def user_get_organizations(db_session: db.Session, user_id: UUID) -> response.Ap
     return response.ApiResponse(message="Success", data=organizations)
 
 
+@user_blueprint.get("/<uuid:user_id>/agencies")
```
chouinar: We should make this a POST endpoint, our list/search endpoints do that as that way we can pass in a request body with filters/pagination - even if we don't need that for a first iteration. We don't use any request params, mainly for consistency.

api/src/api/users/user_routes.py
```diff
@@ -227,6 +229,34 @@ def user_get_organizations(db_session: db.Session, user_id: UUID) -> response.Ap
...
+@user_blueprint.output(UserAgenciesResponseSchema)
+@user_blueprint.doc(responses=[200, 401, 403])
+@user_blueprint.auth_required(api_jwt_auth)
```
chouinar: Do we want to allow API keys as well? That is, do we want folks to be able to query this via something other than just our frontend? Unless we have a good reason to restrict that, I'd say we should have both JWT and API user key auth.

api/src/api/users/user_routes.py
```diff
@@ -227,6 +229,34 @@ def user_get_organizations(db_session: db.Session, user_id: UUID) -> response.Ap
...
+    logger.info("GET /v1/users/:user_id/agencies")
+
+    user_token_session: UserTokenSession = api_jwt_auth.get_user_token_session()
+
+    # Verify the authenticated user matches the requested user_id
+    if user_token_session.user_id != user_id:
+        raise_flask_error(403, "Forbidden")
+
+    with db_session.begin():
+        agencies = get_user_agencies(db_session, user_id)
```
chouinar: If you want to simplify the queries even further in the service layer:
```suggestion
    user_token_session: UserTokenSession = api_jwt_auth.get_user_token_session()

    # Verify the authenticated user matches the requested user_id
    if user_token_session.user_id != user_id:
        raise_flask_error(403, "Forbidden")

    with db_session.begin():
        user = user_token_session.user
        db_session.add(user)
        agencies = get_user_agencies(db_session, user)
```

Pass the user themselves into the service layer and then we can use the model to fetch the objects. Note you have to do this `db_session.add` because the user gets fetched in a different DB session than the one you have here.

api/tests/src/api/users/test_user_agencies_get.py
```diff
@@ -0,0 +1,115 @@
...
+    UserFactory,
+)
+
+
+class TestUserAgenciesGet:
+    def test_get_user_agencies_200_with_agencies(self, enable_factory_create, client, db_session):
+        """Test getting user agencies returns list of agencies"""
+        # Create user
+        user = UserFactory.create()
+        LinkExternalUserFactory.create(user=user)
```
chouinar: I had suggested this to @dghazvini as well, as we add grantor endpoints and tests, we should make some sort of test utility for creating a user in a given agency with particular roles/privileges.

That'll save about a dozen lines of copy-pasted code in every grantor endpoint test and be quite helpful.

An equivalent of https://github.com/HHS/simpler-grants-gov/blob/main/api/tests/lib/organization_test_utils.py but for agencies.

api/src/services/users/get_user_agencies.py
desaiamit: Agreed, will update to return the DB objects instead of manually constructing dicts.

api/src/api/users/user_routes.py
desaiamit: Good suggestion - Will pass the User object from user_token_session and use the relationship.

api/src/api/users/user_routes.py
desaiamit: Understood, also makes sense for consistency with other list endpoints. Will change from GET to POST to allow for future filters/pagination.

api/src/api/users/user_routes.py
desaiamit: Good point. Will add support for both JWT and API user key auth. Also, user_get_organizations has the same pattern (JWT only) - should we raise a ticket to update that endpoint as well?

api/tests/src/api/users/test_user_agencies_get.py
desaiamit: As per your Slack message on simpler-backend - I assume that you are creating such a test utility so I won't create one

chouinar: @desaiamit - Daniel's recently merged change has that if you want to adjust these to use it. Should remove most of the setup from all of this.

api/tests/src/api/users/test_user_agencies_get.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def test_get_user_agencies_401_no_token(self, client):
+        """Test that accessing agencies without auth token returns 401"""
+        random_user_id = str(uuid.uuid4())
+
+        resp = client.post(f"/v1/users/{random_user_id}/agencies")
+
+        assert resp.status_code == 401
+
+    def test_get_user_agencies_401_invalid_token(self, client):
```
chouinar, line 78: Could you add two more simple tests using the API key auth (`X-API-Key`) since the endpoint accepts that as well.
* Happy path, shows that the auth works
* One that shows it'll fail if the ID doesn't correspond to anything

Don't need to duplicate everything since the logic after AuthN is the same either way, just verifies things are all properly connected.

desaiamit: 1. Refactored all tests to use `create_user_in_agency_with_jwt `and `create_user_in_agency_with_jwt_and_api_key` from `opportunity_test_utils`. Also merged main to pick up the utility.
2. Added two tests using `X-API-Key` auth:
-   `test_get_user_agencies_200_api_key_happy_path` — verifies the endpoint works with API key auth
-   `test_get_user_agencies_403_api_key_wrong_user` — verifies 403 is returned when the key belongs to a different user than the one in the URL

Issue #8530, opened 2026-02-12
Bug Fix: Update config file to correctly parse out the soap jwt auth url


jakobpederson:
### Summary

The current way we calculate `api_name` is breaking because the new jwt auth url `service_name` is different:

- Old way: `grantsws-agency`
- New way: `grantsws-agency-partner`

I'm adding `grantsws-agency-partner` to the list of acceptable `service_name` for the `SimplerSoapAPI.GRANTORS` type. [See this error.](https://onenr.io/0BQrWxKknwZ)

### Reproducibility

None

### Browser

None

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #8531, opened 2026-02-12
[Issue #8530] Fix SimplerSoapAPI.get_api_name to handle jwt service name

jakobpederson:
## Summary

Fixes / Work for #8530 

## Changes proposed

Added  `"grantsws-agency-partner"` to the check in `SimplerSoapAPI.get_api_name` method. Also added a few tests.

## Context for reviewers

The new SOAP jwt has a slightly different `service_name` in the url so things were breaking because the check was specifically for `"grantsws-agency"`.

### DIFF
--- a/api/src/legacy_soap_api/legacy_soap_api_config.py
+++ b/api/src/legacy_soap_api/legacy_soap_api_config.py
@@ -62,10 +62,7 @@ class SimplerSoapAPI(StrEnum):
 
     @staticmethod
     def get_soap_api(service_name: str, service_port_name: str) -> SimplerSoapAPI | None:
-        if (
-            service_name in ("grantsws-agency", "grantsws-agency-partner")
-            and service_port_name == "AgencyWebServicesSoapPort"
-        ):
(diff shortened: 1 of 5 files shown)


api/src/legacy_soap_api/legacy_soap_api_config.py
```diff
@@ -62,7 +62,10 @@ class SimplerSoapAPI(StrEnum):
 
     @staticmethod
     def get_soap_api(service_name: str, service_port_name: str) -> SimplerSoapAPI | None:
-        if service_name == "grantsws-agency" and service_port_name == "AgencyWebServicesSoapPort":
+        if (
+            service_name in ("grantsws-agency", "grantsws-agency-partner")
+            and service_port_name == "AgencyWebServicesSoapPort"
+        ):
```
chouinar: If I'm reading the code right, this is checking that the value passed into the API by the user, right?

I don't think we'd want to add this here, or at least not without other changes. I don't think the proxy should ever expect someone to call with `grantsws-agency-partner`, we should change `grantsws-agency` to that behind the scenes in the JWT auth path.

PR #8539, opened 2026-02-13
[Issue #8070] Setup localstack SQS creation as part of our local setup

### LINKED ISSUE #8070: Setup localstack to have SQS and create a queue as part of our local setup
### Summary

Similar to how we use localstack for S3, we’ll want to have an SQS queue that we are running locally. We’ll need to add SQS to the localstack config in our docker-compose, and then make a script that runs as part of `make init` to make the SQS queue.

https://docs.localstack.cloud/aws/services/sqs/

The script that sets up localstack for s3 can be extended to also setup SQS https://github.com/HHS/simpler-grants-gov/blob/main/api/bin/setup_localstack.py

In this script:
* Create an SQS queue from an environment variable pulled from an SQS config (let's call the env var `WORKFLOW_SQS_QUEUE` and name the queue something like `local_workflow_queue`)
* The queue should only be created if it does not already exist
* This does not need to use our SQS client, fine to use boto3 directly for checking if the queue exists / creating it.
* The queue should be created against localstack, make sure to set the AWS creds to some dummy set of values (like the s3 code).

### Acceptance criteria

- [ ] Localstack sqs setup to run in our docker-compose files
- [ ] SQS queue should be created automatically as part of setting up localstack via a python script

hao10282025-sudo:
## Summary

Fixes #8070 

## Changes proposed

Setup localstack SQS creation as part of our local setup

## Context for reviewers

Similar to how we use localstack for S3, we’ll want to have an SQS queue that we are running locally. We’ll need to add SQS to the localstack config in our docker-compose, and then make a script that runs as part of make init to make the SQS queue.

## Validation steps

- Localstack sqs setup to run in our docker-compose files
- SQS queue should be created automatically as part of setting up localstack via a python script

### DIFF
--- a/api/src/adapters/aws/sqs_adapter.py
+++ b/api/src/adapters/aws/sqs_adapter.py
@@ -11,7 +11,7 @@ logger = logging.getLogger(__name__)
 
 
 class SQSConfig(PydanticBaseEnvConfig):
-    workflow_queue_url: str = Field(alias="WORKFLOW_SQS_QUEUE")
+    workflow_queue_url: str = Field(alias="WORKFLOW_QUEUE_URL")
     s3_endpoint_url: str = Field(alias="S3_ENDPOINT_URL")
 
 
--- a/api/local.env
+++ b/api/local.env
@@ -103,7 +103,7 @@ SEARCH_VERIFY_CERTS=FALSE
 # If you want to connect to localstack outside of docker
 # use localhost:4566 instead
 S3_ENDPOINT_URL=http://localstack:4566
-WORKFLOW_SQS_QUEUE=http://sqs.us-east-1.localhost.localstack.cloud:4566/000000000000/local_workflow_queue
+WORKFLOW_QUEUE_URL=http://sqs.us-east-1.localhost.localstack.cloud:4566/000000000000/local_workflow_queue
 
 ############################
 # S3
--- a/api/tests/src/adapters/aws/test_sqs_adapter.py
+++ b/api/tests/src/adapters/aws/test_sqs_adapter.py
@@ -40,17 +40,13 @@ class TestGetBotoSQSClient:
         mock_get_session.return_value = mock_session
         get_boto_sqs_client()
         mock_get_session.assert_called_once()
-        mock_session.client.assert_called_once_with(
-            "sqs", endpoint_url=SQSConfig().workflow_queue_url
-        )
+        mock_session.client.assert_called_once_with("sqs", endpoint_url=SQSConfig().s3_endpoint_url)
 
     def test_uses_provided_session(self):
         """Verify that the factory function uses a specifically provided AWS session."""
         mock_session = Mock()
         get_boto_sqs_client(session=mock_session)
-        mock_session.client.assert_called_once_with(
-            "sqs", endpoint_url=SQSConfig().workflow_queue_url
-        )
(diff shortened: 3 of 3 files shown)


api/src/adapters/aws/sqs_adapter.py
```diff
@@ -13,6 +13,8 @@
 class SQSConfig(PydanticBaseEnvConfig):
     # SQS Queue URL will be required by the service using this adapter
     workflow_queue_url: str = Field(alias="WORKFLOW_QUEUE_URL")
+    # SQS Queue name for workflow processing
+    workflow_queue_name: str = Field(alias="WORKFLOW_QUEUE_NAME", default="local_workflow_queue")
```
chouinar: We don't want to add a new parameter, I think this wouldn't work with how we connect to SQS which with how we have it configured expects the URL to be something like `https://sqs.us-east-1.amazonaws.com/1234/MyQueue`.

Since the path has the queue name in it, where we need it elsewhere for creating the SQS queue, just do something like `queue_name = workflow_queue_url.split("/")[-1]` - it's fine if the local setup is a bit hacky, we'll define it once and not change it later.

api/src/adapters/aws/sqs_adapter.py
```diff
@@ -31,10 +33,20 @@ class SQSDeleteBatchResponse(BaseModel):
...
-def get_boto_sqs_client(session: boto3.Session | None = None) -> botocore.client.BaseClient:
+def get_boto_sqs_client(
+    sqs_config: SQSConfig | None = None, session: boto3.Session | None = None
+) -> botocore.client.BaseClient:
+    if sqs_config is None:
+        sqs_config = SQSConfig()
+
+    params = {}
+    if sqs_config.workflow_queue_url is not None:
+        params["endpoint_url"] = sqs_config.workflow_queue_url
```
chouinar: The endpoint url is not the queue URL. The queue URL would be a full path to the SQS queue. The endpoint URL is just something we set to make it so we can point to localstack. We'd want what we set for the endpoint url to be different - really we should reuse what `S3_ENDPOINT_URL` has although the name probably would need to be changed which might require a bit of rework in places.

api/src/adapters/aws/sqs_adapter.py
hao10282025-sudo: updated

api/src/adapters/aws/sqs_adapter.py
hao10282025-sudo: updated

hao10282025-sudo:
> We don't need unit tests (or more accurately - unit tests wouldn't work), but have you tested that our SQS client can interact with localstack?
> 
> Easiest way to test it is make a simple script. Something like:
> 
> ```python
> @task_blueprint.cli.command("my-script")
> def my_script():
>     pass
> 
>     # put whatever code in here you want to run
> ```
> 
> Then you can run it by doing `make cmd args="workflow my-script"`. The reason we want to run this way is that attaching it to a blueprint attaches it to all of our setup code / avoids any issues with using other parts of the code. You do need to make sure the file you put it in is imported (see what we do in a bunch of places like https://github.com/HHS/simpler-grants-gov/blob/main/api/src/data_migration/__init__.py ) - otherwise the app never imports your file and doesn't know about the command.

test with local script as below:
image

Issue #8198, opened 2026-01-27
Update CSV Export to Match Frontend Search Results


joshtonava:
### Summary

The search results CSV download currently exports 49 columns of opportunity data, creating an overwhelming experience for users who expect columns matching what they see in the search results table. Update the CSV export to output only the 7 columns displayed in the frontend (close date, status, title, opportunity number, agency, award min/max), and an environment specific link to the opportunity.

### Implementation Details                                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                               
- **Main file:** `api/src/services/opportunities_v1/opportunity_to_csv.py`
- Replace 49-field CSV_FIELDS list with 8 fields
- **Column order:** close_date, opportunity_status, opportunity_title, opportunity_number, agency_name, award_floor, award_ceiling, url

### Acceptance criteria

- [ ] CSV export contains exactly 8 columns in the correct order matching frontend display
- [ ] URL column contains full opportunity links using FRONTEND_BASE_URL environment variable
- [ ] Opportunity number is a separate column (not combined with title)
- [ ] Dates remain in ISO format (YYYY-MM-DD) and currency as raw numbers
- [ ] Missing/null values (close_date, award_floor, award_ceiling) output as empty strings without errors
- [ ] Tests cover column headers, status mapping, URL construction, and null handling
- [ ] **Local testing:** CSV downloads successfully with correct format via POST /v1/opportunities/search with format="csv"
- [ ] **Manual verification:** Downloaded CSV opens correctly in Excel/Numbers/Google Sheets with expected columns

babebe:
Changes to AC:
Columns are largely kept the same with the following changes:
Add new "url" field to csv_fields for both extracts and search endpoint
Remove "summary_description" from search endpoint csv

PR #8547, opened 2026-02-13
[Issue #8198] Update CSV export match UI

babebe:
## Summary

Fixes / Work for #8198  

## Changes proposed
Added new column "URL" computed using FRONTEND_BASE_URL environment variable
Added `has_full_fields` flag to `opportunities_to_csv`. Exports will have all the CSV fields while the `opportunity_search` endpoint will not have `summary_description` included.

Ensured:

- Dates remain ISO format (YYYY-MM-DD)
- Currency values remain raw numeric values (no formatting, commas, or symbols)
- Missing values (close_date, award_floor, award_ceiling) render as empty strings

Added tests for:
- URL construction
- ISO date validation
- Raw numeric currency validation
- Null handling behavior

## Validation steps

Locally, search for opportunities and export to csv
Confirm can open file successfully.
Validate columns
Confirm data format

[grants-search-202602171352.csv](https://github.com/user-attachments/files/25372351/grants-search-202602171352.csv)

### DIFF
--- a/api/src/auth/login_gov_jwt_auth.py
+++ b/api/src/auth/login_gov_jwt_auth.py
@@ -35,7 +35,8 @@ class LoginGovConfig(PydanticBaseEnvConfig):
 
     client_id: str = Field(alias="LOGIN_GOV_CLIENT_ID")
     acr_value: str = Field(alias="LOGIN_GOV_ACR_VALUE", default="urn:acr.login.gov:auth-only")
-    scope: str = Field(alias="LOGIN_GOV_SCOPE", default="openid email")
+    scope: str = Field(alias="LOGIN_GOV_SCOPE", default="openid email x509:presented")
+    is_piv_required: bool = Field(alias="IS_PIV_REQUIRED", default=False)
 
     # While all of these endpoints are under the same root, we define the full
     # path each time because the local mock uses a different naming convention
@@ -91,6 +92,7 @@ def get_config() -> LoginGovConfig:
 class LoginGovUser:
     user_id: str
     email: str
+    x509_presented: bool | None = None
 
 
 def _refresh_keys(config: LoginGovConfig) -> None:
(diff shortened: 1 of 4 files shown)


api/src/services/opportunities_v1/opportunity_to_csv.py
```diff
@@ -1,66 +1,35 @@
...
 
 
-def opportunities_to_csv(opportunities: Sequence[dict], output: io.StringIO) -> None:
-    writer = csv.DictWriter(output, fieldnames=CSV_FIELDS, quoting=csv.QUOTE_ALL)
+def _build_opportunity_url(opportunity_id: str, base_url: str | None = None) -> str:
+    """
+    Build the full frontend URL for an opportunity.
+    """
+    if base_url is None:
+        base_url = os.getenv("FRONTEND_BASE_URL")
```
chouinar: This gets called inside the loop, doing `os.getenv` for every opportunity (which if you export the default search is ~1500) seems like a lot, I'd make it so the base url has to be passed in so we fetch it once outside of the loop.

babebe: Yes, that makes sense updated.

api/src/api/opportunities_v1/opportunity_routes.py
```diff
@@ -214,7 +214,19 @@ def opportunity_search(
...
+            "opportunity_title",
+            "opportunity_number",
+            "agency_name",
+            "award_floor",
+            "award_ceiling",
+            "url",
+            "opportunity_id",
+            "post_date",
+        ]
+        opportunities_to_csv(opportunities, output, csv_fields)
```
chouinar: Didn't we say we just wanted to remove the summary description + add the URL?

I think what we want for the `opportunities_to_csv` function is a switch for full/non-full (whatever we call that). The API uses the non-full (with summary). The one we generate for extracts is the full one. Both should have the URL added.

We then define the fields in that code, and just pass in `has_full_fields=True/False`

babebe: Yes haven't applied the change yet. And you mean the API won't have the summary correct ?

Issue #8416, opened 2026-02-06
A11y bug: Filter headings missing focus state


andycochran:
### Summary

**[Loom](https://www.loom.com/share/2c4386d17e1041e88d27351370129167)**

When tab navigating to filter accordions, some of them do not get the appropriate `:focus` styles. It seems this is a result of differing classes, e.g.… 

This `.usa-accordion__heading` gets a focus style: 
```
<h2 class="usa-accordion__heading overflow-visible"><button type="button" class="usa-accordion__button" aria-expanded="true" aria-controls="opportunity-filter-fundingInstrument" data-testid="accordionButton_opportunity-filter-fundingInstrument">Funding instrument</button></h2>
```

This one does not: 

```
<h2 class="usa-accordion__heading maxh-mobile-lg overflow-auto position-relative"><button type="button" class="usa-accordion__button" aria-expanded="false" aria-controls="opportunity-filter-eligibility" data-testid="accordionButton_opportunity-filter-eligibility">Eligibility</button></h2>
```

(When I recorded the Loom, I thought this might be a Safari issue (as it handles `:focus-visible` differently), but seems the bug is there in Chrome too.) 

### Reproducibility

I tried multiple times and saw the bug every time

### How to reproduce the bug

- Open the filter pane
- Use Tab to navigate with the keyboard to the various filters
- Note how filters with specific classes no not get the appropriate `:focus` states

### Browser

Safari

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

ErinPattisonNava:
Just confirming that you turned on highlighting specifically in Safari.  Alas, it isn't on by default: https://www.a11y-collective.com/blog/how-to-activate-keyboard-navigation-on-macos/

see section "Highlight each focusable item in Safari"

andycochran:
Yeah. I was sus that this was a Safari issue. They've made… [choices](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button#clicking_and_focus). But checked it's present in Chrome too

ErinPattisonNava:
Thanks for confirming @andycochran !

@chelsea-nava and @eric-serrato95 note that this does put us out of ADA compliance on our largest use case, so I'd recommend prioritizing work on this.

andycochran:
Should be a pretty easy fix of standardizing the heading classes. IDK why some of them are different. And I don't think they need to be.

chelsea-nava:
[...]

PR #8550, opened 2026-02-13
[Issue #8416] Filter focus state bug

andycochran:
## Summary

Work for #8416

## Changes proposed

- Fix CSS bug where some search filter accordion headings are not getting a focus style

## Context for reviewers

This bug is a result of the Truss component applying `className` to both its `headingClasses` and `contentClasses` (see [code](https://github.com/trussworks/react-uswds/blob/15d032a3417d75b55f6cf08cb13d1a7435c4e7f1/src/components/Accordion/Accordion.tsx#L32-L37)). So the utility classes we had applied to make the content scrollable, also try to make the heading scrollable. That broke the focus styles. 

I removed all these utility classes and added a new semantic custom style class: `.force-scrolling`. It does the same thing as the utility classes were doing. But I scope the styles to `.usa-accordion__content` (and avoid applying them to `.usa-accordion__heading`). 

Not part of the reported bug, but I also improved the layout of search filter accordions so focus styles are not cropped by the container's hidden overflow-x. I moved the top & bottom border from the drawers scrollable area to the divs above/below it. I added negative margin and positive padding to the scrollable div, so it extends a bit to the left but its contents still align. The prevents the focus styles being cropped. 

I also refined some of the whitespace, simplifying how we apply margin and padding to the drawer. Each filter is no longer concerned w/ setting its own padding

https://www.loom.com/share/8fa08cd746d048b0a7b165f61ad4f573

## Validation steps

- Load the search page
- Open the filter drawer
- See the improved styles
- Navigate with the keyboard, tabbing to each filter
- Make sure all the filter headings receive a focus style
- Make sure all the filters that have a max height are scrollable (eligibility, agency, and category)

### DIFF
--- a/api/src/legacy_soap_api/legacy_soap_api_proxy.py
+++ b/api/src/legacy_soap_api/legacy_soap_api_proxy.py
@@ -48,7 +48,10 @@ def get_proxy_response(soap_request: SOAPRequest, timeout: int = PROXY_TIMEOUT) 
                 config, soap_auth.certificate.legacy_certificate
             )
         }
-        proxy_url = join(config.soap_partner_gateway_uri, soap_request.full_path.lstrip("/"))
+        proxy_url = join(
+            config.soap_partner_gateway_uri,
+            "grantsws-agency-partner/services/v2/AgencyWebServicesSoapPort",
+        )
     else:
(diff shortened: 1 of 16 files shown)


andycochran:
cc @eric-serrato95

Issue #7025, opened 2025-11-10
Code Refactors Ahead of Building Expandable Field Groups


doug-s-nava:
### Summary

It will be easier to implement expandable field groups having cleaned up code around form generation

### Acceptance criteria

- [ ] build on https://github.com/HHS/simpler-grants-gov/pull/6909 if not merged
- [ ] split up getFieldConfig function
- [ ] reorganize utils file into individual files (saveUtils, pathFormatUtils, etc.)
- [ ] Unittest should be split, but they should all continue to pass
- [ ] Do a solid regression test on SF424 to ensure we haven't broken anything.

chris-kuryak:
@doug-s-nava I assume you are tagging me here for a regression test on SF-424?

If so, confirming everything you need me to review is in staging?

doug-s-nava:
Yes, this change is up in staging. As discussed last week this regression test should cover the minimum set of forms representing unique functionality. If I recall we settled on:

* sf424
* sf424a
* NEH cover sheet
* sfLLL

chris-kuryak:
UAT performed on this Application:
https://staging.simpler.grants.gov/applications/fe2a7310-bb5c-49ed-bfd6-9eb32804edb4

- [x] Application can be submitted successfully when no errors exist

- [x] sf424
  - [x] Required fields are accurate
  - [x] Conditionally required fields are accurate
  - [x] Errors within forms are accurate
  - [x] Form fields appear as read-only after submission
  - [x] PDF generates as expected with accurate information
  - [x] Pre-population fields populate as expected
  - [x] Pre-population fields cannot be changed upon saving form
  - [x] Post-population fields appearing as read only
  - [x] Post-population fields populating as expected
  - [x] Tabbing navigation works as expected
  - [x] Adding attachments on form works successfully
  - [x] Removing attachments on form works successfully
  - [x] Application cannot be submitted successfully when forms does not pass validation
  - [x] Application errors appear after unsuccessful submission attempt
  - [x] Instructions are present (if available) and open as expected
  - [x] Field selection options (if available) are accurate
- [x] sf424a
  - [x] Required fields are accurate
  - [x] Conditionally required fields are accurate
  - [x] Errors within forms are accurate
  - [x] Form fields appear as read-only after submission
  - [x] PDF generates as expected with accurate information
  - [x] Pre-population fields populate as expected
  - [x] Pre-population fields cannot be changed upon saving form
  - [x] Post-population fields appearing as read only
[...]

PR #6909, opened 2025-10-31
[Issue #7025] remove unused support for deeply nested form fields

doug-s-nava:
## Summary

Work for #7025 (partial, see also #8552) 

## Changes proposed

Some refactors to the form generation process for clarity, and removing unused code branches. Does not affect functionality of any existing forms.

Note that this work is based off of https://github.com/HHS/simpler-grants-gov/pull/8552 blocked until that change is merged

## Context for reviewers

When reviewing form generation code for the the expandable field group tech spec (https://navasage.atlassian.net/wiki/x/AgCqgg), I was confused by a few points that I thought could be easily cleaned up.

* logic branches existed in `<FormFields>` that allow for arbitrary nesting within the UI schema, even though the system doesn't support any idea of nesting here beyond a single depth for "sections". We may want to support subsections or some other more deeply nested implementation in the future, but it is not necessary now and adds complexity, so taking that out
* adding some comments
* renaming some variables
* more deliberate error handling in getFieldConfig

## Validation steps

1. load up and save each form with some data
2. _VERIFY_: everything works as expected

PR #8552, opened 2026-02-17
[Issue #7025] refactor getFieldConfig functionality in apply flow

doug-s-nava:
## Summary

Work for #7025 (partial, see also https://github.com/HHS/simpler-grants-gov/pull/6909)

## Changes proposed

Various code refactors to the `getFieldConfig` section of the apply form data transformation flow, in order to include code readability, testability, extensibility

## Validation steps

1. do a full regression test on existing forms
2. _VERIFY_: all forms continue to work as expected


frontend/src/utils/applyForm/getFieldConfig.ts
```diff
@@ -0,0 +1,401 @@
...
+    value,
+    fieldSchema,
+    rawErrors,
+    fieldName,
+    // not used on multifield fields but returning here to keep things symmetrical
+    htmlFieldName: "",
+  };
+};
+
+const getEnumOptions = ({
```
doug-s-nava: this is new, but largely contains code split out from getFieldConfig - splitting it out to decrease the complexity of the main function

frontend/src/components/applyForm/utils.ts
```diff
@@ -239,69 +233,6 @@ export const buildWarningTree = (
   return [];
 };
 
-// json schema doesn't describe UI so types are infered if widget not supplied
-export const determineFieldType = ({
```
doug-s-nava, line 243: moving to getFieldConfig file, as it is only used to support that function

frontend/src/components/applyForm/utils.ts
```diff
@@ -323,24 +254,7 @@ export const getFieldSchema = ({
   return schema as RJSFSchema;
 };
 
-export const getNameFromDef = ({
```
doug-s-nava, line 326: moving to getFieldConfig file, as it is only used to support that function

frontend/src/components/applyForm/utils.ts
```diff
@@ -421,183 +335,6 @@ const jsonPointerToPath = (jsonPointer: string) => {
...
-      maxLength: fieldSchema?.maxLength,
-      schema: fieldSchema,
-      rawErrors,
-      value,
-      options,
-    },
-  };
-};
-
-export const getWarningsForField = ({
```
doug-s-nava, line 579: moving to getFieldConfig file, as it is only used to support that function

frontend/src/components/applyForm/types.ts
```diff
@@ -74,9 +74,9 @@ export type WidgetTypes =
   | "Budget424aSectionE"
   | "Budget424aSectionF";
 
-export type DefinitionPath =
-  | `/properties/${string}`
-  | [`/properties/${string}`];
+type PropertyPath = `/properties/${string}`;
```
doug-s-nava, line 77: this was incorrectly typed before, though it didn't break anything until I tried to do some more granular testing

frontend/src/components/applyForm/FormFields.tsx
```diff
@@ -55,20 +52,25 @@ export const FormFields = ({
...
         uiSchema.forEach((node) => {
           if ("children" in node) {
+            // if handling a section with an array of children, build the tree for the children within the section
             buildFormTree(node.children as unknown as UiSchema, {
               label: node.label,
               name: node.name,
               description: node.description,
             });
           } else if (!parent && ("definition" in node || "schema" in node)) {
+            // if handling a single field outside of a section ???
```
ErinPattisonNava, line 72: Can we remove the code here then if it has never ben tested, and just throw an error as an unreachable branch?

frontend/src/components/applyForm/types.ts
```diff
@@ -47,7 +47,7 @@ export type FormValidationWarning = {
   field: string;
   message: string;
   type: string;
-  value: string;
+  value: string | null;
```
ErinPattisonNava, line 50: Do you have a use case to test this with?

frontend/src/utils/applyForm/getFieldConfig.ts
```diff
@@ -0,0 +1,413 @@
+import { EnumOptionsType, RJSFSchema } from "@rjsf/utils";
```
ErinPattisonNava, line 1: Is this file a straight copy from the files where code is deleted above, or has it been further refactored and needs a complete review?

ErinPattisonNava (commented):
For verification, when you say that all forms needs to be verified, is that verified that they can be saved?  Submitted?  That errors show up?  Do we need to verify this on every form, or could we just do the sf424?

frontend/src/components/applyForm/FormFields.tsx
doug-s-nava: good, question - that stuff will get handled in the next pr, and these comments will all go away https://github.com/HHS/simpler-grants-gov/pull/6909

frontend/src/components/applyForm/types.ts
doug-s-nava: yeah, if you save an empty sf424, all of the required field validation errors should come back with a null value

frontend/src/utils/applyForm/getFieldConfig.ts
doug-s-nava: These functions have been copied with no change:

* determineFieldType
* getWarningsForField
* getNamesFromDef

The previous functionality of getFieldConfig has been split across 4 new functions
* getFieldInfo - this doesn't do much other than handle the branching for normal vs multifield fields
* getBasicFieldInfo - determines widget type and props for a non-multifield field. Encompasses lines 477 - 499 from the previous implementation
* getBasicMultiFieldInfo - determines widget type and props for a multifield field. Encompasses lines 445 - 475 from the previous implementation
* getEnumOptions - determines values for an enum if a field needs it. Encompasses lines 518 - 561 from the previous implementation

doug-s-nava:
@ErinPattisonNava I know we have a lot more forms now than we used to, and I don't think they all need to be checked. We should check a subset that accounts for all unique functionality though. That may be the sf424 and sf424a, but does the sf424 have a multi-attachment field? If not maybe we need to include a form that has one of those. The sfLLL may be good for complex nesting, and probably the NEH cover sheet to make sure nested allOfs are still good

Issue #8432, opened 2026-02-06
[BE] Create new database table and models for organization saved opportunities


freddieyebra:
### Summary

Create a new db table and models for `organization_saved_opportunities` with the following fields:

organization_id - UUID - pk
opportunity_id - UUID - pk
created_at - timestamp
updated_at - timestamp

```
class OrganizationSavedOpportunity(ApiSchemaTable, TimestampMixin):
...
```

### Acceptance criteria

- Necessary migrations are executed and the `organization_saved_opportunities` table exists in all environments.
- Unit tests pass

PR #8563, opened 2026-02-17
[Issue #8432] Create organization saved opportunities table

babebe:
## Summary

Fixes / Work for #8432 

## Changes proposed

Added new model OrganizationSavedOpportunity
Added corresponding factory
Created migration file
Updated seed data:
Modified _add_saved_opportunities to handle both users and organizations
Added saved opportunities for org3

## Validation steps

Run migrations
Run make db-seed-local
Confirm that the saved opportunities table contains records for both users and org3

### DIFF
--- a/frontend/tests/e2e/search/features/filterDrawer/filter-drawer-sort.feature
+++ b/frontend/tests/e2e/search/features/filterDrawer/filter-drawer-sort.feature
@@ -1,0 +1,63 @@
+Feature: Filter Drawer - Sort
+  As a user
+  I want the sort UI to appear in the correct location based on my device
+  So that sorting behaves consistently while matching the intended layout
+
+Background:
+    Given I am on the Search Funding Opportunity page
+    And my login state is "<loginState>"
+    And the viewport is set to "<viewport>"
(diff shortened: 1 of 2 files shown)


api/src/db/models/opportunity_models.py
```diff
@@ -125,6 +126,12 @@ def agency(self) -> str | None:
         cascade="all, delete-orphan",
         foreign_keys="[ReferencedOpportunity.derived_opportunity_id]",
     )
+    saved_opportunities_by_organizations: Mapped[list[OrganizationSavedOpportunity]] = relationship(
+        "OrganizationSavedOpportunity",
+        back_populates="opportunity",
+        uselist=True,
+        cascade="all, delete-orphan",
+    )
```
chouinar, line 134: Not a fault of this implementation, but in our get_opportunity service function we exclude certain relationships from the GET opportunity call to avoid fetching a ton of info in the DB. Could you add this relationship to that list?

babebe: Makes sense. Added there

PR #8577, opened 2026-02-18
[unticketed] hardcode storybook to english to work around vite asset path bug

doug-s-nava:
## Summary

unticketed

## Changes proposed

fixes bug that was causing all storybook stories to throw an error (see screenshot) based on inability to find messages files

## Context for reviewers

The function we're using to grab messages for next-intl relies on a hardcoded path that vite was not able to properly resolve in storybook builds. Hardcoding to English works around this

## Validation steps

1. `npm ci && npm run storybook-build && npx http-server ./storybook-static` on this branch
2. _VERIFY_: all storybook stories load as expected

### Error state

Screenshot 2026-02-18 at 12 13 17 PM

### DIFF
--- a/frontend/.storybook/preview.tsx
+++ b/frontend/.storybook/preview.tsx
@@ -7,7 +7,7 @@ import { Loader, Preview } from "@storybook/react";
 import "src/styles/styles.scss";
 
 import { defaultLocale, locales } from "src/i18n/config";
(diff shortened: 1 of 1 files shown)


doug-s-nava:
@nathan-stricker - theoretically, but we have no plans to support Spanish any time soon

PR #8582, opened 2026-02-18
[Issue #8389] Add approvals to workflow logic

### LINKED ISSUE #8389: Add approval logic to workflow
### Summary

This will consist of:
* Adding the approval configuration setup to the workflow config
* Adjusting the initial prototype workflow to use this approval config in its steps (we want at least 2 different types of approvals).
* Adding two new privileges for these approvals

An approval event will populate the metadata of an event like so:

```py
{
  # .. other data already present
  "metadata": {
    "approval_response_type": "approved", # This can be one of the values from ApprovalResponseType
    "comment": "str", # optional
  }
}
```

For each approval event that we process, we'll need to handle 3 possible approval response types of approved, declined, and requires_modification. Each of these types will need to affect the logic like so:

Image

### Acceptance criteria

- [ ] Approval configuration setup
- [ ] Initial prototype workflow has at least two approval steps configured
- [ ] Documentation added for how approvals work (high and low-level)

chouinar:
## Summary
Fixes #8389 

## Changes proposed
* Add approval support to workflow logic, including supporting multiple different types of approvals, with varying amounts required in that workflow.
* Add a way to generate the state machine diagrams via a script (helps visualize what is going on here).

## Context for reviewers
There's a lot of little changes to pipe data around that is needed that somewhat bloat the size of this PR, but if you're looking for what to focus on:
* The state machine definition and how approval states are connected
* The approval processor itself, as well as how it connects to everything

First, a crash course in approvals. Approvals have 3 "approval response types" of `approved`, `declined`, `requires_modification` which will affect how the state of the state machine needs to change. Broadly, approved means it can potentially move onto the next state. Declined will move it to some sort of end state (also called declined in this case) and requires modification moves it earlier in the flow. These aren't hard requirements, we could adjust how this all works (could make declined and requires modification go to the same place), but that's how I've setup the prototype.

On the state machine side to do this we need to check the approval response type and map it to these transitions. This is done via this chunk of code:

```py
    ## Program officer approvals
    receive_program_officer_approval = Event(
        # If Approved -> Add approval event and then check if enough approvals have occurred to determine next state
        states.PENDING_PROGRAM_OFFICER_APPROVAL.to.itself(
            cond=WorkflowConstants.IS_APPROVAL_EVENT_APPROVED,
            on=WorkflowConstants.ON_AGENCY_APPROVAL_APPROVED,
            after="check_program_officer_approval",
        )
        |
        # If Declined -> Add approval event and move to Declined state
        states.PENDING_PROGRAM_OFFICER_APPROVAL.to(
            states.DECLINED,
            cond=WorkflowConstants.IS_APPROVAL_EVENT_DECLINED,
            on=WorkflowConstants.ON_AGENCY_APPROVAL_DECLINED,
        )
        |
        # If Requires Modification -> Add approval event and move back to Start
        states.PENDING_PROGRAM_OFFICER_APPROVAL.to(
            states.START,
            cond=WorkflowConstants.IS_APPROVAL_EVENT_REQUIRES_MODIFICATION,
            on=WorkflowConstants.ON_AGENCY_APPROVAL_REQUIRES_MODIFICATION,
        )
    )

    check_program_officer_approval = Event(
        # If it has enough approvals, go to the Budget officer approval
        states.PENDING_PROGRAM_OFFICER_APPROVAL.to(
            states.PENDING_BUDGET_OFFICER_APPROVAL, cond=WorkflowConstants.HAS_ENOUGH_APPROVALS
        )
        # If not, stay in this state
        | states.PENDING_PROGRAM_OFFICER_APPROVAL.to.itself(),
    )
```
The defines a few things:
* What are the transitions. There are actually 4 transitions here off of 2 events. Declined/Requires Modification move if that conditional is satisfied. If it is Approved, it says to run a different event after which checks that there are enough approvals. This is necessary because we have to first add the approval event to the DB before we can check which state it goes to.
* Any "on" handlers, these are where the logic is actually defined, and these alongside the conditionals are all defined in the BaseStateMachine class (ie. we can reuse this pattern for all approvals).

Note that it is very boilerplatey, I'm think that in the future we'll make a utility for generating this that is a bit less verbose, but want to see how things evolve before I do that. An important detail to be aware of, `cond` / `on` / `after` always refer to a function defined in the state machine class itself, in this case they're all defined in the base state machine for reusability.

---

As for approvals, most of that logic is just validation that you can do the approval and it makes sense to do it. After that it's mainly just adding an approval record to the DB. I can imagine a lot more we could expand in that logic, but let's keep it (relatively) simple for now.

---

To give a step-by-step guide to how the state machine handles an event. Let's assume we receive the event `receive_program_officer_approval` with an approval response type of `approved`.
* The state machine will take in that event and see it has several transitions defined in the `receive_program_officer_approval` event. It will look at the conditionals in order to figure out which transition to run. In this case, the conditional will see the approval response type is `approved`, so we do the first one.
* It then runs the `on` function which is the approval processor code and writes the approval event to the approval table.
* It then sees that there is an `after` defined and will send another event of `check_program_officer_approval`.
* That other event also has a conditional which queries the DB and checks the number of approvals of that type against what is required
* Assuming enough approvals are configured, it finally transitions to the `end` state

## Validation steps
All of it is in the unit tests still at this phase as we haven't connected to the service yet.


api/src/workflow/workflow_constants.py
```diff
@@ -1,2 +1,17 @@
 class WorkflowConstants:
     START_WORKFLOW = "start_workflow"
+
+    APPROVAL_RESPONSE_TYPE = "approval_response_type"
+    COMMENT = "comment"
+
+    # Common conditionals
+    IS_APPROVAL_EVENT_APPROVED = "is_approval_event_approved"
```
hao10282025-sudo, line 8: a quick reminder if the naming of this WorkflowConstants is still going to be updated

chouinar: I hadn't settled on how I want to adjust the naming, still not a fan of this way, but at least with constants, adjusting that in a few places will be easier.

I think when I go through and write the documentation, I'll adjust it then

Issue #8104, opened 2026-01-23
Add validation logic to workflow event API


chouinar:
### Summary

https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2374369297/Tech+Spec+Workflow+Management+with+Python+StateMachine#Logic

This validation should include the following:
* (Start workflow event only) - whether the entity IDs passed in exist (404 if not)
* (Start workflow event only) - Whether the workflow type is real (422 if not) (This likely is already handled as part of the API schema validation and does not need to be explicitly implemented)
  * Also account for whether the workflow type has anything configured for it. 
* (Start workflow event only) - Whether the workflow type accepts that type of entity ID (422 if not)
* (Process workflow) - The workflow exists AND the workflow’s is_active flag is true (404 if missing, 422 if flag incorrect)
* (Process workflow) - The event is one the workflow could take in (422 if not - don't worry about solving whether it can currently receive that event)

Several of these checks were built into the core workflow logic, we should reuse any utilities created there. It may be that some of these error cases already get thrown, we'll want to catch those errors and instead do a `raise_flask_error(404, "some message")` with the appropriate error code as above.

This largely consists of fetching the workflow config + state machine class from the WorkflowRegistry.

### Acceptance criteria

- [ ] Above validation rules are added to the event API
- [ ] Unit tests for each scenario are created

ychoy:
@chouinar  will add more details from tech spec

PR #8584, opened 2026-02-18
[Issue #8104] Add validation logic to workflow event API and tests

kkrug:
## Summary

Fixes for #8104  

## Changes proposed

- Added validation logic to the service layer for the workflow events API
- Created tests for the service layer
- Updated route and route tests to reflect service layer changes

## Context for reviewers

The workflow event API was previously created with a stub for the service layer. The service layer now has validation for starting a workflow and processing a workflow event.

## Validation steps

Confirm all tests pass and all validation checks are accounted for.

### DIFF
--- a/api/src/services/workflows/ingest_workflow_event.py
+++ b/api/src/services/workflows/ingest_workflow_event.py
@@ -6,6 +6,7 @@ import src.workflow.state_machine  # noqa: F401  # Import to register all state 
 from src.adapters import db
 from src.api.route_utils import raise_flask_error
 from src.constants.lookup_constants import WorkflowEventType
+from src.logging.flask_logger import add_extra_data_to_current_request_logs
 from src.workflow.event.workflow_event import ProcessWorkflowEventContext, StartWorkflowEventContext
 from src.workflow.registry.workflow_registry import WorkflowRegistry
 from src.workflow.service.workflow_service import (
@@ -17,7 +18,6 @@ from src.workflow.workflow_errors import (
     EntityNotFound,
     InactiveWorkflowError,
     InvalidEntityForWorkflow,
-    InvalidEventError,
     InvalidWorkflowTypeError,
     WorkflowDoesNotExistError,
 )
@@ -41,10 +41,8 @@ def ingest_workflow_event(db_session: db.Session, json_data: dict[str, Any]) -> 
     event_id = uuid.uuid4()
     event_type = json_data.get("event_type")
 
-    logger.info(
-        "Ingesting workflow event",
-        extra={"event_id": event_id, "event_type": event_type},
-    )
+    add_extra_data_to_current_request_logs({"event_id": event_id, "event_type": event_type})
+    logger.info("Ingesting workflow event")
 
     try:
         if event_type == WorkflowEventType.START_WORKFLOW:
@@ -53,30 +51,29 @@ def ingest_workflow_event(db_session: db.Session, json_data: dict[str, Any]) -> 
             _validate_process_workflow_event(db_session, json_data)
 
     except EntityNotFound as e:
-        logger.warning("Entity not found for workflow event", extra={"event_id": event_id})
-        raise_flask_error(404, str(e))
+        logger.info("Entity not found for workflow event", extra={"error": str(e)})
+        raise_flask_error(404, "The specified resource was not found")
     except WorkflowDoesNotExistError as e:
-        logger.warning("Workflow does not exist", extra={"event_id": event_id})
-        raise_flask_error(404, str(e))
-    except (
-        InvalidWorkflowTypeError,
-        InvalidEntityForWorkflow,
-        InactiveWorkflowError,
-        InvalidEventError,
-    ) as e:
-        logger.warning("Invalid workflow event", extra={"event_id": event_id})
-        raise_flask_error(422, str(e))
+        logger.info("Workflow does not exist", extra={"error": str(e)})
+        raise_flask_error(404, "The specified workflow was not found")
+    except InvalidWorkflowTypeError as e:
+        logger.info("Invalid workflow type", extra={"error": str(e)})
+        raise_flask_error(422, "Invalid workflow type specified")
+    except InvalidEntityForWorkflow as e:
+        logger.info("Invalid entities for workflow", extra={"error": str(e)})
+        raise_flask_error(422, "The provided entities are not valid for this workflow type")
+    except InactiveWorkflowError as e:
+        logger.info("Workflow is not active", extra={"error": str(e)})
+        raise_flask_error(422, "This workflow is not currently active")
 
-    logger.info(
-        "Successfully validated workflow event",
-        extra={"event_id": event_id, "event_type": event_type},
-    )
+    logger.info("Successfully validated workflow event")
 
     return event_id
 
 
 def _validate_start_workflow_event(db_session: db.Session, json_data: dict[str, Any]) -> None:
     """Validate a start_workflow event."""
+    # Note: start_workflow_context is guaranteed to be present due to schema validation
     start_context_data = json_data.get("start_workflow_context", {})
     start_context = StartWorkflowEventContext(**start_context_data)
 
@@ -89,6 +86,7 @@ def _validate_start_workflow_event(db_session: db.Session, json_data: dict[str, 
 
 def _validate_process_workflow_event(db_session: db.Session, json_data: dict[str, Any]) -> None:
     """Validate a process_workflow event."""
+    # Note: process_context_data is guaranteed to be present due to schema validation
     process_context_data = json_data.get("process_workflow_context", {})
     process_context = ProcessWorkflowEventContext(**process_context_data)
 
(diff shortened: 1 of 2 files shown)


api/src/services/workflows/ingest_workflow_event.py
```diff
@@ -2,21 +2,106 @@
...
+        logger.warning("Workflow does not exist", extra={"event_id": event_id})
+        raise_flask_error(404, str(e))
+    except (
+        InvalidWorkflowTypeError,
+        InvalidEntityForWorkflow,
+        InactiveWorkflowError,
+        InvalidEventError,
+    ) as e:
+        logger.warning("Invalid workflow event", extra={"event_id": event_id})
+        raise_flask_error(422, str(e))
```
hao10282025-sudo: should these exceptions be more specific for start and process cases?
means to put separate inside line 50 and line 52

api/src/services/workflows/ingest_workflow_event.py
```diff
@@ -2,21 +2,106 @@
...
 
     logger.info(
-        "Ingested workflow event",
-        extra={"event_id": event_id, "event_type": json_data.get("event_type")},
+        "Successfully validated workflow event",
+        extra={"event_id": event_id, "event_type": event_type},
     )
 
     return event_id
```
hao10282025-sudo, line 71: I am curious for the usage of this event_id

api/src/services/workflows/ingest_workflow_event.py
```diff
@@ -2,21 +2,106 @@
...
+
+
+def _validate_start_workflow_event(db_session: db.Session, json_data: dict[str, Any]) -> None:
```
hao10282025-sudo, line 74: should the validate function return boolean value to be more reasonable?
not sure if any follow up action after this validation

api/src/services/workflows/ingest_workflow_event.py
kkrug: I considered separating the try/except blocks, but since the exception types themselves are specific to each validation path (EntityNotFound can only come from START, WorkflowDoesNotExistError can only come from PROCESS). The exception type already documents which validation failed, so the change would really be more for clarity reading the code rather than a functionality difference.

api/src/services/workflows/ingest_workflow_event.py
kkrug: The event_id is helpful for structured logging and possible debugging.

api/src/services/workflows/ingest_workflow_event.py
```diff
@@ -2,21 +2,106 @@
...
+      - Event is valid for the workflow (422 if not)
     """
+    event_id = uuid.uuid4()
+    event_type = json_data.get("event_type")
 
-    # Model conversion to be implemented
+    logger.info(
+        "Ingesting workflow event",
+        extra={"event_id": event_id, "event_type": event_type},
+    )
```
chouinar: Since there is quite a bit of logging, this'll simplify things a bit to just add these to all subsequent logs.

```suggestion
    event_id = uuid.uuid4()
    event_type = json_data.get("event_type")
    
    add_extra_data_to_current_request_logs({"event_id": event_id, "event_type": event_type})

    logger.info("Ingesting workflow event")
```

api/src/services/workflows/ingest_workflow_event.py
kkrug: Boolean returns (i.e. returning false when something is invalid) would lose error context and require extra steps to communicate what failed. Using the exceptions provides better error handling and makes use of all of the classes Michael created for workflow.

api/src/services/workflows/ingest_workflow_event.py
chouinar: We shouldn't do `str(e)`, the message we pass to `raise_flask_error` is the error message the user will receive. We won't necessarily have written all the messages we pass into our error classes for end users. They'll probably be too specific most of the time / referring to internals.

We'll want to construct error messages that are more suited for the API responses.

api/src/services/workflows/ingest_workflow_event.py
```diff
@@ -2,21 +2,106 @@
...
+            _validate_start_workflow_event(db_session, json_data)
+        elif event_type == WorkflowEventType.PROCESS_WORKFLOW:
+            _validate_process_workflow_event(db_session, json_data)
+
+    except EntityNotFound as e:
+        logger.warning("Entity not found for workflow event", extra={"event_id": event_id})
+        raise_flask_error(404, str(e))
+    except WorkflowDoesNotExistError as e:
+        logger.warning("Workflow does not exist", extra={"event_id": event_id})
+        raise_flask_error(404, str(e))
```
chouinar: We won't want to log warnings, I know I've been doing that in the underlying service, but I also probably shouldn't be doing that in some contexts. Since these scenarios are user-error most likely, they're not problems. I set them up as warnings on the workflow service side as if we're seeing them, that means the logic you've written here let them through most likely, so it's worse at that point to see it.

A 404 isn't a warning/problem we need to highlight, should just be an info log.

For context on how we treat the different levels:
* Debug -> Rarely use, we don't have these enabled anywhere, at most they're used locally and that would require you turn them on
* Info -> Most things should fall into this, especially anything that is informational
* Warning -> We want to highlight an oddity, it might be a problem, but we're not sure
* Error / Exception -> We want to be alerted when this happens, we have alerts setup for when there are any error logs.

api/src/services/workflows/ingest_workflow_event.py
chouinar: Any case where it would be invalid would have raised an exception, so it would only ever return True.

api/src/services/workflows/ingest_workflow_event.py
```diff
@@ -2,21 +2,106 @@
...
+        extra={"event_id": event_id, "event_type": event_type},
     )
 
     return event_id
+
+
+def _validate_start_workflow_event(db_session: db.Session, json_data: dict[str, Any]) -> None:
+    """Validate a start_workflow event."""
+    start_context_data = json_data.get("start_workflow_context", {})
+    start_context = StartWorkflowEventContext(**start_context_data)
```
chouinar, line 78: For this and the process workflow path, since our request schema validates that this will be set, we don't need to check again here, right?

Can we add a comment mentioning that? First glance I was wondering "what happens if it's missing"

api/src/services/workflows/ingest_workflow_event.py
```diff
@@ -2,21 +2,106 @@
...
+    # Get the state machine class for this workflow type
+    _, state_machine_cls = WorkflowRegistry.get_state_machine_for_workflow_type(
+        workflow.workflow_type
+    )
+
+    # Validate the event is valid for this workflow
+    if not is_event_valid_for_workflow(process_context.event_to_send, state_machine_cls):
+        raise InvalidEventError(
+            f"Event '{process_context.event_to_send}' is not valid for this workflow"
+        )
```
chouinar: We could use `raise_flask_error` directly from here.

api/tests/src/services/workflows/test_ingest_workflow_event.py
```diff
@@ -0,0 +1,178 @@
...
+                }
+            ],
+        },
+    }
+
+    with pytest.raises(apiflask.exceptions.HTTPError) as exc_info:
+        ingest_workflow_event(db_session, payload)
+
+    assert exc_info.value.status_code == 422
+    assert "do not match expected types" in exc_info.value.message
```
chouinar: You can check the error message with `match` in pytest.raises - it's actually a regex, but basic strings will work as a contains check.

```suggestion
    with pytest.raises(apiflask.exceptions.HTTPError, match="do not match expected types") as exc_info:
        ingest_workflow_event(db_session, payload)

    assert exc_info.value.status_code == 422
```

api/src/services/workflows/ingest_workflow_event.py
hao10282025-sudo: ok not a problem
I was thinking what to do after the function returns

api/tests/src/services/workflows/test_ingest_workflow_event.py
kkrug: I couldn't get this to work - I think it's because the match parameter in pytest.raises checks against the exception's string representation (str(exception)), but APIFlask's HTTPError stores the message in a .message attribute instead.

When the tests ran with match, they all failed with:
AssertionError: Regex pattern did not match.
 Regex: 'event is not valid for this workflow'
 Input: ''

The "Input" is empty because str(HTTPError) returns an empty string.

api/src/services/workflows/ingest_workflow_event.py
kkrug: Updated

api/src/services/workflows/ingest_workflow_event.py
kkrug: I've updated all the error logging

api/src/services/workflows/ingest_workflow_event.py
kkrug: Updated

api/src/services/workflows/ingest_workflow_event.py
kkrug: Added comments to both workflow paths

api/src/services/workflows/ingest_workflow_event.py
kkrug: Updated

PR #8614, opened 2026-02-20
[Issue #8461] Orphaned attachments are not included in application submission

### LINKED ISSUE #8461: Orphaned attachments are not included in application submission
### Summary
Per alignment from #8180, we want to remove the ability for an orphaned attachment to be submitted with an application.

- Orphaned attachment = A user can add an attachment to a form, and then navigate away from the form without saving it. This results in the form being attached to the application, but not associated with the form.

### Acceptance criteria

- [ ] Orphaned attachments are not included in application submission

mikehgrantsgov:
## Summary

Fixes #8461

## Changes proposed

Added `_collect_referenced_attachment_ids` to `attachment_mapping.py` - it inspects each form's `json_to_xml_schema.attachment_fields` config to collect attachment UUIDs that are actually referenced in a form's application_response
Updated create_attachment_mapping` to skip (log WARN) any attachment whose UUID isn't in that set.
Added new tests

## Context for reviewers

Orphaned attachments occur when a user uploads a file to a form but navigates away without saving. The `ApplicationAttachment` record is created but its UUID never gets written into any ApplicationForm.application_response. 

Orphaned attachments are silently excluded from XML generation but still packaged into the submission zip, so this is a non-breaking change and won't block any submission.

## Validation steps

See new tests


api/src/task/apply/create_application_submission_task.py
```diff
@@ -385,8 +385,10 @@ def process_xml_generation(self, submission: SubmissionContainer) -> None:
         logger.info("Generating XML for application submission", extra=log_extra)
 
         # Create attachment mapping once for all forms
-        attachment_mapping = create_attachment_mapping_from_list(
-            submission.application.application_attachments,
+        # Orphaned attachments (not referenced in any form's application_response)
+        # are automatically excluded by create_attachment_mapping.
+        attachment_mapping = create_attachment_mapping(
```
mikehgrantsgov, line 390: Switching here allows us to detect and exclude orphans, since `create_attachment_mapping_from_list` doesn't know which attachments are referenced or orphaned. It just maps everything in the list.

api/tests/src/services/xml_generation/test_attachment_mapping.py
```diff
@@ -0,0 +1,199 @@
...
+
+@pytest.fixture
+def patch_hash():
+    """Patch AttachmentFile.compute_base64_sha1 to return a stable fake hash."""
+    with patch(
+        "src.services.xml_generation.utils.attachment_mapping.AttachmentFile.compute_base64_sha1",
+        return_value=FAKE_HASH,
+    ) as mock_hash:
+        yield mock_hash
+
```
chouinar: Could we move this to the top of the file, not between other tests?

api/tests/src/services/xml_generation/test_attachment_mapping.py
```diff
@@ -0,0 +1,199 @@
...
+
+def make_attachment(attachment_id: uuid.UUID, file_name: str = "file.pdf") -> MagicMock:
+    """Create a mock ApplicationAttachment."""
+    att = MagicMock()
+    att.application_attachment_id = attachment_id
+    att.file_name = file_name
+    att.mime_type = "application/pdf"
+    att.file_location = f"s3://bucket/applications/{attachment_id}/{file_name}"
+    return att
```
chouinar: I am curious why we would use this (and below functions) instead of a factory? If we don't want anything in the DB / want to keep it simple, use `.build()`. If there is something that our factory can't account for, can you add a comment saying that here?

api/tests/src/services/xml_generation/test_attachment_mapping.py
mikehgrantsgov: Makes sense, moved that

api/tests/src/services/xml_generation/test_attachment_mapping.py
mikehgrantsgov: Makes sense, I switched to Application`AttachmentFactory.build()` and `ApplicationFormFactory.build()`

Also found and fixed a missing return in ApplicationAttachmentFactory._build() that was causing .build() to return `None`.

Issue #5903, opened 2025-08-11
Support attachments in SOAP Proxy without error


chris-kuryak:
### Summary

This is a follow-on ticket to the #5735 based on that investigation.

------------------------------
Currently, SOAP requests with attachments error out due to how the soap proxy API handles incoming requests. There are some operation that include attachment file data within the request. 

Normally, we would just forward request to grants.gov s2s soap api and but something about how we are either receiving or preparing the proxy request to gg is failing.

There are 2 paths of resolving this issue:
1. May just be an encoding issue or an issue with how we are processing requests with attachments and forwarding to grants.gov. There may be a possibility that we can just update logic between receiving the request in the route and then forwarding it to grants.gov in the `get_proxy_response` method.
2. We may need update the logic for processing and extracting the attachments and rebuild the request as started in the draft pull request.

Need to use [MTOM](https://en.wikipedia.org/wiki/Message_Transmission_Optimization_Mechanism) like the legacy system does 
 
@chouinar has a script for reproducing both a successful request to trainingws.grants.gov as well as reproducing the failure locally.

### Acceptance criteria

- [ ] Requests with attachments should not fail for requests proxied to gg.gov s2s soap api.

jakobpederson:
Currently working on this. The streaming fix seems to be a viable solution but also trying to get a good test case has been challenging.

PR #8616, opened 2026-02-20
[Issue #5903] Fix SOAP attachment timeout issue

jakobpederson:
## Summary

Fixes / Work for #5903 

## Changes proposed

Stream the data from the request but also allow a "peak" at the first few thousand characters so we can extract some information from the xml in the message.

## Context for reviewers

If the attachments were too large it could cause a timeout issue. To correct that we stream the data from the request to avoid the timeout.

## Validation steps

 1. Using these steps, [deploy the branch to staging](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/1904214069/Deploy+to+lower+environments)
2. Run command to hit the end point with a test file `python repro_attachment.py` [here]()
3. Observe the two different outcomes as the script will hit both dev and staging:

# File filler size: b"a" * 50 * 1024 * 95
```
*************************
url='https://soap.staging.simpler.grants.gov'
response.status_code=500
response.text='--uuid:855b63b1-5ac9-4ea0-a341-665182e9f9d8\r\nContent-Type: application/xop+xml; charset=UTF-8; type="text/xml"\r\nContent-Transfer-Encoding: binary\r\nContent-ID: <root.message@cxf.apache.org>\r\n\r\n<soap:Envelope xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/"><soap:Body><soap:Fault><faultcode>soap:Server</faultcode><faultstring>Unable to accept application submission: 02</faultstring></soap:Fault></soap:Body></soap:Envelope>\r\n--uuid:855b63b1-5ac9-4ea0-a341-665182e9f9d8--'
*************************
*************************
url='https://soap.dev.simpler.grants.gov'
response.status_code=502
response.text='<html>\r\n<head><title>502 Bad Gateway</title></head>\r\n<body>\r\n<center><h1>502 Bad Gateway</h1></center>\r\n</body>\r\n</html>\r\n'
*************************
```

Most importantly, the call to staging returned a failure due to no user certificate:
`02 | E_NO_USER_CERT | No user certificate`
While the dev environment suffered the time out. 

If you go into the script and lower the size of the file filler, you can actually get the same response (filler set to`b"a" * 50 * 1024`

# File filler size: b"a" * 50 * 1024
```
*************************
url='https://soap.staging.simpler.grants.gov'
response.status_code=500
response.text='--uuid:00c88850-99a0-43c0-bab8-2e9d425dbca2\r\nContent-Type: application/xop+xml; charset=UTF-8; type="text/xml"\r\nContent-Transfer-Encoding: binary\r\nContent-ID: <root.message@cxf.apache.org>\r\n\r\n<soap:Envelope xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/"><soap:Body><soap:Fault><faultcode>soap:Server</faultcode><faultstring>Unable to accept application submission: 02</faultstring></soap:Fault></soap:Body></soap:Envelope>\r\n--uuid:00c88850-99a0-43c0-bab8-2e9d425dbca2--'
*************************
*************************
url='https://soap.dev.simpler.grants.gov'
response.status_code=500
response.text='--uuid:a76cc709-5f95-452a-b1ac-f162acd191b1\r\nContent-Type: application/xop+xml; charset=UTF-8; type="text/xml"\r\nContent-Transfer-Encoding: binary\r\nContent-ID: <root.message@cxf.apache.org>\r\n\r\n<soap:Envelope xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/"><soap:Body><soap:Fault><faultcode>soap:Server</faultcode><faultstring>Unable to accept application submission: 02</faultstring></soap:Fault></soap:Body></soap:Envelope>\r\n--uuid:a76cc709-5f95-452a-b1ac-f162acd191b1--'
*************************
```

### DIFF
--- a/api/tests/src/legacy_soap_api/test_legacy_soap_api_schemas.py
+++ b/api/tests/src/legacy_soap_api/test_legacy_soap_api_schemas.py
@@ -14,7 +14,54 @@ from src.legacy_soap_api.legacy_soap_api_schemas import (
     SoapRequestStreamer,
     SOAPResponse,
 )
-from tests.fixtures.legacy_soap_api.mtom_attachment import add_filler, get_base_body
+
+
+def get_base_body(append_to_end: bytes | None = None) -> bytes:
+    base_body = (
+        b"--uuid:a1368612-206e-4fa7-b8c5-4aec08409929\r\n"
+        b'Content-Type: application/xop+xml; charset=UTF-8; type="text/xml"\r\n'
+        b"Content-Transfer-Encoding: 8bit\r\n"
+        b"Content-ID: <root.message@cxf.apache.org>\r\n\r\n"
+        b'<soapenv:Envelope xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/" '
+        b'xmlns:app="http://apply.grants.gov/services/ApplicantWebServices-V2.0" '
+        b'xmlns:glob="http://apply.grants.gov/system/Global-V1.0" '
+        b'xmlns:head="http://apply.grants.gov/system/Header-V1.0" '
+        b'xmlns:gran="http://apply.grants.gov/system/GrantsCommonElements-V1.0" '
+        b'xmlns:ns2="http://apply.grants.gov/services/ApplicantWebServices-V2.0" '
+        b'xmlns:ns3="http://apply.grants.gov/system/GrantsCommonElements-V1.0">'
+        b"<soapenv:Header>"
+        b"<head:GrantSubmissionHeader>"
+        b"<head:OpportunityID>SIMP-COPYING</head:OpportunityID>"
+        b'<glob:HashValue glob:hashAlgorithm="SHA-1">nZXaeALvYJz6QQFdf1bUvr7S6ts=</glob:HashValue>'
+        b"<head:SubmissionTitle>go get it</head:SubmissionTitle>"
+        b"</head:GrantSubmissionHeader>"
+        b"</soapenv:Header>"
+        b"<soapenv:Body>"
+        b"<app:SubmitApplicationRequest>"
+        b"<app:GrantApplicationXML>"
+        b"&lt;test"
+        b"</app:GrantApplicationXML>"
+        b"<gran:Attachment>"
+        b"<gran:FileContentId>budget</gran:FileContentId>"
+        b"<gran:FileDataHandler>"
+        b'<xop:Include xmlns:xop="http://www.w3.org/2004/08/xop/include" href="cid:1157990985709"/>'
+        b"</gran:FileDataHandler>"
+        b"</gran:Attachment>"
+        b"</app:SubmitApplicationRequest>"
+        b"</soapenv:Body>"
+        b"\r\n--uuid:a1368612-206e-4fa7-b8c5-4aec08409929\r\n"
+        b"Content-Type: application/octet-stream\r\n"
+        b"Content-Transfer-Encoding: binary\r\n"
+        b"Content-ID: <1157990985709>\r\n\r\n"
+        b"text"
+    )
+    if append_to_end:
+        base_body += append_to_end
+    return base_body
+
+
+def add_filler(base_body: bytes, size: int = 10000) -> bytes:
+    return base_body + b"a" * size
 
 
 def xml_bytes() -> bytes:


api/src/legacy_soap_api/legacy_soap_api_utils.py
```diff
@@ -350,3 +350,50 @@ def to_snake_case(name: str) -> str:
...
+            chunk_count += 1
+        return content
+
+    def head(self) -> bytes:
+        return self.head_bytes
+
+    def __iter__(self) -> Generator[bytes]:
+        if not self._consumed_head:
+            yield self.head_bytes
+            self._consumed_head = True
```
chouinar: Just checking, with a binary stream, is it possible to move it back to the start.

ie. would:

```py
self.head_bytes = self.get_head_bytes()
self.stream.seek(0) # tell it to go back to the start - could also put in the function, not sure if this is the right way to do it
```

And then in `__iter__` we don't need to merge our fetched value with the stream.

If that doesn't work, that's fine - just wanted to check as it'd be slightly cleaner.

api/src/legacy_soap_api/legacy_soap_api_utils.py
```diff
@@ -350,3 +350,50 @@ def to_snake_case(name: str) -> str:
...
+
+
+class SoapRequestStreamer:
+    def __init__(
+        self, stream: BinaryIO | IO[bytes], chunk_count: int = 4, content_length: int | None = None
+    ) -> None:
+        self.stream = stream
+        self.terminators = [b"</soapenv:Envelope>", b"</env:Envelope>"]
+        self.chunk_count = chunk_count
+        self.chunk_size = 2000
```
chouinar: If terminators and chunk size are static values that we don't override anywhere, can we move them out of the class and statically define them somewhere?

chouinar (commented):
+1 to @freddieyebra 's comments about testing. I know the actual testing for this might be quite complex to setup, we should have that documented for future reference somewhere - just so we don't have to figure it out again.

api/tests/src/legacy_soap_api/test_legacy_soap_api_utils.py
```diff
@@ -273,3 +275,241 @@ def test_diff_soap_dicts_list_of_dicts_with_value_diff():
 def test_hidden_value():
     assert _hide_value("a", False) == "a"
     assert _hide_value("a", True) == HIDDEN_VALUE
+
+
+def test_soap_request_streamer_head_defaults_to_10000_chars():
+    data_1 = (
```
freddieyebra: Not a blocker but in the future, can we move some of these larger payloads into fixture files or separate functions similar to how we do [here](https://github.com/HHS/simpler-grants-gov/tree/main/api/tests/fixtures/legacy_soap_api) or [here](https://github.com/HHS/simpler-grants-gov/tree/main/api/tests/src/legacy_soap_api/soap_request_templates)

api/src/legacy_soap_api/legacy_soap_api_schemas/base.py
```diff
@@ -33,7 +32,7 @@ class SOAPInvalidRequestOperationName(Exception):
 
 
 class SOAPRequest(BaseModel):
-    data: bytes
+    data: Any
```
freddieyebra: Can we ensure the typing here correctly reflects the updates in the PR? There may be some changes needed to make the `SoapRequestStreamer` class inherit the `BaseModel` class but it looks like there are some instances of the `SOAPRequest()` class that take in byte strings and some that take in a `SoapRequestStreamer` class instance.

api/src/legacy_soap_api/legacy_soap_api_utils.py
jakobpederson: It is cleaner but from what I can tell the request.stream doesn't appear to be seekable.

api/src/legacy_soap_api/legacy_soap_api_schemas.py
```diff
@@ -35,7 +33,7 @@ class SOAPInvalidRequestOperationName(Exception):
 
 
 class SOAPRequest(BaseModel):
-    data: bytes
+    data: Any
```
freddieyebra, line 36: Looks like this is an existing duplicated `SOAPRequest` class here. Can we create a follow up ticket to address this duplicate instance?

api/src/legacy_soap_api/legacy_soap_api_utils.py
```diff
@@ -350,3 +352,48 @@ def to_snake_case(name: str) -> str:
...
+        self.stream = stream
+        self.chunk_count = chunk_count
+        self._consumed_head = False
+        self.head_bytes = self.get_head_bytes()
+        self.total_length = content_length
+
+    def get_head_bytes(self) -> bytes:
+        buffer = io.BytesIO()
+        chunk_count = 0
+        while chunk_count <= self.chunk_count:
```
freddieyebra: Should this be 
```suggestion
        while chunk_count < self.chunk_count:
```

If self.chunk_size = 2 the current implementation would loop 3 times. Is that intentional?

1st loop 0
2nd loop 1 (should stop here?)
3rd loop 2

api/src/legacy_soap_api/legacy_soap_api_utils.py
```diff
@@ -350,3 +352,48 @@ def to_snake_case(name: str) -> str:
...
+            chunk = self.stream.read(CHUNK_SIZE)
+            if not chunk:
+                break
+            buffer.write(chunk)
+            content = buffer.getvalue()
```
freddieyebra: This would be evaluating the entire buffer contents on every iteration. could we optimize this to only check for the closing tags on the current chunk before writing it to the buffer that contains the full content?

api/src/legacy_soap_api/legacy_soap_api_utils.py
```diff
@@ -350,3 +352,48 @@ def to_snake_case(name: str) -> str:
...
+        while True:
+            chunk = self.stream.read(CHUNK_SIZE)
+            if not chunk:
+                break
+            yield chunk
+
+    def __len__(self) -> int:
+        if self.total_length is not None:
+            return self.total_length
+        return 0
```
freddieyebra: Can we default the self.total_length to 0 so we can just return `self.total_length` instead of checking for None.

api/tests/src/legacy_soap_api/grantors/schemas/test_legacy_soap_grantor_get_application_zip_schema.py
```diff
@@ -114,7 +115,7 @@ def test_get_pydantic_error_if_request_xml_does_not_have_tracking_number(self, d
             "</soapenv:Envelope>"
         ).encode("utf-8")
         soap_request = SOAPRequest(
-            data=request_xml_bytes,
+            data=SoapRequestStreamer(io.BytesIO(request_xml_bytes)),
```
freddieyebra: Do we need to update other instantiations of the SOAPRequest()? There are some that take in `data=b"..."`, some with `data="xmlstring"` and others that take in  `data=SoapRequestStreamer(io.BytesIO(request_xml_bytes))`. Is that intended?

api/tests/src/legacy_soap_api/test_legacy_soap_api_utils.py
```diff
@@ -273,3 +275,241 @@ def test_diff_soap_dicts_list_of_dicts_with_value_diff():
...
+        b"Content-Transfer-Encoding: binary\r\n"
+        b"Content-ID: <1157990985709>\r\n\r\n"
+        b"text"
+    )
+    data_2 = b"a" * 9000
+    fake_stream = io.BytesIO(data_1 + data_2)
+    soap_request_streamer = SoapRequestStreamer(fake_stream, chunk_count=1)
+    expected = data_1 + b"a" * 2534
+    assert len(soap_request_streamer.head()) == 4000
+    assert soap_request_streamer.head() == expected
```
freddieyebra: I am not sure i understand this test, can you explain a bit what this is doing?

api/tests/src/legacy_soap_api/test_legacy_soap_api_utils.py
jakobpederson: Moved them to the fixtures file under `mtom_attachments.py`. I'm not 100% sure about the name.

api/src/legacy_soap_api/legacy_soap_api_utils.py
jakobpederson: I fixed this, it was causing basically an issue where the number of chunks weren't matching up the way you would expect

api/tests/src/legacy_soap_api/test_legacy_soap_api_utils.py
jakobpederson: Leftover test from when you could modify chunks, I'll pull it.

api/src/legacy_soap_api/legacy_soap_api_schemas.py
jakobpederson: Wrote up a ticket for refinement [here](https://github.com/HHS/simpler-grants-gov/issues/8631).

api/tests/src/legacy_soap_api/test_legacy_soap_api_utils.py
jakobpederson: Reversed that because it kept causing failures but I'll keep it in mind going forward.

api/tests/src/legacy_soap_api/grantors/schemas/test_legacy_soap_grantor_get_application_zip_schema.py
jakobpederson: Updated all the SOAPRequests.

Issue #8007, opened 2026-01-15
Opportunity Table CRUD Endpoint - Update


brianlee-nava:
### Summary

**Background and Context**
There are 4 main tables within the Opportunity Data Model that Plan and Publish will need to work with. Those are: Opportunity, Opportunity Attachment, Opportunity Summary, Opportunity Assistance Listing.

In order to make Opportunity Publishing work in Simpler Grants Management, we need to rebuild endpoints. The basic CRUD Create, Read, Update, Delete operations for each table account for 16 endpoints, with additional simple endpoints for publishing and querying the Opportunity Assistance Listing table. **This ticket is scoped for building the Update endpoint for the Opportunity table.**

**User Story**
As a grantor system
I want to update an existing Opportunity
So that changes can be saved during drafting

### Acceptance criteria

Given an existing Opportunity ID and valid updates
When the update Opportunity endpoint is called
Then the Opportunity record is updated
And the response returns a 200 status

Given an invalid update payload
When the update Opportunity endpoint is called
Then the response returns a 400 status

- [ ] Gate the endpoint (API Key and user privileges)
- [ ] Add to API.grants.gov page lower environment for testing
- [ ] Test for missing required fields
- [ ] If Update command fails, return a descriptive error message

desaiamit:
Hi @brianlee-nava, @dghazvini , @chouinar  — I'm working on this ticket and wanted to confirm the scope of the update endpoint before finalizing.

  The Opportunity table has the following fields: opportunity_title, category, category_explanation, agency_code, is_draft, and revision_number.

  For this ticket, I've implemented the PUT endpoint exposing only opportunity_title, category, and category_explanation as updatable fields, since the others appear to be system-managed (e.g. agency_code ties the
   opportunity to an agency, is_draft and revision_number would likely be managed by dedicated publish/draft workflows).

  Can you confirm this is the right scope, or are there additional fields from the Opportunity table that should be updatable via this endpoint?

chouinar:
@desaiamit - Yeah, the fields are pretty limited since most of an opportunity really lives in other tables, should just be those 3.

PR #8620, opened 2026-02-23
[Issue #8007] Add PUT endpoint to update opportunity

desaiamit:
## Summary

Work for #8007

## Changes proposed

- Added `PUT /v1/grantors/opportunities/<opportunity_id>` endpoint to allow grantor users to update an existing opportunity
- Added `OpportunityUpdateRequestSchema` with required fields (opportunity_title, category) and optional category_explanation
- Added `OpportunityUpdateResponseSchema` returning the full updated opportunity
- Added `update_opportunity` service function that:
  - Fetches the opportunity by ID (404 if not found)
  - Verifies the authenticated user has `UPDATE_OPPORTUNITY` privilege for the opportunity's agency (403 if not)
  - Rejects updates to non-draft opportunities (422)
  - Always updates all fields
- Supports both JWT (`X-SGG-Token`) and API key (`X-API-Key`) authentication via `jwt_or_api_user_key_multi_auth`

## Context for reviewers

The endpoint follows the same pattern as the existing `POST /v1/grantors/opportunities/` create endpoint:
- Uses `jwt_or_api_user_key_multi_auth` for auth, supporting both JWT tokens and API user keys
- Uses `verify_access(user, {Privilege.UPDATE_OPPORTUNITY}, opportunity.agency_record)` to enforce privilege checks
- After flushing the update, reloads the opportunity with `selectinload("*")` and `selectinload(Opportunity.agency_record).selectinload(Agency.top_level_agency)` to avoid `DetachedInstanceError` when serializing the response

Note: The implementation exposes `opportunity_title`, `category`, and `category_explanation` as updatable fields. Other Opportunity table fields (`agency_code`, `is_draft`, `revision_number`) are system-managed and intentionally excluded.

## Validation steps

1. Start the local API
2. Run `make db-seed-local-with-agencies` to seed test users
3. In Postman, create a draft opportunity to test with:
    - Method: POST
    - URL: http://localhost:8080/v1/grantors/opportunities/
    - Header: X-API-Key: one_agency_opp_edit_key
    - Body: { "opportunity_title": "Test Opportunity", "category": "discretionary" }
    - Note the opportunity_id from the response
4. Send a PUT request to update it:
    - Method: PUT
    - URL: http://localhost:8080/v1/grantors/opportunities/<opportunity_id>
    - Header: X-API-Key: one_agency_opp_edit_key
    - Body: { "opportunity_title": "Updated Title", "category": "research" }
    - Verify response is 200 with updated fields in the response body
5. Verify that omitting a required field (e.g. category) returns 422
6. Verify that using an opportunity not belonging to the user's agency returns `403`
7. Verify that a non-existent opportunity ID returns `404`
8. Verify that attempting to update a published (non-draft) opportunity returns `422`
9. Run the test suite: `make test args="tests/src/api/opportunities_grantors_v1/test_opportunity_route_update.py"`

### DIFF
--- a/api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
+++ b/api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
@@ -60,6 +60,7 @@ class OpportunityCreateRequestSchema(Schema):
         },
     )
     category_explanation = fields.String(
+        allow_none=True,
         validate=validators.Length(max=255),
         metadata={
             "description": "Explanation of the category (required when category is 'other')",
@@ -126,6 +127,8 @@ class OpportunityUpdateRequestSchema(Schema):
         },
     )
     category_explanation = fields.String(
+        allow_none=True,
+        load_default=None,
         validate=validators.Length(max=255),
         metadata={
             "description": "Explanation of the category (required when category is 'other')",
--- a/api/src/services/opportunities_grantor_v1/opportunity_update.py
+++ b/api/src/services/opportunities_grantor_v1/opportunity_update.py
@@ -2,12 +2,12 @@ import logging
 import uuid
 
 import src.adapters.db as db
-from src.api.route_utils import raise_flask_error
 from src.auth.endpoint_access_util import verify_access
 from src.constants.lookup_constants import Privilege
 from src.db.models.opportunity_models import Opportunity
 from src.db.models.user_models import User
-from src.services.opportunities_grantor_v1.get_opportunity import _get_opportunity_for_grantors
+from src.services.opportunities_grantor_v1.get_opportunity import get_opportunity_for_grantors
+from src.services.opportunities_grantor_v1.opportunity_utils import validate_opportunity_is_draft
 
 logger = logging.getLogger(__name__)
 
@@ -15,17 +15,12 @@ logger = logging.getLogger(__name__)
 def update_opportunity(
     db_session: db.Session, user: User, opportunity_id: uuid.UUID, opportunity_data: dict
 ) -> Opportunity:
-    opportunity = _get_opportunity_for_grantors(db_session, opportunity_id)
-
-    if opportunity is None:
-        raise_flask_error(404, message=f"Could not find Opportunity with ID {opportunity_id}")
+    opportunity = get_opportunity_for_grantors(db_session, user, opportunity_id)
 
     # Check if user has permission to update opportunities for this agency
     verify_access(user, {Privilege.UPDATE_OPPORTUNITY}, opportunity.agency_record)
 
-    # Only draft opportunities can be updated
-    if not opportunity.is_draft:
-        raise_flask_error(422, message="Only draft opportunities can be updated")
+    validate_opportunity_is_draft(opportunity)
 
     # PUT endpoint — always update all fields
     for field, value in opportunity_data.items():
--- a/api/src/services/opportunities_grantor_v1/opportunity_utils.py
+++ b/api/src/services/opportunities_grantor_v1/opportunity_utils.py
@@ -1,0 +1,12 @@
+from src.api.route_utils import raise_flask_error
+from src.db.models.opportunity_models import Opportunity
+
+
+def validate_opportunity_is_draft(opportunity: Opportunity) -> None:
+    """Raise a 422 error if the opportunity is not a draft.
+
+    Only draft opportunities can be modified. This check is shared across
+    all update endpoints to ensure consistent enforcement.
+    """
+    if not opportunity.is_draft:
+        raise_flask_error(422, message="Only draft opportunities can be updated")
--- a/api/tests/src/api/opportunities_grantors_v1/test_opportunity_route_update.py
+++ b/api/tests/src/api/opportunities_grantors_v1/test_opportunity_route_update.py
@@ -12,7 +12,7 @@ def grantor_auth_data(db_session, enable_factory_create):
     """Create a user with UPDATE_OPPORTUNITY permission and return auth data"""
     user, agency, token, api_key_id = create_user_in_agency_with_jwt_and_api_key(
         db_session=db_session,
-        privileges=[Privilege.UPDATE_OPPORTUNITY],
+        privileges=[Privilege.VIEW_OPPORTUNITY, Privilege.UPDATE_OPPORTUNITY],
     )
     return user, agency, token, api_key_id
 


api/src/services/opportunities_grantor_v1/opportunity_update.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def get_opportunity(db_session: db.Session, opportunity_id: uuid.UUID) -> Opportunity:
+    stmt = select(Opportunity).where(Opportunity.opportunity_id == opportunity_id)
+    opportunity = db_session.execute(stmt).scalar_one_or_none()
+
+    if opportunity is None:
+        raise_flask_error(404, message=f"Could not find Opportunity with ID {opportunity_id}")
+
+    return opportunity
```
chouinar: For this, I'd suggest we move this to its own file / this is eventually going to be the same function that the new GET opportunity endpoint uses.

The benefit of having it central is that we can reuse it in every endpoint so we don't need to do the fetch an opportunity or 404 logic in each of them.

I know the work for building that endpoint is going on in parallel, so fine if we don't change it right this second, but should at least note it down to do it later.

api/src/services/opportunities_grantor_v1/opportunity_update.py
```diff
@@ -0,0 +1,67 @@
...
+    # selectinload("*") eagerly loads all relationships to avoid DetachedInstanceError
+    stmt = (
+        select(Opportunity)
+        .options(
+            selectinload("*"),
+            selectinload(Opportunity.agency_record).selectinload(Agency.top_level_agency),
+        )
+        .filter(Opportunity.opportunity_id == opportunity_id)
+    )
+    opportunity = db_session.execute(stmt).scalar_one()
```
chouinar: None of this should be needed. Anything fetched from the DB will be in the DB session. There isn't anything we need to flush, and the select here would be duplicating what the GET function should have done.

I'm guessing we need to make the GET function above fetch more and that'll solve any issues you're hitting. It should have the `selectinload` pieces.

However, don't do `selectinload("*")` - that fetches every relationship from an opportunity which ends up being about half the DB. We should only have the selectinloads for relationships we want - I'd suggest chatting with @dghazvini who I think is doing the GET opportunity endpoint to work out what we need to fetch.

```suggestion
```

api/src/services/opportunities_grantor_v1/opportunity_update.py
```diff
@@ -0,0 +1,67 @@
...
+    if not opportunity.is_draft:
+        raise_flask_error(422, message="Only draft opportunities can be updated")
+
+    # Apply updates — only update fields that were provided
+    if "opportunity_title" in opportunity_data:
+        opportunity.opportunity_title = opportunity_data["opportunity_title"]
+    if "category" in opportunity_data:
+        opportunity.category = opportunity_data["category"]
+    if "category_explanation" in opportunity_data:
+        opportunity.category_explanation = opportunity_data["category_explanation"]
```
chouinar: Since this is a PUT endpoint, we should always update every field.

We do that intentionally since the update approach ends up more complex (for us and the frontend). We should instead just do:
```py
opportunity.x = opportunity_data["x"]
```

For each field.

If you want to be even fancier, can do:
```py
for field, value in opportunity_data.items():
    setattr(opportunity, field, value)
```

Since the request fields == DB column names.

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
```diff
@@ -72,6 +72,36 @@ class OpportunityGrantorSchema(OpportunityV1Schema):
...
+        metadata={
+            "description": "The opportunity category",
+        },
+    )
+    category_explanation = fields.String(
+        validate=validators.Length(max=255),
+        metadata={
+            "description": "Explanation of the category (required when category is 'other')",
+        },
+    )
```
chouinar, line 136: We'll want these to be required/non-null like the create endpoint - since we're doing a PUT rather than an UPDATE - it's a replacement, not an update (we do that because its much simpler for everyone).

api/src/services/opportunities_grantor_v1/opportunity_update.py
```diff
@@ -0,0 +1,67 @@
...
+    return opportunity
+
+
+def update_opportunity(
+    db_session: db.Session, user: User, opportunity_id: uuid.UUID, opportunity_data: dict
+) -> Opportunity:
+    opportunity = get_opportunity(db_session, opportunity_id)
+
+    # Check if user has permission to update opportunities for this agency
+    verify_access(user, {Privilege.UPDATE_OPPORTUNITY}, opportunity.agency_record)
```
chouinar, line 21: I _think_ it doesn't matter, but as a heads up, `agency_record` can be null in very rare cases for opportunities (generally either test data in lower envs, OR some very rare cases of 15 year old opps in prod).

If it is null, it means the last param here is `None` and so no agency-user records end up getting picked up.

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
desaiamit: Updated opportunity_title and category to required=True. Left category_explanation optional to match the create endpoint, since it's a genuinely nullable field.

api/src/services/opportunities_grantor_v1/opportunity_update.py
desaiamit: Updated. Since this is a PUT (replacement), all fields are now always written. Used the setattr loop since the request field names match the DB column names.

api/src/services/opportunities_grantor_v1/opportunity_update.py
desaiamit: I will coordinate with @dghazvini regarding this comment

api/src/services/opportunities_grantor_v1/opportunity_update.py
desaiamit: I will coordinate with @dghazvini regarding this comment.

api/src/services/opportunities_grantor_v1/opportunity_update.py
desaiamit: Done — removed the local get_opportunity function and switched to using _get_opportunity_for_grantors from get_opportunity.py which was introduced in PR #8649. The shared function also handles the proper
  selectinload of all relationships.

api/src/services/opportunities_grantor_v1/opportunity_update.py
desaiamit: Done — removed the db_session.flush() and the re-query block entirely. Since _get_opportunity_for_grantors already eagerly loads all necessary relationships via selectinload (summaries, agency_record, assistance listings, attachments, competitions), the opportunity object is fully populated after the initial fetch and can be returned directly after the setattr updates.

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
chouinar: I'd suggest making category explanation have `allow_none=True` (here and in create). Nullability != required in json schema.

Even if we treat leaving a field out as null equivalent, we want to also allow `{"category_explanation": null}` in the json itself.

api/src/services/opportunities_grantor_v1/opportunity_update.py
```diff
@@ -0,0 +1,39 @@
...
+
+    # Only draft opportunities can be updated
+    if not opportunity.is_draft:
+        raise_flask_error(422, message="Only draft opportunities can be updated")
+
+    # PUT endpoint — always update all fields
+    for field, value in opportunity_data.items():
+        setattr(opportunity, field, value)
```
chouinar, line 27: To make this work as expected, for any nullable field, the JSON schema needs to set a `load_default` (a default when doing JSON -> python dict).

If you want to see why, we should add a test. Let's have an opportunity with a category explanation set to any value. Then pass in a request with the category explanation not in the request. It's not required, so that's a valid request, but it won't be in this dict so it'll not be nulled out as currently implemented.

api/src/services/opportunities_grantor_v1/opportunity_update.py
```diff
@@ -0,0 +1,39 @@
...
+logger = logging.getLogger(__name__)
+
+
+def update_opportunity(
+    db_session: db.Session, user: User, opportunity_id: uuid.UUID, opportunity_data: dict
+) -> Opportunity:
+    opportunity = _get_opportunity_for_grantors(db_session, opportunity_id)
+
+    if opportunity is None:
+        raise_flask_error(404, message=f"Could not find Opportunity with ID {opportunity_id}")
```
chouinar: Reuse the function that does the check for opportunity being null and 404s for consistency. We should avoid importing things that start with `_` - python won't disallow it, but it's the pythonic way of saying something is private.

It's fine if we this means we check the auth twice - technically anyone doing an update _should_ have GET access as well.

```suggestion
    opportunity = get_opportunity_for_grantors(db_session, opportunity_id)
```

api/src/services/opportunities_grantor_v1/opportunity_update.py
```diff
@@ -0,0 +1,39 @@
...
+
+    # Check if user has permission to update opportunities for this agency
+    verify_access(user, {Privilege.UPDATE_OPPORTUNITY}, opportunity.agency_record)
+
+    # Only draft opportunities can be updated
+    if not opportunity.is_draft:
+        raise_flask_error(422, message="Only draft opportunities can be updated")
```
chouinar: I would suggest moving this check into a function that we call in some shared service file since we'll want it on every update endpoint.

Plus in the future we'll likely want to make this be allowed at least partially and having a single thing to update will be easier.

api/src/services/opportunities_grantor_v1/opportunity_update.py
desaiamit: Updated to use get_opportunity_for_grantors (the public function).

api/src/services/opportunities_grantor_v1/opportunity_update.py
desaiamit: Moved the is_draft check into validate_opportunity_is_draft() in a new opportunity_utils.py shared service file. All future update endpoints can import and call it from one place.

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
desaiamit: Added allow_none=True so the API accepts {"category_explanation": null} explicitly. Added load_default=None so when the field is omitted from the request body, it defaults to None, ensuring setattr always clears any existing value

api/src/services/opportunities_grantor_v1/opportunity_update.py
desaiamit: - added allow_none=True to category_explanation in both the create and update schemas so {"category_explanation": null} is accepted explicitly
- added load_default=None to category_explanation in the update schema

chouinar (approved):

Issue #8564, opened 2026-02-17
Update SQS client to be able to send a message


chouinar:
### Summary

We added a client for interacting with SQS, but we didn't add the ability to send a message, only receive + delete.

We want to add that logic as well to this client.

Update: https://github.com/HHS/simpler-grants-gov/blob/main/api/src/adapters/aws/sqs_adapter.py
We want to call `send_message` - https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs/client/send_message.html

The method we write should take in:
* A message body - a dictionary of unspecified format (we'll build utils that wrap this to build the format for workflows)

The queue URL is already on the client class, and any of the other params aren't likely something we need to be configurable at the moment.

The response should return a python typed object with the 5 fields described in the doc (I doubt we'll really use anything beyond the `messageId` but no reason to skip out on the others).

### Acceptance criteria

- [ ] SQS client adds a function to send a message to the SQS queue
- [ ] Any basic logging/error handling is included
- [ ] Tests are added that use moto3 (the mock boto3) under the hood

PR #8625, opened 2026-02-23
[Issue #8564] Update SQS client to be able to send a message + tests

kkrug:
## Summary

Fixes for #8564  

## Changes proposed

- Updated sqs_adapter.py to add functionality to send a message.
- Updated tests_sqs_adapter.py to test send message functionality.

## Context for reviewers

We added a client for interacting with SQS, but we didn't add the ability to send a message, only receive + delete. The method takes in a message body (a dictionary of unspecified format) and returns a python typed object with the 5 fields described in the [doc](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs/client/send_message.html).

## Validation steps

Confirm tests pass and error logging is handled.

### DIFF
--- a/api/tests/src/adapters/aws/test_sqs_adapter.py
+++ b/api/tests/src/adapters/aws/test_sqs_adapter.py
@@ -1,3 +1,4 @@
+import json
 import logging
 from unittest.mock import Mock, patch
 
@@ -10,6 +11,7 @@ from src.adapters.aws.sqs_adapter import (
     SQSClient,
     SQSConfig,
     SQSDeleteBatchResponse,
+    SQSSendMessageResponse,
     get_boto_sqs_client,
 )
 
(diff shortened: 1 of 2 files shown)


api/tests/src/adapters/aws/test_sqs_adapter.py
```diff
@@ -131,3 +133,82 @@ def test_delete_message_batch_partial_failure(self, workflow_sqs_queue):
...
+
+        message_body = {"test": "data"}
+
+        with pytest.raises(ClientError):
+            with caplog.at_level(logging.ERROR):
+                sqs_client.send_message(message_body)
+
+        assert "Failed to send message to SQS" in caplog.text
+        sqs_record = next(r for r in caplog.records if r.message == "Failed to send message to SQS")
+        assert sqs_record.queue_url == invalid_url
```
hao10282025-sudo, line 197: quick check, this test case is mocking the AWS SQS service outage scenario?

kkrug: This is validating that the errors are logged with queue URL context

hao10282025-sudo: I got that part.
What I mean is the specific scenario (use case in reality like service down, network issue or any other) this test case simulates.
I googled online and see the service down can be simulated by giving invalid input. I don't see the mock_sqs is being used in the function and wonder how the moto works for it

chouinar: Don't know the full internals of moto, but I assume internally it just has something like an in-memory map of queue and messages.

```py
# likely a lot more going on, but at its core, probably as simple as this
queues = {
  "queue1": [ message1, message2 ],
  "queue2": [ message3 ],
}
```

In this case we asked for `queueXYZ` which doesn't exist in the map, so it returns the same error that AWS would if it couldn't find the queue.

hao10282025-sudo: ok I see thanks.
I actually not see much details online either, it looks like only put invalid url and that's it

Issue #8433, opened 2026-02-06
[BE] Implement organization saved opportunities filter to saved opportunities endpoint


freddieyebra:
### Summary

Currently, the list user saved opportunities endpoint returns a list of opportunities that the user has saved. We need to support including organization saved opportunities from the `organization_saved_opportunities` table.

We will need to add a filter to the `SavedOpportunityFilterParams` that will be  `organization_ids` that takes in a list of organization ids that the user is a member of. 

Use cases:
1. User saved opportunities only (current functionality)
organization_ids=None
log message = "user saved opportunities requested"

2. Organizations + user saved opportunities
organization_ids=[]
log message = "all saved opportunities requested"

3. Organization saved opportunities only
organization_ids=["123", "456"]
log message = "organization saved opportunities requested"

Successful response: 
```json
{
  "data": [
    {
      "opportunity_id": "9a86dd26-ba3a-4a85-819c-dc48d3cde65d",
      "opportunity_status": "posted",
      "opportunity_title": "Program 1",
      "summary": {
          "close_date": "2026-02-16",
          "is_forecast": false,
          "post_date": "2026-01-20"
      }
    },
    {
      "opportunity_id": "6121ac3b-e7f5-4bdc-a2b3-f1775a17145b",
      "opportunity_status": "posted",
      "opportunity_title": "Program 2",
      "summary": {
          "close_date": "2026-02-16",
          "is_forecast": false,
          "post_date": "2026-01-12"
      }
    }
  ],
  "message": "Success",
  "status_code": 200
}
```

Unauthorized:
```json
{
  "data": {},
  "message": "Unauthorized",
  "status_code": 401
}
```

User does not have `VIEW_ORG_SAVED_OPPORTUNITIES` privilege to any passed organization ids:
```json
{
  "data": {},
  "message": "User does not have access to requested organization(s) saved opportunities",
  "status_code": 403
}
```

No organization found by id:
```json
{
  "data": {},
  "message": "Organization not found",
  "status_code": 404
}
```

### Acceptance criteria
- [ ] A new endpoint that handles the above scenarios are implemented.
- [ ] Sample request included in implementation ticket.
- [ ] Unit tests pass
- [ ] New filter param that is list of organization.organization_id default to `None`
- [ ] Include saved organization saved opportunities for organizations specified in filter
- [ ] User must have the `VIEW_ORG_SAVED_OPPORTUNITIES` privilege for all organizations included in filter list
[...]

PR #8632, opened 2026-02-24
[Issue #8433] Modify user saved opportunity endpoint to include organization saved opportunities

babebe:
## Summary

Fixes / Work for #8433  

## Changes proposed

- Adds support for organization saved opportunities in the saved opportunities endpoint.
- Updated request schema to support `organization_ids` (list of org IDs).
- Updated query logic to include:
  - User-saved opportunities
  - Organization-saved opportunities
- Ensured:
  - No duplicate opportunities
  - Most recent `saved_at` determines sort order
- Update organization access checks:
  - Added a new privilege `VIEW_ORGANIZATION` 
  - Assigned `VIEW_ORGANIZATION` to static roles `ORG_ADMIN` and `ORG_MEMBER`.
- Updated and expanded test coverage.

## Context for reviewers

This change extends the saved opportunities logic to merge user and organization saves using a unified subquery and aggregation. The current query is intentionally structured to be extendable to add metadata (which org saved an opportunity..) without major restructuring.

## Validation steps

* Verify user-only saves return correctly.
* Verify org-only saves return correctly.
* Verify combined user + org saves return no duplicates.
* Verify invalid org ID returns 404.
* Verify unauthorized org access returns 403.

### DIFF
--- a/api/src/api/opportunities_grantor_v1/opportunity_grantor_routes.py
+++ b/api/src/api/opportunities_grantor_v1/opportunity_grantor_routes.py
@@ -15,6 +15,7 @@ from src.services.opportunities_grantor_v1.get_opportunity_list import (
     get_opportunity_list_for_grantors,
 )
 from src.services.opportunities_grantor_v1.opportunity_creation import create_opportunity
+from src.services.opportunities_grantor_v1.opportunity_update import update_opportunity
 
 logger = logging.getLogger(__name__)
 
@@ -90,5 +91,31 @@ def opportunity_get(db_session: db.Session, opportunity_id: UUID) -> response.Ap
         db_session.add(user)
 
         opportunity = get_opportunity_for_grantors(db_session, user, opportunity_id)
+
+    return response.ApiResponse(message="Success", data=opportunity)
+
+
+@opportunity_grantor_blueprint.put("/opportunities/<uuid:opportunity_id>")
+@opportunity_grantor_blueprint.input(
+    opportunity_grantor_schemas.OpportunityUpdateRequestSchema, location="json"
+)
+@opportunity_grantor_blueprint.output(opportunity_grantor_schemas.OpportunityUpdateResponseSchema())
+@jwt_or_api_user_key_multi_auth.login_required
+@opportunity_grantor_blueprint.doc(
+    responses=[200, 401, 403, 404, 422], security=jwt_or_api_user_key_security_schemes
+)
+@flask_db.with_db_session()
+def opportunity_update(
+    db_session: db.Session, opportunity_id: UUID, json_data: dict
+) -> response.ApiResponse:
+    """Update an existing opportunity"""
+    add_extra_data_to_current_request_logs({"opportunity_id": opportunity_id})
+    logger.info("PUT /v1/grantors/opportunities/:opportunity_id")
+
+    with db_session.begin():
+        user = jwt_or_api_user_key_multi_auth.get_user()
+        db_session.add(user)
+
+        opportunity = update_opportunity(db_session, user, opportunity_id, json_data)
 
     return response.ApiResponse(message="Success", data=opportunity)
--- a/api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
+++ b/api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
@@ -60,6 +60,7 @@ class OpportunityCreateRequestSchema(Schema):
         },
     )
     category_explanation = fields.String(
+        allow_none=True,
         validate=validators.Length(max=255),
         metadata={
             "description": "Explanation of the category (required when category is 'other')",
@@ -105,6 +106,40 @@ class OpportunityGrantorSchema(OpportunityV1Schema):
         fields.Nested(CompetitionAlphaSchema),
         metadata={"description": "List of competitions associated with the opportunity"},
     )
+
+
+class OpportunityUpdateRequestSchema(Schema):
+    """Schema for PUT /v1/grantors/opportunities/<opportunity_id> request"""
+
+    opportunity_title = fields.String(
+        required=True,
+        validate=validators.Length(max=255),
+        metadata={
+            "description": "The title of the opportunity",
+            "example": "Updated Research Grant for Climate Innovation",
+        },
+    )
+    category = fields.Enum(
+        OpportunityCategory,
+        required=True,
+        metadata={
+            "description": "The opportunity category",
+        },
+    )
+    category_explanation = fields.String(
+        allow_none=True,
+        load_default=None,
+        validate=validators.Length(max=255),
+        metadata={
+            "description": "Explanation of the category (required when category is 'other')",
+        },
+    )
+
+
+class OpportunityUpdateResponseSchema(AbstractResponseSchema):
+    """Schema for PUT /v1/grantors/opportunities/<opportunity_id> response"""
+
+    data = fields.Nested(OpportunityGrantorSchema())
 
 
 class OpportunityCreateResponseSchema(AbstractResponseSchema):
--- a/api/src/api/opportunities_v1/opportunity_routes.py
+++ b/api/src/api/opportunities_v1/opportunity_routes.py
@@ -214,7 +214,7 @@ def opportunity_search(
     if search_params.get("format") == opportunity_schemas.SearchResponseFormat.CSV:
         # Convert the response into a CSV and return the contents
         output = io.StringIO()
-        opportunities_to_csv(opportunities, output, False)
+        opportunities_to_csv(opportunities, output)
         timestamp = datetime_util.utcnow().strftime("%Y%m%d-%H%M%S")
         return Response(
             output.getvalue().encode("utf-8"),
--- a/api/src/services/opportunities_grantor_v1/opportunity_update.py
+++ b/api/src/services/opportunities_grantor_v1/opportunity_update.py
@@ -1,0 +1,34 @@
+import logging
+import uuid
+
+import src.adapters.db as db
+from src.auth.endpoint_access_util import verify_access
+from src.constants.lookup_constants import Privilege
+from src.db.models.opportunity_models import Opportunity
+from src.db.models.user_models import User
+from src.services.opportunities_grantor_v1.get_opportunity import get_opportunity_for_grantors
+from src.services.opportunities_grantor_v1.opportunity_utils import validate_opportunity_is_draft
+
+logger = logging.getLogger(__name__)
+
+
+def update_opportunity(
+    db_session: db.Session, user: User, opportunity_id: uuid.UUID, opportunity_data: dict
+) -> Opportunity:
+    opportunity = get_opportunity_for_grantors(db_session, user, opportunity_id)
+
+    # Check if user has permission to update opportunities for this agency
+    verify_access(user, {Privilege.UPDATE_OPPORTUNITY}, opportunity.agency_record)
+
+    validate_opportunity_is_draft(opportunity)
+
+    # PUT endpoint — always update all fields
+    for field, value in opportunity_data.items():
+        setattr(opportunity, field, value)
+
+    logger.info(
+        "Updated opportunity",
+        extra={"opportunity_id": opportunity_id},
+    )
+
+    return opportunity
--- a/api/src/services/opportunities_grantor_v1/opportunity_utils.py
+++ b/api/src/services/opportunities_grantor_v1/opportunity_utils.py
@@ -1,0 +1,12 @@
+from src.api.route_utils import raise_flask_error
+from src.db.models.opportunity_models import Opportunity
+
+
+def validate_opportunity_is_draft(opportunity: Opportunity) -> None:
+    """Raise a 422 error if the opportunity is not a draft.
+
+    Only draft opportunities can be modified. This check is shared across
+    all update endpoints to ensure consistent enforcement.
+    """
+    if not opportunity.is_draft:
+        raise_flask_error(422, message="Only draft opportunities can be updated")
--- a/api/src/services/opportunities_v1/opportunity_to_csv.py
+++ b/api/src/services/opportunities_v1/opportunity_to_csv.py
@@ -68,17 +68,11 @@ def _build_opportunity_url(opportunity_id: str, base_url: str) -> str:
 def opportunities_to_csv(
     opportunities: Sequence[dict],
     output: io.StringIO,
-    has_full_fields: bool,
 ) -> None:
-    output_fields = (
-        CSV_FIELDS
-        if has_full_fields
-        else [field for field in CSV_FIELDS if field != "summary_description"]
-    )
-    writer = csv.DictWriter(output, fieldnames=output_fields, quoting=csv.QUOTE_ALL)
+    writer = csv.DictWriter(output, fieldnames=CSV_FIELDS, quoting=csv.QUOTE_ALL)
     writer.writeheader()
 
-    csv_fields_set = set(output_fields)
+    csv_fields_set = set(CSV_FIELDS)
     base_url = os.getenv("FRONTEND_BASE_URL")
     for opportunity in opportunities:
         opp = flatten_dict(opportunity)
--- a/api/src/task/opportunities/export_opportunity_data_task.py
+++ b/api/src/task/opportunities/export_opportunity_data_task.py
@@ -148,7 +148,7 @@ class ExportOpportunityDataTask(Task):
         logger.info("Creating Opportunity CSV extract", extra={"csv_extract_path": self.csv_file})
 
         with file_util.open_stream(self.csv_file, "w") as outfile:
-            opportunities_to_csv(opportunities, outfile, True)
+            opportunities_to_csv(opportunities, outfile)
 
         csv_size = file_util.get_file_length_bytes(self.csv_file)
         return csv_size
--- a/api/tests/src/api/opportunities_grantors_v1/test_opportunity_route_update.py
+++ b/api/tests/src/api/opportunities_grantors_v1/test_opportunity_route_update.py
@@ -1,0 +1,147 @@
+import uuid
+
+import pytest
+
+from src.constants.lookup_constants import Privilege
+from tests.lib.agency_test_utils import create_user_in_agency_with_jwt_and_api_key
+from tests.src.db.models.factories import OpportunityFactory
+
+
+@pytest.fixture
+def grantor_auth_data(db_session, enable_factory_create):
+    """Create a user with UPDATE_OPPORTUNITY permission and return auth data"""
+    user, agency, token, api_key_id = create_user_in_agency_with_jwt_and_api_key(
+        db_session=db_session,
+        privileges=[Privilege.VIEW_OPPORTUNITY, Privilege.UPDATE_OPPORTUNITY],
+    )
+    return user, agency, token, api_key_id
+
+
+@pytest.fixture
+def existing_opportunity(grantor_auth_data, enable_factory_create):
+    """Create an opportunity belonging to the grantor's agency"""
+    _, agency, _, _ = grantor_auth_data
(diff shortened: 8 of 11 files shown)


api/src/services/organizations_v1/get_organization.py
```diff
@@ -57,11 +58,11 @@ def get_organization_and_verify_access(
     # First get the organization
     organization = get_organization(db_session, organization_id)
 
-    # Check if user has VIEW_ORG_MEMBERSHIP privilege for this organization
+    # Check if user has the correct privilege for this organization
     check_user_access(
         db_session,
         user,
-        {Privilege.VIEW_ORG_MEMBERSHIP},
+        privilege,
```
chouinar: I don't think you want to change this. To view an organization (what this function does) we want that specific privilege.

You've changed what privilege gets passed into this function, several endpoints not related to this work now have different privileges to access them than they did before because of this.

This function verifies they have exactly this privilege, if the endpoint in question wants to check something else, either don't use this function or add additional checks after calling this function.

api/src/api/organizations_v1/organization_routes.py
```diff
@@ -72,7 +73,7 @@ def organization_get(db_session: db.Session, organization_id: UUID) -> response.
 
         # Get organization and verify access using service layer
         organization = get_organization_and_verify_access(
-            db_session, user_token_session.user, organization_id
+            db_session, user_token_session.user, organization_id, {Privilege.VIEW_ORG_MEMBERSHIP}
```
chouinar: Why would the GET organization endpoint require this privilege? That's for getting members of an organization, not the organization itself.

api/src/services/users/get_saved_opportunities.py
```diff
@@ -67,23 +75,86 @@ def add_opportunity_status_filter(
...
+    db_session: db.Session, user: User, raw_opportunity_params: dict
 ) -> tuple[Sequence[Opportunity], PaginationInfo]:
-    logger.info(f"Getting saved opportunities for user {user_id}")
-
     opportunity_params = SavedOpportunityListParams.model_validate(raw_opportunity_params)
-
+    org_ids_param = opportunity_params.organization_ids
+    user_org_ids = [ou.organization_id for ou in user.organization_users]
+    user_id = user.user_id
+    logger.info(f"Getting saved opportunities for user {user_id}")
```
chouinar: Avoid putting variable data in log messages, makes it harder to find them. Always put any IDs in the extra - we can skip it here as the `auth.user_id` field is automatically attached by our auth.

```suggestion
    logger.info("Getting saved opportunities for user")
```

api/src/services/users/get_saved_opportunities.py
```diff
@@ -67,23 +75,86 @@ def add_opportunity_status_filter(
...
 
 
+def _check_access(db_session: db.Session, user: User, organization_ids: list) -> None:
+    for org_id in organization_ids:
+        get_organization_and_verify_access(
+            db_session, user, org_id, {Privilege.VIEW_ORG_SAVED_OPPORTUNITIES}
+        )
+
+
+def _build_saved_union_subquery(
```
chouinar, line 112: Could you add some example SQL that this generates?

api/src/services/organizations_v1/get_organization.py
babebe: Updated to check if specific privilege is required otherwise checks for basic privilege

api/src/services/users/get_saved_opportunities.py
babebe:
```
SELECT anon_1.opportunity_id, max(anon_1.saved_at) AS saved_at 
FROM (
SELECT api.organization_saved_opportunity.opportunity_id AS opportunity_id, api.organization_saved_opportunity.created_at AS saved_at 
FROM api.organization_saved_opportunity 
WHERE api.organization_saved_opportunity.organization_id IN (__[POSTCOMPILE_organization_id_1]) 
UNION ALL
SELECT api.user_saved_opportunity.opportunity_id AS opportunity_id, api.user_saved_opportunity.created_at AS saved_at
FROM api.user_saved_opportunity
WHERE api.user_saved_opportunity.user_id = :user_id_1 AND api.user_saved_opportunity.is_deleted IS NOT true) AS anon_1 
GROUP BY anon_1.opportunity_id
```

api/src/api/organizations_v1/organization_routes.py
babebe: Updated with new privilege

api/src/services/users/get_saved_opportunities.py
```diff
@@ -67,23 +76,102 @@ def add_opportunity_status_filter(
...
+        )
+        org_ids_to_use = [org.organization_id for org in relevant_orgs]
+    elif org_ids_param:
+        # User specified orgs, verify access
+        _check_access(db_session, user, org_ids_param)
+        org_ids_to_use = org_ids_param
+        include_user_saved_opps = False
+    else:
+        # Explicitly empty list, consider only user saved opps
+        org_ids_to_use = []
```
freddieyebra, line 196: This deviates from the [requirements](https://github.com/HHS/simpler-grants-gov/issues/8433). Additionally, we should not be setting any values on the search filter that are not explicitly passed in from the caller. 

Use cases:
1. User saved opportunities only (current functionality)
organization_ids=None
log message = "user saved opportunities requested"

2. Organizations + user saved opportunities
organization_ids=[]
log message = "all saved opportunities requested"

3. Organization saved opportunities only
organization_ids=["123", "456"]
log message = "organization saved opportunities requested"

api/src/services/organizations_v1/get_organization.py
babebe: Spoke with @joshtonava . Added new "view_organization" privilege to be applied here. Added this basic privilege to the relevant static roles.

api/src/services/users/get_saved_opportunities.py
freddieyebra: This is not a blocker bc i do understand there has been a lot progress on this pr already but i do wonder if this can have less impact on other endpoints/services by checking the user org access higher up in this method and avoid changes in privileges in other endpoints. Let me know if my thoughts are way out of line here and we can chat more about it. cc @chouinar for thoughts as well.

I envisioned logic similar to:

```suggestion
    if org_ids_param is None:
        # execute and return existing results
    user_org_ids = get all organization ids the user is a member or admin
    if org_ids_param == []:
        # use all user_org_ids to search and include opportunities for organization_saved_opportunities
        return get_saved_opportunities(user_id, user_org_ids)
    else:
        # assert that all of the passed in organization ids exist in the user_org_ids otherwise error
        return get_saved_opportunities(user_id, org_ids_param)
```

chouinar: @freddieyebra - I see that one of the requirements says it should be a new endpoint, should the existing saved opportunities endpoint be modified for this?

api/src/constants/lookup_constants.py
```diff
@@ -227,6 +227,7 @@ class ApplicationFormStatus(StrEnum):
 class Privilege(StrEnum):
     MANAGE_ORG_MEMBERS = "manage_org_members"
     MANAGE_ORG_ADMIN_MEMBERS = "manage_org_admin_members"
+    VIEW_ORGANIZATION = "view_organization"
```
chouinar: I don't think we want a new privilege with this, `view_org_membership` is the view organization privilege, it's just not consistently named. We should either rename it (needs to verify frontend doesn't have any gates for this) or leave it alone.

api/src/services/organizations_v1/get_organization.py
```diff
@@ -53,15 +54,16 @@ def get_organization_and_verify_access(
...
+    # Retrieve organization (raises 404 if not found)
     organization = get_organization(db_session, organization_id)
 
-    # Check if user has VIEW_ORG_MEMBERSHIP privilege for this organization
+    # Enforce privilege check
     check_user_access(
         db_session,
         user,
-        {Privilege.VIEW_ORG_MEMBERSHIP},
+        {Privilege.VIEW_ORGANIZATION},
```
chouinar: Don't change this - this 100% needs to be left alone, changing this changes the functionality of every endpoint that uses this utility. Don't do that.

api/src/services/organizations_v1/list_legacy_users.py
```diff
@@ -235,8 +234,16 @@ def list_legacy_users_and_verify_access(
         404: Organization not found
         400: Organization does not have a UEI
     """
-    # Get organization and verify access
-    organization = get_organization_and_verify_access(db_session, user, organization_id)
+    # Retrieve organization (raises 404 if not found)
+    organization = get_organization(db_session, organization_id)
+
+    # Enforce privilege check
+    check_user_access(
```
chouinar: Not strictly this PR, but why do we have `verify_access` and `check_user_access`? We seem to have made a duplicate of this same functionality (the fetching of the privileges should always be done).

api/src/services/users/get_saved_opportunities.py
chouinar: Can you add this as a comment?

api/src/services/users/get_saved_opportunities.py
chouinar: It shouldn't affect privileges of any other endpoint - I chatted with Bruk about that, although I see a few things have still been changed.

freddieyebra: Good call. that is a stale requirement. we decided to support an additional parameter to filter instead.

babebe: Thanks for catching that! I had the behavior reversed initially — `None` should correspond to user-saved only, and an empty list to all accessible orgs plus user-saved. I’ve updated the logic to reflect the correct semantics. And added the logs

babebe: Changes to privilege check in other methods is unrelated. It now fallows the pattern we use else where, i.e each endpoints check for specific privilege it enforces.

api/src/services/organizations_v1/get_organization.py
babebe: With the changes implemented. Only the organization_get route uses this function now, and the other functions that were calling it are doing the same checks they were doing before only now explicitly instead of using this "util" func. But if this is intended to be the basic privilege or could be renamed then reverting the changes makes sense.

babebe:
@chouinar  Reverted all the privilege changes. Will have a followup PR for those.

api/src/services/users/get_saved_opportunities.py
```diff
@@ -67,23 +76,122 @@ def add_opportunity_status_filter(
...
+    user: User,
+    org_users: list,
+) -> list[uuid.UUID]:
+    """
+    Returns only the organization ids the user has required access to.
+    """
+    return [
+        org_user.organization_id
+        for org_user in org_users
+        if can_access(user, {Privilege.VIEW_ORG_SAVED_OPPORTUNITIES}, org_user.organization)
```
joshtonava, line 85: When a user passes specific org IDs, each one triggers a separate get_organization DB call + check_user_access. Not blocking but wanted to call out.

api/src/services/users/get_saved_opportunities.py
```diff
@@ -26,20 +30,25 @@ class SavedOpportunityFilterParams(BaseModel):
...
-            else getattr(UserSavedOpportunity, order.order_by)
-        )
+        if order.order_by == "created_at":
+            # Most recent save event across user + org
+            column = saved_union.c.saved_at
+        else:
+            column = (
+                getattr(model_mapping[order.order_by], order.order_by)
+                if order.order_by in model_mapping
+                else getattr(UserSavedOpportunity, order.order_by)
```
joshtonava: Since the schema only allows created_at, opportunity_title, and close_date, and created_at is handled by the if, only opportunity_title, and close_date can hit the else. Both of them are in model mapping so im thinking the ternary is unnecessary.

If an invalid sort field somehow give through, it would throw a KeyError instead of silently failing back to UserSavedOpportunity which is arguably better in my opinion.

api/src/services/users/get_saved_opportunities.py
babebe: Is that not the intended behavior, to check if they have the right access on the specified organization.

chouinar: If a user were in 20 orgs and passed in 3 IDs, you'd be checking all 20 here.

Instead we could do:

```py
for org_id in org_ids_provided:
     organization = fetch_organization_and_verify_access(...)
     verify_access(user, {Privilege.X}, organization)
```

It verifies that the org IDs exist (404ing if not) and verifies they have access to what is relevant.

You have this check elsewhere, shouldn't need it twice.

api/src/api/users/user_schemas.py
```diff
@@ -156,11 +156,11 @@ class UserSavedOpportunitiesFilterSchema(Schema):
 
 class UserSavedOpportunitiesRequestSchema(Schema):
     filters = fields.Nested(UserSavedOpportunitiesFilterSchema(), required=False, allow_none=True)
-
+    organization_ids = fields.List(fields.UUID(required=False), allow_none=True)
```
chouinar: I see why I was a bit confused by this. Why is organization IDs not in the filter schema?

Keep it the same pattern as every other filter, then the null/empty distinction has some meaning.

Although, @freddieyebra / @joshtonava 

```
User saved opportunities only (current functionality)
organization_ids=None
log message = "user saved opportunities requested"

Organizations + user saved opportunities
organization_ids=[]
log message = "all saved opportunities requested"
```

This behavior is confusing / not how filters behave. If I pass `organization_ids=[]`, that's saying I want only the records which match those organization IDs, if I pass nothing, then I should get zero results (think about if you called search with opportunity status of [].)

In the reverse case, passing no filter, should not filter, it should return everything the endpoint can possibly return. But, that'll mean the endpoint returns more by default than it currently does - which I assume you were trying to avoid? If that's a blocker, we need a new endpoint, that's what would make sense from a filter perspective.

api/src/services/users/get_saved_opportunities.py
babebe: For those explicitly passed organization_ids it it is handled separately in the _check_access function. which does the same thing as here. This is one is used to sort out which of the organizations that the user belongs the user has the correct privilege for.

api/src/services/users/get_saved_opportunities.py
babebe: Oh yes! Given the schema constraints, the fallback isn’t reachable in valid scenarios. I’ll remove the ternary and the `else` fallback to keep it explicit and fail fast.

api/src/api/users/user_schemas.py
joshtonava: I agree `organization_ids` should be part of the filters.

We had some discussion on this in the comments of the ticket, 
https://github.com/HHS/simpler-grants-gov/issues/8433#issuecomment-3923209323
I updated the description to be what we settled on.

Important note: user saved opportunities and organization saved opportunities are separate entities in the database. The endpoint returns a union of both, and organization_ids controls how that union is filtered:

organization_ids | Result
-- | --
omitted / null | User saved + ALL accessible org saved
[] (empty list) | Filter out all organization saved opportunities (User saved only)
["123", "456"] | Match only those specific orgs' saved (no user saved)

The `[]` applies the empty filter to the org saved portion of the results (returning zero org saved matches, which is standard behavior), while user saved opportunities are a separate entity that the filter doesnt apply to. When specific org IDs are provided, the intent is to filter org saved results so user saved opportunities are excluded.

> In the reverse case, passing no filter, should not filter, it should return everything the endpoint can possibly return. But, that'll mean the endpoint returns more by default than it currently does - which I assume you were trying to avoid? If that's a blocker, we need a new endpoint, that's what would make sense from a filter perspective.

Regarding this, you are right we are going to be returning more than the API currently does. Its not clear to me that this would break the frontend currently but the follow up ticket does update the response schema so that is worth factoring in. 
Making a new version of this endpoint was discussed in the ticket as well.

cc: @babebe @freddieyebra

api/src/services/users/get_saved_opportunities.py
babebe: Updated logic to get all the organizations requested in one call. If anyone organization is missing, we through a "404". And run the verify access check afterwards.

api/src/api/users/user_schemas.py
babebe: updated accordingly

api/src/services/organizations_v1/list_legacy_users.py
babebe: the `check-user-access` func loads the role for better performance.

api/src/services/users/get_saved_opportunities.py
```diff
@@ -67,23 +72,130 @@ def add_opportunity_status_filter(
...
+    organizations = (
+        db_session.query(Organization)
+        .filter(Organization.organization_id.in_(organization_ids))
+        .all()
+    )
+    org_map = {org.organization_id: org for org in organizations}
+
+    missing_ids = set(organization_ids) - org_map.keys()
+    if missing_ids:
+        raise raise_flask_error(404, f"Organization(s) not found: {missing_ids}")
```
joshtonava: issue: I dont think you need to explicitly do raise raise_flask_error


```suggestion
       raise_flask_error(404, f"Organization(s) not found: {missing_ids}")
```

api/tests/src/api/users/test_user_saved_opportunities_get.py
```diff
@@ -153,16 +159,43 @@ def test_get_saved_opportunities_unauthorized_user(client, enable_factory_create
...
+    """Test that when organization_ids is empty, only user saved opportunities should be returned"""
+
+    # User save
+    user_saved_opp = UserSavedOpportunityFactory.create(user=user)
+
+    response = client.post(
+        f"/v1/users/{user.user_id}/saved-opportunities/list",
+        headers={"X-SGG-Token": user_auth_token},
+        json={
+            "organization_ids": [],
```
joshtonava: **issue:**
If we added organization_ids to the filters, I think this test more accurately would be 

```suggestion
            "filters": {"organization_ids": []},
```

api/src/api/users/user_schemas.py
```diff
@@ -152,15 +152,15 @@ class UserSavedOpportunitiesFilterSchema(Schema):
         .with_one_of(allowed_values=OpportunityStatus, example=OpportunityStatus.POSTED)
         .build()
     )
+    organization_ids = fields.List(fields.UUID(required=False), allow_none=True)
```
chouinar: This should be the same pattern as our other filters. We've got a `UuidSearchSchemaBuilder` for this.

api/src/api/users/user_schemas.py
chouinar: @joshtonava - I'm still struggling to reconcile what I would expect as a user. What would make intuitive sense to me. I think the issue is that changing whether the users personal saved opportunities get returned based on a filter on organizations seems really unintuitive.

I know this isn't backwards compatible, but I really think this should be something like:
* `types_to_include` -> a list of organization / user (needs a better name).
* Org filters (only used if they asked for org).

Even if the frontend doesn't care, we have to consider how the endpoint makes sense to users using it directly.

joshtonava: I'm not sold on this as being objectively better better for a user either.

Some cases that lead to ambiguity:
- Does types_to_include always need to be provided?
- types_to_include: Null, undefined && organization_ids provided?
- types_to_include: 'user' && organization_ids provided?

I don't want to block this ticket for future API consumers since there are frontend dependencies on this work currently. 
Are you okay if we come back to this with a v2 breaking change that could be a different interface for API consumers? (stub: https://github.com/HHS/simpler-grants-gov/issues/8761)

cc: @chouinar

chouinar: I'd say the types to include should always be provided, not empty, not null. We could compromise on a default of `["user"]`.

@joshtonava - We've gone in circles a bit on this discussing it async. Could we chat through this?

PR #8638, opened 2026-02-24
[Issue #8629] Adjust AuthZ enforcement for workflow service

### LINKED ISSUE #8629: Adjust AuthZ approach for approvals in workflow service
### Summary

After discussion, we want to keep AuthZ purely in the event API portion of the workflow service - so we want to remove AuthZ from where we added it in the workflow approval logic - however we still want an easy way to test the privileges when we setup a workflow. To that end we want to do the following:

* Change the workflow config to have the approval privileges configured off of events instead of states
* Add a utility for checking if a user can do event X in workflow Y.
  * This requires quite a few checks as we need to iterate over agencies (albeit in all current cases, will be 1)
  * Use this util for testing that configs are correct + util gets plugged into the event API itself
* Remove direct usage of AuthZ from the workflow service layer
  * This is largely changing tests

This isn't a massive refactor, although the tests will need to be adjusted quite a bit.

### Acceptance criteria

- [ ] AuthZ no longer enforced in service layer
- [ ] Workflow config uses events rather than states for mapping
- [ ] Utility built and used in tests for checking workflow approval

chouinar:
## Summary
Fixes #8629 

## Changes proposed
* Made it so the approval configuration is now event->ApprovalConfig instead of state->ApprovalConfig (Will make AuthZ in endpoint easier)
* Removed AuthZ enforcement from approval in workflow service
* Made a utility for checking AuthZ for a given user + workflow (to be used in the event API)

## Context for reviewers
This is a follow-up to a few discussions and is a first step in simplifying a bit of the AuthZ. The general purpose is to make it so all AuthZ is enforced in our API, and none is done in the workflow service. The utility created and tested in this is what we'll use in the API to do that enforcement.

## Validation steps
Still just tests at the moment

### DIFF
--- a/api/src/workflow/service/approval_service.py
+++ b/api/src/workflow/service/approval_service.py
@@ -3,12 +3,18 @@ import logging
 from sqlalchemy import select
 
 from src.adapters import db
+from src.auth.endpoint_access_util import can_access
 from src.constants.lookup_constants import ApprovalResponseType, ApprovalType
+from src.db.models.agency_models import Agency
 from src.db.models.user_models import User
 from src.db.models.workflow_models import Workflow, WorkflowApproval
 from src.workflow.event.state_machine_event import StateMachineEvent
+from src.workflow.workflow_config import WorkflowConfig
 from src.workflow.workflow_constants import WorkflowConstants
-from src.workflow.workflow_errors import InvalidWorkflowResponseTypeError
+from src.workflow.workflow_errors import (
+    InvalidWorkflowResponseTypeError,
+    OpportunityWithoutAgencyError,
+)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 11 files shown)


api/src/workflow/service/approval_service.py
```diff
@@ -60,3 +66,58 @@ def get_approval_response_type(state_machine_event: StateMachineEvent) -> Approv
...
+            "agency_id": agency.agency_id,
+            "agency_code": agency.agency_code,
+        }
+        logger.info("Checking if user can access agency for approvals", extra=agency_log_extra)
+        if not can_access(user, required_privileges, agency):
+            logger.info("User cannot access agency for approvals", extra=agency_log_extra)
+            return False
+
+    logger.info("User can access agencies for approvals", extra=log_extra)
+    return True
```
kkrug, line 123: Just double checking, it is functionally impossible to end up here with a workflow with zero opportunities and zero applications, right? You would fall through to 'True' here otherwise without a check.

api/src/workflow/service/approval_service.py
```diff
@@ -60,3 +66,58 @@ def get_approval_response_type(state_machine_event: StateMachineEvent) -> Approv
...
+        agencies_map[opportunity.agency_record.agency_id] = opportunity.agency_record
+
+    for application in workflow.applications:
+        opportunity = application.competition.opportunity
+        if opportunity.agency_record is None:
+            raise OpportunityWithoutAgencyError("Opportunity does not have an agency record")
+
+        agencies_map[opportunity.agency_record.agency_id] = opportunity.agency_record
+
+    return list(agencies_map.values())
```
jakobpederson, line 87: Could `agencies_map` just be a list? It seems like we aren't using the key for anything other than maybe overriding agency records?

api/src/workflow/service/approval_service.py
chouinar: Depends on some order of operations, we have checks in other places that validate your workflow has the expected entity types.

For example, the initial prototype workflow has a few different ways that it verifies it is a workflow with 1 opportunity (0 or 2+ would error when trying to run anything). The workflow config also makes certain that only entities of a particular type are present. So, you'd get stopped somewhere.

There are a lot of complexities around the lists of entities, I started a thread in the engineering channel to see if we can simplify it a bit. This is confusing, this is way too complicated, we should make the entities simpler.

api/src/workflow/service/approval_service.py
chouinar: The map is for uniqueness. If the same agency is found twice, it gets added to the map the same both times, then we just grab the values.

PR #8645, opened 2026-02-24
[Unticketed] Don't log an error if someone calls callback directly

chouinar:
## Changes proposed
* Don't raise a 500 / error when someone calls our login callback without required parameters

## Context for reviewers
We received an error earlier today for someone calling our callback endpoint for the login flow without any parameters. This shouldn't happen for any normal operation, and any normal operation should handle this with a proper redirect. This makes it so the error won't get logged as an error by adjusting the validation to be more explicit.

Note that unlike our normal endpoints, we can't do this as part of request validation because that wouldn't allow us to redirect the response.

## Validation steps
You can see this issue right now by going to `http://localhost:8080/v1/users/login/callback` - without this change it'll cause the API to fail and log an error, you won't be redirected. With the fix, you'll redirect through to the frontend and end up at `http://localhost:3000/unauthenticated`.

### DIFF
--- a/api/src/services/users/login_gov_callback_handler.py
+++ b/api/src/services/users/login_gov_callback_handler.py
@@ -22,7 +22,7 @@ logger = logging.getLogger(__name__)
 
 class CallbackParams(BaseModel):
     code: str | None = None
-    state: str
+    state: str | None = None
     error: str | None = None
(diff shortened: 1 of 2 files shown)


mdragon (commented):
👍🏻

Issue #7522, opened 2025-12-17
WF205: Infra set up in lower environment


ychoy:
### Summary

Deploy workflow service infrastructure in development environment

Dependencies:
#7172 
#7519 
#7518 

Blocked by: #8067 

We will have created a script in #8067 that we want to run. This will eventually get more logic added to it.

We want the following:
* ECS service created that runs script (details will be given about entrypoint command once written)
* Ideally same general setup as our API service, that is, same env vars, and a user that can interact with the DB (app user) and search index (whatever we have for reading). We'll need SQS as well, but can handle that when setting that up.
* Any basic healthcheck/etc needed - note this won't be an API service, just a main loop running

### Acceptance criteria

- [x] Infrastructure deployed using infra as code
- [x] Development environment operational
- [x] POC workflow service running in dev environment
- [x] Service accessible to development team
- [x] Basic connectivity and deployment validate

prasnava:
~~Moving this ticket to blocked; we are still waiting on the wild card ssl certs to provision.~~
We can setup the infra for the workflow service; this is not blocked.

PR #8648, opened 2026-02-25
[Issue #7522]: Added new workflow service

prasnava:
## Summary

Added the new workflow service.

## Validation steps

Deployed to the dev environment: [here](https://github.com/HHS/simpler-grants-gov/actions/runs/22396083336)

Link to cloud watch logs: [here](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/service$252Fapi-dev-workflow-fluentbit/log-events/api-dev-workflow$252Fapi-dev-workflow-fluentbit$252Fe6cea1ab1cfc4904891d5a4b9b7504cc)

Seeing the workflow logs executing below: 
Screenshot 2026-02-25 at 9 05 10 AM


prasnava:
In cloudwatch log insights we can filter for the workflow logs using the query below: 

```
fields @timestamp, @message
  | parse @message '"msg"=>"*"' as msg
  | parse @message '"event_data"=>"*"' as event_data
  | parse @message '"event_type"=>"*"' as event_type
  | filter ispresent(msg)
  | display @timestamp, msg
  | sort @timestamp desc
  | limit 100
```

The log group name is: service/api-dev-workflow-fluentbit

infra/api/service/workflow.tf
```diff
@@ -0,0 +1,171 @@
...
+
+      environment = concat(module.service.environment_variables, [
+        {
+          name  = "TMPDIR"
+          value = "/tmp"
+        },
+        {
+          name  = "ENABLE_WORKFLOW_API"
+          value = "true"
+        }
```
chouinar: Shouldn't need to add this, this adds the API on our existing service, we don't have any environment variable gating this new service.

infra/api/service/main.tf
```diff
@@ -60,6 +60,22 @@ locals {
...
+
+  container_secrets = concat(
+    [for secret_name in keys(local.service_config.secrets) : {
+      name      = secret_name
+      valueFrom = module.secrets[secret_name].secret_arn
+    }],
+    module.app_config.enable_identity_provider ? [{
+      # name      = "COGNITO_CLIENT_SECRET"
+      # valueFrom = module.identity_provider_client[0].client_secret_arn
+    }] : [],
```
chouinar, line 71: Should this be commented out? Don't think we'd need an identity provider since we won't be hitting cognito, so can probably trim this out?

infra/api/service/workflow.tf
```diff
@@ -0,0 +1,171 @@
...
+  execution_role_arn = module.service.task_role_arn
+  task_role_arn      = module.service.app_service_arn
+
+  container_definitions = jsonencode([
+    {
+      name                   = local.workflow_service_name,
+      image                  = module.service.image_url,
+      memory                 = local.workflow_service_config.memory - local.workflow_fluent_bit_memory,
+      cpu                    = local.workflow_service_config.cpu - local.workflow_fluent_bit_cpu,
+      networkMode            = "awsvpc",
```
chouinar, line 52: I don't see anything that would make me think it is, but just double-checking that this service isn't available in any way to be called (since it's not an API, just a main loop).

infra/api/service/workflow.tf
```diff
@@ -0,0 +1,171 @@
...
+      essential              = true,
+      readonlyRootFilesystem = true,
+
+      command = ["poetry", "run", "flask", "workflow", "workflow-main"],
+
+      healthCheck = null,
```
chouinar, line 58: Should we have a healthcheck? We don't have endpoints when running in this way, not sure if there's some container check we could do?

infra/api/service/workflow.tf
```diff
@@ -0,0 +1,171 @@
...
+      memory                 = local.workflow_fluent_bit_memory,
+      cpu                    = local.workflow_fluent_bit_cpu,
+      networkMode            = "awsvpc",
+      essential              = true,
+      readonlyRootFilesystem = false,
+      healthCheck = {
+        timeout     = 5,
+        interval    = 10,
+        startPeriod = 30,
+        command     = ["CMD-SHELL", "curl http://localhost:80/api/v1/health"]
```
chouinar: I don't think this would return anything since we won't be running the API service, unless this is just formatted to look very similar to our API's healthcheck endpoint but is something different

infra/api/service/workflow.tf
prasnava: We should add a health check endpoint, or use one if we have one. Right now the container remains healthy as long as the process doesn't crash.

infra/api/service/workflow.tf
prasnava: Correct, we can't invoke it directly. It just runs as a standalone ecs service.

infra/api/service/main.tf
prasnava: We are not using it, but it's just the template module we are using in the rest of the repo. Will remove it.

prasnava: Updated.

Issue #7119, opened 2025-11-18
properly resolve js-yaml vulnerability


doug-s-nava:
### Summary

we patched the js-yaml subdependency here in order to resolve a vulnerability in version 4.1.0
https://github.com/HHS/simpler-grants-gov/pull/7117

long term we don't want this in our top level dependencies. Once all dependencies that include js-yaml as a subdependency have been upgraded we should remove js-yaml from our top level dependencies

dependencies to watch for updates:

* postcss-loader/cosmiconfig
* json-schema-ref-parser
* eslintrc
* @istanbuljs/load-nyc-config

### Acceptance criteria

- [ ] js-yaml is removed from top level dependencies
- [ ] frontend passes vuln scans

PR #8652, opened 2026-02-25
[Issue #7119] remove js-yaml from explicit deps

doug-s-nava:
## Summary

Fixes for #7119 

## Changes proposed

Removes redundant inclusion of js-yaml in explicit npm dependencies

## Context for reviewers

Including js-yaml 4.1.1 explicitly was done to work around a vulnerability in 4.1.0 while subdependencies worked on updating to the patched version. It seems like all those subdependencies are now patched, and there are no longer vulnerabilities appearing in the scan when the explicit inclusion of 4.1.1 is removed from our package.json

## Validation steps

1. _VERIFY_: vulnerability scans pass

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -25,7 +25,6 @@
         "isomorphic-dompurify": "^2.15.0",
         "jose": "^5.9.6",
         "js-cookie": "^3.0.5",
-        "js-yaml": "4.1.1",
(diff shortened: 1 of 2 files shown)


nathan-stricker (approved):
LGTM! All vulnerability checks passing but just noting that flaky E2E test will need to be rerun.

Issue #8102, opened 2026-01-23
Store event history in workflow audit history table


chouinar:
### Summary
We want to populate the `workflow_audit` table as we run the state machine. This can be done using a [listener](https://python-statemachine.readthedocs.io/en/latest/listeners.html).

The proof of concept used this like so (note that the typing / exact params will likely differ):

```py
class AuditListener:

    def __init__(self):
        self.user_id = None

    def on_transition(self, event_data: EventData, model: OpportunityPersistentModel):
        # TODO - there is probably a better way to do this
        event_metadata = event_data.extended_kwargs.get("event_metadata")
        if event_metadata is not None:
            user_id = event_metadata.acting_user_id
        else:
            user_id = None

        model.db_session.add(WorkflowAudit(workflow=model.workflow, source_state=event_data.source.value, target_state=event_data.target.value, transition_event=event_data.event.name, user_id=user_id))
```

When creating the state machine, we did:
```py
ExampleStateMachine(model, listeners=[AuditListener()])
```
which makes it so these audit events are added as part of the logic.

It's possible this might be doable with our persistence classes as well, we'll have to test out what works.

### Acceptance criteria

- [ ] Audit history populated automatically for all current and future workflows
- [ ] Tests added that verify events are added as expected

PR #8656, opened 2026-02-25
[Issue #8102] Store event history in workflow audit history table

kkrug:
## Summary

Fixes for #8102  

## Changes proposed

- Created workflow_audit_listener.py, which is a listener for state machine transitions that automatically creates
    audit records in the workflow_audit table
- Updated event_handler.py to initialize audit listener
- Created test_workflow_audit_listener.py for audit listener unit tests
- Created two necessary __init__.py files

## Context for reviewers

In order to populate the workflow_audit table as we run the state machine, we implemented a [listener](https://python-statemachine.readthedocs.io/en/latest/listeners.html).

## Validation steps

Confirm tests pass.

### DIFF
--- a/api/src/workflow/listener/workflow_audit_listener.py
+++ b/api/src/workflow/listener/workflow_audit_listener.py
@@ -25,46 +25,22 @@ class WorkflowAuditListener:
         """
         self.db_session = db_session
 
-    def on_transition(self, event_data: EventData) -> None:
+    def on_transition(self, state_machine_event: StateMachineEvent, event_data: EventData) -> None:
         """
         Called automatically by the state machine when a transition occurs.
         """
-        # Extract the state_machine_event from the extended_kwargs
-        state_machine_event: StateMachineEvent | None = event_data.extended_kwargs.get(
-            "state_machine_event"
-        )
-
-        if state_machine_event is None:
-            logger.warning(
-                "State machine event not found in transition data, skipping audit record creation",
-                extra={
-                    "source_state": event_data.source.value if event_data.source else None,
-                    "target_state": event_data.target.value if event_data.target else None,
-                    "event_name": event_data.event.name if event_data.event else None,
-                },
-            )
-            return
 
         # Create the audit record
         workflow_audit = WorkflowAudit(
-            workflow_id=state_machine_event.workflow.workflow_id,
-            acting_user_id=state_machine_event.acting_user.user_id,
+            workflow=state_machine_event.workflow,
+            acting_user=state_machine_event.acting_user,
             transition_event=event_data.event.name,
             source_state=event_data.source.value,
             target_state=event_data.target.value,
-            event_id=state_machine_event.workflow_history_event.event_id,
+            event=state_machine_event.workflow_history_event,
             audit_metadata=state_machine_event.metadata,
         )
 
         self.db_session.add(workflow_audit)
 
-        logger.info(
-            "Created workflow audit record for transition",
-            extra={
-                "workflow_id": state_machine_event.workflow.workflow_id,
-                "source_state": event_data.source.value,
-                "target_state": event_data.target.value,
-                "transition_event": event_data.event.name,
-                "acting_user_id": state_machine_event.acting_user.user_id,
-            },
-        )
+        logger.info("Created workflow audit record for transition")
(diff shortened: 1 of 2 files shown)


api/src/workflow/listener/workflow_audit_listener.py
```diff
@@ -0,0 +1,70 @@
...
+    def on_transition(self, event_data: EventData) -> None:
+        """
+        Called automatically by the state machine when a transition occurs.
+        """
+        # Extract the state_machine_event from the extended_kwargs
+        state_machine_event: StateMachineEvent | None = event_data.extended_kwargs.get(
+            "state_machine_event"
+        )
+
+        if state_machine_event is None:
```
chouinar: If you do:

```suggestion
    def on_transition(self, state_machine_event: StateMachineEvent, event_data: EventData) -> None:
        """
        Called automatically by the state machine when a transition occurs.
        """
        ...
```

Does that work? That's how I've setup all the other `on` functions, it passes any parameters you pass into the send function as kwargs - and we do `send(..., state_machine_event=state_machine_event)`.

api/src/workflow/listener/workflow_audit_listener.py
```diff
@@ -0,0 +1,70 @@
...
+                    "target_state": event_data.target.value if event_data.target else None,
+                    "event_name": event_data.event.name if event_data.event else None,
+                },
+            )
+            return
+
+        # Create the audit record
+        workflow_audit = WorkflowAudit(
+            workflow_id=state_machine_event.workflow.workflow_id,
+            acting_user_id=state_machine_event.acting_user.user_id,
```
chouinar: Prefer setting relationships over IDs. That way if you try to do `workflow_audit.workflow` after this, it's set already without a need for a flush. Broadly, work with relationships, let the DB worry about IDs.

```suggestion
            workflow=state_machine_event.workflow,
            acting_user=state_machine_event.acting_user,
```

api/src/workflow/listener/workflow_audit_listener.py
```diff
@@ -0,0 +1,70 @@
...
+
+        logger.info(
+            "Created workflow audit record for transition",
+            extra={
+                "workflow_id": state_machine_event.workflow.workflow_id,
+                "source_state": event_data.source.value,
+                "target_state": event_data.target.value,
+                "transition_event": event_data.event.name,
+                "acting_user_id": state_machine_event.acting_user.user_id,
+            },
```
chouinar: On a few of the classes I added a `get_log_extra` function to help simplify the logging a bit. Just returns a bunch of the IDs/values that are good to log.

```suggestion
            extra=state_machine_event.get_log_extra() | {"source_state": event_data.source.value, "target_state": event_data.target.value, "transition_event": event_data.event.name},
```

api/src/workflow/listener/workflow_audit_listener.py
```diff
@@ -0,0 +1,70 @@
...
+            return
+
+        # Create the audit record
+        workflow_audit = WorkflowAudit(
+            workflow_id=state_machine_event.workflow.workflow_id,
+            acting_user_id=state_machine_event.acting_user.user_id,
+            transition_event=event_data.event.name,
+            source_state=event_data.source.value,
+            target_state=event_data.target.value,
+            event_id=state_machine_event.workflow_history_event.event_id,
```
chouinar: Same here - Although I think for this one we also need to add the relationship so we can set it.

```suggestion
            event=state_machine_event.workflow_history_event,
```

api/src/workflow/listener/workflow_audit_listener.py
kkrug: Yeah this works - updated

api/src/workflow/listener/workflow_audit_listener.py
kkrug: I made these changes, but the workflow models needed a change as well to remove `single_parent=True`

api/src/workflow/listener/workflow_audit_listener.py
kkrug: Updated

api/src/workflow/listener/workflow_audit_listener.py
kkrug: Updated, and the workflow models were updated to add this relationship

api/src/workflow/listener/workflow_audit_listener.py
chouinar: Curious, what issue was `single_parent` causing?

kkrug: When I was using the ids previously, no relationship is established so the single_parent constraint is not checked. Now that it is using the relationship, it checks the constraint. When I set workflow_id=, SQLAlchemy doesn't know or care that I linked multiple audits to the same workflow, it's just storing numbers in a database column.

PR #8659, opened 2026-02-25
[Issue #8658] Change workflow to have a single entity

### LINKED ISSUE #8658: Simplify workflow entity model to be a single entity per workflow
### Summary

Right now we support multiple entities per workflow, that is a given workflow could be associated with 3 opportunities, 2 applications, and a submission, all at the same time. This adds a ton of complexity including for auth (needing to validate a user can access everything). We don't have a known use case for needing this support, we did it because it seemed flexible, but it is adding a ton of overhead to workflow logic.

For that reason, we're going to instead change it so each workflow has a single entity. It can be of each of the different types, but it can only have one. We'll still need a bit of validation logic, but that's amount to checking a type and a single column rather than complex iteration of lists upon lists.

To make certain the table has exactly 1 and only 1 column non-null, we can use a check constraint like `num_nonnulls(opportunity_id, application_id, application_submission_id) = 1` to do so.

This work is largely going to be refactoring everywhere that has a list of entities into a single entity.

### Acceptance criteria

- [ ] Remove existing entity tables for workflows, and change to a new single entity table
- [ ] Fixed logic and utilities to expect a single entity
- [ ] Fixed API route to expect a single entity

chouinar:
## Summary
Fixes #8658 

## Changes proposed
* Remove the one-to-many entity relationships from the workflow table
* Change workflows to instead have a single workflow entity object that can have exactly one entity associated with it.

## Context for reviewers
This PR goes against my own rule - never mix DB changes with code changes. But as we haven't actually deployed this code anywhere yet, and the only way to not mix them would require basically removing the workflow logic, I'll break my own rule.

We've had discussions in a few places about this work, but this largely boils down to simplifying a whole bunch of logic (most of this PR is simplifying logic and a lot of test simplification). 

## Validation steps
The instructions added next to the check constraint are from my own testing of adding additional columns later. Alembic cannot detect check constraints (there's not even an extension I could find, other constraints sure, but not these). So we have to manage this one manually, tried to make that abundantly clear to anyone looking at the model.

### DIFF
--- a/api/tests/workflow/listener/test_workflow_audit_listener.py
+++ b/api/tests/workflow/listener/test_workflow_audit_listener.py
@@ -19,7 +19,7 @@ def test_workflow_audit_created_on_start_workflow(db_session, enable_factory_cre
     event, history_event = build_start_workflow_event(
         workflow_type=WorkflowType.BASIC_TEST_WORKFLOW,
         user=user,
-        entities=[opportunity],
+        entity=opportunity,
     )
 
     event_handler = EventHandler(db_session, event, history_event)
@@ -54,7 +54,7 @@ def test_workflow_audit_created_on_process_workflow(db_session, enable_factory_c
     workflow = WorkflowFactory.create(
         workflow_type=WorkflowType.BASIC_TEST_WORKFLOW,
         current_workflow_state=BasicState.MIDDLE,
-        is_single_opportunity_workflow=True,
+        has_opportunity=True,
     )
 
     event, history_event = build_process_workflow_event(
@@ -91,7 +91,7 @@ def test_workflow_audit_captures_metadata(db_session, enable_factory_create):
     workflow = WorkflowFactory.create(
         workflow_type=WorkflowType.BASIC_TEST_WORKFLOW,
         current_workflow_state=BasicState.MIDDLE,
-        is_single_opportunity_workflow=True,
+        has_opportunity=True,
     )
 
     # Create event with metadata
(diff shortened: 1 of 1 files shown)


api/src/db/models/base.py
```diff
@@ -20,7 +20,7 @@
     naming_convention={
         "ix": "%(table_name)s_%(column_0_name)s_idx",
         "uq": "%(table_name)s_%(column_0_name)s_uniq",
-        "ck": "%(table_name)s_`%(constraint_name)s_check`",
+        "ck": "%(table_name)s_%(constraint_name)s_check",
```
chouinar, line 23: This config is used to generate the name of constraints automatically by alembic. I don't know why we had backticks in the check constraint naming - it didn't cause any issues from what I saw, but I didn't like that.

api/src/db/models/workflow_models.py
```diff
@@ -16,6 +15,77 @@
...
+    )
+
+    workflow_entity_id: Mapped[uuid.UUID] = mapped_column(
+        UUID, primary_key=True, default=uuid.uuid4
+    )
+
+    workflow_id: Mapped[uuid.UUID] = mapped_column(
+        ForeignKey("api.workflow.workflow_id"), unique=True
+    )
+    workflow: Mapped[Workflow] = relationship()
```
kkrug: Adding `ondelete="CASCADE"` to the FK here should prevent orphaned workflows, per @hao10282025-sudo  concern that he brought up in the Slack thread.

```suggestion
    workflow_id: Mapped[uuid.UUID] = mapped_column(
        ForeignKey("api.workflow.workflow_id", ondelete="CASCADE"), unique=True
    )
    workflow: Mapped[Workflow] = relationship()
```

kkrug: We can also use `back_populates` here to establish a bidirectional relationship to ensure they stay in sync. There's a few other places we can make a similar change but I wanted to get your thoughts before suggesting all of them.
```suggestion
    workflow: Mapped[Workflow] = relationship(back_populates="workflow_entity")
```

PR #8660, opened 2026-02-25
[Issue #8434] Add organization saved opportunity endpoint

### LINKED ISSUE #8434: [BE] Add organization saved opportunity endpoint
### Summary

To support saving an opportunity at the organization scope, we need to create a new POST endpoint `/v1/organizations/<organization_id>/saved-opportunities`.

Log message for successfully adding entry into `organization_saved_opportunities`:
```python
logger.info("added organization saved opportunity", extra={"organization_id": "id", "opportunity_id": "id"})
```

This endpoint will take in the following JSON request:
```json
{
  "opportunity_id": "942e4567-e89b-12d3-a456-426614174000",
}
```

Successful response: 
```json
{
  "data": {},
  "message": "Success",
  "status_code": 200
}
```

Opportunity already saved to organization (no-op): 
```json
{
  "data": {},
  "message": "Opportunity already saved to organization",
  "status_code": 200
}
```

Unauthorized:
```json
{
  "data": {},
  "message": "Unauthorized",
  "status_code": 401
}
```

User is not a part of the organization they are trying to access:
```json
{
  "data": {},
  "message": "User does not have access to this organization",
  "status_code": 403
}
```

No organization found by id:
```json
{
  "data": {},
  "message": "Organization not found",
  "status_code": 404
}
```

No opportunity found by id:
```json
{
  "data": {},
  "message": "Opportunity not found",
  "status_code": 404
}
```

Invalid request:
```json
{
  "data": {},
  "message": "Opportunity and organization IDs are required",
  "status_code": 422
}
```

### Acceptance criteria
- [ ] A new endpoint that handles the above scenarios are implemented.
- [ ] Sample request included in implementation ticket.
- [ ] Appropriate logging for successful operation
- [ ] Unit tests pass

mikehgrantsgov:
## Summary

Fixes #8434

## Changes proposed

New service `create_organization_saved_opportunity`. It validates org and opportunity exist, checks user has `VIEW_ORG_MEMBERSHIP `privilege, inserts into `organization_saved_opportunity`, returns a boolean indicating new vs. duplicate
Request/response schemas added to `organization_schemas.py`
Add unit tests

## Context for reviewers

To support saving an opportunity at the organization scope, we need to create a new POST endpoint /v1/organizations/<organization_id>/saved-opportunities.

Also fixed a fragile test causing CI/CD errors:
image

## Validation steps

Run:
`make test args="tests/src/api/organizations_v1/test_organization_save_opportunity_post.py -v"`

### DIFF
--- a/api/src/services/organizations_v1/create_organization_saved_opportunity.py
+++ b/api/src/services/organizations_v1/create_organization_saved_opportunity.py
@@ -5,10 +5,12 @@ from sqlalchemy import select
 
 from src.adapters import db
 from src.api.route_utils import raise_flask_error
+from src.auth.endpoint_access_util import check_user_access
+from src.constants.lookup_constants import Privilege
 from src.db.models.entity_models import OrganizationSavedOpportunity
 from src.db.models.opportunity_models import Opportunity
 from src.db.models.user_models import User
-from src.services.organizations_v1.get_organization import get_organization_and_verify_access
+from src.services.organizations_v1.get_organization import get_organization
 
 logger = logging.getLogger(__name__)
 
@@ -24,8 +26,8 @@ def create_organization_saved_opportunity(
 
     Returns True if a new record was created, False if already saved.
     """
-    # Validate organization exists and fetch it
-    get_organization_and_verify_access(db_session, user, organization_id)
+    organization = get_organization(db_session, organization_id)
+    check_user_access(db_session, user, {Privilege.MODIFY_ORG_SAVED_OPPORTUNITIES}, organization)
 
     # Validate opportunity exists and is not in draft status
     opportunity_id = json_data["opportunity_id"]
--- a/api/tests/src/api/organizations_v1/test_organization_save_opportunity_post.py
+++ b/api/tests/src/api/organizations_v1/test_organization_save_opportunity_post.py
@@ -17,7 +17,7 @@ class TestOrganizationSaveOpportunityPost:
         """New save returns 200 with Success message"""
         opportunity = OpportunityFactory.create(is_draft=False)
         user, organization, token = create_user_in_org(
-            privileges=[Privilege.VIEW_ORG_MEMBERSHIP],
+            privileges=[Privilege.MODIFY_ORG_SAVED_OPPORTUNITIES],
             db_session=db_session,
         )
 
@@ -44,7 +44,7 @@ class TestOrganizationSaveOpportunityPost:
         """Saving an already-saved opportunity returns 200 without creating a duplicate"""
         opportunity = OpportunityFactory.create(is_draft=False)
         user, organization, token = create_user_in_org(
-            privileges=[Privilege.VIEW_ORG_MEMBERSHIP],
+            privileges=[Privilege.MODIFY_ORG_SAVED_OPPORTUNITIES],
             db_session=db_session,
         )
         OrganizationSavedOpportunityFactory.create(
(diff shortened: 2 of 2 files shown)


api/src/api/organizations_v1/organization_routes.py
```diff
@@ -294,3 +299,32 @@ def organization_ignore_legacy_user(
...
+        # Add the user from the token session to our current session
+        db_session.add(user_token_session)
+
+        is_new = create_organization_saved_opportunity(
+            db_session, user_token_session.user, organization_id, json_data
+        )
+
+    if is_new:
+        return response.ApiResponse(message="Success", data={}), 201
+    return response.ApiResponse(message="Opportunity already saved to organization", data={}), 200
```
chouinar: We don't currently return non-200s for success in our endpoints, I'm hesitant to have us start for consistency. If we wanted to do this, we'd probably want to do it for 30+ endpoints, we'd at the very least want it consistent with the other saved opportunity endpoint.

api/src/services/organizations_v1/create_organization_saved_opportunity.py
```diff
@@ -0,0 +1,78 @@
...
+    # Validate organization exists
+    organization = db_session.execute(
+        select(Organization).where(Organization.organization_id == organization_id)
+    ).scalar_one_or_none()
+
+    if organization is None:
+        raise_flask_error(404, message=f"Could not find Organization with ID {organization_id}")
+
+    # Check user has access to organization
+    check_user_access(db_session, user, {Privilege.VIEW_ORG_MEMBERSHIP}, organization)
```
chouinar: This is all in an existing util.
```suggestion
    # Validate organization exists and fetch it
    organization = get_organization_and_verify_access(db_session, user, organization_id)
```

joshtonava: @babebe is working on something that touches this util as well

joshtonava: Im wondering if this privilege still makes sense for what we are doing here

joshtonava: In @babebe's PR we are adding a privilege `VIEW_ORG_SAVED_OPPORTUNITIES`, im wondering if `MANAGE_ORG_SAVED_OPPORTUNITIES` could be a privilege to add that would be more clear

api/src/services/organizations_v1/create_organization_saved_opportunity.py
```diff
@@ -0,0 +1,78 @@
...
+            Opportunity.opportunity_id == opportunity_id,
+            Opportunity.is_draft.is_(False),
+        )
+    ).scalar_one_or_none()
+
+    if opportunity is None:
+        raise_flask_error(404, message=f"Could not find Opportunity with ID {opportunity_id}")
+
+    # Check if already saved
+    existing = db_session.execute(
```
joshtonava, line 45: note: This is not blocking but commenting but wanted to point out that there is potential race condition here.
If two people are trying to save the same opportunity at the same time one could throw an error on the save.

chouinar: These endpoints probably take ~100ms to respond, so as long as they're not in that window, we shouldn't hit an issue. If they're slightly further apart, this'll see that it already exists, and just get the slightly different message (that we probably won't show anyways).

If after this select, and before the commit a separate request caused it to be saved, we would probably hit a constraint issue on the DB and 500. But that's true of several endpoints, and probably like a ~5ms leeway to happen. If we see that happen for all of our endpoints more than once a year, I'll be surprised. Could work around by writing these as upsert queries (which adds complexity), but I'm happy to accept that edge case for clearer code.

Issue #8499, opened 2026-02-11
Update Opportunities List page to use single-agency view


glenblosser-nava:
### Summary

Instead of listing all Opportunities for all Agencies a given User belongs to, only display the Opportunities for a single Agency at a time.
If a user belongs to multiple Agencies, there should be a drop down selector above the table to choose which Agency's Opportunities to display. It should default to the first listed Agency.
If a user belongs to a single Agency, do not display the dropdown.
The Agency selected should be reflected in the URL in some way, so that a user can save and/or send links to a given Agency's Opportunity.
If a User attempts to access a page for an Agency they are not a part of, an error should be displayed instead of showing the Agency's Opportunities.

### Acceptance criteria

- [ ] A multi-agency user can select an agency in the dropdown
- [ ] Changing agency in the dropdown changes which Opportunities are displayed
- [ ] A single-agency user cannot see the dropdown
- [ ] Copying the url while on a specific agency view results in the user seeing that agency's opportunities
- [ ] A user not associated with a given Agency cannot see that Agency's Opportunities

crystabelrangel:
@desaiamit [here is the page in Figma](https://www.figma.com/design/o5g0d2W5RKHYHjdvVCqTV2/Grantor-Designs-%7C-Quad-5-%7C-Simpler-Grants-2026?node-id=5741-42559&t=JA2oTLm2ngbJRwHe-1)

brianlee-nava:
UATed this ticket with @desaiamit. Approved. Moving to Done.

PR #8671, opened 2026-02-26
[Issue #8499] Update Opportunities List page to single-agency view

desaiamit:
## Summary

Work for #8499

## Changes proposed

- Add `userAgenciesFetcher.ts` — server-only fetcher that calls `POST /v1/users/<user_id>/agencies` to retrieve the logged-in user's associated agencies
- Add `AgencySelector.tsx` — client component that renders a dropdown for multi-agency users and updates the URL on change
- Update `OpportunitiesListPage` to:
  - Redirect to `?agency=<first_agency_id>` if no agency param is in the URL
  - Show an error if the agency in the URL is not in the user's agencies
  - Show an error if the user has no agency associations
  - Filter opportunities by the selected agency's `agency_code`
  - Show the agency dropdown only for users with 2+ agencies
- Add i18n strings: `agencySelector`, `agencyNotAuthorized`, `noAgencies`

## Context for reviewers

The page now uses URL state (`?agency=<agency_id>`) to track the selected agency. The frontend enforces the authorization check by comparing the URL agency against the list returned by the user agencies endpoint — no separate API call is needed for this check.

The `AgencySelector` is a client component to support `useRouter` for navigation on dropdown change. The page itself remains a server component.

### Prerequisites
```
cd api && make db-seed-local-with-agencies
```
Ensure API and frontend are running locally.

### Scenario 1 — Single agency: no dropdown

1. Click **Sign in**, enter `one_agency_opp_edit` on the Mock OAuth2 Sign-in page
2. Navigate to `http://localhost:3000/opportunities`
3. **Expected:** Redirects to `?agency=<uuid>` automatically
4. **Expected:** No "Select agency" dropdown visible
5. **Expected:** Table shows opportunities (or empty state) for USAID-ETH only

### Scenario 2 — Multi-agency: dropdown shown

1. Sign out, click **Sign in**, enter `two_agency_opp_pub`
2. Navigate to `http://localhost:3000/opportunities`
3. **Expected:** Redirects to `?agency=<uuid>` for the first agency
4. **Expected:** "Select agency" dropdown visible with USAID-ETH and USAID-SAF
5. Select the other agency from the dropdown
6. **Expected:** URL updates to `?agency=<other-uuid>` and table refreshes

### Scenario 3 — Unauthorized agency in URL

1. While logged in as `one_agency_opp_edit`, navigate to:
   `http://localhost:3000/opportunities?agency=00000000-0000-0000-0000-000000000000`
2. **Expected:** Red alert — "You do not have access to this agency's opportunities."

### Scenario 4 — Copy/paste URL preserves agency view

1. Log in as `two_agency_opp_pub`, navigate to `/opportunities`
2. Use the dropdown to switch to the second agency — URL updates to `?agency=<uuid-B>`
3. Copy the full URL, open a new tab, paste and hit Enter
4. **Expected:** Loads directly into that agency's view without redirecting

### DIFF
--- a/frontend/src/app/[locale]/(base)/opportunities/page.tsx
+++ b/frontend/src/app/[locale]/(base)/opportunities/page.tsx
@@ -157,9 +157,13 @@ const OpportunitiesTable = ({
 
 async function OpportunitiesListPage(props: OpportunitiesListProps) {
   const { searchParams } = props;
-  const resolvedSearchParams = searchParams ? await searchParams : {};
-  const selectedAgencyParam = resolvedSearchParams.agency;
-  const selectedAgencyId = Array.isArray(selectedAgencyParam)
+  const resolvedSearchParams: Record<string, string | string[] | undefined> =
+    searchParams ? await searchParams : {};
+  const selectedAgencyParam: string | string[] | undefined =
+    resolvedSearchParams.agency;
+  const selectedAgencyId: string | undefined = Array.isArray(
+    selectedAgencyParam,
+  )
     ? selectedAgencyParam[0]
     : selectedAgencyParam;
 
--- a/frontend/src/services/fetch/fetchers/userAgenciesFetcher.test.ts
+++ b/frontend/src/services/fetch/fetchers/userAgenciesFetcher.test.ts
@@ -1,4 +1,3 @@
-import { UnauthorizedError } from "src/errors";
 import {
   fetchUserAgencies,
   getUserAgencies,


myduong-navapbc (approved):

frontend/src/app/[locale]/(base)/opportunities/page.tsx
```diff
@@ -107,14 +155,45 @@ const OpportunitiesTable = ({
   );
 };
 
-async function OpportunitiesListPage(_props: OpportunitiesListProps) {
-  const searchParams = convertSearchParamsToProperTypes({});
+async function OpportunitiesListPage(props: OpportunitiesListProps) {
+  const { searchParams } = props;
+  const resolvedSearchParams = searchParams ? await searchParams : {};
+  const selectedAgencyId = resolvedSearchParams.agency;
```
Comment (myduong-navapbc) [inline frontend/src/app/[locale]/(base)/opportunities/page.tsx]: `resolvedSearchParams.agency` may be `string` | `string[] `| `undefined`, but it is treated as a `string` throughout the page. Would it be safer to normalize this value before using it in `find()` and passing it into `AgencySelector`?

(https://nextjs.org/docs/app/api-reference/file-conventions/page#searchparams)

TLDR: the example provided

```
export default function Page({
  searchParams,
}: {
  searchParams?: { [key: string]: string | string[] | undefined }
}) {
```

(youo can see the types in `generalTypes.ts`)

frontend/src/app/[locale]/(base)/opportunities/page.tsx
```diff
@@ -107,14 +155,45 @@ const OpportunitiesTable = ({
...
+    }
+    return <OpportunitiesErrorPage />;
+  }
+
+  if (!userAgencies.length) {
+    return <NoAgenciesPage />;
+  }
+
+  if (!selectedAgencyId) {
+    redirect(`?agency=${userAgencies[0].agency_id}`);
```
Comment (myduong-navapbc) [inline frontend/src/app/[locale]/(base)/opportunities/page.tsx]: [future] When defaulting the user to the first agency, we redirect to `?agency=...,` which replaces the full query string. That works for this ticket, but it would drop future params like `pagination`, `filters`, or `sorting`. It might be safer to preserve existing params and only set the agency value.

the example that comes to mind is if I have pagination enabled, these are the params `?agency=1?page=2`. i

again, this is future proofing.

frontend/src/components/workspace/AgencySelector.tsx
```diff
@@ -0,0 +1,38 @@
...
+    <div className="usa-form-group margin-bottom-4">
+      <label className="usa-label text-bold" htmlFor="agency-selector">
+        {t("agencySelector")}
+      </label>
+      <select
+        className="usa-select"
+        id="agency-selector"
+        name="agency"
+        defaultValue={currentAgencyId}
+        onChange={(e) => router.push(`?agency=${e.target.value}`)}
```
myduong-navapbc, line 28: [future] `router.push("?agency=...")` replaces the entire query string. If the page later adds `pagination`, `sorting`, or `filters`, switching agencies would clear them. It may be safer to merge the new agency value into the existing search params instead of replacing them.

frontend/src/services/fetch/fetchers/userAgenciesFetcher.ts
```diff
@@ -0,0 +1,33 @@
...
+  token: string,
+  userId: string,
+): Promise<UserAgency[]> => {
+  const subPath = `${userId}/agencies`;
+  const resp = await fetchUserWithMethod("POST")({
+    subPath,
+    additionalHeaders: { "X-SGG-Token": token },
+    body: {},
+  });
+  const json = (await resp.json()) as { data: UserAgency[] };
```
myduong-navapbc, line 32: The fetcher parses `resp.json()` without checking `resp.ok.` If the upstream endpoint returns an error response, we could end up with a runtime error instead of a clear fetch failure. Could we check `resp.ok` before reading `json.data`?

example: take a look at `src/services/fetch/fetchers/organizationsFetcher.ts`

you can do something like:
```
if (!resp.ok) {
  if (resp.status === 401) {
    throw new UnauthorizedError("Unauthorized");
  }

  throw new Error(`Failed to fetch user agencies: ${resp.status}`);
}

const json = await resp.json();
return json.data;
```

myduong-navapbc:
[low priority] `UserAgency` is now used across the fetcher, page, component, and tests. It might make sense to move it into a shared types module so UI code does not depend on the fetcher layer.

```
export interface UserAgency {
  agency_id: string
  agency_name: string
  agency_code: string
}
```
consider adding `agencyTypes` to `frontend/src/types`

myduong-navapbc:
[question] Since the default agency depends on `userAgencies[0]`, do we know whether the agencies endpoint returns a stable ordering? Just want to confirm the default behavior is deterministic.

frontend/src/app/[locale]/(base)/opportunities/page.tsx
Comment (desaiamit) [reply frontend/src/app/[locale]/(base)/opportunities/page.tsx]: Great callout. You’re right that searchParams.agency can be string | string[] | undefined and should be normalized before use. I’ll update this PR to normalize the value to a single string | undefined before find() and before passing to AgencySelector.

frontend/src/app/[locale]/(base)/opportunities/page.tsx
Comment (desaiamit) [reply frontend/src/app/[locale]/(base)/opportunities/page.tsx]: Good point. Right now this page only uses agency in the query string, so replacing the query is functionally safe for current scope. I agree we should preserve existing params as soon as pagination/sorting/filter params are introduced. I’ll track this as follow-up and update navigation to merge params rather than replace them.

frontend/src/services/fetch/fetchers/userAgenciesFetcher.ts
desaiamit: Good catch. I’ll add explicit `resp.ok` handling before reading `json.data` (including 401 -> `UnauthorizedError`) to align with existing fetcher patterns and provide clearer failures.

frontend/src/components/workspace/AgencySelector.tsx
desaiamit: As noted in the page.tsx comment, I’ll track this selector-level merge behavior as follow-up and update router.push to retain existing query params once additional query-driven state is introduced.

desaiamit:
> [low priority] `UserAgency` is now used across the fetcher, page, component, and tests. It might make sense to move it into a shared types module so UI code does not depend on the fetcher layer.
> 
> ```
> export interface UserAgency {
>   agency_id: string
>   agency_name: string
>   agency_code: string
> }
> ```
> 
> consider adding `agencyTypes` to `frontend/src/types`

Thanks for the suggestion. Agreed this type may belong in a shared types module as usage grows. For this PR, I kept it scoped to the fetcher/feature implementation to minimize surface-area changes. I’ll open a follow-up to move UserAgency into frontend/src/types once we confirm all intended consumers for the opportunity publishing flow.

desaiamit:
> [question] Since the default agency depends on `userAgencies[0]`, do we know whether the agencies endpoint returns a stable ordering? Just want to confirm the default behavior is deterministic.

Agreed, relying on API order without an explicit contract can be brittle. I’ll make the updates so first-load behavior is stable.

PR #8691, opened 2026-02-26
[Issue #6667]: Add ConfirmApplicationDeliverySchema

jakobpederson:
## Summary

Fixes / Work for #6667 

## Changes proposed

Add schema for SOAP/Proxy `ConfirmApplicationDelivery`.

## Context for reviewers

As part of the step of building out the endpoint we need the schemas for the `Request` and `Response` aspect of `ConfirmApplicationDelivery`. The structures can be found [here](https://www.grants.gov/system-to-system/grantor-system-to-system/web-services/confirm-application-delivery).

## Validation steps

- [ ] unit tests passing

### DIFF
--- a/api/src/legacy_soap_api/grantors/schemas/confirm_application_delivery_schemas.py
+++ b/api/src/legacy_soap_api/grantors/schemas/confirm_application_delivery_schemas.py
@@ -1,39 +1,24 @@
 import re
 
-from pydantic import AliasChoices, Field, field_validator
+from pydantic import Field, field_validator
 
-from src.legacy_soap_api.applicants.fault_messages import OpportunityListRequestInvalidParams
+from src.legacy_soap_api.applicants.fault_messages import (
+    ConfirmApplicationDeliveryInvalidTrackingNumber,
+)
 from src.legacy_soap_api.legacy_soap_api_schemas import BaseSOAPSchema
 from src.legacy_soap_api.legacy_soap_api_utils import SOAPFaultException
 
-CONFIRM_APPLICATION_DELIVERY_NO_TRACKING_NUMBER_ERR = "No grants_gov_tracking_number provided."
 CONFIRM_APPLICATION_DELIVERY_INVALID_TRACKING_NUMBER_ERR = (
     "Invalid grants_gov_tracking_number provided."
 )
 
 
-class ConfirmApplicationDeliveryResponse(BaseSOAPSchema):
+class ConfirmApplicationDeliveryResponseSOAPEnvelope(BaseSOAPSchema):
     grants_gov_tracking_number: str | None = Field(default=None, alias="GrantsGovTrackingNumber")
     response_message: str | None = Field(default=None, alias="ResponseMessage")
 
-
-class ConfirmApplicationDeliveryResponseSOAPBody(BaseSOAPSchema):
-    confirm_application_delivery_response: ConfirmApplicationDeliveryResponse = Field(
-        alias="ns2:ConfirmApplicationDeliveryResponse",
-        # From testing it looks like the response comes in with the namespace ns2 but when we convert
-        # it to a dict via the SOAPPayload the namespace can be lost so using validation_alias in order
-        # to handle both cases
-        validation_alias=AliasChoices(
-            "ns2:ConfirmApplicationDeliveryResponse", "ConfirmApplicationDeliveryResponse"
-        ),
-    )
-
-
-class ConfirmApplicationDeliveryResponseSOAPEnvelope(BaseSOAPSchema):
-    body: ConfirmApplicationDeliveryResponseSOAPBody = Field(alias="Body")
-
     def to_soap_envelope_dict(self, operation_name: str) -> dict:
-        return {"Envelope": self.model_dump(by_alias=True)}
+        return super().to_soap_envelope_dict(f"ns2:{operation_name}")
 
 
 class ConfirmApplicationDeliveryRequest(BaseSOAPSchema):
@@ -42,14 +27,9 @@ class ConfirmApplicationDeliveryRequest(BaseSOAPSchema):
     @field_validator("grants_gov_tracking_number", mode="after")
     @classmethod
     def validate_required_properties(cls, value: str) -> str:
-        if not value:
-            raise SOAPFaultException(
-                CONFIRM_APPLICATION_DELIVERY_NO_TRACKING_NUMBER_ERR,
-                fault=OpportunityListRequestInvalidParams,
-            )
-        if not re.fullmatch(r"GRANT[0-9]{8}", value):
+        if not value or not re.fullmatch(r"GRANT[0-9]{8}", value):
             raise SOAPFaultException(
                 CONFIRM_APPLICATION_DELIVERY_INVALID_TRACKING_NUMBER_ERR,
-                fault=OpportunityListRequestInvalidParams,
+                fault=ConfirmApplicationDeliveryInvalidTrackingNumber,
             )
         return value
--- a/api/src/legacy_soap_api/legacy_soap_api_config.py
+++ b/api/src/legacy_soap_api/legacy_soap_api_config.py
@@ -77,7 +77,6 @@ class SOAPOperationConfig:
     request_operation_name: str
     response_operation_name: str
     compare_endpoints: bool = False
-    is_mtom: bool = False
     always_call_simpler: bool = False
 
     # These are the privileges needed for these endpoints:
@@ -142,19 +141,16 @@ SIMPLER_SOAP_OPERATION_CONFIGS: dict[SimplerSoapAPI, dict[str, SOAPOperationConf
         "GetApplicationZipRequest": SOAPOperationConfig(
             request_operation_name="GetApplicationZipRequest",
             response_operation_name="GetApplicationZipResponse",
-            is_mtom=True,
             privileges={Privilege.LEGACY_AGENCY_GRANT_RETRIEVER},
         ),
         "GetSubmissionListExpandedRequest": SOAPOperationConfig(
             request_operation_name="GetSubmissionListExpandedRequest",
             response_operation_name="GetSubmissionListExpandedResponse",
-            is_mtom=True,
             privileges={Privilege.LEGACY_AGENCY_GRANT_RETRIEVER},
         ),
         "ConfirmApplicationDeliveryRequest": SOAPOperationConfig(
             request_operation_name="ConfirmApplicationDeliveryRequest",
             response_operation_name="ConfirmApplicationDeliveryResponse",
-            is_mtom=True,
             privileges={Privilege.LEGACY_AGENCY_GRANT_RETRIEVER},
         ),
     },
--- a/api/tests/src/legacy_soap_api/grantors/schemas/test_confirm_application_delivery_schema.py
+++ b/api/tests/src/legacy_soap_api/grantors/schemas/test_confirm_application_delivery_schema.py
@@ -14,7 +14,7 @@ RESPONSE_MESSAGE = "Success"
 
 
 class TestLegacySoapGrantorConfirmApplicationRequestSchema:
-    def test_can_convert_confirm_application_delivery_request_soap_payload_dict(self, db_session):
+    def test_can_convert_confirm_application_delivery_request_soap_payload_dict(self):
         request_xml = (
             "<soapenv:Envelope "
             'xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/" '
@@ -44,7 +44,7 @@ class TestLegacySoapGrantorConfirmApplicationRequestSchema:
         }
         assert soap_payload_dict == expected
 
-    def test_can_convert_confirm_application_delivery_request_xml_bytes_to_dict(self, db_session):
+    def test_can_convert_confirm_application_delivery_request_xml_bytes_to_dict(self):
         request_xml = (
             "<soapenv:Envelope "
             'xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/" '
@@ -68,7 +68,7 @@ class TestLegacySoapGrantorConfirmApplicationRequestSchema:
         }
 
     def test_confirm_application_delivery_request_validates_there_is_a_grants_gov_tracking_number_exists(
-        self, db_session
+        self,
     ):
         request_xml = (
             "<soapenv:Envelope "
@@ -88,10 +88,10 @@ class TestLegacySoapGrantorConfirmApplicationRequestSchema:
         )
         with pytest.raises(SOAPFaultException) as e:
             grantors_schemas.ConfirmApplicationDeliveryRequest(**soap_operation_dict)
-        assert e.value.message == "No grants_gov_tracking_number provided."
+        assert e.value.message == "Invalid grants_gov_tracking_number provided."
 
     def test_confirm_application_delivery_request_validates_there_is_a_grants_gov_tracking_number_is_in_correct_format(
-        self, db_session
+        self,
     ):
         BAD_NUMBER = "123456"
         request_xml = (
@@ -116,7 +116,7 @@ class TestLegacySoapGrantorConfirmApplicationRequestSchema:
 
 
 class TestLegacySoapGrantorConfirmApplicationResponseSchema:
-    def test_can_convert_confirm_application_delivery_response_xml_bytes_to_dict(self, db_session):
+    def test_can_convert_confirm_application_delivery_response_xml_bytes_to_dict(self):
         response_xml_bytes = (
             f"--uuid:{BOUNDARY_UUID}"
             'Content-Type: application/xop+xml; charset=UTF-8; type="text/xml"'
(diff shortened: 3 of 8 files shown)


api/src/legacy_soap_api/legacy_soap_api_config.py
```diff
@@ -151,6 +151,12 @@ class SOAPOperationConfig:
             is_mtom=True,
             privileges={Privilege.LEGACY_AGENCY_GRANT_RETRIEVER},
         ),
+        "ConfirmApplicationDeliveryRequest": SOAPOperationConfig(
+            request_operation_name="ConfirmApplicationDeliveryRequest",
+            response_operation_name="ConfirmApplicationDeliveryResponse",
+            is_mtom=True,
+            privileges=set(),
```
chouinar: This can be `{Privilege.LEGACY_AGENCY_GRANT_RETRIEVER}` - double-checked what I had written up in the tech spec / pulled from grants.gov.

https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2125398123/Tech+Spec+Soap+Cert+Mapping+to+Simpler+Auth#Agency-SOAP-Operations-and-Required-Privileges

api/src/legacy_soap_api/legacy_soap_api_config.py
```diff
@@ -151,6 +151,12 @@ class SOAPOperationConfig:
             is_mtom=True,
             privileges={Privilege.LEGACY_AGENCY_GRANT_RETRIEVER},
         ),
+        "ConfirmApplicationDeliveryRequest": SOAPOperationConfig(
+            request_operation_name="ConfirmApplicationDeliveryRequest",
+            response_operation_name="ConfirmApplicationDeliveryResponse",
+            is_mtom=True,
```
chouinar: Does this need to be mtom? The response looks like it's just some very basic XML - no complex file/streaming needed.

api/src/legacy_soap_api/grantors/schemas/confirm_application_delivery_schemas.py
```diff
@@ -0,0 +1,47 @@
...
+    def to_soap_envelope_dict(self, operation_name: str) -> dict:
+        envelope_dict = {"Envelope": self.model_dump(by_alias=True)}
+        return envelope_dict
+
+
+class ConfirmApplicationDeliveryRequest(BaseSOAPSchema):
+    grants_gov_tracking_number: str | None = Field(default=None, alias="GrantsGovTrackingNumber")
+
+    @model_validator(mode="after")
+    def validate_required_properties(self) -> Self:
```
chouinar: I don't think it matters much either way, but could this be a field validator? https://docs.pydantic.dev/latest/concepts/validators/#field-validators

Also, we should probably validate the format is `GRANT[0-9]{8}`. You'll have to test to see what the error message would be in that case.

api/src/legacy_soap_api/legacy_soap_api_config.py
jakobpederson: It looks like if it's NOT mtom then we just return the proxy response
```
        if self.operation_config.is_mtom is False:
            return proxy_response
```

api/src/legacy_soap_api/grantors/schemas/confirm_application_delivery_schemas.py
jakobpederson: updated to field_validator. I think we could also not allow `None` for the `grants_gov_tracking_number` but  that would break the established pattern.

api/src/legacy_soap_api/legacy_soap_api_config.py
chouinar: That seems wrong? We don't want mtom for this endpoint, and do want to return our implementation (once built). How would that work?

How do other endpoints work that don't need mtom that we built?

api/src/legacy_soap_api/grantors/schemas/confirm_application_delivery_schemas.py
```diff
@@ -0,0 +1,48 @@
...
+
+    @field_validator("grants_gov_tracking_number", mode="after")
+    @classmethod
+    def validate_required_properties(cls, value: str) -> str:
+        if not value:
+            raise SOAPFaultException(
+                GET_APPLICATION_ZIP_REQUEST_ERR,
+                fault=OpportunityListRequestInvalidParams,
+            )
+        return value
```
chouinar: If the format isn't correct, what error does it return / does it match grants.gov? From the test you added, it's a pydantic error, which isn't going to be right. I think we need to do the regex check in the validator so we can raise the right error.

jakobpederson:
@chouinar I think the `mtom` thing is a naming issue. When `is_mtom` is true, it constructs the xml message 
```
        # 1. --uuid: {boundary_uuid}\n
        # 2. headers:
        #    'Content-Type: application/xop+xml; charset=UTF-8; type="text/xml"\n'
        #    "Content-Transfer-Encoding: binary\n"
        #    "Content-ID: <root.message@cxf.apache.org>\n\n"
        # 3. MTOM xml body
        # 4. --uuid: {boundary_uuid}
        # 5. the file bytes from the file being attached
        # 6. --uuid: {boundary_uuid}--
```
And if there is an `mtom_file_stream` it injects them. So if `is_mtom` is false, just return the proxy, if it's true construct our own xml. 

I think this is correct behavior though maybe it's more like `build_xml_response` would be a better name.

chouinar:
> @chouinar I think the `mtom` thing is a naming issue. When `is_mtom` is true, it constructs the xml message
> 
> ```
>         # 1. --uuid: {boundary_uuid}\n
>         # 2. headers:
>         #    'Content-Type: application/xop+xml; charset=UTF-8; type="text/xml"\n'
>         #    "Content-Transfer-Encoding: binary\n"
>         #    "Content-ID: <root.message@cxf.apache.org>\n\n"
>         # 3. MTOM xml body
>         # 4. --uuid: {boundary_uuid}
>         # 5. the file bytes from the file being attached
>         # 6. --uuid: {boundary_uuid}--
> ```
> 
> And if there is an `mtom_file_stream` it injects them. So if `is_mtom` is false, just return the proxy, if it's true construct our own xml.
> 
> I think this is correct behavior though maybe it's more like `build_xml_response` would be a better name.

Please correct me if my understanding is wrong. There are two types of responses we can form:
* A regular XML response - for any endpoint that doesn't need to return files. This doesn't have to have all of the boundary and other stuff in it. By default, this should be the behavior of most endpoints
* mtom - Which has all the boundary stuff because of how it structures file responses

This endpoint wants the ordinary case.

As written, the way `is_mtom` works is that if it's true, it does the latter. But if it's false for a grantor endpoint, it returns the proxy response - that is, the response from grants.gov.

How do we return an ordinary response without the mtom wrapper for a grantor endpoint?

jakobpederson:
> > @chouinar I think the `mtom` thing is a naming issue. When `is_mtom` is true, it constructs the xml message
> > ```
> >         # 1. --uuid: {boundary_uuid}\n
> >         # 2. headers:
> >         #    'Content-Type: application/xop+xml; charset=UTF-8; type="text/xml"\n'
> >         #    "Content-Transfer-Encoding: binary\n"
> >         #    "Content-ID: <root.message@cxf.apache.org>\n\n"
> >         # 3. MTOM xml body
> >         # 4. --uuid: {boundary_uuid}
> >         # 5. the file bytes from the file being attached
> >         # 6. --uuid: {boundary_uuid}--
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And if there is an `mtom_file_stream` it injects them. So if `is_mtom` is false, just return the proxy, if it's true construct our own xml.
> > I think this is correct behavior though maybe it's more like `build_xml_response` would be a better name.
> 
> Please correct me if my understanding is wrong. There are two types of responses we can form:
> 
> * A regular XML response - for any endpoint that doesn't need to return files. This doesn't have to have all of the boundary and other stuff in it. By default, this should be the behavior of most endpoints
> * mtom - Which has all the boundary stuff because of how it structures file responses
> 
> This endpoint wants the ordinary case.
> 
> As written, the way `is_mtom` works is that if it's true, it does the latter. But if it's false for a grantor endpoint, it returns the proxy response - that is, the response from grants.gov.
> 
> How do we return an ordinary response without the mtom wrapper for a grantor endpoint?

I may be mistaken here but I don't think there's any "regular XML responses". Every response I've seen has the MTOM boundary stuff. For example, here's a failed response from `ConfirmApplicationDelivery` through Postman
```
--uuid:4583e02b-9320-4d05-a147-b4a00634a447
Content-Type: application/xop+xml; charset=UTF-8; type="text/xml"
Content-Transfer-Encoding: binary
Content-ID: <root.message@cxf.apache.org>

<soap:Envelope xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/"><soap:Body><soap:Fault><faultcode>soap:Server</faultcode><faultstring>Failed to confirm application delivery.(Expected an Application status of:'Validated' , but found a status of 'Received by Agency' for GRANT00000525)</faultstring></soap:Fault></soap:Body></soap:Envelope>
--uuid:4583e02b-9320-4d05-a147-b4a00634a447--
```

So we can return proxy response, which is also MTOM boundaried, or we response with an MTOM boundaried message.

chouinar:
> > > @chouinar I think the `mtom` thing is a naming issue. When `is_mtom` is true, it constructs the xml message
> > > ```
> > >         # 1. --uuid: {boundary_uuid}\n
> > >         # 2. headers:
> > >         #    'Content-Type: application/xop+xml; charset=UTF-8; type="text/xml"\n'
> > >         #    "Content-Transfer-Encoding: binary\n"
> > >         #    "Content-ID: <root.message@cxf.apache.org>\n\n"
> > >         # 3. MTOM xml body
> > >         # 4. --uuid: {boundary_uuid}
> > >         # 5. the file bytes from the file being attached
> > >         # 6. --uuid: {boundary_uuid}--
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > And if there is an `mtom_file_stream` it injects them. So if `is_mtom` is false, just return the proxy, if it's true construct our own xml.
> > > I think this is correct behavior though maybe it's more like `build_xml_response` would be a better name.
> > 
> > 
> > Please correct me if my understanding is wrong. There are two types of responses we can form:
> > 
> > * A regular XML response - for any endpoint that doesn't need to return files. This doesn't have to have all of the boundary and other stuff in it. By default, this should be the behavior of most endpoints
> > * mtom - Which has all the boundary stuff because of how it structures file responses
> > 
> > This endpoint wants the ordinary case.
> > As written, the way `is_mtom` works is that if it's true, it does the latter. But if it's false for a grantor endpoint, it returns the proxy response - that is, the response from grants.gov.
> > How do we return an ordinary response without the mtom wrapper for a grantor endpoint?
> 
> I may be mistaken here but I don't think there's any "regular XML responses". Every response I've seen has the MTOM boundary stuff. For example, here's a failed response from `ConfirmApplicationDelivery` through Postman
> 
> ```
> --uuid:4583e02b-9320-4d05-a147-b4a00634a447
> Content-Type: application/xop+xml; charset=UTF-8; type="text/xml"
> Content-Transfer-Encoding: binary
> Content-ID: <root.message@cxf.apache.org>
> 
> <soap:Envelope xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/"><soap:Body><soap:Fault><faultcode>soap:Server</faultcode><faultstring>Failed to confirm application delivery.(Expected an Application status of:'Validated' , but found a status of 'Received by Agency' for GRANT00000525)</faultstring></soap:Fault></soap:Body></soap:Envelope>
> --uuid:4583e02b-9320-4d05-a147-b4a00634a447--
> ```
> 
> So we can return proxy response, which is also MTOM boundaried, or we response with an MTOM boundaried message.

I'm missing something in the implementation.

We have [get_simpler_soap_response](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/legacy_soap_api/legacy_soap_api_client.py#L164) defined as a base function. This is overriden for the grantor client, but not for the applicant. These two functions have two fundamentally different approaches. The base method converts the response data to XML. The grantor approach converts the data into an mtom file stream but I don't see where it calls any conversion of dicts to XML.

I'm really lost on how these work. I had been up to this point interpreting the MTOM concept as separate from XML (yes, they both have some UUID boundaries, but all the file stuff is what separates them).

This isn't just a config question, I'm also just not getting how a response even gets built in either case. In the applicant endpoint, where does the UUID come from? In the grantor case where does the XML go? Am I missing something here?

api/src/legacy_soap_api/grantors/schemas/confirm_application_delivery_schemas.py
```diff
@@ -0,0 +1,55 @@
...
+class ConfirmApplicationDeliveryRequest(BaseSOAPSchema):
+    grants_gov_tracking_number: str | None = Field(default=None, alias="GrantsGovTrackingNumber")
+
+    @field_validator("grants_gov_tracking_number", mode="after")
+    @classmethod
+    def validate_required_properties(cls, value: str) -> str:
+        if not value:
+            raise SOAPFaultException(
+                CONFIRM_APPLICATION_DELIVERY_NO_TRACKING_NUMBER_ERR,
+                fault=OpportunityListRequestInvalidParams,
```
freddieyebra: Can we update this to use an appropriate fault exception for this?

api/src/legacy_soap_api/grantors/schemas/confirm_application_delivery_schemas.py
```diff
@@ -0,0 +1,55 @@
...
+            "ns2:ConfirmApplicationDeliveryResponse", "ConfirmApplicationDeliveryResponse"
+        ),
+    )
+
+
+class ConfirmApplicationDeliveryResponseSOAPEnvelope(BaseSOAPSchema):
+    body: ConfirmApplicationDeliveryResponseSOAPBody = Field(alias="Body")
+
+    def to_soap_envelope_dict(self, operation_name: str) -> dict:
+        return {"Envelope": self.model_dump(by_alias=True)}
```
freddieyebra: Could we remove this class since we already have `to_soap_envelope_dict` method in the BaseSOAPSchema?

api/src/legacy_soap_api/legacy_soap_api_config.py
freddieyebra: I agree with Michael. We dont need mtom here. The following logic is incorrect. Whether we are returning mtom format or not should not impact whether we return proxy response or not.
```python
if self.operation_config.is_mtom is False:
   return proxy_response```

File: api/tests/src/legacy_soap_api/grantors/schemas/test_confirm_application_delivery_schema.py [hunk 7]
```diff
@@ -0,0 +1,276 @@
...
+    get_soap_operation_dict,
+)
+
+GRANTS_GOV_TRACKING_NUMBER = "GRANT80000000"
+BOUNDARY_UUID = "9b15c243-6601-47b3-8f90-3877f0646f7d"
+RESPONSE_MESSAGE = "Success"
+
+
+class TestLegacySoapGrantorConfirmApplicationRequestSchema:
+    def test_can_convert_confirm_application_delivery_request_soap_payload_dict(self, db_session):
```
freddieyebra: we should not need db_session param in this test. there are also other tests that have this param and do not need it. please remove if not needed.

api/src/legacy_soap_api/grantors/schemas/confirm_application_delivery_schemas.py
jakobpederson: I can swap in a `super()` but I do need to modify the `operation_name` to make sure the right prefix gets added to the dict.

api/src/legacy_soap_api/legacy_soap_api_config.py
jakobpederson: I'll pull that check, and since that's the only place it's used I could pull the entire variable. I'm confused on what MTOM means for our responses. Is the MTOM stuff specifically about the file data?

freddieyebra:
> > > > @chouinar I think the `mtom` thing is a naming issue. When `is_mtom` is true, it constructs the xml message
> > > > ```
> > > >         # 1. --uuid: {boundary_uuid}\n
> > > >         # 2. headers:
> > > >         #    'Content-Type: application/xop+xml; charset=UTF-8; type="text/xml"\n'
> > > >         #    "Content-Transfer-Encoding: binary\n"
> > > >         #    "Content-ID: <root.message@cxf.apache.org>\n\n"
> > > >         # 3. MTOM xml body
> > > >         # 4. --uuid: {boundary_uuid}
> > > >         # 5. the file bytes from the file being attached
> > > >         # 6. --uuid: {boundary_uuid}--
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > And if there is an `mtom_file_stream` it injects them. So if `is_mtom` is false, just return the proxy, if it's true construct our own xml.
> > > > I think this is correct behavior though maybe it's more like `build_xml_response` would be a better name.
> > > 
> > > 
> > > Please correct me if my understanding is wrong. There are two types of responses we can form:
> > > 
> > > * A regular XML response - for any endpoint that doesn't need to return files. This doesn't have to have all of the boundary and other stuff in it. By default, this should be the behavior of most endpoints
> > > * mtom - Which has all the boundary stuff because of how it structures file responses
> > > 
> > > This endpoint wants the ordinary case.
> > > As written, the way `is_mtom` works is that if it's true, it does the latter. But if it's false for a grantor endpoint, it returns the proxy response - that is, the response from grants.gov.
> > > How do we return an ordinary response without the mtom wrapper for a grantor endpoint?
> > 
> > 
> > I may be mistaken here but I don't think there's any "regular XML responses". Every response I've seen has the MTOM boundary stuff. For example, here's a failed response from `ConfirmApplicationDelivery` through Postman
> > ```
> > --uuid:4583e02b-9320-4d05-a147-b4a00634a447
> > Content-Type: application/xop+xml; charset=UTF-8; type="text/xml"
> > Content-Transfer-Encoding: binary
> > Content-ID: <root.message@cxf.apache.org>
> > 
> > <soap:Envelope xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/"><soap:Body><soap:Fault><faultcode>soap:Server</faultcode><faultstring>Failed to confirm application delivery.(Expected an Application status of:'Validated' , but found a status of 'Received by Agency' for GRANT00000525)</faultstring></soap:Fault></soap:Body></soap:Envelope>
> > --uuid:4583e02b-9320-4d05-a147-b4a00634a447--
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So we can return proxy response, which is also MTOM boundaried, or we response with an MTOM boundaried message.
> 
> I'm missing something in the implementation.
> 
> We have [get_simpler_soap_response](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/legacy_soap_api/legacy_soap_api_client.py#L164) defined as a base function. This is overriden for the grantor client, but not for the applicant. These two functions have two fundamentally different approaches. The base method converts the response data to XML. The grantor approach converts the data into an mtom file stream but I don't see where it calls any conversion of dicts to XML.
> 
> I'm really lost on how these work. I had been up to this point interpreting the MTOM concept as separate from XML (yes, they both have some UUID boundaries, but all the file stuff is what separates them).
> 
> This isn't just a config question, I'm also just not getting how a response even gets built in either case. In the applicant endpoint, where does the UUID come from? In the grantor case where does the XML go? Am I missing something here?

@chouinar these are valid concerns and correct. there is some functionality that we will need to revisit and refactor. there has been some divergence in functionality across the applicants and grantors client that we will need to get back aligned. The scope of this ticket is just defining schemas and having unit tests for the new operation so i will not consider these concerns as blockers for this ticket. all of what you mentioned does need to be considered and addressed asap before this and other operations start getting traffic.

Issue #8412, opened 2026-02-06
Eslint Upgrade and Updates for Next 16 Upgrade


doug-s-nava:
### Summary

Depends on https://github.com/HHS/simpler-grants-gov/pull/8404 and https://github.com/HHS/simpler-grants-gov/pull/8407

Upgrading to Next 16 requires updating eslint tooling:
* Next 16 requires update to eslint-config-next package
  * which requires updating typescript-eslint/parser and other eslint related packages
    * which requires lots of linting fixes

### Acceptance criteria

- [ ] all linting and type checks pass

doug-s-nava:
make sure the skipped test linting rule is successfully applied after these changes

PR #8694, opened 2026-02-26
[Issue #8412] eslint updates for next 16 upgrade

doug-s-nava:
## Summary

Fixes for #8412

## Changes proposed

The Next 16 upgrade requires a major version bump to the eslint package, which has some downstream affects that we need to work out

* moves eslint configuration into new eslint.config.js format
* removes some unused `eslint-disable` annotations
* adds some necessary `eslint-disable` annotations
* some test refactors to fix linting errors

## Validation steps

1. _VERIFY_: all linting passes in CI
2. make some updates in the code locally that should break linting rules on this branch
3. _VERIFY_: expected linting errors appear when running `npm run lint`


frontend/src/app/[locale]/(base)/award-recommendation/page.test.tsx
```diff
@@ -218,14 +219,12 @@ describe("AwardRecommendationPage", () => {
...
     it("redirects to /maintenance", async () => {
-      try {
-        await AwardRecommendationPage({
+      await wrapForExpectedError(() => {
+        return AwardRecommendationPage({
           params: localeParams,
         });
-        throw new Error("Expected redirect to throw");
-      } catch (error) {
-        expect(mockRedirect).toHaveBeenCalledWith("/maintenance");
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/award-recommendation/page.test.tsx:227]: this change is to work around an eslint error prohibiting expects in catch blocks

frontend/eslint.config.js
```diff
@@ -0,0 +1,123 @@
+const { defineConfig } = require("eslint/config");
```
doug-s-nava, line 1: the main change necessitated by the version bump is moving from the previous `.eslintrc.js` to this new `eslint.config.js` file. See https://eslint.org/docs/latest/use/configure/migration-guide

frontend/package.json
```diff
@@ -13,7 +13,7 @@
     "debug": "NODE_OPTIONS='--inspect' next dev",
     "format": "prettier --write '**/*.{js,json,md,ts,tsx,scss,yaml,yml}'",
     "format-check": "prettier --check '**/*.{js,json,md,ts,tsx,scss,yaml,yml}'",
-    "lint": "next lint",
+    "lint": "eslint",
```
doug-s-nava, line 16: `next lint` is deprecated in next 16

frontend/src/components/applyForm/widgets/budget/Budget424aSectionB.tsx
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/no-unsafe-assignment */
```
doug-s-nava, line 1: note that the removal of file level disables like this is not because the linter is no longer checking, it's because the new linter was able to notice that there were no violations in the file so the disable was not doing anything

Issue #8413, opened 2026-02-06
Perform Next Upgrade to v16


doug-s-nava:
### Summary

Depends on eslint upgrade ticket (which in turn depends on storybook and testing update tickets)

Will include updates related to Next 16s move from webpack to turbopack, largely affecting how style compilation and imports work.

### Acceptance criteria

- [ ] all checks pass
- [ ] all application functionality works as expected when running on Next 16

PR #8701, opened 2026-02-27
[Issue #8413] Next 16 actual upgrade

doug-s-nava:
## Summary

Work for #8413 

## Changes proposed

Upgrades next from version 15 to version 16

## Context for reviewers

Some breaking changes in v16 that we are dealing with here:

* change from middleware file to proxy file
* change from webpack to turbopack for bundling
  * requires updates to our sass compilation, as turbopack doesn't accept JS functions 
* removal of `--no-lint` flag on next build as lint is no longer called by default

## Validation steps

1. _VERIFY_: ci passes
2. _VERIFY_: general app manual regression tests pass

### DIFF
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -49,7 +49,7 @@ ENV NEXT_TELEMETRY_DISABLED 1
 ENV NEXT_BUILD true
 
 # Skip lint because it should have happened in the CI already
-RUN npm run build -- --no-lint
+RUN npm run build
 
 # Run the Next.js server
 # =====================================


frontend/tsconfig.json
```diff
@@ -12,7 +12,7 @@
     "moduleResolution": "bundler",
     "resolveJsonModule": true,
     "isolatedModules": true,
-    "jsx": "preserve",
+    "jsx": "react-jsx",
```
doug-s-nava, line 15: I believe Next updates this automatically - react-jsx is the "preferred" value as it lets react optimize jsx compilation, and it doesn't seem to break anything

frontend/package.json
```diff
@@ -108,11 +108,14 @@
...
     "style-loader": "^3.3.2",
-    "typescript": "^5.0.0",
+    "typescript": "^5.9.2",
     "vite": "^7.0.0"
   },
   "overrides": {
     "@types/react": "19.2.7",
-    "@types/react-dom": "19.2.3"
+    "@types/react-dom": "19.2.3",
+    "next-navigation-guard": {
```
doug-s-nava, line 117: is pinned at v15, but all reports suggest it works fine on 16 as well https://github.com/LayerXcom/next-navigation-guard/issues/50

Issue #8393, opened 2026-02-05
Set the correct user in the workflow audit history


chouinar:
### Summary

Let's assume a workflow has the follow steps `A -> B -> C`.  If a user does an event for the A->B transition, and then B->C automatically happens afterwards, our audit history would show them as doing both. We'd prefer any of these automated steps to instead be done by our workflow system user so that it's clearer to a user looking at the audit history that they weren't the one sending an email or whatever other internal processes happened.

For this, we'll need to track how many transitions we do, and if we're past the first one, instead swap the acting user out for system user when we write the audit record.

### Acceptance criteria

- [ ] If multiple steps occur automatically in a workflow, the system user is the "acting user" for any later ones

ychoy:
@hao10282025-sudo @kkrug if you finish the SOAP Proxy work sooner, please pick up this work next (in the next sprint).

kkrug:
@ychoy I'll probably have capacity this sprint once I finish up 8102 (which is the one blocking this).

PR #8706, opened 2026-02-27
[Issue #8393] Set the correct user in the workflow audit history

kkrug:
## Summary

Fixes for #8393  

## Changes proposed

- ~Updated workflow_service.py to add function for getting the system workflow user~ Used env variable instead
- ~Updated state_machine_event.py to track the number of transitions that have occurred to determine if we are in an automatic transition~ moved transition counter internally to audit listener
- Updated workflow_audit_listener.py to increment and check the transition counter so that it can log the appropriate user in the audit history
- Updated test_workflow_audit_listener.py to add a test to confirm that automatic transitions (via 'after' parameter) use the system user
- Updated conftest.py to add `autouse=True` to the workflow_user fixture

## Context for reviewers

Let's assume a workflow has the follow steps A -> B -> C. If a user does an event for the A->B transition, and then B->C automatically happens afterwards, our audit history would show them as doing both. We'd prefer any of these automated steps to instead be done by our workflow system user so that it's clearer to a user looking at the audit history that they weren't the one sending an email or whatever other internal processes happened.

## Validation steps

For now, we can only confirm that the tests pass.

### DIFF
--- a/api/src/workflow/listener/workflow_audit_listener.py
+++ b/api/src/workflow/listener/workflow_audit_listener.py
@@ -45,32 +45,24 @@ class WorkflowAuditListener:
         # Determine which user to use for this transition
         # If this is the first transition (transition_count == 0), use the original acting user
         # Otherwise, use the system user ID for automatic transitions
+        audit_kwargs = {
+            "workflow": state_machine_event.workflow,
+            "transition_event": event_data.event.name,
+            "source_state": event_data.source.value,
+            "target_state": event_data.target.value,
+            "event": state_machine_event.workflow_history_event,
+            "audit_metadata": state_machine_event.metadata,
+        }
         if self.transition_count == 0:
             # First transition - use the actual user who triggered the workflow
-            audit_kwargs = {
-                "workflow": state_machine_event.workflow,
-                "acting_user": state_machine_event.acting_user,
-                "transition_event": event_data.event.name,
-                "source_state": event_data.source.value,
-                "target_state": event_data.target.value,
-                "event": state_machine_event.workflow_history_event,
-                "audit_metadata": state_machine_event.metadata,
-            }
+            audit_kwargs["acting_user"] = state_machine_event.acting_user
             acting_user_id = state_machine_event.acting_user.user_id
         else:
             # Subsequent automatic transitions - use the system user ID from config
             # This avoids a DB query and will fail at commit time if the user doesn't exist
             config = WorkflowServiceConfig()
-            audit_kwargs = {
-                "workflow": state_machine_event.workflow,
-                "acting_user_id": config.workflow_service_internal_user_id,
-                "transition_event": event_data.event.name,
-                "source_state": event_data.source.value,
-                "target_state": event_data.target.value,
-                "event": state_machine_event.workflow_history_event,
-                "audit_metadata": state_machine_event.metadata,
-            }
             acting_user_id = config.workflow_service_internal_user_id
+            audit_kwargs["acting_user_id"] = acting_user_id
 
         # Increment the transition count for subsequent transitions
         self.transition_count += 1


api/src/workflow/service/workflow_service.py
```diff
@@ -108,3 +109,16 @@ def get_and_validate_workflow(
         raise InactiveWorkflowError("Workflow is not active - cannot receive events")
 
     return workflow
+
+
+def get_system_workflow_user(db_session: db.Session) -> User:
+    """Get the system workflow user for automated workflow actions."""
+    system_user = db_session.scalar(
+        select(User).where(User.user_type == UserType.INTERNAL_SYSTEM_USER)
+    )
```
chouinar: For the internal user, don't use the user type - just in case we add more than one system user.

There's an env variable with the user ID set https://github.com/HHS/simpler-grants-gov/blob/main/api/src/workflow/config/workflow_service_config.py - use that to fetch the user.

Although I might actually suggest in this case we don't fetch the user and just do `acting_user_id=<that env var>`. That way we don't even go to the DB. This means it would fail when trying to flush/commit to the DB, but that'd effectively be the same behavior as this (the whole transaction fails catastrophically). And that'd only happen if an env is misconfigured which after we get it working once, would only be possible if someone manually deletes the user from the DB I think.

api/tests/workflow/listener/test_workflow_audit_listener.py
```diff
@@ -244,3 +244,82 @@ def test_workflow_audit_different_users(db_session, enable_factory_create):
     assert len(audit_records) == 2
     assert audit_records[0].acting_user_id == user1.user_id
     assert audit_records[1].acting_user_id == user2.user_id
+
+
+def test_workflow_audit_automatic_transitions_use_system_user(db_session, enable_factory_create):
+    """Test that automatic transitions (via 'after' parameter) use the system user."""
+    # Create a system user for the test
+    system_user = UserFactory.create(user_type=UserType.INTERNAL_SYSTEM_USER)
```
chouinar: Don't create the system user - I made a fixture for this user to be setup how they actually are. https://github.com/HHS/simpler-grants-gov/blob/main/api/tests/workflow/conftest.py#L19

We should probably setup that fixture to have `autouse=True` so it always gets setup for tests / you don't need to pull that into every test we do.

api/src/workflow/event/state_machine_event.py
```diff
@@ -43,6 +43,12 @@ class StateMachineEvent:
     # Metadata for processing an event.
     metadata: dict | None = None
 
+    # Track the number of transitions that have occurred for this event.
+    # This is used to determine if we're in an automatic transition (e.g., via 'after' parameter)
+    # versus the initial user-triggered transition. Mutable so it can be incremented.
+    # Value of 0 = first transition (user-initiated), >0 = automatic transitions
+    transition_count: int = dataclasses.field(default=0)
```
chouinar: Another way that we can keep internal to the audit listener.

Add the count to the audit listener. In the init do:
```py
self.count = 0
```

Increment it just the same as this. We reuse the same audit listener for the entire set of transitions from a single event (it's created once in the event handler before sending an event). Did a quick test, should be all the same.

api/src/workflow/service/workflow_service.py
kkrug: Oh yeah that's definitely better - updated

api/tests/workflow/listener/test_workflow_audit_listener.py
kkrug: Fixed

api/src/workflow/event/state_machine_event.py
kkrug: Updated

api/src/workflow/listener/workflow_audit_listener.py
```diff
@@ -17,30 +18,73 @@ class WorkflowAuditListener:
...
+                "transition_event": event_data.event.name,
+                "source_state": event_data.source.value,
+                "target_state": event_data.target.value,
+                "event": state_machine_event.workflow_history_event,
+                "audit_metadata": state_machine_event.metadata,
+            }
+            acting_user_id = config.workflow_service_internal_user_id
+
+        # Increment the transition count for subsequent transitions
+        self.transition_count += 1
```
chouinar, line 68: Most of these are identical, pull the kwargs out of the if-statements and only have the if-statements act on what is different.

```suggestion
        # Otherwise, use the system user ID for automatic transitions
        
           audit_kwargs = {
                "workflow": state_machine_event.workflow,
                "transition_event": event_data.event.name,
                "source_state": event_data.source.value,
                "target_state": event_data.target.value,
                "event": state_machine_event.workflow_history_event,
                "audit_metadata": state_machine_event.metadata,
            }
        
        if self.transition_count == 0:
            # First transition - use the actual user who triggered the workflow
            audit_kwargs["acting_user"] = state_machine_event.acting_user
            acting_user_id = state_machine_event.acting_user.user_id
        else:
            # Subsequent automatic transitions - use the system user ID from config
            # This avoids a DB query and will fail at commit time if the user doesn't exist
            config = WorkflowServiceConfig()
            acting_user_id = config.workflow_service_internal_user_id
            audit_kwards["acting_user_id"] = acting_user_id

        # Increment the transition count for subsequent transitions
        self.transition_count += 1
```

kkrug: Fixed

PR #8721, opened 2026-03-02
[Issue #7668] Refactor duplicate get_organization_and_verify_access implementations

### LINKED ISSUE #7668: Refactor duplicate get_organization_and_verify_access implementations
### Summary

Currently two functions with identical names but different behaviors exist in the organizations service layer, creating confusion about which privilege level is being checked (VIEW_ORG_MEMBERSHIP vs MANAGE_ORG_MEMBERS). This makes the codebase error-prone as someone might import the wrong function for their use case. Consolidate these use cases.

### Technical Notes

We may want to pass in the privileges to check or do the privilege check in the caller.

Current state:
- get_organization.py:41 - checks VIEW_ORG_MEMBERSHIP, eager loads sam_gov_entity
- list_organization_invitations.py:114 - checks MANAGE_ORG_MEMBERS, no eager loading
- 7 files import one implementation or the other

Affected imports:
- list_organization_users.py (uses VIEW)
- list_organization_roles.py (uses VIEW)
- organization_routes.py (uses VIEW)
- list_organization_invitations.py (uses MANAGE)
- list_legacy_users.py (uses MANAGE)
- ignore_legacy_user_organization.py (uses MANAGE)

### Acceptance criteria

- Consolidate the two functions. No duplicate function names.
- Update tests

mikehgrantsgov:
## Summary

Fixes #7668

## Changes proposed

- Added `privileges` parameter to `get_organization_and_verify_access` in `get_organization.py` (defaults to `{Privilege.VIEW_ORG_MEMBERSHIP}`)
- Removed the duplicate `get_organization_and_verify_access` from `list_organization_invitations.py`
- Updated `list_legacy_users.py` and `ignore_legacy_user_organization.py` to import from `get_organization.py` and pass `{Privilege.MANAGE_ORG_MEMBERS}`
- Fixed a bug in `list_organization_roles.py` where the access check call was malformed
- Added unit tests for the unified function covering both privilege levels

## Context for reviewers

Currently two functions with identical names but different behaviors exist in the organizations service layer, creating confusion about which privilege level is being checked (VIEW_ORG_MEMBERSHIP vs MANAGE_ORG_MEMBERS). This makes the codebase error-prone as someone might import the wrong function for their use case. Consolidate these use cases.

## Validation steps

See new tests to `tests/src/services/organizations_v1/test_get_organization.py`.

### DIFF
--- a/api/tests/src/services/organizations_v1/test_get_organization.py
+++ b/api/tests/src/services/organizations_v1/test_get_organization.py
@@ -13,7 +13,11 @@ from tests.lib.organization_test_utils import create_user_in_org
         # VIEW is always required — user with VIEW and no extra arg succeeds
         ([Privilege.VIEW_ORG_MEMBERSHIP], None, 200),
         # Both VIEW and MANAGE required — user with both succeeds
-        ([Privilege.VIEW_ORG_MEMBERSHIP, Privilege.MANAGE_ORG_MEMBERS], {Privilege.MANAGE_ORG_MEMBERS}, 200),
+        (
+            [Privilege.VIEW_ORG_MEMBERSHIP, Privilege.MANAGE_ORG_MEMBERS],
+            {Privilege.MANAGE_ORG_MEMBERS},
+            200,
+        ),
         # MANAGE-only user fails the base VIEW check
         ([Privilege.MANAGE_ORG_MEMBERS], None, 403),
         # VIEW-only user fails the additional MANAGE check
--- a/api/tests/src/api/organizations_v1/test_ignore_legacy_user_organization.py
+++ b/api/tests/src/api/organizations_v1/test_ignore_legacy_user_organization.py
@@ -14,7 +14,9 @@ class TestIgnoreLegacyUserOrganization:
 
     @pytest.fixture
     def user_org_a(self, db_session):
-        user, org, token = create_user_in_org(db_session, privileges=[Privilege.VIEW_ORG_MEMBERSHIP, Privilege.MANAGE_ORG_MEMBERS])
+        user, org, token = create_user_in_org(
+            db_session, privileges=[Privilege.VIEW_ORG_MEMBERSHIP, Privilege.MANAGE_ORG_MEMBERS]
+        )
         return user, org, token
 
     def test_ignore_legacy_user_organization_200(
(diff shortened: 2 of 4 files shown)


api/src/services/organizations_v1/get_organization.py
```diff
@@ -39,29 +39,36 @@ def get_organization(db_session: db.Session, organization_id: uuid.UUID) -> Orga
...
+        privileges: Set of privileges required. Defaults to {Privilege.VIEW_ORG_MEMBERSHIP}.
 
     Returns:
         Organization: The organization with SAM.gov entity data loaded
 
     Raises:
         FlaskError: 404 if organization not found, 403 if access denied
     """
+    if privileges is None:
+        privileges = {Privilege.VIEW_ORG_MEMBERSHIP}
```
chouinar: I'd say that for this - we want to leave this privilege alone. This should always be a requirement of this function.

I'd say we should do:

```py
    check_user_access(
        db_session,
        user,
        {Privilege.VIEW_ORG_MEMBERSHIP},
        organization,
    )

    if privileges is not None:
        check_user_access(
          db_session,
          user,
          privileges,
          organization,
       )
```

Alternatively, we make it so the check function can do an AND for privileges (it's an OR right now if multiple provided). Then we'd only need to call the function once.

api/tests/src/services/organizations_v1/test_get_organization.py
```diff
@@ -0,0 +1,53 @@
...
+)
+
+
+def _make_user_in_org(privileges):
+    user = UserFactory.create()
+    org = OrganizationFactory.create()
+    role = RoleFactory.create(privileges=privileges, is_org_role=True)
+    org_user = OrganizationUserFactory.create(user=user, organization=org)
+    OrganizationUserRoleFactory.create(organization_user=org_user, role=role)
+    return user, org
```
chouinar: Could we use the `create_user_in_org` function? This is mostly that, but just doesn't add a JWT - for agencies we just made a few functions that returned JWTs/didn't but all called the same underlying function.

mikehgrantsgov: Makes sense, updated this

api/src/services/organizations_v1/get_organization.py
mikehgrantsgov: Updated this function to account for that. Thanks

PR #8628, opened 2026-02-24
[Issue #8470] - Opportunity section for AR protoype

chay-REIsys:
## Summary

Fixes / Work for #8470  

## Changes proposed

To the before award recommendation shell page Opportunity section is added where we will pull the specific opportunity details
Updated unit tests

## Context for reviewers

In award recommendation shell page there is a specific opportunity section in which we will show some opportunity details
As of now we are building a shell page we will fetch id from query params and show the details.
Associated unit tests have been updated

## Validation steps

- fetch changes, do npm run build
- awardRecommendationOff set to false
- navigate to http://localhost:3000/award-recommendation?id={oppurtunityid} (fetch opportunity id from local db.
- verify tests are passing using npm run test
- if we try to fetch with id which doesn't exist opportunity details will not be populated 
- As of now selection method is mocked and won't change for any opportunity id this will be implemented in the following tickets for AR page
- change oppurtunityid in query param and see the changes get reflected on UI too

image
 

image

Issue #8470, opened 2026-02-10
Mockup Opportunity section for AR protoype


kevlyu:
### Summary

Mockup the frontend for the Opportunity section with Opportunity ID, Opportunity Title, Fiscal Year, and Opportunity Summary.

### Acceptance criteria

tbd

chaitanya-duddella:
as of now  opportunity id is hardcoded but will be dynamically rendered once get award recommendation is in place 
i cretaed separate ticket for this wire up

Ticket: https://github.com/HHS/simpler-grants-gov/issues/8789

@kevlyu @nathan-stricker @doug-s-nava  @slowbot

chaitanya-duddella:
https://staging.simpler.grants.gov/award-recommendation/b9c15d13-8f13-4e15-80b8-3cf5acf8481c/edit

Issue #8374, opened 2026-02-04
Create Shell Document for Award Recommendation


kevlyu:
### Description

Create the overall container with all the necessary components for AR

### Risks and assumptions

n/a

PR #8751, opened 2026-03-02
[Issue #8470] - PR Feedback for 8470

chay-REIsys:
## Summary

Fixes for #8470  

## Changes proposed

PR Feedback for #8470 
https://github.com/HHS/simpler-grants-gov/pull/8628#pullrequestreview-3856854141
## Context for reviewers

Pr for 8470 is already merged but there is additional feedback which is accounted in this PR

1.  use of existing oppurtunity description display 
2. Show a message to user in case of organization fetch failed
3. Show No summary Available in case organization doesn't have any summary

## Validation steps

- fetch changes, do npm run build
- awardRecommendationOff set to false
- navigate to http://localhost:3000/award-recommendation/{id} (as of now any guid is fine for id) once get award recommendation is wired up we can see this dynamically loaded
- verify tests are passing using npm run test


frontend/src/components/opportunity/OpportunityDescription.tsx
```diff
@@ -74,9 +75,18 @@ const SummaryDescriptionDisplay = ({
 const OpportunityDescription = ({
   summary,
   attachments,
+  summaryOnly = false,
 }: OpportunityDescriptionProps) => {
   const t = useTranslations("OpportunityListing.description");
 
+  if (summaryOnly) {
```
doug-s-nava: why not just use the `<SummaryDescriptionDisplay>` on its own?

frontend/src/app/[locale]/(base)/award-recommendation/page.tsx
```diff
@@ -107,8 +47,8 @@ const OpportunitySectionComponent = async ({
   const fundingOppName =
     opportunityData.opportunity_title || "Funding Opportunity";
   const fundingOppNumber = opportunityData.opportunity_number || "--";
-  const summary =
-    opportunityData.summary.summary_description || "No summary available";
+  const summaryDescription = opportunityData.summary.summary_description || "";
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/award-recommendation/page.tsx]: this still doesn't work for an opportunity that doesn't have a summary, since it will hit `opportunityData.summary` as undefined, then try to reference `summary_description` of undefined. Let me know if you need help with testing. My previous suggestion should work

frontend/src/components/opportunity/OpportunityDescription.tsx
chaitanya-duddella: updated

frontend/src/app/[locale]/(base)/award-recommendation/page.tsx
Comment (chaitanya-duddella) [reply frontend/src/app/[locale]/(base)/award-recommendation/page.tsx]: added optional chaining on summary

chaitanya-duddella:
> Nathan's PR moved the ID into the url path, but the code is still looking for it on the search params, so the opportunity section is not loading

@doug-s-nava  the id which nathan added  is for award recommendation id but i need one for opportunityid and i believe that should be a query param right? please advise

doug-s-nava:
@chay-REIsys ah, good call. You shouldn't need that query param - the award recommendation should have the opportunity id attached, so you can chain the calls

chaitanya-duddella:
> @chay-REIsys ah, good call. You shouldn't need that query param - the award recommendation should have the opportunity id attached, so you can chain the calls

@doug-s-nava  as of now no award recommendation tables are setup so i will have it for now and update it later will add a ticket for this

frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.test.tsx
```diff
@@ -71,7 +95,7 @@ jest.mock("src/components/award-recommendation/AwardRecommendationHero", () => {
 });
 
 const mockOpportunityDetail = {
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.test.tsx:77]: can you use an existing fixture and make updates as necessary? rather not recreate a full fixture from scratch if we don't need to

frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.test.tsx
```diff
@@ -62,6 +62,30 @@ jest.mock("src/services/fetch/fetchers/awardRecommendationFetcher", () => ({
   }),
 }));
 
+jest.mock("next-intl", () => ({
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.test.tsx:65]: this can be removed

frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.test.tsx
```diff
@@ -11,7 +11,7 @@ import { FunctionComponent, ReactNode } from "react";
 type onEnabled = (props: LocalizedPageProps) => ReactNode;
 
 jest.mock("next-intl/server", () => ({
-  getTranslations: () => identity,
+  getTranslations: () => Promise.resolve(identity),
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.test.tsx]: why this change? seems to pass without it

frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.test.tsx
```diff
@@ -62,6 +62,30 @@ jest.mock("src/services/fetch/fetchers/awardRecommendationFetcher", () => ({
   }),
 }));
 
+jest.mock("next-intl", () => ({
+  useTranslations: () => identity,
+}));
+
+jest.mock("next/link", () => ({
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.test.tsx]: why the need to mock this? I don't recall ever needing to mock next/link in other tests

frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.test.tsx
```diff
@@ -62,6 +62,30 @@ jest.mock("src/services/fetch/fetchers/awardRecommendationFetcher", () => ({
...
+  default: ({
+    children,
+    href,
+  }: {
+    children: React.ReactNode;
+    href: string;
+  }) => <a href={href}>{children}</a>,
+}));
+
+jest.mock("src/components/opportunity/OpportunityDescription", () => ({
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.test.tsx]: is this being mocked in order to more fully wall off the behavior of the page? That hasn't been our approach in the past and will make things very complex as we add more and more functionality here. What would we lose if we didn't mock this component?

frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.test.tsx
```diff
@@ -168,29 +196,18 @@ describe("AwardRecommendationPage", () => {
...
-        searchParams: Promise.resolve({}),
-      });
-      render(component);
-
-      expect(opportunityFetcher.getOpportunityDetails).not.toHaveBeenCalled();
-      expect(await screen.findByText("pageTitle")).toBeVisible();
-    });
-
-    it("calls getOpportunityDetails when id search param is provided", async () => {
+    it("calls getOpportunityDetails after award recommendation call happens", async () => {
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.test.tsx]: this should and will be the case down the line, but right now the two calls are independent, and the order of calls is not being tested. Still a useful test but we should rename it something like "calls getOpportunityDetails with expected id"

frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.test.tsx
```diff
@@ -11,7 +11,7 @@ import { FunctionComponent, ReactNode } from "react";
 type onEnabled = (props: LocalizedPageProps) => ReactNode;
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.test.tsx:11]: can we add two more tests:
* one to make sure that the proper opportunity details appear on the page
* one to test the case where the summary is not present on the opportunity

frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.tsx
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.tsx:135]: I'm ok with this for now, but in an upcoming ticket we will need to:
* make the call for award recommendation details at the page level rather than in the hero component
* chain the call for opportunity details off of the result of the award recommendation call

frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.test.tsx
Comment (chaitanya-duddella) [reply frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.test.tsx:77]: there is no existing fixture for opportunityApiResponse it was mocked as below in other file
jest.mock("src/services/fetch/fetchers/opportunityFetcher", () => ({
  getOpportunityDetails: () => opportunity() as Promise<OpportunityApiResponse>,
}));

frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.test.tsx
Comment (chaitanya-duddella) [reply frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.test.tsx]: intially some tests were failing so i was trying to mock each and every component but eventually i forgot to remove it.. thank you

frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.test.tsx
Comment (chaitanya-duddella) [reply frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.test.tsx]: we are using  SummaryDescriptionDisplay from the OpportunityDescription File Not properly mocking this causing not rendering component correctly in this case if we not mock this it is rendering empty div within a body causing test failures for award-recommendation Hero and Opportunity Section not loaded
image

@doug-s-nava

frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.test.tsx
Comment (chaitanya-duddella) [reply frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.test.tsx:65]: removed

frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.test.tsx
Comment (chaitanya-duddella) [reply frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.test.tsx]: reverted

frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.test.tsx
Comment (chaitanya-duddella) [reply frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.test.tsx:11]: added

frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.tsx
Comment (chaitanya-duddella) [reply frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.tsx:135]: will add a ticket for this

Comment (chaitanya-duddella) [reply frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.tsx:135]: added
https://github.com/HHS/simpler-grants-gov/issues/8374?issue=HHS%7Csimpler-grants-gov%7C8789

frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.test.tsx
Comment (doug-s-nava) [reply frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.test.tsx:77]: gotcha - we should create a fixture for this in the future

frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.test.tsx
Comment (doug-s-nava) [reply frontend/src/app/[locale]/(base)/award-recommendation/[id]/page.test.tsx]: tests pass fine for me locally with this mock removed...

doug-s-nava (approved):
✅ ✅ ✅ ✅ 

thank you for bearing with this drawn out review!

Issue #7027, opened 2025-11-10
FieldList Update getFieldConfig logic 


doug-s-nava:
### Summary

Building on the refactor, update the form generation code as described in the tech spec at https://navasage.atlassian.net/wiki/x/AgCqgg to incorporate the FieldList (nee ExpandableGroup) widget.

Changes include:

* update addPrintWidgetToFields to handle the “fieldList” type. This should mirror the handling for “section” closely
* Will likely need a new branch of getFieldConfig to handle the “fieldList” type. This should:
  * Build the props for the the FieldList component
    * groupDefinition: iterate over the children definitions in referenced form schema to determine widgets and props using existing functionality which should not need to change
    * baseId: use existing functionality as much as possible to determine the id prop for each widget, but will need a new function that knows how to set a placeholder for the array index at the right point in the string.
      * ex: normal functionality will produce something like `top_field--expandable_group--grouped_field or maybe top_field--expandable_group[0]--grouped_field and we’ll need to output top_field--expandable_group[~~index~~]--grouped_field` or something like that, so we can swap in the correct index for each entry
    * defaultSize: pull directly from UI schema
    * rawErrors:  depending on how we build out the error mapping logic within the FieldList logic, this could be all errors for the form, or we can create new functionality here to filter down to only errors related to the group. The latter is preferable.

### Acceptance criteria

- [ ] Solid unittests of every usecase.
- [ ] Solid regression of other widgets, especially in SF 424, SF 424A, SF LLL, NEH cover sheet

PR #8755, opened 2026-03-02
[Issue #7027] FieldList widget plumbing 

myduong-navapbc:
## Summary

Work for #7027 

This PR introduces initial support for a new UI schema node type, fieldList, as groundwork for the FieldList (formerly ExpandableGroup) feature.

### Changes proposed

- Adds fieldList to UI schema validation (AJV schema)
- Extends apply-form types to represent UiSchemaFieldList
- Updates utilities to treat fieldList similarly to section where applicable:
  - print widget mapping
  - warning tree traversal
- Adds a getFieldConfig branch for fieldList that builds a FieldList widget config:
  - Generates a groupDefinition describing the template fields for each list entry
  - Extracts reusable widget props into generalProps (including schema)
  - Strips instance-specific props (id, value, key)
  - Generates a baseId placeholder pattern (`[~~index~~]`) for list row instances
- Adds validation guardrails ensuring fieldList children are limited to field/multiField nodes
- Updates and extends tests to cover the new node type and ensure existing behavior remains unchanged

### Context for reviewers

Some target forms require dynamically sized arrays of grouped fields (e.g., contact lists).

The tech spec proposes introducing a fieldList UI schema node that defines a repeatable group template. This PR focuses on the plumbing required for the form system to recognize and configure this node type without affecting existing forms.

Specifically, this PR introduces the schema/type support and configuration generation needed so a future FieldList widget can render repeatable rows of fields.

### Scope

- No implementation of dynamic add/remove entry behavior yet
- No runtime index substitution logic for the `[~~index~~]` placeholder yet 
- No error-mapping/filtering logic specific to field groups yet (rawErrors currently passed through as empty)
- fieldList children are currently restricted to field/multiField nodes; nested fieldList structures are not supported at this time
- No changes to existing form JSON fixtures (SF424 / SF424A / SFLLL / NEH cover sheet) beyond ensuring they continue to validate/render

### Implementation notes

- groupDefinition represents the template for fields inside each list row
- generalProps contains reusable widget configuration derived from existing widget configs
- Instance-specific props (id, value, key) are removed so rows can generate their own values
- baseId includes a placeholder (`[~~index~~]`) that future work will replace with the actual row index

FieldList rows are generated from a template defined in `groupDefinition`.

Each entry in `groupDefinition` represents a field that should appear inside a single row of the list and contains:

- `widget`: the widget type determined using existing `getFieldConfig` logic
- `baseId`: a placeholder-based ID pattern that will be used to generate per-row IDs
- `generalProps`: reusable widget configuration (including `schema`, validation constraints, and options)

Example structure produced by `getFieldConfig`:

FieldList
 -groupDefinition
       widget: "Text"
         baseId: `"contacts[~~index~~]--firstName"`
         generalProps:
         schema: {...}
         required: false
         disabled: false
      
        widget: "AttachmentArray"
          baseId: `"contacts[~~index~~]--docs"`
          generalProps:
            schema: {...}

Instance-specific props (`id`, `value`, `key`) are intentionally stripped out so each rendered row can generate its own values using the `baseId` pattern.

Screenshot 2026-03-10 at 9 55 48 AM
Screenshot 2026-03-10 at 9 54 36 AM

### Validation steps
`npm run test`

Confirmed:

- utils.test.ts passes
- validate.test.ts passes
- getFieldConfig.test.ts passes

Manual / Smoke

Load apply form pages for:

- SF 424
- SF 424A
- SF LLL
- NEH Cover Sheet

Confirm:

- Existing widgets render normally (Select, MultiSelect, Attachment, AttachmentArray, Budget widgets)
- No console errors related to UI schema parsing or validation


frontend/src/components/applyForm/validate.ts
```diff
@@ -136,6 +142,43 @@ export const UiJsonSchema: RJSFSchema = {
...
+          type: "string",
+        },
+        description: {
+          type: "string",
+        },
+        defaultSize: { type: "integer", minimum: 0 },
+        children: {
+          type: "array",
+          items: {
+            anyOf: [
```
doug-s-nava: I think for now we can limit this to just fields as children

myduong-navapbc: I was about to send the link to you for a prelim look over! Filling in the notes. and basic testing steps

frontend/src/components/applyForm/validate.ts
```diff
@@ -136,6 +142,33 @@ export const UiJsonSchema: RJSFSchema = {
...
+          type: "string",
+        },
+        description: {
+          type: "string",
+        },
+        defaultSize: { type: "integer", minimum: 0 },
+        children: {
+          type: "array",
+          items: {
+            $ref: "#/$defs/field",
```
myduong-navapbc, line 165: When we are ready, we can add other children here

```
items: {
  anyOf: [
    {
      $ref: "#/$defs/field",
    },
    {
      $ref: "#/$defs/fieldList",
    },
    {
      $ref: "#/$defs/section",
    },
  ],
},
```

and make changes to the `FormFields.buildFormTree` & `buildWarningTree` to pass down the warnings

frontend/src/components/applyForm/utils.ts
```diff
@@ -586,6 +578,11 @@ export function addPrintWidgetToFields(uiSchema: UiSchema): UiSchema {
         ...item,
         children: addPrintWidgetToFields(item.children),
       };
+    } else if (item.type === "fieldList") {
```
myduong-navapbc, line 583: This would make a great switch statement refactor here, imo.
But I left it to match current styles

frontend/src/components/applyForm/types.ts
```diff
@@ -98,11 +99,21 @@ export interface UiSchemaSection {
   type: "section";
   label: string;
   name: string;
-  children: Array<UiSchemaField | UiSchemaSection>;
+  children: UiSchema;
   description?: string;
 }
 
-export type UiSchema = Array<UiSchemaSection | UiSchemaField>;
```
myduong-navapbc, line 105: I consolidated `UiSchemaField | UiSchemaSection | UiSchemaFieldList` into a `UiSchemaNode` _union type_ to improve readability and maintainability as additional node types are introduced.

frontend/src/components/applyForm/utils.test.ts
```diff
@@ -585,6 +585,92 @@ describe("buildWarningTree", () => {
...
+      type: "object" as const,
+      properties: {
+        name: { type: "string" as const, title: "Name" },
+      },
+    };
+
+    const errors = buildWarningTree(uiSchema, null, warnings, formSchema);
+
+    expect(errors.length).toBeGreaterThan(0);
+    expect(errors[0].message).toBe("name is required");
```
doug-s-nava, line 628: so this works without making any changes to the error handling logic, just by updating the schema definition? cool

frontend/src/components/applyForm/utils.test.ts
```diff
@@ -910,4 +996,93 @@ describe("addPrintWidgetToFields", () => {
...
+            definition: "/properties/docs",
+            schema: { title: "Docs", type: "array" },
+            widget: "PrintAttachment",
+          },
+        ],
+      },
+    ]);
+  });
+
+  it("handles nested fieldList nodes", () => {
```
doug-s-nava: this might work, but I'd rather not add this test since we're otherwise not supporting this functionality and I'd rather not give the impression that we are

frontend/src/utils/applyForm/getFieldConfig.ts
```diff
@@ -352,10 +372,28 @@ export const getFieldConfig = <V extends string | Record<string, unknown>>({
...
-}) => {
-  const { definition, type: uiSchemaFieldType } = uiFieldObject;
+}): FieldConfig => {
+  if (uiFieldObject.type === "fieldList") {
+    return {
+      type: "FieldList",
+      props: {
+        id: uiFieldObject.name,
+        key: uiFieldObject.name,
+        schema: {
```
doug-s-nava, line 463: I think schema will be nested under `generalProps` since each field in the list will share the same schema

frontend/src/utils/applyForm/getFieldConfig.ts
```diff
@@ -352,10 +372,28 @@ export const getFieldConfig = <V extends string | Record<string, unknown>>({
   errors: FormattedFormValidationWarning[] | null;
   formSchema: RJSFSchema;
   formData: object;
-  uiFieldObject: UiSchemaField;
+  uiFieldObject: UiSchemaField | UiSchemaFieldList;
   requiredField: boolean;
-}) => {
-  const { definition, type: uiSchemaFieldType } = uiFieldObject;
+}): FieldConfig => {
```
doug-s-nava, line 419: since the fieldList component is going to need to take some props that other widgets don't take (groupDefinition, defaultSize, requiredFields) do we want to add those as general optional props for all widgets, or do we want to define the set of fieldList props as their own type?

frontend/src/components/applyForm/utils.test.ts
myduong-navapbc: Yep! Trying to keep FieldList as section-like as possible (for traversal right now). Warnings still attach to  field definitions.

frontend/src/components/applyForm/utils.test.ts
myduong-navapbc: I agree, removed.

frontend/src/utils/applyForm/getFieldConfig.test.ts
```diff
@@ -10,7 +10,10 @@ import {
 } from "src/utils/applyForm/getFieldConfig";
 import { fakeValidationError } from "src/utils/testing/fixtures";
```
myduong-navapbc, line 13: This looks like a large diff, but the functional change is focused in scope: 

FieldList now carries a schema (to satisfy the WidgetRenderers contract) and introduces `groupDefinition.generalProps` to standardize how repeated child widgets are configured. 

Tests cover:
- unchanged behavior for standard fields, enums, multifields, and warning mapping, plus 
- new FieldList-specific behavior: config shape, groupDefinition construction, baseId placeholder behavior, and guard rails for invalid/nested children.

frontend/src/utils/applyForm/getFieldConfig.ts
myduong-navapbc: I added its own type to the `types.ts`

frontend/src/utils/applyForm/getFieldConfig.ts
myduong-navapbc: schema is now included inside `generalProps`.
`generalProps` contains the reusable widget configuration extracted from the child widget (schema, options, validation props, etc.), while instance-specific props (`id`, `value`, `key`) are stripped out.

Each repeated row uses `baseId` to generate its own instance id while sharing the same schema via `generalProps`.

and look something like this:

```
groupDefinition: [
  {
    widget: "Text",
    baseId: "contacts[~~index~~]--firstName",
    generalProps: {
      schema: {...},      <---- 
      required: false,
      disabled: false,
      options: {},
      rawErrors: [],
      minLength: undefined,
      maxLength: undefined
    }
  }
]
```

frontend/src/components/applyForm/types.ts
```diff
@@ -72,10 +72,38 @@ export type WidgetTypes =
   | "Budget424aSectionC"
   | "Budget424aSectionD"
   | "Budget424aSectionE"
-  | "Budget424aSectionF";
+  | "Budget424aSectionF"
+  | "FieldList";
 
 type PropertyPath = `/properties/${string}`;
 
+export type FieldListWidgetProps = {
```
freddieyebra, line 80: Can you explain why there are 2 definitions of the `FieldListWidgetProps`?

frontend/src/components/applyForm/utils.ts
```diff
@@ -175,8 +168,10 @@ export const buildWarningTree = (
     const childErrors = uiSchema.reduce<FormattedFormValidationWarning[]>(
       (errors, node) => {
         if ("children" in node) {
+          const children =
+            node.type === "fieldList" ? node.children : node.children;
```
freddieyebra: could this just be 
```suggestion
          const children = node.children;
```?

File: frontend/src/components/applyForm/types.ts [hunk 12]
```diff
@@ -77,6 +77,33 @@ export type WidgetTypes =
...
+  id: string;
+  key: string;
+
+  // Needed because WidgetRenderers expects UswdsWidgetProps-ish props,
```
doug-s-nava, line 84: Yeah, that creates some complexity. This is true on budget form section widgets too - schema gets passed and referenced in types, but is not used. How much would it take to make schema optional, or allow widgets to take a generic prop type that could make this dynamic per widget?

(note this is not a blocker at all right now but let's maybe make a note to think about it later)

frontend/src/components/applyForm/utils.ts
```diff
@@ -581,7 +584,21 @@ export function addPrintWidgetToFields(uiSchema: UiSchema): UiSchema {
     } else if (item.type === "fieldList") {
       return {
         ...item,
-        children: addPrintWidgetToFields(item.children),
+        children: item.children.map((child) => {
+          if (child.type !== "field") {
```
doug-s-nava, line 590: what does this case represent? Could be worth adding a comment

frontend/src/components/applyForm/types.ts
```diff
@@ -77,6 +77,33 @@ export type WidgetTypes =
...
+
+  label: string;
+  description?: string;
+  defaultSize: number;
+  groupDefinition: FieldListGroupItem[];
+  rawErrors?: FormattedFormValidationWarning[] | string[];
+  requiredFields?: string[];
+};
+
+export type FieldListChildWidgetTypes = Exclude<WidgetTypes, "FieldList">;
```
doug-s-nava, line 99: this is used / defined here too, can we reuse? https://github.com/HHS/simpler-grants-gov/pull/8755/changes/55c4e972395cceda9434e853892f0a4a62170d8b#diff-bfff64fdbdc69f27d362c2fb229320a65a09299b161d2cd50d7f8e8ce6ca01aeR54

frontend/src/utils/applyForm/getFieldConfig.ts
```diff
@@ -29,25 +31,37 @@ type FieldInfo<V extends BroadlyDefinedWidgetValue> = {
...
-    rawErrors?: string[] | FormattedFormValidationWarning[] | undefined;
-    value?: BroadlyDefinedWidgetValue;
-    options?: WidgetOptions;
-    label?: string;
-    description?: string;
-    defaultSize?: number;
-  };
+const FIELD_LIST_INDEX_PLACEHOLDER = "~~index~~" as const;
+
+function buildFieldListBaseId({
```
doug-s-nava: this stuff is complex enough that I think a brief jsdoc or just a single comment explaining the usage of these functions will be helpful - for example, what are the use cases where we're replacing the field list name, and when are we just adding the placeholder? Seems like this is dependent on whether the field list is nested in the schema or not?

frontend/src/utils/applyForm/getFieldConfig.ts
```diff
@@ -432,7 +486,7 @@ export const getFieldConfig = <V extends string | Record<string, unknown>>({
       : {};
 
   return {
-    type: widgetType,
+    type: widgetType as Exclude<WidgetTypes, "FieldList">,
```
doug-s-nava: This is defined elsewhere as well, can we reuse?

frontend/src/utils/applyForm/getFieldConfig.test.ts
```diff
@@ -25,474 +25,402 @@ jest.mock("src/i18n/getMessagesSync", () => ({
...
-      definition: "/properties/email",
-    };
-
-    const formSchema: RJSFSchema = {
-      type: "object",
-      properties: {
-        email: { type: "string", title: "Email" },
-      },
-    };
+  describe("getEnumOptions", () => {
```
doug-s-nava: I'm gonna be super annoying here and request not moving the tests around / moving them back, or alternatively pointing out what's new. It's hard for me to tell what's new here

doug-s-nava (commented):
This looks good! Is this the stopping point for now? It looks pretty darn close

frontend/src/components/applyForm/types.ts
myduong-navapbc: That is an oversight on my part- 
I moved it and forgot to delete it from the component. Good catch!

frontend/src/components/applyForm/utils.ts
myduong-navapbc: Definitely! 

some context:
it was me trying to narrow down types:

```
if (node.type === "fieldList") {
  // children: UiSchemaField[]
} else {
  // children: UiSchema
}
```

and in the updated version we no longer needed to narrow the focus

frontend/src/components/applyForm/types.ts
myduong-navapbc: Agreed. I added schema here to stay compatible with the current widget rendering contract, which already expects `UswdsWidgetProps-like` shapes for special widgets as well. 

[Follow-up]: revisit widget renderer typing so special widgets like FieldList and budget section widgets do not need to carry unused schema props solely to satisfy shared rendering contracts.

frontend/src/components/applyForm/utils.ts
myduong-navapbc: I added some comments:

but the line of thinking that i had was: _If something unexpected appears here, I don’t want try to transform it._

and the case that i was guarding against was this super simple example:

```
const uiSchema: UiSchema = [
  {
    type: "fieldList",
    name: "contacts",
    label: "Contacts",
    defaultSize: 1,
    children: [
      {
        type: "field",
        definition: "/properties/firstName",
      },
      {
        type: "section",   <--------------------- not a field
        name: "extra",
        label: "Extra Info",
        children: [],
      },
    ],
  },
];
```

frontend/src/components/applyForm/types.ts
myduong-navapbc: that one is, my bad ^

updated.

frontend/src/utils/applyForm/getFieldConfig.ts
myduong-navapbc: OMG Absolutely! easily one of the most complex pieces of this entire feature. Thank you for calling this out.

```
/**
 * Builds the "baseId" used by FieldList widgets for each child field.
 *
 * FieldList entries represent a repeatable array of grouped fields.
 * Instead of generating a concrete id for a specific row (e.g. `[0]`),
 * we generate a template id containing a placeholder:
 *
 *   contacts[~~index~~]--firstName
 *
 * The FieldList widget later replaces `~~index~~` with the actual
 * array index when rendering each entry.
 *
 * This function adapts the id produced by existing field logic and
 * injects the `[~~index~~]` placeholder at the correct position.
 *
 * Example transformations:
 *
 *   childId: "contacts--firstName"
 *   > "contacts[~~index~~]--firstName"
 *
 *   childId: "top_field--contacts--firstName"
 *   > "top_field--contacts[~~index~~]--firstName"
 *
 * We check for an existing `--fieldListName--` segment so we only modify
 * the correct portion of the id while preserving any parent structure.
 */
```

frontend/src/utils/applyForm/getFieldConfig.ts
myduong-navapbc: done

frontend/src/utils/applyForm/getFieldConfig.test.ts
myduong-navapbc: This one:
i worried quite a bit while doing it. since it shuffled things around- i was trying to avoid just taking things onto the end of the file. i will undo this.

frontend/src/components/applyForm/utils.ts
doug-s-nava: got it. I wonder if we want to error out before we get here on that kind of a shape? In the end that's not going to produce a useable UI

frontend/src/utils/applyForm/getFieldConfig.test.ts
```diff
@@ -496,3 +499,100 @@ describe("getBasicMultifieldInfo", () => {
     ).toEqual(expected);
   });
 });
+
+describe("fieldList", () => {
```
doug-s-nava: since this is testing `getFieldConfig` specifically, can we add that to the describe or next this under the existing `getFieldConfig` describe?

doug-s-nava: can we add testing around `buildFieldListBaseId`, and an expect in the `returns FieldList config for a fieldList node"` test for the baseId?

frontend/src/components/applyForm/utils.ts
myduong-navapbc: The `validate.ts` should catch it but this is the fallback.

myduong-navapbc: `
// fieldList children should only be `field` nodes (validated by validateUiSchema).
// This guard prevents us from applying print widget logic to unexpected node
// types if an invalid schema slips through.
`

frontend/src/utils/applyForm/getFieldConfig.test.ts
myduong-navapbc: ok, i moved it into the proper section

myduong-navapbc: added:
```
  describe("buildFieldListBaseId", () => {
    it("inserts the index placeholder when the fieldList token exists in the child id", () => {
      expect(
        buildFieldListBaseId({
          fieldListName: "contacts",
          childId: "topField--contacts--firstName",
        }),
      ).toBe("topField--contacts[~~index~~]--firstName");
    });

    it("falls back to prefixing the fieldList name when the token is not present", () => {
      expect(
        buildFieldListBaseId({
          fieldListName: "contacts",
          childId: "firstName",
        }),
      ).toBe("contacts[~~index~~]--firstName");
    });
  });
```
  
  and then we also assert it within the return of `FieldList` config for a `fieldList` node, like so:
  
```
  expect(result.props.groupDefinition[0].baseId).toBe(
        "contacts[~~index~~]--firstName",
      );
```

frontend/src/utils/applyForm/getFieldConfig.test.ts
```diff
@@ -295,6 +299,126 @@ describe("getFieldConfig", () => {
       },
     });
   });
+
+  describe("buildFieldListBaseId", () => {
```
doug-s-nava, line 303: technically not a part of the `getFieldConfig` function so should be in its own root describe block, but that's a tiny nit

PR #8758, opened 2026-03-03
fill SFLLL form

aoysimmons-GH:
## Summary

Fixes / Work for #8370 

## Changes proposed

- Implement **[test-action-utils.ts], [test-navigation-utils.ts], [test-report-utils.ts]**  to support **[fill-sfll-form-utils.ts]**

- Implement **[fill-sfll-form-utils.ts]** function. See below summary:

## Context for reviewers

 **This utility function: [fill-sfll-form-utils.ts]:**
- Clicks the "Disclosure of Lobbying" link on the application overview page
- Fills out the form with test data
- Saves it
- Verifies that the save was successful with no validation errors

### DIFF
--- a/frontend/tests/e2e/utils/test-navigation-utils.ts
+++ b/frontend/tests/e2e/utils/test-navigation-utils.ts
@@ -128,7 +128,7 @@ export async function safeHelp_safeGoto(
           contentType: "text/plain",
         });
       }
-    } catch (error) {
+    } catch (_error) {
       // Header not found or timeout - continue silently
     }
   });


frontend/tests/e2e/utils/forms/fill-sfll-form-utils.ts
```diff
@@ -0,0 +1,347 @@
...
+
+export async function fillSfllFormUtils(
+  testInfo: TestInfo,
+  page: Page,
+): Promise<void> {
+  const startTime = new Date();
+
+  try {
+    // Click the form link to open it from the application overview page
+    await page.getByRole("link", { name: ENTITY_DATA.form.name }).click();
```
doug-s-nava, line 110: if this function is specific to the SFLLL we don't need this name to by dynamic. I like the idea of a dynamic "load form" utility but it's not necessary right now

frontend/tests/e2e/utils/test-report-utils.ts
```diff
@@ -0,0 +1,260 @@
...
+
+export async function safeHelp_softStep(
+  testInfo: TestInfo,
+  label: string,
+  action: () => Promise<void>,
+): Promise<void> {
+  await executeStep(testInfo, label, action);
+}
+
+export async function safeHelp_safeStep(
```
doug-s-nava, line 156: doesn't seem like this function does anything - can we just call execute step directly?

frontend/tests/e2e/utils/test-report-utils.ts
```diff
@@ -0,0 +1,260 @@
...
+    const endTime = new Date();
+    const durationMs = endTime.getTime() - startTime.getTime();
+    await testInfo.attach(`${label} - timing`, {
+      body: `Step "${label}" completed in ${durationMs}ms`,
+      contentType: "text/plain",
+    });
+  }
+}
+
+export async function safeHelp_softStep(
```
doug-s-nava, line 148: doesn't look like this is used and also doesn't do anything, can we remove it?

frontend/tests/e2e/utils/forms/fill-sfll-form-utils.ts
```diff
@@ -0,0 +1,347 @@
...
+    await safeHelp_safeWaitForLoadState(testInfo, page);
```
doug-s-nava, line 111: don't we want the test to fail if the page never loads?

frontend/tests/e2e/utils/test-navigation-utils.ts
```diff
@@ -0,0 +1,282 @@
+// ============================================================================
```
doug-s-nava, line 1: a couple of thoughts on this file:
 
* there are a lot of unused functions in here. Someday they may be useful but it is hard to tell and I don't want to delay this review trying to make sure they are in good shape if they won't ever be used. Feel free to stash them somewhere so you can bring them back later if we need them, but please remove them from this PR
* can we stick to a function naming convention without the namespace? for example `safeNavigateToForm` instead of `safeHelp_navigateToForm`? I know that kind of namespacing is helpful in some languages but in JS it just makes things longer

frontend/tests/e2e/utils/forms/fill-sfll-form-utils.ts
```diff
@@ -0,0 +1,347 @@
...
+    title: "TitleTest",
+    phone: "9999999999",
+  },
+};
+
+// ============================================================================
+// Main Function to Fill Form
+// ============================================================================
+
+export async function fillSfllFormUtils(
```
doug-s-nava, line 102: this looks like the full test. Can you move it into a spec file?

PR #8767, opened 2026-03-03
[Issue #7063]  Fix CommonGrants OpenAPI docs

### LINKED ISSUE #7063: [SGG] Fix auto-generated OpenAPI docs for CommonGrants routes
### Summary

Currently the open API docs for the Search endpoint don't list the appropriate values for the filters in the open API documentation.  This causes the POST call to fail when trying to run it from the open API page

### Acceptance criteria

- [ ] The Search Opportunities API endpoint has been updated so that the filter values don't all say string
- [ ] The openAPI POST call to search for opportunities is able to be made successfully

jcrichlake:
## Summary

Fixes #7063 Common grants open-api endpoint not working

## Changes proposed

The open-api endpoint for common-grants search were not working due to missing example values in the marshmallow schema.  This caused the search request to populate the various keys with the value types e.g. "string" instead of an actual example string.

## Context for reviewers

The only changes in this PR that are under our control are the changes to the python file.  The workflow to make changes to the yaml is to update the Python file, then run `make openapi-spec-common-grants` to generate the updated yaml file that is included in this PR.

## Validation steps

1. Checkout branch
2. Run the following commands to start the API `# Start API with data
make init 
make run-logs
```
3. Go to `http://localhost:8080/docs` in your browswer
4. Scroll down to `Common Grants Protocol`
5. Open the POST `search` request and select `try it out`
6. Note that the request body now populates with the correct values and when clicking `Execute` the server returns a response.

### CODE REVIEW

File: api/src/api/common_grants/common_grants_schemas.py [hunk 1]
```diff
@@ -436,23 +549,32 @@ class StringComparisonFilter(Schema):
...
-        metadata={"description": "The operator to apply to the filter value"},
+        required=True,
+        metadata={
+            "description": "The operator to apply to the filter value",
+            "example": "between",
+        },
     )
     value = fields.Nested(
-        DateRange, allow_none=True, metadata={"description": "The date range value"}
+        DateRange, required=True, metadata={"description": "The date range value"}
```
jcrichlake, line 557: This change was needed in order to have the marshmallow schema evaluate to the pydantic schema

api/openapi-cg.generated.yml
widal001, line 1: Is this something we should be committing?

kkrug: I don't think this file should be committed, it's a locally generated artifact.

jcrichlake: I removed this file, sorry about that!

api/src/api/common_grants/common_grants_schemas.py
```diff
@@ -26,31 +26,50 @@
...
 
-    amount = fields.String(required=True, metadata={"description": "The amount of money"})
+    amount = fields.String(
+        required=True,
+        metadata={"description": "The amount of money", "example": "1000000.0"},
+    )
     currency = fields.String(
         required=True,
-        metadata={"description": "The ISO 4217 currency code (e.g., 'USD', 'EUR')"},
+        metadata={"description": "The ISO 4217 currency code", "example": "USD, EUR"},
```
widal001: This should just be either `"USD"` or `"EUR"` since the `currency` value should map to a single ISO 4217 currency code.

widal001: Preferably `"USD"` since that matches the default currency in the SGG responses

Issue #8020, opened 2026-01-15
Run Local E2E Tests On All Merges to Main


doug-s-nava:
### Summary

Whenever any API, frontend or infrastructure code is merged (pushed, in the parlance of Github Actions) to main, a run of the full E2E test suite against a locally spun up environment in Github Actions should be started in parallel with other CI tasks. 

Since deployment jobs are defined on a API vs frontend basis, and we want to run this on all merges that touch either piece of code, but not run duplicate runs when both services are being deployed, it is easiest for now to define this run as a new trigger on the e2e job.

To accomplish this, the trigger definition of the ci-frontend-e2e job can be updated roughly following the cd-frontend job:

```
  push:
    branches:
      - "main"
    paths:
      - "frontend/**"
      - "infra/frontend/**"
      - "infra/modules/**"
      ... etc. to include all relevant API and infra code
```

Since we'll be running this independent of merge based deploys, this will not be able to block deploys until / unless we refactor our CI/CD job tree, which is out of scope of this ticket

### Acceptance criteria

- [ ] whenever any API, frontend or infrastructure code is merged to main, the ci-frontend-e2e job is invoked once

kkrug:
@doug-s-nava Are we at the point where this is ready to be implemented?

doug-s-nava:
to avoid conflicts I'd like to wait until I'm done with this (should be very soon) https://github.com/HHS/simpler-grants-gov/pull/8446

PR #8769, opened 2026-03-03
[Issue #8020] Run Local E2E Tests On All Merges to Main

kkrug:
## Summary

Fixes for #8020  

## Changes proposed

- Updated ci-frontend-e2e.yml to add push trigger for merges to main

## Context for reviewers

Whenever any API, frontend or infrastructure code is merged (pushed, in the parlance of Github Actions) to main, a run of the full E2E test suite against a locally spun up environment in Github Actions should be started in parallel with other CI tasks.
Since deployment jobs are defined on a API vs frontend basis, and we want to run this on all merges that touch either piece of code, but not run duplicate runs when both services are being deployed, it is easiest for now to define this run as a new trigger on the e2e job.

## Validation steps

Confirm that the ci-frontend-e2e job is invoked once whenever any API, frontend or infrastructure code is merged to main.

### DIFF
--- a/.github/workflows/ci-frontend-e2e.yml
+++ b/.github/workflows/ci-frontend-e2e.yml
@@ -1,5 +1,15 @@
 name: E2E Tests (Local Github Target)
 on:
+  push:
+    branches:
+      - "main"
+    paths:
+      - "api/**"
+      - "frontend/**"
+      - "infra/api/**"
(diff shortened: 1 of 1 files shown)


kkrug:
@doug-s-nava Based on the ticket and your other work around testing, I believe this is the only change needed. Let me know if there is anything I am missing here.

Issue #8105, opened 2026-01-23
Add AuthZ to workflow event API


chouinar:
### Summary

https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2374369297/Tech+Spec+Workflow+Management+with+Python+StateMachine#AuthZ

In #8629 we created a new utility for enforcing AuthZ given a user, workflow, config, and event - all fields we should have in our event logic.

We want to add logic like so:
* If a user has the `internal_workflow_event_send` privilege, they can send an event (this is mainly so we can send events for testing purposes). This should be done by doing something like `if can_access(user, {Privilege.INTERNAL_WORKFLOW_EVENT_SEND}, None)`. If they can access, skip the next check.
* Otherwise, a user needs to be validated as being able to access sending the event with the utility we added.

This logic we added elsewhere handles determining which entities and agencies we need to fetch. For the sake of testing, it's fine to focus solely on single-opportunity workflows like our initial prototype workflow - don't worry about the more complex scenarios (we're debating adjusting that structure anyways).

### Acceptance criteria

- [ ] Internal workflow event send privilege added that lets a user through to send an event
- [ ] Otherwise we validate a user can send an event using the new utility
- [ ] Tests added for various scenarios

PR #8775, opened 2026-03-03
[Issue #8105] Add AuthZ to workflow event API

chouinar:
## Summary
Fixes / Work for #8105

## Changes proposed
* Adds AuthZ to workflow event API
* Sets up a user locally we can use to send events

## Context for reviewers
This enforces authentication for the event API. There are two ways to be authenticated:
* We have an internal user privilege that lets you send events and we won't check - this is mostly for testing and demoing.
* Having privileges for doing the approval event against the agency that owns the entity associated with the workflow

Importantly, starting a workflow is not an approval event, except for our internal user, you actually can't start a workflow via this API (and that's just allowed for testing purposes). All start workflow events will be done by other endpoints, for example to start the publish opportunity workflow, publish an opportunity and that endpoint will do it.

## Validation steps

Found an opportunity locally (doesn't matter which) and grabbed their ID. Authenticated with the User API Key Auth with the new user using the `internal_workflow_user_key` key. Can then call the event API like so.

```json
{
  "start_workflow_context": {
    "entity_id": "e43617f3-19e9-4ab2-a1ca-e6a57326090d",
    "entity_type": "opportunity",
  }
}
```

If you try with any other user, you'll get rejected. More granular testing is done in the unit tests.

### DIFF
--- a/api/tests/lib/seed_orgs_and_users.py
+++ b/api/tests/lib/seed_orgs_and_users.py
@@ -7,11 +7,7 @@ from sqlalchemy import select
 
 import src.adapters.db as db
 import tests.src.db.models.factories as factories
-from src.constants.lookup_constants import (
-    ApplicationStatus,
-    LegacyUserStatus,
-    OpportunityStatus,
-)
+from src.constants.lookup_constants import ApplicationStatus, LegacyUserStatus, OpportunityStatus
 from src.constants.static_role_values import (
     INTERNAL_WORKFLOW_USER_ROLE,
     NAVA_INTERNAL_ROLE,


api/src/api/workflows/workflow_routes.py
```diff
@@ -24,14 +24,17 @@
     security=jwt_or_user_api_key_security_schemes,
     responses=[200, 401, 403, 404, 422],
 )
-@jwt_or_user_api_key_multi_auth.login_required
+@jwt_or_api_user_key_multi_auth.login_required
```
chouinar, line 27: We have two similar, but slightly different auth headers that were built around the same time, we want to use this one. I'll make a separate PR to clean this up.

kkrug: Is the only difference between these 2 the return type?

kkrug:
Looks good, just a few unused imports in the seed files that the linter is catching.

chouinar:
> Looks good, just a few unused imports in the seed files that the linter is catching.

The joys of a new laptop, if I run the linter it sometimes fails, and sometimes succeeds without any changes.

kkrug:
> > Looks good, just a few unused imports in the seed files that the linter is catching.
> 
> The joys of a new laptop, if I run the linter it sometimes fails, and sometimes succeeds without any changes.

Huh interesting. Well if it isn't catching them all, this is what I found:

seed_data_utils - Privilege and RoleType
`from src.constants.lookup_constants import Privilege, RoleType`

seed_e2e - Privilege
`from src.constants.lookup_constants import Privilege`

seed_orgs_and_users - Privilege and RoleType
`from src.constants.lookup_constants import (
    ApplicationStatus,
    LegacyUserStatus,
    OpportunityStatus,
    Privilege,
    RoleType,
)`

chouinar:
> > > Looks good, just a few unused imports in the seed files that the linter is catching.
> > 
> > 
> > The joys of a new laptop, if I run the linter it sometimes fails, and sometimes succeeds without any changes.
> 
> Huh interesting. Well if it isn't catching them all, this is what I found:
> 
> seed_data_utils - Privilege and RoleType `from src.constants.lookup_constants import Privilege, RoleType`
> 
> seed_e2e - Privilege `from src.constants.lookup_constants import Privilege`
> 
> seed_orgs_and_users - Privilege and RoleType `from src.constants.lookup_constants import ( ApplicationStatus, LegacyUserStatus, OpportunityStatus, Privilege, RoleType, )`

I don't know what was going on exactly. The ruff command was being passed `src tests bin` and sometimes wouldn't run through the tests folder for me. I just changed it to `.` (the default anyways) - there's a single other file that this was ignoring and I fixed its issues. No idea what changed with my new laptop to cause it to get confused

api/src/api/workflows/workflow_routes.py
chouinar: The auth itself is all the same, main difference is that the way you get the user is a lot simpler on this way, the old way accounts for a lot of different scenarios and it ends up more complex than necessary.

PR #8779, opened 2026-03-03
[Issue #8333] Setup lower env agencies for users and assign them the opportunity publisher role

### LINKED ISSUE #8333: Temporary approaches to assign user roles for Opportunity Management
### Summary

**Background/Context**

**Resources**

**USER STORIES** 

**Backend Example**

**Frontend Example**

**TECHNICAL NOTES** 

### Acceptance criteria
- [ ] Create a script that runs automatically each day in dev/staging and gives a user an agency of their own, and their roles with it.

Role type association: Agency

Dao-REI:
## Summary
Fixes #8333 

## Changes proposed

- Added setup_lower_env_agencies.py - Task that creates fake Agencies, one for each user, and assigns them the opportunity_publisher role.
- Added test_setup_lower_env_agencies.py - Test the setup_lower_env_agencies.py script
- ON-HOLD: Update scheduled_jobs.tf - Schedule a job to run this script: Every day at 8:15am Eastern Time during DST. 9:15am during non-DST. (Michael: we should first run this script manually in each environment to test it.)

## Context for reviewers

Note: Fake agencies created in this script are prefixed with "AUTO". Psuedo code:
- Check that the ENVIRONMENT is local, dev or staging.
- For each user, first check if the user already has a fake agency and if so, skip. Otherwise:
1. Create a unique agency_code (and check it against existing DB records). 
2. Create a new Agency record with this unique agency_code
3. Create a UserAgency record to associate this user with this new agency
4. Create a UserAgencyRole record with the OPPORTUNITY_PUBLISHER role

## Validation steps

1. Code review
2. Run the test script:
`make test args="tests/src/task/agencies/test_setup_lower_env_agencies.py"`

### DIFF
--- a/api/tests/src/task/agencies/test_setup_lower_env_agencies.py
+++ b/api/tests/src/task/agencies/test_setup_lower_env_agencies.py
@@ -52,7 +52,7 @@ def test_setup_lower_env_agencies(enable_factory_create, db_session):
         # For this user, if no fake agency or user role was found
         if not agency_and_role_found:
             errors.append(
-                "User does not have a fake agency and/or a role; user_id = " + str(user.user_id)
+                f"User does not have a fake agency and/or a role; user_id = {user.user_id}"
             )
 
     # Assert that all users were checked successfully
@@ -85,7 +85,7 @@ def test_setup_lower_env_agencies(enable_factory_create, db_session):
                 agency_count += 1
 
         if agency_count != 1:
-            errors2.append("User has more than 1 fake agency; user_id = " + str(user.user_id))
+            errors2.append(f"User has more than 1 fake agency; user_id = {user.user_id}")
 
     # Assert that all users have no more than 1 fake agency
     assert len(errors2) == 0, ". ".join(errors2)


api/src/task/agencies/setup_lower_env_agencies.py
```diff
@@ -0,0 +1,139 @@
...
+        return True
+
+    def create_fake_agency(self) -> Agency:
+        """Create a fake agency"""
+        unique_code = self.generate_agency_code()
+        agency = Agency(
+            agency_code=unique_code,
+            agency_name="Agency for " + unique_code,
+            assistance_listing_number="00.000",
+            agency_submission_notification_setting=3,
```
kkrug: Per Michael's response to you in Slack, you can change this from the int value to:
```suggestion
            agency_submission_notification_setting=AgencySubmissionNotificationSetting.ALWAYS,
```

api/src/task/agencies/setup_lower_env_agencies.py
```diff
@@ -0,0 +1,139 @@
...
+                select(User).options(
+                    selectinload(User.agency_users).selectinload(AgencyUser.agency)
+                )
+            )
+            .scalars()
+            .all()
+        )
+        return users
+
+    def has_no_agency(self, user: User) -> bool:
```
chouinar: Could we invert this? `has_no_agency` means the boolean is a bit of a double-negative.

Would just mean you use it elsewhere as `if not self.has_agency(...):`

infra/api/app-config/env-config/scheduled_jobs.tf
```diff
@@ -93,6 +93,13 @@ locals {
     grantee1 = ["poetry", "run", "flask", "task", "sam-extracts"]
     prod     = ["poetry", "run", "flask", "task", "sam-extracts"]
   }
+  setup-lower-env-agencies-state = {
+    dev      = "ENABLED"
+    staging  = "ENABLED"
+    training = "ENABLED"
```
chouinar: The code says training can't be run, but you have it enabled here.

I'd say that training probably shouldn't be enabled - training is supposed to be prod-like, giving every user an agency wouldn't be prod-like.

infra/api/app-config/env-config/scheduled_jobs.tf
```diff
@@ -214,5 +221,16 @@ locals {
       mem              = try(local.scheduled_jobs_config[var.environment].mem, null)
       environment_vars = try(local.scheduled_jobs_config[var.environment].environment_vars, null)
     }
+    setup-lower-env-agencies = {
+      task_command = ["poetry", "run", "flask", "task", "setup-lower-env-agencies"]
```
chouinar: I'd recommend not enabling this with the job creation. Run it manually once in an environment, you'll hit some issues you didn't expect otherwise.

api/src/task/agencies/setup_lower_env_agencies.py
```diff
@@ -0,0 +1,140 @@
...
+
+        raise Exception("Could not generate a unique agency_code after 5 attempts")
+
+
+@task_blueprint.cli.command(
+    "setup-lower-env-agencies", help="Utility to automatically create agencies for users"
+)
+@flask_db.with_db_session()
+@ecs_background_task(task_name="setup-lower-env-agencies")
+def generate_agency_sql(db_session: db.Session) -> None:
```
chouinar, line 142: You're missing one thing to be able to run this, import this file in https://github.com/HHS/simpler-grants-gov/blob/main/api/src/task/__init__.py

You can run this script locally by doing `make cmd args="task setup-lower-env-agencies"`. True for all our tasks if you want to run them.

api/tests/src/task/agencies/test_setup_lower_env_agencies.py
```diff
@@ -0,0 +1,52 @@
...
+from sqlalchemy.orm import selectinload
+
+from src.constants.static_role_values import OPPORTUNITY_PUBLISHER
+from src.db.models.user_models import AgencyUser, User
+from src.task.agencies.setup_lower_env_agencies import SetupLowerEnvAgenciesTask
+
+
+def test_setup_lower_env_agencies(enable_factory_create, db_session):
+    task = SetupLowerEnvAgenciesTask(db_session)
+    task.run()
```
chouinar, line 19: I see why you didn't hit an issue running this locally. When you run this test on its own, what users exist in the DB? This would get an empty list and end up doing nothing. When you run all the tests together, users will have been made in other tests, so it finds data.

Create a few users - even just `UserFactory.create_batch(size=5)` to be safe.

api/src/task/agencies/setup_lower_env_agencies.py
```diff
@@ -0,0 +1,140 @@
...
+    def get_users(self) -> Sequence[User]:
+        """Get all users"""
+        users = (
+            self.db_session.execute(
+                select(User).options(
+                    selectinload(User.agency_users).selectinload(AgencyUser.agency)
+                )
+            )
+            .scalars()
+            .all()
```
chouinar, line 67: Let's only do this for standard users (the user type of any non-system user). Giving this to system users isn't a big deal, but there's no need to do it for them. This serves a second purpose - to avoid your test picking up a bunch of users from other tests, we could either delete all users (don't - it's slow, it's going to break things), OR update every user to have a different user type. That'll just make them be hidden from this entirely.

```suggestion
            self.db_session.execute(
                select(User).options(
                    selectinload(User.agency_users).selectinload(AgencyUser.agency)
                )
                .where(User.user_type == UserType.STANDARD)
            )
            .scalars()
            .all()
```

api/tests/src/task/agencies/test_setup_lower_env_agencies.py
```diff
@@ -0,0 +1,52 @@
...
+                        agency_and_role_found = True
+                        break
+
+        # For this user, if no fake agency or user role found then stop and fail the test
+        if not agency_and_role_found:
+            test_success = False
+            break
+
+    # If all users were checked successfully, this will assert True
+    assert test_success
```
chouinar: Could you add something to verify if it runs a second time it doesn't give them a second agency?

infra/api/app-config/env-config/scheduled_jobs.tf
Dao-REI: I'm not sure what you mean. Should I not schedule the job for now? Until after we run it in each environment manually? I think you did mention something like that before.

api/src/task/agencies/setup_lower_env_agencies.py
Dao-REI: - from src.task import __init__ -- gave me a lint error because line 36 has def __init__(....
- make cmd args="task setup-lower-env-agencies" -- gave Error: No such command 'setup-lower-env-agencies'.
- make test args="tests/src/task/agencies/test_setup_lower_env_agencies.py" -- this is working for me

infra/api/app-config/env-config/scheduled_jobs.tf
Dao-REI: I have removed the schedule for now.

api/src/task/agencies/setup_lower_env_agencies.py
```diff
@@ -0,0 +1,152 @@
...
+        """Generate an agency code"""
+        for _ in range(5):
+            new_code = "AUTO" + "".join(secrets.choice(string.ascii_uppercase) for _ in range(8))
+
+            existing_record = self.db_session.execute(
+                select(Agency).where(Agency.agency_code == new_code)
+            ).scalar_one_or_none()
+
+            if existing_record is None:
+                self.increment(self.Metrics.NEW_AGENCY_COUNT)
```
chouinar: I don' t think it would affect the numbers - but should we put this in the function that creates the agency?

api/tests/src/task/agencies/test_setup_lower_env_agencies.py
```diff
@@ -0,0 +1,95 @@
...
+
+def test_setup_lower_env_agencies(enable_factory_create, db_session):
+
+    # Create a few users & run the task
+    UserFactory.create_batch(size=5)
+    task = SetupLowerEnvAgenciesTask(db_session)
+    task.run()
+
+    # Get the user records (use expire_all to "commit" the data)
+    db_session.expire_all()
```
chouinar, line 22: Expire all doesn't commit the data, but clears the SQLAlchemy cache. Effectively making it so any query after it will load from the DB rather than use the cache. All to say, yes, this is a good way to work around the responses being a bit odd.

The commit was already done by this point, in run_task, you have `with db_session.begin()` - the end of that block commits.

api/src/task/agencies/setup_lower_env_agencies.py
chouinar: If the `make cmd` command isn't working, this isn't able to be run at all, even non-locally.

You have to go to the above file and import this file otherwise you can't run the command as this file needs to be imported by the app to be known.

Tests work because they import the file, but tests are just tests, they aren't how we actually run.

api/tests/src/task/agencies/test_setup_lower_env_agencies.py
```diff
@@ -0,0 +1,95 @@
...
+    users = (
+        db_session.execute(
+            select(User)
+            .options(selectinload(User.agency_users).selectinload(AgencyUser.agency))
+            .where(User.user_type == UserType.STANDARD)
+        )
+        .scalars()
+        .all()
+    )
+    assert len(users) >= 5
```
chouinar, line 33: We make hundreds of users across our other tests, this'll be a pretty large list that you'll iterate through below.

What if we make this select have a `where(User.user_id.in_(user_ids))` just to cut down on checking other data.

api/src/task/agencies/setup_lower_env_agencies.py
```diff
@@ -0,0 +1,152 @@
+import logging
```
joshtonava, line 1: **issue:** I think you need an `__init__.py` in this folder.

api/tests/src/task/agencies/test_setup_lower_env_agencies.py
```diff
@@ -0,0 +1,95 @@
...
+            agency = agency_user.agency
+            if agency.agency_code.startswith("AUTO"):
+                agency_count += 1
+
+        if agency_count != 1:
+            test2_success = False
+            errors2.append("User has more than 1 fake agency; user_id = " + user.user_id)
+
+    # Assert that all users have no more than 1 fake agency
+    assert len(errors) == 0, ". ".join(errors)
```
joshtonava: issue: Are you intending to verify errors2 is empty?

api/tests/src/task/agencies/test_setup_lower_env_agencies.py
```diff
@@ -0,0 +1,95 @@
...
+                # Fake agency found; check for the user role
+                for agency_user_role in agency_user.agency_user_roles:
+                    if agency_user_role.role_id == OPPORTUNITY_PUBLISHER.role_id:
+                        agency_and_role_found = True
+                        break
+
+        # For this user, if no fake agency or user role was found
+        if not agency_and_role_found:
+            errors.append(
+                "User does not have a fake agency and/or a role; user_id = " + user.user_id
```
joshtonava: issue: I think this will have a runtime TypeError since user.user_id is a UUID

api/src/task/agencies/setup_lower_env_agencies.py
Dao-REI: I figured out what needed to be updated - I got it wrong the first time. Now `make cmd` runs and I can see the updates in my local DB. Thanks

api/tests/src/task/agencies/test_setup_lower_env_agencies.py
```diff
@@ -0,0 +1,95 @@
...
+    for user in users:
+        agency_count = 0
+        for agency_user in user.agency_users:
+            agency = agency_user.agency
+            if agency.agency_code.startswith("AUTO"):
+                agency_count += 1
+
+        if agency_count != 1:
+            test2_success = False
+            errors2.append("User has more than 1 fake agency; user_id = " + user.user_id)
```
joshtonava: **issue:** I think this will have a runtime TypeError since user.user_id is a UUID

api/src/task/agencies/setup_lower_env_agencies.py
Dao-REI: I was following the example from setup_lower_env_sam_extracts.py. I'll remove the metrics.

api/tests/src/task/agencies/test_setup_lower_env_agencies.py
Dao-REI: Hmm... my test script won't work without the `db_session.expire_all()`. Weird.

api/tests/src/task/agencies/test_setup_lower_env_agencies.py
Dao-REI: Oh, yes. Good catch. That was a copy-and-paste error.

api/tests/src/task/agencies/test_setup_lower_env_agencies.py
Dao-REI: I did a quick test and you are correct. I've updated the code to use `str(user.user_id)`

api/src/task/agencies/setup_lower_env_agencies.py
Dao-REI: I think you mean to add my file as an import to this file https://github.com/HHS/simpler-grants-gov/blob/main/api/src/task/__init__.py, which is Michael's instructions above. `make cmd args="task setup-lower-env-agencies"` now works for me and I can see the updates in my local DB.

Dao-REI: That is the __init__.py file under the .../src/task/ folder.

chouinar: What Josh was saying is that we should have an `__init__.py` file in every python folder. Perfectly fine to be entirely blank in most cases.

Things work, but sometimes the behavior ends up a bit unexpected with regards to imports if you leave it out. This file won't be imported anywhere (outside of the tests / the import you had to add for making it known to the app). There's no reason not to include the file, so I'd say we should try to include it.

I don't know the full context on why, just has to do with how modules conceptually work in Python https://docs.python.org/3/tutorial/modules.html#packages

api/tests/src/task/agencies/test_setup_lower_env_agencies.py
chouinar: Generally, use f-strings which'll do any `str(whatever)` for you and are easier to write.

api/tests/src/task/agencies/test_setup_lower_env_agencies.py
chouinar: Was saying it's fine. Most tests don't need it (API tests don't need it generally as the API call happens in a separate DB session I think?) but backend scripts end up a bit of an oddity.

api/src/task/agencies/setup_lower_env_agencies.py
Dao-REI: Done. I added an empty __init__.py file to my new folder, api/tests/src/task/agencies.

Dao-REI:
@chouinar - completed updates and checks have passed. Please re-review.

Issue #8782, opened 2026-03-03
NewRelic browser metrics not working


mdragon:
### Summary

The upgrade of NewRelic (or the removal of the explicit install of axios broke FE client data flowing into NR)

### Acceptance criteria

- NR data is flowing again

PR #8783, opened 2026-03-03
[Issue #8782] Go back to prior Newrelic version to fix client metrics

mdragon:
## Summary

Work for #8782 

## Changes proposed

Newer version of NewRelic failing to initialize in Prod

## Validation steps

- Vuln Scans pass
- Test for NR data flowing once in Staging

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -28,7 +28,7 @@
         "json-pointer": "^0.6.2",
         "json-schema-merge-allof": "^0.8.1",
         "lodash": "^4.17.23",
-        "newrelic": "^13.12.0",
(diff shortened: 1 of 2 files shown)


doug-s-nava (approved):
note that I think we'd be safe running back to 13.3.3 given this comment and this diff
* https://github.com/vercel/vercel/issues/14098#issuecomment-3895463132
* https://github.com/newrelic/node-newrelic/compare/v13.3.3...v13.4.0

For posterity - the issue here is with the New Relic node agent not loading at next server startup with this error `Error: Cannot find module '/frontend/node_modules/newrelic/lib/util/is-string.js'\n`

Issue #8712, opened 2026-03-02
Create Lookup Enums and Lookup Tables for Award Rec work


nathan-stricker:
### Summary

See [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Create-lookup-enums-and-lookup-tables)

Before the Award Recommendation table can be created, we'll need to set up all of its relevant lookup tables and enums. See the [Lookup Tables](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Lookup-Tables) section of the tech spec for details.

### Acceptance criteria

- [x] Lookup tables and enums created:
    - [ ] `lk_award_selection_method`
    - [ ] `lk_award_recommendation_attachment_type`
    - [ ] `lk_award_recommendation_status`
    - [ ] `lk_award_recommendation_type`
    - [ ] `lk_award_recommendation_review_type`
    - [ ] `lk_award_recommendation_risk_type`
    - [ ] `lk_award_recommendation_audit_event`

PR #8790, opened 2026-03-04
[Issue #8712] Create award recommendation lookup enum tables

nathan-stricker:
## Summary

Work for #8712

## Changes proposed

- Created lookup tables:
   - lk_award_selection_method
   - lk_award_recommendation_attachment_type
   - lk_award_recommendation_status
   - lk_award_recommendation_type
   - lk_award_recommendation_review_type
   - lk_award_recommendation_risk_type
   - lk_award_recommendation_audit_event
- Added migration

## Context for reviewers

See [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Create-lookup-enums-and-lookup-tables)

Before the Award Recommendation tables can be created, we need to set up all of its relevant lookup tables and enums. See the [Lookup Tables](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Lookup-Tables) section of the tech spec for details.

### DIFF
--- a/api/src/db/models/lookup_models.py
+++ b/api/src/db/models/lookup_models.py
@@ -654,7 +654,7 @@ class LkAwardRecommendationAttachmentType(LookupTable, TimestampMixin):
     description: Mapped[str]
 
     @classmethod
-    def from_lookup(cls, lookup: Lookup) -> "LkAwardRecommendationAttachmentType":
+    def from_lookup(cls, lookup: Lookup) -> LkAwardRecommendationAttachmentType:
         return LkAwardRecommendationAttachmentType(
             award_recommendation_attachment_type_id=lookup.lookup_val,
             description=lookup.get_description(),
@@ -669,7 +669,7 @@ class LkAwardRecommendationStatus(LookupTable, TimestampMixin):
     description: Mapped[str]
 
     @classmethod
-    def from_lookup(cls, lookup: Lookup) -> "LkAwardRecommendationStatus":
+    def from_lookup(cls, lookup: Lookup) -> LkAwardRecommendationStatus:
         return LkAwardRecommendationStatus(
             award_recommendation_status_id=lookup.lookup_val,
             description=lookup.get_description(),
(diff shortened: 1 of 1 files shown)


api/src/db/models/lookup_models.py
```diff
@@ -551,6 +646,96 @@ def from_lookup(cls, lookup: Lookup) -> LkApplicationStatus:
...
+    def from_lookup(cls, lookup: Lookup) -> LkAwardRecommendationStatus:
+        return LkAwardRecommendationStatus(
+            award_recommendation_status_id=lookup.lookup_val,
+            description=lookup.get_description(),
+        )
+
+
+@LookupRegistry.register_lookup(AWARD_RECOMMENDATION_TYPE_CONFIG)
+class LkAwardRecommendationType(LookupTable, TimestampMixin):
+    __tablename__ = "lk_award_recommendation_type"
```
chouinar, line 681: Missed the comment on the tech spec until now - but yes, good catch, this is the right name.

api/src/constants/lookup_constants.py
```diff
@@ -224,6 +233,55 @@ class ApplicationFormStatus(StrEnum):
...
+    EXCEPTION_CREATED = "exception_created"
+    EXCEPTION_UPDATED = "exception_updated"
+    EXCEPTION_DELETED = "exception_deleted"
+    RISK_CREATED = "risk_created"
+    RISK_UPDATED = "risk_updated"
+    RISK_DELETED = "risk_deleted"
+    APPLICATION_SUBMISSION_UPDATED = "application_submission_updated"
+    REVIEW_CREATED = "review_created"
+    REVIEW_UPDATED = "review_updated"
+    REVIEW_DELETED = "review_deleted"
```
chouinar, line 282: A few of these like exceptions, and some of the review statuses might not happen yet as some functionality was trimmed, but I think are fine to leave as I expect they'll exist in some form eventually.

PR #8794, opened 2026-03-04
[Issue #8101] Modify workflow service to fetch and process SQS event

### LINKED ISSUE #8101: Modify workflow service logic to fetch and process SQS events
### Summary
We want to use our new SQS client to fetch SQS events from a queue, pass them off to the event handler logic and handle the result. As part of this, we'll need to convert the SQS message from JSON into our pydantic model for events. If this errors, we'll need to recover gracefully, but can't process that event and should leave it alone (for now, might change that approach in the future to delete, but keep it simple for the moment).

| Scenario | Behavior |
| - | - |
| Event processed successfully | Event is deleted from SQS queue |
| A non-retryable error is raised | Event is deleted from SQS queue |
| A retryable error is raised | Event is NOT deleted from SQS queue |
| Any other exception | Event is NOT deleted from SQS queue + log an error |

All scenarios should have logging to make clear which scenario occurred. A log message should exist that we can easily use to make a metric from (says which scenario it hit for all 4).

If a retryable/general exception occurs, we should do a rollback, this is so that if any error did occur partially through updating the DB, we don't keep that state.

### Acceptance criteria

- [ ] Convert the SQS message into our event format with Pydantic models - if this errors, log an error, don't delete the message.
- [ ] 4 scenarios above are covered and tested

hao10282025-sudo:
## Summary

Fixes #8101

## Changes proposed

We want to use our new SQS client to fetch SQS events from a queue, pass them off to the event handler logic and handle the result. As part of this, we'll need to convert the SQS message from JSON into our pydantic model for events. If this errors, we'll need to recover gracefully, but can't process that event and should leave it alone (for now, might change that approach in the future to delete, but keep it simple for the moment).

## Context for reviewers

| Scenario | Behavior |
| - | - |
| Event processed successfully | Event is deleted from SQS queue |
| A non-retryable error is raised | Event is deleted from SQS queue |
| A retryable error is raised | Event is NOT deleted from SQS queue |
| Any other exception | Event is NOT deleted from SQS queue + log an error |

All scenarios should have logging to make clear which scenario occurred. A log message should exist that we can easily use to make a metric from (says which scenario it hit for all 4).

If a retryable/general exception occurs, we should do a rollback, this is so that if any error did occur partially through updating the DB, we don't keep that state.

## Validation steps

- Convert the SQS message into our event format with Pydantic models - if this errors, log an error, don't delete the message.
- 4 scenarios above are covered and tested

### DIFF
--- a/api/src/workflow/manager/workflow_manager.py
+++ b/api/src/workflow/manager/workflow_manager.py
@@ -157,27 +157,38 @@ class WorkflowManager:
 
         logger.info("Finished processing workflow events")
 
-    def process_batch(self) -> list[str]:
+    def process_batch(self) -> tuple[list[str], list[str]]:
         """Fetch and process a batch of events from SQS."""
         message_event_dict = self.fetch_messages()
         messages_to_delete: list[str] = []
+        messages_to_keep: list[str] = []
 
         # Note that the initial approach won't be multi-threaded
         # We'll follow-up on that later
 
-        if message_event_dict:
-            for receipt_handle, event in message_event_dict.items():
-                event_result = handle_event(event)
-                if event_result in [
-                    WorkflowEventProcessingResult.SUCCESS,
-                    WorkflowEventProcessingResult.NON_RETRYABLE_ERROR,
-                ]:
-                    messages_to_delete.append(receipt_handle)
+        for receipt_handle, event in message_event_dict.items():
+            messages_to_keep.append(receipt_handle)
 
-                logger.info(
-                    "Processed workflow event",
-                    extra=event.get_log_extra() | {"event_result": event_result},
+            try:
+                event_result = handle_event(event)
+            except Exception:
+                logger.exception(
+                    "Fail to handle current event",
+                    extra={"receipt handle:": receipt_handle, "event id": event.event_id},
                 )
+                continue
+
+            if event_result in [
+                WorkflowEventProcessingResult.SUCCESS,
+                WorkflowEventProcessingResult.NON_RETRYABLE_ERROR,
+            ]:
+                messages_to_delete.append(receipt_handle)
+                messages_to_keep.remove(receipt_handle)
+
+            logger.info(
+                "Processed workflow event",
+                extra=event.get_log_extra() | {"event_result": event_result},
+            )
 
         if messages_to_delete:
             self.delete_messages(messages_to_delete)
@@ -186,17 +197,19 @@ class WorkflowManager:
         self.metrics["batches_processed"] += 1
         self.metrics["events_processed"] += len(message_event_dict)
 
-        # return messages to delete receipt handles for testing purposes
-        return messages_to_delete
+        # return messages to delete and messages to keep handles for testing purposes
+        return messages_to_delete, messages_to_keep
 
     def fetch_messages(self) -> dict[str, WorkflowEvent]:
-        event_dict = {}
+        event_dict: dict[str, WorkflowEvent] = {}
+        messages = []
         try:
             messages = self.sqs_client.receive_messages(
                 wait_time=self.config.workflow_cycle_duration
             )
         except Exception:
             logger.exception("Failed to fetch messages from SQS")
+            return event_dict
 
         for message in messages:
             try:
@@ -213,7 +226,7 @@ class WorkflowManager:
         try:
             delete_result = self.sqs_client.delete_message_batch(receipt_handles)
             if delete_result.failed_deletes:
-                logger.exception(
+                logger.error(
                     "Failed to delete messages from SQS queue",
                     extra={"failed_deletes": list(delete_result.failed_deletes)},
                 )
--- a/api/tests/workflow/manager/test_workflow_manager.py
+++ b/api/tests/workflow/manager/test_workflow_manager.py
@@ -20,14 +20,12 @@ from src.workflow.manager.workflow_manager import (
     WorkflowManagerConfig,
     handle_event,
 )
-from src.workflow.workflow_background_task import workflow_background_task
 from tests.src.db.models.factories import OpportunityFactory, UserFactory, WorkflowFactory
 from tests.workflow.state_machine.test_state_machines import BasicState
 
 logger = logging.getLogger(__name__)
 
 
-@workflow_background_task()
 def test_workflow_manager(workflow_sqs_queue, app, valid_sqs_message):
     """Test process_events() processes batches and tracks metrics correctly."""
     boto_client = boto3.client("sqs", region_name="us-east-1")
@@ -82,13 +80,13 @@ def test_convert_sqs_message_to_workflow_event_success(valid_sqs_message):
     assert str(result.event_id) == json.loads(valid_sqs_message.body)["event_id"]
 
 
-@workflow_background_task()
 def test_workflow_sqs_messages_process_batch_success(workflow_sqs_queue, app):
-    """Test process_events() processes batches and tracks metrics correctly."""
+    """Test process_batch() processes batches and tracks metrics correctly, for success case"""
     boto_client = boto3.client("sqs", region_name="us-east-1")
     sqs_client = SQSClient(queue_url=workflow_sqs_queue, sqs_client=boto_client)
 
     messages_to_delete_handles: list[str] = []
+    messages_to_keep_handles: list[str] = []
     message_deleted = False
 
     user = UserFactory.create()
@@ -109,7 +107,7 @@ def test_workflow_sqs_messages_process_batch_success(workflow_sqs_queue, app):
         "event_id": test_event_id,
         "acting_user_id": user.user_id,
         "event_type": WorkflowEventType.START_WORKFLOW,
-        "process_workflow_context": process_workflow_context.__dict__,
+        "process_workflow_context": process_workflow_context.model_dump(),
         "start_workflow_context": {
             "workflow_type": WorkflowType.BASIC_TEST_WORKFLOW,
             "entity_type": WorkflowEntityType.OPPORTUNITY,
@@ -124,7 +122,7 @@ def test_workflow_sqs_messages_process_batch_success(workflow_sqs_queue, app):
     workflow_manager = WorkflowManager(config=config)
 
     with app.app_context():
-        messages_to_delete_handles = workflow_manager.process_batch()
+        messages_to_delete_handles, messages_to_keep_handles = workflow_manager.process_batch()
 
     # change_message_visibility will throw invalid handle exception, if the message is already deleted
     try:
@@ -147,8 +145,184 @@ def test_workflow_sqs_messages_process_batch_success(workflow_sqs_queue, app):
     metrics = workflow_manager.metrics
     assert metrics["batches_processed"] == 1
     assert metrics["events_processed"] >= 1
+    assert len(messages_to_delete_handles) == 1
+    assert len(messages_to_keep_handles) == 0
+    # Verify the message was deleted and is no longer in the queue
     assert message_deleted
     assert not message_post_process.get("Messages")
+
+
+def test_workflow_sqs_messages_process_batch_retryable(workflow_sqs_queue, app):
+    """Test process_batch() processes batches and tracks metrics correctly, for retryable case"""
+    boto_client = boto3.client("sqs", region_name="us-east-1")
+    sqs_client = SQSClient(queue_url=workflow_sqs_queue, sqs_client=boto_client)
+
+    messages_to_delete_handles: list[str] = []
+    messages_to_keep_handles: list[str] = []
+
+    # Create test user and opportunity
+    user = UserFactory.create()
+    opportunity = OpportunityFactory.create()
+
+    workflow = WorkflowFactory.create(
+        workflow_type=WorkflowType.BASIC_TEST_WORKFLOW,
+        current_workflow_state="not-a-valid-state",
+        has_opportunity=True,
+    )
+
+    process_workflow_context = ProcessWorkflowEventContext(
+        workflow_id=workflow.workflow_id, event_to_send="middle_to_end"
+    )
+
+    test_event_id = uuid.uuid4()
+
+    test_message_body_retryable = {
+        "event_id": test_event_id,
+        "acting_user_id": user.user_id,
+        "event_type": WorkflowEventType.PROCESS_WORKFLOW,
+        "process_workflow_context": process_workflow_context.model_dump(),
+        "start_workflow_context": {
+            "workflow_type": WorkflowType.BASIC_TEST_WORKFLOW,
+            "entity_type": WorkflowEntityType.OPPORTUNITY,
+            "entity_id": opportunity.opportunity_id,
+        },
+    }
+
+    sqs_client.send_message(test_message_body_retryable)
+
+    # Execute: create manager with max 3 batches and 0 sleep time
+    config = WorkflowManagerConfig(workflow_cycle_duration=0, workflow_maximum_batch_count=1)
+    workflow_manager = WorkflowManager(config=config)
+
+    with app.app_context():
+        messages_to_delete_handles, messages_to_keep_handles = workflow_manager.process_batch()
+
+    # Change message visiblity
+    response = boto_client.change_message_visibility(
+        QueueUrl=workflow_sqs_queue, ReceiptHandle=messages_to_keep_handles[0], VisibilityTimeout=0
+    )
+    logger.info(f"Change message visibility response: {response}")
+
+    message_post_process = sqs_client.receive_messages(max_messages=5, wait_time=2)
+
+    # Verify: check metrics reflect 3 batches processed
+    metrics = workflow_manager.metrics
+    assert metrics["batches_processed"] == 1
+    assert metrics["events_processed"] >= 1
+    assert len(messages_to_delete_handles) == 0
+    assert len(messages_to_keep_handles) == 1
+    # Verify the message was not deleted and is still in the queue
+    assert len(message_post_process) == 1
+    assert str(json.loads(message_post_process[0].body)["event_id"]) == str(test_event_id)
+
+
+def test_workflow_sqs_messages_process_batch_mix_cases(workflow_sqs_queue, app, valid_sqs_message):
+    """Test process_batch() processes batches and tracks metrics correctly.
+    This test includes a mix of successful processing, retryable errors, and non-retryable errors to ensure the manager handles each case as expected.
+    """
+    boto_client = boto3.client("sqs", region_name="us-east-1")
+    sqs_client = SQSClient(queue_url=workflow_sqs_queue, sqs_client=boto_client)
+
+    messages_to_delete_handles: list[str] = []
+    messages_to_keep_handles: list[str] = []
+
+    # Create successful instance
+    test_event_id = uuid.uuid4()
+    user = UserFactory.create()
+    opportunity = OpportunityFactory.create()
+    workflow = WorkflowFactory.create(
+        workflow_type=WorkflowType.BASIC_TEST_WORKFLOW,
+        current_workflow_state=BasicState.MIDDLE,
+        has_opportunity=True,
+    )
+
+    process_workflow_context = ProcessWorkflowEventContext(
+        workflow_id=workflow.workflow_id, event_to_send="middle_to_end"
+    )
+
+    test_message_body_success = {
+        "event_id": str(test_event_id),
+        "acting_user_id": user.user_id,
+        "event_type": WorkflowEventType.PROCESS_WORKFLOW,
+        "process_workflow_context": process_workflow_context.model_dump(),
+        "start_workflow_context": {
+            "workflow_type": WorkflowType.BASIC_TEST_WORKFLOW,
+            "entity_type": WorkflowEntityType.OPPORTUNITY,
+            "entity_id": opportunity.opportunity_id,
+        },
+    }
+    sqs_client.send_message(test_message_body_success)
+
+    # Create a retryable instance
+    test_event_id = uuid.uuid4()
+    user = UserFactory.create()
+    opportunity = OpportunityFactory.create()
+    workflow = WorkflowFactory.create(
+        workflow_type=WorkflowType.BASIC_TEST_WORKFLOW,
+        current_workflow_state="not-a-valid-state",
+        has_opportunity=True,
+    )
+
+    process_workflow_context = ProcessWorkflowEventContext(
+        workflow_id=workflow.workflow_id, event_to_send="middle_to_end"
+    )
+
+    test_message_body_retryable = {
+        "event_id": test_event_id,
+        "acting_user_id": user.user_id,
+        "event_type": WorkflowEventType.PROCESS_WORKFLOW,
+        "process_workflow_context": process_workflow_context.model_dump(),
+        "start_workflow_context": {
+            "workflow_type": WorkflowType.BASIC_TEST_WORKFLOW,
+            "entity_type": WorkflowEntityType.OPPORTUNITY,
+            "entity_id": opportunity.opportunity_id,
+        },
+    }
+
+    sqs_client.send_message(test_message_body_retryable)
+
+    # Create a non-retryable instance
+    test_event_id = uuid.uuid4()
+    user = UserFactory.create()
+    opportunity = OpportunityFactory.create()
+    workflow = WorkflowFactory.create(
+        workflow_type=WorkflowType.BASIC_TEST_WORKFLOW,
+        current_workflow_state=BasicState.MIDDLE,
+        has_opportunity=True,
+    )
+
+    process_workflow_context = ProcessWorkflowEventContext(
+        workflow_id=workflow.workflow_id, event_to_send="middle_to_end"
+    )
+
+    # setup invalid user id for non-retryable error
+    test_message_body_non_retryable = {
+        "event_id": test_event_id,
+        "acting_user_id": "abcded1e-8a2f-4e5a-8b1c-9d2e3f4abcde",
+        "event_type": WorkflowEventType.PROCESS_WORKFLOW,
+        "process_workflow_context": process_workflow_context.model_dump(),
+        "start_workflow_context": {
+            "workflow_type": WorkflowType.BASIC_TEST_WORKFLOW,
+            "entity_type": WorkflowEntityType.OPPORTUNITY,
+            "entity_id": opportunity.opportunity_id,
+        },
+    }
+
+    sqs_client.send_message(test_message_body_non_retryable)
+
+    # Execute: create manager with max 3 batches and 0 sleep time
+    config = WorkflowManagerConfig(workflow_cycle_duration=10, workflow_maximum_batch_count=10)
+    workflow_manager = WorkflowManager(config=config)
+
+    with app.app_context():
+        messages_to_delete_handles, messages_to_keep_handles = workflow_manager.process_batch()
+
+    # Verify: check metrics reflect 3 batches processed
+    metrics = workflow_manager.metrics
+    assert metrics["batches_processed"] == 1
+    assert metrics["events_processed"] >= 3
+    assert len(messages_to_delete_handles) == 2
+    assert len(messages_to_keep_handles) == 1
 
 
 def test_process_sqs_event_success(app, db_session):
--- a/api/tests/conftest.py
+++ b/api/tests/conftest.py
@@ -30,6 +30,7 @@ from src.db.models.opportunity_models import Opportunity
 from src.db.models.staging import metadata as staging_metadata
 from src.db.models.user_models import AgencyUser, UserApiKey
 from src.util.local import load_local_env_vars
+from src.workflow.workflow_background_task import _init_newrelic_app
 from tests.lib import db_testing
 from tests.lib.auth_test_utils import mock_oauth_endpoint
 from tests.lib.db_testing import cascade_delete_from_db_table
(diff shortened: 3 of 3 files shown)


api/src/workflow/manager/workflow_manager.py
```diff
@@ -160,42 +159,77 @@ def process_events(self) -> None:
...
-            # Just verify that the DB connection works for now
-            # by doing a very simple query.
-            # Will replace later with more meaningful logic.
-            result = db_session.scalar(text("SELECT 1 AS healthy"))
-            if result != 1:
-                raise Exception("Cannot query DB")
+            try:
+                with workflow_transaction(f"process-{event.event_id}"):
+                    history_event = WorkflowEventHistory(
+                        event_data=None,  # will update with actual data after processing
```
hao10282025-sudo: any template for this event data?

api/src/workflow/manager/workflow_manager.py
```diff
@@ -160,42 +159,77 @@ def process_events(self) -> None:
...
+                    return WorkflowEventProcessingResult.SUCCESS
 
-        logger.info(
-            "Finished processing event",
-            extra={"event_data": event.event_data, "event_type": event.event_type},
-        )
+            except NonRetryableWorkflowError:
+                logger.warning("Encountered non-retryable workflow error", exc_info=True)
+                # Rollback any partial changes
+                db_session.rollback()
```
hao10282025-sudo: if non-retryable, should it be commit or rollback?
for a workflow having non-retryable issue, it will be suspended and have a manual fix?

api/src/workflow/manager/workflow_manager.py
```diff
@@ -160,42 +159,77 @@ def process_events(self) -> None:
...
-            # Will replace later with more meaningful logic.
-            result = db_session.scalar(text("SELECT 1 AS healthy"))
-            if result != 1:
-                raise Exception("Cannot query DB")
+            try:
+                with workflow_transaction(f"process-{event.event_id}"):
+                    history_event = WorkflowEventHistory(
+                        event_data=None,  # will update with actual data after processing
+                        sent_at=datetime_util.utcnow(),
+                        is_successfully_processed=False,
```
hao10282025-sudo: should this "is_successfully_processed" be set to "True" here and always have "True" value?

api/src/workflow/manager/workflow_manager.py
chouinar: Event data is the SQS event that we processed, the raw version of the workflow event

api/src/workflow/manager/workflow_manager.py
chouinar: This'll depend on whether it is successful.

* No errors -> Add a workflow event history record with this set to true
* Non-retryable error -> Add a workflow event history record with this set to false
* Retryable / any other error -> Don't add

Note that the way this should look in the logic might need to be a bit of a different order.
```py
# Make the event outside of the context of the DB session transaction
# as if we error, that'll rollback and then we need to add it again
history_event = WorkflowEventHistory(..., is_successfully_processed=True)

try:
    with db_session.begin():
        db_session.add(history_event)
        event_handler.process(...)
except NonRetryableWorkflowError:
    with db_session.begin():
         history_event.is_successfully_processed = False
         db_session.add(history_event)
# other excepts don't do anything with the history event but still exist
```

---

Although I believe there's a follow-up ticket for actually putting the history event in the DB, I think you have to do it now because the underlying logic expects it to be there. We could have that one be where we handle the non-retryable case?

api/src/workflow/manager/workflow_manager.py
```diff
@@ -160,42 +159,77 @@ def process_events(self) -> None:
...
+            except NonRetryableWorkflowError:
+                logger.warning("Encountered non-retryable workflow error", exc_info=True)
+                # Rollback any partial changes
+                db_session.rollback()
+                return WorkflowEventProcessingResult.NON_RETRYABLE_ERROR
+
+            except RetryableWorkflowError:
+                logger.warning("Encountered retryable workflow error", exc_info=True)
+                # Rollback any partial changes
+                db_session.rollback()
```
chouinar: Rollbacks should almost never be used. Here's how the DB session works

```py
with db_session.begin(): # Begin starts a transaction
     db_session.do_something()
# If the code gets here without any error, the transaction is ended and a flush+commit is done.
# If the with block is exited with any sort of error (even if it's not from the DB session itself), it
# doesn't complete the transaction.
```

In our entire app, we have two rollbacks, the only time we really need to use them is if you do the following:

```py
try:
    with db_session.begin():
          db_session.do_something()
except Exception as e:
     if not db_session.is_active:
          db_session.rollback()

    with db_session.begin():
          # do something with the DB session, usually adding an error to a table
```

With what we're doing in this bit of work, we only would need a rollback for the Non-retryable case (and only if it is active).

api/src/workflow/manager/workflow_manager.py
```diff
@@ -160,42 +159,77 @@ def process_events(self) -> None:
...
+                    history_event = WorkflowEventHistory(
+                        event_data=None,  # will update with actual data after processing
+                        sent_at=datetime_util.utcnow(),
+                        is_successfully_processed=False,
+                        event_id=event.event_id,
+                    )
+
+                    event_handler = EventHandler(db_session, event, history_event)
+                    event_handler.process()
+                    db_session.commit()
```
chouinar: Don't commit, `db_session.begin()` does that for us.

api/src/workflow/manager/workflow_manager.py
```diff
@@ -160,42 +159,77 @@ def process_events(self) -> None:
...
         )
         with db_session.begin():
-            # Just verify that the DB connection works for now
-            # by doing a very simple query.
-            # Will replace later with more meaningful logic.
-            result = db_session.scalar(text("SELECT 1 AS healthy"))
-            if result != 1:
-                raise Exception("Cannot query DB")
+            try:
+                with workflow_transaction(f"process-{event.event_id}"):
```
chouinar: Just make this the event type. For context, this value gets used to break the transactions in New Relic into different types. We don't want these to be unique for every event, we want them to overlap.

api/src/workflow/manager/workflow_manager.py
```diff
@@ -96,20 +93,22 @@ def handle_interrupt(self, signum: int, frame: FrameType | None) -> None:
...
-        logger.info("Fetching workflow events")
-
-        # Generate 1-5 random events
-        events = []
-        for i in range(random.randint(1, 5)):
-            event_type = random.choice(STUB_EVENT_TYPES)
-            events.append(StubEvent(f"Random event {i}", event_type))
-
-        return events
+    def fetch_event(self, message: SQSMessage) -> WorkflowEvent:
```
chouinar: I think I'd name this something like `parse_event` - it's not fetching, it's parsing/converting.

api/src/workflow/manager/workflow_manager.py
```diff
@@ -96,20 +93,22 @@ def handle_interrupt(self, signum: int, frame: FrameType | None) -> None:
...
+                "Failed to parse SQS message body as JSON",
+                extra={"message_id": message.message_id, "error": str(e)},
+            )
+            raise ValueError(f"Invalid JSON in SQS message body: {e}") from e
+        except ValidationError as e:
+            logger.error(
+                "Failed to validate SQS message as WorkflowEvent",
+                extra={"message_id": message.message_id, "error": str(e)},
+            )
+            raise
```
chouinar, line 111: Don't log the error. Do `logger.exception` which is the same as `logger.error` but `exc_info=True` is set. This adds the stack trace and a bunch of other error info to the logs for you in a consistent pattern. For context, when the `exc_info` (which is a tuple of stack trace info) is present in logs, this function is what does that https://github.com/HHS/simpler-grants-gov/blob/main/api/src/logging/flask_logger.py#L232

```suggestion
        except json.JSONDecodeError as e:
            logger.exception(
                "Failed to parse SQS message body as JSON",
                extra={"message_id": message.message_id},
            )
            raise ValueError(f"Invalid JSON in SQS message body: {e}") from e
        except ValidationError as e:
            logger.exception(
                "Failed to validate SQS message as WorkflowEvent",
                extra={"message_id": message.message_id},
            )
            raise
```

api/src/workflow/manager/workflow_manager.py
```diff
@@ -160,42 +159,77 @@ def process_events(self) -> None:
...
+        for message in messages:
+            workflow_event = self.fetch_event(message)
+            events.append(workflow_event)
+            event_result = handle_event(workflow_event)
+            if event_result in [
+                WorkflowEventProcessingResult.SUCCESS,
+                WorkflowEventProcessingResult.NON_RETRYABLE_ERROR,
+            ]:
+                messages_to_delete.append(message.receipt_handle)
+
```
chouinar, line 168: Add a log message so we can use that for making some metric graphs.

```suggestion
        for message in messages:
            workflow_event = self.fetch_event(message)
            events.append(workflow_event)
            event_result = handle_event(workflow_event)
            if event_result in [
                WorkflowEventProcessingResult.SUCCESS,
                WorkflowEventProcessingResult.NON_RETRYABLE_ERROR,
            ]:
                messages_to_delete.append(message.receipt_handle)
                
            logger.info("Processed workflow event", extra=workflow_event.get_log_extra() | {"event_result": event_result}) 

```

api/src/workflow/manager/workflow_manager.py
```diff
@@ -160,42 +159,77 @@ def process_events(self) -> None:
...
+        # Delete messages that were successfully processed or had non-retryable errors
+        if messages_to_delete:
+            logger.info(f"Deleting {len(messages_to_delete)} messages from SQS queue")
```
chouinar: Put any metrics in the extra. If it's in the extra, we can make a graph out this.

Same with the failed deletes one below.

```suggestion
            logger.info("Deleting messages from SQS queue", extra={"message_count": len(messages_to_delete)})
```

api/tests/workflow/manager/test_workflow_manager.py
```diff
@@ -1,19 +1,170 @@
...
+from src.workflow.workflow_background_task import workflow_background_task
+from src.workflow.workflow_errors import NonRetryableWorkflowError, RetryableWorkflowError
 
-    Will be refactored/iterated on significantly
-    as we build out more functionality.
-    """
-    # Don't actually sleep in the test
+
+@patch("src.workflow.manager.workflow_manager.SQSClient")
+@patch("src.workflow.manager.workflow_manager.EventHandler")
```
chouinar: Don't patch the SQS client, we should be able to use the real SQS queue (well, the mocked one from moto) for any of these tests.

I'd say event handler also shouldn't be mocked, we want to make sure what we send to the event handler works. Although it might be necessary for the general error case if we don't have a way to cause it.

Unless we're talking about external calls - I'd say mocking is rarely what we want to do.

api/src/workflow/manager/workflow_manager.py
hao10282025-sudo: updated

api/src/workflow/manager/workflow_manager.py
hao10282025-sudo: updated

api/src/workflow/manager/workflow_manager.py
hao10282025-sudo: updated

api/src/workflow/manager/workflow_manager.py
hao10282025-sudo: updated

api/src/workflow/manager/workflow_manager.py
hao10282025-sudo: updated

api/src/workflow/manager/workflow_manager.py
hao10282025-sudo: updated with the new added utility

api/src/workflow/manager/workflow_manager.py
hao10282025-sudo: updated

api/src/workflow/manager/workflow_manager.py
hao10282025-sudo: updated together with other feedback

api/src/workflow/manager/workflow_manager.py
hao10282025-sudo: got it, updated

api/tests/workflow/manager/test_workflow_manager.py
hao10282025-sudo: updated
also checked the below error/exception test cases, did not find proper way or scenario to cause retryable and non-retryable error, still having event handler mock at this moment

api/src/workflow/manager/workflow_manager.py
hao10282025-sudo: got it, updated

api/src/workflow/manager/workflow_manager.py
```diff
@@ -160,42 +159,88 @@ def process_events(self) -> None:
...
+        except RetryableWorkflowError:
+            logger.warning(
+                "Encountered retryable workflow error while processing event", exc_info=True
+            )
+            return WorkflowEventProcessingResult.RETRYABLE_ERROR
+
+        except Exception:
+            # log specific error for any other exception
+            logger.exception("Unexpected error processing workflow event")
+            return WorkflowEventProcessingResult.GENERAL_ERROR
```
kkrug, line 281: Should we also be doing rollbacks on retryable and general errors?

api/src/workflow/manager/workflow_manager.py
```diff
@@ -160,42 +159,88 @@ def process_events(self) -> None:
...
-                handle_event(event)
-            except RetryableWorkflowError:
-                logger.warning("Encountered retryable workflow error", exc_info=True)
-            except NonRetryableWorkflowError:
-                logger.warning("Encountered non-retryable workflow error", exc_info=True)
-                # In a future ticket, we'll write back the event to SQS
-            except Exception:
-                logger.exception("Failed to process event")
+        for message in messages:
+            workflow_event = self.parse_event(message)
```
chouinar: Because this function can error, we'll want a try-except around it, otherwise it would fail the whole process.

```suggestion
            try:
                workflow_event = self.parse_event(message)
            except Exception:
                 logger.exception("Failed to parse SQS message")
                 continue
```

api/src/workflow/manager/workflow_manager.py
```diff
@@ -160,42 +159,88 @@ def process_events(self) -> None:
 
     def process_batch(self) -> None:
         """Fetch and process a batch of events from SQS."""
-        events = self.fetch_events()
+        messages = self.sqs_client.receive_messages(wait_time=self.config.workflow_cycle_duration)
```
chouinar: We'll want a try/catch around this and the delete call. I'd recommend to keep this a bit readable we make them a function. We need this because one minor network issue shouldn't take down this whole process.

Thinking something like:
```py
def fetch_messages(self) -> list[WorkflowEvent]:
     events = []
     try:
         messages = self.sqs_client.receive_messages(wait_time=self.config.workflow_cycle_duration)
     except Exception:
          logger.exception("Failed to fetch messages from SQS")
          return []

     for message in messages:
         try:
             workflow_event = self.parse_event(message)
             events.append(workflow_event)
         except Exception:
             logger.exception("Failed to convert SQS message")

    return events
```

Then the calling logic simplifies a bit to just fetching messages and processing without needing to do a bunch of nested try/catches. We'll want something similar for the deletes, to put them in a function.

api/src/workflow/manager/workflow_manager.py
```diff
@@ -160,42 +159,88 @@ def process_events(self) -> None:
...
         # Very simple metrics for test purposes
         self.metrics["batches_processed"] += 1
         self.metrics["events_processed"] += len(events)
 
 
 @flask_db.with_db_session()
-def handle_event(db_session: db.Session, event: StubEvent) -> None:
-    with workflow_transaction(event.event_type):
+def handle_event(db_session: db.Session, event: WorkflowEvent) -> WorkflowEventProcessingResult:
+    with workflow_transaction(f"process-{event.event_type}"):
```
chouinar: Just do the event type.

```suggestion
    with workflow_transaction(event.event_type):
```

api/src/workflow/manager/workflow_manager.py
```diff
@@ -160,42 +159,88 @@ def process_events(self) -> None:
...
-            # Will replace later with more meaningful logic.
-            result = db_session.scalar(text("SELECT 1 AS healthy"))
-            if result != 1:
-                raise Exception("Cannot query DB")
 
-        logger.info(
-            "Finished processing event",
-            extra={"event_data": event.event_data, "event_type": event.event_type},
+        history_event = WorkflowEventHistory(
+            event_data=json.dumps(event.__dict__, default=json_encoder),
```
chouinar: `model_dump` turns a pydantic model into a python dict, which is what I did when sending events. Technically `__dict__` sorta does that (nearly every object in python is secretly a dict), but the nested objects might not have the right types.

```suggestion
            event_data=json.dumps(event.model_dump(), default=json_encoder),
```

api/src/workflow/manager/workflow_manager.py
kkrug: If `parse_event` raises an error here, wouldn't it propagate up to the caller? This would mean that one error while iterating through `messages` would break the whole loop and subsequent messages would not be processed. I think a `try/except` with `continue` would allow the error to be logged but the processing to continue on.

api/src/workflow/manager/workflow_manager.py
```diff
@@ -160,42 +159,88 @@ def process_events(self) -> None:
...
 
 @flask_db.with_db_session()
-def handle_event(db_session: db.Session, event: StubEvent) -> None:
-    with workflow_transaction(event.event_type):
+def handle_event(db_session: db.Session, event: WorkflowEvent) -> WorkflowEventProcessingResult:
+    with workflow_transaction(f"process-{event.event_type}"):
         logger.info(
             "Processing event",
-            extra={"event_data": event.event_data, "event_type": event.event_type},
+            extra={"event_type": event.event_type},
```
chouinar: Added a utility function to the workflow event class for getting useful logging info. Could you adjust the logs in this function to use that?

```suggestion
            extra=event.get_log_extra()
```

api/tests/workflow/manager/test_workflow_manager.py
```diff
@@ -1,19 +1,174 @@
...
+
+def test_convert_sqs_message_to_workflow_event_success(valid_sqs_message):
+    """Test successful fetching of a workflow event from an SQS message."""
+    wfm = WorkflowManager(config=WorkflowManagerConfig())
+    result = wfm.parse_event(valid_sqs_message)
+    assert isinstance(result, WorkflowEvent)
+    assert str(result.event_id) == json.loads(valid_sqs_message.body)["event_id"]
+
+
+@workflow_background_task()
```
chouinar: Why did you add this to the tests? Is there an issue if it's not included?

api/tests/workflow/manager/test_workflow_manager.py
```diff
@@ -1,19 +1,174 @@
-from src.workflow.manager.workflow_manager import WorkflowManager, WorkflowManagerConfig
+import json
```
kkrug, line 1: Adding some tests for various error parsing would be helpful.

api/tests/workflow/manager/test_workflow_manager.py
```diff
@@ -1,19 +1,174 @@
...
+def test_convert_sqs_message_to_workflow_event_success(valid_sqs_message):
+    """Test successful fetching of a workflow event from an SQS message."""
+    wfm = WorkflowManager(config=WorkflowManagerConfig())
+    result = wfm.parse_event(valid_sqs_message)
+    assert isinstance(result, WorkflowEvent)
+    assert str(result.event_id) == json.loads(valid_sqs_message.body)["event_id"]
+
+
+@workflow_background_task()
+def test_process_sqs_event_success(app, valid_sqs_message):
```
chouinar: For these tests, we should check:
* Is the history event in the DB
* Was the event deleted from the SQS queue - this one might be tricky as the SQS queue hides events for a while, probably need to make it so we hide the message for 0 seconds (one of the params we can pass into the receive function).

api/tests/workflow/manager/test_workflow_manager.py
```diff
@@ -1,19 +1,174 @@
...
+    with app.app_context():
+        """Test successful processing of an SQS event."""
+        # Setup mocks
+        mock_session = MagicMock()
```
chouinar: I'd like us not to mock this out if possible - these tests should make sure that everything works together. Setting up the scenarios shouldn't be too hard, you can look at what we do in the event handler tests, I've built several utilities for setting up the data. Happy to help if you have any questions about what is needed.

Suggestions for how to hit each scenario:
* Fine if all these are a similar case, a process workflow event is probably easiest.
* Do this against the `BasicTestStateMachine`, not the initial prototype. They're similar, but we're less likely to change the test one so easier to maintain. Note you have to import that file for it to be found.
* For the error scenarios, what I'd recommend:
  * A retryable error -> `UnexpectedStateError` is probably the easiest to hit, event handler hits this - just set the workflow up in a fake state.
  * Non-retryable error -> A bad event that doesn't exist is the easiest
  * Generic exception -> This one is trickier, this is a case where mocking might be useful, but don't mock the entire handler function, mock a function inside the handler and tell it to throw an error.

api/src/workflow/manager/workflow_manager.py
hao10282025-sudo: As discussed in chat, what I understand is the db_session.begin() will rollback DB operations automatically when exception happens. The rollback added for the Non-retryable case, is only to make it safe to ensure all data is removed and current session is closed, before inserting the history_event data manually

api/tests/workflow/manager/test_workflow_manager.py
hao10282025-sudo: No there is no error, may it get copied from other test case, removed

api/tests/workflow/manager/test_workflow_manager.py
hao10282025-sudo: test cases updated

api/tests/workflow/manager/test_workflow_manager.py
hao10282025-sudo: Yes, updated the 4 test cases for handle_event() with real data setup, added a test case for process_batch() for success case
Did not find a way for process_batch() function error cases handling, will check back if those are necessary

api/tests/workflow/manager/test_workflow_manager.py
hao10282025-sudo: Implemented the 1st item, for 2 out of 4 cases in handle_event() test cases
For 2nd item, I created a test case "test_workflow_sqs_messages_process_batch_success" which is I can think of at this moment, it is checking if the change_message_visibility() can be called for a deleted message (it will throw error), and no new message is received. It does look like tricky for other cases test with error handling

api/src/workflow/manager/workflow_manager.py
hao10282025-sudo: updated

api/src/workflow/manager/workflow_manager.py
hao10282025-sudo: updated

api/src/workflow/manager/workflow_manager.py
hao10282025-sudo: updated

api/src/workflow/manager/workflow_manager.py
hao10282025-sudo: updated

api/src/workflow/manager/workflow_manager.py
hao10282025-sudo: updated

api/src/workflow/manager/workflow_manager.py
hao10282025-sudo: updated

api/src/workflow/manager/workflow_manager.py
```diff
@@ -158,44 +157,112 @@ def process_events(self) -> None:
...
-                handle_event(event)
-            except RetryableWorkflowError:
-                logger.warning("Encountered retryable workflow error", exc_info=True)
-            except NonRetryableWorkflowError:
-                logger.warning("Encountered non-retryable workflow error", exc_info=True)
-                # In a future ticket, we'll write back the event to SQS
-            except Exception:
-                logger.exception("Failed to process event")
+
+        if message_event_dict:
```
chouinar: Just to remove an indent - don't check the `message_event_dict`
```suggestion

```

Unless there's some bit of special logic about an iterable being empty, usually not worth checking if it's empty before iterating over it. If it is, well, then the loop does nothing and that's fine.

api/src/workflow/manager/workflow_manager.py
```diff
@@ -158,44 +157,112 @@ def process_events(self) -> None:
...
+
+        for message in messages:
+            try:
+                event = self.parse_event(message)
+                event_dict[message.receipt_handle] = event
+            except Exception:
+                logger.exception("Failed to convert SQS message")
+                continue
+
+        return event_dict
```
kkrug, line 222: If you hit the Exception in the first `try/except` block, `messages` will be undefined and caused an error when you get to `for message in messages`. You could initialize `messages = []` earlier or `return event_dict` inside the except block.

api/src/workflow/manager/workflow_manager.py
```diff
@@ -158,44 +157,112 @@ def process_events(self) -> None:
...
+
+    def delete_messages(self, receipt_handles: list[str]) -> None:
+        # Delete messages that were successfully processed or had non-retryable errors
+        try:
+            delete_result = self.sqs_client.delete_message_batch(receipt_handles)
+            if delete_result.failed_deletes:
+                logger.exception(
+                    "Failed to delete messages from SQS queue",
+                    extra={"failed_deletes": list(delete_result.failed_deletes)},
+                )
```
kkrug: There is no exception context to log here.
```suggestion
                logger.error(
                    "Failed to delete messages from SQS queue",
                    extra={"failed_deletes": list(delete_result.failed_deletes)},
                )
```

api/src/workflow/manager/workflow_manager.py
chouinar: @kkrug - when we write

```py
with db_session.begin():
    # do stuff
```

That's really doing at least this:
```py
try:
    db_session.begin_transaction() # this isn't actually something you can call directly

    # do stuff
    db_session.flush()
    db_session.commit() # / end transcation
except:
    db_session.put_in_error_state() # also not a real function, it makes the session not work anymore
```

We have to do a rollback because that doesn't happen when it errors, it actually is in a failed state, it won't let you commit or do anything until you undo the buffer basically. But if we do zero more DB operations with the session (like for this case) then no reason to rollback. It's why in say the API endpoints we never do rollbacks because if it errored, we're just going to return an error message and do nothing.

api/tests/workflow/manager/test_workflow_manager.py
```diff
@@ -1,19 +1,304 @@
...
+
+def test_convert_sqs_message_to_workflow_event_success(valid_sqs_message):
+    """Test successful fetching of a workflow event from an SQS message."""
+    wfm = WorkflowManager(config=WorkflowManagerConfig())
+    result = wfm.parse_event(valid_sqs_message)
+    assert isinstance(result, WorkflowEvent)
+    assert str(result.event_id) == json.loads(valid_sqs_message.body)["event_id"]
+
+
+@workflow_background_task()
```
chouinar: Ah, I see why you added this for New Relic setup - however we shouldn't add those to the tests.

Figured out a workaround. Can you add to this file and remove those?

```py
@pytest.fixture(scope="session", autouse=True)
def init_new_relic_app():
    """Setup the new relic app to be initialized so the transaction logic
       won't error when running tests. This won't actually connect to New Relic.
    """
    _init_newrelic_app()
```

api/tests/workflow/manager/test_workflow_manager.py
```diff
@@ -1,19 +1,304 @@
...
+    }
+
+    test_event = WorkflowEvent.model_validate(test_message_body)
+
+    # Process the event - should trigger UnexpectedStateError which is a RetryableWorkflowError
+    with app.app_context():
+        result = handle_event(test_event)
+
+    # Verify the result is a retryable error
+    assert result == WorkflowEventProcessingResult.RETRYABLE_ERROR
```
chouinar, line 413: For the retryable case, could we check that the message is still in the queue / not deleted?

api/src/workflow/manager/workflow_manager.py
hao10282025-sudo: Ah right, the same logic is also provided from Michael, not sure why I did not copy that return or delete by mistake
Added back the exception return

api/tests/workflow/manager/test_workflow_manager.py
hao10282025-sudo: Add 2 new test cases for retryable message and mix case messages for process_batch function

api/tests/workflow/manager/test_workflow_manager.py
hao10282025-sudo: Add the test cases for retryable message and mix messages, for process_batch()

api/tests/workflow/manager/test_workflow_manager.py
hao10282025-sudo: updated

PR #8799, opened 2026-03-04
Fix nofos deploy workflow to use nofos-specific pipeline

sean-navapbc:
## Summary
- Restores `cd-nofos.yml` to use `deploy-nofos.yml` instead of the generic `deploy.yml`
- PR #8548 replaced the nofos CD workflow with the generic template-infra deploy pipeline, but nofos is an external app without a source directory in this repo
- The generic `build-and-publish.yml` fails with `fatal: ambiguous argument 'nofos': unknown revision or path not in the working tree` because it expects a `nofos/` directory for `git log ... main nofos`
- Restores CI checks (`ci-nofos.yml`) and vulnerability scans (`vulnerability-scans-nofos.yml`) that were also removed
- Restores the original environment options (dev, training, grantee1, prod) and push trigger paths

## Test plan
- [ ] Trigger a manual nofos deploy to dev via workflow_dispatch
- [ ] Verify the deploy completes successfully using `deploy-nofos.yml`

Fixes deploy failure: https://github.com/HHS/simpler-grants-gov/actions/runs/22681433564

### DIFF
--- a/.github/workflows/cd-nofos.yml
+++ b/.github/workflows/cd-nofos.yml
@@ -1,17 +1,14 @@
-name: Deploy nofos
-# Need to set a default value for when the workflow is triggered from a git push
-# which bypasses the default configuration for inputs
-run-name: Deploy ${{ inputs.version || 'main' }} to nofos ${{ inputs.environment || 'dev' }}
+name: Deploy NOFOs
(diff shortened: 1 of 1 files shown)


mdragon (approved):
Approved

PR #8804, opened 2026-03-04
rename e2e files and directories

doug-s-nava:
## Summary

unticketed

## Changes proposed

renames some e2e files and directories for consistency


frontend/tests/e2e/search/features/search-results/andor-search-term-switch.feature
Bhavna-Ramachandran, line 1: This should be moved under search/features/search-results, as they main check how search results change for each toggle.

Issue #8238, opened 2026-01-29
Audit Find Dashboards and Business Metrics


joshtonava:
### Summary

Before adding new observability instrumentation for the Find user journey, we need to understand what business metrics and dashboards we currently have. This audit will inventory existing New Relic dashboards for Find-related functionality (search, opportunity views, saved opportunities, saved searches), and document what business metrics are tracked today. The goal is to avoid duplicating existing work and identify opportunities to improve our current dashboards and business metrics for better insights.

Note: This audit focuses on dashboards and business metrics. New Relic feature utilization is covered in the companion ticket. Operational alerting and incident response is covered in a separate audit ticket. Designing new user journey event tracking is covered in #8247.


### What to Document

**Existing Dashboards:**
- What New Relic dashboards exist for Find?
- What metrics do they show?
- Who uses them and for what purpose?
- Are they actively maintained?
- links, screenshots and descriptions of each dashboard
- What gaps exist in operational visibility?
                                                                                                                                                                                                                                                                             
**Existing Business Metrics:**
- Document any business metrics currently tracked (search volume, save rates, etc.)
- Who uses these metrics and how?
- Are they effective for current needs?
- How are they calculated/collected?
- Note: Designing NEW business metrics is covered in #8231 and Event Instrumentation ticket


### Tech Notes
- Scope includes: Search endpoints, OpenSearch performance, opportunity view endpoints, saved opportunity endpoints, saved search endpoints, organization endpoints (All find created/maintained APIs)
- Focus on documenting current state of dashboards and business metrics

                                                                                                                                                                                                                                                                               
### Acceptance Criteria

- Document existing dashboards (screenshots, descriptions, usage patterns, users)
- Documentation of any existing business metrics, their usage, and how they are calculated/collected
- Recommendations provided for:
  - Dashboard improvements or consolidation for better business insights
  - Better business metrics collection approaches
- Team has reviewed findings
- Timebox, after spending 3 points of effort on this we can spin up another ticket to complete the remaining effort

joshtonava:
Do we care about the organization management endpoints here as well?

PR #8808, opened 2026-03-04
[Issue #8686] Add structured logging to saved opportunity/search routes

### LINKED ISSUE #8686: Add structured logging to saved opportunity/search routes
### Summary

The saved opportunity and saved search API routes don't log enough context for us to analyze save behavior in New Relic.  Unlike the search endpoint, which logs query parameters, filters, and result counts as structured attributes, the save  routes only log basic identifiers. This means we can't answer questions like "which searches do users save?" or "how many results matched a saved search?" without additional instrumentation. This was identified as a gap in #8238 

### Tech Notes
| Route | What to Add |                                                                                                    
  |-------|-------------|                                                                                                    
  | Save opportunity | Log `opportunity_id` via `add_extra_data_to_current_request_logs()`. Currently uses inline `extra`  dict only. |                                                                                                               
  | Delete saved opportunity | Log `opportunity_id`. Currently logs nothing structured. |
  | List saved opportunities | Log response pagination info (total_pages, total_records). Currently logs nothing structured. |
  | Save search | Log flattened search query params (via `flatten_dict`), saved search name, and `response.matched_opportunity_count`. Currently only logs `user_id` and `saved_search_id` via inline `extra`. |
  | Delete saved search | Log `saved_search_id` |
  | List saved searches | Log response pagination info (total_pages, total_records). Currently logs nothing structured. |
  | Update saved search | Log `saved_search_id` and any updated fields. |

### Acceptance criteria

- [ ] Routes updated with structured logging:
  - [ ] Save opportunity route
  - [ ] Delete saved opportunity route
  - [ ] List saved opportunities route
  - [ ] Save search route 
  - [ ] Delete saved search route
  - [ ] List saved searches route
  - [ ] Update saved search route
- [ ] Unit tests updated / created for log statements
- [ ] All new attributes are queryable via NRQL

mikehgrantsgov:
## Summary

Fixes #8686

## Changes proposed

Structured logging added to all 8 routes in `user_routes.py`
Adds new tests

| Route | What to Add |
|---|---|
| Save opportunity | Log `opportunity_id` via `add_extra_data_to_current_request_logs()` |
| Delete saved opportunity | Log `opportunity_id` |
| List saved opportunities | Log pagination info (`total_pages`, `total_records`) |
| Save search | Log flattened search query params (via `flatten_dict`), saved search name, `response.matched_opportunity_count` |
| Delete saved search | Log `saved_search_id` |
| List saved searches | Log pagination info |
| Update saved search | Log `saved_search_id` and updated fields |

## Context for reviewers

The saved opportunity and saved search API routes don't log enough context for us to analyze save behavior in New Relic. Unlike the search endpoint, which logs query parameters, filters, and result counts as structured attributes, the save routes only log basic identifiers. This means we can't answer questions like "which searches do users save?" or "how many results matched a saved search?" without additional instrumentation. This was identified as a gap in https://github.com/HHS/simpler-grants-gov/issues/8238

## Validation steps

See new updates to unit tests to confirm fields are being logged.

### DIFF
--- a/api/src/api/users/user_routes.py
+++ b/api/src/api/users/user_routes.py
@@ -429,10 +429,7 @@ def user_save_search(
     search_client: search.SearchClient, db_session: db.Session, user_id: UUID, json_data: dict
 ) -> response.ApiResponse:
     add_extra_data_to_current_request_logs(
-        {
-            **flatten_dict(json_data.get("search_query", {}), prefix="search_query"),
-            "search_query.name": json_data.get("name"),
-        }
+        flatten_dict(json_data.get("search_query", {}), prefix="search_query")
     )
     logger.info("POST /v1/users/:user_id/saved-searches")
 
@@ -447,7 +444,7 @@ def user_save_search(
 
     add_extra_data_to_current_request_logs(
         {
-            "response.matched_opportunity_count": len(saved_search.searched_opportunity_ids or []),
+            "response.matched_opportunity_count": len(saved_search.searched_opportunity_ids),
         }
     )
     logger.info(
@@ -534,9 +531,7 @@ def user_get_saved_searches(
 def user_update_saved_search(
     db_session: db.Session, user_id: UUID, saved_search_id: UUID, json_data: dict
 ) -> response.ApiResponse:
-    add_extra_data_to_current_request_logs(
-        {"saved_search_id": saved_search_id, **flatten_dict(json_data, prefix="update")}
-    )
+    add_extra_data_to_current_request_logs({"saved_search_id": saved_search_id})
     logger.info("PUT /v1/users/:user_id/saved-searches/:saved_search_id")
 
     user_token_session: UserTokenSession = api_jwt_auth.get_user_token_session()
--- a/api/tests/src/api/users/test_user_delete_saved_opportunity.py
+++ b/api/tests/src/api/users/test_user_delete_saved_opportunity.py
@@ -1,5 +1,4 @@
 import logging
-from unittest.mock import patch
 
 import pytest
 
@@ -179,17 +178,13 @@ def test_user_delete_saved_opportunity_logging(
     opportunity = OpportunityFactory.create()
     UserSavedOpportunityFactory.create(user=user, opportunity=opportunity, is_deleted=False)
 
-    with patch(
-        "src.api.users.user_routes.add_extra_data_to_current_request_logs"
-    ) as mock_extra_data:
-        caplog.set_level(logging.INFO)
-        response = client.delete(
-            f"/v1/users/{user.user_id}/saved-opportunities/{opportunity.opportunity_id}",
-            headers={"X-SGG-Token": user_auth_token},
-        )
+    caplog.set_level(logging.INFO)
+    response = client.delete(
+        f"/v1/users/{user.user_id}/saved-opportunities/{opportunity.opportunity_id}",
+        headers={"X-SGG-Token": user_auth_token},
+    )
 
     assert response.status_code == 200
-    mock_extra_data.assert_any_call({"opportunity_id": opportunity.opportunity_id})
     assert any("Deleted saved opportunity" in r.message for r in caplog.records)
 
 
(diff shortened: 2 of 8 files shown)


api/src/api/users/user_routes.py
```diff
@@ -342,6 +345,9 @@ def user_delete_saved_opportunity(
         # Delete the saved opportunity
         delete_saved_opportunity(db_session, user_id, opportunity_id)
 
+    logger.info(
+        "Deleted saved opportunity", extra={"user_id": user_id, "opportunity_id": opportunity_id}
+    )
```
chouinar, line 350: If we did the add_extra call, the opportunity ID would be added here.

Auth also adds the user, although as `auth.user_id` to every request.

api/src/api/users/user_routes.py
```diff
@@ -409,6 +428,12 @@ def user_get_saved_opportunities(
 def user_save_search(
     search_client: search.SearchClient, db_session: db.Session, user_id: UUID, json_data: dict
 ) -> response.ApiResponse:
+    add_extra_data_to_current_request_logs(
+        {
+            **flatten_dict(json_data.get("search_query", {}), prefix="search_query"),
+            "search_query.name": json_data.get("name"),
```
chouinar: We probably should be careful about putting the name of the saved search.

Since that's something the user enters, and could be something like "Bob Smith's example query" - so could potentially have someone's names or other info we might not want, probably better to leave it out. Don't know if there'd be much value in logging it anyways.

api/src/api/users/user_routes.py
```diff
@@ -420,6 +445,11 @@ def user_save_search(
     with db_session.begin():
         saved_search = create_saved_search(search_client, db_session, user_id, json_data)
 
+    add_extra_data_to_current_request_logs(
+        {
+            "response.matched_opportunity_count": len(saved_search.searched_opportunity_ids or []),
```
chouinar: I don't think there is a way to have that not be a list from the logic
```suggestion
            "response.matched_opportunity_count": len(saved_search.searched_opportunity_ids),
```

api/src/api/users/user_routes.py
```diff
@@ -495,6 +534,9 @@ def user_get_saved_searches(
 def user_update_saved_search(
     db_session: db.Session, user_id: UUID, saved_search_id: UUID, json_data: dict
 ) -> response.ApiResponse:
+    add_extra_data_to_current_request_logs(
+        {"saved_search_id": saved_search_id, **flatten_dict(json_data, prefix="update")}
```
chouinar: Similarly, here I don't think we want to log the name of the saved search.

api/tests/src/api/users/test_user_delete_saved_opportunity.py
```diff
@@ -168,3 +171,43 @@ def test_user_delete_saved_opportunity_unauthorized(
...
+    ) as mock_extra_data:
+        caplog.set_level(logging.INFO)
+        response = client.delete(
+            f"/v1/users/{user.user_id}/saved-opportunities/{opportunity.opportunity_id}",
+            headers={"X-SGG-Token": user_auth_token},
+        )
+
+    assert response.status_code == 200
+    mock_extra_data.assert_any_call({"opportunity_id": opportunity.opportunity_id})
+    assert any("Deleted saved opportunity" in r.message for r in caplog.records)
```
chouinar: There's no need to do patches for these tests - I'd strongly recommend not patching unless there's no alternative. Just check caplog, it's the actual log records that are created. If you want to make sure the extras are added, check the last log message (the `end request`) and verify opportunity ID is set. Validate the outcome.

But more broadly, we don't write tests for logging. If you want to look at the logs to verify something as part of a test, or there's a particularly complex logging approach, that's fine. But if we're just doing `logger.info("something", extra={"opportunity_id": opportunity.opportunity_id})` - I'm fine not testing that, it's not going to work in an unknown way. Only thing in this PR that I might check are the flatten_dict ones for search params just to check that it's the format you expect.

mikehgrantsgov: Thanks, updated these tests to read from caplog.

api/src/api/users/user_routes.py
mikehgrantsgov: Updated this, too.

api/src/api/users/user_routes.py
mikehgrantsgov: Thanks, updated this.

api/src/api/users/user_routes.py
mikehgrantsgov: 👍

Issue #8791, opened 2026-03-04
Fix SF-424a PDF output to include entries from Section B


chris-kuryak:
#### Current Behavior: 
Entries in form SF-424a Section B are not being captured in the the PDF generated upon submission. They are showing as blank across all fields of Section B. See Below.

From webform:
<img width="859" height="1098" alt="Image" src="https://github.com/user-attachments/assets/df355c05-8190-4bd3-ac9b-655fa2f0e36d" />

From PDF:
<img width="441" height="1132" alt="Image" src="https://github.com/user-attachments/assets/67a3e3a7-bb90-4407-90b0-d0115be75b55" />

#### Expected Behavior: 
Entries in form SF-424a Section B are appear exactly the same in the PDF generated upon submission.

### Acceptance criteria

- [ ] Entries in form SF-424a Section B are appear exactly the same in the PDF generated upon submission.
- [ ] UAT passed

chris-kuryak:
Consider making a unit test for this issue as a follow-on ticket to be picked up in future sprints.

chris-kuryak:
@myduong-navapbc Here is the follow-on ticket for a longer term fix that you can reference.
Please update the ticket with any known details that I am missing:
https://github.com/HHS/simpler-grants-gov/issues/8818

PR #8811, opened 2026-03-04
[Issue #8791] Fix SF-424a PDF output to include entries from Section B

myduong-navapbc:
## Summary

Fixes for #8791  

### What was broken

`Budget424aSectionB` expects to read the entire SF-424A root object so it can build the **Section B table** from:

- `activity_line_items[] `(activity columns 1–4)
- `total_budget_categories` (column 5 totals)

Inside the widget we derive the data using:

```
const rootFormDataFromContext =
  (formContext as { rootFormData?: unknown } | undefined)?.rootFormData;
const rawValue: unknown = rootFormDataFromContext ?? value ?? {};
const activityItemsUnknown = get(rawValue as object, "activity_line_items");
```

And they were all returning `undefined` from the Print Page
On the print page (PrintForm → FormFields) was not passing any `formContext` into the widget tree:

```
formContext was undefined
rootFormDataFromContext was undefined
rawValue fell back to value
```

In RJSF widgets, value is only the value at the widget’s field path, not the full form data. In this case, value did not include `activity_line_items` at the top level, so:

```
activityItemsUnknown was undefined
activityItems normalized to []
```

This is why the logs showed nothing and why the table inputs for Section B couldn’t resolve the expected rows/columns.

### Fix

`Pass the root data into formContext from the print page` so the widget can access the root form shape consistently:

```
<FormFields
  errors={null}
  formData={savedFormData}
  schema={formSchema}
  uiSchema={uiSchema}
  formContext={{ rootFormData: savedFormData, rootSchema: formSchema }}
/>
```

I've added some unit tests around context forwarding.

Screenshot 2026-03-04 at 1 36 49 PM

### Test Steps

- Sign in
- Do a search for `test` and look for `Test Opportunity for SF424A 1.0`
- Fill out the form and save it
- [verify] when you navigate to the print page (http://localhost:3000/print/application/[app_id]/form/[form_id]), that section B is not blank.

### Other Context

The longer more involved fixed will be covered in this this [enhancement ticket](https://github.com/HHS/simpler-grants-gov/issues/8818).

### DIFF
--- a/frontend/src/components/applyForm/FormFields.test.tsx
+++ b/frontend/src/components/applyForm/FormFields.test.tsx
@@ -219,4 +219,45 @@ describe("buildFormTreeRecursive", () => {
     expect(screen.getByTestId("section--field1")).toBeInTheDocument();
     expect(screen.getByTestId("section--field2")).toBeInTheDocument();
   });
+
+  describe("FormFields formContext forwarding", () => {
+    it("forwards formContext to rendered widgets", () => {
+      const schema: RJSFSchema = {
+        type: "object",
+        properties: {
+          example: { type: "string", title: "Example" },
+        },
+      };
+
+      const uiSchema: UiSchema = [
+        {
+          type: "field",
+          definition: "/properties/example",
+        },
+      ];
+
+      const formContext = {
+        rootFormData: { activity_line_items: [{ activity_title: "Test" }] },
+        rootSchema: schema,
+      };
+
+      render(
+        <FormFields
+          errors={null}
+          formData={{ example: "hello" }}
+          schema={schema}
+          uiSchema={uiSchema}
+          formContext={formContext}
+        />,
+      );
+
+      // grab the rendered input
+      const input = screen.getByTestId("example");
+
+      expect(input).toBeInTheDocument();
+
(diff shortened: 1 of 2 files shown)


myduong-navapbc:
> was section b the only one using the forwarded context? Seems like this should have been broken on all sections, but I must be missing something.

Short Answer: No, not just Section B.

Long Answer: They all do the same `rootFormData ?? value ?? {}` pattern. Section B was more sensitive to missing root data: it expects `activity_line_items` and `total_budget_categories` **directly from the root form data** and lacks the normalization logic used in Sections A, C, and E; when `formContext.rootFormData` is not provided or the widget receives only a partial value, Section B cannot safely reconstruct the data shape and ends up with undefined rows.

I have an [enhancement ticket](https://github.com/HHS/simpler-grants-gov/issues/8818):

The enhancement for Section B to be more stable like Sections A/C/E, we would introduce a `normalizeSectionBValue() `layer that can accept multiple incoming shapes (root object, array-only, or indexed-object forms) and always return a predictable `{ items, totals }` structure with safe defaults. This would allow Section B to degrade gracefully when the widget receives partial value data.

 
> can you add test steps?

Certainly!

Issue #8813, opened 2026-03-04
Skip flaky test in Firefox


mdragon:
### Summary

The test "should refresh and retain filters in a new tab" is failing randomly in Firefox. Skip that test in FF for now to allow the E2E tests to be stable and trusted by developers.

### Acceptance criteria

- [ ] Test is skipped in FF

Issue #8444, opened 2026-02-09
Split "should refresh and retain filters in a new tab" test into manageable pieces


doug-s-nava:
### Summary

The "should refresh and retain filters in a new tab" e2e test [located here](https://github.com/HHS/simpler-grants-gov/blob/03618db92f6d32c288fcdbd52f1034cb3ae3c96f/frontend/tests/e2e/search/search.spec.ts#L45) is attempting to test a lot of things at once. Many of the things it is testing are not otherwise tested outside of this one test, and so we currently relying on this one test for much of our basic search testing.

We should take the pieces of functionality being tested here, and split them into individual tests. Pieces to split out:

* updated sort order selection is maintained in UI on refresh
* updated search query is maintained in the UI on refresh
* updated agency filter is maintained in UI on refresh
* updated funding instrument filter is maintained in UI on refresh
* updated eligibility filter is maintained in UI on refresh
* updated category filter is maintained in UI on refresh

**Note:** the existing test is still potentially valuable as a full end-to-end experience so we should not fully delete it, but wait until we have the ability to include it in a non-blocking context before re-enabling.

**Note:** a goal of this ticket is to make our search tests more reliable. However, 100% reliability should not be a blocker for this work. If the new tests do not fully solve the problem of more reliably testing basic search behavior, we will make a follow up ticket to investigate and continue to improve reliability.

### Acceptance criteria

- [ ] current "should refresh and retain filters in a new tab" test is skipped
- [ ] all functionality tested in the existing "should refresh and retain filters in a new tab" test is handled by a group of smaller, more targeted tests

Bhavna-Ramachandran:
All PRs for this task has been merged and the associated sub-tasks are closed. Hence marking this ticket as closed.

PR #8815, opened 2026-03-04
[Issue #8813] Test flaky, skip for now

mdragon:
## Summary

Work for #8813 

## Changes proposed

Skips the test that is randomly failing

## Validation steps

Test "should refresh and retain filters in a new tab" is skipped.


mdragon:
Handling this via #8444 instead.

PR #8816, opened 2026-03-04
[Issue #8106] Workflow event API sends message to SQS

### LINKED ISSUE #8106: Make workflow event API send events to SQS queue
### Summary

After all validations and auth have been checked, we want the workflow event API to send a request to our SQS queue. The components for doing this have already been built, we want to:
* Build the `WorkflowEvent` pydantic object, most of this is already done, but we'll have to add the event ID + user ID to the request we take into the API.
* Convert the WorkflowEvent into JSON - use Pydantic's built-in ability to do this - https://docs.pydantic.dev/latest/concepts/serialization/#serializing-data
* The dict created can be passed to the SQS client's method for sending an event - the workflow queue URL can be loaded from the SQS config

### Acceptance criteria

- [ ] When all validation/authZ passes, an event is sent to SQS
- [ ] Tests verify events are written to the queue & the JSON within can be converted back into a `WorkflowEvent`

chouinar:
## Summary
Fixes #8106

## Changes proposed
* At the end of the event API, send the message to SQS
* A few adjustments to the SQS setup to be a bit easier to work with
* Moved an existing json serializer to a central location as we need it for SQS (currently used for logging)
* Fixed a duplicate env var for enabling the workflow API endpoints 

## Context for reviewers
Sending the event is pretty simple, just needed to make the workflow event and use Pydantic to make it into a dict. Had to fix the dict->json logic in the SQS client to handle UUIDs and a few other types - we already had a util for this for logging, so moved it to be reusable.

## Validation steps

Find an opportunity, you'll need the ID for the request

Can call this endpoint via the openapi docs, use the user API key of `internal_workflow_user_key`

```json
{
  "start_workflow_context": {
    "entity_id": "77f61f5e-caed-4722-99a0-51063e1e4ba4",
    "entity_type": "opportunity",
  }
}
```

This should give a success, and you should be able to query the localstack sqs by doing:

```shell
aws --endpoint-url http://localhost:4566 sqs receive-message --queue-url http://localhost:4566/000000000000/local_workflow_queue
```

This gave me:
Screenshot 2026-03-04 at 4 26 50 PM


api/src/adapters/aws/sqs_adapter.py
```diff
@@ -57,7 +59,7 @@ def get_boto_sqs_client(
     if session is None:
         session = get_boto_session()
 
-    return session.client("sqs", **params)
+    return session.client("sqs", region_name=sqs_config.aws_region, **params)
```
chouinar, line 62: This fixed a weird issue where it wasn't finding the queue because it was defaulting to us-west-2.

api/src/adapters/aws/sqs_adapter.py
```diff
@@ -7,13 +7,15 @@
...
 from src.util.env_config import PydanticBaseEnvConfig
+from src.util.json_util import json_encoder
 
 logger = logging.getLogger(__name__)
 
 
 class SQSConfig(PydanticBaseEnvConfig):
     workflow_queue_url: str = Field(alias="WORKFLOW_QUEUE_URL")
-    s3_endpoint_url: str = Field(alias="S3_ENDPOINT_URL")
+    s3_endpoint_url: str | None = Field(alias="S3_ENDPOINT_URL", default=None)
```
kkrug, line 17: Should we maybe rename this env var for clarity since it is used for both s3 and sqs stuff? If it would affect too many other things it's fine to keep it as is, just a thought.

api/src/services/workflows/send_workflow_event.py
```diff
@@ -0,0 +1,16 @@
...
+from src.workflow.event.workflow_event import WorkflowEvent
+
+logger = logging.getLogger(__name__)
+
+
+def send_workflow_event_to_queue(workflow_event: WorkflowEvent) -> None:
+    """Send a workflow event to the workflow SQS queue."""
+    config = SQSConfig()
+    sqs_client = SQSClient(queue_url=config.workflow_queue_url)
+    sqs_client.send_message(workflow_event.model_dump())
```
kkrug, line 13: I know you changed the `send_message` function to do `message_body_str = json.dumps(message_body, default=json_encoder)`, but could you just do `sqs_client.send_message(workflow_event.model_dump(mode='json'))` to avoid the encoder altogether?

api/src/services/workflows/ingest_workflow_event.py
```diff
@@ -74,17 +79,20 @@ def ingest_workflow_event(
       - Workflow exists and is active (404 if missing, 422 if inactive)
       - Event is valid for the workflow (422 if not)
     """
-    event_id = uuid.uuid4()
-    event_type = json_data.get("event_type")
 
-    add_extra_data_to_current_request_logs({"event_id": event_id, "event_type": event_type})
+    # Construct the
```
kkrug: Just an unfinished comment

api/src/adapters/aws/sqs_adapter.py
chouinar: I added a follow-up ticket to fix the naming inconsistency here https://github.com/HHS/simpler-grants-gov/issues/8826

api/src/services/workflows/send_workflow_event.py
chouinar: Both would work, went with this approach as it makes the SQS client more capable.

PR #8817, opened 2026-03-04
Aoy test suite sflll v2 form

aoysimmons-GH:
## Summary

Fixes / Work for #8370 

## Changes proposed

1. Added **[test-lifecycle-helpers.ts]** to support .spec files that run multiple scenarios, ensuring the page is closed after each test to prevent resource leaks.

2. Added **[test-select-dropdown-utils.ts]** to select dropdown options by value first (for dynamic dropdowns with explicit values), with fallback to visible label text.

3. Added reusable utilities for UI verification, required-field validation, and form filling:
- **[test-verify-ui-utils.ts]**
- **[test-validate-and-fill-field-utils.ts]**
- **[test-form-fill-utils.ts]**

4. Added reusable SFLLL form utilities for UI verification and required-field validation across test types:
- **[user-interface-sflll-form-utils.ts]**
- **[validation-sflll-form-utils.ts]**
- **[fill-sflll-form-utils.ts]**

5. Added **[test-suite-sflll-v2-form.spec.ts]** to run SFLLL form tests by type:
- Smoke: UI-only checks
- Regression: UI, validation, and data-fill coverage

### DIFF
--- a/frontend/tests/e2e/apply/sflll-form.spec.ts
+++ b/frontend/tests/e2e/apply/sflll-form.spec.ts
@@ -3,7 +3,7 @@ import { createSpoofedSessionCookie } from "tests/e2e/loginUtils";
 import playwrightEnv from "tests/e2e/playwright-env";
 import { createApplication } from "tests/e2e/utils/create-application-utils";
 import { fillSflllForm } from "tests/e2e/utils/forms/fill-sflll-form-utils";
-import { validateSflllUI} from "tests/e2e/utils/forms/user-interface-sflll-form-utils";
+import { validateSflllUI } from "tests/e2e/utils/forms/user-interface-sflll-form-utils";
 import { validateSflllFormRequiredFields } from "tests/e2e/utils/forms/validation-sflll-form-utils";
 import { ensurePageClosed } from "tests/e2e/utils/lifecycle-helpers";
 
@@ -12,7 +12,6 @@ const OPPORTUNITY_ID = "f7a1c2b3-4d5e-6789-8abc-1234567890ab";
 const OPPORTUNITY_URL = `/opportunity/${OPPORTUNITY_ID}`;
 
 test.describe("Test Suite - SFLLL Form", () => {
-
   // Ensure the page is closed after each test
   test.afterEach(async ({ page }) => {
     await ensurePageClosed(page);
@@ -24,7 +23,9 @@ test.describe("Test Suite - SFLLL Form", () => {
   // Set a timeout of 120 seconds for the test
   test.setTimeout(120000);
 
-  test("Should fill all required form fields with valid test data - SFLLL Form", async ({ page }, testInfo) => {
+  test("Should fill all required form fields with valid test data - SFLLL Form", async ({
+    page,
+  }, testInfo) => {
     await page.goto(`${baseUrl}${OPPORTUNITY_URL}`, {
       waitUntil: "load",
       timeout: 30000,
@@ -34,7 +35,9 @@ test.describe("Test Suite - SFLLL Form", () => {
     await fillSflllForm(testInfo, page);
   });
 
-  test("Form fields contain expected labels and descriptions - SFLLL Form", async ({ page }, testInfo) => {
+  test("Form fields contain expected labels and descriptions - SFLLL Form", async ({
+    page,
+  }, testInfo) => {
     await page.goto(`${baseUrl}${OPPORTUNITY_URL}`, {
       waitUntil: "load",
       timeout: 30000,
@@ -44,7 +47,9 @@ test.describe("Test Suite - SFLLL Form", () => {
     await validateSflllUI(testInfo, page);
   });
 
-  test("Required form fields validate and show error messages - SFLLL Form", async ({ page }, testInfo) => {
+  test("Required form fields validate and show error messages - SFLLL Form", async ({
+    page,
+  }, testInfo) => {
     await page.goto(`${baseUrl}${OPPORTUNITY_URL}`, {
       waitUntil: "load",
       timeout: 30000,
@@ -54,7 +59,9 @@ test.describe("Test Suite - SFLLL Form", () => {
     await validateSflllFormRequiredFields(testInfo, page);
   });
 
-  test("Smoke test: Form UI, fill, and navigation complete successfully - SFLLL Form", async ({ page }, testInfo) => {
+  test("Smoke test: Form UI, fill, and navigation complete successfully - SFLLL Form", async ({
+    page,
+  }, testInfo) => {
     await page.goto(`${baseUrl}${OPPORTUNITY_URL}`, {
       waitUntil: "load",
       timeout: 30000,
@@ -65,7 +72,9 @@ test.describe("Test Suite - SFLLL Form", () => {
     await fillSflllForm(testInfo, page);
   });
 
-  test("Regression test: Form UI, validation, and fill complete successfully - SFLLL Form", async ({ page }, testInfo) => {
+  test("Regression test: Form UI, validation, and fill complete successfully - SFLLL Form", async ({
+    page,
+  }, testInfo) => {
     await page.goto(`${baseUrl}${OPPORTUNITY_URL}`, {
       waitUntil: "load",
       timeout: 30000,


frontend/tests/e2e/apply/sflll-form.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+
+test.describe("Test Suite - SFLLL Form", () => {
+  test.beforeEach(async ({ context }) => {
+    await createSpoofedSessionCookie(context);
+  });
+  // Set a timeout of 120 seconds for the test
+  test.setTimeout(120000);
+
+  // Ensure the page is closed after each test to prevent resource leaks
+  ensurePageClosedAfterEach(test);
```
doug-s-nava: this only needs to be run once per file or describe block https://playwright.dev/docs/api/class-test#test-after-each

I am also not sure what the behavior is when adding multiple after each blocks. My guess is that it runs all of them, but I think we should avoid it in any case. Can you refactor this utility to be something that runs within a defined block rather than defining the block itself?

frontend/tests/e2e/apply/sflll-form.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  test("Fill - SFLLL Form", async ({ page }, testInfo) => {
+    await page.goto(`${baseUrl}${OPPORTUNITY_URL}`, {
+      waitUntil: "load",
+      timeout: 30000,
+    });
+
+    await createApplication(page, OPPORTUNITY_URL, testOrgLabel);
+    await fillSflllFormUtils(testInfo, page);
```
doug-s-nava: it doesn't look like there are any expects in here. What is this test testing?

frontend/tests/e2e/apply/sflll-form.spec.ts
```diff
@@ -0,0 +1,88 @@
+import { test } from "@playwright/test";
```
doug-s-nava, line 1: not sure about this file name. We can remove `test-suite`. What does the v2 mean?

frontend/tests/e2e/utils/verify-ui-utils.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    // Normalize whitespace in both expected and actual text for comparison
+    // This handles cases where DOM has multiple spaces or line breaks
+    const normalizeWhitespace = (text: string) =>
+      text.replace(/\s+/g, " ").trim();
+
+    const normalizedActual = normalizeWhitespace(elementText || "");
+    const normalizedExpected = normalizeWhitespace(field.expectedText);
+
+    if (!normalizedActual.includes(normalizedExpected)) {
```
doug-s-nava: why not use `expect(normalizedActual).toContain(normalizedExpected)`

frontend/tests/e2e/apply/sflll-form.spec.ts
doug-s-nava: functions in general should be named based on what they are doing rather than what they are. For example, this function could be `fillSflllForm` and `userInterfaceSflllFormUtils` could be `validateSflllUI`

frontend/tests/e2e/apply/sflll-form.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+    });
+
+    await createApplication(page, OPPORTUNITY_URL, testOrgLabel);
+    await fillSflllFormUtils(testInfo, page);
+  });
+
+  // Ensure the page is closed after each test to prevent resource leaks
+  ensurePageClosedAfterEach(test);
+
+  test("UI - SFLLL Form", async ({ page }, testInfo) => {
```
doug-s-nava: can we be more explicit about what is being tested? For example `Form fields contain expected labels and descriptions - SFLLL Form`

frontend/tests/e2e/utils/validate-and-fill-field-utils.ts
```diff
@@ -0,0 +1,101 @@
...
+    // Wait longer for validation errors to appear
+    await page.waitForTimeout(2000);
+
+    // Check page state before clicking error link
+    if (page.isClosed()) {
+      throw new Error("Page was closed after save click");
+    }
+
+    // Try to find and click the error link if it exists
+    const errorLink = page.getByRole("link", { name: field.errorLinkText });
```
doug-s-nava, line 41: I like this idea, but in some cases a field might have multiple possible error cases - for example a required field that expects a number could error either if not filled in, or if a string is provided - and each would have a different message. I think for now we can assume that the form does not have any errors, and we can test the case of filling a form with errors separately later

frontend/tests/e2e/utils/validate-and-fill-field-utils.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    // Fill the field
+    if (field.isDropdown) {
+      await selectDropdownByValueOrLabel(page, field.testId, field.value);
+    } else {
+      await fieldLocator.fill(field.value);
+    }
+
+    // Save again after filling the field
+    await page.getByTestId("apply-form-save").click();
```
doug-s-nava, line 78: what is your thinking around saving after every field? This could get pretty time consuming on a long form. Seems to make more sense to:
* fill each field
* save
* validate that there are no errors

frontend/tests/e2e/apply/sflll-form.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+    });
+
+    await createApplication(page, OPPORTUNITY_URL, testOrgLabel);
+    await validationSflllFormUtils(testInfo, page);
+  });
+
+  // Ensure the page is closed after each test to prevent resource leaks
+  ensurePageClosedAfterEach(test);
+
+  test("Smoke test - SFLLL Form", async ({ page }, testInfo) => {
```
doug-s-nava: what is the benefit of this and the next test? What do we gain from these that we wouldn't get by running the scenarios individually?

frontend/tests/e2e/utils/form-fill-utils.ts
```diff
@@ -0,0 +1,70 @@
...
+// ============================================================================
+
+/**
+ * Fills a single form field with error handling and test reporting
+ *
+ * @param testInfo - Playwright TestInfo for attaching test results
+ * @param page - Playwright Page object
+ * @param field - Field definition including testId/selector, value, type, and section
+ */
+export async function fillField(
```
doug-s-nava, line 38: for this file and the fill utils, there is nothing form specific about the functionality here - can we make general testing utils, and separate the fixture data into its own fixture files?

frontend/tests/e2e/utils/form-fill-utils.ts
```diff
@@ -0,0 +1,70 @@
+/**
```
doug-s-nava, line 1: please remove "test" from filenames

frontend/tests/e2e/apply/sflll-form.spec.ts
aoysimmons-GH: As per the existing form the current user interface of this SFLLL form is version 2.
I updated the name from 'test-suite-sflll-v2-form.spec.ts' to 'sflll-form.spec.ts'

Ref: api/src/task/opportunities/build_automatic_opportunities.py
Image

frontend/tests/e2e/apply/sflll-form.spec.ts
aoysimmons-GH: Thanks for pointing that out. I refactored the utility so it no longer registers its own afterEach. Instead it just handles the page-closing logic, and the afterEach hook is defined within the test/describe block.

frontend/tests/e2e/apply/sflll-form.spec.ts
aoysimmons-GH: The intent of this test is to verify that the SFLLL form utility works correctly by completing the form and confirming there are no validation errors in the happy path. If you think this doesn't provide enough value as a test case, I can remove or refactor it.

aoysimmons-GH: Good point. I renamed all three functions and updated the corresponding calls.

frontend/tests/e2e/apply/sflll-form.spec.ts
aoysimmons-GH: Got it. I updated.

frontend/tests/e2e/utils/form-fill-utils.ts
aoysimmons-GH: Got it. Removed.

frontend/tests/e2e/utils/validate-and-fill-field-utils.ts
aoysimmons-GH: I want to keep the current approach because this is a dedicated validation test that needs to verify each field's validation rules independently, test the complete user experience of error feedback, and provide granular failure diagnostics when issues occur. The thorough approach is justified for validation-critical testing, while faster filling methods can be used in smoke and regression tests that don't focus on validation behavior.

frontend/tests/e2e/utils/validate-and-fill-field-utils.ts
aoysimmons-GH: That makes sense. This scenario only verifies the required field validation messages, and other validation cases can be tested separately. I also renamed the function from validateAndFillField to validateSflllFormRequiredFields to better reflect its purpose.

frontend/tests/e2e/utils/form-fill-utils.ts
aoysimmons-GH: This is a great idea!

I did a bit of research on it. I think only the **ENTITY_DATA** (test data) should be moved to **sflll-fixture.ts**, since it’s duplicated across three files and represents shared test data. The other constants seem to be form-specific.

Another option is to move the constants into a **Page Object** for the form to improve encapsulation, reduce duplication, and make the tests easier to read and maintain.

I think we could combine both approaches:

**Page Object (sflll-form.page.ts)** – structure, selectors, and methods

**Fixture (sflll-fixture.ts)** – reusable test data

I’d need some time to refactor this. Happy to discuss more in the meeting.

aoysimmons-GH (commented):
I’m interested in refactoring this using the [Fixture and Page Object], but I’m not sure if the timeline allows for it in this PR. We can discuss and decide if it should be done now or in a follow-up.

Thanks for the review and the suggestions. I really appreciate the feedback.

frontend/tests/e2e/utils/forms/fill-sflll-form-utils.ts
```diff
@@ -0,0 +1,489 @@
...
+    middleName: "Middle Name for test Signature",
+    lastName: "Last Name for test Signature",
+    suffix: "SuffixTest",
+    title: "TitleTest",
+    phone: "9999999999",
+  },
+};
+
+// ============================================================================
+// Field Definitions for Form Filling
```
Bhavna-Ramachandran, line 85: Suggestion/Improvement: Currently all fields are located using 'selector' or 'testid. Consider adding dynamic field locators / fallbacks so the test does not get flaky. 
Example below:
  const titleInputs = page.locator(
    'input[name*="title" i], input[placeholder*="title" i], textarea[name*="title" i], textarea[placeholder*="title" i]',
  );

you can check e2e/utils/forms/fill-sf424b-form-utils.ts for the dynamic locator logic

frontend/tests/e2e/utils/forms/fill-sflll-form-utils.ts
```diff
@@ -0,0 +1,489 @@
...
+  try {
+    // Extract ApplicationURL and attach to test report
+    const applicationURL = page.url();
+    await testInfo.attach("fillSflllForm-applicationURL", {
+      body: `Application URL: ${applicationURL}`,
+      contentType: "text/plain",
+    });
+
+    // Navigate to the SF-LLL form from the application overview page
+    await page.getByRole("link", { name: ENTITY_DATA.form.name }).click();
```
Bhavna-Ramachandran, line 454: Instead of getByRole, I suggest to use this utility function frontend/tests/e2e/utils/forms/form-link-utils.ts , this is flexible.
Above line would look like - await getFormLink(page, ENTITY_DATA.form.name).click();

frontend/tests/e2e/utils/forms/user-interface-sflll-form-utils.ts
```diff
@@ -0,0 +1,447 @@
...
+  try {
+    // Extract ApplicationURL and attach to test report
+    const applicationURL = page.url();
+    await testInfo.attach("validateSflllUI-applicationURL", {
+      body: `Application URL: ${applicationURL}`,
+      contentType: "text/plain",
+    });
+
+    // Navigate to the SF-LLL form from the application overview page
+    await page.getByRole("link", { name: ENTITY_DATA.form.name }).click();
```
Bhavna-Ramachandran, line 414: Instead of getByRole, I suggest to use this utility function frontend/tests/e2e/utils/forms/form-link-utils.ts , this is flexible.
Above line would look like - await getFormLink(page, ENTITY_DATA.form.name).click();

frontend/tests/e2e/utils/forms/validation-sflll-form-utils.ts
```diff
@@ -0,0 +1,235 @@
...
+  try {
+    // Extract ApplicationURL and attach to test report
+    const applicationURL = page.url();
+    await testInfo.attach("validateSflllFormRequiredFields-applicationURL", {
+      body: `Application URL: ${applicationURL}`,
+      contentType: "text/plain",
+    });
+
+    // Navigate to the SF-LLL form from the application overview page
+    await page.getByRole("link", { name: ENTITY_DATA.form.name }).click();
```
Bhavna-Ramachandran, line 202: Instead of getByRole, I suggest to use this utility function frontend/tests/e2e/utils/forms/form-link-utils.ts , this is flexible.
Above line would look like - await getFormLink(page, ENTITY_DATA.form.name).click();

frontend/tests/e2e/utils/forms/fill-sflll-form-utils.ts
```diff
@@ -0,0 +1,489 @@
...
+
+    // Fill all fields using the field definitions array
+    for (const field of FILL_FIELDS) {
+      await fillField(testInfo, page, field);
+    }
+
+    // Add small delay to allow form to stabilize after all entries
+    await page.waitForTimeout(500);
+
+    // Click Save Button
```
Bhavna-Ramachandran, line 471: the below block:
// Click Save Button
await page.getByTestId("apply-form-save").click();

// Verify No Validation Errors
await page
  .getByText("No errors were detected")
  .waitFor({ state: "visible", timeout: 10000 });

can be replaced with : await saveForm(page);
This is the existing save form utils, e2e/utils/forms/save-form-utils.ts

Issue #8437, opened 2026-02-06
[BE] Implement including saved_to_organizations field to saved opportunities response 


freddieyebra:
### Summary

If any organization ids are specified in the `SavedOpportunityFilterParams`, we need to include a new `"saved_to_organizations"` property on the opportunities in the response. `"saved_to_organizations"` will contain other organizations that the user is a member of and has saved the opportunity to.

Success response:
```json
{
  "data": [
    {
      "opportunity_id": "942e4567-e89b-12d3-a456-426614174000",
      "opportunity_status": "posted",
      "opportunity_title": "Opportunity Title",
      "saved_to_organizations": [
        {
          "organization_name": "Org Name",
          "organization_id": "123e4567-e89b-12d3-a456-426614174000"
        },
        {
          "organization_name": "Org Name 2",
          "organization_id": "891e4567-e89b-764d-a456-426614174000"
        }
      ],
      "summary": {
        "close_date": "2026-08-01",
        "is_forecast": false,
        "post_date": "2026-01-01"
      }
    }
  ],
  "message": "Success",
  "status_code": 200
}
```

### Acceptance Criteria

- [ ] The above property is included in the list saved opportunity response if organization ids passed in filter
- [ ] Unit tests pass

chelsea-nava:
Proposed design is [this page](https://www.figma.com/design/bQUTJCMj0JYbAAnJQ4p4tX/Grantee-Designs-%7C-Quad-5-%7C-Simpler-Grants-2026?node-id=2716-44016&t=tujK3QI4UXs1zzcf-1)


Would prefer to have a single page with ability to filter between "all saved opportunities". We would use the existing dropdown pattern from this page, to include a default setting of "all saved opportunities" and allow users to change this to focus on a single org from their list OR their privately saved opportunities.

If we follow this cadence, any/all mentions of 'view saved opportunities' will result in the user landing on this Saved Opportunities page. If a user enters via an organization-specific entry (e.g. Organization page - view saved opportunities) we should pre-filter the page with that organization's details. 



If this is NOT a possibility, we can have individual pages for each org as well as the individually saved opportunities, but I don't see that as the best long-term solution. (Those designs also are in the figma w/routes).

chelsea-nava:
@freddieyebra 

**Ways to access this page:** 

- [ ] Workspace (nav dropdown) —Saved Opportunities —Saved Opportunities (default all)
[...]

PR #8823, opened 2026-03-05
[Issue #8437] Add organization details to saved opportunity endpoint

babebe:
## Summary

Fixes / Work for #8437  

## Changes proposed

- Updated get_saved_opportunities endpoint to include a new property saved_to_organizations in the opportunity response.
- saved_to_organizations contains all organizations (that the user is a member of) which have saved the opportunity.
- Behavior:
   - User-only saved opportunities: saved_to_organizations is omitted.
   - Organization saved opportunities (specific orgs or all saved opportunities): saved_to_organizations is included.
- If no org has saved it, it returns an empty list [].

## Context for reviewers

The saved_to_organizations field helps frontends distinguish between user-only saved opportunities and organization-saved opportunities.

## Validation steps
Covered in tests

### DIFF
--- a/api/src/services/users/get_saved_opportunities.py
+++ b/api/src/services/users/get_saved_opportunities.py
@@ -173,8 +173,7 @@ def _build_saved_union_subquery(
 def _enrich_with_saved_organizations_sql(
     db_session: db.Session,
     opportunities: Sequence[Opportunity],
-    org_ids_to_use: list[uuid.UUID] | None,
-) -> Sequence[Opportunity]:
+) -> None:
     """
     Enrich paginated Opportunity objects with `saved_to_organizations`.
 
@@ -182,12 +181,9 @@ def _enrich_with_saved_organizations_sql(
     -  If the request is for **user-saved opportunities only**, the `saved_to_organizations` field is omitted.
     - If the request involves organizations the `saved_to_organizations` field is always included.
     """
-    if not org_ids_to_use:
-        return opportunities
-
     opp_ids = [opp.opportunity_id for opp in opportunities]
 
-    rows = db_session.execute(
+    result = db_session.execute(
         select(OrganizationSavedOpportunity.opportunity_id, Organization)
         .join(
             Organization,
@@ -198,13 +194,12 @@ def _enrich_with_saved_organizations_sql(
 
     # Build a mapping: opportunity_id to list of Organization objects
     saved_orgs_map: dict[uuid.UUID, list[Organization]] = {}
-    for opp_id, org in rows:
+    for opp_id, org in result:
         saved_orgs_map.setdefault(opp_id, []).append(org)
 
     # Assign attribute to each organization saved opportunity
     for opp in opportunities:
-        setattr(opp, "saved_to_organizations", saved_orgs_map.get(opp.opportunity_id, []))  # noqa: B010
-    return opportunities
+        opp.saved_to_organizations = saved_orgs_map.get(opp.opportunity_id, [])  # type: ignore[attr-defined]
 
 
 def get_saved_opportunities(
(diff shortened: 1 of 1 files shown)


api/src/api/opportunities_v1/opportunity_schemas.py
```diff
@@ -584,6 +594,9 @@ class SavedOpportunityResponseV1Schema(Schema):
     )
 
     summary = fields.Nested(SavedOpportunitySummaryV1Schema())
+    saved_to_organizations = fields.Nested(
+        SavedOpportunityOrganizationSchema, many=True, allow_none=True
+    )
```
chouinar: I'm not sure if `many=True` will cause any different behavior from `fields.List` - it might on the request side, but just in the interest of consistency, can we use list?

```suggestion
    saved_to_organizations = fields.List(
        fields.Nested(SavedOpportunityOrganizationSchema), allow_none=True
    )
```

api/src/services/users/get_saved_opportunities.py
```diff
@@ -170,9 +171,57 @@ def _build_saved_union_subquery(
...
+    for opp in opportunities:
+        saved_to_orgs = [
+            {
+                "organization_id": str(save.organization_id),
+                "organization_name": (
+                    save.organization.organization_name if save.organization else None
+                ),
+            }
+            for save in opp.saved_opportunities_by_organizations
+            if org_ids_to_use and save.organization_id in org_ids_to_use
```
chouinar: I don't think iterating over this relationship is what we'd want to do. This relationship is every organization that has saved a given opportunity, it could be hundreds or thousands.

I think we'd either want to do a query for this or find a way to put it in the current query. Since the current query is pretty complex, I think the easier thing would be to do a second query to do this instead.

api/src/services/users/get_saved_opportunities.py
```diff
@@ -170,9 +171,57 @@ def _build_saved_union_subquery(
...
+    response: list = []
+
+    for opp in opportunities:
+        saved_to_orgs = [
+            {
+                "organization_id": str(save.organization_id),
+                "organization_name": (
+                    save.organization.organization_name if save.organization else None
+                ),
+            }
```
chouinar: The organization class has the ID + has a property for organization name. Shouldn't need to convert to a dict, a list of orgs would be fine here, marshmallow will convert it just fine.

```suggestion
        saved_to_orgs = [
            organization
```

api/src/services/users/get_saved_opportunities.py
```diff
@@ -170,9 +171,57 @@ def _build_saved_union_subquery(
...
+            for save in opp.saved_opportunities_by_organizations
+            if org_ids_to_use and save.organization_id in org_ids_to_use
+        ]
+        base: dict[str, Any] = {
+            "opportunity_id": str(opp.opportunity_id),
+            "opportunity_title": opp.opportunity_title,
+            "opportunity_status": opp.opportunity_status,
+            "summary": opp.summary,
+        }
```
chouinar: An easier approach that we've done elsewhere to enrich a DB model. Leave it a db model, and do:

```py
setattr(opportunity, "field_i_want_to_add", value)
```

Technically this is the same as `opportunity.field_i_want_to_add = value` but works around mypy complaining about that not being a value.

This won't cause any DB issues, it's not a mapped column, it'll just be there, but it works for turning it into the response.

api/src/services/users/get_saved_opportunities.py
babebe: Updated

api/src/services/users/get_saved_opportunities.py
babebe: When using setattr ruff also complains, using direct assignment and suppressing the mypy-related complaint

Issue #8828, opened 2026-03-05
Upgrade npm packages that have fixes


mdragon:
### Summary

`npm ci` was flagging some packages that had fixes available. Run `npm audit fix` to pick up these "easy fixes."

### Acceptance criteria

Smoke test the FE after upgrades

PR #8830, opened 2026-03-05
[Issue #8828] Update some npm packages with available fixes

mdragon:
## Summary

Work for #8828 

## Changes proposed

`npm ci` was suggesting there was packages with fixes available. Installed the "clean" ones by running `npm audit fix`
## Validation steps

1. Checkout branch
2. Run `npm ci`
3. Start the BE services in another terminal with `make run-logs`
4. Start the app with `npm run start:dev`
5. Validate that various functionalities in the site continue to work

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -542,12 +542,11 @@
       }
     },
     "node_modules/@babel/runtime": {
-      "version": "7.26.0",
+      "version": "7.28.6",
+      "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.28.6.tgz",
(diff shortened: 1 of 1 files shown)


frontend/package-lock.json
```diff
@@ -16258,16 +16251,6 @@
         "node": ">=10"
       }
     },
-    "node_modules/serialize-javascript": {
```
doug-s-nava, line 16261: nice, this was problematic re vulns recently

mdragon:
> However, what do we do to make sure nothing broke? Especially if we want to make this a more repeatable process we should probably have a set of manual regression tests to run through

My thinking was E2E test give us a first backstop. Additionally I will test everything I can think of locally and then maybe we loop in Product for some Staging testing?

On one hand we do this somewhat regularly without extensive testing in response to the vuln scans, but that's a little more targeted and usually fewer packages. On the other hand, it's not a good look to opt into upgrades and then break something important.

doug-s-nava (approved):
ran a regression locally on search and apply functionality, everything seems good as far as I can tell

✅ ✅ ✅ ✅

Issue #8600, opened 2026-02-19
Opportunity Summary CRUD Endpoint - Create


glenblosser-nava:
### Summary

As a grantor, I need to be able to create an Opportunity Summary. In order to do so, I must provide an opportunity ID. Each Opportunity can have no more than 1 of each type of summary (forecast or non-forecast).

### Acceptance criteria

Create Endpoint functional as outlined in [the Phase 2 Tech Spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2613215233/Phase+2+Feature+Build+Out#Endpoints)

dghazvini:
@chouinar 

Hey Michael, just wanted to run this ticket by you before beginning work on it. If anything on the Phase 2 tech spec linked above for this ticket looks off please call it out, I'll be referencing that in configuring the endpoint. Just some questions I had:

-From my understanding with this new endpoint we'll first call `get_opportunity` in the service layer passing in the opp id as a parameter. Will call `verify_access` with `edit_opportunity` privilege to verify AuthZ. From there, should we query the DB to check for an existing summary in OpportunitySummary using Opp Id as our filter? Or we can just check `all_opportunity_summaries`? 

-After the Opportunity Summary creation is done, do we need to make any updates to the Opportunity table, or is the association just between opportunity_id in both tables? I noticed `current_opportunity_summary` and wasn't sure if that needs updating.

-Validation Rules - Are there any specific validation rules for the opp summary fields beyond what's in the model we need to take into account? I noticed there's `award_floor` and `award_cieling` fields. Same for `post_date` and `close_date`.

-Tech Spec mentions returning 422 if a summary of the same type already exists; I'm assuming `is_forecast` is the field we'd check. Any other expected validation error responses?

chouinar:
@dghazvini 

Yes, effectively any of the endpoints besides create opportunity should always:
* Fetch the opportunity, 404 if not found
* Verify they have the view opportunity privilege
* Then verify they have access to whatever the endpoint needs (so in this case, the edit privilege)
* Then do its own logic

Since you'll know whether it's the forecast/non-forecast from the request, I'd say the easiest thing is to use the two properties you added since that's what those are.

[...]

Issue #8006, opened 2026-01-15
Opportunity Table CRUD Endpoint - Read


brianlee-nava:
### Summary

**Background and Context**

There are 4 main tables within the Opportunity Data Model that Plan and Publish will need to work with. Those are: Opportunity, Opportunity Attachment, Opportunity Summary, Opportunity Assistance Listing.

In order to make Opportunity Publishing work in Simpler Grants Management, we need to rebuild endpoints. The basic CRUD Create, Read, Update, Delete operations for each table account for 16 endpoints, with additional simple endpoints for publishing and querying the Opportunity Assistance Listing table. **This ticket is scoped for building the Read endpoint for the Opportunity table. We need to expand the endpoint to accept an user ID and return all opportunities for that user GROUPED by Agency. ** 

**User Story**
As a grantor system
I want to retrieve an Opportunity by ID
So that I can view its current state

### Acceptance criteria

Given an existing Opportunity ID
When the read Opportunity endpoint is called
Then the Opportunity record is returned
And the response returns a 200 status

Given a non existent Opportunity ID
When the read Opportunity endpoint is called
Then the response returns a 404 status

- [ ] User ID should be optional parameter
- [ ] Opportunities are grouped by Agency ID
- [ ] Update the API.grants.gov page

chouinar:
The current opportunity response looks like:
```json
{
   // various fields
   "summary": {
      ...
   }
}
```

Instead we should do:
```json
{
   // various fields
   "forecast_summary": {
      ...
   },
    "non_forecast_summary": {

    }
}
```

PR #8844, opened 2026-03-06
[Issue #8006] Opportunity Summary Create Endpoint

dghazvini:
## Summary

Work for #8600  

## Changes proposed

Implementing the Create Opportunity Summary endpoint as laid out in the [Opportunity Publishing MVP Phase 2: Feature Build Out Tech Spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2613215233/Phase+2+Feature+Build+Out).

## Context for reviewers

Endpoint Path: `POST /v1/opportunities/:opportunity_id/summary` - create a new opportunity summary

Parameters: `opportunity_id:uuid`

* Create Request:
{
  "legacy_opportunity_id": string,
  "is_forecast": boolean,
  "summary_description": string,
  "is_cost_sharing": boolean,
  "post_date": date,
  "close_date": date,
  "close_date_description": string, 
  "archive_date": date,
  "expected_number_of_awards": int,
  "estimated_total_program_funding": int,
  "award_floor": int,
  "award_ceiling": int,
  "additional_info_url": string,
  "additional_info_url_description": string,
  "funding_categories": string[],
  "funding_category_description": string,
  "funding_instruments": string[],
  "applicant_types": string[],
  "applicant_eligibility_description": string,
  "agency_contact_description": string,
  "agency_email_address": string, 
  "agency_email_address_description": string,
  "forecasted_post_date": date,
  "forecasted_close_date": date,
  "forecasted_close_date_description": date,
  "forecasted_award_date": date, 
  "forecasted_project_start_date": date,
  "fiscal_year": int
}

* Response
{
  "opportunity_summary_id": string,
  "legacy_opportunity_id": string,
  "is_forecast": boolean,
  "summary_description": string,
  "is_cost_sharing": boolean,
  "post_date": date,
  "close_date": date,
  "close_date_description": string, 
  "archive_date": date,
  "expected_number_of_awards": int,
  "estimated_total_program_funding": int,
  "award_floor": int,
  "award_ceiling": int,
  "additional_info_url": string,
  "additional_info_url_description": string,
  "funding_categories": string[],
  "funding_category_description": string,
  "funding_instruments": string[],
  "applicant_types": string[],
  "applicant_eligibility_description": string,
  "agency_contact_description": string,
  "agency_email_address": string, 
  "agency_email_address_description": string,
  "forecasted_post_date": date,
  "forecasted_close_date": date,
  "forecasted_close_date_description": date,
  "forecasted_award_date": date, 
  "forecasted_project_start_date": date,
  "fiscal_year": int
}

## Validation steps

- [x] 7 Unit Tests Added for expected responses/cases
- [x] Verified data persists in local DB
- [x] Verified expected responses on local testing


api/src/api/opportunities_v1/opportunity_schemas.py
```diff
@@ -597,3 +601,48 @@ class OpportunityVersionSchema(OpportunityV1Schema):
         fields.Nested(OpportunityVersionAttachmentSchema),
         metadata={"description": "List of attachments associated with the opportunity"},
     )
+
+
+class OpportunitySummaryCreateRequestV1Schema(OpportunitySummaryV1Schema):
+    legacy_opportunity_id = fields.Integer(
```
chouinar: I'd leave this out of the schema. A user wouldn't be setting it, only way this gets set in our system is by processes interacting with grants.gov.

api/src/api/opportunities_v1/opportunity_schemas.py
```diff
@@ -597,3 +601,48 @@ class OpportunityVersionSchema(OpportunityV1Schema):
         fields.Nested(OpportunityVersionAttachmentSchema),
         metadata={"description": "List of attachments associated with the opportunity"},
     )
+
+
+class OpportunitySummaryCreateRequestV1Schema(OpportunitySummaryV1Schema):
```
chouinar: The opportunity summary schema we have right now doesn't have validations for fields, that needs to be added for things like min/max lengths or values.

api/src/api/opportunities_v1/opportunity_routes.py
```diff
@@ -264,3 +265,30 @@ def opportunity_get(db_session: db.Session, opportunity_id: UUID) -> response.Ap
         opportunity = get_opportunity(db_session, opportunity_id)
 
     return response.ApiResponse(message="Success", data=opportunity)
+
+
+@opportunity_blueprint.post("/opportunities/<uuid:opportunity_id>/summary")
```
chouinar: Should this be under the grantor blueprint instead?

api/src/services/opportunities_v1/get_opportunity.py
```diff
@@ -64,3 +64,48 @@ def get_opportunity_by_legacy_id(db_session: db.Session, legacy_opportunity_id:
...
     return opportunity
+
+
+def _get_opportunity_including_drafts(
+    db_session: db.Session, where_clause: ColumnExpressionArgument[bool]
+) -> Opportunity | None:
+    stmt = (
+        select(Opportunity)
+        .where(where_clause)
+        .options(selectinload("*"))
```
chouinar: Don't use `*`, I know we do right now for GET opportunity, I want to change that. Put the specific selectinloads we want. You shouldn't need to do this anyways as `_get_opportunity_for_grantors` was already done that way. Reuse that.

api/src/services/opportunities_v1/opportunity_summaries.py
```diff
@@ -0,0 +1,158 @@
...
+        fiscal_year=request.fiscal_year,
+    )
+
+
+def create_opportunity_summary(
+    db_session: db.Session, opportunity_id: UUID, summary_data: dict, user: User
+) -> OpportunitySummary:
+    """Create a new opportunity summary"""
+    # Parse and validate the request data
+    request = OpportunitySummaryCreateRequest(**summary_data)
```
chouinar: Rather than do dict -> pydantic model -> opportunity summary, what if we just do:

```py
opportunity_summary = OpportunitySummary(
   opportunity=opportunity,
   **summary_data
)
```

We're not doing much with that pydantic model other than checking one or two values, can just pull those out of the dict.

api/src/services/opportunities_v1/opportunity_summaries.py
```diff
@@ -0,0 +1,158 @@
...
+
+    db_session.add(opportunity_summary)
+
+    # Handle relationships for loading
+    if request.funding_instruments:
+        for instrument in request.funding_instruments:
+            link_funding_instrument = LinkOpportunitySummaryFundingInstrument(
+                funding_instrument=instrument
+            )
+            opportunity_summary.link_funding_instruments.append(link_funding_instrument)
```
chouinar: This shouldn't be needed in a special way.

If you do:
```py
opportunity_summary = OpportunitySummary(..., funding_instruments=[FundingInstrument.X, FundingInstrument.Y])
```
That should work (same for the other two) as we setup a special association proxy relationship for those. That's how the factories work uneventfully, it just does that.

api/src/services/opportunities_grantor_v1/opportunity_summaries.py
```diff
@@ -0,0 +1,158 @@
...
+            link_applicant_type = LinkOpportunitySummaryApplicantType(applicant_type=applicant_type)
+            opportunity_summary.link_applicant_types.append(link_applicant_type)
+
+    logger.info(
+        f"Created {'forecast' if request.is_forecast else 'non-forecast'} opportunity summary",
+        extra={
+            "opportunity_id": str(opportunity_id),
+            "opportunity_summary_id": str(opportunity_summary.opportunity_summary_id),
+        },
+    )
```
chouinar, line 51: Put any variables in the extra of the log - makes it so it's easily queryable in New Relic. Also, never do `str(a uuid)` for logs - it's won't cause any issue, it's just not necessary as our logging setup does that automatically.

```suggestion
    logger.info(
        "Created opportunity summary",
        extra={
            "opportunity_id": opportunity_id,
            "opportunity_summary_id":.opportunity_summary.opportunity_summary_id,
            "is_forecast": is_forecast
        },
    )
```

api/src/services/opportunities_v1/opportunity_summaries.py
```diff
@@ -0,0 +1,158 @@
...
+) -> OpportunitySummary:
+    """Create a new opportunity summary"""
+    # Parse and validate the request data
+    request = OpportunitySummaryCreateRequest(**summary_data)
+
+    opportunity = get_opportunity_including_drafts(db_session, opportunity_id)
+
+    # Check if user has permission to view/edit opportunities for this agency
+    agency = opportunity.agency_record
+    verify_access(user, {Privilege.VIEW_OPPORTUNITY}, agency)
```
chouinar: The get opportunity function should verify they have VIEW access. The `get_opportunity_for_grantors` already has this in place.

api/tests/src/api/opportunities_v1/test_opportunity_summary.py
```diff
@@ -0,0 +1,252 @@
...
+
+    response = client.post(
+        f"/v1/opportunities/{opportunity.opportunity_id}/summary",
+        json=invalid_dates_request,
+        headers={"X-SGG-Token": token},
+    )
+
+    assert response.status_code == 422
+    response_json = response.get_json()
+    assert "post date" in str(response_json).lower() and "close date" in str(response_json).lower()
```
chouinar: The response JSON is parseable, don't just check that post date appeared somewhere in the string, look at the actual error. Something like `assert response_json["errors"][0] == ...` If there are multiple errors, check for the set of them.

api/tests/src/api/opportunities_v1/test_opportunity_summary.py
```diff
@@ -0,0 +1,252 @@
...
+        json=invalid_award_request,
+        headers={"X-SGG-Token": token},
+    )
+
+    assert response.status_code == 422
+    response_json = response.get_json()
+    assert (
+        "award floor" in str(response_json).lower()
+        and "award ceiling" in str(response_json).lower()
+    )
```
chouinar, line 177: Same as above, shouldn't ever be check against the raw text of the response, it's easily parsed, find the error.

api/src/api/opportunities_v1/opportunity_routes.py
dghazvini: If it's only the grantors dealing with these opportunity summary endpoints then yeah it should probably be under the grantor blueprint, I don't think we specified in the tech spec. CC @glenblosser-nava ?

Screenshot 2026-03-09 125706

api/tests/src/api/opportunities_v1/test_opportunity_summary.py
dghazvini: Updated

api/tests/src/api/opportunities_v1/test_opportunity_summary.py
dghazvini: Updated

api/src/services/opportunities_v1/opportunity_summaries.py
dghazvini: Updated to remove the unnecessary verify_access call.

api/src/services/opportunities_grantor_v1/opportunity_summaries.py
dghazvini: Updated

api/src/services/opportunities_v1/opportunity_summaries.py
dghazvini: Removed this block - unneeded after changing to call OpportunitySummary.

api/src/services/opportunities_v1/opportunity_summaries.py
dghazvini: Updated to just call OpportunitySummary instead.

api/src/services/opportunities_v1/get_opportunity.py
dghazvini: I had some confusion with if we should be using the grantor service methods or not but yes; I updated it to use the `get_opportunity_for_grantors` method after moving everything to grantor blueprint/services and removed this statement.

api/src/api/opportunities_v1/opportunity_routes.py
dghazvini: Moved to be under the grantor blueprint and moved the according schemas and services/updated test references.

api/src/api/opportunities_v1/opportunity_schemas.py
dghazvini: Sorry, meant to make a note to revisit this after looking over the tech spec again on Friday... I added several validations in the request schema based on what I see on that [tech spec chart](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2613215233/Phase+2+Feature+Build+Out) at the bottom.

api/src/api/opportunities_v1/opportunity_schemas.py
dghazvini: Updated to remove.

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
```diff
@@ -218,3 +227,125 @@ class OpportunityListResponseSchema(AbstractResponseSchema, PaginationMixinSchem
...
+        metadata={"description": "Whether the opportunity is forecasted", "example": False},
+    )
+
+    summary_description = fields.String(
+        required=True,
+        validate=validators.Length(max=18000),
+        metadata={"description": "Opportunity summary", "example": "This opportunity..."},
+    )
+
+    is_cost_sharing = fields.Boolean(required=True)
```
chouinar: Could we have a description for each of these fields? Should be fine to copy-paste whatever we have in the summary response schema.

And for any fields that are basic strings, add an example. Other types don't really need it (they have reasonable default examples). This is mostly so the example generated on the docs is at least somewhat meaningful looking like having a url that is an actual url rather than "string".

api/src/services/opportunities_grantor_v1/opportunity_summaries.py
```diff
@@ -0,0 +1,92 @@
...
+    Privilege,
+)
+from src.db.models.opportunity_models import Opportunity, OpportunitySummary
+from src.db.models.user_models import User
+from src.services.opportunities_grantor_v1.get_opportunity import get_opportunity_for_grantors
+
+logger = logging.getLogger(__name__)
+
+
+class OpportunitySummaryCreateRequest(BaseModel):
```
chouinar: Don't think this is used anymore with the other changes you made.

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
```diff
@@ -218,3 +227,125 @@ class OpportunityListResponseSchema(AbstractResponseSchema, PaginationMixinSchem
...
+                        MarshmallowErrorContainer(
+                            ValidationErrorType.INVALID,
+                            "Post date must be less than or equal to close date",
+                        )
+                    ]
+                )
+
+
+class OpportunitySummaryCreateResponseV1Schema(AbstractResponseSchema):
+    data = fields.Nested(OpportunitySummaryV1Schema())
```
chouinar: The `OpportunitySummaryV1Schema` doesn't have the `opportunity_summary_id` in the schema, we'd probably want to return that so a user could in turn call the PUT endpoint for updating this (once built). Can still reuse that schema, just make a subclass of that one and add the ID.

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
```diff
@@ -218,3 +227,125 @@ class OpportunityListResponseSchema(AbstractResponseSchema, PaginationMixinSchem
...
+    )
+
+    is_cost_sharing = fields.Boolean(required=True)
+
+    post_date = fields.Date(required=True)
+
+    expected_number_of_awards = fields.Integer(
+        required=False,
+        validate=validators.Range(min=0, max=999_999_999_999_999),
+    )
```
chouinar, line 296: Something to be careful of - required != nullability. By default, `allow_none` is generally False (if you set the default to None, it'll default to True).

What is the distinction? Required just says whether it needs to be present, `allow_none` says whether it can be null (if present). If someone wanted to create (or update when we get there) a field to be null that isn't required, the only way they could do it is leave it out. `{"expected_number_of_awards": null}` would fail.

We'll need to go through each of these and if a field is nullable, add `allow_none=True`.

api/tests/src/api/opportunities_grantors_v1/test_opportunity_summary.py
```diff
@@ -0,0 +1,257 @@
...
+    assert FundingInstrument.COOPERATIVE_AGREEMENT in summary_data["funding_instruments"]
+
+    assert len(summary_data["funding_categories"]) == 1
+    assert FundingCategory.AGRICULTURE in summary_data["funding_categories"]
+
+    assert len(summary_data["applicant_types"]) == 1
+    assert ApplicantType.CITY_OR_TOWNSHIP_GOVERNMENTS in summary_data["applicant_types"]
+
+
+def test_opportunity_summary_create_duplicate_summary(
```
chouinar, line 231: I'd recommend adding some tests that verify the schema, since in this endpoint a lot of the logic is really baked into the schema.

Min/max/required/nullability probably? I wouldn't worry as much about things like enum checks.

Easiest way to do something like this is just setup the data with all relevant fields wrong and then just verify which fields have errors.

```py
# For example, for max length
request = {
  "field_a": "a" * 41, # assuming this was a max of 40
}

# request stuff

expected_error_fields = ["field_a"]

erroring_fields = []
for err in response_json["errors"]:
    assert "something common" in err["message"]

    field = err["field"]
    erroring_fields.append(field)
    assert field in expected_error_fields

assert len(erroring_fields) == len(expected_error_fields)


```

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
dghazvini: Added descriptions and examples for the fields.

api/src/services/opportunities_grantor_v1/opportunity_summaries.py
dghazvini: Good call out, removed `OpportunitySummaryCreateRequest`.

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
dghazvini: Got it, I added an `OpportunitySummaryDetailSchema` class that takes in `opportunity_summary_id`.

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
dghazvini: Thanks for the call out, didn't realize the default behavior! I added `allow=True` for all the respective fields; think I covered them all, used the `OpportunitySummaryV1Schema` as reference.

api/tests/src/api/opportunities_grantors_v1/test_opportunity_summary.py
dghazvini: Added a `test_opportunity_summary_create_schema_validation` test.

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
```diff
@@ -218,3 +227,214 @@ class OpportunityListResponseSchema(AbstractResponseSchema, PaginationMixinSchem
...
+        required=True,
+        allow_none=True,
+        metadata={
+            "description": "Whether or not the opportunity has a cost sharing/matching requirement",
+        },
+    )
+
+    post_date = fields.Date(
+        required=True,
+        allow_none=True,
```
chouinar: I'd probably not allow post date to be none. It's allowed in the DB because grants.gov has a handful of these, but if it is null, this summary can never be public since a summary can't be shown to a user until the post date is reached. I don't see any use case for that, better to disallow it.

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
```diff
@@ -218,3 +227,214 @@ class OpportunityListResponseSchema(AbstractResponseSchema, PaginationMixinSchem
         fields.Nested(OpportunityGrantorSchema),
         metadata={"description": "List of opportunities"},
     )
+
+
+class OpportunitySummaryCreateRequestV1Schema(OpportunitySummaryV1Schema):
```
chouinar: At this point, it might be best to just fully make a separate schema from that schema since you override most fields anyways. Sorry for the churn on this, this schema is probably the most complex one of all the opportunity publish work, and has so many pieces.

KevinJBoyer:
Drive-by ask to please update the PR title before you merge to follow our convention of starting with `[Issue #8006] ...` rather than the branch name :)

dghazvini:
> Drive-by ask to please update the PR title before you merge to follow our convention of starting with `[Issue #8006] ...` rather than the branch name :)

Done, thanks for calling it out!

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
dghazvini: Removed `allow_none=True` parameter for `post_date`.

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
dghazvini: No worries, the feedback is great and it's best to get everything right on the first PR 👍 

So I updated this to `OpportunitySummaryCreateRequestV1Schema(Schema)` and added in all the corresponding fields needed (ignoring the 4 unneeded fields mentioned on the tech spec chart), and updated the validation schema unit test where needed.

Also I realized I didn't have the functionality in place that was actually setting `archive date` to 30 days from the close date, so i added a `@validates_schema` for that as well to set it if no value is provided (if it's better to be in the service layer i can move it).

Also for `fiscal_year` - Copied that over as well and current behavior is `required=False, allow_none=True` (along with the validation for checking between years 1900-2100). Not sure if we needed to change this or if that should reflect the current behavior of only applying to forecasted summaries.

Issue #8107, opened 2026-01-23
Create endpoint for fetching a workflow


chouinar:
### Summary

We want to create a `GET /v1/workflows/:workflow_id` endpoint to fetch information about a workflow.

### Request
No request body, just the workflow ID uuid from the path

### Response

```python
{
   "data": {
     "workflow_type": "", # the WorkflowType enum
     "current_workflow_state": "",
      "is_active": true,
      "created_at": "", # datetime
      "updated_at": "", # datetime
      # These should be ordered by created_at timestamp
      "workflow_audit_events": [
        {
          "workflow_audit_id": "<some uuid>",
          "acting_user": {
             # we should reuse our user models from other request schemas, the email + profile should be included
          },
          "transition_event": "",
          "source_state": "",
          "target_state": "",
          "event_id": "<some uuid>",
          "audit_metadata": {}, # freeform dict/json
          "created_at": "", # datetime
          "updated_at": "", # datetime
        }
      ],
     # Should be sorted by created_at
     "workflow_approvals": [
        "workflow_approval_id": "<some uuid>",
        "approving_user": {
            "user_id": "<uuid>",
            "email": "",
            "first_name": "",
            "last_name": "",
        },
        "event_id": "<some uuid>",
        "is_still_valid": True,
        "comment": "",
        "approval_type": "", # ApprovalType enum
        "approval_response_type": "approved", # the ApprovalResponseType enum
         "created_at": "", # datetime
         "updated_at": "", # datetime
     ],
     "workflow_approval_config": {
         "waiting_for_director_approval": {
            "approval_type": "", # ApprovalType enum,
            "required_privileges": [..], # Privilege enum
            # Not sure on exact naming, but we'll want to get this
            # as described below
            "possible_users": [ { } # same user as approving_user above
            ]
         }
     },
   },
   
   # For entities, we'll just return the IDs
   "opportunity_id": "",
   "application_id": "",
   "application_submission_id": "",
}

```

### Auth
Allow JWTs or user API keys

For the auth, we'll determine whether a user can access it by checking the entity associated with the workflow, and using the following corresponding privilege:
* opportunity -> `view_opportunity`
* application -> `view_application`
* application submission -> nothing yet, don't have a mechanism for submissions to be viewed by a grantor yet (this case should just always 403)

In all of these cases, use the entity to find the agency on the opportunity (403 in the unlikely event the agency is null), and verify against that.

### Logic
* Fetch the workflow from the DB, 404 if not found
[...]

PR #8853, opened 2026-03-06
[Issue #8107] Create endpoint for fetching a workflow

kkrug:
## Summary

Fixes for #8107  

## Changes proposed

- Created get_workflow.py service
- Updated workflow_routes.py to add new GET route
- Updated workflow_schemas.py to add various new schemas
- Updated endpoint_access_util.py to add function to get users with specific privileges in an agency
- Updated test_workflow_routes to add test for new functionality
- Updated test_endpoint_util.py to test new function
- Updated approval_service.py to make get_agency_for_workflow public

## Context for reviewers

This is fairly dense as far as GET endpoints go as there is a lot we need to return (see the ticket for more info). We want to create a `GET /v1/workflows/:workflow_id` endpoint to fetch information about a workflow. The most important parts of this review are the various new schemas and ensuring the correct data is fetched in the service layer (and as efficiently as possible.)

## Validation steps

Confirm tests pass.

### DIFF
--- a/api/src/auth/endpoint_access_util.py
+++ b/api/src/auth/endpoint_access_util.py
@@ -151,6 +151,7 @@ def get_users_with_privileges_for_agency(
             selectinload(User.linked_login_gov_external_user),
             selectinload(User.profile),
         )
+        .order_by(User.created_at.asc())
     )
 
     users = db_session.execute(stmt).scalars().unique().all()
--- a/api/src/services/workflows/get_workflow.py
+++ b/api/src/services/workflows/get_workflow.py
@@ -90,7 +90,7 @@ def get_workflow_and_verify_access(
         404: If workflow doesn't exist
         403: If user doesn't have required privilege or entity type not supported
     """
-    log_extra: dict[str, str] = {"user_id": str(user.user_id), "workflow_id": str(workflow_id)}
+    log_extra: dict = {"user_id": user.user_id, "workflow_id": workflow_id}
     logger.info("Fetching workflow for user", extra=log_extra)
 
     workflow = _get_workflow(db_session, workflow_id)
@@ -169,16 +169,20 @@ def build_workflow_approval_config(
 
     # Build config for each approval event
     for event_name, approval_config in config.approval_mapping.items():
-        log_extra["event_name"] = event_name
-        log_extra["approval_type"] = approval_config.approval_type.value
-        logger.debug("Processing approval config for event", extra=log_extra)
+        event_log_extra = {
+            "event_name": event_name,
+            "approval_type": approval_config.approval_type.value,
+        }
+        logger.debug("Processing approval config for event", extra=log_extra | event_log_extra)
 
         possible_users = get_users_with_privileges_for_agency(
             db_session, agency, approval_config.required_privileges
         )
 
-        log_extra["possible_users_count"] = len(possible_users)
-        logger.debug("Found possible users for approval", extra=log_extra)
+        logger.debug(
+            "Found possible users for approval",
+            extra=log_extra | event_log_extra | {"possible_users_count": len(possible_users)},
+        )
 
         approval_config_dict[event_name] = {
             "approval_type": approval_config.approval_type,
--- a/api/tests/src/auth/test_endpoint_access_util.py
+++ b/api/tests/src/auth/test_endpoint_access_util.py
@@ -406,9 +406,29 @@ def test_get_users_with_privileges_for_agency_deduplicates_user_with_same_privil
     AgencyUserRoleFactory.create(agency_user=agency_user, role=role_a)
     AgencyUserRoleFactory.create(agency_user=agency_user, role=role_b)
 
+    # Another user with the same privilege (should be included)
+    user_with_privilege = UserFactory.create()
+    agency_user_with_privilege = AgencyUserFactory.create(agency=agency, user=user_with_privilege)
+    AgencyUserRoleFactory.create(agency_user=agency_user_with_privilege, role=role_a)
+
+    # Another user in the agency without that privilege (should not be included)
+    user_without_privilege = UserFactory.create()
+    agency_user_without_privilege = AgencyUserFactory.create(
+        agency=agency, user=user_without_privilege
+    )
+    role_without_target_privilege = RoleFactory.create(
+        privileges=[Privilege.MANAGE_AGENCY_MEMBERS], is_agency_role=True
+    )
+    AgencyUserRoleFactory.create(
+        agency_user=agency_user_without_privilege, role=role_without_target_privilege
+    )
+
     results = get_users_with_privileges_for_agency(
         db_session, agency, [Privilege.PROGRAM_OFFICER_APPROVAL]
     )
 
-    assert len(results) == 1
-    assert results[0].user_id == user.user_id
+    assert len(results) == 2
+    result_user_ids = {result.user_id for result in results}
+    assert user.user_id in result_user_ids
+    assert user_with_privilege.user_id in result_user_ids
+    assert user_without_privilege.user_id not in result_user_ids


api/src/api/workflows/workflow_routes.py
```diff
@@ -38,3 +45,45 @@ def workflow_event_put(db_session: db.Session, json_data: dict) -> response.ApiR
...
+            "is_active": workflow.is_active,
+            "created_at": workflow.created_at,
+            "updated_at": workflow.updated_at,
+            "workflow_audit_events": workflow.workflow_audits,
+            "workflow_approvals": workflow.workflow_approvals,
+            "workflow_approval_config": approval_config,
+            "opportunity_id": workflow.opportunity_id,
+            "application_id": workflow.application_id,
+            "application_submission_id": workflow.application_submission_id,
+        }
```
chouinar: Can we put all of this in a single service layer function? We want to keep as little as possible in the API route definitions. Also makes it easier to reuse functionality if needed on other endpoints.

---

For the data, return the workflow itself. A trick for "adding" something to the DB models for responses that we've done:

```py
approval_config = ...
setattr(workflow, "workflow_approval_config", approval_config)
```

Note, setattr is generally the same as `workflow.workflow_approval_config = approval_config` but avoids mypy complaining about typing. This is also perfectly fine to do, if a field is added to a SQLAlchemy model but isn't `Mapped` it just doesn't do anything, but it's still in the object so works great for turning into JSON.

api/src/services/workflows/get_workflow.py
```diff
@@ -0,0 +1,218 @@
...
+            .selectinload(Competition.opportunity)
+            .selectinload(Opportunity.agency_record),
+        )
+    )
+
+    workflow = db_session.execute(stmt).scalar_one_or_none()
+    return workflow
+
+
+def _get_agency_for_workflow(workflow: Workflow) -> Agency:
```
chouinar: This already exists in the approval service - remove the underscore from it and then we can just reuse it.

api/src/services/workflows/get_workflow.py
```diff
@@ -0,0 +1,218 @@
...
+        raise_flask_error(404, message=f"Could not find Workflow with ID {workflow_id}")
+
+    try:
+        agency = _get_agency_for_workflow(workflow)
+    except OpportunityWithoutAgencyError:
+        logger.warning("Opportunity associated with workflow has no agency", extra=log_extra)
+        raise_flask_error(403, message="Forbidden")
+    except ImplementationMissingError:
+        logger.warning("No way to determine agency for workflow", extra=log_extra)
+        raise_flask_error(403, message="Forbidden")
```
chouinar: This case probably should let the error through as that is a genuine issue that we should address / would want to see a 500.

api/src/services/workflows/get_workflow.py
```diff
@@ -0,0 +1,218 @@
...
+    if opportunity is None:
+        raise ImplementationMissingError("No approach implemented to find agency for workflow")
+
+    if opportunity.agency_record is None:
+        raise OpportunityWithoutAgencyError("Opportunity does not have an agency record")
+
+    return opportunity.agency_record
+
+
+def get_workflow_for_user(db_session: db.Session, user: User, workflow_id: uuid.UUID) -> Workflow:
```
chouinar: I might adjust this to be `get_workflow_and_verify_access` or something that says "does auth" rather than "for user".

api/src/api/workflows/workflow_schemas.py
```diff
@@ -127,3 +135,212 @@ class WorkflowEventResponseDataSchema(Schema):
...
+
+    @pre_dump
+    def extract_user_fields(self, user: User, **kwargs: Any) -> dict:
+        """Extract user fields from User model using properties."""
+        return {
+            "user_id": user.user_id,
+            "email": user.email,
+            "first_name": user.first_name,
+            "last_name": user.last_name,
+        }
```
chouinar: Do we need this? It should be able to parse that out of the user record.

api/src/api/workflows/workflow_schemas.py
```diff
@@ -127,3 +135,212 @@ class WorkflowEventResponseDataSchema(Schema):
...
+                data["workflow_audit_events"], key=lambda x: x.created_at
+            )
+
+        # Sort approvals by created_at
+        if data.get("workflow_approvals"):
+            data["workflow_approvals"] = sorted(
+                data["workflow_approvals"], key=lambda x: x.created_at
+            )
+
+        return data
```
chouinar: There's two other ways we might prefer to do this:
* In the relationship itself for a workflow, we can sort the relationship. Since I don't think we always care about sorting, probably not what we want.
* Just sort it after fetching. Here's where [we do that for application forms on an application](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/services/applications/get_application.py#L88). Had to do it there as the sorting was actually on a different joined table and it works fine.

api/src/services/workflows/get_workflow.py
```diff
@@ -0,0 +1,218 @@
...
+            "receive_program_officer_approval": {
+                "approval_type": "PROGRAM_OFFICER_APPROVAL",
+                "required_privileges": ["PROGRAM_OFFICER_APPROVAL"],
+                "possible_users": [...]
+            }
+        }
+    """
+    log_extra: dict[str, str | int] = workflow.get_log_extra() | {
+        "agency_id": str(agency.agency_id)
+    }
```
chouinar: A trick that should work. MyPy complains if you don't type the dict and then assign different types to it (eg. `log_extra["a"] = "hello"` is fine but if you do `log_extra["b"] = 1` after it says it expected a string). If you just type it as a dict with no params, it shouldn't complain. Then you can avoid any conversions of types that we already support conversions for like UUIDs and enums.

```suggestion
    log_extra: dict = workflow.get_log_extra() | {
        "agency_id": agency.agency_id
    }
```

api/src/auth/endpoint_access_util.py
```diff
@@ -117,6 +120,43 @@ def get_log_info_for_resource(resource: Organization | Application | Agency | No
...
+            AgencyUser.agency_id == agency.agency_id,
+            LinkRolePrivilege.privilege.in_(required_privileges),
+        )
+        .options(
+            selectinload(User.linked_login_gov_external_user),
+            selectinload(User.profile),
+        )
+    )
+
+    users = db_session.execute(stmt).scalars().unique().all()
```
chouinar, line 157: You have both distinct and unique here, is that to avoid the case where a given user has the same privilege twice via different roles?

Could we add a test for that specifically (doesn't need to be for the route, just this function)? I'm not sure distinct/unique will account for that, we might need a different query shape.

api/src/api/workflows/workflow_routes.py
kkrug: The linter did not like the use of `setattr` - it says 'Do not call `setattr` with a constant attribute value. It is not any safer than normal property access.' It suggested using `workflow.workflow_approval_config = approval_confi` so I made that change instead with `# type: ignore[attr-defined]`.

api/src/services/workflows/get_workflow.py
kkrug: Updated

api/src/services/workflows/get_workflow.py
kkrug: Updated

api/src/services/workflows/get_workflow.py
kkrug: Makes sense, updated.

api/src/api/workflows/workflow_schemas.py
kkrug: Yeah you're right, this was redundant - fixed.

api/src/api/workflows/workflow_schemas.py
kkrug: Updated to sort after fetching

api/src/services/workflows/get_workflow.py
kkrug: Updated

api/src/auth/endpoint_access_util.py
kkrug: Yeah, that's the case it was avoiding - I created a test for that scenario.

api/src/services/workflows/get_workflow.py
```diff
@@ -0,0 +1,193 @@
...
+
+    config, _ = WorkflowRegistry.get_state_machine_for_workflow_type(workflow.workflow_type)
+
+    approval_config_dict = {}
+
+    # Build config for each approval event
+    for event_name, approval_config in config.approval_mapping.items():
+        log_extra["event_name"] = event_name
+        log_extra["approval_type"] = approval_config.approval_type.value
+        logger.debug("Processing approval config for event", extra=log_extra)
```
chouinar: If you add event name + approval type to the log extra, when you log after the loop, it'll have whatever the last one is.

Might want to do something like:

```py
event_log_extra = {"event_name": event_name, "approval_type": approval_config.approval_type}

logger.info("whatever", extra=log_extra|event_log_extra)
```

The `|` will merge dicts together.

api/src/auth/endpoint_access_util.py
```diff
@@ -117,6 +120,43 @@ def get_log_info_for_resource(resource: Organization | Application | Agency | No
...
+        .join(LinkRolePrivilege, LinkRolePrivilege.role_id == Role.role_id)
+        .where(
+            AgencyUser.agency_id == agency.agency_id,
+            LinkRolePrivilege.privilege.in_(required_privileges),
+        )
+        .options(
+            selectinload(User.linked_login_gov_external_user),
+            selectinload(User.profile),
+        )
+    )
```
chouinar, line 155: We should add some sort to this. Not sure on what, user ID or created at? That way if you call GET workflow multiple times, it'll give you users in the same order every time which can help if the frontend calls GET after updating something else.

```suggestion
        ).order_by(User.created_at.asc())
    )
```

api/src/services/workflows/get_workflow.py
```diff
@@ -0,0 +1,193 @@
...
+        workflow_id: UUID of the workflow to fetch
+
+    Returns:
+        Workflow object if user has access
+
+    Raises:
+        404: If workflow doesn't exist
+        403: If user doesn't have required privilege or entity type not supported
+    """
+    log_extra: dict[str, str] = {"user_id": str(user.user_id), "workflow_id": str(workflow_id)}
```
chouinar:
```suggestion
    log_extra: dict = {"user_id": user.user_id, "workflow_id":workflow_id}
```

api/tests/src/auth/test_endpoint_access_util.py
```diff
@@ -384,3 +386,29 @@ def count(*args):
...
     # Assert no additional queries made
     assert len(queries) == 0
+
+
+def test_get_users_with_privileges_for_agency_deduplicates_user_with_same_privilege_via_multi_roles(
+    db_session, enable_factory_create
+):
+    """User assigned the same privilege via two different roles should appear only once."""
+    agency = AgencyFactory.create()
+    user = UserFactory.create()
```
chouinar, line 396: As a sanity check, could we add a few users? Another user with the same privilege + another user in the agency without that privilege (that doesn't get picked up). This or a separate test.

api/src/services/workflows/get_workflow.py
kkrug: Updated

api/src/auth/endpoint_access_util.py
kkrug: Sort by created at works for me - updated

api/src/services/workflows/get_workflow.py
kkrug: Fixed

api/tests/src/auth/test_endpoint_access_util.py
kkrug: Test updated

PR #8866, opened 2026-03-06
[Issue #8333] schedule task to create agencies in lower env

### LINKED ISSUE #8333: Temporary approaches to assign user roles for Opportunity Management

Dao-REI:
## Summary

Fixes #8333 

## Changes proposed

The task was created in #8779. This update is just to schedule the task.

## Context for reviewers

This task is scheduled to run Every day at 8:15am Eastern Time during DST. 9:15am during non-DST.

### DIFF
--- a/infra/api/app-config/env-config/scheduled_jobs.tf
+++ b/infra/api/app-config/env-config/scheduled_jobs.tf
@@ -223,10 +223,10 @@ locals {
     }
     setup-lower-env-agencies = {
       task_command = ["poetry", "run", "flask", "task", "setup-lower-env-agencies"]
-      # Every day at 8:15am Eastern Time during DST. 9:15am during non-DST.
-      # Runs just before the search load job that runs 30 minutes after the hour.
-      schedule_expression = "cron(15 13 * * ? *)"
-      # Only enable in dev/staging/training, do not run in prod
+      # Every day at 7:45am Eastern Time during DST. 8:45am during non-DST.
+      # Runs just before the agency search load job that runs at the top of the hour.
+      schedule_expression = "cron(45 12 * * ? *)"
+      # Only enable in dev/staging, do not run in training/prod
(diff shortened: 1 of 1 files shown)


Dao-REI:
@chouinar I don't know how to validate scheduled jobs.

infra/api/app-config/env-config/scheduled_jobs.tf
```diff
@@ -214,5 +221,16 @@ locals {
       mem              = try(local.scheduled_jobs_config[var.environment].mem, null)
       environment_vars = try(local.scheduled_jobs_config[var.environment].environment_vars, null)
     }
+    setup-lower-env-agencies = {
+      task_command = ["poetry", "run", "flask", "task", "setup-lower-env-agencies"]
+      # Every day at 8:15am Eastern Time during DST. 9:15am during non-DST.
+      # Runs just before the search load job that runs 30 minutes after the hour.
+      schedule_expression = "cron(15 13 * * ? *)"
+      # Only enable in dev/staging/training, do not run in prod
```
chouinar: A few comment suggestions, but I'd suggest running at :45 after the hour when we run if we want to be near the agency search load job (which if you look above, runs at the top of the hour). 

```suggestion
      task_command = ["poetry", "run", "flask", "task", "setup-lower-env-agencies"]
      # Every day at 7:45am Eastern Time during DST. 8:45am during non-DST.
      # Runs just before the agency search load job that runs at the top of the hour.
      schedule_expression = "cron(45 12 * * ? *)"
      # Only enable in dev/staging, do not run in training/prod
```

PR #8867, opened 2026-03-06
Sfll form filling branch aoy

aoysimmons-GH:
## Summary

Fixes / Work for #8370 

## Changes proposed

**Test Utilities (Support Infrastructure)**

select-dropdown-utils.ts — Reusable dropdown interaction helper
lifecycle-helpers.ts — Reusable lifecycle/setup helper functions
playwright-env.ts - Update Opportunity IDs for each environment

**Form Filling Framework (Generic Form Handler)**

general-forms-filling.ts — Abstract form-fill logic (fillAnyForm(), fillField())
form-defaults.ts — Shared form config constants (noErrorsText, saveButtonTestId)

**SFLLL Form Test Data (Fixture)**

test-data-for-sflll-forms.fixture.ts — Type-safe test data object (SflllEntityData)

**SFLLL Form Page Object (Page Object Model)**

sflll-form.page.ts — Form metadata & field mapping:
SFLLL_FORM_CONFIG — Form name + inherited defaults
getSflllFillFields(data) — Maps test data to fillable form fields (selector/testId + value)

**SFLLL Form Test Case (End-to-End Test)**

fill-sflll-form.spec.ts — Main test implementation for issue #8370

## Context for reviewers

_**This implementation follows the Page Object Model (POM) pattern combined with a data-driven approach**_

### DIFF
--- a/frontend/tests/e2e/playwright-env.ts
+++ b/frontend/tests/e2e/playwright-env.ts
@@ -21,7 +21,7 @@ const OPPORTUNITY_IDS: Record<string, string> = {
 };
 
 // Determine environment: can be overridden via PLAYWRIGHT_TARGET_ENV
-const targetEnv = process.env.PLAYWRIGHT_TARGET_ENV || "staging";
+const targetEnv = process.env.PLAYWRIGHT_TARGET_ENV || "local";
 
 if (!Object.prototype.hasOwnProperty.call(BASE_URLS, targetEnv)) {
   throw new Error(


aoysimmons-GH:
@doug-s-nava @Bhavna-Ramachandran 

**Reference PRs:**

https://github.com/HHS/simpler-grants-gov/pull/8570
https://github.com/HHS/simpler-grants-gov/pull/8758
https://github.com/HHS/simpler-grants-gov/pull/8817

frontend/tests/e2e/apply/page-objects/sflll-form.page.ts
```diff
@@ -0,0 +1,341 @@
+import type { SflllEntityData } from "tests/e2e/apply/fixtures/test-data-for-forms.fixture";
+import { FillFieldDefinition } from "tests/e2e/utils/forms/general-forms-filling";
+
+export const SFLLL_FORM_CONFIG = {
+  formName: "Disclosure of Lobbying Activities (SF-LLL)",
+  noErrorsText: "No errors were detected",
```
doug-s-nava: this will be the same on all forms - we can store that as a globally referenced constant somewhere

frontend/tests/e2e/apply/page-objects/sflll-form.page.ts
```diff
@@ -0,0 +1,341 @@
...
+  saveButtonTestId: "apply-form-save",
```
doug-s-nava: this will also be the same on all forms

frontend/tests/e2e/utils/forms/general-forms-filling.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    throw new Error(
+      `Failed to fill ${field.section}: ${
+        error instanceof Error ? error.message : String(error)
+      }`,
+    );
+  }
+}
+
+export async function fillAnyForm(
```
doug-s-nava: worth noting that this function assumes you are on an application page. That is probably worth documenting in a comment

frontend/tests/e2e/utils/forms/general-forms-filling.ts
```diff
@@ -0,0 +1,103 @@
...
+      .waitFor({ state: "visible", timeout: formLoadTimeoutMs });
+
+    for (const field of fields) {
+      await fillField(testInfo, page, field);
+    }
+
+    await page.waitForTimeout(500);
+    await page.getByTestId(saveButtonTestId).click();
+
+    await expect(page.getByText(noErrorsText)).toBeVisible({
```
doug-s-nava: I would return here and let each test decide what they want the expectation to be here - it'd be nice to be able to use this same function for the failure test cases.

frontend/tests/e2e/utils/forms/general-forms-filling.ts
```diff
@@ -0,0 +1,103 @@
...
+  try {
+    if (field.type === "dropdown" && field.selector) {
+      await selectDropdownByValueOrLabel(page, field.selector, field.value);
+    } else if (field.type === "text" && field.testId) {
+      const locator = page.getByTestId(field.testId);
+      await locator.waitFor({ state: "attached", timeout: 5000 });
+      await locator.fill(field.value);
+    }
+
+    await testInfo.attach(`fillField-${field.section}-success`, {
```
doug-s-nava, line 34: why are these debuggers denoting section rather than field? I don't think that will give you the information you want since each section can contain multiple fields

frontend/tests/e2e/apply/page-objects/sflll-form.page.ts
```diff
@@ -0,0 +1,341 @@
...
+
+export function getSflllFillFields(
+  data: SflllEntityData,
+): FillFieldDefinition[] {
+  return [
+    {
+      selector: "#federal_action_type",
+      value: data.federalAction.type,
+      type: "dropdown",
+      section: "Section 1: Type of Federal Action",
```
doug-s-nava: see my comment on `fillField` - it's probably more useful to include the field name rather than the section

frontend/tests/e2e/apply/fixtures/test-data-for-sflll-forms.fixture.ts
```diff
@@ -0,0 +1,86 @@
+import { FillFieldDefinition } from "tests/e2e/utils/forms/general-forms-filling";
+
+export const FORMS_TEST_DATA = {
+  sflll: {
```
doug-s-nava, line 4: This file will get too big if we include test data for all forms in here. Let's isolate this to one file per form

frontend/tests/e2e/apply/fill-sflll-form.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+      timeout: 30000,
+    });
+
+    await createApplication(page, OPPORTUNITY_URL, testOrgLabel);
+
+    const sflllData = FORMS_TEST_DATA.sflll;
+
+    await fillAnyForm(testInfo, page, {
+      formName: SFLLL_FORM_CONFIG.formName,
+      fields: getSflllFillFields(sflllData),
```
doug-s-nava: This works! I like the idea of abstracting this since we'll have some relatively large forms and it'd be nice not to have to reinvent the wheel for each one. I have a couple of thoughts on how this could be improved, but those can be follow ups

just so I don't forget:

store the field information for each form, and the data, as static data in a dictionary keyed off of the locator for each field. Match the data to the field based on locator. This would let us fill any form easily, just passing those two dicts

frontend/tests/e2e/utils/forms/general-forms-filling.ts
aoysimmons-GH: Those are field names that I initially grouped as sections. I can also include properties like testId, value, and type if needed. Alternatively, I can rename section to field to avoid confusion.

Image

frontend/tests/e2e/apply/fixtures/test-data-for-sflll-forms.fixture.ts
aoysimmons-GH: Rename test-data-for-forms.fixture.ts to test-data-for-sflll-forms.fixture.ts

frontend/tests/e2e/apply/fill-sflll-form.spec.ts
Bhavna-Ramachandran, line 1: I see that this spec currently covers the standalone SF‑LLL form. Can you clarify if the objective of this spec is to verify the application submission workflow using only SF‑LLL or is it just to check whether SFLLL form can be filled and saved?

If the goal is to verify the full application workflow, the spec does not currently cover the submission user action, and it also does not check whether SF‑LLL is a required or conditional form. In the real-world user flow, the application can proceed without filling a conditional form, but a required form must be completed before submission.

The opportunity data used in this spec is for SF424B required and SFLLL conditional. If this is a standalone spec file, we need to use the FON which has SFLLL as required form.

Could you confirm the intended scope so the spec aligns with the apply workflow requirements? If you have any questions, feel free to reach out.

doug-s-nava: I think it's best to keep the scope of this PR smaller, and create a separate ticket for the form submission

frontend/tests/e2e/utils/forms/general-forms-filling.ts
doug-s-nava: Yes, please use field instead of section, since sections can contain multiple fields (though they may not in this particular form)

frontend/tests/e2e/apply/page-objects/sflll-form.page.ts
aoysimmons-GH: I will update these and will reflex the comment you mentioned below as well.

frontend/tests/e2e/apply/page-objects/sflll-form.page.ts
aoysimmons-GH: I created a form-defaults keep this

frontend/tests/e2e/apply/page-objects/sflll-form.page.ts
aoysimmons-GH: I created a form-defaults keep this

frontend/tests/e2e/apply/fill-sflll-form.spec.ts
aoysimmons-GH: I make a note about this. Or do you want me to create a ticket to implement it before we pick up other form?

frontend/tests/e2e/apply/fill-sflll-form.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+import playwrightEnv from "tests/e2e/playwright-env";
+import { createApplication } from "tests/e2e/utils/create-application-utils";
+import { fillAnyForm } from "tests/e2e/utils/forms/general-forms-filling";
+import {
+  clearPageState,
+  ensurePageClosed,
+} from "tests/e2e/utils/lifecycle-helpers";
+
+const { baseUrl, testOrgLabel } = playwrightEnv;
+const OPPORTUNITY_ID = "f7a1c2b3-4d5e-6789-8abc-1234567890ab";
```
doug-s-nava: will need a separate id for testing against staging. Will also need to handle logging into staging for when this runs there. That can be a fast follow on this ticket though.

frontend/tests/e2e/utils/forms/general-forms-filling.ts
```diff
@@ -0,0 +1,113 @@
...
+    );
+  }
+}
+
+/**
+ * Fills a form from the application page and saves it.
+ * Assumes the current page is already an application page
+ * where the form link (`formName`) is visible and clickable.
+ */
+export async function fillAnyForm(
```
doug-s-nava: since this function by default fills the form and then does an assertion against the post save state, I would either:

* make that clear in the function name (`fillFormAndAssertNoErrors`)
* even better, don't do any assertions here and do any assertions back in the main test function

I swear I'm not being arbitrary here, but doing what I can to live by Uncle Bob's guidelines https://herbertograca.com/2016/09/03/clean-code-3-functions-by-robert-c-martin-uncle-bob/

doug-s-nava (commented):
this is so close. Can you open for review?

frontend/tests/e2e/apply/fill-sflll-form.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+const OPPORTUNITY_URL = `/opportunity/${OPPORTUNITY_ID}`;
+
+test.describe("fill SF-LLL Form", () => {
+  test.beforeEach(async ({ context }) => {
+    await createSpoofedSessionCookie(context);
```
doug-s-nava: this will only work in local runs. Since we've enabled staging runs as of this week we'll need to include that in the scope here as well. You can reference Bhavna's PR for how she is handling that, but also maybe keep an eye on this thread to see if we need to make any adjustments https://betagrantsgov.slack.com/archives/C09QGE0HTJA/p1773143121233979

aoysimmons-GH: @doug-s-nava I updated

frontend/tests/e2e/playwright-env.ts
```diff
@@ -14,16 +14,16 @@ const BASE_URLS: Record<string, string> = {
   staging: process.env.STAGING_BASE_URL || "https://staging.simpler.grants.gov",
 };
 
-// Determine environment: can be overridden via PLAYWRIGHT_TARGET_ENV
```
doug-s-nava, line 17: I'm not sure what the purpose of the changes in this file are. Can we revert them?

frontend/tests/e2e/playwright-env.ts
```diff
@@ -14,6 +14,12 @@ const BASE_URLS: Record<string, string> = {
   staging: process.env.STAGING_BASE_URL || "https://staging.simpler.grants.gov",
 };
 
+// Opportunity IDs for each environment
+const OPPORTUNITY_IDS: Record<string, string> = {
```
doug-s-nava, line 18: different tests may need to use different opportunities for any number of reasons, so I don't think the concept of a global opportunity id constant is going to be very helpful. Not a blocker, though, we can fix this up later.

Issue #597, opened 2023-10-16
[Task]: Fix health check for frontend container


acouch:
### Summary

.

PR #8869, opened 2026-03-06
[Issue #8868] Upgrade python state machine to 3.0

### LINKED ISSUE #8868: Upgrade python state machine from 2.x to 3.0
### Summary

A recent version of python state machine came out that adds a lot of features. We want to upgrade to potentially use them, and just because it'll make referencing the docs a bit easier.

We don't want any meaningful functionality to change, although we likely need a few adjustments as the changes aren't fully backwards compatible.

### Acceptance criteria

- [ ] Upgraded to 3.0 https://python-statemachine.readthedocs.io/en/latest/releases/upgrade_2x_to_3.html

chouinar:
## Summary
Fixes #8868

## Changes proposed
* Upgrades python state machine to version 3.0

## Context for reviewers
This release changed a lot https://python-statemachine.readthedocs.io/en/latest/releases/3.0.0.html

It includes a ton of features, some of which we might consider using, but for now, we'll keep things mostly the same. I did have to adjust a bit of error logic but still checking the same actual things.

Importantly, there are two configurations that we kept as they were that are defined in the base class. While these are two things that are making it more like the SCXML specification that it now supports, I don't think we want them (transition states being null temporarily AND errors triggering events).

## Validation steps
Tests continue working with a few adjustments to error cases. Wanted to get this change in before we actually start running everything as that way we don't need two passes at testing after a major version upgrade.

### DIFF
--- a/api/pyproject.toml
+++ b/api/pyproject.toml
@@ -46,7 +46,7 @@ PyYAML = "^6.0.3"
 jsonref = "^1.1.0"
 pydantic-core = "2.41.5"
 urllib3 = "^2.6.3"
-python-statemachine = "3.0.0"
+python-statemachine = "^3.0.0"
 beautifulsoup4 = "^4.14.3"
 
 [tool.poetry.group.dev.dependencies]
--- a/api/poetry.lock
+++ b/api/poetry.lock
@@ -3709,4 +3709,4 @@ test = ["pytest", "pytest-cov"]
 [metadata]
 lock-version = "2.1"
 python-versions = "~3.14"
-content-hash = "256bb6e227e5ea5bdb086426366a8a1d8a5658dd4caa2ae4764c254565913de9"
+content-hash = "0b14ca7e5ad2e2cf19f916f43717120c7817a0400cc2c295b3236024cbced5b5"


api/pyproject.toml
```diff
@@ -46,7 +46,7 @@ PyYAML = "^6.0.3"
 jsonref = "^1.1.0"
 pydantic-core = "2.41.5"
 urllib3 = "^2.6.3"
-python-statemachine = "^2.5.0"
+python-statemachine = "3.0.0"
```
kkrug: Is there a reason why we don't want to use `^3.0.0` to allow for minor updates?

chouinar: Nope, just a side-effect of how I upgraded it - didn't look at what it put here. Good catch

fgmacedo:
Hi! I'm the author of python-statemachine. I came across this PR while searching for projects using the library on GitHub, and I was really happy to see you folks adopting v3!

If you have a moment, I'd love to hear any feedback — was the upgrade smooth? Is there anything that could be improved to make your experience with the library better?

Best regards!

chouinar:
@fgmacedo Upgrade was smooth, thanks for documenting the differences + adding configurations to be able to keep some of the changed behavior.

One detail I'm still a little uncertain of is the actual underlying state management. Since there is a possibility for a given state machine to be in multiple states now, how should we work with the library to ensure we only setup state machines to only ever have a single state? In 2.x there was an example for setting up a persistence model that I followed for persisting our state to the database - you can see our implementation [here](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/workflow/state_persistence/base_state_persistence_model.py). The `workflow` object is just a DB ORM model, so setting/getting `current_workflow_state` is against that column in the database table.

For our use case, it's preferable that we always have a single state, not multiple due to the fact that we'll be using this for folks from various government agencies to approve grants (ie. a lot of auditing will be involved, so keeping things linear as far as state-history is for the best).

If we did want to use the persistence model approach and leverage any of the multi-state setups, how would you recommend that be done - assuming the underlying persistence was a database.

fgmacedo:
Hi @chouinar, great to hear the upgrade went well!

**For your current use case (linear, single-state workflows):** you're perfectly fine. If your state machines use only flat states (no `State.Compound` or `State.Parallel`), `model.state` will always be a single scalar value — exactly like v2. Your current DB persistence pattern (get/set a single column) works as-is, no changes needed.

**On the broader multi-state persistence question:** that's a really valuable point you raised. Your feedback actually made me realize this is an important ergonomics gap in the library that I want to address properly.

The core issue is: when a state chart uses compound or parallel states, the configuration can have multiple active states simultaneously. Today, the library stores *all* active states (including redundant parent states) as an `OrderedSet` on the model — which doesn't map cleanly to a single DB column.

The library calls `getattr`/`setattr` on the model's state field, so this property pattern works transparently — no changes needed on the state chart side. There's a full [SQLite example](https://github.com/fgmacedo/python-statemachine/blob/develop/tests/examples/sqlite_persistent_model_machine.py) in the repo (based on develop) if you want to see a complete working implementation.


The good news is that parent states are always redundant — they can be reconstructed from just the leaf (atomic) states. So for compound state machines (nested, but one active path at a time), the persisted value could still be a single string. Only truly parallel regions would require multiple values.

Given your feedback, I'm considering a few possible approaches and would love your input:

1. **A class-level opt-in flag** (e.g., `persist_leaf_states_only = True`) — the state chart would automatically filter to leaf states before writing to the model. Simple, but the library decides the strategy.

2. **A Model Protocol** — the model could implement an opt-in interface (e.g., `serialize_state(value, states_map)` / `deserialize_state(raw, states_map)`). If detected, the state chart would pass enough context (the states map, state hierarchy info) for the model to make its own informed decisions about serialization — CSV, JSON, multiple columns, or whatever fits the persistence layer. This keeps the responsibility where it belongs: on the model.

3. **A codec/transform parameter** — something like `StateChart(model=doc, state_codec=LeafOnlyCodec())` that intercepts reads/writes between the configuration and the model.

Each has different trade-offs around simplicity vs. flexibility. We follow semantic versioning, so whichever approach we settle on would ship as **opt-in in the 3.x series** (fully backward compatible) and potentially become the default in 4.x.

**If you want to use compound/parallel states right now**, before these improvements land, there's a straightforward workaround: implement serialization/deserialization on your model's state property. For example, you could store the state as a comma-separated string (or JSON) in your existing DB column:

```python
from statemachine.orderedset import OrderedSet

class MyModel:
    @property
    def state(self):
        raw = self._read_from_db()  # e.g., "legal_pending,tech_pending"
        if raw is None:
            return None
        parts = raw.split(",")
        return parts[0] if len(parts) == 1 else OrderedSet(parts)

    @state.setter
    def state(self, value):
        if value is None:
            raw = None
        elif isinstance(value, OrderedSet):
            raw = ",".join(str(v) for v in value)
        else:
            raw = str(value)
        self._write_to_db(raw)
```

Your perspective as someone building a real-world audit-heavy workflow on top of the library is exactly the kind of feedback that helps shape this. Do any of these approaches resonate with your use case? Or do you see other patterns we should consider?

I've opened [#597](https://github.com/fgmacedo/python-statemachine/issues/597) to track this properly — it maps out the problem in detail and lays out the approaches we're considering. Your feedback there (or here) is very welcome!

Issue #8597, opened 2026-02-19
Add Create Opportunity button to Opportunities List page


glenblosser-nava:
### Summary

Add a Create Opportunity button to the Opportunities List page

### Acceptance criteria

- [ ] Create Opportunity button is visible and takes you to the Create Opportunity page

doug-s-nava:
figma: https://www.figma.com/design/o5g0d2W5RKHYHjdvVCqTV2/Grantor-Designs-%7C-Quad-5-%7C-Simpler-Grants-2026?node-id=8507-16776&p=f&m=dev

desaiamit:
UAT for this ticket to be conducted as part of #9336

PR #8873, opened 2026-03-06
[Issue #8597] Add Create Opportunity button to opportunities list

desaiamit:
## Summary

Work for #8597

## Changes proposed

- add a `Create Opportunity` button to the opportunities list page
- route the button to the agency-specific create page
- update the opportunities page header layout for single-agency users
- update the opportunities page header layout for multi-agency users
- add and update test coverage for single-agency and multi-agency behavior

## Context for reviewers

- the page previously rendered the create button inside the table section
- the header/action area is now rendered at the page level
- the multi-agency selector was moved into the same layout row as the create button to match the Figma design

### Automated local validation

- `cd frontend && npm test  --runTestsByPath '/Users/adesai/projects/simpler-grants-gov/frontend/src/app/[locale]/(base)/opportunities/page.test.tsx'`

### Manual local validation

- verify single-agency user:
  - count appears on top row
  - `Showing opportunities for ...` appears on second row
  - `Create Opportunity` is aligned on the right
  - Confirm button navigates to the Create Opportunity page
- verify multi-agency user:
  - count appears on top row
  - agency dropdown appears on the left of the second row
  - `Create Opportunity` appears on the right of the second row
  - Confirm button navigates to the Create Opportunity page

### DIFF
--- a/frontend/src/components/workspace/AgencySelector.tsx
+++ b/frontend/src/components/workspace/AgencySelector.tsx
@@ -4,6 +4,7 @@ import { UserAgency } from "src/services/fetch/fetchers/userAgenciesFetcher";
 
 import { useTranslations } from "next-intl";
 import { useRouter } from "next/navigation";
+import { Label, Select } from "@trussworks/react-uswds";
 
 export const AgencySelector = ({
   agencies,
@@ -19,11 +20,10 @@ export const AgencySelector = ({
 
   return (
     <div className={className ?? "usa-form-group margin-bottom-4"}>
-      <label className="usa-label text-bold" htmlFor="agency-selector">
+      <Label htmlFor="agency-selector" className="text-bold">
         {t("agencySelector")}
-      </label>
-      <select
-        className="usa-select"
+      </Label>
+      <Select
         id="agency-selector"
         name="agency"
         defaultValue={currentAgencyId}
@@ -34,7 +34,7 @@ export const AgencySelector = ({
             {agency.agency_name}
           </option>
         ))}
-      </select>
+      </Select>
     </div>
   );
 };


frontend/src/app/[locale]/(base)/opportunities/page.test.tsx
```diff
@@ -65,10 +65,22 @@ jest.mock("next/navigation", () => ({
 }));
 
 jest.mock("src/components/workspace/AgencySelector", () => ({
-  AgencySelector: ({ agencies }: { agencies: UserAgency[] }) => (
+  AgencySelector: ({
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunities/page.test.tsx:68]: curious why the component is being mocked here

frontend/src/components/workspace/AgencySelector.tsx
```diff
@@ -8,15 +8,17 @@ import { useRouter } from "next/navigation";
...
   agencies: UserAgency[];
   currentAgencyId: string;
+  className?: string;
 }) => {
   const router = useRouter();
   const t = useTranslations("Opportunities");
 
   return (
-    <div className="usa-form-group margin-bottom-4">
+    <div className={className ?? "usa-form-group margin-bottom-4"}>
```
doug-s-nava, line 22: curious why we're using this component and why it isn't using Trussworks components for Label and Select? I would move this directly into the OpportunitiesHeader where it's used, since I don't think it's very reusable

frontend/src/app/[locale]/(base)/opportunities/page.tsx
```diff
@@ -112,17 +113,60 @@ const transformTableRowData = (
...
+          ) : (
+            <AgencySelector
+              agencies={agencies}
+              currentAgencyId={currentAgencyId}
+              className="usa-form-group margin-bottom-0"
+            />
+          )}
+        </div>
+        <Link
+          href={`/opportunities/create/${agencyCode}`}
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunities/page.tsx]: we're going to need to adjust this when we change the create page to take a query param instead of path anyway, but to minimize the rework, this should pass the agency id rather than code since that's what the create page expects

frontend/src/app/[locale]/(base)/opportunities/page.test.tsx
Comment (desaiamit) [reply frontend/src/app/[locale]/(base)/opportunities/page.test.tsx:68]: AgencySelector is a "use client" component that depends on useRouter for navigation. The mock isolates the page-level test from that client-side behavior and provides a data-testid so the single-agency vs multi-agency assertions can cleanly verify whether the selector is present or absent, without depending on specific label text from the i18n mock.

frontend/src/components/workspace/AgencySelector.tsx
desaiamit: Good catch — the HTML components pre-dated this PR and this PR only added the className prop; switched to Label and Select from @trussworks/react-uswds in this commit. 

On inlining: because this component uses useRouter, it requires the "use client" directive at the module level — that boundary can't live inside a server component file like page.tsx, so it has to remain a separate file. Agreed it's not meaningfully reusable though, so it makes sense to keep it co-located with the page rather than in src/components/workspace/

frontend/src/app/[locale]/(base)/opportunities/page.tsx
Comment (desaiamit) [reply frontend/src/app/[locale]/(base)/opportunities/page.tsx]: Fixed — now passes currentAgencyId (the UUID) instead of agencyCode. Removed agencyCode from OpportunitiesHeader's props entirely since it was only used for this link. Tests updated accordingly.

frontend/src/app/[locale]/(base)/opportunities/page.test.tsx
Comment (doug-s-nava) [reply frontend/src/app/[locale]/(base)/opportunities/page.test.tsx:68]: we can talk about this more in sync soon, so not a blocker, but I'd like for us to continue defaulting to mocking functionality rather than full components. While this does not achieve full encapsulation it more directly speaks to what functionality is not under test, and should be less complicated and less verbose. There are lots of examples of mocking useRouter specifically across the app if you search for `jest.mock("next/navigation"`

frontend/src/components/workspace/AgencySelector.tsx
doug-s-nava: do you want to move this now or make a follow up ticket?

PR #8880, opened 2026-03-06
[Issue #8495] Adding custom fields

### LINKED ISSUE #8490: [Custom fields] Update Simpler Grants APIs
### Summary

We are expanding our feature support for custom fields to include having API endpoints that will take our pre-selected fields and populate those values into the customFields property of the response.  The transformation.py file in the Simpler-grants-gov is where we will need to update this code.

### Acceptance criteria

- [ ]  The preselected fields are populating into the customField for the `get` endpoint
- [ ] The preselected fields are populating into the customField for the `search` endpoint

### LINKED ISSUE #8495: [CommonGrants] Update OpenAPI docs to include custom fields
### Summary

We need to update the auto-generated openAPI in the Simpler grants Gov repository documents once initial round of  custom fields are selected and implemented. 

### Note
 This ticket is explicitly for the SGG OpenAPI docs that should include the custom fields that we are adding.

### Acceptance criteria

- [ ] The openAPI documentation has been automatically generated and committed to the repository
- [ ] The examples in the repository have also been updated 
- [ ] Determine whether Marshmallow models are still required for OpenAPI doc generation, or if the newer API Flask version supports generating docs directly from Pydantic models

jcrichlake:
## Summary

Fixes #8495 and fixes #8490

## Changes proposed

This PR adds CustomFields to the `common grants` OpenAPI endpoints  via the updates to the Marshmallow schema.  It also supports populating the custom fields from the SGG datasource into the expected field.  This is done in the `populate_custom_fields` function.  

## Context for reviewers

Part of this PR is that this establishes the pattern for custom fields in the common grants endpoints going forward.  There is a new file added `common_grants_custom_fields.py` that will serve as the container for all custom fields going forward.  

These custom fields can then be imported and registered to any base object that is returned in an endpoint by adding it to the CustomFields Marshmallow class.  Below is an example of the class registration. 

```python
class CustomFields(Schema):
    legacyId = fields.Nested(LegacyId, allow_none=True)
    federalOpportunityNumber = fields.Nested(FederalOpportunityNumber, allow_none=True)
    assistanceListing = fields.Nested(AssistanceListing, allow_none=True)
    agency = fields.Nested(Agency, allow_none=True)
    attachments = fields.Nested(Attachments, allow_none=True)
    category = fields.Nested(Category, allow_none=True)
    fiscalYear = fields.Nested(FiscalYear, allow_none=True)
    costSharing = fields.Nested(CostSharing, allow_none=True)
    additionalInfo = fields.Nested(AdditionalInfo, allow_none=True)
    agencyContact = fields.Nested(AgencyContact, allow_none=True)
    yourNewestField = fields.Nexted(YourNewestField, allow_none=True)

```

As well as registering the custom fields in the schema it also adds the population of the fields from the SGG base data.  This occurs in `populate_custom_fields`.  There  are also changes in  `transform_opportunity_to_cg` which is called from the GET endpoint before the populate call is made.   These changes are to pre-populate some of the dictionary fields to ensure the data is pulled out of the initial request so that it can be added to the eventual response. 

## Validation steps

1. Checkout the branch `jeff-8495-Ad…s-to-openapi`
2. Run `cd /api`
3. Run `make openapi-spec-common-grants` to generate the new openAPI yaml file
4. Run `make init make db-seed-local && make populate-search-opportunities make run-logs` to stand up the API endpoints.
5. Open the `SIMPLER-GRANTS-PROTOCOL` repo and navigate to `lib/python-sdk/examples`6. 
7. Update the `ports` in the  3 custom fields examples to be 8080
8. Update the Custom field examples to not take in an opportunity base parameter and update the print statement to list all custom fields (See below for examples of each file. 
9. from `lib/python-sdk` Run `poetry run python examples/list_opportunities.py and validate the custom fields
10. Grab an Opportunity ID then run `poetry run python examples/get_opportunity.py` repeat validation
11. Run `poetry run python examples/search_opportunities.py` repeat validation

# Updated list_opportunities.py
```python
#!/usr/bin/env python3
"""Example script demonstrating basic client usage.

Run with: poetry run python list_opportunity.py
"""

from common_grants_sdk.client import Client
from common_grants_sdk.client.config import Config
from common_grants_sdk.extensions.specs import CustomFieldSpec
from common_grants_sdk.schemas.pydantic import OpportunityBase, CustomFieldType

config = Config(
    base_url="http://localhost:8080",
    api_key="two_org_user_key",
    timeout=5.0,
    page_size=10,
)
client = Client(config)


fields = {
    "legacyId": CustomFieldSpec(
        field_type=CustomFieldType.INTEGER,
        value=int,
        name="Legacy ID",
        description="Federal id legacy value",
    ),
    "groupName": CustomFieldSpec(field_type=CustomFieldType.STRING, value=str),
}

opp = OpportunityBase.with_custom_fields(custom_fields=fields, model_name="Opportunity")



response = client.opportunities.list(page=1)

print(f"Found {len(response.items)} opportunities:")


for item in response.items:

    if item.custom_fields is not None:

        print(
            f"  - {item.id}: {item.title}, custom field value: {item.custom_fields}"  # type: ignore[union-attr]
        )
```

# Updated get_opportunity.py

```python
#!/usr/bin/env python3
"""Example script demonstrating how to fetch a single opportunity by ID.

Run with: poetry run python get_opportunity.py <oppId>
"""

import sys

from common_grants_sdk.client import Client
from common_grants_sdk.client.config import Config
from common_grants_sdk.extensions.specs import CustomFieldSpec
from common_grants_sdk.schemas.pydantic import OpportunityBase, CustomFieldType

if len(sys.argv) < 2:
    print("Usage: get_opportunity.py <oppId>", file=sys.stderr)
    sys.exit(1)

opp_id = sys.argv[1]
config = Config(
    base_url="http://localhost:8080",
    api_key="two_org_user_key",
    timeout=5.0,
)
client = Client(config)

fields = {
    "legacyId": CustomFieldSpec(
        field_type=CustomFieldType.INTEGER,
        value=int,
        name="legacy_id",
        description="Federal ID Legacy Field",
    ),
    "groupName": CustomFieldSpec(field_type=CustomFieldType.STRING, value=str),
}

opp = OpportunityBase.with_custom_fields(custom_fields=fields, model_name="Opportunity")


opportunity = client.opportunities.get(opp_id)

print(f"Opportunity {opp_id}:")
print(f"  Title: {opportunity.title}")
print(f"  ID: {opportunity.id}")
print(f" Custom Fields: {opportunity.custom_fields}")

```

# Updated search_opportunities.py

```python
"""Example script demonstrating how to search for opportunities.

Run with: poetry run python search_opportunity.py <searchTerm>
"""

import sys

from common_grants_sdk.client import Client
from common_grants_sdk.client.config import Config
from common_grants_sdk.schemas.pydantic.models.opp_status import OppStatusOptions
from common_grants_sdk.extensions.specs import CustomFieldSpec
from common_grants_sdk.schemas.pydantic import OpportunityBase, CustomFieldType

if len(sys.argv) < 1:
    print("Usage: search_opportunity.py <searchTerm>", file=sys.stderr)


config = Config(
    base_url="http://localhost:8080",
    api_key="two_org_user_key",
    timeout=5.0,
    page_size=10,
)

search = sys.argv[1]
client = Client(config)


fields = {
    "legacyId": CustomFieldSpec(field_type=CustomFieldType.INTEGER, value=int),
    "groupName": CustomFieldSpec(field_type=CustomFieldType.STRING, value=str),
}

opp = OpportunityBase.with_custom_fields(custom_fields=fields, model_name="Opportunity")


response = client.opportunities.search(
    search=search, status=[OppStatusOptions.OPEN], page=1
)

print(f"Found {len(response.items)} opportunities: ")

for item in response.items:
    print(
        f" - {item.id}: {item.title} custom field value: {item.custom_fields}"  # type: ignore[union-attr]
    )

```


jcrichlake:
@chouinar  see below for the 3 curl commands to hit each of the endpoints. 

ListOpportunities: ```curl -H 'accept: application/json' -H 'X-Api-Key: two_org_user_key' -X 'GET'  'http://localhost:8080/common-grants/opportunities'```


Get Opportunity: ```curl -H 'accept: application/json' -H 'X-Api-Key: two_org_user_key' -X 'GET'  'http://localhost:8080/common-grants/opportunities/3d862f45-a525-460a-8134-5eeffd877611'```

SearchOpportunities:  ```curl -H 'accept: application/json' -H 'X-Api-Key: two_org_user_key' -X 'POST' 'http://localhost:8080/common-grants/opportunities/search' -d '{ "filters": { "status": { "operator": "in", "value": ["open", "forecasted"] }, "pagination": { "page": 1, "pageSize": 10 }, "search": "education", "sorting": { "sortBy": "lastModifiedAt", "sortOrder": "desc" } }' ```

api/src/services/common_grants/transformation.py
```diff
@@ -310,6 +344,152 @@ def transform_search_result_to_cg(opp_data: dict) -> OpportunityBase | None:
...
+                "lastModifiedAt": attachment.get("updated_at"),
+            }
+            for attachment in attachments
+        ]
+        custom_fields["attachments"] = CustomField(
+            name="attachments",
+            fieldType=CustomFieldType.ARRAY,
+            value=attachment_values,
+            description="Attachments such as NOFOs and supplemental documents for the opportunity",
+        )
```
widal001, line 381: This is probably better solved in the Python SDK itself, but I wanted to flag that I'm getting an IDE warning from Pylance/Pyright when we try to populate `CustomField` without explicitly setting `CustomField.schema`

In the next Python SDK release, we should make the default value for this be `None`.

Image

widal001: It's not touched by this PR, but the same is true for `SingleDateEvent` which raises a type error unless you explicitly pass `SingleDateEvent.eventType`

We should have that default to `EventType.SINGLE_DATE` in the SDK since that's a literal value for that sub-class of `EventBase`

Image

api/src/services/common_grants/transformation.py
```diff
@@ -310,6 +344,152 @@ def transform_search_result_to_cg(opp_data: dict) -> OpportunityBase | None:
         return None
 
 
+def populate_custom_fields(opp_data: dict) -> dict[str, CustomField] | None:
```
widal001, line 348: This doesn't need to be changed in this PR, but we should discuss the pros and cons of creating dedicated schemas for these custom fields, instead of just using the generic `CustomField` (similar to what we're doing for the Marshmallow schemas). Maybe this is a future KT topic with Bryan.

While the current pattern formats the data, we're not actually doing any validation of the data going into these custom fields at the transformation layer (via pydantic), which makes it harder to catch and handle malformed data before we try to serialize it with the Marshmallow models.

The plugin framework we're working on now will make it easier to define and auto-generate those pydantic schemas.

api/tests/src/services/common_grants/test_transformation.py
```diff
@@ -208,6 +344,37 @@ def __init__(self):
                 self.updated_at = datetime(
                     2024, 1, 1, 12, 0, 0
                 )  # Provide a default datetime instead of None
+                self.legacy_opportunity_id = 67890
+                self.opportunity_number = "2024-010"
+                self.category = "Mandatory"
+                self.agency_code = "A2345"
+                self.agency_name = "Testing Agency"
+                self.top_level_agency_name = "Testing top level agency"
+                self.opportunity_assistance_listings = [
```
widal001: Instead of defining this over and over again in each mock test, can we bubble this up to a fixture or even just a separate utility function that returns default values that we can override as needed?

Also I notice that we're setting the same attributes in each test but the values are changing slightly, but there doesn't seem to be a reason for having them be different across tests.

Seeing the same values set over and over again makes it hard to track what's actually be tested.

widal001: Also can you add `assert result is not None` before these other asserts to remove the Pylance warnings?

Image

api/tests/src/services/common_grants/test_transformation.py
```diff
@@ -208,6 +344,37 @@ def __init__(self):
...
+                        {
+                            "assistance_listing_number": "10.557",
+                            "program_title": "Special Supplemental Nutrition Program",
+                        },
+                    )()
+                ]
+                self.opportunity_attachments = [
+                    type(
+                        "MockAttachment",
+                        (),
```
widal001, line 348: Where is this `"MockAttachment"` defined and what purpose is it solving?

I only see the phrase `MockAttachment` in this file. Same  with the other `Mock<CustomField>` examples

api/tests/src/services/common_grants/test_transformation.py
```diff
@@ -791,3 +1069,90 @@ def test_transformation_with_invalid_url_logs_but_succeeds(self, caplog):
             and invalid_url in record.message
             for record in caplog.records
         )
+
+    def test_populate_custom_fields(self):
+        """Test that populate_custom_fields correctly maps all fields from opp_data."""
```
widal001: Can you move this out into its own `TestPopulateCustomFields` class and test a few additional cases, each as a separate method:

1. This base transformation test case, where everything is populated as expected.
2. A test case where some or all of these custom fields are missing values -- we want to check that `None` values are handled correctly.
3. A test case where some of these values have the incorrect type -- we want to check how we handle malformed data types

For test cases 2 and 3 we might need to also load the Marshmallow schemas you created to ensure that malformed data is caught before it gets to those values. Because otherwise we'll likely pass on some internal 500 errors to the API consumers -- which we don't want.

On a related we should have a similar set of rout-level test cases that tests how the new marshmallow models handle invalid data (e.g. does it raise a 500 error or a different error type like 422 or something else)

api/src/api/common_grants/common_grants_custom_fields.py
```diff
@@ -0,0 +1,231 @@
+""" Marshmallow schemas for CommonGrants Protocol customFields
+
+This file contains Marshmallow schemas that are part of the custom field portion of the specification.
+
+NOTE: Once added here the fields should be imported into another Marshmallow file to register them under the CustomFields class there.
+At this time there is only one such file and that is common_grants_schemas.py which only supports Opportunity.py
```
widal001, line 6: Thanks for adding this note!

api/src/api/common_grants/common_grants_schemas.py
```diff
@@ -356,6 +334,24 @@ class OppTimeline(Schema):
     )
 
 
+class CustomFields(Schema):
```
widal001: Small nit but we should probably rename this `OpportunityCustomFields` or something similar to indicate that it's specific to the `OpportunityBase` schema.

api/src/api/common_grants/common_grants_custom_fields.py
```diff
@@ -0,0 +1,231 @@
...
+    )
+
+
+# ===========================================================================================
+# Custom Field Implementations
+# ===========================================================================================
+
+
+class LegacyId(CustomField):
+    """Storing legacy id for compatibility iwth legacy systems"""
```
widal001: I like the pattern you followed here! But we should update these to match the schemas in the custom field catalog. I'll tag each one with the associated custom field page.

[`legacySerialId`](https://commongrants.org/custom-fields/legacySerialId/)

api/src/api/common_grants/common_grants_custom_fields.py
```diff
@@ -0,0 +1,231 @@
...
+    value = fields.Integer(required=True, metadata={"example": "12345"})
+    description = fields.String(
+        allow_none=True,
+        metadata={
+            "example": "An integer ID for the opportunity, needed for compatibility with legacy systems"
+        },
+    )
+
+
+class FederalOpportunityNumber(CustomField):
```
widal001: I think this one looks good but tagging it for reference [`federalOpportunityNumber`](https://commongrants.org/custom-fields/federalOpportunityNumber/)

api/src/api/common_grants/common_grants_custom_fields.py
```diff
@@ -0,0 +1,231 @@
...
+
+
+class AssistanceListing(CustomField):
+    """The assistance listing number and program title for this opportunity"""
+
+    name = fields.String(required=True, metadata={"example": "assistanceListing"})
+    fieldType = fields.String(required=True, metadata={"example": "array"})
+    value = fields.List(fields.Nested(AssistanceListingValue), required=True)
+    description = fields.String(
+        allow_none=True,
```
widal001, line 93: [`assistanceListings`](https://commongrants.org/custom-fields/assistanceListings/) Note the change in the `customFields` key (with the `s` on the end to indicate it's an array field)

api/src/api/common_grants/common_grants_custom_fields.py
```diff
@@ -0,0 +1,231 @@
...
+
+    agencyCode = fields.String(required=True, metadata={"example": "US-ABC"})
+    agencyName = fields.String(allow_none=True, metadata={"example": "Department of Examples"})
+    topLevelAgencyName = fields.String(
+        allow_none=True, metadata={"example": "Department of Examples"}
+    )
+
+
+class Agency(CustomField):
+    """Information about the agency offering this opportunity"""
```
widal001, line 111: [`agency`](https://commongrants.org/custom-fields/agency/)

api/src/api/common_grants/common_grants_custom_fields.py
```diff
@@ -0,0 +1,231 @@
...
+class Category(CustomField):
+    """The category type of the grant opportunity"""
+
+    name = fields.String(required=True, metadata={"example": "category"})
+    fieldType = fields.String(required=True, metadata={"example": "string"})
+    value = fields.String(required=True, metadata={"example": "discretionary"})
+    description = fields.String(
+        allow_none=True,
+        metadata={"example": "The category type of the grant opportunity"},
+    )
```
widal001: [`federalFundingSources`](https://commongrants.org/custom-fields/federalFundingSources/)

api/src/api/common_grants/common_grants_custom_fields.py
```diff
@@ -0,0 +1,231 @@
...
+        metadata={"example": "For more information, reach out to Jane Smith at agency US-ABC"},
+    )
+    emailAddress = fields.String(allow_none=True, metadata={"example": "fake_email@grants.gov"})
+    emailDescription = fields.String(
+        allow_none=True, metadata={"example": "Click me to email the agency"}
+    )
+
+
+class AgencyContact(CustomField):
+    """Contact information for the agency managing this opportunity"""
```
widal001, line 178: [`contactInfo`](https://commongrants.org/custom-fields/contactInfo/)

api/src/api/common_grants/common_grants_custom_fields.py
```diff
@@ -0,0 +1,231 @@
...
+
+class AdditionalInfoValue(Schema):
+    """Schema for populating the AdditionalInfo value field"""
+
+    url = fields.String(allow_none=True, metadata={"example": "grants.gov"})
+    description = fields.String(allow_none=True, metadata={"example": "Click me for more info"})
+
+
+class AdditionalInfo(CustomField):
+    """URL and description for additional information about the opportunity"""
```
widal001, line 198: [`additionalInfo`](https://commongrants.org/custom-fields/additionalInfo/)

api/src/api/common_grants/common_grants_custom_fields.py
```diff
@@ -0,0 +1,231 @@
...
+        allow_none=True,
+        metadata={
+            "example": "URL and description for additional information about the opportunity"
+        },
+    )
+
+
+class CostSharing(CustomField):
+    """Whether cost sharing or matching funds are required for this opportunity"""
+
```
widal001: [`costSharing`](https://commongrants.org/custom-fields/costSharing/)

api/src/api/common_grants/common_grants_custom_fields.py
```diff
@@ -0,0 +1,231 @@
...
+    description = fields.String(
+        allow_none=True,
+        metadata={
+            "example": "Whether cost sharing or matching funds are required for this opportunity"
+        },
+    )
+
+
+class FiscalYear(CustomField):
+    """The fiscal year associated with this opportunity"""
```
widal001: I'm okay with leaving this in, but just wanted to note that we decided to take it out of the catalog after Laura's analysis on the frequency with which it's populated.

api/tests/src/services/common_grants/test_transformation.py
jcrichlake: It's mocking the `opportunity_attachments` custom fields  property.

api/src/api/common_grants/common_grants_custom_fields.py
widal001: One thing I'm noticing as I cross-reference this with the existing `OpportunityV1.category` field in the SGG API is that it only accepts a single value right now.

I'll update this in the CommonGrants catalog so the main change will be `federalFundingSource` and it will be just a single object instead of an array.

api/tests/src/services/common_grants/test_transformation.py
```diff
@@ -791,3 +1012,142 @@ def test_transformation_with_invalid_url_logs_but_succeeds(self, caplog):
...
+        # Serialize to the format the Marshmallow schema expects and validate
+        serialized = {k: v.model_dump(by_alias=True) for k, v in result.items()}
+        schema = OpportunityCustomFields()
+        errors = schema.validate(serialized)
+
+        # Marshmallow should catch the malformed integer fields
+        assert "legacySerialId" in errors
+        assert "value" in errors["legacySerialId"]
+        assert "fiscalYear" in errors
+        assert "value" in errors["fiscalYear"]
```
widal001: I don't think this is quite testing the behavior we want.

The malformed data should be caught in `populate_custom_fields()` and omitted _before_ we try to validate it with Marshmallow so that Marshmallow doesn't raise any errors.

api/tests/src/api/common_grants/test_opportunities_routes.py
```diff
@@ -394,3 +395,57 @@ def test_search_opportunities_response_schema(
...
+
+    def test_response_schema_error_returns_500(
+        self,
+        client: FlaskClient,
+        enable_factory_create,
+        db_session,
+        user_api_key_id,
+        monkeypatch,
+    ):
+        """Marshmallow ValidationError raised during response loading → 500 with runtime message."""
```
widal001, line 433: Paired with [this test](https://github.com/HHS/simpler-grants-gov/pull/8880/changes#r2940083462), it _looks_ like if we let malformed data pass through to the marshmallow models that are used for serialization, then we'd raise a 500 error for API consumers _any_ time an opportunity record contains a custom field with a malformed data type.

500 errors should only ever be raised if it is truly impossible for a server to recover from an event, not something that we expect to happen with some amount of frequency like misshapen data.

If you manually deployed this branch to `dev` without merging it, then I'd expect you'd find we get a _lot_ of 500 errors, especially on the opportunity list and search endpoints.

api/src/api/common_grants/common_grants_custom_fields.py
widal001: The fields in the `agency.value` section still don't match the ones in the catalog.

Right now it's

```json
"agency": {
  "description": "Information about the agency offering this opportunity",
  "fieldType": "object",
  "name": "agency",
  "schema": null,
  "value": {
    "agencyCode": "HHS-IHS",
    "agencyName": "Indian Health Service",
    "topLevelAgencyName": "Department of Health and Human Services"
  }
},
```

And it should be:

```json
"agency": {
  "name": "agency",
  "fieldType": "object",
  "value": {
    "code": "CMS",
    "name": "Centers for Medicare & Medicaid Services",
    "parentCode": "HHS",
    "parentName": "Health and Human Services"
  },
  "description": "Information about the agency offering this opportunity"
}
```

api/src/services/common_grants/transformation.py
```diff
@@ -310,6 +344,152 @@ def transform_search_result_to_cg(opp_data: dict) -> OpportunityBase | None:
...
+        custom_fields["agency"] = CustomField(
+            name="agency",
+            fieldType=CustomFieldType.OBJECT,
+            value={
+                "agencyCode": agency,
+                "agencyName": opp_data.get("agency_name"),
+                "topLevelAgencyName": opp_data.get("top_level_agency_name"),
+            },
+            description="Information about the agency offering this opportunity",
+        )
```
widal001, line 438: These would also need to get changed to match the values in the marshmallow schema once they're updated to match [`agency`](https://commongrants.org/custom-fields/agency/)

api/tests/src/api/common_grants/test_opportunities_routes.py
widal001: You can also simulate this issue locally by simply changing the value of the data coming out of `populate_custom_fields()` to an invalid type.

For example. By wrapping `legacy_opportunity_id` in a string that can't be coerced into an integer:

Image

You get a 500 error when you call the list opportunities endpoint:

Image

Image

api/src/services/common_grants/transformation.py
jcrichlake: Fixed

jcrichlake:
@widal001 sorry about that with the agency fields, those should be updated as of my most recent !!

Issue #8934, opened 2026-03-10
SOAP/Proxy: fix issue with capitalization in serial numbers


jakobpederson:
### Summary

Currently capitalization on serial numbers affects search. They can be capitalized in the SG DB, which means they're stored capitalized in our DB, but the algorithm that gets the serial numbers from the post requests are lower case. This is to make sure that we ignore case.

This code gets the serial number from the cert string:
```
    cert_str = unquote(urlencoded_cert)
    cert = load_pem_x509_certificate(cert_str.encode(), default_backend())
    serial_number_hex = format(int(cert.serial_number), "032x")
```

But that cert is lowercased. In the event that the serial number is upper-cased in the DB, this search will fail:
```
    legacy_certificate = db_session.execute(
        select(LegacyCertificate).where(LegacyCertificate.serial_number == str(serial_number_hex))
    ).scalar_one_or_none()
```

### Acceptance criteria

- [ ] unittests pass
- [ ] case does not impact serial number search

PR #8935, opened 2026-03-10
[Issue #8934] Ignore case in legacy certificate search

jakobpederson:
## Summary

Fixes / Work for #8934  

## Changes proposed

Update the `LegacyCertificate` search to ignore case of `serial_number`.

## Context for reviewers

Serial numbers can be upper case or lower case, but when they are extracted from an incoming cert they are lower case. This makes it so that the `LegacyCertificate` search ignores case.

## Validation steps

- [x] Can curl the training env with
```
--cert "/Users/$(whoami)/Downloads/grants_s2s_soap_certs/localsetup/staging_grantors_soap_api.crt" \
--key "/Users/$(whoami)/Downloads/grants_s2s_soap_certs/localsetup/staging_grantors_soap_api.key" \
```
And get a valid response.

### DIFF
--- a/api/src/legacy_soap_api/legacy_soap_api_auth.py
+++ b/api/src/legacy_soap_api/legacy_soap_api_auth.py
@@ -10,7 +10,7 @@ from cryptography.hazmat.primitives import hashes
 from cryptography.x509 import load_pem_x509_certificate
 from pydantic import BaseModel, ConfigDict
 from requests.adapters import HTTPAdapter
-from sqlalchemy import select
+from sqlalchemy import func, select
 
 import src.adapters.db as db
 from src.db.models.user_models import LegacyCertificate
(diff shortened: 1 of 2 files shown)


api/src/legacy_soap_api/legacy_soap_api_auth.py
```diff
@@ -109,15 +109,29 @@ def get_soap_auth(mtls_cert: str | None, db_session: db.Session) -> SOAPAuth | N
...
+
+
 def get_soap_client_certificate(
     urlencoded_cert: str, db_session: db.Session
 ) -> SOAPClientCertificate:
     cert_str = unquote(urlencoded_cert)
     cert = load_pem_x509_certificate(cert_str.encode(), default_backend())
+    # We convert the integer representation of the serial number to lower case hexadecimal here
+    # Note: the 'x' in '032x' returns it lower cased
     serial_number_hex = format(int(cert.serial_number), "032x")
```
freddieyebra, line 133: If we verify that all serial numbers in the db are uppercase, could we just change the format to `"032X"` instead of `"032x"`? Additionally, can we use the integer value to find the cert in `legacy_certificate.cert_id`/`staging.tcertificates.currentcertid`? it looks like that column holds the integer representation of the serial number.

jakobpederson: The serial numbers in the DB are a mix of uppercase and lowercase. If we're sure that the `currentcertid` is the intefied version of the `serial_num` then I think that's the actual best solution.

Issue #8674, opened 2026-02-26
AR data model


kevlyu:
### Summary

create ar data model

### Acceptance criteria

tbd

PR #8946, opened 2026-03-10
[Issue #8674] Create core Award Recommendation tables

nathan-stricker:
## Summary

Work for #8674 

## Changes proposed

Database updates for Award Recommendation work.
- Tables and factories added:
   - `award_recommendation`
   - `award_recommendation_application_submission`
   - `award_recommendation_submission_detail`

## Context for reviewers

See the Award Recommendation tech spec for specifications for the tables created:
- [Award Recommendation](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Award-Recommendation)
- [Award Recommendation Application Submission](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Award-Recommendation-Application-Submission)
- [Award Recommendation Submission Detail](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Award-Recommendation-Submission-Detail)

Also note: it was suggested we add a relationship between Award Recommendation to `funding_strategy`, but this does not yet exist and will be added later.

## Validation steps

No manual verification; added Factory methods for new tables.

### DIFF
--- a/api/src/db/migrations/versions/2026_03_10_add_core_award_recommendation_tables.py
+++ b/api/src/db/migrations/versions/2026_03_10_add_core_award_recommendation_tables.py
@@ -1,8 +1,8 @@
 """add core award recommendation tables
 
-Revision ID: caec4a3aa8d1
+Revision ID: dc49bb9e25ab
 Revises: 21761f4206c4
-Create Date: 2026-03-10 16:54:52.190638
+Create Date: 2026-03-10 20:24:20.040032
 
 """
 
@@ -10,7 +10,7 @@ import sqlalchemy as sa
 from alembic import op
 
 # revision identifiers, used by Alembic.
-revision = "caec4a3aa8d1"
+revision = "dc49bb9e25ab"
 down_revision = "21761f4206c4"
 branch_labels = None
 depends_on = None
@@ -56,14 +56,13 @@ def upgrade():
         "award_recommendation",
         sa.Column("award_recommendation_id", sa.UUID(), nullable=False),
         sa.Column("opportunity_id", sa.UUID(), nullable=False),
-        sa.Column("funding_strategy_id", sa.UUID(), nullable=True),
         sa.Column("award_recommendation_number", sa.Text(), nullable=False),
         sa.Column("award_recommendation_status_id", sa.Integer(), nullable=False),
         sa.Column("additional_info", sa.Text(), nullable=True),
         sa.Column("award_selection_method_id", sa.Integer(), nullable=True),
-        sa.Column("selection_method_detail", sa.Text(), nullable=True),
-        sa.Column("funding_strategy", sa.Text(), nullable=True),
         sa.Column("other_key_information", sa.Text(), nullable=True),
+        sa.Column("is_deleted", sa.Boolean(), nullable=False),
+        sa.Column("review_workflow_id", sa.UUID(), nullable=True),
         sa.Column(
             "created_at",
             sa.TIMESTAMP(timezone=True),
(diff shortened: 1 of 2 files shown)


api/src/constants/lookup_constants.py
```diff
@@ -245,6 +245,7 @@ class AwardRecommendationStatus(StrEnum):
     DRAFT = "draft"
     IN_REVIEW = "in_review"
     APPROVED = "approved"
+    IN_PROGRESS = "in_progress"
```
chouinar: Why did we add this status? Draft is the in-progress status.

api/src/db/models/opportunity_models.py
```diff
@@ -505,6 +519,100 @@ class OpportunityChangeAudit(ApiSchemaTable, TimestampMixin):
     is_loaded_to_version_table: Mapped[bool | None] = mapped_column(index=True)
 
 
+class AwardRecommendation(ApiSchemaTable, TimestampMixin):
```
chouinar: Can we move the award recommendation tables to their own file `award_recommendation_models.py`? Note that you'll need to import these in the `__init__.py` file in this directory so alembic picks them up.

api/src/db/models/opportunity_models.py
```diff
@@ -505,6 +519,100 @@ class OpportunityChangeAudit(ApiSchemaTable, TimestampMixin):
...
+    __tablename__ = "award_recommendation"
+
+    award_recommendation_id: Mapped[uuid.UUID] = mapped_column(
+        UUID, primary_key=True, default=uuid.uuid4
+    )
+    opportunity_id: Mapped[uuid.UUID] = mapped_column(
+        UUID, ForeignKey(Opportunity.opportunity_id), nullable=False, index=True
+    )
```
chouinar: Can we leave out `nullable=X`? SQLAlchemy will infer it from the python typing, if we specify the same thing twice, but mess it up, it won't get recognized and can lead to issues.

api/src/db/models/opportunity_models.py
```diff
@@ -505,6 +519,100 @@ class OpportunityChangeAudit(ApiSchemaTable, TimestampMixin):
...
+    opportunity: Mapped[Opportunity] = relationship(
+        Opportunity, back_populates="award_recommendations"
+    )
+    # TODO - add a relationship to the funding strategy
+    funding_strategy_id: Mapped[uuid.UUID | None] = mapped_column(UUID, nullable=True)
```
chouinar: Ah, this is a holdover from an earlier version, missed removing it from the tech spec table. We can delete this. The funding strategy info used to be a separate table, but I moved all the fields to this table.

I apologize, I actually seem to have made a few mistakes, as a few things in my diagram didn't get ported over to the tech spec for this table. We also want:

* `is_deleted` - boolean, not null, default False
* `review_workflow_id` - UUID, nullable, foreign key to the workflow table
* I think `selection_method_detail` isn't needed anymore?

api/tests/src/db/models/factories.py
```diff
@@ -940,6 +943,71 @@ class Meta:
...
+class AwardRecommendationSubmissionDetailFactory(BaseFactory):
+    class Meta:
+        model = opportunity_models.AwardRecommendationSubmissionDetail
+
+    award_recommendation_submission_detail_id = Generators.UuidObj
+
+    recommended_amount = sometimes_none(
+        factory.Faker("pydecimal", left_digits=7, right_digits=2, positive=True)
+    )
+    scoring_comment = sometimes_none(factory.Faker("paragraph"))
```
chouinar: I think for the prototype the scores were all going to be random numbers between 1-100 but as strings - so for test data probably want the same just so local development that uses these sees something similar?

api/src/db/models/opportunity_models.py
nathan-stricker: No worries, thanks for the clarification! This is fixed in the new version.

PR #8951, opened 2026-03-10
[Issue #8845] Refactor `returnToGrants` link

### LINKED ISSUE #8845: New "Return to Grants.gov" pattern
### Summary

The `returnToGrants` link is meant to give users who arrived on SGG a convenient way back to GG. Links to SGG search from GG's global navigation include a query parameter (`?utm_source=Grants.gov`) that sets a boolean in the current session. It was specifically meant for the search page and opportunities. But, as-implemented, it follows you around the site once the boolean is set. It shows on every page that page includes breadcrumbs — even pages like the Newsletter pages that don't logically need to help the user "return" to GG. 

Instead of the `returnToGrants` link being part of the `<Breadcrumbs>` component, it should be easy for users to get to GG from any page, regardless of whether it has breadcrumbs. It does not make sense in the source order (it's follows the breadcrumbs), needs to be placed in a better location in the layout, is oddly positioned on mobile, and it breaks the layout when breadcrumbs fill the full width of the page: 

Image

The `returnToGrants` link is an antipattern that tries to replicate functionality of the browser's back button. GG has also implemented its link to SGG as a redirect that breaks browser history so that the browser back button does not function (see [bug discussed in Slack thread](https://betagrantsgov.slack.com/archives/C08MTT10SKU/p1765901544495809)).

**Solution:** 
- [Loom showing the replacement](https://www.loom.com/share/89852c186f8a49918c4b6701687b914e)
- [Figma draft](https://www.figma.com/design/bQUTJCMj0JYbAAnJQ4p4tX/Grantee-Designs-%7C-Quad-5-%7C-Simpler-Grants-2026?node-id=3560-1711&t=K6UWWF3xzClp2eMi-0)

Image

### Acceptance criteria

- [ ] Remove the `returnToGrants` link from our custom Breadcrumbs component 
- [ ] Deprecate the `?utm_source=Grants.gov` query parameter
- [ ] Ask MicroHealth to link directly to Simpler Search w/o the query param (this is not necessary to resolve this Issue, but the query param will do nothing) 
- [ ] Replace the `returnToGrants` functionality with a global action alongside the site header for navigating to to GG

andycochran:
## Summary

Fixes #8845 

## Changes proposed

- Delete the "Return to Grants.gov" link 
  - rm `ReturnToGrantsNotification` component and test files
  - Nix `Suspense` and `ReturnToGrantsNotification` from the `<Breadcrumbs>` component 
  - Nix the unnecessary flex classes and container around the breadcrumbs, since the "Return to GG" link no longer 
needs to sit beside it in the layout
  - Nix the `utm_source` query parameter, which is no longer needed to conditionally show the "Return to GG" link 
- rm the unnecessary `SearchCallToAction` component (which was just the `h1` and "Return to GG" link), and move Search's `h1` directly into the page
- Add a new `ClassicSearchBanner` component that's always visible on the search page below the site header 
- Adjust margin/padding of `/search` heading 
- Add basic test coverage for search page heading and a11y

image

image

## Context for reviewers

This PR is the first step in refactoring the Breadcrumbs component so that all pages are managing them consistently. Removing the "Return to GG" link from Breadcrumbs will make that easier. 

## Validation steps

- Run the app and check that all pages with breadcrumbs are still working properly: 
  - The breadcrumbs should still be there
  - The "Return to GG" link should not be there 
- The Search page (only) should display a new banner that inform users that Simpler is the new way to search and they must sign in on GG to access classic search


frontend/src/app/[locale]/(base)/search/page.tsx
```diff
@@ -77,9 +77,21 @@ function Search({ searchParams, params }: SearchPageProps) {
...
         newRelicEnabled={environment.NEW_RELIC_ENABLED === "true"}
       />
-      <div className="bg-base-lightest padding-top-5">
+      <div className="bg-primary-lightest line-height-body-2 font-body-3xs padding-y-1">
         <div className="grid-container">
-          <SearchCallToAction />
+          {t.rich("goToGG", {
+            "search-link": (chunks) => <Link href="/search">{chunks}</Link>,
+            "gg-link": (chunks) => (
+              <a href="https://grants.gov/" target="_blank">
```
Comment (btabaska) [inline frontend/src/app/[locale]/(base)/search/page.tsx]: One thing we’ll want to fix before merge: this external link uses target="_blank" without rel="noopener noreferrer". That can expose us to reverse-tabnabbing. Let’s add the rel attrs on this anchor.

Example:
`<a
                    href={}
                    target="_blank"
                    rel="noopener noreferrer"
                    className="usa-link--external"
                  >`

frontend/src/app/[locale]/(base)/search/page.tsx
```diff
@@ -77,9 +77,21 @@ function Search({ searchParams, params }: SearchPageProps) {
         params={resolvedSearchParams}
         newRelicEnabled={environment.NEW_RELIC_ENABLED === "true"}
       />
-      <div className="bg-base-lightest padding-top-5">
+      <div className="bg-primary-lightest line-height-body-2 font-body-3xs padding-y-1">
         <div className="grid-container">
-          <SearchCallToAction />
+          {t.rich("goToGG", {
```
Comment (btabaska) [inline frontend/src/app/[locale]/(base)/search/page.tsx]: This refactor implements new behavior in t.rich("goToGG") with both internal and external links, but I don’t see replacement coverage for this flow. Would be super helpful to add a test that asserts the banner renders and that the Grants.gov link includes the expected safe external-link attrs.

frontend/src/app/[locale]/(base)/search/page.tsx
```diff
@@ -77,9 +77,21 @@ function Search({ searchParams, params }: SearchPageProps) {
...
+            "search-link": (chunks) => <Link href="/search">{chunks}</Link>,
+            "gg-link": (chunks) => (
+              <a href="https://grants.gov/" target="_blank">
+                {chunks}
+              </a>
+            ),
+          })}
+        </div>
```
Comment (btabaska) [inline frontend/src/app/[locale]/(base)/search/page.tsx]: Nit: I could be wrong, but since this is an important banner-style message, have we thought about giving it alert semantics (for example a USWDS alert/site-alert pattern)? Right now it’s just plain text in a div, so screen reader folks may not get any extra signal that this is important context.

btabaska:
Totally makes sense to remove the old component and its tests, but we did drop all coverage around the “return to Grants.gov” messaging in the same PR. Let’s make sure we replace that with search-page-level tests so we still cover the user-facing path end to end.

frontend/src/types/search/searchQueryTypes.ts
```diff
@@ -27,7 +27,6 @@ export const validSearchQueryParamKeys = [
 // an allow list for any paramaters we expect and would like to keep track of in New Relic or elsewhere
 export const expectedQueryParamKeys = [
   ...validSearchQueryParamKeys,
-  "utm_source",
```
doug-s-nava, line 30: removing this will remove the ability to track `utm_source` params in New Relic. Just confirming that we're ok with that? cc @joshtonava

andycochran: Thanks for this catch. I'll add that back in. There's no need to delete it until it's not being used by GG.

frontend/src/app/[locale]/(base)/search/page.tsx
Comment (andycochran) [reply frontend/src/app/[locale]/(base)/search/page.tsx]: Even though this is styled very similarly to the `usa-alert ` component, I don't think it's an alert. 

Do you think this _needs_ to be an alert? Or is the problem that it's a new pattern that looks similar to an alert? 

I don't see much value in using `usa-alert `, as we'd then have to override many of its styles (the left border, content width/container, font size, etc…). And the only thing it'd get us is an ARIA `role`, but the `usa-alert ` doesn't _require_ an ARIA `role` unless you wanna raise the importance of its message. If we do want to make this an alert, I could be convinced to add `role="region"`. But our `<SimplerAlert>` component only has two roles: `"error"` or `"alert"`. And this message doesn't rise to either of those levels of importance. 

I don't see much value in refactoring our `<SimplerAlert>` component and overriding all its styles for this piece of content (which is a temporary solution btw). I see this as similar to the "Official website…" `usa-banner`, which does not have an ARIA `role`.

andycochran:
Moved the banner to a component and added tests. @btabaska I think I've addressed all your concerns. Re-review?

frontend/src/app/[locale]/(base)/search/page.tsx
Comment (andycochran) [reply frontend/src/app/[locale]/(base)/search/page.tsx]: Added minimal tests (tho there's no good way to test for external-link attrs in intl'd content).

andycochran:
~This is "blocking" (kinda) #9027, or at least making it more complicated.~

frontend/src/components/search/ClassicSearchBanner.tsx
```diff
@@ -0,0 +1,31 @@
...
+import Link from "next/link";
+import { GridContainer } from "@trussworks/react-uswds";
+
+const ClassicSearchBanner = () => {
+  const t = useTranslations("Search");
+
+  return (
+    <div className="bg-primary-lightest line-height-body-2 font-body-3xs padding-y-1">
+      <GridContainer>
+        {t.rich("goToGG", {
```
doug-s-nava, line 13: if you want this to be more easily testable you can do it without t.rich

```suggestion
          <GridContainer>
            <Link href="/search">{t("search")}</Link>
            <a
              href={ExternalRoutes.GRANTS_HOME}
              target="_blank"
              rel="noopener noreferrer"
              className="usa-link--external"
            >
              {t.goToGG}
            </a>
          </GridContainer>


```

frontend/src/app/[locale]/(base)/search/page.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+    locale: "en",
+  })),
+}));
+
+jest.mock("next-intl/server", () => ({
+  getTranslations: () => identity,
+  setRequestLocale: identity,
+}));
+
+describe("Search", () => {
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/search/page.test.tsx]: ok this one took me a minute and it's one of the reasons I hate default exports. This test is actually testing the home page, because the component imported is pointing to the homepage file rather than the search page. This fails if you run it against the search apge.

this is tricky, take a look at what this test used to look like before we deleted it. https://github.com/HHS/simpler-grants-gov/pull/4304/changes#diff-ac24cc7e46b23a04ffb13417d5c23e02bda262e77d1574b62504dc6819993497

I know there is a desire to get some testing back here, but I think we should be careful with it. Are you up to getting this back up and running as part of this PR?

frontend/src/components/search/ClassicSearchBanner.tsx
```diff
@@ -0,0 +1,31 @@
...
+              {chunks}
+            </a>
+          ),
+        })}
+      </GridContainer>
+    </div>
+  );
+};
+
+export default ClassicSearchBanner;
```
doug-s-nava: normally I wouldn't care but seeing as I just lost 20 minutes to tracking down an issue with default exports, can you do this as a named export?

frontend/src/components/search/ClassicSearchBanner.tsx
andycochran: I'm confused. The links are in the middle of the sentence. w/o `rich`, I'd need a string for each segment of the sentence, no?

frontend/src/app/[locale]/(base)/search/page.test.tsx
Comment (andycochran) [reply frontend/src/app/[locale]/(base)/search/page.test.tsx]: That's reason enough for me to pull this test out of this PR. The new `ClassicSearchBanner` component has it's own test, and this test (intended for search) isn't really testing anything crucial for search.

frontend/src/components/search/ClassicSearchBanner.tsx
andycochran: done

frontend/src/app/[locale]/(base)/search/page.test.tsx
Comment (andycochran) [reply frontend/src/app/[locale]/(base)/search/page.test.tsx]: Nixed this file. If we wanna resurrect search tests, we'll do that separate from this.

frontend/src/components/search/ClassicSearchBanner.tsx
andycochran: Only other way I see of making test more robust would be mocking the content. But IDK if that's worth doing. TBH, this banner is a temporary solution, and I don't mind how minimal its test is — the content renders and has no accessibility errors. You okay with this as-is?

doug-s-nava: ah, true. Can leave as is

Issue #8953, opened 2026-03-10
Make analytics data inserts work line by line as first test of not getting OOM killed


mdragon:
### Summary

We're running out of memory so switch to processing the files into the DB line by line to see what if anything that changes. Expect this will slow things down but we can optimize from here if necessary.

### Acceptance criteria

- Job doesn't crash with Out of Memory errors

mdragon:
Working as intended in Prod, no memory errors since this went live.

PR #8954, opened 2026-03-10
[Issue #8953] Ingest files line by line

mdragon:
## Summary

 Work for #8953

## Changes proposed

Rather than reading and passing the entire file contents to the SQL COPY command, run it through line by line

## Context for reviewers

Job has been failing with OOM errors even with 2GB (or 8GB not clear) of RAM, despite largest file only being about 300MB

## Validation steps

- Unit tests have been run successfully
- Checks pass
- Promote to main and see if it fixes things with the Release tomorrow

### DIFF
--- a/analytics/src/analytics/integrations/extracts/load_opportunity_data.py
+++ b/analytics/src/analytics/integrations/extracts/load_opportunity_data.py
@@ -77,7 +77,7 @@ def _fetch_insert_opportunity_data(conn: Connection) -> None:
             )
             copy = stack.enter_context(cursor.copy(query))
 
-            while data := file.read():
-                copy.write(data)
+            for line in file:
+                copy.write(line)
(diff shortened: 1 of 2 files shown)


analytics/src/analytics/integrations/extracts/load_opportunity_data.py
```diff
@@ -77,7 +77,7 @@ def _fetch_insert_opportunity_data(conn: Connection) -> None:
             )
             copy = stack.enter_context(cursor.copy(query))
 
-            while data := file.read():
-                copy.write(data)
+            for line in file:
+                copy.write(line)
```
chouinar, line 81: I'm not too worried about performance for a daily job of a niche feature, but we should watch if the performance is significantly affected.

PR #8963, opened 2026-03-10
[Issue #8846] remove newsletter breadcrumbs

### LINKED ISSUE #8846: Remove breadcrumbs from /newsletter pages
### Summary

For both the confirmation page and the unsubscribe page, breadcrumbs are completely unnecessary. If the user wants to go back to the signup page from the confirmation page, there's the back button, and the link in the global navigation. The unsubscribe page content includes a link to the parent ("Change your mind? [Re-subscribe](https://simpler.grants.gov/newsletter)"). If the user wants to go back to the Home page, it's in the global navigation.

### Acceptance criteria

- [ ] Nix the breadcrumbs on all Newsletter pages

andycochran:
## Summary

Fixes #8846 

## Changes proposed

- Remove breadcrumbs component from the newsletter pages
- Add padding to page content (since breadcrumbs aren't there) 
- Delete the newsletter breadcrumbs constants 

(left=local / right=prod) 

image

image

## Context for reviewers

This is part of a larger effort to remove unnecessary breadcrumbs.  

## Validation steps

- Load the newsletter subpages (you don't have to actually subscribe or unsubscribe; you can go directly to `/newsletter/confirmation` and `/newsletter/unsubscribe`) 
- View that neither page displays breadcrumbs


btabaska:
This looks super clean overall. Only nit is that the more iteration is done, the more I notice that the `newsletter/unsubscribe` and the `newsletter/confirmation` components share a lot of code, especially on the layout section from lines 26-37. Would it make sense to share a small wrapper component for these newsletter result pages so we keep spacing/content structure consistent as folks iterate?

andycochran:
> Would it make sense to share a small wrapper component for these newsletter result pages…

I'm not too concerned about reusability here. Hopefully any iteration on the newsletter pages will happen in conjunction with refactoring for a service other than Sendy, and we'll need to change a lot.

PR #8960, opened 2026-03-10
[Issue #8184] allow usage of playwright tags in e2e test jobs

doug-s-nava:
## Summary

Fixes for #8184

## Changes proposed

Adds support for limiting which playwright tests are run in our e2e testing ci jobs by passing playwright tags.

## Context for reviewers

Includes definition of groups to two tests - these are meant to facilitate verification and manual testing of this PR, but the assigned groups are also the correct groups for these, so leaving them in the PR.

Also cleans up the implementation of the "api_logs" param in the local ci job, which was previously accepting free text

## Validation steps

1. note that an experimental / test only job definition can be found in the first commit on this PR
2. open up previous runs of the experimental job at https://github.com/HHS/simpler-grants-gov/actions/workflows/e2e-experiment.yml
3. _VERIFY_: the first run of the job was run with no tags assigned, and proceeded to run all tests
4. _VERIFY_: the first run of the job was run with a single tag assigned, and proceeded to run one test
5. _VERIFY_: the first run of the job was run with a two tags assigned, and proceeded to run two tests

If you'd like to test this change yourself, feel free to add the experimental job back and play around with it

Issue #8942, opened 2026-03-10
Run staging targeted e2e tests with login only on chrome


doug-s-nava:
### Summary

when running e2e tests against staging, any tests that require login all use the same account. When running the tests against all 4 browser configurations in parallel, this creates a situation where logins will usually fail, due to the user already being logged in elsewhere.

To work around this until we find a better solution (either multiple accounts or login spoofing), we will run all staging targeted tests that require login to run only against Chrome

### Acceptance criteria

- [ ] all tests that require login contain logic such that they will only run against Chrome when targeting staging

Bhavna-Ramachandran:
PR #8965 for this task is under review.

Bhavna-Ramachandran:
PR #8965 - for this task has been reviewed, approved and merged.

PR #8965, opened 2026-03-10
[Issue #8942] Split staging E2E workflow: sharded no-auth, single-worker auth

Bhavna-Ramachandran:
## Summary

Work for : Task #8942 

## Changes proposed
- Tests that require Login.gov authentication now contain logic to run only against Chrome when targeting staging.
- The staging workflow has been split into two jobs:
  - e2e-tests-no-auth-sharded : Runs in parallel shards similar to local E2E runs.
  - e2e-tests-auth-single-worker : Runs auth-related tests in a single shard to avoid parallel Login.gov sessions.

## Context for reviewers
Due to Staging Login.gov limitations - authentication tests cannot run in parallel. Splitting the workflow ensures non-auth tests still benefit from parallelization while auth tests run safely in a single worker.

During this PR, we noticed that - Even with serially run tests, there were few more challenges with Login.gov OTP codes, this PR mainly addresses the fix for these challenges + extra tuning on other tests.

- Login.gov OTPs are valid for 30 seconds and single-use. In longer test runs or in serially run tests:
  - By the time a Test in series reaches the MFA step, the generated code becomes invalid before login.gov processes it or
  - If a test was faster the same code used in previous test login may be attempted again by another test (as the OTP was still valid), which login.gov rejects even if it is still within the same window.

This PR ensures a fresh OTP is generated and handles retry logic if the code is rejected, making the login flow more stable for the E2E test run.

## Credit
Tagging approach inspired by @doug-s-nava's  PR #8960 , where tags are passed via the test options object instead of embedding them in the test name.

### DIFF
--- a/frontend/tests/e2e/saved-opportunities.spec.ts
+++ b/frontend/tests/e2e/saved-opportunities.spec.ts
@@ -19,7 +19,7 @@ test("Saved opportunities page shows unauthenticated state if not logged in", as
 });
 
 // Skip non-Chrome browsers in staging
-test.beforeEach((_, testInfo) => {
+test.beforeEach(({ page: _ }, testInfo) => {
   if (targetEnv === "staging") {
     test.skip(
       testInfo.project.name !== "Chrome",
--- a/frontend/tests/e2e/apply/failure-path-sf424b.spec.ts
+++ b/frontend/tests/e2e/apply/failure-path-sf424b.spec.ts
@@ -25,7 +25,7 @@ const sf424bErrors = [
 ];
 
 // Skip non-Chrome browsers in staging
-test.beforeEach((_, testInfo) => {
+test.beforeEach(({ page: _ }, testInfo) => {
   if (targetEnv === "staging") {
     test.skip(
       testInfo.project.name !== "Chrome",
--- a/frontend/tests/e2e/apply/fill-sflll-form.spec.ts
+++ b/frontend/tests/e2e/apply/fill-sflll-form.spec.ts
@@ -18,7 +18,7 @@ const OPPORTUNITY_URL = `/opportunity/${opportunityId}`;
 
 test.describe("fill SF-LLL Form", () => {
   // Skip non-Chrome browsers in staging
-  test.beforeEach((_, testInfo) => {
+  test.beforeEach(({ page: _ }, testInfo) => {
     if (targetEnv === "staging") {
       test.skip(
         testInfo.project.name !== "Chrome",
--- a/frontend/tests/e2e/apply/happy-path-application-submission.spec.ts
+++ b/frontend/tests/e2e/apply/happy-path-application-submission.spec.ts
@@ -22,7 +22,7 @@ const OPPORTUNITY_ID = "f7a1c2b3-4d5e-6789-8abc-1234567890ab"; // TEST-APPLY-ORG
 const OPPORTUNITY_URL = `/opportunity/${OPPORTUNITY_ID}`;
 
 // Skip non-Chrome browsers in staging
-test.beforeEach((_, testInfo) => {
+test.beforeEach(({ page: _ }, testInfo) => {
   if (targetEnv === "staging") {
     test.skip(
       testInfo.project.name !== "Chrome",
--- a/frontend/tests/e2e/login/login-with-login-gov.spec.ts
+++ b/frontend/tests/e2e/login/login-with-login-gov.spec.ts
@@ -22,7 +22,7 @@ const TIMEOUT_REDIRECT = 90000;
 // Tagging the test for config separation
 test.describe("Login.gov based authentication tests", () => {
   // Skip non-Chrome browsers in staging
-  test.beforeEach((_, testInfo) => {
+  test.beforeEach(({ page: _ }, testInfo) => {
     if (targetEnv === "staging") {
       test.skip(
         testInfo.project.name !== "Chrome",
--- a/frontend/tests/e2e/vision.spec.ts
+++ b/frontend/tests/e2e/vision.spec.ts
@@ -2,7 +2,7 @@ import { expect, test } from "@playwright/test";
 
 test.beforeEach(async ({ page }) => {
   // the waitUntil change here is to work around a temporary bug with some staging assets
-  await page.goto("/vision", { waitUntil: "domcontentloaded" });
+  await page.goto("/vision", { waitUntil: "load" });
 });
 
 test("has title", async ({ page }) => {


Bhavna-Ramachandran:
**Credit:** Tagging approach inspired by @doug-s-nava's PR #8960 , where tags are passed via the test options object instead of embedding them in the test name.

.github/workflows/e2e-create-report.yml
```diff
@@ -45,7 +45,8 @@ jobs:
           path: frontend/all-blob-reports
           # artifact-ids: ${{ inputs.artifact-ids }}
           run-id: ${{ inputs.run_id }}
-          pattern: blob-report-shard-*
+          # Merges both auth and no-auth test artifacts
+          pattern: blob-report*-shard-*
```
mdragon: Wouldn't this need to track the change above that names these off the report_name_prefix?

Bhavna-Ramachandran: Acknowledged. I have made the update.

mdragon (dismissed):
👍🏻

.github/actions/e2e/action.yml
```diff
@@ -110,8 +114,23 @@ runs:
         npm run test:e2e -- --grep "${{ inputs.playwright_tags }}"
       shell: bash
 
+    - name: Run e2e tests excluding ${{ inputs.playwright_tags_invert }} (Shard ${{ inputs.current_shard }}/${{ inputs.total_shards }})
```
doug-s-nava: with this in here there's a chance that if a user supplies both a positive list and a negative list, the same tests could run twice. 

I think the ability to run a negative list is not a super high priority and outside the scope of this change, so I'd like to save implementing this to a separate PR if that's ok

.github/workflows/e2e-staging.yml
```diff
@@ -20,43 +20,61 @@ on:
...
-
       - name: Run E2E tests
         uses: ./.github/actions/e2e
         with:
           version: ${{ inputs.version || github.ref }}
           target: ${{ env.defaulted_target }}
           total_shards: ${{ matrix.total_shards }}
           current_shard: ${{ matrix.shard }}
-          playwright_tags: ${{ inputs.playwright_tags }}
+          playwright_tags_invert: "@auth"
```
doug-s-nava: this isn't quite what I had in mind for this tag. The `@auth` tag in my conception is for tests of `login specific behavior, or behavior that is tied to specific permissions`, not just any functionality that requires the user to be logged in. It's not a big deal since we haven't really started implementing tags yet, but if we're using `@auth` to tag any test where a user logs in we may need to define a different tag to track tests of auth specific behavior

frontend/tests/e2e/utils/authenticate-e2e-user-utils.ts
doug-s-nava, line 94: I know this is out of scope but why do we want to always open the mobile nav after logging in?

frontend/tests/e2e/utils/create-application-utils.ts
```diff
@@ -5,6 +5,39 @@ import playwrightEnv from "tests/e2e/playwright-env";
...
 
+// Retry config for transient network errors (e.g. net::ERR_NETWORK_CHANGED)
+// which are common in Codespaces when navigating to staging URLs.
+const NAVIGATION_RETRIES = 3;
+const NAVIGATION_RETRY_DELAY_MS = 3000;
+
+/**
+ * Navigates to a URL with retry logic to handle transient network errors.
+ */
+async function gotoWithRetry(
```
doug-s-nava: this seems like a general util rather than something tied to "create-application". Can we move to a more general file?

frontend/tests/e2e/utils/perform-login-utils.ts
```diff
@@ -17,6 +17,28 @@ if (
 const TIMEOUT_HOME = playwrightEnv.targetEnv === "staging" ? 180000 : 60000;
 const TIMEOUT_MFA = 120000;
 
+// TOTP codes are valid for 30s windows. If we're within this many seconds
```
doug-s-nava, line 20: very cool!

frontend/tests/e2e/utils/perform-login-utils.ts
```diff
@@ -130,12 +156,25 @@ export const performStagingLogin = async (
...
   }
+
+  // If already logged in, return the sign-out button immediately.
+  // This handles the case where authenticateE2eUser is called on a page
+  // where a session is already active (e.g. storageState was injected).
+  const existingSignOut = page.locator(
+    'button:has-text("Sign out"), a:has-text("Sign out")',
+  );
+  if (await existingSignOut.isVisible({ timeout: 3000 }).catch(() => false)) {
+    // console.log("performStagingLogin: already logged in, skipping login flow");
```
doug-s-nava: can remove this

frontend/tests/e2e/saved-opportunities.spec.ts
```diff
@@ -1,15 +1,14 @@
...
-import { performStagingLogin } from "./utils/perform-login-utils";
+import { waitForURLChange } from "./playwrightUtils";
+import { authenticateE2eUser } from "./utils/authenticate-e2e-user-utils";
 
 const { targetEnv } = playwrightEnv;
 
-test.afterEach(async ({ context }) => {
-  await context.close();
-});
+// Note: do NOT close context in afterEach — Playwright manages context lifecycle
```
doug-s-nava, line 9: don't think we need this comment here - if this should be documented let's put it somewhere more general. Should we start a document for things like this either in /documentation or in Confluence?

.github/workflows/e2e-staging.yml
```diff
@@ -20,43 +20,61 @@ on:
...
+      - name: Checkout repository
+        uses: actions/checkout@v6
+      - name: Set target
+        run: if [[ -z "${{ inputs.target }}" ]]; then echo "defaulted_target=staging" >> "$GITHUB_ENV"; else echo "defaulted_target=${{ inputs.target }}" >> "$GITHUB_ENV"; fi
+      - name: Run E2E tests
+        uses: ./.github/actions/e2e
+        with:
+          version: ${{ inputs.version || github.ref }}
+          target: ${{ env.defaulted_target }}
+          total_shards: 1
```
doug-s-nava, line 39: I think this is still going to run all browsers, just in 1 shard. The configurations are defined in the playwright config so unless we're overriding that or supplying a new config this doesn't really solve the problem

doug-s-nava (commented):
What if we solved this by opting out of anything but chrome within the affected tests rather than updating the jobs?

.github/actions/e2e/action.yml
Bhavna-Ramachandran: Yes I will remove this from action and handle the exclusion directly in staging workflow. Will make this update now.

frontend/tests/e2e/utils/authenticate-e2e-user-utils.ts
Bhavna-Ramachandran: In some login tests, post login the flow is expected to open mobile nav (on mobile browser it is hidden behind hamburger menu) if needed, since this function is a shared login helper, we have it part of this. Based on need we can also move it part of each test too.

.github/workflows/e2e-staging.yml
Bhavna-Ramachandran: Based on our discussion. I am removing the auth tag.

frontend/tests/e2e/utils/create-application-utils.ts
Bhavna-Ramachandran: Acknowledged. I have moved this to lifecycle-utils

frontend/tests/e2e/utils/perform-login-utils.ts
Bhavna-Ramachandran: acknowledged.

frontend/tests/e2e/saved-opportunities.spec.ts
Bhavna-Ramachandran: Yes it is useful to document this. I saw failures in this test in Staging and I made this note. If its ok I would suggest to keep this in this spec too, it would serve as reminder when we make updates or debug this flow. Let me know your thoughts.

.github/workflows/e2e-staging.yml
Bhavna-Ramachandran: Yes, you are right. However, the browser execution is controlled in the test command we pass, where we had explicitly specified the Chrome project:
npx playwright test --config ./tests/playwright.config.ts --project=Chrome --workers=1 --reporter=list --grep @auth
So the auth workflow ran only against the Chrome project with a single worker, which avoided the parallel session in staging.

frontend/tests/e2e/utils/authenticate-e2e-user-utils.ts
```diff
@@ -32,12 +38,61 @@ export async function authenticateE2eUser(
       await page.waitForTimeout(2000);
     }
   } else if (targetEnv === "staging") {
+    const isChrome = !!test.info().project.name.match(/^Chrome/);
```
doug-s-nava: for now I'd hope that this function would not be called in staging except for Chrome. Rather than the early return, which would allow the test to continue, I'd rather either:

* throw an error
* assume that the opt out behavior is being enforced upstream of this function and ignore it

Issue #8760, opened 2026-03-03
Modify workflow event API to validate required metadata fields are present for approvals


chouinar:
### Summary

When we process a workflow approval event, we need the metadata to contain `approval_response_type` in the request, and that it is one of the expected types.

* When we process a `process_workflow` event - check if it is an approval event (That is, the event is in the approval config)
* If it is an approval event then the metadata needs to exist and contain `approval_response_type`
* That approval response type must be one of the allowed values according to the enum

A decent amount of this logic already exists in https://github.com/HHS/simpler-grants-gov/blob/main/api/src/workflow/service/approval_service.py#L51 - consider rearranging it a bit to be reusable in the API which doesn't have a state machine event - perhaps by passing in the metadata

### Acceptance criteria

- [ ] Validation added to API that verifies approval events have the required metadata to process them
- [ ] Tests updated to continue working and validate the success/failure cases (failures should include both missing and invalid value)

PR #8966, opened 2026-03-10
[Issue #8760] Modify workflow event API to validate required metadata fields are present for approvals

kkrug:
## Summary

Fixes for #8760  

## Changes proposed

- Updated ingest_workflow_event.py to validate required metadata for an approval event and add `InvalidWorkflowResponseTypeError`
- Updated approval_service.py to modify `get_approval_response_type` to call a new function, `get_approval_response_type_from_metadata`, which gets the approval response type from the metadata rather than from the state machine event
- Updated test_workflow_routes.py to test new functionality and add metadata to existing tests
- Updated test_approval_processor.py to fix a test that needed a different match value

## Context for reviewers

When we process a workflow approval event, we need the metadata to contain `approval_response_type` in the request, and that it is one of the expected types.

- When we process a `process_workflow event` - check if it is an approval event (That is, the event is in the approval config)
- If it is an approval event then the metadata needs to exist and contain `approval_response_type`
- That approval response type must be one of the allowed values according to the enum

## Validation steps

Confirm tests pass

### DIFF
--- a/api/src/workflow/service/approval_service.py
+++ b/api/src/workflow/service/approval_service.py
@@ -48,14 +48,18 @@ def get_approvals_for_workflow(
     return list(approvals)
 
 
-def get_approval_response_type_from_metadata(metadata: dict | None) -> ApprovalResponseType:
+def get_approval_response_type_from_metadata(
+    metadata: dict | None, log_extra: dict | None = None
+) -> ApprovalResponseType:
     """Get the approval response type from a metadata dict."""
+    if log_extra is None:
+        log_extra = {}
     raw_value = None
     if metadata is not None:
         raw_value = metadata.get(WorkflowConstants.APPROVAL_RESPONSE_TYPE)
 
     if raw_value is None:
-        logger.warning("Approval response type not found in metadata")
+        logger.warning("Approval response type not found in metadata", extra=log_extra)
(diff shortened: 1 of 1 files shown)


api/src/workflow/service/approval_service.py
```diff
@@ -48,28 +48,28 @@ def get_approvals_for_workflow(
     return list(approvals)
 
 
-def get_approval_response_type(state_machine_event: StateMachineEvent) -> ApprovalResponseType:
-    """Get the approval response type from the state machine event."""
-    raw_value = state_machine_event.get_metadata_value(WorkflowConstants.APPROVAL_RESPONSE_TYPE)
+def get_approval_response_type_from_metadata(metadata: dict | None) -> ApprovalResponseType:
```
chouinar: Just to keep the log_extra we had before, sometimes I like to have methods take in an optional context of the logging info. Then when we call it in the new function below, pass in `state_machine_event.get_log_extra()`.

```suggestion
def get_approval_response_type_from_metadata(metadata: dict | None, log_extra: dict | None = None) -> ApprovalResponseType:
    if log_extra is None:
        log_extra = {}
```

kkrug: Added the logging back in

PR #8967, opened 2026-03-10
[Issue #6788] Footer column spacing bug

### LINKED ISSUE #6788: footer text squished at small tablet widths
### Summary

Image

### Reproducibility

None

### How to reproduce the bug

- resize window to just before mobile breakpoint
- notice footer text overlapping

### Browser

None

### Code of Conduct

andycochran:
## Summary

Fixes #6788 

## Changes proposed

- Adjust the footer columns/spacing to account for variable content widths
- Keeps the logo above the other columns for mobile and desktop
- On tablet, makes the "Explore" column (which has narrow content) size `auto` while the other two are `fill`
- On larger viewports, spaces out the columns appropriately
- Adds responsive margin to footer headings so each group is spaced better

### Mobile (left=local / right=prod)

- Everything the same 

image

### Tablet (left=local / right=prod)

- Logo on its own row
- Left column narrower (`auto` while others are `fill`) 

image

### Tablet Large (left=local / right=prod)

- Columns are a bit wider, spaced nicely

image

### Desktop (left=local / right=prod)

- Logo in same row as other columns 

image

 
## Context for reviewers

The "Simpler.Grants.gov" text was wider than the column at a specific viewport width. This change makes the footer columns more dynamic. 

## Validation steps

- View any page and resize your viewport from mobile to desktop sizes
- Make sure to look at every possible size in between
- At every width, the footer columns should not overlap


frontend/src/components/Footer.tsx
```diff
@@ -63,9 +67,11 @@ const Footer = () => {
             </Grid>
             <Grid
               className="margin-top-3 tablet:margin-top-0"
-              tablet={{ col: 3 }}
+              tablet={{ col: "fill" }}
+              tabletLg={{ col: 5 }}
+              desktop={{ col: 3 }}
```
btabaska, line 73: Nit: At tabletLg, the three content columns share a row: Explore (auto) + Simpler (5) + Agency (5). The two fixed-width columns consume 10/12 columns, leaving little space for the auto-width Explore column. So when I use large text size accessibility I note that it can cause wrap, which isn't the end of the world. It may make sense to constrict the Explore column at tabletLg.

`<Grid
  className="usa-footer__nav border-0"
  tablet={{ col: "auto" }}
  tabletLg={{ col: 2 }}
  desktop={{ col: 2 }}
>`

btabaska:
Small nit, overall LGTM

frontend/src/components/Footer.tsx
andycochran: Yeah. Should prob specify `col: 2` at that size.

andycochran: Good find, @btabaska! That was also causing a misaligned left edge on the "Explore" column at `tabletLg` size. FIxed.

Issue #8605, opened 2026-02-19
Add emails to workflow approvals


chouinar:
### Summary

We want to make it so when a workflow enters an approval state, we send an email to every user that could approve the workflow.

In #8399 we will have setup a utility for doing that email, in this ticket we want to call that utility whenever a workflow enters an approval state.

We need to be careful about this, we don't want to send too many emails out.

Let's assume we have a simple workflow of `start` -> `pending_example_approval` -> `end` (we can ignore any branching cases). When a workflow first enters the `pending_example_approval` state, we want to send an email, but if it goes `pending_example_approval` -> `pending_example_approval` (which it will every time we add an approval to check the count), we do not want to keep sending emails.

A few options:
* Figure out if there is a way to add a global handler for this so we don't need to configure it. We could do something like define an `enter` handler and if the prior state != the current state and the current state is one of the configured ones for approvals, send an email.
  * It might that we want to use a [listener](https://python-statemachine.readthedocs.io/en/latest/listeners.html) for this 
* Otherwise, we'll want to explicitly configure this as an `after` function on any state transitions that go into the approval pending state. 

### Acceptance criteria

- [ ] Setup a pattern for adding email notifications to state transitions INTO an approval state
- [ ] Add tests that verify this behavior, should be able to add to existing approval related tests

PR #8990, opened 2026-03-11
[Issue #8399 / #8605] Add email processor for workflow management

### LINKED ISSUE #8399: Create the common logic for sending emails for workflow management, including specific logic for approvals
### Summary

We already have a client for sending emails via AWS pinpoint, but want to make a wrapper around that that'll be fairly easy to reuse where needed in the workflow management work.

For sending an email, we need the following:
* Who we are sending it to (email which is attached to a user)
* Subject
* Message

Additionally, there are some configurational fields like an app ID and trace ID that we should handle as part of the utility we build.

What we should build is a utility like:

```py
def send_workflow_email(email: str, subject: str, message: str) -> None:
    pass
```

This should then be called by:
```py
def send_approval_email_for_grantor(privilege: Privilege, agency: Agency): # maybe more params as needed to build email contents
     pass
```

For a given privilege + workflow, determine who should receive the email. For now, only worry about grantors based on the agency passed in. This should send multiple emails if multiple users are found. In the future, we might add more specific filtering, but the first pass doesn't need to worry about that.

Construct an email with:
 Subject - `An approval is required for simpler grants workflow {workflow_id}` 
 Message - 
```
An approval is required for simpler grants workflow {workflow_id}.

This is required for agency {agency.agency_code}: {agency.agency_name}
```

Don't worry about the contents being super thorough yet, we'll have design look at this later / probably add some links.

### Acceptance criteria

- [ ]  Setup utility for sending emails to grantors for approvals

chouinar:
## Summary
Fixes #8399 + #8605 

## Changes proposed
* Added logic to workflows for sending emails when entering a workflow state that requires approvals

## Context for reviewers
We want to send approval emails only when first entering an approval state. We know this by checking:
* Is the state we're entering an approval state - this was determined by flipping the configuration for approvals to be state->approval config (calculated from the existing config)
* Is the state we're entering the same as the prior state - if so, we're already in that state, we don't want to send emails again

The email logic is mostly a lot of error checks, logging and guard rails. The actual email contents aren't 100% finalized, but are good enough for a prototype and testing.

## Validation steps
See tests - still not able to run service on its own. Soon though.

### DIFF
--- a/api/src/workflow/base_state_machine.py
+++ b/api/src/workflow/base_state_machine.py
@@ -2,7 +2,6 @@ import logging
 from typing import Any, cast
 
 from statemachine import StateChart
-from statemachine.event_data import EventData
 
 from src.constants.lookup_constants import ApprovalResponseType
 from src.db.models.workflow_models import Workflow
@@ -115,6 +114,3 @@ class BaseStateMachine(StateChart):
             get_approval_response_type(state_machine_event)
             == ApprovalResponseType.REQUIRES_MODIFICATION
         )
-
-    def on_enter(self, state_machine_event: StateMachineEvent, event_data: EventData) -> None:
-        print(event_data)
--- a/api/src/workflow/listener/workflow_approval_email_listener.py
+++ b/api/src/workflow/listener/workflow_approval_email_listener.py
@@ -54,7 +54,7 @@ class WorkflowApprovalEmailListener:
             )
             return
 
-        # If the state machine's state is changing as part of this
+        # If the state machine's state is NOT changing as part of this
         # then we don't want to do anything. We only want to send
         # emails when first entering the state.
         if event_data.source == target_state:
--- a/api/src/workflow/workflow_config.py
+++ b/api/src/workflow/workflow_config.py
@@ -35,4 +35,9 @@ class WorkflowConfig:
         self.state_approval_mapping = {}
 
         for approval_config in self.approval_mapping.values():
+            if approval_config.approval_state in self.state_approval_mapping:
+                raise Exception(
+                    f"Approval state {approval_config.approval_state} is configured on two separate approvals - must be unique"
+                )
+
             self.state_approval_mapping[approval_config.approval_state] = approval_config
--- a/api/tests/workflow/test_workflow_config.py
+++ b/api/tests/workflow/test_workflow_config.py
@@ -1,0 +1,30 @@
+import pytest
+
+from src.constants.lookup_constants import ApprovalType, Privilege, WorkflowEntityType, WorkflowType
+from src.workflow.state_persistence.opportunity_persistence_model import OpportunityPersistenceModel
+from src.workflow.workflow_config import ApprovalConfig, WorkflowConfig
+from tests.workflow.state_machine.test_state_machines import BasicState
+
+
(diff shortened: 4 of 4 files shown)


api/src/workflow/base_state_machine.py
```diff
@@ -114,3 +115,6 @@ def is_approval_event_requires_modification(
             get_approval_response_type(state_machine_event)
             == ApprovalResponseType.REQUIRES_MODIFICATION
         )
+
+    def on_enter(self, state_machine_event: StateMachineEvent, event_data: EventData) -> None:
+        print(event_data)
```
kkrug: Was this supposed to do anything else or is it a stub for future use?

api/src/workflow/util/workflow_util.py
```diff
@@ -0,0 +1,51 @@
...
+        send_pinpoint_email_raw(
+            to_address=user.email,
+            subject=subject,
+            message=message,
+            app_id=config.app_id,
+            trace_id=trace_id,
+        )
+
+    except Exception:
+        logger.exception("Failed to send email for workflow", extra=log_extra)
```
kkrug, line 51: I know we're not fully up and running yet, but are we ok with this failing silently?

api/src/workflow/listener/workflow_approval_email_listener.py
```diff
@@ -0,0 +1,104 @@
...
+        approval_config = state_machine_event.config.state_approval_mapping.get(target_state, None)
+        if approval_config is None:
+            logger.debug(
+                "State does not have approval, no email notification required", extra=log_extra
+            )
+            return
+
+        # If the state machine's state is changing as part of this
+        # then we don't want to do anything. We only want to send
+        # emails when first entering the state.
```
kkrug:
```suggestion
        # If the state machine's state is NOT changing as part of this
        # then we don't want to do anything. We only want to send
        # emails when first entering the state.
```

api/src/workflow/workflow_config.py
```diff
@@ -22,3 +23,16 @@ class WorkflowConfig:
...
+    # and is automatically calculated in the post_init below.
+    state_approval_mapping: dict[str, ApprovalConfig] = dataclasses.field(
+        init=False, default_factory=dict
+    )
+
+    def __post_init__(self) -> None:
+        self.state_approval_mapping = {}
+
+        for approval_config in self.approval_mapping.values():
+            self.state_approval_mapping[approval_config.approval_state] = approval_config
```
kkrug, line 43: Can two different events map to the same `approval_state` (or will it be possible in the future)?

api/src/workflow/base_state_machine.py
chouinar: Whoops - was from testing behavior - will clean up

api/src/workflow/util/workflow_util.py
chouinar: I went back and forth on this a bit - should failing to send an email fail a workflow / cause whatever event was sent to get ignored? I wasn't sure, but erred on the side of not letting emails break a workflow.

api/src/workflow/workflow_config.py
chouinar: They shouldn't - I will add a check to this logic that makes it so that can't happen.

Issue #8596, opened 2026-02-19
"Create Opportunity" Page


glenblosser-nava:
### Summary

When the user clicks Create Opportunity on the Opportunity List page, they are taken to this page. This contains only the basic fields required for the opportunity to be created:
Agency ID (inherited from the Opportunity List page), Title, Category, Opportunity Number
Once all are provided, the Submit button enables, which calls the Opportunity-Create endpoint. The user will then be redirected to the Edit Opportunity page for the new opportunity's ID.

### Acceptance criteria

- [ ] The user can enter data in each required field
- [ ] Submit button is only enabled when all fields are filled in
- [ ] Submit calls the create opportunity endpoint and redirects the user
- [ ] Agency ID is pulled from the list opportunities page, but can be changed via dropdown by user

brianlee-nava:
@crystabelrangel we updated and added one AC up top for where this should live "opportunities/create"

brianlee-nava:
@chrisw-nava CC: @brianlee-nava suggests creating a table....check the backlog refinement meeting recording for context. Chris will share an example. Brian and @glenblosser-nava will update the team async on guidance so this is ready for Sprint 5.5

brianlee-nava:
Update: @doug-s-nava has volunteered to take a pass at adjusting this ticket and draw up an example of what this sort of ticket should look like

brianlee-nava:
@doug-s-nava CC: @glenblosser-nava We went over this as a team and posted the answers to the questions above.

doug-s-nava:
### Open questions from 3/3

* what happens in the UI on submit? We don't have anywhere to redirect the user to yet, and likely don't want to build out a "saved" state that will not be used later
  * for now, it takes you opportunity list page.
* mention was made of being able to change / select agency on page but this is not in the designs. Will that come later? How will the page know which agency to attach to the opportunity? Should we throw agency id into the path or query param?
  * This should be added as a dropdown to the design. When click on "create opportunity" button, it should pass currently selected agency ID as part of query string. You can manually change that on "Create Opportunity" Page as needed by choosing on the dropdown.
* what does the cancel button do? redirect to opportunities list page?
  * redirects you back to opportunity list page. no updates need to be made
* where does translation of opportunity category to opportunity category id happen? Can we pass up the string, and let the backend assign the id, or should we hardcode the id values into the form options? 
[...]

PR #8991, opened 2026-03-11
[Issue #8596] Create Opportunity page layout and components

Dao-REI:
## Summary

Work for #8596 

## Changes proposed

1. Created the Create Opportunity page (step 1) for Grantors to start the "create opportunity listing" process
2. Created Common/Shared Form Fields for consistent look-and-feel and reusability, for Grantors.

## Context for reviewers

- 11 files added
- 6 files updated
- 18 ACs
- This page only creates the opportunity record in the database. The opportunity summary will be a separate page (step 2).
- The page is using server-side action to call the backend API. 
- Two additional tickets will be created for common/shared components: Trussworks CharacterCount component and DynamicFieldLabel.

## Validation steps

Acceptance Criteria:
- [ ] Verify that the page matches the Create Opportunity page mockup on Figma, https://www.figma.com/design/o5g0d2W5RKHYHjdvVCqTV2/Grantor-Designs-%7C-Quad-5-%7C-Simpler-Grants-2026?node-id=4707-11806&p=f&m=dev
- [ ] Test/check the ACs on the task #8596

### DIFF
--- a/frontend/src/components/opportunities/create/CreateOpportunityForm.test.tsx
+++ b/frontend/src/components/opportunities/create/CreateOpportunityForm.test.tsx
@@ -144,7 +144,7 @@ describe("createOpportunityForm field change events", () => {
 
     // 2. Fill in Opportunity Number
     const textboxOppNbr = screen.getByRole("textbox", {
-      name: "opportunityNumber *",
+      name: "CreateOpportunityForm.opportunityNumber *",
     });
     expect(textboxOppNbr).toBeInTheDocument();
     expect(textboxOppNbr).toHaveValue("");
@@ -155,7 +155,7 @@ describe("createOpportunityForm field change events", () => {
 
     // 3. Fill in Opportunity Title
     const textareaOppTitle = screen.getByRole("textbox", {
-      name: "opportunityTitle *",
+      name: "CreateOpportunityForm.opportunityTitle *",
     });
     expect(textareaOppTitle).toBeInTheDocument();
     expect(textareaOppTitle).toHaveValue("");
@@ -168,7 +168,7 @@ describe("createOpportunityForm field change events", () => {
 
     // 4. Select a Category that is not "other"
     const selectCategory = screen.getByRole("combobox", {
-      name: "category *",
+      name: "CreateOpportunityForm.category *",
     });
     expect(selectCategory).toBeInTheDocument();
     expect(selectCategory).toHaveValue("");
@@ -178,7 +178,7 @@ describe("createOpportunityForm field change events", () => {
     expect(saveButton).toBeEnabled();
     // The Explanation field should be hidden
     const testExplain = screen.queryByRole("textbox", {
-      name: "categoryExplanation *",
+      name: "CreateOpportunityForm.categoryExplanation *",
     });
     expect(testExplain).not.toBeInTheDocument();
 
@@ -189,7 +189,7 @@ describe("createOpportunityForm field change events", () => {
     expect(saveButton).toBeDisabled();
     // The Explanation field should now be displayed
     const textareaExplain = screen.getByRole("textbox", {
-      name: "categoryExplanation *",
+      name: "CreateOpportunityForm.categoryExplanation *",
     });
     expect(textareaExplain).toBeInTheDocument();
     expect(textareaExplain).toHaveValue("");


frontend/src/app/[locale]/(base)/opportunities/create/[agencyId]/page.tsx
```diff
@@ -0,0 +1,116 @@
...
+      throw error;
+    }
+    return <PageErrorMessage />;
+  }
+
+  return (
+    <>
+      <GridContainer>
+        <PageHeader/>
+        <CreateOpportunityPage1
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunities/create/[agencyId]/page.tsx]: Should avoid arbitrary variable naming - what about `CreateOpportunityForm`?

frontend/src/components/opportunities/create/CreateOpportunityPage1.tsx
```diff
@@ -0,0 +1,244 @@
...
+const categoryList: KeyValuePair[] = [
+  { key: '1', value: 'Discretionary' },
+  { key: '2', value: 'Mandatory' },
+  { key: '3', value: 'Continuation' },
+  { key: '4', value: 'Earmark' },
+  { key: '5', value: 'Other' },
+];
+
+// Field level error messages
+let oppNbrVldtnError = "";
```
doug-s-nava: we should track things like this, if necessary, in component state rather than in module globals. React will handle this better than the module will (as the file may be re-loaded at weird times, and state may be reset as result)

frontend/src/components/opportunities/create/CreateOpportunityPage1.tsx
```diff
@@ -0,0 +1,244 @@
...
+  userAgencies: RelevantAgencyRecord[];
+}
+export const CreateOpportunityPage1 = ({
+  agencyId,
+  userAgencies,
+}: CreateOpportunityPage1Props) => {
+
+  // All labels/text on this page with support for international languages
+  const t = useTranslations("CreateOpportunity.CreateOpportunityPage1");
+  const oppNbrLabel = t("opportunityNumber");
```
doug-s-nava: Since defining variables like this leads to some bloat in the file, I'd rather follow the usual pattern and call `t` directly where these values are being used

frontend/src/components/opportunities/create/CreateOpportunityPage1.tsx
```diff
@@ -0,0 +1,244 @@
...
+  const agencyDesc = ""
+  const categoryLabel = t("category");
+  const categoryDesc = t("categoryDesc");
+  const explainLabel = t("categoryExplanation");
+  const explainDesc = t("categoryExplanationDesc");
+  const charsAllowed40 = t("charactersAllowed40");
+  const charsAllowed255 = t("charactersAllowed255");
+
+  // Agencies: sort alphabetically, convert to key-value pairs, set the default
+  const sortedAgencies = [...userAgencies].sort((a, b) =>
```
doug-s-nava: my guess is that the agency endpoint supports sorting, will be easier to deal with things coming in pre-sorted

frontend/src/components/opportunities/create/CreateOpportunityPage1.tsx
```diff
@@ -0,0 +1,244 @@
...
+    a.agency_name.localeCompare(b.agency_name)
+  );
+  const keyValueList: KeyValuePair[] = sortedAgencies.map(agency => ({
+    key: agency.agency_id,
+    value: agency.agency_name,
+  }));
+  keyValueList.forEach((item) => {
```
doug-s-nava: as it is this will run on each render - can you wrap this and the keyValueList mapping in a useEffect?

frontend/src/components/opportunities/create/CreateOpportunityPage1.tsx
```diff
@@ -0,0 +1,244 @@
...
+    selectedAgencyId = e.target.value;
+    checkRequiredFields();
+  };
+  const onExplanationChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    categoryExplanation = e.target.value;
+    checkRequiredFields();
+  };
+
+  // Custom CSS for the cancel button
+  const cancelButtonStyle: React.CSSProperties = {
```
doug-s-nava: unless there's a really compelling reason we should not use inline styling

frontend/src/components/opportunities/create/CreateOpportunityPage1.tsx
```diff
@@ -0,0 +1,244 @@
...
+    </>    
+  )
+}
+
+
+//----- Handle Save and Cancel -----
+const handleCancel = () => {
+  redirect("/opportunities")
+};
+const handleSave = () => {
```
doug-s-nava: let's use a server action for this - that may change the general approach here a little

frontend/src/constants/defaultFeatureFlags.ts
```diff
@@ -3,5 +3,6 @@ export type FeatureFlags = { [name: string]: boolean };
 export const defaultFeatureFlags: FeatureFlags = {
   applyFormPrototypeOff: false,
   opportunitiesListOff: false,
+  createOpportunityOff: false,
```
doug-s-nava: if this PR is creating a new feature flag, it should also update the terraform, see https://github.com/HHS/simpler-grants-gov/blob/main/documentation/frontend/featureFlags.md

frontend/src/components/grantor/CommonFormFields.tsx
```diff
@@ -0,0 +1,236 @@
+// These fields are meant to be reusable and provide a consistent look and feel for all pages.
```
doug-s-nava, line 1: I like the idea here a lot, but I think we should talk about this as a group before jumping in. It probably makes the most sense to use custom / barebones implementation rather than shared  / general components at this point,

frontend/src/components/opportunities/create/CreateOpportunityPage1.tsx
Dao-REI: @doug-s-nava useState() is not working the way I need it to. The value is not set right away but after all code execution is done. I need the value set right away in order to check values for all required fields before enabling the Save button. 
The code:
`   const [category, setCategory] = useState<string>("");
    const onCategorySelection = (e: React.ChangeEvent<HTMLSelectElement>) => {
    console.log("DEBUG: e.target.value is ", e.target.value);
    setCategory(e.target.value);
    console.log("DEBUG: category is ", category);
`
When I select Continuation, the field category is still the old value.  
Console: 
`DEBUG: e.target.value is  Continuation
DEBUG: category is  Mandatory
`
I think I have to use "let" variables for my situation.

frontend/src/components/opportunities/create/CreateOpportunityPage1.tsx
Dao-REI: The v1/users/[user_id}/agencies endpoint is currently not sorting the agencies.
`def get_user_agencies(user: User) -> list[Agency]:
    """Get agencies for a user using the SQLAlchemy relationship."""
    return [agency_user.agency for agency_user in user.agency_users]`

Do you want me to add sorting, or ask someone to add it?

frontend/src/components/opportunities/create/CreateOpportunityPage1.tsx
Dao-REI: I moved this code from the child (form) to the parent (page). I think that's only getting called once on form-load.

frontend/src/constants/defaultFeatureFlags.ts
Dao-REI: Per @glenblosser-nava we should use one feature flag for the entire Opportunity Publishing epic. Thus, I've updated the code to reuse Glen's opportunitiesListOff flag.

frontend/src/app/[locale]/(base)/opportunities/create/[agencyId]/actions.tsx
```diff
@@ -0,0 +1,80 @@
+"use server";
+
+import { getSession } from "src/services/auth/session";
+import { handleCreateOpportunity } from "src/services/fetch/fetchers/createOpportunityFetcher";
+import { CreateOpportunityResponse } from "src/types/grantor/createOpportunityTypes";
+import { z } from "zod";
+
+const validateFormFields = async (formData: FormData) => {
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunities/create/[agencyId]/actions.tsx]: if frontend validations are out of scope we can remove this function

frontend/src/app/[locale]/(base)/opportunities/create/[agencyId]/page.tsx
```diff
@@ -0,0 +1,133 @@
...
+          {t("basicInstructions")}
+        </div>
+      </div>
+    </>
+  );
+};
+
+
+// --- Main Page ---
+interface formPageProps {
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunities/create/[agencyId]/page.tsx]: a couple of conventions to follow here:
* type names should be capitalized
* types should be preferred to interfaces unless inheritance is in play https://github.com/HHS/simpler-grants-gov/blob/main/documentation/frontend/code-style.md#interfaces-vs-types

frontend/src/app/[locale]/(base)/opportunities/create/[agencyId]/page.tsx
```diff
@@ -0,0 +1,133 @@
...
+    if (error instanceof UnauthorizedError) {
+      throw error;
+    }
+    return <PageErrorMessage />;
+  }
+  // Agencies: sort alphabetically, convert to key-value pairs, set the default
+  const sortedAgencies = [...userAgencies].sort((a, b) =>
+    a.agency_name.localeCompare(b.agency_name)
+  );
+  const mapAgencyToKeyValuePair = (agency: RelevantAgencyRecord): KeyValuePair => ({
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunities/create/[agencyId]/page.tsx]: this can be split out from the component render

frontend/src/app/[locale]/(base)/opportunities/create/[agencyId]/page.tsx
```diff
@@ -0,0 +1,133 @@
...
+    key: agency.agency_id.toString(), // Example mapping
+    value: agency.agency_name,
+  });
+  const keyValueList: KeyValuePair[] = sortedAgencies.map(mapAgencyToKeyValuePair);
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunities/create/[agencyId]/page.tsx]: this is a pretty vague variable name - how about `SortedUserAgencyRecords`?

frontend/src/app/[locale]/(base)/opportunities/create/[agencyId]/page.tsx
```diff
@@ -0,0 +1,133 @@
...
+  // const keyValueList: KeyValuePair[] = sortedAgencies.map(agency => ({
+  //   key: agency.agency_id,
+  //   value: agency.agency_name,
+  // }));
+  keyValueList.forEach((item) => {
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunities/create/[agencyId]/page.tsx]: this will be more performant and remove the need for tracking a mutable variable

```suggestion
  const defaultAgencyId = userAgencies.find((agency) => agency.agency_id === agencyId)
```

frontend/src/app/[locale]/(base)/opportunities/create/[agencyId]/page.tsx
```diff
@@ -0,0 +1,133 @@
...
+    </>
+  );
+};
+
+
+// --- Main Page ---
+interface formPageProps {
+  params: Promise<{ agencyId: string; locale: string }>;
+}
+async function FormPage({ params }: formPageProps) {
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunities/create/[agencyId]/page.tsx]:
```suggestion
async function CreateOpportunityPage({ params }: formPageProps) {
```

frontend/src/components/opportunities/create/CreateOpportunityForm.tsx
```diff
@@ -0,0 +1,256 @@
...
+
+// ----- Main Form -----
+export function CreateOpportunityForm({
+  defaultAgencyId,
+  userAgencies,
+}: {
+  defaultAgencyId: string;
+  userAgencies: KeyValuePair[];
+}) {
+  const t = useTranslations("CreateOpportunity.CreateOpportunityForm");
```
doug-s-nava: rather than using multiple `t` functions, call useTranslations with the lowest level common namespace for the component. Not sure about common labels - I like the idea but the pattern we use now is to not worry about duplicating translation messages. Feel free to duplicate any strings you need into the "CreateOpportunity" namespace

frontend/src/components/opportunities/create/CreateOpportunityForm.tsx
```diff
@@ -0,0 +1,256 @@
...
+  const router = useRouter();
+  useEffect(() => {
+    window.scrollTo({
+      top: 0,
+      behavior: "smooth", // Smooth scrolling animation
+    });
+    if (response?.success) {
+      router.push("/opportunities");
+    } else {
+      setDisableSave(true); // on text change does not seem to reset this?
```
doug-s-nava: not sure what your intention is here, but if you want to set this state based on interaction with a field you'll need to do that separately

frontend/src/components/opportunities/create/CreateOpportunityForm.tsx
```diff
@@ -0,0 +1,256 @@
...
+            labelId="label-for-opportunityTitle"
+            labelText={t("opportunityTitle")}
+            description={t("opportunityTitleDesc")}
+            fieldId="opportunityTitle"
+            fieldMaxLength={255}
+            isRequired={true}
+            onTextChange={onOppTitleChange}
+            defaultValue={response?.data?.opportunity_title || ""}
+          />
+          <CommonText textContent={t("charactersAllowed255")} />
```
doug-s-nava: use the truss CharacterCount instead of implementing field length tracking elsewhere https://github.com/trussworks/react-uswds/blob/main/src/components/forms/CharacterCount/CharacterCount.tsx

frontend/src/components/grantor/CommonFormFields.tsx
```diff
@@ -0,0 +1,233 @@
...
+        {textContent}
+      </div>
+    </>
+  );
+};
+
+// ----------------------------------------------------------
+// Common Label with ErrorMessage
+// ----------------------------------------------------------
+export const CommonLabel = ({
```
doug-s-nava, line 31: can we use DymamicFieldLabel? https://github.com/HHS/simpler-grants-gov/blob/main/frontend/src/components/applyForm/widgets/DynamicFieldLabel.tsx

frontend/src/components/opportunities/create/CreateOpportunityFormFields.tsx
```diff
@@ -0,0 +1,236 @@
...
+        />
+      </FormGroup>
+    </>
+  );
+};
+
+// ----------------------------------------------------------
+// Common Select input with error block
+// ----------------------------------------------------------
+export interface KeyValuePair {
```
doug-s-nava: I'm a little bit confused about why we need this. Couldn't you get the same functionality out of a regular string object: `{ [key:string]: string }`?

frontend/src/services/fetch/fetchers/agenciesFetcher.test.ts
```diff
@@ -1,24 +1,37 @@
...
   .mockReturnValue(["forecasted"]);
 
+jest.mock("src/services/auth/session", () => ({
+  getSession: (arg: unknown): unknown => mockUserSession(arg),
+}));
+
 jest.mock("src/services/fetch/fetchers/fetchers", () => ({
   fetchAgencies: (arg: unknown): unknown => mockFetchAgencies(arg),
   searchAgencies: (arg: unknown): unknown => mockSearchAgencies(arg),
+  fetchUserWithMethod: jest.fn(),   // Initialize as a mock -- this works
```
doug-s-nava: this works unless you want to assert anything about how fetchUserWithMethod is called

frontend/src/services/fetch/fetchers/createOpportunityFetcher.ts
```diff
@@ -0,0 +1,27 @@
+"server-only";
+
+import { fetchGrantorWithMethod } from "src/services/fetch/fetchers/fetchers";
+import { CreateOpportunityRecord } from "src/types/grantor/createOpportunityTypes";
+
+export const handleCreateOpportunity = async (
+  type: "POST", // "POST" | "PUT" | "DELETE"
```
doug-s-nava: we can create different methods for put (update) and delete when we need them

frontend/src/components/opportunities/create/CreateOpportunityForm.tsx
Dao-REI: That text was in the mockup. The component you suggest would be super nice, but I need approval from the Design team (Crystabel) and it will extend this ticket.

frontend/src/components/grantor/CommonFormFields.tsx
Dao-REI: Perhaps yes, but 
1. I will have a spend a little time to put it in and test it out.  And this ticket is already overdue and blocking other pages.
2. It is under .../applyForm/widgets/budget. If Apply decides to make any changes to this component, it will affect us as well.

frontend/src/services/fetch/fetchers/agenciesFetcher.test.ts
Dao-REI: This was the only way I could get the two tests I added to work. It was the third approach. I'm open to trying other things, but maybe later.

frontend/src/components/opportunities/create/CreateOpportunityPage1.tsx
Dao-REI: I figured it out. I'm using useState() and useEffect(), and now it's working properly.

frontend/src/components/grantor/CommonFormFields.tsx
Dao-REI: For DynamicFieldLabel, the styles are different and it is missing the error message handling. We cannot use it right now / as is. 
image

frontend/src/components/opportunities/create/CreateOpportunityForm.tsx
Dao-REI: I don't think we can use the CharacterCount component as-is. I see at least two issues. (1) We need to stop the user's input when the max length has been reached. (2) We need to override the font of the bottom character count text to be consistent with our app. We'll need time to do some customization. Do we want to create a separate ticket for this?
image

frontend/src/app/[locale]/(base)/opportunities/create/[agencyId]/page.tsx
```diff
@@ -0,0 +1,116 @@
...
+    if (error instanceof UnauthorizedError) {
+      throw error;
+    }
+    return <PageErrorMessage />;
+  }
+  // set the default agency if it's valid
+  const defaultAgency = userAgencies.find((agency) => agency.agency_id.toString() === agencyId);
+  const defaultAgencyId = defaultAgency?.agency_id.toString() ?? "";
+  // convert to key-value list for the combobox
+  type AgencyMap = { [key: string]: string };
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunities/create/[agencyId]/page.tsx]: * don't define a type within a function
* you can use `Record` for this https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type

frontend/src/components/opportunities/create/CreateOpportunityForm.tsx
doug-s-nava: bumping this

frontend/src/components/opportunities/create/CreateOpportunityForm.tsx
```diff
@@ -0,0 +1,225 @@
...
+      setShowExplain(true);
+    } else {
+      setShowExplain(false);
+    }
+    // Check for required fields to enable the Save button
+    const allReqFieldsFilled = 
+      opportunityNumber.trim() !== '' &&  
+      opportunityTitle.trim() !== '' && 
+      selectedAgencyId.trim() !== '' &&
+      ( (selectedCategoryId.trim() !== '' && selectedCategoryId.trim() !== 'other') ||
```
doug-s-nava: can simplify this slightly since you've already determined that `selectedCategoryId.trim() !== ''` on the line above

frontend/src/components/opportunities/create/CreateOpportunityForm.tsx
```diff
@@ -0,0 +1,225 @@
...
+    } else {
+      setDisableSave(true);
+      if (selectedCategoryId.trim() !== 'other') {
+        setExplain('');   // need to manually set this for checks below to work correctly
+      }
+    }
+  }, [response, router]);
+
+  // Use useEffect to check fields when inputs change
+  useEffect(() => {
```
doug-s-nava: we should have a test around this behavior

frontend/src/i18n/messages/en/index.ts
```diff
@@ -1929,4 +1908,36 @@ export const messages = {
...
+        'If "Other" was selected, please describe the specific process used to evaluate and award these funds.',
+      charactersAllowed40: "40 characters allowed",
+      charactersAllowed255: "255 characters allowed",
+      successMessage: "Opportunity started. Continuing shortly...",
+    },
+  },
+  CommonLabels: {
+    cancel: "Cancel",
+    saveAndContinue: "Save and continue",
+    pending: "Pending..",
```
doug-s-nava: * imagine this should have three dots to be a full ellipsis
* Are we sure about "pending"? Previous implementation have all used "loading, I think

Dao-REI: Added a dot. I followed another example (I forget which one) that was using Pending. I think "Loading" is appropriate for getting data, but "Pending" is better for saving data.

frontend/src/components/opportunities/create/CreateOpportunityForm.tsx
Dao-REI: It's selectedAgencyId that is check in the line above.

Issue #8957, opened 2026-03-10
Create tables/models for tracking retrieval of submissions


mdragon:
### Summary

Create a tables related to Agency pick up of Submissions

In the legacy system the SOAP API only allows for retrieving a submitted application once across all API consumers as the status of the record is changed to make it "Received by Agency."

We want to create a table named `application_submission_retrieved` with the following columns:
* `application_submission_retrieved_id` - UUID, primary key
* `application_submission_id` - UUID, not null, FK to application_submission
*  `created_by_user_id` - UUID, not null, FK to user
*  `modified_by_user_id` - UUID, not null, FK to user

We should make sure to create a factory for this new table.

### Acceptance criteria

- [ ] Table created
- [ ] Migration created
- [ ] Factory created and manually verified it works (future tickets will need it)

mdragon:
@kkrug can you coordinate with @jakobpederson on whether we need this table or if we should redirect this work to add legacy_is_received to the existing submission table instead.

PR #8993, opened 2026-03-11
[Issue #8957] Create tables/models for tracking retrieval of submissions

kkrug:
## Summary

Fixes for #8957  

## Changes proposed

- Updated competition_models.py to add `ApplicationSubmissionRetrieved` model
- Generated new db migration
- Updated factories.py to add `ApplicationSubmissionRetrievedFactory`

## Context for reviewers

In the legacy system the SOAP API only allows for retrieving a submitted application once across all API consumers as the status of the record is changed to make it "Received by Agency, so we created a table named `application_submission_retrieved` with the following columns:
- `application_submission_retrieved_id` - UUID, primary key
- `application_submission_id` - UUID, not null, FK to application_submission
- `created_by_user_id` - UUID, not null, FK to user
- `modified_by_user_id` - UUID, not null, FK to user

## Validation steps

Confirm that the migrations successfully run. The factory can be manually tested by running the following command in the `api` folder with the backend running locally:
`docker compose exec grants-api poetry run python -c "from src.adapters.db.clients.postgres_client import PostgresDBClient; from tests.src.db.models import factories; db_client = PostgresDBClient(); session = db_client.get_session(); db_session = session.__enter__(); factories._db_session = db_session; record = factories.ApplicationSubmissionRetrievedFactory.create(); print(f'ID: {record.application_submission_retrieved_id}'); print(f'submission_id: {record.application_submission_id}'); print(f'created_by: {record.created_by_user_id}'); print(f'modified_by: {record.modified_by_user_id}'); db_session.rollback(); db_session.close()"`

It should print out 4 UUIDs, something like this:
`ID: c139669a-5799-4852-a1d5-cba36498aea4`
`submission_id: 344b2753-cba7-45d5-ae80-99309007a6e1`
`created_by: a8f5e1ad-d4e9-4961-97b8-64f919f9c14f`
`modified_by: 4925f2a7-bdc1-4b55-8663-1f275a32b1e3`


api/src/db/models/application_submission_retrieved_models.py
```diff
@@ -0,0 +1,28 @@
+import uuid
+
+from sqlalchemy import ForeignKey
+from sqlalchemy.dialects.postgresql import UUID
+from sqlalchemy.orm import Mapped, mapped_column, relationship
+
+from src.db.models.base import ApiSchemaTable, TimestampMixin
+
+
+class ApplicationSubmissionRetrieved(ApiSchemaTable, TimestampMixin):
```
chouinar: Put this in the file with the application submission. There's a balance with the DB model files, a single file with everything would be too long, but a file-per-table will lead to an incredibly complex set of imports.

api/src/db/models/application_submission_retrieved_models.py
```diff
@@ -0,0 +1,28 @@
...
+    created_by_user_id: Mapped[uuid.UUID] = mapped_column(
+        UUID, ForeignKey("api.user.user_id"), nullable=False
+    )
+    modified_by_user_id: Mapped[uuid.UUID] = mapped_column(
+        UUID, ForeignKey("api.user.user_id"), nullable=False
+    )
+
+    application_submission = relationship("ApplicationSubmission")
+    created_by_user = relationship("User", foreign_keys=[created_by_user_id])
+    modified_by_user = relationship("User", foreign_keys=[modified_by_user_id])
```
chouinar: For foreign keys/relationships, use class types if possible (eg. `User` instead of `"User"`). Forward references (a field declared later in the file) are the only case we should be doing this.

---

Don't put `nullable` in the columns, just rely on the python types - don't define it twice - if they don't match, it won't complain but will behave badly. Better to do it one way than two.

---

Add the relationships on the other side - at least for application submission.

chouinar: @mdragon - I know you had put these, but I hate columns like these. If 3 people update, you don't keep that info. We don't need to solve it now, but really should have an audit table for this sort of thing in the future.

hao10282025-sudo: For the HRSA EHB system, the 2nd column is named as "LastUpdatedBy", which the naming is closer to its purpose, to log only the last user updated this record.
Are we targeting to have an audit table to log all the users updated for a record? It sounds like a history is more properly fit.
Could there be a way to auto-add these two user ids audit columns as the created_at and updated_at?

api/src/db/models/application_submission_retrieved_models.py
kkrug: Moved to competition_model

api/src/db/models/application_submission_retrieved_models.py
kkrug: Updated

chouinar: @hao10282025-sudo - we don't have history tables - it adds a lot of complexity to build and maintain those - the foreign keys especially. Grants.gov has them and it made getting data from them much more complex. I prefer the approach of tracking changes separately.

For example, we have a few `x_audit` tables that give a record of any time someone did an action related to a particular entity like an application.

PR #8994, opened 2026-03-11
[Issue #8925] Add application submission columns for award recommendation

### LINKED ISSUE #8925: Backend - Adjust the application submission table to have additional columns
### Summary

See [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Adjust-the-application-submission-table-to-have-additional-columns)

We want to adjust the application submission table to add:
- `application_submission_number` - nullable text
- `project_title` - nullable text
- `total_requested_amount` - nullable decimal

### Acceptance criteria

- [ ] Columns added
- [ ] Tests created/updated as necessary

chouinar:
## Summary
Fixes #8925

## Changes proposed
* Add 3 columns we'll need for award recommendation to the application submission table - will be populated in a later PR.

## Context for reviewers
https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints has context on how we'll populate these in the future.

## Validation steps
Factories are working. Can do `make console` and then use the updated application submission factory and it generates data uneventfully.
Screenshot 2026-03-11 at 3 05 18 PM


api/tests/src/db/models/factories.py
```diff
@@ -1662,6 +1662,14 @@ class Meta:
         lambda s: f"s3://local-mock-public-bucket/applications/{s.application_id}/submissions/{fake.uuid4()}/submission.zip"
     )
 
+    application_submission_number = factory.LazyAttribute(
+        lambda f: f"{f.application.competition.opportunity.opportunity_number}-{fake.unique.pystr_format(string_format="##???").upper()}"
+    )
+    project_title = factory.Faker("sentence")
```
nathan-stricker, line 1668: This might be a basic question, but is there a reason why `application_submission_number` and `project_title`  don't have the `sometimes_none()` wrapper and `total_requested_amount` does, despite them all being nullable? Is it less relevant for strings than decimals?

chouinar: I usually put the `sometimes_null` on fields that are often null. Application submission number should rarely be null, only being nullable because we already have data in other environments (We might want to backfill it eventually and make it non-null). Project title - eh it could be nullable, can adjust that later if it makes more sense for the frontend when testing.

Issue #8586, opened 2026-02-18
Create Idea of Smoke Tests


doug-s-nava:
### Summary

Depends on https://github.com/HHS/simpler-grants-gov/issues/7959

For identified scenarios TBD based on outcome of #7959, run smoke test grouped tests defined in outcome of #7959

### Acceptance criteria

- [ ] smoke tests run on identified cadence(s)

PR #9006, opened 2026-03-12
[Issue #8586] create the idea of smoke tests

doug-s-nava:
## Summary

Fixes for #8586 

## Changes proposed

Adds tags to the smoke tests according to this list https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2811854876/Initial+Test+Groupings+and+Cadences

## Validation steps

1. _VERIFY_: test CI runs run the expected smoke tests (see screenshot or https://github.com/HHS/simpler-grants-gov/actions/runs/23158076343/job/67278127324)
2. to test yourself, use the experimental test from the second to last commit in this branch

## Screenshots

Screenshot 2026-03-16 at 4 18 27 PM

### DIFF
--- a/frontend/tests/e2e/login/login-with-login-gov.spec.ts
+++ b/frontend/tests/e2e/login/login-with-login-gov.spec.ts
@@ -43,13 +43,7 @@ test.describe("Login.gov based authentication tests", () => {
     "Login.gov authentication with MFA",
     { tag: ["@smoke", "@auth"] },
     async ({ page, context }: { page: Page; context: BrowserContext }) => {
-      const isChrome = !!test.info().project.name.match(/^Chrome/);
       const isMobileProject = !!test.info().project.name.match(/[Mm]obile/);
-
-      // for now MFA is failing if we run this test too frequently, so limiting to one browser until we can figure that out
-      if (!isChrome) {
-        return;
-      }
 
       // Only start tracing manually on the first attempt. On retries, Playwright's
       // config (trace: "on-first-retry") has already started tracing, and calling


frontend/tests/e2e/apply/happy-path-application-submission.spec.ts
```diff
@@ -34,54 +34,58 @@ test.beforeEach(({ page: _ }, testInfo) => {
...
+      page,
+      "complete",
+      "SF-424B",
+      applicationUrl,
+    );
 
-  // Extra wait for page to fully render forms table after navigation
-  await page.waitForTimeout(10000);
+    // Extra wait for page to fully render forms table after navigation
+    await page.waitForTimeout(10000);
```
aoysimmons-GH, line 78:
```suggestion
    // Wait for the SF-LLL row to render before interacting with inclusion radios
    await page
      .locator("tr", {
        hasText: /Disclosure of Lobbying Activities \(SF-LLL\)/i,
      })
      .first()
      .waitFor({ state: "visible", timeout: 15000 });
```

frontend/tests/e2e/login/login-with-login-gov.spec.ts
```diff
@@ -39,154 +39,152 @@ test.describe("Login.gov based authentication tests", () => {
...
+        );
+        const isAlreadyLoggedIn = await signOutLocator
+          .isVisible({ timeout: 15000 })
+          .catch(() => false);
+
+        if (isAlreadyLoggedIn) {
+          //  console.log(
+          //    "login-with-login-gov: session already active, skipping MFA login",
+          //  );
+          await expect(signOutLocator).toHaveCount(1, {
```
aoysimmons-GH, line 85:
```suggestion
          await expect(signOutLocator.first()).toBeVisible({
```

Bhavna-Ramachandran: This (signOutLocator.first()).toBeVisible({ -> would not throw when the locator matches nothing + we are not just checking visibility. So we should stick with what it is currently.

frontend/tests/e2e/apply/happy-path-application-submission.spec.ts
Bhavna-Ramachandran: Below utils await selectFormInclusionOption(..) does the same + extras.

frontend/tests/e2e/apply/happy-path-application-submission.spec.ts
```diff
@@ -34,54 +34,71 @@ test.beforeEach(({ page: _ }, testInfo) => {
...
+    // Select 'No' for including SF-LLL form in submission
+    await selectFormInclusionOption(
+      page,
+      "Disclosure of Lobbying Activities (SF-LLL)",
+      "No",
+    );
+
+    // Submit the application and verify success
+    await submitApplicationAndVerify(page);
+    // Application ID is now available in appId variable for further use if needed
```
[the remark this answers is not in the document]
Bhavna-Ramachandran: This is just a reminder FYI to say we have it extracted as part of this utils (gets validated if its generated), so in future we don't have to work on creating a new utils which does the same.

frontend/tests/e2e/apply/happy-path-application-submission.spec.ts
doug-s-nava: @Bhavna-Ramachandran - I'm happy to do whatever you two feel is best here, but I can't tell what that is. It sounds like we don't need Aoy's code, and we can keep the timeout, but please clarify for me

Bhavna-Ramachandran: Sorry if my earlier message wasn’t clear-
 
We dont need this suggested code. The selectFormInclusionOption(...) utility already handles the same plus extras 
 
yes we should keep the extra timeout call (as notied in the annotation above it - I added that as I noticed forms table takes time to load)

PR #9007, opened 2026-03-12
[Issue #8956] Remove orphaned attachments from submission zip file

### LINKED ISSUE #8956: BUG: Orphaned attachments are included in the submission zip file
### Summary

After having orphaned attachments included in the application, they are still being included in the submission zip file.

Expected behavior: Orphaned attachments are not included in the submission zip file.

This was proven on this application:
https://staging.simpler.grants.gov/applications/27f75a8a-bdd1-4d2a-8526-4783703c9344 

### Acceptance criteria

- [ ] Orphaned attachments are not included in the submission zip file
- [ ] UAT Passed
  - [ ] Orphaned attachment triggered but not included in the submission zip file

mikehgrantsgov:
## Summary

Fixes #8956 

## Changes proposed

- `process_application_attachments` now calls `_collect_referenced_attachment_ids()` to determine which attachments are actually referenced in form responses before adding them to the submission zip
- Orphaned attachments (uploaded but not referenced in any form field) are skipped with a warning log
- Updated `test_run_task` to reflect that attachments are orphaned and correctly excluded
- Added test case to cover behavior

## Context for reviewers

The orphan detection logic already existed in `create_attachment_mapping()` to exclude unreferenced attachments from the XML, but wasn't being applied when building the zip. This change wires that same logic into the zip-building step so the zip and XML are consistent.

## Validation steps

See updated test cases.

### DIFF
--- a/api/src/services/xml_generation/utils/attachment_mapping.py
+++ b/api/src/services/xml_generation/utils/attachment_mapping.py
@@ -6,6 +6,7 @@ from typing import Any
 from src.db.models.competition_models import Application, ApplicationAttachment
 from src.form_schema.rule_processing.json_rule_context import JsonRuleConfig, JsonRuleContext
 from src.form_schema.rule_processing.json_rule_processor import process_rule_schema_for_context
+from src.services.applications.application_validation import is_form_required
 from src.services.xml_generation.models.attachment import HASH_ALGORITHM, AttachmentFile
 
 logger = logging.getLogger(__name__)
@@ -63,8 +64,7 @@ def _collect_referenced_attachment_ids(application: Application) -> set[str]:
     """
     referenced: set[str] = set()
     for app_form in application.application_forms:
-        is_required = app_form.competition_form.is_required
-        if not is_required and app_form.is_included_in_submission is not True:
+        if not is_form_required(app_form) and app_form.is_included_in_submission is not True:
             continue
 
         context = JsonRuleContext(app_form, config=_ATTACHMENT_COLLECTION_CONFIG)


api/src/task/apply/create_application_submission_task.py
```diff
@@ -349,7 +352,24 @@ def process_application_attachments(self, submission: SubmissionContainer) -> No
         }
         logger.info("Processing attachments for application submission", extra=log_extra)
 
+        # Only include attachments that are referenced in form responses.
+        # Orphaned attachments (not referenced anywhere) are excluded from the zip.
+        referenced_ids = _collect_referenced_attachment_ids(submission.application)
```
chouinar, line 357: I'm probably missing some context on this work. The way it currently works is very intentional, I set it up to add all attachments to the ZIP regardless of whether they're referenced from a form because attachments don't need to be on a form in our system. I know we adjusted this in the XML because the legacy system does need that, but ours doesn't.

Is it causing issues for working with the old systems if there are are attachments present in the ZIP? If not, I'd strongly push back against this change as it's removing functionality that is intentionally different from the legacy approach. If a user uploads a file to an application without putting it on a form, what is the user experience to let them know that it will not be included when sent to a grantor?

mikehgrantsgov:
@chouinar I think this captures the spirit of what we're going after from yesterday's meeting. Open to your thoughts!

api/src/services/xml_generation/utils/attachment_mapping.py
```diff
@@ -44,25 +54,23 @@ def to_dict(self) -> dict[str, Any]:
...
-            value = response.get(field_name)
-            if isinstance(value, str):
-                referenced.add(value)
-            elif isinstance(value, list):
-                for item in value:
-                    if isinstance(item, str):
-                        referenced.add(item)
+        is_required = app_form.competition_form.is_required
+        if not is_required and app_form.is_included_in_submission is not True:
+            continue
```
chouinar: Not a blocker, but I wonder if this should be some sort of utility. I also had to add a check for this in a PR I have out.

One very minor nitpick - instead of using `app_form.competition_form.is_required` directly there's a utility function that the app form validation logic uses that we should reuse. It's the same, but we assume we'll change the logic in the future, so the util makes it the same.

mikehgrantsgov: Ah makes sense, updated that!

Issue #8997, opened 2026-03-11
SOAP/Proxy: Stop logging an error when there's an HTTPError in auth


jakobpederson:
### Summary

Currently if there's an `HTTPError` when checking permission and it fails, it falls through and logs an exception. We want to add another check here:
https://github.com/HHS/simpler-grants-gov/blob/main/api/src/legacy_soap_api/legacy_soap_api_routes.py#L101

To catch the HTTPError that doesn't log an exception.

### Acceptance criteria

- [ ] not having the correct permission does not log the exception

PR #9012, opened 2026-03-12
[Issue #8997] SOAP/Proxy: handle HTTPError Privilege issue without logging exception

jakobpederson:
## Summary

Fixes / Work for #8997 

## Changes proposed

Handled the HTTPError that results when user does not have Privilege to access an application.

## Context for reviewers

Previously this was being logged as an exception which was setting off alarms in New Relic. This handles it more gracefully and lessens alarm noise.

### DIFF
--- a/api/src/legacy_soap_api/legacy_soap_api_routes.py
+++ b/api/src/legacy_soap_api/legacy_soap_api_routes.py
@@ -1,6 +1,5 @@
 import logging
 
-from apiflask.exceptions import HTTPError
 from flask import request
 
 import src.adapters.db as db
@@ -8,6 +7,7 @@ import src.adapters.db.flask_db as flask_db
 from src.legacy_soap_api.legacy_soap_api_auth import (
     MTLS_CERT_HEADER_KEY,
     USE_SOAP_JWT_HEADER_KEY,
+    SOAPClientUserDoesNotHavePermission,
     get_soap_auth,
 )
 from src.legacy_soap_api.legacy_soap_api_blueprint import legacy_soap_api_blueprint
@@ -99,7 +99,7 @@ def simpler_soap_api_route(
         return get_simpler_soap_response(
             soap_request, soap_proxy_response, db_session
         ).to_flask_response()
-    except HTTPError:
+    except SOAPClientUserDoesNotHavePermission:
         msg = "soap_client_certificate: User did not have permission to access this application"
         logger.info(
             msg=msg,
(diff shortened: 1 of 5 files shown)


api/src/legacy_soap_api/legacy_soap_api_routes.py
```diff
@@ -98,6 +99,15 @@ def simpler_soap_api_route(
...
         ).to_flask_response()
+    except HTTPError:
+        msg = "soap_client_certificate: User did not have permission to access this application"
+        logger.info(
+            msg=msg,
+            extra={
+                "soap_api_event": LegacySoapApiEvent.ERROR_CALLING_SIMPLER,
+            },
+        )
+        return soap_proxy_response.to_flask_response()
```
kkrug: Is this overly broad? Is it possible to hit this HTTPError from something other than a permissions issue?

api/tests/src/legacy_soap_api/test_legacy_soap_api_routes.py
```diff
@@ -244,3 +261,42 @@ def test_simpler_getapplicationzip_operation_returns_not_found_response_includes
...
+        response = client.post(
+            full_path, data=etree.tostring(envelope), headers={"Use-Simpler-Override": "true"}
+        )
+    assert response.status_code == 500
+    post_message = next(
+        record
+        for record in caplog.records
+        if record.message == "User did not have permission to access this application"
+    )
+    assert post_message.message == "User did not have permission to access this application"
```
kkrug: Is this testing the correct log message (i.e. the one you just added, `"soap_client_certificate: User did not have permission to access this application"`)? I think this is testing the inner log message from get_application_zip_response.py. 

----------------------------------------------------------------------------------------------------------

Also, is there actually an assertion here that verifies the AC of "not having the correct permission does not log the exception?" I would think we need something verifies that logging did not occur, something like:
`error_records = [r for r in caplog.records if r.levelno >= logging.ERROR and "Unable to process Simpler SOAP proxy response" in r.message]
assert len(error_records) == 0`

api/tests/src/legacy_soap_api/test_legacy_soap_api_routes.py
```diff
@@ -244,3 +261,42 @@ def test_simpler_getapplicationzip_operation_returns_not_found_response_includes
...
+        mock_get_auth.return_value = SOAPAuth(certificate=mock_client_cert)
+        response = client.post(
+            full_path, data=etree.tostring(envelope), headers={"Use-Simpler-Override": "true"}
+        )
+    assert response.status_code == 500
+    post_message = next(
+        record
+        for record in caplog.records
+        if record.message == "User did not have permission to access this application"
+    )
```
kkrug: I think the `assert` after this is redundant - if the message isn't found within `next()` it will raise an error and fail the test, though that would be with an unhelpful failure message. You could do a list comprehension with  `assert len(...) == 1`.

api/src/legacy_soap_api/legacy_soap_api_routes.py
jakobpederson: I think the thinking is we only ever explicitly raise an HTTPError in one place in the SOAP/Proxy. I think in the space between this check and the actual response we won't see any other HTTPError. That being said I think you're right and we could easily raise a custom exception so I'll add that.

PR #9019, opened 2026-03-12
[Issue #8862] Improve /saved-search-queries breadcrumbs

### LINKED ISSUE #8862: Improve "Saved search queries" breadcrumbs
### Summary

Swap "Home" in the "Saved search queries" breadcrumbs for "Workspace" which will link to `/dashboard`.  

Note: Blocks #8848, which is a separate task to update the Workspace section routes. The "Saved search queries" route will change from `/saved-search-queries` to `/workspace/saved-search-queries`, `/dashboard` will change to `/workspace`, and other pages will have `/workspace` added to the root of their routes. All those changes need happen together. 

### Acceptance criteria

- [ ] "Saved search queries" page's breadcrumbs are "Workspace > Saved search queries"

andycochran:
## Summary

Fixes #8862 

## Changes proposed

- Delete `SAVED_SEARCHES_CRUMBS` constant
- Define breadcrumbs inline in the page where they're used

image

## Context for reviewers

Per new guidance, all breadcrumbs are to be managed consistently, i.e. on their page (not in the constant file). This is part of a larger effort to refactor breadcrumbs. In a subsequent ticket, `/saved-search-queries` will be moved to `/workspace/saved-search-queries`. But that'll happen along with other route changes, when `/dashboard` will also change to `/workspace`. 

## Validation steps

- Sign in
- View the `/saved-search-queries` page
- The breadcrumbs should read "Workspace > Saved search queries"
- The "Workspace" breadcrumb should take you to `/dashboard`

### DIFF
--- a/frontend/src/app/[locale]/(base)/saved-search-queries/page.tsx
+++ b/frontend/src/app/[locale]/(base)/saved-search-queries/page.tsx
@@ -91,11 +91,11 @@ export default async function SavedSearchQueries({
         <Breadcrumbs
           breadcrumbList={[
             {
-              title: `${t("breadcrumbWorkspace")}`,
+              title: t("breadcrumbWorkspace"),
               path: `/dashboard`,
             },
             {
-              title: `${t("breadcrumbSavedQueries")}`,
+              title: t("breadcrumbSavedQueries"),
               path: `/saved-search-queries`,
             },
           ]}


frontend/src/app/[locale]/(base)/saved-search-queries/page.tsx
```diff
@@ -89,7 +88,18 @@ export default async function SavedSearchQueries({
...
+        <Breadcrumbs
+          breadcrumbList={[
+            {
+              title: `${t("breadcrumbWorkspace")}`,
+              path: `/dashboard`,
+            },
+            {
+              title: `${t("breadcrumbSavedQueries")}`,
+              path: `/saved-search-queries`,
+            },
```
Comment (freddieyebra) [inline frontend/src/app/[locale]/(base)/saved-search-queries/page.tsx:100]: Glad these will be translated now! Since the translation function returns strings, can we remove the interpolation here to read:
```suggestion
            {
              title: t("breadcrumbWorkspace"),
              path: `/dashboard`,
            },
            {
              title: t("breadcrumbSavedQueries"),
              path: `/saved-search-queries`,
            },
```

frontend/src/constants/breadcrumbs.ts
```diff
@@ -10,15 +10,10 @@ const SAVED_OPPORTUNITIES: Breadcrumb = {
...
-  path: "/saved-search-queries/",
-};
 
 // page breadcrumbs
 export const SEARCH_CRUMBS: Breadcrumb[] = [HOME, SEARCH];
 export const SAVED_OPPORTUNITIES_CRUMBS: Breadcrumb[] = [
   HOME,
   SAVED_OPPORTUNITIES,
 ];
-export const SAVED_SEARCHES_CRUMBS: Breadcrumb[] = [HOME, SAVED_SEARCHES];
```
freddieyebra, line 24: Just want to double check and confirm that we _do_ want to remove **Home** from the breadcrumb path.

frontend/src/i18n/messages/en/index.ts
```diff
@@ -955,6 +955,8 @@ export const messages = {
   },
   SavedSearches: {
     heading: "Saved search queries",
+    breadcrumbWorkspace: "Workspace",
```
freddieyebra, line 958: Could we add a unit test here that verifies that breadcrumb text we would expect to see on the page does indeed exist?

freddieyebra (changes requested):

frontend/src/i18n/messages/en/index.ts
freddieyebra: Not a blocker, but may be worth a follow up ticket; would it be beneficial to make the breadcrumb values that will be used across other pages to be constants such as `"Workspace"`?  so if we need/want to change name in future we can just update the constant value? We will also be able to use that constant for other breadcrumbs.

Another idea would be have a key in the translations index.ts for breadcrumb values.

frontend/src/constants/breadcrumbs.ts
andycochran: Yes. Per [guidance](https://docs.google.com/document/d/1gerqOGrsXJy3BX2nPGOskMHCT3-o1jiqrT54eQ7qAg4/edit?usp=sharing)… 
> Do not include "Home" in breadcrumbs. It's [not recommended](https://www.nngroup.com/articles/breadcrumbs/#:~:text=However%2C%20note%20that%20duplicating%20the%20Home%20link%20in%20both%20the%20global%20navigation%20and%20the%20breadcrumb%20trail%20is%20not%20recommended%20%E2%80%94%20one%20or%20the%20other%20is%20fine.) to duplicate "Home" in both the global navigation and breadcrumbs. This deviates from USWDS guidance — which says to [include "Home" breadcrumbs](https://designsystem.digital.gov/components/breadcrumb/#guidance:~:text=Start%20with%20the%20word%20%E2%80%9CHome%E2%80%9D.%20Rather%20than%20using%20a%20house%20icon%2C%20spell%20out%20the%20word%20%E2%80%9CHome%E2%80%9D%20as%20the%20first%20link%20in%20the%20breadcrumbs.) — as the USWDS Header component does not prescribe/require a "Home" link.

freddieyebra: thanks for confirming Andy

PR #9020, opened 2026-03-12
[Issue #8926] During application submission, populate fields needed for award rec

### LINKED ISSUE #8926: Backend - Modify our application submission job to populate the new fields
### Summary

See [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Modify-our-application-submission-job-to-populate-the-new-fields)

Following the new fields being added to the application submission table, we need to modify the application submission job to populate them as follows:

- `application_submission_number` should be populated as TODO
- `project_title` should be pulled from the forms as described [in the tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Pull-Fields-from-Forms)
- `total_requested_amount` should be pulled from the forms as described [in the tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Pull-Fields-from-Forms)

### Acceptance criteria

- [ ] Job updated with the three fields listed above
- [ ] Tests created/updated as necessary

chouinar:
## Summary
Fixes #8926 

## Changes proposed
* Populate `application_submission_number` / `project_title` / `total_requested_amount` when creating the application submission
* Added a simple util for fetching a field from the forms that may be on an application.

## Context for reviewers
These are all fields we want for award recommendation work as they're needed in the UI we display to grantors. They're all nullable, although with the way this is written, the new submission number won't ever be. The others can be null if the fields for them aren't present.

https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Application-Submission

## Validation steps
Will test manually (besides the tests written) once it's in staging - my local setup needs a bit of fixing to work with the frontend / handle submission all the way through.

### DIFF
--- a/api/src/services/applications/get_field_from_application.py
+++ b/api/src/services/applications/get_field_from_application.py
@@ -1,0 +1,106 @@
+import dataclasses
+import logging
+import uuid
+from decimal import Decimal
+from typing import Any
+
+from src.db.models.competition_models import Application
+from src.form_schema.forms import ProjectAbstractSummary_v2_0, SF424_v4_0, SF424a_v1_0
+from src.services.applications.application_validation import is_form_required
+from src.util.decimal_util import convert_monetary_field, quantize_decimal
+from src.util.dict_util import get_nested_value
+
+logger = logging.getLogger(__name__)
+
+
+@dataclasses.dataclass
+class FormField:
+    form_id: uuid.UUID
+    field: str  # field should be defined as a path in the JSON schema of format a.b.c
+
+
+def get_field_from_application(application: Application, form_fields: list[FormField]) -> Any:
+    """
+    Function to get a field from an application's forms.
+
+    A list of forms + fields can be passed in and the first non-None value
+    will be returned. If no form has the value set, None will be returned.
+
+    Typing is not validated at this level, you should validate the value
+    in the calling function.
+
+    Form fields can be a jsonpath (eg. "a.b.c" against {"a": {"b": {"c": 123}}}" will return 123)
+    """
+
+    # Setup a mapping of form_id to application form
+    # Skipping all application forms that aren't going
+    # to be in the submission.
+    application_form_map = {}
+    for application_form in application.application_forms:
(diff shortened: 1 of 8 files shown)


api/src/services/applications/get_field_from_application.py
```diff
@@ -0,0 +1,106 @@
...
+        logger.warning(
+            "Project title for application is not a string",
+            extra={"application_id": application.application_id},
+        )
+        return None
+
+    return project_title
+
+
+REQUESTED_AMOUNT_FORM_FIELDS = [
```
ErinPattisonNava, line 84: I'm concerned this isn't going to scale well, especially when we start adding new forms or versioning them.  Thoughts on shifting to a more configuration driven approach, where we could annotate fields or have a mapping object per form template that talks about how that form template's fields map to SGM concepts?

chouinar: @ErinPattisonNava - This won't scale, I don't expect it to, and don't expect it to even be expanded on before we replace it. I agree, the proper approach is likely in some of the work we've talked about of being able to flag types in the JSON schema and generate something from that that says "field X is SGG type A". But as that's likely a quad-long deliverable, I tried to make the most minimal / out-of-the-way approach that doesn't add any burden to the apply team for now (eg. putting this config IN a given form would mean syncing all these forms and maintaining it).

This is for a very rough prototype that I know we'll have to rework already - didn't want to weave that too much into the apply code when I know we'll probably delete / rewrite this in the next 6 months.

joshtonava (approved):
Approving with the knowledge we will end up needing to push this information into the JSON schema so it can be better tracked for auditability

PR #9022, opened 2026-03-12
[Unticketed] Correct a recent incorrect detail in docs

chouinar:
## Changes proposed
* Fixes an incorrect bit of guidance in the API dics

## Context for reviewers
You can't add these values to the override file because if you do, the API service won't work right. Instead we need the values set in your shell.

## Validation steps
This is my actually local dev setup.


documentation/api/development.md
```diff
@@ -144,19 +144,40 @@ Running in the native/local approach may require additional packages to be insta
 
 4. This still requires Docker as the database and services all still run within Docker.
 5. Certain environment variables require different paths when running outside of Docker.
-   You can handle this by setting these into your `override.env` file.
+   You can handle this by setting these environment variables yourself via a means other
+   than our .env files. One approach is to have an `.envrc` in the api directory that works with [direnv](https://direnv.net/)
```
mdragon:
```suggestion
   than our .env files. One approach is to have an `.envrc` in the api directory that works with [direnv](https://direnv.net/) or set these in your `.bashrc` or `.zshrc`
```

chouinar: Reworded it a bit as it was a little clunky - mentioned those shell files before my alternative.

PR #9023, opened 2026-03-12
[Issue #9021] Add retries if Login.gov fails when we ask for token details

### LINKED ISSUE #9021: Add retries for when Login.gov Token Request fails
### Summary

We occasionally see failures when we call Login.gov to get information about the user based on the token returned to us when Login.gov calls back to us.  

### Acceptance criteria

- [ ] Add a basic retry to attempt the call 3 times before giving up. 
- [ ] Successful requests break out of the retry cycle.
- [ ] Ensure we're testing both retries allowing for success and retries that still fail on subsequent retries

mdragon:
## Summary

Fixes #9021  

## Changes proposed

Introduce a retry of up to 3 times for our call to Login.gov to fetch information about the token holder

## Context for reviewers

We see occasional failures, once every few weeks, where upon a callback from Login.gov our request back to Login to get information about the holder of the token fails with a unspecified 500 error. This currently results in the user not successfully signing in. This change would give that user an opportunity to still get logged in.

## Validation steps

- [x] Unit tests pass
- [ ] Login still works in Staging after deploy following merge to main

### DIFF
--- a/api/src/services/users/login_gov_callback_handler.py
+++ b/api/src/services/users/login_gov_callback_handler.py
@@ -181,7 +181,8 @@ def handle_login_gov_token(
                     extra={"tries": tries, "limit": limit},
                 )
                 continue
-
+        # if it's not an error we should break out of the loop since it was a successful call
+        break
     # Process the token response from login.gov
     # which will create/update a user in the DB
     return _process_token(db_session, response.id_token, login_gov_data.nonce)
--- a/api/tests/src/api/users/test_user_route_login.py
+++ b/api/tests/src/api/users/test_user_route_login.py
@@ -274,6 +274,12 @@ def test_user_callback_new_user_302(
     )
     assert db_state is None
 
+    # this checks that we didn't have to retry for this to return it's result
+    # when no retries are specified we start at 0 and decrement for each attempt
+    # so -1 means we only tried once
+    # without the fix Michael noted in the PR, this would incorrectly be -3, failing this test
+    assert mock_oauth_client.retries[code] == -1
+
 
 def test_user_callback_existing_user_302(
     client, db_session, enable_factory_create, mock_oauth_client, private_rsa_key


api/src/adapters/oauth/login_gov/mock_login_gov_oauth_client.py
```diff
@@ -7,15 +7,29 @@ class MockLoginGovOauthClient(BaseOauthClient):
     def __init__(self) -> None:
         self.responses: dict[str, OauthTokenResponse] = {}
 
+        # Used to control testing of retry behavior for Login.gov token lookup calls
+        self.retries: dict[str, int] = {}
+
     def add_token_response(self, code: str, response: OauthTokenResponse) -> None:
         self.responses[code] = response
```
chouinar, line 14: Rather than setting the retries by using the dict directly, what if we add a param to the add token function to do it?

```suggestion
        self.retries: dict[str, int] = {}

def add_token_response(self, code: str, response: OauthTokenResponse, retry_count: int = 0) -> None:
      self.responses[code] = response
      self.retries[code] = retry_count
```

Then that simplifies the logic below a bit as it'll always be set.

api/src/services/users/login_gov_callback_handler.py
```diff
@@ -161,15 +161,26 @@ def handle_login_gov_token(
...
+        # If this request failed, we'll assume we're the issue and 500
+        if response.is_error_response():
+            if tries == limit:
+                raise_flask_error(500, response.error_description)
+            else:
+                logger.info(
+                    "Retrying call to Login.gov after receiving error",
+                    extra={"tries": tries, "limit": limit},
+                )
+                continue
```
chouinar, line 183: I think you meant to do this - as written it would always call 3 times as the loop never breaks before 3 tries.

```suggestion
            else:
                logger.info(
                    "Retrying call to Login.gov after receiving error",
                    extra={"tries": tries, "limit": limit},
                )
                break
```

Alternatively, if we wanted to use tenacity, seems they have a `TryAgain` exception we could use - although I don't think it's that big a deal.

api/src/services/users/login_gov_callback_handler.py
```diff
@@ -161,15 +161,26 @@ def handle_login_gov_token(
...
+            OauthTokenRequest(
+                code=login_gov_data.code, client_assertion=get_login_gov_client_assertion()
+            )
         )
-    )
 
-    # If this request failed, we'll assume we're the issue and 500
-    if response.is_error_response():
-        raise_flask_error(500, response.error_description)
+        # If this request failed, we'll assume we're the issue and 500
```
chouinar: Probably should update this as with the retries it isn't quite right?

api/src/services/users/login_gov_callback_handler.py
mdragon: I think this is right? break would exit the loop? Where as continue will jump back to the top of the loop? Thought I do think now as written, it's extra, just making the loop explicit?

analytics/pyproject.toml
```diff
@@ -31,7 +31,7 @@ requests = "^2.32.5"
 urllib3 = "^2.6.3"
 
 [tool.poetry.group.dev.dependencies]
-black = "^24.0.0"
+black = "^26.3.1"
```
chouinar: I think you might have mixed up your PRs, you put some of the dependency upgrades here for analytics.

api/src/services/users/login_gov_callback_handler.py
chouinar: I was reading it wrong, but the issue is still there.

How does a successful get token request break out of the loop? The way this is written, regardless of success/error, it will always call exactly 3 times.

analytics/pyproject.toml
mdragon: Thanks. Fixed.

api/src/services/users/login_gov_callback_handler.py
mdragon: The top level if is only entered if the response was an error. So the success path skips both of these second tier ifs.

mdragon: Let me know if you think there's a way to make that easier to read.

chouinar: I checked out the code. I added a print statement to the mock get_token function.

For the happy path test of `test_user_login_flow_happy_path_302` I ran it. It printed out 3 times.

Yes, if it's a success, it skips these if-statements. So your loop is:

```py
while tries < limit:
    tries += 1
    response = get_token(...)

    if response.is_error():
         # this doesn't matter for success
```

If the first time we call get_token it is successful, how does it exit the loop?

mdragon: Ok, i see I was printing in the mock, which only loops when it has to (correctly) so I appreciate your patience here. You were correct, we were missing a `break`.

I added a test which fails without the break, and now passes with that added.

api/tests/src/api/users/test_user_route_login.py
```diff
@@ -274,6 +274,12 @@ def test_user_callback_new_user_302(
     )
     assert db_state is None
 
+    # this checks that we didn't have to retry for this to return it's result
+    # when no retries are specified we start at 0 and decrement for each attempt
+    # so -1 means we only tried once
+    # without the fix Michael noted in the PR, this would incorrectly be -3, failing this test
+    assert mock_oauth_client.retries[code] == -1
```
mdragon, line 281: Without the break, this test fails:
```
>       assert mock_oauth_client.retries[code] == -1
E       assert -3 == -1

tests/src/api/users/test_user_route_login.py:281: AssertionError
```

PR #9024, opened 2026-03-12
Refactor Node setup input variable names

aoysimmons-GH:
Updated input variable names for Node setup.

## Summary

Fixes / Work for #ISSUE 

## Changes proposed

Updated the three references in to use the declared lowercase input names.


aoysimmons-GH (commented):
Small fix:
Updated the three references in to use the declared lowercase input names.

Issue #8672, opened 2026-02-26
Selection Methodology


kevlyu:
### Summary

Static view
Editable view

### Acceptance criteria

tbd

kevlyu:
@chay-REIsys fill in ACs

chaitanya-duddella:
https://staging.simpler.grants.gov/award-recommendation/b9c15d13-8f13-4e15-80b8-3cf5acf8481c/edit

PR #9028, opened 2026-03-12
[Issue #8672] - Dynamic Hero buttons for edit/preview page/ Recommendation Selection Method Section

chay-REIsys:
## Summary

 Work for #8672 

## Changes proposed

Recommendation Section Added for preview/Edit AR page
Buttons are configured as per preview/edit page
(figma link in ticket)
## Context for reviewers

As of now buttons for award recommendation header are hardcoded and this is used across pages for Award recommendation based of button configuration loading buttons on header. Now the buttons are configured based off the  preview/edit page and recommendation section is added for preview and edit page of award recommendation ( data is mocked/hardcoded - this will be changed after get award recommendation endpoint implementation)

## Validation steps

- fetch new changes
- ensure awardRecommendationOff flag is set to false
- npm run build and run locally
- http://localhost:3000/award-recommendation/4626af88-631a-4c68-8bfb-f4b75145420b/edit takes us to edit page and http://localhost:3000/award-recommendation/4626af88-631a-4c68-8bfb-f4b75145420b takes us to preview page (as of now data is mocked)
- clicking on edit button takes us to edit page 
- clicking on preview button on edit page takes us to view only page
- save and submit buttons doesn't do anything as of now


frontend/src/components/award-recommendation/AwardRecommendationHero.tsx
```diff
@@ -62,16 +85,38 @@ export default async function AwardRecommendationHero({
...
+                    return (
+                      <Link key={index} href={button.href}>
+                        <Button
+                          type="button"
+                          outline={button.outline}
+                          className="width-auto"
+                        >
+                          {button.label}
+                        </Button>
+                      </Link>
```
btabaska: accessibility:  We’re nesting a button inside a Link (<a>), which creates nested interactive controls and can cause confusing keyboard/screen-reader behavior. We’ll want to render a single interactive element here (for example, style the link as a button instead of wrapping Button inside Link).

frontend/src/components/award-recommendation/AwardRecommendationHero.tsx
```diff
@@ -62,16 +85,38 @@ export default async function AwardRecommendationHero({
...
+                      <form key={index} action={button.formAction}>
+                        <Button
+                          type="submit"
+                          outline={button.outline}
+                          disabled={button.disabled}
+                          className="width-auto"
+                        >
+                          {button.label}
+                        </Button>
+                      </form>
```
btabaska: These action forms only contain the submit button, but the editable fields live elsewhere on the page. That means `saveAwardRecommendation` / `submitAwardRecommendationForReview` won’t receive the section inputs in FormData. We should wire buttons to the same form that wraps `additional_info, award_selection_method, award_selection_details, and other_key_information`.

frontend/src/app/[locale]/(base)/award-recommendation/[id]/actions.ts
```diff
@@ -0,0 +1,69 @@
...
+export async function submitAwardRecommendationForReview(
+  formData: FormData,
+): Promise<AwardRecommendationActionResponse> {
+  // Extract form data
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const rawFormData = {
+    additionalInfo: formData.get("additional_info") as string,
+    awardSelectionMethod: formData.get("award_selection_method") as string,
+    awardSelectionDetails: formData.get("award_selection_details") as string,
+    otherKeyInformation: formData.get("other_key_information") as string,
```
Comment (btabaska) [inline frontend/src/app/[locale]/(base)/award-recommendation/[id]/actions.ts]: `formData.get(...) `as string is unsafe because FormData.get can return null or File. We’ll want a small helper/type guard so we don’t silently accept invalid types and pass bad payloads once the endpoint is wired.

frontend/src/app/[locale]/(base)/award-recommendation/[id]/edit/page.test.tsx
```diff
@@ -239,6 +239,62 @@ describe("AwardRecommendationEditPage", () => {
       expect(screen.queryByText("selectionMethod")).not.toBeInTheDocument();
       expect(screen.queryByText("Merit Review")).not.toBeInTheDocument();
     });
+
+    it("renders the recommendation section on edit page", async () => {
```
Comment (btabaska) [inline frontend/src/app/[locale]/(base)/award-recommendation/[id]/edit/page.test.tsx:243]: Nice coverage on rendering fields.

frontend/src/app/[locale]/(base)/award-recommendation/[id]/edit/page.tsx
```diff
@@ -132,6 +139,104 @@ const OpportunitySection = ({
   );
 };
 
+interface RecommendationSectionProps {
+  locale: string;
+}
+
+const RecommendationSection = ({
+  locale: _locale,
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/award-recommendation/[id]/edit/page.tsx]: if you're not using locale, feel free to leave it out here

frontend/src/app/[locale]/(base)/award-recommendation/[id]/edit/page.tsx
```diff
@@ -132,6 +139,104 @@ const OpportunitySection = ({
...
+                  })}
+                </p>
+                <p className="text-base margin-top-1 margin-bottom-2">
+                  {t("recommendationMethodDetails.description", {
+                    defaultValue:
+                      "Add any additional information - including the selection factors used in the NOFO",
+                  })}
+                </p>
+                <CharacterCount
+                  id="award_selection_details"
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/award-recommendation/[id]/edit/page.tsx]: what's the rationale behind snake case here vs. dash case on the radio buttons? Ideally we should have the same format for all ids

frontend/src/app/[locale]/(base)/award-recommendation/[id]/edit/page.tsx
```diff
@@ -132,6 +139,104 @@ const OpportunitySection = ({
   );
 };
 
+interface RecommendationSectionProps {
+  locale: string;
+}
+
+const RecommendationSection = ({
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/award-recommendation/[id]/edit/page.tsx]: I'd recommend putting this in its own file. This file will get very big very quickly at this rate

frontend/src/components/award-recommendation/AwardRecommendationHero.tsx
doug-s-nava: this is a good point - if the submit button is in the hero, this form will need wrap more or less the entire page. I don't ... think ... that will cause any problems

frontend/src/components/award-recommendation/AwardRecommendationHero.tsx
doug-s-nava: Took me a while to find an example of doing it the right way. https://github.com/HHS/simpler-grants-gov/blob/c46cd41c09ca28b63b6f03958d4ea5d6af01c20c/frontend/src/app/%5Blocale%5D/(base)/saved-opportunities/page.tsx#L30

I'll make a ticket for cleaning up the nested buttons on other pages

frontend/src/app/[locale]/(base)/award-recommendation/[id]/edit/page.tsx
Comment (chaitanya-duddella) [reply frontend/src/app/[locale]/(base)/award-recommendation/[id]/edit/page.tsx]: updated as other id's

frontend/src/components/award-recommendation/AwardRecommendationHero.tsx
chaitanya-duddella: updated to use Link with class usa-button

frontend/src/app/[locale]/(base)/award-recommendation/[id]/actions.ts
Comment (chaitanya-duddella) [reply frontend/src/app/[locale]/(base)/award-recommendation/[id]/actions.ts]: done

frontend/src/app/[locale]/(base)/award-recommendation/[id]/edit/page.tsx
Comment (chaitanya-duddella) [reply frontend/src/app/[locale]/(base)/award-recommendation/[id]/edit/page.tsx]: done

frontend/src/components/award-recommendation/AwardRecommendationHero.tsx
doug-s-nava: for the form to work you'll need to move the `<form>` out to wrap the entire area of the page that contains inputs

Issue #8996, opened 2026-03-11
E2E form test refactors for consistency


doug-s-nava:
### Summary

after writing initial e2e tests for apply forms, revise approaches for consistency so that we can move forward with easy implementation of all future happy path and failure path tests

### Acceptance criteria

- [ ] sf424b happy path test created based on existing submission happy path test
- [ ] all happy path form tests use the same helper functionality
- [ ] helper functionality updated to use form and data config objects

PR #9034, opened 2026-03-13
[Issue #8996] e2e refactors for apply form tests

doug-s-nava:
## Summary

Fixes for #8996 

## Changes proposed

Refactors a number of pieces of apply form e2e testing work for standardization and ease of use as we add tests for more forms

## Validation steps

1. _VERIFY_: all tests pass in CI against local
2. _VERIFY_: all tests pass in CI against staging (see https://github.com/HHS/simpler-grants-gov/actions/runs/23056917555)

### DIFF
--- a/frontend/tests/e2e/utils/forms/verify-form-status-utils.ts
+++ b/frontend/tests/e2e/utils/forms/verify-form-status-utils.ts
@@ -85,8 +85,8 @@ export async function assertFormRowStatus(
     status === "complete"
       ? /no issues detected\.?|complete/i
       : /some issues found\.?|in progress/i;
-
-  await expect(formRow.getByText(statusPattern)).toBeVisible({
+  const text = formRow.getByText(statusPattern);
+  await expect(text).toBeVisible({
     timeout: 10000,
   });
 }
--- a/frontend/tests/e2e/apply/failure-path-sf424b.spec.ts
+++ b/frontend/tests/e2e/apply/failure-path-sf424b.spec.ts
@@ -62,7 +62,7 @@ test("SF-424B error validation - required fields and inline errors", async ({
     // On application page — verify form row status/messages
     await verifyFormStatusOnApplication(
       page,
-      "complete",
+      "incomplete",
       "SF-424B",
       applicationUrl,
     );


frontend/tests/e2e/apply/fixtures/sf424b-field-definitions.ts
```diff
@@ -0,0 +1,21 @@
+import { FORM_DEFAULTS } from "tests/e2e/utils/forms/form-defaults";
+import { FormFillFieldDefinitions } from "tests/e2e/utils/forms/general-forms-filling";
+
+export const fieldDefinitionsSF424B: FormFillFieldDefinitions = {
+  title: {
+    testId: "title",
```
doug-s-nava, line 6: simplified to use testIds since that is what Aoy's system expects. Do we lose any fidelity here?

frontend/tests/e2e/apply/fixtures/sfLLL-fill-data.ts
```diff
@@ -1,395 +1,337 @@
-import type { SflllEntityData } from "tests/e2e/apply/fixtures/test-data-for-sflll-forms.fixture";
 import { FORM_DEFAULTS } from "tests/e2e/utils/forms/form-defaults";
-import { FillFieldDefinition } from "tests/e2e/utils/forms/general-forms-filling";
+import { FormFillFieldDefinitions } from "tests/e2e/utils/forms/general-forms-filling";
+
+const fieldDefinitionsSFLLL: FormFillFieldDefinitions = {
```
doug-s-nava, line 4: change to key off of a common identifier that can be referenced for both the data and form definition

frontend/tests/e2e/apply/happy-path-sfLLL.spec.ts
```diff
@@ -1,53 +1,31 @@
...
+  await authenticateE2eUser(page, context, !!isMobile);
+
+  await createApplication(page, OPPORTUNITY_URL, testOrgLabel);
+
+  await fillForm(testInfo, page, SFLLL_FORM_CONFIG, SFLLL_TEST_DATA, false);
+
+  await page.waitForTimeout(5000);
+
+  // Verify form status after save
+  await verifyFormStatusAfterSave(page, "complete");
```
doug-s-nava: use Bhavna's status confirmation helper here

frontend/tests/e2e/apply/happy-path-sfLLL.spec.ts
```diff
@@ -1,53 +1,31 @@
...
+
+  const isMobile = testInfo.project.name.match(/[Mm]obile/);
+
+  await authenticateE2eUser(page, context, !!isMobile);
+
+  await createApplication(page, OPPORTUNITY_URL, testOrgLabel);
+
+  await fillForm(testInfo, page, SFLLL_FORM_CONFIG, SFLLL_TEST_DATA, false);
+
+  await page.waitForTimeout(5000);
```
doug-s-nava: ran into some flakiness without this timeout. Unsure if there is a better way to handle this systematically...

frontend/tests/e2e/utils/forms/general-forms-filling.ts
```diff
@@ -4,47 +4,61 @@ import { selectDropdownByValueOrLabel } from "tests/e2e/utils/select-dropdown-ut
 export interface FillFieldDefinition {
   testId?: string;
   selector?: string;
-  value: string;
   type: "text" | "dropdown";
-  section: string;
-  field?: string;
+  section?: string;
```
doug-s-nava, line 8: not all forms have sections

frontend/tests/e2e/utils/forms/general-forms-filling.ts
```diff
@@ -84,8 +95,10 @@ export async function fillForm(
       .first()
       .waitFor({ state: "visible", timeout: 35000 });
 
-    for (const field of fields) {
-      await fillField(testInfo, page, field);
+    for (const fieldDefinition of Object.entries(fields)) {
```
doug-s-nava, line 98: an important piece - this way we don't have to hardcode data into configuration, but correlate based on the shared identifier key

frontend/tests/e2e/utils/forms/verify-form-status-utils.ts
Bhavna-Ramachandran, line 1: ✅

frontend/tests/e2e/apply/happy-path-application-submission.spec.ts
Bhavna-Ramachandran, line 1: Everything looks good. Can we make this file also to use your new fixtures or do we want to do that in a new PR?

frontend/tests/e2e/utils/forms/general-forms-filling.ts
```diff
@@ -84,8 +95,10 @@ export async function fillForm(
...
+      const [fieldIdentifier, fieldConfig] = fieldDefinition;
+      const dataForField = data[fieldIdentifier];
```
Bhavna-Ramachandran, line 100: This will work now as we have all matching data, but we add more forms (more data and more probability to miss a value) this could silently return 'Undefined' and the same can get filled into the text field as string. Can we add a guard ( this is not necessarily a blocker) like below -

if (dataForField === undefined) {
  throw new Error(`No data provided for field "${fieldIdentifier}"`);
}

frontend/tests/e2e/apply/happy-path-application-submission.spec.ts
doug-s-nava: my thought was we could clean that up later

frontend/tests/e2e/utils/forms/general-forms-filling.ts
doug-s-nava: we may have cases where we want to skip filling some fields. We could still add some support here, but we'd need to align on some conventions - would we want the user to always enter an empty string instead of leaving the field undefined, for example? It's probably easier to leave it as is and allow a field be undefined, but it would be more explicit and maybe clearer to not allow undefined, and have the user enter `null` or `""` for any empty fields. Thoughts?

Bhavna-Ramachandran: Agreed. We can leave it as is , revisit when we expand and if this is an issue for any form schema

frontend/tests/e2e/apply/fixtures/sf424b-field-definitions.ts
aoysimmons-GH: Great question. Yes, there is a semantic fidelity loss, but it may be worth the tradeoff depending or do hybrid to   keep one flat, testId-keyed field map as the single source of truth for automation, while also keeping semantic groups (like federalAction and signature) so the data stays readable and maintainable.

aoysimmons-GH: Example below:

 // Section 1
  federal_action_type: {
    selector: "#federal_action_type",
    **value: "Grant",**
    type: "dropdown" as const,
    section: "Section 1",
    field: "Type of Federal Action",
  },
  // Section 2
  federal_action_status: {
    selector: "#federal_action_status",
    **value: "BidOffer",**
    type: "dropdown" as const,
    section: "Section 2",
    field: "Status of Federal Action",
  },
  // Section 3
  report_type: {
    selector: "#report_type",
    **value: "MaterialChange",**
    type: "dropdown" as const,
    section: "Section 3",
    field: "Report Type",
  },
  material_change_year: {
    testId: "material_change_year",
    **value: "2025",**
    type: "text" as const,
    section: "Section 3",
    field: "Material Change Year",

frontend/tests/e2e/utils/forms/general-forms-filling.ts
aoysimmons-GH: Additional, about Hybrid Data-Driven Testing I recommend above we will not have to worries about some form does not have section and same label on same page.

Issue #9025, opened 2026-03-12
SOAP/Proxy: modify setup-cert-user script to take parameters


jakobpederson:
### Summary

Currently we cannot easily set up a legacy_certificate for testing. If we modify th esetup-cert-user script to take parameters we could generate a legacy_certificate. This would facilitate general testing and the future plans to support local testing.

### Acceptance criteria

- [ ] can run command and get a legacy certificate with a specified serial number, using a specific agency, with designated privileges

jakobpederson:
Closing as not the correct solution as per conversation with Michael C.

PR #9037, opened 2026-03-13
[Issue #9025] Update setup-cert-user for manually set values

jakobpederson:
## Summary

Fixes / Work for #9025 

## Changes proposed

Update `setup-cert-user` to take more variables.

## Context for reviewers

In order to easily create test certificates we needed an easy way to generate all the necessary data. Updating the task was the easiest way to get that.

## Validation steps

- [ ] run command locally and check to see created `LegacyCertificate`
- [ ] is backwards compatible
- [ ] unittests passing

### DIFF
--- a/api/tests/src/task/certificates/test_setup_cert_user_task.py
+++ b/api/tests/src/task/certificates/test_setup_cert_user_task.py
@@ -58,10 +58,11 @@ class TestSetupCertUserTask(BaseTestClass):
     ):
         role = RoleFactory.create(is_agency_role=True)
         CERT_ID = f"{uuid.uuid4()}"
+        SERIAL_NUMBER = f"{random.randint(1000, 9999)}"
         result = SetupCertUserTask(
-            db_session, None, [str(role.role_id)], None, "garbage-agency-code", CERT_ID
+            db_session, None, [str(role.role_id)], SERIAL_NUMBER, "garbage-agency-code", CERT_ID
         ).setup_cert()
-        assert result == SetupCertUserTaskStatus.MISSING_REQUIRED_INPUT_VALUES
+        assert result == SetupCertUserTaskStatus.AGENCY_NOT_FOUND
         legacy_certificate = (
             db_session.query(LegacyCertificate).filter(LegacyCertificate.cert_id == CERT_ID).first()
         )


api/tests/src/task/certificates/test_setup_cert_user_task.py
```diff
@@ -42,11 +43,86 @@ def test_setup_cert_user_raises_error_if_no_tcertificate(
         self, enable_factory_create, db_session, caplog
     ):
         role = RoleFactory.create(is_agency_role=True)
-        tcert_id = str(uuid.uuid4())
-        result = SetupCertUserTask(db_session, tcert_id, [str(role.role_id)]).setup_cert()
-        assert result == SetupCertUserTaskStatus.TCERTIFICATE_NOT_FOUND
-        warning_messages = [r.getMessage() for r in caplog.records if r.levelname == "WARNING"]
-        assert "Tcertificate not found" in warning_messages
+        result = SetupCertUserTask(db_session, None, [str(role.role_id)]).setup_cert()
+        assert result == SetupCertUserTaskStatus.MISSING_REQUIRED_INPUT_VALUES
```
kkrug, line 47: This test was changed to pass in `None` instead of a random but invalid `tcertificate_id`. I think changing to `None` better aligns with the name of this test, but now there is now test for `TCERTIFICATE_NOT_FOUND`. We may want to consider adding another test for `test_setup_cert_user_raises_error_if_tcertificate_not_found`.

api/tests/src/task/certificates/test_setup_cert_user_task.py
```diff
@@ -42,11 +43,86 @@ def test_setup_cert_user_raises_error_if_no_tcertificate(
...
+
+    def test_setup_cert_user_fails_to_create_test_cert_if_missing_required_agency(
+        self, enable_factory_create, db_session, caplog
+    ):
+        role = RoleFactory.create(is_agency_role=True)
+        CERT_ID = f"{random.randint(1000, 9999)}"
```
kkrug: Is there anything preventing us from using `str(uuid.uuid4())` here instead`? It's low risk but using `randint` does provide a small chance of collisions.

api/src/task/certificates/setup_cert_user_task.py
```diff
@@ -36,24 +36,62 @@ class SetupCertUserTaskStatus(StrEnum):
...
         super().__init__(db_session)
         self.tcertificates_id = tcertificates_id
         self.role_ids = role_ids
 
+        """
+        In order to manually create a test certificate these values are required and enter no value for tcertificates-id
+
+        To generate a test cert locally run:
+            make cmd args="task setup-cert-user --role-ids 446bafb9-41ee-46ac-8584-889aedcd5142 --agency-code <agency code> --serial-number <serial number> --cert-id <cert id>"
+        """
```
kkrug: This may just be a style nitpick but it is unconventional to use a docstring in place of inline comments (I know multiline comments in Python are annoying though).

api/tests/src/task/certificates/test_setup_cert_user_task.py
jakobpederson: Good catch, I'll update it.

api/tests/src/task/certificates/test_setup_cert_user_task.py
jakobpederson: I didn't mostly because the cert isn't actually formatted like a uuid. But it would work if I put one in, so I'll swap it in.

api/src/task/certificates/setup_cert_user_task.py
jakobpederson: I'll take them out, I'd rather adhere to convetion.

api/tests/src/task/certificates/test_setup_cert_user_task.py
```diff
@@ -42,11 +43,93 @@ def test_setup_cert_user_raises_error_if_no_tcertificate(
...
-        warning_messages = [r.getMessage() for r in caplog.records if r.levelname == "WARNING"]
-        assert "Tcertificate not found" in warning_messages
+
+    def test_setup_cert_user_fails_to_create_test_cert_if_missing_required_agency(
+        self, enable_factory_create, db_session, caplog
+    ):
+        role = RoleFactory.create(is_agency_role=True)
+        CERT_ID = f"{uuid.uuid4()}"
+        result = SetupCertUserTask(
+            db_session, None, [str(role.role_id)], None, "garbage-agency-code", CERT_ID
```
kkrug: Could this be returning `SetupCertUserTaskStatus.MISSING_REQUIRED_INPUT_VALUES` because the `serial_number` is `None` and not because the `agency` is invalid (which is what this should be testing)?

jakobpederson:
> I do not see any ACs around ensuring that this must be backwards compatible. Should that be an AC for this update?

It shouldn't break any current functionality, and I think that's proven by the tests still passing. I'll add the AC.

api/src/task/certificates/setup_cert_user_task.py
```diff
@@ -36,24 +36,57 @@ class SetupCertUserTaskStatus(StrEnum):
...
-@click.option("--tcertificates-id", "-t", help="tcertificates_id on Staging Tcertificate")
-@click.option("--role-ids", "-t", help="role_id of role that needs to be added", multiple=True)
+@click.option(
+    "--tcertificates-id",
+    help="tcertificates_id on Staging Tcertificate, if set to none then check for manual settings",
+)
+@click.option("--role-ids", help="role_id of role that needs to be added", multiple=True)
+@click.option("--serial-number", default=None, help="manually set the serial number of certificate")
+@click.option("--agency-code", default=None, help="manually set the agency_code of the certificate")
+@click.option("--cert-id", default=None, help="manually set the cert_id of the certificate")
```
chouinar, line 50: Is the serial number something that we would want logged / are they PII? Parameters passed into an ECS task are logged / not kept secret.

When I suggested we setup a cert, I was thinking we'd do that in the DB manually since we just need to do it once - this task serves a different purpose - I don't think we should be changing this to serve a new purpose of creating a brand new cert.

chouinar: + I don't think we should be creating fake certs in this way in prod - this would need to be blocked from running in certain envs (maybe training?)

Issue #9017, opened 2026-03-12
SF424B Submission E2E: Add submission failure and application history validation


Bhavna-Ramachandran:
### Summary

#8117 : The initial SF424B E2E test covers - starting the application, filling out the SF424B form, and submitting the application with the conditional Lobbying form marked "No".

This ticket focuses on automating the remaining validation steps, including:
-  Confirming error messages appear when attempting to submit without filling the required form
-  Verifying error messages for empty mandatory fields within the SF424B form
-  Confirming the application history reflects all relevant events (organization added, failed submit, form saved, successful submit)
-  Verifying that previously editable mandatory fields are locked after submission

These steps ensure full coverage of the form SF424B Submission E2E workflow.

### Acceptance criteria

- Attempting to submit the application before completing the SF424B form triggers appropriate error messages about required fields and forms
- Saving the SF424B form with empty mandatory fields displays validation errors.
- The application history correctly logs:
  - Organization added
  - Failed submission attempt
  - SF424B form saved
  - Successful submission
- Previously editable fields in the SF424B form become read-only after submission.
- The E2E test runs successfully against both staging and localhost environments.
- Utility functions are updated or created accordingly.

Bhavna-Ramachandran:
PR #9038 for this task is under review.

Bhavna-Ramachandran:
PR #9038 for this task is has been reviewed and approved by @doug-s-nava. Hence marking this task as closed.

PR #9038, opened 2026-03-13
[Issue #9017] SF424B: Add submission failure and application history validation

Bhavna-Ramachandran:
## Summary

Work for : Task #9017 

## Changes proposed
Automate the below validation steps:
- Confirm error messages appear when attempting to submit without filling the required form
- Verify error messages for empty mandatory fields within the SF424B form
- Confirm the application history reflects all relevant events (organization added, failed submit, form saved, successful submit)
- Verify that previously editable mandatory fields are locked after submission

## Context for reviewers
- The structure allows standalone Playwright tests to run independently and in CI

## Validation steps
- Run the standalone Playwright test and it should complete the workflow per requirement in #9017

### DIFF
--- a/frontend/tests/e2e/apply/application-submission-validation.spec.ts
+++ b/frontend/tests/e2e/apply/application-submission-validation.spec.ts
@@ -27,20 +27,16 @@ import {
 import {
   SF424B_FORM_CONFIG,
   SF424B_FORM_MATCHER,
+  SF424B_REQUIRED_FIELD_ERRORS,
 } from "./fixtures/sf424b-field-definitions";
-import { sf424BHappyPathTestData } from "./fixtures/sf424b-fill-data";
+import {
+  sf424BHappyPathTestData,
+  sf424BReadonlyFields,
+} from "./fixtures/sf424b-fill-data";
 
 const { testOrgLabel, targetEnv } = playwrightEnv;
 const OPPORTUNITY_ID = "f7a1c2b3-4d5e-6789-8abc-1234567890ab"; // TEST-APPLY-ORG-IND-ON01
 const OPPORTUNITY_URL = `/opportunity/${OPPORTUNITY_ID}`;
-
-const sf424bErrors = [
-  { fieldId: "title", message: "Title is required" },
-  {
-    fieldId: "applicant_organization",
-    message: "Applicant Organization is required",
-  },
-];
 
 // Validation errors specific to required SF-424B and conditional SF-LLL form combination
 const EXPECTED_SUBMISSION_ERRORS = {
@@ -61,12 +57,6 @@ const EXPECTED_HISTORY_ENTRIES = [
   "Application submission failed", // 4
   "Organization Added", // 5
   "Application created", // 6
-];
-
-// SF-424B field values filled during the test, verified as read-only after submission
-const SF424B_READONLY_FIELDS = [
-  { fieldId: "title", expectedValue: "TESTER" },
-  { fieldId: "applicant_organization", expectedValue: testOrgLabel },
 ];
 
 // Skip non-Chrome browsers in staging
@@ -113,7 +103,11 @@ test(
     await saveForm(page, true); // expect validation errors
 
     // Checks error alert list at top of form page and inline field errors
-    await verifyFormStatusAfterSave(page, "incomplete", sf424bErrors);
+    await verifyFormStatusAfterSave(
+      page,
+      "incomplete",
+      SF424B_REQUIRED_FIELD_ERRORS,
+    );
 
     // On application page — verify form row shows "Some issues found"
     await verifyFormStatusOnApplication(
@@ -164,7 +158,7 @@ test(
       page,
       SF424B_FORM_MATCHER,
       "SF-424B",
-      SF424B_READONLY_FIELDS,
+      sf424BReadonlyFields(testOrgLabel),
     );
   },
 );
--- a/frontend/tests/e2e/apply/fixtures/sf424b-fill-data.ts
+++ b/frontend/tests/e2e/apply/fixtures/sf424b-fill-data.ts
@@ -1,4 +1,19 @@
-export const sf424BHappyPathTestData = (orgLabel: string) => ({
+import { ReadonlyFieldCheck } from "tests/e2e/utils/post-submission-utils";
+
+export const sf424BHappyPathTestData = (
+  orgLabel: string,
+): Record<string, string> => ({
   title: "TESTER",
   organization: orgLabel,
 });
+
+// Readonly field checks derived from fill data — fieldIds match testIds in sf424b-field-definitions.ts
+export const sf424BReadonlyFields = (
+  orgLabel: string,
+): ReadonlyFieldCheck[] => [
+  { fieldId: "title", expectedValue: sf424BHappyPathTestData(orgLabel).title },
+  {
+    fieldId: "applicant_organization",
+    expectedValue: sf424BHappyPathTestData(orgLabel).organization,
+  },
+];
--- a/frontend/tests/e2e/apply/failure-path-sf424b.spec.ts
+++ b/frontend/tests/e2e/apply/failure-path-sf424b.spec.ts
@@ -14,19 +14,14 @@ import {
   verifyFormStatusOnApplication,
 } from "tests/e2e/utils/forms/verify-form-status-utils";
 
-import { SF424B_FORM_MATCHER } from "./fixtures/sf424b-field-definitions";
+import {
+  SF424B_FORM_MATCHER,
+  SF424B_REQUIRED_FIELD_ERRORS,
+} from "./fixtures/sf424b-field-definitions";
 
 const { testOrgLabel, targetEnv } = playwrightEnv;
 const OPPORTUNITY_ID = "f7a1c2b3-4d5e-6789-8abc-1234567890ab"; // TEST-APPLY-ORG-IND-ON01
 const OPPORTUNITY_URL = `/opportunity/${OPPORTUNITY_ID}`;
-
-const sf424bErrors = [
-  { fieldId: "title", message: "Title is required" },
-  {
-    fieldId: "applicant_organization",
-    message: "Applicant Organization is required",
-  },
-];
 
 // Skip non-Chrome browsers in staging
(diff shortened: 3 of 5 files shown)


frontend/tests/e2e/utils/post-submission-utils.ts
```diff
@@ -0,0 +1,77 @@
...
+  await expect(page.locator("#main-content")).toContainText(
+    "Application History",
+  );
+
+  // Verify each history entry in order (index 0 = most recent)
+  for (let i = 0; i < expectedHistoryEntries.length; i++) {
+    const entryLocator = page.getByTestId(`responsive-data-${i}-1`);
+    await expect(entryLocator).toBeVisible();
+    await expect(entryLocator).toContainText(expectedHistoryEntries[i]);
+  }
```
aoysimmons-GH, line 48:
```suggestion
  }
  // Ensure there are no unexpected extra entries beyond the expected set.
  await expect(
    page.getByTestId(`responsive-data-${expectedHistoryEntries.length}-1`),
  ).toHaveCount(0);
```

frontend/tests/e2e/utils/post-submission-utils.ts
```diff
@@ -0,0 +1,77 @@
...
+  fieldId: string;
+  expectedValue: string;
+};
+
+/**
+ * Verifies the application status card shows "Submitted" and the
+ * Application History section is visible with the expected history entries.
+ *
+ * History entries are verified in order from most recent (index 0) to oldest
+ * (last index), matching the responsive-data-{index}-1 test IDs.
```
aoysimmons-GH, line 20:
```suggestion
 * (last index), matching the responsive-data-{index}-1 test IDs. * (last index), matching the responsive-data-{index}-1 test IDs, and verifies
 * no extra history rows are present.
```

frontend/tests/e2e/apply/happy-path-application-submission.spec.ts
```diff
@@ -70,7 +70,7 @@ test(
     );
 
     // Submit the application and verify success
-    await submitApplicationAndVerify(page);
+    await submitApplicationAndVerify(page, "success");
     // Application ID is now available in appId variable for further use if needed
```
aoysimmons-GH: // Application ID is now available in appId variable for further use if needed

I would remove this

frontend/tests/e2e/apply/application-submission-validation.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+      // Scrolls down and checks "Some issues found" in form row
+      await verifyFormStatusAfterSave(
+        page,
+        "incomplete",
+        "SF-424B",
+        applicationUrl,
+        sf424bErrors,
+      );
+    }
+
```
aoysimmons-GH: 89-104: I will do this


```suggestion
    const openedSf424bForValidation = await openForm(page, SF424B_FORM_MATCHER);
    if (!openedSf424bForValidation) {
      throw new Error(
        "Could not find or open SF-424B form link on the application forms page",
      );
    }

    // Do not enter anything and click save
    await saveForm(page, true); // expect validation errors

    // Checks error alert list at top of form page
    // Scrolls down and checks inline field errors on form page
    // Navigates to application landing page
    // Scrolls down and checks "Some issues found" in form row
    await verifyFormStatusAfterSave(
      page,
      "incomplete",
      "SF-424B",
      applicationUrl,
      sf424bErrors,
    );
```

frontend/tests/e2e/apply/happy-path-application-submission.spec.ts
Bhavna-Ramachandran: This is just a reminder FYI to say we have it extracted as part of this utils (gets validated if its generated), so in future we don't have to work on creating a new utils which does the same.

frontend/tests/e2e/utils/post-submission-utils.ts
Bhavna-Ramachandran: Acknowledged. I have made updates

frontend/tests/e2e/utils/post-submission-utils.ts
Bhavna-Ramachandran: Acknowledged. I have made updates

frontend/tests/e2e/apply/application-submission-validation.spec.ts
```diff
@@ -0,0 +1,170 @@
...
+    await submitApplicationAndVerify(
+      page,
+      EXPECTED_SUBMISSION_ERRORS.outcome,
+      EXPECTED_SUBMISSION_ERRORS.errors,
+    );
+
+    const applicationUrl = page.url();
+
+    const openedSf424bForValidation = await openForm(page, SF424B_FORM_MATCHER);
```
doug-s-nava, line 95: One thing we didn't quite resolve in my refactor - do we need all of the complexity of `openForm` vs the more simple `await page.getByRole("link", { name: formName }).click();` pattern used within `fillForm` on other tests?

We don't need another full refactor here but we should be clear going forward what the best way is to open a form. The only change I'd call for specifically at this point is using a hardcoded string for the form name rather than a regex

frontend/tests/e2e/apply/fixtures/sf424b-fill-data.ts
```diff
@@ -1,4 +1,4 @@
-export const sf424BHappyPathTestData = {
+export const sf424BHappyPathTestData = (orgLabel: string) => ({
   title: "TESTER",
-  organization: "Test org",
-};
+  organization: orgLabel,
```
doug-s-nava, line 7: why is this important?

frontend/tests/e2e/apply/application-submission-validation.spec.ts
```diff
@@ -0,0 +1,170 @@
...
+  "Form updated: Disclosure of Lobbying Activities (SF-LLL)", // 1
+  "Form updated: Assurances for Non-Construction Programs (SF-424B)", // 2
+  "Form updated: Assurances for Non-Construction Programs (SF-424B)", // 3
+  "Application submission failed", // 4
+  "Organization Added", // 5
+  "Application created", // 6
+];
+
+// SF-424B field values filled during the test, verified as read-only after submission
+const SF424B_READONLY_FIELDS = [
```
doug-s-nava: would love it if we could re-use the same fill data object for this, would just need to rethink the shared keys for the data and field definitions

frontend/tests/e2e/utils/authenticate-e2e-user-utils.ts
```diff
@@ -32,11 +32,10 @@ export async function authenticateE2eUser(
     await page.waitForTimeout(1000);
     await page.goto(baseUrl, { waitUntil: "domcontentloaded" });
 
-    const signInModal = page.locator('[role="dialog"], #sign-in-modal');
-    if (await signInModal.isVisible()) {
-      await selectLocalTestUser(page, testOrgLabel);
-      await page.waitForTimeout(2000);
-    }
+    // Always switch to the org user so apply flow tests have org associations.
```
doug-s-nava, line 35: interesting - we should be able to set up the spoofed user with org associations if they aren't set up that way already. We can dig into this more next quad

frontend/tests/e2e/utils/post-submission-utils.ts
```diff
@@ -0,0 +1,83 @@
+import { expect, Page } from "@playwright/test";
+import { openForm } from "tests/e2e/utils/forms/form-navigation-utils";
+
+/**
+ * Expected application history entries in display order (index 0 = most recent).
+ * Pass entries matching the actions taken in your specific test scenario.
+ */
+export type HistoryEntry = string;
```
doug-s-nava, line 8: what's the purpose of defining this type? Do we expect to narrow it further in the future?

frontend/tests/e2e/apply/application-submission-validation.spec.ts
Bhavna-Ramachandran: We’re using openForm here since this flow only needs to open the form and save it empty. fillForm is used where we actually fill fields with data.
 
We did initially start with the simple page.getByRole("link", { name: formName }).click() approach, but saw some flakiness locally (form link not immediately visible / interactable depending on scroll and rendering). That's the same reason we moved from hardcoded form name string to regex.

frontend/tests/e2e/apply/fixtures/sf424b-fill-data.ts
Bhavna-Ramachandran: In the current Grants.gov flow, the organization will be forward populated from core form which populates the value from the logged in user's UEI + if there are issues, it will be easy to check the logs as the org will match what is in the env. In future I have plans for test cases which is trying to submit with expired UEI Org.

frontend/tests/e2e/apply/application-submission-validation.spec.ts
Bhavna-Ramachandran: Acknowledged. Moved this to the filldata fixture.

frontend/tests/e2e/utils/post-submission-utils.ts
Bhavna-Ramachandran: The current requirement is to validate only the Activity text. However, I noticed there are other fields in that table. Anticipating that we may have to extend validation to those in future I defined it this way. Type makes it easier to adapt without changing how the function is being called.

frontend/tests/e2e/apply/fixtures/sf424b-fill-data.ts
doug-s-nava: ok. Not a blocker, but in the future let's not build testing functionality out for app functionality that doesn't exist yet. While this could support future functionality in the future, it also introduces complexity and confusion in the present

frontend/tests/e2e/apply/application-submission-validation.spec.ts
doug-s-nava: ok, please work with Aoy to make sure there is a shared understanding of what functionality to use when opening a form. Having both the form matcher approach and the hardcoded string approach is confusing. Regardless of whether a test uses fillForm or not, it should use the same functionality for opening a form. No need to adjust any previous tests if they are working correctly, but something to standardize in future tests

Issue #8601, opened 2026-02-19
Opportunity Summary CRUD Endpoint - Update


glenblosser-nava:
### Summary

As a grantor, I need to be able to update an Opportunity Summary. In order to do so, I must provide an opportunity ID and an opportunity summary ID. I cannot use the update endpoint to change a summary's type (forecast or non-forecast).

### Acceptance criteria

Update Endpoint functional as outlined in [the Phase 2 Tech Spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2613215233/Phase+2+Feature+Build+Out#Endpoints)

PR #9041, opened 2026-03-13
[Issue #8601] Opportunity Summary CRUD Endpoint - Update

dghazvini:
## Summary

Fixes for #8601 

## Changes proposed

Implementing the Update Opportunity Summary endpoint as laid out in the [Opportunity Publishing MVP Phase 2: Feature Build Out Tech Spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2613215233/Phase+2+Feature+Build+Out).

## Context for reviewers

Endpoint Path: `PUT /v1/grantors/opportunities/:opportunity_id/summary/:opportunity_summary_id` - Update an opportunity summary

Parameters: `opportunity_id:uuid`, `opportunity_summary_id:uuid`

Auth: A user must have the UPDATE_OPPORTUNITY privilege for the Agency associated with the opportunity.

Logic & Error Cases: 

Fetch the opportunity by ID and the summary by ID

404 if either don’t exist

Do the AuthZ check

403 if they don’t have the privilege

is_draft check

422 error if the is_draft field for the Opportunity is false; should not be able to update published opportunities.

Update all fields from the request body

Success: Return the updated opportunity summary and a 200 status code

Fail: Return a 500 status code
## Validation steps

 - [ ] 9 Unit Tests Added for expected responses/cases
 - [ ] Verified data persists in local DB
 - [ ] Verified expected responses on local testing


dghazvini:
Notes:


- Would appreciate a once-over on the Update request schema in particular - I copied over the required/allow_none fields mainly from the create update since my understanding is that the front end form will be pre-filled with existing data and sending a complete object when updating, but if that's not the case and I should update most fields to not be required, please call it out.
- Added `validate_opportunity_is_draft` and `validate_opportunity_created_in_simpler_grants` methods to the Create Opportunity Summary Endpoint as well as this new Update endpoint and verified tested 
- Separated `test_opportunity_summary.py` into `test_opportunity_summary_route_update.py` and `test_opportunity_summary_route_create.py`

api/src/api/opportunities_grantor_v1/opportunity_grantor_routes.py
```diff
@@ -149,3 +152,37 @@ def opportunity_summary_create(
             db_session, opportunity_id, json_data, user
         )
     return response.ApiResponse(message="Success", data=opportunity_summary)
+
+
+@opportunity_grantor_blueprint.put(
+    "/opportunities/<uuid:opportunity_id>/summary/<uuid:opportunity_summary_id>"
```
chouinar: All entities in URLs should be plural by convention.

```suggestion
    "/opportunities/<uuid:opportunity_id>/summaries/<uuid:opportunity_summary_id>"
```

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
```diff
@@ -525,3 +525,289 @@ class OpportunitySummaryDetailSchema(OpportunitySummaryV1Schema):
 
 class OpportunitySummaryCreateResponseV1Schema(AbstractResponseSchema):
     data = fields.Nested(OpportunitySummaryDetailSchema())
+
+
+class OpportunitySummaryUpdateRequestV1Schema(Schema):
```
chouinar: Is there anything about the PUT schema that would be different from the create schema? Only field I can think of that a user shouldn't be able to change is the `is_forecast` boolean. Could we use inheritance here to only define this once, and then have create have the additional fields?

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
```diff
@@ -525,3 +525,289 @@ class OpportunitySummaryDetailSchema(OpportunitySummaryV1Schema):
...
+    """Schema for PUT /v1/grantors/opportunities/<opportunity_id>/summary/<opportunity_summary_id> request"""
+
+    legacy_opportunity_id = fields.Integer(
+        required=False, allow_none=True, metadata={"description": "Legacy opportunity ID"}
+    )
```
chouinar: Legacy ID wouldn't be something a user would ever be allowed to set.

api/src/services/opportunities_grantor_v1/opportunity_summaries.py
```diff
@@ -25,6 +29,28 @@ def _check_existing_summary(opportunity: Opportunity, is_forecast: bool) -> None
...
+            break
+
+    if summary is None:
+        raise_flask_error(
+            404, f"Could not find Opportunity Summary with ID {opportunity_summary_id}"
+        )
+
+    # Ensure we're not trying to update is_forecast
+    if "is_forecast" in summary_data:
+        raise_flask_error(422, "Cannot update is_forecast field")
```
chouinar: Any fields not in the request schema wouldn't be added to the JSON - this shouldn't be something that could happen.

We should have a unit test that verifies that if I put `is_forecast` in the request, nothing happens to the field.

api/src/services/opportunities_grantor_v1/opportunity_summaries.py
```diff
@@ -51,3 +79,39 @@ def create_opportunity_summary(
...
+    validate_opportunity_is_draft(opportunity)
+    validate_opportunity_created_in_simpler_grants(opportunity)
+
+    # Get and validate the opportunity summary
+    summary = _get_opportunity_summary(opportunity, opportunity_summary_id, summary_data)
+
+    # Update all fields from the request body
+    for field, value in summary_data.items():
+        if hasattr(summary, field):
+            setattr(summary, field, value)
```
chouinar: What is the `hasattr` check guarding against? The request schema shouldn't have any fields in it that the DB model doesn't have. Any extra fields a user might have put in their request are filtered out by the schema validation that Marshmallow does.

api/tests/lib/opportunity_test_utils.py
```diff
@@ -158,3 +158,90 @@ def create_opportunity_summary_request(
...
+    agency_email_address_description="Updated email the agency",
+):
+    """Create a request body for updating an opportunity summary"""
+
+    # Set default dates if not provided
+    if post_date is None:
+        post_date = date.today()
+
+    if close_date is None:
+        close_date = date.today().replace(year=date.today().year + 1)
```
chouinar: Use timedelta's for doing math on dates.

```suggestion
        close_date = date.today() + timedelta(year=1)
```

api/tests/src/api/opportunities_grantors_v1/test_opportunity_summary_route_update.py
```diff
@@ -0,0 +1,360 @@
...
+    """Test successful update of an opportunity summary"""
+    opportunity, summary = opportunity_with_summary
+    _, _, token, _ = opportunity_summary_auth_data
+
+    # Modify the request to have clearly different values
+    opportunity_summary_update_request["summary_description"] = (
+        "UPDATED: This is a modified description"
+    )
+    opportunity_summary_update_request["award_floor"] = 30000
+    opportunity_summary_update_request["award_ceiling"] = 180000
```
chouinar: Why make `opportunity_summary_update_request` a fixture that calls a function that builds the request? Why not just call the function directly in the tests. You'd be able to do:

```py
opportunity_summary_update_request(summary_description="UPDATED: This is a modified description", award_floor=30000, award_ceiling=180000)`
```

api/tests/src/api/opportunities_grantors_v1/test_opportunity_summary_route_update.py
```diff
@@ -0,0 +1,360 @@
...
+    AgencyFactory,
+    OpportunityFactory,
+    OpportunitySummaryFactory,
+)
+
+
+@pytest.fixture
+def opportunity_summary_update_request():
+    """Return a valid opportunity summary update request"""
+    return create_opportunity_summary_update_request()
```
chouinar: Why is this a fixture? Why not call the function in each of the tests, especially if you want to pass values into this to set them.

api/tests/src/api/opportunities_grantors_v1/test_opportunity_summary_route_update.py
```diff
@@ -0,0 +1,360 @@
...
+
+@pytest.fixture
+def opportunity_summary_auth_data(db_session, enable_factory_create):
+    """Create a user with VIEW_OPPORTUNITY and UPDATE_OPPORTUNITY permissions and return auth data"""
+    agency = AgencyFactory.create()
+
+    user, agency, token, api_key_id = create_user_in_agency_with_jwt_and_api_key(
+        db_session=db_session,
+        privileges=[Privilege.VIEW_OPPORTUNITY, Privilege.UPDATE_OPPORTUNITY],
+    )
```
chouinar: The agency you create here isn't used as it's not passed into the function.

```suggestion
    user, agency, token, api_key_id = create_user_in_agency_with_jwt_and_api_key(
        db_session=db_session,
        privileges=[Privilege.VIEW_OPPORTUNITY, Privilege.UPDATE_OPPORTUNITY],
    )
```

api/tests/src/api/opportunities_grantors_v1/test_opportunity_summary_route_update.py
```diff
@@ -0,0 +1,360 @@
...
+
+    response = client.put(
+        f"/v1/grantors/opportunities/{opportunity.opportunity_id}/summary/{summary.opportunity_summary_id}",
+        json=opportunity_summary_update_request,
+        headers={"X-SGG-Token": token},
+    )
+
+    assert response.status_code == 403
+    response_json = response.get_json()
+    assert "forbidden" in str(response_json).lower()
```
chouinar: Please don't do any checks against `str(response_json).lower()`. These tests have to parse the responses / index into them.

This is a really roundabout way of doing: `assert response_json["message"] == "Forbidden"`.

api/src/api/opportunities_grantor_v1/opportunity_grantor_routes.py
dghazvini: Thanks for catching this - I updated the route name here and in the Create opportunity summary endpoint as well to reflect plural convention.

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
dghazvini: I created a `OpportunitySummaryBaseRequestSchema` and accompanying `OpportunitySummaryCreateRequestV1Schema` and `OpportunitySummaryUpdateRequestV1Schema` to inherit it. `OpportunitySummaryUpdateRequestV1Schema` not technically required but thought it makes sense for clarity/documentation - could remove if its better practice.

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
dghazvini: Removed from this schema.

api/src/services/opportunities_grantor_v1/opportunity_summaries.py
dghazvini: Removed and created a `test_opportunity_summary_update_cannot_change_forecast_status` unit test.

api/src/services/opportunities_grantor_v1/opportunity_summaries.py
dghazvini: Good point, wasn't taking into account the filtering the schema validation Marshmallow is already doing. Removed.

api/tests/lib/opportunity_test_utils.py
dghazvini: Updated to use timedelta - although timedelta doesn't take `year` as a parameter it seems, just set it to `days=365`.
Also I updated this method name from `create_opportunity_summary_update_request` to `update_opportunity_summary_request` to reduce confusion.

api/tests/src/api/opportunities_grantors_v1/test_opportunity_summary_route_update.py
dghazvini: Makes sense, inititally was trying to reduce code reuse but agree it doesn't really save anything using a fixture here. Updated tests to call `update_opportunity_summary_request()`

api/tests/src/api/opportunities_grantors_v1/test_opportunity_summary_route_update.py
dghazvini: Removed as part of above fix.

api/tests/src/api/opportunities_grantors_v1/test_opportunity_summary_route_update.py
dghazvini: Similar to the below issue- copied this over from previous iterations and I didn't catch it before; I went ahead and removed the unneeded agency creation call from all instances where I see this being used.

api/tests/src/api/opportunities_grantors_v1/test_opportunity_summary_route_update.py
dghazvini: Sorry, I copied this over from previous test iterations and overlooked this; I updated this as well as I went ahead and updated any previous cases in other test files of using `str(response_json)` to properly use `response_json["message"]`.

Issue #9044, opened 2026-03-13
Address Analytics Vulnerabilities


mdragon:
### Summary

There's a handful of minor package bumps that AWS ECR is flagging the Analytics container as needing. Unclear why it's so out of wack vs our CI/CD scans which are clean while there's 15 different CVEs (many against the same packages) flagged by ECR.

### Acceptance criteria

- ECR scans improve after merge/release.

PR #9045, opened 2026-03-13
[Issue #9044] Updates per fix versions noted by AWS ECR

mdragon:
## Summary

Work for #9045

## Changes proposed

Update several packages that ECR has flagged as having fixed vulnerabilities

## Context for reviewers

These were not getting flagged by our CI/CD scans, but were getting picked up by ECR scanning. 

Noting here and will comment in the PR at the appropriate spot. Despite many of these packages flagging as in our poetry lock in ECR, it looks like we weren't installing them directly previously. Not sure if that means we should ignore these for now rather than doing this approach.

## Validation steps

- ECR scans pass after merge to main and container builds.


analytics/pyproject.toml
```diff
@@ -26,9 +26,15 @@ boto3 = "^1.35.56"
...
-newrelic = "11.4.0"
+newrelic = "^11.5.0"
 requests = "^2.32.5"
 urllib3 = "^2.6.3"
+nltk = "^3.9.3"
+werkzeug = "^3.1.6"
+marshmallow = "^4.1.2"
+filelock = "^3.20.3"
+authlib = "^1.6.6"
+poetry = "^2.2.1"
```
mdragon: Did this trying to avoid the change in the poetry lock saying we used the older poetry than the prior version of the file. Guessing I just need to do this a different way.

analytics/pyproject.toml
```diff
@@ -26,9 +26,15 @@ boto3 = "^1.35.56"
 boto3-stubs = "^1.35.56"
 psycopg = { extras = ["binary"], version = "^3.2.3" }
 smart-open = "^7.0.5"
-newrelic = "11.4.0"
+newrelic = "^11.5.0"
 requests = "^2.32.5"
 urllib3 = "^2.6.3"
+nltk = "^3.9.3"
+werkzeug = "^3.1.6"
+marshmallow = "^4.1.2"
```
chouinar, line 34: This is a major version upgrade, I haven't made a ticket for it, but this isn't something I'd do without quite a bit of testing. They changed the error handling internals of Marshmallow so we likely need to / should rewrite the hacky logic I made to change the old way error handling worked.

analytics/pyproject.toml
```diff
@@ -26,9 +26,15 @@ boto3 = "^1.35.56"
...
+filelock = "^3.20.3"
+authlib = "^1.6.6"
```
mdragon, line 36: ECR flagged these as needing to be updated in the poetry.lock, but looks like we weren't directly installing them before. Curious if we think this is a good approach to these types of vulnerability results.

analytics/pyproject.toml
chouinar: We don't install poetry as a dependency - it's not one. If there's a poetry version update needed, that would need to change in the dockerfile.

analytics/pyproject.toml
```diff
@@ -26,9 +26,15 @@ boto3 = "^1.35.56"
 boto3-stubs = "^1.35.56"
 psycopg = { extras = ["binary"], version = "^3.2.3" }
 smart-open = "^7.0.5"
-newrelic = "11.4.0"
+newrelic = "^11.5.0"
 requests = "^2.32.5"
 urllib3 = "^2.6.3"
+nltk = "^3.9.3"
+werkzeug = "^3.1.6"
```
chouinar, line 33: This and a few others are not packages we depend on directly, I don't know if upgrading them without them being upgraded in the libraries we use is what we want to do as we might break assumptions in those libraries. Werkzeug for example is what Flask is heavily built on, I don't know if upgrading it would break Flask in fundamental ways that we wouldn't immediately recognize.

analytics/pyproject.toml
chouinar: Ah, wait, this is analytics. I'm less worried there - wait, what uses Marshmallow in analytics?

analytics/pyproject.toml
chouinar: What is using werkzeug in analytics?

analytics/pyproject.toml
mdragon: I think it's a sub-dependency it's it a minor upgrade for whatever package is bringing it in

analytics/pyproject.toml
chouinar: We should first upgrade our direct dependencies - https://github.com/HHS/simpler-grants-gov/pull/8881 - analytics has an open PR for that already. That might resolve at least some of these.

analytics/pyproject.toml
mdragon: These are all minor version upgrades compared to the version ECR is currently seeing.  I'm ok, not making these changes, I thought they were upgrades to stuff we were already using until I saw the diff after doing the ECR noted upgrades (but were actually installs). I'm wondering if we should build this and push it into ECR manually just to see if it changes the scans even?

Issue #8555, opened 2026-02-17
Prototype FieldList widget in form and print view


ErinPattisonNava:
### Summary

We want to do a basic prototype of of the FieldList widget to see what issues might occur.  The scope here is only the happy path, to make sure it is stored in the correctly.  We'll handle error cases and mandatory fields in a follow up ticket.

See tech spec here: https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2192179202/Tech+Spec+Expandable+Form+Field+Groups

### Acceptance criteria

- [ ] Create a dummy form to play around with.  This can be checked in, but should not be synced to any environment. 
- [ ] There is a basic prototype that allows us to create an list of text widgets.
    - [ ] It should have a plus that we can add additional rows.
    - [ ] It should have a delete that we can delete rows. 
    - [ ] Keep minimal investment in UI, only enough to make it functional.  We'll come back in later and design it. 
- [ ] The data should be stored correctly.
    - [ ] display label and description for the group
    - [ ] generate ids for each field based on their array position and base id
- [ ] Data needs to appear correctly in the print screen view.
- [ ] Unittests where it makes sense

ErinPattisonNava:
Breaking up https://github.com/HHS/simpler-grants-gov/issues/7026 into smaller parts

ErinPattisonNava:
Investigate if USWDS has a component/UI we could steal just to cut down further on UI time.

PR #9046, opened 2026-03-13
[Issue #8555] Prototype FieldList widget in form and print view

myduong-navapbc:
## Summary

Work for #8555 

This PR introduces a `FieldList` widget to support repeatable groups of fields within a form.

The widget allows users to dynamically `add` and `remove` rows of related inputs while maintaining compatibility with the existing form rendering system and data model.

A small local demo form was also added to allow development and QA of FieldList behavior.

### New FieldList widget

Added a new `FieldListWidget` that:

- renders repeatable rows of grouped fields
- allows users to add and delete rows
- manages local row state for interactive editing
- forwards changes through the optional onChange handler when available
- integrates with renderWidget so existing widgets can be reused inside rows

Helper utilities were added for:

- row normalization
- index placeholder replacement
- row add/delete operations
- safe value conversion for child widgets

### Form renderer updates

`FormFields.tsx` was updated to recognize `fieldList` nodes in the UI schema.

- introduced `isRenderableFieldNode` helper
- allowed `fieldList` nodes to pass validation checks

### getFieldConfig improvements

A helper (`buildFieldListBaseId`) was updated to generate stable base IDs for `FieldList` child widgets.

This introduces an index placeholder pattern:

`fieldListName[~~index~~]--fieldName`

The placeholder is replaced during rendering to generate row-specific field IDs.

This approach allows FieldList values to serialize correctly as arrays of objects.

### Types updates

New type definitions were added to support the FieldList widget:

- UiSchemaFieldList
- FieldListGroupItem
- FieldListWidgetProps

## Context for reviewers

Some forms require users to enter multiple instances of the same group of fields (for example, multiple contact people). The current form system supports individual fields and custom widgets but does not provide a generic mechanism for repeatable field groups.

This PR introduces a FieldList widget that:

- renders repeatable rows of fields
- allows users to add and remove rows dynamically
- serializes values as an array of objects
- integrates with the existing widget rendering pipeline
- supports print rendering

## Validation steps

3. navigate to `search`
4. search for `test`
5. you should see an opportunity called `Test Opportunity with ALL forms`
6. sign in
7. start new application
8. [VERIFY] click on all forms and verify that all forms render properly without failure
9. click `FieldList Demo Form - DO NOT SYNC PRODUCTION`
10. VERIFY the form loads
11. fill out the form, save
12. run `make generate-internal-token` from `api` directory
13. grab the token and place in request header "X-SGG-Internal-Token" and the token (I use a plugin called Modheader)
14. navigate to http://localhost:3000/print/application/APP_ID/form/FORM_ID
15. VERIFY the form data saves correctly

### DIFF
--- a/api/src/form_schema/forms/__init__.py
+++ b/api/src/form_schema/forms/__init__.py
@@ -36,5 +36,5 @@ def get_active_forms() -> list[Form]:
         SupplementaryNEHCoverSheet_v3_0,
         GG_LobbyingForm_v1_1,
         EPA_FORM_4700_4_v5_0,
-        EPA_KEY_CONTACT_v2_0
+        EPA_KEY_CONTACT_v2_0,
     ]


frontend/src/components/applyForm/widgets/MultipleAttachmentUploadWidget.tsx
```diff
@@ -69,7 +69,12 @@ const MultipleAttachmentUploadWidget = ({
 
     let parsedValue: string[] = [];
 
-    if (Array.isArray(initialValue)) {
```
myduong-navapbc, line 72: because initialValue can now be:

- string[]
- GeneralRecord[]

frontend/src/styles/_uswds-theme-custom-styles.scss
```diff
@@ -752,7 +752,28 @@ dl {
   display: none;
 }
```
myduong-navapbc, line 754: Basic styles were added to make the field lists look halfway decent.

The controls will be hidden in print view

frontend/src/components/applyForm/types.ts
```diff
@@ -77,23 +78,73 @@ export type WidgetTypes =
...
+ * onChange
+ *   Included for compatibility with the standard widget interface.
+ *   The current FieldList implementation manages row updates internally,
+ *   but this allows future versions to integrate with the normal form
+ *   change pipeline.
+ */
 export type FieldListWidgetProps = {
   id: string;
   key: string;
-
```
myduong-navapbc, line 83: moved this into the larger comment block above

frontend/src/components/applyForm/PrintForm.tsx
```diff
@@ -24,13 +24,15 @@ export default function PrintForm({
     <AttachmentsProvider
       value={{ attachments: attachments ?? [], setAttachmentsChanged }}
     >
-      <FormFields
-        errors={null}
-        formData={savedFormData}
-        schema={formSchema}
-        uiSchema={uiSchema}
-        formContext={{ rootFormData: savedFormData, rootSchema: formSchema }}
-      />
```
myduong-navapbc, line 33: Adds a wrapper to the form fields when we render for print to allow much better control

api/src/form_schema/forms/__init__.py
```diff
@@ -37,4 +38,5 @@ def get_active_forms() -> list[Form]:
         GG_LobbyingForm_v1_1,
         EPA_FORM_4700_4_v5_0,
         EPA_KEY_CONTACT_v2_0,
+        FIELDLIST_DEMO_FORM,
```
chouinar: Do we want this purely for local development, or would we want this in dev/staging/training?

If it's purely local dev, I'd remove it from this list, and in the seed script where the `get_active_forms` is called just add this form there.

api/src/form_schema/forms/do_not_sync_to_production__sandbox.py
```diff
@@ -0,0 +1,461 @@
...
+    omb_number="LOCAL-DEMO",
+    form_json_schema=FORM_JSON_SCHEMA,
+    form_ui_schema=FORM_UI_SCHEMA,
+    form_rule_schema=FORM_RULE_SCHEMA,
+    json_to_xml_schema=FORM_XML_TRANSFORM_RULES,
+    form_instruction_id=uuid.UUID("b8d44b2a-36c8-4e41-9f47-b3c5c7c7a9b0"),
+    form_type=FormType.SF424,  # placeholder until a generic demo type exists
+    sgg_version="1.0",
+    is_deprecated=False,
+)
```
chouinar: I'd suggest, even if just local, give this a form type. We don't use form type for anything yet, but it'll matter when we get to versioning/deprecating forms. Having it overlap with SF424 might cause issues, better to make a new form type to be safe.

```suggestion
FIELDLIST_DEMO_FORM = Form(
    form_id=uuid.UUID("d9d3e9a1-5b5e-4b36-a8b2-9c61c9d2b3b1"),
    legacy_form_id=None,
    form_name="FieldList Demo Form - DO NOT SYNC PRODUCTION",
    short_form_name="FIELDLIST_DEMO_DO_NOT_SYNC",
    form_version="1.0",
    agency_code="LOCAL",
    omb_number="LOCAL-DEMO",
    form_json_schema=FORM_JSON_SCHEMA,
    form_ui_schema=FORM_UI_SCHEMA,
    form_rule_schema=FORM_RULE_SCHEMA,
    json_to_xml_schema=FORM_XML_TRANSFORM_RULES,
    form_instruction_id=uuid.UUID("b8d44b2a-36c8-4e41-9f47-b3c5c7c7a9b0"),
    form_type=FormType.TEST_FORM,
    sgg_version="1.0",
    is_deprecated=False,
)
```

api/src/form_schema/forms/do_not_sync_to_production__sandbox.py
```diff
@@ -0,0 +1,461 @@
...
+        "form_name": "WIDGET_DEMO_FORM",
+        "namespaces": {},
+        "xml_structure": {
+            "root_element": "WIDGET_DEMO_FORM",
+            "version": "1.0",
+        },
+    },
+}
+
+FIELDLIST_DEMO_FORM = Form(
```
chouinar: Even if this is just a local-only test form, I might suggest that we add some basic unit tests for the form like our other forms - just to verify it's all configured as expected.

api/src/form_schema/forms/do_not_sync_to_production__sandbox.py
chouinar, line 1: Even if this form is different than other forms, we should still follow the same naming convention for files of naming the file and the form the same. If we want to separate this a bit from the other forms, perhaps we add a new directory to the forms folder like `development` and put it in there with a README explaining what that folder is for?

frontend/src/components/applyForm/widgets/FieldListWidget.tsx
```diff
@@ -1,19 +1,320 @@
...
+ *
+ * If no value exists yet, or if the current value contains fewer rows than
+ * `defaultSize`, this helper pads the result with empty row objects so the
+ * widget always renders the expected minimum number of rows.
+ */
+const normalizeFieldListRows = ({
+  value,
+  defaultSize,
+}: {
+  value: unknown;
```
doug-s-nava: can we type this a little more strongly? Looks like we can go with `GeneralRecord[] | undefined`

frontend/src/components/applyForm/widgets/FieldListWidget.tsx
```diff
@@ -1,19 +1,320 @@
...
+                id: generatedId,
+                value: currentValue,
+                onChange: handleChildChange,
+                isFormLocked,
+                formContext,
+              };
+
+              return (
+                <React.Fragment key={generatedId}>
+                  {renderWidget({
```
doug-s-nava: renderWidget will honor a key prop - seems like you could pass that in rather than wrapping in a fragment here

frontend/src/components/applyForm/widgets/FieldListWidget.tsx
```diff
@@ -1,19 +1,320 @@
...
+              <button
+                type="button"
+                onClick={() => handleDeleteRow(rowIndex)}
+                disabled={isInteractionDisabled}
+              >
+                Delete
+              </button>
+            </div>
+
+            {groupDefinition.map((groupItem: FieldListGroupItem) => {
```
doug-s-nava: might be more performant to pull this logic out into its own useEffects / useCallbacks rather than running on each render, but not important for this iteration of the feature. We can come back to it

frontend/src/components/applyForm/widgets/FieldListWidget.tsx
```diff
@@ -1,19 +1,320 @@
...
+  }
+
+  if (typeof value === "object" && value !== null && !Array.isArray(value)) {
+    return value as GeneralRecord;
+  }
+
+  return undefined;
+};
 
 function FieldListWidget(widgetProps: FieldListWidgetProps): JSX.Element {
```
doug-s-nava: don't think we need to specify the return type here

frontend/src/components/applyForm/widgets/FieldListWidget.tsx
```diff
@@ -1,19 +1,320 @@
...
-      </p>
+
+      {rows.map((rowValue, rowIndex) => {
+        return (
+          <div
+            key={`${id}--row-${rowIndex}`}
+            className="field-list-widget__row"
+          >
+            <div className="field-list-widget__controls">
+              <strong>Row {rowIndex + 1}</strong>
```
doug-s-nava: are we aligned on calling these "Rows"? Visually they probably won't really look like rows, even if the concept is solid. I wonder if something like "entry" or "item" would make more sense to users?

frontend/src/components/applyForm/widgets/FieldListWidget.tsx
```diff
@@ -1,19 +1,320 @@
...
+              <button
+                type="button"
+                onClick={() => handleDeleteRow(rowIndex)}
+                disabled={isInteractionDisabled}
+              >
+                Delete
```
doug-s-nava: can we put these through translation?

frontend/src/components/applyForm/widgets/FieldListWidget.tsx
```diff
@@ -1,19 +1,320 @@
...
+
+              const currentValue = toBroadlyDefinedWidgetValue(
+                rowValue[storageKey],
+              );
+
+              /**
+               * Updates a single field within the current row while preserving
+               * the rest of the row and all sibling rows.
+               */
+              const handleChildChange = (nextValue: unknown): void => {
```
doug-s-nava: is the purpose of this largely to pass on new values to the onChange handler? My hope would be that any change to the values on save would be reflected in the dom directly so we shouldn't need to track the state of individual widgets outside of custom behavior from the onChange

frontend/src/components/applyForm/widgets/MultipleAttachmentUploadWidget.tsx
doug-s-nava: good catch - do we want to error if this somehow gets the wrong type?

frontend/src/components/applyForm/FormFields.tsx
```diff
@@ -13,6 +18,47 @@ type RootBudgetFormContext = {
...
+ * This allows downstream code to safely access properties such as
+ * `definition` or `schema` without additional type assertions.
+ *
+ * Extra context:
+ *
+ * - `fieldList` nodes are always renderable
+ * - `field` nodes contain either `definition` or `schema`
+ * - `section` nodes contain `children` instead and are excluded
+ */
+const isRenderableFieldNode = (
```
doug-s-nava, line 56: 👍

frontend/src/components/applyForm/FormFields.tsx
```diff
@@ -36,8 +82,8 @@ export const FormFields = ({
 
   try {
     requiredFieldPaths = getRequiredProperties(schema);
-  } catch (e) {
-    console.error(e);
+  } catch (error: unknown) {
```
doug-s-nava: let's not forget our code style conventions here https://github.com/HHS/simpler-grants-gov/blob/main/documentation/frontend/code-style.md#errors

frontend/src/components/applyForm/FormFields.tsx
```diff
@@ -52,23 +98,34 @@ export const FormFields = ({
...
-      } else if (!("definition" in node || "schema" in node)) {
+      } else if (!isRenderableFieldNode(node)) {
         throw new Error("child field missing definition and schema");
       } else if (!parent) {
-        // treat as valid non-child field
-        const requiredField = isFieldRequired(
-          (node.definition || node.schema.title || "") as string,
-          requiredFieldPaths,
-        );
+        // FieldList is a renderable composite widget and does not have its own
```
doug-s-nava, line 119: one thing we haven't thought about is that we'll probably need to support a minimum size for a fieldList, as a way of backing into the idea of a required field. Not a requirement right now, but we should maybe ticket that for later since I bet it will come up eventually

frontend/src/styles/_uswds-theme-custom-styles.scss
doug-s-nava: any reason we can't use uswds classes for this?

frontend/src/utils/applyForm/getFieldConfig.ts
```diff
@@ -68,14 +71,10 @@ export function buildFieldListBaseId({
...
-    return childId.replace(
-      token,
-      `--${fieldListName}[${FIELD_LIST_INDEX_TOKEN}]--`,
-    );
-  }
-  return `${fieldListName}[${FIELD_LIST_INDEX_TOKEN}]--${childId}`;
+  const childIdParts = childId.split("--");
+  const finalFieldKey = childIdParts[childIdParts.length - 1];
+
+  return `${fieldListName}[${FIELD_LIST_INDEX_TOKEN}]--${finalFieldKey}`;
```
doug-s-nava, line 94: I'm a little worried this will bring the saved data json out of sync with the UI schema if we remove some parts of the field path from the id. How hard would it be to preserve the full path?

frontend/src/utils/applyForm/getFieldConfig.ts
```diff
@@ -471,6 +475,7 @@ export const getFieldConfig = <V extends string | Record<string, unknown>>({
         groupDefinition,
         rawErrors: [],
         requiredFields: [],
+        value: fieldListValue as GeneralRecord[] | undefined,
```
doug-s-nava: this is nitpicky, but can we move this logic (419 - 481) out to its own function, in order to keep the pattern intact where ach specific type of field handling (basic, multifield, field list) gets its own function (see getBasicFieldInfo, getBasicMultifieldInfo)

api/src/form_schema/forms/__init__.py
myduong-navapbc: ~~Awesome! Will move it to the seed script. this is dev only.~~

This is for a prototype only, sorry for tagging you in the PR. I learned that we will not have the form in the PR at all. So, there's no need for now. but this is very useful, thank you for the info.

frontend/src/components/applyForm/FormFields.tsx
myduong-navapbc: I totally forgot, ty for the reminder

frontend/src/components/applyForm/FormFields.tsx
myduong-navapbc: @chris-kuryak has an upcoming ticket that has the min/max size

frontend/src/components/applyForm/widgets/FieldListWidget.tsx
myduong-navapbc: This one was pretty gnarly, I wrote down a bunch of bullet points when I was testing this: 

I originally tried to rely on the standard widget onChange flow, but during implementation/debugging I found that this form architecture does not currently provide a usable onChange handler for FieldList. That works for standard widgets, but FieldList is a composite widget that also owns entry add/remove behavior. Without local state, add/delete operations computed the next value correctly but had no upstream state change to re-render from, so the UI did not update. Maybe I could use some help here.

Anywho, that left me with just the option to let FieldList manage its entry state locally and rehydrates from the saved value prop. This keeps dynamic entry behavior responsive while remaining aligned with the existing save/hydrate form architecture.

TL;DR:
I tried to rely on the normal onChange flow, but the prop was not consistently propagated through the widget rendering pipeline, so we added local state to keep the UI responsive while still forwarding changes when onChange exists

frontend/src/components/applyForm/widgets/MultipleAttachmentUploadWidget.tsx
myduong-navapbc: I opted to log a warning rather than throw an error here. Since this is a user-facing form widget, throwing could break the entire form render. Let me know if you feel differently here, I can totally throw an error instead

frontend/src/styles/_uswds-theme-custom-styles.scss
myduong-navapbc: not at all, will update to use uswds classes

api/src/form_schema/forms/do_not_sync_to_production__sandbox.py
myduong-navapbc: im going to mark these as resolved because this form will not make it into the codebase at the end

frontend/src/components/applyForm/widgets/FieldListWidget.tsx
myduong-navapbc: Great catch! done

frontend/src/components/applyForm/widgets/FieldListWidget.tsx
myduong-navapbc: Entry it is, at least for now

frontend/src/components/applyForm/widgets/FieldListWidget.tsx
myduong-navapbc: done! added the usecallback, was a bit bigger change than i thought

frontend/src/utils/applyForm/getFieldConfig.ts
myduong-navapbc: Not difficult at all- 
i updated it to include the full path- and the tests as well

frontend/src/utils/applyForm/getFieldConfig.ts
myduong-navapbc: I nit this too, originally. - done -

frontend/src/components/applyForm/widgets/FieldListWidget.tsx
doug-s-nava: interesting. I'll play around with it a bit this morning

doug-s-nava: I think the issue here may be that because we are tracking changes to the number of rows in local client side state, we're opting the whole child tree into a dependency on local state. However, I'm not able to fully test because there seems to be a bug around the input naming. Ex. I get `contact_people_test--items--contact_people_test[0]--prefix` when I should get `contact_people_test[0]--prefix`. This is causing the form not to properly update on save

myduong-navapbc: fixed the ID issue

myduong-navapbc:
> ✅ ✅ ✅ ✅
> 
> this looks good to me, nice job. One thing to clarify - do we need to test out what this would look like with different types of widgets as children, or is the use case we're supporting at the moment just text fields? I think this would work for any type of field but it might make sense to test that hypothesis by switching out some input types on the test form if that's something we want to try and support now

I totally understand the concern! We will add those other input types in a future version.  Including, error handling, required fields, attachments, and test out the multifield (dun dun dun).

I think the strategy here is to take this feature in tiny bitsized chunks.

Issue #9049, opened 2026-03-16
Investigate and Fix Frontend dev environment url redirect issue


prasnava:
Investigate and fix the url redirect issue in the frontend dev environment.

PR #9050, opened 2026-03-16
[Issue #9049]: Fix for frontend url redirect issue dev environment

prasnava:
## Summary

Updated the cloudfront config to https only to fix the url redirect issue in the dev environment. 

We didn't notice this issue in other environment's because only the dev env. has cloudfront enabled. The root case of the issue is related to the platform upgrade merge which introduced a new ALB listener rule http_to_https_redirect. 

## Validation steps

Deployed to the dev environment: [here](https://github.com/HHS/simpler-grants-gov/actions/runs/23148151573)


chouinar:
@prasnava - is your change currently in dev? Still seeing the redirects.

prasnava:
@chouinar yes, it's still in dev. Have you tried incognito window? I'm able to load the page both in regular browser mode and incognito. 
Screenshot 2026-03-16 at 12 22 43 PM

chouinar:
> @chouinar yes, it's still in dev. Have you tried incognito window? I'm able to load the page both in regular browser mode and incognito. 

It seems to behave fine in incognito, but getting redirected to a cloudfront URL otherwise - guessing that's the CDN cache? Is that something that'll uncache quickly?

prasnava:
It could be the browser cache ttl, I cleared the cloudfront cache a couple of times.

PR #9056, opened 2026-03-16
[Issue #8972] Create db tables for organization saved opportunity notifications

### LINKED ISSUE #8972: [BE] Create db tables for organization saved opportunity notifications
### Summary

Create the following tables to store organization saved opportunities notification settings

`user_saved_opportunities_notifications` with the following columns:

id - uuid pk (auto-generated uuid)
user_id - uuid fk
organization_id - uuid fk (optional -- if not there, assume self)
email_enabled - bool
created_at - timestamp
updated_at - timestamp

Set a unique index on user_id and organization_id

When processing this table, if there is no entry with organization_id is null (e.g. a setting for self) we assume a default of email_enabled true.

For each organization the user_id is part of, if there is no entry with the same organization_id (e.g. the setting for that organization) we assume a default of email_enabled false.

Upon POST, if the entry isn't present we add it, and if the entry is present we updated email_enabled to the value sent in and updated_at timestamp to the current time.

Create the appropriate ORM models associated for this table as well.

### Acceptance criteria

- [ ] Migrations ran and tables and indexes exist in all environments
- [ ] Seed data is created for testing
- [ ] ORM models exist

mikehgrantsgov:
## Summary

Fixes #8972

## Changes proposed

Add `user_saved_opportunity_notification` model with `user_id`,`organization_id` (nullable), and `email_enabled`
Add migration  with partial unique indexes and cascade deletes
Add `UserSavedOpportunityNotificationFactory`

## Context for reviewers

`organization_id` being nullable means we have two row types: a NULL org means the user's own default settings; a non-null org means settings scoped to that organization. Because PostgreSQL treats NULLs as distinct in unique constraints, two partial indexes enforce uniqueness instead of a standard UniqueConstraint. Both FKs use ON DELETE CASCADE so notification rows are cleaned up automatically when a user or organization is deleted.

## Validation steps

Check `make db-migrate` to see migration applied

### DIFF
--- a/api/src/db/migrations/versions/2026_03_12_create_application_submission_retrieved_.py
+++ b/api/src/db/migrations/versions/2026_03_12_create_application_submission_retrieved_.py
@@ -11,7 +11,7 @@ from alembic import op
 
 # revision identifiers, used by Alembic.
 revision = "f8eaafc955ec"
-down_revision = "b4c9d2e8f1a3"
+down_revision = "3bbb6f5e037e"
 branch_labels = None
 depends_on = None
 
--- a/api/src/db/migrations/versions/2026_03_16_add_user_saved_opportunity_notification.py
+++ b/api/src/db/migrations/versions/2026_03_16_add_user_saved_opportunity_notification.py
@@ -1,8 +1,8 @@
 """add user_saved_opportunity_notification table
 
-Revision ID: b4c9d2e8f1a3
-Revises: 3bbb6f5e037e
-Create Date: 2026-03-13 00:00:00.000000
+Revision ID: a3f7e2c9d1b8
+Revises: f8eaafc955ec
+Create Date: 2026-03-16 00:00:00.000000
 
 """
 
@@ -10,8 +10,8 @@ import sqlalchemy as sa
 from alembic import op
 
 # revision identifiers, used by Alembic.
-revision = "b4c9d2e8f1a3"
-down_revision = "3bbb6f5e037e"
+revision = "a3f7e2c9d1b8"
+down_revision = "f8eaafc955ec"
 branch_labels = None
 depends_on = None
 
@@ -49,7 +49,7 @@ def upgrade():
         sa.UniqueConstraint(
             "user_id",
             "organization_id",
-            name=op.f("user_saved_opportunity_notification_user_id_organization_id_key"),
+            name=op.f("user_saved_opportunity_notification_user_id_uniq"),
             postgresql_nulls_not_distinct=True,
         ),
         sa.PrimaryKeyConstraint(


api/src/db/models/user_models.py
```diff
@@ -162,6 +169,58 @@ class UserSavedOpportunity(ApiSchemaTable, TimestampMixin):
...
+        # When organization_id IS NOT NULL, one row per (user, org) pair.
+        Index(
+            "user_saved_opportunity_notification_user_id_org_id_idx",
+            "user_id",
+            "organization_id",
+            unique=True,
+            postgresql_where="organization_id IS NOT NULL",
+        ),
+        ApiSchemaTable.__table_args__,
+    )
```
chouinar: Does this need to be two constraints - would having a single index without the where work?

You can use the `NULLS_NOT_DISTINCT` flag to change how null columns behave in constraints https://www.postgresql.org/docs/current/indexes-unique.html - and that can be set I believe by passing `postgresql_nulls_not_distinct` into a unique constraint.

api/src/db/models/user_models.py
```diff
@@ -162,6 +169,58 @@ class UserSavedOpportunity(ApiSchemaTable, TimestampMixin):
...
+        UUID, ForeignKey(User.user_id, ondelete="CASCADE"), index=True
+    )
+    user: Mapped[User] = relationship(User, back_populates="saved_opportunity_notifications")
+
+    organization_id: Mapped[uuid.UUID | None] = mapped_column(
+        UUID,
+        ForeignKey(Organization.organization_id, ondelete="CASCADE"),
+        index=True,
+        nullable=True,
+    )
```
chouinar: Why did we add a cascade for these? If you delete this row, the user/org table should not be affected.

Also don't put `nullable` - just use the python type, don't define it twice.

```suggestion
    user_id: Mapped[uuid.UUID] = mapped_column(
        UUID, ForeignKey(User.user_id), index=True
    )
    user: Mapped[User] = relationship(User, back_populates="saved_opportunity_notifications")

    organization_id: Mapped[uuid.UUID | None] = mapped_column(
        UUID,
        ForeignKey(Organization.organization_id),
        index=True,
    )
```

api/src/db/models/user_models.py
```diff
@@ -162,6 +169,58 @@ class UserSavedOpportunity(ApiSchemaTable, TimestampMixin):
...
 
+class UserSavedOpportunityNotification(ApiSchemaTable, TimestampMixin):
+    """Notification settings for a user's saved opportunities.
+
+    A row with organization_id=None represents the user's own default settings.
+    A row with an organization_id represents settings scoped to that organization.
+
+    Business logic defaults:
+    - No row for user's own settings -> email_enabled=True
+    - No row for an org -> email_enabled=False
```
chouinar, line 180: If this is the case, I'd err on the side of not specifying the default - you might be able to set the value based on whether the organization_id is set - like what we do for `updated_at` being the same as `created_at`

api/src/db/models/user_models.py
mikehgrantsgov: Updated this

api/src/db/models/user_models.py
mikehgrantsgov: Replaced both partial indexes with a single UniqueConstraint here. Added `test_user_saved_opportunity_notification.py` to confirm the behavior.

api/src/db/models/user_models.py
mikehgrantsgov: Makes sense. email_enabled now uses a context-sensitive default: a lambda that reads organization_id at insert time and sets `email_enabled=True` if it's `None` and `False` otherwise.

Issue #8998, opened 2026-03-11
SOAP/Proxy:  Add verify_access check to the GetSubmissionListExpanded


jakobpederson:
### Summary

Currently we do not check for permissions when hitting the GetSubmissionListExpanded. Add the check there so we verify the user has the correct permissions.

### Acceptance criteria

- [ ] user will need the correct permissions to access data on that endpoint

PR #9062, opened 2026-03-16
[Issue #8998] Add privilege check on GetSubmissionListExpanded

jakobpederson:
## Summary

Fixes / Work for #8998  

## Changes proposed

- Added a verify privilege check for cert user in the `GetSubmissionListExpanded` service
- Updated the privilege on the `GetSubmisisonList` from `LEGACY_AGENCY_GRANT_RETRIEVER` to `LEGACY_AGENCY_VIEWER`
- Change `Use-Simpler-Override` from "true" to "1"

## Context for reviewers

- The current setup does not check the privilege of the cert use against the agency, and we wanted to check that to make sure correct users can access the submissions correctly
- Updated the privileges to be in line with comments about what privileges should actually be
- Changed the "true" to "1" in order to be consistent with the `Use-Soap-Jwt` header

## Validation steps

- [ ] unittest pass
- [ ] Sending request to endpoint fields fail if the privilege `LEGACY_AGENCY_VIEWER` is not on the user

### DIFF
--- a/api/src/legacy_soap_api/legacy_soap_api_auth.py
+++ b/api/src/legacy_soap_api/legacy_soap_api_auth.py
@@ -5,7 +5,6 @@ from typing import Any
 from urllib.parse import unquote
 
 import jwt
-from apiflask.exceptions import HTTPError
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import hashes
 from cryptography.x509 import load_pem_x509_certificate
@@ -14,7 +13,7 @@ from requests.adapters import HTTPAdapter
 from sqlalchemy import func, select
 
 import src.adapters.db as db
-from src.auth.endpoint_access_util import verify_access
+from src.auth.endpoint_access_util import can_access
 from src.db.models.agency_models import Agency
 from src.db.models.user_models import LegacyCertificate
 from src.legacy_soap_api.legacy_soap_api_config import SimplerSoapAPI, SOAPOperationConfig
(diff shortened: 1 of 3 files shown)


api/src/legacy_soap_api/legacy_soap_api_config.py
```diff
@@ -146,7 +146,7 @@ class SOAPOperationConfig:
         "GetSubmissionListExpandedRequest": SOAPOperationConfig(
             request_operation_name="GetSubmissionListExpandedRequest",
             response_operation_name="GetSubmissionListExpandedResponse",
-            privileges={Privilege.LEGACY_AGENCY_GRANT_RETRIEVER},
+            privileges={Privilege.LEGACY_AGENCY_VIEWER},
```
jakobpederson, line 149: Changing this as per comments above:
```
    # These are the privileges needed for these endpoints:
    # GetSubmissionListExpandedRequest = {Privilege.LEGACY_AGENCY_VIEWER}
    # GetApplicationRequest = {Privilege.LEGACY_AGENCY_GRANT_RETRIEVER}
    # GetApplicationZipRequest = {Privilege.LEGACY_AGENCY_GRANT_RETRIEVER}
    # ConfirmApplicationDeliveryRequest = {Privilege.LEGACY_AGENCY_GRANT_RETRIEVER}
    # UpdateApplicationInfoReqest = {Privilege.LEGACY_AGENCY_ASSIGNER}
```

api/src/legacy_soap_api/legacy_soap_api_auth.py
```diff
@@ -207,3 +210,27 @@ def generate_soap_jwt(
...
+                "User did not have permission to access this application",
+                extra={
+                    "user_id": certificate.user.user_id,
+                    "agency_id": agency.agency_id if agency else None,
+                    "privileges": soap_config.privileges,
+                },
+            )
+            raise SOAPClientUserDoesNotHavePermission(
+                "User did not have permission to access this application"
+            ) from e
```
chouinar: Rather than catch the exception, use the non-exception version of the function.

```suggestion
     if not can_access(
          certificate.user,
          soap_config.privileges,
          agency,
      )
      logger.info(
          "User did not have permission to access this application",
          extra={
              "user_id": certificate.user.user_id,
              "agency_id": agency.agency_id if agency else None,
              "privileges": soap_config.privileges,
          },
      )
      raise SOAPClientUserDoesNotHavePermission(
          "User did not have permission to access this application"
      )
```

api/src/legacy_soap_api/legacy_soap_api_auth.py
```diff
@@ -207,3 +210,27 @@ def generate_soap_jwt(
         "certId": cert_id,
     }
     return jwt.encode(payload=payload, key=soap_partner_gateway_auth_key, algorithm="HS256")
+
+
+def verify_certificate_access(
+    certificate: LegacyCertificate, soap_config: SOAPOperationConfig, agency: Agency | None
```
chouinar, line 215: If agency is None, this should just error.

I know it can technically be none because of the way the relationship for opportunities works - but that happening should outright block everything.

api/src/legacy_soap_api/legacy_soap_api_auth.py
```diff
@@ -207,3 +210,27 @@ def generate_soap_jwt(
...
+) -> None:
+    if soap_config.privileges is not None:
```
chouinar: If privileges is None - this should probably error. I don't think there's any endpoint on the grantor side that doesn't require privileges.

Issue #6669, opened 2025-10-08
Updating the data and application status for ConfirmApplicationDelivery


chris-kuryak:
### Summary

Updating the data and application status. Currently blocked by #6668 

### Acceptance criteria

- [ ] Updates the status of a submission to "Received by Agency" for a given Grants Gov Tracking Number
- [ ] A submission must have a status of "Validated" for this service to be executed

ychoy:
Needs to be done at same time as #6668, necessary for testing

kkrug:
Noting that this ticket will be reconfigured to account for the new approach (see #8957)

freddieyebra:
I am working on the documentation AC for #8850 which does not block this ticket. So removing it as blocker for now  @kkrug @mdragon

mdragon:
> Noting that this ticket will be reconfigured to account for the new approach (see [#8957](https://github.com/HHS/simpler-grants-gov/issues/8957))

@kkrug I updated the ACs to account for #8957, let me know if you have any questions.

PR #9064, opened 2026-03-16
[Issue #6669] Checking and updating the data for ConfirmApplicationDelivery

kkrug:
## Summary

Fixes for #6669  

## Changes proposed

- Created confirm_application_delivery_response.py to create the new service `confirm_application_delivery_response`
- Updated __init__.py to add the new service
- Updated legacy_soap_api_config.py to add `always_call_simpler=True` to `ConfirmApplicationDeliveryRequest`
- Updated legacy_soap_client.py to add `confirm_application_delivery_request` to dispatch the new service
- Created test_confirm_application_delivery_response.py to test new service
- Updated competition_models.py to add `cascade="all, delete-orphan"` to `application_submission_retrievals`

## Context for reviewers

Some notes on this:

- Only `ACCEPTED` status is valid, `SUBMITTED` maps to "Received" in grants.gov, not "Validated"
- The duplicate check is scoped per-user, meaning different users can each retrieve the same submission, but the same user calling twice gets the fault response
- Schemas and operation config for `ConfirmApplicationDelivery` already existed, this PR adds the service logic

## Validation steps

Confirm tests pass

### DIFF
--- a/api/src/legacy_soap_api/grantors/fault_messages.py
+++ b/api/src/legacy_soap_api/grantors/fault_messages.py
@@ -11,16 +11,16 @@ MissingGrantsGovTrackingNumber = FaultMessage(
 )
 
 ConfirmDeliverySubmissionNotFound = FaultMessage(
-    faultstring="Failed to confirm application delivery.(Authorization Failure)",
+    faultstring="Unable to find application from tracking number. Failed to confirm application delivery.",
     faultcode="soap:Server",
 )
 
 ConfirmDeliveryInvalidStatus = FaultMessage(
-    faultstring="Failed to confirm application delivery.(Authorization Failure)",
+    faultstring="Invalid application status. Expected Accepted status. Failed to confirm application delivery.",
     faultcode="soap:Server",
 )
 
 ConfirmDeliveryAlreadyRetrieved = FaultMessage(
-    faultstring="Failed to confirm application delivery.(Authorization Failure)",
+    faultstring="This application submission has already been retrieved. Failed to confirm application delivery.",
     faultcode="soap:Server",
 )


api/src/db/models/competition_models.py
```diff
@@ -467,6 +467,7 @@ class ApplicationSubmission(ApiSchemaTable, TimestampMixin):
         "ApplicationSubmissionRetrieved",
         back_populates="application_submission",
         uselist=True,
+        cascade="all, delete-orphan",
```
kkrug, line 470: The other changes were causing issues when the full test suite ran (seemed like some test pollution) so this change fixed that.

api/src/legacy_soap_api/grantors/services/confirm_application_delivery_response.py
```diff
@@ -0,0 +1,135 @@
...
+    soap_request: SOAPRequest,
+    confirm_application_delivery_request: grantor_schemas.ConfirmApplicationDeliveryRequest,
+    soap_config: SOAPOperationConfig,
+) -> grantor_schemas.ConfirmApplicationDeliveryResponseSOAPEnvelope:
+    legacy_tracking_number = confirm_application_delivery_request.grants_gov_tracking_number
+
+    if not legacy_tracking_number:
+        return grantor_schemas.ConfirmApplicationDeliveryResponseSOAPEnvelope(
+            grants_gov_tracking_number=legacy_tracking_number,
+        )
```
freddieyebra: We can omit this since we validate and raise error if the tracking number is missing/invalid in the `GrantsGovTrackingNumberRequiredSchema` schema.

api/src/legacy_soap_api/grantors/services/confirm_application_delivery_response.py
```diff
@@ -0,0 +1,135 @@
...
+
+    if legacy_tracking_number.startswith("GRANT"):
+        legacy_tracking_number = legacy_tracking_number.split("GRANT")[1]
+
+    application_submission = db_session.execute(
+        select(ApplicationSubmission).where(
+            ApplicationSubmission.legacy_tracking_number == int(legacy_tracking_number),
+        )
+    ).scalar()
```
freddieyebra: Can we put this into a service file method? `get_application_submission_by_legacy_tracking_number()`. I also do not see any unique constraint on the legacy_tracking_number column. Can we use `.scalar_one_or_none()` instead of `.scalar()` here. that way we will not silently ignore the case of multiple rows returned for the same tracking number.

api/src/legacy_soap_api/grantors/services/confirm_application_delivery_response.py
```diff
@@ -0,0 +1,135 @@
...
+        logger.info(
+            f"Unable to find submission legacy_tracking_number {legacy_tracking_number}.",
+            extra={
+                "soap_api_event": LegacySoapApiEvent.ERROR_CALLING_SIMPLER,
+                "response_operation_name": "ConfirmApplicationDeliveryResponse",
+            },
+        )
+        return grantor_schemas.ConfirmApplicationDeliveryResponseSOAPEnvelope(
+            grants_gov_tracking_number=confirm_application_delivery_request.grants_gov_tracking_number,
+        )
```
freddieyebra: We should return None or raise an exception here. The legacy api returns a `<Fault>` for unsuccessful requests for this operation. You can handle/raise the exceptions wherever makes most sense but the errors should return a fault message instead of a operation response envelope. you can see [this class](https://github.com/HHS/simpler-grants-gov/blob/d42ccabcb5e1f238911d11426a6978e2ce4e89e3/api/src/legacy_soap_api/legacy_soap_api_schemas.py#L138) for example of fault. Same for the other error cases.

api/src/legacy_soap_api/grantors/services/confirm_application_delivery_response.py
```diff
@@ -0,0 +1,135 @@
...
+        return grantor_schemas.ConfirmApplicationDeliveryResponseSOAPEnvelope(
+            grants_gov_tracking_number=confirm_application_delivery_request.grants_gov_tracking_number,
+            response_message=CONFIRM_DELIVERY_FAULT_STRING,
+        )
+
+    certificate = validate_certificate(
+        db_session, soap_auth=soap_request.auth, api_name=soap_request.api_name
+    )
+
+    if soap_config.privileges is not None:
```
freddieyebra: We can remove this conditional so we always call verify_access for this operation.

api/src/legacy_soap_api/grantors/services/confirm_application_delivery_response.py
```diff
@@ -0,0 +1,135 @@
...
+            ) from e
+
+    # Check if this user has already retrieved this submission.
+    existing_retrieval = db_session.execute(
+        select(ApplicationSubmissionRetrieved).where(
+            ApplicationSubmissionRetrieved.application_submission_id
+            == application_submission.application_submission_id,
+            ApplicationSubmissionRetrieved.created_by_user_id == certificate.user.user_id,
+        )
+    ).scalar()
```
freddieyebra: If we only need to check for existence here, could we do .scalar().first()? or combine this with an order by if we need latest.

api/src/legacy_soap_api/grantors/services/confirm_application_delivery_response.py
```diff
@@ -0,0 +1,135 @@
...
+)
+from src.legacy_soap_api.legacy_soap_api_config import SOAPOperationConfig
+from src.legacy_soap_api.legacy_soap_api_constants import LegacySoapApiEvent
+from src.legacy_soap_api.legacy_soap_api_schemas import SOAPRequest
+
+logger = logging.getLogger(__name__)
+
+CONFIRM_DELIVERY_FAULT_STRING = "Failed to confirm application delivery.(Authorization Failure)"
+
+VALID_STATUSES_FOR_DELIVERY = {ApplicationStatus.ACCEPTED}
```
freddieyebra, line 27: I believe this should have submitted status as well. 

```suggestion
VALID_STATUSES_FOR_DELIVERY = {ApplicationStatus.ACCEPTED, ApplicationStatus.SUBMITTED}
```

api/src/legacy_soap_api/legacy_soap_api_client.py
```diff
@@ -242,6 +243,18 @@ def get_application_zip_request(
             soap_config=self.operation_config,
         )
 
+    def confirm_application_delivery_request(
```
freddieyebra, line 246: nice!

api/src/legacy_soap_api/grantors/services/confirm_application_delivery_response.py
kkrug: If I remove this check, mypy sees `soap_config.privileges` as `set[Privilege] | None`, which causes a linter error. I can add something like `assert soap_config.privileges is not None` or just keep the conditional as is.

kkrug: Actually on second thought I think the `assert` is the better option, since we wouldn't want the verify check to ever be skipped.

api/src/legacy_soap_api/grantors/services/confirm_application_delivery_response.py
mdragon: I don't think they could get something that was just submitted, because it wouldn't have the ZIP built for them to retrieve it. Maybe in a world where we supported the XML (not ZIP) endpoint for delivery this could be true. But for now we probably just want ACCEPTED.

AND we should make sure only ACCEPTED are returned when they call GetSubmissionListExpanded (that may already be the case)

freddieyebra: @kkrug Matt is correct. ACCEPTED should be the only valid application status.

api/src/legacy_soap_api/grantors/services/confirm_application_delivery_response.py
freddieyebra: we can also just raise an exception if no privileges are defined

```python
if soap_config.privileges is None:
    raise Exception
```

api/src/legacy_soap_api/grantors/services/confirm_application_delivery_response.py
kkrug: Removed

api/src/legacy_soap_api/grantors/services/confirm_application_delivery_response.py
kkrug: Updated

api/src/legacy_soap_api/grantors/services/confirm_application_delivery_response.py
kkrug: Updated and added new fault classes

api/src/legacy_soap_api/grantors/services/confirm_application_delivery_response.py
kkrug: Updated

api/src/legacy_soap_api/grantors/services/confirm_application_delivery_response.py
kkrug: Updated

api/src/legacy_soap_api/grantors/fault_messages.py
```diff
@@ -9,3 +9,18 @@
...
+
+ConfirmDeliveryInvalidStatus = FaultMessage(
+    faultstring="Failed to confirm application delivery.(Authorization Failure)",
+    faultcode="soap:Server",
+)
+
+ConfirmDeliveryAlreadyRetrieved = FaultMessage(
+    faultstring="Failed to confirm application delivery.(Authorization Failure)",
+    faultcode="soap:Server",
+)
```
freddieyebra, line 26: Can we make these have faultstrings that are more descriptive based on the error that happens.
```suggestion
ConfirmDeliverySubmissionNotFound = FaultMessage(
    faultstring="Unable to find application from tracking number. Failed to confirm application delivery.",
    faultcode="soap:Server",
)

ConfirmDeliveryInvalidStatus = FaultMessage(
    faultstring="Invalid application status. Expected Accepted status. Failed to confirm application delivery.",
    faultcode="soap:Server",
)

ConfirmDeliveryAlreadyRetrieved = FaultMessage(
    faultstring="This application submission has already been retrieved. Failed to confirm application delivery.",
    faultcode="soap:Server",
)
```

kkrug: Updated

PR #9067, opened 2026-03-16
Close this

aoysimmons-GH:
Remove


doug-s-nava (commented):
see previous reviews

frontend/tests/e2e/sign-in.feature
```diff
@@ -0,0 +1,15 @@
+Feature: Sign in
```
doug-s-nava: It should be made clear that this test is only a test of signing in against the local mock 0auth server. Is it worth mentioning the separate feature file at https://github.com/HHS/simpler-grants-gov/blob/main/frontend/tests/e2e/login/features/login-with-login-gov.feature

can we put this in the same directory as that feature?

frontend/tests/e2e/sign-in.feature
```diff
@@ -0,0 +1,15 @@
...
+  As a visitor
+  I want to sign in from the homepage
+  So I can access my account
+
+  Background:
+    Given I open the homepage
+
+  Scenario: Complete sign in flow
+    When I click the sign-in button
+    And I continue with sign-in
```
doug-s-nava: what does this mean?

frontend/tests/e2e/sign-in.feature
```diff
@@ -0,0 +1,15 @@
...
+    And I enter my username
+    And I submit the login
+    Then I should be logged in
+    And I should see my email in the user menu
```
doug-s-nava: I think this is outdated - the email no longer appears in the header

doug-s-nava (commented):
this test is currently skipped - I've made a new ticket around re-enabling it, and we can put off this work to be part of that ticket instead, if that will make things clearer https://github.com/HHS/simpler-grants-gov/issues/9152

Issue #9079, opened 2026-03-17
Stabilize flaky 404 test


Bhavna-Ramachandran:
### Summary

The 404 test is intermittently failing on staging due to errors on static assets causing a blank page. This task is aimed at the root cause.

### Acceptance criteria

The 404 e2e test no longer fails on staging and continues to pass when staging is healthy.

Bhavna-Ramachandran:
PR #9080 has been reviewed, approved and merged by @doug-s-nava . Hence marking this ticket as closed.

PR #9080, opened 2026-03-17
[Issue #9079] Stabilize 404.spec.ts for E2E staging run

Bhavna-Ramachandran:
## Summary
Work for : Issue #9079 

## Changes proposed
Updated waitUntil from domcontentloaded to load and added explicit timeouts to assertions. Added a health check in beforeEach to skip the test rather than fail when staging is unhealthy.

### DIFF
--- a/frontend/tests/e2e/404.spec.ts
+++ b/frontend/tests/e2e/404.spec.ts
@@ -6,12 +6,19 @@ const { targetEnv } = playwrightEnv;
 test.beforeEach(async ({ page }) => {
   const timeout = targetEnv === "staging" ? 180000 : 60000;
   await page.goto("/imnothere", {
-    waitUntil: "domcontentloaded",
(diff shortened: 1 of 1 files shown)


frontend/tests/e2e/404.spec.ts
```diff
@@ -6,12 +6,19 @@ const { targetEnv } = playwrightEnv;
...
   await page.goto("/imnothere", {
-    waitUntil: "domcontentloaded",
+    waitUntil: "load",
     timeout,
   });
   // Wait for staging to stabilize
   if (targetEnv === "staging") {
     await page.waitForTimeout(3000);
+    // Skip if staging is unhealthy
+    const isUnhealthy = await page.evaluate(() =>
```
doug-s-nava, line 16: seems like this will likely be useful in general, but good to add it here for now and see how it does

PR #9085, opened 2026-03-17
[Issue #8103] Restructure logging/metrics for workflow service

### LINKED ISSUE #8103: Adjust logic to add more thorough logging and metrics
### Summary

Rough placeholder, but effectively after we've setup the workflow service, take a second pass at the logic and add any metrics or logging.

We’ll have added some during earlier steps, this is to take another pass and add more. We want to focus on:
* Observability (what did we process / how / when)
* Timing metrics
* Event types / workflow types / workflow tracking
* Error cases and counts

### Acceptance criteria

- [ ] Additional metrics and logging added

chouinar:
## Summary
Fixes #8103

## Changes proposed
* Rework / reconfigure the workflow management logging/metrics
* Changed some logic in the SQS parsing to make it easier to link SQS messages to underlying events (created an SQS Container)
* Had to rework all the tests that interact with the event handler to use the new SQS container - no meaningful changes
* Adjusted the main loop to no longer sleep - the sqs polling waits 10 seconds now, so the sleep isn't needed

## Context for reviewers
This is mostly just cleaning up logging and adding a few in places that were missing it. I ran things locally and just kept adjusting it when it felt like there was too much/too little info for a given event. The biggest change by far is making the SqsMessageContainer class. I wanted to pass in a metric context to the event handler and reworking it again to pass another field down convinced me that we should instead do this approach. That way the event handler input doesn't need to change if we add more later - we can just add it to this class.

Made it so the process tracks a few different useful metrics:
* Duration of processing request (one for just event handling, one factoring in SQS queue time)
* Error classes clearly separated out
* More consistent info on each log
* An obvious message to check against for metrics / "what happened" for a given workflow

## Validation steps
You'll need the API running locally, and also want to run the workflow service with `make cmd args="workflow workflow-main"`. To send events, go to http://localhost:8080/docs#/Workflow/put_v1_workflows_events - you can use the TODO Api User Key for sending events.

You'll need to find an opportunity in your DB, you can then send workflow events against that opportunity with (swapping in the appropriate ID):

```py
{
  "metadata": {},
  "start_workflow_context": {
    "entity_id": "717b8d3a-f4a5-4d45-854a-607ae6aed554",
    "entity_type": "opportunity",
  }
}
```

If your service is running, it'll immediately consume the events - for me locally these events take maybe 20ms to get processed from sending them.

This screenshot shows a few loops of finding no messages.
Screenshot 2026-03-17 at 12 06 03 PM

This shows it processing an event
Screenshot 2026-03-17 at 12 05 41 PM

### DIFF
--- a/api/src/workflow/event/workflow_metric_context.py
+++ b/api/src/workflow/event/workflow_metric_context.py
@@ -18,8 +18,6 @@ class WorkflowMetricContext:
         WORKFLOW_TRANSITION_COUNT = "workflow_transition_count"
         WORKFLOW_EVENT_COUNT = "workflow_event_count"
 
-    metrics: dict[str, Any]
-
     def __init__(self) -> None:
         self.metrics: dict[str, Any] = {metric: 0 for metric in self.Metrics}
         self.log_extra: dict[str, Any] = {}


api/src/workflow/handler/event_handler.py
```diff
@@ -52,29 +53,34 @@ class EventHandler:
...
     def process(self) -> BaseStateMachine:
         """Process an event."""
+        logger.info("Processing event", extra=self.event.get_log_extra())
         state_machine_event = self._pre_process_event()
         return self._process_event(state_machine_event)
 
     def _process_event(self, state_machine_event: StateMachineEvent) -> BaseStateMachine:
         """Run the state machine event against the state machine."""
+        log_extra = state_machine_event.get_log_extra()
+        logger.info("Processing event against state machine", extra=log_extra)
```
kkrug: It's a bit of a nitpick but do we think `"Processing event"` and `"Processing event against state machine"` should be differentiated more?

api/src/workflow/event/workflow_metric_context.py
```diff
@@ -0,0 +1,46 @@
...
+
+        WORKFLOW_TRANSITION_COUNT = "workflow_transition_count"
+        WORKFLOW_EVENT_COUNT = "workflow_event_count"
+
+    metrics: dict[str, Any]
+
+    def __init__(self) -> None:
+        self.metrics = {metric: 0 for metric in self.Metrics}
+        self.log_extra: dict[str, Any] = {}
+        self.start_time = time.perf_counter()
```
kkrug: Nothing really functional, more so style
```suggestion
    def __init__(self) -> None:
        self.metrics: dict[str, Any] = {metric: 0 for metric in self.Metrics}
        self.log_extra: dict[str, Any] = {}
        self.start_time = time.perf_counter()
```

api/src/workflow/handler/event_handler.py
chouinar: The logs are mostly to show where it's at in the process, helpful for debugging if anything goes wrong. But yeah, can clean up the messages a bit.

chouinar: Realized in the pre process function I have a log for start/process workflow, so this first message is a bit redundant.

api/src/workflow/event/workflow_metric_context.py
```diff
@@ -21,7 +21,7 @@ class Metrics(StrEnum):
     metrics: dict[str, Any]
```
kkrug: I should've made it more clear in my original comment but you don't need this line anymore.

chouinar: Ah, looked at just the other part of what was highlighted - fixed

Issue #7884, opened 2026-01-09
Add Host Monitoring to OpenSearch


joshtonava:
### Summary

OpenSearch is not shown in the Hosts section for new relic (nothing is shown currently)

### Acceptance criteria

Ensure our OpenSearch services are properly instrumented and host monitoring is properly configured

PR #9091, opened 2026-03-17
[Issue #7884]: Add es log forwarding via lambda to newrelic

prasnava:
## Summary

Implemented a lambda function to forward all opensearch logs to newrelic. 

## Validation steps

We should see opensearch logs in newrelic, but in the dev environment we don't have much es logging. That last log entry was written in 2024-10-30.   

Deployed the api / database module manually since we don't have that part automated in gh actions pipeline.


mdragon:
@sean-navapbc can you review this?

sean-navapbc:
## Review

### 1. New Relic license key exposed as plaintext

The Lambda env var `NR_LICENSE_KEY` is set directly from the SSM parameter value:

```hcl
NR_LICENSE_KEY = data.aws_ssm_parameter.newrelic_license_key.value
```

The checkov skip comment says "NR license key is passed via SSM, not plaintext" but that's misleading — the key **is** stored as plaintext in the Lambda's environment configuration (visible in the AWS console, Terraform state, and CloudFormation). It's only *sourced* from SSM.

Better approach: have the Lambda read the SSM parameter at runtime. This would require an IAM policy granting `ssm:GetParameter` but keeps the secret out of the environment and state.

### 2. No error handling in the Lambda

`urllib.request.urlopen` has no try/except. If New Relic is unreachable or returns a 5xx, the Lambda fails with an unhandled exception. CloudWatch will retry on Lambda errors, but it would be better to:
- Log the error so it's visible in the Lambda's own log group
- Distinguish retryable vs non-retryable errors

### 3. Lambda log group retention

`retention_in_days = 1827` (5 years) seems excessive for a utility Lambda whose only job is forwarding logs. Consider 30–90 days since the actual OpenSearch logs are being forwarded to New Relic anyway.

### 4. Confirm account-level CloudWatch log resource policy

This PR removes `aws_cloudwatch_log_resource_policy.opensearch` and the `depends_on` from the OpenSearch domain, with a comment that it's now managed at the account level. Existing deployments will see a plan that destroys the module-level policy — need to confirm the account-level policy already exists and covers this log group before merging, otherwise OpenSearch loses the ability to write to CloudWatch.

### 5. Filter pattern

`filter_pattern = ""` forwards ALL log events. This is intentional per the PR description but worth validating that New Relic ingest costs are acceptable if OpenSearch log groups become chatty.

sean-navapbc:
## Suggested Fixes

### 1. New Relic license key — read from SSM at runtime

Instead of passing the key as an environment variable, have the Lambda fetch it at runtime:

**Lambda (`newrelic_log_forwarder.py`):**
```python
import boto3
import os

ssm = boto3.client("ssm")

def get_nr_license_key():
    response = ssm.get_parameter(
        Name=os.environ["NR_LICENSE_KEY_SSM_PATH"],
        WithDecryption=True,
    )
    return response["Parameter"]["Value"]
```

Then in the handler, call `get_nr_license_key()` instead of `os.environ["NR_LICENSE_KEY"]`. You can cache the value outside the handler for reuse across warm invocations.

**Terraform (`log_forwarding.tf`):**

Replace the environment variable with the SSM path:
```hcl
environment {
  variables = {
    NR_LICENSE_KEY_SSM_PATH = "/monitoring/newrelic-license-key"
  }
}
```

Add an IAM policy for SSM access:
```hcl
resource "aws_iam_role_policy" "lambda_ssm" {
  name = "${local.lambda_name}-ssm"
  role = aws_iam_role.newrelic_log_forwarder.id

  policy = jsonencode({
    Version = "2012-10-17"
    Statement = [{
      Effect   = "Allow"
      Action   = "ssm:GetParameter"
      Resource = data.aws_ssm_parameter.newrelic_license_key.arn
    }]
  })
}
```

Then you can remove the checkov skip comment since the secret is no longer in the environment.

### 2. Error handling in the Lambda

Wrap the HTTP call:
```python
def handler(event, context):
    payload = decode_event(event)
    nr_license_key = get_nr_license_key()

    request = urllib.request.Request(
        NR_LOG_ENDPOINT,
        data=json.dumps(payload).encode(),
        headers={
            "Content-Type": "application/json",
            "Api-Key": nr_license_key,
        },
    )

    try:
        response = urllib.request.urlopen(request)
        print(f"New Relic response: {response.status}")
    except urllib.error.HTTPError as e:
        print(f"New Relic HTTP error: {e.code} {e.reason}")
        if e.code >= 500:
            raise  # Retryable — let Lambda/CloudWatch retry
        print(f"Non-retryable error, dropping {len(payload)} log events")
    except urllib.error.URLError as e:
        print(f"New Relic connection error: {e.reason}")
        raise  # Retryable — network issue
```

This logs errors clearly and only raises (triggering CloudWatch retry) for transient failures.

### 3. Lambda log group retention

Change from 5 years to 90 days:
```hcl
resource "aws_cloudwatch_log_group" "newrelic_log_forwarder" {
  name              = "/aws/lambda/${local.lambda_name}"
  retention_in_days = 90
}
```

### 4. Confirm account-level log resource policy

Before merging, verify the account-level policy exists:
```bash
aws logs describe-resource-policies
```

Look for a policy that allows `es.amazonaws.com` to call `logs:PutLogEvents` and `logs:CreateLogStream`. If it doesn't exist, either:
- Add it to `infra/accounts/logs.tf` in a prerequisite PR, or
- Keep the module-level policy in this PR until the account-level one is confirmed

### 5. Filter pattern

If you want to reduce noise/cost, you can filter to specific log levels:
```hcl
filter_pattern = "?ERROR ?WARN ?Error ?Warning"
```

Or keep `""` for all events if that's the intent — just document the cost expectation.

prasnava:
@sean-navapbc Resolved all of the suggested fixes.  Idea is to capture all the logs into newrelic and then create alerts based on those. 

Per the checkov findings: CKV_AWS_338 -> recommends to retain logs for at least 1 year. 

Ran the command: `aws logs describe-resource-policies` - The proper permissions are there in the policyDocument fields. 

```
{
    "resourcePolicies": [
        {
            "policyName": "AWSLogDeliveryWrite20150319",
            "policyDocument": "...",
            "lastUpdatedTime": 1711060999991,
            "policyScope": "ACCOUNT"
        },
        {
            "policyName": "OpenSearchService-api-training-Application-logs",
            "policyDocument": "...",
            "lastUpdatedTime": 1755614289812,
            "policyScope": "ACCOUNT"
        },
        {
            "policyName": "OpenSearchService-api-training-Audit-logs",
            "policyDocument": "...",
            "lastUpdatedTime": 1755594652381,
            "policyScope": "ACCOUNT"
        },
        {
            "policyName": "OpenSearchService-api-training-Index-logs",
            "policyDocument": "...",
            "lastUpdatedTime": 1755613865944,
            "policyScope": "ACCOUNT"
        },
        {
            "policyName": "OpenSearchService-api-training-Search-logs",
            "policyDocument": "...",
            "lastUpdatedTime": 1755613083379,
            "policyScope": "ACCOUNT"
        },
        {
            "policyName": "account-level-logs",
            "policyDocument": "...",
            "lastUpdatedTime": 1730217422732,
            "policyScope": "ACCOUNT"
        },
        {
            "policyName": "api-grantee1-opensearch-logs",
            "policyDocument": "...",
            "lastUpdatedTime": 1769573914646,
            "policyScope": "ACCOUNT"
        },
        {
            "policyName": "api-prod-opensearch-logs",
            "policyDocument": "...",
            "lastUpdatedTime": 1771435890585,
            "policyScope": "ACCOUNT"
        },
        {
            "policyName": "api-staging-opensearch-logs",
            "policyDocument": "...",
            "lastUpdatedTime": 1770841642318,
            "policyScope": "ACCOUNT"
        },
        {
            "policyName": "api-training-opensearch-logs",
            "policyDocument": "...",
            "lastUpdatedTime": 1771436586098,
            "policyScope": "ACCOUNT"
        }
    ]
}
```

PR #9094, opened 2026-03-17
[Issue#9051] Update API Docs Banner

### LINKED ISSUE #9051: Update API Docs Banner
### Summary

The banner at the top of the Simpler Grants API docs currently reads as an alpha-era disclaimer warning users it is not suitable for production. The API is now in full production and in active daily use. We need to determine the final copy for the updated banner and implement it in the docs.

### Acceptance criteria

- [ ] Final banner copy is reviewed by design. 
  - [ ] The updated banner no longer references alpha status or discourages production use.
  - [ ] The banner accurately communicates that the API is production-ready while setting clear expectations that endpoints may change with advance notice.
- [ ] The banner update is live in the deployed docs environment and confirmed visually by the implementer.

btabaska:
## Summary

Fixes #9051  

## Changes proposed

Updated API docs banner based on current development state
[Context](https://betagrantsgov.slack.com/archives/C05TSL64VUH/p1773430798775569)

## Context for reviewers
https://api.dev.simpler.grants.gov/docs
[Design context](https://betagrantsgov.slack.com/archives/C05TGEL3C6Q/p1773672507119629)

## Validation steps
N/A

### DIFF
--- a/api/src/api/opportunities_v1/opportunity_routes.py
+++ b/api/src/api/opportunities_v1/opportunity_routes.py
@@ -34,6 +34,7 @@ SHARED_ALPHA_DESCRIPTION = """
 This API is in active development as we build out new functionalities for Simpler.Grants.gov.
 It is currently stable for everyday use, and will be versioned with advance notice for any breaking changes.
 
+Learn more in our [API documentation](https://wiki.simpler.grants.gov/product/api).
 See [Release Phases](https://github.com/github/roadmap?tab=readme-ov-file#release-phases) for further details.
 """
 
--- a/api/openapi.generated.yml
+++ b/api/openapi.generated.yml
@@ -8,7 +8,7 @@ info:
     
      It is currently stable for everyday use, and will be versioned with advance notice for any breaking changes.
 
-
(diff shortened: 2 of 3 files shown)


api/src/api/opportunities_v1/opportunity_routes.py
```diff
@@ -31,11 +31,8 @@
 
 # Descriptions in OpenAPI support markdown https://swagger.io/specification/
 SHARED_ALPHA_DESCRIPTION = """
-__ALPHA VERSION__
-
-This endpoint in its current form is primarily for testing and feedback.
-
-Features in this endpoint are still under heavy development, and subject to change. Not for production use.
+This API is in active development as we build out new functionalities for Simpler.Grants.gov.
```
chouinar, line 34: We should update these, but this won't be what appears at the top of the page. That's https://github.com/HHS/simpler-grants-gov/blob/main/api/src/app.py#L52

Could we update that and have it link to the developer docs you've written?

btabaska: Iterated based on this feedback

Issue #8708, opened 2026-02-27
Implement testing groups and cadences


doug-s-nava:
### Summary

General implementation ticket following on https://github.com/HHS/simpler-grants-gov/issues/7959

Note that #8586 and #8588 already exists, this is a catch all for implementing all other test runs not covered by those tickets

Some idea of this work is written up here, but the outcome of 7959 may be slightly different https://github.com/HHS/simpler-grants-gov/issues/7257

### Acceptance criteria

- [ ] all test runs and cadences outline in #7959 not already accounted for in other tickets are implemented

PR #9102, opened 2026-03-18
[Issue #8708] tag all e2e tests and introduce tag constants

doug-s-nava:
## Summary

Fixes for #8708 

## Changes proposed

In order to support running groups of tests, adds tags to all e2e tests in line with classification here https://navasage.atlassian.net/wiki/x/HICZpw

## Context for reviewers

Lots of indentation changes in the diff make this look like a bigger change that in it is. Any changes to note outside of adding tags are noted with comments.

## Validation steps

1. _VERIFY_: test runs of experimental job against multiple sets of tags all run the expected number of tests.

## Links to test runs

* https://github.com/HHS/simpler-grants-gov/actions/runs/23345445393/job/67909488055
  * "@auth" + "@static" = 6 + 13 = 19 tests * 2 browsers = 38 tests
  *  only 36 if running local, as the login-with-login-gov spec only runs in staging 

```
 Running 38 tests using 1 worker, shard 1 of 1
```  

* https://github.com/HHS/simpler-grants-gov/actions/runs/23346063120
  * "@smoke" = 10 tests * 2 browsers = 20 tests

```
Running 20 tests using 1 worker, shard 1 of 1
```

### DIFF
--- a/.github/actions/e2e/action.yml
+++ b/.github/actions/e2e/action.yml
@@ -56,13 +56,7 @@ runs:
       shell: bash
       env:
         PLAYWRIGHT_TARGET_ENV: ${{ inputs.target }}
-        NEEDS_SETUP: ${{ inputs.needs_node_setup }}
-        TAGS: ${{ inputs.playwright_tags }}
       run: |
-        echo $PLAYWRIGHT_TARGET_ENV
-        echo $NEEDS_SETUP
-        echo $TAGS
-        echo
         if [[ $PLAYWRIGHT_TARGET_ENV != "local" && $PLAYWRIGHT_TARGET_ENV != "staging" ]]; then echo "invalid e2e test target provided" && exit 1; fi
 
     - name: Check out code


.github/actions/e2e/action.yml
```diff
@@ -57,7 +57,6 @@ runs:
       env:
         PLAYWRIGHT_TARGET_ENV: ${{ inputs.target }}
       run: |
-        echo $PLAYWRIGHT_TARGET_ENV
```
Bhavna-Ramachandran, line 60: Was this removed because we now have clear distinctions with two workflows?

frontend/tests/e2e/apply/happy-path-application-submission.spec.ts
```diff
@@ -40,7 +43,7 @@ test.beforeEach(({ page: _ }, testInfo) => {
 
 test(
   "Application submission happy path - application with required SF424B and unsubmitted conditional SFLLL",
-  { tag: ["@smoke", "@grantee", "@apply"] },
+  { tag: [SMOKE, GRANTEE, APPLY] },
```
Bhavna-Ramachandran, line 46: Good call 👍

.github/actions/e2e/action.yml
doug-s-nava: This was in there for debugging, don't think we need it anymore

PR #9107, opened 2026-03-18
[Issue #9044] Upgrade packages flagged by ECR

mdragon:
## Summary

Work for #9045

## Changes proposed

Update several packages that ECR has flagged as having fixed vulnerabilities

## Context for reviewers

These were not getting flagged by our CI/CD scans, but were getting picked up by ECR scanning. 

Noting here and will comment in the PR at the appropriate spot. Despite many of these packages flagging as in our poetry lock in ECR, it looks like we weren't installing them directly previously. Not sure if that means we should ignore these for now rather than doing this approach.

## Validation steps

- ECR scans pass after merge to main and container builds.

### DIFF
--- a/analytics/pyproject.toml
+++ b/analytics/pyproject.toml
@@ -12,7 +12,7 @@ analytics = "analytics.cli:app"
 db-migrate = "analytics.cli:migrate_database"
 
 [tool.poetry.dependencies]
-dynaconf = "^3.2.4"
+dynaconf = "^3.2.13"
 pandas = "^2.0.3"
 pandas-stubs = "^2.0.2.230605"
 pydantic = "^2.0.3"
--- a/analytics/poetry.lock
+++ b/analytics/poetry.lock
@@ -989,14 +989,14 @@ profile = ["gprof2dot (>=2022.7.29)"]
 
 [[package]]
 name = "dynaconf"
-version = "3.2.12"
+version = "3.2.13"
 description = "The dynamic configurator for your Python Project"
 optional = false
 python-versions = ">=3.8"
 groups = ["main"]
 files = [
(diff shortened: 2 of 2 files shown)


mdragon:
@chouinar just cut a clean branch because otherwise we would have had to deal with conflict resolution in the poetry lock, etc.

Just merging the renovate updates didn't clear many of the flagged vulnerabilities. This clears more but not all of them.

analytics/pyproject.toml
```diff
@@ -26,12 +26,17 @@ boto3 = "^1.35.56"
...
 smart-open = "^7.0.5"
-newrelic = "11.5.0"
+newrelic = "^11.5.0"
 requests = "^2.32.5"
 urllib3 = "^2.6.3"
+nltk = "^3.9.3"
+werkzeug = "^3.1.6"
+marshmallow = "^4.1.2"
+filelock = "^3.20.3"
+authlib = "^1.6.6"
```
chouinar: Checking where these dependencies come from.

* `nltk` - The safety library
* `werkzeug` - Looks like just moto which is a test dependency. I think it's fine to put here.
* `marshmallow` - also just from the safety library
* `filelock` - Also just the safety library
* `authlib` - also the safety library

Safety is a dev dependency - and I'm not even sure what it is, or if it is used. It says its for CLI package scanning, but I'm not sure if that was ever added? It was added in the initial commit for this library, but I don't see any usage of it.

An easier fix might be to get rid of safety which'll get rid of several of these.

mdragon: Ok, this cleared up everything but 3 High and a Medium from the version of urllib3 that the newrelic package is installing for itself, which I don't think we can address easily as we're already installing the newest NR ourselves.

PR #9110, opened 2026-03-18
[Issue #9109] Change timing and add failure Slack notification

### LINKED ISSUE #9109: Adjust Sprint Rollover Action
### Summary

The roll over job failed today and we didn't get any notification. We noticed missing tickets (since they hadn't rolled over they fell off the Current Sprint board) and re-running the job allowed it to succeed. Moving it to a different time may keep us from exhausting quota.

### Acceptance criteria

- Change the time away from other GitHub scheduled jobs
- Include a failure notification to Slack

mdragon:
## Summary

Fixes #9109 

## Changes proposed

* Change timing of the run to avoid other GitHub heavy processes potentially affecting quota
* Include a Slack notification if the job fails

## Context for reviewers

This job failed this week (for the first time in recently history) and didn't roll over all the tickets. Trying to avoid the failures and make it a little more obvious if it does fail again in the future.

## Validation steps
- [x] Test runs succeed

### DIFF
--- a/.github/workflows/lint-sprint-rollover.yml
+++ b/.github/workflows/lint-sprint-rollover.yml
@@ -1,4 +1,6 @@
 name: Lint - Sprint rollover
+env:
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
 
 on:
   workflow_dispatch: # for manual runs


.github/workflows/lint-sprint-rollover.yml
```diff
@@ -1,12 +1,14 @@
 name: Lint - Sprint rollover
+env:
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
```
mdragon, line 3: I think this is fine to just leave in, at least for now. If something newer than 24 gets upgraded to in the future at the action level I don't think this will hold us back on 24.

PR #9114, opened 2026-03-18
[Issue #8995] Add saved opportunities notifications endpoint

### LINKED ISSUE #8995: [BE] Create saved opportunities notifications endpoint
### Summary

We want to create a new notifications endpoint which returns all the notification settings valid for a user for saved opportunities.  There are two types of saved opportunity notification preferences: 

* Notification of what opportunities the user has saved to their profile, which we will call self notification preferences
* Notification of opportunities that have been saved to an organization that the user is part of, even if that user didn't do the saving
 
This endpoint returns both sets of preferences.

Because default values change per scenario, the endpoint will return not just the data stored in the underlying table but what the settings should be for all circumstances. 

When processing the underlying table, if there is no entry with organization_id is null (e.g. a setting for self) we assume a default of email_enabled true.

For each organization the user_id is part of, if there is no entry with the same organization_id (e.g. the setting for that organization) we assume a default of email_enabled false.

Users should only see notification settings for organizations where they can add saved opportunities, so check the VIEW_ORG_SAVED_OPPORTUNITIES privilege before returning the notification setting for that organization. 

```
GET {user_id}/saved-opportunities/notifications 

{
    self: {
        email_enabled: true/false
    },
    organizations: [
        { 
	         organization_id: guid
	         email_enabled: true/false
        },
    ]
}
```

This endpoint should only be available for users with with a valid JWT token.
A 401 should be returned if the user isn't authenticated as the {user_id} in the GET request.

### Acceptance criteria
- [ ] Endpoint behaviour
   - [ ] Saved Opportunity notification setting are always returned for self, even if no explicit settings are listed in the underlying table
   - [ ] Saved Opportunity notification settings are always returned for all organizations that user is a member of, even if no explicit settings are listed in the underlying table. 
   - [ ] By default the saved opportunity notification setting for self is true.
   - [ ] By default the saved opportunity notification setting for that user for an organization is false.
   - [ ] If an explicit saved opportunity notification setting has been recorded for self it is returned.
   - [ ] If an explicit saved opportunity notification setting has been recorded for that user for an organization it is returned.
- [ ] A 401/Unauthorized should be returned if the authenticated user is not the same user as the {user_id} in the request or invalid/missing token. 
- [ ] Only saved opportunity notification settings should be returned for organizations where the user has the VIEW_ORG_SAVED_OPPORTUNITIES privilege for that organization. 
- [ ] Unittests

mikehgrantsgov:
## Summary

Fixes #8995  

## Changes proposed

Add `GET /<user_id>/saved-opportunities/notifications` returning notification preferences for a user's saved opportunities
Service layer queries `UserSavedOpportunityNotification` and applies business logic defaults: `email_enabled=True`, org-scoped settings default to `email_enabled=False`
Organization settings are only returned for orgs where the user has `VIEW_ORG_SAVED_OPPORTUNITIES` privilege

## Context for reviewers

The response shape is:

```json
{
  "self": { "email_enabled": true },
  "organizations": [{ "organization_id": "...", "email_enabled": false }]
}
```
`self` is always present even with no DB row. Organization entries only appear for orgs the user can access.

## Validation steps

See updated unit tests.

### DIFF
--- a/api/src/api/users/user_routes.py
+++ b/api/src/api/users/user_routes.py
@@ -826,22 +826,22 @@ def user_response_org_invitation(
 
 @user_blueprint.get("/<uuid:user_id>/saved-opportunities/notifications")
 @user_blueprint.output(UserSavedOpportunityNotificationsResponseSchema)
-@user_blueprint.doc(responses=[200, 403])
-@user_blueprint.auth_required(api_jwt_auth)
+@user_blueprint.doc(responses=[200, 401, 403], security=jwt_or_api_user_key_security_schemes)
+@jwt_or_api_user_key_multi_auth.login_required
 @flask_db.with_db_session()
 def user_get_saved_opportunity_notifications(
     db_session: db.Session, user_id: UUID
 ) -> response.ApiResponse:
     logger.info("GET /v1/users/:user_id/saved-opportunities/notifications")
-    user_token_session: UserTokenSession = api_jwt_auth.get_user_token_session()
+    user = jwt_or_api_user_key_multi_auth.get_user()
 
     # Verify the authenticated user matches the requested user_id
-    if user_token_session.user_id != user_id:
+    if user.user_id != user_id:
         raise_flask_error(403, "Forbidden")
 
     with db_session.begin():
-        db_session.add(user_token_session)
-        result = get_saved_opportunity_notification_preferences(db_session, user_token_session.user)
+        db_session.add(user)
+        result = get_saved_opportunity_notification_preferences(db_session, user)
 
     logger.info("Successfully fetched saved opportunity notification preferences")
     return response.ApiResponse(message="Success", data=result)


api/src/api/users/user_routes.py
```diff
@@ -818,3 +822,26 @@ def user_response_org_invitation(
         )
 
     return response.ApiResponse(message="Success", data=invitation_response)
+
+
+@user_blueprint.get("/<uuid:user_id>/saved-opportunities/notifications")
+@user_blueprint.output(UserSavedOpportunityNotificationsResponseSchema)
+@user_blueprint.doc(responses=[200, 403])
+@user_blueprint.auth_required(api_jwt_auth)
```
chouinar: I'd suggest for any new endpoints, we make them support JWT + User API Key auth unless we have a strong reason not to. One major benefit of API key auth - getting an API key is a lot easier than figuring out a JWT in staging if you want to call it directly for testing/demos.

mikehgrantsgov: Sounds good, updated this!

Issue #8210, opened 2026-01-28
Update SOAP/Proxy to be able to run and be tested locally.


jakobpederson:
### Summary

Currently the SOAP/PRoxy can be run and tested locally but it takes some extra work to make sure that the certificates are set up correctly and some other hacking. Ideally, it should be set up with local set up scripts. 

### Acceptance criteria

- [ ] Response from local SOAP proxy endpoint matches response from other lower environments
- [ ] SOAP/Proxy docs on confluence are updated to reflect new setup

ychoy:
FYI @freddieyebra we'll likely push to Sprint 5.4

jakobpederson:
[Confluence doc](https://navasage.atlassian.net/wiki/x/igCkr)

PR #9121, opened 2026-03-18
[Issue #8210] Create make command to setup local SOAP testing

jakobpederson:
## Summary

Fixes / Work for #8210  

## Changes proposed

- Added make command to create a legacy certificate and application submission to allow for local testing
- Changed `download_path` to `file_location` in `GetApplicationZip` when trying to get the zip file
- Updated tests to handle the `download_path` change

## Context for reviewers

Currently local testing is hard because we do not set up a local certificate. This script creates a local certificate (and a test submission) that allows the local instance to be hit by Postman and get data back.

## Validation steps

- [ ] unittests pass
- [ ] run script, hit endpoint `GetApplicationZip` with endpoint after putting the encoded url certificate into the header

### DIFF
--- a/api/.gitignore
+++ b/api/.gitignore
@@ -15,7 +15,7 @@ coverage.*
 /tests/validate_legacy_soap_api/cache/*
 !/tests/validate_legacy_soap_api/cache/__init__.py
 !/tests/lib/cache/__init__.py
-/cert_cache/*
+/cert_cache_dump/*
 
 # Python virtual environments
 .venv
--- a/api/tests/lib/seed_local_soap_certificate.py
+++ b/api/tests/lib/seed_local_soap_certificate.py
@@ -147,7 +147,7 @@ def _build_legacy_certificate_and_submission(
 # in order to hit the locally running instance
 # the command looks like
 # `make seed-local-soap-certificate DIR_PATH="~/test/cache/"`
-# dir_path defaults to test_cache dir that is in gitignore
+# dir_path defaults to cert_cache_dump dir that is in gitignore
 @click.command()
 @click.option(
     "--dir-path",
@@ -163,7 +163,7 @@ def seed_local_soap_certificate(dir_path: str | None, agency_code: str | None = 
         logger.info("Running seed script for local soap certificate testing")
         db_client = db.PostgresDBClient()
         with db_client.get_session() as db_session:
-            dir_path = dir_path if dir_path else "test_cache"
+            dir_path = dir_path if dir_path else "cert_cache_dump"
             directory = Path(dir_path)
             directory.mkdir(parents=True, exist_ok=True)
             factories._db_session = db_session


api/.gitignore
```diff
@@ -14,6 +14,8 @@ coverage.*
 .pytest_cache/
 /tests/validate_legacy_soap_api/cache/*
 !/tests/validate_legacy_soap_api/cache/__init__.py
+/tests/lib/cache/*
+!/tests/lib/cache/__init__.py
```
jakobpederson, line 17: Ignoring the cache where the cert & key are stashed.

api/tests/src/legacy_soap_api/test_legacy_soap_api_client.py
```diff
@@ -53,6 +53,7 @@
 GRANTS_GOV_TRACKING_NUMBER = "GRANT80000000"
 CID_UUID = "aaaaaaaa-1111-2222-3333-bbbbbbbbbbbb"
 BOUNDARY_UUID = "cccccccc-1111-2222-3333-dddddddddddd"
+ADDITIONAL_UUID = "eeeeeeee-1111-2222-3333-ffffffffffff"
```
jakobpederson, line 56: Changing the download_path seems to have increased the number of times uuid is called.

api/tests/lib/seed_local_soap_certificate.py
```diff
@@ -0,0 +1,186 @@
...
+
+    with open(f"{TARGET_DIR}/local.crt") as f:
+        cert_text = f.read()
+
+    encoded = quote(cert_text, safe="")
+    print(
+        "For local testing use this url in Postman: 'http://localhost:8080/grantsws-agency/services/v2/AgencyWebServicesSoapPort'"
+    )
+    print("\nCOPY AND PASTE THIS STRING INTO THE POSTMAN HEADER AS 'X-Amzn-Mtls-Clientcert'")
+    print(encoded)
```
jakobpederson, line 141: Because we're currently caching the crt and key, you should only have to copy and paste this into your Postman header once as it will only change if the cached files are deleted.

api/tests/lib/seed_local_soap_certificate.py
```diff
@@ -0,0 +1,196 @@
...
+
+def create_opportunity(db_session, agency, status):
+    # Due to issues with agency being overridden an Opportunity is created without a factory
+    opportunity = Opportunity(
+        opportunity_number=f"{uuid.uuid4()}",
+        opportunity_title=f"Test-Opportunity-{status}",
+        agency_code=agency.agency_code,
+        is_draft=False,
+    )
+    opportunity.agency_record = agency
```
chouinar: This line won't do anything, agency_record isn't an ordinary relationship, there's no actual foreign key.

api/tests/lib/seed_local_soap_certificate.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+def create_opportunity(db_session, agency, status):
+    # Due to issues with agency being overridden an Opportunity is created without a factory
+    opportunity = Opportunity(
+        opportunity_number=f"{uuid.uuid4()}",
+        opportunity_title=f"Test-Opportunity-{status}",
+        agency_code=agency.agency_code,
+        is_draft=False,
+    )
```
chouinar: What issue were you encountering? This won't make an opportunity with much in it, and it won't really be a realistic scenario.

All you need to do to set an agency for an opportunity is set the agency code.

api/tests/lib/seed_local_soap_certificate.py
```diff
@@ -0,0 +1,196 @@
...
+        )
+        .sign(key, algorithm=hashes.SHA256())
+    )
+    with open(f"{TARGET_DIR}/local.crt", "wb") as f:
+        f.write(cert.public_bytes(serialization.Encoding.PEM))
+
+
+def get_or_create_agency(db_session):
+    stmt = select(exists().where(Agency.agency_code == AGENCY_INFO["agency_code"]))
+    agency_exists = db_session.execute(stmt).first()[0]
```
chouinar: You're making this query a lot more complex than it needs to be here and in other places. A simpler pattern.

```py
agency = db_session.scalar(select(Agency).where(Agency.agency_code == "X"))
if agency is None:
   agency = AgencyFactory.create(...)

return agency
```

If it exists, you fetch it and return it, otherwise you get nothing and then make it.

api/tests/lib/seed_local_soap_certificate.py
```diff
@@ -0,0 +1,196 @@
...
+    )
+    print("\nCOPY AND PASTE THIS STRING INTO THE POSTMAN HEADER AS 'X-Amzn-Mtls-Clientcert'")
+    print(encoded)
+
+
+if __name__ == "__main__":
+    db_client = db.PostgresDBClient()
+    with db_client.get_session() as db_session:
+        factories._db_session = db_session
+        _build_legacy_certificate_and_submission(db_session)
```
chouinar: I'd suggest following what we did in `seed_local_db` for the entry function. Click gives some nice utilities like specifying parameters easily + click has a nicer print functionality, although you should just use our logging.

api/tests/lib/seed_local_soap_certificate.py
```diff
@@ -0,0 +1,196 @@
...
+    if key:
+        create_cert(key)
+    with open(f"{TARGET_DIR}/local.crt", "rb") as f:
+        cert_data = f.read()
+    cert = x509.load_pem_x509_certificate(cert_data, default_backend())
+    serial_number = hex(cert.serial_number).lower().lstrip("0x")
+
+    agency = get_or_create_agency(db_session)
+    for status in ApplicationStatus:
+        create_application_submission(db_session, agency, serial_number, status=status)
```
chouinar: Why are we making submissions in application statuses that aren't `accepted`? That's not really a case that will happen.

api/tests/lib/seed_local_soap_certificate.py
```diff
@@ -0,0 +1,196 @@
...
+):
+    opportunity = create_opportunity(db_session, agency, status)
+    db_session.add(opportunity)
+    competition = factories.CompetitionFactory(
+        opportunity=opportunity,
+    )
+    application = factories.ApplicationFactory.create(
+        competition=competition, with_forms=True, application_status=status
+    )
+    submission = factories.ApplicationSubmissionFactory.create(application=application)
```
chouinar: If you want a submission that is at least mildly realistic, the main seed script creates a bunch of opportunities in various scenarios - I might suggest just relying on those instead of making more here. See for example the `_add_application` function that is trying to emulate the actual app submission job.

api/tests/lib/seed_local_soap_certificate.py
```diff
@@ -0,0 +1,196 @@
...
+    )
+    opportunity.agency_record = agency
+    return opportunity
+
+
+def get_or_create_legacy_certificate(db_session, agency, serial_number):
+    stmt = select(exists().where(LegacyCertificate.serial_number == serial_number))
+    certificate_check = db_session.execute(stmt).first()
+    legacy_certificate_exists = certificate_check[0] if certificate_check else None
+    if not legacy_certificate_exists:
```
chouinar: Same here as above, just fetch it, don't check if it exists, just get it.

api/tests/lib/seed_local_soap_certificate.py
```diff
@@ -0,0 +1,196 @@
...
+import src.adapters.db as db
+import tests.src.db.models.factories as factories
+from src.constants.lookup_constants import ApplicationStatus, Privilege
+from src.db.models.agency_models import Agency
+from src.db.models.opportunity_models import Opportunity
+from src.db.models.user_models import LegacyCertificate
+
+logger = logging.getLogger(__name__)
+CURRENT_DIR = Path(__file__).parent
+TARGET_DIR = CURRENT_DIR / "cache"
```
chouinar: I'd make the directory a parameter a user has to pass in. Don't make it a hard-coded directory in the package, even with gitignores, just want to avoid folks checking in certs.

api/Makefile
```diff
@@ -307,6 +307,9 @@ seed-local-legacy-tables:
 generate-internal-token:
 	$(FLASK_CMD) task generate-internal-token $(args)
 
+seed-local-soap-certificate:
+	$(PY_RUN_CMD) python3 -m tests.lib.seed_local_soap_certificate
```
chouinar: This should have a path similar to the db-seed-local script. shouldn't ever need to put `python` in this.

api/tests/lib/seed_local_soap_certificate.py
jakobpederson: The agency kept being swapped out. It was pretty unclear why, it's working now so I'm switching back to factories.

api/tests/lib/seed_local_soap_certificate.py
jakobpederson: I was thinking it would be helpful for testing to represent all states. I'll take that out.

api/tests/lib/seed_local_soap_certificate.py
jakobpederson: Yeah that's a lot easier.

api/tests/lib/seed_local_soap_certificate.py
jakobpederson: I've updated it to leverage existing opportunities & competitions and use the `_add_application` to create the submissions. The result does look a lot more legitimate.

api/tests/lib/seed_local_soap_certificate.py
```diff
@@ -0,0 +1,192 @@
...
+def get_or_create_legacy_certificate(db_session, agency, serial_number):
+    legacy_certificate = db_session.scalar(
+        select(LegacyCertificate).where(LegacyCertificate.serial_number == serial_number)
+    )
+    if legacy_certificate is None:
+        legacy_certificate = factories.LegacyAgencyCertificateFactory(
+            agency_id=agency.agency_id,
+            agency=agency,
+            serial_number=serial_number,
+            expiration_date=datetime.datetime.now(datetime.UTC).date() + timedelta(days=365),
```
chouinar: For any datetime stuff, just use our datetime util - every version of python likes to slightly adjust the right way to do dates - easier to have it all in one place.

```suggestion
            expiration_date=datetime_util.utcnow() + timedelta(days=365),
```

Same with the other datetime usages above.

api/tests/lib/seed_local_soap_certificate.py
```diff
@@ -0,0 +1,192 @@
...
+        .scalars()
+        .all()
+    )
+    competition = (
+        db_session.execute(
+            select(Competition).where(Competition.opportunity_id.in_(opportunity_ids)).limit(4)
+        )
+        .scalars()
+        .first()
+    )
```
chouinar: If our main seed script creates application submissions, would it be easier to just rely on that? Otherwise, there's a few issues to address here.

* Don't do a LIKE against the agency code - a cert is configured to access a specific set of agencies - unless we recursively setup for all subagencies here, we could setup an app the user can't even access.
* If you want competitions for opportunities, get the opportunities and iterate over them, checking the competitions. A little more direct.

api/tests/lib/seed_local_soap_certificate.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+# Before running this command run `make db-seed-local-with-agencies`
+# This method creates a cert and key if it cannot find them in the cache
+# it then gets or creates an agency with an agency_code of 'SOAP'
+# then it creates an opportunity -> competition -> application -> application_submission
+# then it gets or creates a legacy _certificate for that agency and cert serial_number
+# it prints off the encoded cert that can be put into the Postman header
+# in order to hit the locally running instance
+# the command looks like `make seed-local-soap-certificate DIR_PATH="~/test/cache/" AGENCY_CODE="EPA"`
```
chouinar: Should this comment go on the command / main function below?

api/tests/lib/seed_local_soap_certificate.py
```diff
@@ -0,0 +1,192 @@
...
+    "--agency-code",
+    help="Agency code to use ('DOD' is default)",
+)
+def seed_local_soap_certificate(dir_path: str, agency_code: str | None = None) -> None:
+    with src.logging.init("seed_local_soap_certificate"):
+        logger.info("Running seed script for local soap certificate testing")
+        db_client = db.PostgresDBClient()
+        with db_client.get_session() as db_session:
+            directory = Path(dir_path).expanduser()
+            directory.mkdir(parents=True, exist_ok=True)
```
chouinar: What does expanduser do here? If I pass in `/absolute/path/to/folder` or `~/path/to/folder` would those end up changed?

jakobpederson: `expandduser` takes the `~` and converts it the home directory. Unfortunately, that seems to be in the docker instance. If you just pass in `/absolute/path/to/folder` you'll get`simpler-grants-gov/api/absolute/path/to/folder`. If you don't have it and pass in `~/path/to/folder` you'll get `api/~/path/to/folder` https://docs.python.org/3/library/os.path.html#os.path.expanduser

api/tests/lib/seed_local_soap_certificate.py
jakobpederson: 100% should be there. I'll move it.

api/tests/lib/seed_local_soap_certificate.py
jakobpederson: I'll just grab a submission that already exists then. Makes it a lot easier.

api/tests/lib/seed_local_soap_certificate.py
```diff
@@ -0,0 +1,192 @@
...
+            agency=agency, user=legacy_certificate.user
+        )
+        role = factories.RoleFactory.create(privileges=PRIVILEGES, is_agency_role=True)
+        factories.AgencyUserRoleFactory.create(agency_user=agency_user, role=role)
+        db_session.add(legacy_certificate)
+    else:
+        logger.info("Test legacy_certificate already exists")
+
+
+def create_application_submission(db_session, agency, serial_number):
```
kkrug: Do we need `serial_number` as an argument here? It looks unused.

api/tests/lib/seed_local_soap_certificate.py
```diff
@@ -0,0 +1,192 @@
...
+def seed_local_soap_certificate(dir_path: str, agency_code: str | None = None) -> None:
+    with src.logging.init("seed_local_soap_certificate"):
+        logger.info("Running seed script for local soap certificate testing")
+        db_client = db.PostgresDBClient()
+        with db_client.get_session() as db_session:
+            directory = Path(dir_path).expanduser()
+            directory.mkdir(parents=True, exist_ok=True)
+            factories._db_session = db_session
+            _build_legacy_certificate_and_submission(
+                db_session, directory, agency_code if agency_code else "EPA"
```
kkrug: The help comment says 'DOD' is the default, but the code uses 'EPA' as the default.

api/tests/lib/seed_local_soap_certificate.py
```diff
@@ -0,0 +1,192 @@
...
+    )
+    print("\nCOPY AND PASTE THIS STRING INTO THE POSTMAN HEADER AS 'X-Amzn-Mtls-Clientcert'")
+    print(encoded)
+
+
+@click.command()
+@click.option(
+    "--dir-path",
+    required=True,
+    help="Directory path for the certificte and key file",
```
kkrug:
```suggestion
    help="Directory path for the certificate and key file",
```

api/tests/lib/seed_local_soap_certificate.py
```diff
@@ -0,0 +1,192 @@
...
+    )
+    competition = (
+        db_session.execute(
+            select(Competition).where(Competition.opportunity_id.in_(opportunity_ids)).limit(4)
+        )
+        .scalars()
+        .first()
+    )
+    if competition:
+        for i in range(0, 2):
```
kkrug: Minor nitpick but `range(2)` would do the same thing as `range(0, 2)`

api/tests/lib/seed_local_soap_certificate.py
```diff
@@ -0,0 +1,192 @@
...
+        logger.info("cert and key file exist")
+    else:
+        key = create_private_key(path_key)
+        create_cert(key, path_crt)
+    with open(path_crt, "rb") as f:
+        cert_data = f.read()
+    cert = x509.load_pem_x509_certificate(cert_data, default_backend())
+    serial_number = hex(cert.serial_number).lower().lstrip("0x")
+
+    agency = get_agency(db_session, agency_code)
```
kkrug, line 125: If `get_agency()` returns `None`, you could get some nondescript errors when `agency` is used in the next 2 subsequent function calls. Maybe throwing an exception to ensure the issue is clear, something like:
```suggestion
    agency = get_agency(db_session, agency_code)
    if agency is None:
    raise click.ClickException(
        f"Agency '{agency_code}' not found. Run `make db-seed-local-with-agencies` first."
    )
```

api/tests/lib/seed_local_soap_certificate.py
jakobpederson: Not anymore! (I pulled the method)

api/tests/lib/seed_local_soap_certificate.py
jakobpederson: Good catch, neither is true now

api/tests/lib/seed_local_soap_certificate.py
jakobpederson: 100% correct. I just forgot.

api/tests/lib/seed_local_soap_certificate.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+# Before running this command run `make db-seed-local-with-agencies`
+# This method creates a cert and key if it cannot find them in the cache
+# it then gets or creates an agency with an agency_code of 'SOAP'
+# then it creates an opportunity -> competition -> application -> application_submission
+# then it gets or creates a legacy _certificate for that agency and cert serial_number
+# it prints off the encoded cert that can be put into the Postman header
+# in order to hit the locally running instance
+# the command looks like `make seed-local-soap-certificate DIR_PATH="~/test/cache/" AGENCY_CODE="EPA"`
```
kkrug: I think your docs here stale following your recent changes.

api/tests/lib/seed_local_soap_certificate.py
```diff
@@ -0,0 +1,176 @@
...
+        submission = db_session.execute(select(ApplicationSubmission)).scalars().first()
+        if not submission:
+            raise click.ClickException(
+                "No Submissions not found. Run `make db-seed-local-with-agencies` first."
+            )
+        agency_code = (
+            agency_code
+            if agency_code
+            else submission.application.competition.opportunity.agency_code
+        )
```
kkrug: This is already inside `if not agency_code` so `agency_code` is always false.

api/tests/lib/seed_local_soap_certificate.py
chouinar: I think I'd avoid any `~` handling - if you run outside Docker than this probably breaks as they'd just be different paths.

If we do relative paths like `./certs/whatever` does that work for in Docker? If so, I'd change the example to that, and just do nothing. That way if someone wants to use absolute paths outside Docker, that works all the same. Basically, don't try to do anything in code as doing anything means needing to account for more scenarios than just telling the user a standard way.

jakobpederson: `./certs/whatever` -> `api/certs/whatever` without the `expanduser`. I'll pull it.

api/tests/lib/seed_local_soap_certificate.py
```diff
@@ -0,0 +1,172 @@
...
+
+    with open(path_crt) as f:
+        cert_text = f.read()
+
+    encoded = quote(cert_text, safe="")
+    print(
+        "\nFor local testing use this url in Postman: 'http://localhost:8080/grantsws-agency/services/v2/AgencyWebServicesSoapPort'"
+    )
+    print("\nCOPY AND PASTE THIS STRING INTO THE POSTMAN HEADER AS 'X-Amzn-Mtls-Clientcert'")
+    print(encoded)
```
chouinar, line 141: Just checking, when I run this, the output as some encoding oddities like `-----END%20CERTIFICATE-----%0A` - is that fine/expected from the quote above?

jakobpederson: Yes, that should be fine. That's expected.

Issue #8883, opened 2026-03-09
Application Form Happy Path E2E Test - SF-424


doug-s-nava:
### Summary

Create E2E spec file for a happy path completion test case for the SF-424 form, including:

* creating a new application
* accessing the form
* filling out all fields in the form with valid information
* saving the form
* validating that no validation errors are displayed

Follow patterns established in the previous happy path tests.

Note that the following opportunities can be used for this testing:
* staging: opportunity 39cf0a5c-5fed-40b4-8f46-5374101ae419 - https://staging.simpler.grants.gov/opportunity/39cf0a5c-5fed-40b4-8f46-5374101ae419
* local: opportunity c3c59562-a54f-4203-b0f6-98f2f0383481, or directly use the application set up at 892a6aa8-92b1-46c2-a26b-21a6aaab825d

### Acceptance criteria

- [ ] happy path playwright test and spec file are created for the SF-424 form, covering successful filling and saving of the form with no validation errors

PR #9123, opened 2026-03-18
Add happy-path SF424 E2E spec and all supporting files

aoysimmons-GH:
Includes:
- happy-path-sf424.spec.ts (main test)
- sf424-field-definitions.ts, sf424-fill-data.ts (fixtures)
- general-forms-filling.ts (shared form utility)
- Test-data.docx (file upload fixture)

## Summary

Fixes / Work for #8883

PR #9148, opened 2026-03-19
Aoy/8883 application form happy path e2 e test sf 424

aoysimmons-GH:
## Summary

Fixes / Work for #8883 

## Changes proposed

Includes:

happy-path-sf424.spec.ts (main test)
sf424-field-definitions.ts, sf424-fill-data.ts (fixtures)
general-forms-filling.ts (shared form utility)
Test-data.docx (file upload fixture)

### DIFF
--- a/frontend/tests/e2e/utils/forms/general-forms-filling.ts
+++ b/frontend/tests/e2e/utils/forms/general-forms-filling.ts
@@ -207,29 +207,36 @@ export async function fillField(
 
       const inputName = await locator.getAttribute("name");
       const inputId = await locator.getAttribute("id");
-      const fileName = data.split(/[/\\]/).pop() ?? data;
-      const fieldContainer = locator.locator(
-        "xpath=ancestor::*[contains(concat(' ', normalize-space(@class), ' '), ' usa-form-group ') or contains(concat(' ', normalize-space(@class), ' '), ' simpler-formgroup ')][1]",
-      );
-
-      await locator.setInputFiles(data);
-
-      await fieldContainer
-        .locator("span")
-        .filter({ hasText: fileName })
-        .first()
-        .waitFor({ state: "visible", timeout: 30000 });
-
       const hiddenInputSelector = inputName
         ? `input[type="hidden"][name="${inputName}"]`
         : inputId
           ? `input[type="hidden"][name="${inputId}"], input[type="hidden"]#${inputId}`
           : null;
 
+      await locator.setInputFiles(data);
+
+      const fileName = data.split(/[/\\]/).pop() ?? data;
+
+      // Wait for the uploaded filename to appear in the UI before proceeding.
+      // Webkit renders the post-upload filename span more slowly, so use a
+      // generous timeout matching the file-input wait above.
       if (hiddenInputSelector) {
-        // Wait for the uploaded filename to appear in the UI before proceeding.
-        // Webkit renders the post-upload filename span more slowly, so use a
-        // generous timeout matching the file-input wait above.
+        await page
+          .locator(hiddenInputSelector)
+          .locator(
+            "xpath=ancestor::*[contains(concat(' ', normalize-space(@class), ' '), ' usa-form-group ') or contains(concat(' ', normalize-space(@class), ' '), ' simpler-formgroup ')][1]",
+          )
+          .locator("span")
+          .filter({ hasText: fileName })
+          .first()
+          .waitFor({ state: "visible", timeout: 30000 });
+      } else {
+        await page
+          .locator(`span:has-text("${fileName}")`)
+          .waitFor({ state: "visible", timeout: 30000 });
+      }
+
+      if (hiddenInputSelector) {
         await page.waitForFunction(
           ({ selector, uploadedFileName }) => {
             const hiddenInput =


frontend/tests/e2e/apply/happy-path-sf424.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+
+const { testOrgLabel } = playwrightEnv;
+const OPPORTUNITY_ID = "c3c59562-a54f-4203-b0f6-98f2f0383481";
+const OPPORTUNITY_URL = `/opportunity/${OPPORTUNITY_ID}`;
+
+test("Application form completion happy path - SF424", async ({
+  page,
+  context,
+}: { page: Page; context: BrowserContext }, testInfo: TestInfo) => {
+  test.setTimeout(300_000); // 5 min timeout
```
doug-s-nava: should add functionality skip all but chrome when running in staging https://github.com/HHS/simpler-grants-gov/blob/f632be4758e13a667297256c8b38d66634109503/frontend/tests/e2e/apply/happy-path-sfLLL.spec.ts#L17

aoysimmons-GH:
Reference old PR before I changes the branch name: https://github.com/HHS/simpler-grants-gov/pull/9123

frontend/tests/e2e/utils/forms/general-forms-filling.ts
```diff
@@ -1,10 +1,22 @@
+import path from "path";
 import { Page, TestInfo } from "@playwright/test";
 import { selectDropdownByValueOrLabel } from "tests/e2e/utils/select-dropdown-utils";
 
 export interface FillFieldDefinition {
   testId?: string;
   selector?: string;
-  type: "text" | "dropdown";
+  optionTestIdPrefix?: string;
+  hasTextRegex?: string;
```
aoysimmons-GH, line 10: @doug-s-nava comments: it doesn't look like this is being used, what is the use case for this?

aoysimmons-GH: These fields are used for advanced form interactions: optionTestIdPrefix customizes the test ID prefix for combo box options, and hasTextRegex allows filtering radio buttons or checkboxes by matching their text with a regex. They provide flexibility for handling non-standard or dynamic form elements.

frontend/tests/e2e/utils/forms/general-forms-filling.ts
```diff
@@ -1,10 +1,22 @@
...
+  getByText?: string;
+  textExact?: boolean;
+  buttonName?: string;
+  type:
+    | "text"
+    | "dropdown"
+    | "file"
+    | "radiobutton"
+    | "checkbox"
+    | "combo-box-input";
```
aoysimmons-GH, line 24: @doug-s-nava  comments: can we commit to a convention here? I'm fine with anything but let's make sure we're committed, ex:
```suggestion
    "radio-button" | "combo-box-input";
    OR
    "radioButton" | "comboBoxInput";
    OR
    "radio_button" | "combo_box_input";
```

frontend/tests/e2e/utils/forms/general-forms-filling.ts
```diff
@@ -25,24 +37,142 @@ export interface FormsFixtureData {
   fields: FillFieldDefinition[];
 }
 
+function shouldActivateField(data: string | boolean | undefined): boolean {
```
aoysimmons-GH, line 47: @doug-s-nava comments: can you add a comment describing how this function should be used? I figured out that it's meant to determine if a radio or checkbox should be checked, but that's not entirely clear

frontend/tests/e2e/utils/forms/general-forms-filling.ts
```diff
@@ -25,24 +37,142 @@ export interface FormsFixtureData {
...
       await locator.fill(data);
+    } else if (
+      field.type === "file" &&
+      field.buttonName &&
+      typeof data === "string"
+    ) {
+      const locator = page.getByRole("button", { name: field.buttonName });
+      await locator.waitFor({ state: "visible", timeout: 5000 });
+
+      const absolutePath = path.isAbsolute(data)
```
aoysimmons-GH: @doug-s-nava comments:
can we commit to always expecting either:
* an absolute path
* a specific hardcoded path, so that we can only supply the file name?

frontend/tests/e2e/utils/forms/general-forms-filling.ts
aoysimmons-GH: comment added

frontend/tests/e2e/utils/forms/general-forms-filling.ts
aoysimmons-GH: Yes, we can commit to always expecting either an absolute path or a specific hardcoded directory path, so that only the file name needs to be supplied. This will simplify file handling and make the code more predictable.

frontend/tests/e2e/utils/forms/general-forms-filling.ts
```diff
@@ -27,23 +44,151 @@ export interface FormsFixtureData {
...
+        field.type === "text" ||
+        field.type === "file") &&
+      typeof data !== "string"
+    ) {
+      throw new Error(
+        `Field ${fieldIdentifier} requires string data, received ${typeof data}`,
+      );
+    }
+
+    if (field.type === "dropdown" && typeof data === "string") {
```
doug-s-nava: [nit] you shouldn't need to check for `typeof data === "string"` here - if this was a drop down and a non-string value, this would have thrown an error on line 80

frontend/tests/e2e/utils/forms/general-forms-filling.ts
```diff
@@ -27,23 +44,151 @@ export interface FormsFixtureData {
...
+      if (field.selector) {
+        await selectDropdownByValueOrLabel(page, field.selector, data);
+      } else if (field.testId) {
+        const resolvedTestId = field.testId.includes("{value}")
```
doug-s-nava: whats the case where you'd include `{value}` in the test id? It doesn't seem like that's being done in this test unless I'm missing something

frontend/tests/e2e/utils/forms/general-forms-filling.ts
```diff
@@ -27,23 +44,151 @@ export interface FormsFixtureData {
...
+        await locator.click();
+      } else {
+        throw new Error(
+          `Dropdown field ${fieldIdentifier} is missing selector/testId`,
+        );
+      }
+    } else if (
+      field.type === "combo-box-input" &&
+      field.testId &&
+      typeof data === "string"
```
doug-s-nava: same as above, shouldn't have to check for data type here

frontend/tests/e2e/utils/forms/general-forms-filling.ts
```diff
@@ -27,40 +44,155 @@ export interface FormsFixtureData {
...
     } else if (field.type === "file" && (field.testId || field.selector)) {
+      if (typeof data !== "string") {
+        throw new Error(
+          `File field ${fieldIdentifier} requires string data (file path), received ${typeof data}`,
+        );
+      }
       const locator = field.selector
         ? page.locator(field.selector)
         : page.getByTestId(field.testId!);
-      // Use a generous timeout: Mobile Chrome renders the file input more slowly
```
Bhavna-Ramachandran, line 50: Please dont remove annotations. They are there as a note to self in the future

frontend/tests/e2e/utils/forms/general-forms-filling.ts
```diff
@@ -27,40 +44,155 @@ export interface FormsFixtureData {
...
-      // than desktop Chrome, so 5000ms is insufficient.
       await locator.waitFor({ state: "attached", timeout: 30000 });
       await locator.scrollIntoViewIfNeeded();
       await locator.setInputFiles(data);
-      // Wait for the uploaded filename to appear in the UI before proceeding.
```
Bhavna-Ramachandran, line 55: Same as previous annotations comment. Please dont remove them.

frontend/tests/e2e/utils/forms/general-forms-filling.ts
```diff
@@ -27,40 +44,155 @@ export interface FormsFixtureData {
...
-      // Webkit renders the post-upload filename span more slowly, so use a
-      // generous timeout matching the file-input wait above.
-      const fileName = data.split("/").pop() ?? data;
```
Bhavna-Ramachandran, line 58: This full block is needed for all the existing specs that involves file upload. Removing this will introduce a race condition and will lead to subsequent steps to fire before the upload is processed. I have seen this happen in CI, Webkit and Mobile Chrome.

frontend/tests/e2e/utils/forms/general-forms-filling.ts
```diff
@@ -6,7 +7,23 @@ import { getFormLink } from "./form-navigation-utils";
 export interface FillFieldDefinition {
   testId?: string;
   selector?: string;
-  type: "text" | "dropdown" | "file";
+  optionTestIdPrefix?: string;
+  hasTextRegex?: string;
+  getByText?: string;
+  textExact?: boolean;
+  buttonName?: string;
+  type:
```
Bhavna-Ramachandran, line 18: I see fillField logic only handles 'radiobutton' and 'combo-box-input', other aliases are dead code. Do we see a need for all other aliases?

Bhavna-Ramachandran: Can we just keep it simple and clean by using just one canonical name per type (in case of issues, this will be easier for us to debug):

"dropdown" | "text" | "checkbox" | "radio" | "combobox" | "file"

frontend/tests/e2e/utils/forms/general-forms-filling.ts
```diff
@@ -27,40 +44,155 @@ export interface FormsFixtureData {
...
+        await selectDropdownByValueOrLabel(page, field.selector, dropdownValue);
+      } else if (field.testId) {
+        const locator = page.getByTestId(`${field.testId}${dropdownValue}`);
+        await locator.waitFor({ state: "visible", timeout: 5000 });
+        await locator.click();
+      } else {
+        throw new Error(
+          `Dropdown field ${fieldIdentifier} is missing selector/testId`,
+        );
+      }
```
Bhavna-Ramachandran, line 115: We can do this block in the below way (I have done this way in #9317) , it groups by type + makes the logic easier to follow + it cleanly supports multiple locator strategies within the same field type.

We can apply this same pattern across other field types as well, which helps keep the implementation consistent and easier to maintain.

Note: What you have looks good. I'm just suggesting this to make the logic easier to follow which in turn will make things easier for us when we need to debug or extend this in the future.

if (field.type === "dropdown") {
  // Validate data type
  if (typeof data !== "string") {
    throw new Error(
      `Dropdown field ${fieldIdentifier} requires string data, received ${typeof data}`,
    );
  }

// Handle selector-based dropdown (native <select>)
   if (field.selector) {
    await selectDropdownByValueOrLabel(page, field.selector, data);
    return;
  }
  
  // Handle testId-based dropdown (custom component)
  if (field.testId) {
    const locator = page.getByTestId(`${field.testId}${data}`);
    await locator.waitFor({ state: "visible", timeout: 5000 });
    await locator.click();
    return;
  }

  // Fail fast if misconfigured
  throw new Error(
    `Dropdown field ${fieldIdentifier} is missing selector or testId`,
  );
}

frontend/tests/e2e/utils/forms/general-forms-filling.ts
```diff
@@ -27,40 +44,155 @@ export interface FormsFixtureData {
...
+      field.type === "file" &&
+      field.buttonName &&
+      field.field &&
+      typeof data === "string"
+    ) {
+      const locator = page.getByRole("button", { name: field.buttonName });
+      await locator.waitFor({ state: "visible", timeout: 5000 });
+
+      const filePath = path.resolve(__dirname, "../../test-upload-files", data);
+      await locator.setInputFiles(filePath);
```
Bhavna-Ramachandran: This needs to go into the fill-data fixture to be consistent with other form specs

frontend/tests/e2e/utils/forms/general-forms-filling.ts
aoysimmons-GH: The format changed it when we have more

frontend/tests/e2e/utils/forms/general-forms-filling.ts
aoysimmons-GH: I removed

frontend/tests/e2e/utils/forms/general-forms-filling.ts
Bhavna-Ramachandran: Sorry not sure why the formatting is not keeping up. It looked fine when I typed, i tried editing still not helping.

frontend/tests/e2e/utils/forms/general-forms-filling.ts
aoysimmons-GH: I added it

frontend/tests/e2e/utils/forms/general-forms-filling.ts
```diff
@@ -27,23 +43,142 @@ export interface FormsFixtureData {
...
+        field.type === "text" ||
+        field.type === "file") &&
+      typeof data !== "string"
+    ) {
+      throw new Error(
+        `Field ${fieldIdentifier} requires string data, received ${typeof data}`,
+      );
+    }
+
+    if (field.type === "dropdown") {
```
Bhavna-Ramachandran, line 95:
```suggestion

    if (field.type === "dropdown") {
  // Validate data type
  if (typeof data !== "string") {
    throw new Error(
      `Dropdown field ${fieldIdentifier} requires string data, received ${typeof data}`,
    );
  }

  // Handle selector-based dropdown (native <select>)
  if (field.selector) {
    await selectDropdownByValueOrLabel(page, field.selector, data);
    return;
  }

  // Handle testId-based dropdown (custom component)
  if (field.testId) {
    const locator = page.getByTestId(`${field.testId}${data}`);
    await locator.waitFor({ state: "visible", timeout: 5000 });
    await locator.click();
    return;
  }

  // Fail fast if misconfigured
  throw new Error(
    `Dropdown field ${fieldIdentifier} is missing selector or testId`,
  );
}

```

frontend/tests/e2e/utils/forms/general-forms-filling.ts
```diff
@@ -52,15 +199,20 @@ export async function fillField(
...
-      await page
-        .locator(`span:has-text("${fileName}")`)
-        .waitFor({ state: "visible", timeout: 30000 });
+      // Wait for the uploaded filename to appear for this specific field.
+      // The same file can be uploaded in multiple sections, so avoid a
+      // page-wide text query that can match more than one preview element.
+      const fileName = data.split(/[/\\]/).pop()!;
+      const fieldContainer = locator.locator(
+        "xpath=ancestor::*[starts-with(@id, 'form-section-')][1]",
+      );
```
Bhavna-Ramachandran:
```suggestion
      const fileName = data.split(/[/\\]/).pop() ?? data;
      await page
        .locator(`span:has-text("${fileName}")`)
        .waitFor({ state: "visible", timeout: 60000 });
    }
```

Bhavna-Ramachandran: This current block intent is good but the xpath is too fragile hence the form spec with file uploads were failing. Use the above suggestion it should work.

frontend/tests/e2e/utils/forms/general-forms-filling.ts
```diff
@@ -52,15 +199,20 @@ export async function fillField(
...
+      await fieldContainer
+        .getByText(fileName, { exact: true })
+        .waitFor({ state: "visible", timeout: 60000 });
```
Bhavna-Ramachandran: this block needs to go:
  await fieldContainer
        .getByText(fileName, { exact: true })
        .waitFor({ state: "visible", timeout: 60000 });

aoysimmons-GH:
@doug-s-nava @Bhavna-Ramachandran please let me know if you want me to update anything. I was passed in stg

frontend/tests/e2e/apply/fixtures/sf424-field-definitions.ts
```diff
@@ -0,0 +1,372 @@
...
+    section: "Section 11",
+    field: "Agency Name",
+  },
+  assistance_listing_program_title: {
+    testId: "assistance_listing_program_title",
+    type: "text",
+    section: "Section 12",
+    field: "Assistance Listing Program Title",
+  },
+  areas_affected_attachment: {
```
Bhavna-Ramachandran, line 192: This field defines both 'buttonname' and 'selector'. If we refer to the general-forms-fillings utils for these corresponding types -
- fillField has logic for both 'buttonname' and 'selector'
- 'buttonname' branch in 'fillField' does not have file upload logic, it just finds the button and waits.
- 'selector' branch in 'fillField' has the file upload logic. 
- Since 'buttonname' branch comes before 'selector' it will take precedence and the flow will skip actual upload logic -> making the test silently skip file upload.
If the UI requires just file upload we can use just the 'selector' and remove 'buttonname' from here or if UI requires both button click and upload - > update fillField to handle both

frontend/tests/e2e/utils/forms/general-forms-filling.ts
Bhavna-Ramachandran: Ah I see. I understand that formatting introduced inconsistencies. The issue is our logic only handles specific values (e.g. "radiobutton"), so other variants won’t work and may fail silently. 
It would be good to standardize on a single name and use it consistently. New specs can follow the same format as well.

frontend/tests/e2e/apply/fixtures/sf424-field-definitions.ts
```diff
@@ -0,0 +1,372 @@
...
+    section: "Section 14",
+    field: "Areas Affected Attachment",
+  },
+  project_title: {
+    testId: "project_title",
+    type: "text",
+    section: "Section 15",
+    field: "Project Title",
+  },
+  additional_project_title_attachment: {
```
Bhavna-Ramachandran, line 204: Same as above comment. This field defines both 'buttonname' and 'selector' and will have same issue. Hence if the UI requires just file upload we can use just the 'selector' and remove 'buttonname' from here or if UI requires both button click and upload - > update fillField to handle both

frontend/tests/e2e/apply/fixtures/sf424-fill-data.ts
```diff
@@ -0,0 +1,65 @@
...
+  project_end_date: "2036-10-31",
+  federal_estimated_funding: "1000000",
+  applicant_estimated_funding: "500000",
+  state_estimated_funding: "250000",
+  local_estimated_funding: "150000",
+  other_estimated_funding: "50000",
+  program_income_estimated_funding: "25000",
+  application_subject_to: "c. Program is not covered by E.O. 12372.",
+  state_review_available_date: "2026-12-01",
+  delinquent_federal_debt: "true",
```
Bhavna-Ramachandran: I see in UI we are actually selecting 'No'. Are we having 'True' here to tell the logic to activate or look for this field?

frontend/tests/e2e/apply/fixtures/sf424-field-definitions.ts
```diff
@@ -0,0 +1,372 @@
...
+    field:
+      "Is Application Subject to Review By State Under Executive Order 12372 Process?",
+  },
+  state_review_available_date: {
+    testId: "state_review_available_date",
+    type: "text",
+    section: "Section 19",
+    field: "State Review Available Date",
+  },
+  delinquent_federal_debt: {
```
Bhavna-Ramachandran, line 289: delinquent_federal_debt: true - the field is a radio button that selects 'No', but the label in the fixture says getByText: No with data: true meaning 'activate this'. That's confusing -> selecting "No" with shouldActivateField('true') works but reads as contradictory.

frontend/tests/e2e/utils/forms/general-forms-filling.ts
```diff
@@ -96,6 +245,7 @@ export async function fillForm(
   data: Record<string, string>,
```
Bhavna-Ramachandran, line 301: This current type forces all values to be strings, which is causing confusion mappings in some fields eg: delinquent_federal_debt: "true" - which actually selects the 'No' radio button. In other words fields that are inherently boolean (checkboxes,radios) using a string type hides intent and makes the test/fixtures logic hard to read.

This form has been tested and works both in local and staging, I see this as design clarity issue not a functional blocker.

Below are some of my suggestions. I don’t have full visibility on the task priority or intended direction, so I’ll let @doug-s-nava decide whether this specific issue can be approved as-is or if it needs further updates.
- The current type only allows strings, but some fields actually use booleans, so the type should be updated to support both string and boolean.
- The way fixture values map to UI actions is not always clear (for example, "true" ends up selecting "No"), so this mapping should be made consistent and easier to understand.
- These changes are needed to improve readability, clarity, and type safety across all form fields.

cc: @aoysimmons-GH

frontend/tests/e2e/apply/fixtures/sf424-field-definitions.ts
Bhavna-Ramachandran: This issue needs to be fixed. 

cc: @doug-s-nava @aoysimmons-GH

frontend/tests/e2e/apply/fixtures/sf424-field-definitions.ts
Bhavna-Ramachandran: This issue needs to be fixed. 

cc: @doug-s-nava @aoysimmons-GH

frontend/tests/e2e/utils/forms/general-forms-filling.ts
```diff
@@ -6,7 +6,23 @@ import { getFormLink } from "./form-navigation-utils";
...
+  optionTestIdPrefix?: string;
+  hasTextRegex?: string;
+  getByText?: string;
+  textExact?: boolean;
+  buttonName?: string;
+  type:
+    | "text"
+    | "dropdown"
+    | "file"
+    | "radiobutton"
```
doug-s-nava, line 22: is there a reason we need to support `radiobutton` `radioButton` and `radio_button`? Our logic seems to only be looking for `radiobutton` so there's potential confusion there

frontend/tests/e2e/utils/forms/general-forms-filling.ts
doug-s-nava: I see the awkwardness here but I'd like to move this PR forward and come back to it.

One other option for a fix might be, for each radio button group, to supply a string representing the label of the radio button to select, or `null` to keep the group unselected.

Can we make a ticket to improve that, and make any other non blocking improvements we see from this PR?

Bhavna-Ramachandran: Acknowledged.

frontend/tests/e2e/apply/fixtures/sf424-field-definitions.ts
aoysimmons-GH: I removed 4 buttonName

frontend/tests/e2e/apply/fixtures/sf424-field-definitions.ts
aoysimmons-GH: I removed 4 buttonName

frontend/tests/e2e/utils/forms/general-forms-filling.ts
aoysimmons-GH: I just see this comment. It is a bug. I will fix this.

frontend/tests/e2e/utils/forms/general-forms-filling.ts
aoysimmons-GH: Remove the duplicated.
Note: I added those types because during the record playwright captured those in different data id, type named in html but I think we better remove it.

Issue #8973, opened 2026-03-11
[BE] Create endpoints for enabling/disabling organization saved opportunity notifications


freddieyebra:
### Summary

We want to add an endpoint to allow users to update their saved opportunity notification settings.  We will use the same endpoint to record their settings for them opportunities they save for themselves and the settings for the opportunities saved to organizations they area part of. 

```
POST {user_id}/saved-opportunities/notifications
{
   organization_id: guid (optional, if not present assume self)
   email_enabled: true/false
}
```

If the entry isn't present in `user_saved_opportunities_notifications` it should be added with appropriate timestamps; if the entry is present it should be updated email_enabled to the value sent in and updated_at timestamp to the current time.

If organization_id is not specified, then the notification setting refers to opportunities they have saved themselves rather than those shared with their organization.

Log message for successfully adding/modifying their saved opportunity notifications:

```
logger.info("modified saved opportunity notification setting", extra={"organization_id": {organization_id}, "email_enabled":email_enabled)
```

This endpoint requires JWT authentication ensuring that the authenticated user matches the {user_id} in the URL.  If they don't, return a 401/Unauthorized.
 
If the organization_is is specified, then only set the notification if the user has the VIEW_ORG_SAVED_OPPORTUNITIES privilege for that organization.  If they don't, return a 401/Unauthorized.

...

### Acceptance criteria
- [ ] A new endpoint that handles 
   - [ ] The organization_id can be optional/null, which means we are setting the notification value for self
   - [ ] If the entry doesn't exist, create the entry (even if it is the same as the default value so we record that they explicitly set a preference)
   - [ ] If the entry does exist, update it with the new email_enabled value and updated timestamp
      - [ ] If the value of email_enabled is the same as the existing entry, just update the updated timestamp 
- [ ] Clear open API request schema documentation that organization_id is optional and if it is not included we are assuming the notification preference is set for self.
- [ ] A 401/Unauthorized should be returned if the authenticated user is not the same user as the {user_id} in the request.
- [ ] Only saved opportunity notification settings should be returned for organizations where the user has the VIEW_ORG_SAVED_OPPORTUNITIES privilege for that organization.
- [ ] Appropriate logging for successful operation
- [ ] Unit tests pass
...

joshtonava:
- We should probably include the response body in this ticket
- If organization_id is provided and does not exist : 404?
- if user isnt a member of that org? 403?
- validation for email_enabled missing?

PR #9155, opened 2026-03-19
[Issue #8973] Enable saved opportunities notifications endpoint

babebe:
## Summary

Fixes / Work for #8973  

## Changes proposed

New endpoint `POST /v1/users/:user_id/saved-opportunities/notifications`
New service `set_saved_opportunity_notification_settings` to create/update notification settings.
Required but nullable `organization_id` → defaults to user’s own saved opportunities if None.
Required bool `email_enabled` to set the notification preference
Access control: `VIEW_ORG_SAVED_OPPORTUNITIES` required for org-scoped settings.
Logging on successful create/update.
Unit tests covering self/org settings, unauthorized access, and non-existent organizations.

## Validation steps

Covered in test


api/src/api/users/user_schemas.py
```diff
@@ -844,3 +844,19 @@ class SavedOpportunityNotificationsSchema(Schema):
...
+
+
+class SetUserSavedOpportunityNotificationRequestSchema(Schema):
+    organization_id = fields.UUID(
+        required=False,
+        metadata={
+            "description": "The ID of the organization for which to set notification.If not provided, the setting applies to the user's own saved opportunities."
+        },
+    )
+    email_enabled = fields.Boolean(
```
mikehgrantsgov, line 857: Should this be marked `required=True`? The pydantic model has no default, so would `model_validate` raise a `ValidationError` rather than a clean 422 when the field is omitted? Might be worth testing. cc @chouinar if you can have a second pair of eyes on this.

api/tests/src/api/users/test_set_saved_opportunity_notification_settings.py
```diff
@@ -0,0 +1,211 @@
...
+    """
+    Test that a user cannot create or update notification settings for an organization
+    they do not have permission to access, resulting in a 403 Forbidden response.
+    """
+    user, org, token = create_user_in_org(db_session, privileges=[Privilege.VIEW_ORG_MEMBERSHIP])
+    org_id = org.organization_id
+
+    response = client.post(
+        f"/v1/users/{user.user_id}/saved-opportunities/notifications",
+        headers={"X-SGG-Token": user_auth_token},
```
mikehgrantsgov: I think this should be `token` and not `user_auth_token`. The 403 fires because the fixture user has no org membership, not because the privilege check correctly rejected an insufficient privilege. I would update this and confirm the test passes.

api/src/api/users/user_schemas.py
chouinar: Default is `required=False` so if you did this, it would cause the endpoint to 500 I think. Make it required.

I'd say for any endpoints, add tests that will hit request validation issues we expect to be safe / know that we need to adjust.

Another issue - Required != nullability - I think both of these fields should be required, but organization_id should be nullable (that way a user needs to explicitly say what behavior they want rather than assuming).

api/src/api/users/user_routes.py
```diff
@@ -845,3 +850,32 @@ def user_get_saved_opportunity_notifications(
 
     logger.info("Successfully fetched saved opportunity notification preferences")
     return response.ApiResponse(message="Success", data=result)
+
+
+@user_blueprint.post("/<uuid:user_id>/saved-opportunities/notifications")
+@user_blueprint.input(SetUserSavedOpportunityNotificationRequestSchema)
+@user_blueprint.output(SetUserSavedOpportunityNotificationResponseSchema)
+@user_blueprint.doc(responses=[200, 401, 403, 404])
+@user_blueprint.auth_required(api_jwt_auth)
```
chouinar: If there isn't a reason to not have a particular auth, can we default to having JWT + user api key auth on all our endpoints from now on?

api/tests/src/api/users/test_set_saved_opportunity_notification_settings.py
babebe: Yes! Updated.

api/src/api/users/user_schemas.py
babebe: Updated fields to be required. Added validation test.

api/src/api/users/user_routes.py
babebe: Sure. Updated auth.

api/src/api/users/user_routes.py
```diff
@@ -845,3 +850,34 @@ def user_get_saved_opportunity_notifications(
...
+    db_session: db.Session, user_id: UUID, json_data: dict
+) -> response.ApiResponse:
+    add_extra_data_to_current_request_logs(
+        {
+            "user_id": user_id,
+        }
+    )
+    logger.info("POST /v1/users/:user_id/saved-opportunities/notifications")
+
+    user_token_session: UserTokenSession = api_jwt_auth.get_user_token_session()
```
mikehgrantsgov: I think this needs to be: user = jwt_or_api_user_key_multi_auth.get_user()

The GET and POST endpoints are both decorated with @jwt_or_api_user_key_multi_auth.login_required and use jwt_or_api_user_key_security_schemes, meaning both should support JWT and API key auth. But as it's implemented this only works for JWT.

api/src/api/users/user_schemas.py
```diff
@@ -844,3 +844,20 @@ class SavedOpportunityNotificationsSchema(Schema):
...
 class UserSavedOpportunityNotificationsResponseSchema(AbstractResponseSchema):
     data = fields.Nested(SavedOpportunityNotificationsSchema)
+
+
+class SetUserSavedOpportunityNotificationRequestSchema(Schema):
+    organization_id = fields.UUID(
+        required=True,
+        allow_none=True,
+        metadata={
+            "description": "The ID of the organization for which to set notification.If not provided, the setting applies to the user's own saved opportunities."
```
mikehgrantsgov: I would add a space between the '.' and 'If not provided'

api/src/api/users/user_routes.py
babebe: Makes sense! Updated.

Issue #6671, opened 2025-10-08
Fetching the data from Simpler for UpdateApplicationInfo


chris-kuryak:
### Summary

Input Parameters for the endpoint:
GrantsGovTrackingNumber (String)
AssignAgencyTrackingNumber (String)
SaveAgencyNotes (String)

Required for testing

[Docs for existing SOAP endpoint 
](https://www.grants.gov/system-to-system/grantor-system-to-system/web-services/update-application-info)
### Acceptance criteria

- [ ] If you put in a Simpler tracking number to UpdateApplicationInfo, you get a response from Simpler
- [ ] Identify what backend changes need to be made to support this
- [ ] Implement the backend changes

ychoy:
This will be blocked by new tickets for creating the tables. @mdragon please add the relationship there when it's done.

PR #9174, opened 2026-03-19
[Issue #6671] Fetching the data from Simpler for UpdateApplicationInfo

kkrug:
## Summary

Fixes for #6671  

## Changes proposed

- Created update_application_info_response.py service
- Created test_update_application_info_response.py to test `update_application_info_response`
- Updated fault_messages.py to add 3 new fault messages
- Updated update_application_info_schemas.py to add `UpdateApplicationInfoResponseSOAPEnvelope`
- Updated grantors/schemas/__init__.py to add exports
- Updated grantors/services/__init__.py to add exports and imports
- Updated legacy_soap_api_client.py to add `update_application_info_request`
- Updated legacy_soap_api_utils.py to add `get_application_submission_by_legacy_tracking_number` function
- Updated confirm_application_delivery_response.py to use new `get_application_submission_by_legacy_tracking_number` utility
- Updated get_application_zip_response.py to use new `get_application_submission_by_legacy_tracking_number` utility
- Updated test_legacy_soap_client.py to account for refectoring changes

## Context for reviewers

This implements the Simpler backend for the UpdateApplicationInfo SOAP proxy endpoint. When a Simpler tracking number is passed to UpdateApplicationInfo, the API now looks up the corresponding `ApplicationSubmission` and performs the these operations:

- AssignAgencyTrackingNumber: Creates an `ApplicationSubmissionTrackingNumber` record (can only be assigned once per submission)
- SaveAgencyNotes: Creates an `ApplicationSubmissionNote` record (can be saved multiple times)
- Validates application status (must be ACCEPTED), certificate auth, and `LEGACY_AGENCY_ASSIGNER` privilege

## Validation steps

Confirm tests pass

### DIFF
--- a/api/src/legacy_soap_api/grantors/services/update_application_info_response.py
+++ b/api/src/legacy_soap_api/grantors/services/update_application_info_response.py
@@ -1,8 +1,6 @@
 import logging
 from typing import cast
 
-from sqlalchemy import select
-
 import src.adapters.db as db
 from src.auth.endpoint_access_util import can_access
 from src.constants.lookup_constants import ApplicationStatus
@@ -79,18 +77,7 @@ def _assign_agency_tracking_number(
     legacy_tracking_number: str,
     user: User,
 ) -> grantor_schemas.AssignAgencyTrackingNumberResult:
-    existing_tracking_number = (
-        db_session.execute(
-            select(ApplicationSubmissionTrackingNumber).where(
-                ApplicationSubmissionTrackingNumber.application_submission_id
-                == application_submission.application_submission_id,
-            )
-        )
-        .scalars()
-        .first()
-    )
-
-    if existing_tracking_number:
+    if application_submission.application_submission_tracking_numbers:
         logger.info(
             "Agency tracking number has already been assigned.",
             extra={
--- a/api/tests/src/legacy_soap_api/grantors/services/test_update_application_info_response.py
+++ b/api/tests/src/legacy_soap_api/grantors/services/test_update_application_info_response.py
@@ -74,7 +74,6 @@ class TestUpdateApplicationInfoResponse:
         agency = AgencyFactory.create()
         submission = _setup_submission(agency, ApplicationStatus.ACCEPTED)
         tracking_number = f"GRANT{submission.legacy_tracking_number}"
-        db_session.commit()
 
         _, _, soap_client_certificate = setup_cert_user(agency, {Privilege.LEGACY_AGENCY_ASSIGNER})
         soap_request = _make_soap_request(soap_client_certificate, tracking_number)
@@ -110,7 +109,6 @@ class TestUpdateApplicationInfoResponse:
         agency = AgencyFactory.create()
         submission = _setup_submission(agency, ApplicationStatus.ACCEPTED)
         tracking_number = f"GRANT{submission.legacy_tracking_number}"
-        db_session.commit()
 
         _, _, soap_client_certificate = setup_cert_user(agency, {Privilege.LEGACY_AGENCY_ASSIGNER})
         soap_request = _make_soap_request(soap_client_certificate, tracking_number)
@@ -148,7 +146,6 @@ class TestUpdateApplicationInfoResponse:
         agency = AgencyFactory.create()
         submission = _setup_submission(agency, ApplicationStatus.ACCEPTED)
         tracking_number = f"GRANT{submission.legacy_tracking_number}"
-        db_session.commit()
 
         _, _, soap_client_certificate = setup_cert_user(agency, {Privilege.LEGACY_AGENCY_ASSIGNER})
         soap_request = _make_soap_request(soap_client_certificate, tracking_number)
@@ -189,7 +186,6 @@ class TestUpdateApplicationInfoResponse:
 
     def test_submission_not_found_returns_fault(self, db_session, enable_factory_create):
         agency = AgencyFactory.create()
-        db_session.commit()
 
         _, _, soap_client_certificate = setup_cert_user(agency, {Privilege.LEGACY_AGENCY_ASSIGNER})
         tracking_number = "GRANT99999999"
@@ -213,7 +209,6 @@ class TestUpdateApplicationInfoResponse:
         agency = AgencyFactory.create()
         submission = _setup_submission(agency, ApplicationStatus.IN_PROGRESS)
         tracking_number = f"GRANT{submission.legacy_tracking_number}"
-        db_session.commit()
 
         _, _, soap_client_certificate = setup_cert_user(agency, {Privilege.LEGACY_AGENCY_ASSIGNER})
         soap_request = _make_soap_request(soap_client_certificate, tracking_number)
@@ -250,7 +245,6 @@ class TestUpdateApplicationInfoResponse:
             created_by_user=user,
             modified_by_user=user,
         )
-        db_session.commit()
 
         soap_request = _make_soap_request(soap_client_certificate, tracking_number)
 
@@ -283,7 +277,6 @@ class TestUpdateApplicationInfoResponse:
         agency = AgencyFactory.create()
         submission = _setup_submission(agency, ApplicationStatus.ACCEPTED)
         tracking_number = f"GRANT{submission.legacy_tracking_number}"
-        db_session.commit()
 
         _, _, soap_client_certificate = setup_cert_user(agency, {Privilege.LEGACY_AGENCY_ASSIGNER})
         soap_request = _make_soap_request(soap_client_certificate, tracking_number)
@@ -329,7 +322,6 @@ class TestUpdateApplicationInfoResponse:
         agency = AgencyFactory.create()
         submission = _setup_submission(agency, ApplicationStatus.ACCEPTED)
         tracking_number = f"GRANT{submission.legacy_tracking_number}"
-        db_session.commit()
 
         # Give user VIEWER privilege instead of ASSIGNER
         _, _, soap_client_certificate = setup_cert_user(agency, {Privilege.LEGACY_AGENCY_VIEWER})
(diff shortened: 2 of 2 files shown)


api/src/legacy_soap_api/grantors/services/update_application_info_response.py
```diff
@@ -0,0 +1,179 @@
...
+def update_application_info_response(
+    db_session: db.Session,
+    soap_request: SOAPRequest,
+    update_application_info_request: grantor_schemas.UpdateApplicationInfoRequest,
+    soap_config: SOAPOperationConfig,
+) -> grantor_schemas.UpdateApplicationInfoResponseSOAPEnvelope:
+    legacy_tracking_number = update_application_info_request.grants_gov_tracking_number
+
+    # grants_gov_tracking_number is validated as non-None by the pydantic schema
+    assert legacy_tracking_number is not None
```
chouinar: Don't use asserts outside of tests, if we know it's guaranteed to be non-null due to schema validation but mypy is complaining, do `cast(str, legacy_tracking_number)`. When actually running that does nothing, but it talks to the type checker.

api/src/legacy_soap_api/grantors/services/update_application_info_response.py
```diff
@@ -0,0 +1,179 @@
...
+
+    if legacy_tracking_number.startswith("GRANT"):
+        legacy_tracking_number = legacy_tracking_number.split("GRANT")[1]
+
+    application_submission = get_application_submission_by_legacy_tracking_number(
+        db_session, int(legacy_tracking_number)
+    )
```
chouinar: Since this is identical logic for each of the endpoints that takes in a tracking number, could we move this to a utility?

api/src/legacy_soap_api/grantors/services/update_application_info_response.py
```diff
@@ -0,0 +1,179 @@
...
+    validate_certificate,
+)
+from src.legacy_soap_api.legacy_soap_api_config import SOAPOperationConfig
+from src.legacy_soap_api.legacy_soap_api_constants import LegacySoapApiEvent
+from src.legacy_soap_api.legacy_soap_api_schemas import SOAPRequest
+from src.legacy_soap_api.legacy_soap_api_utils import SOAPFaultException
+
+logger = logging.getLogger(__name__)
+
+VALID_STATUSES_FOR_UPDATE = {ApplicationStatus.SUBMITTED, ApplicationStatus.ACCEPTED}
```
chouinar: I think we'd only want to allow it if it accepted. Since these statuses are used in our system for generating an application submission, the only way something could be in submitted AND have an application submission record is if:
* It was previously submitted and had reached accepted
* It was kicked back to in progress (we don't have a mechanism for this yet, but eventually will)
* It was resubmitted by the user
* The new submission hasn't been created yet

Which would mean we're about to make a new submission record - which would have a different tracking number. I'd probably err on the side of not allowing submitted here

api/src/legacy_soap_api/grantors/services/update_application_info_response.py
```diff
@@ -0,0 +1,179 @@
...
+            "User did not have permission to update application info",
+            extra={
+                "user_id": certificate.user.user_id,
+                "application_submission_id": application_submission.application_submission_id,
+                "privileges": soap_config.privileges,
+            },
+        )
+        raise SOAPClientUserDoesNotHavePermission(
+            "User did not have permission to update application info"
+        ) from e
```
chouinar: Don't catch the exception - use the function that doesn't throw an exception. That function is useful for all of our ordinary endpoints since we can just use raise_flask_error

```suggestion
      if not can_access(
          certificate.user,
          soap_config.privileges,
          application.competition.opportunity.agency_record,
      ):
        logger.info(
            "User did not have permission to update application info",
            extra={
                "user_id": certificate.user.user_id,
                "application_submission_id": application_submission.application_submission_id,
                "privileges": soap_config.privileges,
            },
        )
        raise SOAPClientUserDoesNotHavePermission(
            "User did not have permission to update application info"
        )
```

api/src/legacy_soap_api/grantors/services/update_application_info_response.py
```diff
@@ -0,0 +1,179 @@
...
+            modified_by_user_id=certificate.user.user_id,
+        )
+        db_session.add(note_record)
+
+        notes_result = grantor_schemas.SaveAgencyNotesResult(
+            success="true",
+        )
+
+    return grantor_schemas.UpdateApplicationInfoResponseSOAPEnvelope(
+        grants_gov_tracking_number=update_application_info_request.grants_gov_tracking_number,
```
chouinar, line 174: Could we split this function up a bit for organization/readability? It's ~140 lines long.

api/src/legacy_soap_api/grantors/services/update_application_info_response.py
```diff
@@ -0,0 +1,179 @@
...
+                extra={
+                    "soap_api_event": LegacySoapApiEvent.ERROR_CALLING_SIMPLER,
+                    "legacy_tracking_number": legacy_tracking_number,
+                    "response_operation_name": "UpdateApplicationInfoResponse",
+                },
+            )
+            raise SOAPFaultException(
+                "Agency tracking number has already been assigned",
+                fault=UpdateApplicationInfoTrackingNumberAlreadyAssigned,
+            )
```
chouinar: For the actual SOAP endpoint for this, if a request has both a tracking number + notes, and the tracking number already exists, will notes be added? If so, we'd not want an error here.

api/src/legacy_soap_api/grantors/services/update_application_info_response.py
```diff
@@ -0,0 +1,179 @@
...
+
+        tracking_number_record = ApplicationSubmissionTrackingNumber(
+            application_submission_id=application_submission.application_submission_id,
+            tracking_number=update_application_info_request.assign_agency_tracking_number,
+            created_by_user_id=certificate.user.user_id,
+            modified_by_user_id=certificate.user.user_id,
+        )
```
chouinar: Use the relationships instead of IDs - it's shorter / it sets the IDs / if any code needs to reference the relationship, it'll work without any flushes

```suggestion
        tracking_number_record = ApplicationSubmissionTrackingNumber(
            application_submission=application_submission,
            tracking_number=update_application_info_request.assign_agency_tracking_number,
            created_by_user=certificate.user,
            modified_by_user=certificate.user,
        )
```

api/src/legacy_soap_api/grantors/services/update_application_info_response.py
```diff
@@ -0,0 +1,179 @@
...
+
+    # Save agency notes if provided
+    if update_application_info_request.save_agency_notes is not None:
+        note_record = ApplicationSubmissionNote(
+            application_submission_id=application_submission.application_submission_id,
+            note=update_application_info_request.save_agency_notes,
+            created_by_user_id=certificate.user.user_id,
+            modified_by_user_id=certificate.user.user_id,
+        )
```
chouinar:
```suggestion
        note_record = ApplicationSubmissionNote(
            application_submission=application_submission,
            note=update_application_info_request.save_agency_notes,
            created_by_user=certificate.user,
            modified_by_user=certificate.user,
        )
```

api/src/legacy_soap_api/grantors/services/update_application_info_response.py
kkrug: Fixed, also made a similar change in confirm_application_delivery_response.py.

api/src/legacy_soap_api/grantors/services/update_application_info_response.py
kkrug: Utility made and other services updated

api/src/legacy_soap_api/grantors/services/update_application_info_response.py
kkrug: Fixed, and tests updated

api/src/legacy_soap_api/grantors/services/update_application_info_response.py
kkrug: Updated

api/src/legacy_soap_api/grantors/services/update_application_info_response.py
kkrug: Refactored this function

api/src/legacy_soap_api/grantors/services/update_application_info_response.py
kkrug: You're correct, updated

api/src/legacy_soap_api/grantors/services/update_application_info_response.py
kkrug: Updated

api/src/legacy_soap_api/grantors/services/update_application_info_response.py
kkrug: Updated

api/src/legacy_soap_api/grantors/services/update_application_info_response.py
```diff
@@ -0,0 +1,191 @@
...
+            select(ApplicationSubmissionTrackingNumber).where(
+                ApplicationSubmissionTrackingNumber.application_submission_id
+                == application_submission.application_submission_id,
+            )
+        )
+        .scalars()
+        .first()
+    )
+
+    if existing_tracking_number:
```
chouinar: Can we use the relationship for this?

```suggestion
    if application_submission.application_submission_tracking_numbers:
```

api/tests/src/legacy_soap_api/grantors/services/test_update_application_info_response.py
```diff
@@ -0,0 +1,400 @@
...
+            .all()
+        )
+        assert len(tracking_numbers) == 1
+        assert tracking_numbers[0].tracking_number == "AGENCY-123"
+
+    def test_successful_save_agency_notes(self, db_session, enable_factory_create):
+        agency = AgencyFactory.create()
+        submission = _setup_submission(agency, ApplicationStatus.ACCEPTED)
+        tracking_number = f"GRANT{submission.legacy_tracking_number}"
+        db_session.commit()
```
chouinar: In these tests, we shouldn't need this commit - the factories do a commit at the end of the create function.

api/src/legacy_soap_api/grantors/services/update_application_info_response.py
kkrug: Updated

api/tests/src/legacy_soap_api/grantors/services/test_update_application_info_response.py
kkrug: Fixed

Issue #8669, opened 2026-02-26
Create GET award recommendation endpoint


kevlyu:
### Summary

See [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Create-the-GET-award-recommendation-endpoint)

`GET /alpha/award-recommendations/:award_recommendation_id`

See [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Award-Recommendation.1) for JSON response

Note that this endpoint should be done first as a lot of endpoints will reuse the pieces built in it including:
- The schema (half the other endpoints return the same/a subset of the response schema)
- The logic to fetch an award recommendation and 404 if not found (literally every endpoint does this as its first bit of logic)

### Acceptance criteria

- [ ] Endpoint Created
- [ ] Tests Created

PR #9176, opened 2026-03-20
[Issue #8669] Create Get Award Recommendation endpoint

nathan-stricker:
## Summary

Work for #8669 

## Changes proposed

Create new endpoint to get an Award Recommendation

## Context for reviewers

See [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Award-Recommendation.1) for endpoint specifications.

## Validation steps

- [ ] Confirm implementation aligns with specifications outlined in the [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Award-Recommendation.1)
- [ ] Review created tests


api/src/api/award_recommendations_alpha/award_recommendation_schemas.py
```diff
@@ -0,0 +1,152 @@
...
+
+    opportunity_id = fields.UUID(metadata={"description": "The opportunity ID"})
+    opportunity_number = fields.String(
+        allow_none=True,
+        metadata={"description": "The opportunity number"},
+    )
+    opportunity_title = fields.String(
+        allow_none=True,
+        metadata={"description": "The opportunity title"},
+    )
```
chouinar, line 47: For text fields, I'd recommend adding examples since the default isn't too helpful. Most other types (including enums, uuids, dates, and other string fields that have some format restriction) can leave them out as they get defaults that are reasonable.

api/src/api/award_recommendations_alpha/award_recommendation_schemas.py
```diff
@@ -0,0 +1,152 @@
...
+    summary = fields.Nested(
+        AwardRecommendationOpportunitySummarySchema,
+        allow_none=True,
+        metadata={"description": "Summary details of the opportunity"},
+    )
+
+
+class AwardRecommendationUploadingUserSchema(Schema):
```
chouinar: We have a SimpleUserSchema that we should use for this

api/src/app.py
```diff
@@ -195,6 +195,8 @@ def register_blueprints(app: APIFlask) -> None:
     app.register_blueprint(form_blueprint)
     app.register_blueprint(competition_blueprint)
 
+    app.register_blueprint(award_recommendation_blueprint)
```
chouinar: We should put this behind a feature flag - otherwise these endpoints will be available in prod.

Just add something to the endpoint_config like `enable_award_recommendation_endpoints` - default it to False, and then set it to true in our local.env + dev/staging terraform configs. Basically, the same places that I have it for the workflow endpoint env var of a similar purpose: https://github.com/search?q=repo%3AHHS%2Fsimpler-grants-gov%20enable_workflow_endpoint&type=code

api/tests/src/api/award_recommendations/test_award_recommendation_get.py
```diff
@@ -0,0 +1,258 @@
...
+        assert data["other_key_information"] is None
+        assert data["review_workflow_id"] is None
+
+
+####################################
+# 404 Tests
+####################################
+
+
+class TestGetAwardRecommendation404:
```
chouinar, line 156: Y'know, I hadn't ever thought about organizing the tests like this, but I like it.

api/src/api/award_recommendations_alpha/award_recommendation_routes.py
```diff
@@ -0,0 +1,48 @@
...
+    responses=[200, 401, 403, 404],
+)
+@jwt_or_api_user_key_multi_auth.login_required
+@flask_db.with_db_session()
+def award_recommendation_get(
+    db_session: db.Session, award_recommendation_id: uuid.UUID
+) -> response.ApiResponse:
+    add_extra_data_to_current_request_logs(
+        {"award_recommendation_id": str(award_recommendation_id)}
+    )
```
chouinar: For anything with our logging in the extra param (which `add_extra_data_to_current_request_logs` is doing), you can avoid converting most common types to strings - https://github.com/HHS/simpler-grants-gov/blob/main/api/src/util/json_util.py - dates get isoformatted, and if we don't have a conversion we do `str(value)` anyways.

```suggestion
    add_extra_data_to_current_request_logs(
        {"award_recommendation_id": award_recommendation_id}
    )
```

nathan-stricker:
@chouinar thanks for the first review, all of your comments should be addressd. Curious, I saw [this thread](https://betagrantsgov.slack.com/archives/C05TSL64VUH/p1774297229693999?thread_ts=1774278194.991569&cid=C05TSL64VUH) about the failing tfsec check- any suggestions on best next steps?

api/src/api/award_recommendations_alpha/award_recommendation_schemas.py
chouinar: Could you add some examples to the other text fields?

chouinar:
@nathan-stricker - something went weird with your merges from main, you've got a lot of changes from other branches it looks like

nathan-stricker:
@chouinar something weird happened with my rebase for the Trivy issue- should be fixed now

PR #9183, opened 2026-03-20
[Issue 9005] Standardize case and suffix of page titles

### LINKED ISSUE #9005: Make page titles case consistent
### Summary

In a recent PR, @btabaska [noted in a comment](https://github.com/HHS/simpler-grants-gov/pull/9000#discussion_r2924459655) how our use of sentence/title case in page headings (the text that shows in browser tabs) is inconsistent with our decision to use "sentence case for headings, in the site navigation, and for buttons. Our page titles are using title case. 

### Acceptance criteria

- [ ] Design guidance clarifies intent 
- [ ] Page titles have be updated to be consistent with design guidance

andycochran:
## Summary

Fixes #9005 

## Changes proposed

- Standardize all page titles to sentence case (some were using title case)
- Drop some unnecessary word in titles, e.g.: 
  - `Search Funding Opportunities` => `Search`
  - `Workspace Dashboard` => `Workspace`
  - `Application landing page` => `Application`
- Append `" | Simpler.Grants.gov"` to all page titles (some had it, some didn't) 
  - _There might be a better way of doing this globally for all pages (except the index) without repeating it in every content string. Maybe revisit that later._ 
- Fix the `/applications` page's `<h1>` to use `pageHeading` appropriately 
  - _It was using the `pageTitle` (which would've added the " | Simpler.Grants.gov" suffix in the page content)_ 
- Make developers page consistent: 
  - Drop "portal" from the `pageTitle` and make "Developers" plural
  - ~Change the route to plural `/developers`~ <= will make a separate ticket for this
  - Add `<h1>Developers</h1>` to the top of page (following the pattern of other pages)
- Fix margins on the `/api-dashboard` page

## Context for reviewers

This change affects the `<title>` tag contents, what content shows in the browser tab not e.g. the `<h1>` (tho a couple h1s were changed too). 

I noticed the application form page title had no fallback, so I added an `else` condition. Not sure if this is necessary, or if it's possible for `formName` to be null. But this fallback is similar to the "Application" page title, which does not show the actual application name. 

## Validation steps

- Navigate through all pages of the site
- Look at the browser tab to read the page titles 
- They should all be sentence case

### DIFF
--- a/frontend/src/app/[locale]/(base)/developer/page.tsx
+++ b/frontend/src/app/[locale]/(base)/developer/page.tsx
@@ -1,6 +1,7 @@
 import { Metadata } from "next";
 import { LocalizedPageProps } from "src/types/intl";
 
+import { useTranslations } from "next-intl";
 import { getTranslations } from "next-intl/server";
 import { GridContainer } from "@trussworks/react-uswds";
 
@@ -17,9 +18,10 @@ export async function generateMetadata({ params }: LocalizedPageProps) {
 }
 
 export default function Developer() {
+  const t = useTranslations("Developer");
   return (
     <GridContainer className="padding-y-4 grid-container tablet-lg:padding-y-6">
-      <h1 className="margin-bottom-5">Developers</h1>
+      <h1 className="margin-bottom-5">{t("h1")}</h1>
       <DeveloperPageSections />
     </GridContainer>
   );


frontend/src/i18n/messages/en/index.ts
```diff
@@ -602,7 +602,12 @@ export const messages = {
     },
   },
   Developer: {
-    pageTitle: "Developer Portal | Simpler.Grants.gov",
+    // TODO: Make these consistent.
+    // - menu item is plural "Developers"
+    // - route is singular "/developer"
+    // - page title is "Developer portal"
+    // - page has no `h1` tag (infoTitle is an `h2`)
+    pageTitle: "Developer portal | Simpler.Grants.gov",
```
andycochran: @btabaska what should we do here?

andycochran: Talked to the Design team about this at crit. Decision: 
- Drop "portal" from the `pageTitle` and make "Developers" plural
- Change the route to plural `/developers`
- Add `<h1>Developers</h1>` to the top of page (following the pattern of other pages)

frontend/src/app/[locale]/(base)/applications/[applicationId]/page.tsx
```diff
@@ -25,7 +25,7 @@ export const dynamic = "force-dynamic";
 
 export function generateMetadata() {
   const meta: Metadata = {
-    title: `Application landing page`,
+    title: `Application | Simpler.Grants.gov`,
```
Comment (mdragon) [inline frontend/src/app/[locale]/(base)/applications/[applicationId]/page.tsx:28]: Longer term, would there be value in this having their Application name instead? Doesn't have to happen in this ticket, just longer term thought, what if I have 3 Applications going, how could we distinguish the tabs by title.

Comment (andycochran) [reply frontend/src/app/[locale]/(base)/applications/[applicationId]/page.tsx:28]: Yeah. I dipped my toe into that but it was complicating this change quite a lot. The Apply workstream can do that in another ticket?

Comment (andycochran) [reply frontend/src/app/[locale]/(base)/applications/[applicationId]/page.tsx:28]: There's an argument to be made for keeping these shorter.

PR #9185, opened 2026-03-20
[Issue #8763] Add integration tests for workflow management

### LINKED ISSUE #8763: Create full integration tests for workflow
### Summary

We want to have tests that verify our workflow processes work end-to-end.

That is, we want to send a series of events to process a workflow from start to finish. This will mean repeating the pattern of sending an event and running the event service logic for each event.

These tests mainly focus on making sure everything is connected together, and serve as regression tests as we change and modify this logic going forward. Things we should test:
* Happy path(s)
* Verify certain events can't be sent/processed are certain times (validation blocks it)

We do not need to focus on every possible scenario - we have a lot of very specific tests for that, we want to focus on tests that verify the API and service layer work together. Largely, the API should prevent most bad scenarios from happening in the service layer.

### Acceptance criteria

- [ ] Tests setup - should be against our basic test workflow (not a real one)
- [ ] Established patterns for how we might do this for future workflows, if needed

chouinar:
## Summary
Fixes #8763

## Changes proposed
* adds integration tests for workflow management to verify API + service work together

## Context for reviewers
These tests mainly exist so we know that the event API + workflow service work together. We aren't testing every possible scenario, we have a lot of tests for more specific scenarios - these tests exist to make sure if changes are made that would cause the SQS connection to have issues, we'll detect it.

## Validation steps
They're tests - they are their own validation.

### DIFF
--- a/api/tests/workflow/integration/test_workflow_integration.py
+++ b/api/tests/workflow/integration/test_workflow_integration.py
@@ -67,14 +67,31 @@ def build_process_workflow_request(
     return payload
 
 
-def send_event_and_process(client, app, payload: dict, user_jwt: str) -> str:
+def send_event_and_process(
+    client,
+    app,
+    payload: dict,
+    user_jwt: str,
+    is_successful_in_api: bool = True,
+    has_service_messages: bool = True,
+) -> str:
     response = client.put("/v1/workflows/events", json=payload, headers={"X-SGG-Token": user_jwt})
-    assert response.status_code == 200
-    event_id = response.get_json()["data"]["event_id"]
+    if is_successful_in_api:
+        assert response.status_code == 200
+        event_id = response.get_json()["data"]["event_id"]
+    else:
+        assert response.status_code >= 400
+        event_id = None
 
     with app.app_context():
         messages_to_delete, messages_to_keep = WorkflowManager().process_batch()
-        assert len(messages_to_delete) == 1
+
+        if has_service_messages:
+            assert len(messages_to_delete) == 1
+            assert len(messages_to_keep) == 0
+        else:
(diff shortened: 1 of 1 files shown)


api/tests/workflow/integration/test_workflow_integration.py
```diff
@@ -0,0 +1,305 @@
...
+
+
+def send_event_and_process(client, app, payload: dict, user_jwt: str) -> str:
+    response = client.put("/v1/workflows/events", json=payload, headers={"X-SGG-Token": user_jwt})
+    assert response.status_code == 200
+    event_id = response.get_json()["data"]["event_id"]
+
+    with app.app_context():
+        messages_to_delete, messages_to_keep = WorkflowManager().process_batch()
+        assert len(messages_to_delete) == 1
```
kkrug: Should we also assert here that `messages_to_keep` is empty?

```suggestion
        assert len(messages_to_delete) == 1
        assert len(messages_to_keep) == 0
```

kkrug (commented):
Should we add some tests for validation? I know we do a lot of this at the service layers, but the ticket specified "Verify certain events can't be sent/processed at certain times (validation blocks it)." Perhaps also a test for the `DECLINED` state.

chouinar:
> Should we add some tests for validation? I know we do a lot of this at the service layers, but the ticket specified "Verify certain events can't be sent/processed at certain times (validation blocks it)." Perhaps also a test for the `DECLINED` state.

I'll add a bit more, but didn't want to do too much on validation because if the API validation doesn't send something to SQS (which we verify in those tests) then of course the service won't do anything. But can add a check or two to be safe.

PR #9187, opened 2026-03-20
[Issue #9186] Set agency contact information when transforming forecasts

### LINKED ISSUE #9186: Forecasts don't have agency contact information
### Summary

Full context: https://betagrantsgov.slack.com/archives/C05TSL64VUH/p1773950533080269

Rough summary:
* Forecasts and synopsis have different ways they define agency contact information in grants.gov
* We only implemented support for the way synopsis work
* Easiest fix - copy `ac_name` + `ac_phone` to our agency contact description field, separated by a "\n".

### Reproducibility

### How to reproduce the bug

Look at any opportunity with a forecast, the agency contact description isn't set.

### Browser

Not applicable

### Code of Conduct

chouinar:
## Summary

Fixes #9186

## Changes proposed
* During the transformation of a forecast record, populate the agency contact description from the forecasts AC_NAME + AC_PHONE fields
 
## Context for reviewers
In grants.gov, the opportunity summary is actually two tables, TSYNOPSIS and TFORECAST - they mostly overlap, but in this case, only the synopsis has an agency contact description field. They both have these AC_NAME + AC_PHONE fields, but we very intentionally don't use those as they're often wrong (there was a prior escalation for this). But, forecasts do use these, this logic we've added it just mirroring what grants.gov does when rendering their frontend of grabbing these two fields as if they were the singular agency contact description - we're just doing it at the database level rather than the API rendering level.

## Validation steps
There's going to be a bit of manual testing in staging after I merge this as there's also going to need to be a backfill for these. But if you want to verify locally. Assuming you have your local API and frontend running, you can do:

* Run `make console`
* Create a forecast that we can transform by doing `f.StagingTforecastFactory.create(opportunity__opportunity_id=1111, opportunity__cfdas=[])` - if you want to create multiple, you'll need to change the opportunity ID for each.
* Run the transformation job by doing: `make cmd args="data-migration load-transform --no-load"`
* Navigate to http://localhost:3000/opportunity/1111 
* You should see the agency contact description populated for this:
Screenshot 2026-03-20 at 1 30 03 PM

Note that a separate minor issue exists in the frontend where new lines aren't handled for this field. If you look in the DB, it's there.

### DIFF
--- a/api/src/data_migration/transformation/transform_util.py
+++ b/api/src/data_migration/transformation/transform_util.py
@@ -279,7 +279,7 @@ def transform_opportunity_summary(
     # Grants.gov stores agency contact information differently for forecasts
     # Rather than a single field, they have 2 and when they go to display it
     # they put both of them. To keep things simpler, we'll put those two together
-    # here. If they're both null, we'll have an empty string.
+    # here. If they're both null, we'll set it to null to be consistent with synopsis.
     if source_summary.is_forecast:
         agency_name = getattr(source_summary, "ac_name", None)
         agency_phone = getattr(source_summary, "ac_phone", None)
@@ -290,7 +290,8 @@ def transform_opportunity_summary(
         if agency_phone is not None:
             values.append(agency_phone)
 
-        target_summary.agency_contact_description = "\n".join(values)
+        if len(values) > 0:
+            target_summary.agency_contact_description = "\n".join(values)
 
     transform_update_create_timestamp(source_summary, target_summary, log_extra=log_extra)
 


api/tests/src/data_migration/transformation/conftest.py
```diff
@@ -110,7 +110,6 @@ def setup_cfda(
 
 def setup_synopsis_forecast(
     is_forecast: bool,
-    revision_number: int | None,
```
chouinar, line 113: We should have cleaned this up a while ago, we never don't set this null / don't even make history records anymore. Just trimming it out while I'm here.

mdragon (dismissed):
Looks good. Looks like something else broke the tests.

api/src/data_migration/transformation/transform_util.py
```diff
@@ -276,6 +276,22 @@ def transform_opportunity_summary(
...
+        agency_name = getattr(source_summary, "ac_name", None)
+        agency_phone = getattr(source_summary, "ac_phone", None)
+
+        values = []
+        if agency_name is not None:
+            values.append(agency_name)
+        if agency_phone is not None:
+            values.append(agency_phone)
+
+        target_summary.agency_contact_description = "\n".join(values)
```
mikehgrantsgov: Does the frontend ever check for null? I noticed the agency_contact_description maps to `None` when the source field is null. Just checking if we ever use this anywhere that would try null-checking but the UI might get an empty string back.

chouinar:
> Looks good. Looks like something else broke the tests.

Was looking at it. Some older tests manually set agency codes or use an older factory generator for them which has a list of like 70 codes - we stopped using that, but there's a few spots it's still used. This is a newer test that failed, I think what happened is it just happened to randomly generate an agency code that looks like and overlapped with one of those 70.

api/src/data_migration/transformation/transform_util.py
chouinar: I double-checked locally. Null/empty both end up looking like:
Screenshot 2026-03-20 at 2 21 48 PM

Although the frontend looks like it does check for null by doing `{opValue ?? "--"}` which if I'm reading right should only put `--` for exactly null, not the empty string.

I guess to be safe I should make it null to be consistent.

PR #9195, opened 2026-03-20
[Issue #8842] Create an opportunity publish workflow

### LINKED ISSUE #8842: Create an opportunity publish workflow
### Summary

We want to create the workflow that will run as part of publishing an opportunity.

This workflow should have the following states at the moment:
* `start`
* `pending_publish` This state just exists for the start workflow event to put it into
* `draft_flag_flipped`
* `current_opportunity_summary_calculated`
* `opportunity_written_to_search`
* `end`

We will need the following transitions to be defined:
* `start_workflow`: `start` --> `pending_publish`
* `flip_is_draft`: `pending_publish` -> `draft_flag_flipped`
* `calculate_current_opportunity_summary`: `draft_flag_flipped` -> `current_opportunity_summary_calculated`
* `write_opportunity_to_search`: `current_opportunity_summary_calculated` -> `opportunity_wrote_to_search`
* `finish_publish`: `opportunity_written_to_search` -> `end`

Each of these transitions should be setup so that they have an `after` that calls the next transition. That is, if we call the `start_workflow` event, it'll in turn call the next 4 steps automatically and put it in the end state.

We do not need to add any behavior to any of these events yet, just getting things setup and defined. A later PR will handle making these steps do what they say.

### Configuration Details
* `workflow_type` - `opportunity_publish` - already a possible value
* `persistence_model_cls` - `OpportunityPersistenceModel`
* `entity_type` - Opportunity
* No approval mapping

### Acceptance criteria

- [ ] Setup workflow as defined above
- [ ] Create tests that verify the workflow will progress through all the states automatically after starting the workflow
- [ ] Verify we can send an event via the event API for this workflow
- [ ] Document the process for setting up the workflow as you go - we want to make sure other teams can build these in the future

chouinar:
## Summary
Fixes #8842

## Changes proposed
* Sets up an opportunity publish workflow - no functionality on transitions yet
* Add docs for how to create a workflow - some more specific details will come in a later ticket

## Context for reviewers
This sets up a workflow for opportunity publishing that handles automating a few things when an opportunity gets published, although none of that logic is connected yet.

## Validation steps
Run the API locally while running the workflow service with `make cmd args="workflow workflow-main"`.

Find an opportunity in your local DB and you can call the PUT workflow endpoint by setting that as the entity ID in a request like below using the `internal_workflow_user_key` user API key.

```json
{
  "start_workflow_context": {
    "entity_id": "1b0f1981-35c4-42d9-840f-bdc7054a7842",
    "entity_type": "opportunity",
  }
}
```

This'll immediately run through the whole workflow. If you look in the `api.workflow_event_history` table you can find the workflow ID using the event ID - you'll see just one event in this table, but 5 transitions in the workflow_audit table.

### DIFF
--- a/api/src/workflow/state_machine/initial_prototype_state_machine.py
+++ b/api/src/workflow/state_machine/initial_prototype_state_machine.py
@@ -51,7 +51,6 @@ class InitialPrototypeStateMachine(BaseStateMachine):
         InitialPrototypeState,
         initial=InitialPrototypeState.START,
         final=[InitialPrototypeState.END, InitialPrototypeState.DECLINED],
-        use_enum_instance=True,
     )
 
     ### Events + transitions
--- a/api/tests/workflow/state_machine/test_state_machines.py
+++ b/api/tests/workflow/state_machine/test_state_machines.py
@@ -65,7 +65,6 @@ class BasicTestStateMachine(BaseStateMachine):
         BasicState,
         initial=BasicState.START,
         final=[BasicState.END, BasicState.DECLINED],
-        use_enum_instance=True,
     )
 
     ### Events + transitions


api/src/workflow/state_machine/opportunity_publish_state_machine.py
```diff
@@ -0,0 +1,73 @@
...
+
+@WorkflowRegistry.register_workflow(opportunity_publish_state_machine_config)
+class OpportunityPublishStateMachine(BaseStateMachine):
+    ### States
+
+    states = States.from_enum(
+        OpportunityPublishState,
+        initial=OpportunityPublishState.START,
+        final=[OpportunityPublishState.END],
+    )
```
kkrug, line 40: Should we passing in `use_enum_instance=True` here like we do for `InitialPrototypeStateMachine` and `BasicTestStateMachine`? I don't think it matters much now but not sure if it could have any consequences down the road.

api/src/workflow/README.md
```diff
@@ -33,6 +33,119 @@ flowchart LR
...
+
+### Create the state machine
+
+```python
+from statemachine import Event
+from statemachine.states import States
+
+from src.workflow.event.state_machine_event import StateMachineEvent
+
+from src.workflow.base_state_machine import BaseStateMachine
```
kkrug, line 101: Minor nit but if someone was to copy/paste this in the future they would be missing `from src.workflow.registry.workflow_registry import WorkflowRegistry`

api/src/workflow/state_machine/opportunity_publish_state_machine.py
chouinar: In the new version of the state machine library, that became the default behavior - I set the others up before that change.

chouinar: Removed it from the others for consistency.

Issue #8950, opened 2026-03-10
Make Received by Agency and Tracking Number Applied "statuses" work without statuses in Simpler SOAP Notary


mdragon:
### Summary

Apply team is looking to separate out User facing statuses from system or internal statuses.

We can come back to "Submitted" vs "Accepted" but for now:
1. Replace "Agency Tracking Number Assigned" with a check for if any Agency Tracking Number records exist for the Submission, written by the API key user
2. Replace "Received by Agency" with a new table that tracks the API key that has received the submission. (This is a feature request for existing users of Grants.gov who sometimes want to ingest a submission into two different systems) ([This is done here](https://github.com/HHS/simpler-grants-gov/issues/8957))..
3. Not move the new statuses we were going to add forward and/or remove them if they've already merged ([This is done](https://github.com/HHS/simpler-grants-gov/pull/8793)).

Going forward these two would no longer be statuses on the submission record itself, that would stay "Accepted" but we would use these other data points to allow for filtering by the existing legacy statuses when calling the SOAP Notary with the associated statuses as filters. 

### GrantsGovApplicationStatus Hierachy
Tracking
Received
(Existing Application DB Status)

- If `Agency Tracking Number Assigned` check is `True` AND `Received by Agency` check is `True` then the `GrantsGovApplicationStatus` is `Agency Tracking Number Assigned`
- If either `Agency Tracking Number Assigned` check is `True`  OR `Received by by Agency` check is `True` then the `GrantsGovApplicationStatus` is whichever is `True`
- If `Agency Tracking Number Assigned` check is `False` AND `Received by Agency` check is `False` then the `GrantsGovApplicationStatus` is `ApplicationStatus`

## transform_submission
The `transform_submission` method takes an `ApplicationSubmission` and returns a `dict`. Currently the `ApplicationStatus` is calculated by a `dict`:
```
GRANTS_APPLICATION_STATUSES = {
    None: None,
    ApplicationStatus.IN_PROGRESS: None,
    ApplicationStatus.SUBMITTED: "Received",
    ApplicationStatus.ACCEPTED: "Validated",
}
```
This takes the status in simpler and converts it to a legacy status. We'll want to check this and then also check the two tables.

## DB Query Filter
```
        if status := submission_filters.get("Status"):
            # If more than one status in filter then just return nothing
            if status and len(status) > 1:
                return []
            simpler_status = STATUS_TRANSFORM.get(str(status[0])) if status else None
            if simpler_status:
[...]

PR #9196, opened 2026-03-20
[Issue #8950] Make new SOAP statuses filterable and returnable

jakobpederson:
## Summary

Fixes / Work for #8950 

## Changes proposed

Added filtering and serializing the new statuses: `Received by Agency` and `Agency Tracking Number Assigned`.

## Context for reviewers

We need to add these two statuses but they aren't stored with the `ApplicationStatuses` so we need to expand the filter to filter the `ApplicationSubmissions` correctly and also return the correct status (the new ones override `ApplicationStatus`).

## Validation steps

- [ ] filtering on the new statuses return correct submissions
- [ ] Submissions in new statuses cannot be filtered by their application_status
- [ ] unittests pass

### DIFF
--- a/api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
+++ b/api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
@@ -40,12 +40,14 @@ STATUS_TRANSFORM = {
 }
 
 
-# This method gets the GrantsGovTrackingNumber
-# NOTE: The order of operations is VERY important to get the correct GrantsGovTrackingNumber
-# IF there's a row on the ApplicationSubmissionTrackingNumber table -> "Agency Tracking Number Assigned"
-# IF there's NOT a row on the ApplicationSubmissionTrackingNumber table but there IS a row on the ApplicationSubmissionRetrieval table -> "Received by Agency"
-# IF there's not a row on either table -> application.application_status
 def get_grants_gov_application_status(submission: ApplicationSubmission) -> str | None:
+    """
+    This method gets the GrantsGovTrackingNumber
+    NOTE: The order of operations is VERY important to get the correct GrantsGovTrackingNumber
+    IF there's a row on the ApplicationSubmissionTrackingNumber table -> "Agency Tracking Number Assigned"
+    IF there's NOT a row on the ApplicationSubmissionTrackingNumber table but there IS a row on the ApplicationSubmissionRetrieval table -> "Received by Agency"
+    IF there's not a row on either table -> application.application_status
+    """
     grants_gov_application_status = GRANTS_APPLICATION_STATUSES.get(
         submission.application.application_status
     )
--- a/api/tests/src/legacy_soap_api/grantors/services/test_get_submission_list_expanded_response.py
+++ b/api/tests/src/legacy_soap_api/grantors/services/test_get_submission_list_expanded_response.py
@@ -77,29 +77,28 @@ def _setup_submission(agency, application_status=ApplicationStatus.ACCEPTED):
     )
 
 
-@pytest.fixture(scope="class")
-def setup_data(db_session, enable_factory_create):
-    agency = AgencyFactory.create()
-    submission = _setup_submission(agency, ApplicationStatus.ACCEPTED)
-    ApplicationSubmissionTrackingNumberFactory.create(application_submission=submission)
-    ApplicationSubmissionRetrievedFactory.create(application_submission=submission)
-    for _ in range(3):
-        sub = _setup_submission(agency, ApplicationStatus.ACCEPTED)
-        ApplicationSubmissionRetrievedFactory.create(
-            application_submission=sub,
-        )
-    _setup_submission(agency, ApplicationStatus.ACCEPTED)
-    tracking_number = f"GRANT{submission.legacy_tracking_number}"
-    _, _, soap_client_certificate = setup_cert_user(agency, {Privilege.LEGACY_AGENCY_VIEWER})
-    return {
-        "agency": agency,
-        "submission": submission,
-        "tracking_number": tracking_number,
-        "soap_client_certificate": soap_client_certificate,
-    }
-
-
 class TestGetSubmissionListExpandedResponseStatusFilter(BaseTestClass):
+    @pytest.fixture(scope="class")
+    def setup_data(self, db_session, enable_factory_create):
+        agency = AgencyFactory.create()
+        submission = _setup_submission(agency, ApplicationStatus.ACCEPTED)
+        ApplicationSubmissionTrackingNumberFactory.create(application_submission=submission)
+        ApplicationSubmissionRetrievedFactory.create(application_submission=submission)
+        for _ in range(3):
+            sub = _setup_submission(agency, ApplicationStatus.ACCEPTED)
+            ApplicationSubmissionRetrievedFactory.create(
+                application_submission=sub,
+            )
+        _setup_submission(agency, ApplicationStatus.ACCEPTED)
+        tracking_number = f"GRANT{submission.legacy_tracking_number}"
+        _, _, soap_client_certificate = setup_cert_user(agency, {Privilege.LEGACY_AGENCY_VIEWER})
+        return {
+            "agency": agency,
+            "submission": submission,
+            "tracking_number": tracking_number,
+            "soap_client_certificate": soap_client_certificate,
+        }
+
     def test_get_submission_list_expanded_filters_on_agency_tracking_number_assigned(
         self, db_session, enable_factory_create, setup_data
     ):


api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
```diff
@@ -22,26 +22,35 @@
...
 
+def get_grants_gov_application_status(submission: ApplicationSubmission) -> str | None:
+    grants_gov_application_status = GRANTS_APPLICATION_STATUSES.get(
+        submission.application.application_status
+    )
+    if len(submission.application_submission_tracking_numbers) > 0:
+        grants_gov_application_status = AGENCY_TRACKING_NUMBER_ASSIGNED_STATUS
+    elif len(submission.application_submission_retrievals) > 0:
+        grants_gov_application_status = RECEIVED_BY_AGENCY_STATUS
+    return grants_gov_application_status
```
chouinar, line 58: This function needs a very detailed comment to explain that the order of operations is very important - mostly as a way of saying "don't touch this unless you have a good reason".

api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
```diff
@@ -51,7 +60,7 @@ def transform_submission(submission: ApplicationSubmission) -> dict[str, str | d
         "FundingOpportunityNumber": opportunity.opportunity_number,
         "CFDANumber": assistance_listing.assistance_listing_number if assistance_listing else None,
         "GrantsGovTrackingNumber": f"GRANT{submission.legacy_tracking_number}",
-        "GrantsGovApplicationStatus": GRANTS_APPLICATION_STATUSES[application.application_status],
+        "GrantsGovApplicationStatus": get_grants_gov_application_status(submission),
```
chouinar: If something ends up with a null status, should it be returned at all? Not actually sure on the answer myself.

api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
```diff
@@ -112,9 +121,23 @@ def get_submissions(
             # If more than one status in filter then just return nothing
             if status and len(status) > 1:
                 return []
-            simpler_status = STATUS_TRANSFORM.get(str(status[0])) if status else None
-            if simpler_status:
-                stmt = stmt.where(Application.application_status == simpler_status)
+            elif status[0] == AGENCY_TRACKING_NUMBER_ASSIGNED_STATUS:
```
chouinar: Just to avoid indexing so often below. Also fine to not connect these if-statements to the above since it would return anyways.

```suggestion
            status_value = str(status[0])
            if status_value == AGENCY_TRACKING_NUMBER_ASSIGNED_STATUS:
```

api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
```diff
@@ -112,9 +121,23 @@ def get_submissions(
...
+                    ~ApplicationSubmission.application_submission_tracking_numbers.any(),
+                )
+            else:
+                simpler_status = STATUS_TRANSFORM.get(str(status[0]))
+                if simpler_status:
+                    stmt = stmt.where(
+                        ~ApplicationSubmission.application_submission_retrievals.any(),
+                        ~ApplicationSubmission.application_submission_tracking_numbers.any(),
+                        Application.application_status == simpler_status,
+                    )
```
chouinar, line 156: I'd add an explanation for why we do these checks.

api/tests/src/legacy_soap_api/grantors/services/test_get_submission_list_expanded_response.py
```diff
@@ -0,0 +1,256 @@
...
+        submission = _setup_submission(agency, ApplicationStatus.ACCEPTED)
+        ApplicationSubmissionTrackingNumberFactory.create(application_submission=submission)
+        ApplicationSubmissionRetrievedFactory.create(application_submission=submission)
+        for _ in range(3):
+            sub = _setup_submission(agency, ApplicationStatus.ACCEPTED)
+            ApplicationSubmissionRetrievedFactory.create(
+                application_submission=sub,
+            )
+        tracking_number = f"GRANT{submission.legacy_tracking_number}"
+        db_session.commit()
```
chouinar: Any call to create on a factory automatically commits - this ends up redundant.

```suggestion

```

api/tests/src/legacy_soap_api/grantors/services/test_get_submission_list_expanded_response.py
```diff
@@ -0,0 +1,256 @@
...
+    opportunity = OpportunityFactory.create(agency_code=agency.agency_code)
+    competition = CompetitionFactory.create(opportunity=opportunity)
+    application = ApplicationFactory.create(
+        competition=competition,
+        application_status=application_status,
+    )
+    return ApplicationSubmissionFactory.create(application=application)
+
+
+class TestGetSubmissionListExpandedResponseStatusFilter:
```
chouinar: Something I might suggest for these tests to trim down the amount of setup needed for all of them, don't do separate setups.

Have a single setup fixture function that we run once. Create a bunch of submissions in various states for one or two agencies. Then the queries just run against that collective dataset. It also implicitly sets up other data that shouldn't be returned (eg. a test for status A also shows that it doesn't get status B/C/D without any extra work.)

You'd just need to do something like:
```py
class TestWhatever(BaseTestClass):

     @pytest.fixture(scope="class")
     def setup_data(self, db_session, enable_factory_create):
            # do whatever

     # the tests
```

The `BaseTestClass` mainly exists as an easy way to get a class-scoped DB session/factory so we can do this. You'd probably also want a fixture for the user / agency that get pulled into this so the other tests can get them.

api/tests/src/legacy_soap_api/grantors/services/test_get_submission_list_expanded_response.py
```diff
@@ -0,0 +1,256 @@
...
+
+
+def _setup_submission(agency, application_status=ApplicationStatus.ACCEPTED):
+    opportunity = OpportunityFactory.create(agency_code=agency.agency_code)
+    competition = CompetitionFactory.create(opportunity=opportunity)
+    application = ApplicationFactory.create(
+        competition=competition,
+        application_status=application_status,
+    )
+    return ApplicationSubmissionFactory.create(application=application)
```
chouinar: I'm not saying it's particularly readable in this case, but something you can do with factories if you want to set a value in a nested one is do something like:

```suggestion
    return ApplicationSubmissionFactory.create(application__competition__opportunity__agency_code=agency.agency_code, application__application_status=application_status)
```

Probably a few too many levels nested to be useful here, but can save a bit of work at times.

api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
jakobpederson: I think the Pydantic model ignores null fields, but it wouldn't be hard to pull it just to be safe.

api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
```diff
@@ -112,9 +128,28 @@ def get_submissions(
...
+            status_value = str(status[0])
+            # GrantsGovTrackingNumber comes from three different places with a hierarchy
+            # A submissions is in "Agency Tracking Number Assigned Status" IF it has any rows on application_submission_tracking_numbers table
+            # A submissions is in "Received by Agency" IF it has any rows on application_submission_retrievals table AND no rows on application_submission_tracking_numbers tabke
+            # A submissions is in  application.application_status IF there are no rows on either of the above tables
+            if status_value == AGENCY_TRACKING_NUMBER_ASSIGNED_STATUS:
+                stmt = stmt.where(
+                    ApplicationSubmission.application_submission_tracking_numbers.any()
+                )
+            if status_value == RECEIVED_BY_AGENCY_STATUS:
```
kkrug: Should this be `elif`?
```suggestion
            elif status_value == RECEIVED_BY_AGENCY_STATUS:
```

api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
```diff
@@ -112,9 +128,28 @@ def get_submissions(
             # If more than one status in filter then just return nothing
             if status and len(status) > 1:
                 return []
-            simpler_status = STATUS_TRANSFORM.get(str(status[0])) if status else None
-            if simpler_status:
-                stmt = stmt.where(Application.application_status == simpler_status)
+            status_value = str(status[0])
+            # GrantsGovTrackingNumber comes from three different places with a hierarchy
+            # A submissions is in "Agency Tracking Number Assigned Status" IF it has any rows on application_submission_tracking_numbers table
+            # A submissions is in "Received by Agency" IF it has any rows on application_submission_retrievals table AND no rows on application_submission_tracking_numbers tabke
```
kkrug:
```suggestion
            # A submissions is in "Received by Agency" IF it has any rows on application_submission_retrievals table AND no rows on application_submission_tracking_numbers table
```

api/tests/src/legacy_soap_api/grantors/services/test_get_submission_list_expanded_response.py
```diff
@@ -0,0 +1,260 @@
...
+            proxy_response=SOAPResponse(data="", status_code=200, headers={}),
+            soap_config=_make_operation_config(),
+        )
+        assert result.success is True
+        assert result.available_application_number == 1
+        assert result.submission_info[0].grants_gov_tracking_number == setup_data["tracking_number"]
+        assert (
+            result.submission_info[0].grants_gov_application_status
+            == "Agency Tracking Number Assigned"
+        )
```
kkrug: It looks like this test is functionally identical to `test_get_submission_list_expanded_filters_on_agency_tracking_number_assigned`. Based on the name I'm assuming this test should verify that a submission with both tracking number and retrieval records shows up under `Agency Tracking Number Assigned` and not under `Received by Agency`, but the test doesn't actually verify the "not under Received by Agency" part.

api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
kkrug, line 123: I think we should add `application_submission_tracking_numbers` and `application_submission_retrievals` as part of the eager loading here since we know call `get_grants_gov_application_status()` as part of `transform_submission()`. `get_grants_gov_application_status()` references both of these relationships but it will default to lazy loading, causing 2 extra queries per submission, which could potentially snowball.
```suggestion
selectinload(ApplicationSubmission.application_submission_tracking_numbers),
selectinload(ApplicationSubmission.application_submission_retrievals),
selectinload(ApplicationSubmission.application).options(
     selectinload(Application.organization),
     selectinload(Application.competition).options(
          selectinload(Competition.opportunity),
          selectinload(Competition.opportunity_assistance_listing),
```

kkrug: The suggestion is displaying the changes weirdly but the only difference is adding those top 2 lines (115-116).

api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
jakobpederson: I'm not entirely sure since the status_value will always be just one thing. I'll swap it in though since I don't think it will hurt.

api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
jakobpederson: Good call. I didn't realize there was a snowball situation here.

api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
```diff
@@ -22,26 +22,40 @@
...
-    "Received by Agency": ApplicationStatus.ACCEPTED,
-    "Agency Tracking Number Assigned": ApplicationStatus.ACCEPTED,
 }
 
 
+# This method gets the GrantsGovTrackingNumber
+# NOTE: The order of operations is VERY important to get the correct GrantsGovTrackingNumber
+# IF there's a row on the ApplicationSubmissionTrackingNumber table -> "Agency Tracking Number Assigned"
+# IF there's NOT a row on the ApplicationSubmissionTrackingNumber table but there IS a row on the ApplicationSubmissionRetrieval table -> "Received by Agency"
+# IF there's not a row on either table -> application.application_status
```
chouinar: For comments about a function, put it in the function with doc strings - this both is the general convention and makes it work nicer with IDEs (hovering over a function shows the comment).

```py
def my_function():
     """
     words
     more words
     """
```

api/tests/src/legacy_soap_api/grantors/services/test_get_submission_list_expanded_response.py
```diff
@@ -0,0 +1,260 @@
...
+
+def _setup_submission(agency, application_status=ApplicationStatus.ACCEPTED):
+    return ApplicationSubmissionFactory.create(
+        application__competition__opportunity__agency_code=agency.agency_code,
+        application__application_status=application_status,
+    )
+
+
+@pytest.fixture(scope="class")
+def setup_data(db_session, enable_factory_create):
```
chouinar: A scope of class means it needs to be in a class - I'm surprised pytest doesn't error for this actually.

No idea what the behavior is - but shouldn't do that. Put it in the test class

api/src/legacy_soap_api/grantors/services/get_submission_list_expanded_response.py
jakobpederson: Thanks, I wasn't sure when to use docstrings.

PR #9197, opened 2026-03-20
Include saved_to_organizations in individual saved opportunities

freddieyebra:
## Summary

Fixes / Work for #9203  

## Changes proposed

Include `saved_to_organizations` property in saved opportunities response. The initial requirement for this endpoint was to not include this property for individual scoped saved opportunities. Upon further investigation, it was decided that we want to include this property to ensure consistency in the saved opportunity card API response and the UI.

## Context for reviewers

When testing the functionality of adding/removing organization saved opportunities, there was unexpected functionality around saving a user saved opportunity. This PR adds the saved_to_organizations property to ensure it is included in the API request for getting individual scoped saved opportunities.

## Validation steps

- Unit tests pass


api/tests/src/api/users/test_user_saved_opportunities_get.py
```diff
@@ -191,7 +191,47 @@ def test_user_get_saved_opportunities_with_empty_org_filter_returns_only_user_sa
...
+            },
+        },
+    )
+
+    assert response.status_code == 200
+    data = response.json["data"]
+
+    assert len(data) == 1
+    assert data[0]["opportunity_id"] == str(user_saved_opp.opportunity_id)
+    assert data[0]["saved_to_organizations"] == [
```
babebe, line 229: My understanding is that when organization_ids is [], then we should only return user saved opportunities, per [ticket](https://github.com/HHS/simpler-grants-gov/issues/8433). Why are we returning information about what organization the user belongs to also saved this opportunity. If we want this behavior, will need to make more changes

freddieyebra:
> Include `saved_to_organizations` property in saved opportunities response.

Hi @babebe as we had discussed, the initial AC for the individual filter on the api was to not include the "saved_to_organization" property but as we had discussed earlier this week and as result of alignment with @ErinPattisonNava and @chelsea-nava we do want to ensure this property is included in the API response so that the UI can consume it.

freddieyebra:
@joshtonava @mikehgrantsgov mind giving this a final look over? @babebe and I both have commits in this PR.

PR #9216, opened 2026-03-23
[Issue #8978] Ensure organization saved opportunity notifications are removed when a user is no longer a member of an organization

### LINKED ISSUE #8978: [BE] Ensure organization saved opportunity notifications are removed when a user is no longer a member of an organization
### Summary

We need to ensure that when a user is removed from an organization through the [remove user from organization endpoint ](https://github.com/HHS/simpler-grants-gov/blob/89ba21b4ba5453d051acb57bdd842cfe45d573e0/api/src/api/organizations_v1/organization_routes.py#L145) `/v1/organizations/{organization_id}/users/{user_id}` any saved organization opportunity notifications database entries in the `user_saved_opportunities_notifications` table are deleted.

### Acceptance criteria
- [ ] When a user is removed from any organization they are a member or admin on, we need to ensure that any notification configurations they have around organization saved opportunities are deleted from the `user_saved_opportunities_notifications` table for only that organization
- [ ] This change should not affect self notification settings or settings for notifications for that user around other organizations they are part of
- [ ] If there are any errors we should log an error to ensure we are alerted but this should not return any error responses in the endpoint if the user was successfully removed from the organization.
- [ ] Unit tests

mikehgrantsgov:
## Summary

Fixes #8978 

## Changes proposed

Modified `remove_user_from_organization` to call a new `_delete_org_notification_preferences` helper after removing the user
The helper deletes the `UserSavedOpportunityNotification` row scoped to the specific organization the user is leaving
Self notifications (organization_id=NULL) and notifications for other organizations are left untouched

## Context for reviewers

Deletion failures in `_delete_org_notification_preferences` are caught and logged via `logger.exception` without re-raising, so a notification cleanup failure doesn't block the user removal response.

## Validation steps

See new unit tests.

### DIFF
--- a/api/src/services/organizations_v1/remove_user_from_organization.py
+++ b/api/src/services/organizations_v1/remove_user_from_organization.py
@@ -111,26 +111,17 @@ def remove_user_from_organization(
 def _delete_org_notification_preferences(
     db_session: db.Session, user_id: UUID, organization_id: UUID
 ) -> None:
-    """Delete the user's notification preferences for the given organization.
+    """Delete the user's notification preferences for the given organization."""
+    notification = db_session.execute(
+        select(UserSavedOpportunityNotification).where(
+            UserSavedOpportunityNotification.user_id == user_id,
+            UserSavedOpportunityNotification.organization_id == organization_id,
+        )
+    ).scalar_one_or_none()
 
-    Failures are logged but do not block the user removal.
-    """
-    try:
-        notification = db_session.execute(
-            select(UserSavedOpportunityNotification).where(
-                UserSavedOpportunityNotification.user_id == user_id,
-                UserSavedOpportunityNotification.organization_id == organization_id,
-            )
-        ).scalar_one_or_none()
-
-        if notification is not None:
-            db_session.delete(notification)
-            logger.info(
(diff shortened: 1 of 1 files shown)


api/src/services/organizations_v1/remove_user_from_organization.py
```diff
@@ -102,3 +106,31 @@ def remove_user_from_organization(
...
+
+
+def _delete_org_notification_preferences(
+    db_session: db.Session, user_id: UUID, organization_id: UUID
+) -> None:
+    """Delete the user's notification preferences for the given organization.
+
+    Failures are logged but do not block the user removal.
+    """
+    try:
```
chouinar: I'm not sure if this try/catch would help - in the event that we tried to delete and hit some sort of error, that'd break the whole DB session - and we have DB logic before and after this call, so there's not really a way to gracefully resolve.

Plus with how this is implemented, the only way this could error is if someone the execute statement returned an object that couldn't then be deleted? Seems improbable.

mikehgrantsgov: Make sense. I tweaked this.

Issue #7723, opened 2025-12-30
create a maintenance flag to disable the site in cases of emergency


ErinPattisonNava:
### Summary

We need a way to turn off the grants website in the case something goes wrong, for instance if a deploy goes sideways or we are having a DOS attack.  We want this solve to be straightforward and universal, so it has less of a chance of failing and covers more cases.  At a minimum we need a site-wide maintenance page.  We may also want this flag to disable our APIs, or we may wish to handle that a different way.

### Acceptance criteria

- [ ] Create a way to put the site into "maintenance mode".  This most likely will be a feature flag that triggers all pages to show the maintenance page.
- [ ] Consider if we need to extend this "maintenance mode" to our APIs.
   - [ ] Get an analysis and agreement to the Platform team on options here.
   - [ ] write follow up implementation tickets.

PR #9234, opened 2026-03-24
[Issue #7723] Create a maintenance flag to disable the site in cases of emergency

kkrug:
## Summary

Fixes / Work for #7723  

## Changes proposed

- Updated defaultFeatureFlags.ts to remove `opportunitiesListOff: false` and add `maintenanceMode: false`
- Updated environments.ts to add `FEATURE_MAINTENANCE_MODE` and mapping and remove `FEATURE_OPPORTUNITIES_LIST_OFF` and mapping
- Updated proxy.ts to add maintenance mode check in the middleware. When `maintenanceMode` is enabled, all non-API, non-maintenance-page requests are redirected to `/maintenance`
- Updated environment_variables.tf to add `FEATURE_MAINTENANCE_MODE` environment variable and remove `FEATURE_SEARCH_DRAWER_ON` and `FEATURE_SEARCH_TABLE_ON` environment variables
- Updated /opportunities/page.tsx to remove `withFeatureFlag` wrapper as the middleware now handles this instead
- Updated /opportunities/page.test.tsx to account for the removal of `withFeatureFlag`
- Updated providerMocks.tsx to removed `opportunitiesListOff` from mock feature flags

## Context for reviewers

We wanted to create a way to turn off the site in case of emergency. This creates a feature flag that will redirect all pages to the maintenance page by changing the flag in SSM.

**Note:** The new SSM parameter will need to be created and set to "false" in all environments: `/frontend/${var.environment}/feature-maintenance-mode`.

The following SSM parameters can be removed:
- `/frontend/${var.environment}/feature-search-table-on`
- `/frontend/${var.environment}/feature-search-drawer-on`

## Validation steps

1. Run the frontend locally. You should be able to navigate to every page as normal. 
2. To enable maintenance mode (i.e. setting the flag to true), navigate to this url: http://localhost:3000?_ff=maintenanceMode:true
3. You will be redirected to the maintenance page. Navigating to any other page will similarly redirect to this same maintenance page.
4. To disable maintenance mode, navigate to this url: http://localhost:3000?_ff=maintenanceMode:false
5. You will be redirected to the home page and can now freely navigate to any page on the site as normal.


kkrug:
Noting that we are not considering the CI check "Check compliance with tfsec" failing as a blocker at this time.

frontend/src/proxy.ts
```diff
@@ -87,10 +87,37 @@ export default function proxy(request: NextRequest): NextResponse {
...
+  });
+  const isMaintenanceMode = featureFlagsManager.isFeatureEnabled(
+    "maintenanceMode",
+    request.cookies,
+    searchParamsDict,
+  );
+  const isMaintenancePage = request.nextUrl.pathname.includes("/maintenance");
+
+  if (isMaintenanceMode && !isMaintenancePage && !isApiRoute) {
+    const maintenanceUrl = new URL("/maintenance", request.url);
```
doug-s-nava, line 111: til that you can do this. Awesome! https://developer.mozilla.org/en-US/docs/Web/API/URL_API/Resolving_relative_references

PR #9237, opened 2026-03-24
 [Issue #9054] Workflow architecture/system documentation

### LINKED ISSUE #9054: Write workflow documentation describing the architecture of the system
### Summary

Focus on documenting how the system works together, the data model, and the general flow of data through the system between the API and workflow service.

Does not need to focus on low-level state machines information yet, a separate ticket will document that.

### Acceptance criteria

- [ ] Architecture documents included defining how the system connects
- [ ] Data model described and explained
- [ ] Describe the lifecycle of an event

chouinar:
## Summary
Fixes #9054

## Changes proposed
Adds documentation describing the architecture and system of the workflow API + system.

## Context for reviewers
It's docs - we have a few more tickets coming for describing how to build a workflow, and how to use the API, so those aren't included here. Focused on the more high-level system architecture and data model.

You'll want to look at the rich diff so the mermaid diagrams render.

## Validation steps
It's docs

### DIFF
--- a/documentation/api/workflow-service.md
+++ b/documentation/api/workflow-service.md
@@ -43,7 +43,7 @@ Our architecture consists of a few components:
   * `PUT /v1/workflows/events` - Send an event against a workflow, will write to SQS if valid
   * `GET /v1/workflow/:workflow_id` - Get a workflow
 * An SQS queue for storing the events
-* A SQS dead-letter-queue for storing events that fail to process multiple times
+* An SQS dead-letter-queue for storing events that fail to process multiple times
 * Workflow Service - An always running service that processes events async
   * Fetches events from the SQS queue
   * Processes them, running them against the workflow + state machine configuration
@@ -55,8 +55,10 @@ If we receive a surge in events, the queue will fill up, but the service
 can work through processing it.
 
 Our SQS queue uses the most general configuration, the order we receive
-events, and whether we may receive a duplicate event is not defined, and
-should be expected, and handled.
+events, and whether we may receive a duplicate event is not defined. It
+is expected that any work on our handler logic needs to be aware of and account
+for this. As we go forward and add multi-threading support and scale the service
+up we will account for this.
 
 ## Auth
 Authentication is handled purely in the APIs, the backend service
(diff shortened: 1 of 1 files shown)


documentation/api/workflow-service.md
```diff
@@ -0,0 +1,217 @@
...
+  privilege is required to approve it.
+* Event - An event represents the set of information we send to a workflow to process
+  it. Every event maps 1:1 with SQS events that we send, and all events get processed
+  although some error scenarios may cause an event to be processed without changing a workflow.
+* Transition - The arrows on the state machine linking states together. Transitions
+  are caused by events, but a given event can cause multiple transitions to occur in sequence.
+
+
+# Architecture
+![Local DB CLI](./images/workflow-arch.png)
```
kkrug, line 39: Should this be something like "Workflow Architecture Diagram" instead of "Local DB CLI?"

documentation/api/workflow-service.md
```diff
@@ -0,0 +1,217 @@
...
+    
+    Opportunity {}
+    ApplicationSubmission {}
+    
+    Workflow ||--o| WorkflowEventHistory: Maintains
+    Workflow ||--o| WorkflowAudit: Records
+    Workflow ||--o| WorkflowApproval: Records
+    
+    WorkflowAudit ||--o| WorkflowEventHistory: Connected
+    WorkflowApproval ||--o| WorkflowEventHistory: Connected
```
kkrug: Doesn't `||--o|` describe 1-to-1 relationships? These should be 1-to-many, which I believe is `||--o{`.

documentation/api/workflow-service.md
```diff
@@ -0,0 +1,217 @@
...
+could it be associated with nothing. This is enforced as a hard requirement
+in the database.
+
+Workflow audit, workflow approval, and event history are all
+history tables, but serve different purposes.
+
+WorkflowEventHistory tracks every event put into the SQS queue.
+If events fail to process, but we don't think the event could be processed,
+(eg. the event doesn't exist for the workflow type), then we will
+remove it from the queue, and still record it to the database.
```
kkrug: Sentence structure that's a bit clearer
```suggestion
If an event fails to process because it's invalid for the workflow type (e.g. the event doesn't exist for that workflow type), we still remove it from the queue and record it to the database.
```

documentation/api/workflow-service.md
```diff
@@ -0,0 +1,217 @@
...
+
+
+# Architecture
+![Local DB CLI](./images/workflow-arch.png)
+
+Our architecture consists of a few components:
+* API Endpoints (part of our existing)
+  * `PUT /v1/workflows/events` - Send an event against a workflow, will write to SQS if valid
+  * `GET /v1/workflow/:workflow_id` - Get a workflow
+* An SQS queue for storing the events
```
hao10282025-sudo, line 45: change to "an" SQS queue in line 46?
it is hard to not pay attention when next line is different between "a" and "an"

documentation/api/workflow-service.md
```diff
@@ -0,0 +1,217 @@
...
+  * If the event was successfully processed, deletes the message from the queue
+* Database - our existing one, same as used by the API
+
+Importantly, this approach separates event sending from processing.
+If we receive a surge in events, the queue will fill up, but the service
+can work through processing it.
+
+Our SQS queue uses the most general configuration, the order we receive
+events, and whether we may receive a duplicate event is not defined, and
+should be expected, and handled.
```
hao10282025-sudo: Good point, as I see the order should not be matter, as one workflow can only have one state in the queue/processing at any moment?

chouinar: I meant this more as "as a developer, you have to be aware of this" not that it's handled for you without needing to think about it. This only matters if you're working on the event handling logic, not the underlying state machine.

When we add multi-threading next quad, we will have to account for this quite a bit.

chouinar: Adjusted it to be clearer

documentation/api/workflow-service.md
chouinar: I cleaned up the wording a bit.

documentation/api/workflow-service.md
chouinar: I think I've fixed it

Issue #8829, opened 2026-03-05
Force Pre-population of data on all forms after transferring application ownership


chris-kuryak:
### Summary

After transferring ownership of an application to an organization, the pre-population data on forms does not automatically update to reflect the new information associated with the organization. Instead, the user must go to each form and update them individually (just hitting save works).

Risk:
- User confusion seeing some data that is not accurate, and they cannot change (since it is a read-only field)
- User could submit incorrect information in their submission if they do not trigger an update/save of pre-population on forms before submitting

This video shows details on the use case being described.
https://www.loom.com/share/a3b933419a6d4b8ea0e5dd1691db09d1

From Josh O in comment below:
These fields are automatically populated via getter functions in `json_rule_field_population.py`:

| Form | Field | Data Source |
  |---|---|---|
  | SF-424 | Funding Opportunity Number | Opportunity record |
  | SF-424 | Funding Opportunity Title | Opportunity record |
  | SF-424 | Agency Name | Opportunity's associated agency |
  | SF-424 | UEI (Unique Entity Identifier) | Applicant's organization (defaults to `00000000INDV` for individuals) |
  | SF-424 | Assistance Listing Number | Competition record |
  | SF-424 | Assistance Listing Program Title | Competition record |
  | SF-424 | Competition ID | Competition record |
  | SF-424 | Competition Title | Competition record |
  | SF-424 | Current Date | System clock |
  | SF-424 | Signature | Submitting user's email |
  | SF-LLL | Federal Program Name | Competition record |
  | SF-LLL | Assistance Listing Number | Competition record |

### Acceptance criteria
- [ ] Identify solution and align with leads
- [ ] Transferring ownership of an application repopulates all pre-populated data fields on all forms correctly
- [ ] QA includes FE testing
- [x] UAT Passed
  - [x] Pre-populated data re-populates correctly upon transferring application ownership
    - [x] SF-424 | Funding Opportunity Number
    - [ ] SF-424 | Funding Opportunity Title
    - [ ] SF-424 | Agency Name 
    - [ ] SF-424 | UEI (Unique Entity Identifier) 
    - [ ] SF-424 | Assistance Listing Number 
    - [ ] SF-424 | Assistance Listing Program Title
    - [ ] SF-424 | Competition ID
    - [ ] SF-424 | Competition Title
    - [ ] SF-424 | Current Date
    - [ ] SF-424 | Signature
    - [ ] SF-LLL | Federal Program Name
[...]

PR #9243, opened 2026-03-25
[Issue 8829] Update logic to add org to application

babebe:
## Summary

Work for #8829  

## Changes proposed

Set the application → organization relationship directly:
`application.organization = organization`
This ensures the organization object is available in the ORM context during pre-population.
## Context for reviewers

This PR fixes an issue where pre-populated fields were not updating after transferring an application to an organization.
The root cause was that the Application was only being updated via organization_id, but the ORM relationship (application.organization) was not set. As a result, prepopulation logic that depends on the organization object did not have access to the necessary data at runtime.

**scope**
The only impacted field is `sam_uei`
This is because:
`sam_uei` is organization-derived
Other prepopulated fields come from competition/opportunity, which do not change when adding an organization
`sam_uei` exists in two forms:
- SF-424
- SF-4700

## Validation steps

Added service-level test to confirm pre-population has access to organization data after assigning the organization

Local Test
1. Login as a user associated with an org:  can use `one_org_user`
2. Search for opportunity: `STATIC Opportunity with ALL forms`
3. Start a new application as an individual
4. Open forms `SF-424` and `SF-4700` and confirm UEI is the individual value: `00000000INDV`
5. Transfer application ownership to the organization
6. Re-open the same forms and confirm:
    - UEI is now updated to the organization’s UEI
7. No manual save is required to trigger the update


joshtonava:
FYI I think this will need to be QAd before the ticket can fully move to done

PR #9262, opened 2026-03-25
[Issue #8871] When publishing an opportunity, calculate the current summary + status

### LINKED ISSUE #8871: Make the opportunity publish workflow calculate the current opportunity summary + opportunity status
### Summary

When the opportunity publish workflow does the `calculate_current_opportunity_summary` event, it should calculate the current opportunity summary + opportunity status and set them in the database. This logic already exists in https://github.com/HHS/simpler-grants-gov/blob/main/api/src/task/opportunities/set_current_opportunities_task.py - we will need to refactor this a bit so we can reuse the logic that determines the status, this should not change that job other than to move the code around.

The way we want to define these handlers is using decorators. We'll want to do:

```python
@calculate_current_opportunity_summary.on
def handle_ calculate_current_opportunity_summary(self, state_machine_event: StateMachineEvent) -> None:
    pass
```

Note that similar to the existing logic, it's entirely possible that an opportunity will not get an opportunity status / current opportunity summary. That is perfectly fine, we'll add branching logic in the next step that checks whether it is set.

### Acceptance criteria

- [ ] Current opportunity summary logic refactored to allow reuse of it in this logic
- [ ] Current opportunity summary logic added to workflow
- [ ] Tests added that verify value is calculated

chouinar:
## Summary
Fixes #8871

## Changes proposed
* During the opportunity publish workflow, calculate the opportunity summary + status like we do in our hourly job

## Context for reviewers
Largely just moving logic we do in an hourly job to a shared location and reusing it. 

## Validation steps

First setup an opportunity that doesn't have a summary attached by running `make console` and doing something like:
```py
opportunity = f.OpportunityFactory.create(is_draft=True, no_current_summary=True)
f.OpportunitySummaryFactory.create(opportunity=opportunity, is_forecast=False)
```

Run the API locally while running the workflow service with make cmd args="workflow workflow-main".

Find an opportunity in your local DB and you can call the PUT workflow endpoint by setting that as the entity ID in a request like below using the `internal_workflow_user_key` user API key.

```json
{
  "start_workflow_context": {
    "entity_id": "4c061063-dc35-4a18-bafa-25b9c541c619",
    "entity_type": "opportunity",
  }
}
```

This'll run the whole workflow, you should see the summary calculated in the logs + by checking in the database.

### DIFF
--- a/api/tests/workflow/state_machine/test_opportunity_publish_state_machine.py
+++ b/api/tests/workflow/state_machine/test_opportunity_publish_state_machine.py
@@ -143,11 +143,8 @@ def test_opportunity_publish_calculate_current_opportunity_summary(
             opportunity.current_opportunity_summary.opportunity_summary.is_forecast
             == expected_is_forecast
         )
-        assert (
-            opportunity.current_opportunity_summary.opportunity_status
-            == OpportunityStatus.FORECASTED
-            if expected_is_forecast
-            else OpportunityStatus.POSTED
+        assert opportunity.current_opportunity_summary.opportunity_status == (
+            OpportunityStatus.FORECASTED if expected_is_forecast else OpportunityStatus.POSTED
         )
 
 


api/tests/workflow/state_machine/test_opportunity_publish_state_machine.py
```diff
@@ -64,6 +72,85 @@ def test_opportunity_publish_happy_path(db_session, enable_factory_create, is_dr
...
+        assert (
+            opportunity.current_opportunity_summary.opportunity_summary.is_forecast
+            == expected_is_forecast
+        )
+        assert (
+            opportunity.current_opportunity_summary.opportunity_status
+            == OpportunityStatus.FORECASTED
+            if expected_is_forecast
+            else OpportunityStatus.POSTED
+        )
```
kkrug: This could give some unintended results when `expected_is_forecast` is `false` without the parentheses. 
```suggestion
        assert (
            opportunity.current_opportunity_summary.opportunity_status
            == (OpportunityStatus.FORECASTED
            if expected_is_forecast
            else OpportunityStatus.POSTED)
        )
```

kkrug (approved):

Issue #8400, opened 2026-02-05
Workflow prevents a user from starting a workflow if one already exists for a given entity


chouinar:
### Summary

We want to add a field to our workflow configuration called `allow_concurrent_workflow_for_entity`. Default it to True.

Then, we want to add a function that given a workflow type, entity type + the entity ID (or the entity itself which we'll have fetched), check if:
* That entity has a workflow of that type
* That workflow `is_active=True` - we're fine if we start another one later (might change that in the future)

If another active workflow does exist for the entity + workflow type - we want to error.

We want to add this check to:
* The start workflow logic in the event handler (should be a non-retryable error)
* The start workflow validation in the event API - should be a 422(? I think?)

### Acceptance criteria

- [x] Added configuration that disallows multiple of the same type of workflow for a given entity
- [x] Added a function to validate this - added it to the API + event handler

PR #9276, opened 2026-03-26
[Issue #8400] Workflow prevents a user from starting a workflow if one already exists for a given entity

kkrug:
## Summary

Fixes for #8400  

## Changes proposed

- Updated workflow_config.py to add `allow_concurrent_workflow_for_entity` boolean with a default of `True`
- Updated workflow_errors.py to add `ConcurrentWorkflowError`
- Updated workflow_service.py to add `validate_no_concurrent_workflow` function
- Update event_handler.py to add call to `validate_no_concurrent_workflow`
- Updated ingest_workflow_event.py to add catch for `ConcurrentWorkflowError` and call to `validate_no_concurrent_workflow`
- Updated lookup_constants.py to add `NO_CONCURRENT_TEST_WORKFLOW` enum value for testing
- Updated lookup_models.py to add lookup config for `NO_CONCURRENT_TEST_WORKFLOW`
- Updated test_state_machine.py to add `NoConcurrentTestStateMachine` with `allow_concurrent_workflow_for_entity = False`
- Updated test_workflow_service.py to add tests for `validate_no_concurrent_workflow`
- Updated test_event_handler.py to add concurrent workflow tests
- Updated test_ingest_workflow_event.py to add concurrent workflow tests

## Context for reviewers

We needed to add a configuration that disallows multiple of the same type of workflow for a given entity.

## Validation steps

Confirm tests pass.

### DIFF
--- a/api/src/workflow/service/workflow_service.py
+++ b/api/src/workflow/service/workflow_service.py
@@ -5,7 +5,7 @@ from typing import Any
 from sqlalchemy import select
 
 from src.adapters import db
-from src.constants.lookup_constants import WorkflowEntityType, WorkflowType
+from src.constants.lookup_constants import WorkflowEntityType
 from src.db.models.base import ApiSchemaTable
 from src.db.models.competition_models import Application
 from src.db.models.opportunity_models import Opportunity
@@ -17,7 +17,6 @@ from src.workflow.workflow_errors import (
     EntityNotFound,
     ImplementationMissingError,
     InactiveWorkflowError,
-    InvalidEntityForWorkflow,
     WorkflowDoesNotExistError,
 )
 
@@ -26,7 +25,6 @@ logger = logging.getLogger(__name__)
 
 def get_workflow_entity(
     db_session: db.Session,
-    entity_type: WorkflowEntityType,
     entity_id: uuid.UUID,
     config: WorkflowConfig,
 ) -> dict[str, ApiSchemaTable]:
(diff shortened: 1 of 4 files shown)


api/src/workflow/service/workflow_service.py
```diff
@@ -108,3 +109,53 @@ def get_and_validate_workflow(
...
+    if entity_column is None:
+        raise ImplementationMissingError(
+            f"Entity type {entity_type} is not configured for concurrent workflow validation"
+        )
+
+    existing_workflow = db_session.scalar(
+        select(Workflow).where(
+            Workflow.workflow_type == workflow_type,
+            entity_column == entity_id,
+            Workflow.is_active == True,  # noqa: E712
```
chouinar: SQLAlchemy has an `is_` function for doing bool comparisons to avoid type-checking complaints.

```suggestion
            Workflow.is_active.is_(True),
```

api/src/workflow/service/workflow_service.py
```diff
@@ -108,3 +109,53 @@ def get_and_validate_workflow(
...
+    WorkflowEntityType.APPLICATION: Workflow.application_id,
+}
+
+
+def validate_no_concurrent_workflow(
+    db_session: db.Session,
+    workflow_type: WorkflowType,
+    entity_type: WorkflowEntityType,
+    entity_id: uuid.UUID,
+    config: WorkflowConfig,
```
chouinar: Workflow type and entity type we can get from the config, if we want to avoid having to pass quite as much in - although not sure how easy that is to test then. Leave it up to you to decide.

api/src/workflow/service/workflow_service.py
kkrug: Updated

api/src/workflow/service/workflow_service.py
kkrug: Makes sense, I made the change. I also made a similar change for `get_workflow_entity` and updated all the tests.

PR #1, opened 2023-04-05
Setting up repo

lucasmbrown-usds:
Initial work on repo structure.

Issue #9229, opened 2026-03-24
Configure ALB logs to flow into Newrelic


prasnava:
Configure ALB logs to flow into newrelic to help with debugging the 502 errors that the testing team is seeing in our automated e2e tests.

PR #9281, opened 2026-03-26
[Issue #9229]: forward alb logs into newrelic

prasnava:
## Summary

Forwarded alb access logs into newrelic to help debug what is causing the intermittent errors we are seeing in the staging environment. 

## Validation steps

Alb access logs for frontend and api are accessible in newrelic for the dev environment. 

Screenshot 2026-03-26 at 3 14 11 PM

### DIFF
--- a/api/src/services/users/get_saved_opportunities.py
+++ b/api/src/services/users/get_saved_opportunities.py
@@ -173,24 +173,31 @@ def _build_saved_union_subquery(
 def _enrich_with_saved_organizations_sql(
     db_session: db.Session,
     opportunities: Sequence[Opportunity],
+    org_ids_filter: list[uuid.UUID] | None = None,
 ) -> None:
     """
     Enrich paginated Opportunity objects with `saved_to_organizations`.
 
     Behavior:
-    -  If the request is for **user-saved opportunities only**, the `saved_to_organizations` field is omitted.
-    - If the request involves organizations the `saved_to_organizations` field is always included.
+    Always enriches opportunities with organizations that have saved them.
+    - When `org_ids_filter` is provided, only organizations in that list are included.
+    - When `org_ids_filter` is None, all organizations that saved the opportunities are included.
+
     """
     opp_ids = [opp.opportunity_id for opp in opportunities]
 
-    result = db_session.execute(
+    query = (
         select(OrganizationSavedOpportunity.opportunity_id, Organization)
         .join(
             Organization,
             Organization.organization_id == OrganizationSavedOpportunity.organization_id,
         )
         .where(OrganizationSavedOpportunity.opportunity_id.in_(opp_ids))
-    ).all()
+    )
+    if org_ids_filter is not None:
+        query = query.where(OrganizationSavedOpportunity.organization_id.in_(org_ids_filter))
+
+    result = db_session.execute(query).all()
 
     # Build a mapping: opportunity_id to list of Organization objects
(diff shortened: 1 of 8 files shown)


infra/api/app-config/env-config/variables.tf
```diff
@@ -245,3 +245,15 @@ variable "workflow_service_desired_count" {
   type        = number
   default     = 1
 }
+
+variable "service_newrelic_entity_guid" {
+  type        = string
+  description = "New Relic entity GUID for the primary ALB, used to correlate logs with the infrastructure entity in New Relic."
+  default     = ""
```
sean-navapbc: what happens if these are left blank? what's kind of behavior will happen? Is it a required variable?

sean-navapbc:
## Review

Nice work — the ALB log forwarding approach is solid. The Lambda reads the NR license key from SSM at runtime (addressing the feedback from the ES log forwarder PR), has proper error handling, batching, and gzip compression. A few items to address:

### 1. Entity GUID variables default to `""` — what happens when blank?

The `service_newrelic_entity_guid` and `service_newrelic_mtls_entity_guid` variables default to `""` in all three layers (module, api env-config, frontend env-config). When left blank:
- The empty string is passed as the `NR_ENTITY_GUID` / `NR_MTLS_ENTITY_GUID` Lambda env vars
- The Python code handles this correctly (`if NR_ENTITY_GUID:` is falsy for `""`)
- Logs still get forwarded, just without `entity.guid` in the common attributes — so no entity correlation in New Relic

This works, but consider using `default = null` instead of `default = ""` to make the "not configured" intent explicit. Then in the Lambda env vars block, use `coalesce(var.newrelic_entity_guid, "")`. Or just keep `""` and add a comment — either way is fine, the behavior is safe.

### 2. Frontend doesn't pass `newrelic_mtls_entity_guid`

`infra/frontend/service/main.tf` passes `newrelic_entity_guid` but not `newrelic_mtls_entity_guid`. This works because the module variable has a default, but it means the frontend mTLS ALB logs (if any) won't have entity correlation. Is that intentional? (Frontend likely doesn't have an mTLS ALB, so probably fine.)

### 3. Lambda log group retention is 365 days

`retention_in_days = 365` for the Lambda's own log group. The checkov skip comment says "Forwarding Lambda logs don't need long retention" — consider dropping to 30-90 days since this Lambda's output is just status messages about forwarding. The actual ALB logs are in New Relic.

### 4. S3 notification only covers primary ALB prefix

```hcl
filter_prefix = "${var.service_name}-lb/"
```

This only triggers on the primary ALB's logs. If the mTLS ALB writes to a different prefix (e.g., `${var.service_name}-mtls-lb/`), those logs won't trigger the Lambda. The Lambda code handles mTLS ALB names but may never receive their log files. Is the mTLS ALB configured to write to the same prefix, or does it need a separate notification?

### Looks good overall
- SSM runtime fetch for NR license key
- Proper error handling with retryable vs non-retryable distinction
- Batching with BATCH_SIZE = 1000
- Gzip compression for New Relic payloads
- ALB log parsing with shlex handles quoted fields correctly
- S3 event notification scoped to `.log.gz` files

CI all green.

prasnava:
@sean-navapbc Resolved #1, 3 and 4. The mtls alb and regular alb write to the same s3 destination. For #2, we don't have an mtls endpoint for the frontend.

PR #9282, opened 2026-03-26
[Issue #9081] Update and validate custom fields

### LINKED ISSUE #9081: [CommonGrants] Update and validate custom fields
### Summary

Update the set of custom fields supported by the SGG API to align with the fields listed on [the CommonGrants custom field catalog](https://commongrants.org/custom-fields/), enforce validation of custom field values with Pydantic models in `populate_custom_fields()` and gracefully handle validation errors.

**Note:** This may also require some corresponding changes in the custom field catalog #HHS/simpler-grants-protocol/issues/610

### Acceptance criteria

- [ ] SGG's custom fields match the corresponding schemas in the custom field catalog
- [ ] The data passed to these custom fields is validated at the service layer (i.e. with Pydantic models) before they reach the corresponding Marshmallow schemas for serialization in the routing layer
- [ ] Pydantic validation errors are caught and gracefully handled by:
  - [ ] Omitting the field with validation errors from the custom fields object (preference)
  - [ ] Omitting the record that contains an invalid custom field from the results (backup)
- [ ] OpenAPI docs have more detailed schemas for the `CustomField.value` attribute

jcrichlake:
## Summary

Fixes #9081  

## Changes proposed

 - Added typed Pydantic schemas for each CommonGrants custom field in a new common_grants_pydantic_custom_fields.py module
  - Added a validate_custom_field() helper function in transformation.py that validates custom field data against its typed Pydantic schema before including it in the response
  - Updated populate_custom_fields() to use the new typed field classes and validate_custom_field() instead of constructing generic CustomField instances directly
  - Tightened Marshmallow schema field types in common_grants_custom_fields.py — replacing fields.Raw() with typed alternatives (fields.Integer, fields.String, fields.Boolean,
  fields.Nested, fields.List)
  - Reorganized schemas into a schemas/ subdirectory (common_grants_custom_fields.py and common_grants_schemas.py moved under schemas/)
  - Updated imports in common_grants_routes.py to reflect the new module paths

## Context for reviewers

  Previously, populate_custom_fields() constructed all custom fields using the generic CustomField base class with untyped value fields, meaning invalid data could silently pass through
   without validation. This PR introduces a typed Pydantic subclass for each custom field (e.g., LegacySerialIdField, AgencyField, AttachmentsField) that enforces value types at
  construction time.

  The validate_custom_field() wrapper catches validation errors, logs a warning with the OPPORTUNITY_VALIDATION_ERROR event, and returns None so the caller can skip including invalid
  fields in the response rather than crashing or silently serving bad data.

  The AttachmentValue model also includes special handling for downloadUrl: it attempts to parse the value as a valid HttpUrl and falls back to None if it's malformed, rather than
  failing the entire field.

## Validation steps

Start the common grants endpoints with data

```bash
# Install dependencies
poetry install

# Start API with data
make init 
make run-logs
```

Run the following curl commands to validate.

List:
`curl -H 'accept: application/json' -H 'X-Api-Key: two_org_user_key' -X 'GET'  'http://localhost:8080/common-grants/opportunities'`

Get by ID:
`curl -H 'accept: application/json' -H 'X-Api-Key: two_org_user_key' -X 'GET'  'http://localhost:8080/common-grants/opportunities/8b5f468a-8530-4e4f-9d72-a5713adbc0ec'`

Search:
`curl -H 'accept: application/json' -H 'X-Api-Key: two_org_user_key' -X 'POST' 'http://localhost:8080/common-grants/opportunities/search' -d '{ "filters": { "status": { "operator": "in", "value": ["open", "forecasted"] }, "pagination": { "page": 1, "pageSize": 10 }, "search": "education", "sorting": { "sortBy": "lastModifiedAt", "sortOrder": "desc" } }' `


api/src/services/common_grants/transformation.py
```diff
@@ -373,30 +405,23 @@ def populate_custom_fields(opp_data: dict) -> dict[str, CustomField] | None:
             }
             for attachment in attachments
         ]
-        custom_fields["attachments"] = CustomField(
-            name="attachments",
-            fieldType=CustomFieldType.ARRAY,
-            value=attachment_values,
-            description="Attachments such as NOFOs and supplemental documents for the opportunity",
-        )
+        field = validate_custom_field(AttachmentsField, value=attachment_values)
```
SnowboardTechie: `AttachmentsField` validates the entire list at once. If a single attachment has invalid data (e.g. a null `file_name` in the DB), Pydantic rejects the whole list, `validate_custom_field` returns `None`, and all attachments are silently dropped from the response.
Concrete scenario: an opportunity has 50 valid attachments and 1 with a missing `file_name`. All 50 are omitted.
The existing test (`test_attachment_missing_required_fields_is_omitted`) only covers the single-attachment case, so this mixed valid/invalid behavior isn't exercised.
Following the existing pattern in this file (e.g. `validate_url()`), we could try validating and filtering with something like:
```
validated = []
for av in attachment_values:
    try:
        validated.append(AttachmentValue(**av))
    except Exception:
        logger.warning(...)
if validated:
    field = validate_custom_field(AttachmentsField, value=validated)
    if field:
        custom_fields["attachments"] = field
```
Same applies to `assistanceListings` (~line 432), though lower risk since `AssistanceListingValue` fields are all optional.

api/src/api/common_grants/schemas/pydantic/custom_fields.py
```diff
@@ -0,0 +1,199 @@
+"""Pydantic schemas for CommonGrants Protocol customFields.
```
SnowboardTechie, line 1: This new `schemas/` directory is missing an `__init__.py`. Other package directories in the project include one (e.g. `api/src/api/schemas/__init__.py`, `api/src/api/common_grants/__init__.py`). While Python 3.3+ implicit namespace packages may work at runtime, the absence can cause issues with mypy module resolution, pytest collection, and packaging tools. An empty `__init__.py` here would match the project convention.

api/src/api/common_grants/schemas/common_grants_custom_fields.py
```diff
@@ -227,7 +227,7 @@ class fiscalYear(CustomField):
 
     name = fields.String(required=True, metadata={"example": "fiscalYear"})
     fieldType = fields.String(required=True, metadata={"example": "number"})
```
SnowboardTechie: Nit: this metadata still says "number" but the new Pydantic model (`FiscalYearField`, line 180 in `common_grants_pydantic_custom_fields.py`) uses `CustomFieldType.INTEGER`. These are distinct values in the SDK — "number" allows decimals, "integer" is whole numbers only. The change to INTEGER is more correct for fiscal year, but this metadata should be updated to match.

jcrichlake: Fixed, thanks for catching that!

api/src/api/common_grants/schemas/pydantic/custom_fields.py
jcrichlake: Added!

api/src/services/common_grants/transformation.py
jcrichlake: Just pushed a fix for this and assistanceListings

api/src/services/common_grants/transformation.py
```diff
@@ -345,6 +358,27 @@ def transform_search_result_to_cg(opp_data: dict) -> OpportunityBase | None:
...
+    Returns:
+        A validated CustomField instance, or None if validation fails
+    """
+    try:
+        return field_class(**kwargs)
+    except Exception as e:
+        logger.warning(
+            f"Custom field validation failed for {field_class.__name__}: {e}",
+            extra={"cg_event": CommonGrantsEvent.OPPORTUNITY_VALIDATION_ERROR},
+        )
```
widal001, line 386: I know it would require some extra metadata being passed to this function, but it would be helpful if we could include the opportunity ID in the metadata otherwise we'd have a hard time actually troubleshooting these errors.

api/src/api/common_grants/schemas/pydantic/custom_fields.py
```diff
@@ -0,0 +1,199 @@
...
+# Custom Field Implementations
+# ===========================================================================================
+
+
+class LegacySerialIdField(CustomField):
+    """An integer ID for the opportunity, needed for compatibility with legacy systems."""
+
+    name: str = "legacySerialId"
+    field_type: CustomFieldType = CustomFieldType.INTEGER
+    schema_url: HttpUrl | None = HttpUrl("https://commongrants.org/custom-fields/legacySerialId/")
```
widal001, line 76: Hmm I'm on the fence about this. This URL currently maps to an HTML page not the hosted JSON schema.

This page provides some valuable context about this field, but it's not explicitly the schema.

We _could_ host the schemas the same way we do for core fields and have this point to that, or we could leave it as is.

api/src/api/common_grants/schemas/pydantic/custom_fields.py
```diff
@@ -0,0 +1,199 @@
...
+
+class AssistanceListingsField(CustomField):
+    """The assistance listing number and program title for this opportunity."""
+
+    name: str = "assistanceListings"
+    field_type: CustomFieldType = CustomFieldType.ARRAY
+    schema_url: HttpUrl | None = HttpUrl(
+        "https://commongrants.org/custom-fields/assistanceListings/"
+    )
+    value: list[AssistanceListingValue]
```
widal001, line 107: The `AssistanceListingValue` and the shape of the field linked at [`schema_url`](https://commongrants.org/custom-fields/assistanceListings/) don't match.

api/src/api/common_grants/schemas/pydantic/custom_fields.py
```diff
@@ -0,0 +1,199 @@
...
+
+class FederalFundingSourceField(CustomField):
+    """The category type of the grant opportunity."""
+
+    name: str = "federalFundingSource"
+    field_type: CustomFieldType = CustomFieldType.STRING
+    schema_url: HttpUrl | None = HttpUrl(
+        "https://commongrants.org/custom-fields/federalFundingSource/"
+    )
+    value: str
```
widal001, line 147: These don't match either. It's a string here and an object (technically an array of objects, but that was an oversight on my part).

api/src/api/common_grants/schemas/pydantic/custom_fields.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+class ContactInfoField(CustomField):
+    """Contact information for the agency managing this opportunity."""
+
+    name: str = "contactInfo"
+    field_type: CustomFieldType = CustomFieldType.OBJECT
+    schema_url: HttpUrl | None = HttpUrl("https://commongrants.org/custom-fields/contactInfo/")
+    value: ContactInfoValue
+    description: str | None = "Contact information for the agency managing this opportunity"
```
widal001: These don't match either. The value object and the custom field in the registry have different properties.

api/src/api/common_grants/schemas/common_grants_schemas.py
```diff
@@ -7,7 +7,7 @@
 
 from typing import Any
 
-from src.api.common_grants.common_grants_custom_fields import (
+from src.api.common_grants.schemas.common_grants_custom_fields import (
```
widal001: Since we're reorganizing and grouping the schema files, can we think more intentionally about the file names and import paths?

This schemas directory now has three files right?

- `common_grants_custom_fields` This is the _marshmallow_ implementation of custom fields, so it should probably indicate that somewhere?
- `common_grants_pydantic_custom_fields` This is the pydantic equivalent of the file above.
- `common_grants_schemas` This has the top level marshmallow schemas used for (de)serialization at the API route layer, and should probably indicate that.

api/src/api/common_grants/schemas/pydantic/custom_fields.py
jcrichlake: Updated this

api/src/services/common_grants/transformation.py
jcrichlake: Good suggestion! Added

api/src/api/common_grants/schemas/pydantic/custom_fields.py
jcrichlake: Fixed

api/src/api/common_grants/schemas/pydantic/custom_fields.py
widal001: After reflecting further I think we leave these as is!

api/src/api/common_grants/schemas/pydantic/custom_fields.py
```diff
@@ -0,0 +1,206 @@
...
+from common_grants_sdk.schemas.pydantic import CustomField, CustomFieldType
+from pydantic import BaseModel, HttpUrl, ValidationError, field_validator
+
+# ===========================================================================================
+# Value Models
+# ===========================================================================================
+
+
+class AssistanceListingValue(BaseModel):
+    assistanceListingNumber: str | None = None
```
widal001: This attribute is getting dropped because it doesn't match the marshmallow attribute `identifier`

api/src/api/common_grants/schemas/marshmallow/custom_fields.py
widal001: This doesn't match the pydantic schema now

widal001: But the pydantic schema _does_ correctly match the custom field catalog

jcrichlake: This matches the pydantic schema

api/src/api/common_grants/schemas/pydantic/custom_fields.py
jcrichlake: Made this change

PR #9289, opened 2026-03-27
[Issue #9150] Enable OpenSearch Query Explanation Logging

### LINKED ISSUE #9150: Enable OpenSearch Query Explanation Logging
### Summary

When users report poor search results or we notice ranking anomalies in our metrics, we currently have no way to understand why a document ranked where it did. OpenSearch can provide per-field scoring breakdowns that answer this. This does add some additional query overhead so we will want to start with this enabled and monitor the performance impact using the `search.took_ms` metric from #9147, with the ability to quickly disable it via environment variable if we see degradation.

### Tech Notes

Raw OpenSearch _explanation (what comes back):

```                                                              
  {                                                                                                                     
    "_explanation": {                                                                                                   
      "value": 22.18,                                                                                                   
      "description": "sum of:",                                                                                         
      "details": [                                                                                                      
        {                                                                                                               
          "value": 16.5,                                                                                                
          "description": "weight(opportunity_title:transportation in 0) [BM25], result of:",                            
          "details": [...]                                                                                              
        },                                                                                                              
        { 
          "value": 5.68,                                                                                                
          "description": "weight(summary.summary_description:transportation in 0) [BM25], result of:",
          "details": [...]                                                                                              
        }                                                                                                               
      ]                                                                                                                 
    }                                                                                                                   
  } 

```

Example Event: `SearchResultExplanation`                                                                                     
```
{
    correlation_id: "abc-123"
    query: "transportation"
    scoring_rule: "DEFAULT"
    opportunity_id: 4
    position: 1
    total_score: 22.18
    opportunity_title: 16.5
    summary.summary_description: 5.68
    agency_code: 0.0
    opportunity_number: 0.0
}
```

### Acceptance criteria

- [ ] Add environment variable (e.g., `OPENSEARCH_EXPLAIN_ENABLED`, default true) to control whether `explain: true` is sent in the search query
- [ ] When enabled, a SearchResultExplanation custom event is recorded for each of the top 10 results per search
- [ ] Each event includes `correlation_id`, `query`, `scoring_rule`, `opportunity_id`, `position`, `total_score`
- [ ] Each scored field appears as an attribute with its score contribution as the value 
- [ ] Events are queryable in NR via `FROM SearchResultExplanation SELECT ...`
- [ ] unittests
- [ ] Compare `search.took_ms` with explain on vs off to measure performance impact

mikehgrantsgov:
## Summary

Fixes #9150

## Changes proposed

Adds `opensearch_explain.py`. Parses OpenSearch _explanation trees into flat field -> score mappings and emits them as `SearchResultExplanation` custom events in New Relic for the top 10 results per search.
Added `took_ms` and `raw_hits` fields so the search response carries both query latency and the raw hit objects (including _explanation)
Added `OPENSEARCH_EXPLAIN_ENABLED` env var (default: true).
Added explain: bool = False parameter to search() — merges {"explain": True} into the query body when set.
OPENSEARCH_EXPLAIN_ENABLED=true added to environment_variables.tf .

## Context for reviewers

When users report poor search results we have no visibility into why results ranked where they did. This adds OpenSearch explanation logging to surface per-field scoring breakdowns as New Relic SearchResultExplanation custom events for the top 10 results per search, along with search.took_ms instrumentation for latency monitoring.

OPENSEARCH_EXPLAIN_ENABLED (default: true) serves as a kill switch if explain queries cause latency regressions.

## Validation steps

See updated unit tests.

### DIFF
--- a/api/tests/src/adapters/search/test_opensearch_explain.py
+++ b/api/tests/src/adapters/search/test_opensearch_explain.py
@@ -188,9 +188,7 @@ class TestLogSearchResultExplanations:
 
     @patch("src.adapters.search.opensearch_explain.record_custom_event")
     def test_no_boost_when_field_has_no_boost(self, mock_record_event):
-        explanation = _make_explanation(
-            [_make_weight_detail_no_boost("summary_description", 5.0)]
-        )
+        explanation = _make_explanation([_make_weight_detail_no_boost("summary_description", 5.0)])
         hits = [_make_hit("opp-1", "NUM-1", "DOC", 5.0, explanation=explanation)]
         log_search_result_explanations(hits, query="test", scoring_rule="default")
 


joshtonava:
I agree with your concern @babebe that is why we have that kill switch added so we can turn it off if we need to quickly

Issue #9147, opened 2026-03-19
Log OpenSearch query performance metrics on every search request


joshtonava:
### Summary

Today we can't distinguish slow OpenSearch queries from slow API responses, can't detect when users receive incomplete results due to timeouts or shard failures, and have no engine-level search performance baseline. This data is already available in every response, we just need to capture it.

This will enable latency dashboards, timeout/shard-failure alerts, zero-result rate tracking.

Related Document: https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2852945962/OpenSearch+Metrics+Research

**Tech Notes**:
- Add took, timed_out, shards_failed fields to SearchResponse in `api/src/adapters/search/opensearch_response.py`
- Compute score distribution (min, mean, max, stdev) across `hits.hits[*]._score` for each results page
- Add structured logging in `api/src/services/opportunities_v1/search_opportunities.py` using existing `add_extra_data_to_current_request_logs` pattern                                                      
- Derive `is_zero_result` boolean from total_records == 0

### Acceptance criteria

- [ ] Every search request logs `search.took_ms`, `search.timed_out`, `search.shards_failed`, `search.total_records`, `search.is_zero_result` as structured attributes
- [ ] Every search request logs `search.score_min`, `search.score_max`, `search.score_mean`, `search.score_stdev`
- [ ] unit tests 
- [ ] Fields are queryable in New Relic Logs via NRQL

PR #9294, opened 2026-03-27
[Issue #9149] Log OpenSearch Relevance and Data Completeness Metrics 

### LINKED ISSUE #9149: Log OpenSearch Relevance and Data Completeness Metrics
### Summary

Beyond the core performance metrics in #9147, OpenSearch returns relevance and data completeness signals that help us monitor search quality over time. Without these we can't detect scoring degradation, verify result count accuracy, or know when filter facets are silently truncating options. Important for ongoing search quality monitoring as we experiment with scoring changes.

- `max_score` tells us if scoring quality is degrading
- `total.relation` tells us if result counts are becoming approximate instead of exact
- `search_scoring_rule` provides context for which scoring path was used (`DEFAULT`, `AGENCY`, `EXPANDED`, etc.), 
- `sum_other_doc_count` on the agency aggregation reveals when our facet counts are silently truncating agencies. Without these, we'd only notice relevance problems when users complain.

**Tech Note:**

OpenSearch has a `track_total_hits` setting. When it's true (our current config), you get an exact count of matching documents. If it's ever disabled or the count exceeds 10k, OpenSearch returns `hits.total.relation` = "gte" instead of  "eq", meaning "there are at least this many results" rather than an exact number.

### Acceptance criteria

 - Every search request logs `search.max_score`, `search.total_relation`, `search.scoring_rule` as structured attributes     
 - Agency aggregation searches log `search.agency_sum_other_doc_count`
 - unit tests
 - Fields are queryable in New Relic Logs via NRQL

mikehgrantsgov:
## Summary

Fixes #9149 

## Changes proposed

Added max_score, total_relation, scoring_rule, and per-aggregation agg_overflow to structured request logs for OpenSearch searches.

- max_score and total_relation are parsed from the OpenSearch response and logged at the adapter    
layer via `add_extra_data_to_current_request_logs`, consistent with the pattern established in #9296
for `took_ms`, `timed_out`, and `shards_failed`.                                                          
- `agg_overflow` captures sum_other_doc_count for all aggregations generically (not just agency),
logged as `search.agg_overflow.<name>`. Non-zero values indicate a facet is silently truncating       
buckets.
- `scoring_rule` is logged at the service layer since it derives from search params, not the          
OpenSearch response.   

## Context for reviewers

Beyond the core performance metrics in https://github.com/HHS/simpler-grants-gov/issues/9147, OpenSearch returns relevance and data completeness signals that help us monitor search quality over time. Without these we can't detect scoring degradation, verify result count accuracy, or know when filter facets are silently truncating options. Important for ongoing search quality monitoring as we experiment with scoring changes.

## Validation steps

See updated unit tests.

### DIFF
--- a/api/src/services/opportunities_v1/search_opportunities.py
+++ b/api/src/services/opportunities_v1/search_opportunities.py
@@ -6,8 +6,8 @@ from pydantic import BaseModel, Field
 
 import src.adapters.search as search
 from src.adapters.search.opensearch_response import SearchResponse
-from src.logging.flask_logger import add_extra_data_to_current_request_logs
 from src.api.opportunities_v1.opportunity_schemas import OpportunityV1Schema, SearchQueryOperator
+from src.logging.flask_logger import add_extra_data_to_current_request_logs
 from src.pagination.pagination_models import PaginationInfo, PaginationParams, SortDirection
 from src.search.search_config import get_search_config
 from src.search.search_models import (
--- a/api/tests/src/adapters/search/test_opensearch_response.py
+++ b/api/tests/src/adapters/search/test_opensearch_response.py
@@ -3,6 +3,8 @@ Unit tests for SearchResponse parsing edge cases — missing and None fields
 that can't easily be exercised via integration tests against a real OpenSearch instance.
 """
 
+import pytest
+
 from src.adapters.search.opensearch_response import SearchResponse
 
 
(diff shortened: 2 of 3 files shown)


api/tests/src/adapters/search/test_opensearch_client.py
```diff
@@ -259,35 +259,44 @@ def test_search_returns_expected_metadata(search_client, generic_index):
         search_query={"query": {"match_all": {}}},
     )
 
-    # Validate top-level attributes exist
```
joshtonava, line 262: I dont think we need to do this

PR #9296, opened 2026-03-27
[Issue #9147] Add opensearch query performance metrics

babebe:
## Summary

Fixes / Work for #9147  

## Changes proposed

Added `_compute_score_stats` to compute min, max, mean, stdev of relevancy_score from search records.
Added support for top-level search metadata: took_ms, timed_out, shards_failed in SearchResponse.
Ensures type-safety by filtering out None and non-numeric scores.

## Context for reviewers

The goal is to capture search metrics in structured logs for monitoring and analysis.

## Validation steps

Covered in test

### DIFF
--- a/analytics/Dockerfile
+++ b/analytics/Dockerfile
@@ -1,4 +1,4 @@
-FROM ghcr.io/hhs/python-base-image:b94fa25@sha256:11c8f0e9766de10bc7a986bd16e4cdf7f05c3e357e48efeb6af0f98c5bdb151b AS base
+FROM ghcr.io/hhs/python-base-image:8db23e7@sha256:8a6fedd44b3255e975e24160e3b58335aaf71dbbae8776d130986015acd11305 AS base
 
 ARG RUN_UID
 ARG RUN_USER
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -1,6 +1,4 @@
-
-FROM ghcr.io/hhs/python-base-image:b94fa25@sha256:11c8f0e9766de10bc7a986bd16e4cdf7f05c3e357e48efeb6af0f98c5bdb151b AS base
-
+FROM ghcr.io/hhs/python-base-image:8db23e7@sha256:8a6fedd44b3255e975e24160e3b58335aaf71dbbae8776d130986015acd11305 AS base
 ARG RUN_UID
 ARG RUN_USER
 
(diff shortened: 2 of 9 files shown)


api/src/adapters/search/opensearch_client.py
```diff
@@ -211,14 +212,28 @@ def search(
...
             body=search_query,
             params=params,
             _source_includes=includes,
             _source_excludes=excludes,
         )
-        return SearchResponse.from_opensearch_response(response, include_scores)
+
+        response = SearchResponse.from_opensearch_response(raw_response, include_scores)
+        # Structured logging enrichment
+        add_extra_data_to_current_request_logs(
```
joshtonava, line 225: **(non blocking) note:** the search method is also called in some background tasks in which cases `add_extra_data_to_current_request_logs` will silently fail since there is no request context. Not a big deal just thought I would mention.

api/tests/src/adapters/search/test_opensearch_client.py
```diff
@@ -239,3 +244,51 @@ def test_cleanup_old_indices(search_client):
...
+    response = search_client.search(
+        index_name=generic_index,
+        search_query={"query": {"match_all": {}}},
+    )
+
+    # Validate top-level attributes exist
+    assert hasattr(response, "took_ms")
+    assert hasattr(response, "timed_out")
+    assert hasattr(response, "shards_failed")
+    assert hasattr(response, "shards_failed")
```
joshtonava: duplicate, was this supposed to be another field?

api/src/adapters/search/opensearch_response.py
```diff
@@ -62,8 +71,18 @@ def from_opensearch_response(
 
         raw_aggs: dict[str, dict[str, typing.Any]] = raw_json.get("aggregations", {})
         aggregations = _parse_aggregations(raw_aggs)
+        score_stats = _compute_score_stats(records)
```
joshtonava: **suggestion:** We may want to skip the computation when the scores arent included (include_scores = False)

Something like 

```python
score_stats = _compute_score_stats(records) if include_scores else {}
```

api/src/adapters/search/opensearch_client.py
```diff
@@ -211,14 +212,28 @@ def search(
...
+            {
+                "search.took_ms": response.took_ms,
```
joshtonava, line 228: Can we also add in `"search.index": index_name` here?

api/src/adapters/search/opensearch_client.py
babebe: Yes, I figured that’s okay —we are only logging in API requests, so background tasks can safely skip it.

api/src/adapters/search/opensearch_response.py
babebe: Adapted change

api/tests/src/adapters/search/test_opensearch_client.py
babebe: Removed duplicate.

Issue #8969, opened 2026-03-10
[FE] Make the saved opportunity star icon only look at individual saved opportunities


freddieyebra:
### Summary

Currently on the opportunity detail page, the icon for indicating whether the opportunity has been saved checks against the list of "All saved opportunities" instead of just the "Individual saved opportunities". This is causing the icon to indicate that the opportunity has been saved to the individual which is wrong.

We need to update the logic for this icon to check against the user individual saved opportunities instead of all saved opportunities (individual + org saved opportunities).

### Acceptance criteria

- [ ] Saved icon on opportunity details only reflects individual saved opportunity
    - [ ] If a user stars an opportunity, goes to the saved opportunity page and adds an organization, goes to the opportunity page and unstars it, then when the refresh the opportunity page the star should remain unstarred.
- [ ] Rules should also apply to the star/saved icon on the search page.
- [ ] Unit test

freddieyebra:
Summary of how the existing save functionality should work:

The existing places that a user can click the star icon to save/un-save an opportunity should and will remain the same which will save/un-save the opportunity to their own individual saved opportunities.

An org saved opportunity can only be added or removed through the modal when the "show all" filter is applied or the list by organization is applied. This does create an additional step of saving the opportunity first their own individual opportunities and needing to add it to the org through modal.

If a user saves an opportunity to their individual list and then saves/shares it to other orgs, and then removes it by clicking the star icon, it will be removed from the individual saved opportunities and the only way they would be able to see that org saved opportunity is by setting the filter to show the saved organizations for a specific org or the "show all" filter.

The functionality of save/un-saving through the icon should not change in the search results page or through the opportunity details page which is the issue that came up for the modal pr. The opportunities that are listed are depend on the filter that is set in the ui.

PR #9299, opened 2026-03-27
[Issue #8969] Ensure save opportunity icon works as expected

freddieyebra:
## Summary

Fixes / Work for #8969 

## Changes proposed

This pr adds logic for ensuring that the save opportunity icon on the search results and opportunity detail page correctly reflects the saved status against the users individually saved opportunities.

This pr also includes methods for determining the scope of the request based on organization_id and scope params. It will only return the default until filters are supported.

## Context for reviewers

Prior to this change, the saved opportunity icon was checking if the opportunity was saved against the "all" scope of saved opportunity instead of the "individual" scope saved opportunities response. This updates those checks to check against the individual saved opportunities.

## Validation steps

1. Ensure db is seeded by running `make db-seed-local`
2. Find the `isShareWithOrganizationEnabled` and set it to true and run the api and frontend
3. Login as two_org_user
4. Navigate to workspaces > organizations and ensure you have 2 organizations listed
5. Navigate to the saved opportunities page and assert there are not saved opportunities
6. Navigate to the search page and save 2 opportunities by clicking the star icon
7. Navigate to the saved opportunities page and verify 2 saved opportunities
8. On one saved opportunity listing, click on the title to navigate to the opportunity detail view.
9. Verify the saved icon correctly reflects the saved status (yellow)
10. Click to un-save the opportunity and navigate back to the saved opportunities page. Refresh and verify the opportunity is no longer there
11. On the other saved opportunity, click on the "Share with organization" to open the modal and select one of the organizations to save the opportunity to.
12. Now click on that opportunity title to navigate to the opportunity detail page.
13. Verify the saved icon correctly reflects the saved status (yellow)
14. Click the star icon to un-save it and navigate back to the saved opportunities page.
15. Verify you still see the opportunity in the saved opportunity list and verify there is a checked organization in the share modal.
16. Navigate to the search page to find the opportunity in the search results and verify the saved icon does not indicate it is saved
17. Go back to the saved opportunities page and uncheck the organization in the modal and verify the saved opportunity page renders empty

### DIFF
--- a/frontend/src/services/fetch/fetchers/savedOpportunityFetcher.ts
+++ b/frontend/src/services/fetch/fetchers/savedOpportunityFetcher.ts
@@ -86,7 +86,7 @@ export const getSavedOpportunities = async (
     additionalHeaders: ssgToken,
     body,
   });
-  const json = (await resp.json()) as { data: [] };
+  const json = (await resp.json()) as { data: MinimalOpportunity[] };
   return json.data;
 };
 


myduong-navapbc:
i believe we have a typo: `savedOpportunitiiesUtils`, two `ii`s

frontend/src/services/fetch/fetchers/savedOpportunityFetcher.ts
```diff
@@ -80,12 +90,16 @@ export const getSavedOpportunities = async (
   return json.data;
 };
 
-export const getSavedOpportunity = async (
+export const getUserSavedOpportunity = async (
```
myduong-navapbc, line 93: I like this, very concise 👍

frontend/src/services/fetch/fetchers/savedOpportunityFetcher.ts
myduong-navapbc: I know you didnt do this part but, we now know what return: 
`as { data: MinimalOpportunity[]`

myduong-navapbc:
Overall this looks really solid 👍 nice work here.

The scope-based approach makes the intent much clearer, and I like how the saved-opportunity check is now explicitly tied to the individual scope instead of relying on the broader saved list. The changes feel well-contained and easy to follow.

The new utility tests are also great - they cover the scope logic well and make the behavior easier to reason about.

Just a couple small things:

Minor typo in `savedOpportunitiiesUtils` filename (extra `i`)
Might be worth adding one test closer to the fetch/handler path to cover the individual-vs-org edge case

Otherwise looks good to me - appreciate the thoughtful implementation here.

PR #9300, opened 2026-03-27
[Issue #8872] Opportunity publish workflow writes to search index

### LINKED ISSUE #8872: Make the opportunity publish workflow conditionally write the opportunity to the search index
### Summary

When handling the `write_opportunity_to_search` event we want to add conditional logic:
* If the opportunity has current opportunity summary OR the opportunity's agency is a test agency, transition it to the end state
* Otherwise we want to write the opportunity to the search index (Details below)

For doing a conditional transition, see https://python-statemachine.readthedocs.io/en/latest/guards.html - we also have a few of these defined on the initial workflow we've built for approvals. The conditional function we build should just need to check if the opportunity's current opportunity summary is null + whether the agency is a test agency.

---

For writing to search, we have a [backend task that does this](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/search/backend/load_opportunities_to_index.py) - it does quite a bit more than we need to worry about, mainly we want to:

* Convert the opportunity to JSON with `OpportunityV1Schema().dump(opportunity)`
* Write the opportunity to the search index. Use the alias_name from the load opportunity to search index config. Other parameters should be the same as the search index parameter.
* We should have basic retries like that logic does - the requests are occasionally a bit flakey, although we're much less likely to hit an issue 

We'll need to figure out how we get a search client. Ideally we wouldn't be making a new one each time we run this, but the alternative is likely instantiating it when the task starts up with the `@flask_opensearch.with_search_client()` decorator, which'll require quite a bit of piping through the client, and would make every test need to be updated. I'm fine with us setting it up at the top-level, and maybe registering it somewhere that we can reference globally? That way only tests that use it need to do that setup. 

### Acceptance criteria

- [ ] Conditional logic added to branch search load event based on opportunity summary + test agency
- [ ] Opportunity loaded to search index when values set as appropriate
- [ ] Search client is setup in an easy to use way that avoids remaking search clients every request
- [ ] Tests added for validating the different scenarios

chouinar:
## Summary
Fixes #8872

## Changes proposed
* In the opportunity publish workflow, write the opportunity to the search index
* Adds a workflow client registry for getting the search client when needed in workflow logic

## Context for reviewers
The implementation here is slightly different than the ticket. There's no branching logic with conditionals if there isn't a current opportunity summary because adding that would have needed more steps/events as that check would have to happen after we handle setting the opportunity summary in the prior step. So regardless of what we do, the same events would happen, we could add a conditional so it doesn't even go into the new handler function, but an if-statement at the start seems more straightforward / intuitive. Will leave that to the publish folks to sort out in the future if they don't like the history this leaves.

In the interest of not needing to plumb more clients through in the future (like an eventual client for integrating our two services with each other) especially when they're only needed for specific workflows - we'll instead have this global registry that we can use in a very similar manner. It might be we want to adjust how it works down the road, but the search client is fine to be shared like this - we initialize one client and share it across all endpoints in the API, so it's well tested.

## Validation steps
Make sure your API is running locally.

Run `make console` and create an opportunity that won't have a current summary, but will get one.

```py
opportunity = f.OpportunityFactory.create(is_draft=True, no_current_summary=True)
f.OpportunitySummaryFactory.create(opportunity=opportunity, is_forecast=False)
```

Send an event to the event API using the `internal_workflow_user_key` API key, setting the opportunity ID from what you created above.
```json
{
  "metadata": {},
  "start_workflow_context": {
    "entity_id": "4b1cf004-1034-407b-90ab-9e16c2008ad5",
    "entity_type": "opportunity",
  }
}
```

Run the workflow service with: `make cmd args="workflow workflow-main"` - you shouldn't see any errors processing the event.

The opportunity you created should now be viewable in search locally (but wasn't before running):
Screenshot 2026-03-27 at 2 56 53 PM


api/src/workflow/state_machine/opportunity_publish_state_machine.py
```diff
@@ -158,3 +167,52 @@ def handle_calculate_current_opportunity_summary(
...
+
+        schema = OpportunityV1Schema()
+        records = [schema.dump(self.opportunity)]
+        config = SearchConfig()
+        search_client = get_workflow_client_registry().search_client
+
+        log_extra |= {"search_index_alias": config.opportunity_search_index_alias}
+        logger.info("Writing opportunity to search index", extra=log_extra)
+
+        # Writing to the search index is a nice-to-have, in the vent there
```
kkrug:
```suggestion
        # Writing to the search index is a nice-to-have, in the event there
```

api/src/workflow/state_machine/opportunity_publish_state_machine.py
```diff
@@ -158,3 +167,52 @@ def handle_calculate_current_opportunity_summary(
...
+                "Couldn't write opportunity to search index to due to intermittent issue",
+                extra=log_extra,
+                exc_info=True,
+            )
+            state_machine_event.increment(self.Metrics.OPP_PUBLISH_ERROR_WRITING_TO_SEARCH_INDEX)
+        except Exception:
+            logger.exception("Failed to write opportunity to search index", extra=log_extra)
+            state_machine_event.increment(self.Metrics.OPP_PUBLISH_ERROR_WRITING_TO_SEARCH_INDEX)
+
+        state_machine_event.increment(self.Metrics.OPP_PUBLISH_WRITTEN_TO_SEARCH_INDEX)
```
kkrug: Should the increment of `Metrics.OPP_PUBLISH_WRITTEN_TO_SEARCH_INDEX` be within the `try` block?

chouinar: That would make more sense, will fix

PR #9301, opened 2026-03-27
Fix 404 pages returning HTTP 200 by using notFound() instead of JSX

### LINKED ISSUE #5774: Opportunity Not Found page is not returning a 404 Status
### Summary

Previously Opportunities that were not found in the data would return a proper 404 Status Code. The page currently does not do that, at least for Int IDs. This might be related to our Int to UUID conversion/redirect logic.
https://simpler.grants.gov/opportunity/359212

### Reproducibility

None

### How to reproduce the bug

Access the https://simpler.grants.gov/opportunity/359212 or another Int ID that doesn't exist and see that the Status code in the Network tab is a 200 not a 404 even though the language about Not Found is displayed.

### Browser

Not applicable

### Code of Conduct

vijaygovindaraja:
## Summary
- Pages for opportunities, applications, and forms were returning **HTTP 200** instead of **404** when resources weren't found
- Root cause: components used `return <NotFound />` (renders JSX) instead of calling `notFound()` (Next.js function that sets the 404 status code)
- Fixed across **4 pages** that had the same bug pattern

## Root cause

```tsx
// Before — renders "not found" UI but HTTP status stays 200
if (parseErrorStatus(error) === 404) {
  return <NotFound />;
}

// After — Next.js sets HTTP 404 and renders the nearest not-found.tsx boundary
if (parseErrorStatus(error) === 404) {
  notFound();
}
```

In Next.js App Router, `notFound()` from `next/navigation` throws a special error that Next.js catches to:
1. Set the HTTP response status to 404
2. Render the nearest `not-found.tsx` boundary

Simply returning a `<NotFound />` component as JSX renders the UI but skips the status code — the response is still 200.

## Files changed

| File | Change |
|------|--------|
| `opportunity/[id]/page.tsx` | `return <NotFound />` → `notFound()`, remove unused import |
| `applications/[applicationId]/page.tsx` | Same fix, add `notFound` to import |
| `applications/[applicationId]/form/[appFormId]/page.tsx` | Same fix, add `notFound` to import |
| `print/.../form/[appFormId]/page.tsx` | Same fix, add `notFound` to import |

## Test plan
- [ ] Visit `/opportunity/359212` (nonexistent int ID) → verify HTTP 404 in Network tab
- [ ] Visit `/opportunity/00000000-0000-0000-0000-000000000000` (nonexistent UUID) → verify HTTP 404
- [ ] Visit a valid opportunity → verify HTTP 200 and content loads normally
- [ ] Visit `/applications/nonexistent-id` → verify HTTP 404
- [ ] Verify the "not found" UI still renders correctly on all pages

Fixes #5774

### DIFF
--- a/frontend/src/app/[locale]/(base)/applications/[applicationId]/page.tsx
+++ b/frontend/src/app/[locale]/(base)/applications/[applicationId]/page.tsx
@@ -1,6 +1,5 @@
 import { Metadata } from "next";
 import TopLevelError from "src/app/[locale]/(base)/error/page";
-import NotFound from "src/app/[locale]/(base)/not-found";
 import { ApiRequestError, parseErrorStatus } from "src/errors";
 import { getSession } from "src/services/auth/session";
 import withFeatureFlag from "src/services/featureFlags/withFeatureFlag";
@@ -13,7 +12,7 @@ import { getOpportunityDetails } from "src/services/fetch/fetchers/opportunityFe
 import { OpportunityDetail } from "src/types/opportunity/opportunityResponseTypes";
 
 import { getTranslations } from "next-intl/server";
-import { redirect } from "next/navigation";
+import { notFound, redirect } from "next/navigation";
 import { GridContainer } from "@trussworks/react-uswds";
 
 import ApplicationContainer from "src/components/application/ApplicationContainer";
@@ -78,7 +77,7 @@ async function ApplicationLandingPage({ params }: ApplicationLandingPageProps) {
         `Error retrieving application details for application (${applicationId})`,
         e,
       );
-      return <NotFound />;
+      notFound();
     }
     return <TopLevelError />;
   }
--- a/frontend/src/app/[locale]/(base)/search/page.tsx
+++ b/frontend/src/app/[locale]/(base)/search/page.tsx
@@ -13,6 +13,7 @@ import { useTranslations } from "next-intl";
 import { getTranslations, setRequestLocale } from "next-intl/server";
 import { Suspense, use } from "react";
 
+import Loading from "src/components/Loading";
 import { DrawerUnit } from "src/components/drawer/DrawerUnit";
 import { AndOrPanel } from "src/components/search/AndOrPanel";
 import { ClassicSearchBanner } from "src/components/search/ClassicSearchBanner";
@@ -98,11 +99,15 @@ function Search({ searchParams, params }: SearchPageProps) {
                   iconName="filter_list"
                   buttonClass="tablet:margin-x-auto"
                 >
-                  <SearchDrawerFilters
-                    searchParams={convertedSearchParams}
-                    searchResultsPromise={searchResultsPromise}
-                    agencyListPromise={filteredAgencyListPromise}
-                  />
+                  <Suspense
+                    fallback={<Loading message={t("drawer.loading")} />}
+                  >
+                    <SearchDrawerFilters
+                      searchParams={convertedSearchParams}
+                      searchResultsPromise={searchResultsPromise}
+                      agencyListPromise={filteredAgencyListPromise}
+                    />
+                  </Suspense>
                 </DrawerUnit>
               </div>
               <div className="flex-3 flex-align-self-end display-none desktop:display-block">
--- a/frontend/src/app/[locale]/(base)/search/page.tsx
+++ b/frontend/src/app/[locale]/(base)/search/page.tsx
@@ -13,8 +13,8 @@ import { useTranslations } from "next-intl";
 import { getTranslations, setRequestLocale } from "next-intl/server";
 import { Suspense, use } from "react";
 
-import Loading from "src/components/Loading";
 import { DrawerUnit } from "src/components/drawer/DrawerUnit";
+import Loading from "src/components/Loading";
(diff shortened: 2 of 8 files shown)


btabaska:
Thank you so much for this contribution. We will look into this early next week.

frontend/src/app/[locale]/(print)/print/application/[applicationId]/form/[appFormId]/page.tsx
```diff
@@ -65,7 +65,7 @@ export default async function FormPage({ params }: FormPageProps) {
   });
 
   if (error || !data) {
-    if (error === "NotFound") return <NotFound />;
+    if (error === "NotFound") notFound();
```
Comment (btabaska) [inline frontend/src/app/[locale]/(print)/print/application/[applicationId]/form/[appFormId]/page.tsx:68]: I could be wrong, but have we confirmed this route has a reachable not-found.tsx boundary in its ancestor tree? We only seem to have app/[locale]/(base)/not-found.tsx, and (base) is a sibling of (print), not an ancestor. If that’s right, notFound() here may fall back to Next’s default 404 UI instead of our custom one. We’ll want to verify this in-browser and, if needed, add a shared ancestor not-found.tsx (or a (print) one) so folks get the expected not-found experience plus the correct 404 status.

frontend/src/app/[locale]/(base)/applications/[applicationId]/page.tsx
```diff
@@ -78,7 +77,7 @@ async function ApplicationLandingPage({ params }: ApplicationLandingPageProps) {
         `Error retrieving application details for application (${applicationId})`,
         e,
       );
-      return <NotFound />;
+      notFound();
```
Comment (btabaska) [inline frontend/src/app/[locale]/(base)/applications/[applicationId]/page.tsx:80]: Nice fix moving to notFound(

frontend/src/app/[locale]/(base)/search/page.tsx
```diff
@@ -98,11 +99,15 @@ function Search({ searchParams, params }: SearchPageProps) {
                   iconName="filter_list"
                   buttonClass="tablet:margin-x-auto"
                 >
-                  <SearchDrawerFilters
-                    searchParams={convertedSearchParams}
-                    searchResultsPromise={searchResultsPromise}
-                    agencyListPromise={filteredAgencyListPromise}
-                  />
+                  <Suspense
```
Comment (btabaska) [inline frontend/src/app/[locale]/(base)/search/page.tsx:102]: Love this Suspense boundary fix.

btabaska:
It looks like the CI/CD checks are failing on linting. If you could run `npm run format` and commit those changes, that should fix the issue.

Small nit: We should add/extend automated coverage asserting actual HTTP 404 status for these not-found routes (opportunity/application/form/print form), not just UI rendering. This is the core behavior this PR changes, and having one regression test here will keep us from slipping back to 200 + not-found UI later. We could do this in follow-on work if you don't want to add that to this PR

Overall: Very close to done, thank you so much for taking this backlog bug and fixing it.

vijaygovindaraja:
@btabaska Ran the formatter and added a regression test for the opportunity page asserts that `notFound()` is called on 404 and that non-404 errors propagate normally. Kept the scope to the opportunity route since it's the most common path; happy to extend to the application/form routes if you'd prefer that in this PR.

btabaska:
Thanks for running that command. I just ran the CI/CD checks and it looks like I am still seeing some failures. Claude should be able to clean these up pretty quickly. 



/home/runner/work/simpler-grants-gov/simpler-grants-gov/frontend/src/app/[locale]/(base)/opportunity/[id]/page.test.tsx
Error:   94:7  error  'mockOpportunityData' is assigned a value but never used  @typescript-eslint/no-unused-vars

/home/runner/work/simpler-grants-gov/simpler-grants-gov/frontend/src/components/RootLayoutWrapper.tsx
Warning:   57:7  warning  Do not use `<head>` element. Use `<Head />` from `next/head` instead. See: https://nextjs.org/docs/messages/no-head-element  @next/next/no-head-element

/home/runner/work/simpler-grants-gov/simpler-grants-gov/frontend/src/components/opportunities/create/CreateOpportunityForm.tsx
Warning:   71:6  warning  React Hook useEffect has a missing dependency: 'selectedCategoryId'. Either include it or remove the dependency array  react-hooks/exhaustive-deps

✖ 3 problems (1 error, 2 warnings)

Error: Process completed with exit code 1.


No need to extend the tests further past what you have already proposed. Thanks again for the effort!

vijaygovindaraja:
@btabaska Removed the unused variable. The two warnings are in files outside this PR.

btabaska:
> @btabaska Removed the unused variable. The two warnings are in files outside this PR.

I think we just need 1 more tiny change to get the typescript checks to pass. 

Exact lines to change in frontend/src/app/[locale]/(base)/opportunity/[id]/page.test.tsx:
`notFound: (...args: unknown[]) => mockNotFound(...args),`

should be changed to

`notFound: () => mockNotFound(),`

After that just run npm run format one more time to be safe with that one line change, and this will be ready to merge. Thanks for your patience while we work through the CI/CD failures. We will be sure to document the requirements that our CI/CD checks against in the near future to prevent this back and forth. Appreciate your help!

vijaygovindaraja:
@btabaska Done, updated and ran format. Thanks for walking me through the CI requirements.

btabaska (approved):

Issue #9249, opened 2026-03-25
Use Trussworks Character Count in Shared Grantor Form Components


doug-s-nava:
### Summary

The shared grantor facing form field components are using a custom implementation of a character count that may be out of step with USWDS, and should be replaced with the Trussworks CharacterCount component

* Fields in question: https://github.com/HHS/simpler-grants-gov/blob/main/frontend/src/components/grantor/CommonFormFields.tsx
* Trussworks component: https://github.com/trussworks/react-uswds/blob/main/src/components/forms/CharacterCount/CharacterCount.tsx

### Acceptance criteria

- [ ] custom implementation of character count replaced with Trussworks CharacterCount

Dao-REI:
For UAT I demo'ed this ticket to the team at Stand-up After Party on 4/1/2026, including @crystabelrangel, @doug-s-nava, @glenblosser-nava and @desaiamit. Now moving this ticket to Done.

PR #9304, opened 2026-03-27
[Issue #9249] Use Trustworks Character Count in Shared Grantor Form Components

Dao-REI:
## Summary

Work for #9249

## Changes proposed

- Update the Grantor's common components, TextInput and Textarea, with TrustWorks' CharacterCount component.
- Update the CreateOpportunityForm that uses the Grantor's common components
- Update test cases

## Context for reviewers

The Grantor's common components are planned to be used as shared/common components by other pages for SSG in the future.

## Validation steps

1. Code Review
2. UAT:  Open the Create Opportunity page, e.g. http://localhost:3000/opportunities/create/4dcd3a02. For the Opportunity Number field:
     1. Verify the Character Count (below the textbox) reads, "40 characters allowed"
     2. As you type into the textbox, verify that the Character Count starts decrementing and reads, "x characters left"
     3. Continue to enter text past the maximum number of characters allowed. Verify that the negative character count increases and reads, "x characters over limit"
     4. Enter values into the other required fields to enable the 'Save and continue' button. Leave the Opportunity Number field with some characters over the limit. Click on the Save button. Verify that an error message pops up pointing to the Opportunity Number field and reads, "The content is too long".

### DIFF
--- a/frontend/src/components/grantor/CommonFormFields.tsx
+++ b/frontend/src/components/grantor/CommonFormFields.tsx
@@ -12,8 +12,6 @@ import {
   TextInput,
 } from "@trussworks/react-uswds";
 
-import "@trussworks/react-uswds/lib/index.css";
-
 // ----------------------------------------------------------
 // Common Text with the same styles
 // ----------------------------------------------------------


frontend/src/components/grantor/CommonFormFields.tsx
```diff
@@ -97,22 +96,21 @@ export const CommonTextInput = ({
           isRequired={isRequired}
           validationError={validationError}
         />
-        <TextInput
-          type="text"
-          name={fieldId}
+        <CharacterCount
```
doug-s-nava: I don't know if we want to make the assumption that all text and textarea fields will want a character count. Maybe something to confirm with design. If not, I'd make it clear in the name of the components that these include character counts

frontend/src/components/grantor/CommonFormFields.tsx
```diff
@@ -135,6 +133,7 @@ export const CommonTextArea = ({
   defaultValue?: string;
   validationError?: string;
 }) => {
+  const ariaDescBy = fieldId + "-info " + fieldId + "-hint";
```
doug-s-nava: not sure how this is supposed to be working. Assume you're trying to link this to the id of the label component for the field? It looks to me lie that should be more like `label-for-${fieldId}`

Dao-REI: You're right, the aria-describedby tag is not necessary. I was following one of the examples, but I should have removed this. Done.

frontend/src/components/grantor/CommonFormFields.tsx
Dao-REI: Sure. I have renamed the component to CommonCharacterCount, put back CommonTextInput and CommonTextArea, and updated the test cases.

frontend/src/components/grantor/CommonFormFields.test.tsx
```diff
@@ -139,6 +140,71 @@ describe("CommonTextArea", () => {
   });
 });
 
+// --- Test Common CharacterCount ---
+let bigTextValue2 = "";
+const onCharacterCountChange = (e: React.ChangeEvent<HTMLTextAreaElement>) => {
```
doug-s-nava: this works, but it'll be easier to manage if you use a mock function, since you don't actually care what the change function does, just that it's called with the expected arguments.

frontend/src/components/grantor/CommonFormFields.tsx
```diff
@@ -211,7 +261,7 @@ export const CommonSelectInput = ({
           name={fieldId}
           onChange={handleChange}
           value={selectedValue}
-          style={{ maxWidth: "550px" }}
+          style={{ maxWidth: "480px" }}
```
doug-s-nava: can we use a USWDS class here

frontend/src/components/grantor/CommonFormFields.tsx
doug-s-nava: I wasn't saying that the aria-describedby wasn't necessary as much as I don't think it was pointing to the right element. I think it is still nice to have, and likely we will need to add it later. Not a blocker, but feel free to add it back here rather than doing it later if you want.

frontend/src/components/grantor/CommonFormFields.tsx
Dao-REI: I applied the USWDS class, maxw-mobile-lg = 480px

frontend/src/components/grantor/CommonFormFields.tsx
```diff
@@ -12,6 +12,8 @@ import {
   TextInput,
 } from "@trussworks/react-uswds";
 
+import "@trussworks/react-uswds/lib/index.css";
```
doug-s-nava: shouldn't need this

PR #9308, opened 2026-03-30
[unticketed]: Increase the memory for staging api container

prasnava:
## Summary

In the staging environment, we are seeing 5xx errors; investigated the load balancers logs. The ecs container showed an exit code status of 137; indicating the container for forcibly exited. 

Increased the memory for the staging api container from 2gb to 4gb. 

## Validation steps

The step function jobs and the api should continue to work without producing 5xx errors in the staging environment.


prasnava:
Deployed to staging: [here](https://github.com/HHS/simpler-grants-gov/actions/runs/23748954067/job/69185241739)

Issue #9254, opened 2026-03-25
SOAP/Proxy: Update Use-Soap-Jwt to be on by default


jakobpederson:
### Summary

In order to facilitate testing by an outside party we need to flip the Use-Soap-Jwt to be on by default and flipped off when in the header.

Currently `"Use-Soap-Jwt"` is a header that when set to `'1'` will send the request down the jwt path. Since the people testing will be using that, we probably want that to be the default instead. We want to update to something like `Use-Proxy` in order to control that.

Replace `'Use-Soap-Jwt'` with `'Use-Soap-Cert'`.

### Acceptance criteria

- [ ] Can control legacy vs simpler path with the header

PR #9310, opened 2026-03-30
[Issue #9254] Use SOAP jwt auth by default

jakobpederson:
## Summary

Fixes / Work for #9254  

## Changes proposed

Swapped `Use-Soap-Jwt` for `Use-Soap-Cert` header and updated code to check against the `Use-Soap-Cert`.

## Context for reviewers

We want to make the soap jwt auth the default path so we swapped in `Use-Soap-Cert` in order to flag the path to use the older cert auth.

## Validation steps

- [ ] tests pass
- [ ] Use the `Use-Soap-Cert: 1` in dev and check the logs to see if the `use jwt auth` log appears

### DIFF
--- a/api/tests/lib/data_factories.py
+++ b/api/tests/lib/data_factories.py
@@ -150,7 +150,7 @@ def setup_cert_user(
 
 
 def create_soap_request(
-    soap_payload: bytes, use_soap_cert: bool = True, log_local: bool = False
+    soap_payload: bytes, use_soap_cert: bool = False, log_local: bool = False
 ) -> SOAPRequest:
     _, _, soap_certificate = setup_cert_user(
         AgencyFactory.create(), [Privilege.LEGACY_AGENCY_VIEWER]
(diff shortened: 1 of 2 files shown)


api/tests/lib/data_factories.py
```diff
@@ -150,7 +150,7 @@ def setup_cert_user(
 
 
 def create_soap_request(
-    soap_payload: bytes, use_soap_jwt: bool = False, log_local: bool = False
+    soap_payload: bytes, use_soap_cert: bool = True, log_local: bool = False
```
kkrug: We default changed from `use_soap_jwt=False` (header not set) to `use_soap_cert=True` (header IS set). This means calls to `create_soap_request(SOAP_PAYLOAD)` without an explicit argument now produce a request with the `Use-Soap-Cert: 1` header. This sends the request down the cert path by default in tests.

This seems like it's the opposite of what production will see (no header = JWT path). It might be better to default `use_soap_cert=False` in the factory to match production defaults, and explicitly pass `use_soap_cert=True` in tests that need the cert path. It's not really a functional difference but I think it's more representative of how it will work in reality.

Issue #9319, opened 2026-03-30
Scheduled E2E jobs failing


doug-s-nava:
### Summary

After the change in https://github.com/HHS/simpler-grants-gov/commit/fe75b3387c792a230143b8e3335be6de35ccd00b some things are not working with scheduled jobs:

* jobs are not properly parallel - if local job fails, staging job is cancelled and vice versa
* job reports trying to save with the same name for local and staging jobs
* staging jobs are not running on push to main
* concurrency groups do not account for valid runs across environments (runs get cancelled because they are coming from the same job for the same git hash)

### Reproducibility

None

### Browser

None

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #9320, opened 2026-03-30
[Issue #9319] fix errors on scheduled e2e jobs

doug-s-nava:
## Summary

Work for #9319

## Changes proposed

Fixes a number of issues with scheduled e2e test runs as introduced in https://github.com/HHS/simpler-grants-gov/commit/fe75b3387c792a230143b8e3335be6de35ccd00b

* jobs are not properly parallel - if local job fails, staging job is cancelled and vice versa
* job reports trying to save with the same name for local and staging jobs
* staging jobs are not running on push to main
* concurrency groups do not account for valid runs across environments (runs get cancelled because they are coming from the same job for the same git hash)

## Validation steps

1. _VERIFY_: test job ran "sucessfully" https://github.com/HHS/simpler-grants-gov/actions/runs/23753844466/job/69203072233

This means that:
* both jobs were allowed to fully run even though the other failed
* reports were created for both staging a local (yeah I know local failed, but it was only because the local tests didn't fully run because of a blip)

### DIFF
--- a/.github/workflows/e2e-daily.yml
+++ b/.github/workflows/e2e-daily.yml
@@ -13,11 +13,11 @@ jobs:
     with:
       version: main
       target: staging
-      playwright_tags: "@full-regression"
+      playwright_tags: "@smoke|@core-regression|@full-regression"
   daily-e2e-tests-local:
     name: Daily E2E tests (local)
     uses: ./.github/workflows/ci-frontend-e2e.yml
     if: always()
     secrets: inherit
     with:
-      playwright_tags: "@full-regression"
+      playwright_tags: "@smoke|@core-regression|@full-regression"
--- a/.github/workflows/e2e-staging.yml
+++ b/.github/workflows/e2e-staging.yml
@@ -47,7 +47,7 @@ jobs:
         shell: bash
         run: |
           if [[ $GITHUB_EVENT_NAME = "push" ]]; then
-            echo "test_group_tags=@core-regression" >> "$GITHUB_ENV"
+            echo "test_group_tags=@smoke|@core-regression" >> "$GITHUB_ENV"
           else
             echo "test_group_tags=${{ inputs.playwright_tags }}" >> "$GITHUB_ENV"
           fi
--- a/.github/workflows/e2e-weekly.yml
+++ b/.github/workflows/e2e-weekly.yml
@@ -13,11 +13,11 @@ jobs:
     with:
       version: main
       target: staging
-      playwright_tags: "@extended"
+      playwright_tags: "@smoke|@core-regession|@full-regression|@extended"
   weekly-e2e-tests-local:
     name: Weekly - all E2E tests (local)
     uses: ./.github/workflows/ci-frontend-e2e.yml
     if: always()
     secrets: inherit
     with:
-      playwright_tags: "@extended"
+      playwright_tags: "@smoke|@core-regession|@full-regression|@extended"
--- a/.github/workflows/cd-release.yml
+++ b/.github/workflows/cd-release.yml
@@ -91,7 +91,7 @@ jobs:
     uses: ./.github/workflows/ci-frontend-e2e.yml
     secrets: inherit
     with:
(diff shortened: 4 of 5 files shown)


frontend/Makefile
```diff
@@ -73,3 +73,9 @@ load-test-staging: # Load test the staging environment in aws
 
 load-test-prod: # Load test the production environment in aws. Please test responsibly
 	artillery run -e prod artillery-load-test.yml
+
+test-group-counts: # Print the number of tests currently assigned to each test group
```
doug-s-nava, line 77: can be used for debugging to make sure CI is running the right number of tests for each group

.github/workflows/e2e-weekly.yml
```diff
@@ -1,19 +1,23 @@
...
+    secrets: inherit
     with:
       version: main
       target: staging
       playwright_tags: "@extended"
   weekly-e2e-tests-local:
-    name: Weekly - all E2E tests (local)
+    name: Weekly E2E tests (local)
     uses: ./.github/workflows/ci-frontend-e2e.yml
+    if: always()
```
doug-s-nava, line 20: make sure each job always runs regardless of whether the other one fails

.github/workflows/e2e-staging.yml
```diff
@@ -30,7 +33,7 @@ on:
         description: "pipe separated list of @tags denoting groups of tests to run (e.g. @smoke|@core-regression). Leave blank to run all tests"
         type: string
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ inputs.target }}-${{ github.ref }}
```
doug-s-nava: make sure local and staging runs are in separate concurrency groups so they don't cancel each other

.github/workflows/e2e-create-report.yml
```diff
@@ -54,6 +58,6 @@ jobs:
       - name: Upload Merged HTML Report
         uses: actions/upload-artifact@v6
         with:
-          name: merged-html-report
+          name: merged-html-report-$${{ inputs.target }}
```
doug-s-nava: make sure local reports are not overwritten by staging or vice versa

.github/workflows/e2e-staging.yml
Bhavna-Ramachandran: 'inputs.target' will be empty on new push trigger here. Property 'inputs' is available for workflow_dispatch and workflow_call triggers - not for push. Ref: https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#inputs-context

This line would pass -- for concurrency group and create-report would be named **report-local for staging run triggered by push to main

group: ${{ github.workflow }}-${{ inputs.target || 'staging' }}-${{ github.ref }}

.github/workflows/e2e-staging.yml
```diff
@@ -68,6 +71,7 @@ jobs:
     secrets: inherit
     with:
       run_id: ${{ github.run_id }}
+      target: ${{ inputs.target }}
```
Bhavna-Ramachandran: Similar to the comment above, a fallback has to be added.

target: ${{ inputs.target || 'staging' }}

.github/workflows/e2e-create-report.yml
Bhavna-Ramachandran: I believe this is a typo and extra '$' needs to be removed.

name: merged-html-report-${{ inputs.target }}

image

.github/workflows/e2e-daily.yml
```diff
@@ -8,12 +8,16 @@ jobs:
...
+    secrets: inherit
     with:
       version: main
       target: staging
       playwright_tags: "@full-regression"
   daily-e2e-tests-local:
     name: Daily E2E tests (local)
     uses: ./.github/workflows/ci-frontend-e2e.yml
+    if: always()
+    secrets: inherit
```
Bhavna-Ramachandran, line 21: 👍

Issue #9260, opened 2026-03-25
Create reusable scripts for spinning up certificate CSR requests and consuming the responses


mdragon:
### Summary

To facilitate getting the grantee environments stood up, lets go ahead and request individual certificates for:
FE: grantee1.teams.simpler.grants.gov, grantee2.teams.simpler.grants.gov
API: api.grantee1.teams.simpler.grants.gov, api.grantee2.teams.simpler.grants.gov
Files: files.grantee1.teams.simpler.grants.gov, files.grantee2.teams.simpler.grants.gov

We shouldn't need NOFOs, Data (Metabase), SOAP, ???

To facilitate this work, we should create some scripts and documentation around:
- Create CSR
   - named to match the domain
   - append the simpler.grants.gov suffix automatically if it's not provided
   - etc
- Process certificate
   - Given a short domain name, append the simpler.grants.gov suffix if not provided
   - Add the certificate to AWS, converting formats if necessary

### Acceptance criteria

- [x] Scripts created and dogfooded on these certificates
- [x] Certificates requested
- [ ] Certificates received and added to AWS

prasnava:
Update: sent out the request for the new certs and the dns change request forms.

PR #9332, opened 2026-03-30
[Issue #9260]: Add csr helper scripts into the make file

prasnava:
## Summary

Created csr helper make commands for generating a csr request and imported a certificate into acm. 

## Validation steps

Verified the make commands are working as intended.  

Screenshot 2026-03-30 at 4 24 47 PM

### DIFF
--- a/infra/frontend/app-config/staging.tf
+++ b/infra/frontend/app-config/staging.tf
@@ -19,6 +19,8 @@ module "staging_config" {
   # With a minimum of 2, so CPU doesn't spike to infinity on deploys.
   instance_desired_instance_count = 2
   instance_scaling_min_capacity   = 2
+  instance_memory                 = 2048
+  instance_cpu                    = 1024
   # instance_scaling_max_capacity is 5x the instance_scaling_min_capacity
   instance_scaling_max_capacity = 10
 


bin/create-csr
```diff
@@ -0,0 +1,79 @@
...
+ORGANIZATION="Simpler Grants Gov"
+KEY_BITS=4096
+
+OUTPUT_DIR="certs"
+
+usage() {
+  echo "Usage: $0 <domain>"
+  echo
+  echo "  domain    Short name (e.g. 'api') or fully qualified (e.g. 'api.simpler.grants.gov')"
+  echo "            '.simpler.grants.gov' is appended automatically if not present."
```
mdragon, line 29: Just checking for the Prod FE, that would be the full domain, so can we call this script with simpler.grants.gov and get the correct request created? If not we should just flag that in the usage/comments (maybe with a pointer to the steps to use for that specific cert)

.gitignore
```diff
@@ -158,3 +158,6 @@ frontend/tests/artillery/params.json
 
 # XSD cache for XML validation
 xsd_cache/
+
+# ssl certificates directory
+certs/
```
mdragon: Should we make this work in like /tmp or something so we completely remove the ability to accidentally commit this stuff?

bin/create-csr
prasnava: Updated the csr script to support base domain for prod frontend.

.gitignore
prasnava: Updated to use the /tmp dir.

Issue #8599, opened 2026-02-19
"Edit Opportunity" Page


glenblosser-nava:
### Summary

Create an "Edit Opportunity" page for draft opportunities. This page will support editing the base opportunity fields and a single non-forecast summary only. Forecast summaries are out of scope for this ticket.

The page should allow a grantor to enter and update the core opportunity fields and the non-forecast summary fields needed for publishing workflow. The experience should support saving draft changes and returning later with values 
preserved.

[Figma mockup](https://www.figma.com/design/o5g0d2W5RKHYHjdvVCqTV2/Grantor-Designs-%7C-Quad-5-%7C-Simpler-Grants-2026?node-id=9750-33889&t=48cgcmwA0bU3xqPq-1)

### Acceptance criteria

- [ ] An "Edit Opportunity" page exists for draft opportunities
- [ ] The page supports editing the base opportunity fields
- [ ] The page supports editing one non-forecast summary
- [ ] Forecast summary fields and controls are not included
- [ ] A grantor can save changes without publishing
- [ ] Saved draft values are preserved when the grantor revisits the page

brianlee-nava:
@desaiamit CC: @glenblosser-nava @crystabelrangel based on discussions from pre-refinement on 3/5/2026. We decided to remove Forecasts summaries from the scope of this ticket.

@glenblosser-nava can you work with @desaiamit to make the necessary updates to this ticket and the tech spec (if appropriate)?

desaiamit:
### Mapping from Figma label -> frontend -> API -> DB

  | Figma label | Frontend field | Backend API field | Database field / column |
  |---|---|---|---|
  | Opportunity number | Opportunity number | opportunity_number | opportunity.opportunity_number |
  | Title | Opportunity title | opportunity_title | opportunity.opportunity_title |
  | Agency | Agency selector on create | agency_id | opportunity.agency_id |
  | Award selection method | Award selection method | category | opportunity.category |
  | Award selection method explanation | Award selection method explanation | category_explanation | opportunity.category_explanation |
  | Description | Opportunity summary description | summary_description | opportunity_summary.summary_description |
  | Funding type | Funding instruments (multi-select or selector) | funding_instruments | link_opportunity_summary_funding_instrument |
  | Cost sharing or matching requirements? | Cost sharing required | is_cost_sharing | opportunity_summary.is_cost_sharing |
  | Publish date | Posted date (non-forecast) | post_date | opportunity_summary.post_date |
[...]

Issue #7302, opened 2025-12-02
Ensure that opportunity page supports multiple uploads


brianlee-nava:
### Summary

We need support attachments and associate them with opportunities. Suggestion: Refer to Apply side's file upload component for error messages. Change messages as needed.

### Acceptance criteria

- [ ] Opportunity supports multiple uploads
- [ ] File type validation
- [ ] 2gb file upload limit
- [ ] Show error messages when upload fails

ssatia-SGM:
Based on discussion during refinement on 03/30, Chris W. and Doug has confirmed that File type Validation in Acceptance criteria is removed as this has not been implemented so far in Apply side component.

desaiamit:
Questions:

1. After upload, do we show list of attachments? (Yes)
2. Is user allowed to delete an attachment (Yes)
3. Is user allowed to download an attachment (Yes)
4. Who can delete an attachment once uploaded? (User with Edit priveleges)
5. Till what point in the state of an opportunity (_draft - yes, published - no, etc._) can a user with delete privileges be allowed to delete? (Yes, delete if user has permission)
6. Is sorting within the attachment list table required? (No)
7. Can we have the screenshots from "Apply" attached to this ticket, even if we don't create/update Figma file (Done by Crystabel)
8. Does a grantor API endpoints exist for attachments? — Do we create a new one to support this ticket or we make the existing /v1/grantors/opportunities/{id}/attachments endpoint more generic ? (Create new endpoint)
9. Does upload happen immediately on file selection (like Apply side), or only when the form Save button is clicked? (same as Apply)
10. Are there any allowed/blocked file types (doc,pdf,jpeg, etc.) (End point should provide the functionality to allow/block certain document types)

doug-s-nava:
figma: https://www.figma.com/design/o5g0d2W5RKHYHjdvVCqTV2/Grantor-Designs-%7C-Quad-5-%7C-Simpler-Grants-2026?node-id=9750-33889&m=dev

brianlee-nava:
UATed this ticket with @desaiamit. ACs pass, but we may need a follow up ticket to test file size and file type limits. Staging environment may have different configurations from local, Training, or Production.

PR #9334, opened 2026-03-30
[Issue #8599] Build edit opportunity page for grantor draft editing

desaiamit:
## Summary

Work for #8599

## Changes proposed

- **New edit page** at `/opportunity/[id]/edit` — fetches opportunity data, guards non-draft access (renders a warning and disables all fields if `is_draft` is false), and initialises form values from the existing forecast or non-forecast summary
- **Server action** (`actions.ts`) — validates form via Zod (field-level errors surfaced inline), POSTs a new opportunity summary on first save when none exists yet, PUTs on subsequent saves; maps 403/404/422/generic API errors to user-facing messages
- **`OpportunityEditForm` component** — full form with camelCase `name=` attributes, hidden inputs for `title` and `awardSelectionMethod` (needed for Zod validation and first-save POST), `useActionState` + `startTransition` (avoids React 19 uncontrolled form reset), and disabled field state for non-draft opportunities
- **`opportunityEditFormConfig.ts`** — `buildOpportunitySummaryUpdateRequest` helper that maps form data to the API payload; enum option lists for funding type, funding category, and eligibility
- **`opportunitySummaryGrantorFetcher.ts`** — create and update summary API fetcher functions
- **Create opportunity redirect** — after a successful create, the form now redirects to `/opportunity/[id]/edit` (Part 2 of the create flow)
- **i18n strings** — all `OpportunityEdit` labels, hints, alerts, and validation error messages in `en/index.ts`
- **Tests** — full coverage for the server action (validation, POST/PUT routing, error codes) and the form component (hidden fields, enum controls, non-draft warning)

## Context for reviewers

The opportunity data model has two separate records:
1. **Base opportunity** (`opportunity_id`, `opportunity_title`, `agency_id`, `category`) — created in step 1 by the Create Opportunity page
2. **Summary record** (`opportunity_summary_id`) — holds detailed fields (funding amounts, dates, eligibility, contact info); NOT created by the create page

On the first save from the edit page, `opportunitySummaryId` will be empty (no summary exists yet), so the action POSTs to create the summary. The response returns `newOpportunitySummaryId`, which is stored in component state so subsequent saves use PUT. This is required to satisfy the AC "A grantor can save changes without publishing" for newly created opportunities.

The `useActionState` + `startTransition` pattern (rather than `<form action={formAction}>`) was chosen to prevent React 19's automatic uncontrolled form reset on action completion, which was causing field values to be cleared after save.

## Validation steps

1. Run the seed command to create test agencies and users:
   ```bash
   make db-seed-local-with-agencies
   ```

2. Sign in as `one_agency_opp_edit` (Opportunity Editor for USAID-ETH)

3. Create a new opportunity via the Create Opportunity form — on success you should be redirected to `/opportunity/[id]/edit`

4. On the edit page:
   - Verify all fields are editable (the "draft only" warning should not appear)
   - Fill in the required fields and click **Save** — a green "Saved successfully" alert should appear
   - Reload the page — saved values should be restored from the API
   - Click **Save** again (second save should use PUT, not POST) — success alert should appear again

5. To test the non-draft guard: navigate directly to the edit page for a non-draft opportunity — all fields should be disabled and the warning banner should appear

6. Run tests locally:
   ```
   cd frontend && npm run all-checks
   ```


desaiamit:
Fixed the 2 issues, requesting re-review

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
```diff
@@ -0,0 +1,246 @@
...
+  return (
+    stageLabels[opportunityStatus] ??
+    opportunityStatus
+      .split("_")
+      .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
+      .join(" ")
+  );
+}
+
+export default async function OpportunityEditPage({
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: this should be feature flagged, shouldn't it?

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
```diff
@@ -0,0 +1,246 @@
...
+  searchParams?: Promise<Record<string, string>>;
+};
+
+export const dynamic = "force-dynamic";
+
+export function generateMetadata(): Metadata {
+  return {
+    title: "Edit opportunity",
+    description:
+      "Edit draft opportunity information and non-forecast summary fields.",
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: should pull this from the translation messages

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
```diff
@@ -0,0 +1,246 @@
...
+type PageProps = {
+  params: Promise<{ id: string; locale: string }>;
+  searchParams?: Promise<Record<string, string>>;
+};
+
+export const dynamic = "force-dynamic";
+
+export function generateMetadata(): Metadata {
+  return {
+    title: "Edit opportunity",
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: can we make this dynamic using the name of the opportunity? see how it's done on the opportunity detail

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
```diff
@@ -0,0 +1,246 @@
...
+    description:
+      "Edit draft opportunity information and non-forecast summary fields.",
+  };
+}
+
+export function generateStaticParams() {
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: this can be removed - we'll never want to generate this page statically

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
```diff
@@ -0,0 +1,246 @@
...
+function canEditOpportunity(opportunity: GrantorOpportunityDetail) {
+  return opportunity.is_draft === true;
+}
+
+function formatOpportunityStage(opportunityStatus: string | null | undefined) {
+  if (!opportunityStatus) {
+    return "";
+  }
+
+  const stageLabels: Record<string, string> = {
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: remove constant from function body

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
```diff
@@ -0,0 +1,246 @@
...
+  params,
+  searchParams,
+}: PageProps) {
+  const { id } = await params;
+  const resolvedSearchParams = searchParams ? await searchParams : {};
+  const isNewlyCreated = resolvedSearchParams.fromCreate === "true";
+
+  const session = await getSession();
+  if (!session || !session.token) {
+    redirect(`/opportunity/${id}`, RedirectType.push);
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: should show an unauthenticated message instead of redirecting

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
```diff
@@ -0,0 +1,246 @@
...
+      notFound();
+    }
+    if (status === 403) {
+      return <UnauthorizedMessage />;
+    }
+    throw error;
+  }
+
+  if (id !== opportunityData.opportunity_id) {
+    redirect(
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: this seems like a strange edge case, but if it happens I'd expect an error message rather than a redirect

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
```diff
@@ -0,0 +1,246 @@
...
+        <div className="usa-in-page-nav-container">
+          <ApplyFormNav title="On this page" fields={navigationItems} />
+
+          <section className="order-2 width-full maxw-tablet-xl padding-top-4">
+            <OpportunityEditForm
+              opportunityId={opportunityData.opportunity_id}
+              opportunitySummaryId={opportunitySummaryId}
+              isForecast={opportunityData.is_draft}
+              initialValues={initialValues}
+              isDraft={isEditableOpportunity}
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]:
```suggestion
              isDraft={!!opportunity.is_draft}
```

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
```diff
@@ -0,0 +1,246 @@
...
+    {
+      title: "Opportunities",
+      path: "/opportunities",
+    },
+    {
+      title: `${opportunityData.opportunity_title || ""}: ${opportunityData.opportunity_number}`,
+      path: `/opportunity/${opportunityData.opportunity_id}/`,
+    },
+  ];
+  const isEditableOpportunity = canEditOpportunity(opportunityData);
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: putting this in its own function seems to add confusion to the situation, as the name `canEditOpportunity` suggests this is authorization related. See my suggestion on like 237

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
```diff
@@ -0,0 +1,246 @@
...
+      primaryAssistanceListing?.assistance_listing_number || "",
+    opportunityNumber: opportunityData.opportunity_number || "",
+    opportunityStage: opportunityData.is_draft
+      ? "Draft"
+      : formatOpportunityStage(opportunityData.opportunity_status),
+    awardSelectionMethod: opportunityData.category || "",
+    awardSelectionMethodExplanation: opportunityData.category_explanation || "",
+  };
+  const rawLastUpdated =
+    opportunityData.summary?.post_date || opportunityData.updated_at || "";
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: this works, but curious where this lives in the requirements. I don't see anything in the ticket or figma about how this date should be calculated. Can you point to that requirement?

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
```diff
@@ -0,0 +1,246 @@
...
+  const lastUpdated = rawLastUpdated
+    ? new Date(rawLastUpdated).toLocaleDateString("en-US", {
+        month: "2-digit",
+        day: "2-digit",
+        year: "numeric",
+      })
+    : "";
+  const navigationItems = [
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx:150]: can you put this through translation?

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
```diff
@@ -0,0 +1,246 @@
...
+    { text: "Attachments", href: "attachments" },
+  ];
+  return (
+    <div className="bg-white">
+      <section className="bg-base-lightest border-bottom border-base-lightest padding-y-6">
+        <div className="grid-container">
+          <div className="padding-y-2">
+            <Breadcrumbs
+              breadcrumbList={breadcrumbs}
+              className="padding-y-0 bg-transparent font-sans-sm"
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: We should make sure our breadcrumbs are consistent across pages. Please match this to existing pages rather than the designs

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
```diff
@@ -0,0 +1,246 @@
...
+              <div className="display-flex flex-column maxw-mobile-lg">
+                <div className="font-sans-md line-height-sans-5 margin-bottom-2">
+                  <span className="text-bold">Last updated:</span> {lastUpdated}
+                </div>
+                <div className="display-flex flex-align-center">
+                  <span className="text-bold font-sans-md line-height-sans-5 margin-right-1">
+                    Status:
+                  </span>
+                  <span className="display-inline-flex flex-align-center bg-accent-warm text-ink padding-y-05 padding-x-1 radius-sm">
+                    <svg
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: use `<USWDSIcon>`

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
```diff
@@ -0,0 +1,246 @@
...
+
+      <div className="grid-container padding-bottom-4">
+        <div className="usa-in-page-nav-container">
+          <ApplyFormNav title="On this page" fields={navigationItems} />
+
+          <section className="order-2 width-full maxw-tablet-xl padding-top-4">
+            <OpportunityEditForm
+              opportunityId={opportunityData.opportunity_id}
+              opportunitySummaryId={opportunitySummaryId}
+              isForecast={opportunityData.is_draft}
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: seems like this will be the same value as `isDraft`. Will this change later?

frontend/src/components/opportunity/OpportunityEditForm.tsx
```diff
@@ -0,0 +1,1104 @@
...
+import { RequiredFieldIndicator } from "src/components/RequiredFieldIndicator";
+import {
+  ELIGIBILITY_OPTIONS,
+  FUNDING_CATEGORY_OPTIONS,
+  FUNDING_INSTRUMENT_OPTIONS,
+  OPPORTUNITY_CATEGORY_OPTIONS,
+  OpportunityEditFormValues,
+} from "./opportunityEditFormConfig";
+
+type Props = {
```
doug-s-nava:
```suggestion
type OpportunityEditFormProps = {
```

frontend/src/components/opportunity/opportunityEditFormConfig.ts
```diff
@@ -0,0 +1,305 @@
...
+  eligibleApplicants: string[];
+  additionalEligibilityInfo: string;
+  additionalInfoUrl: string;
+  additionalInfoUrlText: string;
+  grantorContactDetails: string;
+  contactEmail: string;
+  contactEmailText: string;
+};
+
+export type CheckboxOption = {
```
doug-s-nava: we have the type `{ label: string, value: string }` already defined twice in the app, don't need to define two more times. See https://github.com/HHS/simpler-grants-gov/blob/3df65dcf2c9c1c1e21b14cf5babe29b484016f69/frontend/src/types/search/searchSortTypes.ts#L14 and https://github.com/HHS/simpler-grants-gov/blob/3df65dcf2c9c1c1e21b14cf5babe29b484016f69/frontend/src/components/manageUsers/RoleManager.tsx#L16

Can you add this type to a generic types file as something like "SelectOptionDefinition"?

frontend/src/components/opportunity/opportunityEditFormConfig.ts
```diff
@@ -0,0 +1,305 @@
...
+  label: string;
+  value: string;
+};
+
+const emptyString = (value: string | null | undefined) => value ?? "";
+
+const numberToString = (value: number | null | undefined) =>
+  value === null || value === undefined ? "" : String(value);
+
+export const ELIGIBILITY_OPTIONS: CheckboxOption[] = [
```
doug-s-nava: we should make sure all of these options match up with the other parts of the app where they are used. See

* https://github.com/HHS/simpler-grants-gov/blob/2bf1185a92e6ddc494c285ada164c992f76a1dc1/frontend/src/components/opportunities/create/CreateOpportunityForm.tsx#L16
* https://github.com/HHS/simpler-grants-gov/blob/2293ece983d3b7b7ca31682536488b18078e1420/frontend/src/constants/opportunity.ts#L1
* https://github.com/HHS/simpler-grants-gov/blob/2293ece983d3b7b7ca31682536488b18078e1420/frontend/src/constants/searchFilterOptions.ts#L36
* https://github.com/HHS/simpler-grants-gov/blob/2293ece983d3b7b7ca31682536488b18078e1420/frontend/src/constants/searchFilterOptions.ts#L63

frontend/src/i18n/messages/en/index.ts
```diff
@@ -273,6 +273,172 @@ export const messages = {
...
+      contactEmailText: "Email display text",
+      yes: "Yes",
+      no: "No",
+      eligibilityBusiness: "Business",
+      eligibilityEducation: "Education",
+      eligibilityGovernment: "Government entities",
+      eligibilityNonprofit: "Nonprofit",
+      eligibilityMiscellaneous: "Miscellaneous",
+    },
+    eligibilityOptions: {
```
doug-s-nava: [note] I wouldn't worry too much about translating these - our option values are hardcoded in the rest of the app to make things more streamlined

frontend/src/components/opportunity/opportunityEditFormConfig.ts
```diff
@@ -0,0 +1,305 @@
...
+    .trim()
+    .replace(/,/g, "");
+  if (!normalizedValue) {
+    return null;
+  }
+  const parsedValue = Number(normalizedValue);
+  return Number.isNaN(parsedValue) ? null : parsedValue;
+}
+
+function getMultiValueField(
```
doug-s-nava, line 131: can we test this?

frontend/src/components/opportunity/opportunityEditFormConfig.ts
```diff
@@ -0,0 +1,305 @@
...
+  }
+
+  return formData
+    .getAll(fallbackFieldName)
+    .flatMap((value) =>
+      typeof value === "string" && value.trim() ? [value.trim()] : [],
+    );
+}
+
+export function buildOpportunitySummaryUpdateRequest(
```
doug-s-nava, line 153: this all looks pretty complex compared to how our other actions are reading values from forms. Does that mean that:
* our other form actions are not doing enough to standardize inputs
* this one is doing too much?

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/actions.ts
```diff
@@ -0,0 +1,261 @@
...
+    "OpportunityEdit.validationErrors",
+  );
+  const alerts = await getTranslations("OpportunityEdit.content.alerts");
+
+  return { validationErrors, alerts };
+}
+
+function validateOpportunityEditForm(
+  formData: FormData,
+  validationErrors: Awaited<
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunity/[id]/edit/actions.ts]: this typing is confusing. What do we gain by passing this in rather than just calling to getTranslations within the function?

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/actions.ts
```diff
@@ -0,0 +1,261 @@
...
+
+      if (closeDate < publishDate) {
+        ctx.addIssue({
+          code: z.ZodIssueCode.custom,
+          path: ["closeDate"],
+          message: validationErrors("closeDateOrder"),
+        });
+      }
+    })
+    .superRefine(({ awardMinimum, awardMaximum }, ctx) => {
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunity/[id]/edit/actions.ts:106]: Going forward we should confirm with product how much frontend validation is necessary on these forms. This sort of detail is definitely nice to have, but will also be checked on the backend. We may end up opting for a middle ground in terms of frontend validation for required fields or data types, and leave more complex logic like this to the backend.

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/actions.ts
```diff
@@ -0,0 +1,261 @@
...
+  }
+
+  const opportunityId = readStringValue(formData.get("opportunityId")).trim();
+  const opportunitySummaryId = readStringValue(
+    formData.get("opportunitySummaryId"),
+  ).trim();
+  const isForecast =
+    readStringValue(formData.get("isForecast")).trim() === "true";
+
+  if (!opportunityId) {
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunity/[id]/edit/actions.ts:170]: seems like we should do this before performing full validation

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/actions.ts
```diff
@@ -0,0 +1,261 @@
...
+    return {
+      errorMessage: alerts("missingSummaryContext"),
+    };
+  }
+
+  const session = await getSession();
+  if (!session?.token) {
+    return { errorMessage: alerts("forbidden") };
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunity/[id]/edit/actions.ts]: if there is no token, this should be an unauthenticated message rather than "forbidden"

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/actions.ts
```diff
@@ -0,0 +1,261 @@
...
+    readStringValue(formData.get("isForecast")).trim() === "true";
+
+  if (!opportunityId) {
+    return {
+      errorMessage: alerts("missingSummaryContext"),
+    };
+  }
+
+  const session = await getSession();
+  if (!session?.token) {
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunity/[id]/edit/actions.ts:177]: this check should happen before form validation

frontend/src/components/opportunity/OpportunityEditForm.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+    expect(
+      screen.getByRole("combobox", { name: /labels\.fundingType/i }),
+    ).toBeInTheDocument();
+    expect(
+      screen.getByRole("combobox", { name: /labels\.category/i }),
+    ).toBeInTheDocument();
+    expect(screen.getByRole("radio", { name: /yes/i })).toBeInTheDocument();
+    // Verify checkbox value attribute so the correct enum string is submitted with the form
+    // (checkboxes without a value= prop default to "on" and fail API validation)
+    // eslint-disable-next-line jest-dom/prefer-to-have-value
```
doug-s-nava, line 93: why doesn't `.toHaveValue` work here?

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
```diff
@@ -0,0 +1,246 @@
...
+                      <use href="/uswds/img/sprite.svg#schedule" />
+                    </svg>
+                    {opportunityKeyInformation.opportunityStage}
+                  </span>
+                </div>
+              </div>
+              <div className="display-flex flex-wrap flex-align-center">
+                <Button
+                  type="submit"
+                  form="opportunity-edit-form"
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx:192]: ❤️

frontend/src/components/opportunity/OpportunityEditForm.tsx
```diff
@@ -0,0 +1,1104 @@
...
+    useState<OpportunityEditFormValues>(initialValues);
+  const [currentSummaryId, setCurrentSummaryId] =
+    useState(opportunitySummaryId);
+  const [formState, formAction] = useActionState(saveOpportunityEditAction, {
+    validationErrors: {},
+  });
+  const validationErrors: OpportunityEditValidationErrors | undefined =
+    formState.validationErrors;
+
+  function updateField<K extends keyof OpportunityEditFormValues>(
```
doug-s-nava, line 119: why do we need to hold form values in state?

frontend/src/components/opportunity/OpportunityEditForm.tsx
```diff
@@ -0,0 +1,1104 @@
...
+      label: t("labels.awardSelectionMethodExplanation"),
+      value:
+        opportunityKeyInformation.awardSelectionMethodExplanation ||
+        t("content.notAvailable"),
+    },
+  ];
+
+  const leftColumnItems = keyInformationItems.slice(0, 3);
+  const rightColumnItems = keyInformationItems.slice(3);
+  const eligibilityGroups = {
```
doug-s-nava, line 200: grouping is handled already in the opportunity detail definitions, can we reuse that instead?

frontend/src/components/opportunity/OpportunityEditForm.tsx
```diff
@@ -0,0 +1,1104 @@
...
+        </div>
+      </Fieldset>
+    );
+  }
+
+  return (
+    <form
+      id="opportunity-edit-form"
+      onSubmit={(e) => {
+        e.preventDefault();
```
doug-s-nava, line 239: what is this needed for?

frontend/src/components/opportunity/OpportunityEditForm.tsx
```diff
@@ -0,0 +1,1104 @@
...
+        const formData = new FormData(e.currentTarget);
+        startTransition(() => {
```
doug-s-nava, line 241: any reason not to use `action={formAction}` instead?

frontend/src/components/opportunity/OpportunityEditForm.tsx
```diff
@@ -0,0 +1,1104 @@
...
+        </h2>
+        <p className="margin-top-0 margin-bottom-4 font-sans-lg text-base-dark">
+          {t("content.fundingDetailsIntro")}
+        </p>
+
+        <div className="display-flex flex-column gap-3">
+          <div className="grid-row grid-gap-lg">
+            <div className="tablet:grid-col-6">
+              <FormGroup error={!!getFieldError("fundingType")}>
+                <Label
```
doug-s-nava: please use `<DynamicFieldLabel>`

frontend/src/components/opportunity/OpportunityEditForm.tsx
```diff
@@ -0,0 +1,1104 @@
...
+                    <RequiredFieldIndicator> *</RequiredFieldIndicator>
+                  </span>
+                </Label>
+                <p className="margin-top-1 margin-bottom-2 font-sans-sm text-base">
+                  {t("content.fundingTypeHint")}
+                </p>
+                {getFieldError("fundingType") ? (
+                  <ErrorMessage>{getFieldError("fundingType")}</ErrorMessage>
+                ) : null}
+                <Select
```
doug-s-nava, line 395: is there a reason not to use the "common" field components that Dao created?

frontend/src/services/fetch/fetchers/opportunitySummaryGrantorFetcher.ts
```diff
@@ -0,0 +1,78 @@
...
+  token: string;
+};
+
+type CreateOpportunitySummaryForGrantorParams = {
+  opportunityId: string;
+  body: OpportunitySummaryCreateRequest;
+  token: string;
+};
+
+const getGrantorOpportunityRequest = requesterForEndpoint(
```
doug-s-nava: these should be defined in fetchers.ts

frontend/src/types/opportunity/opportunityResponseTypes.ts
```diff
@@ -52,6 +52,37 @@ export interface Summary extends MinimalSummary {
...
+export interface OpportunitySummaryDetail extends Summary {
+  opportunity_summary_id: string;
+}
+
+export type OpportunitySummaryCreateRequest =
+  OpportunitySummaryUpdateRequest & {
+    is_forecast: boolean;
+  };
+
+export type OpportunitySummaryUpdateRequest = {
```
doug-s-nava, line 60: if this is going to be extended I'd prefer we use an interface

doug-s-nava (commented):
* looks like there is an attachments item in the left nav but no actual attachments section
* header looks scrunched on mobile

Image

doug-s-nava:
Also, a minor nit but the font for the key info description looks lighter than the others
Screenshot 2026-04-07 at 1 52 30 PM

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
Comment (doug-s-nava) [reply frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: not a blocker, can be follow up

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
Comment (doug-s-nava) [reply frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: not a blocker, but confirm with product

frontend/src/components/opportunity/opportunityEditFormConfig.ts
doug-s-nava: this can be a follow up

frontend/src/components/opportunity/opportunityEditFormConfig.ts
doug-s-nava: for follow up

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/actions.ts
Comment (doug-s-nava) [reply frontend/src/app/[locale]/(base)/opportunity/[id]/edit/actions.ts:106]: follow up with product

frontend/src/components/opportunity/OpportunityEditForm.tsx
doug-s-nava: can be a follow up

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
Comment (desaiamit) [reply frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: Reused "opportunitiesListOff" flag for edit page, remove unused opportunityOff flag

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
Comment (desaiamit) [reply frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: Fixed.

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
Comment (desaiamit) [reply frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: Fixed. The title will now render as "Edit opportunity - {opportunity_title}", falling back to just "Edit opportunity" if the fetch fails

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
Comment (desaiamit) [reply frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: Removed

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
Comment (desaiamit) [reply frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: Fixed, moved the stageLabels from inside formatOpportunityStage to above it at module level

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
Comment (desaiamit) [reply frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: Fixed

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
Comment (desaiamit) [reply frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: Fixed. shows an <Alert type="error" instead of redirect

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
Comment (desaiamit) [reply frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: Fixed

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
Comment (desaiamit) [reply frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: Fixed as part of the above comment, removed isEditableOpportunity = canEditOpportunity(opportunityData);

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
Comment (desaiamit) [reply frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx:150]: Fixed, passed locale as param instead of hard-coding "en-US"

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
Comment (desaiamit) [reply frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: Fixed the font / size and background color

Breadcrumb will now be:
  - From create: Home > Opportunities > Create opportunity > Opportunity details
  - Direct access: Home > Opportunities > Opportunity details

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
Comment (desaiamit) [reply frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: Fixed

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
Comment (desaiamit) [reply frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: Fixed, isForecast now correctly reflects whether a forecast summary exists using opportunityData.forecast_summary instead of relying on opportunityData.is_draft

frontend/src/components/opportunity/OpportunityEditForm.tsx
desaiamit: Fixed

frontend/src/components/opportunity/opportunityEditFormConfig.ts
desaiamit: Fixed. By importing from eligibilityTypes, fundingOptions, and categoryOptions (the same sources the rest of the app uses), the options now stay in sync automatically.

frontend/src/i18n/messages/en/index.ts
desaiamit: Fixed. Removed the eligibilityOptions block from en/index.ts and stopped using t("eligibilityOptions.*") in OpportunityEditForm.tsx. Labels now come directly from the shared constants.

frontend/src/types/opportunity/opportunityResponseTypes.ts
desaiamit: Converted to interface and updated OpportunitySummaryCreateRequest to use extends instead of the type intersection.

frontend/src/services/fetch/fetchers/opportunitySummaryGrantorFetcher.ts
desaiamit: Fixed - moved getGrantorOpportunityRequest, updateGrantorOpportunitySummaryRequest, and createGrantorOpportunitySummaryRequest to fetchers.ts as named exports, and updated opportunitySummaryGrantorFetcher.ts to import them from there.

frontend/src/components/opportunity/OpportunityEditForm.tsx
desaiamit: action={formAction} would trigger React 19's automatic form reset after submission, clearing the uncontrolled DatePicker fields. The onSubmit + startTransition pattern prevents the reset while still calling the server action correctly. Let me know if I am missing something.

frontend/src/components/opportunity/OpportunityEditForm.tsx
desaiamit: The USWDS checkbox/radio components need controlled checked props to pre-populate from server data, and several fields are conditionally rendered based on current selections — both require tracking state. Let me know if I am missing something.

frontend/src/components/opportunity/OpportunityEditForm.test.tsx
desaiamit: Added extra comments.

For an unchecked checkbox, .toHaveValue() always returns null regardless of the value attribute — it reflects the current form value (null when unchecked), not the HTML attribute. .toHaveAttribute("value", ...) reads the attribute directly, which is what we need to verify

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/actions.ts
Comment (desaiamit) [reply frontend/src/app/[locale]/(base)/opportunity/[id]/edit/actions.ts]: Fixed, by making validateOpportunityEditForm async and call getTranslations inside it directly

frontend/src/components/opportunity/opportunityEditFormConfig.ts
desaiamit: - Attachment link will be fixed with #7302 
- Fixed for scrunching on mobile
- Added tests

desaiamit:
> Also, a minor nit but the font for the key info description looks lighter than the others Screenshot 2026-04-07 at 1 52 30 PM

The design has changed https://www.figma.com/design/o5g0d2W5RKHYHjdvVCqTV2/Grantor-Designs-%7C-Quad-5-%7C-Simpler-Grants-2026?node-id=9750-33888&p=f&t=3Ior9Iy5XbpdIdRf-0

**I think a process is required. Once a figma file is moved to "Ready for Dev" it cannot be changed without a new ticket**

frontend/src/components/opportunity/OpportunityEditForm.tsx
desaiamit: Checked the CreateOpportunityForm, the CreateOpportunityForm can use action={formAction} directly because it redirects on success — the native form reset that triggers on submission doesn't matter there. OpportunityEditForm stays on the same page after save, so a native form reset would wipe the user's inputs.
Let me know if I am missing something.

frontend/src/types/opportunity/opportunityResponseTypes.ts
desaiamit: Reverted back to type. Using interface broke the TypeScript build because these types are passed as the request body to the fetcher, which expects JSONRequestBody — a generic { [key: string]: unknown } type. TypeScript is stricter with interface declarations: it won't allow an interface to be used where an index signature is expected unless the interface explicitly declares one too. Since type aliases don't have this restriction, keeping them as type is the right call here. Happy to revisit if there's a pattern elsewhere in the codebase that handles this differently.

frontend/src/components/opportunity/OpportunityEditForm.tsx
desaiamit: These components were developed in parallel with the Create Opportunity work, so the shared field components weren't available when this ticket was started. We've since adopted DynamicFieldLabel (addressed in this PR). Happy to refactor the remaining field wrappers to use Dao's components in a follow-up ticket if that's preferred.

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/actions.ts
```diff
@@ -0,0 +1,253 @@
...
+  successMessage?: string;
+  validationErrors?: OpportunityEditValidationErrors;
+  newOpportunitySummaryId?: string;
+};
+
+function readStringValue(value: FormDataEntryValue | null): string {
+  return typeof value === "string" ? value : "";
+}
+
+async function getOpportunityEditTranslations() {
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunity/[id]/edit/actions.ts]: this doesn't seem necessary - a call directly to getTranslations should do the same thing

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
```diff
@@ -0,0 +1,263 @@
...
+};
+
+function formatOpportunityStage(opportunityStatus: string | null | undefined) {
+  if (!opportunityStatus) {
+    return "";
+  }
+
+  return (
+    stageLabels[opportunityStatus] ??
+    opportunityStatus
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: what are the use cases where an opportunity would not have a valid status, and what are we assuming the status looks like in this fallback case? This would be a nice function to test to demonstrate the expected data and behavior

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
```diff
@@ -0,0 +1,263 @@
...
+    {
+      title: "Home",
+      path: "/",
+    },
+    {
+      title: "Opportunities",
+      path: "/opportunities",
+    },
+    ...(isNewlyCreated
+      ? [{ title: "Create opportunity", path: "/opportunities/create" }]
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: does this make sense? In what case would the user want to navigate back to the create page from here?

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
```diff
@@ -0,0 +1,263 @@
...
+
+  if (!hasVerifiedGrantorEditAccess) {
+    return <UnauthorizedMessage />;
+  }
+
+  const primaryAssistanceListing =
+    opportunityData.opportunity_assistance_listings[0];
+  const breadcrumbs: Breadcrumb[] = [
+    {
+      title: "Home",
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: we're no longer including "home" in our breadcrumbs. Not on you to know that since I don't think the new documentation has been published yet...

frontend/src/components/opportunity/OpportunityEditForm.test.tsx
doug-s-nava: interesting! thanks for looking into it

frontend/src/components/opportunity/OpportunityEditForm.tsx
```diff
@@ -0,0 +1,967 @@
...
+      const currentFieldValues = currentValues[key];
+      const nextFieldValues = currentFieldValues.includes(value)
+        ? currentFieldValues.filter((currentValue) => currentValue !== value)
+        : [...currentFieldValues, value];
+
+      return { ...currentValues, [key]: nextFieldValues };
+    });
+  }
+
+  function formatNumber(value: string): string {
```
doug-s-nava: this seems like a general utility that can live outside of the render

frontend/src/components/opportunity/OpportunityEditForm.tsx
```diff
@@ -0,0 +1,967 @@
...
+    ),
+    misc: ELIGIBILITY_OPTIONS.filter((option) =>
+      ["individuals", "other", "unrestricted"].includes(option.value),
+    ),
+  };
+  const eligibilityDisplayLabels: Record<string, string> = Object.fromEntries(
+    ELIGIBILITY_OPTIONS.map(({ value, label }) => [value, label]),
+  );
+
+  function renderEligibilityCheckboxGroup(
```
doug-s-nava: would be more performant to make this its own component and pass in any missing values. That lets React do its job better in terms of optimizing the render

frontend/src/components/opportunity/OpportunityEditForm.tsx
doug-s-nava: sure, that can be a follow up

frontend/src/components/opportunity/opportunityEditFormConfig.ts
```diff
@@ -0,0 +1,210 @@
...
+  { label: "Continuation", value: "continuation" },
+  { label: "Earmark", value: "earmark" },
+  { label: "Other", value: "other" },
+];
+
+export const FUNDING_INSTRUMENT_OPTIONS: CheckboxOption[] = fundingOptions.map(
+  ({ label, value }) => ({ label: label.trim(), value }),
+);
+
+export const FUNDING_CATEGORY_OPTIONS: CheckboxOption[] = categoryOptions.map(
```
doug-s-nava: this looks like an identity function. Is this doing anything?

frontend/src/components/opportunity/opportunityEditFormConfig.ts
doug-s-nava: I meant to suggest unit tests for this function. Can you do that?

frontend/src/components/opportunity/opportunityEditFormConfig.ts
```diff
@@ -0,0 +1,210 @@
...
+
+// Private helpers for FormData → API payload mapping
+
+function readStringFromFormData(value: FormDataEntryValue | null): string {
+  return typeof value === "string" ? value : "";
+}
+
+function emptyToNull(value: FormDataEntryValue | null): string | null {
+  const normalizedValue = readStringFromFormData(value).trim();
+  return normalizedValue || null;
```
doug-s-nava, line 115: why do we need to do this? I think the API can handle converting undefined values to null.

frontend/src/services/fetch/fetchers/opportunitySummaryGrantorFetcher.ts
```diff
@@ -0,0 +1,65 @@
...
+  token: string;
+};
+
+export async function getOpportunityForGrantor(
+  opportunityId: string,
+  token: string,
+): Promise<GrantorOpportunityApiResponse> {
+  const response = await getGrantorOpportunityRequest({
+    subPath: opportunityId,
+    additionalHeaders: { "X-SGG-Token": token },
```
doug-s-nava, line 33: you can take this out and use the endpoint config based auth as I described last week

frontend/src/types/opportunity/opportunityResponseTypes.ts
doug-s-nava: Got it. This is interesting! https://github.com/microsoft/TypeScript/issues/15300

I definitely would have expected the interface to satisfy that type! Fwiw our general pattern is to not strictly type the request bodies. We probably should allow a generic type on the fetcher functions so that we can be more typesafe there. For now this is fine

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/actions.ts
```diff
@@ -0,0 +1,253 @@
...
+      };
+    }
+
+    if (status === 404) {
+      return {
+        errorMessage: alerts("notFound"),
+      };
+    }
+
+    if (status === 422) {
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunity/[id]/edit/actions.ts:233]: 422s are thrown for any failed validation. This is ok but we need a follow up ticket to properly handle back end validation issues. This will mean handling 422s separately from other errors rather than throwing on them so that we can preserve the errors and process them to display on the frontend

frontend/src/components/opportunity/OpportunityEditForm.tsx
doug-s-nava: got it - we can work around this on certain fields by setting a key. It's a somewhat hacky workaround for https://github.com/facebook/react/issues/30580 but so is the `onSubmit`

frontend/src/components/opportunity/OpportunityEditForm.tsx
```diff
@@ -0,0 +1,967 @@
...
+
+  function updateField<K extends keyof OpportunityEditFormValues>(
+    key: K,
+    value: OpportunityEditFormValues[K],
+  ) {
+    setValues((currentValues) => ({ ...currentValues, [key]: value }));
+  }
+
+  function toggleArrayValue(
+    key: "fundingType" | "fundingCategories" | "eligibleApplicants",
```
doug-s-nava: are funding type and funding category actually something where we're supporting selecting multiple values? Or is this just how the data is modeled to support legacy?

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
Comment (doug-s-nava) [reply frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: throw an error on an unexpected status instead

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
Comment (doug-s-nava) [reply frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: my take is to remove this, but follow up with @andycochran

frontend/src/components/opportunity/opportunityEditFormConfig.ts
doug-s-nava: this can be a follow up

frontend/src/components/opportunity/OpportunityEditForm.tsx
doug-s-nava: follow up with product about how we want to model this going forward - do we want to support users supplying multiple types or categories? If so we should allow that in the UI, if not we should handle the complexity of modeling as arrays on the API side

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
Comment (andycochran) [reply frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: I'd suggest for now we just leave breadcrumbs off of this page altogether. I'm not sure they're not needed for this create/publish flow. 

As @doug-s-nava mentioned in the other comment, not on you since this guidance isn't published in a good place yet: 

We don't rely on breadcrumbs for navigate back to previous pages, and explicitly don't use them to indicate browser history. Breadcrumbs are intended for pages you're dropped directly into, so you know where you are in the site hierarchy. They're not meant for navigating linear page flows. 

Let's revisit the idea of breadcrumbs here once we have a better idea of the header navigation and information architecture. 

cc @crystabelrangel, are you okay with leaving Breadcrumbs off this design for now and we can address them in a follow-up ticket?

Comment (crystabelrangel) [reply frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: Thanks for the heads up, leaving breadcumbs off for now works for me!

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/actions.ts
Comment (desaiamit) [reply frontend/src/app/[locale]/(base)/opportunity/[id]/edit/actions.ts]: Fixed - removed the wrapper and calling getTranslations("OpportunityEdit.content.alerts") directly in saveOpportunityEditAction

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
Comment (desaiamit) [reply frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: Fixed - the fallback to raw status string was masking unexpected API values. Now throws an error for any status that isn't one of the four known values (archived, closed, forecasted, posted) or is_draft. The four known statuses are the complete set defined by the API enum, so anything outside them indicates bad data that should surface as an error rather than render silently.

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
Comment (desaiamit) [reply frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: Removed breadcrumbs entirely

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
Comment (desaiamit) [reply frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: Resolved - removed breadcrumbs from this page entirely per Andy's guidance in the thread above.

frontend/src/components/opportunity/OpportunityEditForm.tsx
desaiamit: Fixed - moved formatNumber to module level outside the component

frontend/src/components/opportunity/OpportunityEditForm.tsx
desaiamit: Fixed - extracted EligibilityCheckboxGroup as a proper module-level React component with explicit props (title, options, baseId, selectedValues, onToggle, disabled). Also moved eligibilityDisplayLabels to module level alongside it since it's derived from a constant

frontend/src/components/opportunity/opportunityEditFormConfig.ts
desaiamit: Tested the API locally: sending post_date: "" returns 422 "Not a valid date." since FormData always produces "" for blank inputs, never null. The conversion to null is required before the API call regardless of whether validation is done on the frontend or deferred to the backend. Ok to keeping emptyToNull ?

frontend/src/services/fetch/fetchers/opportunitySummaryGrantorFetcher.ts
desaiamit: Sure, will park this for now and apply it as soon as #9431 is merged to main

frontend/src/components/opportunity/OpportunityEditForm.tsx
desaiamit: Simplified - fundingType and fundingCategories are now string (not string[]) since the UI is single-select. The state, initial values, and form bindings all use plain strings. The API still receives a single-element array at the call boundary since that's what the schema expects. If product decides multi-select is needed in future, we can create a new ticket, but now will go with what UI shows

frontend/src/components/opportunity/opportunityEditFormConfig.ts
desaiamit: Fixed - ELIGIBILITY_OPTIONS and FUNDING_CATEGORY_OPTIONS now assign directly from eligibilityTypes and categoryOptions without the redundant .map(). FUNDING_INSTRUMENT_OPTIONS keeps its map since it trims whitespace from labels

frontend/src/components/opportunity/OpportunityEditForm.tsx
desaiamit: Fixed - DynamicFieldLabel is now used for all field labels throughout the form

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/actions.ts
Comment (desaiamit) [reply frontend/src/app/[locale]/(base)/opportunity/[id]/edit/actions.ts:177]: Fixed - same as above, session check is now before form validation

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/actions.ts
Comment (desaiamit) [reply frontend/src/app/[locale]/(base)/opportunity/[id]/edit/actions.ts]: Fixed - when !session?.token, the action now returns alerts("unauthenticated")

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/actions.ts
Comment (desaiamit) [reply frontend/src/app/[locale]/(base)/opportunity/[id]/edit/actions.ts:170]: Fixed - session check now happens before validateOpportunityEditForm is called

frontend/src/components/opportunity/opportunityEditFormConfig.ts
desaiamit: The complexity is driven by what the API requires that FormData doesn't handle natively: empty strings must become null (confirmed by local API test — "" returns 422 for date fields), number strings need parsing to number | null, and checkbox groups submit multiple values via getAll. After simplifying fundingType and fundingCategories to single-select strings, the remaining complexity is the minimum needed to bridge FormData to the API schema

frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx
Comment (desaiamit) [reply frontend/src/app/[locale]/(base)/opportunity/[id]/edit/page.tsx]: As discussed with the Product Team during after party, we will use the last updated from either 'Opportunity' or 'Opportunity Summary' which every is the latest

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
```diff
@@ -114,7 +114,7 @@ class OpportunityGrantorSchema(OpportunityV1Schema):
     )
 
     forecast_summary = fields.Nested(
-        OpportunitySummaryV1Schema(),
+        lambda: OpportunitySummaryDetailSchema(),
```
doug-s-nava, line 117: not familiar enough with flask to know what this is doing. Why do we need to make this change?

frontend/src/services/fetch/fetchers/opportunitySummaryGrantorFetcher.ts
doug-s-nava: yeah, sorry I had forgotten it hadn't merged to main yet. You can leave this for now and I can fix it up later

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
desaiamit: OpportunityGrantorSchema (line 107) references OpportunitySummaryDetailSchema, which is defined later in the same file (line 545). Python executes files top-to-bottom, so writing OpportunitySummaryDetailSchema() directly would fail at import time — the class doesn't exist yet at that point.

Wrapping it in lambda: OpportunitySummaryDetailSchema() defers the instantiation: Python stores a small anonymous function and only calls it when the schema is actually used at runtime, by which point the class is fully defined. Marshmallow's fields.Nested() supports this pattern exactly for this forward-reference scenario.

The alternative would be to reorder the classes in the file, but OpportunitySummaryDetailSchema extends OpportunitySummaryV1Schema which has its own dependencies, so reordering would require moving a lot more code around. Hence, went with using lambda here

desaiamit:
> Everything seems to work except that there are no failed validations for many of the fields marked as required. I'm able to save with no warnings even with these fields missing:
> 
> * expected number of awards
> * estimated total program funding
> * award minimum
> * award maximum
> * description
> * link to additional information
> * link display text
> * grantor contact details
> * contact email
> * email display text
> 
> My guess is that none of these fields are not actually required - if they were actually required the API would not have allowed the record to save properly and would have thrown a 422. For now I'd suggest removing the asterisks from these fields, and follow up to see if any of these fields should actually be required, and we can make a follow up if so.

- As discussed in after party, will remove asterisk (mandatory fields) for now. We can look into how to handle this scenario with the "Publish Opportunity" ticket

- For the issue with "X-SGG-Token", I can fix that along with other follow-up items. Feel free to add it to the follow-up ticket

doug-s-nava (approved):

PR #9337, opened 2026-03-31
[Issue #6876] automatically include x-sgg-token header for authorized fetch calls

doug-s-nava:
## Summary

Work for #6876 

## Changes proposed

This change:

* introduces functionality to automatically add the necessary X-SGG-Token header to all API requests that require authentication
* applies that functionality to application related API requests

## Context for reviewers

Note that this change does not update all authenticated fetch calls, will complete this work in a follow up PR. All existing calls should still work using the old manual system until they are translated over to the new system.

As part of this work I produced an audit of all authenticated calls, which can be found here https://docs.google.com/spreadsheets/d/1T23xaq6AMl7qi6OErAzGFRHXs2TjGe-vVI5S9ImdtSs/edit?usp=sharing

A follow up ticket has been created for cleaning up some unrelated standardization issues discovered when performing the audit https://github.com/HHS/simpler-grants-gov/issues/9373

Fetcher helpers file cleaned up a bit so that files that just want to import the types it exported don't opt in to session related stuff.

## Validation steps

2. run through all application regression test scenarios below
3. _VERIFY_: everything works as expected, particularly in relation to authentication - you are allowed to take the actions you expect, and are denied as expected

### Application regression test scenarios

2. visit /applications
3. verify you get a suitable error message
1. log in as `many_app_user`
4. click on the link for the `Test Opportunity for BudgetNarrativeAttachments_1_2 1.2` opportunity
5. open a new application as an individual
6. transfer application ownership to an organization
7. select `Yes` to include the Budget Narrative Attachment Form in the submission
8. complete the Budget Narrative Attachment Form and save
9. return to the application page and submit
10. check the audit history to make sure it is correctly populated

## Todos

- [x] audit app for all calls using x-sgg-token header
- [x] determine whether there are any unauthed calls from namespaced endpoint configs that generally need auth
- [x] split out any functions found in the above audit into their own endpoint configs
- [x] build auth into the main fetcher function
- [x] write unit tests
- [x] add auth to endpoint configs for relevant endpoints
- [x] remove direct inclusions of auth header

### DIFF
--- a/frontend/src/services/fetch/fetcherHelpers.ts
+++ b/frontend/src/services/fetch/fetcherHelpers.ts
@@ -24,11 +24,9 @@ import { QueryParamData } from "src/types/search/searchRequestTypes";
 export async function getDefaultHeaders({
   addContentType = true,
   requiresUserAuthToken = false,
-  // userAuthToken,
 }: {
   addContentType?: boolean;
   requiresUserAuthToken?: boolean;
-  // userAuthToken?: string;
 }): Promise<Record<string, string>> {
   const headers: Record<string, string> = {};
 
@@ -46,9 +44,7 @@ export async function getDefaultHeaders({
   if (requiresUserAuthToken) {
     const session = await getSession();
     if (!session?.token) {
-      // Should make sure this works ok - expect API to send back a 403 if the header is missing
-      // we could throw an error, but maybe best to make the API request and let it fail there?
-      // maybe bring this to the group.
+      // May want to throw here
       console.warn("no user token present for authorized endpoint call");
     } else {
       headers["X-SGG-Token"] = session.token;
--- a/frontend/src/services/fetch/fetchers/fetchers.ts
+++ b/frontend/src/services/fetch/fetchers/fetchers.ts
@@ -46,7 +46,6 @@ export function requesterForEndpoint({
       nextOptions?: NextFetchRequestConfig;
       allowedErrorStatuses?: number[];
       addContentType?: boolean;
-      // userAuthToken?: string;
     } = {},
   ): Promise<Response> {
     const {
@@ -56,7 +55,6 @@ export function requesterForEndpoint({
       nextOptions,
       allowedErrorStatuses = [],
       addContentType = true,
-      // userAuthToken,
     } = options;
     const url = createRequestUrl(
       method,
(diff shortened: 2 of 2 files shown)


frontend/src/services/fetch/fetchers/fetchers.ts
```diff
@@ -154,6 +160,5 @@ export const fetchOrganizationBySavedOpportunities = (
 
 export const fetchLocalUsers = requesterForEndpoint(getLocalUsersEndpoint);
 
-export const fetchGrantorWithMethod = (
-  type: "POST" | "DELETE" | "PUT" | "GET",
-) => requesterForEndpoint(toDynamicGrantorsEndpoint(type));
+export const fetchGrantorWithMethod = (type: "POST") =>
```
doug-s-nava, line 160: the rest of these were not being used, we should move towards not defining endpoint configs until we'll be using them. Easier to keep track that way

frontend/src/services/fetch/fetchers/applicationFetcher.ts
```diff
@@ -209,19 +170,14 @@ export const updateApplicationFilingName = async (
  * Application Forms
  */
 
-export const getApplicationFormDetails = async (
-  token: string,
+export const getApplicationFormDetailsForPrint = async (
```
doug-s-nava, line 173: where there was previously a logic branch in `getApplicationFormDetails` to handle either an SGG token or an internal token, splitting this into separate functions in order to opt the general application endpoint config into auth, while letting the print related endpoint function on its own with its own manual authentication

frontend/src/services/fetch/fetcherHelpers.ts
```diff
@@ -40,6 +40,17 @@ export function getDefaultHeaders(addContentType = true): HeadersDict {
   if (addContentType) {
     headers["Content-Type"] = "application/json";
   }
+
+  if (requiresUserAuthToken) {
+    const session = await getSession();
+    if (!session?.token) {
+      // May want to throw here
```
glenblosser-nava, line 47: Did you decide if we need to throw here? Is it okay to return the headers without the session.token?

doug-s-nava: Wanted to bring this up to the group and haven't had a chance yet, so no decision but I'm leaning towards throwing here. This would let us standardize more on behavior for a case where we know a user isn't logged in. As it stands, some fetch functions check the token and error early if it's not there, and others don't.

As for how this works without throwing, in this case the request would be missing the header, so the API will send back a 403. This seems to work fine, but would be best to save the round trip if we already know it will fail.

Issue #9232, opened 2026-03-24
Handling agency behavior on create opportunity page


brianlee-nava:
### Summary

This is a list of follow up work related to #8596 . Must be discussed, refined, prioritized.

(note that this depends on moving the agency id from path to query param in ticket tbd)

### Acceptance criteria

- [ ] check for valid agency id in query param on page load
- [ ] convert to using query param for the agency id
- [ ] make sure any inbound links into the page are not broken
- [ ] check that agency id in query param exists in users agencies, and if not, present message [content TBD]

brianlee-nava:
@Dao-REI CC: @brianlee-nava As a follow up for refinement conversation, @crystabelrangel we would like you to weigh on design treatment for validating agency on page load [refer to bullets above]. Let's plan to have after party on this during Grantor Standup Wednesday 3/25.

ssatia-SGM:
@Dao-REI to  update the acceptance criteria

Dao-REI:
For UAT I demo'ed this ticket to the team at Stand-up After Party on 4/1/2026, including @crystabelrangel, @doug-s-nava, @glenblosser-nava and @desaiamit. Now moving this ticket to Done.

PR #9355, opened 2026-03-31
[Issue #9232] Create Opportunity page to use a Query Param for the Agency Id

Dao-REI:
## Summary

Work for #9232

## Changes proposed

Change from reading the Agency Id in the URL to using a query param. And update inbound links into the page.

## Context for reviewers

To remove the Agency Id from the URL, files were moved from the src/app/[locale]/(base)/opportunities/create/[agencyId] folder to the src/app/[locale]/(base)/opportunities/create folder and the [agencyId] folder was deleted. Thus, you will not be able to see any code diff. The main code change was in the definition of the page and assignment of the agencyId variable:

```
type CreateOpportunityProps = LocalizedPageProps & WithFeatureFlagProps;

async function CreateOpportunityPage(props: CreateOpportunityProps) {
  const { searchParams } = props;
  const resolvedSearchParams: Record<string, string | string[] | undefined> =
    searchParams ? await searchParams : {};
  const selectedAgencyParam: string | string[] | undefined =
    resolvedSearchParams.agency;
  const agencyId: string | undefined = Array.isArray(selectedAgencyParam)
    ? selectedAgencyParam[0]
    : selectedAgencyParam;
```

## Validation steps

Code review and UAT:

1. Verify that the Create Opportunity button on the Opportunity List page will correctly take the user to the Create Opportunity page and default select the agency in the query param. 
2. Verify that the page loads with an agency param and the Agency dropdown default selects that agency. E.g. http://localhost:3000/opportunities/create?agency=4dcd3a02-60d3-49a9-b2f8-8e42ad8b1a6b
3. Verify that the page loads without an agency param and the Agency dropdown has no default selection. E.g. http://localhost:3000/opportunities/create
4. Verify that the page loads even with an invalid agency id or an agency that the user does not belong to. In this case the Agency dropdown has no default selection. E.g. http://localhost:3000/opportunities/create?agency=a-bad-agency-id

### DIFF
--- a/frontend/src/app/[locale]/(base)/opportunities/create/page.tsx
+++ b/frontend/src/app/[locale]/(base)/opportunities/create/page.tsx
@@ -3,9 +3,7 @@ import { UnauthorizedError } from "src/errors";
 import { getSession } from "src/services/auth/session";
 import withFeatureFlag from "src/services/featureFlags/withFeatureFlag";
 import { fetchUserAgencies } from "src/services/fetch/fetchers/agenciesFetcher";
-import { LocalizedPageProps } from "src/types/intl";
 import { RelevantAgencyRecord } from "src/types/search/searchFilterTypes";
-import { WithFeatureFlagProps } from "src/types/uiTypes";
 
 import { useTranslations } from "next-intl";
 import { redirect } from "next/navigation";


frontend/src/app/[locale]/(base)/opportunities/create/page.tsx
```diff
@@ -62,11 +64,18 @@ const PageHeader = () => {
...
 
 // --- Main Page ---
-type FormPageProps = {
-  params: Promise<{ agencyId: string; locale: string }>;
-};
-async function CreateOpportunityPage({ params }: FormPageProps) {
-  const { agencyId } = await params;
+type CreateOpportunityProps = LocalizedPageProps & WithFeatureFlagProps;
+
+async function CreateOpportunityPage(props: CreateOpportunityProps) {
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunities/create/page.tsx]: [nit] this can be cleaned up a bit, noting that in your change you're implicitly using the typing for searchParams from the NextJS doc, but explicitly using the way we have generally typed it in our app (not expecting an array). Definitely not a bad idea to cover the case where we get an array, though for that to happen the user would have to add the query param to the url twice (ex. /create?agency=1&agency=2)

```suggestion

async function CreateOpportunityPage({
  searchParams,
}: {
  searchParams: Promise<{ [key: string]: string | string[] | undefined }>;
}) {
  const resolvedSearchParams = await searchParams;
  const selectedAgencyParam = resolvedSearchParams?.agency;
  const agencyId = Array.isArray(selectedAgencyParam)
    ? selectedAgencyParam[0]
    : selectedAgencyParam;
```

Comment (Dao-REI) [reply frontend/src/app/[locale]/(base)/opportunities/create/page.tsx]: I actually copied the code from the Opportunity List page and didn't think about it too much. But I'd like to accept your suggested changes.

doug-s-nava (approved):

PR #9370, opened 2026-03-31
[Issue #9202] Replace localstack with s3mock + elasticmq

### LINKED ISSUE #9202: Find and implement an alternative to localstack for local development
### Summary

We currently use localstack for having a mock AWS for local development - specifically for s3 and SQS. Localstack is changing their pricing model and more-or-less making it difficult to continue using them. We should look for an alternative. Note - our unit tests are unaffected by this as we don't use localstack for anything but running the API service locally (which DOES affect frontend e2e tests).

One possibility to consider - moto, the library we use for unit tests has a server mode we should test out.
https://docs.getmoto.org/en/latest/docs/server_mode.html

### Acceptance criteria

- [ ] Localstack replaced - new approach added to make commands
- [ ] Docs updated to replace references to localstack

chouinar:
## Summary
Fixes #9202

## Changes proposed
* Replace localstack with s3mock + elasticmq

## Context for reviewers
Localstack community edition is going away - https://blog.localstack.cloud/the-road-ahead-for-localstack - we temporarily worked around this problem by statically defining the image we use, but that was always temporary. 

While I considered using [moto](https://docs.getmoto.org/en/latest/docs/server_mode.html) - they do not have any way to persist data if you stop and start containers, a feature we really want for at least s3.

Instead I found s3mock + elasticmq which conveniently both have a way to configure s3 buckets / sqs queues to be created for you, letting us remove the create script we had before.

Because we've had issues before with disk space in our github actions, I verified how large the images are.
* `localstack` - 1.8gb
* `s3mock` - 230mb
* `elasticmq` - 131mb

So this should give us more leeway there.

## Validation steps
Don't have 100% exact step-by-step instructions, you can check it out and try running the commands. `make init` should be all you need to do to get this stood up. Just giving a checklist of what I checked.

---

Verified CI/CD works - e2e tests use the s3 setup.

Verified that I could use the AWS cli to interact with the s3 bucket + sqs queue - added a few of these commands to the docs.

Verified that the seed script is able to load data (many things load data to s3).

Verified via the frontend (and thus also our APIs) that files could be fetched/uploaded for:
* Opportunity attachments on the opportunity page / GET opp endpoint
* Can upload attachments to an application
* Can fetch an application submission after that is generated on the application page

The above means it works for both our draft and public bucket, although locally those behave identically since there is no CDN, although the logic branches a bit before converging.

---

For SQS, we only have one usage of it, I was able to send events via the workflow event API, and the service was able to pull from them.


documentation/api/development.md
kkrug: This is now port 9090, right?
```suggestion
S3_ENDPOINT_URL=http://localhost:9090
```

api/docker-compose.yml
```diff
@@ -51,24 +51,29 @@ services:
...
 
-  localstack:
-    container_name: "${LOCALSTACK_DOCKER_NAME:-localstack-main}"
-    image: localstack/localstack:4.14.0
-    ports:
-      - "127.0.0.1:4566:4566" # LocalStack Gateway
-      - "127.0.0.1:4510-4559:4510-4559" # external services port range
+  s3mock:
+    container_name: s3mock
+    image: adobe/s3mock:latest
```
kkrug, line 56: When we were using localstack, we pinned the image to a specific version. Are we ok with using the latest for this (and also the sqsmock below which implicitly uses latest)?

api/docker-compose.yml
```diff
@@ -51,24 +51,29 @@ services:
...
     networks:
       - default
 
+  sqsmock:
+    container_name: sqsmock
+    image: softwaremill/elasticmq-native
+    ports:
+      - 9324:9324
+    volumes:
+      - ./mock-sqs/custom.conf:/opt/elasticmq.conf
```
kkrug, line 74: Should this also have `networks: - default` for consistency? I don't think this affects anything in practice but it's the only services without this.

api/docker-compose.yml
chouinar: We were using latest for localstack until I changed it last week when their latest stopped being supported without reconfiguring all of it.

I'd say unless we see a reason to not use latest, it's probably fine to use it. I'll take a look and see if there's any recommendation, or how often they upgrade/have major versions.

kkrug: Ah gotcha, I think latest is ok then - we can reevaluate if it causes any issues down the road.

chouinar: S3Mock - https://hub.docker.com/r/adobe/s3mock/tags
elasticMq - https://hub.docker.com/r/softwaremill/elasticmq-native/tags

Neither of them have maintained major versions that I could pick (eg. I can't just do like `1` or `1.6`). So our options are exact tag or latest. I think I'm fine with latest, these look pretty slow changing / if they break, we can always fix easily by selecting a tag like I did for localstack.

kkrug (approved):

PR #9374, opened 2026-04-01
[Issue #8768] Create endpoint for paginated workflow audit event

### LINKED ISSUE #8768: Create a workflow endpoint that gets the history of transition events
### Summary

Create`POST /v1/workflows/:workflow_id/audit` - this endpoint will return a paginated list of audit events about a given workflow.

### Request

```python
{
  # No filters yet
  "pagination": {
    "page_offset": 1,
    "page_size": 25,
    "sort_order": [
       # Only need to sort by created_at at the moment.
      {
        "order_by": "created_at",
      }
    ]
  }
}
```

### Response

```python
{
"data": [
   {
      "workflow_audit_id": "<uuid>",
      "acting_user": {
         "user_id": "<uuid>",
         "first_name": "",
         "last_name": "",
         "email": "",
      },
      "transition_event": "",
      "source_state": "",
      "target_state": "",
      "event": {
         "event_id": "<uuid>",
         "sent_at": "",
      },
      "audit_metadata": {}, # a dictionary, non-specific contents, nullable
   }
 ],
  "message": "Success",
  "pagination_info": {
    "page_offset": 1,
    "page_size": 25,
    "sort_order": [
      {
        "order_by": "created_at",
        "sort_direction": "ascending"
      }
    ],
    "total_pages": 2,
    "total_records": 42
  },
  "status_code": 200
}
```

### AuthZ
Allows JWT or user API key

The same as what we implemented in the GET workflow endpoint
https://github.com/HHS/simpler-grants-gov/issues/8107

### Logic
* Fetch the workflow, 404 if not found
* Enforce AuthZ (same as GET workflow endpoint), 403 if not authorized
* Using our pagination utils, construct a request, and return a paginated response

### Acceptance criteria

- [ ] Endpoint created
- [ ] Tests added that verify audit events can be returned with pagination

hao10282025-sudo:
## Summary

Fixes #8768  

## Changes proposed

Create POST /v1/workflows/:workflow_id/audit - this endpoint will return a paginated list of audit events about a given workflow.

## Context for reviewers

Request and response schemas can be found at the ticket summary: https://github.com/HHS/simpler-grants-gov/issues/8768

AuthZ
Allows JWT or user API key
The same as what we implemented in the GET workflow endpoint
https://github.com/HHS/simpler-grants-gov/issues/8107

Logic

- Fetch the workflow, 404 if not found
- Enforce AuthZ (same as GET workflow endpoint), 403 if not authorized
- Using our pagination utils, construct a request, and return a paginated response

## Validation steps

- Endpoint created
- Tests added that verify audit events can be returned with pagination

### DIFF
--- a/api/src/services/workflows/get_workflow_audits.py
+++ b/api/src/services/workflows/get_workflow_audits.py
@@ -35,13 +35,16 @@ def get_workflow_audits(
 
     # First verify the user has access to the workflow
     # This will raise appropriate errors if workflow doesn't exist or user lacks access
-    _ = get_workflow_and_verify_access(db_session, user, workflow_id)
+    get_workflow_and_verify_access(db_session, user, workflow_id)
 
     # Build the base query for workflow audit events
     stmt = (
         select(WorkflowAudit)
         .where(WorkflowAudit.workflow_id == workflow_id)
-        .options(selectinload(WorkflowAudit.acting_user))
+        .options(
+            selectinload(WorkflowAudit.acting_user).options(selectinload(User.profile)),
+            selectinload(WorkflowAudit.event),
+        )
     )
 
     stmt = apply_sorting(stmt, WorkflowAudit, params.pagination.sort_order)
--- a/api/tests/src/api/workflows/test_workflow_audit_routes.py
+++ b/api/tests/src/api/workflows/test_workflow_audit_routes.py
@@ -1,4 +1,3 @@
-import time
 import uuid
 
 import pytest
@@ -43,8 +42,14 @@ class TestWorkflowAuditEndpoint:
     ):
         """Test successfully fetching an opportunity workflow with audits and approvals."""
         # Create a user with VIEW_OPPORTUNITY privilege in the agency
+        # if first_name or last_name provided, user profile will be created
         user, _, token = create_user_in_agency_with_jwt(
-            db_session, agency=agency, privileges=[Privilege.VIEW_OPPORTUNITY]
+            db_session,
+            agency=agency,
+            privileges=[Privilege.VIEW_OPPORTUNITY],
+            email="testWorkflowAudit@example.com",
+            first_name="testFirstNameWA",
+            last_name="testLastNameWA",
         )
 
         # Create workflow
@@ -55,8 +60,11 @@ class TestWorkflowAuditEndpoint:
             opportunity=opportunity,
         )
 
+        event_ids = []
+
         # Create audit events
         start_event = WorkflowEventHistoryFactory.create()
+        event_ids.append(str(start_event.event_id))
 
         audit_events = []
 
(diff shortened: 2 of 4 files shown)


api/src/services/workflows/get_workflow_audits.py
```diff
@@ -0,0 +1,56 @@
...
+
+    Raises:
+        404: If workflow doesn't exist
+        403: If user doesn't have required privilege
+    """
+    params = WorkflowAuditRequest.model_validate(json_data)
+
+    # First verify the user has access to the workflow
+    # This will raise appropriate errors if workflow doesn't exist or user lacks access
+    _ = get_workflow_and_verify_access(db_session, user, workflow_id)
```
chouinar:
```suggestion
    get_workflow_and_verify_access(db_session, user, workflow_id)
```

api/src/services/workflows/get_workflow_audits.py
```diff
@@ -0,0 +1,56 @@
...
+
+    # Build the base query for workflow audit events
+    stmt = (
+        select(WorkflowAudit)
+        .where(WorkflowAudit.workflow_id == workflow_id)
+        .options(selectinload(WorkflowAudit.acting_user))
```
chouinar: We'll need to also select the `User.profile` in the load I think since if that exists we get the name out of it? Can you make sure at least one of the users attached in the test has a profile set to be safe?

api/tests/src/api/workflows/test_workflow_audit_routes.py
```diff
@@ -0,0 +1,220 @@
...
+            WorkflowAuditFactory.create(
+                workflow=workflow,
+                acting_user=user,
+                transition_event="start_workflow",
+                source_state=BasicState.START,
+                target_state=BasicState.MIDDLE,
+                event_id=start_event.event_id,
+            )
+        )
+        time.sleep(0.1)
```
chouinar: Don't put sleeps in tests, even if it seems small, we don't want to start that pattern when we have 3000+ tests. The `created_at` timestamps should already be distinct since datetime is 6 digits of precision (nanoseconds I think?). Should be safe to assume the order matches the create order, we do that in some other tests.

If for whatever reason you do end up with ties and the sorting ends up messed up, a simpler approach would be something like:
```py
for i in range(4):
    WorkflowAuditFactory.create(
         ...,
         created_at = datetime(2026, 1, 1, 12, 0, i, tzinfo=...)
```

Or could make a base timestamp of now, and add i seconds, either would work.

api/src/services/workflows/get_workflow_audits.py
hao10282025-sudo: updated

api/tests/src/api/workflows/test_workflow_audit_routes.py
hao10282025-sudo: updated, removed the lines for sleep

api/src/services/workflows/get_workflow_audits.py
hao10282025-sudo: Added the User.profile
Also update the event_id to event object, in workflow_schemas.py

Issue #9379, opened 2026-04-01
E2E tests not creating correct reports on deploy


doug-s-nava:
### Summary

When running e2e tests on prod deploy, an error occurs when uploading test run artifacts due to local and staging jobs running on the same run id. See https://github.com/HHS/simpler-grants-gov/actions/runs/23857695114/job/69555413632

```
[9](https://github.com/HHS/simpler-grants-gov/actions/runs/23857695114/job/69555413632#step:5:1554)
Run actions/upload-artifact@v6
With the provided path, there will be 1 file uploaded
Artifact name is valid!
Root directory input is valid!
Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run
```

### Reproducibility

None

### How to reproduce the bug

Would need to either run a prod deploy or an experimental job set up similar to the e2e runs in the release job

### Browser

None

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #9384, opened 2026-04-01
[Issue #9379] fix e2e reports on deploy

doug-s-nava:
## Summary

Fixes for #9379

## Changes proposed

The prod release job runs e2e jobs against staging and local, and as a result, the run ids used as identifiers on report artifacts are not unique across the staging and local runs. This change adds environment specificity to the artifact names to fix this issue.

## Validation steps

1. _VERIFY_: test run completed and successfully created test reports for both staging and local test runs, see https://github.com/HHS/simpler-grants-gov/actions/runs/23859955747/job/69567950377


btabaska:
E2E tests were failed in the linked action, but I think it was a flaky timeout, on second run everything passed and the reports were generated as expected. Code looks good to me, no issues that I can see.

Issue #8714, opened 2026-03-02
Backend - Create non-core Award Recommendation tables


nathan-stricker:
### Summary

See [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Create-the-rest-of-the-tables)

Implement the DB tables for:
- [Award Recommendation Attachment](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Award-Recommendation-Attachment)
- [Award Recommendation Risk](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Award-Recommendation-Risk)
- [Award Recommendation Risk Submission](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Award-Recommendation-Risk-Submission)
- [Award Recommendation Audit](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Award-Recommendation-Audit)
- [Award Recommendation Review](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Award-Recommendation-Review)

Factories should always be setup as part of DB table creation tickets.

### Acceptance criteria

- [x] Tables and factories created:
    - [x] `award_recommendation_attachment`
    - [ ] `award_recommendation_risk`
    - [ ] `award_recommendation_risk_submission`
    - [ ] `award_recommendation_audit`
    - [ ] `award_recommendation_review`

PR #9387, opened 2026-04-01
[Issue #8714] Create non-core Award Recommendation tables

nathan-stricker:
## Summary

Work for #8714 

## Changes proposed

Database updates for Award Recommendation work.
- Tables and factories added:
    - `award_recommendation_attachment`
    - `award_recommendation_risk`
    - `award_recommendation_risk_submission`
    - `award_recommendation_audit`
    - `award_recommendation_review`

## Context for reviewers

See the Award Recommendation tech spec for specifications for the tables created:
- [Award Recommendation Attachment](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Award-Recommendation-Attachment)
- [Award Recommendation Risk](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Award-Recommendation-Risk)
- [Award Recommendation Risk Submission](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Award-Recommendation-Risk-Submission)
- [Award Recommendation Audit](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Award-Recommendation-Audit)
- [Award Recommendation Review](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Award-Recommendation-Review)

For more context, see [earlier PR creating the core Award Rec tables](https://github.com/HHS/simpler-grants-gov/pull/8946)

## Validation steps

No manual verification; added Factory methods for new tables.

### DIFF
--- a/api/src/db/models/award_recommendation_models.py
+++ b/api/src/db/models/award_recommendation_models.py
@@ -4,7 +4,7 @@ import uuid
 from decimal import Decimal
 from typing import TYPE_CHECKING
 
-from sqlalchemy import ForeignKey, Numeric
+from sqlalchemy import ForeignKey, Numeric, and_
 from sqlalchemy.dialects.postgresql import JSONB, UUID
 from sqlalchemy.orm import Mapped, mapped_column, relationship
 
@@ -83,11 +83,44 @@ class AwardRecommendation(ApiSchemaTable, TimestampMixin):
     )
     award_recommendation_attachments: Mapped[list[AwardRecommendationAttachment]] = relationship(
         "AwardRecommendationAttachment",
-        back_populates="award_recommendation",
         uselist=True,
-        cascade="all, delete-orphan",
+        primaryjoin=lambda: and_(
+            AwardRecommendation.award_recommendation_id
+            == AwardRecommendationAttachment.award_recommendation_id,
+            AwardRecommendationAttachment.is_deleted.isnot(True),
+        ),
+        # This version of the relationship is view-only and excludes deleted records.
+        # For the one that can be used to modify, use _all_award_recommendation_attachments.
+        viewonly=True,
     )
+
+    # Relationship that gets all award recommendation attachments INCLUDING DELETED
+    # We likely don't want to use this in most cases, preferring the above
+    # one which has only non-deleted ones.
+    _all_award_recommendation_attachments: Mapped[list[AwardRecommendationAttachment]] = (
+        relationship(
+            "AwardRecommendationAttachment",
+            back_populates="award_recommendation",
+            uselist=True,
+            cascade="all, delete-orphan",
+        )
+    )
+
     award_recommendation_risks: Mapped[list[AwardRecommendationRisk]] = relationship(
+        "AwardRecommendationRisk",
+        uselist=True,
+        primaryjoin=lambda: and_(
+            AwardRecommendation.award_recommendation_id
+            == AwardRecommendationRisk.award_recommendation_id,
+            AwardRecommendationRisk.is_deleted.isnot(True),
+        ),
+        viewonly=True,
+    )
+
+    # Relationship that gets all award recommendation risks INCLUDING DELETED
+    # We likely don't want to use this in most cases, preferring the above
+    # one which has only non-deleted ones.
+    _all_award_recommendation_risks: Mapped[list[AwardRecommendationRisk]] = relationship(
         "AwardRecommendationRisk",
         back_populates="award_recommendation",
         uselist=True,
@@ -117,7 +150,7 @@ class AwardRecommendationAttachment(ApiSchemaTable, TimestampMixin):
         index=True,
     )
     award_recommendation: Mapped[AwardRecommendation] = relationship(
-        AwardRecommendation, back_populates="award_recommendation_attachments"
+        AwardRecommendation, back_populates="_all_award_recommendation_attachments"
     )
     file_location: Mapped[str]
     file_name: Mapped[str]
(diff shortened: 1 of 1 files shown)


api/src/db/models/award_recommendation_models.py
```diff
@@ -69,6 +81,94 @@ class AwardRecommendation(ApiSchemaTable, TimestampMixin):
...
+        ForeignKey("api.user.user_id"),
+    )
+    uploading_user: Mapped[User] = relationship("User")
+    is_deleted: Mapped[bool] = mapped_column(default=False)
+
+    @property
+    def download_path(self) -> str | None:
+        if self.file_location:
+            return presign_or_s3_cdnify_url(self.file_location)
+        return None
```
chouinar: Doesn't change anything practically, but can simplify a little. The file location can't be null/empty, and we should use just the presign function since these files would never be in the public bucket. 

```suggestion
    @property
    def download_path(self) -> str:
        return pre_sign_file_location(self.file_location)
```

api/src/db/models/award_recommendation_models.py
```diff
@@ -1,28 +1,40 @@
...
     LkAwardRecommendationType,
     LkAwardSelectionMethod,
 )
+from src.util.file_util import presign_or_s3_cdnify_url
 
 if TYPE_CHECKING:
     from src.db.models.competition_models import ApplicationSubmission
     from src.db.models.opportunity_models import Opportunity
-    from src.db.models.workflow_models import Workflow
+    from src.db.models.user_models import User
```
chouinar: If these don't have a circular dependency, can we import them normally? that way we can refer to them below (eg. using `User` instead of `"User"`).

api/tests/src/db/models/factories.py
```diff
@@ -968,6 +972,46 @@ class Meta:
     review_workflow_id = factory.LazyAttribute(lambda s: s.review_workflow.workflow_id)
 
 
+class AwardRecommendationAttachmentFactory(BaseFactory):
```
chouinar, line 975: Could you add the `_build` / `_create` functions that we have on some of the other factories like the opportunity attachment factory?

In short, makes it so when you use the factory, it'll also create a file in s3. Will also need a `file_contents` value with something in it. Fine if the actual contents of the file is just random text, just makes it easier when we later are setting up data for local dev / testing to actually have files so download links work.

api/tests/src/db/models/factories.py
```diff
@@ -1010,6 +1070,66 @@ class Meta:
...
+
+    award_recommendation = factory.SubFactory(AwardRecommendationFactory)
+    award_recommendation_id = factory.LazyAttribute(
+        lambda s: s.award_recommendation.award_recommendation_id
+    )
+    award_recommendation_review_type = factory.fuzzy.FuzzyChoice(AwardRecommendationReviewType)
+    is_reviewed = False
+
+
+class AwardRecommendationAuditFactory(BaseFactory):
```
chouinar, line 1114: I'd probably set this up closer to the `ApplicationAuditFactory` with different scenarios via Params - we wouldn't have all of the various IDs set for every audit record, most of them are only going to be set in specific cases (eg. the attachment is only set if it was for an attachment).

Don't worry about getting every possible scenario figured out, put some obvious ones, we can always add more and adjust later.

api/src/db/models/award_recommendation_models.py
```diff
@@ -69,6 +81,92 @@ class AwardRecommendation(ApiSchemaTable, TimestampMixin):
         uselist=True,
         cascade="all, delete-orphan",
     )
+    award_recommendation_attachments: Mapped[list[AwardRecommendationAttachment]] = relationship(
```
chouinar, line 84: I didn't think of it until now, but for any of these relationships where the table has an `is_deleted` flag, we'll want the default of these relationships to not include the deleted record. That way the GET award rec endpoint won't return deleted rows.

Two ways to go about it that I can see.

For either case, we'd need a relationship with a slightly different name saying to get all of them, same as what you have, just change the name.
```py
     all_award_recommendation_attachments: Mapped[list[AwardRecommendationAttachment]] = relationship(...)
```

Then we can do either:
```py
@property
def award_recommendation_attachments(self) -> list[AwardRecommendationAttachment]:
     return [att for att in self.all_award_recommendation_attachments if att.is_deleted is False]
```

or keeping it purely in SQLAlchemy, define the join on the relationship (we do this for application's application attachments).

```py
    award_recommendation_attachments: Mapped[list[AwardRecommendationAttachment]] = relationship(
        "AwardRecommendationAttachment",
        uselist=True,
        primaryjoin=lambda: and_(
            AwardRecommendation.award_recommendation_id == AwardRecommendationAttachment.award_recommendation_id,
            AwardRecommendationAttachment.is_deleted.isnot(True),
        ),
        # This version of the award recommendation attachment relationship is view-only
        # For the one that can be used to modify, use the all_award_recommendation_attachments above.
        viewonly=True,
    )
```

---

Note - don't filter out deleted in the joins in the audit table. The audit table is the one place we'd want the deleted records (it's the main user of deleted records).

nathan-stricker: Good catch, thanks for calling this out- just updated to include those changes

Issue #9252, opened 2026-03-25
SOAP/Proxy: Expand validation command (GetApplicationZip)


jakobpederson:
### Summary

The current validate endpoint command needs to be expanded in order to actually validate response from endpoints building off of Freddie's spike. 

### Context
There's this code:
```
    schema_tree = etree.parse(filepath)
    schema_validator = etree.XMLSchema(schema_tree)
    schema_validator.assertValid(operation_element)
```
The [etree.XMLSchema](https://lxml.de/api/lxml.etree.XMLSchema-class.html) should allow us to pass ing a schema or file and then validate the response. (This needs confirming)

So we will want to potentially:
1. Hit a locally running endpoint
2. Validate the response against the schemas we get

### Acceptance criteria

- [ ] actually hit an endpoint (GetApplicationZip)
- [ ] validate endpoint against xsd/wsdls

PR #9389, opened 2026-04-01
[Issue #9252] Add GetApplicationZip soap endpoint validation

jakobpederson:
## Summary

Fixes / Work for #9252 

## Changes proposed

- Actually hit an environment (I was hitting staging)
- Update the requests to also support hitting local instance
- New method to build a schema validator

## Context for reviewers

The older version was testing static responses, this actually hits an endpoint and then validates the response.

## Validation steps

- [ ] add cert, key, and uri to the soap-api.env file
- [ ] Run command and check to see if stuff validates for GetApplicationZip

### DIFF
--- a/api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
+++ b/api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
@@ -9,18 +9,20 @@ from urllib.parse import quote
 import click
 import requests
 from lxml import etree
+from pydantic import Field
 
 import src.adapters.db as db
 import src.logging
 import tests.src.db.models.factories as factories
 from src.adapters.db import PostgresDBClient
+from src.util.env_config import PydanticBaseEnvConfig
 from src.util.local import error_if_not_local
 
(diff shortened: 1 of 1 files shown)


api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
```diff
@@ -34,9 +31,11 @@
 To run use: `make validate-simpler-endpoints``
 """
 
-
-SOAP_URI = os.getenv("SOAP_URI", b"")
-HEADERS = {"Content-Type": "application/xml"}
+SOAP_URI = os.getenv(
+    "SOAP_URI",
+    b"http://host.docker.internal:8080/grantsws-agency/services/v2/AgencyWebServicesSoapPort",
+)
```
chouinar: Don't use getenv, use a pydantic class - same with other usages of getenv below - makes it a lot easier to organize all env vars together and give defaults (when they make sense). Don't configure any docker default here, put it in local.env. There are different ways of running our code, don't assume Docker (I can't run this).

PR #9395, opened 2026-04-02
[Issue #3423] Fix header nav a11y bug

### LINKED ISSUE #3423: [Brand FE] improve low-contrast Header links
### Summary

The links in the Header component are currently failing a11y contrast checks. The gray default state barely passes AA (fails AAA). But the green hover state fails AA. This is a result of adding our custom color palette. To remedy, we need to define explicit colors to use for the Header, as the defaults are too low-contrast. 

Unfortunately, this is not as simple as defining a theme setting since `$nav-link-color` [is hardcoded](https://github.com/uswds/uswds/blob/44e464c848edc08bc0b0b0ac80c50e68b9d2488e/packages/usa-nav/src/styles/_usa-nav.scss#L12) to use `base-dark`. 

[A bug has been filed for this](https://github.com/uswds/uswds/issues/6176) and the USWDS team has decided to add a theme setting. 

If we do not want to wait for this patch and want to change this now, it will require custom CSS to override the hardcoded default. 

### Acceptance criteria

- [ ] All states of Header links pass AA on https://webaim.org/resources/contrastchecker/

andycochran:
## Summary

Fixes #3423 

## Changes proposed

This PR adds custom style overrides for the primary nav and mobile "Menu" button. 

#### Dropdown (new = bottom): 

Only change is the default menu color is a slightly darker gray. 

image

#### Hover (new = bottom): 

The green hover color is darker. 

image

#### Menu button (new = right): 

Menu button darker and styled like our other buttons. 

image

#### Mobile menu (new = right):

Only change is the default menu color is a slightly darker gray. 

image

#### Mobile hover (new = right):

image

## Context for reviewers

This is necessary because USWDS hardcodes a bunch of color variables in their menu SCSS, and it's not likely a fix is coming soon. 

## Validation steps

- View the site at mobile viewport size
  - The "MENU" button should be darker and pass AA contrast check
  - Open the menu
    - The menu items should be darker and pass AA contrast check (also when hovering) 
- View the site at desktop viewport size
  - The menu items should be darker and pass AA contrast check
  - When hovering, the green should be darker
  - Open a dropdown
    - It should be unaffected (text is still white)

### DIFF
--- a/api/src/legacy_soap_api/grantors/schemas/get_submission_list_expanded_schemas.py
+++ b/api/src/legacy_soap_api/grantors/schemas/get_submission_list_expanded_schemas.py
@@ -5,6 +5,7 @@ from typing import Any
 from pydantic import BaseModel, Field, field_serializer, field_validator, model_validator
 
 from src.legacy_soap_api.legacy_soap_api_schemas import BaseSOAPSchema, SOAPInvalidEnvelope
+from src.util.datetime_util import make_timezone_aware
 
 
 class SubmissionInfo(BaseSOAPSchema):
@@ -19,6 +20,13 @@ class SubmissionInfo(BaseSOAPSchema):
     delinquent_federal_debt: str | None = Field(alias="DelinquentFederalDebt")
     active_exclusions: str | None = Field(alias="ActiveExclusions")
     uei: str | None = Field(alias="UEI")
+
+    @field_validator("received_date_time", mode="before")
+    @classmethod
+    def ensure_timezone_aware(cls, received_date_time: datetime | None) -> datetime | None:
+        if isinstance(received_date_time, datetime) and received_date_time.tzinfo is None:
+            return make_timezone_aware(received_date_time, "US/Eastern")
+        return received_date_time
 
     @field_serializer("received_date_time")
     def serialize_dt(self, dt: datetime) -> str:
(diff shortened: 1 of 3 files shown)


frontend/src/styles/_uswds-theme-custom-styles.scss
```diff
@@ -495,6 +495,63 @@ h6 {
...
+    }
+
+    @include at-media-max($theme-header-min-width) {
+      button,
+      .usa-current {
+        &:hover {
+          background-color: color("base-lightest");
+        }
+      }
+    }
```
andycochran, line 538: This fixes another existing bug where, on mobile, links (`a` elements) get a hover background but dropdowns (`button` elements) don't.

btabaska:
Nice work on the contrast fix. We now have two top-level `.usa-nav__primary` blocks in the same file, which makes cascade behavior harder for folks to reason about later. Would it make sense to fold the new a11y section into the existing `.usa-nav__primary` block so nav behavior lives in one place?

frontend/src/styles/_uswds-theme-custom-styles.scss
```diff
@@ -495,6 +495,63 @@ h6 {
...
+  // $nav-link-color hardcoded to "base-dark" (which we theme to "gray-warm-50")
+  // &:hover is hardcoded to "primary" (which we theme to "mint-30")
+  // We need a magic number of 50+ to pass WCAG 2.0 AA contrast (70+ for AAA).
+  // Note that hover & current states have "gray-warm-5" bg on mobile,
+  // so we need to use color grades > 55 for AA.
+  $simpler-nav-link-color: "gray-warm-60";
+  $simpler-nav-link-color-hover: "mint-60";
+
+  > .usa-nav__primary-item {
+    > a,
```
btabaska, line 510: Are both >a and the more specific selector needed or would it be cleaner to just have the second more specific selector here?

Issue #8670, opened 2026-02-26
Seed data for GET award recommendation endpoint


kevlyu:
### Summary

Need to identify seed data for GET AR endpoint

### Acceptance criteria

tbd

kevlyu:
Pickup when there's available cycle @nathan-stricker @chay-REIsys. 

Please connect with Michael to get additional information.

kevlyu:
Copy instruction from another ticket

PR #9401, opened 2026-04-02
[Issue#8670] - setup local data for an award recommendation

chay-REIsys:
## Summary

Fixes  for #8670 

## Changes proposed

Seed data for Award Recommendation and Submission tables

- Main entry point: _build_award_recommendations(db_session)
- Integrated with existing seed infrastructure via seed_local_db.py
- Can be run independently: make db-seed-local args="-s award_recommendations"

## Context for reviewers

Created a comprehensive seed script (seed_award_recommendations.py) to populate test data for the award recommendation feature, enabling local development and demo preparation.

## Validation steps

- fetch changes
- navigate to api folder
- run command  make db-seed-local args="-s award_recommendations"
- verify records get added in award recommendation, award recommendation submission and award recommendation submission details

### DIFF
--- a/frontend/src/components/Header.tsx
+++ b/frontend/src/components/Header.tsx
@@ -307,6 +307,7 @@ const NavLinks = ({
       items={navItems}
       mobileExpanded={mobileExpanded}
       onToggleMobileNav={onToggleMobileNav}
+      className="padding-bottom-05"
     ></PrimaryNav>
   );
 };
--- a/frontend/src/components/homepage/sections/HomepageBuilding.tsx
+++ b/frontend/src/components/homepage/sections/HomepageBuilding.tsx
@@ -1,13 +1,13 @@
 import BuildingImage from "public/img/homepage-building.jpg";
+import { ExternalRoutes } from "src/constants/routes";
 
-import { useMessages, useTranslations } from "next-intl";
+import { useTranslations } from "next-intl";
 import Image from "next/image";
+import Link from "next/link";
 import { Grid, GridContainer } from "@trussworks/react-uswds";
 
 const BuildingContent = () => {
   const t = useTranslations("Homepage.sections.building");
-  const messages = useMessages() as unknown as IntlMessages;
-  const { paragraphs } = messages.Homepage.sections.building;
 
   return (
     <GridContainer
@@ -38,14 +38,25 @@ const BuildingContent = () => {
                 span: (chunks) => <span className="text-italic">{chunks}</span>,
               })}
             </h2>
-            {paragraphs.map((paragraph: string, paragraphIdx: number) => (
-              <p
-                className="font-sans-md tablet-lg:font-sans-lg"
-                key={paragraphIdx}
-              >
-                {paragraph}
-              </p>
-            ))}
+            <p className="font-sans-md tablet-lg:font-sans-lg">
+              {t("paragraph1")}
+            </p>
+            <p className="font-sans-md tablet-lg:font-sans-lg">
+              {t.rich("paragraph2", {
+                code: (chunks) => (
+                  <a
+                    href={ExternalRoutes.GITHUB_REPO}
+                    target="_blank"
+                    rel="noopener noreferrer"
+                    className="usa-link--external"
+                  >
+                    {chunks}
+                  </a>
+                ),
+                roadmap: (chunks) => <Link href="/roadmap">{chunks}</Link>,
+                vision: (chunks) => <Link href="/vision">{chunks}</Link>,
+              })}
+            </p>
           </Grid>
         </Grid>
       </Grid>
--- a/frontend/src/components/homepage/sections/HomepageExperimental.tsx
+++ b/frontend/src/components/homepage/sections/HomepageExperimental.tsx
@@ -1,3 +1,4 @@
+import { ExternalRoutes } from "src/constants/routes";
 import { UswdsIconNames } from "src/types/generalTypes";
 
 import { useMessages, useTranslations } from "next-intl";
@@ -24,7 +25,21 @@ const ExperimentalContent = () => {
         {t("tryLink")}
       </Link>
       <h4>{t("cantDoHeader")}</h4>
-      <p>{t("cantDoParagraph")}</p>
+      <p>
+        {t.rich("cantDoParagraph", {
+          grantsLink: (chunks) => (
+            <a
+              href={ExternalRoutes.GRANTS_HOME}
+              target="_blank"
+              rel="noopener noreferrer"
+              className="usa-link--external"
+            >
+              {chunks}
+            </a>
+          ),
+          roadmapLink: (chunks) => <Link href="/roadmap">{chunks}</Link>,
+        })}
+      </p>
       <Grid row className="padding-y-2" gap="md">
         {iconSections.map((iconSection, iconSectionIdx) => (
           <Grid col={6} key={`experimental-iconsection-${iconSectionIdx}`}>
--- a/frontend/src/components/homepage/sections/HomepageInvolved.tsx
+++ b/frontend/src/components/homepage/sections/HomepageInvolved.tsx
@@ -1,12 +1,10 @@
+import { ExternalRoutes } from "src/constants/routes";
+
 import { useTranslations } from "next-intl";
 import { Grid } from "@trussworks/react-uswds";
 
 import HomePageSection from "src/components/homepage/homePageSection";
 import IconInfo from "src/components/homepage/IconInfoSection";
-
-const codeLink = "https://wiki.simpler.grants.gov/get-involved/github-code";
-const discourseLink = "https://forum.simpler.grants.gov/";
-const participateLink = "https://ethn.io/91822";
 
 const InvolvedContent = () => {
   const t = useTranslations("Homepage.sections.involved");
@@ -26,12 +24,35 @@ const InvolvedContent = () => {
           <IconInfo
             description={t("technicalDescription")}
             iconName="code"
-            link={codeLink}
-            linkText={t("technicalLink")}
             title={t("technicalTitle")}
           />
           <p>
-            <a href={discourseLink} target="_blank" rel="noopener noreferrer">
+            <a
+              href={ExternalRoutes.FIDER}
+              target="_blank"
+              rel="noopener noreferrer"
+              className="usa-link--external"
+            >
+              {t("fiderLink")}
+            </a>
+          </p>
+          <p>
+            <a
+              href={ExternalRoutes.WIKI_CONTRIBUTE_CODE}
+              target="_blank"
+              rel="noopener noreferrer"
+              className="usa-link--external"
+            >
+              {t("technicalLink")}
+            </a>
+          </p>
+          <p>
+            <a
+              href={ExternalRoutes.FORUM}
+              target="_blank"
+              rel="noopener noreferrer"
+              className="usa-link--external"
+            >
               {t("discourseLink")}
             </a>
           </p>
@@ -40,10 +61,18 @@ const InvolvedContent = () => {
           <IconInfo
             description={t("participateDescription")}
             iconName="chat"
-            link={participateLink}
-            linkText={t("participateLink")}
             title={t("participateTitle")}
           />
+          <p>
+            <a
+              href={ExternalRoutes.ETHNIO_VOLUNTEER}
+              target="_blank"
+              rel="noopener noreferrer"
+              className="usa-link--external"
+            >
+              {t("participateLink")}
+            </a>
+          </p>
         </Grid>
       </Grid>
     </HomePageSection>
--- a/frontend/src/components/vision/sections/VisionGetThere.tsx
+++ b/frontend/src/components/vision/sections/VisionGetThere.tsx
@@ -1,3 +1,5 @@
+import { ExternalRoutes } from "src/constants/routes";
+
 import { useTranslations } from "next-intl";
 
 import VisionPageSection from "src/components/vision/visionPageSection";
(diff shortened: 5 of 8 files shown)


api/tests/lib/seed_award_recommendations.py
```diff
@@ -0,0 +1,485 @@
...
+            select(Competition)
+            .join(Application)
+            .join(ApplicationSubmission)
+            .where(Application.application_status == ApplicationStatus.ACCEPTED)
+            .distinct()
+            .limit(5)
+        )
+        .scalars()
+        .all()
+    )
```
chouinar: The seed script shouldn't assume / expect data to already exist. Its the one that makes the data. The only time we should be fetching data is if we are avoiding primary key issues.

I'd change this script to instead just make the scenarios you want, don't worry about what may already be there, if every time it runs it creates 5 more opportunities, that's fine, that's how it should behave.

This also has too much variance in what it could do that it'd be hard to use the data. These competitions wouldn't necessarily be associated with agencies that your users can access.

api/tests/lib/seed_award_recommendations.py
```diff
@@ -0,0 +1,485 @@
...
+        "Creating a competition with 25 accepted applications for award recommendation testing..."
+    )
+
+    # Create a competition (without instructions to avoid S3 dependency)
+    competition = factories.CompetitionFactory.create(
+        opportunity__opportunity_title="Award Recommendation Test Opportunity",
+    )
+
+    # Get existing organizations from database and reuse them
+    from src.db.models.entity_models import Organization
```
chouinar: Put all imports at the top of the file

api/tests/lib/seed_award_recommendations.py
```diff
@@ -0,0 +1,485 @@
...
+                competition_form=competition_form,
+                application_response={"test": "data"},
+            )
+
+        # Create application submission (skip S3 file creation)
+        factories.ApplicationSubmissionFactory(
+            application=app,
+            file_location=f"s3://test-bucket/submissions/{app.application_id}.zip",
+            file_contents="SKIP",
+        )
```
chouinar: We'll probably need to move it to somewhere more central, but for creating application submissions, I'd suggest using the `_add_application` in `seed_orgs_and_users.py`. It handles a lot of this for you and more - there's a ton of extra considerations for accepted applications that'll break if you don't include all of it.

api/tests/lib/seed_award_recommendations.py
```diff
@@ -0,0 +1,485 @@
...
+            )
+            continue
+
+        logger.info(
+            f"Processing competition {competition.opportunity.opportunity_number} with {len(applications)} accepted applications"
+        )
+
+        # Scenario 1: Draft award recommendation with few applications
+        draft_ar = factories.AwardRecommendationFactory.create(
+            opportunity_id=competition.opportunity_id,
```
chouinar: Always prefer using actual models over IDs. With factories especially, since with what you had it'll make an opportunity in the factory and then you set the ID separately. It'll work, but you just made an entire opportunity in the DB that isn't used.

```suggestion
        draft_ar = factories.AwardRecommendationFactory.create(
            opportunity=competition.opportunity,
```

api/tests/lib/seed_award_recommendations.py
```diff
@@ -0,0 +1,485 @@
...
+    logger.info(
+        f"Detail created - ID: {detail.award_recommendation_submission_detail_id}, "
+        f"Type stored: {detail.award_recommendation_type}"
+    )
+
+    factories.AwardRecommendationApplicationSubmissionFactory.create(
+        award_recommendation=award_recommendation,
+        application_submission=application_submission,
+        award_recommendation_submission_detail=detail,
+    )
```
chouinar: A trick to avoid needing factories for factories like this function. Add the variability into the factory that you want with traits.

In the opportunity factory we want to be able to control what type of opportunity it is and the status, which is info stored in the current opportunity summary + opportunity summary tables.

So, to avoid needing to setup factories of factories, we added traits. So if I do:
```py
OpportunityFactory.create(is_forecasted_summary=True)
```

That affects a few tables all at once. I'd probably suggest we do something similar in the `AwardRecommendationApplicationSubmissionFactory` for setting up various scenarios of the `AwardRecommendationSubmissionDetail`. Then you never really interact with the detail table directly, just its "parent" table.

api/tests/lib/seed_award_recommendations.py
```diff
@@ -0,0 +1,485 @@
...
+    AwardRecommendationType,
+    AwardSelectionMethod,
+)
+from src.db.models.award_recommendation_models import AwardRecommendation
+from src.db.models.competition_models import Application, ApplicationSubmission, Competition
+
+logger = logging.getLogger(__name__)
+
+
+def _build_award_recommendations(db_session: db.Session) -> None:
```
chouinar, line 30: This function is doing way too much, you've got for-loops that are 200+ lines long in places.

I do think you can clean this up a bit (don't fetch data, just make it), but even then, can we move some of those loops to be something like:

```py
for x in y:
    call_function(x)
```

Just to help un-indent things a bit?

api/tests/lib/seed_award_recommendations.py
```diff
@@ -0,0 +1,485 @@
...
+
+        if not applications:
+            logger.warning(
+                f"No accepted applications found for competition {competition.competition_id}"
+            )
+            continue
+
+        logger.info(
+            f"Processing competition {competition.opportunity.opportunity_number} with {len(applications)} accepted applications"
+        )
```
chouinar: Don't fetch apps, just make them.

api/tests/lib/seed_award_recommendations.py
```diff
@@ -0,0 +1,485 @@
...
+
+                _add_application_to_award_recommendation(
+                    db_session,
+                    draft_ar,
+                    app.application_submissions[0],
+                    recommended_amount=amount,
+                    award_recommendation_type=rec_type,
+                )
+                apps_added_to_draft += 1
+
```
chouinar: Just make the apps yourself and don't rely on arbitrary indexes in unknown length arrays. If we want 6 of a scenario, make 6 with the factory (factories can be called as `MyFactory.create_batch(size=5)` and it's the same as `.create()`)

api/tests/lib/seed_award_recommendations.py
```diff
@@ -0,0 +1,485 @@
...
+            award_recommendations_created.append(
+                (exception_ar, "With Exception", competition.opportunity.opportunity_number)
+            )
+
+    # Create a static award recommendation with known ID for testing
+    static_ar_id = uuid.UUID("b9c15d13-8ff1-4e15-80b8-3cf5acf84851")
+    existing_static_ar = db_session.get(AwardRecommendation, static_ar_id)
+
+    if existing_static_ar:
+        logger.info(f"Static award recommendation {static_ar_id} already exists, skipping creation")
```
chouinar: This is the one case of select that is fine to have here.

chouinar (changes requested):
There's too many branching paths and complexities in this - don't fetch or try to guess what data already exists in this script - just make what you need without worrying about what already exists. The seed script should ONLY care about state when it's making something with static IDs, which is only needed for certain test scenarios.

I'd also say that an input to this needs to be what agencies you're associating with the opportunities as we'll need agencies associated with these so any of the auth works.

api/tests/lib/seed_award_recommendations.py
chaitanya-duddella: after i create the applications instead of fetching i think this unknown length array issues won't persist..correct me if i am wrong ..

api/tests/lib/seed_local_db.py
```diff
@@ -527,4 +530,10 @@ def run_seed_logic(db_session: db.Session, seed_config: SeedConfig) -> None:
         _build_agencies_and_users(db_session)
     if seed_config.seed_e2e:
         _build_users_and_tokens(db_session)
+    if seed_config.seed_award_recommendations:
+        # Fetch agencies to associate with opportunities for auth
+        from src.db.models.agency_models import Agency
+
+        agencies = db_session.query(Agency).limit(5).all()
+        _build_award_recommendations(db_session, agencies=agencies if agencies else None)
```
chouinar: For the agencies, we setup a few that users can access in `_build_agencies_and_users` - we could have that return its agencies instead. Or, just call the award recommendation building from that logic, passing in the agencies it just made. That way there isn't any variance in which agencies we have award recs for / they're easy to find.

The pattern we have here of turning on/off parts of the seed script isn't really used and it's all fairly connected.

api/tests/lib/seed_award_recommendations.py
```diff
@@ -0,0 +1,435 @@
...
+        logger.info(f"Associating opportunity with agency: {agencies[0].agency_code}")
+
+    competition = factories.CompetitionFactory.create(**competition_kwargs)
+
+    logger.info("Creating 8 fresh organizations for applications")
+    organizations = []
+    for _ in range(8):
+        unique_uei = f"UEI{random.randint(10000000, 99999999):08d}"
+        org = factories.OrganizationFactory.create(sam_gov_entity__uei=unique_uei)
+        organizations.append(org)
```
chouinar, line 468: The factory handles setting a UEI already in the sam.gov entity - can exclude and just make the list of orgs in one line.

```suggestion
    organizations = OrganizationFactory.create_batch(size=8)
```

api/tests/lib/seed_award_recommendations.py
```diff
@@ -0,0 +1,435 @@
...
+            app_owner=org,
+            application_status=ApplicationStatus.SUBMITTED,
+        )
+
+        factories.ApplicationSubmissionFactory(
+            application=app,
+            file_location=f"s3://test-bucket/submissions/{app.application_id}.zip",
+            file_contents="SKIP",
+        )
+        app.application_status = ApplicationStatus.ACCEPTED
```
chouinar, line 486: Is there something preventing us from putting the status as ACCEPTED and letting the function setup the submission? Was it running into any issues?

api/tests/lib/seed_award_recommendations.py
```diff
@@ -0,0 +1,435 @@
...
+        f"Processing competition {competition.opportunity.opportunity_number} with {len(applications)} accepted applications"
+    )
+
+    award_recommendations_created = []
+    award_recommendations_created.extend(
+        _create_draft_scenario(db_session, competition, applications)
+    )
+    award_recommendations_created.extend(
+        _create_in_review_scenario(db_session, competition, applications)
+    )
```
chouinar, line 55: For these, rather than the function having to assume the list is 25 apps and find some subset, what if we pulled that up here.


```suggestion
    award_recommendations_created.extend(
        _create_draft_scenario(db_session, competition, applications[:10])
    )
    award_recommendations_created.extend(
        _create_in_review_scenario(db_session, competition, applications[10:15])
    )
```
and so on. Then those functions don't care about how big the list that came in is, they just process whatever as a list.

api/tests/lib/seed_award_recommendations.py
```diff
@@ -0,0 +1,435 @@
...
+            app.application_submissions[0],
+            recommended_amount=amount,
+            award_recommendation_type=rec_type,
+        )
+        apps_added += 1
+
+    logger.info(
+        f"Created draft AR {draft_ar.award_recommendation_number} with {apps_added} applications"
+    )
+    return [(draft_ar, "Draft", competition.opportunity.opportunity_number)]
```
chouinar, line 169: It's not a big deal since it's a test util, but generally if I start dealing with tuples of more than 3 things, I start just making container classes (dataclasses make that useful) just to make adjusting it and typing a bit less tedious.

api/tests/lib/seed_award_recommendations.py
```diff
@@ -0,0 +1,435 @@
...
+    - Multiple application submissions per award recommendation
+    - Different recommendation types (recommended for funding, not recommended, etc.)
+    - Various selection methods
+    """
+    logger.info("Creating award recommendations with application submissions")
+
+    competition, applications = _create_competition_with_accepted_applications(db_session, agencies)
+
+    logger.info(
+        f"Processing competition {competition.opportunity.opportunity_number} with {len(applications)} accepted applications"
```
chouinar: Nitpick: Probably just call it an opportunity in the logs - award rec doesn't really care about competitions except that they're needed for applications to be setup / the number is for the opportunity.

```suggestion
        f"Processing opportunity {competition.opportunity.opportunity_number} with {len(applications)} accepted applications"
```

api/tests/lib/seed_award_recommendations.py
```diff
@@ -0,0 +1,435 @@
...
+def _create_static_scenario(
+    db_session: db.Session, competition: Competition, applications: list[Application]
+) -> list[tuple]:
+    """Create static award recommendation with known ID for E2E testing."""
+    static_ar_id = uuid.UUID("b9c15d13-8ff1-4e15-80b8-3cf5acf84851")
+    existing_static_ar = db_session.get(AwardRecommendation, static_ar_id)
+
+    if existing_static_ar:
+        logger.info(f"Static award recommendation {static_ar_id} already exists, skipping creation")
+        return []
```
chouinar, line 330: Just confirming you want this behavior. Since the way the script works is first creating a new opportunity + applications, they award recommendation will only be added to the first opportunity you create. If you run the script a second time, another opportunity will be made, but won't have this.

Also with how the script is setup, we're making several award recommendations for each opportunity, is that fine? Not sure if multiple per is a common case.

api/tests/lib/seed_award_recommendations.py
chaitanya-duddella: yes related to some s3 bucket issues

api/tests/lib/seed_award_recommendations.py
chaitanya-duddella: THis is giving unique UEI issues
image

api/tests/lib/seed_award_recommendations.py
chaitanya-duddella: yes i don't think it is a common case in real scenario but for testing puropses i chose this approach..

api/tests/lib/seed_award_recommendations.py
```diff
@@ -0,0 +1,502 @@
...
+        logger.info(f"Created agency: {agency.agency_code}")
+
+    user1_id = uuid.UUID("660e8400-e29b-41d4-a716-446655440000")
+    if not db_session.get(User, user1_id):
+        UserBuilder(user1_id, db_session, "AR User - Award Recommendation User").with_oauth_login(
+            "ar_rec_user1"
+        ).with_api_key("ar_rec_user1_key").with_jwt_auth().with_agency(
+            agency, roles=[AWARD_RECOMMENDATION_USER]
+        ).build()
+        logger.info("Created user: ar_rec_user1 (Award Recommendation User)")
```
chouinar: You don't need to check if it exists, the builder handles that, although in a slightly roundabout way.

It builds a user (which makes the object, but doesn't put it in the DB) and then merges it into the DB which works like an upsert in this case.

You can trim out all the checks and just rely solely on the builder to handle whether the user already exists.

Issue #9404, opened 2026-04-03
SOAP/Proxy: Validate GetSubmissionListExpanded


jakobpederson:
### Summary

Expand the validation tests to validate the response from `GetSubmissionListExpanded`.

Example chunk:
```
--uuid:12737dea-50a2-4766-bceb-5a4df9079570
Content-Type: application/xop+xml; charset=UTF-8; type="text/xml"
Content-Transfer-Encoding: binary
Content-ID: 
<root.message@cxf.apache.org>
    <soap:Envelope xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/">
        <soap:Body>
            <ns2:GetSubmissionListExpandedResponse xmlns:ns12="http://schemas.xmlsoap.org/wsdl/soap/" xmlns:ns11="http://schemas.xmlsoap.org/wsdl/" xmlns:ns10="http://apply.grants.gov/system/GrantsFundingSynopsis-V2.0" xmlns:ns9="http://apply.grants.gov/system/AgencyUpdateApplicationInfo-V1.0" xmlns:ns8="http://apply.grants.gov/system/GrantsForecastSynopsis-V1.0" xmlns:ns7="http://apply.grants.gov/system/AgencyManagePackage-V1.0" xmlns:ns6="http://apply.grants.gov/system/GrantsPackage-V1.0" xmlns:ns5="http://apply.grants.gov/system/GrantsOpportunity-V1.0" xmlns:ns4="http://apply.grants.gov/system/GrantsRelatedDocument-V1.0" xmlns:ns3="http://apply.grants.gov/system/GrantsTemplate-V1.0" xmlns:ns2="http://apply.grants.gov/services/AgencyWebServices-V2.0" xmlns="http://apply.grants.gov/system/GrantsCommonElements-V1.0">
                <ns2:Success>true</ns2:Success>
                <ns2:AvailableApplicationNumber>781</ns2:AvailableApplicationNumber>
                <ns2:SubmissionInfo>
```

## Important Note
From what I can tell the official response coming back from Grants is not actually valid by the wsdl. If I'm reading it correctly the response says that if there's no namespace use "http://apply.grants.gov/system/GrantsCommonElements-V1.0" while the wsdl says if there is no namespace use "http://apply.grants.gov/services/AgencyWebServices-V2.0"'.

Currently I just swap them out in the response and it validates against the wsdl. I'm not sure if that's the correct action or if we need to update our Simpler response to match the wsdl.

[...]

PR #9410, opened 2026-04-03
[Issue #9404] Add GetSubmissionListExpanded to SOAP validation

jakobpederson:
## Summary

Fixes / Work for #9404 

## Changes proposed

Added validation method to validate GetSubmissionListExpanded.

## Context for reviewers

This script is expanded to validate both the GetApplicationZip endpoint and now the GetSubmissionListExpanded.

## Validation steps

- [ ] update local `soap-api.env`
- [ ] run `make validate-simpler-endpoints`

### DIFF
--- a/api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
+++ b/api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
@@ -59,12 +59,12 @@ class ValidateSoapContext:
 def get_response(soap_context: ValidateSoapContext, request_operation: str) -> requests.Response:
     cert = _config.cert_data
     encoded = quote(cert, safe="")
-    HEADERS.update({"X-Amzn-Mtls-Clientcert": encoded})
+    headers = {"X-Amzn-Mtls-Clientcert": encoded, **HEADERS}
     data = REQUEST_BODY[request_operation]
     return requests.post(
         _config.soap_uri,
         data=data,
-        headers=HEADERS,
+        headers=headers,
         cert=(soap_context.cert, soap_context.key),
         timeout=10,
     )
@@ -118,7 +118,7 @@ def get_grantors_get_submission_list_expanded_request_body() -> bytes:
 
 
 REQUEST_BODY = {
-    "GetSubmissionListRequest": get_grantors_get_submission_list_expanded_request_body(),
+    "GetSubmissionListExpandedRequest": get_grantors_get_submission_list_expanded_request_body(),
     "GetApplicationZipRequest": get_grantors_get_application_zip_request_body(),
 }
 
@@ -169,6 +169,7 @@ def validate_response_xml(
         raise Exception("Envelope not found")
     xml_tree = etree.fromstring(simpler_match.group(0))
     ns = {
+        None: "http://apply.grants.gov/system/GrantsCommonElements-V1.0",
         "xop": "http://www.w3.org/2004/08/xop/include",
         "soap": "http://schemas.xmlsoap.org/soap/envelope/",
         "ns2": "http://apply.grants.gov/services/AgencyWebServices-V2.0",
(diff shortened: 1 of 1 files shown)


api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
```diff
@@ -72,19 +89,40 @@ def get_temp_files(stack: ExitStack) -> dict:
...
+        <soapenv:Body>
+        <agen:GetSubmissionListExpandedRequest>
+        </agen:GetSubmissionListExpandedRequest>
+        </soapenv:Body>
+        </soapenv:Envelope>
+    """.encode()
+
+
+REQUEST_BODY = {
+    "GetSubmissionListRequest": get_grantors_get_submission_list_expanded_request_body(),
```
kkrug: This doesn't change anything functionally but I think we should keep naming consistent. 
```suggestion
    "GetSubmissionListExpandedRequest": get_grantors_get_submission_list_expanded_request_body(),
```

api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
```diff
@@ -167,16 +193,40 @@ def validate_grantors_get_application_zip_request(soap_context: ValidateSoapCont
...
+    resp = get_response(soap_context, "GetApplicationZipRequest")
+    assert resp.status_code == 200
+    validate_response_xml(resp.content, "GetApplicationZipResponse", soap_context)
+    print("Validation: GetApplicationZip is validated")
+
+
+def validate_grantors_get_submission_list_expanded_request(
+    soap_context: ValidateSoapContext,
+) -> None:
+    resp = get_response(soap_context, "GetSubmissionListRequest")
```
kkrug: Accounting for the name change I mentioned above.
```suggestion
    resp = get_response(soap_context, "GetSubmissionListExpandedRequest")
```

api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
```diff
@@ -56,9 +56,26 @@ def __post_init__(self) -> None:
         self.cert, self.key = get_credentials(self.stack)
 
 
+def get_response(soap_context: ValidateSoapContext, request_operation: str) -> requests.Response:
+    cert = _config.cert_data
+    encoded = quote(cert, safe="")
+    HEADERS.update({"X-Amzn-Mtls-Clientcert": encoded})
```
kkrug: If we make this change, we're not mutating the global `HEADERS` dict on every call, which can prevent issues in the future if validations are ever run in parallel. 
```suggestion
    headers = {**HEADERS, "X-Amzn-Mtls-Clientcert": encoded}
```

api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
```diff
@@ -167,16 +193,40 @@ def validate_grantors_get_application_zip_request(soap_context: ValidateSoapCont
...
         handler.remove(xop_include)
         handler.text = None
-
     # This step is defensive in case that extracting the GetApplicationZipResponse breaks namespace connections
     etree.cleanup_namespaces(operation_element)
-
     temp_files = get_temp_files(soap_context.stack)
     schema_validator = build_schema_validator(
-        temp_files["AgencyWebServices-V2.0"], "GetApplicationZipResponse"
+        temp_files["AgencyWebServices-V2.0.wsdl"], f"{operation_response_name}"
```
kkrug: You don't need the f-string here.

api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
jakobpederson: Good catch, I wouldn't want it constantly mutating.

api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
jakobpederson: Just so you know I LOVE f-strings.

PR #9411, opened 2026-04-03
[Issue #9426] Clean up custom `.usa-nav__primary` styles

### LINKED ISSUE #9426: Clean up primary nav styles
### Summary

There are a bunch of custom styles that we no longer need (from when we implemented the user menu as a custom dropdown. Recent changes to the nav colors revealed cleanup needed. 

### Acceptance criteria

- [ ] Unused styles have been removed so there's a single custom code block for `.usa-nav__primary` styles

andycochran:
## Summary

Fixes #9426

## Changes proposed

This is a follow up to #9395 and [a comment](https://github.com/HHS/simpler-grants-gov/pull/9395#issuecomment-4183415152) asking why we have multiple CSS blocks for custom `.usa-nav__primary` styles, and [a comment](https://github.com/HHS/simpler-grants-gov/pull/9395#discussion_r3032829862) asking if the CSS selectors for the a11y fix need to be so complex. 

These changes… 

- Remove all the custom styles that are no longer needed from when we had implemented the user menu as a custom dropdown (the user menu is now just part of the default USWDS nav)
- Use the `.padding-bottom-05` USWDS utility class in `Header.tsx` instead of the custom styles for customizing the padding  in`.usa-header` 
  - We want the bottom border of current/hovered menu items to be flush with the edge of the header, but a recent USWDS upgrade changed the padding in the menu. We'd fixed it with a custom style but an existing utility class works. 
- Simplify the selector fixing the a11y bug 
  - We don't need both `> a` and `> a:not(.usa-button):not(.usa-current)` 
  - `> a:not(.usa-current)` will suffice 

## Context for reviewers

Nothing should change visually. This is only a cleanup of of styles. 

## Validation steps

- Load the app
- Use the nav at various screen sizes
- Make sure there are no visual diffs

### DIFF
--- a/frontend/src/styles/_uswds-theme-custom-styles.scss
+++ b/frontend/src/styles/_uswds-theme-custom-styles.scss
@@ -392,9 +392,19 @@ h6 {
   }
 }
 
-// a11y bug: https://github.com/HHS/simpler-grants-gov/issues/3423
-// USWDS to implement theme settings: https://github.com/uswds/uswds/issues/6176
+// Primay Nav Menu
+
 .usa-nav__primary {
+  // Match padding of non-dropdown nav links (Home, Search, Sign In) to dropdown items on mobile (12px)
+  @include at-media-max($theme-header-min-width) {
+    .usa-nav__primary-item > .usa-nav__link {
+      padding-top: 0.75rem;
+      padding-bottom: 0.75rem;
+    }
+  }
+
+  // a11y bug: https://github.com/HHS/simpler-grants-gov/issues/3423
+  // USWDS to implement theme settings: https://github.com/uswds/uswds/issues/6176
   // $nav-link-color hardcoded to "base-dark" (which we theme to "gray-warm-50")
   // &:hover is hardcoded to "primary" (which we theme to "mint-30")
   // We need a magic number of 50+ to pass WCAG 2.0 AA contrast (70+ for AAA).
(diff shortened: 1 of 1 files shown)


frontend/src/components/Header.tsx
```diff
@@ -307,6 +307,7 @@ const NavLinks = ({
       items={navItems}
       mobileExpanded={mobileExpanded}
       onToggleMobileNav={onToggleMobileNav}
+      className="padding-bottom-05"
```
andycochran, line 310: This could've been `.desktop:padding-bottom-05` to exactly match what the custom style was doing visually. But the slightly smaller bottom padding on the mobile nav menu does not cause any layout issues.

frontend/src/styles/_uswds-theme-custom-styles.scss
```diff
@@ -392,112 +392,19 @@ h6 {
...
-}
-
-// revert to behavior before upgrade https://github.com/uswds/uswds/commit/2785b67e58ae19f161101f4859afff0e9664fbda
-.usa-header--basic {
-  @include at-media($theme-header-min-width) {
-    .usa-nav {
-      padding: 0 0 units(0.5) units(1);
-    }
-  }
-}
```
andycochran, line 496: This was removed in favor of `.padding-bottom-05` on the menu

Issue #9412, opened 2026-04-03
SOAP/Proxy: Validate ConfirmApplicationDelivery Endpoint


jakobpederson:
### Summary

Add the ConfirmApplicationDelivery to the SOAP validation script.

Ideally it would be as simple as:
```
def validate_confirm_application_delivery_request(soap_context: ValidateSoapContext) -> None:
    resp = get_response(soap_context, "ConfirmApplicationDeliveryRequest")
    assert resp.status_code == 200
    validate_response_xml(resp.content, "ConfirmApplicationDeliveryResponse", soap_context)
    print("Validation: ConfirmApplicationDelivery is validated")
```

and:
```
VALIDATIONS = [
    validate_grantors_get_application_zip_request,
    validate_grantors_get_submission_list_expanded_request,
    validate_confirm_application_delivery_request
]
```

### Acceptance criteria

- [ ] add soap validation for ConfirmAppliationDelivery
- [ ] can successfully validate endpoint

PR #9422, opened 2026-04-06
[Issue #9412] Add validation for ConfirmApplicationDelivery

jakobpederson:
## Summary

Fixes / Work for #9412  

## Changes proposed

Add validation method for `ConfirmApplicationDelivery` endpoint.

## Context for reviewers

This is one of four methods we're intending to do schema validation on.

## Validation steps

- [ ] fill out `soap-api.env` data
- [ ] run `make validate-simpler-endpoints`


api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
kkrug, line 239: Just want to confirm that we don't need to use `clean_xml_namespaces` for this one.

jakobpederson: Yes, this one is the exception because it doesn't have any un-prefixed elements. So there's no "wrong" defaulting to the wrong url.

Issue #9373, opened 2026-03-31
Clean up and standardize fetcher patterns


doug-s-nava:
### Summary

* deduplicate apiKeyFetcher.fetchApiKeys and handleListApiKeys
* more TK

### Acceptance criteria

- [ ] apiKeyFetchers are deduplicated
- [ ] error is either thrown or not thrown prior to fetch if token is not present for authorized requests

PR #9431, opened 2026-04-07
[Feature] X-SGG-Token Header Refactor

doug-s-nava:
This change:

* introduces functionality to automatically add the necessary X-SGG-Token header to all API requests that require authentication
* applies that functionality to application related API requests

Note that this change does not update all authenticated fetch calls, will complete this work in a follow up PR. All existing calls should still work using the old manual system until they are translated over to the new system.

As part of this work I produced an audit of all authenticated calls, which can be found here https://docs.google.com/spreadsheets/d/1T23xaq6AMl7qi6OErAzGFRHXs2TjGe-vVI5S9ImdtSs/edit?usp=sharing

A follow up ticket has been created for cleaning up some unrelated standardization issues discovered when performing the audit https://github.com/HHS/simpler-grants-gov/issues/9373

Fetcher helpers file cleaned up a bit so that files that just want to import the types it exported don't opt in to session related stuff.

## Summary

Fixes / Work for #6876

### DIFF
--- a/frontend/src/app/[locale]/(base)/opportunities/page.tsx
+++ b/frontend/src/app/[locale]/(base)/opportunities/page.tsx
@@ -363,7 +363,6 @@ async function OpportunitiesListPage(props: OpportunitiesListProps) {
   );
   try {
     userPrivilegeResult = await checkRequiredPrivileges(
-      userSession.token,
       userSession.user_id,
       userPrivilegeDef,
     );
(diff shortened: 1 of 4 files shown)


myduong-navapbc:
## Tested locally

Navigated

- All available links on application
- signed in as
  - no org related user
  - org associated user
  - un-signed in
- created application
- submitted application
- created api key
- shared / unshared opportunity
- Search
  - saved search
  - saved query
- Account
  - updated name


Everything works without issue,
Good job

Issue #8988, opened 2026-03-11
[FE] Add subscribe/unsubscribe to user notification preferences page


ErinPattisonNava:
### Summary

In this ticket we're going to add the behaviour to the checkboxes were added in previous tickets.

Figma: https://www.figma.com/design/bQUTJCMj0JYbAAnJQ4p4tX/Grantee-Designs-%7C-Quad-5-%7C-Simpler-Grants-2026?node-id=3541-53498&t=qMIPFfwbFStH3SaH-1

### Acceptance criteria

- [ ] Display the correct statuses for each of the notification checkboxes
   - [ ] Call the [{user_id}/saved-opportunities/notifications endpoint](https://github.com/HHS/simpler-grants-gov/issues/8995) to get the current state of notifications
       - [ ] Have the checkbox in their personal preferences update to the correct state, where `email_enabled` === true means the checkbox should be checked
       - [ ] For each of their organizations, have the checkbox update to the correct state, where `email_enabled` === true means the checkbox should be checked
- [ ] If the user checks/unchecks the checkmark under `Manage your preferences`
   - [ ] Post the API `{user_id}/saved-opportunities/notifications` with these parameters
      - [ ] `organization_id` is empty/null
      - [ ] `email_enabled` is the opposite of the current state of the checkbox 
   - [ ] On successful return from the API call the checkbox should change
   - [ ] On failure return from the api the error message
- [ ] If the user checks/unchecks a checkmark under `Organization preferences`
   - [ ] Post the API `{user_id}/saved-opportunities/notifications` with these parameters
      - [ ] `organization_id` to match the section's organization_id
      - [ ] `email_enabled` is the opposite of the current state of the checkbox 
   - [ ] On successful return from the API call the checkbox should change
- [ ] Full unittests
- [ ] If any of these API calls fail, show the error banner at the top
- [ ] Design Review before PR is submitted
- [ ] UAT

ErinPattisonNava:
Need to do loading state follow on ticket

eric-serrato95:
@myduong-navapbc UAT ✅ closing ticket

Issue #8987, opened 2026-03-11
[FE] Build user notification preferences page


ErinPattisonNava:
### Summary

We want to build a new notifications page for the application.  Instead of incurring the cost of a feature flag, we're going to build the page first with no links anywhere on the website, so it must be accessed by the specific URL (that way it isn't discoverable by our users).  

This first ticket will create the skeleton of the page, we'll add the functionality to the checkboxes in a later [ticket](https://github.com/HHS/simpler-grants-gov/issues/8988). 

Figma: https://www.figma.com/design/bQUTJCMj0JYbAAnJQ4p4tX/Grantee-Designs-%7C-Quad-5-%7C-Simpler-Grants-2026?node-id=3541-53498&t=qMIPFfwbFStH3SaH-1

### Acceptance criteria
- [x] Basic page is created at `/notifications`
   - [x] Require users to be logged in to view
- [x] `Manage your preferences` section appears for all users. 
   - [x] Create Saved opportunities section with checkbox, but checkbox does nothing 
- [x] If the user is part of an organization, the `Organization preferences` appears
   - [x] Sections are listed for each organization they are part of 
   - [ ] Check boxes are present but they do nothing
- [x] If their list of organizations is not available, show the general error 
- [ ] Mobile view also implemented
- [ ] Check keyboard and voice over a11y
- [ ] Design Review before PR is submitted

PR #9441, opened 2026-04-07
[#8987] Notication preferences page

hayeshorus:
## Summary
Notifications preferences page creation

Work for #8987 

## Important
This is a mock page
It does not connect to the api yet

The link to the notifications page has been commented out — 
The Loading / Saving / Action states will be handled ->
in https://github.com/HHS/simpler-grants-gov/issues/8988 is complete

to test the Navigation Submenu, uncomment this block from the `header.tsx` file (_line 285_)

```
// <NavLink
//   href="/notifications"
//   key="notifications"
//   onClick={closeDropdownAndMobileNav}
//   text={t("notifications")}
// />,
```

## Changes proposed

- Add Notifications page content with:
  - user preference section
  - organization preference sections
- Implement loading and error states:
  - inline (row-level)
  - page-level (alert)
- Improve accessibility:
  - `aria-live` for dynamic updates
  - `aria-busy` for loading states
  - improved labeling for organization context
- Replace toggle switch with checkbox per team feedback
- Add subtle error animation (respects reduced motion)
- Add/clean unit tests for components and page behavior

## Validation steps

Screenshot 2026-04-13 at 3 53 32 PM

### No organization associated user flow
- log into the application using `test`
- navigate to /notification
  - via url
  - or by dropdown under `account`
    - must uncomment the code in `header.tsx` around line 285, `// <NavLink`
- [x] `VERIFY` the org section does not display 

--- 

Screenshot 2026-04-13 at 3 53 09 PM

### Organization associated user
- log into the application using `two_org_user`
- navigate to /notification
  - via url
  - or by dropdown under `account`
    - must uncomment the code in `header.tsx` around line 285, `// <NavLink`
- [x] `VERIFY` the org section does display user's org
- [x] `VERIFY` screen reader announces errors and loading

---

### Error state

Screenshot 2026-04-13 at 3 53 32 PM

### DIFF
--- a/frontend/src/components/notifications/NotificationsPageContent.test.tsx
+++ b/frontend/src/components/notifications/NotificationsPageContent.test.tsx
@@ -1,4 +1,13 @@
+import {
+  act,
+  fireEvent,
+  render,
+  screen,
+  waitFor,
+} from "@testing-library/react";
+
 import { NotificationPreferenceCardProps } from "./NotificationPreferenceCard";
+import { NotificationsPageContent } from "./NotificationsPageContent";
 import { OrganizationPreferenceSectionProps } from "./OrganizationPreferenceSection";
 
 type MockNotificationPreferenceCardProps = Pick<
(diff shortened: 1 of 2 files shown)


andycochran:
Can we discuss this new "Switch" design pattern at the next Design Crit? 

cc @doug-s-nava @shannonmcharg-nava @chelsea-nava

frontend/src/app/[locale]/(base)/notifications/layout.tsx
```diff
@@ -0,0 +1,25 @@
...
+  };
+  return meta;
+}
+
+export default function NotificationsLayout({ children }: LayoutProps) {
+  return (
+    <>
+      <AuthenticationGate>{children}</AuthenticationGate>
+    </>
+  );
```
Comment (freddieyebra) [inline frontend/src/app/[locale]/(base)/notifications/layout.tsx]: Could we simplify this to 
```suggestion
  return <AuthenticationGate>{children}</AuthenticationGate>;
```

shannonmcharg-nava:
This PR should not be merged without @andycochran 's approval.

frontend/src/components/notifications/NotificationPreferenceCard.tsx
```diff
@@ -0,0 +1,70 @@
...
+      aria-busy={isLoading}
+    >
+      <div className="notification-preference-card__field">
+        <SimplerSwitchField
+          id={checkboxId}
+          ariaLabel={ariaLabel}
+          label={label}
+          description={description}
+          checked={isChecked}
+          disabled={isDisabled || isLoading}
```
freddieyebra: should we disable the switch if there is an error?

frontend/src/components/notifications/NotificationPreferenceCard.tsx
```diff
@@ -0,0 +1,70 @@
...
+          onCheckedChange={onCheckedChange ?? (() => {})}
+          errorMessage={hasError ? " " : undefined}
```
freddieyebra: Are we passing empty string here if there is an error since we display errors elsewhere?

frontend/src/components/notifications/NotificationPreferenceCard.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+        <div
+          className="notification-preference-card__status"
+          aria-live="polite"
+        >
+          {isLoading ? (
+            <span className="notification-preference-card__loading-text">
+              Saving...
+            </span>
+          ) : null}
```
freddieyebra, line 98: Could we simplify this to
```suggestion
          {isLoading && (
            <span className="notification-preference-card__loading-text">
              Saving...
            </span>
          )}
```

frontend/src/components/notifications/NotificationsPageContent.tsx
```diff
@@ -0,0 +1,181 @@
...
+  async function handlePreferenceToggle(
+    preferenceKey: NotificationPreferenceKey,
+  ): Promise<void> {
+    setPageLevelErrorMessage(null);
+    setErrorPreferenceKey(null);
+    setSavingPreferenceKey(preferenceKey);
+
+    await new Promise((resolve) => {
+      window.setTimeout(resolve, 1200);
+    });
```
freddieyebra, line 110: Is this for demonstration/uat purposes? If so, can we add a comment specifying that.

frontend/src/components/notifications/OrganizationPreferenceSection.test.tsx
```diff
@@ -0,0 +1,138 @@
+import { fireEvent, render, screen } from "@testing-library/react";
+
+import { OrganizationPreferenceSection } from "./OrganizationPreferenceSection";
+
+type NotificationPreferenceCardProps = {
```
freddieyebra: Curious on why we are not re-using the type definition in the file with `NotificationPreferenceCard` component.

frontend/src/components/Header.tsx
```diff
@@ -281,6 +281,12 @@ const NavLinks = ({
               onClick={closeDropdownAndMobileNav}
               text={t("settings")}
             />,
+            <NavLink
+              href="/notifications"
+              key="notifications"
+              onClick={closeDropdownAndMobileNav}
+              text={t("notifications")}
+            />,
```
freddieyebra: We should comment this out since we only want this page to be available by url until the feature functionality is complete.

frontend/src/components/SimplerSwitch.tsx
```diff
@@ -0,0 +1,269 @@
...
+  - Retry/loading workflows
+  - Page-specific settings row layout
+  - Indeterminate or mixed states
+*/
+
+function joinClassNames(
+  ...classNames: Array<string | undefined | false | null>
+): string {
+  return classNames.filter(Boolean).join(" ");
+}
```
freddieyebra: this seems like a re-usable util. if we do not have a util for this already, can we move it into a shared util file?

frontend/src/components/notifications/NotificationPreferenceCard.tsx
```diff
@@ -0,0 +1,70 @@
...
+  checkboxId: string;
+  label: string;
+  description: string;
+  isChecked: boolean;
+  organizationHeadingId?: string;
+  ariaLabel?: string;
+  isDisabled?: boolean;
+  isLoading?: boolean;
+  hasError?: boolean;
+  onCheckedChange?: (checked: boolean) => void;
```
freddieyebra, line 16: Not a blocker, but it would be good to have the props related to checked consistent. For example

```javascript
   checked: bool;
   onCheckedChange?: (checked: boolean) => void;
```
or
```javascript
   isChecked: bool;
   onIsCheckedChange?: (checked: boolean) => void;
```

preferably the former since that is the name of the prop on the `SimplerSwitch` component.

frontend/src/components/SimplerSwitch.tsx
```diff
@@ -0,0 +1,269 @@
...
+  ariaLabelledBy?: string;
+  ariaDescribedBy?: string;
+  showStateText?: boolean;
+  onText?: string;
+  offText?: string;
+  hasError?: boolean;
+  className?: string;
+};
+
+export type SimplerSwitchFieldProps = Omit<SimplerSwitchProps, "hasError"> & {
```
freddieyebra: 👍

frontend/src/components/notifications/NotificationPreferenceCard.tsx
myduong-navapbc: Yes, for this, it is a blank- 
we will end up show the error under the description (attached a screenshot)

Screenshot 2026-04-13 at 2 02 33 PM

BUT, that said, that is for the next story when we actually implement the action to call the api- that way we can display the error message explictly.

TL;DR: 
Error handling will happen in this ticket https://github.com/HHS/simpler-grants-gov/issues/8988
this AC `On failure return from the api the error message`

frontend/src/components/notifications/NotificationPreferenceCard.tsx
myduong-navapbc: Good question. I don’t think we should disable the control in the error state.

If we disable it, the user has no way to retry the action without refreshing the page, which creates a dead-end interaction. In this case, the error is likely transient (network, service issue, etc.), so allowing the user to try again directly is important.

frontend/src/components/notifications/NotificationsPageContent.tsx
myduong-navapbc: Yes, this is for demo- i will add some comments to this.

frontend/src/components/notifications/OrganizationPreferenceSection.test.tsx
myduong-navapbc: thanks for calling it out:
i am removing this extra declaration to use the existing one.

myduong-navapbc:
> Can we discuss this new "Switch" design pattern at the next Design Crit?
> 
> cc @doug-s-nava @shannonmcharg-nava @chelsea-nava

@andycochran / @shannonmcharg-nava / @chelsea-nava 
I removed the Switch and brought back the Checkbox (FYI)

andycochran:
Sounds good, @myduong-navapbc. We talked about the switch at today's crit. There are likely some good cases for introducing this new pattern, but we'll first need to establish some guidance around when to use what. Checkboxes should suffice here for now.

frontend/src/components/notifications/NotificationsPageContent.test.tsx
```diff
@@ -0,0 +1,238 @@
...
+  act,
+  fireEvent,
+  render,
+  screen,
+  waitFor,
+} from "@testing-library/react";
+
+import { NotificationsPageContent } from "./NotificationsPageContent";
+
+type MockNotificationPreferenceCardProps = {
```
freddieyebra: This is missing required props that are defined on the NotificationPreferenceCardProps interface
```
  description: string;
  isChecked: boolean;
```

To prevent drift across prop types for unit tests and for consistency for these and other mocked component prop types, can we use the actual types that are defined and used in the non-test components? At minimum, we should should at least use something like 
```
Pick<NotificationPreferenceCardProps,  | "checkboxId" | "label" ...>
```

Happy to talk through this

frontend/src/components/notifications/NotificationsPageContent.test.tsx
```diff
@@ -0,0 +1,238 @@
...
+  checkboxId: string;
+  label: string;
+  isDisabled?: boolean;
+  isLoading?: boolean;
+  hasError?: boolean;
+  errorMessage?: string;
+  onCheckedChange?: (checked: boolean) => void;
+};
+
+jest.mock("./NotificationPreferenceCard", () => ({
```
freddieyebra, line 24: I may have looked past this on first review, but can we use the NotificationPreferenceCard component  instead of mocking it here? Same for the other unit tests following this pattern.

myduong-navapbc: I am mocking `NotificationPreferenceCard` here intentionally to keep the test focused on `NotificationsPageContent` behavior.

`NotificationPreferenceCard` already has its own unit tests, so using the real component here would couple this test to its internal implementation and result in duplicate coverage. By mocking it, we can isolate and verify:

- loading state handling
- error state handling
- interaction flows

without introducing additional complexity or brittleness from nested component behavior.

freddieyebra: Got it, i understand the context now. thanks for explaining.

myduong-navapbc:
> left some minor feedback.
> 
> During dev testing, i noticed when the new switch is loading state, the text "Saving..." is a bit off center. Also, is it intended that clicking the label of switches toggles the switch?
> 
> Image

The current 'saving..." text is a placeholder and will be replaced with a spinner in the implementation ticket (https://github.com/HHS/simpler-grants-gov/issues/8988)

frontend/src/components/notifications/NotificationsPageContent.test.tsx
myduong-navapbc: I love it! updated the Preference Card and Organization Section to Pick types

Issue #8573, opened 2026-02-18
Prototype FieldList Mandatory Fields and Error Messages


ErinPattisonNava:
### Summary

We want to do a basic prototype of of the VariableList widget to see what issues might occur.  In this ticket we're going to be looking at constraints, like mandatory subfields, and error messages both at the list level and the individual field level. 

See tech spec here: https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2192179202/Tech+Spec+Expandable+Form+Field+Groups

### Acceptance criteria

- [ ] Building off of our dummy form from our previous [ticket](https://github.com/HHS/simpler-grants-gov/issues/8555), add a new FieldList of a more complex type like address. 
- [ ] Mark mandatory fields correctly and confirm they are still recognized as mandatory.
- [ ] Allow a minimum required number of entries in the list.
   - [ ] Show the minimum number of entries  as blank when starting the form. 
   - [ ] The most important minimum is between 0 or 1 entries (e.g. if it is required to have any item in the list).
- [ ] It should also allow for a maximum number of entries. 
- [ ] Map rawErrors on to the correct fields inside each element of the list.
- [ ] Display general errors for the group as a whole.
- [ ] Confirm that deleting erroneous data removes the errors from the overall list. 
- [ ] Unittests where it makes sense

ErinPattisonNava:
Breaking up https://github.com/HHS/simpler-grants-gov/issues/7026 into smaller parts

PR #9445, opened 2026-04-07
[Issue #8573] FieldList Required Fields and Error Messages

myduong-navapbc:
## Summary

Work for #8573

This PR adds full `FieldList` (repeatable field) support with correct row-aware validation. It enables child field changes, maps validation errors to the correct row, preserves multiple row-specific warnings, and links summary messages directly to the affected inputs. 

## Changes proposed

- Added full `FieldList` support in the form system
- Implemented row-aware validation handling for `FieldList` child fields
- Updated validation pipeline to support **multiple warnings per field definition**
- Ensured inline errors render on the correct row only
- Enabled summary links to anchor to the exact errored input
- Improved validation deduplication to preserve row-specific warnings
- Added _FieldList-specific_ helpers for group and child error handling
- Updated getFieldConfig to properly wire `FieldList` data, schema, and errors
- Added group-level error rendering for array-level validation (e.g. empty list)

## Validation steps

- Sign in
- Type "test" in search
- Open "**Test Opportunity for FIELDLIST_DEMO_DO_NOT_SYNC 1.0"**

### Test FieldList behavior
- Leave all fields empty and click Save
  - Verify:
    - [ ] Summary shows validation errors
    - [ ] Links navigate to the correct inputs

  - Add multiple FieldList entries (e.g. 3 rows)
    - Fill First Name in row 1 only
    - Leave rows 2 and 3 empty
    - Click Save
    - Verify:
      - [ ] Errors appear only on rows 2 and 3 (not row 1)
      - [ ] Summary shows multiple entries (e.g. Entry 2, Entry 3)
      - [ ] Each link navigates to the correct row

  - Fix one invalid row (e.g. fill First Name in row 2)
    - Click Save
    - Verify:
      - [ ] Error for that row disappears
      - [ ] Remaining row still shows error

  - Add a new row after validation errors exist
  - Verify:
    - [ ] New row does not inherit errors incorrectly
    - [ ] Remove a row with errors

  - Verify:
    - [ ] Errors update correctly in both UI and summary

  - Submit with all fields valid
  - Verify:
    - [ ] No validation errors shown
    - [ ] Success message appears


frontend/src/components/applyForm/utils.ts
```diff
@@ -265,6 +278,39 @@ export const getFieldNameForHtml = ({
...
+function getHtmlFieldForWarning({
+  definition,
+  field,
+  schema,
+}: {
+  definition?: string;
+  field?: string;
+  schema?: SchemaField;
+}): string | undefined {
+  if (definition) {
```
myduong-navapbc: Builds a row-aware html field id for FieldList child inputs when possible, otherwise falls back to standard field id generation

frontend/src/utils/applyForm/getFieldConfig.ts
```diff
@@ -436,6 +467,16 @@ const getFieldListConfig = ({
         throw new Error("fieldList children must be field nodes");
       }
```
myduong-navapbc, line 470: These checks ensure each `FieldList` child maps to a single, valid schema definition so the form can correctly bind data and validation.

frontend/src/utils/applyForm/getFieldConfig.ts
```diff
@@ -94,6 +94,37 @@ export function buildFieldListBaseId({
   return `${fieldListName}[${FIELD_LIST_INDEX_TOKEN}]--${finalFieldKey}`;
 }
```
myduong-navapbc, line 96: Extracts required field names from the `FieldList` item schema so each row can correctly mark required child inputs

frontend/src/components/applyForm/utils.ts
```diff
@@ -103,13 +98,20 @@ const validationWarningOverrides = (
     .replace("[] should be non-empty", `${formattedTitle} is required`)
     .replace("is a required property", "is required");
 };
-
-const findValidationError = (
+const formatValidationWarning = (
```
myduong-navapbc: Renamed to reflect that the function formats and returns all matching validation warnings, not just a single error

frontend/src/components/applyForm/widgets/fieldListHelpers.ts
```diff
@@ -0,0 +1,93 @@
+import { FormattedFormValidationWarning } from "src/components/applyForm/types";
```
doug-s-nava: can we move this out of the widgets directory? I know there's at least one other util file in there, but would rather not perpetuate that idea. This probably belongs in utils/applyForm?

frontend/src/utils/applyForm/fieldListHelpers.ts
```diff
@@ -0,0 +1,93 @@
...
+          }
+
+          // Checks if the warning field includes a row index (e.g. [1])
+          const hasIndexedFieldPath = /\[\d+\]/.test(warning.field);
+
+          if (hasIndexedFieldPath) {
+            return false;
+          }
+
+          return warning.definition === childDefinition;
```
doug-s-nava: what is the use case where the path doesn't have an index and we get to this point? If we're only running this on field list children I'd imagine they'd all have an index

frontend/src/components/applyForm/widgets/FieldListWidget.tsx
```diff
@@ -287,6 +338,33 @@ function FieldListWidget(widgetProps: FieldListWidgetProps) {
 
   const isInteractionDisabled = Boolean(disabled || readOnly || isFormLocked);
 
+  /**
```
doug-s-nava, line 352: does this have to do with the error handling in some way, or is this a fix for standardizing the entry values?

frontend/src/components/applyForm/ApplyFormMessage.tsx
```diff
@@ -3,6 +3,48 @@ import { Alert } from "@trussworks/react-uswds";
...
+ *
+ * becomes:
+ *   "First Name is required (Contact People, Entry 2)"
+ */
+
+export const getWarningLinkText = (
+  warning: FormattedFormValidationWarning,
+): string => {
+  const baseText = warning.formatted ?? warning.message;
+  const rowMatch = warning.htmlField?.match(/\[(\d+)\]--/);
```
doug-s-nava: didn't we decide to use `entry` instead of row? https://github.com/HHS/simpler-grants-gov/pull/9046#discussion_r2940652611

frontend/src/components/applyForm/ApplyFormMessage.tsx
```diff
@@ -21,6 +63,31 @@ export const ApplyFormMessage = ({
     ),
     p: (content) => <p>{content}</p>,
   });
+
+  /**
+   * Deduplicate validation warnings before rendering in the summary.
+   *
+   * Some warnings (particularly for FieldList child fields) can be emitted
+   * multiple times during validation tree construction. This ensures each
```
doug-s-nava: curious why this is, and if it's worth fixing.

frontend/src/components/applyForm/ApplyFormMessage.tsx
```diff
@@ -44,13 +111,21 @@ export const ApplyFormMessage = ({
...
               <a href={`#${warning.field}`}>{warning.message}</a>
             ) : (
-              <a href={`#${warning.htmlField || ""}`}>{warning.formatted}</a>
+              <a href={`#${warning.htmlField || ""}`}>
+                {getWarningLinkText(warning)}
+              </a>
+            );
+            return (
+              <li
+                key={`${warning.htmlField ?? warning.field ?? warning.definition}-${warning.message}`}
```
doug-s-nava: huge nit, since this is the same as the template constructed in line 84, maybe it deserves to be split out to it's own function

frontend/src/components/applyForm/utils.ts
doug-s-nava: this looks good, but I got a bit hung up for a minute thinking that this logic meant that any non-field list widget would not be passing in a definition, when the actual test is whether the definition matches the field list regexp. 

very nitpicky, but might be more readable if you happy pathed it with something like
```
const match = definition ? definition.match(/^\/properties\/([^/]+)\/items\/properties\/([^/]+)$/) : null

if (!match) return getFieldNameForHtml({
    definition,
    schema,
  });
```

frontend/src/components/applyForm/utils.ts
```diff
@@ -265,6 +305,85 @@ export const getFieldNameForHtml = ({
...
+    }
+  }
+
+  return getFieldNameForHtml({
+    definition,
+    schema,
+  });
+}
+
+// Returns validation messages for a specific FieldList group,
```
doug-s-nava: is this the right comment? This seems to be dealing with field labels rather than errors

frontend/src/components/applyForm/utils.ts
```diff
@@ -174,20 +209,23 @@ export const buildWarningTree = (
...
             node.schema,
             formSchema,
+            resolvedRootUiSchema,
           );
-          if (error) {
-            return errors.concat([error]);
+          if (matchingWarnings.length > 0) {
+            return errors.concat(matchingWarnings);
           }
+          return errors;
```
doug-s-nava: I think this is fine, but if I'm reading things correctly, it's not strictly necessary, since it would return errors on line 230 anyway?

frontend/src/components/applyForm/utils.ts
doug-s-nava: I find myself wishing that these helper functions had their own tests - they're certainly complex enough to the deserve them. Not a blocker, but maybe something to consider before this goes into wider use

frontend/src/utils/applyForm/getFieldConfig.ts
```diff
@@ -94,6 +94,37 @@ export function buildFieldListBaseId({
...
 
+const getFieldListRequiredFields = ({
+  formSchema,
+  fieldListName,
+}: {
+  formSchema: RJSFSchema;
+  fieldListName: string;
+}): string[] => {
+  const fieldListSchema = formSchema.properties?.[fieldListName] as
```
doug-s-nava, line 105: seems like this won't work for a field list that's nested at all. Is it a safe assumption that these won't be nested?

frontend/src/utils/applyForm/fieldListHelpers.ts
myduong-navapbc: When the FieldList starts with a blank default row and validation fails at the array/group level, the warning can still reach the child mapping step without an index.

Example:

- `contact_people_test` has minItems: 1
- UI renders a blank row 0 by default
- backend/frontend warning comes through as:
  - `field: $.contact_people_test`
- but during formatting/mapping, that warning can still be associated to the child definition:
  - `/properties/contact_people_test/items/properties/first_name`

So by the time `getFieldListChildErrors` runs, we can have:

- child definition present
- but no indexed field path

TL:DR

- the blank initial row would not show the inline child error state
- only row-aware warnings from explicitly indexed rows would work

frontend/src/components/applyForm/ApplyFormMessage.tsx
myduong-navapbc: This is a good one to ask about and to call out - this is happening because of how the warning tree is constructed.

At a high level:

- We start with a flat list of validation warnings from the backend
- In buildWarningTree, we traverse the UI schema and attempt to match warnings at multiple levels:
  - direct matches (field → warning)
  - nested matches (parent → child mapping, especially for FieldList)

In some cases, the same underlying warning can be picked up more than once - for example:

- once as a direct match
- once again through nested matching logic

By the time we reach ApplyFormMessage, we can end up with duplicate entries like:

- "First Name is required"
- "First Name is required"

The dedupe here ensures the summary only shows each message once.

——————

I do think it would be better to address this upstream in `buildWarningTree` so we don’t generate duplicates in the first place. That said, that would involve tightening the matching logic and could affect other parts of the validation pipeline, so I kept the fix localized here for now.


Happy to follow up with a ticket to clean this up at the source.

frontend/src/components/applyForm/utils.ts
myduong-navapbc: very valid! will write test for them

frontend/src/components/applyForm/utils.ts
myduong-navapbc: eagle-eye doug

frontend/src/utils/applyForm/getFieldConfig.ts
myduong-navapbc: correct! this helper currently assumes `FieldLists` are top-level and looks them up directly from `formSchema.properties`. That matches the current FieldList usage in this PR, but it would not work for nested FieldLists. If nested FieldLists become a supported pattern, we should update this to resolve from a schema definition/path rather than from the root field name.

I'll drop a comment in here to state the assumption

frontend/src/components/applyForm/widgets/fieldListHelpers.ts
myduong-navapbc: I agree we shouldn’t keep adding utils into the widgets directory BUT - Actually, i wanted to  propose that we treat applyForm as more of a feature/module rather than scattering pieces across different directories.

Since the form system is fairly self-contained (components, utils, types, helpers), it might make sense to group everything under a single feature folder instead of splitting by type.

Something like:

```
src/libs/apply-form/
  components/
    ApplyForm.tsx
    ApplyFormMessage.tsx
    FieldListWidget.tsx
    FieldListEntry.tsx
  utils/
    buildWarningTree.ts
    fieldListHelpers.ts
    validation.ts
  types/
    formTypes.ts
  config/
    getFieldConfig.ts
```

That keeps all form-related logic co-located and makes it easier to reason about the system as a whole, instead of having form-specific utils live alongside unrelated widgets or global utilities.

For this PR, I’m happy to move this util out of widgets/ into a more appropriate place - either a local applyForm/utils or wherever we think it best fits - and we can follow up with a larger cleanup if we want to standardize the structure.

frontend/src/components/applyForm/widgets/FieldListWidget.tsx
myduong-navapbc: This is not tied to validation. it's for state consistency. In short standardizing / normalizing values

Child widgets can emit values as unknown, so we normalize them before updating the row data. The rest of the function ensures the update is scoped to the correct row and maintains immutability of the rows array.

frontend/src/components/applyForm/ApplyFormMessage.tsx
myduong-navapbc: Ah, I assumed it was Entry for Ui purposes.

I kind of see it as rows of entries. I can update the var names

frontend/src/components/applyForm/ApplyFormMessage.tsx
myduong-navapbc: I see it now and separating it into a dry compute function

frontend/src/components/applyForm/utils.ts
myduong-navapbc: I like it! updated the code to be a little clearer

frontend/src/components/applyForm/utils.ts
myduong-navapbc: Its a tad stale; updated to match the updated function

frontend/src/components/applyForm/ApplyFormMessage.tsx
myduong-navapbc: Here is the ticket: https://github.com/HHS/simpler-grants-gov/issues/9495

myduong-navapbc:
> manual testing results:
> 
> * click save on blank form then add an entry
>   
>   * expected: first name field on new entry has no errors
>   * actual: first name field on new entry has errors - note that this seems to maybe only be an issue on first save?
> * from state with two entries - entry 1 is valid, entry 2 is missing first name -> Click save -> remove entry one
>   
>   * expected: entry 2 error is still visible
>   * actual: entry 2 error is no longer visible

I fixed the fallback path to apply entryIndex === 0

updated*:

I've updated the current code to keep a local state.

frontend/src/utils/applyForm/rebaseFieldListWarningsAfterDelete.ts
```diff
@@ -0,0 +1,73 @@
...
+ * the remaining entries shift indices. This helper:
+ *
+ * - removes warnings for the deleted entry
+ * - shifts warnings for subsequent entries down by one index
+ * - leaves earlier entries unchanged
+ *
+ * This keeps the UI aligned with the current entry state until the next
+ * backend validation pass.
+ */
+export const rebaseFieldListWarningsAfterDelete = ({
```
doug-s-nava, line 17: gnarly

frontend/src/utils/applyForm/rebaseFieldListWarningsAfterDelete.ts
```diff
@@ -0,0 +1,73 @@
...
+  rawErrors?: FormattedFormValidationWarning[] | null;
+  fieldListPath: string;
+  deletedEntryIndex: number;
+}): FormattedFormValidationWarning[] | null => {
+  if (!rawErrors?.length) {
+    return rawErrors ?? null;
+  }
+
+  const escapedFieldListPath = fieldListPath
+    .replace(/^\$\./, "")
```
doug-s-nava, line 31: a few of these regexes are used in some other places as well. Especially since regexes are hard to understand at a quick glance, maybe split them out into their own named vars?

frontend/src/utils/applyForm/rebaseWarnings.ts
```diff
@@ -0,0 +1,70 @@
+import { FormattedFormValidationWarning } from "src/types/applyForm/types";
```
doug-s-nava: is the same file as `rebaseFieldListWarnings`?

frontend/src/components/applyForm/ApplyForm.tsx
```diff
@@ -103,20 +119,68 @@ const ApplyForm = ({
...
   );
+
   const navFields = useMemo(() => getFieldsForNav(uiSchema), [uiSchema]);
+
+  const displayValidationWarnings = useMemo(() => {
+    if (!validationWarnings) {
+      return null;
+    }
+
+    return Object.entries(deletedEntryIndexesByFieldListPath).reduce<
```
doug-s-nava, line 147: somewhat complex here, maybe an example in a comment would help?

doug-s-nava (approved):
✅ ✅ ✅ ✅ 

good job this is tough stuff!

ErinPattisonNava (approved):
Approving removing the temp form, have not reviewed the rest of the PR. 

It appears rebaseWarnings.ts and the associated test file were removed because the enclosed function was refactored into a different file.  

Did basic regression testing on cd511 and sf424, looks good.

Issue #9429, opened 2026-04-06
SOAP/Proxy: Add prefix/namespaces to UpdateApplicationInfo


jakobpederson:
### Summary

It appears our response is missing `ns9: SaveAgencyNoteResult` and `ns2: Success` and `ns9:AssignAgencyTrackingNumberResult` in the response. The current response is lacking `ns2` and `ns9`. This card is to add those in by modifying the response schema. The two nested successes are `ns9`.

<img width="1073" height="860" alt="Image" src="https://github.com/user-attachments/assets/1e5eeb09-b046-4756-a5ae-e2b8544f9efd" />
<img width="1072" height="875" alt="Image" src="https://github.com/user-attachments/assets/762a8855-0a42-4cd2-85c2-c435afd9ff12" />

The fix turned out to be update the alises on the schema object.

### Acceptance criteria

- [ ] updates tests
- [ ] tests passing

PR #9446, opened 2026-04-07
[Issue #9249] Fix missing UpdateApplicationInfo SOAP prefixes

jakobpederson:
## Summary

Fixes / Work for #9429  

## Changes proposed

Added prefixes for the `UpdateApplicationInfo` response schema.

## Context for reviewers

We were missing some prefixes needed for validation, so we're adding them here and updating some tests.

## Validation steps

- [ ] validation tests pass
- [ ] unit tests pass

### DIFF
--- a/api/tests/src/legacy_soap_api/grantors/schemas/test_legacy_soap_grantor_get_application_zip_schema.py
+++ b/api/tests/src/legacy_soap_api/grantors/schemas/test_legacy_soap_grantor_get_application_zip_schema.py
@@ -4,11 +4,20 @@ from unittest.mock import patch
 
 import pytest
 
+from src.constants.lookup_constants import ApplicationStatus, Privilege
 from src.legacy_soap_api.grantors import schemas as grantors_schemas
+from src.legacy_soap_api.legacy_soap_api_auth import SOAPAuth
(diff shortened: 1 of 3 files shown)


api/tests/src/legacy_soap_api/grantors/schemas/test_legacy_soap_grantor_get_application_zip_schema.py
```diff
@@ -4,42 +4,65 @@
...
+    CompetitionFactory,
+    OpportunityFactory,
+)
 
 GRANTS_GOV_TRACKING_NUMBER = "GRANT80000000"
 CID_UUID = "aaaaaaaa-1111-2222-3333-bbbbbbbbbbbb"
 BOUNDARY_UUID = "cccccccc-1111-2222-3333-dddddddddddd"
 
 
 class TestLegacySoapGrantorGetApplicationZipSchema:
```
jakobpederson, line 27: Updated this seemingly unrelated test because it started failing when I started messing around with actual certificates while trying to fix the prefix issue. I think the hardcoded submission legacy tracking number was causing the issue so I added the factories to make sure that wouldn't happen again going forward if anyone else added more submissions for testing.

If it's too far out of scope I'll pull it for it's own pr.

Issue #9397, opened 2026-04-02
Opportunity List Page - Enhancements


ssatia-SGM:
### Summary

Updates for the Opportunity List page

Show draft opportunities
Show/enable the edit button for draft opportunities only
When user click on Edit link, redirect to Edit Opportunity page
On click of the Edit Opportunity button, verify that the user has the edit_opportunity privilege, show an error message as needed
On click of the Create Opportunity button, verify that the user has the create_opportunity privilege, show an error message as needed

NOTE: make a copy of the checkRequiredPrivileges() method from AuthenticationGate.tsx instead of calling it directly. This will call the can_access endpoint on the backend to get the user's privileges. 

### Acceptance criteria

 Updates for the Opportunity List page:
- [x] Show draft opportunities
- [x] Show/enable the edit button for draft opportunities only
- [x] If the opportunity is Published, do not show/allow any action buttons
- [x] When user click on Edit link, redirect to Edit Opportunity page (note: the Edit Opportunity page may not be completed before these enhancements are completed.)
- [x] If the user does not have the edit_opportunity privilege, do not enable the Edit Opportunity link
- [x] If the user does not have the create_opportunity privilege, do not enable/show the Create Opportunity button.
- [x] If the agencyId provided in the url parameter is invalid OR the user does not belong to that agency, display an error but also show the Agency List (with no selection) so that the user may select a different agency.

Dao-REI:
Demo'ed to @brianlee-nava and @dghazvini for UAT. Now closing this ticket.

PR #9447, opened 2026-04-07
[Issue #9397] Add enhancements to the Opportunity List page

Dao-REI:
## Summary

Work for #9397 

## Changes proposed

- Enhancements were made to the Opportunity List page per Issue #9397. 

## Context for reviewers

To support the enhancements to the Opportunity List page, these additional changes were made:
- createOpportunityFetcher was renamed to grantorOpportunitiesFetcher and the searchOpportunitiesByAgency method was added to retrieve draft as well as published opportunities. (NOTE: Some backend work is needed to pull the published opportunities with that same endpoint. This will be in a separate ticket.)
- The component UserPrivileges.tsx was created to copy the checkRequiredPrivileges() function from AuthorizationGate.tsx to be used as a standalone component without the AuthorizationGate (per @doug-s-nava's instructions).

## Validation steps

- Code review
- UAT: validate the AC's in the ticket #9397

### DIFF
--- a/frontend/src/app/[locale]/(base)/opportunities/page.test.tsx
+++ b/frontend/src/app/[locale]/(base)/opportunities/page.test.tsx
@@ -3,11 +3,13 @@ import { axe } from "jest-axe";
 import OpportunitiesListPage from "src/app/[locale]/(base)/opportunities/page";
 import { UnauthorizedError } from "src/errors";
 import { UserAgency } from "src/services/fetch/fetchers/userAgenciesFetcher";
+import { UserSession } from "src/types/authTypes";
 import { LocalizedPageProps } from "src/types/intl";
 import { BaseOpportunity } from "src/types/opportunity/opportunityResponseTypes";
 import { FeatureFlaggedPageWrapper } from "src/types/uiTypes";
 import { DeepPartial } from "src/utils/testing/commonTestUtils";
 import { localeParams, useTranslationsMock } from "src/utils/testing/intlMocks";
+import { UserPrivilegeResult } from "src/utils/userPrivileges";
 
 import { FunctionComponent, ReactNode } from "react";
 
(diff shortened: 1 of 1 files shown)


frontend/src/app/[locale]/(base)/opportunities/page.tsx
```diff
@@ -201,41 +262,104 @@ async function OpportunitiesListPage(props: OpportunitiesListProps) {
...
+  ];
   try {
-    userOpportunities = (await searchForOpportunities(opportunitySearchParams))
-      .data;
+    userPrivilegeResult = await checkRequiredPrivileges(
+      userSession.token,
+      userSession.user_id,
+      userPrivilegeDef,
+    );
   } catch (error) {
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunities/page.tsx:370]: the way that `checkRequiredPrivileges` is written it will never throw. Do you want that behavior or is that copy pasta from the function in the auth gate?

frontend/src/utils/userPrivileges.ts
```diff
@@ -0,0 +1,47 @@
+import { parseErrorStatus } from "src/errors";
```
doug-s-nava, line 1: this seems more like a utility than a component

frontend/src/utils/userPrivileges.ts
```diff
@@ -0,0 +1,47 @@
...
+  authorized: boolean;
+  error?: string;
+}
+
+// calls the API to check all required privileges,
+// and formats results into a format to be used downstream
+// note that if a non-403 is returned, that error will be passed along
+// since we'll consider a non-403 as unauthorized, children should check for errors
+// first before checking "authorized"
+export const checkRequiredPrivileges = async (
```
doug-s-nava, line 23: this seems to be copied verbatim from the auth gate. Is that intentional? My assumption is that you would want different error handling in this case. Also, some of these comments only make sense in the context of the auth gate. We can talk about how to make a generic function that will leave error handling to the individual implementation if you want

frontend/src/services/fetch/fetchers/grantorOpportunitiesFetcher.test.ts
```diff
@@ -0,0 +1,145 @@
...
+      sort_direction: "ascending",
+    },
+  ],
+};
+type PaginationBody = {
+  pagination: PaginationRequestBody;
+};
+const pageBody: PaginationBody = { pagination: pageRequest };
+
+jest.mock("src/services/fetch/fetchers/fetchers");
```
doug-s-nava: as discussed in sync last week, please don't use this pattern https://betagrantsgov.slack.com/archives/C09RKTHP19N/p1775491404902519

frontend/src/services/fetch/fetchers/grantorOpportunitiesFetcher.test.ts
```diff
@@ -0,0 +1,145 @@
...
+  page_offset: 1,
+  page_size: 25,
+  sort_order: [
+    {
+      order_by: "opportunity_title",
+      sort_direction: "ascending",
+    },
+  ],
+};
+type PaginationBody = {
```
doug-s-nava: you shouldn't need to define types in a test. Can we find a way around this?

frontend/src/utils/userPrivileges.ts
```diff
@@ -0,0 +1,47 @@
+import { parseErrorStatus } from "src/errors";
+import { checkUserPrivilege } from "src/services/fetch/fetchers/userFetcher";
+
+export type GatedResourceTypes = "application" | "organization" | "agency";
+
+export type Privileges =
```
doug-s-nava, line 6:
```suggestion
export type AgencyOpportunityPrivileges =
```

frontend/src/app/[locale]/(base)/opportunities/page.tsx
```diff
@@ -75,6 +80,17 @@ const AgencyNotAuthorizedPage = () => {
   );
 };
 
+const NoViewPrivilegeMsg = () => {
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunities/page.tsx]: please avoid using abbreviations in variable names

frontend/src/utils/userPrivileges.ts
Dao-REI: I will put it inside userUtils.ts

frontend/src/services/fetch/fetchers/grantorOpportunitiesFetcher.test.ts
Dao-REI: I tried and it gave me a bunch of errors. I'll try again but I may need you to help me with this.

Dao-REI: I got it working!

frontend/src/utils/userPrivileges.ts
Dao-REI: There is "server only" code in the require path of the userFetcher, so bringing that in to userUtils means that anything importing userUtils now needs to be server only. Thus, we are putting the new check for privileges function in its own file under the utils folder.

frontend/src/utils/userPrivileges.ts
```diff
@@ -0,0 +1,43 @@
+import { parseErrorStatus } from "src/errors";
+import { checkUserPrivilege } from "src/services/fetch/fetchers/userFetcher";
+
+export type GatedResourceTypes = "application" | "organization" | "agency";
```
doug-s-nava, line 4: some of this looks identical to types already defined in userTypes, can we import those?

frontend/src/utils/userPrivileges.ts
```diff
@@ -0,0 +1,43 @@
...
+  privilege: Privileges;
+}
+
+export interface UserPrivilegeResult extends UserPrivilegeRequest {
+  authorized: boolean;
+  error?: string;
+}
+
+// Check the user's privileges
+export const checkRequiredPrivileges = async (
```
doug-s-nava, line 23: to distinguish this from the other `checkRequiredPrivileges` can we call it something different? Even better, though not a blocker, is to fold this behavior into the auth gate behavior (since it seems identical other than the error handling, we could pull the error handling out this function, call it from the auth gate, and allow errors to fall through for each implementation to handle errors how they want to)

frontend/src/services/fetch/fetchers/grantorOpportunitiesFetcher.test.ts
```diff
@@ -0,0 +1,142 @@
...
+} from "src/services/fetch/fetchers/grantorOpportunitiesFetcher";
+import { PaginationRequestBody } from "src/types/search/searchRequestTypes";
+import { fakeAgencyResponseData } from "src/utils/testing/fixtures";
+
+// Mock the main fetchGrantorWithMethod and the sub-method it calls, fetch
+const mockFetcher = jest.fn();
+const mockFetchGrantorWithMethod = jest.fn((_args: unknown) => mockFetcher);
+jest.mock("src/services/fetch/fetchers/fetchers", () => ({
+  fetchGrantorWithMethod: (arg: unknown): unknown =>
+    mockFetchGrantorWithMethod(arg),
```
doug-s-nava, line 13: ❤️

frontend/src/components/opportunities/OpportunityListComponents.tsx
```diff
@@ -0,0 +1,235 @@
...
+  canUpdate,
+  opportunityId,
+}: {
+  canUpdate: boolean;
+  opportunityId: string;
+}) => {
+  const t = useTranslations("Opportunities");
+  return canUpdate ? (
+    <span>
+      <a href={`/opportunity/${opportunityId}/edit`}>
```
doug-s-nava: sorry I missed this earlier, but for internal links we should use next/link in terms of a tags. They may not get us much, but good for consistency

PR #9516, opened 2026-04-10
[Issue #9514] Endpoint Update - Get List of Opportunities Missing Agency ID

### LINKED ISSUE #9514: Endpoint Update - Get List of Opportunities Missing Agency ID
### Summary

Update POST
[/v1/grantors/agencies/{agency_id}/opportunities](http://localhost:8080/docs#/Opportunity%20v1%20-%20for%20Grantors/post_v1_grantors_agencies__agency_id__opportunities)
Get paginated list of opportunities by agency

To include opportunities that have a null `agency_id` field (this applies to existing opportunities not created in Simpler Grants).

### Acceptance criteria

- [ ] Endpoint returns all opportunities for a given agency, including any that have a null `agency_id` field
- [ ] The List of Opportunities page shows all opportunities for a given agency

dghazvini:
## Summary

Fixes for #9514 

## Changes proposed

Added a join statement on the SQL of the Endpoint Get logic to grab opportunities using the `agency_code` as reference if the `agency_id` field is null.

## Context for reviewers

With the way the POST
 /v1/grantors/agencies/{agency_id}/opportunities
 -Get paginated list of opportunities by agency

endpoint is set up currently, it will not return any opportunities that have a null `agency_id` field. This isn't an issue for new opportunities created in Simpler Grants, but existing opportunities from grants.gov all have this field null. The purpose of this PR is to adjust the WHERE condition and add an Outer JOIN condition that will first check for the `agency_id` of the opportunity, and if it is null, it will reference the `agency_code` field instead and then reference the Agency table to determine the correct agency the opportunity belongs to.

## Validation steps

- [x] Confirmed that the endpoint is now returning all opportunities for a given agency, including opportunities where the `agency_id` field is null.
- [x] On the Opportunity List Page all opportunities for a given agency are being populated.
- [x] Added a unit test `test_get_opportunity_list_with_null_agency_id`

### DIFF
--- a/api/src/services/opportunities_grantor_v1/get_opportunity_list.py
+++ b/api/src/services/opportunities_grantor_v1/get_opportunity_list.py
@@ -70,7 +70,13 @@ def list_opportunities_with_filters(
             selectinload(Opportunity.opportunity_attachments),
             selectinload(Opportunity.competitions),
         )
-        .where(Opportunity.agency_id == agency_id)
+        # If opportunity.agency_id is null, use the opportunity.agency_code
+        .outerjoin(
+            Agency,
+            (Opportunity.agency_id == Agency.agency_id)
+            | ((Opportunity.agency_id.is_(None)) & (Opportunity.agency_code == Agency.agency_code)),
+        )
+        .where(Agency.agency_id == agency_id)
     )
 
     # Apply sorting in the database query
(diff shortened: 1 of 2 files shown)


api/src/services/opportunities_grantor_v1/get_opportunity_list.py
```diff
@@ -70,7 +70,13 @@ def list_opportunities_with_filters(
             selectinload(Opportunity.opportunity_attachments),
             selectinload(Opportunity.competitions),
         )
-        .where(Opportunity.agency_id == agency_id)
+        # If opportunity.agency_id is null, use the opportunity.agency_code
+        .outerjoin(
```
chouinar, line 74: Could you add a unit test to verify this?

dghazvini: Good point, I added a `test_get_opportunity_list_with_null_agency_id` and updated the Validation steps.

chouinar (approved):

Issue #9313, opened 2026-03-30
Create an endpoint that given an event ID will return the workflow


chouinar:
### Summary

One slight complexity with our workflow logic is that events are processed asynchronously so when a workflow is started, it can be hard to find the workflow that was created by that event.

We want to make an endpoint of `GET /v1/workflows/events/:event_id` that given an event ID will return a workflow.

#### Request
n/a - just the event ID path parameter

#### Response
Exactly the same as the GET workflow endpoint we already have

#### AuthZ
Identical to the GET workflow endpoint, see logic section below

Both API User Key + JWT should work for this endpoint

#### Logic
The logic of this endpoint should:
* Fetch an event from the `workflow_event_history` by the passed in `event_id`
* 404 if not found
* Grab the attached workflow via the relationship
* Do all of the same logic on the workflow as we did for GET workflow

We'll likely need to slightly refactor the GET workflow endpoints logic to be reusable. Everything in `get_workflow_and_verify_access` should be reused except for the way we find the workflow.

### Acceptance criteria

- [ ] API endpoint created
- [ ] Tests added for common cases (success, errors, auth)

PR #9519, opened 2026-04-10
[Issue #9313] Create an endpoint that given an event ID will return the workflow

kkrug:
## Summary

Fixes for #9313 

## Changes proposed

- Updated get_workflow.py to extract `_verify_workflow_access_and_build_config()` from `get_workflow_and_verify_access()` to make the auth/access logic reusable and added `get_workflow_by_event_id_and_verify_access()`
- Updated workflow_routes.py to add `GET /v1/workflows/events/<uuid:event_id>` endpoint
- Updated test_workflow_routes.py to add tests for new endpoint

## Context for reviewers

One slight complexity with our workflow logic is that events are processed asynchronously so when a workflow is started, it can be hard to find the workflow that was created by that event.

We want to make an endpoint of `GET /v1/workflows/events/:event_id` that given an event ID will return a workflow.

## Validation steps

Confirm tests pass.

### DIFF
--- a/api/src/services/workflows/get_workflow.py
+++ b/api/src/services/workflows/get_workflow.py
@@ -12,7 +12,12 @@ from src.db.models.agency_models import Agency
 from src.db.models.competition_models import Application, ApplicationSubmission, Competition
 from src.db.models.opportunity_models import Opportunity
 from src.db.models.user_models import User
-from src.db.models.workflow_models import Workflow, WorkflowApproval, WorkflowAudit
+from src.db.models.workflow_models import (
+    Workflow,
+    WorkflowApproval,
+    WorkflowAudit,
+    WorkflowEventHistory,
+)
 from src.workflow.registry.workflow_registry import WorkflowRegistry
 from src.workflow.service.approval_service import get_agency_for_workflow
 from src.workflow.workflow_errors import OpportunityWithoutAgencyError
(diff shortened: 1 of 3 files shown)


api/src/services/workflows/get_workflow.py
```diff
@@ -140,6 +137,78 @@ def get_workflow_and_verify_access(
...
+        logger.info("Event has no associated workflow", extra=log_extra)
+        raise_flask_error(404, message=f"Could not find Workflow for Event with ID {event_id}")
+
+    workflow = _get_workflow(db_session, event.workflow_id)
+
+    if workflow is None:
+        logger.info("Workflow not found for event", extra=log_extra)
+        raise_flask_error(404, message=f"Could not find Workflow for Event with ID {event_id}")
+
+    return _verify_workflow_access_and_build_config(db_session, user, workflow)
```
chouinar: We can probably simplify this down like so. I think the main thing we get out of the _get_workflow call is all the data fetches it does that we need in the response? We could put that in the query above, or make a util that sets them all that the two endpoints share.

```suggestion
    if event.workflow is None:
        logger.info("Event has no associated workflow", extra=log_extra)
        raise_flask_error(404, message=f"Could not find Workflow for Event with ID {event_id}")

    return _verify_workflow_access_and_build_config(db_session, user, event.workflow)
```

kkrug: I made some updates to extract out shared functionality

hao10282025-sudo (approved):

PR #9522, opened 2026-04-10
[Issue #9362] Refactor simpler_soap_api_routes

### LINKED ISSUE #9362: SOAP/Proxy: Refactor simpler_soap_api_route
### Summary

Currently `simpler_soap_api_route` is more complicated than our other API endpoints. We would like to simplify it down to be more in-line with our other apis. 

For example this route is much less complicated than the `simpler_soap_api_route`:
Image

The goal would be to simplify the route down to:
`POST -> logging -> response`

We would extract all the logic out to another file like `simpler_soap_api.py` and just call the method from there.

This is just to extract out the logic, further refactoring will be on another ticket.

### Acceptance criteria

- [ ] `simpler_soap_api_route` is refactored to be much simpler
- [ ] logic extracted out to another file
- [ ] tests passing

jakobpederson:
## Summary

Fixes / Work for #9362  

## Changes proposed

Move code out of the routes file.

## Context for reviewers

Our standard pattern is keep logic out of the api route, so I moved the code out of it and also separated out a logging check instead of checking it twice.

## Validation steps

- [ ] tests passing

### DIFF
--- a/api/src/legacy_soap_api/legacy_soap_api_proxy.py
+++ b/api/src/legacy_soap_api/legacy_soap_api_proxy.py
@@ -60,11 +60,6 @@ def get_proxy_response(soap_request: SOAPRequest, timeout: int = PROXY_TIMEOUT) 
 
     soap_auth = soap_request.auth
     use_soap_cert = soap_request.headers.get(USE_SOAP_CERT_HEADER_KEY) == "1"
-    if not use_soap_cert:
-        logger.info(
-            "soap_client_certificate: Using the soap jwt",
-            extra={"soap_api_event": LegacySoapApiEvent.CALLING_WITH_JWT},
-        )
     should_log_response = soap_request.headers.get(LOG_LOCAL_RESPONSE_HEADER_KEY) == "1"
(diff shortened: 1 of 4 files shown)


api/src/legacy_soap_api/legacy_soap_api_proxy.py
```diff
@@ -60,6 +60,11 @@ def get_proxy_response(soap_request: SOAPRequest, timeout: int = PROXY_TIMEOUT)
 
     soap_auth = soap_request.auth
     use_soap_cert = soap_request.headers.get(USE_SOAP_CERT_HEADER_KEY) == "1"
+    if not use_soap_cert:
+        logger.info(
+            "soap_client_certificate: Using the soap jwt",
```
kkrug: Is it intentional to have this log message twice? This log will be triggered for every non-cert request and then when we do go through the JWT path, we hit the same log message below on line 79.

jakobpederson: This was actually to preserve current behavior, so it turns out we are logging this message twice. I'll pull it.

kkrug (approved):

PR #9528, opened 2026-04-10
[Issue #9040] Update the current email task for user/self saved opportunities to respect the new preference

### LINKED ISSUE #9040: [BE] Update the current email task for user/self saved opportunities to respect the new preference
### Summary

We need to update the current task that sends saved opportunities to respect the new preference for self notifications on this feature.

This notification will be denoted as enabled in the `user_saved_opportunities_notifications` database if there is an entry that contains the following values:

```sql
organization_id=null,  user_id=<user_id>, email_enabled=true
```

### Notes

See https://github.com/HHS/simpler-grants-gov/issues/9011#issuecomment-4121044937 comment for running locally

### Acceptance criteria

- [ ] Emails should not be sent to users who have set a saved opportunities notification for self where email_enabled is false.
   - [ ] Code should be refactored to add reuse in a service rather than logic duplicated with the API end point.
- [ ] If users disable and then later re-enable notifications, they should not get a backlog of past notifications.  They should only receive notifications from this time forward.
   - [ ] Practically, this means that opportunity updates should be marked as "notification sent" when the user has a preference of email_enabled is false without an email being sent; what this field really means is "notification processed" and  whether we sent an email or not depends on the user's preference.
- [ ] Unittests
- [ ] UAT

mikehgrantsgov:
## Summary

Fixes #9040

## Changes proposed

Added `NOTIFICATIONS_SKIPPED_EMAIL_DISABLED` metric to track how many users are skipped per run.
Added `_get_users_with_email_disabled()` which queries for users with a personal preference row. In `collect_email_notifications()`, these users are now skipped for email sending. In `post_notifications_process()`, their `last_notified_at` is still advanced,  so re-enabling notifications later won't re-email them.
New tests covering the disabled case and the re-enable/no-backlog case.

## Context for reviewers

We need to update the current task that sends saved opportunities to respect the new preference for self notifications on this feature.

## Validation steps

See new tests:
`make test args="tests/src/task/notifications/test_opportunity_notification.py -v"`


babebe (approved):
LGTM. Tests confirm behavior.

PR #9529, opened 2026-04-10
[Issue #9061] Breadcrumbs tests

### LINKED ISSUE #9061: Improve Breadcrumbs unit tests
### Summary

With all pages what consume the `<Breadcrumbs>` component standardized in how they manage content, and the returnToGrants logic removed, we can improve our test coverage for breadcrumbs. We should create robust test for the `<Breadcrumbs>` component itself. Then, on pages that call the component, simply test that the component exists (check it's created and renders, instead of asserting that it functions). 

As `Breadcrumbs` no longer contains `ReturnToGrantsNotification` (once #8951 is merged), we should also make sure breadcrumb tests are not concerned w/ `ReturnToGrantsNotification` and does not include/import anything it required. 

### Acceptance criteria

- [ ] `<Breadcrumb>` component has its own tests
- [ ] Pages with breadcrumb have standardized test methods

andycochran:
## Summary

Fixes #9061 

## Changes proposed

- Add `Breadcrumbs.test.tsx` asserting that:
  - It renders a link 
  - It renders the current item without a link
  - It passes a11y scan 

## Context for reviewers

The Breadcrumbs component did not have a test file. 

## Validation steps

- Run tests
- Tests pass


andycochran:
@doug-s-nava / @btabaska, do you have any thoughts on how we might simplify all the tests for pages with breadcrumbs? They don't need to assert that the component does what it's already being tested to do? Currently, pages with breadcrumbs are mocking the component, then asserting that they've been called with the right object array. That's testing content and seems unnecessary.

doug-s-nava:
@andycochran I think what should be tested is the idea of "breadcrumbs are on the page". Seems like the easiest way to do this is `expect(screen.getByLabel("Breadcrumbs")).toBeInTheDocument()`, since the BreadCrumbBar has an aria-label we can target https://github.com/trussworks/react-uswds/blob/main/src/components/breadcrumb/BreadcrumbBar/BreadcrumbBar.tsx#L29. That doesn't seem immediately relevant to this PR though?

andycochran:
The ticket this PR is addressing has the AC: 

> - [ ] Pages with breadcrumb have standardized test methods

I didn't see any pages doing anything particularly weird w/ breadcrumb tests (especially since all the `ReturnToGrantsNotification` stuff is removed). I was just thinking they might be simplified. But I'm happy to leave them as-is for now. Might we need to revisit w/ i18n deprecation work?

andycochran:
> … what should be tested is the idea of "breadcrumbs are on the page"

That seems right. There's no reason to test the content of what's mocked if the component has its own test. Mocked breadcrumbs don't prove the real content is being rendered; they just retest what the new test in this PR already does. I'll update the breadcrumb tests on all the pages so they just look for the component's existence.

andycochran:
@doug-s-nava this PR has been sitting dormant for a while. I think it's fine as-is. It just adds a missing test for the `Breadcrumbs` component. Are you okay with merging this and handling separately any rework of page tests that assert "breadcrumbs are on the page"?

frontend/src/components/Breadcrumbs.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+    title: "Home",
+    path: `/`,
+  },
+  {
+    title: "Current",
+  },
+];
+
+describe("Breadcrumbs", () => {
+  it("Has a 'Home' item that is a link", () => {
```
doug-s-nava: I'd prefer more generic test descriptions ex `contains segments with links and text that match passed list` and `has a final segment that is not a link`

doug-s-nava (dismissed):
✅ ✅ ✅ ✅ 

nothing blocking!!!

frontend/src/components/Breadcrumbs.test.tsx
andycochran: Updated. Re-review/approve please?

doug-s-nava (approved):

Issue #9530, opened 2026-04-10
[FE] Add new relic log metric for search query length


freddieyebra:
### Summary

We need to capture metrics on the length of opportunity search queries.

This ticket adds functionality to capture that metric in the frontend application.

### Acceptance criteria

- [ ] Log search query length and metric is shown in new relic
- [ ] unit tests pass

Issue #9358, opened 2026-03-31
Search Engagement & Usage Dashboard - Filter & Sort Deep Dive [3 of 4]


joshtonava:
### Summary

Add a Search Engagement page showing how users interact with search tools: what they search for, which filters they use, how deeply they paginate, and when they're most active. Consolidates panels from the existing Search Parameters and User/Usage Metrics dashboards into the Find Metrics dashboard. All panels use existing data.

**Panels:**
- **Device Type Distribution**
  Breakdown of sessions by device type. Clicking a device type filters all other panels on this page.
- **Filter Usage Frequency**
  Which filters are used most (agency, category, eligibility, status, etc.). Helps prioritize which filters to improve or promote.
- **Query Length Distribution**
  Word/character count buckets for search queries. Short queries may indicate broad exploration while long queries suggest users know what they want but can't find it.
- **Sort Usage Frequency**
  How often each sort option is used. Low usage of non-default sorts may mean users don't know the option exists or don't find it useful.

**Not included (pending future instrumentation):**                                                                                                                 
- Authenticated vs anonymous search patterns (needs correlation ID — #8687, #8689)                                                                             
- Saved search execution rate vs manual search rate (needs frontend saved search context)                                           
- Filter depth vs save rate correlation (needs frontend save context)                                    
- Common filter combinations (derivable from backend logs but may require complex NRQL, evaluate feasibility during implementation)

### Acceptance criteria
                                                                                                                           
- [ ] Page 3 "Search Engagement & Usage" added to Find Metrics dashboard                                                                                           
- [ ] Device type panel with facet linking filters all other panels on this page                                                                                   
- [ ] Panels using existing NR Log and UserAction data are populated and functional
- [ ] UAT with Eric
- [ ] NRQL queries documented for each panel

freddieyebra:
@joshtonava @ErinPattisonNava there is no support for calculating string length in nrql. for the Query Length Distribution metric, well need to add a query length metric. i created a ticket for that [here](https://github.com/HHS/simpler-grants-gov/issues/8250?issue=HHS%7Csimpler-grants-gov%7C9530)

freddieyebra:
Created diagrams for metrics in this issue here [here](https://one.newrelic.com/dashboards/detail/NTI0OTgwOXxWSVp8REFTSEJPQVJEfGRhOjEyMzAzNTgy?account=5249809&state=722993d1-31b8-5759-797a-d00056b21680)

[...]

PR #9570, opened 2026-04-13
[Issue #9530] Add search query length log metric to be pushed to new relic

freddieyebra:
## Summary

Fixes / Work for #9530 

## Changes proposed

This pr adds a custom new relic log metric for search query length.

## Context for reviewers

As part of #9358, we need to capture how the length/character count of search queries impacts user engagement. New relic  does not support calculating string metric lengths so this pr ensures we have insights into that.
 
## Validation steps
Run the following query in new relic: 
```
FROM BrowserInteraction SELECT search_param_query, search_param_query_length WHERE search_param_query_length IS NOT NULL
```

This pr has already been deployed and tested during dev testing. To validate this pr you can either deploy this branch to lower environments and make some search queries or just run the above since they already exist in rew relic.

### DIFF
--- a/frontend/src/utils/analyticsUtil.ts
+++ b/frontend/src/utils/analyticsUtil.ts
@@ -40,7 +40,7 @@ const getNewRelicBrowserInstance = (): NewRelicBrowser | null => {
 
 export const setNewRelicCustomAttribute = (
   key: string,
-  value: string,
+  value: string | number,
 ): undefined => {
   const newRelic = getNewRelicBrowserInstance();
   if (!newRelic) {
@@ -56,4 +56,5 @@ export const unsetAllNewRelicQueryAttributes = () => {
   validSearchQueryParamKeys.forEach((key) => {
     setNewRelicCustomAttribute(key, "");
   });
+  setNewRelicCustomAttribute("query_length", 0);
 };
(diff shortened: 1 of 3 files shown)


frontend/src/utils/analyticsUtil.ts
```diff
@@ -56,4 +56,5 @@ export const unsetAllNewRelicQueryAttributes = () => {
   validSearchQueryParamKeys.forEach((key) => {
     setNewRelicCustomAttribute(key, "");
   });
+  setNewRelicCustomAttribute("query_length", 0);
```
myduong-navapbc, line 59: Rather than setting the value to `0`, set it to `null`

https://docs.newrelic.com/docs/browser/new-relic-browser/browser-apis/setcustomattribute/

_Passing a `null` value unsets any existing attribute of the same name._

---

But just to confirm, you're "unsetting" the value right? not setting it to the value of `0`?

frontend/src/utils/analyticsUtil.ts
```diff
@@ -40,7 +40,7 @@ const getNewRelicBrowserInstance = (): NewRelicBrowser | null => {
 
 export const setNewRelicCustomAttribute = (
   key: string,
-  value: string,
+  value: string | number,
```
myduong-navapbc, line 43: see comment below:

if you pass `null`, then this type becomes `value: string | number | null`

frontend/src/utils/analyticsUtil.ts
freddieyebra: yeah i considered that but the type defined for the value from the package were using has the value type as 

```
type SimpleType = string | number;
```

It could be different version of the package. i am looking now.

myduong-navapbc:
Additionally, I dont think i can completely test until this is merged

myduong-navapbc (approved):

Issue #8927, opened 2026-03-09
Backend - Create the create award recommendation endpoint


nathan-stricker:
### Summary

See [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Create-the-create-award-recommendation-endpoint)

#### Endpoint
`POST /alpha/award-recommendations`

#### Purpose
Create an award recommendation

#### AuthZ
Requires `create_award_recommendation` privilege

#### Request
```json
{
  "opportunity_id": "",
  "additional_info": "",
  "award_selection_method": "", # Enum
  "funding_strategy": "",
  "other_key_information": "",
  # is_deleted doesn't need to be included
  "additional_info": "",
}
```

#### Response
The same as GET award-recommendation

#### Logic

- Fetch the opportunity, 404 if not found
- AuthZ, 403 on error
    - There’s a technically possible scenario where the opportunity has no agency attached. This should always 403 to be safe.
- Generate an award recommendation number of `agency_code` + a random 10 character string (all capital letters + numbers).
- For the opportunity, get all application submissions (even if multiple competitions):
    - Link the application submission to the award recommendation
    - Create an award recommendation submission detail with the following defaults:
        - Set `scoring_comment` to a random integer between 1-100 (this is to simulate scoring having been done, which does not exist at all yet)
        - Set the `award_recommendation_type` to `not_recommended` 
        - Set the `recommended_amount` to whatever is in the `application_submission.requested_amount` column.
- For each type of award recommendation review, we want to add a record to the award recommendation review table with that type, and the `is_reviewed` flag set to False

When creating the award recommendation, set the status to `in_progress`.
**Store an audit event to the audit table with the audit type of `award_recommendation_created`**

### Acceptance criteria

- [ ] Endpoint Created
- [ ] Tests Created

PR #9574, opened 2026-04-13
[Issue #8927] Backend - Create the create award recommendation endpoint

chay-REIsys:
## Summary

 Work for #8927 

## Changes proposed

Create Award Recommendation Endpoint

## Context for reviewers

We should be able to create award recommendation for an opportunity

## Validation steps

- fetch new changes
- Search for Create Award Recommendation Endpoint and click Try it out
- use below input to test the award recommendation success scenario.

{
 "additional_info": "test additional info",
  "award_selection_method": "merit_review_ranking_only",
  "funding_strategy": "test funding strategy",
 "opportunity_id": "1aa2baf1-294f-4109-812b-427954ff11eb",
 "other_key_information": "test other key info"
}


api/tests/src/api/award_recommendations/test_award_recommendation_create.py
```diff
@@ -0,0 +1,196 @@
...
+        assert data["award_recommendation_number"].startswith(agency.agency_code)
+        assert len(data["award_recommendation_number"]) == len(agency.agency_code) + 10
+
+        assert len(data["award_recommendation_reviews"]) == len(AwardRecommendationReviewType)
+        assert all(
+            review["is_reviewed"] is False for review in data["award_recommendation_reviews"]
+        )
+
+
+class TestCreateAwardRecommendation400:
```
kkrug:
```suggestion
class TestCreateAwardRecommendation422:
```

api/tests/src/api/award_recommendations/test_award_recommendation_create.py
```diff
@@ -0,0 +1,196 @@
...
+        assert data["additional_info"] == "Supporting materials reviewed"
+        assert data["other_key_information"] == "Priority funding for rural proposals"
+        assert data["opportunity"]["opportunity_id"] == str(opportunity.opportunity_id)
+        assert data["award_recommendation_number"].startswith(agency.agency_code)
+        assert len(data["award_recommendation_number"]) == len(agency.agency_code) + 10
+
+        assert len(data["award_recommendation_reviews"]) == len(AwardRecommendationReviewType)
+        assert all(
+            review["is_reviewed"] is False for review in data["award_recommendation_reviews"]
+        )
```
kkrug, line 88: Can we add an assertion for ensuring `selection_method_detail` matches `funding_strategy`?

`assert data["selection_method_detail"] == "Allocate across two grants"`

------------------------------------------------------------------------------

Also, I think it is worthwhile to add db level assertions for submission linkage and scoring defaults. Since `award_recommendation_application_submissions` isn't included in the GET response schema, after the POST we need to add something like:

```
linked = db_session.query(AwardRecommendationApplicationSubmission).filter_by(
    award_recommendation_id=data["award_recommendation_id"]
).all()
assert len(linked) == 2

for link in linked:
    detail = link.award_recommendation_submission_detail
    assert detail.award_recommendation_type == AwardRecommendationType.NOT_RECOMMENDED
    assert detail.recommended_amount in (Decimal("1000.00"), Decimal("2000.00"))
    assert 1 <= int(detail.scoring_comment) <= 100
```

kkrug (commented):
For documentation purposes, let's rename the PR "[Issue #8927] Backend - Create the create award recommendation endpoint"

api/tests/src/api/award_recommendations/test_award_recommendation_create.py
chay-REIsys: actually both are different and in the following PR upon discussion with michael we need to either create a column or use any existing column wich michael suggests

kkrug: Do you mean that `funding_strategy` should not map to `selection_method_detail`? Is this just a temporary measure or a bug that needs to be fixed?

chay-REIsys: yes as of now it is wired up like that but we have recommendation method and recommendation method details on UI and also funding strategy. 
image

kkrug: Got it, so then my only suggestion is the db level assertions mentioned above.

kkrug (approved):

PR #9582, opened 2026-04-14
[Issue #9579] E2E tests: Update radio button selection logic

### LINKED ISSUE #9579: E2E tests: Update radio button selection logic
### Summary

Updated the radio button fill data value and selector in the e2e test to ensure the correct option is reliably located and clicked during form submission.

### Acceptance criteria

- The happy path e2e test passes consistently across all browsers
- The correct radio button option is clicked during form fill
- The form saves successfully and displays the success alert after submission

Bhavna-Ramachandran:
## Summary
Work for: Task #9579 

## Changes proposed
- Updated the radio button fill logic in general-forms-filling.ts so that when a field definition specifies getByText (e.g., for a "No" or "Yes" radio option), the corresponding radio button is always clicked, regardless of the data value.
- This ensures that required radio fields are always filled during E2E tests, even when the test data is "false" (for "No").
- The change prevents form validation failures caused by unselected radio buttons when the value is "falsy" but the field definition explicitly encodes which option to select.

## Context for reviewers
- Previously, the code only clicked a radio button if the test data was "truthy" (e.g., "true" or "Yes"). If the data was "false", the radio was not clicked, even if the field definition specified a label to select.
- This caused E2E tests to fail for forms that required a "No" radio selection, as the field was left empty and validation failed.
- The new logic always clicks the radio button when getByText is set, making the test behavior match the intent of the field definition and test data.

## Validation steps
- Run the E2E test suite in Local and Staging and confirm that all tests pass

### DIFF
--- a/frontend/tests/e2e/utils/forms/general-forms-filling.ts
+++ b/frontend/tests/e2e/utils/forms/general-forms-filling.ts
@@ -153,7 +153,10 @@ export async function fillField(
       field.type === "radiobutton" &&
       (field.testId || field.selector || field.getByText || field.useDataAsText)
     ) {
-      if (shouldActivateField(data)) {
+      // If getByText is specified, the field definition already encodes which
+      // specific radio option to click (eg - "No"), so always activate it
+      // regardless of the data value. Otherwise, rely on shouldActivateField.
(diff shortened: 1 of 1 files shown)


doug-s-nava (approved):
✅ ✅ ✅ ✅

Issue #9591, opened 2026-04-14
Upgrade Next to 16.2.3


mdragon:
### Summary

Upgrade to newest Next, 16.2.3

### Acceptance criteria

- Scans pass
- Tests pass
- App smoke tested

PR #9618, opened 2026-04-15
[Issue #9591] Upgrade next

mdragon:
## Summary

Work for #9591 

## Changes proposed

Upgraded next via codemod
Adjusted Lint config/rules and resolved outstanding lint errors

## Context for reviewers

Had a vulnerability fix that required us jumping from 16.1.x to 16.2.x that introduced some lint stuff that we're working through.
Opened 2 tickets to track future follow up evaluation/fixes for things the linter is mad about and we're ignoring for now on this PR: #9633 #9637

## Validation steps

- Checks pass
- Smoke test app


frontend/eslint.config.js
```diff
@@ -31,7 +31,7 @@ module.exports = defineConfig([
       "nava",
       "plugin:storybook/recommended",
       "prettier",
-      "next/core-web-vitals",
+      // "next/core-web-vitals",
```
mdragon: This is crashing the linter so commenting out while I research further. If this drags out, I'll open a new ticket to fix this part and let the 16.2.3 upgrade move forward.

mdragon: I have tried several potential fixes for this issue but haven't had success yet. I'm going to time box this to today, and open a follow up ticket if we can't figure it out and get this re-introduced.

mdragon: Think I got this working. But there's 21 lint issues I have to clean up.

mdragon:
I think this has stabilized and is ready for review.

frontend/eslint.config.mjs
```diff
@@ -50,6 +55,12 @@ module.exports = defineConfig([
           ],
         },
       ],
+      // TODO #9637 remove this off line and address the issues where this rule flags stuff
+      "promise/catch-or-return": "off",
```
doug-s-nava: can we make this "warn" instead of "off" so that we can track the issues better (and maybe keep a fire lit to address them)?

api/poetry.lock
```diff
@@ -1654,14 +1654,14 @@ htmlsoup = ["BeautifulSoup4"]
 
 [[package]]
 name = "mako"
-version = "1.3.10"
+version = "1.3.11"
```
doug-s-nava, line 1657: is this related to the next change? if not can we take it out?

frontend/src/components/application/InformationCard.tsx
```diff
@@ -373,7 +373,7 @@ export const InformationCard = ({
     );
   };
 
-  return (
+  const detailsRender = (shouldBlockSubmitForTransferOwnership: boolean) => (
```
doug-s-nava: no idea why this fixes the error, but the issue is still there - the actual issue is that we're creating components within the render

* ApplicationInstructionsDownload
* ApplicationSubmissionDownload
* NoApplicationInstructionsDownload
* SpecialInstructions
* NeedsTransferOwnershipCta
* InformationCardDetails

All need to be defined outside of the component render for InformationCard

frontend/src/services/featureFlags/withFeatureFlag.tsx
```diff
@@ -22,16 +25,11 @@ const withFeatureFlag = <P, R extends ReactNode>(
     return WrappedComponent;
   }
 
-  // top level component to grab search params from the top level page props
-  const ComponentWithFeatureFlagAndSearchParams = async (
+  function wrapComponent(
```
doug-s-nava: similar to the above, not sure why this satisfies the linter, as it doesn't fix the problem. I have a solution up here that passes unit tests and linting, but have not manually tested https://github.com/HHS/simpler-grants-gov/pull/9682

frontend/package.json
```diff
@@ -88,10 +88,12 @@
     "@typescript-eslint/parser": "^8",
     "dotenv": "^17.2.2",
     "eslint": "^9.39.0",
-    "eslint-config-next": "^15",
+    "eslint-config-next": "16.2.3",
     "eslint-config-prettier": "^8.7.0",
     "eslint-plugin-jest": "^28",
     "eslint-plugin-jest-dom": "^5.0.1",
+    "eslint-plugin-promise": "^7.2.1",
```
doug-s-nava, line 95: why add promise and react hooks? Were those previously included in another package that dropped them?

mdragon: Yeah, I had added the ignores and then in shifting around the packages to get stuff working again the rules disappeared and so they were flagging so I started to put them back in via their original packages. I could pull them back out, but it also seemed like being a little more strict could make sense.

api/poetry.lock
mdragon: I pulled it in so the checks would pass on this branch. It's in main now so I think it might drop when I rebase. If not I can get rid of it.

frontend/src/components/application/InformationCard.tsx
mdragon: Reverted this, changed off to warn

frontend/src/services/featureFlags/withFeatureFlag.tsx
mdragon: Reverted this, changed off to warn

frontend/package.json
doug-s-nava: any idea what it was that caused them to get removed? Agree we should have them in here, just curious. The main thing that seems like could have happened was that the next lint package dropped them in v16, but that's a weird decision if true

frontend/src/components/application/InformationCard.tsx
doug-s-nava: I don't see the change here, was it pushed?

frontend/eslint.config.mjs
```diff
@@ -50,6 +55,14 @@ module.exports = defineConfig([
...
         },
       ],
+      // TODO #9637 remove this warn line after addressing the issues where this rule flags stuff
+      "promise/catch-or-return": "warn",
+      // TODO #9637 remove this warn line after addressing the issues where this rule flags stuff
+      "promise/always-return": "warn",
+      // TODO #9637 remove this warn line after addressing the issues where this rule flags stuff
+      "react-hooks/error-boundaries": "warn",
+      // TODO evaluate the right way to fix this later
+      "react-hooks/static-components": "warn",
```
doug-s-nava: can you ignore this in the specific locations where it is being thrown currently rather than globally?

mdragon: Done

frontend/src/components/application/InformationCard.tsx
mdragon: Yeah weird, the push is in my terminal but there's no output from it running, maybe I CTRL+c'd it and then forgot to go back. Pushed now.

frontend/package.json
mdragon: I thought it was when I mirrored your change dropping nava and putting in eslint recommended but if you're not sure why it happened either, then I'm not sure exactly when they fell off.

doug-s-nava: not sure what the story with promises is, so I'm good with including that here, but the react-hooks plugin seems to still be in the project on your branch, not sure why we need to reinclude it. Can you try removing the explicit declaration here and see what happens?

mdragon: Looks like it can be removed without issue. Pushed that up

doug-s-nava (approved):
ran through full apply workflows with no errors

✅ ✅ ✅ ✅

PR #9627, opened 2026-04-15
[Issue #9396] Create Opportunity - Enhancements

### LINKED ISSUE #9396: Create Opportunity - Enhancements
### Summary

Turn Feature Flag off
Create the Menu item & link to the Opportunity List page

### Acceptance criteria

- [ ] Turn Feature Flag on (update opportunitiesListOff to be true) - This will hide the feature from the user
-  [ ] Create Opportunity page: on Agency selection, check for the create_opportunity privilege, show an error message as needed

dghazvini:
## Summary

Fixes for #9396 

## Changes proposed

Turn Feature Flag on (update opportunitiesListOff to be true) - This will hide the feature from the user

Create Opportunity page: on Agency selection, check for the create_opportunity privilege, show an error message as needed

## Validation steps

- [x] Verified on localhost that top alert populates and Save and Continue button is disabled when an agency that the user does not have the create_opportunity privilege for is selected.
- [x] Tests added in `actions.test.tsx`
- [x] Updates to `CreateOpportunityForm.test.tsx` to resolve any issues with adding the agency validation.
- [x] Renamed the `checkRequiredPrivileges` function being used by Opportunity Publishing (not in the AuthorizationGate) to `checkUserRequiredPrivileges` to satisfy linter and not cause confusion with AuthorizationGate; I also updated all references using this function in utils/userPrivileges.ts


frontend/src/app/[locale]/(base)/opportunities/create/actions.tsx
```diff
@@ -50,3 +54,53 @@ export const createOpportunityAction = async (
...
+      userPrivilegeResult.forEach((result) => {
+        if (result.privilege === "create_opportunity" && result.authorized) {
+          canCreate = true;
+        }
+      });
+    }
+
+    if (canCreate) {
+      return { success: true };
+    } else {
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunities/create/actions.tsx]: no need for this else due to early return

frontend/src/app/[locale]/(base)/opportunities/create/actions.tsx
```diff
@@ -50,3 +54,53 @@ export const createOpportunityAction = async (
...
+      return {
+        error:
+          "You do not have access to create opportunities for this agency.",
+      };
+    }
+  } catch (_error) {
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunities/create/actions.tsx]: can we also log the error here

frontend/src/app/[locale]/(base)/opportunities/create/actions.tsx
```diff
@@ -50,3 +54,53 @@ export const createOpportunityAction = async (
...
+    return { error: "Session error" };
+  }
+
+  try {
+    const userPrivilegeResult = await checkUserRequiredPrivileges(
+      session.user_id,
+      getUserPrivilegeDefinition(agencyId),
+    );
+
+    let canCreate = false;
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunities/create/actions.tsx]: [nit] we can avoid a mutable variable with a reduce

```suggestion
    const canCreate = userPrivilegeResult.length ? userPrivilegeResult.reduce((permissionFound, result) => result.privilege === "create_opportunity" && result.authorized, false): false
```

frontend/src/app/[locale]/(base)/opportunities/create/actions.tsx
```diff
@@ -50,3 +54,53 @@ export const createOpportunityAction = async (
...
+
+export async function validateAgencyAccessAction(agencyId: string) {
+  const session = await getSession();
+
+  if (!session?.token || !session?.user_id) {
+    return { error: "Session error" };
+  }
+
+  try {
+    const userPrivilegeResult = await checkUserRequiredPrivileges(
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunities/create/actions.tsx:66]: we're only checking one user privilege, right? Is there a reason we can't make a single call to /can_access?

frontend/src/constants/defaultFeatureFlags.ts
```diff
@@ -5,5 +5,5 @@ export const defaultFeatureFlags: FeatureFlags = {
   awardRecommendationOff: true,
   featureFlagAdminOff: false,
   maintenanceMode: false,
-  opportunitiesListOff: false,
+  opportunitiesListOff: true,
```
doug-s-nava: just a note that this will only affect local environments - deployed envs will take their default value from env vars defined in SSM

doug-s-nava (commented):
You can probably ignore most of my individual comments here, since we should be doing this with the AuthorizationGate instead of implementing custom code to check for auth status. Let me know if you need help implementing that

frontend/src/constants/defaultFeatureFlags.ts
dghazvini: Reverted

frontend/src/app/[locale]/(base)/opportunities/create/actions.tsx
Comment (dghazvini) [reply frontend/src/app/[locale]/(base)/opportunities/create/actions.tsx]: Done

frontend/src/app/[locale]/(base)/opportunities/create/actions.tsx
Comment (dghazvini) [reply frontend/src/app/[locale]/(base)/opportunities/create/actions.tsx]: Done

frontend/src/app/[locale]/(base)/opportunities/create/actions.tsx
Comment (dghazvini) [reply frontend/src/app/[locale]/(base)/opportunities/create/actions.tsx]: Done

Issue #9315, opened 2026-03-30
Modify the GET /workflow endpoint response to include events that could be sent


chouinar:
### Summary

To make it easier for folks to work with our workflows, we want to modify the GET /workflow endpoint to return a list of events that could be sent for the current state of a workflow.

For the response object, add a list of strings called `valid_events` that we'll populate.

To determine what events are valid, do the following (some of this might be present already, reuse it):

```python
# Get the state machine class + config
config, state_machine_cls = WorkflowRegistry.get_state_machine_for_workflow_type(workflow_type)

# create a state machine
state_machine = state_machine_cls(config.persistence_model)
valid_events = state_machine.get_valid_events_for_current_state()
```

The `get_valid_events_for_current_state` function should be setup so it's only returning events relevant for the current state.

Note that if a workflow's `is_active` flag is False, we can shortcut this and just return `[]` since no events are ever valid once a workflow has finished.

### Acceptance criteria

- [ ] `valid_events` added to GET /workflow endpoint schema
- [ ] Field populated as described above
- [ ] Tests added

PR #9631, opened 2026-04-15
[Issue #9315] Modify the GET /workflow endpoint response to include events that could be sent

kkrug:
## Summary

Fixes for #9315  

## Changes proposed

- Updated worflow_schemas.py to add `valid_events` field to `WorkflowGetResponseDataSchema`
- Updated get_workflow.py to add `get_valid_events_for_workflow()`
- Updated test_workflow_routes.py to add tests and account for changes

## Context for reviewers

To make it easier for folks to work with our workflows, we want to modify the GET /workflow endpoint to return a list of events that could be sent for the current state of a workflow.

## Validation steps

Confirm tests pass.


api/tests/src/api/workflows/test_workflow_routes.py
```diff
@@ -1017,6 +1049,14 @@ def test_get_workflow_by_event_id_200(
         assert "workflow_approval_config" in data
         assert isinstance(data["workflow_approval_config"], dict)
 
+        # Verify valid_events for MIDDLE state
+        assert "valid_events" in data
+        assert set(data["valid_events"]) == {
+            "middle_to_end",
```
hao10282025-sudo, line 1055: Current test case changes are okay.
I wonder if there should be more test cases for other state, like "PENDING_PROGRAM_OFFICER_APPROVAL", and the next valid events for it

kkrug: Makes sense, I added an approval state test

chouinar (approved):

Issue #9139, opened 2026-03-19
Backend - Create an endpoint to update reviews on an award recommendation


nathan-stricker:
### Summary

See [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Create-an-endpoint-to-update-reviews-on-an-award-recommendation)

#### Endpoint
`PUT /alpha/award-recommendations/:award_recommendation_id/reviews/:award_recommendation_review_id`
#### Purpose
Update a review
#### AuthZ
Requires `update_award_recommendation` privilege
#### Request
```json
{
  "is_reviewed": True
}
```
#### Response
```json
{
  "data": {
      # Note that the object here is the same as the award object returned by the GET award rec endpoint
      # should be able to reuse the model
      "award_recommendation_review_id": "<uuid>",
      "award_recommendation_review_type": "", # enum
      "is_reviewed": True,
    }
}
```
#### Logic

- Fetch the award recommendation, 404 if not found
- AuthZ, 403 on error
- Fetch the award recommendation review, 404 if not found
- Set the `is_reviewed` flag - this should be able to set it to true/false (this endpoint backs a checkbox on the frontend)

### Acceptance criteria

- [ ] Endpoint created
- [ ] Tests created/updated as necessary

PR #9643, opened 2026-04-16
[Issue #9139] Backend - Create an endpoint to update reviews on an award recommendation

kkrug:
## Summary

Fixes for #9139  

## Changes proposed

- Updated award_recommendation_schemas.py to add `AwardRecommendationReviewUpdateRequestSchema` and `AwardRecommendationReviewUpdateResponseSchema`
- Updated award_recommendation_routes.py to add the PUT route handler for the new endpoint
- Created update_award_recommendation_review.py to house the service layer for the new endpoint
- Created test_award_recommendation_review_update.py to test the new endpoint

## Context for reviewers

Basic update endpoint to update a review.

## Validation steps

1. Get JWT token of user with `update_award_recommendation`
2. Create an award recommendation - grab the `award_recommendation_id` and `award_recommendation_review_id` from the response
3. Call the new endpoint `PUT /alpha/award-recommendations/{award_recommendation_id}/reviews/{award_recommendation_review_id}` with `{"is_reviewed": true}` and verify a 200 response with the updated review object
4. Confirm tests pass

### DIFF
--- a/api/src/services/award_recommendations/update_award_recommendation_review.py
+++ b/api/src/services/award_recommendations/update_award_recommendation_review.py
@@ -1,0 +1,68 @@
+import logging
+import uuid
+
+import src.adapters.db as db
+from src.api.route_utils import raise_flask_error
+from src.auth.endpoint_access_util import verify_access
+from src.constants.lookup_constants import AwardRecommendationAuditEvent, Privilege
+from src.db.models.award_recommendation_models import (
+    AwardRecommendationAudit,
+    AwardRecommendationReview,
+)
+from src.db.models.user_models import User
+from src.services.award_recommendations.get_award_recommendation import (
+    get_award_recommendation_and_verify_access,
+)
+
+logger = logging.getLogger(__name__)
+
+
+def update_award_recommendation_review(
+    db_session: db.Session,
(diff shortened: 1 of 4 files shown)


api/src/services/award_recommendations/update_award_recommendation_review.py
```diff
@@ -0,0 +1,68 @@
...
+    # checks VIEW; we additionally need UPDATE)
+    agency = award_recommendation.opportunity.agency_record
+    verify_access(user, {Privilege.UPDATE_AWARD_RECOMMENDATION}, agency)
+
+    # Find the review on this award recommendation
+    review: AwardRecommendationReview | None = None
+    for r in award_recommendation.award_recommendation_reviews:
+        if r.award_recommendation_review_id == award_recommendation_review_id:
+            review = r
+            break
```
chouinar: Could we do a select query instead? It'll hit the in-memory SQLAlchemy cache (I think?), and be a bit more direct than iterating.

kkrug: Updated

chouinar (approved):

PR #9648, opened 2026-04-16
[Issue #9646] Fix Simpler soap error response

### LINKED ISSUE #9646: SOAP/Proxy: Fix ConfirmApplicationDelivery response to match Legacy
### Summary

Currently, the response from Simpler is a default that the certificate is not configured. This is returned even when a specific SOAPFaultException is raised, the proxy response overrides this. This can be confusing because it's not a certificate error, it's an issue like the Application is in the wrong status. This makes sure that the Simpler error response bubbles up if the `Use-Simpler-Override` is set.

### Acceptance criteria

- [ ] response matches the Legacy Response

jakobpederson:
## Summary

Fixes / Work for #9646  

## Changes proposed

Added a new exception case (`SOAPFaultException`) in the response.

## Context for reviewers

The error response was defaulting to the proxy legacy response, even when there was a simpler response that is more accurate. This modifies that to return the Simpler response when a `SOAPFaultException` is raised and the Simpler response is expected.

## Validation steps

- [ ] hit `ConfirmApplicationDelivery` with a `legacy_tracking_number` that's Simpler and in `in_progress` and observe response

### DIFF
--- a/api/src/legacy_soap_api/simpler_soap_api.py
+++ b/api/src/legacy_soap_api/simpler_soap_api.py
@@ -173,15 +173,13 @@ def process_simpler_request(
             },
         )
     except SOAPFaultException as e:
-        use_simpler = soap_request.headers.get("Use-Simpler-Override") == "1"
         logger.exception(
             msg=e.fault.faultstring,
             extra={
-                "used_simpler_response": use_simpler,
                 "soap_api_event": LegacySoapApiEvent.ERROR_CALLING_SIMPLER,
             },
         )
-        if use_simpler:
+        if soap_proxy_response.status_code == 500:
             return get_soap_error_response(
                 faultcode=e.fault.faultcode, faultstring=e.fault.faultstring
             ).to_flask_response()
--- a/api/src/legacy_soap_api/simpler_soap_api.py
+++ b/api/src/legacy_soap_api/simpler_soap_api.py
@@ -23,6 +23,7 @@ from src.legacy_soap_api.legacy_soap_api_schemas import (
 )
 from src.legacy_soap_api.legacy_soap_api_schemas.base import SOAPRequest, SoapRequestStreamer
 from src.legacy_soap_api.legacy_soap_api_utils import (
+    SOAPFaultException,
     get_alternate_proxy_response,
     get_invalid_path_response,
     get_soap_error_response,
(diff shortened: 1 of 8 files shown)


api/src/legacy_soap_api/simpler_soap_api.py
```diff
@@ -171,7 +172,17 @@ def process_simpler_request(
...
             },
         )
-        return soap_proxy_response.to_flask_response()
+    except SOAPFaultException as e:
+        logger.exception(
+            msg=e.fault.faultstring,
+            extra={
+                "soap_api_event": LegacySoapApiEvent.ERROR_CALLING_SIMPLER,
+            },
+        )
```
kkrug, line 182: Do we want to use `.exception()` here? `SOAPFaultException` here represents expected logic (wrong status, already retrieved, etc.), not an unexpected failure, so the stack traces could by noisy and it could trigger error alerts. We could us `logger.warning` or `logger.info` with `exc_info=True`.

api/tests/src/legacy_soap_api/test_legacy_soap_api_routes.py
```diff
@@ -104,6 +107,276 @@ def test_successful_confirm_application_delivery_request(
...
+    submission = ApplicationSubmissionFactory.create(
+        application=application, legacy_tracking_number="00837443"
+    )
+    full_path = "/grantsws-agency/services/v2/AgencyWebServicesSoapPort"
+    mock_data = (
+        "<soapenv:Envelope "
+        'xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/" '
+        'xmlns:agen="http://apply.grants.gov/services/AgencyWebServices-V2.0" '
+        'xmlns:gran="http://apply.grants.gov/system/GrantsCommonElements-V1.0">'
+        "<soapenv:Header/>"
```
kkrug, line 352: Should we explicitly set `Use-Simpler-Override: 1` for this test?

api/tests/src/legacy_soap_api/test_legacy_soap_api_routes.py
```diff
@@ -104,6 +107,276 @@ def test_successful_confirm_application_delivery_request(
...
+        "            <faultcode>soap:Server</faultcode>\n"
+        f"            <faultstring>Failed to confirm application delivery.(Expected an Application status of:'Validated' , but found a status of 'Received by Agency' for GRANT{submission.legacy_tracking_number})</faultstring>\n"
+        "        </soap:Fault>\n"
+        "    </soap:Body>\n"
+        "</soap:Envelope>\n"
+    )
+    assert response.data.decode() == expected
+
+
+def test_successful_confirm_application_delivery_request_when_in_tracking_number_assiged_status(
```
kkrug:
```suggestion
def test_successful_confirm_application_delivery_request_when_in_tracking_number_assigned_status(
```

kkrug (approved):

PR #9656, opened 2026-04-17
[Issue #9331] Opportunity Publishing Upload Endpoint

### LINKED ISSUE #9331: Opportunity Publishing Multiple Uploads - Endpoint
### Summary

Create API to support Multiple Upload functionality in Opportunity Publishing Page

### Acceptance criteria

- [ ] An endpoint exists that accepts an array of files, and adds them to the associated Opportunity
- [ ] If I call the endpoint while there are already files associated with the Opportunity, the newly passed files are added to the existing ones
- [ ] An endpoint exists that accepts an attachment ID and deletes it from the associated Opportunity

dghazvini:
## Summary

Fixes for #9331 

## Changes proposed

Create grantor API endpoints to support ~Multiple~ Upload functionality in Opportunity Publishing Page.

## Context for reviewers

2 Endpoints are included in this PR.

[Associated Tech Spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2979627009/Opportunity+Publishing+Multiple+Uploads+API+Endpoint)

`POST /v1/grantors/opportunities/:opportunity_id/attachments`
~An endpoint that accepts an array of files, and adds them to the associated Opportunity.~
**EDIT** - This endpoint will only support one file to be uploaded, not multiple.
An endpoint that accepts a file and adds it to the associated Opportunity.
 If I call the endpoint while there are already files associated with the Opportunity, the newly passed files are added to the existing ones.

`DELETE /v1/grantors/opportunities/:opportunity_id/attachments/:opportunity_attachment_id`
An endpoint that accepts an attachment ID along with an opportunity ID and deletes it from the associated Opportunity.

Both endpoints will have validations to check if the user has the `UPDATE_OPPORTUNITY` privilege for the Agency associated with the opportunity being passed.
Both endpoints will check that the opportunity is in draft status.

## Validation steps

- [x] Verified all expected responses on local Flask documentation API
- [x] Verified entries are being updated as expected in local DB
- [x] Unit Tests Added in `test_opportunity_attachments.py`


dghazvini:
@chouinar 
Hi Michael, PR is ready for full review, and just had a small callout-

In `upload_opportunity_attachments` in `opportunity_upload.py`:

Just wanted to call out I referenced the existing `attachment.util` functions to build the S3 URL/Validate the file names. I also referenced `file_util.py` on how to write the S3 Upload, but wasn't sure if there's an existing function I should be calling for that here instead. Please advise in particular if any edits in this section should be made. I believe everything is working as intended according to the tech spec - entries are being created/deleted in the `opportunity_attachment` table as expected.

The other thing is I know in the tech spec we do not explicitly call out valid file types and upload size limits; not sure if that's something that needs to explicitly be defined and constraints be set for. CC @glenblosser-nava @desaiamit

chouinar:
@dghazvini 

A major thing to be aware of - [file upload will significantly change in the future](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2948792998/Tech+Spec+File+Upload+for+Large+Files+Virus+Scanning) - that's not meant to be a blocker here, but I would say that multiple file upload isn't going to be supported. I'd actually recommend not having that functionality even now as there's a lot of edge cases we'd want to account for like what happens if one file uploads to s3 and the second fails? Since I don't intend that we'll support this going forward, I'd err on the side of single file upload.

Max file size is configured on the app itself (and with the eventual above work, will be in some of the shared utils). That's configured to 2gb right now. We don't have any file type restriction logic right now as there isn't really anything we would want to restrict outside of executables.

api/src/services/opportunities_grantor_v1/opportunity_upload.py
```diff
@@ -0,0 +1,132 @@
...
+            attachment_ids.append(str(attachment_id))
+
+            logger.info(
+                f"Added attachment {file_name} to opportunity",
+                extra={
+                    "opportunity_id": opportunity_id,
+                    "attachment_id": attachment_id,
+                    "file_size": file_size_bytes,
+                },
+            )
```
chouinar: Put any variables in the extra for logging.
```suggestion
            logger.info(
                "Added attachment to opportunity",
                extra={
                    "opportunity_id": opportunity_id,
                    "attachment_id": attachment_id,
                    "file_size": file_size_bytes,
                    "file_name": file_name
                },
            )
```

api/src/services/opportunities_grantor_v1/opportunity_upload.py
```diff
@@ -0,0 +1,132 @@
...
+    except Exception as e:
+        logger.error(f"Error uploading attachments: {str(e)}")
+        db_session.rollback()
+        raise
```
chouinar: This try/except isn't necessary. All exceptions that raise out of the API get logged (except 4xx ones from `raise_flask_error`). So long as you don't reach the end of `db_session.begin` without erroring, the results won't be committed, so a rollback isn't needed.

api/src/services/opportunities_grantor_v1/opportunity_upload.py
```diff
@@ -0,0 +1,132 @@
...
+            file_path = get_s3_attachment_path(
+                file_name=file_name,
+                opportunity_attachment_id=attachment_id,
+                opportunity=opportunity,
+                s3_config=s3_config,
+            )
+
+            # Write the file to S3
+            with file_util.open_stream(file_path, "wb", content_type=mime_type) as f:
+                f.write(file_content)
```
chouinar: Don't do `.read()` above - that reads the entire file into memory, we don't want to do that if it's a 1gb file. We want to just stream it through to s3.

Do this instead:

```py
with file_util.open_stream(...) as f:
    file_data.save(f)
```

Effectively rather than grabbing the data and writing it, tell the data to go to that file.

api/src/services/opportunities_grantor_v1/opportunity_upload.py
```diff
@@ -0,0 +1,132 @@
...
+
+            attachment = OpportunityAttachment(
+                attachment_id=attachment_id,
+                opportunity_id=opportunity_id,
+                file_location=file_path,
+                mime_type=mime_type,
+                file_name=file_name,
+                file_description=file_description,
+                file_size_bytes=file_size_bytes,
+                legacy_attachment_id=0,
```
chouinar: This should be null, we'll have to change the table first. Same as the other legacy IDs we've dealt with.

api/src/services/opportunities_grantor_v1/opportunity_upload.py
```diff
@@ -0,0 +1,132 @@
...
+            attachment_id = uuid.uuid4()
+
+            # Extract file metadata
+            if not file_data.filename:
+                raise_flask_error(422, "File must have a filename")
+
+            file_name = adjust_legacy_file_name(file_data.filename)
+            file_content = file_data.read()
+            mime_type = file_data.mimetype or "application/octet-stream"
+            file_description = ""
```
chouinar: Why isn't the file description being passed in? That's a field that the user should be able to populate / we show it on the opportunity page.

api/src/services/opportunities_grantor_v1/opportunity_upload.py
```diff
@@ -0,0 +1,132 @@
...
+            file_size_bytes = len(file_content)
```
chouinar: An easier way - after we write the file to s3, we have a utility you can call:

```py
file_util.get_file_length_bytes(file_path)
```

We basically ask s3 how big the file is rather than figure it out ourselves.

api/tests/src/api/opportunities_grantors_v1/test_opportunity_attachments.py
```diff
@@ -0,0 +1,315 @@
...
+    client, grantor_auth_data, existing_opportunity, mock_s3_bucket, other_mock_s3_bucket
+):
+    """Test successful upload of an attachment"""
+    _, _, token, _ = grantor_auth_data
+
+    # Create test file data
+    file_content = b"This is a test file content"
+    file_data = FileStorage(
+        stream=io.BytesIO(file_content), filename="test_file.pdf", content_type="application/pdf"
+    )
```
chouinar: It's fine either way, but you can avoid the Werkzeug FileStorage entirely by specifying the request as:

```py
data={"file_attachment": (BytesIO(b"whatever contents", "file_name.txt"))}
```

api/tests/src/api/opportunities_grantors_v1/test_opportunity_attachments.py
```diff
@@ -0,0 +1,315 @@
...
+
+    assert resp.status_code == 200
+    response_json = resp.get_json()
+    assert response_json["message"] == "Attachments uploaded successfully"
+
+    assert "data" in response_json
+    assert "opportunity_attachment_id" in response_json["data"]
+    assert isinstance(response_json["data"]["opportunity_attachment_id"], list)
+    assert len(response_json["data"]["opportunity_attachment_id"]) == 1
+
```
chouinar, line 82: Could you check that the file is in the DB and on s3?

We have a util for checking s3 easily - assuming you've fetched it from the DB:
```py
assert file_util.file_exists(opportunity_attachment.file_location) is True
```

brianlee-nava:
@chrisw-nava CC: @chouinar @dghazvini @glenblosser-nava @desaiamit Michael highlighted that the [way we handle multiple uploads will change moving forward](https://github.com/HHS/simpler-grants-gov/pull/9656#issuecomment-4281964685). Our Quad 5 deliverable AC3 says:

> The system supports a method for Grantors to upload accompanying documents (i.e. via API) in a format that can be used downstream within the GrantSolutions and SGG/SGM ecosystem for Grantors

Question: 
- The way its set up right now, we are uploading multiple attachments in one API call. Based on our understanding, we will NOT be able to support this functionality with the changes to the file upload soon. 
  - Should we change this right now to be an endpoint for a single upload OR keep it "as is"  and take on tech debt to change this in the future? The AC seems to be agnostic of the approach, but we want to confirm.

chouinar:
@brianlee-nava - if you keep it, I'm going to add half a dozen more things you need to account for in this implementation. I'd strongly recommend doing single file upload.

chrisw-nava:
> @brianlee-nava - if you keep it, I'm going to add half a dozen more things you need to account for in this implementation. I'd strongly recommend doing single file upload.

@brianlee-nava @chouinar the deliverable AC is not prescriptive for this reason and we should complete the technical implementation following the appropriate best practices and in line with our standards. Apply already has a pattern for accepting document uploads. **The most important thing we need to do** is demonstrate an ability to complete an upload by Friday's deadline for the AC review since this is an enhanced deliverable. If that's one upload at a time, that is fine.

dghazvini:
Updated the AC to reflect that this feature is for a single item upload.

api/src/services/opportunities_grantor_v1/opportunity_upload.py
dghazvini: Moved variable to extra.

api/src/services/opportunities_grantor_v1/opportunity_upload.py
dghazvini: Removed the unnecessary try except.

api/src/services/opportunities_grantor_v1/opportunity_upload.py
dghazvini: Implemented this streaming method.

api/src/services/opportunities_grantor_v1/opportunity_upload.py
dghazvini: Updated the `opportunity_attachment` table to set `legacy_attachment_id` to be nullable and ran migration and updated here to `None`.

api/src/services/opportunities_grantor_v1/opportunity_upload.py
dghazvini: Thanks for this callout - meant to implement but overlooked it since we didn't define this on the initial tech spec... 
I added `file_description` the request/response schema and logger and updated to pass it in here.

api/src/services/opportunities_grantor_v1/opportunity_upload.py
dghazvini: Implemented

api/tests/src/api/opportunities_grantors_v1/test_opportunity_attachments.py
dghazvini: Implemented

api/tests/src/api/opportunities_grantors_v1/test_opportunity_attachments.py
dghazvini: Implemented and updated corresponding unit tests

api/src/db/models/opportunity_models.py
```diff
@@ -481,7 +481,7 @@ class OpportunityAttachment(ApiSchemaTable, TimestampMixin):
 
     attachment_id: Mapped[uuid.UUID] = mapped_column(UUID, primary_key=True, default=uuid.uuid4)
 
-    legacy_attachment_id: Mapped[int] = mapped_column(BigInteger, index=True)
+    legacy_attachment_id: Mapped[int | None] = mapped_column(BigInteger, index=True, nullable=True)
```
chouinar: Don't specify nullable on the right side, just use the mapped type - specifying in both ways if they mismatch won't fail, so leads to weird outcomes (one takes precedence), better to only define one and avoid mismatch, and we have to type it anyways.

```suggestion
    legacy_attachment_id: Mapped[int | None] = mapped_column(BigInteger, index=True)
```

api/src/services/opportunities_grantor_v1/opportunity_upload.py
```diff
@@ -0,0 +1,126 @@
...
+            OpportunityAttachment.opportunity_id == opportunity_id,
+            OpportunityAttachment.attachment_id == opportunity_attachment_id,
+        )
+    ).scalar_one_or_none()
+
+    if not attachment:
+        raise_flask_error(404, "Attachment not found")
+
+    db_session.delete(attachment)
+
```
chouinar, line 123: We also need to delete the file from s3. Have a util for that.

```suggestion

    db_session.delete(attachment)
    file_util.delete_file(attachment.file_location)

```

api/src/db/migrations/versions/2026_04_21_merge_multiple_heads.py
```diff
@@ -0,0 +1,21 @@
+"""Merge multiple heads
```
chouinar: I'd remake these migrations so we don't have a merge one - we try to only use that when a mistake happens when merging to main. Can easily just fix it in the migration you generated (either undo them and remake it, and it'll be happy, or can try adjusting the down revision).

api/src/services/opportunities_grantor_v1/opportunity_upload.py
dghazvini: Added

api/src/db/models/opportunity_models.py
dghazvini: Updated

api/src/db/migrations/versions/2026_04_21_merge_multiple_heads.py
dghazvini: Removed this file and the previous nullable script and remade a new one.

chouinar (approved):

Issue #9311, opened 2026-03-30
Wireup Get AwardRecommendation with existing front end


chay-REIsys:
### Summary

Remove mock data and call api to fetch award recommendation
Write/Update Unit Tests

### Acceptance criteria

See [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Create-the-GET-award-recommendation-endpoint)

GET /alpha/award-recommendations/:award_recommendation_id

See [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Award-Recommendation.1) for JSON response

1. Remove the existing Mock Api response
2. Replace with the above get award recommendation endpoint and ensure existing front end works as expected
3. Make sure unit tests pass and add new test cases for the api fetcher

KtejanKella:
Nathan PR reviewer

PR #9659, opened 2026-04-17
[Issue#9311]Get award recommendation endpoint Integration with UI

chay-REIsys:
## Summary

 Work for #9311  

## Changes proposed

Integrating front end with backend get award recommendation endpoint
Adding missing funding_strategy column to award recommendation table
## Context for reviewers

As of now data is mocked on the frontend but these changes are to communicate with get award recommendation endpoint and load UI dynamically

## Validation steps

- Fetch Changes and do npm run build on frontend and make remake-backend && docker compose up on backend
- try to access  http://localhost:3000/award-recommendation/22f926c6-cd63-4628-b88d-40b4ba0ea640 and see not authenticated message
- signin using ar_rec_user1 or ar_rec_user2 (configured access to view award recommendation pages)
- fetch any award recommendation id from award recommendation table (you should see some recors in table when you run  make remake-backend && docker compose up on backend as part of seed script)
- after login with either of the above user try to access  http://localhost:3000/award-recommendation/awardRecommendationId replace awardRecommendationId with id from award recommendation table and observe data loads dynamically upon changing id 
- click on edit button and observe redirecting to edit page where you can see the data load dynamically


frontend/src/components/award-recommendation/AwardRecommendationHero.tsx
```diff
@@ -37,7 +37,14 @@ export default async function AwardRecommendationHero({
   buttons,
 }: AwardRecommendationHeroProps) {
   const t = await getTranslations("AwardRecommendation");
-  const { recordNumber, datePrepared, status } = awardRecommendationDetails;
+
+  const number = awardRecommendationDetails.award_recommendation_number;
```
nathan-stricker: nit: `number` is a bit generic for a variable name, maybe change it to something like `awardRecNumber` to make its meaning a bit more explicit

frontend/src/components/award-recommendation/AwardRecommendationStatusTag.tsx
```diff
@@ -23,24 +23,14 @@ const AwardRecommendationStatusTag = ({ status }: Props) => {
...
-          <div
-            className="usa-tag bg-accent-warm-light text-ink radius-2 font-sans-sm text-no-uppercase display-flex flex-align-center"
-            data-testid="award-recommendation-status-in-progress"
-          >
-            <USWDSIcon name="schedule" className="margin-right-05" />
-            {t("inProgress")}
-          </div>
-        );
-      case "pendingReview":
+      case "in_review":
```
nathan-stricker, line 26: Are we intentionally combining the inProgress and pendingReview statuses here? I was checking the designs to see if we had removed one of them but didn't see anything: https://www.figma.com/design/o5g0d2W5RKHYHjdvVCqTV2/Grantor-Designs-%7C-Quad-5-%7C-Simpler-Grants-2026?node-id=5195-161054&m=dev

api/src/api/award_recommendations_alpha/award_recommendation_schemas.py
```diff
@@ -159,6 +159,13 @@ class AwardRecommendationDataSchema(Schema):
             "example": "Selection factors included technical merit, past performance, and cost.",
         },
     )
+    funding_strategy = fields.String(
+        allow_none=True,
+        metadata={
+            "description": "Funding strategy for award recommendations",
+            "example": "Full funding for top 10 applications, partial funding for next 15 based on available budget.",
+        },
+    )
```
nathan-stricker, line 168: I think it would be ideal to add the `funding_strategy` column to the table in its own PR, but this is probably fine for now since we're approaching the end of the quad and this is a bit time sensitive. This should at least be called out in your "Changes proposed" section since it's not directly tied with the integration

frontend/src/components/award-recommendation/AwardRecommendationStatusTag.tsx
chay-REIsys: i saw my local db  and i see only 3 status
image

nathan-stricker: Ah I see- I think this is the correct mapping:

- draft -> inProgress
- in_review -> pendingReview
- approved -> approved

So in that case, we should be removing the separate "draft" status tag, instead of merging inProgress/pendingReview

frontend/src/i18n/messages/en/index.ts
```diff
@@ -2044,9 +2044,8 @@ export const messages = {
     datePrepared: "Date prepared",
     status: "Status",
     statusTag: {
-      draft: "Draft",
-      inProgress: "In progress",
-      pendingReview: "Pending review",
+      draft: "In Progress",
```
nathan-stricker, line 2047: This is good, but I believe you also need to tweak `AwardRecommendationStatusTag.tsx` to replace the deleted "in progress" tag styling with the variable name "draft", and delete the old draft tag, that way the three tags remaining match the designs

chay-REIsys: got it

frontend/src/services/fetch/fetchers/awardRecommendationFetcher.ts
```diff
@@ -1,20 +1,23 @@
...
 
-  // TODO: Replace mock data above with this once the backend endpoint exists.
-  // const response = await fetchAwardRecommendation({ subPath: id });
-  // const responseBody = (await response.json()) as { data: AwardRecommendationDetails };
-
-  // return responseBody.data;
+  return {
+    ...apiData,
+    award_recommendation_summary: apiData.award_recommendation_summary || {
+      total_received_count: 200,
```
doug-s-nava, line 16: is this hardcoded data for demo purposes? Can we remove it for now? What is the expected behavior when there is no `award_recommendation_summary`?

doug-s-nava (commented):
just one small question from me

frontend/src/services/fetch/fetchers/awardRecommendationFetcher.ts
chay-REIsys: this is as part of low priority endpoint ticket which is yet to be accomplished and eventually as part of that ticket we will remove this. @doug-s-nava

chay-REIsys: https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Lower-Priority-Tickets

chay-REIsys: i don't think there will be a case with no award_recommendation_summary as it has some fields which has some value always such as recommended amount etc

nathan-stricker (approved):
LGTM!

Issue #8551, opened 2026-02-17
fix css warnings post next 16 upgrade


doug-s-nava:
### Summary

after the sass and tooling upgrades necessary for the next 16 upgrade, some css warnings have begun appearing. full summary of warnings below

the underlying issues behind these warnings should all be resolved

```

./src/styles/styles.scss.webpack[javascript/auto]!=!./node_modules/next/dist/build/webpack/loaders/css-loader/src/index.js??ruleSet[1].rules[14].oneOf[11].use[2]!./node_modules/next/dist/build/webpack/loaders/postcss-loader/src/index.js??ruleSet[1].rules[14].oneOf[11].use[3]!./node_modules/next/dist/build/webpack/loaders/resolve-url-loader/index.js??ruleSet[1].rules[14].oneOf[11].use[4]!./node_modules/next/dist/compiled/sass-loader/cjs.js??ruleSet[1].rules[14].oneOf[11].use[5]!./src/styles/styles.scss
Deprecation Warning on line 12, column 24 of file:///Users/dougschrashun/code/simpler-grants-gov/frontend/node_modules/@uswds/uswds/packages/uswds-core/src/styles/functions/general/error-not-token.scss:12:24:
The Sass if() syntax is deprecated in favor of the modern CSS syntax.

Suggestion: if(sass($valid-token-map): " Valid tokens: #{map.keys($valid-token-map)}"; else: "")

More info: https://sass-lang.com/d/if-function

12 |   $valid-token-message: if(
    $valid-token-map,
    " Valid tokens: #{map.keys($valid-token-map)}",
    ""
  );

node_modules/@uswds/uswds/packages/uswds-core/src/styles/functions/general/error-not-token.scss 13:25  @forward
node_modules/@uswds/uswds/packages/uswds-core/src/styles/functions/general/_index.scss 3:1             @use
node_modules/@uswds/uswds/packages/uswds-core/src/styles/functions/utilities/color.scss 14:1           @use
node_modules/@uswds/uswds/packages/uswds-core/src/styles/variables/border-high-contrast.scss 2:1       @forward
node_modules/@uswds/uswds/packages/uswds-core/src/styles/variables/_index.scss 2:1                     @forward
node_modules/@uswds/uswds/packages/uswds-core/src/styles/_index.scss 2:1                               @forward
node_modules/@uswds/uswds/packages/uswds-core/_index.scss 1:1                                          @use
[...]

PR #9663, opened 2026-04-17
[Issue #8551] silence uswds sass warnings

doug-s-nava:
## Summary

Fixes for #8551 

## Changes proposed

Silences sass warnings from uswds packages that started appearing during next 16 upgrade. 

## Context for reviewers

Since these warnings are coming from the uswds package, we are ignoring them rather than fixing them. Based on the warnings, it sounds like these will become errors at some point in the future, but we can deal with that when and if it happens

## Validation steps

1. `npm run local` on this branch
2. load a page
3. _VERIFY_: no sass warnings in next js logs

### DIFF
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -163,6 +163,7 @@ const nextConfig = {
   // https://nextjs.org/docs/app/api-reference/next-config-js/output
   output: "standalone",
   sassOptions: {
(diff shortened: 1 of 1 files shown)


Dao-REI (approved):
Verified. 

Before: 
Image

After:
Image

Issue #9146, opened 2026-03-19
Backend - Create an endpoint to get the audit history of an award recommendation


nathan-stricker:
### Summary

See [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Create-an-endpoint-to-get-the-audit-history-of-an-award-recommendation)

#### Endpoint
`POST /alpha/award-recommendations/:award_recommendation_id/audit_history`

#### Purpose
Get the audit history of an award recommendation

#### AuthZ
Requires `view_award_recommendation` privilege

#### Request
```json
{
  # Filters are optional but frontend'll like it if there are certain events they haven't yet built support for
  "filters": {
    "award_recommendation_audit_event": {
      "one_of": ["", ""], # The award recommendation audit event type enum
    }
  },
  "pagination": {
    "page_offset": 1,
    "page_size": 25,
    "sort_order": [
      {
        # Can sort by just created_at at the moment
        "order_by": "created_at",
        "sort_direction": "ascending"
      }
    ]
  }
}
```

#### Response
```json
{
  "data": [
    {
      "award_recommendation_audit_id": "<uuid>",
      "award_recommendation_audit_event": "", # The award recommendation audit event type enum
      "user": {
        "user_id": "",
        "email": "",
        "first_name": "",
        "last_name": "",
      },
      # Note that many of these will only be set for specific audit events
      # as these are the thing that the event affected (eg. an attachment event will have an attachment)
      # These do not contain all fields, we only put enough for display purposes in a table.
      # So that means no nested lists or actual files.
      "award_recommendation_risk": {
        "award_recommendation_risk_id": "<uuid>",
        "award_recommendation_risk_type_id": "", # enum
        "award_recommendation_risk_number": "",
        "is_deleted": True,
      },
      "award_recommendation_attachment": {
        "award_recommendation_attachment_id": "<uuid>",
        "file_name": "",
        "award_recommendation_attachment_type": "", # Enum
        "is_deleted": True,
      },
      "award_recommendation_review": {
        "award_recommendation_review_id": "<uuid>",
        "is_reviewed": True,
      },
      "award_recommendation_application_submission": {
        "award_recommendation_application_submission_id": "<uuid>",
        "application_submission_id": "<uuid>",
        "application_submission_number": "",
      },
      "workflow_approval": {
        "workflow_approval_id": "<uuid>",
[...]

PR #9679, opened 2026-04-20
[Issue #9146] Backend - Create an endpoint to get the audit history of an award recommendation

kkrug:
## Summary

Fixes for #9146 

## Changes proposed

- Updated award_recommendation_schemas.py to add the request/response schemas
- Updated award_recommendation_routes.py to add the POST route handler for the new endpoint
- Created list_award_recommendation_audit.py to house the service layer for the new endpoint
- Created test_award_recommendation_audit_list.py to test the new endpoint

## Context for reviewers

Endpoint to get the audit history of an award recommendation.

## Validation steps

1. Get JWT token of user with `view_award_recommendation`
2. Create an award recommendation if needed - grab the `award_recommendation_id` from the response
3. Call the new endpoint `POST /alpha/award-recommendations/{award_recommendation_id}/audit_history` with:
```
{
  "pagination": {
    "page_offset": 1,
    "page_size": 25
  }
}
```
and verify a 200 response with at least the `award_recommendation_created` audit event
5. Test filtering by re-running with filters:
```
{
  "filters": {
    "award_recommendation_audit_event": {
      "one_of": ["award_recommendation_created"]
    }
  },
  "pagination": { "page_offset": 1, "page_size": 25 }
}
```
6. Confirm tests pass

### DIFF
--- a/api/src/api/award_recommendations_alpha/award_recommendation_schemas.py
+++ b/api/src/api/award_recommendations_alpha/award_recommendation_schemas.py
@@ -5,7 +5,10 @@ from src.api.schemas.response_schema import AbstractResponseSchema, PaginationMi
 from src.api.schemas.search_schema import BoolSearchSchemaBuilder, StrSearchSchemaBuilder
 from src.api.schemas.shared_schema import SimpleUserSchema
 from src.constants.lookup_constants import (
+    ApprovalResponseType,
+    ApprovalType,
     AwardRecommendationAttachmentType,
+    AwardRecommendationAuditEvent,
     AwardRecommendationReviewType,
     AwardRecommendationRiskType,
     AwardRecommendationStatus,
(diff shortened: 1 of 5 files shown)


api/src/api/award_recommendations_alpha/award_recommendation_schemas.py
```diff
@@ -343,3 +347,169 @@ class AwardRecommendationSubmissionListResponseSchema(
...
+        generate_pagination_schema(
+            "AwardRecAuditPaginationSchema",
+            ["created_at"],
+            default_sort_order=[{"order_by": "created_at", "sort_direction": "descending"}],
+        ),
+        required=True,
+    )
+
+
+class AwardRecommendationAuditRiskSchema(Schema):
```
chouinar, line 385: You'll want to coordinate with @nathan-stricker - who added some of these schemas in his latest PR - they'll be the same thing most likely, just make sure when the two are merged, we don't lose anything.

api/src/services/award_recommendations/list_award_recommendation_audit.py
```diff
@@ -0,0 +1,132 @@
...
+        # Transform the workflow approval if present
+        if approval := audit_event.workflow_approval:
+            transformed_approval = {
+                "workflow_approval_id": approval.workflow_approval_id,
+                "workflow_id": approval.workflow_id,
+                "approval_type": approval.approval_type,
+                "approval_response_type": approval.approval_response_type,
+            }
+        else:
+            transformed_approval = None
```
chouinar: What are these transformations doing? I think the approval one is the same shape as far as I can tell and for the submission, we could handle this by adding this to the award recommendation application submission model:

```py
@property
def application_submission_number(self) -> str | None:
     return self.application_submission.application_submission_number
```

kkrug: Updated

kkrug:
@nathan-stricker I merged your recent changes into this branch to avoid conflicts if you wanted to ensure that everything looks correct

chouinar (approved):

PR #9681, opened 2026-04-20
[Issue #9680] Move alert errors to verifyFormStatusAfterSave

### LINKED ISSUE #9680: Fix E2E test failure due to shared alert vs inline error list assumption
### Summary

Fixes a failing E2E test where it is assumed that there is a 1:1 mapping between alert banner errors and inline field errors.

For few forms (e.g., SF-424A), this assumption does not hold:
- The alert banner includes both array-level and field-level validation errors.
- Inline validation only applies to fields with stable, mappable UI elements.

This mismatch led to incorrect validation when a single shared error list was used for both checks.

The scope of this ticket ensures that alert and inline errors are handled appropriately, aligning test expectations with actual form behavior.

### Acceptance criteria

- The E2E test passes consistently without assertion mismatches.
- Tests clearly support divergent error lists (either via separate parameters or explicit utility calls).
- No regressions in other failure-path or happy-path E2E tests.

Bhavna-Ramachandran:
## Summary
Work for: Task #9680 

## Changes proposed
Remove standalone verifyAlertErrors call from failure-path-sf424a.spec.ts
Add optional alertErrors parameter - verifyFormStatusAfterSave in verify-form-status-utils.ts

## Context for reviewers
The structure allows standalone Playwright tests to run independently and in CI

Note: This test failure-path-sf424a.spec.ts previously passed in both local and staging CI runs, however the recent scheduled job run exposed this failure

## Validation steps
Run the standalone Playwright tests and they should complete the workflow without any issues

### DIFF
--- a/frontend/tests/e2e/apply/failure-path-sf424a.spec.ts
+++ b/frontend/tests/e2e/apply/failure-path-sf424a.spec.ts
@@ -14,7 +14,6 @@ import { authenticateE2eUser } from "tests/e2e/utils/authenticate-e2e-user-utils
 import { createApplication } from "tests/e2e/utils/create-application-utils";
 import { openForm } from "tests/e2e/utils/forms/form-navigation-utils";
 import { saveForm } from "tests/e2e/utils/forms/save-form-utils";
-import { verifyAlertErrors } from "tests/e2e/utils/forms/verify-form-errors-utils";
 import {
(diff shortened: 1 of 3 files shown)


doug-s-nava (approved):
✅ ✅ ✅ ✅

Issue #9141, opened 2026-03-19
Backend - Create an endpoint to create a risk on an award recommendation


nathan-stricker:
### Summary

See [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Create-an-endpoint-to-create-a-risk-on-an-award-recommendation)

#### Endpoint
`POST /alpha/award-recommendations/:award_recommendation_id/risks`
#### Purpose
Create an award recommendation risk
#### AuthZ
Requires `update_award_recommendation` privilege
#### Request
```json
{
  "summary": "",
  "award_recommendation_risk_type": "", # Enum
  "award_recommendation_application_submission_id": ["<uuid>", "<uuid>"]
}
```
#### Response
```json
{
  "data": {
    "award_recommendation_risk_id": "<uuid>",
    "summary": "",
    "award_recommendation_risk_number": "",
    "award_recommendation_risk_type": "", # Enum
    "award_recommendation_application_submission_id": ["<uuid>", "<uuid>"]
  }
}
```
#### Logic

- Fetch the award recommendation, 404 if not found
- AuthZ, 403 on error
- Fetch all the award recommendation application submissions with the given IDs
    - 404 if any are not found
    - 404 if any are not attached to this award recommendation (just write the query to join to the award rec table)
- Create the award recommendation risk, linking all the fetched submissions to it

### Acceptance criteria

- [ ] Endpoint created
- [ ] Tests created/updated as necessary

PR #9684, opened 2026-04-20
[Issue #9141] Add endpoint to create a risk on an award recommendation

nathan-stricker:
## Summary

Work for #9141 

## Changes proposed

Created an endpoint to create a risk on an Award Recommendation and added associated tests

## Context for reviewers

See [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Risks)

## Validation steps

- [ ] Confirm implementation aligns with specifications outlined in the [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Risks) (note that I used the variable name `comment` instead of `summary` in the request/response to remain consistent with the data model)
- [ ] Review created tests


api/src/services/award_recommendations/create_award_recommendation_risk.py
```diff
@@ -0,0 +1,138 @@
...
+    user: User,
+    award_recommendation_id: uuid.UUID,
+    request_data: dict,
+) -> dict:
+    award_recommendation = _get_award_recommendation_for_update(
+        db_session, user, award_recommendation_id
+    )
+
+    agency = award_recommendation.opportunity.agency_record
+    assert agency is not None  # Guaranteed by _get_award_recommendation_for_update
```
chouinar: Don't use asserts outside of tests, asserts are ignored when python runs in optimized mode - which we do have enabled in some capacity.

Do a basic check instead:

```py
if agency is None:
    raise Exception("...")
```

api/src/services/award_recommendations/create_award_recommendation_risk.py
```diff
@@ -0,0 +1,138 @@
...
+    return {
+        "award_recommendation_risk_id": risk.award_recommendation_risk_id,
+        "comment": risk.comment,
+        "award_recommendation_risk_number": risk.award_recommendation_risk_number,
+        "award_recommendation_risk_type": risk.award_recommendation_risk_type,
+        "award_recommendation_application_submission_ids": [
+            rs.award_recommendation_application_submission_id
+            for rs in risk.award_recommendation_risk_submissions
+        ],
+    }
```
chouinar: Return the risk itself instead of making an object since they're the same shape. A trick for getting values like that list of IDs - on the risk class, add a property like:

```py
@property
def award_recommendation_application_submission_ids(self) -> list[uuid]:
     return [rs.award_recommendation_application_submission_id for rs in self.award_recommendation_risk_submissions]
```

api/src/api/award_recommendations_alpha/award_recommendation_schemas.py
```diff
@@ -343,3 +344,57 @@ class AwardRecommendationSubmissionListResponseSchema(
...
     )
+
+
+class AwardRecommendationRiskCreateRequestSchema(Schema):
+    """Schema for POST /alpha/award-recommendations/:award_recommendation_id/risks request"""
+
+    comment = fields.String(
+        required=True,
+        metadata={"description": "Summary of the risk"},
+    )
```
chouinar, line 383: Could you add an example to any of these string fields so that they don't just have the default. Other field types are fine since they get reasonable defaults (an enum value, a uuid, etc.)

api/src/services/award_recommendations/create_award_recommendation_risk.py
```diff
@@ -0,0 +1,138 @@
...
+    )
+
+    submissions = list(db_session.execute(stmt).scalars().all())
+
+    if len(submissions) != len(submission_ids):
+        found_ids = {s.award_recommendation_application_submission_id for s in submissions}
+        missing_ids = [sid for sid in submission_ids if sid not in found_ids]
+        raise_flask_error(
+            404,
+            message=f"Could not find Award Recommendation Application Submissions: {missing_ids}",
```
chouinar: We'll need to format this in a way that would be machine readable.

For that we could add a list of validation error details like:

```py
missing_ids = ...

validation_errors = []
for missing_id in missing_ids:
    validation_errors.append(ValidationErrorDetail(type=ValidationErrorType.APPLICATION_SUBMISSION_NOT_FOUND, message="...", field="award_recommendation_application_submission_ids", value=missing_id))

raise_flask_error(404, message="...", validation_issues=validation_errors)
```

Any validation error details will be put in the error section of the response schema.

api/src/services/award_recommendations/create_award_recommendation_risk.py
```diff
@@ -0,0 +1,138 @@
...
+
+    for submission in submissions:
+        risk.award_recommendation_risk_submissions.append(
+            AwardRecommendationRiskSubmission(
+                award_recommendation_application_submission=submission,
+            )
+        )
+
+    db_session.add(risk)
+    db_session.flush()
```
chouinar: If you want to avoid a flush to get the ID, just set it when we create it above. That'll save like 5ms on the API call. Not required, but simpler.

api/src/services/award_recommendations/create_award_recommendation_risk.py
```diff
@@ -0,0 +1,138 @@
...
+        raise_flask_error(403, message="Forbidden")
+
+    verify_access(user, {Privilege.UPDATE_AWARD_RECOMMENDATION}, agency)
+
+    return award_recommendation
+
+
+def _generate_risk_number(db_session: db.Session, agency_code: str) -> str:
+    if not agency_code:
+        raise_flask_error(403, message="Forbidden")
```
chouinar: This shouldn't happen with your other checks - if it did happen I think we'd want to 500 as that's probably a bug in our system at that point.

api/src/services/award_recommendations/create_award_recommendation_risk.py
```diff
@@ -0,0 +1,145 @@
...
+            submission_ids
+        ),
+        AwardRecommendationApplicationSubmission.award_recommendation_id == award_recommendation_id,
+    )
+
+    submissions = list(db_session.execute(stmt).scalars().all())
+
+    if len(submissions) != len(submission_ids):
+        found_ids = {s.award_recommendation_application_submission_id for s in submissions}
+        missing_ids = [sid for sid in submission_ids if sid not in found_ids]
```
kkrug, line 91: If two of the same `submission_ids` are passed in, the query will return one row, but `len(submission_ids)` will be 2. We can deduplicate the `submission_ids` before the query with something like:
`submission_ids = list(set(request_data["award_recommendation_application_submission_ids"]))`

api/src/api/award_recommendations_alpha/award_recommendation_schemas.py
```diff
@@ -343,3 +344,63 @@ class AwardRecommendationSubmissionListResponseSchema(
...
+        required=True,
+        metadata={"description": "The type of risk"},
+    )
+    award_recommendation_application_submission_ids = fields.List(
+        fields.UUID(),
+        required=True,
+        metadata={
+            "description": "List of award recommendation application submission IDs to link to this risk"
+        },
+    )
```
kkrug, line 396: Should it be possible to have a risk with zero linked submissions? If not, we can add validation here to prevent that:
```suggestion
    award_recommendation_application_submission_ids = fields.List(
        fields.UUID(),
        required=True,
        validate=validate.Length(min=1),
        metadata={
            "description": "List of award recommendation application submission IDs to link to this risk"
        },
    )
```

api/src/api/award_recommendations_alpha/award_recommendation_routes.py
```diff
@@ -116,3 +121,30 @@ def award_recommendation_submission_list(
...
+
+    with db_session.begin():
+        user = jwt_or_api_user_key_multi_auth.get_user()
+        db_session.add(user)
+
+        risk = create_award_recommendation_risk(
+            db_session, user, award_recommendation_id, json_data
+        )
+
+    return response.ApiResponse(message="Success", data=risk)
```
kkrug, line 198: Should this `return` statement be within the `with db_session.begin()` block?

chouinar: We don't do that because the end of the begin block does the commit. In the event there is an issue with the commit, we would want the API to return a 500. I'm not actually sure the order of operations of returns/exits of context managers, so it might work, but better to be clear/safe

api/src/services/award_recommendations/create_award_recommendation_risk.py
```diff
@@ -0,0 +1,145 @@
...
+        candidate = agency_code + "".join(secrets.choice(alphabet) for _ in range(10))
+        already_exists = db_session.execute(
+            select(
+                exists().where(
+                    AwardRecommendationRisk.award_recommendation_risk_number == candidate
+                )
+            )
+        ).scalar()
+        if not already_exists:
+            return candidate
```
chouinar: Just in case, can we add a maximum iteration count? It _shouldn't_ matter, but just in case some weird edge case in randomness gets hit, best to be safe. I can't imagine this iterates more than once 99.9% of the time, so setting it to 5 iterations is probably more than safe.

api/src/services/award_recommendations/create_award_recommendation_risk.py
chouinar: We've hit this a few times, should really add a set concept to the API schema, but not here.

But I'd say just make them a set since I don't think we need a list. Just need an iterable.

nathan-stricker: Good catch! Fixing this

api/src/api/award_recommendations_alpha/award_recommendation_routes.py
nathan-stricker: ~I was checking a few of our other routes and it seems like we usually have `return` statements outside the block, but there are a few exceptions. I don't have a strong opinion either way, but having it outside matches the pattern used by the other routes in this file- happy to move it inside if that improves clarity though~
Edit: just saw Michael's comment, ignore this ^

kkrug (approved):

PR #9686, opened 2026-04-20
[Issue #9669] Enable backend call for Publish button

### LINKED ISSUE #9669: Enable backend call for Publish button
### Summary

Metric 1 for #7806 

> At least 3 Opportunities are published in the training environment

- At 4/17 standup, the team highlighted a gap to fulfill this metric. We agreed that when a user clicks "Publish", isDraft flag should be turned to false. A backend call still needs to be implemented.

### Acceptance criteria

- [ ] When "Publish" button is clicked, make backend call to make "isDraft = false"

dghazvini:
## Summary

Fixes for #9669

## Changes proposed

Create an endpoint to queue up an Opportunity to be set into the Published status.

## Context for reviewers

As part of the Opportunity Publishing workflow, the Opportunity table needs to have the `is_draft` column toggled to false to signify it has been published. This will be handled by this endpoint, which will send a message to the SQS queue that will in turn be handled by the workflow service to set the proper flag.

## Validation steps

- [x] Confirmed expected responses on Flask API for queuing into SQS queue
- [x] After running workflow service locally, confirmed that the opportunity is no longer in draft status.
- [x] 5 Unit Tests for expected results


api/src/services/opportunities_grantor_v1/publish_opportunity.py
```diff
@@ -0,0 +1,48 @@
...
+    if not opportunity.is_draft:
+        from src.api.route_utils import raise_flask_error
+
+        raise_flask_error(422, message="Opportunity is already published")
+
+    # Not Needed?
+    # opportunity.is_draft = False
+
+    # TODO Logic for queueing Opportunity to be published goes here.
+
```
chouinar, line 57: All you have to do is queue up the event:

```py
event_id = uuid.uuid4()
send_workflow_event_to_queue(
    WorkflowEvent(event_id=event_id, acting_user_id=user.user_id, event_type=WorkflowEventType.START_WORKFLOW, start_workflow_context=StartWorkflowEventContext(workflow_type=WorkflowType.OPPORTUNITY_PUBLISH, entity_type=WorkflowEntityType.OPPORTUNITY, entity_id=opportunity.opportunity_id)
)
```

This will put a message in the SQS queue that our workflow service listens to. The workflow service runs automatically (non-locally) every few seconds, so this'll process near instantly. https://github.com/HHS/simpler-grants-gov/blob/main/api/src/workflow/state_machine/opportunity_publish_state_machine.py has the logic, but it'll handle flipping the flag, calculating the opportunity status, and putting it in search.

For testing this, you can run the workflow service locally by doing `make cmd args="workflow workflow-main"` which'll spin up the service and start processing events, it'll process anything sitting in the queue as soon as it starts up. For unit tests, you can make the service run by doing:
```py
    with app.app_context():
        WorkflowManager().process_batch()
```

See for example how we test this https://github.com/HHS/simpler-grants-gov/blob/main/api/tests/src/workflow/integration/test_workflow_integration.py - you can get the app from a fixture.

dghazvini: Thanks! I was able to get it working! Opening PR for full review.

api/src/services/opportunities_grantor_v1/publish_opportunity.py
```diff
@@ -0,0 +1,63 @@
...
+    # Check if user has permission to publish opportunities for this agency
+    verify_access(user, {Privilege.PUBLISH_OPPORTUNITY}, opportunity.agency_record)
+
+    # Validate that the opportunity was created in Simpler Grants
+    validate_opportunity_created_in_simpler_grants(opportunity)
+
+    # Check if the opportunity is already published
+    if not opportunity.is_draft:
+        from src.api.route_utils import raise_flask_error
```
chouinar: Put all imports at the top of the file

api/tests/src/api/opportunities_grantors_v1/test_opportunity_publish.py
```diff
@@ -0,0 +1,186 @@
...
+    UserProfileFactory.create(user=user, first_name="System", last_name="User")
+
+    role = RoleFactory.create(
+        privileges=[Privilege.INTERNAL_WORKFLOW_ACCESS], role_types=[RoleType.INTERNAL]
+    )
+    InternalUserRoleFactory.create(user=user, role=role)
+
+    monkeypatch.setenv("WORKFLOW_SERVICE_INTERNAL_USER_ID", str(user.user_id))
+
+    return user
```
chouinar: Rather than add this here, I think we should move the `workflow_user` fixture in the workflow conftest to the top-level conftest. If we need it in more places, it's fine to move up.

However, the reason I hadn't done that is because I wanted it to be marked as `autouse=True` and we don't want that on every test. I'd move it up to the top-level conftest without that. Then where the function currently exists just do:

```
@pytest.fixture(autouse=True)
def workflow_user_init(workflow_user):
     pass
```

Effectively making it autoused where it is now without globally doing it.

api/tests/src/api/opportunities_grantors_v1/test_opportunity_publish.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+@pytest.fixture(scope="session")
+def workflow_client_registry(search_client):
+    """Initialize the workflow client registry"""
+    from src.workflow.registry.workflow_client_registry import init_workflow_client_registry
+
+    return init_workflow_client_registry(search_client=search_client)
```
chouinar: Rather than copy this, also move this function up to the top-level conftest file.

chouinar (commented):
Your unit test failures are the fault of some pretty complicated search index setup that a proper fix is going to be a bit of effort to sort out. As I know this has some urgency, here's a workaround. Add this to the top of your new test file: 

```py
@pytest.fixture(scope="module")
def publish_index(search_client):
    # create a random index name just to make sure it won't ever conflict
    # with an actual one, similar to how we create schemas for database tests
    index_name = f"test-opportunity-index-{uuid.uuid4().int}"

    search_client.create_index(index_name)

    try:
        yield index_name
    finally:
        # Try to clean up the index at the end
        # Use a prefix which will delete the above (if it exists)
        # and any that might not have been cleaned up due to issues
        # in prior runs
        search_client.delete_index("test-opportunity-index-*")

@pytest.fixture(scope="module", autouse=True)
def setup_index(
    search_client, publish_index, opportunity_index_alias
):
    """Setup the index - making sure the alias is set"""
    search_client.swap_alias_index(publish_index, opportunity_index_alias)
```

The short version of the problem, various tests interact with the search index. Sometimes they try to recreate the search index - and rename it to itself which it doesn't like. We just so happen to not have this issue elsewhere due to luck, but the order tests run in with this change ends up causing issues (not at all the fault of what you've written). This works around that by just making a different index name for this file. I didn't run all the tests, but I think this'll fix it?

dghazvini:
@chouinar Thank you for the quick turnaround and analysis for this fix! I implemented those changes as you mentioned and looks like all tests are passing now.

chouinar (approved):

PR #9652, opened 2026-04-17
[Issue #9551] Resolved all broken internal links.

MozamilS:
## Summary

A total of more than 140 internal links were found with lychee (excluding external links). Each broken link was investigated to find the appropriate remedy. The majority of links were repaired since they were only incorrectly formatted. Some links were clearly referencing to external resources but had internal links instead and those were replaced with external links. Some internal links did not exist anymore and they were removed. 

There was only one instance where a directory was created to rectify the broken link. That was in `documentation/wiki/product/decisions/README.md`. The context was moving deprecated ADRs to a decisions/deprecated directory. It did not make sense to remove this link along with its context and making this directory seemed to be the appropriate step instead. 

Fixes #9551 

## Changes proposed

Merging the commit to fix all broken internal documentation links. 

## Files Modified

**Note:** Link(s) rectified means one or more links were found to point to existing addresses incorrectly and were corrected.

### Link(s) Rectified

The following files had one or more links corrected to point to existing addresses:

- api/README.md
- api/src/api/users/README.md
- api/src/auth/README.md
- api/src/form_schema/forms/README.md
- api/src/services/users/README.md
- api/src/services/xml_generation/README.md
- api/src/workflow/README.md
- documentation/analytics/technical-overview.md
- documentation/api/development.md
- documentation/api/legacy-transforms.md
- documentation/api/local-seed-data.md
- documentation/api/lookup-values.md
- documentation/api/package-dependency-management.md
- documentation/api/technical-overview.md
- documentation/api/workflow-service.md
- documentation/architecture/README.md
- documentation/frontend/development.md
- documentation/frontend/featureFlags.md
- documentation/frontend/internationalization.md
- documentation/infra/environment-variables-and-secrets.md
- documentation/infra/set-up-app-env.md
- documentation/infra/set-up-aws-account.md
- documentation/infra/set-up-infrastructure-tools.md
- documentation/wiki/get-involved/policies-and-guidelines/README.md
- documentation/wiki/product/decisions/infra/README.md
- documentation/wiki/product/release-notes.md
- fluentbit/README.md
- infra/README.md

#### CONTRIBUTING.md
- Link for `.github/internal-contributors.yml` was removed along with its sentence.

#### documentation/README.md
- Link(s) rectified. Link and sentence for `deliverables/individual_deliverables` removed since directory doesn't exist.

#### documentation/analytics/testing.md
[...]

Issue #9551, opened 2026-04-13
[Documentation] Fix Broken Documentation Links


btabaska:
### Summary

The simpler-grants-gov repository has a substantial documentation surface spanning more than 10 subdirectories under `documentation/` plus root-level documents like `DEVELOPMENT.md`, `CONTRIBUTING.md`, and various `README.md` files. As the project has evolved, internal cross-references between these documents have rotted: files have been moved, renamed, or deleted without updating the links that point to them.

Broken links frustrate contributors who click through to referenced content and land on a 404. They also undermine confidence in the documentation as a whole — if links don't work, contributors reasonably wonder whether the content itself is current. This is particularly damaging for onboarding documentation where a new contributor may be trying to follow a chain of references to understand the system.

The project already has the infrastructure to detect broken links: a `lychee`-based CI workflow exists at `.github/workflows/ci-wiki-links.yml`. The immediate work is to run lychee locally, triage all reported broken links, and fix the internal references. Known problem areas include `infra/README.md`, `DEVELOPMENT.md`, and `documentation/frontend/development.md`, but a full lychee run should be treated as the authoritative source of truth rather than fixing only the known cases.

This ticket is scoped to internal (relative) links. External URL breakage is a separate concern and is explicitly out of scope.

## Current State
- Lychee CI workflow exists at `.github/workflows/ci-wiki-links.yml` and runs on PRs, but broken links have accumulated without being fixed.
- Known broken link patterns exist in:
  - `infra/README.md`
  - `DEVELOPMENT.md` (root)
  - `documentation/frontend/development.md`
- The documentation tree spans 10+ subdirectories under `documentation/` plus root-level files.
- No recent audit of broken internal links has been performed.

## Technical Approach
Install lychee locally (`brew install lychee` on macOS or via the GitHub releases page). Run it against the repository root with the `--include-verbatim` flag and scope it to internal links only using the appropriate lychee configuration or flags (e.g., `--exclude 'https?://'` to skip external URLs, or configure `lychee.toml` if one does not already exist).

For each broken link, determine the correct fix: if the target file was moved, update the path; if the target file was deleted and the content no longer exists anywhere, remove the link and the surrounding sentence or note the deletion clearly. Do not create stub files to satisfy links — fix or remove the references.

After making all fixes, run lychee again to confirm zero errors before opening the PR.

### Acceptance criteria

- [ ] `lychee` is run locally against the full documentation tree and the output is used to identify all broken internal links.
[...]

PR #9687, opened 2026-04-20
[Follow up #9551] Added README.md to decisions/deprecated

MozamilS:
## Summary

This is a follow up to pull request #9652 . 

Work for #9551

## Changes proposed

In auditing some of the major changes, it was found that the `.../decisions/deprecated` directory was not tracked due to it being an empty directory. To fix this issue, a generic `README.md` file was made in the directory. 

There was also a minor issue with `...decisions/README.md` where the link for `ADR Github Organization` was an internal link. This was changed to an external link. 

## Context for reviewers
This is an added fix for the #9551 issue. 

## Validation steps

`lychee .` still returns 0 broken internal links.


serverless-mom:
Great PR! There's a small internal debate about relative vs. absolute links, and once that's resolved we'll get you a response on this.

MozamilS:
> Great PR! There's a small internal debate about relative vs. absolute links, and once that's resolved we'll get you a response on this.


In that case, my initial PR to solve #9551 has used relative links for all internal links that were broken. 

With lychee, I found there's 600+ internal links that need to be investigated. I can make all internal links absolute this weekend.

serverless-mom:
Okay final decision @MozamilS is to use root relative links, if it's helpful @btabaska wrote a Python to convert all links in the repo:

```
#!/usr/bin/env python3
"""Convert internal markdown links to root-relative form.

Scans tracked .md files. For each inline link ``[text](url)`` or image
``![alt](url)`` whose URL resolves to a path inside the repository, rewrites
the URL to start with ``/`` (root-relative, resolved from the repo root).
External links, anchor-only links, and already-root-relative links are left
untouched. Reports rewrites and cases that could not be validated.

Does NOT rewrite HTML ``<a href>`` tags, reference-style link definitions
(``[ref]: url``), or content inside fenced code blocks (``` / ~~~) or inline
backticks.

Usage:
    python3 scripts/convert-markdown-links.py --dry-run --report /tmp/r.csv
    python3 scripts/convert-markdown-links.py --report /tmp/r.csv
"""

from __future__ import annotations

import argparse
import csv
import os
import re
import subprocess
import sys
from pathlib import Path
from urllib.parse import unquote


INLINE_LINK_RE = re.compile(r"(!?)\[([^\]\n]*)\]\(([^)\n]*)\)")
FENCE_RE = re.compile(r"^\s{0,3}(`{3,}|~{3,})")
INLINE_CODE_RE = re.compile(r"`+[^`\n]*`+")
SCHEME_RE = re.compile(r"^[a-z][a-z0-9+.\-]*:", re.IGNORECASE)


def is_external_url(url: str) -> bool:
    return bool(SCHEME_RE.match(url)) or url.startswith("//")


def split_url_and_title(target: str) -> tuple[str, str]:
    """Split ``(...)`` contents into ``(url, title_suffix)``.

    ``title_suffix`` begins with the whitespace that separates the URL from an
    optional title and is preserved verbatim so the reassembled link is stable.
    Returns ``("", "")`` for angle-bracketed ``<url>`` forms — those are rare
    and we leave them untouched rather than risk a bad round-trip.
    """
    if target.lstrip().startswith("<"):
        return "", ""
    m = re.search(r"\s", target)
    if m:
        return target[: m.start()], target[m.start() :]
    return target, ""


def split_path_and_anchor(url: str) -> tuple[str, str]:
    """Split ``path#anchor`` or ``path?query`` into ``(path, suffix)``.

    The suffix retains the leading ``#`` or ``?``.
    """
    idx = len(url)
    for ch in "#?":
        i = url.find(ch)
        if i != -1 and i < idx:
            idx = i
    return url[:idx], url[idx:]


def mask_inline_code(line: str) -> str:
    """Replace inline backtick spans with NULs so link regex skips them."""
    return INLINE_CODE_RE.sub(lambda m: "\x00" * len(m.group(0)), line)


def tracked_md_files(repo_root: Path) -> list[Path]:
    out = subprocess.run(
        ["git", "ls-files", "-z", "*.md"],
        cwd=repo_root,
        capture_output=True,
        check=True,
    )
    return [repo_root / p.decode() for p in out.stdout.split(b"\x00") if p]


def rewrite_target(
    repo_root: Path, source: Path, url: str
) -> tuple[str, str, str]:
    """Return ``(new_url, status, reason)``.

    status is one of: ``rewritten`` / ``unchanged`` / ``unable to validate``.
    """
    path, suffix = split_path_and_anchor(url)
    if not path:
        return url, "unchanged", "anchor-only or query-only"

    decoded = unquote(path)
    had_trailing_slash = decoded.endswith("/")
    joined = os.path.normpath(os.path.join(str(source.parent), decoded))
    resolved = Path(joined)

    try:
        rel = resolved.relative_to(repo_root)
    except ValueError:
        return url, "unable to validate", "resolves outside repo root"

    if not resolved.exists():
        return url, "unable to validate", "target does not exist"

    new_path = "/" + str(rel).replace(os.sep, "/")
    if had_trailing_slash and not new_path.endswith("/"):
        new_path += "/"
    if "%20" in path:
        new_path = new_path.replace(" ", "%20")
    new_url = new_path + suffix
    if new_url == url:
        return url, "unchanged", "already equivalent"
    return new_url, "rewritten", ""


def process_file(repo_root: Path, source: Path) -> tuple[str, list[dict]]:
    text = source.read_text(encoding="utf-8")
    lines = text.splitlines(keepends=True)
    rows: list[dict] = []
    in_fence = False
    new_lines: list[str] = []

    for lineno, line in enumerate(lines, start=1):
        if FENCE_RE.match(line):
            in_fence = not in_fence
            new_lines.append(line)
            continue
        if in_fence:
            new_lines.append(line)
            continue

        masked = mask_inline_code(line)
        replacements: list[tuple[int, int, str]] = []
        for match in INLINE_LINK_RE.finditer(masked):
            bang = match.group(1)
            text_part = match.group(2)
            target = match.group(3)

            url, title_suffix = split_url_and_title(target)
            if not url:
                continue
            if url.startswith("#") or is_external_url(url) or url.startswith("/"):
                continue

            new_url, status, reason = rewrite_target(repo_root, source, url)
            if status == "rewritten":
                new_whole = f"{bang}[{text_part}]({new_url}{title_suffix})"
                replacements.append((match.start(), match.end(), new_whole))
                rows.append({
                    "file": str(source.relative_to(repo_root)),
                    "line": lineno,
                    "old": url,
                    "new": new_url,
                    "status": "rewritten",
                    "reason": "",
                })
            elif status == "unable to validate":
                rows.append({
                    "file": str(source.relative_to(repo_root)),
                    "line": lineno,
                    "old": url,
                    "new": "",
                    "status": "unable to validate",
                    "reason": reason,
                })

        if replacements:
            new_line = line
            for start, end, new_whole in reversed(replacements):
                new_line = new_line[:start] + new_whole + new_line[end:]
            new_lines.append(new_line)
        else:
            new_lines.append(line)

    return "".join(new_lines), rows


def main() -> int:
    parser = argparse.ArgumentParser(description=__doc__)
    parser.add_argument("--dry-run", action="store_true", help="do not write files")
    parser.add_argument("--report", type=Path, help="CSV report path")
    parser.add_argument("--verbose", action="store_true")
    args = parser.parse_args()

    repo_root = Path(
        subprocess.run(
            ["git", "rev-parse", "--show-toplevel"],
            capture_output=True,
            check=True,
            text=True,
        ).stdout.strip()
    ).resolve()

    files = tracked_md_files(repo_root)
    all_rows: list[dict] = []
    changed_files: list[str] = []

    for path in files:
        try:
            new_content, rows = process_file(repo_root, path.resolve())
        except Exception as e:  # pragma: no cover
            print(f"ERROR {path}: {e}", file=sys.stderr)
            continue
        rewrites = [r for r in rows if r["status"] == "rewritten"]
        all_rows.extend(rows)
        if rewrites:
            changed_files.append(str(path.relative_to(repo_root)))
            if not args.dry_run:
                path.write_text(new_content, encoding="utf-8")

    if args.report:
        with args.report.open("w", newline="", encoding="utf-8") as f:
            writer = csv.DictWriter(
                f, fieldnames=["file", "line", "old", "new", "status", "reason"]
            )
            writer.writeheader()
            writer.writerows(all_rows)

    n_rewrite = sum(1 for r in all_rows if r["status"] == "rewritten")
    n_invalid = sum(1 for r in all_rows if r["status"] == "unable to validate")
    mode = "dry-run" if args.dry_run else "applied"
    print(f"[{mode}] files scanned:        {len(files)}")
    print(f"[{mode}] files with changes:   {len(changed_files)}")
    print(f"[{mode}] links rewritten:      {n_rewrite}")
    print(f"[{mode}] unable to validate:   {n_invalid}")
    if args.report:
        print(f"[{mode}] report: {args.report}")

    if args.verbose:
        for f in changed_files:
            print(f"  changed: {f}")

    return 0


if __name__ == "__main__":
    sys.exit(main())
```

I want to highlight that you also found a number of links that were incorrectly pointing to within the repository when they should have been out to the named tools, great work! I'll have to check out [lychee](https://github.com/lycheeverse/lychee-action)

btabaska:
Hey, just wanted to jump in and ask if you needed any help with this PR. We'd be happy to help get it to done. Thanks again for all your contributions.

MozamilS:
Hello, 

I'm sorry. I have been occupied the past couple of days and couldn't get back to work. I will try and fix this by the end of the day.

MozamilS:
A brief update for the day. 

Unfortunately, git wouldn't allow me to attach a multi-line commit message so I thought I should share the changes in this comment. 

* The local branch was updated to upstream/main: 
* New broken doc links were resolved. They only required a re-formatting. 
* One doc link was removed from `.github/workflows/README.md` that referenced `markdownlint-config.json`. This file couldn't be found so the corresponding sentence was removed. 
* `lychee.toml` file was updated to consider root-relative links as well. 
* The python script was added to the `/scripts/utilities/` directory. It was applied and all doc links were converted to root relative links. 

There are some issues with the linters that was found after the commit was pushed. I will be working on these issues and push a new commit as soon as I can.

btabaska:
Thank you so much for the update. Feel free to discard the script instead of adding it to utilities. totally fine with it being a "one shot" to get the links fixed and then discarded. The rest sounds great, again, thank you so much for being such a prolific contributor. We appreciate you!

MozamilS:
@btabaska 

Hello, 

Thank you so much for your continued and responsive support. 

I had a look at why the CI workflows are failing. 

* The E2E tests, as was mentioned in #9855, are unrelated. 
* The `CI Project Linters` test seems to fail due to some `*.py` files which this commit hasn't and should not have altered. 
* The`Deliverable mapping` test is related to a `GITHUB_API_TOKEN` environment variable, which this PR also should not have altered since only `*.md` files were modified. 

Regarding the script, it was able to find some broken doc links that lychee couldn't identify. Also, it will be useful in the future since in the week that this PR did not see much movement, 20 new broken doc [incorrect formatting] were introduced. It may be a useful tool for quickly validating lychee's output as well as potentially conducting conversions.

MozamilS:
Hello, 

I resolved some conflicts that appeared between the branch and upstream. There was an issue in `documentation/api/package-dependency-management.md` which I could not find a reasonable resolution for. 

The document had a link referencing an `/api/poetry.lock` file which I could not identify and which was not found with lychee.

MozamilS:
Hello @btabaska 

I hope you are doing well. I wanted to check up on the health of this commit and found a couple of new broken links to fix. There were a number of links in the `monitoring-and-observability/logging-configuration.md` that were referencing some scripts in the `api/src/logging` directory which was empty. I was able to find replacements for them in the `/backend/grants_shared/src/grants_shared/logs` directory. 

Furthermore, I ran the script again and there were 5 new links needing to be rewritten and that was completed as will in this latest commit.

btabaska (commented):
Thanks for the persistence on this one, Sayed — chasing 600+ links to root-relative form across the whole tree is genuinely tedious work, and the `deprecated/README.md` fix plus the lychee config tweak are right on the money. I did a full review and I can't recommend merge just yet: somewhere in the rebases, two docs got reverted to stale content (UV → Poetry, and the `grants_shared.logs` package → the old empty `src.logging`). Those aren't link fixes — they roll back real documentation. Let's get those restored and we're basically there.

## Blocking

**`documentation/api/package-dependency-management.md`**

bug: Heads up — this file got reverted from UV back to Poetry, and that's a regression. On `main` today the API uses UV: there's an `api/uv.lock` (no `poetry.lock`), and `api/pyproject.toml` has `[project]` with `requires-python = ">=3.14,<3.15"`. This PR changes all of that back to `poetry`, `poetry.lock`, `[tool.poetry.dependencies]`, and `python = "~3.12"`. This is the merge conflict you mentioned — looks like the resolution kept the old side of the file. The `/api/poetry.lock` link you couldn't find doesn't exist *because* we're on UV now. Let's restore the UV version of this doc (the `main` version) and apply the root-relative link conversion on top of it. Happy to hop on a quick call if that's easier.

**`documentation/api/monitoring-and-observability/logging-configuration.md`**

bug: Same root cause here. This reverts the package references from `grants_shared.logs` back to `src.logging` — including the code samples (`import src.logging`, `with src.logging.init(...)`). The logging code actually lives at `backend/grants_shared/src/grants_shared/logs/` now; `api/src/logging/` is empty. So this both makes the code snippets wrong and points the first link (`[src.logging](/api/src/logging/)`) at an empty directory, while other links in the same file correctly point to `/backend/grants_shared/...`. The file ends up internally inconsistent. Let's take the `main` version (which already uses `grants_shared.logs`) as the base and just convert its links.

> Since both of these came out of conflict resolution, it'd be worth a quick `git diff main -- documentation/api/` on your branch to confirm nothing else in that directory got rolled back the same way.

## Should fix before merge

**`scripts/utilities/markdown_links_to_root.py`**

suggestion: Per our earlier chat, let's go ahead and drop this from the PR — totally fine as a one-shot to get the links converted, but we'd talked about not committing it to `utilities/`. If you feel strongly it's worth keeping as a maintenance tool, I'm open to it, but then it needs a few fixes first: the docstring usage examples reference `scripts/convert-markdown-links.py` (wrong filename), there are no tests, and the percent-encoding round-trip only re-encodes spaces — any other encoded char (e.g. `%28`, non-ASCII) gets decoded but never re-encoded, which can mangle a link. Easiest path is just to remove it.

**`lychee.toml`**

question: Does lychee actually resolve root-relative links correctly with `root_dir = "."`? My understanding is lychee expects `root_dir` to be an *absolute* path for resolving `/`-prefixed links, and a relative `.` may be ignored. Since the whole PR now depends on root-relative links resolving, can we confirm the CI lychee run passes with this value (and ideally paste the run output)? Want to make sure the config matches how it runs in `ci-wiki-links.yml`.

## Nits

**`documentation/wiki/product/decisions/deprecated/README.md`**

nit: Couple of small typos in the new README: "For current **tural** decisions" should be "architectural decisions," and the naming-convention line `yyyymmdd-xxx.md)` has a stray trailing `)`. Also that last link points to a directory (`/documentation/wiki/product/decisions/adr`) without a trailing slash — worth confirming lychee is happy with directory links given the new config.

**`.github/linters/scripts/check-wiki-pages-linked-to-summary.sh`**

nit: Nice catch normalizing the root-relative prefix here — right adjustment so the linter keeps matching after the conversion. Just double-check the SUMMARY check still passes in CI with both link forms, since this script is the gate for that.

---

**Bottom line:** the link conversion and new README are solid work. The two reverted docs (`package-dependency-management.md`, `logging-configuration.md`) are the blockers — restore the `main` content and re-apply the conversion on top. Drop the script per our earlier conversation, confirm the lychee `root_dir` value actually works in CI, and fix the small typos. Then I think we're good to go. Thanks again for sticking with this, Sayed.

MozamilS:
Hello @btabaska 

Thank you so much for taking the time to go through everything; given the scope of this commit, it definitely takes some time to make sure everything is on point. I will try and fix the issues raised in your comment above. 

Regarding the link conversion tool, I think it's definitely useful to keep it in the code base; It's safe to assume more links will be introduced that would need fixing and having this tool would save a lot of time. I'll work on the test cases, encoding and docstring issues to get it fit enough to keep.

MozamilS:
Hello @btabaska 

A bit of an update on the issues: 

___
The rebase-caused issues with `package-dependency-management.md` and `logging-configuration.md` were resolved. Albeit, things had to get dirty before I was able to clean it up. The commit history is a little messy now but I did my best to provide descriptive commit messages. 

___
The typos in `deprecated/README.md` were resolved as well. Thank you so much for catching those. 

___
I was able to make a list of test cases for the convert-markdown-links.py (it was renamed; I thought it was appropriate to go with the intended initial name). The encoding-decoding steps shouldn't cause issues. The program was modified so the output will always be the same as the output file name. I also added a fallback so the program checks for situations with special characters are literally in the file name (e.g., literal file%20name.md instead of file name.md with a space) It would be great if you could review the test cases in case I have missed any edge scenarios.

___
For `lychee.toml`, please see the screenshot below as a confirmation that the root-relative links do not throw errors and get passed fine. Also, you will notice in the screenshot that there are new broken links introduced in the codebase. 

Convenient list: 

* `.github/workflows/pr-environment-destroy.yml`
* `docs/infra/destroy-infrastructure.md`
* `docs/infra/develop-and-test-infrastructure-in-isolation-using-workspace.md`
* `docs/infra/pull-request-environments.md`
*  `.github/workflows/template-only-ci-infra.yml`

Given that these new links were added very recently, I thought I should be cautious in deleting them right away. Please do let me know what should be done with these. Thank you again. 




Screenshot 2026-06-11 at 11 34 16 AM

PR #9689, opened 2026-04-20
[Issue #9723] Re-use existing Docker images for API during E2E tests

mdragon:
## Summary

Work for #9723 

## Changes proposed

1. During CI/CD make the API start/build smarter about using existing API container images if they already are cached in GH Actions Cache.
2. Do the NPM/Playwright installs and build the API outside the matrix job and then pass them in
3. We can't move the FE build outside the matrix because it needs data from the BE image (the Shared Key and Testing Token)

## Context for reviewers

1. As we lean into E2E we're trying to ensure stuff is running efficiently and is using already built images for stuff a branch or merge doesn't change. 
2. Things are complicated a bit by the fact that these playwright and npm caches in particular GH seems to like to fake cache miss on, so we back the caches (which will survive run to run) with the artifacts (that will survive only job to job). Unfortunately that slows stuff down, but otherwise either npm or playwright would fail to load from cache 100% of test runs.

## Validation steps

- Checks pass
- Builds build

### DIFF
--- a/api/Makefile
+++ b/api/Makefile
@@ -106,9 +106,11 @@ start-debug:
 run-logs: start ## Start the API and follow the logs
 	docker compose logs --follow --no-color $(APP_NAME)
 
-init: setup-env-override-file build init-db init-opensearch init-s3mock init-sqsmock init-mock-soap-services ## Run all of the init tasks (including build) to set up DB, Search, s3/sqs mocks, and Mock services. To do this "from scratch" use remake-backend instead
+init-components: init-db init-opensearch init-s3mock init-sqsmock init-mock-soap-services
 
-init-nobuild: setup-env-override-file init-db init-opensearch init-s3mock init-sqsmock init-mock-soap-services ## Do not build, but do run all of the init tasks to set up DB, Search, s3/sqs mocks, and Mock services. To do this "from scratch" use remake-backend instead
+init: setup-env-override-file build init-components ## Run all of the init tasks (including build) to set up DB, Search, s3/sqs mocks, and Mock services. To do this "from scratch" use remake-backend instead
+
+init-nobuild: setup-env-override-file init-components ## Do not build, but do run all of the init tasks to set up DB, Search, s3/sqs mocks, and Mock services. To do this "from scratch" use remake-backend instead
 
 clean-volumes: ## Remove project docker volumes - which includes the DB, and OpenSearch state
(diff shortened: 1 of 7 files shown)


mdragon:
@btabaska This is the other PR I'm working in `api/Makefile`

api/Makefile
```diff
@@ -103,7 +103,9 @@ start-debug:
 run-logs: start ## Start the API and follow the logs
 	docker compose logs --follow --no-color $(APP_NAME)
 
-init: setup-env-override-file build init-db init-opensearch init-s3mock init-sqsmock init-mock-soap-services ## Run all of the init tasks to set up DB, Search, s3/sqs mocks, and Mock services. To do this "from scratch" use remake-backend instead
+init: setup-env-override-file build init-db init-opensearch init-s3mock init-sqsmock init-mock-soap-services ## Run all of the init tasks (including build) to set up DB, Search, s3/sqs mocks, and Mock services. To do this "from scratch" use remake-backend instead
+
+init-nobuild: setup-env-override-file init-db init-opensearch init-s3mock init-sqsmock init-mock-soap-services ## Do not build, but do run all of the init tasks to set up DB, Search, s3/sqs mocks, and Mock services. To do this "from scratch" use remake-backend instead
```
chouinar: I'd probably split this into a few building blocks. If we add any other `init-X` - easier to add to one place than multiple.

Something like:

```
init-components: init-db init-opensearch init-s3mock init-sqsmock init-mock-soap-services

init-nobuild: setup-env-override-file init-components

init: setup-env-override-file build init-components
```

api/README.md
```diff
@@ -91,3 +91,4 @@ Poetry CLI commands are of the form `<task group> <task name> <any other params>
 * [Writing Tests](../documentation/api/writing-tests.md)
 * [Logging configuration](../documentation/api/monitoring-and-observability/logging-configuration.md)
 * [Logging conventions](../documentation/api/monitoring-and-observability/logging-conventions.md)
+a
```
chouinar:
```suggestion

```

doug-s-nava: a

mdragon:
This should be ready for actual review now that it's finally working.

prasnava (approved):
👍 Tested the pipeline: [here](https://github.com/HHS/simpler-grants-gov/actions/runs/25169636186/job/73785139845)

Issue #9692, opened 2026-04-20
SOAP/Proxy: Add correct header and footer to the ConfirmApplicationDelivery error response


jakobpederson:
### Summary

The current response is missing the:
```
--uuid:<uuid>
Content-Type: application/xop+xml; charset=UTF-8; type="text/xml"
Content-Transfer-Encoding: binary
Content-ID: 
<root.message@cxf.apache.org>
```
header and the
```
--uuid:<uuid>-
```
footer

<img width="1624" height="970" alt="Image" src="https://github.com/user-attachments/assets/c1b8655d-f20b-451b-bc0d-ae10849ce082" />

<img width="1624" height="970" alt="Image" src="https://github.com/user-attachments/assets/d58d4e64-5774-4775-860c-c0c50e940723" />

### Acceptance criteria

- [ ] correct response comes back from error

PR #9694, opened 2026-04-20
[Issue #9692] Add mtom wrapper to soap fault exception response

jakobpederson:
## Summary

Fixes / Work for #9692  

## Changes proposed

Add MTOM wrapper for `SOAPFaultException` error response and updated tests.

## Context for reviewers

The response was missing the MTOM header and footer, so it was not completely mimicking the legacy behavior. This was added to make sure the responses from Simpler and Legacy math.

## Validation steps

- [ ] hit the `ConfirmApplicationDelivery` endpoint with an invalid `GrantsGovTrackingNumber`
- [ ] see responses do not match

### DIFF
--- a/api/src/legacy_soap_api/legacy_soap_api_utils.py
+++ b/api/src/legacy_soap_api/legacy_soap_api_utils.py
@@ -155,12 +155,23 @@ def get_soap_fault_error_response(
     mtom_response = (
         f"--uuid:{boundary_id}\r\n"
         f'Content-Type: application/xop+xml; charset=UTF-8; type="text/xml"\r\n'
-        f"Content-Transfer-Encoding: 8bit\r\n"
+        f"Content-Transfer-Encoding: binary\r\n"
         f"Content-ID: <root.message@cxf.apache.org>\r\n\r\n"
         f"{err}\r\n"
         f"--uuid:{boundary_id}--\r\n"
     ).encode()
-    return get_soap_response(data=mtom_response, status_code=500, headers=headers)
+    response_headers = {
+        "Content-Type": (
+            "multipart/related;"
+            ' type="application/xop+xml";'
+            f' boundary="uuid:{boundary_id}";'
+            ' start="<root.message@cxf.apache.org>";'
+            ' start-info="text/xml"'
+        )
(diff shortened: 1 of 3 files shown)


api/src/legacy_soap_api/legacy_soap_api_utils.py
```diff
@@ -136,6 +136,33 @@ def get_soap_error_response(
...
+                <faultstring>{faultstring}</faultstring>
+            </soap:Fault>
+        </soap:Body>
+    </soap:Envelope>
+    """.strip()
+    boundary_id = str(uuid.uuid4())
+    mtom_response = (
+        f"--uuid:{boundary_id}\r\n"
+        f'Content-Type: application/xop+xml; charset=UTF-8; type="text/xml"\r\n'
+        f"Content-Transfer-Encoding: 8bit\r\n"
```
kkrug: Up on line 53 it uses `binary` but here we use `8bit`. Just wanted to confirm this is what we want.

api/src/legacy_soap_api/legacy_soap_api_utils.py
```diff
@@ -136,6 +136,33 @@ def get_soap_error_response(
...
+        f"Content-ID: <root.message@cxf.apache.org>\r\n\r\n"
+        f"{err}\r\n"
+        f"--uuid:{boundary_id}--\r\n"
+    ).encode()
```
kkrug, line 162: Do we need a response header similar to what we see on line 79? The response body includes MTOM boundary markers but the Content-Type response header never declares the boundary. Something like:
```
    ).encode()
    response_headers = {
        "Content-Type": (
            "multipart/related;"
            ' type="application/xop+xml";'
            f' boundary="uuid:{boundary_id}";'
            ' start="<root.message@cxf.apache.org>";'
            ' start-info="text/xml"'
        ),
    }
    if headers:
        response_headers.update(headers)
```

jakobpederson: Great catch, we do need to include that.

kkrug (approved):

Issue #8928, opened 2026-03-09
Backend - Create the endpoint to update an award recommendation


nathan-stricker:
### Summary

See [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Create-the-endpoint-to-update-an-award-recommendation)

#### Endpoint
`PUT /alpha/award-recommendations/:award_recommendation_id`

#### Purpose
Update an award recommendation

#### AuthZ
Requires `update_award_recommendation` privilege

#### Request
```json
{
  "additional_info": "",
  "award_selection_method": "", # Enum
  "funding_strategy": "",
  "other_key_information": "",
  "additional_info": "",
}
```

#### Response
The same as GET award-recommendation

#### Logic
- Fetch the award recommendation, 404 if not found
- AuthZ, 403 on error
- This will override values for all fields, if a field isn’t provided, it should be treated as null (any non-nullable fields should be required and non-null in the schema)

### Acceptance criteria

- [ ] Endpoint Created
- [ ] Tests Created

PR #9695, opened 2026-04-20
[Issue#8928]update award recommendation

chay-REIsys:
## Summary

Work for #8928  

## Changes proposed

Update Award Recommendation Endpoint
Resolving mapping issues between funding strategy and selection method detail
## Context for reviewers

We should be able to update an existing award recommendation

## Validation steps

- Pull latest
- Search for Update Award Recommendation Endpoint and click Try it out (to test this either create a new recommendation (create award recommendation endpoint) or use an existing from award recommendation table.


api/src/api/award_recommendations_alpha/award_recommendation_schemas.py
```diff
@@ -33,6 +33,36 @@ class AwardRecommendationCreateRequestSchema(Schema):
...
+        metadata={"description": "Additional info about the award recommendation"},
+    )
+    funding_strategy = fields.String(
+        allow_none=True,
+        metadata={"description": "Funding strategy information for the award recommendation"},
+    )
+    selection_method_detail = fields.String(
+        allow_none=True,
+        metadata={"description": "Additional details about the selection method"},
+    )
```
nathan-stricker, line 89: Would be good to add `example` metadata to all these String fields

nathan-stricker (commented):
LGTM aside from one comment on adding examples!

kkrug (commented):
Can we add logging to the update service similar to what is in [this service](https://github.com/HHS/simpler-grants-gov/blob/1f418f20a0c9852ef6e3c4beaee6bec092292ceb/api/src/services/award_recommendations/update_award_recommendation_review.py#L62)? It should probably be in at least the create service as well (since it is also mutating), but I'll leave that up to you if the scope is creeping up to be too large.

kkrug (dismissed):

api/src/services/award_recommendations/update_award_recommendation.py
```diff
@@ -0,0 +1,76 @@
...
+            "award_selection_method": (
+                award_recommendation.award_selection_method.value
+                if award_recommendation.award_selection_method
+                else None
+            ),
+            "additional_info": award_recommendation.additional_info,
+            "selection_method_detail": award_recommendation.selection_method_detail,
+            "funding_strategy": award_recommendation.funding_strategy,
+            "other_key_information": award_recommendation.other_key_information,
+            "user_id": user.user_id,
```
kkrug: Can we just confirm that nothing in here could contain PII? Same with the create service.

chaitanya-duddella: yes i don't think there is any PII info but i have removed user id and pushed changes

kkrug (approved):

Issue #9685, opened 2026-04-20
SOAP/Proxy: Update check to see if GrantsGovTrackingNumber is legacy or simpler


jakobpederson:
### Summary

Currently we only do this check for `GetApplicationZip` and we'd like to expand that to include `UpdateApplicationInfo` and 'ConfirmApplicationDelivery`.

### Acceptance criteria

- [ ] Skip sending extra traffic to legacy when we don't need to

PR #9696, opened 2026-04-20
[Issue #9685] Update methods that skip calling legacy soap

jakobpederson:
## Summary

Fixes / Work for #9685  

## Changes proposed

Added more alternate SOAP operations and tests. Added methods to generate "not found" responses.

## Context for reviewers

We don't want to send a request to the legacy system for two additional endpoints, ConfirmApplicationDelivery and UpdateApplicationInfo, so we added this to the list to check if we skip it, and just construct our own proxy response message.

## Validation steps

- [ ] hit the endpoint with a simpler tracking number
- [ ] check logs to see if it gets sent to legacy

### DIFF
--- a/api/src/legacy_soap_api/legacy_soap_api_utils.py
+++ b/api/src/legacy_soap_api/legacy_soap_api_utils.py
@@ -34,6 +34,8 @@ HIDDEN_VALUE = "hidden"
 class AlternateSoapOperation(StrEnum):
     GET_APPLICATION_ZIP = "GetApplicationZipRequest"
     GET_APPLICATION = "GetApplicationRequest"
+    CONFIRM_APPLICATION_DELIVERY = "ConfirmApplicationDeliveryRequest"
+    UPDATE_APPLICATION_INFO = "UpdateApplicationInfoRequest"
 
 
 def format_local_soap_response(response_data: bytes, boundary_id: str | None = None) -> bytes:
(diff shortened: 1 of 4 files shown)


api/src/legacy_soap_api/legacy_soap_api_utils.py
```diff
@@ -88,6 +90,82 @@ def get_soap_proxy_grant_application_not_found_response(
     return get_soap_response(response_data, 500, headers=response_headers)
 
 
+def get_soap_proxy_failed_to_confirm_delivery_response(
+    grants_gov_tracking_number: str, headers: dict, is_get_application_zip: bool = True
```
kkrug: Do we need to pass in `is_get_application_zip: bool = True` here? Same for `get_soap_proxy_failed_to_update_application_info()`.

Also `grants_gov_tracking_number` looks unused here.

api/src/legacy_soap_api/legacy_soap_api_utils.py
```diff
@@ -336,10 +414,16 @@ def get_alternate_proxy_response(soap_request: SOAPRequest) -> SOAPResponse | No
     if soap_request.operation_name in AlternateSoapOperation:
         tracking_number = get_gov_grants_tracking_number(xml_bytes)
         is_zip = soap_request.operation_name == AlternateSoapOperation.GET_APPLICATION_ZIP
+        DEFAULT_RESPONSES = {
+            AlternateSoapOperation.GET_APPLICATION_ZIP: get_soap_proxy_grant_application_not_found_response,
+            AlternateSoapOperation.GET_APPLICATION: get_soap_proxy_grant_application_not_found_response,
+            AlternateSoapOperation.CONFIRM_APPLICATION_DELIVERY: get_soap_proxy_failed_to_confirm_delivery_response,
+            AlternateSoapOperation.UPDATE_APPLICATION_INFO: get_soap_proxy_failed_to_update_application_info,
+        }
```
kkrug: This could be a module level constant so we don't rebuild it every time.

api/tests/src/legacy_soap_api/test_legacy_soap_api_routes.py
```diff
@@ -519,6 +522,128 @@ def test_getapplicationzip_operation_returns_not_found_response_if_simpler_id_is
     )
 
 
+@mock.patch("uuid.uuid4")
+@mock.patch("src.legacy_soap_api.legacy_soap_api_proxy._get_soap_response")
+def test_getapplicationzip_operation_returns_not_found_response_if_simpler_id_is_used_for_confirm_application_delivery(
```
kkrug: Should this be named something like:
```suggestion
def test_confirm_application_delivery_returns_not_found_response_if_simpler_id_is_used(
```

api/tests/src/legacy_soap_api/test_legacy_soap_api_routes.py
```diff
@@ -519,6 +522,128 @@ def test_getapplicationzip_operation_returns_not_found_response_if_simpler_id_is
...
+    assert expected == response.data
+    assert (
+        response.headers["Content-Type"]
+        == f'multipart/related; type="application/xop+xml"; boundary="uuid:{test_uuid}"; start="<root.message@cxf.apache.org>"; start-info="text/xml"'
+    )
+
+
+@mock.patch("uuid.uuid4")
+@mock.patch("src.legacy_soap_api.legacy_soap_api_proxy._get_soap_response")
+def test_getapplicationzip_operation_returns_not_found_response_if_simpler_id_is_used_for_update_application_info(
```
kkrug: Should this be named something like:
```suggestion
def test_update_application_info_returns_not_found_response_if_simpler_id_is_used(
```

api/src/legacy_soap_api/legacy_soap_api_utils.py
```diff
@@ -161,11 +137,42 @@ def get_soap_proxy_failed_to_update_application_info(
             ' start="<root.message@cxf.apache.org>";'
             ' start-info="text/xml"'
         ),
-        "Set-Cookie": (f"{headers.get('Cookie')}; Path=/grantsws-agency; Secure; HttpOnly"),
```
kkrug: Why did `Set-Cookie` get removed here?

jakobpederson: I think it was a mistake, adding it back

kkrug (approved):

PR #9621, opened 2026-04-15
[Issue #9003] Email notification content

babebe:
## Summary

Fixes for #9003  

## Changes proposed
Implemented new email layout based on Figma
Added dynamic messaging:
1 opportunity → singular phrasing
2–3 opportunities → plural phrasing
3 opportunities → show first 3 + “+ X more”

Made opportunity titles clickable (first 3 only)

Added links to:
Notifications page (/notifications)
Saved opportunities page (filtered by organization)

## Validation steps

Covered in test

PR #9703, opened 2026-04-21
[Issue #8976] Add org saved opportunity email notification task

### LINKED ISSUE #8976: [BE] Create new email notification task for organization saved opportunity notifications
### Summary

We want to create a new email notification task for when new opportunities have been saved to organizations that a user is a member of. We will need to implement a new notification class derived from the [base notification class](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/task/notifications/base_notification.py) and add that new organization saved opportunity notification task to be run under the `EmailNotificationTask` class [here](https://github.com/HHS/simpler-grants-gov/blob/2977d3387db9cb07bbd5750bed4d7bdbae7c6eac/api/src/task/notifications/email_notification.py#L43)

This email notification should only be sent to users who have organization saved opportunity notifications configured in their **Notification Preferences** page. The cadence of this email will be the same as other email notifications which is `Every day at 11:40am Eastern Time during DST. 12:40pm during non-DST.` as stated [here](https://github.com/HHS/simpler-grants-gov/blob/0ea5b65caefa23a52fcdc03927a42908a56a7620/infra/api/app-config/env-config/scheduled_jobs.tf#L188).

The email should utilize the templates built as part of #9003.  Note that this email takes in a list of opportunities and truncates the number down for display.

Here is the algorithm.  Note that it is best effort, send-at-most once.  If we get halfway through the list of users and the task errors, there will be no recovery and the users won't receive notice of those opportunities in the next run.

* For every organization in the organization saved opportunity table that does not have an notification_processed_at timestamp
   * Get a list of the members of that organization
   * Filter the list of members to include only those with notification enabled for organization opportunities notifications
   * Select all opportunities saved to that organization but haven’t been notified yet, i.e. where organization = organization_id and  notification_processed_at is null
      * Mark notification_processed_at on all rows as the timestamp now (e.g. of when the processing this organization begins)
      * For each member (who has already been filtered by their notification preferences) send the email with list of organizations with list of opportunities

#### Notes
- See [this](https://github.com/HHS/simpler-grants-gov/issues/9011#issuecomment-4121044937) comment for running locally

### Acceptance criteria

- [ ] New env var for enabling this is created and set to default false. This should be set to true in staging and dev only.
```python
enable_org_saved_opportunity_notifications: bool = Field(default=False, alias="ENABLE_ORG_SAVED_OPPORTUNITY_NOTIFICATIONS")
```
- [ ] New email notification task has been created with algorithm above, pay careful attention to
   - [ ] multiple opportunities saved for a single org
   - [ ] notification settings
- [ ] Unit tests pass
- [ ] UAT to ensure correct and expected data is included in the email
   - [ ] UAT happy path through manual invocation:
      - [ ] Setup notifications for a user that is member of organization through API or UI
      - [ ] Manually run task once and ensure email is sent
      - [ ] Run task again ensure no email is sent
      - [ ] Add a new organization saved opportunity
      - [ ] Run task again and ensure email is sent with new opportunity
   - [ ] UAT scheduled task:
      - [ ] Carefully setup the correct data in lower environments and expect + verify the email is sent at appropriate time
      - [ ] Verify the email design and content is correct
   - [ ] UAT re-notify:
      - [ ] Add new saved opportunity to org and manually run task and ensure email is sent
      - [ ] Remove the saved opportunity from the org and re-save it and manually run the task and ensure the opportunity is included in the email notification
   - [ ] UAT notification link functionality:
      - [ ] Ensure link brings up the notification preferences pages as expected
   - [ ] UAT user with multiple organizations:
      - [ ] Run through the happy path through manual invocation UAT but the user will have notifications enabled for both organizations
      - [ ] Ensure new opportunities from both organization are included in the email notification
   - [ ] UAT user subscribed but no new opportunities saved:
      - [ ] Ensure no email is sent
   - [ ] UAT one user in an org who has notifications enabled receives an email and another user in the same organization who does not have notifications enabled does not receive the email.

mikehgrantsgov:
## Summary

Fixes #8976

## Changes proposed

Implements `OrgSavedOpportunityNotificationTask`, a task that emails organization members when new opportunities are saved to their org
Integrates the new task into the existing EmailNotificationTask runner behind a `ENABLE_ORG_SAVED_OPPORTUNITY_NOTIFICATIONS` feature flag (default false)
Enables the flag in dev and staging environments via Terraform config

## Context for reviewers

We want to create a new email notification task for when new opportunities have been saved to organizations that a user is a member of. We will need to implement a new notification class derived from the [base notification class](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/task/notifications/base_notification.py) and add that new organization saved opportunity notification task to be run under the EmailNotificationTask class [here](https://github.com/HHS/simpler-grants-gov/blob/2977d3387db9cb07bbd5750bed4d7bdbae7c6eac/api/src/task/notifications/email_notification.py#L43)

The task:
Queries all `organization_saved_opportunity` rows where notification_processed_at IS NULL
Groups by org, finds eligible members (org notifications enabled, email not suppressed)
Marks rows as processed (notification_processed_at = now) before sending
Aggregates across orgs so users in multiple orgs receive a single email
Truncates to 5 opportunities per org with a "and N more..." link to /saved-opportunities

## Validation steps

See unit test coverage. 

For local confirmation:
Make sure db is up to date.

Run:
```
make console

user = f.UserFactory.create()
f.LinkExternalUserFactory.create(user=user, email="your-real-email@example.com")
org = f.OrganizationFactory.create(sam_gov_entity=None)
f.OrganizationUserFactory.create(organization=org, user=user)
f.UserSavedOpportunityNotificationFactory.create(user=user, organization=org, email_enabled=True)
opp = f.OpportunityFactory.create()
f.OrganizationSavedOpportunityFactory.create(organization=org, opportunity=opp)
dbs.commit()
```

Ensure `RESET_EMAILS_WITHOUT_SENDING=false` is set in `api/override.env`, then run the task:
```
docker compose run --rm -e ENABLE_ORG_SAVED_OPPORTUNITY_NOTIFICATIONS=true grants-api poetry run flask task email-notifications
```
Confirm`notification_processed_at` is set on the row (email won't be received sending locally).

Expected log output:
```
Processing org saved opportunity notifications  organization_id=... opportunity_count=1 eligible_member_count=1
Successfully delivered notification to user     pinpoint_delivery_status=SUCCESSFUL pinpoint_status_code=200
Completed OrgSavedOpportunityNotificationTask   org_saved_opportunities_tracked=1 users_notified=1
```

Re-run the task and confirm no second email is sent (users_notified=0, notifications_reset=0).


api/src/task/notifications/org_saved_opportunity_notification.py
```diff
@@ -0,0 +1,263 @@
...
+    "Closed on federal holidays"
+)
+
+
+class OrgSavedOpportunityNotificationTask(BaseNotificationTask):
+
+    def __init__(
+        self,
+        db_session: db.Session,
+        notification_config: EmailNotificationConfig | None = None,
```
freddieyebra: Can this can be a required parameter?

api/src/task/notifications/org_saved_opportunity_notification.py
```diff
@@ -0,0 +1,263 @@
...
+        if not user_org_opportunities:
+            logger.info("No eligible members found for org saved opportunity notifications")
+            return []
+
+        # Build one email per user covering all their organizations' new opportunities
+        notifications: list[UserEmailNotification] = []
+        for user_id, org_opp_list in user_org_opportunities.items():
+            email = user_emails[user_id]
+            all_opportunity_ids = [opp.opportunity_id for _, opps in org_opp_list for opp in opps]
+            content = self._build_notification_message(org_opp_list)
```
freddieyebra: Not a blocker and not mentioned in the algorithm on the ticket, but a minor performance enhancement could be building the email content per organization once all orgs and opportunities have been aggregated. that way the email content is build per organization instead of per user per organization.

api/src/task/notifications/org_saved_opportunity_notification.py
```diff
@@ -0,0 +1,263 @@
...
+            org_to_rows.setdefault(row.organization_id, []).append(row)
+
+        # user_id -> list of (organization, list of opportunities)
+        user_org_opportunities: dict[UUID, list[tuple[Organization, list[Opportunity]]]] = {}
+        user_emails: dict[UUID, str] = {}
+
+        for org_id, org_rows in org_to_rows.items():
+            organization = org_rows[0].organization
+            opportunities = [row.opportunity for row in org_rows]
+
```
freddieyebra, line 169: Can we calculate `displayed` and `remaining` here so that it is calculated per organization instead of per user per organization?

```suggestion
            organization = org_rows[0].organization
            opportunities = [row.opportunity for row in org_rows]
            displayed = opportunities[:MAX_OPPORTUNITIES_DISPLAYED]
            remaining = len(opportunities)-len(displayed)
```

May want to consider stuffing these values into an object as well.

api/src/task/notifications/org_saved_opportunity_notification.py
```diff
@@ -0,0 +1,263 @@
...
+                opp_title = html.escape(opportunity.opportunity_title or "")
+                message += f"• <a href='{opp_url}' target='_blank'>{opp_title}</a>\n"
+
+            if remaining > 0:
+                saved_opps_url = (
+                    f"{self.notification_config.frontend_base_url}/saved-opportunities{UTM_TAG}"
+                )
+                message += (
+                    f"• <a href='{saved_opps_url}' target='_blank'>"
+                    f"And {remaining} more opportunity{'s' if remaining > 1 else ''}...</a>\n"
```
freddieyebra:
```suggestion
                    f"And {remaining} more opportunit{'ies' if remaining > 1 else 'y'}...</a>\n"
```

api/src/task/notifications/org_saved_opportunity_notification.py
```diff
@@ -0,0 +1,263 @@
...
+                )
+
+            message += "\n"
+
+        notification_prefs_url = (
```
freddieyebra: Can we move this as global or a part of a config somewhere. or a cached method would be great too.

api/src/task/notifications/org_saved_opportunity_notification.py
```diff
@@ -0,0 +1,263 @@
...
+        message += (
+            f"<a href='{notification_prefs_url}' target='_blank'>"
+            "Manage your notification preferences</a>\n\n"
+            "<b>Questions?</b>\n"
+            "If you have questions about an opportunity, please contact the grantor using "
+            "the contact information on the listing page.\n\n"
+        )
+        message += CONTACT_INFO
+
+        return message.replace("\n", "<br/>")
```
freddieyebra: Can we just replace `"\n"` with `"<br/>"` in the string building initially instead of needing to replace it?

api/src/task/notifications/org_saved_opportunity_notification.py
joshtonava: issue: This doesn't quite give the correct plural. `opportunities` vs `opportunity`

api/src/task/notifications/org_saved_opportunity_notification.py
```diff
@@ -0,0 +1,263 @@
...
+        stmt = (
+            select(OrganizationUser)
+            .options(selectinload(OrganizationUser.user))
+            .where(OrganizationUser.organization_id == org_id)
+            .where(
+                exists().where(
+                    and_(
+                        UserSavedOpportunityNotification.user_id == OrganizationUser.user_id,
+                        UserSavedOpportunityNotification.organization_id == org_id,
+                        UserSavedOpportunityNotification.email_enabled.is_(True),
```
joshtonava, line 264: discussion:
This is optional but if we want to get metrics on for example `NOTIFICATIONS_SKIPPED_EMAIL_DISABLED`
we could remove this filter from the query and do the filtering in the code so we can increment the metric.

api/src/task/notifications/org_saved_opportunity_notification.py
```diff
@@ -0,0 +1,263 @@
...
+                    and_(
+                        SuppressedEmail.email == LinkExternalUser.email,
+                        LinkExternalUser.user_id == OrganizationUser.user_id,
+                    )
+                )
+            )
+        )
+        return list(self.db_session.execute(stmt).scalars().all())
+
+    def _build_notification_message(
```
freddieyebra: We should be utilizing the email template that was built as part of [this](https://github.com/HHS/simpler-grants-gov/pull/9621) ticket since that went through design review.

api/src/task/notifications/org_saved_opportunity_notification.py
freddieyebra: This may already be addressed from https://github.com/HHS/simpler-grants-gov/pull/9621

freddieyebra (changes requested):

api/src/task/notifications/org_saved_opportunity_notification.py
```diff
@@ -0,0 +1,263 @@
...
+                    OrganizationSavedOpportunity.organization_id == org_id,
+                    OrganizationSavedOpportunity.notification_processed_at.is_(None),
+                )
+                .values(notification_processed_at=now)
+            )
+
+            self.increment(Metrics.ORG_SAVED_OPPORTUNITIES_TRACKED, len(opportunities))
+
+            # Find members of this org who have notifications enabled and are not suppressed
+            eligible_members = self._get_eligible_members(org_id)
```
joshtonava, line 183: **Discussion:** This runs SQL query per org. We may need to monitor this for performance. 

We could do something like opportunity_notification.py. One query fetching all `OrganizationUser` rows for the set of org_ids, joined/filtered by notification preference and suppressed email checks and then finally grouped by organization_id in python.

api/src/task/notifications/org_saved_opportunity_notification.py
```diff
@@ -0,0 +1,263 @@
...
+                )
+                message += (
+                    f"• <a href='{saved_opps_url}' target='_blank'>"
+                    f"And {remaining} more opportunity{'s' if remaining > 1 else ''}...</a>\n"
+                )
+
+            message += "\n"
+
+        notification_prefs_url = (
+            f"{self.notification_config.frontend_base_url}/notification-preferences{UTM_TAG}"
```
joshtonava: issue: I dont think this is the correct path.

`frontend/src/app/[locale]/(base)/notifications/page.tsx`

I think its just `notifications`

api/tests/src/task/notifications/test_org_saved_opportunity_notification.py
```diff
@@ -0,0 +1,398 @@
+import pytest
```
joshtonava, line 1: issue: can you add a test for truncation logic? I dont see anything that tests the more than 5 opportunities for an org.

api/tests/src/task/notifications/test_org_saved_opportunity_notification.py
```diff
@@ -0,0 +1,398 @@
...
+                    UserNotificationLog.notification_reason
+                    == NotificationReason.ORG_SAVED_OPPORTUNITY
+                )
+            )
+            .scalars()
+            .all()
+        )
+        assert len(notification_logs) == 1
+
+    def test_user_in_multiple_orgs_receives_single_email(
```
joshtonava: issue: Can we also test the HTML in at least one test?

api/src/task/notifications/org_saved_opportunity_notification.py
```diff
@@ -0,0 +1,263 @@
...
+          - Aggregate per-user across all their orgs
+        - Send one email per user with all their orgs' new opportunities
+        """
+        now = datetime_util.utcnow()
+
+        # Fetch all unprocessed org-opportunity rows with relationships loaded
+        stmt = (
+            select(OrganizationSavedOpportunity)
+            .options(
+                selectinload(OrganizationSavedOpportunity.organization).selectinload(
```
joshtonava, line 142: nit: Do you think you could write a comment for why we are doing this chaining?

api/src/task/notifications/org_saved_opportunity_notification.py
```diff
@@ -0,0 +1,297 @@
...
+                    and_(
+                        SuppressedEmail.email == LinkExternalUser.email,
+                        LinkExternalUser.user_id == OrganizationUser.user_id,
+                    )
+                )
+            )
+        )
+        return list(self.db_session.execute(stmt).scalars().all())
+
+    def _build_notification_message(self, org_opp_list: list[OrgOpportunityGroup]) -> str:
```
babebe: This functionality was handled:  in the `build_notification_content` func . Which could then be modified if needed. Feel free to move that function into this file, since you created a new one. The same with the tests that were created could be moved into the test files here.

mikehgrantsgov: Thanks, I moved build_notification_content into `org_saved_opportunity_notification.py` and updated it to use `MAX_OPPORTUNITIES_DISPLAYED` with the pluralization fix. `_build_notification_message` now delegates to it for the single-org case. 

The three tests from test_saved_opportunity_notification.py have been moved into test_org_saved_opportunity_notification.py (updating the ">3 items" test to ">5 items" to match the new threshold), and test_saved_opportunity_notification.py was deleted.

api/src/task/notifications/org_saved_opportunity_notification.py
mikehgrantsgov: Added

api/tests/src/task/notifications/test_org_saved_opportunity_notification.py
mikehgrantsgov: Added `test_html_content_includes_opportunity_details` which asserts the opportunity title, the /opportunity/{id} link, and the /notifications preferences URL all appear in the rendered email HTML.

api/tests/src/task/notifications/test_org_saved_opportunity_notification.py
mikehgrantsgov: Added `test_truncation_more_than_max_opportunities` which creates 7 saved opportunities for an org, runs the task, and asserts "+ 2 more opportunities" appears in the email HTML.

api/src/task/notifications/org_saved_opportunity_notification.py
mikehgrantsgov: Thanks. Fixed this.. the URL is now /notifications in both `build_notification_content` and the multi-org path in _build_notification_message

api/src/task/notifications/org_saved_opportunity_notification.py
mikehgrantsgov: Agreed, this is a real concern at scale. I'd suggest tracking it as a follow-up rather than blocking this PR? Happy to implement the bulk query pattern now if you'd prefer to address it here though.

api/src/task/notifications/org_saved_opportunity_notification.py
mikehgrantsgov: Thanks, did this. `build_notification_content` (from PR #9621) has been moved into this and `_build_notification_message` now delegates to it

api/src/task/notifications/org_saved_opportunity_notification.py
mikehgrantsgov: Updated

api/src/task/notifications/org_saved_opportunity_notification.py
mikehgrantsgov: the `notification_prefs_url` construction was removed entirely from `_build_notification_message` when the method was refactored to delegate to `build_notification_content`

api/src/task/notifications/org_saved_opportunity_notification.py
mikehgrantsgov: @freddieyebra yep!

api/src/task/notifications/org_saved_opportunity_notification.py
babebe: I think it makes sense to consolidate notification message construction into a single function. We can update build_notification_content to accept a list of OrgOpportunityGroups and handle both single-organization and multi-organization cases by branching on the length.

This removes duplicated body-building HTML and formatting logic currently split between build_notification_content and _build_notification_message, and easier to make future updates if/when it occurs

api/src/task/notifications/org_saved_opportunity_notification.py
mikehgrantsgov: Thanks, this is done on both counts, both displayed and remaining are now computed in the per-org loop, and the values are wrapped in the OrgOpportunityGroup dataclass along with organization and opportunities.

api/src/task/notifications/org_saved_opportunity_notification.py
mikehgrantsgov: Good point. The per-org HTML section could theoretically be pre-built and reused, but I think `build_notification_content` has to run per-user because the subject line and intro copy branch on how many orgs that specific user belongs to (single-org vs. multi-org paths produce different text)? For now I'd lean toward leaving it as-is unless you feel strongly?

api/src/task/notifications/org_saved_opportunity_notification.py
mikehgrantsgov: Made this change

freddieyebra (commented):
There is one test scenario that needs to be addressed/updated. The following test cases have been verified:

1. CLI testing
Notification expected to be sent
Image
Image

No notifications expected to be sent
Image

2. Rendering of email body content cases
Single opportunity
Image

Multiple opportunities
Image

More than 5 opportunities
Image

3. Testing cases through UI
- 2 users have email notifications enabled for the same organization and received emails as expected based on logs and `notification_processed_at` column
- Saved opportunity re-added after notification is included in email as well.
- New user enables notifications after task ran and does not get notifications.

The one scenario that we need to fix/align on is the scenario where a user has notifications enabled for multiple organizations. The following outlines the current functionality for this case:
1. User enables notifications for 2 organizations Fred's Fabric Hut and Sally's Soup Emporium
2. Task runs
3. User receives 1 email with the following content 
Image

Expected functionality:
3. 2 **separate** emails per organization is sent to the same user. 

I realize this may not have been as clear in the ticket as was described in the figma. we can discuss and align on this.

api/src/task/notifications/base_notification.py
```diff
@@ -128,5 +128,7 @@ def run_task(self) -> None:
         """Override to define the task logic"""
         with self.db_session.begin():
             notifications = self.collect_email_notifications()
-            self.send_notifications(notifications)
+        # Commit marks before sending so a crash mid-send doesn't retry and duplicate emails
```
freddieyebra: can we override this `run_task` in the OrgSavedOpportunityNotificationTask class to ensure existing notifications are not impacted.

freddieyebra (approved):
looks good!

- verified email per organization
- single new saved opportunity 
Image

- multiple new saved opportunity
Image

- more than 5 new saved opportunities

Image

- multiple org users with notifications enabled receive emails 
- view all saved opportunities links work (logged in & logged out)

Issue #7769, opened 2026-01-05
Bypass login in e2e tests against staging


doug-s-nava:
### Summary

Depends on https://github.com/HHS/simpler-grants-gov/issues/7768

See https://github.com/HHS/simpler-grants-gov/issues/7256 for background

In order to allow for circumventing the full login.gov authorization process in staging for purposes of efficient end to end testing, we want to be able to:

* allow for providing an `e2e test user` API key to the playwright process via env var, matching one set up in the corresponding simpler DB for an environment
* whenever a test user token (`E2E_USER_AUTH_TOKEN`) is not found on an environment variable, fetch a viable auth token for a test user using the passed in test user API key
* set this token in memory to be used by any playwright tests that need it
* use as much as possible of the existing spoofing system found here https://github.com/HHS/simpler-grants-gov/blob/6a2e4bba811bf7b10949406361c11ccb478f126e/frontend/tests/e2e/loginUtils.ts#L55

To get this working against CI, we will need to:
* ensure that a test user exists in the staging DB with the correct privileges and an API key
* store the test user's API key in Github secrets
* update any staging facing E2E Github Actions with an env var for the e2e test user API key (sourced from Github secrets)

### Acceptance criteria

- [ ] tests that rely on spoofing login, such as the saved opportunity tests, pass when run against staging locally
- [ ] tests that rely on spoofing login, such as the saved opportunity tests, pass when run against staging from CI
- [ ] Standalone test that still tests the Login.gov login flow and successful access to SGG (like that the Account page loads or something) - MD

doug-s-nava:
moving back to 5.5 due to work on other workstreams

PR #9720, opened 2026-04-21
[Issue #7769] bypass staging login

doug-s-nava:
## Summary

Fixes for #7769 

## Changes proposed

Allows playwright to spoof a login to staging, resulting in a logged in state for a test user without going through the full login UX.

## Context for reviewers

Using this system in staging requires setting some environment variables:

* `STAGING_TEST_USER_API_KEY` - a special API key used to authenticate a special test user created in the staging database
* `SESSION_SECRET` - the secret used to encode server side JWTs as client side JWTs. Previously this was only needed for local test processes - the staging spoof process will need the staging secret

If you need these values for testing, obtain them from 1pass or ask a team member

## Validation steps

1. ensure that STAGING_TEST_USER_API_KEY and SESSION_SECRET are set in .env.local with the correct values for testing against staging, and that PLAYWRIGHT_TARGET_ENV is set to "staging"
2. run `npx playwright test --config ./tests/playwright.config.ts`
3. _VERIFY_: tests pass (this was flaky on my end, likely fine as long as the CI versions pass)
4. _VERIFY_: staging test run passes https://github.com/HHS/simpler-grants-gov/actions/runs/25322401661/job/74234457188
5. _VERIFY_: local test run passes https://github.com/HHS/simpler-grants-gov/actions/runs/25322436069/job/74234961317

### DIFF
--- a/.github/actions/e2e/action.yml
+++ b/.github/actions/e2e/action.yml
@@ -47,6 +47,13 @@ inputs:
   force-webkit-install:
     description: "webkit dependencies aren't all cached, so we need to force it to install"
     default: "false"
+  staging_test_user_api_key:
+    description: "test user API key used in spoofing staging user logins"
+    default: ""
+  staging_client_session_secret:
+    description: "encode key for client side JWTs, used in spoofing staging user logins"
+    default: ""
+
 outputs:
   artifact-id:
     description: "artifact ids for uploaded test reports"
--- a/.github/actions/e2e/action.yml
+++ b/.github/actions/e2e/action.yml
@@ -111,7 +111,7 @@ runs:
         STAGING_TEST_USER_PASSWORD: ${{ inputs.staging_test_user_password }}
         STAGING_TEST_USER_MFA_KEY: ${{ inputs.staging_test_user_mfa_key }}
         STAGING_TEST_USER_API_KEY: ${{ inputs.staging_test_user_api_key }}
-        SESSION_SECRET: ${{ inputs.staging_client_session_secret }}
+        SESSION_SECRET_OVERRIDE: ${{ inputs.staging_client_session_secret }}
 
       run: |
         npm run test:e2e
@@ -130,7 +130,7 @@ runs:
         STAGING_TEST_USER_PASSWORD: ${{ inputs.staging_test_user_password }}
         STAGING_TEST_USER_MFA_KEY: ${{ inputs.staging_test_user_mfa_key }}
         STAGING_TEST_USER_API_KEY: ${{ inputs.staging_test_user_api_key }}
-        SESSION_SECRET: ${{ inputs.staging_client_session_secret }}
+        SESSION_SECRET_OVERRIDE: ${{ inputs.staging_client_session_secret }}
 
       run: |
         npm run test:e2e -- --grep "${{ inputs.playwright_tags }}"
--- a/.github/actions/e2e/action.yml
+++ b/.github/actions/e2e/action.yml
@@ -110,6 +110,9 @@ runs:
         STAGING_TEST_USER_EMAIL: ${{ inputs.staging_test_user_email }}
         STAGING_TEST_USER_PASSWORD: ${{ inputs.staging_test_user_password }}
         STAGING_TEST_USER_MFA_KEY: ${{ inputs.staging_test_user_mfa_key }}
+        STAGING_TEST_USER_API_KEY: ${{ inputs.staging_test_user_api_key }}
+        SESSION_SECRET: ${{ inputs.staging_client_session_secret }}
+
       run: |
         npm run test:e2e
       shell: bash
@@ -126,6 +129,9 @@ runs:
         STAGING_TEST_USER_EMAIL: ${{ inputs.staging_test_user_email }}
         STAGING_TEST_USER_PASSWORD: ${{ inputs.staging_test_user_password }}
         STAGING_TEST_USER_MFA_KEY: ${{ inputs.staging_test_user_mfa_key }}
+        STAGING_TEST_USER_API_KEY: ${{ inputs.staging_test_user_api_key }}
+        SESSION_SECRET: ${{ inputs.staging_client_session_secret }}
+
       run: |
         npm run test:e2e -- --grep "${{ inputs.playwright_tags }}"
       shell: bash
--- a/documentation/frontend/testing.md
+++ b/documentation/frontend/testing.md
@@ -1,24 +1,28 @@
 # Testing
 
-## E2E testing
+## End to End (E2E) testing
+
+E2E tests are run using Playwright. See [development.md](/DEVELOPMENT.md) for more general info!
 
 ### Spoofing logins
 
-There are situations where we want to be able to test a "logged in" experience without having to script the test through the full login flow. In order to support this we have built a system to spoof the user login by placing a session cookie into the browser context. THis system works by creating a client side cookie on the browser context within Playwright that will effectively log in a fake user with the API.
+There are situations where we want to be able to test a "logged in" experience without having to script the test through the full login flow. In order to support this we have built a system to spoof the user login by placing a session cookie into the browser context. This system works by creating a client side cookie on the browser context within Playwright that will function the same as the session cookie. produced as the output of the real login process.
 
-#### With a local server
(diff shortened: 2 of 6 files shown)


frontend/tests/e2e/playwright-env.ts
```diff
@@ -14,6 +14,13 @@ const BASE_URLS: Record<string, string> = {
   staging: process.env.STAGING_BASE_URL || "https://staging.simpler.grants.gov",
 };
 
+// API URLs for each environment, read from .env.local if present, else fallback to defaults
+const API_URLS: Record<string, string> = {
+  local: process.env.LOCAL_BASE_URL || "http://127.0.0.1:8080",
+  staging:
+    process.env.STAGING_BASE_URL || "https://api.staging.simpler.grants.gov",
```
Bhavna-Ramachandran: Should we change this : STAGING_BASE_URL (frontend url) to api url?

frontend/tests/e2e/utils/authenticate-e2e-user-utils.ts
```diff
@@ -37,6 +70,17 @@ export async function authenticateE2eUser(
...
   } else if (targetEnv === "staging") {
+    if (attemptSpoof) {
+      try {
+        const spoofSuccessful = await attemptStagingSpoof(context);
+        if (spoofSuccessful) {
+          await page.goto(baseUrl, { waitUntil: "domcontentloaded" });
+          return;
+        }
+      } catch (e) {
+        console.warn("unable to spoof staging login", (e as Error).message);
```
Bhavna-Ramachandran: I noticed we’re logging the error both inside attemptStagingSpoof and again here - is that intentional?

frontend/tests/e2e/utils/authenticate-e2e-user-utils.ts
```diff
@@ -37,6 +70,17 @@ export async function authenticateE2eUser(
     await selectLocalTestUser(page, testUserLabel);
     await page.waitForTimeout(2000);
   } else if (targetEnv === "staging") {
+    if (attemptSpoof) {
+      try {
+        const spoofSuccessful = await attemptStagingSpoof(context);
+        if (spoofSuccessful) {
```
Bhavna-Ramachandran, line 71: Is the intent to fall back to the MFA login flow if spoofing fails? It looks like if spoofing fails, it will silently continue to the next block, check if the user is already logged in, and then proceed with MFA if not.

frontend/tests/e2e/playwright-env.ts
```diff
@@ -23,8 +30,16 @@ if (!Object.prototype.hasOwnProperty.call(BASE_URLS, targetEnv)) {
   );
 }
 
+if (!Object.prototype.hasOwnProperty.call(API_URLS, targetEnv)) {
+  throw new Error(
+    `Unsupported PLAYWRIGHT_TARGET_ENV: ${targetEnv}. Allowed values: ${Object.keys(BASE_URLS).join(", ")}`,
```
Bhavna-Ramachandran: Should this be API_URLS ?

documentation/frontend/testing.md
```diff
@@ -1,20 +1,28 @@
...
-- copy the token variable declaration from the e2e_token.tmp file into your frontend .env.local file
-- that's it! Running e2e tests using the functionality mentioned above should now work locally
+- copy the token variable declaration from the e2e_token.tmp file into the `E2E_USER_AUTH_TOKEN` env var declaration in your frontend .env.local file
+- that's it! Running e2e tests using spoofing should now work. [This process is encapsulated in our CI process here](https://github.com/HHS/simpler-grants-gov/blob/c5f29978c45d658329f2466d652da743d83d6a73/.github/workflows/ci-frontend-e2e.yml#L97).
+
+#### Staging setup
+
+Since staging tests run on a deployed server that does not expose a testing token directly, to spoof a login is a bit more involved, but only requires proper env vars to be set in order to work. A test user is set up in staging that can be spoofed. In order to obtain a session token for this user, Playwright needs to request it from a staging-only internal endpoint.
+
+To run spoofed logins (locally or in CI), you will too have the SESSION_SECRET and STAGING_TEST_USER_API_KEY env vars are correctly set in .env.local. Values can be found 1password, or ask a team member.
```
Bhavna-Ramachandran: Just a minor typo here

documentation/frontend/testing.md
```diff
@@ -1,20 +1,28 @@
...
+## End to End (E2E) testing
 
-### Spoofing logins
+E2E tests are run using Playwright. See [development.md](/DEVELOPMENT.md) for more general info!
 
-There are situations where we want to be able to test a "logged in" experience without having to script the test through the full login flow. In order to support this we have built a system to spoof the user login by placing a session cookie into the browser context.
+### Spoofing logins
 
-This system is based on a `createSpoofedSessionCookie` which will create a client side cookie on the browser context that will effectively log in a fake user with the API.
+There are situations where we want to be able to test a "logged in" experience without having to script the test through the full login flow. In order to support this we have built a system to spoof the user login by placing a session cookie into the browser context. This system works by creating a client side cookie on the browser context within Playwright that will function the same as the session cookie. produced as the output of the real login process.
```
Bhavna-Ramachandran: Typo here : session cookie produced by the output....

frontend/tests/e2e/playwright-env.ts
doug-s-nava: refactored this slightly, let me know what you think

frontend/tests/e2e/utils/authenticate-e2e-user-utils.ts
doug-s-nava: yes, that was my intention, but let me know if you think we should error out instead. The downside of falling back to manual login is that tests would run longer and we wouldn't necessarily know that the spoofing is failing unless we checked the logs... Should we console.error instead of console.warn in that case?

frontend/tests/e2e/utils/authenticate-e2e-user-utils.ts
doug-s-nava: I think that's just left over from my debugging. Will clean up

frontend/tests/e2e/playwright-env.ts
doug-s-nava: good catch

frontend/tests/e2e/utils/authenticate-e2e-user-utils.ts
Bhavna-Ramachandran: Yes, falling back to MFA makes sense to avoid blocking CI on non-UI issues. However, we should still have visibility when spoofing is not used, so we can detect when it’s failing or being bypassed unintentionally. Now it would happen silently, so it might be helpful to log when the fall back happens.

For example, we can add a warning like: "Spoofing unsuccessful. Switching to MFA login flow." or something similar.

What are your thoughts?

.github/actions/e2e/action.yml
```diff
@@ -93,6 +99,9 @@ runs:
         STAGING_TEST_USER_EMAIL: ${{ inputs.staging_test_user_email }}
         STAGING_TEST_USER_PASSWORD: ${{ inputs.staging_test_user_password }}
         STAGING_TEST_USER_MFA_KEY: ${{ inputs.staging_test_user_mfa_key }}
+        STAGING_TEST_USER_API_KEY: ${{ inputs.staging_test_user_api_key }}
+        SESSION_SECRET: ${{ inputs.staging_client_session_secret }}
```
Bhavna-Ramachandran: For local tests, staging_client_session_secret isnt passed and this is currently setting "" in local and I believe this is shadowing the values from .env.local. Hence local tests are now throwing 'Error: Unable to spoof login, missing auth key'

Ref: https://github.com/HHS/simpler-grants-gov/actions/runs/25120771676/job/73620556990?pr=9720

.github/actions/e2e/action.yml
```diff
@@ -109,6 +118,9 @@ runs:
...
```
Bhavna-Ramachandran: Should we change this to : SESSION_SECRET_OVERRIDE: ${{ inputs.staging_client_session_secret }}

Bhavna-Ramachandran (approved):

Issue #9667, opened 2026-04-17
Change behavior for buttons


brianlee-nava:
### Summary

At 4/17 standup, the team highlighted some changes action buttons in order to meet Quad 5 ACs

### Acceptance criteria

- [ ] Change button label from "Submit for view" to "Publish"
- [ ] Enable "Publish" button after all required fields are filled out. There should be 4 required fields
- [ ] Disable "Preview" button
- [ ] When user clicks "Publish", user is redirected to Opportunity List Page and Opportunity is shown as published

PR #9725, opened 2026-04-21
[Issue #9667] Change button behavior on Edit Opportunity page

desaiamit:
## Summary

Work for #9667

## Changes proposed

- Renamed "Submit for review" button to "Publish"
- Disabled "Preview" button (feature not yet built)
- Enabled "Publish" button only when all 4 required fields are filled: Publish Date, Funding Type, Funding Category, and at least one Eligible Applicant
- Wired "Publish" button to backend `POST /v1/grantors/opportunities/{id}/publish` endpoint — on success, redirects user to `/opportunities`

## Context for reviewers

The Publish button lives in `OpportunityEditHeaderActions` (a `"use client"` component) while the form state lives in `OpportunityEditForm` (also a client component). Since they are siblings in the tree, a custom DOM event on the shared form element is used to communicate field value changes from the form to the header actions component.

The `publishOpportunityAction` server action calls the grantor publish endpoint, which queues the opportunity for the publish workflow via SQS. The workflow service then asynchronously sets `is_draft = false`. The frontend redirects to `/opportunities` immediately after the API call succeeds (the queue is accepted), not after the workflow completes.

## Validation steps

- [ ] Load the edit page for a draft opportunity with all 4 required fields empty — Publish button is disabled
- [ ] Fill in Publish Date, Funding Type, Funding Category, and select an Eligible Applicant — Publish button enables
- [ ] Clear Publish Date — Publish button disables. Re-fill — Publish button re-enables
- [ ] Clear any one of the other 3 fields — Publish button disables
- [ ] Click Save — form saves correctly, no regression
- [ ] Click Publish (logged in as a user with `publish_opportunity` privilege) — page redirects to `/opportunities`
- [ ] Preview button is always disabled


frontend/src/components/opportunity/OpportunityEditHeaderActions.tsx
```diff
@@ -0,0 +1,100 @@
...
+      } = (
+        e as CustomEvent<{
+          publishDate: string;
+          fundingType: string;
+          fundingCategories: string;
+          eligibleApplicants: string[];
+        }>
+      ).detail;
+      setPublishEnabled(
+        publishDate.trim() !== "" &&
```
glenblosser-nava, line 67: If you enter an invalid value for this field, you can still click "publish" despite being unable to save. We need to either do validation before enabling the button, or have the appropriate error display when trying to publish an invalid value.
To test: I entered "Apple" as a date. The field allowed me to do so. While I couldn't save (it displayed the validation error), it allowed me to click Publish and navigated me away from the page

desaiamit: Updated the behavior - the Publish button now stays enabled when a non-empty invalid date is entered (so the user isn't left wondering why Publish is disabled), but on click it triggers the form's Save validation first. If the date is invalid, the validation error displays in the form and the publish call is not made. Only when the date passes validation does it proceed to publish.

glenblosser-nava (approved):

frontend/src/components/opportunity/OpportunityEditHeaderActions.tsx
```diff
@@ -0,0 +1,116 @@
+"use client";
```
doug-s-nava, line 1: since `actions` mean something specific in the context of NextJS, and this is a component, I'd suggest renaming this something like `OpportunityHeaderButtons`

frontend/src/components/opportunity/OpportunityEditHeaderActions.tsx
```diff
@@ -0,0 +1,116 @@
...
+import { OpportunityEditFormValues } from "./opportunityEditFormConfig";
+
+type OpportunityEditHeaderActionsProps = {
+  opportunityId: string;
+  initialValues: OpportunityEditFormValues;
+  previewLabel: string;
+  publishLabel: string;
+};
+
+function isValidDate(str: string): boolean {
```
doug-s-nava, line 18: can use https://day.js.org/docs/en/parse/is-valid instead

frontend/src/components/opportunity/OpportunityEditHeaderActions.tsx
```diff
@@ -0,0 +1,116 @@
...
+        onClick={() => {
+          if (!isValidDate(currentPublishDate)) {
+            const form = document.getElementById(
+              "opportunity-edit-form",
+            ) as HTMLFormElement | null;
+            form?.requestSubmit();
+            return;
+          }
+          setIsPublishing(true);
+          publishOpportunityAction(opportunityId)
```
doug-s-nava, line 101: use `useActionState` with actions, otherwise they won't function as server actions, just regular functions

frontend/src/services/fetch/fetchers/opportunitySummaryGrantorFetcher.ts
```diff
@@ -63,3 +64,15 @@ export async function createOpportunitySummaryForGrantor({
...
   return (await response.json()) as OpportunitySummaryDetailApiResponse;
 }
+
+export async function publishOpportunityForGrantor(
+  opportunityId: string,
+  token: string,
+): Promise<GrantorOpportunityApiResponse> {
+  const response = await publishGrantorOpportunityRequest({
+    subPath: `${opportunityId}/publish`,
+    additionalHeaders: { "X-SGG-Token": token },
```
doug-s-nava, line 74: remove and handle in endpoint config

frontend/src/components/opportunity/OpportunityEditForm.tsx
```diff
@@ -116,6 +116,27 @@ export default function OpportunityEditForm({
   const validationErrors: OpportunityEditValidationErrors | undefined =
     formState.validationErrors;
 
+  useEffect(() => {
+    const form = document.getElementById("opportunity-edit-form");
+    if (!form) return;
+    form.dispatchEvent(
```
doug-s-nava, line 122: Since what you're doing here is setting state for the header buttons based on the state of the form, this should be done by passing a shared function from the page down to the form and the buttons component so that they can share state that is controlled at a higher level.

Since the page component is a server component and can't use "useState" you'd need to create an intermediate component to wrap the form and the buttons that could hold the state, but that is still preferable to having to deal with event handlers and custom events. You could also consider using context here, but that might be overkill.

For the publish date, you can also trigger a refetch of the opportunity on submit to grab the latest data. This is how things are done on the apply side. I need to refresh my memory on how that's currently being done

doug-s-nava: I believe that in a case where a server action is mutating data, it will trigger a page refresh. You can thus safely expect the page to refetch any data after a completed server action, and so it should pull the new timestamps automatically

PR #9730, opened 2026-04-21
[Issue #7302] Wire up attachment upload widget to opportunity edit form

### LINKED ISSUE #7302: Ensure that opportunity page supports multiple uploads

desaiamit:
## Summary

Work for #7302

## Changes proposed

- **`OpportunityAttachmentUploadWidget`** — Fully implemented: i18n via `useTranslations`, 2GB client-side file size validation, error `Alert` on upload/delete failure, `isDraft` prop (disables input and hides Remove buttons for non-draft opportunities), `deletable` flag (initially-loaded attachments are display-only since the grantor GET response does not return `opportunity_attachment_id`), `crypto.randomUUID()` for stable unique keys on display-only files
- **`OpportunityEditForm`** — Added Attachments section wired to the widget; section layout matches Figma (`display-flex flex-column gap-3`, `padding-bottom-4`); accepts `initialAttachments` prop
- **`opportunityResponseTypes`** — `GrantorOpportunityDetail` now uses `Omit<OpportunityDetail, "attachments">` to avoid type conflict between `OpportunityDocument[]` (public endpoint) and `OpportunityAttachment[]` (grantor endpoint)
- **`page.tsx`** — Passes `initialAttachments` to `OpportunityEditForm`; passes `attachments: []` to `buildOpportunityEditInitialValues` (which does not use attachments) to satisfy the `OpportunityDetail` type
- **`i18n/messages/en/index.ts`** — Added `content.attachmentsIntro` and `attachments.*` error/action strings

## Context for reviewers

**Why `deletable: false` for initially-loaded attachments?** The grantor GET endpoint (`/grantors/opportunities/{id}`) returns attachments via `OpportunityAttachmentV1Schema` which does not serialize `opportunity_attachment_id`. Without an ID we cannot call the delete endpoint, so these files are shown as read-only. Only freshly uploaded files (which get an ID back from the upload response) can be deleted.

**Why `Omit<OpportunityDetail, "attachments">`?** The public opportunity detail uses `OpportunityDocument[]` (with `download_path`, `file_description`) while the grantor endpoint uses `OpportunityAttachment[]` (with `opportunity_attachment_id`, `mime_type`, `file_size`). These are incompatible types so we omit and re-declare rather than extend directly.

**Why `crypto.randomUUID()` for display-only files?** Initially-loaded files have no `opportunity_attachment_id`, so we generate a stable UUID at init time to ensure unique React list keys even when multiple files share the same filename.

## Validation steps

1. Load the edit page for a draft opportunity with existing attachments — files appear in the list (no Remove button)
2. Upload a new file — it appears in the list with a Remove button
3. Upload the same filename twice — no duplicate key console errors
4. Upload a file >2GB — error alert appears, file not added to list
5. Click Remove → confirm in modal → file removed from list
6. Click Remove → cancel → file remains
7. Load the edit page for a non-draft opportunity — file input is disabled, no Remove buttons


frontend/src/services/fetch/fetchers/opportunityAttachmentFetcher.ts
```diff
@@ -0,0 +1,52 @@
...
+const fetchGrantorOpportunityWithMethod = (type: "POST" | "DELETE" | "GET") =>
+  requesterForEndpoint(toDynamicGrantorOpportunityEndpoint(type));
+
+export const listOpportunityAttachments = async (
+  opportunityId: string,
+  token: string,
+): Promise<OpportunityAttachmentListResponse> => {
+  const response = await fetchGrantorOpportunityWithMethod("GET")({
+    subPath: `${opportunityId}/attachments`,
+    additionalHeaders: { "X-SGG-Token": token },
```
doug-s-nava: remove and handle in endpoint config

frontend/src/services/fetch/fetchers/opportunityAttachmentFetcher.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export const uploadOpportunityAttachment = async (
+  opportunityId: string,
+  token: string,
+  file: FormData,
+): Promise<OpportunityAttachmentUploadResponse> => {
+  const response = await fetchGrantorOpportunityWithMethod("POST")({
+    subPath: `${opportunityId}/attachments`,
+    additionalHeaders: {
+      "X-SGG-Token": token,
```
doug-s-nava: remove and handle in endpoint config

frontend/src/services/fetch/fetchers/opportunityAttachmentFetcher.ts
```diff
@@ -0,0 +1,52 @@
...
+};
+
+export const deleteOpportunityAttachment = async (
+  opportunityId: string,
+  attachmentId: string,
+  token: string,
+): Promise<{ status_code: number; message: string }> => {
+  const response = await fetchGrantorOpportunityWithMethod("DELETE")({
+    subPath: `${opportunityId}/attachments/${attachmentId}`,
+    additionalHeaders: { "X-SGG-Token": token },
```
doug-s-nava: remove and handle in endpoint config

frontend/src/components/opportunity/OpportunityAttachmentUploadInput.tsx
```diff
@@ -0,0 +1,197 @@
...
+  FileInputRef,
+  FormGroup,
+  ModalRef,
+  ModalToggleButton,
+} from "@trussworks/react-uswds";
+
+import { DeleteAttachmentModal } from "src/components/application/attachments/DeleteAttachmentModal";
+import { USWDSIcon } from "src/components/USWDSIcon";
+
+const MAX_FILE_SIZE_BYTES = 2 * 1024 * 1024 * 1024; // 2GB
```
doug-s-nava, line 20: test to make sure, but I think this is enforced at the next config level, shouldn't need to do anything about it here https://github.com/HHS/simpler-grants-gov/blob/d774e6038b7e5e35ffb83cf6daf28e6aa61b4491/frontend/next.config.js#L191

frontend/src/components/opportunity/OpportunityAttachmentUploadWidget.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+export function OpportunityAttachmentUploadWidget({
+  opportunityId,
+  initialAttachments = [],
+  isDraft = false,
+}: OpportunityAttachmentUploadWidgetProps) {
+  const t = useTranslations("OpportunityEdit.attachments");
+
+  const [uploadedFiles, setUploadedFiles] = useState<UploadedFile[]>(
+    initialAttachments.map((a) => ({
```
doug-s-nava:
```suggestion
    initialAttachments.map((attachment) => ({
```

frontend/src/components/opportunity/OpportunityAttachmentUploadWidget.tsx
```diff
@@ -0,0 +1,197 @@
...
+      if (file.size > MAX_FILE_SIZE_BYTES) {
+        setErrorMessage(t("errorFileTooLarge", { fileName: file.name }));
+        continue;
+      }
+
+      const formData = new FormData();
+      formData.append("file", file);
+
+      try {
+        const res = await fetch(
```
doug-s-nava: use `useClientFetch`

frontend/src/components/opportunity/OpportunityAttachmentUploadWidget.tsx
```diff
@@ -0,0 +1,197 @@
...
+          };
+          setUploadedFiles((prev) => [
+            ...prev,
+            {
+              id: data.opportunity_attachment_id,
+              name: file.name,
+              deletable: true,
+            },
+          ]);
+        } else {
```
doug-s-nava: log error info here (since there is no error, this would probably just be status), and in the catch block

frontend/src/components/opportunity/OpportunityAttachmentUploadWidget.tsx
```diff
@@ -0,0 +1,197 @@
...
+    fileInputRef.current?.clearFiles();
+  };
+
+  const confirmDelete = async (): Promise<void> => {
+    if (!fileToDelete) return;
+    setErrorMessage(null);
+    setDeletePending(true);
+
+    try {
+      const res = await fetch(
```
doug-s-nava: use `useClientFetch`

frontend/src/components/opportunity/OpportunityAttachmentUploadWidget.tsx
```diff
@@ -0,0 +1,197 @@
...
+        `/api/opportunities/${opportunityId}/attachments/${fileToDelete.id}`,
+        { method: "DELETE" },
+      );
+      if (res.ok) {
+        setUploadedFiles((prev) =>
+          prev.filter((f) => f.id !== fileToDelete.id),
+        );
+        setFileToDelete(null);
+        deleteModalRef.current?.toggleModal();
```
doug-s-nava: do we want to keep the modal open if there's an error? seems like this should maybe go in the finally block

frontend/src/components/opportunity/OpportunityAttachmentUploadInput.tsx
```diff
@@ -0,0 +1,197 @@
+"use client";
```
doug-s-nava, line 1: let's reserve `Widget` for apply specific form field components and call this `OpportunityAttachmentUploadInput`? or maybe there's something better

frontend/src/components/opportunity/OpportunityAttachmentUploadInput.tsx
```diff
@@ -0,0 +1,197 @@
...
+      <FileInput
+        id="opportunity-attachment-upload"
+        name="opportunity-attachment-upload"
+        ref={fileInputRef}
+        type="file"
+        multiple
+        disabled={isUploading || !isDraft}
+        onChange={(e) => {
+          handleFileChange(e.currentTarget.files).catch(() =>
+            setErrorMessage(t("errorUploadFailed", { fileName: "" })),
```
doug-s-nava, line 147: handleFileChange is swallowing pretty much all errors, other than anything that might happen when appending the file to form data. Do you want it to rethrow? or maybe make a note that this would only capture errors related to appending to form data?

doug-s-nava (commented):
in terms of functionality the only issue I see is that after save, the option to delete the attachment disappears

frontend/src/components/opportunity/OpportunityAttachmentUploadInput.tsx
desaiamit: Renamed to OpportunityAttachmentUploadInput - keeping 'Widget' apply-specific

frontend/src/components/opportunity/OpportunityAttachmentUploadWidget.tsx
desaiamit: Fixed - switched to useClientFetch for both the upload and delete calls.

frontend/src/components/opportunity/OpportunityAttachmentUploadWidget.tsx
desaiamit: Fixed - switched to useClientFetch for both the upload and delete calls.

frontend/src/components/opportunity/OpportunityAttachmentUploadWidget.tsx
desaiamit: Fixed - logging the status code on non-ok responses and the error in the catch block.

frontend/src/components/opportunity/OpportunityAttachmentUploadInput.tsx
desaiamit: Added a comment clarifying that the .catch() here only covers errors thrown synchronously before the loop (e.g. appending to FormData) - per-file upload errors are already caught individually inside the loop.

frontend/src/components/opportunity/OpportunityAttachmentUploadWidget.tsx
desaiamit: Fixed

frontend/src/components/opportunity/OpportunityAttachmentUploadInput.tsx
desaiamit: Verified - next.config.js already enforces sizeLimit: "2000mb" and bodySizeLimit: "2000mb". The client-side check is for immediate user feedback before the upload starts.

frontend/src/services/fetch/fetchers/opportunityAttachmentFetcher.ts
desaiamit: Fixed

frontend/src/services/fetch/fetchers/opportunityAttachmentFetcher.ts
desaiamit: Fixed

frontend/src/services/fetch/fetchers/opportunityAttachmentFetcher.ts
desaiamit: Fixed

desaiamit:
Fixed the issue with "option to delete the attachment disappears" - The opportunity_attachment_id was missing from OpportunityAttachmentV1Schema, so the GET response never included it. After save the page re-fetched and all attachment IDs came back as empty, causing the delete button to hide. Added the field to the schema so the ID is now serialized and the delete button persists after save.

api/src/api/opportunities_v1/opportunity_schemas.py
```diff
@@ -308,6 +308,10 @@ class OpportunityV1Schema(Schema):
 
 
 class OpportunityAttachmentV1Schema(FileResponseSchema):
+    opportunity_attachment_id = fields.UUID(
```
doug-s-nava, line 311: note that this will effect grantee facing opportunity endpoints as well, probably not a big deal

frontend/src/components/opportunity/OpportunityAttachmentUploadInput.tsx
doug-s-nava: did you test this? does Next deny the upload without this bespoke error handling?

frontend/src/components/opportunity/OpportunityAttachmentUploadInput.tsx
```diff
@@ -0,0 +1,194 @@
...
+export function OpportunityAttachmentUploadInput({
+  opportunityId,
+  initialAttachments = [],
+  isDraft = false,
+}: OpportunityAttachmentUploadInputProps) {
+  const t = useTranslations("OpportunityEdit.attachments");
+
+  const { clientFetch: uploadFetch } = useClientFetch<{
+    opportunity_attachment_id: string;
+  }>("upload attachment");
```
doug-s-nava, line 43: the string passed in here is the error message that will be displayed on a non 200 status - can you make this more helpful?

doug-s-nava (approved):
nothing blocking!

✅ ✅ ✅ ✅

api/src/api/opportunities_v1/opportunity_schemas.py
desaiamit: ok, also I had tested the MAX_FILE_SIZE_BYTES

frontend/src/components/opportunity/OpportunityAttachmentUploadInput.tsx
desaiamit: Will be Fixed - Will update it to "Error uploading opportunity attachment" and "Error deleting opportunity attachment"

doug-s-nava:
On the file size upload limit, this is what I see when I upload a 3gb file without the client side file size check.
* request errors out and I receive a standard error message (see image below)
* the NextJS log suggests that the 2gb limit set at the application level is working

```
Request body exceeded 1.95GB for /api/opportunities/c08211c1-f9b5-4899-b799-7e021d031a8c/attachments. Only the first 1.95GB will be available unless configured. See https://nextjs.org/docs/app/api-reference/config/next-config-js/middlewareClientMaxBodySize for more details.
⨯ TypeError: Failed to parse body as FormData.
    at ignore-listed frames {
  [cause]: TypeError: expected boundary after body
      at ignore-listed frames
}
```
Screenshot 2026-04-23 at 9 48 37 AM


So I don't think we need the client side check

PR #9732, opened 2026-04-21
[Issue #9724] Drop opensearch-dashboard from CI/CD runs of docker

### LINKED ISSUE #9724: Remove open search dashboards from API build used in local e2e test runs
### Summary

remove open search dashboards from API build used in local e2e test runs

### Acceptance criteria

- [ ] remove open search dashboards from API build used in local e2e test runs

mdragon:
Replaced by #9736 

## Summary

Fixes #9724 

## Changes proposed

- Remove opensearch-dashboard from the make entry CI/CD uses to do `docker compose up` 
- Keeps local behavior the same so make start still includes the search dashboard for any search development that needs it.

## Validation steps

- Checks/tests still work in CI/CD
- Tested that `docker compose up` still works if someone doesn't know about our make commands


prasnava (approved):
👍

api/docker-compose-both.yml
```diff
@@ -0,0 +1,161 @@
+services:
```
chouinar, line 1: We actually don't need to do any docker shenanigans for this. I had written up a whole thing about structuring these files, but here's a far easier fix.

`make start` runs `docker compose up --detach`. Change that to `docker compose up --detach $(APP_NAME)`.

That's it. Basically right now we run `docker compose up` which starts everything in this file. Instead we tell it just start the API, and what the API depends on. Anything else won't start.

mdragon: Nice, thanks replacing this PR with #9732. Not the App Name doesn't match up with the docker compose name so just hardcoded that.

Issue #9336, opened 2026-03-30
Integrate Merged Features and Stabilize End-to-End Create Opportunity Flow


ssatia-SGM:
### Summary

This includes:

- Integrating and aligning all merged features across the flow
- Resolving interface mismatches, data flow issues, and dependencies
- Addressing integration-related defects identified during implementation
- Ensuring smooth end-to-end functionality across key user flows
- Coordinating across engineers to close gaps between independently developed components

### Acceptance criteria

TBD @glenblosser-nava Could you add the detailed acceptance criteria per the discussion during 03/30 refinement

ssatia-SGM:
#9396 and #9397 have been created per the acceptance criteria above

brianlee-nava:
This ticket is now unblocked [updated to reflect that]. It is a critical final step to closing out AC2 and AC3 for next Friday's deadline.

Dao-REI:
New tickets were created for the last two ACs (updated to require basic Publish functionality). This ticket is now blocked by #9331, #9669 and #9667.

PR #9739, opened 2026-04-22
Fixing publish workflow

glenblosser-nava:
## Summary

Work for #9336 

## Changes proposed

- Flagging legacy id as optional
- Publishing an opportunity also saves it

## Validation steps

- Log in as an agency user
- Navigate to /opportunities
- Create a new opportunity
- While editing the opportunity, fill in the required fields and click publish without clicking save first
- Run `docker exec grants-api poetry run flask workflow workflow-main`
- The opportunity should appear in search, and should have your entered values

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -151,6 +151,7 @@
       "integrity": "sha512-CGOfOJqWjg2qW/Mb6zNsDm+u5vFQ8DxXfbM09z69p5Z6+mE1ikP2jUXw+j42Pf1XTYED2Rni5f95npYeuwMDQA==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.29.0",
         "@babel/generator": "^7.29.0",
@@ -708,6 +709,7 @@
         }
       ],
       "license": "MIT",
+      "peer": true,
       "engines": {
         "node": ">=18"
       },
@@ -728,6 +730,7 @@
         }
       ],
       "license": "MIT",
+      "peer": true,
       "engines": {
         "node": ">=18"
       }
@@ -770,6 +773,7 @@
       "version": "6.1.2",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "cssesc": "^3.0.0",
         "util-deprecate": "^1.0.2"
@@ -885,6 +889,7 @@
         }
       ],
       "license": "MIT",
+      "peer": true,
       "engines": {
         "node": "^14 || ^16 || >=18"
       },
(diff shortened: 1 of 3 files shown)


frontend/package-lock.json
```diff
@@ -151,7 +151,6 @@
       "integrity": "sha512-CGOfOJqWjg2qW/Mb6zNsDm+u5vFQ8DxXfbM09z69p5Z6+mE1ikP2jUXw+j42Pf1XTYED2Rni5f95npYeuwMDQA==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
```
doug-s-nava, line 154: can you revert the package change?

doug-s-nava (commented):
I feel like making legacy id non-required is fine, but will let a backend user weigh in on that before approving

frontend/src/components/opportunity/OpportunityEditHeaderActions.tsx
desaiamit, line 1: On the frontend change: the intent is right (save before publish) but there's a timing issue. 

- form.requestSubmit() fires the save and returns immediately without waiting for it to finish. So publishOpportunityAction gets called right after, before the save has actually completed on the server. The two requests end up racing each other, which means publish can still win.

**Suggested approach**: the publish call needs to explicitly wait for the save to succeed before firing. One approach is to call the save server action directly (instead of via requestSubmit) and await its result, then only proceed to publish if it returns no errors.

Also, the current PR (#9739) fixes the immediate production bug (500 on search due to legacy_opportunity_id: null), which is the higher priority. 
The frontend race condition is a correctness issue but only affects the edge case where a user clicks Publish without saving first. This can be handled as a follow-up ticket.

desaiamit: Approving the PR - follow-up issue created - https://github.com/HHS/simpler-grants-gov/issues/9747

desaiamit (dismissed):
Created a follow-up ticket - https://github.com/HHS/simpler-grants-gov/issues/9747

desaiamit:
**Save result not checked before publishing**

  `saveOpportunityEditAction` doesn't throw on validation errors - it returns them in the result object (`{ validationErrors: {...} }`). So `.then()` runs regardless of whether save succeeded or failed, meaning publish can still fire even when there are validation errors.

  Suggested fix:
  ```ts
  saveOpportunityEditAction(null, formData)
    .then((result) => {
      if (result?.errorMessage || result?.validationErrors) {
        form.requestSubmit(); // routes through useActionState so errors display in the form UI
        return;
      }
      // proceed to publish
    })

  This also fixes validation errors not displaying to the user.

desaiamit (approved):

PR #9742, opened 2026-04-22
[Issue #7654] Change auth on our opportunity search endpoint to JWT + User Key Auth

kkrug:
## Summary

Fixes for #7654 

## Changes proposed

- Updated opportunity_routes.py to change the auth on `POST /v1/opportunities/search` and `POST /v1/opportunities/search/csv` to `jwt_or_api_user_key_multi_auth`
- Updated test_opportunity_auth.py to add `test_opportunity_unauthorized_401_jwt`
- Updated test_opportunity_route_search.py to replace the `api_auth_token / X-Auth` usage with `user_api_key_id / X-API-Key`

## Context for reviewers

We want to modify our `POST /v1/opportunities/search` endpoint to remove the legacy API key auth, we should instead make it support JWT auth + our new api key auth.

## Validation steps

Confirm tests pass and that the endpoint uses the proper auth via Swagger docs.

### DIFF
--- a/api/src/api/opportunities_v1/opportunity_routes.py
+++ b/api/src/api/opportunities_v1/opportunity_routes.py
@@ -13,7 +13,7 @@ import src.api.opportunities_v1.opportunity_schemas as opportunity_schemas
 import src.api.response as response
 import src.util.datetime_util as datetime_util
 from src.api.opportunities_v1.opportunity_blueprint import opportunity_blueprint
-from src.auth.multi_auth import api_key_multi_auth, jwt_or_api_user_key_multi_auth
+from src.auth.multi_auth import jwt_or_api_user_key_multi_auth
 from src.logging.flask_logger import add_extra_data_to_current_request_logs
 from src.services.opportunities_v1.get_opportunity import (
     get_opportunity,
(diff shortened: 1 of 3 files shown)


chouinar:
@joshtonava - FYI - I think this related to some metrics work in the past - for logged in users, we'd have user-specific search info from this.

api/src/api/opportunities_v1/opportunity_routes.py
```diff
@@ -213,7 +213,7 @@ def _build_opportunities_csv_response(opportunities: Sequence[dict]) -> Response
     examples=examples,
 )
 @opportunity_blueprint.output(opportunity_schemas.OpportunitySearchResponseV1Schema())
-@opportunity_blueprint.auth_required(api_key_multi_auth)
+@opportunity_blueprint.auth_required(jwt_or_api_user_key_multi_auth)
```
chouinar, line 216: Arbitrarily threading, have you tested locally that search works logged in / out via our frontend? I know that search at some point had auth setup a bit different, I think it was fixed, but want to make sure we didn't break search for logged out users.

kkrug: I was able to search locally while both logged in and out

chouinar: After we merge this to staging, we'll want to double-check to be safe just in case there's any configuration weirdness - just want to be very careful to not break search.

chouinar (approved):

Issue #8930, opened 2026-03-09
Backend - Create an endpoint to update award recommendation submission details


nathan-stricker:
### Summary

See [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Create-an-endpoint-to-update-award-recommendation-submission-details)

#### Endpoint
`PUT /alpha/award-recommendations/:award_recommendation_id/submission-details`

#### Purpose
Batch update application submission details

#### AuthZ
Requires `update_award_recommendation` privilege

#### Request
```json
{
  "award_recommendation_submissions": {
    "<uuid>": { # This maps to the award_recommendation_application_submission_id
      "recommended_amount": "", # decimal
      "scoring_comment": "",
      "general_comment": "",
      "award_recommendation_type": "", # AwardRecommendationType enum
      "has_exception": True,
      "exception_detail": "",
    }
  }
}
```

#### Response
```json
{
  "data": [
    {
      # same as the list award recommendation submission function
    }
  ]
}
```

#### Logic

- Fetch the award recommendation, 404 if not found
- AuthZ, 403 on error
- For each of the submission objects in the request:
    - Verify that the `award_recommendation_application_submission` exists for the given ID + verify it’s attached to the award recommendation
    - Replace all values in the submission detail table with the provided values
- If anything errors, no updates should be persisted (if it’s all in one DB transaction - should happen implicitly)

Note - if a submission does not appear in the request, leave it alone. We’re only doing the PUT on the provided fields. Conceptually this endpoint should work the same as if we had a PUT endpoint for an individual submission and each were called one after the other.

### Acceptance criteria

- [ ] Endpoint created
- [ ] Tests created/updated as necessary

PR #9752, opened 2026-04-22
[Issue#8930]Update Award Recommendation Submissions

chay-REIsys:
## Summary

Work for #8930 

## Changes proposed

Batch Update Award Recommendation Submission
## Context for reviewers

We should be able to update multiple update award recommendation submission at once.

## Validation steps

- Fetch changes 
- Fetch award recommendation if from award recommendation table and award recommendation application submission id from award recommendation application submission table 
- make sure tests pass
command:  make test args="tests/src/api/award_recommendations/test_award_recommendation_submission_batch_update.py"

example input:
{
  "award_recommendation_submissions": {
    "46c02038-04c4-4c8c-be25-e980df34c636": {
      "recommended_amount": "175000.00",
      "scoring_comment": "Excellent methodology, strong team qualifications.",
      "general_comment": "Recommend full funding with minor budget adjustments.",
      "award_recommendation_type": "recommended_for_funding",
      "has_exception": false
    },
    "6a40efc6-f3fb-4cb7-875d-ad5a81289dc6": {
      "recommended_amount": "125000.00",
      "scoring_comment": "Promising approach but lacks some implementation details.",
      "general_comment": "Partial funding recommended with revised scope.",
      "award_recommendation_type": "recommended_for_funding",
      "has_exception": true,
      "exception_detail": "Budget requires 15% reduction; recommend focusing on core deliverables."
    },
    "8b263e13-136e-4ba3-b1ff-b5343105ab94": {
      "award_recommendation_type": "not_recommended",
      "scoring_comment": "Proposal does not address core requirements.",
      "general_comment": "Project goals not aligned with program priorities.",
      "has_exception": false
    }
  }
}


api/src/api/award_recommendations_alpha/award_recommendation_routes.py
```diff
@@ -227,3 +232,47 @@ def award_recommendation_risk_create(
...
+            )
+
+        return response.ApiResponse(message="Success", data=submissions)
+    except Exception as e:
+        logger.error(
+            f"Error in batch update endpoint: {str(e)}",
+            extra={"award_recommendation_id": str(award_recommendation_id), "error": str(e)},
+            exc_info=True,
+        )
+        raise
```
chouinar: We don't need a try-except here, actually it would be an issue to include it. Any errors other than 4xx `HTTPError` (what `raise_flask_error` throws) automatically get logged and include a stack trace. For any 4xx errors, you'd be logging an error which would send an alert to be investigated.  A 403 or 404 does not need investigation, those are just user error.

api/src/db/models/lookup_models.py
```diff
@@ -293,6 +293,7 @@
         LookupStr(AwardRecommendationAuditEvent.RISK_UPDATED, 10),
         LookupStr(AwardRecommendationAuditEvent.RISK_DELETED, 11),
         LookupStr(AwardRecommendationAuditEvent.APPLICATION_SUBMISSION_UPDATED, 12),
+        LookupStr(AwardRecommendationAuditEvent.AWARD_RECOMMENDATION_SUBMISSION_UPDATED, 16),
```
chouinar: Nitpick - put this at the bottom so the count increments. That way when we add another it's clear what the next number should be.

api/src/services/award_recommendations/update_award_recommendation_submissions.py
```diff
@@ -0,0 +1,176 @@
...
+    except Exception as e:
+        logger.error(
+            f"Error updating award recommendation submissions: {str(e)}",
+            extra={
+                "award_recommendation_id": str(award_recommendation_id),
+                "error": str(e),
+            },
+            exc_info=True,
+        )
+        raise
```
chouinar: We don't need this try/except either, the error is already added to the logs when any non-4xx exception happens.

api/src/services/award_recommendations/update_award_recommendation_submissions.py
```diff
@@ -0,0 +1,176 @@
...
+                ),
+            )
+            .all()
+        )
+
+        # Build a mapping of submission ID to submission object for quick lookup
+        submissions_map = {
+            str(sub.award_recommendation_application_submission_id): sub
+            for sub in submissions_query
+        }
```
chouinar: We should make sure that the number of records fetched match what was passed in and 404 if not.

api/src/services/award_recommendations/update_award_recommendation_submissions.py
```diff
@@ -0,0 +1,176 @@
...
+
+        # Process updates for each submission in the request
+        updated_submissions = []
+        for submission_id_str, submission_data in update_data.items():
+            submission_id = uuid.UUID(str(submission_id_str))
+
+            # Skip if submission not found in query results
+            if str(submission_id) not in submissions_map:
+                continue
```
chouinar: We shouldn't skip if it's missing, that's just a 404 error.

api/src/services/award_recommendations/update_award_recommendation_submissions.py
```diff
@@ -0,0 +1,176 @@
...
+        # Build a mapping of submission ID to submission object for quick lookup
+        submissions_map = {
+            str(sub.award_recommendation_application_submission_id): sub
+            for sub in submissions_query
+        }
+
+        # Process updates for each submission in the request
+        updated_submissions = []
+        for submission_id_str, submission_data in update_data.items():
+            submission_id = uuid.UUID(str(submission_id_str))
```
chouinar: To make this function a bit shorter, could we move what happens in this for loop to a function. Then we have:

```py
for x in y:
    process_x(x)

def process_x(x):
     ...
```

Which makes it easier to see what one iteration does and makes the logic here a bit clearer.

api/src/services/award_recommendations/update_award_recommendation_submissions.py
```diff
@@ -0,0 +1,176 @@
...
+                AwardRecommendationApplicationSubmission.award_recommendation_application_submission_id.in_(
+                    submission_ids
+                ),
+            )
+            .all()
+        )
+
+        # Build a mapping of submission ID to submission object for quick lookup
+        submissions_map = {
+            str(sub.award_recommendation_application_submission_id): sub
```
chouinar: We make this a string several different times, but I think it's always a UUID when we would receive it since that's the type both in the request and the DB. Can we just leave these all as uuids and avoid the back and forth conversions?

api/src/services/award_recommendations/update_award_recommendation_submissions.py
```diff
@@ -0,0 +1,176 @@
...
+                    "award_recommendation_type"
+                )
+
+            if "has_exception" in submission_data:
+                has_exception_value = submission_data.get("has_exception")
+                if has_exception_value is not None:
+                    submission_detail.has_exception = bool(has_exception_value)
+
+            if "exception_detail" in submission_data:
+                submission_detail.exception_detail = submission_data.get("exception_detail")
```
chouinar: We try to avoid updates and prefer doing PUT/replace. Instead of checking the values, we should just fully replace the values. The request schema validation should only allow valid values and we can just do:

```py
for field, value in submission_data.items():
     setattr(submission_detail, field, value)
```

The types should already be fine, any bools or other types would have been converted by the schema.

api/src/services/award_recommendations/update_award_recommendation_submissions.py
```diff
@@ -0,0 +1,176 @@
...
+
+            # Add audit entry for this update
+            award_recommendation.award_recommendation_audit_events.append(
+                AwardRecommendationAudit(
+                    user=user,
+                    award_recommendation_audit_event=AwardRecommendationAuditEvent.AWARD_RECOMMENDATION_SUBMISSION_UPDATED,
+                    award_recommendation_application_submission=submission,
+                    audit_metadata={
+                        "submission_id": str(submission_id),
```
chouinar: The metadata wouldn't need the submission ID since that's on the submission you added above.

api/src/services/award_recommendations/update_award_recommendation_submissions.py
```diff
@@ -0,0 +1,176 @@
...
+                        AwardRecommendationApplicationSubmission.award_recommendation_submission_detail
+                    ),
+                    selectinload(AwardRecommendationApplicationSubmission.application_submission)
+                    .selectinload(ApplicationSubmission.application)
+                    .selectinload(Application.organization)
+                    .selectinload(Organization.sam_gov_entity),
+                )
+            )
+
+            return list(db_session.execute(stmt).scalars().all())
```
chouinar: Don't requery - just return the list you already have, just move the `selectinload` stuff up above and you have all the data you need.

api/src/api/award_recommendations_alpha/award_recommendation_schemas.py
```diff
@@ -484,3 +484,67 @@ class AwardRecommendationRiskCreateResponseSchema(AbstractResponseSchema):
...
         metadata={"description": "The created award recommendation risk"},
     )
+
+
+class AwardRecommendationSubmissionDetailUpdateSchema(Schema):
+    """Schema for updating a single award recommendation submission detail"""
+
+    recommended_amount = fields.Decimal(
+        allow_none=True,
+        as_string=True,
```
chouinar, line 498: If you don't pass this, it'll be a python decimal which would be easier to work with. Was there something that we wanted this for?

api/src/services/award_recommendations/update_award_recommendation_submissions.py
```diff
@@ -0,0 +1,132 @@
...
+    if len(submissions_map) != len(submission_ids):
+        missing_ids = [
+            submission_id
+            for submission_id in submission_ids
+            if submission_id not in submissions_map
+        ]
+        missing_id_strs = [str(mid) for mid in missing_ids]
+        raise_flask_error(
+            404, message=f"Could not find submission(s) with ID(s): {', '.join(missing_id_strs)}"
+        )
```
chouinar: We don't want the IDs to be in the message. If someone was calling us and wanted to know which IDs were wrong, they wouldn't want to parse a string, we need something machine readable. We did something like this in `_get_validated_submissions`.

For simplicity, the logic in `_get_validated_submissions` that does this should probably be made into a utility.

api/src/services/award_recommendations/update_award_recommendation_submissions.py
```diff
@@ -0,0 +1,132 @@
...
+        logger.info(
+            "Successfully updated award recommendation submissions",
+            extra={
+                "award_recommendation_id": str(award_recommendation_id),
+                "submission_count": len(updated_submissions),
+            },
+        )
+        return updated_submissions
+
+    return []
```
chouinar: If `updated_submissions` is empty, we'd still want to have a log message. If we wanted something like a metric for how many submissions are updated, 0 is useful to know, although I think 0 wouldn't happen as if 0 were passed in, we'd immediately return, and if there's no match with DB data, it errors.

```suggestion
      logger.info(
          "Successfully updated award recommendation submissions",
          extra={
              "award_recommendation_id": award_recommendation_id,
              "submission_count": len(updated_submissions),
          },
      )
      return updated_submissions
```

api/src/api/award_recommendations_alpha/award_recommendation_schemas.py
```diff
@@ -489,6 +489,70 @@ class AwardRecommendationRiskCreateResponseSchema(AbstractResponseSchema):
...
+    """Schema for PUT /alpha/award-recommendations/:award_recommendation_id/submission-details request"""
+
+    award_recommendation_submissions = fields.Dict(
+        keys=fields.UUID(),
+        values=fields.Nested(AwardRecommendationSubmissionDetailUpdateSchema),
+        required=True,
+        metadata={
+            "description": "Dictionary of submission IDs to update details, where keys are award_recommendation_application_submission_id"
+        },
+    )
```
chouinar, line 546: Would it be useful to add?
```py
validate=validators.Length(min=1)
```

To make certain we always have at least 1 item since passing an empty dict isn't really useful.

api/src/services/award_recommendations/utils.py
```diff
@@ -0,0 +1,107 @@
...
+            .filter(
+                AwardRecommendationApplicationSubmission.award_recommendation_id
+                == award_recommendation_id,
+                AwardRecommendationApplicationSubmission.award_recommendation_application_submission_id.in_(
+                    submission_ids_set
+                ),
+            )
+            .options(
+                selectinload(
+                    AwardRecommendationApplicationSubmission.award_recommendation_submission_detail
```
chouinar: I don't think the eager loading check is necessary - the joins it'll do will take a fraction of a second (I'd guess 5ms?) and I think simplicity over optimization wins out there.

We shouldn't use `db_session.query`, use `db_session.execute(select(...))` - it's just the legacy way of querying and sqlalchemy will probably eventually drop it - it's that way in the else of this block.

api/src/services/award_recommendations/utils.py
```diff
@@ -0,0 +1,107 @@
...
+        validation_errors = []
+        for missing_id in missing_ids:
+            validation_errors.append(
+                ValidationErrorDetail(
+                    type=ValidationErrorType.APPLICATION_SUBMISSION_NOT_FOUND,
+                    message=f"Could not find Award Recommendation Application Submission with ID {missing_id}",
+                    field="award_recommendation_application_submission_ids",
+                    value=str(missing_id),
+                )
+            )
```
chouinar: With the util idea I was suggesting, I was thinking just this and below would go into it. We could leave the querying separate if there was a good reason to. Just pass into two list of submission IDs?

api/src/services/award_recommendations/utils.py
```diff
@@ -0,0 +1,107 @@
...
+def get_validated_submissions(
+    db_session: db.Session,
+    award_recommendation_id: uuid.UUID,
+    submission_ids: list[uuid.UUID] | set[uuid.UUID],
+    eager_load: bool = False,
+    return_dict: bool = False,
+) -> (
+    list[AwardRecommendationApplicationSubmission]
+    | dict[uuid.UUID, AwardRecommendationApplicationSubmission]
+):
```
chouinar: Unless there's a strong need, don't return different types from a utility function as that makes the type checking when using it more complex.

In this case, we'd just return the dict. If you need a list, you can grab the `.values()` from the response.

api/src/api/award_recommendations_alpha/award_recommendation_schemas.py
```diff
@@ -489,6 +489,71 @@ class AwardRecommendationRiskResponseSchema(AbstractResponseSchema):
...
+
+
+class AwardRecommendationSubmissionDetailsBatchUpdateRequestSchema(Schema):
+    """Schema for PUT /alpha/award-recommendations/:award_recommendation_id/submission-details request"""
+
+    award_recommendation_submissions = fields.Dict(
+        keys=fields.UUID(),
+        values=fields.Nested(AwardRecommendationSubmissionDetailUpdateSchema),
+        required=True,
+        validate=validate.Length(min=1),
```
kkrug: @chouinar Should we be using our custom validator like this?
```
from src.api.schemas.extension import validators
     validate=validators.Length(min=1),
```
I thought it was odd the test for this was asserting a 500 response when it should be 422. There's another instance of this in this file from one of the other award rec endpoints we just did, but we can fix that on a new ticket.

chouinar (dismissed):

api/tests/src/api/award_recommendations/test_award_recommendation_submission_batch_update.py
```diff
@@ -0,0 +1,672 @@
...
+
+        # Check audit record details
+        audit = audit_records[0]
+        assert audit.created_at is not None, "Audit record timestamp is missing"
+        assert audit.audit_metadata is not None, "Audit metadata should be present"
+        assert (
+            "updated_fields" in audit.audit_metadata
+        ), "Audit should contain updated fields list in metadata"
+
+    def test_batch_update_empty_submissions_200(
```
kkrug: Can we rename this test to not have the 200 response code and move it with the other error response tests?

api/src/services/award_recommendations/update_award_recommendation_submissions.py
```diff
@@ -0,0 +1,117 @@
...
+
+        db_session.add(submission_detail)
+        updated_submissions.append(submission)
+
+    db_session.flush()
+
+    logger.info(
+        "Successfully updated award recommendation submissions",
+        extra={
+            "award_recommendation_id": str(award_recommendation_id),
```
kkrug: It doesn't hurt but casting a a string here isn't needed.
```suggestion
            "award_recommendation_id": award_recommendation_id,
```

api/tests/src/api/award_recommendations/test_award_recommendation_submission_batch_update.py
```diff
@@ -0,0 +1,672 @@
...
+        assert "award_recommendation_application_submission_id" in sub1
+        assert "application_submission" in sub1
+        assert "submission_detail" in sub1
+
+        # Check second submission updates
+        sub2 = updated_submissions[str(submission2.award_recommendation_application_submission_id)]
+        # Now that we're properly loading relationships, submission_detail should be present
+        assert "award_recommendation_application_submission_id" in sub2
+        assert "application_submission" in sub2
+        assert "submission_detail" in sub2
```
kkrug, line 175: Can we add assertions for actual values rather than just the keys?

api/tests/src/api/award_recommendations/test_award_recommendation_submission_batch_update.py
```diff
@@ -0,0 +1,672 @@
...
+        assert detail2.exception_detail is None, "Exception detail was not updated correctly"
+        assert (
+            detail2.general_comment == "Updated: Reconsidered after discussion"
+        ), f"General comment was not updated correctly. Expected: 'Updated: Reconsidered after discussion', Got: '{detail2.general_comment}'"
+
+        # Allow for potential differences in enum handling between environments
+        assert (
+            detail2.award_recommendation_type == AwardRecommendationType.RECOMMENDED_FOR_FUNDING
+            or detail2.award_recommendation_type.value == "recommended_for_funding"
+        ), f"Award recommendation type was not updated correctly. Expected: {AwardRecommendationType.RECOMMENDED_FOR_FUNDING}, Got: {detail2.award_recommendation_type}"
```
kkrug: Why would we need to be asserting using an `or` here (and also another time on line ~345)? Shouldn't the enum be returned the same way consistently?

api/src/services/award_recommendations/update_award_recommendation_submissions.py
```diff
@@ -0,0 +1,117 @@
...
+    award_recommendation = get_award_recommendation_and_verify_access(
+        db_session, user, award_recommendation_id
+    )
+
+    agency = award_recommendation.opportunity.agency_record
+    verify_access(user, {Privilege.UPDATE_AWARD_RECOMMENDATION}, agency)
+
+    submission_ids = list(update_data.keys())
+    if not submission_ids:
+        return []
```
kkrug: With the schema validation, we shouldn't need this anymore.

api/tests/src/api/award_recommendations/test_award_recommendation_submission_batch_update.py
```diff
@@ -0,0 +1,672 @@
...
+                },
+            }
+        }
+
+        resp = client.put(
+            f"{API_URL}/{award_recommendation.award_recommendation_id}/submission-details",
+            json=update_data,
+            headers={"X-SGG-Token": token},
+        )
+        # pdb.set_trace()
```
kkrug: Can we remove this if it is not needed?

api/tests/src/api/award_recommendations/test_award_recommendation_submission_batch_update.py
```diff
@@ -0,0 +1,672 @@
...
+        assert "application_submission" in sub2
+        assert "submission_detail" in sub2
+
+        # Note: For test client requests with SQLAlchemy, we need to clear the session cache
+        # to ensure we get fresh data from the database in the same test method
+        db_session.expire_all()
+
+        # Query directly from database to ensure we get fresh data
+        query1 = (
+            db_session.query(AwardRecommendationSubmissionDetail)
```
kkrug: A bit of a nit but we tend to use `db_session.execute(select(...))` instead of `db_session.query()` throughout the project ( there are 2 other instances of these queries below).

kkrug (approved):

PR #9760, opened 2026-04-22
[Issue #9143] Add endpoint to update a risk on an award recommendation

nathan-stricker:
## Summary

Work for #9143 

## Changes proposed

Created an endpoint to update a risk on an Award Recommendation and added associated tests

## Context for reviewers

See [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Risks)

## Validation steps

- [ ] Confirm implementation aligns with specifications outlined in the [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Risks) (note that I used the variable name `comment` instead of `summary` in the request/response to remain consistent with the data model and the create endpoint)
- [ ] Review created tests


api/src/api/award_recommendations_alpha/award_recommendation_schemas.py
```diff
@@ -450,8 +450,33 @@ class AwardRecommendationRiskCreateRequestSchema(Schema):
     )
 
 
-class AwardRecommendationRiskCreateResponseDataSchema(Schema):
-    """Schema for the created risk response data"""
+class AwardRecommendationRiskUpdateRequestSchema(Schema):
```
kkrug: `AwardRecommendationRiskUpdateRequestSchema` and `AwardRecommendationRiskCreateRequestSchema` are basically identical - do we want to reuse/inherit the schema? I'm ok with keeping as is for clear separation especially if there is a chance for divergence in the future.

api/src/services/award_recommendations/update_award_recommendation_risk.py
```diff
@@ -0,0 +1,98 @@
...
+) -> AwardRecommendationRisk:
+    _get_award_recommendation_for_update(db_session, user, award_recommendation_id)
+
+    risk = _get_risk_for_update(db_session, award_recommendation_id, award_recommendation_risk_id)
+
+    risk.comment = request_data["comment"]
+    risk.award_recommendation_risk_type = request_data["award_recommendation_risk_type"]
+
+    submission_ids = set(request_data["award_recommendation_application_submission_ids"])
+    validated_submissions = _get_validated_submissions(
```
chouinar: If `_get_validated_submissions` should be used outside of that file, could we remove the `_`?

api/src/services/award_recommendations/update_award_recommendation_risk.py
```diff
@@ -0,0 +1,98 @@
...
+        for rs in risk.award_recommendation_risk_submissions
+        if rs.award_recommendation_application_submission_id not in ids_to_remove
+    ]
+
+    for sid in ids_to_add:
+        risk.award_recommendation_risk_submissions.append(
+            AwardRecommendationRiskSubmission(
+                award_recommendation_application_submission=validated_submissions_by_id[sid],
+            )
+        )
```
chouinar, line 70: I think we could simplify this a little. Don't need the submission IDs because the dict has that in its keys I think.

Then I think you can adjust the set operations to just do the checks in the loop, although not certain whether that's better or worse necessarily.

```suggestion
def _sync_risk_submissions(
    risk: AwardRecommendationRisk,
    validated_submissions_by_id: dict[uuid.UUID, AwardRecommendationApplicationSubmission],
) -> None:
    existing_ids = {
        rs.award_recommendation_application_submission_id
        for rs in risk.award_recommendation_risk_submissions
    }
    
    risk.award_recommendation_risk_submissions = [
        rs
        for rs in risk.award_recommendation_risk_submissions
        if rs.award_recommendation_application_submission_id not in validated_submissions_by_id
    ]

    for id, award_recommendation_application_submission in validated_submissions_by_id.items():
        if id not in existing_ids:
          risk.award_recommendation_risk_submissions.append(
              AwardRecommendationRiskSubmission(
                  award_recommendation_application_submission=award_recommendation_application_submission,
              )
        )
```

api/src/api/award_recommendations_alpha/award_recommendation_schemas.py
nathan-stricker: You know, I combined the create/update response schemas into`AwardRecommendationRiskResponseSchema` but didn't do the same for the request schema and I'm not quite sure why- I will combine them and we can split them out again in the future if need be

api/src/services/award_recommendations/update_award_recommendation_risk.py
nathan-stricker: Good catch- will make that change and do the same for `_get_award_recommendation_for_update`

kkrug (approved):

chouinar (approved):

PR #9775, opened 2026-04-23
[Issue #8740] Annotate vision.spec.ts with feature references

### LINKED ISSUE #8740: Annotate vision spec with feature file references
### Summary

Depends on:

* feature file existing for functionality
* guidance around annotation of spec files

Based on annotation guidance, update spec file with references to feature file / steps

### Acceptance criteria

- [ ] spec is updated with annotations

Bhavna-Ramachandran:
## Summary
Work for: Task #8740 

## Changes proposed
- Update /workspaces/simpler-grants-gov/frontend/tests/e2e/vision.spec.ts to add annotation header with references to feature file
- Add inline comments consistent with feature file

## Context for reviewers
- Reviewers should make sure that the spec file has annotation header and inline comments in spec file reflects the flow in feature file.

## Validation steps
- Confirm the spec file references the feature file


frontend/tests/e2e/vision.spec.ts
```diff
@@ -1,59 +1,90 @@
+/**
+ * @feature Vision page
+ * @featureFile frontend/tests/e2e/vision.feature
```
aoysimmons-GH, line 3: I am thinking we should do the file name without path to make it easier re-organize location when we expand in the future without comeback to change the path to each files.

aoysimmons-GH (approved):

frontend/tests/e2e/vision.spec.ts
Bhavna-Ramachandran: I understand the concern about refactors making paths brittle, and that makes sense for long-term maintenance.

I was thinking we could maintain consistency across the repo - reason is if the feature file and spec stay in same directory this is fine, but there are feature files which has spec in different directory, so if we remove paths entirely we lose traceability in those cases.

I am open to aligning on whichever convention we prefer, as long as it’s consistent across the repo.

Issue #9144, opened 2026-03-19
Backend - Create an endpoint to delete a risk on an award recommendation


nathan-stricker:
### Summary

See [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Create-an-endpoint-to-delete-a-risk-on-an-award-recommendation)

#### Endpoint
`DELETE /alpha/award-recommendations/:award_recommendation_id/risks/:award_recommendation_risk_id`

#### Purpose
Delete an award recommendation risk

#### AuthZ
Requires `update_award_recommendation` privilege

#### Request
n/a

#### Response
```json
{
  "data": null,
  "message": "Success",
  "status_code": 200
}
```

#### Logic
- Fetch the award recommendation, 404 if not found
- AuthZ, 403 on error
- Fetch the award recommendation risk, 404 if not found
- Set the `is_deleted` flag to True

### Acceptance criteria

- [ ] Endpoint created
- [ ] Tests created/updated as necessary

PR #9839, opened 2026-04-24
[Issue #9144] Add endpoint to delete a risk on an award recommendation

aaronbrown-nava:
## Summary

Work for [#9144](https://github.com/HHS/simpler-grants-gov/issues/9144) 

## Changes proposed

Created an endpoint to delete an Award Recommendation Risk and added tests

## Context for reviewers

See [TechSpec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Risks)

## Validation steps

- [ ] Confirm implementation aligns with specifications outlined in the [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Risks) 
- [ ]  Review created tests


api/src/services/award_recommendations/delete_award_recommendation_risk.py
```diff
@@ -0,0 +1,63 @@
...
+from src.adapters import db
+from src.api.route_utils import raise_flask_error
+from src.auth.endpoint_access_util import verify_access
+from src.constants.lookup_constants import Privilege
+from src.db.models.award_recommendation_models import AwardRecommendation, AwardRecommendationRisk
+from src.db.models.opportunity_models import Opportunity
+from src.db.models.user_models import User
+
+
+def _get_award_recommendation_risk_for_update(
```
chouinar: This function is probably something we'd want more central since an update endpoint would also use it (I don't think that's been built yet?). We could make the `update_award_recommendation_risk.py` file and throw it in there for when we do that.

api/src/api/award_recommendations_alpha/award_recommendation_routes.py
```diff
@@ -313,3 +317,40 @@ def award_recommendation_audit_list(
     return response.ApiResponse(
         message="Success", data=audit_events, pagination_info=pagination_info
     )
+
+
+@award_recommendation_blueprint.delete(
+    "/award-recommendations/<uuid:award_recommendation_id>/risks/<uuid:award_recommendation_risk_id>"
+)
+@award_recommendation_blueprint.output(AbstractResponseSchema)
```
chouinar: While we don't strictly need anything more here, adding a specific schema like:

```py
class AwardRecommendationRiskDeleteResponseSchema(AbstractResponseSchema):
     data = fields.MixinField(metadata={"example": None})
```

is useful as it makes it so the OpenAPI/swagger docs have a named type for the response.

api/src/services/award_recommendations/delete_award_recommendation_risk.py
kkrug: Update actually exists already - you can see _get_risk_for_update here: https://github.com/HHS/simpler-grants-gov/blob/246cf0c771e997f1ecf260cc5a732c830c8e53e1/api/src/services/award_recommendations/update_award_recommendation_risk.py#L20
Update also reuses in `get_award_recommendation_for_update` from create, so the same could be done here. Using these would also add `is_deleted.isnot(True)` for the risk and parent award rec which is needed to avoid "deleting" an already deleted risk.

api/tests/src/api/award_recommendations/test_award_recommendation_risk_delete.py
```diff
@@ -0,0 +1,221 @@
...
+        ).scalar_one()
+        assert deleted_risk.is_deleted is True
+
+
+####################################
+# 404 Tests
+####################################
+
+
+class TestDeleteAwardRecommendationRisk404:
```
kkrug, line 94: Once the changes are made to reuse those get functions we mentioned in the service, you can add a test for the situation of deleting and already deleted risk since it will now have the `is_deleted` filter. Similarly, we can add a test for deleting a risk on a deleted award rec (both of these tests should return 404).

aaronbrown-nava: tests added

api/src/api/award_recommendations_alpha/award_recommendation_routes.py
aaronbrown-nava: type added

api/src/services/award_recommendations/delete_award_recommendation_risk.py
aaronbrown-nava: Yeah update exists now, wasn't there when I started. Reusing it now. I've also removed the _ prefix on it

kkrug:
@aaronbrown-nava just a heads up once someone has reviewed a PR and you make changes, you can re-request a review by clicking the circular arrows next to the reviewers name in the top right. This will put it back in their PR queue so it shows up [here](https://github.com/pulls/review-requested)

kkrug (approved):

Issue #8826, opened 2026-03-05
Adjust SQS/s3 configurations to have a consistent/shared pattern for common fields


chouinar:
### Summary

For connecting to SQS and s3 we have configurations that load environment variables for a variety of cases.

Importantly we have an `S3_ENDPOINT_URL` that we use locally to point our clients to localstack rather than real AWS. This is even the name for SQS - we should change this to be more generically named.

We also have a few env vars for AWS things like region scattered around that we should centralize somewhere.

The endpoint URLs will stay service-specific since S3 mock and SQS mock run on different hosts, but they should have a standardized naming convention (e.g. prefix with `AWS_`). Both fields should also use `Field(alias=...)`, which is currently missing on S3. 

The rest of this work encompasses:

- `AWS_REGION` moves into `BaseAwsConfig` as the single source of truth
- `get_boto_session()` uses that config region instead of hardcoding `us-west-2`
- `S3Config` and `SQSConfig` read region from `BaseAwsConfig` rather than defining their ow
- local.env gets `AWS_REGION=us-east-1` explicitly se
- S3 client gets `region_name` passed to it (matching what SQS already does)

### Acceptance criteria

- [ ] `AWS_REGION` is defined in one place (`BaseAwsConfig`) and consumed by all AWS adapters
- [ ] Endpoint URL fields follow a consistent naming pattern across S3 and SQS configs
- [ ] local.env is updated with the new env var names and includes `AWS_REGION`
- [ ] S3 client passes `region_name` to `session.client()` like SQS already does
- [ ] `get_boto_session()` uses the configured region instead of hardcoding us-west-2
- [ ] Test fixture `reset_aws_env_vars` in conftest.py is updated to match new env var names

PR #9847, opened 2026-04-24
[Issue #8826] Adjust SQS/s3 configurations to have a consistent/shared pattern for common fields

kkrug:
## Summary

Fixes for #8826 

## Changes proposed

- Updated local.env to rename environment variables
- Updated init.py to add `get_base_aws_config` to exports
- Updated aws_session.py to incorporate `aws_region`
- Updated s3_adapter.py to change variable name to `aws_s3_endpoint_url` and incorporate `get_base_aws_config()`
- Updated sqs_adapter.py to change variable name to `aws_sqs_endpoint_url` and incorporate `get_base_aws_config()`
- Updated file_util.py to change variable names
- Updated conftest.py to change environment variable names and reset `base_aws_config` with other env vars
- Updated test_sqs_adapter.py to correctly call url variable 

## Context for reviewers

For connecting to SQS and s3 we have configurations that load environment variables for a variety of cases that we wanted to centralize and standardize.

## Validation steps

Confirm full test suite passes.

### DIFF
--- a/api/src/adapters/aws/aws_session.py
+++ b/api/src/adapters/aws/aws_session.py
@@ -4,29 +4,29 @@ from pydantic import Field
 from src.util.env_config import PydanticBaseEnvConfig
 
 
-class BaseAwsConfig(PydanticBaseEnvConfig):
+class AwsConfig(PydanticBaseEnvConfig):
     is_local_aws: bool = False
     aws_region: str = Field(alias="AWS_REGION", default="us-east-1")
 
 
-_base_aws_config: BaseAwsConfig | None = None
+_aws_config: AwsConfig | None = None
 
 
-def get_base_aws_config() -> BaseAwsConfig:
-    global _base_aws_config
-    if _base_aws_config is None:
-        _base_aws_config = BaseAwsConfig()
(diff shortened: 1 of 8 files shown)


api/src/adapters/aws/aws_session.py
```diff
@@ -1,10 +1,12 @@
 import boto3
+from pydantic import Field
 
 from src.util.env_config import PydanticBaseEnvConfig
 
 
 class BaseAwsConfig(PydanticBaseEnvConfig):
```
chouinar: Haven't looked too close, but I wonder if this "Base" config should actually be a true base with the s3/sqs configs derived from it.

Not sure if that would help the patterns, but would make the naming a bit clearer. Otherwise maybe we rename this just `AwsConfig`?

chouinar (dismissed):
LGTM, a suggestion added if we want to incorporate it.

api/src/adapters/aws/aws_session.py
kkrug: I went with renaming to just `AwsConfig`.

chouinar (approved):

Issue #8626, opened 2026-02-23
RBAC based restriction for list of funding opportunities


kevlyu:
### Summary

Implement RBAC to ensure that only the funding opportunities displayed on screen, during the AR creation process, is limited to the role and agency of the user.

### Acceptance criteria

- [ ] RBAC implemented

kevlyu:
@kevlyu check with Michael on this in tech spec and may be able to use P&P codes with added filter on state of the NOFO

kevlyu:
@doug-s-nava check on #8006

doug-s-nava:
The goal of this ticket is to allow retrieval of all opportunities which:
* belong to a given agency
* are not in draft
* are available for starting an award recommendation

There is an existing endpoint that allows retrieval of all opportunities which:
* belong to a given agency
* are either in draft or published

As such, the work for this would be to modify the existing endpoint to:
* support a filter for draft status, defaulting to allowing all statuses if filter value is not provided
* support for a filter for award recommendation readiness status

Depending on how the backend changes are implemented, this could break existing uses of the agency opportunities fetch endpoint, and may require frontend changes to fix that.

doug-s-nava:
`available for starting an award recommendation` = 
* status == published
* is_simpler_grants_opportunity == true
* at least 1 competition for the opportunity has at least 1 submission (???)
* no associated award recommendations

AC:

- [ ] existing grantor opportunity GET endpoint is extended with a filter for `available for award recommendation`, defaulting to ANY (meaning that if you don't provide the filter value, it returns all opportunities regardless of their award rec availability status)

KtejanKella:
@chouinar is the reviewer.

PR #9850, opened 2026-04-24
[Issue#8626] opportunity endpoint update for ar

chay-REIsys:
## Summary

Work for #8626 

## Changes proposed

Modify Opportunity get endpoint to retrieve opportunities which are ready for award recommendation
No tests were created earlier for the endpoint so i went ahead and added tests
## Context for reviewers

Added filter flag award recommendation ready , if it is passed only opportunities which are posted and has submissions

## Validation steps

- fetch new changes
- run make remake-backend && docker compose up 
- test the endpoint http://localhost:8080/v1/grantors/agencies/{agencyId}/opportunities
- authorize the endpoint specific to agency 
- executing the endpoint without any filters should return all opportunities but with
     "award_recommendation_ready": true
 it would bring opportunities that are posted with atleast one submission and doesn't have an existing award recommendation 
 -run make test args="tests/src/api/opportunities_grantors_v1/test_opportunity_route_list.py -v" to make sure tests pass


api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
```diff
@@ -245,6 +245,16 @@ class OpportunityListRequestSchema(Schema):
...
 
+    filters = fields.Dict(
+        metadata={
+            "description": "Optional filters for opportunity list",
+            "examples": {
+                "draft_status": "Boolean value to filter by draft status. If not provided, shows all opportunities.",
+                "award_recommendation_ready": "Boolean value to filter for opportunities ready for award recommendations.",
+            },
+        }
+    )
```
chouinar, line 265: For any filters, use our existing filter utilities - we want to have the same pattern for ease of use by our callers. I know when it comes to booleans, a list is weird since `[true, false]` isn't really meaningful (we _should_ come up with a pattern for exact rather than just `one_of`)

That would be using the builders to construct the fields like:

```py
class OpportunityListFilterSchema(Schema):
     draft_status = fields.Nested(BoolSearchSchemaBuilder("DraftStatusFilterSchema").with_one_of(example=False).build()
     award_recommendation_ready = fields.Nested(BoolSearchSchemaBuilder("AwardRecommendationReadySchema").with_one_of(example=True).build()
```

Then use that schema here for filters.

api/src/services/opportunities_grantor_v1/get_opportunity_list.py
```diff
@@ -25,8 +25,10 @@
 class OpportunityFilterSchema(BaseModel):
     """Optional filters for opportunity list"""
 
-    # Placeholder for any required filters
-    pass
+    award_recommendation_ready: bool | None = Field(
+        default=None,
+        description="Filter for opportunities ready for award recommendations. If True, returns only opportunities that can have award recommendations created for them.",
+    )
```
chouinar: If you adjust the schema like I mentioned in the request, we have  a`BoolSearchFilter` class for the filter type/shape.

api/src/services/opportunities_grantor_v1/get_opportunity_list.py
```diff
@@ -87,10 +95,20 @@ def list_opportunities_with_filters(
         .where(Agency.agency_id == agency_id)
     )
 
-    # Apply sorting in the database query
+    if params.filters and params.filters.award_recommendation_ready is True:
+        stmt = stmt.where(
+            Opportunity.current_opportunity_summary.has(
+                CurrentOpportunitySummary.opportunity_status == OpportunityStatus.POSTED
+            ),
```
chouinar: Do not look at the opportunity status, the opportunity status is purely a value for search and has no meaning to anything in apply. An opportunity can have applications and have never been in search - nothing in the current opportunity summary / summary classes should be factored into this. Think of the summary as a copy of info about the opportunity, it doesn't have any actual authority, it's just info (the post/close dates also aren't used, that's from the competition).

I think the only checks that matter you have right:
* Opportunity has application submissions - there might be additional nuance here about whether the submissions are still valid, but I think we ignore that for now since we haven't put that concept into the data yet.
* Is one of our opportunities - although we don't import application submissions, so the above check would account for it, safe to be thorough
* No existing award recs

api/src/services/opportunities_grantor_v1/get_opportunity_list.py
```diff
@@ -136,8 +153,12 @@ def get_opportunity_list_for_grantors(
...
     agency = get_agency(db_session, agency_id)
 
-    # Verify user has VIEW_PRIVILEGE for the agency
-    verify_access(user, {Privilege.VIEW_OPPORTUNITY}, agency)
+    # If filtering by award_recommendation_ready, verify user has VIEW_AWARD_RECOMMENDATION privilege
+    if params.filters and params.filters.award_recommendation_ready:
+        verify_access(user, {Privilege.VIEW_AWARD_RECOMMENDATION}, agency)
+    else:
+        # Verify user has VIEW_OPPORTUNITY privilege for the agency
+        verify_access(user, {Privilege.VIEW_OPPORTUNITY}, agency)
```
chouinar: I wouldn't make any sort of branching logic like this - keep authZ simple. If the filter significantly changes functionality, we either want a privilege dedicated to just this endpoint, or we want a separate endpoint for the purposes of this filter.

api/tests/lib/seed_award_recommendations.py
```diff
@@ -437,11 +455,99 @@ def _add_application_to_award_recommendation(
...
+    competition = factories.CompetitionFactory.create(
+        opportunity=opportunity,
+    )
+
+    logger.info("Creating 5 organizations for applications")
+    organizations = []
+    for _ in range(5):
+        unique_uei = f"UEI{random.randint(10000000, 99999999):08d}"
+        org = factories.OrganizationFactory.create(sam_gov_entity__uei=unique_uei)
+        organizations.append(org)
```
chouinar: Let's move this into the factory itself. The org already makes its entity, but I'm guessing the issue is that because it's a sequence, that causes issues with other UEIs already created in the DB?

```suggestion
    organizations = OrganizationFactory.create_batch(size=5)
```

However we'd need a separate change - the `SamGovEntityFactory` should be changed so that uei isn't based on a sequence anymore and is instead random. Can do that with something like:

```py
uei = factory.LazyFunction(
        lambda: f"UEI{fake.random_int(min=1, max=999999999):09d}"
    )
```

chouinar (approved):

PR #9855, opened 2026-04-25
[Issue #8265] Fixed different filter behavior.

### LINKED ISSUE #8265: Filters do not behave the same when same selection is selected twice
### Summary

For some filters, notably Posted date, closing date and cost sharing, when selecting the same option twice, it does not reset the filters (like on agency, eligbility, funding instrument) 

### Reproducibility

### How to reproduce the bug

1. Naviate to simpler.grants.gov/search 
2. Open the filter drawer in the top right 
3. Select an Agency on the Agency filter 
4. Re-click on the same selection 
5. The radio buttong automatically moves back into Any agency 
6. Now select Closing Date range
7. Select any option
8. Re-click on the same option
9. Filter does not reset like Agency 

### Browser

Chrome

### Code of Conduct

MozamilS:
## Summary

The difference in filter behavior documented in #8265 was due to the difference in the selection toggle functions declared in `CheckboxFilter.tsx` (functioning correctly) and in `RadioButtonFilter.tsx` (minimal functionality). Two changes were made to fix the behavior of `RadioButtonFilter.tsx` to mirror `CheckboxFilter.tsx`

Fixes #8265 

## Changes proposed

1) In `src/components/search/SearchFilterRadio.tsx` the toggle selection function was assigned to an `onChange` prop. This caused issues since the toggle selection function was not getting triggered the second time for the selected radio button because nothing was being changed to trigger the event handler. An `onClick` prop was added to fix this issue and the `toggleRadioSelection` function was changed to take `React.MouseEvent` instead of `React.ChangeEvent`.

2) The `toggleRadioSelection` function was changed to de-select the clicked radio button and assign empty string as the value for `setQueryParam()`. This causes the default radio button to be selected.

## Context for reviewers

The `RadioButtonFilter.tsx` did not have a completed toggle selection function. This caused a difference between the selection behavior compared to `CheckboxFilter.tsx`. The difference was that for a selected selection, selecting it the second time would cause the default selection to be automatically selected. This was not happening in the filters that used radio buttons. 

## Validation steps

Clicking on a radio button filter twice will now select the default selection automatically.

### DIFF
--- a/frontend/src/components/search/Filters/RadioButtonFilter.tsx
+++ b/frontend/src/components/search/Filters/RadioButtonFilter.tsx
@@ -21,8 +21,15 @@ export function RadioButtonFilter({
 }: SearchFilterProps) {
   const { setQueryParam } = useSearchParamUpdater();
 
-  const toggleRadioSelection = (event: React.ChangeEvent<HTMLInputElement>) => {
-    setQueryParam(queryParamKey, event.target.value);
+  const toggleRadioSelection = (event: React.MouseEvent<HTMLInputElement>) => {
+    const value = event.currentTarget.value;
+
+    if (query.has(value)) {
+      setQueryParam(queryParamKey, "");
+      event.currentTarget.checked = false;
+    } else {
+      setQueryParam(queryParamKey, value);
+    }
   };
 
   const isNoneSelected = useMemo(() => query.size === 0, [query]);
@@ -44,7 +51,7 @@ export function RadioButtonFilter({
                 id={`${title}-any-radio`}
                 name={`Any ${title}`}
                 label={`Any ${title.toLowerCase()}`}
-                onChange={toggleRadioSelection}
+                onClick={toggleRadioSelection}
                 value={undefined}
                 facetCount={undefined}
                 checked={isNoneSelected}
@@ -57,7 +64,7 @@ export function RadioButtonFilter({
                 id={option.id}
                 name={option.label}
                 label={option.label}
-                onChange={toggleRadioSelection}
+                onClick={toggleRadioSelection}
                 value={option.value}
                 facetCount={facetCounts && (facetCounts?.[option.value] || 0)}
                 checked={query.has(option.value)}
--- a/frontend/src/components/search/Filters/RadioButtonFilter.tsx
+++ b/frontend/src/components/search/Filters/RadioButtonFilter.tsx
@@ -21,16 +21,21 @@ export function RadioButtonFilter({
 }: SearchFilterProps) {
   const { setQueryParam } = useSearchParamUpdater();
 
-  const toggleRadioSelection = (event: React.MouseEvent<HTMLInputElement>) => {
-    const value = event.currentTarget.value;
-
-    if (query.has(value)) {
-      setQueryParam(queryParamKey, "");
-      event.currentTarget.checked = false;
-    } else {
-      setQueryParam(queryParamKey, value);
-    }
+  const toggleRadioSelection = (event: React.ChangeEvent<HTMLInputElement>) => {
+		  const value = event.target.value;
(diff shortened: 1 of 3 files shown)


btabaska:
Thank you for this contribution, I have gone ahead and notified the team and we will get this reviewed as soon as possible.

ErinPattisonNava (commented):
Code looks good, but could you add unittests for the new behavior?

MozamilS:
> Code looks good, but could you add unittests for the new behavior?

Hello, 

I'll try and add the unit tests tomorrow, God willing. I'll ping you when I push the commit.

MozamilS:
@ErinPattisonNava 

Hello, 

When testing the new behavior, I found that since all `onChange` props were changed to `onClick`, a lot of the tests were failing since they depended on the `onChange` behavior. 

To fix this, a new function `handleRadioClickedTwice` was made to encapsulate the new behavior and it was only applied on non-default radio buttons.

A new test case was made for this new behavior and all tests pass successfully.

frontend/src/components/search/Filters/RadioButtonFilter.tsx
```diff
@@ -22,7 +22,20 @@ export function RadioButtonFilter({
...
 
   const toggleRadioSelection = (event: React.ChangeEvent<HTMLInputElement>) => {
-    setQueryParam(queryParamKey, event.target.value);
+    const value = event.target.value;
+    if (!query.has(value)) {
+      setQueryParam(queryParamKey, event.target.value);
+    }
+  };
+
+  const handleRadioClickedTwice = (
```
btabaska, line 31: The second-click deselect logic depends on query.has(value) inside onClick, but query is URL/router-driven and can lag behind fast user clicks. If someone clicks the same radio twice quickly, the second click can be ignored (because query is still stale), so behavior becomes timing-dependent instead of deterministic.

frontend/src/components/search/Filters/RadioButtonFilter.tsx
```diff
@@ -22,7 +22,20 @@ export function RadioButtonFilter({
...
+    event: React.MouseEvent<HTMLInputElement>,
+  ) => {
+    const value = event.currentTarget.value;
+    if ((event.currentTarget as HTMLInputElement).checked && query.has(value)) {
+      setQueryParam(queryParamKey, "");
+      event.currentTarget.checked = false;
+    }
```
btabaska, line 38: We’re mutating event.currentTarget.checked = false on a controlled input (checked={query.has(option.value)}). That creates two sources of truth (DOM + React state) and can cause transient mismatch/flicker. We should let state updates drive checked status only.

btabaska (changes requested):
The main blockers are event-order/state-sync issues in radio deselection logic; once those are fixed, this looks done. Thank you for your contribution!

MozamilS:
@btabaska 

Thank you so much for your review and feedback. 

I thought just checking whether the radio button is checked or not would be a sufficient fix; There will be only one source of truth and it shouldn't cause any delays on the frontend. 

Please let me know what you think.

MozamilS:
@btabaska 

Hello and sorry to bother you again,

I'm not sure why the E2E tests are failing now. I checked the logs and they're trying to get `MozamilS/8265-filter-behavior-fix*` with the additional wildcard `*` at the end. Is this causing the problem?

btabaska:
The E2E test failures are not your fault, an engineer on our team made a change and it is causing fork branches to fail a git hash check. We are looking to get a fix out soon, but please ignore them until then. Thank you again for all your hard work on this project. We really appreciate you

btabaska (approved):
Overall these changes LGTM, thanks for working with us on this one. Merging. Thank you for the contribution!

PR #9866, opened 2026-04-27
[Unticketed] Add linting rule to exclude imports where we've got derived implementations

chouinar:
## Changes proposed
* Added certain imports to a configuration that will cause a linting error if used.

## Context for reviewers
We create our own derived versions of these validators and schemas with changed error handling and formatting. We don't want to mistakenly import the wrong ones as the behavior is undefined for our error handling (probably just 500s in certain cases).

Note that we do use these to make our implementation, so added an ignore where necessary.

## Validation steps
Just run `make lint`. Try adding these imports in other files and it'll complain.

### DIFF
--- a/api/pyproject.toml
+++ b/api/pyproject.toml
@@ -191,6 +191,8 @@ allowed-confusables = [
 # For these libraries, we do use them, but have our own derived versions we want
 # to use instead as we've rewritten fundamental components like error messaging.
 "marshmallow.validate".msg = "Do not import marshmallow.validate directly - instead use src.api.schemas.extension.validators as we've rewritten the error messaging"
+"apiflask.validators".msg = "Do not import apiflask.validators directly - instead use src.api.schemas.extension.validators as we've rewritten the error messaging"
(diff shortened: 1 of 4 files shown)


api/pyproject.toml
```diff
@@ -187,6 +187,11 @@ allowed-confusables = [
     "PT018"
 ]
 
+[tool.ruff.lint.flake8-tidy-imports.banned-api]
+# For these libraries, we do use them, but have our own derived versions we want
+# to use instead as we've rewritten fundamental components like error messaging.
+"marshmallow.validate".msg = "Do not import marshmallow.validate directly - instead use src.api.schemas.extension.validators as we've rewritten the error messaging"
+"apiflask.Schema".msg = "Do not import apiflask.Schema directly - instead use src.api.schemas.extension"
```
kkrug, line 197: Do we want to consider adding a couple of other ones while we are doing this?
```
"apiflask.fields".msg = "Do not import apiflask.fields directly - instead use src.api.schemas.extension.fields"
"apiflask.validators".msg = "Do not import apiflask.validators directly - instead use src.api.schemas.extension.validators"
```

kkrug (approved):

Issue #9869, opened 2026-04-27
Update AwardRecommendationRiskRequestSchema to use proper validation


kkrug:
### Summary

When the `AwardRecommendationRiskRequestSchema` was first created, the validation to prevent an empty list from being passed in was using the Marshmallow `validate` instead of our custom `validators` found in `src.api.schemas.extension`. This causes a response code of 500 instead of the expected 422. The import `from src.api.schemas.extension.field_validators import Length` is already present in the schema file. This ticket will:

- Update to use `AwardRecommendationRiskRequestSchema` to use `validate=[Length(min=1)]`
- Add an empty list 422 test to test_award_recommendation_risk_create.py
- Add an empty list 422 test to test_award_recommendation_risk_update.py

### Acceptance criteria

- [x] `AwardRecommendationRiskRequestSchema` is updated to use `validate=[Length(min=1)]`
- [x] 422 test is added to test_award_recommendation_risk_create.py
- [x] 422 test is added to test_award_recommendation_risk_update.py

PR #9870, opened 2026-04-27
[Issue #9869] Update AwardRecommendationRiskRequestSchema to use proper validation

kkrug:
## Summary

Fixes for #9869 

## Changes proposed

- Updated award_recommendation_schemas.py to update `AwardRecommendationRiskRequestSchema` to use `validate=[Length(min=1)]`
- Updated test_award_recommendation_risk_create.py to add empty list 422 test
- Updated test_award_recommendation_risk_update.py to add empty list 422 test

## Context for reviewers

When the `AwardRecommendationRiskRequestSchema` was first created, the validation to prevent an empty list from being passed in was using the Marshmallow validate instead of our custom validators found in `src.api.schemas.extension`. This causes a response code of 500 instead of the expected 422.

## Validation steps

Confirm tests pass. The endpoints can also be tested via Swagger, where the input of an empty list should return a 422 response.

### DIFF
--- a/api/src/api/award_recommendations_alpha/award_recommendation_schemas.py
+++ b/api/src/api/award_recommendations_alpha/award_recommendation_schemas.py
@@ -1,5 +1,3 @@
-from marshmallow import validate
-
 from src.api.schemas.extension import Schema, fields
 from src.api.schemas.extension.field_validators import Length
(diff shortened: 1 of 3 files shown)


nathan-stricker (approved):
LGTM!

PR #9899, opened 2026-04-27
Dschrashun/9596 authorization gate docs

doug-s-nava:
## Summary

Fixes / Work for #ISSUE


frontend/src/app/[locale]/(base)/workspace/organizations/[id]/manage-users/page.test.tsx
```diff
@@ -140,37 +106,93 @@ describe("manage-users page", () => {
...
-    expect(gateProps.requiredPrivileges).toEqual([
-      {
-        resourceId: "org-123",
-        resourceType: "organization",
-        privilege: "manage_org_members",
+    expect(ManageUsersPageContentMock).toHaveBeenCalledWith({
+      organizationId: "org-123",
+      authorizedData: {
+        fetchedResources: {
+          invitedUsersList: {
```
Comment (ErinPattisonNava) [inline frontend/src/app/[locale]/(base)/workspace/organizations/[id]/manage-users/page.test.tsx:113]: For readability it would be nice if this could be pulled out as variables we could do deep comparisons against, instead of repeated.

frontend/src/app/[locale]/(base)/workspace/organizations/[id]/manage-users/page.test.tsx
```diff
@@ -46,91 +30,73 @@ jest.mock("next/navigation", () => ({
...
+const getOrganizationPendingInvitationsMock = jest.fn().mockResolvedValue({});
+const getOrganizationUsersMock = jest.fn().mockResolvedValue({});
+const getOrganizationRolesMock = jest.fn().mockResolvedValue({});
 
-type RequiredPrivilege = {
-  resourceId: string;
-  resourceType: string;
-  privilege: string;
-};
+jest.mock("src/services/fetch/fetchers/organizationsFetcher", () => ({
```
Comment (ErinPattisonNava) [inline frontend/src/app/[locale]/(base)/workspace/organizations/[id]/manage-users/page.test.tsx:41]: Love that this is simpler.

documentation/frontend/frontend-authorization-checks.md
```diff
@@ -0,0 +1,305 @@
...
+			users: {
+				data: { ...fetchedUserData },
+				status: 200
+			}
+		}
+	}
+}
+```
+
+If any call results in a 403, this info will be passed to the page, or, if an onUnauthorized callback is provided to the gate, the result of that callback will be rendered instead of the page.
```
ErinPattisonNava, line 78: Above it says if any call succeeds the page gets the data, but here it says if any call results in a 403 the onUnauthorized callback is triggered and displayed instead of the page.  If one call succeeds and one fails, what happens?

documentation/frontend/frontend-authorization-checks.md
```diff
@@ -0,0 +1,305 @@
...
+  - if a user shouldn't have access to a page based on their permissions, the gate can prevent them from seeing any page content
+  - note that authentication based gating is built in as well. Any user that is not logged in with a valid session token will be prevented from seeing any page content
+- checks for access to any authorization gated data required by a page
+  - the gate will make calls for any authorized data that the page will need, and can
+- checks for any specific permissions required by the page
+  - the gate will make calls to check for any number of user permissions that are relevant to loading or displaying the page
+
+### General usage
+
+AuthorizationGate is wrapper around a page content child, and depending on how the gate is configured, either supplies the page content component with authorization related data as a prop, or renders the result of a callback function.
```
ErinPattisonNava, line 19: Most of the examples have been at the page level, but would it work around any child component?

documentation/frontend/frontend-authorization-checks.md
```diff
@@ -0,0 +1,305 @@
...
+			users: {
+				error: E,
+				status: 403
+			}
+		}
+	}
+}
+```
+
+If any call errors with a non-403, the error will be passed to the page via props.
```
ErinPattisonNava, line 94: There is an OnError callback that I couldn't get to work correctly when I was building the manage-users/legacy page.  Are we deprecating that?

documentation/frontend/frontend-authorization-checks.md
```diff
@@ -0,0 +1,305 @@
...
+
+AuthorizationGate can take an array of promises that represent API calls for authorization gated resources that the page will use in one way or another. The gate will resolve the promise, and:
+
+- if successful or resolved with a non 403 error
+  - pass the result via props to the gate's child component
+- if unauthorized, resolved with a 403 either
+  - pass the result via props to the gate's child component
+  - gate access to the page by returning
+
+For example, a "manage users" page will need to fetch a list of active users, and a list of invited users from the API, but this data will only be available to users with proper privileges. A promise for these API calls can be passed to the AuthorizationGate.
```
ErinPattisonNava, line 45: My biggest concern with our current pattern is that it breaks the explicit props expectation in the child component it is wrapping.  Instead of declaring what data the child component expects, the child hunts around in the returned data for what it needs.  This makes it difficult to re-use that child, as you need to understand the interior code to know what resources to send in.  Understood that changing the architecture is beyond the scope here, but are there conventions we could add that would make the data contract more clear?

documentation/frontend/frontend-authorization-checks.md
doug-s-nava: I can clarify this but if any of the resource calls fails with a 403, and there is an onUnauthorized callback provided, the gate will run the callback instead of rendering the children. There's currently no way to opt in or out of the `onUnauthorized` handling for a specific call. If we needed to do that we'd need to make that call outside of the gate

documentation/frontend/frontend-authorization-checks.md
doug-s-nava: That's a fair point, and it doesn't sit quite right with me either.

In a regular client side React app we'd solve that issue by using a context. Unfortunately, server components can't pull date from context so that limits the effectiveness of that approach.

We've talked about removing the support that's currently in the AuthorizationGate for a client side context for consuming the output rather than prop drilling, as it adds a layer of complexity we likely don't need. The assumption being that any direct children of AuthorizationGate will be server components, and we can prop drill from there. However, we could look at this the opposite way, where any consumer of the AuthorizationGate output needs to be a client component. I think this could work, the issues I see are something like:

* no component would be able to both consume the auth gate's output AND make server side requests
* potential performance degradation with more data forced to render client side (may not actually be a concern)
* maybe other things I haven't considered...

documentation/frontend/frontend-authorization-checks.md
doug-s-nava: onError should work, but it's limited, as it is only used to catch errors thrown outside of the resource promises themselves. The pattern for error handling for the resource fetch calls is for the gate to catch them and pass them along to the children so that they can deal with them downstream. I figured the complexity of trying to deal with multiple errors would add too much complexity here - probably would need to support error handlers for each promise.

Thinking about it now I wonder if adding `catch`es to each call when creating the promises would work, but even if it did it seems pretty complex to reason about: you're creating a promise that is resolved within the gate, but you're embedding the error handling into the promise itself when you create it.

I'd love to hear how people would like this piece of things to work, as I am not sure exactly what the priorities are here in terms of error handling.

documentation/frontend/frontend-authorization-checks.md
doug-s-nava: The main limitation is that since the AuthorizationGate is a server component we won't be able to include it in a client component.

There are different patterns we could explore, but to keep things simple I kept the guidance here focused on a pattern of including the gate at the top page level. We should talk about what use cases we might have where that doesn't cut it. The main one that comes to mind is where we want different suspensions and loading states for different parts of a page (which we could still support with this component in most cases in my estimation), but with lack of guidance from design around loading state priorities I didn't consider that something we needed to address directly.

documentation/frontend/frontend-authorization-checks.md
doug-s-nava: Build expected interactions with errors with an example

documentation/frontend/frontend-authorization-checks.md
doug-s-nava: establish a convention for child components expressing the shape of the authorized data that it should accept

PR #9911, opened 2026-04-28
[Issue #9904] Dual-validate URLs in CG transformation to fix search 500

### LINKED ISSUE #9904: CommonGrants: 500 error on `/common-grants/search` endpoint
### Summary

I routinely get a 500 error message when I search for `closed` opportunities with the search term `open data`

### Reproducibility

### How to reproduce the bug

Run the following curl command after updating `<YOUR_API_TOKEN>`:

```bash
curl -X 'POST' \
  'https://api.simpler.grants.gov/common-grants/opportunities/search' \
  -H 'accept: application/json' \
  -H 'X-API-Key: <YOUR_API_TOKEN>' \
  -H 'Content-Type: application/json' \
  -d '{
  "filters": {
    "status": {
      "operator": "in",
      "value": [
        "closed"
      ]
    }
  },
  "pagination": {
    "page": 22,
    "pageSize": 1
  },
  "search": "open data"
}'
```

Image

Notably the record before and after 22 are fine:

```bash
curl -X 'POST' \
  'https://api.simpler.grants.gov/common-grants/opportunities/search' \
  -H 'accept: application/json' \
  -H 'X-API-Key: <YOUR_API_TOKEN>' \
  -H 'Content-Type: application/json' \
  -d '{
  "filters": {
    "status": {
      "operator": "in",
      "value": [
        "closed"
      ]
    }
  },
  "pagination": {
    "page": 23,
    "pageSize": 1
  },
  "search": "open data"
}'
```

```bash
curl -X 'POST' \
  'https://api.simpler.grants.gov/common-grants/opportunities/search' \
  -H 'accept: application/json' \
  -H 'X-API-Key: <YOUR_API_TOKEN>' \
  -H 'Content-Type: application/json' \
  -d '{
  "filters": {
    "status": {
      "operator": "in",
      "value": [
        "closed"
      ]
    }
  },
  "pagination": {
    "page": 23,
    "pageSize": 1
  },
  "search": "open data"
}'
```

Image

### Browser

Not applicable

### Code of Conduct

SnowboardTechie:
## Summary

Fixes #9904.

`POST /common-grants/opportunities/search` was 500ing when the result page contained a URL that pydantic accepts but marshmallow rejects (most often comma-separated URLs from grants.gov data). The CG response is loaded back through marshmallow on the way out, so one bad record killed the whole batch. Fix validates URLs against both libraries at the transformation source so divergent strings get dropped instead of crashing the response.

## Changes proposed

- New `api/src/services/common_grants/url_utils.py` with `validate_url_compatible` — a dual-validator that requires both pydantic `HttpUrl` (strict) and marshmallow `fields.URL` to accept the input, returning the pydantic-normalized form so the exact string downstream callers see is the one the response load re-validates.
- `validate_url` and `AttachmentValue.validate_download_url` both delegate to the new helper. Removed `UrlValidator` and `_AttachmentUrlValidator` (consolidated onto one strict-mode pydantic validator). Introduced a single module-level `_marshmallow_url_field` singleton in `url_utils.py` for the dual-validation path.
- URL-validation log: fixed message string at WARNING severity; URL plus field path and `opportunity_id` in `extra` (so operators can answer "why is this field missing for opportunity X?" without paging the source record).
- Tests: `test_url_utils.py` (6 unit tests for `validate_url_compatible` covering both divergent cases — comma-separated URLs and scheme-only-host) and additions to `test_transformation.py` (divergent-URL rejection at `validate_url`, end-to-end reproduction through `OpportunitiesSearchResponseSchema().load(...)`, the attachment sibling case, and a guard that the drop-log carries `field` + `opportunity_id`).

## Context for reviewers

Prod traceback (from #9904):
```
ValidationError({'items': {0: {'customFields': {'additionalInfo': {'value':
{'url': [...message='Not a valid URL.']}}}}}})
```
Stack pointed at `OpportunitiesSearchResponseSchema().load(response_json)` in [common_grants_routes.py:173](api/src/api/common_grants/common_grants_routes.py#L173).

Confirmed divergent inputs (pydantic accepts, marshmallow rejects):
- `https://a.gov,https://b.gov` — pydantic mangles to `https://a.gov,https//b.gov`
- `http://example` — pydantic accepts TLD-less hosts

Two behavior changes worth flagging:
- `AttachmentValue.downloadUrl` now stores the pydantic-normalized form (trailing slash, lowercase host) instead of raw input — matches what `validate_url` has done since #6987.
- Attachment URLs now run through strict-mode pydantic. The deleted `_AttachmentUrlValidator` was non-strict; the marshmallow load already rejected curly-brace URLs etc., this just moves the rejection earlier.

Suggested follow-ups (not in this PR):
- Lock the marshmallow side to `http`/`https` (the dual-check already restricts to those via pydantic strict `HttpUrl`, but defense-in-depth is cheap).
- Surface a `dataWarnings` envelope so consumers can see suppressed fields.
- Apply per-record fail-soft to v1 `search_opportunities` `SCHEMA.load(many=True)` for the same atomic-batch brittleness.

## Validation steps

```bash
make test args="tests/src/services/common_grants -v"
```
Suite covers the dual-validator unit cases and the prod-failing reproduction through the marshmallow response load.

After deploy, confirm the prod-failing payload returns 200:
```bash
curl -X POST 'https://api.simpler.grants.gov/common-grants/opportunities/search' \
  -H 'X-API-Key: <KEY>' -H 'Content-Type: application/json' \
  -d '{"filters":{"status":{"operator":"in","value":["closed"]}},"pagination":{"page":22,"pageSize":1},"search":"open data"}'
```
Expect 200, a WARNING log with `cg_event=url_validation_error`, the URL, the field path, and the `opportunity_id`, and the New Relic "API Prod - Error Logs" alert to clear.

### DIFF
--- a/api/src/services/common_grants/url_utils.py
+++ b/api/src/services/common_grants/url_utils.py
@@ -9,7 +9,7 @@ exposes a single helper that requires both validators to accept, so divergent
 strings are dropped at the source.
 """
 
-from urllib.parse import urlunsplit, urlsplit
+from urllib.parse import urlsplit, urlunsplit
 
 from marshmallow import ValidationError as MarshmallowValidationError
 from marshmallow import fields as marshmallow_fields
--- a/api/src/services/common_grants/url_utils.py
+++ b/api/src/services/common_grants/url_utils.py
@@ -76,4 +76,8 @@ def redact_url_userinfo(value: str | None) -> str:
             host = f"{host}:{parts.port}"
         return urlunsplit((parts.scheme, host, parts.path, parts.query, parts.fragment))
     except ValueError:
-        return "<unparseable-url>"
+        # The URL could not be parsed (e.g. a non-numeric port). We can't
+        # safely strip credentials from a string we can't parse, so the
+        # original value is suppressed entirely. The sentinel is greppable
+        # in log aggregation.
+        return "<malformed-url-redaction-skipped>"
--- a/api/src/services/common_grants/url_utils.py
+++ b/api/src/services/common_grants/url_utils.py
@@ -4,9 +4,11 @@ Pydantic's HttpUrl and marshmallow's fields.URL disagree on a handful of inputs
 (comma-separated URLs, scheme-only-host like "http://example", etc.). Pydantic
 accepts and normalizes; marshmallow rejects. Because the CG response is loaded
 through marshmallow on the way out, any URL accepted by pydantic but rejected
-by marshmallow blows up the entire batch with a 500 (issue #9904). This module
-exposes a single helper that requires both validators to accept, so divergent
-strings are dropped at the source.
+by marshmallow blows up the entire batch with a 500. This module exposes a
+single helper that requires both validators to accept, so divergent strings
+are dropped at the source.
+
+Background and reproduction: https://github.com/HHS/simpler-grants-gov/issues/9904
 """
 
 from urllib.parse import urlsplit, urlunsplit
@@ -34,10 +36,9 @@ def validate_url_compatible(value: str | None) -> str | None:
     marshmallow's fields.URL; otherwise None.
 
     The CG response is loaded through marshmallow on the way out, so a URL
-    accepted only by pydantic blows up the entire batch (issue #9904).
-    Returning the pydantic-normalized form means downstream callers see a
-    consistent string and that exact string is what the response load
-    re-validates.
+    accepted only by pydantic blows up the entire batch. Returning the
+    pydantic-normalized form means downstream callers see a consistent string
+    and that exact string is what the response load re-validates.
     """
     if value is None or value == "":
         return None
--- a/api/src/services/common_grants/url_utils.py
+++ b/api/src/services/common_grants/url_utils.py
@@ -17,7 +17,11 @@ from pydantic import BaseModel, Field, HttpUrl, ValidationError
 
 
 class _UrlValidator(BaseModel):
-    """Pydantic strict HttpUrl validator."""
+    """Pydantic strict HttpUrl validator.
+
+    Mirrors the HttpUrl field in OpportunityBase and other CommonGrants models.
+    TODO(@widal001): Replace this with a new field from SDK or relax strictness in SDK.
+    """
 
     url: HttpUrl = Field(strict=True)
 
@@ -55,16 +59,21 @@ def redact_url_userinfo(value: str | None) -> str:
     validation-failure logs verbatim — leaking the credentials to log
     aggregation. Strip userinfo here. Falls back to a fixed sentinel if the
     URL is unparseable so the caller always has something logger-safe.
+
+    Note: ``SplitResult.port`` lazily raises ``ValueError`` for non-numeric
+    ports (e.g. ``host:abc``), so the *entire* parse + reconstruct sequence
+    runs inside the try — re-raising here would re-introduce the fail-hard
+    behavior this module exists to remove.
     """
     if value is None:
         return "<none>"
     try:
         parts = urlsplit(value)
+        if not parts.username and not parts.password:
+            return value
+        host = parts.hostname or ""
+        if parts.port:
+            host = f"{host}:{parts.port}"
+        return urlunsplit((parts.scheme, host, parts.path, parts.query, parts.fragment))
     except ValueError:
         return "<unparseable-url>"
-    if not parts.username and not parts.password:
-        return value
-    host = parts.hostname or ""
-    if parts.port:
-        host = f"{host}:{parts.port}"
-    return urlunsplit((parts.scheme, host, parts.path, parts.query, parts.fragment))
--- a/api/src/services/common_grants/url_utils.py
+++ b/api/src/services/common_grants/url_utils.py
@@ -1,0 +1,70 @@
+"""URL validation shared across CG transformation and CG custom-field validators.
+
+Pydantic's HttpUrl and marshmallow's fields.URL disagree on a handful of inputs
+(comma-separated URLs, scheme-only-host like "http://example", etc.). Pydantic
+accepts and normalizes; marshmallow rejects. Because the CG response is loaded
+through marshmallow on the way out, any URL accepted by pydantic but rejected
+by marshmallow blows up the entire batch with a 500 (issue #9904). This module
+exposes a single helper that requires both validators to accept, so divergent
+strings are dropped at the source.
+"""
+
+from urllib.parse import urlunsplit, urlsplit
+
+from marshmallow import ValidationError as MarshmallowValidationError
+from marshmallow import fields as marshmallow_fields
+from pydantic import BaseModel, Field, HttpUrl, ValidationError
+
+
+class _UrlValidator(BaseModel):
+    """Pydantic strict HttpUrl validator."""
+
+    url: HttpUrl = Field(strict=True)
+
+
+_marshmallow_url_field = marshmallow_fields.URL()
+
+
+def validate_url_compatible(value: str | None) -> str | None:
+    """Return the normalized URL if accepted by *both* pydantic HttpUrl and
+    marshmallow's fields.URL; otherwise None.
+
+    The CG response is loaded through marshmallow on the way out, so a URL
+    accepted only by pydantic blows up the entire batch (issue #9904).
+    Returning the pydantic-normalized form means downstream callers see a
+    consistent string and that exact string is what the response load
+    re-validates.
+    """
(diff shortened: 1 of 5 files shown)


api/src/services/common_grants/url_utils.py
```diff
@@ -0,0 +1,79 @@
...
+    try:
+        parts = urlsplit(value)
+        if not parts.username and not parts.password:
+            return value
+        host = parts.hostname or ""
+        if parts.port:
+            host = f"{host}:{parts.port}"
+        return urlunsplit((parts.scheme, host, parts.path, parts.query, parts.fragment))
+    except ValueError:
+        return "<unparseable-url>"
```
jcrichlake: Could we update this string to be a little more descriptive as to what happened?

api/tests/src/services/common_grants/test_transformation.py
```diff
@@ -715,6 +715,155 @@ def test_validate_url_with_nasa_url_bug(self):
         assert old_result is not None, "_legacy_validate_url() should accept NASA URL"
         assert new_result is None, "validate_url() should reject NASA URL"
 
+    def test_validate_url_rejects_urls_that_marshmallow_url_field_rejects(self):
+        """Reproduces issue #9904: validate_url() must also reject URLs that pass
```
jcrichlake: I'd say let's not mention issue numbers in code comments.  I think it's confusing given that there are different project boards and most of the code changes in this repo are not going to trace back to the same board that most of the folks who maintain this board use. 

If you really wanted some form of traceability you could link to the public issue URL

SnowboardTechie: Fair point, I'll consolidate them down to a single full url link to clear the noise a bit

api/src/services/common_grants/url_utils.py
```diff
@@ -0,0 +1,79 @@
...
+    if value is None:
+        return "<none>"
+    try:
+        parts = urlsplit(value)
+        if not parts.username and not parts.password:
+            return value
+        host = parts.hostname or ""
+        if parts.port:
+            host = f"{host}:{parts.port}"
+        return urlunsplit((parts.scheme, host, parts.path, parts.query, parts.fragment))
```
jcrichlake: It may be too challenging, but I'm wondering if we should log something that we are dropping data here and perhaps what type of data is being dropped.  It could help us troubleshoot if an API user in the future starts asking why info that they expect to be there is being dropped

SnowboardTechie: Good call. Done in [90eff5276](https://github.com/HHS/simpler-grants-gov/pull/9911/commits/90eff5276):

- Log message bumped from INFO `"URL validation failed"` to WARNING `"Dropping URL field from CG response: failed dual validation"`. Data suppression is a warning, not a routine validation note.
- `extra` now carries `field` (one of `source`, `customFields.additionalInfo.url`, `customFields.attachments[].downloadUrl`) and `opportunity_id`, so a log query can correlate "field X went missing for opportunity Y".
- New test pins all three field paths in WARNING records with `opportunity_id` attached.

That covers the operator side. For the consumer side, surfacing a `dataWarnings` envelope on the response would be a real refactor and probably needs team alignment plus an ADR.

jcrichlake:
Overall really nice job getting these changes up quickly and overall I really like the pattern you've established! I especially like the cleaning up of the URL to attempt to avoid logging any sensitive information

jcrichlake (dismissed):

api/src/services/common_grants/url_utils.py
```diff
@@ -0,0 +1,84 @@
...
+from marshmallow import ValidationError as MarshmallowValidationError
+from marshmallow import fields as marshmallow_fields
+from pydantic import BaseModel, Field, HttpUrl, ValidationError
+
+
+class _UrlValidator(BaseModel):
+    """Pydantic strict HttpUrl validator.
+
+    Mirrors the HttpUrl field in OpportunityBase and other CommonGrants models.
+    TODO(@widal001): Replace this with a new field from SDK or relax strictness in SDK.
```
widal001: Small note but if we're moving this we should probably either drop this TODO or change the target user to @SnowboardTechie or @jcrichlake

api/src/services/common_grants/url_utils.py
```diff
@@ -0,0 +1,84 @@
...
+    except (ValidationError, MarshmallowValidationError):
+        return None
+
+
+def redact_url_userinfo(value: str | None) -> str:
+    """Strip ``user:password@`` from a URL before logging.
+
+    Grants.gov data occasionally puts free text or markdown into URL fields.
+    If a record contains ``https://user:secret@host/path`` (intentional or
+    accidental), the raw string flows through pydantic and lands in
```
widal001: @SnowboardTechie I like the spirit of this extra measure to limit leaking sensitive, but this is the one section of this code that makes me a bit nervous for a few reasons:

- The existing logger already has some sanitization capabilities built in. Have we checked that something like this isn't already happening at the logger level?
- This kind of hand-rolled sanitization feels like an area where we run the risk of introducing new bugs in an attempt to fix previous bugs
- Also FWIW, grants.gov and SGG already display the full unredacted version of these strings to anyone who visits the site, so this feels like we're bending over backward to redact what (at least from the point of view of the system) is already considered public information.

If there's a desire to sanitize URLs for potential usernames and passwords, it feels like it's better handled in a separate PR at the logger level or at the SGG service layer where we can really stress test this.

api/src/services/common_grants/url_utils.py
SnowboardTechie: Done in [139371bed](https://github.com/HHS/simpler-grants-gov/pull/9911/commits/139371bed). Dropped the TODO entirely — it had no concrete next action and would have just rotted attached to whoever's name was on it.

api/src/services/common_grants/url_utils.py
SnowboardTechie: Agreed — pulled the helper out in [139371bed](https://github.com/HHS/simpler-grants-gov/pull/9911/commits/139371bed). `validate_url` now logs the raw URL in `extra["url"]`.

You're right that grants.gov and SGG already serve these unredacted, so the threat model didn't justify the bug surface. The `SplitResult.port` ValueError I tripped on mid-review was a sign this was the wrong layer — if we do want URL scrubbing later, the logger or service layer is the right place to do it once across the whole app.

Net: -154 / +6, helper gone, redaction-specific tests gone, the dual-validation tests still cover the actual fix.

api/tests/src/services/common_grants/test_url_utils.py
```diff
@@ -0,0 +1,26 @@
...
+        # Pydantic accepts (mangled), marshmallow rejects → must return None.
+        assert validate_url_compatible("https://a.gov,https://b.gov") is None
+
+    def test_rejects_scheme_only_host(self):
+        # Pydantic accepts TLD-less hosts, marshmallow rejects.
+        assert validate_url_compatible("http://example") is None
+
+    def test_rejects_garbage(self):
+        assert validate_url_compatible("not-a-url") is None
+        assert validate_url_compatible("sam.gov") is None  # missing scheme
```
widal001, line 37: Just to confirm: Does one of these closely match the actual URL that caused the 500 error?

Also small suggestion: It might help in the tests to demonstrate that it passes pydantic and fails marshmallow to recreate the issue and prove that our fix actually addresses the bug

SnowboardTechie: Honest answer on the first part: we don't know the literal prod URL. Couldn't isolate the offending record without API access, and grants.gov public-API sort order diverges from CG so we couldn't reverse-engineer it. The two synthetic forms in the test are confirmed-divergent representatives, and the end-to-end test in `test_transformation.py` produces a `ValidationError` with the byte-identical shape and message as the New Relic trace. Confident the failure mode is captured even if the exact string isn't.

Good suggestion on the second part. Added a parametrized test in [111785d45](https://github.com/HHS/simpler-grants-gov/pull/9911/commits/111785d45) that asserts each input passes pydantic alone, fails marshmallow alone with the exact prod `Not a valid URL.` message, then `validate_url_compatible` drops it. Pins the divergence and would catch a regression that loosens either side.

widal001 (approved):
Thanks for making the changes I requested!

While I haven't been able to stress test this locally, the code changes look good to me. My one request is that after you merge, you work with @jcrichlake to run the suite of tests he has that validate these changes in staging to make sure there weren't any regressions or new bugs introduced.

PR #9934, opened 2026-04-28
[unticketed]: dedupe nr api logs

prasnava:
## Summary

Resolve the duplicate logs issue for the api service. Updated fluentbit and terraform to bypass sending logs directly to newrelic from the fluentbit container since we have the lambda cloudwatch subscription that is forwarding the logs. 

## Validation steps

Successfully deployed to the dev environment: [here](https://github.com/HHS/simpler-grants-gov/actions/runs/25075766953/job/73469652136) and logs are flowing.

### DIFF
--- a/fluentbit/fluentbit.yml
+++ b/fluentbit/fluentbit.yml
@@ -40,7 +40,7 @@ pipeline:
 
   outputs:
     - name: nrlogs # new relic logs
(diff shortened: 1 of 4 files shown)


mdragon (approved):
👍🏻

PR #9936, opened 2026-04-28
[Issue #9590] Edit opportunity page followups

### LINKED ISSUE #9590: Edit Opportunity Page General Followups
### Summary

After merging https://github.com/HHS/simpler-grants-gov/pull/9334 there are a few stray things to fix up that are not accounted for in the larger followup tickets (TBD & TBD)

- add unit test for getMultiValueField
- establish a single type / interface for `{ label: string, value: string }` and use it anywhere this type is needed across the application
- reuse eligibility grouping already defined in opportunity detail definitions
- use shared input components and DynamicFieldLabel

### Acceptance criteria

- [ ] unit tests for `getMultiValueField` are created
- [ ] opportunityEditForm has 90% unit test coverage across statements, branches, functions and lines
- [ ] only one type / interface exists for the definition `{ label: string, value: string }`
- [ ] edit opportunity eligibility section uses groups as defined in opportunity detail definitions
- [ ] static opportunity related values (opportunity categories, eligibilities, funding categories, funding instruments) are referenced from a shared location (not grantor or grantee specific)
- [ ] editOpportunityForm uses shared input components and DynamicFieldLabel
- [ ] edit opportunity page works as expected

desaiamit:
## Summary

Fixes #9590

## Changes proposed

**AC1 - Single canonical type for `{ label, value }`**
- Added `LabelValueOption` to `generalTypes.ts` as the single shared definition
- `FilterOption` in `searchFilterTypes.ts` now extends `LabelValueOption` instead of redeclaring `label` and `value`
- Removed the local `SelectOption` / `CheckboxOption` definitions from `opportunityEditFormConfig.ts`

**AC2 - Eligibility grouping driven by data**
- Replaced 5 hardcoded `filter()` calls in `OpportunityEditForm.tsx` with a single `reduce()` over `eligibilityTypes`, which already carries a `group` field per entry
- Fixed a pre-existing mismatch: the form was keying the miscellaneous group as `misc` while `eligibilityTypes` uses `miscellaneous`

**AC3 - Static options defined once, at the source**
- Moved `OPPORTUNITY_CATEGORY_OPTIONS` from `opportunityEditFormConfig.ts` to `constants/opportunity.ts`
- Removed `FUNDING_INSTRUMENT_OPTIONS` and `FUNDING_CATEGORY_OPTIONS` wrapper re-exports from `opportunityEditFormConfig.ts` - `OpportunityEditForm.tsx` now imports `fundingOptions` / `categoryOptions` directly from `searchFilterOptions.ts`
- Fixed a pre-existing trailing space in the "Procurement Contract" label in `fundingOptions`

**AC4 - Shared input components and `DynamicFieldLabel`**
- `DynamicFieldLabel` (from `applyForm/widgets`) is already used consistently across every interactive field in `OpportunityEditForm.tsx` - no changes needed here
- `CommonTextInput`, `CommonSelectInput`, `CommonTextArea`, and `CommonCharacterCount` from `CommonFormFields.tsx` were intentionally not adopted in this PR. These components originated in the Create Opportunity flow and are not yet established as a shared component library (no formal API, docs, or Storybook coverage at the `frontend/` level). When a proper shared USWDS wrapper library is ready, all pages using raw USWDS primitives - including this one - will be migrated in one deliberate pass. Adopting these components prematurely would mean the edit form goes through that migration twice, which is the tech debt we are trying to avoid.

## Context for reviewers

These changes are purely organizational - no new behavior is introduced. The goal across all four ACs is to establish a single source of truth: one type definition for `{ label, value }`, one place for each options array, and eligibility grouping driven by the data that already exists in `constants/opportunity.ts`. The edit form's logic is unchanged; only where things are defined and imported from has moved.

## Validation steps

1. Navigate to an opportunity edit page (e.g. `/opportunity/<id>/edit`)
2. Verify all sections render correctly: Funding Details, Dates, Eligibility, Description, Contact
3. Verify eligibility checkboxes are correctly grouped (Government, Education, Business, Nonprofit, Miscellaneous) and the Miscellaneous group is populated
4. Make edits and save - confirm changes persist
5. Run `npm run all-checks` from `frontend/` - all 280 test suites pass

### DIFF
--- a/frontend/src/components/opportunity/OpportunityEditForm.tsx
+++ b/frontend/src/components/opportunity/OpportunityEditForm.tsx
@@ -4,11 +4,11 @@ import {
   saveOpportunityEditAction,
   type OpportunityEditValidationErrors,
 } from "src/app/[locale]/(base)/opportunity/[id]/edit/actions";
-import { eligibilityTypes } from "src/constants/opportunity";
 import {
   categoryOptions,
+  eligibilityTypes,
   fundingOptions,
-} from "src/constants/searchFilterOptions";
+} from "src/constants/opportunity";
 import { OpportunityAttachment } from "src/types/opportunity/opportunityAttachmentTypes";
 
 import { useTranslations } from "next-intl";
--- a/frontend/src/components/opportunity/OpportunityEditForm.tsx
+++ b/frontend/src/components/opportunity/OpportunityEditForm.tsx
@@ -7,6 +7,7 @@ import {
 import {
   categoryOptions,
   eligbilityValueToGroup,
+  ELIGIBILITY_OPTIONS,
   fundingOptions,
 } from "src/constants/opportunity";
 import { OpportunityAttachment } from "src/types/opportunity/opportunityAttachmentTypes";
@@ -29,7 +30,6 @@ import {
 import { DynamicFieldLabel } from "src/components/applyForm/widgets/DynamicFieldLabel";
 import { OpportunityAttachmentUploadInput } from "src/components/opportunity/OpportunityAttachmentUploadInput";
 import {
-  ELIGIBILITY_OPTIONS,
   OPPORTUNITY_CATEGORY_OPTIONS,
   OpportunityEditFormValues,
 } from "./opportunityEditFormConfig";
--- a/frontend/src/components/opportunity/OpportunityEditForm.tsx
+++ b/frontend/src/components/opportunity/OpportunityEditForm.tsx
@@ -5,6 +5,10 @@ import {
   type OpportunityEditValidationErrors,
 } from "src/app/[locale]/(base)/opportunity/[id]/edit/actions";
 import { eligibilityTypes } from "src/constants/opportunity";
+import {
+  categoryOptions,
+  fundingOptions,
+} from "src/constants/searchFilterOptions";
 import { OpportunityAttachment } from "src/types/opportunity/opportunityAttachmentTypes";
 
 import { useTranslations } from "next-intl";
@@ -26,8 +30,6 @@ import { DynamicFieldLabel } from "src/components/applyForm/widgets/DynamicField
 import { OpportunityAttachmentUploadInput } from "src/components/opportunity/OpportunityAttachmentUploadInput";
 import {
   ELIGIBILITY_OPTIONS,
-  FUNDING_CATEGORY_OPTIONS,
-  FUNDING_INSTRUMENT_OPTIONS,
   OPPORTUNITY_CATEGORY_OPTIONS,
   OpportunityEditFormValues,
 } from "./opportunityEditFormConfig";
@@ -404,7 +406,7 @@ export default function OpportunityEditForm({
                   disabled={!isDraft}
                 >
                   <option value="">{t("content.select")}</option>
-                  {FUNDING_INSTRUMENT_OPTIONS.map((option) => (
+                  {fundingOptions.map((option) => (
                     <option key={option.value} value={option.value}>
                       {option.label}
                     </option>
@@ -472,7 +474,7 @@ export default function OpportunityEditForm({
                   disabled={!isDraft}
                 >
                   <option value="">{t("content.selectFundingCategory")}</option>
-                  {FUNDING_CATEGORY_OPTIONS.map((option) => (
+                  {categoryOptions.map((option) => (
                     <option key={option.value} value={option.value}>
                       {option.label}
                     </option>
--- a/frontend/src/components/opportunity/OpportunityEditForm.tsx
+++ b/frontend/src/components/opportunity/OpportunityEditForm.tsx
@@ -9,6 +9,7 @@ import {
   eligbilityValueToGroup,
   ELIGIBILITY_OPTIONS,
   fundingOptions,
+  OPPORTUNITY_CATEGORY_OPTIONS,
 } from "src/constants/opportunity";
 import { OpportunityAttachment } from "src/types/opportunity/opportunityAttachmentTypes";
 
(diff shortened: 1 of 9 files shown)


glenblosser-nava (dismissed):

frontend/src/components/opportunity/OpportunityEditForm.tsx
```diff
@@ -4,6 +4,11 @@ import {
   saveOpportunityEditAction,
   type OpportunityEditValidationErrors,
 } from "src/app/[locale]/(base)/opportunity/[id]/edit/actions";
+import { eligibilityTypes } from "src/constants/opportunity";
+import {
+  categoryOptions,
+  fundingOptions,
```
doug-s-nava, line 11: I'd recommend moving these to a common file rather than relying a search specific file

frontend/src/components/opportunity/OpportunityEditForm.tsx
```diff
@@ -224,41 +227,14 @@ export default function OpportunityEditForm({
...
-        "other_native_american_tribal_organizations",
-        "nonprofits_non_higher_education_with_501c3",
-        "nonprofits_non_higher_education_without_501c3",
-      ].includes(option.value),
-    ),
-    misc: ELIGIBILITY_OPTIONS.filter((option) =>
-      ["individuals", "other", "unrestricted"].includes(option.value),
-    ),
-  };
+  const eligibilityGroups = eligibilityTypes.reduce(
```
doug-s-nava: can we reuse https://github.com/HHS/simpler-grants-gov/blob/e1166d03141ada7bb0e3f82c9956e7e5ac330e2f/frontend/src/components/opportunity/OpportunityEligibility.tsx#L7

doug-s-nava (commented):
* found two option types to clean up in RoleManager.tsx and SavedOpportunityOwnershipFilter.tsx
* if the team has decided not to use shared form input components in their work until global core components are build (likely towards the end of quad 6), I am ok with that. However, my impression was that the team was aligned on the approach of using the form elements created for use in the create form across their forms in the meantime, and using inconsistent and ad hoc implementations will make future refactors more difficult. In the end, though, this is the team's decision and I want to make sure there is consensus on the team about this approach

frontend/src/components/opportunity/OpportunityEditForm.tsx
doug-s-nava: recommend moving to constants/opportunity

desaiamit: My proposal: 

1.  Move `fundingOptions` and `categoryOptions` to a new neutral `constants/filterOptions.ts` and update the edit form import to point there. 

2. `searchFilterOptions.ts` would remain unchanged for now - updating it to import from the shared file touches the search filter pipeline and test coverage and deserves its own focused effort.

Would you be OK with me taking care of the Edit page and you may open a follow-up ticket to migrate the search side? OR you want to handle it differently

glenblosser-nava: If we move the options to a new constants file, we should remove them from searchFilterOptions as well. Anything that was importing from searchFilterOptions can be changed to import from the new file without any change of functionality or test coverage.

KtejanKella:
@doug-s-nava @glenblosser-nava are reviewers.

frontend/src/components/opportunity/OpportunityEditForm.tsx
desaiamit: Thank you both for the feedback.

**Revised Plan:**
- Move fundingOptions and categoryOptions definitions to `constants/opportunity.ts`
- Remove them from `searchFilterOptions.ts` - it will import them back for use in `allFilterOptions`
- Update `OpportunityEditForm.tsx` and `SearchDrawerFilters.tsx `to import from `constants/opportunity.ts`

No test changes needed. Will implement in this PR.

frontend/src/constants/searchFilterOptions.ts
```diff
@@ -1,6 +1,12 @@
-import { eligibilityTypes } from "src/constants/opportunity";
+import {
+  categoryOptions,
+  eligibilityTypes,
+  fundingOptions,
+} from "src/constants/opportunity";
 import { FilterOption } from "src/types/search/searchFilterTypes";
 
+export { categoryOptions, fundingOptions };
```
doug-s-nava: This works, but I'd rather update any files that use these options to import from the new file rather than passing the imports through like this. I feel like this pass through pattern can get confusing in terms of where things are actually defined.

frontend/src/components/opportunity/OpportunityEditForm.tsx
```diff
@@ -227,8 +227,9 @@ export default function OpportunityEditForm({
 
   const leftColumnItems = keyInformationItems.slice(0, 3);
   const rightColumnItems = keyInformationItems.slice(3);
-  const eligibilityGroups = eligibilityTypes.reduce(
-    (acc, { group, label, value }) => {
+  const eligibilityGroups = ELIGIBILITY_OPTIONS.reduce(
```
doug-s-nava, line 228: it looks like `ELIGIBILITY_OPTIONS` is defined i`opportunitEditFormConfig` but isn't actually used there. Would be best to define it where it is used

frontend/src/constants/searchFilterOptions.ts
desaiamit: Done - removed the re-export from `searchFilterOptions.ts` entirely. `allFilterOptions` now imports `categoryOptions` and `fundingOptions` directly from `constants/opportunity.ts`. Also caught and fixed the same pattern for `OPPORTUNITY_CATEGORY_OPTIONS` which was being re-exported through `opportunityEditFormConfig.ts` - that's been cleaned up as well

frontend/src/components/opportunity/OpportunityEditForm.tsx
desaiamit: Moved `ELIGIBILITY_OPTIONS` to `constants/opportunity.ts` alongside `eligibilityTypes` and `eligbilityValueToGroup` - all three are derived from the same source data.

doug-s-nava (approved):
✅ ✅ ✅ ✅

PR #9964, opened 2026-04-29
[Issue #9360] Switch from poetry to uv

### LINKED ISSUE #9360: Migrate from poetry to uv on the backend
### Summary

We want to look into migrating from poetry to uv for our dependency management on the backend.

This is likely not that complicated BUT is a very large change as we run all of our backend tasks as `poetry run ...` which uv also can do, but means we have to change the commands. See: https://github.com/search?q=repo%3AHHS%2Fsimpler-grants-gov%20poetry&type=code

Importantly, we need to make sure the system continues to work running command outside of docker as well as inside of docker - before you do any of this work you need to have your local setup capable of working in both ways.

A few high-level things we'll need to change:
* The pyproject.toml file / poetry.lock file will be replaced, https://github.com/mkniewallner/migrate-to-uv should be able to help automate the file format conversion
* Our makefile should replace all usages of poetry / commands defined in the pyproject with uv
* The [dockerfile](https://github.com/HHS/simpler-grants-gov/blob/main/api/Dockerfile) should use uv instead of poetry - https://docs.astral.sh/uv/guides/integration/docker/
* The [base docker image for openapi generation](https://github.com/HHS/simpler-grants-gov/blob/0f01404aa0b49412c849ecfa618d796768fcd0ff/api/docker-python-openapi#L29)also needs to be adjusted
* Our [terraform defined jobs](https://github.com/HHS/simpler-grants-gov/blob/main/infra/api/app-config/env-config/scheduled_jobs.tf) need to use uv
* [Workflow service command](https://github.com/HHS/simpler-grants-gov/blob/0f01404aa0b49412c849ecfa618d796768fcd0ff/infra/api/service/workflow.tf#L56)
* Documentation
* and more

We do not need to update analytics yet, that can be a follow-up.

If possible, see if there is a way we can migrate across several PRs, if so, break this into several tickets.

### Acceptance criteria

- [ ] Replace all references of poetry with uv in the API folder + documentation
- [ ] Test that all make commands that were using poetry now work with uv, both when running inside docker and outside.
- [ ] Verify that things like installing dependencies and packages is generally faster or the same as before
- [ ] Documentation updated with common commands (links to common command pages in uv docs is fine)
- [ ] All tests, e2e tests, and CI/CD still function
- [ ] Test that ECS tasks still run successfully in a lower environment with new code

chouinar:
## Summary
Fixes #9360 

## Changes proposed
* Replace poetry with uv
* Change both API and analytics to use uv
* Remove all references to poetry, change run commands that were previously `poetry run xyz` to just `xyz` as the $path is setup to allow that

## Context for reviewers
Importantly, this work should only affect how we get and install packages. The actual running of commands shouldn't meaningfully change (although I changed many from `poetry run x 123` to just `x 123` inside Docker as paths were setup to allow that). This is because the actual way of running python code and how it's organized is a python level concept that both Poetry and UV do in the same way (using venv's). This is largely just affecting how those venvs get populated.

In doing this work, I had to fix a bug with analytics linting/formatting not working. Not sure on exactly what was causing it, but when switching to UV, rules that have always existed in ruff started getting triggered that weren't before. The version of ruff did not change as part of this. I think it has to do with me moving where the `config.py` file was located (which had to be moved as the build for uv didn't work with it in the root). The changes to analytics should mostly be minor import adjustments and a few added comments. No practical changes.

## Validation steps
If you'd like to test any of this, simple check out the branch and run something like `make stop init` - init contains the build command and will rebuild your container.

If you'd like to try uv outside of the docker container, follow the instructions to setup uv: https://docs.astral.sh/uv/getting-started/installation/ - running `uv sync --all-groups --locked` (or just `make setup-local`) will install all the packages for you outside of docker. 

Documentation updates contain a bunch of commands that I ran, verified all the different ways of adding/upgrading packages work with links to docs.

Other things tested:
* Verified that commands that run python code inside or outside the docker container (based on the `PY_RUN_APPROACH` env var) work.
* Can run tests with `make test`
* Can run the format and lint commands
* DB migrations work
* Can run our ecs tasks like `make cmd args="workflow workflow-main"` or `make cmd args="task build-automatic-opportunities"`
* `make create-erds` - which uses a slightly different way of running
* `make console` - the interactive shell and has a weird command as well that still needs `uv run` to work properly.
* Verified common commands in analytics like tests and DB migrations work


infra/api/service/workflow.tf
```diff
@@ -53,7 +53,7 @@ resource "aws_ecs_task_definition" "workflow" {
       essential              = true,
       readonlyRootFilesystem = true,
 
-      command = ["poetry", "run", "flask", "workflow", "workflow-main"],
+      command = ["uv", "run", "flask", "workflow", "workflow-main"],
```
kkrug: The scheduled jobs in api/app-config/env-config/scheduled_jobs.tf use `["flask", ...]` directly without `uv run`. I know this previously used `poetry run`, but is that something we could do for consistency? I think if the release Docker image PATH includes `/api/.venv/bin`, we can call `flask` directly. If not, maybe the scheduled jobs should also have `uv run`. 

This is also the case in run-setup-foreign-tables.

bin/run-setup-foreign-tables
```diff
@@ -26,6 +26,6 @@ migrator_role_arn=$(terraform -chdir="infra/${app_name}/service" output -raw mig
 echo "Using migrator role: ${migrator_role_arn}"
 echo
 
-command='["poetry", "run", "flask", "data-migration", "setup-foreign-tables"]'
+command='["uv", "run", "flask", "data-migration", "setup-foreign-tables"]'
```
kkrug: Same thing as the comment on workflow.tf

api/tests/validate_legacy_soap_api/validate_soap_endpoints.py
```diff
@@ -29,7 +29,7 @@
 For cert and key use awk command to replace newlines
 `awk 'NF {sub(/\r/, ""); printf "%s\\n",$0;} bps_grantors.crt`
 `awk 'NF {sub(/\r/, ""); printf "%s\\n",$0;} bps_grantors.key`
-May need to run `poetry install` to update command
+May need to run `uv install` to update command
```
kkrug: Looks like `uv install` isn't a valid command
```suggestion
May need to run `uv sync` to update command
```

documentation/analytics/testing.md
```diff
@@ -12,14 +12,14 @@ Run `make unit-test` from the root of the `analytics/` sub-directory to run all
 
 ### Running specific set of tests
 
-To have greater control over which tests are run, you'll need to use `poetry run pytest` to access the pytest command line. The following commands show you how to invoke progressively smaller subsets of tests:
+To have greater control over which tests are run, you'll need to use `uv run pytest` to access the pytest command line. The following commands show you how to invoke progressively smaller subsets of tests:
 
-- `poetry run pytest` Runs all tests automatically collected in [test discovery](https://docs.pytest.org/en/stable/goodpractices.html#conventions-for-python-test-discovery)
-- `poetry run pytest tests/datasets/` Runs all collected tests in the
+- `yv run pytest` Runs all tests automatically collected in [test discovery](https://docs.pytest.org/en/stable/goodpractices.html#conventions-for-python-test-discovery)
+- `yv run pytest tests/datasets/` Runs all collected tests in the
```
kkrug:
```suggestion
- `uv run pytest` Runs all tests automatically collected in [test discovery](https://docs.pytest.org/en/stable/goodpractices.html#conventions-for-python-test-discovery)
- `uv run pytest tests/datasets/` Runs all collected tests in the
```

infra/frontend/app-config/env-config/scheduled_jobs.tf
```diff
@@ -1,7 +1,7 @@
 locals {
   # The `cron` here is the literal name of the scheduled job. It can be anything you want.
   # For example "file_upload_jobs" or "daily_report". Whatever makes sense for your use case.
-  # The `task_command` is what you want your scheduled job to run, for example: ["poetry", "run", "flask"].
+  # The `task_command` is what you want your scheduled job to run, for example: [flask", "example-cmd"].
```
kkrug:
```suggestion
  # The `task_command` is what you want your scheduled job to run, for example: ["flask", "example-cmd"].
```

infra/api/service/workflow.tf
chouinar: I had originally switched it all to `uv run` before I realized that wasn't needed, missed fixing these two after the fact.

documentation/api/package-dependency-management.md
```diff
@@ -34,92 +34,100 @@ Upgrading the base image involves going to a newer (or the `latest`) tagged base
...
-As I always lose [the command](https://python-poetry.org/docs/cli/#self-update), for your convenience:
+[The command](https://docs.astral.sh/uv/reference/cli/#uv-self), for your convenience:
 ```shell
-poetry self update
+uv self update
 ```
 
 ### For build
-Update the `ARG POETRY_VERSION` [line](https://github.com/HHS/simpler-grants-gov/blob/d48d969d68d41a04fa7b4f5052b93594c829feb8/api/docker-python-base#L8) in the Dockerfile for the [base image](#docker-base-image-updates) `api/docker-python-base`
+Update the `COPY --from=ghcr.io/astra-sh/uv:...` [line](https://github.com/HHS/simpler-grants-gov/blob/main/api/docker-python-base) in the Dockerfile for the [base image](#docker-base-image-updates) `api/docker-python-base`
```
kkrug:
```suggestion
Update the `COPY --from=ghcr.io/astral-sh/uv:...` [line](https://github.com/HHS/simpler-grants-gov/blob/main/api/docker-python-base) in the Dockerfile for the [base image](#docker-base-image-updates) `api/docker-python-base`
```

infra/api/app-config/env-config/scheduled_jobs.tf
```diff
@@ -9,8 +9,7 @@ locals {
   load-transform-args = {
     # Runs, but with everything disabled.
     dev = [
-      "poetry",
-      "run",
+
```
kkrug: We should probably clean up these 6 blank lines since the formatter won't catch stuff in the tf files

kkrug (dismissed):

hao10282025-sudo (commented):
I did a global search for "poetry", are these 3 documents still in use and need an update?

Image

kkrug (approved):

chouinar:
> I did a global search for "poetry", are these 3 documents still in use and need an update?
> 

These are from old ADRs, they're not docs we'd update. We might want something to replace them, but I'll follow up elsewhere to see if that's needed.

PR #9965, opened 2026-04-29
[Issue #9933] Update SOAP Cert Check & Handling

### LINKED ISSUE #9933: SOAP/Proxy: Add checks against LegacyCertificate and tcertificates table
### Summary

We need to handle this situation:
```
No certificate provided or certificate provided is expired
Fail without sending any request to the JWT SOAP or Simpler, no lookup in tcertificates, etc

Certificate sent, not found in tcertificates
Fail without sending any request to the JWT SOAP or Simpler, etc

Certificate sent, exists in legacy (tcertificates), not configured in Simpler (no legacy_certificate)
Make JWT call
Don't make Simpler calls, no error or failure due to no Simpler permissions being specified 

Certificate sent, exists in legacy (tcertificates), configured in Simpler
Make JWT call
Make Simpler call
```

Image

### Acceptance criteria

- [ ] hitting these scenarios yields the expected results

jakobpederson:
## Summary

Fixes / Work for #9933  

## Changes proposed

Added checks/handling for scenarios:
1. `GetOpportunityList` - no auth checking
2. No mtls cert on request -> short circuit error response
3. No tcertificate -> short circuit error response
4. tcertificate expired -> short circuit error response
5. valid tcertificate, no legacy certificate -> only call legacy
6. valid tcertificate, legacy certificate expired -> short circuit error response
7. valid tcertificate, valid legacy certificate -> call both simpler and legacy

Remove legacy `use-soap-cert` path, now only use JWT or certificate-less call to the original url.

Tests updated.

## Context for reviewers

We need to be able to facilitate users that are not configured in Simpler (they have tcertificates but no legacy certificate) as well as handling calls to `GetOpportunityList` (no auth). Throw error responses unless caller has a valid tcertificate and/or a valid legacy certificate. Remove the soap-cert path file writing.

## Validation steps

- [ ] Hit the endpoints, note that the Use-Soap-Cert header no longer works.
- [ ] Notice that unless tcert and legacycert are set up you'll get correct errors

### DIFF
--- a/api/src/legacy_soap_api/legacy_soap_api_auth.py
+++ b/api/src/legacy_soap_api/legacy_soap_api_auth.py
@@ -107,9 +107,13 @@ class SessionResumptionAdapter(HTTPAdapter):
 def get_tcertificate(
     serial_number: str, db_session: db.Session
 ) -> staging.certificates.Tcertificates | None:
+    # Since we lower the input serial_number we wrap the column
+    # value in func.lower() to make the match case insensitive. In hexadecimal
+    # there is no distinction between upper and lower case. The serial number
+    # can be either upper or lower cased in the DB.
     return db_session.scalars(
         select(staging.certificates.Tcertificates).where(
-            staging.certificates.Tcertificates.serial_num == serial_number
+            func.lower(staging.certificates.Tcertificates.serial_num) == serial_number.lower()
         )
     ).one_or_none()
 
@@ -124,13 +128,15 @@ def get_soap_auth(mtls_cert: str | None, db_session: db.Session) -> SOAPAuth:
     logger.info("soap_client_certificate: certificate received header")
     cert_str = unquote(mtls_cert)
     cert = load_pem_x509_certificate(cert_str.encode(), default_backend())
+    # We convert the integer representation of the serial number to lower case hexadecimal here
+    # Note: the 'x' in '032x' returns it lower cased
     serial_number_hex = format(int(cert.serial_number), "032x")
 
     tcertificate = get_tcertificate(serial_number_hex, db_session)
     if not tcertificate:
         logger.info(
             "soap_client_certificate: no tcertificate",
-            extra={"soap_api_event": LegacySoapApiEvent.NO_HEADER_CERT},
+            extra={"soap_api_event": LegacySoapApiEvent.TCERT_NOT_FOUND},
         )
         raise SOAPClientTcertificateNotFound("No tcertificate")
     if tcertificate.expirationdate and tcertificate.expirationdate <= get_now_us_eastern_date():
@@ -141,6 +147,7 @@ def get_soap_auth(mtls_cert: str | None, db_session: db.Session) -> SOAPAuth:
         raise SOAPClientCertificateIsExpired("tcertificate is expired")
 
     legacy_certificate = get_legacy_certificate_by_serial_number(db_session, serial_number_hex)
+
     if not legacy_certificate:
         logger.info(
             "soap_client_certificate: no LegacyCertificate",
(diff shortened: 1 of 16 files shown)


api/src/legacy_soap_api/legacy_soap_api_auth.py
```diff
@@ -91,29 +104,74 @@ def init_poolmanager(self, *args: Any, **kwargs: Any) -> None:
...
+            cert_id=tcertificate.currentcertid,
+        )
+        return SOAPAuth(certificate=soap_client_certificate)
+
+    if legacy_certificate and legacy_certificate.expiration_date <= get_now_us_eastern_date():
+        logger.info(
+            "soap_client_certificate: LegacyCertificate is expired",
+            extra={"soap_api_event": LegacySoapApiEvent.CERT_EXPIRED},
+        )
+        raise SOAPClientCertificateIsExpired("No tcertificate or tcertificate expired")
```
kkrug:
```suggestion
        raise SOAPClientCertificateIsExpired("LegacyCertificate is expired")
```

api/src/legacy_soap_api/legacy_soap_api_proxy.py
```diff
@@ -155,17 +104,22 @@ def _get_soap_response(
...
+            extra={"soap_api_event": LegacySoapApiEvent.JWT_CREATED},
+        )
+        return base64.b64encode(jwt_string.encode("utf-8")).decode("utf-8")
     logger.info(
-        "soap_client_certificate: created SOAP JWT",
-        extra={"soap_api_event": LegacySoapApiEvent.JWT_CREATED},
+        "soap_client_certificate: No cert_id",
     )
-    return base64.b64encode(jwt_string.encode("utf-8")).decode("utf-8")
+    return ""
```
kkrug: If `cert_id` is None, does this just fail silently?

api/src/legacy_soap_api/legacy_soap_api_auth.py
```diff
@@ -91,29 +104,74 @@ def init_poolmanager(self, *args: Any, **kwargs: Any) -> None:
...
+    cert = load_pem_x509_certificate(cert_str.encode(), default_backend())
+    serial_number_hex = format(int(cert.serial_number), "032x")
+
+    tcertificate = get_tcertificate(serial_number_hex, db_session)
+    if not tcertificate:
         logger.info(
-            "soap_client_certificate: successfully extracted certificate and serial number",
-            extra={"soap_api_event": LegacySoapApiEvent.PARSED_CERT},
+            "soap_client_certificate: no tcertificate",
+            extra={"soap_api_event": LegacySoapApiEvent.NO_HEADER_CERT},
```
kkrug: Should this have a more distinct message instead of reusing `NO_HEADER_CERT`? Maybe something like `TCERT_NOT_FOUND`.

api/src/legacy_soap_api/legacy_soap_api_auth.py
```diff
@@ -91,29 +104,74 @@ def init_poolmanager(self, *args: Any, **kwargs: Any) -> None:
...
+    serial_number: str, db_session: db.Session
+) -> staging.certificates.Tcertificates | None:
+    return db_session.scalars(
+        select(staging.certificates.Tcertificates).where(
+            staging.certificates.Tcertificates.serial_num == serial_number
+        )
+    ).one_or_none()
+
+
+def get_soap_auth(mtls_cert: str | None, db_session: db.Session) -> SOAPAuth:
```
kkrug, line 121: The old `get_soap_client_certificate()` that's still on line 191 had the call to `add_extra_data_to_current_request_logs()` for when there was a `legacy_certificate`. Is this something we want to continue doing so we don't lose this level of logging?

api/src/legacy_soap_api/legacy_soap_api_proxy.py
```diff
@@ -155,17 +104,22 @@ def _get_soap_response(
...
+        )
+        logger.info(
+            "soap_client_certificate: created SOAP JWT",
+            extra={"soap_api_event": LegacySoapApiEvent.JWT_CREATED},
+        )
+        return base64.b64encode(jwt_string.encode("utf-8")).decode("utf-8")
     logger.info(
-        "soap_client_certificate: created SOAP JWT",
-        extra={"soap_api_event": LegacySoapApiEvent.JWT_CREATED},
+        "soap_client_certificate: No cert_id",
```
kkrug, line 124: The other log statements all have an `extra` dict with `soap_api_event`. It may be worthwhile having one here for consistency.

api/src/legacy_soap_api/legacy_soap_api_auth.py
kkrug: Is this function being used anymore? If not, we should remove it along with any tests.

api/src/legacy_soap_api/legacy_soap_api_auth.py
```diff
@@ -91,29 +104,74 @@ def init_poolmanager(self, *args: Any, **kwargs: Any) -> None:
...
+            cert=cert_str,
+            fingerprint=cert.fingerprint(hashes.SHA256()).hex(),
+            issuer=cert.issuer.rfc4514_string(),
+            serial_number=str(serial_number_hex),
+            legacy_certificate=None,
+            cert_id=tcertificate.currentcertid,
+        )
+        return SOAPAuth(certificate=soap_client_certificate)
+
+    if legacy_certificate and legacy_certificate.expiration_date <= get_now_us_eastern_date():
```
kkrug: I think the `legacy_certificate and` part of this condition is unnecessary since we already check `if not legacy_certificate` above.

api/src/legacy_soap_api/legacy_soap_api_auth.py
```diff
@@ -91,29 +104,74 @@ def init_poolmanager(self, *args: Any, **kwargs: Any) -> None:
...
+    if not legacy_certificate:
+        logger.info(
+            "soap_client_certificate: no LegacyCertificate",
+            extra={"soap_api_event": LegacySoapApiEvent.NOT_CONFIGURED_CERT},
+        )
+        soap_client_certificate = SOAPClientCertificate(
+            cert=cert_str,
+            fingerprint=cert.fingerprint(hashes.SHA256()).hex(),
+            issuer=cert.issuer.rfc4514_string(),
+            serial_number=str(serial_number_hex),
```
kkrug: I think we can do without the `str()` casting here since we already use `format()` on `serial_number_hex` above. 
```suggestion
            serial_number=serial_number_hex,
```

api/src/legacy_soap_api/legacy_soap_api_proxy.py
jakobpederson: I was thinking on that and, while it should never be `None` by this point, I was thinking you just create an invalid JWT token that gets sent and rejected by the endpoint. Whether that's behavior we want or not is a good question. It would probably be better to raise an exceptiong iven our patterns, so I'll add that.

jakobpederson: Also there's the log so it's not completely silent.

api/src/legacy_soap_api/legacy_soap_api_auth.py
jakobpederson: good catch

kkrug (approved):

PR #9966, opened 2026-04-29
[Unticketed] Enable Grantor Opportunity endpoints in Training for testing

mdragon:
## Changes proposed

Enable API endpoints for Grantor Opportunity in Training


infra/api/app-config/training.tf
```diff
@@ -53,6 +53,7 @@ module "training_config" {
     # Email notification
     RESET_EMAILS_WITHOUT_SENDING               = "false"
     ENABLE_ORG_SAVED_OPPORTUNITY_NOTIFICATIONS = "true"
+    ENABLE_GRANTOR_OPPORTUNITY_ENDPOINTS       = 1
```
chouinar, line 56: Publish depends on the workflow service as the actual publish logic is in that, and we don't have workflows enabled in training. We'd have to do that as well.

prasnava (dismissed):
👍

infra/api/app-config/training.tf
mdragon: I think they're doing testing short of publish, but I'll raise that. Thanks.

chouinar: Which actually requires a few things to be setup including manually making a DB user. This is a fair bit bigger

sean-navapbc:
LGTM

sean-navapbc (dismissed):

mdragon:
Resolved merge conflict

prasnava (approved):
👍

PR #9973, opened 2026-04-30
[Unticketed] Enable workflow in training

chouinar:
## Changes proposed
Enables workflows in training + add docs

## Context for reviewers

I created the user in the training DB as part of this work.


mdragon (approved):
👍🏻

PR #9992, opened 2026-04-30
[Issue #9961] Update Seed File to Populate Agency_Id in Opportunity Table

### LINKED ISSUE #9961: Update Seed File to Populate Agency_Id in Opportunity Table
### Summary

The `agency_id` field for all of the entries in the Opportunity table is currently `null` for the data in the seed file(s). The file should be updated to populate this field with the ID of the corresponding agency the opportunity is assigned to.

### Acceptance criteria

- [ ] All seeded opportunities have an `agency_id` populated to their corresponding agency.

dghazvini:
## Summary

Fixes for #9961 

## Changes proposed

-In `seed_local_db.py`, updated the seed order so that agencies are seeded before opportunities. This is necessary to resolve the correct `agency_id` in the opportunity table.

-Added a `lookup_seed_agency_id_by_code` method to resolve `agency_id` and populate the opportunity table in the factory.

~~-Added `ondelete="SET NULL"` to the `agency_id` foreign key constraint in the Opportunity table model and added corresponding migration file. After doing the two above changes noticed some unit tests were failing - As noted from some tests in `test_relationships.py`, deleting an agency should not affect an opportunity. This change to the constraint will allow for an agency to still be deleted if an opportunity has an `agency_id` field referencing an agency.~~

-Updated affected unit tests to nullify the opportunity's agency_id prior to agency removal/deletion to avoid foreign key conflicts.

## Context for reviewers

The `agency_id` field for all of the entries in the Opportunity table is currently null for the data in the seed file(s). The file should be updated to populate this field with the ID of the corresponding agency the opportunity is assigned to.

~~I don't believe there are any data concerns with the seeding order change and the slight change to the foreign key constraint, but please review and leave any feedback if so.~~
EDIT - Removed foreign key constraint change and just updated unit tests.

## Validation steps

- [x] `agency_id` is populating in the opportunity table after re-seeding
- [x] Agency table seeding as expected
- [x] All unit tests are passing


api/src/db/models/opportunity_models.py
```diff
@@ -46,7 +46,7 @@ class Opportunity(ApiSchemaTable, TimestampMixin):
     opportunity_number: Mapped[str | None]
     opportunity_title: Mapped[str | None] = mapped_column(index=True)
     agency_id: Mapped[uuid.UUID | None] = mapped_column(
-        UUID, ForeignKey(Agency.agency_id), index=True
+        UUID, ForeignKey(Agency.agency_id, ondelete="SET NULL"), index=True
```
chouinar: Which tests were failing with this? I'd rather we fix tests than add this since this changes non-test behavior as well.

We eventually would want this column to never be nullable once we backfill this column.

---

If this is due to us doing `cascade_delete_from_db_table(db_session, Agency)` in tests, if we aren't able to easily remove that call, we could just do an `update opportunity set agency_id = null` at the same time to get the same effectively behavior, but localize it to just unit tests.

dghazvini: All of the tests here (also pasting full log. 1 Fail 4 Erroring out)
` tests/src/db/models/test_relationships.py::test_delete_agency `
`tests/src/services/opportunities_v1/test_opportunity_version.py::test_save_opportunity_version`
`tests/src/services/opportunities_v1/test_opportunity_version.py::test_save_opportunity_version_with_attachments`
`tests/src/services/opportunities_v1/test_opportunity_version.py::test_save_opportunity_version_draft`
`tests/src/services/opportunities_v1/test_opportunity_version.py::test_save_opportunity_version_with_prior_versions`

[log.txt](https://github.com/user-attachments/files/27376763/log.txt)

At least for the `test_delete_agency` test I believe it's just this call `db_session.delete(agency)`. If we don't want to alter the Foreign Key constraint I suppose we could nullify the `agency_id` here before the delete call. At first I was thinking of removing the opportunity which would slightly defeat the purpose of that test but I can take a look at the set null approach.

dghazvini:
I reverted the foreign key constraint change and the migration and just updated the corresponding tests to nullify agency IDs before removing agencies so they now pass.

chouinar (approved):

Issue #9134, opened 2026-03-19
Backend - Create an endpoint to list award recommendations


nathan-stricker:
### Summary

See [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Create-an-endpoint-to-list-award-recommendations)

This is needed to navigate back to an award recommendation.

#### Endpoint
`POST /alpha/award-recommendations/list`
#### Purpose
List award recommendations, filtering on passed in parameters
#### AuthZ
Requires `view_award_recommendation` privilege for each agency queried for
#### Request
```json
{
  "filters": {
    "agency_id": {
      "one_of": ["<uuid>"]
    }, # Required filter, we won't assume agency from the user to keep it simpler
  },
  "pagination": {
    "page_offset": 1,
    "page_size": 25,
    "sort_order": [
      {
        # Will allow for sorting by created_at only for the moment
        "order_by": "created_at",
        "sort_direction": "ascending"
      }
    ]
  },
}
```
#### Response
```json
{
  "data": [
    {
      # This object is the same as the response from GET award recommendation
    },
    {..}
  ],
  "pagination_info": {
    "page_offset": 1,
    "page_size": 25,
    "sort_order": [
      {
        "order_by": "created_at",
        "sort_direction": "ascending"
      }
    ],
    "total_pages": 2,
    "total_records": 42
  }
}
```
#### Logic
- Verify that any agencies passed in exist, 404 if not
- Verify that for any agencies passed in, the user has `view_award_recommendation` privilege in that agency
- Construct the query, adding any filters, and use our pagination util to get the results

### Acceptance criteria

- [ ] Endpoint created
- [ ] Tests created/updated as necessary

PR #10072, opened 2026-05-04
[Issue #9134] Create endpoint to list award recommendations and associated tests

nathan-stricker:
## Summary

Work for #9134 

## Changes proposed

Created endpoint to list award recommendations and added associated tests

## Context for reviewers

See [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Create-an-endpoint-to-list-award-recommendations)

## Validation steps

- [ ] Confirm implementation aligns with specifications outlined in the [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Award-Recommendation.1)
- [ ] Review created tests


api/src/api/award_recommendations_alpha/award_recommendation_schemas.py
```diff
@@ -579,6 +583,50 @@ class AwardRecommendationRiskListResponseSchema(AbstractResponseSchema, Paginati
...
+class AwardRecommendationListRequestSchema(Schema):
+    """Schema for POST /alpha/award-recommendations/list request"""
+
+    filters = fields.Nested(AwardRecommendationListFilterSchema(), required=True)
+
+    pagination = fields.Nested(
+        generate_pagination_schema(
+            "AwardRecommendationListPaginationSchema",
+            ["created_at"],
+            default_sort_order=[{"order_by": "created_at", "sort_direction": "ascending"}],
```
aaronbrown-nava: Looks like we normally go with descending sort on created_at in other calls (newest first). Any reason to go ascending on this one?

chouinar: Generally do descending as we want to show recent records at the top. Would say we should keep that pattern.

api/src/services/award_recommendations/list_award_recommendations.py
```diff
@@ -0,0 +1,117 @@
...
+    db_session: db.Session, user: User, request: dict
+) -> tuple[Sequence[AwardRecommendation], PaginationInfo]:
+    params = AwardRecommendationListRequest.model_validate(request)
+
+    agency_ids = set(params.filters.agency_id.one_of or [])
+    if not agency_ids:
+        raise_flask_error(
+            422,
+            message="filters.agency_id.one_of must contain at least one agency_id",
+        )
```
chouinar: For consistency in formatting a length error, we should put this into the schema itself which can validate a min length.

The UUID schema builder doesn't have that, but you could copy how we do it from the [StrSearchSchemaBuilder](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/api/schemas/search_schema.py#L92) to have a list of validators

api/src/services/award_recommendations/list_award_recommendations.py
```diff
@@ -0,0 +1,117 @@
...
+
+    agencies = verify_agencies_access(db_session, user, agency_ids)
+    agency_codes = [a.agency_code for a in agencies]
+
+    stmt = (
+        select(AwardRecommendation)
+        .join(Opportunity, Opportunity.opportunity_id == AwardRecommendation.opportunity_id)
+        .where(
+            AwardRecommendation.is_deleted.isnot(True),
+            Opportunity.agency_code.in_(agency_codes),
```
chouinar, line 85: A note that eventually we should be able to use the agency ID itself, but probably not until the switch to the grantor system

api/src/services/award_recommendations/list_award_recommendations.py
```diff
@@ -0,0 +1,117 @@
...
+        )
+        .options(
+            selectinload(AwardRecommendation.opportunity).selectinload(Opportunity.agency_record),
+            selectinload(AwardRecommendation.opportunity)
+            .selectinload(Opportunity.current_opportunity_summary)
+            .selectinload(CurrentOpportunitySummary.opportunity_summary),
+            selectinload(AwardRecommendation.award_recommendation_attachments).selectinload(
+                AwardRecommendationAttachment.uploading_user
+            ),
```
chouinar: We might want to adjust the response model to not have attachments returned / not reuse the general award recommendation model.

If we return 25 award recommendations, that each have 5 attachments, that 125 attachments. While we don't return the files themselves, we do presign the URLs which takes like 10-20ms each. Not a big deal for a handful on a single award rec, but might make this request really, really slow.

---

Don't need to fully redefine an award rec, just make a new base model without attachments, and then add a new model with attachments. We do this pattern already with the opportunity schema https://github.com/HHS/simpler-grants-gov/blob/main/api/src/api/opportunities_v1/opportunity_schemas.py#L329 as search doesn't need attachments, but other places do.

api/src/api/award_recommendations_alpha/award_recommendation_schemas.py
nathan-stricker: In that case, should I go ahead and update the request schema for listing risks (`AwardRecommendationSubmissionListRequestSchema`) to be descending as well?

chouinar (approved):

PR #10099, opened 2026-05-04
[Issue 9636] Transition form template definitions to new directory structure

### LINKED ISSUE #9636: transition form template definitions to new directory structure
### Summary

In this quad we're going to be both adding new form templates and adding in more robust versioning of the form templates as laid out in the [form template versioning tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2862907394/Form+Versioning+and+Auditability+Tech+Spec).  To allow these initiatives to go on in parallel, we're going to

1. Move the form templates first and update the two command line tools that update the database as it exists now.
2. Then we'll build out the registry solve while also adding new templates.  
3. At the end of the versioning phase 1 work we'll be deleting the two command line tool.  

It means we do some work at the beginning of the quad that will be deleted by the end of the quad, but the freedom in timing it buys us is worth it. 

### Acceptance criteria

- [ ] Git move (to preserve git version history of the files!) the existing form template definition files into the new directory structure, effectively making them all version 1.0 for that form template.
- [ ] Update the list-forms task to compare the database against the new directory structure.  Assume all form templates will be 1.0.  
- [ ] Update the update-form task to read the directory structure and update the database tables.  Assume all for templates will be 1.0. 
- [ ] Unittests.
- [ ] Confirm that functionality is equivalent to what it is today.

mikehgrantsgov:
## Summary

Fixes #9636 

## Changes proposed

Relocates all 16 form template definition files from a flat `forms/*.py` layout into a versioned directory structure:

`forms/<form_name>/<major>/<minor>/form_json.py`

All existing forms are placed at 1/0 (their first SGG version). Git history is preserved via `git mv`.

Each form directory also gets:

`__init__.py` loads `1/0/form_json.py` via `importlib` and re-exports all public symbols, so all existing import paths continue to work unchanged
`config.py` holds `FORM_ID` and `SHORT_FORM_NAME`, the identifiers that are stable across versions

## Context for reviewers

No changes made to forms/__init__.py, either CLI task, or any test/seed file.

Notes on the mypy linting errors and remediations:
- mypy validates that any directory containing `__init__.py` has a valid Python identifier as its name. 1 and 0 aren't valid identifiers, so mypy hard-errored. 

- Without `__init__.py` in the numeric dirs, mypy couldn't reconstruct the full module path for the `form_json.py` files and treated them all as a top-level module named form_json, causing a duplicate-module collision. Fixed that.

## Validation steps

Confirm all existing XML tests pass.


api/src/form_schema/forms/attachment_form/__init__.py
```diff
@@ -0,0 +1,17 @@
+import importlib.util
+from pathlib import Path
+
+_path = Path(__file__).parent / "1" / "0" / "form_json.py"
+_spec = importlib.util.spec_from_file_location(__name__ + "._v1_0", _path)
+if _spec is None or _spec.loader is None:
+    raise RuntimeError(f"Could not load form module from {_path}")
+_module = importlib.util.module_from_spec(_spec)
+_spec.loader.exec_module(_module)
```
freddieyebra: could me modularize this logic? So it can read something like following:

```python
def load_versioned_form(path, version):
    ...

_mod = load_versioned_form(Path(__file__).parent, "1.0")
FORM_JSON_SCHEMA = _mod.FORM_JSON_SCHEMA
...
```

api/src/form_schema/forms/epa_key_contacts/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+_spec = importlib.util.spec_from_file_location(__name__ + "._v1_0", _path)
+if _spec is None or _spec.loader is None:
+    raise RuntimeError(f"Could not load form module from {_path}")
+_module = importlib.util.module_from_spec(_spec)
+_spec.loader.exec_module(_module)
+
+FORM_JSON_SCHEMA = _module.FORM_JSON_SCHEMA
+FORM_UI_SCHEMA = _module.FORM_UI_SCHEMA
+FORM_XML_TRANSFORM_RULES = _module.FORM_XML_TRANSFORM_RULES
+EPA_KEY_CONTACT_v2_0 = _module.EPA_KEY_CONTACT_v2_0
```
freddieyebra: I see that we currently do not have a `FORM_RULE_SCHEMA` Do we need to have a `FORM_RULE_SCHEMA` for this form?

api/pyproject.toml
```diff
@@ -220,6 +220,13 @@ warn_unused_ignores = true
 
 plugins = ["pydantic.mypy"]
 
+# form_json.py files live inside numeric directories (e.g. sf424/1/0/form_json.py)
+# that are not valid Python package names, so mypy cannot resolve their module paths.
+# They are loaded at runtime via importlib in each form package's __init__.py.
+exclude = [
+    "src/form_schema/forms/[^/]+/[0-9]+/[0-9]+/form_json\\.py",
+]
```
freddieyebra, line 240: nice.

api/src/form_schema/forms/supplementary_neh_cover_sheet/config.py
```diff
@@ -0,0 +1,5 @@
+import uuid
+
+# Stable identifiers for this form — do not change across versions.
+FORM_ID = uuid.UUID("b0541e24-3fd5-444e-99ac-ba4a764582ed")
+SHORT_FORM_NAME = "SupplementaryCoverSheetforNEHGrantPrograms"
```
freddieyebra, line 5: can we add some unit tests that ensure that any files outside of this pattern will fail if the structure deviates?

freddieyebra (changes requested):

api/src/form_schema/forms/attachment_form/__init__.py
mikehgrantsgov: Good call. I created `api/src/form_schema/forms/_loader.py` with a single `load_versioned_form(path, version)` function
Updated all 16 form __init__.py files to use it

api/src/form_schema/forms/epa_key_contacts/__init__.py
mikehgrantsgov: It's the only form without `FORM_RULE_SCHEMA` because it has no conditional fields (just name/address/phone contacts). Added a `# No FORM_RULE_SCHEMA` comment to make this explicit.

api/src/form_schema/forms/supplementary_neh_cover_sheet/config.py
mikehgrantsgov: Good call. Created `api/tests/src/form_schema/forms/test_form_structure.py` with four parametrized tests that run against every form directory

freddieyebra (dismissed):
looks good

freddieyebra (approved):

Issue #9774, opened 2026-04-23
Modify our LoadOracleData logic to filter queries by a set timestamp


chouinar:
### Summary

We've had an annoying issue for a long time where our transforms will fail for a given run because of how we do the loading. Let's assume we load the `topportunity` and `tsynopsis` tables in that order. We get the following potential scenario:

* 01:00:00 - we load + copy records from `topportunity` to our staging table
* 01:00:10 - a new opportunity is created in grants.gov - this adds a row to both `topportunity` and `tsynopsis`
* 01:00:20 - We load + copy records from `tsynopsis`
* 01:01:00 - We transform `topportunity` (new opportunity wasn't copied, nothing happens)
* 01:01:05 - We transform `tsynopsis` (new synopsis can't be transformed as the opportunity doesn't exist)
* ...
* 02:00:00 - we load + copy records from `topportunity` to our staging table - this gets that record from the prior hour

What we propose doing is instead have that query to get records should have a filter that only grabs things created/updated at 12:00:00 or earlier. Effectively ignoring anything that happened after we started loading data. In the above scenario, we'd not copy the `tsynopsis` until the next run avoiding the problem. Effectively, make the batch all play by the same rules.

We actually have 3 types of queries, inserts, updates, and deletes.
* Inserts would just need to check against the `created_date`
* Updates would check against the `last_upd_date` (they're already using that to see if there is an update compared to our system)
* Deletes can't compare against a date since the row in the grants.gov system was deleted, we won't modify delete logic.

The timestamp in grants.gov's data is in EST, our infrastructure runs in UTC, use our datetime util to get an EST datetime. A naive comparison will be off by 4 to 5 hours, causing the filter to skip or include the wrong records. 

Related [thread](https://betagrantsgov.slack.com/archives/C09FCPWPSEM/p1776896010290099) 

### Acceptance criteria

* [x] A single batch cutoff timestamp is captured at load task start and applied to insert/update queries across all tables (deletes unchanged)
* [x] Cutoff timestamp is logged at task start for traceability
* [x] unit tests
  * [x] verify the cutoff timestamp passed to queries is in EST (not UTC or naive local time). Fail if the datetime is off by any offset 
  * [x] record with created_date after cutoff is excluded from the batch; record at or before cutoff is included
  * [x] a record deferred on run N is picked up on run N+1
* [x] Staging validation. This must be tested in staging for at least a few days before it goes to prod
[...]

PR #10125, opened 2026-05-05
[Issue #9774] Modify our LoadOracleData logic to filter queries by a set timestamp

kkrug:
## Summary

Fixes / Work for #9774  

## Changes proposed

- Updated sql.py to add optional cutoff parameter to `build_select_new_rows_sql` (filters on created_date) and `build_select_updated_rows_sql` (filters on last_upd_date)
- Updated load_oracle_data_task.py to capture an EST batch cutoff timestamp at task start, log it, and pass it to the insert/update SQL builders
- Updated test_sql.py to add unit tests verifying cutoff clause appears in generated SQL for both insert and update queries
- Updated test_load_oracle_data_task.py to add integration tests: cutoff is timezone-aware Eastern, records after cutoff are excluded, deferred records are picked up on next run, and NULL created_date records are still included. Also updated some existing tests to use `scalar` instead of `query`

## Context for reviewers

Tables are loaded sequentially from Oracle, so a record created between loading table A and table B can result in a child row being copied without its parent, causing transform failures. This was the root cause of repeated "Opportunity instruction cannot be processed as the competition for it does not exist" errors. The fix captures a single cutoff timestamp (in EST to match grants.gov data) at the start of the load and filters INSERT/UPDATE queries to only include records at or before that time. Deletes are unchanged since the source row no longer exists to compare against.

## Validation steps
- Confirm new and existing test pass
- Staging validation: monitor for zero occurrences of "Opportunity instruction cannot be processed as the competition for it does not exist" in New Relic over several days before promoting to prod


api/src/data_migration/load/load_oracle_data_task.py
```diff
@@ -72,6 +72,11 @@ def __init__(
 
     def run_task(self) -> None:
         """Main task process, called by run()."""
+        self.batch_cutoff = datetime_util.get_now_us_eastern_datetime()
+        logger.info(
+            "batch cutoff timestamp captured",
+            extra={"batch_cutoff": self.batch_cutoff.isoformat()},
```
chouinar: FYI - our logging automatically converts dates/datetimes for you to isoformat, safe to just do 
```suggestion
            extra={"batch_cutoff": self.batch_cutoff},
```

api/src/data_migration/load/load_oracle_data_task.py
```diff
@@ -72,6 +72,11 @@ def __init__(
 
     def run_task(self) -> None:
         """Main task process, called by run()."""
+        self.batch_cutoff = datetime_util.get_now_us_eastern_datetime()
```
chouinar: Generally good practice to set values like this in the init function - makes it so it's not seen as nullable by any type checkers, and you can even pass in a datetime if you really wanted that way, although freeze time works just fine for tests.

api/src/data_migration/load/sql.py
```diff
@@ -81,9 +91,13 @@ def build_select_updated_rows_sql(
             )
             < source_table.c.last_upd_date
         )
-        .order_by(*source_table.primary_key.columns)
     )
 
+    if cutoff is not None:
+        query = query.where(source_table.c.last_upd_date <= cutoff)
+
+    return query.order_by(*source_table.primary_key.columns)
```
chouinar, line 99: Might not want to make the change in this particular PR, but I don't actually know why we sort here - we don't do anything with that data being sorted.

api/src/data_migration/load/sql.py
```diff
@@ -3,28 +3,37 @@
...
+            # `(SELECT (id1, id2, id3, ...) FROM <destination_table>)`    (subquery)
+            sqlalchemy.select(*destination_table.primary_key.columns)
         )
-        .order_by(*source_table.primary_key.columns)
     )
 
+    if cutoff is not None:
+        query = query.where(
+            sqlalchemy.or_(
+                source_table.c.created_date.is_(None),
```
chouinar, line 30: Good catch on this, it's incredibly rare, but some data might have this as null.

api/src/data_migration/load/load_oracle_data_task.py
kkrug: Updated

api/src/data_migration/load/load_oracle_data_task.py
kkrug: Updated

api/src/data_migration/load/sql.py
kkrug: I'll leave it as is for now.

chouinar:
Change looks good, but want us to monitor this one for a few days in a lower environment - mainly to make sure that the timestamp comparison works right / the timezone is EST.

kkrug:
@chouinar Prod release is live so we should be good to go here after final approval

chouinar (approved):

PR #10138, opened 2026-05-05
[Issue #9399] Updating URL Namespace for Grantor Paths

### LINKED ISSUE #9399: Update Page Url Namespace
### Summary

Update namespace the page urls under [simpler.grants.gov/grantors/](http://simpler.grants.gov/grantors/), e.g., [simpler.grants.gov/grantors/opportunities](http://simpler.grants.gov/grantors/opportunities)

### Acceptance criteria

@glenblosser- Update the acceptance criteria

glenblosser-nava:
## Summary

Fixes #9399 

## Changes proposed

Moves the following paths to /grantor:

- /opportunity/{opportunity_id}/edit
- /opportunities
- /opportunities/create

## Context for reviewers

As part of our data separation migration, all grantor pages need to exist under a /grantor namespace

## Validation steps

Verify that none of the above listed urls can be accessed without the /grantor/ namespace
Navigate through the create opportunity workflow and ensure everything works as before


frontend/src/app/[locale]/(base)/opportunity/[id]/page.tsx
```diff
@@ -109,7 +109,7 @@ async function OpportunityListing({ params }: OpportunityListingProps) {
   }
   if (id !== opportunityData.opportunity_id) {
     redirect(
-      `/opportunity/${opportunityData.opportunity_id}`,
+      `/grantor/opportunity/${opportunityData.opportunity_id}`,
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/opportunity/[id]/page.tsx]: this doesn't seem like it should be changed

frontend/tests/e2e/login/specs/login-redirect.spec.ts
```diff
@@ -98,7 +98,7 @@ test.describe("Login Page Redirect", () => {
     { tag: [AUTH, FULL_REGRESSION] },
     async ({ page }) => {
       await page.evaluate(() => {
-        sessionStorage.setItem("login-redirect", "/opportunities");
+        sessionStorage.setItem("login-redirect", "/grantor/opportunities");
```
doug-s-nava, line 100: these tests shouldn't change

doug-s-nava (commented):
can we set up redirects from the existing urls to the new ones in the next config? see https://github.com/HHS/simpler-grants-gov/blob/de5cd5a1aa7483bee642e57543227b2512468f16/frontend/next.config.js#L196

frontend/src/app/[locale]/(base)/opportunity/[id]/page.tsx
Comment (glenblosser-nava) [reply frontend/src/app/[locale]/(base)/opportunity/[id]/page.tsx]: It should not have, good catch. Fixed

frontend/tests/e2e/login/specs/login-redirect.spec.ts
glenblosser-nava: Why not?

glenblosser-nava:
> can we set up redirects from the existing urls to the new ones in the next config? see
> 
> https://github.com/HHS/simpler-grants-gov/blob/de5cd5a1aa7483bee642e57543227b2512468f16/frontend/next.config.js#L196

Added redirects

frontend/tests/e2e/login/specs/login-redirect.spec.ts
doug-s-nava: so, strange that they were set up this way to begin with - it looks like I wrote the test to use `/opportunities` before that route existed, but the tests pass because they're just making sure the url matches the value that's provided in the spoofed session storage, and don't care if the page actually exists. As such, the test would continue to work with `/opportunities`, but it makes more sense to update the url to one that actually exists, so let's go with it

doug-s-nava (dismissed):
✅ ✅ ✅ ✅

doug-s-nava (approved):

PR #10146, opened 2026-05-05
[Issue #9702] Setup local version of DynamoDB for development

### LINKED ISSUE #9702: Setup a local version of DynamoDB for development
### Summary

We want to have a local DynamoDB instance running so we can use it for development

It should be possible to do this via Docker as Amazon provides a local DynamoDB instance you can run in Docker https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocal.DownloadingAndRunning.html

Additionally, as part of any make commands to setup this, we should create a table similar to how some of our other make commands setup after initializing. We should create a table called `local-virus-scan` and it should have the following attributes (setup any that have to be setup, might not need to fully do this):
* `attachment_id` - the partition key
* `user_id` - a string
* `status` -  a string

### Acceptance criteria

- [ ] Local DynamoDB setup - should be able to interact with it from command line
- [ ] Table automatically created for local dev

hao10282025-sudo:
## Summary

Fixes #9702  

## Changes proposed

We want to have a local DynamoDB instance running so we can use it for development

## Context for reviewers

It should be possible to do this via Docker as Amazon provides a local DynamoDB instance you can run in Docker https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocal.DownloadingAndRunning.html

Additionally, as part of any make commands to setup this, we should create a table similar to how some of our other make commands setup after initializing. We should create a table called local-virus-scan and it should have the following attributes (setup any that have to be setup, might not need to fully do this):

- attachment_id - the partition key
- user_id - a string
- status - a string

## Validation steps

- Local DynamoDB setup - should be able to interact with it from command line
- Table automatically created for local dev


api/src/db/setup_local_dynamodb.py
```diff
@@ -0,0 +1,55 @@
...
+    )
+
+    try:
+        table = dynamodb.create_table(
+            TableName=LOCAL_VIRUS_SCAN_TABLE_NAME,
+            KeySchema=[
+                {"AttributeName": "attachment_id", "KeyType": "HASH"},  # Partition key
+            ],
+            AttributeDefinitions=[
+                {"AttributeName": "attachment_id", "AttributeType": "S"},  # String type
```
hao10282025-sudo, line 46: for DynamoDB, the columns definition is in sync between KeySchema and AttributeDefinitions, only the index column can be created at this step
**_user_id_** and **_status_** columns can be manipulated at data insertion (tested in a separate function which not necessary to check-in)

image
<br>
image

api/src/db/setup_local_dynamodb.py
```diff
@@ -0,0 +1,53 @@
...
+            ],
+            BillingMode="PAY_PER_REQUEST",  # On-demand billing for local development
+        )
+        table.wait_until_exists()
+        logger.info("Successfully created table '%s'", LOCAL_VIRUS_SCAN_TABLE_NAME)
+    except dynamodb.meta.client.exceptions.ResourceInUseException:
+        logger.info("Table '%s' already exists", LOCAL_VIRUS_SCAN_TABLE_NAME)
+    except Exception as e:
+        logger.error("Error creating table '%s': %s", LOCAL_VIRUS_SCAN_TABLE_NAME, {e})
```
kkrug: Using `logger.exception()` will preserve the stack trace. 

Also, using `{}` here will create a set literal instead of just the string representation
```suggestion
        logger.exception("Error creating table '%s': %s", LOCAL_VIRUS_SCAN_TABLE_NAME, e)
```

api/Makefile
```diff
@@ -224,6 +224,21 @@ start-sqsmock:
...
 
+##################################################
+# DynamoDB Local
+##################################################
+
+init-dynamodb: start-dynamodb setup-dynamodb
+
+start-dynamodb:
+	docker compose up --detach dynamodb-local
+	chmod 755 ./bin/wait-for-local-dynamodb.sh
```
chouinar: This shouldn't need to be in the makefile, if you chmod a file, it'll keep those permissions when it goes to git.

api/docker-compose.yml
```diff
@@ -112,8 +120,15 @@ services:
...
       - sqsmock
+      - dynamodb-local
       - mock-oauth2-server
       - mock-applicants-soap-api
+    links:
+      - dynamodb-local
+    environment:
+      AWS_ACCESS_KEY_ID: 'local'
+      AWS_SECRET_ACCESS_KEY: 'local'
+      REGION: 'us-east-1'
```
kkrug: If you take a look at [local.env](https://github.com/HHS/simpler-grants-gov/blob/2ed35de4ad7f4c7c96ba4e5a4016feffaa1e30e5/api/local.env#L105), we manage the region in `AwsConfig`, not env vars. For other AWS services, we have endpoint URL vars in local.env (like `AWS_S3_ENDPOINT_URL`), so we could add something like `AWS_DYNAMODB_ENDPOINT_URL=http://dynamodb-local:8000`. Then we wouldn't need this `environment` section at all.

api/docker-compose.yml
```diff
@@ -112,8 +120,15 @@ services:
       - opensearch-node
       - s3mock
       - sqsmock
+      - dynamodb-local
       - mock-oauth2-server
       - mock-applicants-soap-api
+    links:
+      - dynamodb-local
```
kkrug: `links` is a legacy feature of Docker compose. I think we can remove this section altogether and not have to make any other changes.

api/src/db/setup_local_dynamodb.py
```diff
@@ -0,0 +1,53 @@
...
+    """Set up local DynamoDB tables for development"""
+    with src.logging.init(__package__):
+        error_if_not_local()
+
+        dynamodb = boto3.resource(
+            "dynamodb",
+            endpoint_url="http://dynamodb-local:8000",
+            region_name="us-east-1",
+            aws_access_key_id="local",
+            aws_secret_access_key="local",
```
chouinar: Endpoint URL should be an env var you put in local.env. For development that happens outside of Docker (like I do), I need to set this differently since the docker container name is only usable inside docker.

api/docker-compose.yml
chouinar: You added these, but they're not used anywhere, is this a leftover from an old approach? We would want any env vars defined in the local.env file if they were needed

api/docker-compose.yml
chouinar: By default, docker containers should be able to reach other containers by just their container name if they're on the default network - was there an issue this fixed?

api/src/db/setup_local_dynamodb.py
```diff
@@ -0,0 +1,53 @@
...
+        )
+
+        _create_virus_scan_table(dynamodb)
+
+
+def _create_virus_scan_table(dynamodb: boto3.resource) -> None:
```
kkrug: `boto3.resource` is a callable function, not a type. I don't think it would cause an error, but we should change it anyway.
```suggestion
def _create_virus_scan_table(dynamodb: boto3.resources.base.ServiceResource) -> None:
```

api/src/db/setup_local_dynamodb.py
```diff
@@ -0,0 +1,53 @@
...
+
+import boto3
+
+import src.logging
+from src.util.local import error_if_not_local
+
+logger = logging.getLogger(__name__)
+
+# DynamoDB table configuration
+LOCAL_VIRUS_SCAN_TABLE_NAME = "local-virus-scan"
```
chouinar: Should make this an env var as well since later logic we'll implement will need us to be able to configure that anyways.

Looks like infra already created DynamoDB in the envs and pass that on the `FILE_SCAN_CACHE_TABLE_NAME` env var, so name it that

api/src/db/setup_local_dynamodb.py
chouinar: We have `boto3.resource` in a few places in the code - the way types are defined in boto (being created by a central factory function) means it's basically untyped as far as any type-checker is concerned.

You are probably, but we'd never know.

chouinar: Actually - if we wanted - https://pypi.org/project/boto3-stubs/ seems they do have type stubs. Don't do that here, it'll probably complain about a lot.

api/Makefile
hao10282025-sudo: Updated
Good to know!

api/src/db/setup_local_dynamodb.py
hao10282025-sudo: Updated

api/docker-compose.yml
hao10282025-sudo: hmm I set this up as the AWS documents, but looks like not affect anything (verified after removing these lines)
I deleted these lines at this moment

api/docker-compose.yml
hao10282025-sudo: Deleted this links setup.
Same as above, just setup as the AWS document.

api/src/db/setup_local_dynamodb.py
hao10282025-sudo: Updated

api/src/db/setup_local_dynamodb.py
hao10282025-sudo: updated

api/docker-compose.yml
hao10282025-sudo: updated

api/docker-compose.yml
hao10282025-sudo: updated

api/src/db/setup_local_dynamodb.py
hao10282025-sudo: yeah I also searched online for information about all these three definitions, looks like either boto3.resource or boto3.resources.base.ServiceResource is not the best for the type hinting.
If we are not setting up boto3-stubs at this moment, I am updating it with "boto3.resources.base.ServiceResource", as from online info it is more reasonable.
I guess it is alright as long as it is not throwing error, we can discuss if any other concern.

api/local.env
```diff
@@ -86,6 +86,13 @@ DB_SSL_MODE=allow
 # could contain sensitive information.
 HIDE_SQL_PARAMETER_LOGS=TRUE
 
+############################
+# DynamoDB Mock
+############################
+
+# If you want to connect to dynamodb-local outside of docker, use localhost instead of dynamodb-local
+DYNAMODB_ENDPOINT_URL=http://dynamodb-local:8000
```
kkrug: Can we prefix this with `AWS_` to be consistent with other AWS endpoints?
```suggestion
AWS_DYNAMODB_ENDPOINT_URL=http://dynamodb-local:8000
```

api/src/db/setup_local_dynamodb.py
```diff
@@ -0,0 +1,56 @@
...
+
+import boto3
+
+import src.logging
+from src.util.local import error_if_not_local
+
+logger = logging.getLogger(__name__)
+
+# DynamoDB table configuration
+file_scan_cache_table_name = os.getenv("FILE_SCAN_CACHE_TABLE_NAME", "local-virus-scan")
```
kkrug: The `FILE_SCAN_CACHE_TABLE_NAME` env var is not in local.env

api/src/db/setup_local_dynamodb.py
```diff
@@ -0,0 +1,56 @@
...
+        )
+
+        _create_virus_scan_table(dynamodb)
+
+
+def _create_virus_scan_table(dynamodb: boto3.resources.base.ServiceResource) -> None:
+    """Create the local-virus-scan table if it doesn't already exist"""
+    logger.info(
+        "Creating DynamoDB table '%s' if it does not already exist", file_scan_cache_table_name
+    )
```
kkrug, line 37: We generally try to make log messages static strings and use the `extra` dict to house variable data. There are a few instances of this in this file.
```suggestion
    logger.info(
        "Creating DynamoDB table if it does not already exist",
        extra={"table_name": file_scan_cache_table_name},
    )
```

api/src/db/setup_local_dynamodb.py
```diff
@@ -0,0 +1,56 @@
...
+# DynamoDB table configuration
+file_scan_cache_table_name = os.getenv("FILE_SCAN_CACHE_TABLE_NAME", "local-virus-scan")
+
+
+def setup_local_dynamodb() -> None:
+    """Set up local DynamoDB tables for development"""
+    with src.logging.init(__package__):
+        error_if_not_local()
+
+        dynamodb_endpoint_url = os.getenv("DYNAMODB_ENDPOINT_URL", "http://dynamodb-local:8000")
```
chouinar: Can we put this and the above env var in the src/adapters/aws folder as in a `dynamodb_adapter` file using Pydantic? We'll add more to it later, but it'll be something we need to fetch there.

api/src/db/setup_local_dynamodb.py
chouinar: If we make this a pydantic model, I'd say we shouldn't have a default in the code, it should be a required env var to be set - we'd have to add it to local.env

api/src/db/setup_local_dynamodb.py
kkrug: We generally use Pydantic config/settings classes for all environment variables instead of `os.getenv`. If you look at the [S3 adapter for example](https://github.com/HHS/simpler-grants-gov/blob/66e0581fd472ecb1119f51c81dd0cc62d946446a/api/src/adapters/aws/s3_adapter.py#L10), you can see we do `aws_sqs_endpoint_url: str | None = Field(alias="AWS_SQS_ENDPOINT_URL", default=None)`. So for this env and `FILE_SCAN_CACHE_TABLE_NAME` above, we could do:
```
class DynamoDBConfig(PydanticBaseEnvConfig):
    aws_dynamodb_endpoint_url: str | None = Field(alias="AWS_DYNAMODB_ENDPOINT_URL", default=None)
    file_scan_cache_table_name: str = Field(alias="FILE_SCAN_CACHE_TABLE_NAME", default="local-virus-scan")
```
I think it would make the most sense to do this in a new file at `src/adapters/aws/dynamodb_adapter.py` and then import `from src.adapters.aws.dynamodb_config import DynamoDBConfig`

hao10282025-sudo: Updated

api/src/db/setup_local_dynamodb.py
hao10282025-sudo: Updated, set it as required variable with no "default" field

api/local.env
hao10282025-sudo: updated

api/src/db/setup_local_dynamodb.py
hao10282025-sudo: Updated

api/src/db/setup_local_dynamodb.py
hao10282025-sudo: Updated for all instances

api/src/db/setup_local_dynamodb.py
hao10282025-sudo: Updated as the example, thank you for all the details!

api/src/db/setup_local_dynamodb.py
```diff
@@ -0,0 +1,56 @@
...
+        )
+
+        _create_virus_scan_table(dynamodb, dynamodbconfig.file_scan_cache_table_name)
+
+
+def _create_virus_scan_table(
+    dynamodb: boto3.resources.base.ServiceResource, file_scan_cache_table_name: str
+) -> None:
+    """Create the local-virus-scan table if it doesn't already exist"""
+    logger.info("Creating DynamoDB table if it does not already exist")
```
kkrug: Let's add the `extra` dict here for consistency
```suggestion
    logger.info("Creating DynamoDB table if it does not already exist", extra={"table_name": file_scan_cache_table_name})
```

api/src/db/setup_local_dynamodb.py
```diff
@@ -0,0 +1,56 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def setup_local_dynamodb() -> None:
+    """Set up local DynamoDB tables for development"""
+    with src.logging.init(__package__):
+        error_if_not_local()
+
+        dynamodbconfig = DynamoDBConfig()
```
kkrug: Can we make this snake case for readability?
```suggestion
        dynamodb_config = DynamoDBConfig()
```

api/src/db/setup_local_dynamodb.py
```diff
@@ -0,0 +1,56 @@
...
+        table.wait_until_exists()
+        logger.info("Successfully created table", extra={"table_name": file_scan_cache_table_name})
+    except dynamodb.meta.client.exceptions.ResourceInUseException:
+        logger.info("Table already exists", extra={"table_name": file_scan_cache_table_name})
+    except Exception as e:
+        logger.exception(
+            "Unexpected error creating table",
+            extra={"table_name": file_scan_cache_table_name, "error_info": e},
+        )
+        raise
```
kkrug: I don't think we need this general exception at all - any uncaught exception will propagate automatically

api/src/db/setup_local_dynamodb.py
hao10282025-sudo: updated

api/src/db/setup_local_dynamodb.py
hao10282025-sudo: updated

api/src/db/setup_local_dynamodb.py
hao10282025-sudo: removed this part
I did a comparison of error messages for these two cases, looks not necessary
image

kkrug (approved):

PR #10148, opened 2026-05-05
[Issue #9603] Remove client-side state management from edit opportunity form

### LINKED ISSUE #9603: Remove client side state management from edit opportunity form
### Summary

With more extensive use of the server action to maintain state, client side state management should only be necessary on a couple of fields within the edit opportunity form. Any unnecessary usage of client side state should be removed and replaced by state managed by the server action.

See https://github.com/HHS/simpler-grants-gov/pull/9580 for some ideas on implementation

### Acceptance criteria

- [ ] any unnecessary use of client side state management in the edit opportunity form is replaced with server action led state management

desaiamit:
## Summary

Fixes #9603

## Changes proposed

- Removed `useState<OpportunityEditFormValues>` that tracked all form fields and re-rendered on every keystroke; replaced inputs with `defaultValue`/`defaultChecked` - the browser natively preserves values
- Retained minimal `useState` for 5 fields that require it: `fundingCategory`, `closeDate` (conditional rendering), `selectedEligibility`, `fundingType`, `publishDate` (CustomEvent payload)
- Replaced `useEffect` + dependency array dispatch with an explicit `dispatchValuesChange` helper using an override pattern
- Consolidated 5 duplicate inline eligibility toggle handlers into a single shared `handleEligibilityToggle`
- Fixed missing `<Label>` on the attachment file input in `OpportunityAttachmentUploadInput` - a11y violation caught by jest-axe
- Added `uploadLabel` translation key to `en/index.ts`

## Context for reviewers

The form previously tracked every field in a single `useState<OpportunityEditFormValues>` object, triggering a re-render on every keystroke. The bulk of this change replaces that with uncontrolled inputs - the browser natively preserves field values without React needing to manage them.

5 fields still require state: `fundingCategory` and `closeDate` drive conditional show/hide of explanation textareas and must be controlled; `selectedEligibility`, `fundingType`, and `publishDate` are needed to build the `opportunity-values-change` CustomEvent that keeps `OpportunityEditHeaderActions`' publish button in sync.

The override pattern in `dispatchValuesChange`: React state updates are async - calling `setState(newValue)` and immediately reading the state variable gives the old value. The override parameter passes the new value explicitly for the field that just changed, while reading current state for all other fields.

## Validation steps

1. Log in as `one_agency_opp_pub` (seed: `make db-seed-local-with-agencies` from `api/`)
2. Navigate to a draft opportunity edit page
3. Change Category to `Other` - verify Category Explanation textarea appears; change to another category - verify it disappears
4. Clear the Close Date - verify Close Date Explanation appears; set a date - verify it disappears
5. Check/uncheck eligibility checkboxes - verify the Publish button enables/disables correctly as eligibility becomes empty or non-empty
6. Clear the Publish Date - verify Publish button disables; set a date - verify it re-enables
7. Change the Category to a new value and Save - verify the selected category is preserved after save
8. Verify all typed field values are preserved after a save (not reset to original values)
9. Verify the Attachments section shows an "Upload files" label above the file input


glenblosser-nava (commented):
This is looking good to me, and has passed every manual test I've tried. @doug-s-nava would you like to take another look at it before I approve?

frontend/src/components/opportunity/OpportunityEditForm.tsx
```diff
@@ -110,53 +110,64 @@ export default function OpportunityEditForm({
...
   });
   const validationErrors: OpportunityEditValidationErrors | undefined =
     formState.validationErrors;
 
-  useEffect(() => {
+  // Dispatch CustomEvent so OpportunityEditHeaderActions can update
+  // the publish button enabled state in real time.
+  // Accepts overrides for the field that just changed — React state updates
+  // are async so reading state immediately after setState gives the old value.
+  function dispatchValuesChange(
```
doug-s-nava: confused why this function is necessary if we are tracking data in state. Can we do this with a useEffect keyed off of the state values instead?

doug-s-nava (commented):
based on my testing everything works, but I think we can simplify this a bit further

frontend/src/components/opportunity/OpportunityEditForm.tsx
desaiamit: That's exactly what the dependency array on line 137 does. The reason it dispatches a DOM event rather than directly setting state is that `OpportunityEditHeaderActions` is a sibling component with no shared parent - it has no access to OpportunityEditForm's state. A useEffect alone doesn't solve cross-sibling communication; something has to bridge the two components.

The two options are:

- Keep current approach: DOM event is contained and self-cleaning - listener registers on mount, removes on unmount. No changes to the page component.
- Lift state to the page: Move values state up to the page component, pass it as props to both OpportunityEditForm and OpportunityEditHeaderActions. Eliminates the DOM event entirely but adds state management to the page.


Let me know which option to go forward with

doug-s-nava: Yeah, I'm not proposing removing the event handler stuff as that's a bigger refactor, and we already have that ticketed elsewhere. However, I do not see an issue with removing manual calls to the dispatch handler and replacing with a useEffect. This seems to work for me https://github.com/HHS/simpler-grants-gov/commit/90f86ad7ae94ba7a5c663a37aa663c1c8de920c5

desaiamit: Understood - replaced dispatchValuesChange and its 3 manual call sites with a single useEffect keyed off [publishDate, fundingType, fundingCategory, selectedEligibility]. The effect fires after React commits each state change, so the event always carries current values with no override pattern needed.

desaiamit: Done - replaced dispatchValuesChange and its 3 manual call sites with a single useEffect keyed off [publishDate, fundingType, fundingCategory, selectedEligibility]. The effect fires after React commits each state change, so the event always carries current values. Let me know if anything looks off.

doug-s-nava (approved):
✅ ✅ ✅ ✅

PR #10158, opened 2026-05-06
test(util): add unit tests for src.util.local

### LINKED ISSUE #9815: Add unit tests for `src.util.local`
### Summary

Add unit tests for `src.util.local`, a small module with two functions that guard local-only operations. `error_if_not_local()` is used across the API (12 call sites including `src/app.py`, `src/task/forms/*`, and `src/services/local/get_local_users.py`) to prevent accidental execution of local-only code in deployed environments, but has no automated coverage.

## Background & Context

`api/src/util/local.py` exposes two functions: `load_local_env_vars(env_file)` loads a dotenv file only when `ENVIRONMENT` is unset or equals `"local"`; `error_if_not_local()` raises if `ENVIRONMENT` is anything other than `"local"`. Every other utility under `api/src/util/` has a sibling test file in `api/tests/src/util/`. This file is an orphan.

Rule reference: `documentation/rules/cross-domain.md` (CCP-3: tests ship with code).

## Getting Started

- File under test: `api/src/util/local.py`
- New test file: `api/tests/src/util/test_local.py`
- Local setup: `cd api && make format-check && make test`

## Technical Approach

Source (`api/src/util/local.py:9-31`):

```python
def load_local_env_vars(env_file: str = "local.env") -> None:
    environment = os.getenv("ENVIRONMENT", None)
    if environment is None or environment == "local":
        load_dotenv(env_file)

def error_if_not_local() -> None:
    if (env := os.getenv("ENVIRONMENT")) != "local":
        logger.error("Environment %s is not local - cannot run operation", env)
        raise Exception("Local-only process called when environment was set to non-local")
```

Use `monkeypatch` for env-var manipulation and `tmp_path` for a throwaway dotenv file. See `api/tests/src/util/test_deploy_metadata.py` for the canonical `monkeypatch.setenv` pattern.

## Testing Requirements

- Location: `api/tests/src/util/test_local.py`
- Fixtures: `monkeypatch`, `tmp_path`
- No DB session needed
- Run: `cd api && uv run pytest tests/src/util/test_local.py -v`

### Acceptance criteria

- [ ] A new test file exists at `api/tests/src/util/test_local.py`.
- [ ] `test_load_local_env_vars_when_environment_unset_loads_file` uses `monkeypatch.delenv("ENVIRONMENT", raising=False)` and asserts `load_local_env_vars` is invoked.
- [ ] `test_load_local_env_vars_when_environment_is_local_loads_file` sets `ENVIRONMENT=local` and asserts the same behavior.
- [ ] `test_load_local_env_vars_skips_in_non_local_env` sets `ENVIRONMENT=dev` and asserts no variables from the temp env file leak into `os.environ`.
- [ ] `test_error_if_not_local_raises_in_non_local` asserts `pytest.raises(Exception)` when `ENVIRONMENT=dev`.
- [ ] `test_error_if_not_local_succeeds_in_local` asserts returns `None` when `ENVIRONMENT=local`.
- [ ] `test_error_if_not_local_raises_when_environment_unset` asserts raise when env var absent.
- [ ] `cd api && make test` passes.

mvanhorn:
## Summary

Resolves #9815. Adds the missing unit tests for `api/src/util/local.py`. Every other module under `api/src/util/` has a sibling test file in `api/tests/src/util/`; `local.py` was the orphan despite `error_if_not_local()` being on the call path of `src/app.py`, `src/task/forms/*`, and `src/services/local/get_local_users.py`.

## Why this matters

`error_if_not_local()` is the guard that prevents accidental execution of local-only code paths in deployed environments. With no tests, a regression that swaps the early-return condition (e.g., a wrong `==` -> `!=`) would silently flip the production safety net to a no-op. The tests pin both the raise paths (`ENVIRONMENT=dev`, `ENVIRONMENT` unset) and the success path (`ENVIRONMENT=local`).

The cross-domain rule (`documentation/rules/cross-domain.md` CCP-3: tests ship with code) was already cited in the issue.

## Changes

- New file `api/tests/src/util/test_local.py` covering all six cases listed in the issue's acceptance criteria:
  - `test_load_local_env_vars_when_environment_unset_loads_file`
  - `test_load_local_env_vars_when_environment_is_local_loads_file`
  - `test_load_local_env_vars_skips_in_non_local_env`
  - `test_error_if_not_local_raises_in_non_local`
  - `test_error_if_not_local_succeeds_in_local`
  - `test_error_if_not_local_raises_when_environment_unset`

Pattern follows `api/tests/src/util/test_deploy_metadata.py` (`monkeypatch.setenv` / `monkeypatch.delenv(..., raising=False)` for env manipulation, `tmp_path` for the throwaway dotenv).

## Testing

- `pytest tests/src/util/test_local.py -v` — 6/6 pass.
- `black --check tests/src/util/test_local.py` — clean.
- `ruff format --check tests/src/util/test_local.py` — clean.

Fixes #9815


btabaska:
This PR is super close and the core intent is great — adding coverage for src.util.local closes a real gap. I like that the tests hit both local and non-local behavior and use monkeypatch/tmp_path patterns consistent with the repo. Main feedback is tightening the test shape to match our API test conventions

mvanhorn:
Thanks @btabaska. Pushed a4d06fd that tightens the test shape to match the conventions in `api/tests/src/util/test_datetime_util.py` and `test_dict_util.py`:

- Parametrized the two 'loads file' cases (environment unset vs 'local') since the assertion is identical - now one test with `pytest.param(..., id='environment_unset')` / `id='environment_local'`.
- Grouped the three `error_if_not_local` cases into `TestErrorIfNotLocal`, matching the `TestParseGrantsGovDate` pattern - class with docstring, method docstrings.
- Kept the `monkeypatch` / `tmp_path` patterns since you flagged those as good.

Same 6 scenarios covered, all pass. Let me know if you had a different convention in mind.

api/tests/src/util/test_local.py
```diff
@@ -0,0 +1,73 @@
+import os
+
+import pytest
+
+from src.util.local import error_if_not_local, load_local_env_vars
```
btabaska: bug: Heads up — the ground shifted under this one. Commit `0dee615ac` ([#10629], "Pull logging and a few other utilities over from grants_shared") **deleted** `api/src/util/local.py`; the module now lives at `backend/grants_shared/src/grants_shared/util/local.py` and everything imports it as `from grants_shared.util.local import ...` (see `api/src/app.py:13`). So two things need to change together:

1. The import `from src.util.local import ...` will throw `ModuleNotFoundError` at collection time on current `main` — the whole file silently runs zero tests. It needs to become `from grants_shared.util.local import error_if_not_local, load_local_env_vars`.
2. Per our `api-tests` convention (tests mirror the source tree), the file itself should move to `backend/grants_shared/tests/grants_shared/util/test_local.py`, right alongside the `test_datetime_util.py` / `test_deploy_metadata.py` siblings the PR description references — those moved too.

Worth a rebase on latest `main` and re-running `pytest` from the `grants_shared` package to confirm collection. Let me know if you want a hand finding the new home.

api/tests/src/util/test_local.py
```diff
@@ -0,0 +1,73 @@
...
+    else:
+        monkeypatch.setenv("ENVIRONMENT", environment_value)
+    monkeypatch.delenv("LOCAL_TEST_FROM_DOTENV", raising=False)
+
+    env_file = tmp_path / "local.env"
+    env_file.write_text("LOCAL_TEST_FROM_DOTENV=loaded\n")
+
+    load_local_env_vars(env_file=str(env_file))
+
+    assert os.environ["LOCAL_TEST_FROM_DOTENV"] == "loaded"
```
btabaska: testing: The `load_dotenv` no-override behavior is the safety-critical bit and it's currently uncovered. python-dotenv won't overwrite an env var that's already set, which is exactly what protects real secrets in `local`. Could we add a case where `ENVIRONMENT=local`, `LOCAL_TEST_FROM_DOTENV` is pre-set to something like `"pre-existing"` via `monkeypatch.setenv`, the `.env` file sets it to `"loaded"`, and we assert it's still `"pre-existing"` afterward? That pins the contract so a future `override=True` regression can't slip through.

btabaska: testing: We only ever exercise the explicit `env_file=str(env_file)` path, so the default (`env_file="local.env"`) wiring is untested. A small companion test with `monkeypatch.chdir(tmp_path)`, a `local.env` written into that dir, and a bare `load_local_env_vars()` call would cover the default resolution too.

btabaska (changes requested):

mvanhorn:
Updated in 7c1ac78d. Merged main to pick up the #10629 move - since grants_shared brought its own baseline test_local.py, I removed the now-redundant api/tests copy rather than relocating it wholesale. The two gaps you flagged are added to backend/grants_shared/tests/grants_shared/util/test_local.py: a no-override case (LOCAL_TEST_FROM_DOTENV pre-set to 'pre-existing' survives load_local_env_vars) and a default-resolution case (bare load_local_env_vars() with monkeypatch.chdir picking up local.env from the working directory). 16 tests pass, ruff clean.

PR #10223, opened 2026-05-06
[Issue #9644] Remove auto summation from SF424A section A 

### LINKED ISSUE #9644: Fix editable pre-population field in SF-424A
### Summary

SF-424A Bug filed to highlight 4 changes required:
1. Section A's Totals for rows 1-4 need to be editable.
2. Section C column A needs to be editable. 
3. Section E Column A needs to be editable.
4. We should not pre-populate Sections A, C and E unless we can set parameters (e.g., force=1) to give precedence to the user's intended pre-filled values.

**Description**
In order to avoid blocking necessary edits for various grant applicant types, we should make the Total Column of Section A of SF-424A editable and avoid pre-populating Section A totals unless we can force accepting user-entered values. The DAT Business Rules for Section A of SF-424A indicate that Total values are “calculated to be the sum of the row, however value may be overwritten” for A-1-7, A-2-7, A-3-7, and A-4-7. This aligns with the SF-424A [DAT SF424A-V1.0_F241.xls](https://github.com/user-attachments/files/26791104/DAT.SF424A-V1.0_F241.xls) instructions for TOTALS in Section A, which show variability in auto-summation based on the kind of application. So until we are allowing filtering by application type (explored in Q5 Budget Form concept testing), we should allow the user to edit their Totals column in Section A.

Furthermore, the DAT Business Rules for Sections C and E of SF-424A  stipulates that for Section C column A and Section E Column A, “Forward populated value may be overwritten. If no entry is made, and corresponding data is entered, this field will populate with 'N/A' upon submission.”

Images for reference to expected behavior. Please note: these mockups do not imply style changes; we're only addressing the field input behavior at this time.

**Section A:** Make "Total 1-4" editable - do not auto-sum and do not pre-populate unless we can set parameters (e.g., force=1) to give precedence to the user's intended pre-filled values.

Current Behavior: Section A
Image

Expected Behavior: Section A
Image

**Section C:** Make column A editable - do not pre-populate unless we can set parameters (e.g., force=1) to give precedence to the user's intended pre-filled values.
Current Behavior: Section C
Image

Expected Behavior: Section C
Image

**Section E:** Make column A editable - do not pre-populate unless we can set parameters (e.g., force=1) to give precedence to the user's intended pre-filled values.
Current Behavior: Section E
Image

Expected Behavior: Section E
Image

### Reproducibility

None

### How to reproduce the bug

View current SF-424A example in all environments.

### Browser

None

### Code of Conduct

babebe:
## Summary

Fixes / Work for #9644  

## Changes proposed

- Made budget_summary.total_amount a user-input field
  -   Removed auto-calculation of row G (per activity line item)
  -   Updated fixtures to explicitly include total_amount
- Updated total budget summary calculation (row 5)
  - total_budget_summary.total_amount is now calculated only from columns C–F
  - Explicitly ignores activity_line_items[*].budget_summary.total_amount (row G)
- Updated tests
  - Removed assumptions that row G is computed and updated fixtures to reflect new behavior
  - Added test to ensure:
     - Row G is preserved as user input
     - Row 5 total is derived strictly from C–F values
     

## Validation steps

1.   Run frontend and api locally ensure ensure 
2.  Find application  for  testing form SF424A, Can login with `many_app_user` and use `Test Opportunity for SF424A`
3.  Verify:
   - Each line item’s total_amount (row G) is editable and remains exactly as entered when saved
   - total_budget_summary.total_amount equals the sum of C–F across all line items and is pre-populated
   - Changing row G does not affect the total


api/src/form_schema/forms/sf424a.py
```diff
@@ -495,13 +480,23 @@
...
                 "rule": "sum_monetary",
-                "fields": ["activity_line_items[*].budget_summary.total_amount"],
-                # Run this in the 2nd iteration after the total_amount of the activity line items is calculated
+                "fields": [
+                    "activity_line_items[*].budget_summary.federal_estimated_unobligated_amount",
+                    "activity_line_items[*].budget_summary.non_federal_estimated_unobligated_amount",
+                    "activity_line_items[*].budget_summary.federal_new_or_revised_amount",
+                    "activity_line_items[*].budget_summary.non_federal_new_or_revised_amount",
+                ],
+                # Run after line-item budget_summary values are finalized
```
chouinar, line 511: Slight optimization. This would have to grab up to 16 items (C1->C4, D1->D4, E1->E4, F1-F4). If we're still calculating C5, D5, E5, and F5, then we can instead use those to sum since those are calculated from these fields.

```suggestion
                "fields": [
                    "total_budget_summary.federal_estimated_unobligated_amount",
                    "total_budget_summarynon_federal_estimated_unobligated_amount",
                    "total_budget_summary.federal_new_or_revised_amount",
                    "total_budget_summary.non_federal_new_or_revised_amount",
                ],
                # Run after line-item budget_summary values are finalized
```

chouinar (approved):

PR #10229, opened 2026-05-06
[Issue #10224] Ensure we use a committed DB model for JWT creation (login) logging.

### LINKED ISSUE #10224: Fix logging of user_id when logging login/logout
### Summary

Log in (JWT creation) isn't successfully logging the user_id who signed in. 

### Acceptance criteria

- [ ] Logs include "auth.user_id" for JWT log entry
- [ ] We don't use str() in the logs so that null won't be logged as "None"
- [ ] Tests pass

mdragon:
## Summary

Fixes #10224 

## Changes proposed

Get user_id from the committed user object not the uncommitted token object were the value doesn't resolve until commit. Remove str so we get nulls not "None" in NR

## Context for reviewers

We're not currently outputting the user_id on Login ("JWT Created") events because we haven't committed the token record yet.

## Validation steps

- Checks pass
- Value logged

### DIFF
--- a/api/src/api/users/user_routes.py
+++ b/api/src/api/users/user_routes.py
@@ -160,8 +160,8 @@ def user_token_logout(db_session: db.Session) -> response.ApiResponse:
     logger.info(
         "Logged out a user",
         extra={
-            "user_token_session.token_id": str(user_token_session.token_id),
-            "user_token_session.user_id": str(user_token_session.user_id),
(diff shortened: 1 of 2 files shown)


mdragon:
Confirmed during local testing that the "Created JWT token" log entry now has a user_id present.

chouinar (approved):

Issue #10233, opened 2026-05-06
Remove defaultSize from FieldList and rely on minItems / maxItems


myduong-navapbc:
### Summary

Remove support for the `defaultSize` property from `FieldList`. `FieldList` entry count should be determined only by `minItems` and `maxItems`.

### Acceptance criteria

- [ ] `defaultSize` is removed from:
  - [ ] `FieldList` widget props
  - [ ] UI schema definition
  - [ ] JSON schema validation (if applicable)
- [ ] `FieldList` initial render behavior:
  - [ ] If `minItems` > 0, render `minItems` number of entries
  - [ ] If `minItems` = 0 or undefined, render no entries
- [ ] `maxItems` continues to:
  - [ ] prevent adding entries beyond the limit
  - [ ] disable or hide the add control when reached
- [ ] All existing usages of `defaultSize` are removed
- [ ] Unit tests updated to reflect removal of defaultSize

megreed1:
Reviewer: @jesababon 
@chris-kuryak will need to huddle with @myduong-navapbc for UAT on this after Jes reviews.

ErinPattisonNava:
We are going to do UAT in https://github.com/HHS/simpler-grants-gov/issues/9761 so marking this as closed/done and we will track UAT in a future sprint there.

PR #10258, opened 2026-05-07
[Issue #10233] Removed defaultSize from the fieldList feature

myduong-navapbc:
## Summary

Work for #10233 

Remove `defaultSize` from `FieldList` configuration and rely on `minItems / maxItems` to control entry count behavior.

This updates `FieldList` so initial rows are driven by schema constraints: lists with `minItems` > 0 render that number of blank entries, while optional lists with `minItems: 0` or no `minItems` render with no entries. Existing `maxItems` behavior continues to prevent adding entries beyond the configured limit.

## Validation steps

- [x] Sign in and open the FieldList min/max demo form.
- [x] Verify a `FieldList` with `minItems: 0` starts with no entries.
- [x] Verify a `FieldList` with no `minItems` starts with no entries.
- [x] Verify a `FieldList` with `minItems: 1` starts with one blank entry.
- [x] Verify a `FieldList` with `minItems: 2 `starts with two blank entries.
- [x] Verify users cannot delete entries below the configured `minItems`.
- [x] Verify users cannot add entries beyond the configured `maxItems`.
- [x] Verify the add control is disabled or hidden when maxItems is reached.
- [x] Save the form and confirm `FieldList` validation behavior still works.
- [x] Confirm no `defaultSize` references remain in source or test fixtures.


frontend/src/components/applyForm/widgets/FieldListWidget.tsx
```diff
@@ -300,27 +302,43 @@ function FieldListWidget(widgetProps: FieldListWidgetProps) {
    * The entries are rehydrated from the incoming `value` prop so the widget
    * stays aligned with saved form data after save / reload.
    */
-  const [rows, setRows] = useState<GeneralRecord[]>(
-    normalizeFieldListRows({
-      value,
-      defaultSize,
-    }),
+  const normalizedInitialRows = useMemo(
```
ErinPattisonNava: What does normalized mean here?

frontend/src/components/applyForm/utils.test.ts
```diff
@@ -596,7 +596,6 @@ describe("buildWarningTree", () => {
         type: "fieldList" as const,
         name: "contacts",
         label: "Contacts",
-        defaultSize: 1,
```
ErinPattisonNava, line 599: Should some of these tests have a minLength of 1?  I'm confused if we aren't passing it in that the tests are passing, as I assume no value is 0.

ErinPattisonNava (commented):
1)  If there are 3 entries and you delete the second one, it actually deletes the third entry.


2) If I make the partner organizations have a min of 2 and a max of 2 it looks right, but errors talk about an array.
 
Image


3)  Have a form where the default fields are filled in correctly and saved.  If you go to the field list entry where the min is 0 and add two entries a value in each entry, this error appears.  Save the form.  The values in the entries you just added are erased.  Go back to the application and then click into the form again.  The values reappear.

Image

frontend/src/components/applyForm/widgets/FieldListWidget.tsx
myduong-navapbc: i renamed the variable to: `initialFieldListRows`.

frontend/src/components/applyForm/utils.test.ts
myduong-navapbc: I removed the `defaultSize` reference here since that prop was removed from `FieldList`. We should not add `minItems` to the UI schema because entry constraints belong to the JSON schema, not the UI schema.

myduong-navapbc:
> 1. If there are 3 entries and you delete the second one, it actually deletes the third entry.
> 2. If I make the partner organizations have a min of 2 and a max of 2 it looks right, but errors talk about an array.
> 
> Image
> 3. Have a form where the default fields are filled in correctly and saved.  If you go to the field list entry where the min is 0 and add two entries a value in each entry, this error appears.  Save the form.  The values in the entries you just added are erased.  Go back to the application and then click into the form again.  The values reappear.
> 
> Image

I made fixes for all three issues in the form data shaping/pruning utilities.

The cause was that the prune logic was stripping empty objects from FieldList arrays before save. That caused the saved payload to no longer match the schema-defined minItems expectations, which then produced array-level validation errors and incorrect row/index behavior.

The updates now:

- preserve empty FieldList rows required by minItems
- maintain stable row ordering/indexes during add/remove operations
- prevent newly added row values from disappearing after save
- allow validation to correctly target row-level required fields instead of array-level errors

After refresh/reload, the saved values now hydrate correctly from persisted form data.

myduong-navapbc:
Here's an example of a fieldList with 2 empty entries that have required fields

(min: 2 / max: 2)

Screenshot 2026-05-14 at 1 36 25 PM


Screenshot 2026-05-14 at 1 35 47 PM

frontend/src/components/applyForm/widgets/FieldListWidget.tsx
```diff
@@ -13,67 +13,84 @@ import {
...
  * `BroadlyDefinedWidgetValue` inside `handleFieldChange`.
  */
 type FieldListChangeParams = {
-  entryIndex: number;
+  entryId: string;
   storageKey: string;
   nextValue: unknown;
 };
 
+type FieldListEntryValue = {
```
ErinPattisonNava: This is just the FieldListEntry now, right?

frontend/src/components/applyForm/widgets/FieldListWidget.tsx
```diff
@@ -13,67 +13,84 @@ import {
...
+ * Saved form data is stored as `GeneralRecord[]`, but the widget needs stable
+ * client-side entry ids so React can preserve each entry's local state when
+ * entries are added or removed. The entry ids are only used for rendering and
+ * are stripped before values are sent back to the parent form.
  *
+ * If the schema defines `minItems`, blank entries are appended so the UI shows
+ * the required minimum number of entries.
  */
-const normalizeFieldListRows = ({
+const normalizeFieldListEntryValues = ({
```
ErinPattisonNava: What does normalized mean here?

frontend/src/components/applyForm/widgets/FieldListWidget.test.tsx
```diff
@@ -34,6 +34,29 @@ jest.mock("src/components/applyForm/widgets/WidgetRenderers", () => ({
...
+
+Object.defineProperty(globalThis, "crypto", {
+  value: {
+    randomUUID: randomUuidMock,
+  },
+});
+
+let randomUuidIndex = 0;
+
+Object.defineProperty(globalThis, "crypto", {
```
ErinPattisonNava: Isn't this a duplicate and will overwrite the property on line 39?

frontend/src/components/applyForm/widgets/FieldListWidget.tsx
```diff
@@ -13,67 +13,84 @@ import {
...
+    ...startingEntries,
+    ...Array.from({ length: missingEntryCount }, () => ({
+      entryId: createFieldListEntryValueId(),
+      value: {},
+    })),
   ];
 };
 
+const createFieldListEntryValueId = (): string => {
+  return crypto.randomUUID();
```
ErinPattisonNava: What are the implications of using something in the browser here?  I wish from a testing perspective it was a library so it wouldn't require a mock, but that isn't a blocking issue.  Is this always going to be a client component so the browser is always going to be available?

frontend/src/components/applyForm/widgets/FieldListWidget.tsx
```diff
@@ -244,24 +259,28 @@ function FieldListEntry({
...
+         * must update local state before add/delete operations so unsaved entry values
+         * are not lost when indexes shift.
          */
         const childWidgetProps: UswdsWidgetProps = {
           ...groupItem.generalProps,
           schema: groupItem.generalProps.schema as RJSFSchema,
           id: generatedId,
           name: generatedId,
-          key: generatedId,
+          key: `${entryId}-${storageKey}`,
```
ErinPattisonNava, line 274: Can this just be entryId?

ErinPattisonNava (commented):
Functionality looks great, some questions on naming.

frontend/src/components/applyForm/widgets/FieldListWidget.tsx
myduong-navapbc: technically, function is doing two things:

- converting raw form values into internal `FieldList` entry objects
- shapes the missing entries of `minItems`

we could say `createInitialFieldListEntries` but normalize is the more accurate description

frontend/src/components/applyForm/widgets/FieldListWidget.test.tsx
myduong-navapbc: Good catch!

frontend/src/components/applyForm/widgets/FieldListWidget.tsx
myduong-navapbc: It actually cannot-

The `entryId` alone is enough here because each `FieldList` entry renders multiple child widgets. If every child field in the same entry used the same `entryId` key, those sibling widgets would share a key. 

(And this was the issue I had originally - and the react warning that kept popping up)

The `storageKey` makes the key unique per child field while still keeping it stable to the entry. So `${entryId}-${storageKey}` gives us stable identity for each field inside a stable entry.

frontend/src/components/applyForm/widgets/FieldListWidget.tsx
myduong-navapbc: It should, since we already use client-side feats like `useRef, useMemo, and useStatte`and other browser-driven input event.

I do understand your point.

Since these ids are only used for temporary client-side entry identity and React rendering, I removed the crypto.randomUUID dependency entirely. We do not persist these values outside the widget, so a lightweight generated id is sufficient here and avoids browser/runtime assumptions and test mocking overhead.

I added the counter- just-in-case we have create  multiple entries within the same milliseconds (not probable but could happen)

frontend/src/components/applyForm/widgets/FieldListWidget.tsx
myduong-navapbc: It can! ill update it.

ErinPattisonNava (approved):
I didn't get a chance to test the new way of assigning entry id, but the code otherwise is good.  Let's put that as an AC on the next ticket for the FieldEntry so we can move this ticket forward.

PR #10289, opened 2026-05-07
[DO NOT MERGE] Prototype of splitting out an API + a shared package

chouinar:
## Changes proposed
* Created a new `grants_shared` python package where we put all of our 
* Created a new minimal working `example` API that uses 

## Context for reviewers

The purpose of this work was to investigate what challenges we'll face in doing the data separation. It's not meant to be merged, it's purely illustrative to show how much work this is and document the challenges. I've documented many of the challenges on https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/3059286063/Data+Separation+-+Backend+Requirements+Research#Prototype-separation

Note that at the time of creating this, we haven't moved to uv (although I'll merge that shortly). I just set this up using UV because it's easier.

Important things that haven't been done:
* Dockerfile working
* Actually publishing the grants_shared code
* All tests (got lazy, stopped copying the unit tests over - some need a bit of work to work due to paths / utils)
* Any meaningful functionality for API endpoints or actual DB tables
* Anything auth related (we're going to investigate whether that's a service or library)
* There's a bunch of env configuration missing - my setup might even make that not fail for you

## Validation steps
Mainly just getting things working locally - which requires having uv installed.

I don't know how much of this is necessary, but roughly:
* In the shared folder run `uv sync --all-groups` and `uv build`
* In the example folder run `uv sync --all-groups`
* To run the API natively, run `uv run flask --app src.app run --host 0.0.0.0 --port 8080 --reload`


chouinar:
I'd kept this open for a bit as a reference as we did the work that this was prototyping out, but since we've done almost all of it at this point, no need to keep it anymore.

Issue #10279, opened 2026-05-07
SOAP/Proxy: fix proxy url


jakobpederson:
### Summary

The `proxy_url` on the JWT path is being generated incorrectly.

Currently it's:
`proxy_url='https://{gateway_uri}/grantsws-agency/services/v2/AgencyWebServicesSoapPort'`

When it should be:
`proxy_url='https://{gateway_uri}/grantsws-agency-partner/services/v2/AgencyWebServicesSoapPort'`

This is causing requests to fail.

### Acceptance criteria

- [ ] can hit the endpoint and get correct response

PR #10293, opened 2026-05-07
[Issue #10279] Fix SOAP proxy url

jakobpederson:
## Summary

Fixes / Work for #10279  

## Changes proposed

Updated the `proxy_url` calculated from data in the AWS Parameter Store.
Removed check for `legacy_certificate` on the headers. This is no longer functionality we want to check. 

## Context for reviewers

The url was incorrect so JWT path was failing. This should correct that.
If there's no `legacy_certificate` on on the `SOAPAuth` it should still use JWT.

If we go with env variables, these will need to be added to each env before merge and deploy.

## Validation steps

- [ ] hit url in lower env and get correct response

### DIFF
--- a/api/src/legacy_soap_api/legacy_soap_api_proxy.py
+++ b/api/src/legacy_soap_api/legacy_soap_api_proxy.py
@@ -16,7 +16,11 @@ from src.legacy_soap_api.legacy_soap_api_auth import (
     SOAPClientMissingCertificate,
     generate_soap_jwt,
 )
-from src.legacy_soap_api.legacy_soap_api_config import LegacySoapAPIConfig, get_soap_config
+from src.legacy_soap_api.legacy_soap_api_config import (
+    LegacySoapAPIConfig,
+    SimplerSoapAPI,
+    get_soap_config,
+)
 from src.legacy_soap_api.legacy_soap_api_constants import LegacySoapApiEvent
 from src.legacy_soap_api.legacy_soap_api_schemas import SOAPResponse
 from src.legacy_soap_api.legacy_soap_api_schemas.base import SOAPRequest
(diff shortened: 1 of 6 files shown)


api/src/legacy_soap_api/legacy_soap_api_proxy.py
```diff
@@ -59,6 +59,7 @@ def get_proxy_response(soap_request: SOAPRequest, timeout: int = PROXY_TIMEOUT)
             config.soap_partner_gateway_uri,
             soap_request.full_path.lstrip("/"),
         )
+        proxy_url = proxy_url.replace("grantsws-agency", "grantsws-agency-partner")
```
chouinar: If the proxy_url is just an environment variable, why not change that to be this value rather than do a replace?

jakobpederson: I was debating that, but I was worried about the application route. 

The incoming url can either be:
`grantsws-agency/services/v2/AgencyWebServicesSoapPort`

or
`grantsws-applicant/services/v2/ApplicantWebServicesSoapPort`

And while we don't have any JWT endpoints using the Applicant path, I thought we should just be able to handle it.

api/tests/src/legacy_soap_api/test_legacy_soap_api_proxy.py
```diff
@@ -97,12 +101,41 @@ def test_request_gets_correct_proxy_url_when_no_auth(enable_factory_create):
     soap_request.auth = None
     with patch("src.legacy_soap_api.legacy_soap_api_proxy._get_soap_response") as mock_request:
         get_proxy_response(soap_request)
-        expected = (
-            "https://google.com/xyz/grantsws-agency-partner/services/v2/AgencyWebServicesSoapPort"
-        )
+        expected = f"https://google.com/xyz/{soap_request.full_path.lstrip('/')}"
```
kkrug: Would it make sense to have kept the expected url hardcoded? If there was an issue with `soap_request.full_path`, this test wouldn't catch it.

api/tests/src/legacy_soap_api/test_legacy_soap_api_proxy.py
```diff
@@ -97,12 +101,41 @@ def test_request_gets_correct_proxy_url_when_no_auth(enable_factory_create):
...
         assert mock_request.call_args_list[0][0][0].url == expected
 
 
+def test_request_gets_correct_proxy_url_when_request_has_auth(enable_factory_create, monkeypatch):
```
kkrug: Looks like we can remove `monkeypatch` as an arg here.

api/src/legacy_soap_api/legacy_soap_api_proxy.py
kkrug: I think this replace is a bit fragile. If the url contained "grants-agency-partner" then the replace would make it "grants-agency-partner-partner" (not sure if this is feasible but is theoretically possible).

chouinar: Yeah, I'd rather we just define what we need when the values are static. If applicant needs a different path, we should just define two env vars and add logic to pick one based on what was passed in.

api/tests/src/legacy_soap_api/test_legacy_soap_api_proxy.py
jakobpederson: I'm conflicted because the issue with the `soap_request.full_path` shouldn't change the fact that at that part of the code that's where it gets the value. But also hard coding it won't be too big a problem either. So I'll update it.

kkrug (approved):

Issue #7960, opened 2026-01-13
Audit and adjust our DB user privileges


chouinar:
### Summary

We have the following users in our DB:
* `root` - The default superuser with access to everything in the DB.
* `migrator` - A user we give the ability to create schemas and tables, only used by DB migrations.
* `app` - The user we use for interacting with the database. Has read/write privileges on tables (as well as sequences, functions, and anything else).

We know that some of these users have incorrect privileges (app has the ability to create tables in the legacy schema). We want to correct those and make sure the processes we have in place keep them aligned.

All schemas, tables, and anything else in them (sequences/functions/triggers) should be owned by the `migrator` user.

The `app` user should have read/write access on all of these, but should not be able to create tables/schemas/etc.

---

Right now, the `migrator` user is used by our DB migrations, but that's not the only place where we create tables. Our legacy DB tables (that connect to grants.gov's Oracle DB) were created by a mix of the root user and the `app` user. This is because we have a backend ECS task that we use to create the tables that runs as the app user. This `setup-foreign-tables` script only gets run manually, but should use the migrator user, not sure how we want to account for that.

### Acceptance criteria

- [ ] Noted down current access for users in the DB
- [ ] Corrected, or created follow-up tickets to correct the access for users in our DB

kkrug:
@chouinar I created a doc with some tables showing my audit results, let me know what you think before I make any code changes
https://docs.google.com/document/d/16I6cAMrwYEIdCrFtvUo_2Y0MI30z90-W8d9D_s2tWCA/edit?usp=sharing

virajkothari-svg:
@chouinar  will review

PR #10301, opened 2026-05-08
[Issue #7960] Audit and adjust our DB user privileges

kkrug:
## Summary

Fixes for #7960  

## Changes proposed

- Updated run-setup-foreign-tables to add `DB_USER` env var override to connect as migrator instead of app
- Updated manage.py to do the following:
  - Add `get_all_schema_names()` to dynamically discover schemas by querying `pg_namespace`
  - Updated `configure_database()` to loop over all schemas instead of just the primary one
  - Updated `configure_schema()` to revoke CREATE from app, grant on existing objects, and fix object ownership
  - Add `fix_schema_object_ownership()` to reassign to reassign tables/foreign tables/sequences/views not owned by migrator
  - Updated `configure_default_privileges()` to set default privileges for all discovered schemas

## Context for reviewers

We audited DB user privileges across api-dev and api-prod and found that the legacy schema and all 31 foreign tables within it are owned by root or app instead of migrator. The app user also has CREATE privilege on the legacy schema, which it shouldn't. This happened because `setup-foreign-tables` was using the migrator IAM role but never overrode `DB_USER`, so it connected to Postgres as app. The original batch of foreign tables was created by root directly.

The role manager lambda previously only configured the primary api schema — legacy and staging were never managed. These changes make the role manager discover and configure all schemas on every deploy, and fix existing ownership. See the ticket for the full audit tables

## Validation steps

- Deploy to dev and verify the role manager lambda logs show it discovering and configuring api, legacy, staging, and EGRANTSADMIN schemas
- Re-run audit queries against api-dev to confirm fixes. Connect to the app database via RDS Query Editor:
  - Schema ownership and privileges - verify legacy is owned by migrator and app only has USAGE (no CREATE):
    ```
      SELECT nspname, nspowner::regrole::text, nspacl::text[]
      FROM pg_namespace
      WHERE nspname NOT LIKE 'pg_%'
      AND nspname NOT IN ('information_schema', 'public');
    ```
  - Foreign table ownership - verify all tables in legacy are owned by migrator:
    ```
      SELECT c.relname, c.relkind::text, pg_roles.rolname AS owner
      FROM pg_class c
      JOIN pg_namespace n ON n.oid = c.relnamespace
      JOIN pg_roles ON pg_roles.oid = c.relowner
      WHERE n.nspname = 'legacy'
      ORDER BY c.relname;
    ```
  - Default privileges - verify migrator has schema-scoped default privileges for legacy and staging:
    ```
      SELECT pg_roles.rolname, nspname, defaclobjtype::text, defaclacl::text[]
      FROM pg_default_acl
      JOIN pg_namespace ON pg_namespace.oid = defaclnamespace
      JOIN pg_roles ON pg_roles.oid = defaclrole;
    ```
- Verify the API application still functions normally (app can still read/write all tables)
- Run `setup-foreign-tables` in dev to confirm it connects as migrator
- Repeat audit queries against api-prod after prod deploy

### DIFF
--- a/infra/modules/database/role_manager/manage.py
+++ b/infra/modules/database/role_manager/manage.py
@@ -17,7 +17,7 @@ def manage(config: dict):
         roles, schema_privileges = print_current_db_config(master_conn)
         roles_with_groups = get_roles_with_groups(master_conn)
 
-    configure_default_privileges()
+    configure_default_privileges(config)
 
     return {
         "roles": roles,
@@ -75,33 +75,12 @@ def get_schema_privileges(conn: Connection) -> list[tuple[str, str]]:
     ]
 
 
-def get_all_schema_names(conn: Connection, primary_schema: str) -> list[str]:
-    """Discover all application schemas in the database.
-
-    Returns the primary schema first, followed by any other non-system schemas.
-    """
-    rows = db.execute(
-        conn,
-        """
-        SELECT nspname
-        FROM pg_namespace
-        WHERE nspname NOT LIKE 'pg_%'
-        AND nspname NOT IN ('information_schema', 'public')
-        ORDER BY nspname
-        """,
-        print_query=False,
-    )
-    schemas = [row[0] for row in rows]
-    if primary_schema in schemas:
-        schemas.remove(primary_schema)
-    return [primary_schema] + schemas
-
-
 def configure_database(conn: Connection, config: dict) -> None:
     print("-- Configuring database")
     app_username = os.environ.get("APP_USER")
     migrator_username = os.environ.get("MIGRATOR_USER")
-    schema_name = os.environ.get("DB_SCHEMA")
+    schema_name = config.get("schema_name", os.environ.get("DB_SCHEMA"))
+    primary_schema = os.environ.get("DB_SCHEMA")
     database_name = os.environ.get("DB_NAME")
 
     # In Postgres 15 and higher, the CREATE privilege on the public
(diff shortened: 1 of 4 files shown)


kkrug:
@chouinar These were the small code changes I mentioned. The audit document is linked in the ticket and also [here](https://docs.google.com/document/d/16I6cAMrwYEIdCrFtvUo_2Y0MI30z90-W8d9D_s2tWCA/edit?usp=sharing) for reference.

infra/modules/database/role_manager/manage.py
```diff
@@ -98,7 +120,10 @@ def configure_database(conn: Connection, config: dict) -> None:
     )
 
     configure_roles(conn, [migrator_username, app_username], database_name)
-    configure_schema(conn, schema_name, migrator_username, app_username)
+
+    for s in get_all_schema_names(conn, schema_name):
+        configure_schema(conn, s, migrator_username, app_username)
```
chouinar: This changes a little bit of how this script works - it grabs schemas that might not be right to grab/adjust. Right now it wouldn't be an issue, but let me pitch a future scenario in the grantor system - we have multiple schemas, let's say `grantor` and `payment`. `grantor` contains most things, but `payment` is a separate schema we want locked down to a specific user that only a payment process can interact with.

As written, this script would grant the same permissions for our users to everyone. Migrator, that's fine probably, but the app user shouldn't have that. As written, we wouldn't be able to configure that.

Instead, what if we just know we need to run this script for each type of user? Let's assume we have a `grantor` and `payment` user. We'd run the script once for each of them, passing in that user + schema name. Effectively, the change is just to run this with `DB_SCHEMA` overriden.

infra/modules/database/role_manager/manage.py
```diff
@@ -131,7 +156,7 @@ def configure_role(conn: Connection, username: str, database_name: str) -> None:
 
 
 def configure_schema(conn: Connection, schema_name: str, migrator_username: str, app_username: str) -> None:
-    print("---- Configuring schema")
+    print(f"---- Configuring schema: {schema_name}")
     print(f"------ Creating schema: {schema_name=}")
     db.execute(conn, f"CREATE SCHEMA IF NOT EXISTS {identifier(schema_name)}")
```
chouinar, line 140: To go along with my other comment - if we get the list of non-main DB schemas from the DB, how would we create those other schemas?

infra/modules/database/role_manager/manage.py
kkrug: I think I see what you mean. I updated it to now read `schema_name` from the event config payload instead of the env var, and processes a single schema per invocation. The invocation script (create-or-update-database-roles) reads `additional_schema_names` from the database config and loops, calling the lambda once per schema with the `schema_name` overridden in the payload

chouinar (approved):
LGTM! This'll also be helpful to have when we setup the grantor DB.

PR #10322, opened 2026-05-08
[Unticketed] Misc test cleanup on the backend

chouinar:
## Changes proposed
* Various cleanup pieces in our tests including:
  * Removing print statements + added a lint rule to catch
  * Removing `db_session.commit()` right after a call to and `create` call on our factory (which commits)
  * A few minor performance improvements to slow tests that had sleeps/other extra stuff that made them take several seconds (a lot more to do here, but got a few simple ones)

## Context for reviewers
Just routine upkeep - was looking at any performance improvements we could make as part of some of the data separation work, but there weren't many easy/obvious wins without quite a bit more refactor.

## Validation steps
Other than one comment change, this is all tests, the tests themselves say whether they still work.

### DIFF
--- a/api/tests/src/services/opportunities_v1/test_opportunity_version.py
+++ b/api/tests/src/services/opportunities_v1/test_opportunity_version.py
@@ -1,22 +1,12 @@
-import pytest
-
 from src.constants.lookup_constants import OpportunityCategory
-from src.db.models.agency_models import Agency
-from src.db.models.opportunity_models import Opportunity, OpportunityVersion
+from src.db.models.opportunity_models import OpportunityVersion
 from src.services.opportunities_v1.opportunity_version import save_opportunity_version
-from tests.lib.db_testing import cascade_delete_from_db_table
 from tests.src.db.models.factories import (
     AgencyFactory,
     OpportunityAssistanceListingFactory,
     OpportunityAttachmentFactory,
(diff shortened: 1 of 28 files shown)


api/tests/src/services/opportunities_v1/test_opportunity_version.py
```diff
@@ -62,7 +52,11 @@ def test_save_opportunity_version(db_session, enable_factory_create):
     save_opportunity_version(db_session, opp)
 
     # Verify Record created
-    saved_opp_version = db_session.query(OpportunityVersion).all()
+    saved_opp_version = (
+        db_session.query(OpportunityVersion)
+        .where(OpportunityVersion.opportunity_id == opp.opportunity_id)
+        .all()
+    )
```
kkrug, line 59: Do we want to change these from `db_session.query()` to `db_session.execute(select(...))` since we are making changes here anyway? There's a few instances in this file.

chouinar: I do want to do that, but.. `grep -r "db_session.query" tests | wc -l` -> `180` - there's a lot. I made a ticket for specifically that https://github.com/HHS/simpler-grants-gov/issues/10331

kkrug (approved):

PR #10328, opened 2026-05-08
[Issue #10327]: copy params script from dev/staging to grantee1/2 env

### LINKED ISSUE #10327: Create github action and script to replicate ssm params to grantee environments.
Create github actions script to copy ssm params from dev/staging to grantee1 and grantee2 environments.

prasnava:
## Summary

Created a new github actions and shell script to copy ssm params from dev/staging to grantee1 and grantee2 environments. 

## Validation steps

Ran the shell script locally against grantee1. The search and login are working in grantee1 environment. (The login is redirecting to staging, but that is a separate issue from this ticket.)


sean-navapbc (changes requested):
## PR Review: Copy SSM Parameters Script

### Overview
This PR adds a script and GitHub Actions workflow to copy SSM parameters from dev/staging to grantee1/grantee2 environments. While the approach is sound, there are **several critical issues** that need to be addressed.

---

### 🚨 Critical Issues

**1. AWS Credentials Configuration Bug**

`.github/workflows/copy-ssm-params.yml:42-44`:
```yaml
- name: Configure AWS credentials
  uses: ./.github/actions/configure-aws-credentials
  with:
    app_name: api
    environment: ${{ inputs.source_environment }}  # ❌ WRONG
```

**Problem**: The workflow configures AWS credentials for the **source** environment, but the script needs:
- **READ** access to the source environment
- **WRITE** access to the target environment

The script will fail when trying to write parameters to the target environment because it doesn't have credentials for it.

**Solution**: You need to configure credentials for the **target** environment (where you're writing to), since both environments are presumably in the same AWS account and the credentials will have read access across environments.

```yaml
- name: Configure AWS credentials
  uses: ./.github/actions/configure-aws-credentials
  with:
    app_name: api
    environment: ${{ inputs.target_environment }}  # ✅ CORRECT
```

Or verify both environments are in the same AWS account and the role has cross-environment access.

---

**2. No Backup Before Overwrite**

`.github/scripts/copy-ssm-params.sh:115-121`:
```bash
aws ssm put-parameter \
  --name "${TARGET_NAME}" \
  --value "${VALUE}" \
  --type "${TYPE}" \
  --tier "${TIER}" \
  --overwrite \  # ❌ Overwrites without backup
  --no-cli-pager > /dev/null
```

**Problem**: The script overwrites existing parameters without creating backups. If something goes wrong, you can't roll back.

**Recommendations**:
1. Add a backup step that exports existing target params before overwriting
2. Or check if the parameter exists and its value before overwriting
3. Or add a confirmation prompt showing what will be changed

---

**3. Error Handling Stops on First Failure**

`.github/scripts/copy-ssm-params.sh:11`:
```bash
set -euo pipefail  # ❌ Will exit on first error
```

**Problem**: If one parameter fails to copy (e.g., due to permissions or size limits), the entire script exits and remaining parameters aren't copied.

**Recommendation**: Track failures and continue processing:
```bash
FAILED_PARAMS=()

# In the loop:
if ! aws ssm put-parameter ...; then
  FAILED_PARAMS+=("${NAME}")
  echo "  ERROR: Failed to copy"
  continue
fi

# At the end:
if [[ ${#FAILED_PARAMS[@]} -gt 0 ]]; then
  echo "Failed to copy ${#FAILED_PARAMS[@]} parameter(s)"
  exit 1
fi
```

---

### ⚠️ Medium Priority Issues

**4. Hardcoded Skip List May Become Stale**

`.github/scripts/copy-ssm-params.sh:45-50`:
```bash
SKIP_PARAMS=(
  "frontend-login-redirect-url"
  "frontend-base-url"
  "api-url"
)
```

**Concern**: As the system evolves, more environment-specific parameters may be added. This hardcoded list could miss them.

**Recommendation**: 
- Document clearly in the code comments that this list must be updated when adding environment-specific params
- Or consider a naming convention (e.g., params ending in `-url` or starting with `env-specific-`)

---

**5. Scheduled Jobs Enabled Without Testing**

`infra/api/app-config/env-config/scheduled_jobs.tf`:

All 4 jobs are being enabled in grantee1/grantee2:
- feed-lookups
- sam-extracts  
- create-analytics-db-csvs
- email-notification-opportunity

**Question**: Have these jobs been tested in the grantee environments? Especially the email notifications - you don't want to accidentally send test emails to real users.

**Recommendation**: 
- If these are truly test/demo environments, consider keeping email notifications disabled
- Or verify email configs point to test addresses in grantee environments

---

**6. No Success Notification for Audit Trail**

`.github/workflows/copy-ssm-params.yml:55-58`:
```yaml
send-slack-notification:
  if: failure()  # ❌ Only notifies on failure
  needs: [copy-ssm-params]
```

**Recommendation**: Add success notification too for audit purposes.

---

### ✅ Good Practices

1. **Dry run mode** - Excellent safety feature
2. **Input validation** - Restricts source/target to specific environments
3. **GitHub step summary** - Good user experience
4. **Manual trigger only** - `workflow_dispatch` prevents accidental runs
5. **Environment protection** - Uses GitHub environment for target, enabling manual approvals
6. **Tier detection** - Automatically uses Advanced tier for large params
7. **Good logging** - Clear output showing what's being copied

---

### 📋 Testing Recommendations

Before merging:

1. **Test the dry-run mode** against real environments
2. **Verify AWS credentials** work for writing to target environment
3. **Test with a parameter that already exists** in the target to verify overwrite behavior
4. **Test with a SecureString parameter** to verify encryption works correctly
5. **Test failure scenario** - what happens if you don't have permission to write to target?
6. **Verify the skip list** - check that those 3 params actually exist and are environment-specific

---

### 🎯 Must-Fix Before Merge

1. ✅ Fix AWS credentials configuration (use target environment, not source)
2. ✅ Add error handling to continue on failure and report all failures at end
3. ✅ Add backup mechanism or at least document the risk

### Should-Fix Before Merge

4. ⚠️ Add success notifications for audit trail
5. ⚠️ Verify email notification jobs won't send real emails in grantee environments
6. ⚠️ Add comment documenting that skip list must be updated manually

---

### Final Verdict

**Request changes** - The AWS credentials bug is critical and will cause the workflow to fail. Once that's fixed and error handling is improved, this will be a useful addition to the infrastructure tooling.

sean-navapbc (approved):

PR #10332, opened 2026-05-08
[unticketed]: update login gov client id

prasnava:
## Summary

Updated grantee1 and 2 login.gov client id's to match staging. 

## Validation steps

Deployed and tested in grantee2. The login functionality is working in the grantee2 environment.


mdragon (approved):
👍🏻

Issue #10342, opened 2026-05-11
SOAP/Proxy: Correctly returned Legacy response


jakobpederson:
### Summary

The SOAP/Proxy is overriding the Legacy response with the Simpler response when it really should be returning the Legacy response. Add some logic to only call and return Legacy response if the original post is Legacy specific.

### Acceptance criteria

- [ ] SOAP/Proxy only calls Simpler or Legacy for cases where that's appropriate

PR #10343, opened 2026-05-11
[Issue #10342] Return Legacy response correctly

jakobpederson:
## Summary

Fixes / Work for #10342  

## Changes proposed

Added logic to either call just Simpler, Legacy, or both.

## Context for reviewers

The Use-Simpler-Override header was controlling which response was returned but going forward we will want to have that controlled by whether or not the request is Simpler or Legacy.

## Validation steps

- [ ] Hit `UpdateApplicationInfo` endpoint with valid Legacy tracking number and get correct success response

### DIFF
--- a/api/src/legacy_soap_api/simpler_soap_api.py
+++ b/api/src/legacy_soap_api/simpler_soap_api.py
@@ -39,6 +39,9 @@ from src.logging.flask_logger import add_extra_data_to_current_request_logs
 
 logger = logging.getLogger(__name__)
 
+GET_SUBMISSION_LIST_EXPANDED_REQUEST = "GetSubmissionListExpandedRequest"
+GET_OPPORTUNITY_LIST_REQUEST = "GetOpportunityListRequest"
+
 
 def get_simpler_soap_response(
     soap_request: SOAPRequest, soap_legacy_response: SOAPResponse, db_session: db.Session
@@ -139,7 +142,7 @@ def process_simpler_request(
     logger.info("SOAP request received")
 
     is_get_opportunity_list = (
-        operation_name == "GetOpportunityListRequest" and api_name == SimplerSoapAPI.APPLICANTS
+        operation_name == GET_OPPORTUNITY_LIST_REQUEST and api_name == SimplerSoapAPI.APPLICANTS
     )
     auth = None
     # GetOpportunityListRequest does not have any auth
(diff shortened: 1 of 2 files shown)


api/tests/src/legacy_soap_api/test_legacy_soap_api_routes.py
```diff
@@ -1059,3 +977,146 @@ def test_request_fails_if_legacy_certificate_is_expired(
...
+        <soapenv:Envelope xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/" xmlns:agen="http://apply.grants.gov/services/AgencyWebServices-V2.0" xmlns:gran="http://apply.grants.gov/system/GrantsCommonElements-V1.0">
+        <soapenv:Header/>
+        <soapenv:Body>
+        <agen:GetSubmissionListExpandedRequest>
+        </agen:GetSubmissionListExpandedRequest>
+        </soapenv:Body>
+        </soapenv:Envelope>
+    """
+    envelope = etree.fromstring(mock_data)
+    envelope = etree.fromstring(mock_data)
```
kkrug: Duplicate line

api/tests/src/legacy_soap_api/test_legacy_soap_api_routes.py
```diff
@@ -1059,3 +977,146 @@ def test_request_fails_if_legacy_certificate_is_expired(
...
+    agency = AgencyFactory.create()
+    privileges = {Privilege.LEGACY_AGENCY_GRANT_RETRIEVER}
+    _, _, _, mtls_cert = setup_cert_user(agency, privileges)
+    client.post(
+        full_path,
+        data=etree.tostring(envelope),
+        headers={MTLS_CERT_HEADER_KEY: mtls_cert},
+    )
+    mock_get_soap_response.assert_called_once()
+    mock_get_simpler_soap_response.assert_called_once()
```
kkrug: I think we should assert the status code == 200 here as well.

api/tests/src/legacy_soap_api/test_legacy_soap_api_routes.py
```diff
@@ -1059,3 +977,146 @@ def test_request_fails_if_legacy_certificate_is_expired(
...
+    mock_get_soap_response.assert_called_once()
+    mock_get_simpler_soap_response.assert_not_called()
+
+
+@mock.patch("src.legacy_soap_api.legacy_soap_api_proxy._get_soap_response")
+@mock.patch("src.legacy_soap_api.simpler_soap_api.get_simpler_soap_response")
+def test_get_submission_list_expanded_always_calls_legacy_and_simpler(
+    mock_get_simpler_soap_response, mock_get_soap_response, client, enable_factory_create
+) -> None:
+    full_path = "/grantsws-applicant/services/v2/ApplicantWebServicesSoapPort"
```
kkrug: Should this be:
```suggestion
    full_path = "/grantsws-agency/services/v2/AgencyWebServicesSoapPort"
```

api/src/legacy_soap_api/simpler_soap_api.py
```diff
@@ -176,19 +179,25 @@ def process_simpler_request(
...
         elif alternate_legacy_response := get_alternate_legacy_response(soap_request):
             logger.info(
                 "simpler_soap_api: skipping legacy call",
             )
             soap_legacy_response = alternate_legacy_response
-        # Fallback: get legacy response
-        else:
+        # GetSubmissionListExpanded will call both if use_simpler is true
+        # handled in the get_simpler_response
+        elif operation_name == GET_SUBMISSION_LIST_EXPANDED_REQUEST:
```
kkrug: Would it make sense to also check `api_name == SimplerSoapAPI.GRANTORS` here like we do for `is_get_opportunity_list` or is that redundant?

api/tests/src/legacy_soap_api/test_legacy_soap_api_routes.py
jakobpederson: Yeah good catch!

api/src/legacy_soap_api/simpler_soap_api.py
jakobpederson: It might be redundant, but I don't have a problem being explicit and consistent.

kkrug (approved):

PR #10350, opened 2026-05-11
[Issue #10142] Modify transform opportunity to set agency_id

### LINKED ISSUE #10142: Modify Transform Logic to Fill Agency IDs from Grants.Gov
### Summary

Corresponding ticket to #9538 to modify the transform logic to fill in `agency_id` in the opportunity table for the opportunities coming from grants.gov

We transform opportunities in https://github.com/HHS/simpler-grants-gov/blob/main/api/src/data_migration/transformation/subtask/transform_opportunity.py 

and create them in this function:
https://github.com/HHS/simpler-grants-gov/blob/main/api/src/data_migration/transformation/transform_util.py#L125

In order to connect the agency ID, we'll need to:
* Fetch the agency from our agency table based on the agency code (it's fine if this isn't a part of the big query, not worried about performance since that only matters for a first time load)
* If the agency doesn't exist then we won't add it (this'll be covered by the below logic).

Separately, we'll want to create a new automation step in our job that handles backfilling agencies.

### Acceptance criteria

- [ ] New opportunities from grants.gov will have the `agency_id` field in the opportunity table be filled with the corresponding correct agency ID based on the agency code/Agency table.

chouinar:
## Summary
Fixes  #10142

## Changes proposed
* Adjust the transform_opportunity logic to set the agency ID on the opportunity when transforming
* Add a new task that handles backfilling anything that isn't already set

## Context for reviewers
We're halfway through pivoting from our current approach of managing the agency foreign key to the agency table (joining on the `agency_code` field in both tables) to a proper foreign key using `agency_id`. This'll make it so that opportunities we pull from grants.gov attempt to set the `agency_id` when we transform, as well as handle backfilling anything that might have been missed.

I'm not 100% certain, but there may be some timing edge cases where we could have the transform bit not properly link everything, so I'm erring on the side of being cautious and we'll run the backfill every hour. It only will pickup opportunities that can be connected to an agency that aren't already connected. So, if after a few weeks, the metric is always 0 processed, we can clean it up.

## Validation steps
These are mainly tested via our unit tests. I did do a quick run locally of the load-transform jobs with 10k opportunities that had no agency ID, and it took only ~1 second to run, so even though non-locally it'll be a bit slower, I think that's on the order of 1-2 minutes.

### DIFF
--- a/api/src/data_migration/transformation/subtask/transform_opportunity.py
+++ b/api/src/data_migration/transformation/subtask/transform_opportunity.py
@@ -121,16 +121,16 @@ class TransformOpportunity(AbstractTransformSubTask):
                         "agency_id": agency.agency_id,
                     },
                 )
-                transformed_opportunity.agency_id = agency.agency_id
                 self.increment(self.Metrics.OPPORTUNITY_TRANSFORMED_HAS_AGENCY)
+                transformed_opportunity.agency_id = agency.agency_id
             else:
-                transformed_opportunity.agency_id = None
                 logger.info(
-                    "Attaching agency to opportunity",
+                    "No agency found to attach to opportunity",
                     extra=extra
                     | {"agency_code": transformed_opportunity.agency_code, "agency_id": None},
                 )
                 self.increment(self.Metrics.OPPORTUNITY_TRANSFORMED_NULL_AGENCY)
+                transformed_opportunity.agency_id = None
 
             if is_insert:
                 self.increment(
--- a/api/src/data_migration/transformation/subtask/transform_opportunity.py
+++ b/api/src/data_migration/transformation/subtask/transform_opportunity.py
@@ -1,5 +1,8 @@
 import logging
+from enum import StrEnum
 from typing import cast
+
+from sqlalchemy import select
 
 import src.data_migration.transformation.transform_constants as transform_constants
 import src.data_migration.transformation.transform_util as transform_util
@@ -7,6 +10,7 @@ from src.adapters.aws import S3Config
 from src.data_migration.transformation.subtask.abstract_transform_subtask import (
     AbstractTransformSubTask,
 )
+from src.db.models.agency_models import Agency
 from src.db.models.competition_models import Competition
 from src.db.models.opportunity_models import Opportunity, OpportunityAttachment
 from src.db.models.staging.opportunity import Topportunity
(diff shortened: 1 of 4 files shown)


api/src/data_migration/transformation/subtask/transform_opportunity.py
```diff
@@ -100,6 +110,28 @@ def process_opportunity(
...
+                    },
+                )
+                transformed_opportunity.agency_id = agency.agency_id
+                self.increment(self.Metrics.OPPORTUNITY_TRANSFORMED_HAS_AGENCY)
+            else:
+                transformed_opportunity.agency_id = None
+                logger.info(
+                    "Attaching agency to opportunity",
+                    extra=extra
+                    | {"agency_code": transformed_opportunity.agency_code, "agency_id": None},
```
kkrug, line 130: Would we rather use two distinct log messages here?

Also more of a nit but could we have the `transformed_opportunity.agency_id =...` and `self.increment(...` lines either both be before or both after the log messages for more consistency?

chouinar: Whoops, meant to have those be distinct message - good catch

api/src/data_migration/transformation/subtask/transform_opportunity.py
```diff
@@ -172,3 +204,57 @@ def _move_competition_instructions_to_correct_bucket(self, competition: Competit
...
+
+    def fetch_opportunities_and_agencies(self) -> list[tuple[Opportunity, Agency]]:
+        """Fetch all opportunities without an agency_id set that could have it set"""
+        return cast(
+            list[tuple[Opportunity, Agency]],
+            self.db_session.execute(
+                select(Opportunity, Agency)
+                # This join means we'll only fetch records where
+                # the agency exists for the opportunity that we'll be updating.
+                .join(Agency, Opportunity.agency_code == Agency.agency_code)
```
hao10282025-sudo, line 256: quick check is the Opportunity.agency_code or Agency.agency_code always NOT None for the join?
not see it from the model files

chouinar: The agency code can be null in the opportunity table, although very rarely (generally means bad data from grants.gov). The agency code in the agency table can't be null.

This is all defined on the model files - the python typing is used to determine nullability. `agency_code: Mapped[str]` isn't nullable. `agency_code: Mapped[str | None]` is.

hao10282025-sudo: Ah right, all good then, thanks for the context

kkrug (approved):

Issue #10242, opened 2026-05-07
Don't run scheduled Actions on forks


mdragon:
### Summary

As we get more folks forking the repo we've noticed that many of our scheduled Actions only make sense to run in the context of the original repository and either fail or just aren't actionable/useful on a fork. Figure out how to prevent the jobs from running on a fork or short circuit the actual task and exit successfully immediately.

### Acceptance criteria

- [ ] Identify how to best prevent scheduled Actions from running at all (ideally) or at least exit quickly without failing (fallback option).
- [ ] Implement that solution on scheduled jobs
- [ ] Document what we're doing and how to apply it to future scheduled jobs

PR #10365, opened 2026-05-12
[Issue #10242] One stop action to disable scheduled Actions in forks

mdragon:
## Summary

Work for #10242 

## Changes proposed

Create a single Action that will disable all the scheduled jobs that don't make sense to waste cycles executing in forks.

## Context for reviewers

Now that we have a lot of scheduled Actions, folks who fork our repo run those Actions and many fail because they don't have AWS connectivity or API Keys for other systems, etc. Since GH doesn't support this directly, the easiest solution seems to be a single job that will disable all the other jobs. Any new scheduled jobs can be added to the list and will be disabled automatically in any forks that update, thereby picking up the new job, and the update the disabler job.

## Validation steps

- ✅ Tested that this job doesn't run the disable commands in the mainline repo
- After merge will test that the jobs are disabled successfully in a fork

### DIFF
--- a/.github/workflows/fork-disable-unneeded-actions.yml
+++ b/.github/workflows/fork-disable-unneeded-actions.yml
@@ -1,0 +1,38 @@
+name: Fork - Disable Unneeded Actions
+env:
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
+
+on:
+  workflow_dispatch: # for manual runs
+  schedule:
+    # Runs "at 06:00 UTC, every week on Monday" (see https://crontab.guru)
+    # Sprints are 2 weeks long, but crons don't easily support biweekly schedules
+    # and this GitHub action won't move tickets out of the current sprint
+    - cron: "0 6 * * 1"
+
+jobs:
+  disable-scheduled:
+    name: Disable scheduled jobs
+    runs-on: ubuntu-latest
+    steps:
+      - id: disable-scheduled-step
+        # only run this on forks
+        if: ${{ github.repository != 'hhs/simpler-grants-gov' }}
+        run: |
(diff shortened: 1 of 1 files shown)


prasnava (approved):
👍

.github/workflows/fork-disable-unneeded-actions.yml
```diff
@@ -0,0 +1,47 @@
...
+      - id: disable-scheduled-step
+        name: Disable jobs not needed in the fork
+        # only run this on forks
+        if: ${{ github.repository != 'hhs/simpler-grants-gov' }}
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          # NOTE: Add to this list in alphabetical order, if more jobs shouldn't run on forks
+          gh workflow disable "Check Infra deploy status"
+          gh workflow disable "CI Cron Vulnerability Scans"
```
chouinar, line 33: Any way we could have a secret/env var configured that we use to enable/disable things?

Something like (psuedo-code):

```py
if secrets.get("ENABLE_SCHEDULED_GH_ACTION", False):
     # skip the workflow
# proceed
```

Then we enable it, but it wouldn't be enabled by default in forks.

Alternatively, any way to add a tag to an action to fetch which ones we disable here so it's easier to maintain?

mdragon: Unfortunately no way to exit early with success. So we'd have to go to every step and make it contain logic to skip on the fork side and run on our side.

I think this will work the way you're thinking for the second idea though. Our "disabled job list" lives in the repo. So whatever jobs exist in that version of the repo, we have added to this action at that same version of the repo. If we add a job and we was the disable, they'll reach the fork together. The one hole here is if we forget to add it and they update their fork after the job exists but before we set it to get disabled.

Issue #10347, opened 2026-05-11
SOAP/Proxy: Add a variable to turn on/off SOAP simpler route and add header to override that variable


jakobpederson:
### Summary

This is similar to the existing `Use-Soap-Override` but the use case is a little different. For one, we want an emergency kill switch for the Simpler route in case the wrong data starts going out, and then we want to be able to flip it per request for testing purposes to track down a fix. 

So something like `ENABLE_SIMPLER_ROUTE`.

### Acceptance criteria

- [ ] Can turn on/off Simpler route
- [ ] Add environment variable to AWS Parameter Store

PR #10370, opened 2026-05-12
[Issue #10347] Add ENABLE_SIMPLER_ROUTE key and env var

jakobpederson:
## Summary

Fixes / Work for #10347  

## Changes proposed

-`ENABLE_SIMPLER_ROUTE` env var 
- '"Enable-Simpler-Override"' header in order to override `ENABLE_SIMPLER_ROUTE`
- tests

## Context for reviewers

We want the ability to deactive Simpler route and force just the Legacy response (`ENABLE_SIMPLER_ROUTE`) and a header var that will override that when we want to do testing on individual requests ('"Enable-Simpler-Override"' ).

## Validation steps

-  [ ]  Can deactivate the Simpler route
- [ ] Can use the header to enable Simpler route on a request

### DIFF
--- a/api/src/legacy_soap_api/legacy_soap_api_config.py
+++ b/api/src/legacy_soap_api/legacy_soap_api_config.py
@@ -33,7 +33,7 @@ class LegacySoapAPIConfig(PydanticBaseEnvConfig):
     soap_partner_gateway_auth_key: str = Field("", alias="SOAP_PARTNER_GATEWAY_AUTH_KEY")
     soap_grantors_path: str = Field("", alias="SOAP_GRANTORS_PATH")
     soap_applicants_path: str = Field("", alias="SOAP_APPLICANTS_PATH")
-    enable_simpler_route: bool = Field(False, alias="ENABLE_SIMPLER_ROUTE")
+    enable_simpler_route: bool = Field(True, alias="ENABLE_SIMPLER_ROUTE")
 
     @property
     def gg_url(self) -> str:
--- a/api/src/legacy_soap_api/legacy_soap_api_config.py
+++ b/api/src/legacy_soap_api/legacy_soap_api_config.py
@@ -33,6 +33,7 @@ class LegacySoapAPIConfig(PydanticBaseEnvConfig):
     soap_partner_gateway_auth_key: str = Field("", alias="SOAP_PARTNER_GATEWAY_AUTH_KEY")
     soap_grantors_path: str = Field("", alias="SOAP_GRANTORS_PATH")
     soap_applicants_path: str = Field("", alias="SOAP_APPLICANTS_PATH")
+    enable_simpler_route: bool = Field(False, alias="ENABLE_SIMPLER_ROUTE")
 
     @property
     def gg_url(self) -> str:
--- a/api/src/legacy_soap_api/legacy_soap_api_constants.py
+++ b/api/src/legacy_soap_api/legacy_soap_api_constants.py
@@ -37,4 +37,4 @@ class LegacySoapApiEvent(StrEnum):
 
     RETURNING_SIMPLER_RESPONSE = "returning_simpler_response"
     RETURNING_LEGACY_SOAP_RESPONSE = "returning_legacy_soap_response"
-    SIMPLER_ROUTE_DISABLED = "simpler-route-disabled"
+    SIMPLER_ROUTE_DISABLED = "simpler_route_disabled"
--- a/api/src/legacy_soap_api/legacy_soap_api_constants.py
+++ b/api/src/legacy_soap_api/legacy_soap_api_constants.py
@@ -37,3 +37,4 @@ class LegacySoapApiEvent(StrEnum):
 
     RETURNING_SIMPLER_RESPONSE = "returning_simpler_response"
     RETURNING_LEGACY_SOAP_RESPONSE = "returning_legacy_soap_response"
+    SIMPLER_ROUTE_DISABLED = "simpler-route-disabled"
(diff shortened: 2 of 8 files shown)


api/src/legacy_soap_api/legacy_soap_api_config.py
```diff
@@ -33,6 +33,7 @@ class LegacySoapAPIConfig(PydanticBaseEnvConfig):
     soap_partner_gateway_auth_key: str = Field("", alias="SOAP_PARTNER_GATEWAY_AUTH_KEY")
     soap_grantors_path: str = Field("", alias="SOAP_GRANTORS_PATH")
     soap_applicants_path: str = Field("", alias="SOAP_APPLICANTS_PATH")
+    enable_simpler_route: bool = Field(False, alias="ENABLE_SIMPLER_ROUTE")
```
kkrug: I think the default should be `True`. If it's `False`, when this code is deployed before the SSM param is created, it will immediately disable all simpler route processing (i.e. it's backwards incompatible).

api/tests/src/legacy_soap_api/test_legacy_soap_api_routes.py
```diff
@@ -1120,3 +1121,124 @@ def test_calls_simpler_if_using_simpler_tracking_number(
...
+                full_path,
+                data=mock_data,
+                headers={MTLS_CERT_HEADER_KEY: mtls_cert},
+            )
+    assert response.status_code == 200
+    mock_legacy_response.assert_called_once()
+    mock_simpler_response.assert_called_once()
+
+
+def test_x(db_session, client, enable_factory_create, monkeypatch) -> None:
```
kkrug: Can we rename this to something more meaningful, something like:
```suggestion
def test_enable_simpler_route_header_overrides_disabled_config db_session, client, enable_factory_create, monkeypatch) -> None:
```

api/tests/src/legacy_soap_api/test_legacy_soap_api_routes.py
```diff
@@ -1120,3 +1121,124 @@ def test_calls_simpler_if_using_simpler_tracking_number(
     assert response.status_code == 200
     mock_get_soap_response.assert_not_called()
     mock_get_simpler_soap_response.assert_called_once()
+
+
+def test_enable_simpler_routes_set_to_false_stops_simpler_calls(
```
kkrug: All three new tests reuse quite a bit of setup code, it may be worth breaking it out into a helper.

api/src/legacy_soap_api/simpler_soap_api.py
```diff
@@ -180,6 +181,12 @@ def process_simpler_request(
             auth and auth.certificate and not auth.certificate.legacy_certificate
         )
 
+        if (
+            not get_soap_config().enable_simpler_route
+            and request.headers.get(ENABLE_SIMPLER_ROUTE_KEY, None) != "1"
+        ):
+            return get_legacy_response(soap_request).to_flask_response()
+
         # If it is GetOpportunityList or is valid legacy certificate but not configured in Simpler
```
kkrug, line 194: Can we add logging here for consistency? I didn't show it here, but you might also consider a new constant instead of reusing `RETURNING_LEGACY_SOAP_RESPONSE`, something like `SIMPLER_ROUTE_DISABLED = "simpler_route_disabled"`.
```suggestion
        if (
            not get_soap_config().enable_simpler_route
            and request.headers.get(ENABLE_SIMPLER_ROUTE_KEY, None) != "1"
        ):
            logger.info(
                "simpler_soap_api: simpler route disabled, returning legacy response",
                extra={"soap_api_event": LegacySoapApiEvent.RETURNING_LEGACY_SOAP_RESPONSE},
            )
            return get_legacy_response(soap_request).to_flask_response()

# If it is GetOpportunityList or is valid legacy certificate but not configured in Simpler
```

api/src/legacy_soap_api/legacy_soap_api_config.py
chouinar: If we're using an SSM param, we need to have those defined before this can be merged regardless. Deploys will fail if we try to fetch an undefined SSM param. But I'd probably default to True regardless

api/tests/src/legacy_soap_api/test_legacy_soap_api_routes.py
jakobpederson: Nice catch, I would never want a test named `test_x`.

api/src/legacy_soap_api/simpler_soap_api.py
jakobpederson: Logging is a good idea, I'll ad it

api/src/legacy_soap_api/legacy_soap_api_constants.py
```diff
@@ -37,3 +37,4 @@ class LegacySoapApiEvent(StrEnum):
 
     RETURNING_SIMPLER_RESPONSE = "returning_simpler_response"
     RETURNING_LEGACY_SOAP_RESPONSE = "returning_legacy_soap_response"
+    SIMPLER_ROUTE_DISABLED = "simpler-route-disabled"
```
kkrug: I know `simpler-route-disabled` was my suggestion but let's use underscores for naming consistency. 
```suggestion
    SIMPLER_ROUTE_DISABLED = "simpler_route_disabled"
```

jakobpederson: That one was my mistake.

kkrug (approved):

PR #10385, opened 2026-05-12
[Issue #10242] Fixes from fork testing

mdragon:
## Summary

Work for #10242 

## Changes proposed

Fix a typo.
Handle that disabling an already disabled Action will exit with error so we have to protect against that.

## Context for reviewers

Testing in the fork found a few issues we'll want to correct.

## Validation steps

- Will confirm in fork after merge to main.


prasnava (approved):
👍

PR #10405, opened 2026-05-13
[Issue #10403] Only install webkit dependencies in the Webkit Shard, Shard 3

### LINKED ISSUE #10403: Only install Webkit in shard 3 of E2E tests
### Summary

Webkit requires an extra install because it doesn't install to a nice clean, cacheable location. Right now we're installing Webkit dependencies in every shard, tweak this to only install it in the Webkit shard, shard 3.

### Acceptance criteria

- Scans pass
- Shards other than 3 don't install Webkit 
- Shard 3 installs Webkit

mdragon:
## Summary

Fixes #10403  

## Changes proposed

Vary the require webkit install input when calling the E2E action based on whether we're running shard 3, the WebKit shard

## Context for reviewers

Webkit doesn't cleanly contain itself and it's dependencies when installing through Playwright like the other browsers do. As a result we have to re-install the playwright deps every time we run E2E. This change limits that install cost to just shard 3, as that's the only one running Webkit.

## Validation steps

- Checks pass

### DIFF
--- a/.github/actions/e2e/action.yml
+++ b/.github/actions/e2e/action.yml
@@ -91,6 +91,7 @@ runs:
         npm install @playwright/test
         npx playwright install --with-deps
 
+    # this should only be needed on shard 3, the WebKit shard
     - name: Install Webkit dependencies
       if: ${{ inputs.force-webkit-install == 'true' }}
       shell: bash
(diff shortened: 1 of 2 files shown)


Bhavna-Ramachandran:
FYI - staging also downloads the WebKit binary on every run via `npx playwright install --with-deps` in the action, but that's expected since `e2e-staging.yml` has no playwright cache job unlike `ci-frontend-e2e.yml`. Irrelevant for this PR, just adding a note here - if we want to optimize staging runs in the future, particularly if we expand it to run browser-specific shards.

cc: @doug-s-nava

Bhavna-Ramachandran (approved):

Issue #9863, opened 2026-04-27
Create an endpoint for updating the status of a pending file scan


chouinar:
### Summary

We want to create `POST /v1/files/:file_id` that will take in a scan status and update the `pending_file` table for that record.

#### Request

```py
{
  "file_scan_status": "success", # The file scan status enum
}
```

#### Response

A generic, empty response. No data needs to be sent back beyond the status code.

#### Auth
This endpoint _only_ allows API User Key Auth (Not JWT auth).

This endpoint requires a user have the `internal_s3_scan` privilege which is an internal one that only the lambda will have.

#### Logic

* Verify the calling user has the `internal_s3_scan` privilege (403 if not)
* Fetch the `pending_file` from the DB (404 if not found)
* Update the file scan status of the pending file record (don't need to bother checking if they match already)
* Make sure to log the status / user id / pending file ID
* Log the duration that this took. That should be the current UTC time minus the `created_at` timestamp of the pending record.

### Acceptance criteria

- [x] Pending file update endpoint created
- [x] Tests added

PR #10418, opened 2026-05-14
[Issue #9863] Create an endpoint for updating the status of a pending file scan

kkrug:
## Summary

Fixes for #9863  

## Changes proposed

- Created the api/src/api/files_v1 directory to house the newly created:
  - init.py
  - file_blueprint.py
  - file_routes.py
  - file_schemas.py - currently has `FileScanStatusUpdateRequestSchema` and `FileScanStatusUpdateResponseSchema`
- Created update_pending_file_scan_status.py for the service layer to verify the `internal_s3_scan` privilege, fetch the pending file, update the scan status, and log the transaction
- Created test_update_pending_file_scan_status.py and the associated init.py to test the new endpoint
- Updated app.py to add the `enable_file_upload_endpoints` feature flag and import/register `files_v1_blueprint`
- Updated local.env to add `ENABLE_FILE_UPLOAD_ENDPOINTS=TRUE`

## Context for reviewers

We want to create `POST /v1/files/:file_id` that will take in a scan status and update the `pending_file table` for that record.

## Validation steps

Confirm tests pass.


api/src/services/files/update_pending_file_scan_status.py
```diff
@@ -0,0 +1,44 @@
...
+
+    stmt = select(PendingFile).where(PendingFile.pending_file_id == pending_file_id)
+    pending_file = db_session.execute(stmt).scalar_one_or_none()
+
+    if pending_file is None:
+        raise_flask_error(404, "Pending file not found")
+
+    pending_file.file_scan_status = file_scan_status
+
+    now = datetime.now(timezone.utc)
```
chouinar: Same functionality, just cleaner / makes sure we're consistent when Python changes the proper way to do datetimes again.

```suggestion
    now = datetime_util.utcnow()
```

api/src/services/files/update_pending_file_scan_status.py
```diff
@@ -0,0 +1,44 @@
...
+    scan_duration_seconds = (now - pending_file.created_at).total_seconds()
+
+    logger.info(
+        "Updated pending file scan status",
+        extra={
+            "pending_file_id": pending_file.pending_file_id,
+            "user_id": user.user_id,
+            "file_scan_status": file_scan_status.value,
```
chouinar:
```suggestion
            "file_scan_status": file_scan_status,
```

api/tests/src/api/files_v1/test_update_pending_file_scan_status.py
```diff
@@ -0,0 +1,160 @@
...
+    factories.LinkRoleRoleTypeFactory.create(role=role, role_type=RoleType.INTERNAL)
+    factories.InternalUserRoleFactory.create(user=user, role=role)
+
+    return user
+
+
+class TestUpdateFileScanStatusSuccess:
+    def test_update_scan_status_complete(self, client, db_session, enable_factory_create):
+        api_key = str(uuid.uuid4())
+        _setup_s3_scan_user(db_session, api_key)
```
chouinar: Make the user / api key creation a fixture. It's fine if the factory generates the key for you

```suggestion

class TestUpdateFileScanStatusSuccess(BaseTestClass):
    
    @pytest.fixture
    def s3_scan_user_key(db_session, enable_factory_create):
        api_key = factories.UserApiKeyFactory.create(user=user)

        role = factories.RoleFactory.create(
            privileges=[Privilege.INTERNAL_S3_SCAN]
        )
        factories.LinkRoleRoleTypeFactory.create(role=role, role_type=RoleType.INTERNAL)
        factories.InternalUserRoleFactory.create(user=api_key.user, role=role)
             
        return api_key.key_id

    def test_update_scan_status_complete(self, client, db_session, s3_scan_user_key):
        ...
```

api/tests/src/api/files_v1/test_update_pending_file_scan_status.py
```diff
@@ -0,0 +1,160 @@
...
+            _build_url(pending_file.pending_file_id),
+            headers={"X-API-Key": api_key},
+            json={"file_scan_status": "complete"},
+        )
+
+        assert resp.status_code == 200
+
+        db_session.expire_all()
+        updated = db_session.get(PendingFile, pending_file.pending_file_id)
+        assert updated.file_scan_status == FileScanStatus.COMPLETE
```
chouinar: If you've already got the DB object, but you want to get it fresh from the DB, rather than expire + query, can just refresh it:
```suggestion
        db_session.refresh(pending_file)
        assert updated.file_scan_status == FileScanStatus.COMPLETE
```

Under the hood I think it does an expire + get.

api/src/api/files_v1/file_schemas.py
```diff
@@ -0,0 +1,18 @@
...
+from src.constants.lookup_constants import FileScanStatus
+
+
+class FileScanStatusUpdateRequestSchema(Schema):
+    file_scan_status = fields.Enum(
+        FileScanStatus,
+        required=True,
+        metadata={
+            "description": "The scan status to set on the pending file",
+            "example": "complete",
```
chouinar: With enums, we can exclude the example as it'll pick a value from the enum, or use the enum to define a specific one. That way if we ever change enum values, we won't miss fixing it here (really niche concern).

```suggestion
            "example": FileScanStatus.COMPLETE,
```

api/src/services/files/update_pending_file_scan_status.py
kkrug: Updated

api/src/services/files/update_pending_file_scan_status.py
kkrug: Updated

api/tests/src/api/files_v1/test_update_pending_file_scan_status.py
kkrug: Updated

api/tests/src/api/files_v1/test_update_pending_file_scan_status.py
kkrug: Updated

api/src/api/files_v1/file_schemas.py
kkrug: Updated

chouinar (approved):

PR #10444, opened 2026-05-15
[Unticketed] Modify the opportunity notification code to batch queries

chouinar:
## Changes proposed
* Change the opportunity notification query that finds prior notification versions to be batched

## Context for reviewers
With an unrelated backfill of data, we created a version of every opportunity in our system, which made the query size massive. The query that happens in the `_get_last_notified_versions` function is incredibly complex and does a lot of joins that have a lot of fields in them. With that large backfill, we're hitting a limit on Postgres of 65k fields defined in the query

```
psycopg.OperationalError) sending query and params failed: number of parameters must be between 0 and 65535
```

While in the long-term, we should consider simplifying the query (pulling logic out into our python code), that's a bigger refactor than we want to do when fixing a bug. This instead just batches that query into chunks.

## Validation steps
I loaded 50k opportunities, and users with those opportunities saved into my local DB using `make console` and our factories. Without my change, I get the same error we're seeing in prod, with the change, we're able to get past that query and fetch the data as appropriate. Batch sizes of 5000 or more still hit the error, so don't think we'd want to increase it further. Also tested with a batch size of 1 and the unit tests to verify nothing breaks with the batching logic (since 1000 is way more than the tests would ever setup)

### DIFF
--- a/api/src/task/notifications/opportunity_notifcation.py
+++ b/api/src/task/notifications/opportunity_notifcation.py
@@ -1,5 +1,6 @@
+import itertools
 import logging
-from collections.abc import Sequence
+from collections.abc import Iterable, Sequence
 from datetime import datetime
 from typing import cast
 from uuid import UUID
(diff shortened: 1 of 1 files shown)


mdragon (approved):
👍🏻

PR #10446, opened 2026-05-15
[Issue #5152] Move newsletter subscription to API route for improved test stability

### LINKED ISSUE #5152: Successfully mock external requests for e2e testing
### Summary

We have some tests for newsletter subscription actions that are attempting to mock out fetch calls, but it seems like the calls are not actually being fetched

(for more, see https://github.com/vercel/next.js/blob/canary/packages/next/src/experimental/testmode/playwright/README.md - since our code is attempting to mock calls to /subscribe, this won't work given the method we're using)

I've worked a bit on this and I think what we should be doing is:

* using built in playwright route mocking utilities (see https://playwright.dev/docs/api/class-route)
* mocking the call to the sendy service rather than the call to our Next API endpoint. This will involve somewhow telling playwright about the URL for the sendy service, which may be somewhat difficult. If we are ok with exposing this url publicly, we can just put it in an env var on the call during CI, maybe in the package.json script so that it can be used locally, or in a dedicated .env file we can read in with dotenv. If it cannot be public, we will need to add a call to AWS SSM into our E2E CI process to grab the value

Of course if that's all too complex we could just remove the test

### Acceptance criteria

- [ ] e2e tests pass reliably locally and in ci (this leave the door open to a decision to remove the tests in question)

Bhavna-Ramachandran:
## Summary
Work for : Task #5152 

## Changes proposed
- Newsletter subscription currently uses a Server Action, but E2E tests are unreliable because Next.js experimental test mode cannot consistently mock the server-side Sendy request in CI. This PR introduces an API route so the browser makes a real HTTP request that Playwright can reliably intercept in all environments.

## Context for reviewers
- To improve test stability, we are introducing a standard API route (/api/newsletter/subscribe)
- This change does not affect user behavior or Sendy integration, but improves test reliability and removes dependency on experimental Next.js testing features.

## Validation steps
- All tests pass locally and in CI after these changes

### DIFF
--- a/frontend/src/app/[locale]/(base)/newsletter/_components/SubscriptionForm.test.tsx
+++ b/frontend/src/app/[locale]/(base)/newsletter/_components/SubscriptionForm.test.tsx
@@ -41,10 +41,12 @@ describe("SubscriptionForm", () => {
 
   it("calls /api/newsletter/subscribe on submit", async () => {
     global.fetch = jest.fn(() =>
-      Promise.resolve({
-        json: () => Promise.resolve({ success: true }),
-      }),
-    );
+      Promise.resolve(
+        ({
+          json: () => Promise.resolve({ success: true }),
+        } as unknown) as Response,
+      ),
+    ) as unknown as typeof global.fetch;
 
     render(<SubscriptionForm />);
 
@@ -68,10 +70,12 @@ describe("SubscriptionForm", () => {
 
   it("redirects to confirmation page on success", async () => {
     global.fetch = jest.fn(() =>
-      Promise.resolve({
-        json: () => Promise.resolve({ success: true }),
-      }),
-    );
+      Promise.resolve(
+        ({
+          json: () => Promise.resolve({ success: true }),
+        } as unknown) as Response,
+      ),
+    ) as unknown as typeof global.fetch;
 
     render(<SubscriptionForm />);
 
@@ -92,10 +96,12 @@ describe("SubscriptionForm", () => {
 
   it("shows server error message on failure", async () => {
     global.fetch = jest.fn(() =>
-      Promise.resolve({
-        json: () => Promise.resolve({ success: false, errorCode: "server" }),
-      }),
-    );
+      Promise.resolve(
+        ({
+          json: () => Promise.resolve({ success: false, errorCode: "server" }),
+        } as unknown) as Response,
+      ),
+    ) as unknown as typeof global.fetch;
 
     render(<SubscriptionForm />);
 
--- a/frontend/src/app/[locale]/(base)/newsletter/_components/SubscriptionForm.test.tsx
+++ b/frontend/src/app/[locale]/(base)/newsletter/_components/SubscriptionForm.test.tsx
@@ -1,4 +1,16 @@
-import { act, render, screen, waitFor } from "@testing-library/react";
+import { act, render, screen, waitFor, fireEvent } from "@testing-library/react";
+
+// Ensure the client fetch used by the component delegates to `global.fetch`,
+// which the tests mock below. Must mock before importing the component.
+jest.mock("src/hooks/useClientFetch", () => ({
+  useClientFetch: () => ({
+    clientFetch: async (url: string, options?: RequestInit) => {
+      const res = await (global.fetch as unknown as jest.Mock)(url, options);
+      return res.json();
+    },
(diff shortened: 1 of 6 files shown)


frontend/src/app/[locale]/(base)/newsletter/_components/SubscriptionForm.tsx
```diff
@@ -34,8 +39,57 @@ export default function SubscriptionForm() {
...
   };
 
+  const handleSubmit = async (event: {
+    preventDefault(): void;
+    currentTarget: HTMLFormElement;
+  }) => {
+    event.preventDefault();
+    const formData = new FormData(event.currentTarget);
+
+    const response = await fetch("/api/newsletter/subscribe", {
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/newsletter/_components/SubscriptionForm.tsx]: can we use `useClientFetch` here?

frontend/src/app/[locale]/(base)/newsletter/_components/SubscriptionForm.tsx
```diff
@@ -18,10 +17,16 @@ export type ValidationErrors = {
...
+  errorMessage: ReactNode;
+  validationErrors: ValidationErrors;
+};
+
 export default function SubscriptionForm() {
   const t = useTranslations("Subscribe");
+  const router = useRouter();
 
-  const [state, formAction] = useActionState(subscribeEmail, {
+  const [state, setState] = useState<FormState>({
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/newsletter/_components/SubscriptionForm.tsx]: to make this more clear, can we call this `subscribeErrorState`?

frontend/src/app/api/newsletter/subscribe/route.ts
```diff
@@ -0,0 +1,84 @@
+import { environment } from "src/constants/environments";
```
doug-s-nava, line 1: can we add a unit test for this? If one exists for the previous server action we can reuse that. If not, and you feel like it's out of scope, let's create a new ticket for follow up

Bhavna-Ramachandran: I checked the repo and did not find any existing test for server action, but I found this - we can reuse the pattern from this route test /workspaces/simpler-grants-gov/frontend/src/app/api/auth/login/route.test.ts

I can do this as a follow up task (you can create a ticket and assign it to me). Let me know your thoughts.

frontend/src/app/[locale]/(base)/newsletter/_components/SubscriptionForm.tsx
Comment (Bhavna-Ramachandran) [reply frontend/src/app/[locale]/(base)/newsletter/_components/SubscriptionForm.tsx]: Updates made.

frontend/src/app/[locale]/(base)/newsletter/_components/SubscriptionForm.tsx
Comment (Bhavna-Ramachandran) [reply frontend/src/app/[locale]/(base)/newsletter/_components/SubscriptionForm.tsx]: Updates made.

frontend/src/app/[locale]/(base)/newsletter/_components/SubscriptionForm.test.tsx
```diff
@@ -1,13 +1,35 @@
-import { act, render, screen } from "@testing-library/react";
+import { fireEvent, render, screen, waitFor } from "@testing-library/react";
 import SubscriptionForm from "src/app/[locale]/(base)/newsletter/_components/SubscriptionForm";
 
-const mockSubscribeEmail = jest.fn();
+// Mock useClientFetch to delegate to the already-mocked `global.fetch` in tests.
+jest.mock("src/hooks/useClientFetch", () => ({
+  useClientFetch: () => ({
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/newsletter/_components/SubscriptionForm.test.tsx:6]: [non-blocking] one upside of using `useClientFetch` is that you don't have to mock the global - you can just mock the return value of the hook, ex: https://github.com/HHS/simpler-grants-gov/blob/0e074ff4da1f5dd8b840b132d7f22b3a09144fed/frontend/src/app/%5Blocale%5D/(base)/search/_components/SaveSearchModal.test.tsx#L19

frontend/src/app/api/newsletter/subscribe/route.ts
doug-s-nava: https://github.com/HHS/simpler-grants-gov/issues/10793

doug-s-nava (approved):
✅ ✅ ✅ ✅

PR #10449, opened 2026-05-15
[Issue #10255] Add an in-memory form template registry

### LINKED ISSUE #10255: In memory form template registry
### Summary

Form templates are currently synced from a database table at runtime, which creates a fragile dependency. The app can't serve form data unless the DB has been populated by `UpdateFormTask`. We need an in-memory `FormTemplateRegistry` that loads form definitions at startup, resolves all `$ref` schema references, and replaces DB lookups as the source of truth for form data.

### Tech Notes
- `get_by_id_and_major_version()` raises `ValueError` on miss. Return type is `Form`, not `Form | None`
- Tests that need an isolated registry should instantiate `FormTemplateRegistry()` directly rather than using global class state
- The `resolve_jsonschema()` safety net test (calling it on every real registered form) belongs in #10264, not here
- Reference: [Form Template Registry Tech Spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/3074359301)

### Acceptance criteria

- [ ] `FormTemplateRegistry` exists with `register()`, `get_by_id_and_major_version()`, and `get_all()`
- [ ] `register()` raises on all invalid inputs: bad UUID, non-positive `major_version`, duplicate key, `$ref` resolution failure
- [ ] `get_by_id_and_major_version()` raises `ValueError` when the form isn't found
- [ ] Unit tests cover all raise cases and happy path

mikehgrantsgov:
## Summary

Fixes #10255 

## Changes proposed

Added `FormTemplateRegistry` class
`register()` validates the `form_id` is a UUID, major_version is a positive integer, the key isn't a duplicate, and resolves all $ref references in form_json_schema at registration time (raises immediately on any failure)
`get_by_id_and_major_version()` raises `ValueError` on miss
Added unit tests

## Context for reviewers

`register()` mutates form.form_json_schema in place with the resolved schema. These are module-level form objects and resolving an already-resolved schema is idempotent.

## Validation steps

See updated unit tests.

### DIFF
--- a/api/src/form_schema/registry/form_template_registry.py
+++ b/api/src/form_schema/registry/form_template_registry.py
@@ -60,16 +60,16 @@ class FormTemplateRegistry:
 
         self._registry[key] = form
 
-    def get_by_id_and_major_version(self, form_id: uuid.UUID, major_version: int) -> Form:
-        """Return the registered form for the given (form_id, major_version) pair.
+    def get_by_id_and_major_version(self, form_key: FormTemplateKey) -> Form:
+        """Return the registered form for the given FormTemplateKey.
 
         Raises:
             ValueError: if no form is registered with that key.
         """
-        form = self._registry.get(FormTemplateKey(form_id, major_version))
+        form = self._registry.get(form_key)
         if form is None:
             raise ValueError(
-                f"No form registered with form_id={form_id}, major_version={major_version}"
+                f"No form registered with form_id={form_key.form_id}, major_version={form_key.major_version}"
             )
         return form
 
--- a/api/src/form_schema/registry/form_template_registry.py
+++ b/api/src/form_schema/registry/form_template_registry.py
@@ -1,7 +1,13 @@
 import uuid
+from typing import NamedTuple
 
 from src.db.models.competition_models import Form
 from src.form_schema.jsonschema_resolver import resolve_jsonschema
+
+
+class FormTemplateKey(NamedTuple):
+    form_id: uuid.UUID
+    major_version: int
 
 
 class FormTemplateRegistry:
@@ -12,15 +18,18 @@ class FormTemplateRegistry:
     inlined schemas. Raises immediately on any invalid input or broken $ref so the
     application fails fast rather than serving partial data.
 
-    Production code uses a module-level instance; tests that need isolation should
-    instantiate FormTemplateRegistry() directly.
+    The module-level `form_template_registry` singleton is the production instance.
+    Tests that need isolation should instantiate FormTemplateRegistry() directly.
     """
 
     def __init__(self) -> None:
-        self._registry: dict[tuple[uuid.UUID, int], Form] = {}
+        self._registry: dict[FormTemplateKey, Form] = {}
 
     def register(self, form: Form, major_version: int) -> None:
         """Register a form at the given major version.
+
+        In Phase 1 all forms are registered at major_version=1. In Phase 2 the value
+        will be derived from the form directory structure once multi-version support lands.
 
         Raises:
             ValueError: if form_id is not a uuid.UUID instance, major_version is not a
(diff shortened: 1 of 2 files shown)


api/src/form_schema/registry/form_template_registry.py
```diff
@@ -0,0 +1,70 @@
...
+            raise ValueError(f"major_version must be a positive integer, got {major_version!r}")
+
+        key = (form.form_id, major_version)
+        if key in self._registry:
+            raise ValueError(
+                f"Form already registered: form_id={form.form_id}, major_version={major_version}"
+            )
+
+        # Resolve $refs at registration time; raises jsonref.JsonRefError on failure.
+        # Mutates form_json_schema in place — these are module-level form objects and
```
freddieyebra: This looks to function by getting an updated deep copy of the form_json_schema and not mutating in place according to the json_ref.replace_refs docs.

api/src/form_schema/registry/form_template_registry.py
```diff
@@ -0,0 +1,70 @@
...
+    Resolves all $ref references at registration time so callers always receive fully
+    inlined schemas. Raises immediately on any invalid input or broken $ref so the
+    application fails fast rather than serving partial data.
+
+    Production code uses a module-level instance; tests that need isolation should
+    instantiate FormTemplateRegistry() directly.
+    """
+
+    def __init__(self) -> None:
+        self._registry: dict[tuple[uuid.UUID, int], Form] = {}
```
freddieyebra: Can we create a dedicate type or dataclass for the form_id, int tuple so it can be easily referenced and type checked.

api/src/form_schema/registry/form_template_registry.py
```diff
@@ -0,0 +1,70 @@
+import uuid
+
+from src.db.models.competition_models import Form
+from src.form_schema.jsonschema_resolver import resolve_jsonschema
+
+
+class FormTemplateRegistry:
+    """
+    In-memory cache of form template definitions, keyed by (form_id, major_version).
```
freddieyebra, line 15: I am not sure I understand where in memory is this cached? Would we need to have something like a singleton FormTemplateRegistry instance + helpers that utilize the single instance?

api/src/form_schema/registry/form_template_registry.py
```diff
@@ -0,0 +1,70 @@
...
+    application fails fast rather than serving partial data.
+
+    Production code uses a module-level instance; tests that need isolation should
+    instantiate FormTemplateRegistry() directly.
+    """
+
+    def __init__(self) -> None:
+        self._registry: dict[tuple[uuid.UUID, int], Form] = {}
+
+    def register(self, form: Form, major_version: int) -> None:
```
freddieyebra, line 28: Will the major_version parameter be a value derived from the form registry directory structure?

freddieyebra (changes requested):
minor feedback

api/src/form_schema/registry/form_template_registry.py
joshtonava: I think the intention is to have a singleton of this class

api/src/form_schema/registry/form_template_registry.py
mikehgrantsgov: Yes, later on we'll derive it from the directory path. To start we pass 1 explicitly. Added a note to the docstring.

api/src/form_schema/registry/form_template_registry.py
mikehgrantsgov: Made this a singleton, thanks. Added form_template_registry = FormTemplateRegistry() at module level.

api/src/form_schema/registry/form_template_registry.py
mikehgrantsgov: Makes sense, added FormTemplateKey(NamedTuple).

api/src/form_schema/registry/form_template_registry.py
mikehgrantsgov: ah yeah, I fixed the comment. `resolve_jsonschema` returns a new dict.

api/src/form_schema/registry/form_template_registry.py
```diff
@@ -0,0 +1,81 @@
...
+                f"Form already registered: form_id={form.form_id}, major_version={major_version}"
+            )
+
+        # resolve_jsonschema returns a new dict with all $ref pointers inlined;
+        # it does not modify the original schema. Raises jsonref.JsonRefError on failure.
+        form.form_json_schema = resolve_jsonschema(form.form_json_schema)
+
+        self._registry[key] = form
+
+    def get_by_id_and_major_version(self, form_id: uuid.UUID, major_version: int) -> Form:
```
freddieyebra: Can we utilize the FormTemplateKey type here as well may be worth renaming this method as well.
```suggestion
    def get_by_id_and_major_version(self, form_key: FormTemplateKey) -> Form:
```

freddieyebra (changes requested):

api/src/form_schema/registry/form_template_registry.py
mikehgrantsgov: Good call, updated all these refs

freddieyebra (approved):
looks good!

Issue #10286, opened 2026-05-07
Edit opportunity follow ups part 3


brianlee-nava:
### Summary

Cleanup from https://github.com/HHS/simpler-grants-gov/issues/9602 and https://github.com/HHS/simpler-grants-gov/pull/9725

* remove manual token management from publishOpportunityForGrantor
* manage opportunity edit page state without event handlers

### Acceptance criteria

* [ ] manual token management is removed from publishOpportunityForGrantor
* [ ] edit page state management is done with out custom events and handlers

Issue #9841, opened 2026-04-24
Edit opportunity follow ups part 2


doug-s-nava:
### Summary

Cleanup from https://github.com/HHS/simpler-grants-gov/issues/9602 and https://github.com/HHS/simpler-grants-gov/pull/9725

* change name of `OpportunityEditHeaderActions
* switch out isValidDate for dayJs is-valid
* use server action for submit
* remove manual token management from publishOpportunityForGrantor
* manage opportunity edit page state without event handlers
* update error message strings for clientFetch calls in attachment upload input

### Acceptance criteria

* [ ] `OpportunityEditHeaderActions` name is changed to remove `actions`
* [ ] dayJs is-valid is used instead of manual date helper
* [ ] server action using `useActionState` is used to submit opportunity
* [ ] error message strings for clientFetch calls in attachment upload input reflect helpful user messaging

brianlee-nava:
This ticket is related to #10286

PR #10458, opened 2026-05-15
[Issue #9841] Edit opportunity follow-ups part 2

desaiamit:
## Summary

Work for #9841

## Changes proposed

**Part 1: Addressing issue #9841 acceptance criteria**

- Renamed `OpportunityEditHeaderActions` --> `OpportunityEditHeader`
- Replaced manual date validation with `dayjs` strict parsing (`dayjs(date, "YYYY-MM-DD", true).isValid()`) in `saveOpportunityEditAction`
- Replaced direct API calls in Publish flow with `submitOpportunityAction` server action wired via `useActionState` in `OpportunityEditHeader`
- Updated `useClientFetch` error labels in `OpportunityAttachmentUploadInput` to descriptive strings (`"Error uploading opportunity attachment"`, `"Error deleting opportunity attachment"`)

**Part 2: Fixes raised by AI Coding Toolkit review**

  Must-Fix (3 of 4 fixed; 1 documented and deferred):
  - Added errorHeading i18n key under OpportunityEdit.content.alerts (heading was a hardcoded English string)
  - Renamed misleading test cases for invalid-date scenarios in actions.test.ts
  - Un-exported publishOpportunityAction to prevent accidental direct server action calls
  - Documented double-save side effect in form.requestSubmit() path; tracked in #10286 for proper fix

  Should-Fix (all addressed):
  - Consolidated dayjs plugin registration into getConfiguredDayJs() from dateUtil - removed duplicate extend() call
  - Fixed validationErrors truthy check - replaced {} check with Object.keys(...).length > 0 guard
  - Removed unused successMessage field from submitOpportunityAction return
  - Added OpportunityEditHeader.test.tsx with jest-axe scan and behavior coverage

  Nit promoted to Must-Fix in round 2:
  - Extracted duplicate successfulUpdateResponse fixture to module-level constant in actions.test.ts

## Context for reviewers

**React 19 external button limitation:** Browsers route form submission through the form element's own action when a submit button uses `form="..."` to associate with an external form. In React 19, this means `formAction` on the button is ignored - the form's registered action fires instead. The fix is `type="button"` with an `onClick` that calls the `useActionState` dispatch directly via `startTransition`.

**Double-save side effect (known, deferred):** When `submitOpportunityAction` returns validation errors, the header calls `form.requestSubmit()` to route those errors through the form's own `useActionState` so they appear in field-level UI. This causes a second save round-trip. The correct fix requires removing the CustomEvent-based state bridge first - tracked in #10286.

**Portal approach for publish error:** The publish error alert is owned by `OpportunityEditHeader` (which holds the `submitOpportunityAction` state) but must render in the form body alongside save alerts. `createPortal` into `#publish-error-container` achieves this without prop drilling or shared state. Falls back to inline rendering in test environments where the target div is absent.

## Validation steps

1. **Publish button disabled:** Fill in the form with `publishDate`, `fundingType`, `fundingCategories`, and `eligibleApplicants` all empty - confirm Publish button is disabled. Populate all four - confirm it enables.
2. **Publish button disabled while submitting:** Click Publish and confirm the button is disabled while the request is in-flight.
3. **Validation errors on publish:** Clear `publishDate` from the hidden input (browser devtools) and click Publish - confirm field-level validation errors appear in the form.
4. **Permission error on publish:** Log in as a user without Publisher role. Click Publish - confirm a "forbidden" error alert appears in the form body (same location as save success alert), not inline in the header buttons row.
5. **Successful publish:** Log in as a user with Publisher role on a draft opportunity. Click Publish - confirm redirect to `/opportunities`.
6. **Save still works independently:** Click Save (not Publish) - confirm save success alert appears and no publish action fires.
7. **Attachment error labels:** Attempt to upload an invalid file and delete an attachment while offline - confirm error messages read "Error uploading opportunity attachment" and "Error deleting opportunity attachment".


glenblosser-nava (dismissed):
Code LGTM, and I am able to verify the behavior of each validation step. Approved.

frontend/src/app/[locale]/(base)/grantor/opportunity/[id]/edit/actions.test.ts
```diff
@@ -26,9 +30,14 @@ jest.mock(
   () => ({
     createOpportunitySummaryForGrantor: jest.fn(),
     updateOpportunitySummaryForGrantor: jest.fn(),
+    publishOpportunityForGrantor: jest.fn(),
   }),
 );
 
+jest.mock("next/navigation", () => ({
+  redirect: jest.fn(),
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/grantor/opportunity/[id]/edit/actions.test.ts]: you should avoid importing redirect - something like this will cause fewer problems.

```suggestion
const mockRedirect = jest.fn()
...
  redirect: () => mockRedirect()
```

frontend/src/app/[locale]/(base)/grantor/opportunity/[id]/edit/actions.ts
```diff
@@ -268,5 +274,27 @@ export async function publishOpportunityAction(
...
+    return saveResult;
+  }
+
+  const opportunityId = readStringValue(formData.get("opportunityId")).trim();
+  const publishResult = await publishOpportunityAction(opportunityId);
+  if (publishResult.errorMessage) {
+    return publishResult;
+  }
+
+  redirect("/opportunities");
```
Comment (doug-s-nava) [inline frontend/src/app/[locale]/(base)/grantor/opportunity/[id]/edit/actions.ts]: ???

```suggestion
  redirect("/grantor/opportunities");
```

frontend/src/components/opportunity/OpportunityEditHeader.tsx
```diff
@@ -81,39 +95,36 @@ export default function OpportunityEditHeaderActions({
...
+          const alert = (
+            <div className="margin-top-2">
+              <Alert type="error" headingLevel="h3" heading={t("errorHeading")}>
+                {submitState.errorMessage}
+              </Alert>
+            </div>
+          );
+          const target =
+            typeof document !== "undefined"
+              ? document.getElementById("publish-error-container")
```
doug-s-nava, line 124: the correct react way to do this would be with shared state or a shared ref

doug-s-nava (commented):
looks like this doesn't address removing the event handlers from OpportunityEditForm

frontend/src/app/[locale]/(base)/grantor/opportunity/[id]/edit/actions.test.ts
Comment (desaiamit) [reply frontend/src/app/[locale]/(base)/grantor/opportunity/[id]/edit/actions.test.ts]: Good catch - updated. Using `const mockRedirect = jest.fn()` with `redirect: (url: string) => mockRedirect(url)` in the factory.

frontend/src/app/[locale]/(base)/grantor/opportunity/[id]/edit/actions.ts
Comment (desaiamit) [reply frontend/src/app/[locale]/(base)/grantor/opportunity/[id]/edit/actions.ts]: This slipped through - the path was written before the grantor namespace move in #10138 landed on main. Fixed.

frontend/src/components/opportunity/OpportunityEditHeader.tsx
desaiamit: Agreed. The `createPortal` approach was a patch added late during manual testing when I discovered the error needed to render alongside the save alerts - at that point the `useActionState` placement was already settled. 
The right fix is to lift `useActionState(submitOpportunityAction)` to the page level and pass submitState down as props to both components.

desaiamit: As discussed with @doug-s-nava - DEFERRED to #10286

desaiamit:
> looks like this doesn't address removing the event handlers from OpportunityEditForm

~~DEFERRED~~ this AC is part of #10286 - "edit page state management is done with out custom events and handlers"

doug-s-nava (approved):
✅ ✅ ✅ ✅

Issue #10043, opened 2026-05-01
Reduce pre-signed URL expiry for application submission downloads


joshtonava:
### Summary

The "Download submitted application" link is a 2-hour S3 pre-signed URL, which makes the download usable across browsers and after logout for that full window. Submission packages contain PII, so the exposure window should be much shorter (~15 minutes) and aligned closer to a session's lifetime. Opportunity attachments and other public-ish files can keep the longer expiry.

### Tech Notes:
- Auth is already enforced when the URL is generated; this change reduces blast radius if a URL leaks.
- Global default lives at `api/src/adapters/aws/s3_adapter.py:14` 
- Helper: `pre_sign_file_location()` in `api/src/util/file_util.py:99` currently uses the global duration only
- Submission-download call sites that need the shorter expiry:
  - `api/src/db/models/competition_models.py:406` (application attachment download)
  - `api/src/db/models/competition_models.py:478` (submission package download)
- let `pre_sign_file_location` accept an optional `expires_in` override (or add a separate config knob like `presigned_submission_duration`), and pass the shorter value from the submission download paths only. Don't change the default for opportunity/competition public attachments. 

#### Scenarios:
So after the link expires . . .

The user logs back into their account, navigates back to the Application page, it would generate a new link for the download button?
If a user has NOT logged out or navigated away from Application page, and they click the link to download the application again, what happens?

#### How to handle these scenarios: 
Whenever the API is called we'll generate a new presigned URL. We don't cache them in any manner on the backend. If you call the endpoint 10x, we'll make 10 different links which'll probably only differ on the expiration time meaningfully.

So for your cases:

If they come back to the page later, that'll be a new call so long as the frontend itself isn't doing any sort of caching (probably don't want that on the application pages anyways)
Depends on the frontend's behavior. If the frontend calls the backend when first loading the page, then the expiration timer starts on page load effectively. If they call the backend only when clicking, then it could theoretically expire if the timer is shorter than their session token

For the latter scenario, the frontend fetches it on page load since they'd display other info about the submission from fetching it.

### Acceptance criteria

- [x] Submission download pre-signed URLs expire after the new shorter window (default 15 min, env-configurable)
- [x] Opportunity attachments and other public files are unaffected because they use the CDN URL, not pre-signed URLs, so no separate config is needed. The single duration applies to every remaining presigned URL (submissions included)
[...]

PR #10465, opened 2026-05-18
[Issue #10043] Reduce pre-signed URL expiry for application submission downloads

kkrug:
## Summary

Work for #10043 

## Changes proposed

- Updated s3_adapter.py to change the default of `presigned_s3_duration` on `S3Config` from 7200 (2 hours) to 900 (15 minutes)
- Updated test_file_util.py to add a test asserting `pre_sign_file_location()` produces URLs whose `X-Amz-Expires` matches the configured duration

## Context for reviewers

Submission pre-signed URLs currently live for 2 hours, which is longer than necessary for files containing PII. The 2 hour window was originally needed for opportunity attachments, but those now use the CDN URL rather than presigned URLs, so there's no remaining use case that requires the longer duration. This drops the global default to 15 minutes.

## Validation steps

Confirm tests pass.

For manual UAT:
- Grab a submission download URL from the API
- Wait past 15 minutes, then curl or paste it in a browser. This should get a 403 from S3

### DIFF
--- a/api/src/adapters/aws/s3_adapter.py
+++ b/api/src/adapters/aws/s3_adapter.py
@@ -12,6 +12,10 @@ class S3Config(PydanticBaseEnvConfig):
     # locally to use s3mock
     aws_s3_endpoint_url: str | None = Field(alias="AWS_S3_ENDPOINT_URL", default=None)
     presigned_s3_duration: int = 7200  # 2 hours in seconds
+    # Shorter expiry used for application submission downloads, which contain PII.
+    # Submissions are re-fetched on page load, so a tighter window reduces blast
+    # radius if a URL leaks without harming the normal download flow.
+    presigned_submission_duration: int = 900  # 15 minutes in seconds
 
     # CDN URL for public files - if set, will be used instead of presigned URLs
     cdn_url: str | None = None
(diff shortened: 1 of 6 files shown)


api/src/adapters/aws/s3_adapter.py
```diff
@@ -12,6 +12,10 @@ class S3Config(PydanticBaseEnvConfig):
     # locally to use s3mock
     aws_s3_endpoint_url: str | None = Field(alias="AWS_S3_ENDPOINT_URL", default=None)
     presigned_s3_duration: int = 7200  # 2 hours in seconds
+    # Shorter expiry used for application submission downloads, which contain PII.
+    # Submissions are re-fetched on page load, so a tighter window reduces blast
+    # radius if a URL leaks without harming the normal download flow.
+    presigned_submission_duration: int = 900  # 15 minutes in seconds
```
chouinar: Can we just change `presigned_s3_duration` to 15 minutes?

We don't have a use case for presigning urls where we'd actually want 2 hours. We did that because we used to presign all urls we return, and the opportunity attachments we wanted to have a URL that lasted for longer than the frontend cache. Opportunity attachments aren't presigned anymore for any public file, they're the CDN URL.

kkrug: Makes sense, updated

hao10282025-sudo (commented):
LGTM

chouinar (approved):

Issue #8295, opened 2026-02-01
Frontend - Select funding opportunity to initiate Award Recommendation


kevlyu:
### Summary
As the creator of the award recommendation, when I click "create new award recommendation", I need to be able to see a list funding opportunities, so I can select one of those opportunities to initiate an award recommendation document. 

Figma: 

### Acceptance criteria
- [ ] I can see a list of funding opportunities
- [ ] Opportunity List should include:
   - Funding Opportunity # or ID
   - Title
   - Fiscal Year
   - Number of submitted applications
   - CTA: Select
- [ ] When I click CTA "Select", an Award Recommendation will be created for the Funding Opportunity

kevlyu:
@kevlyu refine the AC to match Figma

PR #10466, opened 2026-05-18
8295 select opportunity for award rec page shell

aaronbrown-nava:
## Summary

Work for #8295 

## Changes proposed

Adds new files for page shell. Updates "before you start" page to match header per @slowbot 
image

image

## Context for reviewers

This page will be used for selecting which opportunity the user will be creating an award recommendation for. The page shell just sets up the header, the table functionality will be added in a future PR

## Validation steps
- [ ] navigate to http://localhost:3000/award-recommendation/select-opportunity
- [ ] verify page header matches [figma](https://www.figma.com/design/o5g0d2W5RKHYHjdvVCqTV2/Grantor-Designs-%7C-Quad-5-%7C-Simpler-Grants-2026?node-id=7692-108379&t=yrkvejQANL1natp8-0)


frontend/src/app/[locale]/(base)/award-recommendation/select-opportunity/page.tsx
```diff
@@ -0,0 +1,76 @@
...
+}: SelectOpportunityPageProps) {
+  await params;
+
+  const t = await getTranslations(
+    "AwardRecommendationSelectFundingOpportunity",
+  );
+
+  return (
+    <>
+      <div className="bg-white">
```
Comment (nathan-stricker) [inline frontend/src/app/[locale]/(base)/award-recommendation/select-opportunity/page.tsx]: The header's background color looks off- should be `bg-base-lightest`.  Updating this will also reveal that some padding needs updating- you can take a look at [AwardRecommendationHero.tsx](https://github.com/HHS/simpler-grants-gov/blob/main/frontend/src/components/award-recommendation/AwardRecommendationHero.tsx) for reference to keep the styling similar

frontend/src/app/[locale]/(base)/award-recommendation/select-opportunity/page.tsx
```diff
@@ -0,0 +1,76 @@
...
+              {
+                title: "Award Recommendations",
+                path: `/workspace`,
+              },
+              {
+                title: "Create",
+                path: `/award-recommendation/create`,
+              },
+            ]}
+          />
```
Comment (nathan-stricker) [inline frontend/src/app/[locale]/(base)/award-recommendation/select-opportunity/page.tsx]: I didn't see the breadcrumbs in the original designs, wasn't sure if that was intentional or not. Did Jesse/Kevin mention that they should be included here?
If they are supposed to be included, I'm not sure that "Create" is the right option here, since that would take us back to the previous page. Maybe need design input here?

frontend/src/app/[locale]/(base)/award-recommendation/select-opportunity/page.test.tsx
```diff
@@ -0,0 +1,130 @@
...
+      );
+    });
+
+    it("renders the page title", async () => {
+      const component = await SelectOpportunityPage({
+        params: selectOpportunityParams,
+      });
+      render(component);
+      expect(await screen.findByText("Award Recommendations")).toBeVisible();
+    });
```
Comment (nathan-stricker) [inline frontend/src/app/[locale]/(base)/award-recommendation/select-opportunity/page.test.tsx]: This test will end up finding "Award Recommendations" in the breadcrumbs instead of the title as intended- something like the following would ensure you're asserting on the right component:
```typescript
expect(
    screen.getByRole("heading", { level: 1, name: "pageHeading" }),
  ).toBeVisible();
```

frontend/src/app/[locale]/(base)/award-recommendation/select-opportunity/page.tsx
Comment (aaronbrown-nava) [reply frontend/src/app/[locale]/(base)/award-recommendation/select-opportunity/page.tsx]: removed - redid header area for both this page and before you start page after a conversation with Jesse yesterday

frontend/src/app/[locale]/(base)/award-recommendation/select-opportunity/page.tsx
Comment (aaronbrown-nava) [reply frontend/src/app/[locale]/(base)/award-recommendation/select-opportunity/page.tsx]: Designs have been updated

frontend/src/i18n/messages/en/index.ts
```diff
@@ -2385,4 +2386,12 @@ export const messages = {
     alertMessage:
       "Hard refresh your page when done changing Flags for the changes to fully apply.",
   },
+  AwardRecommendationSelectFundingOpportunity: {
+    pageTitle: "Select funding opportunity | Simpler.Grants.gov",
+    pageHeading: "Award Recommendations",
```
nathan-stricker, line 2391: nit: I believe `pageHeading` is unused and can be deleted

nathan-stricker (approved):

Issue #10019, opened 2026-05-01
Adjust our existing ECS tasks to each use the new job type lookup enum for the ecs background task


chouinar:
### Summary

https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/3026255873/Tech+Spec+Backend+ECS+Task+Job+Locks

Right now on each of our ECS tasks we have `@ecs_background_task("some-job-name"). In #10003 we added a new job type enum that we want to use instead. This should match exactly what it was before for every job, the value should not be changing. We just want the enum for consistency.

Note, the ECS background task isn't tested in our jobs, you do not need to add tests for that here.

### Acceptance criteria

- [ ] Updated `ecs_background_task` calls - double-check that ALL changed values match exactly with their new enum's string value (if not, we should adjust the enum).
- [ ] Updated docs that reference `ecs_background_task` to say to use the enum

PR #10474, opened 2026-05-18
[Issue #10019] Update ecs tasks to use JobType

jakobpederson:
## Summary

Fixes / Work for #10019  

## Changes proposed

For ecs_background_tasks, swapping out the string input for a `JobType` enum.

## Context for reviewers

This allows for consistency.

## Validation steps

- [ ] docs updated
- [ ] ecs_background_tasks enums match exactly the str inputs

### DIFF
--- a/api/src/task/ecs_background_task.py
+++ b/api/src/task/ecs_background_task.py
@@ -10,6 +10,7 @@ from typing import ParamSpec, TypeVar
 import newrelic.agent
 import requests
 
+from src.constants.lookup_constants import JobType
 from src.logging.flask_logger import add_extra_data_to_global_logs
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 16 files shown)


api/src/task/ecs_background_task.py
kkrug: We can tighten up this param type now that `JobType` is always used
```suggestion
def ecs_background_task(task_name: JobType) -> Callable[[Callable[P, T]], Callable[P, T]]:
```

kkrug:
```suggestion
      task_name (JobType): Job type of the ECS task
```

kkrug (dismissed):

api/src/task/ecs_background_task.py
chouinar: I'd actually prefer we leave it a string.

Not every job uses the JobType, we have tests that use a string, and I don't want to add unit test jobs to the JobType class.

We also will soon move this code to the shared repo, and we'd not be able to specify the enum type since it will be different between grantor/applicant systems.

chouinar (approved):

Issue #9984, opened 2026-04-30
Multithreading event handling in workflow service


chouinar:
### Summary

Right now our logic for processing workflow events is to fetch a batch of up to 10 and sequentially process each one. This won't scale very well, and if each event is going to be processed separately, we should process them in parallel. The code was setup to generally support this, although some modifications may be needed.

Importantly, this multithreading should be around the `handle_event` functionality - https://github.com/HHS/simpler-grants-gov/blob/main/api/src/workflow/manager/workflow_manager.py#L184. Everything in the core loop should at least for now, stay single threaded (eg. we process a batch of events in parallel, we're not processing batches in parallel for now). A few callouts:
* `handle_event` has a decorator on it for getting the DB session automatically injected. I don't know how well this'll work with multithreading which usually uses a function - may need a small wrapper for it. Importantly, each thread should still get the db session _after_ the thread spins up - that should work just fine.
* We should have a timeout for how long it takes to process a thread of 30 seconds (feel free to make this configurable to make testing easier by setting it much lower). Any threads that don't respond in 30 seconds should be killed - since everything in the thread should be a single DB transaction already, this should implicitly handle the rollback.
* We can fetch up to 10 events from SQS at a time, 10 threads is fine so no need to worry about a max thread count (but only create those that you need, if we get 3 events, just 3 threads)
* Wait for all threads to complete (successful or not) before we delete messages
* The logic for whether to delete a message for certain error conditions should not change - any errors specifically due to multithreading might need special attention depending on when they occurred, but can probably be classified the same as our unknown exceptions.
* Any individual thread erroring should not impact the other threads. Process them separately

Keep the threading simple, we're not looking to rework everything with async or multiprocessing. Technically multithreading doesn't let two threads execute python code in parallel and the gains are only from IO waiting - but most time we'll spend in a workflow will be IO (DB or other service calls), so it should still be a pretty substantial performance improvement.

### Acceptance criteria

- [x] Multithreading added for processing individual events
- [x] Tests continue to work / aren't significantly complicated by the threading logic
- [x] Manually verified behavior by running service locally with many events in parallel

PR #10478, opened 2026-05-19
[Issue #9984] Multithreading event handling in workflow service

kkrug:
## Summary

Fixes for #9984  

## Changes proposed

- Updated workflow_manager.py to add support for multithreading. `process_batch` now submits each `SqsMessageContainer` to a `ThreadPoolExecutor` (sized to the batch, up to 10) and waits per-future with a per-event timeout
- Updated test_workflow_manager.py to add two new tests

## Context for reviewers

Our logic for processing workflow events was to fetch a batch of up to 10 and sequentially process each one. This won't scale very well, and if each event is going to be processed separately, we should process them in parallel.

## Validation steps

Confirm tests pass.
The multithreading can also be manually tested locally by starting `make cmd args="workflow workflow-main"` in one terminal, and in another terminal, firing ~10 `PUT /v1/workflows/events` requests in parallel against a fresh opportunity ID using the seeded `X-API-Key: internal_workflow_user_key`. The log should show one `Fetched SQS messages message_count=N` followed by N interleaved `Processing event` lines within milliseconds of each other and a `batch_duration_sec` close to a single event's duration.

### DIFF
--- a/api/src/workflow/manager/workflow_manager.py
+++ b/api/src/workflow/manager/workflow_manager.py
@@ -185,52 +185,8 @@ class WorkflowManager:
         """Fetch and process a batch of events from SQS."""
         sqs_containers = self.fetch_messages()
         logger.info("Fetched SQS messages", extra={"message_count": len(sqs_containers)})
-        messages_to_delete: list[str] = []
-        messages_to_keep: list[str] = []
 
-        if sqs_containers:
-            # Each container is processed on its own thread so that IO waits
-            # (DB calls, downstream services) overlap. The core loop stays
-            # single-threaded - we wait for every thread in this batch to
-            # finish before deleting any messages.
-            app = current_app._get_current_object()  # type: ignore[attr-defined]
-            timeout = self.config.workflow_event_processing_timeout_sec
-
-            with ThreadPoolExecutor(max_workers=len(sqs_containers)) as executor:
-                future_to_container = {
-                    executor.submit(_handle_event_in_thread, app, container): container
-                    for container in sqs_containers
-                }
-
-                for future, sqs_container in future_to_container.items():
-                    try:
-                        event_result = future.result(timeout=timeout)
-                    except FuturesTimeoutError:
-                        # Python threads can't be forcibly killed - the thread
-                        # will keep running until its DB call errors or returns,
-                        # at which point the surrounding transaction rolls back.
-                        # We stop waiting on it and treat the message as failed.
-                        logger.warning(
-                            "Workflow event handler exceeded timeout",
-                            extra=sqs_container.get_log_extra() | {"timeout_sec": timeout},
-                        )
-                        messages_to_keep.append(sqs_container.receipt_handle)
-                        continue
-                    except Exception:
-                        logger.exception(
-                            "Failed to handle current event",
-                            extra=sqs_container.get_log_extra(),
-                        )
-                        messages_to_keep.append(sqs_container.receipt_handle)
-                        continue
-
-                    if event_result in [
-                        WorkflowEventProcessingResult.SUCCESS,
-                        WorkflowEventProcessingResult.NON_RETRYABLE_ERROR,
-                    ]:
-                        messages_to_delete.append(sqs_container.receipt_handle)
-                    else:
-                        messages_to_keep.append(sqs_container.receipt_handle)
+        messages_to_delete, messages_to_keep = self._handle_containers(sqs_containers)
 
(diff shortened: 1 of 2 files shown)


api/src/workflow/manager/workflow_manager.py
```diff
@@ -177,26 +188,49 @@ def process_batch(self) -> tuple[list[str], list[str]]:
...
-            ]:
-                messages_to_delete.append(sqs_container.receipt_handle)
-            else:
-                messages_to_keep.append(sqs_container.receipt_handle)
+        if sqs_containers:
+            # Each container is processed on its own thread so that IO waits
+            # (DB calls, downstream services) overlap. The core loop stays
+            # single-threaded - we wait for every thread in this batch to
+            # finish before deleting any messages.
+            app = current_app._get_current_object()  # type: ignore[attr-defined]
```
chouinar: Curious what the `_get_current_object()` call was needed for? We just use `current_app` directly in the place where we're actually grabbing the DB session from the db client attached to the app.

I assume something with the multithreading caused the app to act up?

api/src/workflow/manager/workflow_manager.py
```diff
@@ -177,26 +188,49 @@ def process_batch(self) -> tuple[list[str], list[str]]:
...
-                continue
-
-            if event_result in [
-                WorkflowEventProcessingResult.SUCCESS,
-                WorkflowEventProcessingResult.NON_RETRYABLE_ERROR,
-            ]:
-                messages_to_delete.append(sqs_container.receipt_handle)
-            else:
-                messages_to_keep.append(sqs_container.receipt_handle)
+        if sqs_containers:
```
chouinar: Just because this has ended up pretty indented, what if we did something like:

```py
def process_containers(sqs_containers):
     if not sqs_containers:
          return

     # thread stuff

# In this function, instead call that function
process_containers(sqs_containers)
```

api/src/workflow/manager/workflow_manager.py
```diff
@@ -177,26 +188,49 @@ def process_batch(self) -> tuple[list[str], list[str]]:
...
+
+                for future, sqs_container in future_to_container.items():
+                    try:
+                        event_result = future.result(timeout=timeout)
+                    except FuturesTimeoutError:
+                        # Python threads can't be forcibly killed - the thread
+                        # will keep running until its DB call errors or returns,
+                        # at which point the surrounding transaction rolls back.
+                        # We stop waiting on it and treat the message as failed.
+                        logger.warning(
```
chouinar: I wouldn't expect timeouts to be commonplace, probably want to log an error for that.

```suggestion
                        logger.exception(
```

api/src/workflow/manager/workflow_manager.py
```diff
@@ -177,26 +188,49 @@ def process_batch(self) -> tuple[list[str], list[str]]:
...
+                        )
+                        messages_to_keep.append(sqs_container.receipt_handle)
+                        continue
+                    except Exception:
+                        logger.exception(
+                            "Failed to handle current event",
+                            extra=sqs_container.get_log_extra(),
+                        )
+                        messages_to_keep.append(sqs_container.receipt_handle)
+                        continue
```
chouinar: Having the errors here have their own logic to add to the list feels iffy. What if instead we just set the status and let the below logic cover them?

```suggestion
                        event_result = WorkflowEventProcessingResult.GENERAL_ERROR
                    except Exception:
                        logger.exception(
                            "Failed to handle current event",
                            extra=sqs_container.get_log_extra(),
                        )
                        event_result = WorkflowEventProcessingResult.GENERAL_ERROR
```

api/src/workflow/manager/workflow_manager.py
kkrug: Yeah it was the multithreading. `current_app` resolves per-thread and a new worker thread starts with no Flask app context, causing a "Working outside of application context" runtime error.

api/src/workflow/manager/workflow_manager.py
kkrug: Updated

api/src/workflow/manager/workflow_manager.py
kkrug: You know I originally did this and then second guessed myself changed it to a warning. Updated

api/src/workflow/manager/workflow_manager.py
kkrug: Updated

chouinar (approved):

PR #10437, opened 2026-05-14
[Issue #9697] Remove requirement of internal token for form print view

freddieyebra:
## Summary

Fixes / Work for #9697 

## Changes proposed

This pr removes the requirement of needing an internal JWT token for previewing the print view of a form on an application. 

## Context for reviewers

Previously, the form print view page was only able to viewed if there was an existing header that contained an internal token on the get request. This pr removes that requirement so that if an authenticated user has the appropriate access to the form, they can view the print preview.

### Local
You will want to install [modheader](https://chromewebstore.google.com/detail/modheader-modify-http-hea/mknbkgdikangjebcpfcbkiioojainjnh?hl=en-US&utm_source=ext_sidebar) extension to test the internal token functionality

1. Run the api and frontend
2. Login as a user with an submitted or saved application
3. Navigate to a form that has been saved and click to form link
4. The url will have the path like
`/workspace/applications/fd4437a7-a39d-455c-bd01-ce187698ce0f/form/a7f97c87-a5e5-41e2-9eab-bcf8f555c522`
You will want to modify that url to be
`/print/application/fd4437a7-a39d-455c-bd01-ce187698ce0f/form/a7f97c87-a5e5-41e2-9eab-bcf8f555c522`
5. Ensure that you can view the print view
6. Logout
7. and run the following in the api/ directory: `make generate-internal-token`
8. Use modheader extension to set `X-SGG-Internal-Token` with the JWT from step 6
9. Navigate the print view url from step 4 and ensure you can view

### UAT Validation
1. Sign in to https://grantee1.team.simpler.grants.gov
2. Create an individual owned application and save at least 1 form
3. Navigate to view the form. The URL will look like `/workspace/applications/fd4437a7-a39d-455c-bd01-ce187698ce0f/form/a7f97c87-a5e5-41e2-9eab-bcf8f555c522`. To get the print view, modify the form view URL to be `/print/application/fd4437a7-a39d-455c-bd01-ce187698ce0f/form/a7f97c87-a5e5-41e2-9eab-bcf8f555c522`
4. Create an organization owned application and save at least 1 form
5. Navigate to view the form. To obtain the print view URL as described in step 3
6. You should now have 2 print view URLs. 1 individual owned and 1 organization owned
7. Still logged in, navigate to the individual owned print view url and verify you can view it
8. Still logged in, navigate to the organization owned print view url and verify you can view it 
9. Share the individual owned print view to a member of an organization you are a member of and verify it cannot be viewed
[...]

PR #10479, opened 2026-05-19
Improve print application form authorization/error handling

Bhavna-Ramachandran:
## Summary
This PR builds on top of PR #10437 to retain the related in-progress changes and adds few follow-up improvements to print application form authorization/error handling.

## Changes proposed
- Added handling for unauthorized vs forbidden API responses in getFormData.ts
- Added redirect handling for:
  - unauthenticated users -> /unauthenticated
  - authenticated users without access -> /unauthorized
- Expanded inline documentation around supported print/PDF access patterns
- Added a new /unauthorized page using the standard base layout and existing messaging

## Context for reviewers
- These changes were discovered while reviewing PR #10437 and were scoped separately to keep the original review focused.
- Previously, all error cases in this flow showed a generic error message - “We’re sorry. There seems to have been an error. Please try again.”, regardless of whether the issue was authentication, authorization, or missing data.
- This PR has been verified in local. The following behaviors were tested:
  - No session - > Redirects to /unauthenticated
  - Logged in without access -> Redirects to /unauthorized (Indv user cannot access Org user print view and vice versa)
  - Valid access -> Form renders normally
  - Valid Docraptor token -> Form renders normally

## Validation steps
Run the flow locally and verify that the above behaviors work as expected.


frontend/src/app/[locale]/(print)/print/application/[applicationId]/form/[appFormId]/page.tsx
```diff
@@ -49,10 +49,13 @@ interface FormPageProps {
...
-  by Docraptor. Docraptor will not be able to log in as a user to access application information.
-  Instead, requests from Docraptor will contain an "internal token" header ("X-SGG-Internal-Token")
-  containing an alternate authorization token that can be used to fetch the form data without logging in.
+  This page supports two access patterns:
+  1. Automated PDF generation (e.g., Docraptor):
+     - Requests include an "internal token" header ("X-SGG-Internal-Token") for authorization.
+     - Used by services that cannot log in as a user.
+  2. Regular authenticated users:
+     - Users logged in via session can browse directly to this page to view/print completed application forms.
+  If neither a valid internal token nor a valid user session is present, the users are redirected to appropriate error page.
```
Comment (freddieyebra) [inline frontend/src/app/[locale]/(print)/print/application/[applicationId]/form/[appFormId]/page.tsx:58]: 👍 this is great.

freddieyebra (changes requested):
These are great improvements.

For the unauthenticated use case, if a user who has authorization for a print view URL for example, if they are not logged in and try to navigate to that URL, they will be redirected to `/unauthenticated` which is correct. The issue is that once they log in, they should be redirected back to their initial page they tried to access and not the /unauthenticated page.

An example of the expected behavior of this would be the /workspace/applications page. If a user is not logged in and tries to navigate there, they are shown and error. and if the user logs in successfully, they are redirected to that initial page.

PR #10480, opened 2026-05-19
[Issue #9705] Setup DynamoDB client in backend code

### LINKED ISSUE #9705: Setup DynamoDB client in our backend code
### Summary

We want to setup logic to wrap the DynamoDB boto3 client https://docs.aws.amazon.com/boto3/latest/reference/services/dynamodb.html

This is similar to the other clients we've created for interacting with other AWS services like SQS to make the interface a bit easier to work with.

We want to wrap the following functions:
* `get_item` - With the ability to pass in a table name, key, whether to use consistent read (default to true)

We only need to handle `get_item` as our API will only query for a record, updates are handled elsewhere.

Other considerations:
* Like our SQS/s3 clients, we likely need to handle local development for connecting to the local instance (like how we have the `S3_ENDPOINT_URL` and `SQS_ENDPOINT_URL` variables)
* Tests should work using the moto3 library like we do elsewhere - will need to setup a dynamoDB table https://docs.getmoto.org/en/latest/docs/services/dynamodb.html

### Acceptance criteria

- [ ] DynamoDB adapter client created that can `get_item`
- [ ] Adapter client can talk to local dynamoDB (outside unit tests)
- [ ] Adapter client works with moto3 setup table (in unit tests)

hao10282025-sudo:
## Summary

Fixes #9705  

## Changes proposed

We want to setup logic to wrap the DynamoDB boto3 client
https://docs.aws.amazon.com/boto3/latest/reference/services/dynamodb.html

## Context for reviewers

This is similar to the other clients we've created for interacting with other AWS services like SQS to make the interface a bit easier to work with.

We want to wrap the following functions:

get_item - With the ability to pass in a table name, key, whether to use consistent read (default to true)
We only need to handle get_item as our API will only query for a record, updates are handled elsewhere.

Other considerations:

Like our SQS/s3 clients, we likely need to handle local development for connecting to the local instance (like how we have the S3_ENDPOINT_URL and SQS_ENDPOINT_URL variables)
Tests should work using the moto3 library like we do elsewhere - will need to setup a dynamoDB table https://docs.getmoto.org/en/latest/docs/services/dynamodb.html

## Validation steps

- DynamoDB adapter client created that can get_item
- Adapter client can talk to local dynamoDB (outside unit tests)
- Adapter client works with moto3 setup table (in unit tests)

### DIFF
--- a/api/src/adapters/aws/dynamodb_adapter.py
+++ b/api/src/adapters/aws/dynamodb_adapter.py
@@ -1,8 +1,94 @@
-from pydantic import Field
+import logging
+from typing import Any
 
+import boto3
+import botocore.client
+from pydantic import BaseModel, Field
+
+from src.adapters.aws import get_aws_config, get_boto_session
 from src.util.env_config import PydanticBaseEnvConfig
+
+logger = logging.getLogger(__name__)
 
 
 class DynamoDBConfig(PydanticBaseEnvConfig):
     aws_dynamodb_endpoint_url: str | None = Field(alias="AWS_DYNAMODB_ENDPOINT_URL", default=None)
     file_scan_cache_table_name: str = Field(alias="FILE_SCAN_CACHE_TABLE_NAME")
+
+
+class DynamoDBGetItemResponse(BaseModel):
+    """Represents the response from getting an item from DynamoDB."""
+
+    item: dict[str, Any] | None = Field(default=None)
+
+
+def get_boto_dynamodb_client(
+    dynamodb_config: DynamoDBConfig | None = None, session: boto3.Session | None = None
+) -> botocore.client.BaseClient:
+    if dynamodb_config is None:
+        dynamodb_config = DynamoDBConfig()
+
+    params = {}
+    if dynamodb_config.aws_dynamodb_endpoint_url is not None:
+        params["endpoint_url"] = dynamodb_config.aws_dynamodb_endpoint_url
+
+    if session is None:
+        session = get_boto_session()
+
+    return session.client("dynamodb", region_name=get_aws_config().aws_region, **params)
+
+
+class DynamoDBClient:
+    def __init__(self, dynamodb_client: botocore.client.BaseClient | None = None):
+        self.client = dynamodb_client or get_boto_dynamodb_client()
+
+    def get_item(
+        self,
+        table_name: str,
+        key: dict[str, Any],
+        consistent_read: bool = True,
+    ) -> DynamoDBGetItemResponse:
+        """
+        Get an item from DynamoDB by its key.
+
+        Args:
+            table_name: The name of the DynamoDB table
+            key: The key of the item to retrieve (in DynamoDB format)
+            consistent_read: Whether to use consistent read (default: True)
+
+        Returns:
+            DynamoDBGetItemResponse containing the item if found, None otherwise
+        """
+        try:
+            logger.info(
+                "Getting item from DynamoDB",
+                extra={"table_name": table_name, "key": str(key)},
+            )
+
+            response = self.client.get_item(
+                TableName=table_name,
+                Key=key,
+                ConsistentRead=consistent_read,
+            )
+
+            item = response.get("Item")
+
+            if item:
+                logger.info(
+                    "Successfully retrieved item from DynamoDB",
+                    extra={"table_name": table_name, "key": str(key)},
+                )
+            else:
+                logger.info(
+                    "Item not found in DynamoDB",
+                    extra={"table_name": table_name, "key": str(key)},
+                )
+
+            return DynamoDBGetItemResponse(item=item)
+
+        except Exception:
+            logger.exception(
+                "Failed to get item from DynamoDB",
+                extra={"table_name": table_name, "key": str(key)},
+            )
+            raise
--- a/api/src/adapters/aws/dynamodb_adapter.py
+++ b/api/src/adapters/aws/dynamodb_adapter.py
@@ -92,7 +92,13 @@ class DynamoDBClient:
         try:
             logger.info(
                 "Getting item from DynamoDB",
-                extra={"table_name": table_name, "key": key},
+                extra={
+                    "table_name": table_name,
+                    "key_name": key_name,
+                    "key_type": key_type,
+                    "value": value,
+                    "consistent_read": consistent_read,
+                },
             )
 
             response = self.client.get_item(
@@ -106,12 +112,24 @@ class DynamoDBClient:
             if item:
                 logger.info(
                     "Successfully retrieved item from DynamoDB",
-                    extra={"table_name": table_name, "key": key},
+                    extra={
+                        "table_name": table_name,
+                        "key_name": key_name,
+                        "key_type": key_type,
+                        "value": value,
+                        "consistent_read": consistent_read,
+                    },
                 )
             else:
                 logger.info(
                     "Item not found in DynamoDB",
-                    extra={"table_name": table_name, "key": key},
+                    extra={
+                        "table_name": table_name,
+                        "key_name": key_name,
+                        "key_type": key_type,
+                        "value": value,
+                        "consistent_read": consistent_read,
+                    },
                 )
 
             return DynamoDBGetItemResponse(item=item)
@@ -119,6 +137,12 @@ class DynamoDBClient:
         except Exception:
             logger.exception(
                 "Failed to get item from DynamoDB",
-                extra={"table_name": table_name, "key": key},
+                extra={
+                    "table_name": table_name,
+                    "key_name": key_name,
+                    "key_type": key_type,
+                    "value": value,
+                    "consistent_read": consistent_read,
+                },
             )
             raise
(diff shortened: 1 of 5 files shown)


api/src/adapters/aws/dynamodb_adapter.py
```diff
@@ -1,8 +1,94 @@
...
+                Key=key,
+                ConsistentRead=consistent_read,
+            )
+
+            item = response.get("Item")
+
+            if item:
+                logger.info(
+                    "Successfully retrieved item from DynamoDB",
+                    extra={"table_name": table_name, "key": str(key)},
```
kkrug: We don't need to cast to str in the `extra` dict, it handles type conversion automatically.
```suggestion
                    extra={"table_name": table_name, "key": key},
```

api/src/adapters/aws/dynamodb_adapter.py
```diff
@@ -1,8 +1,94 @@
...
+                )
+            else:
+                logger.info(
+                    "Item not found in DynamoDB",
+                    extra={"table_name": table_name, "key": str(key)},
```
kkrug:
```suggestion
                    extra={"table_name": table_name, "key": key},
```

api/src/adapters/aws/dynamodb_adapter.py
```diff
@@ -1,8 +1,94 @@
...
+                )
+
+            return DynamoDBGetItemResponse(item=item)
+
+        except Exception:
+            logger.exception(
+                "Failed to get item from DynamoDB",
+                extra={"table_name": table_name, "key": str(key)},
```
kkrug:
```suggestion
                extra={"table_name": table_name, "key": key},
```

api/tests/src/adapters/aws/test_dynamodb_adapter.py
```diff
@@ -0,0 +1,195 @@
...
+        )
+
+    def test_uses_provided_session(self):
+        mock_session = Mock()
+        get_boto_dynamodb_client(session=mock_session)
+        mock_session.client.assert_called_once_with(
+            "dynamodb",
+            region_name="us-east-1",
+            endpoint_url=DynamoDBConfig().aws_dynamodb_endpoint_url,
+        )
```
kkrug, line 53: I don't know that this tests are necessary. Your `test_get_item_` tests below already exercise the real path end-to-end, making these redundant.

api/tests/src/adapters/aws/test_dynamodb_adapter.py
```diff
@@ -0,0 +1,195 @@
...
+        dynamodb_client = DynamoDBClient(dynamodb_client=boto_client)
+
+        with pytest.raises(ClientError):
+            with caplog.at_level(logging.ERROR):
+                dynamodb_client.get_item(
+                    table_name=invalid_table,
+                    key={"attachment_id": {"S": "test-id"}},
+                )
+
+        assert "Failed to get item from DynamoDB" in caplog.text
```
kkrug: We can drop this line. The `caplog.records` lookup that immediately follows makes it redundant. Same for the other two instances below.

api/src/adapters/aws/dynamodb_adapter.py
```diff
@@ -1,8 +1,94 @@
...
+
+
+class DynamoDBClient:
+    def __init__(self, dynamodb_client: botocore.client.BaseClient | None = None):
+        self.client = dynamodb_client or get_boto_dynamodb_client()
+
+    def get_item(
+        self,
+        table_name: str,
+        key: dict[str, Any],
```
chouinar: I wonder if it would be useful to obfuscate away the format of the key / split it up into several fields.

Since it seems you have to pass in `{"attachment_id": {"S": "test-id-123"}}` what if instead that was 3 values:
* key name
* key type (maybe with a default of `S`)
* value

Otherwise, I'd add an example to the docstrings to explain the format since that's a really weird format.

api/tests/conftest.py
```diff
@@ -513,6 +514,30 @@ def workflow_sqs_queue(mock_sqs, monkeypatch):
...
+        KeySchema=[
+            {"AttributeName": "attachment_id", "KeyType": "HASH"},
+        ],
+        AttributeDefinitions=[
+            {"AttributeName": "attachment_id", "AttributeType": "S"},
+        ],
+        BillingMode="PAY_PER_REQUEST",
+    )
+    monkeypatch.setenv("FILE_SCAN_CACHE_TABLE_NAME", table_name)
+    return table_name
```
chouinar, line 538: Since what we're creating here is specifically the file scan mock, should we name it that? Also, just to make sure we're setting it up as expected, the table's key should be `file_id`

```suggestion
@pytest.fixture
def file_scan_dynamodb_table(mock_dynamodb, monkeypatch):
    dynamodb = boto3.client("dynamodb", region_name="us-east-1")
    table_name = "test-local-virus-scan"
    dynamodb.create_table(
        TableName=table_name,
        KeySchema=[
            {"AttributeName": "file_id", "KeyType": "HASH"},
        ],
        AttributeDefinitions=[
            {"AttributeName": "file_id", "AttributeType": "S"},
        ],
        BillingMode="PAY_PER_REQUEST",
    )
    monkeypatch.setenv("FILE_SCAN_CACHE_TABLE_NAME", table_name)
    return table_name
```

hao10282025-sudo: Updated for this file and the "setup_local_dynamodb.py" altogether

api/src/adapters/aws/dynamodb_adapter.py
hao10282025-sudo: updated with separate parameters

api/tests/src/adapters/aws/test_dynamodb_adapter.py
hao10282025-sudo: removed for all instances

api/tests/src/adapters/aws/test_dynamodb_adapter.py
hao10282025-sudo: hmm I copied the pattern from sqs test, I guess it is fine as it covers 2 types of client initialization?
added docstrings for it

api/src/adapters/aws/dynamodb_adapter.py
hao10282025-sudo: updated
I compared the print-out before and after the str() removed, and it looks same. Claude fraud me
image

chouinar: By default in python, it wouldn't properly handle the extra containing a UUID or many other common types (dates, datetimes, enums), but our logging is configured to handle those. Unless you tell an AI that we have a configuration for that, it's not going to know.

api/src/adapters/aws/dynamodb_adapter.py
```diff
@@ -1,8 +1,124 @@
...
+                value="test-id",
+                consistent_read=False
+            )
+        """
+        key = {key_name: {key_type: value}}
+
+        try:
+            logger.info(
+                "Getting item from DynamoDB",
+                extra={"table_name": table_name, "key": key},
```
kkrug: Using a nested dict in `extra` can be awkward for New Relic querying. It may be better to log the individual values as flat strings (e.g. `"key_name": key_name, "key_type": key_type,` etc.)

api/tests/src/adapters/aws/test_dynamodb_adapter.py
kkrug: That's fine, we can keep it as is

api/src/adapters/aws/dynamodb_adapter.py
hao10282025-sudo: hmm ok, will check more on that

api/src/adapters/aws/dynamodb_adapter.py
hao10282025-sudo: updated
I searched online looks like consistent_read also worth to be included for the new relic

api/src/adapters/aws/dynamodb_adapter.py
hao10282025-sudo: ok I revisit the previous slack thread and PRs, now I think I have better understanding of it

api/src/adapters/aws/dynamodb_adapter.py
```diff
@@ -1,8 +1,148 @@
...
+        try:
+            logger.info(
+                "Getting item from DynamoDB",
+                extra={
+                    "table_name": table_name,
+                    "key_name": key_name,
+                    "key_type": key_type,
+                    "value": value,
+                    "consistent_read": consistent_read,
+                },
```
kkrug: Since this is used a few times, we can clean this up a bit and do `extra=log_extra` and create the variable at the top of the method:
```
log_extra = {
    "table_name": table_name,
    "key_name": key_name,
    "key_type": key_type,
    "value": value,
    "consistent_read": consistent_read,
}
```

hao10282025-sudo: updated

kkrug (approved):

PR #10481, opened 2026-05-19
[Issue #10264] Wire form templates into app startup

### LINKED ISSUE #10264: Wire form templates into app startup via registry
### Summary

Once the `FormTemplateRegistry` exists (#10255), the app needs to populate it at startup so form data is available in memory before any requests are served. This ticket wires the registration loop into `forms/__init__.py`, guarantees the registry is always populated via an explicit `app.py` import, and adds a safety net test that catches broken `$ref` references at PR time rather than at deployment. After this ticket, the app no longer needs to query the database to serve form data.

**Tech Notes**
- Today the registry is transitively populated through the task import chain. Once those files are deleted, the explicit `app.py` import added here becomes the only guarantee.
- `get_active_forms()` becomes a thin wrapper delegating to `FormTemplateRegistry.get_all()` so callers stay in sync with the registry without a separate list
- Reference: [Form Template Registry Tech Spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/3074359301)

### Acceptance criteria

- [x] All 16 form templates are registered in `FormTemplateRegistry` on app startup
- [x] `get_active_forms()` delegates to `FormTemplateRegistry.get_all()`
- [x] Existing unit tests pass
- [x] Unit test passes that calls `resolve_jsonschema()` on every registered form

babebe:
## Summary

Fixes for #10264  

## Changes proposed

Added form registration loop in `src/form_schema/forms/__init__.py`
Updated `get_active_forms()` to delegate to `FormTemplateRegistry.get_all()`
Added explicit function call  in` app.py` to guarantee registry population at startup
Added startup registration test verifying all 16 form templates are registered

## Validation steps

Covered in Test.
Can also Run Make console--

```
from src.form_schema.forms import get_active_forms

forms = get_active_forms()
len(forms)

```
should return 16 forms.


api/tests/src/form_schema/test_jsonschema_resolver.py
```diff
@@ -161,3 +162,17 @@ def test_resolve_jsonschema_invalid_ref():
 
     with pytest.raises(jsonref.JsonRefError, match="Unresolvable JSON pointer"):
         resolve_jsonschema(schema)
+
+
+def test_resolve_jsonschema_succeeds_for_all_registered_forms():
+    """Verify all registered forms successfully resolve JSON schema references."""
+    import src.form_schema.forms  # noqa: F401
```
mikehgrantsgov: Can imports go at the top of the file?

api/tests/src/form_schema/registry/test_form_template_registry.py
```diff
@@ -209,3 +209,15 @@ def test_get_raises_for_wrong_major_version():
...
+# ---------------------------------------------------------------------------
+# Startup wiring / integration behavior
+# ---------------------------------------------------------------------------
+
+
+def test_all_forms_are_registered_on_startup():
+    import src.form_schema.forms  # noqa: F401 ; triggers registration at import time
+
+    forms = form_template_registry.get_all()
+    assert len(forms) == 16
```
mikehgrantsgov: assert len(forms) == len(_ALL_FORMS). 16 will break when we add new forms

api/src/form_schema/forms/__init__.py
```diff
@@ -17,24 +18,30 @@
...
+    CD511_v1_1,
+    SupplementaryNEHCoverSheet_v3_0,
+    GG_LobbyingForm_v1_1,
+    EPA_FORM_4700_4_v5_0,
+    EPA_KEY_CONTACT_v2_0,
+]
+
+
+for _form in _ALL_FORMS:
+    form_template_registry.register(_form, major_version=1)
```
chouinar: Because this register function calls `resolve_jsonschema` - I'm not sure what happens if we attempt to re-resolve a json schema we already resolved (at the very least, extra cycles). RIght now we do that resolution in a few places like the form unit tests, seed script, and form upload scripts.

--

Also, is there a way we can load this explicitly rather than when this file gets imported? This means that anything that imports this file will resolve this, even if it's not needed, which will be basically anything in apply, and mean the resolution of this always has to happen for running things like unit tests.

Could we do something like:

```py
def init_form_registry():
     logger.info("Registering forms")
     if len(form_template_registry.get_all()) == 0:
           for _form in _ALL_FORMS:
                 form_template_registry.register(_form, major_version=1)
     logger.info("Finished registering forms")
```

Then in create_app, we call this function? If we need it anywhere else like for unit tests, we can just call the function in a fixture.

The log messages also make it so we can track how long this takes, and that can only be added if this runs after logging initializes in create_app

babebe: ~~Agreed — that approach makes the initialization flow much more explicit. I can refactor this into an init_form_registry() function~~

api/tests/src/form_schema/registry/test_form_template_registry.py
babebe: Good catch!! Updated

api/tests/src/form_schema/test_jsonschema_resolver.py
babebe: Moved!

api/src/form_schema/forms/__init__.py
babebe: Moving to an explicit init function would break existing consumers that rely on import-time registration via forms/__init__.py. Keeping registration tied to the import chain preserves current behavior without requiring changes across scripts and tests, and will be replaced later by the planned directory-based loader in #[9636](https://github.com/HHS/simpler-grants-gov/issues/9636).
Yes, `resolve_jsonschema` is called in multiple places, and re-running it will re-process the schema and incur some extra cycles. In practice though, each of those flows (registry registration, seed scripts, form upload scripts, and tests) runs in isolation.

mikehgrantsgov (dismissed):
Looks good from my end, but would like @chouinar to also look again since it seems like his feedback was addressed but would like him to confirm.

api/src/form_schema/forms/__init__.py
chouinar: Could you talk more about what breaks?

I am hesitant of adding something that does resolution during import time as we add more forms, this could take a considerable amount of time, and I'd like us to at least track the metrics, but you can't do any logging until after the app initializes.

chouinar (changes requested):
want to hold on merging this until we talk through the import issues

api/src/form_schema/forms/__init__.py
babebe: One example is the generate-xml CLI flow. It indirectly loads forms through from `src.form_schema.forms import get_active_forms`, and today that import path is what ensures forms are available to the execution flow.

If registration only happens via an explicit `create_app()` initialization step, flows like this one or the `seed
_local_db.py` would also need to explicitly initialize the registry before use.

chouinar: Yes, adding that explicit initialization function is what I'm suggesting we do. If you put it in create_app, that gets most of them as any CLI command that is attached to a blueprint calls create_app.

Otherwise, I believe we'd just need to account for:
* Our seed script
* The `load_active_forms` fixture in conftest.py
* The XML code would likely need some adjustments since it probably shouldn't be grabbing things globally anyways 
  * xml_generation_cli - could just be fine with not having exact validation on the input parameter, fine to grab the list after the task starts
  * test_runner - If this is just run as part of the cli script, can we pass the value into the init of the test runner since if we adjusted the cli, we'd have it.
  * The submission XML assembler runs as part of the create submission ECS task, so would have initialized it

api/src/task/forms/form_task_shared.py
```diff
@@ -35,6 +35,7 @@ class FormTaskConfig(PydanticBaseEnvConfig):
 class BaseFormTask(abc.ABC):
     def __init__(self) -> None:
         self.config = FormTaskConfig()
+        init_form_registry()
```
chouinar, line 38: You shouldn't need to init it here, although it doesn't necessarily hurt. These tasks all run as CLI commands.

chouinar (approved):

api/src/task/forms/form_task_shared.py
babebe: Yeah, both tasks that inherit from BaseFormTask depend on this initialization, so the base class felt like the safest/common place to enforce it.

PR #10493, opened 2026-05-19
[Issue #10406] Setup PyPi integration for grants_shared

### LINKED ISSUE #10406: Publish grants_shared to PyPi
### Summary

We want to publish our package to PyPi and want to setup the process to do that.

We should look into our options for running this as a github action, but in the short-term we'd be fine with a makefile command to do it.

### Acceptance criteria

- [ ] Package published to PyPi
- [ ] README updated with information on how to publish

chouinar:
## Summary

Fixes #10406 

## Changes proposed
* Adjust grants_shared code to be able to publish to PyPi
* Setup a github action that we can manually run (for now) that'll publish the package to PyPi
* Removed a lot of unnecessary dockerfile parts that we don't need as we don't have to deploy the image and only need local

## Context for reviewers
PyPi is the global dependency store for all of Python. All the dependencies we fetch come from there, this is just adding our new shared package to it.

I setup the publish to PyPi to use an API key connected to only the grants_shared repo. There's another approach that authenticates via Github without any credentials, but it doesn't work with our docker approach. We'd have to go fully natively running for that (doable, but not sure we want to go against the patterns we have). 

Because we don't need to have any concept of a release image as this code won't ever be deployed as an image, I heavily modified the Dockerfile and removed everything we didn't need. The Dockerfile is purely for local development + generating the wheels for releasing to PyPi. It's fine if we have tests and dev dependencies because none of that is included in the package we import elsewhere (triple-checked that by adding to the API temporarily).

## Validation steps
I tested all of this through the github action I created, you can see the releases I pushed over https://pypi.org/project/grants-shared/ - will clean up some of those and also move this to `0.2.0` after this is merged


.github/workflows/publish-grants-shared.yml
```diff
@@ -0,0 +1,57 @@
+name: Release Grants Shared Package to PyPi
+on:
+  workflow_call:
```
kkrug: The PR notes mention being able to run this manually. Would this need `workflow-dispatch` in order to do that?

backend/grants_shared/Makefile
```diff
@@ -118,4 +121,25 @@ lint-mypy:
...
+
+##################################################
+# Release functionality
+##################################################
+
+bump-patch-version: ## Bump the patch version of the code (eg. 1.0.1 -> 1.0.2)
+	uv version --bump patch
+
+bump-minor-version: ## Bump the minor version of the code (eg. 1.0.1 -> 1.1.0)
+	uv version --bump minor
```
kkrug: A lot of our other commands here use `$(PY_RUN_CMD)` to run through Docker by default. Do we need to add that for these two as well?

backend/grants_shared/tests/grants_shared/smoke_test.py
```diff
@@ -0,0 +1,21 @@
+"""
```
kkrug, line 1: Will this run as part of every `make test` invocation?

.github/workflows/publish-grants-shared.yml
```diff
@@ -0,0 +1,57 @@
...
+          make docker-cmd args="uv run --isolated --with dist/grants_shared-${{steps.get-version.outputs.project_version}}.tar.gz tests/grants_shared/smoke_test.py"
+
+      # This publish relies on an API key configured
+      # for the repo and stored in a Github secret.
+      #
+      # See the following for more information:
+      # * https://docs.astral.sh/uv/guides/integration/github/#publishing-to-pypi
+      - name: Release to PyPi
+        run: |
+          make docker-cmd args="uv publish"
```
kkrug: Do we want a guard here against running this from a non-main branch?

backend/grants_shared/tests/grants_shared/smoke_test.py
chouinar: No, it won't get picked up by pytest since the file isn't named `test_...`. Nothing imports it either, so it won't run. 

I can add a note.

.github/workflows/publish-grants-shared.yml
chouinar: Good catch, put the wrong one here copying from some other workflows we have. This is for calling workflows from other workflows, which we do a lot of.

.github/workflows/publish-grants-shared.yml
chouinar: Probably? We don't have that on any other workflows, but I think that would enforce the "get this reviewed before publishing" bit.

chouinar:
At @mdragon 's suggestion, I went to a native build approach instead of with Docker so we could get publish working.

Tested it by making the github action run on merge (now removed) if you want to take a look:
https://github.com/HHS/simpler-grants-gov/actions/runs/26301348236

kkrug (approved):

Issue #10502, opened 2026-05-20
Only install Chrome for Staging Playwright that only runs Chrome


mdragon:
### Summary

We only run Chrome and Chrome mobile on the Staging path but we're installing many browsers because we used the default install command.

### Acceptance criteria

- Chrome is only browser installed
- Tests pass

PR #10506, opened 2026-05-20
[Issue #10502] Install only the playwright browsers needed for the run

mdragon:
## Summary

Work for #10502

## Changes proposed

Expose each playwright browser and dependency install as a input on the job. Set those based on the shard in the Local (CI/CD) runs (pre-install chrome since it's used in Shards 1 and 4). 

## Context for reviewers

We were installing all browsers and dependencies for all tests. For CI/CD we've re-worked that to cache chrome and firefox, but we always were having to install Webkit as it installs all over and isn't easily cachable. For Staging though we were installing all browsers and then only testing with chrome. So with this change we'll only cache chrome, and install Firefox and Webkit when needed.

## Validation steps

- Tests pass

### DIFF
--- a/.github/actions/e2e/action.yml
+++ b/.github/actions/e2e/action.yml
@@ -5,7 +5,7 @@ description: Composite action to run Playwright against specified environment
 # This action conditionally installs Playwright browsers (Chrome, Firefox, WebKit)
 # based on the do-chrome-install, do-firefox-install, and do-webkit-install inputs.
 # Only the browsers needed for the current test shard are installed, improving CI efficiency.
-# At least one browser input must be set to 'true' for a valid test run.
+# Chrome is cached outside this job, and will be available even if not instructed to install.
 # ---
 
 inputs:
--- a/.github/actions/e2e/action.yml
+++ b/.github/actions/e2e/action.yml
@@ -1,6 +1,14 @@
 name: Run Playwright
 description: Composite action to run Playwright against specified environment
 
+# ---
+# This action conditionally installs Playwright browsers (Chrome, Firefox, WebKit)
+# based on the do-chrome-install, do-firefox-install, and do-webkit-install inputs.
+# Only the browsers needed for the current test shard are installed, improving CI efficiency.
+# At least one browser input must be set to 'true' for a valid test run.
+# ---
+
+
 inputs:
   version:
     description: "ref / sha / branch to run tests against"
--- a/.github/actions/e2e/action.yml
+++ b/.github/actions/e2e/action.yml
@@ -44,8 +44,14 @@ inputs:
   workers:
     description: "number of workers to use for test run"
     default: ""
-  force-webkit-install:
-    description: "webkit dependencies aren't all cached, so we need to force it to install"
+  do-chrome-install:
+    description: "install chrome for playwright"
+    default: "false"
+  do-firefox-install:
+    description: "install firefox for playwright"
+    default: "false"
+  do-webkit-install:
+    description: "install webkit for playwright"
     default: "false"
   staging_test_user_api_key:
     description: "test user API key used in spoofing staging user logins"
@@ -83,17 +89,30 @@ runs:
         cache: ${{ inputs.package_manager }}
         cache-dependency-path: ${{ inputs.lockfile_path }}
 
-    - name: Install Node dependencies & Playwright
+    - name: Install Playwright
       if: ${{ inputs.needs_node_setup == 'true' }}
       shell: bash
       working-directory: ./frontend
       run: |
         npm install @playwright/test
-        npx playwright install --with-deps
+
+    - name: Install Firefox and dependencies
+      if: ${{ inputs.do-firefox-install == 'true' }}
+      shell: bash
+      working-directory: ./frontend
+      run: |
+        npx playwright install --with-deps firefox
+
+    - name: Install Chrome and dependencies
+      if: ${{ inputs.do-chrome-install == 'true' }}
+      shell: bash
+      working-directory: ./frontend
+      run: |
+        npx playwright install --with-deps chrome
 
     # this should only be needed on shard 3, the WebKit shard
-    - name: Install Webkit dependencies
-      if: ${{ inputs.force-webkit-install == 'true' }}
+    - name: Install Webkit and dependencies
+      if: ${{ inputs.do-webkit-install == 'true' }}
       shell: bash
       working-directory: ./frontend
       run: |
--- a/.github/actions/e2e/action.yml
+++ b/.github/actions/e2e/action.yml
@@ -8,7 +8,6 @@ description: Composite action to run Playwright against specified environment
 # At least one browser input must be set to 'true' for a valid test run.
 # ---
 
-
 inputs:
   version:
     description: "ref / sha / branch to run tests against"
@@ -97,13 +96,14 @@ runs:
         cache: ${{ inputs.package_manager }}
         cache-dependency-path: ${{ inputs.lockfile_path }}
 
-    - name: Install Playwright
+    - name: Install Playwright core package
       if: ${{ inputs.needs_node_setup == 'true' }}
       shell: bash
       working-directory: ./frontend
       run: |
         npm install @playwright/test
 
+    # Conditionally install Chrome if requested
     - name: Install Chrome and dependencies
       if: ${{ inputs.do-chrome-install == 'true' }}
       shell: bash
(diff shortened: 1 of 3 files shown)


.github/workflows/ci-frontend-e2e.yml
```diff
@@ -429,8 +430,10 @@ jobs:
           staging_test_user_email: ${{ secrets.STAGING_TEST_USER_EMAIL }}
           staging_test_user_password: ${{ secrets.STAGING_TEST_USER_PASSWORD }}
           staging_test_user_mfa_key: ${{ secrets.STAGING_TEST_USER_MFA_KEY }}
+          # shard 2 is the webkit shard
```
Bhavna-Ramachandran: This comment needs to be updated to :  # shard 2 is the Firefox shard

.github/workflows/ci-frontend-e2e.yml
```diff
@@ -429,8 +430,10 @@ jobs:
...
+          do-firefox-install: ${{ matrix.shard == 2 && 'true' || 'false' }}
```
Bhavna-Ramachandran, line 432: The PR description mentions caching both Chrome and Firefox, but the caching step only installs Chrome. Is Firefox intentionally left out (installed fresh per-shard instead)?

.github/workflows/ci-frontend-e2e.yml
```diff
@@ -429,8 +430,10 @@ jobs:
...
           # shard 3 is the webkit shard
-          force-webkit-install: ${{ matrix.shard == 3 && 'true' || 'false' }}
+          do-webkit-install: ${{ matrix.shard == 3 && 'true' || 'false' }}
```
Bhavna-Ramachandran, line 434: Now that Firefox and WebKit are no longer included in the cache, the do-webkit-install step here will run the existing npx playwright install-deps webkit, which installs system dependencies but not the browser binary. Shard 3 will fail because the WebKit binary is never downloaded.

.github/actions/e2e/action.yml
Bhavna-Ramachandran: This needs to be changed to:  npx playwright install --with-deps webkit.

aoysimmons-GH, line 2:
```suggestion
description: Composite action to run Playwright against specified environment

# ---
# This action conditionally installs Playwright browsers (Chrome, Firefox, WebKit)
# based on the do-chrome-install, do-firefox-install, and do-webkit-install inputs.
# Only the browsers needed for the current test shard are installed, improving CI efficiency.
# At least one browser input must be set to 'true' for a valid test run.
# ---

```

aoysimmons-GH, line 77:
```suggestion
  steps:
      # --- Input validation: Ensure at least one browser is selected ---
    - name: Validate browser install inputs
      shell: bash
      run: |
        if [ "${{ inputs.do-chrome-install }}" != "true" ] && [ "${{ inputs.do-firefox-install }}" != "true" ] && [ "${{ inputs.do-webkit-install }}" != "true" ]; then
          echo "ERROR: At least one browser install input (do-chrome-install, do-firefox-install, do-webkit-install) must be set to 'true'."
          exit 1
        fi
    # --- End input validation ---
```

aoysimmons-GH, line 91:
```suggestion
    # Install Playwright core package
    - name: Install Playwright core
```

.github/actions/e2e/action.yml
```diff
@@ -83,17 +89,30 @@ runs:
...
 
-    - name: Install Node dependencies & Playwright
+    - name: Install Playwright
       if: ${{ inputs.needs_node_setup == 'true' }}
       shell: bash
       working-directory: ./frontend
       run: |
         npm install @playwright/test
-        npx playwright install --with-deps
+
```
aoysimmons-GH, line 105:
```suggestion
    # Conditionally install Firefox if requested
```

.github/actions/e2e/action.yml
```diff
@@ -83,17 +89,30 @@ runs:
...
+    - name: Install Firefox and dependencies
+      if: ${{ inputs.do-firefox-install == 'true' }}
+      shell: bash
+      working-directory: ./frontend
+      run: |
+        npx playwright install --with-deps firefox
+
```
aoysimmons-GH, line 113:
```suggestion
    # Conditionally install Chrome if requested
```

.github/actions/e2e/action.yml
```diff
@@ -83,17 +89,30 @@ runs:
...
+    - name: Install Chrome and dependencies
+      if: ${{ inputs.do-chrome-install == 'true' }}
+      shell: bash
+      working-directory: ./frontend
+      run: |
+        npx playwright install --with-deps chrome
 
     # this should only be needed on shard 3, the WebKit shard
```
aoysimmons-GH:
```suggestion
    # Conditionally install WebKit if requested (typically only needed on the WebKit shard)
```

aoysimmons-GH (commented):
@mdragon  I finished reviewing action.yml and added some suggestions—please take a look and see if they make sense.
I am now reviewing ci-frontend-e2e.yml. I think we can improve it by moving the browser install steps into the matrix job steps.

.github/workflows/ci-frontend-e2e.yml
aoysimmons-GH, line 235:
```suggestion
 include:
          - shard: 1
            browser: chrome
          - shard: 2
            browser: firefox
          - shard: 3
            browser: webkit
          - shard: 4
            browser: chrome
```

aoysimmons-GH, line 238:
```suggestion
    steps:
      # Install the required browser for this shard
      - run: npx playwright install --with-deps chrome
        if: matrix.browser == 'chrome'
      - run: npx playwright install --with-deps firefox
        if: matrix.browser == 'firefox'
      - run: npx playwright install --with-deps webkit
        if: matrix.browser == 'webkit'
```

.github/workflows/ci-frontend-e2e.yml
aoysimmons-GH, line 432:
```suggestion
          do-firefox-install: ${{ matrix.browser == 'firefox' && 'true' || 'false' }}
```

.github/workflows/ci-frontend-e2e.yml
aoysimmons-GH, line 434:
```suggestion
          do-webkit-install: ${{ matrix.browser == 'webkit' && 'true' || 'false' }}
```

.github/workflows/ci-frontend-e2e.yml
```diff
@@ -146,10 +146,11 @@ jobs:
         run: |
           npm ci
 
-      - run: npx playwright install --with-deps
-        if: steps.playwright-cache.outputs.cache-hit != 'true'
-      - run: npx playwright install-deps
+      # 2 of the 4 shards need chrome, so cache that, install the others as needed
```
aoysimmons-GH:
```suggestion
      # Install Chrome only for caching; other browsers installed per-shard in test job
```

aoysimmons-GH (commented):
@mdragon  I added a suggestion for ci-frontend-e2e.yml regarding moving the browser install steps into the matrix job steps.

.github/actions/e2e/action.yml
mdragon: This isn't strictly true as we cache chrome, and I'll work on caching FF, so you don't have to do the installs (except for Webkit which will fail if you don't install it and it's the Webkit shard.

.github/workflows/ci-frontend-e2e.yml
mdragon: These don't seem to work, which I think makes sense because the matrix only has access to what you pass in? So left this shard based.

.github/workflows/ci-frontend-e2e.yml
mdragon: Oh, sorry didn't see this first. Since the shard browsers are actually controlled by the config file, I'd rather not also encode this here. If we ever change the browsers per shard, stuff will fail because it won't have the right browser installed so I'm not too worried about making this code more obvious about what shard is what browser.

mdragon: This is happening in the e2e/action and I think it has to, because the container running this part, isn't the same as what will run that? I'm actually not clear on that, but I think having it in the e2e/action makes more sense since we call that from Staging and from here.

.github/actions/e2e/action.yml
```diff
@@ -1,6 +1,13 @@
 name: Run Playwright
 description: Composite action to run Playwright against specified environment
 
+# ---
+# This action conditionally installs Playwright browsers (Chrome, Firefox, WebKit)
+# based on the do-chrome-install, do-firefox-install, and do-webkit-install inputs.
+# Only the browsers needed for the current test shard are installed, improving CI efficiency.
+# At least one browser input must be set to 'true' for a valid test run.
```
Bhavna-Ramachandran: Can we remove this line, as this does not factually reflect the set up here.

Bhavna-Ramachandran (approved):

Issue #10438, opened 2026-05-14
List of opportunities not displaying in Training env


brianlee-nava:
### Summary

@mdragon and I troubleshooted this live and could not figure out why the list of opportunities is not showing in Training.

### Reproducibility

I tried multiple times and saw the bug every time

### How to reproduce the bug

**Steps to reproduce**

1. Log in to training
2. Navigate to https://training.simpler.grants.gov/opportunities?_ff=opportunitiesListOff%3Afalse
  - If agency dropdown exists for you, select Simpler Agency
3. Check to see if list of opportunities displays

**Expected behavior**

1. List of opportunities is shown

**Current behavior**
1. Frontend error: We have encountered an error loading your opportunities, please try again later.
2. 

<img width="2922" height="936" alt="Image" src="https://github.com/user-attachments/assets/bdd612fb-a003-4ba9-810d-6a84b8014cce" />

When run from API docs, data is returning empty with a 200 status code.

Note: Making API request directly succeeds, but finds no records.

### Browser

Chrome

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Dao-REI:
Thank you, @mdragon for doing the troubleshooting. Cause: we had removed the limit on the number of records returned for this API since the frontend had not implemented pagination yet (ticket #10323). We only have 221 opportunities in Training but it exceeded the 10MB APIGateway limit with a response of 111Mb (over 500kb per opportunity). The page only shows 3-4 fields for each opportunity, but the API is returning everything (all data) for each opportunity. Suggested fixes:

1.  Implement pagination for both the frontend and backend.
2.  Only return the fields needed for that page. This will also speed up performance greatly.

PR #10511, opened 2026-05-20
[Issue #10438] Add logging to help troubleshooting Grantor Opportunities List

mdragon:
## Summary

Work for #10438

## Changes proposed

Log additional information (server-side FE logs) when we have issues with token parsing or data fetching on the new Opportunities List page. 

## Context for reviewers

This was driven by lack of information available to troubleshoot why this page wasn't loading correctly in Training. Ultimately that was due to the size of the data response causing it to get eaten by API Gateway in between the FE and the BE but until more of this logging was added we couldn't determine what was happening.

## Validation steps

- This branch was deployed to Training to help troubleshoot the issues we were having at the time, validating the new logging was either seen or it not being seen helped narrow the root cause.
- Tests pass


frontend/src/services/fetch/fetchers/fetchers.ts
```diff
@@ -106,7 +107,16 @@ export function requesterForEndpoint({
           `bad Json from error response at ${url} with status code ${response.status}`,
         );
       }
-      return throwError(jsonBody, url);
+      console.error(
```
doug-s-nava: this is potentially confusing, given that throwError will log a different error message immediately after. can we consolidate those into one log?

frontend/src/services/auth/session.ts
```diff
@@ -81,6 +81,7 @@ export const getSession = async (): Promise<UserSession | null> => {
   if (!clientSessionToken) return null;
   const payload = await decryptClientToken(clientSessionToken);
   if (!payload) {
+    console.error("No session payload");
```
doug-s-nava: don't think we need this extra log - this should be logging already here https://github.com/HHS/simpler-grants-gov/blob/3ecb1e954de5a1733f6ae912cd57cfd8f6c1e292/frontend/src/services/auth/sessionUtils.ts#L38

frontend/src/services/fetch/fetchers/fetchers.ts
mdragon: Looking at this.

mdragon: @doug-s-nava Addressed this.

doug-s-nava (approved):
✅ ✅ ✅ ✅

PR #10514, opened 2026-05-20
[Issue #10352] CREATE Endpoint for Competitions

### LINKED ISSUE #10352: CREATE endpoint for competitions
### Summary

Create a POST endpoint at `/v1/grantors/opportunities/:opportunity_id/competitions` to create a new competition. API Spec [here](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/3027992702/Tech+Spec+Quad+6+Enable+Grantors+to+Post+Application+Packages+With+Opportunities#API-%2F-Integration-Details)

Note that all of the provided fields should be nullable.

### Acceptance criteria

- [ ] A user with update_opportunity permission can create a new competition

dghazvini:
## Summary

Fixes / Work for #10352 

## Changes proposed

Create a competition to be associated with the opportunity.
~-Added blueprint/schema/route for Competition v1 and the Create Competition Endpoint.~
-Added schema/route for competition creation under the Opportunities Grantor v1 blueprint at `POST /v1/grantors/opportunities/:opportunity_id/competitions`
-Created `competition_creation` service in opportunities_grantor_v1 module
~-Added a `get_opportunity_assistance_listing` function under the Grantor Opportunities Services.~
~-Added feature flag for the new Competition v1 blueprint.~
-Added unit tests.

## Context for reviewers

API Tech Spec [here](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/3027992702/Tech+Spec+Quad+6+Enable+Grantors+to+Post+Application+Packages+With+Opportunities#API-%2F-Integration-Details)

## Validation steps

- [ ] ~Data populating in local DB in the `competition`, `competition_instruction`, and `link_competition_open_to_applicant` tables.~
- [x] Data populating in local DB in `competition` and `link_competition_open_to_applicant` tables
- [x] Manual API testing confirmed correct response structure
- [x] 9 Unit Tests Passing


api/src/api/competitions_v1/competition_schemas.py
```diff
@@ -0,0 +1,85 @@
...
+            "example": "competition_instructions.pdf",
+        },
+    )
+    download_path = fields.String(
+        required=True,
+        metadata={
+            "description": "The URL/path to the instruction file",
+            "example": "s3://simpler-grants-gov-dev/competition-instructions/file.pdf",
+        },
+    )
```
chouinar: A user can't provide an s3 path, users don't have access or any awareness of s3 in our system, you can't ask for a path, files have to be uploaded in their own dedicated endpoint.

api/src/services/competitions_v1/competition_creation.py
```diff
@@ -0,0 +1,111 @@
...
+
+    # Create competition instructions
+    for instruction_data in request.competition_instructions:
+        instruction = CompetitionInstruction(
+            competition_instruction_id=uuid.uuid4(),
+            competition_id=competition.competition_id,
+            file_name=instruction_data.file_name,
+            file_location=instruction_data.download_path,
+        )
+        db_session.add(instruction)
```
chouinar: Anything with file upload needs its own separate endpoint. Don't try to add competition instructions as part of this endpoint.

File upload is fundamentally changing soon (a few sprints and platform will have a different approach setup to use). But if you want a pattern to follow, look at the application attachment route for an example of how we do it right now.

api/src/api/competitions_v1/competition_schemas.py
```diff
@@ -0,0 +1,85 @@
...
+    is_simpler_grants_enabled = fields.Boolean(
+        required=True, metadata={"description": "Whether simpler grants are enabled"}
+    )
+    open_to_applicants = fields.List(
+        fields.Enum(CompetitionOpenToApplicant),
+        required=True,
+        validate=validators.Length(min=1),
+        metadata={
+            "description": "List of applicant types eligible for this competition",
+            "example": ["individual", "organization"],
```
chouinar:
```suggestion
            "example": [CompetitionOpenToApplicant.APPLICANT, CompetitionOpenToApplicant.ORGANIZATION],
```

api/src/api/competitions_v1/competition_schemas.py
```diff
@@ -0,0 +1,85 @@
...
+            "example": "Proposal for Advanced Research",
+        },
+    )
+    opening_date = fields.Date(
+        required=True,
+        metadata={"description": "The opening date of the competition", "example": "2026-05-18"},
+    )
+    closing_date = fields.Date(
+        required=True,
+        metadata={"description": "The closing date of the competition", "example": "2026-06-18"},
```
chouinar: All of these are technically nullable, although not sure we want title to be.

A null opening date just means it's always open.

api/src/api/competitions_v1/competition_schemas.py
```diff
@@ -0,0 +1,85 @@
...
+    contact_info = fields.String(
+        required=True,
+        metadata={
+            "description": "Contact information for the competition",
+            "example": "Bob Smith\nFakeMail@fake.com",
+        },
+    )
+    is_simpler_grants_enabled = fields.Boolean(
+        required=True, metadata={"description": "Whether simpler grants are enabled"}
+    )
```
chouinar: This shouldn't be in the request schema, we should just set this to True for anything we create. This is needed as we have to manually setup anything we pulled from grants.gov

api/src/api/competitions_v1/competition_schemas.py
```diff
@@ -0,0 +1,85 @@
...
+        metadata={"description": "List of instruction files"},
+    )
+    opportunity_assistance_listing = fields.Nested(
+        OpportunityAssistanceListingV1Schema,
+        required=True,
+        metadata={
+            "description": "Assistance listing information",
+            "example": {"assistance_listing_number": "43.012", "program_title": "Space Technology"},
+        },
+    )
```
chouinar: This probably should just be the ID of the assistance listing, which would mean the frontend needs to fetch these values first to display to the user. Might need to talk with design about how we want it, but should definitely not be the values, should be an ID.

api/src/api/competitions_v1/competition_schemas.py
```diff
@@ -0,0 +1,85 @@
...
+            "example": "s3://simpler-grants-gov-dev/competition-instructions/file.pdf",
+        },
+    )
+
+
+class CompetitionCreateRequestSchema(Schema):
+    """Schema for POST /v1/competitions/ request"""
+
+    opportunity_id = fields.UUID(
+        required=True, metadata={"description": "The opportunity ID this competition belongs to"}
```
chouinar: Rather than pass in the opportunity ID here, what if the request was `POST .../opportunities/:opportunity_id/competitions` rather than it being a competition specific route. That's what we've generally done as a competition lives within an opportunity

api/src/services/competitions_v1/competition_creation.py
```diff
@@ -0,0 +1,111 @@
...
+    stmt = (
+        select(Competition)
+        .options(
+            selectinload(Competition.competition_forms),
+            selectinload(Competition.competition_instructions),
+            selectinload(Competition.opportunity_assistance_listing),
+            selectinload(Competition.link_competition_open_to_applicant),
+        )
+        .where(Competition.competition_id == competition.competition_id)
+    )
```
chouinar: This shouldn't be necessary.

When you create each of the parts, just do:

```py
competition = Competition(...)

competition.link_competition_open_to_applicant = set(...)
competition.xyz = abc
```

Then they'll all be set.

chouinar (changes requested):
Competition instructions can't be in this route, file upload is always a separate endpoint.

api/src/api/competitions_v1/competition_schemas.py
dghazvini: Removed the `competition_instructions` field entirely from the create endpoint based on the updated tech spec. Will be handled in a separate endpoint in the future.

api/src/services/competitions_v1/competition_creation.py
dghazvini: Removed the competition instructions logic entirely from this endpoint, will be handled in a separate file upload endpoint

api/src/api/competitions_v1/competition_schemas.py
dghazvini: Updated to use enum values

api/src/api/competitions_v1/competition_schemas.py
dghazvini: Removed this field from the request schema and set `is_simpler_grants_enabled=True` in the service layer.

api/src/api/competitions_v1/competition_schemas.py
dghazvini: Removed the `opportunity_assistance_listing` field from the request entirely based on the updated tech spec.

api/src/api/competitions_v1/competition_schemas.py
dghazvini: Endpoint has been updated to reflect this based on the updated tech spec.

api/src/api/competitions_v1/competition_schemas.py
dghazvini: For `opening_date` and `closing_date`, set `required=False` and `allow_none=True`. Added a `test_competition_create_with_null_dates` test as well.

api/src/services/competitions_v1/competition_creation.py
dghazvini: Remove the stmt and added this block to avoid any lazy load errors:

```
    # Explicitly initialize all relationships that will be serialized
    competition.open_to_applicants = set(request.open_to_applicants)
    competition.competition_forms = []
    competition.competition_instructions = []
    competition.opportunity_assistance_listing = None
```

dghazvini:
Made significant updates based on the updated tech spec and updated the original ticket text to reflect that.

Sample request/response:

Request
```
{
  "closing_date": "2026-05-11",
  "competition_title": "Proposal for Advanced Research",
  "contact_info": "Bob Smith\nFakeMail@fake.com",
  "open_to_applicants": [
    "individual",
    "organization"
  ],
  "opening_date": "2026-05-11"
}
```


Response
```
{
  "data": {
    "closing_date": "2026-05-11",
    "competition_forms": [],
    "competition_id": "7ace8a93-9910-4aa0-bee0-a432f834ed7e",
    "competition_instructions": [],
    "competition_title": "Proposal for Advanced Research",
    "contact_info": "Bob Smith\nFakeMail@fake.com",
    "is_open": false,
    "is_simpler_grants_enabled": true,
    "open_to_applicants": [
      "individual",
      "organization"
    ],
    "opening_date": "2026-05-11",
    "opportunity_assistance_listing": null,
    "opportunity_id": "0b3439be-27d2-4fa6-aa03-8d81230e7b59"
  },
  "message": "Success",
  "status_code": 200
}
```

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
```diff
@@ -621,3 +622,50 @@ class DeleteAttachmentResponseV1Schema(ResponseWithErrorsSchema):
...
+    )
+    opening_date = fields.Date(
+        required=False,
+        allow_none=True,
+        metadata={"description": "The opening date of the competition", "example": "2026-05-11"},
+    )
+    closing_date = fields.Date(
+        required=False,
+        allow_none=True,
+        metadata={"description": "The closing date of the competition", "example": "2026-05-11"},
```
chouinar, line 645: I would defer to design, but I these should both be required, but keep allow_none. If someone wants to set these to null, they should explicitly do that.

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
```diff
@@ -621,3 +622,50 @@ class DeleteAttachmentResponseV1Schema(ResponseWithErrorsSchema):
...
+    )
+    contact_info = fields.String(
+        required=True,
+        metadata={
+            "description": "Contact information for the competition",
+            "example": "Bob Smith\nFakeMail@fake.com",
+        },
+    )
```
chouinar, line 653: Don't need to solve here, but since these usually have some amount of formatting, we should make sure the frontend is setup to allow for that.

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
```diff
@@ -621,3 +622,50 @@ class DeleteAttachmentResponseV1Schema(ResponseWithErrorsSchema):
 
 class OpportunityPublishResponseV1Schema(AbstractResponseSchema):
     data = fields.Nested(OpportunityGrantorSchema())
+
+
+class CompetitionCreateRequestSchema(Schema):
```
chouinar, line 627: I don't see a grace period field, is that intentional (cc @glenblosser-nava) ? It's a commonly used field, did some research recently for design.

api/src/api/opportunities_grantor_v1/opportunity_grantor_routes.py
```diff
@@ -262,3 +263,28 @@ def opportunity_publish(db_session: db.Session, opportunity_id: UUID) -> respons
...
+    """Create a new competition for an opportunity"""
+    add_extra_data_to_current_request_logs({"opportunity_id": opportunity_id})
+    logger.info(f"POST /v1/grantors/opportunities/{opportunity_id}/competitions")
+
+    with db_session.begin():
+        user = jwt_or_api_user_key_multi_auth.get_user()
+        db_session.add(user)
+
+        json_data["opportunity_id"] = opportunity_id
+        competition = create_competition(db_session, user, json_data)
```
chouinar: Pass the opportunity ID in separately, don't alter the request object.

api/src/services/opportunities_grantor_v1/competition_creation.py
```diff
@@ -0,0 +1,64 @@
...
+    # Create the competition
+    competition = Competition(
+        competition_id=uuid.uuid4(),
+        opportunity_id=request.opportunity_id,
+        competition_title=request.competition_title,
+        opening_date=request.opening_date,
+        closing_date=request.closing_date,
+        contact_info=request.contact_info,
+        is_simpler_grants_enabled=True,
+    )
```
chouinar, line 31: You could do this if you wanted to trim a bit of this out.

```suggestion
def create_competition(db_session: db.Session, user: User, opportunity_id: uuid.UUID, competition_data: dict) -> Competition:

    # Check if opportunity exists
    opportunity = get_opportunity_for_grantors(db_session, user, opportunity_id)

    # Check if user has permission to update opportunities for this agency
    verify_access(user, {Privilege.UPDATE_OPPORTUNITY}, opportunity.agency_record)

    # Create the competition
    competition = Competition(
        competition_id=uuid.uuid4(),
        is_simpler_grants_enabled=True,
        **competition_data
    )
```

We usually make them into pydantic if we're going to interact with the fields a bit as a typed-object is more convenient, but if we're just shoving them into the competition, we can skip it.

api/src/services/opportunities_grantor_v1/competition_creation.py
```diff
@@ -0,0 +1,64 @@
...
+
+    # Explicitly initialize all relationships that will be serialized
+    competition.open_to_applicants = set(request.open_to_applicants)
+    competition.competition_forms = []
+    competition.competition_instructions = []
+    competition.opportunity_assistance_listing = None
```
chouinar: How is this going to be set if not from this endpoint?

api/src/services/opportunities_grantor_v1/competition_creation.py
dghazvini: Removed the pydantic model and simplified the competition creation.

api/src/api/opportunities_grantor_v1/opportunity_grantor_routes.py
dghazvini: Updated to pass in opportunity ID

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
dghazvini: Updated `required` to True.

api/src/services/opportunities_grantor_v1/competition_creation.py
dghazvini: @chouinar My understanding was that both `competition_forms` and `competition_instructions` would be set in separate file upload endpoints. `opportunity_assistance_listing` I can see might want to be set here, but in that case we'd need to keep assistance listing number in the request and add a get_opportunity_assistance_listing function, no? I previously removed it since we removed ALN from the request. CC @glenblosser-nava

glenblosser-nava: Yes, we will need to create the competition first with empty forms and instructions, and then add them in later to preserve the ability to save incomplete work on a competition.

api/src/api/opportunities_grantor_v1/opportunity_grantor_schemas.py
glenblosser-nava: The most recent designs for the competition creation page don't include a grace period. @crystabelrangel indicated that it's a more complicated topic and we won't be handling it for now.

chouinar: We could include it in the API even if the frontend never sets it for completeness.

chouinar (approved):

PR #10525, opened 2026-05-21
[Issue #10020] Add job lock context manager

### LINKED ISSUE #10020: Build the ECS task job lock logic
### Summary

Information and a prototype you can start from can be found on: 
https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/3026255873/Tech+Spec+Backend+ECS+Task+Job+Locks

A few additional notes:
* We should have an environment variable of `ENABLE_JOB_LOCK` that if set to False makes the job lock logic automatically pass through (The entry/exit functions should just instantly return).
* Consider how we might generify how we interact with the database, with upcoming data separation work, we will want this usable in both systems. Separate any DB interactions out to be overrideable.

### Acceptance criteria

- [ ] Job lock created and tested (will likely need a bit of manual testing too)
- [ ] No jobs actually have the job lock yet
- [ ] An environment variable exists that can disable the job lock logic entirely

jakobpederson:
## Summary

Fixes / Work for #10020  

## Changes proposed

- Added `TaskJobLock` context manager
- Added tests for `TaskJobLock`
- Added `ENABLE_JOB_LOCK` env variable

## Context for reviewers

Want to add locks to prevent tasks from running over each other.

## Validation steps

- [ ] tests passing
- [ ] jobs do not run if job type is currently locked


api/tests/src/task/test_task_job_lock.py
```diff
@@ -0,0 +1,231 @@
...
+    with TaskJobLock(db_session, job_type=JobType.SETUP_CERT_USER, lock_duration_minutes=60):
+        SetupCertUserTask(db_session, tcert_id, [str(role.role_id)]).setup_cert()
+    job_lock = (
+        db_session.query(JobLock).where(JobLock.job_type == JobType.SETUP_CERT_USER).one_or_none()
+    )
+    assert job_lock is not None
+    assert job_lock.job_type == JobType.SETUP_CERT_USER
+
+
+def test_task_job_lock_updates_a_job_lock_if_enable_job_lock_is_false(
```
kkrug:
```suggestion
def test_task_job_lock_updates_a_job_lock_if_enable_job_lock_is_true(
```

api/src/task/task_job_lock.py
```diff
@@ -0,0 +1,127 @@
...
+
+        with self.db_session.begin_nested():
+            job_lock = self.get_job_lock()
+
+            if job_lock is None:
+                job_lock = JobLock(job_type=self.job_type, is_locked=False)
+
+            now = datetime_util.utcnow()
+            if job_lock.is_locked and job_lock.locked_until > now:
+                logger.error("Job is currently locked", extra=self.extra)
```
kkrug: Should we use `logger.warning` here instead since this is expected logic?

api/src/task/task_job_lock.py
```diff
@@ -0,0 +1,127 @@
...
+    def __enter__(self) -> None:
+        logger.info("Entering the lock", extra=self.extra)
+        if not self.config.enable_job_lock:
+            return
+
+        with self.db_session.begin_nested():
+            job_lock = self.get_job_lock()
+
+            if job_lock is None:
+                job_lock = JobLock(job_type=self.job_type, is_locked=False)
```
chouinar: I'd just move this if check into the `get_job_lock` function. That way it always gets a lock, even if it's brand new.

A secondary reason why, we'll soon need to generify this logic a bit and make it usable by more than one service. They won't be able to directly use a specific DB table object, so it'll serve as an interface for handling all of that.

api/src/task/task_job_lock.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+@cache
+def get_task_job_lock_config() -> TaskJobLockConfig:
+    return TaskJobLockConfig()
+
+
+class TaskJobLock(contextlib.AbstractContextManager[None]):
+
+    def __init__(
```
chouinar, line 39: Could we add some docs to this and include an example of how it is used?

api/src/task/task_job_lock.py
```diff
@@ -0,0 +1,127 @@
...
+            if job_lock.is_locked and job_lock.locked_until > now:
+                logger.error("Job is currently locked", extra=self.extra)
+                raise TaskJobLockIsLockedError
+
+            job_lock.locked_until = datetime_util.utcnow() + timedelta(
+                minutes=self.lock_duration_minutes
+            )
+            job_lock.is_locked = True
+            job_lock.locked_by = self.internal_lock_id
+            self.db_session.add(job_lock)
```
chouinar: We need a secondary check after this logic to verify that the lock was acquired by this process. If two processes both come into this at the same time, we need to make certain it got the lock.

api/src/task/task_job_lock.py
```diff
@@ -0,0 +1,127 @@
...
+                    logger.error(
+                        "JobLock ids do not match",
+                        extra=updated_extra,
+                    )
+                    raise TaskJobLockInternalIDError
+
+                job_lock.is_locked = False
+                self.db_session.add(job_lock)
+        except Exception as e:
+            logger.exception("Failed to free the job lock", extra=self.extra)
```
kkrug: These two paths in the `try` block both have a `logger.error`, plus the `logger.exception`, so there will be two distinct error logs for the same event.

api/src/task/task_job_lock.py
```diff
@@ -0,0 +1,127 @@
...
+            "internal_lock_id": self.internal_lock_id,
+            "job_locked_enabled": self.config.enable_job_lock,
+        }
+
+    def __enter__(self) -> None:
+        logger.info("Entering the lock", extra=self.extra)
+        if not self.config.enable_job_lock:
+            return
+
+        with self.db_session.begin_nested():
```
chouinar: This shouldn't be nested, it needs to just be its own transaction. If you're hitting a nesting issue with tests, we probably need to adjust the test setup.

api/tests/src/task/test_task_job_lock.py
```diff
@@ -0,0 +1,231 @@
...
+):
+    get_task_job_lock_config.cache_clear()
+    monkeypatch_session.setenv("ENABLE_JOB_LOCK", "False")
+    caplog.set_level(logging.INFO)
+    db_session.query(JobLock).filter_by(job_type=JobType.SETUP_CERT_USER).delete()
+    db_session.flush()
+    role = RoleFactory.create(is_agency_role=True)
+    agency = AgencyFactory(agency_code=f"XYZ-{uuid.uuid4()}", is_multilevel_agency=False)
+    tcertificate = StagingTcertificatesFactory(agencyid=agency.agency_code)
+    tcert_id = str(tcertificate.tcertificates_id)
```
chouinar: Don't use the cert job.

The job locks aren't even connected to an actual task, so I'd argue we should just ignore them entirely for these tests for simplicity.

api/tests/src/task/test_task_job_lock.py
```diff
@@ -0,0 +1,231 @@
...
+
+def get_or_create_job_lock(db_session, update_date=None, locked_until_date=None, is_locked=None):
+    is_locked = is_locked or False
+    if update_date:
+        updated_at = update_date
+    else:
+        now = datetime_util.utcnow()
+        updated_at = now - timedelta(days=366)
+    job_lock = (
+        db_session.query(JobLock).where(JobLock.job_type == JobType.SETUP_CERT_USER).one_or_none()
```
kkrug: Let's change all of the `db_session.query` instances to use `db_session.execute(select(...)) instead
```suggestion
        db_session.execute(select((JobLock).where(JobLock.job_type == JobType.SETUP_CERT_USER).one_or_none())
```

api/src/task/task_job_lock.py
```diff
@@ -0,0 +1,127 @@
...
+                raise TaskJobLockIsLockedError
+
+            job_lock.locked_until = datetime_util.utcnow() + timedelta(
+                minutes=self.lock_duration_minutes
+            )
+            job_lock.is_locked = True
+            job_lock.locked_by = self.internal_lock_id
+            self.db_session.add(job_lock)
+
+    def __exit__(self, exc_type: Any, exc_value: Any, traceback: Any) -> None:
```
chouinar, line 103: We need to adjust the logging to satisfy what I outlined in the tech spec
https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/3026255873/Tech+Spec+Backend+ECS+Task+Job+Locks#Metrics-%2F-Monitoring

You've got job type added to all the logs, but we should add:
* success/error
* Some sort of "what the error was" as a field
* How long the lock was held for here in the exit function

api/src/task/task_job_lock.py
chouinar: Can we add info to this error message about the job that is locking it?

api/tests/src/task/test_task_job_lock.py
```diff
@@ -0,0 +1,231 @@
...
+        SetupCertUserTask(db_session, tcert_id, [str(role.role_id)]).setup_cert()
+    db_session.refresh(job_lock)
+    assert job_lock.updated_at == old_date
+
+
+def test_task_job_lock_creates_job_lock_if_one_does_not_exist(
+    caplog, monkeypatch_session, db_session, enable_factory_create
+):
+    get_task_job_lock_config.cache_clear()
+    monkeypatch_session.setenv("ENABLE_JOB_LOCK", "True")
```
chouinar: I'd argue that we get rid of the cache entirely. In a given job run, we'll at most fetch that config once. All it's adding is extra complexity to tests.

We should also add `ENABLE_JOB_LOCK` to local.env as true so the tests don't need to do this.

api/src/task/task_job_lock.py
```diff
@@ -0,0 +1,127 @@
...
+class TaskJobLockConfig(PydanticBaseEnvConfig):
+    enable_job_lock: bool = Field(alias="ENABLE_JOB_LOCK")
+
+
+@cache
+def get_task_job_lock_config() -> TaskJobLockConfig:
+    return TaskJobLockConfig()
+
+
+class TaskJobLock(contextlib.AbstractContextManager[None]):
```
kkrug, line 37: Per the ticket, we should separate out the DB interactions so we can reuse it. ("Consider how we might generify how we interact with the database, with upcoming data separation work, we will want this usable in both systems. Separate any DB interactions out to be overrideable.")

api/src/task/task_job_lock.py
```diff
@@ -0,0 +1,127 @@
...
+class TaskJobLockInternalIDError(Exception):
+    pass
+
+
+class TaskJobLockNotFoundError(Exception):
+    pass
+
+
+class TaskJobLockConfig(PydanticBaseEnvConfig):
+    enable_job_lock: bool = Field(alias="ENABLE_JOB_LOCK")
```
chouinar: The alias you gave is identical. Give it a default so we don't need to configure every env ahead of time.

```suggestion
    enable_job_lock: bool = False
```

infra/api/app-config/env-config/environment_variables.tf
```diff
@@ -124,6 +124,11 @@ locals {
       secret_store_name = "/api/${var.environment}/enable-simpler-route"
     }
 
+    ENABLE_JOB_LOCK = {
+      manage_method     = "manual"
+      secret_store_name = "/api/${var.environment}/enable-job-lock"
+    }
```
chouinar: We don't generally add our feature flags as secrets on the backend as we don't need to change them at a moments notice and have ways of overriding them for a given job run if needed.

Can this just be set like our other non-secret env vars?

api/tests/src/task/test_task_job_lock.py
```diff
@@ -0,0 +1,231 @@
...
+    monkeypatch_session.setenv("ENABLE_JOB_LOCK", "False")
+    caplog.set_level(logging.INFO)
+    db_session.query(JobLock).filter_by(job_type=JobType.SETUP_CERT_USER).delete()
+    db_session.flush()
+    role = RoleFactory.create(is_agency_role=True)
+    agency = AgencyFactory(agency_code=f"XYZ-{uuid.uuid4()}", is_multilevel_agency=False)
+    tcertificate = StagingTcertificatesFactory(agencyid=agency.agency_code)
+    tcert_id = str(tcertificate.tcertificates_id)
+    with TaskJobLock(db_session, job_type=JobType.SETUP_CERT_USER, lock_duration_minutes=60):
+        SetupCertUserTask(db_session, tcert_id, [str(role.role_id)]).setup_cert()
```
kkrug: I don't know that calling `SetupCertUserTask()` is necessary. To strictly test the lock context manager, we could probably `pass` within this block or assert the lock state directly. Same with the other instances below.

api/tests/src/task/test_task_job_lock.py
kkrug: There's a lot of duplicated code here that could be broken out into a fixture

api/tests/src/task/test_task_job_lock.py
```diff
@@ -0,0 +1,231 @@
...
+
+    with pytest.raises(TaskJobLockError) as excinfo:  # noqa
+        with TaskJobLock(db_session, job_type=JobType.SETUP_CERT_USER, lock_duration_minutes=60):
+            job_lock = get_or_create_job_lock(db_session, update_date=old_date)
+            job_lock.locked_by = uuid.uuid4()
+            db_session.flush()
+
+    assert isinstance(excinfo.value.__cause__, TaskJobLockInternalIDError)
+    assert "JobLock ids do not match" in caplog.messages
+    assert "Failed to free the job lock" in caplog.messages
```
chouinar: You're doing a lot here in these tests that can be simplified.

I don't think we should be checking log messages to find paths in code that happened, just rely on the error type since you created separate ones for each scenario anyways.

kkrug: Instead of checking this via string matching, we should iterate over `caplog.records` and assert whatever is needed (message, levelname, job_type, etc.). Same in the test below.

api/tests/src/task/test_task_job_lock.py
```diff
@@ -0,0 +1,231 @@
...
+)
+from tests.src.db.models.factories import (
+    AgencyFactory,
+    JobLockFactory,
+    RoleFactory,
+    StagingTcertificatesFactory,
+)
+
+
+def get_or_create_job_lock(db_session, update_date=None, locked_until_date=None, is_locked=None):
```
kkrug: This helper queries for an existing lock and updates it, which couples the test to whatever state previous tests left. Each test can create exactly what it needs via `JobLockFactory.create(...)`. The factory already supports these args, and it makes test ordering irrelevant.

api/src/task/task_job_lock.py
```diff
@@ -0,0 +1,127 @@
...
+    ) -> None:
+        self.db_session = db_session
+        self.job_type = job_type
+        self.lock_duration_minutes = lock_duration_minutes
+        self.internal_lock_id = uuid.uuid4()
+        self.config = get_task_job_lock_config()
+        self.extra = {
+            "job_type": self.job_type,
+            "internal_lock_id": self.internal_lock_id,
+            "job_locked_enabled": self.config.enable_job_lock,
```
kkrug:
```suggestion
            "job_lock_enabled": self.config.enable_job_lock,
```

kkrug (commented):
Should we be adding `ENABLE_JOB_LOCK=TRUE` to local.env as well?

api/tests/src/task/test_task_job_lock.py
jakobpederson: I got it reversed in my head!

api/src/task/task_job_lock.py
jakobpederson: made the two errors into info.

api/tests/src/task/test_task_job_lock.py
jakobpederson: Just pulling the log messages per Michael's comment.

api/tests/src/task/test_task_job_lock.py
```diff
@@ -0,0 +1,183 @@
...
+    TaskJobLockNotFoundError,
+)
+from tests.src.db.models.factories import JobLockFactory
+
+
+def test_task_job_lock__does_not_create_a_job_lock_if_enable_job_lock_is_false(
+    monkeypatch_session, db_session, enable_factory_create
+):
+    monkeypatch_session.setenv("ENABLE_JOB_LOCK", "False")
+    db_session.query(JobLock).filter_by(job_type=JobType.SETUP_CERT_USER).delete()
```
kkrug: We should switch all the instances of `db_session.query` to `db_session.execute(select(...))`. There are a few in this file. 

Additionally, is this query meant to clean up from prior tests? If so, it would probably be better to have a fixture like `clean_job_locks` that deletes the row in teardown.

api/src/task/task_job_lock.py
```diff
@@ -0,0 +1,201 @@
...
+
+        try:
+            with self.db_session.begin():
+                job_lock = self.get_job_lock()
+
+                if not job_lock:
+                    logger.info(
+                        "JobLock not found",
+                        extra={**self.extra, "success": False, "error": "TaskJobLockNotfoundError"},
+                    )
```
kkrug: Do we need this log statement at all? The outer exception will already have `TaskJobLockNotFoundError`, so the same event will be logged twice.

api/tests/src/task/test_task_job_lock.py
```diff
@@ -0,0 +1,183 @@
...
+def test_task_job_lock_updates_a_job_lock_if_enable_job_lock_is_true(
+    caplog, monkeypatch_session, db_session, enable_factory_create
+):
+    monkeypatch_session.setenv("ENABLE_JOB_LOCK", "True")
+    caplog.set_level(logging.INFO)
+    now = datetime_util.utcnow()
+    old_date = now - timedelta(days=366)
+    db_session.query(JobLock).filter_by(job_type=JobType.SETUP_CERT_USER).delete()
+    job_lock = JobLockFactory.create(job_type=JobType.SETUP_CERT_USER, updated_at=old_date)
+    db_session.commit()
```
kkrug: Could we just configure the factory to commit instead of calling `db_session.commit()` in all of these tests?

api/src/task/task_job_lock.py
```diff
@@ -0,0 +1,201 @@
...
+            return
+
+        try:
+            with self.db_session.begin():
+                job_lock = self.get_job_lock()
+
+                if not job_lock:
+                    logger.info(
+                        "JobLock not found",
+                        extra={**self.extra, "success": False, "error": "TaskJobLockNotfoundError"},
```
kkrug:
```suggestion
                        extra={**self.extra, "success": False, "error": "TaskJobLockNotFoundError"},
```

api/src/task/task_job_lock.py
```diff
@@ -0,0 +1,201 @@
...
+        )
+        job_lock.is_locked = True
+        job_lock.locked_by = self.internal_lock_id
+        self.add_job_lock(job_lock)
+        return job_lock
+
+    def set_job_lock_to_unlocked(self, job_lock: JobLock) -> JobLock:
+        job_lock.is_locked = False
+        self.add_job_lock(job_lock)
+        return job_lock
```
kkrug, line 191: Do we need to return `job_lock` here? It doesn't look like it gets used

api/tests/src/task/test_task_job_lock.py
```diff
@@ -0,0 +1,183 @@
...
+
+def test_job_lock_raises_when_lock_deleted_between_enter_and_exit(
+    caplog, db_session, enable_factory_create
+):
+    now = datetime_util.utcnow()
+    old_date = now - timedelta(days=366)
+    db_session.query(JobLock).filter_by(job_type=JobType.SETUP_CERT_USER).delete()
+    JobLockFactory.create(job_type=JobType.SETUP_CERT_USER, updated_at=old_date)
+    db_session.commit()
+    with pytest.raises(TaskJobLockError) as excinfo:  # noqa
```
kkrug: I don't think you need the # noqa here
```suggestion
    with pytest.raises(TaskJobLockError) as excinfo:
```

api/src/task/task_job_lock.py
```diff
@@ -0,0 +1,201 @@
...
+            # Otherwise raise the specific exception we encountered for the job lock update failure.
+            raise TaskJobLockError from e
+
+    def get_duration_seconds(self) -> float | None:
+        if self.lock_acquired_at is not None:
+            return (datetime_util.utcnow() - self.lock_acquired_at).total_seconds()
+        return None
+
+    def get_task_job_lock_config(self) -> TaskJobLockConfig:
+        return TaskJobLockConfig()
```
kkrug: Is it necessary for this to be its own function? It could just be an inline `self.config = TaskJobLockConfig()` in `__init__`

jakobpederson: I was aiming for it be easy to swap things out, but I don't see a good reason why it has to be it's own method. I'll swapi it.

api/tests/src/task/test_task_job_lock.py
jakobpederson: It looks like if I remove it I get `# W: `pytest.raises()` block should contain a single simple statement`

api/src/task/task_job_lock.py
jakobpederson: I meant to grab it, I added it back. It's what I passed into the validation.

api/tests/src/task/test_task_job_lock.py
jakobpederson: All the `db_session.commit()` are there to keep the `db_session.begin()` working. As far as I can tell the factory is already configured to commit
```
class BaseFactory(factory.alchemy.SQLAlchemyModelFactory):

    class Meta:
        abstract = True
        sqlalchemy_session = Session
        sqlalchemy_session_persistence = "commit"
```

Which may be different than just the committing I'm doing.

api/src/task/task_job_lock.py
jakobpederson: I think we can just remove it then.

api/src/task/task_job_lock.py
```diff
@@ -0,0 +1,194 @@
...
+                        **self.extra,
+                        "locking_job_lock_id": job_lock.job_lock_id,
+                        "success": False,
+                        "error": "TaskJobLockIsLockedError",
+                    },
+                )
+                raise TaskJobLockIsLockedError
+
+            job_lock = self.set_job_lock_to_locked(job_lock)
+            self.verify_job_lock(job_lock)
```
chouinar: Verifying the job lock should be in a separate DB transaction from this above code.

Should be:
* Transaction 1
  * Grab lock
  * Error if not available
  * Update lock
  * Commit - technically if there were an issue with another process grabbing it, this probably fails at the DB level
* Transaction 2
  * Verify job lock

api/src/task/task_job_lock.py
```diff
@@ -0,0 +1,194 @@
...
+            minutes=self.lock_duration_minutes
+        )
+        job_lock.is_locked = True
+        job_lock.locked_by = self.internal_lock_id
+        self.add_job_lock(job_lock)
+        return job_lock
+
+    def set_job_lock_to_unlocked(self, job_lock: JobLock) -> JobLock:
+        job_lock.is_locked = False
+        self.add_job_lock(job_lock)
```
chouinar, line 190: Is the flush needed for anything when adding to the DB? Don't think we're getting anything from the DB we care about?

api/src/task/task_job_lock.py
```diff
@@ -0,0 +1,194 @@
...
+            logger.info(
+                "Job lock ids do not match",
+                extra={
+                    **self.extra,
+                    "locked_by": job_lock.locked_by,
+                    "success": False,
+                    "error": "TaskJobLockInternalIDError",
+                },
+            )
+            raise TaskJobLockInternalIDError
```
chouinar:
```suggestion
            logger.error(
                "Job lock ids do not match",
                extra={
                    **self.extra,
                    "locked_by": job_lock.locked_by,
                    "success": False,
                    "error": "TaskJobLockInternalIDError",
                },
            )
            raise TaskJobLockInternalIDError
```

api/src/task/task_job_lock.py
chouinar: What I'm imagining we do to split this, and other utils like this:
* Define a base SQLAlchemy model in the shared code. It uses that here for a type definition
* Define any of the endpoints that interact with fetching that to be abstract
* On each system that uses this, they make a subclass of task job lock and define the logic to fetch / plug in the concrete type

Doesn't need to be fully built that way right this second, but moving the DB interactions to a function will make it easy later

api/src/task/task_job_lock.py
```diff
@@ -0,0 +1,194 @@
...
+
+        try:
+            with self.db_session.begin():
+                job_lock = self.get_job_lock()
+
+                if not job_lock:
+                    raise TaskJobLockNotFoundError
+
+                job_lock = self.set_job_lock_to_unlocked(job_lock)
+                self.verify_job_lock(job_lock)
```
chouinar, line 116: As written, you update the lock and then call verify with does a refresh which would reload it from the DB.

I think what we should do is:
* get job lock
* Check that it's our lock
* Update

The verify function shouldn't do a refresh, just use what was passed in. That means in the enter function we'll want to re-fetch it first.

api/src/task/task_job_lock.py
jakobpederson: I'll remove the flush, I don't think we need it.

api/src/task/task_job_lock.py
```diff
@@ -0,0 +1,191 @@
...
+class TaskJobLockConfig(PydanticBaseEnvConfig):
+    enable_job_lock: bool = False
+
+
+class TaskJobLock(contextlib.AbstractContextManager[None]):
+
+    def __init__(
+        self,
+        db_session: db.Session,
+        job_type: str,
```
kkrug, line 42: Should this be `JobType` instead of `str`?

api/tests/src/task/test_task_job_lock.py
```diff
@@ -0,0 +1,184 @@
...
+            pass
+    record = next(r for r in caplog.records if r.message == "Job is currently locked")
+    assert record.success is False
+    assert record.locking_job_lock_id == job_lock.job_lock_id
+    assert record.error == "TaskJobLockIsLockedError"
+
+
+def test_verify_job_lock_raises_when_locked_by_does_not_match(
+    caplog, db_session, enable_factory_create
+):
```
kkrug, line 147: Should we have `clean_job_locks` here as well?

api/src/task/task_job_lock.py
jakobpederson: I'm not sure, the JobType is technically a string and we left it more generic in other places.

chouinar: Leave it string. Any enums that are `StrEnum` are also a string (StrEnum is just `MyEnum(str, Enum)`). When we separate this to the shared repo, we won't have a set list of enums since each system will track it differently.

api/tests/src/task/test_task_job_lock.py
```diff
@@ -0,0 +1,184 @@
...
+    TaskJobLockNotFoundError,
+)
+from tests.src.db.models.factories import JobLockFactory
+
+
+@pytest.fixture
+def clean_job_locks(db_session):
+    db_session.execute(delete(JobLock))
+    db_session.commit()
+    return
```
chouinar: How required is this for the tests? I want us to avoid any and all DB cleanup for tests that isn't absolutely essential. I want to start getting rid of it where possible.

We've got a lot of job types, could you use those? None of those should ever be run as part of other tests.

api/src/task/task_job_lock.py
```diff
@@ -0,0 +1,191 @@
...
+
+                logger.info(
+                    "Job lock held duration",
+                    extra={
+                        **self.extra,
+                        "lock_duration_seconds": self.get_duration_seconds(),
+                        "success": True,
+                        "lock_acquired_at": self.lock_acquired_at,
+                    },
+                )
```
chouinar, line 127: The duration is a nice thing to put, but in reality this is the "success" message - so let's word it like that - feel free to adjust. Also, move it outside the transaction just in case the commit at the end of the `with` block fails.

Also, it's basically identical, but a slightly different format for merging dicts. Can do `dict1 | dict2`

```suggestion
              logger.info(
                  "Job lock released",
                  extra=self.extra | {
                      "lock_duration_seconds": self.get_duration_seconds(),
                      "success": True,
                      "lock_acquired_at": self.lock_acquired_at,
                  },
              )
```

api/tests/src/task/test_task_job_lock.py
chouinar: What issue are you hitting?

The nested transaction issue is usually something else causing the problem.

api/tests/src/task/test_task_job_lock.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+def test_task_job_lock__does_not_create_a_job_lock_if_enable_job_lock_is_false(
+    clean_job_locks, monkeypatch_session, db_session, enable_factory_create
+):
+    monkeypatch_session.setenv("ENABLE_JOB_LOCK", "False")
```
chouinar: You're having to monkeypatch all these tests because you're patching the env var for the entire session of tests. A session being the full test run.

Switch it to this and you won't need to monkeypatch back everything as by default it'll use local.env's value.

```suggestion
    monkeypatch.setenv("ENABLE_JOB_LOCK", "False")
```

api/tests/src/task/test_task_job_lock.py
chouinar: That's saying that you shouldn't put multiple lines of code into a `pytest.raises` - put just the line that you expect to fail.

Since you want this to fail in the `__exit__` function, you could do that as:

```py
context_manager = TaskJobLock(db_session, job_type=...)
context_manager.__enter__()

# do whatever

with pytest.raises(Whatever):
    context_manager.__exit__()
```

All a with block is doing is running enter/exit and doing some error handling around the exit block to call it even if there is an exception.

api/tests/src/task/test_task_job_lock.py
```diff
@@ -0,0 +1,184 @@
...
+):
+    now = datetime_util.utcnow()
+    old_date = now - timedelta(days=366)
+    JobLockFactory.create(job_type=JobType.SETUP_CERT_USER, updated_at=old_date)
+    db_session.commit()
+    with pytest.raises(TaskJobLockError) as excinfo:  # noqa
+        with TaskJobLock(db_session, job_type=JobType.SETUP_CERT_USER, lock_duration_minutes=60):
+            db_session.execute(delete(JobLock).where(JobLock.job_type == JobType.SETUP_CERT_USER))
+            db_session.flush()
+            db_session.commit()
```
chouinar: commiting flushes, you shouldn't need both?

api/tests/src/task/test_task_job_lock.py
jakobpederson: I'll see if I can just rotate them since there are 18 or so types.

api/tests/src/task/test_task_job_lock.py
```diff
@@ -0,0 +1,188 @@
...
+    caplog.set_level(logging.INFO)
+    now = datetime_util.utcnow()
+    old_date = now - timedelta(days=366)
+    job_lock = JobLockFactory.create(
+        job_type=JobType.LOAD_OPPORTUNITY_DATA_OPENSEARCH,
+        updated_at=old_date,
+        is_locked=True,
+        locked_until=now + timedelta(days=2),
+    )
+    db_session.commit()
```
chouinar: factory create does a commit, you still have this on most of the tests.

```suggestion
    job_lock = JobLockFactory.create(
        job_type=JobType.LOAD_OPPORTUNITY_DATA_OPENSEARCH,
        updated_at=old_date,
        is_locked=True,
        locked_until=now + timedelta(days=2),
    )
```

hao10282025-sudo (commented):
Overall looks good to me.
I wonder if it is feasible the below scenario for the unit testing, like one by one create 2 or more same type job_locks?
Image

api/tests/src/task/test_task_job_lock.py
```diff
@@ -0,0 +1,184 @@
...
+        db_session.add(job_lock)
+        db_session.flush()
+    job_lock = db_session.execute(
+        select(JobLock).where(
+            JobLock.job_type == JobType.MIGRATE_DOWN,
+            JobLock.locked_by == context_manager.internal_lock_id,
+        )
+    ).scalar_one_or_none()
+    assert job_lock is not None
+    assert job_lock.job_type == JobType.MIGRATE_DOWN
```
kkrug: I think this test should actually enter the context manager instead of manually invoking the helpers. What if we did something like this:
```
def test_task_job_lock_creates_job_lock_if_one_does_not_exist(
    db_session, enable_factory_create
) -> None:
    pre_existing = db_session.execute(
        select(JobLock).where(JobLock.job_type == JobType.MIGRATE_DOWN)
    ).scalar_one_or_none()
    assert pre_existing is None

    context_manager = TaskJobLock(
        db_session, job_type=JobType.MIGRATE_DOWN, lock_duration_minutes=60
    )
    with context_manager:
        pass

    job_lock = db_session.execute(
        select(JobLock).where(JobLock.job_type == JobType.MIGRATE_DOWN)
    ).scalar_one_or_none()
    assert job_lock is not None
    assert job_lock.locked_by == context_manager.internal_lock_id
    assert job_lock.is_locked is False 
```

jakobpederson:
> Overall looks good to me. I wonder if it is feasible the below scenario for the unit testing, like one by one create 2 or more same type job_locks? Image


Should we add a unique constraint on the job locks? If there's more than one of the same type it throws a multiple rows found error.

api/tests/src/task/test_task_job_lock.py
jakobpederson: Had to add a commit() to get around some nested db_session stuff. I wonder if it's worth just deleting all MIGRATE_DOWNS, but I know we want to minimize that kind of thing.

hao10282025-sudo:
> > Overall looks good to me. I wonder if it is feasible the below scenario for the unit testing, like one by one create 2 or more same type job_locks? Image
> 
> Should we add a unique constraint on the job locks? If there's more than one of the same type it throws a multiple rows found error.

Hmm looks not necessary as your code logic has restrict only one record creation, the function is all okay.
I was originally thinking a test scenario like below:

create_job_lock(job_type = A)
create_job_lock(job_type = A)
create_job_lock(job_type = A)

assert *there is only one record for job_type = A*

I looked over again, this can be covered in your other test case at a lower level, so it is alright

chouinar:
> Should we add a unique constraint on the job locks? If there's more than one of the same type it throws a multiple rows found error.

Yeah, I didn't have that originally in the schema because it was going to work a different way, but this way assumes that's unique. Can we add that?

chouinar (approved):

Issue #9714, opened 2026-04-21
Create an endpoint that streams the results of a scan back to a user


chouinar:
### Summary

https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2948792998/Tech+Spec+File+Upload+for+Large+Files+Virus+Scanning#Scan-Endpoint

We want to create `GET /v1/files/:file_id/results`.

### Request

n/a - just the ID from the path

### Response

```json
{
  "data": {
    "status": "", # FileScanStatus enum
  }
}
```

### Auth

API User Key Auth or JWT Auth as usual

AuthZ will be a little different, rather than check a privilege, when we query the DynamoDB table, if the user ID associated with the row doesn't match the calling user's, we should 403. Effectively the user who started the file upload is the only one allowed to check its status.

### Logic

This endpoint will be a little different, we'll stream a response back rather than return a standard response. To do this we can follow this pattern:

```py
from flask import stream_with_context

@...
def my_route() -> Any:
     def generate() -> Any:
           for i in range(10):
                yield "something"

     return stream_with_context(generate())
```

This general pattern will let us stream back a response in chunks to a caller. Whatever we yield will be what is returned in each chunk.

For our logic we want to do the following in each iteration (any errors should end the loop immediately):
* Call DynamoDB for the given file ID (404 if not found)
* Validate that the user calling has the same user_id as the record (403 if not)
* Grab the status and return that
  * if the status is `success` or `infected` - that should break the loop
  * if the status is `pending` or `in_progress` - return it, but keep looping
* If more than 60 seconds have passed, end the stream regardless of result
* Each iteration should sleep for 3 seconds.

For the sake of unit testing, we might need to have this be somewhat configurable in terms of iteration duration, or time to sleep as we don't want 60 second tests.

### Acceptance criteria

- [ ] Streaming endpoint added that calls DynamoDB
- [ ] Values of how long we stream, and sleep are environment variables with defaults that we can override
- [ ] Tests do not run for 60 seconds

PR #10536, opened 2026-05-21
[Issue #9714] Create an endpoint that streams the results of a scan back to a user

kkrug:
## Summary

Fixes for #9714 

## Changes proposed

- Updated file_routes.py to add `GET /v1/files/<pending_file_id>/results`, returning a streaming NDJSON Response
- Updated file_schemas.py to add `FileScanResultsDataSchema` and `FileScanResultsResponseSchema`
- Created stream_file_scan_results.py service to poll DynamoDB on a configurable interval, yield each status, and end on terminal status (complete/infected) or max duration
- Updated local.env to add `FILE_SCAN_RESULTS_POLL_INTERVAL_SECONDS=3` and `FILE_SCAN_RESULTS_MAX_DURATION_SECONDS=60`
- Created test_get_file_scan_results.py to house tests for the new endpoint

## Context for reviewers

Implements the file-scan-results streaming endpoint on top of the DynamoDB adapter. AuthZ deviates from our usual privilege pattern, as only the user who started the upload can read its status, This is enforced by comparing the calling user against `user_id` on the DynamoDB record.

## Validation steps

Confirm tests pass

### DIFF
--- a/api/src/services/files/stream_file_scan_results.py
+++ b/api/src/services/files/stream_file_scan_results.py
@@ -1,0 +1,169 @@
+import json
+import logging
+import time
+import uuid
+from collections.abc import Iterator
+
+from pydantic import Field
+
+from src.adapters.aws.dynamodb_adapter import DynamoDBClient, DynamoDBConfig
+from src.api.route_utils import raise_flask_error
+from src.constants.lookup_constants import FileScanStatus
+from src.db.models.user_models import User
+from src.util import datetime_util
+from src.util.env_config import PydanticBaseEnvConfig
+
+logger = logging.getLogger(__name__)
+
+
+# When the status reaches one of these values, the scan is finished and we
+# stop streaming further updates.
+TERMINAL_STATUSES: frozenset[FileScanStatus] = frozenset(
+    {FileScanStatus.COMPLETE, FileScanStatus.INFECTED}
+)
+
+# Attribute names on the DynamoDB scan-cache record. The partition key matches
+# what setup_local_dynamodb.py / infra define.
+SCAN_RECORD_FILE_ID_KEY = "file_id"
+SCAN_RECORD_USER_ID_ATTR = "user_id"
+SCAN_RECORD_STATUS_ATTR = "status"
+
+
+class FileScanStreamConfig(PydanticBaseEnvConfig):
+    poll_interval_seconds: float = Field(
+        default=3.0, alias="FILE_SCAN_RESULTS_POLL_INTERVAL_SECONDS"
+    )
+    max_duration_seconds: float = Field(
+        default=60.0, alias="FILE_SCAN_RESULTS_MAX_DURATION_SECONDS"
+    )
+
+
+def _get_scan_record(
+    dynamodb_client: DynamoDBClient,
+    table_name: str,
+    pending_file_id: uuid.UUID,
+) -> dict | None:
+    response = dynamodb_client.get_item(
+        table_name=table_name,
+        key_name=SCAN_RECORD_FILE_ID_KEY,
+        value=str(pending_file_id),
+        consistent_read=True,
+    )
+    return response.item
+
+
+def _get_string_attr(record: dict, attr: str) -> str | None:
+    # DynamoDB returns items in attribute-value format: {"attr": {"S": "value"}}.
+    # We only need the string-typed attributes on the scan record.
+    raw = record.get(attr)
+    if not isinstance(raw, dict):
+        return None
+    return raw.get("S")
+
+
+def stream_file_scan_results(
+    pending_file_id: uuid.UUID,
+    user: User,
+    dynamodb_client: DynamoDBClient,
+    config: FileScanStreamConfig | None = None,
+    dynamodb_config: DynamoDBConfig | None = None,
+) -> Iterator[str]:
+    """Stream file scan status updates for the given pending file.
+
+    The initial DynamoDB lookup happens before any chunks are yielded so we can
+    return a 404 or 403 with the proper HTTP status. After streaming starts, any
+    subsequent error (missing record, user mismatch) just ends the stream.
+    """
+    if config is None:
+        config = FileScanStreamConfig()
+    if dynamodb_config is None:
+        dynamodb_config = DynamoDBConfig()
+
+    table_name = dynamodb_config.file_scan_cache_table_name
+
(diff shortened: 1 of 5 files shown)


api/src/services/files/stream_file_scan_results.py
```diff
@@ -0,0 +1,184 @@
...
+            except ValueError:
+                logger.warning(
+                    "Unable to parse file scan status from record, ending stream",
+                    extra={
+                        "pending_file_id": pending_file_id,
+                        "user_id": user.user_id,
+                        "raw_status": raw_status,
+                    },
+                )
+                return
```
chouinar: This function is pretty long, any way we could break it up a bit?

Might be a bit complex/too much to pass around, but something like this loop being just:
```
while True:
      result = _get_result(...)
      # all the various checks
      if stop_iterating(...):
           return

      return { ... }
```

api/src/services/files/stream_file_scan_results.py
```diff
@@ -0,0 +1,184 @@
...
+            if status in TERMINAL_STATUSES:
+                logger.info(
+                    "File scan results stream reached terminal status",
+                    extra={
+                        "pending_file_id": pending_file_id,
+                        "user_id": user.user_id,
+                        "file_scan_status": status,
+                    },
+                )
+                return
```
chouinar, line 197: Could we add the elapsed time to this log extra? Will be a useful metric to record to see how long people are usually waiting.

api/src/services/files/stream_file_scan_results.py
```diff
@@ -0,0 +1,184 @@
...
+    table_name: str,
+    pending_file_id: uuid.UUID,
+) -> dict[str, Any] | None:
+    response = dynamodb_client.get_item(
+        table_name=table_name,
+        key_name=SCAN_RECORD_FILE_ID_KEY,
+        value=str(pending_file_id),
+        consistent_read=True,
+    )
+    return response.item
```
chouinar: Just to save on parsing logic in the loop, could we parse the response here into a manageable object (dataclass?).

kkrug: Updated

api/src/services/files/stream_file_scan_results.py
kkrug: Updated

api/src/services/files/stream_file_scan_results.py
kkrug: Did some refactoring and broke this out

api/src/services/files/stream_file_scan_results.py
```diff
@@ -0,0 +1,237 @@
...
+
+    Fields are typed as Optional because a record in DynamoDB can be partially
+    written or carry an unknown status value; callers decide what to do with
+    a missing or unparseable field.
+    """
+
+    user_id: str | None
+    status: FileScanStatus | None
+    # Kept for diagnostic logging when ``status`` couldn't be parsed.
+    raw_status: str | None
```
chouinar: Is that partial DynamoDB case something we should expect or should we consider that some sort of error? If it means our DynamoDB setup is broken, I'd probably just respond to that by raising a 500 and killing the stream.

If the status is an invalid value, I'd say we just error immediately.

If we're streaming a result back and then raise an exception/give our usual error response format, does that work, or do we need to account for that in a different way? Not sure how status codes work in a stream

api/src/services/files/stream_file_scan_results.py
```diff
@@ -0,0 +1,237 @@
...
+        raise_flask_error(403, "Forbidden")
+
+    def generate() -> Iterator[dict[str, Any]]:
+        current_record = record
+        start_time = datetime_util.utcnow()
+
+        while True:
+            status = _resolve_status(current_record, pending_file_id, user.user_id)
+            if status is None:
+                return
```
chouinar: We have a lot of `is not None` checks inside the `_resolve_status` call that add a lot of branches to ultimately end up at this if-statement.

If we're able to do `raise_flask_error` and that response object works with the streamed approach, I'd say we just do that in that logic. If not, we could do:


```suggestion
            try:
                status = _resolve_status(current_record, pending_file_id, user.user_id)
            except StatusParseError:
                logger.exception("something")
                return
```

Change all that branching logic to short-circuit and immediately jump to this exception. Then we would also be able to keep so many fields from being nullable that can't be in reality

api/src/services/files/stream_file_scan_results.py
```diff
@@ -0,0 +1,237 @@
...
+                )
+                return
+
+            time.sleep(config.poll_interval_seconds)
+
+            next_record = _fetch_next_record(
+                dynamodb_client, table_name, pending_file_id, user.user_id
+            )
+            if next_record is None:
+                return
```
chouinar: This being None would mean we got a record at least once to enter the loop and then the record vanished? I think this should definitely be an error case that we at least log.

kkrug: Updated

api/src/services/files/stream_file_scan_results.py
kkrug: Made some updates

api/src/services/files/stream_file_scan_results.py
kkrug: Added `InvalidFileScanRecordError` to get raised for some missing values, plus some tests to verify

api/src/services/files/stream_file_scan_results.py
```diff
@@ -0,0 +1,236 @@
...
+            time.sleep(config.poll_interval_seconds)
+
+            try:
+                next_record = _fetch_next_record(
+                    dynamodb_client, table_name, pending_file_id, user.user_id
+                )
+            except InvalidFileScanRecordError:
+                # Streaming has already started so we can't change the HTTP
+                # status; log explicitly (Flask's error processor only fires
+                # for HTTPError) and re-raise to abort the connection.
```
chouinar, line 220: Flask will log any exception, although there's special logic for HTTPError that we've registered. For everything else, it'll fallback to this:
https://flask.palletsprojects.com/en/stable/errorhandling/#unhandled-exceptions

That said, still good to log an error here as we can add context to the log message like the file ID and user ID.

chouinar (approved):

Issue #8687, opened 2026-02-26
Generate and Send Journey Session ID


babebe:
### Summary

Implement frontend logic to generate a UUIDv4 SID on first page load, store it in a secure cookie , and send it with every API request via header. This enables tracking of anonymous user activity across multiple tabs and authentication transitions.

### Acceptance criteria

- [ ] SID is generated as a valid UUIDv4 on first page load if none exists.
- [ ] SID is stored in a Secure + HttpOnly cookie.
- [ ] SID is included in a custom header (`X-Journey-Session-ID`) on all API requests.
- [ ] SID persists across multiple tabs and browser sessions.

freddieyebra:
@joshtonava @babebe can i get confirmation that we want a 7 day sliding window TTL on the correlation id cookie and not 1 day.

joshtonava:
Im not 100% certain, the spec goes back in forth in the comments. We can start 1 and then always change it later.

freddieyebra:
Going to proceed with 1 day TTL and httpOnly true. This will provide a bit more hardening to the cid cookie to aid in organic correlation events. If we need to eventually support emitting client side logging events, we will need a follow up ticket for creating a separate nextjs api endpoint that client side components can request to emit logging events if needed. otherwise, server side logging should have access to this cookie.

PR #10537, opened 2026-05-21
[Issue #8687] Add correlation id support

freddieyebra:
## Summary

Fixes / Work for #8687 

## Changes proposed

This pr adds creating an anonymous correlation id within the frontend middleware.

## Context for reviewers

This pr updates the current middleware in the frontend to add the new correlation id cookie.

## Validation steps

1. Run the frontend application
2. Right click and open inspect > application > cookies > ensure `correlation_id` cookie is set
3. Look into logs and verify that there was a logging event similar to
```
{"level":"info","time":1779397898328,"event":"anonymous_session_started","correlation_id":"cc96d951-2974-43e5-aea7-fbd24a2c49c6","reason":"missing"}
```
4. Set the cid cookie to a random non uuid string and reload the page
5. Ensure you see a log similar to the following 
```
{"level":"info","time":1779398448805,"event":"anonymous_session_started","correlation_id":"e275549d-abe4-4ccb-8897-f77677ecfbbd","reason":"invalid"}
```
6. Optionally set `const CORRELATION_ID_MAX_AGE_SECONDS = 1;` and refresh the browser and assert a new uuid cid is assigned.
7. Ensure new windows persist the correlation_id cookie


myduong-navapbc:
Overall, I think everything looks good with some minor suggestions:

myduong-navapbc:
Can we update the task/comments to reflect the finalized behavior? 

The original AC mentioned a 7-day TTL and httpOnly=false, but implementation is now using a 1-day sliding TTL and httpOnly=true.

myduong-navapbc:
Minor suggestion: 
instead of matching against the full URL, could we use something like:
const isApiRoute = request.nextUrl.pathname.startsWith("/api/");

frontend/src/services/correlationId/correlationIdMiddleware.ts
```diff
@@ -0,0 +1,75 @@
...
+ */
+export const applyCorrelationId = (
+  request: NextRequest,
+  response: NextResponse,
+): NextResponse => {
+  const existing = request.cookies.get(CORRELATION_ID_COOKIE)?.value;
+
+  let correlationId: string;
+  if (isValidCorrelationId(existing)) {
+    correlationId = existing as string;
```
myduong-navapbc: [Small nit]:
we can avoid the as string cast by checking existing !== undefined before validation. 

That keeps this fully type-safe without the assertion.

frontend/src/services/correlationId/correlationIdMiddleware.ts
```diff
@@ -0,0 +1,75 @@
...
+export const isValidCorrelationId = (value: string | undefined): boolean =>
+  typeof value === "string" && UUID_V4_REGEX.test(value);
+
+type CreationReason = "missing" | "invalid";
+
+const logAnonymousSessionStarted = (
+  correlationId: string,
+  reason: CreationReason,
+): void => {
+  logger.info({
```
myduong-navapbc, line 34: Me like ^

frontend/src/services/correlationId/correlationIdMiddleware.ts
myduong-navapbc: `if (existing !== undefined && isValidCorrelationId(existing)) {
  correlationId = existing;
}`

myduong-navapbc (dismissed):

myduong-navapbc (approved):

PR #10545, opened 2026-05-21
[Unticketed] Fix LoadOracleData filter logic timezone issue

kkrug:
## Summary

Fixes for Unticketed

## Changes proposed

- Updated load_oracle_data_task.py to strip the timezone from the `batch_cutoff`
- Updated test_load_oracle_data_task.py to account for change in tests 

## Context for reviewers

The Oracle source data is EST/EDT wall-clock time stored as UTC in Postgres. Our cutoff was a timezone-aware EDT value. When Postgres compared them, it normalized both to UTC, which made the source data appear to be 4 hours earlier than it really was, letting post-cutoff records sneak through.

## Validation steps

Confirm tests pass. Confirm no New Relic errors occur in lower environments before the next prod release.


chouinar (approved):
LGTM!

Issue #9971, opened 2026-04-30
Backend - Create the endpoint to delete an award recommendation


nathan-stricker:
### Summary

See [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Award-Recommendation.1)

#### Endpoint
`DELETE /award-recommendations/:award_recommendation_id`

#### Purpose
Delete an award recommendation - does a soft-delete for auditing purposes

#### AuthZ
Requires `update_award_recommendation` privilege

#### Request
n/a

#### Response
```json
{
  "data": null,
  "message": "Success",
  "status_code": 200
}
```

#### Logic

- Fetch the award recommendation, 404 if not found
- AuthZ, 403 on error
- Set the `is_deleted` flag to True on the award recommendation - we don’t actually delete it.

### Acceptance criteria

- [ ] Endpoint created
- [ ] Tests created/updated as necessary

PR #10550, opened 2026-05-22
[Issue#9971]- Delete Award Recommendation

chay-REIsys:
## Summary

Work for #9971 

## Changes proposed

Soft Delete Award Recommendation

## Validation steps

Call the endpoint passing award recommendation id and see is_deleted is updated to true in db
make sure local tests pass


api/src/services/award_recommendations/delete_award_recommendation.py
```diff
@@ -0,0 +1,20 @@
...
+    db_session: db.Session,
+    user: User,
+    award_recommendation_id: UUID,
+) -> None:
+    award_recommendation = get_award_recommendation_for_update(
+        db_session, user, award_recommendation_id
+    )
+
+    award_recommendation.is_deleted = True
+    db_session.add(award_recommendation)
```
nathan-stricker, line 33: I think we should probably be adding an `AwardRecommendationAudit` event here, like we do in [update_award_recommendation.py](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/services/award_recommendations/update_award_recommendation.py)

nathan-stricker (approved):
Approving changes, just make sure those E2Es finish before merging, may need to restart the task

PR #10553, opened 2026-05-22
[Issue #10549] Support GetSubmissionList

### LINKED ISSUE #10549: Support Grantor GetSubmissionList
### Summary

Need to support GetSubmissionList not GetSubmissionListExpanded for GS integration.
They're very similar, maybe just have to remove the DUNS and UEI fields from the Expanded response to make it compatible with the non-Expanded response.
https://grants.gov/system-to-system/grantor-system-to-system/web-services/get-submission-list

### Acceptance criteria

- GetSubmissionList requests return data from Simpler and Grants.gov

jakobpederson:
## Summary

Fixes / Work for #10549  

## Changes proposed

- Add `GetSubmissionList` support
- Added tests
- Updated schemas, services

## Context for reviewers

We need to support `GetSubmissionList` for Grant Solutions.

## Validation steps

- [ ] hit the `GetSubmissionList` endpoint and see correct response

### DIFF
--- a/api/src/legacy_soap_api/grantors/services/get_submission_list_response.py
+++ b/api/src/legacy_soap_api/grantors/services/get_submission_list_response.py
@@ -102,7 +102,7 @@ def transform_submission(submission: ApplicationSubmission) -> dict[str, str | d
 
 def get_submissions(
     db_session: db.Session,
-    request: schemas.GetSubmissionListExpandedRequest | schemas.GetSubmissionListRequest,
+    request: schemas.GetSubmissionListRequest,
     soap_request: SOAPRequest,
     soap_config: SOAPOperationConfig,
 ) -> list[ApplicationSubmission]:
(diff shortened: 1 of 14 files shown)


mdragon (dismissed):
👍🏻

api/tests/src/legacy_soap_api/grantors/services/test_get_submission_list_response.py
```diff
@@ -0,0 +1,278 @@
...
+            soap_request=soap_request,
+            soap_config=_make_operation_config(),
+        )
+        result = get_submission_list_response(
+            simpler_submissions=simpler_submissions,
+            proxy_response=SOAPResponse(data="", status_code=200, headers={}),
+        )
+        assert result.success is True
+        assert result.available_application_number == 1
+        assert "GrantGovApplicationStatus" not in dict(result.submission_info[0])
```
kkrug: I think there are some issues with this test. The test name says `grants_gov_tracking_number` but the only assertions are for `GrantGovApplicationStatus`. Also, should it be `GrantsGovApplicationStatus`. with an 's?' And then for `dict(result.submission_info[0])`, is this iterating through the field names or the aliases? If it's not the aliases, this test is essentially 'assert True`.

api/src/legacy_soap_api/grantors/services/get_submission_list_response.py
```diff
@@ -102,7 +102,7 @@ def transform_submission(submission: ApplicationSubmission) -> dict[str, str | d
 
 def get_submissions(
     db_session: db.Session,
-    request: schemas.GetSubmissionListExpandedRequest,
+    request: schemas.GetSubmissionListExpandedRequest | schemas.GetSubmissionListRequest,
```
chouinar: If the expanded one is a subclass, and identical, just have the one type.

```suggestion
    request: schemas.GetSubmissionListRequest,
```

api/src/legacy_soap_api/grantors/services/get_submission_list_response.py
```diff
@@ -230,6 +230,26 @@ def get_submission_list_expanded(
...
+    database submissions into SubmissionInfoExpanded objects.
+
+    Returns a list of SubmissionInfoExpanded from Simpler.
+    """
+    submissions = get_submissions(db_session, request, soap_request, soap_config)
+    return [
+        schemas.SubmissionInfoExpanded(**transform_submission(submission))
+        for submission in submissions
+    ]
+
```
chouinar: Any way we can minimize how much logic we need to have that is different between the two basically identical endpoints?

That way if we have to make any adjustments to logic, we just make it to one spot rather than two.

Any way we could do something like:

```
def get_submission_list(..., is_expanded: bool):
    # fetch stuff

    if is_expanded:
         schema = schemas.SubmissionInfoExpanded
    else:
         schema = schemas.SubmissionInfo

    return [schema(**transform_submission(submission)) for submission in submissions]
```

api/src/legacy_soap_api/grantors/services/get_submission_list_response.py
jakobpederson: I was thinking about that, I was worried it would get confusing if there were just one. I'll change it.

kkrug (approved):

PR #10555, opened 2026-05-22
[Issue #10502] Need headless not headed Chrome

mdragon:
## Summary

Work for #10502  

## Changes proposed

We need to cache headless chrome, not regular chrome

## Context for reviewers

This was changed in the last PR around test infra, but because it was using a cached Playwright install that had all the browsers, it didn't fail when we swapped out the all browser install to the just chrome install because we hit the cache that included headless chrome.

## Validation steps

- Cleared existing playwright cache so that we would do a clean headless only install and test that completely.
- Checks pass

### DIFF
--- a/.github/actions/e2e/action.yml
+++ b/.github/actions/e2e/action.yml
@@ -109,7 +109,7 @@ runs:
       shell: bash
       working-directory: ./frontend
       run: |
-        npx playwright install --with-deps chrome
+        npx playwright install --with-deps chromium-headless-shell
 
     # Conditionally install Firefox if requested (Shard 2 on the CI/CD run)
     - name: Install Firefox and dependencies
--- a/.github/workflows/ci-frontend-e2e.yml
+++ b/.github/workflows/ci-frontend-e2e.yml
@@ -147,7 +147,7 @@ jobs:
           npm ci
 
       # Install Chrome only for caching; other browsers installed per-shard in test job
-      - run: npx playwright install --with-deps chrome
+      - run: npx playwright install --with-deps chromium-headless-shell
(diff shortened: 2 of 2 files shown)


doug-s-nava (approved):

.github/actions/e2e/action.yml
```diff
@@ -109,7 +109,7 @@ runs:
       shell: bash
       working-directory: ./frontend
       run: |
-        npx playwright install --with-deps chrome
+        npx playwright install --with-deps chromium-headless-shell
```
Bhavna-Ramachandran, line 112: I think we should use `chromium`. As far as i know `chromium-headless-shell` cannot emulate mobile devices.

.github/workflows/ci-frontend-e2e.yml
```diff
@@ -147,7 +147,7 @@ jobs:
           npm ci
 
       # Install Chrome only for caching; other browsers installed per-shard in test job
-      - run: npx playwright install --with-deps chrome
+      - run: npx playwright install --with-deps chromium-headless-shell
```
Bhavna-Ramachandran, line 150: Same comment as above. Update command to use `chromium`

.github/actions/e2e/action.yml
mdragon: So I'm not picking, our existing config failed when I installed chromium only and passes when i install chromium-headless-shell only. I got the headless-shell hint from the error messages that said that couldn't find it's files to execute. So I think we've been using this all along and now we're just no longer doing the "install all browsers".  For example, with the cache cleared and just this install happening, Shard 4, Mobile Chrome, executed without issue?

Issue #10424, opened 2026-05-14
Fix existing skipped XSD validation tests


joshtonava:
### Summary

Fix the XSD validation failures that #10425 explicitly skipped, then remove the skip decorators so the suite is fully running.

We assumed ~ 3 tests to be fixed. If time required to fix tests exceeds ticket points, up ticket points to correct representation.

Current results:
```
============================================================
XML VALIDATION TEST SUMMARY
============================================================
Total Tests: 39
Successful: 30
Failed: 9
Success Rate: 76.9%

FAILURE BREAKDOWN:
  xsd_validation: 9 tests
    - sf424_with_single_attachment
    - sf424_with_multiple_attachments
    - sf424_with_all_attachment_types
    - sf424a_minimal_non_federal_resources_only
    - sf424a_budget_sections_with_array_decomposition
    - sf424a_with_forecasted_cash_needs
    - sf424a_complete_all_sections
    - sf424a_complete_all_sections
    - sf424a_complete_all_sections
```

### Acceptance criteria

- [ ] All `@pytest.mark.skip` decorators introduced by #10425 (referencing this ticket) are removed once the underlying issue is fixed
- [ ] tests all pass

Issue #10543, opened 2026-05-21
Improve XML fetcher to auto discover XSD dependencies


joshtonava:
### Summary

The XSD fetcher only knows about hardcoded dependencies for one form, so transitive schema imports for every other form are silently missed. Update the fetcher to discover dependencies directly from each XSD instead of relying on a static map. 

Reference #10425 [comments ](https://github.com/HHS/simpler-grants-gov/issues/10425#issuecomment-4500890510)for inspiration.

### Acceptance criteria

- [x] Fetcher discovers dependencies by parsing XSDs; static dependency map removed.
- [x] `make fetch-xsds` pulls every transitive dependency for all supported forms.
- [x] Validation tests still resolve cross schema references against the flat cache.
- [x] Parser tested against imports, nested transitive imports, and circular imports.

PR #10560, opened 2026-05-22
[Issue #10425] xsd schema commit to repo and failed test skip

### LINKED ISSUE #10425: Commit XSD schemas to repo and run XML validation tests in CI
### Summary

Commit grants.gov's XSD schemas into the repo and run XSD validation as part of the normal unit test suite in CI. Any tests that fail get explicitly skipped (tracked in #10424) so the pipeline lands green and we can iterate on fixes.

**Concerns / discussion points:**

- `KNOWN_XSD_DEPENDENCIES` in `xsd_fetcher.py` is incomplete and only declares deps for SF424. Other forms likely have transitive imports the fetcher doesn't currently follow. The plan here is to iterate: commit what we know, find missing imports when tests fail, fix as needed. A more permanent fix (auto parsing imports) is worth a follow up but is out of scope for this ticket.
- After this ticket, running `make fetch-xsds` modifies tracked files in the working copy 

### Acceptance criteria

- [x] Download XSDs from Grants.gov form repository [for supported forms ](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2766897293/Apply+Supported+Forms+and+Opportunities)(and known dependencies) and committed to `api/src/services/xml_generation/xsds/`
- [x] `xsd_cache/` removed from `.gitignore`
- [x] XSD fetcher writes to the new committed path
- [x] `xml_validation` marker removed from the codebase
- [x] Runtime skip guards removed from XSD validation test files; validator paths updated
- [x] Any currently failing tests explicitly skipped with reasons referencing #10424
- [x] CI runs the XSD validation tests as part of the normal test suite

babebe:
## Summary

Fixes / Work for #10425  

## Changes proposed

Downloaded XSDs from the Grants.gov form repository for supported forms and their known dependencies.
Committed all XSD files to api/src/services/xml_generation/xsds/.
Removed xsd_cache/ from .gitignore.
Updated XSD fetcher (xsd_fetcher.py) to write directly to the committed XSD directory instead of cache.
Removed xml_validation marker and related conditional logic.
Removed runtime skip guards from XSD validation test files.
Updated validator paths to point to committed schema location.
Explicitly skipped currently failing tests with references to #10424 for follow-up remediation.
Ensured CI now runs XSD validation tests as part of the standard unit test suite.

## Validation steps

make fetch-xsds - Confirm files are written to: `api/src/services/xml_generation/xsds/`
Run full test suite - Make Test:
Confirm:
XSD validation tests execute in CI (no xml_validation marker skips)
Previously failed tests remain skipped with references to #10424
No xsd_cache/ usage or .gitignore dependency remains
CI pipeline:
XSD validation run as part of standard test stage
Pipeline remains green with expected skips documented


.gitignore
```diff
@@ -156,5 +156,3 @@ dmypy.json
 # Artilery
 frontend/tests/artillery/params.json
 
-# XSD cache for XML validation
-xsd_cache/
```
babebe, line 160: @joshtonava @mikehgrantsgov Do we want to ignore the validation_results.json file  here ?

api/src/services/xml_generation/xsds/validation_results.json
mikehgrantsgov, line 1: We don't want to commit this file. It should be git ignored.

api/tests/src/services/xml_generation/test_submission_xsd_validation.py
mikehgrantsgov: Wouldn't this path need to be updated?

mikehgrantsgov (dismissed):

api/tests/src/services/xml_generation/test_submission_xsd_validation.py
```diff
@@ -28,26 +28,23 @@
...
+        """Create XSD validator with directory."""
         from pathlib import Path
 
-        xsd_cache_dir = Path(__file__).parent.parent.parent.parent.parent / "xsd_cache"
-        if not xsd_cache_dir.exists():
-            pytest.skip(
-                "XSD cache directory not found. Run 'flask task fetch-xsds' to download schemas."
-            )
-        return XSDValidator(xsd_cache_dir)
+        xsd_dir = Path(__file__).parent.parent.parent.parent.parent / "services/xml_generation/xsds"
```
mikehgrantsgov: The XSDs live at api/src/services/xml_generation/xsds. is this right?

babebe: Correct. Updated these

api/src/cli/xml_generation_cli.py
```diff
@@ -149,15 +149,15 @@ def generate_xml_command(
     help="Form name to test (e.g., SF424_4_0). If not specified, runs all forms.",
 )
 @click.option(
-    "--cache-dir",
+    "--xsd_dir",
```
mikehgrantsgov: should be --xsd-dir to match Click convention and be consistent with the fetch-xsds command

api/src/cli/xml_generation_cli.py
```diff
@@ -202,23 +202,23 @@ def validate_xml_generation_command(
     if verbose:
         logging.getLogger().setLevel(logging.DEBUG)
     try:
-        # Use default cache directory if not specified
-        if not cache_dir:
-            cache_dir = "./xsd_cache"
+        # Use default XSD directory if not specified
+        if not xsd_dir:
+            xsd_dir = str(Path(__file__).resolve().parents[4] / "src/services/xml_generation/xsds")
```
mikehgrantsgov: What does parents[4] resolve to?

api/src/cli/xml_generation_cli.py
```diff
@@ -232,13 +232,23 @@ def validate_xml_generation_command(
                 click.echo("Error: No test cases found", err=True)
                 sys.exit(1)
 
+        # APPLY SKIP FILTER
+        test_cases, skipped = filter_skipped(test_cases)
```
mikehgrantsgov, line 236: Either remove filter_skipped and let the CLI runner just run whatever pytest runs, or move the skip list to test_cases.py where it can serve both. filter_skipped creates a second skip list that's parallel to the @pytest.mark.skip decorators added to the pytest tests. These two lists will diverge

.gitignore
```diff
@@ -156,5 +156,5 @@ dmypy.json
...
+# XML validation results
+validation_results.json
```
mikehgrantsgov: Scope this to api/src/services/xml_generation/xsds/validation_results.json, we don't want to suppress all validation_results.json files.

api/src/cli/xml_generation_cli.py
babebe: Wrong change here. Fixed

api/src/cli/xml_generation_cli.py
babebe: The skip logic is currently form-level in the CLI runner vs test-level in pytest, so they don’t map cleanly 1:1.
Trying to consolidate skip mechanisms, does not really make sense to me.

freddieyebra: I agree with @mikehgrantsgov. We can still get rid of the `xml_validation` filter but we will still need to specify pytest.mark.skip. We should utilize that for skipping any failing validation test.

api/src/services/xml_generation/validation/test_runner.py
```diff
@@ -16,34 +15,38 @@
...
         """
         self.xml_service = XMLGenerationService()
-        self.xsd_validator = XSDValidator(xsd_cache_dir)
+        self.xsd_validator = XSDValidator(xsd_dir)
         self.results: list[dict[str, Any]] = []
+        self.xml_form_map = xml_form_map
+
+    def _get_transform_config(self, form_name: str) -> dict | None:
+        """Return transform config for a form, matching case-insensitively."""
+        return self.xml_form_map.get(form_name.upper())
```
freddieyebra, line 49: can we move this to be a shared function? this will likely be needed in other xml related functionality.

api/src/services/xml_generation/submission_xml_assembler.py
```diff
@@ -36,6 +37,7 @@ def __init__(
         self.application_submission = application_submission
         self.xml_service = XMLGenerationService()
         self.attachment_mapping = attachment_mapping
+        init_form_registry()
```
freddieyebra, line 40: will this not have already been called when the app starts up? cc: @mikehgrantsgov

api/tests/src/services/xml_generation/test_cd511_xml_generation.py
```diff
@@ -240,30 +239,27 @@ def test_generate_cd511_xml_element_order_matches_xsd(self):
...
-        """Create XSD validator with cache directory."""
-        xsd_cache_dir = Path(__file__).parent.parent.parent.parent.parent / "xsd_cache"
-        if not xsd_cache_dir.exists():
-            pytest.skip(
-                "XSD cache directory not found. Run 'flask task fetch-xsds' to download schemas."
-            )
+        """Create XSD validator."""
+        xsd_dir = Path(__file__).parents[4] / "src/services/xml_generation/xsds"
+        if not xsd_dir.exists():
+            pytest.skip("XSD directory not found. Run 'flask task fetch-xsds' to download schemas.")
```
freddieyebra, line 250: Can we create a helper function to get the xsd file or file path based on a form_name or other relevant attribute? for these tests, we will have already downloaded the XSD for all of the supported forms and should not need to check for existence in every forms test suite. If the XSD is not found, we should fail instead of silently skipping.

freddieyebra (changes requested):

api/src/services/xml_generation/submission_xml_assembler.py
babebe: Pytest never starts the app, so the registry would be empty without it in tests that call the assembler directly.

api/tests/src/services/xml_generation/test_cd511_xml_generation.py
babebe: There is a ticket created for this. https://github.com/HHS/simpler-grants-gov/issues/10543

api/src/cli/xml_generation_cli.py
babebe: I added filter_skipped here as the skip mechanism for this CLI validation layer since these are plain dicts, not pytest tests. An alternative would be adding a skip field directly to the test case dicts in test_cases.py — happy to go that route if preferred.

api/src/services/xml_generation/validation/test_runner.py
babebe: Good call — config.py already has load_xml_transform_config but the test runner needs slightly different behavior — it returns None for unsupported forms rather than raising, so it can skip them gracefully.

freddieyebra (commented):
I think the `SupplementaryCoverSheetforNEHGrantPrograms_3_0-V3.0.xsd` schema may be missing as well as `ProjectNarrativeAttachments_1_2-V1.2.xsd`

freddieyebra (approved):
this looks good!

Verified `make fetch-xsds` works as expected:
- Results from running without any changes on branch

Image

- Results from removing an XSD from xsd_schemas directory

Image

Verified xml_validation tests are not skipped (excluding existing failing tests) and ran as part of the CI/CD

Image

Issue #10282, opened 2026-05-07
Fixing award Min Max displays in search table


eric-serrato95:
### Summary

This ticket will address fixing the [issues](https://github.com/HHS/simpler-grants-gov/issues/9513) with award quantities sometimes getting cropped into two lines as opposed to showing the full amount in a single line.

### Acceptance criteria

- [x] See @chelsea-nava 's comment below for beginning style changes on the page.
- [x] Make the dollar amounts no wrap in css.
- [x] Ensure number can handle up to `$99,000,000,000.00` without wrapping
- [x] Support multiple breakpoints including desktop, tablet, mobile
- [x] Set up 30 min with Chelsea to review page at different breakpoints and ensure that it looks appropriate in all cases.
- [ ] UAT passed in staging with Product

chelsea-nava:
- [x] Award min/max containers: Increase container width to support 12 characters plus punctuation
- [x] Award min/max containers: No wrap

To help accommodate this change, [please take a look here: ](https://www.figma.com/design/bQUTJCMj0JYbAAnJQ4p4tX/Grantee-Designs-%7C-Simpler-Grants-2026?node-id=7796-4423&t=CWPqHxQ8G64HgWmq-1)

- [x] Change **status pill text size** to match the size used for "expected awards"
- [x] Change **date text size** to match the size used for "expected awards"

- [x] Design review/colab with @chelsea-nava to get this right (we want to make the minimal change possible to achieve no wrapping)

freddieyebra:
Synced with @chelsea-nava on this. In addition to to the min/max no-wrapping, it was suggested to increase the width of agency and title

freddieyebra:
Met with @chelsea-nava and without making other content smaller/truncating, we can set max width on the award min. However, that would only impact lower environments since in production, there have been historically no award min with more than 8 digits (11 with , and $). We will not be resizing columns as part of this change unless there is another need to.

freddieyebra:
@eric-serrato95 this is ready for UAT. It is deployed to on [grantee1](https://grantee1.teams.simpler.grants.gov/search)

eric-serrato95:
@freddieyebra the amounts now look good, but I think there might a new issue popping up which is, for quantities in the Billions , the agency and date are now getting squished in very tightly 
@ErinPattisonNava @chelsea-nava 
question here:
how do we feel about the current solution? see attached screenshot for reference 

[...]

PR #10561, opened 2026-05-22
[Issue #10282] Fix min/max award rows in search results

freddieyebra:
## Summary

Fixes / Work for #10282 

## Changes proposed

This changes some font sizes as well as preventing award min/max monetary values from wrapping in the table.

## Context for reviewers

To improve the UI, we want to ensure there is no wrap in the award values on the search results table.

## Validation steps

1. Find an opportunity id in the search results you want to use to test.
2. Update the `opportunity_summary.award_ceiling` in the DB with an award ceiling value with 11 digits
3. Run `make populate-search-opportunities` to update the search index.
4. Run the frontend
5. Find that opportunity in the search table, assert that the value does not wrap like the following:
Screenshot 2026-05-22 at 2 51 55 PM


ErinPattisonNava (dismissed):
Looks great, tested both award min and award max at multiple breakpoints and the numbers stayed on one line.

ErinPattisonNava (approved):

PR #10578, opened 2026-05-26
[unticketed] remove token headers from award rec calls

doug-s-nava:
## Summary

unticketed
Brings award rec API fetch related code up to date in terms of fetcher / endpoint definition patterns and token usage

## Validation steps

1. run general regression test of award rec functionality


chay-REIsys (commented):
As i am reviewing changes in removing token from award recommendation calls but i am seeing unauthorzed error though i am using pre defined useClientFetch function
with server side calls from next js i didn't see this issue
that is the reason why i added token from session in those functions

@doug-s-nava

doug-s-nava:
closing this in favor of https://github.com/HHS/simpler-grants-gov/pull/10657

Issue #10567, opened 2026-05-22
GrantSolutions calls to SOAP Notary failing


mdragon:
### Summary

-

### Acceptance criteria

-

PR #10589, opened 2026-05-26
[Issue #10567] Pass increased header limits as cmd args

mdragon:
## Summary

Work for #10567  

## Changes proposed

Pass increased header size limits via cmd line args since config entries didn't seem to be getting picked up as expected.

## Validation steps

- Checks pass
- ✅ Super large headers tested after deploying this to Dev


prasnava (approved):
👍

Issue #9836, opened 2026-04-24
Upgrade from v15.x to v16.x from template infra


sean-navapbc:
### Summary

Upgrade from v15.x to v16.x from template infra

### Acceptance criteria

Deploy into dev and into prod succesfully

sean-navapbc:
This work has been merged and completed.

sean-navapbc:
code is merged

PR #10625, opened 2026-05-27
[Issue #10486] Create client for calling SES

### LINKED ISSUE #10486: Create a client for calling SES
### Summary

Related to #9836. Infra is handling infra portion setup and add SES. Backend engineers need to build and test backend and remove pinpoint

**Create this client in our grants_shared package - do not create it under api**

We want to add an adapter for calling SES, similar to the one we have for calling Pinpoint. We just need the client to wrap the [send_email](https://docs.aws.amazon.com/boto3/latest/reference/services/ses/client/send_email.html) function of the SES client.

The client should have a configuration with `AWS_SES_FROM_EMAIL` - the email we'll send from. This'll be configured by infra, set something in local.env.

Our function should take in:
* Who to send the email to
* Subject
* Message

We should send an email, handling any errors gracefully.

The response from SES looks to just be a message ID, log and return that.

Unlike pinpoint, SES has its functionality mocked out by moto3 (the mock boto3 we use for unit tests) so we can set that up instead of creating the hacky mock for pinpoint: https://docs.getmoto.org/en/latest/docs/services/ses.html

This ticket won't yet swap existing functionality to use this client, we'll have follow-up tickets to switchover.

### Acceptance criteria

- [ ] Create a client for interacting with SES in our `grants_shared` code
- [ ] Setup any test utils in conftest for SES - use them to write unit tests for sending the emails
- [ ] When running locally, we shouldn't actually send anything, we're not going to have a service to call for real - should just do a log

hao10282025-sudo:
## Summary

Fixes #10486 

## Changes proposed

We want to add an adapter for calling SES, similar to the one we have for calling Pinpoint. We just need the client to wrap the [send_email](https://docs.aws.amazon.com/boto3/latest/reference/services/ses/client/send_email.html) function of the SES client.

## Context for reviewers

Related to https://github.com/HHS/simpler-grants-gov/issues/9836. Infra is handling infra portion setup and add SES. Backend engineers need to build and test backend and remove pinpoint

Create this client in our grants_shared package - do not create it under api

The client should have a configuration with AWS_SES_FROM_EMAIL - the email we'll send from. This'll be configured by infra, set something in local.env.

Our function should take in:

Who to send the email to
Subject
Message
We should send an email, handling any errors gracefully.

The response from SES looks to just be a message ID, log and return that.

Unlike pinpoint, SES has its functionality mocked out by moto3 (the mock boto3 we use for unit tests) so we can set that up instead of creating the hacky mock for pinpoint: https://docs.getmoto.org/en/latest/docs/services/ses.html

This ticket won't yet swap existing functionality to use this client, we'll have follow-up tickets to switchover.

## Validation steps

- Create a client for interacting with SES in our grants_shared code
- Setup any test utils in conftest for SES - use them to write unit tests for sending the emails
- When running locally, we shouldn't actually send anything, we're not going to have a service to call for real - should just do a log

### DIFF
--- a/backend/grants_shared/src/grants_shared/aws/ses_adapter.py
+++ b/backend/grants_shared/src/grants_shared/aws/ses_adapter.py
@@ -23,7 +23,7 @@ def get_ses_client(session: boto3.Session | None = None) -> botocore.client.Base
     if session is None:
         session = get_boto_session()
 
-    return session.client("ses")
+    return session.client("sesv2")
 
 
 def send_email(
@@ -33,16 +33,16 @@ def send_email(
     ses_client: botocore.client.BaseClient | None = None,
 ) -> str:
     """
-    Send an email using AWS SES.
+    Send an email using AWS SESv2.
 
     Args:
         to_address: Email address to send to
         subject: Email subject line
         message: Email body (supports both HTML and plain text)
-        ses_client: Optional SES client (for testing)
+        ses_client: Optional SESv2 client (for testing)
 
     Returns:
-        Message ID from SES
+        Message ID from SESv2
 
     Raises:
         Exception: If email fails to send
@@ -65,7 +65,7 @@ def send_email(
 
     try:
         logger.info(
-            "Sending email via SES",
+            "Sending email via SESv2",
             extra={
                 "to_address": to_address,
                 "subject": subject,
@@ -74,14 +74,16 @@ def send_email(
         )
 
         response = ses_client.send_email(
-            Source=config.aws_ses_from_email,
+            FromEmailAddress=config.aws_ses_from_email,
             Destination={"ToAddresses": [to_address]},
-            Message={
-                "Subject": {"Data": subject, "Charset": "UTF-8"},
-                "Body": {
-                    "Text": {"Data": message, "Charset": "UTF-8"},
-                    "Html": {"Data": message, "Charset": "UTF-8"},
-                },
+            Content={
+                "Simple": {
+                    "Subject": {"Data": subject, "Charset": "UTF-8"},
+                    "Body": {
+                        "Text": {"Data": message, "Charset": "UTF-8"},
+                        "Html": {"Data": message, "Charset": "UTF-8"},
+                    },
+                }
             },
         )
 
@@ -89,7 +91,7 @@ def send_email(
         message_id = response_object.message_id
 
         logger.info(
-            "Successfully sent email via SES",
+            "Successfully sent email via SESv2",
             extra={
                 "message_id": message_id,
                 "to_address": to_address,
@@ -98,20 +100,18 @@ def send_email(
 
         return message_id
 
-    except ClientError as e:
+    except ClientError:
         logger.exception(
-            "Failed to send email via SES",
+            "Failed to send email via SESv2",
             extra={
                 "to_address": to_address,
                 "subject": subject,
-                "error_code": e.response.get("Error", {}).get("Code"),
-                "error_message": e.response.get("Error", {}).get("Message"),
             },
         )
-        raise Exception("Failed to send email via SES") from e
+        raise
     except Exception:
         logger.exception(
-            "Unexpected error sending email via SES",
+            "Unexpected error sending email via SESv2",
             extra={
                 "to_address": to_address,
                 "subject": subject,
--- a/backend/grants_shared/src/grants_shared/aws/ses_adapter.py
+++ b/backend/grants_shared/src/grants_shared/aws/ses_adapter.py
@@ -1,0 +1,120 @@
+import logging
+
+import boto3
+import botocore.client
+from botocore.exceptions import ClientError
+from pydantic import BaseModel, Field
+
+from grants_shared.aws.aws_session import get_boto_session, is_local_aws
+from grants_shared.util.env_config import PydanticBaseEnvConfig
+
+logger = logging.getLogger(__name__)
+
+
+class SESConfig(PydanticBaseEnvConfig):
+    aws_ses_from_email: str = Field(alias="AWS_SES_FROM_EMAIL")
+
+
+class SESResponse(BaseModel):
+    message_id: str = Field(alias="MessageId")
+
+
+def get_ses_client(session: boto3.Session | None = None) -> botocore.client.BaseClient:
+    if session is None:
+        session = get_boto_session()
+
+    return session.client("ses")
+
+
+def send_email(
+    to_address: str,
+    subject: str,
+    message: str,
+    ses_client: botocore.client.BaseClient | None = None,
+) -> str:
+    """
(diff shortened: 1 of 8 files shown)


backend/grants_shared/src/grants_shared/aws/ses_adapter.py
```diff
@@ -0,0 +1,120 @@
...
+
+class SESResponse(BaseModel):
+    message_id: str = Field(alias="MessageId")
+
+
+def get_ses_client(session: boto3.Session | None = None) -> botocore.client.BaseClient:
+    if session is None:
+        session = get_boto_session()
+
+    return session.client("ses")
```
chouinar: This is on me, but just noticed that there's actually a newer version of SES (`sesv2`), which the infra did set us up to use, can we adjust this to that?

https://docs.aws.amazon.com/boto3/latest/reference/services/sesv2/client/send_email.html

I think it's just a slight change in the request schema from looking at it.

backend/grants_shared/tests/grants_shared/aws/test_ses_adapter.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def test_send_email_success(ses_client, monkeypatch):
+    """Test successfully sending an email via SES"""
+    from_email = os.getenv("AWS_SES_FROM_EMAIL")
+    to_email = "recipient@example.com"
+    message_subject = "Test Subject"
+    message_body = "This is a test email message"
+
+    ses_client.verify_email_identity(EmailAddress=from_email)
```
chouinar: Move this repeated setup to a fixture.


```suggestion

# in conftest
@pytest.fixture
def ses_from_email(ses_client, monkeypatch):
     from_email = "example@example.com"
     monkeypatch.setenv("AWS_SES_FROM_EMAIL", from_email)
     ses_client.verify_email_identity(EmailAddress=from_email)
     return from_email

def test_send_email_success(ses_client, monkeypatch, ses_from_email):
    """Test successfully sending an email via SES"""
    to_email = "recipient@example.com"
    message_subject = "Test Subject"
    message_body = "This is a test email message"
```

backend/grants_shared/tests/grants_shared/aws/test_ses_adapter.py
```diff
@@ -0,0 +1,145 @@
...
+
+import pytest
+from moto.core import DEFAULT_ACCOUNT_ID
+from moto.ses.models import ses_backends
+
+import grants_shared.aws.aws_session as aws_session_module
+from grants_shared.aws import send_email
+
+
+def test_send_email_success(ses_client, monkeypatch):
```
kkrug: There are a few tests in here that have `monkeypatch` as an arg but never use it

backend/grants_shared/src/grants_shared/aws/ses_adapter.py
```diff
@@ -0,0 +1,120 @@
...
+        raise Exception("Failed to send email via SES") from e
+    except Exception:
+        logger.exception(
+            "Unexpected error sending email via SES",
+            extra={
+                "to_address": to_address,
+                "subject": subject,
+            },
+        )
+        raise
```
kkrug: You can drop the `raise Exception(...) from e` wrapper and re-raise the original `ClientError`. Wrapping in a bare `Exception` forces callers to catch the broadest type and loses the ClientError semantics. You may also need to update `test_send_email_unverified_sender`. 
```suggestion
    except ClientError:
        logger.exception(
            "Failed to send email via SES",
            extra={
                "to_address": to_address,
                "subject": subject,
            },
        )
        raise
    except Exception:
        logger.exception(
            "Unexpected error sending email via SES",
            extra={
                "to_address": to_address,
                "subject": subject,
            },
        )
        raise
```

backend/grants_shared/src/grants_shared/aws/ses_adapter.py
hao10282025-sudo: Updated

backend/grants_shared/tests/grants_shared/aws/test_ses_adapter.py
hao10282025-sudo: Updated.
I actually combine the "ses_from_email" to the "ses_client" to be more succinct, the 3rd test case is added with delete_email_identity()

backend/grants_shared/tests/grants_shared/aws/test_ses_adapter.py
hao10282025-sudo: removed unused monkeypatch

backend/grants_shared/src/grants_shared/aws/ses_adapter.py
hao10282025-sudo: yeah I checked again, updated with "raise" only
the test case is updated to compare with partial of the general error message

backend/grants_shared/src/grants_shared/aws/ses_adapter.py
```diff
@@ -0,0 +1,120 @@
...
+    except Exception:
+        logger.exception(
+            "Unexpected error sending email via SESv2",
+            extra={
+                "to_address": to_address,
+                "subject": subject,
+            },
+        )
+        raise
```
kkrug: Do we need this generic exception at all? Anything that isn't a `ClientError` would propagate with it's original traceback, right?

backend/grants_shared/src/grants_shared/aws/ses_adapter.py
```diff
@@ -0,0 +1,120 @@
...
+    if ses_client is None:
+        ses_client = get_ses_client()
+
+    try:
+        logger.info(
+            "Sending email via SESv2",
+            extra={
+                "to_address": to_address,
+                "subject": subject,
+                "from_address": config.aws_ses_from_email,
```
chouinar: I would leave out logging any of this. The to_address and potentially subject would be identifying info about a user, and we want to avoid that.

The from_address would be statically defined for an env, so not as important to log either.

backend/grants_shared/src/grants_shared/aws/ses_adapter.py
hao10282025-sudo: Yes, I looked over online and that is the Python build-in logging feature!
Removed the last exception block.

backend/grants_shared/src/grants_shared/aws/ses_adapter.py
hao10282025-sudo: yeah I agree!
Removed email info fields for non-local part, the local logger is still keeping these fields for easy debugging

chouinar (dismissed):

kkrug (dismissed):

chouinar (approved):

Issue #9454, opened 2026-04-08
Arrange existing components based on results of audit


doug-s-nava:
### Summary

Following #9453, move all components that can be easily relocated as directed in the audit document.

Any files that require any work before relocation should be adjusted and moved as part of TBD 

Note that 
* unit test files will also be moved to remain colocated with the components they are testing
* all necessary file references will be updated
* any directories that are empty after migrating can be deleted

### Acceptance criteria

- [ ] all component files that do not require any adjustment are moved to their new locations as dictated by the audit document

doug-s-nava:
see https://docs.google.com/spreadsheets/d/11DGUIvcgatvWu6L1yi6LiHvnADmF-XnjkmGR07nXrjY/edit?gid=0#gid=0 for audit results

Issue #9453, opened 2026-04-08
Audit existing components for type and usage


doug-s-nava:
### Summary

In order to properly organize our components based on the plan laid out in the [tech spec here](https://navasage.atlassian.net/wiki/x/L4DsrQ), we first need to categorize our existing components to know where to put them and whether any changes need to be made to them to make them more able to fit into our new categorization scheme.

The work of this ticket is to perform such an audit and capture the results in a document. For each component file currently in the system, we should capture:

* it's assignment as a core, domain specific, or local component file
* whether the file needs any work in order to be properly categorized 
  * for example, if a file contains both local and domain specific files, we should note which components should be split into new files
  * or if theres is something that is very close to being a core component but needs some change in order to make it more universal, we should note that
* new location for the file

Note that

* layout and page components will not be part of this audit
* no change should be made to the code base as part of this audi

### Acceptance criteria

- [ ] document is produced covering the topics mentioned above
- [ ] document is reviewed and approved by at least one other frontend developer

PR #10632, opened 2026-05-27
[Issue #9454] core organization pieces that need a closer look

doug-s-nava:
## Summary

Work for #9454 

## Changes proposed

As part 1 of moving all components identified as core components into a new structure, this change includes moving files that needed a bit more love in order to be properly defined within the new core structure.

Based on the audit from https://github.com/HHS/simpler-grants-gov/issues/9453

## Context for reviewers

A few things to call out:

* SummaryDescriptionDisplay renamed ExpandableTextContent and moved out of opportunity file
* TestApplicationLink and SignOutNavLink were split out of existing "UserControl" file

## Validation steps

1. General regression test


frontend/src/components/core/ExpandableTextContent.tsx
```diff
@@ -0,0 +1,62 @@
+import DOMPurify from "isomorphic-dompurify";
+import { splitMarkup } from "src/utils/generalUtils";
+
+import ContentDisplayToggle from "src/components/ContentDisplayToggle";
+
+export const ExpandableTextContent = ({
+  textContent = "",
+  showCallToAction,
```
doug-s-nava, line 8: some new props to support core flexibility

frontend/src/app/[locale]/(base)/award-recommendation/[id]/edit/page.tsx
```diff
@@ -21,14 +21,14 @@ import {
...
-import ApplyFormNav from "src/components/applyForm/ApplyFormNav";
 import AwardRecommendationAttachments from "src/components/award-recommendation/AwardRecommendationAttachments";
 import AwardRecommendationHero, {
   HeroButtonConfig,
 } from "src/components/award-recommendation/AwardRecommendationHero";
 import { RecommendationSection } from "src/components/award-recommendation/RecommendationSection";
 import { RecommendationSummarySection } from "src/components/award-recommendation/RecommendationSummarySection";
-import { SummaryDescriptionDisplay } from "src/components/opportunity/OpportunityDescription";
+import { ExpandableTextContent } from "src/components/core/ExpandableTextContent";
+import ApplyFormNav from "src/components/core/forms/LeftHandFormNav";
```
Comment (glenblosser-nava) [inline frontend/src/app/[locale]/(base)/award-recommendation/[id]/edit/page.tsx]: Why do you have ApplyFormNav moved down the list in some of these files? I believe our standard is for imports to be alphabetical

Comment (glenblosser-nava) [reply frontend/src/app/[locale]/(base)/award-recommendation/[id]/edit/page.tsx]: Alternatively, it should be renamed to LeftHandFormNav rather than leaving the name as ApplyFormNav

Comment (doug-s-nava) [reply frontend/src/app/[locale]/(base)/award-recommendation/[id]/edit/page.tsx]: yeah, good catch on the renaming. I noticed but left it as I didn't want to confuse the diff on the next PR, but I'll go ahead and fix. For import order, that's defined in our prettier config https://github.com/HHS/simpler-grants-gov/blob/5a5e165058af77d9d42ad95c0469c9207f99679f/frontend/.prettierrc.js#L14. I always let prettier handle that on save because it's too complicated to reason about otherwise.

In this case I think it's been moved because this is the way it works out alphabetically:
* "src/components/award-recommendation"
* "src/components/core/ExpandableTextContent"
* "src/components/core/forms/LeftHandFormNav"

Comment (myduong-navapbc) [reply frontend/src/app/[locale]/(base)/award-recommendation/[id]/edit/page.tsx]: This makes me wonder if we should have a lint check for import ordering as part of CI. If import order is already defined in Prettier, it might be nice to fail builds when files don't match the expected order

myduong-navapbc:
Looks good.
The renames seem to align with the audit's goal of making the shared components more generic. 

Thanks for cleaning that up while doing the migration. ❤️

myduong-navapbc (approved):

frontend/src/app/[locale]/(base)/award-recommendation/[id]/edit/page.tsx
Comment (doug-s-nava) [reply frontend/src/app/[locale]/(base)/award-recommendation/[id]/edit/page.tsx]: we run a prettier check in ci that will fail if the imports are out of order, so that should already be covered

PR #10633, opened 2026-05-27
[Issue #9454] core component reorganization - simple file moves

doug-s-nava:
## Summary

Work for #9454 

## Changes proposed

Moves all components designated as core components during the audit from https://github.com/HHS/simpler-grants-gov/issues/9453 into proper locations. This change should not include any changes beyond moving files.

## Context for reviewers

All changes in this PR should be either:

* a file moved to the core directory without changes
* an update to import paths to reflect the new file location

The Headers test file is for some reason showing up as a completely new file rather than being moved, but I think that is just a quirk of git

Note that not all components organized into the core directory in this change are fully ready to be thought of as "core", but anything that is in there now can be thought of as either already "core" or in the pipeline to become "core" by the end of the quad.

## Validation steps

1. General regression test


frontend/src/components/application/ApplicationHistoryTable.tsx
```diff
@@ -7,7 +7,7 @@ import { Alert } from "@trussworks/react-uswds";
 import {
   TableCellData,
   TableWithResponsiveHeader,
-} from "src/components/TableWithResponsiveHeader";
+} from "src/components/core/TableWithResponsiveHeader";
```
ErinPattisonNava, line 10: Storybook name should probably be updated: https://github.com/HHS/simpler-grants-gov/blob/32b38534befcb881703ed1a7ff239539fb32c445/frontend/stories/components/TableWithResponsiveHeaders.stories.tsx#L9

frontend/src/components/applyForm/widgets/MultiSelectWidget.tsx
```diff
@@ -16,9 +16,9 @@ import {
   type ComboBoxRef,
 } from "@trussworks/react-uswds";
 
-import { FieldErrors } from "src/components/applyForm/FieldErrors";
-import { Pill } from "src/components/Pill";
-import { DynamicFieldLabel } from "./DynamicFieldLabel";
+import { DynamicFieldLabel } from "src/components/core/forms/DynamicFieldLabel";
+import { FieldErrors } from "src/components/core/forms/FieldErrors";
+import { Pill } from "src/components/core/Pill";
```
ErinPattisonNava, line 21: Storybook name should probably be updated: https://github.com/HHS/simpler-grants-gov/blob/32b38534befcb881703ed1a7ff239539fb32c445/frontend/stories/components/Pill.stories.tsx#L4

frontend/src/components/award-recommendation/AwardRecommendationAttachments.tsx
```diff
@@ -8,13 +8,13 @@ import { useTranslations } from "next-intl";
 import { useEffect, useState } from "react";
 import { Pagination } from "@trussworks/react-uswds";
 
-import { PopoverMenu } from "src/components/PopoverMenu";
-import SimplerAlert from "src/components/SimplerAlert";
-import Spinner from "src/components/Spinner";
+import { PopoverMenu } from "src/components/core/PopoverMenu";
+import SimplerAlert from "src/components/core/SimplerAlert";
```
ErinPattisonNava, line 12: Storybook name should probably be updated: https://github.com/HHS/simpler-grants-gov/blob/32b38534befcb881703ed1a7ff239539fb32c445/frontend/stories/components/SimplerAlert.stories.tsx#L4

frontend/src/components/core/forms/CommonFormFields.tsx
```diff
@@ -10,8 +10,8 @@ import {
   TextInput,
 } from "@trussworks/react-uswds";
 
-import { FieldErrors } from "src/components/applyForm/FieldErrors";
-import { DynamicFieldLabel } from "src/components/applyForm/widgets/DynamicFieldLabel";
+import { DynamicFieldLabel } from "src/components/core/forms/DynamicFieldLabel";
```
ErinPattisonNava, line 13: Should this really be a core component, or is it part of only the apply widgets and really tied to the designs for those Forms (vs general forms).

frontend/src/components/core/header/Header.tsx
```diff
@@ -24,10 +24,10 @@ import {
   Header as USWDSHeader,
 } from "@trussworks/react-uswds";
 
-import NavDropdown from "src/components/NavDropdown";
-import { RouteChangeWatcher } from "src/components/RouteChangeWatcher";
-import { TestUserSelect } from "src/components/TestUserSelect";
-import { USWDSIcon } from "src/components/USWDSIcon";
+import NavDropdown from "src/components/core/header/NavDropdown";
+import { RouteChangeWatcher } from "src/components/core/header/RouteChangeWatcher";
```
ErinPattisonNava, line 28: Documentation needs updating: https://github.com/HHS/simpler-grants-gov/blob/32b38534befcb881703ed1a7ff239539fb32c445/documentation/frontend/authentication.md#L39

frontend/src/components/core/header/Header.tsx
```diff
@@ -24,10 +24,10 @@ import {
   Header as USWDSHeader,
 } from "@trussworks/react-uswds";
 
-import NavDropdown from "src/components/NavDropdown";
-import { RouteChangeWatcher } from "src/components/RouteChangeWatcher";
-import { TestUserSelect } from "src/components/TestUserSelect";
-import { USWDSIcon } from "src/components/USWDSIcon";
+import NavDropdown from "src/components/core/header/NavDropdown";
```
ErinPattisonNava, line 27: Nice grouping.

frontend/src/components/core/ClientSideUrlUpdater.test.tsx
```diff
@@ -1,6 +1,6 @@
 import { render } from "@testing-library/react";
 
-import { ClientSideUrlUpdater } from "src/components/ClientSideUrlUpdater";
+import { ClientSideUrlUpdater } from "src/components/core/ClientSideUrlUpdater";
```
ErinPattisonNava, line 3: This feels search specific, rather than a core component.  In depends on `useSearchParamUpdater`

frontend/src/components/core/ContentDisplayToggle.test.tsx
```diff
@@ -1,7 +1,7 @@
 import { act, render, screen, waitFor } from "@testing-library/react";
 import { Breakpoints } from "src/types/uiTypes";
 
-import ContentDisplayToggle from "src/components/ContentDisplayToggle";
+import ContentDisplayToggle from "src/components/core/ContentDisplayToggle";
```
ErinPattisonNava, line 4: Storybook name should be updated: https://github.com/HHS/simpler-grants-gov/blob/32b38534befcb881703ed1a7ff239539fb32c445/frontend/stories/components/ContentDisplayToggle.stories.tsx#L6

frontend/src/components/core/Snackbar.test.tsx
```diff
@@ -5,7 +5,7 @@ import { axe } from "jest-axe";
 
 import React from "react";
 
-import Snackbar from "src/components/Snackbar";
+import Snackbar from "src/components/core/Snackbar";
```
ErinPattisonNava, line 8: Storybook name should be updated: https://github.com/HHS/simpler-grants-gov/blob/32b38534befcb881703ed1a7ff239539fb32c445/frontend/stories/components/Snackbar.stories.tsx#L4

frontend/src/components/search/Filters/FilterOptionLabel.tsx
```diff
@@ -1,7 +1,7 @@
 import clsx from "clsx";
 import { FilterOption } from "src/types/search/searchFilterTypes";
 
-import InfoTooltip from "src/components/InfoTooltip";
+import InfoTooltip from "src/components/core/tooltip/InfoTooltip";
```
ErinPattisonNava, line 4: Storybook name should be updated: https://github.com/HHS/simpler-grants-gov/blob/32b38534befcb881703ed1a7ff239539fb32c445/frontend/stories/components/InfoTooltip.stories.tsx#L6

ErinPattisonNava (commented):
Tested basics of search, apply, and manage users.  Functionality looks right.  The requested changes shouldn't require retesting.  

This PR was really too big, and could have been split easily into several smaller PRs per component.  Please be kind to reviewers going forward and split things this large.

frontend/src/components/core/forms/CommonFormFields.tsx
doug-s-nava: At the moment the `CommonFormFields` are only being used in one form (the create opportunity form - not apply widgets), but this is one case that I'm referring to in the PR description comment `Note that not all components organized into the core directory in this change are fully ready to be thought of as "core", but anything that is in there now can be thought of as either already "core" or in the pipeline to become "core" by the end of the quad.`

Agreed it could go either for in the interim here, but I'd rather think of these as core starting now so that folks don't add domain specific logic to them before I have time to take a pass at them to turn them into true core components.

frontend/src/components/core/ClientSideUrlUpdater.test.tsx
doug-s-nava: I'm torn on this one. Since it's only used in one place right now, I'm ok with making this local. 

However, it _should_ be applicable to anywhere on the site where we want to do the very specific thing it's doing. 

However, also, do we need it at all? Why wouldn't a redirect work instead of updating the url manually like this?

Regardless:

* I don't want to necessarily let the scope on this ticket creep into research around whether we need to keep this component
* The verbiage around `useSearchParamUpdater` is mirroring NextJS's perferred term for query params (see https://nextjs.org/docs/app/api-reference/functions/use-search-params). This has caused issues in the past, since `search params` really does make things sound search specific, and I've tried to use the term `query params` instead, but have not cleaned up the verbiage in this component. Again, I think that's out of scope

I'll add a ticket for the find team to look into whether we need this component, and along with that make a decision about where it should live (whether they would like to take ownership or leave that to platform). Does that work?

doug-s-nava:
@ErinPattisonNava Can we leave the storybook naming to the storybook reorganization work later this quad? I called that out as out of scope in my tech spec https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2917957679/Tech+Spec+Frontend+Component+Sharing+and+Organization

frontend/src/components/core/ClientSideUrlUpdater.test.tsx
ErinPattisonNava: Thanks, that makes sense.

ErinPattisonNava (dismissed):
Thanks for making the documentation change!

ErinPattisonNava (approved):
Re-did minimal regression after merge conflicts, still looks good

PR #10635, opened 2026-05-27
[Issue #9708] Setup local logic that mocks out what the virus scanner actually does

### LINKED ISSUE #9708: Setup local logic that mocks out what the virus scanner actually does
### Summary

https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2948792998/Tech+Spec+File+Upload+for+Large+Files+Virus+Scanning#Scanning-Replacement

We don't need scanning locally, but we need something that will handle taking a file that was uploaded to s3 and moving it as if it was scanned.

To set this up we'll use a background thread that runs in our API. In our `create_app` function we want to setup a background thread like so:

```py
def create_app():
    # ... existing logic
    setup_local_background_thread()


def scan_files():
      # TODO - this is where the meaningful logic will go we describe below
      pass

def setup_local_background_thread():
        if os.environ.get("ENVIRONMENT", None) != "local":
            return
 
        # TODO - add something to account for unit tests, this shouldn't run there either
        # likely another env var that we set in local.env, and in our conftest set to disable it.

        if os.environ.get("WERKZEUG_RUN_MAIN") == "true":
            logger.info("Setting up background process")
            thread = threading.Thread(target=background_process, name="some-thread-name", daemon=True)
            thread.start()
```

This will get us a background thread where we can run code. What do we want to do?

Watch the directory where we store s3 files for changes (Something like [watchfiles](https://pypi.org/project/watchfiles/) ) - files are stored in `locals3root` - note that they have a different naming scheme than s3, but can watch for any changes to any files named `objectMetadata.json` in the file scanning bucket.

For each file that is changed do the following logic:
* If the file's key starts with anything other than `unscanned` - do nothing
* If the file's key has `scenario-` in it, we're handling a specific test scenario:
  * `scenario-infected` - Move the file to the `/infected` folder in the file scanning bucket.
  * `scenario-wait10s` - Wait 10 seconds to process the file - first changing the status to `in_progress`
  * We may add more scenarios over time
* Regardless of the scenario, we'll want to do the following:
  * Call the `POST /v1/files/:file_id` endpoint with the file scan status (default to `complete`)
  * Update the status in DynamoDB.

### Acceptance criteria

- [ ] Created automation that runs only locally for acting as the scanning lambda
- [ ] Make sure that we only process files when they're done being uploaded. Might want to test what happens when we upload a file that takes a few seconds - do we see the change when the upload starts or at the end? Can we do anything to filter out in-progress uploads? If no way to detect, add a delay in processing.

kkrug:
## Summary

Fixes #9708  

## Changes proposed

- Created local_file_scanner.py to mock the s3 virus-scanning lambda locally 
- Created test_local_file_scanner.py to test the mock virus scanning functionality
- Updated app.py to call `setup_local_file_scanner()` in `create_app()`
- Updated s3_adapter.py to add `required file_scan_bucket_path` on `S3Config`
- Updated local.env to add `FILE_SCAN_BUCKET`, `LOCAL_S3_STORE_PATH`, `ENABLE_LOCAL_FILE_SCANNER=TRUE`, `LOCAL_FILE_SCANNER_PRE_PROCESS_DELAY_SECONDS`, `LOCAL_FILE_SCANNER_WAIT_SCENARIO_DELAY_SECONDS`, and `LOCAL_FILE_SCANNER_USER_ID`
- Updated docker-compose.yml to add `local-mock-file-scan-bucket` to s3mock's initial buckets
- Updated pyproject.toml to add `watchfiles>=1.0.0,<2` to dev group
- Updated uv.lock to add `watchfiles>=1.0.0,<2`
- Updated conftest.py tp set `ENABLE_LOCAL_FILE_SCANNER=FALSE` for the test session and add `s3_scanner_user` fixture
- Updated dynamodb_adapter.py to add `put_item` and pass `aws_access_key_id="local"` when `AWS_DYNAMODB_ENDPOINT_URL` is set so PutItem against `dynamodb-local` works
- Updated test_dynamodb_adapter.py to update two existing assertions for the new local-creds kwargs
- Updated seed_orgs_and_users.py to seed the local file scanner user with the `INTERNAL_S3_SCANNER_ROLE`

## Context for reviewers

This implements the local stand-in for the virus-scanning lambda. It watches s3mock's on-disk store and writes scan status to postgres + DynamoDB and is gated to `ENVIRONMENT=local` + Flask reloader's worker process so it never runs in tests, CI, or prod. The change to dynamodb_adapter.py also fixes a latent bug (`NO_CREDS` is rejected by `dynamodb-local` on `PutItem`) that didn't surface until we wrote to DynamoDB at runtime.

## Validation steps

Confirm tests pass. The three scenarios (happy path, infected, and wait10s) can also be tested manually via `make console` commands, but that is a lot to add here so I can walk anyone through it if necessary.


api/local.env
```diff
@@ -143,6 +143,18 @@ FILE_SCAN_RESULTS_MAX_DURATION_SECONDS=60
...
+############################
+# Local file scanner
+############################
+
+# The local file scanner watches this directory (s3mock's store, bind-mounted
+# into the api container) and mocks the virus-scanning lambda. Set
+# DISABLE_LOCAL_FILE_SCANNER=TRUE to skip the background thread (the test
+# conftest does this so unit tests don't spawn it).
+LOCAL_S3_STORE_PATH=/api/locals3root
+DISABLE_LOCAL_FILE_SCANNER=FALSE
```
chouinar: Just in the interest of avoiding double negatives, can we flip this?

```suggestion
ENABLE_LOCAL_FILE_SCANNER=TRUE
```

api/src/services/files/local_file_scanner.py
```diff
@@ -0,0 +1,270 @@
...
+import logging
+import threading
+import time
+import uuid
+from pathlib import Path
+from typing import Any
+
+from pydantic import Field
+from sqlalchemy import select
+from watchfiles import Change, watch
```
chouinar: As an extra safety check, we could make `watchfiles` a dev dependency so it only exists in local dev, and not in the release image. We'd have to move this import down below to the `_run_scanner` function but since I think that's the only place it's used, that'd be fine.

api/src/services/files/local_file_scanner.py
```diff
@@ -0,0 +1,270 @@
...
+# non-happy paths without needing a real virus scanner.
+SCENARIO_INFECTED = "scenario-infected"
+SCENARIO_WAIT_10S = "scenario-wait10s"
+
+WAIT_10S_DELAY_SECONDS = 10
+
+# Pause briefly after a metadata change so we don't read s3mock's file mid-write
+# during a slow / multi-part upload. We have no signal from the file system that
+# tells us the upload is "done", so we rely on this short fixed delay.
+PRE_PROCESS_DELAY_SECONDS = 1.0
```
chouinar: Some of the times for this might be better as an environment variable, would simplify some of the test setup too where we want to lower these to reasonable numbers.

Fine if we re-grab a pydantic setting class in each thread - we don't change env vars during processing, so don't need to worry about thread safety, and it's local only

api/src/services/files/local_file_scanner.py
```diff
@@ -0,0 +1,270 @@
...
+        pending_file = db_session.execute(stmt).scalar_one_or_none()
+        if pending_file is None:
+            logger.warning(
+                "Local file scanner could not find pending file row",
+                extra=log_extra,
+            )
+        else:
+            pending_file.file_scan_status = status
+            user_id = pending_file.user_id
+    return user_id
```
chouinar: Since there might be some other changes we want to make for the pending file logic, could we reuse some of what we implemented for the endpoint? Instead of fetching from the DB ourselves, just reuse `update_pending_file_scan_status` that we built.

Only issue is we'd need to have a user with that privilege created, could have an env var with their user ID, and add logic to create them in our seed script / unit tests. Did that for the user we need in the workflow service.

api/src/services/files/local_file_scanner.py
```diff
@@ -0,0 +1,270 @@
...
+            "status": {"S": status.value},
+        },
+    )
+
+
+def _move_to_infected(s3_key: str) -> None:
+    s3_config = S3Config()
+    bucket_path = s3_config.file_scan_bucket_path
+    infected_key = INFECTED_PREFIX + s3_key[len(UNSCANNED_PREFIX) :]
+    file_util.move_file(f"{bucket_path}/{s3_key}", f"{bucket_path}/{infected_key}")
```
chouinar: Might want to make this slightly more generic / when a file is successfully scanned it should also be moved to another folder - based on some security related convos I had with infra.

I think they'll move them to `s3://whatever-bucket/scanned` in that case?

api/local.env
kkrug: Updated

api/src/services/files/local_file_scanner.py
kkrug: Updated

api/src/services/files/local_file_scanner.py
kkrug: Updated to use env vars

api/src/services/files/local_file_scanner.py
kkrug: Updated to use existing logic and created user

api/src/services/files/local_file_scanner.py
kkrug: Updated

chouinar (approved):

Issue #10650, opened 2026-05-28
Define approach to frontend unit test coverage


doug-s-nava:
### Summary

While we're collecting unit test coverage for jest tests, we aren't doing anything with them.

Questions to answer:

* what sort of thresholds do we want to set as standards for coverage?
* how will we enforce these standards?
* how do we want to report coverage and when?

Depending on the answers we can use this ticket to handle next steps (documentation, implementation) or we can create new tickets for follow up work

### Acceptance criteria

- [ ] questions above are answered
- [ ] tickets are written for follow up tasks

PR #10651, opened 2026-05-28
[unticketed] print frontend unit test coverage to std out

doug-s-nava:
## Summary

unticketed

## Changes proposed

* prints jest test coverage to std out in CI
* removes test coverage report step in CI that was not doing anything helpful

see https://github.com/HHS/simpler-grants-gov/issues/10650 for follow up

## Context for reviewers

Currently we are using an action in CI that is designed to calculate diffs on coverage for a PR by running tests against main and comparing. This action by default reports results in a comment on the PR. These comments likely would get noisy, as a new comment is added with each test run. In our implementation this commenting is turned off, in which case the action sends a markdown version of the report as an output to the next CI step. However, this report is not being used anywhere. Thus, this change comments out the report action for now while we figure out next steps.

With this change, jest will still create a json test coverage report in addition to printing to CI so that we can easily reimplement a step to make coverage more visible, though for now it will not be used anywhere.

## Validation steps

1. _VERIFY_: CI std out shows full coverage report like this
Screenshot 2026-05-28 at 10 18 14 AM

### DIFF
--- a/.github/workflows/ci-frontend.yml
+++ b/.github/workflows/ci-frontend.yml
@@ -49,17 +49,17 @@ jobs:
       - name: Run format
         run: npm run format-check
 
-      - run: npm run test -- --testLocationInResults --json --outputFile=coverage/report.json
-      - uses: ArtiomTr/jest-coverage-report-action@v2
-        with:
-          coverage-file: coverage/report.json
-          test-script: npm test
-          working-directory: ./frontend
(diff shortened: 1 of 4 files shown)


prasnava (approved):
👍

mdragon (approved):
👍🏻

chay-REIsys (approved):

PR #10652, opened 2026-05-28
[unticketed]: update a11y workflow to rm puppeteer cache

prasnava:
## Summary

Updated a11y workflow file to remove puppeteer cache before install. 

## Validation steps

The pa11y workflow should succeed in the pr.


mdragon (approved):
👍🏻

PR #10654, opened 2026-05-28
[Issue #10653] Fix GetSubmissionList ExpandedFilter schema

### LINKED ISSUE #10653: SOAP/Proxy: Fix GetSubmissionList schema
### Summary

The schema incorrectly sets `FilterValue` as a `list` when it should be `str | int`. This does not match the WSDL.

### Acceptance criteria

- [ ] `FilterValue` is typed correctly

jakobpederson:
## Summary

Fixes / Work for #10653  

## Changes proposed

- Remove `list` typing from the `GetSubmissionList` schema

## Context for reviewers

This was not a correct typing and could possibly be causing or obscuring some issues.

## Validation steps

- [ ] typing is correct and filters perform correctly

### DIFF
--- a/api/src/legacy_soap_api/grantors/schemas/get_submission_list_schemas.py
+++ b/api/src/legacy_soap_api/grantors/schemas/get_submission_list_schemas.py
@@ -52,28 +52,22 @@ class GetSubmissionListResponse(BaseSOAPSchema):
 
 
 class ExpandedApplicationFilter(BaseModel):
-    filter_value: list[str | int] = Field(default_factory=list, alias="FilterValue")
+    filter_value: str | int = Field(alias="FilterValue")
     filter_type: str = Field(alias="FilterType")
 
     @field_validator("filter_value", mode="before")
     @classmethod
-    def convert_grants_gov_tracking_numbers_to_int(cls, v: Any, info: Any) -> Any:
-        return [
-            (
-                int(item.split("GRANT")[1])
-                if isinstance(item, str) and item.startswith("GRANT")
-                else item
-            )
-            for item in v
-        ]
+    def convert_grants_gov_tracking_numbers_to_int(cls, v: Any) -> Any:
+        if isinstance(v, str) and v.startswith("GRANT"):
+            return int(v.split("GRANT")[1])
+        return v
 
 
 def update_consolidated(consolidated: dict, filter_type: str, filter_value: str | list) -> dict:
-    value = filter_value if isinstance(filter_value, list) else [filter_value]
     filter_item = ExpandedApplicationFilter.model_validate(
-        {"FilterType": filter_type, "FilterValue": value}
+        {"FilterType": filter_type, "FilterValue": filter_value}
     )
-    consolidated[filter_item.filter_type].extend(filter_item.filter_value)
+    consolidated[filter_item.filter_type].append(filter_item.filter_value)
     return consolidated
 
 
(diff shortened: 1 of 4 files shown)


api/src/legacy_soap_api/grantors/schemas/get_submission_list_schemas.py
```diff
@@ -52,28 +52,22 @@ def to_soap_envelope_dict(self, operation_name: str) -> dict:
...
 
 
 def update_consolidated(consolidated: dict, filter_type: str, filter_value: str | list) -> dict:
-    value = filter_value if isinstance(filter_value, list) else [filter_value]
     filter_item = ExpandedApplicationFilter.model_validate(
-        {"FilterType": filter_type, "FilterValue": value}
+        {"FilterType": filter_type, "FilterValue": filter_value}
     )
-    consolidated[filter_item.filter_type].extend(filter_item.filter_value)
+    consolidated[filter_item.filter_type].append(filter_item.filter_value)
```
chouinar: Does this still need to be a list? Was it that you could technically pass in multiple of the same filter like this (in JSON):
```
[
  {"filter_type": "abc", "filter_value": 123},
  {"filter_type": "abc", "filter_value": 456},
]
```

jakobpederson: Right now that's how the filter is consumed. We could change it up but it would be more involved than what we did here.

chouinar: I'm fine if we make it into a field -> list here if that makes it simpler later, I didn't mean to say to change this, was checking which made sense.

jakobpederson: I think it does makes sense if only because it makes things more readable in a space where things can get convoluted.

api/src/legacy_soap_api/grantors/services/get_submission_list_response.py
```diff
@@ -158,46 +158,66 @@ def get_submissions(
                         Application.application_status == simpler_status,
                     )
         # Each one of these filters is Last One Wins so if multiple of the same type are entered the last one is the only one that matters
-        if grants_gov_tracking_numbers := submission_filters.get("GrantsGovTrackingNumber"):
+        grants_gov_tracking_numbers = [
+            d.filter_value for d in submission_filters if d.filter_type == "GrantsGovTrackingNumber"
+        ]
```
chouinar: Since we need this pattern in a few places, and it's a bit clunky, what if instead we added a function to the `ConsolidatedFilter` class to hide it away? Something like:

```py
def get_filter_values(self, filter_type: str) -> list[ExpandedApplicationFilter]:
     return [f.filter_value for f in self.filters if f.filter_type == filter_type]
```

That way here you'd just do:
```
grants_gov_tracking_numbers = request.expanded_application_filters.get_filter_values("GrantsGovTrackingNumber")
```
Basically having the simpler looking pattern that you had before.

api/src/legacy_soap_api/grantors/services/get_submission_list_response.py
```diff
@@ -158,46 +158,66 @@ def get_submissions(
...
+        if grants_gov_tracking_numbers:
             stmt = stmt.where(
                 [
                     ApplicationSubmission.legacy_tracking_number == tracking_number
                     for tracking_number in grants_gov_tracking_numbers
                 ][-1]
             )
-        if cfda_numbers := submission_filters.get("CFDANumber"):
+        cfda_numbers = [d.filter_value for d in submission_filters if d.filter_type == "CFDANumber"]
+        # if cfda_numbers := submission_filters.get("CFDANumber"):
```
chouinar: Should this be commented out or removed?

jakobpederson: Good catch,it should be removed.

chouinar (approved):

PR #10657, opened 2026-05-28
[unticketed] - Refactoring client fetch to next js fetch from award rec risks handlers

chay-REIsys:
## Summary

unticketed

## Changes proposed

refactoring client fetchers to pre defined  next js fetcher method
## Context for reviewers

not to pass token separately instead use predefined fetchers to do so

## Validation steps

make sure regression testing for award recommendation risks.


frontend/src/app/api/award-recommendations/[id]/risks/[riskId]/handler.test.tsx
```diff
@@ -1,13 +1,18 @@
-import * as sessionModule from "src/services/auth/session";
-import * as fetcherModule from "src/services/fetch/fetchers/awardRecommendationFetcherClient";
+import * as fetcherModule from "src/services/fetch/fetchers/awardRecommendationFetcher";
```
Comment (doug-s-nava) [inline frontend/src/app/api/award-recommendations/[id]/risks/[riskId]/handler.test.tsx:1]: [non blocking] the better way to do this is to mock the fetcher, that way you don't need to mock the session utils. You can see how it's being done here as an example https://github.com/HHS/simpler-grants-gov/blob/main/frontend/src/app/api/user/saved-searches/list/route.test.ts

frontend/src/services/fetch/fetchers/awardRecommendationFetcher.test.ts
```diff
@@ -16,17 +18,24 @@ const mockJson = jest.fn().mockResolvedValue({
...
     mockFetchAwardRecommendation(params) as Promise<Response>,
-  fetchAwardRecommendationWithMethod: (_type: "POST" | "PUT") =>
-    mockFetchAwardRecommendationWithMethod,
+  fetchAwardRecommendationWithMethod: (
+    type: "POST" | "PUT" | "DELETE",
+  ): jest.Mock => {
+    mockInnerFetch.mockResolvedValue({
+      ok: true,
+      json: jest.fn().mockResolvedValue({
+        data: type === "POST" ? mockAwardRecommendationSubmissions : null,
```
doug-s-nava, line 32: [non-blocking] in a case like this where you want the mock to behave differently for each test I find it easier to define the mock implementation within that test rather than globally, see https://github.com/HHS/simpler-grants-gov/blob/main/frontend/src/services/fetch/fetchers/organizationsFetcher.test.ts as an example

frontend/src/services/fetch/fetchers/awardRecommendationFetcher.test.ts
```diff
@@ -114,3 +123,33 @@ describe("getAwardRecommendationSubmission", () => {
...
+    expect(mockInnerFetch).toHaveBeenCalled();
+  });
+});
+
+describe("deleteAwardRecommendationRisk", () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it("deletes risk successfully", async () => {
```
doug-s-nava, line 149: this doesn't actually assert that anything was deleted, more that the function returns the expected data and calls the mock. Can you update the name?

frontend/src/services/fetch/fetchers/awardRecommendationFetcher.ts
```diff
@@ -65,3 +67,36 @@ export const getAwardRecommendationSubmission = async (
     ) ?? null
   );
 };
+
+// Client-safe fetchers for award recommendation risks
```
doug-s-nava: can delete this comment

doug-s-nava (approved):
✅ ✅ ✅ ✅

PR #10665, opened 2026-05-28
[Issue #9454] delete unused frontend components

doug-s-nava:
## Summary

Work for #9454 

## Changes proposed

Deletes any components identified as either unused or not necessary during the audit from https://github.com/HHS/simpler-grants-gov/issues/9453

## Context for reviewers

Note that some consuming files were updated slightly to move away from dependency on an unnecessary component

## Validation steps

1. General regression test


freddieyebra (approved):
Looks good. Visual regression looks good also

PR #10674, opened 2026-05-29
[Issue #10512] Move DB client and base DB to grants_shared

### LINKED ISSUE #10512: Move DB client to grants_shared
### Summary

We want to move all functionality in this directory https://github.com/HHS/simpler-grants-gov/tree/main/api/src/adapters/db to our grants shared repo including its tests.

We also want to move [base.py](https://github.com/HHS/simpler-grants-gov/blob/main/api/src/db/models/base.py) except for the ApiSchemaTable which should be left alone.

This will be a few steps, all outlined in Guide on how to move code from API to grants_shared: https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/3159818413/How+to+move+code+from+API+to+grants_shared

### Acceptance criteria

- [ ] DB client and base DB moved to grants_shared

chouinar:
## Summary
Fixes #10512

## Changes proposed
* Move db client, and base DB table to grants_shared
* Setup connection to the local shared DB recently setup for API+Analytics

## Context for reviewers
Note, these were copied as-is with the following adjustments:
* The DB configuration requires we specify the schemas that the DB has as we remap DB schemas during unit tests. In the API code, this is an enum, but it'll vary by derived service. I moved the value to an env var that we'll setup in each service (will add to API when I make it use this code)
* The base DB class had a `copy` function for copying an object, we do not ever use it, and I didn't want to commit to supporting it without investigating a bit further in whether it's the right way to do it.
* Added a few extra tests for DB functionality

Importantly, we don't have DB migrations setup for the shared code, but can still setup DB tables for tests. We don't want these models available outside of the code, so they're defined IN our tests. 

## Validation steps
Unit tests work, able to interact with the DB


kkrug (dismissed):

chouinar:
@kkrug - Merged my other approved branch into this one so they both will get merged to main together so that the files get moved rather than delete + create in the git history.

kkrug (approved):

PR #10675, opened 2026-05-29
[Issue #9715] Create an endpoint for getting a presigned url for uploading a file by a user

### LINKED ISSUE #9715: Create an endpoint for getting a presigned url for uploading a file by a user
### Summary
https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2948792998/Tech+Spec+File+Upload+for+Large+Files+Virus+Scanning

We want to create `POST /v1/files`. This endpoint will handle generating a presigned url for a user to upload a file. This is the entrypoint to all of our file scanning work.

#### Request

```py
{
  "file_name": "example.txt", # String, limit it to say 100 characters max
  "mime_type": "plain/text" # String
}
```

#### Response

```py
{
  "message": "Success",
  "data": {
    "url": "https://...",
    "body": { # Just make this a mapping, we can add an example later if we want this to be vaguely real looking
      ...
    },
    "pending_file_id": "<uuid>"
  }
}
```

#### Auth

API JWT Auth or API User Key auth are allowed. There's not AuthZ because a user is allowed to upload files, but some of the logic will restrict a user from uploading endless files.

#### Logic

First validate the following:
* Query the `pending_file` table for the count of files they've uploaded in the past hour. If this is more than 100, error (429 status code - "Too Many Requests")

For these errors, we want to make sure the error is machine readable. We should make sure to a validation error object to the `raise_flask_error` call which says which scenario they hit.

Assuming that's all good, we can process the file:
* Adjust the file name using the `get_secure_file_name` file util.
* Create a record in the `pending_file` table 
* Generate a presigned POST URL - see the [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2948792998/Tech+Spec+File+Upload+for+Large+Files+Virus+Scanning#Presigned-URLs) for specific details.
  * We should include a param of `{"IfNoneMatch": "*"}` which should prevent a user from uploading to a presigned URL that already has a file.
* Create a record in the DynamoDB table with the `file_id`, `user_id` and `status` (pending at this point).
* Return the result

### Acceptance criteria

- [ ] Endpoint created - if this is the first files endpoint created, make sure it's hidden behind a env var feature flag. We don't want this enabled in training/prod yet.
- [ ] Tests added
- [ ] Verified the endpoint works locally and that you can use the presigned url for uploading files to our local mock s3

kkrug:
## Summary

Fixes #9715 

## Changes proposed

- Updated file_routes.py tp add `POST /v1/files` route
- Updated file_schemas.py to add `CreatePresignedUploadRequestSchema`, `CreatePresignedUploadResponseSchema`, and `PresignedUploadDataSchema` schemas
- Created create_presigned_upload.py to house the service that rate-limits, secures the file name, generates the presigned POST URL, writes the `pending_file row`, generates the presigned POST URL via the new `file_util.pre_sign_upload` helper, and writes the DynamoDB scan record
- Updated file_util.py to add a reusable `pre_sign_upload` helper that builds the presigned POST URL + body
- Updated validation_constants.py to add `PENDING_FILE_UPLOAD_LIMIT_EXCEEDED` validation error type for the 429 response
- Updated conftest.py to add `mock_file_scan_s3_bucket` fixture and wired `FILE_SCAN_BUCKET` into the `s3_config` fixture
- Created test_create_presigned_upload.py to test the new service layer

## Context for reviewers

This is the endpoint for the file upload with virus scanning flow that generates a presigned s3 POST URL the caller uses to upload directly to the scanning bucket. Per-user limit is 100 pending files per hour, configurable via env, returning a 429.

## Validation steps

Confirm tests pass. 
To test locally:
- Hit `POST /v1/files` with a JWT or API key + `{"file_name": "x.txt", "mime_type": "text/plain"}`
- POST the returned body (with a file attached under the file field) to the returned url
- Verify the object lands in `s3://local-mock-file-scan-bucket/unscanned/<pending_file_id>/<file_name>`, a `pending_file` row exists with `file_scan_status = complete`, and a DynamoDB row exists with matching `file_id` / `user_id` / `status`

### DIFF
--- a/api/src/api/files_v1/file_routes.py
+++ b/api/src/api/files_v1/file_routes.py
@@ -46,7 +46,6 @@ def create_presigned_upload_route(db_session: db.Session, json_data: dict) -> re
             db_session=db_session,
             user=user,
             request_data=json_data,
-            dynamodb_client=DynamoDBClient(),
         )
 
     add_extra_data_to_current_request_logs({"pending_file_id": result.pending_file_id})
--- a/api/src/api/files_v1/file_routes.py
+++ b/api/src/api/files_v1/file_routes.py
@@ -13,10 +13,52 @@ from src.adapters.aws.dynamodb_adapter import DynamoDBClient
 from src.api.files_v1.file_blueprint import file_blueprint
 from src.auth.api_user_key_auth import api_user_key_auth
 from src.auth.multi_auth import jwt_or_api_user_key_multi_auth
+from src.services.files.create_presigned_upload import create_presigned_upload
 from src.services.files.stream_file_scan_results import stream_file_scan_results
 from src.services.files.update_pending_file_scan_status import update_pending_file_scan_status
 
 logger = logging.getLogger(__name__)
+
+
+@file_blueprint.post("")
+@file_blueprint.input(file_schemas.CreatePresignedUploadRequestSchema, location="json")
+@file_blueprint.output(file_schemas.CreatePresignedUploadResponseSchema)
+@file_blueprint.doc(
+    summary="Create a presigned upload URL",
+    description=(
+        "Generate a presigned s3 URL the caller can POST a file to. Creates a "
+        "pending file record and a DynamoDB row tracking the scan status. The "
+        "caller is responsible for performing the multipart POST to the "
+        "returned URL with the returned body fields and the file attached."
+    ),
+    responses=[200, 401, 429],
+)
+@file_blueprint.auth_required(jwt_or_api_user_key_multi_auth)
+@flask_db.with_db_session()
+def create_presigned_upload_route(db_session: db.Session, json_data: dict) -> response.ApiResponse:
+    user = jwt_or_api_user_key_multi_auth.get_user()
+    add_extra_data_to_current_request_logs({"user_id": user.user_id})
+    logger.info("POST /v1/files")
+
+    with db_session.begin():
+        db_session.add(user)
+        result = create_presigned_upload(
+            db_session=db_session,
+            user=user,
+            request_data=json_data,
+            dynamodb_client=DynamoDBClient(),
+        )
+
+    add_extra_data_to_current_request_logs({"pending_file_id": result.pending_file_id})
+
+    return response.ApiResponse(
+        message="Success",
+        data={
+            "url": result.url,
+            "body": result.body,
+            "pending_file_id": result.pending_file_id,
+        },
+    )
 
 
 @file_blueprint.post("/<uuid:pending_file_id>")
--- a/api/src/services/files/create_presigned_upload.py
+++ b/api/src/services/files/create_presigned_upload.py
@@ -10,7 +10,7 @@ from sqlalchemy import func, select
 
 import src.adapters.db as db
 import src.util.file_util as file_util
-from src.adapters.aws import S3Config, get_boto_session, get_s3_client
+from src.adapters.aws import S3Config
 from src.adapters.aws.dynamodb_adapter import DynamoDBClient, DynamoDBConfig
 from src.api.response import ValidationErrorDetail
 from src.api.route_utils import raise_flask_error
@@ -21,12 +21,6 @@ from src.util.env_config import PydanticBaseEnvConfig
 from src.validation.validation_constants import ValidationErrorType
 
 logger = logging.getLogger(__name__)
-
-
-# 2 GB - max file size we allow for uploads. The presigned POST policy enforces
-# this at the s3 layer so the user can't bypass it by sending a larger file.
-_MAX_FILE_UPLOAD_SIZE_BYTES = 2 * 1024 * 1024 * 1024
-_MIN_FILE_UPLOAD_SIZE_BYTES = 1
 
 
 class PresignedUploadConfig(PydanticBaseEnvConfig):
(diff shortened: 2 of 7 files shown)


api/src/api/files_v1/file_routes.py
```diff
@@ -13,12 +13,54 @@
...
+    add_extra_data_to_current_request_logs({"user_id": user.user_id})
+    logger.info("POST /v1/files")
+
+    with db_session.begin():
+        db_session.add(user)
+        result = create_presigned_upload(
+            db_session=db_session,
+            user=user,
+            request_data=json_data,
+            dynamodb_client=DynamoDBClient(),
```
chouinar: Can we leave the client to be fetched inside `create_presigned_upload`?

api/src/services/files/create_presigned_upload.py
```diff
@@ -0,0 +1,214 @@
...
+from src.db.models.user_models import User
+from src.util.env_config import PydanticBaseEnvConfig
+from src.validation.validation_constants import ValidationErrorType
+
+logger = logging.getLogger(__name__)
+
+
+# 2 GB - max file size we allow for uploads. The presigned POST policy enforces
+# this at the s3 layer so the user can't bypass it by sending a larger file.
+_MAX_FILE_UPLOAD_SIZE_BYTES = 2 * 1024 * 1024 * 1024
```
chouinar: We have an env var for this in our AppConfig (not sure why it's there, could move), can we reuse that so that we have a single place where we define this?

api/src/services/files/create_presigned_upload.py
```diff
@@ -0,0 +1,214 @@
...
+    user_id: uuid.UUID,
+    mime_type: str,
+) -> dict[str, Any]:
+    """Generate a presigned POST URL + body for uploading a single file to s3.
+
+    The returned dict has ``url`` and ``fields`` keys per boto3's
+    ``generate_presigned_post``. We pass ``IfNoneMatch=*`` to ensure callers
+    can't reuse a presigned URL to overwrite an existing s3 object.
+    """
+    s3_client = get_s3_client(s3_config, get_boto_session())
```
chouinar: If the session passed into `get_s3_client` is null, it does `get_boto_session()`, can leave that out for simplicity.

api/src/services/files/create_presigned_upload.py
```diff
@@ -0,0 +1,214 @@
...
+        pending_file_id=pending_file_id,
+        user=user,
+        file_name=file_name,
+        file_location=s3_file_location,
+        mime_type=mime_type,
+        file_scan_status=FileScanStatus.PENDING,
+    )
+    db_session.add(pending_file)
+    # Flush so the row is in place before we write to DynamoDB.
+    db_session.flush()
```
chouinar: Flush doesn't write to the DB, it just tells SQLAlchemy to make the query to the DB, but this all happens in a transaction so it's not committed yet.

api/src/services/files/create_presigned_upload.py
```diff
@@ -0,0 +1,214 @@
...
+        },
+        Conditions=[
+            ["content-length-range", _MIN_FILE_UPLOAD_SIZE_BYTES, _MAX_FILE_UPLOAD_SIZE_BYTES],
+            {"Content-Type": mime_type},
+            {"x-amz-meta-file-id": str(pending_file_id)},
+            {"x-amz-meta-user-id": str(user_id)},
+            {"IfNoneMatch": "*"},
+        ],
+        ExpiresIn=s3_config.presigned_s3_duration,
+    )
```
chouinar: I think we should move this to our file_util with the other presigning functions we have.

I'd probably make the metadata fields something we pass into that function, and then this function handles defining those.

Also probably worth mentioning what some of these fields/conditions do in a comment. If I didn't describe them on the tech spec, here's some info:
* Fields end up as metadata on the s3 object, we set the file ID + user ID so the scanning logic knows what it has to call the API with later.
* Conditions restrict what a user can upload, we have to include the fields, and we restrict them to whatever we know the value is (so a user doesn't change it).
* Length range is min/max length
* `IfNoneMatch` prevents a user from uploading over a file they already uploaded.

api/tests/src/api/files_v1/test_create_presigned_upload.py
```diff
@@ -0,0 +1,290 @@
...
+        client,
+        user,
+        user_auth_token,
+        aws_mocks,
+        monkeypatch,
+    ):
+        # Use a smaller limit so the test stays fast
+        monkeypatch.setenv("PENDING_FILE_UPLOAD_RATE_LIMIT", "3")
+        for _ in range(3):
+            factories.PendingFileFactory.create(user=user)
```
chouinar:
```suggestion
        factories.PendingFileFactory.create_batch(size=3, user=user)
```

api/tests/src/api/files_v1/test_create_presigned_upload.py
```diff
@@ -0,0 +1,290 @@
...
+        db_session,
+        monkeypatch,
+    ):
+        monkeypatch.setenv("PENDING_FILE_UPLOAD_RATE_LIMIT", "2")
+        monkeypatch.setenv("PENDING_FILE_UPLOAD_RATE_WINDOW_HOURS", "1")
+
+        old_pending_file = factories.PendingFileFactory.create(user=user)
+        # Backdate the row past the rate-limit window so it doesn't count
+        old_pending_file.created_at = datetime_util.utcnow() - timedelta(hours=2)
+        db_session.commit()
```
chouinar: You can set any field when using factories, even created_at/updated_at despite us not defining them in the factory/never usually setting that ourselves in actual code.

```suggestion
        # Backdate the row past the rate-limit window so it doesn't count
        old_pending_file = factories.PendingFileFactory.create(user=user, created_at = datetime_util.utcnow() - timedelta(hours=2))
```

api/src/api/files_v1/file_routes.py
kkrug: Updated

api/src/services/files/create_presigned_upload.py
kkrug: Updated, kept it in AppConfig for now

api/src/services/files/create_presigned_upload.py
kkrug: Updated

api/src/services/files/create_presigned_upload.py
kkrug: Updated

api/src/services/files/create_presigned_upload.py
kkrug: Moved to file_util and updated

api/tests/src/api/files_v1/test_create_presigned_upload.py
kkrug: Updated both instances

api/tests/src/api/files_v1/test_create_presigned_upload.py
kkrug: Updated

chouinar (approved):

PR #10678, opened 2026-05-29
[Issue #10512] Switch API to use grants_shared for DB client + DB base

### LINKED ISSUE #10512: Move DB client to grants_shared

chouinar:
Note that this will be merged into https://github.com/HHS/simpler-grants-gov/pull/10674 before going to main, that way it should recognize that we moved files rather than deleted them when merging to main.

## Summary
Fixes #10512

## Changes proposed
* Upgrade dependency version of grants_shared which includes DB client + base 
* Changed imports to be from grants_shared (~99% of the changes)
* Added a new env var that defines the DB schemas - to go along with a change to how we set those up in PostgresDbConfig
* Slight renaming/adjustment to where ApiSchemaTable is defined since the rest of the base file moved

## Context for reviewers
Like 99% of this is import changes, I'm sorry for the size - we use the db in a lot of places.

## Validation steps
Tests still work, which should catch any import issues, but also verified a few more:
* `make console` - can use the factories and interact with the db
* DB migrations still work as well as the seed script
* `make run-logs` works after you rebuild the image with `make build`. Am able to query multiple endpoints


api/pyproject.toml
```diff
@@ -46,7 +46,7 @@ dependencies = [
     "urllib3>=2.7.0,<3",
     "python-statemachine>=3.0.0,<4",
     "beautifulsoup4>=4.14.3,<5",
-    "grants-shared>=0.2.1",
+    "grants-shared==0.2.3",
```
kkrug, line 49: Just wanted to confirm that the switch from `>=` to `==` was intentional

kkrug (approved):

api/pyproject.toml
chouinar: Yeah, want it to be specifically a version rather than someone upgrade it unintentionally, at least for the moment as we keep incrementally moving chunks over

Issue #9975, opened 2026-04-30
Backend - Modify GET Award Recommendation Endpoint to add summary amounts for award recommendation submissions


nathan-stricker:
### Summary

See [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Calculate-summary-amounts-for-award-recommendation-submissions-and-add-to-GET-award-recommendation-endpoint)

On the main landing page, we want to add data that looks like the following to the GET award recommendation endpoint:
```json
{
  "data": {
    "award_recommendation_id": "",
    # other existing fields
    # ...
    # New stuff we want to add
    "award_recommendation_summary": {
       "total_received_count": 10,
       "recommended_for_funding_count": 5,
       "recommended_without_funding_count": 5,
       "not_recommended_count": 5,
       "total_recommended_amount": 1000,
    }
  }
}
```

These values need to be calculated. Because a given award recommendation may have hundreds of submissions, we want to do all this math in database queries. Naively, these queries would be as simple as:
```SQL
select sum(recommended_amount) from api.award_recommendation_submission_detail
  join api.award_recommendation_application_submission on api.award_recommendation_submission_detail.award_recommendation_submission_detail_id = api.award_recommendation_application_submission.award_recommendation_submission_detail_id
  where award_recommendation_type_id = X and award_recommendation_id = 'y';
```
Leaving out the filter for award recommendation type for the total values.

It may be possible to combine some of these into a single query, if this isn’t particularly complicated, I’d recommend doing that to avoid several separate queries, but if it’s a very complex query, we are fine with a few queries.

Once we’ve fetched these values, add them to the award recommendation we return. Can do the following to add an arbitrary item to the DB object (will not affect the DB in any way):
```python
setattr(award_recommendation, "award_recommendation_summary", {...})
```

### Acceptance criteria

- [ ] Endpoint includes summary data
- [ ] Tests created/updated as necessary

kevlyu:
@chay-REIsys to research this ticket and maybe seek clarification with Michael C

PR #10680, opened 2026-05-29
[Issue #9975] Modify GET Award Recommendation endpoint to include summary data

nathan-stricker:
## Summary

Work for #9975 

## Changes proposed

- Added `get_award_recommendation_summary` to aggregate submission counts and total recommended amount in a single DB query
- Extended `GET /alpha/award-recommendations/:award_recommendation_id` to include `award_recommendation_summary` on the response
- Added tests for populated and empty summary cases
- Removed the hardcoded summary fallback in `awardRecommendationFetcher.ts` so the UI uses API data

## Context for reviewers

See [tech spec](https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/2735046679/Tech+Spec+-+Award+Recommendation+Data+Model+Initial+Endpoints#Calculate-summary-amounts-for-award-recommendation-submissions-and-add-to-GET-award-recommendation-endpoint)

Note that "Total available" is still a frontend placeholder (`totalAvailable = 250000` in `RecommendationSummaryDisplay.tsx`) and is not part of this API change.

## Validation steps

- [ ] Confirm `award_recommendation_summary` values match submission data for an award recommendation
- [ ] Confirm empty award recommendations return zero counts and "0" for total amount
- [ ] Review `test_award_recommendation_get.py` summary tests
- [ ] Verify Recommendations summary on the award recommendation view/edit pages shows live data for the five API-backed fields


api/src/services/award_recommendations/get_award_recommendation_summary.py
```diff
@@ -0,0 +1,91 @@
...
+from src.constants.lookup_constants import AwardRecommendationType
+from src.db.models.award_recommendation_models import (
+    AwardRecommendationApplicationSubmission,
+    AwardRecommendationSubmissionDetail,
+)
+
+
+def get_award_recommendation_summary(
+    db_session: db.Session, award_recommendation_id: uuid.UUID
+) -> dict[str, int | Decimal]:
```
chouinar: Rather than return a dict, could we have this return an object? Just makes it easier to know what we're working with - dataclasses are good for this.

api/src/api/award_recommendations_alpha/award_recommendation_schemas.py
```diff
@@ -187,6 +187,30 @@ class AwardRecommendationReviewSchema(Schema):
...
+    not_recommended_count = fields.Integer(
+        metadata={"description": "Number of submissions not recommended"},
+    )
+    total_recommended_amount = fields.Decimal(
+        as_string=True,
+        metadata={
+            "description": "Sum of recommended amounts across all submissions",
+            "example": "250000.00",
+        },
+    )
```
chouinar, line 211: With the complexity of the query, we may want to do a bit of testing to see how much time it takes to calculate the values on something with like 1000 submissions. If that ends up being a fairly long calculation (100ms+), it might be better to separate this into a separate endpoint, that way we aren't calculating this value anytime we're doing `GET /award recommendation` which is probably going to be in several places that don't need this.

api/src/services/award_recommendations/get_award_recommendation_summary.py
```diff
@@ -0,0 +1,91 @@
...
+                        (
+                            AwardRecommendationSubmissionDetail.award_recommendation_type
+                            == without_funding_type,
+                            1,
+                        ),
+                        else_=0,
+                    )
+                ),
+                0,
+            ).label("recommended_without_funding_count"),
```
chouinar: Since this ends up very tabbed-over due to how much nesting there is, might be worth making a function to wrap each of these chunks? Something like:

```py
def get_coalesced_count(award_recommendation_type, label):
     # I'm not going to fix the indenting in this example, but it'd be a bit further over now
     return func.coalesce(
                func.sum(
                    case(
                        (
                            AwardRecommendationSubmissionDetail.award_recommendation_type
                            == award_recommendation_type,
                            1,
                        ),
                        else_=0,
                    )
                ),
                0,
            ).label(label),
```

Then here you'd just need:
```py
get_coalesced_count(AwardRecommendationType.RECOMMENDED_WITHOUT_FUNDING, "recommended_without_funding_count"),
```

api/src/api/award_recommendations_alpha/award_recommendation_schemas.py
nathan-stricker: Added a test to check performance to make sure request stays under 100ms for 1000 submissions, so decided to keep the summary data under the same endpoint for now. Let me know if you think this should be modified at all!

chouinar: Testing here would need to be manual, a unit test isn't going to be able to mimic the real situation here, it would need to be an e2e test that runs in staging, not locally. Why?

* A unit test (or any local manual test for that matter) doesn't have any latency with the DB, you're hitting an in-memory DB rather than having to make a network call and deal with network transfer times.
* Slow queries are often due to the data you're not fetching. Since in this case we're effectively doing several `select * where x = y`, if there's a lot of data to filter on (eg. a lot of other award recs) then that's where the slowness 

This is a case that it's probably fine right now, but as the system scales, it might get bad. Different query, but we recently had a similar issue with the GET opportunity endpoint. Adding other data to the database caused the query to jump from ~75ms to ~10000ms on average.

Not saying we need to do anything about it now, but @aaronbrown-nava - you should be aware of this and keep an eye on it as you build the system out. I'd recommend making this into a separate endpoint, but it might not be necessary to do that yet. Even if the latency is ~100 from this call on average, that's already doubling the call time for this endpoint (most of our standard GET endpoints are ~50-100).

api/tests/src/services/award_recommendations/test_get_award_recommendation_summary.py
```diff
@@ -0,0 +1,86 @@
...
+# If this query exceeds 100ms at scale, consider moving summary to a dedicated endpoint
+# so GET /award-recommendations/:id stays fast for callers that do not need it.
+LARGE_SUBMISSION_COUNT = 1000
+MAX_SUMMARY_QUERY_SECONDS = 0.1
+
+
+@pytest.fixture
+def award_recommendation(enable_factory_create):
+    opportunity = OpportunityFactory.create()
+    return AwardRecommendationFactory.create(opportunity=opportunity)
```
chouinar: You can just do `AwardRecommendationFactory.create()` - the factory already creates an opportunity.

Most of our factories create other required data, unless you need to set those sub factories up differently, it's fine to leave it out.

api/tests/src/services/award_recommendations/test_get_award_recommendation_summary.py
```diff
@@ -0,0 +1,86 @@
...
+        )
+
+        assert result.total_received_count == 0
+        assert result.recommended_for_funding_count == 0
+        assert result.recommended_without_funding_count == 0
+        assert result.not_recommended_count == 0
+        assert result.total_recommended_amount == Decimal("0")
+
+
+class TestGetAwardRecommendationSummaryPerformance:
```
chouinar: Separate from the issue I described elsewhere, this unit test takes over 3 minutes to run right now, probably all for creating the award rec submission records. That's way too much time for one unit test.

api/src/api/award_recommendations_alpha/award_recommendation_schemas.py
nathan-stricker: Ah yeah that makes sense, thanks for the explanation. I'll leave the summary data on this endpoint for now but will make sure to sync up with Aaron when he's back from PTO to make a plan for this going forwards!

api/tests/src/services/award_recommendations/test_get_award_recommendation_summary.py
nathan-stricker: Removed the test since it wasn't doing anything meaningful anyway, thanks again Michael!

chouinar (approved):

Issue #10636, opened 2026-05-27
SOAP/Proxy: lower serial_number in setup-cert-user command


jakobpederson:
### Summary

Lower the serial_number when converting tcertificates to LegacyCertificate. This is to help standardize the certificates to make it easier to search for.

### Acceptance criteria

- [ ] command, when run, will lower case the serial_numbers

PR #10706, opened 2026-06-01
[Issue #10636] Lower values for tcerts and LegacyCerts

jakobpederson:
## Summary

Fixes / Work for #10636  

## Changes proposed

- add validator for `certemail` on tcertificates model 
- add validator for `serial_num` on tcertificates model
- add validataor for `serial_number` on legacycertificate model

These validators lower the strings going into these fields.

## Context for reviewers

The serial numbers can have a upper cased values so we wanted to remove that. I added that check for certemail too.

## Validation steps

- [ ] values are lowered when models are created


api/tests/src/task/certificates/test_setup_cert_user_task.py
```diff
@@ -48,10 +50,42 @@ def test_setup_cert_user_raises_error_if_no_tcertificate(
...
+        SERIAL_NUM = "ABC"
+        CERTEMAIL = "XYZ@TEST.COM"
+        tcert = staging.certificates.Tcertificates(
+            currentcertid="1234",
+            expirationdate=date(2024, 12, 31),
+            serial_num=SERIAL_NUM.upper(),
+            created_date=date(2024, 12, 30),
+            creator_id="1235",
+            certemail=CERTEMAIL.upper(),
+        )
```
kkrug: Could we use the `StagingTcertificatesFactory` here instead?

api/src/db/models/legacy_mixin/certificates_mixin.py
```diff
@@ -30,3 +30,11 @@ class TcertificatesMixin:
...
     serial_num: Mapped[str | None]
     system_name: Mapped[str | None]
+
+    @validates("serial_num")
+    def validate_serial_num(self, key: str, value: str | None) -> str | None:
+        return value.lower() if value else value
+
+    @validates("certemail")
+    def validate_certemail(self, key: str, value: str) -> str:
+        return value.lower()
```
chouinar: I'm not sure I want to do this on the SQLAlchemy model itself, I'd have a lot of questions about when this runs during the SQLAlchemy lifecycle, and I think it's easier to not have to solve for that.

In the code that writes to the table, lowercase it. If we want to fix existing data, make a DB migration.

api/src/data_migration/load/load_oracle_data_task.py
```diff
@@ -139,6 +142,13 @@ def do_insert(self, foreign_table: sqlalchemy.Table, staging_table: sqlalchemy.T
                 extra=log_extra | {"excluded_columns": excluded_columns},
             )
 
+        lowercase_columns = self.lowercase_columns.get(foreign_table.name, [])
+        if excluded_columns:
```
kkrug: Would it make more sense for this to be `if lowercase_columns` instead?

api/src/task/certificates/setup_cert_user_task.py
```diff
@@ -105,7 +105,7 @@ def process_cert_user(
             agency=agency,
             cert_id=tcertificate.currentcertid,
             expiration_date=valid_expiration_date,
-            serial_number=tcertificate.serial_num,
+            serial_number=tcertificate.serial_num.lower() if tcertificate.serial_num else None,
```
kkrug, line 108: Does this need the `if` conditional here? `setup_cert` checks
```
        if not tcertificate.serial_num:
            logger.warning("Tcertificate is missing serial number")
            return SetupCertUserTaskStatus.TCERTIFICATE_IS_MISSING_SERIAL_NUMBER
```

api/src/db/migrations/versions/2026_06_02_lower_case_serial_num.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def upgrade():
+    op.execute(
+        "UPDATE staging.tcertificates SET serial_num = LOWER(serial_num), certemail = LOWER(certemail)"
+    )
+
+
+def downgrade():
+    pass
```
hao10282025-sudo: overall looks good to me
a quick question is this downgrade "pass" auto-generated? I wonder if it should be downgrade to the previous "no-lower" mode

jakobpederson: I thought we needed to include it but I'm not sure we do. UPPER-ing everything really wouldn't actually be that worthwhile.

api/src/task/certificates/setup_cert_user_task.py
jakobpederson: It's a typing thing, it thrwos an error if I don't account for a None situation even if it won't happen.

api/src/db/migrations/versions/2026_06_02_lower_case_serial_num.py
hao10282025-sudo: ok no worries then

api/src/data_migration/load/load_oracle_data_task.py
```diff
@@ -55,6 +55,9 @@ def __init__(
         # Note: This does not work on UPDATES and can cause errors because the fields won't be excluded
         # if the column has invalid data
         self.columns_to_exclude: dict[str, list[str]] = {"tcertificates": ["is_selfsigned"]}
+        self.lowercase_columns: dict[str, list[str]] = {
```
chouinar: Please do not change this code, this is not where we would want this change, load oracle data copies data exactly as it is from grants.gov, it does not do transformations, that is breaking the separation of these processes entirely.

chouinar (changes requested):
Let's step back and discuss this. This change has moved significantly from its starting point and is now changing code that is way upstream of where it should be.

The staging tables should be unmodified by any of these changes. If you want to lowercase values, do that when copying over from the staging table to the actual table we use, not before.

api/src/data_migration/load/load_oracle_data_task.py
jakobpederson: I'll roll this back. I misunderstood and thought we needed to update the staging table.

api/src/task/certificates/setup_cert_user_task.py
chouinar: Could also do a `cast` instead if you want to tell the type checker you know better.

chouinar (approved):

PR #10728, opened 2026-06-01
[Unticketed] Adjust windows setup to not require manual commands

chouinar:
## Changes proposed
* Adjust bash commands run as part of setup to use `#!/usr/bin/env sh` which should work better with windows
* Add a `.gitattributes` file and tell it to automatically handle line ending differences between windows and unix on checkout/checkin

## Context for reviewers
Right now our setup for windows users requires running commands to change permissions on files. I looked into how to do it better, and this seems to be the recommended approach from what I could find.

For reference, I looked at what [uv](https://github.com/astral-sh/uv) does and basically stole its approach (ontop of the other references I found to doing it this way). UV is used a lot, it probably sorted these issues out a while ago.

There are additional things we could do like an `.editorconfig` and also adjusting other bash files, but I'm just trying to get the scope narrowly focused on the ones needed to build on the backend as those are all local-only + impact development. I also intend to move most of those scripts soon as we refactor our local setup.

## Validation steps
I can't test it, but I'm getting windows users to test that it works.

### DIFF
--- a/documentation/api/windows-setup.md
+++ b/documentation/api/windows-setup.md
@@ -7,11 +7,9 @@ This guide will help you set up the Simpler Grants.gov API for local development
 1. Install WSL 2 with Ubuntu (not docker-desktop)
 2. Install Docker Desktop and enable Ubuntu integration
 3. Install `make` and `postgresql-client` in Ubuntu
-4. Navigate to the `api` directory
-5. Fix shell script line endings: `sed -i 's/\r$//' bin/*.sh && chmod +x bin/*.sh`
-6. Run `make init` (takes 5-10 minutes first time)
-7. Run `make start`
-8. Verify with `curl http://localhost:8080/health`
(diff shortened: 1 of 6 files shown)


documentation/api/windows-setup.md
```diff
@@ -7,11 +7,9 @@ This guide will help you set up the Simpler Grants.gov API for local development
 1. Install WSL 2 with Ubuntu (not docker-desktop)
 2. Install Docker Desktop and enable Ubuntu integration
 3. Install `make` and `postgresql-client` in Ubuntu
-4. Navigate to the `api` directory
```
hao10282025-sudo, line 10: this old No.4 step should be kept?
either to setup api or any other folder when applicable

hao10282025-sudo (approved):

dghazvini (approved):

Issue #10521, opened 2026-05-20
Move the lookup registry functionality from API to grants_shared


chouinar:
### Summary

We want to move the functionality defined in https://github.com/HHS/simpler-grants-gov/tree/main/api/src/db/models/lookup to our grants_shared repo.

This will be a few steps, all outlined in Guide on how to move code from API to grants_shared: https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/3159818413/How+to+move+code+from+API+to+grants_shared

NOTE: We do not want to move the `_sync_roles` function in sync_lookup_values, at least not yet. Leave that alone for now. We also do not need to move any of the actual lookup tables defined in https://github.com/HHS/simpler-grants-gov/blob/main/api/src/db/models/lookup_models.py - those are domain specific and won't be moved (well, some will move to grantor, but that's a later problem).

### Acceptance criteria

- [ ] Moved lookup registry to grants_shared

Issue #10515, opened 2026-05-20
Move API schema utils and generic functionality from API to grants_shared


chouinar:
### Summary

This will be a few steps, all outlined in Guide on how to move code from API to grants_shared: https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/3159818413/How+to+move+code+from+API+to+grants_shared

We want to move the following code:
* Everything in https://github.com/HHS/simpler-grants-gov/tree/main/api/src/api/schemas
* Do not move the `shared_schema.py` file, that's specific to the API's code, and shouldn't be moved.
* ValidationErrorType should not be moved over, instead we should make a new constants class with all the generic values on the grants_shared side, but make any functions/variables that take in this enum instead just be a string. That way it's reusable by APIs without them needing to define specific error codes in a separate python package.
* Some of the tests for this code is in https://github.com/HHS/simpler-grants-gov/blob/main/api/tests/src/api/test_route_error_format.py which tests in the context of an APIFlask configured route. We likely need some refactoring here to do much since we're not going to have some of the functionality it is depending on from create_app generified in a way we can use here.

### Acceptance criteria

- [ ] Moved most of the schemas folder to grants_shared
- [ ] Adjusted schemas to not directly use the ValidationErrorType type - create a new one
- [ ] Fixed tests to still test schema

PR #10731, opened 2026-06-01
[Issue #10408, #10515, #10521] Move generic utils, API schema utils, and lookup registry to grants_shared

### LINKED ISSUE #10408: Move generic utils from API to grants_shared
### Summary

We want to move several of our generic utilities from the API to grants_shared. This will be a few steps, all outlined in Guide on how to move code from API to grants_shared: https://navasage.atlassian.net/wiki/spaces/Grantsgov/pages/3159818413/How+to+move+code+from+API+to+grants_shared

For this ticket we want to move the following functionality:
* https://github.com/HHS/simpler-grants-gov/blob/main/api/src/util/api_key_gen.py
* https://github.com/HHS/simpler-grants-gov/blob/main/api/src/util/decimal_util.py
* https://github.com/HHS/simpler-grants-gov/blob/main/api/src/util/dict_util.py
* https://github.com/HHS/simpler-grants-gov/blob/main/api/src/util/string_utils.py

These should all transfer over uneventfully.

### Acceptance criteria

- [ ] Moved code over to shared repo and created a release after merging
- [ ] Updated version in API code and removed code previously copied
- [ ] All tests on both sides work, no routes or other functionality are broken by this change or require any non-import related modification to work.

chouinar:
## Summary
Fixes #10408, #10515, #10521

## Changes proposed
* Moved several generic utils over to grants_shared ( api key gen, decimal util, dict util, string util)
* Moved API schema utils over to grants_shared (all the marshmallow schema  extensions / error handling changes we've made)
* Moved the lookup logic over to grants_shared
* Added a handful of tests

## Context for reviewers
This is 3 tickets, but doing each of these moves requires a lot of little steps, so I decided to group these 3 tickets together.

Will have a second part of this PR out soon to move the API to use all of it once this is approved. Will not merge this until I've merged the API changes as well.

## Validation steps
Tests all work, added a few and had to rewrite some of the lookup table ones to not rely on API DB tables.


api/src/db/models/lookup/sync_lookup_values.py
```diff
@@ -30,6 +30,7 @@ def sync_lookup_values(db_client: PostgresDBClient | None = None) -> None:
         for table, lookup_config in sync_values.items():
             _sync_lookup_for_table(table, lookup_config.get_lookups(), db_session)
 
+        # TODO - need to move this somewhere else
```
chouinar: In the 2nd part of the PR, I'll handle this

kkrug (dismissed):

api/src/api/schemas/extension/field_validators.py
```diff
@@ -163,18 +163,14 @@ def _format_error(self, value: _T, message: str) -> list[MarshmallowErrorContain
         # The method this overrides returns a string, but we'll modify it to return one of
         # our error containers instead which works, but MyPy doesn't like.
 
-        is_min = False
-        is_max = False
-        if self.min is not None or self.max_inclusive is not None:
-            is_min = True
-        if self.max is not None or self.max_inclusive is not None:
-            is_max = True
```
hao10282025-sudo, line 171: can we double check this update is fine (replace with new line 166 and 167)? not quite sure what is exactly checked before but looks like the logic is simplified
There looks also a bug for this removed block, line 168 and line 170 are both checking "**self.max_inclusive**"

chouinar: Yes, I did change this to fix a bug I noticed in copying it over.

Before if you set a minimum and no maximum, it would always give you the `min_or_max` error rather than just min. Noticed it when I added a handful of additional unit tests and they didn't work.

Note that we only even have a handful of fields that even use it yet, and I double-checked that our frontend never refers to it.

backend/grants_shared/src/grants_shared/api/schemas/extension/field_validators.py
hao10282025-sudo, line 1: file shows as renamed/moved 👍

kkrug (approved):

PR #10734, opened 2026-06-02
[Issue #10038] Add an ECS task job lock to our load-transform job in dev/staging

### LINKED ISSUE #10038: Add an ECS task job lock to our load-transform job in dev/staging
### Summary

To test that our ECS task job locks work, we want to add it to our load transform job.

TODO - exact configuration depends on some other work in the quad that may have lowered the time. Assuming everything is as it currently works:
* Lock should encompass ALL of the tasks
* Lock should be configured to hold the lock for 90 minutes

We'll need to enable the job lock env var in dev/staging - we won't enable it in training/prod yet, and even though we've added the lock, it shouldn't do anything.

### Acceptance criteria

- [ ] Job lock added to the load-transform job
- [ ] Job lock only enabled in dev/staging

kkrug:
## Summary

Fixes #10038  

## Changes proposed

- Updated load_transform.py to wrap load, transform, set-current, and store-version in a single `TaskJobLock` context manager
- Updated dev.tf to set `ENABLE_JOB_LOCK = "true"`
- Updated staging.tf to set `ENABLE_JOB_LOCK = "true"`

## Context for reviewers

This is the first usage of the `TaskJobLock` context manager. It wraps the full load-transform job (load → transform → set-current → store-version) with a 90-minute hold. `ENABLE_JOB_LOCK` is set at the environment level for dev/staging only.

## Validation steps

After dev/staging deployment:
- Terraform plan shows `ENABLE_JOB_LOCK=true` on the load-transform task in dev/staging, and no env var diff in training/grantee1/grantee2/prod
- In dev, trigger the job and confirm a row appears in `api.job_lock` for `job_type = "load-transform"` and the `"Job lock released"` log line fires with `job_lock_enabled: true`
- Trigger a second run while the first is in flight and confirm it fails with `TaskJobLockIsLockedError`

### DIFF
--- a/infra/api/app-config/env-config/scheduled_jobs.tf
+++ b/infra/api/app-config/env-config/scheduled_jobs.tf
@@ -117,6 +117,26 @@ locals {
     grantee2 = "ENABLED"
     prod     = "ENABLED"
   }
+  # Only enable the job lock in dev/staging so we can validate it before
+  # turning it on in training/prod.
+  load-transform-environment-vars = {
+    dev = [
+      {
+        Name  = "ENABLE_JOB_LOCK"
+        Value = "true"
+      }
+    ]
+    staging = [
+      {
+        Name  = "ENABLE_JOB_LOCK"
+        Value = "true"
+      }
+    ]
+    training = null
+    grantee1 = null
+    grantee2 = null
+    prod     = null
+  }
   sam-extracts-state = {
     dev      = "ENABLED"
     staging  = "ENABLED"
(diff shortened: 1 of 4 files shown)


infra/api/app-config/env-config/scheduled_jobs.tf
```diff
@@ -117,6 +117,26 @@ locals {
...
+      {
+        Name  = "ENABLE_JOB_LOCK"
+        Value = "true"
+      }
+    ]
+    training = null
+    grantee1 = null
+    grantee2 = null
+    prod     = null
+  }
```
chouinar: I think we should just enable/disable the job lock globally for the environments. We've only attached the job lock to a single job, and don't really intend to do it in others yet, so it ends up the same. But if we do add it to the other jobs, don't want to be confused why it doesn't work for them.

kkrug: Makes sense, updated

prasnava (approved):
Verified the terraform plan against dev. 

```
Plan: 2 to add, 20 to change, 2 to destroy.
```

chouinar (approved):

Issue #9456, opened 2026-04-08
Clean up from x-sgg token refactor


doug-s-nava:
### Summary

After merging https://github.com/HHS/simpler-grants-gov/pull/9431 there will be some clean up to do:

* any fetch functions created since the refactor PRs were opened will need to be updated to use the new pattern
* based on the decision made by the team, change should be made to handle requests lacking tokens in a universal way

### Acceptance criteria

- [ ] any fetch functions created since the refactor PRs were opened will need to be updated to use the new pattern
- [ ] based on the decision made by the team, change should be made to handle requests lacking tokens in a universal way

PR #10737, opened 2026-06-02
[Issue #9456] auth token cleanup

doug-s-nava:
## Summary

Work for #9456 

Blocked by https://github.com/HHS/simpler-grants-gov/pull/10736

## Changes proposed

Following up on work from https://github.com/HHS/simpler-grants-gov/pull/10736, this change 
* removes unnecessary checking for session tokens in cases where fetch functionality will naturally handle the case downstream
* standardizes handling of errors thrown from requests that are missing tokens
* cleans up some fetch handler functions

## Validation steps

1. `npm run local` on this branch
2. For each page on the site that requires authentication, visit the page as a logged in user, then log out
3. _VERIFY_: an `Unauthenticated` message is displayed

#### Bonus Points

For each action on the site that requires authentication:

1. visit the page where the action takes place
2. log in
3. manually remove the "session" cookie from your browser session
4. perform the action
5. _VERIFY_: the correct handling of the unauthenticated action is shown

### DIFF
--- a/frontend/src/app/api/applications/[applicationId]/transfer-ownership/handler.ts
+++ b/frontend/src/app/api/applications/[applicationId]/transfer-ownership/handler.ts
@@ -1,5 +1,4 @@
 import { readError } from "src/errors";
-import { getSession } from "src/services/auth/session";
 import { addOrganizationToApplication } from "src/services/fetch/fetchers/addOrganizationToApplication";
 
 import { NextResponse } from "next/server";
@@ -13,11 +12,6 @@ export const transferOwnershipHandler = async (
   options: { params: Promise<{ applicationId: string }> },
 ): Promise<Response> => {
   try {
-    const session = await getSession();
-    if (!session?.token) {
-      return NextResponse.json({ error: "Unauthorized" }, { status: 401 });
-    }
-
     const { applicationId } = await options.params;
 
     let parsedBody: TransferOwnershipRequestBody;
--- a/frontend/src/app/api/user/saved-searches/list/handler.ts
+++ b/frontend/src/app/api/user/saved-searches/list/handler.ts
@@ -1,16 +1,8 @@
-import { readError, UnauthorizedError } from "src/errors";
-import { getSession } from "src/services/auth/session";
+import { readError } from "src/errors";
 import { fetchSavedSearches } from "src/services/fetch/fetchers/savedSearchFetcher";
 
 export const listSavedSearches = async (_request: Request) => {
   try {
-    const session = await getSession();
-    if (!session || !session.token) {
-      throw new UnauthorizedError(
-        "No active session to fetch saved opportunities",
-      );
-    }
-
     const savedSearches = await fetchSavedSearches();
 
     return new Response(JSON.stringify(savedSearches), {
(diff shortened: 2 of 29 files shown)


frontend/src/services/fetch/fetchers/apiKeyFetcher.ts
```diff
@@ -1,43 +1,11 @@
 import "server-only";
 
-import { getSession } from "src/services/auth/session";
 import { fetchUserWithMethod } from "src/services/fetch/fetchers/fetchers";
-import { ApiKey } from "src/types/apiKeyTypes";
-import { APIResponse } from "src/types/apiResponseTypes";
-
-// this is a duplicate - resolve with handleListApiKeys
-export const fetchApiKeys = async (): Promise<ApiKey[]> => {
```
doug-s-nava, line 9: this was deleted, and usage of the function replaced with `handleListApiKeys`. The two functions do the same thing but were typed slightly differently, thus the slight changes to typing in API key fetching scattered around here

frontend/src/app/api/user/saved-searches/list/handler.ts
```diff
@@ -1,16 +1,8 @@
...
 
 export const listSavedSearches = async (_request: Request) => {
   try {
-    const session = await getSession();
-    if (!session || !session.token) {
-      throw new UnauthorizedError(
-        "No active session to fetch saved opportunities",
-      );
-    }
-
```
nathan-stricker, line 13: Can you explain the reasoning for removing this check? I'm not super familiar with saved searches, but it seems like this would result in a 200 response for an unauthenticated user

frontend/src/app/api/applications/[applicationId]/transfer-ownership/handler.ts
```diff
@@ -13,11 +12,6 @@ export const transferOwnershipHandler = async (
   options: { params: Promise<{ applicationId: string }> },
 ): Promise<Response> => {
   try {
-    const session = await getSession();
-    if (!session?.token) {
-      return NextResponse.json({ error: "Unauthorized" }, { status: 401 });
-    }
-
     const { applicationId } = await options.params;
```
Comment (nathan-stricker) [inline frontend/src/app/api/applications/[applicationId]/transfer-ownership/handler.ts:15]: It looks like `MissingAuthError` doesn't have `status = 401` by default - wouldn't that make removal of checks like these default to 500?

nathan-stricker (commented):
Left a few comments- also looks like there are some conflicts that need to be resolved

frontend/src/app/api/user/saved-searches/list/handler.ts
doug-s-nava: In this case `fetchSavedSearches` will return an empty array if the user isn't logged in, which I think is the expected behavior

frontend/src/app/api/applications/[applicationId]/transfer-ownership/handler.ts
Comment (doug-s-nava) [reply frontend/src/app/api/applications/[applicationId]/transfer-ownership/handler.ts:15]: good question - since in the system a request never gets made at all, there's no status code to think about. The error is thrown before we make the request

nathan-stricker (approved):
Approved!

PR #10738, opened 2026-06-02
[Issue #10408, #10515, #10521] Use generic utils, API schema utils, and lookup registry from grants_shared

### LINKED ISSUE #10408: Move generic utils from API to grants_shared

chouinar:
This will be merged into https://github.com/HHS/simpler-grants-gov/pull/10731 before merging to main.

## Summary
Fixes #10408, #10515, #10521

## Changes proposed
* Changed API code to use generic utils, API schema utils, and lookup registry from the grants_shared code
* Slight adjustment to syncing lookup values as the role portion of the logic wasn't moved
* Added a new `ApiLookupTable` class that adds the api schema to the lookup tables as that was removed when moving it to grants_shared
* Adjusted docs to use correct import paths/links

## Context for reviewers
Another set of code moved over, puts us around halfway. Same as the prior PRs, mostly just changing imports beyond what I called out above.

## Validation steps
Verified DB migrations still work as I changed some lookup logic. Lookup tables & roles still get populated as expected. Also didn't cause anything to change from Alembic's perspective about the lookup table changes, it doesn't see the tables any differently.

API still works locally, can query it uneventfully.


api/src/db/models/lookup/sync_lookup_values.py
```diff
@@ -19,51 +18,15 @@ def sync_lookup_values(db_client: PostgresDBClient | None = None) -> None:
...
-    with db_client.get_session() as db_session, db_session.begin():
-        sync_values = LookupRegistry.get_sync_values()
-
-        for table, lookup_config in sync_values.items():
-            _sync_lookup_for_table(table, lookup_config.get_lookups(), db_session)
+    grants_shared_lookup.sync_lookup_values(db_client)
 
-        # TODO - need to move this somewhere else
+    with db_client.get_session() as db_session, db_session.begin():
         _sync_roles(db_session)
```
kkrug, line 27: Could these two syncs being in separate transactions potentially cause an issue if one raises an exception?

chouinar: Can't think of a scenario where it would. While we do need to run the lookup value sync first since some of the roles use those lookup values, there shouldn't be any requirement that they're in the same session. Not 100% on this, but under the hood when we sync the roles to the DB, we're converting the lookup values to integers anyways (how they're represented in the DB), so all this gets enforced at the DB.

If either part raises an exception then it just fails there - shouldn't be an issue if the role sync failed but we already did the lookup sync.

kkrug (approved):

